PR #3197, opened 2018-03-31
Update chinese help texts for Server command

BramChen:
* support environment variables for specifying port number


Jermolene:
Thanks @BramChen

PR #3203, opened 2018-04-06
draggable widget: actions on drag-start and drag-end

BurningTreeC:
this allows actions on drag-start and drag-end

why: it allows selective/conditional applying of "pointer-events: none" on certain elements. Those elements can be "blocking elements" like iframes. Or droppable widgets that contain other droppable widgets, where one can either apply conditionals on the action-string to prevent undesired actions from one or more  droppable widgets, or apply different classes to the droppable div when another droppable should be triggered... and I can imagine many other things

### DIFF
--- a/core/modules/widgets/draggable.js
+++ b/core/modules/widgets/draggable.js
@@ -52,6 +52,8 @@ DraggableWidget.prototype.render = function(parent,nextSibling) {
 		domNode: domNode,
 		dragTiddlerFn: function() {return self.getAttribute("tiddler");},
 		dragFilterFn: function() {return self.getAttribute("filter");},
+		dragStartActions: self.dragStartActions,
+		dragEndActions: self.dragEndActions,
 		widget: this
 	});
 	// Insert the link into the DOM and render any children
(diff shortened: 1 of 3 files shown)


BurningTreeC:
Hi @Jermolene - is this considerable when I add the docs?

core/modules/widgets/draggable.js
```diff
@@ -67,6 +69,8 @@ DraggableWidget.prototype.execute = function() {
 	// Pick up our attributes
 	this.draggableTag = this.getAttribute("tag","div");
 	this.draggableClasses = this.getAttribute("class");
+	this.dragStartActions = this.getAttribute("startactions");
```
Jermolene: I think we should use "startActions", "endActions" for consistency with e.g. LinkCatcher widget.

Jermolene:
Hi @BurningTreeC 

> Hi @Jermolene - is this considerable when I add the docs?

Yes, looks good.

core/modules/widgets/draggable.js
BurningTreeC: Ok! :+1:

BurningTreeC:
@Jermolene 

done , it's ready to merge :smiley:

Jermolene:
Thanks @BurningTreeC

PR #3206, opened 2018-04-07
Add hooks for canceling editing a tiddler, new tiddler and renaming a…

inmysocks:
… tiddler

These hooks are used by the multi-user plugin and are required to make the live two-way updating between the file system and the browser


Jermolene:
Thanks @inmysocks

pmario:
@inmysocks see: th-renaming-tiddler - double use - with different return values #4023

PR #3209, opened 2018-04-11
Add data attributes to ViewTemplate

pmario:
Examples + new and updated docs can be found at [2737.tiddlyspot.com](http://2737.tiddlyspot.com/#How%20to%20apply%20custom%20styles:%5B%5BHow%20to%20apply%20custom%20styles%5D%5D%20%5B%5BCustom%20styles%20by%20user-class%5D%5D)

* Adds a `class` field to the ViewTemplate. Similar to the `color` field for tag-pills
* Adds a `data-tags` attribute, which can be used for styling. .. See docs 
* Deprecates `tc-tagged-*` because of usability problems. 
* Adds description for how to use `data-tiddler-title` for styling
* changes 3 x `<$set` to 1x `<$vars`


twMat:
Q: if I have some tweaks for the docs, how would I propose these? @pmario , if I make a copy on tiddlyspot, would you take a look at it and transfer it if you agree?

pmario:
It depends how big the changes are. ... Just open the file here and github should create a new branch, that you can edit. ... Once finished, create a PR against my repo. I'll merge it, if we can agree ;)

Jermolene:
Thanks @pmario @BurningTreeC

PR #3212, opened 2018-04-12
dynaview unset functionality when scrolling out

BurningTreeC:
I obviously lied :grin:  - found the time

This sets a specified unset-value to a specified unset-tiddler if the element has triggered and scrolls out of view

### DIFF
--- a/plugins/tiddlywiki/dynaview/dynaview.js
+++ b/plugins/tiddlywiki/dynaview/dynaview.js
@@ -138,7 +138,37 @@ function checkVisibility() {
 	var elements = document.querySelectorAll(".tc-dynaview-set-tiddler-when-visible");
 	$tw.utils.each(elements,function(element) {
 		// Bail if we've already triggered this element
-		if(element.getAttribute("data-dynaview-has-triggered") === "true" && element.getAttribute("data-dynaview-unset-tiddler") === undefined) {
+		if(element.getAttribute("data-dynaview-has-triggered") === "true") {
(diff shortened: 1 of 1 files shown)


plugins/tiddlywiki/dynaview/dynaview.js
```diff
@@ -139,6 +139,36 @@ function checkVisibility() {
...
+					elementRect.right < viewportRect.left || 
+					elementRect.top > viewportRect.bottom ||
+					elementRect.bottom < viewportRect.top) {
+					// Set the tiddler value
+					var tiddler = element.getAttribute("data-dynaview-unset-tiddler"),
+						value = element.getAttribute("data-dynaview-unset-value") || "";
+					if(tiddler && $tw.wiki.getTiddlerText(tiddler) !== value) {
+						$tw.wiki.addTiddler(new $tw.Tiddler({title: tiddler, text: value}));
+					}
+					element.setAttribute("data-dynaview-has-triggered",false);
```
Jermolene: For consistency with the test in line 141 perhaps this should be set to the string `"false"`.

Jermolene:
Terrific thanks @BurningTreeC. I've left one minor comment; could you kindly also update the docs?

plugins/tiddlywiki/dynaview/dynaview.js
BurningTreeC: @Jermolene - I'm reading "false" here - do I have to be worried about dyslexia? ;)

BurningTreeC:
@Jermolene - ready!

PR #3214, opened 2018-04-15
add swipe,tap and press widget to hammerjs

BurningTreeC:
this is a try to add a swipe widget to the hammerjs plugin

it's supposed to trigger actions and/or a popup state when a swipe gesture is detected on an element that is specified by its class attribute

it's called like this: `<$swipe targets="tc-targetclass" actions=<<myactions>>><div class="tc-targetclass"/></$swipe>`

it accepts attributes:

- targets ... the css class of the target element(s ... multiple elements allowed)
- actions ... actions that get triggered when gesture is recognized
- popup ... like the button widget, a popup state tiddler
- pointers ... 1(default) to I-don't-know-how-many pointers , 2 and 3 do work so I guess 10
- velocity ... the velocity value that must be reached to trigger (default 0.1 - appears to be a good value)
- direction ... allows to specify left/up/right/down - if specified, doesn't trigger on swipes in other directions
- threshold ... minimal distance needed to travel with pointer before recognizing the gesture

I'm adding this so that we can maybe discuss it and correct what's necessary
the swipe widget is one of the simpler ones I made and the swipe gesture adds many possibilities

Simon


Jermolene:
Thanks @BurningTreeC

BurningTreeC:
Hi @Jermolene , thank you very much for merging this!

I guess you are aware that this may still need some work?

the refresh mechanism(s) needs a second look - I'm not really getting how it works and I'm just trying to cover many different cases. I guess I'm not covering all

It's a bit hard to understand what needs to be checked ... I'm talking about lines 40 - 53
Anyone who understands what I'm trying to accomplish is invited to give this a look

This is some tinkering that became somehow necessary because the target html elements often change because of another hammer widget doing something at the same time - and I haven't figured the correct tiddly way of handling that


Simon

Jermolene:
Hi @BurningTreeC 

> I guess you are aware that this may still need some work?

No, I hadn't realised (and didn't give it the scrutiny that I would for a core component). Not to worry, perhaps the best thing would be to revert these commits, and keep the core hammer plugin as just a means to share the hammer.js library, and then experiment with these widgets in a different, non-core plugin?

BurningTreeC:
> perhaps the best thing would be to revert these commits, and keep the core hammer plugin as just a means to share the hammer.js library, and then experiment with these widgets in a different, non-core plugin?

yes, that would be it

Jermolene:
Great thanks @BurningTreeC -- I'm very pleased to see you experimenting in this area, a truly touch-centric TiddlyWiki 5 remains of great interest to me.

BurningTreeC:
thanks @Jermolene 

are there some resources, docs or examples for advanced widget-refreshing somewhere?
That would be a great help

I've put this set of widgets on http://hammerwidgets.tiddlyspot.com
There are docs and some examples, too

Jermolene:
> that the widget refreshes its state when it cannot be rendered because the involved dom nodes are changing because of another widget working on them

Broadly, a widget cannot interact with DOM nodes other than those that it creates itself. As you've probably discovered, breaking that rule causes things to fail in subtle, intermittent ways.

BurningTreeC:
could I add the created domNodes to $tw.hammerwidget.domNodes and refresh if they change?

BurningTreeC:
... Now thinking about it it's a bad idea

... but I can access the parent widget from all the hammer widgets and check for their domNodes if I add them with this.domNodes.push()

what would be a check for those domNodes if they're changed or undefined?

PR #3217, opened 2018-04-21
simplified hammer widgets

BurningTreeC:
the widgets now create the domnode that gets hammer-enabled

that simplifies how they work and should keep possible errors low

demo and docs on http://hammerwidgets.tiddlyspot.com

BurningTreeC


BurningTreeC:
@Jermolene 

I see - how can we organize further to get them into shape for a future tw plugin and would you intend going on with them? Should I make a repo for them on my account?

BurningTreeC:
@Jermolene

edit:

I see - how can we organize further to get them into shape for a future tw plugin and/or how would you intend going on with them? Should I make a repo for them on my account?

PR #3224, opened 2018-04-25
Fix a typo in my name on the prerelease

inmysocks:
So it points to the correct GitHub user.


Jermolene:
Thanks @inmysocks, good catch!

PR #3227, opened 2018-04-25
Add a mechanism allowing the sync adaptor to notify the syncer of changes

Arlen22:
This PR adds a mechanism allowing the sync adapter to notify the syncer of change notifications from the server as they happen. I have added the skeleton code to the filesystem and tiddlyweb adapters, and implemented the handling code in syncer.js. 

The sync adapter emits a `server-update` event with type and tiddlerfields (a skinny tiddler) arguments whenever it feels like it. The syncer then handles the event as it would a normal skinny tiddler. 

I've also added a delete tiddler mechanism but it is incomplete and someone needs to make sure it works correctly. This all needs testing -- I haven't even tested it yet myself. But I got the idea and it seemed simple, so I did it. This didn't take very long to put together.

### DIFF
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -211,6 +211,13 @@ Syncer.prototype.syncFromServer = function() {
 		}
 		
 		this.syncadaptor.getSkinnyTiddlers(function(err,tiddlers) {
+			//get an array of the titles being updated
+			var titles = tiddlers.map(e => e.title);
+			//get the titles in the tiddlerInfo hashmap that are not on the server
+			var removes = Object.keys(self.tiddlerInfo).filter(e => titles.indexOf(e) === -1);
+			//assume a server delete and handle it locally
+			removes.forEach(title => self.handleServerDelete(title));
+			
 			// Trigger the next sync
 			self.pollTimerId = setTimeout(function() {
 				self.pollTimerId = null;
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -33,6 +33,13 @@ function Syncer(options) {
 	var self = this;
 	this.wiki = options.wiki;
 	this.syncadaptor = options.syncadaptor;
+	this.syncadaptor && this.syncadaptor.addEventListener && this.syncadaptor.addEventListener('server-update', function(type, tiddlerFields){
+		if(type === "update" || type === "create") { 
+			self.handleServerUpdate(tiddlerFields);
+		} else if (type === "delete") {
+			self.handleServerDelete(tiddlerFields.title);
+		}
+	});
 	this.disableUI = !!options.disableUI;
 	this.titleIsLoggedIn = options.titleIsLoggedIn || this.titleIsLoggedIn;
 	this.titleUserName = options.titleUserName || this.titleUserName;
@@ -202,6 +209,7 @@ Syncer.prototype.syncFromServer = function() {
 			clearTimeout(this.pollTimerId);
 			this.pollTimerId = null;
 		}
+		
 		this.syncadaptor.getSkinnyTiddlers(function(err,tiddlers) {
 			// Trigger the next sync
 			self.pollTimerId = setTimeout(function() {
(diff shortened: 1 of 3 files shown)


plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -13,16 +13,23 @@ A sync adaptor module for synchronising with TiddlyWeb compatible servers
 "use strict";
 
 var CONFIG_HOST_TIDDLER = "$:/config/tiddlyweb/host",
-	DEFAULT_HOST_TIDDLER = "$protocol$//$host$/";
+	DEFAULT_HOST_TIDDLER = "$protocol$//$host$/",
+    	WikiExport = require('$:/core/modules/wiki.js');
```
pmario: Why is the variable named "WikiExport"  ... There is no "export" handling going on. right?

core/modules/syncer.js
```diff
@@ -202,7 +209,15 @@ Syncer.prototype.syncFromServer = function() {
 			clearTimeout(this.pollTimerId);
 			this.pollTimerId = null;
 		}
+		
 		this.syncadaptor.getSkinnyTiddlers(function(err,tiddlers) {
+			//get an array of the titles being updated
+			var titles = tiddlers.map(e => e.title);
```
pmario: I think TW doesn't use this type of function calls in the core. ... @Jermolene .. do you allow to use "arrow functions"?

sukima: Can't use arrow functions because TW still supports ES5 browsers. Unless we run things through a transpiler. (I run my own plugins through babel but I don't think the core does).

Arlen22: Yeah, that is the code that I didn't know if it is correct or not in the handling of deletions. If the server and client sync filters are different then this will cause problems, but if they are a perfect snapshot of each other then we are fine.

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
Arlen22: The wiki.js file exports a bunch of functions, and we use three of them. So I just called it that.

danielo515:
There is an issue that should be linked to this PR

core/modules/syncer.js
```diff
@@ -33,6 +33,15 @@ function Syncer(options) {
 	var self = this;
 	this.wiki = options.wiki;
 	this.syncadaptor = options.syncadaptor;
+	if (this.syncadaptor && this.syncadaptor.addEventListener) {
```
Jermolene: Why do we make the sync adaptor be an event listener? Wouldn't it make more sense for the syncer to be an event listener?

I don't actually think it makes much sense to use the event pattern here, because the syncer and sync adaptor are tightly integrated components. Better to for the syncer to pass a reference to itself to the syncadaptor constructor, and then have the syncadaptor invoke methods on it.

core/modules/syncer.js
```diff
@@ -33,6 +33,15 @@ function Syncer(options) {
...
+		this.syncadaptor.addEventListener('server-update', function(type, tiddlerFields){
```
Jermolene: Coding standards:

* should use double quotes for string constants
* should not have spaces after the commas in parameter lists

core/modules/syncer.js
```diff
@@ -281,6 +280,38 @@ Syncer.prototype.handleLazyLoadEvent = function(title) {
 	}
 };
 
+/* 
+Handle server update event
+*/
+Syncer.prototype.handleServerUpdate = function(tiddlerFields){
```
Jermolene: The way that this method is written is expecting a skinny tiddler, is that what you want? Perhaps it would be more useful for the syncadaptor to invoke a method on the syncer that passes both the full tiddler fields and the tiddlerInfo?

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -14,16 +14,23 @@ A sync adaptor module for synchronising with the local filesystem via node.js AP
 
 // Get a reference to the file system
 var fs = $tw.node ? require("fs") : null,
-	path = $tw.node ? require("path") : null;
+	path = $tw.node ? require("path") : null,
+    	WikiExport = require('$:/core/modules/wiki.js');
```
Jermolene: Rather than reusing those three methods from wiki.js directly we should abstract them into a utility module.

core/modules/syncer.js
Arlen22: In the actual startup code, both are initialized by the "startup" startup module, and the syncadaptor is passed to the syncer constructor. The syncer constructor would have to then call a method on the syncadaptor with `this` as an argument. If that's fine, then we're good.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
Arlen22: I would rather give the syncadaptor a reference to the syncer anyway, so I'll just remove this code again.

core/modules/syncer.js
Arlen22: And yes, I extracted that code out of syncFromServer. It is correct. However, as I said, it is more likely that the sync adaptor would want to call storeTiddler.

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
pmario: I just had a look again.. IMO ``WikiExport` isn't used?!

Arlen22: You're right. I missed that when I updated the other pages per our discussion here.

Issue #3231, opened 2018-04-27
is-filter issue with 5.1.16


pmario:
IMO started with: https://github.com/Jermolene/TiddlyWiki5/pull/2982

I'll try to create a PR .... @Jermolene ... We will need a hot-fix for this one.

mklauber:
Can you be more explicit about the issue?  Is it behavior, or performance?

mklauber:
Started work on #3232

mklauber:
Some work at https://github.com/Jermolene/TiddlyWiki5/pull/3240

PR #3232, opened 2018-04-27
Fix for #3231: Is operator multiple categories

mklauber:


core/modules/filters/is.js
```diff
@@ -42,24 +42,30 @@ exports.is = function(source,operator,options) {
...
+		} else {	
+			return [$tw.language.getString("Error/IsFilterOperator")];	
+		}
+	} else { 
+		for (var t=0; t<subops.length; t++) {
+			var subop = isFilterOperators[subops[t]];
+			if(subop) {
+				filteredResults[subops[t]] = subop(source,operator.prefix,options);
+			} else {
+				return [$tw.language.getString("Error/IsFilterOperator")];
```
pmario: duplicate

core/modules/filters/is.js
```diff
@@ -42,24 +42,30 @@ exports.is = function(source,operator,options) {
...
+			}
+		}
+		for (var t=0; t<subops.length; t++) {
+			if (filteredResults[subops[t]].indexOf(title) != -1){
+				results.push(title);
```
pmario: imo creates duplicated array elements.

pmario:
The point also is: If I say `"$:/test" test  +[is[tiddler+system]]` ... do I want it to be sorted? 

I would say: yes. ... 

So Why do you want to keep the order?

core/modules/filters/is.js
mklauber: Wait, not this shouldn't work at all.  It doesn't have a variable named title.  That why I had the source function in the first version.

mklauber:
Because If I say `[tag[Test]] +[is[tiddler+shadow]]` I want to get things in the order my tag specifies.

pmario:
Tthat's a point

core/modules/filters/is.js
mklauber: I can't exactly combine that, because in one I need to test every element in a list, and the other a single item.

mklauber:
Does your is[alias] filter return results that are not part of the original list?

pmario:
Does your is[alias] filter return results that are not part of the original list?

yes ... aliases are always lowercase, but the element canbe "test, Test, TEst" ...

mklauber:
Ok, so now I understand why this is breaking for you.  I don't think I can maintain order when you're changing the titles, but I may be able to make it append anything in the output that's not in the input.  Would that work for you?

Evolena:
While you are at it, in the "is" documentation tiddler, the railroad diagram has been copied from `all` operator. For `is` operator, it should be (no final "s", more options):
[{: ("current" | "image" | "missing" |: "orphan" | "shadow" | "system" | "tag" | "tiddler" ) +"+" }]

mklauber:
Thanks @Evolena!  I'll include that in whatever fix we decide on here.

pmario:
>Ok, so now I understand why this is breaking for you. I don't think I can maintain order when you're changing the titles, but I may be able to make it append anything in the output that's not in the input. Would that work for you?

I think so. .... 

But as I wrote. `is` is a wrapper, that imo is not allowed to do additional filtering. 

---------------- 

I think error handling code can be simplified. ...

pmario:
>Every is operator except yours returns a subset of the input parameters, and so it doesn't do any additional filtering except in that instance.

IMO Calling source() does additional work, that has already been done.

pmario:
I did change the `is[alias]` function too. .... But I need to check, why I did it that way in the first place. And if there are other side effects. ... 

But I still thing the existing implementation needs improvement. 

What I'd like to see is: 

 - fast lane, if there is only 1 parmeter
   - same code as it was 
 - new lane .. 
   - new stuff.

pmario:
Why did you close it. I still think there is a problem.

mklauber:
Because I didn't think it was ready for merge, so I closed it until I got a
chance to review it. I'm planning to come back to it, no worries.

pmario:
OK :)

mklauber:
Take a look here.  https://github.com/Jermolene/TiddlyWiki5/pull/3240

PR #3234, opened 2018-04-28
keyboard widget: global shortcuts

BurningTreeC:
this allows triggering global actions with keyboard shortcuts, using the keyboard widget, setting the attribute "global" to "yes"

for testing I've tagged a tiddler $:/tags/BelowStory and put this inside:

<$keyboard key="ctrl+shift+N" actions="""<$action-sendmessage $message="tm-new-tiddler"/>""" global="yes"/>

### DIFF
--- a/core/modules/widgets/keyboard.js
+++ b/core/modules/widgets/keyboard.js
@@ -38,29 +38,47 @@ KeyboardWidget.prototype.render = function(parent,nextSibling) {
 		tag = this.tag;
 	}
 	// Create element
-	var domNode = this.document.createElement(tag);
-	// Assign classes
-	var classes = (this["class"] || "").split(" ");
-	classes.push("tc-keyboard");
-	domNode.className = classes.join(" ");
+	var domNode,
+	    classes;
+	if(this.global !== "yes") {
+		domNode = this.document.createElement(tag);
+		// Assign classes
+		classes = (this["class"] || "").split(" ");
+		classes.push("tc-keyboard");
+		domNode.className = classes.join(" ");
+	} else {
+		domNode = document.body;
+	}
 	// Add a keyboard event handler
+	var actionInvoked = null,
+	    messageInvoked = null;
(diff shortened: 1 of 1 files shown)


BurningTreeC:
Note: the tiddler with the global keyboard widget should be closed and the wiki reloaded for testing

BurningTreeC:
@Jermolene @pmario @allothers

are you interested in global keyboard shorcuts?

this PR is obviously just a quick hack of the keyboard widget and has the downside that the global-shortcuts pagetemplate must be closed and the wiki refreshed so that it doesn't add multiple event-listeners for the same key-combo, but it could be tuned right. On linux using a keyboard-driven tiling window manager like i3/sway/awesome etc this is a real big plus

Jermolene:
Apologies for not commenting on this before @BurningTreeC. I don't think it's the right way to support global keyboard shortcuts.

One simple option is to use the HTML5 `accesskey` attribute on a button; the only reason that the keyboard widget exists was originally to cope with the situations where `accesskey` couldn't be used (typically because of focus issues).

BurningTreeC:
No problem @Jermolene 

that `accesskey` attribute - is it used to only focus buttons that are visible or can it be used to dispatch wikitext-actions?

I'd like to have something where I can define a key-combo that triggers an action, like showing/hiding the sidebar, creating a new tiddler...

Jermolene:
> that accesskey attribute - is it used to only focus buttons that are visible or can it be used to dispatch wikitext-actions?

There are docs here:

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/accesskey

Pressing the key named in the accesskey attribute of a visible element will trigger that element as if it was clicked. Try visiting https://classic.tiddlywiki.com/#readOnly:no and then pressing ctrl-alt-N on a Mac (might be a different combination of modifiers on other platforms).

BurningTreeC:
Thank you, I should google it for myself next time

PR #3240, opened 2018-04-30
Refactored the is operator for simplicity and efficiency.

mklauber:


mklauber:
One note:  I'm using the Javascript `Set` object which is part of the relatively new ECMAScript 6 standard.  However, according to http://kangax.github.io/compat-table/es6/#test-Set it's supported by the majority of modern browsers.  I can refactor this if desired.

core/modules/filters/is.js
```diff
@@ -27,40 +27,44 @@ Export our filter function
 */
 exports.is = function(source,operator,options) {
 
+	var isFilterOperators = getIsFilterOperators(),
+		subops = operator.operand.split("+");
 
-	if( !operator.operand) {
-		// Return all tiddlers if the operand is missing
+	//Make sure all the operands are defined.
+	for (var t = 0; t < subops.length; t++){
```
pmario: subops.length is used 4 times. So we may copy it to it's own variable, for a minimal speed boost.

core/modules/filters/is.js
```diff
@@ -27,40 +27,44 @@ Export our filter function
...
+
+	if(subops.length === 0) { 	// Return all tiddlers if the operand is missing
 		var results = [];
 		source(function(tiddler,title) {
 			results.push(title);
 		});
 		return results;
-	}
+	} else if(subops.length === 1) {	// Shortcut the Single Operator
+		var subop = isFilterOperators[subops[0]];
```
pmario: `subop` is very close to `subops` ... So I was confused the very first time.... IMO it should be a variable name with is a bit more difference.

core/modules/filters/is.js
```diff
@@ -27,40 +27,44 @@ Export our filter function
...
-		var subop = isFilterOperators[subops[t]];
-		if(subop) {
-			filteredResults[subops[t]] = subop(source,operator.prefix,options);
-		} else {
-			return [$tw.language.getString("Error/IsFilterOperator")];
+	} else {	// Handle multiple operators
+		var filtered_results = new Set(),
+			results = [];
+		for(var t=0; t < subops.length; t++){
+			var subop = isFilterOperators[subops[t]];
```
pmario: `var subop` is already defined. ... IMO we should move the variable declaration outside the if() construct.

pmario:
Did test it with uni-link and it seems to work. ... 

I'm not sure about the new "set" function @Jermolene  ??

Jermolene:
Hi @mklauber we haven't yet adopted ECMAScript 6  for TW5; it's useful to be conservative so that we can run on old Raspberry PIs and so on.

mklauber:
Alright, I'll rework this to use keys in a object instead of a formal set.

Jermolene:
Hi @mklauber sorry to press you, but I wondered if you'd had a chance to look at this so that we can release v5.1.17? Many thanks!

mklauber:
Take a look, should be good to go now.

pmario:
Did test it with uni-link plugin and it works :)

mklauber:
Out of curiosity, when you test are you using multiple operators?  Because I can understand why it would work for you in the single case, but I think the multiple would still fail, and I don't think I can avoid that.

mklauber:
For aliases at least.

pmario:
I'm sure, you run your tests with multiple inputs and you may also test the new sorting function. right?

pmario:
... We may need a lot more automatic tests :)

core/modules/filters/is.js
```diff
@@ -27,40 +27,44 @@ Export our filter function
...
+			results = [];
+		for(var t=0; t < num_of_subops; t++){
+			var operator = isFilterOperators[subops[t]];
+			operator(source,operator.prefix,options).forEach(function(element) { filtered_results[element] = True});
 		}
-		
+
+		// Sort the output by the input (There may be a better way to do this)
+		source(function(tiddler,title) {
+			if(filtered_results[title] === True) {
```
Jermolene: Is "True" correct?

core/modules/filters/is.js
```diff
@@ -27,40 +27,44 @@ Export our filter function
...
+	} else {	// Handle multiple operators
+		var filtered_results = {},
+			results = [];
+		for(var t=0; t < num_of_subops; t++){
+			var operator = isFilterOperators[subops[t]];
+			operator(source,operator.prefix,options).forEach(function(element) { filtered_results[element] = True});
 		}
-		
+
+		// Sort the output by the input (There may be a better way to do this)
```
Jermolene: I don't understand the comment about sorting?

mklauber: I'm using the `source` function to iterate through the source list and check if the item is in the `filtered_results`.  I'm doing this so I can maintain the order of the input list when I output the matching items.  There may be a better way to get the input list, but I wasn't sure about it.  @pmario had some concerns that `source` did other things, and was possibly too expensive.

core/modules/filters/is.js
mklauber: It appears to work probably because I'm using the smae value when  I'm setting it in the for loop above, but for clarity's sake, I'll make it explicitly a string.

Jermolene:
Thanks @mklauber @pmario

pmario:
@mklauber @Jermolene  .... This seems to break the Recent-tab ... It only shows system tiddlers atm.

pmario:
@mklauber 
We need to test all the `is` appearances in the core to be sure, there are no unwanted side effects.

BurningTreeC:
this breaks many things for me - tested right now

BurningTreeC:
@mklauber 

num_of_subops may never be "0" (?)

Jermolene:
I'l look at this now

mklauber:
Well, I'm embarrassed.  If possible, I think this should be reverted.  Notable issues I've identified now that I'm looking at it with new eyes include 

a) num_of_subops may never be zero as @BurningTreeC identified.  Instead that test should be `if(operand === "")`
b) I'm pretty sure I completely lost the ability to invert this operator somewhere along the line.  I'm going to have to look into how to bring that back.

BurningTreeC:
@mklauber 
please forgive, no need to feel embarrassed. Let's just look at it and see what it takes to make it work :)

the old filter is still in [the file history](https://github.com/mklauber/TiddlyWiki5/commits/8ba1c37e9b12a455dc5e593ea9142ed2889d4839/core/modules/filters/is.js) here on github, shouldn't be complicated to revert

Jermolene:
I'm afraid that I think you're right @mklauber. I've found a number of problems now that I come to look at it in detail -- a big problem was that I didn't give any of this proper oversight, my apologies.

Issue #3239, opened 2018-04-30
package.json still defines prerelease version


TheDiveO:
... please bump up the version defined in `package.json` to `5.1.16` (stable/final).

Jermolene:
Fixed with v5.1.17 release

PR #3241, opened 2018-04-30
Bump version number from release

mklauber:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "tiddlywiki",
   "preferGlobal": "true",
-  "version": "5.1.16-prerelease",
+  "version": "5.1.16",
```
pmario: 5.1.16 is already released. So it needs to be 5.1.17-prerelease now.

mklauber: The issue is 5.1.16 was released without removing the prerelease flag.  So stable 5.1.16 lists it's version as 5.1.16-prerelease.  This is an attempt to fix that.

mklauber: See https://github.com/Jermolene/TiddlyWiki5/issues/3239

pmario: ahh .. OK. .. my bad

Jermolene:
Thanks @mklauber I'll push a fix shortly

PR #3243, opened 2018-05-01
Marked alternative Markdown parser

icarito:
Hi I don't know if this will miss some features from the original, but it's a faster lighter weight parser that allows to put raw HTML in between the Markdown, that I use a lot and was needing here. Hopefully it will be useful!


icarito:
I understand this is a plugin - is there a plugin registry somewhere?

I thought I'd share but don't know how to package plugins to be helpful
for others.

I managed to install my plugin in existing wiki by building the plugin
library and then serving a local copy of it (while changing the server
URL from https://tiddlywiki.com/library/v5.1.15/index.html to localhost).

But I can only deactivate the previously installed markdown plugin, not
remove it.

I'd really like code to be separated from data to be able to hack the
code and not worry about my wiki, any pointer welcome!

icarito:
This is working for me. The original Markdown plugin was more complex (it turned Markdown into specific `widgets`) whereas this one only makes a single "raw" widget with the HTML inside.
Is there a drawback to this? I get the advantage of being able to include some HTML snippets and have them rendered (which is part of the point of Markdown, imho).

Let me know if I should rip this from here and put it in a separate repo? Is https://tiddlywiki.com/#Resources the only registry for "community" plugins? I'd be cool if it was easy to publish them for anyone - i'm sure we'd find some jewels.

I guess if this is not desired in core, we can close this PR! Otherwise I can improve it.

Jermolene:
Hi @icarito 

> The original Markdown plugin was more complex (it turned Markdown into specific widgets) whereas this one only makes a single "raw" widget with the HTML inside.
Is there a drawback to this?

The problem with producing raw HTML from the parser is that it precludes any further processing which may matter in some contexts (eg fixing up tiddler links). But it should work in many situations.

The plugin library doesn't yet really accommodate third party plugins; it's just got the plugins that the core team is committed to maintaining. So the usual first step for a new plugin is to publish it independently and then link to it from tw.com "Resources"

icarito:
Thanks @Jermolene I really appreciate your thoughtful reply. Today I noticed the links issue so I understand what you mean. We can close this issue then and I can work on it some more.

Thanks for your work on creating and maintaining this wonderful software, @Jermolene @pmario and everyone else.

PR #3244, opened 2018-05-02
Add write condition for DAT protocol

icarito:
In order to write to a DAT with Beaker Browser, it needs to have access to the file via `dat:` protocol.

### DIFF
--- a/core/modules/savers/beaker.js
+++ b/core/modules/savers/beaker.js
@@ -51,7 +51,7 @@ BeakerSaver.prototype.info = {
 Static method that returns true if this saver is capable of working
 */
 exports.canSave = function(wiki) {
-	return !!window.DatArchive;
+	return !!window.DatArchive && location.protocol==="dat:";
 };
 
 /*


core/modules/savers/beaker.js
```diff
@@ -51,7 +51,7 @@ BeakerSaver.prototype.info = {
 Static method that returns true if this saver is capable of working
 */
 exports.canSave = function(wiki) {
-	return !!window.DatArchive;
+	return !!window.DatArchive && location.protocol==="dat:";
```
pmario: yea, but the first condition will fail in any browser except the beaker browser. IMO this is a feature test. If the API is available it should be able to use it. 

And beaker browser is in flux. ... There was a "workspace:" protocol that also made dat API available. .. So this check would have been a "false", which would have been an issue.

pmario: I don't think, that this should be merged.

Jermolene:
@pmario the problem is that Beaker provides `window.DatArchive` even to sites loaded over http; right now you get an error if you visit https://tiddlywiki.com/ in Beaker and click the "save" button

PR #3254, opened 2018-05-04
[doc] Update Sharing your tiddlers with others.tid

twMat:
Just a little polishing up.
Ideally the icons should be styled to not be so big.


Jermolene:
Thanks @twMat

PR #3259, opened 2018-05-06
[widget.js] - avoid producing negative hash-values for qualify macro

BurningTreeC:
Hi,

a simple Math.abs() in getStateQualifier() avoids qualified tiddlers of the form

`tiddlername--9843502839` <- note the double " - "

having this character twice in a qualified title causes problems with some filters

also, IF I would want to create multiple css class-names using `<<qualify>>` , such a name with double - would not be valid as class-name ... the css fails

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -229,7 +229,7 @@ Widget.prototype.getStateQualifier = function(name) {
 			}
 			node = node.parentWidget;
 		}
-		var value = $tw.utils.hashString(output.join(" ")).toString().replace(/^-/g, "0");
+		var value = $tw.utils.hashString(output.join("")).toString().replace(/^-/g, "0");
 		this.qualifiers[name] = value;
 		return value;
 	}
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -229,7 +229,7 @@ Widget.prototype.getStateQualifier = function(name) {
 			}
 			node = node.parentWidget;
 		}
-		var value = Math.abs($tw.utils.hashString(output.join(" ")));
+		var value = $tw.utils.hashString(output.join(" ")).toString().replace(/^-/g, "0");
 		this.qualifiers[name] = value;
 		return value;
 	}
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -229,7 +229,7 @@ Widget.prototype.getStateQualifier = function(name) {
 			}
 			node = node.parentWidget;
 		}
-		var value = $tw.utils.hashString(output.join(""));
+		var value = Math.abs($tw.utils.hashString(output.join(" ")));
 		this.qualifiers[name] = value;
 		return value;
 	}


BurningTreeC:
@sukima ... I'll have to dig deeper into it, I saw it coming :)

I was using the qualify macro to create classes like tc-my-class-12345 and the ones with double dashes seemed not to be valid. This was constructed in a complex macro where also some state tiddlers with or without double dashes were created ... the whole thing failed when there were double dashes.

Now I got to figure out if it's a filter problem or a parsing problem or if my discovery was wrong.
I'll check the last point in the first place, I'll report back then

core/modules/widgets/widget.js
```diff
@@ -229,7 +229,7 @@ Widget.prototype.getStateQualifier = function(name) {
 			}
 			node = node.parentWidget;
 		}
-		var value = $tw.utils.hashString(output.join(""));
+		var value = Math.abs($tw.utils.hashString(output.join(" ")));
```
pmario: IMO you can't do that. You are deliberately combining 2 different values to 1. There is a reason why 2 different strings should result in 2 different hashes.

pmario: replace the `-` with `x` or `0` and you are probably good to go

BurningTreeC: thanks @pmario - I've changed it

sukima:
>the ones with double dashes seemed not to be valid.

Still not understanding this. Who or what is telling you it is _invalid_?

sukima:
Ok. I looked into this. This is a **bug** err _feature_ in TiddlyWiki. It seems that when TW parses WikiText a `--` it converts it to `\n&emdash;\n` so that `--` represents a full emdash

screen shot 2018-05-06 at 2 54 57 pm

The problem here is that TW should not be parsing WikiText in side a `<style>` tag.

BurningTreeC:
@sukima , thanks for looking deeper into it

ah I understand, but a stylesheet tiddler of type vnd.tiddlywiki gets parsed that way, and there it should work

also, do you think the same problem with the emdash could cause issues with filters?
wouldn't a first (not really bad) intervention be to avoid the double-dashes where they could be created?

what, do you think?

sukima:
> but a stylesheet tiddler of type vnd.tiddlywiki gets parsed that way, and there it should work

These tiddler are also parsed through WikiText.

> do you think the same problem with the emdash could cause issues with filters?

Maybe. I don't know that much about the WikiText parsing to that degree. I hope someone else here would be able to chime in.

Jermolene:
Just to clarify, the problem is indeed with overzealous wikitext parsing of stylesheet tiddlers. The core stylesheets avoid the problem by explicitly restricting the parse rules right at the top:

```
\rules only filteredtranscludeinline transcludeinline macrodef macrocallinline macrocallblock
```

See https://github.com/Jermolene/TiddlyWiki5/blob/master/themes/tiddlywiki/vanilla/base.tid#L42

Jermolene:
Also don't forget the new control panel stylesheet feature for debugging stylesheet wikification. It makes it easier to check the the stylesheet text seen by the browser (ie after macro processing and other wikification)

screenshot 2018-05-08 18 27 40

BurningTreeC:
Thanks @Jermolene, I'll try those parse rules - didn't even think about using them

The Stylesheets tab is very useful, too - but I'm not yet used to having it

I'll close this PR

PR #3263, opened 2018-05-09
Update RenderCommand.tid

danielo515:
Adds a link that opens the advanced search with a filter to list all the template tiddlers.
This is instead of #3260 

Not sure if we should put it between small tags

```
,,<$button class="tc-btn-invisible" style="text-decoration:underline">
Show available rendering templates
<$action-setfield $tiddler="$:/temp/advancedsearch" text="[all[shadows]prefix[$:/core/templates/]]"/>
<$action-setfield $tiddler="$:/state/tab--1498284803" text="$:/core/ui/AdvancedSearch/Filter"/>
<$action-navigate $to="$:/AdvancedSearch"/>
</$button>,,
```


Jermolene:
Thanks @danielo515 looks good

PR #3265, opened 2018-05-10
[new-journal] don't overwrite existing journals

BurningTreeC:


BurningTreeC:
@pmario @Jermolene 

this builds on top of the new `<$wikify>` solution and uses action-createtiddler to create a UNIQUE journal - so it doesn't overwrite, and then navigates to it

core/ui/PageControls/new-journal.tid
```diff
@@ -20,4 +21,4 @@ description: {{$:/language/Buttons/NewJournal/Hint}}
 <$set name="journalTags" value={{$:/config/NewJournal/Tags}}>
 <$set name="journalText" value={{$:/config/NewJournal/Text}}>
```
pmario: IMO we should also combine all those set-widgets to 1 vars widget!

core/ui/PageControls/new-journal.tid
```diff
@@ -6,7 +6,8 @@ description: {{$:/language/Buttons/NewJournal/Hint}}
 \define journalButton()
 <$button tooltip={{$:/language/Buttons/NewJournal/Hint}} aria-label={{$:/language/Buttons/NewJournal/Caption}} class=<<tv-config-toolbar-class>>>
 <$wikify name="journalTitle" text="""<$macrocall $name="now" format=<<journalTitleTemplate>>/>""">
-<$action-sendmessage $message="tm-new-tiddler" title=<<journalTitle>> tags=<<journalTags>> text=<<journalText>>/>
+<$action-createtiddler $basetitle=<<journalTitle>> $savetitle="$:/temp/new-journal-tiddler" tags=<<journalTags>> text=<<journalText>>/>
+<$action-navigate $to={{$:/temp/new-journal-tiddler}} $scroll="yes"/>
```
pmario: IMO navigate is not enough. .. It needs to be opened for edit mode.

pmario: eg: 

`<$action-sendmessage $message="tm-edit-tiddler" $param={{$:/temp/new-journal-tiddler}}/>`

Jermolene:
Hi @BurningTreeC I've got a slightly different solution that retains the use of tm-new-tiddler. I'll commit shortly.

BurningTreeC:
@Jermolene ok, I've seen what you did :+1: 

I have one issue, maybe it's the third problem:

if one creates the journal "10th May 2018", and adds text, then creates another one the same day, your new solution still overwrites the old tiddler but doesn't delete its text which is perfectly ok - 
But if one wants to create another new journal tiddler that doesn't work and forces to create a timestamp in minutes and seconds for the new-journal titles to avoid this

I would have preferred more single journal tiddlers over this solution but it's you to decide

Jermolene:
> if one creates the journal "10th May 2018", and adds text, then creates another one the same day, your new solution still overwrites the old tiddler but doesn't delete its text which is perfectly ok -
But if one wants to create another new journal tiddler that doesn't work and forces to create a timestamp in minutes and seconds for the new-journal titles to avoid this

Are you asking for the second click of the new journal tiddler button to create a new journal tiddler? That is not functionality that people have asked for; if you want instantaneous new journals then I think using a timestamp is the usual way to work.

BurningTreeC:
> Are you asking for the second click of the new journal tiddler button to create a new journal tiddler?

yes that's what I'm asking because I understood it that way. That's my misunderstanding, I'm fine with how it's now

PR #3272, opened 2018-05-12
Add name field to server start hook

Arlen22:
This allows other server implementations (such as TiddlyServer) to provide custom support for features a plugin wishes to add by access the Node server object. The main example is WebSockets, since many data folders are loaded into TiddlyServer and therefore cannot extend the NodeJS server object. 

Even if they could, TiddlyServer uses two separate server instances. So it's better to do it like this. TiddlyServer uses the SimpleServer router so the first argument of the hook still works fine and is not affected.


Jermolene:
Thanks @Arlen22 presumably TiddlyServer invokes the same hook? This fix doesn't seem right: is the argument to add an extra parameter to indicate the entity triggering the hook being TiddlyWiki?

Arlen22:
This PR ended up being merged in another way and so this PR was not closed. 

https://github.com/Jermolene/TiddlyWiki5/commit/dd6575dd8a39f23868f7778c61ba22dea1af33a3

PR #3273, opened 2018-05-13
[button widget] add accesskey attribute

BurningTreeC:
adds the `accesskey` attribute to buttons for keyboard shortcuts

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -64,6 +64,9 @@ ButtonWidget.prototype.render = function(parent,nextSibling) {
 	if(this["aria-label"]) {
 		domNode.setAttribute("aria-label",this["aria-label"]);
 	}
+	if(this.accessKey) {
+		domNode.setAttribute("accesskey",this.accessKey);
+	}
 	// Add a click event handler
 	domNode.addEventListener("click",function (event) {
 		var handled = false;
@@ -183,6 +186,7 @@ ButtonWidget.prototype.execute = function() {
 	this.buttonTag = this.getAttribute("tag");
 	this.dragTiddler = this.getAttribute("dragTiddler");
 	this.dragFilter = this.getAttribute("dragFilter");
+	this.accessKey = this.getAttribute("accessKey");
 	// Make child widgets
 	this.makeChildWidgets();
 };
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -125,6 +125,9 @@ LinkWidget.prototype.renderLink = function(parent,nextSibling) {
 	if(this["aria-label"]) {
 		domNode.setAttribute("aria-label",this["aria-label"]);
 	}
+	if(this.accessKey) {
+		domNode.setAttribute("accesskey",this.accessKey);
+	}
 	// Add a click event handler
 	$tw.utils.addEventListeners(domNode,[
 		{name: "click", handlerObject: this, handlerMethod: "handleClickEvent"},
@@ -183,6 +186,7 @@ LinkWidget.prototype.execute = function() {
 	this.isMissing = !this.wiki.tiddlerExists(this.to);
 	this.isShadow = this.wiki.isShadowTiddler(this.to);
 	this.hideMissingLinks = ($tw.wiki.getTiddlerText(MISSING_LINK_CONFIG_TITLE,"yes") === "no");
+	this.accessKey = this.getAttribute("accessKey");
 	// Make the child widgets
 	this.makeChildWidgets();
 };
--- a/core/ui/PageControls/newtiddler.tid
+++ b/core/ui/PageControls/newtiddler.tid
@@ -3,7 +3,7 @@ tags: $:/tags/PageControls
 caption: {{$:/core/images/new-button}} {{$:/language/Buttons/NewTiddler/Caption}}
 description: {{$:/language/Buttons/NewTiddler/Hint}}
 
-<$button message="tm-new-tiddler" tooltip={{$:/language/Buttons/NewTiddler/Hint}} aria-label={{$:/language/Buttons/NewTiddler/Caption}} class=<<tv-config-toolbar-class>>>
+<$button message="tm-new-tiddler" tooltip={{$:/language/Buttons/NewTiddler/Hint}} aria-label={{$:/language/Buttons/NewTiddler/Caption}} class=<<tv-config-toolbar-class>> accessKey="n">
 <$list filter="[<tv-config-toolbar-icons>prefix[yes]]">
 {{$:/core/images/new-button}}
 </$list>
--- a/core/ui/PageControls/newtiddler.tid
+++ b/core/ui/PageControls/newtiddler.tid
@@ -3,7 +3,7 @@ tags: $:/tags/PageControls
 caption: {{$:/core/images/new-button}} {{$:/language/Buttons/NewTiddler/Caption}}
 description: {{$:/language/Buttons/NewTiddler/Hint}}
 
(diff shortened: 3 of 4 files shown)


Jermolene:
Hi @BurningTreeC I think we can merge the enhancements to the link and button widgets (although we do need documentation), but I think we should consider carefully before using this feature for the new tiddler button and the hide sidebar button.

The trouble is that the browser's built in accesskey handling doesn't let us control the keyboard modifier keys, and the modifiers required to trigger the shortcuts are browser dependent. That enormously complicates documentation and tooltips. Meanwhile, those problems are solved within the keyboard handling infrastructure that we've already got.

I've experimented with adding global keyboard shortcuts via the page template but the problem is that the focus won't necessarily lie within the keyboard widget; on startup many browsers give the focus to the body or html elements.

So I think the solution is a new global keyboard shortcut mechanism that would listen for key events on the document. Then we can use the existing keyboard shortcut editor etc to set those shortcuts.

Anyhow, for the moment, let's just merge the accesskey support.

pmario:
- What happens, if 2 buttons have the same key?
 - What happens if a 1 button has a key and 1 "hidden" button has the same key
   - what if the hidden button gets visible ??

I think this can get very confusing very fast? ....  I didn't do any tests atm.

BurningTreeC:
@Jermolene - we could just not define keys by default and mention the config tiddlers in the "hidden settings" section - then we should discuss about a list of buttons which should get

`accessKey={{$:/config/accessKey/buttonName}}`

... I've read about the different modifiers for this and tested on chrome and firefox. both work, even if the newest FF needs a bit of intervention here on linux

I think accesskey support doesn't hurt if we don't define defaults. Works pretty well for me..

<hr>

@pmario 

I've tried 2 buttons with the same key. Only one of both gets triggered and it seems that it's the alphabetically "lower" (?) key

Hidden buttons can be triggered if they're hidden like the sidebar with "retain" (don't know what's the mechanism). If the sidebar is closed, the top pagecontrol buttons can get triggered. Buttons in the "Tools" tab only if that tab is open. And if for example the newtiddler button and the newjournal button both have "n" as the accesskey but the journal button is only shown in the "More" tab then:

- if the more tab is open a new journal gets created
- if the more tab is closed a new tiddler gets created

that's indeed confusing but why should I want to have the same key for two different buttons. I understand that it can happen though

BurningTreeC:
my assumption that it sorts alphabetically may be wrong, I've tried "x" as accesskey for the close-tiddler button and it closes the last tiddler in the story river. So maybe the last one in the dom or something like that

pmario:
>my assumption that it sorts alphabetically may be wrong, I've tried "x" as accesskey for the close-tiddler button and it closes the last tiddler in the story river. So maybe the last one in the dom or something like that

Yea. ... The signal may be bubbling up to the parent and then handled by one of the siblings...

pmario:
>that's indeed confusing but why should I want to have the same key for two different buttons. I understand that it can happen though

As you found out. Close Tiddler and transcluded template buttons are my concerns.

pmario:
>I think this can get very confusing very fast?

But depending on the user and configuraiton, it may be very powerful.

pmario:
We would have to make it clear in the docs, that it's for advanced users only. ... and they are responsible to make it work in a sensible way.

BurningTreeC:
> We would have to make it clear in the docs, that it's for advanced users only. ... and they are responsible to make it work in a sensible way.

except they're used for unique buttons only

BurningTreeC:
```
accessKey={{{` [list[$:/StoryList]nth[1]regexp<currentTiddler>addprefix<key>removesuffix<currentTiddler>] }}}}
```

BurningTreeC:
... thought that would work but doesn't :-1:

BurningTreeC:
... no this isn't good. if the storylist changes the accesskey remains "attached" to the tiddler that was previously on position 1 (even if my filter above is wrong)

BurningTreeC:
... this should either be used for unique buttons only, or we create that global mechanism aka keyboard widget for document

BurningTreeC:
@Jermolene 

If we would implement a global keyboard event listener that uses the mechanisms the keyboard widget uses but listens on the document - should it be a startup module or a widget? A core feature or a core plugin?

Jermolene:
Hi @BurningTreeC — I'd propose it be a core startup module.

BurningTreeC:
Hi @Jermolene @pmario @allothers

I've put a small startup module together that seems to work, but it's not in final shape:

```js
/*\
title: $:/core/modules/startup/keyboard.js
type: application/javascript
module-type: startup

Keyboard shortcut handling

\*/
(function(){

/*global $tw: false */
"use strict";

// Export name and synchronous status
exports.name = "global-keyboard";
exports.platforms = ["browser"];
exports.after = ["startup"];
exports.synchronous = true;

	if($tw.browser) {
		$tw.utils.each($tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[keyboardShortcut]!has[draft.of]]"),function(title) {
			var self = this;
			var key = $tw.wiki.getTiddlerText(title);
			var keyInfoArray = $tw.keyboardManager.parseKeyDescriptors(key);
			document.addEventListener("keydown",function(event) {
				if($tw.keyboardManager.checkKeyDescriptors(event,keyInfoArray)) {
						$tw.rootWidget.invokeActionString($tw.wiki.getTiddlerText(title + '/action'));
						event.preventDefault();
						event.stopPropagation();
						return true;
				}
				return false;
			},false);
		});
	}
};

})();

```
It looks for tiddlers tagged "keyboardShortcut" and reads its key descriptor from the text field
Then it looks for the tiddler with the same title but with suffix "/action" and gets the action string from its text field. I'm just lacking a better idea. But if you want to try how it works, that's it

Jermolene:
Hi @BurningTreeC looks good. A few thoughts on your code:

* The tag should be `$:/tags/KeyboardShortcut`
* Rather than using two separate tiddlers, I think it would be better to use a field "key" for the keyboard shortcut(s), and the text field for the actions (that's similar to the way that startup actions work)

However, there's more: we want to reuse the existing shortcut editor in control panel. Can you can figure out a bit more about how it works by doing an advanced filter search for `[all[shadows+tiddlers]prefix[$:/config/shortcuts]]`, and inspecting the resulting tiddlers alongside the source for the editor: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/ControlPanel/KeyboardShortcuts.tid

So, to make a shortcut appear in the editor for, say, "new tiddler" we'd need to make a tiddler called `$:/config/shortcuts/new-tiddler` and put the keycode (eg `alt-N`) in the text field. Then, the startup module needs to keep track of these mappings, which it can do by listening to changes in the wiki and filtering by the prefix `$:/config/shortcuts`.

BurningTreeC:
Thanks @Jermolene 

Yes, I'll take a look at the shortcut config of the keyboard widget and try to implement it that way
I've seen that "changes" listener before somewhere, shouldn't be that hard to copy.

I'll make a separate PR which will need a bit of real testing

BurningTreeC:
Now I have the change listener working for both changing the key and the action string

[See commit](https://github.com/BurningTreeC/TiddlyWiki5/commit/6b15d6a4e283a00314c512dc9e268a3c02379d09)

Is there a possible fix for the action string not triggering some action widgets?

BurningTreeC:
I have to correct myself: tm-full-screen does work, tm-new-tiddler does not

BurningTreeC:
should I close this in favor of the startup module or are you interested in both?

Jermolene:
Hi @BurningTreeC I was thinking that we should merge the support for accessKey attributes, but not bother to use them in the core. Now I'm thinking that perhaps we shouldn't bother; if we add it we'll have to provide support to users who ask about it, and it's brittle and complicated to use properly. So, I'd propose we close this PR.

BurningTreeC:
:) thanks

Issue #3274, opened 2018-05-13
Tiddler type field malfunctions if setting "links to missing tids" is disabled


twMat:
As noted [here](https://groups.google.com/forum/#!topic/tiddlywiki/m8O6LlhsS6k).
To reproduce:

Ctrlpanel > Settings .... uncheck "Enable links to missing tiddlers"

Then open a tid in edit mode and click on the type dropdown. The list is not correct.

BurningTreeC:
Hi @twMat , I've proposed a possible fix here: #3275

twMat:
wow, that was fast! Thank you. I'll wait til your PR is accepted before closing this issue.

Jermolene:
Thanks @twMat @BurningTreeC I'm beginning to think that I should never have merged the "disable links to missing tiddlers" setting; it's poorly thought through, evidently insufficiently tested and disables critical wiki functionality.

One thing I don't quite understand is why people might actually use it; I'm wondering if it's perhaps being used to mask some other problem?

pmario:
>One thing I don't quite understand is why people might actually use it; I'm wondering if it's perhaps being used to mask some other problem?

I can understand the desire. eg: If you create some documents for end-users, that are "pure" documentation "consumers". Users may have no idea, that TW is actually used to create the docs.  It may be confusing  for them, if they click a link and an "empty" tiddler opens. 

For me this falls into the same topic as "hiding tiddler toolbar buttons" ... or "read only mode"

pmario:
I think the problem could be fixed with CSS only.

BurningTreeC:
Hi ... I can imagine that it's useful in some cases to have it

@pmario - I don't know how to fix it with CSS, because of the linkcatcher widget catching the link, which isn't functional in the current state with "disable links to missing tiddlers"

pmario:
You are right. ... I didn't think about the functionality, of the buttons ... Just the block formatting :) 
Function is completely removed, here. .... 

So we may need ot split the whole thing into 2 different functions, that can be used independently

BurningTreeC:
... for the new-field-name drop-down the problem is the same
I'm ok with other ideas

twMat:
AFAICT this is now solved.
Thanks everyone.
Closing.

PR #3275, opened 2018-05-13
[type dropdown] fix for #3274

BurningTreeC:
this is a possible fix for the type dropdown where links don't show when "show missing links" is disabled

### DIFF
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -39,7 +39,7 @@ LinkWidget.prototype.render = function(parent,nextSibling) {
 		useWikiLinks = wikiLinksMacro ? (wikiLinksMacro.trim() !== "no") : true,
 		missingLinksEnabled = !(this.hideMissingLinks && this.isMissing && !this.isShadow);
 	// Render the link if required
-	if(useWikiLinks && missingLinksEnabled) {
+	if(useWikiLinks && missingLinksEnabled || this.showLinks) {
 		this.renderLink(parent,nextSibling);
 	} else {
 		// Just insert the link text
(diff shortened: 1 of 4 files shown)


Jermolene:
Thanks for tackling this @BurningTreeC. I've left a comment over at https://github.com/Jermolene/TiddlyWiki5/issues/3274#issuecomment-388829333

core/modules/widgets/link.js
```diff
@@ -39,7 +39,7 @@ LinkWidget.prototype.render = function(parent,nextSibling) {
 		useWikiLinks = wikiLinksMacro ? (wikiLinksMacro.trim() !== "no") : true,
 		missingLinksEnabled = !(this.hideMissingLinks && this.isMissing && !this.isShadow);
 	// Render the link if required
-	if(useWikiLinks && missingLinksEnabled) {
+	if(useWikiLinks && missingLinksEnabled || this.showLinks) {
```
pmario: I'm not so happy with the name. ... For me it is more like (lacking a better name) "forceLinks" ...

pmario: or may be "enforceLinks"

BurningTreeC: no problem, I was just lacking a better one

enforceLinks sounds good

BurningTreeC: or retainLinks or keepLinks
someone other than me should decide that

BurningTreeC:
@pmario - what I think is, that users may want to use that "linkcatcher + link to missing-tiddler" thing sometimes if they build something. This would enable that

PR #3276, opened 2018-05-13
fixing edit-field state mirrored in each tiddler

BurningTreeC:
this fixes that editing a field shows the text for name and value in each tiddler that's opened for editing


BurningTreeC:
I'll have to put some more work into this for the save-tiddler button and ctrl-enter for saving

Issue #3039, opened 2017-12-04
Feature Request: Suggesting improvements for BibTex plugin


diego898:
I am very interested in using the bibtex plugin in my TW. I use Mendely, and it automatically creates a bib file of my entire library. I made the following small edits to: 

1. made title of resulting tiddler be the parsed title
2. added an alias field, for my own uses
3. added a bibtex-citationkey field to store citationKey.

resulting in the following code change:

```
        // Convert each entry
	$tw.utils.each(data,function(entry) {
		var fields = {
			title: entry.entryTags.title,
			"alias": entry.citationKey,
			"bibtex-citationKey": entry.citationKey,
			"bibtex-entry-type": entry.entryType
		};
```

I then repackaged my plugin and refreshed. I tried uploading my library, but all tiddlers had their **titles** surrounded by `{}`. The actual bibitex file generated by Mendely has `title={{my title}}` for all entries, resulting in the titles all being surrounded by one set of `{}`. 

Even in the test file provided with the default plugin at:

https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/bibtex/samples/first-jabref-bibtex-db.bib

there are 12 entries with the double braces around the title. This seems like a common enough thing for TiddlyWiki to try and deal with. 

I'm not sure that ORCID/bibtexParseJs should be the one to deal with these special characters, since its job is just to parse the file. TiddlyWiki is the one that doesn't want tiddler names with `| [ ] { }`. 

Suggested Improvements:

- Automatically handle special characters in titles 
- Include built-in support for Zotero and Mendely
  - Within these cases, add option to tag posts with author names
  - Specifically for Mendely, it exports a `mendeley-tags` field in the bibtex-entry. We should be able to automatically add these tags to posts as well.

diego898:
I am still learning about plugins and how to pass them parameters/settings, etc. But so far I've made the following changes:

* carefully removed all `{{` and `}}` from the `title` field of my `.bib` file.
  * not that much of a pain using regular expressions
* Made the following changes to `$:/plugins/tiddlywiki/excel-utils/deserializer.js`:

```
// Convert each entry
    $tw.utils.each(data,function(entry) {
        var fields = {
            title: entry.entryTags.title,
            "bibtex-citationKey": entry.citationKey,
            "bibtex-entry-type": entry.entryType
        };
        $tw.utils.each(entry.entryTags,function(value,name) {
            fields["bibtex-" + name] = value;
[...]

PR #3281, opened 2018-05-15
fix bibtex import problem with unix systems 

pmario:
fix a problem with ubuntu 18.04, which returns text/x-bibtex instead of application/x-bibtex. ...

reported at: https://github.com/Jermolene/TiddlyWiki5/issues/3039#issuecomment-388743869

@gse-cc-git  can you test the attached index.zip It contains empty.html. ... It does contain the fix, but no plugins !!!!

[index.zip](https://github.com/Jermolene/TiddlyWiki5/files/2006010/index.zip)

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1999,6 +1999,7 @@ $tw.boot.startup = function(options) {
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.wordprocessingml.document","base64",".docx");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet","base64",".xlsx");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.presentationml.presentation","base64",".pptx");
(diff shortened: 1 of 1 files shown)


pmario:
I did test it with win10 and ubuntu 18.04

pmario:
@Jermolene ... can you have a look, wich mime-type the mac file explorer returns?

BurningTreeC:
works like a charm on arch linux

pmario:
@BurningTreeC ... could you test empty 5.1.17 from tiddlywiki.com, if it has a problem on "arch"

BurningTreeC:
@pmario I've already tested a current prerelease without your fix, the problem is there

BurningTreeC:
... and after testing it - also the empty.html 5.1.17 has the problem

gse-cc-git:
@pmario, @BurningTreeC . Lately but I confirm it worked on my Manjaro Linux.

pmario:
@Jermolene ... It seems, it can be merged. ... I did test with win10 and ubunutu. BTC and gse.. did test with other unix version. ... So if you test it with Mac, we should be good to merge it :)

PR #3283, opened 2018-05-15
keyboard shortcuts startup module

BurningTreeC:


BurningTreeC:
@Jermolene 

The startup actions cannot create new tiddlers - `<$action-sendmessage $message="tm-new-tiddler"/>` doesn't work
I say that, because I'm using that same mechanism here and cannot use the `action-sendmessage` widget

BurningTreeC:
... it's the `<$action-sendmessage>` widget that doesn't work in both the startup actions and this keyboard module. So tm-full-screen and tm-new-tiddler don't work, where the latter is one of the most useful

BurningTreeC:
I have to correct myself: tm-full-screen does work, tm-new-tiddler does not

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,43 @@
...
+			var key = $tw.wiki.getTiddler(title).fields["key"];
+			$tw.wiki.addEventListener("change",function(changes) {
+				if($tw.utils.hop(changes,title)) {
+					key = $tw.wiki.getTiddler(title).fields["key"];
+				}
+			});
+
+			document.addEventListener("keydown",function(event) {
+				if(key !== undefined && $tw.keyboardManager.checkKeyDescriptors(event,$tw.keyboardManager.parseKeyDescriptors(key))) {
+						$tw.rootWidget.invokeActionString($tw.wiki.getTiddlerText(title,$tw.rootWidget));
```
pmario: indentation? 1 tab more then if should be enyough

pmario:
if that doesn't work, you may need both: .after rootwidget and .before story ... But I'd test .after first.

BurningTreeC:
... that's why I initially thought that if I make this a simple widget and put it in the pagetemplate below the navigator widget it should work without these problems here

BurningTreeC:
@pmario ... I tried exports.after=["rootwidget"] and exports.before=["story"] and other variants and combinations without luck

Jermolene:
Hi @BurningTreeC @pmario 

> my suspicion is that the messages that get handled by the navigator widget don't work here, because it's defined in the PageTemplate and is missing here

That's correct. The current workaround is to wrap the action widgets in an instance of the navigator widget, with the appropriate settings for the story and history list.

Things are like this because I was keen to avoid permeating the whole core with the assumption that there would be a single story river.

Perhaps we could handle the navigation messages in `$:/core/modules/startup/story.js` in addition to via the navigator widget (story.js already knows about the default story list and history list). But let's consider that as a separate issue.

BurningTreeC:
@Jermolene 

> Perhaps we could handle the navigation messages in $:/core/modules/startup/story.js in addition to via the navigator widget (story.js already knows about the default story list and history list). But let's consider that as a separate issue.

Ok

Now the config part missing and I have no clue how it works.

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,43 @@
...
+exports.platforms = ["browser"];
+exports.after = ["startup"];
+exports.synchronous = true;
+
+exports.startup = function() {
+	if($tw.browser) {
+		$tw.utils.each($tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]!has[draft.of]]"),function(title) {
+			var self = this;
+			var key = $tw.wiki.getTiddler(title).fields["key"];
+			$tw.wiki.addEventListener("change",function(changes) {
```
Jermolene: The reason I suggested using a change event listener is so that we can dynamically track the active keyboard shortcuts as tiddlers with the required tag come and go. If we only read the shortcuts at startup then the shortcut editor won't work.

So, you'd have a function that reads all the keyboard shortcut tiddlers similar to what you've got here, but you'd run it at startup, and every time you get a change event for one of the tiddlers you're already tracking, or a new tiddler with the required tag.

Jermolene:
> Now the config part missing and I have no clue how it works.

Did my comments [here](https://github.com/Jermolene/TiddlyWiki5/pull/3273#issuecomment-389190582) help?

I'd suggest exploring those tiddlers prefixed `$:/config/shortcuts` while using the shortcut editor.

core/modules/startup/keyboard.js
pmario: Does this additional setting have a performance impact?

BurningTreeC:
> Did my comments here help?

Yes they did but I was fighting with them

Now I understand ... the lookupName() function in keyboard.js handles those config tiddlers that are "connected" through their titles

I can just add `$:/config/shortcuts/new-tiddler` and `$:/config/ShortcutInfo/new-tiddler` and in the control panel I can edit the key for new-tiddler which then changes in `$:/config/shortcuts/new-tiddler`

I've seen that the EditorToolbar Buttons have a field `shortcut` with for example `((strikethrough))` and I guess I'll have to put `((new-tiddler))` somewhere and the action in that tiddlers text field


Now what I don't understand is how I get the corresponding actionstring into the startup module

core/modules/startup/keyboard.js
BurningTreeC: ok I'll wrap this whole construct in a change listener

BurningTreeC:
@Jermolene @pmario 

>That's correct. The current workaround is to wrap the action widgets in an instance of the navigator widget, with the appropriate settings for the story and history list.

I need an example, a how-to for that or I'm lost in space

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,43 @@
...
+				if($tw.utils.hop(changes,title)) {
+					key = $tw.wiki.getTiddler(title).fields["key"];
+				}
+			});
+
+			document.addEventListener("keydown",function(event) {
+				if(key !== undefined && $tw.keyboardManager.checkKeyDescriptors(event,$tw.keyboardManager.parseKeyDescriptors(key))) {
+					$tw.rootWidget.invokeActionString('<$navigator story="$:/StoryList" history="$:/HistoryList">' + $tw.wiki.getTiddlerText(title) + '</$navigator>',$tw.rootWidget);
```
Jermolene: I'm against embedding the navigator widget like this (I meant that the workaround **for users** was to use the navigator widget around any action widgets that need it). Otherwise we're further driving the specific singular default story river into the core.

BurningTreeC:
@Jermolene , I've changed it so that one can add and change shortcuts without having to reload

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,56 @@
...
+
+// Export name and synchronous status
+exports.name = "global-keyboard";
+exports.platforms = ["browser"];
+exports.after = ["startup"];
+exports.synchronous = true;
+
+exports.startup = function() {
+	if($tw.browser) {
+		var shortcutTiddlers = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]!has[draft.of]]");
```
pmario: imo if `shortcutTiddlers` has a falsy value, we can return, without doing enything else. Like: "return early"

BurningTreeC:
@Jermolene @pmario 

I think we don't even need a change listener, the latest code works for all mentioned cases if I'm not totally off

if done like this it checks the shortcut tiddlers on each keystroke but hey, it seems to be fast and it's simple

Jermolene:
Sorry @BurningTreeC I do think we need the change event listener and ancillary parts. We've got to consider performance on low end hardware, and this is something that affects the performance of every keystroke.

BurningTreeC:
Ok, I'll see what I can do

pmario:
You can use this code in the console. It creates 20000 tiddlers and then try again. 
make sure you have a **HelloThere** tiddler, because it will be used as a template. 

```
var ntext = "" +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " +"123456789 123456789 123456789 123456789 " +
"123456789 123456789 123456789 123456789 " + "123456789 123456789 123456789 123456789 ";

var tiddler = $tw.wiki.getTiddler("HelloThere");

for (var i = 0; i<200; i++) {
   tag = "t-"+i;
   for (var k = 0; k<100; k++) {
      $tw.wiki.addTiddler(new $tw.Tiddler(tiddler, {title: "a-"+i+"-"+k, text: ntext, tags: [tag]}));
   }
}
```

BurningTreeC:
@pmario - not gonna try, I got the point. But thanks for the beautiful example :)

BurningTreeC:
@Jermolene - sorry for my previous overkill with the listener checking everything on each keystroke :grin:
Now I think it's much better, even if it caches all shortcuts again, if one changes or gets added

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,77 @@
...
+exports.platforms = ["browser"];
+exports.after = ["startup"];
+exports.synchronous = true;
+
+exports.startup = function() {
+	if($tw.browser) {
+		var shortcutTiddlers = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]!has[draft.of]]"),
+			shortcutKeysList = [],
+			shortcutActionList = [],
+			shortcutParsedList = [];
```
pmario: As I wrote:

```
if (!shortcutTiddlers) return;
```

IMO If no shortcuts are defined, there is no need to activate any listeners, so the function "exits early"

pmario: uups. .. The change has to be registered. ... It's a startup module. ... my bad!

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,77 @@
...
+				if($tw.utils.hop(changes,title) || shortcutTiddlers.indexOf(title) === -1) {
+					hasChanged = true;
+				}
+			});
+
+			if(hasChanged || newList !== shortcutTiddlers) {
+				shortcutTiddlers = newList;
+				shortcutKeysList = [];
+				shortcutActionList = [];
+				shortcutParsedList = [];
```
pmario: hmmm, Are those variables defined?

BurningTreeC: Seems so

BurningTreeC: A little help with the change listener is much appreciated

BurningTreeC: The current is wrong I see that

core/modules/startup/keyboard.js
BurningTreeC: yes, that's the point of this ... trying to cache those tiddlers, keys and actions and update them through the change listener

BurningTreeC:
I finally understand now how ((new-tiddler)) is used in combination with parseKeyDescriptors() and the $:/ControlPanel KeyboardShortcuts Tab :1st_place_medal: 

Are there some keys/actions where we'd like to add the config tiddlers so that they show up in the control-panel tab with or without defining the keys for them?

core/modules/startup/keyboard.js
```diff
@@ -65,7 +65,7 @@ exports.startup = function() {
 				}
 			}
 			if(!hasChanged) {
-				var shortcutConfigTiddlers = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]prefix[$:/config/shortcuts/]]");
+				var shortcutConfigTiddlers = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]prefix[$:/config/shortcuts]]");
 				for(i = 0; i < shortcutConfigTiddlers.length; i++) {
```
pmario: IMO trailing slash is needed! otherwise prefix handling will fail for `$:/config/shortcutsXXX/` which would be a "false positive"

pmario: Just found: `$:/config/shortcuts-not-mac/` .. how does the existing mechanism handle it?

BurningTreeC: in the control panel you can set shortcuts for linux only. that tiddler will be named $:/config/shortcuts-linux/something

pmario: yea saw it. .. replied to early :)

BurningTreeC: but you're right speaking about false-positives because this will update also if only a editor-shortcut changes

BurningTreeC: I'm not handling the different mac, linux, not-mac, not-linux ... config tiddlers, just all with prefix `$:/config/shortcuts` - I was expecting that you'll want that to be changed. The `$:/core/modules/keyboard.js` has that handling, I can use parts of it I guess

For distinguishing between editor and global-keyboard shortcuts I'm lacking ideas

BurningTreeC:
@Jermolene @pmario 

What do you think, is it the right approach now?

I don't know if it's worth distinguishing between changes of editor-shortcut tiddlers and global-shortcut tiddlers. The check if the key is defined in a tiddler tagged $:/tags/KeyboardShortcut and if that tiddler is one of the tiddlers that has changed will also take time, as it does simply re-caching all shortcut tiddlers at once if one of them has changed. I can't estimate the difference of performance-impact - what's your opinion?

pmario:
We'll need to test it. Are you satisfied with the functionality, atm?

I'll test it soon.

BurningTreeC:
@pmario yes functionality is what I expect, also the keyboard shortcuts controlpanel-tab works for the global shortcuts like for the editor shortcuts

to test:

create tiddler `$:/config/ShortcutInfo/new-tiddler` with a descriptive text in its text field
create a tiddler tagged `$:/tags/KeyboardShortcut` with the field `key` and its value `((new-tiddler))`
in the text field of the tagged tiddler put the action(s) for new-tiddler
in the controlpanel-tab set the shortcut
then you can test:
- does the combo work
- does adding/removing/changing work without reloading (either through the controlpanel-tab and through adding only the tagged tiddler with the key directly in its `key` field. both methods work)

Jermolene:
Great work @BurningTreeC.

As part of this PR we should add some documentation, and perhaps some core shortcuts for new tiddler etc.

The code suffers from some duplication. Perhaps you can refactor the code that reads the keyboard shortcut tiddlers into a helper function, and call it once at startup, and again when a relevant change event happens. The easiest way to do it is to keep track of the tiddler titles you got last time, and look for any changes.

BurningTreeC:
@Jermolene - I hope I've addressed the code duplication you mentioned

1) what shortcuts do you want in?
2) do you want the keys to be already defined?

a useful list of shortcuts:

- sidebar toggle
- new tiddler
- new journal
- fullscreen toggle ?
- import
- save-wiki ?

BurningTreeC:
now we can even navigate up and down the story list

pmario:
Is it possible to have "navigate next / previous"?

BurningTreeC:
> IMO F11 is already full-screen toggle?

yes, the browser fullscreen which isn't the same as the tw-fullscreen but I get it

> Is it possible to have "navigate next / previous"?

I don't think so - we can't access the current tiddler from outside the story river without ugly hacks, right?

BurningTreeC:
... maybe the dynaview plugin could keep track about which tiddler is probably the one that's viewed and add shortcuts for next/previous as a bonus

BurningTreeC:
@pmario 

>Is it possible to have "navigate next / previous"?

with a workaround: a keyboard shortcut that navigates to a tiddler and sets a state tiddler to its number in the story list
then we can navigate up/down using that number

Jermolene:
@BurningTreeC thank you again for your sterling efforts on this. I'll look properly when I'm on my laptop, but a quick thought is to limit ourselves to new journal and new tiddler at the moment; the story river navigation stuff is experimental (and I think there are other ways that we might approach it).

One thing that seems clumsy is that we're having to duplicate the logic of those new tiddler and new journal buttons. I am wondering if we can refactor things so that there's just one instance of shared logic; the easiest might be to refactor the buttons to use the actions from the keyboard shortcut tiddlers text fields?

BurningTreeC:
@Jermolene sure we can

I'll remove the navigation and make the buttons use the shortcut-actions but I'm leaving now and will continue tomorrow.
The fix for the action-navigate widget is still needed

For the docs, do you have a preference where to link them/put them?

BurningTreeC:
>> refactor things so that there's just one instance of shared logic

> sure we can

not so sure anymore @Jermolene - would you propose a way how to do it?

BurningTreeC:
I'm done here, let me know if something needs to be changed

BurningTreeC:
I found a problem with the CodeMirror editor:

Launching a global shortcut from within it triggers it twice - this doesn't happen in the title, tags, type or fields inputs or the standard editor. Only from within the CodeMirror textarea.

Does anybody have an idea where to look for a solution?

BurningTreeC:
@Jermolene @pmario - I found a fix for the case that the texteditor isn't framed. I'm testing within factory.js if domNodes[2] is an iframe - only then a new Event gets propogated to the container (that causes global shortcuts launched from within CM to trigger twice)

My test works but I guess there are better ones ..?
apart from this, I've been testing and I'd say it works

BurningTreeC:
* should I fix prop **o** gate or do you write it like that in UK?

Jermolene:
Hi @BurningTreeC it's "propagate" in the UK, too. It's too late to change the method names but we could certainly change the documentation references :)

BurningTreeC:
no, nevermind ;) ... it's fine, it has character

BurningTreeC:
@Jermolene - propogateKeydownEvent isn't referenced anywhere, neither in the code, nor on tiddlywiki.com/dev ... but on tiddlywiki.com there's the same typo - I make a PR

kookma:
@Jermolene 
Is there any change to merge this in new release (5.1.18). Shortcutkey module is a demanding one and can facilitate working with TW when you are with keyboard.

core/modules/editor/factory.js
```diff
@@ -285,17 +285,19 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	Propogate keydown events to our container for the keyboard widgets benefit
 	*/
 	EditTextWidget.prototype.propogateKeydownEvent = function(event) {
-		var newEvent = this.document.createEventObject ? this.document.createEventObject() : this.document.createEvent("Events");
-		if(newEvent.initEvent) {
-			newEvent.initEvent("keydown", true, true);
+		if(this.engine.constructor.name === "FramedEngine") {
```
Jermolene: Apologies if I've missed something in the discussion, but why do we need to check explicitly for the framed engine?

Jermolene:
Hi @BurningTreeC my apologies for leaving you in the lurch. I'm keen to merge this, but it still needs some work -- sorry it's been such a long slog; this is an ambitious change to the core, and we need to get it right.

I'll leave some review points shortly.

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,97 @@
...
+// Export name and synchronous status
+exports.name = "global-keyboard";
+exports.platforms = ["browser"];
+exports.after = ["startup"];
+exports.synchronous = true;
+
+exports.startup = function() {
+	if($tw.browser) {
+		// Get existing shortcut-tiddlers
+		var shortcutTiddlers = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]!has[draft.of]]"),
```
Jermolene: First, I think we could get away with calling `$tw.wiki.getTiddlersWithTag()`, which is much faster than compiling and executing a filter.

But more importantly, I think we can improve things here. There are three places where we're duplicating the same logic.

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,97 @@
...
+			shortcutKeysList = [],
+			shortcutActionList = [],
+			shortcutParsedList = [];
```
Jermolene: It would be worth a brief comment explaining what these lists are for, and what goes in them (something like [this](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/syncer.js#L55-L59))

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,97 @@
...
+
+		function updateShortcutLists(tiddlerList) {
+			shortcutTiddlers = tiddlerList;
+			for(var i=0; i < tiddlerList.length; i++) {
+				var title = tiddlerList[i],
+					tiddlerFields = $tw.wiki.getTiddler(title).fields;
+				shortcutKeysList[i] = tiddlerFields["key"] !== undefined ? tiddlerFields["key"] : undefined;
```
Jermolene: We would usually only use the `["key"]` notation when the string contains characters that aren't legal in JS names; here you can say `tiddlerFields.key` etc.

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,97 @@
...
+			// First inspect the existing shortcut-tiddlers
+			hasChanged = changedTiddlers(shortcutTiddlers);
+			
+			// Check if there are new shortcut-actions
+			if(!hasChanged) {
+				hasChanged = changedTiddlers(newList);
+			}
+			
+			// Check if key-combinations have changed
+			if(!hasChanged) {
```
Jermolene: I think that this check for changes to tiddlers with the config prefix would be more efficient as a loop through the `Object.keys(changes)` looking for the prefix.

core/modules/startup/keyboard.js
```diff
@@ -0,0 +1,97 @@
...
+			return false;
+		},false);
+
+		// Listener for changing shortcuts "on-the-fly"
+		$tw.wiki.addEventListener("change",function(changes) {
+			var newList = $tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]!has[draft.of]]");
+			var hasChanged = false;
+			
+			// Returns true if a passed array contains a Tiddler that has changed
+			function changedTiddlers(tiddlerList) {
```
Jermolene: This function might be better named as `hasAnyTiddlerChanged`?

core/modules/widgets/action-navigate.js
```diff
@@ -56,7 +56,7 @@ Invoke the action associated with this widget
 */
 NavigateWidget.prototype.invokeAction = function(triggeringWidget,event) {
 	var bounds = triggeringWidget && triggeringWidget.getBoundingClientRect && triggeringWidget.getBoundingClientRect(),
-		suppressNavigation = event.metaKey || event.ctrlKey || (event.button === 1);
+		suppressNavigation = event !== undefined ? event.metaKey || event.ctrlKey || (event.button === 1) : false;
```
Jermolene: Here we're really protecting against a missing parameter value, and so we can use a construction that we use elsewhere:

```
NavigateWidget.prototype.invokeAction = function(triggeringWidget,event) {
event = event || {};
```

core/ui/Actions/new-journal.tid
```diff
@@ -0,0 +1,14 @@
+title: $:/core/ui/Actions/new-journal
```
Jermolene: This works well, is there a reason we don't do it for the new tiddler action?

editions/tw5.com/tiddlers/howtos/KeyboardShortcuts.tid
```diff
@@ -1,18 +1,62 @@
 created: 20140526204527418
 modified: 20160610082117953
-tags: Features [[Working with TiddlyWiki]]
+tags: Features [[Working with TiddlyWiki]] Keyboard
```
Jermolene: Is there a reason for the "Keyboard" tag?

core/modules/editor/factory.js
BurningTreeC: Hi @Jermolene - the propogateKeydownEvent method creates a new Event also in a CodeMirror instance and that new Event is because of the iframe of the framed engine not propagating the event up to the container by default. In the CodeMirror text editor a global keyboard shortcut would get dispatched twice, that's why I'm testing if we're within a FramedEngine here

editions/tw5.com/tiddlers/howtos/KeyboardShortcuts.tid
BurningTreeC: not really, no

core/ui/Actions/new-journal.tid
BurningTreeC: it's just one line there that would get replaced by another line and I didn't see the benefit of it

core/modules/widgets/action-navigate.js
BurningTreeC: ok :+1:

core/modules/startup/keyboard.js
BurningTreeC: I'll change it

core/modules/startup/keyboard.js
BurningTreeC: I'll look into it

core/modules/startup/keyboard.js
BurningTreeC: ah yes, I've seen that - I'll change this

core/modules/startup/keyboard.js
BurningTreeC: I'll add comments here

core/modules/editor/factory.js
BurningTreeC: There are probably better fixes but this is the best I've come up with

core/ui/Actions/new-journal.tid
BurningTreeC: one line (two counting both the button and the shortcut) would get replaced by other two (or more) lines + the new action tiddler in the new-tiddler case.

I can see the benefit for users simply transcluding action tiddlers - if that's the way you want to go I'm in :sunglasses: ... then we could probably create more action-tiddlers and refactor buttons

BurningTreeC: @Jermolene now we do it for new-tiddler, too

core/modules/editor/factory.js
BurningTreeC: @Jermolene do you know a better fix?

BurningTreeC:
@Jermolene - I've removed the check for "FramedEngine" in factory.js - event.preventDefault() and event.stopPropagation() in the keyboard startup module now do the trick

I don't know what else to change, it appears to be very reliable now

pmario:
@BurningTreeC ... Why did you close your PR's? ... Is there a problem?

PR #3285, opened 2018-05-15
Contribute webmodule code from TiddlyServer

Arlen22:
This is the code that I use to load data folders. I wrote it myself, and have now modified it slightly for using TiddlyWiki directly. It is useful for loading a data folder as a route in express or other route-based servers but must have access to the original request and response object. 

Requests are handled by calling `server.requestHandler(req, res)`.

### DIFF
--- a/webmodule.js
+++ b/webmodule.js
@@ -68,6 +68,6 @@ exports.loadTiddlyWiki = function loadTiddlyWiki(mount, folder, callback) {
 		$tw.hooks.invokeHook('th-server-command-post-start', server, events, "webmodule");
 
 		//return the server and event emitter
-		callback(null, server, events);
+		callback(null, server, events, $tw);
 	}
 };
--- a/webmodule.js
+++ b/webmodule.js
@@ -1,0 +1,83 @@
+// Written by @Arlen22 for TiddlyServer, modified for TiddlyWiki
+
+/*
+Call loadTiddlyWiki to load the datafolder specified at the mount path specified and 
+return the SimpleServer instance to be used for handling requests.
+
+Requests may be handled by calling server.requestHandler(req, res)
+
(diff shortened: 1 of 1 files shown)


pmario:
What are "data folders"

webmodule.js
```diff
@@ -0,0 +1,73 @@
+// Written by @Arlen22 for TiddlyServer, modified for TiddlyWiki
```
pmario: There is no TW module header?
What should this module do, and how do I use it?

pmario: ok

Arlen22:
Well, it's more like it allows you to load a data folder easily into your Express app on a specific mount path. So it just attaches the data folder and handles requests. If that's what you mean, then yes. :)

Arlen22:
That being said, I don't mind turning this into a gist or just posting it on TiddlyWiki.com. Either one would work fine, and it would cut down on clutter.

pmario:
@Arlen22 ... 5.1.18 will have big changes for the server command. Pls check if this PR is still valid.

Jermolene:
I think @Arlen22's approach, once updated, is still useful, but as mentioned above I think it makes more sense as a code sample in the /dev wiki because it's not useful for general TW5 users on its own.

Arlen22:
> @Arlen22 ... 5.1.18 will have big changes for the server command. Pls check if this PR is still valid.

TiddlyServer will contain the new implementation once I update it. I think the code is mostly self-contained and fairly self-explanatory. Unless someone really wants to go to that much trouble, a note in the dev wiki pointing out that it can be done should be enough, and maybe linking to my implementation in TiddlyServer. I wrote the first code for Express, then used that idea to make TiddlyServer, so this has wider uses, including external authentication and [now] authorization.

Issue #3268, opened 2018-05-10
[Print Tiddler Button] print dialogue fires too early


BurningTreeC:
To replicate: print a tiddler with some heavier content that doesn't render instantly  (more images, something that takes a bit to load) when the new window opens. The print dialogue doesn't show the content then. Closing the dialogue causes the window to finish rendering and afterwards one can print the content

sukima:
Not sure how to fix this but to reproduce this in development you can use this PHP script:

```php
<?php
sleep(5);
header("Cache-Control: no-store, no-cache, must-revalidate, max-age=0");
header("Cache-Control: post-check=0, pre-check=0", false);
header("Pragma: no-cache");
header("Content-Type: image/jpg");
echo readfile("editions/tw5.com/tiddlers/images/Blurry Lawn.jpg");
?>
```

1.  Place the above PHP script in the root TiddlyWiki5 folder. (`test.php`)
2.  Run the PHP dev server: `php -Slocalhost:8888`
3.  Run the tiddlywiki server in a seperate terminal: `./tiddlywiki.js editions/tw5.com --server`
4.  Open the browser to `http://localhost:8080`
5.  Create a new tiddler with the following in it: `<img src="http://localhost:8888/test.php">`
6.  Attempt to **print in new window**

BurningTreeC:
Hm. How does this work in your test?

line 65 in windows.js

```js
                widgetNode.render(srcDocument.body,srcDocument.body.firstChild,function(){
                // Print the window if required
                if(print) {
                                srcWindow.print();
                }
                });
```

BurningTreeC:
-.- I thought that worked but testing it now after restarting node I've seen that I was still using the setTimeout method :+1:

Jermolene:
Thanks @BurningTreeC @sukima. I wondered if we could find a good example tiddler for testing? My current thinking is that we should push the feature back to v5.1.18 if we can't fix this overnight.

sukima:
@Jermolene We can't really have a test tiddler because the bug only presents itself if the content of such tiddler is remote *and* the remote takes forever to resolve. Since most all of the tiddlers in the current repository are built-in we would not see this happen.

The reason I posted the test Node server above was to artificially introduce a large delay to the act of serving a remote image. It used an image in the repo but has nothing to do with the tiddlywiki server. Once a tiddler that references that slow image (by means of a `<img src="http://localhost:8888/">` in the tiddler body) is created the bug will present itself.

[...]

PR #3287, opened 2018-05-16
Add print tiddler toolbar button

Jermolene:
We pulled this from 5.1.16 because of timing issues -- see #3268.

Perhaps we should adopt a long timeout (say 3 seconds) but include an on-screen timer, and provide a hidden setting for altering the timeout.


BurningTreeC:
I've built something here: http://print-tiddler.tiddlyspot.com

It allows to open a tiddler in a new window with the pagecontainer, story-river and tiddler-classes defined so that styles will be kept. Further one can add more tiddlers for a multi-tiddler document and have titles, subtitles and tags shown or hidden

Jermolene:
Hi @BurningTreeC that's great, works really well. Maybe two minor issues:

* The three preview buttons to hide the tiddler body elements are genius, but maybe a little inscrutable
* The button overlay can be hard to read when overlaid on tiddler content

Perhaps we could make the overlay panel have an opaque background, reusing the styling of `.tc-plugin-reload-warning` (although we should perhaps add `.tc-print-tiddler-window-overlay` as an alias).

Are you able to push more commits to this PR? I'm not sure I understand GitHub's permissions model well enough to predict it.

BurningTreeC:
> Are you able to push more commits to this PR? I'm not sure I understand GitHub's permissions model well enough to predict it.

I'll try - I'm going to make your proposed changes at the tiddlyspot page, then I try PR-ing

BurningTreeC:
The preview buttons are inscrutable, I wanted to use icons but there are no suitable icons for title and subtitle ... it would'nt be easy to find/make some... So I'll go for text buttons I think

Jermolene:
Great, thanks @BurningTreeC

BurningTreeC:
@Jermolene I can only clone the repo and add to my clone / change there, no push access.
I've addressed your ideas on the [demo page](http://print-tiddler.tiddlyspot.com) if you want to take a look

BurningTreeC:
@Jermolene I've made another PR #3331

BurningTreeC:
I thought it might be good to be able to pass an attribute to `tm-open-window` to set a window title that will be used for the document's name when it gets printed. Possibly the `SiteTitle` + a timestamp

BurningTreeC:
ping for 5.1.20

BurningTreeC:
ping

PR #3289, opened 2018-05-16
Add WebSocket Server to TiddlyWiki core

Arlen22:
This adds a websocket server to the server.js and changes the startup hook to an event emitter since the primary reason for sending the Node server in the hook was to attach a websocket handler. All websocket connections are forwarded directly to the event emitter.

### DIFF
--- a/core/modules/commands/server.js
+++ b/core/modules/commands/server.js
@@ -314,15 +314,16 @@ Command.prototype.execute = function() {
 		$tw.utils.warning("Warning: Plugins required for client-server operation (\"tiddlywiki/filesystem\" and \"tiddlywiki/tiddlyweb\") are missing from tiddlywiki.info file");
 	}
 	//call the server post-start hook with the SimpleServer and EventEmitter
-	$tw.hooks.invokeHook('th-server-command-post-start', this.server, eventer, "tiddlywiki");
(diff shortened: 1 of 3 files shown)


Arlen22:
The primary case for this is to add basic WebSocket support to the server command so that plugins can begin to implement WebSocket support without creating a whole bunch of incompatible implementations. The way I have implemented it here allows TiddlyServer to support WebSockets based plugins originally written for the NodeJS server without modifying them in any way.

Arlen22:
This doesn't just benefit TiddlyServer however. I know that this is a massive change, but this allows anyone to load a datafolder as part of a NodeJS server and be able to use plugins that require websockets without any modification.

diego898:
How does this play in with the changes Jeremy proposed recently in the dev group?

Jermolene:
Apologies @Arlen22 I see I hadn't responded to this. It's early days for websocket support in the core, and so I think we'd be best to explore the space via experimental plugins, and consider migrating the functionality to the core once things are more mature. (Personally, I'm more interested in [server sent events](https://www.html5rocks.com/en/tutorials/eventsource/basics/) as a better match for the basic update notification stuff, primarily because they behave better with proxy servers and other bits of old school Internet infrastructure — but I'd like to explore both).

Arlen22:
Closing in favor of https://github.com/Jermolene/TiddlyWiki5/pull/4609

Issue #738, opened 2014-08-17
Add a WebDAV Saver


Jermolene:
As discussed in this thread, it turns out to be pretty simple to get WebDAV up and running:

https://groups.google.com/d/topic/tiddlywikidev/eb5JHNm_eak/discussion

@FND has hacked together a working implementation:

https://gist.github.com/FND/b3e969f5ed471b55c831

edrex:
@FND, does your store load and save individual .tids in the same format as used by the NodeJS server? I'm very much interested in continuing this work.

Here's a sketch of what I'm planning:

Phase 1:

- Minimal HTML TiddlyWiki loads first, then loads all tiddlers in store by walking directory indexes. This will be slow with HTTP/1.1, but wicked fast with HTTP/2 pipelining.
- Reference WebDAV server with HTTP/2 support, with a similar interface to `tiddlywiki server`. Either in Golang or Node.js. I'm already pretty familiar with the http2 implementation in Go. This is just for convenience - the goal is that any WebDAV server will work.

Phase 2:

Introduce a [Service Worker](https://github.com/slightlyoff/ServiceWorker/blob/master/explainer.md) for cache management and offline data access, which will provide a reliable offline experience on [supporting browsers](http://caniuse.com/#feat=serviceworkers). I think this will be independent of WebDAV, still need to work out the details.

FND:
My implementation there uses TiddlyWeb's tiddler RFC822-style serialization (i.e. `key: value` pairs, blank line, body) - which, as I recall, is also what TW5's Node implementation was based on. So I imagine it should be fully compatible.

It would be very exciting to see this actually working out, so you're more than welcome to pick up the code and use it however you see fit. Unfortunately, since I stalled out on this effort quite a while ago, I'm not sure that I'll be much help at this point - though I can try, so let me know if anything's unclear.

By the way, that Taskard repo has moved to https://github.com/FND/taskard.original - so I'm afraid the URLs above aren't quite right anymore (it's all still there though, just add the ".original" suffix).

edrex:
## Timeline

### Developer preview

* [ ] Test suite for client
* [ ] If on HTTP, check server headers for DAV
* [ ] Queries root for list of tiddlers with `[PROPFIND](http://www.webdav.org/specs/rfc4918.html#METHOD_PROPFIND) Depth: infinity` on root resource
* [ ] Root defaults to the current URL with file extension and index.html stripped and '/' appended:
[...]

PR #2277, opened 2016-02-08
add a simple put saver, for saving to a webdav or REST server

edrex:
Fixes #738

PR #3292, opened 2018-05-18
Webdav adaptor

icarito:


icarito:
I am able to test this with wsgidav but I guess it should work with other WebDAV solutions (@pmario, @Arlen22 please try it!). Thanks in advance!

editions/tw5tiddlyweb/tiddlywiki.info
```diff
@@ -1,7 +1,8 @@
 {
```
pmario: The TiddlyWeb spec is much more then the basic tiddlyweb adapter implemented in TW5. ... So imo it should have a different name. 

I think "webdav-adaptor" would be OK for me. ... So the PUT_save we implemented could be: "webdav-put-etag" and a version I'm experimenting with could be "webdav-put-modified" ... Which uses "if-unmodified-since" header. IMO it's much more robust against different server settings as "compression" on/off and so on.

editions/tw5tiddlyweb/tiddlywiki.info
```diff
@@ -1,7 +1,8 @@
...
 	"description": "Base edition for TiddlySpace integration",
 	"plugins": [
-		"tiddlywiki/tiddlyweb"
+		"tiddlywiki/tiddlyweb",
+		"tiddlywiki/filesystem"
 	],
```
pmario: hmmm. do we need them here?

editions/webdav/tiddlers/TiddlerListTemplate.tid
```diff
@@ -0,0 +1,3 @@
+title: TiddlerListTemplate
```
pmario: I think, new templates should be part of the plugins, **if** the are essential for plugin function. If they are docs only, edition is OK.

The edition shouldn't be needed, if a user wants to use new functionality

editions/webdav/tiddlers/system/SiteTitle.tid
```diff
@@ -0,0 +1,3 @@
+title: $:/SiteTitle
+
+TiddlyWiki in the Sky (WebDAV)
```
pmario: :) nice choice!

editions/webdav/tiddlywiki.info
```diff
@@ -0,0 +1,14 @@
+{
```
pmario: why are there 2 different editions?

plugins/tiddlywiki/webdav/html-div-tiddler.tid
```diff
@@ -0,0 +1,9 @@
+title: $:/core/templates/html-div-tiddler
```
pmario: IMO this template is essential and only used to store tiddler content in the single-file version. ... It should _not_ be modified, or used in other ways. ... Saving tiddlers in .tid format looks different!

plugins/tiddlywiki/webdav/plugin.info
```diff
@@ -0,0 +1,7 @@
+{
+	"title": "$:/plugins/tiddlywiki/webdav",
+	"description": "WebDAV sync adaptor",
+	"author": "Sebastian Silva",
+	"core-version": ">=5.0.0",
```
pmario: There are no supported 5.0.0 versions ... TW [started with 5.1.0](https://tiddlywiki.com/#Releases) excluding alphas and betas.

Did you test the plugin with 5.1.0? ... It would be cool if it would work with older versions. BUT we have to be sure. Otherwise we will get a lot of support requests!

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
+/*\
+title: $:/plugins/tiddlywiki/webdav/webdavadaptor.js
+type: application/javascript
+module-type: syncadaptor
+
+A sync adaptor module for synchronising with WebDAV compatible servers
+Heavily based on WebDAV Adaptor
```
pmario: based on WebDAV adaptor or "TiddlyWeb" adaptor?

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+
+WebDAVAdaptor.prototype.name = "webdav";
+WebDAVAdaptor.prototype.knownDirs = [];
+
+WebDAVAdaptor.prototype.isReady = function() {
+	return this.hasStatus;
+};
+
+WebDAVAdaptor.prototype.getHost = function() {
+	var text = this.wiki.getTiddlerText(CONFIG_HOST_TIDDLER,DEFAULT_HOST_TIDDLER),
```
pmario: IMO this isn't a tiddler name, its a configuration value. right? I think DEFAULT_HOST should be ok.

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+	var self = this;
+	$tw.utils.httpRequest({
+		url: this.tiddlersBaseURL + "/" + title + ".tid",
+		headers: {accept: "charset=UTF-8"},
+		callback: function(err,data,request) {
+			if(err) {
+				return callback(err);
+			}
+			// Invoke the callback
+			var type = "application/x-tiddler";
```
pmario: tiddlers are type: `text/vnd.tiddlywiki`, `application/x-tiddler` is tiddlywiki-classic

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+			// Invoke the callback
+			callback(null);
+		}
+	});
+};
+
+if($tw.browser && document.location.protocol.substr(0,4) === "http" ) {
+	exports.adaptorClass = WebDAVAdaptor;
+}
+
```
pmario: I'm not 100% sure, but shouldn't there be a `.canSave()` function, which tells the saver module it can use the adaptor?

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+		var s = substitutions[t];
+		text = $tw.utils.replaceString(text,new RegExp("\\$" + s.name + "\\$","mg"),s.value);
+	}
+	return text;
+};
+
+/*
+Make a directory recursively
+*/
+WebDAVAdaptor.prototype.mkdir_p = function (path, callback) {
```
pmario: Is it reliable, if the `tiddlers` directory doesn't exist on the server?

editions/webdav/tiddlers/WebDAV sync adaptor plugin.tid
```diff
@@ -0,0 +1,15 @@
...
+! Features
+
+* Loads skinny tiddlers from entire shared directory at startup/login
+* Subsequently syncs changes back to the server
+* Polls for changes from the server
+
+! Issues
+
+* ''[[$:/DefaultTiddlers]]'' doesn't work because thanks to lazy loading it is only loaded after it is needed
+* Should probably disable PUT saver
```
pmario: I don't understand this line?

editions/webdav/tiddlers/WebDAV sync adaptor plugin.tid
```diff
@@ -0,0 +1,15 @@
...
+* Ignores ''if-match'' header, so doesn't detect clashes on save
```
pmario: ouch ... We shold use `if-unmodified-since` header instead. .. I did some experiments replacing the `if-match` stuff with the PUT save and it works great.

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+			}
+
+			if (doc!==undefined) {
+				const ns = "DAV:";
+				var tiddlers = [];
+				var items = doc.children[0].children;
+				// Start at 1, because the 0th is the same as self.
+				for(var i=1; i< items.length; i++) {
+					var response	 = items[i];
+					var href		 = response.getElementsByTagNameNS(ns, 'href')[0].firstChild.nodeValue;
```
pmario: coding style: not tabs in front of `=`

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+					href = href.replace(/\/$/, ''); // Strip trailing slash
+					href = href.replace(self.tiddlersBaseURL + "/", ""); // Strip path
+					href = decodeURIComponent(href);
+					var propstat	 = response.getElementsByTagNameNS(ns, 'propstat')[0];
+					var prop		 = propstat.getElementsByTagNameNS(ns, 'prop')[0];
+					var resourcetype = prop.getElementsByTagNameNS(ns, 'resourcetype')[0];
+					var collection   = resourcetype.getElementsByTagNameNS(ns, 'collection')[0];
+
+					if (prop.getElementsByTagNameNS(ns, 'getcontenttype').length > 0) {
```
pmario: Do we need a sanitation step here?

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+						// var type = prop.getElementsByTagNameNS(ns, 'getcontenttype')[0].innerHTML;
+						if (href.endsWith(".tid") && href!=="$__StoryList.tid") {
+							var type = "application/x-tiddler";
```
pmario: IMO should be: 

```
var type = TIDDLER_TYPE;
``` 

and

```
var TIDDLER_TYPE = "text/vnd.tiddlywiki"

File: plugins/tiddlywiki/webdav/html-div-tiddler.tid (hunk 6 above)
Comment (pmario) [reply plugins/tiddlywiki/webdav/html-div-tiddler.tid]: Why do you want to change it?

File: plugins/tiddlywiki/webdav/html-div-tiddler.tid [hunk 18]
```diff
@@ -0,0 +1,9 @@
...
+
+<!--
+
+This template is used for saving tiddlers as an HTML DIV tag with attributes representing the tiddler fields. This version includes the tiddler changecount as the field `revision`.
+
+-->`<div`<$fields exclude='text revision bag' template=' $name$="$encoded_value$"'></$fields>` revision="`<<changecount>>`" bag="default">
```
pmario: server specific fields need to be built dynamically at load time. .... They are _not_ saved in the .tid file!

pmario: server fields IMO should not saved in the "Download" single file version

pmario:
Very nice work! ... 

There are several things, which I think need to be improved. Comments incoming in the code :)

plugins/tiddlywiki/webdav/webdavadaptor.js
```diff
@@ -0,0 +1,248 @@
...
+			if(err) {
+				return callback(err);
+			}
+			// Invoke the callback
+			callback(null);
+		}
+	});
+};
+
+if($tw.browser && document.location.protocol.substr(0,4) === "http" ) {
```
pmario: hmmm, I personally consider unencrypted webdav write access, to be extreamly dangerous. ... 

Our users will use the stuff and expose it to the web. Many of them don't ask, if it save to do so, they expect it to be save. Which it isn't.

With IIS its simple to create a "self-signed" cert even for local testing. And I think, we should force it.

editions/webdav/tiddlywiki.info
icarito: This is the new one - the other was already there.

editions/webdav/tiddlers/WebDAV sync adaptor plugin.tid
icarito: +1    but it's additional to what tiddlyweb had. Currently I'm just trying to get it to work! :-)

plugins/tiddlywiki/webdav/webdavadaptor.js
icarito: Okay :-)

plugins/tiddlywiki/webdav/html-div-tiddler.tid
icarito: Sorry I'm still not failiar with what server fields are etc - I'm seeing some issues with the saver but I don't understand yet what's wrong.

plugins/tiddlywiki/webdav/webdavadaptor.js
icarito: I think it should be the user's responsibility! We could add a note.

icarito:
Please feel free to patch this or work on it - I won't have time till next week. Your review is sincerely appreciated!

editions/tw5tiddlyweb/tiddlywiki.info
pmario: Ah, yea. the original tiddlyweb plugin needs both. .. But I don't know, what this has to do with webdav. ... 

If it's just a general fix. please consider a second PR with just this fix.

editions/webdav/tiddlers/TiddlerListTemplate.tid
pmario: I see. We have to have a closer look here! ...

editions/webdav/tiddlywiki.info
pmario: ok. As I wrote. Please create a second PR for the other one.

editions/webdav/tiddlers/WebDAV sync adaptor plugin.tid
pmario: I see.

plugins/tiddlywiki/webdav/html-div-tiddler.tid
pmario: I think I know which problems you have. I think I had the same problems when I was messing around with the dat-adaptor, which wasn't released, because the dat-api needed was removed. ... 

I'll have a closer look, too. ... But this weekend will be limited time, because of "family business" :)

plugins/tiddlywiki/webdav/webdavadaptor.js
pmario: I tend to disagree a little bit here. ... We need a ---- big note ----

plugins/tiddlywiki/webdav/webdavadaptor.js
Arlen22: Um, I think that's the other way around. `application/x-tiddler` is TiddlyWiki 5 if I'm reading boot.js correctly (line 1300).

plugins/tiddlywiki/webdav/plugin.info
Arlen22: This is what almost all core plugins specify. But still a good thing to change :)

Arlen22:
This does not work with TiddlyServer because TiddlyServer does not support everything specified here. However, I think that is fine, because this is intended to work with data folders and TiddlyServer already has that taken care of with the TiddlyWeb adaptor and server command that come default with TiddlyWiki. This adaptor is more for wikis running on other kinds of servers. 

In other words, if someone is already running TiddlyServer, they will want to use TiddlyWeb anyway, since it is designed for TiddlyWiki, whereas this is more of an adaptor to make TiddlyWiki work in places where NodeJS is not available. 

That's my very long way of saying "No, TiddlyServer won't ever work with this plugin, but it doesn't need to. PLEASE DON'T LET THAT DISCOURAGE YOU!!! THIS IS A REALLY USEFUL TOOL!!! AND A **_USEFUL CONCEPT_** AS WELL!!! It's just that TiddlyServer already has this use case covered thanks to NodeJS Tiddlywiki and the server command.". I hope you get my point :) :+1: 

I want to use this exact concept to make a Dropbox sync adapter some day.

pmario:
@Arlen22 ... It's not intended to be active when a "tiddlyweb" nodejs server is used as the backend. 

With 5.1.17 we made a new WebDAV PUT saver, that can deal better with etags and different server backends. .. apache, nginx and microsoft IIS. ... The PUT saver is relatively simple since it only saves the "single-file" version. 

This adaptor uses the "single-tiddler" approach, similar to tiddlyweb, but with WebDAV. It's a very interesting concept and it should work very well with a local network, since we can use "full-tiddlers" instead of "skinny-tiddlers" ...

pmario:
Once we got this going, there should be only a very small step, for a "single-tiddler" multi-user dat-adaptor. ...

Jermolene:
Hi @icarito thanks for this, it's a great idea – there's a lot of Sharepoint users out there, and it would be pretty cool if this can be made to work on that. I'm grateful you're contributing it to the core; there's a higher bar before acceptance than for you to publish it yourself, but there's more scope for collaboration and in the end it's likely to be more useful this way.

I'll leave @pmario to discuss the code in detail, but wanted to quickly make the point that we might want to rename it `$:/plugins/tiddlywiki/webdavsync`, just to reduce confusion with the existing webdav saver.

Arlen22:
The one thing I wasn't sure about is how is this going to work with the standard skinny tiddler list? Doesn't the skinny tiddler list need to have all the fields except the text field? One thing you could do is load all tid files and all other meta files for the skinny tiddler list, I suppose. Maybe you already got this covered -- I didn't read the code through in detail.

Arlen22:
I'm also not sure that it works correctly to use the file title as part of the skinny tiddler list. Maybe it does, I haven't looked.

One thing you could do, though, is to load all the files (just the relevant json, tid, and meta files) on startup and cache the actual skinny tiddler array (all fields except the text field). Then every time getSkinnyTiddlers is called, you would check for changes to the file system (as you do presently) and if any files are added or changed, you would then read the skinny part of the file (the .tid or .meta) and get the updated skinny fields. You would also update the revision field, and then give the whole list to the callback. 

And just so you know, the revision field is entirely under the control of the sync adaptor. So you can store whatever revision marker you want in it -- it doesn't matter to the wiki. 

Anyway, that's just the idea from my head. It's up to you if you want to use it. 

Have fun :+1:

edrex:
@Jermolene the one I implemented in #2277 is named `put`, because that's all it does. Any webdav server at all concerned about security wouldn't serve a user-supplied file as html. I shouldn't have closed #738 with that, not sure what I was thinking.

I was looking at the state of webdav in TW5 to try to answer https://www.reddit.com/r/TiddlyWiki5/comments/ar0515/using_tw_with_nextcloud/ and it looks like this is the frontier :) @icarito are you still interested in pushing this forward? I'd be happy to review.

Jermolene:
Hi @edrex @icarito I'd still be very keen to see this completed. As I said above it doesn't have to be in the core but ultimately might be more valuable if it were.

pmario:
@edrex ... The last time I did have a closer look at NextCloud, it only lets you connect a file-explorer with webdav. It doesn't serve files as HTML. 

But there are several other open-source projects, that let you serve TW as HTML files from webDav.

pmario:
The second big problem with the existing webDav saver is the etag handling and server side compression. 
Every server seems to have a different opinion, how etags are handled, if compression is on.

I did test with: nginx, IIS, apache ... With all of them you need to switch compression OFF, which is a shame. 

I did create a proof of concept, where I used the `last-modified` header info, which seems to work well with server side compression.

Issue #3296, opened 2018-05-20
sortByList is incorrect when placement specifications are in inconvenient order


flibbles:
Simple example: three tiddlers, all tagged with `foo`

A {list-after: B}
B {list-after: C}
C

Results?

A, C, B

Tiddler A is not after B, because in sortByList, it was placed after B, then B was placed after C, which put it back after A. This problem can manifest in many little ways when using `list-after` and `list-before`, but this is the most rudimentary example.

I already have a fix, and I can make a PR at any time. I'm unfamiliar with the Tiddlywiki dev process. Am I allowed to go straight to making the PR? Or should I continue making an issue about it first?

Jermolene:
Thanks @flibbles it would be very much appreciated if you can produce a PR for this.

There is already a single test for tag ordering. As part of the PR it would be very useful to expand the tests:

https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-tags.js

flibbles:
Done! Let me know if there's any issues with the code. Formatting, style, tests...

pmario:
As I wrote in PR #3301 I don't think there is an easy fix. eg: A (list-after B) and B (list-after A)

flibbles:
As it turns out, it actually _is_ an easy fix. @Jermolene and @pmario, PR #3301 is ready to merge.

If not, please provide a reason why it shouldn't be. This issue and PR has been stagnating for nearly 6 months when it was ready to merge in the first week. It fixes a legitimate bug.

flibbles:
Thank you all for merging the fix for this issue. This proves that there is a brighter future ahead for our children and our children's children.

PR #3301, opened 2018-05-20
List order

flibbles:
Fix for (#3296), custom tag ordering could mess up when some instances of `list-after` and `list-before` are processed in a bad order.

This solution creates a function which is called for every title in the list instead of just a loop. That way, this function can be recursively called on titles further down the list if they need to be placed before the current title can be ordered around it.

For tests, I created new tests that individually test the problems, instead of working with the amalgamate test that was already there. In my experience, those kind of individual tests are easier to diagnose and understand when something goes wrong. Also added a test for the blank `list-after` feature added last month.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -559,10 +559,10 @@ exports.sortByList = function(array,listTitle) {
 		}
 		// Finally obey the list-before and list-after fields of each tiddler in turn
 		var sortedTitles = titles.slice(0),
-			replacedTitles = {},
+			replacedTitles = Object.create(null),
 			self = this;
 		function replaceItem(title) {
-			if(!replacedTitles[title]) {
+			if(!$tw.utils.hop(replacedTitles, title)) {
 				replacedTitles[title] = true;
 				var newPos = -1,
 					tiddler = self.getTiddler(title);
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -558,39 +558,49 @@ exports.sortByList = function(array,listTitle) {
 			}
 		}
 		// Finally obey the list-before and list-after fields of each tiddler in turn
-		var sortedTitles = titles.slice(0);
-		for(t=0; t<sortedTitles.length; t++) {
-			title = sortedTitles[t];
-			var currPos = titles.indexOf(title),
-				newPos = -1,
-				tiddler = this.getTiddler(title);
-			if(tiddler) {
-				var beforeTitle = tiddler.fields["list-before"],
-					afterTitle = tiddler.fields["list-after"];
-				if(beforeTitle === "") {
-					newPos = 0;
-				} else if(afterTitle === "") {
-					newPos = titles.length;
-				} else if(beforeTitle) {
-					newPos = titles.indexOf(beforeTitle);
-				} else if(afterTitle) {
-					newPos = titles.indexOf(afterTitle);
-					if(newPos >= 0) {
-						++newPos;
+		var sortedTitles = titles.slice(0),
+			replacedTitles = {},
+			self = this;
+		function replaceItem(title) {
+			if(!replacedTitles[title]) {
+				replacedTitles[title] = true;
+				var newPos = -1,
+					tiddler = self.getTiddler(title);
+				if(tiddler) {
+					var beforeTitle = tiddler.fields["list-before"],
+						afterTitle = tiddler.fields["list-after"];
+					if(beforeTitle === "") {
+						newPos = 0;
+					} else if(afterTitle === "") {
+						newPos = titles.length;
+					} else if(beforeTitle) {
+						replaceItem(beforeTitle);
+						newPos = titles.indexOf(beforeTitle);
+					} else if(afterTitle) {
+						replaceItem(afterTitle);
+						newPos = titles.indexOf(afterTitle);
+						if(newPos >= 0) {
+							++newPos;
+						}
 					}
-				}
-				if(newPos === -1) {
-					newPos = currPos;
-				}
-				if(newPos !== currPos) {
-					titles.splice(currPos,1);
-					if(newPos >= currPos) {
-						newPos--;
+					// We get the currPos //after// figuring out the newPos, because recursive replaceItem calls might alter title's currPos
+					var currPos = titles.indexOf(title);
+					if(newPos === -1) {
+						newPos = currPos;
 					}
-					titles.splice(newPos,0,title);
+					if(newPos !== currPos) {
+						titles.splice(currPos,1);
+						if(newPos >= currPos) {
+							newPos--;
+						}
+						titles.splice(newPos,0,title);
+					}
 				}
 			}
-
+		};
+		for(t=0; t<sortedTitles.length; t++) {
+			title = sortedTitles[t];
+			replaceItem(title);
 		}
 		return titles;
 	}
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -588,7 +588,7 @@ exports.sortByList = function(array,listTitle) {
 					if(newPos === -1) {
 						newPos = currPos;
 					}
-					if(newPos !== currPos) {
+					if(currPos >= 0 && newPos !== currPos) {
 						titles.splice(currPos,1);
 						if(newPos >= currPos) {
 							newPos--;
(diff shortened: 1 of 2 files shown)


flibbles:
HOLD ON! Found a bug that results from this. It's a little obscure, but I'll fix it as soon as I can.

flibbles:
...And corrected. Good to merge.

There was a small problem that could occur if a tiddler had a `list-after` that referenced another tiddler that _wasn't_ part of the list, and that tiddler had a blank `list-after` or `list-before`.

It's good now though. And a test is included.

core/modules/wiki.js
```diff
@@ -558,39 +558,49 @@ exports.sortByList = function(array,listTitle) {
...
-				} else if(afterTitle === "") {
-					newPos = titles.length;
-				} else if(beforeTitle) {
-					newPos = titles.indexOf(beforeTitle);
-				} else if(afterTitle) {
-					newPos = titles.indexOf(afterTitle);
-					if(newPos >= 0) {
-						++newPos;
+		var sortedTitles = titles.slice(0),
+			replacedTitles = {},
```
Jermolene: The core avoids this pattern because of the `var a={};console.log(!!a["__proto__"])` problem (it returns true!). The workaround is to use `$tw.utils.hop()`, a shortcut for `Object.hasOwnProperty`.

core/modules/wiki.js
```diff
@@ -558,39 +558,49 @@ exports.sortByList = function(array,listTitle) {
...
-					newPos = titles.indexOf(beforeTitle);
-				} else if(afterTitle) {
-					newPos = titles.indexOf(afterTitle);
-					if(newPos >= 0) {
-						++newPos;
+		var sortedTitles = titles.slice(0),
+			replacedTitles = Object.create(null),
+			self = this;
+		function replaceItem(title) {
+			if(!$tw.utils.hop(replacedTitles, title)) {
```
flibbles: Tested and corrected. Object.create and $tw.utils.hop are now used.

flibbles:
Tested and corrected. Object.create and $tw.utils.hop are now used.

(I am _not_ used to github's inline review features. I'm just shotgunning comments and reviews all over this PR.)

flibbles:
Also, let me know if you prefer if I squash the PR into a single commit. I don't know the etiquette here, and I know some projects prefer that.

pmario:
>This solution creates a function which is called for every title in the list instead of just a loop. That way, this function can be recursively called on titles further down the list if they need to be placed before the current title can be ordered around it.

I didn't test the code yet, but "recursively" calling a sort function will immediately fail. 

A (list-after B) ... B (list-after A) -> endless loop. This can't be easily resolved.

The "list-after" parameter for list sorting has been designed in TW beta days, where the template tiddlers changed very often. So if a user did change a "shadow" template tiddler is was cloned and therefore this "system" tiddler allways won over the next updated "shadow" tiddler. ... This behaviour caused problem in the early days.

That's why we needed a way to modify list order, without modifying shadow tiddlers. ... I think your PR doesn't take this behaviour into account, but imo it should.

flibbles:
It won't infinite loop. The following lines guard against that.

```javascript
if(!$tw.utils.hop(replacedTitles, title)) {
  replacedTitles[title] = true;
```

In your example, A will recursively call B, which will recursively call A, but A is already in `replacedTitles`, so it aborts. True, the ordering of such a tagList is undefined, but my changes will handle it no less gracefully than before.

I have a test on my computer testing your scenario. If anyone would like, I can include that test with the PR.

flibbles:
Also, I'm not sure about other people, but I've been making great use of the `list-after` and `list-before` feature with my own projects. Besides just non-shadow use, I'm using the feature with in plugins that add to `$:/tags/ViewTemplate` tagList, so my plugins can add to the view without modifying core.

core/modules/wiki.js
```diff
@@ -558,39 +558,49 @@ exports.sortByList = function(array,listTitle) {
...
+				replacedTitles[title] = true;
+				var newPos = -1,
```
flibbles: The upper two lines guard against infinite loops.

pmario:
>... I'm using the feature with in plugins that add to $:/tags/ViewTemplate tagList, so my plugins can add to the view without modifying core.

That's exactly the usecase it was designed for.

flibbles:
> That's exactly the usecase it was designed for.

Right. I guess I was confused by what you were saying. It sounded like you were implying that `list-after` and `list-before` were just a niche tool to solve an old, forgotten problem. I was worried that line of thought was moving in the direction of deprecating those features.

core/modules/wiki.js
pmario: That's ok. .. BUT where is the difference now, between your algorithm and the existing one. ... 

If the users expectes a different sorting result as the algorithm can create, the user still needs to manually fix it. ... 

So why shouldn't we tell them in the first place, that they have to manually fix the relations.

pmario:
My main concern here is performance! We use lists everywhere. .. and in my opinion the existing core UI should be improved performance wise!

core/modules/wiki.js
flibbles: Circular dependencies still produce undefined behavior.

My fix aims to correct issues that may occur with chained dependencies.
```
A (list-after B)
B (list-after C) 
```
This simple case _does_ have defined behavior, but because the old algorithm always processed them in order, it would mess that example up. A would not be after B. That's what recursive calls fix.

flibbles:
Yeah. In running through the code, I did notice that Tiddlywiki calls `sortByList` approx 1,000,000X per keyboard input.

My change, while recursive, still maintains the same O(n) speed. It's adding an extra check to each iteration through the tagList, and it fills up a Hash with all the listItems by the end. I haven't profiled, but it doesn't seem that expensive considering every `list-after` and `list-before` is "splicing the `titles` list, effectively duplicating the list anyway, and it also creates `sortedTitles` already, another duplicate.

I _can_ make this more efficient by removing the old duplication of `titles` into `sortedTitles` with some fancy iterator manipulation. (Instead of creating "sortedTitles" to iterate so we don't have to deal with `titles` changing under our feet, we can just "deal" with `titles` changing under our feet, if that makes sense)

pmario:
>This simple case does have defined behavior, but because the old algorithm always processed them in order, it would mess that example up. A would not be after B. That's what recursive calls fix.

I think you usecase is valid. ... But I would like to have a possibility to fix it, without adding CPU cycles to the core.

pmario:
>I can make this more efficient by removing the old duplication of titles into sortedTitles with some fancy iterator manipulation.

simple to understand imo is preferable!

flibbles:
> simple to understand imo is preferable!

Okey Dokey. I can get that done tomorrow, but if you still want to take a crack at it, that's fine too.

pmario:
I'm not 100% sure if this test touches all possible code paths. .. It seems to sort tiddler 250 after 50 ... right?
So the recursion won't be touched, right?

pmario:
I'll have a closer look

flibbles:
That's true. It only has a single recursive path that terminates immediately. In all of the shadow tiddler tagLists that exist, there are only two shadows that have a `list-after` or `list-before`, so heavily testing it wouldn't really make sense if we're worried about how taxing the new sortAsList will have on core functionality. In 99% of the cases, recursion won't happen at all, so I didn't test it in my load test.

flibbles:
If no one has any reason why this shouldn't be merged, I would voice again that this really should be included. It fixes an existing bug, And it's been three months.

flibbles:
I might be ignorant of the the way Tiddlywiki issues are addressed and merged (maybe they are revisited near release), but I'm concerned that this bug has fallen through the cracks. And it's a critical issue for the likes of me.

@pmario, if you're still concerned about the performance of this fix, then I'd recommend opening another issue that discusses optimizing core code such as the portion that this fix touches. However, I still believe that the performance hit for this fix is negligible. I've been using this fix on my fork for ages now.

If this issue is already slated to be considered for next release, just let me know, and I'll stop raising a holler.

flibbles:
@Jermolene, I'm not familiar with the process. But I'd like to promote this PR for the v5.1.18 final release. It's well unit tested and field tested.

If I'm following the wrong protocol, please forgive me.

Jermolene:
Thanks @pmario @flibbles

BurningTreeC:
Hi @flibbles, if you read this - we've refactored your code a little bit to fix #3555, if you want to have a look if it's OK :+1: 

all the best

PR #3305, opened 2018-05-21
Update cla-individual.md

AnthonyMuscio:
I would like to assist in the development of TiddlyWiki Documentation by providing WikiText examples and generic forms of a given widget call so users can copy to clipboard and use this as basis for their own code.


Jermolene:
Hi @AnthonyMuscio welcome, great to see you here – I think you've now been involved with TiddlyWiki for more than a decade.

PR #3317, opened 2018-05-31
doc "tip" "warning" and "note" boxes

BurningTreeC:


Jermolene:
Hi @BurningTreeC thanks - can you share a before and after picture to show what these changes do?

BurningTreeC:
Hi @Jermolene, I once made this mockup-page for that purpose: http://dochighlighting.tiddlyspot.com/

Jermolene:
Hi @BurningTreeC thanks, that's very helpful, now I remember some discussion about this before. I think there are a couple of minor issues to address.

Firstly, the top margin of the icons seems too small; it feels to me like the top of the icon should align with the top of the text:

image

Secondly, I think the boxes need to have width 100% as it looks odd to have different widths (also shown in the screenshot above).

Thirdly, I'm not convinced that reversing out the icons works well. The lightbulb one looks wrong because it is now a dark lightbulb on a light background:

image

Perhaps the usual black versions of the icons could be used: there would still be reasonable contrast with the background.

BurningTreeC:
@Jermolene , I believe this is now how you imagined it

BurningTreeC:
Mockup: http://doctipnotewarning.tiddlyspot.com/

00SS:
I would imagine the explanation tiddler for all this would need to be updated too: Documentation Macros.tid

I was working on it earlier in reference to another pull request - so it has many changes - I added in <<.note ... >> as well.
Something like the attached TXT file which is actually a .tid file.
[Documentation Macros.txt](https://github.com/Jermolene/TiddlyWiki5/files/2719233/Documentation.Macros.txt)

kookma:
@BurningTreeC  and @Jermolene 

I have developed some macro for this and used the common name as info, warning, danger, primary, secondary

https://kookma.github.io/Shiraz/#%24%3A%2Fplugins%2F.mr%2Fshiraz%2Fdemo%2Falert
https://kookma.github.io/Shiraz/#%24%3A%2Fplugins%2F.mr%2Fshiraz%2Fdemo%2Falert-hline

Also some other notes

https://kookma.github.io/Shiraz/#%24%3A%2Fplugins%2F.mr%2Fshiraz%2Fdemo%2Fnote

The icons are note mines but have no licenses (at least I did not see them

I would be happy to be of any help.

Cheers

BurningTreeC:
@kookma , @all 

Now I've thought about this and I'm more in favor of restructuring the documentation than to add too much of additional eye-catchers

I've read about [Banner blindness](https://en.wikipedia.org/wiki/Banner_blindness) and figured that that's my behavior and it must be the same for many others

So I believe we should put such things on ice and concentrate on well-structured documentation first.

00SS:
I agree with both thoughts. The banners style of presenting info can draw attention, but can also be ignored/blocked out, and can tend towards being distracting. 

So, back to concentrating on a more rewarding task - restructuring - some reorganising - tweaking the display format.

The more I am looking at the documentation  - the less daunting it feels - as there is so so so much already there - 90% of the work has been done - it actually will not take all that much to lay it out in a way that is friendlier to new users.

Experienced users already know how to find what they want quite quickly. They do not have much problem digging out the answers. It might be a little frustrating, but being experienced, they probably won't give up easily.

Optimising the first 20 to 30 topics for the newer users - allowing them to get familiar before getting lost - could be a goal to strive for.

I really wish there would be a little feedback on issue #3664  - if only to know if this is useful - or if it is not the right way.

00SS:
@BurningTreeC

Do you think making simple link-lists into an internal/external-nav toc is a good idea for the website?

BurningTreeC:
Hi @00SS 

> Do you think making simple link-lists into an internal/external-nav toc is a good idea for the website?

No, not really.. That's one of the problems why this is more difficult than it appears :grin: 

Q: What is the problem with a simple link list?
A: It can become very long

By using a toc-internal-nav (external) to address that we ignore simpler solutions like figuring out if we can break a list and show it side by side

00SS:
@BurningTreeC

That's a great point! I never thought about it that way.
If we could break a list and show it side by side, the list would be just as long though it would display using a less long (length wise) tiddler. 

The issue might still be the same though - how would a user open the tiddlers behind each link? 

Some possibilities:

1: Click a link - the page scrolls down to show that tiddler. After reading - close that tiddler. The page does not scroll back up to the list of links. So, scroll back up manually, identify the last clicked tiddler, and repeat.

2: CTRL click each link one by one. This opens the clicked tiddlers in reverse order below the list-links tiddler - the latest clicked one just below it. Scrolling right to the bottom and reading from there, then closing the tiddler has the same problem of the next tiddler not being displayed properly - as it is above and not below.

3: CTRL click in reverse order - this solves the scrolling issue.

I wonder if this is the right way to find what is needed from a long list.

With a toc-internal-nav - each tiddler opens in the same place - with the list of links to the left - one can scan each tiddler quickly - decide if it's worth reading - move to the next in the list quickly. After all - isn't this what the toc-internal-nav style of Table of Contents is designed to do?

PR #3320, opened 2018-05-31
fix "propogate" typo + add highlighting

BurningTreeC:


Evolena:
I'm not sure "tip" is the best one for highlighting the general purpose of a widget/macro/... It should be reserved to actual tips.
The new "note" highlight that you've just added would better fit this usage. Or maybe another type of macro, with a colored box but not necessarily an icon.

BurningTreeC:
@Evolena yes that may be better, I'd let others decide about that: I can change it to use the new note macro which displays the core (motovun jack) icon or add a "simplenote" macro with the same colorbox but no icon

... @Jermolene @pmario @allvoters

PR #3324, opened 2018-06-01
[doc] Rearrange ListWidget.tid

twMat:
Move section "Content and Attributes" to before "Examples" to be consistent with several other widget docs (e.g RevealWidget). 

The Examples section is better at bottom because we can then more freely add more examples. (I intend to add example of how to make html tables using the ListWidget.)


twMat:
If I understand right, #3802  makes this draft here superfluous so I'm closing.

PR #3328, opened 2018-06-07
tiddler-body: wrap around floating elements

BurningTreeC:
found this solution here: http://maxdesign.com.au/articles/floatsample/

when an element within a tiddler is made floating this fix forces the tiddler body to wrap around it

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1055,6 +1055,19 @@ canvas.tc-edit-bitmapeditor  {
 	clear: both;
 }
 
+.tc-tiddler-body:before, .tc-tiddler-body:after {
+	content: "";
+	display: table;
+}
+ 
+.tc-tiddler-body:after {
+	clear: both;
+}
+ 
+.tc-tiddler-body {
+	zoom: 1;
+}
+
 .tc-tiddler-frame .tc-tiddler-body {
 	font-size: {{$:/themes/tiddlywiki/vanilla/metrics/bodyfontsize}};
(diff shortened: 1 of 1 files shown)


BurningTreeC:
@Jermolene , reopening this because it might be useful, especially for new users

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1055,6 +1055,19 @@ canvas.tc-edit-bitmapeditor  {
...
+	content: "";
+	display: table;
+}
+ 
+.tc-tiddler-body:after {
+	clear: both;
+}
+ 
+.tc-tiddler-body {
+	zoom: 1;
```
pmario: This is marked with IE6 / 7 ... which we can ignore, since TW won't work with them anyway.

pmario:
see: https://css-tricks.com/snippets/css/clear-fix/

this should be enough. .... 

```
.clearfix:after {
  content: "";
  display: table;
  clear: both;
}
```

... but it's not usable for users. .... So we should create a proper clearfix class, that can be used everywhere. 

It would be needed, to assign the `clearfix` class to the tiddler body. Which means the ViewTemplate would be changed.

pmario:
@BurningTreeC ... Why did you close it. I think it's worth fixing. ... but in a more generic way.

BurningTreeC:
reopening...

@pmario, would you make the PR with your fix?

BurningTreeC:
@pmario, do you want to add a fix for tiddler bodies not wrapping around floating elements?

PR #3331, opened 2018-06-10
[print-tiddler] Add print-river using a PrintTemplate

BurningTreeC:
this uses a template tiddler `$:/core/ui/PrintTemplate` that lists all tiddlers that users add to the Print-List `$:/PrintList` by clicking the print-tiddler button

Tiddlers to print are shown in a simple "print-river" where Titles, Subtitles and Tags can be hidden/shown, page-breaks can be added and Tiddlers can be sorted reusing the `Sidebar/Open` mechanism

In order to reuse that `Sidebar/Open` mechanism some tweaks are made in `$:/core/ui/PageTemplate`, `$:/core/ui/PageTemplate/story` and `$:/core/ui/SideBar/Open`:
- in the base Page Template `<<storyList>>` and `<<historyList>>` variables are defined (transcluding the config tiddlers `$:/config/storylist`and `$:/config/historylist`) that get used by the navigator widget there, the list widget in the story PageTemplate and in the `$:/core/ui/SideBar/Open` tiddler
- in the Print Template those variables are set to `$:/PrintList` and `$:/PrintHistory` which results in the transcluded `$:/core/ui/SideBar/Open` tiddler to sort and navigate the Print List

Two icons are planned, one for the Sort-Dropdown and one for the PageBreak buttons. Because of licensing questions atm they are text-buttons


BurningTreeC:
Note: we'll need to take a look at the popup-cancelling mechanism in order to make it work in new windows

themes/tiddlywiki/vanilla/print.tid
```diff
@@ -0,0 +1,195 @@
...
+
+</$reveal>
+
+\end
+
+.tc-print-tiddler-window-overlay p {
+	display: inline-block;
+	margin: 0;
+	-webkit-margin-before: 0em;
+	-webkit-margin-after: 0em;
```
pmario: Is it only `-webkit` or do other browsers need prefixed parameters too?

themes/tiddlywiki/vanilla/print.tid
```diff
@@ -0,0 +1,195 @@
...
+
+.tc-print-tiddler-list-view .tc-print-tiddler-button {
+	display: none;
+}
+
+.tc-print-sort-drop-down {
+	text-align: left;
+	-webkit-box-shadow: 2px 2px 10px rgba(0, 0, 0, 0.5);
+	-moz-box-shadow: 2px 2px 10px rgba(0, 0, 0, 0.5);
+	box-shadow: 2px 2px 10px rgba(0, 0, 0, 0.5);
```
pmario: I think `box-shadow` https://caniuse.com/#search=box-shadow doesn't need prefixing anymore. ... IMO we don't need shadows for printing. .. Can / Should we remove it?

themes/tiddlywiki/vanilla/print.tid
```diff
@@ -0,0 +1,195 @@
...
+tags: $:/tags/Stylesheet
+
+\define set-page-break-tiddlers()
+
+@media print {
+	[data-tiddler-title='$(currentTiddler)$'].tc-tiddler-frame {
+		page-break-before: always;
+	}
+}
+
```
pmario: do we need those empty lines?

themes/tiddlywiki/vanilla/print.tid
BurningTreeC: it's only -webkit as I've read on a source - let me see if I find the link. Seems to be something some browsers prefer over "margin-top" and "margin-bottom" ... chrome does, firefox does not

themes/tiddlywiki/vanilla/print.tid
BurningTreeC: this box-shadow is for the "sort" dropdown in the small top print-toolbar only.
for the tiddler frames it's removed

themes/tiddlywiki/vanilla/print.tid
BurningTreeC: No I don't think that we need them, I'll format it better

themes/tiddlywiki/vanilla/print.tid
BurningTreeC: I added these styles because my browser added such `-webkit-margin-before` and after on `<p>` tags. Now I realize that it's just some sanitizing of vanilla/base that's missing in the new window.

BurningTreeC: Or maybe it's not missing sanatizing but something the browser applies there but not in the root window. For tables I had to add separate font-size and line-height styles because they got some default style applied.

Sounds like missing sanatizing, but is it?

pmario: >Sounds like missing sanatizing, but is it?

may be possible.

themes/tiddlywiki/vanilla/print.tid
pmario: ah. ok

BurningTreeC:
I thought it might be good to be able to pass an attribute to tm-open-window to set a window title that will be used for the document's name when it gets printed. Possibly the SiteTitle + a timestamp

BurningTreeC:
I've made a PR #3332 that handles cancelling popups in new windows

Then I've made a small adjustment in the windows.js startup module that enables specifying the document-title for the new window by passing `documentTitle` through the paramObject hashmap of `tm-open-window`. This document-title gets used as default filename when printing to pdf

PR #3334, opened 2018-06-16
make tm-full-screen work in all windows

BurningTreeC:
this makes tm-full-screen work in new windows, too


Jermolene:

PR #3335, opened 2018-06-17
make modals display in their corresponding window

BurningTreeC:
this makes modals display in the window where they were triggered

### DIFF
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -26,9 +26,9 @@ Display a modal dialogue
 Options include:
 	downloadLink: Text of a big download link to include
 */
-Modal.prototype.display = function(title,options,currentdocument) {
+Modal.prototype.display = function(title,options,currentDocument) {
 	options = options || {};
-	currentdocument = currentdocument || document;
+	currentDocument = currentDocument || document;
 	var self = this,
 		refreshHandler,
 		duration = $tw.utils.getAnimationDuration(),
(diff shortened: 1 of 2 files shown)


core/modules/utils/dom/modal.js
```diff
@@ -26,8 +26,9 @@ Display a modal dialogue
 Options include:
 	downloadLink: Text of a big download link to include
 */
-Modal.prototype.display = function(title,options) {
+Modal.prototype.display = function(title,options,currentdocument) {
 	options = options || {};
+	currentdocument = currentdocument || document;
```
pmario: IMO should be `currentDocument` if possible from the API

Jermolene:
Hi @BurningTreeC good catch -- it's great that you're finding so many wrinkles with the multiwindow support. It was added fairly recently; perhaps we should have audited usages of document and window and been more methodical about tracking down problems.

I think @pmario is right that `currentDocument` would be preferred.

BurningTreeC:
@Jermolene - thanks, I'd still like to find some more ... anyway, I've changed the name to "currentDocument"

I understand that parts of the core are designed without multi-window functionality in mind but I want to tackle what I think I understand. The storyviews are tricky, though.

I've made [another PR](#3332) recently for cancelling popups in new windows... would you mind having a look at it to see if it's any good if you find the time?

Jermolene:
Hi @BurningTreeC I hope everything is OK; I see you've been closing quite a few PRs, is that intentional? I can understand if you've grown frustrated with the delays with merging, but would like to resolve things more constructively if possible.

PR #3355, opened 2018-07-13
[doc update] Custom_styles_by_data-tags.tid

twMat:
A few minor improvements that I propose are merged directly.

But, @pmario , we also need a clarifying example showing how to target a tiddler using *multiple* tags. So, under the examples there ought to be something like this:... BUT I fail to get this to work (syntax?):

```
[data-tags*="[[How to apply custom styles]] example-test"] {
  outline: 2px solid green;
}
```

Further, the text states that

`Important: Tiddler tags are not sorted, so the order in the rendered output may be different!`

...but I'm not sure what this means. Does it mean:

```
Important: A tiddlers tags are displayed alphabetically but rendered styles may come out in a different order. As per standard CSS behaviour, styles rendered later override styles rendered earlier.
```


Jermolene:
Thanks @twMat

PR #3359, opened 2018-07-15
feat(utils): add set tags

danielo515:
added a small helper function to override the tags of a tiddler.
This is something that I find doing all the time, and I wanted to have an easier way. The code is simple, but the amount of work it saves it's big.
I added a function description using JSDoc format, not sure if you are ok with it.

If think this is not the correct place for such function, please let me know

Regards

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -799,5 +799,22 @@ exports.strEndsWith = function(str,ending,position) {
 		return lastIndex !== -1 && lastIndex === position;
 	}
 };
+	
+/**
+ * @function setTags
+ * @description Helper function to override the tags of a tiddler
+ * If the tiddler does not exist it will be created
+ * @param  {String} title Tiddler you want to override its' tags
+ * @param  {[String]} tags  List of tags that will be the new tiddler tags
+ * @return {Void} This function does not return anything
+ */
+exports.setTags = (title, tags) => {
+    $tw.wiki.addTiddler(
+        new $tw.Tiddler(
+            $tw.wiki.getTiddler(title),
+            {title: title, tags: tags}
+        )
+    )
+};	
 
 })();


core/modules/utils/utils.js
```diff
@@ -799,5 +799,22 @@ exports.strEndsWith = function(str,ending,position) {
 		return lastIndex !== -1 && lastIndex === position;
 	}
 };
+	
+/**
+ * @function setTags
+ * @description Helper function to override the tags of a tiddler
+ * If the tiddler does not exist it will be created
```
pmario: If tiddlers are created, they need the `created` and `modified` fields, otherwise, they don't show up in the "recent" tab
If tiddlers are changed, the modified field has to be updated

core/modules/utils/utils.js
```diff
@@ -799,5 +799,22 @@ exports.strEndsWith = function(str,ending,position) {
...
+ * @param  {String} title Tiddler you want to override its' tags
+ * @param  {[String]} tags  List of tags that will be the new tiddler tags
```
pmario: There is no check in the code, if the parameter is valid

pmario: IMO it should look more like: FieldManglerWidget.prototype.handleAddTagEvent

core/modules/utils/utils.js
danielo515: What do you think will be best ? To just handle the update and if the toddler does not exist return false (true in case of update) or to handle both scenarios ?

core/modules/utils/utils.js
pmario: I suggested, that your function should implement several checks, if the parameters are valid. Similar to: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/fieldmangler.js#L128

pmario: >Or are you suggesting to use that method to accomplish this task ?

no

Jermolene:
Thanks @danielo515 

Firstly, there are some minor issues: the function is a static utility method and hardwires a reference to the global wiki. It should probably be a method of the wiki object instead. Then we should consider whether it is generic enough: for example, it hardcodes the tags field, but might be useful with the list field.

But, a secondary consideration is that I'm not keen on adding utility functions to the core that aren't actually used by the core. Such functions are apt to become orphans, not properly maintained, and lack examples of usage for people reading the source.

Jermolene:
> I added a function description using JSDoc format, not sure if you are ok with it.

If we're going to adopt JSDoc then we need to do so consistently across the core.

ght:
No response in almost 3 years probably means that this can be closed.

danielo515:
Yeah, from now I will ask before putting any effort into any PR

Jermolene:
> Yeah, from now I will ask before putting any effort into any PR

Thanks @danielo515 I would always advise people to consult before investing in a complex PR.

I don't think that this is a case of the PR being ignored; I gave feedback explaining the issues I saw with the PR, and I don't think anything has changed since then.

PR #3360, opened 2018-07-16
Add <select> color setting to the palette

talha131:


pmario:
Hi "talha131". You should have pointed to the discussion, as some minimum info, which imo makes the PR valid.

see: https://groups.google.com/forum/#!topic/tiddlywiki/1lAK2VcMJQU

talha131:
Thank you @pmario for adding it. It slipped my mind.

Jermolene:
I found a minor issue with this PR: the explicit specification of the background and foreground colours for the select element forces some browsers to render the select controls differently.

For example, in Safari, the current TiddlyWiki select boxes look like this:

screenshot 2018-07-19 21 15 07

With this PR, they look like this:

screenshot 2018-07-19 21 15 44

So, I'm going to leave those palette entries blank by default to prevent this behaviour. Commit coming up.

PR #2679, opened 2016-12-29
Module-ize server routes and add static file support (#2510)

snlhnk:
* Refactor server routes to modules

New module type: serverroute

Caveats: Loading order is not deterministic but this would only matter
if two route modules attempted to use the same path regexp (that would
be silly).

* Add static assets plugin

This plugin allows the node server to fetch static assets in the /assets
directory. I felt that this was a feature that goes above the core
functionality. That is why I added it as a plugin. with the modular
route extensions this was a breeze.

* Add serverroute description to ModuleTypes

PR #3362, opened 2018-07-18
Allow all methods and add bodyFormat property to route definition

Arlen22:
This allows all methods to be used as routes and also adds a `bodyFormat` property to allow routes to specify which format they want the request body in. 

 - `string` is the current method of receiving the body as a string and is set as default.
 - `buffer` dumps the chunks into an array and then calls `Buffer.concat` with the array.
 - `stream` does nothing and immediately calls the route handler.

If a route doesn't specify `bodyFormat` it defaults to `string`. 

GET and HEAD requests are always treated as `stream` requests -- `bodyFormat` is ignored.

### DIFF
--- a/core/modules/server/server.js
+++ b/core/modules/server/server.js
@@ -200,21 +200,29 @@ Server.prototype.requestHandler = function(request,response) {
 	// TODO: Presumably this would need tweaking if we supported PUTting binary tiddlers
 	request.setEncoding("utf8");
 	// Dispatch the appropriate method
-	switch(request.method) {
-		case "GET": // Intentional fall-through
-		case "DELETE":
+	if(route.bodyFormat === "stream" || !route.bodyFormat || request.method === "GET" || request.method === "HEAD"){
+		//let the route handle the request stream itself if there is one
+		route.handler(request,response,state);
+	} else if(route.bodyFormat === "string"){
+		var data = "";
+		request.on("data",function(chunk) {
+			data += chunk.toString();
+		});
+		request.on("end",function() {
+			state.data = data;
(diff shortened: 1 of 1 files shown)


Arlen22:
The number one use case for the `bodyFormat` property is implementing server-sent events as a route. The second use case is handling uploads. I get the feeling this was going to be in the work -- maybe I beat you to the punch. :)

Arlen22:
Relates to https://github.com/Jermolene/TiddlyWiki5/pull/2679

core/modules/server/server.js
```diff
@@ -196,25 +196,33 @@ Server.prototype.requestHandler = function(request,response) {
...
-	// Set the encoding for the incoming request
-	// TODO: Presumably this would need tweaking if we supported PUTting binary tiddlers
-	request.setEncoding("utf8");
-	// Dispatch the appropriate method
-	switch(request.method) {
-		case "GET": // Intentional fall-through
-		case "DELETE":
+
+	//receive the request body if necessary and hand off to the route handler
+	if(route.bodyFormat === "stream" || request.method === "GET" || request.method === "HEAD"){
```
Jermolene: Shouldn't we check for "DELETE" too here?

Jermolene:
Thanks @Arlen22 much appreciated.

core/modules/server/server.js
Arlen22: Good question. I looked it up, and according to [this answer on Stack Overflow](https://stackoverflow.com/a/51112818/258482) the HTTP spec allows all methods except TRACE to have a request body.

PR #3364, opened 2018-07-19
Update chinese translations for built-in HTTP server

BramChen:
* Add chinese help texts for the new listen command
* Add chinese descriptions of "authenticator" and "route" in Docs/ModuleTypes
* Revised chinese help texts for the server command


Jermolene:
Thanks @BramChen

PR #3366, opened 2018-07-20
Update Chinese translations

BramChen:
* Add chinese help texts for the new listen command
* Add chinese texts for color setting of `<select>` element


Jermolene:
Thanks @BramChen

PR #3367, opened 2018-07-22
[keyboard widget]refresh if config tiddler changes

BurningTreeC:
this makes the keyboard widget refresh, if the config tiddler changes ("$:/config/shortcuts/...")
so one can change a shortcut for the text-editor without having to close and open the tiddler again in order to have the new shortcut applied

### DIFF
--- a/core/modules/widgets/keyboard.js
+++ b/core/modules/widgets/keyboard.js
@@ -79,6 +79,7 @@ KeyboardWidget.prototype.execute = function() {
 	this.tag = this.getAttribute("tag");
 	this.keyInfoArray = $tw.keyboardManager.parseKeyDescriptors(this.key);
 	this["class"] = this.getAttribute("class");
+	this.shortcutKeyDescriptor = this.key.match(/^\(\(.*\)\)$/g) ? this.key.replace(/^\(\(/,'').replace(/\)\)$/,'') : undefined;
 	// Make child widgets
 	this.makeChildWidgets();
 };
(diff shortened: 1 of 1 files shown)


core/modules/widgets/keyboard.js
```diff
@@ -88,7 +89,8 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 KeyboardWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.message || changedAttributes.param || changedAttributes.key || changedAttributes["class"] || changedAttributes.tag) {
+	if(changedAttributes.message || changedAttributes.param || changedAttributes.key || changedAttributes["class"] || changedAttributes.tag ||
+	  (this.shortcutKeyDescriptor && changedTiddlers["$:/config/shortcuts/" + this.shortcutKeyDescriptor])) {
```
pmario: hmmm. Is it needed to hardcode the `$:/config/shortcuts` tiddlers ?? ... When ever the store is changed, the refresh should be triggered automatically. ...

For me this seems to be "fighting the symptoms" of an issue and not fixing the issue!!

BurningTreeC: There are the different namespaces for the different operating systems... I've realized that now, therefor I need to change it... I can check for the os like it's done in parseKeyDescriptors()

In $:/core/modules/keyboard.js it's already sort of hardcoded, but it considers the different OS

PR #3369, opened 2018-07-23
[doc] Update DateFormat.tid

twMat:
The pre-requisite for the `format` attribute in order to use the `template` attribute has tripped me several times.

### DIFF
--- a/editions/tw5.com/tiddlers/features/DateFormat.tid
+++ b/editions/tw5.com/tiddlers/features/DateFormat.tid
@@ -4,7 +4,7 @@ tags: Features
 title: DateFormat
 type: text/vnd.tiddlywiki
 
-The ViewWidget accepts a `template` attribute that allows the format of date values to be specified. The format string is processed with the following substitutions:
+If the ViewWidget has a `format=date` attribute, the date can be formatted using the `template` attribute. Specify this format by setting `template="..."` where `...` is a combinaton of the following substitution strings:
 
 |!Token |!Substituted Value |
 |`DDD` |Day of week in full (eg, "Monday") |


editions/tw5.com/tiddlers/features/DateFormat.tid
```diff
@@ -4,7 +4,7 @@ tags: Features
 title: DateFormat
 type: text/vnd.tiddlywiki
 
-The ViewWidget accepts a `template` attribute that allows the format of date values to be specified. The format string is processed with the following substitutions:
+If the ViewWidget has a `format=date` attribute, the date can be formatted using the `template` attribute. Specify this format by setting `template="..."` where `...` is a combinaton of the following substitution strings:
```
pmario: I'm not sure, if this is an improvement.

pmario: There is no "if" ... The description tiddler should tell about facts. ... Examples can start with "If"

twMat:
But it is a fact. It's a conditional `if`. Should I change it to `Iff` ("If and only if")? How can it be communicated that `format=date` is a pre-requisite for `template` to work?

pmario:
>`format=date` is a pre-requisite for `template` to work?

... This needs to be included into https://tiddlywiki.com/#ViewWidget

twMat:
I'm beginning to see your point and will probably retract this PR. Possibly there should be a more "secondary" mention of where the date templates can be used, like in a "See also" list or similar.

Still I want to know how to solve this because it has confused me several times. In the ViewWidget, do **all** `template` values require that there is also a set a `format` attribute?

pmario:
>Still I want to know how to solve this because it has confused me several times. In the ViewWidget, do all template values require that there is also a set a format attribute?

Honestly ... I don't know.

twMat:
>Still I want to know how to solve this because it has confused me several times. In the ViewWidget, do all template values require that there is also a set format attribute?

@Jermolene - do you know?

I ask because that would much simplify rewriting the relevant parts in  the ViewWidget doc in a general way.

Jermolene:
Hi @twMat @pmario the "template" attribute of the view widget is ignored unless the "format" attribute is set to "date". I'll commit a clarification shortly

PR #3386, opened 2018-08-10
Improvements to the static single tiddler view as well as documentation.

Drakor:
This adds the following things:

* Path to the single tiddler view is now `/view/<tiddler title>`. This prevents name conflicts with api endpoints such as `/status` or `/login-basic` as well as any future plugin-defined paths (since the current regex almost always takes precedence)

* Added per tiddler options to overwrite the default render type and render template via the fields `_render_type` and `_render_template`

* Sanitized parameters to conform to the same format

This also comes with updated documentation.

/Andreas

### DIFF
--- a/core/modules/server/routes/get-tiddler-html.js
+++ b/core/modules/server/routes/get-tiddler-html.js
@@ -14,7 +14,7 @@ GET /:title
 
 exports.method = "GET";
 
-exports.path = /^\/view\/([^\/]+)$/;
+exports.path = /^\/([^\/]+)$/;
 
 exports.handler = function(request,response,state) {
 	var title = decodeURIComponent(state.params[0]),
--- a/core/modules/server/routes/get-tiddler-html.js
+++ b/core/modules/server/routes/get-tiddler-html.js
@@ -18,10 +18,10 @@ exports.path = /^\/view\/([^\/]+)$/;
 
 exports.handler = function(request,response,state) {
 	var title = decodeURIComponent(state.params[0]),
-	    tiddler = state.wiki.getTiddler(title);
+		tiddler = state.wiki.getTiddler(title);
 	if(tiddler) {
 		var renderType = tiddler.getFieldString("_render_type"),
-		    renderTemplate = tiddler.getFieldString("_render_template");
+			renderTemplate = tiddler.getFieldString("_render_template");
 		// Tiddler fields '_render_type' and '_render_template' overwrite
 		// system wide settings for render type and template
 		if(state.wiki.isSystemTiddler(title)) {
--- a/core/modules/server/routes/get-tiddler-html.js
+++ b/core/modules/server/routes/get-tiddler-html.js
@@ -14,22 +14,24 @@ GET /:title
 
 exports.method = "GET";
 
-exports.path = /^\/([^\/]+)$/;
+exports.path = /^\/view\/([^\/]+)$/;
 
 exports.handler = function(request,response,state) {
 	var title = decodeURIComponent(state.params[0]),
 		tiddler = state.wiki.getTiddler(title);
 	if(tiddler) {
-		var renderType,template;
-		// Render ordinary tiddlers as HTML, and system tiddlers in plain text
+        var renderType = tiddler.getFieldString("_render_type"),
+            renderTemplate = tiddler.getFieldString("_render_template");
+        // Tiddler fields '_render_type' and '_render_template' overwrite
+        // system wide settings for render type and template
 		if(state.wiki.isSystemTiddler(title)) {
-			renderType = state.server.get("system-tiddler-render-type");
-			template = state.server.get("system-tiddler-template");
+			renderType = renderType || state.server.get("system-tiddler-render-type");
+			renderTemplate = renderTemplate || state.server.get("system-tiddler-render-template");
 		} else {
-			renderType = state.server.get("tiddler-render-type");
-			template = state.server.get("tiddler-template");
+			renderType = renderType || state.server.get("tiddler-render-type");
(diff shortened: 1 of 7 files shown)


core/modules/server/routes/get-tiddler-html.js
```diff
@@ -18,12 +18,12 @@ exports.path = /^\/view\/([^\/]+)$/;
 
 exports.handler = function(request,response,state) {
 	var title = decodeURIComponent(state.params[0]),
-		tiddler = state.wiki.getTiddler(title);
+	    tiddler = state.wiki.getTiddler(title);
```
pmario: TW uses tabs as whitespace. tabs and spaces should __not__ be mixed!

Drakor: My apologies, I thought I fixed that earlier, should be cleaned up now.

Jermolene:
Thanks @Drakor looks good.

> * Path to the single tiddler view is now `/view/<tiddler title>`. This prevents name conflicts with api endpoints such as `/status` or `/login-basic` as well as any future plugin-defined paths (since the current regex almost always takes precedence)

This makes sense, but is a bit of a shame; I really like the clean "pure tiddler" URLs (the current URL encoding is ugly, but I plan to integrate a slugification module as an alternative).

So, I wonder about changing `/status` and `/login-basic` to make them contain slashes so that they don't clash with tiddler titles. For example, `/system/status` and `/system/login-basic`. There would be a minor backwards compatibility issue, but equally there's a lot of cruft in the current HTTP API, thanks to the early decision to remain compatible with TiddlyWeb, and so it might just be opening the floodgates to a long overdue overhaul of the API.

There are two very annoying issues from the top of my head:

* The way that unknown fields are moved down into a 'fields' property
* The use of tiddler fields to pass meta information like the revision number

One way to proceed would be by freezing/mothballing the current `tiddlywebadaptor`, and forking it to a new `tiddlywiki5adaptor`. But that would exacerbate the backwards compatibility issues; might be better to stick the (mis)naming "tiddlywebadaptor", and fork the tiddlyweb 1.x adaptor as "tiddlyweb1adaptor" instead.

Drakor:
> This makes sense, but is a bit of a shame; I really like the clean "pure tiddler" URLs (the current URL encoding is ugly, but I plan to integrate a slugification module as an alternative).

It is and you got a point, `/view/` was the next best thing that came to mind. I can change the path back for now, so that we can get the tiddler parameters in there, because thats the feature that I'd like to add. [1]

When it comes to the larger API issue at hand, I think if you plan to break compatibility, it might be best to do it right and keep the "legacy" `tiddlywebadaptor` and make a separate `tiddlywiki5adaptor` which then serves a completely new API where pure tiddler URL's could be the very centerpiece to PUT, GET and DELETE tiddlers and even `/status`, `/login-basic` and the likes could be turned into "tiddlers".

Another route I thought I'd mention here would be the other direction to completely implement the tiddlyweb API in the future and implement bags and recipes into the core and add abilities to interact and use them. Not sure if TW5 is headed this way though.


[1] My main motivation to introduce `/view/` was because the server route regex will almost always take precedence over any "user" added plugin routes that try to serve `/<something>`.

Jermolene:
Thanks @Drakor 

> When it comes to the larger API issue at hand, I think if you plan to break compatibility, it might be best to do it right and keep the "legacy" `tiddlywebadaptor` and make a separate `tiddlywiki5adaptor` which then serves a completely new API where pure tiddler URL's could be the very centerpiece to PUT, GET and DELETE tiddlers and even `/status`, `/login-basic` and the likes could be turned into "tiddlers".

I agree that that would be the logical approach. But the problem is that we've got 7 years of instructions and forum posts for TW5 that talk of tiddlywebadaptor. And meanwhile there is probably zero people actually using the tiddlywebadaptor to talk to TiddlyWeb. Hence my suggestion that we move legacy TiddlyWeb compatibility to a new plugin, and update the tiddlywebadaptor.

> Another route I thought I'd mention here would be the other direction to completely implement the tiddlyweb API in the future and implement bags and recipes into the core and add abilities to interact and use them. Not sure if TW5 is headed this way though.

That is my goal, but my plan is to first update the TW5 core wiki object to support revisions, bags and recipes, and then patch in the API.

editions/tw5.com/tiddlers/webserver/Using the read-only single tiddler view.tid
```diff
@@ -1,22 +1,22 @@
...
 
 TiddlyWiki's experimental single tiddler per page, read-only view uses a simplified page layout, and implements links between tiddlers, but there are no other interactive features. Compared to a full TiddlyWiki user interface, it is very lightweight and usable even over very slow connections.
 
-Alongside serving the full interactive wiki at the path `/` (e.g. http://127.0.0.1:8080/), TiddlyWiki serves each tiddler at the path `/<url-encoded-tiddler-title>`. For example:
+Alongside serving the full interactive wiki at the path `/` (e.g. http://127.0.0.1:8080/), TiddlyWiki serves each tiddler at the path `/view/<url-encoded-tiddler-title>`. For example:
 
-* http://127.0.0.1:8080/HelloThere
-* http://127.0.0.1:8080/Philosophy%20of%20Tiddlers
+* http://127.0.0.1:8080/view/HelloThere
+* http://127.0.0.1:8080/view/Philosophy%20of%20Tiddlers
```
Jermolene: Hi @Drakor I think these two examples should not have the `/view` part?

Drakor:
I apologize for the long delay, I believe I have fixed the documentation on this now.

Jermolene:
Thanks @Drakor apologies for the delay.

PR #3393, opened 2018-08-11
Updated documentation with description of filter operator each suffix: value

jho1965us:
This suffix to each was previous undocumented


Jermolene:
Thanks @jho1965us

PR #3403, opened 2018-08-18
Fixed type mismatch in get-file.js server route

Drakor:
Sorry for the small pull request, but this just fixes a typo, so that we now set the `ContentType` for file requests properly, if its a known file extension.

Please let me know whether these types of pull requests are more trouble than they are worth, I could have also just opened an issue. What's the preferred way of handling this?

/Andreas


core/modules/server/routes/get-file.js
```diff
@@ -38,7 +38,7 @@ exports.handler = function(request,response,state) {
 		} else {
 			status = 200;
 			content = content;
-			type = $tw.config.fileExtensionInfo[extension] || "application/octet-stream";
+			type = $tw.config.fileExtensionInfo[extension].type || "application/octet-stream";
```
Jermolene: Might we need to guard against `$tw.config.fileExtensionInfo[extension]` being undefined?

Jermolene:
Thanks @Drakor

core/modules/server/routes/get-file.js
Drakor: Indeed we do

PR #3411, opened 2018-08-25
Adding configuration options to tiddlywiki.info

ibnishak:
viz wikiTiddlersSubDir, wikiLanguagesSubDir, wikiThemesSubDir, wikiPluginsSubDir

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1818,7 +1818,7 @@ $tw.loadWikiTiddlers = function(wikiPath,options) {
 	$tw.loadPlugins(wikiInfo.themes,$tw.config.themesPath,$tw.config.themesEnvVar);
 	$tw.loadPlugins(wikiInfo.languages,$tw.config.languagesPath,$tw.config.languagesEnvVar);
 	// Load the wiki files, registering them as writable
-	var resolvedWikiPath = path.resolve(wikiPath,$tw.config.wikiTiddlersSubDir);
+	var resolvedWikiPath = path.resolve(wikiPath,wikiInfo.wikiTiddlersSubDir || $tw.config.wikiTiddlersSubDir);
 	$tw.utils.each($tw.loadTiddlersFromPath(resolvedWikiPath),function(tiddlerFile) {
 		if(!options.readOnly && tiddlerFile.filepath) {
 			$tw.utils.each(tiddlerFile.tiddlers,function(tiddler) {
@@ -1845,9 +1845,9 @@ $tw.loadWikiTiddlers = function(wikiPath,options) {
 		$tw.wiki.addTiddler({title: "$:/config/OriginalTiddlerPaths", type: "application/json", text: JSON.stringify(output)});
 	}
 	// Save the path to the tiddlers folder for the filesystemadaptor
-	$tw.boot.wikiTiddlersPath = path.resolve($tw.boot.wikiPath,config["default-tiddler-location"] || $tw.config.wikiTiddlersSubDir);
+	$tw.boot.wikiTiddlersPath = path.resolve($tw.boot.wikiPath,config["default-tiddler-location"] || wikiInfo.wikiTiddlersSubDir || $tw.config.wikiTiddlersSubDir);
 	// Load any plugins within the wiki folder
-	var wikiPluginsPath = path.resolve(wikiPath,$tw.config.wikiPluginsSubDir);
+	var wikiPluginsPath = path.resolve(wikiPath,wikiInfo.wikiPluginsSubDir || $tw.config.wikiPluginsSubDir);
 	if(fs.existsSync(wikiPluginsPath)) {
 		var pluginFolders = fs.readdirSync(wikiPluginsPath);
 		for(var t=0; t<pluginFolders.length; t++) {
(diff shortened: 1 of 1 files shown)


pmario:
I'm not happy with those hardcoded parameters! ... 

Why should I want to have your eg: unix directory configuration stored on my windows harddisk? ... or vice versa.

We do have environment variables for exactly that purpose. see: https://tiddlywiki.com/#Environment%20Variables%20on%20Node.js

Jermolene:
Hi @ibnishak can you explain a little of the motivation for the change? As @pmario suggests, using the environment variables might be an alternative approach?

ibnishak:
The focus is on "tiddlers" folder. If a user wants to change the default folder, currently there is no mechanism per se. The "default-tiddler-location" option is slightly misleading, as server will not read from the location, only write. There is no environment variable pertaining this.

The rest of the options(languages, themes, plugins etc) are added to bring uniformity to the workflow. However consider a situation where there are two groups of wikis subscribing to  different  plugin stores. Setting the environment variable each time user switches between wikis is an awkward solution. Setting it once at the level of a wiki is much more acceptable.  

However, coming back to the original issue. Setting an environment variable is not feasible regarding the tiddlers folder location. Currently that option is hardcoded to "./tiddlers" in the boot script. This PR adds the option to change this.

Now, regarding platform differences in paths, remember that tiddlywiki.info already contains options that involve paths, viz "includeWiki", "default-tiddler-locations" etc. All of these will fail when user switches  between windows-unix platforms. The obvious solution is to keep platform specific info files and synchronise the tiddlers folder. This PR will not add any more issues than those already existing regarding paths.


Finally, take note that none of the proposed changes break anything or makes it mandatory to set these options. They are all falling back to existing mechanisms and options unless user specifically adds any of these in their "tiddlywiki.info" file. So I do not see the issue in providing more configuration options to the user.

Jermolene:
> Setting the environment variable each time user switches between wikis is an awkward solution

I suspect that the environment variable approach implies having some sort of batch file arrangements to set things up in a convenient way.

> However, coming back to the original issue. Setting an environment variable is not feasible regarding the tiddlers folder location. Currently that option is hardcoded to "./tiddlers" in the boot script. This PR adds the option to change this.

I'm uncertain why this is needed. There's a general benefit to the community if everyones wiki folders look the same. The fixed subdirectory names provide a consistent baseline that makes it easier for users/developers to comfortably navigate any wiki folder.

pmario:
>The "default-tiddler-location" option is slightly misleading, as server will not read from the location, only write. There is no environment variable pertaining this.

The "includeWikis" parameter is responsible to load content from different locations.

pmario:
>Hoping you won't take offense, may I point out that not everyone might want to have a folder named "tiddlers" in their filesystem.
>
>If you stll feel against it, I understand.

There's no offense, in searching and exploring improved solutions! ...  

The issue I have with the PR is, that for me it seems to be very "use-case" specific. .. That's OK. .. But once we merge something to the core, we have to be sure, that it's implemented in a generic way. .. Because we have to live with it for a very long time :) ...  and we have an existing mechanism, that is very similar. ...

I'm not against a possibility to improve the configuration. I think we need to discuss the solution. ... 

IMO there is "value" in arguing "back and forth", discussing "pros and cons", because that's the way  how new ideas come up.

Jermolene:
> may I point out that not everyone might want to have a folder named "tiddlers" in their filesystem.

No offence taken, you're quite right: we do sometimes get the feedback that the terms "tiddler" and "tiddlywiki" make some people uncomfortable. If we thought it was a big problem I think we'd have to get rid of both terms, since they are both found all over a TW5 installation.

My main problem with this PR is the same problem I have with lots of prospective enhancements to boot.js: the code has grown by accretion from a modest 200 line boot kernel into a monolithic monster, and so backwards compatible enhancements tend to look messy and repetitive -- we have to patch constructions that are already quite unwieldy and hard to follow.

In cases like this it can be useful to pull back and consider whether we've got a strong enough conceptual model for what we've ended up trying to do.

I have been toying with a way of reconceptualising a "wiki folder". The idea is to adopt a pure JSON format that is capable of containing an entire wiki (both configuration and tiddlers), with a generic capability to drop out into the file system for any part of the JSON structure.

So, a simple all-in-one wiki might be represented as:

```
{
	"description": "Documentation from https://tiddlywiki.com",
	"tiddlers": {
		"HelloThere": {
			"title": "HelloThere",
			"text": "This is a tiddler."
		}
	}
}
```

Or, if the tiddlers were stored in a folder called "mydata" in the same folder as the tiddlywiki.info file:

```
{
	"description": "Documentation from https://tiddlywiki.com",
	"tiddlers": "./mydata"
}
```

The idea is that any object in the JSON can be replaced with a string giving a relative or absolute pathname to the folder or file containing that item. The system would understand further JSON files, or equivalent file formats like a redesigned .tid file format that can cope with multiline fields, and an update to the .meta file format.

The reason that all of this has been on my mind is because I'm hoping to add bag/recipe support to the TW5 server, which will involve significantly extending the model of a wiki folder.

Anyhow, perhaps we could experiment with a more modest goal along the lines of adding support for an optional "tiddlers" property of the "tiddlywiki.info" file that can either be an object map of tiddlers, or a string containing a filepath.

Issue #1938, opened 2015-09-14
Storyview Zoomin - remake to only one tid open at a time?


twMat:
Depending on the answer to this question, it is a proposal or not; 

The doc tiddler [Performance](http://tiddlywiki.com/#Performance) reads (my emphasis):

> Don't have too many tiddlers open at once. Every tiddler you have open will require processing to keep it up to date as the store changes (for example, while you type into a draft tiddler). **It is particularly easy when using zoomin story view to end up with dozens of tiddlers listed** in the Open tab in the sidebar. Get into the habit of periodically closing all open tiddlers with the close all button

Is there any particular reason why storyview Zoomin keeps the tiddlers open? I mean, they show one at a time so if the above is an issue could they not simply be closed an push/popped from history?

Admittedly, that doc tiddler has this as a suggestion for older/slower browers, but I'm thinking that if it is important enough to be 1 out of 4 suggestions _and_ it is a storyview that might be particularly suitable for **mobile**, then it's worth bringing it up.

If _"Yeah Mat, that's a brilliant observation."_... then consider this is a proposal.

Evolena:
I think that even in zoomin view it can still be useful to have access to opened tiddlers.

However, this "only one tiddler opened" could be a new view.

tobibeer:
Interesting point which seems to beg for a "last seen" tab instead of an "open" tab ...or for the latter to actually function as such when some **$:/config/num-display-tids** is set to 1.

pmario:
`$:/config/max-open-tids` could be an option.

Jermolene:
Thanks 3 year old @twMat! I've finally got around to doing this in #3412, so I'm closing this.

PR #3412, opened 2018-08-26
Add true "single tiddler mode"

Jermolene:
As @twMat noted in #1938 nearly three years ago, the current "zoomin" storyview suffers from some annoying limitations/quirks. This PR adds a proper "single tiddler mode" to the core in which only a single tiddler is displayed in the story river.

The changes involved are:

* Storyviews now expose a static "singleTiddlerMode" property to determine if they require single tiddler mode
* Navigator widget adds a "singleTiddlerMode" attribute; when set to "yes", the story list is entirely replaced during navigation
* Navigator widget now updates "list" field of $:/HistoryList with a more accessible version of the history list
* Add a getstoryviewsingletiddlermode[] operator so that we can obtain the story view STM status for passing to the navigator widget
* Add "History" sidebar tab because it is much more useful than the existing "Open" tab when in STM. However, I do not want to add another top level sidebar tab, so will need to find another way to handle this

There are still some issues to be resolved:

* The story river isn't yet adjusted when switching to a story view with STM enabled
* The usual zoomin animation is broken
* `$:/DefaultTiddlers` isn't yet adjusted, so multiple tiddlers are currently being opened at startup

### DIFF
--- a/core/modules/filters/getstoryviewsingletiddlermode.js
+++ b/core/modules/filters/getstoryviewsingletiddlermode.js
@@ -1,0 +1,35 @@
+/*\
+title: $:/core/modules/filters/getstoryviewsingletiddlermode.js
+type: application/javascript
+module-type: filteroperator
+
+Filter operator for retrieving the single tiddler mode status of a storyview.
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter functions
+*/
+
+exports.getstoryviewsingletiddlermode = function(source,operator,options) {
+	// Initialise the storyviews if they've not been done already
+	var storyviews = {};
+	$tw.modules.applyMethods("storyview",storyviews);
+	var results = [];
+	source(function(tiddler,title) {
+		var storyview = storyviews[title];
+		if(storyview && storyview.singleTiddlerMode) {
+			results.push("yes");			
+		} else {
+			results.push("no");
+		}
+	});
+	return results;
+};
+
+})();
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -116,7 +116,11 @@ NavigatorWidget.prototype.replaceFirstTitleInStory = function(storyList,oldTitle
 };
 
 NavigatorWidget.prototype.addToStory = function(title,fromTitle) {
-	this.wiki.addToStory(title,fromTitle,this.storyTitle,{openLinkFromInsideRiver: this.getAttribute("openLinkFromInsideRiver","top"),openLinkFromOutsideRiver: this.getAttribute("openLinkFromOutsideRiver","top")});
+	this.wiki.addToStory(title,fromTitle,this.storyTitle,{
+		openLinkFromInsideRiver: this.getAttribute("openLinkFromInsideRiver","top"),
+		openLinkFromOutsideRiver: this.getAttribute("openLinkFromOutsideRiver","top"),
+		singleTiddlerMode: this.getAttribute("singleTiddlerMode","no") === "yes",
+	});
 };
 
 /*
--- a/core/ui/PageTemplate.tid
+++ b/core/ui/PageTemplate.tid
@@ -22,7 +22,7 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 
 <div class=<<containerClasses>>>
 
-<$navigator story="$:/StoryList" history="$:/HistoryList" openLinkFromInsideRiver={{$:/config/Navigation/openLinkFromInsideRiver}} openLinkFromOutsideRiver={{$:/config/Navigation/openLinkFromOutsideRiver}} relinkOnRename={{$:/config/RelinkOnRename}}>
+<$navigator story="$:/StoryList" history="$:/HistoryList" singleTiddlerMode={{{ [{$:/view}getstoryviewsingletiddlermode[]] }}} openLinkFromInsideRiver={{$:/config/Navigation/openLinkFromInsideRiver}} openLinkFromOutsideRiver={{$:/config/Navigation/openLinkFromOutsideRiver}} relinkOnRename={{$:/config/RelinkOnRename}}>
 
 <$dropzone>
 
--- a/core/language/en-GB/SideBar.multids
+++ b/core/language/en-GB/SideBar.multids
@@ -4,6 +4,7 @@ All/Caption: All
 Contents/Caption: Contents
 Drafts/Caption: Drafts
 Explorer/Caption: Explorer
+History/Caption: History
 Missing/Caption: Missing
 More/Caption: More
 Open/Caption: Open
--- a/core/modules/story.js
+++ b/core/modules/story.js
@@ -37,44 +37,48 @@ Story.prototype.getStoryList = function() {
 Story.prototype.addToStory = function(navigateTo,navigateFromTitle,options) {
 	options = options || {};
 	var storyList = this.getStoryList();
-	// See if the tiddler is already there
-	var slot = storyList.indexOf(navigateTo);
-	// Quit if it already exists in the story river
-	if(slot >= 0) {
-		return;
-	}
-	// First we try to find the position of the story element we navigated from
-	var fromIndex = storyList.indexOf(navigateFromTitle);
-	if(fromIndex >= 0) {
-		// The tiddler is added from inside the river
-		// Determine where to insert the tiddler; Fallback is "below"
-		switch(options.openLinkFromInsideRiver) {
-			case "top":
-				slot = 0;
-				break;
-			case "bottom":
-				slot = storyList.length;
-				break;
-			case "above":
-				slot = fromIndex;
-				break;
-			case "below": // Intentional fall-through
-			default:
-				slot = fromIndex + 1;
-				break;
-		}
+	if(options.singleTiddlerMode) {
+		storyList = [navigateTo];
 	} else {
-		// The tiddler is opened from outside the river. Determine where to insert the tiddler; default is "top"
-		if(options.openLinkFromOutsideRiver === "bottom") {
-			// Insert at bottom
-			slot = storyList.length;
-		} else {
-			// Insert at top
-			slot = 0;
+		// See if the tiddler is already there
+		var slot = storyList.indexOf(navigateTo);
+		// Quit if it already exists in the story river
+		if(slot >= 0) {
+			return;
 		}
+		// First we try to find the position of the story element we navigated from
+		var fromIndex = storyList.indexOf(navigateFromTitle);
+		if(fromIndex >= 0) {
+			// The tiddler is added from inside the river
+			// Determine where to insert the tiddler; Fallback is "below"
+			switch(options.openLinkFromInsideRiver) {
+				case "top":
+					slot = 0;
+					break;
+				case "bottom":
+					slot = storyList.length;
+					break;
+				case "above":
+					slot = fromIndex;
+					break;
+				case "below": // Intentional fall-through
+				default:
+					slot = fromIndex + 1;
+					break;
+			}
+		} else {
+			// The tiddler is opened from outside the river. Determine where to insert the tiddler; default is "top"
+			if(options.openLinkFromOutsideRiver === "bottom") {
+				// Insert at bottom
+				slot = storyList.length;
+			} else {
+				// Insert at top
+				slot = 0;
+			}
+		}
+		// Add the tiddler
+		storyList.splice(slot,0,navigateTo);
 	}
-	// Add the tiddler
-	storyList.splice(slot,0,navigateTo);
 	// Save the story
 	this.saveStoryList(storyList);
 };
@@ -93,11 +97,20 @@ Story.prototype.saveStoryList = function(storyList) {
 Story.prototype.addToHistory = function(navigateTo,navigateFromClientRect) {
 	var titles = $tw.utils.isArray(navigateTo) ? navigateTo : [navigateTo];
 	// Add a new record to the top of the history stack
-	var historyList = this.wiki.getTiddlerData(this.historyTitle,[]);
+	var historyList = this.wiki.getTiddlerData(this.historyTitle,[]),
+		historyTitles = this.wiki.getTiddlerList(this.historyTitle);
 	$tw.utils.each(titles,function(title) {
 		historyList.push({title: title, fromPageRect: navigateFromClientRect});
+		var p;
+		do {
+			p = historyTitles.indexOf(title);
+			if(p !== -1) {
+				historyTitles.splice(p,1);
+			}
+		} while(p !== -1);
+		historyTitles.unshift(title);
 	});
-	this.wiki.setTiddlerData(this.historyTitle,historyList,{"current-tiddler": titles[titles.length-1]});
+	this.wiki.setTiddlerData(this.historyTitle,historyList,{"current-tiddler": titles[titles.length-1], list: historyTitles});
 };
 
 Story.prototype.storyCloseTiddler = function(targetTitle) {
--- a/core/modules/storyviews/zoomin.js
+++ b/core/modules/storyviews/zoomin.js
@@ -39,6 +39,9 @@ var ZoominListView = function(listWidget) {
 	});
 };
 
+// Engage single tiddler mode
+ZoominListView.singleTiddlerMode = true;
+
 ZoominListView.prototype.navigateTo = function(historyInfo) {
 	var duration = $tw.utils.getAnimationDuration(),
 		listElementIndex = this.listWidget.findListItem(0,historyInfo.title);
--- a/core/templates/single.tiddler.window.tid
+++ b/core/templates/single.tiddler.window.tid
@@ -12,7 +12,7 @@ title: $:/core/templates/single.tiddler.window
 
 <$importvariables filter="[[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]">
 
-<$navigator story="$:/StoryList" history="$:/HistoryList">
+<$navigator story="$:/StoryList" history="$:/HistoryList" singleTiddlerMode={{{ [{$:/view}getstoryviewsingletiddlermode[]] }}}>
 
 <$transclude mode="block"/>
 
--- a/core/ui/SideBar/History.tid
+++ b/core/ui/SideBar/History.tid
@@ -1,0 +1,5 @@
+title: $:/core/ui/SideBar/History
+tags: $:/tags/SideBar
+caption: {{$:/language/SideBar/History/Caption}}
+
+<$list filter="[list[$:/HistoryList]limit[100]]" template="$:/core/ui/ListItemTemplate"/>
--- a/core/wiki/tags/SideBar.tid
+++ b/core/wiki/tags/SideBar.tid
@@ -1,3 +1,3 @@
 title: $:/tags/SideBar
-list: [[$:/core/ui/SideBar/Open]] [[$:/core/ui/SideBar/Recent]] [[$:/core/ui/SideBar/Tools]] [[$:/core/ui/SideBar/More]]
+list: [[$:/core/ui/SideBar/Open]] [[$:/core/ui/SideBar/History]] [[$:/core/ui/SideBar/Recent]] [[$:/core/ui/SideBar/Tools]] [[$:/core/ui/SideBar/More]]
 
--- a/editions/tw5.com/tiddlers/workingwithtw/Performance.tid
+++ b/editions/tw5.com/tiddlers/workingwithtw/Performance.tid
@@ -1,5 +1,5 @@
 created: 20150330155120127
-modified: 20160607145222633
+modified: 20180824083527982
 tags: [[Working with TiddlyWiki]]
 title: Performance
 type: text/vnd.tiddlywiki
@@ -9,4 +9,4 @@ TiddlyWiki ships with defaults that are designed to get the best out of modern d
 * ''Avoid the "Recent" tab''. It is computationally slow to generate and update in response to tiddler changes.
 * ''Use the "Vanilla" theme''. The default "Snow White" theme includes visual effects like shadows, transparency and blurring that can be slow to render on older devices
 * ''Avoid large tiddlers''. Large bitmaps can significantly slow TiddlyWiki's performance. For example, an image taken with a modern smartphone will often be 5MB or more. Use ExternalImages whenever possible
-* ''Don't have too many tiddlers open at once''. Every tiddler you have open will require processing to keep it up to date as the store changes (for example, while you type into a draft tiddler). It is particularly easy when using zoomin story view to end up with dozens of tiddlers listed in the ''Open'' tab in the sidebar. Get into the habit of periodically closing all open tiddlers with the {{$:/core/images/close-all-button}} ''close all'' button
+* ''Don't have too many tiddlers open at once''. Every tiddler you have open will require processing to keep it up to date as the store changes (for example, while you type into a draft tiddler). It is easy to end up with dozens of tiddlers listed in the ''Open'' tab in the sidebar. Get into the habit of periodically closing all open tiddlers with the {{$:/core/images/close-all-button}} ''close all'' button, or use zoomin story view which only keeps one tiddler open at a time.


Jermolene:
There has been some discussion of these changes in this [Google Groups thread](https://groups.google.com/d/topic/tiddlywiki/gXAewvbZAX0/discussion).

Based on that feedback, I've posted an update to https://tiddlywiki.com/prerelease with a couple of further changes:

* The tiddler 'close' button now navigates to the previous tiddler in the history stack
* The "open" tab in the sidebar now mutates into a "history" tab when single tiddler mode is engaged

I consider this to be a major change to TiddlyWiki, and so would welcome feedback from everyone but I'll mention @giffmex @jdjdjdjdjdjd @telmiger as they may not be tracking this PR.

The questions I'm curious about:

* What do people think of the new behaviour of the "open" tab? It doesn't feel like a great solution to me -- there's no precedent for tabs coming and going elsewhere. But equally, I didn't like having "history" as a separate tab
* Is the navigating backwards action on "close tiddler" working as you'd expect?

It would also be interesting to test these changes against the kind of plugins and hacks that manipulate the story list.

giffmex:
Hi Jeremy

I finally got back in my office and went to try out the single tiddler
mode. I downloaded the prerelease. Now I am looking everywhere to find the
control to make it single tiddler mode. I tried the down arrow in the
tiddler viewtoolbar, the control panel, starting with the story view tab,
theme tab, and settings tab, the edit toolbar, the tools tab in the
sidebar, and a search for "single". Nothing. If you want me to try it out,
you have to tell me how to get to that mode. Blessings.

David Gifford
Mexico team leader, Mexico City

*Resonate Global Mission*
*Engaging People. Embracing Christ.*
A Ministry of the Christian Reformed Church
resonateglobalmission.org

giffmex:
BTW, two unrelated observations:

1) The tools tab in the sidebar should be in alphabetical order. There is
no clear logic to the order.

2) After being used to the way the Open tab works, the History tab makes me
feel like those are still open and need to be closed. But that is probably
just me. I don't think a brand new user would see it that way, only a long
term user. But I mention my reaction in case there is an easy way to
diminish that feeling.

Blessings.

David Gifford
Mexico team leader, Mexico City

*Resonate Global Mission*
*Engaging People. Embracing Christ.*
A Ministry of the Christian Reformed Church
resonateglobalmission.org

giffmex:
oops, sorry, I hit reply in my work email and that adds the long and religious signature that some people do not appreciate.

pmario:
@giffmex ... It's the "zooming" story view

Jermolene:
Hi Dave

Apologies, I forgot to push the latest update to tw.com <http://tw.com/>; there should be an introductory paragraph explaining what’s going on.

https://tiddlywiki.com/prerelease/ <https://tiddlywiki.com/prerelease/>

Best wishes

Jeremy

core/modules/filters/getstoryviewsingletiddlermode.js
```diff
@@ -0,0 +1,35 @@
...
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter functions
+*/
+
+exports.getstoryviewsingletiddlermode = function(source,operator,options) {
```
pmario: wouldn't it be better to have a function "getstoryviewmode", that returns the mode name? ... 

IMO we would need a new function for every other existing and future mode too!

Jermolene: Hi @pmario good point, I'll fix.

giffmex:
Ahh thank you, Mario and Jeremy. I assumed zoomin was going to continue being its own thing and that single tiddler mode was going to be separate. I have to run, but when I come back I will experiment and give feedback.

core/ui/PageTemplate.tid
```diff
@@ -14,7 +14,7 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 
 <$set name="storyviewTitle" value={{$:/view}}>
 
-<$set name="tv-storyview-single-tiddler-mode" value={{{ [<storyviewTitle>getstoryviewsingletiddlermode[]] }}}>
+<$set name="tv-storyview-single-tiddler-mode" value={{{ [<storyviewTitle>getstoryviewmode[singletiddlermode]] }}}>
```
pmario: I'm not sure, but I think the history-tab caption needs to be changed too?

Jermolene: Hi @pmario the open/history tab caption is handled here:

https://github.com/Jermolene/TiddlyWiki5/blob/single-tiddler-mode/core/ui/SideBar/Open.tid#L3

giffmex:
Feedback viewing on Windows laptop: great! 1) in other story views, the tiddlers slide to the top of the screen. I never liked before that sometimes tiddlers would open toward the bottom of the screen. 2) In Zoom only one tiddler open at a time, probably helps avoid the sluggishness of having a heap of tiddlers open at a time. And as probably everyone knows, zoomin is one of the first changes I make to a TW. 3) I do like the alternating between history and open when you change storyviews. However I had an idea: what about having only a history tab, where the first section is open tiddlers and below the open tiddlers is another list of 'recently opened' tiddlers? It would look similar to search results, which are similarly divided into two sections, title matches and all matches. That would avoid both problems: having two separate tabs, and having a tab change names, which could be confusing to new users. Regardless, I really like this new feature of single tiddler mode. Very nice!

giffmex:
Forgot to add: in my idea of having two sections in a history tab, in story view mode, the 'Open tiddlers' section at the top would only have one tiddler listed, and the 'Recently opened' tiddlers would be longer. In other modes, the 'Open tiddlers' section at the top would have more tiddlers, but usually not too many, and the 'Recently opened' tiddlers list would be shorter, only those tiddlers actively closed. And you could have a close all tiddlers at the top.

Jermolene:
Thanks @giffmex!

> Feedback viewing on Windows laptop: great! 1) in other story views, the tiddlers slide to the top of the screen. I never liked before that sometimes tiddlers would open toward the bottom of the screen. 2) In Zoom only one tiddler open at a time, probably helps avoid the sluggishness of having a heap of tiddlers open at a time. And as probably everyone knows, zoomin is one of the first changes I make to a TW. 3) I do like the alternating between history and open when you change storyviews.

Great!

> However I had an idea: what about having only a history tab, where the first section is open tiddlers and below the open tiddlers is another list of 'recently opened' tiddlers? It would look similar to search results, which are similarly divided into two sections, title matches and all matches. That would avoid both problems: having two separate tabs, and having a tab change names, which could be confusing to new users. Regardless, I really like this new feature of single tiddler mode. Very nice!

The first issue is that one list above the other makes it hard to find the second list if the first is long.

But the deeper problem is that the "open" list is still displayed when in single tiddler mode, and it's always going to be redundant: it'll never contain more than a single tiddler, and using the close button won't have the expected effect.

Meanwhile, I'd argue that the "history" list isn't useful when one is not in single tiddler mode because one can accomplish everything with the "open" tab and more (drag and drop, using the mini close buttons etc).

giffmex:
My second comment above was trying to say that the second list wouldn't be difficult to find if the first list is one tiddler in storyview mode. Also, while redundant, the one item open list might give a new user a comforting feeling that this tab works, because what he sees on the screen is what is in the open tiddlers list. 

In other modes, I doubt that in one session of use of a file that the user will have either too many open tiddlers or too many recently closed tiddlers. Some users probably leave everything open, in which case the history list is not as needed, like you say. And other users may close most tiddlers except those they are working on, so the open list will be short and the recently closed list will be accesible. I understand your reserve, but I still think it is worth experimenting with.

core/ui/PageTemplate.tid
pmario: ok

telmiger:
Great improvements, thanks for inviting me to this discussion! After a first sight some thougts from a user perspective.

- Is the term "History" clear enough? In shops you would see "Last visited" or "Recently viewed"
- Closing tiddlers: If you open the prerelease, click a link to "TiddlyWiki", navigate back to "TiddlyWiki Pre-release" in the History and close the tiddler, the story is empty. I would have expected to return to "TiddlyWiki" (as the last tiddler in the History tab – maybe not the same as the history stack?).
- When you have several tiddlers open in classic mode, switch to STM/zoomin view and back to classic again, you still have several tiddlers open. I would have expected only one, but switching view mode seems not to count as "navigation". I guess that's mentioned by @Jermolene in the first post here:

> The story river isn't yet adjusted when switching to a story view with STM enabled

I am really fond of the idea to unite History and Open in one tab (thanks @giffmex for bringing this up). Just thinking aloud: Why not call the tab "Story" and show a section "Open" and a section "Last opened" – maybe one or both in collapsible sections? The order in both sections can easily differ, thanks to open above/below as well as drag & drop. I often miss a "back" button and the history list would always be a nice alternative. 
Configuration could offer an option to only show the Open section if more than one tiddler is open (this is redundant information in all view modes, not only STM, at times when only one tiddler is open).

jdjdjdjdjdjd:
Hi Jeremy, 

Thanks for including me in the conversation as well. Although it seems even the mention got buried in my gmail inbox.

Some observations, related to how I personally use my wikis:
1. Saving the wiki does not save the History list inside the tab. Is this is an expected function? I would like the ability to save/remember it... 
2. Maybe a "clear history" button is needed in this tab, for when the list is too populated to be useful. 
3. "Draft of 'New Tiddler'" remains in the history tab, even after saving the New Tiddler. Clicking on that link will open a tiddler titled "Draft of 'New Tiddler'", with the missing tiddler notice.
4. After opening a lot of tiddlers, then closing them all, I am left with an empty story, and a populated History tab. Opening a single tiddler from this tab, somewhere along the middle of the list, and then closing it again, will not lead me back to an empty story, like I expect. I am instead cycled through the tiddlers that had been listed above that single tiddler before it jumped to the top of the list.
5. I echo @giffmex and @telmiger for a "History" partition, but only on non-zoomin story views. In zoomin view's History tab, maybe what's needed is to style the title of the currently open tiddler, if it being on the top of the list is not enough visual indication. At first I thought it would be nice if it was in bold, until I realized that shadow tiddlers are already in bold. Maybe italicized? But then, missing tiddlers are italicized...

telmiger:
> At first I thought it would be nice if it was in bold, until I realized that shadow tiddlers are already in bold. Maybe italicized? But then, missing tiddlers are italicized...

The most logical differentiation in my eyes would be making it **not a link** (no neeed for a link as it is the only one allready open). Something similar to the dates in the "Recent" tab.

Jermolene:
Thanks to everyone for the feedback. I think it's clear that single tiddler mode needs more work with the basic logic of how history is retained and navigated. 5.1.18 already has some big changes in the new web server features, so I'm inclined to defer STM until a subsequent release.

Jermolene:
Over in #4257, @akizuha makes a case for single tiddler mode to resemble the Simplenote UI.

AnthonyMuscio:
I have come late to this thread, but in relation to the history and how its represented in the side bar in zoomin view (single tiddler mode) **recent** could refer to the tiddlers recently navigated to as much as those recently edited. 

Perhaps the recent tab could be enhanced to allow selection of either modified (existing) or navigated (History) recent list. The Zoomin view could toggle this to the history for convenience when selected.

Such a feature would be helpful in other cases without the zoomin view as well.

Regards
Tony

kookma:
@Jermolene 
Refering to https://groups.google.com/d/msg/tiddlywiki/uwgQMe9qgGc/HsXI0l9mAwAJ

I would appreciate if you could kindly have a look and improve Tiddlywiki story view.

kookma:
@Jermolene 
Does the new merge f256b77 included the single tiddler mode?
Should I give a try?

Jermolene:
Hi @kookma this PR isn't merged yet. But the way that it uses the DOM during the animation is the same as the current code so I would not expect it to affect your problem.

kookma:
Hi @Jermolene 
Thank you for your clarification.

Jermolene:
Thinking about @kookma's use case has suggested an alternative approach:

* 5f25b21c0dc64a2648ab2b48de8681de7432708e - adds a new "solo" storyview that is a single tiddler story view without any animation
* 5272b13af11b5a1ff92a8bd198af792e8482bb72 - restores the "zoomin" storyview to its original behaviour (because making it a single tiddler mode broke the animations)

With these latest changes, the only effect is the new "solo" storyview, everything else should be as it was before.

Jermolene:
@kookma here is a build for you to experiment with:

[snapshot-5272b13af11b5a1ff92a8bd198af792e8482bb72.zip](https://github.com/Jermolene/TiddlyWiki5/files/4080399/snapshot-5272b13af11b5a1ff92a8bd198af792e8482bb72.zip)

kookma:
> @kookma here is a build for you to experiment with:
> 
> [snapshot-5272b13af11b5a1ff92a8bd198af792e8482bb72.zip](https://github.com/Jermolene/TiddlyWiki5/files/4080399/snapshot-5272b13af11b5a1ff92a8bd198af792e8482bb72.zip)

@Jermolene 
I am testing the solo and Tiddlyshow!
Those animation in slideshow mode are gone. But it seem I need to differently set the width and height of tiddler in solo mode! I will update you in brief in this part!

BUT more questions

- why History has replaced Open tab? Is there a plan to remove open tab for all story view? If not why  not to keep solo similar to other story view?

- I used $:/HistoryList and $:/StoryList for managing my navigation, but seems in solo mode these are different! I can correct Tiddlyshow script to work with new setting but I appreciate let me know if this is temporarily and only for solo story view or will be used for all story view

I come and update you with more results

kookma:
@Jermolene 
I cannot set tiddler to cover the whole screen. I am not sure if any new CSS classes is used for solo story view!

In the current zoomin story view, I set the below CSS to let a tiddler cover the whole screen! but
in solo story view this seems not work!


```
body.tc-body .tc-story-river {
  width:100%;
}

body.tc-body .tc-storyview-zoomin-tiddler{
top:0; left:0; margin:0;
width:100%;
}

body.tc-body .tc-tiddler-frame {
    border: 0;
    height:100vh;
}
```

kookma:
@Jermolene 

Also there seems the padding and margin setting for tiddler-frame, body, story are different!
More important it seems the selector are also different.

core/modules/widgets/navigator.js
```diff
@@ -60,6 +60,10 @@ NavigatorWidget.prototype.execute = function() {
 	// Get our parameters
 	this.storyTitle = this.getAttribute("story");
 	this.historyTitle = this.getAttribute("history");
+	this.openLinkFromInsideRiver = this.getAttribute("openLinkFromInsideRiver","top");
+	this.openLinkFromOutsideRiver = this.getAttribute("openLinkFromOutsideRiver","top");
+	this.singleTiddlerMode = this.getAttribute("singleTiddlerMode","no") === "yes";
+	this.relinkOnRename = this.getAttribute("relinkOnRename","no").toLowerCase().trim() === "yes";
 	this.setVariable("tv-story-list",this.storyTitle);
```
pmario: this is all super ugly!  IMO it shouldnt be that way

Jermolene: This is consistent with how all the core widgets manage their attributes.

pmario: Yes, but that's not the point I wanted to make. ... The navigator widget shouldn't be responsible to handle views. It's the story and storyviews in combination, which need to handle the stuff. 

So If an event wants to pass any parameter through, it need to be dynamic and not hardcoded like this one. 

We do have the `paramObject` which can hold any additional parameters the users and plugin authors may want to send to the navigator event handler. 

I want the stuff to be dynamic and there shouldn't be a core PR necessare if new parameters like this one are needed by a new function. 

It's OK if you make it. It gets merged, but if anyone else wants new params the chance is way to low. ... So I do want a solution, similar to the linkcatcher that I used for the new "story manager stuff" I'm creating. see: https://github.com/pmario/TiddlyWiki5/blob/dbf32fdef66196fb389ad55665b2f2c84991b65f/core/modules/widgets/linkcatcher.js#L75

linkcatcher has parameters that are OK, but can't be used for the new `tm-load-story` message. I needed different params. With the mechanism I propose it will be simple, to create new tm-xxx plugins without the need to modify linkcatcher or navigator. They just pass through the `paramObject`

also see: https://github.com/wikilabs/plugins/blob/8b85e7b8d66b0b4dd7c7f3fc213a22501e633f44/wikilabs/wltm-open-story/tiddlers/open-story.js#L42 where the paramObject defined above AND sent by a user can be used. 

Usage can be see at: https://github.com/wikilabs/editions/blob/eca970252a441c2612239453c01d406da7519d75/wltm-open-story/tiddlers/%24__core_ui_AdvancedSearch_Story_Tabs_template.tid#L8

and: https://github.com/wikilabs/editions/blob/eca970252a441c2612239453c01d406da7519d75/wltm-open-story/tiddlers/%24__core_ui_SideBar_Stories.tid#L20

I think, that's the mechanism we should use to add unknown future parameters to navigator and story related widgets.

Jermolene: Hi @pmario I think you're proposing a mechanism whereby widgets that dispatch events can specify the `paramObject` hashmap by using unknown attributes. Why wouldn't one use the `action-sendmessage` widget in such situations?

I don't quite understand how the proposal would be applied in this case? Which message is involved?

Jermolene:
Hi @kookma the zoomin storyview uses some intricate CSS to make the animations work. The new solo storyview uses the same CSS as the classic storyview; it's exactly like classic storyview, except there are no animations, and the storylist is restricted to only hold a single tiddler.

kookma:
> Hi @kookma the zoomin storyview uses some intricate CSS to make the animations work. The new solo storyview uses the same CSS as the classic storyview; it's exactly like classic storyview, except there are no animations, and the storylist is restricted to only hold a single tiddler.

@Jermolene 
Is there any reason History replaced Open tab?
Is there any plan to use History tab for all story view?

I am using solo and it seems it works much better than zoomin! However I uses $:/StoryList and $:/HistoryList for navigation, which does not work with solo.

I also appreciate to advise me how different are stylesheets (e.g selectors) from classic story view. I need to modify these stylesheet to get the desired behavior for Tiddlyshow plugin (like covering the whole screen, ...

Jermolene:
> Is there any reason History replaced Open tab?

There's a lot of discussion further up this thread: the "open" tab is useless when using a single tiddler mode story view, because it would only ever show a single tiddler. The History tab is offered as an alternative means of navigating.

> Is there any plan to use History tab for all story view?

We have considered it, but it seems confusing to have both tabs present at once.

> I am using solo and it seems it works much better than zoomin! However I uses $:/StoryList and $:/HistoryList for navigation, which does not work with solo.

I don't follow: all story views use $:/StoryList and $:/HistoryList.

> I also appreciate to advise me how different are stylesheets (e.g selectors) from classic story view. I need to modify these stylesheet to get the desired behavior for Tiddlyshow plugin (like covering the whole screen, ...

The "solo" storyview uses the same CSS as the "classic" storyview.

kookma:
@Jermolene 
Many thanks for all your clarification! So, all css as the classic story view and this makes life much easier.
I will update you how solo works with Tiddlyshow

kookma:
> > Is there any reason History replaced Open tab?
> 
> There's a lot of discussion further up this thread: the "open" tab is useless when using a single tiddler mode story view, because it would only ever show a single tiddler. The History tab is offered as an alternative means of navigating.
> 
> > Is there any plan to use History tab for all story view?
> 
> We have considered it, but it seems confusing to have both tabs present at once.
> 
> > I am using solo and it seems it works much better than zoomin! However I uses $:/StoryList and $:/HistoryList for navigation, which does not work with solo.
> 
> I don't follow: all story views use $:/StoryList and $:/HistoryList.
> I mean for navigation using shortcut keys I use $:/StoryList to find a specific tiddler to navigate to it! It seems in solo story view it is not work the same as $:/StoryList in zoomin! 
Following to your reply, I will look close to see what is wrong in my script.

One more question:
In solo story view if I want the opened tiddler occupies the whole screen what are the correct css settings? I confused with tc-tiddler-frame, tc-tiddler-view , tc-story-river

Thank you

Jermolene:
> In solo story view if I want the opened tiddler occupies the whole screen what are the correct css settings? I confused with tc-tiddler-frame, tc-tiddler-view , tc-story-river

I don't know, but I would expect to need to target tc-tiddler-frame along the lines suggested by @pmario above.

kookma:
> > In solo story view if I want the opened tiddler occupies the whole screen what are the correct css settings? I confused with tc-tiddler-frame, tc-tiddler-view , tc-story-river
> 
> I don't know, but I would expect to need to target tc-tiddler-frame along the lines suggested by @pmario above.

Thanks! by the way I cannot see any comment by @pmario  on css!

kookma:
Some findings: These are have been previously raised in the forum or Gihub

Solo story view

- click create new tiddler button
- save it as Test
- There is still missing "_Draft of New Tiddler 1_" in the History tab

kookma:
Hi Jeremy,

These are some comments by [Mat ](https://groups.google.com/d/msg/tiddlywiki/s-e2Ukz_OY0/dCgJgmuDAQAJ)


I downloaded the attached zip and extracted, getting two files: empty and index.

First trying out empty. I open it, go to ctrlpanel and select Solo storyview (for testing purposed it would probably make more sense to have this pre-set). Anyway, OK I close ctrlpanel so I only have GetingStarted and now click "+" to create a new tid. It opens in edit mode along with the already active GettingStarted. (Maybe I misunderstood the whole premise but I was expecting it to be only one tiddler open at a time? But OK, maybe it's not only one becaseu one is in view mode and one is in edit mode so maybe the latter "doesn't count"...)

After I've saved the tiddler I note that the sidebar tab is not named Open but History. And it lists: New Tiddler, Draft of 'New Tiddler', GettingStarted. I would think that second title should not be there. Clicking it confuses things even more.

If I create yet another "New Tiddler 1", I get the same issue with a missing Draft showing in the History list. 

If I reload the wiki and forget to set the Solo storyview but instead first click "+" to create a new tiddler but realize my mistake so I cancel the editview tiddler and then go to ctrlpanel and activate Solo, then upon closing the ctrlpanel the missing Draft pops up (hm!) and clicking "+" right away saves it as an existing tiddler but named "Draft of 'New Tiddler'". 

OK, deciding that the empty version is not quite ready for prime time I go to "index". 

This starts up with the usual default tiddlers open. I go to ctrpanel and note that Solo is pre-set. Closing the ctrlpanel... every tiddler closes/disappears. Actually what happened what that when I went to ctrlpanel, those default closed but I didn't see it because they were below the ctrlpanel. Not sure how to ideally handle that.

As with empty, the drafts seem to be listed in the History tab. I'd suggest that only existing drafts are actually listed - or is there a thought behind it? This time, clicking "+" does not seem to save it though but instead a "New Tiddler" is opened in draft mode and the History list is updated accordingly. (This makes me wonder if empty and index don't actually have the same version of the machanism...)

core/modules/widgets/navigator.js
```diff
@@ -122,12 +126,17 @@ NavigatorWidget.prototype.replaceFirstTitleInStory = function(storyList,oldTitle
 };
 
 NavigatorWidget.prototype.addToStory = function(title,fromTitle) {
-	if(this.storyTitle) {
-		this.wiki.addToStory(title,fromTitle,this.storyTitle,{
-			openLinkFromInsideRiver: this.getAttribute("openLinkFromInsideRiver","top"),
-			openLinkFromOutsideRiver: this.getAttribute("openLinkFromOutsideRiver","top")
-		});
-	}
+	this.wiki.addToStory(title,fromTitle,this.storyTitle,{
```
pmario: This PR will invalidate PR #4200 "Remove dependency on wiki.js for story-startup and navigator" where we don't have a decision yet. 

PR #4200 tries to fix a problem, that is also discussed in issue: #4180. Which links to many other issues, which are closed, but _not_ resolved. 

Proposal: make navigator.js + friends "multi - story" aware. #4354 also discusses several elements that this PR touches. ... I would like to discuss the main mechanism first.

pmario:
This PR will invalidate PR #4200 "Remove dependency on wiki.js for story-startup and navigator" where we don't have a decision yet. 

PR #4200 tries to fix a problem, that is also discussed in issue: #4180. Which links to many other issues, which are closed, but _not_ resolved. 

Proposal: make navigator.js + friends "multi - story" aware. #4354 also discusses several elements that this PR touches. ... I would like to discuss the main mechanism first.

Jermolene:
Hi @pmario I'm in no hurry to merge this PR, I think it needs quite a lot more work.

kookma:
@Jermolene 
I have tested the new story view and asked other people to do so!
The comments are given in above posts!

I would appreciate to have a look! I also wish to give some priority to this PR. It makes presentation simple and more semantic.

Thank you

Jermolene:
Hi @kookma I'm currently still finishing up the big refactoring of the client-server implementation (see #4373), and once that's merged I'll come back to this.

kookma:
Many thanks @Jermolene!

kookma:
> Hi @kookma I'm currently still finishing up the big refactoring of the client-server implementation (see #4373), and once that's merged I'll come back to this.

Hi @Jermolene 
I hope we could have the solo finalized soon!

This is a small reminder.

Thank you

Jermolene:
Just to say that I don't propose to merge this for v5.1.22 because we haven't resolved the issues around the handling of the "Open" sidebar tab.

kookma:
@Jermolene 
No problem, 5.1.22 introduces alot of new features! I will wait sometimes to see this in 5.1.23!

Thank you

kookma:
> Just to say that I don't propose to merge this for v5.1.22 because we haven't resolved the issues around the handling of the "Open" sidebar tab.

Hi @Jermolene - This is mild reminder.
By the way I am not sure with respect to https://github.com/Jermolene/TiddlyWiki5/issues/4473 you prefer to add solo story or not! Any way I think the solo can be used instead of zoomin and has a positive impact on the performance.

ght:
(I don't mean to nag, sorry if this is being worked on and just takes its time. I was just combing through old open issues and PRs and this seemed stale)

Quick summary:
- this was [postponed for 5.1.22 and planned for 5.1.23](https://github.com/Jermolene/TiddlyWiki5/pull/3412#issuecomment-607950231)
- then considered for **post-v5.1.23** in #5231
- *not* mentioned for **5.2.0** in #5717

Does this, and the above-mentioned #4473, mean this PR is obsolete?
The branch also seems to decay, having several merge conflicts already.

Jermolene:
> (I don't mean to nag, sorry if this is being worked on and just takes its time. I was just combing through old open issues and PRs and this seemed stale)

Good catch! This issue is one of those threads that once I started pulling, more and more inter-related stuff popped out. As sometimes happen, it will be a multi-year journey to get things right.

> Does this, and the above-mentioned #4473, mean this PR is obsolete?

Pretty much, yes. I've ended up exploring #4473 as part of my work on https://github.com/TWPUB/TWPUB-Tools (see https://twpub-tools.org). You can see there a sketch of the multi-column architecture. I'm currently working on bringing up a new implementation of the basic story river architecture that is entirely implemented in wikitext, without the complex encapsulated logic of the `<$navigator>` widget. This is essentially #3967.

pmario:
@Jermolene ... Can you remember, why this PR wasn't merged at the end?

kookma:
I was waiting for long time to see this merged! Solo ;-)

Jermolene:
Hi @pmario @kookma I think my earlier comments still apply:

* The way that the "open" tab switches into a "history" tab when the solo view is engaged is not entirely satisfactory
* I'm still working on revamping navigation into wikitext as part of https://github.com/TWPUB/TWPUB-Tools

But I guess it might make sense to revisit this for the next release.

Jermolene:
> But I guess it might make sense to revisit this for the next release.

By which I meant the release after v5.2.3

linonetwo:
When there is only a single tiddler, is it still a "story river"? Since we have layout mechanism, I think it is better to be a single tiddler layout.

And I would expect a layout that looks like Obsidian or VSCode, so it is easier to write article to earn users from those group.

Jermolene:
> When there is only a single tiddler, is it still a "story river"? Since we have layout mechanism, I think it is better to be a single tiddler layout.

The reason is backwards compatibility (as usual!). However it certainly wouldn't be unreasonable to approach this as a new page layout that ignores `$:/StoryList`.
 
> And I would expect a layout that looks like Obsidian or VSCode, so it is easier to write article to earn users from those group.

I agree 100%. I would also like to have an option to present the layout switcher as a strip of vertical tabs on the left that can be toggled via a button.

PR #3070, opened 2017-12-22
preventDefault drag behaviour inside textarea

BurningTreeC:
fixes errors when trying to drag inside textarea while an eventhandler is listening for drag events

PR #3414, opened 2018-08-28
revert preventing dragging within textarea

BurningTreeC:
this reverts #3070 which prevents dragging text from one codemirror instance to another

there are other ways to solve the issue I tried to solve with this, so we should revert this to the standard behavior

... deletes `t.preventDefault&&t.preventDefault(),` at character 138562


pmario:
@BurningTreeC ... Did you change the default codemirror.js file? .. If yes, that would be a general issue.

Because if I would have updated the CM library to a new version I would use the default files, without any mods ... So the next update done by me would fail. ...

BurningTreeC:
@pmario Look at #3070 and you see what i did. It was a PR before the new modularized Plugin. This PR reverts it

BurningTreeC:
> So the next update done by me would fail. ...

It wouldn't fail

Jermolene:
Thanks @BurningTreeC @pmario

BurningTreeC:
Thank you!

pmario:
So now the codemirror.js is the same again as downloaded from the codemirror.net homepage. right?

PR #3433, opened 2018-09-11
Add `text/asciidoc` MIME type

bimlas:
This type is not registered yet officially (https://github.com/asciidoctor/asciidoctor/issues/2502), but used by "Asciidoctor for TiddlyWiki" plugin (https://bimlas.gitlab.io/tw5-asciidoctor/).

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1998,6 +1998,7 @@ $tw.boot.startup = function(options) {
 	$tw.utils.registerFileType("audio/mp4","base64",[".mp4",".m4a"]);
 	$tw.utils.registerFileType("text/markdown","utf8",[".md",".markdown"],{deserializerType:"text/x-markdown"});
 	$tw.utils.registerFileType("text/x-markdown","utf8",[".md",".markdown"]);
+	$tw.utils.registerFileType("text/asciidoc","utf8",[".ad",".adoc",".asciidoc"]);
 	$tw.utils.registerFileType("application/enex+xml","utf8",".enex");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.wordprocessingml.document","base64",".docx");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet","base64",".xlsx");


pmario:
IMO you should not add / use a [standards-tree](https://tools.ietf.org/html/rfc6838#section-3.1) mime-type, that doesn't exist! 

The issue is about 1 year old and it seems nothing happened.

bimlas:
As I understand, should I add `x-` prefix, so register as `text/x-asciidoc`?

pmario:
The unregistered `x.` tree is discouraged. `x-` is not part of `x.`!! see: https://tools.ietf.org/html/rfc6838#section-3.4

So may be `text/x.asciidoc` ... I think this would be ok. ...

Jermolene:
Hi @pmario part of the problem here is that the way that we need to use tiddler `type` field values that are not "legal" IETF registered MIME types. For example, internally we use types such as `application/x-tiddler-html-div` or `text/vnd.tiddlywiki2-recipe`. In particular, the parser architecture requires us to choose one or more `type` field values to select that parser.

So, perhaps we should be clearer in saying that values of the `type` field in TiddlyWiki **MAY** be IETF registered MIME types, rather than **MUST**?

pmario:
> text/vnd.tiddlywiki

I think it should be best practice to use IETF conform mime types. ... So there is a chance to register them. 
If they are not conform all the content that uses them will have problems in the future. 

>application/x-tiddler-html-div

It's an experimental type. They are strongly discouraged, but not forbidden. ... There is a very low chance to get them registered. It happened in the past, with some "de-facto" standards.

pmario:
@bimlas -- I think this PR is still valid. I did have a look at the communication at: https://github.com/asciidoctor/asciidoctor/issues/2502 which stopped late 2020. 

So if you change it to `text/vnd.asciidoc` or `text/x.asciidoc` I would support the PR. IMO `text/asciidoc` as it stands is a no-go

Leilei332:
According to the issue, `text/asciidoc` is now registered as a temporary MIME type:

https://www.iana.org/assignments/provisional-standard-media-types/provisional-standard-media-types.xhtml

PR #3435, opened 2018-09-11
[Further TOC fixes] - FollowUp: fix tag macros

BurningTreeC:
this fixes the tag macros when titles are used that contain quotes, especially triple quotes and more

it also adds a class to the tag-icon so that it can get the same contrastcolor as the tag name assigned

### DIFF
--- a/core/wiki/macros/tag.tid
+++ b/core/wiki/macros/tag.tid
@@ -10,18 +10,18 @@ color:$(foregroundColor)$;
 \define tag-pill-inner(tag,icon,colour,fallbackTarget,colourA,colourB,element-tag,element-attributes,actions)
 <$vars foregroundColor=<<contrastcolour target:"""$colour$""" fallbackTarget:"""$fallbackTarget$""" colourA:"""$colourA$""" colourB:"""$colourB$""">> backgroundColor="""$colour$""">
 <$element-tag$ $element-attributes$ class="tc-tag-label tc-btn-invisible" style=<<tag-pill-styles>>>
-$actions$<$transclude tiddler="""$icon$"""/> <$view tiddler="""$tag$""" field="title" format="text" />
+$actions$<span class={{{ [<foregroundColor>removeprefix[$colourA$]addprefix[tc-tag-icon-dark]] [<foregroundColor>removeprefix[$colourB$]addprefix[tc-tag-icon-light]] }}}><$transclude tiddler="""$icon$"""/></span> <$view tiddler=<<__tag__>> field="title" format="text" />
 </$element-tag$>
 </$vars>
 \end
 
 \define tag-pill-body(tag,icon,colour,palette,element-tag,element-attributes,actions)
-<$macrocall $name="tag-pill-inner" tag="""$tag$""" icon="""$icon$""" colour="""$colour$""" fallbackTarget={{$palette$##tag-background}} colourA={{$palette$##foreground}} colourB={{$palette$##background}} element-tag="""$element-tag$""" element-attributes="""$element-attributes$""" actions="""$actions$"""/>
+<$macrocall $name="tag-pill-inner" tag=<<__tag__>> icon="""$icon$""" colour="""$colour$""" fallbackTarget={{$palette$##tag-background}} colourA={{$palette$##foreground}} colourB={{$palette$##background}} element-tag="""$element-tag$""" element-attributes="""$element-attributes$""" actions="""$actions$"""/>
 \end
 
 \define tag-pill(tag,element-tag:"span",element-attributes:"",actions:"")
 <span class="tc-tag-list-item">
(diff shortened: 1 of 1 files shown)


Jermolene:
Great, thanks @BurningTreeC

core/wiki/macros/tag.tid
```diff
@@ -10,18 +10,18 @@ color:$(foregroundColor)$;
 \define tag-pill-inner(tag,icon,colour,fallbackTarget,colourA,colourB,element-tag,element-attributes,actions)
 <$vars foregroundColor=<<contrastcolour target:"""$colour$""" fallbackTarget:"""$fallbackTarget$""" colourA:"""$colourA$""" colourB:"""$colourB$""">> backgroundColor="""$colour$""">
 <$element-tag$ $element-attributes$ class="tc-tag-label tc-btn-invisible" style=<<tag-pill-styles>>>
-$actions$<$transclude tiddler="""$icon$"""/> <$view tiddler="""$tag$""" field="title" format="text" />
+$actions$<span class={{{ [<foregroundColor>removeprefix[$colourA$]addprefix[tc-tag-icon-dark]] [<foregroundColor>removeprefix[$colourB$]addprefix[tc-tag-icon-light]] }}}><$transclude tiddler="""$icon$"""/></span> <$view tiddler=<<__tag__>> field="title" format="text" />
```
pmario: This change doesn't work as intended! There are 2 separate changes in the PR, which imo don't belong together. 

The colors are defined in line 11 and re-defined in line 13. IMO they should be only defined once and then used accordingly. 

The additional span causes styling problems and also creates an unnecessary element. IMO this styling needs to be reworked!

pmario: If you create a tiddler named: `a` and tag it with: `a`   add an icon eg: `$:/core/images/clone-button` this mechanism doesn't work. 

The assigned class to the span is `tc-tag-icon-dark` which is OK, but the assigned class isn't strong enough to go down to the `<svg>` element! 

We need to rework this PR. ... May be we make 2: 1 for the quotes and 1 for the styling.

BurningTreeC: Hi @pmario 

> This change doesn't work as intended! There are 2 separate changes in the PR, which imo don't belong together.

yes let's split them up

BurningTreeC: @pmario I tried your testcase and the following would solve that if we access the `svg` with an external stylesheet through `.tc-tag-label.tc-tag-label-[dark/light] svg {}`

```
\define tag-pill-classes()
tc-tag-label tc-btn-invisible $(textColorClass)$
\end

\define tag-pill-inner(tag,icon,colour,fallbackTarget,colourA,colourB,element-tag,element-attributes,actions)
<$vars foregroundColor=<<contrastcolour target:"""$colour$""" fallbackTarget:"""$fallbackTarget$""" colourA:"""$colourA$""" colourB:"""$colourB$""">> backgroundColor="""$colour$""">
<$vars textColorClass={{{ [<foregroundColor>removeprefix[$colourA$]addprefix[tc-tag-label-dark]] [<foregroundColor>removeprefix[$colourB$]addprefix[tc-tag-label-light]] }}}>
<$element-tag$ $element-attributes$ class=<<tag-pill-classes>> style=<<tag-pill-styles>>>
$actions$<$transclude tiddler="""$icon$"""/> <$view tiddler=<<__tag__>> field="title" format="text" />
</$element-tag$>
</$vars>
</$vars>
\end
```

pmario: >`.tc-tag-label.tc-tag-label-[dark/light] svg {}`

That's a very unspecific CSS rule, that will need to watch every svg-element on the page. :/ But imo that's the only way we can do this, since there are some hard-coded `style=` elements involved, which make styling by class impossible. ... 

We'll have to have a very close look and "inspect" the created DOM tree. May be we can eliminate the `style=` stuff.

...(rant) we probably missed that train in TW

pmario: The `<$vars` widget should not be nested like `<$set` ... it can handle several variables in 1 definition.

BurningTreeC: @pmario @Jermolene I added a PR #3446 for a css-only fix

pmario: I'm not happy with the pseudo-class. it needs to watch every svg. and now also check all assigned classes instead of stop early.

BurningTreeC: @pmario alternativelly, we could add a `style=<<...>>` to the icons directly?

BurningTreeC: > @pmario alternativelly, we could add a style=<<...>> to the icons directly?

in their svg tiddlers

BurningTreeC: `<svg style=<<tag-pill-styles>> ... </svg>`

BurningTreeC: @pmario 

what also works is just deleting the line from `.tc-tag-label svg`

`fill: <<colour tag-foreground>>;`

from the vanilla base stylesheet

PR #3437, opened 2018-09-12
Fix tag macros

BurningTreeC:
This fixes the tag macros

The color for the icon will get a separate issue


Jermolene:
Thanks @BurningTreeC, apologies for the delay

BurningTreeC:
Thank you, @Jermolene 

there's another PR (#3442) which fixes the list-tagged-draggable macro, too. Now that bad tags don't break the interface, they still break the tag-dropdown's draggable list - that PR should fix that

BurningTreeC:
Last but not least the Tag-Manager uses tag-names for its states for the icon-dropdown and the info-dropdown. certain tags break those states, like my very-bad-tag example with `!!` in its name

BurningTreeC:
you proposed the `stateTitle` ... attributes
another idea would be an `escapestate="true"` which escapes the state's breaking characters

pmario:
I think, we should be sure about the performance implications first. Since those changes are made to the TOC macros we have a problem there!

PR #3439, opened 2018-09-12
popups/states: add "noStateReference" attribute

BurningTreeC:
this adds an optional attribute to the reveal and button widgets:

`noStateReference`

if set to `"true"` it causes the button widgets to set its states to the full title, without interpreting the string as a text reference. The reveal widget - if the attribute is true - also skips the text-reference.

The popup handler in utils/dom/popup.js gets and additional `noStateReference: true` passed with the options from the button widget and sets the tiddler text directly, without textreference.

This could be used to fix the TagManager which doesn't do it right with tag-names that contain `!!` or `##`

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -167,7 +167,7 @@ ButtonWidget.prototype.triggerPopup = function(event) {
 };
 
 ButtonWidget.prototype.setTiddler = function() {
-	if(this.getVariable("noStateReference") !== "true") {
+	if(this.noStateReference !== "true") {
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @BurningTreeC. I think actually that we should take a slightly different approach: I've come to the conclusion that it was a mistake for text references (meaning the special !! and ## syntaxes) to exist beyond the parser layer. Widgets should have separate attributes for the title, field, and index. So, we would add new `stateTitle`, `stateField` and `stateIndex` attributes to these widgets, and have them take priority over any `state` attribute.

BurningTreeC:
@Jermolene yes let's do that, sounds more elegant

PR #3441, opened 2018-09-12
fix list-tagged-draggable for special characters

BurningTreeC:
apparently there wasn't more to do than this


BurningTreeC:
nope

PR #3446, opened 2018-09-13
fix tag svg-icon colors css only

BurningTreeC:


BurningTreeC:
alternativelly, we could add `style=<<tag-pill-styles>>` to all svg icons

BurningTreeC:
or better, we just delete the line from `.tc-tag-label svg`

`fill: <<colour tag-foreground>>;`

from the vanilla base stylesheet

BurningTreeC:
closing this PR in favor of #3448

PR #3447, opened 2018-09-13
show tag-icons in tags edittemplate

BurningTreeC:
this PR makes tag icons show up within the edit template


Jermolene:
Thanks @BurningTreeC

PR #3448, opened 2018-09-14
make tag-icons use the tag-pill-styles

BurningTreeC:
this line inherits the `fill: $(foregroundColor)$` from `<<tag-pill-styles>>` in `$:/core/macros/tags`
deleting the line makes the tag-icons fill with the computed color


Jermolene:
Thanks @BurningTreeC

PR #3453, opened 2018-09-17
Update DateFormat.tid

twMat:
very minor tweak


sukima:
What does this do? How does the `0` differ from the not-`0`?

twMat:
E.g times:  19:2:3 vs 19:02:03

twMat:
Good idea. I added so there are now two examples to clarify the diff.

Jermolene:
Thanks @twMat sorry I missed this, I fixed it in a75434a347eb2fb31ff7a5fce4d3cba301c5eeb0 (and added the YYYY for the year).

PR #3458, opened 2018-09-30
Suggest https instead of http

iblech:


editions/tw5.com/tiddlers/howtos/Adding a Twitter Follow button.tid
```diff
@@ -17,5 +17,5 @@ The only change from the version published at https://dev.twitter.com/docs/follo
 Note that the src URL is given without a protocol (ie "http" or "https"). It's done this way so that it works on sites whether they are hosted on a HTTP or HTTPS domain. If you want the Twitter button to work while using a TiddlyWiki offline on a "file://" URL, then you'll need to manually add the protocol. For example:
```
sukima: According to this line the correct example should be:

```
src="//platform.twitter.com/widgets/follow_button.html?screen_name=jermolene"
```

iblech: Sorry, I'm afraid I don't quite understand. If your Tiddly Wiki is served using either http or https, then the link `//platform.twitter.com/widgets/...` will do nicely (as described on the page, and not touched on by my pull request). If on the other hand it is served using file, then the link `//platform.twitter.com/widgets/...` won't work (as described on the page). The page then went on to recommend to fall back on `http://platform.twitter.com/widgets/...`; I'm proposing to change the suggestion to use https instead of http.

sukima: >Note that the src URL is given without a protocol (ie "http" or "https"). It's done this way so that it works on sites whether they are hosted on a HTTP or HTTPS domain. If you want the Twitter button to work while using a TiddlyWiki offline on a "file://" URL, then you'll need to manually add the protocol. For example:

This line read very different for me. Instead when I read it I understood it as suggesting the user should use `//` not `http://` or `https://` and yet in the previous example it showed `http://` and in your PR it is `https://` both of which seem to contradict that statement one line above it.

Can that **Note** be reworded to make it clear what the _example_ is really trying to illustrate?

Jermolene:
Thanks @iblech

PR #3466, opened 2018-10-05
Create plugin-info.js

inmysocks:
This is similar to edition-info.js but instead of returning the info for all available editions it returns the info for all available plugins.

Bob uses this to let you change which plugins are listed in tiddlywiki.info from inside the wiki, with the new improvements planned with the server-browser and browser-server communication in the core this could be added to the core as well.

### DIFF
--- a/core/modules/utils/plugin-info.js
+++ b/core/modules/utils/plugin-info.js
@@ -1,0 +1,51 @@
+/*\
+title: $:/plugins/OokTech/Bob/plugin-info.js
+type: application/javascript
+module-type: utils-node
+
+Information about the available plugins
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @inmysocks it may be a bit early to merge this: I avoid having functions in the core that are not used elsewhere in the core. Otherwise, orphan functions build up and it's hard to know where things are being used.

> Bob uses this to let you change which plugins are listed in tiddlywiki.info from inside the wiki, with the new improvements planned with the server-browser and browser-server communication in the core this could be added to the core as well

I'm also interested in this. I've not dug into Bob in detail, but the approach I'm interested in is to change the listen/server commands so that they can serve external wiki folders - ie a different wiki folder than is running the main wiki, thus avoiding the problems of the main wiki altering its own parameters.

PR #3467, opened 2018-10-06
correct scrolling with tc-adjust-top-of-scroll

BurningTreeC:
not multiplying `offset` with t causes jumps at the first animation steps, where the offset value is bigger than `endY - scrollPosition.y`


BurningTreeC:
@Jermolene - I think it's still not correct, my "fix" just eliminates the first jumps, but leads to slow scrolling at the beginning and fast scrolling at the end. the " - offset" must go to the `getBounds()` function

BurningTreeC:
@Jermolene , I just want to bump this. There's another PR here:  #3475 

That PR should be correct but it's better to give it a thought before merging

PR #3472, opened 2018-10-08
[proposal] add editorManager to edit widgets / engines

BurningTreeC:
this adds `$tw.editorManager` which in this PR is used in the CodeMirror engine to:

- fix the focus-loss when clicking toolbar buttons
- restore cursor position / selections when opening/closing the preview
- scrolling those cursor positions / selections into view

this could be used for all kinds of improvements, like restoring focus on the tags input when clicking "add" or the fieldname input when clicking "add" ...

also the framed engine could profit, restoring selections, cursor position, setting focus on the previously focused input field when re-opening a tiddler etc

### DIFF
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -162,6 +162,9 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	Compute the internal state of the widget
 	*/
 	EditTextWidget.prototype.execute = function() {
+		this.editorStateQualifier = this.getVariable("storyTiddler");
+		$tw.editorManager = $tw.editorManager || [],
+		$tw.editorManager[this.editorStateQualifier] = $tw.editorManager[this.editorStateQualifier] || {};
 		// Get our parameters
 		this.editTitle = this.getAttribute("tiddler",this.getVariable("currentTiddler"));
 		this.editField = this.getAttribute("field","text");
--- a/core/modules/widgets/edit.js
+++ b/core/modules/widgets/edit.js
@@ -40,6 +40,9 @@ var EDITOR_MAPPING_PREFIX = "$:/config/EditorTypeMappings/";
 Compute the internal state of the widget
 */
 EditWidget.prototype.execute = function() {
+	this.editorStateQualifier = this.getVariable("storyTiddler");
+	$tw.editorManager = $tw.editorManager || [];
+	$tw.editorManager[this.editorStateQualifier] = $tw.editorManager[this.editorStateQualifier] || {};
 	// Get our parameters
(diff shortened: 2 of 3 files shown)


core/modules/editor/factory.js
```diff
@@ -162,6 +162,9 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	Compute the internal state of the widget
 	*/
 	EditTextWidget.prototype.execute = function() {
+		this.editorStateQualifier = this.getVariable("storyTiddler");
+		$tw.editorManager = $tw.editorManager || [],
+		$tw.editorManager[this.editorStateQualifier] = $tw.editorManager[this.editorStateQualifier] || {};
```
pmario: The above 2 lines don't make much sense to me. You copy the `$tw.editorManager[this.editorStateQualifier]` variable to itself, or init it with an empty object. 

`$tw` is a global structure. So initialisation should be done at startup or registering elements. ... But imo not as a "side-effect" in execute.

core/modules/widgets/edit.js
```diff
@@ -40,6 +40,9 @@ var EDITOR_MAPPING_PREFIX = "$:/config/EditorTypeMappings/";
 Compute the internal state of the widget
 */
 EditWidget.prototype.execute = function() {
+	this.editorStateQualifier = this.getVariable("storyTiddler");
+	$tw.editorManager = $tw.editorManager || [];
+	$tw.editorManager[this.editorStateQualifier] = $tw.editorManager[this.editorStateQualifier] || {};
```
pmario: same as above. IMO init-problem

plugins/tiddlywiki/codemirror/engine.js
```diff
@@ -91,6 +91,7 @@ function CodeMirrorEngine(options) {
 	this.value = options.value;
 	this.parentNode = options.parentNode;
 	this.nextSibling = options.nextSibling;
+	this.editorManager = $tw.editorManager[self.widget.editorStateQualifier];
```
pmario: I didn't review this file.

core/modules/editor/factory.js
BurningTreeC: aha, I was initializing it differently originally, like "if it's undefined, then it's `[]` "

I think I could move it to a startup module - or - like the keyboardManager - a "global" module-type. I haven't looked at what that global type is, but I imagine the editorManager to work similarily

plugins/tiddlywiki/codemirror/engine.js
BurningTreeC: here the manager gets actually used - init problems beside

BurningTreeC:
I'll make a new proposal for this one

PR #3473, opened 2018-10-08
[pagescroller] add callback function option

BurningTreeC:
this allows using the pagescroller for scrolling elements into view where the rect gets calculated somewhere else

Example: CodeMirror has the `cm.cursorCoords()` function that returns the rect of the textselection (or cursor coordinates)

this scrolls the cursor or text selection into view using tiddlywikis pagescroller:

`$tw.pageScroller.scrollIntoView(undefined, function() { return self.cm.cursorCoords(true,"window"); });`


Jermolene:
Thanks @BurningTreeC

PR #3475, opened 2018-10-10
correct tc-adjust-top-of-scroll

BurningTreeC:
the first fix made the scrolling jumps disappear, but the calculations were wrong

it must be done in `getBounds()`


Jermolene:
Thanks @BurningTreeC

PR #3476, opened 2018-10-10
Fix for lazy loaded state in Reveal widget

Lioric:
In Reveal Widget execute function, the state is read "this.readState()", this method calls "wiki.getTextReference()" with the state's tiddler title as the textreference (so the actual tiddler text is requested), as no field or index is used, getTiddlerText is called to lazy load the tiddler (as it is the first time the tiddler contents are requested)

getTiddlerText dispatch the "lazyLoad" event with the title as parameter and right after that it returns "null". This return result is bubbled up to Reveal widget's readState where state is set to this "null" value and right after that, the switch paths use this "state" value (now null) in "compareStateText", which tries to execute a localCompare on this null "state" object (state.localeCompare(this.text,....)), and this results in a error exception

This fix uses the widget default value until the real state is (lazy) loaded

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -116,6 +116,11 @@ Read the state tiddler
 RevealWidget.prototype.readState = function() {
 	// Read the information from the state tiddler
 	var state = this.stateTitle ? this.wiki.getTextReference(this.stateTitle,this["default"],this.getVariable("currentTiddler")) : this["default"];
+	
+	// If state tiddler is not yet loaded (lazily loaded), use default value until it gets udpated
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @Lioric thanks, that's great (very good sleuthing to track it down, too).

Apologies for the bureaucracy, but would you mind please signing the CLA (in a separate PR) as described in contributing.md in the root of this repo? Many thanks!

Lioric:
Hi @Jermolene, ok, I have signed the CLA in a separated PR

Jermolene:
Thanks @Lioric!

Issue #3484, opened 2018-10-15
Search dropdown doesn't display when create wikitext link was recently used


hoelzro:
Steps to reproduce:

  * Load https://tiddlywiki.com (I loaded a version I built from d50e2df57 - the latest revision at the time of writing)
  * Click new tiddler button
  * Hit Ctrl-L in the editor to prompt for a wikitext link - type something (I did `tag`)
  * Click on a link to insert a link to that tiddler into the tiddler being edited
  * Click on the search editor in the sidebar, typing something (I did `full`)
  * Note that the search results popup doesn't display

I've managed to reproduce this on Firefox (62.0.3) and Chromium (69.0.3497.100-1)

Here's a GIF demoing the bug:

![peek 2018-10-15 14-24](https://user-images.githubusercontent.com/107804/46973233-3b118880-d086-11e8-8d33-eed207a282b9.gif)

I think I've seen something similar with the tag dropdown, but I haven't taken the time to reduce that to a series of reproducible steps.

hoelzro:
Looking at the code, I think what's happening is that the `tc-popup-keep` is preventing the popup mechanism from cleaning up the popup state when the link is clicked; I'm not sure what the best way to fix this is.  My feeling, having only superficial experience with the codebase, would be to monitor for the destruction of the popup state tiddler (which is deleted by the link dropdown actions) and cancel the corresponding popup, but maybe Jeremy has a better idea.  An alternative would be that when creating a new popup, any existing popups with an ancestor hierarchy independent of the new popup would be canceled.

twMat:
I suspect the wikitext link button doesn't delete the dropdown state tiddler properly.
If that is the case, then this is probably a sub-question to #2959

BurningTreeC:
Hi @hoelzro , @twMat ... the bug here is that the popup gets cancelled clicking on the page, but not when clicking within input fields. I solved that "my way" in the keeboord plugin at https://burningtreec.github.io/KeeBoord (in development) / http://keeboord.tiddlyspot.com (stable)

hoelzro:
@twMat You're right - it doesn't delete the dropdown state, because `tc-popup-keep` prevents the popup logic from doing so.  Thanks for linking to that issue - it looks like this issue is related, if not a duplicate!

[...]

PR #3490, opened 2018-10-19
Cancel non-ancestor popups when showing a new popup

hoelzro:
Addresses GH #3484

As far as I can tell, the popup level checks in this module are meant to handle nested popups.  It seems to me that the goal is for at most a single hierarchy of popups to exist at any given time - bearing that in mind, this change checks any popups currently tracked by the module, canceling any that don't share an element hierarchy with the new popup.


hoelzro:
My cursory tests with this change work; I'm going to add this code to my own wiki and see if any bugs fall out over the next few days.

Jermolene:
Thanks @hoelzro

PR #3491, opened 2018-10-19
Problem: revealed dropdown menu on mobiles

### LINKED ISSUE #3486: Revealed dropdown menu on mobiles
On mobile phones, tiddler's dropdown menu stays partially off-screen. Any plans to mitigate this issue? I might be able to look into it a bit later.

screenshot_20181015-140200

yrashk:
On mobile phones, tiddler's dropdown menu stays partially off-screen.

Solution: ensure that the revealed coordinates are never negative

Closes #3486

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -58,23 +58,23 @@ RevealWidget.prototype.positionPopup = function(domNode) {
 	domNode.style.zIndex = "1000";
 	switch(this.position) {
 		case "left":
-			domNode.style.left = (this.popup.left - domNode.offsetWidth) + "px";
+			domNode.style.left = Math.max(0, this.popup.left - domNode.offsetWidth) + "px";
 			domNode.style.top = this.popup.top + "px";
 			break;
 		case "above":
 			domNode.style.left = this.popup.left + "px";
-			domNode.style.top = (this.popup.top - domNode.offsetHeight) + "px";
+			domNode.style.top = Math.max(0, this.popup.top - domNode.offsetHeight) + "px";
 			break;
 		case "aboveright":
 			domNode.style.left = (this.popup.left + this.popup.width) + "px";
-			domNode.style.top = (this.popup.top + this.popup.height - domNode.offsetHeight) + "px";
+			domNode.style.top = Math.max(0, this.popup.top + this.popup.height - domNode.offsetHeight) + "px";
 			break;
 		case "right":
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @yrashk thanks for looking at this. I think this doesn't address popups that are offscreen to the right, presumably because users can scroll to see those popups? I do wonder if we might not introduce some code to slide popups leftwards or upwards if they overlap the right or bottom of the viewport, what do you think? I think @BurningTreeC may have some views here too.

BurningTreeC:
@Jermolene - I thought maybe `max-width: 100%` would already do it, but now thinking about what you say, when resizing a window while a popup is open, or just opening/closing the sidebar (with a keyboard shortcut), its popup-button moves but the popup stays on its position. The reveal widget could listen for such a resize or change and call positionPopup again?

BurningTreeC:
I tried this in the reveal widget which handles re-positioning popups when the window resizes:

```

if(this.popup) {
		var options = {
			domNode: this.domNodes[0],
			title: this.popup,
			wiki: this.wiki
		};
		this.domNodes[0].ownerDocument.defaultView.addEventListener("resize", function() {
			if(isPoppedUp) {
				$tw.popup.show(options);
			}
		});
}

```

I don't know how to track if the popup-button position changes due to something else, like opening/closing the sidebar

BurningTreeC:
... I guess an event listener in popup.js would be better :grin:

BurningTreeC:
something like this in popup.js :+1: 

```

Popup.prototype.handleResizeEvent = function() {
	var numPopups = this.popups.length;
	for (var i=0; i<numPopups.length; i++) {
		this.show(this.popups[i]);
	}
};

```

and the resize listener that calls this in rootwidget.js

yrashk:
@Jermolene you are right, this patch does not address all potential cases. However, it is meant as more of a stopgap measure to fix the problem that happens when rendering on mobiles in pretty  much 100% cases.

BurningTreeC:
> it is meant as more of a stopgap measure to fix the problem that happens when rendering on mobiles in pretty much 100% cases.

for that it's good enough and I think further fixes should get new PR's

Jermolene:
Great, thanks @yrashk I'll go ahead and merge.

@BurningTreeC -- I didn't mean that we reposition popups when the browser viewport changes, just that we might avoid initially positioning them so that they're clipped by the right edge of the viewport.

Issue #3494, opened 2018-10-21
$tw.platform.isLinux wrong on latest firefox


BurningTreeC:
in startup.js the platforms are tested like so:

```
        $tw.platform.isMac = /Mac/.test(navigator.platform);
        $tw.platform.isWindows = /win/i.test(navigator.platform);
        $tw.platform.isLinux = /Linux/i.test(navigator.appVersion);
```

`navigator.appVersion` contains the "Linux" string on Chrome on my Linux Desktop, but not on latest Firefox

I'd appreciate if someone could verify this on another machine, because I have a nightly build and somehow my stable builds got replaced by it... 

anyway, if I use navigator.platform for the Linux test, the platform gets reported correctly. Also on Chrome navigator.platform contains "Linux" so it would be fine there, too

Jermolene:
Hi @BurningTreeC ouch! It sounds like a good idea to go with the navigator.platform.

Jermolene:
Hi @BurningTreeC great, perhaps @pmario can confirm your findings.

pmario:
I can confirm, that there is a problem.  ubuntu 18.04  ff 62.0.3

BurningTreeC:
@pmario @Jermolene is it safe to simply use `navigator.platform` or will there problems with older firefoxes?

pmario:
seems to be ok: see: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorID/platform

Jermolene:
OK, please go ahead with a PR. Many thanks @BurningTreeC

PR #3495, opened 2018-10-22
fix for #3494 (Linux - wrong platform detection)

BurningTreeC:
this fixes the wrong platform-detection on firefox linux


BurningTreeC:
Thanks @pmario

Jermolene:
Thanks @BurningTreeC

PR #3501, opened 2018-10-26
fix file-save mechanism with external-js setting

### LINKED ISSUE #3500: external-js: auto-save doesn't work. PR on the way.
2 problems: 

 - If a file-TW is used with the tiddly-fox like auto-save mechanism an invalid html file is created. 
 - this [commit](https://github.com/Jermolene/TiddlyWiki5/commit/f2a38960fce1bb70f1090012334bee452d59f45f) breaks the default download savers. all of them. It only seems to work with `tw --server`

How to reproduce. 

 - create an empty.html file with tiddlywiki.js
 - install file-backups or savefiles browser addOn
 - open it from the Downloads folder
 - edit a tiddler with autosave enabled. 

PR is on the way!!

pmario:
- this fixes #3500 
 - it reverts the problem "Cannot load tiddlywiki.js" for file-TWs
 - adds `external-js` build parameter to empty

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -245,7 +245,7 @@ NavigatorWidget.prototype.handleDeleteTiddlerEvent = function(event) {
 	this.removeTitleFromStory(storyList,title);
 	this.saveStoryList(storyList);
 	// Trigger an autosave
-	$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki"});
+	$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki", param: this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template")});
 	return false;
 };
 
@@ -343,7 +343,7 @@ console.log("Relinking '" + draftOf + "' to '" + draftTitle + "'");
 					}
 				}
 				// Trigger an autosave
-				$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki"});
+				$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki", param: this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template")});
 			}
 		}
 	}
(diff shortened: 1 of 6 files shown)


core/modules/widgets/navigator.js
```diff
@@ -564,7 +564,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	// Navigate to the $:/Import tiddler
 	this.addToHistory([event.param]);
 	// Trigger an autosave
-	$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki"});
+	$tw.rootWidget.dispatchEvent({type: "tm-auto-save-wiki", param: this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template")});
```
Jermolene: It seems awkward to have to add this parameter each time we invoke auto-save-wiki. Can we avoid it?

core/templates/external-js/tiddlywiki5-external-js.html.tid
```diff
@@ -37,5 +37,5 @@ title: $:/core/templates/tiddlywiki5-external-js.html
 <!--~~ Ordinary tiddlers ~~-->
 {{$:/core/templates/store.area.template.html}}
 </body>
-<script src="%24%3A%2Fcore%2Ftemplates%2Ftiddlywiki5.js" onerror="alert('Error: Cannot load tiddlywiki.js');"></script>
+<script src="tiddlywiki.js" onerror="alert('Error: Cannot load tiddlywiki.js');"></script>
```
Jermolene: This change breaks the web server usage. Perhaps we need to make the path to the script be a retained in a configuration tiddler? Alternately, we need to parameterise this template so that we can pass the script path as a variable, and then create two new wrapper templates that set the variable and invoke this template.

core/modules/widgets/navigator.js
pmario: You are right. It feels a bit hacky!

pmario: The problem I see is, that the tiddler which causes the update is: `$:/config/SaveWikiButton/Template`. So we need to work with the active wiki-store.

pmario: There are only 2 files, where `.dispatchEvent({type: "tm-auto-save-wiki", ...` is called:

 - navigator.js
 - plugin-maker.js

pmario:
will be superseded by: Fix problems with single-file wikis when using the external-js template #5570

PR #3502, opened 2018-10-26
Enhance search operator

Jermolene:
It is proposed to enhance the search operator:

* to permit an arbitrary list of fields to be specified
* to add support for flags to control the type of search performed: the current token-based search, a literal search, a whitespace-tolerant search, and a regexp search

For example:

```
[search:caption,description[this]]
[search:caption,description:literal[this phrase]]
[search:caption,description:literal,casesensitive[this phrase]]
[search::literal,casesensitive[this phrase]]
[search::regexp,casesensitive[(bl|d)o(ggs|e)]]
```

The extended syntax for operator suffixes has been implemented in a manner that allows other filter operators to opt in to using it if they want to. We should review carefully whether the proposed format will be adequate for other operators that we may want to extend.

These changes are backwards compatible as long as existing wikis are not using the search operator with fields whose names contain colons or commas.

We could extend the advanced search box to allow the new search options to be used but I'm concerned that it might be overkill for most users while adding a lot to the core, and so might be best in a plugin.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -1047,8 +1047,12 @@ Options available:
 	exclude: An array of tiddler titles to exclude from the search
 	invert: If true returns tiddlers that do not contain the specified string
 	caseSensitive: If true forces a case sensitive search
-	literal: If true, searches for literal string, rather than separate search terms
-	field: If specified, restricts the search to the specified field
+	field: If specified, restricts the search to the specified field, or an array of field names
+	The search mode is determined by the first of these boolean flags to be true
+		literal: searches for literal string
+		whitespace: same as literal except runs of whitespace are treated as a single space
+		regexp: treats the search term as a regular expression
+		words: (default) treats search string as a list of tokens, and matches if all tokens are found, regardless of adjacency or ordering
 */
 exports.search = function(text,options) {
 	options = options || {};
@@ -1064,6 +1068,22 @@ exports.search = function(text,options) {
 		} else {
 			searchTermsRegExps = [new RegExp("(" + $tw.utils.escapeRegExp(text) + ")",flags)];
 		}
+	} else if(options.whitespace) {
+		terms = [];
+		$tw.utils.each(text.split(/\s+/g),function(term) {
+			if(term) {
+				terms.push($tw.utils.escapeRegExp(term));
+			}
+		});
+		searchTermsRegExps = [new RegExp("(" + terms.join("\\s+") + ")",flags)];
+	} else if(options.regexp) {
+		try {
+			searchTermsRegExps = [new RegExp("(" + text + ")",flags)];			
+		} catch(e) {
+			searchTermsRegExps = null;
+console.log("Regexp error",e)
+		}
+console.log("Regexps are ",searchTermsRegExps)
 	} else {
 		terms = text.split(/ +/);
 		if(terms.length === 1 && terms[0] === "") {
@@ -1075,6 +1095,21 @@ exports.search = function(text,options) {
 			}
 		}
 	}
+	var fields = [];
+	if(options.field) {
+		if($tw.utils.isArray(options.field)) {
+			$tw.utils.each(options.field,function(fieldName) {
+				fields.push(fieldName);
+			});
+		} else {
+			fields.push(options.field);
+		}
+	}
+	if(fields.length === 0) {
+		fields.push("title");
+		fields.push("tags");
+		fields.push("text");
+	}
 	// Function to check a given tiddler for the search term
 	var searchTiddler = function(title) {
 		if(!searchTermsRegExps) {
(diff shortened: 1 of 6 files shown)


xcazin:
Hi @Jermolene,

Well, this proposal doesn't seem too controversial :-) I find it quite elegant myself, but I wish it could also produce several outputs when the search leads to several matches.

For instance, I'd like to obtain the following list `FMCBlWgcrGc 390` out of the following string input: 
`https://www.youtube.com/watch?v=FMCBlWgcrGc&start=390`. I could imagine that  the `removeprefix` and `removesuffix` operators get extended to support a regexp flag and combine them into appropriate runs. Even nicer if I can get the string directly in the `url` field of the input tiddlers.

Still, in order to build a more ubiquitous YouTube dropper, I'd also need to take in account the multiple variants of a YouTube video url, depending on whether the host name is _youtube.com_ or _youtu.be_, whether the video is part of a playlist like in https://www.youtube.com/watch?v=FMCBlWgcrGc&index=3&list=PLJXb5DyfMUF7UwlTeFqIuiCLz4wHanAdt&t=390, with or without a starting time, etc. 

Such aim leads to monster regexps like `https:\/\/(youtu\.be\/|www\.youtube\.com\/(watch\?(.*&)?v=|(embed|v)\/))([^\?&"'>]+)&t=([0-9]+)(.*)` for which the prefix/suffix game is no fun anymore. On the other hand, if the `search` operator could retrieve every groups in one go, parsing with filters would become magic again. The issue then is that some groups are bound to be empty. How could I ask such a `search` operator for matches 5, 6 and 7 (see http://rubular.com/r/yDtKjZ56eR)?

Jermolene:
Hi @xcazin 

>  I'd like to obtain the following list `FMCBlWgcrGc 390` out of the following string input: 
`https://www.youtube.com/watch?v=FMCBlWgcrGc&start=390`

I think that this sort of thing needs a full-blown widget that sets variables to the results of a match:

```
<$match text={{$:/IncomingURL}} regexp="""https:\/\/(youtu\.be\/|www\.youtube\.com\/(watch\?(.*&)?v=|(embed|v)\/))([^\?&"'>]+)&t=([0-9]+)(.*)""" casesensitive="no" global="yes" multiline="yes">

Matching text: <$text text=<<match>>/>

Matching index: <$text text=<<index>>/>

Last index: <$text text=<<last-index>>/>

Capture groups:  <$text text=<<match-1>>/> <$text text=<<match-2>>/> etc.

<$match

</$match>

```

We might also support attributes for specifying different variable names for the various outputs (`output-match="my-match-var"`).

editions/tw5.com/tiddlers/filters/examples/search.tid
```diff
@@ -1,10 +1,16 @@
...
-
+<$macrocall $name=".operator-example" n="4" eg="[search:*[arch]]" ie="tiddlers containing `arch` in any field"/>
+<$macrocall $name=".operator-example" n="5" eg="[search:-title,caption[arch]]" ie="tiddlers containing `arch` in any field except <<.field title>> and <<.field caption>>"/>
+<$macrocall $name=".operator-example" n="6" eg="[!is[system]search[the first]]" ie="non-system tiddlers containing a case-insensitive match for both the <<.word 'the'>> and <<.word 'first'>>"/>
+<$macrocall $name=".operator-example" n="7" eg="[!is[system]search::literal[the first]]" ie="non-system tiddlers containing a case-insensitive match for the literal phrase <<.word 'the first'>>"/>
+<$macrocall $name=".operator-example" n="8" eg="[!is[system]search::literal,casesensitive[The first]]" ie="non-system tiddlers containing a case-sensitive match for the literal phrase <<.word 'The first'>>"/>
+<$macrocall $name=".operator-example" n="9" eg="[search:caption,description:casesensitive,words[arch]]" ie="any tiddlers containing a case-sensitive match for the word `arch` in their <<.field caption>> or <<.field description>> fields"/>
+ d
```
pmario: is this a typo?

editions/tw5.com/tiddlers/filters/search.tid
```diff
@@ -1,21 +1,44 @@
...
-op-suffix: optionally, the name of a [[field|TiddlerFields]]
-op-suffix-name: F
-op-parameter: one or more search terms, separated by spaces
+op-suffix: the <<.op search>> operator uses a rich suffix, see below for details
+op-parameter: one or more search terms, separated by spaces, or a literal search string
 op-output: those input tiddlers in which <<.em all>> of the search terms can be found in the value of field <<.place F>>
 op-neg-output: those input tiddlers in which <<.em not>> all of the search terms can be found
 
-When used with a suffix, the <<.op search>> operator is similar to <<.olink regexp>> but less powerful.
+<<.from-version "5.1.18">> Earlier versions do not support searching multiple fields, or the ''literal'' or ''casesensitive'' options.
```
pmario: `<<.from-version "5.1.18">> ` usually shows a "new in Version 5.1.18" tag .... but the text talks about "Earlier versions"... So IMO it should be rephrased, to talk about the new possibilities.

core/modules/wiki.js
```diff
@@ -1064,6 +1069,21 @@ exports.search = function(text,options) {
...
+				terms.push($tw.utils.escapeRegExp(term));
+			}
+		});
+		searchTermsRegExps = [new RegExp("(" + terms.join("\\s+") + ")",flags)];
+	} else if(options.regexp) {
+		try {
+			searchTermsRegExps = [new RegExp("(" + text + ")",flags)];			
+		} catch(e) {
+			searchTermsRegExps = null;
+console.log("Regexp error",e)
```
pmario: ???

editions/tw5.com/tiddlers/filters/examples/search.tid
Jermolene: Thanks @pmario

editions/tw5.com/tiddlers/filters/search.tid
Jermolene: Thanks @pmario

core/modules/wiki.js
Jermolene: It's not clear what we should do with a regexp error. Dumping to the console gives the user a chance to discover that there's something wrong. I'll improve the message to include the actual regexp.

Marxsal:
> Matching text: <$text text=<<match>>/>
> 
> Matching index: <$text text=<<index>>/>
> 
> Last index: <$text text=<<last-index>>/>
> 
> Capture groups:  <$text text=<<match-1>>/> <$text text=<<match-2>>/> etc.
> 
> </$match>
> ```

If you're doing that, shouldn't there be a <<match-list>>, that returns all groups? Otherwise you're forced to know in advance what matches (<<match-1>>, <<match-4>>, etc.) will be there.

Jermolene:
Hi @Marxsal 

If you're doing that, shouldn't there be a `<<match-list>>`, that returns all groups? Otherwise you're forced to know in advance what matches (`<<match-1>>`,` <<match-4>>`, etc.) will be there.

Yes to the first, and I added a comment in 
https://github.com/Jermolene/TiddlyWiki5/pull/2963#issuecomment-434387016 to the same effect. But the reason for having independent access to each match is so that you can use multiple capture groups to grab the different components of a URL, say.

twMat:
@Jermolene - first, thanks for your input on this matter. It seems a few of us don't fully understand the reasons why a widget is preferable over a filter op on this matter. For example, the snip you quote from @xcazin is:

>I'd like to obtain the following list FMCBlWgcrGc 390 out of the following string input:
https://www.youtube.com/watch?v=FMCBlWgcrGc&start=390

...but this output is trivial to get - if using two separate filterings. Is that ("two") the problem? (BTW, with #4452 it'd be one). Or is your reasoning perhaps about @xcazin 's observation that:

>Such aim leads to monster regexps like https:\/\/(youtu\.be\/|www\.youtube\.com\/(watch\?(.*&)?v=|(embed|v)\/))([^\?&"'>]+)&t=([0-9]+)(.*) for which the prefix/suffix game is no fun anymore.

...but a widget would not simplify this aspect, or?

Also, other than the effort to construct it, would it be inappropriate with *both* filter ops and a widget? As I note in #4479 , string manipulation does "feel" like a filter operator matter.

Thanks.

Jermolene:
It's fair that regexp feels like a filter operator matter but at the moment the limitations of filter operators make that difficult. In particular, a filter operator is able to return a list of results but a regular expression operation yields a match string, a list of captured groups, and a numeric position for the match, all of which can be useful. We could jam all of that information into a single list but it the filter code to manipulate it would become very clumsy and inscrutable.

Issue #3503, opened 2018-10-27
Option to tweak Sidebar background in Palette Editor


abesamma:
Hello,

**Problem**: With reference to this issue's title, the current release of TW5 as well as the pre-release versions do not have a convenient means of changing the sidebar background colour and/or gradient. You have to create a separate stylesheet tiddler to achieve the desired effect.

This can be limiting for end users who'd rather not deal with css directly.

**Proposal**: introduce an extra field in the Theme editor under the control panel's _Palette_ tab labelled _Sidebar background_ for colour picking and perhaps gradient adjustment.

PR #3507, opened 2018-10-31
Sidebar background field

abesamma:
PR to address issue #3503. Adds a field to palette editor to allow users to change the appearance of the sidebar background. 

Default values for ContrastLight is `white` and ContrastDark is `black`. Remains blank for other themes.

### DIFF
--- a/core/language/en-GB/Docs/PaletteColours.multids
+++ b/core/language/en-GB/Docs/PaletteColours.multids
@@ -47,6 +47,7 @@ pre-border: Preformatted code border
 primary: General primary
 select-tag-background: `<select>` element background
 select-tag-foreground: `<select>` element text
+sidebar-background: Sidebar background
 sidebar-button-foreground: Sidebar button foreground
 sidebar-controls-foreground-hover: Sidebar controls foreground hover
 sidebar-controls-foreground: Sidebar controls foreground
--- a/core/palettes/Blanca.tid
+++ b/core/palettes/Blanca.tid
@@ -51,6 +51,7 @@ pre-border: #cccccc
 primary: #7897f3
 select-tag-background:
 select-tag-foreground:
+sidebar-background: 
 sidebar-button-foreground: <<colour foreground>>
 sidebar-controls-foreground-hover: #000000
 sidebar-controls-foreground: #ccc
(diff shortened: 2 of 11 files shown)


Jermolene:
Thanks @abesamma I'm not 100% convinced that this is right for the core. The issue to me is that colouring the background of tc-sidebar-scrollable differently than the page background looks poor because the scrollable area abuts the tiddlers on the right:

screenshot 2018-10-31 16 56 14

It also feels inconsistent that the background isn't colored when the media query breakpoint kicks in and the sidebar moves to the top.

I can see why this might be useful in some customisations, but I think it's not universal enough to warrant inclusion in the core.

By the way, I have used a very similar technique on some of my own wikis. For example, this one uses a background colour to the tc-sidebar-scrollable but smoothly fades the colour towards the tiddler river:

https://manuals.annafreud.org/ambit/#%24%3A%2F_AFC%2FStyles%2FFadedSidebar

screenshot 2018-10-31 17 03 12

abesamma:
> 
> Thanks @abesamma I'm not 100% convinced that this is right for the core. The issue to me is that colouring the background of tc-sidebar-scrollable differently than the page background looks poor because the scrollable area abuts the tiddlers on the right:
>

I understand the concern @Jermolene. I wasn't sure whether to hard code a linear gradient to address the *abutting effect* to make it prettier (was worried restrict customisation), but you could add this for example to the field and it would help produce a more graceful appearance:

`linear-gradient(to right, transparent, black 5%)`

>
>It also feels inconsistent that the background isn't colored when the media query breakpoint kicks in and the sidebar moves to the top.
>

Right! Sorry I forgot about that breakpoint. I could redo it if you allow.

>I can see why this might be useful in some customisations, but I think it's not universal enough to warrant inclusion in the core.
>

Do you suggest I create a plugin instead? It feels too small a thing to warrant a plugin. I might as well just create a single style tiddler that everyone could use (but again, it would exclude folks who don't want to mess with too much css).

Jermolene:
> Do you suggest I create a plugin instead? It feels too small a thing to warrant a plugin. I might as well just create a single style tiddler that everyone could use (but again, it would exclude folks who don't want to mess with too much css).

It is indeed a bit small to be a plugin as it stands, but maybe worth bundling with some other similar tweaks into a worthwhile plugin.

pmario:
@abesamma -- I think this PR may be closed in favour of this **[IDEA] - provide a way to extend palettes #7736**

I do not know, if the idea will be implemented, but this PR has not been merged for 5 years. So the chance is low.

pmario:
A lot has been changed in palettes till the PR was created. So will close as not-planned

PR #3511, opened 2018-11-02
Kin filter: Recursively looking for kinship between tiddler titles

bimlas:
Usage examples:
* [Analyzing family/ToC trees](https://groups.google.com/d/msg/tiddlywiki/YZlPGP0qX1o/Qh-sAWEuCQAJ)<br>(uses an older name: `kindred[]` = `kin[]`; `kindredup[]` = `kin::to[]`; `kindreddown[]` = `kin::from[]`)
* [Display](https://groups.google.com/d/msg/tiddlywiki/YZlPGP0qX1o/tcgTCvoJCgAJ)/[export](https://groups.google.com/d/msg/tiddlywiki/YZlPGP0qX1o/Wz0UHBCSCQAJ) only a part of ToC
* [Generate ToC wihtout duplicated entries](https://groups.google.com/forum/#!topic/tiddlywiki/Z98wFWKsHCs)
* [Avoid using many tags unnecessarily](https://groups.google.com/d/msg/tiddlywiki/bL8KKIvL03Y/SWCEjO8-AwAJ)

Discussion: https://groups.google.com/forum/#!topic/tiddlywiki/YZlPGP0qX1o
Demo: https://bimlas.gitlab.io/demo/tw5/kin-filter.html

Voting: https://goo.gl/forms/bywjS7UiuaosfRlR2
Current results:
* Would you like to use this filter? 
  * 5 yes
* Do you think it should be part of TiddlyWiki itself? 
  * 5 yes
* Which name would be the best to describe the behaviour?
  * 3 kin
  * 1 kindred
  * 1 related

---

I know the suffix is not used in the usual way, but I think it's the most efficient way to pass certain parameters to the filter. For example, if you passing `from` / `to` as variables, you would not be able to crawl the tree in both directions in a single phrase, you should be writing two separate queries.

```
In my way:
<$list filter="[kin::from[Filter Expression]kin::to[Filters]]">...</$list>

Using variables (I'm not sure that this would work):
<$var name="kin-direction" value="from">
  <$list filter="[kin[Filter Expression]]">
    <$var name="kin-direction" value="to">
      <$list filter="[kin[Filters]]">
        ...
      </$list>
    </$var>
  </$list>
</$var>
```
Since I want to add more options ("depth", "field contains a single title instead of a list"), so I understand that the suffix can not be expanded indefinitely, thus the variables will be needed: the options used by all filters ("field", "field contains a single title instead of a list") could be passed as variables, but for passing options applicable to each filter individuallys, I think the suffix would be the best.

What do you think? Can I use the suffix in this form? If not, what are your suggestions for passing the various options?


pmario:
Sorry for my ignorance. ... BUT what is the SINGLE main purpose of this new operator. ... No references to discussions. maximum 3 sentences please!

bimlas:
Recursively find "direct" (linear) connections between tiddlers (based on a field like "tags").

bimlas:
It seems that I need to add an easy-to-understand picture to the reference. Maybe this helps to understand:

tree

twMat:
@bimlas - first, let me say you've done a wonderful job in presenting this. Docs and all.

IMO, this is a filter operator that **outputs as if it were a widget or a macro**. It kind of feels unlikely that it would be used as a filter operator in the usual sense, e.g where succeeding operators do something with its output. Or maybe my imagination is too limited?

The key question is: Is it justified to ship this to *everyone* who downloads a TW? Will people *likely* have use of it? IMO: no (...but, admittedly, I'd say the same about a few other things that *are* in the standard distro). One idea might be to re-shape it into e.g a macro or so but I'm not sure exactly what it should do. Or could it, or part of it, be made to extend the existing ToC functionality somehow?

I have a feeling I will use this some day and I hope it'll be easily found when I will need it one day. Big thanks for making it.

bimlas:
It does not seem clear what the filter is doing. This is a general filter, *I use ToC only to visualize the results*, only a list of titles that match the condition appears at the output of it (as with any other filter), you can try with Advanced Search.

> What is the SINGLE main purpose of this new operator?

Let me try to explain again.

* Finds the "leaves" of the branch of the specified element in a tree-like structure (where the given tag / field item is a leave).
* Finds the super- and subsets / groups of a mathematical set (where the given tag / field item is a set).
* Finds the ancestors and successors of a family member.

So, it finds out where a given item originates and what other elements originate from it.

It can be used in a variety of ways, thus I've put some examples in the opening comment.

Jermolene:
Hi @bimlas apologies for the brief interjection, but I do wonder if this new operator might make it possible to re-implement the TOC macros more efficiently/readably.

If so, then I think it's a good candidate for the core. If not, then perhaps it's better to keep it as a plugin while users experiment with it and assess where it might be most useful.

bimlas:
Hi @Jermolene !

What do you mean by "more efficiently / readably"? I do not know the exact way, but the filter makes a macro able to

* Display an item at the deepest level that is linked to multiple tags, for example in the demo this means "ToC -> Reference -> Concepts ->Filters" for "Filters" ("ToC -> Reference -> Filters" is in a "higher" level).
* Based on the above example I can imagine a "dynamic table of contents" where you do not have to look ahead to ToC to know where to place a tiddler, just add all the tags that characterize it (Filters, Example, Syntax) and it "falls into the appropriate compartment" - only the hierarchy of the tags should be established.

pmario:
@bimlas Thank you very much for 3 point explanation and you great contribution!! 

IMO this info should be the "short description" for the docs. It makes things much clearer, without the need for heavy testing ;)

Jermolene:
Hi @bimlas apologies I don't think I had grasped what you're trying to do with this operator, but I think it's clearer now.

The usual determination for whether something should go into the core centres on whether it is universally useful. That's sometimes a tricky debate because it leaves room for matters of opinion. But, if we were able to use this new filter to improve an existing core feature then the debate would be much easier.

In the meantime, I noticed a few coding style issues and will leave some comments. (Our coding standards are not well documented -- see https://tiddlywiki.com/dev/#TiddlyWiki%20Coding%20Style%20Guidelines).

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+Filter operator that recursively finds kindred between tiddlers
+
+\*/
+(function () {
+
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  "use strict";
+
+  // TODO: Should I set global tw to true?
```
Jermolene: Yes, I think it should be true here (I notice that quite a few source files set it wrongly)

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+
+  function collectFamilyMembers(tiddler, title, fieldname, direction) {
```
Jermolene: Coding style: don't include spaces in parameter lists

bimlas:
I just created a demo for ToC: It's removing the redunant elements, puting them on the "deepest" level. I have to work on it to remove the duplicated items ("String in music/computer science").

See the tags, it's just a draft, thus I did not documented it well, but I hope it's clear.

https://bimlas.gitlab.io/demo/tw5/kin-filter-toc-demo.html

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+    var family_members_from = [],
```
Jermolene: Coding style: favour camelCase over underscores in identifier names

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+      family_members_to = [];
+
+    function addToResultsIfNotFoundAlready(list, title) {
+      if (list.includes(title)) {
```
Jermolene: Coding style: don't include a space between `if` and its bracket

pmario:
@Jermolene I think it should be part of the core. ... but imo it isn't ready for prime time yet. ... 

I did review the code a little bit. There are some internal functions eg: ` function uniqueArray(input) {` that are already part of the core. Or at least they are very close. So this should be optimized. 

The documentation needs to start with "simpler" more basic examples. The existing ones are sophisticated but too complicated for new users. 

We should play a little bit more with a plugin and then see how it goes. 

I see it more as an "analysing tool" for the internal document structure. 

It also extends the possibility to have more then 1 suffix parameter. We need to check if the mechanism is generic enough, that we can use it for future filters, or extend it even more ... I'm not sure here!

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+        return false;
+      }
+      list.push(title);
+      return true
+    }
+
+    function findKindredFrom(tiddler, title) {
+      if (addToResultsIfNotFoundAlready(family_members_from, title)) {
+        if (tiddler) {
+          tiddler.getFieldList(fieldname).forEach(function (target_title) {
```
Jermolene: Coding style: We avoid ES2015 `forEach` in favour of the `$tw.utils.each` method -- this is partly for performance, but mainly to retain compatibility with older JS platforms.

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+
+    function findKindredTo(title) {
+      if (addToResultsIfNotFoundAlready(family_members_to, title)) {
+        $tw.wiki.findListingsOfTiddler(title, fieldname).forEach(function (target_title) {
+          findKindredTo(target_title);
+        });
+      }
+    }
+
+    if ((direction === 'from') || (direction === 'with')) {
```
Jermolene: Coding style: prefer double quotes for string constants

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+      findKindredFrom(tiddler, title);
+    }
+    if ((direction === 'to') || (direction === 'with')) {
+      findKindredTo(title);
+    }
+    return uniqueArray(family_members_from.concat(family_members_to));
+  }
+
+  // TODO: Is there a better way for unique?
+  function uniqueArray(input) {
```
Jermolene: There's a core function `$tw.utils.pushTop()` that you may be able to use instead?

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+        output[j++] = item;
+      }
+    }
+    return output;
+  }
+
+  /*
+    Export our filter function
+
+    TODO: May I add tests? (editions/test/tiddlers/tests)
```
Jermolene: Yes indeed, and if necessary add a new JS test file

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+    */
+  exports.kin = function (source, operator, options) {
+    var results = [],
+      needs_exclusion = operator.prefix === '!',
+      suffix_list = (operator.suffix || '').split(':'),
```
Jermolene: The PR #3502 adds support for a new `operator.suffixes` structure for standardised parsing of the suffix. Can it be used here?

core/modules/filters/kin.js
```diff
@@ -0,0 +1,107 @@
...
+      });
+    } else {
+      source(function (tiddler, title) {
+        results = results.concat(collectFamilyMembers(tiddler, title, fieldname, direction));
+      });
+      results = uniqueArray(results);
+    }
+
+    return results;
+  }
```
Jermolene: BTW GitHub is showing the code as being indented with two spaces. We use tabs, not spaces, and treat them as 4 characters wide.

Jermolene:
> It also extends the possibility to have more then 1 suffix parameter. We need to check if the mechanism is generic enough, that we can use it for future filters, or extend it even more ... I'm not sure here!

We also did that in #3502; I've left a code review comment to that effect.

pmario:
>We also did that in #3502; I've left a code review comment to that effect.

So would it be possible to extract the new logic and make it available for other filters, so we can avoid code duplication?

pmario:
Just some thoughts. 

The existing TOC builds the structure with wikitext, starting from the "root" tag down to the leafs. This mechanism has the problem, that tiddlers with more than 1 tag show up in the TOC at every "level". 

The kin filter internally goes down to the leafs and lets us build a "less" cluttered TOC.

pmario:
Depending on the "settings" it also has some additional possibilities. 

> - Finds the super- and subsets / groups of a mathematical set (where the given tag / field item is a set).
> - Finds the ancestors and successors of a family member.

Jermolene:
Hi @pmario 

> So would it be possible to extract the new logic and make it available for other filters, so we can avoid code duplication?

That's already been done as part of that PR.

bimlas:
Found a shorter way to build ToC:

```

<<toc 'Example tree for kin filter' '] -[tag<currentTiddler>tagging[]kin::to[]' >>
</div>
```

bimlas:
I apologize for the short comment, but I was on the bus and I can not write text on the phone.

I've updated the ToC demo, I've included a table of contents that (almost) no longer contains duplicate elements.

https://bimlas.gitlab.io/demo/tw5/kin-filter-toc-demo.html

bimlas:
> In the meantime, I noticed a few coding style issues and will leave some comments. (Our coding standards are not well documented -- see https://tiddlywiki.com/dev/#TiddlyWiki%20Coding%20Style%20Guidelines).

Since I had to format the source code, I thought it would be useful to set up my IDE (IntelliJ IDEA) auto-formatting style, so I did it and I like to share the settings to facilitate the work of others. If you find it good, you can put a link to it in the documentation.

https://gitlab.com/bimlas/tw5-code-style-auto-format

Jermolene:
Hi @bimlas that's great, do please make a PR to add it to the "Resources" section of the "Community" listing.

Jermolene:
Hi @bimlas the new processing for the suffix splits it up first by colons, and then by commas.

So you could use commas to separate your parameters so that they are easier to read:

```
[kin:parent,to,2,single[Grandpa]]
```

bimlas:
As I see https://tiddlywiki.com/prerelease/#search%20Operator it seems to be a good idea that the colon character separates the logical elements, and the comma separates the elements of the list, so I think that I will stay in the original form if it suits you.

Jermolene:
Hi @bimlas yes I think that makes sense

bimlas:
How do I find the titles where the field is a specific text? I think the `$tw.wiki.findListingsOfTiddler()` do not works properly in this case, because it looks at the fields as a list, but I want to find the tiddlers where the value of the field is exactly equal to the given value (the target title).

bimlas:
I forgot to write why I need it:

> * "Type" of field, treat as a single title or a list (usefull for PMario's [tocP](https://wikilabs.github.io/editions/tocP/), `single`/`list`)

Jermolene:
Hi @bimlas

> How do I find the titles where the field is a specific text?

There is indeed no support for that at the moment. We could add a new method to the wiki object: `findTiddlersByField`.

One likely issue is the way that the core treats the `list` and `tags` fields specially, presenting their value as an array rather than a string. I imagine you're going to want to search arbitrary fields but interpret them as a list, so you may need to manually parse them (with `$tw.utils.parseStringArray()`).

bimlas:
I've finished writing the filter, but the tests are not ready yet. You can try it on https://bimlas.gitlab.io/demo/tw5/kin-filter.html

For the time being, I doubt about how the type of field (singlet / list) should be handled, I think it should be discussed in another issue before implementation, but the filter works in most cases without it, so this option is not yet provided.

I also dropped the minimum / maximum depth because it would have made the code worse, but it would not have been profitable.

Please try to look at the documentation as a beginner to make sure it is understandable.

bimlas:
I got stuck in writing tests: The following example does not work, the output only includes `one`. If I replace `kin` with `tag`, the output is correct, it's listing `Tiddler Three` and `TiddlerOne`.

```
	it("should handle the kin operator", function() {
		expect(wiki.filterTiddlers("[kin[one]]").join(",")).toBe("one,Tiddler Three,TiddlerOne");
	});
```

Because the filter works, so I think there might be something wrong with the test environment, something different from the "live" system I did not handle correctly in the filter. Do you have any idea what I've been missing?

Jermolene:
Hi @bimlas I don't know of anything obviously different between the environments. Are you runnings the tests in the browser or under Node.js?

bimlas:
I ran `./bin/test.cmd` (I'm on Winodws if it makes matter), then I tried `./bin/serve.cmd editions/test` (so far I did not know it was possible), the results are the same.

bimlas:
Just found the HTML output, the tests show the same results there.

bimlas:
Note: the `findListingsOfTiddler` method is used only by `kin` and `listed`. `listed` is not included in any of the tests, thus the `findListingsOfTiddler` has had no impact on the tests until now. However, it does seem to work correctly, although I did not test it very much yet. Just want to let you know.

Jermolene:
Thanks @bimlas -- could we refactor `findListingsOfTiddler` to use the new `findTiddlersByField` method?

bimlas:
@Jermolene Sorry, but I do not know what you want because I can not find this method anywhere.

bimlas:
I tried to find out the cause of the error, I placed `console.log ()` lines in the code and it looks like these are causing trouble:

https://github.com/Jermolene/TiddlyWiki5/blob/375099470815eedec684dac0237f531bc3dcfa9f/core/modules/filters/kin.js#L33-L35

https://github.com/Jermolene/TiddlyWiki5/blob/375099470815eedec684dac0237f531bc3dcfa9f/core/modules/filters/kin.js#L45-L47

The "each" function blocks are not executed. I tried placing the following lines in the code, the output appeared, so `each` works.

```
$tw.utils.each(["A", "B", "C"], function(current) {
  console.log(current);
});
```

Some functions seems broken *in tests* and *only for `kin`*... o_O Weird... I placed these lines in to the main function (`exports.kin = function(...)`):

```
console.log($tw.wiki.findListingsOfTiddler('one','tags'));
// []
console.log($tw.wiki.getTiddler('TiddlerOne'));
// undefined
```

Is the use of global `$tw` causing trouble? But why are not the other tests wrong? Using `options` instead of `$tw` works:

```
console.log(options.wiki.findListingsOfTiddler('one','tags'));
// [ 'TiddlerOne', 'Tiddler Three' ]
console.log(options.wiki.getTiddler('TiddlerOne'));
// { fields:
//    { title: 'TiddlerOne',
//    ...
```

So it looks like the global `$ tw` is causing trouble. When will I have time, I will continue the investigation.

bimlas:
I tried to put this code at the beginning of `tag` operator:

```
console.log(options.wiki.getTiddlersWithTag(operator.operand));
// correct results
console.log($tw.wiki.getTiddlersWithTag(operator.operand));
// []
```

Then I searched where the `$ tw.wiki` object was used: only I use it. :dagger: 

Somewhere in the documentation we has to write that if you want to access the wiki, you always have to use `options.wiki`, never` $tw.wiki`.

bimlas:
I'm ready, the branch can be merged. Sorry for the forced push, I forgot to delete TODO comment about tests.

bimlas:
> Thanks @bimlas -- could we refactor `findListingsOfTiddler` to use the new `findTiddlersByField` method?

Correction: I have not fixed this yet because I can not find the method.

Jermolene:
Hi @bimlas apologies for the confusion. We need to implement `findTiddlersByField()` as part of this PR.

bimlas:
Got it, OK, I will do it soon. :+1:

bimlas:
@Jermolene I'm done.

bimlas:
Ping @Jermolene : Please merge me!

bimlas:
@Jermolene , if you think that the code is OK and it should be in the core, then merge it please.

Jermolene:
Hi @bimlas apologies, I've not had a chance to properly review this PR yet (it takes a long time to properly review changes like this I'm afraid). I'll leave a couple of notes now for some issues I spotted on a brief glance. Part of the bigger review it needs is deciding whether this should go into the core, I'm open minded at the moment.

core/modules/filters/kin.js
```diff
@@ -0,0 +1,96 @@
...
+		function addToResultsIfNotFoundAlready(list,title) {
+			if(list.indexOf(title) !== -1) {
+				return false;
+			}
+			list.push(title);
+			return true
+		}
+
+		function collectTitlesPointingFrom(tiddler,title,currentDepth) {
+			if((options.depth) && (currentDepth++ > options.depth)) {
```
Jermolene: As a matter of defensive programming style, I'd advise against using the increment operator in this way (the increment will only be performed if that part of the expression is evaluated, and JavaScripts expression shortcutting rules mean that that is not always obvious). It's clearer to perform the increment in a separate statement.

core/modules/filters/kin.js
```diff
@@ -0,0 +1,96 @@
...
+Finds out where a tiddler originates from and what other tiddlers originate from it
+
+\*/
+(function() {
+
+	/*jslint node: true, browser: true */
+	/*global $tw: true */
+	"use strict";
+
+	function collectTitlesRecursively(baseTiddler,baseTitle,options) {
```
Jermolene: A big issue to consider with this operator is efficiency. You'll notice that the wiki object has some features for caching expensive computations, and that these are used for most of the tag-related operators. I think we need to do something similar here to make this moderately efficient.

core/modules/wiki.js
```diff
@@ -522,7 +522,7 @@ exports.getTagMap = function() {
 /*
 Lookup a given tiddler and return a list of all the tiddlers that include it in the specified list field
 */
-exports.findListingsOfTiddler = function(targetTitle,fieldName) {
```
Jermolene: Here you've deleted the original `findListingsOfTiddler()` method. That breaks backwards compatibilty. I was suggesting instead that the `findListingsOfTiddler()` be refactored to use the new `findTiddlersByField()` method.

bimlas:
I apologize for sending you such messages, I do not want to bother you, I just do not know why you do not merging the finished PRs; it seems like you just forget about them. By adding a label to it, I think you might be saying that PR is not forgotten, it just "waiting for review".

I understand you have to think and decide things; if the filter does not get into the core, I do not mind, you know better than me. Ultimately, thanks to your choices and knowledge, this program is stable for 14 years. Thank You!

Jermolene:
Hi @bimlas gosh sorry if I sounded brusque, not intended. And please don't worry about reminding me of things that I may have forgotten -- in fact, I think  I rely on people doing that :)

core/modules/wiki.js
bimlas: I'm still confused about this: `findTiddlersByField()` is not exists anywhere. Earlier (https://github.com/Jermolene/TiddlyWiki5/pull/3511#issuecomment-437330680) you said, that I need to implement. So you would like me to create this method and replace the contents of `findListingsOfTiddler()` with the calling of the new method (so the it would just be an "alias")?

bimlas: In other words: would you like me to extract the body of `findListingsOfTiddler()` as a new `findTiddlersByField()` method, then call the latter from it?

bimlas:
I've updated a demo so you can try it without building it: https://bimlas.gitlab.io/demo/tw5/kin-filter.html

Jermolene:
Hi @bimlas I'm really sorry about all the confusion about `findListingsOfTiddler()`. To explain again what I am recommending:

* That the search logic of your filter be implemented as a new method on the wiki object called "findTiddlersByField"
* To then consider refactoring "findListingsOfTiddler" to use the new method if it is more efficient than the current implementation

To be clear, there would be no need to change "core/modules/filters/listed.js".

bimlas:
Sorry, probably because of my weak English knowledge, we do not understand each other so much.

> @bimlas: In other words: would you like me to extract the body of `findListingsOfTiddler()` as a new `findTiddlersByField()` method, then call the latter from it?
> @Jermolene : That's right: it's a two step process: first add the new function, then second refactor the existing function to use it. That way we retain full backwards compatibility
> ...
> @Jermolene : That the search logic of your filter be implemented as a new method on the wiki object called "findTiddlersByField"

I really do not understand what you want, I'm sorry. :disappointed: Perhaps the simplest thing would be to do it by yourself. :(

If you want the [search mechanism of `kin` filter](https://github.com/Jermolene/TiddlyWiki5/blob/717e6f5b584fc8a0c051f68bc86dedcf7bb23538/core/modules/filters/kin.js#L15-L75) to be a method in the `wiki.js` with the `findTiddlersByField()` method name, then this method should get all the parameters that are used by `kin` (field, depth, direction) as this is a recursive function - you want that? So the `findListingsOfTiddler()` method would call the `kin` filter limited at depth 1.

Jermolene:
Hi @bimlas not to worry, I'll pick this up after we get v5.1.18 released. Please don't hesitate to remind me if necessary.

core/modules/wiki.js
```diff
@@ -101,7 +101,7 @@ exports.deleteTextReference = function(textRef,currTiddlerTitle) {
 exports.addEventListener = function(type,listener) {
 	this.eventListeners = this.eventListeners || {};
 	this.eventListeners[type] = this.eventListeners[type]  || [];
-	this.eventListeners[type].push(listener);
+	this.eventListeners[type].push(listener);
```
pmario: what should these spaces be good for?

bimlas: They are not good at all, but deletion of whitespace is unrelated to this PR, but their removal would make `git blame` dirty.

pmario: ah. ok. ...

bimlas:
I've tried the filter in the wild and indeed it is very useful, but I'm not sure it can make the core better.

The plugin I use to: https://bimlas.gitlab.io/tw5-locator/

I use the filter to list tiddlers associated with the given tags at any depth and to avoid duplication in ToC (filtering out "grandchildren", as I did this earlier in this thread: https://github.com/Jermolene/TiddlyWiki5/pull/3511#issuecomment-436308981). 

Since it is unlikely that this kind of table of contents will get into the core, I think we would only use it to filter out duplications from the classical ToC, which is slow. [Here is a copy of tw.com](https://bimlas.gitlab.io/demo/tw5/locator-filtering-out-duplications-from-toc.html) with the plugin mentioned above: "With duplications" sidebar tab does not filter duplications, "Without duplications" does. For example navigate to Reference, switch between these tabs and see the results (performance instrumentation enabled)

Do you think I should implement a separate plugin? If so, I leave only those parts of the PR that affect the core. If you see other benefits to the filter, I'd be happy if it would be in the core.

Jermolene:
Hi @bimlas

> Do you think I should implement a separate plugin? If so, I leave only those parts of the PR that affect the core. If you see other benefits to the filter, I'd be happy if it would be in the core.

I think it does make sense to go ahead and make it a plugin, that's probably the best way to find out how useful it is to other people. Your experience with performance matches my concerns about the inherent slowness of the filter (but of course we've already got slow filters, like missing and orphans).

bimlas:
While converting the PR to a plugin, I wondered if there is anything in the modifications that has to be merged with the master or should I simply close the PR?

https://github.com/Jermolene/TiddlyWiki5/pull/3511/files#diff-3378be2ff14e6ff5521c34b06144efeb
https://github.com/Jermolene/TiddlyWiki5/pull/3511/files#diff-89406c65507116b5c42accc5f52da5da
https://github.com/Jermolene/TiddlyWiki5/pull/3511/files#diff-4aa6df7cefc567c1f777eacdf3031ec9 (used only by this plugin, but I extracted this method to the plugin itself)

Jermolene:
Hi @bimlas 

> I wondered if there is anything in the modifications that has to be merged with the master or should I simply close the PR?

We should discard the core changes -- assuming you can make the plugin without them?

bimlas:
Yes, I can make it, closing the PR.

bimlas:
Released the plugin: https://bimlas.gitlab.io/tw5-kin-filter/

PR #3512, opened 2018-11-04
add new-tiddler new-journal new-image keyboard shortcuts

BurningTreeC:
this adds global keyboard shortcuts for `new-tiddler`, `new-journal` and `new-image`

it also refactors the involved actions into `action-tiddlers` ($:/core/ui/Actions/new-tiddler ...) and makes the PageControls Buttons use them

for the new-image action it also adds a `$:/config/NewImageType` setting for jpeg/png which is jpeg by default. the corresponding action tiddler uses that config for the new-image type (probably a separate PR but maybe it's ok here?)

### DIFF
--- a/core/ui/Actions/new-image.tid
+++ b/core/ui/Actions/new-image.tid
@@ -1,0 +1,10 @@
+title: $:/core/ui/Actions/new-image
+tags: $:/tags/Actions
+description: create a new image tiddler
+
+\define get-type()
+image/$(imageType)$
+\end
+<$vars imageType={{$:/config/NewImageType}}>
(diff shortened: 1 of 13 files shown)


Jermolene:
> for the new-image action it also adds a $:/config/NewImageType setting for jpeg/png which is jpeg by default. the corresponding action tiddler uses that config for the new-image type (probably a separate PR but maybe it's ok here?)

I think it makes sense to include it here, given the degree of refactoring we're now doing. But perhaps we should add $:/config/NewImageType to the "Hidden Settings" docs tiddler? https://tiddlywiki.com/prerelease/#Hidden%20Settings

Jermolene:
Thanks @BurningTreeC

PR #3514, opened 2018-11-06
let new tiddlers follow openLinkFromOutsideRiver

BurningTreeC:
this makes new tiddlers appear at the bottom of the story list when they get created and `openLinkFromOutsideRiver` is `bottom`


Jermolene:
Thanks @BurningTreeC

PR #3515, opened 2018-11-06
fix error opening new windows with popup-blocker

BurningTreeC:
this prevents errors when a popup blocker blocks opening a new window from within a running TW


Jermolene:
Terrific, thanks @BurningTreeC

PR #3516, opened 2018-11-06
FIX "Observe openLinkFromOutsideRiver ..."

BurningTreeC:
I assumed the attribute to be available but it's not


Jermolene:
Thanks @BurningTreeC

PR #3520, opened 2018-11-07
FIX: classic story view opening/closing tids at bottom

BurningTreeC:
this fixes the classic storyview when tiddlers get inserted or removed at the very bottom of the story river

now that  new tiddlers open at the bottom when `openLinkFromOutsideRiver` is `bottom`, this glitch in the animation should be fixed I think

### DIFF
--- a/core/modules/storyviews/classic.js
+++ b/core/modules/storyviews/classic.js
@@ -45,7 +45,7 @@ ClassicStoryView.prototype.insert = function(widget,options) {
 		currMarginBottom = parseInt(computedStyle.marginBottom,10),
 		currMarginTop = parseInt(computedStyle.marginTop,10),
 		currHeight = targetElement.offsetHeight + currMarginTop,
-	    	isBottom = options.index === options.listLength - 1;
+	    	isBottom = options.listLength !== 1 && options.index === options.listLength - 1;
 	var initialStyle = isBottom ? [{marginTop: (-currHeight) + "px"}] : [{marginBottom: (-currHeight) + "px"}];
 	var finalStyle = isBottom ? [{transition: "opacity " + duration + "ms " + easing + ", " +
 						"margin-top " + duration + "ms " + easing},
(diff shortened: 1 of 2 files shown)


core/modules/storyviews/classic.js
```diff
@@ -33,7 +33,7 @@ ClassicStoryView.prototype.navigateTo = function(historyInfo) {
 	this.listWidget.dispatchEvent({type: "tm-scroll", target: targetElement});
 };
 
-ClassicStoryView.prototype.insert = function(widget) {
+ClassicStoryView.prototype.insert = function(widget,bottom) {
```
Jermolene: Rather than passing this "bottom" flag, perhaps we can update the insert method to take a second parameter called "options", and pass members for the index of the item, and the total number of items?

BurningTreeC: changed it, @Jermolene

BurningTreeC:
@Jermolene - in this PR the animation for tiddlers at the bottom is not exactly the same as anywhere else. I believe I'd have to use the top margin here to make it behave the same way... I'd have to figure it out

Jermolene:
Hi @BurningTreeC I haven't had a chance to try a build of these changes but it seems reasonable to improve things.

In the background, I've been thinking for a while that I'd like to do some major surgery on the Classic storyview, to add the zooming rectangle from the classic TW animation you see at 
https://classic.tiddlywiki.com/

BurningTreeC:
@Jermolene - yes, that would give it an even more `classic` feel, I like the idea

Jermolene:
Hi @bimlas thanks -- I'm mobile this week and so can't run a build, so I wondered if you might be able to put one up somewhere like TiddlySpot? Many thanks.

BurningTreeC:
@Jermolene - I'm working on the animation to be better ... I'll leave a link here

BurningTreeC:
Sorry, closing this for now -  I've seen that this needs more thoughts and testing

PR #3523, opened 2018-11-07
fix NewImageType config

BurningTreeC:
the NewImageType config is erraneous

somehow I couldn't create a config tiddler without new-line at the end other than from within a tiddlywiki


Jermolene:
Hi @BurningTreeC - the best way to do it is to make the text field explicit in the header of the file:

```
title: boo
text: far
```

Jermolene:
See the definition of $:/config/MissingLinks for an example

BurningTreeC:
thanks @Jermolene, now I got it!

Issue #3474, opened 2018-10-09
indentUnit, tabSize and cursorBlinkRate should have field type="integer"


NikolayMakhonin:
Now these configs is a strings, and this leads to problems, for example with tabulation (TAB button)
(2 + "2" = 22)

$:/config/codemirror/indentUnit
$:/config/codemirror/cursorBlinkRate
$:/config/codemirror/tabSize

BurningTreeC:
Thanks @NikolayMakhonin , I've created a PR for it

PR #3524, opened 2018-11-07
update CodeMirror config tiddlers

BurningTreeC:
this uses the "text" field configuration scheme for the CM config tiddlers 

AND

sets the type of cursorBlinkRate, indentUnit and tabSize to `integer` as of issue #3474


Jermolene:
Thanks @BurningTreeC

pmario:
>this uses the "text" field configuration scheme for the CM config tiddlers

Is there a special reason, why you did it that way? .. or is it just optics?

BurningTreeC:
@pmario yes, the select inputs in the ControlPanel tab don't show the default value because of new-lines within the config files ... using a `text field` fixes that issue

PR #3526, opened 2018-11-07
tree macro - fix for double quotes

morosanuae:
This is trying to fix the problem with double quotes (") in branches/leaves names.
One can try to replicate the problem by creating a system tiddler like this: $:/_temp/"abc" or $:/"_temp"/abc and then see how it appears in the "Explorer" sidebar tab.
I'm using double quotes in standard tiddlers titles in my day to day work and this problem appears when I create state tiddlers automatically in lists (e.g. for popups).

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -1,13 +1,18 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
+\define link(full-title, chunk)
+<$link to=<<__full-title__>>>
+<$text text=<<__chunk__>>/>
+</$link>
+\end
+
 \define leaf-node(prefix)
 <li>
-<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]] +[removeprefix[$prefix$]] +[limit[1]]""" 
-emptyMessage="""<$text text="$prefix$$(chunk)$"/>""">
-<span>{{$:/core/images/file}}</span> <$link to="""$prefix$$(chunk)$""">
-<$view field="title"/>
-</$link> 
+<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]]""" variable="full-title">
+<$list filter="""[<full-title>] +[removeprefix[$prefix$]]""" variable="chunk">
+<span>{{$:/core/images/file}}</span> <$macrocall $name="link" full-title=<<full-title>> chunk=<<chunk>>/>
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @morosanuae (and good to see you back!). @BurningTreeC has been looking at similar problems, but I don't think has tackled this yet.

morosanuae:
Thanks @Jermolene, I'm also glad to be back! If you find some mistakes or unnecessary things in the code please feel free to correct them. I've made some tests but I don't know if I managed to resolve all the problems.

Jermolene:
Hi @morosanuae thanks! I'm mobile at the moment and so can't test the code properly, but I'm still seeing some textual parameter substitution (ie `$param$` and `$(var)$` within a macro definition) -- to properly fix things I would expect to need to replace all of them with other formulations.

BurningTreeC:
Hi @morosanuae , I think @Jermolene is talking about the `<<__variable__>>` Syntax

morosanuae:
@Jermolene do you mean to replace all other textual parameter substitutions, for example `$prefix$` with `<<__prefix__>>`?

Jermolene:
Hi @morosanuae yes, just that. There's been a few recent commits where we've addressed these problems:

https://github.com/Jermolene/TiddlyWiki5/commits/6c9dfd7f6294ef7203a9163fffc71509fa6719d6/core/wiki/macros/tag.tid

https://github.com/Jermolene/TiddlyWiki5/commits/698224556ea0901adf2d04cd5d536edd652f700f/core/wiki/macros/toc.tid

@BurningTreeC am I missing any others?

BurningTreeC:
@morosanuae - I was trying to accomplish the same for tag-pills, the tag-dropdowns and states using `bad titles` as I call them :grin: 

One PR got merged recently: #3437 
Another one is open: #3442 

using the `<<__variable__>>` syntax consistently and where needed fixes the rendering problems

there are some problems left with the generation of states within the reveal and button widgets but they are mostly with titles containing `!!` and `##` (state references). a good fix for that problem is still todo

morosanuae:
@Jermolene, @BurningTreeC I was aware of these commits/discussions when I published my fix for this kind of problems. So I decided that this is a good moment to join in. I encountered the same problem with double quotes also in the "toc" macros and this keeps me from using them in my day to day work.

morosanuae:
@Jermolene, @BurningTreeC  I'm working on a second version that I'll publish soon. I'll take the code changes in the commits you have mentioned as a model.

BurningTreeC:
cool, thanks @morosanuae !

PR #3529, opened 2018-11-09
stateTitle stateField stateIndex attributes

BurningTreeC:
this adds following attributes to the reveal widget:

- `stateTitle`
- `stateField`
- `stateIndex`

these attributes are new for the button widget:

-  `setTitle`
- `setField`
- `setIndex`
- `popupTitle`

<hr>

it uses `getTiddlerText` / `getFieldString` / `extractTiddlerDataItem` respectively `setText` when `stateTitle` is used ... instead of `getTextReference` / `setTextReference`

the button widget passes a `noStateReference` property to `triggerPopup` when `popupTitle` is used

from the logic this should be sufficient to make state-breaking titles work. For other attributes like `to` (button widget) there should be workarounds available

if this can get somewhere, I'd add the necessary documentation

### DIFF
--- a/core/modules/utils/dom/popup.js
+++ b/core/modules/utils/dom/popup.js
@@ -134,9 +134,13 @@ Popup.prototype.show = function(options) {
 			height: options.domNode.offsetHeight
 		};
 	}
-	options.wiki.setTextReference(options.title,
-			"(" + rect.left + "," + rect.top + "," + 
-				rect.width + "," + rect.height + ")");
+	var popupRect = "(" + rect.left + "," + rect.top + "," + 
+				rect.width + "," + rect.height + ")";
+	if(options.noStateReference) {
+		options.wiki.setText(options.title,"text",undefined,popupRect);
+	} else {
+		options.wiki.setTextReference(options.title,popupRect);
+	}
 	// Add the click handler if we have any popups
 	if(this.popups.length > 0) {
 		this.rootElement.addEventListener("click",this,true);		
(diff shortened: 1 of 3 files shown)


Jermolene:
Hi @BurningTreeC apologies, just realised that I've merged this and #3531 in the wrong order! :smile:

BurningTreeC:
Hi @Jermolene :grin: no problem!

BurningTreeC:
very happy this is merged, thank you @Jermolene !

morosanuae:
Hi @BurningTreeC I'm a little confused about these new attributes. Why did you split the initial "state" and "set" attributes in multiple attributes? Couldn't these be made to support all text references types (title, field, index) in a single attribute, like before? Apologies, I just want to understand because the new attributes names and purposes are, at least for me, not very intuitive. Thanks.

morosanuae:
The fixes available for special characters (e.g. double quotes) are available only for these new attributes and not for the old ones?

Jermolene:
Hi @morosanuae the current scheme is to use a single attribute containing a text reference (ie, `HelloThere`, `HelloThere!!caption`, `HelloThere##index` etc.). The trouble comes when one wants to use the character sequences `!!` or `##` within a tiddler title. This option of using these new "unpacked" attributes makes it possible.

We realised we needed to do this when @BurningTreeC was working on one of the recent PRs to fix usage of special characters.

Jermolene:
@BurningTreeC I just realised that we need to update the docs; are you happy to take a shot? Many thanks.

morosanuae:
Hi @Jermolene ok, I understand now, thanks The solution is not that nice but we need this. I, personally, need this because I have a lot of problems with special characters.

PR #2347, opened 2016-03-27
Add option to disable linking to missing tiddlers

sukima:
This closes Issue #2340

PR #3530, opened 2018-11-10
tv-override-missing-links: fix links to missing tiddlers

BurningTreeC:
this adds a variable `tv-override-missing-links` that gets detected by the `link widget`

that enables us to render links like in the `Filter` dropdown of the `Advanced Search` , even if  `enable links to missing tiddlers` is unchecked in the Control-Panel-Settings-Tab

See discussion in the groups: https://groups.google.com/forum/#!topic/TiddlyWiki/E7qL1-hOOxM

fixes:

- filter dropdown in AdvancedSearch
- type dropdown in EditTemplate
- field-name dropdown in EditTemplate

### DIFF
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -37,9 +37,10 @@ LinkWidget.prototype.render = function(parent,nextSibling) {
 	// Get the value of the tv-wikilinks configuration macro
 	var wikiLinksMacro = this.getVariable("tv-wikilinks"),
 		useWikiLinks = wikiLinksMacro ? (wikiLinksMacro.trim() !== "no") : true,
-		missingLinksEnabled = !(this.hideMissingLinks && this.isMissing && !this.isShadow);
+		missingLinksEnabled = !(this.hideMissingLinks && this.isMissing && !this.isShadow),
+	    	overrideMissingLinksEnabled = this.getVariable("tv-override-missing-links") === "true";
 	// Render the link if required
-	if(useWikiLinks && missingLinksEnabled) {
+	if(useWikiLinks && (missingLinksEnabled || overrideMissingLinksEnabled)) {
 		this.renderLink(parent,nextSibling);
 	} else {
 		// Just insert the link text
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -191,7 +191,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 LinkWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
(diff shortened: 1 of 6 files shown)


Jermolene:
Hi @BurningTreeC this is a really interesting one. I'm afraid that merging the original PR #2347 was a mistake; it's what happens if I let my guard down and merge PRs without properly thinking them through. (To be clear, the fault here is entirely mine: it's my responsibility to act as a gatekeeper and not blindly wave things through).

It should never have been a global setting, but instead should have always been controlled by a variable `tv-hide-missing-links`, just like a bunch of other settings of the link widget. Then right at the top of the page template we can assign that variable from transcluding `$:/config/MissingLinks`. And then, just as you suggest, we'd need to wrap certain components with `tv-hide-missing-links=no`.

BurningTreeC:
@Jermolene is this fix here that much worse in terms of performance?

Jermolene:
> is this fix here that much worse in terms of performance?

The fix here is more complex than it needs to be because of the earlier mistake. That contributes to making  the core code harder to follow and understand, so I'd be inlined to clean things up properly.

BurningTreeC:
Now the variable used is `tv-hide-missing-links="no"`

Jermolene:
> I understand - what about backwards compatibility @Jermolene? Here I'd be much more concerned than in #3572

I think the semantics of the revised implementation I'm proposing wouldn't introduce any problems; there would still be support for $:/config/MissingLinks but the problem areas would be fixed, and there'd be a new config variable.

BurningTreeC:
@Jermolene , I've updated the PR ... I think this is similar to how you wanted it to be ... ?

Jermolene:
Great stuff, thank you @BurningTreeC!

BurningTreeC:
@Jermolene, is it correct that the link widget doesn't have to refresh anymore if $:/config/MissingLinks changes? because I removed that

Jermolene:
> @Jermolene, is it correct that the link widget doesn't have to refresh anymore if $:/config/MissingLinks changes? because I removed that

Yes, that's right. The set widget in the page template will ensure that the entire page gets refreshed if $:/config/MissingLinks changes

BurningTreeC:
Perfect, thanks!

Jermolene:
Whoops, I just started updating the docs and realised that we may have inadvertently flipped the logic of the $:/config/MissingLinks variable. Before this PR, setting that config tiddler to "no" would cause missing links to be hidden. Afterwards, setting it to "yes" will cause missing links to be hidden.

So, we need to change the variable to "tv-show-missing-links", and default it to "yes" if its not defined.

BurningTreeC:
Ups, my mistake! I didn't think about it at all

PR #3531, opened 2018-11-10
use stateTitle stateField setTitle ...

BurningTreeC:
this replaces `state=...` , `set=...` and `popup=...` where tiddler titles are used for the states with

- `stateTitle=...`
- `setTitle=...`
- `popupTitle=...`

I've grepped for the affected files and I hope I got them all:

- toc macros
- tree macros
- Tag Manager
- folded-state in body-viewtemplate
- import listings (preview dropdowns)


Jermolene:
Great stuff @BurningTreeC thank you.

PR #3532, opened 2018-11-10
Tree macro double quotes fix dev2

morosanuae:
* Adding the 2nd batch of fixes
* Remove a lot of redundant code (hope I didn't break anything)
* Set a default value for the "prefix" parameter ($:/)
* Brushing the code a little bit (indentation)

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -1,13 +1,18 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
+\define link(full-title, chunk)
+<$link to=<<__full-title__>>>
+<$text text=<<__chunk__>>/>
+</$link>
+\end
+
 \define leaf-node(prefix)
 <li>
-<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]] +[removeprefix[$prefix$]] +[limit[1]]""" 
-emptyMessage="""<$text text="$prefix$$(chunk)$"/>""">
-<span>{{$:/core/images/file}}</span> <$link to="""$prefix$$(chunk)$""">
-<$view field="title"/>
-</$link> 
+<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]]""" variable="full-title">
+<$list filter="""[<full-title>] +[removeprefix[$prefix$]]""" variable="chunk">
+<span>{{$:/core/images/file}}</span> <$macrocall $name="link" full-title=<<full-title>> chunk=<<chunk>>/>
+</$list>
 </$list>
 </li>
 \end
@@ -34,10 +39,10 @@ emptyMessage="""<$text text="$prefix$$(chunk)$"/>""">
 
 \define tree-node(prefix)
 <ol>
-<$list filter="[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[!suffix[/]]" variable="chunk">
+<$list filter="""[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[!suffix[/]]""" variable="chunk">
 <<leaf-node """$prefix$""">>
 </$list>
-<$list filter="[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[suffix[/]]" variable="chunk">
+<$list filter="""[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[suffix[/]]""" variable="chunk">
 <<branch-node """$prefix$""">>
 </$list>
 </ol>
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -7,52 +7,56 @@ tags: $:/tags/Macro
 </$link>
 \end
 
-\define leaf-node(prefix)
+\define leaf-node(prefix, chunk)
 <li>
-<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]]""" variable="full-title">
-<$list filter="""[<full-title>] +[removeprefix[$prefix$]]""" variable="chunk">
-<span>{{$:/core/images/file}}</span> <$macrocall $name="link" full-title=<<full-title>> chunk=<<chunk>>/>
+<$list filter="""[[$prefix$$chunk$]is[shadow]] [[$prefix$$chunk$]is[tiddler]]""" variable="full-title">
+<$list filter="""[<full-title>] +[removeprefix<__prefix__>]""" variable="chunk">
+<span>{{$:/core/images/file}}</span>
+<$macrocall $name="link" full-title=<<full-title>> chunk=<<chunk>>/>
 </$list>
 </$list>
 </li>
 \end
 
-\define branch-node(prefix)
-<$reveal type="nomatch" text="" default="""$(chunk)$""">
-<li>      
-<$list filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]] +[limit[1]]""">
-<$reveal type="nomatch" state="""$:/state/tree/$prefix$$(chunk)$""" text="show">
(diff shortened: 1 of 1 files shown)


morosanuae:
@Jermolene I tried to follow the references you pointed an do the same kind of refactoring. I tried to get rid of the `$prefix$$chunk$` construct also but it was more difficult than I initially thought. So please let me know what you think about this second version after you get the chance to investigate the code in depth. Thanks.

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,62 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
-\define leaf-node(prefix)
+\define link(full-title, chunk)
```
BurningTreeC: Hi @morosanuae , maybe it's better to use a more unique macro-name because `<<link>>` would be available everywhere because of the system tag and may be used for other things by users?

morosanuae: You're right, I'll do that. Thanks @BurningTreeC.

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,62 @@
...
+\end
+
+\define leaf-node(prefix, chunk)
 <li>
-<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]] +[removeprefix[$prefix$]] +[limit[1]]""" 
-emptyMessage="""<$text text="$prefix$$(chunk)$"/>""">
-<span>{{$:/core/images/file}}</span> <$link to="""$prefix$$(chunk)$""">
-<$view field="title"/>
-</$link> 
+<$list filter="""[[$prefix$$chunk$]is[shadow]] [[$prefix$$chunk$]is[tiddler]]""" variable="full-title">
```
BurningTreeC: I believe here we should use `<$list filter="[<__prefix__>addsuffix<__chunk__>is[shadow]] ...>`

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,62 @@
...
+<$button set="""$:/state/tree/$prefix$$chunk$""" setTo="hide" class="tc-btn-invisible">
+{{$:/core/images/folder}} <$text text=<<__chunk__>>/>
+</$button>
 </$reveal>
-</$list>
-<$list filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]] +[limit[1]]"""><span>(<$count filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]]"""/>)</span>
-<$reveal type="match" state="""$:/state/tree/$prefix$$(chunk)$""" text="show">
-<$macrocall $name="tree-node" prefix="""$prefix$$(chunk)$"""/>
+<span>
+(<$count filter="""[all[shadows+tiddlers]prefix[$prefix$$chunk$]] -[[$prefix$$chunk$]]"""/>)
```
BurningTreeC: this should be more stable using `filter="[all[shadows+tiddlers]removeprefix<__prefix__>removeprefix<__chunk__>] -[<__prefix__>addsuffix<__chunk__>]`

morosanuae: Got it. I overlooked the "addsuffix" operator because I never used it before.

morosanuae: I'll wait until you finished reviewing the code and then I'll do the changes.

morosanuae: Thanks a lot.

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,62 @@
...
 
-\define branch-node(prefix)
-<$reveal type="nomatch" text="" default="""$(chunk)$""">
-<li>      
-<$list filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]] +[limit[1]]""">
-<$reveal type="nomatch" state="""$:/state/tree/$prefix$$(chunk)$""" text="show">
-<$button set="""$:/state/tree/$prefix$$(chunk)$""" setTo="show" class="tc-btn-invisible">{{$:/core/images/folder}} <$text text="""$(chunk)$"""/></$button>
+\define branch-node(prefix, chunk)
+<li>
+<$reveal type="nomatch" state="""$:/state/tree/$prefix$$chunk$""" text="show">
```
BurningTreeC: here we could use `state={{{ [[$:/state/tree/]addsuffix<__prefix__>addsuffix<__chunk__>] }}}` or we define the state with a `vars` widget here above like `<$vars stateName={{{ [[$:/state/tree/]addsuffix<__prefix__>addsuffix<__chunk__>] }}}>` so that we can reuse it below?

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,62 @@
...
 </$reveal>
-</$list>
-<$list filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]] +[limit[1]]"""><span>(<$count filter="""[all[shadows+tiddlers]prefix[$prefix$$(chunk)$]] -[[$prefix$$(chunk)$]]"""/>)</span>
-<$reveal type="match" state="""$:/state/tree/$prefix$$(chunk)$""" text="show">
-<$macrocall $name="tree-node" prefix="""$prefix$$(chunk)$"""/>
+<span>
+(<$count filter="""[all[shadows+tiddlers]prefix[$prefix$$chunk$]] -[[$prefix$$chunk$]]"""/>)
+</span>
+<$reveal type="match" state="""$:/state/tree/$prefix$$chunk$""" text="show">
+<$macrocall $name="tree-node" prefix="""$prefix$$chunk$"""/>
```
BurningTreeC: here `prefix={{{ [<__prefix__>addsuffix<__chunk__>] }}}`

morosanuae: I have never used filtered transclusions before either so I'm not used to it.

BurningTreeC:
@morosanuae , I'm done reviewing, thanks for your work!

for the button and reveal widgets I have an open PR: #3529
it introduces `stateTitle` and `setTitle` attributes which use tiddler titles like `$:/state/look!!interesting` like full titles, not like a reference to the field `interesting` of the tiddler `$:/state/look`. That would apply also here but should be added in a separate PR if mine gets merged or we have an alternative

core/wiki/macros/tree.tid
BurningTreeC: it's very useful, once you get it you'll use it all over the place. it's like editing a variable in-place
here it's a workaround for the `quotes` problem. this accepts `$:/""""""""` for example

core/wiki/macros/tree.tid
BurningTreeC: @morosanuae sorry, I've spotted another thing:

here you've removed `+[removeprefix[$prefix$]] +[limit[1]]` and the `emptyMessage`

they are for cases where we have a tiddler like `$:/state/something/something-else/another-thing` but there are no tiddlers `$:/state/something` or `$:/state/something/something-else`

removing the check and the emptyMessage causes `$:/state/something/something-else/another-thing` not to be displayed I think

BurningTreeC: the same below where the other emptyMessage was

morosanuae: I will do some tests. Thanks

BurningTreeC: yes, I may be totally wrong here

morosanuae: I gave some thought about it and it looks like this situation is not possible in my opinion.

morosanuae: The code suffered some extensive modifications throughout the time and the overall logic changed. The redundant code is from @Jermolene initial version of the macro.

morosanuae: Anyway if you spot something else please tell me because I want to publish the new version soon. Thanks.

BurningTreeC: thanks, I see :+1:

BurningTreeC: >Anyway if you spot something else please tell me because I want to publish the new version soon. Thanks.

I'm done spotting

morosanuae: :)

morosanuae:
@BurningTreeC I'll publish the modifications in the current form and if your proposal will be accepted it will not be a problem for me to publish another version incorporating "cutting edge" development. Anyway I have learned some very useful things from you today. Thanks again.

PR #3533, opened 2018-11-10
tree macro - fix for double quotes problem

morosanuae:
* Adding the 2nd batch of fixes (including @BurningTreeC suggestions)
* Remove a lot of redundant code (hope I didn't break anything)
* Brushing the code a little bit (indentation)

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -1,53 +1,64 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
-\define leaf-node(prefix)
+\define leaf-link(full-title, chunk)
+<$link to=<<__full-title__>>>
+<$text text=<<__chunk__>>/>
+</$link>
+\end
+
+\define leaf-node(prefix, chunk)
 <li>
-<$list filter="""[[$prefix$$(chunk)$]is[shadow]] [[$prefix$$(chunk)$]is[tiddler]] +[removeprefix[$prefix$]] +[limit[1]]""" 
-emptyMessage="""<$text text="$prefix$$(chunk)$"/>""">
-<span>{{$:/core/images/file}}</span> <$link to="""$prefix$$(chunk)$""">
-<$view field="title"/>
-</$link> 
+<$list filter="[<__prefix__>addsuffix<__chunk__>is[shadow]] [<__prefix__>addsuffix<__chunk__>is[tiddler]]" variable="full-title">
+<$list filter="[<full-title>] +[removeprefix<__prefix__>]" variable="chunk">
+<span>{{$:/core/images/file}}</span>
+<$macrocall $name="leaf-link" full-title=<<full-title>> chunk=<<chunk>>/>
+</$list>
 </$list>
 </li>
 \end
 
-\define branch-node(prefix)
-<$reveal type="nomatch" text="" default="""$(chunk)$""">
-<li>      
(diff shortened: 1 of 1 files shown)


morosanuae:
@Jermolene I think (and hope) this is the final version that can be merged, thanks to @BurningTreeC. You  have the final word on it.

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,64 @@
...
+<$list filter="[all[shadows+tiddlers]] +[removeprefix<__prefix__>] +[splitbefore[/]] +[sort[]] +[suffix[/]]" variable="chunk">
+<$macrocall $name="branch-node" prefix=<<__prefix__>> chunk=<<chunk>>/>
 </$list>
 </ol>
 \end
 
-\define tree(prefix)
+\define tree(prefix: "$:/")
 <div class="tc-tree">
-<span><$text text="""$prefix$"""/></span>
```
Jermolene: Hi @morosanuae I'm on mobile and so can't double check, but I think that the lack of whitespace inside this span was deliberate. With the newlines introduced the text nodes of the tree will gain a newline character, too. It won't be visible in the rendered DOM, but it is significant for JS code that looks directly at the DOM.

(A scenario where this matters is when generating a tree within a static HTML file with JS code for dynamically expanding/collapsing nodes).

core/wiki/macros/tree.tid
```diff
@@ -1,53 +1,64 @@
...
 \define tree-node(prefix)
 <ol>
-<$list filter="[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[!suffix[/]]" variable="chunk">
-<<leaf-node """$prefix$""">>
+<$list filter="[all[shadows+tiddlers]] +[removeprefix<__prefix__>] +[splitbefore[/]] +[sort[]] +[!suffix[/]]" variable="chunk">
+<$macrocall $name="leaf-node" prefix=<<__prefix__>> chunk=<<chunk>>/>
 </$list>
-<$list filter="[all[shadows+tiddlers]removeprefix[$prefix$]splitbefore[/]sort[title]] +[suffix[/]]" variable="chunk">
-<<branch-node """$prefix$""">>
+<$list filter="[all[shadows+tiddlers]] +[removeprefix<__prefix__>] +[splitbefore[/]] +[sort[]] +[suffix[/]]" variable="chunk">
```
Jermolene: Here you've changed `[all[shadows+tiddlers]removeprefix[...` with `[all[shadows+tiddlers]] +[removeprefix[...`. I would expect the former to be marginally faster, but I wondered if there was a reason for the change?

Jermolene: Hi @morosanuae I think it would be more consistent with the rest of the core to only use the `+` prefix when explicitly merging multiple filter runs.

core/wiki/macros/tree.tid
Jermolene: Hi @morosanuae of course, I forgot that you created the original.

(The background is that any text in between HTML elements/widgets is put into text nodes, and that includes whitespace, so the version with newlines is actually equivalent to `<span><$text text=<<new-line-character>>/><$text text-"""$prefix$"""/><$text text=<<new-line-character>>/></span>`. As I say, the newline characters will just appear as whitespace, not as an actual line break, and so the effect is seldom noticeable).

I would recommend changing it to the single line form because it does give neater output.

morosanuae: @Jermolene ok, thanks for explaining. I'll close this one and republish it after making the changes.

Issue #196, opened 2013-11-04
Renaming a tiddler break links, the user should be warned at least


mildred:
Renaming a tiddler break links. However, no hint about that is shown to the user when the title is modified.

Possible solutions in increasing order of complexity:
- explain what happens if the tiddler name is modified (links will be broken), discourage renaming tiddlers.
- display the list of backlinks that will be broken
- modify the linking tiddlers so the links are not brokens, take care of sub-tiddlers as well if there is any.

edrex:
@Jermolene, [$:/core/modules/widgets/navigator.js](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/navigator.js) handles `tm-rename-tiddler` events by calling [`$tw.wiki.renameTiddler`](https://github.com/Jermolene/TiddlyWiki5/blob/72941f3e9a916948222ff7be540d896e2b1901f6/core/modules/wiki-bulkops.js#L18) which updates tags and other references immediately, but it's not triggered when the title is changed in tiddler edit mode. Is that planned? 

I ask because I want to port [mGSD](https://github.com/simonbaird/mGSD)'s [RenameTagsPlugin](https://github.com/simonbaird/MPTW/blob/master/core/RenameTagsPlugin.js) UX (trigger reference updates on any rename, after prompting user) [to gsd5](https://github.com/roma0104/gsd5/issues/76) [1], and if that were in core it would at least give me a pattern to follow.

The major difficulty is getting code to run when a rename happens. I can see a few possibilities:

* `$tw.hooks.addHook("th-saving-tiddler",function(tiddler) {...` This is what I tried first, but the old title isn't available here.
* Replace `$tw.Wiki.addTiddler` with a wrapped version that detects renames and fires an event, I guess via a `wikimethod` module.
* Do something at the widget level with the title field on the edit form. I'm guessing this is more idiomatic TW5, although I'm not sure (and I don't know how to go about it).

IDK TW5 at all, just fumbling around in the dark here. Any advice greatly appreciated. Generally it seems like there should be more tiddler lifecycle hooks, but probably I'm missing something.

[1] gsd5 references aren't tags like mGSD's but regular fields, so `$tw.wiki.renameTiddlers` wouldn't work verbatim. I would probably register another listener for `tm-rename-tiddler` to do the job.

cc @roma0104 @simonbaird

roma0104:
Thanks for the feedback both here and on GSD5.
[...]

Issue #3288, opened 2018-05-16
Add support for search and replace


Jermolene:
The core should directly support search and replace. Desirable features include:

* The ability for the user to review the changes as diffs, and individually accept or reject them
* The ability to restrict the search and replace operation to a specific field, or across all fields

A significant part of the user interface functionality is implemented now that we have the diffing engine. It's also worth noting that a lot of the existing importing mechanism can actually be reused. For example:

1. Navigate to https://tiddlywiki.com/#%24%3A%2FSearchAndReplace
2. Open the browser JavaScript console and execute the following snippet:

```
$tw.wiki.addTiddler({title: "$:/SearchAndReplace", "plugin-type": "import",status: "pending",type: "application/json",text: JSON.stringify({tiddlers:{HelloThere:{title:"HelloThere",text:"Have you never had the feeling that your head is not big enough to hold everything you remember"}}})})
```

3. In the `$:/SearchAndReplace` tiddler, click the disclosure triangle to reveal the preview of the incoming tiddler, and switch the preview type to "Diff"

The result is pretty much what we need for the second part of the search and replace UI: a way for the user to review the proposed changes and to accept or reject them individually. The text still talks of "importing", of course.

<img width="908" alt="screenshot 2018-05-16 18 34 53" src="https://user-images.githubusercontent.com/174761/40133487-dc7395b8-5937-11e8-8876-136df4fcbd4d.png">

pmario:
That's an insteresting approach.

TiddlyTweeter:
> The core should directly support search and replace. Desirable features include:
> -   The ability for the user to review the changes as diffs, and individually accept r reject them
> -   The ability to restrict the search and replace operation to a specific field, or across all fields

In total agreement. Clearly for something so radical review is necessary before accepting the replace. But the scope covering potentially all fields on all Tiddlers would be very useful. At the moment I use external tools to do it.  I would rather not because they lack the intelligent context that TW itself would give more appropriately.

bimlas:
This would be useful for tiddler renaming - though new users might think it's too techie, but in any case it's better than not having the option to update backlinks. However, in the case of renaming, the text and the fields must also be seen, so in this case the two views should be combined.

I could imagine the apply of replace as clicking on the hunk in diff (red / green text).

As I see, `$:/core/modules/widgets/diff-text.js` is responsible for comparison (if someone wants to do it, do not have to look for it).

twMat:
[...]

PR #3534, opened 2018-11-11
List backlinks on rename

bimlas:
Currently, the user is **not warned that the links will not be modified** and the related tiddlers are "difficult" to find (you have to open Advanced Search, or exit from edit mode and look at Info -> References.

image

This PR tries to solve it (partially solves #196 too).

---

### 4c01a811d8a7bb8aaf22d4cb2e2b25386bd7e613 opens up Advanced Search to find backlinks:

image
...
image

---

### 6a09661b121226d370577472854e9b5d2769cab7 listing the references inplace:

image

---

I prefer the latter one, but I'm not sure that it suits for you as well.

### DIFF
--- a/core/language/en-GB/EditTemplate.multids
+++ b/core/language/en-GB/EditTemplate.multids
@@ -21,7 +21,8 @@ Tags/Dropdown/Caption: tag list
 Tags/Dropdown/Hint: Show tag list
 Title/BadCharacterWarning: Warning: avoid using any of the characters <<bad-chars>> in tiddler titles
 Title/Exists/Prompt: Target tiddler already exists
-Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers, backlinks in <<backlinksLink>> do not change
+Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers
+Title/References/Prompt: The following manual links to this tiddler will not be automatically updated
 Type/Dropdown/Caption: content type list
 Type/Dropdown/Hint: Show content type list
 Type/Delete/Caption: delete content type
--- a/core/language/en-GB/EditTemplate.multids
+++ b/core/language/en-GB/EditTemplate.multids
@@ -21,7 +21,7 @@ Tags/Dropdown/Caption: tag list
 Tags/Dropdown/Hint: Show tag list
 Title/BadCharacterWarning: Warning: avoid using any of the characters <<bad-chars>> in tiddler titles
 Title/Exists/Prompt: Target tiddler already exists
-Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers
+Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers, backlinks in <<backlinksLink>> do not change
 Type/Dropdown/Caption: content type list
 Type/Dropdown/Hint: Show content type list
 Type/Delete/Caption: delete content type
--- a/core/ui/EditTemplate/title.tid
+++ b/core/ui/EditTemplate/title.tid
@@ -2,12 +2,17 @@ title: $:/core/ui/EditTemplate/title
 tags: $:/tags/EditTemplate
 
 \define backlinksLink()
-<$button class="tc-btn-invisible tc-tiddlylink">
-<$action-setfield $tiddler="$:/temp/advancedsearch" text="""[[$(fromTitle)$]backlinks[]sort[title]]"""/>
-<$action-setfield $tiddler="$:/temp/search" text=""/>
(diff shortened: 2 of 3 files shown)


BurningTreeC:
Hi @bimlas , this is very much needed ...

could we make something with [this proposal ](https://github.com/Jermolene/TiddlyWiki5/issues/3288) by @Jermolene?

bimlas:
For the time being this seems too big, but it can be extended later on with this option as well.

BurningTreeC:
>For the time being this seems too big, but it can be extended later on with this option as well.

yes you're right

bimlas:
I hope you will like the latest commit: I modified the reference list template to accept another (`operandTitle`) title besides to the current tiddler to prevent code duplication.

core/language/en-GB/EditTemplate.multids
```diff
@@ -21,7 +21,7 @@ Tags/Dropdown/Caption: tag list
 Tags/Dropdown/Hint: Show tag list
 Title/BadCharacterWarning: Warning: avoid using any of the characters <<bad-chars>> in tiddler titles
 Title/Exists/Prompt: Target tiddler already exists
-Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers
+Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers, backlinks in <<backlinksLink>> do not change
```
Jermolene: Rather than re-using the "Relink" prompt, I think we should add a separate block of text with the warning about references to this tiddler, and therefore a separate translatable string. It might be called `Title/References/Prompt`, and contain the text `The following manual links to this tiddler will not be automatically updated:`.

We don't need the `backlinksLink` macro if we instead directly populate the listing (see below).

core/ui/EditTemplate/title.tid
```diff
@@ -35,10 +49,14 @@ tags: $:/tags/EditTemplate
 
 <$checkbox tiddler="$:/config/RelinkOnRename" field="text" checked="yes" unchecked="no" default="no"> {{$:/language/EditTemplate/Title/Relink/Prompt}}</$checkbox>
```
Jermolene: This is where we should include the new warning prompt string mentioned above. It should be preceded with a simple button showing the arrow icon, which won't need the transclusion of `<<lingo TiddlerInfo/References/Caption>>`.

core/ui/EditTemplate/title.tid
```diff
@@ -35,10 +49,14 @@ tags: $:/tags/EditTemplate
...
 
+<$reveal type="match" state="$:/state/showeditreferencing" text="show">
+<$vars operandTitle=<<fromTitle>> >
```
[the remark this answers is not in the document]
bimlas: :+1: I didn't knew about `<$tiddler>`, thanks!

Note that `current` of `<$tiddler>` is not works when transcluding is done via `{{$:/core/ui/TiddlerInfo/References}}`, but `<$transclude tiddler="$:/core/ui/TiddlerInfo/References"/>`: is this intended?

bimlas:
I hope this is how you imagined:

image

Should I handle the translations as well (`languages/ca-ES`, ...) by adding the `Title/References/Prompt` string in English?

core/ui/TiddlerInfo/References.tid
```diff
@@ -4,4 +4,4 @@ caption: {{$:/language/TiddlerInfo/References/Caption}}
 
 \define lingo-base() $:/language/TiddlerInfo/
 <$list filter="[all[current]backlinks[]sort[title]]" emptyMessage=<<lingo References/Empty>> template="$:/core/ui/ListItemTemplate">
-</$list>
+</$list>
```
bimlas: Leave a blank line at the end of the file or not? Both versions are used in source codes, not mentioned in style guide.

bimlas:
Ping @Jermolene : Please merge me!

core/ui/EditTemplate/title.tid
```diff
@@ -35,10 +35,22 @@ tags: $:/tags/EditTemplate
 
 <$checkbox tiddler="$:/config/RelinkOnRename" field="text" checked="yes" unchecked="no" default="no"> {{$:/language/EditTemplate/Title/Relink/Prompt}}</$checkbox>
 
+<$reveal type="nomatch" state="$:/state/showeditreferencing" text="show">
```
pmario: the state name could be more readable if it would be: `$:/state/show/references` or something similar. .. but I'm pretty sure we'll need to use `<<qualify..` .... 

Did you test the code if 2 tiddlers are open at the same time?

bimlas: As for the path of the state tiddler, you are right to be easier to read, but based on the `$:/state/showitpreview`, I would like to name it like it can be easily found. Anyway, I've modified it to `$:/state/edit/references`. I also applied the `qualify` macro.

core/ui/TiddlerInfo/References.tid
pmario: Doesn't really matter for TW. ... but some external editors may have a problem.

pmario: If you add several empty new-lines it may be a problem if you use tiddlers as templates or with transclusion.

Jermolene:
Great stuff, thank you @bimlas.

PR #3536, opened 2018-11-13
Dynaview: add element height to state tiddler

BurningTreeC:
this would allow us to reuse the height of a shown element for the hidden element

I tried it in the new dynaview demo replacing `style="min-height:5em"` with `style="height: $(height-in-statetiddler)$` and scrolling up/down becomes smooth

### DIFF
--- a/editions/dynaviewdemo/tiddlers/ViewTemplate.tid
+++ b/editions/dynaviewdemo/tiddlers/ViewTemplate.tid
@@ -7,6 +7,9 @@ tc-tiddler-frame tc-tiddler-view-frame $(missingTiddlerClass)$ $(shadowTiddlerCl
 \define folded-state()
 $:/state/folded/$(currentTiddler)$
 \end
+\define hidden-styles()
+height: $(tiddlerHeight)$
+\end
 <$vars storyTiddler=<<currentTiddler>> tiddlerInfoState=<<qualify "$:/state/popup/tiddler-info">> userClass={{!!class}}>
 <$tiddler tiddler=<<currentTiddler>>>
 <div data-tiddler-title=<<currentTiddler>> data-tags={{!!tags}} class=<<frame-classes>>>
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @BurningTreeC interesting! Doesn't that mean that all the tiddlers will start with a height of 0? And at that instant wouldn't they all be classes as being within the viewport? Perhaps you could include the changes to the demo in the PR? Many thanks.

BurningTreeC:
@Jermolene I tried reloading the wiki and it starts with a low amount of dynaview-state-tiddlers, `10 - 15` maybe. So I believe they don't collapse all within the viewport at wiki-start

BurningTreeC:
>It could probably be fixed so that the explicit height is only assigned when a placeholder is removed from view, and have it fall back to the min-height. In that way we'd be giving the vacated placeholders (so to speak) their previously rendered height

I thought I was doing so with 

```
tiddlerHeight={{{ [<state>has[height]get[height]addprefix[height:]] [<state>!has[height]addprefix[min-height:5em]removesuffix<state>] }}}
```

>You didn't mention which browser you're using but in my tests things behave very differently in Chrome from other browsers

that's right, I tried it in Firefox, not in Chrome

Jermolene:
Hi @BurningTreeC the problem that needs to be addressed is that the visibility tiddlers are initially missing, and so the height is left unset, defaulting to zero. Adding a `min-height` declaration would avoid that problem.

But, as I say, none of this helps us with the fundamental problem, that browsers other than Chrome will experience jumpy scrolling when elements are added or removed above the viewport. To fix that, we need to save the scroll position before a refresh, and restore it afterwards; the difficult part is that we'll need to save the scroll position relative to the currently viewed element, rather than the top of the page. Another aspect that is concerning is that I can't see a way to detect Chrome's behaviour vs. other browsers (short of browser sniffing).

Jermolene:
> The fix you mentioned sounds too complicated. Maybe firefox/safari will add something like that chrome-feature ... fingers crossed

The fix I described is pretty hellish, but I've long wanted to explore something similar for restoring focus after a refresh, so I may end up going down that rabbit hole...

Annoyingly, I remember reading Chrome's release note when this new feature came in, but I can't find it now, and Google isn't showing up anyone interested in the problem...

BurningTreeC:
>The fix I described is pretty hellish, but I've long wanted to explore something similar for restoring focus after a refresh, so I may end up going down that rabbit hole...

@Jermolene, good luck, it would be a great feature! 

I've found some information here:

https://drafts.csswg.org/css-scroll-anchoring/

and here:

https://stackoverflow.com/questions/42947024/keeping-scroll-position-when-adding-elements-on-top-works-in-firefox-but-not-chr

apparently, the chrome behavior can switched off by

```
body {
overflow-anchor: none;
}
```

BurningTreeC:
I found another resource: https://github.com/WICG/ScrollAnchoring

Jermolene:

PR #3537, opened 2018-11-13
PageScroller: scroll in new windows

BurningTreeC:
this PR makes the pagescroller scroll in new windows, too

for tests, one can open the pagetemplate in a new window


Jermolene:

PR #3538, opened 2018-11-13
New window: add missing click listener for popups

BurningTreeC:
this adds a click listener to new windows which enables us to cancel popups by clicking


Jermolene:
Thanks @BurningTreeC

PR #3539, opened 2018-11-14
Modals: Display in source-window

BurningTreeC:
this makes modals display within the window where they got opened, with the parameter `rootwindow` that, if `yes` or `true`, shows the modal always in the root TW window (`<$action-sendmessage $message="tm-modal" $param="mymodal" rootwindow="yes|true"/>`)

### DIFF
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -26,11 +26,11 @@ Display a modal dialogue
 Options include:
 	downloadLink: Text of a big download link to include
 */
-Modal.prototype.display = function(event,options) {
+Modal.prototype.display = function(title,options) {
 	options = options || {};
-	var title = event.param;
-	this.srcDocument = options.variables && (options.variables.rootwindow === "true" || options.variables.rootwindow === "yes") ? document :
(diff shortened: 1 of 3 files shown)


core/modules/utils/dom/modal.js
```diff
@@ -26,8 +26,12 @@ Display a modal dialogue
 Options include:
 	downloadLink: Text of a big download link to include
 */
-Modal.prototype.display = function(title,options) {
+Modal.prototype.display = function(event,options) {
```
Jermolene: We shouldn't change the type of an existing parameter. Perhaps we can use `options.event` instead?

BurningTreeC: @Jermolene , done!

BurningTreeC:
I'll add a note about the rootwindow attribute in the docs

Jermolene:
Thanks @BurningTreeC ping me when you're ready

BurningTreeC:
Hi @Jermolene , I've added docs about the new `rootwindow` attribute

Jermolene:
Thanks @BurningTreeC

PR #3541, opened 2018-11-14
Keyboard Widget: refresh keyInfoArray

BurningTreeC:
this makes a keyboard widget update its keyInfoArray if its key has the `((descriptor))` format and one of the platform-specific configuration tiddlers with that descriptor-suffix changes


Jermolene:
Thanks @BurningTreeC

PR #3542, opened 2018-11-15
fix action-navigate widget for undefined event

BurningTreeC:
we need to prevent the event from being undefined which happens when using `action-navigate` within a global keyboard shortcut


Jermolene:
Thanks @BurningTreeC

PR #3543, opened 2018-11-15
glob keyboard shtcts: pass event

BurningTreeC:
I figured we should pass the event here, too


BurningTreeC:
no, probably not... we don't want navigation to be suppressed when trying to trigger it with a keyboard shortcut

Issue #3379, opened 2018-08-07
[request] slugify operator


twMat:
Similar to the [encodeuri operator](https://tiddlywiki.com/#encodeuri%20Operator) it would be useful with something that converts a string into a **field name friendly** format - and - a **css identifier friendly** format. I.e:

`this-format : This is a field`

`.and-this-format { content:"This is a style definition" }`

I believe the character rules are very similar for these (very different) use areas so IMO it would be enough with a single general solution to cover both.

Why? Well, it would increase *hackability*; Currently, creating fields or css defs is a very hands-on matter. But, more importantly, it would be possible to integrate the creation of fields etc much more in the workflow, with field names potentially automatically generated from tiddler titles.

I understand if "you" have not come across any situations where you needed an "operator to create a field name" but ever since I realized that this *should* be possible I find more and more situations where it would really simplify things. Incidentally, there is over all surprisingly little discussion about custom fields and even CSS so I suspect both of these areas are way under-exploited in TW.

AnthonyMuscio:
twMat,

I support this idea because I have really started using custom fields of late, I would love to programaticaly add fields, with for example a number prefix or suffix. In mymenus plugin I could not easily automate the creation of menu-item-1 menu-item-2 .... fields as a new menu Item was added. I have also used menu-item-name-from-somewhere and removed the prefix "menu-item-" to get "name-from-somewhere" as the label in front of the fields value (transcluded), It would be nice if we can change that to "name from somewhere" or even "Name from somewhere". users could provide fieldnames from an input text field that are correctly encoded to save the value in.

Using prefix and suffix then add or remove prefix or suffix against fieldnames you can use the existence of a fieldname as a value in itself with the bonus the field also has a value that can be stored. 

I do think this could help hackability.

Regards
Tony

Jermolene:
@twMat are you proposing a new filter operator that converts an arbitrary string to a string of lower case letters, digits and dashes?

That operation is sometimes called to *slugify* a string. For example: https://blog.tersmitten.nl/slugify/

Good slugify routines also transliterate accented characters into their closest ASCII equivalent (eg `é` to `e` or `Œ` to `OE`).

One point to note is that most approaches do not guarantee uniqueness, with multiple strings mapping to the same slug (eg, `HélloThere` and `HelloThere` both mapping to `hellothere`).

[...]

PR #3546, opened 2018-11-17
add escapecss filter

BurningTreeC:
this filter would allow creating valid css classes from titles containing special characters

we assign a class to an element using `encodeuricomponent[]` so that the class name is encoded

in a stylesheet we create the classname by `<title>escapecss[]` which applies the uri encoding and escapes characters that need to be escaped

### DIFF
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -101,15 +101,8 @@ exports.escaperegexp = function(source,operator,options) {
 exports.escapecss = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		// escape any character with a special meaning in CSS if CSS.escape is available
-		var result;
-		try {
-			result = CSS.escape(title);
-		}
-		catch(e) {
-			result = title;
-		}
-		results.push(result);
+		// escape any character with a special meaning in CSS using CSS.escape()
+		results.push(CSS.escape(title));
 	});
 	return results;
 };
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -101,10 +101,8 @@ exports.escaperegexp = function(source,operator,options) {
 exports.escapecss = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		// encode the title
-		var escapedTitle = encodeURIComponent(title);
-		// escape any remaining character with a special meaning in CSS
-		results.push(escapedTitle.replace(/[\!\"\#\$\%\&\'\(\)\*\+\,\-\.\/\:\;\<\=\>\?\@\[\\\]\^\`\{\}\~]/g, '\\$&'));
+		// escape any character with a special meaning in CSS
+		results.push(title.replace(/[\!\"\#\$\%\&\'\(\)\*\+\,\-\.\/\:\;\<\=\>\?\@\[\\\]\^\`\{\}\~]/g, '\\$&'));
 	});
 	return results;
 };
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -101,8 +101,15 @@ exports.escaperegexp = function(source,operator,options) {
 exports.escapecss = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		// escape any character with a special meaning in CSS
-		results.push($tw.utils.escapeCssSelector(title));
+		// escape any character with a special meaning in CSS if CSS.escape is available
+		var result;
+		try {
+			result = CSS.escape(title);
+		}
+		catch(e) {
+			result = title;
+		}
+		results.push(result);
 	});
 	return results;
 };
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -98,4 +98,15 @@ exports.escaperegexp = function(source,operator,options) {
 	return results;
 };
 
+exports.escapecss = function(source,operator,options) {
+	var results = [];
+	source(function(tiddler,title) {
+		// encode the title
+		var escapedTitle = encodeURIComponent(title);
+		// escape any remaining character with a special meaning in CSS
+		results.push(escapedTitle.replace(/[\!\"\#\$\%\&\'\(\)\*\+\,\-\.\/\:\;\<\=\>\?\@\[\\\]\^\`\{\}\~]/g, '\\$&'));
+	});
+	return results;
+};
+
 })();
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -102,7 +102,7 @@ exports.escapecss = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
 		// escape any character with a special meaning in CSS
-		results.push(title.replace(/[\!\"\#\$\%\&\'\(\)\*\+\,\-\.\/\:\;\<\=\>\?\@\[\\\]\^\`\{\}\~]/g, '\\$&'));
+		results.push($tw.utils.escapeCssSelector(title));
 	});
 	return results;
 };
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -780,15 +780,19 @@ exports.makeDataUri = function(text,type) {
 	return parts.join("");
 };
 
+exports.escapeCssSelector = function(selector) {
+	return selector.replace(/[!"#$%&'()*+,\-./:;<=>?@[\\\]^`{\|}~,]/mg,function(c) {
+		return "\\" + c;
+	});
+};
+
 /*
 Useful for finding out the fully escaped CSS selector equivalent to a given tag. For example:
 
 $tw.utils.tagToCssSelector("$:/tags/Stylesheet") --> tc-tagged-\%24\%3A\%2Ftags\%2FStylesheet
 */
 exports.tagToCssSelector = function(tagName) {
-	return "tc-tagged-" + encodeURIComponent(tagName).replace(/[!"#$%&'()*+,\-./:;<=>?@[\\\]^`{\|}~,]/mg,function(c) {
-		return "\\" + c;
-	});
+	return $tw.utils.escapeCssSelector("tc-tagged-" + encodeURIComponent(tagName));
 };
 
 /*
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -780,19 +780,15 @@ exports.makeDataUri = function(text,type) {
 	return parts.join("");
 };
 
-exports.escapeCssSelector = function(selector) {
-	return selector.replace(/[!"#$%&'()*+,\-./:;<=>?@[\\\]^`{\|}~,]/mg,function(c) {
-		return "\\" + c;
-	});
-};
-
 /*
 Useful for finding out the fully escaped CSS selector equivalent to a given tag. For example:
 
 $tw.utils.tagToCssSelector("$:/tags/Stylesheet") --> tc-tagged-\%24\%3A\%2Ftags\%2FStylesheet
 */
 exports.tagToCssSelector = function(tagName) {
-	return $tw.utils.escapeCssSelector("tc-tagged-" + encodeURIComponent(tagName));
+	return "tc-tagged-" + encodeURIComponent(tagName).replace(/[!"#$%&'()*+,\-./:;<=>?@[\\\]^`{\|}~,]/mg,function(c) {
+		return "\\" + c;
+	});
 };
 
 /*


BurningTreeC:
Thanks @pmario , yes it's more complicated. Still, I cannot use any kind of title with `data-tiddler-title`. This proposal would allow that

Jermolene:
By the way @BurningTreeC there is [already a utility method](
https://tiddlywiki.com/#How%20to%20apply%20custom%20styles%20by%20tag) for doing the conversion: `$tw.utils.tagToCssSelector()`

I take your point @pmario, but I'd be inclined to add this; we can mark the new operator as "deprecated" from the beginning.

BurningTreeC:
@Jermolene , that utility function would need to be changed a bit... should another one be made that adds a prefix only if there`s a prefix passed to it? The filter operand could be used as prefix

pmario:
>Thanks @pmario , yes it's more complicated. Still, I cannot use any kind of title with data-tiddler-title. This proposal would allow that

I'm pretty sure, that's not true. ... Can you describe what you want to achieve. ... With the new system it's possible to create dynamic CSS files, similar to your proposal, but without new code.

pmario:
see: https://tiddlywiki.com/#Custom%20Styles%20FAQ:%5B%5BCustom%20Styles%20FAQ%5D%5D%20%5B%5BQ%3A%20Is%20there%20a%20way%20to%20create%20dynamic%20stylesheets%3F%5D%5D%20%5B%5BQ%3A%20How%20can%20I%20use%20a%20custom%20field%20to%20style%20a%20tiddler%3F%5D%5D%20%5B%5BHow%20to%20apply%20custom%20styles%5D%5D

BurningTreeC:
>I'm pretty sure, that's not true. ... Can you describe what you want to achieve. ... With the new system it's possible to create dynamic CSS files, similar to your proposal, but without new code.

@pmario, on tiddlywiki.com there's a good example-tiddler: `"Moments: A little color won't hurt" by Riz`

I'd like to use the title of a tiddler like this to assign a style to it

how can I access it with `[data-tiddler-title=""Moments: A little color won't hurt" by Riz"]` or `[data-tiddler-title='"Moments: A little color won't hurt" by Riz']` ? using double- or triple-quotes isn't valid... I could encode the title first and assign that encoded string to data-tiddler-title , but then we're back at this point here. How would you solve it?

The problem is, that this should not just work for a specific tiddler where we can look for a workaround using special selectors or tailor a special combination that works, it should simply work for any string a filter gives me as output..

pmario:
Use the `class` field instead of the title.

Jermolene:
> @Jermolene , that utility function would need to be changed a bit... should another one be made that adds a prefix only if there`s a prefix passed to it? The filter operand could be used as prefix

Probably best to refactor out a new `escapeCssSelector()` utility method from `tagToCssSelector()`, and then use the new `escapeCssSelector()` in your operator.

But, I'm inclined to agree with @pmario that if it's possible to achieve what you want using new features then we should avoid adding this filter to avoid encouraging use of a deprecated feature.

BurningTreeC:
@Jermolene , yes if it's possible to achieve I agree that we don't need it

@pmario, I'd need to use the tiddler title or create a `qualified` class that's unique for that tiddler - but putting it in a field for each tiddler is inconvenient.

Anyways, I can just use that filter for my own purposes, we don't need to add it. The solution of defining it as "deprecated" from the beginning is a bit weird, I think we should not do this

pmario:
I'm still not convinced. The example you pointed out is a problem, but the only thing for data-attributes would be to escape the quote like this: 

```
[data-tiddler-title^="\"Moments: A little color won't hurt\" by Riz"] {
  border: 1px solid green;
}
```

Which gives us a human readable / human friendly definition.

Jermolene:
Hi @pmario that actually suggests that we *do* need an `escapecss` filter operator, but that it should just do the escaping necessary for a simple CSS string (ie newlines as above, but presumably there's a way to escape non-printable characters, too).

pmario:
>Hi @pmario that actually suggests that we do need an escapecss filter operator, ...

probably yes, but a much simpler one.

pmario:
I'll have a closer look to the specs.

BurningTreeC:
>I'm still not convinced. The example you pointed out is a problem, but the only thing for data-attributes would be to escape the quote like this

@pmario yes escaping the necessary characters only without uri encoding would be good enough

pmario:
in short: attribute values _must_ be unicode chars. So replacing `"` -> `\"` should be the only thing we need. Every other char should be valid. ... Which was the main reason why we implemented it. 

relevant SPECs

 - https://drafts.csswg.org/selectors-4/#attribute-selectors
 - https://drafts.csswg.org/css-syntax-3/#typedef-ident-token
 - https://infra.spec.whatwg.org/#code-point

pmario:
IMO the `<$view` widget also needs a format like this one ?!

Jermolene:
> IMO the `<$view` widget also needs a format parameter like this one ?!

Adding new formats for the view widget is actually much less useful than adding a filter operator, because one is so restricted where the view widget can be used. We're quite close to being able to deprecate the view widget if we wanted to...

Jermolene:
> in short: attribute values must be unicode chars. So replacing " -> \" should be the only thing we need. Every other char should be valid. ... Which was the main reason why we implemented it.

Great, thanks @pmario.

The existing `stringify` operator does that, but also a few other conversions for non-printable characters, but perhaps it's usable?

BurningTreeC:
@Jermolene @pmario , I think using the `stringify` operator for this is fine. Thanks for discussing this with me!

BurningTreeC:
I'm reopening this PR because the `stringify` operator is not sufficient. For example it doesn't handle `ä ö ü` correctly (replaces `ä` with `\xE4`) ...

BurningTreeC:
I guess we need a filter that escapes all necessary characters `and` doesn't encode anything

pmario:
ö and ä are valid if data-attributes are used. .. So imo the only thing is `"`

BurningTreeC:
@pmario @Jermolene, perhaps this `CSS.escape` polyfill could be useful?

https://github.com/mathiasbynens/CSS.escape/blob/master/css.escape.js

Jermolene:
Thanks @BurningTreeC I hadn't come across CSS.escape, it does indeed make sense to use it. It's a shame it's so much more complex than expected...

pmario:
no ... IMO we don't need a CSS-escape here. everything unicode is valid except " which causes a problem with attribute-selectors.

BurningTreeC:
@pmario I did some tests with some very bad titles

- `"` needs to be escaped as you say
- `\` needs to be escaped
- `'` should be escaped, too - we can do: `[data-tiddler-title='my title with a ' single quote']`

but you're right that we don't need an excessive list of characters being escaped

yet I don't know about non-printable characters, I'll try to investigate them

BurningTreeC:
we could call the filter `escapedatatag` / `escapedatacss` / `escapecssdata` / `escapecssdatatag` :1st_place_medal:

Jermolene:
Looks good, have you resolved the concern about non-printable characters?

BurningTreeC:
@Jermolene - I tried copy-pasting some non-printable characters into titles and tiddler-content

- seems that they get removed from titles
- seems that in tiddler content they aren't rendered but their "string" is shown
- so at the moment I'm not concerned about them
- but my test-cases may be unappropriate, can anyone provide a concerning test-tiddler?

Jermolene:
Hi @BurningTreeC the paths by which non-printable characters might get introduced are subtle. For example, one might have a text field containing newlines, and then use an action widget to create a tiddler titled with that text. So I'd be inclined to find a way to test the filter with them even if one can't see a plausible scenario under which they'd be encountered; things like TW5 get used in ways that we don't expect.

BurningTreeC:
@Jermolene, thanks you're right. You're simple test-case just broke it

- create a tiddler `New Tiddler` containing text and a new line
- create a new tiddler with this button: `<$button><$action-setfield $tiddler={{New Tiddler}} text="test"/>test</$button>`
- try to assing some css to it using `data-tiddler-title`
- doesn't work when escaping just `"` `\` and `'`

BurningTreeC:
I would propose stripping them out from titles, whenever such characters get assigned to a title

The action-setfield widget should check if we're assigning to a title field and remove those characters before assigning to that field. The tiddlywiki core will replace such a tiddler title anyways after reloading a wiki (`hello\n` becomes `hello` ). We should also add a prompt in case the tiddler with the title without those characters exists and may get overwritten.

BurningTreeC:
In the case of this PR the problem starts with assigning the title that may contain new-line chars to the `data-tag` ... at that point already new-line characters should get replaced by their html entity ( https://stackoverflow.com/questions/16451553/css-data-attribute-new-line-character-pseudo-element-content-value )

I think avoiding all that hassle would be the best

Jermolene:
> The action-setfield widget should check if we're assigning to a title field and remove those characters before assigning to that field.

It's a bit of hornets nest. At the moment, the semantics of the tiddler store are very simple: a saveTiddler cannot fail, and the tiddler that is stored is always the same as the tiddler that was provided. Lots of code subtly relies on that.

Tackling these things at the UI level, though, is fine, and makes sense.

> The tiddlywiki core will replace such a tiddler title anyways after reloading a wiki (hello\n becomes hello ).

I think of that as one of many artefacts of the storage format we use in the single file version. I think we should consider moving to embedded JSON.

> We should also add a prompt in case the tiddler with the title without those characters exists and may get overwritten.

That's the kind of unexpected consequence of mangling tiddler titles that is apt to increase exponentially...

BurningTreeC:
>I think of that as one of many artefacts of the storage format we use in the single file version. I think we should consider moving to embedded JSON.

It does so in the nodejs version, too. After reloading, `hello\n` doesn't exist anymore, but `hello` does (Linux 64bit)

I cannot say much about the other points, I lack insights there. Maybe we should really try filtering the tiddler-title for those characters and replacing them by their html entity for assigning them to a data-tag? [see my comment above](https://github.com/Jermolene/TiddlyWiki5/pull/3546#issuecomment-441358489)

BurningTreeC:
... and I think we can use the same filter to create a state-title with the suffix of such a tiddler

Jermolene:
> It does so in the nodejs version, too. After reloading, hello\n doesn't exist anymore, but hello does (Linux 64bit)

The .tid file format doesn't support newlines except in body text.

> Maybe we should really try filtering the tiddler-title for those characters and replacing them by their html entity for assigning them to a data-tag?

Ideally, we'd find the most efficient escaping that was guaranteed not to produce any illegal characters. HTML entities are quite slow to process.

BurningTreeC:
Back to the escapecss filter: I've added the above mentioned `escapeCss` function from the polyfill to my `escapecss` filter and it handles the test-tiddler ( `hello +\n` ) correctly!

Jermolene:
> Back to the escapecss filter: I've added the above mentioned escapeCss function from the polyfill to my escapecss filter and it handles the test-tiddler ( hello +\n ) correctly!

Great. It's still a great shame that have to pull in such a big lump of JS for this, since all of this is rather about handling edge cases, but I can't see an alternative: for this to work properly we need to match the processing done by browsers.

BurningTreeC:
the escapeCSS from the polyfill also escapes characters that wouldn't need to be escaped for data-tags

if anyone could provide the "breaking" set of characters for this use-case, I look up how escapeCSS escapes them, we incorporate that escaping for that set of chars only and the filter should be done

pmario:
ok

BurningTreeC:
This is all the code from the escapeCSS polyfill that we need:

https://gist.github.com/BurningTreeC/c9c39e87563d31501bd378877e2dd0fe

BurningTreeC:
pulling the whole thing in would have the benefit of being able to escape class-selectors, too - if needed at some point

Jermolene:
Indeed. There's similar logic behind bringing the entire diff-patch-match library into the core.

BurningTreeC:
@Jermolene, would you want the code in `utils.js` as replacement for `escapeCssSelector` or should it be somewhere else or a separate function?

Jermolene:
> @Jermolene, would you want the code in utils.js as replacement for escapeCssSelector or should it be somewhere else or a separate function?

I think we should add it as a polyfill. I've tended to avoid adding them to the core in the past, so there's no particular infrastructure. We should probably add a "polyfill" module-type, and have one of the early startup modules execute them.

BurningTreeC:
@twMat , you're right, it's the same issue. I think `CSS.escape` is what works for both of us

If we only get this CSS.polyfill added correctly ...

Jermolene:
The slugify function that I posted over in #3379 is designed to produce human readable output, and has the property that many titles might map to the same slug. I think that approach only works for limited situations. Here, we're trying to replicate CSS's encodings so that we can compose CSS declarations to match specific strings.

BurningTreeC:
@Jermolene , I modified the escapecss filter to use CSS.escape , surrounded it by a `try{}catch(e)` which should catch the error if `CSS.escape` isn't present. A simple solution, not going the polyfill way, would now be adding the escapeCSS function to `utils.js` and fall back to it if CSS.escape throws an error in the filter

Jermolene:
The problem with that approach is that it puts a taken try/catch in the code path. They are expensive so we should only put exceptions on the code path in exceptions... What's the problem with the polyfill approach?

BurningTreeC:
>What's the problem with the polyfill approach?

I don't know how to implement it

BurningTreeC:
Hi @Jermolene , how do you want to proceed with this?

BurningTreeC:
This filter would still be useful

Jermolene:
Hi @BurningTreeC if I'm re-reading things correctly, we'd decided to use the `CSS.escape()` function, and include a polyfill in the core. There was some discussion about how to package the polyfill, but on reflection if we're including the polyfill code in the core then it makes just as much sense to go with your suggestion of adding an `escapeCSS` function to `utils.js` that tries to use `CSS.escape()` if available, falling back to the polyfill.

Jermolene:
Apologies, I hadn't seen #3574, let's go ahead and merge with the externalised polyfill.

Jermolene:
Hi @BurningTreeC please could you add some user docs in a separate PR? Many thanks.

BurningTreeC:
@Jermolene - sure, I'll add some docs the next days

PR #3548, opened 2018-11-18
tree macro - fix for double quotes (and other special characters)

morosanuae:
* Added fixes for special characters (e.g. double quotes) (including @BurningTreeC and @Jermolene suggestions)
* Removed a lot of redundant code

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -1,53 +1,57 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
-\define leaf-node(prefix)
+\define leaf-link(full-title, chunk)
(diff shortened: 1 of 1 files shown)


morosanuae:
@Jermolene I have made the adjustments but I noticed there are conflicts with #3531. Would you like me to publish a new version incorporating the new changes added by #3531?

Jermolene:
Thanks @morosanuae yes, you should be able to update your branch with the latest changes, rather than actually deleting and creating the PR, but please feel free to do whichever is easier!

morosanuae:
@Jermolene currently I'm at a very basic level with GitHub (and with Git also) and I don't have much time to learn it, so if it isn't a big problem I will recreate the PR for now.

PR #3549, opened 2018-11-18
Fix the TagManager

BurningTreeC:
now the tag manager needs some fixes for breaking titles, too

- the `<<__variable__>>` syntax needs to be used
- the `tag-pills` don't use the tag-pill macro, but `{{||$:/core/ui/TagTemplate}}` directly
- some already sufficiently qualified states in the `iconEditor` macro are reduced not to use the `qualify` macro anymore


Jermolene:

BurningTreeC:
thanks!

PR #3550, opened 2018-11-19
fix untagged-pill popup position

BurningTreeC:
this corrects the popup-position of the untagged-pill in the tag-manager, showing on top of the Tiddler, not below the pill


Jermolene:
Thanks @BurningTreeC

PR #3552, opened 2018-11-19
tree macro - fix for double quotes (and other special characters)

morosanuae:
* Added fixes for special characters (e.g. double quotes) (including @BurningTreeC and @Jermolene suggestions)
* Removed a lot of redundant code


morosanuae:
@Jermolene are we there yet?

Jermolene:
Thanks @morosanuae looks like the gods of GitHub have smiled upon you...

morosanuae:
Yep, I have prayed for mercy...

Issue #3547, opened 2018-11-17
RSOD - problem with prerelease - preview


pmario:
to reproduce: 

 - open https://tiddlywiki.com/prerelease/#%24%3A%2FControlPanel
 - enter edit mode
 - open preview for
 - parse-tree

This problem doesn't show in 5.1.17

BurningTreeC:
@pmario I can reproduce this, though I don't know the root of the problem

I was able to "fix" the error in `utils/fakedom.js` : 

```
TW_Element.prototype.appendChild = function(node) {
	this.children.push(node);
	try {
		node.parentNode = this;
	}
	catch(e) {}
};
```

... don't know how expensive this is

BurningTreeC:
> This problem doesn't show in 5.1.17

No, it shows in 5.1.17, too. Try the combination of CodeMirror + tools for exploring tiddlywiki internals
The CodeMirror engine.js causes the error when trying to append `cmDomNode` through `self.domNode.appendChild(cmDomNode);` (around line 112 in engine.js)

BurningTreeC:
I found another way to solve this...

line 110 in codemirror's engine.js: 

```
this.cm = window.CodeMirror(function(cmDomNode) {
		Object.defineProperty(cmDomNode,'parentNode',{writable: true});
               self.domNode.appendChild(cmDomNode);
},config);
```

Jermolene:
See my comment over on #3444: https://github.com/Jermolene/TiddlyWiki5/pull/3554#issuecomment-439984398

PR #3554, opened 2018-11-19
fix for #3547

BurningTreeC:
see https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

because of "strict mode" - the `parentNode` property may not be writable and causes an error in `fakedom.js` (appendChild)

to reproduce: install CodeMirror, install "Tools for exploring internals of TW", open a heavy Tiddler like the Control-Panel in edit-mode, toggle preview on, switch to "parse tree" or "widget tree"

### DIFF
--- a/plugins/tiddlywiki/codemirror/engine.js
+++ b/plugins/tiddlywiki/codemirror/engine.js
@@ -109,6 +109,7 @@ function CodeMirrorEngine(options) {
 	// Create the CodeMirror instance
 	this.cm = window.CodeMirror(function(cmDomNode) {
 		// Note that this is a synchronous callback that is called before the constructor returns
+		Object.defineProperty(cmDomNode,"parentNode",{writable: true});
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC it looks like CodeMirror.js is attempting to append a real DOM node into a fake dom. The reason why it's happening is because the "internals" plugin renders the tiddler to a fakedom in order to display the widget tree.

The CM plugin should check explicitly for the fake dom, as we do in e.g. the KaTeX plugin:

https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/katex/wrapper.js#L42

BurningTreeC:
thanks for clarifying, @Jermolene 

Now we're checking for the fake-dom and don't try to append the domnode in that case

Jermolene:
Thanks @BurningTreeC

PR #3559, opened 2018-11-20
fix #3555

### LINKED ISSUE #3555: $tw.wiki.initParsers is not a function (FF38)
in [the group](https://groups.google.com/d/msg/TiddlyWiki/htO5KckxEfw/6Ii6_KNdBgAJ) this error was reported on an old FireFox version (38.0.5) and I could reproduce it using that  version and opening the tiddlywiki prerelease

the window is stuck at the splash screen and yells `$tw.wiki.initParsers` is not a function

see the attached screens
19-11-2018-155653_shoot
19-11-2018-155718_shoot

BurningTreeC:
fix for #3555 - inner function declaration should be on top of its "host" function

error originally appeared on firefox v38.0.5

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -538,6 +538,45 @@ exports.findListingsOfTiddler = function(targetTitle,fieldName) {
 Sorts an array of tiddler titles according to an ordered list
 */
 exports.sortByList = function(array,listTitle) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @BurningTreeC

PR #3565, opened 2018-11-21
[UI] add selectable tags- and search-dropdown

BurningTreeC:
this adds a `saveTiddler` attribute to factory.js

if it's present it's used as a tiddler title where all the inputs/changes made by the user get stored
this can be used for selecting dropdown-entries that are filterable by an input field
the entries should be sorted by the `saveTiddler` and the original tiddler (`<$edit-text tiddler="originaltiddler" .../>`) can be updated with the selected dropdown-entry which will be shown within the input, without changing the filtering of the dropdown-items

<hr>

Update: to the above mentioned changes, this adds:

- a `refreshTiddler` attribute to factory.js (refreshes the input-text when that tiddler changes, sets it to whatever is within the `editTiddler`)
- a full-featured keyboard-selectable tags-dropdown

- up/down selects a dropdown tag
- enter adds the selected tag
- escape clears the input field, or - if empty - cancels editing the tiddler


Jermolene:
Hi @BurningTreeC this is intriguing, but I'm not sure I fully understand. Can you post a more complete example?

BurningTreeC:
Ok, @Jermolene here's a working tag-picker (with a missing refresh logic for the input field when the selected  tag changes):

```
\define add-tag-actions()
<$vars selectedTag={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]get[text]] }}}>
<$list filter="[[$(currentTiddler)$]!contains:tags<selectedTag>]" variable="ignore">
<$action-sendmessage $message="tm-add-tag" $param=<<selectedTag>>/>
</$list>
<$list filter="[[$(currentTiddler)$]contains:tags<selectedTag>]" variable="ignore">
<$action-sendmessage $message="tm-remove-tag" $param=<<selectedTag>>/>
</$list>
<$action-deletetiddler $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}}/>
</$vars>
\end

\define tag-button(selectedClass)
<$button class="tc-btn-invisible $selectedClass$" tag="a">
$(actions)$
<$action-deletetiddler $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}}/>
<$macrocall $name="tag-pill" tag=<<tag>>/>
</$button>
\end

\define get-tags-list()
$(filteredNonSystemTags)$ $(filteredSystemTags)$
\end

\define select-tag-actions(beforeafter,reverse)
<$vars userInput={{{ [[$(saveInputTiddler)$]get[text]] }}} selectedDropDownTag={{{ [[$(newTagNameTiddler)$]get[text]] }}}>
<$set name="filteredNonSystemTags" filter="[tags[]!is[system]search:title<userInput>sort[]]"> <$set name="filteredSystemTags" filter="[tags[]is[system]search:title<userInput>sort[]]">
<$set name="tagsList" filter=<<get-tags-list>>>
<$list filter="[enlist<tagsList>] +[$beforeafter$<selectedDropDownTag>] ~[enlist<tagsList>$reverse$nth[1]]" variable="nextTag">
<$action-setfield $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}} text=<<nextTag>>/>
</$list>
</$set>
</$set>
</$set>
</$vars>
\end

\define tag-picker(actions)
<$vars newTagNameTiddler={{{ [[$:/temp/NewTagName/]addsuffix<currentTiddler>] }}} saveInputTiddler={{{ [[$:/temp/NewTagName/Input/]addsuffix<currentTiddler>] }}} actions=<<__actions__>>>
<div class="tc-edit-add-tag">
<$keyboard key="Down" actions=<<select-tag-actions "after">>>
<$keyboard key="Up" actions=<<select-tag-actions "before" "reverse[]">>>
<$keyboard key="ENTER" actions=<<add-tag-actions>>>
<$edit-text tiddler=<<newTagNameTiddler>> tag="input" default="" placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" saveTiddler=<<saveInputTiddler>>/>
</$keyboard></$keyboard></$keyboard>
</span> <$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button> <span class="tc-add-tag-button">
<$set name="tag" value={{{ [<newTagNameTiddler>get[text]] }}}>
<$button setTitle=<<newTagNameTiddler>> setTo="" class="">
$actions$
<$action-deletetiddler $tiddler=<<newTagNameTiddler>>/>
{{$:/language/EditTemplate/Tags/Add/Button}}
</$button>
</$set>
</span>
</div>
<div class="tc-block-dropdown-wrapper">
<$vars userInput={{{ [<saveInputTiddler>get[text]] }}}>
<$reveal state=<<qualify "$:/state/popup/tags-auto-complete">> type="nomatch" text="" default="">
<div class="tc-block-dropdown">
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter="[tags[]!is[system]search:title<userInput>sort[]]" variable="tag">
<$list filter="[<newTagNameTiddler>get[text]removeprefix<tag>suffix[]]" emptyMessage="""<<tag-button>>""">
<<tag-button "tc-tag-button-selected">>
</$list>
</$list></$list>
<hr>
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter="[tags[]is[system]search:title<userInput>sort[]]" variable="tag">
<$list filter="[<newTagNameTiddler>get[text]removeprefix<tag>suffix[]]" emptyMessage="""<<tag-button>>""">
<<tag-button "tc-tag-button-selected">>
</$list>
</$list></$list>
</div>
</$reveal>
</$vars>
</div>
</$vars>
\end
```

to visualize the selected tag: 

```
.tc-tag-button-selected {
	background-color: <<colour primary>>;
}
```

Filtering the tags through the input writes 2 different tiddlers. Selecting tags with Up/Down sets the `editTiddler` to that tag-name (we need a refresh mechanism for that, so the selected tag is shown within the input). The dropdown is filtered by what's inside the second state-tiddler which isn't changed by selecting tags from the dropdown with Up/Down

BurningTreeC:
An updated tag-picker with auto-refresh and auto-focus:

```
\define add-tag-actions()
<$vars selectedTag={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]get[text]] }}}>
<$list filter="[[$(currentTiddler)$]!contains:tags<selectedTag>]" variable="ignore">
<$action-sendmessage $message="tm-add-tag" $param=<<selectedTag>>/>
</$list>
<$list filter="[[$(currentTiddler)$]contains:tags<selectedTag>]" variable="ignore">
<$action-sendmessage $message="tm-remove-tag" $param=<<selectedTag>>/>
</$list>
<$action-deletetiddler $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}}/>
</$vars>
\end

\define tag-button(selectedClass)
<$button class="tc-btn-invisible $selectedClass$" tag="a">
$(actions)$
<$action-deletetiddler $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}}/>
<$action-deletetiddler $tiddler={{{ [[$:/temp/NewTagName/Input/]addsuffix[$(currentTiddler)$]] }}}/>
<$macrocall $name="tag-pill" tag=<<tag>>/>
</$button>
\end

\define get-tags-list()
$(filteredNonSystemTags)$ $(filteredSystemTags)$
\end

\define select-tag-actions(beforeafter,reverse)
<$vars userInput={{{ [[$(saveInputTiddler)$]get[text]] }}} selectedDropDownTag={{{ [[$(newTagNameTiddler)$]get[text]] }}}>
<$set name="filteredNonSystemTags" filter="[tags[]!is[system]search:title<userInput>sort[]]"> <$set name="filteredSystemTags" filter="[tags[]is[system]search:title<userInput>sort[]]">
<$set name="tagsList" filter=<<get-tags-list>>>
<$list filter="[enlist<tagsList>] +[$beforeafter$<selectedDropDownTag>] ~[enlist<tagsList>$reverse$nth[1]]" variable="nextTag">
<$action-setfield $tiddler={{{ [[$:/temp/NewTagName/]addsuffix[$(currentTiddler)$]] }}} text=<<nextTag>>/>
<$action-setfield $tiddler="$:/temp/SelectedTag" text=<<nextTag>>/>
</$list>
</$set>
</$set>
</$set>
</$vars>
\end

\define tag-picker(actions)
<$vars refreshMe={{$:/temp/SelectedTag}}>
<$vars newTagNameTiddler={{{ [[$:/temp/NewTagName/]addsuffix<currentTiddler>] }}} saveInputTiddler={{{ [[$:/temp/NewTagName/Input/]addsuffix<currentTiddler>] }}} actions=<<__actions__>>>
<div class="tc-edit-add-tag">
<$keyboard key="Down" actions=<<select-tag-actions "after">>>
<$keyboard key="Up" actions=<<select-tag-actions "before" "reverse[]">>>
<$keyboard key="ENTER" actions=<<add-tag-actions>>>
<$edit-text tiddler=<<newTagNameTiddler>> tag="input" default="" placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" saveTiddler=<<saveInputTiddler>> focus={{{ [<newTagNameTiddler>!is[missing]addprefix[yes]removesuffix<newTagNameTiddler>] }}}/>
</$keyboard></$keyboard></$keyboard>
</span> <$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button> <span class="tc-add-tag-button">
<$set name="tag" value={{{ [<newTagNameTiddler>get[text]] }}}>
<$button setTitle=<<newTagNameTiddler>> setTo="" class="">
$actions$
<$action-deletetiddler $tiddler=<<newTagNameTiddler>>/>
{{$:/language/EditTemplate/Tags/Add/Button}}
</$button>
</$set>
</span>
</div>
<div class="tc-block-dropdown-wrapper">
<$vars userInput={{{ [<saveInputTiddler>get[text]] }}}>
<$reveal state=<<qualify "$:/state/popup/tags-auto-complete">> type="nomatch" text="" default="">
<div class="tc-block-dropdown">
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter="[tags[]!is[system]search:title<userInput>sort[]]" variable="tag">
<$list filter="[<newTagNameTiddler>get[text]removeprefix<tag>suffix[]]" emptyMessage="""<<tag-button>>""">
<<tag-button "tc-tag-button-selected">>
</$list>
</$list></$list>
<hr>
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter="[tags[]is[system]search:title<userInput>sort[]]" variable="tag">
<$list filter="[<newTagNameTiddler>get[text]removeprefix<tag>suffix[]]" emptyMessage="""<<tag-button>>""">
<<tag-button "tc-tag-button-selected">>
</$list>
</$list></$list>
</div>
</$reveal>
</$vars>
</div>
</$vars>
</$vars>
\end
```

BurningTreeC:
to the last example: I'd rather have a refresh mechanism within `factory.js` that refreshes the input-text and sets the focus back to the input than using a hack like here

I think about a `refreshTiddler` attribute that refreshes the input-text and sets focus to the input, when the refreshTiddler changes AND the input is focused

BurningTreeC:
@Jermolene - I've updated the first comment with the additions made

BurningTreeC:
we could refactor the logic for dropdown-selection into a macro, so that we can reuse it for the search-dropdown, fieldname-dropdown etc

BurningTreeC:
For cases like the `Dynaview demo` we'll probably want to add an `exclude` filter

pmario:
do you have a minimal-test-case?

BurningTreeC:
@pmario - this is a current prerelease with my additions from this PR:

http://selectabledropdowninput.tiddlyspot.com

- keep the top 3 state tiddlers open
- try the tags Input of `HelloThere`
- the first state is the real text value of the tags input
- the second state is the user input by which the dropdown is filtered
- the third state is the state used to refresh the input field (its text is updated and focus is set)
- in the tags input try:
-- up arrow / down arrow to select a tag from the dropdown
-- write something to filter the dropdown
-- up / down again, to see that the input updates with the selected tag but the dropdown filtering remains the same
-- press Enter to add a tag (dropdown filtering remains the same), one can add more tags by same filter
-- press Escape to reset the input field (all tags will be shown again, input gets cleared)
-- on Escape: if all tags are shown and input field is clear, it cancels editing
-- that's all

pmario:
Great stuff! .... As you wrote, it would be cool to have it for the search input. Users would love this!

BurningTreeC:
thanks @pmario , for the tags and search dropdowns, we have to take into account that we can add tabs within the dropdowns with different filterings

the select-logic by up/down needs the filters that are used within the current open dropdown tab

I think therefor we should put the filters (the tags and search dropdowns have to different filters for non-system and system) for the dropdown lists into a field somewhere and document the mechanism. Then I can make the macro that uses the tab-state to look up for the filters

BurningTreeC:
Demo is up-to-date with both tags-dropdown and search-dropdown working:

http://selectabledropdowninput.tiddlyspot.com/

- up/down selects
- enter adds tag / opens search tiddler in river
- escape clears input / cancels editing tiddler
- search dropdown handles currently active search-dropdown tab

BurningTreeC:
@Jermolene - In case you still have that startup module, would you post it or send me a link to it? I would give it a try testing

BurningTreeC:
@Jermolene - I have a hard time imagining which states may be kept in the dom here

Jermolene:
> I have a hard time imagining which states may be kept in the dom here

Apologies, I hadn't looked at your implementation then. I was expecting that you'd be reusing the browser's existing focus handling, which is generally recommended as being better for accessibility and generality.  But it looks like your implementation is more specific to a text box+dropdown, and therefore quite a bit simpler. I'm afraid I've not had a chance to look at it in great detail.

BurningTreeC:
@Jermolene, no problem!

I tried keeping everything very simple here, the solution with the three state tiddlers is the simplest I found

the focus is handled by a third state-tiddler that's passed as `refreshTiddler` to the edit-text widget. When that tiddler changes, the input gets focus (this is done in factory.js , in the refresh prototype)

I make sure that the tiddler changes when a dropdown item gets selected and when an item gets added or the input is cleared

the refresh tiddler also has the tiddler title in a state field, which assures that we're always focusing the input of that tiddler and not any other tiddler open for editing

BurningTreeC:
I made a startup module that triggers "$:/config/MissingLinks" yes/no every second (with the latest tv-hide-missing-links on top of the pagetemplate) ... is that sufficient for a full refresh? I don't see any problem at the moment

Jermolene:
Hi @BurningTreeC apologies I was about to work on that! The configuration you've got is useful but it's also important to try it while marking all tiddlers as having been modified, so that everything gets rebuilt on each refresh cycle.

BurningTreeC:
- changed the refresh-mechanism to a more generic solution
  - the `refreshTiddler` attribute is the tiddler that must change in order to trigger a refresh and
  - the `refreshCondition` attribute must be `true` or `yes` in order to trigger a refresh
  - the `refresh` if the refreshTiddler changed and the condition is met  is updating the text of the input (changed from "outside" through an action-widget) and setting  focus
- removed the `cancelPopups` attribute, I believe popups will be handled better sooner or later

BurningTreeC:
I'm done here

- made the refresh mechanism even more generic by using a `refreshAction` attribute that lets us select what happens when the refreshTiddler changes and the refreshCondition is met.
- currently there's only "focus-update" which is also the default. It updates the input with changed text and sets focus back to it
- an example: this mechanism can be used for the editor to set the cursor back to it when we open/close the preview panel

Jermolene:
Hi @BurningTreeC thanks for this and #3590. Perhaps it would make sense for me to add a "5.1.19" label to the PRs that we're deferring. Are there any others besides these two?

BurningTreeC:
Thanks @Jermolene - I would consider #3309 , too. Up to you :)

Jermolene:
Thanks @BurningTreeC that makes sense

BurningTreeC:
Awesome, thank you! I'm really looking forward to this 5.1.18 release. Thank you for making this possible

Jermolene:
Not at all, thanks to you @BurningTreeC for your many contributions to 5.1.18, and for being so instantaneously responsive as we've worked up to the release!

BurningTreeC:
Hi @Jermolene ,

in this PR I introduced a mechanism that refreshes an input field on a given condition

Now I've taken this a bit further and made an `inputManager` module like the keyboardManager, that handles these refreshes when they are desired and a given refreshCondition is met, and it also stores cursor positions and selections and the last focused input (connecting a set of inputs that belongs to the same tiddler through a qualifier) ... so that it's able to restore selections, too ... always just only if desired and the condition is met

This `inputManager` is used [here](https://burningtreec.github.io/TW5-navigator/) and can be tried out by focusing the editor and opening/closing the preview (cursor gets set back to previous position/selection) or in the tags input or sidebar-search input, where the `inputManager` just refreshes the input text (because it's changed by an action-setfield widget) and re-focuses the input

the refresh-behavior is defined by the `refreshAction` attribute of the `edit` and `edit-text` widgets, triggered only if the `refreshCondition` attribute is "true" and the tiddler given with the `refreshTiddler` attribute has changed. The message is sent in `factory.js`.

The `framed.js` and `simple.js` engines do save their selections when they get focus or blur, and when they get focus they set a `focusedInput` property to their `qualifier` , which is made of the title of the tiddler they are editing + a state qualifier

This can also be used in combination with #3590 so that if more tiddlers are open for editing and one gets closed, the input of the next one (if it's open for editing) gets focus again...

Jermolene:
Hi @BurningTreeC that's very helpful, thank you. The functionality gained here is very valuable.

My concern is that this design exposes a lot of the details of the refresh mechanism to wikitext authors, making it hard to understand these features without a good grasp of those details. I'm particularly thinking of the `refreshAction/Condition/Tiddler` attributes.

I had been thinking about a different approach. It's not entirely thought through, but the idea was to change the refresh mechanism so that the details of the currently focused element and text selection are saved before the refresh process, and then restored afterwards. Doing that requires identifying the corresponding elements before and after a refresh which could perhaps be done by making the qualified identifier of an element be exposed as an attribute in the DOM, and then we can perhaps use a CSS selector to find the element again.

I think the approach you're taking may have a few advantages: your approach delegates more of the work to the widgets themselves which seems like better encapsulation, and offers significant advantages by retaining a stack of focus/selections.

BurningTreeC:
@Jermolene ... maybe a combination of your approach and the advantages of mine?

BurningTreeC:
@Jermolene - the refreshTiddler/Condition/Action attributes seemed to be the best approach to me, because I felt it was opinionated deciding when it should refresh or trigger the action and being able to have some control here would also make tm-edit-text-operation more awesome, thinking about dropdowns below the selection, choosing text-completions, tiddler-titles ... the attributes could eventually be marked as advanced, but I think they're not that difficult to understand ... what do others think about this?

core/wiki/macros/input.tid
```diff
@@ -0,0 +1,67 @@
...
+<$list filter="[<__saveTiddler__>get[text]minlength[$filterMinLength$]] [[$filterMinLength$]removeprefix[0]suffix[]]">
+<$vars userInput={{{ [<__saveTiddler__>get[text]] }}} selectedDropDownItem={{{ [<__editTiddler__>get[text]] }}}>
+<$set name="dropDownFilter1" filter=<<__filter1__>>>
+<$set name="dropDownFilter2" filter=<<__filter2__>>>
+<$set name="dropDownList" filter=<<get-dropdown-list>>>
+<$list filter="[enlist<dropDownList>] +[$beforeafter$<selectedDropDownItem>] ~[enlist<dropDownList>$reverse$nth[1]]" variable="nextTag">
+<$action-setfield $tiddler=<<__editTiddler__>> text=<<nextTag>>/>
+<$action-deletetiddler $tiddler=<<__refreshTiddler__>>/>
+<$action-setfield $tiddler=<<__refreshTiddler__>> refresh-qualifier={{{ [[$(tv-refresh-input-qualifier)$]] }}}/>
+<$action-setfield $tiddler=<<__refreshTiddler__>> text=<<nextTag>>/>
```
BurningTreeC: theses two lines make the input field refresh after its text has changed through up/down-arrow selections

BurningTreeC:
Hi @Jermolene , are you interested in this?

You proposed a refresh-mechanism for inputs that restores focus. Have you got an idea how to do that?

Jermolene:
> are you interested in this?

Very interested in gaining the functionality, but keen to do it right...

> You proposed a refresh-mechanism for inputs that restores focus. Have you got an idea how to do that?

Tentatively, one possible approach is to have a tiddler called `$:/state/current-focus` that contains a string identifying the widget that owns the focus. Then each widget instance that wants to opt-in to the mechanism would need to have an attribute that gave the qualified ID for that widget. Then at refresh, the widget would detect whether it owned the focus, and set the focus to itself if so.

Jermolene:
btw the other thing that we should be able to fix in this way is maintaining the focus/selection while editing fields within the current tiddler...

BurningTreeC:
>btw the other thing that we should be able to fix in this way is maintaining the focus/selection while editing fields within the current tiddler...

Ah yes, that should be done while we're on it.

So, if I understand it right, the widgets should define a unique ID and populate `$:/state/current-focus`  with it when they get focus. When they leave focus, they check if the ID in that tiddler matches their ID and maybe set `this.widget.shouldFocusAgain` to true (?) and after a refresh of the widget, if `this.shouldFocusAgain` is true, it sets focus ...?

Jermolene:
Hi @BurningTreeC

That's the idea, yes. There are quite a few points to think about:

* Triggering a refresh cycle every time the focus changes might be burdensome, so we may need to e.g. suppress change events for changes to the selection tiddler that have been immediately actioned
* We'll also need to track the start and end positions of the text selection
* I don't think that there is a reasonable basis for the widgets to generate the IDs themselves, but perhaps we should explore whether there are any viable strategies.

BurningTreeC:
>Triggering a refresh cycle every time the focus changes might be burdensome, so we may need to e.g. suppress change events for changes to the selection tiddler that have been immediately actioned

That's the reason why I originally made the `refreshTiddler` attribute, so it does this kind of refresh only if that tiddler refreshes. That tiddler should refresh with the user-action, so it needs to be updated with an action-widget after a user-interaction. I understand that it's a bit complicated.

>We'll also need to track the start and end positions of the text selection

Yes, that's important. I was thinking about an `inputManager` startup module that initializes an array where inputs store their selectionStart and selectionEnd identified by the unique ID we're going to use ...

>I don't think that there is a reasonable basis for the widgets to generate the IDs themselves, but perhaps we should explore whether there are any viable strategies.

I couldn't find another way ... a tiddler can have many inputs that edit the same tiddler and the same field/index and they could all be siblings, so a state qualifier using a combination of the editTitle, editField/Index and stateQualifier wouldn't be unique. I came up with adding the index of the widget within the array of its siblings to that identifier which is working well:

```
this.editQualifiedID = this.editTitle + (this.editField ? this.editField : "") + (this.editIndex ? this.editIndex : "") + this.getStateQualifier() + this.getOwnSiblingsWidgetIndex;
```

But possibly I'm just not seeing the best approach

BurningTreeC:
I see now that my approach would need to register each widget within an array of its parent widget

pmario:
>btw the other thing that we should be able to fix in this way is maintaining the focus/selection while editing fields within the current tiddler...

I think this would be a "killer" ;)

BurningTreeC:
Hi @Jermolene , this PR has merge conflicts that could be resolved. Do you have an idea/a roadmap how to go on with this? The unique widget id seems to be difficult to solve, this mechanism here uses a different approach, demanding that a given condition must be true in order for the input to refresh its text and focus again.

the condition is that a given `refreshTiddler` has changed and that the `refreshCondition` attribute is "true". it's a bit tricky as a mechanism, but it works

Jermolene:
> Do you have an idea/a roadmap how to go on with this? The unique widget id seems to be difficult to solve, this mechanism here uses a different approach, demanding that a given condition must be true in order for the input to refresh its text and focus again.

It's tricky. The problem is that if we merge this then we're committed to supporting this technique forever, and that might limit our ability to address related problems in the future. On balance, experimental stuff like is good to chew over as a plugin. Do you think that's feasible, if we added enough core hooks?

BurningTreeC:
@Jermolene , I know I'm absolutely not convinced that this should be the mechanism to stick with forvever. but it's a mechanism that works, now.

the plugin proposal would be just right. about the hooks ... would you just add a hook in the factory.js refresh method, passing the widget?

Jermolene:
> would you just add a hook in the factory.js refresh method, passing the widget?

Yes I think that's reasonable.

BurningTreeC:
> > would you just add a hook in the factory.js refresh method, passing the widget?
> 
> Yes I think that's reasonable.

ok, there are more points where factory.js would need to be extended, see https://github.com/Jermolene/TiddlyWiki5/pull/3565/files#diff-2788afb61c207f85b558b296f643ce5e

what's the best way to do it?

Jermolene:
> what's the best way to do it?

Too many hooks at random places will also box us in too far. It may be reasonable to look at subclassing the existing factory.js, I'm not sure how much hackery would be needed to make that possible though...

BurningTreeC:
> > what's the best way to do it?
> 
> Too many hooks at random places will also box us in too far. It may be reasonable to look at subclassing the existing factory.js, I'm not sure how much hackery would be needed to make that possible though...

ok, I'll make a minimal working concept and see where subclasses make sense without overdoing it

BurningTreeC:
Hi @Jermolene,

in order to make this functionality work using hooks and a startup module, I need 3 hooks, one in the refresh method, one in the execute method and one in the getEditInfo method of factory.js

it's really simple to extend it this way, I think it'd be a great hackability extension

BurningTreeC:
> in order to make this functionality work using hooks and a startup module, I need 3 hooks, one in the refresh method, one in the execute method and one in the getEditInfo method of factory.js

nah, this was exaggerated, I only need 1 hook in the refresh method

BurningTreeC:
opening this again for discussion - I got a version working that uses the new widget-subclassing here: http://selectable-inputs.tiddlyspot.com/

BurningTreeC:
closing in favor of a new attempt

PR #3567, opened 2018-11-21
fix sticky titles + more-tiddler-actions popups

### LINKED ISSUE #3566: Tiddler popup behind other tiddler titlebars
Tiddler popup menu is partially coverd by other tiddlers title bar. This happens when tiddlers are folded or a tiddlers height is lesser than the popup menus height.

image
All tiddlers folded

image
Empty tiddler is shorter than popup menu

Tested in 
* Firefox 63.0.1 (64-bit)
* Chrome 70.0.3538.110 (Official Build) (64-bit)

With tiddlywiki version 5.1.17

Using Nodejs 8.12.0 on windows 10

BurningTreeC:
this fixes #3566 - more-tiddler-action popups hiding behind sticky titles of next tiddlers in story-river


core/wiki/macros/CSS.tid
```diff
@@ -74,3 +74,47 @@ column-count: $columns$;
 \define if-background-attachment(text)
 <$reveal state="$:/themes/tiddlywiki/vanilla/settings/backgroundimage" type="nomatch" text="">$text$</$reveal>
 \end
+
+\define if-sticky-titles-calc-zindex()
```
Jermolene: Perhaps better called "calc-zindex-for-sticky-titles"?

core/wiki/macros/CSS.tid
```diff
@@ -74,3 +74,47 @@ column-count: $columns$;
...
+``
+[data-tiddler-title="``$(cssEscapedTitle)$``"] .tc-tiddler-title {
+	z-index: calc(500 + ``$(storyCount)$`` - ``$(tiddlerIndex)$``);
```
Jermolene: Perhaps we should have a fallback declaration for browsers that don't have "calc"?

core/wiki/macros/CSS.tid
```diff
@@ -74,3 +74,47 @@ column-count: $columns$;
...
+	position: -moz-sticky;
+	position: -o-sticky;
+	position: -ms-sticky;
+	position: sticky;
+	top: 0px;
+	background: ``<<colour tiddler-background>>``;
+}
+``
+
+<$vars storyCount={{{ [list[$:/StoryList]count[]] }}}>
```
Jermolene: This sequence is very cunning, and now I see why you were interested in the CSS escaping! Great stuff. I've tried and failed to come up with a pure CSS solution, this looks like the next best thing.

core/wiki/macros/CSS.tid
```diff
@@ -74,3 +74,47 @@ column-count: $columns$;
...
+
+<$list filter="[list[$:/StoryList]]" variable="tiddlerTitle">
+
+<$list filter="[list[$:/StoryList]] +[allbefore<tiddlerTitle>count[]]" variable="tiddlerIndex">
```
Jermolene: Could this be:

```
<$set name="tiddlerIndex" value={{{ [list[$:/StoryList]allbefore<tiddlerTitle>count[]] }}}>
```

core/wiki/macros/CSS.tid
```diff
@@ -74,3 +74,47 @@ column-count: $columns$;
...
+
+\define if-sticky-titles-calc-zindex()
+``
+[data-tiddler-title="``$(cssEscapedTitle)$``"] .tc-tiddler-title {
+	z-index: calc(500 + ``$(storyCount)$`` - ``$(tiddlerIndex)$``);
+}
+``
+\end
+
+\define if-sticky-titles()
```
Jermolene: Perhaps this should just be "sticky-titles"?

core/wiki/macros/CSS.tid
BurningTreeC: Thanks @Jermolene ... I also couldn't figure out how to do it in css. I had already figured out this logic for applying z-indices so it was just a matter of doing this when I saw the issue...

core/wiki/macros/CSS.tid
BurningTreeC: Yes probably, I've tried finding some good fallback options but I haven't found any. Can you suggest one?

Alternativelly, I have a fully working version without calc, just `z-index: $(value)$`

core/wiki/macros/CSS.tid
BurningTreeC: yes, I've changed it

core/wiki/macros/CSS.tid
BurningTreeC: this is also changed now

BurningTreeC:
@Jermolene - for the `calc()` fallback, the best I could figure out is a version without `calc()` :+1: 

```
\define set-zindex-for-sticky-titles()
``
[data-tiddler-title="``$(cssEscapedTitle)$``"] .tc-tiddler-title {
	z-index: ``$(zIndex)$``;
}
``
\end

\define get-full-zindex-list()
$(zIndexFilter)$ $(storyList)$
\end

\define sticky-titles()
<$reveal state="$:/themes/tiddlywiki/vanilla/options/stickytitles" type="match" text="yes">

``
.tc-tiddler-title {
	position: -webkit-sticky;
	position: -moz-sticky;
	position: -o-sticky;
	position: -ms-sticky;
	position: sticky;
	top: 0px;
	background: ``<<colour tiddler-background>>``;
}
``

<$vars storyCount={{{ [list[$:/StoryList]count[]] }}}>

<$list filter="[list[$:/StoryList]]" variable="tiddlerTitle">

<!-- addsuffix tiddlerTitle should assure that later on we have a list with unique entries -->
<$set name="zIndexFilter" filter="[range[1,500]addsuffix<tiddlerTitle>]">

<$set name="storyList" filter="[list[$:/StoryList]]">

<!-- a list containing 500 entries + the amount of tiddlers in the story river -->
<$vars zIndexList=<<get-full-zindex-list>>>

<$set name="tiddlerIndex" value={{{ [list[$:/StoryList]] +[allbefore<tiddlerTitle>] [<tiddlerTitle>] +[count[]] }}}>

<$set name="reversedTiddlerTitle" value={{{ [list[$:/StoryList]reverse[]nth<tiddlerIndex>] }}}>

<$list filter="[enlist<zIndexList>] +[allbefore<reversedTiddlerTitle>count[]]" variable="zIndex">

<$vars cssEscapedTitle={{{ [<tiddlerTitle>stringify[]] }}}>

<<set-zindex-for-sticky-titles>>

</$vars>

</$list>

</$set>

</$set>

</$vars>

</$set>

</$set>

</$list>

</$vars>

</$reveal>
\end
```

BurningTreeC:
For cases like the Dynaview demo we'll probably want to add an exclude filter

BurningTreeC:
@Jermolene - thinking about it, if we just change the tiddler-info-state from a qualified state to a state containing the tiddler title, we just list all tiddler-info-states and set the z-index of that tiddler and the next 2-3 in the story-river ...

BurningTreeC:
... or we use/find/create a mechanism that allows extracting the tiddler-title from the qualified state

BurningTreeC:
... all this for performance reasons

pmario:
I'm not sure, but it seems your code has a lot of empty lines..... Please make sure, that no additional paragraphs are created in the rendered DOM!

BurningTreeC:
thanks @pmario - I've removed unnecessary empty lines and added a `whitespace trim` on top. Looks clean now in the inspector

Jermolene:
Hi @BurningTreeC 

>  for the calc() fallback, the best I could figure out is a version without calc()

I meant that maybe we should precede the z-index declaration that uses calc with another declaration that avoids calc. Thus browsers that don't understand calc will still get a z-index.

> For cases like the Dynaview demo we'll probably want to add an exclude filter

The exclusion mechanism looks like it allows individual tiddlers to be opted out of the Z-index calculation, is that right? What's the motivation?

BurningTreeC:
Hi @Jermolene 

>I meant that maybe we should precede the z-index declaration that uses calc with another declaration that avoids calc. Thus browsers that don't understand calc will still get a z-index.

Yes, after a while thinking about it I understood. I simply added the default `500` z-index as fallback
In another version, I don't use `calc()` but it's not that good performance-wise when there are many tiddlers in the river

>The exclusion mechanism looks like it allows individual tiddlers to be opted out of the Z-index calculation, is that right? What's the motivation?

The motivation came from trying this in your dynaview demo with `> 800` tiddlers. This exclusion mechanism allows excluding all that are not shown. It's not necessary calculating their indices...

But the last idea I had was changing the state-titles of the more-actions popup so that they have a defined prefix and the tiddler title as suffix. Then we can set the z-indices for the tiddler with the popup open and the next few tiddlers in the river, which would be the best solution performance-wise I think

Jermolene:
@BurningTreeC if there are performance concerns with this we might be better off making it a plugin? At least until we have some experience of it in the field?

BurningTreeC:
@Jermolene, the idea about getting the tiddler with the dropdown through the state-title would solve any performance issue

Jermolene:
> @Jermolene, the idea about getting the tiddler with the dropdown through the state-title would solve any performance issue

Rather than changing the semantics of the existing tiddlerInfo state tiddler, perhaps we should introduce something new?

In terms of performance, one of the impacts of this change is that the entire vanilla palette has to be re-wikified every time the story list changes. We can mitigate that a little by putting this stuff in it's own stylesheet tiddler.

BurningTreeC:
>Rather than changing the semantics of the existing tiddlerInfo state tiddler, perhaps we should introduce something new?

Yes, how about a mechanism to trace back the tiddler-title from a qualified state?

>In terms of performance, one of the impacts of this change is that the entire vanilla palette has to be re-wikified every time the story list changes. We can mitigate that a little by putting this stuff in it's own stylesheet tiddler.

Oh yes, that makes much sense

Jermolene:
> Yes, how about a mechanism to trace back the tiddler-title from a qualified state?

The motivation for some of the present complexity was to be able to be able to have nested story rivers, so it's worth thinking about the impact on them.

BurningTreeC:
>Rather than changing the semantics of the existing tiddlerInfo state tiddler, perhaps we should introduce something new?

What's your idea about this?

>The motivation for some of the present complexity was to be able to be able to have nested story rivers, so it's worth thinking about the impact on them.

I'll need to understand what impacts what in that case, maybe I get it if I understand your original idea

Jermolene:
> What's your idea about this?

I was thinking about 
https://github.com/Jermolene/TiddlyWiki5/pull/3567#issuecomment-441307048 where it sounds like a change that isn't backwards compatible.

Jermolene:
Should this be closed in favour of #3572

BurningTreeC:
@Jermolene where/how does changing the more-popup state break backwards-compatibility?

Jermolene:
> where/how does changing the more-popup state break backwards-compatibility?

Plugins could manipulate it to programmatically summon the popup... I agree that it's not likely, but I've observed people hardcoding the state tiddlers for the sidebar tabs, and so I've come to regard the state tiddler titles as needing backwards compatibiltiy. We could make a case here that the backwards compatibility risk is low, especially since #3572 is so much nicer...

BurningTreeC:
>We could make a case here that the backwards compatibility risk is low, especially since #3572 is so much nicer...

I vote for making a case. I'll leave a post in the community if people are interested to add an opinion

Jermolene:
> I'll leave a post in the community if people are interested to add an opinion

I'd be inclined to think that the discussion here is sufficient. If we are going to break something, the best way to find out if it matters is often just to go ahead and break it :)

BurningTreeC:
>I'd be inclined to think that the discussion here is sufficient. If we are going to break something, the best way to find out if it matters is often just to go ahead and break it :)

:+1: let's break it!

Jermolene:
Hi @BurningTreeC can we close this?

BurningTreeC:
@Jermolene yes!

PR #3571, opened 2018-11-23
Translate instructions to Spanish about install plugins in node server.

G0erman:
Translate instructions to Spanish about install plugins in node server.


Jermolene:
Thanks @G0erman

PR #3574, opened 2018-11-24
CSS.escape as startup module - seems to work

BurningTreeC:


Jermolene:
Hi @BurningTreeC thinking about it more, perhaps the polyfill should just be a plain old startup module? that way polyfills can mark themselves for execution in the browser/node etc.

Am I right in thinking this is only needed for v5.1.18 if we merge #3572?

Jermolene:
Thanks @BurningTreeC

BurningTreeC:
Hi @Jermolene , how do you want to proceed with this?

I tested this on an old firefox without CSS.escape support and it works

BurningTreeC:
Hi again @Jermolene ... this would be interesting if we add the option of storing the navigated tiddler somewhere, like in #3590 , possibly with an option in the controlpanel to a) enable storing the navigated tiddler and b) showing it emphasized

1) would that be something to be considered for the core?

2) i'm not sure if the way how I add this polyfill here is correct. it works on an old firefox that doesn't have CSS.escape (tested) and then gets support through this (tested, too). but I'm lacking knowledge about the crossbrowser way of adding this polyfill

Jermolene:
I'm hoping we can use the browser focus mechansim, and then we could style the visual appearance of tiddlers that contain the focus with CSS `:focus-within`.

BurningTreeC:
> I'm hoping we can use the browser focus mechansim, and then we could style the visual appearance of tiddlers that contain the focus with CSS `:focus-within`.

... aahhh! now I get what you're up to!

yep, if that's doable, great! :+1:

BurningTreeC:
Should we use this?

Jermolene:
Hi @BurningTreeC apologies, I'm working my way back through the email notifications and am seeing this after #3546. I'll go ahead and merge this.

PR #3579, opened 2018-11-24
Update "SystemTag_ $__tags_SideBarSegment.tid"

twMat:
The possibility to rearrange stuff from "inside" tag pills is not obvious so makes sense to mention.


Jermolene:
Thanks @twMat

PR #3583, opened 2018-11-25
Update to fr-FR language translations

xcazin:
Plus swap read and write in the description of `readers` and `writers` parameters in core/language/en-GB/Help/listen.tid


Jermolene:
Thanks @xcazin

Issue #3582, opened 2018-11-25
No space between tag pills in edit template


Jermolene:
One of the recent changes to tag rendering in the edit template has lost the space that should be separating each tag pill:

![c6c98d2a-4eb2-42d4-a211-d0f3473f9130](https://user-images.githubusercontent.com/174761/48980203-039eee80-f0be-11e8-9ca8-609c7b74cedd.jpeg)

BurningTreeC:
I noticed that, too. I think we should just add a margin-right, maybe not for the last child

pmario:
the pill has a class `tc-tag-label` ... if it would be `tc-tag-label tc-tag-list-item` it would get the right margin with 7px like in view-mode

pmario:
In view mode the elements also contain a text element, which shouldn't be there. imo the text element is a bug.

pmario:
if $:/core/ui/TagTemplate would have a `\whitespace trim` pragma, the text element would be gone

pmario:
ino can be closed.

Jermolene:
Thanks @pmario

PR #3585, opened 2018-11-26
fix missing space between edittemplate tags (#3582)

BurningTreeC:
... and remove a space that doesn't do anything here


BurningTreeC:
fixes #3582

Jermolene:
Thanks @BurningTreeC

PR #3586, opened 2018-11-26
undo #3490 - error in  popup-cancelling logic

BurningTreeC:


BurningTreeC:
the problem #3490 tries to fix is that inputs with dropdowns don't cancel other eventually open popups

to reproduce:

- click into a tags-input field so that its dropdown is shown
- then click into the search field of the sidebar. its dropdown doesn't show up

#3309 adds two attributes that can be used for the `edit-text` and `edit` widgets: `cancelPopups` and `tabIndex`

I've added a more detailed explanation to the PR, [see](https://github.com/Jermolene/TiddlyWiki5/pull/3309#issuecomment-441561012)

sukima:
For the future, Both Git and GitHub have a revert feature.

### Git

GitHub offers the merge commit:

> Jermolene merged commit   into Jermolene:master a day ago

```console
$ git checkout -b revert-pr-3490
$ git revert 09ea592
$ git push -u origin revert-pr-3490
```

Then make a PR in GitHub

### GitHub only

screenshot

Pressing **revert** will create another PR that does the above Git commands.

BurningTreeC:
Thanks for your kind explanation @sukima

Jermolene:
Thanks @BurningTreeC

PR #3587, opened 2018-11-26
fix textnodes between tags in viewtemplate

BurningTreeC:
this removes additional text-nodes in the dom after each `tc-tag-list-item` caused by the last empty line
we could also just remove that line but I don't know if that's a permanent solution or if some mechanism will re-add that line at some point, so I go with the `whitespace trim` pragma


BurningTreeC:
I don't know if all browsers show those nodes in the inspector - Firefox does

Jermolene:
Thanks @BurningTreeC

PR #3596, opened 2018-11-30
style the framed editor

BurningTreeC:
this makes the framed editor look a bit less ugly :sunglasses:

### DIFF
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -39,9 +39,11 @@ function FramedEngine(options) {
 	this.iframeDoc.close();
 	// Style the iframe
 	this.iframeNode.className = this.dummyTextArea.className;
-	this.iframeNode.style.border = "none";
+	this.iframeNode.style.border = "2px solid " + this.widget.wiki.extractTiddlerDataItem(this.widget.wiki.getTiddlerText("$:/palette"),"muted-foreground");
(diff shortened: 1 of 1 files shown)


BurningTreeC:
box-shadow on the iframe would look really good, I just don't know how to set it ... should it be copied from the tiddler frame to know if the theme sets shadows?

BurningTreeC:
ah, nevermind, I found the solution

BurningTreeC:
this removes also the bottom right resize handle which looks bad and has no real value

Jermolene:
Hi @BurningTreeC my fault, I did not realise that this was a new change, I just assumed that it was another of the recent torrent of tweaks.

I think the moral of the story is that if you're going to suggest a change in the core look and feel then it's important to explain what you're doing and why, and ideally show before and after pictures so that as many people as possible can participate.

Jermolene:
Hi @BurningTreeC I'm having trouble reverting this commit because of merge conflicts. Would you be able to prepare a new PR that restores the editor textarea appearance to where we were last week? Many thanks.

BurningTreeC:
>I think the moral of the story is that if you're going to suggest a change in the core look and feel then it's important to explain what you're doing and why, and ideally show before and after pictures so that as many people as possible can participate.

yeah, this was a bit premature of myself - I take the responsability

PR #3600, opened 2018-12-01
look inside! - update katex to 0.10.0

BurningTreeC:
this updates katex to 0.10.0 but we need to look at the minified css without font-faces

I just removed all `@font-face` definitions from the new minified css ... I don't know if there are rendering issues that have been addressed in the past that get undone by this


Jermolene:
Hi @BurningTreeC looks good -- we duplicate those fontface definitions in styles.tid, but they don't seem to have changed in this release.

PR #3601, opened 2018-12-01
katex: add mchem extension for chemical syntax

BurningTreeC:


Jermolene:
Hi @BurningTreeC that's great, thanks. We need an update to the plugin docs, and maybe a new example(s) in the demo edition?

PR #3602, opened 2018-12-01
remove bad outline style from remove-tag-button

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

pmario:
IMO this PR shouldn't have been merged because of this: https://developer.mozilla.org/en-US/docs/Web/CSS/outline#accessibility_concerns

which says: 

>Assigning outline a value of 0 or none will remove the browser's default focus style. If an element can be interacted with it must have a visible focus indicator. Provide obvious focus styling if the default focus style is removed.
>
>    [How to Design Useful and Usable Focus Indicators](https://www.deque.com/blog/give-site-focus-tips-designing-usable-focus-indicators/)
>    WCAG 2.1: [Understanding Success Criterion 2.4.7: Focus Visible](https://www.w3.org/WAI/WCAG21/Understanding/focus-visible.html)

We did remove the focus styling, because it was described as "bad" ... it's a visual glitch ... but we didn't implement an improvement which means, that now we have a visually broken behaviour, which imo is worse. 

It should look similar like this

<details><summary>show the GIF</summary>

outline-missing

</details>

IMO a "perfect" visual behaviour would be like the mockup image below

image

PR #3603, opened 2018-12-01
big oops - katex plugin fix

BurningTreeC:


Jermolene:
No worries, I think we caught it at the same time :)

PR #3614, opened 2018-12-03
revert setting styles in framed.js

BurningTreeC:
this sets the framed-editor styles to 5.1.17 state


Jermolene:
Thanks @BurningTreeC

PR #3622, opened 2018-12-05
toc macro - macrocall wasn't closed

BurningTreeC:


Jermolene:
Good spot, thanks @BurningTreeC

PR #3634, opened 2018-12-10
Warn when importing core tiddlers

bimlas:
See https://groups.google.com/d/msg/tiddlywiki/YZlPGP0qX1o/3tUpyoHXDgAJ

image

### DIFF
--- a/core/language/en-GB/Import.multids
+++ b/core/language/en-GB/Import.multids
@@ -19,3 +19,4 @@ Upgrader/Plugins/Upgraded: Upgraded plugin from <<incoming>> to <<upgraded>>
 Upgrader/State/Suppressed: Blocked temporary state tiddler
 Upgrader/System/Suppressed: Blocked system tiddler
 Upgrader/ThemeTweaks/Created: Migrated theme tweak from <$text text=<<from>>/>
+Warning/Core: You are about importing risky tiddlers (<<titles>>). It is strongly recommend in the strongest terms NOT to use any hack that requires overwriting a core module, because it is likely to make the system unusable. Never import such tiddlers into an important wiki, just into a testing environment!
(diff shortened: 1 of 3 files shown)


Jermolene:
Hi @bimlas this is a great idea. I have two thoughts:

* Checking for the prefix `$:/core/` may not be quite right; there are some templates that are reasonable to override. So perhaps we should check for `$:/core/modules/`
* The implementation can be simplified by plugging into the existing "upgrader" mechanism -- basically, "upgrader" modules are called for each tiddler being imported, and gives them an opportunity to block them, or add a warning message. See the tiddlers prefixed `$:/core/modules/upgraders/` for the three core upgraders, and the `invokeUpgraders()` method in wiki.js. For this we'd add a new upgrader module for handling core module overwrites

pmario:
I think, the text "You are about importing risky tiddlers... " is misleading. The action to overwrite core modules is risky. ... The modules themselves are well tested and not risky.

bimlas:
@pmario Please, write the message that should appear - you are better in English than me. :)

pmario:
I'm not sure, why navigator.js needs to check for this behaviour. It's not responsible for the import mechanism. 

There is the `th-importing-tiddler` hook, which can be used to add additional checks and test them.

bimlas:
I've moved the changes to the $:/core/modules/upgraders/system.js because that performs similar checks.

image

00SS:
Suggested slightly "better English" wording for the warning:

You are about to import a tiddler that will overwrite a core module tiddler. This is a risky action that may make the system unstable. It is strongly recommended that you do NOT overwrite these tiddlers in an important TiddlyWiki; only in a testing environment.

bimlas:
Sorry for the late reaction, I rewrote the text.

Jermolene:
Thanks for your patience @bimlas

PR #3639, opened 2018-12-13
Docs: Add pre-release location info tiddler

00SS:
Apologies if I am doing this wrong.


00SS:
I don't know what else to do to fix this pull request to add the information tiddler: Prerelease.tid

Please let me know if I should close this request and open a new one.

Thanks

00SS:
Perhaps this tiddler can explain why the pre-release exists - who would look at it - how it is useful.

00SS:
@Jermolene 
Is there some reason why providing this information is not a good idea?

The only alternative I can think of is to include it at the top of: https://tiddlywiki.com/#Releases

If that is a better idea, please let me know so I can edit that tiddler instead.

Thanks

Issue #3636, opened 2018-12-10
Issue with search in 5.1.18?


hoelzro:
Hello!

I know that the search operator saw some changes for 5.1.18 - I noticed a change in behavior between 5.1.17 and 5.1.18, and I was wondering if it was deliberate.

If I create a tiddler in the empty edition with the title "Test tiddler" and content "Body content", searching the wiki for "test body" yields no results under either "title matches" or "all matches".  Searching for either word individually turns up "Test tiddler", and repeating this in an empty wiki I created from the 5.1.17 tag causes "Test tiddler" to show up under "all matches".

Was this change intentional, or is this a bug?  If the former, is there a way to configure a wiki to restore the old behavior?

Jermolene:
Thanks @hoelzro I think that's a subtle regression, we should revert to the previous behaviour for v5.1.19

hoelzro:
@Jermolene I'm up for attempting a fix - I read through the search function's source and have a pretty good idea what to do.  Are there tests for the search function anywhere, like in the test edition?

hoelzro:
I have a fixed-up version I'm monkey patching into my own wiki; I'll keep using it and make sure it behaves ok before issuing a PR adding the fix to the codebase:

```js
(function() {
exports.search = function(text,options) {
	options = options || {};
	var self = $tw.wiki,
		t,
		invert = !!options.invert;
	// Convert the search string into a regexp for each term
	var terms, searchTermsRegExps,
		flags = options.caseSensitive ? "" : "i";
	if(options.literal) {
		if(text.length === 0) {
			searchTermsRegExps = null;
		} else {
			searchTermsRegExps = [new RegExp("(" + $tw.utils.escapeRegExp(text) + ")",flags)];
		}
	} else if(options.whitespace) {
		terms = [];
		$tw.utils.each(text.split(/\s+/g),function(term) {
			if(term) {
				terms.push($tw.utils.escapeRegExp(term));
			}
		});
		searchTermsRegExps = [new RegExp("(" + terms.join("\\s+") + ")",flags)];
	} else if(options.regexp) {
		try {
			searchTermsRegExps = [new RegExp("(" + text + ")",flags)];			
		} catch(e) {
			searchTermsRegExps = null;
			console.log("Regexp error parsing /(" + text + ")/" + flags + ": ",e);
		}
	} else {
		terms = text.split(/ +/);
		if(terms.length === 1 && terms[0] === "") {
[...]

PR #3641, opened 2018-12-14
Fix search method for search tokens spread across fields

hoelzro:
Addresses GH #3636, which reports that if you're searching for "test
body", and "test" only appears in the title field, and "body" only appears
in the text field, 5.1.18's search method won't yield that tiddler as a
result, which appears to be a regression from the 5.1.17 behavior


Jermolene:
Thanks @hoelzro, looks good.

PR #3643, opened 2018-12-14
remove \import pragma from main readme

BurningTreeC:


Jermolene:
Hi @BurningTreeC the readme.md in the root of this repo is generated from [this file](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/tw5.com/tiddlers/readme/ReadMe.tid) in the tw5.com edition (see the [build script](https://github.com/Jermolene/build.jermolene.github.io/blob/master/readme-bld.sh)).

The problem in this case is that the readme was built with v5.1.17, which doesn't support the `\import` pragma. It will fix itself for v5.1.19.

BurningTreeC:
I see, thanks @Jermolene

Issue #3648, opened 2018-12-17
Fix: Display of backlinks when renaming a tiddler - Hide when none


AnthonyMuscio:
Show backlinks on rename (#3534) 

**During rename**
_The following references to this tiddler will not be automatically updated:
No tiddlers link to this one_

Please fix this so the "The following references to this tiddler will not be automatically updated:" is not shown unless there are references. at present you must click to see, "No tiddlers link to this one" and it is a wast of valuable screen space for no reason.

It would be useful if we could use a system tag so this function could be extended by users but this is not part of this request.

Jermolene:
Thanks @AnthonyMuscio that's a good idea, but I'm not going to tackle it for v5.1.19 because it's not a bug, and I'm trying to keep this release as low risk as possible.

AnthonyMuscio:
Thanks Diego, To me it may not be a bug but it was in need of better design.

diego898:
I believe #3654 successfully closes this. @AnthonyMuscio what do you think?

AnthonyMuscio:
Agreed Diego, a quick test on tiddlywiki.com 5.1.21 suggests this is fixed - thanks

PR #3654, opened 2018-12-19
Fixing 3648 - now non updated references are shown only when non zero

diego898:
This is a quick fix for #3648 

Signed-off-by: Diego Mesa <diego898@gmail.com>

### DIFF
--- a/core/ui/EditTemplate/title.tid
+++ b/core/ui/EditTemplate/title.tid
@@ -35,6 +35,10 @@ tags: $:/tags/EditTemplate
 
 <$checkbox tiddler="$:/config/RelinkOnRename" field="text" checked="yes" unchecked="no" default="no"> {{$:/language/EditTemplate/Title/Relink/Prompt}}</$checkbox>
 
+<$set name="backlinks" filter="[title<fromTitle>backlinks[]]">
+
+<$reveal type="nomatch" default=<<backlinks>> text="">
+
 <$vars stateTiddler=<<qualify "$:/state/edit/references">> >
 
 <$reveal type="nomatch" state=<<stateTiddler>> text="show">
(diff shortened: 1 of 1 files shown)


core/ui/EditTemplate/title.tid
```diff
@@ -35,6 +35,10 @@ tags: $:/tags/EditTemplate
 
 <$checkbox tiddler="$:/config/RelinkOnRename" field="text" checked="yes" unchecked="no" default="no"> {{$:/language/EditTemplate/Title/Relink/Prompt}}</$checkbox>
 
+<$set name="backlinks" filter="[title<fromTitle>backlinks[]]">
+
+<$reveal type="nomatch" default=<<backlinks>> text="">
+
```
pmario: Please also check the DOM. There are several paragraph elements, that shouldn't be there. Some text nodes that contain no text and so on.

diego898: hey @pmario, Im not exactly familiar with that - I just tried to follow the convention of the file - I guess that resulted in some extra empty elements. Its never really clear to me when a new line will show up or not. Any help you can provide would be greatly appreciated!

diego898: Hey Mario - just a bump on this. I'm eager to try and contribute something to TW!

diego898:
just a bump on asking for advice!

Jermolene:
Hi @diego898 the line spacing you've got looks consistent with the rest of the tiddler. However, I think it may be better to fix this with a `<$list>` widget to avoid introducting the DOM element that the `<$reveal>` widget always creates.

diego898:
Hey @Jermolene - Ive updated the PR to use the list widget. Thank you for the help!

diego898:
A slight update to make sure I dont overwrite the `<<currentTiddler>>` variable

diego898:
@Jermolene could this make it into 5.1.20? If not, what else is necessary?

Jermolene:
Thanks @diego898 looks good.

PR #3657, opened 2018-12-20
Fix outdated external-js template.

BramChen:
Add the missing raw markup sections:
  * the top of the head
  * the top of the body
  * the bottom of the body


pmario:
Why is it outdated? .. Can you be more specific pls.

BramChen:
TiddlyWiki5 has supported more raw markup tiddlers since v5.1.18 for `tiddlywiki5.html.tid`,to see [#c246fab](https://github.com/Jermolene/TiddlyWiki5/commit/cfb2d7c9c8d899729b67864070d1958b35cc0ca2#diff-c246fab0f60acddf67748471c9c4fdec). The contents within those sections would be lost if using the external JS template to generate/render the TW5.

Jermolene:
Thanks @BramChen that's very helpful.

PR #3668, opened 2018-12-29
Fix recursion error in toc() macro in: toc.tid

00SS:
I noticed that if a non-existent TAG name used in a simple Table of Contents macro:
```
<<toc "NonExistentTag">>
</div>
```
It gives a javascript recursion error.

The macro within tiddler **$:/core/macros/toc** is:
```
\define toc(tag,sort:"",itemClassFilter:" ")
<$macrocall $name="toc-body"  tag=<<__tag__>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> />
\end
```
Noticing that all the other ToC macros did not have this problem, I remembered a discussion earlier in GG explaining the significance of the **exclude** parameter.

Modifying the **toc** macro by adding the `exclude` parameter appears to fix this - but I do not know if this is the right fix!

```
\define toc(tag,sort:"",itemClassFilter:" ",exclude)
<$macrocall $name="toc-body"  tag=<<__tag__>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> />
\end
```

I am making a pull request just to practice using github.
Please close it if it is nonsense!

### DIFF
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -26,7 +26,7 @@ tags: $:/tags/Macro
 </ol>
 \end
 
-\define toc(tag,sort:"",itemClassFilter:" ")
+\define toc(tag,sort:"",itemClassFilter:" ",exclude)
 <$macrocall $name="toc-body"  tag=<<__tag__>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> />
 \end
 


pmario:
if you search for "macrocall" you will see, that the "toc" is _not_ called recursively. So IMO this isn't the problem. ... we need to check it, again.

pmario:
toc-body may contain a problem

00SS:
Is this fixed now? I can no longer reproduce the error - but neither can I see where it was fixed!
The tiddler which produced the error is exactly the same as in the present pre-release!
 `$:/core/macros/toc` 

I could not even reproduce it in an older TiddlyWiki! That makes me wonder if it was a javascript error coming from an older version of my browser!

Strange.

pmario:
Then you should close the issue

00SS:
Since it once was an issue, and something unidentified may have changed, I think if it would be useful if the issue no longer happening was confirmed by someone else too.

00SS:
Don't know what changed that fixed this issue, but glad it's fixed!

PR #3669, opened 2018-12-30
Fix diff-widget undefined attributes bug

Lioric:
Calling diff_main method with undefined parameters throws a (uncaught) "Null input" error.
This patch adds default values to the widget "source" and "dst" attributes so the diff method don't throw an error when widget attributes are not defined or their value is not an string object (undefined).

The default value string has been selected to help identifying the error when there is a syntax error in the widget argument names or their value is undefined. You might want to change it as you see fit.

### DIFF
--- a/core/modules/widgets/diff-text.js
+++ b/core/modules/widgets/diff-text.js
@@ -39,7 +39,7 @@ DiffTextWidget.prototype.render = function(parent,nextSibling) {
 	this.execute();
 	// Create the diff
 	var dmpObject = new dmp.diff_match_patch(),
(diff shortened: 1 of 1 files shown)


Lioric:
Hi @Jermolene 

In the first version of this patch (adf72e3), the second argument of widget.getAttribute() method was used, but later I found that it was still causing a NullInput error on some corner cases. That is the reason for the change in the second version of the patch (ef21c57). Right now I don't remember the exact conditions that still caused the error when used with widget.getAttribute() second parameter.

PR #3670, opened 2018-12-31
Example added to: CountWidget.tid

00SS:
Just a simple example added


00SS:
Thanks!
Changed it to use the much easier `<<wikitext-example-without-html>>` macro.
Should be all right now.

Jermolene:
Hi @00SS apologies can you retarget this PR to the tiddlywiki-com branch? Many thanks!

PR #3674, opened 2018-12-31
Add howto for changing highlight languages

dnebauer:
Add howto to the highlight plugin for changing the set of languages supported by the plugin. These instructions work for TW5 under nodejs -- I'm not sure which other TW configurations the instructions would work for, or whether additional information is required for other TW configurations.

These instructions are basically about downloading a custom highlight.js server (which involved selecting the languages to be supported), taking the `highlight.pack.js` file from the downloaded files, and using it to replace the `highlight.pack.js` file from the TW highlight plugin.


Jermolene:
Great stuff, thanks @dnebauer

PR #3682, opened 2019-01-04
Adding tv-pagetemplate variable

ibnishak:
User can define tv-pagetemplate as a global macro to list his own set of pagetemplates without overriding the core or defining multiple config tiddlers.

### DIFF
--- a/core/ui/PageTemplate.tid
+++ b/core/ui/PageTemplate.tid
@@ -27,7 +27,7 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 
 <$dropzone>
 
-<$list filter="[all[shadows+tiddlers]tag[$:/tags/PageTemplate]!has[draft.of]]" variable="listItem">
+<$list filter="[subfilter<tv-pagetemplate>] ~[all[shadows+tiddlers]tag[$:/tags/PageTemplate]!has[draft.of]]" variable="listItem">
 
 <$transclude tiddler=<<listItem>>/>
 


pmario:
I'm not sure, if this makes sense. You can create a theme to modify the default behaviour. Themes work like plugins and seem to be much cleaner for me. 

If the new variable is set, the whole documentation from tiddlywiki.com is not valid anymore. I think it's not obvious for users, while switching themes is a well known workflow.

pmario:
>Why can't such a mechanism co-exist with themes?

IMO it doesn't need to. ... Your theme can include the PageTemplate, that you suggested. So only those users, who want to use the mechanism need to take the overhead. 

If the variable is part of the core, it will be unused in 98% of the TWs. So it needs to do a full filter run, just to see, that the "else" path is the valid one. ... 

If your theme contains the modified PageTemplate, there is no need for the "else" path, since you are sure that the new content should be used. ... Everyone wins and it will be fast.

ibnishak:
You didn't respond to the major objective. Given that `$:/core/ii/PageTemplate` cannot be included in a theme, how would you solve separate sets of toolbar buttons for separate resolutions?

pmario:
I did a test with my [slant-01 theme](https://wikilabs.github.io/#theme-slant-01:GettingStarted%20theme-slant-01). It works perfectly well. Just include your own core/ui/PageTemplate into the theme and you can switch back and forth between your own settings and the default themes.

pmario:
The linked theme doesn't contain a PageTemplate, but I did test it with the same settings.

ibnishak:
You are right.

PR #3686, opened 2019-01-07
Fix typo, password -> username

kyrias:


00SS:
Good catch!

Jermolene:
Thanks @kyrias

PR #3689, opened 2019-01-09
Add en-US localization

benwebber:
I am working on a plugin that references the wiki language for internal localization. The idea is that switching the wiki language will also translate plugin strings.

I need to include a few US English variants in the plugin, so offering it in the drop-down menu would be very helpful.

I sourced the flag icon from [Wikimedia Commons](https://commons.wikimedia.org/wiki/File:Flag_of_the_United_States.svg).


Jermolene:
Hi @benwebber this is a great idea, many thanks.

PR #3696, opened 2019-01-15
Update TextWidget.tid

00SS:
Used Example display method of `<<wikitext-example-without-html>>` and added another example.


Jermolene:
Thanks @00SS

PR #3704, opened 2019-01-17
Improve command line logging

### LINKED ISSUE #3703: Improve command line logging
It is proposed to make the following improvements to the way that TW5 logs messages to the command line:

* Include a datestamp at the start of each log line
* Include a the process ID within logs from the server component
* Merge the server component messages `Serving on http://127.0.0.1:8080` and `(press ctrl-C to exit)` into a single line

```
2019-01-17T15:08:27.299Z [server:1323-451961067283] Serving on http://127.0.0.1:8080 (press ctrl-C to exit) 
2019-01-17T15:08:28.323Z [syncer-server-filesystem] Dispatching 'save' task: $:/StoryList 
2019-01-17T15:08:28.369Z [filesystem] Saved file /TiddlyWiki5/editions/tw5.com/tiddlers/$__StoryList.tid 
```

The motivation for some of these changes is to make easier to debug when TW5 is running within a proxy web server (e.g. Apache or Internet Information Services).

I'm open to suggestion for other improvements.

Jermolene:
To fix #3703

### DIFF
--- a/core/modules/commands/fetch.js
+++ b/core/modules/commands/fetch.js
@@ -120,7 +120,7 @@ Command.prototype.fetchFile = function(url,options,callback,redirectCount) {
 	        }
 	   	});
 	   	response.on("error",function(e) {
-			console.log("Error on GET request: " + e);
+			self.commander.log("Error on GET request: " + e);
 			callback(e);
 	   	});
 	});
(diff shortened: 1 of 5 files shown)


pmario:
@Jermolene -- Did you forget about this one and #3703

Jermolene:
Thanks @pmario I had indeed forgotten about this one. I've added it to #7856

Jermolene:
I've reverted this PR because it seems to be incomplete – for example, it sets a prefix for the logger, but doesn't actually use it. More fundamentally, it doesn't resolve two significant problems:

* That the commander.js uses a different logging mechanism than the syncer and the server
* That there is no way to turn off the colour codes (we should probably automatically turn them off when outputting to a file rather than an interactive terminal)

Issue #3700, opened 2019-01-17
Draggable actions for draggable links


admls:
Hello everyone,

I'm just thinking the $link widget could take a couple extra attributes--startactions and endactions--to be triggered on drag events just like the $draggable widget has. They'd only get triggered if the draggable attribute is "on". It seems like a low overhead fix and would easily be backwards compatible.

Best wishes,
Adam

Issue #2563, opened 2016-08-29
Feature req: Drag'n drop transclusion


twMat:
Drag'n dropping a *title link*, e.g from sidebar list, into an editor adds the title there like `[[Foo bar]]`. Very nice.

I would like to be able to get it added as `{{Foo bar}}`, as an option. Perhaps by holding down a key during the dragging.

This would make reuse of tiddlers even simpler and seems to me like a shortcut for drag'n drop functionality (...or I'm just ignorant of how complex it would be to add `{{` around the title...).

tobibeer:
Can you describe an actual scenario when you would (want to) use this? ...as in "And this would be my workflow: ..."?

twMat:
Sure. Or at least a merge of workflow description and examples:

**1)**
The simplest example is that one is authoring a tiddler and want's it to transclude another tiddler. Say you're authoring a "member presentation", and you want to drag in member profiles so they show up, rather than just their tiddler names. As you edit the encompassing tiddler, you in the search box type in (the prefix) "Member" and then drag in the desired names... i.e with surrounding transclusions.
Currently when you do this, you get the tiddler links.

**2)**
More fun example: A sidebar "toolbox" of items needing no parameter or only needing implicit currentTiddler as parameter. A simple example is a gallery of images. Workflow;

I'm authoring a tiddler in edit mode. Sidebar tab is showing a listing of images (their titles, perhaps with thumbnail). I "trans-drag" the desired image into my text and continue happily.

(Such a toolbox gallery is desirable for my own use In my fiddling I quite often need core images but their difficult-to-remember names actually makes me go to [a thing](https://tobibeer.github.io/tb5/#Core%20Images%20And%20Inline%20Stylesheets) you've made!)

**3)**
Another gallery example; In a social network (yes, twederation) one could imagine an avatar showing up instead of using `@bob`, i.e drag in the avatar from a "contacts list" or a "current participants list" into the edit-text when addressing someone. Such an avatar, i.e the contents of that tiddler, could be a reveal button showing only the avatar by default but if clicked it expands to show relevant stuff. Note that the tiddler is transcluded into the current one so it can show context specific stuff.

**4)**
[...]

PR #3710, opened 2019-01-19
Draggable links

admls:
Hi @Jermolene,

I added startactions and endactions to the link widget so that its draggable behavior is like the draggable widget. I then changed list-links-draggable macro and the list-tagged-draggable macro so that they would support that new link behavior. As far as I can tell, these changes are entirely backwards compatible.

I also fixed what I saw as a minor formatting issue in list-links-draggable. The position of the closing </$type$> tag was placed before the final droppable-placeholder, causing that final placeholder to be of a different format from the rest of the placeholders in the list. I moved the closing tag to include the trailing placeholder.

Best wishes,
Adam

#3700

### DIFF
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -133,6 +133,8 @@ LinkWidget.prototype.renderLink = function(parent,nextSibling) {
 		$tw.utils.makeDraggable({
 			domNode: domNode,
 			dragTiddlerFn: function() {return self.to;},
+			startActions: self.startActions,
+			endActions: self.endActions,
 			widget: this
 		});
 	}
@@ -177,6 +179,8 @@ LinkWidget.prototype.execute = function() {
 	this.overrideClasses = this.getAttribute("overrideClass");
 	this.tabIndex = this.getAttribute("tabindex");
 	this.draggable = this.getAttribute("draggable","yes");
+	this.startActions = this.getAttribute("startactions");
+	this.endActions = this.getAttribute("endactions");
 	this.linkTag = this.getAttribute("tag","a");
 	// Determine the link characteristics
 	this.isMissing = !this.wiki.tiddlerExists(this.to);
@@ -191,7 +195,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 LinkWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip) {
+	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip || changedTiddlers.tag || changedTiddlers["class"]) {
 		this.refreshSelf();
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @admls apologies for the delayed response.

Adding the startactions and endactions attributes to the link widget is a great idea. I need to check carefully about moving the closing `</$type$>` because I faintly remember that it was intentional, but of course can't recall the details...

admls:
Hi @Jermolene,

Thanks! I think it will allow for some nice ui functionality. I've been looking at links specifically because I've been using Firefox and the draggable buttons don't seem to work there. But it occurs to me the same functionality could probably be extended to buttons with similar ease (for Chrome users at least). I'll take a look at it when I get the chance.

The `</$type$>` tag is for the container element of the list. The list items have a `$subtype$` tag that defaults to `<li>`. That last droppable placeholder is a <div> with no option to change it by passing in any parameters to the macro. So it doesn't end up styled like the other list elements (i.e., not bullet point). But if it is within the `</$type$>` tag, it will end up being in the place you expect (i.e., indented to the same degree as the other list items with a predictable amount of padding between it and the last list item). If it falls outside the `</$type$>` tag it isn't entirely clear that it belongs to the list because of this formatting difference.

twMat:
If I understand this correctly, it would be a partial solution to the request in #2563 i.e to feature drag'n drop translusion, i.e to DnD a link and have it display as `{{title}}` instead of `[[title]]`. I'm guessing it would only be a *partial* solution because the request would ideally be solved with a general way to DnD any link i.e not necessarily ones defined with the full linkwidget.

saqimtiaz:
> I need to check carefully about moving the closing `</$type$>` because I faintly remember that it was intentional, but of course can't recall the details...

@Jermolene it has been a while but I remember editing the macro to move the closing typo as well, as has been done in this PR. Otherwise it is not possible for the final droppable placeholder to consistently styled with the other placeholders.

Jermolene:
Thanks @saqimtiaz, I'll commit that change separately.

saqimtiaz:
@Jermolene there is merit to the actual purpose of this PR, which is to **add drag start and drag end actions** to the LinkWidget akin to the DraggableWidget. 

Do we want to re-do this PR to resolve conflicts and implement that? If so I can try and look at it in the coming week.

Jermolene:
> @Jermolene there is merit to the actual purpose of this PR, which is to **add drag start and drag end actions** to the LinkWidget akin to the DraggableWidget.

> Do we want to re-do this PR to resolve conflicts and implement that? If so I can try and look at it in the coming week.

Indeed, I just dealt with the typo to put it to bed, I'd still be interested in the rest of this PR, probably for v5.1.24.

saqimtiaz:
bump

Jermolene:
Thanks @theSherwood @saqimtiaz there's some conflicts to be resolved, but I also noticed that your signature on the contributors license agreement is under a different username (@admls), which would be worth updating with another PR to tiddlywiki-com.

joshuafontany:
@Jermolene @theSherwood @saqimtiaz This seems worth including in an upcoming release. I may have time to resolve these conflicts this week. Thoughts?

saqimtiaz:
@joshuafontany I would welcome that.

pmario:
@theSherwood ... Are you still with us, to resolve the conflicts?

theSherwood:
@pmario Apologies, but no. Unfortunately, I haven't used TiddlyWiki in some years and no longer have the context on any of this.

Feel free to close this or take it over!

pmario:
- Closed in favour of https://github.com/TiddlyWiki/TiddlyWiki5/pull/9621

PR #3712, opened 2019-01-20
Implement and document a shuffle filter operator.

mklauber:
A shuffle operator randomizes the order of tiddlers in the input list.  It can also be used with the `first[]` operator to select a random subset of the input list.

### DIFF
--- a/core/modules/filters/shuffle.js
+++ b/core/modules/filters/shuffle.js
@@ -50,15 +50,14 @@ A filter operator to randomize the order of the list passed to it.
     };
 
     exports.shuffle = function(source, operator, options) {
-      var results = prepare_results(source);
-      if (operator['operand'] == "") {
-        if (options.wiki.tiddlerExists("$:/temp/seed") {
-          var tiddler = options.getTiddler("$:/temp/seed")
-          shuffle(results, tiddler.getFieldString("startup"));
-         }
-      } else {
-        shuffle(results, hash(operator['operand']));
-      }
-      return results;
+       var results = prepare_results(source);
+       if (operator['operand'] !== "") {
+         shuffle(results, hash(operator['operand']));
+       } else if (options.wiki.tiddlerExists("$:/temp/seed")) {
+         shuffle(results, options.wiki.getTiddler("$:/temp/seed").getFieldString("startup"));
+       } else {
+         shuffle(results, "");
+       }
+       return results;
     }
 })();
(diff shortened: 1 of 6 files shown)


mklauber:
Understood.  The desired behavior should be default the same shuffle every time?  `[enlist[1 2 3 4 5]shuffle[]]` always returns `4 2 3 5 1` for example?  I can default to a hard coded seed.  

What do you think of having a hard coded value `[shuffle[randomize]]` that produces a different output each time.  I can think of some use cases where you may want that to be the case.  That way you're explicit about wanting the performance intensive behavior?  

Unless there's a decent way to generate and pass a random value to the filter?  I've been using a button setting the state tiddler to `<now "YYYY0MM0DD0hh0mm0ssXXX">` which is roughly equivalent to a new random seed every time the button is pressed, but I can't pass that directly as a filter parameter.

Jermolene:
Hi @mklauber apologies for the delayed response.

> Understood.  The desired behavior should be default the same shuffle every time?  `[enlist[1 2 3 4 5]shuffle[]]` always returns `4 2 3 5 1` for example?  I can default to a hard coded seed.  

Yes, that's right. Every time a filter is run against a given set of tiddlers in the store, the result should be identical.

> What do you think of having a hard coded value `[shuffle[randomize]]` that produces a different output each time.  I can think of some use cases where you may want that to be the case.  That way you're explicit about wanting the performance intensive behavior?  

I think that should be plugin territory.

> Unless there's a decent way to generate and pass a random value to the filter?  I've been using a button setting the state tiddler to `<now "YYYY0MM0DD0hh0mm0ssXXX">` which is roughly equivalent to a new random seed every time the button is pressed, but I can't pass that directly as a filter parameter.

No, and the filter can't directly alter the tiddler store. So I fear that all we can offer is a hash function...

mklauber:
I've updated this to always be deterministic.  I've created a file core/modules/startup/seedrandom.tid that uses $:/tags/startupAction to set a seed based on the time at startup.

Jermolene:
Thanks @mklauber. I'll give some detailed comments when I have more time, but a couple of top level thoughts:

* I'd prefer to avoid bringing in another external library if not strictly necessary. I think the existing Stanford JavaScript Library already has methods for generating randomness which we should reuse if we can
* Rather than a startup actions tiddler to initialise the seed, I think it is more robust to have a startup module do so. If it's just a line or two we could put it in one of the existing miscellaneous startup modules

PR #3742, opened 2019-01-31
add findParentTransclusionWidget prototype for unique widget-ID's

BurningTreeC:
this prototype returns the parent widget (if present) that changes the `transclusion` variable
this allows us to push widgets to the `transclusionWidgets` array of the `parentTransclusionWidget` , so that we can iterate over that array to get the index of our widget within the same `transclusion-level`

this can be useful when we need unique id's for widgets but there are more widgets with a similar "footprint" within the same "transclusion-level", where `widget.getStateQualifier` also returns the same value for all of them. adding the index we get by iterating over the `transclusionWidgets` array, detecting the current widget within it and adding its index within that array to our ID makes the ID unique

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -42,6 +42,7 @@ Widget.prototype.initialise = function(parseTreeNode,options) {
 	this.document = options.document;
 	this.attributes = {};
 	this.children = [];
+	this.transclusionWidgets = [];
 	this.domNodes = [];
 	this.eventListeners = {};
 	// Hashmap of the widget classes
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC 

> this prototype returns the parent widget (if present) that changes the `transclusion` variable
this allows us to push widgets to the `transclusionWidgets` array of the `parentTransclusionWidget` , so that we can iterate over that array to get the index of our widget within the same `transclusion-level`

I think it's problematic for a widget to reach back up and change the state of an ancestor. It should be the other way around: a widget inherits a token from it's immediate parent and then uniqufies it somehow to identify which child it is.

BurningTreeC:
>I think it's problematic for a widget to reach back up and change the state of an ancestor. It should be the other way around: a widget inherits a token from it's immediate parent and then uniqufies it somehow to identify which child it is.

But is the prototype ok to identify the parent transclusion widget?

BurningTreeC:
If I understand it right, it's something like the findNextSiblingDomNode prototype but iterating over all the child widgets, counting them and returning the count for the widget that asks for it ... ?

BurningTreeC:
a widget could use findParentTranscludeWidget to identify it, then get the index of itself in the children of its parentWidget, if that isn't the parent-transclude-widget it adds the index of its parent-widget in the children of the parent-widget's parent-widget... until the parent widget is the parentTranscludeWidget ...

that would create a unique identifier for all widgets at the same transclusion level

BurningTreeC:
I'm reopening this PR because of #3747

BurningTreeC:
I believe I can close this one again, #3744 does the trick

PR #3744, opened 2019-02-01
widget.js : generate a unique widget ID

BurningTreeC:
this try on generating unique widget ID's does the following:

it generates a "footprint" of the form `0130` where the first digit is the index of the current widget within the array of its parentWidget's children

the next digit is the index of its parentWidget within the array of that widget's parentWidget's children, and so on, until the first widget that creates a transclusion is reached.

this generates a unique footprint among different widgets that are all at the same transclusion levels

adding this footprint to the state qualifier we get through `this.getStateQualifier` by `this.getStateQualifier + "-" + this.generateTransclusionFootprint` creates unique widget-ID's

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -557,6 +557,25 @@ Widget.prototype.allowActionPropagation = function() {
 	return true;
 };
 
+Widget.prototype.generateTransclusionFootprint = function() {
+	var parentTransclusionWidget = this.findParentTransclusionWidget(),
+	    node = this,
+	    footprint = node.parentWidget.children.indexOf(node);
+	while(node = node.parentWidget && node !== parentTransclusionWidget) {
+		footprint += "" + node.parentWidget.children.indexOf(node);
+	}
+	return footprint;
+};
+
+Widget.prototype.findParentTransclusionWidget = function() {
+	var node = this;
+	var transclusionVariable = this.getVariable("transclusion");
+	while(node.getVariable("transclusion") === transclusionVariable) {
+		node = node.parentWidget;
+	}
+	return node !== this ? node : null;
+};
+
 exports.widget = Widget;
 
 })();
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -561,8 +561,12 @@ Widget.prototype.generateTransclusionFootprint = function() {
 	var parentTransclusionWidget = this.findParentTransclusionWidget(),
 	    node = this,
 	    footprint = node.parentWidget.children.indexOf(node);
-	while(node = node.parentWidget && node !== parentTransclusionWidget) {
-		footprint += "" + node.parentWidget.children.indexOf(node);
+	while(node = node.parentWidget) {
+		if(node === parentTransclusionWidget) {
+			break;
+		} else if(node.parentWidget && node.parentWidget.children) {
+			footprint = footprint + "" + node.parentWidget.children.indexOf(node);
+		}
 	}
 	return footprint;
 };
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -565,7 +565,7 @@ Widget.prototype.generateTransclusionFootprint = function() {
 		if(node === parentTransclusionWidget) {
 			break;
 		} else if(node.parentWidget && node.parentWidget.children) {
-			footprint = footprint + "" + node.parentWidget.children.indexOf(node);
+			footprint = footprint + "-" + node.parentWidget.children.indexOf(node);
 		}
 	}
 	return footprint;
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -561,7 +561,8 @@ Widget.prototype.generateTransclusionFootprint = function() {
 	var parentTransclusionWidget = this.findParentTransclusionWidget(),
 	    node = this,
 	    footprint = node.parentWidget.children.indexOf(node);
-	while(node = node.parentWidget) {
+	while(node) {
+		node = node.parentWidget;
 		if(node === parentTransclusionWidget) {
 			break;
 		} else if(node.parentWidget && node.parentWidget.children) {


pmario:
I'm a bit confused. ... If it is 1 digit, the index can be 0..9. What if it has more then 10 siblings? So the ID is unpredictable for users?

If that's true, I think it may be a problem. I think we need a possibility, to show the "footprint" value. Similar to the [transclusion variable](https://tiddlywiki.com/#transclusion%20Variable) used to with the `<<qualify>>` macro.

just some thoughts.

core/modules/widgets/widget.js
```diff
@@ -557,6 +557,29 @@ Widget.prototype.allowActionPropagation = function() {
...
 
+Widget.prototype.generateTransclusionFootprint = function() {
+	var parentTransclusionWidget = this.findParentTransclusionWidget(),
+	    node = this,
+	    footprint = node.parentWidget.children.indexOf(node);
+	while(node = node.parentWidget) {
+		if(node === parentTransclusionWidget) {
+			break;
+		} else if(node.parentWidget && node.parentWidget.children) {
+			footprint = footprint + "" + node.parentWidget.children.indexOf(node);
```
pmario: I'm not sure if the `""` empty string is only there to "force" a string conversion, or if it was `"-"` which I think I'd prefer.

BurningTreeC: Hi @pmario , it's there just to force string conversion, but thinking about what you mentioned above about when there are more than 10 siblings, I also want it to be "-" or another character.
I thought I'd use "-" to concatenate the state-qualifier with this string, that's why here a different character would be good, what do you think?

BurningTreeC: yes, we can also just use "-" and create the id by `this.getStateQualifier() + "_" + this.generateTransclusionFootprint();` , then it will look like `728119345_0-12-4-0-0`

pmario: I'm ok with this structure

core/modules/widgets/widget.js
```diff
@@ -557,6 +557,29 @@ Widget.prototype.allowActionPropagation = function() {
 	return true;
 };
 
+Widget.prototype.generateTransclusionFootprint = function() {
+	var parentTransclusionWidget = this.findParentTransclusionWidget(),
+	    node = this,
+	    footprint = node.parentWidget.children.indexOf(node);
+	while(node = node.parentWidget) {
```
Jermolene: I'm not keen on putting assignments within conditionals, as it's too easy to read it as a mistaken equality comparision. Is it intentional here?

core/modules/widgets/widget.js
```diff
@@ -557,6 +557,29 @@ Widget.prototype.allowActionPropagation = function() {
...
+		if(node === parentTransclusionWidget) {
+			break;
+		} else if(node.parentWidget && node.parentWidget.children) {
+			footprint = footprint + "-" + node.parentWidget.children.indexOf(node);
```
Jermolene: It looks like you're using the index of the widget amongst it's siblings as an additional qualifier. Doesn't that mean that the IDs will change if, say, we have a sequence of 10 input elements and then insert a new one at the top?

core/modules/widgets/widget.js
```diff
@@ -557,6 +557,29 @@ Widget.prototype.allowActionPropagation = function() {
...
+		}
+	}
+	return footprint;
+};
+
+Widget.prototype.findParentTransclusionWidget = function() {
+	var node = this;
+	var transclusionVariable = this.getVariable("transclusion");
+	while(node.getVariable("transclusion") === transclusionVariable) {
```
Jermolene: Variables that are undefined within a widget are passed to the parent widget for resolution. Thus, when you're walking up a stack of ancestors, the first time you get the chosen transclusion value won't be the widget in which it is defined, but rather the deepest descendent widget that doesn't redefine that value.

BurningTreeC: ah, I see. so it works by accident I guess ... I need to think this through, haven't thought about that yet

BurningTreeC:
@Jermolene , thanks for your comments. As you see, this is just an attempt in generating unique ID's ... it's limited by my level of insight to the whole inner mechanics.
Anyways, I'm keen to get some kind of method working and would be glad about some instructions

Thanks!

Jermolene:
Hi @BurningTreeC just thinking aloud here, but I wonder if it might be beneficial to think up a systematic testing procedure. I think there's maybe two requirements:

* That we cannot find a widget tree where the input elements within it do not generate unique IDs
* That we cannot mutate any of those widget trees such that the surviving input elements  get assigned different IDs, or that newly added input elements get assigned non-unique IDs

It would be tough to build a test rig for the entire focus mechanism (we'd probably need to use PhantomJS or Puppeteer under Node.js), but we should be able to build a test rig for getting the ID generation right...

core/modules/widgets/widget.js
BurningTreeC: @Jermolene - here I'm walking up the stack of ancestors until the `parentTransclusionWidget` is reached. that's the one that defines the transclusion value (?) ... the while loop is for detecting all ancestor widgets up to the `parentTransclusionWidget` where the transclusion variable is the same

BurningTreeC: ... hm, does all this mean that I'm one widget off?

BurningTreeC: hm... does this mean that I'm one widget off?

BurningTreeC:
Hi @Jermolene ,

if in a widget tree each widget has a different path than any other widget in that given tree then the id's will be unique. They (should, if my idea works correctly) retrace the path from the first ancestor widget that changes the transclusion variable down to the widget itself. as long as `getStateQualifier()` does always generate different hash-strings for different "transclusion levels", the combination of `getStateQualifier()` and `getTransclusionFootprint()` is unique at any point in the widget tree, or I misunderstand how this is working, that's always possible

BurningTreeC:
>That we cannot mutate any of those widget trees such that the surviving input elements get assigned different IDs, or that newly added input elements get assigned non-unique IDs

with this method the input elements get different IDs when the widget tree changes, there's no other way with this approach. I don't see the need that those IDs should be the same all the time, they only need to be unique (?)

pmario:
>... I don't see the need that those IDs should be the same all the time, they only need to be unique (?)

I think that's right.

admls:
Would/could the widget id be added to the dom node it generates in some way? Maybe as part of the base widget renderdomnodes method? It would be wonderful to be able to get access to the widget tree from the dom.

Best wishes

pmario:
@admls not every widget creates DOM nodes. eg: `<$transclusion />`

admls:
@pmario Sure, but if those that do were to pass their id to the renderDomNodes function, and that function were to add a data attribute with the id to the dom node(s) it creates, that could be very convenient. Being able to find where you are in the widget tree from the dom side of things could be very handy.

BurningTreeC:
I'm reopening this PR because of #3747

core/modules/widgets/widget.js
BurningTreeC: @Jermolene, I think this works ok

core/modules/widgets/widget.js
BurningTreeC: @Jermolene, I think this works fine, if by accident or not I don't know :smile:

BurningTreeC:
[This demo page](http://refocusing-inputs.tiddlyspot.com/) shows a state tiddler that stores the ID of the currently focused input. It's unique for each input in the DOM tree

BurningTreeC:
closing in favor of #4113

PR #3745, opened 2019-02-01
fix bug in reveal widget when stateTitle tiddler is missing

BurningTreeC:
in the group this bug was reported

when the stateTitle is missing it tries to get a field string of an undefined tiddler ...


pmario:
see: https://groups.google.com/d/msg/tiddlywiki/l50vIN7EJlk/LJKPzy1DHgAJ

Jermolene:
Thanks @BurningTreeC

PR #3746, opened 2019-02-02
add inputmanager module

BurningTreeC:
this stores informations about inputs using a unique id

id
selectionStart
selectionEnd
shouldFocusAgain (input had focus and $:/state/current-focus has the input-id in its text field)


core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+function InputManager(options) {
+	options = options || "";
```
pmario: Is options a string or an object? IMO it needs to be `|| {}`

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+
+Input handling utilities
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
```
pmario: Comment block with general info is missing.

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+function InputManager(options) {
+	options = options || "";
+	this.inputs = [];
+}
+
+InputManager.prototype.findInputById = function(identifier) {
+	var result;
+	for(var i=0; i<this.inputs.length; i++) {
+		var object = this.inputs[i];
```
pmario: why is it named `object` ... is there a better name? .... Is it an html `element` ?

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+		return false;
+	}
+};
+
+InputManager.prototype.setValue = function(identifier,name,value) {
+	var object = this.findInputById(identifier);
+	if(object) {
+		object[name] = value;
+	} else {
+		var options = {
```
pmario: `options` ... I'm confused. IMO this is different to the first `options` from `function InputManager(options) {` .. right?

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+			result = object;
+			break;
+		}
+	}
+	return result;
+};
+
+InputManager.prototype.shouldFocusAgain = function(identifier) {
+	var object = this.findInputById(identifier);
+	if(object && object.shouldFocusAgain && $tw.wiki.getTiddlerText("$:/state/current-focus") === identifier) {
```
pmario: hmm. `$tw.wiki` is the global wiki instance. ... imo it should be like this: `this.wiki.getTiddlerText(STATE_CURRENT_FOCUS)`

core/modules/inputmanager.js
pmario: I'd add: 

```
this.wiki = options.wiki || $tw.wiki;
```

see comment below!

core/modules/inputmanager.js
pmario: IMO there should be a global variable, that contains the tiddler name. It makes it easier to find typos.

```
var STATE_CURRENT_FOCUS = "$:/state/current-focus"
```

it's used in comments below.

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+		var options = {
+			id: identifier
+		};
+		options[name] = value;
+		this.inputs.push(options);
+	}
+};
+
+InputManager.prototype.getSelections = function(identifier) {
+	var object = this.findInputById(identifier);
```
pmario: `object` ... for me that's a confusing var name.

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+		var selEnd = object.selectionEnd;
+		if(selStart && selEnd) {
+			return {
+				selectionStart: selStart,
+				selectionEnd: selEnd
+			};
+		} else {
+			return null;
+		}
+	}
```
pmario: what if the `object` doesn't exist. ... IMO else path is missing.

core/modules/inputmanager.js
```diff
@@ -0,0 +1,72 @@
...
+		this.inputs.push(options);
+	}
+};
+
+InputManager.prototype.getSelections = function(identifier) {
+	var object = this.findInputById(identifier);
+	if(object) {
+		var selStart = object.selectionStart;
+		var selEnd = object.selectionEnd;
+		if(selStart && selEnd) {
```
pmario: Are the above 3 lines necessary?

core/modules/inputmanager.js
BurningTreeC: I just copied this over from our keyboardManager

core/modules/inputmanager.js
BurningTreeC: no it's just an object containing 
```
{
id: uniqueID,
selectionStart: value,
selectionEnd: value,
shoudFocusAgain: boolean
}
```

I think inputInfo is a good name

core/modules/inputmanager.js
BurningTreeC: correct. I understand the need to rename variables in order for them to be more self-explanatory

core/modules/inputmanager.js
BurningTreeC: no not necessarily

BurningTreeC:
Using this I already have inputs working that edit the current Tiddler

pmario:
So the only thing missing is a little bit more docs that goes with the code.

pmario:
Especially what return values are expected. eg: undefined or null ... or what ever is expected in success case or error case.

pmario:
Nice!

BurningTreeC:
reopening because of #3747

BurningTreeC:
I'm inclined to close this in favor of a new, cleaner PR

BurningTreeC:
closing in favor of #4113

PR #3747, opened 2019-02-02
focusing inputs after focus left

BurningTreeC:
this PR uses #3744 and #3746 to create unique widget-id's in factory.js and update the previously focused input with the previous selections

this makes also editing the current tiddler work, with some performance caveats that should probably be solved by storing and restoring widget instances, if that's possible


pmario:
You used "should" in your comment. Did you test it? Does it work for you?

BurningTreeC:
>You used "should" in your comment. Did you test it? Does it work for you?

I always test. It works for me but at the point I wrote this there was a small bug that's resolved now

Still, it may have some side effects because it doesn't limit all possible cases when it can refocus. Some conditions when to refocus and when not still have to be defined

BurningTreeC:
I've put a current empty wiki with my three pull requests here: refocusing-inputs.tiddlyspot.com

there's a small test scenario

BurningTreeC:
in the demo page, you may also want to try showing/hiding the editor preview while the editor has focus

there you'll notice a case that's not handled when you click outside the editor (not in another input field) so that the editor looses focus and then change the preview state, the editor will have focus again

pmario:
I did a refresh test and it seems to work well. Very good work! Excellent!

pmario:
There is a refresh problem. ... to reproduce:

 - open 2 new tiddlers, that you can see both edit areas
 - Firefox F12
 - settings: Available Toolbox Buttons: check: Toggle paint flashing
 - activate paint flashing
 - make debug area small. so you can see both edit areas
 - type some text. after 400ms both  ediot areas are repainted.
    - this behaviour is different in 5.1.19

pmario:
above comment was updated

BurningTreeC:
Hi @pmario ,

following your instructions step by step, I cannot see any repainting. where do I have to look for it?

BurningTreeC:
Another thing:

Testing on my smartphone I notice the performance-cost when editing the current tiddler
Typing fast does not work because after each character the widget makes a full refresh (like it did before) ... the selection then is not restored fast enough to insert the next character if it was pressed during the refresh

I believe we need something like a "light" refresh for this scenario

BurningTreeC:
>I believe we need something like a "light" refresh for this scenario

thinking about it, what causes the full refresh/re-rendering when we're editing the current tiddler?

BurningTreeC:
I thought maybe it's possible to save the widget instance when a certain event occurs (the inputManager can be used for that) and restore it afterwards instead of a full refresh

pmario:
the paint flash setting only enables a new toolbar button, which needs to be activated. It shows the elements on the page, that are drawn in different colors

pmario:
did you update the demo on tiddlyspot?

BurningTreeC:
>the paint flash setting only enables a new toolbar button, which needs to be activated. It shows the elements on the page, that are drawn in different colors

I did that, I see flashing elements when they flash, they're colored. But the text areas or inputs don't flash

>did you update the demo on tiddlyspot?

yes, with the latest update here, the double-focus fix

BurningTreeC:
I've updated the demo with the latest fixes:

* correct getSelections() like @pmario proposed
* correct order of when to set focus, after that selectionRange is set
* fix the mentioned "quirk" when selection starts with `0`

pmario:
>There is a refresh problem. ... to 

This problem is gone now!

pmario:
If I open this: http://refocusing-inputs.tiddlyspot.com/#test%20input%20fields:%5B%5Btest%20input%20fields%5D%5D%20%24%3A%2Fstate%2Fcurrent-focus and

 - F12
 - activate paint flashing
 - edit: "another tiddler" the `$:/state/current-focus` tiddler is only updated when the edit field is selected -> OK
 - if I edit: "current tiddler, field test-field" then the `$:/state/current-focus` is redrawn with every key-stroke. .... -> strange behaviour. I didn't expect this. ... There seems to be a "state-write" action, even if the last input area is the same as the new one. ... Is it possible to check this? and leave the state tiddler alone if the value doesn't change?

pmario:
I'm not 100% sure, why we need to remember the selected x/y values. Is there a special reason?

BurningTreeC:
>This problem is gone now!

That's good :+1: 

>I didn't expect this. ... There seems to be a "state-write" action, even if the last input area is the same as the new one. ... Is it possible to check this? and leave the state tiddler alone if the value doesn't change?

Ah yes, it should check if `$:/state/current-focus` already is the current id ... it just overwrites it all the time at the moment

>I'm not 100% sure, why we need to remember the selected x/y values. Is there a special reason?

make a selection in the text editor and toggle the preview -> the selection is kept which is what I'd expect ... if we don't save selections it selects the whole text again and again

this could also make the framed editor not needed anymore (?) ... if I recall correctly, because it's framed so that selections are kept when clicking a toolbar button for a text operation

BurningTreeC:
I've made `$:/state/current-focus` update only if it has changed,
updated some comments in the inputmanager

the demo is updated, too

BurningTreeC:
@pmario, now I got how to make those inspections. I see what you mean now. This repainting should be solved by updating $:/state/current-focus correctly, not with every refresh

BurningTreeC:
... also on my smartphone it's much better now

BurningTreeC:
>This would be nice to explore. The iframe adds a lot of weight to the internal memory usage, since it basically opens a new browser-view-instance.

the framed engine in the demo already updates the selections stored in the global arrays with the new selections of the various text operations ... so I think it's only a matter of removing the iframe and making it a standard textarea

BurningTreeC:
the framed editor is replaced by a normal textarea (in the demo only)

pmario:
There is a little problem, if you type fast and then change the cursor position. After 400ms it returns the input cursor back to the last typing position.

pmario:
The easiest way to see it is:

 - insert some "bla bla" test [enter]
 - more text [enter]
 - cursor up
 - cursor jumps down -> problem

BurningTreeC:
@pmario, thanks! I'm about to look for additional cases when selections need to be updated. Moving the cursor is another one. I've already added onselect

BurningTreeC:
the demo is updated again, with the latest fix:

when updating the editor domNode in factory.js, before updating it, selections need to be updated


(removed iframeless editor from the demo)

BurningTreeC:
for testing, the demo on http://refocusing-inputs.tiddlyspot.com/ again has a frameless editor and the wrap-selection text-operation wraps at the start of tiddlers when there's no text

pmario:
Now the problem seems to be back, that if you open "New Tiddler" and edit one of the text inputs, the "New Tiddler" edit area is redrawn.

BurningTreeC:
@pmario, I'll have a look what may have caused this

BurningTreeC:
@pmario ,

I've tried a standard textarea on tiddlywiki.com by

```
<$edit-text tag="textarea" tiddler="sometiddler" default="" placeholder=""/>
```

such a textarea behaves the same way. there's nothing in the additional code that causes this, just the missing iframe in the demo which is removed for testing purposes only, to show that the editor buttons work correctly. I believe that this redraw is a small one but it's there

pmario:
That's funny, I did a test with the latest version at tiddlyspot. No problem found. .... May be I did my last test with something loaded from the browser cache. ... Seems to be my fault.

pmario:
@BurningTreeC ... Why did you close this and the related PRs? ??

twMat:
@BurningTreeC - I'm also curious as to why you closed this? Was there some error?

BurningTreeC:
reopening this PR because of a recent [discussion in the google dev group](https://groups.google.com/forum/#!topic/tiddlywikidev/A2JM-qyIFeg)

BurningTreeC:
[Here](http://refocusing-inputs.tiddlyspot.com/)'s a Demo of the functionality

BurningTreeC:
Note that [in the Demo](http://refocusing-inputs.tiddlyspot.com/) the text editor also doesn't loose focus if clicking on the preview button - to test, insert some text, make a selection and click the button / press the shortcut alt-P ... this works for the framed engine and the codemirror engine alike

BurningTreeC:
I'm inclined to close this again in favor of a new, clean PR

BurningTreeC:
closing in favor of #4113

PR #3749, opened 2019-02-04
update wrap-selection text-operation

BurningTreeC:
this is only a proposal and I make this because I don't understand why this doesn't remove a prefix and suffix wrapped around the cursor only when it's at the start of the editor and there's no other text

in all other cases it does remove them, so it comes totally unexpected that only in the case when there are only the given prefix and suffix the text-operation doesn't remove them

### DIFF
--- a/core/modules/editor/operations/text/wrap-selection.js
+++ b/core/modules/editor/operations/text/wrap-selection.js
@@ -16,14 +16,12 @@ exports["wrap-selection"] = function(event,operation) {
 	if(operation.selStart === operation.selEnd) {
 		// No selection; check if we're within the prefix/suffix
 		if(operation.text.substring(operation.selStart - event.paramObject.prefix.length,operation.selStart + event.paramObject.suffix.length) === event.paramObject.prefix + event.paramObject.suffix) {
-			// Remove the prefix and suffix unless they comprise the entire text
-			if(operation.selStart > event.paramObject.prefix.length || (operation.selEnd + event.paramObject.suffix.length) < operation.text.length ) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC maddeningly, I'm afraid I can't definitively recall the logic here, but I think it may be due to limitations in the way that browsers implement `execCommand`. I did find the commit where it was introduced: 04ec2cee27e5005ccccc2ee4be514e8f192f8acc

BurningTreeC:
hm, I cannot find anything about such a problem... if it's a problem with internet explorer or some older browser versions I understand that it's reasonable to keep it. testing on recent chromes and firefoxes doesn't reveal any problems

I'll let this PR open, maybe at some point we can bring light into this :smile:

Jermolene:
Hi @BurningTreeC my dim recollection is that the problem was generic, and not browser specific. But, I wonder if we might not be better off going ahead and merging this and see what gets reported...

BurningTreeC:
Hi @Jermolene , I wouldn't mind if we do so, there are no problems with it in **my** browsers

BurningTreeC:
Hi @Jermolene, removing this limitation is fine for the standard editor, I haven't encountered any problem with it. With CodeMirror there seems to be a problem, but the "solution" removed here fights only the symptom I believe. At the start of the first line the CodeMirror editor sometimes reports negative selectionStart and selectionEnd values which is strange ... it appears as if that's the reason why cm behaves weird then. I'm not 100% sure but I believe we were right removing the limitation and should investigate what the problem is

Jermolene:
Thanks @BurningTreeC that makes sense.

PR #3751, opened 2019-02-05
a filter that creates a hash-string from its input titles

BurningTreeC:
this filter takes a list of titles and creates a hashstring from them, like the qualify macro does.
it uses the same function `$tw.utils.hashString`

it's useful for example when feeding it with a list of transclusion variables that model the path to a point at the page, to get the `<<qualify>>` value at that point

### DIFF
--- a/core/modules/filters/qfp.js
+++ b/core/modules/filters/qfp.js
@@ -1,0 +1,40 @@
+/*\
+title: $:/core/modules/filters/qfp.js
+type: application/javascript
+module-type: filteroperator
+
+Filter operator that returns a "qualified" hashstring for a list of input titles
+Example feeding it with a list of transclusion variables that model the path
+from a given point up to the page template (returns the same qualified string
+like the <<qualify>> macro would do at that point):
+
+filter="{|$:/core/ui/PageTemplate/story|||} {|$:/core/ui/PageTemplate|||} +[qfp[]]"
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports.qfp = function(source,operator,options) {
+	var titles = [],
+	results = [],
+	hashString;
+	source(function(tiddler,title) {
+			titles.push(title);
+	});
+	if(titles.length) {
+		hashString = $tw.utils.hashString(titles.join("")).toString();
+		while(hashString.length < 11) {
+			hashString = "-" + hashString;
+		}
+		results.push(hashString);
+	}
+return results;
+};
+
+})();
--- a/core/modules/filters/numhash.js
+++ b/core/modules/filters/numhash.js
@@ -28,11 +28,7 @@ exports.numhash = function(source,operator,options) {
 			titles.push(title);
 	});
 	if(titles.length) {
-		hashString = $tw.utils.hashString(titles.join("")).toString();
-		while(hashString.length < 11) {
-			hashString = "-" + hashString;
-		}
-		results.push(hashString);
+		results.push($tw.utils.hashString(titles.join("")).toString());
 	}
 return results;
 };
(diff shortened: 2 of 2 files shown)


pmario:
Where do you want to use this?

pmario:
With my "slant-01" theme I have a hard time to remove the `<<qualify>>` makro, with predictable strings. ... I want to be able to import "state" tiddlers and define the application state that way. So I can be sure that the input marker is exactly, where I want it to be. Hashes are completely unpredictable, if you don't know the input string.

Jermolene:
Hi @pmario it sounds like your objection is more fundamental in that you want to be able to work with wikis that don't use qualified state tiddler titles. I think @BurningTreeC is trying to satisfy a simple need that I have also found: to be able to tidy up overly long state tiddler titles without using the qualify macro.

Having said that @BurningTreeC I'm wary of merging changes piecemeal that are part of the general exploration you're doing into preserving the DOM state over refresh cycles. Is this filter worth merging now if it wasn't for that work? I've some minor code comments too.

core/modules/filters/qfp.js
```diff
@@ -0,0 +1,40 @@
+/*\
+title: $:/core/modules/filters/qfp.js
```
Jermolene: I'm not a huge fan of acronymic filter operators. `hash` doesn't seem right because this is only one of a number of possibly useful hashing algorithms we might want to expose. So maybe `numhash`?

core/modules/filters/qfp.js
```diff
@@ -0,0 +1,40 @@
...
+exports.qfp = function(source,operator,options) {
+	var titles = [],
+	results = [],
+	hashString;
+	source(function(tiddler,title) {
+			titles.push(title);
+	});
+	if(titles.length) {
+		hashString = $tw.utils.hashString(titles.join("")).toString();
+		while(hashString.length < 11) {
```
Jermolene: Presumably the padding is to make sure that every hash is the same length? Any reason for padding with dashes?

core/modules/filters/qfp.js
BurningTreeC: yes, numhash sounds good!

core/modules/filters/qfp.js
BurningTreeC: the reason for padding with dashes is to get the same string that the `<<qualify>>` macro would produce

BurningTreeC:
Hi, I have to say something about my reason for this. It's not about preserving the DOM state over refresh cycles

It has more to do with getting the tiddler title associated with a qualified hashstring, which allows to do something with that tiddler if its qualified state-tiddler is present, like fixing the sticky-titles popup problems for the more-popup and all the dropdowns in edit-mode

Jermolene:
Thanks @BurningTreeC would you like to go ahead and add documentation?

BurningTreeC:
Thank you, @Jermolene ... I've made a separate PR for the documentation, I forgot I could just add it here

Jermolene:
Thanks @BurningTreeC

BurningTreeC:
Hi @Jermolene , @pmario and all others interested ,

this filter would allow to set a `qualified` state from outside its scope, with a button, a keyboard shortcut (...)

or it would just allow showing the name of the tiddler associated with a qualified state

I understand that a filter should be used in the core to justify its existence as a core filter
in the numhash case, now that the "sticky titles" problem is solved, I think this filter could still be useful, we just need to figure out some usecases

* it could be used to change the active tab in the sidebar
* changing active tab of the tiddler with the tiddler-info open
* opening popups of tiddlers, if there would be a "selected-tiddler" mechanism
* ...

Jermolene:
Hi @BurningTreeC with the immediate need satisfied, I'd be inclined to put this on hold for the moment, pending some of the other discussions we're having.

BurningTreeC:
Ok @Jermolene

PR #3756, opened 2019-02-08
adding optional tabindex attr to simple, framed and cm engines

BurningTreeC:
we could add a configuration option in the control panel to enable a tabindex on the input and textarea fields of the edit template, for example


Jermolene:
Thanks @BurningTreeC

PR #3758, opened 2019-02-08
adding tabindex attributes to edit templates

BurningTreeC:
... in the tags edittemplate, before calling the tag-picker macro, I set the tabIndex variable to `{{$:/config/EditTabIndex}}`, so that when reusing the tagpicker elsewhere, the variable can be set differently

### DIFF
--- a/core/ui/EditTemplate/body-editor.tid
+++ b/core/ui/EditTemplate/body-editor.tid
@@ -5,6 +5,7 @@ title: $:/core/ui/EditTemplate/body/editor
   field="text"
   class="tc-edit-texteditor"
   placeholder={{$:/language/EditTemplate/Body/Placeholder}}
+  tabindex={{$:/config/EditTabIndex}}
 
 ><$set
 
--- a/core/ui/EditTemplate/body.tid
+++ b/core/ui/EditTemplate/body.tid
@@ -13,7 +13,7 @@ $:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
 
 <a href={{!!_canonical_uri}}><$text text={{!!_canonical_uri}}/></a>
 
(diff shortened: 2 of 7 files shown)


BurningTreeC:
#3814 is newer, forgot about this one

this one here does it in a more elegant way

Jermolene:
Does #3814 need closing? And is this ready to merge?

BurningTreeC:
@Jermolene - yes #3814 can be closed, this is better

yet the configuration options are not there, but I'd like to make them when I'm giving the codemirror ui config things a try. documentation for the attribute then follows, too

Jermolene:
Hi @BurningTreeC forgive the stupid question but why are we assigning the same value to every edit-text widget? Shouldn't the tab indices be distinct integers?

BurningTreeC:
@Jermolene 

> why are we assigning the same value to every edit-text widget? Shouldn't the tab indices be distinct integers?

by default all elements are at the same tabindex level (0)

when tabbing, the browser jumps to the next element in the DOM with the **same** tabindex level

when no element is found, it jumps to the higher level


the idea is to leave all elements default, except the ones we're interested in. here the inputs in the edittemplates

they all get a tabindex higher than the default one, like **1** (that's what I use)

so when tabbing from the tags input, the next element with tabindex 1 is the editor, the toolbar buttons all have default 0

BurningTreeC:
short:

first all 0, then all 1, then all 2 ...

BurningTreeC:
when a tiddler is being edited, the autofocus on the title assures that we're automatically at the given tabindex level

Jermolene:
Ah yes of course...

Jermolene:
We should also have docs for the new hidden setting, maybe?

BurningTreeC:
yes I'll add them, thanks for merging :fireworks: !

PR #3759, opened 2019-02-09
Add flag to allow duplicates in filter output

### LINKED ISSUE #3757: Prerelease math filter operators don't allow duplicate inputs
To see a quick demonstration of the problem you can go to tiddlywiki.com/prerelease and make a new tiddler and paste this into the text field:

<$list filter='1 1 1 1 +[sum[]]'/>

the expected output is 4, but because filters enforce uniqueness the output is 1

I am looking at the list widget and filtering code to see if there is a simple way to add an attribute that would allow duplicate entries, but I am a bit concerned about editing any of it because of how carefully the list and filtering has been designed and how deeply integrated into tiddlywiki it is.

inmysocks:
This is one potential way to resolve #3757  I think that it needs more testing before being considered production ready. Also it should use a better character than `?` as the prefix, it was just the first thing I thought of that wasn't in use.

This shouldn't break backwards compatibility in any way other than filters that start with ?, like `? ! . , ' "` would now strip out the `?` character and allow duplicates when they wouldn't before.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -72,13 +72,14 @@ $tw.utils.isArrayEqual = function(array1,array2) {
 Push entries onto an array, removing them first if they already exist in the array
 	array: array to modify (assumed to be free of duplicates)
 	value: a single value to push or an array of values to push
+	allowDuplicate: a boolean flag, if true than duplicate entries are not removed
 */
-$tw.utils.pushTop = function(array,value) {
+$tw.utils.pushTop = function(array,value, allowDuplicate) {
 	var t,p;
 	if($tw.utils.isArray(value)) {
 		// Remove any array entries that are duplicated in the new values
 		if(value.length !== 0) {
-			if(array.length !== 0) {
+			if(array.length !== 0 && !allowDuplicate) {
 				if(value.length < array.length) {
 					for(t=0; t<value.length; t++) {
 						p = array.indexOf(value[t]);
@@ -100,7 +101,7 @@ $tw.utils.pushTop = function(array,value) {
 		}
 	} else {
 		p = array.indexOf(value);
-		if(p !== -1) {
+		if(p !== -1 && !allowDuplicate) {
 			array.splice(p,1);
 		}
 		array.push(value);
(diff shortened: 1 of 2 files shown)


BurningTreeC:
hi @inmysocks , what do you think about borrowing regex syntax like `.*`/`*` or a `+` at the *end*, like `{{{ 
 1 1 1 1 +[sum[]]+ }}}`  ?

inmysocks:
I like it. 

That would remove most or all of the problems with backwards compatibility, or at least not make it worse, and it would be something that already makes sense to at least some people. Having it at the end also gives the idea that it is something that affects the output.

I think that in terms of regular expressions + (one or more) makes more sense than * (0 or more), but * may be better because + is already used as a prefix that means something completely different.

If anyone else has opinions or suggestions about what to use speak up. I will edit the pull request to use * at the end of the filter string sometime later unless there is a better idea.

inmysocks:
I just did a test and filters using {{{ }}} don't work with this. I am not sure why but I will look into fixing that soon.

pmario:
I think, there should be a new function. `pushTop()` is one of the functions, that is called very often in the core. So adding functionality to it, will only make it slower.   AND a big part of the function is to search for duplicates. 

I think the pushDupe() can be implemented with much less code.

just my concerns

inmysocks:
@pmario I don't understand, there are 3 lines in pushTop changed and 8 lines changed in filters.js to make this work, adding another function would require every filter that uses pushTop to have a check to determine if pushTop or pushDupe should be used and the pushDupe function would have to be created. pushDupe would be very simple, but there are many more places where errors could be made and it wouldn't be backwards compatible because every custom filter operator that uses pushTop would have to be updated as well.

Making pushTop slower is something I am worried about, but it is at most one additional check on the value of a boolean in an if statement that already exists. I can do some speed tests but I think that compared to the cost of multiple indexOf and array.splice calls it is almost negligible.

inmysocks:
After some discussion over in #3757 It looks like this will become redundant. I am going to close this in favour of the suggestion that we just remove the uniqueness restriction on filters.

PR #3760, opened 2019-02-09
add "equals" filter that returns [] if not equal

BurningTreeC:


BurningTreeC:
... oh, I noticed now that this needs additional handling for `!equals[value]`

BurningTreeC:
I reopened this for discussion

twMat:
Side note: I believe `regexp[^...$]` achieves "equals". I do think we need a dedicated "equals" but if the regexp does this then just maybe maybe this simplifies the implementation for it. Just sayin'.

BurningTreeC:
@twMat I think you're right and we can probably close this

twMat:
So no mistake; I do think a dedicated "equals" is of value. Regexp is a bit scary and not as obvious.

BurningTreeC:
We have a match filter now

PR #3763, opened 2019-02-10
fix tags not animating in viewtemplate

BurningTreeC:
this fixes the tags not animating when we add/remove tags to a visible tiddler for example with a tm-add/remove-tag message

the PR creates two different tiddlerState-hashes, one with and the other without the tags. then, if the one without the tags has changed, it makes a full tiddler-widget refresh, if only the tags have changed, it gives the animation time to finish, then updates the tiddlerState-hash and refreshes the tiddler's children

### DIFF
--- a/core/modules/widgets/tiddler.js
+++ b/core/modules/widgets/tiddler.js
@@ -62,7 +62,9 @@ TiddlerWidget.prototype.computeTiddlerState = function() {
 		tiddlerTagClasses: this.getTagClasses()
 	};
 	// Compute a simple hash to make it easier to detect changes
-	state.hash = state.currentTiddler + state.missingTiddlerClass + state.shadowTiddlerClass + state.systemTiddlerClass + state.tiddlerTagClasses;
+	state.shortHash = state.currentTiddler + state.missingTiddlerClass + state.shadowTiddlerClass + state.systemTiddlerClass;
+	state.hash = state.shortHash + state.tiddlerTagClasses;
+	
 	return state;
 };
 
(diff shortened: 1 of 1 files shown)


BurningTreeC:
No comments or concerns about this pr?

Jermolene:
Hi @BurningTreeC apologies, I hadn't seen this one.

> this fixes the tags not animating when we add/remove tags to a visible tiddler for example with a tm-add/remove-tag message

OK, so the cause is that the tiddler widget in `$:/core/ui/ViewTemplate` is being refreshed when the tags change, and thus the list widget in `$:/core/ui/ViewTemplate/tags` doesn't get a chance to animate?

If so, presumably that's because of the CSS class variables that are maintained by the tiddler widget?

If so, perhaps another solution might be to compute those CSS classes via a `class={{{...}}}` attribute where needed, and entirely avoid computing those classes in the tiddler widget, and thus avoid it refreshing?

BurningTreeC:
Hi @Jermolene, your idea sounds better

BurningTreeC:
closing this in favor of a better idea

BurningTreeC:
@Jermolene, 

>If so, perhaps another solution might be to compute those CSS classes via a class={{{...}}} attribute where needed, and entirely avoid computing those classes in the tiddler widget, and thus avoid it refreshing?

where did you mean to compute those classes? I'd give it a shot

BurningTreeC:
reopening for discussion

BurningTreeC:
I don't see a different way than waiting for the interval of the animation duration to set the tiddlerTagClasses through the tiddler widget, like it's done in this PR

Jermolene:
Hi @BurningTreeC the first part is fairly easy: we trim the view template down to remove the class assignments for shadow, missing, etc. That's enough to restore the tag animations if you programmatically add/remove them:

```
\define folded-state()
$:/state/folded/$(currentTiddler)$
\end
<$vars storyTiddler=<<currentTiddler>> tiddlerInfoState=<<qualify "$:/state/popup/tiddler-info">> userClass={{!!class}}>
<div data-tiddler-title=<<currentTiddler>> data-tags={{!!tags}} class={{{ [[tc-tiddler-frame tc-tiddler-view-frame]] }}}>
<$list filter="[all[shadows+tiddlers]tag[$:/tags/ViewTemplate]!has[draft.of]]" variable="listItem">
</$list>
</div>
</$vars>
```

Now, all that's needed is a humungous `{{{ }}}` to assign the other classes to the div.

BurningTreeC:
>Now, all that's needed is a humungous {{{ }}} to assign the other classes to the div.

I was at a similar point before, but gave up in front of that `{{{ }}}`

Jermolene:
Hi @BurningTreeC I've got something that seems to work, but it needs a new filter operator:

```
{{{ tc-tiddler-frame tc-tiddler-view-frame [<currentTiddler>is[tiddler]as[tc-tiddler-exists]] [<currentTiddler>is[missing]!is[shadow]as[tc-tiddler-missing]] [<currentTiddler>is[shadow]as[tc-tiddler-exists tc-tiddler-shadow]] [<currentTiddler>is[system]as[tc-tiddler-system]] [{!!class}] [<currentTiddler>tags[]encodeuricomponent[]addprefix[tc-tagged-]] +[join[ ]] }}}
```

I'm not sure it's the best name, but the `as` operator replaces any input titles with the specified operand.

BurningTreeC:
@Jermolene my alternative would be a huge addsuffix removeprefix construct ... I've got no better idea for the new operator-name, sorry

BurningTreeC:
@Jermolene, I'm closing this PR in favor of a new one

PR #3768, opened 2019-02-14
Keyboard Shortcut for saving the wiki

hchaase:
just made a shortcut for the HTML5 fallback saver


pmario:
There is a CTRL-S shortcut now. So close

PR #3770, opened 2019-02-15
proposal: reveal widget popup adjustments

BurningTreeC:
this adds two things if a popup is handled through the positionPopup method and the popup belongs to a tiddler in the story river:

- sets the popup state tiddler's `popup-tiddler` field to the title of the corresponding story-tiddler
- adjusts the left position of the popup in case that its right position is outside of the viewport (moves it to the left for the amount it's outside)

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -99,6 +99,7 @@ RevealWidget.prototype.positionPopup = function(domNode) {
 		    viewportWidth = domNode.ownerDocument.defaultView.innerWidth || domNode.ownerDocument.documentElement.clientWidth,
 		    rect = domNode.getBoundingClientRect();
 		if((rect.x + domNodeWidth) > viewportWidth) {
+			// -19 for the standard width of scrollbars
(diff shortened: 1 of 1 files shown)


BurningTreeC:
closing this - doesn't work like I thought

Jermolene:
No worries @BurningTreeC it's useful to be able to see what you're thinking about!

I just noticed that GitHub added support for "draft" PRs which might be helpful:

https://github.blog/2019-02-14-introducing-draft-pull-requests/

BurningTreeC:
Hi @Jermolene - I desperately need the title of the tiddler with an open popup stored somewhere - the best would be a field in the popup-state-title

the other thing in this PR is popup-positioning when the right of the popup is outside, but that's not working correctly here ... I was assuming that it works for absolutely positioned parents and relatively positioned parent nodes alike and had tested with absolutely positioned only which isn't the case for most popups in the standard tw

BurningTreeC:
>I just noticed that GitHub added support for "draft" PRs which might be helpful

They made those right for me :grin:

Issue #3771, opened 2019-02-16
BUG: list-links macro - Link breaking if caption includes CamelCase


00SS:
On tiddlywiki.com edit the tiddler [Hidden Setting: Search AutoFocus](https://tiddlywiki.com/#Hidden%20Setting%3A%20Search%20AutoFocus)

Add a caption field with value: **`Search AutoFocus`**

Open tiddler: [Hidden Settings](https://tiddlywiki.com/#Hidden%20Settings) that uses `<<list-links "[tag[Hidden Settings]]">>`

Note that the link to the tiddler looks like one link, but it breaks into two values.
**`Search `**  and the space following it links properly.
**`AutoFocus`** becomes a separate CamelCase link.

00SS:
Just as an idea, this seems to address 2 issues in the `<<list-links ... >>` macro
- Fixes this issue's CamelCase broken link
- Fixes #3753 if used in the draggable also

Of course it is possible that it may cause problems in other ways.

```
\define unCamelCase(undo)
\rules except wikilink
$undo$
\end

\define list-links(filter,type:"ul",subtype:"li",class:"",emptyMessage)
<$type$ class="$class$">
<$list filter="$filter$" emptyMessage=<<__emptyMessage__>>>
<$subtype$>
<$link to={{!!title}}>
<$list filter="[all[current]has[caption]]" 
      emptyMessage="""<$view field="title"/>""">
      <$macrocall $name="unCamelCase" undo={{!!caption}}/>
</$list>
</$link>
</$subtype$>
</$list>
</$type$>
\end
```

Jermolene:
Hi @00SS by convention, the `caption` field is wikified. Thus if you include a camelcase link in a caption field it will be rendered as a link. It's done this way so that we can have caption fields that transclude language tiddlers for translations.

00SS:
What I can see happening here is that:
 - The macro doesn't just "display" the `caption` field
  - A `caption` field containing some CamelCase is transcluded, making a string that includes CamelCase link(s)
 - The string containing CamelCase link(s) get used inside the `<$link>` widget, creating what looks like one normal link
 - The non-CamelCase parts of the link, both before and after any CamelCase portion, link normally to the tiddler's  `title` field
 - The CamelCase parts link to each **CamelCase's tiddler** `title`
 - The whole link looks like one normal link, but it isn't. You open different tiddlers depending on where in the link you click.

[...]

PR #3772, opened 2019-02-16
Update: $:/core/macros/list

00SS:
Resolving issue #3771 where CamelCase in caption fields mess up the links created. 

Used the same technique adopted in [$:/core/macros/toc](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/wiki/macros/toc.tid) - by wrapping the `<$transclude`> widget with `<$set name="tv-wikilinks" value="no">`

### DIFF
--- a/core/wiki/macros/list.tid
+++ b/core/wiki/macros/list.tid
@@ -76,11 +76,12 @@ tags: $:/tags/Macro
 </$set>
 \end
 
-\define list-tagged-draggable(tag,subFilter,emptyMessage,itemTemplate,elementTag:"div")
+\define list-tagged-draggable(tag,subFilter,type:"ul",subtype:"li",class:"",emptyMessage,itemTemplate,elementTag:"div")
 <$set name="tag" value=<<__tag__>>>
+<$type$ class="$class$">
 <$list filter="[<__tag__>tagging[]$subFilter$]" emptyMessage=<<__emptyMessage__>>>
 <$elementTag$ class="tc-menu-list-item">
-<$droppable actions="""<$macrocall $name="list-tagged-draggable-drop-actions" tag=<<__tag__>>/>""">
+<$droppable actions="""<$macrocall $name="list-tagged-draggable-drop-actions" tag=<<__tag__>>/>""" tag="""$subtype$""">
 <$elementTag$ class="tc-droppable-placeholder">
 &nbsp;
 </$elementTag$>
@@ -94,6 +95,7 @@ tags: $:/tags/Macro
 </$droppable>
 </$elementTag$>
 </$list>
+</$type$>
 <$tiddler tiddler="">
 <$droppable actions="""<$macrocall $name="list-tagged-draggable-drop-actions" tag=<<__tag__>>/>""">
 <$elementTag$ class="tc-droppable-placeholder">
--- a/core/wiki/macros/list.tid
+++ b/core/wiki/macros/list.tid
@@ -76,12 +76,11 @@ tags: $:/tags/Macro
 </$set>
 \end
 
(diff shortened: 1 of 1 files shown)


00SS:
The list-tagged-draggable macro is sad and feels discriminated against.

All the other list macros in `$:/core/macros/list` have a `type` & `subtype` parameter for creating ordered/unordered lists.

Ref: Google Groups 2019-02-21 [Generate numbered list using list-tagged-draggable](https://groups.google.com/forum/?hl=en#!topic/tiddlywiki/1gIvh_sDjuA)

It is therefore proposed to right this wrong.

core/wiki/macros/list.tid
```diff
@@ -72,11 +76,12 @@ tags: $:/tags/Macro
 </$set>
 \end
 
-\define list-tagged-draggable(tag,subFilter,emptyMessage,itemTemplate,elementTag:"div")
+\define list-tagged-draggable(tag,subFilter,type:"ul",subtype:"li",class:"",emptyMessage,itemTemplate,elementTag:"div")
```
pmario: this would make it the same as list-links-dragable. ... So we have 2 macros that do the same thing. IMO that's wrong!

pmario:
IMO this PR mixes 2 different things. 

 - Fix the problem with CamelCase
 - change the default behaviour 

I'm OK to fix the CamelCase problem. ... but not with the 2nd change.

00SS:
Is there no other difference between `list-tagged-draggable` and `list-links-draggable` ?

The `list-tagged-draggable-drop-actions` macro looks quite different to the `list-links-draggable-drop-actions` macro, but I did not try and see the differences.

bluepintail:
There's a major difference between what `list-tagged-draggable` and `list-links-draggable` do. `list-tagged-draggable` generates the list based on a tag (and can accept a subfilter argument) whereas `list-links-draggable` uses a field from a single tiddler to generate the list.

To be honest I think this is confusing because the `list-links` macro generates a list from a filter, so I think it would make a lot more sense if `list-links-draggable` was just a draggable version of the same thing.

00SS:
Thanks @bluepintail - I was not so clear on the differences to be able to explain it to @pmario  
Here's a simple table that might make it clearer. Comparison below only looks at differences.

 list-**links**-draggable & list-**tagged**-draggable

Macro | Action | Relevant parameters
------------ | ------------- | -------------
links | renders the ListField of a tiddler as a list of links | tiddler field <br>(list)type/subtype/class
tagged | renders the tiddlers with a particular tag as a list of links | tag subfilter <br>emptyMessage

pmario:
@00SS imo that's not, what [the doc](https://tiddlywiki.com/#list-links%20Macro) says:

>The list-links macro returns a formatted list of links to a selection of tiddlers.

00SS:
@pmario yes you are right. But we are not touching the **list-links** macro in this Pull Request.

Only the **list-links-draggable** and the **list-tagged-draggable**

Both my **Action** definitions above are from the [docs](https://tiddlywiki.com/#list-links-draggable%20Macro:%5B%5Blist-links-draggable%20Macro%5D%5D%20%5B%5Blist-tagged-draggable%20Macro%5D%5D).

pmario:
A I wrote. I'm OK with the fix of tv-wikilink. ... But this PR seems to change the the tagged list to an unordered list view. Which imo is a problem. 

Create 2 PRs ... 1 with the fix and a second one as a proposal.

00SS:
As is often the case, @pmario seems to sense when something just doesn't fit right. By making the **list-tagged-draggable** an unordered list view BY DEFAULT, it also makes the drop-down list on any tag into an unordered list. I was not aware this same macro is used there. This is not a desired effect at all. So, I will (try and) revert those changes here.

Perhaps @bluepintail can open an Issue on this for consideration, and a way can be found to allow the default list to be as before with an option to make the unordered (or ordered) list.

PR #3790, opened 2019-02-21
Add switch for enabling duplicates within filters

Jermolene:
This PR explores one possible fix for #3757: to entirely switch off the suppression of duplicates during filter processing.

Despite being an incompatible change to the core, early suggestions are that it has little or no impact on the normal operation of TiddlyWiki.

Thus, I'm making the changes available so that other people can experiment with them to evaluate their impact. 

To enable the new behaviour:

* Add `#dupes` to the browser address bar and refresh the page
* Use the `--dupe` command line switch under Node.js

I've added some suitably eye catching notification messages:

dupe-warning

screenshot 2019-02-21 15 41 37

I've also uploaded this new branch to TiddlySpot:

http://allow-filter-duplicates.tiddlyspot.com/#dupes

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -76,34 +76,42 @@ Push entries onto an array, removing them first if they already exist in the arr
 $tw.utils.pushTop = function(array,value) {
 	var t,p;
 	if($tw.utils.isArray(value)) {
-		// Remove any array entries that are duplicated in the new values
-		if(value.length !== 0) {
-			if(array.length !== 0) {
-				if(value.length < array.length) {
-					for(t=0; t<value.length; t++) {
-						p = array.indexOf(value[t]);
-						if(p !== -1) {
-							array.splice(p,1);
+		if($tw.config._temp_allowDuplicates) {
+			Array.prototype.push.apply(array,value);
+		} else {
+			// Remove any array entries that are duplicated in the new values
+			if(value.length !== 0) {
+				if(array.length !== 0) {
+					if(value.length < array.length) {
+						for(t=0; t<value.length; t++) {
+							p = array.indexOf(value[t]);
+							if(p !== -1) {
+								array.splice(p,1);
+							}
 						}
-					}
-				} else {
-					for(t=array.length-1; t>=0; t--) {
-						p = value.indexOf(array[t]);
-						if(p !== -1) {
-							array.splice(t,1);
+					} else {
+						for(t=array.length-1; t>=0; t--) {
+							p = value.indexOf(array[t]);
+							if(p !== -1) {
+								array.splice(t,1);
+							}
 						}
 					}
 				}
-			}
-			// Push the values on top of the main array
-			array.push.apply(array,value);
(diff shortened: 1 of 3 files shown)


Jermolene:
I'll assemble here a list of things that don't work when duplicates are allowed:

* Control Panel -> Advanced -> Tiddler Fields shows multiple entries for each field:
screenshot 2019-02-21 16 31 46

Jermolene:
The sidebar More -> Explorer tab shows duplicate entries:

screenshot 2019-02-21 16 37 58

pmario:
I'm not sure but I probably miss the point. 

>Despite being an incompatible change to the core, early suggestions are that it has little or no impact on the normal operation of TiddlyWiki.

IMO the current issues are described at: https://tiddlywiki.com/#Dominant%20Append

Since the "dominant append" rule is the default at the moment, 100% of filter outputs are affected. So I do not agree that it has "little to no impact". It heavily depends on the usecase.

On the other hand, I think it may have the potential to speed up filter runs. ... The [$tw.utils.pushTop()](https://github.com/pmario/TiddlyWiki5/blob/add773e74e5497310b162469a384cff1fbf1e8c1/boot/boot.js#L96) is relative performance heavy, since it is called on every input title that is pushed to the output. 

So a `+[unique[]]` filter run at the end of filters that do not need duplicates may be more performant. (may be ... measurements are needed)

We do have the [unique filter operator](https://tiddlywiki.com/#unique%20Operator:%5B%5Bunique%20Operator%5D%5D%20%5B%5Bunique%20Operator%20(Examples)%5D%5D) ... **which does no "dominant append"** ... 

So we would need to add a new `unique-a` filter operator that mimics the current behaviour. As Jeremy's examples show, we would need it

Jermolene:
Hi @pmario I should have been clearer that this PR is currently broken due to subsequent changes in `master`. I'm in the process of fixing it but in the meantime you can experiment with the build at http://allow-filter-duplicates.tiddlyspot.com/#dupes

> So I do not agree that it has "little to no impact". It heavily depends on the usecase.

What I was getting at is that there are almost no practical problems with tiddlywiki.com when building with deduplication disabled. As noted above, I've only found two minor problems so far.

Jermolene:
The motivation for considering this change as part of v5.3.0 is that the current deduplication behaviour confuses new users and makes the documentation significantly more complicated.

pmario:
Yea, since we included mathematical operators the automatic deduplication is a pain point. ...

pmario:
>I've added some suitably eye catching notification messages:

I think the animation is not really needed. Even in "closed" state the animation is way to aggressive for me personally.

saqimtiaz:
Suggest making it possible to toggle this at the level of a filter expression by means of a filter pragma. Will create a separate issue with a proposal for the syntax.

saqimtiaz:
Note that since this PR was created, filters have been architected to use a custom LinkedList class, so we would need to patch the behaviour of the pushTop method in LinkedList.

PR #3796, opened 2019-02-22
add "isHidden" attribute to domNode

BurningTreeC:
this adds an additional `isHidden` attribute to domNodes but not like the "hidden" attribute after the animation is finished, but right after the state is triggered

this isn't used in the core, but it's useful if listening for the change of this hidden attribute. in that case it's better to be informed instantly when it's about to get the hidden attribute. so one listens for the `isHidden` attribute change

a proposal, but a very useful one for me :smile:

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -49,6 +49,7 @@ RevealWidget.prototype.render = function(parent,nextSibling) {
 	}
 	if(!this.isOpen) {
 		domNode.setAttribute("hidden","true");
+		domNode.setAttribute("isHidden","true");
 	}
 	this.domNodes.push(domNode);
 };
@@ -230,13 +231,17 @@ RevealWidget.prototype.updateState = function() {
 	}
 	if(this.isOpen) {
 		domNode.removeAttribute("hidden");
+		domNode.removeAttribute("isHidden");
         $tw.anim.perform(this.openAnimation,domNode);
 	} else {
+		domNode.setAttribute("isHidden","true");
 		$tw.anim.perform(this.closeAnimation,domNode,{callback: function() {
 			//make sure that the state hasn't changed during the close animation
 			self.readState()
 			if(!self.isOpen) {
 				domNode.setAttribute("hidden","true");
+			} else {
+				domNode.removeAttribute("isHidden");
 			}
 		}});
 	}


pmario:
I think that's the wrong name. isHidden and hidden are the same thing, from the name perspective. But you implemented it, to contain different info. 

It contains the "hide animation is active info" and it should be named accordingly.

just my thoughts

pmario:
`ishiding` may be a possibility. BUT it's still a bad one

pmario:
Where is the opposite?

BurningTreeC:
@pmario - yes I'm that bad with names ... `isAboutToHide` ?

BurningTreeC:
... but then, the useful part where I'm using this is watching the node for the change of this attribute and it should stay the same as long as it's hidden, so `isAboutToHide` matches the time of the animation duration but not the hidden state

pmario:
IMO When animation starts, it is not hidden yet. .. So still a strange name.

pmario:
I'm not sure, why you track the DOM. In TW that's not the way to detect something. ... In TW whenever the wiki-store is changed every widget is notified of this change and the redraw themself. 

So what exactly do you try to achieve. ... IMO you should listen for that built in change event.

BurningTreeC:
>I'm not sure, why you track the DOM. In TW that's not the way to detect something. ... In TW whenever the wiki-store is changed every widget is notified of this change and the redraw themself.

As of my experience, widgets refresh what's necessary on changed tiddlers or changed attributes. The refresh method can also be made more complex so that on certain conditions it makes a full refresh.
A full refresh is often too much / too heavy.

There are some wiki-store changes that cannot precisely be tracked. The reveal widget for example:

I have say 5 reveal widgets one within the other. How can my widget - which may be there within those reveal widgets - tell if it's visible or not (the reveal widgets may have retain="yes" so it's just hidden but active). I don't know which state tiddler to track because they may all be qualified states and I don't know their positions in the DOM etc. ... that's complicated

What I can do is walk up the dom from my widget and detect all reveal-widget domNodes, listen on them for attribute changes, if no reveal-widget domNode is hidden, I know the widget is visible. If one of them is hidden, the widget is hidden, too. Now the hidden attribute changes after the animation duration has passed. this attribute here would change immediately and such a change can be tracked when it happens.

I'm detecting using a MutationObserver ...

observer.observe... attributes: true, attributeOldValue: true

I haven't found a built-in way to do all this. Is there one?

>So what exactly do you try to achieve. ... IMO you should listen for that built in change event.

What would be that built-in change event for my example above if I don't know the state tiddlers?

pmario:
I personally would never use "retain=yes", so the inner DOM structure won't exist. So you can't track them.

pmario:
> I don't know which state tiddler to track because they may all be qualified states and I don't know their positions in the DOM etc. ... that's complicated

Yea, the "unpredictability" of qualified state is a problem, we should try to fix.

pmario:
I think, we are fighting the symptoms of a "hidden" problem here. We need to identify the real problem. ... just some thoughts.

BurningTreeC:
>I personally would never use "retain=yes", so the inner DOM structure won't exist. So you can't track them.

Every tiddler's ViewTemplate, the folded-state:

```
<$reveal tag="div" class="tc-tiddler-body" type="nomatch" stateTitle=<<folded-state>> text="hide" retain="yes" animate="yes">
```

the sidebar itself, too

BurningTreeC:
> > I don't know which state tiddler to track because they may all be qualified states and I don't know their positions in the DOM etc. ... that's complicated
> 
> Yea, the "unpredictability" of qualified state is a problem, we should try to fix.

Fixing the unpredictability of qualified states is one thing, but that doesn't fix the issue mentioned here, where it can still be any state title, one using a tiddler title, one using a qualified state, another crazy naming, whatever. That doesn't matter, it can be any name

BurningTreeC:
>I think, we are fighting the symptoms of a "hidden" problem here. We need to identify the real problem. ... just some thoughts.

I believe this is something different, as mentioned in my previous response. This is not related to qualified states. This is related to the unpredictability of where I am in the DOM at any given point and the unpredictability of how that structure around me looks like

PR #3797, opened 2019-02-22
Hook for closing tiddlers

theSherwood:
Hi Folks,

Most of the other common events have hooks associated. But there didn't seem to be a hook for closing tiddlers. Honestly, I'd like to put hooks on all of these events. It seems like a low-overhead way of adding some convenience for those making plugins. I figured a good place to start was with the most obvious event that didn't yet have a hook: closing tiddlers.

Best wishes!

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -144,6 +144,7 @@ NavigatorWidget.prototype.handleNavigateEvent = function(event) {
 
 // Close a specified tiddler
 NavigatorWidget.prototype.handleCloseTiddlerEvent = function(event) {
+	event = $tw.hooks.invokeHook("th-closing-tiddler",event);
 	var title = event.param || event.tiddlerTitle,
 		storyList = this.getStoryList();
 	// Look for tiddlers with this title to close


saqimtiaz:
I ran into a need for a `th-closing-tiddler` hook yesterday.

@Jermolene thoughts on whether we can merge this? Anything else needed? Docs I can add in a separate PR.

pmario:
I also think hooks are relatively low cost and can give authors a lot of flexibility.

Jermolene:
Thanks @theSherwood @saqimtiaz @pmario

One thing we're going to have to watch is that in the future we plan for all of the logic in the navigator widget to be implemented as wikitext actions, at which point it is very hard to see how we could continue to support the current JS hooks. (I imagine that there would be an analogous wikitext mechanism that might execute all actions tagged `$:/tags/Actions/ClosingTiddler` when closing a tiddler, but it wouldn't be directly compatible).

saqimtiaz:
@Jermolene thank you.

I think for most use cases eventually transitioning the approach of invoking tagged tiddlers for hooks should be sufficient. In fact, I already do so for all hooks that I use, e.g:

https://github.com/saqimtiaz/streams/blob/137a2008e2d2ac379e43b35060c2e82a8a76b645/plugins/streams/delete-hook/hook.js#L8-L14

```javascript
$tw.hooks.addHook("th-deleting-tiddler",function(tiddler) {
	if(tiddler.fields["draft.of"]) {
		return;
	}
	
	$tw.rootWidget.invokeActionsByTag("$:/tags/deleteTiddlerAction",null,{currentTiddler : tiddler.fields.title});
});
```

PR #3798, opened 2019-02-23
add resize listener utilities to dom.js + innerwiki dimension refresh

BurningTreeC:
this adds `$tw.utils.addResizeListener(domNode,callback)` `$tw.utils.removeResizeListener(domNode,callback)` and `$tw.utils.isDOMElement(obj)`

this can be used in the innerwiki plugin for example, to listen on the `domWrapper` if its dimensions change and update the iframe accordingly

this is a `Draft PR`

Update: the innerwiki plugin updates its dimensions whenever the dimensions of the domWrapper change ... when resizing the browser, changing the sidebar state, changing the story-river width etc. No full refresh which is too heavy. Just dimensions and scaling

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -205,6 +205,76 @@ exports.addEventListeners = function(domNode,events) {
 };
 
 /*
+Listen for resize events on a give domNode and execute a callback on resize
+http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/
+*/
+exports.addResizeListener = function(domNode,callback) {
+	if(domNode && !domNode.__resizeListeners__) {
+		domNode.__resizeListeners__ = [];
+		if(domNode.ownerDocument.attachEvent) {
+			domNode.__resizeTrigger__ = domNode;
+			domNode.attachEvent('onresize',$tw.utils.resizeListener);
+		}
+		else {
+			if(domNode.ownerDocument.defaultView.getComputedStyle(domNode).position == 'static') {
+				domNode.style.position = 'relative';
+			}
+			var resizeObject = domNode.__resizeTrigger__ = domNode.ownerDocument.createElement('object');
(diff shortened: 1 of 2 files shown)


core/modules/utils/dom/dom.js
```diff
@@ -204,6 +204,77 @@ exports.addEventListeners = function(domNode,events) {
 	});
 };
 
+/*
+Listen for resize events on a give domNode and execute a callback on resize
```
pmario: typo: given?

core/modules/utils/dom/dom.js
```diff
@@ -204,6 +204,77 @@ exports.addEventListeners = function(domNode,events) {
...
+http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/
+*/
+exports.addResizeListener = function(domNode,callback) {
+	if(domNode && !domNode.__resizeListeners__) {
+		domNode.__resizeListeners__ = [];
```
pmario: Strange variable names: `__asdf__`. IMO we shouldn't change variable name conventions in the middle of a single module.

BurningTreeC: that naming is just for marking it and I just followed the naming from the example, see the link
do we have a naming convention for something like that? I don't know how to describe it - is it a not-browser-specific identifier that should look like one (..) ?

or should I just name it domNode.resizeListeners?

core/modules/utils/dom/dom.js
BurningTreeC: fixed, thanks!

BurningTreeC:
@pmario , thanks for your review. it works well, no problems found so far

core/modules/utils/dom/dom.js
pmario: I don't know. @Jermolene ??

Jermolene: This is a custom property on a DOM node. I was trained not to use them 15 years ago with TWC because they used to generate memory leaks. I'm still not keen on the technique, to be honest.

BurningTreeC: @Jermolene, I closed this PR because I found the technique was erratic ... there are probably better ways

PR #3799, opened 2019-02-23
Add means of halting newTiddlerEvent to hook

admls:
Hi Folks,

Most of the other hooks I've been looking at have means of aborting the native event handling by returning an empty event or false value. The "tm-new-tiddler" event appears to be an exception. To my mind, the hooks, without exception, ought to be able to capable of being used to stop the native handling entirely.

I love the hook system by the way. It seems like a fantastic way of opening up Tiddlywiki to really extensive augmentations. I'd love to see it expanded! It is such a low-overhead way of making tiddlywiki endlessly adaptable without overwriting any core tiddlers.

Best wishes


joshuafontany:
@Jermolene IS this worth considering for the next release?

pmario:
@Jermolene ... I think this one is still valid. Do we want to allow it?

I did have a look at other hooks in the navigator and some of them indeed have the requested behaviour. eg: `th-navigating, th-closing-tiddler, th-editing-tiddler`

Issue #3794, opened 2019-02-22
Tree Macro -- Bizarre Behaviour in Certain Important Cases


TiddlyTweeter:
Try this ...

1 -- Create a Tiddler called "_Rabbit"
2 -- Create a Tiddler called "$:/_Rabbit"
3 -- Create a Tiddler containing <<tree "">>

View (3). There is only ONE entry for "_Rabbit". There should be TWO!

cc @morosanuae

pmario:
IMO that's intended. The tree node filter starts like this: `[all[shadows+tiddlers]` which doesn't include system tiddlers.

00SS:
Definition of `tiddlers` in the `[all]` operator is: **all the non-shadow tiddlers that exist**

00SS:
The tree macro starts like this: `\define tree(prefix: "$:/")`
So I believe the issue is actually more in the documentation.
It should say words to the effect of: **Defaults to the prefix: $:/**

morosanuae:
@TiddlyTweeter 
If you dont specify a prefix then the default argument (`$:/`), introduced in v5.18, kicks in and only the tiddlers that are prefixed with `$:/` are taken into account. The tiddler named "_Rabbit" is not prefixed with `$:/` and so it is excluded.
The "tree" macro was conceived primarily for the internal structure of the system: all the tiddlers prefixed with (`$:/`). This includes system and shadow tiddlers (with system role).

> IMO that's intended. The tree node filter starts like this: `[all[shadows+tiddlers]` which doesn't include system tiddlers.

@pmario I thinks this is wrong. The filter you have mentioned (`[all[shadows+tiddlers]`) includes absolutely all tiddler types, including system tiddlers.

TiddlyTweeter:
I will close this issue as I raised it from a slight misunderstanding.

morosanuae:
@TiddlyTweeter It's understandable because the docs are not very explanatory on this matter. I personally haven't participate in making it, but maybe I'll find the time to improve it.

PR #3805, opened 2019-03-01
DOCS: Update: tree Macro.tid

00SS:
In response to #3794


00SS:
This PR addresses the issue:
The `<<tree>>` macro's definition starts like this: `\define tree(prefix: "$:/")`
The documentation should say words to the effect of: **Defaults to the prefix: $:/**

PR #3806, opened 2019-03-01
fix for "Open" tab placeholders in FireFox

BurningTreeC:
this PR fixes the placeholders in FireFox not being removed on drag-leave from time to time

it consists of 2 mods where apparently both are needed:

- creating a `droppable-item` macro where whitespace is trimmed. that macro contains the droppable and inserts the placeholders
- removing the `&nbsp;` entity in favor of an inline style `height:2em;` on the placeholder div, putting it in a macro where whitespace can be trimmed, too

I'm investigating if there's a similar fix for the top page dropzone

### DIFF
--- a/core/ui/SideBar/Open.tid
+++ b/core/ui/SideBar/Open.tid
@@ -26,7 +26,7 @@ $button$
 <div class="tc-sidebar-tab-open">
 <$list filter="[list[$:/StoryList]]" history="$:/HistoryList" storyview="pop">
 <div class="tc-sidebar-tab-open-item">
-<$macrocall $name="droppable-item" button="""<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini">&times;</$button>&nbsp;<$link to={{!!title}}><$view field="title"/></$link>"""/>
+<$macrocall $name="droppable-item" button="""<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini">{{$:/core/images/close-button}}</$button>&nbsp;<$link to={{!!title}}><$view field="title"/></$link>"""/>
 </div>
 </$list>
 <$tiddler tiddler="">
--- a/core/ui/SideBar/Open.tid
+++ b/core/ui/SideBar/Open.tid
@@ -9,23 +9,29 @@ caption: {{$:/language/SideBar/Open/Caption}}
 <$action-listops $tiddler="$:/StoryList" $subfilter="+[insertbefore:currentTiddler<actionTiddler>]"/>
 \end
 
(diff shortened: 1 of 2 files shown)


core/ui/SideBar/Open.tid
```diff
@@ -9,23 +9,29 @@ caption: {{$:/language/SideBar/Open/Caption}}
 <$action-listops $tiddler="$:/StoryList" $subfilter="+[insertbefore:currentTiddler<actionTiddler>]"/>
 \end
 
-<$list filter="[list[$:/StoryList]]" history="$:/HistoryList" storyview="pop">
-<div style="position: relative;">
-<$droppable actions=<<drop-actions>>>
-<div class="tc-droppable-placeholder">
-&nbsp;
+\define placeholder()
+\whitespace trim
```
pmario: I think you don't need the `whitespace trim` here. It's only 1 line with no whitespace

core/ui/SideBar/Open.tid
```diff
@@ -9,23 +9,29 @@ caption: {{$:/language/SideBar/Open/Caption}}
...
+<div class="tc-droppable-placeholder" style="line-height:2em;height:2em;">
 </div>
```
pmario: `</div>` can be on the same line. no content

pmario:
Many thanks for your ongoing work!

core/ui/SideBar/Open.tid
BurningTreeC: I think you're right, I'll try without

core/ui/SideBar/Open.tid
BurningTreeC: yes it can just be a `<div .../>`

BurningTreeC:
I would also change the `&times;` to the close-button svg because on some machines it's really small, see screenshots attached (the good-looking one uses svg icons), and I'm not able to control it using css

01-03-2019-175752_shoot


01-03-2019-175854_shoot

Jermolene:
Thanks @BurningTreeC. I'm testing this now.

> I would also change the `&times;` to the close-button svg because on some machines it's really small, see screenshots attached (the good-looking one uses svg icons), and I'm not able to control it using css

Good idea.

Jermolene:
Hi @BurningTreeC OK I'd prefer to keep the placeholder rectangle how it is at the moment, but perhaps we should talk about that in #3807

BurningTreeC:
@Jermolene , for testing: the inline styles are removed and there's no stylesheet tiddler yet that adds them. but the classes are there. you can use this:

```
.tc-sidebar-tab-open-item { position: relative; }

.tc-sidebar-tab-open .tc-droppable-placeholer {  line-height:2em; height:2em; }
```

BurningTreeC:
Sorry for the styling confusion, there are typos in the styles above and the close buttons aren't styled

```
.tc-sidebar-tab-open-item { position: relative; }

.tc-sidebar-tab-open .tc-droppable-placeholder {  line-height: 2em; height: 2em; }

.tc-sidebar-tab-open .tc-btn-invisible.tc-btn-mini svg {
font-size: 0.7em;
fill: <<colour muted-foreground>>;
}
```

Jermolene:
Thanks @BurningTreeC

PR #3809, opened 2019-03-02
attempt on fixing the sticky dropzone on firefox

BurningTreeC:
in my tests listening for the dragend event and removing the tc-dragover class in that case removes the `tc-dragover` every time it should ...

without this, firefox often doesn't remove it and the green bar sticks at the top

investigating now if the droppable widget has a similar problem

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -42,7 +42,8 @@ DropZoneWidget.prototype.render = function(parent,nextSibling) {
 		{name: "dragover", handlerObject: this, handlerMethod: "handleDragOverEvent"},
 		{name: "dragleave", handlerObject: this, handlerMethod: "handleDragLeaveEvent"},
 		{name: "drop", handlerObject: this, handlerMethod: "handleDropEvent"},
-		{name: "paste", handlerObject: this, handlerMethod: "handlePasteEvent"}
+		{name: "paste", handlerObject: this, handlerMethod: "handlePasteEvent"},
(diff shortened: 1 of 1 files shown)


BurningTreeC:
it appears as if the droppable widget needs that one, too. when I select text within a wiki and drag it around, then move it to the open sidebar tab and drag it over its droppable list-items and repeat that more times, one or more of the droppable widgets in the open tab sometimes keep(s) its/their tc-dragover class. handling the dragend there, too improves this, I don't have sticky placeholders

BurningTreeC:
there are more suspicious things going on in firefox , as I read this article:

https://stackoverflow.com/questions/42775095/dragexit-vs-dragleave-which-should-be-used

and the point where it states that the moment when the dragleave event is fired is after the dragenter

I believe the dropzone and droppable widgets need some investigation what is happening on firefox

looking at the `this.currentlyEntered` array it doesn't seem correct all the time, especially when those strange things happen

I have no real solution and just suspicions though

Jermolene:
Hi @BurningTreeC I hate to think how much time I've burned over the years trying to fix drag and drop in Firefox! Definitely worth continuing to work on it, though, the bugs in Firefox are so egregious they will surely be fixed at some point...

BurningTreeC:
@Jermolene, yes it looks like a mess.

I'm also inclined to just sit back and wait for better ff times :smile:

should I close this PR or do you consider it? it does remove the tc-dragover, even if it just fights symptomps

Jermolene:
Yes I think we can merge this.

PR #3810, opened 2019-03-02
add smartIndent to codemirror

BurningTreeC:
this PR makes smartIndent the default setting

it also adds the functionality that when pressing enter, inserting a new line, it follows the indentation above AND inserts tab(s) correctly. currently it inserts spaces in the new line, even if the previous line is indented with a tab

further it enables indentAuto when pressing tab

I think these are sensible defaults one would like to have when choosing codemirror. it improves usability a lot.

I also made the default indentUnit the same value as the default tabSize (both 2)

### DIFF
--- a/plugins/tiddlywiki/codemirror/config/extraKeysTW.tid
+++ b/plugins/tiddlywiki/codemirror/config/extraKeysTW.tid
@@ -11,5 +11,7 @@ type: json
 	"Alt-T": "transposeChars",
 	"Alt-U": "undoSelection",
 	"Shift-Alt-U": "redoSelection",
-	"Cmd-U": ""
+	"Cmd-U": "",
+	"Tab": "indentAuto()",
+	"Enter": "newLineAndIndent()"
 }
--- a/plugins/tiddlywiki/codemirror/config/indentWithTabs.tid
+++ b/plugins/tiddlywiki/codemirror/config/indentWithTabs.tid
@@ -1,0 +1,3 @@
+title: $:/config/codemirror/indentWithTabs
+type: bool
+text: true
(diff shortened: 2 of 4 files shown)


Jermolene:
Hi @BurningTreeC good stuff. It does make me wonder if it might be useful to expose a UI for more CodeMirror options.

BurningTreeC:
>It does make me wonder if it might be useful to expose a UI for more CodeMirror options.

yes that would be useful

there are some options one may like to change on the fly, like line numbers, tab-size. editor-toolbar buttons would be nice for those

a mechanism to set different indentation styles for differend tiddler types would also be nice, so we could set the tiddlywiki-default 8 spaces for application/javascript tab-size

BurningTreeC:
... and auto-refresh when those settings change

pmario:
8 spaces?? really. IMO 4 at most.

BurningTreeC:
> 8 spaces?? really. IMO 4 at most.

it's the default when editing a core javascript tiddler here on github, that's the reason

pmario:
It's just because browsers can't show tabs as 4 chars wide!

pmario:
TW source size would be significantly bigger if we'd use spaces instead of tabs. Since TW isn't minified, this makes a difference.

pmario:
Indents for standard users are useless, since they don't show up. It's only useful for devs. ... You'll have to do a lot of explanation in the future ;)

BurningTreeC:
yeah I don't want to fight about that value, anyone else than me may decide

I actually don't know anything about what's right :grin:

BurningTreeC:
@pmario could you suggest a set of good defaults?

also looking at codemirror.net/doc/manual.html#config , what should be included, what makes sense and what would be good defaults?

pmario:
IMO default tab should be 4. I think that's the default setting for many editors too. That's the setting that produces the most readable source code in TW. 

If you use spaces for indentation you may use 2 chars as default and 4 as the second best option.

PR #3814, opened 2019-03-02
add tabindices to edittemplates

BurningTreeC:
this adds `tabindex={{$:/config/EditTabindex}}` to all input fields in viewtemplates and the tag-picker

it doesn't add anything to toolbar-button dropdowns, that can be an idea for later on

it doesn't add the `$:/config/EditTabindex` tiddler. I would add a configuration option for it in the controlpanel, not enabling it by default


Jermolene:
Closing as per [this discussion](
https://github.com/Jermolene/TiddlyWiki5/pull/3758#issuecomment-468940847)

PR #3816, opened 2019-03-03
an attempt on a regexpreplace filter

BurningTreeC:
this filter takes the input title, tests it with the operand - regex, like the regexp filter

it takes the operator-suffix as a variable name for the replace-string and replaces the regex match with the value of the given variable

example:

```
<$set name="replaceString" value="replaced ST">

{{{ [[Test String]regexpreplace:replaceString[(i)t.*st]] }}}

</$set>
```

output: `replaced STring`

in my tests it has no problems with replace variables that contain special characters

### DIFF
--- a/core/modules/filters/regexpreplace.js
+++ b/core/modules/filters/regexpreplace.js
@@ -1,0 +1,52 @@
+/*\
+title: $:/core/modules/filters/regexpreplace.js
+type: application/javascript
+module-type: filteroperator
+
+Filter operator for regexp matching
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports.regexpreplace = function(source,operator,options) {
+	var results = [],
+		regexpString, replaceVariable, regexp, flags = "", match;
+	// Process flags and construct regexp
+	regexpString = operator.operand,
+	replaceVariable = operator.suffix;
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC not bad! The limitations of the filter format can be frustrating for things like this. I had wondered about making it be a separate widget just so that we can more easily pass further parameters. It's not at all concise, but I'm thinking that this isn't a particularly common operation.

```
<$set name="replaceString" value="replaced ST">

<$replace output="replacedString" original="Test String" replace=<<replaceString>> match="t.*st" mode="i">

<$text text=<<replacedString>>/>

</$replace>

</$set>
```

BurningTreeC:
Hi @Jermolene ... yes your example would be easier someone that doesn't know how to regex I guess

being able to do it within a single filter-run would be powerful, but I see that your way would not change what's possible ... If we could have both it'd great, still great if just one solution makes it :smile:

00SS:
Ref: `<$replace output="replacedString" original="Test String" replace=<<replaceString>> match="t.*st" mode="i">`

If this is implemented, to keep the same attribute naming style for output to a variable, such as `<$set name=` & `<$wikify name=` & `<$macrocall $name=` - may I suggest using `<$replace name=` instead of `<$replace output=`

BurningTreeC:
@Jermolene I believe what you proposed is absolutely sufficient, no need for an extra filter

@00SS , consistent naming is key, I'd also prefer the "name" attribute

PR #3817, opened 2019-03-03
fix missing clear: both when tiddler-body is hidden

BurningTreeC:
to reproduce the problem, enable sticky titles, open a plugin tiddler and open a javascript tiddler below


BurningTreeC:
edit: can be any tiddler, not only a javascript tid

Jermolene:
Hi @BurningTreeC can you show a screenshot/gif of the problem?

BurningTreeC:
@Jermolene - I'm confused, I can't reproduce it anymore. I'll close this and reopen if it happens again and I can reproduce it consistently

PR #3822, opened 2019-03-03
add outline-color to button, textarea, input

BurningTreeC:


Jermolene:
Thanks @BurningTreeC I'm nervous about including a colour that isn't defined in the core palettes because it makes it so hard for people to find. I also wonder whether "outline" is the right name; we use "border" for other things. Perhaps it should be "control-border"?

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -417,6 +421,10 @@ a.tc-tiddlylink-external:hover {
 ** Buttons
 */
 
+button {
```
Jermolene: Even though this is about buttons, it should go up at the top where we have gathered the other style declarations that target all instances of HTML elements.

BurningTreeC:
>  I'm nervous about including a colour that isn't defined in the core palettes because it makes it so hard for people to find. I also wonder whether "outline" is the right name; we use "border" for other things. Perhaps it should be "control-border"?

should we just use the primary color?

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: ok, I'll move it

Jermolene:
The primary colour is supposed to be a prominent highlight colour. Perhaps the "muted-foreground" colour?

BurningTreeC:
> Perhaps the "muted-foreground" colour?

I see the outlining mostly used to highlight the element that has focus, so the primary color would match for that purpose.

For me it should just match the palette and I'm fine with muted-foreground

Jermolene:
> I see the outlining mostly used to highlight the element that has focus, so the primary color would match for that purpose.

I'd misunderstood; the primary colour would indeed be a reasonable choice.

BurningTreeC:
changed to use primary @Jermolene

Jermolene:
Thanks @BurningTreeC

PR #3823, opened 2019-03-04
fix sticky draggable-placeholders in list macros (firefox)

BurningTreeC:
this fixes the sticky placeholders when using firefox, where the list-tagged-draggable macro is used:

- tags dropdown
- controlpanel themetweaks toolbars

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -388,7 +388,7 @@ a.tc-tiddlylink-external:hover {
 	cursor: move;
 }
 
-.tc-sidebar-tab-open .tc-droppable-placeholder {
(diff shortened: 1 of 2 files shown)


themes/tiddlywiki/vanilla/base.tid
```diff
@@ -388,7 +388,7 @@ a.tc-tiddlylink-external:hover {
 	cursor: move;
 }
 
-.tc-sidebar-tab-open .tc-droppable-placeholder {
+.tc-sidebar-tab-open .tc-droppable-placeholder, .tc-tagged-draggable-list .tc-droppable-placeholder, .tc-links-draggable-list .tc-droppable-placeholder {
```
pmario: IMO you may split it into 3 lines. It will be easier to read.

BurningTreeC:
ping @Jermolene

Jermolene:
Thanks @BurningTreeC

PR #3824, opened 2019-03-04
Codemirror: refresh on config changes

BurningTreeC:
this PR makes the cm instance update its configuration options when a configuration tiddler changes

makes the editor instantly apply configuration changes, no widget refresh required

this refactors @pmario 's `getCmConfig` function into a prototype and its config-assign-logic into another one, so that the refresh proto can use this assign-logic for assigning the changed values

### DIFF
--- a/plugins/tiddlywiki/codemirror/engine.js
+++ b/plugins/tiddlywiki/codemirror/engine.js
@@ -69,12 +69,6 @@ function CodeMirrorEngine(options) {
 			self.domNode.appendChild(cmDomNode);
 		}
 	},config);
-	
-	$tw.hooks.addHook("th-refreshing-input",function(widget,editInfo,changedTiddlers,changedAttributes) {
-		if(widget === self.widget) {
-			self.refreshCodeMirrorOptions(changedTiddlers);
-		}
-	});
 
 	// Set up a change event handler
 	this.cm.on("change",function() {
--- a/plugins/tiddlywiki/codemirror/engine.js
+++ b/plugins/tiddlywiki/codemirror/engine.js
@@ -35,53 +35,6 @@ if($tw.browser && !window.CodeMirror) {
 	}
 }
 
-function getCmConfig() {
-	var type,
-		test,
-		value,
-		element,
-		extend,
-		tiddler,
-		config = {},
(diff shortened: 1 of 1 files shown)


BurningTreeC:
@Jermolene , this PR makes codemirror refresh its internal options when a codemirror-config tiddler changes. it works well in my tests. this doesn't add any ui part

@pmario, what do you think about this?

Jermolene:
Hi @BurningTreeC widgets can't have effects, and so listening to the wiki change events is a no-no. Instead, you can detect changes to tiddlers of interest in the refresh method.

BurningTreeC:
@Jermolene

> Hi @BurningTreeC widgets can't have effects, and so listening to the wiki change events is a no-no. Instead, you can detect changes to tiddlers of interest in the refresh method.

does this rule apply here? I thought this isn't the actual widget...

anyway, I'd need to replace/extend the refresh method in factory.js for this then

BurningTreeC:
> anyway, I'd need to replace/extend the refresh method in factory.js for this then

... which would be another side effect

BurningTreeC:
```
	this.widget.wiki.addEventListener("change",function(changes) {
		self.refresh(changes);
	});
```

this part is the culprit

it adds an event listener, but the codemirror engine adds several listeners. I don't understand what makes this different to the other listeners

it calls `self.refresh(changes)` which isn't the widget's refresh method, but a different one (at the bottom of the codemirror engine) ... it exclusively updates codemirror-internal options

BurningTreeC:
> ```
> 	this.widget.wiki.addEventListener("change",function(changes) {
> 		self.refresh(changes);
> 	});
> ```
> this part is the culprit
> 
> it adds an event listener, but the codemirror engine adds several listeners. I don't understand what makes this different to the other listeners
> 
> it calls `self.refresh(changes)` which isn't the widget's refresh method, but a different one (at the bottom of the codemirror engine) ... it exclusively updates codemirror-internal options

it does not refresh the widget

BurningTreeC:
Hi @Jermolene , if we add a hook to factory.js like in #3861 , is it allowed to use the hook here instead of the change listener?

PR #3825, opened 2019-03-04
CodeMirror editortoolbar button + dropdown

BurningTreeC:


pmario:
What does it do?

BurningTreeC:
@pmario , it's just an options button with fast access to the codemirror configuration

with #3824 option changes refresh instantly, so this is convenient

PR #3826, opened 2019-03-04
fix "cannot read substr of undefined" in removeprefix filter

BurningTreeC:
from time to time this error appears when testing an input with the removeprefix filter and somehow there's no input title

last time it happened was when I changed a tiddler's type and a toolbar button had a show-condition using the removeprefix filter


Jermolene:
Thanks @BurningTreeC

Issue #3811, opened 2019-03-02
Bug in location hash processing loses any characters after second #


Jermolene:
@00SS discovered an interesting bug that has been in TW5 [since 2014](986a20b22be8918cec09ff48bc718f8977059b7b): the following code [is used in boot.js](https://github.com/Jermolene/TiddlyWiki5/blob/master/boot/boot.js#L270) to read the location hash (the portion of the current URL from the first `#` character):

```
$tw.utils.getLocationHash = function() {
	var parts = window.location.href.split('#');
	return "#" + (parts.length > 1 ? parts[1] : "");
};
```

The bug occurs when the location hash contains more than one `#` character. For example `https://tiddlywiki.com/#Foo#Bar` would be split into `https://tiddlywiki.com/`, `Foo` and `Bar`. The first string is thrown away and the second string is returned as the location hash, discarding the final string. This means that it is impossible to use a permalink to a tiddler that includes a `#` character. In practice, permalinks are supposed to escape `#` to `%23` (and the permalinks generated by TiddlyWiki itself use this encoding), so I suspect that nobody has ever encountered this bug in the field.

It's easy enough to fix the bug, but @00SS [also discovered an interesting side effect of the bug](https://groups.google.com/d/topic/tiddlywiki/ikjxwjllFtE/discussion): navigating to a location hash that starts with a `#` character will not trigger TiddlyWiki's own navigation processing, but leave the underlying browser processing intact. This enables anchors to be used within tiddlers for sub-tiddler navigation.

So, rather than fix this bug, I propose to document the behaviour of double hash location hashes.

Jermolene:
@00SS would you be interested in making a PR for the required documentation? We'll need to both document the behaviour and how it can be used in practical terms, along the lines of your post.

00SS:
@Jermolene 
My pleasure!
:relaxed:

BurningTreeC:
> look, it's a bug!

> no, itsa feature!

:grin:

:+1:

Jermolene:
Thanks @00SS

00SS:
Just as an interesting note, if the underlying HTML is inspected, this code produces the same HTML as using the `<a href ... </a>` method :
```
<$set name="tv-get-export-link" value="##s004">
<$link>Test link to id=#s004</$link>
</$set>
```
gives: `<a class="tc-tiddlylink tc-tiddlylink-resolves" href="##s004">Test link to id=#s004</a>`
[...]

PR #3767, opened 2019-02-14
DOCS Update: Documentation Macros.tid

00SS:
Reformatting, additions etc. as detailed in Issue #3711

PR #3836, opened 2019-03-07
DOCS: CREATE: Anchor Links using HTML.tid

00SS:
A new tiddler documenting the process for creating anchors and using links for sub-tiddler navigation.

Arising from issue #3811 & Discussed on Google Groups: [Section Names using ID for internal Links](https://groups.google.com/forum/?hl=en#!topic/tiddlywiki/ikjxwjllFtE)


00SS:
I will have to update this PR and include [Tony's insight](https://groups.google.com/d/msg/tiddlywiki/ikjxwjllFtE/V9ByUWhEBAAJ) into using the `<<qualify>>` macro in both the `id` and `href` attributes in order to overcome the limitations of multiple transclusions.

With this inclusion, the internal links within a single tiddlers becomes much more robust. It may still be a challenge to use this for linking from one open tiddler to another one when the `<<qualify>>` macro is used, since the macro does not have a **tiddler** parameter for creating the string for another "given tiddler."

00SS:
The information on using the `<<qualify>>` macro has been added.

A preview can be seen [here](https://00ss.github.io/documentation/tiddlywiki.html#Internal%20Links%20using%20HTML).

twMat:
@00SS - thank you for your work on this.

IMO the term "Internal link" is ambiguous. It is not obvious that it refers to links within a single tiddler. Maybe a better word would be "Tiddler-internal links" or "Tiddler-local links" or "Tiddler sub-links" or just "Linking within a Tiddler".

00SS:
Thanks @twMat 

My difficulty was trying to describe the effect taking into account that a link in open tiddler1 (or sidebar) can scroll to open tiddler2 (or sidebar). I will think about it some more.

Also, it may be wise to standardize how to refer to the `id` portion - as an **anchor**, **marker** or **target**. Pre-HTML5 it would be called an anchor using the `name` attribute, but now it is referred to in different ways.

> In HTML 4.01, the `<a>` tag could be either a hyperlink or an anchor. In HTML5, the `<a>` tag is always a hyperlink, but if it has no `href` attribute, it is only a placeholder for a hyperlink.

The other standardization is: does clicking the link **scroll** or **navigate** to the **anchor**/**marker**/**target**

Yet another consideration is on this present documentation keeping open the possibility of future modification of widgets like the `$link` and `$action-navigate` to take advantage of this, and if so, to perhaps decide on this tiddler suggesting a best practice method (like the HTML element to use for the `id` attribute or anything else).

One nice feature of being able to use a modified `<$link to="SomeTiddler" id="#somewhere">` would be that using the right javascript, the browser's history could be left untouched, and perhaps the link's `href` value not appended to the web page's URL. Of course the best part would be if a closed tiddler could be opened followed by navigation to the `id`

Jermolene:
Hi @00SS I think we should use the term "anchor" and "anchor link" because that matches HTML nomenclature and avoids terms that we already use in TW.

00SS:
- The tiddler has been renamed from **Internal Links using HTML** to **Anchor Links using HTML**
- Text within the tiddler has been changed to use this terminology.
- [Google Groups post](https://groups.google.com/d/msg/tiddlywiki/ikjxwjllFtE/Jddk0T47BQAJ) has been made to allow others to comment
- A preview can be seen here: [Anchor Links using HTML](https://00ss.github.io/documentation/tiddlywiki.html#Anchor%20Links%20using%20HTML)

I would appreciate a check on the **Hints** section at the bottom of the tiddler to know if it is acceptable to suggest using the `tc-tiddlylink` `tc-tiddlylink-missing` & `tc-tiddlylink-shadow` classes for styling the link.

00SS:
This is ready for merging.
The preview can be seen here: 
https://00ss.github.io/documentation/tiddlywiki.html#Anchor%20Links%20using%20HTML

Jermolene:
Congratulations @00SS this is a very high standard of work, I'm really impressed (I will merge once we've completed the discussion at #3767).

twMat:
@00SS Great stuff! Super small correction - the first example reads:

`<h1 id="#heading-01"> My Heading</h1>`

Because it is a very first and clarifying example, and because the text explicitly discusses space characters, I think the (unsignificant) space character before "My" should be removed to avoid confusion.

00SS:
Thanks @twMat 
Corrected that now.

PR #3841, opened 2019-03-08
CodeMirror: require modules in correct order

BurningTreeC:
this assigns an index to each codemirror module in each plugin by which the modules-array is sorted first, then the modules get required in the correct order, so that a module that gets required by another module is loaded first

### DIFF
--- a/plugins/tiddlywiki/codemirror-autocomplete/files/tiddlywiki.files
+++ b/plugins/tiddlywiki/codemirror-autocomplete/files/tiddlywiki.files
@@ -5,35 +5,40 @@
 			"fields": {
 				"type": "application/javascript",
 				"title": "$:/plugins/tiddlywiki/codemirror/addon/hint/anyword-hint.js",
-				"module-type": "codemirror"
+				"module-type": "codemirror",
+				"require-index": "18"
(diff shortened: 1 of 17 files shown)


plugins/tiddlywiki/codemirror-autocomplete/files/tiddlywiki.files
```diff
@@ -5,35 +5,40 @@
 			"fields": {
 				"type": "application/javascript",
 				"title": "$:/plugins/tiddlywiki/codemirror/addon/hint/anyword-hint.js",
-				"module-type": "codemirror"
+				"module-type": "codemirror",
+				"require-index": "18"
```
pmario: hmmm, we should find a way to do this programmatically. ... Doing it be hand is a lot of work and users can't add their own stuff without problems.

pmario:
I found this one: https://gist.github.com/six8/1732686

pmario:
or this one: https://gist.github.com/RubyTuesdayDONO/5006455

BurningTreeC:
@pmario , would you mind giving it a shot? I cannot do more than this

PR #3848, opened 2019-03-11
fix: reveal-widget - prevent from undefined state

BurningTreeC:
the state didn't fall back to the default when there was no data-item to extract for a stateIndex

see the google group for a bug report: https://groups.google.com/forum/#!topic/TiddlyWiki/3jiFpayAIRc


Jermolene:
Thanks @BurningTreeC

PR #3853, opened 2019-03-12
sort palettes by name, not description

BurningTreeC:
... currently it appears as if there was no sorting logic


Jermolene:
Thanks @BurningTreeC

PR #3854, opened 2019-03-12
Add autoParagraph variable

Arlen22:
This tells the parser when to set auto paragraph tags when parsing content. This way the layout of the page is not broken by the tags, but paragraph breaks still get created for content. To use auto-generated paragraph breaks outside of a story river tiddler body, simply set the `autoParagraph` variable to `"yes"` in the tiddler that is being included elsewhere. Some minor adjustments will need to be made to the stock themes, but overall this makes it much easier to layout the user interface using CSS.


Arlen22:
If you're curious, I exported the tw5.com data folder and uploaded it to my GitHub page to give you an idea of what it looks like.

https://github.com/Arlen22/arlen22.github.io/blob/master/tiddlywiki-auto-paragraph.html

It's supposed to work at this URL as well, but it isn't yet for me.

https://arlen22.github.io/tiddlywiki-auto-paragraph.html

joshuafontany:
> It's supposed to work at this URL as well, but it isn't yet for me.
> 
> https://arlen22.github.io/tiddlywiki-auto-paragraph.html

Github Pages is a bit tricky at first. If you have just created the repository, you have to to to the Settings page and scroll down to the Github Pages options, and select the 'main' branch as the gh-pages repository.

Arlen22:
Ok, now I added a pragma which does the set widget automatically, but you can also do it manually if you want. Either one works, but the pragma is required if you want to affect the current file. However, two unique edge cases are the macrocall widget with text/html output and the list widget's empty message. 

- The empty message of the transclude widget is parsed as part of the tiddler it is in, so it works the same as any other text in the tiddler. 
- The list widget's empty message is a string until needed. Exactly the reverse of the transclude widget.
- A macro is parsed when it is called rather than when it is defined so the context is similar to transclusion. 
- The wikify widget takes a string and parses it. Again similar to transclusion.

There are a few edge cases like startup actions and so forth that aren't really affected. The widgets need give the parser accurate arguments so it works fine to fallback to true because it means it's probably an old piece of code that doesn't need to be updated.

joshuafontany:
Testing modals would be worth doing. I found out that modals are considered 'outside' of the story river (I think... I know I had to work around that in the JsonMangler Setting Tab modal).

pmario:
I'm sorry, but I don't understand the name. Should it add paragraphs, or should it avoid paragraphs?

```
\autoparagraph yes
\autoparagraph no
```

What do I need to add and when?

core/modules/parsers/wikiparser/rules/autoparagraph.js
```diff
@@ -0,0 +1,72 @@
...
+		if(match[1]) {
+			tokens.push(match[1]);
+		}
+		// Match the next token
+		match = reMatch.exec(this.parser.source);
+	}
+	// Set it for the current file
+	if(tokens.length > 0) {
+		if(tokens[0] === "yes") this.parser.autoParagraph = true;
+		else if (tokens[0] === "no") this.parser.autoParagraph = false;
```
pmario: What happens if token[0] is "Yes" or "hugo"?

core/modules/parsers/wikiparser/rules/autoparagraph.js
```diff
@@ -0,0 +1,72 @@
...
+	}
+	// And set it for everything transcluded or otherwise parsed
+	return [{
+		type: "set",
+		attributes: {
+			name: {type: "string", value: "tv-auto-paragraph"},
+			value: {type: "string", value: tokens[0]}
+		},
+		children: []
+	}];
```
pmario: IMO if the token is invalid, we shouldn't add anything to the parsetree.

core/modules/parsers/wikiparser/rules/autoparagraph.js
Arlen22: It just gets ignored, but still goes into the set widget.

core/modules/parsers/wikiparser/rules/autoparagraph.js
Arlen22: Any token value specified will get forwarded to the set widget. I think it should be that way to allow any desired value to be specified in case there are more complex use cases.

Arlen22:
> I'm sorry, but I don't understand the name. Should it add paragraphs, or should it avoid paragraphs?
> 
> ```
> \autoparagraph yes
> \autoparagraph no
> ```
> What do I need to add and when?

I'm sorry, I'm not understanding your question. From the user's perspective, this turns the auto-paragraph insertion feature on or off in the parser. If you set `\autoparagraph no` then a double line break will be ignored and your essay will come out looking like a wall of text (except for block mode things like lists, etc).  But this is exactly what you want to have happen when you're designing a UI that's going to be styled with flexbox or even just regular CSS. You don't want random p tags keeping you from organizing your code properly or forcing you to modify core tiddlers just to get your theme to work because there's an extra line break somewhere.

heronils:
This should be done - but not based on a userdefined setting but automatically, based on the type of the child element! Let me explain:

This is the problem (at least for me, you devs may see this completely different):

TiddlyWiki sometimes generates this html: `<p><div>...</div></p>`

Now static output will be parsed as this by the Browser:

```
p
div
    ...
// followed by an invalid p closing tag
```

Because of [optional tags](http://w3c.github.io/html/syntax.html#optional-tags) the p will be implicitely closed before the div. The resulting html is structurally invalid.

DOM generated via Javascript will be parsed different:

```
p
    div
        ...
```

which is structurally valid, but semantically invalid, because the div is no [phrasing content](https://www.w3.org/TR/html5/dom.html#phrasing-content).

This means, that, amongst other problems, CSS authors have to write different styles based on the type of output generated.

The solution is relatively simple, if i am not wrong. Arlen was on the right path: [In this code line](https://github.com/Jermolene/TiddlyWiki5/pull/3854/commits/d5d893aa36d45f034256fd046a52e24c9f40eb09#diff-778fdf646dead05c92f203795b4009cbR222), when we have the list of children available, we do the following:

1) we create an empty list _newChildren_

2) we compare the tag of each direct child of the original _children_ list against the following set (taken from the above first link, may be that this set is actually shorter, because some of those elements will never appear in this context, i dont know TiddlyWikis internals good enough - may be that this list is just div, blockquote, fieldset, pre, table, lists and headings):

```
address, article, aside, blockquote, details, div, dl, fieldset,
figcaption, figure, footer, form, h1, h2, h3, h4, h5, h6,
header, hr, main, nav, ol, p, pre, section, table, ul
```

3) if the childs tag is in that list, we append the child to _newChildren_, otherwise we wrap the child in a p and then append that p to _newChildren_

4) we return _newChildren_

Maybe this can be done more performant, on the child level, i dont know. Again, i dont understand TiddlyWikis Internals good enough. Jeremy Ruston knows much better how the optimal implementation would look.

To give a concrete example, in [Arlen22s example](https://arlen22.github.io/tiddlywiki-auto-paragraph.html#%24%3A%2FControlPanel) the _p_ wrapping the _div.tc-control-panel_ in the sidebar is gone, and that is good. But the _p_ wrapping the same _div.tc-control-panel_ inside the controlpanel tiddler is still there. It should also go. Concrete, the inner html of the Tiddler in the Story View should be

```
div.tc-control-panel
    ...
```

instead of how it currently is:

```
p
    div.tc-control-panel
        ...
```

BUT if i edit that Tiddler, and replace

```
<div class="tc-control-panel">...</div>
```

with

```
Test
<div class="tc-control-panel">...</div>
Test
```

then it should be

```
p
    "Test"
div.tc-control-panel
    ...
p
    "Test"
```

instead of the current

```
p
    "Test"
    div.tc-control-panel
        ...
    "Test"
```

If i am not wrong, this can be achieved with the checking of the list of childrens, described above. All other changes by Arlen are not needed, It would automagically work out of the box. Correct me if im wrong.

heronils:
There is an error in my thinking.

If we have children like this (taken from ParsingMechanism.tid)

```
[
	{type: "text", text: "Some "},
	{type: "element", tag: "em", children: [
		{type: "text", text: "italics"}
	]},
	{type: "text", text: " and a "},
	{type: "tiddler", attributes:{
		tiddler: {type: "string", value: "Transclusion"}
	}, children:[
		{type: "transclude", attributes:{
			tiddler: {type: "string", value: "Transclusion"}
		}}
	]},
	{type: "text", text: "."}
]
```

then we dont want to wrap every child which is phrasing content into a p, because then the em would be wrapped into a p. Instead we want to go through the list, and when we detect the first phrasing content element, we generate a p. then we append this child and all following phrasing content to that p. (here the first three) As soon as we detect the first non phrasing content (here the tiddler) we append the p and then the tiddler to the _newChildren_. For the last element (the ".") we then once again create a new p, append the "." to it and the p to _newChildren_. Should there exist inline tiddlers which are not divs (but eg spans) then they have to contain this information so that they will be appended to _p_s like other phrasing content. We will need to set the variable referencing the p to null after we appended the p to _newChildren_, so that the "." knows that it needs to generate a new p.

heronils:
Maybe it is even simpler to do this during the rendering phase. We leave the tree as it is currently, but when rendering we create a [documentFragment ](https://davidwalsh.name/documentfragment) instead of the p DOM node. Then we create the neccessary ps (or not) for the child elements according to the above mechanism, and then append the fragment.

Arlen22:
The problem I see with this is that the p tags actually do contain divs at least in Chrome. So that makes things even more complicated.

The problem isn't whether or not there is phrasing content involved. The problem is that too many times we have to write user interface wikitext in a way that ends up causing rogue p tags to be inserted. Rendering user content with auto-paragraph on allows users to keep their paragraphs and rendering the rest of the user interface with it off makes it easier to style. In my mind, we have to keep it there for user content so we don't break existing wikitext of end users and we have to turn it off for code that's intended to generate HTML so we don't have this big mess. 

Technically it's only half the mess, but that's for another PR.

Arlen22:
If the flag is set at the top of a tiddler in the story river, it will affect the rendering of that tiddler, so it is still possible to turn off autoparagraph for a particular tiddler, and I am assuming that most plugin authors would do that as needed for consistency.

pmario:
I think the p-tag in UI elements should be globally removed. We just need to fix some CSS margin-bottoms in the sidebar. The overall consistency will be better.

heronils:
@pmario @Arlen22 

i would do it like this:

(Definition: _pc_ = [phrasing content](https://www.w3.org/TR/html5/dom.html#phrasing-content) (stuff inside a paragraph), _npc_ = non-phrasing content (stuff outside of paragraphs))

First, i would fundamentally separate between _inline Tiddlers_ and _block Tiddlers_. Inline Tiddlers are _pc_ (`display:inline;`) and Block Tiddlers are _npc_ (`display:block;`).

Block Tiddlers can contain everything.

Inline Tiddlers are allowed everywhere but only _pc_ and other inline Tiddlers are allowed within.

Second, if the user uses the `whitespace trim` pragma, then _p_ s are inserted nowhere in the level ([users seem to expect it to behave like this](https://groups.google.com/d/msg/tiddlywiki/VBrRCeYka1I/PyZedy6cBQAJ)) -  it behaves like autoParagraph off on that level.

If the user does not use that pragma, then, inside of block Tiddlers there are two cases:

* The user explicitely types two newlines after _pc_ and _pc_ follows
* _npc_ is followed by _pc_ or _pc_ is followed by _npc_

Here the _pc_ always gets wrapped into a _p_, but the _npc_ doesnt.

Examples of content inside of a block Tiddler (in inline Tiddlers _p_ s are never added, instead a newline will simply be replaced with a _br_):

* `foo bar` --> `<p>foo bar</p>`
* `foo\nbar` --> `<p>foo<br\>bar</p>`
* `foo\n\nbar` --> `<p>foo</p><p>bar</p>`
* `foo\n\n\nbar` --> `<p>foo</p><p></p><p>bar</p>`

(There could be an option to reset this to the way it is currently done in TiddlyWiki, for backwards  compatibility).

If the content contains a block Tiddler, eg a TOC:

* `foo BLOCKTIDDLER` --> `<p>foo</p><block-tiddler>...</block-tiddler>`
* `foo\nBLOCKTIDDLER` --> `<p>foo</p><block-tiddler>...</block-tiddler>`
* `foo\n\nBLOCKTIDDLER` --> `<p>foo</p><block-tiddler>...</block-tiddler>`
* `foo\n\n\nBLOCKTIDDLER` --> `<p>foo</p><p></p><block-tiddler>...</block-tiddler>`

If the content contains an inline Tiddler, eg an autogenerated date:
* `foo INLINETIDDLER` --> `<p>foo <inline-tiddler>...</inline-tiddler></p>`
* `foo\nINLINETIDDLER` --> `<p>foo<br\><inline-tiddler>...</inline-tiddler></p>`
* `foo\n\nINLINETIDDLER` --> `<p>foo</p><p><inline-tiddler>...</inline-tiddler></p>` 
* `foo\n\n\nINLINETIDDLER` --> `<p>foo</p><p></p><p><inline-tiddler>...</inline-tiddler></p>`

I would actually use these tag names (_block-tiddler_ and _inline-tiddler_), because, while unspecified, [this will validate in HTML5 as long as the tag name contains hyphens](https://stackoverflow.com/questions/10830682/is-it-ok-to-use-unknown-html-tags/27869027#comment72073089_27869027). As _Tiddler_ is its own semantic concept, it consequently deserves its own tag. And it makes the html source and the stylesheets look more pretty.

Notice that the inserted empty _p_ s are no problem (as long as they contain nothing, not even ' ') because, if a developer needs this, they can be targeted using the [`:empty` selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:empty) (which is, aside of the increased semantic meaning, the main reason why using _p_ instead of two _br_ s is good if we are inside of a _npc_).

Arlen22:
@heronils I think that this is a modification of the autoparagraph feature of TiddlyWiki. What we are running into in this case is that we want the autoparagraph feature completely disabled for certain parts of the UI because it causes a lot of problems that have nothing to do with phrasing content at all. 

In other words, p tags are getting inserted in places where there is no phrasing content involved (within the sidebar layout divs, for instance) and so it is causing significant problems. 

If you are willing to open a separate issue with this information in it I think that would be a good idea, because it only applies to the cases where the autoparagraph feature is turned on, which is currently the default. In this PR, I am trying to make a way for the autoparagraph feature to be explicitly turned off in certain cases where there is no phrasing content involved and where double line breaks are not supposed to insert p tags ever, and so it doesn't really relate to what you're discussing. What you're saying is, I think, completely valid, but I am working with other content than what you're discussing. 

Thank you very much for your interest in this discussion. I'm not trying to chase you away, because TiddlyWiki needs a lot more people than just me! I just think what you're saying is unrelated in this case and should be in a new Issue. I hope that makes sense.

Arlen22:
@heronils I apologize greatly if I sound terse or rude. I don't mean to be that way at all. I know I sometimes have a hard time knowing how to explain myself. I think the first paragraph sums up what I'm trying to say.

Jermolene:
Hi @Arlen22 @heronils it's great to open a conversation about this. These issues have languished for a long time and I'm certainly keen to explore how we might address them within our backwards compatibility constraints.

To be frank, the HTML generation in TW5 is pretty severely broken from the perspective of the W3C standards. In retrospect, back in 2011/12 I was oblivious of the full impact of nesting rules, largely because my experience of HTML tends to be via the DOM, which is much more permissive than HTML itself.

Prior to @Arlen22's PR, my thinking was that the fundamental problem is that TW5 tries to resolve almost all the block vs. inline handling and paragraph generation at parse time. In fact, any consideration of the nesting rules quickly shows that we need to do it at render time.

I think you're both thinking along compatible lines. I would see @Arlen22's proposal as a cunning hack that could ameliorate the most egregious problems with paragraphs. Meanwhile, @heronils is considering the broader problem and therefore necessarily fundamentally rethinking the logic.

I'd like to explore both approaches:

* Hacks along the lines of this PR and the earlier `\whitespace trim` to ameliorate the problems
* Establishing a new, sensible parser that can be used in parallel to the existing `text/vnd.tiddlywiki` parser

There's a great opportunity with the second approach to fix other fundamental problems with TW5's design, and add important new features. One thing I'm particularly keen on is to explore making parse rules be user defined, so that people can evolve their own custom markup. Something like this:

```
\parserule italics
start-match: //
end-match: //
output-macro: base-italics
\end
```

Given the foregoing, the approach I think we should take with evaluating this current PR is to ensure that we're not restricting our ability to rethink things in the future. One thing I'd like to look at is how easily we can eradicate the extraneous `<p>` tags in the UI using the existing tools at our disposal. A lot of the UI templates in TW5 are very, very old (up to 5 years old), and we have some new techniques at our disposal.

Arlen22:
I am quite sure this does not prevent future improvements as it simply prevents p tags from being inserted in parts of the layout where we definitely do not want them. It also makes it easy for people to selectively turn this feature on or off in their tiddlers. Sometimes we just want it to behave like an XML-ish widget tree without trying to wrangle with the wrenches this particular feature throws in the works. We don't want line-breaks to specify phrasing content, we just want everything to be treated as an XML tree structure and will gladly write `<p></p>` wherever we need it (which is rare) in order to keep the parser from throwing a wrench in the works. Browsers don't mind rendering plain text between elements, so it doesn't cause any real problems in the end.

heronils:
Hi @Jermolene , @Arlen22 , sorry for late answering.

Ok, As Jeremy suggested that it should be done in the rendering phase – which makes sense – i tried to replace the function `Widget.prototype.makeChildWidgets` in `C:\Users\Nils-Hero\AppData\Roaming\npm\node_modules\tiddlywiki\core\modules\widgets` with the following one. Set the _mode_ variable to `'legacy'`, `'xml'` and `'html5'`, then check the DOM of the Wiki. It is also documented in the source. I dont know if this is the right approach. Please review.

```javascript
/*
Make child widgets corresponding to specified parseTreeNodes
*/
Widget.prototype.makeChildWidgets = function(parseTreeNodes) {
    this.children = [];
    var self = this;

    // Todo: The default global mode should of course be read from eg. an option settable in the
    // control panel. I dont know how to do this.
    //
    // Possible modes are:
    //      'legacy'    the current mode used by TiddlyWiki.
    //      'xml'       insert no ps at all (Arlen22´s mode).
    //      'html5'     insert ps based on the tag name of the child nodes (my mode).
    let mode = 'html5';

    // Define what is non phrasing content. Only used by the html5 mode.
    let nonPhrasingContent = new Set([
        'address', 'article', 'aside', 'blockquote', 'details', 'div', 'dl', 'fieldset',
        'figcaption', 'figure', 'footer', 'form', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'header',
        'hr', 'main', 'nav', 'ol', 'p', 'pre', 'section', 'table', 'ul'
    ]);

    // this is how it is originally done.
    if (mode == 'legacy') {
        $tw.utils.each(parseTreeNodes || (this.parseTreeNode && this.parseTreeNode.children),function(childNode) {
            self.children.push(self.makeChildWidget(childNode));
        });
    }

    // Arlen22´s autoparagraph mode (i renamed it to xml mode). Strip away all p´s.
    // This currently makes not much sense as it is currently done globally. We need a local
    // setting to just temporarily set this mode. Could eg. be set and read from the widget tree.
    // Definable by eg. something like
    //      // Jeremys (current) | Arlen22´s | mine | reset to global setting
    //      <$parse mode="legacy|xml|html5|global">
    //          ...
    //      </$parse>
    // or a pragma or whatever. I dont know how to do this. Discuss.
    //
    // This mode does not fix semantically invalid html like '<p><div></div></p>' entered by the
    // user into a div (which may be a feature and not a bug).
    else if (mode == 'xml') {
        $tw.utils.each(parseTreeNodes || (this.parseTreeNode && this.parseTreeNode.children),function(childNode) {
            if(childNode.type == 'element' && childNode.tag == 'p') {
                for (let psChildNode of childNode.children) {
                    self.children.push(self.makeChildWidget(psChildNode));
                }
            } else {
                // if this is not a p, proceed like in the legacy mode.
                self.children.push(self.makeChildWidget(childNode));
            }
        });
    }

    // In this mode for example the first div.tc-control-panel in the control panel or the
    // div.tc-tab-set in the sidebar will not be wrapped in a p any more. However, i found out
    // that this does not work everywhere.  Eg. the div.tc-search in the sidebar is still wrapped
    // inside a p. I guess that is because the child iswrapped into some <$...> element.
    //
    // Also, like in the xml mode, broken html manually entered into a tiddler will not be fixed.
    else if (mode == 'html5') {
        $tw.utils.each(parseTreeNodes || (this.parseTreeNode && this.parseTreeNode.children),function(childNode) {
            if(childNode.type == 'element' && childNode.tag == 'p') {
                //~ console.log('we are in a p');
                let p = self.makeChildWidget(childNode);
                for (let psChildNode of childNode.children) {
                    if(psChildNode.type == 'element' && nonPhrasingContent.has(psChildNode.tag)) {
                        //~ console.log('we are in a child of p, which is non phrasing content');
                        if (p.children.length) {
                            self.children.push(p);
                            p = self.makeChildWidget(childNode);
                        }
                        self.children.push(self.makeChildWidget(psChildNode));
                    } else {
                        //~ console.log('we are in a child of p, which is phrasing content');
                        p.children.push(p.makeChildWidget(psChildNode))
                    }
                }
                // End of the level reached, insert the remaining stuff.
                if (p.children.length) {
                    self.children.push(p);
                }
            } else {
                // if this is not a p, proceed like in legacy mode.
                self.children.push(self.makeChildWidget(childNode));
            }
        });
    }
};

/*
Before i implemented the above, i logged some infos. Then i played around with the wiki a bit,
and after i while i got output like the following (sorted). Those are child nodes of p´s.
Probably an incomplete list. First is the attribute 'type', second is the attribute 'tag'.

action-deletetiddler $action-deletetiddler
action-sendmessage $action-sendmessage
button $button
checkbox $checkbox
edit-text $edit-text
element a
element code
element div
element em
element i
element span
element strike
element strong
element sub
element sup
element u
element ul
image undefined
link undefined
linkcatcher $linkcatcher
list $list
macrocall $macrocall
macrocall undefined
radio $radio
reveal $reveal
select $select
set $set
text $text
text undefined
tiddler $tiddler
tiddler undefined
transclude $transclude
transclude undefined
vars $vars
*/

/*
// This is what i used to log this stuff.
function pp(t,n) {console.log(t + ':', n.type, n.tag);}
function log(m){console.log(m)}
function isa(e, tag) {return e.type == 'element' && e.tag == 'p'}
let coll = new Set();
let size = 0;

Widget.prototype.makeChildWidgets = function(parseTreeNodes) {
    this.children = [];
    var self = this;

    $tw.utils.each(parseTreeNodes || (this.parseTreeNode && this.parseTreeNode.children),function(childNode) {
        if(isa(childNode, 'p')) {
            for (let psChildNode of childNode.children) {
                let k = `${psChildNode.type} ${psChildNode.tag}`;
                coll.add(k);
            }
        }
        self.children.push(self.makeChildWidget(childNode));
    });
    if (size < coll.size) {
        log('--------------------');
        for(let e of coll) {
            log(e);
        }
        size = coll.size;
    }
};
*/
```

Arlen22:
The auto-paragraph feature needs to be turned off globally in the UI, but not within displayed tiddlers because we want things to be precise when writing HTML that affects the page layout, but we need to keep the current behavior when rendering tiddlers so we don't break backward compatibility. 

I'm actually having a hard time understanding all this as it's unrelated to what I'm trying to accomplish and I don't have time to investigate it properly, but I think I'm starting to.

```js
    else if (mode == 'html5') {
        $tw.utils.each(parseTreeNodes || (this.parseTreeNode && this.parseTreeNode.children),function(childNode) {
            if(childNode.type == 'element' && childNode.tag == 'p') {
```

I think the proper way to do this is to disable auto-paragraph in the parser, which changes the parse tree node (`childNode.type == 'text'`). Then we should iterate through the children (which includes both elements and text), and if we see paragraph-like content we would add the appropriate p tags. This just kicks the ball further down the road but it does allow the renderer to examine the child elements that have been specified. 

The problem that I see is that we still need to make an educated guess at what is text and what is xml. 

 - We could make exception for the HTML elements that are allowed inside paragraphs, but when we start making exception for widget elements we are instantly back to the whole point of this PR. My brain overheats trying to make exceptions for widgets. 
 - If the next non-whitespace after a double line-break is an element, then it would be xml, but if it is text then it could be a paragraph, except for HTML elements allowed inside paragraphs.
 - If you write an explicit p tag, everything inside it would be forced to conform to the html standard. 
 - A widget element with text immediately after it (e.g. list widget), without a line break, will be 
considered a paragraph, but not if it has another widget or a non-paragraph html element ANYWHERE in the paragraph. 
 - If the p tag is allowed to have p tags inside it, then when the user explicitly enters a p tag, double line-breaks would function the same as when inside a div, otherwise, they would be inserted into the dom verbatim as text.

Even with everything I wrote, and everything you wrote, I still want the autoparagraph feature completely turned off for all UI code. I think it just makes sense because it makes it a lot harder to use the widget tree to generate clean HTML that you can properly style. It throws all kinds of problems into the mix. Everything you are proposing applies to rendered tiddler content -- not UI code -- and I think it's a good thing to consider, but it's too hard to work with when you're trying to just generate predictable clean HTML.

Arlen22:
If someone sees the right way to do it, I'm not saying don't do it. I just run into too many problems figuring out how to do it.

heronils:
@Arlen22 There is a serious 'bug' in my above code (actually it is just me misunderstanding the internals). I need not just to move around the widgets but i need also to detach the parse trees of the _non phrasing content_ childs of p and append them to the parse tree of the parent of the p (At least that is my guess how to fix it). Currently, when in html5 mode, it will for example show a duplicated search entry in the _$:/AdvancedSearch_ tiddler. And the _get more plugins_ overlay is also duplicated. I guess the xml mode also is broken. Working on it and posting an update soon.

I also just become aware that this is not clean and that we actually need to do this in the parser, as @Jermolene pointed out already. I have an [interesting idea](https://pegjs.org/online). But first im gonna fix this. I then want to implemented the xml mode in the parser.

I am just posting this to inform you that im currently working on the fix of the above code, so that we dont do duplicate work.

heronils:
BTW, [this](http://cjhunt.github.io/) was helpful

pmario:
Browsers simply don't care what you put inside a p-tag, even if the HTML spec says it's wrong. Nobody can avoid "out of spec" html code. ... I think we should deal with wikitext content, the same way as browsers. Leave it, as it is at the moment. 

IMO this PR is about the UI and p-tags don't belong there.

heronils:
I posted the fix of my previous code example [here](https://github.com/Jermolene/TiddlyWiki5/issues/3913). I have removed the xml mode, see below. But i will re-add it if you want it.

@pmario may i point you to [this post of mine](https://github.com/Jermolene/TiddlyWiki5/pull/3854#issuecomment-478205255)? i explained there why it is a problem – dynamic and static output will be parsed differently by the browser, which results in for example styles like `p > div` not function properly in static output.

---

Regarding the autoparagraph feature:

I have thought about this again and i think that this is actually not needed. Here is the reason why:

### Empty p tags

TiddlyWiki does not generate empty p tags, so this is not a problem for the interface designer. Tiddly Wiki devs also do not generate empty p tags manually, why would they? Users may generate empty p tags manually, but that is not the business of the interface designer. In his development environment he does not generate empty p tags and if the user does, it is his problem.

### P tags with non phrasing contents in them

Such p tags are a problem and should be removed. Manually from the Tid files and long term by improving the parser. But this can done automatically with my approach.

those p tags manually generated by the user are not the problem of the developers. We can add an entry to the wiki explaining users why they should not put non phrasing content inside of p tags and that it results in different types of DOM (because the p tag will be [implicitely closed](http://w3c.github.io/html/syntax.html#optional-tags)). Im willing to write that entry.

### P tags with phrasing content inside them

I dont see why this is a problem. The interface designer can avoid them by not inserting double line breaks in his templates. For example, if you remove the first line break in the _$:/core/ui/SideBarSegments/search_ tiddler, with my [html5 mode](https://github.com/Jermolene/TiddlyWiki5/issues/3913) enabled, then the search entry will be not be wrapped in a p any more. All we need is some sensible review of the core tiddlers.

Removing all p tags outside of the story river also creates problems for CSS design, because this text can then in some cases not easily be accessed, for example, try to hide the 'foo', but not the 'bar' in `<div><p>foo</p><span>bar</span></div>` and in `<div>foo<span>bar</span></div>`.

So, all in all i now see no need for adding the autoparagraph feature. We should just, as @Jermolene suggested already, review the core tid files and work on getting my html5 mode into the parser.

pmario:
> ... All we need is some sensible review of the core tiddlers.

I think that's a good point. We need to improve readability AND have a close look, at the actual DOM, that is created. ... eg: remove a lot of redundant class-names, that can't be used for styling anyway. 

The TW parser has 2 modes: **inline** and **block** mode. The parse-logic goes a long way to detect (guess) the mode, that is intended by the user. 

1st essential rule for prose text is: 2 line-feeds mean to start a new paragraph. 
2nd rule is: 2 line-feeds also start the block-mode for eg: transclusions. 

At the moment TW UI templates are "pure" wikitext and both rules are applied. ... **Rule 1 is always wrong for UI templates**. Because we know it's wrong, we could enforce the rule, by hard-coding it.

We also know, that making rules configurable is preferable over enforcing them.

So I think, we need a dynamic way to switch paragraph creation on or off. Similar to the PR. 

-----------------------

@heronils ... You point out that paragraphs should only contain "phrasing" content. .. That's right. But I think it's close to impossible, to do so. eg: 

```
some prose text can contain {{a transclusion}}. 

a second paragraph
```

If the "a transclusion" tiddler looks like this: 

```
<div>some text<div>
```

We can't avoid the creation of invalid DOM, because the user told us to do so. ... The only way to fix this, is "best practice documentation".

Arlen22:
@Jermolene, what's the chance that we could include this PR in 5.1.20?

The other part of this long discussion only relates to when autoparagraph is turned on and I don't really want to worry about that part right now. I'm not in that section of the work and I'm fine with letting someone else worry about those things. 

This feature has been a real hassle in UI design and makes code very finicky so I'm wondering if we could add this to give ourselves some more room to work without worrying about random p and div tags. Without this, they're almost unavoidable because transclusion throws a wrench into the works for some reason. I've seen random p or div tags pop up and then move around when I try to correct the problem. Those random tags popping up make it almost impossible to style the UI without touching the core code, which is definitely not the intended result of this feature. 

I'm hoping to get this into 5.1.20 so I can start working on some of the major themes that I've envisioned in the past. But I guess it depends on what this all affects, although I can't really think of any negatives, because I think all theme designers will have an easier life.

Jermolene:
> This feature has been a real hassle in UI design and makes code very finicky so I'm wondering if we could add this to give ourselves some more room to work without worrying about random p and div tags. Without this, they're almost unavoidable because transclusion throws a wrench into the works for some reason. I've seen random p or div tags pop up and then move around when I try to correct the problem. Those random tags popping up make it almost impossible to style the UI without touching the core code, which is definitely not the intended result of this feature.

I appreciate the problem. For the most part, paragraphs can be avoided by using `\whitespace trim` and making sure that you use inline transclusion. I know that's not ideal, but it does make it possible to avoid inadvertent paragraphs. I'm afraid that that rules it out for v5.1.20 at this late stage.

Stepping back, I feel that this PR rather adds to our existing tower of hacks. It introduces a new mechanism for controlling parsing via a variable, with a lot of attendant complexity. It's a devilish sort of complexity: lots of conditional code (the good sort of complexity is e.g. filters where we have a large number of small similar but different components). We may be better off stepping back a bit and revising more of the parser architecture.

In another thread about the wikitext parser I make the point that the fundamental problem is that paragraph creation should not be done at parse time, but at render time. Otherwise it's impossible to respect HTML's nesting rules (which we thoroughly ignore at present).

Arlen22:
I'm trying to turn off paragraph creation, as this seems to be the one thing that prevents the widget tree from being completely declarative and intuitive. To me, the auto-paragraph feature is the hack, and this is the first step to removing it from a good portion of our code base where it is never wanted. I think if this would happen, the widget tree would finally be completely predictable. I don't understand block vs inline, because I'm usually just trying to generate an HTML tree anyway, but I never really run into them. If whitespace trim accomplishes that, then I'm OK with that, although it feels like this solution is more global and should have been there from the beginning. 

That's just my two cents.

heronils:
The waiting may have ended :-)

pmario:
@Arlen22 ... There's something wrong with your PR. 452 files changed. ...

Arlen22:
I was trying to reset the branch to the current master. Does anyone know what the correct way to do that is or does that invalidate the PR?

Arlen22:
What's wierd is the compare page looks normal. I guess I could just open a new PR. 

https://github.com/Jermolene/TiddlyWiki5/compare/master...Arlen22:autoparagraph

PR #3857, opened 2019-03-12
fix jumpy pop storyview: don't x-overflow page

BurningTreeC:
... this makes the animation when inserting tiddlers / navigating to tiddlers in the pop storyview less jumpy

it simply sets `overflow-x` to `hidden` for the time of the insert-animation


BurningTreeC:
Ping

Jermolene:
Thanks @BurningTreeC

PR #3858, opened 2019-03-13
add EditTabIndex ControlPanel setting

BurningTreeC:
this adds a tabindex option in the ControlPanel without setting a default tabindex


Jermolene:
Hi @BurningTreeC what's the motivation for exposing this setting?

Renaming all those language strings has a significant knock-on impact on all the existing translations, and the translators edition.

BurningTreeC:
Hi @Jermolene, the motivation is that we don't have a default for this and it will likely stay undiscovered

I woul however make `1` default and close this PR

the language strings have not been renamed, just ordered alphabetically like its done everywhere else

PR #3862, opened 2019-03-14
invoke th-changed-tiddlers hook when there are ...

BurningTreeC:
changed tiddlers ...

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -153,6 +153,7 @@ exports.enqueueTiddlerEvent = function(title,isDeleted) {
 			self.changedTiddlers = Object.create(null);
 			self.eventsTriggered = false;
 			if($tw.utils.count(changes) > 0) {
+				$tw.hooks.invokeHook("th-changed-tiddlers",changes);
 				self.dispatchEvent("change",changes);
 			}
 		});


Jermolene:
Hi @BurningTreeC it looks like this is a synchronous version of the existing "change" event on the wiki, which is undesirable because it will encourage developers to sidestep the correct way of doing things. Why is it needed?

BurningTreeC:
Hi @Jermolene, I have no clue, I'm just trying to get a grip on the refresh functionality, looking for

1) a way to implement a refresh method from outside the scope of a widget
2) a way that does things "the way it's supposed to be done"
3) find suitable places for hooks to intercept refresh cycles

... I'm trying by trial and error and I more and more feel like wasting my time here, but that's the way I learn

Arlen22:
The line you added is next to the current method, which is the wiki event handler. You can listen for the change event by using the global `$tw` object to access the wiki.

```ts
$tw.wiki.addEventListener("change", function(changes:{[title: string]: "deleted" | "modified"}){ 
    
});
```

If we were going to use the hook for some reason, we would have to at least add the `self` variable as an argument to the hook: `$tw.hooks.invokeHook("th-wiki-changed",self,changes)`

BurningTreeC:
Hi @Arlen22 , thanks for your help!

I'm not sure anymore what I'm even looking for and I close this, the :bulb: may come at some point

PR #3863, opened 2019-03-15
add th-refreshing-input hook to factory.js

BurningTreeC:
this adds a `th-refreshing-input` hook in the EditTextFactory's refresh method

it passes the widget, the editInfo, the changedTiddlers object and the newly computed attributes

### DIFF
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -214,7 +214,6 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 			this.refreshSelf();
 			return true;
 		} else if(changedTiddlers[this.editTitle]) {
-			$tw.hooks.invokeHook("th-refreshing-input-text",this,editInfo,changedTiddlers,changedAttributes);
 			this.updateEditor(editInfo.value,editInfo.type);
 		}
 		this.engine.fixHeight();
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -206,13 +206,14 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
 	*/
 	EditTextWidget.prototype.refresh = function(changedTiddlers) {
-		var changedAttributes = this.computeAttributes();
+		var changedAttributes = this.computeAttributes(),
+			editInfo = this.getEditInfo();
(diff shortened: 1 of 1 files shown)


BurningTreeC:
this hook can be used to update codemirror's options on the fly without having to open a new cm instance in order to take effect

I'm also successfully using it for `arrow-key selectable inputs`

core/modules/editor/factory.js
```diff
@@ -206,13 +206,14 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
 	*/
 	EditTextWidget.prototype.refresh = function(changedTiddlers) {
-		var changedAttributes = this.computeAttributes();
+		var changedAttributes = this.computeAttributes(),
+			editInfo = this.getEditInfo();
+		$tw.hooks.invokeHook("th-refreshing-input",this,editInfo,changedTiddlers,changedAttributes);
```
Jermolene: This hook is specific to the edit text widget so it should be called something like `th-edit-text-widget-refreshing`

Jermolene:
Hi @BurningTreeC I am still wondering whether we might not be better off approaching this by (a) implementing a general purpose mechanism for plugins to subclass core widgets and then (b) refactoring the methods of these widgets so that there are the right methods for you to override.

core/modules/editor/factory.js
BurningTreeC: Hi @Jermolene , I'll change it

BurningTreeC:
Hi @Jermolene,

> I am still wondering whether we might not be better off approaching this by (a) implementing a general purpose mechanism for plugins to subclass core widgets and then (b) refactoring the methods of these widgets so that there are the right methods for you to override.

Yes, I didn't know anything about such a plan, haven't seen any work in that direction. I don't know how to make that mechanism and so I'm limited to simple solutions, sorry for that. But it would be very welcome to many people out there

core/modules/editor/factory.js
BurningTreeC: @Jermolene, the name is changed to `th-edit-text-widget-refreshing`

note: the edit widget is also accessible through this, it's a direct parent so no need for a second hook if one wants to access it

Jermolene:
Hi @BurningTreeC I've made a ticket with a proposal for a widget subclassing mechanism -- see #3865.

Issue #345, opened 2014-01-14
Introduce vocabulary parameter to wikitext content type


Jermolene:
To aid in the interchange of content, it is proposed that we introduce a parameter to the TiddlyWiki5 content type (`text/vnd.tiddlywiki`) to specify the vocabulary with which the text is intended to be parsed. For example:

A vocabulary specified via a URL:

```
text/vnd.tiddlywiki; vocab=http://vocabs.tiddlywiki.com/standard.json
```

A vocabulary defined in a local tiddler:

```
text/vnd.tiddlywiki; vocab=myBirdwatchingVocabulary
```

In either case, the vocabulary contains information specifying:
- The parse rules to be applied to the text
- Optional configuration parameters for those parse rules

It might be useful for vocabularies to be able to cascade, so that one could describe a custom vocab as, say, "standard TW markup plus these three plugins".

Telumire:
I like that idea a lot. Maybe instead of "vocab", this could be called "xmlns" or namespace, to follow the html spec ?

Jermolene:
> I like that idea a lot. Maybe instead of "vocab", this could be called "xmlns" or namespace, to follow the html spec ?

"namespace" might work, but seems a little technical. I think of these vocabularies as being very visible for end users. The dream would be that we'd evolve vocabularies for specific niches (eg a vocabulary for dance choregraphers that incorporates a notation for dance moves, or a special vocabulary for writing D&D games).

PR #3876, opened 2019-03-22
Replace Markdown parsing library with Remarkable

### LINKED ISSUE #3108: markdown plugin does not work with links with '
Something like `[foo](It's about me)` does not work. And this `[foo](It about me)` does

### LINKED ISSUE #3338: markdown plugin maruku variant does not work with fenced code blocks
[This](https://michelf.ca/projects/php-markdown/extra/#fenced-code-blocks) does not work
captura de pantalla_2018-06-21_19-36-38

### LINKED ISSUE #3341: Replace markdown-js with another markdown plugin
There are a lot of errors with markdown plugin right now (see [1](https://github.com/Jermolene/TiddlyWiki5/issues/3338) and [2](https://github.com/Jermolene/TiddlyWiki5/issues/3108) for instance). It seems markdown-js is not maintained (2017 is latest uptodate version and they don't respond to issues). Please replace with other markdown plugins

MidnightLightning:
fixes #3341, fixes #3338, fixes #3108

Replaces the internal library used by the "markdown" plugin to use the [Remarkable](https://github.com/jonschlinkert/remarkable) library. Remarkable has capabilities to parse out fenced code blocks and other modern addons to the Markdown language, so works well as a replacement for the prior library.

This change also modifies the Parser logic of the Markdown plugin to parse all possible Markdown entities first, but then if the Remarkable library thinks a run of text is just text, that text is also passed into the WikiText parser, such that syntax could be intermixed (allowing for an easier transition for a Tiddler that was written in WikiText, moving to Markdown).

----

I implemented this Parser for my own personal TiddlyWiki instance, but realized it could serve as a good replacement for the existing library, which there are several bug cases claiming to have issue with.

- [x] Add configuration settings for "Convert '\n' in paragraphs into line breaks", linkification, and quotes beautification settings (core Remarkable settings).
- [x] Add configuration setting for "pass through" WikiText parsing.
  - [x] Add configuration setting for pragma string to be used with WikiText parsing
- [x] Evaluate Marked library as alternative

To test, run `node ./tiddlywiki.js editions/markdowndemo --build index` and then check `editions/markdowndemo/output/index.html`.


Jermolene:
Thanks @MidnightLightning this looks good. Before we dig into the details, I wondered if you'd mind making your CLA signature be a separate PR targeting the "tiddlywiki-com" branch? Many thanks.

Jermolene:
Hi @MidnightLightning great stuff, looks good.

The fallback wikitext parsing is a great idea. I think we need to make it be configurable for backwards compatibility sake. I wonder also if we might expose some of the other configuration settings Remarkable offers.

We should also consider whether it's worth offering both the old and new Markdown plugins in the plugin library so that people can choose the old one if they need full backward compatibility.

MidnightLightning:
Sure thing! I created a separate pull request for the CLA signature.

With regard to the configuration options that Remarkable has, I had considered exposing some of those, but several configuration options are with regard to the final HTML output, not the node/tree representation (`xhtmlOut`, `langPrefix`), and for "backwards compatibility" to Markdown dialects like the original Gruber style, Remarkable _adds_ options, (not taking options away), so anything that was written in Gruber-style Markdown just simply wouldn't have fenced code blocks (etc.) in it, so wouldn't break or look different if suddenly those became available as an option?

Remarkable has a few "syntax extensions" that are off by default (`sup`, `sub`, `abbreviations`, `ins`, and `mark`) that might be good candidates for a configuration setting to enable, though?

Jermolene:
Hi @MidnightLightning

>  Remarkable adds options, (not taking options away), so anything that was written in Gruber-style Markdown just simply wouldn't have fenced code blocks (etc.) in it, so wouldn't break or look different if suddenly those became available as an option

OK, I think we'd rather not have to maintain two plugins.

> Remarkable has a few "syntax extensions" that are off by default (sup, sub, abbreviations, ins, and mark) that might be good candidates for a configuration setting to enable, though?

Yes, and maybe the "Convert '\n' in paragraphs into <br>" setting, and linkification, quotes beautification.

plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -12,78 +12,142 @@ Wraps up the markdown-js parser for use in TiddlyWiki5
...
+				attributes: {
+				tooltip: { type: 'string', value: currentNode.alt },
+				source: { type: 'string', value: currentNode.src }
+				}
+			});
+		} else if (currentNode.type == 'inline') {
+			out = out.concat(convertNodes(currentNode.children));
+		} else if (currentNode.type == 'text') {
+			let wikiParser = $tw.wiki.parseText('text/vnd.tiddlywiki', currentNode.content, {
+				parseAsInline: false
```
Jermolene: I notice that wikitext is parsed in block mode. Might that cause difficulties with it e.g. recognising `!` at the start of a string as a heading?

It may also be worth prepending a pragma to the parse string to restrict the parse rules that are applied e.g.

```
\rules only html image macrocallinline syslink transcludeinline wikilink filteredtranscludeblock macrocallblock transcludeblock
```

plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -12,78 +12,142 @@ Wraps up the markdown-js parser for use in TiddlyWiki5
 /*global $tw: false */
 "use strict";
 
-var markdown = require("$:/plugins/tiddlywiki/markdown/markdown.js");
+var Remarkable = require('$:/plugins/midnight/remarkable/remarkable.js');
```
Jermolene: There are a few coding style things to watch: TW5 uses double quotes for string constants.

plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -12,78 +12,142 @@ Wraps up the markdown-js parser for use in TiddlyWiki5
...
+		} else if (currentNode.type == 'heading_open') {
+			i = wrappedElement('h' + currentNode.hLevel, i, 'heading_close', remarkableTree);
+		} else if (currentNode.type == 'bullet_list_open') {
+			i = wrappedElement('ul', i, 'bullet_list_close', remarkableTree);
+		} else if (currentNode.type == 'list_item_open') {
+			i = wrappedElement('li', i, 'list_item_close', remarkableTree);
+		} else if (currentNode.type == 'link_open') {
+			var j = findTagWithType(remarkableTree, i + 1, 'link_close');
+
+			if (currentNode.href.substr(0, 4) == 'http') {
```
Jermolene: Coding style: we use triple equals in preference to double equals.

plugins/tiddlywiki/markdown/wrapper.js
MidnightLightning: Added as a config option now

MidnightLightning:
Okay, updated now! How's this looking, @Jermolene?

pmario:
hi ... Does Remarkable sanitize HTML input?

pmario:
Hi, I did have a closer look about the possibilities we have atm. see: https://www.npmtrends.com/markdown-it-vs-marked-vs-remarkable-vs-showdown

Why did you choose "Remarkable" instead of "marked" ... The biggest problem I have with Remarkable is, that the last commits have been 3 years ago. So it seems to be dead. 

If we include it into the "core libs" we need to maintain it :/

MidnightLightning:
Good points, @pmario! I originally looked into a few libraries, after finding [this blog post] (https://css-tricks.com/choosing-right-markdown-parser/), though I realize that's from 2016 itself. I looked into the Showdown and Remarkable libraries, and found Showdown does use a Lexer/AST style (just regexes on the original text), so harder to integrate with the TiddlyWiki Parser/Widget tree idea.

Looks like Marked [does provide access] (https://marked.js.org/#/USING_PRO.md) to the Lexer/Parser level, and the syntax of the Lexer tokens looks very similar to the Remarkable format I already coded for. I'll try dropping Marked in and see how that goes!

MidnightLightning:
All right, I took a look at Marked, and specifically at their Lexer functionality, and my assessment is that it has a few rough edges:

- Latest version is 0.6.1: That library isn't even considered up to "version 1.0" yet
- In the Lexer, Headlines store their importance level (1-6) in an attribute called "depth", which seems like a confusing label? It's not referring to how deeply nested the headline is in the DOM (e.g. inside a table, inside a div)?
- Nested block level items in the Lexer tree output don't have a way to match up what level/depth of the tree you're at. Meaning, when a "start of unordered list" node is encountered, if it has an other unordered list inside it, you can't just scan ahead for the next "end of unordered list" node, you have to note how many levels you went down. Remarkable's node tree has that information already in the nodes, making searching a lot easier.

So, in my assessment, while there does seem to be more active development around Marked, that would seem to be because there's still a lot of polish that can be put on the library. So, I'd recommend sticking with Remarkable for the time being.

pmario:
hmmm, That's sensible. Do you know the [What if?](https://groups.google.com/forum/#!msg/tiddlywiki/iC1ZNZHVU8A/quMw-cDtBAAJ) thread

MidnightLightning:
@pmario I wasn't aware of that thread, but reading through it, it seems your premise is to try and use Markdown (CommonMark) as a starting point, and layer additional Wiki functionality into it, to make "TiddlyWiki syntax" be "Markdown plus some extras"?

That sort of idea is indeed how I would use it as well, and the main reason I added in the "if Markdown just thinks this is text, hand it off to the Wiki parser to see if there's more there", to enable TiddlyWiki syntax like `<$view />` widget calls to be made in a Markdown tiddler.

I think the Abstract Syntax Tree (AST) representation of the content should be the "hub" language that all the content-types should use as a central conversion point, but as far as what language/form the tiddlers get persisted to disk as, I'm fine with a blend. I'm in the process of converting my previous note-taking system (a Git repo of a folder of Markdown files) to a TiddlyWiki structure, by using the NodeJS server method. Using the Node hosting method, the tiddlers are saved in a folder as individual files, with Markdown files getting proper `.md` extensions. That makes it very similar to my previous setup, and allows me to use external editors (namely the [GitLab Web IDE](https://docs.gitlab.com/ee/user/project/web_ide/) to edit right in the repo UI) to edit the `.md` files directly, even when I'm on a machine without the ability to host the Node engine. Going that route, having each tiddler have its own file extension based on content-type I think gives a lot of flexibility for editing the content outside the TiddlyWiki infrastructure, using whatever editor is most appropriate for that particular tiddler.

pmario:
Reading your comments, I think we are on the same line here. ... 

I think, the only thing we need to do is define a new type: `text/markdown; variant=tiddlywiki` as described [in the GG thread](https://groups.google.com/d/msg/tiddlywiki/iC1ZNZHVU8A/hoTHllayBQAJ) and [RFC 7763 SPEC example](https://tools.ietf.org/html/rfc7763#section-5)

Details about variants can be found at: [Guidance on Markdown: Design Philosophies, Stability Strategies, and Select Registrations](https://tools.ietf.org/html/rfc7764#page-10)

pmario:
previous comment updated.

MidnightLightning:
Ah, so you'd be in favor of not having a configuration parameter for "also parse Wikitext" (which would make it a global, per-wiki setting), but rather on a tiddler-by-tiddler basis (setting the content-type of an individual tiddler to a different setting)?

In my mind that seems overly-complicated; if a user doesn't want to do TiddlyWiki-specific things within their Markdown files, it wouldn't hurt to have the TiddlyWiki parsing enabled (it just wouldn't find anything additional to parse, since they didn't add anything in). In what situation would a user want to put in TiddlyWiki specific markup (e.g. `<$view />`) raw in a markdown file, and not want it parsed? If they wanted that text to show as text, they'd probably be wanting it in a code/fenced-codeblock structure, which wouldn't be wiki-parsed, right?

pmario:
I'm OK with a global setting. I just wanted to point out, that "commonmark" spec could deal with a specific setting and we probably should use it.

MidnightLightning:
> "commonmark" spec could deal with a specific setting and we probably should use it

So, you'd like to see a config option for `useCommonmark`, which would set the `breaks`, `linkify`, `quotes`, `typographer`, and `renderWikiText` config options to `false`?

Jermolene:
>  It wouldn't hurt to have the TiddlyWiki parsing enabled (it just wouldn't find anything additional to parse, since they didn't add anything in). In what situation would a user want to put in TiddlyWiki specific markup (e.g. `<$view />`) raw in a markdown file, and not want it parsed?

One situation is where the use of TW markup is accidental. For instance, double slash in TW starts italic text, which could plausibly appear in a Markdown document.

MidnightLightning:
Ah, good point! Perhaps the default `pragma` config option should only have the WikiText rules that don't have a direct Markdown representation? I'd say Markdown has ways to define `bold`, `codeinline`, `entity`, `extlink`, `hardlinebreaks`, `image`, `italic`, `prettyextlink`, `prettylink`, `strikethrough`, `wikilink`, `codeblock`, `commentblock`, `heading`, `horizrule`, and `quoteblock`, so the default config could be a `\rules except` that list?

pmario:
If we need `\rules except` I'd highly recommend the `variant=tiddlywiki` since this makes the users intent clear. ....

But there is still some potential for hickups. ... but we'll see.

MidnightLightning:
So where are we at with this? @Jermolene, do you have an opinion on the "Content-Type" of `text/markdown; variant=tiddlywiki`? The semicolon-delimited modifier on a Content-Type is in RFC7763 as a way to demark different flavors of Markdown from among its accepted standards. TiddlyWiki already uses the ["Vendor" style](https://en.wikipedia.org/wiki/Media_type#Vendor_tree) media types (defined in RFC6838) to demark Tiddlers. So, to me it comes down to whether the TiddlyWiki wiki text formatting is a "vendor-specific" formatting, or is a "markdown variant"? Personally, since it's indicating markup that's specifically not markdown-formatted, it doesn't really seem to me to be a "markdown variant", so I'd lead toward a vendor-specific content-type, if we wanted to do tiddler-specific parsing of wikitext inside a markdown-formatted tiddler.

And do we really need a global setting for "parse WikiText in markdown tiddlers"? I'm still thinking that if the way that WikiText is parsed (using a `pragma` config) removes any "accidental" WikiText formatting, then it could be on all the time (only a minor performance hit), and users that don't want it just won't use WikiText transclusions, macros, etc., and it will render just fine.

Jermolene:
> So where are we at with this? @Jermolene, do you have an opinion on the "Content-Type" of `text/markdown; variant=tiddlywiki`?

I like it (see #345 for a related proposal for the TW5 parser).

You asked over on the mailing list whether we can get this into v5.1.20. Given my usual backwards compatibility concerns, perhaps the quickest way to get this merged might be to make it be a separate plugin that is offered alongside the existing one in the plugin library. Then we can get it out there and get some feedback from the field.

Jermolene:
> So where are we at with this? @Jermolene, do you have an opinion on the "Content-Type" of `text/markdown; variant=tiddlywiki`?

I like it (see #345 for a related proposal for the TW5 parser).

You asked over on the mailing list whether we can get this into v5.1.20. Given my usual backwards compatibility concerns, perhaps the quickest way to get this merged might be to make it be a separate plugin that is offered alongside the existing one in the plugin library. Then we can get it out there and get some feedback from the field.

MidnightLightning:
Okay, I can reconfigure this as a separate plugin, but that seems a bit backward to me as a testing method (and runs the risk of falling into a limbo of never harmonizing back to one Markdown plugin and always having multiple plugins): to me that testing method seems more like "test it in production", which is generally looked upon as a not-ideal testing methodology. Will there be any "release candidates" of 5.1.20 before final release, where the community can test their Markdown content with it before finalizing it? There is a Markdown demo project in the main repo, which if it's not sufficient to cover all the use-cases for Markdown by the TiddlyWiki community, I'd encourage that any feedback received from this plugin switch of edge cases where things didn't work, that they get added as examples into that project, such that future Markdown updates can be more assured that testing before release is a good indicator of a smooth transition for users?

Jermolene:
Hi @MidnightLightning the suggestion to make it be a separate plugin is just a tactic to get the PR into v5.1.20 given that I'm not sure we've got time to settle all the outstanding questions. A subsequent release could replace the current Markdown plugin. Alternatively, we can continue as at present, perhaps merging this immediately after v5.1.20 so that it spends plenty of time in the prerelease for v5.1.21.

> Will there be any "release candidates" of 5.1.20 before final release, where the community can test their Markdown content with it before finalizing it? 

The prerelease is built continuously, and the closest we have to release candidates is when we make an announcement on the mailing list that the release is imminent and request a round of testing. In practice, we get very little in the way of bug reports at that point.

MidnightLightning:
I'm not in a rush to get this in, so would be fine with leaving it as-is and merging it in early in the 5.1.21 pipeline. 

For the content-type, should we have two different content-types handled? `text/markdown` for only Markdown content (no further processing of Tiddlywiki syntax), and `text/markdown; variant=tiddlywiki` which does attempt to recognize Tiddlywiki syntax after resolving all Markdown syntax? That would allow users to pick on a per-tiddler basis, and we could keep the default `pragma` config option for further modifying how the second content-type behaves?

Jermolene:
Hi @MidnightLightning great.

> For the content-type, should we have two different content-types handled? text/markdown for only Markdown content (no further processing of Tiddlywiki syntax), and text/markdown; variant=tiddlywiki which does attempt to recognize Tiddlywiki syntax after resolving all Markdown syntax? That would allow users to pick on a per-tiddler basis, and we could keep the default pragma config option for further modifying how the second content-type behaves?

Yes exactly that.

We do have a bit of a challenge figuring out the best way to extend the content type. You'll have seen how we use the content type as an index in various hashmaps. When processing the variant we'd want to strip whitespace, I guess? I've wondered in the past if it isn't better to keep the variant as a separate field `type.variant`. Anyhow, given that we'd like to be able to use the variant mechanism for other parsers, I think it's likely to be a fairly significant refactoring, and may be worth putting into its own PR.

pmario:
@MidnightLightning ... It seems they are back with the development. https://github.com/jonschlinkert/remarkable/releases

Can you have a look, if your PR still works?

FlashSystems:
I gave remarkable 2.0.0 a shot on my TiddlyWiki node.js install. The following steps where necessary to get 2.0.0 running:

- Replace `files/remarkable.js` with the new 2.0.0 version. I was lazy and used the minified [CDN version](https://cdn.jsdelivr.net/npm/remarkable@2.0.0/dist/remarkable.min.js).
- Modify the import in `wrapper.js` to import the new module correctly. Line 15 now reads `const { Remarkable, linkify, utils } = require("$:/plugins/tiddlywiki/markdown/remarkable.js");`
- That's almost all but `linkify` was moved to a plugin. It gives you a warning as soon as you start TiddlyWiki. So I remove it (line 28) from the remarkableOpts variable and put the loading of the plugin below the instantiation of Remarkable: `if (parseAsBoolean("$:/config/markdown/linkify")) md = md.use(linkify);`

To answer MidnightLightning's question: On first glance it seems that the upgrade to 2.0.0 does not take much effort. Maybe it should be done before merging this pull request.

Beware! This was only a quick hack to see if anything breaks and I only tested it on node.js. So please take all of this with a grain (or better a bucket) of salt.

PS: Thank you for all the effort so far. I'd really love to have a better markdown parser in TiddlyWiki :)

MidnightLightning:
Thank you @FlashSystems for doing that initial testing! I've updated this pull request to use v2.0.0 of the Remarkable library, and tested on my own system, and it appears to work for the various Markdown examples. @pmario, how's this one look now, for getting merged in?

MidnightLightning:
Checking in again; @Jermolene, can this get merged in?

FlashSystems:
I tested this iteration of the pull request (979f553) with TiddlyWiki on node.js. It displayed all my markdown documents without problems.
I'm going to keep this PR installed on my server an hope that it will be merged soon.

Jermolene:
Thank you everyone especially @MidnightLightning, I'm going to merge this now, so it will be included in v5.1.22

somenxavier:
I think this deserves a release.

PR #3880, opened 2019-03-26
Add uniquetitle macro

inmysocks:
This adds the uniquetitle macro that gives access to the $tw.wiki.generateUniqueTitle function to user space. A way to generate titles that are guaranteed to be unique has been requested in the forums multiple times, this macro does that.

### DIFF
--- a/core/modules/macros/uniquetitle.js
+++ b/core/modules/macros/uniquetitle.js
@@ -1,8 +1,8 @@
 /*\
-title: $:/core/modules/macros/uniquetitle.js
+title: $:/core/modules/macros/unusedtitle.js
 type: application/javascript
 module-type: macro
-Macro to return a unique title given a base name
+Macro to return a new title that is unused in the wiki. It can be given a name as a base.
 \*/
 (function(){
 
@@ -14,7 +14,7 @@ Macro to return a unique title given a base name
 Information about this macro
 */
 
-exports.name = "uniquetitle";
+exports.name = "unusedtitle";
 
 exports.params = [
 	{name: "baseName"},
@@ -25,7 +25,7 @@ exports.params = [
 Run the macro
 */
 exports.run = function(baseName, options) {
-	if (!baseName) {
+	if(!baseName) {
 		baseName = $tw.language.getString("DefaultNewTiddlerTitle");
 	}
 	return this.wiki.generateNewTitle(baseName, options);
--- a/core/modules/macros/uniquetitle.js
+++ b/core/modules/macros/uniquetitle.js
@@ -1,0 +1,34 @@
+/*\
+title: $:/core/modules/macros/uniquetitle.js
+type: application/javascript
+module-type: macro
+Macro to return a unique title given a base name
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Information about this macro
+*/
+
+exports.name = "uniquetitle";
+
+exports.params = [
+	{name: "baseName"},
+	{name: "options"}
+];
+
+/*
+Run the macro
+*/
+exports.run = function(baseName, options) {
+	if (!baseName) {
+		baseName = $tw.language.getString("DefaultNewTiddlerTitle");
+	}
+	return this.wiki.generateNewTitle(baseName, options);
+};
+
+})();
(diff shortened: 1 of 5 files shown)


core/modules/macros/uniquetitle.js
```diff
@@ -0,0 +1,34 @@
...
+exports.params = [
+	{name: "baseName"},
+	{name: "options"}
+];
+
+/*
+Run the macro
+*/
+exports.run = function(baseName, options) {
+	if (!baseName) {
```
Jermolene: Our coding style is not to have any white space between `if` and the following opening bracket.

core/modules/macros/uniquetitle.js
```diff
@@ -0,0 +1,34 @@
...
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Information about this macro
+*/
+
+exports.name = "uniquetitle";
```
Jermolene: I think "unique" isn't quite right; perhaps it should be called "unusedtitle" or "newtitle"? I think I prefer the former.

core/modules/macros/uniquetitle.js
inmysocks: I have been working on too many different projects lately. It is fixed.

core/modules/macros/uniquetitle.js
inmysocks: I agree. I have updated the code and documentation to unusedtitle.

core/modules/macros/uniquetitle.js
Jermolene: I sympathise! I'd love to have automated tooling to check/fix the simple stuff.

Jermolene:
Thanks @inmysocks

kookma:
Hi all,
I have two suggestions, 
I would like to have this with tm-new-tiddler, as the other method of creation tiddler has it (action-createtiddler), but if you like to have this a separate macro to be used in other usecases, 

A. 
I would suggest to let (or by default) title zero padded!

So, 

New Tiddler 001
New Tiddler 002

or if baseName: Hirad then have

Hirad 001 
Hirad 002

B. 
I would suggest to change the baseName parameter to be RENAMED to **basetitle**
first because we use title instead of name for tiddler and also it is inline with action-createtiddler

inmysocks:
kookma,

For tm-new-tiddler you can give a title, so this works with it, changing the inputs to tm-new-tiddler would make it inconsistent with other tiddler messages.

For zero padding that would be a change to how tiddlywiki makes new names in general with zero padding on all new tiddlers like New Tiddler 001, so that would be another issue or feature request.

As far as changing the name goes, being consistent is a good idea. I am not sure if I can edit this after the pull request has been merged.

Jermolene:
>  I am not sure if I can edit this after the pull request has been merged.

No problem, just start a new PR.

kookma:
One more thing:
Create a tiddler called: **MyTest** with the below content:

`<<unusedtitle Hirad>> `

and save it
It will show: **Hirad**
Then create another tiddler with title: **Hirad** and save it.
I expect the first tiddler MyTest shows now **Hirad 1** but `<<unusedtitle Hirad>> ` still shows the unusedtitle as "Hirad"

It should be noted that, if you use this inside a button, it will generate the unused title correctly!

00SS:
Seems like `<<unusedtitle>>` does not refresh when a new tiddler is created

I also agree that **baseName** should not be used. The tiddler[ tiddlywiki.files Files](https://tiddlywiki.com/prerelease/#tiddlywiki.files%20Files) defines:

_basename_ the filename of the file containing the tiddler without any extension

I also looked at all the Core Macros.
Just about every 2-word parameter is in the formats below. The only odd one is the _tag-pill_ macro.

Thus the parameter name I suggest is **baseTitle** and not __basetitle__

Thanks for the effort for making this new macro!

Macro|  two word parameters
 ----- |  -----
contrastcolour Macro |fallbackTarget colourA colourB
Table-of-Contents Macros | selectedTiddler unselectedText 
list-tagged-draggable Macro | subFilter itemTemplate emptyMessage
list-links Macro | emptyMessage
timeline Macro | dateField
tag-pill Macro | element-tag element-attributes
tabs Macro | tabsList buttonTemplate

kookma:
@00SS 
Thank you for supporting the idea!
Also the documentation in tiddlywiki.com is not inline with other documentation!
I suggest Jeremy push a standard format for tiddler contains the macro documentation (a template) , like the one I used in Yazd

Template: https://tw-scripts.github.io/Yazd/#List%20of%20Macros
Samle: https://tw-scripts.github.io/Yazd/#List%20of%20Macros

Then authors only need to fill the required parts.

PR #3882, opened 2019-03-28
update the file name from uniquetitle to unusedtitle

inmysocks:
Also changed the baseName parameter to basename to be more consistent with other macros.

### DIFF
--- a/core/modules/macros/unusedtitle.js
+++ b/core/modules/macros/unusedtitle.js
@@ -24,11 +24,11 @@ exports.params = [
 /*
 Run the macro
 */
-exports.run = function(baseName, options) {
+exports.run = function(basename, options) {
 	if(!baseName) {
-		baseName = $tw.language.getString("DefaultNewTiddlerTitle");
+		basename = $tw.language.getString("DefaultNewTiddlerTitle");
 	}
-	return this.wiki.generateNewTitle(baseName, options);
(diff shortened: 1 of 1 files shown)


00SS:
For the parameter to be more consistent with other core macros, the second word starts with a Capital letter. Please look at the table in the [previous PR](https://github.com/Jermolene/TiddlyWiki5/pull/3880#issuecomment-477318721)

Also I believe the parameter should **not** be called basename or baseName, but should be **baseTitle**.

**basename** already has a different definition in TiddlyWiki.

inmysocks:
If you have specific changes you would like than feel free to open a pull request with them.

kookma:
@inmysocks 
I think basetitle is correct to be consistent with basetitle in action-createtiddler!
Please correct this!

Thank you

00SS:
I am simply pointing out the trends I see in the documentation. Consistency is really useful in the long run.

$basetitle is an **attribute** of the $action-createtiddler widget
Global macros have **parameters** and as mentioned normally follow a slightly different construction - thus my suggestion of **baseTitle**
The suggestion by @kookma is also valid, as it is consistent with the attribute that has exactly the same function.

Someone with more experience might have to decide.

inmysocks:
Closing this in favour of #3885

Issue #3883, opened 2019-03-29
Implement wikitext shortcuts as macros


Jermolene:
It is proposed to investigate refactoring the existing wikitext parsing rules so that they generate a macrocall instead of a specific widget node. Each parse rule would map to a macrocall whose default global definition would just render the same widgets as at present, but it would be possible for users to redefine those macros so that some other sequence of widgets is generated.

For example, the **prettylink** parse rule currently generates the following parse tree node for internal links:

```
return [{
	type: "link",
	attributes: {
		to: {type: "string", value: link}
	},
	children: [{
		type: "text", text: text
	}]
}];
```

Under this proposal, the parse rule would instead generate the following parse tree node:

```
return [{
	type: "macrocall",
	name: "tm-prettylink-internal",
	params: [
		{name: "to", value: link},
		{name: "text", value: text}
	]
}];
```

The default definition of `tm-prettylink-internal` would be:

```
\define tm-prettylink-internal(to,text)
<$link to=<<__to__>>><$text text=<<__text__>>/></$link>
\end
```

Or, to render internal links as buttons, the following definition might be used:

```
\define tm-prettylink-internal(to,text)
<$button>
<$action-navigate $to=<<__to__>>/>
<$text text=<<__text__>>/>
</$button>
\end
```

This proposal reflects one of my goals for this phase of TiddlyWiki 5's development: to try to refactor more of the special, canned behaviours of the core so that they are expressed as wikitext and can be more easily customised. There are examples where this is already possible: for example, the linkcatcher widget allows navigation actions to be customised globally or locally. But there are still many examples of complex behaviour built into the core that is opaque and inaccessible.

I've experimented with this proposal and it is viable, but I'm concerned with performance -- every instance of wikitext shortcut syntax will now involve an extra parsing step. It's possible that this change will require more detailed performance evaluation and optimisation. I'll push a PR shortly that demonstrates the idea.

pmario:
I do like this approach. ... It will increase hackability. ... But it may lead to some weird problems, that are difficult to find. ...

pmario:
It would be nice, if aka-macros + utilities could become part of the core ;)

Jermolene:
> I do like this approach. ... It will increase hackability. ... But it may lead to some weird problems, that are difficult to find. ...

I think in most cases the resulting widget tree will be the same; it's just an extra indirection.

> It would be nice, if aka-macros + utilities could become part of the core ;)

What do you mean?

[...]

PR #3884, opened 2019-03-29
Explore implementing wikitext primitives as macros

Jermolene:
An experiment for #3883

### DIFF
--- a/core/modules/parsers/wikiparser/rules/prettylink.js
+++ b/core/modules/parsers/wikiparser/rules/prettylink.js
@@ -49,13 +49,12 @@ exports.parse = function() {
 		}];
 	} else {
 		return [{
-			type: "link",
-			attributes: {
-				to: {type: "string", value: link}
-			},
-			children: [{
-				type: "text", text: text
-			}]
+			type: "macrocall",
+			name: "tm-prettylink-internal",
+			params: [
+				{name: "to", value: link},
+				{name: "text", value: text}
+			]
 		}];
 	}
 };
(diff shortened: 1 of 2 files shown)


core/modules/parsers/wikiparser/rules/prettylink.js
```diff
@@ -49,13 +49,12 @@ exports.parse = function() {
...
 	} else {
 		return [{
-			type: "link",
-			attributes: {
-				to: {type: "string", value: link}
-			},
-			children: [{
-				type: "text", text: text
-			}]
+			type: "macrocall",
```
pmario: The problem now is, that backlinks don't work anymore :)  ... That's the exact same behavior, that unilink-plugin uses. 

I did use an [ugly hack](https://github.com/wikilabs/plugins/blob/master/wikilabs/uni-link/tiddlers/modules/uni-link.js#L84) so I didn't need to modify the core backlink handling. 

alias-handling has it's own [backlink functions](https://github.com/wikilabs/plugins/blob/master/wikilabs/uni-link/tiddlers/modules/wiki-methodes.js)

macros can be found [here](https://github.com/wikilabs/plugins/tree/master/wikilabs/uni-link/tiddlers): uni-link-macro.tid, aka-macros.tid ...

Jermolene: I think we can address that by updating the link detection code to also look for the corresponding macro.

pmario: yea, I know. ... but for the plugin, I didn't want to mess with the core code.

PR #3891, opened 2019-04-04
Use list widget for reveals in core that obstruct styling

Arlen22:
There are certain places in the core where a reveal widget is used without any class added. In these cases this behavior obstructs CSS inheritance because the reveal widget is sometimes parent to the div you are trying to work with. If it is child to the div, it is just a minor annoyance, but still unnecessary.

At the beginning of this PR I modified the reveal widget to not insert a div tag unless it is actually needed, but after feedback from Jeremy I changed the offending locations to use the list widget (in the case of `$:/tags/SideBarSegment` and `$:/tags/PageControls`) or just added the class to the reveal widget and removed the extra div (in the case of `tc-sidebar-header`). A hidden reveal widget has the hidden attribute set so it is still possible to style it accordingly if desired. This is probably the preferred method. 

I've also added a global macro that streamlines using the list widget to "OPTIONALLY" display items like the page toolbar, edit toolbar, and view toolbar. Since this is a common paradigm it makes sense to use a single filter macro for two reasons. First because it cuts down on code duplication. Second because it can easily be updated if we find a filter string which has better performance.

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -34,30 +34,23 @@ RevealWidget.prototype.render = function(parent,nextSibling) {
 	if(this.revealTag && $tw.config.htmlUnsafeElements.indexOf(this.revealTag) === -1) {
 		tag = this.revealTag;
 	}
-	if(this.revealTag || this.class || this.style 
-		|| this.type === "popup" || this.animate === "yes" 
-		|| this.retain === "yes" || this.nowrap === "no"
-	){
-		var domNode = this.document.createElement(tag);
-		var classes = this["class"].split(" ") || [];
-		classes.push("tc-reveal");
-		domNode.className = classes.join(" ");
-		if(this.style) {
-			domNode.setAttribute("style",this.style);
-		}
-		parent.insertBefore(domNode,nextSibling);
-		this.renderChildren(domNode,null);
-		if(!domNode.isTiddlyWikiFakeDom && this.type === "popup" && this.isOpen) {
-			this.positionPopup(domNode);
-			$tw.utils.addClass(domNode,"tc-popup"); // Make sure that clicks don't dismiss popups within the revealed content
-		}
-		if(!this.isOpen) {
-			domNode.setAttribute("hidden","true");
-		}
-		this.domNodes.push(domNode);
-	} else {
-		this.renderChildren(parent, nextSibling);
+	var domNode = this.document.createElement(tag);
+	var classes = this["class"].split(" ") || [];
+	classes.push("tc-reveal");
+	domNode.className = classes.join(" ");
+	if(this.style) {
+		domNode.setAttribute("style",this.style);
 	}
+	parent.insertBefore(domNode,nextSibling);
+	this.renderChildren(domNode,null);
+	if(!domNode.isTiddlyWikiFakeDom && this.type === "popup" && this.isOpen) {
+		this.positionPopup(domNode);
+		$tw.utils.addClass(domNode,"tc-popup"); // Make sure that clicks don't dismiss popups within the revealed content
+	}
+	if(!this.isOpen) {
+		domNode.setAttribute("hidden","true");
+	}
+	this.domNodes.push(domNode);
 };
 
 RevealWidget.prototype.positionPopup = function(domNode) {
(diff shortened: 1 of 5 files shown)


Jermolene:
Hi @Arlen22 I think of the distinguishing feature of `<$reveal>` vs. `<$list>` is that the former uses a DOM element so that the hide/show can be animated. Are there situations where one doesn't want an element to be generated, but can't use `<$list>` instead?

Jermolene:
I should add that I've found that the disadvantage of this kind of conditional behaviour is that it makes the documentation hard -- the `<$set>` widget is a great case in point.

core/modules/widgets/reveal.js
```diff
@@ -34,23 +34,27 @@ RevealWidget.prototype.render = function(parent,nextSibling) {
...
+		if(!domNode.isTiddlyWikiFakeDom && this.type === "popup" && this.isOpen) {
+			this.positionPopup(domNode);
+			$tw.utils.addClass(domNode,"tc-popup"); // Make sure that clicks don't dismiss popups within the revealed content
+		}
+		if(!this.isOpen) {
+			domNode.setAttribute("hidden","true");
+		}
+		this.domNodes.push(domNode);
+	} else {
+		this.renderChildren(parent, nextSibling);
```
pmario: hmmm, I'm not 100% sure, but if `retain` is `no`, and reveal is closed, nothing should be rendered. right?

Arlen22: The current behavior is that the node is still render with the hidden attribute set and no content.

Arlen22:
> Hi Arlen22 I think of the distinguishing feature of <$reveal> vs. <$list> is that the former uses a DOM element so that the hide/show can be animated. Are there situations where one doesn't want an element to be generated, but can't use <$list> instead?

I tried using the list widget for this but it seemed like it made my wiki super slow and typing in CodeMirror became very laggy. The reveal widget is much simpler and quicker in its calculations. This PR checks whether the animated attribute is set. A third-party animation would always set either a custom tag or a custom class, which would also cause the wrapping element to be rendered. 

The other option would be to use a different attribute like `nowrap` to indicate that no wrapping element should be used. This would conflict with all the situations listed above, so one of them would have to take precedence. I think `nowrap` should only apply if all of the other conditions are false. We would hopefully be able to overhaul the core UI because there are very many places where `nowrap` is preferred because the reveal widget's class attribute is not taken advantage of, and others (such as page controls) where you simply don't want clutter.

sukima:
What about making a custom widget the extands the reveal widget. This way you would have controll over the existence of a wapping dom element or not. This way the core remains untouched and functional.

Arlen22:
I actually want to change the core in this situation, because the core is hard to style, but a `nowrap` attribute is probably a better idea than just conditional behavior.

Jermolene:
Hi @Arlen22 

> I tried using the list widget for this but it seemed like it made my wiki super slow and typing in CodeMirror became very laggy. The reveal widget is much simpler and quicker in its calculations.

I'm not keen to change the reveal widget to have two such fundamentally different behaviours. The present situation is that the list widget is recommended for situations where no DOM element is desired. If we can show that the reveal widget is significantly faster than the list widget then I'd consider adding a `<$if>` widget that had the proposed behaviour.

If you're running into problems styling `<$reveal>` elements in the core I'd be happy to consider adding new classes etc. to make them addressable.

Arlen22:
I made another commit to use the list widget, which actually simplified things slightly. In the sidebar, where the reveal widget is needed for animation I merged it with the parent div so it now has the `tc-sidebar-header` class on the reveal widget itself. 

The PR https://github.com/Jermolene/TiddlyWiki5/pull/3854 to add the autoparagraph variable is the other side of this coin. Both of them together will make creating new themes much easier.

Arlen22:
@Jermolene Shall I make a new PR with only one commit or are you able to merge everything into one commit and take out the change to reveal.js?

Arlen22:
@Jermolene, what's the chance that we can add this one to 5.1.20? The other one (https://github.com/Jermolene/TiddlyWiki5/pull/3854) is a lot bigger, and doesn't necessarily need to be added in this release in my opinion. It does seem like something that would be added to the styling improvements that are coming.

Jermolene:
Hi @Arlen22 thanks for the updates. The OP mentions that the reveal widgets obstruct styling, I just wanted to check what the original problem was, and how this new PR addresses it (in fact, perhaps you could update the OP to describe the current changes, so that future historians will be able to piece things together).

Arlen22:
@Jermolene thanks. I've updated the OP to give an overview of the PR. The cases that I've addressed seemed to be the main offenders. 

I'm sure other places like the page toolbar, view toolbar, and edit toolbar could benefit from the macro as well but I'll leave that up to others so as not to clutter this PR with too much stuff.

Arlen22:
Closing this in favor of https://github.com/Jermolene/TiddlyWiki5/pull/4160

Issue #3888, opened 2019-04-03
Advanced authentication


newhinton:
Currently only basic authentication and setting a header are the only way to authenticate a user.
I think we could build a ui for that, so that a user can use password manager or similar. I have experimented with the basic.js to serve a html webpage and then redirect the user to a site using the user:pass@domain.tld scheme, but that promts a browser popup (which is exactly what i do not want)

So, my idea would be to create the ability to set a session variable or something similar server side, so that we can call a route and pass the password and username to some sort of backend-plugin that then authenticates a user (that would also allow for many different methods of authentication since we then can create plugins that add eg. ldap). 

Any ideas or suggestions on this topic?

AnthonyMuscio:
A Key part of this is to allow a user to take ownership of the wiki, check it out for edit, check it back so someone else can edit. this requires consideration of single file wiki saves.

I have done some experimental work on this but do not have the js and plugin dev skills, local and session storage is already available and a great place to maintain session authentication information. 

Regards
Tony

newhinton:
@AnthonyMuscio I have since my original post made further progress with migrating from http-header authentication to cookiebased authentication which looks promising. The main problem i am facing in my prototype is the server-side, i have not yet figured out how to store a session-information.

Do you know how to do that and can you provide me with hints on where i can find the session store?

Regarding your first comment, if i undestand you correctly that should not be an issue. The basic authentication is more or less only for the nodejs version which runs as a server, not for single file wiki's

AnthonyMuscio:
Hinton

I am no expert on this. In what server environment are you doing this or want to? Are you saying you only want a similtaniouse edit on a folder based wiki?

I have been exploring php user authentication and single file wikis. However this has led me down the path of the local storage plugin in the new release and access to local rather than only session based storage. 

I have made a number discoveries of late such as the passwordwidget already saves in the browser session. I believe with a checkin/out process and the ability to store changes until the users access permits and then commit changes I have a solution.

The other key is using a php based user login and auth process and reloading the wiki with auth detail fed into the wiki. Within php you can set and interogate server session info the logged in user cant see.

Any way. This is early in my journey

Tony

newhinton:
@AnthonyMuscio 
[...]

PR #3892, opened 2019-04-04
Implemented UI and Cookiebased authentication for nodejs

newhinton:
This is a fully working prototype for an ui-based approch to a login to tiddlywiki. It uses cookies and a temporary authentication file to authorize users to tiddlywiki. It currently replaces the basic authentication, so i need a little bit advice on how to move it to a standalone authentication method.

Todo:

- Translation of the Loginpage
- Check if authfile cleanup is in the proper place (currently boot.js)
- make it standalone and not replace basic auth
- hash client side cookie and check if client has tampered with its cookies (and invalidate them if they have been changed)
- the template is loaded from the current working directory, in dev instances that is fine but npm i -g tiddlywiki will crash the server


newhinton:
see #3888

newhinton:
2019-04-04 23_09_34-Tiddlywiki Authentication – Opera

This is the current userinterface

Jermolene:
Hi @newhinton thanks for this, and apologies for the delay in responding. I'll give some detailed code comments shortly, but overall I think there may be quite a lot of work to get this into a form where it's universal enough for the core. I'm wondering if instead we can refactor things into a plugin -- on the face of it, it's only the deletion of `auth.temp` in boot.js that couldn't be done in a plugin, and I think that could be handled in a startup module.

boot/boot.js
```diff
@@ -28,6 +28,11 @@ if($tw.node) {
 	fs = require("fs");
 	path = require("path");
 	vm = require("vm");
+	try {
```
Jermolene: To avoid having to modify boot.js I think this could be done in a startup module.

core/modules/server/authenticators/basic.js
```diff
@@ -60,35 +61,290 @@ Returns true if the request is authenticated and assigns the "authenticatedUsern
 Returns false if the request couldn't be authenticated having sent an appropriate response to the browser
 */
 BasicAuthenticator.prototype.authenticateRequest = function(request,response,state) {
```
Jermolene: As you noted, you should add a new authenticators/cookie.js file. Things should just work if you implement the methods correctly, but we may need to tweak the logic in server.js

package.json
```diff
@@ -23,7 +23,9 @@
     "tiddlywiki5",
     "wiki"
   ],
-  "dependencies": {},
+  "dependencies": {
+    "uuid": "^3.3.2"
```
Jermolene: To ensure that TW is fully self contained we don't use NPM for installing dependencies. Instead we copy the code into the core (e.g. https://github.com/Jermolene/TiddlyWiki5/tree/master/core/modules/utils/diff-match-patch)

plugins/tiddlywiki/jasmine/files/reporter.js
```diff
@@ -176,7 +176,7 @@
       var results = runner.results();
       var specs = runner.specs();
       var msg = '';
-      msg += specs.length + ' test' + ((specs.length === 1) ? '' : 's') + ', ';
+      msg += specs.length + ' template.html' + ((specs.length === 1) ? '' : 's') + ', ';
```
Jermolene: What's the idea behind this change?

core/modules/server/authenticators/basic.js
```diff
@@ -60,35 +61,290 @@ Returns true if the request is authenticated and assigns the "authenticatedUsern
...
+			response.writeHead(406,"Wrong Password");
+			response.end();
+			return false;
+		}
+
+		id=uuid.v4();
+		timestamp = date.getTime();
+
+		var appDir = path.resolve('core/modules/server/authenticators/');
+		//var resp=fs.readFileSync(appDir+'/template.html', 'utf8');
```
Jermolene: You can read the template directly from the tiddler using something like `$tw.wiki.getTiddlerText(title)` where title is the title of the tiddler containing the template.

plugins/tiddlywiki/jasmine/files/reporter.js
newhinton: I guess this was a mistakenly commited file after refactoring my template :D shouldn't be here

PR #3894, opened 2019-04-05
Update wikitext-macros.tid

AnthonyMuscio:
Added new macro wikitext-example-compact to support the publishing of Widget variants in a more compact form.

Initially to support changes to the ViewWidget Documentation, this compact form may be used in other Widget documentation.

Variants demonstrate alternative forms of a specific widget or macro with the ability for users to copy to clipboard and use in their own wiki. In some widgets that have multi-functions there is no single form of the widget, and demonstrating some variations highlights this.

PR #3895, opened 2019-04-05
Update ViewWidget.tid to include common variants

AnthonyMuscio:
The provision of common variates to the view widget provides new an experience user examples of the different variants of the ViewWidget. The copy to clipboad method allows each variant to be quickly accessed and pasted into their wiki.

To facilitate this a new macro wikitext-example-compact has being added to 
$:/editions/tw5.com/wikitext-macros

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ViewWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ViewWidget.tid
@@ -39,6 +39,19 @@ The following formats can be specified in the `format` attribute:
 |''stripcomments'' |The field is interpreted as JavaScript source code and any lines beginning `\\#` are stripped |
 |''jsencoded'' |The field is displayed as a JavaScript encoded string |
 
+;The following examples are common variants
+
+!!!!View the value in a field of a named tiddler
+:<<wikitext-example-compact '<$view tiddler="ViewWidget" field="caption"/>'>>
+!!!!View the value in a date field of the currentTiddler as a relative date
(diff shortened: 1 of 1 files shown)


AnthonyMuscio:
The attached image demonstrates the insertion of these changes in the view Widget Tiddler

VariantsInViewWidget

Regards
Tony

pmario:
@Jermolene -- This is doc-only. Please check

Jermolene:
Thank you @AnthonyMuscio

saqimtiaz:
@Jermolene all of the examples that this PR adds seem broken on tiddlywiki.com, see https://tiddlywiki.com/#ViewWidget and compare it to the screenshot above.

It seems that the macro used for the examples does not exist.

AnthonyMuscio:
This is a very old request, is not 

> @Jermolene -- This is doc-only. Please check

As @pmario indicated. I was to introduce the new macro, that @saqimtiaz indicated was missing. 

The intention was to also display the raw code examples, such that people could copy and use the different forms.

Perhaps I could submit an new change but would appreciate advice on how to display the code in the documentation standards.

pmario:
@Jermolene ... Sorry. I did miss that this PR depends on **Update wikitext-macros.tid #3894**, which is not merged yet.

saqimtiaz:
@Jermolene this PR needs to be reverted as the changes merged are incomplete and non-functional.

Jermolene:
Thanks – I've reverted this for now.

PR #3896, opened 2019-04-05
Update Filter Operators.tid

AnthonyMuscio:
To add an additional paragraph at the top to direct the user to common uses of Filters, in ListWidget, SetWidget and Filtered Transclusions.

Please add to this if there is common uses elsewhere for filter. 

By providing this guidance to users at the early stage of learning they will know the key places in which Widgets use filters thus have the opportunity to experiment and learn more about filters, Otherwise they remain an abstract concept.

PS I am not sure why my recent proposed changes delete the last <enter> in the file and if this is a problem, nor do I know how to reverse this.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/Filter Operators.tid
+++ b/editions/tw5.com/tiddlers/filters/Filter Operators.tid
@@ -21,6 +21,8 @@ A <<.def "filter operator">> is a predefined keyword attached to an individual [
 
 ''Important:'' Each first [[step|Filter Step]] of a [[filter run|Filter Run]] not given any input titles receives the output of <$link to="all Operator">[all[tiddlers]]</$link> as its input.
 
+[[Filters]] and Filter Operators are most commonly used in the ListWidget, SetWidget and [[Filtered Transclusions|Transclusion in WikiText]] `{{{ filter }}}`
+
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @AnthonyMuscio I wonder if this might not be better in the tiddler "Filters", which is the top level introduction to the topic, with "Filter Operators" subordinate to it.

editions/tw5.com/tiddlers/filters/Filter Operators.tid
```diff
@@ -21,6 +21,8 @@ A <<.def "filter operator">> is a predefined keyword attached to an individual [
 
 ''Important:'' Each first [[step|Filter Step]] of a [[filter run|Filter Run]] not given any input titles receives the output of <$link to="all Operator">[all[tiddlers]]</$link> as its input.
 
+[[Filters]] and Filter Operators are most commonly used in the ListWidget, SetWidget and [[Filtered Transclusions|Transclusion in WikiText]] `{{{ filter }}}`
```
pmario, line 24: @AnthonyMuscio ... I think this PR is obsolete now. It probably wasn't as you created it. 

The `[[Filters]]` link is a duplicate of the `[[filter|Filters]]` link in line 20 ... 
`[[Filtered Transclusions|Transclusion in WikiText]]` .. Is linked at the [Filters](https://tiddlywiki.com/#Filters) tiddler. 
Lists are also mentioned in the Filters tiddler. 
------
@AnthonyMuscio 
IMO this PR can be closed, because it only contains duplicated info now.

AnthonyMuscio:
Taking advice given to close and it was neglected too long

PR #3898, opened 2019-04-06
Add support of language aliases in highlight.js plugin

talha131:
listLanguages() returns a list of supported languages. It does not
return language aliases.

highlight.js has extensive support of language aliases which can be 
viewed here

https://highlightjs.readthedocs.io/en/latest/css-classes-reference.html#language-names-and-aliases

But because of using listLanguages(), TW does not take advantage of 
alias.

getLanguage() method on the other hand supports aliases.

https://highlightjs.readthedocs.io/en/latest/api.html#getlanguage-name

To summarize, now user can use javascript, js or jsx for the code block. getLanguage() will return an object which means highlight.js supports
this language.

### DIFF
--- a/plugins/tiddlywiki/highlight/highlightblock.js
+++ b/plugins/tiddlywiki/highlight/highlightblock.js
@@ -27,7 +27,7 @@ CodeBlockWidget.prototype.postRender = function() {
 	if(tiddler) {
 		language = tiddler.fields.text || "";
 	}
-	if(language && hljs.listLanguages().indexOf(language) !== -1) {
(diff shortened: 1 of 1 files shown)


talha131:
Details are here

https://groups.google.com/forum/#!topic/tiddlywiki/LUFCr2-KcP0

talha131:
@Jermolene Can you please merge this PR?

I have been using the fork in my local setup and I have not noticed any bug so far.

Jermolene:
Apologies @talha131, and thank you, this looks good.

Issue #3840, opened 2019-03-08
DOCS: Missing: QualifyWidget


00SS:
I believe the documentation for the <**$qualify>** widget is missing.

Perhaps something as simple as this to start with, though it would be better to give some real use case examples:

The qualify widget returns a unique string that encodes its position within the widget tree, as identified by the stack of transcluded tiddlers that lead to that position, and assigns the result to a specified variable. The new value of the variable is available to the content within the qualify widget.

Attribute |Description 
------------ | -------------
name |The name of the variable to assign 
title |The prefix for the returned string, normally a tiddler title. Defaults to `undefined-` 

Example:
```
<$qualify name="some-name" title="Some Title"> <<some-name>>

</$qualify>
```

See also the **qualify Macro**

Jermolene:
Hi @00SS that's looks great, thank you! I'd be very happy if you can prepare a PR.

By the way, I'm aware that you have a number of outstanding PRs, and I'm sorry not to have dealt with them quicker. Are they all ready for review? If the answer is mixed perhaps it would be worth making a new issue for keeping track of the status of your contributions (call it something like "00SS Pull Requests", and make a list of the issues sorted into "ready for review", "in progress" etc.).

00SS:
@Jermolene I will make a PR for the qualify widget soon.

Except for the most recent PR - [**Internal Links**](https://github.com/Jermolene/TiddlyWiki5/pull/3836) which is still getting active feedback, all of [**my Pull Requests**](https://github.com/Jermolene/TiddlyWiki5/pulls/00SS) are ready for review.

I like the new ability to make a "Draft PR" first, and changing its status when it has gone through  discussion and is ready for review. I will use that method in the future so that it's easy to spot the completed ones.

Maintaining a separate Issue for all "00SS Pull Requests" would probably end up confusing us both.

AnthonyMuscio:
**I would propose more user friendly content, and move the above proposed introductory paragraph below**

The qualify macro generates a unique title with numbers based on the current tiddler in which it is operating. The title parameter permits the provision of a prefix to that unique title.

For example;
`<<qualify $:/temp/>> `
Will result in the output of "$:/temp/-1105666819" where the number if unique to the current Tiddler.
Using `<<qualify $:/temp/>> ` again, later in the same tiddler will produce the same result.

`<<qualify $:/state/display>> `
Will result in the output of "$:/state/display-1105666819"

[...]

PR #3900, opened 2019-04-09
DOCS: Create: qualifyWidget.tid

00SS:
Documentation for this tiddler was missing as per #3840


00SS:
Preview can be seen here: https://00ss.github.io/documentation/tiddlywiki.html#qualifyWidget

PR #3905, opened 2019-04-11
Check vars earlier in GitHub saver

bimlas:
If the path was not specified, RSOD error occurred when we wanted to download the wiki:

```
Uncaught TypeError: Cannot read property 'substring' of undefined
```


bimlas:
@Jermolene Please merge as soon as possible, because prerelease cannot be downloaded right now without this fix.

Jermolene:
Thanks @bimlas

PR #3908, opened 2019-04-12
DOCS: Small update: Headings in WikiText.tid

00SS:
Changed the example to use the **wikitext-example** macro and pointed out that wikitext only works till `<h6>`


Jermolene:
Thanks @00SS

PR #3911, opened 2019-04-14
Update chinese language files

BramChen:
* Add chinese help texts for the new savewikifolder command


Jermolene:
Thanks @BramChen very quick!

BramChen:
Many thanks @Jermolene

Issue #3912, opened 2019-04-14
Prefix line


RunningUtes:
A tm-edit-text-operation is supposed to preform an operation on selected text but it doesn't seem to do as the documentation lists.

https://tiddlywiki.com/#WidgetMessage%3A%20tm-edit-text-operation

If selected text is bolded, the '' is added as a prefix and as a suffix. If the same text is already bolded and the operation is run again, the '' is removed. See the rules in $:/core/ui/EditorToolbar/list-bullet.

This doesn't seem to happen for the prefix-lines operation.

For example, select several lines of text and press the unordered list button. A * will be added as a prefix according to the rules in $:/core/ui/EditorToolbar/list-bullet. If the same text is selected and the unordered list button is pressed again, the * is not removed. In fact, nothing happens. The documentation states that the prefix should be removed from the selected lines.

Is this a bug in the code of the current TW?

The problem seems to be in $:/core/modules/editor/operations/text/prefix-lines.js

AnthonyMuscio:
Perhaps with prefix items such as * # and : we should allow add level and subtract level

I know I would like to select multiple rows and simply add another level deep, then a tools to reverse this one level at a time. Perhaps once fixed the existing button can simply reset it to one level or none.

+
*  +
** +
*** +
** -
* -
-


Regards
Tony

sukima:
# Confirmed

This is a bug. [`prefix-lines.js` L:36](https://github.com/Jermolene/TiddlyWiki5/blob/df416814b13fadd541ca30dffe3f77d24e63f8b0/core/modules/editor/operations/text/prefix-lines.js#L36) has a type error.

```js
		if(count !== event.paramObject.count) {
```

in this case `count` is a number while `event.paramObject.count` is a string and the `!==` comparison is type sensitive.

pmario:
@sukima ... Nice catch!

PR #3919, opened 2019-04-17
Cast count to a number for comparison

sukima:
Prior to this change, the `count` provided was a string. However, our use for it is as a number. Though there are some cases where type coercion is desired in this case it conflates the readbility of the code.

This change casts the `count` to a number before using it in numeric comparisons.

Fixes issue #3912

### DIFF
--- a/core/modules/editor/operations/text/prefix-lines.js
+++ b/core/modules/editor/operations/text/prefix-lines.js
@@ -17,8 +17,10 @@ exports["prefix-lines"] = function(event,operation) {
 	operation.cutStart = $tw.utils.findPrecedingLineBreak(operation.text,operation.selStart);
 	// Cut to just past the following line break, or to the end of the text
 	operation.cutEnd = $tw.utils.findFollowingLineBreak(operation.text,operation.selEnd);
+	// paramsObject offers count as a string not number
+	var prefixCount = parseInt(event.paramObject.count,10);
 	// Compose the required prefix
-	var prefix = $tw.utils.repeat(event.paramObject.character,event.paramObject.count);
+	var prefix = $tw.utils.repeat(event.paramObject.character,prefixCount);
 	// Process each line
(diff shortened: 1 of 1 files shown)


ght:
This can be closed. The linked issue #3912 has already been fixed in a054d100e73d in pretty much the same way, so in a way this was already merged :)

core/modules/editor/operations/text/prefix-lines.js
```diff
@@ -17,8 +17,10 @@ exports["prefix-lines"] = function(event,operation) {
 	operation.cutStart = $tw.utils.findPrecedingLineBreak(operation.text,operation.selStart);
 	// Cut to just past the following line break, or to the end of the text
 	operation.cutEnd = $tw.utils.findFollowingLineBreak(operation.text,operation.selEnd);
+	// paramsObject offers count as a string not number
```
pmario, line 20: The comment text is confusing. ... It doesn't fit to the code. right?

ght: It refers to the line below, and I didn't find it confusing. The merged fix has no comment at all, and in addition to what this PR does, [casts `paramsObject.count` to a string before parsing it](https://github.com/Jermolene/TiddlyWiki5/commit/a054d100e73db95071299e92c4321c2aa8e42382#diff-130da145d09ea1c35cea010cf80d0e79860765998846dd294da956ab093b04d5R16) which could be confusing to new contributors:
```diff
 exports["prefix-lines"] = function(event,operation) {
+	var targetCount = parseInt(event.paramObject.count + "",10);
 	// Cut just past the preceding line break, or the start of the text
```
Either way, it's a separate issue (and a very minor one).

This PR can be closed. If the current code needs a comment, it should be added in a new PR IMHO.

Jermolene:
Thanks @ght

PR #3921, opened 2019-04-17
Fix "Filtered Attribute Values" example

AnthonyMuscio:
syntax error, an additional "]" has being removed, and the P changed to p


00SS:
@AnthonyMuscio 
Nice catch! Good fix!
This should be merged right away.

pmario:
@Jermolene ... This PR can be closed, because the docs has been fixed in the meantime. Probably with a different PR.

AnthonyMuscio:
closed

PR #3922, opened 2019-04-17
Update PageTemplate.tid - change 6x$set to 1x$vars

00SS:
Would there be any reason not to use 1 `$vars` widget instead of 6 `$set` widgets?


pmario:
Hi @00SS is there something wrong with this PR?

PR #3931, opened 2019-04-22
Add GitLab saver, apply common lingo to Git savers

bimlas:
The GitHub and GitLab APIs are different in many ways, so it is not enough to change only the URL of the API, thus the differences are stored in separate modules so you can choose between providers.

I could not implement the BitBucket API: https://developer.atlassian.com/bitbucket/api/2/reference/resource/repositories/%7Busername%7D/%7Brepo_slug%7D/src#post


Jermolene:
Many thanks @bimlas, this a very complete piece of work. Refactoring it as a generic Git saver is quite a significant change and we need to understand all the alternatives before we decide how to proceed.

My first concern is that from a user perspective the generalised UI prompts and documentation are a bit clumsier and harder to follow for beginners. I'm not sure of the best solution, perhaps we just need to add more explanatory text explaining what a "git service" is etc.

My second concern is about the way that the commonality is structured between the two savers. The current approach abstracts a handful of operations to the modules that implement GitHub or GitLab specific behaviour (composing URIs, headers, commit data). To judge whether this is a sufficient level of abstraction we need to consider the scenarios in which we might want to extend the current implementation. For example, if adding a new service provider, we'd have to consider whether the abstractions are adequate the express the behaviour of the new service. It seems that the current abstractions might be quite specific to "Github-like" APIs

The alternative approach would be to have two separate saver modules, sharing as much code as possible between them. They could use the same lingo strings, or we could parameterise the service name so that e.g. `Saving/GitService/Caption` becomes `<<service-name>> Saver`.

There's a couple of minor code comments I'll make in a moment, but they're really subservient to these bigger concerns about structure.

core/modules/savers/git-service.js
```diff
@@ -17,26 +17,23 @@ var base64utf8 = require("$:/core/modules/utils/base64-utf8/base64-utf8.module.j
...
-		filename = this.wiki.getTiddlerText("$:/GitHub/Filename"),
-		branch = this.wiki.getTiddlerText("$:/GitHub/Branch") || "master",
-		endpoint = this.wiki.getTiddlerText("$:/GitHub/ServerURL") || "https://api.github.com",
-		headers = {
-			"Accept": "application/vnd.github.v3+json",
-			"Content-Type": "application/json;charset=UTF-8",
-			"Authorization": "Basic " + window.btoa(username + ":" + password)
-		};
+		titleOfSelectedService = this.wiki.getTiddlerText("$:/GitService/ServiceType"),
+		service = require(titleOfSelectedService),
```
Jermolene: I think it's dangerous to use a user-supplied tiddler value as a the name of a module to be required. It might set up a security vulnerability if the user chose a malicious value.

It would be better for service provider modules to be a new module type, handling them something like the way that the `is` filter operator invokes the submodules that implement each specific type of check.

core/modules/savers/git-service.js
```diff
@@ -17,26 +17,23 @@ var base64utf8 = require("$:/core/modules/utils/base64-utf8/base64-utf8.module.j
...
+		username = this.wiki.getTiddlerText("$:/GitService/Username"),
+		password = $tw.utils.getPassword("gitservice"),
+		repo = this.wiki.getTiddlerText("$:/GitService/Repo"),
+		path = this.wiki.getTiddlerText("$:/GitService/Path"),
+		filename = this.wiki.getTiddlerText("$:/GitService/Filename"),
+		branch = this.wiki.getTiddlerText("$:/GitService/Branch") || "master",
+		apiUrl = this.wiki.getTiddlerText("$:/GitService/ServerURL") || this.wiki.getTiddler(titleOfSelectedService).fields['default-api-url'];
```
Jermolene: Here you're using a field of a JS module to hold the default API url. It's clearer to keep everything in JavaScript, and have the JS module export the value.

core/modules/savers/git-service.js
```diff
@@ -61,24 +58,27 @@ GitHubSaver.prototype.save = function(text,method,callback) {
...
+				requestType = service.requestTypeIfFileNotExists;
 			}
-			var data = {
-					message: "Saved by TiddlyWiki",
-					content: base64utf8.base64.encode.call(base64utf8,text),
-					branch: branch,
-					sha: sha
-				};
+			var data = service.formatDataForCreatingCommit(
+				"Saved by TiddlyWiki",
```
Jermolene: While we're working on this, we might want to make this string be translatable.

bimlas:
Sorry, but I don't have time to do it now, but I will answer on the weekend.

core/modules/savers/git-service.js
linonetwo: @Jermolene Why so many wiki.getTiddlerText? Shouldn't those fields merged into a single tiddler, and use different fields of a tiddler?

And what do you mean by "keep everything in JavaScript"?

linonetwo:
I'm trying to make a solid saver, and learning from here. But many things here is undocumented in https://tiddlywiki.com/dev/, so I asked questions in comments in this PR, sorry for the disturbing!

core/modules/savers/git-service.js
```diff
@@ -17,26 +17,23 @@ var base64utf8 = require("$:/core/modules/utils/base64-utf8/base64-utf8.module.j
 /*
 Select the appropriate saver module and set it up
 */
-var GitHubSaver = function(wiki) {
+var GitServiceSaver = function(wiki) {
```
linonetwo: Why not use class syntax? All major browser supports it now.

core/modules/savers/git-services/github.js
```diff
@@ -0,0 +1,45 @@
...
+exports.formatDataForCreatingCommit = function(message,encodedContent,branch,sha) {
+	return {
+		message: message,
+		content: encodedContent,
+		branch: branch,
+		sha: sha
+	}
+}
+
+})();
```
linonetwo: Why wrap class with a (function (){})() ?

If is's for variable locality, why not just:

```js
{
  const aaa = 3;
}
```
to create a scope?

core/modules/savers/git-service.js
linonetwo: Oh, I know!

The text you added to `core/language/en-GB/ControlPanel.multids` becomes several tiddlers on boot! And they are automatically registered as control panel fields, so you are actually retrieving control panel fields!

core/language/en-GB/ControlPanel.multids
```diff
@@ -89,15 +89,16 @@ Saving/DownloadSaver/Hint: These settings apply to the HTML5-compatible download
...
-Saving/GitHub/Branch: Target branch for saving (defaults to `master`)
-Saving/GitHub/Caption: ~GitHub Saver
-Saving/GitHub/Description: These settings are only used when saving to ~GitHub
-Saving/GitHub/ServerURL: Server URL (defaults to `https://api.github.com`)
-Saving/GitHub/Filename: Filename of target file (e.g. `index.html`)
-Saving/GitHub/Password: Password, OAUTH token, or personal access token
-Saving/GitHub/Path: Path to target file (e.g. `/wiki/`)
-Saving/GitHub/Repo: Target repository (e.g. `Jermolene/TiddlyWiki5`)
-Saving/GitHub/UserName: Username
+Saving/GitService/Branch: Target branch for saving (defaults to `master`)
```
linonetwo: What does `multids` mean? Does it mean a quick way to create multiple tiddlers?

And how to push text into this file by a plugin, if I want to register some fields into control panel?
I want to create a plugin that can read some custom config from control panel.

core/modules/savers/git-service.js
```diff
@@ -17,26 +17,23 @@ var base64utf8 = require("$:/core/modules/utils/base64-utf8/base64-utf8.module.j
...
-		endpoint = this.wiki.getTiddlerText("$:/GitHub/ServerURL") || "https://api.github.com",
-		headers = {
-			"Accept": "application/vnd.github.v3+json",
-			"Content-Type": "application/json;charset=UTF-8",
-			"Authorization": "Basic " + window.btoa(username + ":" + password)
-		};
+		titleOfSelectedService = this.wiki.getTiddlerText("$:/GitService/ServiceType"),
+		service = require(titleOfSelectedService),
+		username = this.wiki.getTiddlerText("$:/GitService/Username"),
+		password = $tw.utils.getPassword("gitservice"),
```
linonetwo: Why not use `<$password name="gitservice"/>`?

core/language/en-GB/ControlPanel.multids
pmario: language multids are for translation text only. you cant store values there

core/modules/savers/git-services/github.js
pmario: TW implements its own module loader that is common.js compatible and works in the browser and from the command line with node.js, without changing a line of code. No "browserify" or other code transformations needed.

core/modules/savers/git-service.js
pmario: Because of coding style consistency.

core/modules/savers/git-service.js
pmario: >Why so many wiki.getTiddlerText? Shouldn't those fields merged into a single tiddler, and use different fields of a tiddler?

Experience shows, that separating config values has advantages. eg: config values are also be used by the UI templates. If you change 1 tiddler, the refresh mechanism only updates those parts that use this tiddler. ...

Refresh is triggered by tiddler change and not per field. So if you change 1 of 100 fields, you'd need to update them all

bimlas:
@Jermolene 

> It seems that the current abstractions might be quite specific to "Github-like" APIs
> 
> The alternative approach would be to have two separate saver modules, sharing as much code as possible between them. They could use the same lingo strings, or we could parameterise the service name so that e.g. Saving/GitService/Caption becomes <<service-name>> Saver.

I also feel that the current solution is not the best because it only supports GitHub-like services. I did not know that parameters / variables can be used in translations, this makes easier the implementation of common parts. :+1: It will certainly be best to create separate savers, so I will change the code in this direction.

core/modules/savers/git-service.js
bimlas: If we break the code into separate savers, I think this problem will be resolved, BTW I've got your idea.

core/modules/savers/git-service.js
bimlas: @Jermolene I wanted to minimize the code duplication, so I wanted to get the value from one place, either I'm using it in Javascript or in a tiddler. What is the right solution for this?

core/modules/savers/git-service.js
bimlas: Ok, but then we need to change it in the other savers as well.

core/modules/savers/git-service.js
Jermolene: Hi @linonetwo we intentionally stick to ES5 to give maximum compatibility; there are a lot of old JS environments out there. There are lots of reasons we don't use transpilation chiefly because browsers don't support source maps for single files like TW.

Jermolene:
Hi @bimlas thanks and good luck. I'm keen to be able to offer GitLab support in the core.

core/modules/savers/git-service.js
bimlas: @Jermolene 

> have the JS module export the value

I don't know how to fix it. For a widget, `setVariable()` works, but this is a saver module where the method is unavailable (as I know). How do I create a "global" variable? Or is it so minimal duplication not to deal with it and declare the URL in two places (in the config interface and in Javascript)?

bimlas:
Sadly, it doesn't seem to make much sense to search for common parts in the code, because there might be a Git service that is very different from the GitHub API (such as BitBucket). Therefore, I feel that Git savers are not able to share other than the translations.

While GitLab and GitHub are very similar to each other, maybe this could be used and GitLab would extend the GitHub saver, but I'm not sure it would be a good idea: if GitLab API will be more and more different from GitHub , the code will be more unmanageable.

I think the best thing is to undo the changes (expect for lingo) and copy the original GitHub saver, modify it and it will be the GitLab saver. This is not the first case that two savers are very similar to each other, but their code is separate (tiddlye.js and fsosaver.js).

@Jermolene, what is your opinion?

bimlas:
I think, I'm done, @Jermolene please review!

core/modules/savers/git-service.js
bimlas: Should I convert to lingo in each saver in this PR or should I open another request?

core/ui/ControlPanel/Saving/GitHub.tid
```diff
@@ -0,0 +1,16 @@
+title: $:/core/ui/ControlPanel/Saving/GitHub
+tags: $:/tags/ControlPanel/Saving
+caption: {{$:/language/ControlPanel/Saving/GitService/GitHub/Caption}}
```
bimlas: It doesn't seems to possible to use macros in caption (`<<service-name>>`), but if possible, there is no need for separate captions.

bimlas:
It's not exactly about savers, but the password might be encoded in some form.

image

core/modules/savers/git-service.js
bimlas: Sorry, I misunderstood what to translate, corrected it.

dubiouscript:
> it doesn't seem to make much sense to search for common parts in the code, because there might be a Git service that is very different from the GitHub API (such as BitBucket).

>if GitLab API will be more and more different from GitHub , the code will be more unmanageable.

as it happens i have recently made some changes to a script (snip.sh) for accessing gitlab snippet api 
https://gist.github.com/dubiouscript/5553dce89497ffd9805dd0de16503e8d 
 
and also found a others for bitbucket snippets and github gist
(bucketsnip.sh)
https://gist.github.com/mgoellnitz/c7c5effb4045d5484bf649eca674a374
(gist.sh)
https://gist.github.com/mgoellnitz/103757dea40f51c1c1376580a89a7773

splitting the api makes sense imho

---
... ftr
git saving in browser reminded me of creationix/js-git 
& i found demo 
https://es-git-examples.mariusgundersen.net/push.html

dubiouscript:
> ... so splitting the api related things seems to makes sense imho (eyes) ,..

more related to this tw-gitlab and other tw-save remote api methods 
the above scripts have examples of using the api to save "snippets" / "gists" 
which could be used as an alternative to full git repo / 'project'

*gitlab currently do not use git to store snippets 
https://gitlab.com/gitlab-org/gitlab-ce/issues/13426

bimlas:
It seems that you are online, thus I'm pinging at you, @Jermolene :)

bimlas:
@Jermolene, can you please review?

core/modules/savers/gitlab.js
```diff
@@ -0,0 +1,122 @@
...
+type: application/javascript
+module-type: saver
+
+Saves wiki by pushing a commit to the GitLab REST API
+
+\*/
+(function(){
+
+	/*jslint node: true, browser: true */
+	/*global $tw: true */
```
pmario: the `(function(){` ... doesn't count for indentation. So you'll need to go 1 level in.

core/modules/savers/github.js
```diff
@@ -8,113 +8,113 @@ Saves wiki by pushing a commit to the GitHub v3 REST API
 \*/
 (function(){
 
-/*jslint node: true, browser: true */
-/*global $tw: false */
-"use strict";
+	/*jslint node: true, browser: true */
+	/*global $tw: true */
+	"use strict";
```
pmario: should stay 1 indent left. the initial function() doesn't count

Jermolene:
Hi @bimlas apologies for the delay. I think this looks good, apart from the indentation issues raised by @pmario which make it hard to track the changes.

bimlas:
@pmario, @Jermolene: I've fixed the indentation.

bimlas:
@Jermolene , if you want GitLab saver to be included in v5.1.20, then please merge.

bimlas:
@Jermolene, merge please if you like it.

Jermolene:
Thank you @bimlas, and apologies for the delay in merging.

Issue #3881, opened 2019-03-27
BUG: toc macro recursion error


00SS:
On tiddlywiki.com - Open tiddler: **Contents** and add the tag:
tags: **First**

Make a: **New Tiddler**
text:
```
<div class="tc-table-of-contents">
<<toc "Contents">>
</div>
```
Save: **New Tiddler**

Javascript Error: too much recursion

I had opened a pull request some months ago Fix recursion error in toc() macro in: toc.tid #3668 and recently closed it as very strangely I could not reproduce that recursion error of having given a non-existent tag.

I found this present error because I was trying to figure out how the **toc macros** call themselves recursively without going into an **endless loop**. I studied the code to the best of my ability and could not find any way this was being accomplished. The **exclude** and **excluded** variables appear to ALWAYS BE EMPTY - never actually filled up with any results of a `<$list>` filter.

I think **ALL** the macros that make table of Contents do not stop recursive calling.

The RECURSIVE ERROR does not show up simply because all the other macros have $reveal areas.

Following the same above, this time in the : New Tiddler
text:
```
<div class="tc-table-of-contents">
<<toc-expandable "Contents">>
</div>
```
You can expand the link **First** and **Contents** for ever! The same can be seen with ALL the toc macros.

@pmario  Please also add this to the top of Meta: Issues with Table of Contents Macros #3627

Jermolene:
Thanks @00SS this is indeed a regression, and an embarrassing one!

Fix is upcoming, but what I realised while working through it is that arguably we don't actually need the recursion protection for the selectable TOCs (i.e. where the user has to click to disclose child entries). In those cases, there's no possibility of a JS recursion error. Cutting out the recursion detection makes the macros significantly simpler and therefore should improve performance.

00SS:
@Jermolene 
There has been some [discussion in Google Groups](https://www.mail-archive.com/search?q=Table+Contents+toc+exclude&l=tiddlywiki%40googlegroups.com) centered around using the **exclude** parameter to "exclude" certain tags from being included in a Table of Contents.

I imagine some resistance to removing this parameter.

00SS:
I noticed the Commit today of: [Restore exclude parameter for selective TOC macros](https://github.com/Jermolene/TiddlyWiki5/commit/6e81122fa3ac0b74764b86347f84e2805f0ece34)

Please bear with me, as the toc macros are quite confusing for me to unravel. I may be way off base, but I will still try and give an opinion.

[...]

Issue #3627, opened 2018-12-07
Meta: Issues with Table of Contents Macros


pmario:
Proposal to have a closer look at the TOC macros

 - [ ] add new "auto-close-others" if a new path is selected. see [code from Andreas](https://groups.google.com/d/msg/tiddlywiki/DNNRxc94MJM/e2pzQVjNABAJ)
 - [x] remove `path` and `itemClassFilter` from `<<toc-body>>` macro. 
    - They are not needed there.  
 - [x] simplify `path` construction, since new code doesn't do, what it is supposed to do. 
   - The problem is the `<<qualify>>` macro and not how the `path` variable was constructed
 - [x] fix `toc-item` - `toc-item-selected` class assignment
 - [x] make the `exclude` parameter usable
   - At the moment it is an internal parameter, which can be "misuesed" 
   - misusing the param is very error prone for new users
 - [x] Table of Contents toc-caption macro awkward behaviour #3624
 - [x] summary - issues found while refactoring core TOC macros #2646
 - [x] BUG: toc macro recursion error #3881
 - [ ] Find a way to implement "auto-expand" feature with the existing TOC. 
   - I did fail with this one, because the existing toc allows multiple tags to be used

BurningTreeC:
Hi @pmario , @Jermolene ... I'd like to add a short piece of code here that comes from a short [discussion in the group](https://groups.google.com/forum/#!searchin/tiddlywiki/wikify$20help|sort:date/tiddlywiki/rnC8doLyI0M/iaKgd_oNAgAJ) and looked very interesting to me, I was thinking that this could be a fast way to create toc-structures (I've changed it a bit to be two-level, but like the `class_sub` part I think it can be made (double-) recursive)

```
\define create-tiddler-title()
[[[[$(currentTiddler)$]]]]
\end

\define class_sub() 
<$list filter="[all[tiddlers]contains:tags{!!title}]"> 
<<create-tiddler-title>>
<$macrocall $name="class_sub">/> 
</$list> 
\end

<$list filter="[[TableOfContents]]">
<$wikify name="all_subclasses" text=<<class_sub>> output="text">
<$list filter="[enlist<all_subclasses>tag{!!title}]">
<<list-links filter:"[enlist<all_subclasses>tag{!!title}]">>
</$list> 
</$wikify> 
</$list>
```

BurningTreeC:
@Jermolene no it doesn't I hadn't checked that ...

[...]

Issue #3624, opened 2018-12-06
Table of Contents toc-caption macro awkward behaviour


00SS:
When using the Table of Content macros:

The top (first) macro in the Shadow file: [$:/core/macros/toc](https://tiddlywiki.com/#%24%3A%2Fcore%2Fmacros%2Ftoc) is -

```
\define toc-caption()
<$set name="tv-wikilinks" value="no">
  <$transclude field="caption">
    <$view field="title"/>
  </$transclude>
</$set>
\end
```

This means that if a Table of Contents macro lists a  **tiddler** that has a **caption** field, but the field is **empty**, the a table of contents listings shows an **blank** space for the title.

Would something like the below address this issue?

```
\define toc-caption()
<$set name="tv-wikilinks" value="no">
<$list filter="[all[shadows+tiddlers]has[caption]]" 
      emptyMessage="""<$transclude field="title"><$view field="title"/></$transclude>""">
  <$transclude field="caption">
    <$view field="title"/>
  </$transclude>
  </$list>
</$set> 
\end
```

pmario:
I think you are right. ... but an empty caption creates problems with all other lists too! ... It should not be empty!

pmario:
If it is empty, it should be removed!

Jermolene:
Hi @00SS the reason that the toc-caption macro is separated out is so that it can be overridden for customisation purposes. So you can use your definition of toc-caption with a local definition, or a `<$importvariables>` widget, and then invoke the core TOC macro which will then use your definition of toc-caption. We don't want to make the core toc-caption macro too complicated because it is typically rendered many times.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

00SS:
An empty title in a list can be quite confusing & disconcerting. Worse might be **not noticing** that such items are **not appearing** at all. Perhaps this can be adjusted in other lists that use captions as well?

So as not to increase overhead, could it be done this way with similar widgets as presently used?:

```
\define toc-caption()
<$set name="tv-wikilinks" value="no">
<$view field="caption"><$transclude field="title"/></$view>
</$set>
\end
```

Jermolene:
Hi @00SS in that sample you've transposed the view and transclude widgets so that the caption field is being rendered as raw text, and the title field is wikified. That isn't the convention; the caption field is generally wikified (and so can include formatting etc.) while the title is rendered as plain text.

pmario:
[...]

Issue #2646, opened 2016-12-04
summary - issues found while refactoring core TOC macros


pmario:
just a reminder.

**doc needed**

 - [ ] set-widget: "select" attribut - [documentation is missing.](https://github.com/Jermolene/TiddlyWiki5/commit/b35544bf4946596e60f7b0c7f0a55fa490ac2a24#commitcomment-20064623)

**fix it**

Code like this, will never set the filter, since it has a hardcoded "value". 

```
<$set name="myVariable" filter="a b c" value="ASDF">
<$text text=<<myVariable>>/>
</$set>
```
 - [x] todo
A construction like this is used at the [core/macros/toc](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/wiki/macros/toc.tid#L15) and won't work

 - [ ] check hacky CSS

```
.tc-table-of-contents li li a {
	font-weight: normal;
}

.tc-table-of-contents li li li {
	font-size: 0.95em;
	font-weight: 200;
	line-height: 1.5;
}
```

pmario:
@Jermolene 

I'll remove the hardcoded: toc-item-selected, since the standard `<<toc>>` macro sets the value for every element, which is wrong.

pmario:
Close this one in favour of #3627 

Doc was done with TW V5.1.14++

PR #3939, opened 2019-05-02
WIP TOC refactor

pmario:
Hi Folks, 
This is "Work in Progress" atm. 

see: Meta: Issues with Table of Contents Macros #3627

[Here's a video](https://youtu.be/iaevaODdKyA), that shows what's new. 

### Overview 

 - [ ] add new "auto-close-others" if a new path is selected. see [code from Andreas](https://groups.google.com/d/msg/tiddlywiki/DNNRxc94MJM/e2pzQVjNABAJ)
 - [x] remove `path` and `itemClassFilter` from `<<toc-body>>` macro. 
    - They are not needed there.  
 - [x] simplify `path` construction, since new code doesn't do, what it is supposed to do. 
   - The problem is the `<<qualify>>` macro and not how the `path` variable was constructed
 - [x] fix `toc-item` - `toc-item-selected` class assignment
     - [x] make them available for all toc-types. (atm it's only toc-tabbed-xx)
 - [x] make the `exclude` parameter usable for "ordinary users"
 - [x] convert "internal" parameters into variables, so users can't misuse them. 
 - [x] Table of Contents toc-caption macro awkward behaviour #3624
 - [x] summary - issues found while refactoring core TOC macros #2646
 - [x] BUG: toc macro recursion error #3881
 - [ ] Find a way to implement "auto-expand" feature with the existing TOC. 
   - I did fail with this one, because the existing toc allows multiple tags to be used

### New Features

 - New "focused-tiddler" plugin
    - update "focused-tiddler" with a scroll event
    - Function atm: If the tiddler reaches `view-port-height / 4` the variable will be set
 - Set a "focused-tiddler" variable in `tv-history-list` so it's possible to highlight focused tiddler
  - Make `toc-tabbed-xxx` more flexible. So future versions may be used as "in-tiddler-tocs", that can work with headings instead of tiddler titles.


pmario:
Hi folks, 

[Here's a video](https://youtu.be/iaevaODdKyA), that shows what's new. PR will follow soon.

twMat:
@pmario - you're doing really fantastic things here! As you noted over in #3934 your solution might cover needs that "compound tiddlers" address. 

I see a need for styling: For example, one must be able to manipulate the display of the ToC index + left field, including the option to not show these at all but only the actual tiddlers. Maybe this is some programmatic toggle or just a styling matter, I don't know. But overall, I'd propose more user styling control for these ToC widgets. There is [currently](https://tiddlywiki.com/#Table-of-Contents%20Macros) no class parameter in the widget call so all examples in the docs are surrounded with a div.

**A different matter:**
I love the highlighting of the focused tiddler title! Might this technique be used for "parallel scrolling"? - for example in edit mode, scroll the editor wikitext and have the preview follow in sync?

Jermolene:
Hi @pmario thanks for this. I'm not very good at watching videos, and it makes this ticket hard to read when one is on mobile. Please could you update the OP with a list of the changes, their impacts, and perhaps screenshots of anything that can't be readily visualised. We can't make watching a video a mandatory condition of participating in a ticket.

I'll dive in and leave some comments on the code.

core/templates/save-all.tid
```diff
@@ -1,6 +1,6 @@
 title: $:/core/save/all
 
 \define saveTiddlerFilter()
-[is[tiddler]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
+[is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
```
Jermolene: Can we avoid making changes to the default save filters? These are commonly customised by end users.

core/ui/SideBar/Open.tid
```diff
@@ -10,22 +10,31 @@ caption: {{$:/language/SideBar/Open/Caption}}
...
-<$droppable actions=<<drop-actions>>>
-<div class="tc-droppable-placeholder">
-&nbsp;
-</div>
-<div>
-<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini">&times;</$button> <$link to={{!!title}}><$view field="title"/></$link>
-</div>
-</$droppable>
-</div>
+  <div style="position: relative;">
```
Jermolene: It's unclear what changes have been made to this file (if any) because of the addition of indenting. It would be clearer to either not change it, or change the indenting in a separate PR that can be merged more quickly.

core/wiki/macros/toc_template_external-link.tid
```diff
@@ -0,0 +1,7 @@
+title: $:/core/macros/toc/template/external-link
+type: text/vnd.tiddlywiki
+
+<$navigator story="$:/StoryList" history="$:/HistoryList" openLinkFromInsideRiver={{$:/config/Navigation/openLinkFromInsideRiver}} openLinkFromOutsideRiver={{$:/config/Navigation/openLinkFromOutsideRiver}} relinkOnRename={{$:/config/RelinkOnRename}}>
```
Jermolene: The navigator widget is quite heavy weight, do we really have to have an instance for each external link in the TOC?

editions/tw5.com/tiddlers/features/Performance Instrumentation.tid
```diff
@@ -0,0 +1,31 @@
+created: 20190316163428191
+modified: 20190316163657710
+tags: Features
+title: Performance Instrumentation
```
Jermolene: I think this is here because it's in tiddlywiki-com but hasn't been merged into master? I've just pulled the latest changes across so hopefully this will disappear by itself.

core/modules/story.js
```diff
@@ -97,7 +97,8 @@ Story.prototype.addToHistory = function(navigateTo,navigateFromClientRect) {
 	$tw.utils.each(titles,function(title) {
 		historyList.push({title: title, fromPageRect: navigateFromClientRect});
 	});
-	this.wiki.setTiddlerData(this.historyTitle,historyList,{"current-tiddler": titles[titles.length-1]});
+	var current = titles[titles.length-1];
+	this.wiki.setTiddlerData(this.historyTitle,historyList,{"current-tiddler": current, "focused-tiddler": current});
```
Jermolene: The 'focused-tiddler' stuff appears to be mostly but not entirely within a plugin, why is that?

core/ui/SideBar/Open.tid
pmario: There should be indentation only. ... I'll have to check this. I didn't add the style thing intentionally.

pmario: It's already there at tiddlywiki.com see: https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FSideBar%2FOpen ... May be a problem with github highlighting.

core/wiki/macros/toc_template_external-link.tid
pmario: Yea, I know. .. I'll have a look if it can be done with `tv-` variables only. .. Or may be linkcatcher.

editions/tw5.com/tiddlers/features/Performance Instrumentation.tid
pmario: Yea, ... There are 2 commits at the beginning, that I wasn't aware of. ... strange.

core/modules/story.js
pmario: I'm using the built in core mechanism, that is triggered by $:/HistoryList changes only. There is no new code.

The new focus plugin only catches changes if you scroll. 

The problem is: If you click a link in the right sidebar, it triggers a "scroll-into-view". Which is triggered by changing the $:/HistoryList in navigator.js. navigator catches link events. 

The `module-type: storyview` modules like classic.js. are responsible for the animation. After animation is done there is no other signal, to highlight the "now focused" tiddler.

The new focus-highlighting in toc.tid is done with a filter like this:

```
[all[current]]-[<tv-history-list>get[focused-tiddler]]
```

The $:/HistoryList is updated in navigator.js with `this.addToHistory(event.navigateTo,event.navigateFromClientRect);`

which then calls a wiki.js function: `this.wiki.addToHistory(title,fromPageRect,this.historyTitle);` 

which is implemented like this: 

```
exports.addToHistory = function(title,fromPageRect,historyTitle) {
	var story = new $tw.Story({wiki: this, historyTitle: historyTitle});
	story.addToHistory(title,fromPageRect);
};
```

So story.js is the "lowest" module which does the job. -> That's why I needed it to update `focused-tiddler`. Which now can be used also without the plugin in a backwards compatible way. 

Which could be seen in the video :/

pmario: I think, calling `this.wiki.addToHistory` in navigator.js is not the right thing to do. ... But it's the only way do get the work done without changing more elements of the core.

core/templates/save-all.tid
pmario: Not really, if we don't want users to store huge HistoryLists ... IMO it should have been implemented with  a prefix from the beginning. ... 

TWs design should be able to handle several independent "stories". The video explains, why I think, the "tabbed-toc" ... is an independent story. ... There is huge potential in the new concept. 

Like: "In Tiddler Anchor Navigation" and "In Tiddler TOCs", which are requested quite a lot. ...

The new "focused-tiddler" mechanism will enable all of those possibilities. ... Which the video shows.

pmario:
@Jermolene ... I did update the OP

pmario:
@Jermolene ... At the moment the `focused-tiddler` plugin is a `startup` module, which adds a scroll eventhandler to the main window. ... Which is suboptimal. 

IMO the best place would be a "story" based event handler, that is part of the "story-list". ... The problem is, that a "story" consists of 3 elements. 

- the navigator-widget
- the list widget
   - the storyview modules. 

storyview is the only place, that really knows the "scroll-to-top" position. eg: With your experimental "stacked" storyview the "top-most" div is not necessarily the focused-tiddler. 

So the question is, where and HOW should we add the "scroll" event handler?

pmario:
>I love the highlighting of the focused tiddler title! Might this technique be used for "parallel scrolling"? - for example in edit mode, scroll the editor wikitext and have the preview follow in sync?

That's something completely different. ... TOCs work with many tiddlers. ... synced-scrolling would work with rendered DOM elements. ... There is a scrollable-widget, which may be used here, but it would need some work. 

The linkcatcher-widget in combination with the scrollable-widget may be able to handle stuff like this. ... BUT there is no code atm.

Jermolene:
Hi @pmario thanks for the updates. I appreciate that watching the video may make things a bit clearer, but we can't make watching a video be a prerequisite for understanding a PR. Personally, I don't watch videos, listen to podcasts or any of that "real-time" stuff through choice, but I'm more concerned for the people who can't watch videos because of language or accessibility issues. Let's keep to written communication to keep things clearer.

Anyhow, standing back, this is a huge PR that touches on many considerations, and adds a bunch of new features to the core, all of which need proper consideration. Is there any way that it can be split up into smaller chunks? For example, you mentioned that there are bugs in the current TOC, can we just do a PR that fixes those bugs?

pmario:
>...you mentioned that there are bugs in the current TOC, can we just do a PR that fixes those bugs?

I think so. ... The problem are the tabbed-toc-xxx implementations. ... They are a bit hacky. .. But I think we could use the `$:/HistoryList/current-tiddler` field instead of the new `focused-tiddler` field and disable the "highlight" CSS settings.

So users don't see a difference to the existing behavior of the TOC macros. ... I'm not 100% sure about the tabbed-macros. ...

Jermolene:
Hi @pmario great, then I'd suggest first the simplest possible PR to fix the bugs; do you think the other stuff could be made into a plugin? People are evidently eager to experiment with these capabilities...

pmario:
ok

Jermolene:
Thanks @pmario

pmario:
Close this one in favor of #3973 ... Some more PRs will follow with new features. the focused-tiddler plugin will be published as a wikilabs-plugin.

PR #3942, opened 2019-05-12
Scale embedded videos and audios

donmor:
Modified files:
$:/core/modules/parsers/audioparser.js
$:/core/modules/parsers/videoparser.js
Added an attribute to each to make video imported scaled to fit the parent container:
`style: {type: "string", value: "width: 100%; height: 100%; object-fit: contain"}`


donmor:
Well, I applied some changes and this pull request is possibly being closed, then a new one created.

PR #3944, opened 2019-05-14
Make the tag pills looks like buttons

bimlas:
Based on their appearance, it is not clear that the tag pill is a button that can be pressed.

Add a shadow to make it clearer.

Before:

vivaldi_2019-05-15_00-27-35

After:

vivaldi_2019-05-15_00-27-44

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -641,6 +641,8 @@ button.tc-tag-label, span.tc-tag-label {
 	vertical-align: baseline;
 	background-color: <<colour tag-background>>;
 	border-radius: 1em;
+	border-top: 1px solid rgba(255, 255, 255, 0.9);
+	box-shadow: 0px 2px 2px 0px rgba(0, 0, 0, 0.3);
 }
 
 .tc-untagged-separator {


pmario:
I think the days of elevated buttons are long gone. 

You could add a very subtle hover effect but I think that's theme / plugin territory,

pmario:
I'm not a big fan of MS, but the new design language used with Windows Web Apps is great. 
see: https://developer.microsoft.com/en-us/fabric#/controls/web/button

Just to be sure. I'm not proposing to use any of the background tools. I want to point out, that there are a lot of good ideas, which are worth considering.

Jermolene:
Hi @bimlas I'm definitely interested in improvements to the tag pill appearance, but I'm not convinced by this. Part of the problem is that every OS has different conventions for making things look "clickable"; it's hard to make TW's tag pills look more clickable without making the appearance more specific to the conventions of one particular OS. Right now the clickability of the appearance of tags pills just relies on them being visually distinct: we don't have any other UI elements that are the same shape/colour.

Just some general thoughts: I don't think TW should create new visual design conventions, we should reflect the familiar conventions that people are used to from other systems -- in other words, as far as UI components go, we should copy the prevailing conventions, and not try to invent new conventions except for the few things that are truly unique to TW. The trouble with tag pills is that they've almost died out from mainstream web services, in favour of hashtags. So maybe there we'll struggle to find the right thing to copy.

Anyhow, in terms of the implementation, the other issue is that the PR uses fixed colours: for something like this we'd need to adapt to the current palette and tag pill colour in order to work in every colour scheme.

bimlas:
> I'm not a big fan of MS, but the new design language used with Windows Web Apps is great.

> Just some general thoughts: I don't think TW should create new visual design conventions, we should reflect the familiar conventions that people are used to from other systems

OK, the majority won, forget the shadows. :) I'm closing the PR. I read in the Google Group or in another issue that you are planning to apply a CSS framework, so I will not continue experimenting: the framework will provide a solution for a unified and clear look. :+1:

> Anyhow, in terms of the implementation, the other issue is that the PR uses fixed colours: for something like this we'd need to adapt to the current palette and tag pill colour in order to work in every colour scheme.

That's why I used black and white colors with alpha to lighten or darken the covered color.

PR #3948, opened 2019-05-15
Hash based auth mechanism.

harishjp:
Adding a hash based auth which does not require transport of password in plaintext.


pmario:
Hi, Nice work, 

But could you add a little bit of documentation. Also about the workflow. 

Does it also work with: https://tiddlywiki.com/#WebServer%20Parameter%3A%20credentials ?

harishjp:
Thanks, I'll update the documentation.

It currently does not work with the credential file. Should I assume credentials file will contain hashed password instead of the plain text?

I'll have to add a few more params to make it work.
* Take authType as a parameter to activate the right authentication module. This way rest of the parameters can be shared b/w the modules and only password format will change based on the 
Either take an authType and assume that file or username and password will be of the format accepted by the authentication handler.
* Or create new set of parameters which if it is set will activate this authentication handler. This is currently being used, but not with credential file.

pmario:
>It currently does not work with the credential file. Should I assume credentials file will contain hashed password instead of the plain text?

As a user, I'd expect that. ... IMO also better for security.

harishjp:
I've added support for credentials file and also added documentation. Should I merge the commits into one and raise a new pull request? Or is it alright to have any number of commits?

pmario:
>Should I merge the commits into one and raise a new pull request?

It's OK to have several commits in 1 PR.

harishjp:
Please review when possible.

Jermolene:
Hi @harishjp firstly, congratulations! This is a very well worked through PR, and addresses an important new capability. It's always a pleasure to get a PR that includes documentation, too. However, this is still a relatively big and complex PR and I'm afraid it will take a bit of patience to get it into the core.

One area of concern is that it seems that the user experience is incomplete: it doesn't look like there's a way for web users to create an account and register their password. I think we really need a smooth, complete web-based user experience that's suitable for general users before this can go in the core. The trouble of course is that user needs will vary greatly the further we get from the provision of the basic functionality, and so we'll need a good deal of customisability with those higher level features.

I'll leave some comments separately on the code, but I'm afraid there's also some bureaucracy: please could you sign the Contributor License Agreement as described in [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md).

core/modules/server/authenticators/hash.js
```diff
@@ -0,0 +1,91 @@
...
+Returns false if the request couldn't be authenticated having sent an appropriate response to the browser
+*/
+HashAuthenticator.prototype.authenticateRequest = function(request,response,state) {
+	var header = request.headers.cookie || "";
+	var cookies = header.split(';');
+	var username;
+	for (var i = 0; i < cookies.length; ++i) {
+		var parts = cookies[i].match(/([^=]*)=(.*)$/);
+		if (parts && parts[1].trim() === "a") {
+			username = $tw.utils.validateCookie(parts[2], this.secretKey);
```
Jermolene: Coding style is not to use spaces after commas in function parameter lists.

core/modules/server/routes/get-login-hash.js
```diff
@@ -0,0 +1,28 @@
...
...===t.status)return void n(t.response);"Object"==typeof t.response?document.getElementById("error").innerHTML=t.response.error:document.getElementById("error").innerHTML="Server request failed with response code:"+t.status,loginForm.onsubmit=login}};var r="";for(var o in e)r&&(r+="&"),r+=o+"="+encodeURIComponent(e[o]);t.send(r)}return function(){var t=loginForm;return t.onsubmit=function(){return!1},u({user:t.user.value,sid:o},function(e){var r=function(e){for(var n=new a(e.length/2),t=0;t<e.length;t+=2)n[t/2]=parseInt(e.substr(t,2),16);return n}(e.salt);s(i.encode(t.key.value),r).then(function(e){e=new a(e);var n=i.encode(o),t=new a(e.length+n.length);return t.set(e),t.set(n,e.length),s(t,r)}).then(function(e){var n=function(e){for(var n="",t=0;t<e.length;++t)n+=(e[t]<16?"0":"")+e[t].toString(16);return n}(new a(e));u({user:t.user.value,sid:o,key:n},function(e){window.location=e.location})})}),!1}}()</script></head><body><form name="loginForm" method="post" onsubmit="return login()">Username: <input type="text" name="user" size="20"> Password: <input type="text" name="key" size="20"> <input type="submit" value="Login"><div id="error" style="color: red"></div></form></body></html>';
```
Jermolene: It would be better to place the HTML in a separate tiddler so that it can be more easily edited and reviewed. Perhaps we might add a webserver parameter giving the name of the tiddler containing the login template, and for it to default `$:/core/ui/webserver/login-hash`.

core/modules/server/authenticators/hash.js
```diff
@@ -0,0 +1,91 @@
...
+	this.server = server;
+}
+
+/*
+Returns true if the authenticator is active, false if it is inactive, or a string if there is an error
+*/
+HashAuthenticator.prototype.init = function() {
+	// Read the credentials data
+	var authType = this.server.get("auth-type");
+	if (authType !== 'hash') {
```
Jermolene: Coding style is not to have a space between `if` and the opening parenthesis

core/modules/utils/auth.js
```diff
@@ -0,0 +1,172 @@
...
+/*jslint node: true, browser: false */
+/*global $tw: false */
+"use strict";
+if($tw.node) {
+	var crypto = require("crypto"),
+			fs = require("fs"),
+			path = require("path");;
+}
+
+//	iv + '$' + cipher('aes-256-gcm', randomBytes + date , secret) + '$' + tag
```
Jermolene: It would be helpful to have some references to authoritative sources so that others can verify the logic here (I'm assuming this is standard crypto stuff but I'm no expert).

harishjp:
Thanks Jermolene for the review. I'll fix the styling issues. Trying to figure out how to render a tiddler into html before authentication. I'm also looking at changing the password validation algorithm to SCRAM or SRP. Provides more security than hash of hash being used now.

I was thinking of this as a safer alternative to plain text password auth and not a full blown system to create and add users. And so currently I'm not targeting creating users.

PR #3951, opened 2019-05-18
Add indexes to the wiki store to improve performance

Jermolene:
When it is running, TiddlyWiki stores tiddlers in pretty much the simplest possible way: as a JavaScript [object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object) where the properties are the fields, and the values of those properties are the values of the fields. In general, operations such as finding all the tiddlers with a particular tag require scanning through all the tiddlers in the store.

The implementation of the tiddler store has always been carefully encapsulated to permit it to be changed without disturbing the rest of the core. In particular, it was expected that some form of indexing would be required to get reasonable performance with large numbers of tiddlers.

As it's turned out, the optimisations that have been implemented so far have all been small tweaks that haven't changed the overall logic:

* Every time there's a tag lookup we cache the results (although the cache is cleared whenever any tiddler is modified, so this optimisation doesn't help much if tiddlers are changing rapidly)
* We cache the list of tiddler titles because `Object.keys()` can be slow in some situations

Over the last two years @Jermolene's work at https://github.com/Federatial has involved working with some very large wikis (up to 80MB with 60,000 tiddlers). Performance has been reasonable for simple operations but analysing the CPU usage shows that filter operations take a substantial amount of the overall time to refresh the page (the recent work on performance instrumentation in e8d1fbba6c109ccbeaedc1b34e47677c0ada7529 makes it much easier to make these measurements).

So, I've started exploring adding proper indexes to the store: the idea is to maintain, say, an index of which tiddler fields have which value, so that one can instantly look them up.

There are several parts to this work:

* Establishing a new "indexer" module type that handles creating, updating and utilising particular indexes
* Updating the wiki store to update these indexes whenever a tiddler value changes. This is done by patching the store to invoke a set of "indexer" objects for each tiddler modification
* Utilising the indexes when performing lookups. This is done by adding special indexed methods to the "source" tiddler iterator
* Thinking about efficiency (for example, the field indexer included in the initial commit only indexes fields when a search has been performed on them)

Most of those issues are addressed in the initial version of the code, but there are several things to do:

* The store object in boot.js currently "knows" about individual indexers, whereas it shouldn't have any prior knowledge of any of them
* The field indexer is incomplete (it only adds its indexed method to `wiki.eachShadowPlusTiddlers`, and not to the other tiddler iterators

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1037,6 +1037,18 @@ $tw.Wiki = function(options) {
 			return shadowTiddlerTitles;
 		};
 
+	var indexers = [],
+		indexersByName = Object.create(null);
+
+	this.addIndexer = function(indexer,name) {
+		indexers.push(indexer);
+		indexersByName[name] = indexer;
+	};
+
+	this.getIndexer = function(name) {
+		return indexersByName[name] || null;
+	};
+
 	// Add a tiddler to the store
 	this.addTiddler = function(tiddler) {
 		if(!(tiddler instanceof $tw.Tiddler)) {
@@ -1046,6 +1058,7 @@ $tw.Wiki = function(options) {
 		if(tiddler) {
 			var title = tiddler.fields.title;
 			if(title) {
+				var oldTiddler = this.getTiddler(title);
 // Uncomment the following line for detailed logs of all tiddler writes
 // console.log("Adding",title,tiddler)
 				tiddlers[title] = tiddler;
@@ -1054,6 +1067,9 @@ $tw.Wiki = function(options) {
 				}
 				this.clearCache(title);
 				this.clearGlobalCache();
+				$tw.utils.each(indexers,function(indexer) {
+					indexer.update(oldTiddler,tiddler);
+				});
 				this.enqueueTiddlerEvent(title);
 			}
 		}
@@ -1064,6 +1080,7 @@ $tw.Wiki = function(options) {
 // Uncomment the following line for detailed logs of all tiddler deletions
 // console.log("Deleting",title)
 		if($tw.utils.hop(tiddlers,title)) {
+			var oldTiddler = this.getTiddler(title);
 			delete tiddlers[title];
 			if(tiddlerTitles) {
 				var index = tiddlerTitles.indexOf(title);
@@ -1073,6 +1090,10 @@ $tw.Wiki = function(options) {
 			}
 			this.clearCache(title);
 			this.clearGlobalCache();
+			var newTiddler = this.getTiddler(title);
+			$tw.utils.each(indexers,function(indexer) {
+				indexer.update(oldTiddler,newTiddler);
+			});
 			this.enqueueTiddlerEvent(title,true);
 		}
 	};
@@ -1105,9 +1126,23 @@ $tw.Wiki = function(options) {
 		}
 	};
 
+	this.each.byTag = function(tag) {
+		var titles = getTiddlerTitles();
+		return self.getIndexer("TagIndexer").lookup(tag).filter(function(title) {
+			return titles.indexOf(title) !== -1;
+		});
+	};
+
 	// Get an array of all shadow tiddler titles
 	this.allShadowTitles = function() {
 		return getShadowTiddlerTitles().slice(0);
+	};
+
+	this.allShadowTitles.byTag = function(tag) {
+		var titles = getShadowTiddlerTitles();
+		return self.getIndexer("TagIndexer").lookup(tag).filter(function(title) {
+			return titles.indexOf(title) !== -1;
+		});
 	};
 
 	// Iterate through all shadow tiddler titles
@@ -1139,6 +1174,10 @@ $tw.Wiki = function(options) {
 		}
 	};
 
+	this.eachTiddlerPlusShadows.byTag = function(tag) {
+		return self.getIndexer("TagIndexer").lookup(tag).slice(0);
+	};
+
 	// Iterate through all the shadows and then the tiddlers
 	this.eachShadowPlusTiddlers = function(callback) {
 		var index,titlesLength,title,
@@ -1159,7 +1198,14 @@ $tw.Wiki = function(options) {
 				callback(tiddlers[title],title);
 			}
 		}
+	};
 
+	this.eachShadowPlusTiddlers.byTag = function(tag) {
+		return self.getIndexer("TagIndexer").lookup(tag).slice(0);
+	};
+
+	this.eachShadowPlusTiddlers.byField = function(name,value) {
+		return self.getIndexer("FieldIndexer").lookup(name,value).slice(0);
 	};
 
 	// Test for the existence of a tiddler (excludes shadow tiddlers)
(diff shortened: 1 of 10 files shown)


Jermolene:
In tests on a wiki with 60K tiddlers, startup time is reduced by 25% with these optimisations, and refresh time is reduced by a factor of three.

pmario:
Hi @Jermolene ... Very interesting. ... For my alias-plugin I was also looking for a fast index method. I wanted to implement a [trie-structure](https://github.com/trekhleb/javascript-algorithms/tree/master/src/data-structures/trie) The same would be possible for tags. 

The advantage of a trie would be, that dropdown-lookups as the tag-dropdown would be extremely fast, with 1 or 2 characters entered, since the trie already contains all the available possibilities. 

I think all the indices should be implemented as plugins. Even the indices in this PR. They probably need a lot of additional internal memory. So if a plugin implements a different index. eg: tag-trie ... The core would still need memory, but it may not be used. 

just some thoughts.

Jermolene:
Hi @pmario 

> Very interesting. ... For my alias-plugin I was also looking for a fast index method. I wanted to implement a trie-structure The same would be possible for tags.

It's tricky to know how far to go with optimisation. Behind the scenes the browser is using tries  to represent object hashmaps, so one wants to lean on the optimisations that they've done as much as possible.

The tag indexer currently stores an array of tiddler titles to keep track of which tiddlers have a particular tag; I do wonder if it might be more efficient to use a hashmap, because I think they might be more optimised for insertions and deletions.

> The advantage of a trie would be, that dropdown-lookups as the tag-dropdown would be extremely fast, with 1 or 2 characters entered, since the trie already contains all the available possibilities.

Ah, you mean that the attraction of a trie is that one can do prefix searches? That's certainly true, and would be interesting to explore.

> I think all the indices should be implemented as plugins. Even the indices in this PR. They probably need a lot of additional internal memory. So if a plugin implements a different index. eg: tag-trie ... The core would still need memory, but it may not be used.

The thing is that we already ship with a tag indexer with almost exactly the same memory consumption, but much worse performance because we throw it away and rebuild it rather than updating it. So, I'm pretty sure we should ship with the tag indexer.

As to the field indexer, the thing here is that it only selectively indexes the fields that have been queried. The tw5.com edition boots up only using three indexes for the `role`, `draft.of` and `plugin-type` fields.

You can see which fields are currently being indexes with the following command in the console:

```
Object.keys($tw.wiki.getIndexer("FieldIndexer").index)
```

You can get a sense of the entire size of the index with this:

```
JSON.stringify($tw.wiki.getIndexer("FieldIndexer").index)
```

In the case of a freshly booted tw5.com, the index is 1000 characters in JSON.

So, I'd be inclined to ship with both these indexers (and I've love to explore more sophisticated ones for plugins). The great thing is that there's nothing stopping us pulling or changing the indexers in the future because they're not visible to end users.

pmario:
>Ah, you mean that the attraction of a trie is that one can do prefix searches?

That's right. 

>The thing is that we already ship with a tag indexer with almost exactly the same memory consumption, but much worse performance because we throw it away and rebuild it rather than updating it. So, I'm pretty sure we should ship with the tag indexer.

do you mean "global cache"? ... I once wanted to use it, but I found out, that it is thrown away most of the time :/ ... So if the tag-index replaces this one .. I'm fine.

>So, I'd be inclined to ship with both these indexers (and I've love to explore more sophisticated ones for plugins). The great thing is that there's nothing stopping us pulling or changing the indexers in the future because they're not visible to end users.

I'm convinced :)

pmario:
Did you consider, to implement the `xxx.rebuild()` function as a worker thread? so it doesn't slow down the boot process. 

Do you know, if and how much does it slow down boot.js?

Jermolene:
Hi @pmario the present architecture wouldn't allow for asynchronous indexers. But in any case, the indexing is deferred until it's needed, and the initial indexing is no more burdensome than the previous code that scanned every tiddler. So, we should see the initial lookup being only a little slower than previously, with subsequent lookups greatly accelerated. This PR doesn't really introduce any additional computation, it just shifts it around so that we can more efficiently avoid some re-computation.

pmario:
OK. I have to have a closer look at the code :)

Jermolene:
Hi @pmario those were all good questions, and rather matched my concerns when I started this...

Jermolene:
Hi @pmario 

> In the case of a freshly booted tw5.com, the index is 1000 characters in JSON.

As another check, I booted tw5.com, and then entered `[text[foo]]` in the advanced search "filters" tab, and finally ran that same snippet to get the index size. It comes out as 7.1MB!

So, I think I'll explore adding a maximum length for the indexed values, and fall back to a scanning search if searching for values above that limit.

Jermolene:
> So, I think I'll explore adding a maximum length for the indexed values, and fall back to a scanning search if searching for values above that limit.

d989fc97e0122b4cc85f12f56b0eecfe926747bc adds this restriction. The maximum length is not yet easily configurable.

Jermolene:
> As another check, I booted tw5.com, and then entered [text[foo]] in the advanced search "filters" tab, and finally ran that same snippet to get the index size. It comes out as 7.1MB!

Rerunning that experiment when restricting the indexed value length to 50 characters, the JSON index size is now reduced to 122KB.

pmario:
So if you run `[text[foo]]` the most important info should be in the first 128 chars. .. right?

Jermolene:
> So if you run [text[foo]] the most important info should be in the first 128 chars. .. right?

No, the field indexer doesn't index fields that have values longer than 128 characters, and rejects attempts to lookup such values. So searching for "foo" will use the index, but searching for 1,000 character text of "HelloThere" will not build nor use the index.

Jermolene:
I'm going to merge this; I've been using it intensively without running into any problems, and so I think the best way to shake out any further issues now is to merge it so that it gets more widely used.

boot/boot.js
```diff
@@ -1054,6 +1071,9 @@ $tw.Wiki = function(options) {
 				}
 				this.clearCache(title);
 				this.clearGlobalCache();
```
pmario: Do we still need the GlobalCache()? ... I think it isn't used and it is cleared all the time you want to use it.

pmario: @Jermolene  Or should we avoid clearing the global cache, if tiddlers are changed, that don't trigger the "dirty" flag?? Would this make sense?

Jermolene: Hi @pmario I left the global cache mechanism (ie the stuff in boot.js) in place for backwards compatibility with plugins that may be using it. I also left the existing tag caching (ie the stuff in wiki.js) so that it's possible to turn the new indexing stuff off for testing/comparison purposes.

pmario: Did you think about the "dirty" flag possibility. Would it be an option?

I did try to use the global cache. But it's unusable. It's always empty.... At least I couldn't make sense of it.

Jermolene: The cache will only have anything in it if it is being used; try turning off the indexing and you'll see the cache being used instead.

core/modules/filters/has.js
```diff
@@ -45,7 +45,7 @@ exports.has = function(source,operator,options) {
 				if(tiddler && $tw.utils.hop(tiddler.fields,operator.operand) && !(tiddler.fields[operator.operand] === "" || tiddler.fields[operator.operand].length === 0)) {
 					results.push(title);
 				}
-			});
+			});
```
pmario: Seems to be not intended?

Jermolene: Indeed! We really need to add a step to test.sh that will fail if there are any JS source code lines with trailing whitespace...

PR #3960, opened 2019-05-31
Update Saving on TiddlySpot.tid

twMat:
Included mention of community created (i.e created by me) "shortcut" to set up new TW5 on TS.

I have previously hesitated to propose this for the docs because cousins Baird might prefer that users go to their front page and not via an iframe as in the shortcut. I did contact them with this question a few years back but heard nothing back. BUT I figure that because they do have the service running they do want people to use it and my "shortcut" should definitely support this. Without the shortcut, it is a bit messy to get a TW5 onto TS. E.g, prof Steve Schneider rhetorically wondered on how he ever managed to use TS with his students before I created this.


Jermolene:
Thanks @twMat

PR #3961, opened 2019-05-31
Update TiddlyDrive info on Tiddlywiki.com

LordRatte:
I wasn't sure how much I can change but I thought I'd just include a more practical link.

Depending on how much free reign I have for that tiddler, I could put a bunch of other useful info there too in the future. :+1:


Jermolene:
Thanks @LordRatte 

> Depending on how much free reign I have for that tiddler, I could put a bunch of other useful info there too in the future

I think the purpose of these resources tiddlers is:

* To provide the link
* To give the user enough information to decide whether to click on it
* To include keywords that a user might search for

Beyond that I'd be inclined to keep these tiddlers brief.

PR #3964, opened 2019-06-02
Update ListWidget.tid

twMat:
Clarification of emptyMessage attribute, i.e that it is not possible to further process the empty message within the widget. For example, this does not work to get the content of the Alert tiddler:

```
<$list filter="" emptyMessage=Alert>
<$transclude/>
</$list>
```

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ListWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ListWidget.tid
@@ -75,7 +75,7 @@ The content of the `<$list>` widget is an optional template to use for rendering
 |template |The title of a template tiddler for transcluding each tiddler in the list. When no template is specified, the body of the ListWidget serves as the item template. With no body, a simple link to the tiddler is returned. |
 |editTemplate |An alternative template to use for [[DraftTiddlers|DraftMechanism]] in edit mode |
 |variable |The name for a [[variable|Variables]] in which the title of each listed tiddler is stored. Defaults to ''currentTiddler'' |
-|emptyMessage |Message to be displayed when the list is empty |
+|emptyMessage |Message to be displayed when the list is empty. Commands in the widget body are left unevaluated. |
 |storyview |Optional name of module responsible for animating/processing the list |
 |history |The title of the tiddler containing the navigation history |
 


Jermolene:
Hi @twMat I'm confused by what your example is illustrating. If you want the emptyMessage attribute to have the value of the content of the Alert tiddler then you need to transclude the attribute:

```
<$list filter="" emptyMessage={{Alert}}>
<$transclude/>
</$list>
```

It sounds like perhaps the point you are making is that the text of the emptyMessage attribute is used as the content of the list widget, and not as the text of a list entry to use if the list is empty. But I think it might be confusing to point that out, it might be better to better express what the attribute is, rather than what it isn't?

pmario:
I think, he wants to point out, that if the list is empty, the emptyMessage is the only thing that is shown. The "body text" is ignored. .... if the list is empty.

twMat:
Yes, what pmario says. It is not obvious that the *only* output is the emptymessage and that the widget cuts the processing at that, so no further manipulation of the empty message is possible in the widget "body".

Is the proposed text in the PR poorly formulated?

editions/tw5.com/tiddlers/widgets/ListWidget.tid
```diff
@@ -75,7 +75,7 @@ The content of the `<$list>` widget is an optional template to use for rendering
 |template |The title of a template tiddler for transcluding each tiddler in the list. When no template is specified, the body of the ListWidget serves as the item template. With no body, a simple link to the tiddler is returned. |
 |editTemplate |An alternative template to use for [[DraftTiddlers|DraftMechanism]] in edit mode |
 |variable |The name for a [[variable|Variables]] in which the title of each listed tiddler is stored. Defaults to ''currentTiddler'' |
-|emptyMessage |Message to be displayed when the list is empty |
+|emptyMessage |Message to be displayed when the list is empty. Commands in the widget body are left unevaluated. |
```
Jermolene: The trouble with that wording is that it doesn't take into account the `template` attribute, which overrides the widget body. Perhaps it would be clearer to say "Message to be displayed when the list is empty. Note that the message entirely replaces any template specified in the `template` attribute or widget body".

Jermolene:
Hi @twMat I've got an alternative formulation that I'll commit momentarily, let me know what you think.

twMat:
Yes, good clarification. Thanks.

PR #3971, opened 2019-06-07
Implement a data tiddler keyvalues operator

pmario:
This PR includes a new filter operator: `keyvalues`, which allows the output of data-tiddlers key and/or value elements. 

The PR contains the new documentation and groups the existing 3 operators. `indexes, getindex and keyvalues` in the "Filter Operators" overview tiddler. 

Link to a test version will follow


pmario:
Test version can be seen at: http://3971.tiddlyspot.com/

twMat:
Just a note: If this operator is introduced, we should harmonize the lingo. In the [docs](https://tiddlywiki.com/prerelease/#DictionaryTiddlers) it is referred to as *name:value* pairs.

Jermolene:
Hi @pmario this is interesting, and certainly addresses one of the many limitations of working with data tiddlers

As you may recall, my view on data tiddlers has evolved over the years, and I broadly no longer think they were a good idea. The present implementation is limited, and would take a lot of code and complexity to make it universal. The implementation of data tiddlers is spread throughout the core code, and hard to work with. I think that the whole idea was misguided: it was an attempt to avoid the proliferation of tiddlers, but I now think that that is a non-goal.

Against that background, I think of data tiddlers as a candidate for deprecation in v5.2.x, and so I'm reluctant to expand support at this point. It looks like this operator could be delivered as a plugin, which I think is what I would advise at this point, although I'm happy to discuss.

(I quickly glanced at the code and noticed a few issues, notably that it's using some constructions from ECMAScript 2015 or later; we currently keep the core restricted to ECMAScript 5).

twMat:
>As you may recall, my view on data tiddlers has evolved over the years, and I broadly no longer think they were a good idea.

Have you elaborated on this publicly anywhere? I'd be really interested to hear your thoughts. (I'm not asking anyone to comment on this *here*, as this would hijack the thread.)

pmario:
>(I quickly glanced at the code and noticed a few issues, notably that it's using some constructions from ECMAScript 2015 or later; we currently keep the core restricted to ECMAScript 5).

Yea, I thought about a plugin myself because of this "issue". ... But it was worth a try as a PR

pmario:
There has been a thread in the GG about [Count Data dictionary values](https://groups.google.com/forum/#!topic/tiddlywiki/Sxf51oIyhuE), where Tony presented (a lot of) code to solve a relatively "simple" problem.

I did suffer myself some time ago, so I thought I'd create a simpler possibility, to sum up some values, with the new math function of 5.1.20

The main problem with standard tiddler is "deduplication" of filtered output. ... This doesn't happen with data tiddlers. ... That's why I thought it would make a nice fit.

Wattaged:
I'm starting a "don't deprecate data tiddlers" petition now (dddt).

AnthonyMuscio:
Data tiddlers can be a a useful way to bring data in from other files and software so I would hope it is always available by plugin or otherwise. To remove it from the core I think it would be necessary to provide the ability to create tiddlers with key like titles and while without the $:/ prefix can be hidden from default searches. If rather than data tiddlers we convert each "record" to a tiddler we could have thousands of tiddlers we do not want appearing in searches etc, and for that matter competing for Tiddler unique titles.

The fact that I can have more than one DataTiddler containing the same keyname where if it was a tiddler I can only have one is a good reason to keep Data Tiddlers or provide another set of namespace tools. 

Regards
Tony

pmario:
As @Wattaged and @AnthonyMuscio pointed out, data-tiddlers are handy in some cases. eg: If you need an "append only log" as used the [script manager](https://wikilabs.github.io/editions/script-manager/#myNewScript:myNewScript%20myNewScript.links) project I created. 

The list created there was one of the reasons, why I wanted to have the "keyvalues" operator. 

The second usecase that is interesting can be directly seen in the list. ... It would be extremely helpful, if an operator yould assign values to new variables. ... So in the list we would be able to access `<<key>>` and `<<value>>` separated and not as a combined <<currentTiddler>>. @Jermolene Is this possible??

Jermolene:
Don't worry! "Deprecating" doesn't mean removing, it just means that a feature is no longer recommended for new projects and will no longer be improved. We can only deprecate things when there is a better alternative. We're quite a long way from having alternatives for every use case for data tiddlers, but my point above is that I think that the primary use case has been to avoid the proliferation of lots of tiddlers, and that I now think that that is a mistake, in that the system pays too high a price in overall complexity to gain a meagre advantage.

kookma:
Hi @Jermolene 
 Are you also against JSON tiddler or you mean only dictionary tiddler to be deprecated?

I think a small database inside TW can be very useful!

Jermolene:
Hi @kookma 

> Are you also against JSON tiddler or you mean only dictionary tiddler to be deprecated?

I am referring to data tiddlers, whether they are in JSON or "data dictionary" format.

> I think a small database inside TW can be very useful!

Jermolene:
Sorry @kookma I meant to add:

> I think a small database inside TW can be very useful!

Indeed, but data tiddlers are not a database, except in the sense that they permit name lookups.

TiddlyTweeter:
@Jermolene 
>As you may recall, my view on data tiddlers has evolved over the years, and I broadly no longer think they were a good idea ...

>... I think that the whole idea was misguided: it was an attempt to avoid the proliferation of tiddlers, but I now think that that is a non-goal.

Broadly agree. But there are use cases where a structure that imitates a simple "database" is conceptually and practically useful. For instance, say you wanted a wiki of the novels of Dickens. Having ONE data tiddler of the paragraphs in order (one record=one paragraph) for one novel makes sense at one data tiddler per book. I find it conceptually easier and practically easier for constructing and using the data that forms the book.

Jermolene:
> Broadly agree. But there are use cases where a structure that imitates a simple "database" is conceptually and practically useful. For instance, say you wanted a wiki of the novels of Dickens. Having ONE data tiddler of the paragraphs in order (one record=one paragraph) for one novel makes sense at one data tiddler per book. I find it conceptually easier and practically easier for constructing and using the data that forms the book.

Right, you've pretty clearly stated my case: that users often find it conceptually easier to think about fewer tiddlers. I don't think people are making that decision based on any data, it's entirely about what feels conceptually comfortable. But my contention is that that intuition is illusory: everything is actually exponentially easier if you break items down into individual tiddlers. And by "everything" I mean both the implementation/performance and the end user experience.

As I said above, if we're going to deprecate data tiddlers then we need to have viable alternatives. For example: I think that what most people value about data tiddlers is the ease of editing multiple items at once in a single text box. We could build similar functionality for working with multiple tiddlers in a couple of ways. First, we could make a spreadsheet component that makes it easy to jump between the fields of a given set of tiddlers. Secondly, we could make a packer/unpacker component that on "edit" packed the text of a bunch of tiddlers into a single tiddler for editing as one run of text, and then on "done" unpacks the text into the original tiddlers.

TiddlyTweeter:
>... I don't think people are making that decision based on any data, it's entirely about what feels conceptually comfortable. But my contention is that that intuition is illusory: everything is actually exponentially easier if you break items down into individual tiddlers. And by "everything" I mean both the implementation/performance and the end user experience.

I agree. IF there can be a method for "consolidated edit." I think the book example is quite a useful one in that the components have a set order and have substantial content, albeit trivially structured.

One point perhaps worth noting is that if you use external tools to create a single large data Tiddler (I use regex tools to do so) its somewhat easier than having to spawn thousands of Tiddlers, though a JSON (non-nested) importable structure works fairly easily too.

Jermolene:
> One point perhaps worth noting is that if you use external tools to create a single large data Tiddler (I use regex tools to do so) its somewhat easier than having to spawn thousands of Tiddlers, though a JSON (non-nested) importable structure works fairly easily too.

In some situations you can generate a `.multids`; the individual tiddlers within it will be imported separately.

twMat:
>... we could make a packer/unpacker component that on "edit" packed the text of a bunch of tiddlers into a single tiddler for editing as one run of text, and then on "done" unpacks the text into the original tiddlers.

Me like. This has parallels to "compound tiddlers" and also the text slicer. 

Another route might be to use fields. They seem very similar to data tiddler entries, at least from a user PoV.

sycom:
> In some situations you can generate a .multids; the individual tiddlers within it will be imported separately.

In this paradigm, `.multids` could be the _new data tiddlers_.  Except that for now, `.mutlids` are not editable from within the browser (or I missed something). Which is a huge limitation for users and devs.

> First, we could make a spreadsheet component that makes it easy to jump between the fields of a given set of tiddlers. Secondly, we could make a packer/unpacker component that on "edit" packed the text of a bunch of tiddlers into a single tiddler for editing as one run of text, and then on "done" unpacks the text into the original tiddlers.

With tools like this, data tiddlers would be overwhelmed by the new mechanism since users and devs would be able to structure far more complex data than with the current data tiddlers without giving up edition simplicity. May be define a `multid` field telling the _saver_ that the tiddler should be saved in the tiddler file defined by `multid` and not in separate files?

How does this concept compare to the plugins / theme / language mechanism ? basically separate files for the filesystem that become json bundles of tiddlers for the browser and are not simply editable from within the browser.

_side note : Is it just me or we need a more structured (and may be constructive?) way to discuss concepts like this one? Conversation is also scattered on the google group and in some issues and PR comments and it's pretty difficult to know where is best way to post any thought. What is the future of the [gitlab space project](https://gitlab.com/tiddlywiki.org)?_

Jermolene:
Hi @sycom 

> In this paradigm, .multids could be the new data tiddlers. Except that for now, .mutlids are not editable from within the browser (or I missed something). Which is a huge limitation for users and devs.

When a multids file is imported it is deconstructed into the individual tiddlers, and there's no way to distinguish between tiddlers imported individually vs. those that arrived in the same multids file.

> side note : Is it just me or we need a more structured (and may be constructive?) way to discuss concepts like this one? Conversation is also scattered on the google group and in some issues and PR comments and it's pretty difficult to know where is best way to post any thought. What is the future of the gitlab space project?

I think GitHub is still the best place for our practical development discussions. I don't see any good reason to migrate from GitHub at the moment (I understand that some people don't trust Microsoft but I actually trust them a lot more than the previous management at GitHub).

sycom:
> When a multids file is imported it is deconstructed into the individual tiddlers, and there's no way to distinguish between tiddlers imported individually vs. those that arrived in the same multids file.

Thanks for the clarification, that is what I had understood. I'm not aware enough of the saving mechanism, but that is why I suggested making it possible to reconstruct the `.multids` (via a `multids` field while deconstructing?) when saving. That would be a good way to avoid file proliferation, particularly for very little files. I guess it's easier to say it than to implement it :-)

> I think GitHub is still the best place for our practical development discussions. I don't see any good reason to migrate from GitHub at the moment (I understand that some people don't trust Microsoft but I actually trust them a lot more than the previous management at GitHub).

Ok, that's crystal clear. I have nothing against Github (even since Microsoft arrival).

Historically, my main repositories are on a Gitlab CE instance because they used to offer more functionalities with free accounts than Github (.yaml, private repo, fine-grained group management, automatic pull and push...). Plus I tend to make use of FLOSS when I have the choice. But the gap is narrower now and I don't mind. I just didn't know what was the status of the tiddlywiki.org project on Gitlab.

Maybe we just have to be more disciplined and systematic in our discussions. e.g. not pursuing such a conversation in a PR comment, as I'm doing now, when it should probably take place in an issue about data-tiddlers future for broader audience ;-) Enthusiastic communities can be a little messy and hard to manage...

Jermolene:
> Maybe we just have to be more disciplined and systematic in our discussions. e.g. not pursuing such a conversation in a PR comment, as I'm doing now, when it should probably take place in an issue about data-tiddlers future for broader audience ;-) Enthusiastic communities can be a little messy and hard to manage...

Yes indeed, it's quite a problem; I'm forever finding myself trying to find comments I made 4 years ago...

pmario:
Closing. ... Will create a new docs PR including the deprecation info.

pmario:
keyvalues operator will get its own plugin.

kookma:
> This PR includes a new filter operator: `keyvalues`, which allows the output of data-tiddlers key and/or value elements.
> 
> The PR contains the new documentation and groups the existing 3 operators. `indexes, getindex and keyvalues` in the "Filter Operators" overview tiddler.
> 
> Link to a test version will follow

Hi @pmario - this solution worth to be added to TW core. There are many use cases for JSON/DataTiddler. and working on them using the current set of filter operator is tedious. 

Hi @Jermolene - while I know you do not encourage developing data tiddler features in current form, I would like to have a look at the present PR. I have been using dataTiddlers (JSON) in few of my plugins including ToDolist, commander, Vote,. ... and they are quite handy as small database. Right now retrieving values is tedious, sorting based on values is very slow, so I appreciate if you can look at this as a bug fix or little improvement.

pmario:
@kookma see: https://wikilabs.github.io/editions/keyvalues/ 

I've seen the GG thread. Mark S. code will work with data-dictionary tiddlers. With the keyvalues operator, you'd need to use the same sortsub as mark did. But it should also work with json tiddlers

kookma:
Thanks @pmario - please see https://github.com/Jermolene/TiddlyWiki5/discussions/5506#discussioncomment-383457

kookma:
Thanks @pmario - please see https://github.com/Jermolene/TiddlyWiki5/discussions/5506#discussioncomment-383457

PR #3973, opened 2019-06-09
fix toc and prepare for link highlighting if element is selected

pmario:
This PR fixes all the problems mentioned in #3939 **without** the additional new features.

 - remove path and itemClassFilter from `<<toc-body>>` macro.
    - They are not needed there.``

 - simplify path construction, since new code doesn't do, what it is supposed to do.
    - The problem is the `<<qualify>>` macro and not how the path variable was constructed

 - fix toc-item - toc-item-selected class assignment
   - make them available for all toc-types. (atm it's only toc-tabbed-xx)

 - make the exclude parameter usable for "ordinary users"

 - convert "internal" parameters into variables, so users can't misuse them.
 - Table of Contents toc-caption macro awkward behaviour #3624
 - summary - issues found while refactoring core TOC macros #2646
 - BUG: toc macro recursion error #3881

-------------

It adds a new field named: `focused-tiddler` to the `$:/HistoryList`. This behaviour is needed, because there is no way to do this with action widgets or a plugin. -> It causes unneeded UI redraws, if done with action widgets.

------------------

There is a new definition in the CSS base, which uses the already existing `sidebar-tiddler-link-foreground-hover` variable, to define `.toc-item-selected` and the new `tc-item-selected` classes.

`tc-item-selected` will be used in future PRs, that will allow us to highlight the selected links in the "Open", "Recent" and "More" tabs.

### DIFF
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -2,98 +2,107 @@ title: $:/core/macros/toc
 tags: $:/tags/Macro
 
 \define toc-caption()
-<$set name="tv-wikilinks" value="no">
-  <$transclude field="caption">
-    <$view field="title"/>
-  </$transclude>
-</$set>
+\whitespace trim
+<$set name="tv-wikilinks" value=no>
+  <$list filter="[all[current]has[caption]]" 
+      emptyMessage="""<$transclude field=title><$view field=title/></$transclude>""">
+     <$transclude field=caption>
+       <$view field="title"/>
+     </$transclude>
+  </$list>
+</$set> 
 \end
 
-\define toc-body(tag,sort:"",itemClassFilter,exclude,path)
+\define getItemClass() [all[current]]-[<tv-history-list>get[focused-tiddler]]
+
+\define toc-body(tag,sort:"")
+\whitespace trim
 <ol class="tc-toc">
-  <$list filter="""[all[shadows+tiddlers]tag<__tag__>!has[draft.of]$sort$] -[<__tag__>] -[enlist<__exclude__>]""">
-    <$vars item=<<currentTiddler>> path={{{ [<__path__>addsuffix[/]addsuffix<__tag__>] }}}>
-      <$set name="excluded" filter="""[enlist<__exclude__>] [<__tag__>]""">
-        <$set name="toc-item-class" filter=<<__itemClassFilter__>> emptyValue="toc-item" value="toc-item-selected">
-          <li class=<<toc-item-class>>>
-            <$list filter="[all[current]toc-link[no]]" emptyMessage="<$link><$view field='caption'><$view field='title'/></$view></$link>">
-              <<toc-caption>>
-            </$list>
-            <$macrocall $name="toc-body" tag=<<item>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> exclude=<<excluded>> path=<<path>>/>
-          </li>
-        </$set>
+  <$list filter="""[all[shadows+tiddlers]tag<__tag__>!has[draft.of]$sort$] -[<__tag__>] -[enlist<visited>]""">
+    <$set name=visited filter="[enlist<visited>] [<currentTiddler>]">
+      <$set name="toc-item-class" filter=<<getItemClass>> emptyValue="toc-item-selected" value="toc-item">
+        <li class=<<toc-item-class>>>
+          <$list filter="[all[current]toc-link[no]]" emptyMessage="<$link><<toc-caption>></$link>">
+            <<toc-caption>>
+          </$list>
+          <$macrocall $name="toc-body" tag=<<currentTiddler>> sort=<<__sort__>>/>
+        </li>
       </$set>
-    </$vars>
+    </$set>
   </$list>
 </ol>
 \end
 
-\define toc(tag,sort:"",itemClassFilter:" ")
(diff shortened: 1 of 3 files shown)


Jermolene:
Hi @pmario thanks. I'm not sure I understand the purpose of the 'focused-tiddler' field. I'll review the code.

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,214 @@ title: $:/core/macros/toc
...
 \define toc-caption()
-<$set name="tv-wikilinks" value="no">
-  <$transclude field="caption">
-    <$view field="title"/>
-  </$transclude>
-</$set>
+\whitespace trim
+<$set name="tv-wikilinks" value=no>
+  <$list filter="[all[current]has[caption]]" 
+      emptyMessage="""<$transclude field=title><$view field=title/></$transclude>""">
```
Jermolene: It's wrong to transclude the `title` field; it should only ever be displayed as plain text.

In any case, I don't think I quite understand what the changes to `toc-caption` are trying to accomplish.

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,214 @@ title: $:/core/macros/toc
...
+     <$transclude field=caption>
+       <$view field="title"/>
+     </$transclude>
+  </$list>
+</$set> 
 \end
 
-\define toc-body(tag,sort:"",itemClassFilter,exclude,path)
+\define getItemClass() [all[current]]-[<tv-history-list>get[focused-tiddler]]
```
Jermolene: It would be clearer to add a single space before the `-`: 

```
\define getItemClass() [all[current]] -[<tv-history-list>get[focused-tiddler]]
```

Jermolene: Again, I don't really understand what's going on here. Why are we touching the `$:/HistoryList` tiddler at all? Using the TOC is nothing to do with navigation (until the point a user clicks on a link).

pmario:
`focused-tiddler` field is used to calculate the "toc-item-selected" status. This allows us to highlight the "latest link" which was clicked. ... It uses the same color as the link "hover" color. So it can be adjusted with the palette settings in the control panel. 

The `focused-tiddler` is also used in preparation to allow plugins, to update this field with a scroll event. ... We could use current-tiddler field. ... BUT it would be needed, that a plugin contains the toc macro, just to change this setting. ...

core/wiki/macros/toc.tid
pmario: If the `caption` field exists, but it is empty, it should use the tiddler title.

core/wiki/macros/toc.tid
pmario: The history tiddler is updated, whenever a link is clicked. So it contains the actual status, which can be used to calculate the `toc-item-selected` status

Jermolene:
> `focused-tiddler` field is used to calculate the "toc-item-selected" status. This allows us to highlight the "latest link" which was clicked. ... It uses the same color as the link "hover" color. So it can be adjusted with the palette settings in the control panel.

Isn't that a new feature? I thought that this version of the PR was just bug fixes.

It's an interesting idea, and I can understand why it's popular but it would take a lot to convince me that it should go into the core. The problem is that it introduces a whole new dependency between the TOC and navigation which I think will give rise to unexpected behaviours. This is experimental stuff that isn't ready for the core.

> The `focused-tiddler` is also used in preparation to allow plugins, to update this field with a scroll event. ... We could use current-tiddler field. ... BUT it would be needed, that a plugin contains the toc macro, just to change this setting. ...

That's interesting stuff, but we really shouldn't introduce major overhauls to the navigation mechanism under the guise of a bug fix.

pmario:
>The changes to the indentation do get in the way; perhaps we might defer them to a subsequent PR so that it's easier for us all to see what has actually changed in this one?

I'm not really sure, what's wrong with the indentation. .. IMO It is the only way to visualize the variable scopes. ... I'll check if I did miss something.

Jermolene:
Background: I think we've let the TOC macros become overloaded with obscure niche features that negatively impact performance. It's best for the core macros to be the simplest, purest expression of the basic functionality so that they make good starting points for further customisation. Trying to make the core TOC macros cater to every requirement just obscures the basic simplicity of how they work.

Jermolene:
> I'm not really sure, what's wrong with the indentation. .. IMO It is the only way to visualize the variable scopes. ... I'll check if I did miss something.

It's not that the indentation is wrong, it's that the diffs are useless because many lines are marked as different when it's just the indentation that has changed. I was suggesting that you do what I've done [here](fddc5d4ee627232441278284ed6bee2f50fe8b26) and [here](5ae14a16ec8bc47e8ee62276cd14e8c1b6f9e0b4), which is to adjust the indentation in a separate PR/commit.

pmario:
The performance problems came with the change from """variable""" to <<__variable__>> ... If we want to speed up the whole thing we should block tiddler names that contain contain characters that cause problems. 

This would speed up the toc macros by almost 100%

pmario:
I did modify the last comment.

pmario:
The diff will show edits in almost every line. ... There have been so many problems, that this version is almost a rewrite.

pmario:
I'll remove the indents

pmario:
@Jermolene  please have a look at: https://tiddlywiki.com/#%24%3A%2Fcore%2Fmacros%2Ftoc  It wasn't the indentation, that caused problems. ... The whole thing is a rewrite. I messed up this PR and will create a new one

PR #3975, opened 2019-06-10
fix toc and prepare for link highlighting if element is selected

pmario:
This PR fixes all the problems mentioned in #3939 **without** the additional new features.

 - remove path and itemClassFilter from `<<toc-body>>` macro.
    - They are not needed there.

 - simplify path construction, since existing code doesn't do, what it is supposed to do.
    - The problem is the `<<qualify>>` macro and not how the path variable was constructed

 - fix toc-item - toc-item-selected class assignment
   - make them available for all toc-types. (atm it's only toc-tabbed-xx)

 - make the exclude parameter usable for "ordinary users"

 - convert "internal" parameters into variables, so users can't misuse them.
 - Table of Contents toc-caption macro awkward behaviour #3624
 - summary - issues found while refactoring core TOC macros #2646
 - BUG: toc macro recursion error #3881

-------------

It adds a new field named: `focused-tiddler` to the `$:/HistoryList`. This behaviour is needed, because there is no way to do this with action widgets or a plugin. -> It causes unneeded UI redraws, if done with action widgets.

------------------

There is a new definition in the CSS base, which uses the already existing `sidebar-tiddler-link-foreground-hover` variable, to define `.toc-item-selected` and the new `tc-item-selected` classes.

`tc-item-selected` will be used in future PRs, that will allow us to highlight the selected links in the "Open", "Recent" and "More" tabs.

### DIFF
--- a/core/wiki/macros/toc-item-class.tid
+++ b/core/wiki/macros/toc-item-class.tid
@@ -1,4 +1,0 @@
-title: $:/core/macros/toc/item-class
-tags: $:/tags/Macro
-
-\define getItemClass() [all[current]] -[<tv-history-list>get[current-tiddler]]
--- a/core/wiki/macros/toc-item-class.tid
+++ b/core/wiki/macros/toc-item-class.tid
@@ -1,0 +1,4 @@
+tags: $:/tags/Macro
+title: $:/core/macros/toc/item-class
+
+\define getItemClass() [all[current]] -[<tv-history-list>get[current-tiddler]]
--- a/core/wiki/macros/toc-item-class.tid
+++ b/core/wiki/macros/toc-item-class.tid
@@ -1,4 +1,4 @@
-tags: $:/tags/Macro
 title: $:/core/macros/toc/item-class
+tags: $:/tags/Macro
 
 \define getItemClass() [all[current]] -[<tv-history-list>get[current-tiddler]]
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -2,98 +2,107 @@ title: $:/core/macros/toc
 tags: $:/tags/Macro
 
 \define toc-caption()
-<$set name="tv-wikilinks" value="no">
-  <$transclude field="caption">
-    <$view field="title"/>
-  </$transclude>
-</$set>
+\whitespace trim
+<$set name="tv-wikilinks" value=no>
+  <$list filter="[all[current]has[caption]]" 
+      emptyMessage="""<$transclude field=title><$view field=title/></$transclude>""">
+     <$transclude field=caption>
+       <$view field="title"/>
+     </$transclude>
+  </$list>
+</$set> 
 \end
 
-\define toc-body(tag,sort:"",itemClassFilter,exclude,path)
+\define getItemClass() [all[current]]-[<tv-history-list>get[focused-tiddler]]
+
+\define toc-body(tag,sort:"")
+\whitespace trim
 <ol class="tc-toc">
-  <$list filter="""[all[shadows+tiddlers]tag<__tag__>!has[draft.of]$sort$] -[<__tag__>] -[enlist<__exclude__>]""">
-    <$vars item=<<currentTiddler>> path={{{ [<__path__>addsuffix[/]addsuffix<__tag__>] }}}>
-      <$set name="excluded" filter="""[enlist<__exclude__>] [<__tag__>]""">
-        <$set name="toc-item-class" filter=<<__itemClassFilter__>> emptyValue="toc-item" value="toc-item-selected">
-          <li class=<<toc-item-class>>>
-            <$list filter="[all[current]toc-link[no]]" emptyMessage="<$link><$view field='caption'><$view field='title'/></$view></$link>">
-              <<toc-caption>>
-            </$list>
-            <$macrocall $name="toc-body" tag=<<item>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> exclude=<<excluded>> path=<<path>>/>
-          </li>
-        </$set>
+  <$list filter="""[all[shadows+tiddlers]tag<__tag__>!has[draft.of]$sort$] -[<__tag__>] -[enlist<visited>]""">
+    <$set name=visited filter="[enlist<visited>] [<currentTiddler>]">
+      <$set name="toc-item-class" filter=<<getItemClass>> emptyValue="toc-item-selected" value="toc-item">
+        <li class=<<toc-item-class>>>
+          <$list filter="[all[current]toc-link[no]]" emptyMessage="<$link><<toc-caption>></$link>">
+            <<toc-caption>>
+          </$list>
+          <$macrocall $name="toc-body" tag=<<currentTiddler>> sort=<<__sort__>>/>
+        </li>
       </$set>
-    </$vars>
+    </$set>
   </$list>
 </ol>
 \end
 
-\define toc(tag,sort:"",itemClassFilter:" ")
-<$macrocall $name="toc-body"  tag=<<__tag__>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> />
+\define toc(tag,sort:"",exclude)
+\whitespace trim
+<$set name=visited filter=<<__exclude__>> >
+  <$macrocall $name="toc-body"  tag=<<__tag__>> sort=<<__sort__>>/>
+</$set>
 \end
 
-\define toc-linked-expandable-body(tag,sort:"",itemClassFilter,exclude,path)
-<!-- helper function -->
-<$qualify name="toc-state" title={{{ [[$:/state/toc]addsuffix<__path__>addsuffix[-]addsuffix<currentTiddler>] }}}>
-  <$set name="toc-item-class" filter=<<__itemClassFilter__>> emptyValue="toc-item" value="toc-item-selected">
+\define toc-linked-expandable-body()
+\whitespace trim
+<$vars toc-state={{{ [[$:/state/toc]addsuffix[/]addsuffix<name>addsuffix<path>addsuffix[/]addsuffix<currentTiddler>] }}} >
+  <$set name="toc-item-class" filter=<<getItemClass>> emptyValue="toc-item-selected" value="toc-item">
     <li class=<<toc-item-class>>>
-    <$link>
+    <$link tooltip="">
       <$reveal type="nomatch" stateTitle=<<toc-state>> text="open">
         <$button setTitle=<<toc-state>> setTo="open" class="tc-btn-invisible tc-popup-keep">
-          {{$:/core/images/right-arrow}}
+          {{$:/core/images/right-arrow}}&nbsp;
         </$button>
       </$reveal>
       <$reveal type="match" stateTitle=<<toc-state>> text="open">
         <$button setTitle=<<toc-state>> setTo="close" class="tc-btn-invisible tc-popup-keep">
-          {{$:/core/images/down-arrow}}
+          {{$:/core/images/down-arrow}}&nbsp;
         </$button>
       </$reveal>
       <<toc-caption>>
     </$link>
     <$reveal type="match" stateTitle=<<toc-state>> text="open">
-      <$macrocall $name="toc-expandable" tag=<<currentTiddler>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> exclude=<<__exclude__>> path=<<__path__>>/>
+      <$macrocall $name="toc-expandableX" tag=<<currentTiddler>> sort=<<sort>> />
     </$reveal>
     </li>
   </$set>
-</$qualify>
+</$vars>
 \end
 
-\define toc-unlinked-expandable-body(tag,sort:"",itemClassFilter:" ",exclude,path)
-<!-- helper function -->
-<$qualify name="toc-state" title={{{ [[$:/state/toc]addsuffix<__path__>addsuffix[-]addsuffix<currentTiddler>] }}}>
-  <$set name="toc-item-class" filter=<<__itemClassFilter__>> emptyValue="toc-item" value="toc-item-selected">
+\define toc-unlinked-expandable-body()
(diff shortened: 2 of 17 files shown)


pmario:
One main change is: **convert "internal" parameters into variables, so users can't misuse them.** So most `<<__variable__>>`s are gone now. ... Only variables, which are exposed to users will be macro parameters. The rest is internal now.

pmario:
The `exclude` parameter is a real user facing parameter now. ... The old `exclude` was renamed to `visited`

Jermolene:
Hi @pmario I think a number of the points I made in my earlier feedback over at #3973 still apply:

* You're still transcluding the title field, rather than viewing it as text. Titles should never be wikified
* `toc-item-selected` etc is a new feature; this PR is supposed to just be bug fixes

> The performance problems came with the change from `"""variable"""` to `<<__variable__>>` ... If we want to speed up the whole thing we should block tiddler names that contain characters that cause problems

I haven't verified the suggestion that text substitution is faster than variable access. It doesn't sound right, and may be the result of a bug of some sort. Anyhow, even if it was true, it would be preferable to have a slower TOC than to have to explain to people endlessly why their TOC suddenly fails when some unexpected character appears in a title.

pmario:
> - toc-item-selected etc is a new feature; this PR is supposed to just be bug fixes

IMO that's not true. check 5.1.14 it has been there for a long time. It just didn't work consistently.

pmario:
> -  You're still transcluding the title field, rather than viewing it as text. Titles should never be wikified

I'll fix that

pmario:
Documentation changes will follow soon!

Jermolene:
> IMO that's not true. check 5.1.14 it has been there for a long time. It just didn't work consistently

I think that originally `toc-item-selected` was used in the tabbed variants of the TOC, and were not used at all for normal navigation. The semantics only work for the tabbed TOC, when there is a single, unambiguous "selected" tiddler.

I can understand your interest in exploring the selected semantics, but I really think it needs to go into a separate macro.

What we're doing at the moment is making a core macro more and more complicated to cater for a wider variety of use cases. I think that it is the wrong approach because it makes the core macros so much more complex for a casual user who wants to customise them. It's easy to make a super-complicated TW5 macro that tries to do everything, but that's not as useful as a simple, transparent implementation that's easy to customise.

pmario:
>I think that originally toc-item-selected was used in the tabbed variants of the TOC, and were not used at all for normal navigation. The semantics only work for the tabbed TOC, when there is a single, unambiguous "selected" tiddler.

You are right. The `toc-item-selected` was used in the tabbed TOC. BUT the parameter was passed to all other toc-macros. ... The mechanism, that was used only worked for the tabbed-toc ... NOW it works for the tabbed tocs and every other toc. ... So I didn't make it more complicated. I fixed it.

pmario:
`toc-item-selected` is used to modify the class attribute of toc elements. There is no need for the links to be "unambiguous" if several of them are shown in the toc. They are all highlighted. .. IMO that's exactly the behaviour that's expected.

Jermolene:
Hi @pmario I'm sorry my views haven't changed. Just to check, am I right in thinking that this introduces a regression whereby tiddler titles ending in a double quote will break things?

Related, I've yet to see proof that the switch from `<<__x__>>` performs better than `$x$`.

But the fundamental problem is that you've introduced a whole new connection between the history mechanism and the TOC, all under the guise of a bug fix. If we're going to support things like that then we need to properly review whether we're doing it right, and understand all the consequences.

There's also my argument above that we should not make the TOC macros any more complex; if we want to add new features then we should do so via new variants.

pmario:
I think the toc-macros are simple. ... The problem is, that it's not 1 TOC macro. There are 6 of them in 1 tiddler.  .. And they reuse each others code in a relatively complex way. 

We could split them up into 6 tiddlers. BUT we would need to do a lot of "code duplication". Which imo would be a maintenance nightmare. .. Even more then now. 

----------

You wrote

>I haven't verified the suggestion that text substitution is faster than variable access. It doesn't sound right, and may be the result of a bug of some sort. Anyhow, even if it was true, it would be preferable to have a slower TOC than to have to explain to people endlessly why their TOC suddenly fails when some unexpected character appears in a title.

That's OK. But the first implementation that tried to address the problem imo was broken in some ways. It seems the the PRs just did search and replace for stuff like: """$variable$"""` --> `<<__variable__>>`  ... BUT that's wrong in cases like: "$variable$ $append-text$" which is not the same as `<<__variable__>> +<<__variable__>>`

So recursion detection and the "path" variable were broken. ... Recursion detection is a problem in 5.1.19, which this PR fixes. .. The path variable is no problem in 5.1.19, since there are no "auto expand / auto close" plugins available at the moment.

pmario:
@Jermolene bump for 5.1.20?

Jermolene:
Apologies @pmario thanks for clearing up the `<<__x__>>` vs. `$x$` situation.

I'm still not happy with making a connection between the TOC and navigation. The rationale that we're fixing support for the `selectedTiddler` parameter won't fly because that parameter was never intended for the vanilla TOC macros.

Jermolene:
I think it would be good to separate out the focus tiddler parts of this PR so that I can merge the bug fix parts.

pmario:
New PR incoming soon :)

pmario:
@Jermolene

 - I did remove the changes to story.js
 - I did add "new in 5.1.20 flags for new parameter: **exclude** and **name**
 - I did move the `current-tiddler` class calculation into a new tiddler, that can be easily overwritten by plugins, that want to handle focused-tiddler instead.

core/wiki/macros/toc-item-class.tid
```diff
@@ -0,0 +1,4 @@
+tags: $:/tags/Macro
+title: $:/core/macros/toc/item-class
+
+\define getItemClass() [all[current]] -[<tv-history-list>get[current-tiddler]]
```
Jermolene: Is the purpose here to avoid repeating this filter? It's not great that it's a global macro because it isn't really useful in its own right, and the name doesn't suggest what it does. I think it would be better to just transclude the filter from the text field of a tiddler called something like `$:/core/macros/toc/item-class`?

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,212 @@ title: $:/core/macros/toc
 tags: $:/tags/Macro
 
 \define toc-caption()
-<$set name="tv-wikilinks" value="no">
-  <$transclude field="caption">
-    <$view field="title"/>
-  </$transclude>
-</$set>
+\whitespace trim
+<$set name="tv-wikilinks" value=no>
```
Jermolene: The coding style for the core is to default to use double quotes around attribute values.

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,212 @@ title: $:/core/macros/toc
...
+  <$list filter="[all[current]has[caption]]"
```
Jermolene: I assume that this is a "fix" for blank caption fields? I think this isn't the right way to approach it. Those tiddlers with a blank caption are broken, and we should provide the tools to fix them. If we attempt to change the semantics of having a blank caption field then it's a disaster to do it inconsistently; it makes the overall system much more complex, with the only benefit being that certain mistakes can be "masked" in certain situations.

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,212 @@ title: $:/core/macros/toc
...
-<!-- helper function -->
-<$qualify name="toc-state" title={{{ [[$:/state/toc]addsuffix<__path__>addsuffix[-]addsuffix<currentTiddler>] }}}>
-  <$set name="toc-item-class" filter=<<__itemClassFilter__>> emptyValue="toc-item" value="toc-item-selected">
+\define toc-linked-expandable-body()
+\whitespace trim
+<$vars toc-state={{{ [[$:/state/toc]addsuffix[/]addsuffix<name>addsuffix<path>addsuffix[/]addsuffix<currentTiddler>] }}} >
+  <$set name="toc-item-class" filter=<<getItemClass>> emptyValue="toc-item-selected" value="toc-item">
     <li class=<<toc-item-class>>>
-    <$link>
+    <$link tooltip="">
```
Jermolene: Why do we have the blank tooltip?

core/wiki/macros/toc.tid
```diff
@@ -2,182 +2,212 @@ title: $:/core/macros/toc
...
 
 \define toc-expandable-empty-message()
-<$macrocall $name="toc-linked-expandable-body" tag=<<tag>> sort=<<sort>> itemClassFilter=<<itemClassFilter>> exclude=<<excluded>> path=<<path>>/>
+<$macrocall $name="toc-linked-expandable-body"/>
 \end
 
-\define toc-expandable(tag,sort:"",itemClassFilter:" ",exclude,path)
-<$vars tag=<<__tag__>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> path={{{ [<__path__>addsuffix[/]addsuffix<__tag__>] }}}>
-  <$set name="excluded" filter="""[enlist<__exclude__>] [<__tag__>]""">
+\define toc-expandableX(tag,sort:"")
```
Jermolene: This seems an unexpected name for a macro because it doesn't meet the conventions of all the other macro names. Are you trying to mark it as internal or private? If so, I'd prefer to do it unambiguously: `\define toc-expandable_private(...` etc.

editions/tw5.com/tiddlers/demonstrations/TableOfContentsMacro Tabbed Example.tid
```diff
@@ -1,13 +1,15 @@
...
+modified: 20190617113004175
 tags: table-of-contents-example
-order: 5
+title: Example Table of Contents: Tabbed Internal
+type: text/vnd.tiddlywiki
 
 <$macrocall
 	$name="toc-tabbed-internal-nav"
 	tag="TableOfContents"
+        sort=""
```
Jermolene: Is there a reason to add these blank attributes? Can we not rely on the defaults?

core/wiki/macros/toc-item-class.tid
pmario: I did separate the filter, so other plugins can easily replace it. .. I'll create a transclusion as you suggest.

core/wiki/macros/toc.tid
pmario: It's a possibility to find the 2 places for debugging. I didn't remove it, since it's much easier to find the right spots.

core/wiki/macros/toc.tid
pmario: Yea. It's a wrapper, that was needed, to handle the "exclude" parameter in the right way. Also changing from `"""$var$"""` to `<<___var__>>` forced some changes in the logic. ...

I use the capital X for extra ... BUT I'll rename it to `_private`

editions/tw5.com/tiddlers/demonstrations/TableOfContentsMacro Tabbed Example.tid
pmario: Yes. ... That's example tiddlers, which should be copy pasted by users. It's much easier to remove attributes that you don't need, than create typos if you type them on your own.

core/wiki/macros/toc.tid
pmario: This is the discussion why I did implement it: Table of Contents toc-caption macro awkward behaviour #3624 

The reason, why I thought it should be implemented is: 

 - Open tiddlywiki.com 
 - Open HelloThere
   - add an empty caption field
 - Go to the sidebar Content
 - And now as a newbie user try to find the problem, which causes the empty TOC line.

core/wiki/macros/toc.tid
pmario: Should I remove it? ... I'd like to keep it for easier debugging

pmario:
@Jermolene ... The `getItemClass()` macro is needed. toc-tabbed-external redefines it to be able to use the linkcatcher workaround. .... 

But I think the implementation now is better, as it was

pmario:
Working with the TOC, I did find several improvemt possibilities for the navigator, linkcatcher and the story handling. I'll make proposals for 5.1.21

core/wiki/macros/toc.tid
Jermolene: That's the rationale I was expecting, but my point above is that having an empty caption field is a mistake that will cause problems anywhere that the caption field is used. Having special logic to "fix" things in the special case of the TOC (but not others) actually makes the system more complex overall.

core/wiki/macros/toc.tid
Jermolene: I think it should be removed.

editions/tw5.com/tiddlers/demonstrations/TableOfContentsMacro Tabbed Example.tid
Jermolene: Good point.

core/wiki/macros/toc.tid
pmario: I'm OK with your reasoning. I did respond in the same way with [my first post](https://github.com/Jermolene/TiddlyWiki5/issues/3624#issuecomment-444930269)

So I'll restore the "existing" behaviour

core/wiki/macros/toc.tid
pmario: ok

Jermolene:
Hi @pmario I notice that the focused tiddler parts are still in the PR. Just to be super clear, I’m not ready to merge those changes for v5.1.20. For v5.1.20 I would just like the bug fixes.

pmario:
@Jermolene 
I just recognized, that the recursion problem is already fixed in master. 

So .. if I create a new PR, that only contains the tabbed-xxxx cosmetic changes. 

Is there any chance, that we get the/this improved TOC into 5.1.20?

pmario:
I did create a new PR with the: Toc cosmetic changes for tabbed-xxxx macros #4101

pmario:
new discussion at: Meta-Issue: Refactor the Navigation mechanism #4180

PR #3976, opened 2019-06-11
[doc] Update currentTiddler Variable

twMat:


Jermolene:
Hi @twMat 

> The value of currentTiddler is stored in the $:/HistoryList `current-tiddler` field.

That's not quite right. The field is poorly named because it doesn't relate to the `currentTiddler` variable (which is in any case local to a particular part of the widget tree), but rather it is used to store the tiddler that was most recently the target of a navigation.

twMat:
Thanks, I did not realize that. I was hoping to inform about where to find the name of the current tiddler. Someone recently asked about this on the boards. I'm closing this PR because I don't know where/if this info then belongs in some other tiddler - and if it doesn't then I don't know what to do with a one sentence tiddler anyway :-)

PR #3981, opened 2019-06-14
Docs for split operator should clarify that it does not remove duplicates in the output

rmunn:
This is just suggested wording; feel free to improve it.


Jermolene:
That's great, thanks @rmunn

Jermolene:
Thanks @rmunn

PR #3982, opened 2019-06-14
Better examples for mathematics operators

### LINKED ISSUE #3979: Providing documentation changes for 5.1.20 tiddlywiki.com? eg mathematics
What is the best way for me to submitt feedbackl for documentation changes for 5.1.20 tiddlywiki.com? eg mathematics prior to its release?

For example
file:///C:/Data/TW5/EmptyFiles/5.1.20Pre-release2019-06-14.html#Mathematics%20Operators

Examples are not using "=" before each number thus will be de-duped if the same value is used more than once. This is fragile, and miss-leading.

rmunn:
Include `=` before each number in the Mathematics Operators examples, so that people get used to seeing that in math examples where dupes are likely.

Only the Mathematics Operators tiddler really needed to be changed; the example tiddlers for individual operators were already using `=`.

Fixes #3979.


Jermolene:
Good point, thanks @rmunn

Jermolene:
Thanks @rmunn

PR #3988, opened 2019-06-15
fr-FR translation for GitHub saver related strings

xcazin:
Plus an old typo left in the fr-FR/ControlPanel.multids file


Jermolene:
Thanks @xcazin

PR #3994, opened 2019-06-17
Add untrunc operator, the mirror of trunc

### LINKED ISSUE #3983: Maybe add a "round away from zero" operator?
Currently the v5.1.20 preview includes the following rounding operators:

`round` - Rounds to the nearest integer, with .5 values rounding towards positive infinity: 1.2 → 1, 1.5 → 2, 1.8 → 2, -1.2 → -1, -1.5 → -1 (this might surprise some people), -1.8 → -2
`ceil` - Rounds towards positive infinity: 1.2 → 2, -1.2 → -1
`floor` - Rounds towards negative infinity: 1.2 → 1, -1.2 → -2
`trunc` - Rounds towards zero: 1.2 → 1, -1.2 → -1

There is, however, no operator that rounds *away from* zero (1.2 -> 2, -1.2 -> -2). That's partly because the rounding operators are based on what Javascript provides, and Javascript does not provide a "round away from zero" operator. If your filter input is a single number, you can make do without writing any Javascript by combining `abs`, `ceil`, `sign`, and `multiply`: store the sign of your number in a variable, then do `[=num +[abs[]] +[ceil[]] +[multiply<signOfNum>]]`. But if your input is a list of numbers, and you really want to round them all *away* from zero there's no simple workaround and you'll have to write a Javascript filter yourself.

It would be trivial to implement a filter providing Javascript's missing operator. Simply add this snippet to `core/modules/filters/math.js`:

```javascript

exports.whatShouldThisBeCalled = makeNumericBinaryOperator(
	function(a) {return Math.ceil(Math.abs(a)) * Math.sign(a);}
);
```

The tricky part, however, is coming up with a good name. `roundAwayFromZero` or `untrunc` are the best I could come up with. There doesn't seem to be a standard name for this operation that I could find; indeed, the IEEE 754 standard (the one that defines how computer languages should handle floating-point numbers) only requires languages to provide a `ceil`, `floor`, `trunc`, and round-to-nearest operation, and it doesn't require the opposite of `trunc` (always round away from zero). This suggests that it's not often needed and this may not actually be a feature that people desire.

Therefore, I have two questions for anyone who reads this feature request:

1. Would a round-away-from-zero operator be useful to you? Why?
1. What would you prefer for it to be called? Any clever ideas for naming this operator?

rmunn:
Fixes #3983.

### DIFF
--- a/core/modules/filters/math.js
+++ b/core/modules/filters/math.js
@@ -43,6 +43,10 @@ exports.trunc = makeNumericBinaryOperator(
 	function(a) {return Math.trunc(a)}
 );
 
+exports.untrunc = makeNumericBinaryOperator(
+	function(a) {return Math.ceil(Math.abs(a)) * Math.sign(a)}
+);
+
 exports.sign = makeNumericBinaryOperator(
 	function(a) {return Math.sign(a)}
 );
--- a/editions/prerelease/tiddlers/Release 5.1.20.tid
+++ b/editions/prerelease/tiddlers/Release 5.1.20.tid
@@ -9,7 +9,7 @@ type: text/vnd.tiddlywiki
 
 !! Major Improvements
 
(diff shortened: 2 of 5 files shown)


Jermolene:
Great, thanks @rmunn

PR #3996, opened 2019-06-18
Update CheckBox.tid

twMat:


Jermolene:
Thanks @twMat

twMat:
@Jermolene - Forgive me, this PR was an error. Please revoke.

However, the example, as found at very bottom in  https://tiddlywiki.com/prerelease/#CheckboxWidget  does not work correctly. The added tags are faulty.

Jermolene:
Hi @twMat yes indeed -- @Marxsal I think this was your example from #2673, can you recollect what it is supposed to be doing? Many thanks.

Marxsal:
It takes the tags from "Features" and "Encryption" and applies them to the current (CheckBoxWidget) tiddler.
Mat's fix will make things slightly more broken, I think. The fix is to roll back his change and then take out the part inside the triple quoted string:

 `<$action-sendmessage $message="tm-add-tag" $param=<<tag>> />
`

After 2.5 years, I have no idea why this 2nd clause is in there, nor how it escaped testing. Apologies :$

Jermolene:
Thanks @Marxsal. I'm actually inclined to remove the example entirely since it doesn't really demonstrate the checkbox widget and is overly complex.

twMat:
>Thanks @Marxsal. I'm actually inclined to remove the example entirely since it doesn't really demonstrate the checkbox widget and is overly complex.

But the actions (and uncheckactions) attributes are really important and they are by necessity more complex than the the other attributes. An example is absolutely needed IMO. However, maybe it is possible to externalize the argument into a macro, to make the example clearer?

Jermolene:
> But the actions (and uncheckactions) attributes are really important and they are by necessity more complex than the the other attributes. An example is absolutely needed IMO. However, maybe it is possible to externalize the argument into a macro, to make the example clearer?

I couldn't come up with a good example in reasonable time. I don't consider it a particularly useful feature: we already have adequate bindings between the checkbox and the tiddler store, and I struggle to see scenarios where the action string attribute is useful.

twMat:
> I don't consider it a particularly useful feature: we already have adequate bindings between the checkbox and the tiddler store, and I struggle to see scenarios where the action string attribute is useful.

Are you implying one should use other means, presumably then the ButtonWidget, instead of the checkbox when further actions are needed? And then even a second button for uncheckactions....?

Marxsal:
Currently you can use "tag" and "untagged" to flip the presence of a tag in a tiddler. A plausible real-life example where the actions attribute is needed is when you want there to also be an "untagged" Tag that is added when unchecked and taken away when checked.

twMat:
>A plausible real-life example

If we need real-life examples I refer to the [fields2table macro](http://fields2table.tiddlyspot.com/) that I released just the other day. The table "edit toggle" is a styled checkbox that performs multiple actions on both check and uncheck.  The option would be two buttons and two reveal widgets.

Jermolene:
Thanks @Marxsal @twMat that's helpful.

> Currently you can use "tag" and "untagged" to flip the presence of a tag in a tiddler. A plausible real-life example where the actions attribute is needed is when you want there to also be an "untagged" Tag that is added when unchecked and taken away when checked.

In that situation one would try to avoid using the "untagged" tag and instead dynamically read the untagged tiddlers with the `untagged` operator. Part of the problem is that it wouldn't be a two way binding: at present changing the tags/fields on a tiddler automatically changes any checkbox widgets bound to that tiddler. But removing the "untagged" tag wouldn't affect the checkbox display.

> The table "edit toggle" is a styled checkbox that performs multiple actions on both check and uncheck. The option would be two buttons and two reveal widgets.

It's great that it's been convenient, but I don't think that that would be concise enough for the sort of example we want here.

PR #4004, opened 2019-06-20
Example changed to use $param as well

AnthonyMuscio:


Jermolene:
Hi @AnthonyMuscio the button widget attribute is called `param`, not `$param`; that's the action-sendmessage widget.

Issue #3842, opened 2019-03-09
DOCS: tm-modal missing attributes


00SS:
The tiddler: [WidgetMessage: tm-modal](https://tiddlywiki.com/#WidgetMessage%3A%20tm-modal)

I think the first attribute's name should be `$param` and not `param`

Also, I believe there should be mention that in the modal tiddler
- the text in the `subtitle` field will be displayed as the **title** of the modal
  - defaults to no title displayed
- the text in the `footer` field displayed as a right aligned **footer** of the modal
  - defaults to `<$button message="tm-close-tiddler">Close</$button>`

If it can be confirmed that the above is correct, I will attempt to modify [the tiddler](https://github.com/Jermolene/TiddlyWiki5/edit/tiddlywiki-com/editions/tw5.com/tiddlers/messages/WidgetMessage_%20tm-modal.tid) accordingly.

Thanks.

On another note, which deserves its own issue, is that since the `footer` field can be used to display a custom close button, if broken code is used there, there is NO WAY to close the modal. Or is there? `Esc` does not work.

Jermolene:
Hi @00SS there's an underlying issue: all of the widget message documentation was originally written for JS developers. We should rewrite them all consistently. The complication we have to get over with $param vs. param is that both are used e.g. the button widget uses param and the action-sendmessage widget uses $param.

00SS:
Ahhh - I had to read what you wrote a few times before it sunk in. Now it makes sense!

It's almost like we need to write:


~~Name~~ Attribute | Description
-- | --
<br>$param<br>param| Title of the tiddler to be displayed<br>_when used in an ActionWidget_<br>_when used in the $button widget_<br>
paramObject | Hashmap of variables to be provided to the modal
rootwindow |  **yes** or **true** will always display a modal in the wiki-root-window

Please note I changed **Name** to **Attribute** - is this correct?

Would this kind of explanation and display suffice?

I will also at some point (when I understand the concept clearly) make a tiddler to explain what a **Hashmap** means in these types of cases.

PR #4005, opened 2019-06-20
Creation from a template only works if $param is used not param

AnthonyMuscio:
Should paramObject and navigateFromTitle need $


Jermolene:
Hi @AnthonyMuscio there's a bigger underlying issue here: the docs for widget messages is written for JS developers; we should update all of them consistently. But in any case, there's a subtlety because it's `param=x` with (for example0 the button widget, but `$param=x` for the action-sendmessage widget.

(There's some more discussion about this [here](https://github.com/Jermolene/TiddlyWiki5/issues/3842#issuecomment-471175965))

pmario:
Close. Docs change is wrong

PR #4006, opened 2019-06-20
Fix sentencecase operator, add titlecase operator

### LINKED ISSUE #4003: sentencecase behavior is unexpected
The new `sentencecase` operator doesn't produce the results I would expect. Consider the title list "abc [[ten reasons to switch to TiddlyWiki]] Example", containing three titles.

### What I'd expect

I'd expect `sentencecase` to produce a title list consisting of "Abc [[Ten reasons to switch to TiddlyWiki]] Example", where only the *first* word in each title is converted to uppercase and, crucially, the rest of the letters are left alone.

### Current behavior

What `sentencecase` currently produces is "Abc [[Ten Reasons To Switch To Tiddlywiki]] Example". Each word of the multi-word title has been capitalized, and the name TiddlyWiki has become Tiddlywiki. This is more like what I'd expect from an operator called `titlecase`.

### Notes

Why do I say that it's crucial for the rest of the letters to be left alone? Because converting them all to lowercase before capitalizing each one risks converting "TiddlyWiki" to "Tiddlywiki", which may not be the desired result. If `titlecase` (as I'd suggest renaming the `sentencecase` operator to) is made to not touch the rest of the letters, then this has two benefits:

* Scenarios where the user doesn't want "TiddlyWiki" to get a lowercase "W" become possible.
* Scenarios where the user **does** want "TiddlyWiki" to get a lowercase "W" are still possible, by chaining multiple filters together like `lowercase[]titlecase[]`.

So I'd suggest renaming the current implementation of `sentencecase` to `titlecase` (AND editing it to remove the `.toLowerCase()` call in there — the user can do that if they want it, but it won't always be desirable), and creating a new `sentencecase` operator that touches only the first letter of each title, rather than the first letter of each word.

The `titlecase` and `sentencecase` documentation would also need a sentence or two to mention that they *don't* change the case of any other letters, and they should be combined with a `lowercase` filter if the desired result is to change "aBcDeF" into "Abcdef".

rmunn:
This would fix #4003 by changing the `sentencecase` operator to work how I (and others, judging from the upvotes on that issue) would expect it to work. I've also added a `titlecase` operator that capitalizes the first letter of every word in each title (the way `sentencecase` was implemented before this PR) because that's also a very useful operation in some cases.

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -30,6 +30,10 @@ exports.sentencecase = makeStringBinaryOperator(
 	function(a) {return [$tw.utils.toSentenceCase(a)];}
 );
 
+exports.titlecase = makeStringBinaryOperator(
+	function(a) {return [$tw.utils.toTitleCase(a)];}
+);
+
 exports.trim = makeStringBinaryOperator(
 	function(a) {return [$tw.utils.trim(a)];}
 );
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -93,12 +93,19 @@ exports.trim = function(str) {
 };
 
 /*
-Convert a string to sentence case (ie capitalise each initial letter)
+Convert a string to sentence case (ie capitalise first letter)
 */
 exports.toSentenceCase = function(str) {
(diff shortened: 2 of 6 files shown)


Jermolene:
Thanks @rmunn great stuff.

PR #4009, opened 2019-06-21
Update Configuring startup tiddlers.tid

twMat:


ght (approved):
Nice addition

Jermolene:
Thanks @twMat and thanks for the nudge @ght

PR #4010, opened 2019-06-22
Fix examples for sentencecase and titlecase

### LINKED ISSUE #4008: sentencecase examples are wrong
The `sentencecase` examples are using `titlecase`, and the `titlecase Operator (Examples)` tiddler is missing. I'm sure I wrote it, so I probably just missed adding it in my PR. Can't fix it right now as I'm about to head to the airport, but I can fix that this evening.

rmunn:
This will fix #4008. The example tiddlers for titlecase and sentencecase both had the title "sentencecase Operator (Example)", so the wiki was confused and was linking to the wrong one. While I was in there, I fixed their tags as well.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/sentencecase Operator (Examples).tid
+++ b/editions/tw5.com/tiddlers/filters/examples/sentencecase Operator (Examples).tid
@@ -1,6 +1,6 @@
 created: 20190619110741485
 modified: 20190620140353983
-tags: [[uppercase Operator]] [[Operator Examples]]
+tags: [[sentencecase Operator]] [[Operator Examples]]
 title: sentencecase Operator (Examples)
 type: text/vnd.tiddlywiki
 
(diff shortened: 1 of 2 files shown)


AnthonyMuscio:
There is still something wrong. "Tiddler sentencecase Operator (Examples)" needs to be renamed titlecase Operator (Examples) and retagged

And a new Tiddler "sentencecase Operator (Examples)" containing sentence case examples only.

I would do this if I had resolved how to in GitHub/Pre-release

Regards
Tony

rmunn:
@AnthonyMuscio The suggestions you make are already in this pull request. Are you saying "there is still something wrong" based on the current state of https://tiddlywiki.com/prerelease/? Because these fixes haven't been applied to that page yet. Or is there still something wrong even after this PR has been applied? I tested it locally and it seemed to work when I applied this fix on my computer, but if you're still seeing issues even after applying this PR locally, then I probably missed something.

AnthonyMuscio:
RMunn

I am not sure if the faults remain or you addressed them.

I will keep a watching brief and once the changes are applied advise. 

Regards
Tony

Jermolene:
Thanks @rmunn looks good to me.

AnthonyMuscio:
Yes, My observed faults were fixed

All that remains is to also add to **see also titlecase,** in uppercase and lower case operators

PR #4025, opened 2019-06-28
add shortcut to focus sidebar search

BurningTreeC:
ctrl-shift-F (but configurable) sets the focus to the sidebar search input field

it does so by adding and removing a tag `$:/tags/RefreshTag` to the corresponding SidebarSegment which is the simplest solution


Jermolene:
Thanks @BurningTreeC this is cunning stuff! However I wonder if we might not do better with a new widget message `tm-focus-selector` which would take a CSS selector and then attempt to set the focus to the first element identified by the selector.

BurningTreeC:
Hi @Jermolene, your idea sounds much better but I have no idea how to make it happen

Issue #4023, opened 2019-06-28
th-renaming-tiddler - double use - with different return values


pmario:
Hi, 

The following line expects a "tiddler object" as a return value: https://github.com/Jermolene/TiddlyWiki5/blob/274dffe75053b165b4498c42fc70f1fa9488cedf/core/modules/wiki-bulkops.js#L26

The following line expects an "event" as a return value.: https://github.com/Jermolene/TiddlyWiki5/blob/274dffe75053b165b4498c42fc70f1fa9488cedf/core/modules/widgets/navigator.js#L601

IMO **That's a big problem**. ... 

The `tm-rename-tiddler` message isn't documented, so the only affected plugin should be the "multi-user" plugin. ... **PR that introduced the "event" hooks**: https://github.com/Jermolene/TiddlyWiki5/pull/3206

pmario:
My proposal is to rename the "event-hooks"

pmario:
Can you see the problem. 

![th-renaming-tiddler](https://user-images.githubusercontent.com/374655/60330677-434ef100-9993-11e9-86c3-9c18b08bd212.gif)

The same hook is invoked 2 times with different expectations.

Jermolene:
Hi @pmario well spotted!

The original `tm-renaming-tiddler` hook was implemented in 2397f0aa6f7af305807b61ccb1d435c2e6a789ab, and the newer one in #3206. I note that #3206 lacks documentation while there is complete docs in 2397f0aa6f7af305807b61ccb1d435c2e6a789ab (see https://tiddlywiki.com/dev); if documentation had been produced it maybe would have been more likely to have been spotted.

I'd rather fix #3206 to match the existing documentation.

pmario:
>I'd rather fix #3206 to match the existing documentation.

But @inmysocks did create the PR for a special purpose, which expects an event and returns an event. ... So imo it may need a new and different hook-name.

pmario:
We'll see, what Jed says.

Jermolene:
Hi @pmario when we have two things that clash like this the one that "wins" is the one that has docs and/or is first. #3206 should have been done in a backwards compatible manner, and we'll just have to fix that now.

inmysocks:
I am pretty certain that I removed the dependency on the hook from the plugin so there isn't a reason to keep the behaviour, if not I will just update the plugin. I can make a pull request to fix the behaviour in two or three days if no one else gets it first.

pmario:
[...]

PR #4037, opened 2019-06-30
Fix th-renaming-tiddler (see #4023)

inmysocks:
This removes the th-renaming-tiddler hook added in #3206 in favour of the version added in 2397f0a.

As discussed in #4023 the th-renaming-tiddler hook usage is inconsistent.
The invocation added in #3206 is redundant because it is called in the renameTiddler function.

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -601,9 +601,7 @@ NavigatorWidget.prototype.handleRenameTiddlerEvent = function(event) {
 	var paramObject = event.paramObject || {},
 		from = paramObject.from || event.tiddlerTitle,
 		to = paramObject.to;
-	var oldTiddler = this.getTiddler(from),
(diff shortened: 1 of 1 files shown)


pmario:
There are several of them, which may cause problems. see: https://github.com/Jermolene/TiddlyWiki5/pull/3206/files

pmario:
for me the problem is the difference in the return values. all the `th-xxx` hooks return a new tiddler object. in #3206 event objects are returned. ... That's why my suggestion was to rename them

inmysocks:
I don't know what the problem is with those ones. Do you want them to all just take tiddler names instead of events?

To my knowledge there are no conflicts with the usage of those hooks.

inmysocks:
I was copying the `th-navigating` hook which takes an event and returns an event the same way. The reason that I used the event instead of the tiddler object is that you lose flexibility in some of the hooks if you only give the tiddler object or title.

Jermolene:
Thanks @inmysocks @pmario

PR #4044, opened 2019-07-02
fix github readme links formatting at the bottom

BurningTreeC:


BurningTreeC:
... because firefox renders those links differently

Jermolene:
Hi @BurningTreeC the readme file in the root of this repo is actually built from other tiddlers by this script:

https://github.com/Jermolene/build.jermolene.github.io/blob/master/readme-bld.sh

See also this build definition:

https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/tw5.com/tiddlywiki.info#L40-L44

BurningTreeC:
@Jermolene, I was expecting something like this... I don't know where to look for this quirk

Jermolene:
If the problem is that Firefox doesn't like the newlines within those anchor tags, then I think we can fix it by adding a `\whitespace trim` to the `<<list-links>>` macro. I'll have a look.

Issue #4046, opened 2019-07-02
stamp button: wrap selection if prefix and/or suffix fields are given


BurningTreeC:
The idea is to add a condition that looks if a stamp has a prefix or a suffix field and in that case it wraps the selection, adding the defined prefix and suffix and ignoring the text input

BurningTreeC:
a better condition would be, that one or both of the tiddlers named like the original stamp but with a `/prefix` and /or `suffix` suffix are present. In that case it uses those tiddlers - advantage: new lines are preserved

twMat:
>a better condition would be, that one or both of the tiddlers named like the original stamp but with a /prefix and /or suffix suffix are present. In that case it uses those tiddlers - advantage: new lines are preserved

You mean *additonal tiddlers* for that stamp with the *titles* suffixed with `/prefix` or `/suffix` ?

BurningTreeC:
>You mean additonal tiddlers for that stamp with the titles suffixed with /prefix or /suffix ?

Yes that's what I mean. Because if fields are used, we cannot insert new lines, the fields don't accept new line characters

BurningTreeC:
closing, as it's in the core now :)

PR #4047, opened 2019-07-02
make stamp-button optionally wrap selection

BurningTreeC:
This PR makes the stamp button wrap the selection if one of the tiddlers titled like the stamp tiddler but suffixed with either `/prefix` or `/suffix` or `both` exist

if none of both tiddlers exist it works the standard way


BurningTreeC:
see #4046

Jermolene:
Looks good thanks @BurningTreeC. Please could you add an update to the existing docs in `Using Stamp`?

Jermolene:
Thanks @BurningTreeC

PR #4048, opened 2019-07-02
katex dropdown: add description hint

BurningTreeC:
this creates a new snippet with prefilled `description` field as a hint


Jermolene:
Thanks @BurningTreeC

PR #4050, opened 2019-07-03
include wiki in options sent to generateTiddlerFilePath

inmysocks:
This adds options.wiki to the object sent to the generateTiddlerFilePath

The function generateTiddlerFilePath can take a wiki in the options object, but generateTiddlerFileInfo doesn't pass the wiki to it.

This is needed by Bob to save tiddlers in multiple wikis.


Jermolene:
Thanks @inmysocks

PR #4052, opened 2019-07-03
add whitespace trim to pagetemplate, story and sidebar

BurningTreeC:
this PR adds a `\whitespace trim` on top of the three mentioned page templates


Jermolene:
Thanks @BurningTreeC

PR #4053, opened 2019-07-03
Update  "core/modules/utils/utils.js"

BramChen:


Jermolene:
Good thinking @BramChen thank you!

Issue #4054, opened 2019-07-03
Extend HTTP API to retrieve filtered tiddlers


Jermolene:
As [discussed here](https://groups.google.com/d/msgid/tiddlywiki/9cf91e41-dc22-4041-901d-b93bf059c64a%40googlegroups.com?utm_medium=email&utm_source=footer) it would be useful to be able to query the TW5 server for a filtered list of tiddlers.

For example, we could extend the `/recipes/default/tiddlers.json` endpoint with an optional `filter` parameter. For example, to retrieve the tiddlers resulting from the filter `[tag[mytag]]`:

```
GET /recipes/default/tiddlers.json?filter=%5Btag%5Bmytag%5D%5D
```

inmysocks:
I have a setup that is not completely unlike this as part of the Bob server. I am not sure how much of it can be directly copied into the tiddlywiki server but it should be a place to start. I will look at what I have and see if I can do a quick proof of concept with the built-in server.

bimlas:
@Jermolene, I've just opened a PR, please check it out.

bimlas:
Just created a userscript to combine Google and TiddlyWiki search results:

https://github.com/bimlas/userscript-combine-tw5-and-search-engine-results

The webserver API filter option allows you to search for more than one wiki at a time. This could somehow be used in TW-Bob in my opinion (or it is already implemented @inmysocks ?)

inmysocks:
Bob has a way tofetch a filtered list of tiddlers from a wiki. It is a bit more complex because of the multiple wikis and authentication used.

I need to look at the search engine results script you made to see if I can see if it would work with Bob.

I like the idea, it looks like a good step toward federation in the core.

Jermolene:
This is now fixed in #4373

PR #4055, opened 2019-07-03
Extend HTTP API to retrieve filtered tiddlers

bimlas:
See #4054 for details.

I tried to use the `?filter=foobar` form, but the `?` character did not work for some reason, so I followed the [permalink syntax](https://tiddlywiki.com/#PermaLinks) and used the `:` character to mark the filter.

```
GET /recipes/default/tiddlers.json:%5Btag%5Bmytag%5D%5D
```

In the first commit I wrote the code in a separate file, in the second I merged the parametrized and plain routes in to one file.

I have not yet written the documentation and test because I don't know if you are at all satisfied with it.


Jermolene:
Hi @bimlas 

> I tried to use the `?filter=foobar` form, but the `?` character did not work for some reason, so I followed the permalink syntax and used the : character to mark the filter.

I think the main server module needs some adjustments to parse query strings and pass them along to the route handlers. It's something for after v5.1.20.

24king:
good job for this function

Jermolene:
This functionality was included in #4373 so closing this.

PR #4066, opened 2019-07-05
Update core/modules/savers/github.js

BramChen:
Using "$tw.utils.base64Encode()" to encode content of data instead of calling module library directly.


Jermolene:
Thanks @BramChen

PR #4070, opened 2019-07-05
outsource tags, searchresults, type and fields filters

BurningTreeC:
this PR creates $:/core/filters/... tiddlers that store the filters used in the tag-picker and search results

the tag-picker and the search-results make use of the new `subfilter` operator

### DIFF
--- a/core/wiki/Filters.multids
+++ b/core/wiki/Filters.multids
@@ -1,6 +1,9 @@
 title: $:/core/filters/
 
+Fields: [!is[shadow]!is[system]fields[]search:title{$:/temp/newfieldname}sort[]] -created -creator -draft.of -draft.title -modified -modifier -tags -text -title -type
+Fields/System: [fields[]search:title{$:/temp/newfieldname}sort[]] -[!is[shadow]!is[system]fields[]]
 SearchResults/AllMatches: [!is[system]search{$(searchTiddler)$}sort[title]limit[250]]
 SearchResults/Matches: [!is[system]search:title{$(searchTiddler)$}sort[title]limit[250]]
 SystemTags: [tags[]is[system]search:title{$:/temp/NewTagName}sort[]]
 Tags: [tags[]!is[system]search:title{$:/temp/NewTagName}sort[]]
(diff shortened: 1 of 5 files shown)


core/wiki/Filters.multids
```diff
@@ -0,0 +1,10 @@
+title: $:/core/filters/
+
+Fields: [!is[shadow]!is[system]fields[]search:title{$:/temp/newfieldname}sort[]] -created -creator -draft.of -draft.title -modified -modifier -tags -text -title -type
+Fields/System: [fields[]search:title{$:/temp/newfieldname}sort[]] -[!is[shadow]!is[system]fields[]]
+SearchResults/AllMatches: [!is[system]search{$(searchTiddler)$}sort[title]limit[250]]
+SearchResults/Matches: [!is[system]search:title{$(searchTiddler)$}sort[title]limit[250]]
+SystemTags: [tags[]is[system]search:title{$:/temp/NewTagName}sort[]]
+Tags: [tags[]!is[system]search:title{$:/temp/NewTagName}sort[]]
```
pmario: `{$:/temp/NewTagName}` ... imo makes it impossible to reuse it somewhere else. ...

pmario:
What's the advantage of the extra indirection. The hardcoded variables and transclusions seem to make them not really generically reusable.

BurningTreeC:
Hi @pmario - the selectable inputs I introduced with #4062 would make use of those same filters and outsourcing them would make more sense because the mechanism could reuse them

BurningTreeC:
... but the filters could be made more generic

Jermolene:
Hi @BurningTreeC it's a an interesting idea. I guess the concerns might be:

* Does the extra indirection negatively impact performance?
* Are the filter strings generic enough to be reused?
* Future compatibility -- if we subsequently want to change one of these filters we'll have an extra decision: should we change the existing named filter, or leave it as it is for backwards compatibility and create a new named filter with the changes. There are valid reasons to need to do both, so it seems like a lot of friction

On balance, I'd be inclined to keep things as they are. We've already got a series of named filters in the advanced search box, and because they're for end users and not used directly by the core they are much simpler to maintain.

PR #4078, opened 2019-07-07
cancel open popups when input gets focus

BurningTreeC:


BurningTreeC:
I believe we additionally need to query in popup.cancel() if a focused input is embedded within a popup and don't cancel it in that case

BurningTreeC:
>I believe we additionally need to query in popup.cancel() if a focused input is embedded within a popup and don't cancel it in that case

the `link` toolbar button for example would cancel its own popup

if I recall correctly, @Jermolene mentioned a way to query for the element that has focus... we could then detect if it's nested within a popup that would get cancelled

BurningTreeC:
I've added a way in popup.js that detects if a focused input is within a popup that would get cancelled. Now the link dropdown works and the popup cancelling works as one would expect

Jermolene:
> if I recall correctly, @Jermolene mentioned a way to query for the element that has focus... we could then detect if it's nested within a popup that would get cancelled

We can use `this.domNode.ownerDocument.activeElement` to get the element with focus, and then `$tw.utils.domContains()` to check containership.

BurningTreeC:
Hi @Jermolene , I did it a bit differently by querying for `input:focus-within` and detecting if it exists and if it's within a domNode that's about to be cancelled...

BurningTreeC:
Peek 2019-07-08 16-10

PR #4079, opened 2019-07-08
default search-results: remove duplicates

BurningTreeC:
this removes search results from `All` that are already under the first list

### DIFF
--- a/core/ui/DefaultSearchResultList.tid
+++ b/core/ui/DefaultSearchResultList.tid
@@ -5,11 +5,15 @@ caption: {{$:/language/Search/DefaultResults/Caption}}
 \define searchResultList()
 //<small>{{$:/language/Search/Matches/Title}}</small>//
 
-<$list filter="[!is[system]search:title{$(searchTiddler)$}sort[title]limit[250]]" template="$:/core/ui/ListItemTemplate"/>
+<$set name="titleList" filter="[!is[system]search:title{$(searchTiddler)$}sort[title]]">
+
+<$list filter="[enlist<titleList>limit[250]]" template="$:/core/ui/ListItemTemplate"/>
 
 //<small>{{$:/language/Search/Matches/All}}</small>//
 
-<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]] -[!is[system]search:title{$(searchTiddler)$}] +[limit[250]]" template="$:/core/ui/ListItemTemplate"/>
+<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]] -[enlist<titleList>] +[limit[250]]" template="$:/core/ui/ListItemTemplate"/>
+
+</$set>
 
 \end
 <<searchResultList>>
--- a/core/ui/DefaultSearchResultList.tid
+++ b/core/ui/DefaultSearchResultList.tid
@@ -9,7 +9,7 @@ caption: {{$:/language/Search/DefaultResults/Caption}}
 
 <$list filter="[enlist<titleList>limit[250]]" template="$:/core/ui/ListItemTemplate"/>
 
-//<small>{{$:/language/Search/Matches/All}}</small>//
+//<small>{{$:/language/Search/Matches/Other}}</small>//
 
 <$list filter="[!is[system]search{$(searchTiddler)$}sort[title]] -[enlist<titleList>] +[limit[250]]" template="$:/core/ui/ListItemTemplate"/>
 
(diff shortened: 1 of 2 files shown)


pmario:
Why do you close your PRs without any discussion?

Jermolene:
Hi @BurningTreeC @pmario I think one problem with it might have been that it required calling the search operator twice for a single search.

BurningTreeC:
reopening for discussion

core/ui/DefaultSearchResultList.tid
```diff
@@ -9,7 +9,7 @@ caption: {{$:/language/Search/DefaultResults/Caption}}
 
 //<small>{{$:/language/Search/Matches/All}}</small>//
 
-<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]limit[250]]" template="$:/core/ui/ListItemTemplate"/>
+<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]limit[250]] -[!is[system]search:title{$(searchTiddler)$}]" template="$:/core/ui/ListItemTemplate"/>
```
pmario: I think, the `limit[250]` should be at the end, after the list is shortened.

pmario: if that's possible.

pmario:
>... @pmario I think one problem with it might have been that it required calling the search operator twice for a single search.

IMO that's an easy fix. We can use `<$set name="titleList" filter="[!is[system]search:title{$(searchTiddler)$}sort[title]limit[250]]">` as a wrapper and 

` -[enlist<titleList>]` as a second run and the first list filter (without the -)

core/ui/DefaultSearchResultList.tid
BurningTreeC: you're right I think, it's fixed

pmario:
I didn't test the above proposal, but I think it can be done, with only 1 "real" filter evaluation

BurningTreeC:
@pmario - it works like you described

BurningTreeC:
`All Matches` should now probably say `All other Matches`

Jermolene:
Thanks @BurningTreeC. Funnily enough I did experiment with the same idea at some point. But I found that it was apt to be confusing in some situations: I'd find myself running my eye down the "All matches" list and then be confused that there were missing entries that were only in the list above. I think I'm saying that the current presentation of the two lists invites the user to perceive them as two separate lists. For this idea to work, I think we need a presentation that makes it crystal clear that it is now a single list, with some ordering applied.

pmario:
> `All other Matches`

I think `Other Matches` is enough, since `All` isn't true, since it is limited to 250

BurningTreeC:
@Jermolene @pmario, how about a presentation like the following?

Bildschirmfoto vom 2019-08-03 07-23-43

twMat:
>, how about a presentation like the following?

@BurningTreeC  How does this differ from 5.1.19... other than the line on the left which I find redundant?

BurningTreeC:
>How does this differ from 5.1.19

The second list doesn't contain the entries from the first one

twMat:
>The second list doesn't contain the entries from the first one

But it does. Both in 5.1.19 and in your image.

BurningTreeC:
@ twMat, in the image it's only about the presentation. The changes made here are not included

BurningTreeC:
@Jermolene @pmario @twMat ... can we merge this? I think I've addressed all points @pmario mentioned

BurningTreeC:
Hi @Jermolene , I'm currently preparing the keyboard-driven mechanism for the search input and its drop down and it's crucial for the mechanism to work that there aren't double-entries in the lists. This PR would address this problem

Jermolene:
Hi @BurningTreeC

> it's crucial for the mechanism to work that there aren't double-entries in the lists

I can see how that might be the case. I'm still not persuaded by this PR, for the reason given above. The logic for avoiding repeating entries is clear, but I don't think it reflects the way that search results are used in practice: it is about helping the eye to scan a list of entries and efficiently pick out the entry that they are looking for. This proposal makes the second list discontinuous, breaking our ability to pick out entries. It seems that only users who have grasped the logic of the search result layout would be able to use it effectively. That's not the case at the moment.

Earlier, I said:

>   For this idea to work, I think we need a presentation that makes it crystal clear that it is now a single list, with some ordering applied

I don't think the vertical bar does that. The typography still invites the user to see it as two separate lists.

On reflection, perhaps it is impossible to reconcile (a) moving the title matches to a separate list and (b) making them look like a single list, so I'm left with my original reservation about the usability of this change.

The fact that this presents a problem suggests that you might be tracking the text of the currently selected dropdown entry. Perhaps it should be done via qualified identifiers instead?

BurningTreeC:
Thanks for your comment @Jermolene, I understand your point of view.

For the qualified titles to work I'd have to experiment, perhaps with a `qualify[]` operator?

Issue #4084, opened 2019-07-09
Prohibiting tiddlers from starting with '+', '-', or '~'


flibbles:
I propose that tiddlers be restricted from starting with '+', '-', or '~',
OR barring that, I propose that the stringified list syntax specify that such tiddlers must be wrapped in brackets (i.e. "tiddlerA tiddlerB [[-tiddler-starting-with-dash]]")

This is so that lists are genuinely subsets of filters, which it seems like there was a concerted effort to make true. It's not quite there though. `A B C +B` is not equivalent when interpreted as a list or a filter.

Having true equivalency has several benefits:

Some settings like `DefaultTiddler` pretend to accept either lists or filters, but that's not true currently. If a new user inputs...
```
HelloThere
--Dashboard--
```
Into "Default tiddlers:" under Basic settings, they'll be confused as to why it's not working.

There's another benefit, and this is the cool one. If lists are truly subsets of filters, then you can put in the most beautiful optimization in the filter compiler. First, it's pretty easy and efficient to tell the difference between a list filter and a non-list filter. In the filters module, you'd have...
```
exports.isList = function(filterString) {
   var index = 0;
   while ((index = filterString.indexOf('[', index)) >= 0) {
      index += 1;
      if (index >= filterString.length || filterString[index] != '[') {
         return false;
      }
      index += 1;
   }
   return true;
};
```
There may be some edge cases involving illegal filters this isn't accounting for, but otherwise, I'm pretty certain that works. Non-list filters will have a standalone bracket somewhere inside of it, which is what this looks for.

Next, you add (something like but not exactly) this into the beginning of `exports.compileFilter`:
```
if (this.isList(filterString)) {
   var list = $tw.utils.parseStringArray(filterString);
   return function() { return list; }
}
```
Next step: Bask in the glorious optimization. I'm not certain how many filters in tiddlywiki could often be downgraded to quick-return lists like that, but I know at least some would (especially in my own projects), as would any in plugins. The isList check is fairly efficient compared to the actual compilation process, though I could profile a bit to make sure.

So I don't know about you, but I think this is totally worth saying that tiddlers starting in '+', '-', and '~' are either disallowed, or must be wrapped in brackets when specified in lists.

-Flibbles

flibbles:
[...]

PR #4087, opened 2019-07-09
Handling for titles starting with `+-~=` (#4084)

flibbles:
* If a user tries to make a tiddler title that starts with `+-~=`, tiddlywiki warns them (but doesn't stop them)
* $tw.utils.stringifyList will wrap tiddlers starting with `+-~=` in brackets
* Changed tw5.com to mention this change.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -314,13 +314,13 @@ $tw.utils.parseDate = function(value) {
 // Stringify an array of tiddler titles into a list string
 $tw.utils.stringifyList = function(value) {
 	if($tw.utils.isArray(value)) {
-		var result = [];
-		for(var t=0; t<value.length; t++) {
+		var result = new Array(value.length);
+		for(var t=0, l=value.length; t<l; t++) {
 			var entry = value[t] || "";
-			if(entry.indexOf(" ") !== -1) {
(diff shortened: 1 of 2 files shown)


flibbles:
This is discussed more in issue: https://github.com/Jermolene/TiddlyWiki5/issues/4084

flibbles:
Alternatively, the title warning can be removed, since starting with `+-~=` works just fine as long as you're willing to wrap such titles in brackets in any lists.

flibbles:
Bah! Since the languages aren't equivalent, there isn't actually a benefit to applying this change.

flibbles:
Okay. I'm reopening it. It seems like despite how this won't close the disparity between the list and filter language, it may still be a good idea to handle titles starting with `+-~=`.

Jermolene:
Hi @flibbles I'm don't think that we need the edit template warning; my concern is that people tend not to read or understand such warnings. From an end user perspective, I think it's safe to use those prefixes as long as care is taken when using those titles in lists.

flibbles:
Okay. Change to the edit template has been removed, as well as the warning in the tw5.com page.

Jermolene:
Thanks @flibbles

PR #4089, opened 2019-07-10
cancel open popups when inputs get focus

BurningTreeC:
This is a new PR based on #4078 

This uses the domNode passed from factory.js - no need to query for a focused element, that domNode has focus at that point

It also detects if an input is placed within a popup and doesn't cancel that one

### DIFF
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -79,6 +79,7 @@ function FramedEngine(options) {
 	// Add event listeners
 	$tw.utils.addEventListeners(this.domNode,[
 		{name: "click",handlerObject: this,handlerMethod: "handleClickEvent"},
+		{name: "focus",handlerObject: this.widget,handlerMethod: "handleFocusEvent"},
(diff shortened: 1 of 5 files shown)


BurningTreeC:
Peek 2019-07-10 12-13

morosanuae:
@BurningTreeC very nice! Jeremy was right. It is much easier now with the help of animations to understand what you are doing.

BurningTreeC:
thanks @morosanuae 

another one that shows nested popups with inputs... the first popup has the class `tc-popup-keep`

Peek 2019-07-10 14-48

BurningTreeC:
Peek 2019-07-14 06-47

Jermolene:
Hi @BurningTreeC I think this is simple enough to merge for v5.1.20 but I'd appreciate your help double checking things once merged.

BurningTreeC:
>but I'd appreciate your help double checking things once merged.

Thanks @Jermolene, I will test test test

PR #4091, opened 2019-07-11
popup cancelling - use state reference

BurningTreeC:
in the google group there's a discussion: https://groups.google.com/forum/#!topic/tiddlywiki/_mDDZ1jpMgU

buttons allow setting a state-reference for popups but the popup mechanism doesn't respect that


Jermolene:
Thanks @BurningTreeC -- when referencing a discussion elsewhere it's very helpful if you're able to copy and paste the relevant quotes for ease of reference on mobile.

BurningTreeC:
Hi @Jermolene - I was wondering if instead of deleting the tiddler, we shouldn't delete just the field or the index

PR #4083, opened 2019-07-09
add keyboard shortcut to add new field

BurningTreeC:

PR #4093, opened 2019-07-12
add kb shortcut to add new field and use qualified states

BurningTreeC:
this PR add a keyboard shortcut in the `fieldvalue` field of tiddlers (`enter` by default) that adds the fieldname/-value pair to the tiddler and sets focus to the `next fieldname` input

the keyboard shortcut is configurable through the control panel

### DIFF
--- a/core/language/en-GB/EditTemplate.multids
+++ b/core/language/en-GB/EditTemplate.multids
@@ -8,6 +8,7 @@ Field/Remove/Hint: Remove field
 Field/Dropdown/Caption: field list
 Field/Dropdown/Hint: Show field list
 Fields/Add/Button: add
+Fields/Add/Button/Hint: Add the fieldname/-value pair to the tiddler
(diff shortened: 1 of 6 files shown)


BurningTreeC:
Peek 2019-07-12 10-33

BurningTreeC:
The previous PR (#4083 ) didn't work correctly (the button wasn't working anymore) which is fixed here

BurningTreeC:
the last update fixes focusing for the `add` button

BurningTreeC:
This PR makes use of the new `escapecss` filter to reliably focus the fieldname field of the current tiddler

BurningTreeC:
#4098 should get merged first, then this PR will have some conflicts that are going to be resolved

BurningTreeC:
This PR now additionally includes everything from #4098

BurningTreeC:
Peek 2019-07-13 09-44

PR #4095, opened 2019-07-12
shortcut in sidebar search: navigate to advanced search

BurningTreeC:
this PR adds a keyboard shortcut (configurable, enter by default) in the sidebar search field to navigate to the advanced search

### DIFF
--- a/core/ui/SideBarSegments/search.tid
+++ b/core/ui/SideBarSegments/search.tid
@@ -1,16 +1,22 @@
 title: $:/core/ui/SideBarSegments/search
 tags: $:/tags/SideBarSegment
 
+\define advanced-search-actions()
+<$action-setfield $tiddler="$:/temp/advancedsearch" text={{$:/temp/search}}/>
+<$action-setfield $tiddler="$:/temp/search" text=""/>
+<$action-navigate $to="$:/AdvancedSearch"/>
+\end
+
 <div class="tc-sidebar-lists tc-sidebar-search">
 
 <$set name="searchTiddler" value="$:/temp/search">
 <div class="tc-search">
+<$keyboard key="((advanced-search))" actions=<<advanced-search-actions>>>
 <$edit-text tiddler="$:/temp/search" type="search" tag="input" focus={{$:/config/Search/AutoFocus}} focusPopup=<<qualify "$:/state/popup/search-dropdown">> class="tc-popup-handle"/>
+</$keyboard>
 <$reveal state="$:/temp/search" type="nomatch" text="">
 <$button tooltip={{$:/language/Buttons/AdvancedSearch/Hint}} aria-label={{$:/language/Buttons/AdvancedSearch/Caption}} class="tc-btn-invisible">
-<$action-setfield $tiddler="$:/temp/advancedsearch" text={{$:/temp/search}}/>
-<$action-setfield $tiddler="$:/temp/search" text=""/>
-<$action-navigate $to="$:/AdvancedSearch"/>
+<<advanced-search-actions>>
 {{$:/core/images/advanced-search-button}}
 </$button>
 <$button class="tc-btn-invisible">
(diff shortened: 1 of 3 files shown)


pmario:
hmmm, I'm not so sure, if it should be `enter` by default. ... Is it consistent to call a "special function" like open Advanced Search with "enter"?

BurningTreeC:
>Is it consistent to call a "special function" like open Advanced Search with "enter"?

I do think so, the tags field also supports adding a tag with `enter`

morosanuae:
@BurningTreeC I don't think this is a good idea. The `enter` should be used to open a tiddler in the search results list. The following sequence should be more natural IMO:

- Focus the search field by pressing a shortcut
- Enter the keywords into the search field
- The first item in the result list should be opened automatically by pressing `enter`
- Or by navigating with arrow keys (up/down) select an item and then press `enter`

BurningTreeC:
@morosanuae right, that would be the right thing to have

pmario:
IMO the PR shouldn't be closed. We are still discussing possible improvements. I could think about CTRL-Enter ...

BurningTreeC:
@pmario @morosanuae , it's <kbd>ctrl-enter</kbd> now

sukima:
<kbd>Shift</kbd>+<kbd>Enter</kbd> seems ok to IMHO (easier to press with one hand) Another one might be <kbd>Ctrl</kbd>+<kbd>F</kbd>. 🤔

morosanuae:
@sukima 
<kbd>Ctrl</kbd>+<kbd>F</kbd> - This will be in conflict with browser search.
<kbd>Shift</kbd>+<kbd>Enter</kbd> - It's a little awkward IMO if you mean pressing 2 adjacent buttons with the little finger.

@BurningTreeC should we need to establish a shortcut for focusing the search field first and then choose one for the advanced search? I saw you close the #4025. Maybe this should come first.

BurningTreeC:
@morosanuae the prerelease already has a shortcut for focusing the search field (it's ctrl-shift-F)... Jeremy made a tm-focus-selector message for that recently which is better than the hack used in my PR

morosanuae:
@BurningTreeC 
> the prerelease already has a shortcut for focusing the search field (it's ctrl-shift-F)

I guess it's not documented in the pre-release notes yet.

And another thing. Is it a good/normal thing to have identical shortcuts with different scopes in the same app? The <kbd>Ctrl</kbd>+<kbd>Enter</kbd> - @pmario proposal - is already in use for confirming the changes for tiddlers. Wouldn't be this confusing for users? I certainly don't like it and I haven't encountered this situation before.

BurningTreeC:
@morosanuae ... if one sees it like "invoke the action for this input" then I think it isn't confusing

morosanuae:
@BurningTreeC 

> if one sees it like "invoke the action for this input" then I think it isn't confusing

If `Ctrl+Enter` would be some kind of universal generic invoking shortcut maybe. Anyway I don't like the idea that a user must see it in some way. A shortcut is just a shortcut, nothing more. Can we find an alternative?

BurningTreeC:
@morosanuae - ctrl-shift-enter, alt-shift-enter, ctrl-shift-A ... it's configurable anyway

morosanuae:
@BurningTreeC I know it's configurable but the best option should be the default one IMO. If the shortcut for focusing the search field is `ctrl+shift+f` I think a similar approach for advanced search would be `ctrl+shift+enter`. It would be easier to remember that way. What do you say?

Jermolene:
Quick observation: the action of the <kbd>enter</kbd > key is *always* specific to the focused user interface element, and I think most users implicitly understand that. However, why don't we just have a keyboard shortcut for opening advanced search: one that doesn't require the search field to be focused? I'm not sure that it should behave as the current advanced search button: whether it should copy any text from the sidebar search box into the advanced search box.

morosanuae:
@Jermolene 
> However, why don't we just have a keyboard shortcut for opening advanced search: one that doesn't require the search field to be focused?

It would be the simplest solution I guess.

> I'm not sure that it should behave as the current advanced search button: whether it should copy any text from the sidebar search box into the advanced search box.

It would be helpful and a natural thing to do. Why not?

BurningTreeC:
@morosanuae @pmario @Jermolene 

I made a separate PR #4103 for the AdvancedSearch keyboard shortcut

PR #4098, opened 2019-07-13
fields edittemplate: use qualified newfieldname ...

BurningTreeC:
and newfieldvalue tiddlers

this PR makes the fields EditTemplate use different $:/temp/newfieldname and $:/temp/newfieldvalue tiddlers for each Tiddler, so that the user input isn't reflected in every open tiddler


BurningTreeC:
closing this in favor of #4093 which has this functionality, too

PR #4099, opened 2019-07-13
tag-picker macro: remove duplicate add-tag-actions

BurningTreeC:


BurningTreeC:
closing because #4100 does this, too

PR #4101, opened 2019-07-13
Toc cosmetic changes for tabbed-xxxx macros

pmario:
This PR only contains the cosmetic changes, that are needed to get the tabbed-xxxx macros highlighting the selected element.

### DIFF
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -58,7 +58,7 @@ tags: $:/tags/Macro
 </$qualify>
 \end
 
-\define toc-unlinked-expandable-body(tag,sort:"",itemClassFilter:" ",exclude,path)
+\define toc-unlinked-expandable-body(tag,sort:"",itemClassFilter,exclude,path)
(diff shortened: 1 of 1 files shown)


pmario:
The recursion fix was merged 4 months ago.

pmario:
I did only run tests against the 5.1.19

 - recursion problems
 - tabbed-xxx macros highlighting problem

Jermolene:
Hi @pmario can/should this be merged independently of #3975?

pmario:
@Jermolene Yes.This is the fix you requested, in 3975. No additional changes made. It doesn't contain any focus-tiddler related stuff.

-------------

With this version every "internal" variable like "exclude" is exposed to users. And they try to misuse them and promote a problematic workflow, that could establish a defacto standard, which we definitely don't want.

see https://groups.google.com/forum/#!msg/tiddlywiki/x3HVIjJrbx8/iof9Ml7KAgAJ  which is just 1 of them

Jermolene:
Thanks @pmario

PR #4102, opened 2019-07-14
reducing function calls across board

flibbles:
I heard y'all like clock cycles. Between `render`, `renderChildren`, `refresh`, and `refreshChildren`, I believe they account for a substantial  amount of function calls in tiddlywiki. You can witness this using any browser profiler. I at least know for a fact that `refreshChildren` can be called 10,000s of times just for opening or closing a tiddler.

This modification removes two out of four function calls in this tight loop.
`render -> renderChildren -> each -> (anonymous)` to `render -> renderChildren`

If I'm correct about this, then this small adjustment should reduce function calls in Tiddlywiki by about 35%.

I do invite anyone to check my work. It's hard to count function calls.

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -385,9 +385,10 @@ Widget.prototype.previousSibling = function() {
 Render the children of this widget into the DOM
 */
 Widget.prototype.renderChildren = function(parent,nextSibling) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @flibbles looks good.

PR #4109, opened 2019-07-15
make plugin addons indent visually

BurningTreeC:
this PR makes Addons (plugins tagged `$:/tags/Plugins/AddOn`) indent 2em in the ControlPanel, see the screenshot. It also adds that tag to all CodeMirror addons

Bildschirmfoto vom 2019-07-15 07-26-42

### DIFF
--- a/core/ui/Components/plugin-info.tid
+++ b/core/ui/Components/plugin-info.tid
@@ -52,7 +52,7 @@ $:/config/Plugins/Disabled/$(currentTiddler)$
 \define plugin-info(type,default-popup-state)
 <$set name="popup-state" value=<<popup-state-macro>>>
 <$reveal type="nomatch" state=<<plugin-disable-title>> text="yes">
-<$link to={{!!title}} class="tc-plugin-info">
+<$link to={{!!title}} class={{{ [tag[$:/tags/Plugins/AddOn]addprefix[tc-plugin-info ]removesuffix{!!title}addsuffix[tc-plugin-addon]] ~[[tc-plugin-info]] }}}>
 <<plugin-table-body type:"$type$" default-popup-state:"""$default-popup-state$""">>
 </$link>
 </$reveal>
--- a/core/ui/Components/plugin-info.tid
+++ b/core/ui/Components/plugin-info.tid
@@ -62,7 +62,7 @@ $:/config/Plugins/Disabled/$(currentTiddler)$
 </$link>
 </$reveal>
 <$reveal type="match" text="yes" state=<<popup-state>> default="""$default-popup-state$""">
-<div class="tc-plugin-info-dropdown">
+<div class={{{ [tag[$:/tags/Plugins/AddOn]addprefix[tc-plugin-info-dropdown ]removesuffix{!!title}addsuffix[tc-plugin-addon]] ~[[tc-plugin-info-dropdown]] }}}>
 <div class="tc-plugin-info-dropdown-body">
 <$list filter="[all[current]] -[[$:/core]]">
 <div style="float:right;">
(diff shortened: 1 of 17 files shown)


BurningTreeC:
@Jermolene - this is just a visual update for the codemirror `"AddOn"-Plugins`

Jermolene:
Hi @BurningTreeC it's a neat idea, and emphasises that we really do need to clean up the current mess...

However, I'm not super keen on adding a tag to plugins because it's quite an obtrusive change to the data model of plugins (we don't currently use tags for low level things like modules or plugins). We already have the "dependents" field, can we perhaps re-use it?

BurningTreeC:
Hi @Jermolene ... using the dependents field would be a mess, too ... many plugins already have that field, even if it's empty

Jermolene:
Hi @BurningTreeC 

> using the dependents field would be a mess, too ... many plugins already have that field, even if it's empty

The "dependents" field is a list of titles of the plugins on which the current plugin depends. So the `$:/themes/tiddlywiki/centralised` plugin has the "dependents" field containing `$:/themes/tiddlywiki/snowwhite` to indicate that the latter plugin should also be loaded.

The "dependents" field is currently only used by themes and languages, but we could also use it to set each of the codemirror sub plugins to mark their dependence on `$:/plugins/tiddlywiki/codemirror`. The tricky bit would be using the "dependents" field in the control panel listing, instead of the current judicious use of alphabetical ordering.

BurningTreeC:
>The tricky bit would be using the "dependents" field in the control panel listing, instead of the current judicious use of alphabetical ordering.

yeah, too tricky for me

Jermolene:
@BurningTreeC on reflection reusing the "dependents" field wouldn't work; we need a singleton field for "sub-plugins" that points to the parent plugin for presentation purposes.

BurningTreeC:
@Jermolene - the sub-plugins (the addons) still need a naming convention so that they are shown below the "parent plugin"

Jermolene:
Hi @BurningTreeC 

> he sub-plugins (the addons) still need a naming convention so that they are shown below the "parent plugin"

I was thinking that the outer loop would be `[!has[draft.of]plugin-type[$type$]field:parent-plugin[]sort[description]]` to select plugins without a `parent-plugin` field, and then underneath each plugin we'd loop through the plugins whose `parent-plugin` field points to the parent.

We'd also need to list plugins that have a `parent-plugin` field but that the specified plugin is missing. We could do that by changing the outer loop to `[!has[draft.of]plugin-type[$type$]field:parent-plugin[]] [!has[draft.of]plugin-type[$type$]get[parent-plugin]] +[sort[description]]` so that all the specified parent plugins are included, and then we'd need a check for plugins that don't exist, so we can mark them.

BurningTreeC:
@Jermolene, is it sufficient to do this in $:/core/ui/ControlPanel/Plugins only?

BurningTreeC:
... I mean so that it also works for the more-plugins modal...

BurningTreeC:
one problem is that codemirror addons can have cross-dependencies

twMat:
IMO subplugins should be visible only when opening the dropdown of the mother plugin, as a "sub list" therein. A indentation, as in the OP, looks awkward when there are only single subtiddlers and it would get uglier the more plugins with subplugins there are making for an irregular zig-zag.

A mothers unrevealed dropdown could have an indication that it has subplugins.

pmario:
The problem is: 

 - Do we want to "just show" possible dependecies, OR
 - Do we want to resolve dependencies. 

We could define "hand made" bundles, that are difficult to maintain, or we just introduce a short algorithm, that can resolve the problem. 

similar to: https://gist.github.com/RubyTuesdayDONO/5006455

pmario:
CM already causes a lot of problems, since users don't know, what to install, if they need a certain functionality.

PR #4113, opened 2019-07-16
Saving/Restoring input selections

BurningTreeC:
This PR creates an `inputManager` instance ($tw.inputManager) which stores and handles `selectionRanges` / `selections` of inputs or textareas that have the `inputManagement="yes"` attribute.

This makes it possible to keep focus in the tiddler-texteditor for example, when clicking the preview button (or changing the preview state by <kbd>alt-P</kbd>

Further, the inputManager sets a tiddlers' text (`$:/temp/current-focus`) to a qualified ID that's unique for each widget. It also sets the `focused-tiddler` field to the name of the current `storyTiddler` (which can be useful for story-river-navigation like discussed in #4782 ) ... but this functionality can also be removed. Both are set whenever an input gets focus


twMat:
@buggyj - Very cool! One case that doesn't work is to remotely edit at macro tiddler. Maybe this case is something else?

BurningTreeC:
@twMat, remotely editing a macro tiddler doesn't work? Can you show me an example?

pmario:
>One case that doesn't work is to remotely edit at macro tiddler. Maybe this case is something else?

Can you tell, how to reproduce the problem. Step by step?

twMat:
@BurningTreeC (sorry, I mistakenly selected @ buggyj previously) 

http://4113issue.tiddlyspot.com/

BurningTreeC:
@twMat - that issue has to do with the sidebar search field and the way how macros are imported in the pagetemplate... if a macro changes, the whole page gets rendered again and the sidebar search gets focus... try this with  the sidebar closed and it works. This issue will not be that easy to fix if at all

twMat:
Interesting

>This issue will not be that easy to fix if at all

Couldn't the editor sense that it's editing a tiddler tagged "macro" and disable the set-focus-on-search? (I'm not saying this PR of yours should do this, it is just a general question - possibly for a next step in this matter.)

BurningTreeC:
>Couldn't the editor sense that it's editing a tiddler tagged "macro" and disable the set-focus-on-search? (I'm not saying this PR of yours should do this, it is just a general question - possibly for a next step in this matter.)

@twMat - I don't see a real solution for that atm

twMat:
Maybe editors could have an default ID which, when the editor gets focus, is added to a state tiddler. Editing a macro could reset focus after refresh?

Jermolene:
Hi @BurningTreeC thanks for this. I'm keen to address it post v5.1.20...

BurningTreeC:
@Jermolene ... 5.1.20 is already a big release anyways, but I'm keen looking forward to 5.1.21

BurningTreeC:
Hi @Jermolene, how do you think about this PR?

BurningTreeC:
ping @Jermolene

BurningTreeC:
bump

BurningTreeC:
@Jermolene, any thoughts about this?

Jermolene:
Apologies @BurningTreeC, I'm thinking that I need to get v5.1.22 out as soon as possible, and so I want to stick to fairly simple stuff for the moment. I'll post a plan soon.

BurningTreeC:
Ok @Jermolene, this one can wait. What about #4458? It's working now with a workaround

Jermolene:
Hi @BurningTreeC did you re-open this on purpose?

BurningTreeC:
Hi @Jermolene - I reopened this because I almost forgot about this and I think it's still worth discussing. But not necessarily to be included in 5.1.23

Jermolene:
No problem, thanks @BurningTreeC

pmario:
@BurningTreeC .. can you upload the demo file here as a zip?

BurningTreeC:
@pmario, I'll create a new Demo, I need just some minutes

BurningTreeC:
@pmario, here's the Demo in a .zip archive

[inputmanager.zip](https://github.com/Jermolene/TiddlyWiki5/files/5471002/inputmanager.zip)

BurningTreeC:
New version, there was a small bug:

[inputmanager.zip](https://github.com/Jermolene/TiddlyWiki5/files/5471010/inputmanager.zip)

BurningTreeC:
@pmario - input-management is only enabled when the `inputManagement` attribute is set to "yes". In the demo it's only the tiddler editor

BurningTreeC:
An up-to-date demo:

[inputmanager.zip](https://github.com/Jermolene/TiddlyWiki5/files/5471357/inputmanager.zip)

Jermolene:
Hi @BurningTreeC thanks. This PR lacks a clear description of what it is intended to do, and how it works. Could you update the OP to reflect the current state of things? It may be worth trying to outline a docs tiddler for the /dev edition that explains the things that a developer writing a widget would need to know?

BurningTreeC:
@Jermolene - I've updated the OP

BurningTreeC:
@pmario - what do you think about this PR?

Jermolene:
Hi @BurningTreeC how does `Widget.prototype.generateTransclusionFootprint()` practically differ from `Widget.prototype.getStateQualifier()`? In particular, does it distinguish between sibling widgets?

BurningTreeC:
Hi @Jermolene,

>how does Widget.prototype.generateTransclusionFootprint() practically differ from Widget.prototype.getStateQualifier()? In particular, does it distinguish between sibling widgets?

the generateTransclusionFootprint method DOES distinguish between sibling widgets. First we get the state qualifier the usual way, then we cycle up to the widget that changes the transclusion variable and for each widget found on that way, we hang together its index among its parentwidgets' children (I hope I explained it well enough, sorry for my english)

Jermolene:
Hi @BurningTreeC 

> the generateTransclusionFootprint method DOES distinguish between sibling widgets. First we get the state qualifier the usual way, then we cycle up to the widget that changes the transclusion variable and for each widget found on that way, we hang together its index among its parentwidgets' children (I hope I explained it well enough, sorry for my english)

That makes sense. It's an approach I tried back when I was first working on the qualified state stuff. The problem I found then was interactions with the way that the list widget rearranges its child widgets dynamically. Anyhow, if it did work then we should deprecate the existing qualified state mechanism and adopt it instead. It's such a big change that it might be worth putting into a separate PR so that we can more easily figure out how to apply it to the existing core code.

BurningTreeC:
>That makes sense. It's an approach I tried back when I was first working on the qualified state stuff. The problem I found then was interactions with the way that the list widget rearranges its child widgets dynamically. Anyhow, if it did work then we should deprecate the existing qualified state mechanism and adopt it instead. It's such a big change that it might be worth putting into a separate PR so that we can more easily figure out how to apply it to the existing core code.

@Jermolene, sounds like a bigger change. If I understand you right, `Widget.prototype.getStateQualifier` should be replaced by this method?

Jermolene:
Hi @BurningTreeC I think that actually to maintain backwards compatibility we'd need to add a flag to `Widget.prototype.getStateQualifier` to choose between the old and new behaviours, and then a new version of the `<<qualify>>` macro that uses the new mechanism, and a new attribute flag for the (embarrassingly undocumented) `<$qualify>` widget. Then we'd look at the core macros that take a state qualifier and figure out how to adapt them.

BurningTreeC:
Ok @Jermolene, now that my work on the keyboard-driven-inputs is done, I need another challenge ;)

I'll give this a try and on the way I can also try to document the `<$qualify>` widget...

If you don't want to do this work :grin:

saqimtiaz:
> The problem I found then was interactions with the way that the list widget rearranges its child widgets dynamically.

@Jermolene @BurningTreeC this was my first thought as well. Are we sure this wont be an issue? Otherwise just opening/closing tiddlers in the story river would change the qualifier for widgets inside a tiddler in the story.

BurningTreeC:
@saqimtiaz 

>Are we sure this wont be an issue? Otherwise just opening/closing tiddlers in the story river would change the qualifier for widgets inside a tiddler in the story.

in my demo posted above, which makes use of this "transclusion-footprint", I don't see an issue. But it's possible that I don't understand the real problem

pmario:
@saqimtiaz You can test this code with the demo.zip above. 

```
Move the "set start" here

<<qualify>>

Move the "set start" here

<$set name=a value=x>

<<qualify>>

<$edit-text tiddler="StoreAppSettings" field="myconfig"/>

</$set>
```

 - edit the tiddler and 
 - open preview mode
 - now play with the position of the `<$set widget`
 - If you select the input the "footprint" will be shown. 
 - If you add some content
 - select the input 
 - change the set start 
 - select the input

You'll see, that the footprint changes. .. The "first" part of the qualifier stays. ... The footprint is completely "random". **That's OK for the purpose it was defined**. But imo it can't be used to "restore" any state.

BurningTreeC:
closing this in favor of #5414

PR #4115, opened 2019-07-17
important fix: store noStateReference in popup.js

BurningTreeC:


BurningTreeC:
... this fix is important for the TagManager's icon-dropdowns

BurningTreeC:
this should get fixed before the 5.1.20 release imho

Jermolene:
Thanks @BurningTreeC

PR #4122, opened 2019-07-22
Fix possible bug

Arlen22:
Checking for outstanding startup modules to execute before the module in question seems to have incorrect code.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -2364,7 +2364,8 @@ $tw.boot.isStartupTaskEligible = function(taskModule) {
 		for(t=0; t<remaining.length; t++) {
 			var task = remaining[t];
 			if(task.before && task.before.indexOf(name) !== -1) {
-				if($tw.boot.doesTaskMatchPlatform(task) || (task.name && $tw.boot.disabledStartupModules.indexOf(name) !== -1)) {
+				//check if the other task matches the platform and is not disabled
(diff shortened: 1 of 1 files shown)


Arlen22:
This is in a section of code that is checking whether any other modules specify that they need to be executed before this module. In that line of reasoning and the apparent intent of the code, it seems that this line should be what I changed it to.

ght (approved):
Good catch! The previous code can't possible be correct, as the same condition (`$tw.boot.disabledStartupModules.indexOf(name) !== -1`) is checked just a few lines earlier and triggers an early return, so it can never be true here.

The fixed version might cause unexpected behavior changes, since now that the check is correct, the code will bail if the task is not in `disabledStartupModules`.

Jermolene:
Hi @ght @Arlen22 it does look as if the code is buggy, and this is the right fix. As you say, there could well be code out there that depends on the current behaviour. We don't currently have tests for the startup module ordering which would have give a bit more confidence at least as far as the core is concerned.

However, I've verified that the core prerelease and tw5.com editions work correctly with no changes to the startup module ordering, so I think that's good enough.

Jermolene:
@FlashSystems found a bug - see #6579. This PR doesn't fix things – the comment is correct, but the code in the PR doesn't match the comment. Fix coming up.

PR #4123, opened 2019-07-22
Add server-sent events to core

Arlen22:
This adds a simple API to use server-sent events. It adds `serverEvents` to the $tw global. The mechanism is very simple. It should work fine with Bob and TiddlyServer, as far as I know. I only tested this in vanilla TiddlyWiki. 

On the server, this is a simple object with two functions used to handle connections and messages. I added `isServer: true` so anything wanting to use it can quickly determine which side of the connection they are on. This is what the TypeScript definition would look like for it. 

```ts
interface serverEvents { 
  broadcastEvent(event: string, data: string): void; 
  isServer: true;
  /** called by the server route to handle incoming requests */
  handleConnection(request, response, state): void; 
  /** internal use only */
  activeConnections: any[];
}
```

On the client, it is an instance of EventSource. 

It also adds the route `GET /status/events` to the server to handle connections. Because of get-tiddler-html it is unpredictable to not have a slash in the middle of the path (so I could not use `/events`). This may need to be addressed separately in get-tiddler-html.

### DIFF
--- a/core/modules/server/routes/get-status-events.js
+++ b/core/modules/server/routes/get-status-events.js
@@ -1,0 +1,36 @@
+/*\
+title: $:/core/modules/server/routes/get-status-events.js
+type: application/javascript
+module-type: route
+
+Adds the server-sent-events server route at /status/events.
+
+\*/
+(function () {
+
+exports.method = "GET";
+
+exports.path = /^\/status\/events$/
+
+exports.handler = handler;
+
+exports.bodyFormat = "stream";
+
+/**
+ * 
+ * @param {import("http").IncomingMessage} request 
+ * @param {import("http").ServerResponse} response 
+ * @param {ServerState} state 
+ */
+function handler(request, response, state) {
+  
+  if (request.headers.accept && request.headers.accept === 'text/event-stream') {
+    $tw.serverEvents.handleConnection(request, response, state);
+  } else {
+    response.writeHead(404, {});
+    response.write(JSON.stringify(request.headers.accept))
+    response.end();
+  }
+}
+
+})();
--- a/core/modules/server/server-events-global.js
+++ b/core/modules/server/server-events-global.js
@@ -1,0 +1,96 @@
+/*\
+title: $:/core/modules/server/server-events-global.js
+type: application/javascript
+module-type: global
+
+Adds serverEvents to $tw. On the server, its public type is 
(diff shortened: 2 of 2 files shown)


Arlen22:
Here is an edit watcher that I wrote to use this mechanism. It also deletes tiddlers from the browser when they are deleted from the server, which the syncer does not do.

Arlen22:
```js
/*\
title: $:/plugins/tiddlywiki/server-sent-events/edit-watcher.js
type: application/javascript
module-type: startup

\*/
(function () {

exports.name = "edit-watcher";
// exports.platforms = ["browser", "node"]; 
exports.after = ["load-modules", "startup"];
exports.synchronous = true;

exports.startup = function () {
  console.log("isServer", $tw.serverEvents.isServer);
  if ($tw.serverEvents.isServer) {
    $tw.wiki.addEventListener("change", function (changes) {
      let res = {};
      Object.keys(changes).forEach(function (k) { res[k] = changes[k].deleted ? "deleted" : "modified" });
      console.log("broadcastEvent server-refresh", res);
      $tw.serverEvents.broadcastEvent('server-refresh', JSON.stringify(res));
    });
  } else {
    $tw.serverEvents.addEventListener("server-refresh",/** @param {MessageEvent} msg */ function serverRefresh(msg) {
      // console.log(msg);
      var changes = JSON.parse(msg.data);
      var filteredChanges = $tw.syncer.filterFn.call(this.wiki, function (callback) {
        Object.keys(changes).forEach(function (title) {
          callback($tw.syncer.wiki.getTiddler(title), title);
        })
      });
      filteredChanges.forEach(function (title) {
        if (changes[title] === "deleted" && $tw.syncer.tiddlerInfo[title]) {
          //This could delete a newly created tiddler with the same name on the server
          //since the syncer will try deleting it from the server again.
          $tw.wiki.deleteTiddler(title);
          // So we delete the sync task before it can be processed. 
          delete $tw.syncer.taskQueue[title];
        }
      });
      $tw.syncer.syncFromServer();
    }
    );
  }
}

})();
```

Arlen22:
I used this during testing and it works good.

Jermolene:
Thanks @Arlen22, this looks promising. I'm going to have to defer looking at it properly until after v5.1.20, and will label it as such.

Arlen22:
I'm closing this in favor of https://github.com/Jermolene/TiddlyWiki5/pull/4609

PR #4127, opened 2019-07-25
navigate to tiddler above or below when cancelling, closing, deleting

BurningTreeC:


Jermolene:
Thanks @BurningTreeC. Is there any chance that users might want to configure whether the target is the next or previous tiddler?

I also wondered whether this logic should be in the associated message handlers in the navigator widget, so that anyone using those messages will get the same behaviour, without needing a custom navigate action.

twMat:
I would really like this addition but/and agree it should be configurable.

BurningTreeC:
>Thanks @BurningTreeC. Is there any chance that users might want to configure whether the target is the next or previous tiddler?

Yes the chance is high

>I also wondered whether this logic should be in the associated message handlers in the navigator widget, so that anyone using those messages will get the same behaviour, without needing a custom navigate action.

I haven't looked at the navigator widget itself, maybe I can try giving it a shot...

BurningTreeC:
>I haven't looked at the navigator widget itself, maybe I can try giving it a shot...

No, the javascript way is too complicated for me at the moment

Jermolene:
> No, the javascript way is too complicated for me at the moment

I can't disagree there; it's overdue for refactoring.

pmario:
new discussion at: Meta-Issue: Refactor the Navigation mechanism #4180

PR #4132, opened 2019-07-30
add a selectable SideBar TableOfContents to the core

BurningTreeC:


core/ui/TableOfContents.tid
```diff
@@ -0,0 +1,14 @@
+title: $:/core/ui/TableOfContents
+tags: $:/tags/SideBar
+caption: {{$:/language/SideBar/Contents/Caption}}
+
+<div class="tc-table-of-contents">
+<$select tiddler="$:/config/TableOfContents/Tag" default="">
```
pmario: IMO there should be a possibility to switch back to the default setting.

Jermolene:
Hi @BurningTreeC this is an interesting case. I think it's relatively advanced to need switchable TOCs, and feel that that the clutter of the select widget is thus a cognitive tax on the majority who will never use it. I try to keep the default sidebar as simple as possible.

There are more complex solutions (e.g. only showing the select widget if there is more than one TOC registered), but they tend to make the TOC tab more complex which in turn makes it harder to understand and modify.

BurningTreeC:
@Jermolene - I see

PR #4133, opened 2019-07-31
update katex plugin to latest v0.10.2

### LINKED ISSUE #3963: Latex (Katex) \neq with offset forward slash
## Summary
When using \neq the forward slash is offset. This happens in the Katex Demo page as well, as shown in the images below:

Katex Demo offset

Katex Demo offset

This does not occur on the Katex website. 

## Steps to reproduce

Use `$$\neq$$` in a tiddler body in a tiddlywiki that has the Katex plugin. Then save it to see the rendered result.

## Expected behavior

For the not equals symbol to be rendered correctly.

## Context

Windows 10, Chrome Version 74.0.3729.169 (Official Build) (64-bit)

BurningTreeC:
Fixes #3963


BurningTreeC:
This PR also fixes #3963

BurningTreeC:
the deletions come from deleting katex.js which isn't used, we're using the minified version

Jermolene:
Hi @BurningTreeC I've been keeping the full katex.js file around in the repo because it makes debugging a bit easier.

It looks like all the font files have changed, is that right?

BurningTreeC:
>It looks like all the font files have changed, is that right?

I've replaced the old ones with the new ones without checking

>I've been keeping the full katex.js file around in the repo because it makes debugging a bit easier.

Ok I can add it again if desired

Jermolene:
Thanks @BurningTreeC that would be great

BurningTreeC:
It's back in @Jermolene

plugins/tiddlywiki/katex/plugin.info
```diff
@@ -3,5 +3,6 @@
 	"description": "KaTeX: mathematical typography",
 	"author": "JeremyRuston",
 	"core-version": ">=5.0.0",
-	"list": "readme usage"
+	"list": "readme usage",
+	"version": "v0.10.2"
```
Jermolene: The version field of the core plugins is set automatically as part of the build process. Perhaps here better to use a new field like "library-version" for the version of the underlying library.

linonetwo:
How to do this, are files in `/contrib` folder needed? 

截屏2022-05-02 09 57 24

Seems there is mhchem renderer too now.

linonetwo:
截屏2022-05-02 09 59 50

Oh, I see, it used to be in /files folder, now move into the /contrib folder.

linonetwo:
How to test this locally, how to start a wiki like https://tiddlywiki.com/plugins/tiddlywiki/katex/  ?

PR #4134, opened 2019-07-31
make plugins with dependencies indent ...

BurningTreeC:
... through `parent-plugin` field

See #4109


BurningTreeC:
I'm going the wrong way here... closing for now

Issue #4131, opened 2019-07-29
get-tiddler-html route conflicts with other routes


Arlen22:
Because there is no sorting done in the route handlers, the get-tiddler-html route conflicts with other routes because the route order cannot be specified and is not predicted before-hand. Currently they are loaded in alphabetical file order which puts it after the core routes but before any other plugin or custom routes. This makes it hard to add routes that do not have a subpath in them (specifically, that have a slash in the middle). Can we change this route to `/static/*` like it used to be in some builds? Or some alternative? Or can we just institute a list filter so we can use the list-before and list-after field or something like that?

Jermolene:
Hi @Arlen22 single tiddler display mode is a somewhat experimental proof of concept at this point, and we should clean it up. I'd prefer reverting to a `/static/` prefix.

Arlen22:
Using the list mechanism for the routing would be highly useful, but that's a separate issue I guess. I'll investigate that part when I have time.

Jermolene:
Hi @Arlen22 you may have noticed that the module mechanism in TW5 doesn't use any features of the tiddler store (e.g. tags), instead just using a single "module-type" field. This is a visible artefact of a subtlety of the module implementation: it is possible to build TW5 with all the JS modules out-of-band (in a simple script block), without exposing any of them as tiddlers (the bootprefix sets up `$tw.modules` so that code run before the main boot file can add modules). That was done right at the beginning so that a future online version of TW5 could harden security by disallowing JS modules as tiddlers. There's a few more things that would need to be done to make that support practical, but I'd be reluctant to do anything to compromise future support for it.

Overall, having to carefully order routes seems brittle and archaic; I'd prefer to avoid overlapping routes entirely.

Arlen22:
Wow, I didn't know that. That's actually a really good idea and should have some very practical implications for TW Cloud. I like the idea as I've also been trying to figure out if it is possible to webpack the Javascript modules or at least make a custom code file that imports them in a webpack-like bundle as Javascript rather than JSON. If I can accomplish that then I can possibly use Closure Compiler or something similar to minify the entire thing.

Arlen22:
So I spent a few hours playing with the idea, and here is the result: https://github.com/Jermolene/TiddlyWiki5/compare/master...Arlen22:compile-modules

You can try it out by running `node tiddlywiki.js editions/server --listen`. I haven't tried the browser side yet. 

[...]

PR #4136, opened 2019-07-31
Use static prefix for html tiddler rendering

Arlen22:
Per https://github.com/Jermolene/TiddlyWiki5/issues/4131 this adds `/static` as the prefix for HTML tiddlers.


pmario:
@Arlen22 -- Is this PR still a thing?

Arlen22:
Apparently, but if I'm the only person requesting it then probably not. Also, 4 years old.

PR #4138, opened 2019-08-01
Update list-links-draggable Macro.tid

AnthonyMuscio:
It took time for me to discover it will not work unless I do provide the current tiddler myself as it is not the default.

This simple update is to the doc's until perhaps this is fixed.

### DIFF
--- a/editions/tw5.com/tiddlers/macros/list-links-draggable Macro.tid
+++ b/editions/tw5.com/tiddlers/macros/list-links-draggable Macro.tid
@@ -10,7 +10,7 @@ The <<.def list-links-draggable>> [[macro|Macros]] renders the ListField of a ti
 !! Parameters
 
 ;tiddler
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/macros/list-links-draggable Macro.tid
```diff
@@ -10,7 +10,7 @@ The <<.def list-links-draggable>> [[macro|Macros]] renders the ListField of a ti
 !! Parameters
 
 ;tiddler
-: The title of the tiddler containing the list
+: The title of the tiddler containing the list, Use `<<currentTiddler>>` if required because it is not a default
```
pmario, line 13: IMO there should be no comma or "Use" should be "use"

AnthonyMuscio:
Closed as duplicate with  Update list-links-draggable Macro.tid #4184

PR #4139, opened 2019-08-01
tag-picker: delete double-definition of tag-actions

BurningTreeC:


BurningTreeC:
this appears to be an artifact

Jermolene:
Thanks @BurningTreeC

PR #4140, opened 2019-08-01
lower margin-bottom of tiddler-frames on narrow screens

BurningTreeC:
this is a proposal for a lower margin-bottom on narrow screens...

the default 28px take a lot of vertical space on smartphones for example

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -877,6 +877,12 @@ button.tc-btn-invisible.tc-remove-tag-button {
 	border: 1px solid <<colour tiddler-border>>;
 }
 
+@media (max-width: {{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}}) {
+	.tc-tiddler-frame {
+		margin-bottom: 10px;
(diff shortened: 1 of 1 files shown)


BurningTreeC:
Before: 

Bildschirmfoto vom 2019-08-01 14-53-00

After: 

Bildschirmfoto vom 2019-08-01 14-52-38

Jermolene:
Thanks @BurningTreeC. There's a lot of Snow White that we might want to change for smaller screens. I've wondered in the past about switching themes at startup according to customisable criteria (the $:/info stuff like screen size).

BurningTreeC:
I was wondering about the screen size, too but then it wouldn't accommodate small sized windows on big screens...

Jermolene:
> I was wondering about the screen size, too but then it wouldn't accommodate small sized windows on big screens...

We could dynamically switch theme based on window size instead of screensize. We'd perhaps need some dampening so that it only reacts when the window size settles down after a resize.

BurningTreeC:
That'd be great

BurningTreeC:
@Jermolene, I don't know how to proceed with this. Should I close this?

PR #4142, opened 2019-08-02
fix tags not animating in viewtemplate

BurningTreeC:
this PR uses the addprefix removesuffix filters, not a new filter as proposed in #3763


Jermolene:
Thanks @BurningTreeC I preferred the version with the "as" operator. I've added a ticket #4147 for discussion.

BurningTreeC:
Hi @Jermolene, I changed it to use the new `then` operator

Jermolene:
Thanks @BurningTreeC

PR #4144, opened 2019-08-02
fix: correct handling of tc-popup-handle when cancelling ...

BurningTreeC:
... popups through focussing inputs


BurningTreeC:
... important for the `$:/Manager` to work correctly

Jermolene:
Thanks @BurningTreeC, funnily enough I bumped into this yesterday.

BurningTreeC:
Thanks @Jermolene

PR #4145, opened 2019-08-02
pop storyview for pagecontrols

BurningTreeC:
this PR adds the pop storyview to the page-control buttons in the sidebar


Jermolene:

PR #4146, opened 2019-08-02
pop storyview for topright and topleft bar

BurningTreeC:
this adds the nice `pop` storyview to the topright and topleft bars


Jermolene:
Thanks @BurningTreeC

PR #4148, opened 2019-08-02
pop storyview for fields list in EditTemplate

BurningTreeC:
this PR makes the fields animate when removing/adding


twMat:
>this PR makes the fields 

this PR makes the mouth of mine smile. Such a fun little detail.

PR #4149, opened 2019-08-02
Pop storyview for tabs macro

BurningTreeC:


Jermolene:
Oooh! Interesting

pmario:
hmm, what does it do?

BurningTreeC:
>hmm, what does it do?

@pmario ... when tabs are added or removed, they animate to show visually where something has changed

Issue #3967, opened 2019-06-03
Implement more of TiddlyWiki's black box logic as customisable action strings


Jermolene:
We have a general problem that a lot of the fundamental behaviour of TiddlyWiki is locked up in "magic" widgets -- navigation, interpretation of the location hash at startup, actions on dropping a file. This makes it hard to customise the logic beyond any attributes exposed by the widgets.

Now that action widgets are so much more powerful than back in 2013 there is an opportunity to rewrite that logic as "action macros" that can be overwritten locally.

For example, the current logic for navigating to a new tiddler is as follows:

1. Add the target tiddler to `$:/StoryList`, with a bunch of custom logic to determine where it should go in the list
2. Add the target tiddler to `$:/HistoryList`, which is stored in a JSON format that can't be addressed via the index mechanism

With a bit of refactoring of the format of the `$:/HistoryList`, we could express that logic as a series of `<$action-listops>` widgets in a global macro called `tx-navigate`. The action of the `<$navigate>` widget would then be to listen for the `tm-navigate` message and execute the action widgets in `tx-navigate` when it receives the message. The `<$linkcatcher>` would assign a new set of action widgets to `tx-navigate` that would assign the target tiddler title to the specified tiddler.

pmario:
Hi Jeremy, I'm absolutely with you!

While I was refactoring the TOC, I did dig relatively deep into the existing navigation mechanism. ... 

We already have the `th-navigating` hook. Which is called in the navigator.js `tm-navigate` event. 

```
/*
Handle a tm-navigate event
*/
NavigatorWidget.prototype.handleNavigateEvent = function(event) {
	event = $tw.hooks.invokeHook("th-navigating",event);
	if(event.navigateTo) {
		this.addToStory(event.navigateTo,event.navigateFromTitle,event.storyTitle);
		if(!event.navigateSuppressNavigation) {
			this.addToHistory(event.navigateTo,event.navigateFromClientRect,event.historyTitle);
		}
	}
	return false;
};
```

I know, that `th-navigating` needs a plugin, but the principle is pretty much the same. 

The problem for me was, that the `event` variable didn't contain enough information, what the navigator should do. eg: `tm-navigate` action could have been created by the linkcatcher widget, in a different "story context". 

I did try to "simulate" the navigation event using the linkcatcher and some action widgets, that directly manipulate $:/StoryList and $:/HistoryList. ... It works! ... BUT

[...]

PR #4156, opened 2019-08-04
make navigator.js and story.js populate focus-tiddler field ...

BurningTreeC:
... of $:/HistoryList with the last navigated tiddler title or the adjacent tiddler title after closing / cancelling a missing tiddler / deleting a tiddler

### DIFF
--- a/core/modules/startup/story.js
+++ b/core/modules/startup/story.js
@@ -157,7 +157,6 @@ function openStartupTiddlers(options) {
 			if(target.indexOf("[[") === 0 && target.substr(-2) === "]]") {
 				target = target.substr(2,target.length - 4);
 			}
-			$tw.wiki.addToHistory(target);
 			$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,target);
 			$tw.wiki.addToHistory(target);
 		} else if(storyList.length > 0) {
--- a/core/modules/startup/story.js
+++ b/core/modules/startup/story.js
@@ -149,6 +149,7 @@ function openStartupTiddlers(options) {
 	}
 	// Save the story list
 	$tw.wiki.addTiddler({title: DEFAULT_STORY_TITLE, text: "", list: storyList},$tw.wiki.getModificationFields());
+	$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,storyList[0]);
 	// Update history
 	if(!options.disableHistory) {
 		// If a target tiddler was specified add it to the history stack
@@ -157,10 +158,8 @@ function openStartupTiddlers(options) {
 			if(target.indexOf("[[") === 0 && target.substr(-2) === "]]") {
 				target = target.substr(2,target.length - 4);
 			}
-			$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,target);
 			$tw.wiki.addToHistory(target);
 		} else if(storyList.length > 0) {
-			$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,storyList[0]);
 			$tw.wiki.addToHistory(storyList[0]);
 		}
 	}
--- a/core/modules/startup/story.js
+++ b/core/modules/startup/story.js
@@ -158,9 +158,12 @@ function openStartupTiddlers(options) {
 				target = target.substr(2,target.length - 4);
 			}
 			$tw.wiki.addToHistory(target);
+			$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,target);
+			$tw.wiki.addToHistory(target);
 		} else if(storyList.length > 0) {
+			$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,storyList[0]);
 			$tw.wiki.addToHistory(storyList[0]);
-		}		
+		}
 	}
 }
 
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -126,6 +126,17 @@ fromPageRect: page coordinates of the origin of the navigation
 */
 NavigatorWidget.prototype.addToHistory = function(title,fromPageRect) {
 	this.wiki.addToHistory(title,fromPageRect,this.historyTitle);
+	this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+	this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,title);
+};
+
+NavigatorWidget.prototype.findAdjacentStoryTiddler = function(title,storyList) {
+	if(this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler") === title) {
+		var titleIndex = storyList.indexOf(title);
+		return storyList[titleIndex + 1] ? storyList[titleIndex + 1] :
+			(storyList[titleIndex - 1] ? storyList[titleIndex - 1] : null);
+	}
+	return false;
 };
 
 /*
@@ -146,15 +157,24 @@ NavigatorWidget.prototype.handleNavigateEvent = function(event) {
 NavigatorWidget.prototype.handleCloseTiddlerEvent = function(event) {
 	var title = event.param || event.tiddlerTitle,
 		storyList = this.getStoryList();
+	var adjacentTiddler = this.historyTitle ? this.findAdjacentStoryTiddler(title,storyList) : null;
 	// Look for tiddlers with this title to close
 	this.removeTitleFromStory(storyList,title);
 	this.saveStoryList(storyList);
+	if(adjacentTiddler) {
+		this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+		this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,adjacentTiddler);
+	}
 	return false;
 };
 
 // Close all tiddlers
 NavigatorWidget.prototype.handleCloseAllTiddlersEvent = function(event) {
 	this.saveStoryList([]);
+	if(this.historyTitle) {
+		this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+		this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,"");
+	}
 	return false;
 };
 
@@ -162,6 +182,10 @@ NavigatorWidget.prototype.handleCloseAllTiddlersEvent = function(event) {
 NavigatorWidget.prototype.handleCloseOtherTiddlersEvent = function(event) {
 	var title = event.param || event.tiddlerTitle;
 	this.saveStoryList([title]);
+	if(this.historyTitle) {
+		this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+		this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,title);
+	}
 	return false;
 };
 
@@ -209,7 +233,8 @@ NavigatorWidget.prototype.handleDeleteTiddlerEvent = function(event) {
 		storyList = this.getStoryList(),
 		originalTitle = tiddler ? tiddler.fields["draft.of"] : "",
 		originalTiddler = originalTitle ? this.wiki.getTiddler(originalTitle) : undefined,
-		confirmationTitle;
+		confirmationTitle,
+	    	adjacentTiddler;
 	if(!tiddler) {
 		return false;
 	}
@@ -230,6 +255,9 @@ NavigatorWidget.prototype.handleDeleteTiddlerEvent = function(event) {
 			))) {
 		return false;
 	}
+	if(this.historyTitle) {
+		adjacentTiddler = this.findAdjacentStoryTiddler(title,storyList);
+	}
 	// Delete the original tiddler
 	if(originalTitle) {
 		if(originalTiddler) {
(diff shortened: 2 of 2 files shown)


BurningTreeC:
We can use this mechanism to create Keyboard Shortcuts that allow navigating the StoryList and performing actions on the "currently viewed tiddler"

We can emphasize that tiddler visually through a controlpanel-option if desired

BurningTreeC:
this PR also stores the "previously viewed tiddler" in the "previous-focus-tiddler" field

BurningTreeC:
What I'm trying to achieve is that there's always a tiddler title stored in the focus-tiddler field that is contained in the StoryList

Therefor, when closing a tiddler, deleting a tiddler or cancelling a missing tiddler it should populate the focus-tiddler field with the adjacent tiddler which is the next tiddler in the list or the previous if the "viewed tiddler" is the last in the StoryList

BurningTreeC:
There's an issue with the startup not setting the focus-tiddler to the first story-tiddler ... I'll need to investigate

BurningTreeC:
@Jermolene - is there a way to get this functionality for 5.1.20?

core/modules/widgets/navigator.js
```diff
@@ -146,22 +158,35 @@ NavigatorWidget.prototype.handleNavigateEvent = function(event) {
...
 	var title = event.param || event.tiddlerTitle,
 		storyList = this.getStoryList();
+	var adjacentTiddler = this.historyTitle ? this.findAdjacentStoryTiddler(title,storyList) : null;
 	// Look for tiddlers with this title to close
 	this.removeTitleFromStory(storyList,title);
 	this.saveStoryList(storyList);
+	if(adjacentTiddler) {
+		this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+		this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,adjacentTiddler);
+	}
```
pmario: I think this should be handled at a lower level. follow the `.addHistory()` function, down too the root. It should be handled there

core/modules/widgets/navigator.js
```diff
@@ -373,6 +410,10 @@ NavigatorWidget.prototype.handleCancelTiddlerEvent = function(event) {
 					this.removeTitleFromStory(storyList,draftTitle);
 				}
 				this.saveStoryList(storyList);
+				if(adjacentTiddler) {
+					this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+					this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,adjacentTiddler);
+				}
```
pmario: these 2 lines are duplicated way too often. IMO that's a sign, that theres something wrong.

BurningTreeC: @pmario, this is refactored now

core/modules/widgets/navigator.js
BurningTreeC: the underlying problem is that addToHistory is not called when closing, deleting or cancelling a tiddler (and the tiddler is removed)

BurningTreeC:
@pmario @Jermolene, is there anything to do from my part to make this core-ready?

core/modules/startup/story.js
```diff
@@ -149,6 +149,7 @@ function openStartupTiddlers(options) {
 	}
 	// Save the story list
 	$tw.wiki.addTiddler({title: DEFAULT_STORY_TITLE, text: "", list: storyList},$tw.wiki.getModificationFields());
+	$tw.wiki.setText(DEFAULT_HISTORY_TITLE,"focus-tiddler",undefined,storyList[0]);
```
pmario: Hardcoding `DEFAULT_HISTORY_TITLE` imo should be avoided. ... We need to use the `tv-history-list` variable for all of those elements.

pmario: Same thing with: `DEFAULT_STORY_TITLE`

core/modules/widgets/navigator.js
```diff
@@ -126,6 +126,22 @@ fromPageRect: page coordinates of the origin of the navigation
 */
 NavigatorWidget.prototype.addToHistory = function(title,fromPageRect) {
 	this.wiki.addToHistory(title,fromPageRect,this.historyTitle);
+	this.setFocusTiddler(title);
+};
+
+NavigatorWidget.prototype.setFocusTiddler = function(title) {
+	this.wiki.setText(this.historyTitle,"previous-focus-tiddler",undefined,this.wiki.getTiddler(this.historyTitle).getFieldString("focus-tiddler"));
+	this.wiki.setText(this.historyTitle,"focus-tiddler",undefined,title);
```
pmario: As I wrote, this should be done in story.js. I do think it belongs there.

core/modules/startup/story.js
pmario: I should be possible to handle several independent stories at the same time

pmario:
I'm OK with the basic functionality, that is provided, but I do think it needs a more generic focus.

core/modules/startup/story.js
BurningTreeC: Hi @pmario, this is in /startup/story.js , there is no other storylist available

pmario: That's not true. The story and history tiddlers are defined as parameters of the navigator widget.

pmario:
Have a look at this PR: https://github.com/Jermolene/TiddlyWiki5/pull/3939 and this video https://github.com/Jermolene/TiddlyWiki5/pull/3939. It shows, what I want to achieve. 

IMO we should combine our efforts, into a generic change that can do it all.

pmario:
updated the video link

BurningTreeC:
@pmario, your video looks good! I'd welcome a PR by you

pmario:
At https://youtu.be/iaevaODdKyA?t=375 it shows what I mean with different stories at the same time.

BurningTreeC:
@pmario - yes I see, I understand that part now

BurningTreeC:
The important part for me is that the currently viewed tiddler is stored somewhere and if I'm seeing it correctly, your work does that

pmario:
>The important part for me is that the currently viewed tiddler is stored somewhere and if I'm seeing it correctly, your work does that

I'm using the same mechanism as you do. The $:/HistorList is the key tiddler. I'm using a new `focused-tiddler` field, that also is manipulated by the "focused-tiddler plugin"

If this plugin would be activated, it would not be needed to handle "delete" and "cancel" event, since it would be done by "focused-tiddler" plugin.

pmario:
The problem with "focused-tiddler" plugin is, that it only works reliably with the "classic" storyview.

eg: the experimental "stacked" storview from Jeremy it would be wrong.

pmario:
So the function [getTopMost()](https://github.com/Jermolene/TiddlyWiki5/pull/3939/files?file-filters%5B%5D=.info&file-filters%5B%5D=.js&file-filters%5B%5D=.tid#diff-8ac664f1394a73a3c0ac36016bde83b3R80) needs to be calculated in the different story-views. eg: https://github.com/Jermolene/TiddlyWiki5/tree/master/core/modules/storyviews

BurningTreeC:
Ah I see, it's also a bit complicated doing it your way

BurningTreeC:
My approach is basically driven by the navigation itself which I thought would be the easiest way

pmario:
Not really complicated, but there are several moving parts involved - and some of them are part of the core :/

pmario:
I also did some changes to the navigator.js and linkcatcher.js which will allow us to handle as many stories as we want.

pmario:
story.js contains the most basic change, which would solve 80% of all "problems"

pmario:
The PR looks complicated, because there is a lot of docs ... 3 core js files and 1 plugin js file contain the "magic" :)

BurningTreeC:
@pmario - maybe you could make different PR's for each single "problem" we have to solve?

pmario:
Yea, you are right. I'll do that.

pmario:
Why do you need: `previous-focus-tiddler` field ... use-case?

BurningTreeC:
>Why do you need: previous-focus-tiddler field ... use-case?

my use-cases are primarily keyboard-driven... going back to previous tiddler or reopening the last closed tiddler ...

pmario:
ahhh, ok

pmario:
I think the first thing we need is to add some options to 

https://github.com/Jermolene/TiddlyWiki5/blob/a986e4f7d60030ece39b6ed5ec3313e64f937d17/core/modules/story.js#L93

So we can save additional fields to the $:/HistorListy using plugins.

BurningTreeC:
that's a good idea @pmario

BurningTreeC:
have you seen the adjacentTiddler mechanism I'm using? That comes near to the behavior of which tiddlers are being viewed after closing one / cancelling a missing one

I think it could simplify what you're doing using a worker

BurningTreeC:
... / deleting one

pmario:
I think this comment is also important, to see what's going on: https://github.com/Jermolene/TiddlyWiki5/pull/3939#discussion_r281022656

and that there is a problem: in `$tw.wiki.addToHistory()`

pmario:
>... I think it could simplify what you're doing using a worker

I think you are right. I'll have a closer look.

pmario:
Also see: https://github.com/Jermolene/TiddlyWiki5/blob/a986e4f7d60030ece39b6ed5ec3313e64f937d17/core/modules/story.js#L103

We may use the TODOs here

BurningTreeC:
those `TBD` methods in story.js aren't called at the moment, am I right? We'd have to refactor quite a bit

pmario:
Yea, but it would be the right level of the API, without having to use "hacks" like `$tw.wiki.addToHistory()`

pmario:
We want to mess with navigator.js, linkcatcher, story.js anyway ... So if we are at it ;)

pmario:
I forgot the different storyviews

BurningTreeC:
>We want to mess with navigator.js, linkcatcher, story.js anyway ... So if we are at it ;)

:grin:

BurningTreeC:
My PR here has the advantage of not messing too much

pmario:
Yea, but it contains "hack-like" code, which imo is a sign, that the core API is missing generic functions/functionality.

pmario:
Eg: Tony wants to improve the "import" mechanism. see: https://github.com/Jermolene/TiddlyWiki5/issues/4170

I want to have "replay ability". ... for eg: step-by-step tutorials. ... 

So if the $:/HistoryList would contain more info, what was going on, it would be much easier to do an "undo last action" keyboard event. 

If the "undo last actions" code is part of the core, we could import a $:/ReplayHistory tiddler and go back and forth ... Just an other idea. .... 

The funny thing is. We could do it all just reusing the core. ... That's the level of improvements, that I want..

BurningTreeC:
>The funny thing is. We could do it all just reusing the core. ... That's the level of improvements, that I want..

Wow, then where to start?

pmario:
I think most of it can be done in story.js adding some more TBDs :) .. First step 

```
story.addToHistory(title,fromPageRect,options);
```

pmario:
I'll create a PR for this one. ... Let's see, if you can use it for your keyboard handling too. ... If yes, we may be on the right way :)

BurningTreeC:
>I'll create a PR for this one. ... Let's see, if you can use it for your keyboard handling too. ... If yes, we may be on the right way :)

Perfect, I'll see where I can help

Jermolene:
Hi @pmario @BurningTreeC I have a couple of concerns that I hope we can address along with this:

* Refactoring navigator.js to be less of a mess by moving more of the logic to $:/core/modules/story.js
* Decent support for single tiddler mode (see my failed PR #3412)
* Laying the groundwork for moving the critical logic over into wikitext land (#3967)

I've not tracked all of your discussion above, you may have touched on these points.

pmario:
There is quite some overlapping between navigator.js and story.js atm.

It's not 100% clear to me, what should be part of story and what should be part of navigator. 

navigator is a widget. ... So it is "end-user" visible. I think it should "abstract" the "complicated" parts away from the users. It should contain "high level" interface.

story.js is / should be accessible for plugin authors, if they need to implement "special" functions available through plugins. ... plugin authors should know, what they are doing. ...  eg: populate the $:/HistoryList with additional info, that's needed to do several steps of undo. ... 

So for me it should be possible to say: "close others" ... then CTRL - Z should re-open all closed tiddlers ... This functions imo can be implemented in the TBDs in story.js

navigator.js on the other hand does a little bit too much atm. ... So the only function that is needed from story.js is .addToHistory() ... Some of those functions may be moved to story.js, to make navigator.js easier to read. ...

Jermolene:
> There is quite some overlapping between navigator.js and story.js atm.

It's actually worse than that because we also have some story/history helpers in wiki.js. The idea was to have a bunch of helper functions in story.js, and then slim navigator.js down to be much more readable.

> It's not 100% clear to me, what should be part of story and what should be part of navigator.

Ideally, the navigator widget should be just one exposure of a box of logic that exists independently and is also used by other clients like wiki.js and hopefully future plugins.

> navigator is a widget. ... So it is "end-user" visible. I think it should "abstract" the "complicated" parts away from the users. It should contain "high level" interface.

> story.js is / should be accessible for plugin authors, if they need to implement "special" functions available through plugins. ... plugin authors should know, what they are doing. ... eg: populate the $:/HistoryList with additional info, that's needed to do several steps of undo. ...

I think the purpose of story.js is to package up the common, meaningful primitive operations on the story and history, particularly the ones that require policy decisions (e.g. choosing where to open a newly navigated tiddler).

> So for me it should be possible to say: "close others" ... then CTRL - Z should re-open all closed tiddlers ... This functions imo can be implemented in the TBDs in story.js

I'd imagine that as part of a generic undo mechanism that steps back through changes to the tiddler store, and so wouldn't know directly about the story.js mechanism.

> navigator.js on the other hand does a little bit too much atm. ... So the only function that is needed from story.js is .addToHistory() ... Some of those functions may be moved to story.js, to make navigator.js easier to read. ...

As above, I'd like to see navigator.js slimmed down: the success criteria is that it should be readable, with the method names from story.js being descriptive enough that one might not need to look at the code.

BurningTreeC:
@Jermolene, @pmario ... I'd create a PR that initializes story.js in the navigator widget and removes the indirections through wiki.js ... we could then discuss there and take it step by step ... what do you think?

BurningTreeC:
see #4178

PR #4157, opened 2019-08-04
add "preventScroll" option to tm-focus-selector message

BurningTreeC:
this delays the focussing for the time of the animation duration if delayed="yes" or delayed="true"

this is useful when navigating the story river up and down with keyboard shortcuts and the shortcuts focus the title input if a navigated tiddler is in edit mode -> navigation doesn't jump but stays smooth

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -44,7 +44,13 @@ exports.startup = function() {
 			console.log("Error in selector: ",selector)
 		}
 		if(element && element.focus) {
-			element.focus();
+			if(event.paramObject.delayed === "yes" || event.paramObject.delayed === "true") {
+				setTimeout(function() {
+					element.focus();
+				},$tw.utils.getAnimationDuration());
+			} else {
+				element.focus();
+			}
 		}
 	});
 	// Install the scroller
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -44,13 +44,7 @@ exports.startup = function() {
 			console.log("Error in selector: ",selector)
 		}
 		if(element && element.focus) {
-			if(event.paramObject.delayed === "yes" || event.paramObject.delayed === "true") {
-				setTimeout(function() {
-					element.focus();
-				},$tw.utils.getAnimationDuration());
-			} else {
-				element.focus();
-			}
+			element.focus(event.paramObject);
 		}
 	});
 	// Install the scroller
--- a/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-focus-selector.tid
+++ b/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-focus-selector.tid
@@ -9,4 +9,4 @@ The `tm-focus-selector` message sets the focus to the DOM element identified by 
 
 |!Name |!Description |
 |param |Selector identifying the DOM element to be focussed  |
-
+|paramObject |Optional hashmap of additional parameters passed to the `focus` command |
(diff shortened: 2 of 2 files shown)


BurningTreeC:
This gif shows how navigating up/down with the keyboard focuses the titles of the navigated tiddlers, but not instantly... it waits for the animation to finish

Peek 2019-08-04 09-44

Jermolene:
Hi @BurningTreeC interesting!

Presumably the root cause is the focus="true" attribute on the title edit text widget in $:/core/ui/EditTemplate/title? The trouble is that that was always a terrible hack; it just throws the problem of controls competing to take the focus at the browser to solve.

So I'd suggest putting this in the pot with the other considerations we've accumulated for the Great Focus Rearchitecture (post v5.1.20!).

BurningTreeC:
@Jermolene , this is something different

the focus is set to the titles through a keyboard shortcut with the tm-focus-selector message

Jermolene:
Hi @BurningTreeC OK then perhaps I don't understand what problem this is trying to fix.

> this is useful when navigating the story river up and down with keyboard shortcuts and the shortcuts focus the title input if a navigated tiddler is in edit mode -> navigation doesn't jump but stays smooth

Can you unpack that sentence so it is easier to parse?

BurningTreeC:
I have some keyboard shortcuts to navigate the story river up and down

Those shortcuts also focus the title of the navigated tiddlers, if they're open for editing

If they get focused instantly the navigation gets interrupted and is jumpy

If we wait for the navigation to finish everything is fine and smooth

BurningTreeC:
@Jermolene ... in the gif everything works fine, it shows a working example

Jermolene:
> If they get focused instantly the navigation gets interrupted and is jumpy

Why does focusing them interrupt the animation? Does it happen on all browsers?

BurningTreeC:
Happens on firefox and chrome

>Why does focusing them interrupt the animation?

Focusing instantly jumps to the focused input field, even if an animation is still ongoing

Jermolene:
> Focusing instantly jumps to the focused input field, even if an animation is still ongoing

The thing that's strange is that if I set the animation speed to something like 8s, and then navigate to open a draft, then I can happily click in the new tiddler to change the focus while it's animating. So I don't think the act of focusing should disturb the animation. I wonder if the animation is actually being paused by, for example, an attempt to read a dimension that requires the DOM to calculate a reflow?

BurningTreeC:
>I wonder if the animation is actually being paused by, for example, an attempt to read a dimension that requires the DOM to calculate a reflow?

Seems to be a browser issue?

Jermolene:
> Seems to be a browser issue?

Perhaps, but the fact that both Firefox and Chrome exhibit the behaviour is interesting. The trouble is that the settimeout fix doesn't tell us a whole lot about the underlying problem because it's quite a brute force fix.

I found this stack overflow answer: https://stackoverflow.com/questions/26782998/why-does-calling-focus-break-my-css-transition

It turns out that the focus method does have a parameter to prevent scrolling:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus

Can you give it a try?

BurningTreeC:
>It turns out that the focus method does have a parameter to prevent scrolling:
>https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
>Can you give it a try?

Thanks @Jermolene, that option works!

BurningTreeC:
@Jermolene - this PR has changed and includes my try

BurningTreeC:
Now one can simply pass `preventScroll="true"` to the action-sendmessage widget with the `tm-focus-selector` message

Jermolene:
Thanks @BurningTreeC please could you update the docs?

BurningTreeC:
done @Jermolene - it's just a single line, I hope it's ok

PR #4160, opened 2019-08-06
Sidebar items should use list widget for visibility

Arlen22:
The sidebar items should use the list widget for visibility and this filter combines both into one list item. The main purpose of this change is to make it easier to change the look of the page with only CSS.


pmario:
@Arlen22 ... Is this PR still valid?

Arlen22:
Probably not.

Issue #4173, opened 2019-08-12
Bug with custom FileSystemPaths in 5.1.20


hoelzro:
Steps to reproduce:

  * Create a new wiki: `tiddlywiki my-wiki --init server && tiddlywiki my-wiki --listen`
  * Visit the wiki in the browser
  * Create a new tiddler: `$:/config/FileSystemPaths` with the contents `[is[system]removeprefix[$:/]addprefix[_system/]]`
  * The tiddler is saved with the filename `tiddlers/Filter error_ Missing [ in filter expression.tid`

I did this with a clone of 5.1.20 after seeing this in my own wiki - I'll keep digging into this.

hoelzro:
I initially suspected 7fcd2f132 - I tried these steps there.  While I didn't see the "Filter error..." filename, `$:/config/FileSystemPaths` was saved as `$__config_FileSystemPaths.tid`.  So it's still not respecting the custom filesystem path filter, but at least it's not the same breakage.  Trying with the parent commit edd315643 preserves the old behavior - it was saved under `_system/config/FileSystemPaths.tid`

hoelzro:
I just submitted a PR that appears to fix the issue - is there a good place I could add a test for this?

PR #4174, opened 2019-08-12
Pass pathFilters as an array

hoelzro:
Otherwise, when we try to iterate over pathFilters in the filesystem
utils module, we end up iterating over each character in the filter
string, which ends up generating 'Filter error_ Missing [ in filter
expression.tid' as the tiddler's filename

Fixes GH #4173

### DIFF
--- a/plugins/tiddlywiki/filesystem/filesystemadaptor.js
+++ b/plugins/tiddlywiki/filesystem/filesystemadaptor.js
@@ -50,16 +50,10 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
 	var title = tiddler.fields.title,
 		fileInfo = $tw.boot.files[title];
 	if(!fileInfo) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @hoelzro that's definitely an oversight. The docs say that the tiddler should be a newline delimited list of filters so perhaps we can fix it like this:

```
		// Otherwise, we'll need to generate it
		fileInfo = $tw.utils.generateTiddlerFileInfo(tiddler,{
			directory: $tw.boot.wikiTiddlersPath,
			pathFilters: this.wiki.getTiddlerText("$:/config/FileSystemPaths","").split("\n"),
			wiki: this.wiki
		});
```

hoelzro:
Oh, great point @Jermolene - I didn't realize that there could be more than one filter!  I'll fix that up right away.

hoelzro:
@Jermolene Alright, I've applied your suggestion - thanks!

hoelzro:
@Jermolene How does the revised change look to you?

Jermolene:
Thanks @hoelzro

PR #4178, opened 2019-08-15
work on navigator.js and story.js

BurningTreeC:
This removes the indirect way to story.js through wiki.js

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -60,6 +60,13 @@ NavigatorWidget.prototype.execute = function() {
 	// Get our parameters
 	this.storyTitle = this.getAttribute("story");
 	this.historyTitle = this.getAttribute("history");
+	if(this.storyTitle) {
+		this.story = new $tw.Story({
+			wiki: this.wiki,
+			storyTitle: this.storyTitle,
+			historyTitle: this.historyTitle
+		});
+	}
 	this.setVariable("tv-story-list",this.storyTitle);
 	this.setVariable("tv-history-list",this.historyTitle);
 	// Construct the child widgets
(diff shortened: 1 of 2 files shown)


pmario:
Hi, ... I think we need to deal with 4 or 5 different elements here. 

`navigator.js`
`story.js` ... `modules/startup/story.js` and `modules/story.js`
`storyviews/classic, pop, zooming.js`
`linkcatcher.js`

pmario:
In the #3939 PR I did include the possibility that navigator.js can handle different stories. ... So there is only 1 navigator widget needed for the whole wiki. 

As Jeremy pointed out, the navigator widget is heavy weight, so he didn't want me to use it to redirect single links. see: https://github.com/Jermolene/TiddlyWiki5/pull/3939/files#r280725257

That's why I used / modified the linkcatcher, so it can do the job, in a much "simpler" way.

pmario:
I think navigator.js is 1 key part for the whole UI. So we are not allowed to "break" it, if we do refactoring. 

It seems [Jeremy is OK](https://github.com/Jermolene/TiddlyWiki5/pull/4156#issuecomment-521433281) with some refactoring. ... BUT we must be backwards compatible. 

So I think we need to create some tests for the existing functions _first_ so we can be sure, to create a "drop in" replacement.

pmario:
@BurningTreeC discussions should be done at: Meta-Issue: Refactor the Navigation mechanism #4180

pmario:
I did create an "proof of concept" interactive browser based test environment, that will let us and our users create automatic  and / or interactive, wikitext based, tests. 

It should be possible to create all kind of interaction based tests. ... I think keyboard-shortcuts should be testable too. 

Link will follow soon.

PR #4181, opened 2019-08-18
make wikify in save-wiki button evaluate on click

BurningTreeC:
this makes the wikify widget evaluate the content of `$:/config/SaveWikiButton/Filename` when the button is clicked

in case there's a `<<now>>` macro used for example


BurningTreeC:
@Jermolene, is this anything to be considered?

PR #4184, opened 2019-08-19
Update list-links-draggable Macro.tid

AnthonyMuscio:
Added this to the tiddler parameter 

(must be provided as does not default to currentTiddler)

Unless this is done it will interact but fail to save the change as expected.


editions/tw5.com/tiddlers/macros/list-links-draggable Macro.tid
```diff
@@ -10,7 +10,7 @@ The <<.def list-links-draggable>> [[macro|Macros]] renders the ListField of a ti
 !! Parameters
 
 ;tiddler
-: The title of the tiddler containing the list
+: The title of the tiddler containing the list (must be provided as doers not default to currentTiddler)
```
Jermolene: Thanks @AnthonyMuscio -- there's a minor typo ("doers" instead of "does")

Issue #4159, opened 2019-08-05
Extend keyboard manager to support cheat codes


Jermolene:
Elsewhere, @twMat has been seeking a way to make his tiddlyspot hosted wiki present a different UI for him versus other, ordinary users (as I understand it, please correct if I'm wrong).

A somewhat traditional solution to this problem might be to add support for [cheat codes](https://en.wikipedia.org/wiki/Cheating_in_video_games#Cheat_codes): a magic key combination of multiple consecutive keypresses that triggers an action that modifies the UI. The tiddler(s) holding the state could be stored in browser storage, meaning that @twMat would only have to repeat the magic key combination once per browser.

So, @BurningTreeC this might be one for you: could the keyboard manager track enough state that it could use sequences of keyboard shortcuts to trigger an action?

BurningTreeC:
@Jermolene , this appears to be a huge rewrite of the mechanism I'm not able to do

twMat:
>a way to make his tiddlyspot hosted wiki present a different UI for him versus other, ordinary users

Yes. Much appreciated that you bring this up, and that's a nice proposed solution which should be of use for anyone who wants to control/trigger *anything* locally for whatever reason.

>this appears to be a huge rewrite of the mechanism I'm not able to do

@Jermolene - do you have any specific thoughts or pointers that would ease understanding for how to approach this?

Jermolene:
> do you have any specific thoughts or pointers that would ease understanding for how to approach this?

The approach I was thinking would require a change to the core keyboard handling. Each shortcut would get an array of keys that trigger it, instead of a single one. The system would keep track of the last X keypresses, and match against that against the array instead of a single keypress.

AnthonyMuscio:
Would an obscure three key combination not be enough. Can that be handled by the current shortcut methods?, I wanted something similar to trigger a save so I could conceal the save mechanism then save the wiki (for read only users).

pmario:
IMO has been implemented with https://tiddlywiki.com/#Alternative%20page%20layouts

Issue #3784, opened 2019-02-20
Feature Request: Trigger Custom Popups via Custom Strings


diego898:
As @twMat described in: https://github.com/Jermolene/TiddlyWiki5/issues/3764#issuecomment-464076426, it would be incredibly useful to trigger _custom popups_ via _custom strings:_

```
to be able to trigger _custom popups_ via _custom strings_ would be so useful that it deserves place in standard distro. Both newbies and experts should benefit from having custom triggers like `[[`, `<$`, etc show pickable options like

* tiddler titles
* widgets/macros - much needed (e.g `<$` triggers picklist of available widget with parameters)
* transclusion popups - to show arbitrary tiddler (... e.g string `wikipedia:foo` shows a tid with an iframed wikipedia for entry "foo". Or `scrap` transcluding transcludes your clippings tid)
```

An example of a plugin that gets us part way there, is Snowgoons Edit-CompText:

https://snowgoon88.github.io/TW5-extendedit/

which allows you to define custom strings to trigger its standard popup:

https://snowgoon88.github.io/TW5-extendedit/#%24%3A%2Fplugins%2Fsnowgoon88%2Fedit-comptext%2Fusage

twMat:
In deed. This would make the editor **support** the very process of authoring and editing! The current editor merely *enables* authoring/editing. "Pick lists"  (as in Snowgoons plugin) is probably the prime examples, i.e to auto complete tiddler titles, macro- and widget calls as well as signatures or stock phrases by selecting options from a presented list. Super userful.

If both the trigger strings and the popup content is customizable, it would turn the editor into a powerful and contextual *toolbox*: Currently, the closest we get is the rich editor toolbar. While useful, it is static: It takes up real estate so it has to be much more limited and cannot take context into consideration. In contrast, very context specific "trigger strings" like `>packlist` or `>synonym:foo` or `>mathops` or `>help` can show exactly the relevant tools or info. (The `>` prefix is just an example)

*Overall it should lower the threshold to using TW because users can get help right away.*

**User friendly UI:**
By having the popup transclude individual tiddlers, it would also be really simple for users to add new ones. For example, tiddlers could be tagged `$:/tags/editortool`, have a field named `editor-trigger` and the tiddler content (i.e text field) would be what shows in the popup. Super simple and very tiddlyesque.

diego898:
[...]

PR #4185, opened 2019-08-19
add action-popup widget ...

BurningTreeC:
... that allows triggering popups via keyboard shortcuts (popups need to be positioned using css)

### DIFF
--- a/core/modules/widgets/action-popup.js
+++ b/core/modules/widgets/action-popup.js
@@ -39,11 +39,11 @@ ActionPopupWidget.prototype.render = function(parent,nextSibling) {
 };
 
 ActionPopupWidget.prototype.triggerPopup = function(event) {
-		$tw.popup.triggerPopup({
-			domNode: this.domNodes[0],
-			title: this.popup,
-			wiki: this.wiki
-		});
+	$tw.popup.triggerPopup({
+		domNode: this.domNodes[0],
+		title: this.popup,
+		wiki: this.wiki
+	});
 };
 
 /*
--- a/core/modules/widgets/action-popup.js
+++ b/core/modules/widgets/action-popup.js
@@ -1,0 +1,78 @@
+/*\
+title: $:/core/modules/widgets/action-popup.js
+type: application/javascript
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC interesting! Is the idea that the action-popup widget would contain the content that is to go into the popup? Does it work properly if used within an action string attribute, rather than inline within the main widget tree?

BurningTreeC:
Hi @Jermolene , the idea is much simpler. It's supposed to trigger a popup tiddler using keyboard shortcuts which gets deleted when clicking somewhere or triggering it again

twMat:
IMO related to #4159

BurningTreeC:
>Does it work properly if used within an action string attribute, rather than inline within the main widget tree?

@Jermolene could you give me an example to understand what you mean?

BurningTreeC:
>Is the idea that the action-popup widget would contain the content that is to go into the popup?

I think the answer is no. There still needs to be a reveal widget somewhere that makes use of the popup state tiddler

Jermolene:
Hi @BurningTreeC is the idea to extract the popup-triggering logic of the button widget to make it a generic action widget? The trouble is that it relies on the action-popup widget rendering a DOM node, does that work if rendering into the fakedom?

BurningTreeC:
@Jermolene, do you have a tip for how to render into the fakedom?

BurningTreeC:
>Hi @BurningTreeC is the idea to extract the popup-triggering logic of the button widget to make it a generic action widget?

Yes that's the idea

twMat:
>extract the popup-triggering logic of the button widget to make it a generic action widget

+1 (sounds like a step that is also in the right direction for #3784 )

Jermolene:
Hi @BurningTreeC apologies for this languishing for so long. I've actually recently needed the same thing in one of my projects. I've slightly extended it to handle passing the popup coordinates. I'll commit it shortly.

BurningTreeC:
Thanks @Jermolene!

PR #4191, opened 2019-08-22
Update Modules.tid

mdemoss:
As far as I can tell nothing runs unless it is a system tiddler.


Jermolene:
Thanks @mdemoss please see my earlier comment on #4192

Issue #4180, opened 2019-08-18
Meta-Issue: Refactor the Navigation mechanism


pmario:
This issue is based on discussions at: 

 - work on navigator.js and story.js #4178
 - make navigator.js and story.js populate focus-tiddler field ... #4156
 - WIP TOC refactor #3939
 - navigate to tiddler above or below when cancelling, closing, deleting #4127
 - ...
 - and probably others which I'll link if I find them :)

pmario:
There is quite some overlapping between navigator.js and story.js atm.

It's actually worse than that because we also have some story/history helpers in wiki.js. The idea was to have a bunch of helper functions in story.js, and then slim navigator.js down to be much more readable.

    It's not 100% clear to me, what should be part of story and what should be part of navigator.

Ideally, the navigator widget should be just one exposure of a box of logic that exists independently and is also used by other clients like wiki.js and hopefully future plugins.

    navigator is a widget. ... So it is "end-user" visible. I think it should "abstract" the "complicated" parts away from the users. It should contain "high level" interface.

    story.js is / should be accessible for plugin authors, if they need to implement "special" functions available through plugins. ... plugin authors should know, what they are doing. ... eg: populate the $:/HistoryList with additional info, that's needed to do several steps of undo. ...

I think the purpose of story.js is to package up the common, meaningful primitive operations on the story and history, particularly the ones that require policy decisions (e.g. choosing where to open a newly navigated tiddler).

    So for me it should be possible to say: "close others" ... then CTRL - Z should re-open all closed tiddlers ... This functions imo can be implemented in the TBDs in story.js

I'd imagine that as part of a generic undo mechanism that steps back through changes to the tiddler store, and so wouldn't know directly about the story.js mechanism.

    navigator.js on the other hand does a little bit too much atm. ... So the only function that is needed from story.js is .addToHistory() ... Some of those functions may be moved to story.js, to make navigator.js easier to read. ...

As above, I'd like to see navigator.js slimmed down: the success criteria is that it should be readable, with the method names from story.js being descriptive enough that one might not need to look at the code.

pmario:
It

pmario:
The new mechanism _must_ be backwards compatible. .. So IMO we need tests

pmario:
https://tiddlywiki.com/#StartupActions may be used for testing in the browser UI

AnthonyMuscio:
Mario,

[...]

PR #4200, opened 2019-08-25
Remove dependency on wiki.js for story-startup and navigator

BurningTreeC:
This removes the indirection through wiki.js to access addToStory and addToHistory in story.js

It's probably the first step to simplify the story-navigator mechanisms


BurningTreeC:
#4180

Jermolene:
I'm happy with this, but I think we should leave the existing methods on the wiki object for backwards compatibility.

pmario:
Why did you close it? ... We can add a deprecated warning into the removed code, instead of removing it. Then after eg 2 versions we can remove it

BurningTreeC:
@pmario - closed it because I thought this wouldn't be of interest

pmario:
I think the existing implementation is a bug, if you have 2 navigator widgets, that work for different stories. ... So imo it has to be fixed anyway. 

I would add a `console.log()` to the "deprecated" code, instead of removing it, since Jeremy wants the function to be still part of the `$tw.wiki` API

eg: `console.log("$tw.wiki.addToHistory() is deprecated! Use the this.story.addToHistory() from the story-object!")` ... same with addToStory.

So if a developer uses the "old" function they should immediately see, that there is something wrong. .. BUT old plugins still work. 

We could check our plugins and those of other important and core plugins, to see, if someone uses this function. ... But that's definitely much more work.

pmario:
may be: `console.log("$tw.wiki.addToHistory() is deprecated since V5.1.23! Use the this.story.addToHistory() from the story-object!")`

pmario:
I checked my plugins. 1 of them actually uses the "old" function. ... So Jeremy is right!

BurningTreeC:
@pmario @Jermolene - now the old code is still in place, I've put it back. Should I add a console log that marks them as deprecated?

pmario:
>... Should I add a console log that marks them as deprecated?

I think there should be a console.log()

pmario:
@Jermolene ... needs to decide it.

BurningTreeC:
@Jermolene do you have any thoughts on this one?

Jermolene:
Hi @BurningTreeC @pmario we have a precedent in `$:/core/modules/filters.js` for displaying deprecation warnings to the console. But we should probably be smarter about it, and not re-display the same message more than once:

```
if(!errorHasBeenDisplayed) {
  errorHasBeenDisplayed = true;
  console.log("error");
}
```

BurningTreeC:
@Jermolene - I'd need advice where to put the console log and how to handle it it's already been displayed

Jermolene:
Hi @BurningTreeC for the moment we can just use console.log in the usual way; if there are issues with the amount of logging then we can address them later.

BurningTreeC:
Ok @Jermolene, I've added the console logs like @pmario suggested in a post above

Jermolene:
Thanks @BurningTreeC

PR #4207, opened 2019-08-26
Fixed 3 typos in the example in line 88

StefanSTFG:
1.     Removed closing square bracket after <currentTiddler>
2.    Changed capital 'P' in addprefix
3.    Added missing slash at the end


Jermolene:
Thanks @StefanSTFG good catch!

PR #4212, opened 2019-08-29
Webserver API "get all tiddlers" does not list system tiddlers

bimlas:
`GET /recipes/default/tiddlers.json` does not return system tiddlers, make this clear in the documentation.


Jermolene:
Thanks @bimlas I'll fix this via a separate commit because there's a typo here.

PR #4220, opened 2019-08-30
add $:/language/DefaultNewTiddlerTags option for predefined tags

BurningTreeC:
With this PR, one can define additional tags through $:/language/DefaultNewTiddlerTags that are added when tiddlers are created using the new-tiddler button and shortcut and the new-here and new-journal-here buttons

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -556,18 +556,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	$tw.utils.each(importData.tiddlers,function(tiddlerFields) {
 		var title = tiddlerFields.title;
 		if(title && importTiddler && importTiddler.fields["selection-" + title] !== "unchecked") {
-			var tiddler;
-			if(moreTags && tiddlerFields.tags) {
-				tiddler = new $tw.Tiddler(tiddlerFields,{
-					tags: tiddlerFields.tags + " " + moreTags
-				});
-			} else if (moreTags) {
-				tiddler = new $tw.Tiddler(tiddlerFields,{
-					tags: moreTags
-				});
-			} else {
-				tiddler = new $tw.Tiddler(tiddlerFields);
-			}
+			var tiddler = new $tw.Tiddler(tiddlerFields);
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
 			self.wiki.addTiddler(tiddler);
 			importReport.push("# [[" + tiddlerFields.title + "]]");
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -556,9 +556,18 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	$tw.utils.each(importData.tiddlers,function(tiddlerFields) {
 		var title = tiddlerFields.title;
 		if(title && importTiddler && importTiddler.fields["selection-" + title] !== "unchecked") {
-			var tiddler = new $tw.Tiddler(tiddlerFields,{
-				tags: tiddlerFields.tags + " " + moreTags
-			});
+			var tiddler;
+			if(moreTags && tiddlerFields.tags) {
+				tiddler = new $tw.Tiddler(tiddlerFields,{
+					tags: tiddlerFields.tags + " " + moreTags
+				});
+			} else if (moreTags) {
+				tiddler = new $tw.Tiddler(tiddlerFields,{
+					tags: moreTags
+				});
+			} else {
+				tiddler = new $tw.Tiddler(tiddlerFields);
+			}
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
 			self.wiki.addTiddler(tiddler);
 			importReport.push("# [[" + tiddlerFields.title + "]]");
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -549,13 +549,16 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	var self = this,
 		importTiddler = this.wiki.getTiddler(event.param),
 		importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
-		importReport = [];
+		importReport = [],
+	    	moreTags = this.wiki.getTiddlerText("$:/language/DefaultNewTiddlerTags");
 	// Add the tiddlers to the store
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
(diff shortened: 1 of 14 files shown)


pmario:
@BurningTreeC .. You should add a little bit more text into the description field.

BurningTreeC:
@pmario 
With this PR, one can define additional tags through `$:/config/NewTiddler/Tags` that are added when tiddlers are created using the new-tiddler button and shortcut and the `new-here` and `new-journal-here` buttons

Jermolene:
Hi @BurningTreeC the motivation behind hidden settings is to avoid cluttering up both the control panel and the translatable strings with lots of settings that are rarely used. In this case, I think setting the tags on a new tiddler is actually a very common requirement, and it would be worth giving this a user interface.

To @pmario's point, it is much better if the top comment of PRs is a complete self-contained description of what is being changed.

BurningTreeC:
@Jermolene, sorry for the missing description. I'll add a controlpanel interface

BurningTreeC:
@Jermolene @pmario I'm done here

core/ui/Actions/new-journal.tid
```diff
@@ -2,10 +2,13 @@ title: $:/core/ui/Actions/new-journal
...
 
-<$vars journalTitleTemplate={{$:/config/NewJournal/Title}} journalTags={{$:/config/NewJournal/Tags}} journalText={{$:/config/NewJournal/Text}}>
+\define get-tags()
+$(journalTags)$ $(moreTags)$
+\end
+<$vars journalTitleTemplate={{$:/config/NewJournal/Title}} journalTags={{$:/config/NewJournal/Tags}} moreTags={{$:/language/DefaultNewTiddlerTags}} journalText={{$:/config/NewJournal/Text}}>
 <$wikify name="journalTitle" text="""<$macrocall $name="now" format=<<journalTitleTemplate>>/>""">
 <$reveal type="nomatch" state=<<journalTitle>> text="">
-<$action-sendmessage $message="tm-new-tiddler" title=<<journalTitle>> tags=<<journalTags>> text={{{ [<journalTitle>get[]] }}}/>
+<$action-sendmessage $message="tm-new-tiddler" title=<<journalTitle>> tags=<<get-tags>> text={{{ [<journalTitle>get[]] }}}/>
```
Jermolene: Rather than using a macro with text subtitution, it might be cleaner to do:

```
tags={{{ [enlist<journalTags>] [enlist<moreTags>] }}}
```

(Not tested!)

core/ui/ViewToolbar/new-here.tid
```diff
@@ -4,9 +4,14 @@ caption: {{$:/core/images/new-here-button}} {{$:/language/Buttons/NewHere/Captio
...
 
 \whitespace trim
+\define get-tags()
+$(tags)$ $(moreTags)$
+\end
 \define newHereActions()
 <$set name="tags" filter="[<currentTiddler>]">
-<$action-sendmessage $message="tm-new-tiddler" tags=<<tags>>/>
+<$set name="moreTags" value={{$:/language/DefaultNewTiddlerTags}}>
+<$action-sendmessage $message="tm-new-tiddler" tags=<<get-tags>>/>
```
Jermolene: Same comment as above.

core/ui/ViewToolbar/new-journal-here.tid
```diff
@@ -5,7 +5,7 @@ description: {{$:/language/Buttons/NewJournalHere/Hint}}
 
 \whitespace trim
 \define journalButtonTags()
-[[$(currentTiddlerTag)$]] $(journalTags)$
+[[$(currentTiddlerTag)$]] $(journalTags)$ $(moreTags)$
```
Jermolene: It would be great to get rid of this macro, too. Albeit a bit more complicated than the instances above!

core/ui/Actions/new-journal.tid
BurningTreeC: @Jermolene, it doesn't work your way, it just picks the first tag

Jermolene: Gosh perhaps we need a "listify" operator to call `$tw.utils.stringifyList()`...

BurningTreeC: >Gosh perhaps we need a "listify" operator to call $tw.utils.stringifyList()...

Such a filter would be useful

Jermolene: I'm away from my main computer until tomorrow but will look at it then if you don't get a chance first...

BurningTreeC: @Jermolene I'll let you go first

BurningTreeC:
@Jermolene, do you know a way to add a tag defined in $:/language/NewTiddlerTags to tiddlers that are being imported?

core/modules/widgets/navigator.js
```diff
@@ -549,13 +549,25 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
...
+				tiddler = new $tw.Tiddler(tiddlerFields,{
+					tags: tiddlerFields.tags + " " + moreTags
+				});
+			} else if (moreTags) {
+				tiddler = new $tw.Tiddler(tiddlerFields,{
+					tags: moreTags
+				});
+			} else {
+				tiddler = new $tw.Tiddler(tiddlerFields);
+			}
```
pmario: I do have strong concerns about these changes!  IMO the import mechanism should _NOT_ modify content in an invisible and "aggressive" way.

pmario: For me this is an absolute no-go. ... IFF there is a possibility to automatically add tags, it has to be an OPT-IN and switched off by default. 

I also don't see a reason, why an option to add tags with "new - something" buttons needs to add tags for hundreds of imported tiddlers. That' doesn't make sense, since it makes tags useless.

BurningTreeC: @pmario - that are good points! I'll revert the changes for the import mechanism

core/modules/widgets/navigator.js
```diff
@@ -549,13 +549,25 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
...
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
```
pmario: Your "old" code was inserted just above this line. So I think it would be a perfect fit for a hook - plugin.

Jermolene:
Hi @BurningTreeC @pmario I'm afraid this needs to wait until after the v5.1.21 bug fix release. I think it's premature to retrofit this onto the import mechanism because I'd like to do a deep revision as discussed in #3967.

core/ui/ViewToolbar/new-here.tid
BurningTreeC: @Jermolene , how do we proceed here?

PR #4227, opened 2019-09-01
fix whitespace around field inputs

BurningTreeC:
this PR (re)-adds some whitespace around the field-inputs


BurningTreeC:
From: 
Unbenannt1

To:
Unbenannt

Jermolene:
Thanks @BurningTreeC

twMat:
This doesn't look quite right IMO

image

PR #4243, opened 2019-09-06
Fix gitlab saver.

Janno:
This fixes the HTTP request sent to gitlab that is meant to see if the
target file already exists. It did not follow the official gitlab v4 api
documentation. That documentation dictates both `path` and `branch` to
be passed via corresponding GET parameters.

Disclaimer: You probably shouldn't merge this as I am sure this is not the right way to construct this URL. It can't really be safe this way, I think. I just didn't want to keep people waiting as other's probably are running into the same issue. Feel free to edit this PR—GitHub tells me that TiddlyWiki maintainers should have the ability to add commits to this PR.

### DIFF
--- a/core/modules/savers/gitlab.js
+++ b/core/modules/savers/gitlab.js
@@ -48,12 +48,9 @@ GitLabSaver.prototype.save = function(text,method,callback) {
 	var uri = endpoint + "/projects/" + encodeURIComponent(repo) + "/repository/";
 	// Perform a get request to get the details (inc shas) of files in the same path as our file
 	$tw.utils.httpRequest({
-		url: uri + "tree/" + encodeURIComponent(path.replace(/^\/+|\/$/g, '')),
+		url: uri + "tree/?path=" + encodeURIComponent(path.replace(/^\/+|\/$/g, '')) + "&branch=" + encodeURIComponent(branch.replace(/^\/+|\/$/g, '')),
 		type: "GET",
 		headers: headers,
-		data: {
-			ref: branch
-		},
 		callback: function(err,getResponseDataJson,xhr) {
 			var getResponseData,sha = "";
 			if(err && xhr.status !== 404) {


Jermolene:
Hi @Janno great, many thanks. I'm sorry for the bureaucracy, but please could you sign the CLA in a separate PR targetting the tiddlywiki-com branch (see instructions [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

Janno:
Oh, no need to apologize. I should have read the contributing.md file before opening the PR.

See #4244

core/modules/savers/gitlab.js
```diff
@@ -48,12 +48,9 @@ GitLabSaver.prototype.save = function(text,method,callback) {
 	var uri = endpoint + "/projects/" + encodeURIComponent(repo) + "/repository/";
 	// Perform a get request to get the details (inc shas) of files in the same path as our file
 	$tw.utils.httpRequest({
-		url: uri + "tree/" + encodeURIComponent(path.replace(/^\/+|\/$/g, '')),
+		url: uri + "tree/?path=" + encodeURIComponent(path.replace(/^\/+|\/$/g, '')) + "&branch=" + encodeURIComponent(branch.replace(/^\/+|\/$/g, '')),
```
Jermolene: We would generally use the `data` property to pass the `path` and `branch` query strings to be assembled by `$tw.utils.httpRequest` (as the previous code did for the `branch` parameter). Is there a reason not to use that approach here?

bimlas:
I tried it but it gave me the same result: Base64 encoded HTML file. @Janno, could you give me a link to the documentation where you found this format? Originally I used this documentation and the saver worked earlier (as I know): https://docs.gitlab.com/ee/api/repository_files.html#create-new-file-in-repository

bimlas:
There is an option in the [API](https://docs.gitlab.com/ee/api/repository_files.html#create-new-file-in-repository):

```
encoding (optional) - Change encoding to ‘base64’. Default is text.
```

I've tried these:

* send with `encoding: "base64"`
* send `content` without encoding
* send `content` without coding but set `encoding: "base64"`

In all cases, the HTML file was encoded. When I decoded with https://www.base64decode.org/ it worked, I was able to open the working wiki.

Jermolene:
Hi @bimlas can you try with "encoding: utf8" and "encoding: text"?

bimlas:
... because I used the global `tiddlywiki` instead of `./tiddlywiki.js`. `(-_-')`

Anyway, Janno's fix in my case still resulted in a Base64 encoded HTML file, however the above-mentioned `encoding` option fixed the bug.

Opened #4255

Jermolene:
Hi @bimlas I think @Janno's fix is for a different problem.

bimlas:
Sorry, you're right, I'm really crazy today. :smile: :cry:

core/modules/savers/gitlab.js
bimlas: I tried to define the `branch` and` path` parameters in `data`, but it doesn't want to work, only @Janno 's modification works properly: it's a `GET` request, thus you cannot send `data` with it. Although [this bug is present in the GitHub saver](https://github.com/Jermolene/TiddlyWiki5/blob/3732c83ad476fbf8d2c0536da715ed9df456517a/core/modules/savers/github.js#L50-L56) as well, it only contains the `branch` parameter, so we haven't noticed yet (probably everyone uses `master`).

bimlas:
I rested for a bit and now I can see what this request is all about and tried it out. These parameters are used in the [documentation](https://docs.gitlab.com/ee/api/repositories.html#list-repository-tree) as well, but in the current GitLab saver, they are defined incorrectly in the `data`: the GET method cannot send data but only via the URL. The same [bug is also present in the GitHub saver](https://github.com/Jermolene/TiddlyWiki5/blob/3732c83ad476fbf8d2c0536da715ed9df456517a/core/modules/savers/github.js#L50-L56).

I apologize @Janno, you did well!

bimlas:
It's an important PR: without this **you can save a wiki only once** with the same filename; if the file already exists on GitLab, it will drop an error and does not creating the commit!

Jermolene:
Hi @Janno @bimlas apologies for missing this from v5.1.21, I hadn't appreciated this was a showstopper.

Jermolene:
There'll be a new prerelease build in 10 minutes that incorporates this fix for testing: https://tiddlywiki.com/prerelease

bimlas:
Seems to work for me, GitLab is OK, but for safety's sake I would love to have someone else test it. :)

I have tried the **GitHub** saver to see if it is OK to save the wiki multiple times with the same filename: When I first saved the file, everything went fine. When I want to overwrite it, I have to wait a lot of time (minutes) and haven't even saved it, dirty indicator is still red. I'm sorry, but I don't have time right now to find the bug, but @Jermolene, could you please test if it works in the same way for you?

Jermolene:
Thanks @bimlas I won't be able to work on it for a few days, perhaps @Janno can do some testing?

farrago:
How can I help test this and get this into a build?

Gitlab has recently announced [Access Control for Pages](https://docs.gitlab.com/ee/user/project/pages/pages_access_control.html) which is an absolutely killer app together with tiddywiki and the gitlab saver; making it incredibly simple to setup a secure personal wiki, always available online and with history via git.

Only blocker is this bug where you can only save the first time and not subsequent times.  So I'd love to help where I can to get an official release out with this resolved. Just let me know how I can help, thanks!

hvisage:
Any ETA on the "release" of .22 with this fix, or how to incorporate it into 5.1.21 to have the plugins/etc. still available?

PS: Something I noticed that isn't mentioned (at least I didn't yet read it) is that this bug is only present when saving with a directory, but doesn't seem to manifest when saving to the root directory on GitLab.

PR #4245, opened 2019-09-07
cosmetics: no text-shadow for tags in sidebar

BurningTreeC:


Jermolene:
Thanks @BurningTreeC I think this makes sense.

BurningTreeC:
Thanks @Jermolene, I think so, too

pmario:
hmm. Wouldn't it make more sense to remove the CSS setting from `.tc-sidebar-header` which defines it.  This element is very high up in the tree. 

So I think text-shadow should be assigned only at elements that need it, instead of removing it from all other elements.

pmario:
eg: .tc-site-title ... I think that's the only one, where it makes sense.

Jermolene:
Hi @pmario I wondered about that. The rationale for the text shadow being high up in the tree is that it is intended to apply to all of the text in the sidebar. It's a way of getting more readable contrast for the text while using a mid-grey background; it was a common technique back in 2012/13 when Vanilla and Snow White were designed. I still find that I prefer it over removing the shadow.

pmario:
My personal preferences are different. For me the shadow makes the text less readable. The only use-case I see is, if a user "accidentally" defines the same "text and background" colour. They can still see something in the sidebar if a shadow is active.

I'm OK with the setting as it is.  

I do disable "anti aliasing" with the graphic cards settings, so the changes are system wide. I do like "crisp" edges.

Issue #4248, opened 2019-09-07
...So I may have made StringifyList worse


flibbles:
Remember when we had that long back and forth about making StringifyList wrap titles like `-myTiddler` in brackets?

Well... Take a guess what this outputs:
```
<$set name=X filter="[[3]subtract[16]]">

{{{[<X>add[2]]}}}
</$set>
```
Did you guess -11? That's what I would have guessed. It's actually 2. That's because of my fun little change wraps -13 like [[-13]] before setting it equal to X. and since [[-13]] isn't a number, the add operator defaults to 0.

I uh... humbly suggest we may want to roll back my change to stringifyList.

-Flibbles

flibbles:
I created a pull request that reverts the change, and includes a new test so we can make sure this atrocity never happens again.

Jermolene:
Hi @flibbles 

Actually, I think the problem is that the variable X contains a title list (because it's using the set widget with the filter attribute). That means that you have to bring it into the subsequent filter with the `enlist` operator:

```
<$set name=X filter="[[3]subtract[16]]">
<$text text=<<X>>/>
<br/>
{{{[enlist<X>add[2]]}}}
</$set>
```

flibbles:
So you're saying this is expected behavior? Without the modification I proposed a while ago, the negative numbers don't get wrapped, and thus can be used directly even if they're set using a filter attribute.

If you think this is intended behavior, then go ahead and close this issue and the pull request I made, but is it possible that wrapping negative numbers like stringifyList now does may confuse some less experienced users?

Jermolene:
Hi @flibbles 

> So you're saying this is expected behavior? 

It's definitely a bit surprising, and highlights that it wasn't an entirely backwards compatible change. However, the old behaviour was in conflict with the principle that a list should also be a valid filter, which I've generally thought to be a useful property of the system.

> If you think this is intended behavior, then go ahead and close this issue and the pull request I made, but is it possible that wrapping negative numbers like stringifyList now does may confuse some less experienced users?

It's a difficult one. We certainly have a good opportunity to revert the change with the imminent release of v5.1.21. On balance, I think backwards compatibility wins, and we should retract the change. I'll merge your PR momentarily.

flibbles:
Okey dokey.

PR #4249, opened 2019-09-07
Reverting changes to stringifyList

flibbles:
I (Flibbles) changed it so that lists generated by stringifyList
would always be compatible with a filter parser, but since lists
are not, and never will be, a subset of filters, there isn't a
point.

More importantly, wrapping negative numbers like "-7" in brackets
would mess up some math stuff.

This relates to issue: #4248


Jermolene:
Thanks @flibbles

PR #4255, opened 2019-09-10
Fix GitLab saver Base64 encoding

bimlas:
Without this fix, the commit will have a Base64 encoded file instead of HTML.

Because the GitLab API can interpret the request without Base64 encoding, and because the encoded data is larger (meaning more data traffic and time to send over the internet), we should not use Base64 encoding, thus I changed the code to send the data as plain text.

Empty edition without encoding: 2.19MB
Base64 encoded empty edition: 2.91MB


Jermolene:
Thanks @bimlas

PR #4259, opened 2019-09-11
Dynamic loading/unloading of plugins

Jermolene:
This PR allows plugins that do not contain JavaScript modules to be dynamically loaded/unloaded as they are created, modified, or deleted. Plugins containing JS modules will still require the wiki to be reloaded.

The initial commit includes the basic functionality but does not yet detect and warn correctly for plugins containing JS modules.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1237,15 +1237,39 @@ $tw.Wiki = function(options) {
 		return null;
 	};
 
-	// Read plugin info for all plugins
-	this.readPluginInfo = function() {
-		for(var title in tiddlers) {
-			var tiddler = tiddlers[title];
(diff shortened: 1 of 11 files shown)


Jermolene:
The recent commits add detection of plugins containing JS modules. When they are created/modified/deleted, their shadow tiddlers are not processed and a warning message is displayed.

Jermolene:
With the latest commits, there is an indication in the control panel plugin chooser of whether a reload will be required:

image

Jermolene:
The plugin name field is now displayed (or the portion of the title after the last slash if there is no name field).

image

Jermolene:
I think is now ready to merge; there are further UI improvements possible but the basic functionality works and now needs wider testing.

PR #4268, opened 2019-09-21
Scale embedded videos and audios

donmor:
Modified files:
$:/core/modules/parsers/audioparser.js
$:/core/modules/parsers/videoparser.js
Added an attribute to each to make imported videos and audios scaled to fit the parent container:
style: {type: "string", value: "width: 100%; height: 100%; object-fit: contain"}


donmor:
Oops I forgot I have already had a pr before XD

PR #3943, opened 2019-05-14
Scale embedded videos and audios

donmor:
Modified files:
$:/core/modules/parsers/audioparser.js
$:/core/modules/parsers/videoparser.js
Added an attribute to each to make video imported scaled to fit the parent container:
style: {type: "string", value: "width: 100%; object-fit: contain"}

Before:
2019-05-14 12-41-01屏幕截图
2019-05-14 13-06-33屏幕截图

After:
2019-05-14 12-40-15屏幕截图
2019-05-14 13-06-39屏幕截图

PR #4276, opened 2019-09-27
Update andtidwiki.js (The saver for Android apps including AndTidWiki, Tiddloid and Tiddloid Lite)

donmor:
The priority of this saver is 1800 (higher than andtidwiki.js), and will be loaded before andtidwiki.js when loaded in Tiddloid or Tiddloid Lite, which supports both of the two scripts. The new saver contains less code and supports "download" method to support exporting feature.

BTW I'm glad to see if another PR I created (#3943 ) merged into the next version


Jermolene:
Hi @donmor could this be done via an update to andtidwiki.js, rather than adding a whole new saver? It looks like at the moment nobody would be able to use the andtidwiki saver if the tiddloid.js saver were loaded at the same time because they both use the same logic for "canSave". Something along the lines of:

```
if(window.twi.saveWiki) {
	window.twi.saveWiki(text);
} else if(window.twi.saveFile) {
	window.twi.saveWiki(pathname,text);
}
```

core/modules/savers/tiddloid.js
```diff
@@ -0,0 +1,46 @@
...
+
+/*jslint node: true, browser: true */
+/*global $tw: false, netscape: false, Components: false */
+"use strict";
+
+var Tiddloid = function(wiki) {
+};
+
+Tiddloid.prototype.save = function(text,method,callback) {
+  if (method === "download") window.twi.saveDownload(text);
```
Jermolene: Note that our coding standard is to always include braces in an if statement.

donmor:
I have just updated andtidwiki.js and removed tiddloid.js.

Jermolene:
Thanks @donmor does the old AndTidWiki have the saveFile and downloadFile methods? If not, I think this will break things for people using those old version of AndTidWiki.

donmor:
Only saveFile found. If saveWiki not exists, the former code would be called. Have been tested before.

On Tiddloid:
- download->saveDownload(text)
- else->

On AndTidWiki::
- download->nothing
- else->saveFile(path,text)

I think it is necessary to change to the new apps, since scoped storage prevents apps from accessing the file system directly. Unfortunately Tiddloid is also affected, so I'm making the new app Tiddloid Lite, which based on Storage Access Framework, to work on Android Q and above. For AndTidWiki and Tiddloid, users have to use ADB to make them continue to work.

donmor:
I just fixed a bug that accidently found.
BTW, will this or #3943 be merged to the next version? I think it is convenient to have these two new features.

Jermolene:
Hi @donmor apologies for the delay. Have you been able to test that the saver still works with AndTidWiki? I'm pretty sure that it's still quite widely used by people on old Android devices, so I would want to make certain that it still works.

donmor:
I just tested the script again with andtidwiki.  The wiki was saved normally and the only thing different is that ATW no longer calls a browser to open a blob url(obviously useless), instead it calls with a data url.BTW on Android version lower than 4.4 both file with new saver and original are popping up JavaScript errors on exporting tiddlers, and ATW installed on some roms won't call the browsers(for example Android studio avd, while my phone calling the default browser).

donmor:
Hi @Jermolene ,
I just found another PR in my fridge :)
If these changes are helpful, please consider merging this to the next version.
I'm glad to see my contributions in version 5.1.2.2 :)
Best regards,
donmor

donmor:
Hello? Anybody here?
My PR is still in the fridge :P

Jermolene:
Hi @donmor thanks, and apologies for the delay. Usually one nudge is enough to remind me, sorry it took so long in this case.

PR #4279, opened 2019-10-01
add a tree-compact macro to the existing tree macro fixes #4260

### LINKED ISSUE #4260: Proposal: "enhance" tree macro
The tree macro is a very useful tool that helps to understand the code. I have some frustrations with it and I would like to propose some enhancement. Since macros and tools usage is very personal I wished I had some advice about those : 

* For now, when a _folder_ is also a _tiddler_, it will show two times in the tree. In a more "tiddler first" way the **folder could directly link to the tiddler when needed**. Tree will be shorter and clearer.
* For some trees, exploring might be pretty boring since you have to open each folder to go further. **Adding an option to _show all_ or _hide all_** as asked in #3792 may fix this.
* Since it's difficult to remember if you have to put the `separator` or not at the end of the `prefix`, the macro could be caring of this
* Some **sorting capacities with a sort parameter**, could also be useful (since a bit limited)

Here is a mock-up for some of the functionnalities

image

sycom:
* the tree-compact displays folders that are also tiddler only once,
* we expand tree folders by clicking on the icon since the name will send you to the tiddler
* to accomplish this also added three "hidden" macros : _branch-link_, _branch-node-compact_ and _tree-node-compact_
* also updated **tree Macro** documentation (and added french translation) and tree Macro examples

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -5,6 +5,12 @@ tags: $:/tags/Macro
 <$link to=<<__full-title__>>><$text text=<<__chunk__>>/></$link>
 \end
 
+\define branch-link(full-title,chunk,separator:"/")
+<$set name="short-title" value={{{ [<__chunk__>removesuffix<__separator__>] }}}>
+<$list filter="[<__full-title__>is[missing]]"><$text text=<<short-title>> /></$list><$list filter="[<__full-title__>!is[missing]]"><$link to=<<__full-title__>> ><<short-title>></$link></$list>
+</$set>
+\end
+
 \define leaf-node(prefix,chunk)
 <li>
 <$list filter="[<__prefix__>addsuffix<__chunk__>is[shadow]] [<__prefix__>addsuffix<__chunk__>is[tiddler]]" variable="full-title">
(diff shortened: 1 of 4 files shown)


Jermolene:
Hi @sycom many thanks for this. I was hoping that the implementation would be able to share a good deal of code with the existing macros, but it looks like it only shares the existing "leaf-node" macro. I guess that's understandable, but it does mean that this PR is quite large overall. Do you think there's any scope to refactor things so that more code can be shared?

sycom:
Hello @Jermolene , thanks for your feedback. I think we can make it thinner, but with some consequences. My analysis of this (based on my somewhat limited knowledge of TiddlyWiki). I gave an estimation of gained weight for each option. As a reference, the `$:/core/macros/tree` text field is around 2250 char :

1. I didn't manage to merge _leaf-link_ and _branch-link_. Better said I managed but the result with qualifying lists was not more compact and way more complicated. So I decided to go back and have two sub-macros
2. It's possible to delete _branch-node_ and rename _branch-node-compact_ to _branch-node_. But that way, folder that are also tiddler will appear as "folder icon" + "link to tiddler" whether you are using tree-compact. Maybe some people would dislike this. I do not but it **does** change the tree macro and I didn't find a simple way to preserve it. (-550 char)
3. _tree-node_ and _tree-node-compact_ have very different behaviour since second one is testing if the node is a tiddler-folder or a folder only. I don't see any way to merge them.
4. _tree-compact_ is a bit long since I added a functionality that render same result with `$:/` and `$:`.  It's friendly for users, but we can dismiss this since the cost is heavy the user can double check and avoid problems (so sad because I was pretty proud of this ;-) and divide it by two. (-900 char)
5. As my reference was the _toc_ macro and its variations, I didn't explore this, but we may merge _tree_ and _tree-compact_ and pass `-compact` as a parameter : `<<tree prefix separator -compact>>` instead of `<<tree-compact prefix separator>>`. Doing so we will have to define a var before launching the first node so _tree_ will be a bit heavier (-975 char)

I have no good solution for (1) and (3). So now possibilities are : 

A. crunch all we can : do (2) + (5) will remove all _branch-node_ and _tree-compact_ lines
B. preserve _tree_ macro as is : do only (4) or (5) will remove all _tree-compact_ lines
C. preserve the _toc_ macro analogy : do (2) + (4) will remove all _branch-node_ lines and divide _tree-compact_ by two

There are other combinations, but those are the most obvious. I would vote for (C), but tell me what you think, and I will edit the PR.

PR #4285, opened 2019-10-03
Add autoparagraph variable (fresh PR, fixed commit history)

Arlen22:
I messed up the commit history on the other PR, so here's a new PR for the dubious autoparagraph variable which completely turns off the auto-paragraph feature outside of the story river. 

The original PR with discussion is at https://github.com/Jermolene/TiddlyWiki5/pull/3854

### DIFF
--- a/core/modules/parsers/wikiparser/rules/autoparagraph.js
+++ b/core/modules/parsers/wikiparser/rules/autoparagraph.js
@@ -1,0 +1,72 @@
+/*\
+title: $:/core/modules/parsers/wikiparser/rules/autoparagraph.js
+type: application/javascript
+module-type: wikirule
+
+Wiki pragma rule for whether p tags should be inserted for text blocks
+
+```
+\autoparagraph yes
+\autoparagraph no
+```
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.name = "autoparagraph";
+exports.types = {pragma: true};
+
+/*
+Instantiate parse rule
+*/
+exports.init = function(parser) {
+	this.parser = parser;
+	// Regexp to match
+	this.matchRegExp = /^\\autoparagraph[^\S\n]/mg;
+};
+
+/*
+Parse the most recent match
+*/
+exports.parse = function() {
+	// Move past the pragma invocation
+	this.parser.pos = this.matchRegExp.lastIndex;
+	// Parse whitespace delimited tokens terminated by a line break
+	var reMatch = /[^\S\n]*(\S+)|(\r?\n)/mg,
+		tokens = [];
+	reMatch.lastIndex = this.parser.pos;
+	var match = reMatch.exec(this.parser.source);
+	while(match && match.index === this.parser.pos) {
+		this.parser.pos = reMatch.lastIndex;
+		// Exit if we've got the line break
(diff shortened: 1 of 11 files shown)


Arlen22:
@Jermolene I know this idea got hijacked by that other suggestion which I won't link to for fear of getting more people involved in this again. This one feature is very simple and very needed. I was just now trying to style some parts of the TW UI using flex styling and I again ran into this multiple times. It's really frustrating because it completely and totally breaks the predictability of styling and sometimes even completely breaks the ability to style all together. I don't know what whitespace trim does and I don't want to do that, I only want to disable this particular feature completely within the UI. It makes a lot more sense once you understand that the parser builds a widget tree and the widgets generate the output. Being able to turn off linebreaks adding additional widgets makes it much simpler to control the output predictably.

Arlen22:
In the particular case I am working on, whitespace trim did not fix the problem.

Jermolene:
Hi @Arlen22 apologies we're too close to the release of v5.1.23 to merge this now. Happy to pick it up afterwards, but in the meantime do check out `\whitespace trim` to see if it helps

Arlen22:
Although #4290 may be of interest to many, as a structure developer rather than a content developer, I am very desparate to see this feature have the option to be disabled. It has been the bane of my CSS styling every single time I try, because it prevents rearranging the page in any meaningful way to give different layouts. We have already addressed several other problems in the past and this is the only one left that seriously affects page layouts. 

I really wish we could have two content types, though, one for structure and one for content. But since disabling p tags is the only change it would really involve, I don't know if it would be worth it just for that. I imagine there are some other things about wikitext that are also a headache for structure developers, but I don't know what they are. 

I guess you could add something to the toolbar to quickly toggle it, if necessary, since the buttons there already change based on the content type. But it's not something that should often change, since you are normally writing tiddlers in one context or the other, but should rarely need to specify it unless it's a structure tiddler to use inside a content tiddler, in which case selecting the content type from the list is easy enough. 

I don't understand much about how the parser works, so perhaps my proposal misses a lot of what the parser is intended for. But the constant p tags are a game of whackamole. In the current 5.2.0, there are already two in the `.tc-dropzone` of all places, which is obviously a structure block. There are more in the sidebar as well. It completely breaks any attempt to rearrange the structure of the page using flex or float CSS, and unfortunately whitespace trim does not always fix it and I don't know why. 

The biggest reason I think this is both essential and unobtrusive is that we almost never mix paragraph tags and page structure together. P tags are usually used for content but rarely anywhere else, and if you do need them for some reason in page structure, it's much easier to just write `<p></p>` than it is to mess around with line breaks. 

Because TiddlyWiki is so transclusion oriented you rarely put both kinds of wikitext on the same page, and it makes a lot more sense to use a variable than to have to specify it for every single tiddler you make. Then you just wrap your tiddler body transclusion in a set widget which turns on the paragraph feature, and every single tiddler gets rendered with paragraphs turned on.  If you want a structure block on a content page, you are most likely transcluding it anyway, in which case you set its content-type to `structure`, or use the pragma I originally proposed, or wrap the transclusion in a set widget. 

We could use `text/tiddlywiki; content` and `text/tiddlywiki; structure`  for the two content types (which is valid content-type syntax). 

Please don't use `application/tiddlywiki` and `text/tiddlywiki` because that isn't how the content-type system works.

PR #4287, opened 2019-10-04
Add an example to the $fieldmangler widget

mklauber:
Add an example to the field mangler widget demonstrating the `tiddler` param.


Jermolene:
Thanks @mklauber

PR #4288, opened 2019-10-04
Documentation: typo in Range widget documentation

mklauber:
The documentation for the range widget refers to radio buttons, instead of range sliders.


Jermolene:
Thanks @mklauber

PR #4290, opened 2019-10-05
Improve p generation

heronils:
This is a big PR (Pull Request) – not the logic which introduces the feature, but the consequences – so i need to write some lengthy stuff to explain things. Sorry :-) You dont need to read it fully, just have a look at the bold looking text parts ;-)

I also suggest to skim through the code changes commit by commit.

(Should you agree to these changes, Jeremy, please **dont merge** this PR and also not the original  #4195 PR, i will pull another big one, where i adapt to changes meanwhile done in the master branch
and to any critic and suggestions made by you all)

## The problem

The problem: In some locations `p`´s wrap block elements which should not be inside of `p`´s.

Block elements like `div`, `ul`, `ol`, `table`, `blockquote`, etc.

eg.

```
(empty line)
<div>
...
</div>
(empty line)
```
becomes

```
<p><div>...</div></p>
```

but we want it to become

```
<div>...</div>
```

This PR tries to solve this problem.

#4285 and #345 are other interesting approaches. However, if i got this right, they wont allow us to handle it directly inside of the Wiki text on a per paragraph level. Maybe we should do all three, i dont know :-)

For the records, here is what will be changed in this PR, and what will stay the same:

### This will change (code examples are prettyprinted)

```
(empty line)
<div>
text
</div>
(empty line)
```

becomes _before_ the PR

```
<p>
<div>text</div>
</p>
```

and _after_ the PR

```
<div>text</div>
```

### This will stay the same

This

```
(empty line)
<button>text</button>
(empty line)
```

always becomes this

```
<p>
<button>text</button>
</p>
```

And this

```
(empty line)
<div>
(empty line)
text
(empty line)
</div>
(empty line)
```

always becomes this

```
<div>
<p>text</p>
</div>
```

### 1. change the regex to introduce the different parsing behavior.

(implemented in `html.js`)

### 2. update the css

(implemented in `base.tid`)

For now this is just one rule fixing the margins in the side bar.

### 3. _conservatively_ skip some newlines

(implemented in `html.js`, `wikiparser.js`)

Skip only after an opening tag of a code or pre element, to prevent empty first lines.

Skip more conservatively than in my original PR, because Tiddlywiki creates more things than just HTML, which i wasn´t aware of. For example *textslicer* would break with my other PR.

Also, [Whitespaces are not a big problem in HTML](https://medium.com/@patrickbrosset/when-does-white-space-matter-in-html-b90e8a7cdd33) – i incorrectly thought they were. For the locations where they *are* a problem there is the best solution to just avoid them in the HTML and the second best to use [CSS magic](https://css-tricks.com/fighting-the-space-between-inline-block-elements/).

But thats not the real issue of this PR – should it be an issue at all – so lets only trim those newlines after an opening pre or code start tag, because these *are* issues with this PR.

### 4. add a feature to explicitely skip a linebreak, comparable to how python does it

(implemented in `html.js`, `wikiparser.js`)

Should we introduce this PR, then in a lot of code locations we need to remove linebreaks after an opening tag in order for the whole tag to stay wrapped in a p. Buttons are an example.

#### 4.1. Aestetic problem:

An example is this code:

```
(empty line)
<button>
<element-with-lots-of-attributes>
...
<element-with-lots-of-attributes>
</button>
(empty line)
```

Before the PR, the outermost `<button>...</button>` gets wrapped in a p, which we want.

And we want it to also stay wrapped in a p after the PR.

Therefore we would have to change the html to ...

```
(empty line)
<button><element-with-lots-of-attributes>
...
<element-with-lots-of-attributes>
</button>
(empty line)
```

... which is not a nice formatting. The eye wants to have a linebreak after the opening `<button>` tag.

#### 4.2 solution

We introduce a function which skips explicit linebreaks directly after the `>` of an opening html tag.
An explicit linebreak is a `\\\n` (backslash followed by a newline). It will be removed.

So we can write

```
(empty line)
<button>\
<element-with-lots-of-attributes>
...
<element-with-lots-of-attributes>
</button>
(empty line)
```

which gets interpreted by the parser as if we wrote

```
(empty line)
<button><element-with-lots-of-attributes>
...
<element-with-lots-of-attributes>
</button>
(empty line)
```

(The whole block gets wrapped in a p).

It is backwards compatible, because there is no search result for `>\\\n` when searching in all `*.tid` files in the main branch of the code base (the TiddlyWiki5 folder).

It also makes the diff here at Github to be easier to read.

#### Methodology

As an example what i do, this PR shows all changes done on just the tw5.com edition. I would – should you decide to do it this way, Jeremy – then also do this on the other editions.

**In general i do this**

* Create two [static html](https://tiddlywiki.com/#Generating%20Static%20Sites%20with%20TiddlyWiki) documents for the specific edition, one from the master and one from my PR branch.
* Normalize some irrelevant differences away using a conservative normalizing script. This script also removes all invalid `p`´s from the master branch output html file, but not from that of my PR (script attached, quick and dirty, work in progress, best viewed with [Code Browser](https://github.com/heronils/Code_Browser_49)).
* Compare the diffs of the static outputs (using Win Merge).
* Apply the necessary changes manually to the `*.tid` files in my branch. See below how i do that.
* Regenerate the static html documents from time to time.
* Update and re-call the normalize script.
* Compare again
* And so on.

I also partially use [Treesearcher](https://github.com/heronils/Treesearcher) for this job.

This is in 90% of the time adding (go away, p!) and removing (stay, p!) newlines after opening start tags.

I do this until the diffs are equal or until it is clear that the remaining diffs are not a problem or _should_ stay different (like eg. the docs documenting the old behaviour).

I also fix some errors found on the way, when they are relevant to the diffing. I apply some of these to both branches in order to not have them pollute the diff.

With this method now the two tw5.com editions are the same after normalizing, except of five locations where it is clear why it is different and that it must be different (eg. some examples calculating the size of the wiki and some things for which my PR creates better html – manually checked all these).

But you wont see this when diffing this PR with the main branch because i have not pulled everything – only changes made in the tw5.com edition folder or in parent folders are here, changes to the other editions are not here (can upp them too if you are interested).

**The process of manually editing the Tiddlers is most of the time this**

* Add newlines after opening tags, where we want wrapping `p`´s to disappear.
* Remove newlines after opening tags (usually by appending a backslash, see 4.) where we want `p`´s to stay.
* Fix some broken html but only if it is related to this PR (aka if it pollutes the diff). I left other things as they were in order to not make this PR confusing.
* In some rare locations add an empty line between a standalone tag and an `\end` token in a macro definition, or at the end of a tiddler. That because `<standalone/>` gets wrapped in a p, but `<standalone/>\n` does not (not in the master branch and not in my PR).
* In some rare locations, replace standalone `<$macrocall .../>` with `<$macrocall ...></$macrocall>` (same true for some other widgets like *transclude* or *text*) in order to enforce a wrapping p when the referenced html does not create that p.

The last two points are best explained by example:

```
\define foo()
<button/>
\end

<<foo>>

<$macrocall $name='foo'/>

<$macrocall $name='foo'></$macrocall>
```

will generate (prettyprinted)

```
<p><button></button></p>
<p><button></button></p>
<p><button></button></p>
```

but if we add an empty line between `<button/>` and `\end` then only the last call syntax will wrap with a p:

```
<button></button>
<button></button>
<p><button></button></p>
```

Transclusion works the same (add `<button/>` to a tiddler, and transclude it using the different three syntaxes, then – if there is a newline at the end of the transcluded tiddler – only `<$transclude ...></$transclude>` will wrap with a p.

### 6. add tests

(implemented in `test-wikitext-parser.js`)

This PR does not break the old tests. I tests for the new behavior (and some headings for easier overview).

### 8. let 'list items' of `blockquote`´s be `div`´s instead of `p`´s, update the CSS accordingly

(implemented in `list.js` and `base.tid` but not part of this specific PR, Will be part of the final one or one small one – if you want to have it)

A bit unrelated to this PR, but im using it in the diffs because of technical reasons:
The normalizing script removes all `p`´s which wrap block elements from the master branch static output, but not from that of my PR, in order to check if my PR generates correct html. The invalid html generated by eg `> > ...` – results in a `blockquote` inside of a `p` – does not get catched by my PR, so it disturbs the diff.

As it is invalid html anyway and we want html5 compatibility, we should change this.

The added css rule would be something like this, in order to make those divs behave like `p`´s when they are direct childs of `blockquotes`:

```
blockquote > div {
    margin-top: 1em;
    margin-bottom: 1em;
}
```

### 9. (optional) allow prettier formatting in attribute values

Unrelated to my PR, will pull in another PR. (it would currently disturb the diffing, except i apply it to both branches)

I saw that there are quite a few locations where complex html is put into widget attribute strings. Like eg. the *wikitext-example* macrocalls. We could have a nicer formatting for these cases.

So i added a 'dedent' preprocessor, to be optionally prepended to strings

(implemented but can not find it right now)

It allows us to write

```
<$widget attr=d"""

    <some>
        <complex/>
            <html/>
        <here/>
    </some>

"""/>
```

(notice the d ('dedent') in front of the triple string, inspired from pythons format/raw/binary string syntax)

which gets exactly parsed as if we wrote

```
<$widget attr="""<some>
    <complex/>
        <html/>
    <here/>
</some>"""/>
```

Aka,
    newlines at the start of the whole string,
    the maximum common indent of all lines, and
    whitespaces at the end of the string
get stripped away.

Idea: let `'''...'''` be an alias for `d"""..."""`

The dedent logic will only be called when there is a 'd' directly in front of the string.

### 10. please confirm if you agree with these changes, Jeremy.

I would like to ask for a confirmation if you actually have interest in having these things done this way, Jeremy, because it is a bit time consuming.

Ok, thats all folks :-)


Arlen22:
My PR (https://github.com/Jermolene/TiddlyWiki5/pull/4285) allows you to explicitly turn the autoparagraph feature on or off for the entire page using a pragma rule, and for child transclusions using the set widget. The default depends on where the tiddler is transcluded into, based on the variable, which is set to on for the story river and models, in my PR, and off everywhere else, since that's mostly UI code and tags should be explicit. 

If https://github.com/Jermolene/TiddlyWiki5/issues/345 was implemented, I don't think the autoparagraph pragma or set variable would be affected by it. But there is currently no way to turn this feature off at all, so that would need to be added anyway. 

The reason I put forward my PR is because p tag problems turned an hour-long mashup into 6 or 8 hours of editing templates because there was literally no way to get around the problem using only CSS. And that doesn't work if you're trying to ship a drop-in theme that you don't need to update every time a template changes. CSS is massively powerful, as anyone knows who's forgotten to specify a theme in their tiddlywiki.info! :) 

I believe what stalled my PR was that I didn't know how to write the proper tests or research properly what all my change affected. It's obvious you know much more about this. Would it be possible to add the ability to explicitly turn off the feature? I would be glad to contribute what I've written and help integrate it and I'll let you and everyone else decide whether auto-paragraph should be on or off in UI code -- I'm more interested in the option being available. 

In detail, the way my feature works is based on a variable `tv-auto-paragraph` which is used to set the autoParagraph option on the parser. The change to `wikiparser.js` was fairly simple. 

https://github.com/Jermolene/TiddlyWiki5/blob/ae2a0e42a241d4f9a557d4c6266ed42b00fc2e95/core/modules/parsers/wikiparser/wikiparser.js#L30-L36

https://github.com/Jermolene/TiddlyWiki5/blob/ae2a0e42a241d4f9a557d4c6266ed42b00fc2e95/core/modules/parsers/wikiparser/wikiparser.js#L220-L223

The rest of the modifications in the PR are all the different places this option needed to be set based on the variable, as well as the pragma (`autoparagraph.js`), which sets the parser directly for the page, and sets the variable for all child transclusions.

https://github.com/Jermolene/TiddlyWiki5/pull/4285/files

I welcome any questions you might have about this.

Arlen22:
After looking through the changes you made to the different files to keep paragraph tags consistent, I've change my PR to default to autoparagraph turned on. I can understand where having anything else could get a bit sticky. I'm not really sure how that works for me as a theme developer but I guess I'll find ways to work around it. But all of these changes should improve the landscape a lot, especially with the addition of /whitespace trim in several core templates. 

This one is sort of take it or leave it. Maybe It should be restricted to the `wikiparser.js` file itself just so the option is there in Javascript for other libraries to make use of, and we should just leave the whole tv-auto-paragraph thing behind.

Arlen22:
Speaking of `\whitespace trim`, does this conflict at all with that?

Jermolene:
Hi @heronils many thanks for your hard work on this. I appreciate that there is a lot of interest in improving the paragraph generation behaviour, and welcome the rigorous approach you've taken.

One of the problems in this area is that any improvements will inevitably break backwards compatibility, because that's rather the point: this is a situation where we explicitly want to change the current behaviour. Both this PR and @Arlen22's #4285 try to walk a delicate balance between changing things and keeping them the same. In the case of this PR, things change a lot (e.g. all the new trailing backslashes).

I think the way to tackle this is by introducing a new, revised wikitext syntax and have tiddlers explicitly opt-in to using it. This could be done by adding a vocabulary parameter to the MIME type as proposed in #345, or by introducing a completely new MIME type. This allows us to somewhat separate designing the new syntax from "fixing" the paragraph issues in the core templates.

A useful tool to help us appraise changes to the paragraph generation behaviour would be to attempt to write end user documentation (it would also be useful if the current behaviour were documented, too). Our goal should be that paragraph generation is predictable from a users perspective.

Finally, as I've argued elsewhere, I think the only way we can respect all of HTML's nesting rules is to handle paragraph generation at render time, not parse time.

heronils:
Good evening,

@Arlen22

Question, how does it work using the *set* widget?

This trims the wrapping p away for me:

```
\autoparagraph no

foo
```

but this doesnt:

```
<$set name="tv-auto-paragraph" value="no">

foo

</$set>
```

Can you give a working example?

> (...) there was literally no way to get around the problem using only CSS. 

Yes, because CSS can not match elements based on the children they have. That makes it a nasty problem.

> that I didn't know how to write the proper tests (...)

TiddlyWikis [tests](https://tiddlywiki.com/dev/#TestingMechanism) are located in `TiddlyWiki5/editions/test/tiddlers/tests`, and you can run all those tests by doing `node ./tiddlywiki.js ./editions/test` from the root folder. A candidate to add tests for autoparagraph would for example be the `test-wikitext-parser.js`.

I wrote an example test:

```
it("should handle the autoparagraph setting", function() {
	expect(parse("\\autoparagraph no\n\nfoo\n\nbar\n\nbaz")).toEqual(

		[ { type: "set", attributes: { name: { type: "string", value: "tv-auto-paragraph" }, value: { type: "string", value: "no" } }, children: [ { type: "text", text: "foo" }, { type: "text", text: "bar" }, { type: "text", text: "baz" } ] } ]

	);
	// ... more tests, eg, for var syntax
});
```

The other `test-*.js` files have examples how to create tiddlers, etc.

However, lets see how this thread continues, may be not necessary to write any tests.

> (...) or research properly what all my change affected.

My grand information research framework<sup>TM</sup> consists of a revolutionary technology – copy paste text and search globally using the editor of choice. I can very much recommend this :-D If i may do some self ad, my [Treesearcher](https://github.com/heronils/Treesearcher) repo can also be helpful for these things.

For the records, currently i am quite comfortable with the code written inside of `core/modules/widgets/`, `core/modules/parsers/` and `editions/`, in the other places not so much.

> Would it be possible to add the ability to explicitly turn off the feature?

If you are referring to the autoparagraph feature – It should be possible to disable it under *$:/ControlPanel* → *Info* → *Advanced* → *Parsing* but disabling pragma rules there is seemingly ignored. eg, \whitespace trim will also still be in effect when it gets disabled. inline and block rules respond to this setting. Looks like a bug.

In case you was referring to my my feature, nope, it can not be disabled, because it is not a specific parse rule but changes how 'the stuff between the rules' is parsed. I could consider to do this anyhow.

> Speaking of \whitespace trim, does this PR integrate with it?

Yes, my PR integrates well with it. The functionality of \whitespace trim is not affected, because the trimming is done after the parsing of the opening tag (and the newlines after, which defines if and where p wrapping will happen).

@Jeremey

> One of the problems in this area is that any improvements will inevitably break backwards compatibility

But ... you introduced the [previous breaking changes to the paragraph creation in a minor version](https://tiddlywiki.com/#Notes%20for%20upgrading%20to%205.0.8-beta)! Unfair! Nah, just kidding. Its not a bad idea to have another content type. Makes this more smooth, we can do it tiddler by tiddler.

> I think the way to tackle this is by introducing a new, revised wikitext syntax (...)

Do you mean just the changes in in this pull request? Or do you have other things in mind?

> (...) and have tiddlers explicitly opt-in to using it. This could be done by adding a vocabulary parameter to the MIME type as proposed in #345 or by introducing a completely new MIME type. This allows us to somewhat separate designing the new syntax from "fixing" the paragraph issues in the core templates.

Yes. A new content type would make the transformation simpler. I would suggest to call it `wikitext` (I dont like the `text/vnd.` prefix :-) ). And `wikitext.defaultvocabulary` could be an alias, to be prepared for a later addition of vocabulary logic (which i would suggest to delay a bit).

We could also have an alias to be prepared for future changes, like `wikitext.v3` (or whatever wikitext version this PR actually is).

I would also like to suggest another new type: `template`, which we could use for template tiddlers which use @Arlen22´s autoparagraph mechanism plus whitespace trimming by default: I already talked about this [here](https://github.com/Jermolene/TiddlyWiki5/pull/4195#issuecomment-525118866) – middle of the post. This has the advantage that we can do Arlen22´s logic but without passing around global variables. And it would be nice to write templates without worrying about any whitespaces, which is something one still has to care about in my PR.

pmario:
>Yes. A new content type would make the transformation simpler. I would suggest to call it wikitext (I dont like the text/vnd. prefix :-) ). And wikitext.defaultvocabulary could be an alias, to be prepared for a later addition of vocabulary logic (which i would suggest to delay a bit).

:) ... MIME types are standardized and TiddlyWiki is part of the "unregistered" [vendor-tree](https://en.wikipedia.org/wiki/Media_type#Vendor_tree)

attributes are standardized too!

pmario:
>In the case of this PR, things change a lot (e.g. all the new trailing backslashes).

@heronils ... Would it be possible to eliminate the backslashes, if `\whitespace trim` would be used in the UI templates?

Arlen22:
> Question, how does it work using the set widget?

I made a mistake when I first explained it. It actually only applies child widgets (transclude, list, wikify, anything that calls the parser directly), but not to the page itself. 

The autoparagraph pragma sets the parser for the current page AND puts the page inside a set widget (similar to the define macro pragma, which also uses a set widget). 

> In case you was referring to my my feature, nope, it can not be disabled, because it is not a specific parse rule but changes how 'the stuff between the rules' is parsed. I could consider to do this anyhow.

I think it's important to at least have the option in `wikiparser.js` and the change is simple and doesn't seem to conflict with your changes at all, so it should be easy to do.

> I would also like to suggest another new type: template, which we could use for template tiddlers which use @Arlen22´s autoparagraph mechanism plus whitespace trimming by default

I don't think this has any benefits at all over using the autoparagraph pragma alone. Both of them operate on the parser directly and change only the current page. There is no way to change the behavior of child widgets without using a global variable because transcluded tiddlers (transclude, list, wikify, etc) are parsed independant of the page that uses them them.

Arlen22:
> Its not a bad idea to have another content type. Makes this more smooth, we can do it tiddler by tiddler.

When we do add another content type with all the interesting additions we want to add, we should probably add a template content type which has a much simpler xml-ish structure that turns off most wikitext rules and lets you just build widget+element trees. 

If we're going to make a new content-type, it's good to think of some of our other major pain points (if there are any) and see if we can integrate them here.

heronils:
@Jeremy

I will write an example end user documentation as you suggested. Clarifies things. Will append to the commits.

@pmario

> :) ... MIME types are standardized and TiddlyWiki is part of the "unregistered" vendor-tree 
>
>parameters are standardized too!

Im not fluid in these things :-) In my tests it worked without the prefixes: I added `exports["wikitext"] = WikiParser;` at the end of wikiparser.js and then i used `wikitext` as content type. This has led me to believe that this naming scheme is just a convention.

I see that `text/vnd.tiddlywiki` is used in quite a few locations inside of TiddlyWiki. Question: If we use for example `text/vnd.tiddlywiki.v4` (or whatever wikitext version this PR is) for the new content type, will it 'inherit' from the `text/vnd.tiddlywiki` type in these locations? Because otherwise we have to later manually adjust in a few locations, otherwise invalid p wrapping will stay there.

> Would it be possible to eliminate the backslashes, if `\whitespace trim` would be used in the UI templates?

If the question is: Will `\whitespace trim` cause p wrapping, so that we dont need to use the `\`, the answer is no. Whitespace trimming happens, or happens not, independently from what my PR does, see below.

If the question is: Will whitespace trimming work nicely together with this feature, the answer is yes. Whitespace trimming, based on the `\whitespace trim` pragma, happens in [pushTextWidget()](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L342), and that is called the first time inside of parseInlineRun(). The things my PR does are all done before that happens. Here is how it works:

It starts in the [parse()](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/rules/html.js#L44) function. At this point it is already known if the element is wrapped in a p or not. Then we proceed as follows:

* [Skip the linebreak escaper](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/rules/html.js#L50).
* [do the original logic](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/rules/html.js#L52), which determines whether we will parse the following child in block or inline mode,
* [Skip newlines, should we be in a pre or a code element](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/rules/html.js#L60)
* [Decide](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/rules/html.js#L63) based on the collected information if we parse the following child as [block run](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L251) or as [inline run](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L168).

If it is an [inline run](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L168), then go deeper into the rabbit hole: [call parseInlineRunTerminated()](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L170) → [call pustTextWidget](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/core/modules/parsers/wikiparser/wikiparser.js#L190) where finally whitespace trimming may happen or not happen, depending on the `\whitespace trim` pragma.

@Arlen22

Thanks for the `var` explanation, now I've understood it.

> I think it's important to (...) have the option (...)

I will come back to this after we figured out how to proceed in general, ok?

> I don't think this has any benefits at all over using the autoparagraph pragma directly

It has. We get rid of passing around global variables. Copypasting from [an older post of mine](https://github.com/Jermolene/TiddlyWiki5/pull/4195#issuecomment-525118866):

First, add `exports["template"] = WikiParser;` at the end of the `wikiparser.js`.

Then replace

```
return [{type: "element", tag: "p", children: this.parseInlineRun(terminatorRegExp)}];
```

with

```
if (this.type === 'template') {
    return this.parseInlineRun(terminatorRegExp);
} else {
    return [{type: "element", tag: "p", children: this.parseInlineRun(terminatorRegExp)}];
}
```

Thats all we need to do when using a content type instead of a pragma (but we will probably use another content type according to @pmario).

The idea of the `template` content type is to

* behave like normal wikitext (either the current behaviour or that of my PR), except
* no p generation is done in the whole tiddler (aka, behave like a local `\autoparagraph no`)
* whitespaces are trimmed (behave like `\whitespace trim`)
* transclusions or macrocalls do not inherit the behaviour from this `template` tiddler.

heronils:
@Arlen22 

> When we do add another content type with all the interesting additions we want to add, we should probably add a template content type which has a much simpler xml-ish structure that turns off most wikitext rules and lets you just build widget+element trees.

If we would use my `template` type then all we would have to consider is to add two linebreaks before elements which we want to have parsed in block mode. But i am also open to more extreme cutting of wiki syntax in that `template` contenttype, including doing it completely without wikisyntax. I have no opinion on this.

> If we're going to make a new content-type, it's good to think of some of our other major pain points (if there are any) and see if we can integrate them here.

Do you have something specific in mind?

Arlen22:
> > I don't think this has any benefits at all over using the autoparagraph pragma directly
>
> It has. We get rid of passing around global variable

I'm having a hard time explaining this part. The pragma has two parts. One part sets it for ONLY the current page by setting a field/property on the parser, and the other part sets it for ONLY child widgets using the global variable. So we could remove the second part and only use the pragma for setting the parser for the current page. 

In other words, it's incorrect to say _"then we won't have to"_ pass around a global variable because in fact we do need to pass it around to affect child widgets, but **the variable has no affect on the current page** and therefore is useless if the current page is our only objective.   

The pragma has direct access to the parser instance and can change anything on it. The parser also has the option to wrap its output in a set widget, just like the pragma, and it cannot affect child widgets unless it does so, just like the pragma. So it is the same as the pragma. 

The following code snippet shows the two parts. The first part sets the parser option for the current page, and the second part sets the variable for child widgets. So they are unrelated and **either one** could be safely removed. 

https://github.com/Jermolene/TiddlyWiki5/blob/3f3cbf43bac5240721bdf51025b503003ac76760/core/modules/parsers/wikiparser/rules/autoparagraph.js#L56-L69

I think the return statement is actually supposed to be inside the if statement. 

Sorry for repeating myself. I'm having a hard time thinking how to explain it. I've tried to highlight my main points.

Arlen22:
> Do you have something specific in mind?

No, I don't actually. Except maybe the thing of being able to assign classes to the UL or OL of an asterisk or pound list. But I don't do much with wikitext. Mainly widgets, macros, and Javascript.

pmario:
>If the question is: Will \whitespace trim cause p wrapping, so that we dont need to use the \, the answer is no. Whitespace trimming happens, or happens not, independently from what my PR does, see below.

The point is. I would be extreamly unhappy with a backslash `\` at the end of the line. IMO it is super ugly and it has always been. Even in unix. 

And it has the potential to completely mess up windows based systems.

heronils:
Ok, here are some docs, and a code fix and updated tests.

@Jermolene i meanwhile think that another MIME type is not a good idea. There are tons of code locations where we check against ocurrences of `text/vnd.tiddlywiki` - what do we do there?

We should prepare this PR carefully, check it triple and then do that. Should it really break things then this will not be a big problem. All that could break is that p´s are missing. I am willing to do the conversions on all the tiddlers in the code base. Meanwhile i have a good eye for what i need to look.

If you make a clear decision for or against this, let me know. Im now working on other things i am interested in. I actually came to TiddlyWiki to implement a decent tagging application, this here is just fixing things on the way for me.

@pmario so you dont like the `\`, what do you suggest? Shall we remove it?

@everybody I also want to post here an idea for a template syntax, just in case a bus strikes me tomorrow:

```
h1 [class='hello', id='world', bar='baz'] (
    'content'
)

foo()
bar
baz(
    quux()
)

-- whitespace is irrelevant, free formatting! --
```

... which we currently would write as ...

```
<h1 class='hello' id='world' bar='baz'>
content
</h1>

<foo>
</foo>
<bar/>
<baz>
<quux></quux>
</baz>

<!-- whitespace is relevant, no free formatting -->
```

We could e.g. use [PEG.js](https://pegjs.org/online) for this.

heronils:
@pmario ich hab´ auch eine deutsche Doku zum Wikiparser erstellt. Weiß aber erstens nicht, ob das zu lang für´s TW Dokumentationsformat ist, und zweitens auch nicht wo ich das hinpacken soll. Kannst Du Dir das vielleicht mal angcken und Feedback geben ob das gebraucht wird? Und wenn ja, wo es hin soll? Würde das dann ins Englische übersetzen.

Wiki mit der Doku als HML im Anhang.

[wikiparser-docs-de.zip](https://github.com/Jermolene/TiddlyWiki5/files/3747369/wikiparser-docs-de.zip)

heronils:
I can not resolve these conflicts. After leaving that page my resolves are gone.

heronils:
Gentlemen, im going to force push a few changes tomorrow, should anyone of you have made local changes on this PR, backup your data.

I will regulary rebase this branch from now on.

Arlen22:
> i meanwhile think that another MIME type is not a good idea. There are tons of code locations where we check against occurrences of text/vnd.tiddlywiki - what do we do there?

https://github.com/Jermolene/TiddlyWiki5/search?l=JavaScript&p=1&q=%22text%2Fvnd.tiddlywiki%22

According to this search, I don't think it's a problem. Not specifying it as the default means that the new format would be opt-in rather than opt-out. Once we have it completed we can easily change it to be the new default, especially with all the work going into making it backward compatible.

Arlen22:
We could also save ourselves a bunch of hassle by naming the old one something else so it would still be available if someone wants it. But I think with the change as comprehensive as it is we should make a new format and then allow people to start using it whenever they want. Moving core to use it is fairly trivial. Even if it is not a drop-in replacement and we need to make a new type, maintaining backward compatibility as much as possible is greatly desired, as anyone knows who remembers the transition from TWC.

heronils:
Ok, these commits bring master and this branch in sync again.

I added my diff scripts and batches ( 6ac02d9 ). You can now check using the diff scripts, that tw5.com static htmls generated from both branches are nearly identical, after normalisation using the python script (python 3.7, pip install regex). I am not uploading the files directly, because each is 12MB. They anyway wont make sense without understanding the context (after normalisation they are half as big because i strip attributes, script and style tags). See the description in 6ac02d9 for how to generate them. As a Total Commander user (which allows me to filter-as-you-type by ADS file metadata using [plug](https://www.ghisler.ch/board/viewtopic.php?t=22592) [ins](https://totalcmd.net/plugring/NTFS_diz.html)) and long time user of Drive Image XML + Bart PE, i still stick to windose 7. So the scripts are written for this operating system. As the batch scripts are oneliners it is no problem to convert them. The python script should also work on Linux.

I do not compare against master directly but against improve-p-generation-master-compared-against, which i branched from master in order to apply small changes to it ([that branch is in my fork](https://github.com/heronils/TiddlyWiki5/tree/improve-p-generation-master-compared-against), i also rebase it). That way i can apply obvious fixes on master (i also pull most of them in other PRs) and move on with the diffing.

My current state of docs i am writing about this thing is now all inside of a single file tiddler ( b58b9d7 ). It is located in my feature branch, in the editions folder, in improve-p-generation-docs.html. [Download it here](https://github.com/heronils/TiddlyWiki5/blob/improve-p-generation/editions/improve-p-generation-docs.html). Because i have not yet found the right approach to make TWs nice tagging feature scale and be representable, i am just using very few tags in a strictly hierarchical manner. (tocs filterable by deselectable tags aside may be the solution, todo).

@Arlen22, yep we can do it also as new content type or using your pragma, or do it in the rendering phase - but this discussion is fruitless until the administrator has not made a decision how to proceed. @Jeremy suggested two times to do it in the rendering phase, but so far without suggesting any concrete ways how to do it. I found it needs to be done before DOM elements are registered in the widgettree, otherwise things like drag drop and refresh would create duplicated html.

So i will just keep pulling my stuff here. Im doing it the way i think is best. And best is in my opinion to make this the new parsing default. Because the current WikiSyntax is broken. Not just different, but broken. We can not generate `<div>text</div>` in a simple, valid way right now. My PR fixes this by completing [what TiddlyWiki started in 5.0.8.beta](https://tiddlywiki.com/#Notes%20for%20upgrading%20to%205.0.8-beta). The changes my PR introduces are much less dangerous than the changes introduced with that version, because the parsing mode for the children of the element is not affected. Just for the html element itself and just when it stands at the start of a text block (two line breaks before). Yes, standalone elements are also now always parsed in block mode when they stand alone. One can use closing tags to have p wrapping. Implicit p wrapping for standard void elements like `` is not possible, and it is also not in the main branch, if there are linebreaks after the element. In this edge case we need to write `<p></p>` or figure out something else, if this is desired.

Giving the current parsing behaviour a new content type, which we may depreace after a few years, sounds like a good idea, @Arlen22. But giving this PR a new content type makes no sense, as it is, as already mentioned, just a bug fix. Im fine with removing the `\` idea @pmario, that is not much work. Still like the idea. And no, it wont break Windows systems more than it does now. Every windows system which uses `\define my-macro()` as path name, may already break right now.

TiddlyTweeter:
@heronils 
> The problem: In some locations `p`´s wrap block elements which should not be inside of `p`´s.

As a naive user I noticed this recently in the HTML preview in TW editor. I notice its gets worse the more non-WikiText HTML you use. 

Maybe I am an idiot, but I always assumed Type: **text/vnd.tiddlywiki** was supposed to be a mixed format that honoured HTML raw.

The insertion of excess <p>'s does break some constructs in practice. For instance of you have one blank line in a Tiddler after a **<style>** tag then the styles break. That is bizarre. And could confuse a user. It did me.

TiddlyTweeter:
@Jermolene 
> One of the problems in this area is that any improvements will inevitably break backwards compatibility, because that's rather the point: this is a situation where we explicitly want to change the current behaviour. Both this PR and @Arlen22's #4285 try to walk a delicate balance between changing things and keeping them the same. In the case of this PR, things change a lot (e.g. all the new trailing backslashes).

I would say that ...

1.  Semi-brained users like me should actually KNOW about this, not discover it by accident. 

2. WHY? Because if you don't know it weird stuff can happen the user (even informed) can't understand (an example is here: https://groups.google.com/d/msg/tiddlywiki/88rBsOHv2DE/yElGoyYlBwAJ that caused me hassle)  

3. There is virtually no public info on this BIG issue. 

3. The COMPATIBILITY issue  matters for sure. But a solution that MOVES FORWARD beyond this issue also matters. I would embrace a new ContentType. In my usage I mix WikiText and raw HTML tags; I would rather it was not polluted with **P**s! :-)

pmario:
>Maybe I am an idiot, but I always assumed Type: text/vnd.tiddlywiki was supposed to be a mixed format that honoured HTML raw.

It honors raw HTML, but it doesn't fix it. So if your raw HTML contains stuff, that "violates" the spec, TW won't fix it (and it shouldn't). ... As mentioned in some cases, due to the nature of TW wikitext it creates html code, that isn't valid (according to in the spec) and therefore may be ignored by browsers. 

As I mentioned in GG somewhere else, wikitext should only be mixed with raw html, if there is no other way... 

Inserting a `<style>` element into a tiddler is the absolute last resort to apply a style. It will cause a lot of side effects for TW theming. ... So if you do it that way, you have to live with the side effects.

TiddlyTweeter:
I just commented on GG about this again at: [The Profligate "P"](https://groups.google.com/d/msg/tiddlywiki/YGRzd6fd8q0/GjFEmh-3BwAJ).

Any movement?

If this can't be fixed I think it should at least be **Documented!** Undocumented I am sure its causing problems that _users won't understand_. I myself only stumbled on it by accident when a precise CSS cascade stopped working because I wrote ```<h3>``` not ```!!!``` in a tiddler. Took hours to pin down.

Some resolution on this, please!

pmario:
@nilslindemann ... The next TW version will be 5.2.0 ... which indicates: There are incompatible changes. ... 

I think it would be the right time to re-activate this PR!

Jermolene:
> @nilslindemann ... The next TW version will be 5.2.0 ... which indicates: There are incompatible changes. ...
> 
> I think it would be the right time to re-activate this PR!

In the past, I was expecting that v5.2.x might mark a point at which we make big wikitext changes, but as things have turned out, the v5.2.x designation is being made because of the change in the store area format, which is significant because it affects tooling outside of TiddlyWiki itself.

If we do have a release that makes non-backwards compatible wikitext changes then I'd want to together all such changes that we've considered over the years. Off the top of my head:

* Fix the paragraph generation (newline behaviour)
* Turn off camel case by default
* Extending the transclusion syntax to pass parameters

nilslindemann:
Hi @pmario @Jermolene 

Unfortunately I deleted that repo online and I can not find the local backup again.

But as it is documented here, it should be no problem to redo this and create a new PR.

Let me know when you are ready for big changes @Jermolene or if there are any questions about the PR. But the commit messages (not my lengthy bla bla here) explains it quite well.

Greetings, Nils

kookma:
p generation in Tiddlywiki is an area need a real improvement! I am ready for any testing if a new PR is submitted!

Jermolene:
Thanks @nilslindemann @kookma

Arlen22:
> > @nilslindemann ... The next TW version will be 5.2.0 ... which indicates: There are incompatible changes. ...
> > I think it would be the right time to re-activate this PR!
> 
> In the past, I was expecting that v5.2.x might mark a point at which we make big wikitext changes, but as things have turned out, the v5.2.x designation is being made because of the change in the store area format, which is significant because it affects tooling outside of TiddlyWiki itself.
> 
> If we do have a release that makes non-backwards compatible wikitext changes then I'd want to together all such changes that we've considered over the years. Off the top of my head:
> 
> * Fix the paragraph generation (newline behaviour)
> * Turn off camel case by default
> * Extending the transclusion syntax to pass parameters

Please, please, please consider the far simpler proposal that I had originally made which started this whole thing which simply turned off the auto-paragraph generation completely where it was not needed. HTML and wikitext work just fine in a block-oriented XML-ish structure without all the encumbering formatting that p tags introduce.

pmario:
@nilslindemann @Jermolene 

>Unfortunately I deleted that repo online and I can not find the local backup again.

There may be a button here in this thread, with which you can reactivate the branch again. ...

Jermolene:
It is possible to merge this PR without @nilslindemann's original branch, I don't think that's a problem.

nilslindemann:
@Jermolene I would appreciate, I do not know how this works. I can not see this button, sorry.

Arlen22:
It would be so much simpler if we just created a text-block widget that would decide whether or not its contents should be in a p or a div based on the child widgets it had. And then instead of adding a p tag in the parse tree we would add a text-block widget. 

There'd be a few different ways to handle that then in the widget side of things, since that was always the problem to start with. You can walk the child widget tree before render and the dom tree after render to check what the situation is.

I still don't understand why this mammoth change is being considered, when the change I was originally requesting in https://github.com/Jermolene/TiddlyWiki5/pull/4285 was that we simply have the option to turn off p tag generation as desired, not change how and when new lines get added. That was a far simpler proposal that had no other changes connected to it.

Arlen22:
I really wish we could have two content types, though, one for structure and one for content, with an easy way to switch between them. I guess you could add something like that to the toolbar, since the buttons there already change based on the content type. But since this is the only change it would really involve, I don't know if it would be worth it just for that. I imagine there are some other things about wikitext that are also a headache for structure developers, but I don't know what they are. 

`text/tiddlywiki; content` and `text/tiddlywiki; structure` for instance, which is valid content-type syntax. 

Because TiddlyWiki is so transclusion oriented you rarely put both kinds of wikitext on the same page. If you want a structure block on a content page, you are most likely transcluding it anyway, in which case you set the child page content-type to `structure` (or use the pragma I originally proposed).

Arlen22:
Sorry, just realized my original proposal was not technically rejected. I would be happy to see my PR get added and then we could see how much of this PR we still need. I don't think it will be a lot and it would save us a huge change to the codebase.

nilslindemann:
I am just working on a regeneration of this PR on the current code base. The code is done, I am just diffing static generated wikis generated from the current master and from my PR. Give me two days, as I have to do something for tomorrow, afterwards I will finish it. Then I will close this PR and create a new one, where I hope to explain things better and especially _shorter_ :-)

nilslindemann:
@Arlen

Regarding the two content types, I fully agree. This is something we need.

There are simple tiddlers which have more wiki text than html elements or macro calls etc. Tiddlers, which everyday users edit. For these tiddlers this syntax is great.

In the complex tiddlers, which create html templates and other code snippets used throughout the wiki, this HTML style syntax becomes unreadable very quickly.

For those complex tiddlers, I would love to have a syntax like this:

```
div div div
div{ }
div(class="x")
div(class="y"){ }
p(class="z"){
    em{"hi"}
}

```

... which translates to ... 

```html
<div/><div/><div/><div></div><div class="x"/><div class="y"></div><p class="z"><em>hi</em></p>
```

... or (prettyprinted) ...

```html
<div/><div/><div/>
<div></div>
<div class="x"/>
<div class="y"></div>
<p class="z">
    <em>hi</em>
</p>
```

p's and all whitespace (by putting it in "..." strings, like all raw text) would have to be created explicitly. Otherwise, whitespace is insignificant, the resulting html is compressed.

For example, this is how the "tabs" macro definition (in tabs.tid) could look in this syntax:

```
define macro "tabs" (
    tabsList, default, state="$:/state/tab",
    class, template, buttonTemplate, retain,
    actions, explicitState
) {
    -- not sure what some of below syntaxes do exactly, so I just put them in quotes for now --
    LET qualifiedState = qualify(state)
    VARS (tabsState="{{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}") {
        div (class="tc-tab-set " class) {
            div (class="tc-tab-buttons " class) {
                FOR currentTab IN tabsList (storyview="pop") {
                    LET save-currentTiddler = currentTiddler
                    TIDDLER currentTab {
                        BUTTON (
                            set=tabsState, setTo=currentTab, default=default,
                            selectedClass="tc-tab-selected", tooltip="{{!!tooltip}}"
                        ) {
                            TIDDLER save-currentTiddler {
                                LET tv-wikilinks = "no"
                                TRANSCLUDE buttonTemplate (mode="inline") {
                                    TRANSCLUDE currentTab (field="caption") {
                                        CALL currentTab (TYPE="text/plain", OUTPUT="text/plain")
                                    }
                                }
                            }
                            actions
                        }
                    }
                }
            }
            div (class="tc-tab-divider " class)
            div (class=f"tc-tab-content " class) {
                FOR currentTab IN tabsList {
                    REVEAL (type="match", state=tabsState, text=currentTab, default=default, retain=retain) {
                        TRANSCLUDE template (mode="block") {
                            TRANSCLUDE currentTab (mode="block")
                        }
                    }
                }
            }
        }
    }
}
```

This is something we may have a look on, after we fixed this specific issue, if you have interest, @Jermolene, @pmario. We could use PEG.js for this, which is a beautiful parser generator. I have already been successful in 2019 parsing a tiddler with such a generated parser. The generated parsers have god error reporting.

Arlen:
The wrong Arlen.

Thanks,
Arlen

nilslindemann:
Ops, sorry, @Arlen22 :-)

PR #4293, opened 2019-10-06
Missing whitespace trim in PageControls

### LINKED ISSUE #4292: Buttons in the page toolbar have random spacers
### Summary

The buttons in the page toolbar on the right get spacers but not all the time.

### Steps to reproduce

1. Open TiddlyWiki

2. Make the sidebar visible

3. Examine the page toolbar and notice the space between the buttons.

4. Open the settings by clicking the "control panel" button and examine the page toolbar again.

### Expected result

There are just the configured buttons in the toolbar.

### Actual result

Some buttons have spacers, some not:

grafik

This can be visualiazed even better when selecting the whole content in the browser - the blocks between the buttons are additional spacers:

grafik

And when opening the settings ($:/ControlPanel), the "control panel" button gets an additional spacer added:

grafik

This behaviour is at least confusing. It would be better, if there are either no spacers at all or always spacers after each button and not depending on which tiddlers are open or not.

### Context

Server: TiddlyWiki 5.1.21, Ubuntu 16.04 LTS, Node.js 8.16.1
Client: Firefox 69.0.2

BurningTreeC:
fixes #4292

### DIFF
--- a/core/ui/PageControls/controlpanel.tid
+++ b/core/ui/PageControls/controlpanel.tid
@@ -3,6 +3,7 @@ tags: $:/tags/PageControls
 caption: {{$:/core/images/options-button}} {{$:/language/Buttons/ControlPanel/Caption}}
 description: {{$:/language/Buttons/ControlPanel/Hint}}
 
+\whitespace trim
 \define control-panel-button(class)
 <$button to="$:/ControlPanel" tooltip={{$:/language/Buttons/ControlPanel/Hint}} aria-label={{$:/language/Buttons/ControlPanel/Caption}} class="""$(tv-config-toolbar-class)$ $class$""">
 <$list filter="[<tv-config-toolbar-icons>match[yes]]">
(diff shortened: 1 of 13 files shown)


core/ui/PageControls/controlpanel.tid
```diff
@@ -3,6 +3,7 @@ tags: $:/tags/PageControls
 caption: {{$:/core/images/options-button}} {{$:/language/Buttons/ControlPanel/Caption}}
 description: {{$:/language/Buttons/ControlPanel/Hint}}
 
+\whitespace trim
 \define control-panel-button(class)
```
pmario: `\whitespace trim`

also needs to be inside `\define` otherwise it only works on the tiddler body content.

BurningTreeC: it's sufficient here, it already removes the spaces

Jermolene:
Thanks @BurningTreeC @pmario

PR #4304, opened 2019-10-12
change AutoFocus to Autofocus

BurningTreeC:
this PR changes all occurences of `AutoFocus` to `Autofocus`


Jermolene:
Hi @BurningTreeC my apologies I think we just need to change the content of the lingo string (ie the first file you've changed) to avoid the accidental link. We should keep tiddler titles such as $:/config/AutoFocus in camelcase for consistency with $:/config/Search/AutoFocus.

Issue #4282, opened 2019-10-03
Dropdowns should not be dismissed when clicking in a contained text editor


Jermolene:
To reproduce, paste the following into a new tiddler on https://tiddlywiki.com:

```
<$button popup="$:/SamplePopupState">Pop me up!</$button>

<$reveal type="popup" state="$:/SamplePopupState" class="tc-popup-keep">

<div class="tc-drop-down">

{{||$:/core/ui/EditTemplate/body/editor}}

<$edit tiddler="this"/>

</div>

</$reveal>
```

After saving it, click the button to display the dropdown, and scroll down to the two text boxes. Clicking on the lower text box (a plain `<textarea>`) does not dismiss the popup, but clicking on the upper one (a full iframe editor) does erroneously dismiss the popup:

<img width="444" alt="image" src="https://user-images.githubusercontent.com/174761/66112774-6658f480-e5c3-11e9-82d0-dae449632db3.png">

@BurningTreeC I think you did some work in this area?

BurningTreeC:
Hi @Jermolene , I cannot reproduce this. Testing your code on the current prerelease. Both editors don't cancel the popup

BurningTreeC:
Ah, now I see. Codemirror works fine, the standard editor doesn't

Jermolene:
Oh good! I was worried for a moment!

BurningTreeC:
I'm worried about how to fix this

Jermolene:
OK so it's because we cancel popups when the editor gets focus. Perhaps the editor widget can check whether it's within a popup before cancelling popups? It would be reasonable to scan ancestors in the DOM looking for the CSS classes.

BurningTreeC:
> OK so it's because we cancel popups when the editor gets focus. Perhaps the editor widget can check whether it's within a popup before cancelling popups? It would be reasonable to scan ancestors in the DOM looking for the CSS classes.

I thought I'd have already done so in popup.js

BurningTreeC:
@Jermolene , I'd be inclined to undo the latest changes that cancel popups when focusing textareas and give it a new try at some point. I think I haven't covered all possible cases there

Jermolene:
Thanks @BurningTreeC can you remind me of the commit/PR in question?

BurningTreeC:
@Jermolene, this should be fixed now

Jermolene:
Thanks @BurningTreeC

PR #4305, opened 2019-10-12
restore previous popup mechanism (no cancelling when focusing inputs)

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

Jermolene:
(This change was prompted by the discussion at #4282)

PR #4306, opened 2019-10-12
two missing tooltips: add-tag button and add-field button

BurningTreeC:
this adds tooltips to the add-tag button and add-field button


Jermolene:
Thanks @BurningTreeC

PR #4307, opened 2019-10-12
add default new tiddler tags mechanism and interface

BurningTreeC:


core/ui/Actions/new-image.tid
```diff
@@ -6,5 +6,5 @@ description: create a new image tiddler
 image/$(imageType)$
 \end
 <$vars imageType={{$:/config/NewImageType}}>
-<$action-sendmessage $message="tm-new-tiddler" type=<<get-type>>/>
+<$action-sendmessage $message="tm-new-tiddler" type=<<get-type>> tags={{$:/language/DefaultNewTiddlerTags!!tags}}/>
```
Jermolene: Hi @BurningTreeC I don't think that this tiddler should be in the `$:/language/` namespace because it's not part of the set of translatable strings. Instead it should be `$:/config/NewTiddler/tags`.

core/ui/Actions/new-journal.tid
```diff
@@ -2,7 +2,7 @@ title: $:/core/ui/Actions/new-journal
 tags: $:/tags/Actions
 description: create a new journal tiddler
 
-<$vars journalTitleTemplate={{$:/config/NewJournal/Title}} journalTags={{$:/config/NewJournal/Tags}} journalText={{$:/config/NewJournal/Text}}>
+<$vars journalTitleTemplate={{$:/config/NewJournal/Title}} journalTags={{{ [{$:/config/NewJournal/Tags}addsuffix[ ]addsuffix{$:/language/DefaultNewTiddlerTags!!tags}] }}} journalText={{$:/config/NewJournal/Text}}>
```
Jermolene: Should new journal tiddlers get these tags? Also it might be good to add a UI for $:/config/NewJournal/Tags

BurningTreeC: I thought it would be good that also the journal tiddlers get those additional tags

core/ui/Actions/new-image.tid
BurningTreeC: ok @Jermolene

core/ui/Actions/new-journal.tid
Jermolene: I'm not sure what's best. Visualising the control panel with a row for editing new tiddler tags and a row for new journal tags I suspect that users would perceive them as independent, and we'd have to include some instructions to explain what's going on. So I guess I'm in favour of keeping them separate. People can always duplicate the tags; it's hard for them to separate them if we leave it like this.

core/ui/Actions/new-image.tid
BurningTreeC: now it's using the tiddler `$:/config/NewTiddler/Tags` with Tags uppercase like `$:/config/NewJournal/Tags`

core/ui/Actions/new-journal.tid
BurningTreeC: It now uses only the separate new-journal tags for new journal tiddlers. the UI in the controlpanel now is the same for both

Jermolene: Thanks @BurningTreeC

PR #4312, opened 2019-10-12
Revert "add default new tiddler tags mechanism and interface"

BurningTreeC:
Reverts Jermolene/TiddlyWiki5#4307


pmario:
is there some reasoning?

PR #4314, opened 2019-10-13
add default new-tiddler and new-journal tags mechanism and...

BurningTreeC:
... control-panel interface

### DIFF
--- a/core/ui/Actions/new-image.tid
+++ b/core/ui/Actions/new-image.tid
@@ -6,5 +6,5 @@ description: create a new image tiddler
 image/$(imageType)$
 \end
 <$vars imageType={{$:/config/NewImageType}}>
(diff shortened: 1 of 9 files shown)


core/ui/Actions/new-image.tid
```diff
@@ -6,5 +6,5 @@ description: create a new image tiddler
 image/$(imageType)$
 \end
 <$vars imageType={{$:/config/NewImageType}}>
-<$action-sendmessage $message="tm-new-tiddler" type=<<get-type>>/>
+<$action-sendmessage $message="tm-new-tiddler" type=<<get-type>> tags={{$:/config/NewTiddler/Tags!!tags}}/>
```
pmario, line 9: see #4591 ... we should have used the text field similar to all other config tiddlers.

PR #4315, opened 2019-10-13
Update Saving on Android.tid

donmor:
I think it is recommended to change to the new apps, since scoped storage prevents apps from accessing the file system directly. Tiddloid and Tiddloid Lite are easier to use, and are modern apps designed for TW5.


Jermolene:
Hi @donmor as this is a pure docs update it's better to use the branch `tiddlywiki-com`, which causes it to be merged to https://tiddlywiki.com/ immediately.

donmor:
Okay I have had a new PR created in #4316

PR #4316, opened 2019-10-13
Update Saving on Android.tid

donmor:
I think it is recommended to change to the new apps, since scoped storage prevents apps from accessing the file system directly. Tiddloid and Tiddloid Lite are easier to use, and are modern apps designed for TW5.


Jermolene:
Thanks @donmor!

donmor:
That's all right.
BTW it will makes TiddlyWiki better if #3943 and #4276 were merged.

PR #4325, opened 2019-10-15
Update chinese translations

BramChen:
- Imporve chinese translations for UI of basic tab in controlpanel
- Add chinese translations for description of throttle.refresh field
- Add chinese translations for the `none` option of icon dropdown in ta manager


Jermolene:
Thanks @BramChen

Issue #4319, opened 2019-10-14
Proposal: a sidebar tab where we can save and restore stories


BurningTreeC:
The proposal would be a sidebar tab called `Stories` where we can save a current StoryList and restore saved StoryLists

Jermolene:
Hi @BurningTreeC I'm definitely interested in adding such story handling, but I don't think it should be a top level sidebar tab. It's the kind of thing that wouldn't make much sense to a first time user; it's only once the concept of the story river has bedded down that a user would understand the need. So I'd favour putting it under the "More" sidebar tab.

Jermolene:
It also occurs to me that there is a relationship with the permaview functionality, which is also presented as a way to share and save story river content.

BurningTreeC:
Hi @Jermolene, the `More` Sidebar tab makes sense to me

twMat:
>but I don't think it should be a top level sidebar tab. It's the kind of thing that wouldn't make much sense to a first time user

This thread inspired me to post #4323

AnthonyMuscio:
I posted in Proposal: High level UI for add/remove sidebar tabs #4323 but was keen to get some story saving as well.

Some thoughts

I was thinking what If I could save the current story, or as Jeremy suggested the current permaview against the current tiddler say in $:/storylist/currenttiddler and on opening such a tiddler, a button would appear to open that tiddlers saved story and either replace/add to the current story.

Because I wanted to do this for the current Project tiddler I thought about adding a filter as well sto save to the project story the current tiddler and any open tiddler with project=current project.

A little more thought may come up with a more broadly applicable solution and Proposal: High level UI for add/remove sidebar tabs #4323 can determine where it is displayed.

pmario:
I want to be able to define a story list similar to the DefaultTiddlers definition. That means, it should be possible to mix tiddler names with filters. eg: 

```
Learning [tag[HelloThere]] 
Community
```

If you copy paste the above code into the DefaultTiddlers section in the ControlPanel and hit the "home-button" in the sidebar it works. 

There is a `tm-home` event, that creates this functionality. The tm-home event doesn't allow any parameters. 

So if we create a `tm-story` message, that allows us to specify 

 - the tiddler, that contains the filter
 - the story-list tiddler, that should be modified
 - the history tiddler that should be modified

I think we are good to go. .... I think, I'll test this with a plugin.

pmario:
[...]

PR #4330, opened 2019-10-18
first try on a "Stories" MoreSideBar tab

BurningTreeC:
This PR adds a MoreSideBar tab called "Stories"

It provides a simple UI to save the current story list and to delete, open and modify lists

### DIFF
--- a/core/language/en-GB/SideBar.multids
+++ b/core/language/en-GB/SideBar.multids
@@ -10,6 +10,9 @@ Open/Caption: Open
 Orphans/Caption: Orphans
 Recent/Caption: Recent
 Shadows/Caption: Shadows
+Stories/Caption: Stories
+Stories/New/Add: add
+Stories/New/Prompt:  New Story Name
 System/Caption: System
 Tags/Caption: Tags
 Tags/Untagged/Caption: untagged
--- a/core/ui/MoreSideBar/Stories.tid
+++ b/core/ui/MoreSideBar/Stories.tid
@@ -1,0 +1,17 @@
+title: $:/core/ui/MoreSideBar/Stories
+tags: $:/tags/MoreSideBar
+
+\whitespace trim
+\define lingo-base() $:/language/SideBar/Stories/
+\define actions()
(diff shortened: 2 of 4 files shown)


pmario:
hmmm, 

IMO, we should create this as a plugin first.

BurningTreeC:
>IMO, we should create this as a plugin first.

@pmario, that's ok

Jermolene:
Thanks @BurningTreeC -- this PR emerged from discussion on a ticket you raise, in such situations it's very useful if you can link back to it.

@pmario I think the point here is explicitly to explore a version of the functionality that can go in the core. It's actually been on my mind since last year when @joearms often argued to me that it belonged in the core (we called the feature "trails" after Vannevar Bush).

The difficulty, I think, is making a UI that is coherent with what we have already. I mentioned over on the original ticket that there is a relationship with the permaview functionality, which is a simple way to record and share a story.

So, for example (making things up on the fly), the way into the story functionality could be via the permaview button. Clicking it could perform the present functionality but also open a normally hidden sidebar segment consisting of a list of story lists, with the new entry at the top. Clicking on a list entry would swap in the new story list, and also copy the permaview to the clipboard.

There's not much point in adding to the core that which could be a standalone plugin, the opportunity is to weave the enhanced story/trails functionality into the front of the TW5 UI.

pmario:
I'm OK with the functionality in the core, but I think the "permaview" function is not powerful enough. 

As mentioned in https://github.com/Jermolene/TiddlyWiki5/issues/4319 I think a story or trail needs to be able to use the full power of our filter functions.

pmario:
I was also thinking about some possibilities, since the discussion about improved functions of the bundle plugin did start. A proof of concept is on the way, which should nicely fit into several parts of the existing UI.

Jermolene:
Hi @pmario right, it's because the permaview functionality isn't powerful enough that I'm proposing extending it, rather than just bolting on the story functionality as a separate thing. I commented on #4319 that yes it seems clear that one should be able to specify stories as a filter.

PR #4331, opened 2019-10-19
setting focus when cancelling/deleting/closing tiddlers

BurningTreeC:
This PR uses the `tm-focus-selector` message to set focus to the next tiddler in the story river when we cancel/close/delete tiddlers in the river


BurningTreeC:
Peek 2019-10-20 11-03

core/ui/EditToolbar/delete.tid
```diff
@@ -3,11 +3,14 @@ tags: $:/tags/EditToolbar $:/tags/ViewToolbar
 caption: {{$:/core/images/delete-button}} {{$:/language/Buttons/Delete/Caption}}
 description: {{$:/language/Buttons/Delete/Hint}}
 
-<$button message="tm-delete-tiddler" tooltip={{$:/language/Buttons/Delete/Hint}} aria-label={{$:/language/Buttons/Delete/Caption}} class=<<tv-config-toolbar-class>>>
+\define focus-selector()
+[data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame input
+\end
+<$button actions="""<$list filter="[list[$:/StoryList]after<currentTiddler>]"><$set name="cssEscapedTitle" value={{{ [<currentTiddler>escapecss[]] }}}><$action-sendmessage $message="tm-focus-selector" $param=<<focus-selector>>/></$set></$list><$action-sendmessage $message="tm-delete-tiddler"/>""" tooltip={{$:/language/Buttons/Delete/Hint}} aria-label={{$:/language/Buttons/Delete/Caption}} class=<<tv-config-toolbar-class>>>
```
Jermolene: Can we make the actions attribute use a separate macro definition for the action text to make it easier to read?

core/ui/ViewToolbar/close.tid
```diff
@@ -4,7 +4,10 @@ caption: {{$:/core/images/close-button}} {{$:/language/Buttons/Close/Caption}}
 description: {{$:/language/Buttons/Close/Hint}}
 
 \whitespace trim
-<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class=<<tv-config-toolbar-class>>>
+\define focus-selector()
+[data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame input
+\end
+<$button actions="""<$list filter="[list[$:/StoryList]after<currentTiddler>]"><$set name="cssEscapedTitle" value={{{ [<currentTiddler>escapecss[]] }}}><$action-sendmessage $message="tm-focus-selector" $param=<<focus-selector>>/></$set></$list><$action-sendmessage $message="tm-close-tiddler"/>""" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class=<<tv-config-toolbar-class>>>
```
Jermolene: Likewise, it would be useful to refactor the actions string into a macro definition.

core/ui/EditToolbar/delete.tid
BurningTreeC: Thanks @Jermolene, it's more readable now

core/ui/ViewToolbar/close.tid
BurningTreeC: This is also more readable now

core/ui/EditTemplate.tid
```diff
@@ -8,13 +8,26 @@ title: $:/core/ui/EditTemplate
 <$action-deletetiddler $tiddler=<<newFieldValueTiddler>>/>
 <$action-sendmessage $message="tm-save-tiddler"/>
 \end
+\define focus-selector()
+[data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame input
```
Jermolene: This selector will target the first input element in the tiddler frame. Perhaps we should respect the new autofocus setting we use for focussing new tiddlers?

Jermolene: (Plus the same point to the two further examples below)

BurningTreeC: That makes sense @Jermolene. I'm trying to do so but I'm having a hard time targeting the framed text-editor. Maybe you know a trick for doing so?

BurningTreeC:
@Jermolene, now this works for all fields. for the codemirror editor I'd need a macro that overrides a core macro if that's possible

BurningTreeC:
closing, needs work

BurningTreeC:
Now this needs some testing (on firefox there might be a bug with the framed text-editor)

BurningTreeC:
This now works as expected, although Firefox has a bug

BurningTreeC:
@Jermolene I'll create a demo page soon - could this then be considered?

Jermolene:
Hi @BurningTreeC this looks good to me, but a demo would be helpful to see what it feels like.

BurningTreeC:
@Jermolene [Here's a Demo](http://cancelling-closing-tiddlers.tiddlyspot.com/) to play around with

BurningTreeC:
@Jermolene - any thoughts about this?

twMat:
@BurningTreeC - in the demo, if one sets focus to a draft, or just accept default focus on top one, and deletes that draft, then the following scroll is off making the next draft scroll half past viewport. It does not happen for non-drafts, i.e view mode tids.

BurningTreeC:
@twMat could you provide a gif to demonstrate what you mean?

twMat:
@BurningTreeC 
OK, if the following step-by-step isn't enough, I'll try to make a gif:

On Win10 in **Chrome but not on FF**

1. go to demo
2. right away delete "draft 5"
3. now the scrolling/navigation mechanism does *not* make the top of "draft 4" align with viewport top, instead the viewport top stops in the middle of "draft 4" (slightly above draft 4's editor bottom edge).

BurningTreeC:
@twMat this is odd, I cannot reproduce this on Chrome or Firefox, but I'm on Linux. I'll investigate on a Windows machine. Thanks for testing!

twMat:
@BurningTreeC  FYI, I also tried to replicate it on tw.com, i.e opening a lot of drafts, including filling some of them to make them longer than the other ones, and then closing - but they all scroll/navigate correctly.

BurningTreeC:
I'll investigate on this problem. @Jermolene , is this PR still of interest?

saqimtiaz:
@BurningTreeC I feel like it would be good to implement this at the same time as the keyboard handling for story river navigation to ensure that the two work well together. I would be up for working together on the two post 5.1.23 if you would like that.

BurningTreeC:
@saqimtiaz yes that's probably the best. I'd be very grateful about your help on this!

Jermolene:
Hi @BurningTreeC is everything OK? I noticed you closing a bunch of PRs that are still under discussion.

PR #4333, opened 2019-10-20
Let Colour-Picker be valid HTML

heronils:
These [divs](https://w3c.github.io/html-reference/div.html) will be wrapped inside of a [p](https://w3c.github.io/html-reference/p.html), but divs are not allowed there (they are not [phrasing content](https://w3c.github.io/html-reference/common-models.html#common.elem.phrasing)). The [a](https://w3c.github.io/html-reference/a.html) does not change this, because it is [transparent](https://w3c.github.io/html-reference/terminology.html#transparent). Therefore let's replace the divs with spans.

(Accidently also pulled that whitespace removed by Geany).

[Example Colour Picker](https://tiddlywiki.com/#colour-picker%20Macro%20(Example%201)) - currently invalid html.

### DIFF
--- a/core/wiki/macros/colour-picker.tid
+++ b/core/wiki/macros/colour-picker.tid
@@ -15,7 +15,7 @@ $(colour-picker-update-recent)$
 
 $actions$
 
-<div style="background-color: $(colour-picker-value)$; width: 100%; height: 100%; border-radius: 50%;"/>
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @heronils

heronils:
Welcome @Jermolene

PR #4335, opened 2019-10-20
Enable the internals plugin by default in docs

heronils:
1. Why hide such a useful thing from the users?
2. When playing around with code examples from the docs they may want to know how the resulting html looks.


pmario:
The problem is, that this info is only useful for experienced++ users. It's way to confusing for everyone else. 

I think, it should be part of the dev edition and prerelease (as it is)

Jermolene:
The policy is for the main tiddlywiki.com release is to use the minimum necessary plugins. Having said that, I noticed that it currently includes evernote migration and nw.js saver, neither of which are strictly needed. (It also includes the railroad diagram generator and the browser sniffing plugin which is used to customise GettingStarted). I'd be in favour of removing those two unneeded plugins.

I do think @heronils suggestion has merit. tiddlywiki.com is where people are likely to be experimenting with TiddlyWiki. However, to @pmario's point, adding the internals plugin certainly makes the preview dropdown more complicated. To a certain extent that boat has already sailed because last year we added the differencing entries to the preview dropdown. So on balance, I'm inclined to accept this unless anyone has a different objection?

heronils:
We could append a button which links to the _Text Preview_ documentation tiddler. I have written some documentation, see my commit below.

Then i appended ...

```

<hr>

<$button tag="a">
<$action-navigate $to="Text preview"/>
What is this?
</$button>
```

... at the end of [preview-type-dropdown.tid](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/EditorToolbar/preview-type-dropdown.tid), which looks really good – in my [p-branch](https://github.com/Jermolene/TiddlyWiki5/pull/4290) :-)

Jermolene:
Thanks @heronils

editions/tw5.com/tiddlywiki.info
```diff
@@ -5,6 +5,7 @@
 		"tiddlywiki/browser-sniff",
 		"tiddlywiki/railroad",
 		"tiddlywiki/evernote",
+		"tiddlywiki/internals"
```
pmario, line 8: missing comma at the end -  please update.

Jermolene: Thanks @pmario

PR #4337, opened 2019-10-20
String Modifiers

heronils:
Already mentioned this in my 'improve p generation' PR message, under 9. 

you can now write ...

```
<$macrocall $name="my-macro" src=dn"""

  <my>
    <complex>
      <code>inside of the attribute vale</code>
    </complex>
  </my>
  
"""/>
```

... and it gets dedented and stripped as if you wrote ...

```
<$macrocall $name="my-macro" src="""<my>
  <complex>
    <code>inside of the attribute vale</code>
  </complex>
</my>"""/>
```

(notice the `dn` before the string)

Shall `'''...'''` be an alias for `dn"""..."""`?


Leilei332:
Sadly, macrocall widget is considered deprecated now, and IMO dedenting in attributes should be replaced by the new `\widget` syntax.

PR #4340, opened 2019-10-22
Add keyboard-selectable tags-dropdown (tag-picker)

BurningTreeC:
Peek 2019-10-22 11-27


pmario:
Does the text input still filter, if you have many existing tags?

BurningTreeC:
>Does the text input still filter, if you have many existing tags?

Yes it does, if I understand you correctly

BurningTreeC:
I'm closing this because it's too difficult to maintain the macro-code. An easier solution is needed imo

BurningTreeC:
I'll try to make some smaller commits from this if possible

BurningTreeC:
Now I've figured out the bug that bothered me, so I'm reopening this PR

core/modules/widgets/subclasses/edit-text-subclass.js
```diff
@@ -0,0 +1,65 @@
...
+A subclass that intercepts the refresh method of edit-text widgets for enhanced refreshing
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.baseClass = "edit-text";
```
Jermolene: It looks like you're extending the existing edit-text widget, rather than creating a new widget that is a subclass of it. In that case, why not just modify the existing edit-text refresh method, rather than creating a subclass? Or is this a residue from a plugin implementation?

core/modules/widgets/subclasses/edit-text-subclass.js
```diff
@@ -0,0 +1,65 @@
...
+
+	if(changedAttributes.refreshTiddler || changedAttributes.refreshAction || changedAttributes.saveTiddler) {
+		this.refreshSelf();
+		return true;
+	} else if(!changedAttributes.tiddler && !changedAttributes.field && !changedAttributes.index && !changedAttributes["default"] && !changedAttributes["class"] && !changedAttributes.placeholder && !changedAttributes.size && !changedAttributes.autoHeight && !changedAttributes.minHeight && !changedAttributes.focusPopup && !changedAttributes.rows && !changedAttributes.tabindex && !event[HEIGHT_MODE_TITLE] && !event[ENABLE_TOOLBAR_TITLE]) {
+
+		var refreshCondition = this.getAttribute("refreshCondition");
+		if(this.refreshTiddler && event[this.refreshTiddler] && (refreshCondition === "true" || refreshCondition === "yes")) {
+			this.engine.domNode.value = this.getEditInfo().value;
+			this.wiki.deleteTiddler(this.refreshTiddler);
```
Jermolene: Modifying the tiddler store during a refresh cycle is a big no-no (I have thought of putting in some code to detect and warn).

BurningTreeC: maybe it's possible without this line

BurningTreeC: yes, it isn't even needed, I'll delete it

core/modules/widgets/subclasses/edit-text-subclass.js
BurningTreeC: Yes, it's a residue from a plugin implementation. I kept it this way because I didn't want to make the refresh method too big

BurningTreeC:
Now I've realized that it doesn't work without my hack, that's why I'll close all related PR's

core/modules/widgets/subclasses/edit-text-subclass.js
```diff
@@ -0,0 +1,64 @@
...
+			this.engine.domNode.value = this.getEditInfo().value;
+		} else if(event[this.editTitle] && (event[this.editTitle].deleted !== true)) {
+			if(!event[this.refreshTiddler] && refreshCondition !== "true" && refreshCondition !== "yes") {
+				//update the saveTiddler with the new text
+				var saveTiddler = this.wiki.getTiddler(this.saveTiddler),
+					updateFields = {
+						title: this.saveTiddler
+					};
+				updateFields[this.editField] = this.getEditInfo().value;
+				this.wiki.addTiddler(new $tw.Tiddler(this.wiki.getCreationFields(),saveTiddler,updateFields,this.wiki.getModificationFields()));
```
BurningTreeC: Hi @Jermolene , would this line here be allowed or is it another no-no?

PR #4344, opened 2019-10-22
Correct the jsonstringify documentation

mklauber:
The jsonstringify substitution table does not list all substitutions, and it listed `'` as being replaced when it's not.  This updates the table based on the code at https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/utils/utils.js#L537


Jermolene:
Thanks @mklauber

PR #4345, opened 2019-10-23
add edit-text-subclass that allows refreshing the edited

BurningTreeC:
... tiddler with some text (using an action-setfield widget for example) while another tiddler still holds the input-text (for sorting a given dropdown-list for example)

The mechanism:
a `refreshTiddler` attribute specifies a tiddler that must change in order for the input to update its text with the text stored in its tiddler that's being edited (it doesn't do so currently when changing its content via an action-setfield widget through a keyboard shortcut from within the input) ... a `refreshCondition` attribute must also be `true` or `yes` so that the update occurs. That's used to distinguish the input field from others so that the correct one updates

### DIFF
--- a/core/modules/widgets/subclasses/edit-text-subclass.js
+++ b/core/modules/widgets/subclasses/edit-text-subclass.js
@@ -1,0 +1,65 @@
+/*\
+title: $:/core/modules/widgets/subclasses/edit-text-subclass.js
+type: application/javascript
+module-type: widget-subclass
+
+A subclass that intercepts the refresh method of edit-text widgets for enhanced refreshing
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.baseClass = "edit-text";
+
+exports.constructor = function(parseTreeNode,options) {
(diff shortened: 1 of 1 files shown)


BurningTreeC:
update: edited the text above

Jermolene:
Thanks @BurningTreeC -- and thanks for your patience I know you've been working on this stuff for a while.

To help me understand the problem, could you come up with a minimal test case that doesn't currently work, along with a version that works with these changes? Many thanks.

Jermolene:
Hi @BurningTreeC I'm not sure. I need to invest more time in understanding what you're proposing, and whether this would be simplified if we properly addressed focus preservation... I guess focusing on small PRs is good, to the extent that you can make the first one self-contained...

BurningTreeC:
@Jermolene #4340 is self-contained

Jermolene:
Hi @BurningTreeC just to be clear, am I correct in understanding that #4340 is self contained but doesn't work properly without #4345 (this PR)?

BurningTreeC:
@Jermolene - no, #4340 is fully self-contained and includes this PR here

BurningTreeC:
closing this, see #4340

PR #4346, opened 2019-10-23
Replace '<<...>>' with a macrocall widget

heronils:
... because there is another macrocall inside.

This was introduced by the commit 'Fix sizes of SVG icons in documentation' (SHA: 9395d7567179c436d0e8ac26fc976d717eae7f50) where this probably slipped through in a regular expression replacement session.

I searched through the codebase and the other replacements of this type are ok.


Jermolene:
Thanks @heronils, excellent spotting!

heronils:
You are welcome, @Jermolene

Issue #4349, opened 2019-10-25
tm-fold-tiddler handling in navigator.js is not even close to what the docs says!


pmario:
To reproduce:

- create 2 tiddlers `a` and `b`
- with some content !!!
- open both tiddlers a and b
- use the following code
- or import from the zip: [issue-test-tm-fold-tiddler.zip](https://github.com/Jermolene/TiddlyWiki5/files/3772372/issue-test-tm-fold-tiddler.zip)

title: test-tm-fold-tiddler

```
<$button>toggle tiddler "a" fold state
<$action-sendmessage $message="tm-fold-tiddler" $param="a" foldedState="state"/>
</$button>
```

Expected: tiddler a folding state should be toggled from a button in `test-tm-fold-tiddler` according to the doc.

existing behaviour: the content of tiddler `state` is toggled. Nothing else. 

If you use: `$:/state/folded/a` for the `foldedState` parameter it does work. BUT that's not what the documentation says: https://tiddlywiki.com/prerelease/#WidgetMessage%3A%20tm-fold-tiddler

pmario:
docs param name `foldedStatePrefix` is wrong. navigator.js uses: https://github.com/Jermolene/TiddlyWiki5/blob/6fa3e21116338e5fc7b59db2a327eedb04ba9288/core/modules/widgets/navigator.js#L578

pmario:
Technically the `tm-fold-tiddler`  is a `toggle-xxx` function, that stores the result "show" "hide" in a tiddler.

pmario:
`tm-fold-tiddler` is hadcoded in the ViewTemplate, so I think we can only fix the doc in a backwards compatible way.

Jermolene:
Hi @pmario yes the docs do look wrong, I'm not sure how that happened.

pmario:
PR incoming :)

PR #4353, opened 2019-10-28
fixes: #4349 ... bring tm-fold-xxx message docs on par with the code,…

pmario:
bring tm-fold-xxx message docs on par with the code that is executed.


pmario:
This PR only contains doc files. 

There will be a follow-up, that will fix the view-template tiddlers, since they contain a `$param` attribute, which is ignored by the code executed in navigator.js

pmario:
code can be found at: https://github.com/Jermolene/TiddlyWiki5/blob/6fa3e21116338e5fc7b59db2a327eedb04ba9288/core/modules/widgets/navigator.js#L576

pmario:
@Jermolene ... this is docs only.

Jermolene:
Thanks @pmario

PR #4355, opened 2019-10-29
Enable indexers in tag test

hoelzro:
Setting enableIndexers to an empty array ends up disabling all indexers for the wiki


Jermolene:
Thanks @hoelzro

PR #4226, opened 2019-09-01
Upgrade to Jasmine 3

ento:
A more thorough version of #4223: upgrades the version of Jasmine bundled in the jasmine plugin to 3.4.0.

Jasmine's upgrade guides:

* To 2.x: https://jasmine.github.io/tutorials/upgrading_to_2
* To 3.x: https://github.com/jasmine/jasmine/blob/v3.0.0/release_notes/3.0.md
  * 3.x dropped support for Node.js 0.x: I assume it's fine since Travis tests are run only against the latest stable version of Node 

The test suite mostly worked as-is.

Some things I'd like to call out about the changes:
* jasmine-plugin.js: `evalInContext()` does something quite similar to `$tw.modules.execute()` and I'm not sure if there's a smarter way. I didn't see a way to use `$tw.modules.execute()` as-is because I needed to tweak the `sandbox` object.
* update-packages.sh: I'm not sure if it's fine to leave a shell script like this in this directory.

PR #4356, opened 2019-10-29
Fix undefined print_ error

hoelzro:
On newer versions of Node, I get this error:

```
    TypeError: this.print_ is not a function
```

I believe this is because Node has deprecated util.print in favor of
console.log

### DIFF
--- a/plugins/tiddlywiki/jasmine/files/reporter.js
+++ b/plugins/tiddlywiki/jasmine/files/reporter.js
@@ -1,14 +1,4 @@
 (function() {
-  //
-  // Imports
-  //
-  var util;
-  try {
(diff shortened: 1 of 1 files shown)


hoelzro:
An alternative would be to upgrade the bundled version of Jasmine - I'm guessing this has been addressed in a later release!

pmario:
I think, there is an existing (waiting) PR that updates Jasmine.

pmario:
see: #4226

pmario:
@hoelzro ... IMO can be closed. `utils.print` is not part of the TW code anymore. At least I cannot find it

hoelzro:
Agreed - closing!

PR #4357, opened 2019-10-30
make tags-dropdown overflow in y-direction

BurningTreeC:
the tags-dropdown gets very long with many tags. this PR makes it scrollable

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -680,6 +680,11 @@ button.tc-untagged-label {
 	vertical-align: middle;
 }
 
+.tc-edit-tags .tc-block-dropdown {
+	max-height: 70vh;
+	overflow-y: scroll;
+}
+
 .tc-tag-manager-table .tc-tag-label {
 	white-space: normal;
 }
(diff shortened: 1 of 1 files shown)


BurningTreeC:
edit: this pr makes the tag-manager icons-dropdown overflow the same way

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -680,7 +680,7 @@ button.tc-untagged-label {
 	vertical-align: middle;
 }
 
-.tc-edit-tags .tc-block-dropdown {
+.tc-edit-tags .tc-block-dropdown, .tc-tag-manager-table .tc-drop-down {
```
pmario: If a CSS rule is valid for several elements, 2 lines should be used like: 

```
.tc-edit-tags .tc-block-dropdown, 
.tc-tag-manager-table .tc-drop-down {
..
}
```
which makes it much more obvious, what's going on.

BurningTreeC: thanks @pmario, I've changed it

BurningTreeC:
bump

Jermolene:
Hi @BurningTreeC I try to avoid nested scrollable areas in the core because many browsers don't give a good visual hint that there is out of sight content (and those that do use ugly scroll bars). In this case, one can just scroll the page, which is less fiddly eg on mobile than trying to scroll the subframe.

BurningTreeC:
Ok @Jermolene, that's reasonable

PR #4358, opened 2019-10-30
make "type" input look consistent with "fields" input

BurningTreeC:


BurningTreeC:
this PR makes the type input field look like the fields input field, for consistency

BurningTreeC:
bump

core/ui/EditTemplate/fields.tid
```diff
@@ -69,7 +69,7 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 <$fieldmangler>
 <div class="tc-edit-field-add">
 <em class="tc-edit">
-<<lingo Fields/Add/Prompt>>&nbsp;
+<<lingo Fields/Add/Prompt>>&nbsp;&nbsp;
```
pmario: as I wrote ... I'd prefer a CSS solution

Jermolene: I think this is a reasonable quick fix.

PR #4359, opened 2019-10-30
make tags input look cleaner

BurningTreeC:
... by making its background color transparent


BurningTreeC:
bump

Jermolene:
Hi @BurningTreeC I quite like this change but feel that it is inconsistent with our other usage of text boxes. I think it's the kind of thing we might do if we were to create a new default theme, for instance.

---

Here's what the add tag box looks like with this PR:

image

And here's what it looks like now:

image

pmario:
hmmm. For me the existing version makes more sense, since it defines, to which element the dropdown arrow belongs to.

BurningTreeC:
@pmario, yes I think you're right, closing this PR

Issue #3483, opened 2018-10-15
Trailing space character in title should be removed


twMat:
Serious issue: A suffixing space character in the title is apparently not removed when saving the tiddler. So "Foo " and "Foo" appear the same, both as titles in view mode and edit mode and also as tag pills.

**Proposal**: Automatic removal of trailing space characters in title when saving a tiddler and in tags when creating one.

----

Update based on discussion below: The proposal here concerns whether *tags* should be trimmed or not. The concern about *tiddler titles* turned out to be an edge case for #751 (albeit probably the most common one) 

My position regarding tags is that they should follow the same rules as titles... and IMO they should both have trailing space characters trimmed or at least marked out, e.g in pink, in edit mode.

Jermolene:
Hi @twMat I remember this coming up before -- @pmario do you recall the discussion?

twMat:
@Jermolene - You're not thinknig of #3368 are you? I don't think this is the same issue, but maybe it is? This case with a simple space character seems much more critical.

Jermolene:
I think that that is the same problem: the dangers of invisible characters in tiddler titles. 

I also found #1791 "TW overwrites existing tiddler without a warning -> data loss"

There are a lot of variants of this: there are Unicode characters that look the same but have different codes; in some situations one can't distinguish between two spaces or a single space.

Anyhow, I can't quite duplicate the problem you're seeing here. When I create a tiddler with a trailing space in the title it is automatically trimmed. Are you perhaps creating the tiddler programmatically?

Jermolene:
I think #751 does cover the same problem. It sounds like the new thing here is the discussion about tag entry, and whether they should be trimmed too.

pmario:
My opinion didn't change. see  #751, #2726, #1791,

I think the next version of TW should consistently remove whitespace,
We missed the opportunity for tw5.

AnthonyMuscio:
Regardless of core functions handling leading and trailing spaces could we not provide a mechanism to simply remove them from a given string?. I have had mysterious spaces appearing in macros and at the time would have being happy to just have a method to remove them.

hoelzro:
As long as we're gathering issues related to trailing whitespace, I reported and fixed #2850 last year.

diego898:
As tags are tiddlers too, Im reporting here that someone on the groups reported a problem where they accidently tagged their tiddler with a space before the correct string. 

I think trailing and leading white spaces should be removed!

[...]

PR #4360, opened 2019-10-30
add whitespace trims to tag-picker macro

BurningTreeC:


BurningTreeC:
bump

core/wiki/macros/tag-picker.tid
```diff
@@ -15,12 +15,13 @@ tags: $:/tags/Macro
...
 
 \define tag-picker-inner()
+\whitespace trim
 <div class="tc-edit-add-tag">
 <span class="tc-add-tag-name">
 <$keyboard key="ENTER" actions=<<add-tag-actions>>>
 <$edit-text tiddler=<<newTagNameTiddler>> tag="input" default="" placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" tabindex=<<tabIndex>> focus={{{ [{$:/config/AutoFocus}match[tags]then[true]] ~[[false]] }}}/>
 </$keyboard>
-</span> <$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button> <span class="tc-add-tag-button">
+</span>&nbsp;<$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button>&nbsp;<span class="tc-add-tag-button">
```
pmario: non breakable space IMO should not be used as a formatting tool. We need to use `margin-right or margin-left` here. Depending which element can or should push the other elements. ... &nbsp;'s cause problems with copy / paste actions.

Jermolene: This isn't really about formatting, it's just a technique for including whitespace in a tiddler with `\whitespace trim`.

Jermolene:
Thanks @BurningTreeC

twMat:
While this pull makes perfect sense, I'm just wondering if it would not prevent making tags from titles that do contain white space, considering how #3483 is not solved? I.e will my tiddler `Foo   ` be impossible to create a tag to/for?

Jermolene:
Hi @twMat this PR doesn't affect the treatment of the text of tags themselves, it's just to make the wikitext more readable.

BurningTreeC:
Hi @twMat @Jermolene ... On 5.1.21 it's possible to insert a tag like `Foo ` with a space at the end. On the current prerelease it's not possible anymore, so maybe this Pr was a mistake

BurningTreeC:
Hi @twMat @Jermolene ... On 5.1.21 it's possible to insert a tag like `Foo ` with a space at the end. On the current prerelease it's not possible anymore, so maybe this Pr was a mistake

pmario:
I think I did fix the spacing problem with: https://github.com/Jermolene/TiddlyWiki5/pull/4366 ... now it may be broken again.

Jermolene:
I'm struggling to see how this PR could have affected whether tag with a trailing space can be created. I thought at some point we deliberately trimmed any whitespace?

Meanwhile, to your point in #4366 about copying and pasting, it may make more sense to use `&#32;` instead of `&nbsp;`.

PR #4362, opened 2019-10-30
New attempt on a keyboard selectable tag-dropdown

BurningTreeC:
This PR makes much smaller changes in factory.js and simple.js

It removes an apparently superfluous condition in simple.js which blocked the mechanism from working properly, please have a look at it

PR #4365, opened 2019-11-02
New attempt on a keyboard selectable tag-dropdown

BurningTreeC:


BurningTreeC:
This is the follow-up PR to #4362 and it's ready for review

Jermolene:
Thanks @BurningTreeC 

I just wanted to check the purpose of the new "storeTiddler" attribute. It looks like it specifies the title of another tiddler which is also updated with the value of the input element as it changes. Is that right?

BurningTreeC:
@Jermolene yes that's right, but it doesn't update if the edited Tiddler changes otherwise, like through an action-setfield widget

BurningTreeC:
@Jermolene yes that's right, but it doesn't update if the edited Tiddler changes otherwise, like through an action-setfield widget

Jermolene:
Thanks @BurningTreeC what problem does that solve? Can you give a brief explanation of your solution?

BurningTreeC:
It gives us the opportunity to sort a drop down by what is stored within the storeTiddler while changing the content of the Tiddler that's being editrd

Jermolene:
Hi @BurningTreeC that whets my appetite, but isn't really adequate to explain what's going on!

We've discussed before that it's not tremendously helpful to have a PR with a blank description. I shouldn't have to spend hours reading the code to understand what's going on. In this particular case, the new "storeTiddler" attribute rather sticks out as being a hack, so naturally I'm going to want to drill down into why it's needed.

BurningTreeC:
@Jermolene ... apologies!

The new storeTiddler attribute specifies a tiddler that stores the user input by which I sort the tags dropdown. That allows me to change the text of the tiddler that's being edited when using the down/up arrow keys (selecting a dropdown list-item). This way the sorting stays the same because it isn't sorted by the tiddler that's being edited but by the `storeTiddler`. It's more a mechanism than a hack

Jermolene:
Ah, OK. So the problem is that you want to avoid updating the text input as the user steps through the options with the up and down arrows because that would cause the dropdown to be refiltered?

If that's right, why update the value in the text box as the user steps through the values? Why not just update the text box when the user presses enter?

BurningTreeC:
@Jermolene the selected item is stored somewhere for the visual feedback (the selected item has a different background) and I use the Tiddler that's being edited for that. The visual feedback is important but I didn't want to use a third Tiddler for that, it's convenient to use that one that's being edited

Jermolene:
Apologies @BurningTreeC, I'm still struggling to understand this PR.

One way forward might be to update the documentation for the edit text widget. If we can tease out a useful description for end users perhaps it will help clarify what's going on.

BurningTreeC:
@Jermolene I'm away from my computer til the end of the week but I can describe what the `storeTiddler` attribute is for

BurningTreeC:
I want to sort a drop down by the user input in an input field. Then, using the up and down arrow keys I want to update the input fields' text with the selected value but still keep the drop down sorted by the previous user input. That works if all filters use the `storeTiddler` Tiddler for filtering. That Tiddler isn't changed when using the up and down keys or the enter key for selecting an item. Pressing escape for the first time then resets the filtering, pressing escape for the second time cancels editing of the tiddler

BurningTreeC:
The `StoreTiddler`, if given, behaves like the Tiddler that's being edited. The only difference to the default behavior is that there are two tiddlers that store the user input

BurningTreeC:
bump

BurningTreeC:
closing this in favor of a new PR with better description and without conflicts

PR #4366, opened 2019-11-07
remove &nbsp from tag pill in edit mode

pmario:
PR: fix missing space between edittemplate tags #3585 introduced an unbreakable space ...

The `&nbsp` isn't needed and **causes problems**, if users copy&paste the tag text, because the "new" tag in the text input field now contains an space in front of the tag. This space invalidates the tag, so it doesn't function anymore. 

see [comment in GG](https://groups.google.com/d/msg/tiddlywiki/RQEyqPQIZSM/uaU7lgJJAAAJ) .. I also had a problem like this some time ago, which costed me several hours of debugging.


pmario:
uups. Don't merge. ... This now causes a problem if the tag has an icon. ... I'll send an update soon

pmario:
@Jermolene  Can be merged now. ... 

The `<<tag>>` macro had the same problem. It used whitespace to define a margin between the icon and the text. 

the base CSS configuration already knows about special handling for `tc-tag-label svg, tc-tag-label img`  ... so I did add the margin-right there. 

marging-bottom was added, because of better vertical alignment.

Jermolene:
Thanks @pmario

PR #4368, opened 2019-11-10
Fix "list" svg icon

BurningTreeC:
The icon has an artifact which is visible on internet explorer


Jermolene:
Hi @BurningTreeC please can you show before and after screenshots of the icon in IE and a regular browser?

Jermolene:
Fixes in 015306dfc9099f4d5d9415b64266d328a154b119

PR #4369, opened 2019-11-10
Fix "line-width" svg icon

BurningTreeC:
The icon has an artifact which is visible on internet explorer


Jermolene:
Hi @BurningTreeC please can you show before and after screenshots of the icon in IE and a regular browser?

pmario:
I can see something with IE 11.418 with line-width. I don't see a problem with list. 

BTC-PRs

pmario:
Didn't test the PRs.

pmario:
The funny thing is. If I open AdvancedSearch and the image-gallery example tiddler, the problem goes away. If AS is closed, the problem comes back...

Issue #4275, opened 2019-09-25
wiki.sortByList should skip an expensive operation when able to


Arlen22:
https://github.com/Jermolene/TiddlyWiki5/blob/e752ba7c6abcdde229258aa1562e6de518c3556e/core/modules/wiki.js#L564-L634

A discussion on Google Groups about a slow wiki led me to this line, which was the culprit for the slowdown for a tag with 36000 titles. 

https://github.com/Jermolene/TiddlyWiki5/blob/e752ba7c6abcdde229258aa1562e6de518c3556e/core/modules/wiki.js#L593

I've been studying this function for a half hour, and it suddenly occurred to me that

- A: if new position is -1 then current position is useless, 
- B: because right now if new position is -1, then it is set to current position
- C: immediately after that it checks if they are equal and only splices if they are not. 

So, in other words, if new position was set to -1, then it will be set to current position, and therefore is equal to it, and therefore prevents a splice from happening. So we could totally optimize this if new position is still -1 after the first group of if statements and just exit the function if there is nothing instead of looking up the current position first.

```
if(newPos !== -1) {
  var currPos = titles.indexOf(title);
  if(currPos !== -1) {
    titles.splice(currPos, 1);
    if(newPos >= currPos) newPos--;
    titles.splice(newPos, 0, title);
  }
}
```

Of course, you still can't use list-before and list-after on 36000 tiddlers, but who has ever done that?

Arlen22:
The original discussion is here.

https://groups.google.com/d/msg/tiddlywiki/U1ZKK8UIk94/_URkH-GxAwAJ

Jermolene:
Hi @Arlen22 great stuff, thanks for tracking things down. I think it might be helpful to add some tests before we change this. I think the tests would fit best in https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-utils.js. Might you be able to put together a PR?

pmario:
I think the main problem here is, that the tag-operator always does a `.sortByList()`. ... The "list-field" was introduced at the very beginning of TW development cycle, to get a predictable order for "tagged templates" used with the TW UI. 

We rely on the behaviour, because it is convenient. 

But I think, it could also be created like this: `[tag[test]sortby[list]]` 

So my proposal would be to create something like a "stripped down" tag-operator, that doesn't do any sorting at all. 

[...]

Issue #4179, opened 2019-08-18
Proposal : integrate a translation macro into the core


sycom:
There had been many discussions about TiddlyWiki internationalization capabilities, here as issues or on the discussions groups. Such a macro should allow both developers to offer translations for their work (plugins, macros, themes...) and users to enable multi-languages wikis.

I made a working demonstrator called [i18n macro](https://sycom.github.io/TiddlyWiki-Plugins/#i18n%20macro) ([source code](https://framagit.org/sycom/TiddlyWiki-Plugins/blob/master/ext/modules/macros/i18n-macro.js)) that I'm using for my developments  (you know, _eat your own dog-food_) and that could solve a lot of cases listed below. **It's far from perfect** but can be a good start for the discussion about implementation for the core. 

@Jermolene already [said](https://github.com/Jermolene/TiddlyWiki5/pull/2722#issuecomment-521285516) (in #2722 discussion) he'd rather have it as _real_ macro than a javascript macro. There are a lot of other challenges to make it useful. I just have a few in my head right now that I put here. Please put yours : 

* how tiddlers content will still be searchable if their translations are within `$:/some_domain/i18n/lg-LG/some_title` tiddlers ?
* How to deal with titles for whole tiddlers translations. It's a major issue when languages are using non-Latin alphabet?
* maybe the macro doesn't have to solve all those, but can be used by plugins that will do more deep mods?

For people who need to down write ideas, I made a draft of reflexion around a, more radical, full i18n system in TiddlyWiki. That's on a CodiMD. Feel free to go there and make your edits: https://demo.codimd.org/ak7D5_2tQRmPt60zvbJkag

## References

* issue #786 : Plugins that expose strings should be translatable
* Issue #983 : Theme descriptions in ControlPanel/Appearence should be translatable
* issue #1372 : instructions for plugin language support
* issue #1513 : Language Switcher API
* issue #2597 : changing tiddlers in the story when changing the language
* issue [12](https://github.com/roma0104/gsd5/issues/12) in GSD5 : Language support
* discussions around the i18n macro
  * [with a potential Ukrainian user](https://groups.google.com/d/msg/tiddlywiki/ATHJzp0SitM/tPOIutn-AwAJ) who have needs around titles, timelines that imply also templates modifications
  * [with some users](https://groups.google.com/d/msg/tiddlywiki/h4UAjwbkdAk/7dd_A-cnBAAJ) about the best way to do plugins translations.
[...]

Issue #1011, opened 2014-10-26
Move tiddlywebadaptor and filesystemadaptor into the core


Jermolene:
The plugins `tiddlywiki/tiddlyweb` and `tiddlywiki/filesystem` currently need to be specified manually for wikis that need client-server functionality.

This is the reason that the core provides both a `tw5.com` edition and a `tw5.com-server` edition; the latter includes the former, but adds the two plugins.

For end users, it means that the instructions are complex for setting up a wiki that is to be used both to build standalone HTML files and serve over HTTP.

We could move the two plugins into the core to avoid this. They would add approximately 20KB to the size of the core.

Neither plugin currently has any means of configuration; if they are loaded in a wiki then they will attempt to perform their respective tasks.

We'd need to provide a means to disable the filesystem adaptor for wikis that shouldn't write their changes back to the file system. For example, the external images build process currently modifies image tiddlers to add the "_canonical_uri" field. We wouldn't want those modified images to be written back to the wiki folder. So, we could provide a new `config.writable` entry in `tiddlywiki.info` files that disables the filesystem adaptor when set to "no". Sadly, it would not be a backwards compatible change unless we perhaps used "no" as a default, and accepted the presence of `tiddlywiki/filesystem` in the plugins list as an alternative way to trigger "writable" mode.

The tiddlyweb adaptor would need to gain explicit configuration to tell it to attempt to sync to a particular server. Perhaps `$:/config/TiddlyWebServerURL` might be interpreted as the base URI to attempt to sync to; if the config tiddler were missing or blank, then no attempts to sync would be made.

Spangenhelm:
Hi @Jermolene since everyone does not need it, or not all the time, even though it is really lightweight i am against including it in the core but rather leave it as plugins, plus, with the new awesome library feature to find and install them it is easier than before.

Jermolene:
Hi @Spangenhelm the filesystemadaptor is 10KB and the tiddlywebadaptor is 15KB, so they are a very small proportion of the core plugin (currently 1.2MB). Annoyingly, the plugin library can't be used to install these components under Node.js; one has to edit the `tiddlywiki.info` file.

pmario:
I think the biggest win is, that we can get rid of the xxx-server editions. see OP

>This is the reason that the core provides both a tw5.com edition and a tw5.com-server edition; the latter includes the former, but adds the two plugins.

Spangenhelm:
[...]

PR #4373, opened 2019-11-14
Fix syncer to handler errors properly

Jermolene:
## Summary

The syncer ignores errors reported by the syncadaptor, leading to corruption of internal state and incorrect operation.

## Background

The core module [syncer.js](https://github.com/Jermolene/TiddlyWiki5/blob/165538180ef6a89a5f35ae9aa2d764e59f1f6191/core/modules/syncer.js) is responsible for synchronising changes between a wiki and a remote, asynchronous data store. It is used in several different contexts:

* in the browser to synchronise changes back to the server via the [tiddlywebadaptor](https://github.com/Jermolene/TiddlyWiki5/blob/165538180ef6a89a5f35ae9aa2d764e59f1f6191/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js)
* under Node.js to synchronise change changes back to the server via the [filesystemadaptor](https://github.com/Jermolene/TiddlyWiki5/blob/165538180ef6a89a5f35ae9aa2d764e59f1f6191/plugins/tiddlywiki/filesystem/filesystemadaptor.js)
* in the browser, by the [savetrail plugin](https://github.com/Jermolene/TiddlyWiki5/blob/165538180ef6a89a5f35ae9aa2d764e59f1f6191/plugins/tiddlywiki/savetrail/savetrail.js) to manage the download of backup tiddler files

The lack of error handling means that if a network or file system operation fails then user data can be lost: the red "dirty" indicator can incorrectly indicate that there are no changes to be synced when there are in fact changes that have not yet been seen by the server.

## Resolution

The obvious way to fix the problem would be to respond to errors by reverting the internal state of the syncer back to the point before the failed request was issued. The difficulties stem from the fact that the internal state of the syncer is represented as two separate task queues (one for upcoming tasks and one for tasks in progress) plus some per-tiddler tracking of the respective revisions in the browser and on the server. Properly responding to an error would involve putting the in progress task back in the upcoming queue, requiring complex logic for merging the two tasks.

However, it turns out that the current design may be much more complicated than needed. A simpler design would dispense with the queues, and instead use a simple scheduler that examines the current state of the store and the per-tiddler server tracking data, and determines the next highest priority synchronisation task to be performed.

## Status (Updated)

* [x] Synchronise system tiddlers from the server to the browser (previously only non-system tiddlers were synchronised)
* [x] Deletions of tiddlers on the server are now propagated to connected browsers
* [x] Throttling to prevent rapidly changing tiddlers being synchronised on every change
* [x] New extensible popup user interface providing easier access to server and synchronisation controls
* [x] Add a "filter" parameter to the `GET /recipes/default/tiddlers/tiddlers.json` route to return a filtered subset of tiddlers
* [x] Retry when syncadaptor errors occur (instead of failing)
* [x] Verify that tiddlers that change very rapidly on the browser or node side are handled correctly
* [x] Improved error reporting
* [x] Include skinny versions of lazily loaded tiddlers in the index.html (see [this comment](https://github.com/Jermolene/TiddlyWiki5/pull/4373#issuecomment-562214461))
* [x] Suppport for lazy loading
* [x] Support for the savetrails plugin
* [x] Review whether existing logging is too verbose/sparse


Jermolene:
We now track when tiddlers were last saved to the server, enabling us to throttle frequent changes to a tiddler by only saving to the remote data store if the tiddler hasn't been previously been saved in the last 1 second.

This is different than the old behaviour: previously, tiddlers were only saved if their last modification time was more than 1s ago, or their first modification time was more than 10s ago.

The upshot is that, say, typing in a draft without pause will now save changes to the server very 1s instead of every 10s. I think this logic is correct because if the user is typing quickly, we shouldn't wait 10s before transferring the data to the server. More experimentation is needed.

pmario:
There is 1 more thing, that bugs me since some time already. If a tiddler is edited and the saved, the server does 2 things. 

 - delete the draft tiddler
 - save the "new tiddler"

I think the order is wrong. If the second action goes wrong, the information is lost, even if the draft had been saved.

pmario:
>The upshot is that, say, typing in a draft without pause will now save changes to the server very 1s instead of every 10s. I think this logic is correct because if the user is typing quickly, we shouldn't wait 10s before transferring the data to the server. More experimentation is needed.

I'm not sure if 10s is right. Just running a server and it only needs max 3 seconds to by synced to the server.

Jermolene:
> I'm not sure if 10s is right. Just running a server and it only needs max 3 seconds to by synced to the server.

The 10s timeout only kicks in if the tiddler has been continuously modified at a rate of more than once per second for the entire 10s. So you'll see it you mash the keyboard for more than 10s, being careful not to pause for more than a few hundred milliseconds between each key.

Jermolene:
> There is 1 more thing, that bugs me since some time already. If a tiddler is edited and then saved, the server does 2 things. 

>  - delete the draft tiddler
>  - save the "new tiddler"

> I think the order is wrong. If the second action goes wrong, the information is lost, even if the draft had been saved. 

I agree that the ordering is wrong. However, it may not be easy to fix it because the server isn't really aware of the relationship between the two tiddlers.

(I think the best fix may be to support versioning on the server, which I am planning).

core/modules/syncer.js
```diff
@@ -138,7 +136,35 @@ Syncer.prototype.createTiddlerInfo = function(title) {
 Checks whether the wiki is dirty (ie the window shouldn't be closed)
 */
 Syncer.prototype.isDirty = function() {
-	return (this.numTasksInQueue() > 0) || (this.numTasksInProgress() > 0);
+	// Check tiddlers that are in the store and included in the filter function
+	var titles = this.filterFn.call(this.wiki);
+	for(var index=0; index<titles.length; index++) {
```
pmario: Is this efficient? `.isDirty` has been a really simple function. It is now replaced with 2 loops, that both iterate over the complete wiki store.

Jermolene: That is indeed a concern, and needs investigation. I'm hoping it may not be a problem: these loops are fast, pure JS and don't hit the DOM. The old implementation computed the same state, just at different times, and went to a lot more trouble to maintain that state, with relatively expensive hashmap insertions and deletions.

pmario:
Did just test your branch. ... The save / delete order is not a concern anymore. It seems you implemented a  "save first" priority.

pmario:
I did programmatically create 10000 tiddlers on the client and stopped the syncing after about 6100. ... 

It seems the syncing gets slower and slower over time. ... If I stop and restart the server, syncing will be fast again ... for some time. 

BUT I think this is a different problem and not really related to this PR.

pmario:
While syncing the tiddler, the whole time the UI response times where OK. So no real concern here.

pmario:
How do you set a PR as "draft"??

Jermolene:
Hi @pmario thanks for the feedback!

> Did just test your branch. ... The save / delete order is not a concern anymore. It seems you implemented a "save first" priority.

Doh! Yes, of course.

> It seems the syncing gets slower and slower over time. ... If I stop and restart the server, syncing will be fast again ... for some time.

Hmmm I wonder if there's a memory leak.

> How do you set a PR as "draft"??

image

https://github.blog/2019-02-14-introducing-draft-pull-requests/

Jermolene:
## Summary 

The upcoming commit addresses a couple of long-standing shortcomings with the client-server configuration:

- Changes to system tiddlers are not synchronised from the server to the browser
- Deletions of tiddlers on the server are not propagated to browser clients

It also provides a cleaner implementation for #4055 by @bimlas

## Syncing System Tiddlers

The reason that system tiddlers are not synced from the server back to the browser is that the `recipes/default/tiddlers.json` route only returns non-system tiddlers.

That can be fixed by adding a `filter` parameter to the `recipes/default/tiddlers.json` route, with a default of `[all[tiddlers]!is[system]sort[title]]` to match the current behaviour.

However, with that change there are several tiddlers that the browser immediately tries to load from the server because the local copies don't have server revision/bag data because they were baked into the HTML file via a special template:

* $:/library/sjcl.js
* $:/boot/bootprefix.js
* $:/boot/boot.js
* $:/boot/boot.css

To fix this problem we have to update the templates to include bag and revision fields for those tiddlers.

## Propagating Deletes from the Server to the Browser

When the syncer retrieves skinny tiddlers from the server, it should check for any tiddlers that were either baked into the HTML file or previously retrieved from the server that are no longer listed. Those tiddlers must have been deleted on the server, and so can be deleted locally.

Jermolene:
The last commit replaces the core save-wiki button with a popup menu that allows us to show more status information, and provide easier access to refresh and logout.

Cloud Sync

I've added new icons for synchronisation status, two of which are animated to attract attention:

Cloud Sync Icons

plugins/tiddlywiki/tiddlyweb/save-wiki-button.tid
```diff
@@ -0,0 +1,44 @@
+title: $:/core/ui/Buttons/save-wiki
+tags: $:/tags/PageControls
+caption: {{$:/core/images/cloud-sync}} Server status
+description: Status of synchronisation with server
+
```
pmario: It would be nice, if you would use `\whitespace trim` and indent the code properly. So it would be a 100 times more readable and I think much easier to understand. 

The second advantage would be, that there are no unwanted <p> tags, that are a horror for styling.

Jermolene: Hi @pmario I'll give it a go.

Jermolene:
I should also have added that the UI is a work in progress -- the plan is to swap the icon to indicate the sync status.

Also note that the animations are defined in the snow white theme, so won't be visible to users with the plain vanilla theme.

plugins/tiddlywiki/tiddlyweb/save-wiki-button.tid
```diff
@@ -3,52 +3,47 @@ tags: $:/tags/PageControls
...
-	</div>
+<div class="tc-drop-down">
+<$reveal state="$:/status/IsLoggedIn" type="nomatch" text="yes">
+<$button message="tm-login" class="tc-btn-invisible">
+{{$:/core/images/unlocked-padlock}} Login
+</$button>
+</$reveal>
+<$reveal state="$:/status/IsLoggedIn" type="match" text="yes">
+<p>
+Currently logged in<$reveal state="$:/status/UserName" type="nomatch" text="" default=""> as <$text text={{$:/status/UserName}}/></$reveal><$reveal state="$:/status/IsReadOnly" type="match" text="yes" default="no"> (read-only)</$reveal>
```
pmario: If you had `\whitespace trim` problems here, ... that's OK. This text needs to be made translateable in the future. 

There should be _no `&nbsp;` and such stuff. Elements, that need spacing need a class + proper margin or padding. This makes the content "themeable" We have a some problems because of "hacky" UI elements. 

IMO `\whitespace trim` just points out the problems, that should be fixed.

plugins/tiddlywiki/tiddlyweb/syncer-actions-copy-logs.tid
```diff
@@ -0,0 +1,6 @@
+title: $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/copy-logs
+tags: $:/tags/SyncerDropdown
+
+<$button message="tm-copy-syncer-logs-to-clipboard" class="tc-btn-invisible">
+{{$:/core/images/copy-clipboard}} Copy syncer logs to clipboard
```
pmario: I'm OK with those "snippets" for readability ... Just a reminder, that we need to translate this texts ;)

plugins/tiddlywiki/tiddlyweb/save-wiki-button.tid
Jermolene: > Elements, that need spacing need a class + proper margin or padding

Yes, quite. But sadly I think we're stuck with using space characters for spacing in lots of situations for backwards compatibility.

plugins/tiddlywiki/tiddlyweb/syncer-actions-copy-logs.tid
Jermolene: Yes indeed. We've got @sycom's #4179 which could address it. Equally, I still wonder whether #1011 might be a good idea.

pmario: about: #1011 ... I think these are 2 different elements of a server side middleware. `tiddlyweb` contains the tiddlyweb API abstraction for the client. filesystem contains the "storage API" below tiddlyweb api.

So it should be possible to exchange filesystem with an sql-like adapter and it should still work. The same may be true for a tiddlyweb / dat-fs combination.

What I want to say is, that I'm not convinced any-more, that the 2 elements should be combined. 

The problem is, that we would need much clearer API definitions to make it easier for devs to actually implement different types of middleware.

Jermolene: For clarity, the tiddlyweb and filesystem plugins are both instances of a "syncadaptor". The syncer that runs on the server uses the filesystem adaptor, and the syncer that runs in the browser uses the tiddlyweb adaptor. It is certainly possible to replace either syncadaptor with a different one.

If we integrated the tiddlyweb and filesystem plugins in the core we'd need some way to disable them so that users could choose to use different syncadaptors.

pmario:
about translations. I would be OK if the core plugin texts would be part of the core multids, with a little twist, that those snippets can be re-used by any plugin. 

I don't want to ship a plugin with 20+ translations, if only 1 text is needed. ... As a user I would go with plain english instead of wasting the space. 

---------------------

OT

I personally do have a real problem with TWs filesize. TWc empty.html is about 350kByte and TW5 is 2300kByte 

I know, this may be just me and every "crappy" jpg image has 2+times the size, but I do have a problem with it. It doesn't feel right. ... and this feeling doesn't go away. 

It will go away, once the dat-project is more mature, since I can split content from code, in a compatible way. dat is COW so every version from the past can be accessed.

sycom:
> about translations. I would be OK if the core plugin texts would be part of the core multids, with a little twist, that those snippets can be re-used by any plugin.
> I don't want to ship a plugin with 20+ translations, if only 1 text is needed. ... 

I DO understand this. And I mostly agree. BUT, I think that all your point of view stands here : 

> As a user I would go with plain english instead of wasting the space.

You are an English fluent and don't need translations. I'm not so fluent but can deal with a _no translation_ state. However, I think that we miss something if we are not able to address an internationalization issue for users who are not. With your proposal  : 

* We would only able to integrate core plugins translations to multids, or it will become a nightmare,
* We are about to increase translation multids size with plugins translations that most users won't use
* We are still missing a translation solution for community plugins?

To combine both we may need a more "modular" translation system: user can load the translations he needs from the plugin. Or at least the plugin gives a link to get them. The _downside_ is:

* users have to accomplish this manually for each plugin unless some part of the translations are in the core multids
* devs have to make separate bundles / multids for translations.

Bottom line: the purpose of i18n macro was also to make the translation process more friendly and modular. I don't find multids are, and I have no experience with bundles, so I can't say how difficult it is to produce one... Maybe other (more complex) solutions are possible with a global option for automatic plugins translations option loading and a load those requested through an XMLHttpRequest for a JSON tiddler?

_comment: drop here a reference to #4179 for keeping consistency_

plugins/tiddlywiki/tiddlyweb/icon-cloud.tid
```diff
@@ -0,0 +1,11 @@
+title: $:/plugins/tiddlywiki/tiddlyweb/icon/cloud
+tags: $:/tags/Image
+
+<svg class="tc-image-cloud tc-image-button" width="22pt" height="22pt" viewBox="0 0 128 128">
```
sycom: I think this one can be much lighter: 951 char vs 2562

```
<svg xmlns="http://www.w3.org/2000/svg" class="tc-image-cloud tc-image-button" width="22pt" height="22pt" viewBox="0 0 128 128">
<g stroke="none" fill-rule="nonzero">
  <path d="M24 103a24 24 0 01-9.97-45.84 24 24 0 0128-24.83A36 36 0 01106 55v.1a24 24 0 011 47.72v.19H24z"/>
  <path class="tc-image-cloud-idle" d="M57.93 84.7a6 6 0 01-8.49 0L35.3 70.56a6 6 0 118.49-8.49l9.9 9.9L81.97 43.7a6 6 0 018.49 8.48L57.93 84.7z"/>
  <path class="tc-image-cloud-progress tc-animate-rotate-slow" d="M44.8 40a3.6 3.6 0 100 7.2h2.06A24 24 0 0063.6 88h.4v-7.2a16.8 16.8 0 01-16.8-16.48V64c0-5.17 2.33-9.79 6-12.87v4.47a3.6 3.6 0 107.2 0v-12a3.6 3.6 0 00-3.6-3.6h-12zM64 40v7.2a16.8 16.8 0 0110.8 29.67V72.4a3.6 3.6 0 10-7.2 0v12a3.6 3.6 0 003.6 3.6h12a3.6 3.6 0 100-7.2h-2.06A24 24 0 0064 40z"/>
  <path class="tc-image-cloud-error tc-animate-pulse-fast" d="M64 84a6 6 0 110 12 6 6 0 010-12zm0-52a6 6 0 016 6v34a6 6 0 11-12 0V38a6 6 0 016-6z"/>
</g>
</svg>
```

Jermolene:
Let's discuss the plugin translation issues over at #4179, and feel free to open a new ticket for discussion of reducing the core size (I believe there is a lot of "low hanging fruit", such as optimising the SVGs).

Jermolene:
Hi @sycom 

> give SVG a "diet" ;-) ?

I've now done a bit of work on this, see #4379.

inmysocks:
I do not know how practical it would be to resolve this without using an alternate syncer like Bob does, but the principle issue that I have run into while making Bob and during experiments with different database back-ends is asynchronous responses.

For Bob I had to give up on writing a normal syncadaptor for the browser and wrote a completely new syncer module because the responses for success or failure on the server side are completely asynchronous and come back separately from the browser request in the sense that it isn't in the response text of the request.

The database interfaces I have used can often use callbacks and can work, although I have run into some that only return promises.

But using websockets with Bob or server sent events the response from the server is completely unconnected to the task.run() (or dispatchTask in the current version) so calling the callback when the function returns is premature and can't recover from server errors or dropped packets using websockets or server sent events.

I don't know if there is a better option than creating an alternate syncer module. I have been working on making my very hacky syncer for Bob cleaner and hopefully more general purpose. I don't know if it could ever be compatible with the core syncer.

This is off-topic is this thread.
I ran into a very similar problem with the BobSaver. The save function returns a boolean success or failure flag immediately without waiting for the callback. Waiting for a response to an asynchronous http request either makes it impossible to save if the server isn't available because it always indicates success, or it always tries to save using the next saving method as well because it always indicates failure. The same is true for the canSave function in the saver handler.

Jermolene:
Thanks @inmysocks for the feedback. I'd like to step through it to make sure I've understood everything.

> I do not know how practical it would be to resolve this without using an alternate syncer like Bob does,

What do you mean by "using an alternate syncer"? Does it mean something different than "improving the syncer implementation"? I'm confused because the whole point of this PR is that it's a rewrite of the syncer

> but the principle issue that I have run into while making Bob and during experiments with different database back-ends is asynchronous responses.

I'm confused here because the existing syncadaptor API is completely asynchronous.

> For Bob I had to give up on writing a normal syncadaptor for the browser and wrote a completely new syncer module because the responses for success or failure on the server side are completely asynchronous and come back separately from the browser request in the sense that it isn't in the response text of the request.

It would be helpful to unpack that a bit more. The syncadaptor APIs are asynchronous and I'm struggling to understand how there could be a type of asynchronicity that they don't support.

> The database interfaces I have used can often use callbacks and can work, although I have run into some that only return promises.

Right, but unless we're talking at cross purposes it's trivial to convert a promise-based API to use callbacks.

> But using websockets with Bob or server sent events the response from the server is completely unconnected to the task.run() (or dispatchTask in the current version)

Here I think you're saying that the current syncer relies on polling to get changes from the server, but that your implementation uses websockets, and so changes arrive asynchronously without polling.

We can refactor the syncer so that there is a convenient method for you to call when you've got an updated tiddler to deliver.

> so calling the callback when the function returns is premature and can't recover from server errors or dropped packets using websockets or server sent events.

Again, it's hard to visualise this without some example code.

> I don't know if there is a better option than creating an alternate syncer module. I have been working on making my very hacky syncer for Bob cleaner and hopefully more general purpose. I don't know if it could ever be compatible with the core syncer.

So, I think you're saying that you'd like the syncer to support pushing new tiddlers from the server, instead of having to poll for them. That's perfectly reasonable but I'm still confused by this talk of an "alternate syncer".

This is off-topic is this thread.
I ran into a very similar problem with the BobSaver. The save function returns a boolean success or failure flag immediately without waiting for the callback. Waiting for a response to an asynchronous http request either makes it impossible to save if the server isn't available because it always indicates success, or it always tries to save using the next saving method as well because it always indicates failure. The same is true for the canSave function in the saver handler.

Jermolene:
Hi @inmysocks apologies I pressed return too quickly:

>This is off-topic is this thread.
> I ran into a very similar problem with the BobSaver. The save function returns a boolean success or failure flag immediately without waiting for the callback. Waiting for a response to an asynchronous http request either makes it impossible to save if the server isn't available because it always indicates success, or it always tries to save using the next saving method as well because it always indicates failure. The same is true for the canSave function in the saver handler.

I think here your pointing out that synchronous APIs can't call asynchronous APIs. That's true, and the solution is to refactor the first API to be asynchronous. I think that's perfectly feasible in the case of the saver API.

inmysocks:
@Jermolene playing with this is showing just how bad I am with some parts of javascript, I am quite happy to hear that I am wrong about what is going on. One would think that after almost a decade of doing development professionally I would have a better grasp of the language. I am always a bit bemused by that.

As I was typing out an explanation I realised that I should be able to convert the message queues into event emitters and each time an acknowledgement is received from the server it could check and see if the id matches and call the callback if it does. The logic shoudln't be any much more complicated the the queue handling I have now.

Asynchronous was the wrong word, unconnected or independent may be what I was thinking of and using a event emitters should take care of that problem.

Doing it this way I may be able to make the browser component of Bob almost entirely contained in a syncadaptor module.

Jermolene:
Hi @inmysocks 

> @Jermolene playing with this is showing just how bad I am with some parts of javascript, I am quite happy to hear that I am wrong about what is going on. One would think that after almost a decade of doing development professionally I would have a better grasp of the language. I am always a bit bemused by that.

Ha! Your many achievements speak otherwise. Plus there's the phenomenon whereby the more you know the more you realise you don't know.

> As I was typing out an explanation I realised that I should be able to convert the message queues into event emitters and each time an acknowledgement is received from the server it could check and see if the id matches and call the callback if it does. The logic shoudln't be any much more complicated the the queue handling I have now.

Fantastic! It's instructive how writing things out can clarify ones thinking

> Asynchronous was the wrong word, unconnected or independent may be what I was thinking of and using a event emitters should take care of that problem.

That makes sense.

> Doing it this way I may be able to make the browser component of Bob almost entirely contained in a syncadaptor module.

Perfect, do let me know if you run into anything; this is a great time to improve  the API if we need to.

Jermolene:
I'll shortly commit some improvements to the handling of lazy loading. Here's the background.

The current `./bin/lazy.sh` serves using the template $:/core/save/lazy-images that simply omits any image tiddlers from the tiddlers included in the index.html. Then the syncer will retrieve skinny versions of those images just after startup, and subsequently retrieve the body of those images if they are referenced (ie the `lazyLoad` event is triggered on the wiki object).

None of this is particularly optimal. What we'd really like is for the skinny versions of the images to be baked into the initial index.html and then for the bodies to be retrieved after startup.

So, we need to not exclude the images but just to make them skinny. That means changing $:/core/templates/store.area.template.html where we currently have:

```
`<div id="storeArea" style="display:none;">`
<$list filter=<<saveTiddlerFilter>> template="$:/core/templates/html-div-tiddler"/>
`</div>`
```

Which we must change to:

```
`<div id="storeArea" style="display:none;">`
<$list filter=<<saveTiddlerFilter>> template="$:/core/templates/html-div-tiddler"/>
<$list filter=<<skinnySaveTiddlerFilter>> template="$:/core/templates/html-div-skinny-tiddler"/>
`</div>`
```

And then back up in $:/core/save/lazy-images we simply need to set up `skinnySaveTiddlerFilter`.

The catch is that that new list filter will return all tiddlers if the variable `skinnySaveTiddlerFilter` is not defined. We can get around that with an alternative construction:

```
`<div id="storeArea" style="display:none;">`
<$list filter=<<saveTiddlerFilter>> template="$:/core/templates/html-div-tiddler"/>
<$list filter={{{ [<skinnySaveTiddlerFilter>] }}} template="$:/core/templates/html-div-skinny-tiddler"/>
`</div>`
```

Jermolene:
The next commit is the last part of making lazy loading work again. We fix it by introducing a new `_is_skinny` field that, if present, indicates that a tiddler is subject to lazy loading. Previously, we'd assumed that any tiddler without a text field must be skinny, resulting in extraneous loads from the server.

Jermolene:
The new design is designed to tolerate network failures. For testing, I've been using Chrome Dev Tools "Network" tab where one can force the connection to be online, offline or a specific speed/latency.

Taking the network offline and then (say) creating a new tiddler will cause errors, which appear as prominent yellow alert boxes. The cloud indicator will spin red as soon as any attempt is made to change the wiki. Once the network is restored, the syncer automatically retries the failed operation(s), and eventually the cloud returns to its ticked state. However, the alerts will still be visible. The next commit addresses this by automatically removing any outstanding alerts when synchronisation has been restored.

Jermolene:
The next commits make some minor tweaks to the appearance of alerts:

* Changed close icon from trash can to cancel button
* Tightened margins
* Reformatted date stamp to use language specific format


Old:

image

New:

image

Jermolene:
The next commits address two problems with the sync adaptor API:

Firstly, the current sync adaptor API requires that tiddlers have a "revision" field that contains an identifier provided by the remote server identifying the revision of the tiddler known on the server. (The TiddlyWeb sync adaptor also users a "bag" field containing the bag identifier provided by the remote server for a tiddler, but the bag handling is encapsulated within the sync adaptor, and doesn't impact the sync adaptor API).

Secondly, the current sync adaptor API uses the "getSkinnyTiddlers()" method to determine if there are any updates that need to be loaded from the server. The syncer itself handles the logic of examining tiddler revision values to determine if a particular tiddler is new.

It is proposed to refactor two aspects of the current implementation so that the revision data can be stored elsewhere (rather than disturbing the users' tiddlers with an additional field), and to enable the sync adaptor to use its own logic for obtaining updates from the server. These changes are intended to retain backwards compatibility.

The first step is to refactor things so that the sync adaptor is invoked to extract the revision for a tiddler, rather than pulling it straight out of the "revision" field. If the sync adaptor doesn't support the appropriate method then the syncer falls back to using the "revision" field.

The second step is to add support for a new "getRemoteUpdates()" method that is used in preference to the existing "getSkinnyTiddlers()" method. The new method performs all the required logic to identify tiddlers that have been modified or deleted on the server. The older method will still be supported for compatibility with existing sync adaptors.

core/modules/syncer.js
```diff
@@ -64,23 +64,27 @@ function Syncer(options) {
 	this.filterFn = this.wiki.compileFilter(this.wiki.getTiddlerText(this.titleSyncFilter));
 	// Record information for known tiddlers
 	this.readTiddlerInfo();
+	this.titlesToBeLoaded = {}; // Hashmap of titles of tiddlers that need loading from the server
+	this.titlesHaveBeenLazyLoaded = {}; // Hashmap of titles of tiddlers that have already been lazily loaded from the server
+	// Timers
 	this.taskTimerId = null; // Timer for task dispatch
 	this.pollTimerId = null; // Timer for polling server
+	// Number of outstanding requests
 	this.numTasksInProgress = 0;
```
pmario: Why don't you rename the variable to `numOpenRequests` or something similar?

core/modules/syncer.js
```diff
@@ -212,17 +209,15 @@ Syncer.prototype.updateDirtyStatus = function() {
...
 */
-Syncer.prototype.storeTiddler = function(tiddlerFields,hasBeenLazyLoaded) {
+Syncer.prototype.storeTiddler = function(tiddlerFields) {
 	// Save the tiddler
 	var tiddler = new $tw.Tiddler(tiddlerFields);
 	this.wiki.addTiddler(tiddler);
 	// Save the tiddler revision and changeCount details
 	this.tiddlerInfo[tiddlerFields.title] = {
-		revision: tiddlerFields.revision,
+		revision: (self.syncadaptor && self.syncadaptor.getTiddlerRevision && self.syncadaptor.getTiddlerRevision(tiddlerFields.title)) || tiddlerFields.revision,
```
pmario: This line is ugly and it is used several times. So can we create a function for this one too. similar to `getSyncedTiddlers`

core/modules/syncer.js
```diff
@@ -129,47 +133,40 @@ function Syncer(options) {
...
 	this.tiddlerInfo = {};
 	// Record information for known tiddlers
 	var self = this,
-		tiddlers = this.filterFn.call(this.wiki);
+		tiddlers = this.getSyncedTiddlers();
 	$tw.utils.each(tiddlers,function(title) {
 		var tiddler = self.wiki.getTiddler(title);
 		self.tiddlerInfo[title] = {
-			revision: tiddler.fields.revision,
+			revision: (self.syncadaptor && self.syncadaptor.getTiddlerRevision && self.syncadaptor.getTiddlerRevision(tiddler)) || tiddler.fields.revision,
```
pmario: hmmm .. `self` is local to the function in line 218 self is "global"

core/modules/syncer.js
```diff
@@ -129,47 +133,40 @@ function Syncer(options) {
...
 /*
 Read (or re-read) the latest tiddler info from the store
 */
 Syncer.prototype.readTiddlerInfo = function() {
-	// Hashmap by title of {revision:,changeCount:,adaptorInfo:,needsToBeLoaded:,hasBeenLazyLoaded:}
+	// Hashmap by title of {revision:,changeCount:,adaptorInfo:}
+	// "revision" is the revision of the tiddler last seen on the server, and "changecount" is the corresponding local changecount
 	this.tiddlerInfo = {};
 	// Record information for known tiddlers
 	var self = this,
```
pmario: `var self` is also defined at the prototype root. ... Then it is redefined in the functions. This makes the code really hard to read and error-prone.

core/modules/syncer.js
Jermolene: Good thinking

core/modules/syncer.js
Jermolene: Good thinking

core/modules/syncer.js
Jermolene: Ouch!

core/modules/syncer.js
Jermolene: Ouch! Where's the definition at the prototype root? Do you mean line 39? That variable is only local to the constructor.

core/modules/syncer.js
Jermolene: Actually, the code uses "task" quite a few times to refer to the CRUD transactions, so I think it's consistent enough.

Jermolene:
I'm going to merge this PR in the next few days. These are substantial changes so any further feedback is very welcome, particularly from anyone who maintains an alternate syncadaptor. 

I tried to use GitHub search to cc people who have recently worked on syncadaptors for TiddlyWiki. Please forgive me if I've included you in this list in error.

cc. @danielo515 @Arlen22 @linonetwo @fiatjaf @jimpick@cjxgm @ProgrammingIncluded @iilyak @noahlange @neumark @apalmer27516

Jermolene:
Updated to add a space between @jimpick and @cjxgm

Jermolene:
The next commits attempt to present network connection errors more clearly for non-technical users.

Previously, alerts were shown with technical details for each error:

image

With the upcoming changes, detailed technical information is logged and a single generic alert is displayed:

image

(Logs are also available via the cloud sync dropdown).

core/modules/syncer.js
pmario: ok

apalmer27516:
My project was stalled waiting for the backend And my own availability anyway. The fixes/enhancements look good. I’m all for versioning support in the future too. Thanks for the heads up!
—Abraham

Jermolene:
Many thanks @apalmer27516!

core/modules/syncer.js
Arlen22: The situation in https://github.com/Jermolene/TiddlyWiki5/pull/4397 and https://github.com/Jermolene/TiddlyWiki5/issues/4275 involved an Array.indexOf call while iterating through the array, which resulted in length squared indexOf calls. That only became a problem once someone created 60,000 tiddlers, which resulted in 3.6 billion indexOf calls.

Jermolene:
Hi @Arlen22 that makes sense. Perhaps we might have an optional second filter that is configured on the server, and applied after the client provided filter. It would be a much cleaner way to handle the exclusions currently hardcoded in `TiddlyWebAdaptor.prototype.getSkinnyTiddlers()`.

Jermolene:
Hi @Arlen22 just checking: is the DOS attack you're concerned with that the attacker can craft a filter expression that is so complex that it crashes the server? Or is this about restricting access to certain tiddlers (e.g. system tiddlers)?

plugins/tiddlywiki/tiddlyweb/icon-cloud.tid
Jermolene: Thanks @sycom and apologies for the delay. I added the compression here: aaa3b34d2

Jermolene:
Hi @Arlen22 reading your comment more carefully, I see you are concerned about malformed filters. I wonder whether we might handle this by adding an optional configuration setting of a list of  allowed filter expressions?

Jermolene:
Hi @Arlen22 let me know if c9d6f501b7dd0513643d11d95743264309cb5119 addresses your concerns

Arlen22:
> Hi @Arlen22 let me know if c9d6f50 addresses your concerns. 

Yes, it seems to take care of it completely, although I am wondering how we would know if there is a filter expression being blocked that we need to add to the array. 

I was also thinking that it would be fine for each plugin to tag the filters it uses with `$:/tags/get-tiddlers-json-filter` or something similar, which the get-tiddlers-json route would compile the array from. It's more the fact that any public stranger can execute any filter they want on the server. Server code is trusted to start with, so it's fine to let it whitelist its own filters, I think. 

I keep debating whether we should allow named filters, but I think thats too confusing. Plus, we can implement caching more efficiently if we keep the text of the filter in the URL, because if the store hasn't changed, the response can be served straight from cache, and more than one plugin using the same filter can share that benefit, whereas aliases would isolate them.

Arlen22:
Also, I think that should be a 403, because 401 usually indicates the user needs to log in, which is not the case here.

Arlen22:
Also, are you aware that the text field could be retrieved with get-tiddlers-json? In fact, I should be able to get all fields by specifying a single comma in the exclude query, which would split to an array of two empty strings. 

I'm sure that's fine, but just wanted to point it out. I don't have a problem with it.

editions/tw5.com/tiddlers/webserver/WebServer API_ Get All Tiddlers.tid
```diff
@@ -15,8 +15,19 @@ Parameters:
 * ''filter'' - filter identifying tiddlers to be returned (optional, defaults to "[all[tiddlers]!is[system]sort[title]]")
 * ''exclude'' - comma delimited list of fields to excluded from the returned tiddlers (optional, defaults to "text")
 
-Response:
+An optional list of permitted filters may be configured in [[tiddlywiki.info|tiddlywiki.info Files]] in the config setting `server-get-tiddlers-allowed-filters`. If the setting is missing then all filters are permitted. For example, the following setting explicitly permits the filter used by the TiddlyWeb adaptor plugin:
```
pmario: I think security should be default. Extended rights should be optional.
just a thought

Jermolene: Hi @pmario that's what I've been thinking overnight too. I'll address it shortly.

Jermolene:
Hi @Arlen22 it's only a rough guideline but for this kind of server config I tend to stick things in tiddlywiki.info rather than encoding them as tiddlers because it avoids the problem of building a UI.

Named filters are an intriguing idea here, but I do tend to agree that it may be more complex than we need right now.

> Also, I think that should be a 403, because 401 usually indicates the user needs to log in, which is not the case here.

Thanks, fix upcoming.

> Also, are you aware that the text field could be retrieved with get-tiddlers-json? In fact, I should be able to get all fields by specifying a single comma in the exclude query, which would split to an array of two empty strings.

Yup, my thinking was that sometimes it might be useful and practical to read the entire tiddler.

Jermolene:
Hi @Arlen22 

> although I am wondering how we would know if there is a filter expression being blocked that we need to add to the array.

I'll add some logging.

Jermolene:
The commit 0f1a08d5fc3677eed2ff278a2c24038cd6c43204 makes it so that the default configuration is restricted to two allowed filters. It also adds a new config option to allow all filters to be accepted.

Arlen22:
> ...for this kind of server config I tend to stick things in tiddlywiki.info rather than encoding them as tiddlers because it avoids the problem of building a UI.

I don't think a UI would be needed here, because if someone is making HTTP requests from Javascript, they can definitely create a tiddler with a tag. On the other hand, having it in the tiddlywiki.info file works well with the concept of client data folders, which people prefer for low memory environments. 

I wish there was a way for plugins to whitelist the filters they use, since I'm not really worried about plugins and it would make things a lot easier. Can we allow plugins to whitelist their filters in plugin.info? But if we're loading plugin.info, we're loading the rest of the plugin too so then the tag idea would work. 

Besides there are already server-side filters which can be set from the browser using TiddlyWeb, so that's not the security issue we're trying to solve -- we are just preventing GET requests from being able to execute arbitrary filters, since method restriction is the current way to make a wiki readonly to the public.

Arlen22:
> On the other hand, having it in the tiddlywiki.info file works well with the concept of client data folders, which people prefer for low memory environments.

Actually, this doesn't matter because accessing get-tiddlers-json implies a fully loaded server-side wiki. A client-side loaded wiki would not need to call get-tiddlers-json anyway because it has the full data.

Jermolene:
Hi @Arlen22 a downside with the tag-based approach is that each call to `GET /recipes/default/tiddlers/tiddlers.json` would need to read through all the tiddlers with that tag looking for the requested filter string. Instead, we could just look for a single tiddler titled with the filter text prefixed with `$:/config/Server/ExternalFilters/`. I think your use case would still work.

Arlen22:
> Hi @Arlen22 a downside with the tag-based approach is that each call to `GET /recipes/default/tiddlers/tiddlers.json` would need to read through all the tiddlers with that tag looking for the requested filter string. Instead, we could just look for a single tiddler titled with the filter text prefixed with `$:/config/Server/ExternalFilters/`. I think your use case would still work.

That will work great. I didn't think about putting the filter in the title. It will make it easier to process the list on the server-side when dealing with client-side data folders.

xcazin:
Hi @Jermolene,

I just tried this branch for a few hours, with my favorite setup: tiddlywiki servers running on my termux-powered phone, serving wikis over wifi. Simply editing a tiddler with the phone held far from the client browser surely triggered the red cloud and the yellow alert asking to check connectivity . And bingo, once the phone got closer to the workstation, everything got synced smoothfully. A much better experience than before.

I also tried several times to save then edit a tiddler, in order to reproduce the worst error I had seen from time to time: the text content getting empty on editing, leaving you with the only safe option to cancel the editing. I couldn't reproduce this with the fix-syncer branch, so that's another relief.

The only issue that I noticed is the favicon not appearing on my browser tab until after I load the image tiddler explicitly. It didn't happen with a fresh wiki served directly from node to the 0.0.0.0 interface, but with another one served through nginx as a proxy. I don't now if its related to this branch or another recent change.

Jermolene:
Thanks @xcazin.

> The only issue that I noticed is the favicon not appearing on my browser tab until after I load the image tiddler explicitly. It didn't happen with a fresh wiki served directly from node to the 0.0.0.0 interface, but with another one served through nginx as a proxy. I don't now if its related to this branch or another recent change.

I wasn't able to duplicate that problem with the default configuration, or with lazy loading enabled.

xcazin:
Hi @Jermolene,
 
> > The only issue that I noticed is the favicon not appearing on my browser tab until after I load the image tiddler explicitly. It didn't happen with a fresh wiki served directly from node to the 0.0.0.0 interface, but with another one served through nginx as a proxy. I don't now if its related to this branch or another recent change.
> 
> I wasn't able to duplicate that problem with the default configuration, or with lazy loading enabled.

Apologies, here was the actual command line ($:/core/save/lazy-images is needed because of my phone limited memory):

`node ~/src/TW5-fix-syncer/tiddlywiki.js /data/data/com.termux/files/home/wikis/mobile --verbose --listen port=1803 root-tiddler=$:/core/save/lazy-images username= password= host=127.0.0.1`

It looks like with lazy-images, the favicon is indeed lazily loaded with the fix-syncer branch, while it **was not** with master. Which sounds like an additional fix with respect to my previous experience, although I'd probably wish that the favicon was the _only_ image preloaded :-)

Jermolene:
Thanks @xcazin that's encouraging!

Arlen22:
Looking forward to this getting completed. It's "one giant leap" for TiddlyWiki as it will allow consistent tiddler syncing and the implementation of multi-user editing. I've already written a Server Sent Events handler to notify the browser of changes, and this PR will close the gap by making syncing consistent. Looking forward to online multi-user collaboration.

kookma:
@Jermolene 
How is the status of Syncer? Is there any chance to have new features in TW 5.1.22?

kookma:
@Jermolene 
These are my findings during testing new features of Fix-Syncer branch
- When the layout is set to Fluid Story-Fixed Sidebar, on click the synchronization icon reveals a local window below icon which is partly cross the right browser edge (border). TW shall be enough smart to show the local windows bottom right or bottom left based on the space avilable
- When I create a plugin library using the **master branch**, I use the standard plugin library edition and I use this command

```
"build": {
		"library": [
			"--makelibrary","$:/UpgradeLibrary",
   			"--savelibrarytiddlers","$:/UpgradeLibrary","[prefix[$:/]] +[prefix[$:/plugins/kookma]]","recipes/library/tiddlers/","$:/UpgradeLibrary/List",
   			"--savetiddler","$:/UpgradeLibrary/List","recipes/library/tiddlers.json",
			"--rendertiddler","$:/plugins/tiddlywiki/pluginlibrary/library.template.html","index.html","text/plain"]
	}
```

note to `"--savelibrarytiddlers","$:/UpgradeLibrary","[prefix[$:/]] +[prefix[$:/plugins/kookma]]","recipes/library/tiddlers/","$:/UpgradeLibrary/List"`

Repeating this using Fix-Syncer branch Tiddlywiki produces some warning like below

```
Warning: missing plugin.info file in d:\Tiddlywiki\plugins\relink\js
Warning: missing plugin.info file in d:\Tiddlywiki\plugins\relink\tiddlers
 syncer-server-filesystem: Dispatching 'save' task: $:/StoryList
```

I expect no warning, as I excluded those pluging in making the library


* **Suggestion**
Some settings affects the Sync server like

 - $:/config/FileSystemPaths
-  $:/config/SaverFilter
-  PublishFilter variable

are hidden or not in close access so, I may suggest to have a setting tab or a button on syncer local window (when sync server icon clicked) to open the setting tab/tiddler. In other words have all syncer settings in one place to be able view/edit them

More feedback comes soon

Jermolene:
Thanks @kookma 

The first issue with the visibility of the syncer page control button dropdown, is it any different than the behaviour seen with other page control buttons that feature a dropdown (e.g. the "more" button)? It sounds like a problem with pop up positioning, and unless there's a new problem introduced by this branch we should just file a separate GH issue.

As to the second issue with the plugin.info warning, can you confirm that you don't get that warning when using the master branch?

Your third point of an enhanced settings panel is a good idea but I won't delay this PR for it.

kookma:
> Thanks @kookma
> 
> The first issue with the visibility of the syncer page control button dropdown, is it any different than the behaviour seen with other page control buttons that feature a dropdown (e.g. the "more" button)? It sounds like a problem with pop up positioning, and unless there's a new problem introduced by this branch we should just file a separate GH issue.

Yes, it is because pop up positioning issue and not specific to fix-syncer branch
> 
> As to the second issue with the plugin.info warning, can you confirm that you don't get that warning when using the master branch?
Sorry, you are right! The warning is appeared with latest update of master branch
> 
> Your third point of an enhanced settings panel is a good idea but I won't delay this PR for it.
Agree! Hope this be added in 5.1.23

joshuafontany:
> Thanks @kookma
> 
> The first issue with the visibility of the syncer page control button dropdown, is it any different than the behaviour seen with other page control buttons that feature a dropdown (e.g. the "more" button)? It sounds like a problem with pop up positioning, and unless there's a new problem introduced by this branch we should just file a separate GH issue.
> 

This was noticed by the author of the "Mentat" plugin that turns TW5 into a desktop/moveable-dashboard type setup. See:

https://github.com/theSherwood/Mentat/blob/master/src/tiddlers/system/plugins/admls/repopup/startup/repopup.js

https://github.com/theSherwood/Mentat
> Plugins
> Repopup
> 
> A quick hack of a plugin that attempts to reposition tiddlywiki popups so that they don't run off the screen. It's not perfect and it doesn't seem to do much to normal dropdowns. But it's an okay fix until the popup mechanism can get a comprehensive overhaul. This plugin has been designed with the Volant, Engine, and Mentat Plugins in mind. So if you use it for something else, your mileage may vary.

kookma:
@joshuafontany 
Thanks for pointing me to Mentat.

inmysocks:
I finally got around to testing this with Bob and I am getting some strange errors. I do not know if the syncer I wrote for Bob is the problem, but it works with the prerelease branch master, but not on this branch.

- First is that somehow the number of tasks in progress goes below 0 which freezes the syncer
  - as a hack I tried adding a check to the processTaskQueue function that sets the number of tasks in progress to 0 if it is less than 0, this makes things work but I don't think it addresses the problem.
- Second is after editing fields directly, like setting the site title in the control panel, the dirty indicator never goes away unless I make some other edit that gets saved using the normal tiddler editing

As a slightly related thing, which setting changes how quickly the queue is processed?
I have run into a problem where saving hundreds or thousands of new tiddlers can take a long time, over 15 minutes last time I tired.
When Bob was bypassing the syncer module the websocket connection had a much higher message rate and it would take a few seconds to save the same number of tiddlers.

Jermolene:
Thanks @inmysocks. Does Bob entirely replace the core syncer, or does it run its own syncer alongside? The interface to the sync adaptor has only minor changes which are intended to be backward compatible.

> somehow the number of tasks in progress goes below 0

That seems significant. I can't see a code path where that could happen with the normal operation of the updated syncer.

> which setting changes how quickly the queue is processed

If there's a bunch of saves queued up to process then they will happen at 1s intervals by default. It's controlled by the `taskTimerInterval` option on the syncer constructor.

inmysocks:
In the last update to Bob I started using the core syncer and made a sync adaptor that works with it, before that the syncing was completely separate from the built-in syncer.

I found the problem, if a tiddler wasn't supposed to sync there was a missing return so it would call the call back and then try and sync anyway so the callback could be called more than once.
So that was a problem with my syncer and should be fixed now. The question becomes how did it work before, but that isn't something I am going to dig into.

And thank you, changing the `taskTimerInterval` was what I wanted. I will make a pull request to add a configuration setting for it.

I haven't been very active over the winter so I need to catch up on the changes.

The old sync method for Bob used ack messages from the server to maintain consistent state between browser and server and to trigger the next sync task. Bob also uses tiddler hashes to determine changes to tiddlers and for determining what has changed. The stable stringily and hashing function together are pretty quick and about 100 lines of code. I don't know if that is an appropriate method for most syncer types. I will see if I can find a way to allow that without too much change to the syncer. It helped a lot with intermittent connections to the server.

Jermolene:
Hi @inmysocks that's great, glad to hear it's working. My thinking for this update was to go as far as I could without changing the core wiki store or the syncadaptor API. For the future, I'm interested in exploring other representations of the wiki store which will are easier to sync (specifically, an append-only log of some kind).

inmysocks:
After looking at my code a bit more it turns out that both of those things are (or at least can be) outside the scope of the syncer.
Figuring out how to wait for the ack response was what took me so long to make it. I should get the same behaviour as before by just setting the taskTimerInterval to 0 and the server throttles the syncing to an appropriate rate with the ack messages. Using the hashes to check if something should sync is done in the saveTiddler and deleteTiddler functions of the syncadaptor, so that isn't a problem either.

I am continually surprised and impressed by how flexible the core of tiddlywiki is.

Jermolene:
Thanks @inmysocks that's very encouraging, the main risk with this work was breaking things like Bob, so it's kind of cool if the upshot is that we've nailed another bug :smile:

kookma:
@Jermolene 
It seems I found an issue. To reproduce

1. create a new wiki using tiddlywiki mywiki --init server
2. start the new wiki using tiddlywiki mywiki --server
3. open $:/tags/EditToolbar and edit its list filed, for example put $:/core/ui/Buttons/delete at the end of list
4. delete $:/tags/EditToolbar
5. the cloud icon (syncer) turns red and rotating and never stop...
6. refresh the wiki, you see $:/tags/EditToolbar has not been deleted (one expect to have $:/tags/EditToolbar deleted and the original $:/tags/EditToolbar tiddler restored from core, but the modified one is still there)

Jermolene:
Thanks @kookma that's very helpful, I'll commit a fix shortly.

Jermolene:
Thanks to everyone for your help with this. I'm going to merge it now to help flush out any remaining issues.

kookma:
@Jermolene 
Thank you
For test, I appreciate to update the fix-syncer branch also.

Jermolene:
Hi @kookma these changes are merged into "master", so you should conduct further testing on that branch. Many thanks!

kookma:
Sure, Thank you @Jermolene

core/modules/syncer.js
```diff
@@ -358,189 +453,180 @@ Syncer.prototype.handleRefreshEvent = function() {
...
-				}
-				// Adjust the info stored about this tiddler
-				self.tiddlerInfo[task.title] = {
-					changeCount: changeCount,
-					adaptorInfo: adaptorInfo,
-					revision: revision
-				};
-				// Invoke the callback
-				callback(null);
-			},{
```
danielo515, line 507: This is a breaking change/regression compared to a159b5baf3ad91d8defc68cbf81c78d01b69c416. This info should still be provided

Jermolene: Hi @danielo515 sorry, what is it that has changed?

danielo515: Hello @Jermolene , take a look at the linked commit. This method was being called with an extra options object that included tiddlywiki information. I was sure I made a PR to include it back, but seems that I forgot to click the pull request button. Here it is:
https://github.com/Jermolene/TiddlyWiki5/pull/4584

Jermolene: Ah, I see, thank you @danielo515

joshuafontany:
In PR https://github.com/Jermolene/TiddlyWiki5/pull/5467 it is proposed to change the order of arguments for the syncadaptor method signatures. @pmario suggested a comment on each changed method, but @Jermolene would like to reduce re-used code.

The `https://tiddlywiki.com/dev/#SyncAdaptorModules` tiddler references this thread to understand the "evolution of this API". I am stashing the full comment block here, and then reducing the code in the core files. 

```
	// Starting with 5.2.0, all syncadptor method signatures follow the node.js
	// standard of callback as last argument. This catches the previous signature:
	if(typeof callback !== "function" && typeof options === "function"){
		// First stash any non-function third argument, then swap
		var optionsArg = callback;
		callback = options;
		options = optionsArg;
	}
	options = options || {};
```

PR #4374, opened 2019-11-15
add location.hash to browser system info tiddlers

pmario:
Add some missing startup info. 

This will be a way for user, to know, how the wiki was opened.

see discussion in GG: https://groups.google.com/forum/#!topic/tiddlywiki/js4dpXZYMDg

### DIFF
--- a/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
+++ b/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
@@ -10,6 +10,10 @@ type: text/vnd.tiddlywiki
 
 System tiddlers in the namespace `$:/info/` are used to expose information about the system (including the current browser) so that WikiText applications can adapt themselves to available features.
 
+! Important
+
+The values below are "snapshots". The values are initialized at the wiki startup.
+
 ! Information Tiddlers
 
 |!Title |!Description |
--- a/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
+++ b/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
@@ -19,6 +19,7 @@ System tiddlers in the namespace `$:/info/` are used to expose information about
 |[[$:/info/browser/screen/height]] |Screen height in pixels |
 |[[$:/info/node]] |Running under [[Node.js]]? ("yes" or "no") |
 |[[$:/info/url/full]] |<<.from-version "5.1.14">> Full URL of wiki (eg, ''<<example full>>'') |
+|[[$:/info/url/hash]] |<<.from-version "5.1.22">> Hash portion of URL of wiki (eg, ''#InfoMechanism'') |
 |[[$:/info/url/host]] |<<.from-version "5.1.14">> Host portion of URL of wiki (eg, ''<<example host>>'') |
 |[[$:/info/url/hostname]] |<<.from-version "5.1.14">> Hostname portion of URL of wiki (eg, ''<<example hostname>>'') |
 |[[$:/info/url/origin]] |<<.from-version "5.1.14">> Origin portion of URL of wiki (eg, ''<<example origin>>'') |
--- a/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
+++ b/editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
@@ -23,7 +23,7 @@ The values below are "snapshots". The values are initialized at the wiki startup
 |[[$:/info/browser/screen/height]] |Screen height in pixels |
 |[[$:/info/node]] |Running under [[Node.js]]? ("yes" or "no") |
 |[[$:/info/url/full]] |<<.from-version "5.1.14">> Full URL of wiki (eg, ''<<example full>>'') |
-|[[$:/info/url/hash]] |<<.from-version "5.1.22">> Hash portion of URL of wiki (eg, ''<<example hash>>''). If empty click [[here|https://tiddlywiki.com/#InfoMechanism]]|
+|[[$:/info/url/hash]] |<<.from-version "5.1.22">> Hash portion of URL of wiki (eg, ''<<example hash>>'')|
 |[[$:/info/url/host]] |<<.from-version "5.1.14">> Host portion of URL of wiki (eg, ''<<example host>>'') |
 |[[$:/info/url/hostname]] |<<.from-version "5.1.14">> Hostname portion of URL of wiki (eg, ''<<example hostname>>'') |
 |[[$:/info/url/origin]] |<<.from-version "5.1.14">> Origin portion of URL of wiki (eg, ''<<example origin>>'') |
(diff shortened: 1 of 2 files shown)


editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
```diff
@@ -19,6 +19,7 @@ System tiddlers in the namespace `$:/info/` are used to expose information about
 |[[$:/info/browser/screen/height]] |Screen height in pixels |
 |[[$:/info/node]] |Running under [[Node.js]]? ("yes" or "no") |
 |[[$:/info/url/full]] |<<.from-version "5.1.14">> Full URL of wiki (eg, ''<<example full>>'') |
+|[[$:/info/url/hash]] |<<.from-version "5.1.22">> Hash portion of URL of wiki (eg, ''<<example hash>>''). If empty click [[here|https://tiddlywiki.com/#InfoMechanism]]|
```
Jermolene: What's the idea behind the note about the hash being empty? Why is it an external link?

Jermolene:
Hi @pmario the problem with this approach is (of course) that it just presents a snapshot of the location hash when the wiki was loaded, and it won't be dynamically updated if the location hash changes (e.g. by following a link). Having said that it is certainly a useful step to #3967.

Can we change the line in the docs to just say:

```
Snapshot of the hash portion of URL of wiki when it was loaded (eg, ''<<example hash>>'').
```

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
pmario: If the external link is clicked, the hash will have a value, so users can actually see something.

pmario:
IMO all of the `$:/info` elements are snapshots. So I think we should mention this at the top of the tiddler text. ... I will create a new PR

pmario:
New commit created!

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
Jermolene: OK I don't think that's clear. I don't think we need to explain anything here; this is only for users who know what the location hash is

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
```diff
@@ -10,6 +10,10 @@ type: text/vnd.tiddlywiki
 
 System tiddlers in the namespace `$:/info/` are used to expose information about the system (including the current browser) so that WikiText applications can adapt themselves to available features.
 
+! Important
+
+The values below are "snapshots". The values are initialized at the wiki startup.
```
Jermolene: Not entirely. I think that the only other values that can change without the wiki being reloaded are the screensize.

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
pmario: info.js is a startup module, which initializes all info tiddlers as "pseudo plugins", as far as I understand it. I don't see a "url change" mechanism. 

https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/startup/info.js

I think this info shouldn't be dynamic. Dynamic changes to the views needs to come from "storyviews" and may be others

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
pmario: hmm. where is my reply. click here: https://github.com/Jermolene/TiddlyWiki5/pull/4374#discussion_r351220202

editions/tw5.com/tiddlers/mechanisms/InfoMechanism.tid
Jermolene: I'm confused! All we're talking about here is the docs, and I'm suggesting that the "If empty..." part is not needed.

Jermolene: But we absolutely will need a dynamic mechanism to respond to changes in the location hash in order to properly address #3967 .

pmario: At the moment the "If empty" is needed, because we don't update the info automatically. So the hash hast to be there on page reload, otherwise the browser won't see it. 

This PR should implement a "snapshot" because that is what was requested in the [GG thread](https://groups.google.com/forum/#!topic/tiddlywiki/js4dpXZYMDg). 

I didn't want to implement something related to #3967 with this PR. 

I wanted to extend the existing mechanism because the hash element was missing.

Jermolene: > At the moment the "If empty" is needed, because we don't update the info automatically. So the hash hast to be there on page reload, otherwise the browser won't see it. 

I'm saying that we shouldn't include the link that causes the location hash to be updated. It's too obscure, and I think unnecessary.

> This PR should implement a "snapshot" because that is what was requested in the [GG thread](https://groups.google.com/forum/#!topic/tiddlywiki/js4dpXZYMDg). 

I'm happy with it being a snapshot, my point was that the docs need to point this out.

> I didn't want to implement something related to #3967 with this PR. 



> I wanted to extend the existing mechanism because the hash element was missing.

Jermolene: Sorry fat fingers!

> I didn't want to implement something related to #3967 with this PR. 

I'm afraid it is related whether we like it or not. It establishes an exposure of the location hash to userland, which is part of what we need for #3967.

> I wanted to extend the existing mechanism because the hash element was missing. 

It was deliberately missing. I still think we should try to go further with a proper, dynamic exposure of the location hash that allows custom actions to be taken when the hash changes.

pmario:
OK. We could name snapshot hash as: `$:/info/url/hash` and the new dynamic elements `$:/info/url/active/hash` 

So every dynamic element gets an `/active/` prefix once #3967 is active. So we can be backwards compatible and the stuff that needs to be dynamic can be easily identified. 

An other possibility would be `$:/info/url/dynamic/hash`. But it would be more to type.

pmario:
May be it would be better to group all dynamic elements together with: `$:/info/dynamic/xxxx` so screen-width and screen-height could be also part of the dynamic namespace.

Jermolene:
The issue with making a dynamic exposure of the location hash is figuring out a new mechanism to react to those changes. The approach I'm thinking about at the moment would be to have a tag `$:/tags/LocationHashChange` for tiddlers containing actions that are invoked whenever the location hash changes.

pmario:
As [mentioned in the OP and GG](https://groups.google.com/forum/#!topic/tiddlywiki/js4dpXZYMDg) the "snapshot hash" is already very useful to detect "startup info"

So please merge.

pmario:
>The issue with making a dynamic exposure of the location hash is figuring out a new mechanism to react to those changes. The approach I'm thinking about at the moment would be to have a tag $:/tags/LocationHashChange for tiddlers containing actions that are invoked whenever the location hash changes.

@Jermolene ... I would suggest to merge this PR to get the "snapshot" functionality and let's improve the dynamic mechanism with: Trigger actions when tiddlers change #5765

Jermolene:
Hi @pmario I'm not keen to merge this because it will add a backwards compatibility constraint to the future implementation of dynamic tracking. I've been doing some work on re-implementing the navigation mechanism in wikitext in the testbed of the TWPub project, and so will very soon need to address the dynamic tracking issue.

PR #4378, opened 2019-11-20
Add elixir to language support list

dweremeichik:
It appears that Elixir is already a supported language. This can be found by performing a ctrl-F on the [highlight.pack.js](https://github.com/Jermolene/TiddlyWiki5/blob/8f3da69f818940eb5f517da850fb3766b72c7d7d/plugins/tiddlywiki/highlight/files/highlight.pack.js#L2) file that ships with TiddlyWiki.

### DIFF
--- a/plugins/tiddlywiki/highlight/readme.tid
+++ b/plugins/tiddlywiki/highlight/readme.tid
@@ -36,6 +36,7 @@ The plugin includes support for the following languages (referred to as "brushes
 * css
 * diff
 * dockerfile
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @dweremeichik -- strictly speaking we need all contributors to sign the Contributor License Agreement, as described in contributing.md in the root of this repo. To avoid it I'll just commit a change directly, but please feel free to sign the CLA if you might be contributing again.

PR #4384, opened 2019-12-03
Add cacheFileInfo function

Arlen22:
This function caches `$tw.boot.files` to a JSON file in the data folder, allowing TiddlyWeb compatible servers to modify the data folder more efficiently. If this is not desired by default, `$tw.utils.cacheFileInfo` could be made a no-op function which specific plugins would be allowed to implement if needed for specific data folders. There is no good way to watch `$tw.boot.files` without using a setInterval, which is suboptimal.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -2124,6 +2124,7 @@ $tw.loadTiddlersNode = function() {
 	// Load the tiddlers from the wiki directory
 	if($tw.boot.wikiPath) {
 		$tw.boot.wikiInfo = $tw.loadWikiTiddlers($tw.boot.wikiPath);
+		$tw.utils.cacheFileInfo();
 	}
 };
 
--- a/core/modules/utils/filesystem.js
+++ b/core/modules/utils/filesystem.js
@@ -231,6 +231,21 @@ exports.generateTiddlerFileInfo = function(tiddler,options) {
 	return fileInfo;
 };
 
+/**
+ * Save the boot file info to disk.
+ * This allows simple TiddlyWeb access 
+ * without loading a node instance for it.
+ */
+exports.cacheFileInfo = function() {
+	if($tw.boot.files) {
+		fs.writeFileSync(
+			path.join($tw.boot.wikiPath, "fileinfo.json"), 
+			JSON.stringify($tw.boot.files, null, 2), 
+			"utf8"
+		);
+	}
+}
+
 /*
 Generate the filepath for saving a tiddler
 Options include:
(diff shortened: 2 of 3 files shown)


Arlen22:
As a no-op function, it would be implemented by a plugin with the same goal as https://github.com/Arlen22/TW5-storage-plugin, which is to make TiddlyWiki "client-side" data folders that are compatible with a simple TiddlyWeb implementation.

Arlen22:
The reasoning behind this idea is simple. Without this feature, a "client-side" data folder would need to be loaded in its entirety on every request, which is a very disk-intensive process for large wikis, for the sole purpose of building the `$tw.boot.files` object, which it would then use to process the request. This feature would remove the need for an external server to build this array from scratch, making the required server code much simpler and more like the file system adapter.

Arlen22:
I should have tested before creating the PR. It doesn't work in boot because utils-node isn't loaded until later. A plugin needing this functionality would just use `th-server-command-post-start` to do the initial cache. It's only needed in the file system adapter, which I think is fine, since loading shouldn't change anything. The file system adapter usually writes to `$:/StoryList` on startup anyway, so unless that is disabled it would do it then. 

I'm expecting a plugin will be required to activate "client-side" compatibility anyway, and a TiddlyWeb compatible server MUST look for this plugin first in the tiddlywiki.info before making any changes to the data folder.

I still want to put it in utils-node and the file system adapter, but if the file system adapter would be disabled by the plugin anyway, then this whole PR is irrelevant. I'll have to think about that.

Arlen22:
I'm studying the filesystem code and I think I'm going to add an additional stipulation to `$tw.utils.generateTiddlerFilepath` that the filepath must be under the specified directory. In other words, client-side tiddlers may not save outside the data folder. I can do this by wrapping the function. 

I could possibly save tiddlers to another folder beside the tiddlers folder, which would keep it away from the file-system adapter when loaded without the plugin and allow similar manipulation of the server-rendered index file. With tight control of the file system, it is possible to use the File System Adapter method to save directly from the TiddlyWeb routes in NodeJS. I could ignore path filters in order to keep things simple, or I could use a different set of path filters than the file system adapter uses and just make sure the path it returns is a child of the root directory using path.resolve and path.relative. 

Any thoughts on this, or are my thoughts too scattered to follow?

pmario:
@Arlen22 ... I stumbled upon this draft. Is this still something you consider?

Arlen22:
I don't think so. TW5 data folders are really designed as an input format (editions, plugins, etc). Their role as a storage format is shoe-horned in. 

Since I'm assuming git is the primary reason people want this, if I was going to design a folder-based system now, it would probably be some kind of JSON (perhaps the header fields as JSON followed by the text field), and stored in a flat folder with UUIDs for the filenames. Most likely it would also put the filename in a field and that would become the permanent tiddler ID, so any file name could work.

PR #4385, opened 2019-12-07
Make the saving throttle interval configurable

inmysocks:
After switching Bob to use the core syncer the throttle interval makes saving feel very sluggish compared to the message queue setup that I had before.
The editing lock that I use to prevent conflicts with multiple users doesn't go away until the save is completed, and with the 1 second delay it means that if you edit a tiddler and save it than you have to wait one second before you can edit it again.


Jermolene:
Thanks @inmysocks

PR #4389, opened 2019-12-14
Discourage / Deprecate data tiddlers

pmario:
PR: "Implement a data tiddler keyvalues operator #3971 " was rejected, because data-tiddlers are discouraged. 

If this is the case, then this PR can be easily merged.


Jermolene:
Hi @pmario I don't think we're quite at the point that we want to broadly discourage users from using data tiddlers because there are still quite a few use cases where they are the best option. The situation is more that I want to discourage further development of data tiddler features, and instead try to focus on better ways of achieving those remaining use cases.

kookma:
> Hi @pmario I don't think we're quite at the point that we want to broadly discourage users from using data tiddlers because there are still quite a few use cases where they are the best option. The situation is more that I want to discourage further development of data tiddler features, and instead try to focus on better ways of achieving those remaining use cases.

The Todolist plugin is one! The Trashbin (JSON) is another!

DataTiddlers are the best options in some cases!

pmario:
ok

PR #4392, opened 2019-12-19
add 3 new tiddlers, add 1 "enlist" test, fix all tests that failed, 

pmario:
This PR does: 

 - add 3 new tiddlers, 
 - add 1 "enlist" test, 
 - fix all tests that failed, because 3 new tiddlers where added. 
 - stopped ESLint to complain about global vars, 
 - fix some mixed-tab-space indent typos.

It also does enhance readability of failed tests. eg: JohnDo, JaneBloggs, JoeBloggs, DaveBloggs are way to similar to be recognized by humans. So it's incredibly hard, and "time wasting" to see, what actually is the error. 

 - I did change `JohnDo` to `John` which **shouldn't influence** the setting in line 47. 
 - I did change `JaneBloggs` to `Susi`

This PR is intended to serve as a base for new tests using "enlist" and may be other operands.

### DIFF
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -9,7 +9,7 @@ Tests the filtering mechanism.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw: false */
+/*global $tw: false, describe: false, it: false, expect: false, require: false*/
 "use strict";
 
 describe("Filter tests", function() {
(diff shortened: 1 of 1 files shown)


pmario:
@Jermolene ... It would be nice if this one can be merged fast, so I can use it as the base for more tests, that I wanted to implement.

Jermolene:
Thanks @pmario 

Is this part of a larger mission to make the whole codebase ESLint cleanly?

pmario:
>Is this part of a larger mission to make the whole codebase ESLint cleanly?

hmm, I think, this is a "side effect". I'm using the Brackets editor, which uses a [ESLint library](https://github.com/brackets-userland/brackets-eslint) with default settings. 

We have never agreed on a "code style guide" setting. 

I did have a look at existing "best practices" like https://github.com/airbnb/javascript. TWs styles are very close to this one, with some exceptions. eg: tab instead of spaces and 2 or 3 others which are different. 

Some of those recommendations are not used by TW eg: ES2015, which we don't use at all.

PR #4396, opened 2019-12-21
Fixed join filter operator to never return null

flibbles:
`[[Test]butlast[]join[ ]prefix[]]` would cause a Red Screen of Death.

This is because the join operator would return null if passed an
empty list. I've put in a small fix to ensure that doesn't happen,
and included a couple tests to catch that and another possible
pitfall.


Jermolene:
Thanks @flibbles

rmunn:
In case anyone finds this PR in the future, I've changed its behavior with #4853. This PR made the `join` operator return a list of one item (and that item was an empty string) when it had no input. With #4853, the behavior changes so that now the `join` operator will return an empty list if it had no input (i.e., any future filter operators down the line after join would also see no input).

PR #4397, opened 2019-12-27
Reduced indexOf calls in wiki.sortByList

flibbles:
PR for issue #4275

Examined the tests in test-tag. They already cover all the use cases
I could think of. I agree with Arlen22 that this change is functionally
identical to what was.

I also changed the name of the recursive function, and added some
comments to make this function easier to understand.


pmario:
The advantage is?

pmario:
hmm it should be linked to: #4275 right?

flibbles:
Yes, it should. I made the inexcusable error of thinking that linking to the issue in the title would create the link.

Jermolene:
Thanks @flibbles

Issue #2083, opened 2015-11-17
Add tiddler types for audio and video


braykov:
Current TiddlyWiki tiddler types contain Text and Image formats, but no audio or video.
Add these for easier embedding of music and video.

Jermolene:
Hi @braykov there are actually two separate layers at which different content types can be registered within TiddlyWiki:

* The boot kernel defines the low level file types that TiddlyWiki understands - [see them here](https://github.com/Jermolene/TiddlyWiki5/blob/master/boot/boot.js#L1818-L1841). A type being defined here tells TiddlyWiki whether to treat it as binary or text, for example
* The core UI defines tiddlers for each tiddler type that is to be selectable in the "type" dropdown in the edit template - they are defined [here](https://github.com/Jermolene/TiddlyWiki5/tree/master/core/language/en-GB/Types)

Are all the file types that you are interested in included in the first list?

There's a strong reason to keep the size of the dropdown reasonably small; we don't want users to have to scroll to see the entire list. So I think we should restrict it to the types that are commonly used, in particular the types that are commonly worked with interactively. Users can still manually type other types.

Creating `_canonical_uri` tiddlers manually is tedious in any case; I'd be quite interested in adding a special new UI for creating them.

braykov:
I am particularly interested in audio/mp3,mp4,ogg. They are all included in the first group.
If  I had to choose a single type, that would be .ogg.

pmario:
> So I think we should restrict it to the types that are commonly used, in particular the types that are commonly worked with interactively. Users can still manually type other types.

>If I had to choose a single type, that would be .ogg.

hmmm. While the [.ogg](https://en.wikipedia.org/wiki/Ogg) format is the one, that imo fits best to the TW project, because of its "free nature", its the one, that's the least known type. 

IMO [.mp3](https://en.wikipedia.org/wiki/MP3) is the de-facto standard and was the matter of a lot of copyright problems. Hence the .ogg format was created. ... but according to wikipedia

>In Europe mp3 is patent free and in the United States, the technology will be substantially patent-free on 31 December 2017

So there should be no problem, to implement them. .. But imo the "best audio" format discussion is a "hot" topic. If we support one extension, we basically need all of them. 

@Jermolene is it possible to include additional file extensions as plugins? 

This would also solve a lot of problems with other file types, that are often requested. Users can decide, what's important for them, so the dropdown list and the core can be small, and we don't need to carry things around, which we don't use often.

[...]

PR #4404, opened 2020-01-04
add support for video file type webm & ogg theora

s-light:
as discussed in [#2083](https://github.com/Jermolene/TiddlyWiki5/issues/2083#issuecomment-566963210)

added support for file type
- video/webm
- video/ogg
filename extension for Ogg Theora video files based on [this mdn note](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video#Server_support_for_video)


Jermolene:
Looks good, thanks @s-light

Issue #2744, opened 2017-01-31
Future maintenance and support for the ActionListops widget and associated filters


matabele:
@Jermolene @tobibeer 

Hi -- I have been struggling off and on with cancer for the past few years and no longer see that I will be able to maintain the ActionListops widget and extended filters code, nor provide support.

I believe that the cycle[] and keep[] filters should be pulled into the code from [here](http://listops.tiddlyspot.com/)

It might also be a good idea to pull the ActionMaketid widget and the associated filter into the core from [here,](http://maketid.tiddlyspot.com/) as I have had several requests for this functionality -- perhaps in the form of a plugin.

matabele:
@Jermolene @tobibeer 

Hi -- I have been struggling off and on with cancer for the past few years and no longer see that I will be able to maintain the ActionListops widget and extended filters code, nor provide support.

I believe that the cycle[] and keep[] filters should be pulled into the code from [here](http://listops.tiddlyspot.com/)

It might also be a good idea to pull the ActionMaketid widget and the associated filter into the core from [here,](http://maketid.tiddlyspot.com/) as I have had several requests for this functionality -- perhaps in the form of a plugin.

Then there are a couple of other projects I was working on which might be worth someone picking up on:

http://checklist.tiddlyspot.com/

http://simple-tagger.tiddlyspot.com/

If anyone is interested, I will help where I can, but don't expect any technical input such as modifications to algorithms. 

There are two outstanding support requests:

https://github.com/Jermolene/TiddlyWiki5/issues/2312#issuecomment-250885661

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/tiddlywiki/0m5CV3WXXu4/wtJwDwBzAwAJ

Jermolene:
Hi @matabele please accept my best wishes, and warm thanks for your past contributions. You're currently listed as contributor #9 to TiddlyWiki, which is pretty good going!

<img width="490" alt="screenshot 2017-01-31 08 39 49" src="https://cloud.githubusercontent.com/assets/174761/22457715/e276d4bc-e790-11e6-89a5-27971ed3dc10.png">

If anybody is in a position to help @matabele and I by packaging up some PRs, please do shout here.

matabele:
Hi @Jermolene 

Many thanks for the thumbs up. I should be able to answer questions relevant to a handover from time to time for a while -- but code changes or algorithm modification are out of the question.

tobibeer:
[...]

PR #2658, opened 2016-12-13
#2312 Prevent move filter from wrapping

mklauber:
Per Issue #2312 

> given a list `A B C D` if I run `A B C D +[move:-1 [A]]` I get `B C A D`.  However, if I were to do `A B C D +[move:1[D]]` it doesn't wrap around, and I get `A B C D`.  

This fixes that such that `A B C D +[move:-1 [A]]` gives `A B C D`

PR #4409, opened 2020-01-08
More listops tests

pmario:
Working on the new "story related" functions I needed several new lisops test, to be sure not breaking stuff. 

Some of the tests do have weird results. ... See comments. ... but they are as they are. We probably can't change the behaviour without breaking stuff. 

Especially have a look at the negative suffixes `:-x` they produce some confusing results. Most of them can be explained, but some are "just strange". ... 

Negative suffixes are probably not used nor documented, but they are possible, so they should be tested.

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw: false */
+/*global $tw, console, exports */
 "use strict";
 
 /*
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -82,7 +82,7 @@ function parseFilterOperation(operators,filterString,p) {
 				if(rexMatch) {
 					operator.regexp = new RegExp(rexMatch[1], rexMatch[2]);
 // DEPRECATION WARNING
-console.log("WARNING: Filter",operator.operator,"has a deprecated regexp operand",operator.regexp);
+console.log("\nWARNING: Filter",operator.operator,"has a deprecated regexp operand",operator.regexp);
 					nextBracketPos = p + rex.lastIndex - 1;
 				}
 				else {
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw, console, exports */
+/*global $tw:false, console:false, exports:false */
 "use strict";
 
 /*
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw:false, console:false, exports:false */
+/*global $tw: false */
 "use strict";
 
 /*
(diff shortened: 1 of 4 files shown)


pmario:
@matabele ... Can you have a closer look at the tests. ... I know some of them are not intended usecases, but they should be tested anyway.

Jermolene:
Hi @pmario sorry to remind you of #2744, and the passing of @matabele.

pmario:
Ah, So the tests as they are describe the actual behaviour. 

If you have a closer look at the comments, you'll see, where the "strange" things happen. ... We can either define the "wanted" behaviour and fix it OR we can "nail it down" to the existing behaviour. 

Strange things are mainly going on with negative suffixes, which seem to be not intended, except for the "move" operator. .... "move" already [got changed](https://github.com/Jermolene/TiddlyWiki5/pull/2658) in a way, that I don't like too much.

Jermolene:
Hi @pmario indeed some of that behaviour does seem unexpected! My inclination would be to merge these tests so that the current behaviour doesn't get changed inadvertently, but not to make a systematic attempt to "fix" them. Of course, I'd be happy to consider if there's a particular motivation to fix any particular case.

Also, there's again a bunch of eslint-related changes in this PR. I've no objection to updating that stuff across the board, but we probably shouldn't just so it piecemeal.

pmario:
>Also, there's again a bunch of eslint-related changes in this PR. I've no objection to updating that stuff across the board, but we probably shouldn't just so it piecemeal.

Do you use "eslint-like" addOn for your editor?

I could update all files at once, but that would invalidate all >100 existing PRs, which I don't want to happen.

pmario:
The listops operators exist since end of 2015 and we didn't have "functionality issues". So I think, it's OK to merge the tests. So the existing behaviour is "documented" and doesn't change by accident. 

If someone wants to change the behaviour and tests start to fail, we need to discuss the "new" functionality anyway.

pmario:
@Jermolene ... Is there a problem with the new tests?

core/modules/filters.js
```diff
@@ -82,7 +82,7 @@ function parseFilterOperation(operators,filterString,p) {
 				if(rexMatch) {
 					operator.regexp = new RegExp(rexMatch[1], rexMatch[2]);
 // DEPRECATION WARNING
-console.log("WARNING: Filter",operator.operator,"has a deprecated regexp operand",operator.regexp);
+console.log("\nWARNING: Filter",operator.operator,"has a deprecated regexp operand",operator.regexp);
```
Jermolene: Why is this newline being introduced?

pmario: If I run the tests like so: 

```
PS D:\git\tiddly\tiddlywiki\TiddlyWiki5> node .\tiddlywiki.js .\editions\test\ --build index
```
I get this output, which is much better to read, then with no NL and it doesn't create problems in the browser console. The warning is more visible there too. 

```
Randomized with seed 01155
Started
.....................................................................
WARNING: Filter modifier has a deprecated regexp operand /JoeBloggs/

WARNING: Filter modifier has a deprecated regexp operand /Jo/
......................................................
WARNING: Filter modifier has a deprecated regexp operand /JoeBloggs/

WARNING: Filter modifier has a deprecated regexp operand /Jo/
..........................


149 specs, 0 failures
Finished in 0.822 seconds
Randomized with seed 01155 (jasmine --random=true --seed=01155)
```

Without the new-lines the test-output is "scary". IMO with the NLs it's just a warning, which can be ignored.

pmario: With no new-lines, the first thing I saw, was the red text `/JoeBloggs/` and `/Jo/`. The rest of the text is hard to read. So I thought there is a real problem going on. ... 

The NL may avoid this problem for other users in the future.

Jermolene: Hi @pmario that makes sense, but it's still troubling to be changing something unrelated to the core purpose of this PR. I'd prefer to keep it how it is, or to add an option for disabling the warning message.

pmario:
@Jermolene only 1 files changed now.

Jermolene:
Thanks @pmario

PR #4412, opened 2020-01-14
Add a new example that shows using all before to determine the nummer…

AnthonyMuscio:
Numeric position of a title in a list.

See discussion https://groups.google.com/forum/?hl=en#!topic/tiddlywiki/wY6haW2cYMA


Jermolene:
Hi @AnthonyMuscio it's a cunning trick, thanks.

AnthonyMuscio:
The trick was Eric/Mohammads.

PR #4421, opened 2020-01-23
Add backlinks indexer

hoelzro:
This PR adds an indexer module for backlinks, along with some tests and a small tweak to the backlinks filter itself so it actually _uses_ the backlinks indexer.

### DIFF
--- a/core/modules/indexers/backlinks-index.js
+++ b/core/modules/indexers/backlinks-index.js
@@ -38,24 +38,26 @@ BacklinksIndexer.prototype.update = function(updateDescriptor) {
 		return;
 	}
 	var newLinks = [],
-	    oldLinks = [];
+	    oldLinks = [],
+	    self = this;
 	if(updateDescriptor.old.exists) {
 		oldLinks = this._getLinks(updateDescriptor.old.tiddler);
 	}
 	if(updateDescriptor.new.exists) {
 		newLinks = this._getLinks(updateDescriptor.new.tiddler);
 	}
-	for(var link of oldLinks) {
-		if(this.index[link]) {
-			this.index[link].delete(updateDescriptor.old.tiddler.fields.title);
+
+	$tw.utils.each(oldLinks,function(link) {
+		if(self.index[link]) {
+			delete self.index[link][updateDescriptor.old.tiddler.fields.title];
 		}
-	}
-	for(var link of newLinks) {
-		if(!this.index[link]) {
-			this.index[link] = new Set();
+	});
+	$tw.utils.each(newLinks,function(link) {
+		if(!self.index[link]) {
+			self.index[link] = Object.create(null);
 		}
-		this.index[link].add(updateDescriptor.new.tiddler.fields.title);
-	}
+		self.index[link][updateDescriptor.new.tiddler.fields.title] = true;
+	});
 }
 
 BacklinksIndexer.prototype.lookup = function(title) {
@@ -64,16 +66,16 @@ BacklinksIndexer.prototype.lookup = function(title) {
 		var self = this;
 		this.wiki.forEachTiddler(function(title,tiddler) {
 			var links = self._getLinks(tiddler);
-			for(var link of links) {
+			$tw.utils.each(links, function(link) {
 				if(!self.index[link]) {
-					self.index[link] = new Set();
+					self.index[link] = Object.create(null);
 				}
-				self.index[link].add(title);
-			}
+				self.index[link][title] = true;
+			});
 		});
 	}
 	if(this.index[title]) {
-		return Array.from(this.index[title]);
+		return Object.keys(this.index[title]);
 	} else {
 		return [];
 	}
--- a/core/modules/indexers/backlinks-index.js
+++ b/core/modules/indexers/backlinks-index.js
@@ -1,0 +1,99 @@
+/*\
+title: $:/core/modules/indexers/backlinks-indexer.js
+type: application/javascript
+module-type: indexer
+
+Indexes the tiddlers' backlinks
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global modules: false */
+"use strict";
+
+
+function BacklinksIndexer(wiki) {
+	this.wiki = wiki;
+}
+
+BacklinksIndexer.prototype.init = function() {
+	this.index = null;
+}
+
+BacklinksIndexer.prototype.rebuild = function() {
+	this.index = null;
+}
+
+BacklinksIndexer.prototype._getLinks = function(tiddler) {
+	var links = [],
+            parser =  this.wiki.parseText(tiddler.fields.type, tiddler.fields.text, {}),
+	    checkParseTree = function(parseTree) {
+			for(var t=0; t<parseTree.length; t++) {
+				var parseTreeNode = parseTree[t];
+				if(parseTreeNode.type === "link" && parseTreeNode.attributes.to && parseTreeNode.attributes.to.type === "string") {
+					var value = parseTreeNode.attributes.to.value;
+					if(links.indexOf(value) === -1) {
+						links.push(value);
+					}
+				}
+				if(parseTreeNode.children) {
+					checkParseTree(parseTreeNode.children);
+				}
+			}
+		};
+	if(parser) {
+		checkParseTree(parser.tree);
+	}
+	return links;
+}
+
+BacklinksIndexer.prototype.update = function(updateDescriptor) {
+	if(!this.index) {
+		return;
+	}
+	var newLinks = [],
+	    oldLinks = [];
+	if(updateDescriptor.old.exists) {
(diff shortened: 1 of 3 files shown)


core/modules/indexers/backlinks-index.js
```diff
@@ -0,0 +1,108 @@
...
+}
+
+BacklinksIndexer.prototype.rebuild = function() {
+	this.index = null;
+}
+
+BacklinksIndexer.prototype._getLinks = function(tiddler) {
+	var links = [],
+            parser =  this.wiki.parseText(tiddler.fields.type, tiddler.fields.text, {}),
+	    checkParseTree = function(parseTree) {
```
hoelzro: I'm not super thrilled about the duplication betwen here and `getTiddlerLinks` in wiki.js.  I wrote this a while ago and can't exactly remember why I didn't just use `getTiddlerLinks` - maybe I wanted to avoid the caching layer?

hoelzro: Ah, I see why now - I needed to be able to parse an arbitrary tiddler object that I pass in, rather than get from the tiddler store, because the `update` method gets such a tiddler object via the update descriptor.

pmario:
Tiddler backlinks are already indexed by the core.

pmario:
uups. They are not indexed, they are cached.

core/modules/indexers/backlinks-index.js
pmario: your .init does nothing, but it could use the existing core cache, if there is one.

pmario: Your indexer doesn't update the tiddler cache. .. So work has to be done twice. .. I also don't understand why you don't use: `wiki.getTiddlerLinks` .. IMO it does the same thing and it also updated the tiddler cache. 

You are duplicating code here, which makes maintenance problematic. If we update `wiki.getTiddlerLinks` we will forget about the indexer and vice versa. 

So if you need new functionality from `wiki.getTiddlerLinks` you need to update it there. ...

Just my opinion.

pmario: >Ah, I see why now - I needed to be able to parse an arbitrary tiddler object that I pass in, rather than get from the tiddler store, because the update method gets such a tiddler object via the update descriptor.

The `addTiddler` function saves the tiddler at: https://github.com/Jermolene/TiddlyWiki5/blob/b258afea0dc564b737409d88c72ca4180f967799/boot/boot.js#L1077

The indexers are called at: https://github.com/Jermolene/TiddlyWiki5/blob/b258afea0dc564b737409d88c72ca4180f967799/boot/boot.js#L1092

So it's perfectly fine, that you use the `wiki` functions. .. IMO no code duplication needed.

core/modules/indexers/backlinks-index.js
```diff
@@ -0,0 +1,108 @@
...
+BacklinksIndexer.prototype.update = function(updateDescriptor) {
+	if(!this.index) {
+		return;
+	}
+
+	var newLinks = [],
+	    oldLinks = [];
+
+	if(updateDescriptor.old.exists) {
+		oldLinks = this._getLinks(updateDescriptor.old.tiddler);
```
pmario: hmmm. If an old tiddler exists, it should be possible to use the cache, since parsing is "expensive" ..

pmario: Oh, I see. If `.tiddlerDelete()` is activated, the  old.tiddler is already overwritten in the store and caches are destroyed :/ 

https://github.com/Jermolene/TiddlyWiki5/blob/b258afea0dc564b737409d88c72ca4180f967799/boot/boot.js#L1129-L1133

.. That's unfortunate. ... So may be the `updateDescriptor` should provide this info, to avoid parsing for a deleted tiddler, which is complete waste of time. 

@Jermolene what do you think?

core/modules/indexers/backlinks-index.js
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	if(updateDescriptor.new.exists) {
+		newLinks = this._getLinks(updateDescriptor.new.tiddler);
```
pmario: here it's the other way around. The indexer needs to parse the tiddler, to get its info, so imo it would be good to update the cache too.

core/modules/indexers/backlinks-index.js
hoelzro: @pmario I agree with you on the problem on code duplication and would prefer to leverage `getTiddlerLinks` if I can.  The problem is how can I call `getTiddlerLinks` and get the links for the _former_ version of the tiddler (I need to remove any links that have disappeared in the new version), since `getTiddlerLinks` takes a title and uses that to get the tiddler from the tiddler store.

hoelzro: FWIW, I have no objections to updating the cache - I don't remember exactly why I omitted it from my reimplementation.  Maybe because it felt like storing too much data for something that might not be used again?

pmario: >...Maybe because it felt like storing too much data for something that might not be used again?

The cache is also used by Orphan and Missing tiddlers. So it can be reused. ... As you can see by my later replies, I did see the problem.  That's why, I did ask Jeremy.

pmario:
Did you make some "real  live" tests, if it does improve lookup performance? How many tiddlers are needed, so it makes a difference?

core/modules/indexers/backlinks-index.js
hoelzro: Yeah, that's ok with me!

core/modules/filters/backlinks.js
```diff
@@ -16,9 +16,15 @@ Filter operator for returning all the backlinks from a tiddler
 Export our filter function
 */
 exports.backlinks = function(source,operator,options) {
-	var results = [];
+	var results = [],
+	    indexer = options.wiki.getIndexer('BacklinksIndexer');
+
 	source(function(tiddler,title) {
-		$tw.utils.pushTop(results,options.wiki.getTiddlerBacklinks(title));
+		if(indexer) {
```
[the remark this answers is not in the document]
hoelzro: Sure thing - done!

core/modules/indexers/backlinks-index.js
```diff
@@ -0,0 +1,108 @@
...
+		checkParseTree(parser.tree);
+	}
+	return links;
+}
+
+BacklinksIndexer.prototype.update = function(updateDescriptor) {
+	if(!this.index) {
+		return;
+	}
+
```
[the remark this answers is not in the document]
hoelzro: Ok, I've removed the blank lines from my code and pushed the new commit.

hoelzro:
What's the policy on JavaScript language features? I noticed that I use `for(var link of links) { ... }` in a few places, and I was wondering if `for...of` is too recent a feature.

core/modules/indexers/backlinks-index.js
hoelzro: I've extracted that code into its own method and changed the indexer to use it.

hoelzro:
@Jermolene I've made some changes given your feedback - please let me know what you think when you have a chance!

Jermolene:
Hi @hoelzro sorry for the late reply

> What's the policy on JavaScript language features? I noticed that I use `for(var link of links) { ... }` in a few places, and I was wondering if `for...of` is too recent a feature.

We aim to stick to venerable old EcmaScript 5. Every time we introduce newer features we get complaints from users stuck on older Androids etc.

hoelzro:
Ok - I'll have to backport my change to stick with ES5-only features, then.

hoelzro:
By the way, is there an ESLint config or something I could use to run against my changes?  I see from the commit logs that @pmario has an ESLint config that they use?

pmario:
@hoelzro I'm using http://brackets.io as my editor of choice. It includes ESlint settings 

The plugin which they use is: https://github.com/brackets-userland/brackets-eslint

The setting seems to be: https://github.com/adobe/brackets/blob/master/.eslintrc.js

Which I didn't change. So it doesn't fit 100% to TW style, but it is close enough, to be useful.

hoelzro:
Ok, thanks @pmario! I'll give that a try and see how it works for me.

pmario:
There are some plugins, that are needed, so you can really work with it and git. I'm using those plugins see: https://youtu.be/oMMsiIqB8bs?list=PLuiC_HFhI4OxtThoQf92ws875R9zgNGSd&t=143

hoelzro:
Ok, I've removed any ES6isms I could find (or were picked up by eslint), and I've lifted the code to extract links from a tiddler into its own Wiki method. @Jermolene please have a look when you have a chance!

hoelzro:
@Jermolene Sorry to be a bother, but have you had a chance to look at this?

Jermolene:
Thanks @hoelzro I think this is indeed ready to go with this, sorry for the delays.

hoelzro:
@Jermolene No problem - thanks!

Issue #1522, opened 2015-02-23
Provide download link for binary tiddlers


Jermolene:
It should be possible to easily download the content of a binary tiddler as a file.

See this discussion for the background:

https://groups.google.com/d/topic/tiddlywiki/XoyXlWO8SpA/discussion

pmario:
+1

Jermolene:
This is a dupe of #781

Issue #781, opened 2014-09-03
Add ability to export binary tiddlers, including images


tlwt:
It seems to be impossible to access binary data stored in TW5. I get only the message mentioned before without any chance to download it.

r1chard5mith:
I'd like to be able to do this - download individual binary tiddlers (in order to make a manual image externaliser). Most of the export functionality seems to have been added since this ticket was opened and I suspect it may already be possible to do?

tlwt:
Hi @Jermolene, any news on this feature request? I would love to see it being scheduled for a milestone.

Jermolene:
Hi @tlwt apologies for the late reply. I'm afraid that this is just one of those things that I've not got to myself. There's a few fiddly changes to get this running, but I'm happy to be reminded again once I've got 5.1.10 released!

tlwt:
Hi @Jermolene, release is done (I already migrated to it). Here comes the reminder ;-)

Jermolene:
I've done some exploration. Sadly, there's quite a few changes we need:

* Updates to the saver interface and all of the saver implementations to support saving binary files as well as the current support for saving text files
* Add a new `tm-download-tiddler` message that actually downloads the raw text of a tiddler
* Add a new view toolbar button to download the current tiddler

Jermolene:
Fixed in #4423

PR #4423, opened 2020-01-25
Add download button/link to binary tiddler warning banners in view and edit mode

s-l-lee:
This PR adds a download link to the existing warning banners shown for tiddlers containing binary data:
example
It aims to solve https://github.com/Jermolene/TiddlyWiki5/issues/1522 and kind of helps with https://github.com/Jermolene/TiddlyWiki5/issues/781 but does not cover all the described cases. The code was mostly written referencing [$:/core/modules/parsers/imageparser.js](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/parsers/imageparser.js) and some parts of [$:/core/modules/widgets/edit-bitmap.js](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/edit-bitmap.js) as well as the original versions of the modified files.

It piggybacks off the existing warning banner for binary tiddlers code so that it does not need to check for the presence binary data itself. It uses the existing export button image for the download link to avoid the need for more translations. 

I don't have much experience with TiddlyWiki and open-source in general so I apologise for any trouble caused. I made these changes so that I can embed and access small .zip archives in my own TiddlyWiki conveniently and thought others may find it useful. Thank you for your time and attention.

Warm regards,
Sheng

### DIFF
--- a/core/modules/parsers/binaryparser.js
+++ b/core/modules/parsers/binaryparser.js
@@ -3,7 +3,7 @@ title: $:/core/modules/parsers/binaryparser.js
 type: application/javascript
 module-type: parser
 
-The video parser parses a video tiddler into an embeddable HTML element
+The binary parser parses a binary tiddler into a warning message and download link
 
 \*/
 (function(){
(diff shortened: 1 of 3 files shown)


core/ui/BinaryWarning.tid
```diff
@@ -1,8 +1,4 @@
 title: $:/core/ui/BinaryWarning
 
 \define lingo-base() $:/language/BinaryWarning/
-<div class="tc-binary-warning">
```
Jermolene: Why is this class being removed?

s-l-lee: This class is moved from [$:/core/ui/BinaryWarning.tid](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files/43427900ddb11025c2db2a1ceb952e9252d490b8#diff-c58fb2aaa141e94ac70468fbad065ff8) to [$:/core/modules/parsers/binaryparser.js](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files#diff-fc3d3179097ec63eeb5dd9a85c568920) and [$:/core/modules/widgets/edit-binary.js](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files#diff-62ccbeda3fbb154516f949d819aa9178), the binary tiddler view and edit mode respectively, so that [$:/core/ui/BinaryWarning.tid](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files/43427900ddb11025c2db2a1ceb952e9252d490b8#diff-c58fb2aaa141e94ac70468fbad065ff8) (now just the warning message text without the class) can be transcluded into a single instance of the class together with the download link in the binary tiddler view and edit modes. 

It was done in this way because I am not familiar with how the lingo macro and translation in general works, so I avoided messing with it. I have also previously tried a new 2nd instance of the class for the download link but it looked strange to me compared to this option of combining the message and download link in a single instance of the class. I would be interested to know any more elegant or idiomatic ways to achieve this as I am still quite new to TiddlyWiki.

core/modules/widgets/edit-binary.js
```diff
@@ -43,13 +44,55 @@ EditBinaryWidget.prototype.render = function(parent,nextSibling) {
...
+		type: "element",
+		tag: "div",
 		attributes: {
-			tiddler: {type: "string", value: BINARY_WARNING_MESSAGE}
-		}
-	}]);
+			class: {type: "string", value: "tc-binary-warning"}
+		},
+		children: [warn, link]
+	}
```
s-l-lee: `<div class="tc-binary-warning">` moved [$:/core/ui/BinaryWarning.tid](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files/43427900ddb11025c2db2a1ceb952e9252d490b8#diff-c58fb2aaa141e94ac70468fbad065ff8)

core/modules/parsers/binaryparser.js
```diff
@@ -13,14 +13,57 @@ The video parser parses a video tiddler into an embeddable HTML element
...
+	}
+	// Combine warning message and download link in a div
+	var element = {
+		type: "element",
+		tag: "div",
+		attributes: {
+			class: {type: "string", value: "tc-binary-warning"}
+		},
+		children: [warn, link]
+	}
```
s-l-lee: `<div class="tc-binary-warning">` moved from [$:/core/ui/BinaryWarning.tid](https://github.com/Jermolene/TiddlyWiki5/pull/4423/files/43427900ddb11025c2db2a1ceb952e9252d490b8#diff-c58fb2aaa141e94ac70468fbad065ff8)

core/ui/BinaryWarning.tid
Jermolene: Hi @s-l-lee that makes perfect sense, and is a good solution, thank you.

PR #4428, opened 2020-01-29
Update length Operator.tid

twMat:
IMO "length" is ambiguous because we typically associate to *lists* when tiddlyfiddling.

Instead of my proposed formulation "number of characters", maybe "amount of characters" is even clearer? An English tongue should decide.


Jermolene:
That's much better, thanks @twMat

PR #4432, opened 2020-01-31
Fix support for zip files in some environments

s-l-lee:
In some environments (at least on my own machine), TiddlyWiki detects zip files as type `"application/x-zip-compressed"` instead of `"application/zip"`. This commit adds support for zip files with type `"application/x-zip-compressed"` so that they are encoded in `"base64"` like other zip files with type `"application/zip"`.


Jermolene:
Thanks @s-l-lee

PR #4433, opened 2020-01-31
Add new parameters to fields-widget and fields-operator.

pmario:
This PR extends the `fields-widget` and the `fields` operator, so that the functionality is on par. 

## fields-widget

 - The widget has an `exclude` parameter 
   - The PR adds an `include` parameter 
   - if "include" is active, it takes precedence over "exclude"
 - The widget contains an hardcoded `.sort()` function
   - The PR makes the sort behaviour configurable
       - `sort` and `sortReverse` params are added
         - `sort` defaults to "yes", which is the current behaviour
         - `sortReverse` is optional
         - sort and sortReverse can be active at the same time. sort is executed first.
 - The PR adds new info to the docs tiddlers
 - The PR add new examples to the examples tiddler

## fields operator

 - The PR adds 2 new suffixes: `include` and `exclude`
   - eg: `[[HelloThere]fields:include[list title text non-existing]]`
      - If a field name doesn't exist, it will be ignored
      - This behaviour allows the list to dynamically expand, if new fields are added.
   - see examples tiddler for more info
 - The fields operator now uses the operands as a list of "field-titles" for "include/exclude"
 - fields operator doesn't do any sorting 

## This PR is mainly needed for "tools" authors to increase flexibility.

 - "include" allows to show a constant list of fields, without adding fields like: `bag` or `revision` in a server setting.
 - hardcoded sort order is not always wanted. 
   - eg if the resulting list is used with CSS, where order has a meaning


Jermolene:
Hi @pmario the `<$fields>` widgets is one of the early hacks that was needed to get saving working. Nowadays the same thing can be accomplished more consistently with the list widget. As a candidate for deprecation I'm not sure we should expend effort improving it.

As to the improvements for the `[fields[]` filter, this PR introduces a new convention that I don't think we currently have in the core: a space separated list as the operand. The `[search[]]` operator uses a different technique for specifying fields.

But more importantly, I think it might be more flexible to use primitive filter features to build the proposed functionality. For example, exclusion can be done with `[all[tiddlers]fields[]] -text -type`, and inclusion can be done with the `match` operator. For example, to just get the text and type fields (if present): `[all[tiddlers]fields[]match[text]] [all[tiddlers]fields[]match[text]]`.

pmario:
The `match` operator uses a flaglist. `fields` operator does the same. 
`match` uses a filter operand. a title list is a perfect subset of a filter operand. 

In the documentation I wrote and added an example, that exclude could be also achieved with additional filter runs, which is cumbersome and error prone. I wrote that `exclude` is a convenience function.

pmario:
>For example, to just get the text and type fields (if present): [all[tiddlers]fields[]match[text]] [all[tiddlers]fields[]match[text]]

Thx ... Your example is error prone with only 2 parameters. ... Have a very close look at your example. ... There is an error in the filter;)

Now do the same to create a filter like the following. 

 - It only shows fields of the current tiddler that exist.
 - The userList contains a list of all possible fields defined by the user
 - Fields that may or may not be part of the core should _not_ be shown. 
 - output elements must not be sorted.
 - unlimited-and-unknown-user-parameters contains eg: a b x y

```
<$set name=userList value={{unlimited-and-unknown-user-parameters}}>
<$list filter="[all[current]fields:include<userList>]" ><<currentTiddler>><br></$list>
</$set>
```

I think this is a complex example with a simple workflow for developers and end-users.

We do have several GG threads, that try to work with user defined fields. All of them come up with ultra complicated filters, that are very brittle. 

Links will follow

pmario:
> For example, exclusion can be done with [all[tiddlers]fields[]] -text -type

With exclusion alone you can't save the above example `[all[current]fields:include<userList>]` because you don't know future system fields or fields from other plugins. You don't know what to exclude in advance.

pmario:
A much more complex usecase is part of my uni-link plugin, where I use a widget named `uni-fields`, which does the same as fields-widget in this PR. 

As I mentioned in the OP, fields widget is used to create a dynamic CSS rule for "caption" or "subtitle" links, depending on which field us used to create the link. The class has to be adjusted. 

I couldn't find a way to use conventional filters, that would have allowed me to create this functionality, that simple

pmario:
I use the proposed fields-widget to create a dynamic `get-tabs-list()` in the [Story Editor Edition] (https://wikilabs.github.io/editions/wltm-open-story/#%24%3A%2Fcore%2Fui%2FAdvancedSearch%2FStory) that I promoted [at the GG](https://groups.google.com/forum/#!topic/tiddlywiki/z7938E0bi2k)

I couldn't find an easy way without this widget. ... The functionality isn't finished yet. 

The tab list should get a [+] tab, that should open a list with possible tabs. The tab list comes from a user config tiddler.

Jermolene:
Hi @pmario I'm happy with this now, but please can you remove the various changes made to the linting declarations? As discussed elsewhere, I don't mind updating that stuff but it needs to be done consistently and universally in a single PR, not piecemeal like this.

pmario:
@Jermolene .. did remove the eslint settings

Jermolene:
Thanks @pmario

PR #4434, opened 2020-01-31
`th-renaming-tiddler` invocation was missing

bimlas:
If we renamed the tiddler, this hook was not invocated [as the documentation says](https://tiddlywiki.com/dev/#Hook%3A%20th-renaming-tiddler).

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -330,8 +330,6 @@ NavigatorWidget.prototype.handleSaveTiddlerEvent = function(event) {
 				var shouldRelink = this.getAttribute("relinkOnRename","no").toLowerCase().trim() === "yes";
 				if(isRename && shouldRelink && this.wiki.tiddlerExists(draftOf)) {
 					console.log("Relinking '" + draftOf + "' to '" + draftTitle + "'");
-					// TODO: According to docs, we should pass objects, not just titles
-					// https://tiddlywiki.com/dev/#Hook%3A%20th-relinking-tiddler
 					this.wiki.relinkTiddler(draftOf,draftTitle);
 				}
 				// Remove the draft tiddler
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -322,7 +322,7 @@ NavigatorWidget.prototype.handleSaveTiddlerEvent = function(event) {
 				},this.wiki.getModificationFields());
 				if(isRename && this.wiki.tiddlerExists(draftOf)) {
 					console.log("Renaming '" + draftOf + "' to '" + draftTitle + "'");
-					newTiddler = $tw.hooks.invokeHook("th-renaming-tiddler", draftOf, draftTitle);
+					newTiddler = $tw.hooks.invokeHook("th-renaming-tiddler",newTiddler,tiddler);
 				}
 				newTiddler = $tw.hooks.invokeHook("th-saving-tiddler",newTiddler);
 				this.wiki.addTiddler(newTiddler);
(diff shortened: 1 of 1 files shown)


bimlas:
Please do not merge yet, I think I'm not tested it enough. :disappointed:

Jermolene:
Hi @bimlas fyi: https://github.blog/2019-02-14-introducing-draft-pull-requests/ -- sadly it doesn't seem to (yet) be possible to change an existing PR to draft status.

bimlas:
The hook works, but I found another bug: the `th-relinking-tiddler` is [called with string parameters](https://github.com/Jermolene/TiddlyWiki5/pull/4434/commits/09b8db7fb59f7a50da0ad7f0c74a4a2f509d8e66#diff-94f2ff697f3cad8f798cbe4f8d77ee3dR333-R335), not with objects.

If this is indeed a bug, then the Savetrail plugin should [throw an error when renaming and relinking a tiddler](https://github.com/Jermolene/TiddlyWiki5/blob/929b0c98330ad25aa4a32cd201274a6e89cd8902/plugins/tiddlywiki/savetrail/savetrail.js#L57-L65), because the `saveTiddlerFile()` called there gets a string instead of an object. I don't know the Savetrail plugin, so please test what it does in its current state and what happens when you change the hook parameters to an object.

When you agree, I'll change the code.

Jermolene:
Hi @bimlas the `th-relinking-tiddler` hook must indeed be called with tiddler objects. The `wiki.relinkTiddler()` method takes title strings, however.

bimlas:
Sorry, I confused the function call and the hooks. :S I will remove the comment and I think everything should work fine.

Jermolene:
Hi @bimlas apologies for the delay, I'm now looking at this in more detail. The docs are clear that the "th-renaming-tiddler" hook is invoked during processing of the "tm-rename-tiddler" message. Flipping a tiddler out of edit mode isn't considered a rename by that criteria, so I think the present behaviour is as documented.

Perhaps your use case could be met by using the "th-saving-tiddler" hook, perhaps adding an additional parameter of the draft tiddler object?

bimlas:
Sorry, I absolutely don't remember the circumstances anymore and the project has been deleted so I can't verify your suggestion. I close the issue.

rimi:
Hi! I cam accross exectly the issue discussed here and initiated in https://groups.google.com/forum/#!searchin/tiddlywikidev/th-renaming-tiddler%7Csort:date/tiddlywikidev/1-546CioHMc/wchFLFc_FAAJ.

My Use-Case is just as follows: Beside _tags_ and _list_ renaming (that is directly supported by TW5) i want to update my own list-fields (custom named) on ANY title-change (API or manual via editor) that occurs. As far as i understand the TW5-code, the best way to implement that is to listen on exectly the mentioned "th-renaming-tiddler"-event but that is not fired when renaming a tiddler by the standard edit-mode. 

Or is there any other approach i should use that i don't see now (beside changing core-code)?

Thanks,
Mirko

Jermolene:
Hi @rimi the `th-renaming-tiddler` hook is only invoked for renames made by the "tm-rename-tiddler" message, which isn't actually used by the core.

For your purposes, the `th-saving-tiddler` is a closer fit: it fires when saving tiddlers using the usual "tick" icon. However, at the moment it only passes the new tiddler, and doesn't provide access to the previous value of the tiddler.

I'll push an update shortly that adds this information to the hook. Then you should be able to identify renames by checking whether the "title" field of the new tiddler is the same as the "draft.of" field of the previous value.

rimi:
Hi @Jermolene, thanks for the quick reply! Looking forward to this change.

Best,
Mirko

Jermolene:
Hi @rimi the change is available now at https://tiddlywiki.com/prerelease and here on GitHub in the master branch if you want to experiment with it.

rimi:
I saw it right after typing the thanks ;) awesome!

PR #4436, opened 2020-01-31
Action create tiddler improved

pmario:
This PR extends the `action-createtiddler` command with 2 prameters: 

 - `$template` ... A tiddler title used as template ... Implementing a "clone" function will be trivial now.
 - `$overwrite` ... If a new tiddler title already exists, it should be overwritten. 
   - The overwrite parameter will be ignored, if the template tiddler would be overwritten

With those improvements, the action-createtiddler now behaves in the exact same way as a "tm-new-tiddler" message, except the new tiddlers won't be opened in edit mode. 

This PR makes cloning and adding user defined fields to an existing template tiddler is trivial. Links to GG threads will follow.


saqimtiaz:
+1 for adding support for a template to `action-create-tiddler`

This also opens the door to create a tiddler from a template without opening it in a story (which is currently the only option from a template via` tm-new-tiddler` )

pmario:
>This also opens the door to create a tiddler from a template without opening it in a story (which is currently the only option from a template via tm-new-tiddler )

That's 1 reason, why I did create the PR. We do have requests like this in the GG all over the place. 

A second reason was, to have the possibility to create tiddlers similar to tm-new-tiddler, with "open in edit mode" or "unique name" ... but have much more fine grained control, ...
without the need to use action-set-field with 100 parameters, which has high potential to accidentally overwrite tiddler content.

morosanuae:
This is very useful (and also mandatory) stuff. But why is not merged already?

saqimtiaz:
@pmario Would using an `action-deletetiddler` before the `action-createtiddler` allow one to accomplish the same goal as the $overwrite paramter? 

If so, that could allow the code for this pull request to be considerably simplified, and facilitate merging into the core.

pmario:
No. It makes configuration for macro authors so much easier. 

It's possible to have a `$template` tiddler only. ... Which uses the template tiddler name as the base name. In this case it's forbidden to delete the tiddler, because it's the template.

This behaviour is super convenient if you want to create a simple to create Clone function, that also adds new fields. 

Now try to implement this function with a macro, that contains the additional logic for action-delete. 

I can tell you, if you make some typos during development, you'll delete your template not only once. ... This is OK if the template is simple, but it's a pain if you don't have a backup!

Try this: 

```
\define testCreate()
<$action-createtiddler $template="ActionCreateTiddlerWidget Template" $overwrite=yes/>
\end

<$button actions=<<testCreate>> >
Create Tiddler
</$button>
```

pmario:
@Jermolene ... did you see this PR,  There is a lot of docs and examples too

Jermolene:
Thanks @pmario

PR #4437, opened 2020-02-01
Update ViewWidget.tid

twMat:
I'm assuming "text" is default. Please (silently) confirm that this indeed is the case before accepting.


Jermolene:
Thanks @twMat

Issue #4445, opened 2020-02-05
Bug: The StaticRiver exporter does not actually export any tiddler content


lucible:
Currently, `$:/core/templates/exporters/StaticRiver` does not export any tiddlers &/or their content when used with `--rendertiddler` to create a wiki snapshot with internal links. It results in an html page with included custom CSS, but no content.

Because `$:/core/templates/alltiddlers.template.html` is simply a reference back to the `StaticRiver` exporter, it is also broken.

To reproduce, use this command: `--rendertiddler $:/core/templates/exporters/StaticRiver alltiddlers.html text/plain` as listed in [Generating Static Sites with TiddlyWiki](https://tiddlywiki.com/static/Generating%2520Static%2520Sites%2520with%2520TiddlyWiki.html).

lucible:
Hm, after poking around the documentation some more, it would seem this and #4448 were a misunderstanding on my part about how the render command & templates work together due to the incorrect documentation at _Generating Static Sites with TiddlyWiki_.

It would seem that `--rendertiddler "$:/core/templates/exporters/StaticRiver" test.html text/plain "" exportFilter "[tag[example]]"` -- syntax gotten from the [RenderTiddlerCommand](https://tiddlywiki.com/static/RenderTiddlerCommand.html) documentation -- does indeed produce the expected results.

However, `$:/core/templates/alltiddlers.template.html` is still broken when used with the syntax above, and appears to run infinitely without completing.

Jermolene:
Hi @lucible it sounds like you've figured it out -- the export mechanism is separate from the saving mechanism, and it is intentional that it doesn't respect the same filters because there are alternate ways of specifying the payload for an export. As you've probably found, the docs could be improved and help in that regard is very welcome!

The command `tiddlywiki editions/prerelease/ --rendertiddler $:/core/templates/alltiddlers.template.html alltiddlers.html text/plain` works for me, but it is entirely possible that it could hang if used on a wiki that has tiddlers that recurse infinitely.

saqimtiaz:
@Jermolene @lucible this issue can be closed.

Jermolene:
Thanks @saqimtiaz

Issue #4446, opened 2020-02-05
Enhancement: Add publishFilter support to StaticRiver exporter


lucible:
Currently the save mechanism template includes a reference to `$(publishFilter)$` that allows the tiddler save filter to be customized, either in a global variable within the wiki or when using it in combination with the render tiddler command (example: `--rendertiddler $:/plugins/tiddlywiki/tiddlyweb/save/offline test.html text/plain "" "publishFilter" "+[tag[example]]"`).

`$:/core/templates/exporters/StaticRiver` does not currently align with the save mechanism template, and therefore does not accept a publish filter when used with the render command.

PR #4448, opened 2020-02-05
Update StaticRiver exporter to use publishFilter

lucible:
This fixes issues #4445 and #4446, by implementing the `saveTiddlerFilter()` used in the save mechanism. This also allows the use of `publishFilter` in combination with the StaticRiver exporter when used with the render command.

### DIFF
--- a/core/templates/exporters/StaticRiverContent.tid
+++ b/core/templates/exporters/StaticRiverContent.tid
@@ -1,7 +1,5 @@
 title: $:/core/templates/exporters/StaticRiver/Content
 
-\define renderContent()
-{{{ $(exportFilter)$ ||$:/core/templates/static-tiddler}}}
-\end
 \import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @lucible. The exporter mechanism is separate from the saver mechanism, and it is intentional that it doesn't respect the same filters (otherwise it would be hard to export tiddlers that were excluded from saving).

core/templates/exporters/StaticRiverContent.tid
```diff
@@ -1,7 +1,5 @@
 title: $:/core/templates/exporters/StaticRiver/Content
 
-\define renderContent()
-{{{ $(exportFilter)$ ||$:/core/templates/static-tiddler}}}
```
pmario: hmm, I think, that's not backwards compatible. ... you can't remove `exportFilter` without problems.

saqimtiaz:
@lucible @Jermolene this issue as well as #4445 can be closed.

The author admits in #4448 that this was a misunderstanding on their part.

Jermolene:
Thanks @saqimtiaz

PR #4454, opened 2020-02-10
Add isUnique to qualify - and - update core tabs macro

pmario:
This functionality was removed as the qualify macro was introduced. 

The `reveal-widget` and some other widgets, that need to deal with states, had 2 parameters. 

 - A `state` 
 - A `qualifyTiddlerTitles` to make it unique 

see commit: https://github.com/Jermolene/TiddlyWiki5/commit/fd70aa8c45f0f6049e044469b4f7c997ac31ce15

For example every core macro that uses the `<<qualify xxx>>` macro doesn't allow the user to specify a unique state title, because it uses the user-state as the base name. ... This causes problems for tools authors, because it is impossible to provide uniqe states without modifying the core macros. 

This PR fixes the problem.


pmario:
@Jermolene bump ... I know, this is a small change, but imo it will really make plugin authors life much easier.

pmario:
@Jermolene ... Please consider this PR for V5.1.23 ... It extends the functionality without any side effects, but it has a big value for content "creators" that want to use unique state names for predictability.

Jermolene:
Hi @pmario I think it might make more sense to split out an inner "tabs-unqualified" macro that takes an unqualified state tiddler title. It would be used in the ordinary "tabs" macro, and available for users to use directly..

By the way, I noticed some coding style issues in this PR, particularly around formatting and use of curly braces for "if" statements.

Jermolene:
Hi @pmario it looks like @BurningTreeC already addressed this by adding an "explicitState" parameter, see #4902

pmario:
Hi @Jermolene ... #4902 is docs only. The https://tiddlywiki.com/prerelease/ doesn't contain anything like this. 

@BurningTreeC .. please check my code. It primarily changes the `<<qualify` macro and widget. I'm pretty sure, that changing qualify macro is a generic solution. and adding a new parameter to the tabs macro is a "workaround"

Changing the tabs macro is a "prominent usecase" for the `<<qualify ` macro change.

Jermolene:
Hi @pmario see https://github.com/Jermolene/TiddlyWiki5/blob/master/core/wiki/macros/tabs.tid; the code was changed in #4875

pmario:
I still think, that my PR is generic solution for a global problem with the `<<qualify macro` and #4875 is a workaround to fix a specific problem for the tabs macro only. @BurningTreeC what do you think?

BurningTreeC:
Hi @pmario, @Jermolene - I also think that @pmario's solution is more generic and I could live with it... I'd just have to make some adjustments. The *actions* parameter in the `tabs macro` though would be still needed

Jermolene:
Hi @pmario @BurningTreeC I'm not keen on this fix because adding a switch to turn off the functionality of the macro is convoluted and makes things hard to read. It's clearer for the caller to avoid invoking the qualify macro if it is not needed.

pmario:
>... It's clearer for the caller to avoid invoking the qualify macro if it is not needed.

I'm with you. ... BUT at the moment plugin authors don't have a choice. `<<qualify` is all over the places in the TW UI. If we don't want to use it, we HAVE to overwrite core UI tiddlers, which is exactly the same as you shoot in your foot. (maintenance wise)

Eg:
>Hi @pmario it looks like @BurningTreeC already addressed this by adding an "explicitState" parameter, see #4902

The expliciteStat will be carved in stone with the next release. .. On the other hand it invalidates every other plugin, that already needed to modify the core tabs macro for the a **similar** or the same reason. 

I have 2 of them: https://wikilabs.github.io/editions/link-to-tabs/ and edit-tabs. @twMat has one or 2 of them too. Both versions are relatively popular. ... At the moment it is necessary, that those plugins ship with a modified version of the core tabs macro. 

I would like to have a version, that you describe:

>Hi @pmario I think it might make more sense to split out an inner "tabs-unqualified" macro that takes an unqualified state tiddler title. It would be used in the ordinary "tabs" macro, and available for users to use directly..

This would solve the `<<qualify` problem here.

BUT it wouldn't solve my or Mat's problem with link-to-tabs. We probably need a different hook, to extend the tab-macro functionality.

@BurningTreeC @twMat, ... I think we should "split" the tabs macro into several pieces (similar to Jeremy's suggestion), that will allow us to customize and extend it in the future, without the need to overwrite the core tiddlers. 

@BurningTreeC This will delay your other PR #4909 and may be #4875 would need to be revoked, but I think it would be worth it.

I would like to get it ready for 5.1.23!!

pmario:
May be, it's possible to use templates for the tab-button and other elements. So we would only need to modify those templates, or some variables to change the behaviour.

Jermolene:
Hi @pmario adding a "bypass" parameter to the "qualify" macro still requires that any macros that use it be modified to expose the "bypass" parameter. We might just as well modify those macros to provide a path that avoids calling the "qualify" macro in the first place.

Generally, I think that a function accepting a parameter that turns it into a "no op" is an anti-pattern.

pmario:
>Hi @pmario adding a "bypass" parameter to the "qualify" macro still requires that any macros that use it be modified to expose the "bypass" parameter. 

Yea, I thought about that too. I should have used a tv-xxx variable instead, like: https://tiddlywiki.com/#tv-wikilinks%20Variable ... 

IMO Using variables and modify the core functions to know about those variables, is an option, that is backwards compatible and will work without the need to modify core UI tiddlers. 

>We might just as well modify those macros to provide a path that avoids calling the "qualify" macro in the first place.

The problem with a "tabs_body" and a "tabs" macro is the same. If the core uses "tabs", users need to modify it to use "tabs_body". ... As soon as users need to modify core tiddlers we LOST. ... Splitting the macro into several pieces and using templates, we can make the elements, that need to be changed by plugin authors smaller, but we probably can't avoid it. 

>Generally, I think that a function accepting a parameter that turns it into a "no op" is an anti-pattern.

Yea, but sometimes we don't have a choice. https://tiddlywiki.com/#tv-wikilinks%20Variable

Jermolene:
> Yea, I thought about that too. I should have used a tv-xxx variable instead, like: https://tiddlywiki.com/#tv-wikilinks%20Variable

I don't think that will really help; it's just a hidden way of passing the parameter.

> IMO Using variables and modify the core functions to know about those variables, is an option, that is backwards compatible and will work without the need to modify core UI tiddlers.

It adds more hidden behaviour. The qualify macro as it is now is pretty simple, all we're doing is adding complexity. The fact that it's hard to come up with a name for the parameter is an indication that this is the wrong direction; "isUnique" doesn't make sense in this context.

> The problem with a "tabs_body" and a "tabs" macro is the same. If the core uses "tabs", users need to modify it to use "tabs_body". ... As soon as users need to modify core tiddlers we LOST. ... Splitting the macro into several pieces and using templates, we can make the elements, that need to be changed by plugin authors smaller, but we probably can't avoid it.

Now that I've seen @BurningTreeC's commits I'm no longer proposing that we use a separate tabs macro.

> Yea, but sometimes we don't have a choice. https://tiddlywiki.com/#tv-wikilinks%20Variable

That's not really the same. That's a variable that affects the behaviour of a widget, while here we're talking about a proposal to make a very small macro much, much more complex with no overall benefit.

pmario:
>That's not really the same. That's a variable that affects the behaviour of a widget, while here we're talking about a proposal to make a very small macro much, much more complex with no overall benefit.

As I stated in the OP, some useful functionality was removed as the `<<qualify ` macro was introduced. see commit: https://github.com/Jermolene/TiddlyWiki5/commit/fd70aa8c45f0f6049e044469b4f7c997ac31ce15

UI authors had the choice to use the state tiddler as a unique title, or let the core "qualify" it, if needed. ... The ´<<qualify´ macro removed that choice, and mad it impossible for plugin authors to configure the behaviour. 

With my PR I wanted to give more flexibility back to authors. .. It would have had a big benefit for me, since it would have made maintaining existing code easier.

PR #4458, opened 2020-02-15
Tag-Picker: keyboard selectable dropdown

BurningTreeC:
This PR adds a mechanism by which the tag-picker dropdown (or any other dropdown [or list] from within an input field) can be navigated using the `up` and `down` arrow keys. `Enter` adds a selected tag, `Escape` clears the input field or cancels editing the tiddler.

The mechanism adds a new `storeTitle` attribute to factory.js

The storeTitle is used to store whatever the user inserts in the input field. By navigating up/down the input field's value gets updated with the selection but the dropdown is filtered by the `storeTitle` value which doesn't change when the input's value is updated via an `action-setfield` command.

~~I had to edit `simple.js`, too and remove a condition which blocked updating the input field's value via an `action-setfield` command issued by a keyboard widget. I haven't found any issues by doing so.~~

This mechanism can be used in other places like the search field, too.

[HERE](http://tag-picker-kb.tiddlyspot.com/) is a DEMO of the working mechanism


Jermolene:
Hi @BurningTreeC thanks for persisting with this, it remains something that users regularly ask for, and I'd love to see it finally get into the core.

I tried the demo and noticed one minor oddity: after typing say "3" in the tag search box, the list is filtered as expected, and then I can cursor down and enter to select the tag. However, immediately afterwards the list is still filtered even though the tag search box is empty. Is that intentional?

core/modules/editor/engines/simple.js
```diff
@@ -67,9 +67,7 @@ Set the text of the engine if it doesn't currently have focus
 */
 SimpleEngine.prototype.setText = function(text,type) {
 	if(!this.domNode.isTiddlyWikiFakeDom) {
-		if(this.domNode.ownerDocument.activeElement !== this.domNode || text === "") {
-			this.domNode.value = text;
-		}
+		this.domNode.value = text;
```
Jermolene: The test you've removed is intended to prevent text inputs that contain the focus from being updated in response to changes to the underlying tiddler. As I recall, without this check typing into the text input will trigger an update of the text input, losing the cursor position/selection.

Jermolene:
Hi @BurningTreeC just checking, but am I right in understanding that the "storeTitle" attribute contains the title of a tiddler that the edit text widget will update whenever the user manually changes the text input?

BurningTreeC:
Hi @Jermolene,

> just checking, but am I right in understanding that the "storeTitle" attribute contains the title of a tiddler that the edit text widget will update whenever the user manually changes the text input?

Yes that's exactly the purpose of it

>I tried the demo and noticed one minor oddity: after typing say "3" in the tag search box, the list is filtered as expected, and then I can cursor down and enter to select the tag. However, immediately afterwards the list is still filtered even though the tag search box is empty. Is that intentional?

It's not fully intentional and can still be changed. I just didn't think about resetting the input's content to the content of the `storeTitle` (by which the dropdown is sorted). I'll make this update the coming weekend...

BurningTreeC:
I've updated the [DEMO](http://tag-picker-kb.tiddlyspot.com) in the meantime...

Now after adding a tag the Input Shows by what the Dropdown is sorted and the Input doesn't select an "empty" tag anymore

core/modules/editor/engines/simple.js
BurningTreeC: Thanks @Jermolene. This test somehow blocks updating the input's value by navigating up/down through surrounding keyboard widgets. Updating an input's value with a button widget works fine though. I don't know if it's a bug somewhere or if I need to add another test at this point in the code, I just didn't notice any difference with other inputs with or without this test

BurningTreeC: @Jermolene, the following code doesn't update the input fields' value, that's the problem here:

```
<$keyboard key="alt-shift-A" actions="""<$action-setfield $tiddler="$:/temp/test" text="12345"/>""">
<$edit-text tiddler="$:/temp/test" default="" tag="input" placeholder="">
</$keyboard>
```

BurningTreeC: @Jermolene, I don't notice any downsides by removing this test

Jermolene: In my tests, things work OK without that test on Firefox and Chrome, but it doesn't work on Webkit/Safari (Mac and iOS).

The issue is that on Webkit assigning a new value to the input element also moves the cursor to the end of the input text. I appreciate you may not have access to Safari to text it but the steps are:

1. First, make the change in the lines above (ie replace lines 70-72 with line 71)
2. Build the wiki and open it in Safari
3. Click the "new tiddler" button
4. Place the cursor after the word "new" and add the letters "ish"
5. Wait for up to 1 second for the draft debouncing to occur
6. Watch the cursor move to the end of the input box

Back when this code was new, the conditional was needed on Chrome and Firefox too. I think that since then both browsers have added a check to ignore attempts to set a text input to the text that it already contains. This can be verified by changing the line above to something like `this.domNode.value = text + "$";`; in Chrome and Firefox, attempts to type in the middle of a text box will again see the cursor move to the end.

BurningTreeC: Thanks for clarifying this, @Jermolene. Yes, I don't have access to Safari but I may be able to test with a webkit browser on linux. But now I understand the importance of this test.

To make this functionality happen nevertheless I could make the whole macro simpler and renounce to update the inputs' value with the selected dropdown item

Jermolene: Changing the test to checking whether the new text is different than the current text seems to work OK on Webkit, and is more readable:

```
if(this.domNode.value !== text || text === "") {
	this.domNode.value = text;
}
```

BurningTreeC: > Changing the test to checking whether the new text is different than the current text seems to work OK on Webkit, and is more readable:
> 
> ```
> if(this.domNode.value !== text || text === "") {
> 	this.domNode.value = text;
> }
> ```

This would make it work :smile:

BurningTreeC: I've updated the demo with your suggestion now, @Jermolene. works fine on firefox and chrome

Jermolene:
Hi @BurningTreeC thanks for your patience I think we're nearly there on this. One further thing we need is docs for the new edit-text attribute, but I wonder if we might also update the search dropdown along the same lines?

BurningTreeC:
Hi @Jermolene, I've updated the `edit-text` widget documentation.

I'd like to address the search dropdown in a second PR if that's OK for you

Jermolene:
Hi @BurningTreeC 

> Hi @Jermolene, I've updated the edit-text widget documentation.

Great, thanks. I've left a small comment.

> I'd like to address the search dropdown in a second PR if that's OK for you

That's OK, but we'll definitely need it for v5.1.22.

I also wondered whether you think the keyboard driven autocomplete functionality can be made into a reusable macro?

editions/tw5.com/tiddlers/widgets/EditTextWidget.tid
```diff
@@ -30,7 +30,8 @@ The content of the `<$edit-text>` widget is ignored.
 |size |The size of the input field (in characters) |
 |autoHeight |Either "yes" or "no" to specify whether to automatically resize `textarea` editors to fit their content (defaults to "yes") |
 |minHeight |Minimum height for automatically resized `textarea` editors, specified in CSS length units such as "px", "em" or "%" |
-|rows|Sets the rows attribute of a generated textarea |
+|rows |Sets the rows attribute of a generated textarea |
+|storeTitle <<.from-version 5.1.22>> |An additional tiddler title that stores the user's input |
```
Jermolene: The `<<.from-version 5.1.22>>` should go immediately after the second vertical bar for consistency with usage elsewhere. I also try to avoid apostrophes in documentation like this because I've no confidence that I get them right. Perhaps here better to say "An optional tiddler title that is updated with the text content each time it is directly updated by the user". Ideally, we'd also say "This can be used to create keyboard driven autocomplete inputs, please see XXXX", with a new tiddler giving instructions to end users wanting to build their own implementations.

BurningTreeC:
Hi @Jermolene, I'll make the PR ready for the next weekend

The functionality can be made into a reusable macro, that's true. I'll give it a try

Jermolene:
Great, thank you @BurningTreeC

BurningTreeC:
Hi @Jermolene, I've worked the functionality into a macro and updated the documentation. An explanatory tiddler is missing, I'm planning to add that in a separate PR if everything here is cleared.

The problem I'm facing is that the `keyboard-driven-input` macro only works with lists that don't contain items twice. So it isn't usable for the search dropdown in its current state.

In another thread you mentioned working with qualified titles but I don't get the idea how

BurningTreeC:
Hi @Jermolene,

Good News: I managed to make the macro work with the search dropdown, too

Bad News: A new filter operator `indexed` is needed wich indexes the input titles

But overall it works fine now as it is

Jermolene:
Hi @BurningTreeC great stuff - I fear my fix in a65ec87c53fceee69dac87d1051b46ebb7aaaf75 has given us a merge conflict.

BurningTreeC:
Merge conflict resolved @Jermolene

BurningTreeC:
>I also noticed that the support for searching included in the demo doesn't seem to be here?

@Jermolene, I wanted to add that in a second PR

core/language/en-GB/Shortcuts.tid
```diff
@@ -0,0 +1,6 @@
+title: $:/language/Shortcuts/
```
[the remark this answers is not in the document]
BurningTreeC: Ok @Jermolene, I'll make the change

Jermolene:
Hi @BurningTreeC 

>  I wanted to add that in a second PR

OK, it would be quite handy to see everything together if practical.

BurningTreeC:
>OK, it would be quite handy to see everything together if practical.

Then I'll add it in this PR

BurningTreeC:
Now it's all in one place @Jermolene, the sidebar-search functionality, too

core/language/en-GB/Shortcuts.tid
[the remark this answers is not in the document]
BurningTreeC: done @Jermolene

Jermolene:
Thanks @BurningTreeC much appreciated.

BurningTreeC:
@Jermolene, I'd be inclined to remove the `type="search"` on the sidebar search input because the `X` showing up within the search input can mess (a little) with the selected item not being reset correctly. It's better to use the `X` beneath the input or the <kbd>Escape</kbd> shortcut to clear the input field

Jermolene:
Hi @BurningTreeC that would be a shame, because it is more semantic to have type="search". Is it a visual problem, or usability?

BurningTreeC:
@Jermolene, it's not a BIG problem. It's because the mechanism uses three state tiddlers, two of them are connected to the user input and are being reset with the `X` within the search input field. The third one is independent of the user input and is only reset by the `X` beneath the search input or by the <kbd>Escape</kbd> keyboard shortcut. When the user uses the `X` within the search input to reset the field, the "selected item" stays the same and when cycling through the list the cycling may start with the previously selected item, because that state tiddler has not been reset. It's not a big problem but may cause some irritation.

What I'd like to do is make people test this mechanism and see if they get along well with it

core/ui/SideBarSegments/search.tid
```diff
@@ -4,7 +4,7 @@ tags: $:/tags/SideBarSegment
 \whitespace trim
 <div class="tc-sidebar-lists tc-sidebar-search">
 
-<$vars searchTiddler="$:/temp/search/input" selectionTiddler=<<qualify "$:/state/selected-item">>>
+<$vars name="searchTiddler" value="$:/temp/search/input">
```
pmario: imo should be a set-widget

BurningTreeC: thanks @pmario, this was an oversight

core/ui/DefaultSearchResultList.tid
```diff
@@ -5,11 +5,24 @@ caption: {{$:/language/Search/DefaultResults/Caption}}
...
+</span>
+</$list></$list>
 
 //<small>{{$:/language/Search/Matches/All}}</small>//
 
-<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]limit[250]]" template="$:/core/ui/ListItemTemplate"/>
+<$list filter="[!is[system]search{$(searchTiddler)$}sort[title]limit[250]]">
+<$list filter="[<currentTiddler>match{$:/temp/search}]" variable="ignore" emptyMessage="""
+{{||$:/core/ui/ListItemTemplate}}""">
+<span class="tc-list-item-selected">
```
Jermolene: I think this might be done more easily as:

```
<$list filter="[!is[system]search:title{$(searchTiddler)$}sort[title]limit[250]]">
<span class={{{[<currentTiddler>match{$:/temp/search}then[tc-list-item-selected]] }}}>
<$transclude tiddler="$:/core/ui/ListItemTemplate"/>
</span>
</$list>
```

BurningTreeC: thanks @Jermolene, I've applied your suggestion

BurningTreeC:
@Jermolene, I've found a workaround for the refresh-problem. When you have time, would you mind having a look?

The demo is also updated to use the new solution

BurningTreeC:
* fixed a typo in factory.js

pmario:
@BurningTreeC I did find an issue: To reproduce

 - I did open your Demo ... link in the OP
 - Go to "Draft of 'New Tiddler'"
 - Select tag-input
 - curser down
 - hit TAB-key
 - The dropdown doesn't close ... so it causes problems if I eg: select the type-dropdown

pmario:
Second thing:

 - If I select the type-input ... the dropdown opens
 - With mouse I select the tag-input
 - Results in "jumpy" behaviour

I'm using FF latest on Windows 10

BurningTreeC:
@pmario, to your first issue... that's the same on a current prerelease... dropdowns don't close if other inputs get focus

BurningTreeC:
@pmario, to your second issue, do you mean the tag-dropdown that opens for a moment and closes again? That's also behavior you can see on a current prerelease, also on older versions

pmario:
OK. I did only test it with your demo. ... Is it possible to fix it?

BurningTreeC:
>OK. I did only test it with your demo. ... Is it possible to fix it?

@pmario ... I tried to fix those issues with other PR's in the past but failed. I think it's possible to fix but atm I lack ideas how to

BurningTreeC:
Hi @Jermolene, are you still interested in this PR?

Jermolene:
Hi @BurningTreeC yes, very definitely so, and I think it's ready to go now we've got the docs etc. The frustration is that I need to get v5.1.22 released as soon as possible, and I don't think this has had wide enough testing to go out so quickly. So, I'm afraid I'm tending towards deferring this to v5.1.23. The last few release cycles have been too long; we should be aiming to get v5.1.23 out in a couple of months.

BurningTreeC:
Thanks for the new release @Jermolene :smile: !

Should we merge this right away to give it a test-drive on the prerelease?

BurningTreeC:
@Jermolene, I'm happy now how this works, I'd consider it done. How should we test this? The Demo is up to date

saqimtiaz:
> > just checking, but am I right in understanding that the "storeTitle" attribute contains the title of a tiddler that the edit text widget will update whenever the user manually changes the text input?
> 
> Yes that's exactly the purpose of it

 @BurningTreeC really excited about this feature. I am wondering though if we can re-think this part a bit. It would be a lot more flexible and re-usable if we could provide the edit-text widget with custom actions as a parameter, that are invoked everytime the user manually changes the text input, instead of hardwiring the storeTitle behaviour.

I have a similar keyboard driven setup that also involves autocomplete and started with needing to save the input value to a second tiddler as we do here with storeTitle. Eventually I needed to also update a state tiddler for a different part of the autocomplete mechanism that tied in. So that is an example where making this more flexible would increase reusability.

core/wiki/macros/keyboard-driven-input.tid
```diff
@@ -0,0 +1,50 @@
...
+$(filter1)$ $(filter2)$
+\end
+
+\define input-next-actions(afterbefore:"after",reverse:"",tiddler,storeTitle,field,index,filter1,filter2,refreshTitle)
+<$list filter="[<__storeTitle__>get[text]minlength<__filterMinLength__>] [<__filterMinLength__>match[0]] +[limit[1]]" variable="ignore">
+<$vars userInput={{{ [<__storeTitle__>get[text]] }}} selectedItem={{{ [<selectionTiddler>get[text]] }}}>
+<$set name="filter1" filter=<<__filter1__>>>
+<$set name="filter2" filter=<<__filter2__>>>
+<$set name="filterListRaw" value=<<get-filter-list>>>
+<$set name="filterList" filter="[enlist:raw<filterListRaw>indexed[]]">
```
saqimtiaz, line 14: Could we not skip the previous `set widget` for `filterListRaw` and just use the macro `get-filter-list` directly here?

`<$set name="filterList" filter="[enlist:raw<get-filter-list>indexed[]]">`

BurningTreeC:
Hi @saqimtiaz , I don't know if custom actions per input shouldn't be a separate pr... They're not used in this one.

Also, I'd be happy to see your solution somewhere showing up, maybe it works better than mine?

core/wiki/macros/keyboard-driven-input.tid
BurningTreeC: Oh, I think you're right. I'll test and change it

BurningTreeC: Thanks @saqimtiaz , I've changed it now

saqimtiaz:
@BurningTreeC your solution is by far cleaner overall as I resorted to javascript for some pieces.

I am suggesting we replace the storeTitle parameter to edit-text, with a custom actions parameter and an action to achieve the same thing as the storeTitle. 

It would be a more versatile change in edit-text rather than a one off change to support this feature. Doing that as a separate PR would mean that we end up with both the storeTitle parameter and custom actions, which would be redundant.

BurningTreeC:
>I am suggesting we replace the storeTitle parameter to edit-text, with a custom actions parameter and an action to achieve the same thing as the storeTitle.

Ok, I understand. There will be more possible use-cases I guess

saqimtiaz:
@BurningTreeC  yes, exactly. I know you have put a lot of hard work into this PR and it highly appreciated. I just want to get this as polished and as versatile as possible, as anything that increases keyboard handling in TW is a big win.

BurningTreeC:
@saqimtiaz - I'll brew something together and see how it works, thanks for your help on this

BurningTreeC:
Ok @saqimtiaz , I have a working version, so it's easily doable for me... Now I'm interested what @Jermolene thinks about actions triggered per-input

saqimtiaz:
@BurningTreeC I may have an idea that gets us around the need for the `indexed` operator as well. But I am on a tablet at the moment at work, so will need to investigate when I have time to sit down at a computer.

@Jermolene and I have previously discussed actions triggered per-input in a different context so I suspect we are good there, but let's wait for confirmation.

BurningTreeC:
>I may have an idea that gets us around the need for the indexed operator as well

that'd be great

Jermolene:
Hi @BurningTreeC @saqimtiaz I think the idea of adding actions to the edit widget is great, it keeps the edit widget much more generic, and gives us more flexibility here and for other similar use cases.

saqimtiaz:
@BurningTreeC switching to actions is going to help, but after that we need to re-think the implementation a bit more as there is a bit of a design and usability issue that needs to be resolved, and that will affect how we implement this. I ran into this while tweaking your code to avoid the indexed operator.

Currently in search, if the item selected is visible in both the "title" matches and the "all" matches, it is highlighted in both places. This can be very disorienting as you quickly move through the list.

Type "Adding" into the search field at http://tag-picker-kb.tiddlyspot.com/ and hit the down key 5-6 times. It is far more disorienting in longer search results.

Here is an illustration of the problem:
https://drive.google.com/file/d/1GUmvQDCsIYu49T8kQFJQrcs7WSseXC1G/view?usp=sharing

We need to find a way to uniquely qualify the results of each filter in a reproducable manner, so we can do the same where we are selecting which list item gets highlighted. I have an idea but it would involve passing a fixed state tiddler to the macro, and not calling qualify on it so that we can access it in `core/ui/SideBarSegments/search` as well.

Also keep in mind that currently the state tiddler is not reset when the search term changes, this can cause issues as well but we can reset the state tiddlers via actions when the user types.

BurningTreeC:
Ok @saqimtiaz , the demo is up to date with the discussed changes

saqimtiaz:
@BurningTreeC thank you. I have started looking at solutions to the issue of multiple high lighted list entries, as well as possible workarounds to needing the indexed operator. Will report back soon.

BurningTreeC:
Thank you @saqimtiaz!

saqimtiaz:
I have an updated [demo](https://saqimtiaz.github.io/sq-tw/keyboard-driven-saq.html#HelloThere:HelloThere%20GettingStarted%20Community) here that:

1. does not use the indexed operator. Instead the problem of duplicates in the two lists (result of two filter runs) is solved by means of a simple string suffix as qualifier.
2. In the logic to decide what list item to highlight, it uses the same string suffix to compare to the state tiddler (and not the search text). As a result, we no longer get duplicates highlighted in the results list.

The string suffix technique is exceedingly simply and can even feel a bit kludgy at first, but I am unconvinced that we need something more complex. We know we will never have more than 2 lists in this macro, and it seems reasonable to assert that while we can have duplication between the two lists, we cannot within each list.

Code changes diff: https://github.com/saqimtiaz/TiddlyWiki5/compare/04e7ece..58148bc

If we reach concensus over this technique, then we can work further on optimizing and simplifying the macros.

BurningTreeC:
Great stuff, thanks @saqimtiaz! I'm OK with your changes

editions/tw5.com/tiddlers/widgets/EditTextWidget.tid
```diff
@@ -30,7 +30,9 @@ The content of the `<$edit-text>` widget is ignored.
 |size |The size of the input field (in characters) |
 |autoHeight |Either "yes" or "no" to specify whether to automatically resize `textarea` editors to fit their content (defaults to "yes") |
 |minHeight |Minimum height for automatically resized `textarea` editors, specified in CSS length units such as "px", "em" or "%" |
-|rows|Sets the rows attribute of a generated textarea |
+|rows |Sets the rows attribute of a generated textarea |
+|inputactions |<<.from-version 5.1.22>> Optional actions that are triggered every time an input event occurs within the input field or textarea |
+|refreshTitle |<<.from-version 5.1.22>> An optional tiddler title that makes the input field update whenever the specified tiddler changes |
```
pmario, line 35: IMO Version 5.1.23. right?

core/wiki/macros/tag-picker.tid
```diff
@@ -68,16 +69,24 @@ tags: $:/tags/Macro
 <$set name="userInput" value={{{ [<storeTitle>get[text]] }}}>
 <$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
 <$list filter=<<filter1>> variable="tag">
-<$list filter="[<newTagNameTiddler>get[text]match<tag>]" emptyMessage="""<<tag-button>>""">
+<$set name="state-value" filter="[<q-state>get[text]]">
```
saqimtiaz, line 72: We should be able to make do with one less set widget, I was being conservative while trying to get this working. Will optimize later.

editions/tw5.com/tiddlers/widgets/EditTextWidget.tid
BurningTreeC: thanks @pmario , changed

BurningTreeC:
@saqimtiaz I think it's good to go now - what do you think?

editions/tw5.com/tiddlers/filters/indexed Operator.tid
```diff
@@ -0,0 +1,14 @@
+caption: indexed
```
saqimtiaz, line 1: tiddler not needed.

editions/tw5.com/tiddlers/filters/examples/indexed Operator (Examples).tid
```diff
@@ -0,0 +1,7 @@
+created: 20200313085714698
```
saqimtiaz, line 1: tiddler not needed.

saqimtiaz:
@BurningTreeC I think we are almost there! Since the next release is a way off, I think a few more days spent polishing this up will be worth it.

 There are a few small optimizations we should make to reduce the number of set widgets where possible. I can have a look at that this weekend. There are also a few variable names I used while making my tweaks that are worth renaming, like `q-state`, to be more self-explanatory. 

Is this meant to be a macro re-usable by end-users in a different context? It would be useful if that was possible. However, the macro seems quite intimidating with a plethora of parameters, and I fear it will be something we have a lot of end user support issues with. The crux is that the underlying widgets like edit-text have so many parameters that we need to account for, so not sure how much we can simplify it.

I also want to try to re-use the keyboard driven macro in a different context to understand just how re-usable it is, and if we can do anything to make that easier. For example from the tiddlywiki.com content we could test with a filter based on the tag "Filter Operators".  

In the same vein, we should also test and make sure that the macro works well if only supplying one filter. If the macro is meant to be re-usable by end users, I think only having one filter will be a common use case.

Jermolene:
Thanks @BurningTreeC @saqimtiaz this is looking good. It would indeed be great to make sure that we make it easy as possible for users to make their own autocompletes be keyboard driven.

saqimtiaz:
@BurningTreeC I am a bit unclear as to how to propose changes directly to the patch on github.
Here is a revised and optimized version of the tag-picker-inner macro from $:/core/macros/tag-picker

Removed some set widgets and the qualify widget. Renamed q-state variable to tagSelectionState

Please test before merging.

```
\define tag-picker-inner()
\whitespace trim
<$vars tagSelectionState=<<qualify "$:/state/selected-tag">> storeTitle=<<qualify "$:/temp/NewTagName/input">> refreshTitle=<<qualify "$:/temp/NewTagName/refresh">>>
<$set name="filter1" value="[tags[]!is[system]search:title<userInput>sort[]]">
<$set name="filter2" value="[tags[]is[system]search:title<userInput>sort[]]">
<div class="tc-edit-add-tag">
<$macrocall $name="keyboard-driven-input" tiddler=<<newTagNameTiddler>> tag="input" placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" tabindex=<<tabIndex>> focus={{{ [{$:/config/AutoFocus}match[tags]then[true]] ~[[false]] }}} inputactions="""<$action-setfield $tiddler=<<storeTitle>> text={{{ [<newTagNameTiddler>get[text]] }}}/>""" storeTitle=<<storeTitle>> filter1=<<filter1>> filter2=<<filter2>> input-accept-actions=<<add-tag-actions>> input-cancel-actions=<<clear-tags-actions>> filterMinLength={{$:/config/Tags/MinLength}} refreshTitle=<<refreshTitle>> selectionTiddler=<<tagSelectionState>>/>
</span>&nbsp;<$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button>&nbsp;<span class="tc-add-tag-button">
<$set name="tag" value={{{ [<newTagNameTiddler>get[text]] }}}>
<$button set=<<newTagNameTiddler>> setTo="" class="">
<$action-sendmessage $message="tm-add-tag" $param={{{ [<newTagNameTiddler>get[text]] }}}/>
<$action-deletetiddler $tiddler=<<newTagNameTiddler>>/>
<$action-deletetiddler $tiddler=<<storeTitle>>/>
{{$:/language/EditTemplate/Tags/Add/Button}}
</$button>
</$set>
</span>
</div>
<div class="tc-block-dropdown-wrapper">
<$reveal state=<<qualify "$:/state/popup/tags-auto-complete">> type="nomatch" text="" default="">
<div class="tc-block-dropdown">
<$set name="userInput" value={{{ [<storeTitle>get[text]] }}}>
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter=<<filter1>> variable="tag">
<$list filter="[<tag>addsuffix[-list1]] -[<tagSelectionState>get[text]]" emptyMessage="""<<tag-button "tc-tag-button-selected">>""">
<<tag-button>>
</$list>
</$list></$list>
<hr>
<$list filter="[<userInput>minlength{$:/config/Tags/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
<$list filter=<<filter2>> variable="tag">
<$list filter="[<tag>addsuffix[-list2]] -[<tagSelectionState>get[text]]" emptyMessage="""<<tag-button "tc-tag-button-selected">>""">
<<tag-button>>
</$list>
</$list></$list>
</$set>
</div>
</$reveal>
</div>
</$set>
</$set>
</$vars>
\end
```

Hoping to work on this some more tomorrow evening.

BurningTreeC:
Thanks @saqimtiaz , this is looking good

core/ui/SideBarSegments/search.tid
```diff
@@ -4,23 +4,27 @@ tags: $:/tags/SideBarSegment
...
+<$reveal state="$:/temp/search/input" type="nomatch" text="">
 <$button tooltip={{$:/language/Buttons/AdvancedSearch/Hint}} aria-label={{$:/language/Buttons/AdvancedSearch/Caption}} class="tc-btn-invisible">
 <$action-setfield $tiddler="$:/temp/advancedsearch" text={{$:/temp/search}}/>
 <$action-setfield $tiddler="$:/temp/search" text=""/>
+<$action-setfield $tiddler="$:/temp/search/input" text=""/>
+<$action-setfield $tiddler=<<searchListState>> text=""/>
```
saqimtiaz, line 15: Can we just delete the $:/temp/search, $:/temp/search/input and searchListState instead?

We could then replace those 3 action-setfield widgets with an action-deletetiddler widget? Untested code below:
`<$action-deletetiddler $filter=""""[[$:/temp/search]] [[$:/temp/search/input]] [<searchListState>]"""`

saqimtiaz:
@BurningTreeC just to confirm, is the purpose of the refreshTitle parameter to edit-text, to be able to refresh the displayed input value without triggering a refreshSelf and loosing focus?

core/ui/SideBarSegments/search.tid
BurningTreeC: Hi @saqimtiaz , yes that should work! I'll test it

BurningTreeC:
> is the purpose of the refreshTitle parameter to edit-text, to be able to refresh the displayed input value without triggering a refreshSelf and loosing focus?

the first purpose is that it makes it possible to update the input field with what is stored in the tiddler that's being edited. we're setting that tiddler's content through `action-setfield` widgets and the input field doesn't update if we don't tell him to do so afterwards

core/ui/SideBarSegments/search.tid
BurningTreeC: I believe there are more like these somewhere, I'll investigate

BurningTreeC:
@saqimtiaz I think I'm done with cleaning up the action-deletetiddler widgets

Is there something else you wanted to address?

saqimtiaz:
@BurningTreeC there are some optimizations we can make to `core/ui/DefaultSearchResultList.tid`, to replace those multiple set widgets for `state-value` and `qualified-tiddler`, similar to what I did for `\define tag-picker-inner()` macro above.

Basically replace
```
<$set name="state-value" filter="[<searchListState>get[text]]">
<$set name="qualified-tiddler" filter="[<currentTiddler>addsuffix[-list1]]">
<span class={{{[<qualified-tiddler>match<state-value>then[tc-list-item-selected]] }}}>
```

with something like this: (untested code)

`<span class={{{[<currentTiddler>addsuffix[-list1]] -[<searchListState>get[text]] +[then[]else[tc-list-item-selected]] }}}>`

I can work on this tomorrow afternoon (unless you want to take a crack), as well as try to do some testing for re-use.

core/wiki/macros/keyboard-driven-input.tid
```diff
@@ -0,0 +1,40 @@
...
+</$list>
+</$list>
+</$set>
+</$set>
+</$vars>
+</$list>
+\end
+
+\define keyboard-driven-input(tiddler,storeTitle,field:"text",index:"",tag:"input",type,focus:"",inputactions,input-accept-actions,input-accept-variant-actions,input-cancel-actions,placeholder:"",default:"",class,filter1,filter2,focusPopup,rows,minHeight,tabindex,size,autoHeight,filterMinLength:"0",refreshTitle,selectionTiddler)
+<$set name="selectionTiddler" value="$selectionTiddler$">
```
saqimtiaz, line 27: Can this set widget be avoided by directly using the selectionTiddler parameter with `<__selectionTiddler__>`
This is an artifact from when I was testing removing the indexed operator while wanting to make minimal changes.

BurningTreeC: yeah, I was looking at this before... I'll have another, deeper look :grin:

saqimtiaz: PS: If you can post an updated demo at the end of the day when you are done tweaking, that would be helpful. I am realizing that collaborating on a pull request on github and keeping files up to date is tricky.

BurningTreeC: If we avoid the set widget, then it must be passed to the input-nex-actions macro ... the second is more elegant so I'll change it

BurningTreeC: Yes I'll post an up-to-date demo later :)

saqimtiaz: @BurningTreeC many thanks!

BurningTreeC:
[The Demo](http://tag-picker-kb.tiddlyspot.com) is now up to date with the latest changes

saqimtiaz:
@BurningTreeC thank you. A bit under the weather but will be slowly reviewing this today and tomorrow (testing etc).

BurningTreeC:
I've found a way to remove the `X` within the search input field (on webkit browsers). Using it instead the `X` beneath the field confuses our mechanism

source: https://stackoverflow.com/questions/9421551/how-do-i-remove-all-default-webkit-search-field-styling

saqimtiaz:
@BurningTreeC regarding the x within the search field, I had the same problem and resolved it the same way in my autocomplete implementation. The issue is that it clears the input without triggering the actions we need to trigger when cancelling a search.

I worked a bit on this pull request yesterday. I found it incredibly difficult to even get started on re-using the keyboard-driven-input macro to create a different autocompleting text input and gave up after some minutes. It will be even harder for other users, as the actions that need to be provided to the macro require knowledge of the innerworkings of the macro.

I started refactoring a bit to rename variables and parameters for consistency, and moving action strings to macros to improve readability. Still a work in progress and it will be slow going for me this week as I am unwell, but you can monitor progress here: https://github.com/saqimtiaz/TiddlyWiki5/commits/keyboard-driven

BurningTreeC:
>I found it incredibly difficult to even get started on re-using the keyboard-driven-input macro to create a different autocompleting text input and gave up after some minutes. It will be even harder for other users, as the actions that need to be provided to the macro require knowledge of the innerworkings of the macro.

@saqimtiaz yes, it's hard to grasp. I was thinking about a wrapper macro that simplifies things, sets some defaults etc.

saqimtiaz:
@BurningTreeC yes I was thinking the same. Part of the reason I started renaming variables and moving action strings to macros was to make it easier to read and follow, but also to make it easier to see where there might be room for simplifying calling the macro (whether via a wrapper or otherwise). Please do feel free to explore this further. I wont be able to actively work on it until I'm feeling a bit better, needs a clear head.

BurningTreeC:
get well soon @saqimtiaz

BurningTreeC:
I'm now up to date with your changes @saqimtiaz ... but I'm using `selectionStateTitle` instead of `selectionStateTiddler`

saqimtiaz:
@BurningTreeC good call on renaming `selectionStateTitle`. If you are up to please feel free to take a similar pass through tagpicker to see if any variables/parameters need renaming, and move macro action parameters to macros for legibility. 

Will get back to this as soon as I am a little better.

BurningTreeC:
[The Demo](http://tag-picker-kb.tiddlyspot.com) is up to date

BurningTreeC:
I don't know how to simplify this any further @saqimtiaz @Jermolene 

I'm also facing problems in making this a universally usable thing, the possible use-cases seem to be too many and too different from each other, so that it's more convenient to study the tag-picker / the sidebar search and build something with that as a template

saqimtiaz:
@BurningTreeC I will review your updates by tomorrow, thank you for continuing to work on it. It may be that the best we can do, offer not a ready for use macro, but a pattern to follow to implement something similar.

So I wonder if the way to go is to document the pattern rather than the keyboard driven input macro itself. My concern is that if we add end user documentation for that macro there will be an expectation that it's easily reusable.

The problem is that the results of the drop-down can be used for so many things, opening tiddlers, adding tags or fields, or even more complex permutations. And it's not really possible to provide the action parameters for the keyboard-driven-input macro to perform when the user selects a result, without knowing the internals of the macro.

What if we provide some simple examples: 
a) to find titles based on one filter and open the selected titles
b) to add selected values to a field 
In addition to the examples of the search and tagpicker, that might be sufficient for advanced users to follow and make use of the keyboard-driven-input macro.

BurningTreeC:
>What if we provide some simple examples:
a) to find titles based on one filter and open the selected titles
b) to add selected values to a field
In addition to the examples of the search and tagpicker, that might be sufficient for advanced users to follow and make use of the keyboard-driven-input macro.

That would probably be the best

BurningTreeC:
Now the `refreshTitle` and `inputActions` attributes should also work correctly in the framed editor / codemirror editor

core/ui/SideBarSegments/search.tid
```diff
@@ -2,54 +2,78 @@ title: $:/core/ui/SideBarSegments/search
...
 <div class="tc-search">
-<$edit-text tiddler="$:/temp/search" type="search" tag="input" focus={{$:/config/Search/AutoFocus}} focusPopup=<<qualify "$:/state/popup/search-dropdown">> class="tc-popup-handle"/>
-<$reveal state="$:/temp/search" type="nomatch" text="">
+<$macrocall $name="keyboard-driven-input" tiddler="$:/temp/search" storeTitle=<<searchTiddler>> 
+		selectionStateTitle=<<searchListState>> refreshTitle="$:/temp/search/refresh" type="search" 
+		tag="input" focus={{$:/config/Search/AutoFocus}} focusPopup=<<qualify "$:/state/popup/search-dropdown">> 
+		class="tc-popup-handle" filter1=<<titleSearchFilter>> filter2=<<allSearchFilter>> 
+		filterMinLength={{$:/config/Search/MinLength}} inputCancelActions=<<cancel-search-actions>> 
+		inputAcceptActions=<<input-accept-actions>> inputAcceptVariantActions=<<input-accept-variant-actions>> />
+<$reveal state="$:/temp/search/input" type="nomatch" text="">
```
saqimtiaz, line 47: use `<<searchTiddler>>` instead of `$:/temp/search` ?

core/wiki/macros/tag-picker.tid
```diff
@@ -1,31 +1,50 @@
...
 <span class="tc-add-tag-name">
-<$keyboard key="ENTER" actions=<<add-tag-actions>>>
-<$edit-text tiddler=<<newTagNameTiddler>> tag="input" default="" placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" tabindex=<<tabIndex>> focus={{{ [{$:/config/AutoFocus}match[tags]then[true]] ~[[false]] }}}/>
-</$keyboard>
+<$macrocall $name="keyboard-driven-input" tiddler=<<newTagNameTiddler>> tag="input" 
+		placeholder={{$:/language/EditTemplate/Tags/Add/Placeholder}} focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> 
+		class="tc-edit-texteditor tc-popup-handle" tabindex=<<tabIndex>> focus={{{ [{$:/config/AutoFocus}match[tags]then[true]] ~[[false]] }}} 
+		storeTitle=<<storeTitle>> filter1=<<nonSystemTagsFilter>> filter2=<<systemTagsFilter>> inputAcceptActions=<<add-tag-actions>> 
+		inputCancelActions=<<clear-tags-actions>> filterMinLength={{$:/config/Tags/MinLength}} refreshTitle=<<refreshTitle>> 
+		selectionStateTitle=<<tagSelectionState>>/>
```
saqimtiaz, line 42: I think it helps with readability to have all the "title" parameters first: tiddler, storeTitle, refreshTitle, selectionStateTitle

core/ui/SideBarSegments/search.tid
BurningTreeC: thanks, fixed

saqimtiaz:
@BurningTreeC great changes, readability is significantly improved in my opinion. I think the only remaing steps are to:
- document keyboard-driven-input macro
- add examples as previously discussed: 
a) to find titles based on one filter and open the selected titles
b) to add selected values to a field

Can you please document the macro? It will help with creating the examples and Jeremy can make a judgmenet call on whether to include that documentation in the merge.

I can work on the examples but am still under the weather so it may be some time yet. So feel free to take a crack at that if you feel like it, otherwise rest assured that I will get to it. For the examples, we need to just base them off of existing tiddlywiki.com content.

core/wiki/macros/keyboard-driven-input.tid
```diff
@@ -0,0 +1,53 @@
...
+</$set>
+</$vars>
+</$list>
+\end
+
+\define keyboard-driven-input(tiddler,storeTitle,field:"text",index:"",tag:"input",type,focus:"",inputAcceptActions,inputAcceptVariantActions,inputCancelActions,placeholder:"",default:"",class,filter1,filter2,focusPopup,rows,minHeight,tabindex,size,autoHeight,filterMinLength:"0",refreshTitle,selectionStateTitle)
+<$keyboard key="((input-accept))" actions=<<__inputAcceptActions__>>>
+<$keyboard key="((input-accept-variant))" actions=<<__inputAcceptVariantActions__>>>
+<$keyboard key="((input-up))" actions=<<keyboard-input-up-down-actions "reverse[]" "before">>>
+<$keyboard key="((input-down))" actions=<<keyboard-input-up-down-actions "" "after">>>
```
saqimtiaz, line 40: can we call `input-next-actions` directly here?

BurningTreeC: yes, changed it! thanks

saqimtiaz:
@BurningTreeC sorry for the delay, hoping to review this again and work on those examples tomorrow. 

Is the demo up to date?

BurningTreeC:
@saqimtiaz at the moment code and demo are not up to date

saqimtiaz:
@BurningTreeC no worries, will review once updated.

BurningTreeC:
@saqimtiaz in #4725 I made a new PR which is the same as this one. This PR here has too many merge conflicts...

BurningTreeC:
Closing this in favor of #4725

PR #4478, opened 2020-03-08
Use config tiddler title to check filter query (merge into fix-syncer)

Arlen22:
Here is the change referred to in https://github.com/Jermolene/TiddlyWiki5/pull/4373#issuecomment-575702520


Jermolene:
Thanks @Arlen22 much appreciated.

Issue #4485, opened 2020-03-09
Extend Modal to allow custom class for customization


saqimtiaz:
Being able to add a custom class to the modal wrapper allows for customization via css.
As an example, it opens the door for resizable modals using only CSS.

This could be accomplished by either providing the custom class in the parameters to the tm-modal message, or via a field in the tiddler to be transcluded in the modal.

Jermolene:
Hi @saqimtiaz that's a good idea. I think perhaps it would be more in keeping with the current implementation for the class to be a field on the modal tiddler.

PR #4490, opened 2020-03-11
Add support for a custom class to modal wrapper

saqimtiaz:
As per #4485 add support for a custom class to modal wrapper, by means of a field in the modal tiddler. The class is added to the modal wrapper in addition to the default class, allowing for custom styling via css of any part of the modal.

### DIFF
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -58,6 +58,9 @@ Modal.prototype.display = function(title,options) {
 	this.adjustPageClass();
 	// Add classes
 	$tw.utils.addClass(wrapper,"tc-modal-wrapper");
+	if(tiddler && tiddler.fields && tiddler.fields.class) {
+		$tw.utils.addClass(wrapper,tiddler.fields.class);
+	}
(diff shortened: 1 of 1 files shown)


core/modules/utils/dom/modal.js
```diff
@@ -58,6 +58,9 @@ Modal.prototype.display = function(title,options) {
 	this.adjustPageClass();
 	// Add classes
 	$tw.utils.addClass(wrapper,"tc-modal-wrapper");
+	if(tiddler && tiddler.fields && tiddler.fields.class) {
```
Jermolene: There's a test for "tiddler" being falsey on line 40, so you can avoid checking that. I notice there's the same issue on line 107, if you'd like to kindly address that at the same time.

saqimtiaz: Resolved.

Jermolene:
Thanks @saqimtiaz

Jermolene:
Hi @saqimtiaz sorry I just realised that it would be good to have a docs update too! Many thanks

PR #4494, opened 2020-03-12
Add "Gruvbox Dark" color palette

BurningTreeC:
this adds the "gruvbox dark" color palette (https://github.com/morhetz/gruvbox) which is also available for highlight.js and codemirror

### DIFF
--- a/core/palettes/GruvBoxDark.tid
+++ b/core/palettes/GruvBoxDark.tid
@@ -1,0 +1,121 @@
+title: $:/palettes/GruvboxDark
+name: Gruvbox Dark
+description: Retro groove color scheme
+tags: $:/tags/Palette
+type: application/x-tiddler-dictionary
+
+alert-background: #cc241d
+alert-border: #cc241d
+alert-highlight: #d79921
+alert-muted-foreground: #504945
+background: #3c3836
+blockquote-bar: <<colour muted-foreground>>
+button-background: #504945
+button-foreground: #fbf1c7
+button-border: transparent
+code-background: #504945
+code-border: #504945
+code-foreground: #fb4934
+diff-delete-background: #fb4934
+diff-delete-foreground: <<colour foreground>>
+diff-equal-background: 
+diff-equal-foreground: <<colour foreground>>
+diff-insert-background: #b8bb26
+diff-insert-foreground: <<colour foreground>>
+diff-invisible-background: 
+diff-invisible-foreground: <<colour muted-foreground>>
+dirty-indicator: #fb4934
+download-background: #b8bb26
+download-foreground: <<colour background>>
+dragger-background: <<colour foreground>>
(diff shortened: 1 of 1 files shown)


kookma:
@BurningTreeC 
Love this!

The light mode is also great!

kookma:
I drag and drop this into https://tiddlywiki.com/prerelease/
The codeMirror uses the light (white) mode!

Is there any way CodeMirror automatically use the new pallet color?

BurningTreeC:
>Is there any way CodeMirror automatically use the new pallet color?

No, you need to explicitely add the gruvbox codemirror theme. See the "usage" section of the codemirror plugin for how to add color schemes

Jermolene:
Thanks @BurningTreeC it occurs to me that maybe for the palettes that have been adapted from other sources we should have a "credits" or "source" field?

BurningTreeC:
Yes @Jermolene, I'll add a credits field now

Jermolene:
Hi @BurningTreeC could you perhaps retrospectively add one for the other palettes you've ported?

BurningTreeC:
@Jermolene, I've ported the "Nord" palette, which has a "license" field pointing to the original MIT license

BurningTreeC:
... same with the "Solarized Light" palette, it also has a license field pointing to the license. Do you think a "credits" field is still needed in those cases?

Jermolene:
Hi @BurningTreeC apologies, in that case we should probably be using the same "license" field for consistency.

BurningTreeC:
Ok @Jermolene, done :smile:

Jermolene:
Thanks for your patience @BurningTreeC

BurningTreeC:
No problem @Jermolene! Thanks for merging!

kookma:
A minor point: 
Testing on https://tiddlywiki.com/prerelease/ the Comment plugin which is an essential one, on this new palette looks great! Except the border color! Is there any way to quickly adjust the border color?
It is light blue.

BurningTreeC:
@kookma - the comments plugin has its own styles built in. this would be a new issue

kookma:
@BurningTreeC 
Okay, so I think this for Jermey to see if the Comment plugin can gets its color from palettes or not!

kookma:
> > Is there any way CodeMirror automatically use the new pallet color?
> 
> No, you need to explicitely add the gruvbox codemirror theme. See the "usage" section of the codemirror plugin for how to add color schemes

Actually it would be great if plugins can get their color from palettes set by user!

By the way I have attached the two gruvbox-dark styles for codeMirror and Highlight.js here! I will also announce this in Forum.


[gruvbox-dark-for-Highlightjs-codeMirror.zip](https://github.com/Jermolene/TiddlyWiki5/files/4325715/gruvbox-dark-for-Highlightjs-codeMirror.zip)

Thank you both of you!

PR #4495, opened 2020-03-12
Update documentation for Modals

saqimtiaz:
Update documentation for modals to include the custom class introduced in #4485, as well as the footer and subtitle fields that don't appear to be documented.


Jermolene:
Thanks @saqimtiaz and apologies for the delay!

kookma:
A minor issue: There is a period after **nest wizard** button.

- Open https://tiddlywiki.com/prerelease/#Modals
- Click open demo modal
- Look at the button **nest wizard** and period after it
This repeats on all other modals.

PR #4491, opened 2020-03-11
added gitea saver

m42e:

PR #4496, opened 2020-03-12
Use CamelCase for Gitea

m42e:
As @kookma mentioned in https://github.com/Jermolene/TiddlyWiki5/pull/4491 the names should be in CamelCase. I agree and if you @Jermolene agree too, this could be merged, too. else just close it.


kookma:
@m42e 
The tab title under Saving should be **Gitea Saver**

Have a look at other tabs like GitHub Saver

Jermolene:
Hi @m42e @kookma thanks I agree that this looks better

kookma:
Still there is a small typo I had mentioned before.

Opne Saving/ Gitea Saver and read the first line:

`These settings are only used when saving to ~Gitea`

It should be like below. Note to the `~` in `~Gitea`

`These settings are only used when saving to Gitea`

PR #4497, opened 2020-03-13
Palettes: Updated GruvBoxDark for brightness clashes.

joshuafontany:
Love the new dark palette. I noticed some "contrast bugs", especially when viewing the tiddlywiki.com documentation tiddlers. These "clashes", either too bright or too dark versus the usually displayed contexts, are adjusted slightly.

... This is my first pr to the TW5 core, hope I did this right. :)


joshuafontany:
I'm not sure why there are 20 commits for 1 file change & why it was showing things I was experimenting with.... I did not touch anything besides the GruvBoxDark file after branching. Still figuring out GitHub.

kookma:
@joshuafontany 
I did not go through PR, but I think the UI elements of TW or some plugins does not follow the color palettes are set. For example Comment plugin uses its own colors and when one changes the palette to dark, it uses still bright colors. I think these clashes shall be resolved also!

BurningTreeC:
@joshuafontany thanks for your help on improving the palette :)

Jermolene:
Thanks @joshuafontany -- please could you remove the package.lock file from the PR? Many thanks.

BurningTreeC:
@joshuafontany, what I'm wondering is if the new colors are part of the gruvbox palette? See https://github.com/morhetz/gruvbox (scroll down a little bit) for the full palette colors

kookma:
> @joshuafontany, what I'm wondering is if the new colors are part of the gruvbox palette? See https://github.com/morhetz/gruvbox (scroll down a little bit) for the full palette colors

I agree with @BurningTreeC. I think the Tiddlywiki documentation and Comment plugins should follow the colors from palettes. So, the issue is in the other side and the gruvbox colors should not be changed!

joshuafontany:
Thanks everybody!

@BurningTreeC Thanks for pointing to the full color palette! I'll find a good set. I only added 1 hexcode, & moved one, so should not be a problem.

@kookma The documentation macros use the color macro (seen also in this palette as it can be called recursively). It would be worth looking at the comment plugin's style-sheet and applying the <<color NAME>> macro as needed.

@Jermolene Not sure how that file got into my tw folder.... I think I will try this again from a fresh fork of the repo.

Issue #4487, opened 2020-03-10
Request: Update CodeMirror and Highlight.js plugins in 5.1.22p


kookma:
**CodeMirror**
There are some requests for updating the CodeMirror in 5.1.22p. Seems there are some issues reporting here: https://groups.google.com/d/msg/tiddlywiki/4NGCvCJObhc/i_CsytpCAwAJ

**Highlight.js**
There was a bug in highlighting numbers in some language like fortran (e.g 12.23_wp) which addressed in new Highlight.js update. It would be nice to update the Highlight.js in 5.1.22p

BurningTreeC:
Thanks @kookma...

@pmario, what would you say, shall we update?

pmario:
We didn't get feedback from George "bluespire" ... So may be I scared him a bit too much ;)

There have been several changes to the javascript mode for eg: new elements like "arrow functions" and some such. If I did read the release notes right. 

Our last cm version is from April 2018 so I think it would be worth an update. 

----------

I did want to play with CM-V6-next and create a new TW5 syntax-mode + hints, but the new "parser" is a real parser optimized for highlighting now. see: [Lezer-parser](https://lezer.codemirror.net/)

The time needed to deig deeper is much higher, than I anticipated. 

---------------

So adding to CM-5 is worth a new thought.

 - There is also a new option, that would be nice: https://codemirror.net/addon/edit/trailingspace.js ... which would add nicely to an option I did implement as an [experimental parser addOn for Tony](https://groups.google.com/forum/#!topic/tiddlywiki/BJe7X-0jhSc). 

 - edit/continuelist.js ... needs some modifications, but could be nice add for TW too.
 - runmode/runmode.js and
 - runmode/colorize.js ... would make it possible to make highlight.js obsolete, if CM is used. 
 - merge/merge.js ... 
 - mode/overlay.js
 - mode/multiplex.js

All of that stuff would be nice addOns, if we really want to invest some more time into CM-V5.

BurningTreeC:
@pmario I was thinking that it's worth waiting for cm6 and just keep our cm5 plugins up to date. If you'd like to add some plugins, I'm totally fine with that

pmario:
With TWclassic I did use TW exclusively as a development environment, so an up to date CM editor was critical for me. 

With TW5 I'm using brackets.io editor (which uses CM5 :) because it has "inline code viewing and editing" and a really great git-plugin. 

[...]

Issue #4486, opened 2020-03-10
Plugin Library: Feature request


kookma:
**Feature 1: Important**
Assume you open Get More Plugin from $:/ControlPanel. You see list of plugins. No assume there is an update for a plugin you already have it. There is no sign to indicate your installed version is old.

See the below pic

![plugin-library](https://user-images.githubusercontent.com/830394/76279753-10d8f000-62a5-11ea-8719-aa3e9f93735e.png)

Also it is good to show the plugin version instead of Tiddlywiki release number. One solution is to use something different instead of reinstall with blue color (e.g orange color with some suitable message)

I understand, the plugin library is updated per new release, but if it is used by other publisher then their plugin release number may be different.

**Feature ii**
The plugin library when open shows every plugin using a panel with an arrow to open the readme. It is good the panel shows more information like a link to publisher site (here I like a link to demo site for documentation and tutorial)
By the way this is a small improvement and not important as one can click the double arrow and open the readme.

Jermolene:
Hi @kookma I understand feature 1, and agree it would be better to mark out of date plugins more prominently, but I may not be understanding feature 2 properly -- could you kindly make a sketch to show what you mean?

kookma:
@Jermolene 
Right now Plugin library shows each plugin on a panel with this information

- description e.g: `BibTeX: BibTeX importer`
- plugin tiddler name e.g: `$:/plugins/tiddlywiki/bibtex`
- Tiddlywiki release number e.g: `5.1.22prerelease`

Feature ii says, the panel can contains more information like

- link to code e.g.: `https://github.com/Jermolene/TiddlyWiki5/BibText`
- rank (based on the number of install, or stars people give)

I know the last one is difficult.

pmario:
I'm not 100% sure, but I think the "reinstall" button code has info about the "old" version. So it may be possible to write "update" instead of reinstall.

pmario:
or even better "update from: `<<old version>>`"

kookma:
> or even better "update from: `<<old version>>`"

That is much better update from ...

kookma:
@Jermolene 
Is there any chance to add this feature into 5.1.22? An indicator shows there is an update for installed plugin! The current reinstall does not have enough indication and user confuses if he/she can reinstall the current plugin release or a new update.

kookma:
@Jermolene 
I noticed on open the slider (arrows left of plugin title) a line appears shows the installed version!
[...]

PR #4508, opened 2020-03-15
Codemirror split PLUS working local test server

pmario:
This PR is the first take on the Proposal, to split CodeMirror5 into its own library. 

TLDR; To create a local test server
 
 - clone the branch
 - `cd editions/prerelease-server`
 - `npm test`

will run the test server.

 - This command will build the libraries, 
 - copy them to the server into the files directory and 
 - start a local server at http://localhost:8080

If libs are built you can use `npm start` which will only start the server BUT doesn't build / add any changes. 

You can open The ControlPanel : Plugins tab and play with the new settings. 

All CM AddOns have the dependencies defined. So eg: If you only select the hmtlembedded AddOn, all the needed settings will be imported too. 

The ControlPanel plugins tab uses the "name" field and is sorted by the "name" field. 

have fun!

------------

The PR has 19 commits starting from: Commits on Mar 15, 2020 ... So you can review every single step that I made. 

 - move library language texts into their own multids file ... fdfadc4
 - move library configuration files into new libraries directory ... 6723353
 - update path from tiddlywiki to codemirrror5 ... 557a4a0
 - update cm demo build and library creation build ... 7bba205
 - sort plugins by name-field and show name-field in ControlPanel plugin…  ... 9210557
 - make AddPlugins modal readable and remove redundant divs and other el… 67c33a7
 - don't commit prerelease-server/files/** ... they can be created and l… 6461855
 - fix cm paths for codemirrordemo edition  ... 1854373
 - fix plugin name info, since it is also used for pluigins now ... f463a67
 - fix cm path info in Spanish and French editions ... e6e030c
 - fix path for full edition ... 1b86510
 - fix cm path for introduction edition ... e0dadf7
 - remove comdemirror5 from default library ... 8e00c8b
 - cm5 gets its own library ... c74866d
 - new prerelease-server configuration with config tiddlers to test the … 78b53a9
 - adjust prerelease info and split library configuration into 2 files. … 6d3bdea
 - adjust cm5 path in tw5.com edition ... 4a80d13
 - move cm5 plugin to new library ... 3aa0b49
 - make plugin toggle chevron bigger, to make it easier to hit it. ... e8ffcf3


BurningTreeC:
@pmario ... first tests: the dependency-installation works fine, good work :) !

Jermolene:
Hi @pmario thanks for this, it must have been a lot of work. I like the idea of using the built-in server for testing, and think that that should really be it's own separate PR.

But overall I still need convincing that this is the right way to go, as I said over at https://github.com/Jermolene/TiddlyWiki5/issues/4487#issuecomment-598835460 but that is because I haven't seen an explanation of what problems these changes are intended to solve. Can you explain what the problems as you see them?

Jermolene:
Hi @pmario apologies I meant to add that the changes in indentation make it almost impossible to see what you've changed. I'm open to reformatting core templates, but it needs to be done in a separate PR.

core/language/en-GB/Libraries.multids
```diff
@@ -0,0 +1,6 @@
+title: $:/language/Libraries/
```
Jermolene: There are a few implications of adding a new language file, in particular the translator edition will need changing. What's the motivation here?

pmario:
>Hi @pmario apologies I meant to add that the changes in indentation make it almost impossible to see what you've changed. I'm open to reformatting core templates, but it needs to be done in a separate PR.

I needed to indent it, otherwise I have no chance to understand it. I was searching for the right place to add some styling, which I didn't add yet. But I did prepare the structure.

Yea TW does a much better job here. 

Modals-AddPlugins

pmario:
The proposed AddOn modal looks like this: 

 - I do have an overview, what's going on with 1 click on the "open plugin library" button
 - It's only 16 plugins, that all belong together
 - It gives me, as a maintainer, the possibility to add CM-related Themes and Languages
 - No information is hidden.
 - Adding new plugins will be much simpler. see next post
 - Adding [CodeMirror-6](https://codemirror.net/6/) will be simple, since we can use the same mechanism

new-install-01

---------------

>Can you explain what the problems as you see them?

The existing AddOn modal

If I'm looking at the image below, I'll see, that _all_ codemirror elements are installed. But they don't work. .. So I need to file an issue. 

There is absolutely no indication, that I must open the plugin.

install-01

If I'm able to open the plugin dropdown, with the chevron button area that is way to small, I only see 3-5 sub-elements. If I scroll down, the context goes missing.

install-02

If the modal looks like this, I am already lost. 

install-03

I need to zoom out like this, to get an overview. The problem is, I can't read it anymore. 

install-04

pmario:
The proposed ControlPanel uses plugin names for sorting and shows them. 

plugin-CP-01

The TW repo structure will be cleaner

tw5-dir-structure-01

and 

tw5-dir-structure-02

core/language/en-GB/Libraries.multids
pmario: I thought, it would be a cleaner structure, since the library texts where in `Misc.multids`

But we could wait until we get more 3rd party library configs.

pmario:
>Hi @pmario thanks for this, it must have been a lot of work. I like the idea of using the built-in server for testing, and think that that should really be it's own separate PR.

IMO this server atm only makes sense with the proposed configuration, since the `npm test` build requires the structure as it is atm. see: https://github.com/Jermolene/TiddlyWiki5/blob/816e470064858ffa215ef8d52763efef9692b8ca/editions/prerelease-server/package.json#L6-L19

kookma:
@pmario 
Please have a look at https://github.com/Jermolene/TiddlyWiki5/issues/4486
While this is off-topic here, but I recommend if you gonna to have any tweak for Plugin Library, consider the above feature request.

Further more when there several plugin libraries on a Tiddlywiki, then interface (modal window) needs a little more flat and cleaner UI elements

pmario:
@Jermolene ... I did just finish the file-backups update. ... So I could go on now, with updating the 3rd party libraries and do some testing. 

My arguments start here: https://github.com/Jermolene/TiddlyWiki5/pull/4508#issuecomment-599251685

Jermolene:
Hi @pmario sorry for the late reply, and thank you for laying out the issues so clearly. However, I don't think this PR is a good solution to those problems because:

* It burdens the user with understanding why things are setup in this way
* It requires us to rename the CodeMirror plugin, breaking backwards compatibility and burdening every user of the plugin
* It isn't scalable to the wider usability problems stemming from there being so many plugins in the library; we can't keep tearing groups of them off into separate libraries
* It changes the "publisher" portion of the plugin name to mean something other than is intended

There also remains a couple of problems with the PR that we've talked about. In particular, I can't merge it with the changes in indentation because they make it impossible to understand what has changed.

I'll shortly commit some of the changes you've suggested to make the plugin listings more consistent, and leave any further improvements to v5.1.23.

pmario:
>There also remains a couple of problems with the PR that we've talked about. In particular, I can't merge it with the changes in indentation because they make it impossible to understand what has changed.

I did add an image, that exactly shows the changes that I've made with the indentation. see: https://github.com/Jermolene/TiddlyWiki5/pull/4508#issuecomment-599251685

The problem here is made by GitHub. That's not my fault. If you **right click** the image, and select "view image" you can exactly see, what I've changed. 

The main part was the indentation, to make the code readable and maintainable by others. I also did remove some redundant elements, that have no meaning for functionality, but make the DOM more complicated and make CSS styling more difficult.

pmario:
I'll close the PR and create a new external library that includes the latest changes and works.

Jermolene:
> create a new external library that includes the latest changes and works.

Just to be clear, please don't redo this PR. As discussed, I'd rather explore more general ways to fix the UI problems you've identified.

pmario:
It's not only the UI problems. CM5 will be needed to be maintained and [CM6 is on the way](https://codemirror.net/6/)

With TWclassic and TiddlySpace I did use TW as a development platform for my plugins. I didn't use any other external editors. So it was important for me that CM works in the best possible way. 

I did create a Tiddlywiki2 wikitext highlighter for edit and view mode. ... In my own interest I did update the external cm-libraris as soon as they where available. So the stuff was always up to date.

With TW5 that's not the case anymore. So I only invest time for CM if someone at the GG requests it. 

In the future we will have CM5 and CM6 and we should be able to maintain them both. ... I don't see it as part of the core. 

I think it should be moved to it's own libraries and only the ["library tiddler"](https://github.com/Jermolene/TiddlyWiki5/pull/4508/files#diff-91b02d7f88ffe7f4a646c277a66f65e1) should be part of the core.

Jermolene:
Hi @pmario 

> It's not only the UI problems. CM5 will be needed to be maintained and CM6 is on the way

Presumably we'd add a new $:/plugins/tiddlywiki/coremirror6 plugin when that happens? We'd still be left with the problem that the plugin library listing will have too many items in it to be convenient; that's still the main UI problem I'm thinking of.

> With TWclassic and TiddlySpace I did use TW as a development platform for my plugins. I didn't use any other external editors. So it was important for me that CM works in the best possible way.

> I did create a Tiddlywiki2 wikitext highlighter for edit and view mode. ... In my own interest I did update the external cm-libraris as soon as they where available. So the stuff was always up to date.

> With TW5 that's not the case anymore. So I only invest time for CM if someone at the GG requests it.

Is that because of the time lag between TW5 releases?

> In the future we will have CM5 and CM6 and we should be able to maintain them both. ... I don't see it as part of the core.

That's an interesting point that I don't think you've made before. However, I think it's just another expression of the fundamental tension of the plugin library that we've explored elsewhere. (I.e. that plugin developers always want to publish the plugins independently so that they have control, while in my role as guardian of the interests of TiddlyWiki users I very much want the plugins that people depend upon to be pulled into a single repository).

> I think it should be moved to it's own libraries and only the "library tiddler" should be part of the core.

This now feels like a very different discussion than we started with: that we should move the CodeMirror plugins out of the core library in order to be able to make more frequent releases.

Jermolene:
Hi @pmario I've made some further improvements to the presentation of the plugin library, please see my comment over at https://github.com/Jermolene/TiddlyWiki5/issues/4486#issuecomment-611447864

PR #4515, opened 2020-03-18
TiddlyDesktop Wikilist: add colors to Vanilla palette

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

pmario:
What is `wiki-list`???? It seems those elements are not used. @BurningTreeC ... It would really help, if your PRs would have at least 1 line of comment, what they do and maybe "why"!

Issue #4504, opened 2020-03-15
Prerelease ControlPanel Info Basics - Filter links are broken - problem for existing plugins


pmario:
This is a real problem for many plugins, that will also use the "old" state tiddler. 

To reproduce expected

 - At 5.1.21 open https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FBasics
 - Click the last element: "Number of overridden shadow tiddlers"
 - The AdvancedSearch panel will be opend _and_ Filter tab will be selected. 
 - OK

To reproduce ISSUE

 - Open https://tiddlywiki.com/prerelease/#%24%3A%2Fcore%2Fui%2FControlPanel%2FBasics
 - Click the last element: "Number of overridden shadow tiddlers"
 - AdvancedSearch will be opened BUT no Filter tab will be selected. 

In 5.1.21 the state tiddler is: $:/state/tab--1498284803
In 5.1.22-pre it is: $:/state/tab-1822219760

Jermolene:
Many thanks @pmario can you determine which commit introduced the problem?

pmario:
I did have a fast look, but I couldn't find anything. ... It needs to be an additional transclusion somewhere. ... right?

pmario:
5.1.22 output array is

```
output: (6) […]
​​0: "{$:/AdvancedSearch|$:/AdvancedSearch|||}"
​​1: "{$:/AdvancedSearch|$:/core/ui/ViewTemplate/body|||}"
​​2: "{$:/AdvancedSearch|$:/core/ui/ViewTemplate|||}"
​​3: "{|$:/core/ui/PageTemplate/story|||}"
​​4: "{|$:/core/ui/PageTemplate|||}"
​​5: "{|$:/core/ui/PageTemplate|||}"
```

5.1.21 output array for hasString calc is:

```
output: (5) […]
​​0: "{$:/AdvancedSearch|$:/AdvancedSearch|||}"
​​1: "{$:/AdvancedSearch|$:/core/ui/ViewTemplate/body|||}"
​​2: "{$:/AdvancedSearch|$:/core/ui/ViewTemplate|||}"
​​3: "{|$:/core/ui/PageTemplate/story|||}"
​​4: "{|$:/core/ui/PageTemplate|||}"
```

So there seems to be something wrong with the PageTemplate. There is 1 more transclusion, which imo shouldn't be there.

flibbles:
[...]

PR #4518, opened 2020-03-19
Fixes issue with #4504 and importvariable copying

flibbles:
ImportVariables widget was using $tw.utils.extend to copy the
variables from temporary set widgets into itself. However,
$tw.utils.extend does NOT behave like Object.assign. It not only
copies all self-owned variables over, but also all variables
in that object's prototype chain. This led to some redundant copying,
and a problem where some variables might show up more than once
(like transclusion).

Fixed now. importvariables widget does its own copying, since it
can't rely on $tw.utils.extend to do the right job, and it can't
count on Object.assign to be there.

### DIFF
--- a/core/modules/widgets/importvariables.js
+++ b/core/modules/widgets/importvariables.js
@@ -61,7 +61,14 @@ ImportVariablesWidget.prototype.execute = function(tiddlerList) {
 					var widget = widgetPointer.makeChildWidget(node);
 					widget.computeAttributes();
 					widget.execute();
-					$tw.utils.extend(widgetPointer.variables,widget.variables);
+					// We SHALLOW copy over all variables
+					// in widget. We can't use
+					// $tw.utils.assign, because that copies
+					// up the prototype chain, which we
+					// don't want.
+					$tw.utils.each(Object.keys(widget.variables), function(key) {
+						widgetPointer.variables[key] = widget.variables[key];
+					});
 				} else {
 					widgetPointer.makeChildWidgets([node]);
 					widgetPointer = widgetPointer.children[0];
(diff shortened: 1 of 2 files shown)


flibbles:
I STRONGLY CONTEND that if $tw.utils.extend is supposed to be a replacement for Object.assign, then it is wrong.

I looked through all of the core code and I couldn't find any instance that exists where it copies over from an object that has a prototype chain, so this has so far been a hidden bug, but a bug nontheless.

I believe that the extend code which is this:
```javascript
exports.extend = function(object /* [, src] */) {
        $tw.utils.each(Array.prototype.slice.call(arguments, 1), function(source) {
                if(source) {
                        for(var property in source) {
                                object[property] = source[property];
                        }
                }
        });
        return object;
};
```
Should have been this:
```javascript
exports.extend = function(object /* [, src] */) {
        $tw.utils.each(Array.prototype.slice.call(arguments, 1), function(source) {
                if(source) {
                        $tw.utils.each(Object.keys(source), function(property) {
                                object[property] = source[property];
                        });
                }
        });
        return object;
};
```

Jermolene:
Hi @flibbles thanks that's great! I'm afraid the explanation for the vagaries of $tw.utils.extend() is just that it is really, really old! I think it dates from 2011/2 when we didn't have things like `Object.keys()` (which the core does now use elsewhere).

flibbles:
If it won't be fixed for back-compat reasons, it might be worth commenting on it so that dumb people like me don't make that mistake again. Or maybe create $tw.utils.assign...?

Jermolene:
Hi @flibbles I try to avoid adding utility functions that aren't themselves used by the core so I guess one answer could be to update your fix to use a new $tw.utils.assign function. In any case it would certainly be welcome to add a brief warning comment.

PR #4523, opened 2020-03-25
Add hidden setting to set taskTimerInterval, add documentation

inmysocks:
This adds a hidden setting for taskTimerInterval and adds documentation for the hidden settings:

task timer interval
saver filter
syncer filter


Jermolene:
Thanks @inmysocks

Issue #3004, opened 2017-11-07
TiddlyFox saver tiddlyfox.js doesn't pass on options received from saver-handler.js


pmario:
This is just a reminder.

see: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/saver-handler.js#L174 .. `{variables ....`
and: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/savers/tiddlyfox.js#L18

PR will follow.

sukima:
Because `master` changes over time here are the **permalinks** to the above:
* https://github.com/Jermolene/TiddlyWiki5/blob/ced9f315a1c78853bb2ff7e1377eef7d853b7d1e/core/modules/saver-handler.js#L174
* https://github.com/Jermolene/TiddlyWiki5/blob/66a13cb9154f019dc2c28e26ed89fb109654b13c/core/modules/savers/tiddlyfox.js#L18

pmario:
right .. thx. ..

pmario:
some more reminders:

```
SaverHandler.prototype.saveWiki = function(options) {
```
takes an options object, which also can contain a `callback()` function, that is unused atm. 

It will be activated, after the file has saved. see: https://github.com/Jermolene/TiddlyWiki5/blob/29364cbd081458427327dc831b23cc78036f0285/core/modules/saver-handler.js#L162

The default `tm-save-wiki` and `tm-download-wiki` messages are defined here. 
They are activated with the `autosave`, `download` and `save-wiki` buttons. 

see:https://github.com/Jermolene/TiddlyWiki5/blob/29364cbd081458427327dc831b23cc78036f0285/core/modules/saver-handler.js#L89

So a plugin can install eg: `tm-save-wiki-and-custom-notification" message, that also contains a custom callback(), that is activated as seen above. This mechanism is 100% backwards compatible, since it doesn't change any TW API signature.

Jermolene:
Sorry I hadn't seen this before. The original description isn't helpful. Would it be possible to update it to more clearly state what is proposed and why?

Leilei332:
Tiddlyfox extension is deprecated since firefox 57. Its saver should also considered deprecated now.

@pmario IMO we should close all tiddlyfox related issues as not planned.

pmario:
Can not be closed. This saver is still active if you use browser plugins. It is still an issue.

Jermolene:
[...]

PR #4526, opened 2020-03-30
Fix tiddlyfox problem with path like: Téléchargements 

pmario:
**Some more tests needed see text below**

The existing tiddlyfox.js saver gives a wrong name to AddOns like file-backups and Timimi.

That's why saving with path names like: "Téléchargements" fails in unix-like systems. Windows didn't have a problem.

This PR fixes the problem 

### Tests done with file-backups AddOn

 - I did test it with FF-latest on ubuntu 18.04
 - FF latest with Windows on local drive
 - FF latest with Windows and a mounted drive Z:\...

### Special test with FF 56 and TiddlyFox 2.0.2
 
see: https://github.com/TiddlyWiki/TiddlyFox

 - I did test FF56 and the old TiddlyFox extension in debug mode since the xpi isn't signed. 
 - Works as as expected. 

### Tests needed

 - Using Mac OS with FF and file-backups plugin, if that's possible
 - Using Timimi ... any OS


Jermolene:
Hi @pmario these conversions have given us a lot of trouble over the years. I'm a little reluctant to  make changes now because the TiddlyFox saver works OK with paths containing accented characters when used with TiddlyDesktop. Have you tested these changes with TiddlyDesktop?

pmario:
It hasn't been a problem with windows. .. I didn't test it with TD. Where can I find it?

pmario:
:) found it

pmario:
I can't run TD on ubuntu 18.04. If I want to "nw", which is an executable it says: 

"There is no application installed for "shared library" files. Do you want to search for an application to open this file"

gernert:
Hi @pmario 

I am using Mint 19.3 (based on Ubuntu 18.04) and can use TiddlyDesktop.
When I try to start nw with my default file manager (Caja in my case), I get the same message as you. But I can start nw with Double Commander.
I can also start nw by creating an app launcher.
Hope that helps.

pmario:
@gernert ... Thanks a lot. I did start it from the command line. ... that worked.

pmario:
@Jermolene TD has a problem with the new code, but I think, there are 2 problems with the "fix" that TD uses to deal with encoded characters in unix-like systems. 

-----------

`escape` and `unescape` that tiddlyfox.js save uses are functions that shouldn't be used anymore and may be removed in the future. (So we have a ticking bomb here!)

See: [escape docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/escape) and the spec: [Annex B](https://www.ecma-international.org/ecma-262/9.0/index.html#sec-additional-ecmascript-features-for-web-browsers)  B.2.1.1 and B.2.1.2

------------

TD uses a check, that assumes, that filepath is "escaped" based on the OS used, instead of checking the path string, if it really contains escaped elements. So if the path coming from tiddlyfox.js is already fixed, it breaks it. :(

see: 

https://github.com/Jermolene/TiddlyDesktop/blob/837a9711b265e8a94be098da43a47a54821fa4b1/source/js/utils/saving.js#L27-L30

The function used is `new Buffer().toString("utf8")` ... This function isn't available in browsers (at least I didn't find one), so browser extensions can't use it, to fix the problem. 

So it seems as you crated TD you found out about the problem. But instead of fixing the core problem in tiddlyfox.js you created a node.js specific workaround, that can't be used in browser extensions. 

------

I'm not happy to create a workaround, instead of fixing the problem.

Jermolene:
Hi @pmario 

I don't think we need to worry about escape and unescape going away, the docs are just trying to discourage new code using them.

The problem here would seem to be that if we fix tiddlyfox.js in a way that needs a companion fix to TiddlyDesktop then we'd break backwards compatibility between old TiddlyWiki's and TiddlyDesktop.

Perhaps an alternate approach would be for tiddlyfox.js to add a new attribute with the unmodified path alongside the existing garbled path, and then update future versions of TiddlyDesktop/file-backups/Timini can opt to use that attribute if it's present.

Also, as we've discussed before, we really need to document the TiddlyFox protocol somewhere...

pmario:
>Perhaps an alternate approach would be for tiddlyfox.js to add a new attribute with the unmodified path alongside the existing garbled path, and then update future versions of TiddlyDesktop/file-backups/Timini can opt to use that attribute if it's present.

... OK. ... But then I'd also like to fix issue #3004, which will also effect core/modules/saver-handler.js

Since Timimi and file-backups would need to be changed anyway, I'd use the possibility to add a new feature to file-backups: A "Snapshot" save. Which is an "out of order" save, similar to a "git tag" option.

ibnishak:
> alternate approach would be for tiddlyfox.js to add a new attribute

If we are in the process of adding attributes, may I please request a backup path attribute too. I have seen this already in place for tiddlyspot saver. However if this is available for other save objects and user is able to define in control panel, then each tiddlywiki can have its own absolute/relative backup paths.

ibnishak:
Also would this be a good time to discuss the merit of a "lock" attribute too - to prevent it from being saved by anyone else until an unlock attribute appears in a later save object?

Issue #4519, opened 2020-03-20
Range widget does not save value in IE10


saqimtiaz:
The range widget does not save its value on user mouse input in IE 10 on Windows 10.

To reproduce the bug:

- Open https://tiddlywiki.com/#RangeWidget
- Move the range widget sliders in the demos.
- The new value is not displayed as the tiddler the value is to be saved to, is not created.

This is probably because IE 10 & 11 do not fire the `input` event on range sliders on mousemove, but rather the `change` event.
See "Known Issues" at https://caniuse.com/#feat=input-range

Jermolene:
Thanks @saqimtiaz do you have a fix?

saqimtiaz:
> Thanks @saqimtiaz do you have a fix?

I can attempt a fix in the next few days,  I currently do not have a Windows test/debug environment and will need to set one up

Jermolene:
Thanks. I don't pay much attention to IE10 but it's good to fix what we can for IE11

saqimtiaz:
@Jermolene I believe this is the simplest solution and doesn't require resorting to browser sniffing:
https://github.com/Jermolene/TiddlyWiki5/pull/4534

PR #4534, opened 2020-04-03
Add change event listener range widget for IE10/11

saqimtiaz:
As detailed in #4519 the range widget currently does not save its value to the state tiddler on IE 10/11 as they do not support the` input` event, but rather the `change` event is fired instead of the input event. This has patch has been tested in IE11 and should work in IE10 as well.

Note that on Chrome and Firefox, the `change` event will fire only once after the user stops dragging the range slider (In addition to the `input` event). However this does not lead to an extra refresh as the `handleInputEvent` method already checks to see if the current value of the slider is different from the saved value before saving to the store.

### DIFF
--- a/core/modules/widgets/range.js
+++ b/core/modules/widgets/range.js
@@ -49,7 +49,8 @@ RangeWidget.prototype.render = function(parent,nextSibling) {
 	this.inputDomNode.value = this.getValue();
 	// Add a click event handler
 	$tw.utils.addEventListeners(this.inputDomNode,[
-		{name: "input", handlerObject: this, handlerMethod: "handleInputEvent"}
+		{name: "input", handlerObject: this, handlerMethod: "handleInputEvent"},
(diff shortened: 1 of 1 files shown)


pmario:
Is there a guarantee, that only 1 event fires. Otherwise we would get a "double" handling on other browsers.

Jermolene:
Thanks @saqimtiaz 

> Is there a guarantee, that only 1 event fires. Otherwise we would get a "double" handling on other browsers.

The `handleInputEvent` already ignores repeated calls to set the same value.

saqimtiaz:
Also note that Chrome and Firefox only fire change event on mouseup, and not on mousemove (correct behaviour as per specs).

So in these browsers there is one extra event fired now per completed move of the range slider (not per step), which is then debounced by handleInputEvent.

Internet Explorer fires change on mousemove and therefore can be used instead of input.

PR #4540, opened 2020-04-07
Feature has:index

joshuafontany:
It was noticed in the Google Groups that the `has` operator can specify field existence, but not index existence. I think this would add additional flexibility in using stock filters on json tiddlers in a vanilla wiki & it seems like a good core function (even if I slightly overwrite it in my own Json plugins).

Best,
Joshua Fontany


joshuafontany:
I'm not sure if I should update the docs in this branch within the ./editions/tw5.com-docs folder or do so with a different pull request to the `tiddlywiki-com` branch. I can get the `has Operator` doc tiddler updated if this is accepted.

pmario:
IMO docs for new features like this one should be part of the PR. ... But you should make sure first, that Jeremy wants to merge it. .. Otherwise you spend a lot of time for nothing.

Jermolene:
Hi @joshuafontany please do add docs - we do indeed include docs updates in feature PRs targetting master. I merge tiddlywiki-com back into master every so often to try to avoid conflicts.

joshuafontany:
Docs updated in the tw-com edition folder. :)

Jermolene:
Thanks @joshuafontany and sorry for the delay

PR #4552, opened 2020-04-11
Fix 4532 links add unwanted whitespace

### LINKED ISSUE #4532: Links in lists improperly formatted causing unwanted whitespace nodes
In UI list generators the `<$link>` HTML is improperly formatted. For example, in $:/core/ui/ListItemTemplate:
```
<div class="tc-menu-list-item">
<$link to={{!!title}}>
<$view field="title"/>
</$link>
</div>
```

This is causing browsers to collapse the newlines into whitespace nodes that cannot be removed with CSS. Browsers do this for inline elements, that includes `<a>` and `<li>` tags, and maybe more.

This particularly becomes a problem when adding `::before` or `::after` content to the generated links, as shown in the screenshot.

Annotation 2020-04-01 192426

In the link in the sidebar to the right, the spaces before and after "Custom CSS" should not be there.

This is how the code above should be formatted, just need to remove the newlines:
```
<div class="tc-menu-list-item">
<$link to={{!!title}}><$view field="title"/></$link>
</div>
```

Annotation 2020-04-01 193719

Here are some of the files I have found that have improper formatting, there may be more:

- $:/core/ui/ListItemTemplate
- $:/core/macros/timeline
- $:/core/ui/PluginListItemTemplate
- $:/core/ui/PluginInfo/Default/contents

pmario:
This PR fixes #4532 "links add unwanted whitespace", which causes problems with CSS styling

I did test with a stylesheet definition as follows, to see every link (and whitespace) in TW.

Every dropdown, theme, palette, sidebar-tab, ControlPanel tab, tag manager and so on was opened, to see if there are side effects. ... There are none!

```
a::before {
  content: "[b]"; 
}

a::after {
  content: "[a]";
}
```

# Not fixed: TOC macro

I just found out, that one more this macro needs some general fixing .. an other time.


pmario:
It seems I also missed the new "top menu bar" which is still broken with this PR

pmario:
@Jermolene should I fix it?

core/ui/ListItemTemplate.tid
```diff
@@ -1,7 +1,5 @@
 title: $:/core/ui/ListItemTemplate
 
 <div class="tc-menu-list-item">
-<$link to={{!!title}}>
-<$view field="title"/>
-</$link>
+<$link to={{!!title}}><$view field="title"/></$link>
```
twMat, line 4: Is there a reason to not merely use <$link/> ? (Or is perhaps !!title not always same as currentTiddler?)

core/ui/PluginInfoDefaultContents.tid
```diff
@@ -5,9 +5,7 @@ title: $:/core/ui/PluginInfo/Default/contents
 <ul>
 <$list filter="[all[current]plugintiddlers[]sort[title]]" emptyMessage=<<lingo Empty/Hint>>>
 <li>
-<$link to={{!!title}}>
-<$view field="title"/>
-</$link>
+<$link to={{!!title}}><$view field="title"/></$link>
```
twMat, line 8: same here

core/ui/ViewTemplate/subtitle.tid
```diff
@@ -1,10 +1,10 @@
 title: $:/core/ui/ViewTemplate/subtitle
 tags: $:/tags/ViewTemplate
 
+\whitespace trim
 <$reveal type="nomatch" stateTitle=<<folded-state>> text="hide" tag="div" retain="yes" animate="yes">
 <div class="tc-subtitle">
-<$link to={{!!modifier}}>
-<$view field="modifier"/>
-</$link> <$view field="modified" format="date" template={{$:/language/Tiddler/DateFormat}}/>
+<$link to={{!!modifier}}><$view field="modifier"/></$link>
```
twMat, line 7: <$link to={{!!modifier}}/>

pmario:
commit f2a613a removes whitespace from title and adds a little bit of margin to the tiddler icon, if there is one.

core/ui/ListItemTemplate.tid
pmario: hmm, should be faster too. Will do some testing. thx

Jermolene: @twMat's point is worth addressing I think

pmario: I'll add it. I did some testing already and everything seemed to be OK.

pmario:
@Jermolene did change code according to @twMat suggestions.

Jermolene:
Thanks @pmario

Issue #4545, opened 2020-04-09
Feature request: Create Plugin Library from other paths


kookma:
Referring to https://groups.google.com/d/msg/tiddlywiki/EMHF71bzlP0/kCyau0c5AwAJ it seems Tiddlywiki plugin library can be made only from plugins stored under:

`tidllywiki/plugins` folder. For example on Windows 10 if Tiddlywiki is stored in

D:\Tiddlywiki, then the plugins are in `D:\Tiddlywiki\plugins\` 

The recommended script to make plugin library using official pluginlibrary edition

```
tiddlywiki mylib --init pluginlibrary
```

as belows

```
 "build": {
        "library": [
            "--makelibrary",
            "$:/UpgradeLibrary",
            "--savelibrarytiddlers",
            "$:/UpgradeLibrary",
            "[prefix[$:/]] -[[$:/plugins/tiddlywiki/upgrade]] -[[$:/plugins/tiddlywiki/translators]] -[[$:/plugins/tiddlywiki/pluginlibrary]] -[[$:/plugins/tiddlywiki/jasmine]]",
            "recipes/library/tiddlers/",
            "$:/UpgradeLibrary/List",
            "--savetiddler",
            "$:/UpgradeLibrary/List",
            "recipes/library/tiddlers.json",
            "--rendertiddler",
            "$:/plugins/tiddlywiki/pluginlibrary/library.template.html",
            "index.html",
            "text/plain"
        ]
    }
```

On the other hand Tiddlywiki allows to store plugins on other paths: See https://tiddlywiki.com/prerelease/#Environment%20Variables%20on%20Node.js

The request is to be able to create library of plugins from customized path no only from installed path of Tiddlywiki on Node.js

pmario:
Use the environment variables and you are good to go. 

You can use it like: `tiddlywiki --build my-libs`

```
 "build": {
        "library": [
           ... default stuff removed to make the example smaller
        ],
        "my-libs": [
           ... here you can do whatever you want. It just needs to work. So you can use any paths that you need.
        ]
    }
```

pmario:
IMO there is no need for a new feature.

kookma:
Thanks @pmario 

I give a try and back to you!

kookma:
@pmario 
I have all developed library in

`D:\Tiddlywiki\plugins\kookma`

Should I repeat all the command in current  ` "library":`  here? I appreciate if you kindly give me an example

kookma:
[...]

PR #4559, opened 2020-04-13
makelibrary.js uses enviroment variable paths

inmysocks:
This makes makelibrary.js use environment variables to find paths for plugins, themes and languages instead of just using the paths hardcoded in boot.js

### DIFF
--- a/core/modules/commands/makelibrary.js
+++ b/core/modules/commands/makelibrary.js
@@ -51,9 +51,20 @@ Command.prototype.execute = function() {
 				}
 			}
 		};
-	collectPublisherPlugins(path.resolve($tw.boot.corePath,$tw.config.pluginsPath));
-	collectPublisherPlugins(path.resolve($tw.boot.corePath,$tw.config.themesPath));
-	collectPlugins(path.resolve($tw.boot.corePath,$tw.config.languagesPath));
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @inmysocks

kookma:
@Jermolene 

- Can we test this new feature?
- Is this address the issues raised here? https://github.com/Jermolene/TiddlyWiki5/issues/4545#issue-597224453

- Is there any documentation for this new feature? If I can be of any help please let me know

Jermolene:
> Can we test this new feature?

Yes, just download the latest "master" from https://github.com/Jermolene/TiddlyWiki5

> Is this address the issues raised here? #4545 (comment)

Yes, I believe so.

> Is there any documentation for this new feature? If I can be of any help please let me know

It's kind of a bug fix: we had already documented the environment variables as the means to introduce additional plugins to the Node.js configuration, but we weren't respecting those environment variables when building the plugin library.

The main docs we need is a simple guide to setting up a third party plugin library, but I'll have to return to that after v5.1.22 is released.

kookma:
Thanks @Jermolene 

I give several tries and it worked. This PR address 
 https://github.com/Jermolene/TiddlyWiki5/issues/4545 feature request

PR #4564, opened 2020-04-15
Updates for chinese editions

BramChen:
Fix wrong configurations in `tiddlywiki.info` for editions `zh-Hant` and `zh-Hans`


Jermolene:
Thanks @BramChen

PR #4567, opened 2020-04-15
Fix "has" filter to properly interpret an empty list field

saqimtiaz:
For a tiddler x with an empty "list" field the filter currently returns no result for the following: 
`[[x]!has[list]] `
whereas the expected result would be x

This is because the filter checks if the field is equal to an empty string, whereas the list field is represented by an empty array.
I am not set up to run the tests at this time, so welcome anyone that can do so and perhaps add some additional tests.


Jermolene:
Thanks @saqimtiaz I've merged #4643 from @bimlas instead.

PR #4573, opened 2020-04-17
Update Hidden Setting Disable Drag and Drop.tid

silvyn:
Add doc-from-version 5.1.22


Jermolene:
Thanks @silvyn and apologies for the delay. Docs updates should target the 'tiddlywiki-com' branch. Annoyingly, I can't change it so I've just duped the same changes.

Jermolene:
Thanks @silvyn and apologies for the delay. Docs updates should target the 'tiddlywiki-com' branch. Annoyingly, I can't change it so I've just duped the same changes.

PR #4579, opened 2020-04-19
Make input fields dismiss their popups when the ...

BurningTreeC:
... fields loose focus (`blur`)

BEFORE:

Peek 2020-04-19 11-38

AFTER:

Peek 2020-04-19 11-40


Jermolene:

BurningTreeC:
Thanks @Jermolene

BurningTreeC:
@Jermolene , this was a good attempt but causes problems with the search dropdown where links don't open anymore ... so this should be reverted

PR #4580, opened 2020-04-19
fix(syncer): Tiddler should not be a boolean

danielo515:
When a tiddler does not exist, it is skipt from the getting tiddler info.
I opted for this simpler version, but if it is imperative to run this for every tiddler I can come with a  different implementation like this:

```js
var tiddler = self.wiki.tiddlerExists(title) ? self.wiki.getTiddler(title) : new self.Tiddler({title})
```

Right now the code is passing a boolean when the output of the check is false, which is not correct in any way.

### DIFF
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -180,7 +180,8 @@ Syncer.prototype.readTiddlerInfo = function() {
 	var self = this,
 		tiddlers = this.getSyncedTiddlers();
 	$tw.utils.each(tiddlers,function(title) {
-		var tiddler = self.wiki.tiddlerExists(title) && self.wiki.getTiddler(title);
+		if(!self.wiki.tiddlerExists(title)) return;
+		var tiddler = self.wiki.getTiddler(title);
 		self.tiddlerInfo[title] = {
 			revision: self.getTiddlerRevision(title),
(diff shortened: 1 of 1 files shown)


danielo515:
@Jermolene seems that you are taking a loot at simple PRs, this is another one 😄

core/modules/syncer.js
```diff
@@ -180,7 +180,8 @@ Syncer.prototype.readTiddlerInfo = function() {
 	var self = this,
 		tiddlers = this.getSyncedTiddlers();
 	$tw.utils.each(tiddlers,function(title) {
-		var tiddler = self.wiki.tiddlerExists(title) && self.wiki.getTiddler(title);
+		if(!self.wiki.tiddlerExists(title)) return;
```
Jermolene, line 183: Just to note that our coding style is to always use braces around the bodies of "if" statements.

Jermolene:
Thanks @danielo515 I'm about to commit a different fix, let me know if it works for you

danielo515:
> Thanks @danielo515 I'm about to commit a different fix, let me know if it works for you

Yep, works perfectly for me.

Jermolene:
Great thanks @danielo515

PR #4583, opened 2020-04-20
fix: make makeLibrary.js account for non folders

danielo515:
It is too naive to not account for possible non folder files on any subdirectory. This is checking that each path is an actual folder

### DIFF
--- a/core/modules/commands/makelibrary.js
+++ b/core/modules/commands/makelibrary.js
@@ -35,8 +35,10 @@ Command.prototype.execute = function() {
 	var collectPlugins = function(folder) {
 			var pluginFolders = fs.readdirSync(folder);
 			for(var p=0; p<pluginFolders.length; p++) {
+				var currentFolder = path.resolve(folder,"./" + pluginFolders[p]);
+				if(!fs.statSync(currentFolder).isDirectory()) continue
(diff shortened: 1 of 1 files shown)


danielo515:
@Jermolene seems that you are looking at simple PRs, this is one 😄

core/modules/commands/makelibrary.js
```diff
@@ -35,8 +35,10 @@ Command.prototype.execute = function() {
 	var collectPlugins = function(folder) {
 			var pluginFolders = fs.readdirSync(folder);
 			for(var p=0; p<pluginFolders.length; p++) {
+				var currentFolder = path.resolve(folder,"./" + pluginFolders[p]);
+				if(!fs.statSync(currentFolder).isDirectory()) continue
```
Jermolene, line 39: As noted elsewhere, our coding style is to always use braces for the bodies of "if" statements.

Also, we avoid using "continue" in the core code because it makes code hard to follow.

Finally, there is a built in function for checking if a path is a directory: see https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/utils/filesystem.js#L142

Jermolene:
Thanks @danielo515, I'll commit an alternate fix shortly.

PR #4586, opened 2020-04-20
Revert "Make input fields dismiss their popups when the ..."

BurningTreeC:
Reverts Jermolene/TiddlyWiki5#4579


Jermolene:
Thanks @BurningTreeC

PR #4590, opened 2020-04-22
Make selections follow palette colors

BurningTreeC:


Jermolene:
Thanks @BurningTreeC, some people like their default OS of controls so I'm uncertain whether this is the way to go, or whether we should define new palette colours for the selection colours, but leave them undefined by default.  But let's merge it and see what people think.

BurningTreeC:
No risk no fun @Jermolene :grin: ... we can always revert

PR #4593, opened 2020-04-23
Fixing kinda-serious import bug.

flibbles:
So if you go to tiddlywiki.com and create a new tiddler with the following text:
```
<$set name=A value=A>

<$set name=B value=B>

Break!!!!
</$set></$set>
```
And you give this tiddler the tag `$:/tags/Macro`, the moment you hit save, Tiddlywiki will go straight to hell. (It literally goes away. Poof.)

This is my fault. But I've already got the fix here. My importvariables magic didn't actually support having more than one nonMacroDefinition variable imported.

My fix is strange, but I've added tests, and it's all working.

I'd like to point out again how utterly bizarre imported nonMacroDefinitions are. Look at this:
```
\define A() A content
<$set name=B value="B content"></$set>
```
Both A and B load when this is imported, right?

WRONG!!

The $set widget will be wrapped in a `<p>`, which prevents importing. Only "A" loads. You want:

```
\define A() A content
<$set name=B value="B content">

</$set>
```
Do you want a second nonMacro variable? This is how you do it.
```
\define A() A content
<$set name=B value="B content">

<$set name=C value="C content">

</$set></$set>
```
It has to be _inside_ the first.

Isn't that weird? Isn't that whack jack? The widget nature of $set makes you think it's variable should only be present INSIDE of it, but no. It gets picked up by import.

Also, the fact that I completely broke Tiddlywiki if you use 2 or more of them, and _nobody_ complained, kinda shows that nobody is using them, right?

...

Also, if you remove macros from a $:/tags/Macro tiddler. It wouldn't actually be gone until you refreshed the page. This was another problem I caused. I fixed and tested that too.

### DIFF
--- a/core/modules/widgets/importvariables.js
+++ b/core/modules/widgets/importvariables.js
@@ -71,6 +71,13 @@ ImportVariablesWidget.prototype.execute = function(tiddlerList) {
 					});
 				} else {
 					widgetPointer.makeChildWidgets([node]);
+					// No more regenerating children for
+					// this widget. If it needs to refresh,
+					// it'll do so along with the the whole
+					// importvariable tree.
+					if (widgetPointer != this) {
+						widgetPointer.makeChildWidgets = function(){};
+					}
 					widgetPointer = widgetPointer.children[0];
 				}
 				parseTreeNode = parseTreeNode.children && parseTreeNode.children[0];
--- a/core/modules/widgets/importvariables.js
+++ b/core/modules/widgets/importvariables.js
@@ -38,6 +38,8 @@ Compute the internal state of the widget
 */
 ImportVariablesWidget.prototype.execute = function(tiddlerList) {
 	var widgetPointer = this;
+	// Got to flush all the accumulated variables
+	this.variables = new this.variablesConstructor();
 	// Get our parameters
 	this.filter = this.getAttribute("filter");
 	// Compute the filter
@@ -70,7 +72,7 @@ ImportVariablesWidget.prototype.execute = function(tiddlerList) {
 						widgetPointer.variables[key] = widget.variables[key];
 					});
 				} else {
-					widgetPointer.makeChildWidgets([node]);
+					widgetPointer.children = [widgetPointer.makeChildWidget(node)];
 					// No more regenerating children for
 					// this widget. If it needs to refresh,
 					// it'll do so along with the the whole
--- a/editions/test/tiddlers/tests/test-widget.js
+++ b/editions/test/tiddlers/tests/test-widget.js
@@ -518,6 +518,41 @@ describe("Widget module", function() {
 		expect(wrapper.innerHTML).toBe("<p>A B C D</p>");
 	});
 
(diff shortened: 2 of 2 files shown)


saqimtiaz:
> Do you want a second nonMacro variable? This is how you do it.
> It has to be inside the first.

I banged my head against this one on pre-release a few weeks ago and it drove me batty. In the end I gave up assuming there was only support for importing one variable.

My use case was to transclude a tagged tiddler in a shadow tiddler and make a variable (calculated via a filter) available in it, without having to edit said shadow tiddler.

Jermolene:
Hi @flibbles hair-raising stuff! Thanks for this.

Jermolene:
Hi @flibbles some of the problems you mention stem from the fact that the importvariables widget parses macro tiddlers in block mode. I've experimented with forcing inline mode:

```
var parser = widgetPointer.wiki.parseTiddler(title,{parseAsInline: true});
```

Everything still seems to work, and it becomes possible to nest set widgets like this:

```
\define A() A content
<$set name=B value="B content">
<$set name=C value="C content">
</$set>
</$set>
```

Making this change wouldn't be strictly backwards compatible, but it does seem like it might reduce some of the gotchas.

flibbles:
Hold up, @saqimtiaz. Are you saying you actually use this feature? You import `$set` widgets?

How did you know that feature even existed? By accident? It's not documented as far as I can tell.

And the problem you're talking about is about how it only seemed to import the first $set? NOT how Tiddlywiki would completely break down if you tried?

saqimtiaz:
@flibbles tried to use it and failed. As I explained it would be useful for situations where you don't want to edit a shadow tiddler but still make a variable available (page template, story template etc). 

I was trying to import from a tiddler with multiple set widgets and when only the first seemed to work I deemed this unreliable. I do see utility in it, even though I agree that it's jarring that it works in content that's not wrapped in the widget. Or that the vars widget is left out in the cold and not recognized at all from what I remember.

As to how I discovered that this should work, I saw something about the import pragma for variables. This led to unanswered questions about whether that applied to macros, set widgets and the vars widget. So I looked up the import code and that's also how I first came across the ticket/pull request discussing the optimizations you made.

I run a significantly patched version of the core code, so am not always up to date with core changes as I need to merge. This could explain why I had issues with only the first set widgets being picked up, but not with any TW breakdowns.

flibbles:
Yeah, I agree it's not intuitive. I also only discovered it by working with the source.

I feel like if people want filtered or indirect values for \define, it should be like 

```
\define macro(){{{filter}}}
\define macro()<<macro>>
```
or maybe
```
\define macro{{{params}}} filter
\define macro filtered() filter
```
or _something_. Even if the existing feature were documented, I think it would confuse many.

Also, yeah, vars. If import used vars under the hood. Its implementation could be cleaner. It's the superior widget.
```

saqimtiaz:
Yeah, I would happily embrace a cleaner and more intuitive way to declare and import variables that have filtered values.

PR #4600, opened 2020-04-26
Fix #4591: Use text field instead of tags field for $:/config/NewJournal/Tags ...

### LINKED ISSUE #4591: [BUG] Journal tag pill dropdown list includes the system tiddler "$:/config/NewJournal/Tags"
**Bug description**
Clicking the Journal tag pill produces a dropdown list that wrongly includes a single system tiddler as the first tiddler titled: "$:/config/NewJournal/Tags" in TW5 v5.1.22 This behaviour seems to be restricted to only that action. Using list generating macros correctly excludes the system tiddler. Changing that system tiddler to remove the behavior breaks the creation of new Journal tiddlers.

**To Reproduce**
1. Create a non-system tiddler tagged Journal
2. Click on Journal tag pill
3. First entry shows ""$:/config/NewJournal/Tags""

**Expected behavior**
"$:/config/NewJournal/Tags" should not appear in the dropdown list.

**Screenshots**
EWNXSr0X0AAxAtX

BurningTreeC:
... and $:/config/NewTiddler/Tags

The tags input in $:/ControlPanel/ Info/ Basics/ is replaced by simple text input fields

### DIFF
--- a/core/language/en-GB/NewJournalTags.tid
+++ b/core/language/en-GB/NewJournalTags.tid
@@ -1,2 +1,2 @@
 title: $:/config/NewJournal/Tags
-tags: Journal
+text: Journal
--- a/core/language/en-GB/NewJournalTags.tid
+++ b/core/language/en-GB/NewJournalTags.tid
@@ -1,2 +1,3 @@
 title: $:/config/NewJournal/Tags
-text: Journal
+
+Journal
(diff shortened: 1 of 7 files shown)


core/language/en-GB/NewJournalTags.tid
```diff
@@ -1,2 +1,2 @@
 title: $:/config/NewJournal/Tags
-tags: Journal
```
pmario, line 2: hmmm, pls use 

```
title: $:/config/NewJournal/Tags

Journal
```

BurningTreeC: done @pmario

BurningTreeC:
bump. I think this should be changed

Jermolene:
Thanks @BurningTreeC

PR #4601, opened 2020-04-26
Use this.wiki where applicable

Arlen22:
There are a few places where the code references `$tw.wiki` in places where a local `this.wiki` is available. This PR changes those to use the local wiki reference. I haven't tested this. Hopefully there is a way that these can be tested.

### DIFF
--- a/core/modules/commands/fetch.js
+++ b/core/modules/commands/fetch.js
@@ -71,7 +71,7 @@ Command.prototype.fetchFiles = function(options) {
 	if(options.url) {
 		urls = [options.url]
 	} else if(options.urlFilter) {
-		urls = $tw.wiki.filterTiddlers(options.urlFilter);
+		urls = this.commander.wiki.filterTiddlers(options.urlFilter);
 	} else {
 		return "Missing URL";
 	}
--- a/core/modules/commands/savelibrarytiddlers.js
+++ b/core/modules/commands/savelibrarytiddlers.js
@@ -69,7 +69,7 @@ Command.prototype.execute = function() {
 		// Collect the skinny list data
 		var pluginTiddlers = JSON.parse(tiddler.text),
 			readmeContent = (pluginTiddlers.tiddlers[title + "/readme"] || {}).text,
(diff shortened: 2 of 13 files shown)


Arlen22:
`node tiddlywiki.js editions/test --listen root-tiddler=$:/core/save/all` seems to work properly.

pmario:
+1 ... The problem with our test wiki is, that ther isnt too much code coverage atm.

Arlen22:
Testing the changes. For most of these you should get a "cannot read property of undefined" error.

The changes in the server are critical and would be hit instantly. The change in syncer -- there is already another call which uses `this.wiki` two lines above it in the same context.

Here are a couple things I don't know how to test for sure. But one of you should be easily able to test them. 

- Use the `has` operator with suffix `index` to trigger the changed code path. The input must include tiddlers that exist. 
- These are part of the xlsx-import command and all changes here should be testable simply by attempting to import a workbook. 
  - `plugins/tiddlywiki/xlsx-utils/importer.js`
  - `plugins/tiddlywiki/xlsx-utils/xlsx-import-command.js`

I've manually double-checked the rest of these and they look good. I caught a few errors along the way. 

- This one I checked manually using VS Code since it's a closure.
  - `core/modules/upgraders/plugins.js`
- These are commands and refer to `this.commander.wiki`.
  - `core/modules/commands/fetch.js`
  - `core/modules/commands/savelibrarytiddlers.js`
- These are widgets and they properly inherit the base class
  - `plugins/tiddlywiki/innerwiki/innerwiki.js`
  - `plugins/tiddlywiki/railroad/wrapper.js`
- This is a parser and sets the wiki property in the constructor a few lines before.
  - `plugins/tiddlywiki/tw2parser/wikitextparser.js`
- There is one in `core/modules/wiki.js`. All the other functions in wiki.js use `this` to refer to the wiki object, so it seemed like an appropriate change.

Sorry, I think that's the best I can do right now for sake of time. I don't really know all the caveats of how to use most of these.

Jermolene:
Thanks @Arlen22 I've scanned through the changes for a quick check.

PR #4602, opened 2020-04-26
Separate $tw.boot.startup into three functions

Arlen22:
This PR is a change I have tried several times to work in and everytime there have been various problems with the way I tried to do it. So I've taken another shot at it using a different method. 

Because $tw.boot.startup is called asyncly and nothing depends on it, it is simple to override it when necessary so no other changes need to be made. 

The only change I made is to move the line `$tw.wiki = ...` down a few lines, but this should not change anything as none of the methods refer to $tw.wiki.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -2172,8 +2172,7 @@ $tw.loadTiddlersNode = function() {
 /*
 Startup TiddlyWiki
 */
-$tw.boot.startup = function(options) {
-	options = options || {};
+$tw.boot.initStartup = function(options) {
 	// Get the URL hash and check for safe mode
 	$tw.locationHash = "#";
 	if($tw.browser && !$tw.node) {
@@ -2285,8 +2284,6 @@ $tw.boot.startup = function(options) {
 	$tw.utils.registerFileType("application/x-bibtex","utf8",".bib");
 	$tw.utils.registerFileType("application/epub+zip","base64",".epub");
 	$tw.utils.registerFileType("application/octet-stream","base64",".octet-stream");
-	// Create the wiki store for the app
-	$tw.wiki = new $tw.Wiki();
 	// Install built in tiddler fields modules
 	$tw.Tiddler.fieldModules = $tw.modules.getModulesByTypeAsHashmap("tiddlerfield");
 	// Install the tiddler deserializer modules
@@ -2306,6 +2303,10 @@ $tw.boot.startup = function(options) {
 			return result;
 		}
 	}
+};
+$tw.boot.loadStartup = function(options){
+	// Create the wiki store for the app
+	$tw.wiki = new $tw.Wiki();
 	// Load tiddlers
 	if($tw.boot.tasks.readBrowserTiddlers) {
 		$tw.loadTiddlersBrowser();
@@ -2318,6 +2319,8 @@ $tw.boot.startup = function(options) {
 	}
 	// Give hooks a chance to modify the store
 	$tw.hooks.invokeHook("th-boot-tiddlers-loaded");
+}
+$tw.boot.execStartup = function(options){
 	// Unpack plugin tiddlers
 	$tw.wiki.readPluginInfo();
 	$tw.wiki.registerPluginTiddlers("plugin",$tw.safeMode ? ["$:/core"] : undefined);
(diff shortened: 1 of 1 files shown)


Arlen22:
Oops, moving $tw.wiki doesn't work, so I moved it back.

Jermolene:
Hi @Arlen22 if I'm understanding correctly, the objective is to be able to load tiddlers asynchronously in the browser? If so, the way I do that is to have some raw markup that loads the tiddlers and defers booting TiddlyWiki until the loading has completed. You can see an example at a xememex private wiki such as https://xememex.com/jeremy-private/index.html, but the following are the key steps.

First, we prepare the preloaded tiddlers array and stop TiddlyWiki from booting:

```
window.$tw = window.$tw || {};
$tw.boot = $tw.boot || {};
$tw.preloadTiddlers = $tw.preloadTiddlers || [];
$tw.boot.suppressBoot = true;
```

Then, we asynchronously push the tiddlers that are loaded with `$tw.preloadTiddlers.push()`, and when we're ready to boot TiddlyWiki we call `$tw.boot.boot();`.

(I think @danielo515 might do something similar).

Arlen22:
There are many scenarios where that works fine, but I want to do something that fits into the flow of the startup code a little better. Especially since I'm planning to use this with cloud connectors where the files are the same ones loaded with loadTiddlersNode() but they must be retrieved via web requests. 

I don't necessarily want to load the entire data folder beforehand since some folders (such as the files folder) will not get loaded, but I can't know beforehand which ones loadTiddlersNode() would actually load because of the dynamic nature of the loading sequence. 

And I also want to use all of the file types and other parameters registered in the first part of `$tw.boot.startup` since I want this to be as seamless as it can possibly be and require minimal updates for every new TiddlyWiki version. 

Besides, preloadTiddlers only takes JSON objects, not raw files in a file system, which is what I'm working with. I feel like having this option would data folders more useful.

Jermolene:
Hi @Arlen22 that makes sense I'll look over the PR shortly.

> Besides, preloadTiddlers only takes JSON objects, not raw files in a file system, which is what I'm working with

I'm not sure what you mean here?

Arlen22:
> I'm not sure what you mean here?

If I would just use preloadTiddlers, I still have to parse the tiddler from the file system, which means I basically need all the code to run before loadTiddlersNode() anyway, because it sets up the tiddler parsers and all that. The preload tiddlers array takes a JSON fields object, not raw .tid files.

danielo515:
I don't know if it is that I didn't slept enough, but I can  not make sense of the pull request description.
 In Noteself  we do is somethijg similar to what @jernolene has described. All tidlers are loaded asynchronously, but there are certain tiddlers that need to be loaded before tw boots. So what we do is suppress boot, load the tiddlers, inject them and then boot tw. This has worked well for 5 years so far, so I am not sure what is the objective of this change

Arlen22:
> So what we do is suppress boot, load the tiddlers, inject them and then boot tw. This has worked well for 5 years so far, so I am not sure what is the objective of this change

The problem is your tiddlers are already in JSON format. I'm dealing with .tid files that I need to load into TiddlyWiki the normal way, but I have to rewrite part of it to load them from the internet instead of the file system. Otherwise it is identical, but it has to be async. The startup function does a lot of heavy lifting before and after loading the tiddlers and I don't want to duplicate all that code and then make sure I'm keeping it up to date every time a new version comes out. That's too much duplicate code. 

I've run into this many times, and I really want a way that I can make data folder loading async without duplicating the whole startup process. This PR allows me to do that.

danielo515:
> The problem is your tiddlers are already in JSON format. I'm dealing with .tid files that I need to load into TiddlyWiki the normal way, but I have to rewrite part of it to load them from the internet instead of the file system.

What about injecting them into the DOM and then let tiddlywiki boot? I think you can just inject the content of .tid files in a random container (like a div) and that should work.

danielo515:
> The startup function does a lot of heavy lifting before and after loading the tiddlers and I don't want to duplicate all that code and then make sure I'm keeping it up to date every time a new version comes out.

That's true. But the real improvement (from my point of view) will be if tiddlywiki exposed more of its internals as usable functions, so it can be used as a library and hence allowing more flexible usages of it.

Arlen22:
>What about injecting them into the DOM and then let tiddlywiki boot? I think you can just inject the content of .tid files in a random container (like a div) and that should work.

That's too much of a work around. My goal with this PR is to modify boot.js in such a way that I don't need workarounds to do this.

>>The startup function does a lot of heavy lifting before and after loading the tiddlers and I don't want to duplicate all that code and then make sure I'm keeping it up to date every time a new version comes out.
>
>That's true. But the real improvement (from my point of view) will be if tiddlywiki exposed more of its internals as usable functions, so it can be used as a library and hence allowing more flexible usages of it.

That's where I'm headed with this, but I've opened Issues and submitted PRs since TiddlyWiki Classic and I've found that the quickest way forward is usually small changes that give a lot of gain. Big changes rarely make it past the review process.

I'm sorry if I sound aggressive in any way. I really appreciate your contribution to the community, and especially the work you pioneered with NoteSelf.

Arlen22:
@Jermolene how soon do you think this can be merged in?

Jermolene:
Thanks @Arlen22 and apologies for the delay.

PR #4603, opened 2020-04-27
Fix CamelCase issue for TextEditor/Snippet docs

sukima:
Prior to this change, the docs on the site said to use the tag `$:/tags/Texteditor/Snippet` but TiddlyWiki expects it to be `$:/tags/TextEditor/Snippet` (note the casing)


Jermolene:
Thanks @sukima and apologies for the delay

PR #4606, opened 2020-04-29
Update KeyboardWidget.tid

twMat:
Add railroad for key strings.
I don't know the formatting rules so may need to be revised to fit coding standards.


Jermolene:
Thanks @twMat

Issue #4605, opened 2020-04-29
[Question/Bug] TW refreshes every 20-50s


jordifradera:
I'm not sure this is a bug or an intended behavior of TW, but it makes it difficult to work in a multi-user environment. The problem is the following:

- TW is hosted in a server inside an intranet
- TW safe content to the server (no gitHub or gitLab or giTea or...)
- All users browse as (anon)
- Every 20-50s the interface refreshes on its own showing a layout
- The layout is the one corresponding to a user loged-in as writer
- When another user logs-in as writer the interfaces does not refresh, it only happens in Read-Only

I use TW as a collaborative tool, so all users with write permission add content to projects. The rest of users browse TW for info on that projects. Hence, it is annoying for (anon) users to go to a layout other than theirs every now and then.

I've been trying to find a parameter in tools than can change this behavior with no luck.

The following is the list of plugins in case there is any incompatibility:

TiddlyWiki5 core 5.1.22
 encryptTiddler 2.1
 datepicker 0.5.2
 moment 2.24.0
 bibtex 5.1.21
 comments 5.1.21
 d3 5.1.21
 Filesystem 5.1.22
 Highlight 5.1.22
 katex 5.1.21
 TiddlyWeb 5.1.22

All users browse with Chrome.

Thanks in advance.

saqimtiaz:
@Jermolene this is resolved with the fix for system tiddler sync.

saqimtiaz:
@Jermolene bump. Is resolved and can be closed.

Jermolene:
Thanks @saqimtiaz

Issue #4608, opened 2020-04-29
[IDEA] Is now the time to add server sent events to core?


Arlen22:
@Jermolene 

With the recent improvements to the syncer, is it time to add server-sent-events? I don't know what all the use cases would be, but I think it would be useful, as I'm sure you well know. 

We can add it as a route to the Server. I would propose setting aside the `events` route for it. Do we have all events go through one channel or do we allow multiple channels as URLs under `events`?

Server Sent Events allows named events to be sent in addition to the unnamed "message" event. However, I would proposed that each plugin have its own route under events named according to the plugin scope. This way every plugin can own its own channel. 

- `events/core`
- `events/plugins/tiddlywiki/tiddlyweb`
- `events/plugins/arlen22/plugin-name`
- `events/themes/someone/dashboard`

If that's too many connections, we could also have the plugins use the same prefixes for the event names themselves, but it seems like it would be more useful to have separate channels. Also, I don't think a large number of connections is really a problem. Node is capable of handling "1 million concurrent connections". 

We can make a `ServerSentEvents` class which would handle all of this so that plugins could easily setup their own route and would include instructions on the exact parameters to use. Regex can be compiled from a string, so we can set it up based on constructor parameters. 

Please see https://github.com/Jermolene/TiddlyWiki5/pull/4609 for the code that was included here.

joshuafontany:
@Arlen22 , I am looking at implementing an oEmbed library to  TW on Node.js, and this seems exactly what I want. The Server has to send/handle the requests for the Embed json, in order to avoid CORS errors. The Node library I want to use has smart white-listing and a few other methods to handle common oEmbed issues.

I am looking through your code now. What would be the best way to implement this, as its own plugin with its own named route, like `events/plugins/joshuafotnany/oembed`? I will start with reading your tiddlyweb implementation. It would be very useful to have a test or demo setup to see this in action. Thanks!

Best,
Joshua F

NicolasPetton:
@Arlen22 @Jermolene now that #5279 is merged, should we close this one?

Jermolene:
Thanks @NicolasPetton

PR #4609, opened 2020-04-29
Add Server Sent Events

Arlen22:
This is related to Issue https://github.com/Jermolene/TiddlyWiki5/issues/4608 and also fixes Issue https://github.com/Jermolene/TiddlyWiki5/issues/4605. 

To test this, run `node tiddlywiki.js editions/server --listen` then open two pages side-by-side and watch changes propagate in almost real time.


Arlen22:
I used the TypeScript compiler to convert it to ES5, so I'm sorry if it doesn't look great.

Arlen22:
Tested this on TiddlyServer and after fixing this one bug, everything works on TiddlyServer as well, without any changes to TiddlyServer.

plugins/tiddlywiki/tiddlyweb/sse-server.js
```diff
@@ -0,0 +1,62 @@
...
+ * }[]>} */
+var wikis = {};
+/**
+ * Setups up the array for this wiki and adds the change listener
+ * 
+ * @param {$tw.Wiki} wiki The wiki object to listen to changes on
+ */
+function setupWiki(wiki) {
+  // add a new array for this wiki (object references work as keys)
+  wikis[wiki] = [];
```
hoelzro, line 29: I think this would work ok for a single wiki, but IIRC JavaScript will stringify the object reference, so you'll end up with something like `wikis['[object Object]'] = []`, right?

plugins/tiddlywiki/tiddlyweb/sse-client.js
```diff
@@ -0,0 +1,33 @@
...
+  if($tw.syncadaptor.name !== "tiddlyweb") return;
+  //get the mount point in case a path prefix is used
+  var host = $tw.syncadaptor.getHost();
+  //make sure it ends with a slash (it usually does)
+  if(!host.endsWith("/")) host += "/";
+  //setup the event listener 
+  var events = new EventSource(host + "events/plugins/tiddlywiki/tiddlyweb");
+  events.addEventListener("change", function () {
+    $tw.syncer.syncFromServer();
+  });
```
hoelzro, line 30: Would it be a good idea to add some debouncing here to reduce the amount of requests going to the server over a small period of time?

hoelzro: Another idea - you might want to add an event listener for the `error` event.  Otherwise if you lose the connection to the server, you'll need to reload the entire wiki to reestablish it!

hoelzro:
Hi @Arlen22 - this is great! I've been hoping for support for SSE in TiddlyWiki for a while, and I agree that this would be a good time to add it.  I looked at your code and left some ideas I have for improvement - please let me know what you think!

Also, you mentioned how you originally wrote this in TypeScript - do you by chance have a TypeScript definition file for TiddlyWiki's API?  I think it would be very useful!

Arlen22:
> Also, you mentioned how you originally wrote this in TypeScript - do you by chance have a TypeScript definition file for TiddlyWiki's API? I think it would be very useful!

I do not. Because of how often closures are used it is very hard to make it work properly with TypeScript. Plus I don't know how to keep it in sync. It seems like it would be a full time job. I have often wished, though :)

plugins/tiddlywiki/tiddlyweb/sse-server.js
Arlen22: You're right. Many thanks. Fixed.

Arlen22:
@Jermolene any idea when we can get this in?

Arlen22:
No. Jeremy has said that it would be ideal to have SSE in core, and the other two files are tiddlyweb specific so can't really separate them out.

Jermolene:
Thank you @Arlen22 @hoelzro, and apologies for the slow response. There are some coding standards issues that I'll mark up shortly.

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+module-type: library
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/**
+ * @param {string} prefix
```
Jermolene, line 13: The core doesn't use JSDoc, just plain text for code comments

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+      throw new Error("data must be a single-line string");
+    response.write("event: " + event + "\ndata: " + data + "\n\n", "utf8");
+  };
+  ServerSentEvents.prototype.end = function (response) {
+    response.end();
+  };
+  return ServerSentEvents;
+}());
+
+exports.ServerSentEvents = ServerSentEvents;
```
Jermolene, line 72: The construction used to export the class isn't consistent with the rest of the core, which doesn't use the inner IIFE that returns "ServerSentEvents". It should follow the same approach as, say, the core widget classes.

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+ * comes in with the request and state from the
+ * route and an emit function to call.
+ */
+var ServerSentEvents = /** @class */ (function () {
+
+  function ServerSentEvents(prefix, handler) {
+    this.handler = handler;
+    this.prefix = prefix;
+  }
+  ServerSentEvents.prototype.getExports = function () {
```
Jermolene, line 32: The core leaves a blank line between methods

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+    };
+  };
+  /**
+   *
+   * @param {import("http").IncomingMessage} request
+   * @param {import("http").ServerResponse} response
+   * @param {*} state
+   */
+  ServerSentEvents.prototype.handleEventRequest = function (request, response, state) {
+    if (request.headers.accept && request.headers.accept.startsWith('text/event-stream')) {
```
Jermolene, line 47: The core doesn't use a space between `if` and the following parenthesis

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+      response.writeHead(200, {
+        'Content-Type': 'text/event-stream',
```
Jermolene, line 49: The core prefers double quotes for string constants, only using single quotes for strings that need to contain double quotes.

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+        'Connection': 'keep-alive'
+      });
+      this.handler(request, state, this.emit.bind(this, response), this.end.bind(this, response));
+    } else {
+      response.writeHead(406, "Not Acceptable", {});
+      response.end();
+    }
+  };
+  ServerSentEvents.prototype.emit = function (response, event, data) {
+    if (typeof event !== "string" || event.indexOf("\n") !== -1)
```
Jermolene, line 60: The core always wraps the body of an `if` statement in curly braces

core/modules/server/server-sent-events.js
```diff
@@ -0,0 +1,74 @@
...
+      throw new Error("type must be a single-line string");
```
Jermolene, line 61: We would normally capitalise an error message

plugins/tiddlywiki/tiddlyweb/sse-client.js
```diff
@@ -0,0 +1,41 @@
...
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.name = "/events/plugins/tiddlywiki/tiddlyweb";
+exports.after = ["startup"];
+exports.synchronous = true;
+exports.platforms = ["browser"];
+exports.startup = function () {
+  //make sure we're actually being used
```
Jermolene, line 20: Comments should be capitalised, with a space after the `//`

plugins/tiddlywiki/tiddlyweb/sse-client.js
```diff
@@ -0,0 +1,41 @@
...
+  if($tw.syncadaptor.name !== "tiddlyweb") return;
```
Jermolene, line 21: Could we instead install the event handler when we startup the syncadaptor?

plugins/tiddlywiki/tiddlyweb/sse-client.js
```diff
@@ -0,0 +1,41 @@
...
+  //get the mount point in case a path prefix is used
+  var host = $tw.syncadaptor.getHost();
+  //make sure it ends with a slash (it usually does)
+  if(!host.endsWith("/")) host += "/";
```
Jermolene, line 25: We try to target ECMAScript 5 to maintain support for older Android browsers, and so we avoid `endsWith`

plugins/tiddlywiki/tiddlyweb/sse-client.js
Jermolene: I think debouncing is going to be necessary to avoid obvious DDOS vectors. Alternatively, perhaps we should disable SSE by default.

Jermolene (commented):
Thanks @Arlen22. In most cases I've just marked up the first occurrence of repeated issues.

plugins/tiddlywiki/tiddlyweb/sse-client.js
Arlen22: I'm sure we could but I don't really know where. I did it here because that is right after the syncer and sync adapter are initialized. But you could just as easily do it there.

core/modules/server/server-sent-events.js
Arlen22: I use it to give VS Code the proper intellisense. TiddlyWiki is really hard to work with otherwise.

Arlen22:
I'm sure I missed stuff. I don't know really. I've kind of lost interest. Feel free to use it if you feel it's useful. I use VS Code and Typescript for several years now and depend on it for a lot of stuff. So it's kind of hard to work with the internals of TiddlyWiki code because the wrapper functions block the type detection.

Arlen22:
@NicolasPetton I appreciate the work here. I'll close mine as it looks like you've taken care of everything.

core/modules/server/server-sent-events.js
linonetwo: Yes, for us familiar with type script, interface with typing is very nice.

Issue #4618, opened 2020-04-30
[BUG] evalSandboxed causes server to take 4x more memory


Arlen22:
**Describe the bug**

The use of evalSandboxed in TiddlyWiki causes it to take 85 MB of memory, whereas changing it to evalGlobal uses only 20 MB. Are there any particular reasons for using evalSandboxed vs evalGlobal? 

**To Reproduce**

1. Run `node --inspect-brk tiddlywiki.js editions/server --listen`
2. In Google Chrome, navigate to `chrome://inspect` and open the devtools for the node instance. 
3. Take a heap snapshot or simply note the amount of consumed memory on the memory tab.
4. Press run in source code (because `--inspect-brk` starts with it paused)
5. Note the amount of memory now. It should be 80 MB or so. 

**Desktop (please complete the following information):**
 - OS: MacOS
 - NodeJS 12.14.1
 - Version: master

Arlen22:
Here's another idea. This sandboxes all modules together in one context, but still keeps them separate from the main context. 

```diff
diff --git a/boot/boot.js b/boot/boot.js
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -501,7 +501,7 @@ $tw.utils.getTypeEncoding = function(ext) {
 /*
 Run code globally with specified context variables in scope
 */
-$tw.utils.evalGlobal = function(code,context,filename) {
+$tw.utils.evalGlobal = function(code,context,filename,sandbox) {
 	var contextCopy = $tw.utils.extend(Object.create(null),context);
 	// Get the context variables as a pair of arrays of names and values
 	var contextNames = [], contextValues = [];
@@ -516,19 +516,21 @@ $tw.utils.evalGlobal = function(code,context,filename) {
 	if($tw.browser) {
 		fn = window["eval"](code + "\n\n//# sourceURL=" + filename);
 	} else {
-		fn = vm.runInThisContext(code,filename);
+		if(sandbox){
+			fn = require('vm').runInContext(code, sandbox, filename)
+		} else {
+			fn = vm.runInThisContext(code,filename);
+		}
 	}
 	// Call the function and return the exports
 	return fn.apply(null,contextValues);
 };
-
+var sandbox = vm.createContext(); 
 /*
 Run code in a sandbox with only the specified context variables in scope
 */
 $tw.utils.evalSandboxed = $tw.browser ? $tw.utils.evalGlobal : function(code,context,filename) {
[...]

PR #4624, opened 2020-04-30
Change evalSandboxed to use only one context for all modules

Arlen22:
This fixes https://github.com/Jermolene/TiddlyWiki5/issues/4618. Modules may still be sandboxed in their own context as before, if required. Jasmine is the only case I ran into. 

These all work.

```sh
node tiddlywiki.js editions/test
node tiddlywiki.js editions/full
node tiddlywiki.js editions/tw5.com-server
```

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -510,11 +510,11 @@ var polyfill =[
 	"(function() {",
 	"	if (typeof globalThis === 'object') return;",
 	"	// node.green says this is available since 0.10.48",
-	"	Object.prototype.__defineGetter__('__wow__', function() {",
+	"	Object.prototype.__defineGetter__('__temp__', function() {",
 	"		return this;",
 	"	});",
-	"	__wow__.globalThis = __wow__; // lolwat",
-	"	delete Object.prototype.__wow__;",
+	"	__temp__.globalThis = __temp__;",
+	"	delete Object.prototype.__temp__;",
 	"}());"
 ].join("\n");
 
@@ -545,7 +545,7 @@ $tw.utils.evalGlobal = function(code,context,filename,sandbox,allowGlobals) {
 		"  return exports;\n",
 		"})"
 	].join("\n");
-	// code = "(function(" + contextNames.join(",") + ") {(function(){  \n" + code + "\n;})();\nreturn exports;\n})\n";
+
 	// Compile the code into a function
 	var fn;
 	if($tw.browser) {
@@ -565,18 +565,12 @@ $tw.utils.sandbox = !$tw.browser ? vm.createContext({}) : undefined;
 Run code in a sandbox with only the specified context variables in scope
 */
 $tw.utils.evalSandboxed = $tw.browser ? $tw.utils.evalGlobal : function(code,context,filename,allowGlobals) {
-	return $tw.utils.evalGlobal(code,context,filename,(allowGlobals ? vm.createContext({}) : $tw.utils.sandbox),allowGlobals);
+	return $tw.utils.evalGlobal(
+		code,context,filename,
+		allowGlobals ? vm.createContext({}) : $tw.utils.sandbox,
+		allowGlobals
+	);
 };
-
-// if(!$tw.browser) $tw.utils.evalSandboxed(`
-//   setInterval(function(){
-// 		console.log(globalThis)
-// 		if(Object.keys(globalThis).length){
-// 			console.log(globalThis);
-// 			throw "Global assignment is not allowed within modules on node.";
-// 		}
-// 	}, 1000);
-// `, { setInterval, exports: {}, console }, "$:/boot/boot.js-global-monitor");
 
 /*
 Creates a PasswordPrompt object
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -23,12 +23,7 @@ $tw.utils = $tw.utils || Object.create(null);
 
 /////////////////////////// Standard node.js libraries
 
-/** @type {import{"fs"}} */
-var fs;
-/** @type {import{"path"}} */
-var path;
-/** @type {import{"vm"}} */
-var vm;
+var fs, path, vm;
 if($tw.node) {
 	fs = require("fs");
 	path = require("path");
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -503,6 +503,27 @@ $tw.utils.getTypeEncoding = function(ext) {
 	return typeInfo ? typeInfo.encoding : "utf8";
 };
 
+var polyfill = `
+// this polyfills the globalThis variable
+// using the this variable on a getter 
+// inserted into the prototype of globalThis
+(function() {
+	if (typeof globalThis === 'object') return;
+	// node.green says this is available since 0.10.48
+	Object.prototype.__defineGetter__('__wow__', function() {
+		return this;
+	});
+	__wow__.globalThis = __wow__; // lolwat
+	delete Object.prototype.__wow__;
+}());
+`;
+var globalCheck = `{
+	if(Object.keys(globalThis).length){
+		console.log(Object.keys(globalThis));
+		throw "Global assignment is not allowed within modules on node.";
+	}
+}`;
+
 /*
 Run code globally with specified context variables in scope
 */
@@ -515,17 +536,12 @@ $tw.utils.evalGlobal = function(code,context,filename,sandbox,allowGlobals) {
 		contextValues.push(value);
 	});
 	// Add the code prologue and epilogue
-	// and the sandboxed global check
 	code = `
+	${!$tw.browser ? polyfill : ""}
+	//Now we run the module and check the output
 	(function(${contextNames.join(",") }) {
 		(function(){${code};})();
-		${(!$tw.browser && sandbox && !allowGlobals) ? `
-			let globe = typeof global !== "undefined" ? global : this;
-			if(Object.keys(globe).length){
-				console.log(globe);
-				throw "Global assignment is not allowed within server modules.";
-			}
-		` : ""}
+		${(!$tw.browser && sandbox && !allowGlobals) ? globalCheck : ""}
 		return exports;
 	})
 	`
@@ -544,13 +560,23 @@ $tw.utils.evalGlobal = function(code,context,filename,sandbox,allowGlobals) {
 	// Call the function and return the exports
 	return fn.apply(null,contextValues);
 };
-var sandbox = !$tw.browser ? vm.createContext() : undefined; 
+$tw.utils.sandbox = !$tw.browser ? vm.createContext({}) : undefined; 
 /*
 Run code in a sandbox with only the specified context variables in scope
 */
 $tw.utils.evalSandboxed = $tw.browser ? $tw.utils.evalGlobal : function(code,context,filename,allowGlobals) {
-	return $tw.utils.evalGlobal(code,context,filename,(allowGlobals ? vm.createContext() : sandbox),allowGlobals);
+	return $tw.utils.evalGlobal(code,context,filename,(allowGlobals ? vm.createContext({}) : $tw.utils.sandbox),allowGlobals);
 };
+
+// if(!$tw.browser) $tw.utils.evalSandboxed(`
+//   setInterval(function(){
+// 		console.log(globalThis)
+// 		if(Object.keys(globalThis).length){
+// 			console.log(globalThis);
+// 			throw "Global assignment is not allowed within modules on node.";
+// 		}
+// 	}, 1000);
+// `, { setInterval, exports: {}, console }, "$:/boot/boot.js-global-monitor");
 
 /*
 Creates a PasswordPrompt object
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -552,7 +552,6 @@ $tw.utils.evalGlobal = function(code,context,filename,sandbox,allowGlobals) {
 		fn = window["eval"](code + "\n\n//# sourceURL=" + filename);
 	} else {
 		if(sandbox){
-			console.log(code);
 			fn = vm.runInContext(code,sandbox,filename)
 		} else {
 			fn = vm.runInThisContext(code,filename);
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -503,26 +503,27 @@ $tw.utils.getTypeEncoding = function(ext) {
 	return typeInfo ? typeInfo.encoding : "utf8";
 };
 
-var polyfill = `
-// this polyfills the globalThis variable
-// using the this variable on a getter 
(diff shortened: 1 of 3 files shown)


Arlen22:
@Jermolene, so it turns out it's pretty easy to check for global pollution because there are no properties on the global context itself, since they are now passed in through the code. So after the module is done loading it checks to make sure there are no global variables. 

This isn't going to catch everything, though since the global could get polluted later. But I just realized that `"use strict"` prevents this from happening. Should we make that a required feature? 

> 
> Strict mode makes it easier to write "secure" JavaScript.
> 
> Strict mode changes previously accepted "bad syntax" into real errors.
> 
> As an example, in normal JavaScript, mistyping a variable name creates a new global variable. In strict mode, this will throw an error, making it impossible to accidentally create a global variable.
> 
> In normal JavaScript, a developer will not receive any error feedback assigning values to non-writable properties.
> 
> In strict mode, any assignment to a non-writable property, a getter-only property, a non-existing property, a non-existing variable, or a non-existing object, will throw an error.
>

Arlen22:
Nevermind, "use strict" definitely breaks stuff. Not sure why but I'm not going to go that route.

Arlen22:
So any file that specifies "use strict" won't accidentally pollute the global scope. We can turn this on globally in the module wrapper code, and that would probably solve all our problems. Or we can just turn it on in node and leave it off in the browser. 

But that breaks highlight.js, which is a [known issue](https://github.com/highlightjs/highlight.js/issues/2234) and may be resolved in the latest version, although I haven't tested it. 

"use strict" also removes the global variable reference, so I polyfilled `globalThis` using [this article](https://mathiasbynens.be/notes/globalthis). I could also polyfill `global`, and I think I will for consistency, although it has very little use since it is sandboxed.

Arlen22:
I think for now I'm going to leave out "use strict" as is the current case. The globalThis polyfill is required for the check, since it was only added in node v10. If we don't need the check, we also don't need the polyfill.

pmario:
@Arlen22 ... Just be sure, that you also run the `testcommonjs` edition tests with your changes.

Arlen22:
I searched the repo for all occurrences of `$tw.utils.evalSandboxed` and it appears like I covered everything. I had made this change in TiddlyServer in the past, and never got any complaints related to it, so I don't think it will break anything in general usage. I think most people use the `$tw` variable if they want to make something global.

boot/boot.js
```diff
@@ -510,26 +536,47 @@ $tw.utils.evalGlobal = function(code,context,filename) {
 		contextValues.push(value);
 	});
 	// Add the code prologue and epilogue
-	code = "(function(" + contextNames.join(",") + ") {(function(){\n" + code + "\n;})();\nreturn exports;\n})\n";
+	code = `
```
pmario, line 539: I think this type of templating isn't allowed in TW. TW is ES5 only atm.

Arlen22: Oops, thanks. Fixed.

Arlen22:
> @Arlen22 ... Just be sure, that you also run the `testcommonjs` edition tests with your changes.

That runs correctly. Everything says PASS or DONE.

Arlen22:
This change brings the startup time down from 1 second to about 400 ms. Bundling the core and plugins brings it down to about 200 ms. That's pretty fast.

Arlen22:
I wonder if we should remove the global check. It's not going to catch everything, and it's going to be hard to track down offenders because you can't really tell who added it other than by searching the code base for that string. Also, "use strict" prevents accidental assignment. I think it would be better to leave it there commented out so if someone needs to track something down they can do that. I guess this is a really good indicator of bugs, so maybe leaving it in will force the users to open support tickets with developers. What are your thoughts?

Arlen22:
Actually, I think it would be better to leave it in.

Arlen22:
I've been running the bundled version of this for the past week, and it is incredibly fast. The bundled version decreases the number of files that need to be read from disc, but it's unrelated to this. But both of them together definitely feel snappy. I'm regularly seeing boot times around 100ms which is way lower than previously.

Arlen22:
I dug around a bit and there aren't really any other optimizations that can be made on the client side as far as memory goes. It's already 20 MB because there is no partitioning like on the server. There is only one retained instance of the core tiddler text, so nothing to improve there. 

I think most of the memory (beyond 5 MB of strings in empty.html) is consumed by the various states maintained by widgets and parse nodes. All the little pieces that fit together each take a tiny bit of memory that adds up. I don't know if there are any easy improvements there. 

@pmario I checked the closures by stepping through on the browser side, and there are no extra closures. The module context variables are also optimized on the function call so only the ones that are used actually get retained in the context. Not that it really matters since they're all object references. 

The Gmail client is 190 MB, so we're doing quite a bit better than that.

Arlen22:
It also passes `node tiddlywiki editions/testcommonjs` with this in tiddlywiki.js

`$tw.boot.boot(() => {$tw.modules.execute('allTests')});`

pmario:
>@pmario I checked the closures by stepping through on the browser side, and there are no extra closures. The module context variables are also optimized on the function call so only the ones that are used actually get retained in the context. Not that it really matters since they're all object references.

thx for having a closer look. As I wrote it was quite some time ago, as I stumbled upon a difficulty, while single stepping through some code. ... I can't remember it anymore. ...

Jermolene:
Thanks @Arlen22 and apologies for the delay. The improvements sound worthwhile but I was concerned to see the change to `core/modules/startup/css-escape-polyfill.js`; does it imply that this change is not 100% backwards compatible?

Arlen22:
The code wasn't running at the correct time. It was running during file evaluation instead of when the module was executed. In this case it is a browser polyfill, which does not conflict because it only checks for globals when `!$tw.browser`.

Arlen22:
It should be backward compatible because before this any global assignments would stay inside that module. The only time we needed to allow it was for Jasmine to run tests. So that is the only scenario in which it is different. Other than that it is backward compatible. If someone wants to allow globals, they can add the fourth argument `true` to the evalSandboxed call.

boot/boot.js
```diff
@@ -23,7 +23,12 @@ $tw.utils = $tw.utils || Object.create(null);
 
 /////////////////////////// Standard node.js libraries
 
-var fs, path, vm;
```
Jermolene, line 26: For consistency, please revert to a single line `var` declaration.

boot/boot.js
```diff
@@ -498,10 +503,32 @@ $tw.utils.getTypeEncoding = function(ext) {
...
+	"// this polyfills the globalThis variable",
+	"// using the this variable on a getter ",
+	"// inserted into the prototype of globalThis",
+	"(function() {",
+	"	if (typeof globalThis === 'object') return;",
+	"	// node.green says this is available since 0.10.48",
+	"	Object.prototype.__defineGetter__('__wow__', function() {",
+	"		return this;",
+	"	});",
+	"	__wow__.globalThis = __wow__; // lolwat",
```
Jermolene, line 516: Rather than `__wow__` could we use a descriptive variable name? If the "lolwat" comment doesn't serve a purpose then it should be removed too.

boot/boot.js
```diff
@@ -498,10 +503,32 @@ $tw.utils.getTypeEncoding = function(ext) {
...
+	"	delete Object.prototype.__wow__;",
+	"}());"
+].join("\n");
+
+var globalCheck =[
+	"  if(Object.keys(globalThis).length){",
+	"    console.log(Object.keys(globalThis));",
+	"    throw \"Global assignment is not allowed within modules on node.\";",
+	"  }"
+].join('\n');
```
Jermolene, line 526: Am I right in understanding that this polyfill and globalcheck are needed as part of attempts to trap assignments to global variables? If so, perhaps it's better to drop them in the interests of simplicity.

boot/boot.js
```diff
@@ -498,10 +503,32 @@ $tw.utils.getTypeEncoding = function(ext) {
...
 };
 
+var polyfill =[
+	"// this polyfills the globalThis variable",
+	"// using the this variable on a getter ",
+	"// inserted into the prototype of globalThis",
+	"(function() {",
+	"	if (typeof globalThis === 'object') return;",
+	"	// node.green says this is available since 0.10.48",
+	"	Object.prototype.__defineGetter__('__wow__', function() {",
```
Jermolene, line 513: We generally use `Object.defineProperty()`; is there a reason for using `__defineGetter__` here?

boot/boot.js
```diff
@@ -510,26 +537,46 @@ $tw.utils.evalGlobal = function(code,context,filename) {
...
+			fn = vm.runInContext(code,sandbox,filename)
+		} else {
+			fn = vm.runInThisContext(code,filename);
+		}
 	}
 	// Call the function and return the exports
 	return fn.apply(null,contextValues);
 };
-
+$tw.utils.sandbox = !$tw.browser ? vm.createContext({}) : undefined;
```
Jermolene, line 563: It might be safer to check for `$tw.node`; TW 5 also runs under nw.js where both `$tw.browser` and `$tw.node` are `true` (and Node apis are available).

boot/boot.js
```diff
@@ -510,26 +537,46 @@ $tw.utils.evalGlobal = function(code,context,filename) {
...
+
+// if(!$tw.browser) $tw.utils.evalSandboxed(`
+//   setInterval(function(){
+// 		console.log(globalThis)
+// 		if(Object.keys(globalThis).length){
+// 			console.log(globalThis);
+// 			throw "Global assignment is not allowed within modules on node.";
+// 		}
+// 	}, 1000);
+// `, { setInterval, exports: {}, console }, "$:/boot/boot.js-global-monitor");
```
Jermolene, line 579: Can this be removed?

boot/boot.js
```diff
@@ -510,26 +537,46 @@ $tw.utils.evalGlobal = function(code,context,filename) {
...
+$tw.utils.sandbox = !$tw.browser ? vm.createContext({}) : undefined; 
 /*
 Run code in a sandbox with only the specified context variables in scope
 */
-$tw.utils.evalSandboxed = $tw.browser ? $tw.utils.evalGlobal : function(code,context,filename) {
-	var sandbox = $tw.utils.extend(Object.create(null),context);
-	vm.runInNewContext(code,sandbox,filename);
-	return sandbox.exports;
-};
+$tw.utils.evalSandboxed = $tw.browser ? $tw.utils.evalGlobal : function(code,context,filename,allowGlobals) {
```
Jermolene, line 567: The handler function here has the arguments `code,context,filename,allowGlobals` but the definition of `$tw.utils.evalGlobal` above has `code,context,filename,sandbox,allowGlobals`.

This perhaps suggests that some more tests would be useful?

Jermolene (commented):
Thanks @Arlen22, some minor comments.

boot/boot.js
Arlen22: Yes, comments can be removed.

boot/boot.js
Arlen22: Yes, this part needs to be exactly as written. It's probably needed for backward compatibility with IE or something.

boot/boot.js
Arlen22: Yes, does it not matter if the global gets assignments?

boot/boot.js
Arlen22: That's based on evalSandboxed, which uses `$tw.browser`, so this needs to be the opposite of that.

boot/boot.js
Arlen22: evalSandboxed calls evalGlobal with the specified sandbox, and evalGlobal uses the sandbox if specified. I don't really know how or where to write tests. Check the code of evalSandboxed to see how the sandbox argument is used.

Arlen22:
I cleaned up the comments and did a bit of formatting for readability. I think the polyfill definitely needs to stay. I found it in an article which basically found a way to cover every scenario including Workers. This covers most browsers and Node back to 0.10. Otherwise there is no way to dependably access the global in NodeJS. Thus the use of __defineGetter__. 

As far as whether we should have the global check or not, I don't really know. I guess that's up to you. I think we should have it there because that keeps the original behavior as far as security is concerned. In other words, it will throw an error if a module tries to assign anything to the global, which I think is good and is in line with the design goals of TW5.

pmario:
@Arlen22 ... Is this PR still a thing. ... There is a conflict now, with css-escape-polyfill.js ... Which was a concern for backwards compatibility in the discussion above.

Arlen22:
Certainly it still has not been fixed. The evalSandboxed is not intended to be a security feature in NodeJS, so the only use for it is to prevent global assignments or cross-polution with other pieces of server software in the same Node process, so using one context for all modules is still a reasonable choice. I'll check the conflict with css-escape-polyfill when I'm at my computer.

pmario:
> ... The evalSandboxed is not intended to be a security feature in NodeJS, ...

I did read that. ... But the "startup time" improvement should have a value too. Especially if you want to use TW with "lambda functions", where a fast start could make a difference. right?

Jermolene:
Thanks @Arlen22 @pmario 

Apologies, I'd like to merge this but sadly it's too late in the current release cycle, but let's merge it for the following release.

pmario:
@Arlen22 .. I stumbled upon this PR, while searching for some info. .. Is this PR still a thing? Jeremy mentioned, that he would like to merge it -- but there are conflicts now

Arlen22:
The css-escape-polyfill.js change is just to fix the fact that it's factory function is being called on eval instead of on the startup call, which I believe appeared to be a bug. That was the only change. I'll check it tomorrow and try to resolve it.

Arlen22:
Yes, this is still a thing.

boot/boot.js
Arlen22: Yes, the idea here is to throw an error if anything tries to assign something to global so that we don't just silently ignore it and then make everyone wonder why it's not working.

boot/boot.js
Arlen22: AllowGlobals isn't checked unless sandbox is specified, and object assign should throw an error if sandbox is boolean, so I think that covers the possible variations introduced. Back then I thought you meant unit tests, but now I'm not sure if you meant making sure variables are correctly typed.

Arlen22:
Apparently css-escape-polyfill doesnt exist anymore. So I guess just remove it from the PR.

Arlen22:
Ok, looks like it's good now.

Jermolene:
Thanks @Arlen22 – I think the only remaining issue is the variable declaration at the top of the file – we don't use directives like `/** @type {import{"path"}} */`.

Jermolene:
Thanks @Arlen22

Issue #4615, opened 2020-04-30
[BUG] [Node] [FileSystemPaths] tiddler file paths are only set at wiki boot


joshuafontany:
**Describe the bug**
If you have a `$:/config/FileSystemPaths` tiddler set, such as _below_ while running on Node.js
```
[is[system]removeprefix[$:/]addprefix[_system/]]
[has[draft.of]addprefix[drafts/]]
[!is[system]!has[draft.of]tag[Glossary]addprefix[glossary/]]
[!is[system]!has[draft.of]addprefix[wiki/]]
```
Then any tiddler which exists when the wiki is booted will get a `fileInfo.filePath`, but then if the tiddler gains or loses the `Glossary` tag, it will not move to the folder as defined on that line of `$:/config/FileSystemPaths`.

If the Title of the tiddler is changed, the old tiddler is correctly deleted and a new tiddler is created that matches and then is filed in the correct folder (either moving into or out of the `glossary/` directory).

**Expected behavior**
The *.tid file should be deleted and re-saved into the correct path if it matches a new filter after it has been changed.

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10 Node.js
 - Browser [e.g. chrome, safari] Chrome
 - Version [e.g. 22] TW 5.1.22 & TW 5.1.23-pre

**Additional context**

Jermolene:
Hi @joshuafontany that sounds reasonable, I'd welcome a PR if you have any ideas for the fix?

joshuafontany:
Yeah, I'm working on a solution, but debugging the callback functions via VSCode' Node.js options is a headache atm... (breakpoints are not getting hit). Might have something to hare this weekend.

Jermolene:
Thanks @joshuafontany

joshuafontany:
PR here: https://github.com/Jermolene/TiddlyWiki5/pull/4630

joshuafontany:
@Jermolene I believe this can be closed with the 5.1.23 fix. Also cleaning up the callbacks in my latest PR.

pmario:
@joshuafontany ... You did open the issue, so you are able to close it, if you want.

PR #4630, opened 2020-05-03
Fix file info

joshuafontany:
Fixes https://github.com/Jermolene/TiddlyWiki5/issues/4615

I tried to be very minimal with the code changes and keep it to the plugin, but I did have to change a $:/core tiddler.

$:/core/modules/utils/filesystem.js now accepts a `fileSystemPath` value in the incoming `options` object. If set to a string, it will test all final `fullPath` values (in the count loop) against this and break out of the loop immediately if they match. This is because we are now calling this method on ALL tiddler changes that are saved to the server, not just on tiddler "creation".

(I had an infinite $:/StoryList_XXX.tid series until I figured that out, lol.)

I decided to delete the old file _after_ writing the new file (in case of crashes, better duplicate tiddlers than data-loss). I think I did that correctly (chaining it into the callbacks). I have run as many tests as I can think of and this seems to be very stable (finally :P ).


plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -60,7 +66,23 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
 		});
 		$tw.boot.files[title] = fileInfo;
 	}
-	callback(null,fileInfo);
+	else if(fileInfo && fileSystemPaths) {
+		// Otherwise (check for `FileSystemPaths`), we'll need to (re)generate it
+		options.fileInfo = {...fileInfo};
```
pmario, line 71: TW needs to be ES5 only so no Rest functionality.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -43,14 +43,20 @@ Return a fileInfo object for a tiddler, creating it if necessary:
...
+
+If `$:/config/FileSystemPaths` exists, we need to test for a new path and delete the old file after saving.
 */
 FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
 	// See if we've already got information about this file
 	var title = tiddler.fields.title,
-		fileInfo = $tw.boot.files[title];
+		fileInfo = $tw.boot.files[title],
+		fileSystemPaths = this.wiki.tiddlerExists("$:/config/FileSystemPaths"),
+		options = {};
```
pmario, line 59: options is initialized as an empty Object.  In line 79 it is set to null. ... hmmm

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -60,7 +66,23 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
...
+		// Otherwise (check for `FileSystemPaths`), we'll need to (re)generate it
+		options.fileInfo = {...fileInfo};
+		fileInfo = $tw.utils.generateTiddlerFileInfo(tiddler,{
+			directory: $tw.boot.wikiTiddlersPath,
+			pathFilters: this.wiki.getTiddlerText("$:/config/FileSystemPaths","").split("\n"),
+			wiki: this.wiki,
+			fileSystemPath: options.fileInfo.filepath
+		});
+		if(options.fileInfo && options.fileInfo.filepath == fileInfo.filepath) {
+			options = null;
```
pmario, line 79: should it be `null` or `{}`

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -69,11 +91,20 @@ Save a tiddler and invoke the callback with (err,adaptorInfo,revision)
 */
 FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback) {
 	var self = this;
-	this.getTiddlerFileInfo(tiddler,function(err,fileInfo) {
+	this.getTiddlerFileInfo(tiddler,function(err,fileInfo,options) {
 		if(err) {
 			return callback(err);
 		}
-		$tw.utils.saveTiddlerToFile(tiddler,fileInfo,callback);
+		if (options && options.fileInfo !== null) {
```
pmario, line 98: here is a check for `options.fileInfo !== null` .... but fileInfo may not be `null`  ... IMO `options` was set to `null` ... Please check this. I'm not 100% sure, it just looks a little bit suspicious

plugins/tiddlywiki/filesystem/filesystemadaptor.js
joshuafontany: Thanks, changed this to explicitly map the properties.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
joshuafontany: Thanks very much for the additional eyes on this!

options was added to store a copy of the fileInfo to test against the regenerated fileInfo. If the `filepath` property matches, the file is going to be written to the same disk location, so options is set to null.

If the path has changed, the new fileInfo is stored and the existence of options triggers the additional `self.deleteTiddler` callback (your comment below) using the options object to tell it where to delete the old file-copy from.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
joshuafontany: Mahalo. Debugging this was a pain for a little while until I figured out how to step into virtual file in VSCode+Node.js. Please see my above comments, and the newest refactor. Really appreciate the help. :)

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -60,7 +66,27 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
...
+		};
+		fileInfo = $tw.utils.generateTiddlerFileInfo(tiddler,{
+			directory: $tw.boot.wikiTiddlersPath,
+			pathFilters: this.wiki.getTiddlerText("$:/config/FileSystemPaths","").split("\n"),
+			wiki: this.wiki,
+			fileSystemPath: options.fileInfo.filepath
+		});
+		if(options.fileInfo && options.fileInfo.filepath == fileInfo.filepath) {
+			options = null; //if filepaths match, options not needed
+		}
```
Jermolene, line 84: And again here

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -60,7 +66,27 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
 		});
 		$tw.boot.files[title] = fileInfo;
 	}
-	callback(null,fileInfo);
+	else if(fileInfo && fileSystemPaths) {
```
Jermolene, line 69: Minor formatting wonkiness: we'd usually write `} else {` on one line

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -69,11 +95,21 @@ Save a tiddler and invoke the callback with (err,adaptorInfo,revision)
 */
 FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback) {
 	var self = this;
-	this.getTiddlerFileInfo(tiddler,function(err,fileInfo) {
+	this.getTiddlerFileInfo(tiddler,function(err,fileInfo,options) {
 		if(err) {
```
linonetwo, line 99: maybe need whitespace after `if`?

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -60,7 +66,27 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
...
+			type: fileInfo.type,
+			hasMetaFile: fileInfo.hasMetaFile
+		};
+		fileInfo = $tw.utils.generateTiddlerFileInfo(tiddler,{
+			directory: $tw.boot.wikiTiddlersPath,
+			pathFilters: this.wiki.getTiddlerText("$:/config/FileSystemPaths","").split("\n"),
+			wiki: this.wiki,
+			fileSystemPath: options.fileInfo.filepath
+		});
+		if(options.fileInfo && options.fileInfo.filepath == fileInfo.filepath) {
```
linonetwo, line 82: I think modern JS use === instead of ==

saqimtiaz:
This change isn't entirely backwards compatible and in my particular case it is a breaking change.

I create my notes as tiddlers with a type `text/plain` and a tag `note` to ensure that they are saved as `.txt` files with an external `.meta` file, as this is an essential part of my workflow. A filter in `$:/config/FileSystemPaths` ensures that they are saved to a specific directory. 

I am then able to change the type to `text/vnd.tiddlywiki` if need to be able to use wikimarkup, and the tiddler is still saved to the same `.txt` file with an external `.meta` file. With this change, changing the type of the tiddler saves it to a `.tid` file, breaking my workflow entirely. The problem in my case isn't that the file might get saved to a different directory, rather than it can trigger file name/meta file changes as well.

I realize mine is an edge case, but it is good to be aware when a change breaks backwards compatibility no matter how unlikely, and make the decision knowingly. I also welcome alternative suggestions for how I could make sure my tiddlers get saved with an external .meta file (even with type text/vnd.tiddlywiki), so that we can move forwards with this pull request without destroying my worflow.

linonetwo:
Hi, @saqimtiaz , maybe we need a config option to toggle this? 

Because I need this feature, sometimes I create a file with js content, but forget to change the type to `application/text-javascript`, later while I change it to `application/text-javascript`, it still remains `.tid`, through I need a `.js` file with `.meta`.

I don't know whether there is a better solution for this except for a config option.

saqimtiaz:
@linonetwo I understand, and I do think this change is overall in an a positive direction. However we are very careful with regards to backwards compatibility with TiddlyWiki, so any change that breaks backwards compatibility needs to be carefully considered.

Entirely personally, I feel that if this change was accompanied at the same time by also adding the ability to control the file extensions and meta file behaviour for different types of tiddlers, that would be a good solution. It would let me maintain my workflow and let us improve the handling of file paths.  However, it is entirely possible there might be other users whose workflow relies on the same quirk who may feel otherwise.

Perhaps a way to configure or tweak the default file extensions and meta file behaviour for the different tiddler types? Or could we extend $:/config/FileSystemPaths to also allow specifying file extensions? I am open to ideas.

Thoughts @Jermolene ?

joshuafontany:
> @linonetwo I understand, and I do think this change is overall in an a positive direction. However we are very careful with regards to backwards compatibility with TiddlyWiki, so any change that breaks backwards compatibility needs to be carefully considered.

Hi @linonetwo, @saqimtiaz !

Thanks for checking through this change. Ouch, we're in "it's not a bug, it' a feature" territory! That does complicate things..... I do think the existing behaviour is non-optimal ("a bug"), but @linonetwo 's point about backwards compatibility is also very very true. Hmmm...
 
> Entirely personally, I feel that if this change was accompanied at the same time by also adding the ability to control the file extensions and meta file behaviour for different types of tiddlers, that would be a good solution. It would let me maintain my workflow and let us improve the handling of file paths. However, it is entirely possible there might be other users whose workflow relies on the same quirk who may feel otherwise.
> 
> Perhaps a way to configure or tweak the default file extensions and meta file behaviour for the different tiddler types? Or could we extend $:/config/FileSystemPaths to also allow specifying file extensions? I am open to ideas.
> 
> Thoughts @Jermolene ?

These are options worth thinking through. Any other suggestions are also welcome. I'll give this some thought and check back later.

joshuafontany:
> @linonetwo I understand, and I do think this change is overall in an a positive direction. However we are very careful with regards to backwards compatibility with TiddlyWiki, so any change that breaks backwards compatibility needs to be carefully considered.
> 
> Entirely personally, I feel that if this change was accompanied at the same time by also adding the ability to control the file extensions and meta file behaviour for different types of tiddlers, that would be a good solution. It would let me maintain my workflow and let us improve the handling of file paths. However, it is entirely possible there might be other users whose workflow relies on the same quirk who may feel otherwise.

From writing my JSON/CSV plugins, I know that you can register your own fFileTypes, like so

`$tw.utils.registerFileType("application/csv","utf8",".csv");`

Would it be worth spinning something up with that that would allow a plugin to over-ride the default extentions/MIME types?

> Perhaps a way to configure or tweak the default file extensions and meta file behaviour for the different tiddler types? Or could we extend $:/config/FileSystemPaths to also allow specifying file extensions? I am open to ideas.
> 
> Thoughts @Jermolene ?

I am also curious if @Jermolene has any insight  here. Thanks!

pmario:
@saqimtiaz 

>I create my notes as tiddlers with a type text/plain and a tag note to ensure that they are saved as .txt files with an external .meta file, as this is an essential part of my workflow. A filter in $:/config/FileSystemPaths ensures that they are saved to a specific directory.

That's funny. I prefer the other way around. .. Sometimes I do create `text/plain` and really hate it, that it is split into 2 files, which can't be saved in an atomic way. I do prefer the `.tid` format as a default. ... 

I would like to have a configuration to tell TW which type I want to save it, that is completely independent of the "default" setting. .. It's OK, that default splits files. But I want the possibility to overwrite and  remember it. 

It would also be nice to see, how a tiddler will be saved, in the UI

saqimtiaz:
@pmario user preferences and needs will always vary, so flexibility where possible, is key. :)

I think we are in agreemnet with @joshuafontany that a way to over-ride the default file types would be a good way to go. I am not suggesting we change the defaults, just have a way to override them.

An additional part of the issue is that the meta file behaviour for text/vnd.tiddlywiki files is hardcoded:
https://github.com/Jermolene/TiddlyWiki5/blob/fd9938f2ccfd37765c0f17419f398e429aa5e004/core/modules/utils/filesystem.js#L212

Perhaps the fileTypes could be extended to also encompass whether or not to generate a metaFile, with a backwards compatible default?

There is of course the entirely separate question of whether solving my use-case is enough justification to break backwards compatibility, for what is probably extremely unlikely to impact other users.

joshuafontany:
Yeah, there is some interesting fallback logic here and in `boot.js` that registers both `text/vnd-tiddlywiki` and `application/x-tiddler` types to a `*.tid` filetype, and this also touches on a lot of the hard-coded parser logic in `boot.js`....

https://github.com/Jermolene/TiddlyWiki5/blob/fd9938f2ccfd37765c0f17419f398e429aa5e004/core/modules/utils/filesystem.js#L210-L220

Curiouser and curiouser!, quoth Alice.

@saqimtiaz, I want to clarify your use case. You want **all** tiddlers that contain parable wikitext to be `type: text/vnd.tiddlywiki` but to save as `*.txt` and `*.txt.meta`? Would you want this as a global configurable, a per-tiddler (by tag or "_canonical_ext" field, or something), or configurable with `$:/fileSystemPaths`, or a new `$:/fileSystemExtentions` ??

I don't see any problems there with deserializing a tiddler from disk once it has been custom-saved, it will have an accompanying Meta file, and these are always loaded after the raw text/tiddler data is created, and overrides anything from that on import.

To summarize, we need a way to force a Meta file and a way to override the extension?

saqimtiaz:
@joshuafontany A global way to configure all `type: text/vnd.tiddlywiki` tiddlers to save as` *.txt` and `*.txt.meta` will suffice for my use case. However, the focus should be on extending the ability to configure this behaviour in a flexible way as others' use cases may vary. 

Even if I had to do this through a one line configuration plugin that registered a fileType, that would be workable for me, but perhaps not for non-developer users.

> To summarize, we need a way to force a Meta file and a way to override the extension?

For my use case, yes. Though I would generalize it to **being able to configure the file extension and meta file generation for any tiddler type.**

joshuafontany:
@saqimtiaz, well I gave it a shot. There is a try/catch block and probably some debugging setuff to clean up, but I ran this through a series of tests and it seems stable (finally, heh).

Here is the config file I tested with:
```
description: Forces a meta file and custom file extention when running on node.js. Same format as $:/config/FileSystemPaths, the first matching line's output becomes the new file extension.
title: $:/config/FileSystemExtensions
type: text/vnd.tiddlywiki

[!is[system]!has[draft.of]type[text/vnd.tiddlywiki]tag[.txt]then[.txt]]
[!is[system]!has[draft.of]type[text/vnd.tiddlywiki]tag[.tid]then[.tid]]
```

Any feedback welcome!

Jermolene:
Hi @joshuafontany @saqimtiaz sorry for the delay. Making things configurable is a good outcome, thank you. Let me know when you're ready for review.

joshuafontany:
Hi guys. I have cleaned up the debugging and if statements, and left the hard-coded skip for Draft tiddlers. The config tiddler below is a good example of a simple set of filters (remember the Extension includes the dot :) ):

```
description: Forces a meta file and custom file extention when running on node.js. Same format as $:/config/FileSystemPaths, the first matching line's output becomes the new file extension.
title: $:/config/FileSystemExtensions
type: text/vnd.tiddlywiki

[!is[system]type[text/vnd.tiddlywiki]tag[.txt]then[.txt]]
[!is[system]type[text/vnd.tiddlywiki]tag[.tid]then[.tid]]
```

joshuafontany:
Hmm, I've been re-reading my code here, and I think there is a deign bug. With this code, it is possible to override the `hasMetaFile` property to true while retaining the `*.tid` extension. I attempted to code around this bug by trapping the options object and testing for the same file path but different `hasMetaFile` values.

That seemed to work fine after debugging, and was stable through a bunch of tests. Yet, on thinking  on it now- it seem to me the better way to go would be to force `hasMetaFile` to FASLE if the extension === ".tid". Why? Well, 1) we don't break the *.tid file format definition which requires a series of fields before the text field, and 2) as I'm coding up a syntax highlighting extention for VSCode, then point 1 throws a wrench into a whole bunch of gears, 3) the alternate would mean you can force files stored on the server as other filetypes to be _saved_ as text all-in-one *.tid files. I'm not 100% sure there are good use-cases for Three, but it also seems more "backwards compatible".

I'm going to update this PR in that diection.

joshuafontany:
That simplifies things in both files. Cool. 

@Jermolene @saqimtiaz @pmario , anything else you can think of?

I tested a bunch of image and text tiddlers, changed via the Draft mechanism or via Button Actions to add/remove tags in various combinations with the `type` field set to different things, with the following `$:/config/SystemFileExtensions` tiddler:
```
[!is[system]tag[.tid]then[.tid]] [!is[system]type[text/vnd.tiddlywiki]tag[.txt]then[.txt]]
```

Thanks for all the help with this!

saqimtiaz:
@joshuafontany this looks good and well thought out, and should cover my workflow! 

The only remaining use case I can envision that we haven't covered, is if someone wanted to save `text/vnd.tiddlywiki` tiddlers as `.txt` files but without a `meta` file.

Should that be something we allow for?

joshuafontany:
@saqimatiaz, the TW core defines *.txt files as content:

https://github.com/Jermolene/TiddlyWiki5/blob/41a7d98e17470d23164420c0235b10f717a7a009/boot/boot.js#L2252

https://github.com/Jermolene/TiddlyWiki5/blob/41a7d98e17470d23164420c0235b10f717a7a009/boot/boot.js#L1557-L1563

I think if you wanted to treat ALL *.txt files in your wiki's tiddlers folder as tiddlers (i.e. check them for field-value pairs and then a blank line, and then text-content), you might be able to do so in a plugin that redefines a few key modules, for example (completely off the top of my head):

```
$tw.modules.define("$:/boot/tiddlerdeserializer/txt","tiddlerdeserializer",{
	"text/plain": function(text,fields) {
		var split = text.split(/\r?\n\r?\n/mg);
		if(split.length >= 1) {
			fields = $tw.utils.parseFields(split[0],fields);
		}
		if(split.length >= 2) {
			fields.text = split.slice(1).join("\n\n");
		}
		return [fields];
	}
});
```

But I think that's going to clash with the deserializer definitions where

https://github.com/Jermolene/TiddlyWiki5/blob/41a7d98e17470d23164420c0235b10f717a7a009/boot/boot.js#L1483-L1486


So I think that the best options are to force the *.txt.meta file, or loading folders with the tiddlywiki.files method (which allows you to attach fields to body-content loaded from arbitrary files).

---

This all seems stable so far while running vanilla Node.js TW, and also while running Bob while the PR changes are loaded. 

(Although I am having a hell of a time re-creating the Save-flow in Bob's syncadaptor so that the fileInfo object gets regenerated, the tiddler saved & any old files deleted... @inmysocks I may be bugging you about getting that working, as this is proving incredibly useful.)

saqimtiaz:
@joshuafontany good point about the  `tiddlywiki.files` option.

saqimtiaz:
@joshuafontany #4702 presents a good argument for allowing Markdown files with the tiddler metadata inline as opposed to in a .meta file.

joshuafontany:
> @joshuafontany #4702 presents a good argument for allowing Markdown files with the tiddler metadata inline as opposed to in a .meta file.

Interesting, I hadn't heard of frontmatter before. The frontmatter format is very specific. YAML name: value pairs between a set of triple dashes. That additional formatting is going to be a challenge...

saqimtiaz:
My mistake. I glossed over the requirement for triple dashes and assumed
the format was the same as .tid

I do not think we need to implement front matter support, especially in
this pull request. However it would be nice if plugins could implement
custom formats for saving.

joshuafontany:
I have been using this for a couple of weeks and it seems very stable. I have had one or two instances of a tiddler appearing in my tiddlers folder as "Draft of 'Tiddler Name'" but _without_ a "draft.of" field (so it was a regular tiddler in disguise). I'm not 100% sure that was unrelated to this branch.... so it could use a bit more testing.

saqimtiaz:
@joshuafontany I can use this for my daily usage wiki as well so we get the benefit of some extra testing.

joshuafontany:
Fixed a race condition bug in Bob by implementing this fix there.
PR: https://github.com/OokTech/TW5-Bob/pull/134

xeor:
Is this far from getting merged? There is (https://github.com/OokTech/TW5-Bob/pull/134) which requires this. It's a fix for a stability-issue in bob, where notes can get lost (not saved), without any notification..

saqimtiaz:
@joshuafontany Could you please have a look at #4761 ? This is another instance where fileInfo needs updating for an existing tiddler, this time because a multiline field has been added. 

It may make sense to address that along with this fix since they deal with the same underlying issue.

saqimtiaz:
@joshuafontany Also, I am wondering if users might find `$:/config/FileSystemExtensions` to be a bit confusing in that it changes the extension but not the format for saving a tiddler.

For example:
`[has[myfield]then[.json]]` in `$:/config/FileSystemExtensions` will result in all tiddlers with the field `myfield` being saved as `.json` files with the `.json` file containing the text field as plain text, with accompanying `.meta` files. A user may somewhat reasonably expect in this situation that the file would be saved as a `.json` file with a JSON object representing all fields of the tiddler

I wonder if `.json` needs some special handling, as being one of two formats for tiddlers (alongside` .tid`) that is self-contained and does not require a separate `.meta` file for each tiddler.

joshuafontany:
> @joshuafontany Could you please have a look at #4761 ? This is another instance where fileInfo needs updating for an existing tiddler, this time because a multiline field has been added.
> 
> It may make sense to address that along with this fix since they deal with the same underlying issue.

I remember noticing this behavior in my tests, and I just re-tested. I almost thought the tiddlers getting written as .json files without seperate .json.meta files was a bug, but no! Good catch. Yup, my PR fixes this bug as well.

joshuafontany:
> @joshuafontany Also, I am wondering if users might find `$:/config/FileSystemExtensions` to be a bit confusing in that it changes the extension but not the format for saving a tiddler.
> 
> For example:
> `[has[myfield]then[.json]]` in `$:/config/FileSystemExtensions` will result in all tiddlers with the field `myfield` being saved as `.json` files with the `.json` file containing the text field as plain text, with accompanying `.meta` files. A user may somewhat reasonably expect in this situation that the file would be saved as a `.json` file with a JSON object representing all fields of the tiddler
> 
> I wonder if `.json` needs some special handling, as being one of two formats for tiddlers (alongside` .tid`) that is self-contained and does not require a separate `.meta` file for each tiddler.

This is where we have clashing definitions and/or a semantic problem, which makes it harder to code up a clean solution. The does the `type` field of a tiddler describe more than just the IMIE type of the `text` field contents? It seems that TW mainly uses `type` in that fashion. When we say "an image tiddler" or a "a JSON tiddler" we mean a tiddler that has that type of "text-content".

Now TW DOES map these IMIE types to distinct filetypes, and JSON is definitely a special case, where you can have a valid JSON file without an accompanying .json.meta file, and it can contain one OR MORE tiddlers.

While I agree about avoiding confusion, I'd have to think about that a bit.....

saqimtiaz:
@joshuafontany I cannot recall the exact steps at the moment but while testing I also encountered some data loss via tiddler corruption when a tiddler matched a filter specifying to save it with .json file extension and had multiline fields added triggering save as a JSON object. I can try to re-create that if that would be helpful.

linonetwo:
@Jermolene This seems stable enough, and this feature is important for storing private content in TW, and merging this PR will enable a fix in the Bob https://github.com/OokTech/TW5-Bob/issues/130

Can you please review this and consider merge it?

saqimtiaz:
@linonetwo the handling of the .json extension still needs work or at least concensus on how it should be handled, and #4761 should probably be addressed at the same time. Therefore I do not think this is ready to be merged.

Jermolene:
Thanks everyone, always surprising when apparently simple things turn out to have this kind of twisty-turniness.

I just wanted to point out that #4756 may give us an alternative approach to clearing up the inconsistencies in the way that the boot kernel and filesystemadaptor handle syncing. In that context, I'm interested in supporting alternative formats for wiki storage: the client wikis won't be touched by the boot process and so we can handle loading and syncing them much more consistently. I'd like to adopt a disk format that gives us etags that survive across restarts of the server.

joshuafontany:
@Jermolene  This patch (& the parallel patch I made for Bob) are working fine for my needs, atm. Although not being in 'main' is blocking @linonetwo from experimenting with git called from TW. I would be interested in helping whatever solution we go with, especially multi-wiki support in the core. I'll take a look at the referenced thread.

linonetwo:
I try this PR by releasing an npm package `@tiddlygit/tiddlywiki@5.1.23-prerelease.20200815-fix-file-info`, then I use it in the TiddlyGit-Desktop.

I have following `$:/config/FileSystemPaths`, so it will move private tiddler into corresponding sub-folders, which are symlinks to a private Github repo:

```tid
tags: $:/plugins/linonetwo/private-tiddlers-in-private-repo/readme
title: $:/config/FileSystemPaths
type: text/vnd.tiddlywiki

[kin::to[APrivateContent]addprefix[/]addprefix[private-wiki]]
[kin::to[ByteDance]addprefix[/]addprefix[private-wiki-bytedance]]
```

Seems it is not working? Here is a screenshot git:

ezgif com-video-to-gif

I add "APrivateContent" tag to a new tiddler, it truely go to the `tiddlers/private-wiki` folder, but when I remove that tag and add "ByteDance" tag, it doesn't go to the `tiddlers/private-wiki-bytedance` folder. Which it should.

And also, remove tags wont make tiddler go back to the root `tiddlers` folder

pmario:
>(kin-filter will move all tiddlers in the tag-sub-tree into the private folder too, very convenient)

kin-filter isn't part of the core. So an example, that uses core only features would be nice!

pmario:
A node TW keeps track of original tiddler paths in: http://localhost:8080/#%24%3A%2Fconfig%2FOriginalTiddlerPaths So this may be the cause, that tiddlers change the path if you change the tag.

pmario:
also see: http://localhost:8080/#How%20to%20add%20a%20banner%20for%20GitHub%20contributions

pmario:
and: http://localhost:8080/#tiddlywiki.info%20Files

linonetwo:
When 

```
tags: $:/plugins/linonetwo/private-tiddlers-in-private-repo/readme
title: $:/config/FileSystemPaths
type: text/vnd.tiddlywiki

[!is[system]kin::to[APrivateContent]addprefix[subwiki/private-wiki/]]
[!is[system]kin::to[ByteDance]addprefix[subwiki/private-wiki-bytedance/]]
```

what should happen if you add two tags `APrivateContent` and `ByteDance` to a tiddler? @joshuafontany

joshuafontany:
Per the TW Documentation, only the first matching filter from `$:/config/FileSystemPaths` is used to generate the final file-path.

Test for a combination of both tags before testing for each individually (on separate lines/filter-runs).

Best,
Joshua F

joshuafontany:
@saqimtiaz Hey, I was reviewing your excellent work making `tiddlywiki.files` external files savable back to their original location.

I would really like to get some version of this fix merged into the core, is your requirement that you are able to specify an alternate filetype still needed? If not, I can simplify this back to just checking if the new file-path differs from the existing filepath.

Would I need to take any of your `tiddlywiki.files` work into account there? I wouldn't want to change the file-path for externally loaded tiddlers, are they flagged somehow in `loadTiddlersFromSpecification ()`?

Thanks for your help,
Joshua F

saqimtiaz:
@joshuafontany the extension to `tiddlywiki.files` does not obviate the need for specifying an alternate filetype. If anything, that workflow and the need to specify extensions is likely to become more common now.

However, I do not think you need to take into account those changes in any form for this PR. The extension to `tiddlywiki.files` allows the possibility of saving files loaded in that manner back to their original location. The user should still be able to override that path with `$:/config/FileSystemPaths`

I reviewed this PR again last week and in my opinion there are two issues we need to address:
1. Instead of making the mechanism for changing the file path specific to `$:/config/FileSystemPaths` it should be general enough to encompass other situations where the file path or type needs to change. An immediate example is the issue in #4761 where tiddlers that are updated to contain multi-line fields need to be saved as `.json`
2. Currently if the user configures a .json extension in `$:/config/FileSystemExtensions` what is generated is not a valid `.json` file, as it is a plain text file with `.json` extension. I realize hardcoding special behaviour is never elegant but I can see users wanting to save as `.json` as a realistic use case.

joshuafontany:
@saqimtiaz Thank you, that helped me get back up to speed on this one. I agree with your note in https://github.com/Jermolene/TiddlyWiki5/issues/4761 that we may just need to regenerate the fileInfo on each tiddler save to cover all the possible ways that a path or extension may change.

I was primarily concerned with externally loaded tiddlers because my code currently wants to delete the file at the old location once the new file is saved for a tiddler that moves or changes extension. I believe this would be a bug for "externally loaded" tiddlers.

I'll also think on how to make this a more generally applicable test. Do you agree with keeping `$:/config/FileSystemExtensions` as a valid way of setting file extensions for tiddlers in-bulk? Would there be a use for a `_canonical_ext` field on individual tiddlers (which would override any other setting)?

Thanks for your input!

Best,
Joshua Fontany

saqimtiaz:
@joshuafontany So tiddlers loaded via tiddlywiki.files that do not have the `isEditableFile` flag set, wont have their filepath set. So your code should not be able to delete the original files, it just needs to not throw an error in that situation. 

For files that do have `isEditableFile` flag set, I would expect them to be deleted/moved etc just like regular tiddlers. Otherwise we would end up with duplication of files, which is precisely the situation that the flag was introduced to avoid.

In my particular use case I need the file extension not to change when the tiddler type changes, which I would enforce with `$:/config/FileSystemExtensions`

So I don't think you need to do any more handling in this regard. Either this weekend or early next week at the latest, I can do some testing to make sure the behaviour is as I have described above and there isn't something that we are missing. If you can merge the upstream/master into your branch, that will make testing easier.

> Do you agree with keeping `$:/config/FileSystemExtensions` as a valid way of setting file extensions for tiddlers in-bulk?

Yes

> Would there be a use for a `_canonical_ext` field on individual tiddlers (which would override any other setting)?

Not that I can think of for now, and I would recommend taking the conservative approach in situations where we aren't sure of the usage pattern.

joshuafontany:
@saqimtiaz @Jermolene @pmario @linonetwo 

Recent changes to boot.js have forced me to abandon this fork. I have started on a new PR for updating `filesystemadaptor.js` and related methods with what we have learned here. I really need to run a few things past you guys, especially for "user expectations" that I might not be seeing.

So, we have the case of tiddlers loaded from a directory with a `tiddlywiki.files` config with the new "isEditableFile" set to true, and have to keep in mind the desire to generate the `fileInfo` on each save task (in order to stay flexible with the file path and extention filters). I think we have overloaded the `tiddlywiki.files` functionality a bit. I have given it my best shot to match all of the requirements we brought up here. I'll post my PR shortly for testing and update this comment.

saqimtiaz:
@joshuafontany let me know when you need feedback or testing. 

I still think that we do not need any special handling for tiddlers loaded via `tiddlywiki.files` with `isEditableFile` set to `true`.

Without this attribute there is no `filePath` set, so your code just needs to not throw an error when it cannot delete the file at the original path since there will be no original path.

For tiddlers loaded with `isEditableFile` set to true, treat them like any other tiddler. The only exception that I can think of is that a `.meta` file may not exist and trying to delete it should not throw an error.

joshuafontany:
Hey @saqimtiaz. Thanks for that. The problem I am currently having is that "regenerate file Info every time" clashes with "The extension to tiddlywiki.files allows the possibility of saving files loaded in that manner back to their original location. The user should still be able to override that path with $:/config/FileSystemPaths"

Regenerating file Info without FileSystemPaths gives the "normal" title-based (in the flat "/tiddlers" folder) which never matches the fileInfo from loading these. You would need a specific FileSystemPaths filter to put them "back" into their "home location".

While this is configurable- it doesn't exactly match how we have been talking about these side-loaded files "just working" when marked isEditableFile.

saqimtiaz:
@joshuafontany interesting! I hadn't considered that the "regenerate fileInfo every time" would be the part causing problems.

So the real reason to consider that is #4761 

So two thoughts:
a) we could do a check for unsafe fields first. 
b) we could try to recouple regenerating the file type from the file path. Solving #4761 and similar issues only requires a change in file type.

saqimtiaz:
@joshuafontany another thought is that in `generateTiddlerFileInfo`, if there are no `unsafeFields` and no extensions or path being overridden, preserve the original file path in `fileInfo`. I am not at my desk or looking at the code at the moment so I apologize if that is not helpful.

joshuafontany:
@saqimtiaz @linonetwo @pmario @Jermolene 

Closed in favor of https://github.com/Jermolene/TiddlyWiki5/pull/5113

Thanks for any attention you can lend. :)

Best,
Joshua F

PR #4634, opened 2020-05-05
Fix 4596 - High cpu consume after upgrade to 5.1.22

### LINKED ISSUE #4596: [BUG] High cpu consume after upgrade to 5.1.22
**Describe the bug**

After upgrade to 5.1.22, when open tiddlywiki in google chrome, it consume around 30% cpu, even  if there is no tiddler is opened

If move the google chrome window into background, the cpu consume is back to normal.

I also tried to fallback to 5.1.21, which do not have the issue.

Firefox has the same behavior with google chrome.

**Expected behavior**

**Screenshots**

image

 - OS: Archlinux
 - Browser: google-chrome 81.0.4044.122-1
 - Version: 

**Additional context**

```json
{
    "plugins": [
        "tiddlywiki/codemirror",
        "tiddlywiki/codemirror-closebrackets",
        "tiddlywiki/codemirror-closetag",
        "tiddlywiki/filesystem",
        "tiddlywiki/highlight",
        "tiddlywiki/pluginlibrary",
        "tiddlywiki/tiddlyweb"
    ],
    "themes": [
        "tiddlywiki/vanilla",
        "tiddlywiki/snowwhite"
    ],
    "build": {
        "renderfull": [
            "--rendertiddler", "$:/core/save/all", "index_full.html", "text/plain"
        ],
        "render": [
            "--rendertiddler", "$:/core/save/lazy-images", "index_without_img.html", "text/plain"
        ],
        "index": [
            "--rendertiddler",
            "$:/core/save/all",
            "index.html",
            "text/plain"
        ],
        "empty": [
            "--rendertiddler",
            "$:/core/save/all",
            "empty.html",
            "text/plain",
            "--rendertiddler",
            "$:/core/save/all",
            "empty.hta",
            "text/plain"
        ],
        "externalimages": [
            "--savetiddlers",
            "[is[image]]",
            "images",
            "--setfield",
            "[is[image]]",
            "_canonical_uri",
            "$:/core/templates/canonical-uri-external-image",
            "text/plain",
            "--setfield",
            "[is[image]]",
            "text",
            "",
            "text/plain",
            "--rendertiddler",
            "$:/core/save/all",
            "externalimages.html",
            "text/plain"
        ],
        "static": [
            "--rendertiddler",
            "$:/core/templates/static.template.html",
            "static.html",
            "text/plain",
            "--rendertiddler",
            "$:/core/templates/alltiddlers.template.html",
            "alltiddlers.html",
            "text/plain",
            "--rendertiddlers",
            "[!is[system]]",
            "$:/core/templates/static.tiddler.html",
            "static",
            "text/plain",
            "--rendertiddler",
            "$:/core/templates/static.template.css",
            "static/static.css",
            "text/plain"
        ]
    },
    "config": {
        "retain-original-tiddler-path": true
    }
}
```

pmario:
This PR fixes #4596 .. There was a problem with Chrome and Chromium. The "syncing in progress" element was set to opacity: 0 but was animated all the time, which causes CPU load to go up significantly. 

Now the element is set to `display: none` _and_ `opacity:0` so the fade in and fade-out animations still work.


Jermolene:
Thanks @pmario just curious why you used "unset" instead of "inline" when restoring visibility?

pmario:
>Thanks @pmario just curious why you used "unset" instead of "inline" when restoring visibility?

No special reason. 

I didn't know, what the initial value was. I did try "unset" and it worked. 

It will automatically inherit the setting from outer elements. ... If "inline" is hardcoded it will always use inline instead of inheriting the value. So may be "inherit" would have worked too.

PR #4640, opened 2020-05-07
Visual changes to Saving Tiddler

ibnishak:
- Adding a card template in line with Tiddler design
- Allows users to select multiple options at a time, creating combinations
- Authors names displayed in the cards
- Read more links to corresponding tiddlers.

### DIFF
--- a/editions/tw5.com/tiddlers/$__StoryList.tid
+++ b/editions/tw5.com/tiddlers/$__StoryList.tid
@@ -1,5 +1,5 @@
 created: 20191004112211823
-list: [[WidgetMessage: tm-unload-plugin-library]] [[WidgetMessage: tm-load-plugin-library]] [[WidgetMessage: tm-load-plugin-from-library]] HelloThere GettingStarted Community
+list: HelloThere GettingStarted Community
 modified: 20191004113621710
 title: $:/StoryList
 type: text/vnd.tiddlywiki
--- a/editions/tw5.com/tiddlers/Twexe_ Single File Tiddlywiki5 executable.tid
+++ b/editions/tw5.com/tiddlers/Twexe_ Single File Tiddlywiki5 executable.tid
@@ -1,0 +1,29 @@
+caption: Twexe
+color: #ff8a65
+community-author: ihm4u
+created: 20200507214737998
+delivery: App
+description: Single File Tiddlywiki5 executable
+method: save
+modified: 20200507215529213
+tags: Firefox [[Internet Explorer]] Opera Resources Safari Saving Windows Edge Linux
+title: Twexe: Single File Tiddlywiki5 executable
+type: text/vnd.tiddlywiki
+url: https://ihm4u.github.io/twexe/
+
+{{!!description}}
+{{!!url}}
+
+!! Features
+
+* Automatic backups
+* Automatic saving
+* No browser add-ons needed!!
+* Works on Linux and Windows
+* Single file executable, can be moved, copied, etc.
+* Wiki is compressed, occupying less space on your disk (sometimes even a third of the space!)
+* Converts //any// Tiddlywiki5 file to a single file executable  <<twexe>>, no need for special tiddlywiki plugins
+* Automatically uses the saving tab in the control panel (there you can specify a backup directory if you don't like the default)
+* Easily get back the html wiki file from the executable twexe (if you  need it for some reason)
+* Opens up the possibility to run external executables from your wiki (to draw charts, etc.) - stay tuned!!
+* Supports external images (i.e. with _canonical_uri field)
--- a/editions/tw5.com/tiddlers/community/resources/TiddlyDrive Add-on for Google Drive by Joshua Stubbs.tid
+++ b/editions/tw5.com/tiddlers/community/resources/TiddlyDrive Add-on for Google Drive by Joshua Stubbs.tid
@@ -1,9 +1,11 @@
 caption: ~TiddlyDrive
+color: #FFEB3B
+community-author: Joshua Stubbs
 created: 20171109172705241
 delivery: Google Drive Add-on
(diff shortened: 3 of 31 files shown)


ibnishak:
Relevant Screenshots

GSPR1
GSPR2

ibnishak:
If you would like a live demo: https://ibnishak.github.io/Tesseract/demo/docs.html#GettingStarted

Jermolene:
Thank you @ibnishak this is great work, and my apologies for the late response. I love the switch to checkboxes, and the overall look. I have a couple of minor comments:

* I find the large, diffuse drop shadow on the download button distracting. I think we should probably continue to avoid large blur values because they perform poorly on low end devices
* On a typical laptop, only two cards fit across the screen (see below). I think the information density might be a little low, and wonder if we might not either (a) make the cards a bit smaller or (b) put a bit more information into them (e.g. the list of platforms?)
ibnishak github io_Tesseract_demo_docs html(Laptop with HiDPI screen)
* The use of "READ MORE" in all caps might be a bit too Material Design. For us, it seems more natural to present the entire card as a clickable link, without needing any instructional text

The other thought I've been wondering about is whether we could present the same information through a wizard that presented a series of multiple choice questions. Answer (A) for the first question would be something along the lines of "I've no idea what I'm doing", answer (D) would be "I'm a software developer and know exactly what I'm doing", each taking people down an appropriate path. And then there would be a series of questions before arriving at one or more recommendations, presented as cards as per this PR. If something like that worked well, it might make a better "Getting Started" than the picking matrix, which might go into a new tiddler "Saving Methods". Anyhow, that's not a suggestion to modify this PR, just wondering if you might have been thinking along similar lines.

ibnishak:
@Jermolene The changes suggested has been made. The new demo can be found in [doc-demo-tw5.surge.sh](doc-demo-tw5.surge.sh). 

Apart from the obvious ones, some comments about the changes:
1. The font size is reduced by 2px. 
2. The download button box-shadow has been changed, on hover effect removed. on-active, it will change background mildly to give visual feedback of the click.
3. I felt that user might not understand that the card is "clickable". So card title color is changes on hover to give a visual cue.

Please see and comment. I resized the card such that approx 3 cards would be seen in a row in 720p.

ibnishak:
> The other thought I've been wondering about is whether we could present the same information through a wizard that presented a series of multiple choice questions. Answer (A) for the first question would be something along the lines of "I've no idea what I'm doing", answer (D) would be "I'm a software developer and know exactly what I'm doing", each taking people down an appropriate path. And then there would be a series of questions before arriving at one or more recommendations, presented as cards as per this PR. If something like that worked well, it might make a better "Getting Started" than the picking matrix, which might go into a new tiddler "Saving Methods". Anyhow, that's not a suggestion to modify this PR, just wondering if you might have been thinking along similar lines.

Actually I was planning this. However this needs your input. What I had in mind is as follows. 

1. Categorization changes to saver methods. 
I suggest that the savers are broadly classified to 5 groups/tags
  - Browser addons - this will require additional tags of browsers
  - Applications - TiddlyDesktop, Tiddloid, Quine etc
  - Cloud Savers - NoteSelf, TiddlySpot, Git savers, TiddlyDrive, Cloud Connectors
  - Servers and Server protocols - WebDAV, Node JS, TiddlyServer
  - In built - Download saver, Emergency export

This options would be listed under the these subheadings in the matrix. This way all the browser tags in browser independent methods can be removed. User would be able to choose better based on his preferences.

3. The wizard. 
Here is my view regarding giving users options along the lines of advanced/non-advanced user. There is a lot of grey area between those two options. There might be software developers who want a quick-fix. On the other hand there might be beginners who are ready to put in time and effort to build a set up. 

Here is my counter proposal. Provide user with objective choices. The text I had in mind goes along the lines of:

TiddlyWiki comes in 2 major flavors
- Node.js based flat file flavor: All your notes/tiddlers are saved as separate text files. Node.js runtime is required to view your wiki in browser.
- Single HTML file flavor: All your notes/tiddlers are saved in a single HTML file which user can view in any browser. 
  <<.tip "You can convert your wiki between the two flavors any time.">>
  <<.tip "There are third-party community resources providing other back-ends like  [[sqlite, boltdb, dynamodb | widdly]] and [[PouchDB | Noteself]]. Please note that they are published and maintained independently of TW5">>

  Choose your flavor:
   - [[Node.js based flat file flavor|GettingStarted with Node.js]]
   - [[Single HTML file flavor| Link to matrix with user browser checkbox pre-selected]]


---

This way TiddlyWiki distinguishes itself right from the beginning. Hardly any other major note taking platforms provide these many back-ends.

Jermolene:
Thanks @ibnishak 

> Here is my counter proposal

I like it, but I think we should try to cater for those users who would be confused by those options. Perhaps we could add a third option "I don't understand the question" or "Help me to choose which is best for me".

ibnishak:
> Help me to choose which is best for me

In this case what is the next set of questions to be asked?

Jermolene:
> In this case what is the next set of questions to be asked?

The key one perhaps would be to try to establish whether the user needs online or offline access. In the former case I think we'd currently route them towards TiddlySpot, but it would be nice to have docs for options like Glitch and Digital Ocean. The offline branch would presumably then split by operating system, leading them towards to the simplest option for each OS.

ibnishak:
I would start a post in the group regarding this and form a final set of questionnaire so that relatively new users can give their input.

ibnishak:
I just saw that you already did that.

pmario:
For online access, I'd definitely go for GitLab, GitHub and TiddlySpot. .. TS has the problem to be `http://` only, which is a real problem, if you access the wiki from public WiFi's.

linonetwo:
This wizard is very nice, I like this "help me choose".

PR #4643, opened 2020-05-08
The `!has[tags]` filter didn't work because `tags` field is an array

bimlas:
The negated `has` filter only considered empty strings, but not empty arrays (such as the `tags` field).

Steps to reproduce:

* Create a New Tiddler without specifying tags
* Use this filter expression in the AdvancedSearch: `[[New Tiddler]!has[tags]]`
* The output of the filter will be empty, which does not correspond to the behaviour described in the documentation:
> those input tiddlers in which field F does not exist or has an empty value

See also: https://groups.google.com/forum/#!topic/tiddlywiki/2HzNb2V-aIo

### DIFF
--- a/core/modules/filters/has.js
+++ b/core/modules/filters/has.js
@@ -52,13 +52,13 @@ exports.has = function(source,operator,options) {
 	else {
 		if(invert) {
 			source(function(tiddler,title) {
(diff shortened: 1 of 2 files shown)


bimlas:
It seems that this issue is fixed twice, see #4567

Jermolene:
Thanks @bimlas

PR #4647, opened 2020-05-12
Fix the class attribute of Droppable widget

bimlas:
Even though we passed the `class` attribute, it did not take it into
account. For example try to render this:

```
  <$droppable class="custom-css-class">
    ... content ...
  </$droppable>
```

Before applying this patch:

```
  <span class=" tc-droppable">
    ... content ...
  </span>
```

After applying:

```
  <span class="custom-css-class tc-droppable">
    ... content ...
  </span>
```


saqimtiaz:
@bimlas thank you! I ran into this just late last night, so very timely pull request!

Jermolene:
Thanks @bimlas @saqimtiaz

PR #4648, opened 2020-05-12
Move wiki-related $tw.boot variables onto $tw.wiki

Arlen22:
This moves the variables `$tw.boot.files`, `$tw.boot.wikiInfo`, `$tw.boot.wikiPath`, `$tw.boot.wikiTiddlersPath` onto `$tw.wiki`.

- Renamed all occurrences using find and replace all for all Javascript files in the repo. 
- Manually find all occurrences of `$tw.wiki` and replace with the appropriate `this.wiki` or `this.commander.wiki` where applicable. 
- Add a getter and setter on $tw.boot for each variable which gets and sets `$tw.wiki` so current code should not be broken. The getter prints a console message with a stack trace. 
- Test by running `bin/build-site.sh` and `bin/test.sh`. No errors were apparent and no access on `$tw.boot` was logged. 

The main advantage of this is that it keeps file system paths and info with the wiki it pertains to and lets wiki specific code immediately access these variables for the wiki being acted on. 

- NoteSelf has no references to `$tw.boot` in the repo, which makes sense since it is browser-oriented.
- @inmysocks Bob has multiple references to `$tw.boot`, but they will still be able to refer to the global `$tw.wiki` via the getters. Bob may benefit from this change as well by being able to reuse code. 
- TiddlyServer is unaffected.
- I don't know of any other server systems.

The objective I am moving toward is a memory efficient version of Node Server, and one that can handle multiple wikis if desired (through a plugin of course). I did some experiments using the code in TiddlyServer, and the result is very promising. I was able to serve multiple Wiki instances through the same server by setting the pathPrefix and wiki for each request, and attaching a file system adaptor and syncer to each Wiki instance. 

In my testing, because tiddlers are immutable, I was able to save additional memory by sharing one instance of the core tiddler among all data folders. Each core tiddler takes 3.5 MB memory in heap for just the text field string. A modified loader, in this case, would make simple work of this as it could cache this string and use it for the text field of every wiki it loads. This saves another 3.5 MB of memory per wiki.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -803,7 +803,11 @@ $tw.modules.execute = function(moduleName,moduleRoot) {
 				moduleInfo.definition(moduleInfo,moduleInfo.exports,sandbox.require);
 			} else if(typeof moduleInfo.definition === "string") { // String
 				moduleInfo.exports = _exports;
-				$tw.utils.evalSandboxed(moduleInfo.definition,sandbox,tiddler.fields.title);
+				// let isShadow = $tw.wiki.isShadowTiddler(tiddler.fields.title);
+				// let source = isShadow ? $tw.wiki.getShadowSource(tiddler.fields.title) : tiddler.fields.title;
+				let filename = tiddler.fields.title;
+				// console.log(filename);
+				$tw.utils.evalSandboxed(moduleInfo.definition,sandbox,filename&&filename.startsWith("$:/")?filename.slice(3):filename);
 				if(sandbox.module.exports) {
 					moduleInfo.exports = sandbox.module.exports; //more codemirror workaround
 				}
@@ -2076,7 +2080,7 @@ $tw.loadWikiTiddlers = function(wikiPath,options) {
 	$tw.utils.each($tw.loadTiddlersFromPath(resolvedWikiPath),function(tiddlerFile) {
 		if(!options.readOnly && tiddlerFile.filepath) {
 			$tw.utils.each(tiddlerFile.tiddlers,function(tiddler) {
-				$tw.boot.files[tiddler.title] = {
+				$tw.wiki.files[tiddler.title] = {
 					filepath: tiddlerFile.filepath,
 					type: tiddlerFile.type,
 					hasMetaFile: tiddlerFile.hasMetaFile
@@ -2089,8 +2093,8 @@ $tw.loadWikiTiddlers = function(wikiPath,options) {
 	var config = wikiInfo.config || {};
 	if(config["retain-original-tiddler-path"]) {
 		var output = {}, relativePath;
-		for(var title in $tw.boot.files) {
-			relativePath = path.relative(resolvedWikiPath,$tw.boot.files[title].filepath);
+		for(var title in $tw.wiki.files) {
+			relativePath = path.relative(resolvedWikiPath,$tw.wiki.files[title].filepath);
 			output[title] =
 				path.sep === "/" ?
 				relativePath :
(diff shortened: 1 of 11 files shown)


inmysocks:
Bob has multiple wikis on the same server, it just requires a server implementation.
From memory tests with Bob individual wikis are not the cause of the heavy memory usage, each wiki adds less than 20mb of memory even with the core in it. I don't think that any changes to the wiki core is going to help much.

I used to have Bob only use one wiki object and load all the tiddler into that one wiki, the memory usage didn't change much between having one $tw.wiki object for all the wikis and one wiki object for each individual wiki.

Arlen22:
> From memory tests with Bob individual wikis are not the cause of the heavy memory usage, each wiki adds less than 20mb of memory even with the core in it. I don't think that any changes to the wiki core is going to help much.

That's a good point. And 3.5 MB matters less every year. 

> I used to have Bob only use one wiki object and load all the tiddler into that one wiki, the memory usage didn't change much between having one $tw.wiki object for all the wikis and one wiki object for each individual wiki.

Make sense.

Jermolene:
Hi @Arlen22 I appreciate the motivation here but it seems to break backwards compatibility. Experience has shown that we cannot see all the code that depends on the TW5 core, and so we have to make the most conservative assumptions about dependencies.

Arlen22:
> Hi @Arlen22 I appreciate the motivation here but it seems to break backwards compatibility. Experience has shown that we cannot see all the code that depends on the TW5 core, and so we have to make the most conservative assumptions about dependencies.

I'm aware of these possibilities. Would the getter and setter on $tw.boot solve that issue? Should we comment out the console.log line?

Jermolene:
>  Would the getter and setter on $tw.boot solve that issue? Should we comment out the console.log line?

It's a pretty poor experience for everyone, I think. It's also a shame to clutter up the bootprefix.

Arlen22:
These three places are the ones that affect running multiple wikis in core. Changing this allows us to use the server functions unchanged, easing the upgrade process to TiddlyWiki versions. 

The loadTiddlersNode function and its related functions are synchronous so they don't need to be rewritten. At least it's not important right now. 

https://github.com/Jermolene/TiddlyWiki5/blob/6570561d4ec31d9e64c3021bb69c20daec8c9eac/core/modules/server/routes/get-file.js#L24

https://github.com/Jermolene/TiddlyWiki5/blob/6570561d4ec31d9e64c3021bb69c20daec8c9eac/plugins/tiddlywiki/filesystem/filesystemadaptor.js#L24

https://github.com/Jermolene/TiddlyWiki5/blob/6570561d4ec31d9e64c3021bb69c20daec8c9eac/core/modules/server/authenticators/basic.js#L34

Arlen22:
I guess I should just say all the changes in these four files.

- [core/modules/server/authenticators/basic.js](https://github.com/Jermolene/TiddlyWiki5/pull/4648/files#diff-bb66fb42d7621c7676d9982c1d6d4608)
- [core/modules/server/routes/get-file.js](https://github.com/Jermolene/TiddlyWiki5/pull/4648/files#diff-a3001524fc2e41cfa4b70d68377c2b1a)
- [core/modules/server/server.js](https://github.com/Jermolene/TiddlyWiki5/pull/4648/files#diff-799b99f2e1a91901538040f07465f1cd)
- [plugins/tiddlywiki/filesystem/filesystemadaptor.js](https://github.com/Jermolene/TiddlyWiki5/pull/4648/files#diff-304453faeba2d7a772361746dc041c98)

The others were mostly parity changes for commands (changing them to `this.commander.wiki`).

If this PR isn't a good option, I can find a solution for each of these situations. But attaching it to the wiki object would certainly help. Another option would be to make the boot object local in the server and the file system adaptor.

Jermolene:
> I guess I should just say all the changes in these four files.

Ah, were those left off #4601 because they were here?

> If this PR isn't a good option, I can find a solution for each of these situations. But attaching it to the wiki object would certainly help. Another option would be to make the boot object local in the server and the file system adaptor.

Could these components accept an "options" parameter that allows values to be specified in place of those in $tw.boot?

Arlen22:
> Could these components accept an "options" parameter that allows values to be specified in place of those in $tw.boot?

I could add `this.boot` and `state.boot` to all of those files, as applicable.

Arlen22:
> > I guess I should just say all the changes in these four files.
> 
> Ah, were those left off #4601 because they were here?

Adding these to the wiki object resulted in more of those references needing to be changed.

Arlen22:
> > Could these components accept an "options" parameter that allows values to be specified in place of those in $tw.boot?
> 
> I could add `this.boot` and `state.boot` to all of those files, as applicable.

@Jermolene Shall I add these changes to https://github.com/Jermolene/TiddlyWiki5/pull/4649? The changes fit very well with that PR.

Arlen22:
I added it as a single commit to https://github.com/Jermolene/TiddlyWiki5/pull/4649. Feel free to comment on your thoughts or merge with or without that commit.

Arlen22:
Closing in favor of https://github.com/Jermolene/TiddlyWiki5/pull/4649

PR #4649, opened 2020-05-12
Allow setting boot, wiki, and pathPrefix for each request

Arlen22:
This adds the `pathPrefix` to the state object for each request and adds an options argument to the request handler which sets both `wiki` and `pathPrefix` for each request. The options argument is optional.

### DIFF
--- a/core/modules/server/server.js
+++ b/core/modules/server/server.js
@@ -159,11 +159,11 @@ Server.prototype.requestHandler = function(request,response,options) {
 	// Compose the state object
 	var self = this;
 	var state = {};
-	state.wiki = (options?options.wiki:false) || self.wiki;
+	state.wiki = (options ? options.wiki : false) || self.wiki;
 	state.server = self;
 	state.urlInfo = url.parse(request.url);
 	state.queryParameters = querystring.parse(state.urlInfo.query);
-	state.pathPrefix = (options?options.pathPrefix:false) || this.get("path-prefix") || "";
+	state.pathPrefix = (options ? options.pathPrefix : false) || this.get("path-prefix") || "";
 	// Get the principals authorized to access this resource
 	var authorizationType = this.methodMappings[request.method] || "readers";
 	// Check for the CSRF header if this is a write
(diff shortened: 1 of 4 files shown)


core/modules/server/server.js
```diff
@@ -156,14 +155,15 @@ Server.prototype.isAuthorized = function(authorizationType,username) {
 	return principals.indexOf("(anon)") !== -1 || (username && (principals.indexOf("(authenticated)") !== -1 || principals.indexOf(username) !== -1));
 }
 
-Server.prototype.requestHandler = function(request,response) {
+Server.prototype.requestHandler = function(request,response,options) {
 	// Compose the state object
 	var self = this;
 	var state = {};
-	state.wiki = self.wiki;
+	state.wiki = (options?options.wiki:false) || self.wiki;
```
Jermolene, line 162: We use whitespace in the ternary conditional operator: `(options ? options.wiki : false)`

Jermolene: Actually, it would be more in line with existing code to insert a line `options = options || {}` at the top of the function, rather than guarding against each access.

Arlen22: > Actually, it would be more in line with existing code to insert a line `options = options || {}` at the top of the function, rather than guarding against each access.

I considered this, but in the normal use case `options` is always undefined, and I thought this would prevent an extra object from being created but never used. I don't know if this is important or if I'm just over thinking it.

Jermolene: We should be consistent with existing code.

Arlen22: Ok, done.

Arlen22:
I've added the boot property from https://github.com/Jermolene/TiddlyWiki5/pull/4648 to this one since it seemed more fitting. Feel free to comment on your thoughts or merge with or without that commit.

Arlen22:
Ok, I closed the other one in favor of this one. Hopefully this can be merged with that change.

Jermolene:
Thanks @Arlen22

PR #4656, opened 2020-05-15
Git saver: Correctly handle empty path

cd2357:
Skip checking for empty path (which is the default in a new TiddlyWiki and described as optional in the [Git saver documentation](https://tiddlywiki.com/#Saving%20to%20a%20Git%20service)).

The path is already explicitly checked and initialized to "/" if empty.


Jermolene:
Thanks @cd2357! With apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

Issue #4654, opened 2020-05-14
How to fix focusPopups?


BurningTreeC:
**Describe the bug**
focusPopups (dropdowns of input fields) don't cancel correctly when for example field 1 with a focusPopup has focus and then field 2 with a focusPopup gets focus by clicking, the popup of field 2 gets cancelled. If the change of focus occurs by tabbing it happens that no popup gets cancelled. This is especially annoying in Edit Mode

**Possible Solution**

We could explicitely tell each concerned input field with a focusPopup to cancel all popups before triggering its own. Probably through something like a `cancelPopups="yes"` attribute.

I've been trying many different ways to solve these kind of issues and haven't found a universal way to do it, that's why I come up with this idea

PR #4658, opened 2020-05-15
Fix focusPopups like described in #4654

BurningTreeC:
this PR fixes the sticky focusPopups of tiddlers in edit mode and the sidebar search by cancelling all popups when the input gets focus and before an eventual focusPopup is triggered


Jermolene:
Hi @BurningTreeC thanks, the changes look reasonable at a first glance, but given the brittleness of the popup mechanism I think we'll have to do some fairly careful testing of this.

For example, there are a few UI elements that don't look like popups but are implemented that way (eg the tiddler info panel). It might be helpful to identify and verify that these cases aren't impacted by these changes.

BurningTreeC:
The PR can be tested [here](http://focus-popups-fix.tiddlyspot.com/) @Jermolene

BurningTreeC:
Perhaps this could be realized with a `tv-config-cancel-popups` variable like the `tv-config-wiki-links`?

Jermolene:
Thanks @BurningTreeC

Jermolene:
Hi @BurningTreeC I only thought to try it after merging, but I think this breaks usage of text inputs within dropdowns.

BurningTreeC:
Yes @Jermolene, it isn't thought for text inputs within drop downs, only for inputs with the `cancelPopups=yes` attribute

Jermolene:
Thanks @BurningTreeC.

PR #4659, opened 2020-05-16
Compressing $set chains into one $vars

flibbles:
This change makes the code:

1. Easier to read
2. more concise
3. use less memory
4. run faster, since vars is more lightweight, and this will parse faster.
5. trim 12 whole floors off those flame chart skyscrapers Tiddlywiki has.

I've been using this improvement on my machine a while to no effect, and looking at it, I can't see how it possibly changes runtime.

(Quick note: I noticed that a month ago, you added **tv-enable-drag-and-drop** to _ui/PageTemplate_, but not to _templates/single.tiddler.window.tid_. Was that an oversight? I'm not sure)

ANYWAY

This tiny improvement is just an example of why $vars is amazing. And I'm really making this PR to promote the following addition to $vars:
```javascript
VarsWidget.prototype.getVariableInfo = function(name,options) {
    // $vars is the only widget which looks in itself when fetching a variable
    // rather than looking at its parent or higher
    if ($tw.utils.hop(this.variables, name)) {
        var variable = this.variables[name];
        // It doesn't have to perform any param or macro stuff, since vars
        // only ever contains simple variables.
        return {
            text: variable.value,
            params: []
        };
    } else {
        return Widget.prototype.getVariableInfo.call(this,name,options);
    }
};
```

That's right. We make $vars the one widget which can look into itself when scrounging up variables. Why? For chaining definitions.
```html
<$vars	tiddler={{myvalue}}
		another={{{ [<tiddler>get[field]] }}}
		third={{{ [tag[another]] }}}>
```
Defining variables in a vars chain can depend on ones before it. It's exactly like javascript. Everywhere in the JS code, you see stuff like,
```javascript
vars tiddler=this.getVariable("currentTiddler"),
	value=tiddler.fields.value;
```
It makes perfect sense that $vars would behave this way. Completely intuitive. And now $vars can be used over cumbersome $set chains more often.

But there are even more advantages to this.

### wiki.makeWidget

You know all those times in code when you make dummy widgets just to run code with context? For instance, wiki.makeWidget is this weird complicated thing that might actually make a TON of widgets. At least 2.
```javascript
exports.makeWidget = function(parser,options) {
    options = options || {};
    var widgetNode = {
            type: "widget",
            children: []
        },
        currWidgetNode = widgetNode;
    // Create set variable widgets for each variable
    $tw.utils.each(options.variables,function(value,name) {
        var setVariableWidget = {
            type: "set",
            attributes: {
                name: {type: "string", value: name},
                value: {type: "string", value: value}
            },
            children: []
        };
        currWidgetNode.children = [setVariableWidget];
        currWidgetNode = setVariableWidget;
    });
    // Add in the supplied parse tree nodes
    currWidgetNode.children = parser ? parser.tree : [];
    // Create the widget
    return new widget.widget(widgetNode,{
        wiki: this,
        document: options.document || $tw.fakeDocument,
        parentWidget: options.parentWidget
    });
};
```
...gross

But this becomes WAY simpler if all you have to make is a $vars widget instead of a $widget widget. It can hold everything, and then you attach parser tree to that. And if you want to to run filters from this, you don't have to climb to the bottom of the returned widget's children. _You can use the widget returned by makeWidget._

### importVariablesWidget

I've turned $importvariables into a highly efficient yet monstrously complicated widget. With this modification to vars (and uh... one other one I'll talk about another time), I can turn importvariables back into something other people can understand, _and still retain the efficiency!_

So how about it? The change is small, and I'd accompany some tests along with it to make sure it's not breaking anything.

### DIFF
--- a/core/templates/single.tiddler.window.tid
+++ b/core/templates/single.tiddler.window.tid
@@ -6,17 +6,13 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 \end
 \import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
 
-<$set name="tv-config-toolbar-icons" value={{$:/config/Toolbar/Icons}}>
-
-<$set name="tv-config-toolbar-text" value={{$:/config/Toolbar/Text}}>
-
-<$set name="tv-config-toolbar-class" value={{$:/config/Toolbar/ButtonClass}}>
-
-<$set name="tv-show-missing-links" value={{$:/config/MissingLinks}}>
-
-<$set name="storyviewTitle" value={{$:/view}}>
-
-<$set name="languageTitle" value={{{ [{$:/language}get[name]] }}}>
+<$vars
+	tv-config-toolbar-icons={{$:/config/Toolbar/Icons}}
+	tv-config-toolbar-text={{$:/config/Toolbar/Text}}
+	tv-config-toolbar-class={{$:/config/Toolbar/ButtonClass}}
+	tv-show-missing-links={{$:/config/MissingLinks}}
+	storyviewTitle={{$:/view}}
+	languageTitle={{{ [{$:/language}get[name]] }}}>
 
 <div class=<<containerClasses>>>
 
@@ -28,14 +24,4 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 
 </div>
 
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
+</$vars>
--- a/core/ui/PageTemplate.tid
+++ b/core/ui/PageTemplate.tid
@@ -6,19 +6,14 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 \end
 \import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
 
-<$set name="tv-config-toolbar-icons" value={{$:/config/Toolbar/Icons}}>
-
-<$set name="tv-config-toolbar-text" value={{$:/config/Toolbar/Text}}>
-
-<$set name="tv-config-toolbar-class" value={{$:/config/Toolbar/ButtonClass}}>
-
-<$set name="tv-enable-drag-and-drop" value={{$:/config/DragAndDrop/Enable}}>
-
-<$set name="tv-show-missing-links" value={{$:/config/MissingLinks}}>
-
-<$set name="storyviewTitle" value={{$:/view}}>
-
-<$set name="languageTitle" value={{{ [{$:/language}get[name]] }}}>
+<$vars
+	tv-config-toolbar-icons={{$:/config/Toolbar/Icons}}
+	tv-config-toolbar-text={{$:/config/Toolbar/Text}}
+	tv-config-toolbar-class={{$:/config/Toolbar/ButtonClass}}
+	tv-enable-drag-and-drop={{$:/config/DragAndDrop/Enable}}
+	tv-show-missing-links={{$:/config/MissingLinks}}
+	storyviewTitle={{$:/view}}
+	languageTitle={{{ [{$:/language}get[name]] }}}>
 
 <div class=<<containerClasses>>>
 
@@ -38,16 +33,4 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 
 </div>
 
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
-
-</$set>
+</$vars>


saqimtiaz:
@flibbles oddly the diff under `Files Changed` is not showing me the changes to the vars widget javascript code. Did you perhaps forget to merge/commit some changes? Or do you intend that as a separate PR?

Otherwise, very intriguing. I've already taken to using vars (with filtered attributes) over set, to avoid the plethora of set widgets I often end up with otherwise.

pmario:
@saqimtiaz .. I think it's easier to get it merged, if it will be 2 separated PRs. I think, this PR would be an easy merge, since $vars was created to do exactly that. Nobody did take the time to clean the core UI templates. .. yet.

pmario:
@flibbles .. I will +1 everything that makes the core more efficient. We only need to take care, that it doesn't have side effects. 

----------

I haven't been aware of this:

```
<$vars	tiddler={{myvalue}}
	another={{{ [<tiddler>get[field]] }}}
	third={{{ [tag[another]] }}}>
```

That's nice and should probably be documented. 

------

I think you should go on with the other PRs. _All of them_. I'd be interested in the differences of the flame-chart.

Jermolene:
Hi @flibbles I'd thought about this, and agree with the benefits, but I was concerned that using the vars widget in this way relies of the ordering of "Object.keys()", which I thought was browser-specific, but it turns out that it's been specced since ES2015. But the trouble is that we're still targetting ES5/6 to stay compatible with old, cheap Android phones. Anyhow, the ordering isn't even strictly in insertion order (see [here](https://stackoverflow.com/a/38218582)), because keys that look like integers are sorted first.

flibbles:
Hmmmgraaah!

@Jermolene, I see you're right. There might be ways to get around this through lazy attribute evaluation, but that'd be complicated, and it'd raise the issues that all $vars attributes would be in scope for each other, which probably has it's own backward compat problems. ...and possible infinite recursion problems.

I'll let it go. For now. But none of the variables in my PR affect each other, so they can still be merged, right?

@saqimtiaz, @pmario is right. I only put in the tiny core ui change first because I know that's safe. The idea of revamping needed discussion.

Jermolene:
Thanks @flibbles

pmario:
@Jermolene 
>But the trouble is that we're still targetting ES5/6 to stay compatible with old, cheap Android phones. 

Do you want to target ES5 (from 2009) or ES6 (aka ES2015). IMO this makes a big difference. 

There is a loot of [cool stuff](https://medium.com/@madasamy/javascript-brief-history-and-ecmascript-es6-es7-es8-features-673973394df4) in ES6 ... like map, set, weakmap, weakset. map returns keys in order.

pmario:
Browser compatibility table ES6 https://kangax.github.io/compat-table/es6/

Jermolene:
> Do you want to target ES5 (from 2009) or ES6 (aka ES2015). IMO this makes a big difference

Argh, sorry I always get the names wrong. I was describing the JS version that the core currently uses, and I think I mean ES5, but I believe we use some features from ES6 that browsers had already implemented early.

We could consider moving to a later dialect but we'd leave IE11 out in the rain, which is a shame since we've gone to all the work to maintain a high level support for it. For the core, I think we should stay where we are unless there was a significant benefit. All of this only applies to the core, of course, there's nothing to stop plugins from using whatever JS dialect they like.

pmario:
Yea, IE11 and Edge (old) are the new IE6 now :/ That's a shame. .. 

IMO the good thing is, that Chromium based Edge eats into usage number of IE 11 which seems to be at 1.8% already. According to: https://www.stetic.com/market-share/browser/

Wikipedia has an interesting overview too: https://en.wikipedia.org/wiki/Usage_share_of_web_browsers

Jermolene:
Hi @flibbles can this be closed or do you plan to continue?

flibbles:
It can be merged actually. The proposed change for $vars won't work, but I still stand by my suggestion that $vars should be used over $set in all places it can be. The change in this PR does improve the code.

flibbles:
Also, same question:

I noticed that a month ago, you added **tv-enable-drag-and-drop** to _ui/PageTemplate_, but not to _templates/single.tiddler.window.tid_. Was that an oversight? I'm not sure.

Arlen22:
I once needed to use the flame chart to diagnose some performance issues and had to modify the set widget to loop through its children instead of calling refreshChildren because it is a recursive function call and loads up the flame chart with so much data it can hardly even render. It was a pretty simple hack once I figured it out. 

@Jermolene I think flibbles is correct that this can be merged because the self-reference scenario is not used in core.

Arlen22:
@Jermolene oops, maybe I'm missing something, but it doesn't look compete yet.

flibbles:
This is complete.

Changing $vars so it can be self-referential turns out to be impossible with the current architecture. The changes I have here is just changing a cluster of set widgets used in every single tiddler to instead be one vars widget. It trims 12 function calls off the stack depth, and makes it easier to read.

Arlen22:
Right, sorry, my mistake.

Arlen22:
```js
/*
Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
*/
SetWidget.prototype.refresh = function(changedTiddlers, iterated) {
	if(changedAttributes.name || changedAttributes.filter || changedAttributes.select || changedAttributes.tiddler || (this.setTiddler && changedTiddlers[this.setTiddler]) || changedAttributes.field || changedAttributes.index || changedAttributes.value || changedAttributes.emptyValue ||
	   (this.setFilter && this.getValue() != this.variables[this.setName].value)) {
		this.refreshSelf();
		return true;
	} else if(this.children.length === 1 && this.children[0] instanceof SetWidget) {
		if(iterated) return false;
		let child = this.children[0];
		while(child instanceof SetWidget){
			if (child.children.length !== 1)
				return this.refreshChildren(changedTiddlers);
			else if(child.refresh(changedTiddlers, true)) 
				return true;
			else child = child.children[0];
		}
		console.log(new Error("I guess a set widget was last")) 
	} else {
		return this.refreshChildren(changedTiddlers);
	}
};
```

flibbles:
Again, yeah. This PR should just be merged as is. But the perfectionist in me saw that debug-friendly function and thought, "couldn't it be cleaner?"

```javascript

SetWidget.prototype.refresh = function(changedTiddlers) {
    var current = this;
    while (true) {
        var changedAttributes = current.computeAttributes();
        if(changedAttributes.name || changedAttributes.filter || changedAttributes.select || changedAttributes.tiddler || (current.setTiddler && changedTiddlers[current.setTiddler]) || changedAttributes.field || changedAttributes.index || changedAttributes.value || changedAttributes.emptyValue ||
        (current.setFilter && current.getValue() != current.variables[current.setName].value)) {
            current.refreshSelf();
            return true;
        }
        if (current.children.length !== 1 || !(current.children[0] instanceof SetWidget)) {
            break;
        }
        current = current.children[0];
    }
    return current.refreshChildren(changedTiddlers);
};
```
Nobody use this. Please just merge this PR, but I had to post this. I think it should be a little smoother to iterate through in a debugger.

Jermolene:
Thanks @flibbles apologies for the delay.

pmario:
@flibbles .. Should this https://github.com/Jermolene/TiddlyWiki5/pull/4659#issuecomment-643359337 be a separate PR?

saqimtiaz:
@pmario @flibbles it should at least be an issue if not a PR so that we can come back to it once IE support is no longer a consideration.

pmario:
@saqimtiaz ... How is it related to IE support. I may be missing something.

saqimtiaz:
@pmario It is my understanding that the blocker for this change is that we are targeting ES5, which in turn is due to wanting to provide suppor for IE. See:
https://github.com/Jermolene/TiddlyWiki5/pull/4659#issuecomment-629642632

flibbles:
It is problematic to ensure ordering, even given IE6's limitations.
1. Attributes with numbers for names will get reordered on any browser
2. It turns out since we first A) convert it into a parseTreeNode, then B) convert it into a widget, there's some reason why this makes it difficult. (I don't remember why. It's been months since I looked at it. I just remember something about this makes it tricky, and would take more under-the-hood work than initially expected.)

PR #4661, opened 2020-05-16
Delete or update RoadMap.tid

twMat:
Is this still valid? I hope "yes" but would guess "no". Maybe update or delete?


Jermolene:
Thanks @twMat

PR #4671, opened 2020-05-22
Add modifier variable to button actionstring

BurningTreeC:
This PR adds a `modifier` variable to the button actions parameter, which contains the modifier keys pressed when the actions are invoked

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -91,13 +91,7 @@ ButtonWidget.prototype.render = function(parent,nextSibling) {
 			handled = true;
 		}
 		if(self.actions) {
-			var modifierKey = event.ctrlKey && !event.shiftKey && !event.altKey ? "ctrl" : 
-				event.shiftKey && !event.ctrlKey && !event.altKey? "shift" : 
-                		event.ctrlKey && event.shiftKey && !event.altKey ? "ctrl-shift" : 
-				event.altKey && !event.shiftKey && !event.ctrlKey ? "alt" : 
-				event.altKey && event.shiftKey && !event.ctrlKey ? "alt-shift" : 
-				event.altKey && event.ctrlKey && !event.shiftKey ? "ctrl-alt" : 
-				event.altKey && event.shiftKey && event.ctrlKey ? "ctrl-alt-shift" : "normal";
+			var modifierKey = $tw.keyboardManager.getEventModifierKeyDescriptor(event);
 			self.invokeActionString(self.actions,self,event,{modifier: modifierKey});
 		}
 		if(handled) {
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -91,7 +91,14 @@ ButtonWidget.prototype.render = function(parent,nextSibling) {
 			handled = true;
 		}
 		if(self.actions) {
-			self.invokeActionString(self.actions,self,event);
+			var modifierKey = event.ctrlKey && !event.shiftKey && !event.altKey ? "ctrl" : 
+				event.shiftKey && !event.ctrlKey && !event.altKey? "shift" : 
+                		event.ctrlKey && event.shiftKey && !event.altKey ? "ctrl-shift" : 
+				event.altKey && !event.shiftKey && !event.ctrlKey ? "alt" : 
+				event.altKey && event.shiftKey && !event.ctrlKey ? "alt-shift" : 
+				event.altKey && event.ctrlKey && !event.shiftKey ? "ctrl-alt" : 
+				event.altKey && event.shiftKey && event.ctrlKey ? "ctrl-alt-shift" : "normal";
+			self.invokeActionString(self.actions,self,event,{modifier: modifierKey});
 		}
 		if(handled) {
 			event.preventDefault();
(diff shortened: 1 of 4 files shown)


core/modules/widgets/button.js
```diff
@@ -91,7 +91,14 @@ ButtonWidget.prototype.render = function(parent,nextSibling) {
...
 		if(self.actions) {
-			self.invokeActionString(self.actions,self,event);
+			var modifierKey = event.ctrlKey && !event.shiftKey && !event.altKey ? "ctrl" : 
+				event.shiftKey && !event.ctrlKey && !event.altKey? "shift" : 
+                		event.ctrlKey && event.shiftKey && !event.altKey ? "ctrl-shift" : 
+				event.altKey && !event.shiftKey && !event.ctrlKey ? "alt" : 
+				event.altKey && event.shiftKey && !event.ctrlKey ? "alt-shift" : 
+				event.altKey && event.ctrlKey && !event.shiftKey ? "ctrl-alt" : 
+				event.altKey && event.shiftKey && event.ctrlKey ? "ctrl-alt-shift" : "normal";
+			self.invokeActionString(self.actions,self,event,{modifier: modifierKey});
```
pmario, line 101: Did you test them with all OSes. Windows, unix, MacOS?

saqimtiaz: I recommend either modelling after how this is done in the droppable widget, or also updating that widget to match for the sake of consistency: 
https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/droppable.js#L133

BurningTreeC: tested with linux and windows, not macOS @pmario

BurningTreeC: @saqimtiaz you're right. I wonder what's been the reasoning for omitting the alt-key in droppable.js ... maybe it was a macOS thing

Jermolene:
Hi @BurningTreeC are you able to change this along the lines proposed?

BurningTreeC:
@Jermolene - what should I change? I think I've addressed all points above ...  what do you think @pmario @saqimtiaz ?

saqimtiaz:
Looks consistent to me. Droppable.js has also been updated with the same pattern and support for ALT as a modifier.

@Jermolene was there a specific reason for omitting the ALT key in the list of modifiers supported in droppable.js? If so, perhaps worth testing on OSX before merging.

Jermolene:
Hi @saqimtiaz I think it was just an oversight. So perhaps we should add a new static method to the KeyboardManager class to encapsulate the logic?

saqimtiaz:
@Jermolene agreed, will reduce duplication and make maintenance and re-use easier as well.

saqimtiaz:
Here is a quick take, though not using a static method since we have an instance of KeyboardManager available:

```javascript
KeyboardManager.prototype.getEventModifierKeyDescriptor = function(event) {
	return event.ctrlKey && !event.shiftKey && !event.altKey ? "ctrl" : 
		event.shiftKey && !event.ctrlKey && !event.altKey? "shift" : 
		event.ctrlKey && event.shiftKey && !event.altKey ? "ctrl-shift" : 
		event.altKey && !event.shiftKey && !event.ctrlKey ? "alt" : 
		event.altKey && event.shiftKey && !event.ctrlKey ? "alt-shift" : 
		event.altKey && event.ctrlKey && !event.shiftKey ? "ctrl-alt" : 
		event.altKey && event.shiftKey && event.ctrlKey ? "ctrl-alt-shift" : "normal";
};
```

Used as follows:

```javascript
DroppableWidget.prototype.performActions = function(title,event) {
	if(this.droppableActions) {
		var modifierKey = $tw.keyboardManager.getEventModifierKeyDescriptor(event);
		this.invokeActionString(this.droppableActions,this,event,{actionTiddler: title, modifier: modifierKey});
	}
};
```

BurningTreeC:
@saqimtiaz  @Jermolene  @pmario I've updated the code to reflect the proposal of a separate method in our `keyboardManager`

saqimtiaz:
@Jermolene thoughts? Happy to revise if necessary.

pmario:
I'm able to test stuff again starting from sunday.

BurningTreeC:
@saqimtiaz @pmario @Jermolene ... the linkcatcher could also profit of the `modifier` variable available within the `actions string` ... what do you think? Should I add this to the linkcatcher widget, too?

saqimtiaz:
@BurningTreeC good idea

pmario:
IMO linkcatcher should be a new PR. 

... linkcatcher. ... I would like to add more info to linkcatcher widget. IMO it should also get info about the "storylist and historylist" variables. ... 

This will make it much easier, to create "multi-story" navigation. ... And I would like to be able to "nest" linkcatchers with a possibility to "bubble" all the event info. 

just some thoughts.

BurningTreeC:
@pmario HOW would you like to implement your proposal?

saqimtiaz:
@Jermolene at your convenience, please review the addition of the `KeyboardManager.prototype.getEventModifierKeyDescriptor` method and whether you are happy with this approach. If it needs revision please let us know, otherwise this can be merged.

@BurningTreeC I suggest a separate PR for the linkcatcher, adding the modifier key, just in order to get this PR merged sooner rather than later. That PR need not do anything more than add the modifier key and if @pmario feels that would somehow clash with future improvements he has in mind, it can be discussed there. Small gains need not be delayed in favour of larger changes if the two approaches do not inherently clash.

Jermolene:
Thanks @BurningTreeC @saqimtiaz @pmario this looks good to me.

PR #4676, opened 2020-05-23
Incorrect modified field glitches timeline macro

saqimtiaz:


Jermolene:
Ouch! Thanks @saqimtiaz

PR #4680, opened 2020-05-24
Corrected incorrect image source for markdown

flibbles:
Making a plugin for relinking markdown and noticed a problem.

If you have an image link with any kind of escape characters in it, like `Alt text`, it will not render. Escape characters are the only way to specify tiddlers with special characters in markdown, so consequently you can't markdown-image link to those tiddlers at all without this fix.


Jermolene:
Thanks @flibbles. This flags another frustrating issue with the Markdown plugin. Requiring users to apply URL encoding to internal link targets is a poor user experience. But the fact that the Markdown parser doesn't even recognise `[Foo Bar](#Foo Bar)` as a link doesn't give us much option.

Issue #4679, opened 2020-05-23
[BUG] Line breaks rendered incorrectly for Markdown tiddlers


ifethereal:
Currently being addressed by #4684.

---

**Describe the bug**
Line breaks are rendered incorrectly for Markdown tiddlers if `breaks` is `false`. That is, `<br>` appears when `<br>` is meant to be absent; `<br>` is absent when `<br>` is meant to be present.

**To Reproduce**
1. Go to [the _MarkdownExample_ tiddler on the Markdown demo](https://tiddlywiki.com/plugins/tiddlywiki/markdown/#%24%3A%2Fconfig%2Fmarkdown%2Fbreaks:%24%3A%2Fconfig%2Fmarkdown%2Fbreaks%20MarkdownExample).
2. Note that the `breaks` setting is `false` for the Markdown plugin.
3. Observe the multiple `<br>` present in the _MarkdownExample_ tiddler by inspecting the source code of the page (although it can be noticed by eye too).
4. Add two spaces at the end of any line inside a regular plain-text paragraph of the _MarkdownExample_ tiddler which is not followed by a blank line.
5. Observe the disappearance of a `<br>`.

I have observed the same behaviour for Markdown tiddlers on my own TiddlyWiki (running on a Node.js server).

**Expected behavior**
Based on what I understand of the `breaks` setting for Remarkable, since `breaks` is `false`, there should have been no `<br>` within any one paragraph in the _MarkdownExample_ tiddler, and adding the two spaces should have introduced a `<br>`.

**Screenshots**
Using the preview pane to get everything on one screen. (There are two spaces after `This page offers a brief overview of what it's like to use Markdown.`.)

![Screenshot](https://i.imgur.com/JVxq7M3.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Google Chrome 83.0.4103.61, Mozilla Firefox 76.0.1
 - Version: TiddlyWiki 5.1.22

**Additional context**
N/A

pmario:
I can confirm the issue. We will have a closer look.

flibbles:
Weird how we found this issue separately in the same week.

flibbles:
I probably should have just said I have PR fix standing by.

saqimtiaz:
This was resolved in #4684  and be closed @flibbles @Jermolene

Jermolene:
Thanks @saqimtiaz @flibbles

PR #4684, opened 2020-05-27
Corrected issue with markdowns softbreaks AND whitespace between md markup

flibbles:
I was resolving an issue for someone using Relink-markdown when we realized any kind of multi-line wikiSyntax is broken when using markdown. It's because wrapper isn't respecting the $:/config/markdown/breaks setting. It was effectively always "true". So...

```markdown
These lines
should be merged after rendering if breaks==false
but they weren't
```
Also, fixed wrapper so it handles hardbreaks now, which is the proper way to force `<br>` when breaks==false. That's when you put 2 spaces at the end of the line.

Warning, this will change existing rendering somewhat, since markdown is currently rendering everyone's stuff wrong.

### DIFF
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -225,7 +225,8 @@ function convertNodes(remarkableTree, isStartOfInline) {
 		) {
 			// The Markdown compiler thinks this is just text.
 			// Hand off to the WikiText parser to see if there's more to render
-			if (!pluginOpts.renderWikiText) {
+			// But only if it's configured to, and we have more than whitespace
+			if (!pluginOpts.renderWikiText || accumulatedText.match(/^\s*$/)) {
 				out.push({
 					type: "text",
 					text: accumulatedText
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -32,6 +32,10 @@ var remarkableOpts = {
 	quotes: $tw.wiki.getTiddlerText("$:/config/markdown/quotes"),
 	typographer: parseAsBoolean("$:/config/markdown/typographer")
 };
+var accumulatingTypes = {
+	"text": true,
+	"softbreak": true
+};
 
 var md = new Remarkable(remarkableOpts);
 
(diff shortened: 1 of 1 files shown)


flibbles:
Just noticed that someone actually opened an issue about this: #4679

ifethereal:
I "tested" this (by replacing the shadow tiddler for `wrapper.js` --- please let me know if this is invalid) and, while it does now respect the `breaks` setting properly, it seems to gobble spaces after any piece of formatting as well as before any piece of formatting right after a softbreak,

Are you able to reproduce this?

For example, with this Markdown (which is hard-wrapped at 80 characters)
```markdown
There is meant to be a space after this [link](http://example.com) unless there
is something wrong with rendering.

I can try again with this reference-style [link][1] but if there is again no
space then there might be an issue. Seems like it happens **after** _any_ `type`
of formatting?

It also seems not to produce a space if there is a sentence that happens to last
[across][1] a softbreak with the next line (in the source) starting with some
_formatting_?

[1]: http://example.com
```

rendered like this for me

flibbles:
@ifethereal, you are correct. I introduced a bug that erased the whitespace after markdown patterns. I've pushed a fix for that.

But the problem where ``it would happens **after** _any_ `type`
of formatting`` was actually already there. Whitespace between markdown patterns was getting forgotten. I have however, resolved that issue in this same PR (since it affects the already-touched code). I'll rename the title accordingly.

pmario:
>Warning, this will change existing rendering somewhat, since markdown is currently rendering everyone's stuff wrong.

So it's a bugfix, that should be added, because current behaviour is definitely wrong. right?

flibbles:
It is definitely wrong. I only mention it because some people using the plugin today might be depending on that wrong behavior.

We could change the "break" setting to be true by default, though then it would different from the Remarkable default behavior.

I personally say change it. One should be able to expect their imported md files to render in Tiddlywiki the same way they'd render elsewhere.

flibbles:
By "change it", I mean we should change the behavior. Not the default break setting.

ifethereal:
Thanks @flibbles for updating this. Those latest commits fix the issue I had, except for the spacing right after `` `type` `` in my example above or when the formatting is just before a softbreak like in
```markdown
Lastly, there seems to be an issue in that there's no space following _markup_
right before a softbreak.
```

flibbles:
Hell's bells. You're right. I'll patch this up before midnight tonight or I'll turn in my badge.

flibbles:
Okay. One more commit added (and hopefully the last). Problem resolved. It's another existing issue that crept up because the `break` setting now works. So while it was there, I should have caught it. 

I really wish the tiddlywiki plugins had a testing framework so I didn't feel like I'm whacking moles with this bug fixing.

flibbles:
Thanks, by the way, for testing this PR before merge. You're probably saving a lot of future people a headache I would have given them.

ifethereal:
Thanks for the extra commit. I don't see any more spacing/softbreak issues!

chunjiw:
Just wondering, after you (awesome) guys fix it via a few commits, how do I apply it to my own wiki? Just simply update to the latest version? Or do I have to wait till next stable release?

flibbles:
If you're using a standalone html tiddlywiki, you'll need to wait for the next release
If you're using a nodeJS tiddlywiki server, you can get it as soon as it's merged. Just pull.

If you need it right now, you can [download the raw file](https://raw.githubusercontent.com/Jermolene/TiddlyWiki5/7859b96fcb06f086b38a678c9a73ce17802e003e/plugins/tiddlywiki/markdown/wrapper.js) and drop it into your tiddlywiki. It'll override the existing shadow tiddler. Just make sure you remove the override when you do eventually update to a version which has the fix.

chunjiw:
That's super clear, thank you!

Jermolene:
Thanks @flibbles @chunjiw @ifethereal @pmario

PR #4686, opened 2020-05-29
Type EditTemplate: show dropdown at correct position ...

BurningTreeC:
... (below type input field, not at the very left)

This PR adds a wrapper div with the class `tc-type-selector-wrapper` that gets `display: inline-block` what makes the positioning of the dropdown work


Jermolene:
Hi @BurningTreeC please could you resolve the conflicts? It would also be useful to update the OP to include a before and after picture. Many thanks.

PR #4696, opened 2020-06-03
Droppable widget: Check for DOM node existence before removing class.

saqimtiaz:
Can cause issues if dragstart and dragend actions trigger a refresh.


Jermolene:
Thanks @saqimtiaz

PR #4706, opened 2020-06-12
Make tags in tags-dropdown change bg-color on hover

BurningTreeC:
This PR makes the tags in the tags-dropdown (edit-mode) change their background-color to the primary color on hover


saqimtiaz:
@BurningTreeC  appreciate all the attention to detail! A screenshot of the before/after would be very helpful.

BurningTreeC:
Note: the mouse hovers on a tag in the tags-dropdown

Before:
Bildschirmfoto von 2020-06-12 09-44-53
After:
Bildschirmfoto von 2020-06-12 09-44-29

saqimtiaz:
@BurningTreeC Thank you!

BurningTreeC:
Now this PR uses the correct class `tc-btn-invisible` instead of (mistake) `tc-button-invisible`

PR #4711, opened 2020-06-13
Fixes use of stateTitle instead of state resulting in artifacts during import

### LINKED ISSUE #4599: [BUG] Import process leaves behind $:/Import!!popup-filename tiddlers
**Describe the bug**
If during the process of importing a file into TW, you decide to preview the file before hitting Import, it leaves behind a tiddler with name:
$:/Import!!popup-filename for every file previewed. Since these are not state or temp tiddlers, they persist.

**To Reproduce**
1. Open any tiddlywiki.
2. Drag and drop a file for import
3. Click the down arrow to preview the file
4. $:/Import!!popup-filename is created and will persist

Is there a reason we don't use state tiddlers here?

saqimtiaz:
Fixes #4599 

If during the process of importing a file into TW, you decide to preview the file before hitting Import, it leaves behind a tiddler with name:
$:/Import!!popup-filename for every file previewed.

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -1,5 +1,3 @@
-title: $:/core/ui/ImportListing
-
 \define lingo-base() $:/language/Import/
 
 \define messageField()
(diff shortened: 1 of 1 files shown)


core/ui/ImportListing.tid
```diff
@@ -1,5 +1,3 @@
-title: $:/core/ui/ImportListing
```
Jermolene, line 1: Is removing the title a mistake?

saqimtiaz: Indeed! Apologies. Should be resolved.

Jermolene:
Thanks @saqimtiaz!

PR #4713, opened 2020-06-13
Resolve conflicting `padding` property for `pre`

unindented:
`padding: 14px` was getting overridden by `padding: 0 3px 2px` below, which I believe was a mistake.


Jermolene:
Thanks @unindented, and apologies for the delay.

It seems that both `padding` properties were set in the original commit 7175f1cbf1f9966d718b99f294425f17b8a31ac9, so I think in fact it's the `padding: 14px` that shouldn't be there.

PR #4716, opened 2020-06-15
PreferredSaver

xmaysonnave:
This PR supersed the closed PR https://github.com/Jermolene/TiddlyWiki5/pull/4371

### Motivation
The mechanism is meant to let users change from the UI their preferred saver.
This feature help users to save their wiki when a saver has failed.

### Artifacts
1. core/language/en-GB/ControlPanel.multids
2. core/modules/saver-handler.js
3. core/modules/filters/savers.js
4. core/modules/macros/infosaver.js
5. core/ui/ControlPanel/Saving/General.tid
6. core/wiki/config/PreferredSaver.tid

### Remarks
There is a debate whether or not savers should handle multiple savers sequentially during a save session. I see a blocking point here as a plugin can decide to reload its wiki once saved preventing any further saver processing.

### DIFF
--- a/core/modules/filters/savers.js
+++ b/core/modules/filters/savers.js
@@ -1,0 +1,24 @@
+/*\
+title: $:/core/modules/filters/savers.js
+type: application/javascript
+module-type: filteroperator
+
+This Filter operator return savers title
+
+\*/
+;(function () {
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  'use strict'
+
+  /*
+  Export our filter function
+  */
+  exports.savers = function (source, operator, options) {
+    var results = []
+    for (var i in $tw.saverHandler.savers) {
+      results.push($tw.saverHandler.savers[i].title)
+    }
+    return results
+  }
+})()
--- a/core/modules/saver-handler.js
+++ b/core/modules/saver-handler.js
@@ -93,15 +93,15 @@ function SaverHandler(options) {
 	}
 	// Install the save action handlers
 	if($tw.browser) {
-		$tw.rootWidget.addEventListener("tm-save-wiki",function(event) {
-			self.saveWiki({
+		$tw.rootWidget.addEventListener("tm-save-wiki",async function(event) {
+			await self.saveWiki({
 				template: event.param,
 				downloadType: "text/plain",
 				variables: event.paramObject
 			});
 		});
-		$tw.rootWidget.addEventListener("tm-download-file",function(event) {
-			self.saveWiki({
+		$tw.rootWidget.addEventListener("tm-download-file",async function(event) {
+			await self.saveWiki({
 				method: "download",
 				template: event.param,
 				downloadType: "text/plain",
@@ -125,15 +125,22 @@ SaverHandler.prototype.initSavers = function(moduleType) {
 	var self = this;
 	$tw.modules.forEachModuleOfType(moduleType,function(title,module) {
 		if(module.canSave(self)) {
-			self.savers.push(module.create(self.wiki));
+			self.savers.push({title:title,module:module.create(self.wiki)});
 		}
 	});
-	// Sort the savers into priority order
+	// Sort savers
+	this.sortSavers();
+};
+
+/*
+Sort the savers into priority order
+*/
+SaverHandler.prototype.sortSavers = function() {
 	this.savers.sort(function(a,b) {
-		if(a.info.priority < b.info.priority) {
+		if(a.module.info.priority < b.module.info.priority) {
 			return -1;
 		} else {
-			if(a.info.priority > b.info.priority) {
+			if(a.module.info.priority > b.module.info.priority) {
 				return +1;
 			} else {
 				return 0;
@@ -148,7 +155,7 @@ Save the wiki contents. Options are:
 	template: the tiddler containing the template to save
 	downloadType: the content type for the saved file
 */
-SaverHandler.prototype.saveWiki = function(options) {
+SaverHandler.prototype.saveWiki = async function(options) {
 	options = options || {};
 	var self = this,
 		method = options.method || "save";
@@ -172,18 +179,91 @@ SaverHandler.prototype.saveWiki = function(options) {
 				$tw.notifier.display(self.titleSavedNotification);
 				if(options.callback) {
 					options.callback();
-				}
-			}
-		};
-	// Call the highest priority saver that supports this method
-	for(var t=this.savers.length-1; t>=0; t--) {
-		var saver = this.savers[t];
-		if(saver.info.capabilities.indexOf(method) !== -1 && saver.save(text,method,callback,{variables: {filename: variables.filename}})) {
-			this.logger.log("Saving wiki with method",method,"through saver",saver.info.name);
-			return true;
-		}
-	}
-	return false;
+        }
+      }
+    };
+  // Process preferred if any
+  var ignorePreferred = null
+  var preferredSaver = $tw.wiki.getTiddler("$:/config/PreferredSaver")
+  if (preferredSaver !== null && preferredSaver !== undefined) {
+    var title = preferredSaver.getFieldString("text")
+    title =
+      title === undefined || title == null || title.trim() === ''
+        ? null
+        : title.trim()
+    if (title !== null) {
+      ignorePreferred = title
+      // Process preferred saver
+      if (
+        await this.save(
+          this.getSaver(title).module,
+          method,
+          variables,
+          text,
+          callback
+        )
+      ) {
+        return true
+      }
+    }
+  }
+  // Call the highest priority saver that supports this method
+  for (var t = this.savers.length - 1; t >= 0; t--) {
+    // Ignore failed preferred if any
+    if (this.savers[t].title === ignorePreferred) {
+      continue
+    }
+    // Process
+    if (
+      await this.save(
+        this.savers[t].module,
+        method,
+        variables,
+        text,
+        callback
+      )
+    ) {
+      return true
+    }
+  }
+  return false
+};
+
+SaverHandler.prototype.getSaver = function (title) {
+  // Locate saver
+  var saver = null
+  for (var i = 0; i < this.savers.length; i++) {
+    var current = this.savers[i]
+    if (current.title === title) {
+      saver = current
+      break
+    }
+  }
+  return saver
+};
+
+SaverHandler.prototype.save = async function (
+  saver,
+  method,
+  variables,
+  text,
+  callback
+) {
+  if (saver.info.capabilities.indexOf(method) !== -1) {
+    var saved = await saver.save(text, method, callback, {
+      variables: { filename: variables.filename }
+    })
+    if (saved) {
+      this.logger.log(
+        'Saved wiki with method',
+        method,
+        'through saver',
+        saver.info.name
+      )
+      return true
+    }
+  }
+  return false
 };
 
 /*
--- a/core/wiki/config/PreferredSaver.tid
+++ b/core/wiki/config/PreferredSaver.tid
@@ -1,0 +1,3 @@
+title: $:/config/PreferredSaver
+
+$:/core/modules/savers/download.js
--- a/core/language/en-GB/ControlPanel.multids
+++ b/core/language/en-GB/ControlPanel.multids
@@ -159,6 +159,7 @@ Settings/NavigationPermalinkviewMode/UpdateAddressBar/Description: Update addres
 Settings/PerformanceInstrumentation/Caption: Performance Instrumentation
 Settings/PerformanceInstrumentation/Hint: Displays performance statistics in the browser developer console. Requires reload to take effect
 Settings/PerformanceInstrumentation/Description: Enable performance instrumentation
+Settings/PreferredSaver/Caption: Preferred Saver
 Settings/ToolbarButtonStyle/Caption: Toolbar Button Style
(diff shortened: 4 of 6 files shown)


pmario:
>There is a debate whether or not savers should handle multiple savers sequentially during a save session. I see a blocking point here as a plugin can decide to reload its wiki once saved preventing any further saver processing.

If a plugin want's to reload the wiki, there is a user "dialog". So we can add some info, to let the user know, what's going on

pmario:
This PR contains ES2016 code `async ... await`. TW allows ES5(from 2009 spec) only, because of browser compatibility.

So this code needs to be moved into a plugin. ... I didn't try the code yet, but imo it should be relatively easy to make it a "core plugin".

The only thing we would need in the core, would be the language specific elements.

I personally think, that async await is so much easier to read and maintain as the "callback centric" code we need to use atm. So imo rewriting it, wouldn't make too much sense.

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
+/*\
+title: $:/core/modules/filters/savers.js
+type: application/javascript
+module-type: filteroperator
+
+This Filter operator return savers title
+
+\*/
+;(function () {
```
Jermolene, line 9: Please use the IIFE format used by other core modules (ie without the semicolon).

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
...
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  'use strict'
+
+  /*
+  Export our filter function
+  */
+  exports.savers = function (source, operator, options) {
+    var results = []
+    for (var i in $tw.saverHandler.savers) {
```
Jermolene, line 19: Note that coding style is to omit whitespace between `for` and the following parenthesis

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
...
+;(function () {
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  'use strict'
+
+  /*
+  Export our filter function
+  */
+  exports.savers = function (source, operator, options) {
+    var results = []
```
Jermolene, line 18: Note that coding style is not to omit semicolons

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
...
+\*/
+;(function () {
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  'use strict'
+
+  /*
+  Export our filter function
+  */
+  exports.savers = function (source, operator, options) {
```
Jermolene, line 17: Note that coding style is to omit whitespace between `function` and the opening parenthesis, and to omit whitespace after commas in argument lists.

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
...
+    var results = []
+    for (var i in $tw.saverHandler.savers) {
+      results.push($tw.saverHandler.savers[i].title)
+    }
+    return results
+  }
+})()
```
Jermolene, line 24: Coding style is to include a trailing semicolon after the IIFE

core/modules/filters/savers.js
```diff
@@ -0,0 +1,24 @@
...
+type: application/javascript
+module-type: filteroperator
+
+This Filter operator return savers title
+
+\*/
+;(function () {
+  /*jslint node: true, browser: true */
+  /*global $tw: false */
+  'use strict'
```
Jermolene, line 12: Coding style is to prefer double quotes over single quotes.

core/modules/saver-handler.js
```diff
@@ -93,15 +93,15 @@ function SaverHandler(options) {
 	}
 	// Install the save action handlers
 	if($tw.browser) {
-		$tw.rootWidget.addEventListener("tm-save-wiki",function(event) {
-			self.saveWiki({
+		$tw.rootWidget.addEventListener("tm-save-wiki",async function(event) {
```
Jermolene, line 96: As noted, we cannot use async/await.

core/modules/saver-handler.js
```diff
@@ -125,15 +125,22 @@ SaverHandler.prototype.initSavers = function(moduleType) {
 	var self = this;
 	$tw.modules.forEachModuleOfType(moduleType,function(title,module) {
 		if(module.canSave(self)) {
-			self.savers.push(module.create(self.wiki));
+			self.savers.push({title:title,module:module.create(self.wiki)});
```
Jermolene, line 128: This change to the "savers" property isn't backwards compatible, is there another way to approach this?

core/modules/saver-handler.js
```diff
@@ -172,18 +179,91 @@ SaverHandler.prototype.saveWiki = function(options) {
...
-	}
-	return false;
+        }
+      }
+    };
+  // Process preferred if any
+  var ignorePreferred = null
+  var preferredSaver = $tw.wiki.getTiddler("$:/config/PreferredSaver")
+  if (preferredSaver !== null && preferredSaver !== undefined) {
+    var title = preferredSaver.getFieldString("text")
```
Jermolene, line 189: We use 4-character tabs not spaces to indent code

core/modules/saver-handler.js
```diff
@@ -172,18 +179,91 @@ SaverHandler.prototype.saveWiki = function(options) {
...
+    title =
+      title === undefined || title == null || title.trim() === ''
+        ? null
+        : title.trim()
+    if (title !== null) {
+      ignorePreferred = title
+      // Process preferred saver
+      if (
+        await this.save(
+          this.getSaver(title).module,
```
Jermolene, line 199: Does this code cope with there being no saver with the specified title?

core/wiki/config/PreferredSaver.tid
```diff
@@ -0,0 +1,3 @@
+title: $:/config/PreferredSaver
+
+$:/core/modules/savers/download.js
```
Jermolene, line 3: It would be better to use the `info.name` property that is already provided by savers.

Jermolene (commented):
Thanks @xmaysonnave and apologies for the delay. Some feedback below.

Jermolene:
> So this code needs to be moved into a plugin. ... I didn't try the code yet, but imo it should be relatively easy to make it a "core plugin".

@pmario I'm not sure what you're referring to here: do you mean that this submission from @xmaysonnave could be made into a plugin?

pmario:
Yes, Since it contains `async await` ES2016 code, it can't be part of the core. ... but having it as a plugin wouldn't be a big problem. Since users can decide if they want to use it. 

ES 2016 is [widely used (93%)](https://caniuse.com/#search=ES2016) by browsers now. 

ES5 is [almost 99%](https://caniuse.com/#search=ES5) ... which is good for the core and core plugins.

and

It would be easier to test as a plugin. We wouldn't need to clone the branch to test it.

core/modules/saver-handler.js
xmaysonnave: yes

core/wiki/config/PreferredSaver.tid
xmaysonnave: The saver title is used as a key to locate the appropriate saver.
The benefit of using the title is when displayed the $:/config/PreferredSaver provides a link for free to let users to access its content. If we provide the info.name if will be hard to locate the preferred saver.

core/modules/filters/savers.js
xmaysonnave: General Comment below

core/modules/filters/savers.js
xmaysonnave: General Comment below

core/modules/filters/savers.js
xmaysonnave: General Comment below

core/modules/filters/savers.js
xmaysonnave: General Comment below

core/modules/filters/savers.js
xmaysonnave: General Comment below

core/modules/saver-handler.js
xmaysonnave: General Comment below

core/modules/saver-handler.js
xmaysonnave: If we can iterate the savers module array to either locate the saver title of info.name it could be avoided.

core/modules/saver-handler.js
xmaysonnave: General Comment below

xmaysonnave:
General Comment

1 - About the formatting remarks. 
It could be nice to have a standard TiddlyWiki prettier definition preferences used either at build time and git hooked when suggesting commit and PR. I made this setup on my project (eslint and prettier) and I let the prettier-standard tooling to do the job.

2 - I agree with Mario that it could be a plugin if you don't want to introduce async/await at the core level.

core/wiki/config/PreferredSaver.tid
Jermolene: Hi @xmaysonnave that's true. We could use the lingo mechanism to have strings like `$:/config/savers/<savername>/Description` that are displayed when choosing a saver.

Jermolene:
Hi @xmaysonnave 

> It could be nice to have a standard TiddlyWiki prettier definition preferences used either at build time and git hooked when suggesting commit and PR. I made this setup on my project (eslint and prettier) and I let the prettier-standard tooling to do the job.

That would be terrific, I'd definitely be interested. There is a ticket elsewhere discussing adopting "standard.js" but I think it's too late for that; we should be content with codifying the existing standards.

> 2 - I agree with Mario that it could be a plugin if you don't want to introduce async/await at the core level.

Could it be rewritten to avoid async/await? And indeed would it be feasible to make this be a plugin?

pmario:
@xmaysonnave .. I did check the async functions and it seems you missed to handle: https://github.com/Jermolene/TiddlyWiki5/blob/b2138b3a5851530f13e6eeab40e61f674797dff7/core/modules/saver-handler.js#L60 and https://github.com/Jermolene/TiddlyWiki5/blob/b2138b3a5851530f13e6eeab40e61f674797dff7/core/modules/saver-handler.js#L74

... I didn't test the code, so I'm not 100% sure.

xmaysonnave:
> @xmaysonnave .. I did check the async functions and it seems you missed to handle:
> 
> https://github.com/Jermolene/TiddlyWiki5/blob/b2138b3a5851530f13e6eeab40e61f674797dff7/core/modules/saver-handler.js#L60
> 
> and
> https://github.com/Jermolene/TiddlyWiki5/blob/b2138b3a5851530f13e6eeab40e61f674797dff7/core/modules/saver-handler.js#L74
> 
> ... I didn't test the code, so I'm not 100% sure.

@pmario, you're right.

xmaysonnave:
> Hi @xmaysonnave
> 
> > It could be nice to have a standard TiddlyWiki prettier definition preferences used either at build time and git hooked when suggesting commit and PR. I made this setup on my project (eslint and prettier) and I let the prettier-standard tooling to do the job.
> 
> That would be terrific, I'd definitely be interested. There is a ticket elsewhere discussing adopting "standard.js" but I think it's too late for that; we should be content with codifying the existing standards.
> 
> > 2 - I agree with Mario that it could be a plugin if you don't want to introduce async/await at the core level.
> 
> Could it be rewritten to avoid async/await? And indeed would it be feasible to make this be a plugin?

1 - Considering the time you take to comment formatting issues (almost 11/13) It's worth to consider a tool. Personally I use prettier-standard who has the benefit to bundle (eslint, prettier and standard) either used at build time and at commit time. However standard increase the overall size.
On top of that I've also added other tools to improve the overall quality of this project:
- https://snyk.io/ (vulnerabilities)
- https://app.fossa.com/ (licensing)

2  - I can definitely do a plugin however I will keep the async/await. Let me know.

core/wiki/config/PreferredSaver.tid
xmaysonnave: Not sure to understand what you think.
Right now I do the following :

```
\define preferred-saver()
<$text text=<<infosaver "$(currentTiddler)$">>/>
\end

<$select tiddler="$:/config/PreferredSaver">
<$list filter="[savers[]sort[]]">
<option value=<<currentTiddler>>><<preferred-saver>></option>
</$list>
</$select>
```

```
/*
   * Information about this macro
   */
  exports.name = 'infosaver'

  exports.params = [{ name: 'tiddler' }]

  /*
   * Run the macro
   */
  exports.run = function (tiddler) {
    return $tw.saverHandler.getSaver(tiddler).module.info.name
  }

```

core/modules/saver-handler.js
xmaysonnave: Could you elaborate a little more about what you mean with 'backward compatible' ? It sound to me that this data structure is internal.

Jermolene:
Hi @xmaysonnave apologies, I've been slow due illness recently, but hopefully getting back up to speed.

1. I'm keen to improve the tooling around the project (e.g. just recently we improved the CI implementation and moved it to Google Actions). There's a lot more to be done but contributions are welcome.

2. I think the async/await issue settles this as needing to go into a plugin. Let me know if there are additional hooks etc. that you need

xmaysonnave:
Thanks for the informations.
I agree that it could be plugin to preserve the original saver-handler without async/await.

PR #4719, opened 2020-06-19
Allow toggling relink in tm-rename-tiddler

saqimtiaz:
This change allows the user to determine whether the `tm-rename-tiddler` message will trigger renaming in tags and list fields of other tiddlers, while staying backwards compatible.

The current behaviour is that relinking always happens, which is not always desirable.

Two extra parameters are supported in the `tm-rename-tiddler` message, which correspond to options attributes already supported in `wiki.relinkTiddler`:
- [dontRenameInTags](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/wiki-bulkops.js#L50)
- [dontRenameInLists](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/wiki-bulkops.js#L60)

This would be used as follows:
`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> dontRenameInTags="true" dontRenameInLists="true"/>`

or if the user wants the relink behaviour to abide by the $:/config/RelinkOnRename setting:
`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> dontRenameInTags={{$:/config/RelinkOnRename}} dontRenameInLists={{$:/config/RelinkOnRename}}/>`

If there is consensus on this change I will update docs to match.

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -609,10 +609,13 @@ NavigatorWidget.prototype.handleUnfoldAllTiddlersEvent = function(event) {
 };
 
 NavigatorWidget.prototype.handleRenameTiddlerEvent = function(event) {
-	var paramObject = event.paramObject || {},
+	var options = {},
+		paramObject = event.paramObject || {},
 		from = paramObject.from || event.tiddlerTitle,
 		to = paramObject.to;
-	this.wiki.renameTiddler(from,to);
+	options.dontRenameInTags = (paramObject.dontRenameInTags === "true" || paramObject.dontRenameInTags === "yes") ? true : false;
+	options.dontRenameInLists = (paramObject.dontRenameInLists === "true" || paramObject.dontRenameInLists === "yes") ? true : false;
+	this.wiki.renameTiddler(from,to,options);
 };
 
 exports.navigator = NavigatorWidget;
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz that's great, please could you also update the docs for "WidgetMessage: tm-rename-tiddler"?

saqimtiaz:
@Jermolene docs updated.

Jermolene:
Great thanks @saqimtiaz

saqimtiaz:
I've realized there is a minor issue with the merged pull request. While the code is correct and works as expected, the third example provided will not work.

`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> dontRenameInTags={{$:/config/RelinkOnRename}} dontRenameInLists={{$:/config/RelinkOnRename}}/>`

This wont work as expected since `$:/config/RelinkOnRename` is `yes` when we want to allow renaming, but that corresponds to `dontRenameInTags` and `dontRenameInLists` needing to be the opposite: no.

While this isn't necessary for the added functionality, it is a nice to have option that will make life easy for users. I have revised the code and will submit a revised pull request after I have completed my testing.

saqimtiaz:
I have tested the revised code and a new pull request will be made once the first one has been reverted.

---

For those curious about the revision, the parameters `renameinTags` and `renameInLists` are replaced with `dontRenameInTags` and `dontRenameInLists` respectively. This avoids users having to think through double negatives, as well as corresponds better to the setting in `$:/config/RelinkOnRename`

Revised code is here:
https://github.com/saqimtiaz/TiddlyWiki5/commit/48511356f0ce3e146080cf229deea73a58eb9b82

PR #4722, opened 2020-06-19
Revert "Allow toggling relink in tm-rename-tiddler"

saqimtiaz:
Reverts Jermolene/TiddlyWiki5#4719

Revised pull request will follow.

I've realized there is a minor issue with the merged pull request. While the code is correct and works as expected, the third example provided will not work.

`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> dontRenameInTags={{$:/config/RelinkOnRename}} dontRenameInLists={{$:/config/RelinkOnRename}}/>`

This wont work as expected since `$:/config/RelinkOnRename` is `yes` when we want to allow renaming, but that corresponds to `dontRenameInTags` and `dontRenameInLists` needing to be the opposite: `no`.

While this isn't necessary for the added functionality, it is a nice to have option that will make life easy for users. I have revised the code and will submit a revised pull request after I have completed my testing.

---

For those curious about the revision, the parameters `renameinTags` and `renameInLists` are replaced with `dontRenameInTags` and `dontRenameInLists` respectively. This avoids users having to think through double negatives, as well as corresponds better to the setting in `$:/config/RelinkOnRename`

Revised code is here:
https://github.com/saqimtiaz/TiddlyWiki5/commit/48511356f0ce3e146080cf229deea73a58eb9b82

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -609,13 +609,10 @@ NavigatorWidget.prototype.handleUnfoldAllTiddlersEvent = function(event) {
 };
 
 NavigatorWidget.prototype.handleRenameTiddlerEvent = function(event) {
-	var options = {},
-		paramObject = event.paramObject || {},
+	var paramObject = event.paramObject || {},
 		from = paramObject.from || event.tiddlerTitle,
 		to = paramObject.to;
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz apologies for the delay

PR #4723, opened 2020-06-19
Revised: toggling relink in tm-rename-tiddler

saqimtiaz:
This is a revised version of #4719, which I have thoroughly tested.

The parameters `renameInTags` and `renameInLists` replace `dontRenameInTags` and `dontRenameInLists` from the previous PR respectively. This avoids users having to think through double negatives, as well as corresponds better to the setting in `$:/config/RelinkOnRename`

---

Two extra parameters are supported in the `tm-rename-tiddler` message, which inversely correspond to options attributes already supported in `wiki.relinkTiddler`:

- `renameInTags`
- `renameInLists`

This would be used as follows:
`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> renameInTags="no" renameInLists="no"/>`

or if the user wants the relink behaviour to abide by the $:/config/RelinkOnRename setting:
`<$action-sendmessage $message="tm-rename-tiddler" from=<<fromTitle>> to=<<toTitle>> renameInTags={{$:/config/RelinkOnRename}} renameInLists={{$:/config/RelinkOnRename}}/>`

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -609,10 +609,13 @@ NavigatorWidget.prototype.handleUnfoldAllTiddlersEvent = function(event) {
 };
 
 NavigatorWidget.prototype.handleRenameTiddlerEvent = function(event) {
-	var paramObject = event.paramObject || {},
+	var options = {},
+		paramObject = event.paramObject || {},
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz

Issue #2889, opened 2017-06-10
LinkWidget->handleClickEvent should send original event keypress data


saqimtiaz:
[LinkWidget.prototype.handleClickEvent](https://github.com/Jermolene/TiddlyWiki5/blob/8f1114960a20ba3d84dd1633ad1976db9461a41c/core/modules/widgets/link.js#L128) is currently implemented such that it does not pass along any of the original event keypress data (ctrlKey, metaKey, shiftKey, altKey) in the tm-navigate event that is dispatched.  

 ```
this.dispatchEvent({
		type: "tm-navigate",
		navigateTo: this.to,
		navigateFromTitle: this.getVariable("storyTiddler"),
		navigateFromNode: this,
		navigateFromClientRect: { top: bounds.top, left: bounds.left, width: bounds.width, right: bounds.right, bottom: bounds.bottom, height: bounds.height
		},
		navigateSuppressNavigation: event.metaKey || event.ctrlKey || (event.button === 1)
	});
```

For a plugin writer seeking to add on functionality or modify the existing behavior based on modifier keys used when clicking a tiddler link, including this keypress data is essential. 

The handler for this event ([Navigator.prototype.handleNavigationEvent](https://github.com/Jermolene/TiddlyWiki5/blob/eba1c3c160c63ac3dfc15f55df086971a832d657/core/modules/widgets/navigator.js#L174)) provides a hook 'th-navigating' that can be used to modify navigation behavior on clicking a link. However without the event keypress data, adding extra functionality based on say pressing Shift or Alt when clicking a link is not possible.

It should also be considered to provide the ctrlKey and metaKey data, in addition to Shift and Alt - in the tm-navigate event dispatched - in case it is desirable to override core suppressNavigation behavior and use those keys for a different purpose. Example:
 ```
this.dispatchEvent({
		type: "tm-navigate",
		navigateTo: this.to,
		navigateFromTitle: this.getVariable("storyTiddler"),
		navigateFromNode: this,
		navigateFromClientRect: { top: bounds.top, left: bounds.left, width: bounds.width, right: bounds.right, bottom: bounds.bottom, height: bounds.height
		},
		navigateSuppressNavigation: event.metaKey || event.ctrlKey || (event.button === 1),
		metaKey: event.metaKey,
		ctrlKey: event.ctrlKey,
		altKey: event.altKey,
		shiftKey: event.shiftKey
	});
```

danielo515:
[...]

PR #4724, opened 2020-06-21
Link widget: Include original event in tm-navigate

saqimtiaz:
`tm-navigate` messages sent by the `Link` widget do not include the original `event`, whereas [when sent by the Button widget](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/button.js#L151) they do.

This change is needed to have a consistent way to detect modifier keys used when a link or button is clicked, as well as any other properties of interest from the original event.

This also addresses the inconsistency introduced in #2889, when we added modifier keys to the message instead of including the original `event`, which was the established pattern from the Button widget. The change introduced in #2889 is not retracted in the interest of backwards compatibility.

### DIFF
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -158,7 +158,8 @@ LinkWidget.prototype.handleClickEvent = function(event) {
 		metaKey: event.metaKey,
 		ctrlKey: event.ctrlKey,
 		altKey: event.altKey,
-		shiftKey: event.shiftKey
+		shiftKey: event.shiftKey,
+		event: event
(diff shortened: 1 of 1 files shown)


pmario:
+1 for this. Since the button widget does provide the info, it should be implemented for consitency reasons. ... IMO that should be enough reasoning ;) 

For generic "multi-story" support it also needs a "context-object" .. but that's a completely different thing.

Jermolene:
Thanks @saqimtiaz @pmario apologies for the delay.

PR #4727, opened 2020-06-24
Signing the CLA

jdangerx:

PR #4728, opened 2020-06-24
Clarifying plugin installation for Node.js

jdangerx:
I was having trouble installing plugins for Node.js - turns out I had been trying to include them in the `tiddlywiki.info`, which caused 

```
Warning: Cannot find plugin 'flibbles/relink'
```

Many thanks to @matrixbot on the Gitter for explaining what was going on and how to fix it.

### DIFF
--- a/editions/tw5.com/tiddlers/nodejs/Installing custom plugins on Node.js.tid
+++ b/editions/tw5.com/tiddlers/nodejs/Installing custom plugins on Node.js.tid
@@ -7,7 +7,19 @@ type: text/vnd.tiddlywiki
 There are several ways in which custom plugins that are not part of TiddlyWiki's plugin library can be installed when using TiddlyWiki under Node.js. (See [[Installing a plugin from the plugin library]] for instructions on installing plugins from the library).
 
 * Arrange the PluginFolders containing the plugins in a convenient shared location and then use [[environment variables|Environment Variables on Node.js]] to tell TiddlyWiki to search those folders. The plugins can be referenced in `tiddlywiki.info` by their name (e.g. `tiddlytools/magic`)
-* Place the PluginFolders containing the plugins in a `plugins` folder within the [[wiki folder|TiddlyWikiFolders]]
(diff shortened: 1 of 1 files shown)


ibnishak:
I think this should go like:

There are three ways to include plugins in your node js based TiddlyWiki

1. Tiddlywiki.info .....
2. Keep them at a convenient location and set the environment variable ...m
3. Plugin folders within your wiki folder

joshuafontany:
#2 requires listing them in #1 (it does allow you to keep plugins outside of the TW repo).

pmario:
That's absolutely __not__ the way, how to do it!

pmario:
You need to set the environment variables in the right way and then include the plugin into `tiddlywiki.info`. That's the _only_ way it should happen. Other ways you will only have trouble in the future.

pmario:
I think your paths where wrong, that's why you had the problem from the OP. 

eg: `TIDDLYWIKI_PLUGIN_PATH` may be `C:\other\plugins`

and `C:\other\plugins\flibbles\relink` contains the relink code
and `C:\other\plugins\flibbles\relink-markdown` the relink md plugin.

Then your `tiddlywiki.info` needs to look that way: 

```
{
	"plugins": [
		"tiddlywiki/slider",
		"flibbles/relink"
		"flibbles/relink-markdown"
	]
}
```

Working with those settings allows you to add other 3rd party plugins in the exact same way, using the same `c:\other\plugins` path, with a different namespace.

Jermolene:
Thanks @jdangerx. Notwithstanding @joshuafontany and @pmario's tanget, I think this is good to go: it clarifies an existing bullet point with an example, and seems helpful.

However, I'm afraid there is a tiny bit of bureaucracy: please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)). Many thanks!

jdangerx:
Great to hear it @Jermolene ! [Here's the PR to the CLA.](https://github.com/Jermolene/TiddlyWiki5/pull/4727)

Jermolene:
Thanks @jdangerx and apologies for the delay.

PR #4730, opened 2020-06-25
Updates to Getting started, Saving & reclassification of savers

ibnishak:
As discussed in https://github.com/Jermolene/TiddlyWiki5/pull/4640

Demo at: https://ibnishak.github.io/TiddlyWiki5/

### DIFF
--- a/editions/tw5.com/tiddlers/community/resources/Noteself by Danielo Rodriguez.tid
+++ b/editions/tw5.com/tiddlers/community/resources/Noteself by Danielo Rodriguez.tid
@@ -5,9 +5,9 @@ created: 20141122093837330
 delivery: Web Service
 description: Free online service that you can also host yourself
 method: sync
-modified: 20200507203026367
+modified: 20200625180613379
 tags: Android Chrome Firefox [[Internet Explorer]] Linux Mac Opera PHP Resources Safari Saving Windows iOS Edge
-title: "Noteself" by Danielo Rodríguez
+title: Noteself by Danielo Rodríguez
 type: text/vnd.tiddlywiki
 url: https://noteself.github.io/
 
--- a/editions/tw5.com/tiddlers/community/resources/TiddlyDrive Add-on for Google Drive by Joshua Stubbs.tid
+++ b/editions/tw5.com/tiddlers/community/resources/TiddlyDrive Add-on for Google Drive by Joshua Stubbs.tid
@@ -1,12 +1,12 @@
 caption: ~TiddlyDrive
-color: #FFEB3B
+color: #29B6F6
 community-author: Joshua Stubbs
 created: 20171109172705241
-delivery: Google Drive Add-on
+delivery: Web Service
 description: Google Drive add-on to save TiddlyWiki files
 method: save
-modified: 20200507202706013
-tags: Android Chrome Firefox [[Internet Explorer]] Linux Mac Opera PHP Resources Safari Saving Windows iOS plugins Edge
+modified: 20200625180613342
+tags: Resources Saving
 title: TiddlyDrive Add-on for Google Drive by Joshua Stubbs
 type: text/vnd.tiddlywiki
 url: https://lordratte.gitlab.io/tiddlydrive/#installation-guide
--- a/editions/tw5.com/tiddlers/community/resources/TiddlyServer by Arlen Beiler.tid
+++ b/editions/tw5.com/tiddlers/community/resources/TiddlyServer by Arlen Beiler.tid
@@ -2,10 +2,10 @@ caption: ~TiddlyServer
 color: #f06292
 community-author: Arlen Beiler
 created: 20171109171415540
-delivery: DIY
+delivery: Node Server
 description: An extension to TiddlyWiki's Node.js server
 method: sync
-modified: 20200507203417109
+modified: 20200625180929623
 tags: Linux Mac Windows Saving Resources
 title: TiddlyServer by Arlen Beiler
 type: text/vnd.tiddlywiki
--- a/editions/tw5.com/tiddlers/community/resources/Timimi Extension and executable by Riz.tid
+++ b/editions/tw5.com/tiddlers/community/resources/Timimi Extension and executable by Riz.tid
@@ -2,10 +2,10 @@ caption: Timimi
 color: #4db6ac
 community-author: Riz
 created: 20180830194141190
-delivery: Browser Extension & Native host
-description: Browser extension & native host for desktops
+delivery: Browser Extension
+description: Browser extension for Chrome and Firefox.
 method: save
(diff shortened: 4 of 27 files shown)


pmario:
Hi, ... We need to deprecate the existing Beaker Browser. ... The TW saver won't work for BB V 1.0

pmario:
hmmm, There is no information anymore, which savers work for Android or other OSes. Eg. File-backups works with FF for Android. 
I still like the filter options.

pmario:
TiddlyFox needs to be deprecated too!

ibnishak:
One of the inputs I received regarding last model was most filters are pointless as they show most options regardless. I felt that was a valid criticism. In addition, filters were rather confusing on mobile screens, where a bird eye view of every thing is not visible and it is hard to see the changes.

 I have updated the descriptions of platform specific savers. As you can see, in desktop application - each of them mentions specific platforms. Browsers extension mentions specific browsers they support. 
After that only IIS and HTA hack are platform bound - which is again clearly mentioned. This way it is less confusing.

ibnishak:
If necessary, more detailed platform details can be added to individual saver tiddlers.

ibnishak:
By deprecated - should they be taken out of the matrix? Tiddlyfox is of interest to people who depend on old Firefox and waterfox like browsers I thought.

ibnishak:
Also please reconsider the status of HTA hack and PHP server too. While HTA hack technically still works, it is no more advised to use them due to security reasons - similar in principle to depreciating Tiddlyfox but with the added reason that users of old IE editions have alternatives in TiddlyIE. PHP server because its instructions are outdated I gather. Fewer well tested and updated things are better than a plethora of options with potholes.

pmario:
>I have updated the descriptions of platform specific savers. As you can see, in desktop application - each of them mentions specific platforms. Browsers extension mentions specific browsers they support.
After that only IIS and HTA hack are platform bound - which is again clearly mentioned. This way it is less confusing.

That's nice, but nobody reads the description. Text doesn't have enough visual "differences" to help make decisions. ... The number of elements causes the confusion. 

If we would add icons for OSes, Browsers and so on to the "cards" imo it would be much easier for users to decide. Because they can see their preferences immediately.

ibnishak:
I am up for replacing text with icons. I can pack that and removal of outdated savers to another PR

Jermolene:
Thank you @ibnishak for your continued work on this. I'd especially praise the visual design, which is up to your usual standard - if you see scope for other visual improvements to the core please don't hesitate to suggest them.

The demo is really helpful. It does however highlight that there's a fair way to go, which makes me wonder if we might step back and consider whether it's worthwhile to invest your time in it at this point. The fundamental trouble is that because we don't have all the answers up front, this looks like it will require several cycles of iteration where we experiment, hypothesise improvements, and perhaps test with users. That kind of investigative development can be frustrating, because it is time consuming and there's an inevitable risk that individual bits of work may end up being thrown away. Meanwhile, there are of course many other areas you might focus on that are better defined.

With that proviso, if you did want to keep going, I'll give my feedback below :smile:

Firstly, I wonder if it might be helpful to work with a real flow chart so that we can consider the questions and pathways that we need, without getting distracted by the details of the implementation as a wizard. That way we might be able to get a sense of whether the approach is helpful without investing too much time in the implementation.

The implementation of the wizard seems the easy and fun part. I think that many pathways would require multiple steps in the wizard, and perhaps a way to store information in one step (e.g. choice of operating system) and then access it in a subsequent step.  I was imagining a presentation where we show a single question at a time, with a back button to go back to the previous question, and a next button to proceed to the next step.

ibnishak:
I would definitely like to contribute further. I understand that the final iteration of the wizard part will take time and several round of trials to come together. However it is high time this part is correctly tackled by the community to help better user onboarding.

The flow of wizard - various questions, possible answers and pathways - definitely require user testing and a better consensus. The focus group for testing such a change ought to include new users rather than veterans. Unfortunately, I don't have access to that sort of user base. I am definitely up for changes to wizard steps if sound suggestions arise. The current discussion thread in the Google groups seems little derailed from that purpose with focus shifted to relevance of individual savers and the implementation details. Any other suggestions to raise meaningful user inputs?

## Regarding the implementation

The  wizard is built in such a way that the current step is stored in a state tiddler($:/state/GettingStarted) and answers to each question is stored in separate state tiddlers. So answers to each step will be available throughout the wizard. 

The next and previous steps are bound to respective buttons on each step. There is no "list" where order of wizard is stored. I have points for and against this approach.

### For
- It is simple to build and understand if anyone down the lane wants to make changes. 
- It is flexible and will enable branching pathways.

### Against
- It is not scalable.
- Official documentation should be painted on the insides too. If a user want to implement a wizard in his wiki, this would be the wizard he might refer. So it should showcase better practices. The ideal wizard in my imagination is a variation of the toc macro, taking a root tag, sort order and a template.

## Non Wizard parts of PR

1. The filters are removed from the single HTML file flavor matrix. As mentioned above, this was in response to a feedback that most filters are pointless and doesn't significantly narrow options. The alternative approach I have followed is mentioning the platforms wherever it is relevant. If I find proper platform SVGs with appropriate license, I would like to indicate the platforms using them. 
2. The tip/warning/note macros. The SVGs are replaced with respective headings. This is done with accessibility for those who are vision impaired and depend on screen readers in mind. I would to like to suggest the rest of tip/warning/preamble macros to be converted to this.
3. @AnthonyMuscio pointed out that there ought to be a readily available download empty button somewhere. Menu bar would be a good location for catering this to established users.

pmario:
Demo link is broken. Internet archive [link is from 8. Nov. 2020](https://web.archive.org/web/20201108123458/https://ibnishak.github.io/TiddlyWiki5/#Saving:Saving)

@Jermolene PR can be closed, since it can't be merged anymore. PR branch is gone!

Jermolene:
> @Jermolene PR can be closed, since it can't be merged anymore. PR branch is gone!

Interestingly, it looks like the PR can still be merged. @ibnishak was working on this while I was ill in the summer and so I wasn't able to engage with much feedback. I'd certainly be interested to see this investigation continue.

linonetwo:
This wizard is very nice, I like this "help me choose".

!\[\](https://user-images.githubusercontent.com/3746270/167230520-11ee1601-292d-4dc5-bcec-5f8bd45a6661.png)

linonetwo:
@ibnishak will you still resolve conflict for this?

AnthonyMuscio:
Can we use the Info mechanism and and Browser Sniffer to prepopulate the saver wizard?

Of course a user may be seeking instructions for a different OS/Browser than the one they are on but especially new users could start with what they have.
* Based on your current browser (and any other info we can glean)
* Other Browser and Operating system options

PR #4739, opened 2020-06-26
Update Filter Operators.tid

twMat:


Jermolene:
Thanks @twMat for all of these!

PR #4741, opened 2020-06-26
Extend tm-import-tiddlers message

saqimtiaz:
This PR extends the `tm-import-tiddlers` message and the `dropzone` widget to support two extra attributes:

- `importTitle` : which allows to use a different tiddler instead of `$:/Import` for the import process. Note that `tm-perform-import` already allows to [derive the title of the import tiddler from the event](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/navigator.js#L550).
- `autoOpenOnImport` : which allows a `dropzone` widget to override the global setting in `tv-auto-open-on-import`. 

Combined together both of these changes allow greater flexibility in implementing new UI and workflow for the import process. These changes are backwards compatible, the new attributes are optional.


Jermolene:
Great stuff, thank you @saqimtiaz

PR #4754, opened 2020-07-14
Add requirements to build tiddlywiki executable

inmysocks:
This adds the needed dev dependency and scripts to build a single file executable for tiddlywiki.

All the core editions, languages, plugins and themes are packaged in it and it is used in the same way as using npm to install tiddlywiki but you don't need node or npm to use it.

One thing to note is that it won't be globally available unless you place it somewhere included in your system $PATH

Building executables like this and putting them up as part of each release will probably make it much easier for people to use the node version.

### DIFF
--- a/bin/nexe-build-all.js
+++ b/bin/nexe-build-all.js
@@ -1,0 +1,47 @@
+#!/usr/bin/env node
+
+const { compile } = require('nexe')
+
+
+console.log('Building Executable for Linux x64')
+compile({
+  input: './tiddlywiki.js',
+  output: 'tiddlywiki',
+  targets: 'linux-x64',
+  resources: ['./plugins/**/*', './themes/**/*', './editions/**/*', './languages/**/*']
+}).then(() => {
+  console.log('Building Executable for Linux x32')
+  return compile({
+    input: './tiddlywiki.js',
+    output: 'tiddlywiki',
+    targets: 'linux-x32',
+    resources: ['./plugins/**/*', './themes/**/*', './editions/**/*', './languages/**/*']
+  })
+}).then(() => {
+  console.log('Building Executable for OSX x64')
+  return compile({
+    input: './tiddlywiki.js',
+    output: 'tiddlywiki.command',
+    targets: 'mac-x64',
+    resources: ['./plugins/**/*', './themes/**/*', './editions/**/*', './languages/**/*']
+  })
+}).then(() => {
+  console.log('Building Executable for Windows x64')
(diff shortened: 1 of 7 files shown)


pmario:
I'm not happy with such an executable. ... It's not signed. At least I don't see any code for an official signing process. 

If you install nodejs, you get the Installation "Warning" dialogue, that shows you, that the setup exe hasn't been modified and the you install something as intended by the ORIGINAL authors. 

`npm install tiddlywiki` makes sure, that you install tidlywiki as Jeremy "compiled" and uploaded it to npm. ... 

So there is a "very little bit of confidence" that you installed what YOU the user intended. ...

If you install an unsigned exe, for me it's the same as if you "shoot in your foot". There is absolutely _no_ way for users to verify what they do. ... Absolutely no confidence what you execute. 

---------

I personally do want to have a plugin signing and checking mechanism, that will allow users to have better confidence about plugins they use. ... IMO that would be worth working on. 

just my thoughts.

Jermolene:
Hi @inmysocks Nexe is interesting, and I see it would be useful in some situations. However I'm not sure that we should make it one of the official distributions of TiddlyWiki because it adds yet another potentially confusing choice for new users, and would have implications across the documentation. We could still add it to the core repo (in the same way that the core includes the means to make external JS builds of the core, but it's not a configuration that we directly support).

By the way, I'd have expected the virtual file system to include package.json for reading the TW version number.

linonetwo:
I think this can be in a separate repo, just like https://github.com/tiddly-gittly/TidGi-Desktop .

And this executable is similar to Tidgi, because Tidgi is just an wrapper around tiddlywiki too.

linonetwo:
> make it much easier for people to use the node version.

I think TidGi already achieve this!

Jermolene:
Thanks @linonetwo @inmysocks. I'm keen to improve TiddlyWiki's downloadable native app options. It is complicated for end users if we have too many options, so I think that to really move things along we need a platform that gives equivalent or improved functionality to TiddlyDesktop. Roughly:

* Cross platform GUI
* Deep integration with host OS for global keyboard shortcuts etc
* Mobile compatibility
* Compatibility with app stores
* Easy to build a cross platform installer

Nexe seems useful enough, and is still being maintained, but is quite narrow in focus and I don't think satisfies enough of our goals to justify the investment in switching.

inmysocks:
I agree. For my day job I have been working with electron for cross-platform applications and something like that would be a nicer solution.

PR #4757, opened 2020-07-14
Fix #4755 - Indirect colours not shown in palette switcher

### LINKED ISSUE #4755: [BUG] Palette switcher preview doesn't work with indirect colours
The preview swatches in the palette switcher in control panel do not correctly show colours that are defined indirectly via the `<<colour>>` macro

2. Click on the cog icon to open the control panel, and choose the "Appearance" and then "Palette" tabs
3. Scroll down to the "Show Editor" button and click it
4. Locate the entry for "General Background"/"background" (which should currently be set to `#ffffff`)
5. Change it to `<<colour alert-background>>`
6. Note that the background colour of tiddlers changes to yellow, and 
7. The error is that the preview swatch should show the background colour as the second box, but it doesn't show the correct yellow colour

image

This behaviour is observed on all platforms.

BurningTreeC:


joshuafontany:
Nice fix!

BurningTreeC:
There's probably a better fix - I'm investigating

Jermolene:
Thanks @BurningTreeC, much appreciated.

PR #4758, opened 2020-07-15
Add modifier variable to linkcatcher actionstring

BurningTreeC:
This PR adds the **modifier** variable to the actionstring of the linkcatcher widget, like we're doing it within the `button` and `droppable` actionstrings


BurningTreeC:
For discussions I made this a draft PR

saqimtiaz:
@BurningTreeC thank you. I think this is good to go.

@pmario would this is anyway clash with future enhancements that you have in mind?

Jermolene:
Terrific, thanks @BurningTreeC

PR #4763, opened 2020-07-21
Update fixed Operator (Examples).tid

twMat:
To clarify it pads with zeros.


Jermolene:
Thanks @twMat apologies for the delay

PR #4774, opened 2020-07-27
Fix markdown table to honor alignment directives

### LINKED ISSUE #4698: [BUG] Markdown table no center and right-alignment
**Describe the bug**
The markdown plugin does not render tables correctly when there are centered or right-aligned columns. Columns always align to the left.

**To Reproduce**
1. Go to https://tiddlywiki.com/plugins/tiddlywiki/markdown/
2. Click on 'MarkdownExample'
3. Add table with center or right alignment such as 

```
| Tables   |      Are      |  Cool |
|----------|:-------------:|------:|
| col 1 is | left-aligned | $1600 |
| col 2 is | centered in the middle |   $12 |
| col 3 is | right-aligned |    $1 |
```
4. See the table rendered. Here is a screenshot:

Screenshot from 2020-06-04 05-50-22

**Expected behavior**
| Tables   |      Are      |  Cool |
|----------|:-------------:|------:|
| col 1 is | left-aligned | $1600 |
| col 2 is | centered in the middle |   $12 |
| col 3 is | right-aligned |    $1 |

**Desktop**
 - OS: Ubuntu 18.10
 - Browser [brave]

ento:
Fixes #4698

How it looks like:

image

This PR replicates how Remarkable renders tables in the Markdown plugin:
https://github.com/jonschlinkert/remarkable/blob/v2.0.1/lib/rules.js#L197-L211

### DIFF
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -69,7 +69,7 @@ function findTagWithType(nodes, startPoint, type, level) {
 function convertNodes(remarkableTree, isStartOfInline) {
 	let out = [];
 	var accumulatedText = '';
-	function wrappedElement(elementTag, currentIndex, currentLevel, closingType, nodes) {
+	function withChildren(currentIndex, currentLevel, closingType, nodes, callback) {
 		var j = findTagWithType(nodes, currentIndex + 1, closingType, currentLevel);
 		if (j === false) {
 			console.error("Failed to find a " + closingType + " node after position " + currentIndex);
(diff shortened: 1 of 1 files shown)


pmario:
@ento ... So it means remarkable.js 2.0.1 fixes the problem? ... If so, imo we should include the latest version into our plugin, instead of "working around" the problem.

ento:
@pmario I don't believe so; remarkable.js's code for rendering table cells hasn't changed since [v2.0.0](https://github.com/jonschlinkert/remarkable/blob/v2.0.0/lib/rules.js#L197-L211) (nor [v1.7.4](https://github.com/jonschlinkert/remarkable/blob/v1.7.4/lib/rules.js#L206-L220)). It wasn't working for us because the Markdown plugin takes remarkable.js's output in AST form and converts it to TiddlyWiki's AST form, and that conversion step wasn't emulating all of how remarkable.js renders/converts its AST to HTML.

* Remarkable.js: Markdown -> Remarkable AST -> HTML (the code snippets that I linked above get run at this second step)
* Markdown plugin: Markdown -> Remarkable AST -> TiddlyWiki AST

pmario:
ok. yea I did have a look at the changes in remarkable and saw that this part wasn't changed. So thx for the PR.

Jermolene:
Thanks @ento @pmario

PR #4780, opened 2020-07-30
Fix authenticated-user-header is case sensitive in header authenticator

favadi:
From RFC 2616 - "Hypertext Transfer Protocol -- HTTP/1.1", Section 4.2, "Message Headers":

Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive.


Jermolene:
Thanks @favadi

Issue #4784, opened 2020-07-31
[IDEA] Date handling can be improved using: Natively Format JavaScript Dates and Times


pmario:
See: https://elijahmanor.com/format-js-dates-and-times

Browser support is good: https://caniuse.com/#feat=date-tolocaledatestring

eg: 

```
const date = new Date();

console.log(date.toLocaleDateString('en-US'));
// 7/19/2020

const dateOptions = {
  weekday: 'long',
  year: 'numeric',
  month: 'long',
  day: 'numeric',
};

console.log(date.toLocaleDateString('en-US', dateOptions));
// Sunday, July 19, 2020
```

pmario:
also see: https://elijahmanor.com/format-js-numbers for numbers

PR #4785, opened 2020-07-31
Format filter to create formatted date strings

saqimtiaz:
This PR adds an optimized version of a filter I have been using for some time to allow easy creation of **formatted date strings.** While there are workarounds possible using the `view widget` and wikify, this extends the ability of filters and is often more convenient.

**Operand:** The _operand_ to the filter operator is used as date `format`, consistent with the handling in the `View widget`. If the operand is not specified it defaults to `YYYY MM DD 0hh:0mm` which is again consistent with the view widget.

**Suffix:** If the _suffix_ `relative` is used, the operand (if any) is ignored and the string is formatted as a relative date.

An invalid date string returns no result.

Examples:
`[[20200731]asdate[mmm DD]] -> Jul 31`
`[[20200730]asdate:relative[]] -> 45 hours ago`
`[[20200730]asdate:relative[mmm DD]] -> 45 hours ago`

If there is interest in adding this to the core I can work on docs as well.

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -27,6 +27,7 @@ Error/EditConflict: File changed on server
 Error/Filter: Filter error
 Error/FilterSyntax: Syntax error in filter expression
 Error/IsFilterOperator: Filter Error: Unknown operand for the 'is' filter operator
+Error/FormatFilterOperator: Filter Error: Unknown suffix for the 'format' filter operator
 Error/LoadingPluginLibrary: Error loading plugin library
 Error/NetworkErrorAlert: `<h2>''Network Error''</h2>It looks like the connection to the server has been lost. This may indicate a problem with your network connection. Please attempt to restore network connectivity before continuing.<br><br>''Any unsaved changes will be automatically synchronised when connectivity is restored''.`
 Error/RecursiveTransclusion: Recursive transclusion error in transclude widget
--- a/core/modules/filters/format.js
+++ b/core/modules/filters/format.js
@@ -1,0 +1,46 @@
+/*\
+title: $:/core/modules/filters/format.js
+type: application/javascript
+module-type: filteroperator
+Filter operator for formatting strings
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+var formatFilterOperators;
+
+function getFormatFilterOperators() {
+	if(!formatFilterOperators) {
+		formatFilterOperators = {};
+		$tw.modules.applyMethods("formatfilteroperator",formatFilterOperators);
+	}
+	return formatFilterOperators;
+}
+
+/*
+Export our filter function
+*/
+exports.format = function(source,operator,options) {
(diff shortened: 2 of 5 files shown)


pmario:
Does it allow all the possibilities as described here: https://tiddlywiki.com/#DateFormat ?

pmario:
Imo also related to: #4784

saqimtiaz:
> Does it allow all the possibilities as described here: https://tiddlywiki.com/#DateFormat ?

Yes the support is complete and the filter relies on the same code from `$tw.utils` as the `view widget`, so there is no discrepancy in the handling.

> Imo also related to: #4784

I do not think that should impact whether we want to merge this or not. The correct way to implement #4784 for backwards compatability would be to adapt the underlying methods for date handling in `$tw.utils` such as `$tw.utils.formatDateString`, in which case the filter would receive those improvements automatically.

saqimtiaz:
@Jermolene any thoughts? I don't have a vested interested in having this in the core beyond the perceived utility of such a filter to users. If we decide not to merge, I am happy to bundle this as a plugin for those that need it.

pmario:
I think it would be useful in the core. We do have a lot of requests in the GG and the existing "workaround" is highly sophisticated wikitext, which is hard to remember.

Jermolene:
Thanks @saqimtiaz this is useful. I've also got an implementation of formatDate() as a filter somewhere in my backlog. I took the approach of calling the operator "format", with "date" being one of the suffixes (e.g. `[[20200731]format:date[mmm DD]]`), but then I don't think I managed to come up with any other types that would be useful. Perhaps `format:currency[]`?

saqimtiaz:
@Jermolene apologies for the late reply, I had a bunch of things come up unexpectedly at work that are occuping my attention.

I favour your idea of a `format` operator with a suffix, it is a more versatile approach and will allow us to supports other formats as needed, without needing new operators each time.

I think `date` and `relativedate` are two good suffixes to support at first.

If you have the code for this conveniently at hand already then it would be great to get it posted/merged, otherwise I can rework this PR to implement a `format` filter with `date` and `relativedate` as supported suffixes.

Jermolene:
Hi @saqimtiaz I'm afraid I couldn't find the old code; I do usually keep deleted code but not in this case it seems. There wasn't much to it, I think $:/core/modules/filters/format.js was pretty much a copy of $:/core/modules/filters/is.js.

saqimtiaz:
@Jermolene no worries. I'll rework this PR in that direction sometime this week.

Jermolene:
Thanks @saqimtiaz

saqimtiaz:
@Jermolene @pmario PR updated with a `format` filter operator (modelled on `is` operator) and suffixes `date` and `relativedate`.

Jermolene:
Thanks @saqimtiaz -- docs would be appreciated when you get a chance.

saqimtiaz:
> Thanks @saqimtiaz -- docs would be appreciated when you get a chance.

Will do. I had just created an issue to put this on my to-do list when I saw this :)

Jermolene:
Thanks @saqimtiaz

PR #4811, opened 2020-08-13
Add suffix and parameter to trim operator

### LINKED ISSUE #4809: [IDEA] Allow trim operator to take an optional parameter and optional "prefix" and "suffix" suffixes
#4520 contains a [feature idea](https://github.com/Jermolene/TiddlyWiki5/issues/4520#issuecomment-651976032) by @ericshulman which should really have its own dedicated issue. It suggests extending the `trim` operator to take an optional parameter specifying the string to be trimmed at start or end (omitting the parameter would mean "trim whitespace" which is the current behavior). The trim operator could also take optional suffixes "prefix" and "suffix" which would trim only at the start, or only at the end, respectively.

`[[ foo ]] [[bar ]] [[ baz]] +[trim[]]` would produce `[[foo]] [[bar]] [[baz]]`
`[[ foo ]] [[bar ]] [[ baz]] +[trim:prefix[]]` would produce `[[foo ]] [[bar ]] [[baz]]`
`[[ foo ]] [[bar ]] [[ baz]] +[trim:suffix[]]` would produce `[[ foo]] [[bar]] [[ baz]]`

And with parameters:

`[[abba]] +[trim[a]]` would produce `[[bb]]`
`[[abba]] +[trim:prefix[a]]` would produce `[[bba]]`
`[[abba]] +[trim:suffix[a]]` would produce `[[abb]]`

With parameters, `trim` would stop at the first place where the string didn't match:

`[[ abba]] +[trim[a]]` would produce `[[ abb]]` (space doesn't match `a` so the prefix trim stops)
`[[abba ]] +[trim[a]]` would produce `[[bba ]]` (space doesn't match `a` so the suffix trim stops)

The `trim` operator should match as many copies of its parameter as are present at the start or end, and trim them repeatedly until the parameter no longer appears. E.g., `[[aaabbb]] +[trim[a]]` should produce `[[bbb]]`.

Partial matches or out-of-order matches would not be touched: `[[baobab]] [[abacus]] +[trim[ab]]` would produce `[[baob]] [[acus]]`, not `[[o]] [[cus]]`.

`removeprefix` and `removesuffix` also filter by the specified prefix/suffix, but there is no operator that trims a specified string without filtering, so there is no alternative that I know of to this feature idea.

There are two ways that the parameter to `trim` could be considered. One would be as a string that must be matched entirely to trim, e.g. `+[trim:prefix[ab]]` means "As long as the input starts with the two-letter sequence `ab`, trim it, but don't trim an `a` not followed by `b`", so "abacus" becomes "acus". The other way to consider the parameter would be as a *set of characters* that should be trimmed, e.g. `+[trim:prefix[ab]]` would mean "Trim all copies of `a` or `b` from the prefix until you hit a letter that is neither `a` nor `b`", so that "abacus" would become "cus". This might be more familiar to programmers from tools like Python's `string.strip()` method: `"abacus".strip('ab')` produces "cus" in Python 2.7. But that would *not* be how most non-programmers would expect it to work, and as a programmer myself, I would have far more use for the first approach, where the parameter is a complete string that must match at the beginning and end. That would make `trim:prefix` and `trim:suffix` parallel `removeprefix` and `removesuffix` except that `trim` does no filtering.

**Additional context**

rmunn:
Fixes #4809

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -34,9 +34,35 @@ exports.titlecase = makeStringBinaryOperator(
 	function(a) {return [$tw.utils.toTitleCase(a)];}
 );
 
-exports.trim = makeStringBinaryOperator(
-	function(a) {return [$tw.utils.trim(a)];}
-);
+exports.trim = function(source,operator,options) {
+	var result = [],
+		suffix = operator.suffix || "",
+		operand = (operator.operand || ""),
+		fnCalc;
+	if (operand !== "") {
+		// Safely regexp-escape the operand
+		operand = operand.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
+	}
+	if (suffix === "prefix") {
+		fnCalc = function(a,b) {return [$tw.utils.trimPrefix(a,b)];}
+	} else if (suffix === "suffix") {
+		fnCalc = function(a,b) {return [$tw.utils.trimSuffix(a,b)];}
+	} else {
+		if (operand === "") {
+			fnCalc = function(a) {return [$tw.utils.trim(a)];}
+		} else {
+			fnCalc = function(a,b) {return [$tw.utils.trimSuffix($tw.utils.trimPrefix(a,b),b)];}
+		}
+	}
+	source(function(tiddler,title) {
+		Array.prototype.push.apply(result,fnCalc(title,operand));
+	});
+	return result;
+};
+
+// makeStringBinaryOperator(
+// 	function(a) {return [$tw.utils.trim(a)];}
+// );
 
 exports.split = makeStringBinaryOperator(
 	function(a,b) {return ("" + a).split(b);}
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -39,10 +39,6 @@ exports.trim = function(source,operator,options) {
 		suffix = operator.suffix || "",
 		operand = (operator.operand || ""),
 		fnCalc;
-	if(operand !== "") {
-		// Safely regexp-escape the operand
-		operand = operand.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
-	}
 	if(suffix === "prefix") {
 		fnCalc = function(a,b) {return [$tw.utils.trimPrefix(a,b)];}
 	} else if(suffix === "suffix") {
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -39,16 +39,16 @@ exports.trim = function(source,operator,options) {
 		suffix = operator.suffix || "",
 		operand = (operator.operand || ""),
 		fnCalc;
-	if (operand !== "") {
+	if(operand !== "") {
 		// Safely regexp-escape the operand
 		operand = operand.replace(/[\\^$*+?.()|[\]{}]/g, '\\$&');
 	}
-	if (suffix === "prefix") {
+	if(suffix === "prefix") {
 		fnCalc = function(a,b) {return [$tw.utils.trimPrefix(a,b)];}
-	} else if (suffix === "suffix") {
+	} else if(suffix === "suffix") {
 		fnCalc = function(a,b) {return [$tw.utils.trimSuffix(a,b)];}
 	} else {
-		if (operand === "") {
+		if(operand === "") {
 			fnCalc = function(a) {return [$tw.utils.trim(a)];}
 		} else {
 			fnCalc = function(a,b) {return [$tw.utils.trimSuffix($tw.utils.trimPrefix(a,b),b)];}
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -467,6 +467,18 @@ function runTests(wiki) {
 		// Ensure that join correctly handles empty strings in source
 		expect(wiki.filterTiddlers("[[]] Paul +[join[-]]").join(",")).toBe("-Paul");
 		expect(wiki.filterTiddlers("[[ John ]] [[Paul ]] [[ George]] Ringo +[trim[]join[-]]").join(",")).toBe("John-Paul-George-Ringo");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim[]]").join(",")).toBe("abc,def,ghi");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim:prefix[]]").join(",")).toBe("abc ,def ,ghi");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim:suffix[]]").join(",")).toBe(" abc,def, ghi");
+		expect(wiki.filterTiddlers("[[ abacus ]] [[ baobab ]] +[trim[ab]]").join(",")).toBe(" abacus , baobab ");
+		expect(wiki.filterTiddlers("[[ abacus ]] [[ baobab ]] +[trim[a]]").join(",")).toBe(" abacus , baobab ");
+		expect(wiki.filterTiddlers("abacus baobab +[trim[a]]").join(",")).toBe("bacus,baobab");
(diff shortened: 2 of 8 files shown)


rmunn:
I'm working on a one-line summary to add to the "Release 5.1.23" tiddler, but I'm not 100% happy with the text I've written so far. Here's what I have come up with so far; suggestions for improvement would be welcome.

`* [[Extended|https://github.com/Jermolene/TiddlyWiki5/pull/4811]] the [[trim Operator]] to optionally trim a given string instead of whitespace, and trim from front, back, or both sides of input tiddlers`

pmario:
So it does remove "whitespace +  a given string" eg: `[[ abc ]trim[a]]` will output `[[bc]]` ... Is that right?

pmario:
Some new tests would be nice too, which imo may make my question obsolete ;)

https://github.com/Jermolene/TiddlyWiki5/blob/8cf458d3b3f1e38c14a2819529e08dca4a7e297c/editions/test/tiddlers/tests/test-filters.js#L457-L469

Does the test edition still pass? ... If there are questions about the test system. ... Feel free to ask.

pmario:
Would be interested, what happens to `[[ aabcabcab ]trim[a]]`?

rmunn:
> So it does remove "whitespace + a given string" eg: `[[ abc ]trim[a]]` will output `[[bc]]` ... Is that right?

No, it's either-or. If you specify a given string to remove, it does *not* remove whitespace, so your example would output `[[ abc ]]`. I wrote up several examples in the documentation, which I'll turn into unit test cases.

rmunn:
> Would be interested, what happens to `[[ aabcabcab ]trim[a]]`?

That would return the string unchanged since there are no `a` characters at the start or end of the input. If you want to remove whitespace from start and end and *then* remove `a` from start and end, you could do that with `[[ aabcabcab ]trim[]trim[a]]`, which would output `[[bcabcab]]` (First step becomes `[[aabcabcab]]`, then leading a's (2) and trailing a's (none) are removed producing `[[bcabcab]]`).

rmunn:
Yes, the test edition still passes. I'll add a few more unit tests to test `trim` behavior.

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -467,6 +467,18 @@ function runTests(wiki) {
 		// Ensure that join correctly handles empty strings in source
 		expect(wiki.filterTiddlers("[[]] Paul +[join[-]]").join(",")).toBe("-Paul");
 		expect(wiki.filterTiddlers("[[ John ]] [[Paul ]] [[ George]] Ringo +[trim[]join[-]]").join(",")).toBe("John-Paul-George-Ringo");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim[]]").join(",")).toBe("abc,def,ghi");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim:prefix[]]").join(",")).toBe("abc ,def ,ghi");
+		expect(wiki.filterTiddlers("[[ abc ]] [[def ]] [[ ghi]] +[trim:suffix[]]").join(",")).toBe(" abc,def, ghi");
+		expect(wiki.filterTiddlers("[[ abacus ]] [[ baobab ]] +[trim[ab]]").join(",")).toBe(" abacus , baobab ");
+		expect(wiki.filterTiddlers("[[ abacus ]] [[ baobab ]] +[trim[a]]").join(",")).toBe(" abacus , baobab ");
```
pmario, line 474: /*disapointed*/ .. But then there is no difference between removeprefix and removesuffix. .. I think `trim` should _always_ remove whitespace + eventual chars.

pmario: But that's just my opinion. ... What do others think?

rmunn: The difference between `trim` and `removePrefix` would be that `removePrefix` also *filters out* any titles that don't start with the string, whereas `trim` does no filtering. That was the original feature request: `removePrefix` without filtering, so the fact that trim-with-parameter acts exactly like `removePrefix` except for the filtering is precisely the design goal I'm going for here.

If I implement it this way, then what you want (`trim` removes whitespace and then the supplied characters) can easily be achieved by doing `+trim[]trim[a]`. If I implement it the way you're suggesting, then what I want (exactly like `removePrefix` but without filtering) would *not* be possible, since I would have no way to tell it *not* to remove any whitespace that I wanted to keep. Therefore, it's better to implement this the more flexible way, which allows for both use cases; and that's the way I wrote it here.

pmario: `+trim[]trim[a]` you are right. That's much more verbose and will create less support requests. I'm OK thx.

core/modules/filters/strings.js
```diff
@@ -34,9 +34,35 @@ exports.titlecase = makeStringBinaryOperator(
...
 
-exports.trim = makeStringBinaryOperator(
-	function(a) {return [$tw.utils.trim(a)];}
-);
+exports.trim = function(source,operator,options) {
+	var result = [],
+		suffix = operator.suffix || "",
+		operand = (operator.operand || ""),
+		fnCalc;
+	if (operand !== "") {
```
Jermolene, line 42: Our coding style is to omit whitespace before `if` and the opening parenthesis.

Jermolene:
Hi @rmunn @twMat @pmario thanks, looks good apart from a minor coding style issue.

core/modules/filters/strings.js
rmunn: Assuming you mean between `if` and the parentheses, which I see from other .js files (now that I look) is the style of the rest of the code, I'll make that change.

rmunn: Done in commit 73c018eec.

rmunn:
Okay, commit 73c018eec adjusts the coding style to match the rest of the code.

editions/tw5.com/tiddlers/filters/trim Operator.tid
```diff
@@ -1,13 +1,16 @@
 caption: trim
 created: 20190613153740241
 modified: 20190613153820282
+op-purpose: returns each item in the list with whitespace, or a given character string, trimmed from the start and/or end
```
Jermolene, line 4: The docs should clarify that the character string is interpreted as a regular expression, and not as plain text.

Jermolene (commented):
I hadn't realised that the new string parameter is interpreted as a regular expression, and not a static string. While it provides some flexibility, it makes it harder to trim strings that include regexp special characters. I wonder whether situations where the regexp functionality is useful might not be handled by a more generic regexp transformation filter operator.

rmunn:
It's not interpreted as a regexp; if you look at the trim function in strings.js, I escape any regex special characters before passing the operand to trimPrefix and trimSuffix in utils.js. And the tests I wrote in test-filters.js include two tests to ensure that the trim filter operator doesn't hiccup on special characters.

OTOH, perhaps it might be a good idea to do the regex escaping in utils.trimPrefix and utils.trimSuffix, so that if anyone in the future tries to use those functions in other parts of the Javascript code, they won't get caught out by the fact that those functions use regexes and need to have regex characters escaped. So I'll submit a new commit that moves the regex escaping into trimPrefix and trimSuffix. This will have the effect of doing the escaping twice rather than once in the case where both suffixes and prefixes are trimmed, but I doubt that will cause any noticeable difference in performance since most people will be passing very small strings to the trim operator. And the benefit is that utils.trimPrefix and utils.trimSuffix will become safer to call from other parts of the JS code.

rmunn:
Commit 494abc283 moves the regex-escaping into utils.trimPrefix and utils.trimSuffix so that it's easier to call them from other parts of the JS code. As a bonus, the strings.trim function looks cleaner and easier to read as well, so that's nice. The only drawback is a small amount of code duplication, but since it's only one line I'm not too worried about that.

editions/tw5.com/tiddlers/filters/trim Operator.tid
rmunn: It's not interpreted as a regular expression; see https://github.com/Jermolene/TiddlyWiki5/pull/4811#issuecomment-683203044 for details.

Jermolene:
Thanks @rmunn that's great.

PR #4812, opened 2020-08-13
Fix 4800 - remove communication to prodcut hunt on display.

### LINKED ISSUE #4800: [BUG] [Privacy Concern] The "Product Hunt badge" phones home on display - also in file TW
As in the title: The "Product Hunt badge" at https://tiddlywiki.com/#HelloThere phones home on display - also in file TW

To replicate: 

 - Download the whole wiki from Tiddlywiki.com open it in the browser
 - open dev-tools with F12
 - select Network Tab
 - reload the page. 

You will see it does a

```
GET | https://api.producthunt.com/widgets/embed-image/v1/top-post-badge.svg?post_id=193414&theme=dark&period=daily
```

Which causes unwanted traffic and imo also isn't GDPR conform. 

''Expected:''

The badge should link to the right badge, but shouldn't load anything from a 3rd party page. IMO TW can do this, since we are able to use image buttons with svgs. 

-------------

I found out about this issue, as I saw that uBlockOrigin did _not_ block the request. ... I wanted to disable the browser AddOn for tiddlywiki, since it injects some code, which may slow down the TW core, which I don't want for TW. Especially on mobile devices.

pmario:
Fix #4800 - remove communication to product hunt on display.

Did test it with FF latest and windows 10


Jermolene:
Thanks @pmario

PR #4813, opened 2020-08-13
Fix incorrect base64 encoding of astral-plane text

### LINKED ISSUE #4685: [BUG] GitHub Saver Corrupts Emoji
**Describe the bug**

New user here.

I wanted a Tiddler with Emoji's in it. For example 🎥.

Adding these Emoji directly with git works fine.

However, the TW GitHub saver seems to corrupt them. See this Gist of a patch the GitHub saver generated: https://gist.github.com/philips/78767c849be6f7ba625a6f3b68c0194f

**To Reproduce**
1. Go to any Tiddler after enabling GitHub saver
2. Click on Edit
3. Input an emoji like 🎥
4. See broken commits

**Expected behavior**
Emoji's saved without issue.

 - OS: ChromeOS 83
 - Browser Chrome

rmunn:
Fixes #4685.

Most astral-plane text is emojis like U+1F4DA BOOKS (📚), but some languages like Osage have their alphabet entirely in the supplementary multilingual plane as well. For proper support of languages like Osage, and newer emojis, the UTF-8 decode and encode functions need to properly handle codepoints above U+FFFF, which are represented by a surrogate pair in Javascript strings.

The included unit tests should help prevent a regression.


rmunn:
I've also reported this bug, and its fix, "upstream" at https://gist.github.com/Nijikokun/5192472 where the buggy code came from.

Jermolene:
Thanks @rmunn

PR #4814, opened 2020-08-13
Update navigator widget docs

pmario:
relinkOnRename was missing since 5.1.14


twMat:
I vaguely recall @Jermolene saying that "since version"-info only goes back some 2-3 versions, so a note about 5.1.14 would be much too old if I'm not mistaken.

Off topic but it seems very inefficient to have these manually added and removed in the docs like this.  Maybe individual doc tiddlers could have some automated link to the Release tiddler.

pmario:
We don't remove these version infos. .. But may be this one is too early. I'll have a closer look.

pmario:
There are seversal elements, that have a version info 5.1.14. .. But it seems to be the first one. So I'd like to keep it.

Jermolene:
Hi @pmario looks good, but please adjust the PR to target the tiddlywiki-com branch.

pmario:
If I target tiddlywiki-com I do get a list of several 100 changed files.

pmario:
IMO did it wrong. will create a new one

PR #4793, opened 2020-08-09
Add GitHub Actions workflow

ento:
Fixes #4789 (indirectly)

Adds `.github/workflows/ci.yml` that runs the same scripts as `.travis.yml` in three stages (jobs in GitHub Actions parlance): test, build-prerelease, build-tiddlywiki-com.

How it looks like in action (pun intended): https://github.com/ento/TiddlyWiki5/actions/runs/201083222

^This run is for a branch that contains [a commit](https://github.com/ento/TiddlyWiki5/commit/0493f641fb3547d584852ac88f948fa9bff453a6) that allows the `build-*` jobs to run on any branch, so that we can see how these jobs get run. The `build-*` jobs ultimately fail because the necessary push token is not configured.

What happens when this PR is merged without disabling Travis:

- The `test` job will get run on every PR and on every push to `master` and `tiddlywiki-com` branches
- The `build-prerelease` job will get run on every push to `master`, but will fail at the final `git push` command until a `GITHUB_PUSH_TOKEN` secret is added to the repo that has push access to the destination repo
- The `build-tiddlywiki-com` job will get run on every push to `tiddlywiki-com`, but will fail at the final `git push` command until a `GITHUB_PUSH_TOKEN` secret is added to the repo that has push access to the destination repo
- Travis builds will continue to run as well.
  - If  `GITHUB_PUSH_TOKEN` secret is configured and the `build-*` stages/jobs get run in close parallel between Travis and GitHub, I expect one of the runners to fail; first one to run `git push` will succeed and the other one won't be able to push due to their local branch being behind the remote. If builds for one of the services get too delayed, the sites can get reverted to an older state when the delayed builds finally get run.

## Further actions required to complete the switch from Travis

* [Write access needed] Test if the `build-prerelease` and `build-tiddlywiki-com` jobs of the workflow work
  * Create a test branch and add it to `on.push.branches` and adjust the `if` conditions of jobs so that they get run on that branch.  (This needs to be done in a branch in this repo, so that the workflow can access the secret configured in the next step.)
  * [Admin access needed] [Create a secret](https://docs.github.com/en/actions/configuring-and-managing-workflows/creating-and-storing-encrypted-secrets#creating-encrypted-secrets-for-a-repository) named `GITHUB_PUSH_TOKEN` that has push access to `Jermolene/jermolene.github.io`
[...]

PR #4835, opened 2020-08-28
Fix 'if' condition syntax in Actions workflow

ento:
#4793 introduced a new GitHub Actions workflow file, and runs are failing with:

> Invalid workflow file
> The workflow is not valid. .github/workflows/ci.yml (Line: 23, Col: 9): Unexpected symbol: '"refs/heads/master"'. Located at position 15 within expression: github.ref == "refs/heads/master",.github/workflows/ci.yml (Line: 48, Col: 9): Unexpected symbol: '"refs/heads/tiddlywiki-com"'. Located at position 15 within expression: github.ref == "refs/heads/tiddlywiki-com"
> https://github.com/Jermolene/TiddlyWiki5/actions/runs/228978450

The reason appears to be that strings in `if` [need to be in single quotes](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#literals), not double quotes. (🤔 I'm not sure why it didn't fail in my repo. Mysterious. -- Edit: because I commented out these `if` conditions in order to test, hah.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,7 +20,7 @@ jobs:
       - run: "./bin/test.sh"
   build-prerelease:
     runs-on: ubuntu-latest
-    if: github.ref == "refs/heads/master"
+    if: github.ref == 'refs/heads/master'
     needs: test
     env:
       TW5_BUILD_TIDDLYWIKI: "./tiddlywiki.js"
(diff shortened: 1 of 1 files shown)


ento:
Oh, it seems like changes to a workflow file made by PRs from a forked repo get picked up during runs executed for the PR. (I'd assumed otherwise, so this is nice.)

>     if: github.ref == 'refs/heads/master'
> https://github.com/Jermolene/TiddlyWiki5/actions/runs/229017410/workflow

Jermolene:
Thanks @ento

PR #4850, opened 2020-09-12
Removed bad-character warning

flibbles:
@Jermolene, you once said that you wanted to remove the "bad character warning" that shows up if you use brackets (or some such) in a tiddler title. I don't remember the exact context, but it had to do with other problematic titles I tried to add to the warning while I was building Relink.

But in preparation for its removal, I removed that warning in Relink, and someone is pointing out the inconsistency it now has with core. I either should to put that warning back in, or find out if it ever will be removed from core tiddlywiki.

This pull request will remove it and all associated language strings, if that's the direction you want to go. If not, feel free to reject this PR.

-Flibbles

### DIFF
--- a/core/language/en-GB/EditTemplate.multids
+++ b/core/language/en-GB/EditTemplate.multids
@@ -21,7 +21,6 @@ Tags/Add/Button/Hint: add tag
 Tags/Add/Placeholder: tag name
 Tags/Dropdown/Caption: tag list
 Tags/Dropdown/Hint: Show tag list
-Title/BadCharacterWarning: Warning: avoid using any of the characters <<bad-chars>> in tiddler titles
 Title/Exists/Prompt: Target tiddler already exists
 Title/Relink/Prompt: Update ''<$text text=<<fromTitle>>/>'' to ''<$text text=<<toTitle>>/>'' in the //tags// and //list// fields of other tiddlers
 Title/References/Prompt: The following references to this tiddler will not be automatically updated:
--- a/core/ui/EditTemplate/title.tid
+++ b/core/ui/EditTemplate/title.tid
@@ -3,20 +3,6 @@ tags: $:/tags/EditTemplate
 
 <$edit-text field="draft.title" class="tc-titlebar tc-edit-texteditor" focus={{{ [{$:/config/AutoFocus}match[title]then[true]] ~[[false]] }}} tabindex={{$:/config/EditTabIndex}} cancelPopups="yes"/>
 
-<$vars pattern="""[\|\[\]{}]""" bad-chars="""`| [ ] { }`""">
-
-<$list filter="[all[current]regexp:draft.title<pattern>]" variable="listItem">
-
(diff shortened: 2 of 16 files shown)


Jermolene:
Thanks @flibbles I think my comment about removing the warning was because we've tried to remove those limitations from the core, and now those characters are only "bad" if one subsequently tries to reference them in an incompatible way (e.g. not being able to use tiddler titles containing a vertical bar within the `[[link]]` syntax).

I'd be interested to hear from others about this though. For example @pmario @saqimtiaz @BurningTreeC @twMat what do you think?

twMat:
>...removing the warning was because we've tried to remove those limitations from the core, and now those characters are only "bad" if one subsequently tries to reference them in an incompatible way

If something risks causing problems down the line, then of course one would want to know about it. Maybe soften up the warning and include a revealwidget or an external link to a tw.com doc explaining it.

Jermolene:
> If something risks causing problems down the line, then of course one would want to know about it. Maybe soften up the warning and include a revealwidget or an external link to a tw.com doc explaining it.

But the point is that now there are very few situations where those characters are "bad": as far as we know, there are no problems in the core user interface, the risk is just if the user writes some custom wikitext without taking those special characters into account. So it seems premature to give everybody a big scary warning if it only really affects those people that are writing custom wikitext.

twMat:
>But the point is that now there are very few situations

OK, "very few" - fair enough, skip the warning.

pmario:
So I think the warning should stay

----------

I personally would have not allowed the problematic characters in titles at all. They still cause problems, in certain edge cases, that we don't know yet. 

eg `'exportMe'` is a valid tiddler name, but it can't be exported as JSON from the **AdvancedSearch : Filter** tab using a valid filter string like: `"'exportMe'"` .. The tiddler has to exist.

saqimtiaz:
My motivation for the PR I made to the Relink plugin to add in the warning, was out of a desire for consistency between the core and the plugin. I've had one private support request that was due to a pipe character in a tiddler title, and the user had not recalled seeing a warning about using that character in tiddler titles. This was followed by a GG discussion with a [similar issue](https://groups.google.com/d/msg/tiddlywiki/iu-y7CVx6dg/MRz8XJ-sBAAJ) which made me realize this was due to the Relink plugin.

Not only is the inconsistency confusing for end-users but also for those providing support. 

I do think that not being able to reference tiddler titles containing a vertical bar inside `[[link]]` syntax is fairly significant. As a new user I would want to be aware of this before I have created a heap of tiddlers that may then need to be renamed. 

Is there a better way to communicate this to new users if we wish to remove the warning? Perhaps a checkbox toggle to not show the warning again is a compromise.

Jermolene:
Thanks @saqimtiaz @pmario @twMat that sounds like a fairly resounding vote in favour of keeping the warning.

flibbles:
So I'll just close this then.

Jermolene:
Thanks @flibbles

PR #4853, opened 2020-09-14
Joining an empty list should return an empty list

### LINKED ISSUE #4852: [BUG] Join operator returns non-empty output from empty input
**Describe the bug**
An empty filter output if directed as input to a join operator will return something!!!

**To Reproduce**
2. Create a tiddler and put the below wikitext inside its text field

`
<$list filter="[tag[not-exist]]" emptyMessage="I am empty">
I am here
</$list>
`
Save and see the result. It will shows `I am empty`

3. Next change the wikitext with below one
`
<$list filter="[tag[not-exist]join[_]]" emptyMessage="I am empty">
I am here
</$list>
`
Save and see the result: `I am here`

That means an empty filter output if fed to a join operator will return non-empty output.

**Expected behavior**
Nothing shall be displayed!

**Screenshots**

 - OS: Windows 10
 - Browser Chrome 81
 - TW 5.1.22

While I know this behavior may be useful in some situation, but this is confusing!

rmunn:
Fixes #4852.

Note that this changes behavior introduced in TiddlyWiki 5.1.22 by https://github.com/Jermolene/TiddlyWiki5/pull/4396. Specifically, #4396 made join return `[""]` (a list containing a single empty string) if passed an empty input list, while this PR makes join return `[]` (an empty list) if passed an empty input list. Before #4396, join would return `[null]` (a list containing a single null) on empty input, which was causing Red Screen of Death errors. While returning `[""]` does stop the crashes, I believe that it's not what users would expect, and users would expect join to return an empty list when given empty input.

I would also like code-style comments: I've written `if(result.length === 0) return [];` as a single line here. Should I instead write this with braces? It seems short enough to merit a one-liner, but some people don't like that.

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -90,6 +90,9 @@ function makeStringReducingOperator(fnCalc,initialValue) {
 		source(function(tiddler,title) {
 			result.push(title);
 		});
+		if(result.length === 0) {
+			return [];
+		}
 		return [result.reduce(function(accumulator,currentValue) {
 			return fnCalc(accumulator,currentValue,operator.operand || "");
 		},initialValue) || ""];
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @rmunn that makes sense.

> I would also like code-style comments: I've written `if(result.length === 0) return [];` as a single line here. Should I instead write this with braces? It seems short enough to merit a one-liner, but some people don't like that.

Our house style is to always use braces.

rmunn:
@Jermolene Okay, I've rewritten this PR to use braces instead. And while I was at it, I rebased it onto the current state of the `master` branch.

rmunn:
@Jermolene Anything else you want me to change here, or is this ready to merge?

Jermolene:
Thanks @rmunn apologies for the delay.

PR #4859, opened 2020-09-17
Fix tiddler title alignment if no icon is acitve

pmario:
The tiddler title is miss-aligned by 0.1em margin right from an empty span, if no icons are specified. 

The image shows it all. "Hello There" text is moved to the right, because of the `.tc-tiddler-title-icon` margin-right setting. 

The PR fixes the miss-alignment, because the span is only included into the DOM, if a tiddler icon-field or $:/config/DefaultTiddlerIcon is defined.

tiddler-title-not-aligned-if-no-icon


pmario:
@Jermolene ... Don't merge yet. The span also makes the title line higher, which is a side effect that is needed. I will probably need to adjust a CSS setting.

pmario:
@Jermolene .. Can be merged now. .. The line height fix is needed for titles that contain chars like "g".

Jermolene:
Thanks @pmario

Jermolene:
Thanks @pmario

PR #4863, opened 2020-09-22
Fix 4860 parser problem with tripple double qoutes.

### LINKED ISSUE #4860: Advanced Search -> Export Tiddlers drop-down breaks when double quotations are used in Filter tab
I was looking for a way to list a few individual tiddlers for export together into a single JSON file. I entered "Tiddler One" "Tiddler Two" into the Filter tab search box (which properly matched the tiddlers that I wanted) and clicked the export tiddlers button. Instead of the normal options, each option showed a bunch of underlying code and didn't function correctly.

I finally figured out the proper way ( [[Tiddler One]] [[Tiddler Two]] ... also using single tick-mark/apostrophe quotes works ) but it seemed that the intuitive way of using double quotes that properly matched shouldn't break the export feature.

pmario:
This PR fixes #4860 

The attached ZIP file contains 2 TW versions. "parser-problem" is the one with the problem.

There are 3 tiddlers, that show the problem if in preview-mode, parse-tree is selected. 

-----
Edit: This is the new test.html
[test.zip](https://github.com/Jermolene/TiddlyWiki5/files/5263232/test.zip)

These are the new test tiddlers, that can be used with 5.1.22. If you check the parsetree it will fail.

If you import them into the test.html it should all be fine. 

[parser-problem-with-tripple-doublequote.zip](https://github.com/Jermolene/TiddlyWiki5/files/5263205/parser-problem-with-tripple-doublequote.zip)


pmario:
@Jermolene Don't merge yet. The tests pass, but tiddlywiki.com breaks

pmario:
The tests are not good enough. It seems I have a regexp that is OK, but need more testing

pmario:
I did update the initial post with the new files. @Jermolene ... There are new tests and the fixes now. Please test. @rmunn can you test it?

Jermolene:
Thanks @pmario. It looks like the original problem reported in #4860 is fixed by the change to `core/wiki/macros/export.tid`, is that right? What is the intention behind the regexp changes?

pmario:
>Thanks @pmario. It looks like the original problem reported in #4860 is fixed by the change to core/wiki/macros/export.tid, is that right? What is the intention behind the regexp changes?

No the problem is super weird. The changes to the export.tid only seemed to fix the export problem. 
I needed about half a day of single step debugging to find the real problem.

With the regexp fix, core/wiki/macros/export.tid doesn't need to be changed. .. So the regexp fix also hardens the usage of tripple-double-quotes. (Which, sometimes, has advantages over the new macro-like syntax)

eg: if you export `"a" "b"` _and_ "a" and "b" are real tiddlers, only "a" was exported, because the regexp "eats" the second double-quote from "b". So it became "b ... See the new tests. (if you cherrypick them without the regexp stuff.)

The real problem is the regexp and was an oversight in commit: a2ec4c55e3af794e83ef417088a700ee5e299c30, which doesn't take care of "a b" being equal to [[a b]]. 

Check out this text `""""a" "b""""` with the existing regexp and the new regexp. You'll see the difference in your regexp-tool. Also have a close look at `test-html-parser` in the new tests. ... There are some tests with 4 double-quotes at the end and some with 3.

With the "old" regexp all of them with 4 "'s fail. 

As I wrote. It is much trickier as it seems at first and it was a really time consuming fix. All in all with the tests 2 days :/
And my first fix was a complete fail, since the regexp needed to be more specific.

Jermolene:
> No the problem is super weird. The changes to the export.tid only seemed to fix the export problem.
I needed about half a day of single step debugging to find the real problem.

OK, so there's actually two fixes in this PR:

* A fix for the export problem by switching from triple double  quotes and string substitution to the new syntax for accessing macro parameters as variables. This would be the recommended best practice; using triple double  quotes and string substitution is now deprecated
* An improvement to the parsing of triple double  quote delimited strings, making them less greedy so that they work with trailing double quotes

Given that we have a better alternative to triple double quote delimited strings in most situations, I would be inclined not to bother with the second part of this PR in order to avoid unexpected backwards incompatibility. It also looks as though these new regexps are slower than the old ones.

pmario:
>A fix for the export problem by switching from triple double quotes and string substitution to the new syntax for accessing macro parameters as variables. This would be the recommended best practice; using triple double quotes and string substitution is now deprecated

IMO Tripple double quotes can't be deprecated, because that would be a feature regression.  eg:  \define test(a, b) """$a$$b$""" is fundamentally different than `\define test(a, b) <<__a__>><<__b__>>` how it is resolved. The first example makes a proper string concatenation, the second doesn't. 

That's exactly THE problem every users struggles with.

pmario:
>I would be inclined not to bother with the second part of this PR in order to avoid unexpected backwards incompatibility. It also looks as though these new regexps are slower than the old ones.

BUT it's an issue! ... Have a look at: https://github.com/Jermolene/TiddlyWiki5/issues/4860#issuecomment-695933592 .. If you want to export those 2 tiddlers, only the first one will be exported! That's a big problem and can cause data loss. 

If I'm exporting tiddlers from a wiki and then delete them, because I want to import them into a different wiki. I'll definitely loose data.

pmario:
>OK, so there's actually two fixes in this PR:

No, there's only 1 fix. the regexp. The `core/wiki/macros/export.tid` can be reverted back to tripple quotes.

Jermolene:
Hi @pmario I've fixed this differently in ae13a0fee118c50169b7835c950df1dade33788f and 59f233cd46616646fa8889f65aa9cc7d704d8c9a

PR #4865, opened 2020-09-24
Change description for Filesystem and TiddlyWeb plugins

pmario:
[Quote from GG](https://groups.google.com/d/msg/tiddlywiki/cHv4bWvZo00/3Aa6kuTFBgAJ)

So this just cost me about 3 hours and I was really lucky to lose only a few changes to tiddlers.

The plugin is described as "Synchronise changes to the local filesystem" so I thought it had something to do with running TW locally as a single file. But it's actually the mechanism that the node server uses to save tiddlers.

The insidious thing is that it only becomes effective the next time the server is restarted - which could be months after you made the change in the settings and have forgotten about it completely. And then you still don't notice anything wrong because the server keeps changes in memory - until you restart it the next time and all changes in between are lost!

I think it would be a good idea at minimum to change the plugin description, or even make it impossible to disable.

----------

This PR should make the dataflow more obvious.

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw: false */
+/*global $tw, console, exports */
 "use strict";
 
 /*
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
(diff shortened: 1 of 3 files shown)


Jermolene:
Thanks @pmario

pmario:
@Jermolene .. do you have an idea, how I can get rid of all the "unnecessary" commits that are shown here?

Jermolene:
Hi @pmario 

> do you have an idea, how I can get rid of all the "unnecessary" commits that are shown here?

Sorry, no, Git is still largely a mystery to me...

pmario:
@BurningTreeC Hi, do you know a solution for https://github.com/Jermolene/TiddlyWiki5/pull/4865#issuecomment-698827336

PR #4869, opened 2020-09-29
Tests fail, but the test runner shows OK

pmario:
The following commits: 
 - https://github.com/Jermolene/TiddlyWiki5/commit/ae13a0fee118c50169b7835c950df1dade33788f and 
 - https://github.com/Jermolene/TiddlyWiki5/commit/59f233cd46616646fa8889f65aa9cc7d704d8c9a 

don't solve the underlying filter problem, which can be seen with the new tests. They fail, but the expected output is right.

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw: false */
+/*global $tw, console, exports */
 "use strict";
 
 /*
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -9,7 +9,7 @@ Adds tiddler filtering methods to the $tw.Wiki object.
 (function(){
 
 /*jslint node: true, browser: true */
-/*global $tw, console, exports */
+/*global $tw:false */
 "use strict";
 
 /*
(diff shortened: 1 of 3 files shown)


pmario:
@Jermolene ... There seems to be something wrong with the automated test-runner. The 4th job, run ./bin/test.sh fails, but the message here is: "All checks have passed", which imo is wrong!

Jermolene:
Thanks @pmario my concern with trying to fix these issues is the risk that we'll make some hairy, slow regexps even hairier and slower. Also, I think there are now very few situations where we don't have a better alternative to triple double quotes and text substitution.

pmario:
OK. ... but the tests fail and the runners seem to be happy.

pmario:
I did rename the PR, so we can use it to track this problem.

Jermolene:
Hi @pmario 

> There seems to be something wrong with the automated test-runner. The 4th job, run ./bin/test.sh fails, but the message here is: "All checks have passed", which imo is wrong!

Ouch, yes indeed. I think the problem is that `./bin/test.sh` doesn't exit with an error if the tests fail, I'll look into it.

Jermolene:
Thanks @pmario I've had a poke around and looked in the Jasmine docs. Jasmine doesn't seem to call `process.exit()` so I tried to use the `jasmine.onComplete()` callback as documented [here](https://jasmine.github.io/setup/nodejs.html#custom-oncomplete), but that failed because "onComplete" is not a function. Generally, the Jasmine docs aren't much help because of our custom setup but worse than that they completely omit methods e.g. there's no visible docs for the `jasmine.configureDefaultReporter()` method. I'd be grateful if you got a chance to have a look too.

ento:
Looking into this

ento:
Made a PR: https://github.com/Jermolene/TiddlyWiki5/pull/4889

Jermolene:
Thanks @ento @pmario I've merged #4889, let me know how it goes.

pmario:
@Jermolene ... It seems to work now as expected. I'll change the test, so they will pass. Once they pass, the PR should be ready to be merged.

pmario:
Will close it. No merge needed.

PR #4870, opened 2020-09-30
Change some single quotes to double quotes for consistency reasons

pmario:
There has been a discussion at the [GG about frustration](https://groups.google.com/d/msg/tiddlywiki/jQpUdgQQQWc/GyitFu2SCgAJ) because of little docs inconsistencies.

### DIFF
--- a/editions/tw5.com/tiddlers/ActionCreateTiddlerWidget Example 1.tid
+++ b/editions/tw5.com/tiddlers/ActionCreateTiddlerWidget Example 1.tid
@@ -1,5 +1,5 @@
 created: 20200131142401129
-modified: 20200131152023958
+modified: 20201020111232755
 tags: ActionCreateTiddlerWidget
 title: ActionCreateTiddlerWidget Example 1
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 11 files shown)


Jermolene:
Hi @pmario I've seen the original post, but what is the rationale of these changes? Quotes are optional around attribute values, just as they are in HTML, and that we support multiple types of quote symbol. Changing our examples to only use double quotes would seem to obscure this truth.

pmario:
I think, it's about consistency. For new users it can be confusing, if you see one example with quotes and an other without. ...

Most of the time it isn't a problem, but if a user creates a macro, that doesn't work, they have to test one more option. If all examples use quotes, they'd add it and have no problems.

The "naive rule" is: If a value has no spaces, then it works without spaces. As I wrote, it's naive to think so. Values without spaces like: "$:/system" have to have quotes around them. The following code doesn't work:

```
<$vars xx=$:/tags/Stylesheet>
<<xx>>
</$vars> 
```

You reminded me at: https://github.com/Jermolene/TiddlyWiki5/pull/3975#discussion_r300867058, that I didn't use quotes some time ago. I had to add them for consistency reason. 

This change was requested from a new user, because it obviously did create problems for him. ... I'm confused.

kookma:
> Hi @pmario I've seen the original post, but what is the rationale of these changes? Quotes are optional around attribute values, just as they are in HTML, and that we support multiple types of quote symbol. Changing our examples to only use double quotes would seem to obscure this truth.

We can think about good programming style here and for example use double quote everywhere! Then in the note we can state
* Shingle quote can be used instead of double quote
* If the parameter has no space, one can omits quotes 

So we follow a logic and then explain there is other possibilities.

pmario:
Thx for the feedback. I'm not sure, if there is some explanation about the usage of quotes. 

eg: Sometimes we need: `param='The outer quote is single, because I want to see the "inner quotes", which are double'.` and vice versa. The second possibility is: `param="""The outer quote is tripple-double, because I want to see the "inner quotes", which are double."""`

> - If the parameter has no space, one can omits quotes

If you have a closer look at my example, that's not always true!

Issue #1865, opened 2015-07-06
Use automated tooling to enforce coding style


Jermolene:
There is an opportunity to use [JSCS](http://jscs.info/) to automatically enforce our chosen coding style guidelines, as described in [this blog post](http://ariya.ofilabs.com/2015/06/detecting-and-automatically-fixing-javascript-code-style.html).
- Build a `.jscsrc` file that corresponds to our existing coding standards
- Add it to the dev dependencies of the repo
- Add a script for checking conformance, and call it from `bld/test.sh`

Any help with this ticket would be much appreciated.

pmario:
+1
I did install the JSCS plugin for the Brackets editor some time ago. There it works similar to the JSHint plugin. Every time you save you get some warnings that you can fix, or ignore :/

There are several plugins for different popular editors. 

I also had a look at some predefined coding guidelines, but non of them fits to your actual settings. I didn't know about the `auto-configure` option. ... Would be worth a try. 

 - We will need some special settings for our leading comments and comments in general. 
 - Most of the presets prefer spaces instead of tabs, but this can be overwritten. 

And if JSCS implementation is on the way, it would be nice to run all source files through JSHint too. There are a lot of warnings all over the places atm. Especially "semicolons" are a problem. ... But since programs could change behaviour if we don't take care fixing them, I didn't touch them.

sukima:
Our company enforces coding standards with [Hound-ci](https://houndci.com/) but that uses [JSHint](http://jshint.com/about/).

Jermolene:
@sukima and @pmario I'm embarrassed by how bad the TW5 JSHint feedback report has got. I did once keep on top of it. I'll open a ticket about it.

pmario:
I think JSCS looks promising, if we find the right parameters, that fit the existing style.

pmario:
related to:  Remove function wrapper / add matching dprint and eslint configurations #7596

pmario:
Close this one in favour of ongoing effort for pretty printing: 

- https://github.com/TiddlyWiki/TiddlyWiki5/pull/9058
- https://github.com/TiddlyWiki/TiddlyWiki5/pull/7474
- https://github.com/TiddlyWiki/TiddlyWiki5/issues/9113

PR #4872, opened 2020-10-01
Add eslint plus very tolerant starting config

hoelzro:
Addresses GH #1865

This adds eslint as a developer dependency, plus a generated eslint
config that doesn't take a very strong stance on much of anything.  The
goal here to get started using automated style checking, add eslint
checking to the testing flow, and gradually introduce stricter checks
over time.


hoelzro:
I have a few additional thoughts on this:

  * I'd like to eventually add eslint as part of the GitHub actions that run as part of the test suite. To prevent overwhelm, I was thinking we could do a sort of "backwards-compatible" linting, where violations that occur on lines whose blame predates the addition of eslint are prevented from failing the test suite - that is, only new lines or changed lines that violate the rules would fail the test suite.  More thought will need to be put into how this would work with an eslint config that changes over time.
  * All of the settings marked "temporary" were _not_ included in the generated eslint config and generated failures - it would nice to remove these over time and fix up violations.
  * Other settings (especialy `indent`) should probably be turned on, tuned, and have their violations fixed over time.  I think this is going to take a good deal of community participation and input from @Jermolene!

ento:
I worked on [a branch](https://github.com/Jermolene/TiddlyWiki5/compare/master...ento:add-eslint) for adding eslint support a while ago - there might be useful things in there. The branch was put in the freezer while waiting for a GitHub action for running linters that can add line comments ([example](https://github.com/AlexsLemonade/resources-portal/pull/436/files#diff-d1edad457cab496940b4bced907d6a0f)) to [support PRs from forks](https://github.com/wearerequired/lint-action/issues/13).

* Backwards-compatible linting: the branch takes the approach of disabling rules per file when the file doesn't conform to those rules. Upside of this is that fewer new violations will be introduced and it'll be possible to work on fixing existing violations by making small PRs that don't involve sweeping changes. There's a script for generating the config for that (`estlint/gen-todo.js`). Downside could be that the workflow for fixing existing violations may seem complicated.
* TiddlyWiki specific rule(s): The branch has a custom rule for linting TiddlyWiki's code style for function calls (`foo(a,b)` - no spaces between arguments) in `eslint/plugin/`. In general, the branch's rules extend from `eslint:recommended` and enable/disable further rules to match TiddlyWiki's style guide.
* I somehow assumed that it's desirable to keep the top-level package.json's devDependencies empty; there might be unnecessary complexity coming from that in the branch.

hoelzro:
Thanks @ento, I'll check that out!  The backwards-compatible linting thing sounds in line with what I had in mind - I was thinking the that the tooling would cross-reference the list of violations against the Git history for each offending line.  If an offending line entered history before the eslint changes landed, it would be given a pass (these would probably be tracked via a warning or something) - if a line can't be blamed to a commit or the commit came in after the eslint changes, only then would that fail the tests.  This way, you'd only need to fix a violation on a line if you're already changing that code, or if you're introducing new code.

I think it'd be worth merging your branch if @Jermolene approves - the line comment stuff could always get added later!

Jermolene:
Hi @hoelzro @ento thanks, and apologies for the delay. I like the sound of backwards-compatible linting.

PR #4873, opened 2020-10-02
Add documentation for indexer modules

hoelzro:
This adds some documentation on what indexer modules are for and which methods they need to implement.


saqimtiaz:
@Jermolene one more docs update that I think we can merge.

Jermolene:
Great stuff -- sorry I missed this @hoelzro

PR #4875, opened 2020-10-03
Fix #4826 - Keyboard-driven-inputs clashes with customized search results.

### LINKED ISSUE #4826: [BUG] Keyboard-driven-inputs clashes with customized search results.
I am creating an issue for this as I am not sure if notifications work for [already merged PRs](https://github.com/Jermolene/TiddlyWiki5/pull/4725).

@BurningTreeC @Jermolene

I've discovered a significant issue with the way we have implemented `keyboard-driven-inputs` in #4725 

It requires both the keyboard driven input macro that creates the search input, and the code handling displaying the search results to use the same filter to return search results.

However, the search results can be customised:
https://tiddlywiki.com/#Customising%20search%20results

With an alternative $:/tags/SearchResults tiddler like the one provided in the example at the link above, the keyboard navigation is completely broken since the keyboard driven input and the search results display code no longer use the same filter.

I haven't had the opportunity to give this much thought, but the answer might lie in storing the search filter to be used in a common location that can be accessed by both the macro and the search results code.

BurningTreeC:
This PR fixes the problem with customized search results and the keyboard-driven-input macro.

There's some documentation needed, that will follow in a second PR

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -67,8 +67,8 @@ Shortcuts/Input/Accept/Hint: Accept the selected item
 Shortcuts/Input/AcceptVariant/Hint: Accept the selected item (variant)
 Shortcuts/Input/Cancel/Hint: Clear the input field
 Shortcuts/Input/Down/Hint: Select the next item
-Shortcuts/Input/Tab-Left/Hint: Select the next Tab
-Shortcuts/Input/Tab-Right/Hint: Select the previous Tab
+Shortcuts/Input/Tab-Left/Hint: Select the previous Tab
+Shortcuts/Input/Tab-Right/Hint: Select the next Tab
 Shortcuts/Input/Up/Hint: Select the previous item
 SystemTiddler/Tooltip: This is a system tiddler
 SystemTiddlers/Include/Prompt: Include system tiddlers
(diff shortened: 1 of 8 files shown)


saqimtiaz:
@BurningTreeC thank you for looking at this and apologies for the late reply. This looks good at first glance and seems like an elegant solution with placing the search filters in the fields of the search results tiddlers. Do you have a demo where this can be tested?

BurningTreeC:
Thanks @saqimtiaz , there's a Demo at http://kb-driven-search-results.tiddlyspot.com/

saqimtiaz:
> Thanks @saqimtiaz , there's a Demo at http://kb-driven-search-results.tiddlyspot.com/

This looks good @BurningTreeC. Note that it is better to close all tiddlers except the Instructions one, otherwise the input fields in the river keep stealing focus. Perhaps update the demo so that only that tiddler is open, so others can test easily.

@Jermolene we should try to get this into 5.1.23. It addresses an issue introduced by the keyboard driven search results. Namely that while we can have custom search results which might use a different filter, it means the keyboard navigation of the results might not match the custom results leading to haphazard and mismatched navigation.

Jermolene:
Hi @BurningTreeC @saqimtiaz thanks. It looks like you'll need to switch this PR out of draft status before it can be merged.

BurningTreeC:
@Jermolene - done

Jermolene:
Thanks @BurningTreeC

Jermolene:
@BurningTreeC – the documentation updates would be much appreciated

PR #4878, opened 2020-10-04
support save into BitBucket "Downloads" section

OmbraDiFenice:
As noted in https://github.com/Jermolene/TiddlyWiki5/pull/3931 BitBucket doesn't support pushing files into the repo through REST API. But what it does support is the upload of files as repository artifacts, and this is what this saver uses to store the wiki.

The file uploaded in this way won't be versioned by git so every time we save it will be overwritten and it won't be possible to see past versions, but it will be stored and be available in the "Downloads" section of the BitBucket repository.

### DIFF
--- a/core/language/en-GB/ControlPanel.multids
+++ b/core/language/en-GB/ControlPanel.multids
@@ -115,6 +115,9 @@ Saving/GitService/GitLab/Caption: ~GitLab Saver
 Saving/GitService/GitLab/Password: Personal access token for API (see [[GitLab help page|https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html]] for details)
 Saving/GitService/Gitea/Caption: Gitea Saver
 Saving/GitService/Gitea/Password: Personal access token for API (via Gitea’s web interface: `Settings | Applications | Generate New Token`)
+Saving/GitService/BitBucket/Caption: BitBucket
+Saving/GitService/BitBucket/Description: These settings are only used when saving to ~BitBucket. The file will be saved as a downloadable artifact and will be available in the "Downloads" section of the repo
+Saving/GitService/BitBucket/Password: Password or ~AppPassword for your ~BitBucket account
 Saving/TiddlySpot/Advanced/Heading: Advanced Settings
 Saving/TiddlySpot/BackupDir: Backup Directory
 Saving/TiddlySpot/ControlPanel: ~TiddlySpot Control Panel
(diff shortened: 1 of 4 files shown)


pmario:
Hi @OmbraDiFenice ... I think, there is a little problem with the PR. It also changes the core/modules/utils/dom/http.js .. That's not allowed if you only want to add a saver. 

Savers and may be other functions may rely on this module. ... So we have to be sure, that this change doesn't have a side effect to other elements of TW. 

Did you test this?

pmario:
Can you explain, what you did in core/modules/utils/dom/http.js and why?

OmbraDiFenice:
I needed to handle the case when data is an instance of FormData and make sure to propagate the object without changes if it's the case. The bitbucket API requires to upload the file as multipart/form-data, and the current code only supports data being string or a map of strings.

I could have changed the condition on the "if string" to include also my case, since also that is just propagating the value as it is, but I figured that this way it would have been easier to separate behaviors in case they diverge in the future (and in fact I realized that I also need to make sure that the browser will handle the Content-Type header by itself).

In theory my changes impact a case isolated enough to not cause problems to existing code and probably not even to future additions, but I don't have tests for that. 

I didn't notice any test for http.js (I didn't look for them tbh), can you point me to where should I put them?

pmario:
>I didn't notice any test for http.js (I didn't look for them tbh), can you point me to where should I put them?

I think, there are not tests for http.js.. I wanted to know, if you did test it with some other savers. ...

Having a look at the code it looks simple enough _not_ causing side effects.... We'll see, what @Jermolene  says?

Jermolene:
Thanks @OmbraDiFenice, and apologies for the delay. With regards to the modification to http.js, it's worth noting that the core "upload saver" sends data as "multipart/form-data" (see https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/savers/upload.js), can the same technique be used here?

OmbraDiFenice:
Ah interesting, I didn't notice there was something like that. Anyway in my opinion it would be better to change also the "upload saver" and make it use the http.js module, since the code is almost a copy-paste and the http.js purpose is precisely to export an interface to make async http requests...

The difficult part as far as I can see is that in the upload saver there is a "custom" return callback call in a specific case (cfr. https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/savers/upload.js#L60). I'm not sure why is it done like that, but it looks like it's still a "success" case, just calling the callback with different parameters. If that can be generalized then I think it's doable and worth to use http.js instead.

What do you think?

PR #4886, opened 2020-10-09
Simplify ci-push.sh: end script on any error

rmunn:
Use the bash `set -e` option to exit the script immediately if any command returns a non-zero error code, so putting `|| exit 1` at the end of each command is not needed.

As a bonus, this lets us get the actual return code from any failures, instead of the return code always being 1 when something goes wrong. Depending on which command failed, the return code might be meaningful and help with debugging the process.

### DIFF
--- a/bin/ci-push.sh
+++ b/bin/ci-push.sh
@@ -2,19 +2,14 @@
 
 # Push output back to GitHub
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @rmunn and apologies for the delay.

PR #4889, opened 2020-10-10
Make test harness exit with a non-zero code when there's test failure

ento:
Fixes the issue demonstrated by #4869

I believe I sowed the seed for this when I upgraded the jasmine plugin to use Jasmine 3.

What the test harness should've been doing is to call [`Jasmine.prototype.execute()`](https://github.com/jasmine/jasmine-npm/blob/v3.4.0/lib/jasmine.js#L235), which sets up completion callbacks, instead of `env.execute()`, which is a code path shared between both browser and node.

[Here's a check run that failed due to an artificially introduced test failure,](https://github.com/ento/TiddlyWiki5/pull/1/checks?check_run_id=1233895566) based on this branch.

### Review requests
- Would there be any issue in letting Jasmine [call `process.exit()`](https://github.com/jasmine/jasmine-npm/blob/v3.4.0/lib/jasmine.js#L218-L221) in a startup module? Could tests depend on a startup module that was scheduled to run _after_ jasmine's startup module?
  - If it can be problematic, overriding `process.exit` with a function that only sets `process.exitCode` could be a workaround, or making jasmine's startup module run last if possible. (or make it an asynchronous startup module?)

### DIFF
--- a/plugins/tiddlywiki/jasmine/jasmine-plugin.js
+++ b/plugins/tiddlywiki/jasmine/jasmine-plugin.js
@@ -65,7 +65,10 @@ exports.startup = function() {
 		return context.module.exports || contextExports;
 	}
 
-	// Get the core Jasmine exports
+	// Get the core Jasmine exports.
+	// We load 'jasmine-core/jasmine.js' here in order to start with a module
+	// that is shared between browser and Node.js environments. Browser-specific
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @ento that's great, thanks. The approach of setting `process.exitCode` does seem more orderly.

PR #4897, opened 2020-10-20
Fix state tiddlers for action-create-tiddlers examples

pmario:
Fix state tiddlers for action-create-tiddlers examples. Now the buttons work as intended and according to the button description


Jermolene:
Thanks @pmario. We should have some tests to check that the same state tiddlers are always used for the sidebar tabs; I think they changed accidentally a year or two ago.

pmario:
I did create the examples prior to a fix that was released with 5.1.22. The fun fact is, that I did find the problem, but then I forgot to fix my own examples, that used the wrong states.

PR #4899, opened 2020-10-21
Fix bullet points in highlight.js howto

dnebauer:
Change bullet points from `#` to `*`.


Jermolene:
Hi @dnebauer this is an ordered list of instructions that need to be carried out in the specified order, so it is more semantically correct to use an ordered list (`#`) rather than an unordered list (`*`).

pmario:
@Jermolene ... Since the change was rejected, IMO the PR should be closed with a "nofix"

Jermolene:
THank you @dnebauer @pmario

PR #4902, opened 2020-10-23
update TabsMacro.tid with "actions" and "explicitState" attribute

BurningTreeC:
#4875 introduced 2 parameters for the tabs macro:

- actions
- explicitState

they are documented here


Jermolene:
Thanks @BurningTreeC

PR #4906, opened 2020-10-24
Update syncFilter to allow excluding multiple storyLists, historyLists and import tiddlers

saqimtiaz:
Per this [discussion](https://github.com/Jermolene/TiddlyWiki5/commit/707e9d892635d21e2b0535fcbd2902e599a922b7), this PR updates the syncFilter to allow excluding multiple storyLists and historyLists based on prefixes, as well as excluding multiple tiddlers that might be used as import tiddlers.

Issue #4773, opened 2020-07-27
[BUG] plugintiddlers filter doesn't update value for $:/Import


saqimtiaz:
If `$:/Import` already exists when a single file wiki is loaded, calls to `plugintiddlers[]` will always return the same result as exists on startup, even if the tiddler is changed. This manifests in the user being unable to add more tiddlers to the import tiddler. Only deleting `$:/Import`, saving and reloading resolves the problem.

**To Reproduce**
1. In a single file wiki version 5.1.22, drag and drop a tiddler from tiddlywiki.com to trigger the import interface.
2. Without importing or cancelling the import process, save the wiki and reload.
3. Any attempt to import more tiddlers will not be reflected in `$:/Import` in the story, even though editing the tiddler shows the expected JSON data is present.
4. Even deleting the `$:/Import` tiddler and trying to import will show the same tiddler for import as when the wiki was loaded.

When `$:/Import` exists at startup, it has a `pluginInfo` defined in wiki.pluginInfo. Therefore the filter `plugintiddlers[]` always [returns the pluginInfo value from startup, instead of calling getTiddlerDataCached](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/filters/plugintiddlers.js#L21)

Not saving `$:/Import` is one way to resolve this. See also #4772 

Tested in TiddlyDesktop on Windows 10.

pmario:
Strange things can happen ...

saqimtiaz:
Resolved in #4907

PR #4907, opened 2020-10-24
Update filters for single file to match changes to syncFilter

### LINKED ISSUE #4772: [BUG] $:/Import tiddler should not be saved in single file
In single file tiddlywiki's the $:/Import tiddler is saved, whereas on node.js it is not.

Saving this tiddler in single file wikis can lead to accidentally saving data in wikis and increasing the file size, as it is not clear to users that closing this tiddler without clicking cancel, leaves that data from the incompleted import in the wiki. At the very least, closing the $:/Import tiddler with pending data should provide feedback that the data will be saved.

Background: I received a bug report from a colleague I had setup a TW for, that their file size had ballooned to 100+MB and could not be easily shared via email. Investigation showed that at some point they had accidentally dragged a PDF into the wiki, and then closed the $:/Import tiddler and saved and continued using the wiki for several weeks.

saqimtiaz:
This PR updates the saving for the single file version to match the changes to the syncFilter in #4906

Namely:

- use the filter `-[prefix[$:/HistoryList]]` to allow excluding multiple history lists
- exclude tiddlers that match the filter `[status[pending]plugin-type[import]]` to ensure that any tiddlers with an import payload are not saved. This resolves #4772 and #4773 

@pmario  and @Jermolene  please review.

### DIFF
--- a/core/templates/external-js/save-all-external-js.tid
+++ b/core/templates/external-js/save-all-external-js.tid
@@ -2,6 +2,6 @@ title: $:/core/save/all-external-js
 
 \import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
 \define saveTiddlerFilter()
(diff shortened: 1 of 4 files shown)


pmario:
I would like to get this merged.

Jermolene:
Thanks @saqimtiaz

PR #4909, opened 2020-10-25
Add ability to cycle Advanced-search tabs by keyboard ...

BurningTreeC:
... and to open the Advanced-Search panel from within the sidebar search using a keyboard shortcut


pmario:
IMO this PR should be delayed. It uses `explicitState` from tabs macro, but I think we need to finish discussion at PR #4454 first

BurningTreeC:
@pmario I think this PR can be merged as @Jermolene seems to prefer the `explicitState` solution

Jermolene:
Just a thought, but these keyboard shortcuts are fairly undiscoverable as things stand. There's a nice feature on the iPad where you can hold down a modifier key to see a HUD of all the key combinations that can be used. I wonder if we could do something similar: a global message handler that collects the active keyboard shortcuts and displays the HUD. We wouldn't be able to trigger it with a modifier key, we'd have to use <kbd>F1</kbd> or <kbd>ctrl-?</kbd>. (It would be tricky to use a button to trigger it because the act of clicking the button will move the focus away).

saqimtiaz:
@Jermolene that's an interesting idea and definitely worth exploring. It may also be useful to consider separating the keyboard shortcuts in the control panel into groups/categories.

PR #4912, opened 2020-10-25
Get "New Release Banner" from TW5-com edition

pmario:
Get "New Release Banner" from TW5-com edition with `tiddlywiki.files` configuration.

This will only be active, after your next "global" build!


Jermolene:
Hi @pmario I don't think it necessarily matters, but it's worth noting that this patch will require us in future to maintain the banner as a PNG file; in fact, it has switched between JPG and PNG format over recent releases.

PR #4913, opened 2020-10-26
Fix #4910 - "Editor link dropdown is broken in pre-release."

### LINKED ISSUE #4910: [BUG] Editor link dropdown is broken in pre-release.
"https://tiddlywiki.com/prerelease/ -> New tiddler -> Create wikitext link (in toolbar) -> search for "HelloThere", it doesn't give me search results in either Chrome or Firefox." via [GG](https://groups.google.com/forum/?oldui=1#!topic/tiddlywiki/HU7n7NPl9wM)

Presumably this is due to #4875, as I did test the toolbar link dropdown after we introduced the keyboard-driven-macro, including extending it to be usable via a keyboard only workflow.

@BurningTreeC  I suspect configTiddler needs to be defined in [$:/core/ui/EditorToolbar/link-dropdown](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/EditorToolbar/link-dropdown.tid) before transcluding $:/core/ui/SearchResults:
https://github.com/Jermolene/TiddlyWiki5/pull/4875/files#diff-b4f2091e7fe5d0fa2d16dc46b6b88ea843a2702f573923959f2fedc6fbf6e7e0R11

Note that there is also an opportunity here to extend the link dropdown to be usable with an entirely keyboard based workflow.
I did prototype this a while back and the user experience is very good:
https://saqimtiaz.github.io/sq-tw/floats.html#%24%3A%2Fcore%2Fui%2FEditorToolbar%2Flink-dropdown:%5B%5BDraft%20of%20%27link-autocomplete%27%5D%5D%20%24%3A%2Fcore%2Fui%2FEditorToolbar%2Flink-dropdown

BurningTreeC:
This PR fixes #4910 

The Editor Link-dropdown didn't work anymore because I missed that it also uses the default SearchResultList

Some more adjustments where needed in the `keyboard-driven-input macro` and the `SidebarSegments/search` tiddler, because - if using more search-tabs via $:/tags/SearchResults changing those tabs within the Link-dropdown caused a redraw of the sidebar-search input which caused it to steal focus from the Link-dropdown.

Now the Link-dropdown would also be prepared to be keyboard-driven-input-ready

### DIFF
--- a/core/ui/EditorToolbar/link-dropdown.tid
+++ b/core/ui/EditorToolbar/link-dropdown.tid
@@ -37,7 +37,11 @@ title: $:/core/ui/EditorToolbar/link-dropdown
 
 <$linkcatcher actions=<<add-link-actions>> to=<<linkTiddler>>>
 
+<$vars userInput={{{ [<searchTiddler>get[text]] }}} configTiddler={{{ [[$:/state/search/currentTab]!is[missing]get[text]] ~[{$:/config/SearchResults/Default}] }}}>
+
 {{$:/core/ui/SearchResults}}
+
+</$vars>
 
 </$linkcatcher>
 
--- a/core/ui/SideBarSegments/search.tid
+++ b/core/ui/SideBarSegments/search.tid
@@ -19,17 +19,17 @@ tags: $:/tags/SideBarSegment
 
 \define search-results-list()
 \whitespace trim
-<$set name="userInput" value={{$(searchTiddler)$}}>
+<$vars userInput={{$(searchTiddler)$}} configTiddler={{{ [[$:/state/search/currentTab]!is[missing]get[text]] ~[{$:/config/SearchResults/Default}] }}}>
 <$list filter="[<userInput>minlength{$:/config/Search/MinLength}limit[1]]" emptyMessage="""<div class="tc-search-results">{{$:/language/Search/Search/TooShort}}</div>""" variable="listItem">
 
-<$tiddler tiddler="$(configTiddler)$">
+<$tiddler tiddler=<<configTiddler>>>
 
 {{$:/core/ui/SearchResults}}
 
 </$tiddler>
 
 </$list>
-</$set>
+</$vars>
 \end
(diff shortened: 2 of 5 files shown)


saqimtiaz:
@BurningTreeC have you tested the tag keyboard completion with this patch? I am wondering how that is impacted by the change to the macro parameters for keyboard-driven-input.

Also I am wondering if might be better not to change the macro call for keyboard-driven-input.

Being able to pass filters directly seems more user friendly for those wanting to re-use the macro.
Instead we could use filtered parameters to get the filters from the config tiddler and pass it to the keyboard-driven-input macro.

BurningTreeC:
@saqimtiaz - you're right, I have to change the tag-picker macro, too

saqimtiaz:
@BurningTreeC what do you think about not changing the kbd macrocall? It seems like an onerous requirement for those wanting to re-use the macro that they have to create a config tiddler to hold the filters.

BurningTreeC:
>Also I am wondering if it might be better not to change the macro call for keyboard-driven-input.

>what do you think about not changing the kbd macrocall? It seems like an onerous requirement for those wanting to re-use the macro that they have to create a config tiddler to hold the filters.

@saqimtiaz, the change is needed because both the link-dropdown and the sidebar-search use the same configTiddler which changes when there's a second (...) search-tab and we click on it (or change it by kb-shortcut). This change causes the vars widget that surrounds the kb-driven-input macro to refresh - and focus jumps to the sidebar search, which closes the link-dropdown. To resolve this problem, I need to get the configTiddler at the last position possible within the kb-driven-input macro (because that tiddler can change and cause the redraw)

BurningTreeC:
@Jermolene @saqimtiaz this PR is still missing a little adjustment

BurningTreeC:
@Jermolene @saqimtiaz , now this would be ready to go

Jermolene:
Thanks @BurningTreeC

PR #2996, opened 2017-10-16
Add meta options for tiddlywiki.files

sukima:
Fixes #2906

Plain text .txt files or .md files, loaded via a tiddlywiki.files declaration load in the TiddlyWiki just fine. However, editing them is problematic. They are saved to the wrong file name.

With a subdir in my tiddlers directory called "notes", containing a file "mynote.txt", editing results in a file "mynote 1.txt" and "mynote 1txt.meta" whereas I would expect the original "mynote.txt" to be updated and a .meta file created. This behavior leads to duplication of tiddler files and is rather confusing. If the intent was that files loaded in this manner should not be editable, then we should find a way to set them as read only.

Digging into the code in boot.js reveals that .txt or .md files that do not have .meta files and are loaded via a tiddlywiki.files declaration, are never added to $tw.boot.files (as they don't have their filepath set).

With this commit the files are added to $tw.boot.files and on editing are saved to the accurate and original filename, along with a .meta file for non .tid files.

Original Author: https://github.com/saqimtiaz
Original patch: https://github.com/rubaboo/TiddlyWiki5/commit/50fbe23bc97b5c87d9d544506eed4d9df9494700.patch

PR #4914, opened 2020-10-26
Extend tiddlywiki.files to allow saving changes to original location

### LINKED ISSUE #2906: Files loaded via tiddlywiki.files save to incorrect filename
Plain text .txt files or .md files, loaded via a tiddlywiki.files declaration load in the TiddlyWiki just fine. However, editing them is problematic. They are saved to the wrong file name.

With a subdir in my tiddlers directory called "notes", containing a file "mynote.txt", editing results in a file "mynote 1.txt" and "mynote 1txt.meta" whereas I would expect the original "mynote.txt" to be updated and a .meta file created. This behavior leads to duplication of tiddler files and is rather confusing. If the intent was that files loaded in this manner should not be editable, then we should find a way to set them as read only.

Digging into the code in boot.js reveals that .txt or .md files that do not have .meta files and are loaded via a tiddlywiki.files declaration, are never added to $tw.boot.files [(as they don't have their filepath set).](https://github.com/Jermolene/TiddlyWiki5/blob/a4035d34247e2f27efc9cb6228432deea21d73a3/boot/boot.js#L1786)

If we modify line line [1594](https://github.com/Jermolene/TiddlyWiki5/blob/master/boot/boot.js#L1594) in boot.js from 

`tiddlers.push({tiddlers: fileTiddlers});`

to

`tiddlers.push({filepath: pathname, hasMetaFile: !!metadata && !isTiddlerFile, tiddlers: fileTiddlers});`

the files are added to $tw.boot.files and on editing are saved to the accurate and original filename, along with a .meta file for non .tid files.

I propose that this should be the default way of handling .txt and .md files that are loaded via tiddlywiki.files

saqimtiaz:
This PR extends the directory specification for `tiddlywiki.files` files to allow optionally saving changes back to the original location by setting `isEditableFile` to true.

More background at #2996 

Closes #2906 and #2996


Jermolene:
Thanks @saqimtiaz

PR #4915, opened 2020-10-26
Named filter run prefixes

saqimtiaz:
This PR extends the filter run syntax to allow named filter run prefixes.

- `:all` is equivalent to `=[run]`
- `:else` is equivalent to `~[run]`
- `:remove` is equivalent to `-[run]`
- `:intersection` is equivalent to `+[run]`

A new `module-type` has been introduced called `filterprefix`, this makes the filter run prefixes extensible via plugins.


saqimtiaz:
I'd appreciate overall feedback on the **code pattern** as well as:

- the module-type name `filterprefix`. They are actually filter run prefixes but that seems fairly long.
- the `intersection` prefix needs an easier to understand and shorter name.
- should the default behaviour of no prefix also be encapsulated in a `filterprefix` module-type tiddler? If so, what should it be called? At the moment the code for the no prefix case is still inline.

rmunn:
For the `intersection` prefix, how about `and`? And for `remove`, I kind of like `except`. Using `all` for `=` is good since that reminds people that deduplication is turned off for the `=` prefix. So my suggestion would be:

- `:all` is equivalent to `=[run]`
- `:and` is equivalent to `+[run]`
- `:or` is equivelent to `[run]` (no prefix)
- `:else` is equivalent to `~[run]`
- `:except` is equivalent to `-[run]`

Jermolene:
> For the `intersection` prefix, how about `and`? And for `remove`, I kind of like `except`. Using `all` for `=` is good since that reminds people that deduplication is turned off for the `=` prefix. So my suggestion would be:
> 
> * `:all` is equivalent to `=[run]`
> * `:and` is equivalent to `+[run]`
> * `:or` is equivelent to `[run]` (no prefix)
> * `:else` is equivalent to `~[run]`
> * `:except` is equivalent to `-[run]`

I like @rmunn's proposals for the names.

saqimtiaz:
@Jermolene regarding the module type, should we go with `filterrunprefix` or just `filterprefix` ?

Jermolene:
Thanks @saqimtiaz the code looks good.

> * the module-type name `filterprefix`. They are actually filter run prefixes but that seems fairly long.

It is unfortunate that we're using the "prefix" word again when it is so overloaded within the filter processing code.

I don't think it's much of a problem if the module-type is lengthy. I was going to propose "filter run modifiers" but "filterrunmodifiers"  is easy to read as "filter unmodifiers". So maybe "filter run type"? 

> * should the default behaviour of no prefix also be encapsulated in a `filterprefix` module-type tiddler? If so, what should it be called? At the moment the code for the no prefix case is still inline.

That would be good, yes. Perhaps we could call it "accumulate"?

saqimtiaz:
@Jermolene I agree the `prefix` word is overloaded already, in particular there could be confusion with filter operator prefixes. I'll use `filterruntype` as the module type. 

Do note that we refer to `+, - ,  ~` as filter run prefixes in the [documentation](https://tiddlywiki.com/#Filter%20Expression), and changing that now would be confusing.

I will update documentation in a separate PR, introducing the new module-type and how it relates to `+, - ,  ~` filter run prefixes.

rmunn:
Do the prefix names (`:or`, `:else`, and so on) have to be the same as the name of the tiddler module that defines them? If not, then I think "union" and "intersection" might be good names for the Javascript functions (and modules) that define the no-prefix and `+`-prefix behaviors that I suggested `:or` and `:and` for. Use the set-related language (union, intersection, etc.) for the tiddler modules or Javascript functions where the code lives, and use the logic-based language (or, and, etc.) for what we expose to the user. I suggest this because a file named "or.js" or "and.js" might be a little confusing to people maintaining the code later. OTOH, that would require a lookup table somewhere, and it's probably simpler for the `:or` filter prefix to be defined in a `core/modules/filterprefixes/or.js` file.

Jermolene:
Hi @saqimtiaz I was imagining that we would change the documentation, but you're right that is probably too confusing. In which case we're back to using "filterrunprefix" as the module-type.

@rmunn modules like this would normally export a single named property containing the function (or more complex object) representing the operation. There's no enforcement of the relationship between names, but we do use consistent naming to make things easier to read. In this case, I think it would make sense to consistently use the user-exposed vocabulary.

saqimtiaz:
@Jermolene @rmunn  this should be good to go.

- I've renamed the module-type to `filterrunprefix`
- All modules have been renamed in accordance with @rmunn's suggestions. Note that the default case of no prefix is also exposed as `:or `
- I've added an error message to handle unknown filter run prefixes.

Once this is merged, I will work on PR's for:
1.  Updated documentation about filter run prefixes and the new module-type
2. A `filter` filter run prefix

Jermolene:
Excellent thanks @saqimtiaz

PR #4918, opened 2020-10-27
Add new filter run prefix: filter

### LINKED ISSUE #4888: [IDEA] Consider extending filter run prefixes
Should we extend the filter run prefixes with one that behaves similar to the new [filter](https://tiddlywiki.com/prerelease/#filter%20Operator) filter operator?

I've been running a couple of custom extensions to the filter run prefixes (`=, - , + , ~` ) that add convenience, one of which `: ` behaves as the new `filter` operator does. However using a filter run prefix allows for more brevity in inline code. 

I realize the balance between utility and not wishing to inflate the core size, therefore this is posted as an issue and not a PR. I do have working code though it might need to be reviewed from the point of view of performance.

The following would be equivalent:

```
\define larger-than-1k() [get[text]length[]compare:integer:gteq[1000]]

{{{ [tag[HelloThere]filter<larger-than-1k>] }}}
```
and
```
{{{ [tag[HelloThere]] :[get[text]length[]compare:integer:gteq[1000]] }}}
```

An excerpt of the relevant part of the code:
```
case ":":
	return function(results,source,widget) {
		if(results.length > 0) {
			var toRemove = [];
			$tw.utils.each(results,function(result) {
				var filtered = operationSubFunction(self.makeTiddlerIterator([result]),widget);
				if(filtered.length === 0) {
					toRemove.push(result);
				}
			});
			$tw.utils.removeArrayEntries(results,toRemove);
		}
	}
```
Are there any performance considerations to be take into account with this approach?

saqimtiaz:
Adds a new filter run prefix `filter`, that behaves akin to the new `filter` operator as described at #4888 

The following would be equivalent:

`\define larger-than-1k() [get[text]length[]compare:integer:gteq[1000]]`

`{{{ [tag[HelloThere]filter<larger-than-1k>] }}}`

and

`{{{ [tag[HelloThere]] :filter[get[text]length[]compare:integer:gteq[1000]] }}}`

Once this is merged I'll make another PR for updating documentation.

Fixes #4888


pmario:
@saqimtiaz .. IMO some text was eaten by github here:??? {{{ [tag[HelloThere]filter] }}} ... should it be `subfilter`?

saqimtiaz:
@pmario yes it was a github formatting issue in my post, now fixed.

Jermolene:
Thanks @saqimtiaz

PR #4919, opened 2020-10-27
Make link-dropdown navigatable by keyboard

BurningTreeC:
This PR makes the EditorToolbar link-dropdown navigatable by keyboard

Up/Down chooses a list-item, alt-Left/alt-Right cycles through (eventual) search tabs


BurningTreeC:
Just an idea: We could also add a different action when hitting alt-Enter ... possibly the transclusion of the selected list-item

Jermolene:
Thanks @BurningTreeC

PR #4920, opened 2020-10-27
Docs for :filter filter run prefix.

saqimtiaz:
Documentation for #4918


Jermolene:
Thanks @saqimtiaz these are terrific improvements, and very happy with the collaboration that led to them.

saqimtiaz:
@Jermolene likewise! I think the synthesis of our ideas has led to a significant extension to the filter syntax.

Jermolene:
(I was a bit embarrassed by the filter syntax at first, feeling that it was too complex and confusingly different from anything else. And then in 2018 Joe Armstrong very graciously told me that it was OK because I'd (re-)invented monads, exciting for me because I'd never previously understood the term).

saqimtiaz:
@Jermolene I have only just now wrapped my head around the term monad, so thank you. 

To be fair I wasn't too convinced when filters were introduced in TWC, but as they stand now they are extremely powerful and flexible.

kookma:
@saqimtiaz Filter operation needs more examples to be included in TW documentation or even separate document to reveal the power of filter operation in Tiddlywiki!

saqimtiaz:
@kookma this PR adds docs for a specific extension of filter syntax.

As such it isn't the best place to discuss overall documentation for filters. A PR towards this end would be most welcome though.

PR #4921, opened 2020-10-27
Add ability to navigate fieldname-dropdown by keyboard

BurningTreeC:
This PR adds the ability to navigate the fieldname dropdown using the Up/Down arrow keys and Escape to

a) clear the input field if not empty
b) cancel editing the tiddler if the input field is empty

In this case there's need for a separate configuration tiddler that stores the two filters in the `first-search-filter` and `second-search-filter` fields


Jermolene:
Thanks @BurningTreeC

PR #4922, opened 2020-10-28
Fix menubar-plugin search

BurningTreeC:
Because of the recent change in the `keyboard-driven-input macro` this fix is needed for the menubar-search-dropdown to work correctly


Jermolene:
Thanks @BurningTreeC

Issue #4694, opened 2020-06-03
[IDEA] - adding some classesto replace whitespace with spans


BurningTreeC:
There are many places where `&nbsp;` or just ` ` are used for styling of the UI and I'd propose replacing those with `spans` with classes like `tc-span-margin-05`, `tc-span-margin-025`, `tc-span-margin-05-left` for example

pmario:
+1 in principle. 

But if there are `&nbsp;`, there should be html elements that follow. .. If so, there shouldn't be extra spans, if we can use existing elements.

BurningTreeC:
> But if there are &nbsp;, there should be html elements that follow. .. If so, there shouldn't be extra spans, if we can use existing elements.

Yes I'm thinking the same

Jermolene:
Apologies for the delay, I agree this would be a worthwhile improvement.

BurningTreeC:
@Jermolene - so I'll try to come up with a PR in time

BurningTreeC:
@Jermolene, what names would you prefer for the classes? I was thinking about classes that add a margin-right, a margin-left, and both a margin-right and margin-left of 0.5em (which should replace &nbsp;) ... we could call the classes like "tc-margin-05em, tc-margin-05-em-right, tc-margin-05em-left and so on ...?

pmario:
>we could call the classes like "tc-margin-05em, tc-margin-05-em-right, tc-margin-05em-left and so on ...?

IMO you shouldn't use the "05em" part in the name. If we want to change it to 0.4em or rem, the name would be invalidated. 

tc-small-margin-xxxxx, tc-tiny-margin-xxxx  would be 2 possibilities I can think of atm. or even 

tc-push-left and tc-push-right may be more generic. You could use both margin and padding to push "content" out of the way.

saqimtiaz:
Maybe tc-small-gap rather than push?

pmario:
>Maybe tc-small-gap rather than push?

hmmm, gap doesn't have a direction, which should be part of the class so tc-gap-left, tc-gap-right?

BurningTreeC:
Perhaps tc-small-gap for margin-left+margin-right, tc-small-gap-left and tc-small-gap-right?

saqimtiaz:
I am wondering about order, tc-gap-left-small?

BurningTreeC:
I like tc-gap-small, tc-gap-left-small etc.

PR #4923, opened 2020-10-28
Add .tc-small-gap (+left / right) and .tc-big-gap...

BurningTreeC:
(+left / right) classes that define margins of .5em / 1em

See #4694


Jermolene:
Is this ready to merge?

BurningTreeC:
@Jermolene, yes this is ready to merge. I'll add another PR that makes use of those classes later this day

Jermolene:
Thanks @BurningTreeC

PR #4925, opened 2020-10-28
Make AdvancedSearch/Standard, AdvancedSearch/System and AdvancedSearch/Shadows navigatable by keyboard

BurningTreeC:
This PR makes the AdvancedSearch/System, AdvancedSearch/Standard and AdvancedSearch/Shadows tabs navigatable using Up/Down arrow keys, Enter to navigate to the selected item, alt-Enter to edit the selected item and Escape to clear the input field

I'm trying to make this also possible for the Filter tab but that appears to need a bit more investigation and work

### DIFF
--- a/core/ui/AdvancedSearch/Shadows.tid
+++ b/core/ui/AdvancedSearch/Shadows.tid
@@ -1,40 +1,53 @@
 title: $:/core/ui/AdvancedSearch/Shadows
 tags: $:/tags/AdvancedSearch
 caption: {{$:/language/Search/Shadows/Caption}}
+first-search-filter: [all[shadows]search<userInput>sort[title]limit[250]] -[[$:/temp/advancedsearch]] -[[$:/temp/advancedsearch/input]]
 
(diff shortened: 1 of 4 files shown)


Jermolene:
Thanks @BurningTreeC does it make sense to wait before merging?

BurningTreeC:
Hi @Jermolene 

>does it make sense to wait before merging?

No, waiting doesn't make sense, the Filter tab would be a separate PR if I'll be able to wrap my head around it

Jermolene:
Thanks @BurningTreeC

kookma:
I have noted opening $:/AdvancedSearch, curse is blinking in the inputbox (to input the search term)
then using keyboard up/down arrow in the empty inputbox brings and cycle through some tidddlers title.
Is this a bug or it is designed to cycle through possible tiddlers title?

BurningTreeC:
Hi @kookma , this is a bug that will be resolved soon. Thanks for reporting!

PR #4926, opened 2020-10-28
Replace various &nbsp; with tc-small-gap classes

BurningTreeC:
This PR replaces `&nbsp;` in different places with the newly introduced `tc-small-gap`, `tc-small-gap-left` and `tc-small-gap-right` classes

### DIFF
--- a/core/ui/EditTemplate/fields.tid
+++ b/core/ui/EditTemplate/fields.tid
@@ -21,14 +21,6 @@ $:/config/EditTemplateFields/Visibility/$(currentField)$
 <$action-sendmessage $message="tm-focus-selector" $param=<<current-tiddler-new-field-selector>>/>
 \end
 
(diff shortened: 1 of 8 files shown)


core/ui/EditTemplate/fields.tid
```diff
@@ -76,57 +68,46 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 
 <$fieldmangler>
 <div class="tc-edit-field-add">
-<em class="tc-edit">
-<<lingo Fields/Add/Prompt>>&nbsp;&nbsp;
+<em class="tc-edit tc-big-gap-right">
```
pmario, line 71: This should have been a `tc-small-gap-right` see: default edit-template has a visual problem #5424

@Jermolene .. The types input also got a "big-gap", which doesn't create a problem there. Should we also change it to "small"

Issue #4473, opened 2020-03-06
Improve page layout


Jermolene:
UPDATE 6th June 2021:

I've now implemented this new layout as part of the work on the https://twpub-tools.org/ project. I plan to work on it more there, working towards a proposal for the core:

https://github.com/TWPUB/TWPUB-Tools

---

We have a number of long-standing problems with the layout of the story river, sidebar, position sticky headings and any toolbars:

* The relationships between the window width, story river width and the sidebar width is confusing, and hard to customise
* The scrollbars for the story river and the sidebar are both pushed to the far right of the screen, and are hard to distinguish
* #4471 - We can only use a fixed offset to move sticky titles below a top menu, even with dynamically sized top menu bars (like the new menubar plugin)
* We have to use hacks to make sure that scrolling to the top of a tiddler takes into account any top menu bar

the current layout is the result of overlaying three independent components: the story river, the sidebar, and the new menu bar. Their layout is entirely independent: we just arrange margins and padding to try to stop them blotting each other out. That's the fundamental problem with the current layout that we keep running into: whenever we change the layout of one of those three page components we have to meticulously adjust the layout of the others.

It is proposed to explore an alternative layout based on CSS flexbox. It's entirely fluid, with the page components flowing to take up the available space. Flexbox gives us a lot of flexibility for how those components react to the screen size changing too.

As a first step, here's a mockup of TW5's layout built with flexbox: https://codepen.io/jermolene/pen/mdJBqGz

<img width="1272" alt="image" src="https://user-images.githubusercontent.com/174761/76069465-ceed3880-5f8a-11ea-890c-3dbcb474a4f4.png">

The implementation is pleasantly clean; there are no magic numbers, and everything lines up without having to do anything special. However, the required HTML structure is not compatible with the present page template.

The flexbox solution is adaptive; if one inserts another `.toolbar` beneath the closing `</main>` tag then one gets a toolbar fixed to the bottom of the window, with everything smoothly adjusting to make room. Additional sidebars can be added in the same way. CSS Grid requires the entire layout to specified in one place, and doesn't have that same adaptability. We need the adaptability because of the way that TW5 works: we want users to be able to add another horizontal bar by adding a new tiddler with the tag `$:/tags/HorizBar` (say), without having to fiddle with CSS as well. That's rather the point of this change, in fact 

[...]

PR #4928, opened 2020-10-28
Import rename

saqimtiaz:
This PR extends the import listing UI to allow renaming tiddlers, as demoed earlier in the summer to @Jermolene as part of one of my plugins.

[Demo.](https://saqimtiaz.github.io/sq-tw/temp/import-rename.html)

Note that two CSS classes have been added.


saqimtiaz:
@BurningTreeC considering your recent work with CSS adjustments, I was hoping you could take a look at the two CSS classes I've added and see if there is a better approach.

Jermolene:
This looks good, thanks @saqimtiaz

saqimtiaz:
@Jermolene thank you. I've taken it out of draft.

Jermolene:
Thanks @saqimtiaz. Please feel free to tweak the CSS @BurningTreeC

BurningTreeC:
@Jermolene @saqimtiaz - for me everything is fine here

saqimtiaz:
The main room for improvement is probably using `display:flex` to let the input grow to max avaialable width and also align the rename buttons, but as I understand it we don't use `display:flex` for the core.

Jermolene:
Hi @saqimtiaz we do use display: flex for vertical tabs, and haven't had any complaints about them, so I guess we can go ahead. The planned new layout stuff in #4473 also uses display: flex, come to think about it.

kookma:
@saqimtiaz 
Is it possible to show a warning if the tiddler already exist in the wiki? At least show the title in bold or red?

kookma:
@saqimtiaz [IDEA]
It would be great to add extra tag on import!

saqimtiaz:
@kookma The hold up with warning about existing tiddlers is mostly related to UI and presentation. There isn't a lot of space to work with. One option is a text warning in the status column as shown below:

image

Alternatively, if we used colours/icons we would need to provide a legend and should then also use the same for outdated/blocked plugins, etc.

PR #4930, opened 2020-10-28
Fix keyboard-driven fieldname dropdown

BurningTreeC:
I've accidentally overwritten the correct file when adding the tc-small-gap classes ... SORRY!


Jermolene:
No worries, thanks @BurningTreeC

PR #4932, opened 2020-10-28
Make AdvancedSearch/Filter navigatable by keyboard

BurningTreeC:
This PR makes the AdvancedSearch/Filter Tab navigatable like the other three Tabs (Standard, System and Shadows)


BurningTreeC:
@Jermolene - this is ready to get merged

BurningTreeC:
@Jermolene - no this isn't yet ready. I'll make it ready this evening

Jermolene:
Hi @BurningTreeC it looks like the state tiddler used for storing the search text is still being changed. Is that necessary?

BurningTreeC:
Hi @Jermolene ,

>it looks like the state tiddler used for storing the search text is still being changed. Is that necessary?

If you refer to `$:/temp/advancedsearch/input` then yes, it's a necessary change

But if this gets merged I can have another look at all the AdvancedSearch search-tabs and see if I can make the searchTiddler being `$:/temp/advancedsearch` again

Jermolene:
Hi @BurningTreeC I'm afraid that that is a breaking change: we have dependencies on the existing title elsewhere in the core, and can expect end users to do so too. Can we address it before merging?

BurningTreeC:
@Jermolene - yes I'll make the necessary changes so that it uses `$:/temp/advancedsearch`

BurningTreeC:
@Jermolene - this PR is now ready. I've changed it so that it uses `$:/temp/advancedsearch` again as the main filter-tiddler/search-tiddler (also for the other AdvancedSearch search-tabs)

Jermolene:
Thanks @BurningTreeC

BurningTreeC:
Thanks @Jermolene - I'd like to create a follow-up PR but would like to hear your opinion about a detail:

The AdvancedSearch/Filter Tab when pressing <kbd>Escape</kbd> (after some items were selected by up/down) first resets the input field to the filter. Then when pressing <kbd>Escape</kbd> again, it clears the input field.

I'd like to make this behavior the default for all keyboard-driven-inputs... What do you think?

Jermolene:
Hi @BurningTreeC I think that sounds reasonable. Would the second time escape is pressed lead to the dropdown being removed too?

BurningTreeC:
>Would the second time escape is pressed lead to the dropdown being removed too?

Yes, @Jermolene

Jermolene:
Hi @BurningTreeC that's great, and worth doing I think.

PR #4936, opened 2020-10-29
Improve documentation for reduce operator

### LINKED ISSUE #4884: Docs for reduce operator should mention that it always produces a value even with empty input
I like the new `filter` and `reduce` operators that were just added. Looking through the code of the `reduce` operator, I see that it will always produce a value, even when the input was empty (in which case the result of the `reduce` operator is just the value of its accumulator, which is probably an empty string unless the user specified an initial value). This is the right way to implement it, but IMHO it might be a little surprising for some people, who might be expecting that an empty input will produce an empty output. (An empty input should, and does, produce an empty output in the case of `filter`, but in the case of `reduce` it should always produce an output of length 1 no matter what length the input is).

But since the "empty input produces non-empty output" functionality of `reduce` might surprise some people, I think it's a good idea to document that fact in the `reduce` docs, and add an example showing an empty input. (And maybe add a unit test for this case as well, to ensure that if the behavior ever changes, the unit test failure will cause someone to think about whether that changed behavior is the right one).

It would also be a good idea for the `reduce` docs to show how to set the initial value of the accumulator, which might be particularly useful in the case of the `[tag[shopping]reduce<add-price>]` example. Show what happens with `[tag[non-existent]reduce<add-price>]` (result: empty string) and compare to `[tag[non-existent]reduce:0<add-price>]` (result: "0"). There's no good way, AFAICT, to examine the filter expression passed to `reduce` to figure out whether an empty result should be returned as an empty string or the number 0, so defaulting to empty string is best, and the user can choose to let the value be 0 instead by using the suffix correctly. But I think this will not be obvious to anyone just reading the TiddlyWiki documentation and not reading the code, so it's probably best to give an example or two of empty input and show how you can get the output values you want by setting the accumulator's starting value (via the suffix) correctly.

I plan to work on this soon (next week, probably) and will submit a PR soon to add those examples to the `reduce` docs. If you (anyone who reads this) have already written such examples, please comment on this issue so I don't duplicate your work. :-)

rmunn:
Fixes #4884.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/reduce Operator (Examples).tid
+++ b/editions/tw5.com/tiddlers/filters/examples/reduce Operator (Examples).tid
@@ -6,14 +6,19 @@ type: text/vnd.tiddlywiki
 
 \define add-price() [get[price]multiply{!!quantity}add<accumulator>]
 \define num-items() [get[quantity]add<accumulator>]
+\define join-with-commas() [<index>compare:number:gt[0]then<accumulator>addsuffix[, ]addsuffix<currentTiddler>else<currentTiddler>]
 \define display-variable(name)
 ''<$text text=<<__name__>>/>'': <code><$text text={{{ [<__name__>getvariable[]] }}}/></code>
+\end
+\define reduce-tip()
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @rmunn

PR #4937, opened 2020-10-29
Improvements for Import UI

saqimtiaz:
This PR improves upon #4928 which added the ability to rename tiddlers before import.
Changes include:

- use tiddler fields in the import tiddler instead of state tiddlers to track state, to avoid state tiddlers that would get saved in single file version.
- bug fix where the original name was used for a renamed tiddler in the preview popup.
- Change in UI where the input field to rename a tiddler appears in the next table row, providing greater space for longer titles.
- Introduced utility flexbox classes and used them to style the rename UI.

[Demo](https://saqimtiaz.github.io/sq-tw/temp/import-rename.html)


Jermolene:
Thanks @saqimtiaz

PR #4938, opened 2020-10-29
Do not save temp tiddlers in single file

### LINKED ISSUE #4933: [BUG] $:/temp tiddlers should not be saved in single file
As the [documentation ](https://tiddlywiki.com/#Naming%20of%20System%20Tiddlers )suggests, these are temporary tiddlers and should not be saved.

However, they are saved in the single file version and accumulate over time leading to clutter. In particular, the plugin library generates a lot of `$:/temp/RemoteAssetInfo` tiddlers that end up getting saved.

Suggest we modify $:/core/save/all to exclude all `[prefix[$:/temp/]]` tiddlers.

saqimtiaz:
Fixes #4933
Fixes #4938 

Temp tiddlers should not be saved in single file as they accumulate over time. Furthermore their value as temporary storage to plugin writers is negated by the fact that they are saved in single file but not on node.js.

As the [documentation ](https://tiddlywiki.com/#Naming%20of%20System%20Tiddlers ) suggests, these are temporary tiddlers and should not be saved.

### DIFF
--- a/core/templates/external-js/save-all-external-js.tid
+++ b/core/templates/external-js/save-all-external-js.tid
@@ -2,6 +2,6 @@ title: $:/core/save/all-external-js
 
 \import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
 \define saveTiddlerFilter()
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz

PR #4939, opened 2020-10-29
Fix AdvancedSearch Standard, System and Shadows up/down behavior

BurningTreeC:
This fixes the problem that, when pressing up/down in an EMPTY search field, the keyboard-driven-input macro gave some search results


Jermolene:
Thanks @BurningTreeC

PR #4943, opened 2020-10-29
Avoid editing the wrong tiddlers when input fields are empty

BurningTreeC:
This PR adds a test if input fields are empty before sending the tm-edit-tiddler message


Jermolene:
Thanks @BurningTreeC

PR #4947, opened 2020-10-30
Fix bug when location hash contains a #

rmunn:
See #3811 for motivation, especially for why location hashes starting with # are special-cased to be ignored.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -267,8 +267,16 @@ $tw.utils.htmlDecode = function(s) {
 Get the browser location.hash. We don't use location.hash because of the way that Firefox auto-urldecodes it (see http://stackoverflow.com/questions/1703552/encoding-of-window-location-hash)
 */
 $tw.utils.getLocationHash = function() {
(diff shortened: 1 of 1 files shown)


rmunn:
Note that #3811 will *not* be automatically closed when this PR is merged. I did that on purpose, as #3811 contains discussion of the useful behavior of double hashes which still needs to be documented. Once that behavior is documented (by someone else as I'm not the right person to do that), then #3811 can be safely closed.

rmunn:
Anyone who wants to document the double-hash behavior will want to look at https://github.com/Jermolene/TiddlyWiki5/issues/3811#issuecomment-719306525 and the files attached there, as a good start towards the documentation which may (or may not) need to be updated for TW 5.1.23.

Jermolene:
Thanks @rmunn some docs would certainly be appreciated if you get a chance.

rmunn:
Docs PR: #5149. See also #5150 for some "interesting" interactions with the menubar.

PR #4948, opened 2020-10-30
Add string operator enlist-input[]

saqimtiaz:
While [title lists](https://tiddlywiki.com/#Title%20List) are very common in TiddlyWiki, we are missing a filter operator that allows it's input to be parsed as a title list.

For example consider `[{!!parent}get[children]]` where `children` is a field holding a title list. 
Currently this requires two steps to operate on the list of children, first getting `children` and then `enlist<children>`
```
<$vars children={{{ [{!!parent}get[children]] }}}/>
<$list filter="[enlist<children>]">
<!-- do something -->
</$list>
</$vars>
```

With this operator the above would be:

`[{!!parent}get[children]aslist[]]`

I've considered extending `enlist[]` to operate on its input when no operand is present, but that might be confusing for users.

I'm open to better suggestions for a name:
- `parselist[]`
- `tolist[]`
- `titlelist[]`
- ?

See [this table](https://tiddlywiki.com/#Filter%20Operators) for names of existing string operators.

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -56,12 +56,12 @@ exports.trim = function(source,operator,options) {
 	return result;
 };
 
-// makeStringBinaryOperator(
-// 	function(a) {return [$tw.utils.trim(a)];}
-// );
-
 exports.split = makeStringBinaryOperator(
 	function(a,b) {return ("" + a).split(b);}
+);
+
+exports.titlelist = makeStringBinaryOperator(
+	function(a) {return $tw.utils.parseStringArray("" + a);}
 );
 
 exports.join = makeStringReducingOperator(
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -60,7 +60,7 @@ exports.split = makeStringBinaryOperator(
 	function(a,b) {return ("" + a).split(b);}
 );
 
-exports.aslist = makeStringBinaryOperator(
+exports["enlist-input"] = makeStringBinaryOperator(
 	function(a) {return $tw.utils.parseStringArray("" + a);}
 );
 
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -60,7 +60,7 @@ exports.split = makeStringBinaryOperator(
 	function(a,b) {return ("" + a).split(b);}
 );
 
-exports.titlelist = makeStringBinaryOperator(
+exports.aslist = makeStringBinaryOperator(
 	function(a) {return $tw.utils.parseStringArray("" + a);}
 );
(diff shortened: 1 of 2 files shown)


BurningTreeC:
@saqimtiaz I believe this helps me a lot

BurningTreeC:
@saqimtiaz - I've encountered a problem, where a filter is passed as a variable to another filter. The variable can contain both a filter or a text reference to a filter. I'd like to have an operator that leaves the filter output as it is if the variable is a filter and parses the text reference to get the filter behind it if it's a text reference. Is this filter doing just that?

saqimtiaz:
@BurningTreeC no this is a string manipulation filter to turn a string like:
`"[[Title A]] TitleB [[Title C]]" `
into a list which can be manipulated with listops or iterated over:
```
"Title A"
"TitleB"
"Title C"
```

We can currently do this with `enlist[]` but if the string is the result of previous filter steps we can't chain them, leading to the need for an extra step with a set/vars widget.

pmario:
I think `aslist[]` is a good name

Jermolene:
Thanks @saqimtiaz I agree that we should not extend the "enlist" operator to cover this use case. Nonetheless, it might be helpful for users to understand that this is the same operator made available in two different packages. So, decorating the name "enlist" in some way might be a good approach, especially if there are other operators that we wish worked on input titles instead of their operand.

"enlist-input[]" could be a candidate?

Jermolene:
Hi @saqimtiaz sorry, another thought: I think it might be worth adding some tests for this operator to clarify the behaviour of edge cases (eg what happens when there are multiple input titles).

saqimtiaz:
@Jermolene I was trying to maintain the naming pattern of string operators (trim, split etc) but from a user perspective your suggestion makes more sense.

I have updated the PR to use `enlist-input`

As for multiple inputs, this is a string operator so the behaviour is the same as for other string operators like `split`, the output is the union of the output for each input title.

I am not familiar with the test framework we use but can look into it.

saqimtiaz:
@Jermolene tests added and filter renamed to `enlist-input`.

I'll prepare a separate PR with docs once this is merged.

Jermolene:
Thanks @saqimtiaz

rmunn:
#5151 is probably important to consider: I think `enlist-input` should have "dedupe" and "raw" suffixes like `enlist` does, to let the user choose whether or not it should dedupe its input. Right now the behavior appears to be deduplication, with no way to select "raw" mode.

rmunn:
PR #5152 adds "dedupe" and "raw" suffixes to `enlist-input`, similar to `enlist`.

PR #4949, opened 2020-10-31
Fix alignment of tc-tiddler-edit-title

BurningTreeC:
BEFORE:

Bildschirmfoto vom 2020-10-31 09-01-32

AFTER:

Bildschirmfoto vom 2020-10-31 09-02-04


pmario:
I don't see a difference!?

BurningTreeC:
@pmario ... the difference is, that in the second image the "Draft of 'New Tiddler 1'" at the top left is more versus the bottom

Jermolene:
Thanks @BurningTreeC

PR #4950, opened 2020-10-31
Update CodeMirror plugins to v5.58.2

BurningTreeC:
This PR updates all CodeMirror plugins to the latest CM version v5.58.2


Jermolene:
Thanks @BurningTreeC

PR #4952, opened 2020-10-31
Add keyboard shortcut to change the sidebar layout

BurningTreeC:
This PR adds a keyboard shortcut (shift-alt-Down by default) that changes the sidebar layout from fluid-fixed to fixed-fluid or vice versa

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -72,6 +72,7 @@ Shortcuts/Input/Down/Hint: Select the next item
 Shortcuts/Input/Tab-Left/Hint: Select the previous Tab
 Shortcuts/Input/Tab-Right/Hint: Select the next Tab
 Shortcuts/Input/Up/Hint: Select the previous item
+Shortcuts/SidebarLayout/Hint: Change the sidebar layout
 SystemTiddler/Tooltip: This is a system tiddler
 SystemTiddlers/Include/Prompt: Include system tiddlers
 TagManager/Colour/Heading: Colour
--- a/core/ui/KeyboardShortcuts/change-sidebar-layout.tid
+++ b/core/ui/KeyboardShortcuts/change-sidebar-layout.tid
@@ -1,0 +1,8 @@
+title: $:/core/ui/KeyboardShortcuts/change-sidebar-layout
+tags: $:/tags/KeyboardShortcut
+key: ((change-sidebar-layout))
+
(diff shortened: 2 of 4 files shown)


saqimtiaz:
I'm not convinced this is a setting changed often enough by most users to warrant a keyboard shortcut be configured for it by default.

BurningTreeC:
@saqimtiaz - it's the first thing I change when creating a new tiddlywiki and something I change often when using tiddlymap's live view

BurningTreeC:
...But I'm ok if this doesn't get merged

saqimtiaz:
@BurningTreeC however I don't recall ever seeing any users asking about this on GG/Reddit/Discord.

Features should be useful to majority of users to be added to the core in my opinion. Otherwise it's more so plugin territory. 

Keyboard shortcuts are great but having too many by default makes it overwhelming for users to remember and actually decreases utility.

Jermolene:
Hi @BurningTreeC @saqimtiaz it's an interesting case. It is also a setting that I find myself changing often. Generally our criteria for inclusion in the core are built on the idea of being universally useful -- not that all users will use all core features, but that there are a plausible circumstances where anyone might benefit from those features. I think there's little penalty for having too many keyboard shortcuts in terms of complexification of the user experience, indeed software like TiddlyWiki often prides itself on supporting many keyboard shortcuts. So I'd be inclined to go ahead with this, but I'll wait for other views before I merge.

saqimtiaz:
@Jermolene my concern is mostly that the list of keyboard shortcuts in the control panel is long and hard to scan at a glance, especially when plugins are adding their own shortcuts. Also, the more shortcuts that are defined in the core, the less available options there are for plugins without clobbering something the users might expect from the core.

I think a good target for 5.1.24 would be a better way to organize keyboard shortcuts in the control panel tab. Perhaps a way to group them: collapsible sections or sub-tabs for editor shortcuts, UI shortcuts, etc.

`toggle-sidebar` for example, is listed between `superscript` and `underline`, which isn't intuitive when you don't know how the keyboard shortcut might be described.

Jermolene:
Hi @saqimtiaz agreed. The palette editor and settings tab suffer from many of the same problems, perhaps there's scope to bring in a general purpose component that groups items and provides dynamic search (similar to the existing plugins listing).

saqimtiaz:
@Jermolene I've created an issue so we can track that going forwards and hopefully get something in place for 5.1.24

On that basis I retract my objection to merging this.

Jermolene:
Great, thanks @saqimtiaz @BurningTreeC

PR #4954, opened 2020-10-31
Init less parsers

pmario:
@Jermolene ... I did create a new plugin, that contains a parser. I saw, that my parsers are initialized, even if they are not needed. .... 

The problem is caused by: the **list-widget** `emptyMessage` and **$:/palettes/Vanilla** 

This commit in the PR fixes the problem with the list-widget. There will be more commits for the data-tiddler handling.

 - calling `$tw.wiki.parseText("text/vnd.tiddlywiki","",{parseAsInline: true}).tree` will return an empty array: `[]`

So it is much faster to `return []` if we know, that `emptyMessage` is empty.

At wiki startup `.wiki.paresText()` is called about 60 times for no reason. ... Working with the wiki and rendering tiddlers will increase the numbers. ...


Jermolene:
Thanks @pmario that makes sense. Small thing, there are missing braces around the `if` statement.

pmario:
@Jermolene ... The second element is caused by empty "values" in $:/palettes/Vanilla .. I know, where to fix it, but it would need more discussions, since the solution isn't as obvious as here. 

I'll create a new PR for the palette fix / discussion

Jermolene:
Thanks @pmario

PR #4956, opened 2020-11-01
Import UI: add warning about tiddlers that already exist

saqimtiaz:
Adds a message in the status column for tiddlers that already exists, so users can avoid accidental overwrites.

image

This should be the last of the Import UI pull requests for now.


Jermolene:
Thanks @saqimtiaz

PR #4959, opened 2020-11-01
Intersection filter run prefix

### LINKED ISSUE #4775: [IDEA] Intersection operator
From time to time there is need for finding the *intersection* between two lists. I propose an **intersection operator** for this.

For my latest use case, I managed to partly solve this like so:

```
<$set name="b" filter="""[enlist{!!list-b}split[ ]join[|]addprefix[(?=]addsuffix[)]]""" >
<$list filter="""[enlist{!!list-a}regexp<b>]""" >

</$list>
</$set>
```
i.e the regexp checks every item in "list-a" against "list-b" which is in this form: `(?=Item1|Item2|Item3|...)`. Unfortunately this doesn't handle spaced titles sufficiently nor forbidden characers.

If the correct regexp is identified, I'd think it is fairly straightforward to convert this into a  single operator that takes a list as input and another list as parameter and spits out the intersection.

saqimtiaz:
Fixes #4775 

This PR implements a `:intersection` filter run prefix.

As discussed in #4775 a run prefix is more flexible in use as compared to an intersection filter.

Compare:
` [[field Operator]tags[]] :intersection[[after Operator]tags[]]`

vs

```
<$set name="a" filter="[[field Operator]tags[]]" >
{{{ [[after Operator]tags[]] +[intersection<a>] }}}
</$set>
```

However the real question is how to document this to differentiate it from `:and` which is described in docs as an `intersection of sets`.


saqimtiaz:
@Jermolene I would welcome any input on how to document the difference between this `:intersection[run]` filter prefix as compared to the `:and[run]` prefix which [is documented as "intersection of sets"](https://tiddlywiki.com/prerelease/#Filter%20Expression)

@rmunn suggested in #4775 that we change the docs for :and[run] to not talk about set intersection but rather about adding more conditions. 

image

Jermolene:
Thanks @saqimtiaz it seems like the problem is indeed that "intersection" is not a good description for the ":and" prefix. Perhaps "accumulation" might describe it instead?

saqimtiaz:
@Jermolene "accumulation of filter steps", perhaps, rather than "accumulation of sets"?

saqimtiaz:
The issue is of course that the table is supposed to present things in "technical/logical" terms.

Jermolene:
> @Jermolene "accumulation of filter steps", perhaps, rather than "accumulation of sets"?

I think that helps. We'll never map perfectly onto established vocabulary, and need to qualify terms to clarify exactly how they are being applied.

PR #4961, opened 2020-11-01
Make Escape in keyboard-driven-inputs behave like in AdvancedSearch/Filter

BurningTreeC:
This PR updates all keyboard-driven-inputs to have a more sophisticated functionality when hitting <kbd>Escape</kbd>

~~Note the new textoperation "focus-editor" which focuses the Editor when hitting <kbd>Escape</kbd> in an empty link-dropdown~~


BurningTreeC:
I believe what I'm trying to achieve with the focus-editor textoperation is also possible with the wrap-selection textoperation

saqimtiaz:
> I believe what I'm trying to achieve with the focus-editor textoperation is also possible with the wrap-selection textoperation

@BurningTreeC perhaps mark this as a draft while you look into that. Would be good to avoid adding a text-operation if it isn't strictly needed.

BurningTreeC:
@saqimtiaz thanks, I just cannot mark it as draft atm because I'm away from my desk

BurningTreeC:
@Jermolene - now this is ready to get merged

Jermolene:
Thanks @BurningTreeC

PR #4964, opened 2020-11-02
Multiple operands for filter operators

saqimtiaz:
This PR adds support for multiple operands for filter operators.

The syntax is `operator[operand1],[operand2],[operand3]`

I settled on the ` , ` character to separate operands as it is intuitive as a representation of conjoined values, and ` : ` is already overloaded for both separating multiple suffixes, as well as for named filter run prefixes.

I've added tests and this should be completely backwards compatible. 
Filter operators have access to the first operand (if any) via `operator.operand` as before. In addition an `operands` property on the operator is an array of all operands provided to the operator.

I've also temporarily added a `string-replace` operator just to illustrate and test a use case:

`[[The quick brown fox jumps over the lazy dog]string-replace[dog],[cat]]`

outputs: `The quick brown fox jumps over the lazy cat`

The intention is for this PR to serve as a basis for discussion towards the end of having multiple operands in filter operators.


Jermolene:
Great stuff @saqimtiaz, it's an excellent idea, and the implementation is indeed pleasingly neat and self-contained.

saqimtiaz:
@Jermolene it further extends the capabilities of filters quite nicely and readability of the resultant syntax seems pretty good.

Let me know if you have any suggestions for changes, apart from removing the `string-replace` operator which probably deserves separate consideration and is just there for testing.

It is a fundamental change so feedback from others would be good too.

Jermolene:
Hi @saqimtiaz I think it's just a matter of documentation now. I agree that we should tackle `string-replace` separately (perhaps it should be `search-replace` for consistently).

saqimtiaz:
@Jermolene I have oustanding PRs to prepare for docs for `:intersection` and `enlist-input[]`. I will work on those first and then come back to this in a day or two. That will allow us to consider feedback from others, if any.

Very excited for 5.1.23, this release brings a significant slew of new features!

core/modules/filters.js
```diff
@@ -62,43 +62,61 @@ function parseFilterOperation(operators,filterString,p) {
...
+			}
+			p = nextBracketPos + 1;
 		}
+		
 		p = nextBracketPos + 1;
-
+		parseOperand(bracket);
+		
+		// Check for multiple operands
+		while(filterString.charAt(p) === ",") {
```
pmario, line 109: Operands can be {!!title}. right? ... Since titles can have commas, is there a problem?

core/modules/filters.js
```diff
@@ -62,43 +62,61 @@ function parseFilterOperation(operators,filterString,p) {
...
+			p++;
+			if(/^[\[\{<\/]/.test(filterString.substring(p))) {
```
pmario, line 111: The comment above should have been here.

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -19,19 +19,38 @@ describe("Filter tests", function() {
...
 		);
 		expect($tw.wiki.parseFilter("[search: , : [operand]]")).toEqual(
-			[ { prefix : '', operators : [ { operator : 'search', suffix : ' , : ', suffixes : [ [  ], [  ] ], operand : 'operand' } ] } ]
+			[ { prefix : '', operators : [ { operator : 'search', suffix : ' , : ', suffixes : [ [  ], [  ] ], operands: [ { text:'operand' } ] } ] } ]
+		);
+	});
+	
+	
+	it("should parse multiple operands for operators", function() {
+		expect($tw.wiki.parseFilter("[search: , : [operand],[operand2]]")).toEqual(
```
pmario, line 40: plase test: `expect($tw.wiki.parseFilter("[search: , : [oper,and],[operand2]]")).toEqual(`

core/modules/filters.js
saqimtiaz: @pmario no this isn't an issue as the parser reads the complete operand before checking for the comma. I did test this but an explicit test is a good idea and I will add one.

editions/test/tiddlers/tests/test-filters.js
saqimtiaz: @pmario good suggestion. Added the test with a comma in the operand.

saqimtiaz:
@Jermolene this is good to go.

I've removed the string-replace filter and added an extra test for comma characters in tiddler titles.

Docs in #4969 

I will add more docs once we have an operator in the core that uses multiple operands.

saqimtiaz:
@Jermolene anything else that needs attention here?

Merging this opens the door for quite a few plugins possibilities that I am keen to explore while I have the time for it.

Jermolene:
Thanks @saqimtiaz

saqimtiaz:
@Jermolene thank you!

PR #4965, opened 2020-11-02
Add generateTransclusionFootprint method to widget.js ...

BurningTreeC:
... and add switches to the `<<qualify>>` macro and the `<$qualify>` widget (for backwards compatibility) that allow to turn the new functionality on through `footprint="yes"` (or a better name)

### DIFF
--- a/core/modules/macros/qualify.js
+++ b/core/modules/macros/qualify.js
@@ -19,14 +19,15 @@ Information about this macro
 exports.name = "qualify";
 
 exports.params = [
-	{name: "title"}
+	{name: "title"},
+	{name: "footprint"}
 ];
 
 /*
 Run the macro
 */
-exports.run = function(title) {
-	return title + "-" + this.getStateQualifier();
+exports.run = function(title,footprint) {
+	return title + "-" + this.getStateQualifier(undefined,footprint === "yes");
 };
 
 })();
--- a/core/modules/widgets/qualify.js
+++ b/core/modules/widgets/qualify.js
@@ -40,9 +40,10 @@ QualifyWidget.prototype.execute = function() {
 	// Get our parameters
 	this.qualifyName = this.getAttribute("name");
 	this.qualifyTitle = this.getAttribute("title");
+	this.qualifyTransclusionFootprint = this.getAttribute("footprint") === "yes"
 	// Set context variable
 	if(this.qualifyName) {
-		this.setVariable(this.qualifyName,this.qualifyTitle + "-" + this.getStateQualifier());
+		this.setVariable(this.qualifyName,this.qualifyTitle + "-" + this.getStateQualifier(undefined,this.qualifyTransclusionFootprint));
 	}
 	// Construct the child widgets
 	this.makeChildWidgets();
@@ -53,7 +54,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 QualifyWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.name || changedAttributes.title) {
+	if(changedAttributes.name || changedAttributes.title || changedAttributes.footprint) {
 		this.refreshSelf();
 		return true;
 	} else {
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -232,10 +232,34 @@ Widget.prototype.hasVariable = function(name,value) {
 	return false;
 };
 
+Widget.prototype.findParentTransclusionWidget = function() {
+	var node = this;
+	var transclusionVariable = this.getVariable("transclusion");
+	while(node.getVariable("transclusion") === transclusionVariable) {
+		node = node.parentWidget;
+	}
+	return node !== this ? node : null;
+};
+
+Widget.prototype.generateTransclusionFootprint = function() {
+	var parentTransclusionWidget = this.findParentTransclusionWidget(),
+	    node = this,
+	    footprint = node.parentWidget.children.indexOf(node);
+	while(node) {
+		node = node.parentWidget;
+		if(node === parentTransclusionWidget) {
+			break;
+		} else if(node.parentWidget && node.parentWidget.children) {
+			footprint = footprint + "-" + node.parentWidget.children.indexOf(node);
+		}
+	}
+	return footprint;
+};
+
 /*
 Construct a qualifying string based on a hash of concatenating the values of a given variable in the parent chain
 */
-Widget.prototype.getStateQualifier = function(name) {
+Widget.prototype.getStateQualifier = function(name,generateTransclusionFootprint) {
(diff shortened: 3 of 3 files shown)


saqimtiaz:
@BurningTreeC could you please explain exactly what these two switches to? I've had some thoughts around qualify and auto-generating state titles recently and I'm wondering if this might be relevant.

saqimtiaz:
I've pulled some background and context from #4113 but a concise explanation would be great.

BurningTreeC:
@saqimtiaz , @Jermolene and me have briefly discussed this in #4113

the switches, if set to `yes` generate a unique widget-footprint so that also sibling-widgets have different qualified ID's

saqimtiaz:
@BurningTreeC ah excellent. This would allow auto-generating state titles in a situation where they need to be consistent across refresh but we don't care exactly what the titles are.

I have a sliders plugin ala Nested Sliders from TWC and it's quite cumbersome to have to manually assign a state title for each. The user doesn't care what the titles are, they just need to be consistent across refresh so that the state can be restored.

BurningTreeC:
@saqimtiaz I hope this works that way like you describe

core/modules/widgets/widget.js
```diff
@@ -232,10 +232,40 @@ Widget.prototype.hasVariable = function(name,value) {
...
+/*
+Find the widget, walking up the widget-tree, that has a different transclusion variable than the current widget
+*/
+Widget.prototype.findParentTransclusionWidget = function() {
+	var node = this;
+	var transclusionVariable = this.getVariable("transclusion");
+	while(node.getVariable("transclusion") === transclusionVariable) {
+		node = node.parentWidget;
+	}
+	return node !== this ? node : null;
```
pmario, line 244: please use braces here. Ternary operators are always used that way in the core. `(node !== this) ? node : null;` IMO it's more readable.

core/modules/widgets/widget.js
```diff
@@ -249,7 +279,12 @@ Widget.prototype.getStateQualifier = function(name) {
 			}
 			node = node.parentWidget;
 		}
-		var value = $tw.utils.hashString(output.join(""));
+		var value;
+		if(generateTransclusionFootprint) {
+			value = $tw.utils.hashString(output.join("")) + "_" + this.generateTransclusionFootprint();
```
pmario, line 284: Does the `<<qualify` macro and the `<$qualify` widget produce the same, or different hashes?

core/modules/widgets/qualify.js
```diff
@@ -40,9 +40,10 @@ QualifyWidget.prototype.execute = function() {
 	// Get our parameters
 	this.qualifyName = this.getAttribute("name");
 	this.qualifyTitle = this.getAttribute("title");
+	this.qualifyTransclusionFootprint = this.getAttribute("footprint") === "yes"
 	// Set context variable
 	if(this.qualifyName) {
-		this.setVariable(this.qualifyName,this.qualifyTitle + "-" + this.getStateQualifier());
+		this.setVariable(this.qualifyName,this.qualifyTitle + "-" + this.getStateQualifier(undefined,this.qualifyTransclusionFootprint));
```
pmario, line 46: IMO you should not rely on the default value used by the `this.getStateQualifier()`. IMO you should name the variable that is used!

core/modules/macros/qualify.js
```diff
@@ -19,14 +19,15 @@ Information about this macro
...
+	{name: "footprint"}
 ];
 
 /*
 Run the macro
 */
-exports.run = function(title) {
-	return title + "-" + this.getStateQualifier();
+exports.run = function(title,footprint) {
+	return title + "-" + this.getStateQualifier(undefined,footprint === "yes");
```
pmario, line 30: I think we shouldn't use "undefined" here .. IMO it should be named!

pmario:
@Jermolene ... If you are willing to change the behaviour of the `<<qualify` macro and widget. Please also let's discuss the option I wanted to add. see: "Add isUnique to qualify - and - update core tabs macro" #4454

Jermolene:
> @Jermolene ... If you are willing to change the behaviour of the `<<qualify` macro and widget. Please also let's discuss the option I wanted to add. see: "Add isUnique to qualify - and - update core tabs macro" #4454

Hi @pmario I didn't propose changing the behaviour of the `<<qualify>>` macro for precisely the same reasons I gave in our earlier discussion. See https://github.com/Jermolene/TiddlyWiki5/pull/4113#issuecomment-720681235:

> Hi @BurningTreeC I think that actually to maintain backwards compatibility we'd need to add a flag to `Widget.prototype.getStateQualifier` to choose between the old and new behaviours, and then a new version of the `<<qualify>>` macro that uses the new mechanism, and a new attribute flag for the (embarrassingly undocumented) `<$qualify>` widget. Then we'd look at the core macros that take a state qualifier and figure out how to adapt them.

@BurningTreeC I'll add some code review points shortly.

BurningTreeC:
@Jermolene - is this still of interest?

Jermolene:
Apologies @BurningTreeC I got distracted by something before leaving the code comments...

BurningTreeC:
>Apologies @BurningTreeC I got distracted by something before leaving the code comments...

No worries, I just wanted to **ping** you

BurningTreeC:
@Jermolene, pinging you for this one, just in case you forgot about it

BurningTreeC:
@Jermolene, whenever you find the time for this...

Jermolene:
Sorry @BurningTreeC I have been noodling away at this in the background.

My first concern was that the footprints you're generating wouldn't be durable in the face of dynamic widgets, particularly the way that the list widget moves its child widgets around in response to reordering of items. I'm still unsure about that.

The other thing that's been on my mind is whether these footprints could be an alternative to the existing qualified state tiddlers, and I've now realised that they definitely cannot be used for the same purpose. An important property of the existing qualified state titles is that siblings in the render tree will get back the same value. That enables us to have a reveal widget and a button widget alongside each other, each calling qualify and getting the same value.

So I think footprints are at best a separate mechanism that we may be able to use for focus/selection preservation.

BurningTreeC:
I understand @Jermolene. Should I try changing this PR so that it adds only the `generateTransclusionFootprint` method?

See also #4113 where this method is used

Jermolene:
> See also #4113 where this method is used

Indeed, I think #4113 makes a start on the focus/selection preservation I was referring to. So perhaps it makes sense to move the footprint generation into the same PR?

BurningTreeC:
>Indeed, I think #4113 makes a start on the focus/selection preservation I was referring to. So perhaps it makes sense to move the footprint generation into the same PR?

Yes, I'll do so. Thanks @Jermolene

BurningTreeC:
Ah, I forgot. The `generateTransclusionFootprint` method already IS part of #4113 @Jermolene.

I'll close this PR here

PR #4970, opened 2020-11-03
Add missing inputAcceptVariantActions to fieldname, tag-picker and link-dropdown

BurningTreeC:
This PR adds the missing save-tiddler-actions to the fieldname-input, the tag-picker and the link-dropdown


BurningTreeC:
Merging #4971 would make this PR superfluous

saqimtiaz:
@BurningTreeC perhaps add a short explanation as to what this PR is about, and which solution is preferred.

BurningTreeC:
@saqimtiaz - the inputAcceptVariantActions keyboard shortcut is set to <kbd>ctrl-Enter</kbd> in the current prerelease and I'd like to leave it that way. Now that shortcut is also the shortcut to save the current tiddler when it's being edited. In the tags input, the link-dropdown and the fieldname input the inputAcceptVariantActions aren't set so that shortcut is trapped and doesn't do anything. Now we can either define the save-tiddler-actions as the inputAcceptVariantActions, which leads to a little confusing behavior if the user changes the keyboard shortcut for the "input-accept-variant" actions, or we merge #4971 , which is a much better fix

BurningTreeC:
Closing this PR in favor of #4971

PR #4971, opened 2020-11-03
KeyboardWidget: don't trap keyboard shortcut if actions and message are empty and invokeActions isn't handled

BurningTreeC:
This PR changes the KeyboardWidget so that, if there are no actions, no message and no invokeActions to handle, `event.preventDefault()` and `event.stopPropagation()` aren't called.

This is useful for the keyboard-driven-inputs in the tag-picker, the link-dropdown and the fieldname-input, where no actions are defined for the keyboard-widget that has the (default) keyboard-shortcut <kbd>ctrl-Enter</kbd> and would prevent the tiddler-global `save-tiddler-actions` that should be dispatched on <kbd>ctrl-Enter</kbd>

### DIFF
--- a/core/modules/widgets/keyboard.js
+++ b/core/modules/widgets/keyboard.js
@@ -51,7 +51,7 @@ KeyboardWidget.prototype.render = function(parent,nextSibling) {
 				self.invokeActionString(self.actions,self,event);
 			}
 			self.dispatchMessage(event);
-			if(self.actions) {
+			if(self.actions || self.message) {
 				event.preventDefault();
 				event.stopPropagation();
 			}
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@BurningTreeC I don't think this is backwards compatible. Keyboard widgets also invoke any action widgets contained inside them. With this chance the keys would not be trapped in those situations where there the actions attribute is blank but there might be inline action widgets.

BurningTreeC:
You're right, thanks @saqimtiaz

saqimtiaz:
@BurningTreeC just a heads up that when you work on a PR like #4974  that isn't marked as draft, it spams the notifications for every commit you make.

BurningTreeC:
Thanks @saqimtiaz - I've converted it to draft

BurningTreeC:
I've updated the OP

Jermolene:
Thanks @BurningTreeC

PR #4974, opened 2020-11-04
Menubar plugin: make sidebar-scrollable top margin adjust ...

BurningTreeC:
... to height of menubar

Peek 2020-11-04 10-31

### DIFF
--- a/core/ui/TopRightBar/menu.tid
+++ b/core/ui/TopRightBar/menu.tid
@@ -2,8 +2,8 @@ title: $:/core/ui/TopBar/menu
 tags: $:/tags/TopRightBar
 
 <$list filter="[[$:/state/sidebar]get[text]] +[else[yes]!match[no]]" variable="ignore">
-<$button set="$:/state/sidebar" setTo="no" tooltip={{$:/language/Buttons/HideSideBar/Hint}} aria-label={{$:/language/Buttons/HideSideBar/Caption}} class="tc-btn-invisible">{{$:/core/images/chevron-right}}</$button>
+<$button set="$:/state/sidebar" setTo="no" tooltip={{$:/language/Buttons/HideSideBar/Hint}} aria-label={{$:/language/Buttons/HideSideBar/Caption}} class="tc-btn-invisible tc-hide-sidebar-btn">{{$:/core/images/chevron-right}}</$button>
 </$list>
 <$list filter="[[$:/state/sidebar]get[text]] +[else[yes]match[no]]" variable="ignore">
-<$button set="$:/state/sidebar" setTo="yes" tooltip={{$:/language/Buttons/ShowSideBar/Hint}} aria-label={{$:/language/Buttons/ShowSideBar/Caption}} class="tc-btn-invisible">{{$:/core/images/chevron-left}}</$button>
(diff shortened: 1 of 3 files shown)


BurningTreeC:
Peek 2020-11-04 11-37

BurningTreeC:
This PR also rotates the sidebar-chevrons when the menubar has its narrow styles

BurningTreeC:
Peek 2020-11-05 12-10

pmario:
Did you test it in a way, where the menu bar is only 1 line high. .. Does the height calculation still work?

BurningTreeC:
>Did you test it in a way, where the menu bar is only 1 line high. .. Does the height calculation still work?

@pmario - yes I did test that, it works also in that case

BurningTreeC:
for the screen-width between the sidebarbreakpoint and the narrow-breakpoint this needs some work

Jermolene:
Good stuff thank you @BurningTreeC. Using wikitext processing in stylesheets is like a magical power in TiddlyWiki, great to see it used more.

BurningTreeC:
@Jermolene, I've updated the styles and the menu

Peek 2020-11-06 13-55

BurningTreeC:
>Using wikitext processing in stylesheets is like a magical power in TiddlyWiki, great to see it used more.

Yes it's really *magical* :smile:

BurningTreeC:
@Jermolene - are you interested in this?

I was also thinking, how browser support is for calculating media queries directly like:

```

@media (max-width: calc({{$:/config/plugins/menubar/breakpoint}} - 1px) ) {

...

}

```

Jermolene:
> @Jermolene - are you interested in this?

Yes!
> I was also thinking, how browser support is for calculating media queries directly like:
> ```
> @media (max-width: calc({{$:/config/plugins/menubar/breakpoint}} - 1px) ) {
> ...
> }
> ```

We could do that with a filter, couldn't we?

Jermolene:
Sorry @BurningTreeC I misunderstood

BurningTreeC:
@Jermolene - in this PR I'm not calculating media queries like described

the currently used filter that removes the "px" suffix, adds "1" and adds the "px" suffix again works only for widths with px units. if a user chooses em or vw it wouldn't work anymore ... That's why I'm wondering about browser support for media queries that use calc()

Jermolene:
Ouch, that is indeed not very satisfying, I imagine it won't prove to be a problem.

BurningTreeC:
>I imagine it won't prove to be a problem.

Yout think we can use:

```
@media (max-width: calc({{$:/config/plugins/menubar/breakpoint}} - 1px) ) {

...

}
```

?

Jermolene:
Indeed @BurningTreeC Googling suggests that there is only partial browser support at the moment.

kookma:
Hi @BurningTreeC 
I just checked this on Android running cell phone, opening tiddlywiki.com/prerelease see below images

WhatsApp Image 2020-11-26 at 2 40 17 PM
WhatsApp Image 2020-11-26 at 2 39 53 PM

See the sidebar toggle button! Is this a bug? 

I also tested in Firefox using design mode and different WxH. It seems below W=640px the toggle button is not visible and see it like above image. So on some Mobile phone you see this issue.

rmunn:
#5150 is probably related to this.

PR #4975, opened 2020-11-04
Add editor-height button and stamp-button to tiddler types that miss them

BurningTreeC:
This PR adds the editor-height and stamp buttons also to tiddlers of type application/javascript, application/json and application/x-tiddler-dictionary


Jermolene:
Thanks @BurningTreeC

PR #4977, opened 2020-11-04
Added more start/end parser ranges

SmilyOrg:
- [ ] Test all default tiddlers for missing ranges (e.g. `macrocall`)
- [ ] Fix existing tests (`./bin/test`)
- [ ] Add more tests

### DIFF
--- a/core/modules/parsers/wikiparser/wikiparser.js
+++ b/core/modules/parsers/wikiparser/wikiparser.js
@@ -165,6 +165,17 @@ WikiParser.prototype.findNextMatch = function(rules,startPos) {
 	return matchingRule;
 };
 
+WikiParser.prototype.parseRule = function(rule) {
+	var start = this.pos;
+	var blocks = rule.parse();
+	for(var i=0; i<blocks.length; i++) {
+		var block = blocks[i];
+		if (block.start !== undefined) block.start = start;
+		if (block.end !== undefined) block.end = this.pos;
+	}
+	return blocks;
+};
+
 /*
 Parse any pragmas at the beginning of a block of parse text
 */
@@ -184,7 +195,7 @@ WikiParser.prototype.parsePragmas = function() {
 			break;
 		}
 		// Process the pragma rule
-		var subTree = nextMatch.rule.parse();
+		var subTree = this.parseRule(nextMatch.rule);
 		if(subTree.length > 0) {
 			// Quick hack; we only cope with a single parse tree node being returned, which is true at the moment
 			currentTreeBranch.push.apply(currentTreeBranch,subTree);
@@ -208,10 +219,13 @@ WikiParser.prototype.parseBlock = function(terminatorRegExpString) {
 	// Look for a block rule that applies at the current position
 	var nextMatch = this.findNextMatch(this.blockRules,this.pos);
 	if(nextMatch && nextMatch.matchIndex === this.pos) {
-		return nextMatch.rule.parse();
+		return this.parseRule(nextMatch.rule);
 	}
 	// Treat it as a paragraph if we didn't find a block rule
-	return [{type: "element", tag: "p", children: this.parseInlineRun(terminatorRegExp)}];
+	var start = this.pos;
+	var children = this.parseInlineRun(terminatorRegExp);
+	var end = this.pos;
+	return [{type: "element", tag: "p", children, start, end }];
 };
 
 /*
@@ -287,17 +301,17 @@ WikiParser.prototype.parseInlineRunUnterminated = function(options) {
 	while(this.pos < this.sourceLength && nextMatch) {
 		// Process the text preceding the run rule
 		if(nextMatch.matchIndex > this.pos) {
-			this.pushTextWidget(tree,this.source.substring(this.pos,nextMatch.matchIndex));
+			this.pushTextWidget(tree,this.source.substring(this.pos,nextMatch.matchIndex),this.pos,nextMatch.matchIndex);
 			this.pos = nextMatch.matchIndex;
 		}
 		// Process the run rule
-		tree.push.apply(tree,nextMatch.rule.parse());
+		tree.push.apply(tree,this.parseRule(nextMatch.rule));
 		// Look for the next run rule
 		nextMatch = this.findNextMatch(this.inlineRules,this.pos);
 	}
 	// Process the remaining text
 	if(this.pos < this.sourceLength) {
-		this.pushTextWidget(tree,this.source.substr(this.pos));
+		this.pushTextWidget(tree,this.source.substr(this.pos),this.pos,this.sourceLength);
 	}
 	this.pos = this.sourceLength;
 	return tree;
@@ -317,7 +331,7 @@ WikiParser.prototype.parseInlineRunTerminated = function(terminatorRegExp,option
 		if(terminatorMatch) {
 			if(!inlineRuleMatch || inlineRuleMatch.matchIndex >= terminatorMatch.index) {
 				if(terminatorMatch.index > this.pos) {
-					this.pushTextWidget(tree,this.source.substring(this.pos,terminatorMatch.index));
+					this.pushTextWidget(tree,this.source.substring(this.pos,terminatorMatch.index),this.pos,terminatorMatch.index);
 				}
 				this.pos = terminatorMatch.index;
 				if(options.eatTerminator) {
@@ -330,11 +344,11 @@ WikiParser.prototype.parseInlineRunTerminated = function(terminatorRegExp,option
 		if(inlineRuleMatch) {
 			// Preceding text
 			if(inlineRuleMatch.matchIndex > this.pos) {
-				this.pushTextWidget(tree,this.source.substring(this.pos,inlineRuleMatch.matchIndex));
+				this.pushTextWidget(tree,this.source.substring(this.pos,inlineRuleMatch.matchIndex),this.pos,inlineRuleMatch.matchIndex);
 				this.pos = inlineRuleMatch.matchIndex;
 			}
 			// Process the inline rule
-			tree.push.apply(tree,inlineRuleMatch.rule.parse());
+			tree.push.apply(tree,this.parseRule(inlineRuleMatch.rule));
 			// Look for the next inline rule
 			inlineRuleMatch = this.findNextMatch(this.inlineRules,this.pos);
 			// Look for the next terminator match
@@ -344,7 +358,7 @@ WikiParser.prototype.parseInlineRunTerminated = function(terminatorRegExp,option
 	}
 	// Process the remaining text
 	if(this.pos < this.sourceLength) {
-		this.pushTextWidget(tree,this.source.substr(this.pos));
+		this.pushTextWidget(tree,this.source.substr(this.pos),this.pos,this.sourceLength);
 	}
 	this.pos = this.sourceLength;
 	return tree;
@@ -353,12 +367,12 @@ WikiParser.prototype.parseInlineRunTerminated = function(terminatorRegExp,option
 /*
 Push a text widget onto an array, respecting the configTrimWhiteSpace setting
 */
-WikiParser.prototype.pushTextWidget = function(array,text) {
+WikiParser.prototype.pushTextWidget = function(array,text,start,end) {
 	if(this.configTrimWhiteSpace) {
 		text = $tw.utils.trim(text);
 	}
 	if(text) {
-		array.push({type: "text", text: text});		
+		array.push({type: "text", text: text, start: start, end: end});		
 	}
 };
 


pmario:
Could you add more description, what's it good for?

Jermolene:
Hi @SmilyOrg this is a worthwhile improvement thank you.

@pmario if you use the parse tree preview you'll see that some but not all nodes have "start" and "end" properties marking the source text from which they are derived. These values would be useful for example if re-using the wikitext parser for syntax highlighting purposes.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -165,6 +165,17 @@ WikiParser.prototype.findNextMatch = function(rules,startPos) {
 	return matchingRule;
 };
 
+WikiParser.prototype.parseRule = function(rule) {
+	var start = this.pos;
+	var blocks = rule.parse();
+	for(var i=0; i<blocks.length; i++) {
+		var block = blocks[i];
+		if (block.start !== undefined) block.start = start;
```
Jermolene, line 173: `if` statements should not have whitespace between the `if` and the opening parenthesis. We also use braces around the body.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -165,6 +165,17 @@ WikiParser.prototype.findNextMatch = function(rules,startPos) {
 	return matchingRule;
 };
 
+WikiParser.prototype.parseRule = function(rule) {
+	var start = this.pos;
```
Jermolene, line 169: Consecutive `vars` should generally be merged together.

Jermolene (commented):
A couple of minor coding standards issues, but otherwise looks good.

It would be useful to update the test suite where appropriate too.

SmilyOrg:
Sorry for leaving the description hanging, I've added some use-cases and background now.

I've also added a commit with fixes for coding standards. Thanks for the review and describing the PR before I got around to it @Jermolene. 😄 

I'll fix the tests and see how much range coverage we get on default tiddlers when I get the time, there might be easy wins to add still. See the checklist at the end of the PR for progress 😀

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -165,6 +165,21 @@ WikiParser.prototype.findNextMatch = function(rules,startPos) {
 	return matchingRule;
 };
 
+WikiParser.prototype.parseRule = function(rule) {
+	var start = this.pos,
+		blocks = rule.parse();
+	for(var i=0; i<blocks.length; i++) {
```
pmario, line 171: TW core code most of the times uses a construction like: 

```
var i, length  // var defs outside for

for(i=0, length=blocks.length; i<length; i++) {
```

Jermolene: Apologies @pmario that's not quite right, the core does frequently declare variables within a "for" statement.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -165,6 +165,21 @@ WikiParser.prototype.findNextMatch = function(rules,startPos) {
...
+		var block = blocks[i];
+		if(block.start !== undefined) {
```
pmario, line 173: what is the extra `if ()` element needed for? .. If there are no blocks, imo the for shouldn't be executed at all. ... It should be a "return early" approach.

core/modules/parsers/wikiparser/wikiparser.js
ght: Just a FYI: caching the `length` when iterating over an array with a for-loop actually results in worse performance (though probably not measurably). Read http://mrale.ph/blog/2014/12/24/array-length-caching.html for details if you want to know more of what happens under the hood.

TL;DR; of the linked post: When accessing `blocks[i]` inside the loop, the engine has to do a bounds check to make sure `i` is not past the end of `blocks`. When you do not cache the length, but use `i < blocks.length` in the condition, the engine can optimize the bounds check away. If you cache, it cannot (or at least couldn't when the article was written). Don't do premature optimizations like that without measuring first.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -208,10 +223,13 @@ WikiParser.prototype.parseBlock = function(terminatorRegExpString) {
...
 	if(nextMatch && nextMatch.matchIndex === this.pos) {
-		return nextMatch.rule.parse();
+		return this.parseRule(nextMatch.rule);
 	}
 	// Treat it as a paragraph if we didn't find a block rule
-	return [{type: "element", tag: "p", children: this.parseInlineRun(terminatorRegExp)}];
+	var start = this.pos;
+	var children = this.parseInlineRun(terminatorRegExp);
+	var end = this.pos;
+	return [{type: "element", tag: "p", children, start, end }];
```
Jermolene, line 232: Apologies for the delay -- the TW5 core sticks to ECMAScript 5 for compatibility with older browsers, and so we can't use destructuring like this. Instead we should use:

```
return [{type: "element", tag: "p", children: children, start: start, end: end }];
```

pmario: @SmilyOrg ... can you add the requested changes and push a new version. So the tests can be run again. So I can see, where the test error came from.

saqimtiaz: @pmario IIRC most if not all the test errors are from the fact that the tests assume that the parser will not return start and end ranges but now it does. So the fixtures that we compare against need updating.

pmario:
@SmilyOrg ... Are you still with us?

SmilyOrg:
@pmario Hi! Sorry for letting this hang for so long. I'll try to fix it up in the next few days.

core/modules/parsers/wikiparser/wikiparser.js
SmilyOrg: In the ideal case every block would define its own range and this for loop wouldn't be needed at all.

In the real world some blocks define their range and some don't, so this is a fallback for child blocks that don't already have start/end ranges defined (and are therefore `undefined`) - in this case it falls back to using the range of the parent.

I'll add a comment in code to make this clearer.

SmilyOrg: I realize now that it's the opposite of what I intended 🤦‍♂️

SmilyOrg:
Alright, I've added the missing ranges to tests. Also uncovered some bugs with the original implementation of range fallbacks, so I've fixed that or at least made it more sensible.

I think it's now ready for another look. Thanks to everyone who looked at it already!

Jermolene:
Thanks @SmilyOrg 

What are the situations that lead to the need for `WikiParser.prototype.parseRule()`? Can we not just ensure that every parse tree node does have a start/end pair?

SmilyOrg:
> What are the situations that lead to the need for `WikiParser.prototype.parseRule()`? Can we not just ensure that every parse tree node does have a start/end pair?

That would be the ideal way to go, but:
1. It seems like there are a lot of different rules spread around many files, finding and implementing it for each seems like a lot more work than what's in the PR right now, so I thought I would start small and add some useful fallbacks.
2. I'm not sure, but I think new rules can also be added with plugins, in which case you can't guarantee the ranges will be there anyway, so having a fallback seems useful either way.

I'm not married to it though, so if you have a cleaner/easier/better way I'm all for it :)

Jermolene:
> > What are the situations that lead to the need for `WikiParser.prototype.parseRule()`? Can we not just ensure that every parse tree node does have a start/end pair?
> 
> That would be the ideal way to go, but:
> 
> 1. It seems like there are a lot of different rules spread around many files, finding and implementing it for each seems like a lot more work than what's in the PR right now, so I thought I would start small and add some useful fallbacks.
> 2. I'm not sure, but I think new rules can also be added with plugins, in which case you can't guarantee the ranges will be there anyway, so having a fallback seems useful either way.
> 
> I'm not married to it though, so if you have a cleaner/easier/better way I'm all for it :)

That makes sense. I would hope we can resolve (1) between us, but (2) is definitely an issue. But I think the answer should be for consumers of parse trees to be tolerant of nodes that don't have start/end markers, rather than filling them in ourselves. Given that parse trees from other parsers (like KaTeX) don't necessarily include start/end markers, consumers already have to deal with this issue.

saqimtiaz:
I'll need to dig around and find the relevant files but I have done some work on adding start and end ranges for the list parser.

pmario:
@SmilyOrg @Jermolene ... I would like to have this one in 5.2.0

Jermolene:
> @SmilyOrg @Jermolene ... I would like to have this one in 5.2.0

That would be great, but we need to address the concern I raised above.

pmario:
@SmilyOrg ... Can you resolve those concerns: https://github.com/Jermolene/TiddlyWiki5/pull/4977#issuecomment-826368695 ?

SmilyOrg:
Thanks for poking me to take another look at this :)

@Jermolene I agree that consumers will likely still need to handle the missing ranges that aren't covered here.

However, if we can provide useful fallbacks (e.g. by filling in blanks between subsequent ranges), it lifts some of the burden from consumers, since otherwise every consumer will have to implement this logic themselves as a post-processing step (or just give up).

If you have a feeling this is overreaching (and might lead to compatibility issues down the line?), I can try removing `WikiParser.prototype.parseRule()` and see what we're left with [1]. How does that sound?

[1]: From what I remember, without this fallback it seemed a lot less useful (only a few leaf blocks/nodes got the ranges).

Jermolene:
> However, if we can provide useful fallbacks (e.g. by filling in blanks between subsequent ranges), it lifts some of the burden from consumers, since otherwise every consumer will have to implement this logic themselves as a post-processing step (or just give up).

A concern is that different strategies/heuristics might need to be used for filling in the missing data depending upon what is being done with it. On the whole it seems safer to just report what we know, and let clients figure out any gaps.

> If you have a feeling this is overreaching (and might lead to compatibility issues down the line?), I can try removing `WikiParser.prototype.parseRule()` and see what we're left with [1]. How does that sound?

That would be great, thank you.

> [1]: From what I remember, without this fallback it seemed a lot less useful (only a few leaf blocks/nodes got the ranges). 

Where do the gaps come from? Hopefully we can gradually update any parse rules that still need updating.

SmilyOrg:
> A concern is that different strategies/heuristics might need to be used for filling in the missing data depending upon what is being done with it. On the whole it seems safer to just report what we know, and let clients figure out any gaps.

It probably is safer, yes. I'll try to get some time to remove it in the next few days.

I wonder if we could provide some more utilities that can be used internally (core) or externally (plugins) to make it easier to provide these ranges. For example, for the simple case, if you know `this.pos` before and after your rule gets parsed, that is already very valuable. That's that the fallback does, but it's a little trickier (and thus not as trivial) as it's a list of blocks to update and not just one. Maybe having a wrapper that updates ranges for you? I could be overthinking it 😄

> Where do the gaps come from? Hopefully we can gradually update any parse rules that still need updating.

Once I remove the fallback I'll get a clearer picture from the failing tests, but I think a lot of it was in parsing nested HTML embedded in wikitext.

Jermolene:
> It probably is safer, yes. I'll try to get some time to remove it in the next few days.

Thanks @SmilyOrg much appreciated.

> I wonder if we could provide some more utilities that can be used internally (core) or externally (plugins) to make it easier to provide these ranges. For example, for the simple case, if you know `this.pos` before and after your rule gets parsed, that is already very valuable. That's that the fallback does, but it's a little trickier (and thus not as trivial) as it's a list of blocks to update and not just one. Maybe having a wrapper that updates ranges for you? I could be overthinking it 😄

Perhaps it would be worth documenting those techniques in the dev wiki for the moment?

> Once I remove the fallback I'll get a clearer picture from the failing tests, but I think a lot of it was in parsing nested HTML embedded in wikitext.

Thanks.

SmilyOrg:
I've removed `parseRule` and the ranges it added from tests, seems all good now 👍

> Once I remove the fallback I'll get a clearer picture from the failing tests, but I think a lot of it was in parsing nested HTML embedded in wikitext.

From the tests it seems like the missing ranges were `element`, `set` and `entity` types. So those are probably good for further improvement.

> Perhaps it would be worth documenting those techniques in the dev wiki for the moment?

Not a bad idea, but I'll have to leave that for another time 😅

pmario:
I did add this issue: https://github.com/Jermolene/TiddlyWiki5/issues/5739 so we don't forget about it

pmario:
@Jermolene ... I think your concerns have been handled now. Can we merge it?

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -231,7 +231,10 @@ WikiParser.prototype.parseBlock = function(terminatorRegExpString) {
 		return nextMatch.rule.parse();
 	}
 	// Treat it as a paragraph if we didn't find a block rule
-	return [{type: "element", tag: "p", children: this.parseInlineRun(terminatorRegExp)}];
+	var start = this.pos;
+	var children = this.parseInlineRun(terminatorRegExp);
+	var end = this.pos;
```
saqimtiaz, line 236: We would normally write this as:

```
var start = this.pos,
	children = this.parseInlineRun(terminatorRegExp),
	end = this.pos;
```

linonetwo:
I'm trying to create a WYSIWYG editor, but I found for `core/modules/parsers/wikiparser/rules/commentblock.js`, it returns an empty node result

https://github.com/Jermolene/TiddlyWiki5/blob/a51191d334de367ba1872d7ee46d5d551516ba9e/core/modules/parsers/wikiparser/rules/commentblock.js#L49-L54

This make it not useable by the editor...Otherwise this info will be lost after edit.

PR #4978, opened 2020-11-05
Add delete-field keyboard shortcut

BurningTreeC:
This PR adds a delete-field keyboard-shortcut (<kbd>ctrl-D</kbd> by default) that deletes a currently focused field and sets focus to the fieldname-input

### DIFF
--- a/core/ui/EditTemplate/fields.tid
+++ b/core/ui/EditTemplate/fields.tid
@@ -73,7 +73,7 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 </td>
 <td class="tc-edit-field-remove">
 <$button class="tc-btn-invisible" tooltip={{$:/language/EditTemplate/Field/Remove/Hint}} aria-label={{$:/language/EditTemplate/Field/Remove/Caption}}>
-<$action-deletefield $field=<<currentField>>/>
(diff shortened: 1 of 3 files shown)


rmunn:
That would have come in handy for me once or twice, and <kbd>ctrl-D</kbd> looks like a reasonable shortcut. I like it.

pmario:
hmmm. CTRL-D could be used as a "duplicate" shortcut in other software. ... So if a user is used to this action it would cause problems. ... Is there a "are you sure to delete xxxx?" dialogue?

BurningTreeC:
Hi @pmario, there's no prompt for the user to accept deleting the field. But we could change the shortcut to <kbd>shift-alt-D</kbd> ? What do you think?

pmario:
With FireFox Ctrl-D is captured by the browser for "store page as bookmark" .. At least that's the setting for my FF German.

Jermolene:
Hi @BurningTreeC @pmario deleting a field doesn't seem like a frequent operation, so I'd be inclined to make it <kbd>shift-alt-D</kbd>.

BurningTreeC:
Ok @Jermolene

BurningTreeC:
@Jermolene @pmario , I've updated the shortcut to be <kbd>shift-alt-D</kbd>

Jermolene:
Thanks @BurningTreeC

PR #4979, opened 2020-11-05
Support for switching page templates

saqimtiaz:
Minimal changes to add support for switchable page templates by introducing `$:/config/ui/PageTemplate` which is used as a text reference for the root transclusion in render.js 

This is entirely backwards compatible, even if users might have customized the their `$:/core/ui/PageTemplate`
It also follows the same pattern introduced by `$:/config/ui/ViewTemplate` and `$:/config/ui/EditTemplate`

Another thought (not implemented) is to use a filtered attribute for the transclusion rather than a text reference, and default to `$:/core/ui/PageTemplate` if the tiddler defined in `$:/config/ui/PageTemplate` does not exist. The aim being to avoid situations where the user may accidentally end up with a not working wiki.

@Jermolene I know you've done some work in this area, so if you have a different solution in mind I am happy to discuss or close this.


Jermolene:
Thanks @saqimtiaz. I've a partially completed draft of the same thing. Instead of modifiying `wiki.makeTranscludeWidget()`, I assign `var PAGE_TEMPLATE_TITLE = "$:/core/ui/RootTemplate";` and then setting that new tiddler to:

```
<$transclude tiddler={{{ [{$:/layout}has[text]] ~[[$:/core/ui/PageTemplate]] }}} mode="inline"/>
```

I used `$:/layout` for consistency with `$:/language`, `$:/view` etc.

saqimtiaz:
@Jermolene I'm amenable to that approach. One potential concern is whether the extra transclusion in `$:/core/ui/RootTemplate` would have a knock on effect and change all the state titles for tabs etc. That would be problematic for some plugins and break backwards compatibility, which is why I went the route of modifying `wiki.makeTranscludeWidget()`

It would be good to get support for switching page templates into 5.1.23 as it opens up a lot of possibilities for customization. Let me know if you prefer to finish up your draft, otherwise I can also rework this in whichever direction is agreed upon.

joshuafontany:
Ooooh. You guys are doing excellent work in the core. Really looking forward to what this and the new filters will make possible.

Jermolene:
Hi @saqimtiaz 

> One potential concern is whether the extra transclusion in `$:/core/ui/RootTemplate` would have a knock on effect and change all the state titles for tabs etc. That would be problematic for some plugins and break backwards compatibility, which is why I went the route of modifying `wiki.makeTranscludeWidget()`

I worried about this too, and explored two solutions:

* Adding a `recursionMarker` attribute to the `<$transclude>` widget (defaults to "yes") to stop it creating a "recursion" variable, and thus having an impact on qualifiers
* Amending the core instances of the tabs macro to use explicit state tiddlers via the new `explicitState` parameter

I favour the second fix (although the first might be worth having too), because I think that any future refactoring of the page templates is likely to break the current qualified titles.


> 
> It would be good to get support for switching page templates into 5.1.23 as it opens up a lot of possibilities for customization. Let me know if you prefer to finish up your draft, otherwise I can also rework this in whichever direction is agreed upon.

saqimtiaz:
@Jermolene changing to unqualified state titles for the tabs in the core makes sense considering we do want to refactor the layout in the not so distant future. 

I can work rework this PR in the direction of introducing `$:/core/ui/RootTemplate`

@BurningTreeC would you be up to preparing a PR to change usage of the tabs macro in the core to use the explicState parameter?

BurningTreeC:
Hi @saqimtiaz - what should I do exactly? I do have time right now

saqimtiaz:
@BurningTreeC  the goal is change all the places in the core where the tabs is used, use the explicitState parameter so that they always have a fixed state tiddler.

BurningTreeC:
>the goal is change all the places in the core where the tabs macro is used, to use the explicitState parameter so that they always have a fixed state tiddler.

ok @saqimtiaz , I'll create a PR

saqimtiaz:
@BurningTreeC thank you!

Jermolene:
Thanks @saqimtiaz @BurningTreeC

saqimtiaz:
@Jermolene this now adopts the pattern you described. Also added `recursionMarker` attribute to `transclude` widget, so even without #4985 this should be backwards compatible.

`$:/layout` does not exist by default.

My only concern is that it still feels more consistent from a user point of view to use `$:/config/PageTemplate` rather than `$:/layout`, given the precedence set by `$:/config/ViewTemplate` and `$:/config/EditTemplate`.

core/modules/widgets/transclude.js
```diff
@@ -60,8 +61,10 @@ TranscludeWidget.prototype.execute = function() {
...
 	// Set context variables for recursion detection
-	var recursionMarker = this.makeRecursionMarker();
-	this.setVariable("transclusion",recursionMarker);
+	if(this.recursionMarker === "yes") {
+		var recursionMarker = this.makeRecursionMarker();
+		this.setVariable("transclusion",recursionMarker);
+	}
 	// Check for recursion
 	if(parser) {
 		if(this.parentWidget && this.parentWidget.hasVariable("transclusion",recursionMarker)) {
```
Jermolene, line 70: Note that here we use the recursionMarker variable again, so the conditional above should only be around the setVariable, and not the var.

Jermolene (commented):
Thanks @saqimtiaz 

> @Jermolene this now adopts the pattern you described. Also added `recursionMarker` attribute to `transclude` widget, so even without #4985 this should be backwards compatible.
> 
> `$:/layout` does not exist by default.
> 
> My only concern is that it still feels more consistent from a user point of view to use `$:/config/PageTemplate` rather than `$:/layout`, given the precedence set by `$:/config/ViewTemplate` and `$:/config/EditTemplate`.

I was thinking of using `$:/layout` for consistency with existing top level settings like `$:/theme`, `$:/language`; we'll have a UI for switching layouts that is a peer with the others.

core/modules/widgets/transclude.js
saqimtiaz: @Jermolene good catch, thank you.

saqimtiaz:
@Jermolene updated. Should be good to go. 

I've verified that the state tiddlers for the sidebar tabs remain unchanged, whereas without the recursionMarker set to "no" they change.

Jermolene:
Thanks @saqimtiaz

Jermolene:
I think it would be worth trying to get a page layout switcher UI done for v5.1.23, we should be able to copy the existing implementations. We'll also need to add some metadata fields to the existing $:/core/ui/PageTemplate tiddler for a name and description to be included in the switcher. @BurningTreeC is that something you might be able to look at?

BurningTreeC:
>I think it would be worth trying to get a page layout switcher UI done for v5.1.23, we should be able to copy the existing implementations. We'll also need to add some metadata fields to the existing $:/core/ui/PageTemplate tiddler for a name and description to be included in the switcher. @BurningTreeC is that something you might be able to look at?

I can try over the weekend

saqimtiaz:
I was about to write some basic docs for `HiddenSetting: $:/layout` but I'll hold off until we have a UI in place.

Jermolene:
Great thanks @saqimtiaz @BurningTreeC

BurningTreeC:
@Jermolene - should we ship a second layout with the layoutswitcher?

Jermolene:
> should we ship a second layout with the layoutswitcher?

I think not in the core itself at this point, but we certainly could one or more via plugins. I've been working on the multicolumn flexbox layout in #4473 for a client project and hope to bring that into the core post v5.1.23.

Do you have any thoughts? I made a demo based on https://startbootstrap.com/themes/one-page-wonder/ but didn't go very deep in customising the content via tiddlers.

BurningTreeC:
>Do you have any thoughts? I made a demo based on https://startbootstrap.com/themes/one-page-wonder/ but didn't go very deep in customising the content via tiddlers.

I think we should wait for your multicolumn flexbox layout

kookma:
> I think not in the core itself at this point, but we certainly could one or more via plugins. I've been working on the multicolumn flexbox layout in #4473 for a client project and hope to bring that into the core post v5.1.23.

@Jermolene we discussed here https://github.com/Jermolene/TiddlyWiki5/issues/4473 and before to have a layout for slideshow and I explained the issues with current storyview! Is it possible to have a layout in the context of this PR to support slideshow mode! If so, I would appreciate to have that as an example here.

PR #4985, opened 2020-11-06
Update core "tabs" macros to use explicitState

BurningTreeC:
All of the core `tabs macros` are updated to use the `explicitState` parameter, except the tabs macro of the `$:/core/ui/Components/plugin-info` tiddler, where I think it doesn't make sense to change it

### DIFF
--- a/core/ui/ControlPanel/Modals/AddPlugins.tid
+++ b/core/ui/ControlPanel/Modals/AddPlugins.tid
@@ -148,7 +148,7 @@ $:/state/add-plugin-info/$(connectionTiddler)$/$(assetInfo)$
 
 <$set name="transclusion" value=<<connectionTiddler>>>
 
-<<tabs "[[$:/core/ui/ControlPanel/Plugins/Add/Updates]] [[$:/core/ui/ControlPanel/Plugins/Add/Plugins]] [[$:/core/ui/ControlPanel/Plugins/Add/Themes]] [[$:/core/ui/ControlPanel/Plugins/Add/Languages]]" "$:/core/ui/ControlPanel/Plugins/Add/Plugins">>
+<$macrocall $name="tabs" tabsList="[[$:/core/ui/ControlPanel/Plugins/Add/Updates]] [[$:/core/ui/ControlPanel/Plugins/Add/Plugins]] [[$:/core/ui/ControlPanel/Plugins/Add/Themes]] [[$:/core/ui/ControlPanel/Plugins/Add/Languages]]" default="$:/core/ui/ControlPanel/Plugins/Add/Plugins" explicitState="$:/state/tab/control-panel/plugins/modals/add-plugins"/>
 
 </$set>
 
--- a/core/ui/ControlPanel/Modals/AddPlugins.tid
+++ b/core/ui/ControlPanel/Modals/AddPlugins.tid
@@ -148,7 +148,7 @@ $:/state/add-plugin-info/$(connectionTiddler)$/$(assetInfo)$
 
 <$set name="transclusion" value=<<connectionTiddler>>>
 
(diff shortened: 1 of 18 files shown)


Jermolene:
Thanks @BurningTreeC apologies for the confusion, the idea is to hard code the existing state tiddler titles so that existing code targetting them will work as expected.

BurningTreeC:
Thanks @Jermolene - does that mean I'd have to look up every qualified state tiddler of each of those tabs macros?

Jermolene:
> Thanks @Jermolene - does that mean I'd have to look up every qualified state tiddler of each of those tabs macros?

The way I do it is to open advanced search filter tab and select "All recently modified tiddlers, including system tiddlers", maybe adjust the limit to 5, and then click around the various tabs copying and pasting the titles.

BurningTreeC:
Thanks that's a good idea

BurningTreeC:
@Jermolene - I've now changed the states to use the current state-titles

BurningTreeC:
I've also changed the already set `explicitState`s

BurningTreeC:
In the $:/core/ui/SearchResults tiddler explicitState="$:/state/tab/search-results/sidebar" ... I don't know if changing it makes sense, because the qualified state is different in the sidebar, the top menu search and the link-dropdown search

pmario:
I think using states, that are **not** human readable is a big mistake. It's super brittle, eg: if the transclusion level changes. ...

saqimtiaz:
@pmario since we are hardcoding state titles now, would you recommend breaking backwards compatibility and using human readable titles? That is what this PR from BTC originally did before it was suggested to change it to maintain existing state titles.

Jermolene:
Hi @BurningTreeC somehow things haven't gone quite right. The intention was to keep constructions such as the links to advanced search in the control panel "basics" tab working, but they are currently broken:

https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/ControlPanel/Basics.tid#L10

image

BurningTreeC:
Hi @Jermolene, I didn't see this coming :) but I can resolve the issue by using $:/state/tab--1498284803 as explicitState

Jermolene:
Thanks @BurningTreeC

BurningTreeC:
Thanks @Jermolene - there's a follow-up PR coming for the ControlPanel Basics tab with a small adjustment

core/ui/ControlPanel/Modals/AddPlugins.tid
```diff
@@ -148,7 +148,7 @@ $:/state/add-plugin-info/$(connectionTiddler)$/$(assetInfo)$
 
 <$set name="transclusion" value=<<connectionTiddler>>>
 
-<<tabs "[[$:/core/ui/ControlPanel/Plugins/Add/Updates]] [[$:/core/ui/ControlPanel/Plugins/Add/Plugins]] [[$:/core/ui/ControlPanel/Plugins/Add/Themes]] [[$:/core/ui/ControlPanel/Plugins/Add/Languages]]" "$:/core/ui/ControlPanel/Plugins/Add/Plugins">>
+<$macrocall $name="tabs" tabsList="[[$:/core/ui/ControlPanel/Plugins/Add/Updates]] [[$:/core/ui/ControlPanel/Plugins/Add/Plugins]] [[$:/core/ui/ControlPanel/Plugins/Add/Themes]] [[$:/core/ui/ControlPanel/Plugins/Add/Languages]]" default="$:/core/ui/ControlPanel/Plugins/Add/Plugins" explicitState="$:/state/addplugins/tab-1342078386"/>
```
Jermolene, line 151: This is the cause of #5235 we are erroneously using a fixed tab state tiddler for the add plugins modal

PR #4986, opened 2020-11-06
Tag-picker: focus tag-input after clicking a tag from the dropdown

BurningTreeC:
This PR makes the tag-picker input focus again when clicking a tag to add from the tag-dropdown. This is very convenient when adding more tags by clicking (and not using the keyboard for it)


Jermolene:
Great improvement, thanks @BurningTreeC

Issue #4868, opened 2020-09-29
[BUG] StoryList is shared among clients as of 5.1.22


hoelzro:
**Describe the bug**
If you load a TiddlyWiki served via Node, the story list is shared among all clients.

**To Reproduce**
1. Set up a TiddlyWiki served via Node (5.1.22)
2. Load the wiki in two browser windows (let's call them A and B)
3. Open a tiddler or two in A
4. Observe how the tiddlers are then opened in B after a sync interval
5. Close a tiddler in B
6. Observe how that tiddler is then closed in A after a sync interval

**Expected behavior**
Each client should have its own story list.

**Desktop (please complete the following information):**
 - OS: Desktop Linux (Arch)
 - Browser: Firefox 81

**Additional context**

I think that I've seen other system tiddlers, such as the sidebar tab state, synced as well - that's weird to me since `$:/state/` tiddlers should be excluded via `$:/config/SyncFilter`, so I'll need to try to reproduce this.

**Other reports**

  * #4605
  * https://groups.google.com/g/tiddlywiki/c/ENyfB5ZxPWs/m/Xf3K6UnYAgAJ
  * https://groups.google.com/g/tiddlywiki/c/22z0Qse3amw/m/M_0pXIFaAQAJ
  * https://groups.google.com/g/tiddlywiki/c/cil7qRMSdjw/m/25zvFCNAAQAJ
  * https://groups.google.com/g/tiddlywiki/c/MvHY_eelsuQ

pmario:
There is a setting in the ControlPanel for `$:/DefaultTiddlers` which allows to store / reopen the story: `[list[$:/StoryList]]`. For me this is one of the most important settings. So it should _not_ go away. 

If it has to go away, we need a replacement, that also works in browser "private" mode!

pmario:
I think, this wouldn't be a problem, if the TW server would be a TiddlyWeb server. TiddlyWeb knows about "recipes" and "bags". 

IMO the problems which come up now with the whole multiuser-crux have all been thought through already. ...We have just forgotten them. 

More info see: http://tiddlyweb.com/ ... I'm not saying, that it should be a python implementation. I'm saying we should have a closer look about the "battle tested" workflow and configurations.

pmario:
A better link https://tank.peermore.com/tanks/tiddlyweb/HomePage

hoelzro:
[...]

PR #4987, opened 2020-11-06
System tiddler syncing

### LINKED ISSUE #4829: [BUG] TiddlyWeb/Server Edition "Save snapshot for offline use" files are broken
I run a very custom setup, atm, so I stripped it all back to basics/core when I noticed this.

1. Run the simplest Node.js client-server setup (tiddlyweb+filesystem).
2. Load your wiki, change something, and select the Cloud icon, then select "Save snapshot for offline use".
3. Wait for the file to download.
4. Load the file in a browser.

Result: White page (no wiki renders). Investigating the source shows that a large chunk of javascript is missing compared to a "Ctrl+S" save of the same wiki. Is that the "boot.js"?

Am I just mis-configuring something?

Best,
Joshua F

### LINKED ISSUE #4924: [BUG] Save offline snapshot on pre-release is broken
**Describe the bug**
Using the latest pre-release on node.js, the "save offline snapshot" feature results in a tiddlywiki.html file that does not load in the browser.

**To Reproduce**
1. with the latest pre-release code from github run `node tiddlywiki.js editions/server/ --listen`
2. In the browser, click on the syncer dropdown and choose "save snapshot for offline use".
3. Open the saved HTML file in the browser.
4. The TiddlyWiki does not load, giving a blank white screen and no error messages in the developer console.
5. Inspecting the HTML file in a text editor, it looks like a TW file at first glance.

 - OS: Windows 10
 - Browser: latest Chrome and Firefox

Can anyone else confirm that they are having the same issue? 

Stepping backwards at intervals through the git history has so far failed to identify a specific point or commit that causes the problem. I would like to determine whether this is an issue others are experiencing as well or somehow specific to my environment.

saqimtiaz:
This PR is for discussion only, the code needs clean up and optimization.

@Jermolene this is an attempt at restoring the behaviour we had from 5.1.21, that is the storylist syncs from client to server, but not vice versa. 

The steps needed to sync the story list only from client to server are:
- do not exclude the storylist in `config/syncFilter`
- do exclude it from tiddlyweb filters so its not retrieved from the server.
- make the syncer aware not to delete the storylist on the client, as it assumes its been deleted from the server and should be deleted on the client too.

I'm not entirely happy with this approach as it introduces a second sync related filter `$:/config/syncFromServerFilter` that is used to determine which tiddlers, if absent on the server, should be deleted from the client.  

Not only is that confusing for users, but it feels brittle that to setup a tiddler for one way sync from client to server only we need to modify 3 filters: 2 in tiddlyweb and the new `$:/config/syncFromServerFilter` 

However I'm not sure how else to approach the problem short of hardcoding that system tiddlers should not be synced [as in this proposal.
](https://github.com/Jermolene/TiddlyWiki5/issues/4868#issuecomment-701684972)

The changes:
- remove `-[prefix[$:/StoryList]]` from `$:/config/syncFilter` so that the storylist is saved to the server.
- add `-[prefix[$:/StoryList]]` to the tiddlyweb filters so that the server does not send the story list.
- add a new filter temporarily called `$:/config/syncFromServerFilter` that can be used to exclude deleting a tiddler in the client if it is not sent by the client. This is needed so that the client does not delete the storylist when it is not returned by the server.

Feedback is highly appreciated.

Fixes #4829
Fixes #4924


saqimtiaz:
@Jermolene I realize this is complex one and might require time for some reflection, or you might want to try a different approach.

If you think there's something in here worth discussing just let me know when you're reading.

saqimtiaz:
@joshuafontany what you are saying makes sense. One option for the implementation that has been brought up, is implementing something similar to what tiddlyweb had: bags, where a bag is a user specific storage location for tiddlers that are not shared.

The aim with this PR was to find a short term solution that returned us to the story list behaviour of 5.1.21 - that the story List is saved to the server but not synced from it - until we have the time to formulate a better solution.

We will eventually need a way to specify one way sync too, syncFilter applies in both directions. The implementation in this PR feels far from elegant though.

Jermolene:
Thanks @saqimtiaz @joshuafontany 

I think a reasonable goal of a quick fix for v5.1.23 would be to be able to optionally stop syncing all system tiddlers from server to client, not just $:/StoryList. The trouble is that if the implementation of that option is for users to override a hair-raising series of filters, then we've created the usual problems with subsequent updates to those filters.

So, I'm just wondering if in this case we might be best off accepting a more kludgey solution, which would be to have a single `$:/config/SyncSystemTiddlers=yes|no` flag that overrides whatever system tiddlers are specified in our various filters.

Then, in the future, if we adopt @joshuafontany's suggestions or similar to make it safe to start syncing system tiddlers again then we can just drop the flag.

saqimtiaz:
@Jermolene the syncer deletes all tiddlers not sent by the server, but present on the client. So a global flag for system tiddler sync from the server to the client only helps if the syncer also knows not to delete those tiddlers.

For the story list, those using it in Default Tiddlers want to be able to sync it to the server, but not from the server.

Jermolene:
> @Jermolene the syncer deletes all tiddlers not sent by the server, but present on the client. So a global flag for system tiddler sync from the server to the client only helps if the syncer also knows not to delete those tiddlers.

I was imagining indeed that the syncer would check the flag and only delete system tiddlers if enabled.

> For the story list, those using it in Default Tiddlers want to be able to sync it to the server, but not from the server.

Right, I'm suggesting that the flag would only suppress syncing of system tiddlers going from the server to the client, and that that would be sufficient to restore the old functionality.

saqimtiaz:
@Jermolene right. Ok, now we are on the same page. I'll take a look at it tonight or tomorrow morning.

Jermolene:
Excellent, thanks @saqimtiaz

saqimtiaz:
@Jermolene apologies but it will be Friday evening/Saturday morning before I first get to this. Just realized that I have a full calendar tomorrow.

saqimtiaz:
@Jermolene please have a look, this is re-worked according to our recent discussion.

I've add a config tiddler `$:/config/SyncSystemTiddlersFromServer` that by default is set to `no`.
The title is a mouthful. I wanted to convey that it only controls sync from the sever, and not to the server.

When this config option is set to "no":
- `GET` requests to `/recipes/default/tiddlers.json?filter=<filter>` will have system tiddlers excluded.
- the `syncer` will not delete system tiddlers in the client that are missing on the server.

I've also excluded the `[prefix[$:/boot/]]` files from the `syncFilter` to resolve the problem of the "save offline snapshot" not working. There is no real reason to sync those tiddlers.

Jermolene:
Great thanks @saqimtiaz this looks good.

pmario:
@saqimtiaz ... Is it possible again to save `$:/StoryList` to the server and save `$:/DefaultTiddlers` to the server? 

My `$:/DefaultTiddlers` is: `[list[$:/StoryList]]` ... So page refresh on the client will automatically show me the last story settings???

I really miss the "remember last story" function on my local dev environment.

saqimtiaz:
@pmario I use the same settings and yes this aims to restore that functionality. I did test but another set of eyes confirming the behaviour will be very welcome.

BurningTreeC:
I can confirm that this functionality is restored

joshuafontany:
Just curious, but would this prevent tiddler added by serverside-plugins from propagating out to connected clients? I'm working with a remote inventory API at the moment, and while I could try to limit requests to a startup module (which is working), I wanted some way to refresh the data while the server us up...

Thanks for any insights. :)

saqimtiaz:
@joshuafontany if the system tiddlers are added after the client loads then yes, they will not propagate to the client. However this can be toggled.

The goal here was to restore the functionality from 5.1.21 - that system tiddlers do not sync from server to client - while allowing those that need that ability to toggle the flag.

Post 5.1.23 we will try to implement a better solution to the edge cases of syncing system tiddlers.

Arlen22:
@saqimtiaz @joshuafontany @Jermolene @pmario I'm bringing up a solution for the future over at #5342

PR #4989, opened 2020-11-07
Make type input keyboard-driven

BurningTreeC:
This PR makes the `type` input keyboard-driven. When entering a search term the dropdown gets filtered and the <kbd>Up</kbd> and <kbd>Down</kbd> arrows let you cycle through the filtered list


Jermolene:
Excellent, thanks @BurningTreeC it's also great to have another demo

PR #4991, opened 2020-11-07
Operator is[draft]

### LINKED ISSUE #3999: Request: is[draft]
The `is` operator should have a category to enable filtering `is[draft]`.

I hit the same confusion ever time I want to prevent my various lists from updating as I'm drafting and the most intuitive solution is `!is[draft]`

Yes, there is the `!has[draft.of]` but not only is the `.of` suffix really un-obvious but the whole idea of "has" just doesn't make sense; I want to filter out the draft itself, not the tiddler that has a draft ...not to mention the tiddler that has a draft *of* (???)

saqimtiaz:
Fixes #3999 which indicates we want to implement this unless the opinion on that has changed since.

Will add docs if merged.


Jermolene:
Thanks @saqimtiaz I do still think this makes sense, but we should really replace the instances of `has[draft.of]` that are scattered through the coree.

PR #4992, opened 2020-11-07
Update keyboard-driven type-input actions on "Escape"

BurningTreeC:
This PR makes the keyboard-driven type-input behave a bit more user-friendly when hitting Escape


Jermolene:
Thanks @BurningTreeC

PR #4995, opened 2020-11-07
Update AdvancedSearch tabs-state to use $:/state/tab--1498284803

BurningTreeC:
This PR restores the ability to navigate to the $:/AdvancedSearch Filter tab from within the ControlPanel by using the explicitState `$:/state/tab--1498284803`

### DIFF
--- a/core/ui/AdvancedSearch.tid
+++ b/core/ui/AdvancedSearch.tid
@@ -3,5 +3,5 @@ icon: $:/core/images/advanced-search-button
 color: #bbb
 
 <div class="tc-advanced-search">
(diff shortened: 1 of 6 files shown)


BurningTreeC:
This PR would be simpler if I just changed the tabs-state to $:/state/tab/advanced-search-results in $:/core/ui/ControlPanel/Basics

Jermolene:
Hi @BurningTreeC it's not just about fixing the links in control panel, the point is that because the core uses the technique we should assume that it has been adopted by end users too. There are many good reasons that a plugin might want to switch these core tabs, and we need to continue to support any existing code that does so.

The aim is to freeze all the core tabs state tiddlers at their existing, v5.1.22 titles, not just those for advanced search.

BurningTreeC:
Hi @Jermolene , then #4985 should fix that issue

Jermolene:
Apologies @BurningTreeC I was working through the issues in the wrong order :)

BurningTreeC:
No problem @Jermolene - I'll close this PR

PR #4996, opened 2020-11-07
Add simple layout switcher to ControlPanel

BurningTreeC:


core/ui/PageTemplate.tid
```diff
@@ -1,4 +1,6 @@
 title: $:/core/ui/PageTemplate
+name: Core PageTemplate
```
saqimtiaz, line 2: perhaps "default" instead of core? Let's see what @Jermolene thinks about it.

BurningTreeC: @saqimtiaz - I've updated it to "Default PageTemplate" as the name and "the default tiddlywiki layout" as description

BurningTreeC:
@Jermolene @saqimtiaz , while I'm at it, I could also add a keyboard shortcut that reveals the layoutswitcher on a fixed position (reveaeled from within the $:/core/ui/RootTemplate) - so that if users set a layout with no way back, they can still switch back to the default layout

saqimtiaz:
@BurningTreeC I do like the idea of having a fool proof way of switching layouts or reverting to the default layout. Perhaps this could be a separate PR with a demo we can review it independently. Then even if it wasn't merged it would make for a very useful plugin.

I am also wondering if users are going to ask for a way to switch themes and layouts together, or having a layout specify a theme or vice versa. Not something we need to address at the moment but worth keeping in mind.

Jermolene:
> @Jermolene @saqimtiaz , while I'm at it, I could also add a keyboard shortcut that reveals the layoutswitcher on a fixed position (reveaeled from within the $:/core/ui/RootTemplate) - so that if users set a layout with no way back, they can still switch back to the default layout

Perhaps we should use the modal mechanism to display this switcher so that we can make sure it is available via keyboard shortcut even if the user has switched to a non-functional layout.

We could introduce a new `tm-show-switcher` message that takes a parameter for the target: "layout", "theme", "language", "palette". It would use a common template with custom subtemplates for each switcher type so that we can still show palette swatches, flags etc.

I'm guessing though that we'd still need some hacks to make it work when the root template is set to a blank tiddler.

saqimtiaz:
> Perhaps we should use the modal mechanism to display this switcher so that we can make sure it is available via keyboard shortcut even if the user has switched to a non-functional layout.

The modal was my first thought as well.

> 
> We could introduce a new `tm-show-switcher` message that takes a parameter for the target: "layout", "theme", "language", "palette". It would use a common template with custom subtemplates for each switcher type so that we can still show palette swatches, flags etc.

I quite like this idea.

> 
> I'm guessing though that we'd still need some hacks to make it work when the root template is set to a blank tiddler.

I suggest we document the root template as a Hidden Setting that is not recommended to be changed by users, pointing to $:/layout instead for customization purposes.

BurningTreeC:
@Jermolene - implementing the `tm-show-switcher` message is a bit over my head

saqimtiaz:
@BurningTreeC if you want to give it a go:

- add a handler for `tm-show-switcher` to the root widget. You can use the [handler for tm-modal](https://github.com/Jermolene/TiddlyWiki5/blob/31e7dbf020ddbbc5a7b88cbfc9acdd3168fe72c5/core/modules/startup/rootwidget.js#L25) as an example
- have it accept a parameter that specifies what we are switching: "layout", "theme", "language", "palette"
- the handler triggers a modal where you can switch the desired setting.

saqimtiaz:
@BurningTreeC if you decide to work on it perhaps consider a separate PR, so if it doesn't go as desired it doesn't interfere with your work so far.

BurningTreeC:
@saqimtiaz - I'll have to give it some thoughts and try a bit ... If I'm successfull, I make it a different PR

saqimtiaz:
@BurningTreeC feel free to send questions my way.

 You can also work on a separate branch in your github TW repo, and tag for me review etc, and once happy with it trigger a PR here.

Jermolene:

BurningTreeC:
closing this in favor of #5003

saqimtiaz:
@BurningTreeC just to be clear, we do still want the layout switcher in the Control Panel.

BurningTreeC:
>just to be clear, we do still want the layout switcher in the Control Panel.

Ok @saqimtiaz, I can easily add that in #5003

saqimtiaz:
@BurningTreeC its also OK to leave them separate and just re-open this, whichever makes more sense to you.

PR #4999, opened 2020-11-08
Make Advanced-Search Buttons in Basics.tid update Advanced-Search filter input field

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

PR #5002, opened 2020-11-08
Update modifier Variable.tid with a test-button and...

BurningTreeC:
... a tip for meta-key detection on various operating systems


Jermolene:
Thanks @BurningTreeC

PR #5003, opened 2020-11-08
Add tm-show-switcher message + LayoutSwitcher Modal

BurningTreeC:
This is just a Draft PR for discussion

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -39,6 +39,7 @@ Error/XMLHttpRequest: XMLHttpRequest error code
 InternalJavaScriptError/Title: Internal JavaScript Error
 InternalJavaScriptError/Hint: Well, this is embarrassing. It is recommended that you restart TiddlyWiki by refreshing your browser
 InvalidFieldName: Illegal characters in field name "<$text text=<<fieldName>>/>". Fields can only contain lowercase letters, digits and the characters underscore (`_`), hyphen (`-`) and period (`.`)
+LayoutSwitcher/SubTitle: Layout Switcher
 LazyLoadingWarning: <p>Trying to load external content from ''<$text text={{!!_canonical_uri}}/>''</p><p>If this message doesn't disappear, either the tiddler content type doesn't match the type of the external content, or you may be using a browser that doesn't support external content for wikis loaded as standalone files. See https://tiddlywiki.com/#ExternalText</p>
 LoginToTiddlySpace: Login to TiddlySpace
 Manager/Controls/FilterByTag/None: (none)
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -39,6 +39,7 @@ Error/XMLHttpRequest: XMLHttpRequest error code
 InternalJavaScriptError/Title: Internal JavaScript Error
 InternalJavaScriptError/Hint: Well, this is embarrassing. It is recommended that you restart TiddlyWiki by refreshing your browser
 InvalidFieldName: Illegal characters in field name "<$text text=<<fieldName>>/>". Fields can only contain lowercase letters, digits and the characters underscore (`_`), hyphen (`-`) and period (`.`)
+LayoutSwitcher/Description: open the layout switcher
 LayoutSwitcher/Hint: Layout
 LayoutSwitcher/SubTitle: Layout Switcher
 LazyLoadingWarning: <p>Trying to load external content from ''<$text text={{!!_canonical_uri}}/>''</p><p>If this message doesn't disappear, either the tiddler content type doesn't match the type of the external content, or you may be using a browser that doesn't support external content for wikis loaded as standalone files. See https://tiddlywiki.com/#ExternalText</p>
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -74,6 +74,7 @@ Shortcuts/Input/Tab-Left/Hint: Select the previous Tab
 Shortcuts/Input/Tab-Right/Hint: Select the next Tab
 Shortcuts/Input/Up/Hint: Select the previous item
 Shortcuts/SidebarLayout/Hint: Change the sidebar layout
+Snippets/LanguageSwitcher/Hint: Language
 SystemTiddler/Tooltip: This is a system tiddler
 SystemTiddlers/Include/Prompt: Include system tiddlers
 TagManager/Colour/Heading: Colour
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -64,6 +64,8 @@ MissingTiddler/Hint: Missing tiddler "<$text text=<<currentTiddler>>/>" -- click
 No: No
 OfficialPluginLibrary: Official ~TiddlyWiki Plugin Library
 OfficialPluginLibrary/Hint: The official ~TiddlyWiki plugin library at tiddlywiki.com. Plugins, themes and language packs are maintained by the core team.
+PageTemplate/Description: the default tiddlywiki layout
+PageTemplate/Name: Default ~PageTemplate
 PluginReloadWarning: Please save {{$:/core/ui/Buttons/save-wiki}} and reload {{$:/core/ui/Buttons/refresh}} to allow changes to ~JavaScript plugins to take effect
 RecentChanges/DateFormat: DDth MMM YYYY
 RootTemplate/Reset: Reset the ~RootTemplate
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -39,6 +39,7 @@ Error/XMLHttpRequest: XMLHttpRequest error code
 InternalJavaScriptError/Title: Internal JavaScript Error
 InternalJavaScriptError/Hint: Well, this is embarrassing. It is recommended that you restart TiddlyWiki by refreshing your browser
 InvalidFieldName: Illegal characters in field name "<$text text=<<fieldName>>/>". Fields can only contain lowercase letters, digits and the characters underscore (`_`), hyphen (`-`) and period (`.`)
+LayoutSwitcher/Hint: Layout
 LayoutSwitcher/SubTitle: Layout Switcher
 LazyLoadingWarning: <p>Trying to load external content from ''<$text text={{!!_canonical_uri}}/>''</p><p>If this message doesn't disappear, either the tiddler content type doesn't match the type of the external content, or you may be using a browser that doesn't support external content for wikis loaded as standalone files. See https://tiddlywiki.com/#ExternalText</p>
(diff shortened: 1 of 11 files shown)


saqimtiaz:
@BurningTreeC when you're ready for review let me know. Would also be helpful if you can provide a demo for quick testing.

BurningTreeC:
@saqimtiaz , here's a Demo for testing:

[layout_switcher.zip](https://github.com/Jermolene/TiddlyWiki5/files/5506456/layout_switcher.zip)

BurningTreeC:
@saqimtiaz ,  ready for review

kookma:
@BurningTreeC Thank you!
Just a minor suggestion: Assume you have pressed alt+shift+L and you have the modal window for layout switching! Then again you pressed alt+shift+L you have another windows on top of previous one, and so on...

Is it possible somehow to check if the modal is open prevent reopening the second window? or if you press shortcut in an open Modal windows, Tiddlywiki ignores it!

BurningTreeC:
Hi @kookma 

>Is it possible somehow to check if the modal is open prevent reopening the second window? or if you press shortcut in an open Modal windows, Tiddlywiki ignores it!

That's worth a separate issue. Would you mind creating one?

kookma:
@BurningTreeC  Sure, I create a separate issue!

core/ui/LayoutSwitcher/Layout.tid
```diff
@@ -0,0 +1,20 @@
...
+""">
+<$set name="cls" filter="[all[current]field:title[$:/core/ui/PageTemplate]]" value="tc-chooser-item tc-chosen" emptyValue="tc-chooser-item"><div class=<<cls>>><$link to={{!!title}}>''<$transclude field="name"/>'' - <$transclude field="description"/></$link></div>
+</$set>
+</$list>
+</$list>
+</div>
+</$linkcatcher>
+
+<br>
+<$button>{{$:/language/RootTemplate/Reset}}<$action-deletetiddler $tiddler="$:/core/ui/RootTemplate"/></$button>
```
saqimtiaz, line 20: What if we only display this button when the tiddler $:/core/ui/RootTemplate exists as a non-shadow tiddler?
And maybe call it "Reset to layout defaults".

core/language/en-GB/Misc.multids
```diff
@@ -39,6 +39,9 @@ Error/XMLHttpRequest: XMLHttpRequest error code
 InternalJavaScriptError/Title: Internal JavaScript Error
 InternalJavaScriptError/Hint: Well, this is embarrassing. It is recommended that you restart TiddlyWiki by refreshing your browser
 InvalidFieldName: Illegal characters in field name "<$text text=<<fieldName>>/>". Fields can only contain lowercase letters, digits and the characters underscore (`_`), hyphen (`-`) and period (`.`)
+LayoutSwitcher/Description: open the layout switcher
+LayoutSwitcher/Hint: Layout
+LayoutSwitcher/SubTitle: Layout Switcher
```
saqimtiaz, line 44: I'm unsure about describing it as a "layout switcher" since it lets you switch more than just the layout, i.e. themes, languages etc. @Jermolene thoughts?

saqimtiaz (commented):
Great work @BurningTreeC 

I've left a few comments and perhaps we need some input from Jeremy about terminology but otherwise it is good.

core/ui/LayoutSwitcher/Layout.tid
BurningTreeC: That makes sense! I'll add that check

saqimtiaz:
@BurningTreeC I can't get the button for advanced search to work.

saqimtiaz:
@BurningTreeC I can't get the button for the Control Panel, next to the search field, to bring up the Control Panel.

saqimtiaz:
> @BurningTreeC I can't get the button for the Control Panel, next to the search field, to bring up the Control Panel.

Nevermind, resolved. Problem was on my end.

saqimtiaz:
@Jermolene this looks good to me. Thoughts?

Also what shall we call the Modal considering it switches more than just the layout?

Another thought is that I can see potential for a concept called a Design, where a design is a combination of: layout, theme, palette and language; and the user wants to switch between those pre-configured combinations.

This is entirely doable as a plugin which extends the modal with another tab, its just a case of considering whether this is something we want to address in the core off the bat or leave it to the community at least until we understand usage patterns.

core/modules/startup/rootwidget.js
```diff
@@ -25,6 +25,9 @@ exports.startup = function() {
 	$tw.rootWidget.addEventListener("tm-modal",function(event) {
 		$tw.modal.display(event.param,{variables: event.paramObject, event: event});
 	});
+	$tw.rootWidget.addEventListener("tm-show-switcher",function(event) {
```
pmario, line 28: There is already the "tm-modal" event, that can be used here. I don't think there should be a new event only for this 1 config element, which I personally think will be only used once and then never again.

pmario:
I think the "layout switcher" should be implemented in the exact same way as the "language, theme, palette" switcher. ... As a global button with a dropdown. ... It should only be an global keyboard shortcut, if the button is shown. ... It shouldn't be a global keyboard shortcut that is always active.

core/modules/startup/rootwidget.js
BurningTreeC: I was thinking the same, but maybe I didn't understand how @Jermolene wanted to implement this message-handler

saqimtiaz:
@pmario suggest reading from [here](https://github.com/Jermolene/TiddlyWiki5/pull/4996#issuecomment-723481109) onwards to understand the rationale behind the keyboard shortcut and modal.

As for whether we want to introduce a new message or not, the alternative would be using tm-modal alongside setting state tiddlers to display the correct tab, which is less user friendly. I don't have a personal preference either way.

> I think the "layout switcher" should be implemented in the exact same way as the "language, theme, palette" switcher. ... As a global button with a dropdown

Global button with a dropdown? I have no idea what this means, could you please elaborate?

pmario:
>Global button with a dropdown? I have no idea what this means, could you please elaborate?

open the tools sidebar and click one of the buttons: language, themes, palette and story-view ... That are the dialogues that are used for other "switchers". We should be consistent here.

saqimtiaz:
> It should only be an global keyboard shortcut, if the button is shown. ... It shouldn't be a global keyboard shortcut that is always active.

And how does that address the use case for the keyboard shortcut? Namely, to have a way to switch layouts if you've either broken your layout or switched to one with no means to get back?

pmario:
>And how does that address the use case for the keyboard shortcut?

IMO switching a layout is done at the beginning, when new users try different plugins or themes. ... Once they found them it will never be used again. So having it in the ControlPanel IMO is good enough. 

>Namely, to have a way to switch layouts if you've either broken your layout or switched to one with no means to get back?

That's an argument. OK

core/ui/LayoutSwitcher/Layout.tid
Jermolene: I'm not convinced that we need this reset feature. The UI complicates things, and it's hard to explain satisfactorily to end users because it only makes sense for quite sophisticated users.

core/language/en-GB/Misc.multids
Jermolene: Ideally the modal should be captioned according to the type of thing that is being switched: "Theme Switcher", "Layout Switcher" etc.

Jermolene (commented):

kookma:
> IMO switching a layout is done at the beginning, when new users try different plugins or themes. ... Once they found them it will never be used again. So having it in the ControlPanel IMO is good enough.

There are more use case like using Tiddlywiki for Slideshow where you switch between two layouts: author mode / view mode!
But it needs its own shortcuts and may be considered plugin specific.

By the have LayoutSwitcher Modal is a shortcut quick method for changing layout, theme, ...

core/ui/LayoutSwitcher/Layout.tid
BurningTreeC: Ok @Jermolene, I'll remove this button

core/language/en-GB/Misc.multids
BurningTreeC: @Jermolene, I don't know how to do that, do you have an idea?

core/modules/startup/rootwidget.js
```diff
@@ -25,6 +25,9 @@ exports.startup = function() {
...
+		$tw.modal.display("$:/core/ui/LayoutSwitcher",{variables: event.paramObject, event: event});
```
saqimtiaz, line 29: @BurningTreeC event.paramObject is a hashmap of variables to the modal. So if tm-show-switcher is called with a variable indicating switcher type, that gets passed to the modal and you can use the variable in the caption field of the modal tiddler to get it to display in the modal title.

core/language/en-GB/Misc.multids
saqimtiaz: @BurningTreeC see my comment at https://github.com/Jermolene/TiddlyWiki5/pull/5003/files#r519956303

We would also want to choose which tab to display accordingly which is where it gets trickier I think.

core/modules/startup/rootwidget.js
BurningTreeC: Thanks @saqimtiaz - that's an idea. But should that title change when we change the tabs to say "Theme"? Should it then show "Theme Switcher"?

saqimtiaz: @BurningTreeC I guess yes. Though to be honest, having just taken a fresh look at the demo I am not so sure the title of the modal and the tabs needs to match, otherwise it is just redundant. 

@Jermolene perhaps take a fresh look at the demo. It may be more appropriate to come up with a more generalized name for the modal since the tabs themselves are called after their respective switchers.

saqimtiaz: @Jermolene @BurningTreeC bumping this as it would be good to have for 5.1.23

BurningTreeC: @saqimtiaz I just don't know how to proceed

Jermolene:
Apologies I hadn't looked carefully enough at this PR, and hadn't realised that it contained a set of tabs. They make it like a mini popup version of the control panel, which could be a good thing, but I was thinking of something a bit simpler where we'd only show one of the switcher lists at a time, and adjust the text accordingly. Perhaps we might also use a custom modal style (which would be fun to support anyway):

image

I don't think it's necessarily desperately important to include this in v5.1.23, the main thing was to get the underlying mechanism in.

saqimtiaz:
Speaking of custom modals, post 5.1.23 I would like to rewrite the Modal class to support optional modal templates so the entire layout can be customised.

saqimtiaz:
@BurningTreeC this can be closed now. Good teamwork on this :)

BurningTreeC:
Thanks @saqimtiaz ! You did a great job!

Issue #4998, opened 2020-11-08
[BUG] Pressing Enter in Advanced Search Filter Tab Creates Missing Tiddler


kookma:
**Describe the bug**
The new $:/AdvancedSearch tweaks creates a missing tiddler

**To Reproduce**
1. Go to https://tiddlywiki.com/prerelease/
2. Open $:/AdvancedSearch
3. Open the Filter Tab
3. Enter the [tag[book]tag[AI]] in search input
4. Press Enter key
4. See Tiddlywiki creates a missing tiddler _[tag[book]tag[AI]]_ and navigate to it

_It seems this is the case for all search boxes in $:/AdvancedSearch_

**Screenshots**
![image](https://user-images.githubusercontent.com/830394/98459734-cb155400-21b2-11eb-8d88-b8e7bbf5ed89.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome 86
 - Version TW 5.1.23 prerelease

BurningTreeC:
Hi @kookma, I would say that this is by Design. Pressing <kbd>enter</kbd> in such an input field should navigate to whatever is stored in the input field. I think we should let more users test this and see if they find this behavior convenient or buggy

Jermolene:
Hi @kookma @BurningTreeC funnily enough, I've recently talked to users who were confused that a new tiddler was apparently created when they pressed "enter" with text in the search input. Part of the problem is perhaps that the new keyboard handling makes it more likely that "enter" is pressed accidentally.

So, I'd actually favour hiding this behaviour behind a setting, so that by default pressing "enter" in a search field is ignored unless there is a highlighted search result.

BurningTreeC:
Hi @Jermolene, that is doable. I'll create a PR for that

Jermolene:
Thanks @BurningTreeC

BurningTreeC:
I just don't want to be the one that creates PR #5000 :grin:

BurningTreeC:
@Jermolene , how should that hidden-setting-tiddler be called?

Jermolene:
Thanks @BurningTreeC

Jermolene:
> @Jermolene , how should that hidden-setting-tiddler be called?

Maybe $:/config/Search/NavigateOnEnter/enable set to "yes" or "no"?

BurningTreeC:
[...]

PR #5004, opened 2020-11-08
Make navigating and editing Shadow tiddlers work again

BurningTreeC:
While fixing #4998 I disabled the ability to navigate to shadow tiddlers / edit shadow tiddlers by keyboard (<kbd>Enter</kbd> and <kbd>ctrl-Enter</kbd>) . This PR fixes that


Jermolene:
Thanks @BurningTreeC

PR #5005, opened 2020-11-08
Update Release notes 5.1.23

saqimtiaz:
Updated release notes, @Jermolene this should now be up to date until the present moment (Nov 8 included).

Could probably benefit from some re-arranging and you probably need to decide what should be listed under major improvements.

Currently there are a group of keyboard handling improvements under Usability improvements as well as the main ones listed under major improvements.

For filters, the end-user applicable changes like new filters are under "filter improvements" where as multiple filter operands and named filter run prefix support are under "hackability improvements".

Never realized how much work release notes are! Thank you for all the times you've done this and do feel free to ask the community for assistance in the future. :)

### DIFF
--- a/editions/prerelease/tiddlers/Release 5.1.23.tid
+++ b/editions/prerelease/tiddlers/Release 5.1.23.tid
@@ -24,6 +24,8 @@ type: text/vnd.tiddlywiki
 ! Performance Improvements
 
 * [[Updated|https://github.com/Jermolene/TiddlyWiki5/pull/4659]] templates to use a single VarsWidget instead of several [[SetVariableWidgets|SetVariableWidget]], for improved performance and easier debugging
+* [[Updated|https://github.com/Jermolene/TiddlyWiki5/pull/4954]] ListWidget to not initialize parsers for blank `emptyMessage` attributes.
+* [[Refactored|https://github.com/Jermolene/TiddlyWiki5/pull/4200]] `story.js` to remove dependency on `wiki.js` for story start up and navigator.
 
 ! Usability Improvements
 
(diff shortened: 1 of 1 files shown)


kookma:
@saqimtiaz  @Jermolene I know this is difficult and time consuming but it would be great if it was announced to ask some volunteers  help in a blog (or similar page) to explain these features one by one with enough examples something like below examples

- https://devblogs.microsoft.com/commandline/windows-terminal-tips-and-tricks/
- https://4sysops.com/archives/top-10-new-features-of-powershell-7/

I know this is out of your time or even your interest but the documentation in Tiddlywiki is very concise and sometimes one can not understand what a feature does and why it is there!

saqimtiaz:
@kookma I agree in principle that we would benefit from more thorough documentation for the lay person.

Realistically I don't know what we can do about it. We are a pretty small community and even if we ask for volunteers to write the kind of blog posts you reference there are very few, if any, community members that might have both the technical and pedagogical skillset for it.

Not to mention the vast commitment of time that would be needed. You wouldn't believe how much time, focus and effort it takes just to update the release notes.

kookma:
@saqimtiaz yes, that's quite true and I see your efforts and the time you put here!

Jermolene:
Hi @kookma @saqimtiaz it would indeed be great if we could coordinate a bunch of external blog posts/articles/videos to go along with major new core releases. We have relatively long release cycles for major releases, so there should generally be enough time for people to prepare drafts and get some feedback from the community before a simultaneous release.

kookma:
Hi @Jermolene, As @saqimtiaz explained the writeup to explain new features with enough examples is time consuming and needs interested people with both the technical and pedagogical skillset for it.
Considering all these obstacles, still I think it worth to give a try and see if experienced users interested to help or not.

To start I may recommend to setup a simple mechanism (we need  linear, structured docs) and invite people to help and see how it goes.

Issue #2986, opened 2017-09-25
Width of content Type in Editor too narrow


TiddlyTweeter:
The default width of the input field for content Type in the Editor is simply too narrow. It becomes more problematic if you make custom types with long names. Could it either have adaptive width to length of content or just a much wider default?

![capture](https://user-images.githubusercontent.com/28830541/30808966-6d45162e-a200-11e7-940b-27dcff7e4213.PNG)

Marxsal:
While waiting for response, you can alter this using a stylesheet tiddler with contents:

input.tc-edit-typeeditor { width: 80% ; }

saqimtiaz:
@BurningTreeC would you be up to taking a look at this?

BurningTreeC:
@saqimtiaz - I can have a look

PR #5007, opened 2020-11-08
Make fieldvalue-input a bit larger when viewport is narrow

BurningTreeC:
This PR makes the field-value input field display a bit larger when the viewport is below the sidebarbreakpoint

Also it replaces occurences of {{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}} with `<<sidebarbreakpoint>>` (new macro on top) and fixes some edge-cases when screen-width is exactly 960px by setting the `@media (max-width: ... )` to `<<sidebarbreakpoint-minus-one>>`

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -21,8 +21,8 @@ background-size:` {{$:/themes/tiddlywiki/vanilla/settings/backgroundimagesize}}`
 </$set>
 \end
 
-\define sidebarbreakpoint-plus-one()
-<$text text={{{ [{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}removesuffix[px]add[1]addsuffix[px]] ~[{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}] }}}/>
+\define sidebarbreakpoint()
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@BurningTreeC thank you. Please add this to the description:

Fixes #2986 

It means less work as the issue gets closed automatically when this is merged.

BurningTreeC:
@saqimtiaz , this doesn't fix issue #2986 ... that's a different problem, I'm still investigating

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1435,7 +1439,14 @@ html body.tc-body.tc-single-tiddler-window {
 
 .tc-edit-field-add-value {
 	display: inline-block;
-	width: 35%;
+}
+
+@media (min-width: <<sidebarbreakpoint-plus-one>>) {
+
+	.tc-edit-field-add-value {
```
pmario, line 1446: what is the extra whitespace (new lines) needed for

BurningTreeC: readability :grin:

BurningTreeC:
I've updated the OP

Jermolene:
Thanks @BurningTreeC

PR #5008, opened 2020-11-08
Correct breakpoint for no-box-shadow in snowwhite/base stylesheet

BurningTreeC:
When a tiddlywiki is exactly half-screen on a 1920px screen the sidebar does not yet break but the box-shadows are gone. This PR fixes that


Jermolene:
Thanks @BurningTreeC

PR #5009, opened 2020-11-08
MenuBar: Add fallback value to breakpoint-plus-one and ...

BurningTreeC:
... sidebarbreakpoint-minus-one in case that widths aren't given in px units


Jermolene:
Good thinking @BurningTreeC

PR #5012, opened 2020-11-08
Import UI : Fix issue with size of rename button

saqimtiaz:
Fixes an issue where the rename button size could vary between table rows depending on content flow.

Also optimizes table layout so that there is enough space for the titles when there are long status messages, and that if necessary the titles can break and wrap to the next line.

I would be in favour or re-doing the entire UI for a future release, using DIVs and flexbox, instead of a table, to allow for a more flexible layout.

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -29,7 +29,7 @@ title: $:/core/ui/ImportListing
 <table class="tc-import-table">
 <tbody>
 <tr>
-<th align="left" style="width:10%">
+<th align="left">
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz. I agree about redoing the design of the import tiddler. I've been thinking that we might need another step in the wizard to be able to control how the incoming items are interpreted. Browsers provide clipboard/dragged items in severasl different formats and we currently pick one programmatically. Mostly there's no ambiguity, but sometimes it would be useful to be able to explicitly choose which deserializer is to be used.

saqimtiaz:
@Jermolene yes agreed on the need to be able to change/choose the deserializer being used. A good example is `.json` files from non-TW sources.

I will try to work on this post 5.1.23 as time allows.

PR #5010, opened 2020-11-08
Added disabled attribute support to ButtonWidget

saqimtiaz:
Fixes #3390 and I've needed this a few times myself when wanting to validate user input before saving.

Adds a parameter `disabled` to the ButtonWidget which must be set to "yes" to disable the button.

PR #5014, opened 2020-11-09
Added disabled attribute to input widgets

saqimtiaz:
Adds the disabled attribute to input widgets, à la #5010


Jermolene:
Thanks @saqimtiaz good stuff.

PR #5015, opened 2020-11-09
Extend toggle operator

saqimtiaz:
This PR extends the `toggle[]` operator to accept an optional second parameter to toggle between a value pair.
Examples:

`+[toggle[todo],[done]]`

`[<visibility-state>toggle[show],[hide]]`

The behaviour with a single operand is unchanged.

Tests and documentation updates are included in this PR.


pmario:
Hi Saq. I do have a feature request. 

It would be nice to have a **rotate** operator, that works similar to toggle, but allows more params. eg: "open",  "planned", "active", "done" .. and so on.

After "done" it starts with "open" .... May be it's possible to have a `rotate::minus/plus` or `rotate::left/right` to define the direction.

saqimtiaz:
@pmario interesting idea. I'll throw that on the "to do" pile. There are a lot of interesting possibilities to explore with filters now. :)

kookma:
> @pmario interesting idea. I'll throw that on the "to do" pile. There are a lot of interesting possibilities to explore with filters now. :)

@saqimtiaz, see https://tw-scripts.github.io/Yazd/#macro%2Fcycle this is a small macro I have developed for cycling among several values (like week of the days, colors, numbers, ...)

Jermolene:
Thanks @saqimtiaz 

I like @pmario's idea too.

saqimtiaz:
@Jermolene I like the idea too. 

I am just wondering if it might be better as an extension of this operator as opposed to introducing a rotate/cycle operator. I will investigate as time permits.

Jermolene:
> I am just wondering if it might be better as an extension of this operator as opposed to introducing a rotate/cycle operator. I will investigate as time permits.

Yes that's what I was thinking: a nice example of an operator with a variable number of operands, not merely optional operands.

saqimtiaz:
@pmario @Jermolene Regarding extending `toggle[]` to support a variable number of operands:

It is easy enough to do, I wrote the code in a few minutes and it is just a few extra lines of code at that. 
But I do wonder if it will meet user expectations?

I would expect users to want to derive the list of operands from a single text reference or variable.
Something like:
`[[....input-list...]rotate{textreference-with-list}]
`
or
`[[....input-list...]rotate<subfilter-to-get-list>]
`

instead of

`[[...input-list...]toggle{textreference1},{textreference2},{textreference3}]`

It almost suggests the need for a flag to interpret a singular operand as a list.

kookma:
@saqimtiaz I think toggle semantically means switch between two opposite things like on/off show/hide yes/now reader/author, ...
So I am in favor of new operator like `cycle`, `loop`, `rotate` and  `cycle` is more meaningful.

But, If your are insisted in using toggle for both switching and looping/cycling may be we can use a flag like what is used with `search` operator, so

```
toggle:cycle<subfilter-to-get-list>
```

versus default syntax

```
toggle[val1],[val2]
```

Other examples can be

```
[[....input-list...]toggle:cycle[]]
```

So, `cycle` can also work on input with only two inputs where original toggle is for!

Jermolene:
> But I do wonder if it will meet user expectations?
>
> I would expect users to want to derive the list of operands from a single text reference or variable.
> It almost suggests the need for a flag to interpret a singular operand as a list.
> 
> 

I agree, we should keep the toggle operator with two operands, and consider a separate cycle operator.

saqimtiaz:
@Jermolene I'm leaning the same way. Having a suffix that drastically changes the behaviour of `toggle[]` would just lead to confusion amongst users.

kookma:
> @Jermolene I'm leaning the same way. Having a suffix that drastically changes the behaviour of `toggle[]` would just lead to confusion amongst users.

@saqimtiaz @Jermolene That is very true! You should keep Tiddlywiki grammar as simple as possible. As the basic use of Tiddlywiki (create tiddler, add rich text, add tags and fields) is simple, writing scripts, specially filters shall be kept simple. So, when a complicated logic is used for wikitext/scripting it is difficult to be used.

PR #5016, opened 2020-11-09
Fix height of <p> Element after page-container-wrapper ...

BurningTreeC:
... and before tc-page-container


Jermolene:
Hi @BurningTreeC that new `<p>` tag is a bug with the new layout switching stuff, we should be able to remove it. I'll have a look.

BurningTreeC:
Thanks @Jermolene

PR #5019, opened 2020-11-09
Move styles from menubar plugin to vanilla/base

### LINKED ISSUE #5018: Moving some styles from the menubar plugin to vanilla/base
https://github.com/Jermolene/TiddlyWiki5/blob/59c6f4447e25ef73c0678ee3cf8413cf1357bcd2/plugins/tiddlywiki/menubar/styles.tid#L220

We should add these lines of CSS to vanilla/base

BurningTreeC:
Fixes #5018


Jermolene:
Thanks @BurningTreeC

rmunn:
This is having "interesting" interactions with anchor links; see #5150 for details and a proposed solution.

PR #2251, opened 2016-01-24
New cycle[] and keep[] Operators

matabele:
Two new operators especially designed to be used with the ActionListops
widget — facilitates the cycling of an item from a reference list, and
the display of the currently selected item.

PR #2224, opened 2016-01-15
New cycle() filter operator and associated documentation

matabele:
The `cycle[]` filter operator is designed for use with the ActionListops widget, and cycles the value of the first matching item in the current list, from a list of options specified by the operator operand.

Whilst this is already possible using two ActionListops widgets (one, when cycling tags), the required filter expression is difficult to construct. The new `cycle[]` operator renders the construction of the filter expression trivial.

There is a demo for the `cycle[]`filter operation posted [here] (http://listops.tiddlyspot.com/), where the action of the operator may be explored. 

Suggestions are welcome.

PR #5021, opened 2020-11-10
Cycle operator and refactored toggle operator

saqimtiaz:
@Jermolene @kookma @pmario 
This PR refactors `toggle[]` and introduces a new `cycle[]` operator, following the discussion in #5015 

Both use the same underlying code to avoid duplication. 

`cycle[]` accepts one operand which is parsed as a title list.
`toggle[]` accepts unlimited operands, where each operand is a title.

`cycle[]` also accepts the optional suffix `reverse`

Otherwise their behaviour is identical. We could easily limit `toggle[]` to 2 operands if desired, but the extra functionality comes at no cost.

The output of the following is now identical:

`[[...input-list...]cycle[todo soon now done]]`

`[[...input-list...]toggle[todo],[soon],[now],[done]]`

The advantage with `cycle[]` is that we can pass it a text reference or variable from which to derive the list of values to cycle through.

Feedback and thoughts please.
Once we have consensus I will work on documentation.

PS: for some reason the entirety of x-listops.js uses spaces instead of tabs, will fix in a separate PR.

### DIFF
--- a/core/modules/filters/x-listops.js
+++ b/core/modules/filters/x-listops.js
@@ -16,11 +16,11 @@ Extended filter operators to manipulate the current list.
     Fetch titles from the current list
     */
     var prepare_results = function (source) {
-		var results = [];
-		source(function (tiddler, title) {
-			results.push(title);
-		});
-		return results;
+    var results = [];
+        source(function (tiddler, title) {
+            results.push(title);
+        });
+        return results;
     };
 
     /*
@@ -194,11 +194,11 @@ Extended filter operators to manipulate the current list.
 			nextOperandIndex;		
 		for(i; i < operands.length; i++) {
 			resultsIndex = results.indexOf(operands[i]);
-			if(resultsIndex != -1) {
+			if(resultsIndex !== -1) {
 				break;
 			}
 		}
-		if(resultsIndex != -1) {
+		if(resultsIndex !== -1) {
 			i++;
 			nextOperandIndex = (i === operands.length ? 0 : i);
 			if(operands.length > 1) {
--- a/core/modules/filters/x-listops.js
+++ b/core/modules/filters/x-listops.js
@@ -16,7 +16,7 @@ Extended filter operators to manipulate the current list.
     Fetch titles from the current list
     */
     var prepare_results = function (source) {
-    var results = [];
+		var results = [];
         source(function (tiddler, title) {
             results.push(title);
         });
@@ -187,26 +187,45 @@ Extended filter operators to manipulate the current list.
         }, []);
         return set;
     };
-	
-	/*
-	Toggles an item in the current list.
-	*/
-	exports.toggle = function(source, operator) {
-		var results = prepare_results(source),
-			index = results.indexOf(operator.operand),
-			pairIndex = (operator.operands[1] ? results.indexOf(operator.operands[1]) : -1);
-		if(index === -1) {
-			results.push(operator.operand);
-			if(pairIndex !== -1) {
-				results.splice(pairIndex,1);
-			}
-		} else {
-			results.splice(index,1);
-			if(operator.operands[1]) {
-				results.push(operator.operands[1]);
+
+	var cycleValueInArray = function(results,operands) {
+		var resultsIndex,
+			i = 0,
+			nextOperandIndex;		
+		for(i; i < operands.length; i++) {
+			resultsIndex = results.indexOf(operands[i]);
+			if(resultsIndex != -1) {
+				break;
 			}
 		}
-		return results;
-	};
+		if(resultsIndex != -1) {
+			i++;
+			nextOperandIndex = (i === operands.length ? 0 : i);
+			if(operands.length > 1) {
+				results.splice(resultsIndex,1,operands[nextOperandIndex]);
+			} else {
+				results.splice(resultsIndex,1,);
+			}
+		} else {
+			results.push(operands[0]);
+		}
+		return results;		
+	}
+
+	/*
+	Toggles an item in the current list.
+	*/	
+	exports.toggle = function(source,operator) {
+		return cycleValueInArray(prepare_results(source),operator.operands);
+	}
+
+	exports.cycle = function(source,operator) {
+		var results = prepare_results(source),
+			operands = $tw.utils.parseStringArray(operator.operand, "true");
+		if(operator.suffix === "reverse") {
+			operands.reverse();
+		}
+		return cycleValueInArray(results,operands);
+	}
 	
 })();
--- a/core/modules/filters/x-listops.js
+++ b/core/modules/filters/x-listops.js
@@ -17,10 +17,10 @@ Extended filter operators to manipulate the current list.
     */
     var prepare_results = function (source) {
 		var results = [];
-        source(function (tiddler, title) {
-            results.push(title);
-        });
-        return results;
+		source(function (tiddler, title) {
+			results.push(title);
+		});
+		return results;
     };
 
     /*
@@ -221,7 +221,7 @@ Extended filter operators to manipulate the current list.
 
 	exports.cycle = function(source,operator) {
 		var results = prepare_results(source),
-			operands = $tw.utils.parseStringArray(operator.operand, "true");
+			operands = (operator.operand.length ? $tw.utils.parseStringArray(operator.operand, "true") : [""]);
 		if(operator.suffix === "reverse") {
 			operands.reverse();
 		}


kookma:
Hi @saqimtiaz 

My thoughts:
-  I may recommend to limit `toggle` to switch between opposite values, so a two operands is enough
- The `cycle` as it means loop through a list or array of values
- The `reverse` is nice modifier/flag to go forward or backward

One can argue when you use cycle with a list of two values, it acts the same as `toggle`, but I think toggle is semantic for switching two opposite things and the code will be more readable.

rmunn:
Because `cycle` exists, there's no need for `toggle` to take more than two values, and it would confuse users if it did. OTOH, if it would complicate the code too much to disallow more than 2 values for `toggle`, it might be okay to not check for that and just don't document more than 2 values. But I think it would be as simple as checking the length of `operator.operands`, and if the length is greater than 2, return an error result that says "The toggle operator cannot take more than 2 parameters".

BTW, I'm :+1: on `cycle`. My current TiddlyWiki project is a tool for helping the "GM" ("Game Master" or "storyteller") of a Star Wars role-playing game, and a `cycle` operator would be perfect for helping keep track of combat rounds where each character takes a turn one at a time. Thing is, in the game system I'm using, the combat rounds chart might look like ["player", "player", "enemy 1", "player", "enemy 2"], i.e. there are going to be duplicate "player" entries because the game system does not assign specific action slots to specific players, but lets the players decide who will get a turn when the next player action slot comes up. And the way `cycle` is currently implemented, it does not handle duplicates well: the first `player` entry it finds in the list it will assume is the current one. Because I know about this, I can handle this by making sure that the input I give to `cycle` does not actually contain duplicates, so I'll write ["player 1", "player 2", "enemy 1", "player 3", "enemy 2"] in my list, and then when a "player" result comes up I'll ignore the number when I ask the players who wants to take this action slot.

But for the sake of users who aren't programmers and didn't look at the code, it should probably be documented that the `cycle` operator cannot handle duplicates in its input.

rmunn:
Suggestion: add an optional suffix that's a number of items to cycle. E.g., the `move` operator lets you say `move:2[Wednesday]`, which moves the "Wednesday" list entry two places towards the back of the list. I'd like to be able to say `cycle:2[a b c d e f]` and get `[c d e f a b]` as a result. And then `cycle:reverse` could be written as `cycle:-1` instead, or `cycle:reverse` could still be allowed as a special way of writing `-1`. (There would be no need for `cycle:2:reverse` because that could be written as `cycle:-2`).

rmunn:
There are also scenarios I can come up with where the number of items I would want to cycle is stored in a variable. So maybe instead of a number as a suffix (because a suffix can't be a variable), a number as a second parameter might be more useful. E.g., `cycle[a b c d],<num>` would produce `[b c d a]` if `<num>` contained "1", or `[c d a b]` if `<num>` contained "2".

saqimtiaz:
Limiting `toggle[]` to 2 operands is easy, it's as simple as `operator.operands.slice(0,2) `

@rmunn important to note that both toggle and cycle operate on and modify their input, and not their operand.
`[[done]cycle[done,todo,now,maybe]]` called successively several times would give you:
=> todo => now => maybe => done => todo...

The key here is getting the implementation right such that it is both universally useful and intuitive.

rmunn:
@saqimtiaz Thanks, I was getting mixed up. So in my "whose turn is it in the combat round" example, `[[player-1]cycle[player-1 player-2 enemy-1 player-3 enemy-2]]` would work fine and would produce `player-2`, then if I did the same `cycle` again I'd get `enemy-1`, and so on. But if I had `[[player]cycle[player player enemy player enemy]]`, then the first time through it would find "player" at index 0 and return "player" at index 1. The second time through it would see "player" in the input, and find it at index 0 again, and return "player" when I might have expected it to return "enemy". So I'd still need to make sure that the operand I feed to `cycle` is deduplicated, but as long as I ensure that then I'll have no problems.

BTW, what do you think of either `[[a]cycle:2[a b c d e]]` or `[[a]cycle[a b c d e],[2]]`? (Both of which should return "c").

joshuafontany:
Following along. This is a really really good development. (I also have Gamemaster uses in mind like rotating Initiative/Active-Player for the current combat scene, etc).

You could resolve this duplicate snag if you rotate the order of your list after applying the `cycle` operation to the field or whatever holding the "current value". So, if {{Combat Scene 01!!active-turn}} == "player", and your {{!!initative-list}} has generic duplicates like "player player enemy player player", then:

1) call `[{Combat Scene 01!!active-turn}cycle{Combat Scene 01!!initiative-list}]` in an action.
2) then call `[list[Combat Scene 01!!initiative-list]] +[putlast:1[]]` with a list-ops widget to save the result back to the field.

The initiative list now has "player enemy player player player". The next time you call `[{Combat Scene 01!!active-turn}cycle{Combat Scene 01!!initiative-list}]` it will give you "enemy".

Best,
Joshua F

rmunn:
If you're calling `[list[Combat Scene 01!!initiative-list]] +[putlast:1[]]` with a list-ops widget, then there's no real need to use `cycle` here, because you could just do the `putlast` operation first, then do `[list[Combat Scene 01!!initiative-list]first[]]` to get the current value to put into `{{!!active-turn}}`. And you also need to keep in mind that this would *modify* the initiative list, so if you need to reference what the original initiative order was for any reason, you'd have to save that into another variable (or field) before you start modifying it with `putlast`. (Whereas `cycle` won't modify the list, which is why it's not a duplicate of doing `putlast` + `first` and is a truly useful operator).

But thanks: I had been looking for a list-operation filter that would move X items from the beginning of a list to the end, and I somehow missed `putlast` (and `putfirst`). So you helped me out here.

saqimtiaz:
To summarize for those catching up:

1. `toggle[]`and `cycle[]` use the same underlying method. `Toggle` receives distinct multiple operands, `cycle` receives one operand that is parsed as a title list.
1. Consensus thus far is to limit `toggle[]` to two operands.
2. There is a need to document how to handle situations where the operand for cycle might include duplicate values.
3. There is a request to add the option to specify how many places to move each time. `cycle<list>,[2]`

Regarding `1` since the underlying method supports more operands, I suggest we not artificially limit it but just document how it works and how it differs from cycle. There may be use cases for passing 3+ distinct operands and toggling between them that we aren't thinking of.

Regarding `3.`, my personal feeling is that the usage patterns of the operator aren't that well understood so far, and we should keep it relatively simple until we have more user feedback. This is comes from the consideration of backwards compatibility once the operator is added, and the kind of support requests and confusion that would arise from users moving two places in the list a time with 4 operands. We can always extend the operator in future releases.

Would be good to hear from @pmario and @Jermolene before we move forwards.

pmario:
@saqimtiaz .. I did those tests, using the **text** and the **tags** field from tiddler: **target**

```
\define toggle()
<$action-listops $tiddler="target" $field="tags" $subfilter="+[toggle[done],[active]]" />
\end

\define toggle()
<$action-listops $tiddler="target" $field="text" $filter="[enlist{target}] +[toggle[aa],[xx]]" />
\end

\define cycle()
<$action-listops $tiddler="target" $field="text" $subfilter="+[cycle[a b c d]]" />
\end

\define cycle()
<$action-listops $tiddler="target" $field="text" $filter="[enlist{target}] +[cycle[xx aa bb]]" />
\end

<$button actions=<<toggle>> >toggle done</$button>

<$button actions=<<cycle>> >cycle</$button>
```

There is a little bit of "trial and error" involved, but that can be fixed with good examples.

saqimtiaz:
Resolved conflicts with upstream.

Jermolene:
Thanks @saqimtiaz @kookma @rmunn @joshuafontany

saqimtiaz:
The stepsize parameter has been added as a second operand to the cycle operator as requested and the reverse suffix has been removed.

Please see #5069. I would welcome a second set of eyes on the docs. 

PRs for doc improvements are very welcome.

saqimtiaz:
@rmunn @joshuafontany @kookma @pmario Please see above message. Need another set of eyes on docs for `cycle[]`

kookma:
Hi @saqimtiaz
I am testing the new cycle and toggle operators and prepare some examples for docs. Do you think there is any room to improve?

saqimtiaz:
@kookma There is always room to improve documentation! Thank you.

kookma:
Good! I will post theme here to discuss!

kookma:
@saqimtiaz 

**Test** 

1. start with empty field ( a field gets its value from a filter containing cycle operator)
2. use a step size like 2 with odd number of input titles 


```
<$vars colors="blue red green yellow purple" fld={{test!!text}}>
<$button>Cycle
<$action-setfield $tiddler="test" $field="text" $value={{{ [<fld>cycle<colors>,[2]]}}} />
</$button>
</$vars>

Color: {{test!!text}}
```

**Result**
- To test create a tiddler in https://tiddlywiki.com/prerelease and copy paste above code
- Test 1 failed  if the field is empty at start
- Test 2 cycles through all values note that: step size is even and input titles are odd (when cycle should it always starts from first value)

saqimtiaz:
@kookma `cycle[]` and `toggle[]` are **listops** operators.

Untested but try something like this:

```
<$vars colors="blue red green yellow purple">
<$button>Cycle
<$action-listops $tiddler="test" $field="text" $filter={{{ [enlist{test}cycle<colors>,[2]]}}} />
</$button>
</$vars>

Color: {{test}}
```

saqimtiaz:
Also have a look at the examples that pmario posted above.

kookma:
Well yes, with **listops** they works fine also this code


both your code works

```
<$vars colors="blue red green yellow purple">
<$button>Cycle
<$action-listops $tiddler="test" $field="text" $filter={{{ [enlist{test}cycle<colors>,[2]]}}} />
</$button>
</$vars>

Color: {{test}}
```

and mine works too

```
<$vars colors="blue red green yellow purple" fld={{testt!!text}}>
<$button>Cycle
<$action-listops $tiddler="testt" $field="text" $subfilter="+[cycle<colors>,[2]]" />
</$button>
</$vars>

Color: {{testt!!text}}
```

kookma:
There are some good discussion here:
https://github.com/Jermolene/TiddlyWiki5/pull/2251
https://github.com/Jermolene/TiddlyWiki5/pull/2224
http://listops.tiddlyspot.com/

from the late metabele! and I think they are useful here.

kookma:
Repeated values

**Test 2**

Look at the below test where the list has repeated values. 

```
<$vars colors="blue red blue yellow purple" fld={{testt!!text}}>
<$button>Cycle
<$action-listops $tiddler="testt" $field="text" $subfilter="+[cycle<colors>]" />
</$button>
</$vars>

Color: {{testt!!text}}
```

Also raised by [Tobias Beer](https://github.com/Jermolene/TiddlyWiki5/pull/2224#issuecomment-171897919) 

I think we should keep the code as is now!

kookma:
> Also have a look at the examples that pmario posted above.

Good examples for documentation!

kookma:
**Test 3**
Titles (values) with space

Note to [[light blue]]

```
<$vars colors="blue red [[light blue]] yellow purple" fld={{testt!!text}}>

<$button>Cycle
<$action-listops $tiddler="testt" $field="text" $subfilter="+[cycle<colors>]" />
</$button>

Color: <<fld>>
</$vars>
```

The code works and for light blue returns `[[light blue]]`. It seems this is as expected am I right?

saqimtiaz:
@kookma yes, that is correct. We are dealing with lists so titles with spaces always will be in double brackets.

PR #5024, opened 2020-11-10
Fix whitespace in x-listops.js and tweak toggle operator

saqimtiaz:
This PR fixes the whitespace in x-listops.js, replacing leading spaces with tabs.

It also tweaks the current implementation of `toggle[]` to insert in the same list position as the original value, when toggling between a value pair. Use case example: replace a tiddler in the storylist with another. 
I've also added a test for this.


saqimtiaz:
bump.

Merging this will allow me to resolve conflicts with #5021

Jermolene:
Thanks @saqimtiaz

PR #5027, opened 2020-11-13
Modals: add navigator to handle tm-navigate

### LINKED ISSUE #2188: how make tiddlylinks in a modal open a tiddler?
...and perhaps also close the modal?

### LINKED ISSUE #3560: Links to tiddlers not working inside modals?
I don't know if this is by design or not but I discovered that links to tiddlers inside modals are not working. I mean when I press a link to a tiddler it doesn't open that tiddler. It would be a very helpful ability for me.

saqimtiaz:
Fixes #3560 
Fixes #2188

As discussed in #3560, there was a regression at some point due to which links within modals no longer work. There is no `navigator` widget available to handle the navigate message.

This PR wraps a `navigator` widget around the `transclude` widgets that make up the modal.
This is a cleaned up version of a modification I have been using for a couple of years.

The value for the storylist and historylist are derived from the widget that dispatched the `tm-modal` message. However, users can specify different values for `tv-story-list` and `tv-history-list` in the `tm-modal` message, in which case they take precedence.

This is a common pitfall for both new and experienced users and would be good to resolve. I'm open to exploring a different way of addressing this, if there are any suggestions for a better approach.

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -23,10 +23,7 @@ exports.startup = function() {
 	// Install the modal message mechanism
 	$tw.modal = new $tw.utils.Modal($tw.wiki);
 	$tw.rootWidget.addEventListener("tm-modal",function(event) {
-		var paramObject = (event.paramObject || {});
-		paramObject["tv-story-list"] = paramObject["tv-story-list"] || event.widget.getVariable("tv-story-list");
-		paramObject["tv-history-list"] = paramObject["tv-history-list"] || event.widget.getVariable("tv-history-list");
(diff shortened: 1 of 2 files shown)


saqimtiaz:
@Jermolene would appreciate some feedback on this when you get the opportunity.

Jermolene:
Thanks @saqimtiaz this looks reasonable.

saqimtiaz:
@Jermolene I'll work on docs for Modals sometime this week. We need to document these changes and overall the docs for modals are a bit lacking.

Jermolene:
Great thanks @saqimtiaz

PR #5031, opened 2020-11-13
Make keyboard-driven-input Macro reset to typed input ...

BurningTreeC:
... on up/down when reaching the top/bottom of the list

This PR makes the keyboard-driven-input macro reset the input fields' text to what was typed into the input when we reach the top or bottom of the list by selecting with <kbd>Up</kbd> and <kbd>Down</kbd>


Jermolene:
Thanks @BurningTreeC I think this is a good solution.

PR #5032, opened 2020-11-13
Change tag-picker behavior when there is user-input ...

BurningTreeC:
... in the search field

This PR changes the behavior of the tag-picker slightly when there's user-input to filter the search-dropdown. In that case, after selecting/adding a tag, the dropdown remains filtered by the user-input. An `X` appears beneath the input field that allows resetting the input, <kbd>Escape</kbd> still works the same way. This also fixes the tag-pills not animating when adding them from a filtered dropdown. See the GIF and the Demo for more:

Peek 2020-11-13 19-39

[tag_picker_1.zip](https://github.com/Jermolene/TiddlyWiki5/files/5538781/tag_picker_1.zip)

### DIFF
--- a/core/wiki/macros/tag-picker.tid
+++ b/core/wiki/macros/tag-picker.tid
@@ -16,7 +16,7 @@ second-search-filter: [tags[]is[system]search:title<userInput>sort[]]
 $actions$
 </$list>
 </$set>
-<<delete-tag-state-tiddlers>>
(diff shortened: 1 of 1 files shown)


Jermolene:
Looks good thanks @BurningTreeC

kookma:
@BurningTreeC  Right now the first Esc clear the input, the second tries to close the tiddler (ignore changes)! What do you think to close the drop down tag list in this case?

PR #5035, opened 2020-11-14
Make Tag-Picker Macro work with every field

BurningTreeC:
This PR makes the `tag-picker` macro and the `tags EditTemplate` work with every field. It also updates the `$:/core/ui/ControlPanel/Basics` tab where we now can select the tags for New Tiddlers and New Journal Tiddlers using the tag-picker macro but populating the text field of the corresponding configuration tiddlers


Jermolene:
Thanks @BurningTreeC

kookma:
@BurningTreeC 
I am testing this new feature and got some odd behaviors, please have a look below.


081_%pn

**a.**
After selecting a field name and pressing Enter, the drop down list should be closed like what is happened for mouse click.

**b.** Bug
After adding a field name by entering a partial word  (arrow keys + Enter) and want to add the second field, it only shows the filtered dropdown list based on the partial word of previous step, while the field name input is empty.
It seems if you press the _Esc_ key in the empty input the list restores (there is some hidden state)

BurningTreeC:
Thanks @kookma, I've added PR #5045 to fix this

PR #5037, opened 2020-11-14
Action widget: confirm

saqimtiaz:
While action widgets are incredibly powerful in TW, there is currently no way to prompt a user for confirmation before invoking them. Such a confirmation is incredibly important for situations which might incur data loss, or are not easily reversible.
 
Examples: custom delete buttons, batch operations.

I would like to use this PR to discuss potential solutions.

The code in this PR is a renamed version of an [action widget I've been using in all of my wikis](https://github.com/saqimtiaz/streams/blob/master/plugins/streams/widgets/action-ifconfirmed.js) to be able to prompt a user for confirmation. All action-widgets that require confirmation are wrapped in the `action-confirm` widget.

Usage example:
```
<$button>go
<$action-confirm $message="Do you want to proceed?">
	<$action-setfield $tiddler="lego" $value="chopper"/>
</$action-confirm>
</$button>
```

It accepts two parameters:
- `$message`: the text to display to the user
- `$prompt`: a flag to be able to disable the confirmation from a state tiddler/variable.

I'm open to alternative implementations or approaches.

### DIFF
--- a/core/modules/widgets/confirm.js
+++ b/core/modules/widgets/confirm.js
@@ -57,14 +57,15 @@ ConfirmWidget.prototype.refresh = function(changedTiddlers) {
 Invoke the action associated with this widget
 */
 ConfirmWidget.prototype.invokeAction = function(triggeringWidget,event) {
-	var invokeActions = true;
+	var invokeActions = true,
+		handled = true;
 	if(this.message && this.prompt) {
 		invokeActions = confirm(this.message);
 	}
 	if(invokeActions) {
-		this.invokeActions(triggeringWidget,event);
-		return true;
+		handled = this.invokeActions(triggeringWidget,event);
 	}
+	return handled;
 };
 
 ConfirmWidget.prototype.allowActionPropagation = function() {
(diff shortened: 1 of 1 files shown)


kookma:
@saqimtiaz right now the $:/AdvancedSearch, Filter tab uses a confirmation popup before deleting selected tiddler, but the underlaying code is lengthy and complicated. So having a confirmation mechanism for some critical actions can be useful.

Few use cases: delete tiddler, rename tiddler, delete bulk of tiddler via Filter tab in $:/AdvancedSearch

saqimtiaz:
@kookma thank you for that. I've never had reason to come across that button before! It's a clever way to add a confirmation. 

However as you mention the code is lengthy and convoluted and not the most convenient to use. It becomes even more unwieldy when working with actions that are triggered by a keyboard widget.

However, the big plus with this approach is that the confirmation dialog is entirely created from wikitext and can be customized, compared to the browser default and somewhat ugly prompt dialog I have been using.

I really do think we need an easier way for users to add a confirmation step to invoking actions. Ideally it would somehow combine the ease of use of my approach, with the UI flexibility of the delete button in Advanced Search/Filter.

Jermolene:
Hi @saqimtiaz thanks that's a good idea, and helpful towards our goal of #3967 by bringing more of the functionality of the navigator widget into wikitext.

PR #5042, opened 2020-11-15
Remove image-buttons from type image/svg+xml and add editor-height and stamp button

BurningTreeC:
This PR removes the (useless) image-buttons from tiddlers with type `image/svg+xml` and adds the editor-height and stamp Buttons


Jermolene:
Thanks @BurningTreeC

PR #5044, opened 2020-11-15
Remove redundant fieldmangler widget from tags EditTemplate

BurningTreeC:
This PR removes a redundant `fieldmangler` widget from the `tags` EditTemplate


Jermolene:
Thanks @BurningTreeC

PR #5047, opened 2020-11-15
Docs for action-confirm widget, widget improvements

saqimtiaz:
This PR adds docs for the new `action-confirm` widget, as well as adding a default message for it to display if one is not specified, and improves the logic for disabling the confirmation.


Jermolene:
Thanks @saqimtiaz

PR #5050, opened 2020-11-16
Fix minor mistake in ButtonWidget docs

rmunn:
The ButtonWidget documentation refers to a "state" attribute, but this should probably be "set". The name "state" appears to be an artifact from when the setTitle, setField, and setIndex attributes were added to the widget; at one time they were called stateTitle and so on, and then they were renamed to setTitle instead, but not every instance of "state" was found and renamed to "set". This is one such instance, now fixed.


Jermolene:
Thanks @rmunn, great sleuthing!

Issue #5029, opened 2020-11-13
[IDEA] Add actions parameter to "range widget"


pmario:
The idea is to add action parameterS to the "range widget". 

@saqimtiaz pointed out, that all "interactive" widgets that can have an **actions** parameter, should have 1 or more.

Some thoughts about the specs follow.

## Spec / Ideas

 1. Add info about "click" and "click-release" state ... 
   1.1 This means, there at least 2 actions
 2. While the "handle" is clicked, _no_ actions should be fired
   2.1 This avoids overloading the wiki internal "change" tasks 
   2.2 This avoids update problems if changes are auto-saved to a server
 3. Modifier keys ???

Did I miss something?

pmario:
The html range-input fires 

 - a "change" event, ONLY if the start-value is different to the end-value
 - an "input" event, WHENEVER the slider is moved ... this causes a lot of events and IMO should be avoided. 

## Additional Spec

 - There needs to be a difference between "change" and "input" .. Which means we have 4 actions now.

Jermolene:
Hi @pmario good idea.

> While the "handle" is clicked, no actions should be fired
> 2.1 This avoids overloading the wiki internal "change" tasks

I think that might be a bit restrictive. The refresh cycle is often quick enough for interactive use.

> 2.2 This avoids update problems if changes are auto-saved to a server

I think that sort of throttling would be the responsibility of the sync subsystem.

saqimtiaz:
One consideration here is that IE10/11 do not fire an `input` event on range elements, but instead fire the `change` event on `mousemove`.

Other browsers only fire `change` on `mouseup`, and `input` on `mousemove`.

This naturally restricts us to only being able to invoke actions on `mousemove` rather than `mouseup` if we want to maintain IE support.

Note that `handleInputEvent` already does debouncing and can do the same for invoking actions.

See #4534 for some related issues with event handling for range elements.

Jermolene:
I'm much less worried about IE 10/11 support these days, especially for something like this where generally all IE 10/11 users would lose is some interactivity.

saqimtiaz:
@Jermolene to clarify my point was that if we fire actions on mouse-move, as you suggest for interactive use, then we also get IE10/11 compatibility at no cost.

Jermolene:
Apologies @saqimtiaz, but why wouldn't we just use the "input" event and stay away from "mousemove"?

saqimtiaz:
[...]

PR #5052, opened 2020-11-16
Add Range actions

pmario:
This is WIP and contains some `console.log()` functions so you can see, what's going on with F12

There some redundant code to simplify debugging. ... See code-comments! 

This PR adds 4 actions: 

 - `actionsMouseDown`
 - `actionsMouseUp`
 - `actionsInput`
 - `actionsChange`

It behaves the same way in modern browsers and contains some lines of code to **behave well with IE11**. 
I did test it with IE11 wondows 10, FF latest, Edge latest. .. 

[Go to Latest Version:](https://github.com/Jermolene/TiddlyWiki5/pull/5052#issuecomment-730783577)


pmario:
test code looks like this: 

```
\import range-log-actions

\define logTiddler()$:/temp/log

\define clearLogButton()<$button actions=<<clear-range-log>>>Clear Log</$button>

\define actionsMouseUp()
<$macrocall $name="range-log" text="                    mouse up" value=<<attr-value>> />
\end

\define actionsMouseDown()
<$macrocall $name="range-log" text="                    mouse down" value=<<attr-value>> />
\end

\define actionsInput()
<$macrocall $name="range-log" text="input" value=<<attr-value>> />
\end

\define actionsChange()
<$macrocall $name="range-log" text="                    change" value=<<attr-value>> />
\end


<$range tiddler="$:/temp/range/1__" min="-1" max="10" default="1" increment="1"
  actionsMouseDown=<<actionsMouseDown>> actionsMouseUp=<<actionsMouseUp>>
  actionsInput=<<actionsInput>> actionsChange=<<actionsChange>>
/> {{$:/temp/range/1__}}

<$range tiddler="$:/temp/range/__2" min="0" max="1" default=".01" increment=".01"
  actionsMouseDown=<<actionsMouseDown>> actionsMouseUp=<<actionsMouseUp>>
  actionsInput=<<actionsInput>> actionsChange=<<actionsChange>>
/> {{$:/temp/range/__2}} 

<$checkbox tiddler=<<logTiddler>> field="show" checked="show" unchecked="hide"> Show log</$checkbox> <<clearLogButton>>

<$reveal stateTitle=<<logTiddler>> stateField="show" type="match" text="show">
<$transclude tiddler=<<logTiddler>> />
</$reveal>
```

See: example ZIP for the log-actions code.

pmario:
@twMat @rmunn  ... Have a look at the test-ZIP in the OP, if it fits your needs.

pmario:
This PR will fix: [IDEA] Add actions parameter to "range widget" #5029

core/modules/widgets/range.js
```diff
@@ -92,16 +173,30 @@ RangeWidget.prototype.handleInputEvent = function(event) {
 Compute the internal state of the widget
 */
 RangeWidget.prototype.execute = function() {
+	// TODO remove the next 2 lines once IE is gone!
+	this.mouseUp = true; // Needed for IE10
+	this.isIE = (/msie|trident/i.test(window.navigator.userAgent)) ? true : false;
```
Jermolene, line 178: The core already defines `$tw.browser.isIE`

core/modules/widgets/range.js
```diff
@@ -92,16 +173,30 @@ RangeWidget.prototype.handleInputEvent = function(event) {
...
+	this.defaultValue = this.getAttribute("default","");
 	this.elementClass = this.getAttribute("class","");
 	this.isDisabled = this.getAttribute("disabled","no");
+	// Actions since 5.1.23
+	// Next 3 only fire once!
+	this.actionsMouseDown = this.getAttribute("actionsMouseDown","");
+	this.actionsMouseUp = this.getAttribute("actionsMouseUp","");
+	// Change only fires, if start-value is different to end-value
+	this.actionsChange = this.getAttribute("actionsChange","");
+	// input fires very often!!
```
Jermolene, line 198: It's a small thing but we do try to capitalise comments as sentence case to keep things neat

core/modules/widgets/range.js
```diff
@@ -92,16 +173,30 @@ RangeWidget.prototype.handleInputEvent = function(event) {
...
 	this.increment = this.getAttribute("increment");
-	this.defaultValue = this.getAttribute("default");
+	this.defaultValue = this.getAttribute("default","");
 	this.elementClass = this.getAttribute("class","");
 	this.isDisabled = this.getAttribute("disabled","no");
+	// Actions since 5.1.23
+	// Next 3 only fire once!
+	this.actionsMouseDown = this.getAttribute("actionsMouseDown","");
+	this.actionsMouseUp = this.getAttribute("actionsMouseUp","");
+	// Change only fires, if start-value is different to end-value
```
Jermolene, line 196: The comma isn't necessary here

core/modules/widgets/range.js
```diff
@@ -111,15 +206,17 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 RangeWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes['min'] || changedAttributes['max'] || changedAttributes['increment'] || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.disabled) {
+	if(	changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes['min'] || changedAttributes['max'] ||
```
Jermolene, line 209: I don't mind the condition being split to multiple lines, but we wouldn't usually do it by adding whitespace between the opening parenthesis and the condition.

Jermolene (commented):
Thanks @pmario

core/modules/widgets/range.js
rmunn: Using tab characters for alignment (the "if(" line is really "if(→" where → represents a tab character) is a bad idea, because it ensures that the code will only be aligned if someone else uses tab settings that are the same size as yours, thereby removing the one big advantage of tabs for indentation (that people can set their own tab settings). Here you have the right idea by putting a tab after the "if(", so that people with either 4-space or 8-space tab settings will see the code correctly aligned. But if someone has their tab size set to 2 spaces, then this is no longer aligned.

If you are using tabs for indentation, the only correct approach is "tabs for indentation, spaces for alignment". The second line of the `if(` condition should be indented the same level as the `if(` (here that's two tab characters for two indentation levels), with three spaces after the two tab characters to line up the `changedAttributes` text. I.e., the second line should look like `→→···changedAttributes['increment']`, where → represents a tab character and · represents a space. That is the only correct way to do alignment when you have used tabs for indentation, because it ensures that the aligned lines will stay aligned in everyone's editor no matter what tab settings they use.

And the big _problem_ with "tabs for indentation, spaces for alignment" is that most editors are not smart about it, and will mess it up. Which means that if you've aligned your code, you then have to pay careful attention to your whitespace every time you edit that line. Which is a lot of fiddly work that I, for one, don't care for, and that's why my personal preference is to use space characters, not tabs, for indentation.

There are really only four options:
1. Use tab characters naively, for both indentation and alignment, and end up with code that looks different when different editor settings are applied.
1. Use tab characters intelligently, using tabs for indentation and spaces for alignment, and end up paying far more attention to whitespace than it's worth.
1. Use tab characters for indentation and forget trying to align subsequent lines. The rule for function parameters, second lines of "if()" conditions, and so on, is "just indent one more level with one extra tab character". This will still look reasonably good on everyone's editor, and everyone can use their favorite indentation level by tweaking their editor's tab settings.
1. Use space characters for indentation. This allows you to align function parameters, "if()" conditions, and so on, but other coders won't get to use their favorite indentation level and must conform to the indentation level the project has chosen.

My personal preference is for 4. 1 is just bad, 2 causes you to pay way more attention to whitespace than it's worth, and 3 sacrifices alignment, which is a useful tool in many situations. So I'd rather choose option 4 and sacrifice the ability to use my preferred indentation, instead using whatever the project has settled on.

Point is, it's up to you whether you choose 2 or 3 here (4 is not an option since the TiddlyWiki project uses tabs for indentation), but please don't choose option 1. To choose option 2, write your continuation lines as `→→···condition`. To choose option 3, write your continuation lines as `→→→condition`.

core/modules/widgets/range.js
```diff
@@ -52,8 +52,10 @@ RangeWidget.prototype.render = function(parent,nextSibling) {
 	this.inputDomNode.value = this.getValue();
 	// Add a click event handler
 	$tw.utils.addEventListeners(this.inputDomNode,[
-		{name: "input", handlerObject: this, handlerMethod: "handleInputEvent"},
-		{name: "change", handlerObject: this, handlerMethod: "handleInputEvent"}		
+		{name:"mousedown", handlerObject:this, handlerMethod:"handleMouseDownEvent"},
+		{name:"mouseup",   handlerObject:this, handlerMethod:"handleMouseUpEvent"},
+		{name:"change",    handlerObject:this, handlerMethod:"handleChangeEvent"},
+		{name:"input",     handlerObject:this, handlerMethod:"handleInputEvent"},
```
rmunn, line 58: I'd rather see this list built dynamically based on whether there's an attribute specified for that event. Something like this:

```javascript
var eventList = [];
if(this.actionsMouseDown) {
    eventList.push({name:"mousedown", handlerObject:this, handlerMethod:"handleMouseDownEvent"});
}
// Similar for this.actionsMouseUp and this.actionsChange
if(this.actionsInput) {
    eventList.push({name:"input",     handlerObject:this, handlerMethod:"handleInputEvent"});
}
$tw.utils.addEventListeners(this.inputDomNode, eventList);
```

The advantage is that if the end user doesn't care about the `input` event, we don't need to call a dozen times into an event handler that will end up doing nothing. There is a performance cost to the browser firing events, even if the event handler returns quickly. If we can avoid hooking up the event at all, the browser can optimize and not fire the event because it knows nothing is going to handle the event.

core/modules/widgets/range.js
pmario: > 4. Use space characters for indentation. This allows you to align function parameters, "if()" conditions, and so on, but other coders won't get to use their favorite indentation level and must conform to the indentation level the project has chosen.

TW doesn't compress or minify the source code. That makes it easy for every user to start to read the code and start to debug it.

That means the empty.html would contain 4 times more indentation spaces as used with tabs. This would considerably increase the basic file size. So atm TW uses tabs for a reason.

rmunn: Indeed, TW is one of the few projects where the fact that tabs take less storage than spaces is actually relevant. Most of the time compression makes the space savings of tab characters irrelevant, but in TW it actually matters. I wasn't actually arguing that TW should change (I did mention that option 4 is not an option in this case), I was just trying to list all available options for dealing with indentation & alignment. I agree that using tabs for TW makes sense since it isn't minified (though it probably is compressed, given that most servers default to gzipping HTML for delivery).

Though it *is* actually a little hard for every user to read the code, since when you download TW in HTML form via the big "Download Empty" button, the HTML form has had to store the code in a way that looks like "{\n\t\t\t\tresults.push(title);\n\t\t\t}" and so on. Granted, anyone who knows enough about Javascript to want to modify that code also probably knows how to do a search-and-replace to turn "\n" into a real newline and "\t" into a real tab. But there is a small barrier to entry; it's not quite the ideal situation that you described. :-)

core/modules/widgets/range.js
Jermolene: Good point @rmunn

pmario: The problem with input and change event are, that they have to be fired, even if there is no action. Otherwise the target tiddler wouldn't be updated. It doesn't matter if there are actions defined or not. ... Input is fired on all modern browsers and change is fired in IE.

You are right for MouseDown and MouseUp. ATM they are only there for handling actions atm. ... BUT we don't know if this will change in the future. 

Events are only fired if user wants interaction. So IMO there is no real performance overhead, without user interaction. So I'd vote for simple code here.

pmario: I forgot something. MouseUp is needed for IE to trigger the change event. So it doesn't matter if actions are defined.

pmario:
@Jermolene Should I add the `prepareAttributs` to the basic `widget.js`? ... It can be made more generic and may be renamed to `.prepareActionAttributes()`. So we can probably reduce code duplication. 

https://github.com/Jermolene/TiddlyWiki5/pull/5052/files#diff-10dbab80ae6e1067219cf5b2d25d69fd68bc382029e3056c37feb20ccf4d565cR83

saqimtiaz:
@pmario I think a method on the base widget class makes sense.

Perhaps getActionVariables()?

pmario:
Latest commit 8b25443 uses `actionValue` instead of `attr-value` now. .. Will add `actionValue` for **radio widget** too.

core/modules/widgets/range.js
```diff
@@ -111,15 +202,17 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 RangeWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes['min'] || changedAttributes['max'] || changedAttributes['increment'] || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.disabled) {
+	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes['min'] || changedAttributes['max'] ||
```
saqimtiaz, line 205: @pmario if we need to call` refreshSelf()` if any attributes have changed, it is easier to just check the count:

```
if($tw.utils.count(changedAttributes) > 0) {
    this.refreshSelf();
    return true;
}
```

core/modules/widgets/range.js
```diff
@@ -62,23 +64,102 @@ RangeWidget.prototype.render = function(parent,nextSibling) {
...
+};
+
 RangeWidget.prototype.handleInputEvent = function(event) {
+console.log("input",event);
+
+	this.handleEvent(event);
+	// Trigger actions
+	if(this.actionsInput) {
+		// "tiddler" parameter may be missing. See .execute() below
+		var variables = this.prepareAttributes() // TODO this line will go into the function call below.
```
saqimtiaz, line 157: currently `handleEvent()` does debouncing and only calls `wiki.setText` if the value of the widget has changed. We should similarly debounce the actions, at least the ones for events like `input` that fire very frequently:

```
if(this.getValue() !== this.inputDomNode.value) {
    this.invokeActionString(this.actionsInput,this,event,variables);
}
```

pmario: You are right. Isn't this the case for every widget?

core/modules/widgets/range.js
pmario: Isn't this the case for every widget?

saqimtiaz: @pmario We just can't use this trick when we only want to call `refreshSelf()` on some attributes changing but not for others. So there is probably room for some refactoring and optimization in other core widgets.

core/modules/widgets/range.js
pmario: input or change(on IE) are only fired, if the value has been changed. ... IMO this is no debouncing. 

The line was introduced with commit e84c422e5091c02f55db4027faa9ba840e2aee6c ... But I don't see a difference, if I remove it. ... @Jermolene Do you remember the problem?

saqimtiaz: @pmario according to the spec you should be right. I thought I ran into some issue with this last I worked on this widget. I can do some tests to confirm tomorrow.

saqimtiaz: @pmario I can confirm that the debouncing does not seem necessary on the input event. In both Chrome and FF I cannot get the event to fire more than once per change of value.

saqimtiaz:
@pmario @Jermolene @rmunn  I have some concerns regarding usability from the point of view of end users, and a question in this regard please: 

@rmunn has given some great examples of why having separate actions for the `input` and `change` events would be useful. However, what added benefit is there in making the distinction of having separate actions for `mouseUp` and `mouseDown`,  in addition to actions associated with the `change` and `input` events, and how do we reconcile that with the usability perspective? 

I am worried about how accessible this will be to non-technical users who need to distinguish between 4 types of actions:
- `actionsChange`
- `actionsInput`
- `actionsMouseDown`
- `actionsMouseUp`

If we are handling those extra events (`mousedown` and `mouseup`) internally just for IE compatibility, I suggest we not expose them as widget parameters for distinct actions unless or until there is a specific need to do so.

 Furthermore the current attribute names are perhaps too developer centric, especially if there are four of them. These are user facing widget attributes and not internal variables or method names.

Jermolene:
I'm inclined to agree with @saqimtiaz. If we do want to trap mousedown/move/up, then I think we might be better off making a generic event trapping widget that traps events within its content. That widget could be wrapped around any other widget without that widget being aware of the event handling.

saqimtiaz:
> I'm inclined to agree with @saqimtiaz. If we do want to trap mousedown/move/up, then I think we might be better off making a generic event trapping widget that traps events within its content. That widget could be wrapped around any other widget without that widget being aware of the event handling.

I like this idea! In many ways the `button` widget does this for the `click` event, but is overloaded with other features. 

We could generalize this to a `trap` widget that can trap any type of event we specify and associate actions with them. There is a good case for the `dblclick` and `context` events as well, and this would be far more versatile and flexible than adding support for each event to every widget where it might be needed.

Jermolene:
> > I'm inclined to agree with @saqimtiaz. If we do want to trap mousedown/move/up, then I think we might be better off making a generic event trapping widget that traps events within its content. That widget could be wrapped around any other widget without that widget being aware of the event handling.
> 
> 
> 
> I like this idea! In many ways the `button` widget does this for the `click` event, but is overloaded with other features. 
> 
> 
> 
> We could generalize this to a `trap` widget that can trap any type of event we specify and associate actions with them. There is a good case for the `dblclick` and `contextmenu` events as well, and this would be far more versatile and flexible than adding support for each event to every widget where it might be needed.

I've got a very simple prototype `<$event>` widget for click events, but `<$trap>` is a better name. I'll post it in a moment.

Jermolene:
> I've got a very simple prototype `<$event>` widget for click events, but `<$trap>` is a better name. I'll post it in a moment.

Here's the ticket: #5074

pmario:
I was thinking about it too. So I wanted to implement a `actionStart`, `actionStop` and `action`. Where **start** gets `actionValue` as the start-value. Stop gets `actionValue` and `actioinValueHasChanged` info. So the user has easiy access to the info if the start- and end-value are different. 'action' is activated on input only. ...

It's mad sure, that "start" and "stop" is only called once. and action is called whenever the handle is moved.

pmario:
Latest Version: [range-actions.zip](https://github.com/Jermolene/TiddlyWiki5/files/5570646/range-actions.zip)

It has 3 actions now: 

 - actionsStart ... fires 1x if range-handle is clicked
 - actionsStop ... fires 1x if handle is released
 - actions ... fires if handle is moved. 

`actionValue` is available in the actions 
`actionValueHasChanged` can be "yes" / "no" and shows, if the value between start and stop has changed.

pmario:
Latest Version: [range-actions.zip](https://github.com/Jermolene/TiddlyWiki5/files/5579790/range-actions.zip)

pmario:
Latest Version: [range-actions.zip](https://github.com/Jermolene/TiddlyWiki5/files/5579790/range-actions.zip)

All console.log() and redundant code is removed. see commit: 0111333

core/modules/widgets/range.js
```diff
@@ -62,23 +64,88 @@ RangeWidget.prototype.render = function(parent,nextSibling) {
...
 			} else {
-				value = this.defaultValue || "";
+				value = this.defaultValue;
 			}
 		}
 	}
 	return value;
 };
 
+RangeWidget.prototype.prepareAttributes = function(options) {
```
saqimtiaz, line 83: IMHO this name is not intuitive. In fact we are **processing** attributes to **prepare** variables, rather than preparing attributes.

Perhaps `getActionVariables()` so that the context is easier to undertand?

Jermolene: `getActionVariables` would be more consistent with other usages in the core.

twMat:
Bump.
Would be nice if included in 5.1.23

pmario:
Will close this one and create a new and simplified version.

PR #5054, opened 2020-11-16
add the possibility to dynamically exclude tiddlers from being delive…

rimi:
…red out (e.g. temporarily reduce size by kind of 'hiding' stuff to the browser). one can also add the filter '-[subfilter{$:/config/Server/GlobalExclusionFilter}]' to the root-tiddler to get a synchronous behavior

### DIFF
--- a/core/modules/server/routes/get-tiddlers-json.js
+++ b/core/modules/server/routes/get-tiddlers-json.js
@@ -28,6 +28,7 @@ exports.handler = function(request,response,state) {
 			return;
 		}
 	}
+	filter = filter + "-[subfilter{$:/config/Server/GlobalExclusionFilter}]";
 	var excludeFields = (state.queryParameters.exclude || "text").split(","),
 		titles = state.wiki.filterTiddlers(filter);
 	response.writeHead(200, {"Content-Type": "application/json"});


Jermolene:
Hi @rimi there has been some recent discussion about the need for the server side to be able to control which tiddlers are exposed via `GET /recipes/default/tiddlers.json?filter=<filter>`, and this might be a reasonable way to do it. What do you think @saqimtiaz?

saqimtiaz:
@Jermolene my only concern is whether users could just override this by creating a tiddler called `$:/config/Server/GlobalExclusionFilter` and whether that would matter?

We separately still need a way to decouple sync so that it isn't always bi-directional. The current blocker in the StoryList sync handling is that we have no means of specifying a tiddler that should be saved to the server, but not from the server.
See #4987 for more thoughts on this.

Jermolene:
> @Jermolene my only concern is whether users could just override this by creating a tiddler called `$:/config/Server/GlobalExclusionFilter` and whether that would matter?

That's true, but I think that that boat has already sailed, in that a malicious user can already create system tiddlers that disrupt the operation of the server (eg they can already add tiddlers to accept new filter strings).

> We separately still need a way to decouple sync so that it isn't always bi-directional. The current blocker in the StoryList sync handling is that we have no means of specifying a tiddler that should be saved to the server, but not from the server.
> See #4987 for more thoughts on this.

I'll take a look.

saqimtiaz:
@Jermolene a configurable mechanism like this seems like a good idea in addition to the other changes we have made.

However perhaps a more intuitive tiddler title than `$:/config/Server/GlobalExclusionFilter` ?

Jermolene:
> @Jermolene a configurable mechanism like this seems like a good idea in addition to the other changes we have made.

Now that we've applied the temporary fix to system tiddler syncing I'd be inclined to wait until we revisit the syncer post-v5.1.23 before investigating this.

> However perhaps a more intuitive tiddler title than `$:/config/Server/GlobalExclusionFilter` ?

Yes indeed.

saqimtiaz:
> Now that we've applied the temporary fix to system tiddler syncing I'd be inclined to wait until we revisit the syncer post-v5.1.23 before investigating this.

That seems sensible.

rimi:
Hi,

the reason behind it is, that i used to work in multiple TW-instances (private, work, training-system ...) but realized that this is often no good idea especially if it comes to changes in my own plugins used in all or at leas several of them. So i switched to a mode where i have only one TW-instance but now a lot of data in it. It is separated based on naming/folder etc. clearly and i could load only the data i need for my current interests/jobs to do.

Regards

PR #5055, opened 2020-11-16
Correct for new GitHub default branch

kookma:
This PR corrects the default branch of GitHub from master to main. This is ne GitHub repository configuration from Oct 1st, 2020


Jermolene:
Hi @kookma we also need to change the underlying JS code to implement the change. I'll do both in a moment.

Jermolene:
Hi @kookma I just noticed that the documentation tiddler here covers both the GitHub and GitLab savers. If it's only the former that's changed then we're going to need to rephrase things to clarify the different defaults for the different services.

kookma:
@Jermolene Sure, So what do you recommend? Should I divide this into GitHub and GitLab or only change this line

```
* ''Branch'' - (optional) the name of the branch to be used within the Git repository. Defaults to `main`
```

to
```
* ''Branch'' - (optional) the name of the branch to be used within the Git repository. Defaults to `main` for GitHub, and `master` for GitLab
```

Jermolene:
Hi @kookma we should first confirm that the default branch is still "master" on GitLab. If so, then something like "Defaults to `main` (~GitHub) or `master` (~GitLab)"

kookma:
@Jermolene I just checked the GitLab, its default branch still is `master`. So I edited the tiddler as your recommendation.

Jermolene:
Thanks @kookma

rmunn:
I'm not 100% sure that the default branch on GitHub is going to be `main` for most new projects created. At least the way I create new projects, I run `git init` locally, then create a new repo in GitHub and push my local Git repo to it. This results in a default branch of `master`, not of `main`. The GitHub instructions that appear when the new repo is empty do include a step that would rename the default branch to `main`, but if you're an experienced developer who just skips reading those instructions, you're going to end up with a default branch of `master` unless you take active steps to change that.

So it's possible that the default branch in the GitHub plugin should remain `master` because that's the most commonly-used setting. (I'm not quite sure how to find out whether that's true. I know you can find out the default branch on a GitHub repo by [using the GitHub API to query it](https://stackoverflow.com/a/16501903), but how do I find a list of repos created since October 1st, 2020 to survey what master branch name they're using? And how do I detect whether a repo is a TiddlyWiki repo, which would be most relevant to that survey?)

kookma:
@rmunn - I can confirm that

- Creating new repo in GitHub within **browser** creates `main` as default branch
- Creating new repo in GitHub using **GitHub Desktop** creates `main` as default branch
- Creating new repo in GitHub using `git init` from shell command creates `master` as default branch

rmunn:
> @rmunn - I can confirm that
> 
>     * Creating new repo in GitHub within **browser** creates `main` as default branch

If you initialize the repo with a README file or other data, then yes, I suppose it would. I never choose the "initialize the repo" option because I'm always pushing something I created with `git init`, so I always end up with a `master` default branch even though I've created the repo in GitHub within the browser. But given those results, I suppose the `main` default in the TW config makes sense now, since that's probably going to be what most people end up with.

Another solution would be to say "leave this blank to use the repo's default branch", and then use the GitHub API to query the name of the default branch. Or even have the plugin query the default branch when the user fills in the GitHub repo URL (or owner name & repo name), and then have the "which branch?" section auto-populate once the repo's URL is known.

But if we don't want to go with the API solution, then `main` makes sense as a default now.

Jermolene:
> If you initialize the repo with a README file or other data, then yes, I suppose it would. I never choose the "initialize the repo" option because I'm always pushing something I created with `git init`, so I always end up with a `master` default branch even though I've created the repo in GitHub within the browser. But given those results, I suppose the `main` default in the TW config makes sense now, since that's probably going to be what most people end up with.

That's all very messy, but it does sound like it's mostly going to affect people who know Git well enough to figure things out. We could perhaps tweak the labels to clarify things.

> Another solution would be to say "leave this blank to use the repo's default branch", and then use the GitHub API to query the name of the default branch. Or even have the plugin query the default branch when the user fills in the GitHub repo URL (or owner name & repo name), and then have the "which branch?" section auto-populate once the repo's URL is known.

I do quite like that idea. I think there's a lot of interesting potential in tighter integration with the GitHub API, and so I'd welcome further work in that area to improve our infrastructure and understanding.

Of course, GitHub is still a commercial service, and so we'd want to ensure we retained compatibility with GitLab and Gitea.

It would be great to be able to work with tiddler folders stored in GitHub repos. One could imagine having a CI process that builds all of the tiddlers into an HTML file for GitHub Pages, and then at startup the browser would compare commit hashes to load any tiddlers that have changed in the repo since the HTML file was baked.

PR #5059, opened 2020-11-17
Improve status messages in ImportListing 

### LINKED ISSUE #5048: [BUG] New $:/Import in TW 5.1.23 too much information
**Describe the bug**
I tried to re import a plugin two times and I got the below status message

```
(requires reload) Blocked plugin (due to incoming 2.1.1 being older than existing 2.1.1)
A tiddler with this title already exists. 
```

1. It says required reload, but then says it is blocked. This is confusing!
2. It says incoming 2.1.1 is older than existing 2.1.1, but it is not correct! They are the same
3. It says A tiddler with this title already exists.

I may recommend to use a clear status message here

**To Reproduce**
2. Drag and drop a sample plugin (I did this with my own plugin)
3. Save and reload (if required)
4. Repeat step 2
5. See the message

**Expected behavior**
A clear status message is required.

**Screenshots**
image

 - OS: Windows 10
 - Browser Chrome 10
 - Version TW 5.1.23

saqimtiaz:
As reported in #5048, there is too much information displayed in the status column in the import listing, much of which is redundant, or irrelevant for tiddlers that have been suppressed and cannot be imported.

This PR makes the following changes:
- For suppressed tiddlers that cannot be imported:
  - disables rename button and checkbox, since they are just misleading
  - removes the warning `A tiddler with this title already exists`
  - removes warning `(requires reload)` for plugins that are blocked and cannot be imported.
- adds punctuation to status messages, to improve comprehension when multiple messages are combined.

**Before:** 
image

**After:**
image

Fixes #5048

### DIFF
--- a/core/language/en-GB/Import.multids
+++ b/core/language/en-GB/Import.multids
@@ -18,11 +18,11 @@ Listing/Rename/Prompt: Rename to:
 Listing/Rename/ConfirmRename: Rename tiddler
 Listing/Rename/CancelRename: Cancel
 Listing/Rename/OverwriteWarning: A tiddler with this title already exists.
-Upgrader/Plugins/Suppressed/Incompatible: Blocked incompatible or obsolete plugin
-Upgrader/Plugins/Suppressed/Version: Blocked plugin (due to incoming <<incoming>> being older than existing <<existing>>)
-Upgrader/Plugins/Upgraded: Upgraded plugin from <<incoming>> to <<upgraded>>
-Upgrader/State/Suppressed: Blocked temporary state tiddler
-Upgrader/System/Suppressed: Blocked system tiddler
-Upgrader/System/Warning: Core module tiddler
-Upgrader/System/Alert: You are about to import a tiddler that will overwrite a core module tiddler. This is not recommended as it may make the system unstable
-Upgrader/ThemeTweaks/Created: Migrated theme tweak from <$text text=<<from>>/>
+Upgrader/Plugins/Suppressed/Incompatible: Blocked incompatible or obsolete plugin.
+Upgrader/Plugins/Suppressed/Version: Blocked plugin (due to incoming <<incoming>> not being newer than existing <<existing>>).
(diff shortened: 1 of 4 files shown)


saqimtiaz:
I would favour using background colours for rows as follows:

- a light grey background for disabled rows.
- a light red background for rows that will overwrite existing tiddlers.

Thoughts @pmario @kookma ?

BurningTreeC:
If it's possible I would make the colors definable by the palette @saqimtiaz

saqimtiaz:
> If it's possible I would make the colors definable by the palette @saqimtiaz 

That would definitely be the case @BurningTreeC, though I haven't checked if the default palette has appropriate colours.

BurningTreeC:
>That would definitely be the case @BurningTreeC, though I haven't checked if the default palette has appropriate colours.

I don't think so. I think we need lighter colors, maybe even a light green for tiddlers that are being imported

saqimtiaz:
@BurningTreeC  Another option might be using existing palette colour with a lower opacity.

BurningTreeC:
That's a good idea @saqimtiaz

kookma:
@saqimtiaz great job! Now the messages are relevant and not confusing specially the one for importing a plugin of the same version of existing one.

I like the colorful messages!

Jermolene:
Perhaps this is a case where different styles for even/odd rows might be helpful?

saqimtiaz:
@Jermolene so what I wanted to achieve with colours was:
a) ensure that users are aware of tiddlers that might be overwritten
b) make it easier to skim past tiddlers that are suppressed and cannot be imported.

Having different colors for even/odd rows would make it difficult to use colours for the above purposes.

Anyway, given time I'll test a few options and prepare a separate PR for that, with a demo so we can get feedback.

**This PR is salient and complete and can be merged.**

Jermolene:
That makes sense, thanks @saqimtiaz

PR #5062, opened 2020-11-17
Make sidebarsegments/search-results-count work with all search tabs

BurningTreeC:
This PR makes the "matches" button beneath the sidebar search field show the exact number of found entries (duplicates get counted twice). Further it makes it work with any search-tab if there are more and they use the keyboard-driven-input mechanism


Jermolene:
Thanks @BurningTreeC

PR #5063, opened 2020-11-17
Remove 2em of whitespace from EditTemplate when there are no visible fields

BurningTreeC:
This PR removes the 2em of whitespace between the type input and the fieldname input when there are no fields added

If there are visible fields, the margin-top and margin-bottom of the fields-table are 1em. If there are no fields, the margin-top is 0.5em and the margin-bottom is 0.

This makes the EditTemplate look cleaner.

### DIFF
--- a/core/ui/EditTemplate/fields.tid
+++ b/core/ui/EditTemplate/fields.tid
@@ -57,8 +57,7 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 \whitespace trim
 
 <div class="tc-edit-fields">
-<$set name="excludeList" filter="[all[tiddlers+shadows]prefix[$:/config/EditTemplateFields/Visibility/]contains:text[hide]removeprefix[$:/config/EditTemplateFields/Visibility/]]">
-<table class={{{ [all[current]fields[]] -[enlist<excludeList>] +[count[]] +[!match[0]] +[then[tc-edit-fields]] ~[[tc-edit-fields tc-edit-fields-small]] }}}>
+<table class={{{ [all[current]fields[]] :filter[lookup[$:/config/EditTemplateFields/Visibility/]!match[hide]] +[count[]!match[0]] +[then[tc-edit-fields]] ~[[tc-edit-fields tc-edit-fields-small]] }}}>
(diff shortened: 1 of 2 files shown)


BurningTreeC:
BEFORE:

Bildschirmfoto_2020-11-17_19-44-41

AFTER:

Bildschirmfoto_2020-11-17_19-44-20

Jermolene:
Hi @BurningTreeC this fixes an annoying problem, but it does feel like the solution might be more complex than the problem warrants; it introduces quite a bit of complexity into the base stylesheet. Might it be possible to fix it by tweaking the template to skip the fields entirely if none are present?

BurningTreeC:
>this fixes an annoying problem, but it does feel like the solution might be more complex than the problem warrants; it introduces quite a bit of complexity into the base stylesheet. Might it be possible to fix it by tweaking the template to skip the fields entirely if none are present?

I didn't think about that option. Yes it should be possible, I'll investigate a bit

BurningTreeC:
@Jermolene, it's possible to fix it by tweaking the template, but that makes us loose the pop animation when adding/removing fields. I couldn't find a way to do it otherwise, maybe you have more luck with it :+1:

BurningTreeC:
Oh, I've found a way

BurningTreeC:
@Jermolene - I've now fixed this by tweaking the template (and a little bit of css)

saqimtiaz (commented):
I'm not at my desktop now but I'm wondering if we can avoid the set widget by using the new :filter run prefix?

BurningTreeC:
@saqimtiaz do you have an idea how?

saqimtiaz:
@BurningTreeC Probably an even easier approach that requires no CSS: 

wrap the entire table in a list widget, with a filter that checks the number of fields. If there are no fields, we don't display the table.

BurningTreeC:
@saqimtiaz , I'm now using your filter directly on the table classes, which makes also the animations work

BurningTreeC:
This is ready to get merged @Jermolene

Jermolene:
Awesome, thanks @BurningTreeC @saqimtiaz

PR #5065, opened 2020-11-18
Edit Template: hide empty fields table

saqimtiaz:
Alternative approach to #5063 

@BurningTreeC please confirm if this is OK. If you feel it needs further tweaking, just copy this to your PR and go at it.

The new filter syntax is fun :D

### DIFF
--- a/core/ui/EditTemplate/fields.tid
+++ b/core/ui/EditTemplate/fields.tid
@@ -57,6 +57,7 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 \whitespace trim
 
 <div class="tc-edit-fields">
+<$list filter="[all[current]fields[]] :filter[lookup[$:/config/EditTemplateFields/Visibility/]!match[hide]] +[count[]!match[0]]" variable="temp">
 <table class="tc-edit-fields">
 <tbody>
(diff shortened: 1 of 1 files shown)


BurningTreeC:
@saqimtiaz I'm not at my desk, but I'm wondering if the field add animation when adding the first field works with your approach

saqimtiaz:
@BurningTreeC no idea, I never pay attention to animations. But you could also use the same filter to set a class for the table, without the need for a separate set widget.

pmario:
I don't think, that an animation is needed here for the first field. If I add a new field, I'm already 100% focused to the field area.

If there are many fields already it's OK, since it shows where the new field goes to.

saqimtiaz:
@Jermolene #5063 does the same thing as this, but preserves animation for the first field added, at the cost of adding a CSS class.

I don't have a preference so please choose whatever you consider appropriate.

BurningTreeC:
Hi @saqimtiaz , #5063 preserves animation for ALL field, where this PR here I think removes the animations

saqimtiaz:
This PR results in no animations for the first field added, that is correct.

BurningTreeC:
@saqimtiaz, this PR results in no animations for all fields as far as I have tested now

Issue #4394, opened 2019-12-20
Sort filter: Add support for stripping "stop words" from start of strings


Jermolene:
The sort operator largely exposes the underlying JavaScript `sort()` method, which is not language aware.

This means that strings like "The House" are sorted under "T", rather than "H", which would be conventional for alphabetical indexes (see the image below for an example).

It is proposed to extend the sort filter to support automatically removing "stop words" from the start of each string. The list of stopwords would be language specific, with English stopwords being shipped in the core, and others available via plugins.

<img src="https://user-images.githubusercontent.com/174761/71242128-26bdc100-2305-11ea-873f-de9cba29bc62.jpeg" width="300"/>

pmario:
I need 2 or more sets of stop-words to be active at the same time. 
So imo it's not like the language switcher, which replaces tiddlers.

It's just the sort order, that is changed. The returned list-title is the full title. right?

just brainstorming.

Jermolene:
> I need 2 or more sets of stop-words to be active at the same time.
> So imo it's not like the language switcher, which replaces tiddlers.

My rough thinking was that the sort operator would use tiddlers tagged $:/tags/Sort/StopWords each containing a list of words in their list field. Languages plugins would include a tiddler like $:/language/Sort/StopWords with that tag, and thus the default stop words would be switched as one switched language, but it would be easy enough to modify the list.

> It's just the sort order, that is changed. The returned list-title is the full title. right?

That's right, it means using a sort function that removes any prefixes before doing an ordinary string comparison.

AnthonyMuscio:
Could this be implemented through a wsort operator (Word sort) or tsort (title sort) operator, just as we have nsort so people have to explicitly ask for this StopWords sort. Tiddler Titles may be titles of books, movies and other real world items but given the generic nature of tiddlers and their use in many different applications including as data or status or tag etc... I would think it unwise and possibly not backward compatible unless implemented under its own sourt operator.

Jermolene:
Hi @AnthonyMuscio indeed this would be either an entirely new operator, or a new option on the existing "sort" operator.

twMat:
I think this could be easily solved if filter ops like removeprefix have a *flag* to also allow lazy matching, i.e passing of titles that don't have the prefix.

Current behvariour:

`foo bar baz +[removeprefix[b]] ----> ar az`

Proposed, using flag

`foo bar baz +[removeprefix:lazy[b]] ----> foo ar az`

[...]

PR #5066, opened 2020-11-18
Docs: Added example of how to remove stop words

saqimtiaz:
This PR adds an example of how to use the `search-replace` and `sortsub` operators to remove stop words when sorting a title list. 

Also moves some filter example tiddlers to the correct folder.

Prompted by #4394


Jermolene:
Fantastic thanks @saqimtiaz

PR #5067, opened 2020-11-18
Import UI: Colour code rows

saqimtiaz:
This PR adds some color coding to the Import UI to make it easier to:
- identify tiddlers that might overwrite existing content. 
  - updates if the user chooses to rename the tiddler being imported.
- skim over tiddlers that are disabled and cannot be imported

I have taken a minimal route here, adding alpha values to existing palette colours to adjust opacity.

image

Thoughts @Jermolene @kookma @pmario @BurningTreeC ?

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -46,7 +46,7 @@ title: $:/core/ui/ImportListing
 </th>
 </tr>
 <$list filter="[all[current]plugintiddlers[]sort[title]]" variable="payloadTiddler">
-<tr>
+<tr class={{{[<currentTiddler>has<suppressedField>then[tc-row-disabled]] ~[subfilter<payloadTitleFilter>is[tiddler]then[tc-row-warning]] }}}>
 <td>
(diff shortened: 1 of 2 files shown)


BurningTreeC:
I like @saqimtiaz

kookma:
Thank you @saqimtiaz. I am not sure if the colors come from color palette to follow light/dark modes!

kookma:
Thank you Saq! The import UI above is _semantic_ and really useful! I like it! The disabled blocked plugin rows and the pale colors look great and prevent user confusion with old UI.

Jermolene:
Thanks @saqimtiaz

kookma:
Hi @saqimtiaz 

Just a small issue, it seems the pointer changes when you hover on a disabled row in new import UI

085_%pn



Not a big deal, but it may be a small bug!

Jermolene:
It's because the title is actually rendered as a button, and buttons by default show the hand cursor.

kookma:
Hi @Jermolene
Thank you for clarification!

PR #5068, opened 2020-11-18
Removing extra info from resource tiddlers

kookma:
Sorry for this repeated work! That is because I am not good at making PR.

- The creator/modifier field from resource tiddlers has been removed
- Small typo in Saving to a Git service was corrected


Jermolene:
No worries @kookma I've fixed it in a81b7fc9f4b7da42597c7ebfb1718b7bdfdac032

PR #5078, opened 2020-11-19
Extend action-log and subclass it as log widget

saqimtiaz:
All previous usage patterns of the `<$action-log>` widget have been preserved.

In addition:
- the optional parameter `$$filter` allows using a filter to specify variables to log, in addition to widget attributes. Example: `[prefix[tv-]]`
- the optional parameter `$$message` allows adding a description to the console group. This is helpful when there are multiple `<$action-log>` widgets in use in sequence.
- the optional parameter `$$all` when set to `yes` will log all variables in a collapsed table
- if no widget parameters are used, all variables are logged.

The `<$action-log>` widget has been subclassed as `<$log>` which provides the same features but on render/refresh.

The choice of parameter names using a `$$` prefix is to avoid clashes with `<$action-setfield>` parameters and allowing use cases like Jeremy's, i.e. renaming a `<$action-setfield>` widget to `<$action-log>`

image

PS: I've checked and all modern browsers (including IE 11) support `console.group` and `console.groupCollapsed`.

### DIFF
--- a/core/modules/widgets/action-log.js
+++ b/core/modules/widgets/action-log.js
@@ -14,37 +14,81 @@ Action widget to log debug messages
 
 var Widget = require("$:/core/modules/widgets/widget.js").widget;
 
-var LogWidget = function(parseTreeNode,options) {
+var ActionLogWidget = function(parseTreeNode,options) {
 	this.initialise(parseTreeNode,options);
 };
 
 /*
 Inherit from the base widget class
 */
-LogWidget.prototype = new Widget();
+ActionLogWidget.prototype = new Widget();
 
 /*
 Render this widget into the DOM
 */
-LogWidget.prototype.render = function(parent,nextSibling) {
+ActionLogWidget.prototype.render = function(parent,nextSibling) {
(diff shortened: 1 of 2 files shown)


core/modules/widgets/action-log.js
```diff
@@ -28,23 +28,66 @@ Render this widget into the DOM
...
+		allVars = {},
+		filteredVars;
+
+	$tw.utils.each(this.attributes,function(attribute,name) {
+		if(name.substring(0,2) !== "$$") {
+			data[name] = attribute;
+		}		
+	});
+
+	for(var v in this.variables) {
```
pmario, line 68: I think TW variables walk the prototype chain. Do we need to use hasOwnProperty here?

Jermolene:
Thanks @saqimtiaz

saqimtiaz:
@Jermolene thank you.

I will work on docs later today for this.
After that the only remaining item on my to-do list is the `eventcatcher`

Jermolene:
Great thanks @saqimtiaz

PR #5083, opened 2020-11-20
Revert line-height of 3em on tiddler-edit-title

BurningTreeC:
When titles are long, recognizable on narrow screens, the line-height of 3em is way too high. So it's better to revert this like it was before


Jermolene:
Thanks @BurningTreeC

PR #5084, opened 2020-11-20
Correct error in operator documentation

saqimtiaz:
Sorry, this slipped through despite proofreading several times.
This is part of the reason I wanted to get feedback on docs before the release.


Jermolene:
Thanks @saqimtiaz, at least we can fix docs easily after the release 😃

PR #5085, opened 2020-11-21
Automatic light/dark palette switching through matchMedia Listener and toggle Button

### LINKED ISSUE #5023: [IDEA] A Toggle Button for Simple Switching Dark/Light Theme for Tiddlywiki 5.1.23
Tiddlywiki 5.1.23 has a bunch of wonderful new feature. With the advent of toggle operator and the nice working dark pallets, yet the current trend for having toggle dark/light  on many websites, I may recommend

1. Have a toggle button on the **page control** or **top menu bar**
2. Use nice buttons like https://icones.js.org/collection/uil (see the nice light/dark button top right of screen)
3. Toggle between Vanilla / Gruvbox dark or [Neon ](http://j.d.simplemobile.tiddlyspot.com/)from JD

BurningTreeC:
This is a draft PR that should fix #5023 

It sets `$:/state/dark-mode` to true or false at startup and adds an event listener for when the media-query for `prefers-color-scheme: dark` changes.

It requires `$:/config/AutoDarkMode` to be set to `yes` so that the page switches to whatever palette is stored in `$:/config/Palette/DarkMode` when the OS' dark-mode setting changes

It also adds a toggle button that toggles between the palette in `$:/palette` and the one in `$:/config/Palette/DarkMode`

There's no UI yet, but I'm working on that


BurningTreeC:
Here's a DEMO:

[matchMediaDemo.zip](https://github.com/Jermolene/TiddlyWiki5/files/5577516/matchMediaDemo.zip)

BurningTreeC:
Here's the same DEMO but the framed engine does no more extract the background-colors from the $:/palette... background-colors for the iframe and the domNode are set by the vanilla/base stylesheet:

[matchMediaDemo_2.zip](https://github.com/Jermolene/TiddlyWiki5/files/5577628/matchMediaDemo_2.zip)

kookma:
Hi @BurningTreeC 
Many thanks for all your efforts. Both work fine. 

Both manual and automatic  switch dark/light works. For UI I may suggest

1. two slots to set preferred dark/light palettes
2. two slots to set start/stop time for day or night for automatic switch.

I will suggest to use better icons for day/night for page controls.

Tested: Windows 10, Edge 86, Chrome 86, FF 82.

BurningTreeC:
Hi @kookma and thanks for testing!

Currently the palette chosen for $:/palette is used as the light palette and the one chosen for $:/config/Palette/DarkMode is the dark palette. Therefor only one additional UI chooser is needed, which is the one for the dark palette...

If someone provides me two better icons for dark/light I will add them. I'm not an icon-designer at all...

The idea for automatic day/night switch has come up in the google groups, I've seen that... But I don't know how to implement that correctly in this scenario. Maybe I get an idea

kookma:
> If someone provides me two better icons for dark/light I will add them. I'm not an icon-designer at all...

Is it possible to use free open source icons for example from: https://icones.js.org/
great sun/moon (day/night) icons can be find here https://erikflowers.github.io/weather-icons/

But I am not sure if the policy of TW allow to use such icons.

>The idea for automatic day/night switch has come up in the google groups, I've seen that... But I don't know how to implement that correctly in this scenario. Maybe I get an idea

I am not sure if the bimlas code for automatic calculating day/night time is useful here or not. Please have a look at
https://groups.google.com/g/tiddlywiki/c/-I6brIviJJs/m/zFwiP8wHBAAJ

kookma:
I also announced this PR in the Google group in hope of other people come for testing or helping (Sorry if I did not informed you in advance to do so)

BurningTreeC:
An updated DEMO with a very rudimentary UI in ControlPanel -> palette:

[matchMediaDemo_3.zip](https://github.com/Jermolene/TiddlyWiki5/files/5577761/matchMediaDemo_3.zip)

kookma:
@BurningTreeC 
Would you please use the below attached icons. I can edit them in Inkscape if they have any problem.

[sun-moon.zip](https://github.com/Jermolene/TiddlyWiki5/files/5577779/sun-moon.zip)

BurningTreeC:
@kookma , the same DEMO with your icons:

[matchMediaDemo_4.zip](https://github.com/Jermolene/TiddlyWiki5/files/5577796/matchMediaDemo_4.zip)

kookma:
@ The UI is at the bottom of palettes page and a little difficult to find! What do you think if a dedicated tab is used for setting dark mode! But I wait to see what other says!

The icons are good thank you! but it seems they are a little smaller and their lines are thinner than other core icon ! I will work to see if I can improve them.

Jermolene:
> The icons are good thank you! but it seems they are a little smaller and their lines are thinner than other core icon ! I will work to see if I can improve them.

It is indeed important that the weights of the core icons match consistently. I'll try to work up a couple of icons we can use.

BurningTreeC:
>It is indeed important that the weights of the core icons match consistently. I'll try to work up a couple of icons we can use.

That's great. Thank you @Jermolene!

pmario:
For me demo_4 switches back to light mode, whenever I reload the tab

to reproduce:

 - OS is light mode and will never change
 - Open demo
 - it's light mode
 - click light / dark icon
 - save
 - reload the page
 - it switches to light mode

BurningTreeC:
@pmario - Demo 4 sets Light Mode on startup if `$:/config/AutoDarkMode` is set to yes. Maybe the name is a bit misleading

BurningTreeC:
@pmario - What I wanted to say is: Disable `$:/config/AutoDarkMode` by setting it to `no` and it should not automatically switch to Light Mode on startup if you've set a light theme

BurningTreeC:
An up-to-date DEMO:

[matchMediaDemo_5.zip](https://github.com/Jermolene/TiddlyWiki5/files/5579313/matchMediaDemo_5.zip)

kookma:
> An up-to-date DEMO:
> 
> [matchMediaDemo_5.zip](https://github.com/Jermolene/TiddlyWiki5/files/5579313/matchMediaDemo_5.zip)

Testing Demo_5 on Windows 10 + Edge chromium 86 it works

- Change palettes as : light: Spartan day, dark: Spartan night
  - Save and reload
  - Works fine, no problem
- Change to dark 
  - Save and reload 
  -Works fine, no problem

BurningTreeC:
thanks for testing @kookma !

BurningTreeC:
There are yet three things to decide: how the tiddler `$:/config/AutoDarkMode` should be named, how the UI should look like and where in the ControlPanel it should go

kookma:
> where in the ControlPanel it should go and which should be the default dark theme

This needs to be simple and accessible. I hope you get more feedback here, but I like a better UI!
As I said above the current button are located at the bottom of palette page!
There is a confusion for normal users (where they select day palette? it is not clear now)

One **crude** suggestion:
- keep current palette page
- Use current palette page for editing
- Use three rows above the palette lists in current palette page as
 -- **Day palette** label, a Select drop/down to choose the **day** palette (show names or the snapshot as you designed for dark palette now [The select pallet should be shown  (its name or its name with color squares)
-- **Night palette** label, a Select drop/down to choose the **night** palette (show names or the snapshot as you designed for dark palette now (its name or its name with color squares)

-- A row Enable **automatic** setting of Dark/Light Mode

BurningTreeC:
@kookma - I've updated the DEMO with a fresh new UI in the ControlPanel:

[matchMediaDemo_6.zip](https://github.com/Jermolene/TiddlyWiki5/files/5579692/matchMediaDemo_6.zip)

kookma:
@BurningTreeC 
I tested the new demo and I like it, there is only one questions:

1. Why dark palettes are in Dark Palette tab? and why there are some light palettes in the Light Palette tab?

I may recommend to separate them and putt all dark ones under Dark tab and light ones under Light tab!

BurningTreeC:
>I may recommend to separate them and putt all dark ones under Dark tab and light ones under Light tab!

Hm, I thought it was ok like that, so users can decide to have two light or two dark palettes if they want so

kookma:
I am not sure if I understand you! but in any case we need one light or dark palette! Am I right? But they select their palettes from different tabs! So, if a user don't like automatic dark light palette he/she disable the automatic switch and only choose a palette dark or light! If one like both palettes or like the automatic day/light palette switch, then he/she has to select one dark and one light from related tab!

Jermolene:
I think perhaps @BurningTreeC means that it should be possible to set a light palette for the nighttime and a dark palette for the daytime if a user wanted. Therefore we shouldn't constrain the choice of either palette setting.

Jermolene:
> It sets `$:/state/dark-mode` to true or false at startup and adds an event listener for when the media-query for `prefers-color-scheme: dark` changes.

This should be an $:/info/* tiddler; we may need to update the info mechanism to allow for the event handler.

> It also adds a toggle button that toggles between the palette in `$:/palette` and the one in `$:/config/Palette/DarkMode`

I think it would be more consistent to use `$:/paletteDark` for the dark palette, which will ensure that the two settings will be next to each other in alphabetical listings.

BurningTreeC:
>I think perhaps @BurningTreeC means that it should be possible to set a light palette for the nighttime and a dark palette for the daytime if a user wanted. Therefore we shouldn't constrain the choice of either palette setting.

That's what I meant.

BurningTreeC:
>This should be an $:/info/* tiddler; we may need to update the info mechanism to allow for the event handler.

@Jermolene - is $:/info/os/darkmode appropriate?

>I think it would be more consistent to use $:/paletteDark for the dark palette, which will ensure that the two settings will be next to each other in alphabetical listings.

Ok

How about the configuration tiddler currently called `$:/config/AutoDarkMode` - it enables automatic setting of Light/Dark Mode if set to yes. How should it be called?

Jermolene:
> @Jermolene - is $:/info/os/darkmode appropriate?

I think $:/info/darkmode is sufficiently consistent with what we have.

> How about the configuration tiddler currently called `$:/config/AutoDarkMode` - it enables automatic setting of Light/Dark Mode if set to yes. How should it be called?

That's OK. The idea was that $:/theme, $:/layout, $:/palette etc. are all conceptually $:/config/ tiddlers but have been hoisted to the top level to signify their importance. The rationale for changing $:/paletteDark is to keep the two tiddlers next to each other in alphabetical lists.

BurningTreeC:
Thanks @Jermolene

BurningTreeC:
@Jermolene, if you find the time please have a look how I add the event-listeners in platform.js ... it works fine like this with single file wikis and on nodejs

BurningTreeC:
... what we need is a default dark palette. @Jermolene do you want to decide which one to pick?

Jermolene:
> @Jermolene, if you find the time please have a look how I add the event-listeners in platform.js ... it works fine like this with single file wikis and on nodejs

The info mechanism stores all the $:/info tiddlers into a special plugin of type "info" (see $:/core/modules/startup/info.js) so that they don't clutter up the system tiddler listing. Now that we can dynamically reload plugins we should be able to reload the info plugin.

> ... what we need is a default dark palette. @Jermolene do you want to decide which one to pick?

Of the current palettes, "Spartan Night" seems like the best dark mode match for "Vanilla", what do you think?

BurningTreeC:
>Of the current palettes, "Spartan Night" seems like the best dark mode match for "Vanilla", what do you think?

Yes, I think so, too

>The info mechanism stores all the $:/info tiddlers into a special plugin of type "info" (see $:/core/modules/startup/info.js) so that they don't clutter up the system tiddler listing. Now that we can dynamically reload plugins we should be able to reload the info plugin.

... And my "solution" would just add another listener when reloading the info plugin?

BurningTreeC:
About the dark color scheme ... I'm planning to create one based on macOs Dark Mode using the colors from [here](https://blog.verslu.is/xamarin/ios-macos-dark-mode-dynamic-colors-overview/)

Jermolene:
> ... And my "solution" would just add another listener when reloading the info plugin?

I think a reasonable approach would be for $:/core/modules/startup/info.js to pass a callback to `getInfoTiddlerFields()` that info modules can use to subsequently change the value of an info tiddler.

BurningTreeC:
> I think a reasonable approach would be for $:/core/modules/startup/info.js to pass a callback to `getInfoTiddlerFields()` that info modules can use to subsequently change the value of an info tiddler.

@Jermolene, this is far off my skills. Would you mind taking over?

Jermolene:
> @Jermolene, this is far off my skills. Would you mind taking over?

I'm looking at this now. Making it work has exposed some bugs with the refreshing of shadow tiddlers so it's turned out to be a little more involved.

saqimtiaz:
@BurningTreeC looks like Jeremy has been busy:
https://github.com/Jermolene/TiddlyWiki5/commit/c854e518faa2d2661b7b7278634b10607ab0a5f5

BurningTreeC:
@saqimtiaz Thanks for pointing this out! :)

For what I've seen it's not yet updating the `$:/info/darkmode` tiddler dynamically...

saqimtiaz:
@BurningTreeC I'm a bit unwell today so haven't had a close look. Might still be missing something?

BurningTreeC:
Get well soon @saqimtiaz ! Yes there might still be missing something

saqimtiaz:
> For what I've seen it's not yet updating the `$:/info/darkmode` tiddler dynamically...

Just making sure you've seen this @BurningTreeC https://github.com/Jermolene/TiddlyWiki5/commit/c854e518faa2d2661b7b7278634b10607ab0a5f5#diff-5d89adc6853b56bd6cd71a0fb902df15c66523886dcf770838b7ead486f0fb12R43

Jermolene:
Hi @BurningTreeC @saqimtiaz it's working for me in Chrome, Firefox and Safari on macOS, but not in Safari on iOS. I'm investigating that.

BurningTreeC:
Hi @Jermolene, I tried both on FireFox and Chrome on my Linux desktop and the info tiddler `$:/info/darkmode` doesn't change when I change the preferred color scheme. Seems that this is a refresh issue since my console logs showed that the event gets catched and the tiddler should be updated

Jermolene:
> Hi @Jermolene, I tried both on FireFox and Chrome on my Linux desktop and the info tiddler `$:/info/darkmode` doesn't change when I change the preferred color scheme. Seems that this is a refresh issue since my console logs showed that the event gets catched and the tiddler should be updated

Hi @BurningTreeC did you test with the latest commit 94ffb50e047c47e37b1992ca8a9cecf8d7bcf255?

BurningTreeC:
Yes @Jermolene

saqimtiaz:
@Jermolene the problem seems to be that `$:/config/RegisterPluginType/info` is set to `no`, but needs to be `yes` for the plugin to be reloaded.

BurningTreeC:
> the problem seems to be that $:/config/RegisterPluginType/info is set to no, but needs to be yes for the plugin to be reloaded.

@saqimtiaz , you're right, that solves it

Jermolene:
Thanks, I'd forgotten to commit it :P

BurningTreeC:
@Jermolene may we add an event listener for when the Tiddler $:/config/AutoDarkMode changes and update the info Tiddler in that case?

Jermolene:
> @Jermolene may we add an event listener for when the Tiddler $:/config/AutoDarkMode changes and update the info Tiddler in that case? 

Not quite, the info tiddler should always reflect what the browser is telling us, and can't meaningfully be overwritten.

I think you'll need $:/config/AutoDarkMode for choosing whether dark mode is set automatically or not, and also something like $:/config/DarkModeSetting for the user-specified setting of dark vs. light for use when dark mode is not set automatically.

BurningTreeC:
>I think you'll need $:/config/AutoDarkMode for choosing whether dark mode is set automatically or not, and also something like $:/config/DarkModeSetting for the user-specified setting of dark vs. light for use when dark mode is not set automatically.

when $:/config/AutoDarkMode is set to yes, should the light/dark toggle button then be disabled?

BurningTreeC:
I'll prepare this PR when we have a set of icons that fit well @Jermolene

kookma:
> when $:/config/AutoDarkMode is set to yes, should the light/dark toggle button then be disabled?

Having the button gives this flexibility to switch dark/light whenever you like! I personably think have it enabled!
My two cents!

Jermolene:
Hi @kookma @BurningTreeC I was envisaging a set of three icons that would function as radio buttons:

* Manually engage dark mode
* Manually engage light mode
* Automatically switch between dark/light according to the browser

Does that align with your implementation?

BurningTreeC:
Hi @Jermolene, the light /dark button is currently just a toggle button between light and dark palettes and is disabled when AutoDarkMode is yes. But we can make it three - state which seems to be a better idea

Jermolene:
Hi @BurningTreeC were you thinking of a single button that cycles through the three states? That might be confusing.

An alternative would be to show a single icon in the side bar that would be either the dark or light icon to indicate the current status. Clicking the icon would bring up a dropdown menu with three items dark/light/auto, each accompanied by the appropriate icon, and a tick applied to the current setting.

We might also provide the three individual icons as optional page control buttons/menu bar items, so that people who want single click access to one of the modes can enable those buttons.

BurningTreeC:
>were you thinking of a single button that cycles through the three states? That might be confusing.

yes I was thinking about that, but you may be right that it's a bit confusing

>An alternative would be to show a single icon in the side bar that would be either the dark or light icon to indicate the current status. Clicking the icon would bring up a dropdown menu with three items dark/light/auto, each accompanied by the appropriate icon, and a tick applied to the current setting.

That sounds like a much better idea

>We might also provide the three individual icons as optional page control buttons/menu bar items, so that people who want single click access to one of the modes can enable those buttons.

Ok, yes that sounds appropriate

kookma:
Hi @Jermolene 
I absolutely support optional page control button to toggle dark/light whenever user like! So I am with your last suggestion which @BurningTreeC  also agreed!

Jermolene:
Hi @BurningTreeC thank you for all your work on this. I think it will be a popular feature, and it's worth our attention.

However, I now have some concerns about the our approach. The idea of having a separate palette setting for dark vs. light mode seems reasonable, but it does have a few consequences that are becoming clearer:

* It's kind of annoying that every reference to $:/palette now needs to check for dark mode first. It's also not entirely backwards compatible: any plugin or custom layout that overrides $:/palette will not work as expected if auto dark/light mode is engaged, and $:/paletteDark is being used instead
* The semantics of switching palettes could be confusing: does switching palette mean switching the palette used for either dark mode or light mode, or does it mean engaging a new palette that automatically switches between light and dark mode
* #5123 proposes extending the core palette entry identifiers to include app-specific entries, which makes every palette significantly more complex and unwieldy.

So, I'm tentatively thinking that we should maybe move to a mode where palettes themselves have to be upgraded to support dark vs. light mode. We might have a prefix (like `~`, so it sorts after letters) for dark mode entries that overrides any light mode entry. Then the colour macro switches the prefix on or off according to the current light vs. dark mode setting.

We'd just have $:/palette to identify the current palette and $:/dark "yes"/"no" for switching to the dark mode variant within that palette.

Given an ecosystem of skilfully made palettes, that might all works quite well, but in the meantime we might need to help the transition with eg. tools to merge two palettes, making one the dark palette and one the light palette.

Jermolene:
Another thought: it would be useful to be able to make icons that changed appearance according to dark mode vs. light mode. The easiest way to do that would be to include classes `tc-light-mode-only` and `tc-dark-mode-only` that could be applied to elements of the SVG to control which parts are displayed for different palettes. The foreground/background colours of the SVG would still be set via SVG in the usual way.

That implies that we would have a `tc-dark-mode` class on the body element so that we could make the necessary CSS declarations.

saqimtiaz:
@Jermolene @BurningTreeC  I would suggest that we not rush to get this into 5.1.23 so that we can give it proper consideration.

BurningTreeC:
@Jermolene, thanks for your detailed answer on this :)

As @saqimtiaz suggests, we should postpone this to after 5.1.23 so that we have time to rethink the approach

BurningTreeC:
5.1.23 already is a huge release

Jermolene:
I think that's sensible. Having got the $:/info tiddler in means that others will be able to experiment in the meantime.

Jermolene:
Hi @BurningTreeC just to add that I came across the `color-scheme: dark;` property which causes browsers to render dark mode versions of system controls. Seems to be a much better approach than manually changing button backgrounds etc. See https://web.dev/color-scheme/

PR #5086, opened 2020-11-21
Add: Eventcatcher widget

### LINKED ISSUE #5074: [IDEA] Add a generic event trapping widget
In a project I'm working on I need to display tens of thousands of relatively small, simple tiddlers in a container DIV. Users need to be able to click on a tiddler to perform actions on it. The simplest solution is to wrap each tiddler with a button or link widget, but doing so has a significant impact on rendering performance. It is proposed to add a new event trapping widget that caters for this use case and related situations. There was some discussion about the idea [here](https://github.com/Jermolene/TiddlyWiki5/pull/5052#issuecomment-730206852) too.

The code for the prototype `<$event>` widget is here:

https://gist.github.com/Jermolene/663bc8e96285e7eea7987f58ff6ad948

Here's the event handler which demonstrates the essential features:

* Only handles events whose originating DOM node is contained within a DOM node that matches the provided selector (so `.myclass` will match events originating from a DOM node with the class "myclass")
* The attributes of the selected DOM node are made available to action strings as variables prefixed `dom-`
* I haven't yet needed to, but the idea would also be to copy all the useful properties from the event object into attributes prefixed `event-`
 * We might want to make special cases for the modifier keys and mouse buttons for consistency with existing event handlers

```js
function(event) {
	var selector = self.getAttribute("selector"),
		actions = self.getAttribute("actions"),
		selectedNode = event.target,
		variables = {};
	if(selector) {
		// Search ancestors for a node that matches the selector
		while(!selectedNode.matches(selector) && selectedNode !== domNode) {
			selectedNode = selectedNode.parentNode;
		}
		// If we found one, copy the attributes as variables, otherwise exit
		if(selectedNode.matches(selector)) {
			$tw.utils.each(selectedNode.attributes,function(attribute) {
				variables["dom-" + attribute.name] = attribute.value;
			});
		} else {
			return false;
		}
	}
	// Execute our actions with the variables
	if(actions) {
		self.invokeActionString(actions,self,event,variables);
		event.preventDefault();
		event.stopPropagation();
		return true;
	}
	return false;
}
```

@saqimtiaz suggested `<$trap>` for the name, which I think prefer too.

saqimtiaz:
Fixes #5074

@Jermolene here is a [diff with your original version](https://github.com/saqimtiaz/TiddlyWiki5/compare/ac709ec..eventcatcher)

Changes:
- rename widget to <$eventcatcher>
- added variable for modifer key: `modifier`
- added variable for mouse button clicked: `event-mousebutton`
- added variable for co-ordinate string for popup: `tv-popup-coords`
- allowed updating domNode className without `refreshSelf()`
- allowed optionally specifying domNode tag

Thoughts:
- I don't think we should do anything extra here to accommodate keyboard events. It would add significant complexity and we already have the keyboard widget.
- Do we want to provide the event co-ordinates (`pageX`/`clientX`...) as variables? If so, relative to what? The `eventcatcher` domNode? Raw values will likely not be helpful in actions.  Not knowing how they might be used makes it harder to know what format to provide them in.

### DIFF
--- a/core/modules/widgets/eventcatcher.js
+++ b/core/modules/widgets/eventcatcher.js
@@ -39,10 +39,9 @@ EventWidget.prototype.render = function(parent,nextSibling) {
 		tag = this.elementTag;
 	}	
 	var domNode = this.document.createElement(tag);
+	this.domNode = domNode;
 	// Assign classes
-	var classes = (this["class"] || "").split(" ");
-	classes.push("tc-eventcatcher");
-	domNode.className = classes.join(" ");
+	this.assignDomNodeClasses();	
 	// Add our event handler
 	domNode.addEventListener(this.type,function(event) {
 		var selector = self.getAttribute("selector"),
@@ -59,13 +58,18 @@ EventWidget.prototype.render = function(parent,nextSibling) {
 				$tw.utils.each(selectedNode.attributes,function(attribute) {
 					variables["dom-" + attribute.name] = attribute.value;
 				});
+				//Add a variable with a popup coordinate string for the selected node
+				variables["tv-popup-coords"] = "(" + selectedNode.offsetLeft + "," + selectedNode.offsetTop +"," + selectedNode.offsetWidth + "," + selectedNode.offsetHeight + ")";
+				
 			} else {
 				return false;
 			}
 		}
 		// Execute our actions with the variables
 		if(actions) {
+			// Add a variable for the modifier key
 			variables.modifier = $tw.keyboardManager.getEventModifierKeyDescriptor(event);
+			// Add a variable for the mouse button
 			if(event.button === 0) {
 				variables["event-mousebutton"] = "left";
 			} else if(event.button === 1) {
@@ -83,7 +87,7 @@ EventWidget.prototype.render = function(parent,nextSibling) {
 	// Insert element
 	parent.insertBefore(domNode,nextSibling);
 	this.renderChildren(domNode,null);
-	this.domNodes.push(domNode);
+	this.domNodes.push(domNode);	
 };
 
 /*
@@ -93,10 +97,15 @@ EventWidget.prototype.execute = function() {
 	var self = this;
 	// Get attributes that require a refresh on change
 	this.type = this.getAttribute("type");
-	this["class"] = this.getAttribute("class");
 	this.elementTag = this.getAttribute("tag");
 	// Make child widgets
 	this.makeChildWidgets();
+};
+
+EventWidget.prototype.assignDomNodeClasses = function() {
+	var classes = this.getAttribute("class","").split(" ");
+	classes.push("tc-eventcatcher");
+	this.domNode.className = classes.join(" ");	
 };
 
 /*
@@ -104,9 +113,11 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 EventWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.type || changedAttributes["class"]) {
+	if(changedAttributes.type) {
 		this.refreshSelf();
 		return true;
+	} else if(changedAttributes["class"]) {
+		this.assignDomNodeClasses();
 	}
 	return this.refreshChildren(changedTiddlers);
 };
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz good stuff. `tv-popup-coords` is a good idea.

> * Do we want to provide the event co-ordinates (`pageX`/`clientX`...) as variables? If so, relative to what? The `eventcatcher` domNode? Raw values will likely not be helpful in actions.  Not knowing how they might be used makes it harder to know what format to provide them in.

I was thinking that it might be quite fun to do so. Being able to drag "position: absolute" items around the window would be quite cool. Ultimately we might want to be able to make a simple bitmap editor out of action widgets. Maybe we can provide one pair of variables with the coordinates relative to the `<$eventcatcher>` DOM node, and another relative to the selected DOM node.

saqimtiaz:
> > * Do we want to provide the event co-ordinates (`pageX`/`clientX`...) as variables? If so, relative to what? The `eventcatcher` domNode? Raw values will likely not be helpful in actions.  Not knowing how they might be used makes it harder to know what format to provide them in.
> 
> I was thinking that it might be quite fun to do so. Being able to drag "position: absolute" items around the window would be quite cool. Ultimately we might want to be able to make a simple bitmap editor out of action widgets. Maybe we can provide one pair of variables with the coordinates relative to the `<$eventcatcher>` DOM node, and another relative to the selected DOM node.

Sounds good. I'll update in a short while.

saqimtiaz:
> Maybe we can provide one pair of variables with the coordinates relative to the `<$eventcatcher>` DOM node, and another relative to the selected DOM node.

I welcome suggestions for intuitive names for these.

I have considered `event-posx-selected` and `event-posx-catcher` for x co-ordinate respective to the selected DOM node and the event catcher DOM node. My concern is they could be easily confused as the co-ordinate for those DOM Nodes.

saqimtiaz:
I've added the following variables:
- `tv-selectednode-posx`, `tv-selectednode-posy`, `tv-selectednode-width`, `tv-selectednode-height` with the offset left, top, width and height of the selected DOM node. This is equivalent to the information used to create the `tv-popup-coords` variable, just exposed differently.
- `event-fromselected-posx` and `event-fromselected-posy` for the `X` and `Y` position of the event relative to the selected node.
- `event-fromcatcher-posx` and `event-fromcatcher-posy` for the `X` and `Y` position of the event relative to the catcher widget node.

Jermolene:
Terrific thanks @saqimtiaz

saqimtiaz:
@Jermolene I will work on documentation for the `<$eventcatcher>` tomorrow. 

This will be a tricky one to document and explain well, so hopefully we can get some more input on the documentation too.

saqimtiaz:
@Jermolene PS: thank you for this! This is the most excited I have been about something new in TW for quite some time!

Jermolene:
And thank you @saqimtiaz for many really helpful contributions, including this one!

In terms of the docs, I guess we can assume a little familiarity with HTML/CSS/JS; it's hard to imagine it being useful to people without that.

saqimtiaz:
@Jermolene just wanted to share some thoughts and get an opinion, so posting here instead of creating a new issue. 

It occurred to me that the `eventcatcher` could easily be tweaked to support catching multiple events as long as the `selector` is the same. 

The alternative of course is nesting multiple eventcatcher widgets. I get this might be a step too far in terms of complexity for what we want in the core, but I just wanted to share and get your thoughts on this approach. Are there any potential problems that I am not forseeing?

**Background:** I started planning refactoring my Streams plugin to take advantage of the `eventcatcher` today. I need to catch multiple events from the same domNodes: `dblclick`, `contextmenu`, `swiperight`/`swipeleft` (custom events)

**Example code:**

`<$eventcatcher types="click contextmenu" actionsclick=<<actionsclick>> actionscontextmenu=<<actionscontextmenu>>  selector=".item">`

So `types` is a space separated list of events and each event maps to an attribute of the name `actions+type`

Here is diff from some very quick refactoring of the widget, essentially just put 
 the code setting up the event listener in a loop: https://github.com/saqimtiaz/TiddlyWiki5/commit/a3f0e6b913c5a27845879e34c43a06ff0ae04ddd#diff-d214504f7c00351878f798da313df9b8b441f878d297c8325f045c6471d3eae3R46

Jermolene:
Hi @saqimtiaz with more spooky synchronicity, I was thinking the same thing (I was playing about with making a dragging rectangles thing), although I'd been thinking in terms of sharing a single action attribute and it makes much more sense to maintain separate ones.  I think it may be worth trying to get that into v5.1.23 because it's not 100% backwards compatible.

saqimtiaz:
@Jermolene I'm glad I didn't hold back on posting this! I should be able to polish up the code and do a bit more testing and submit a PR tomorrow.

I suggest the action attributes take the form `"actions-"+event` for easier legibility, for example `actions-dblclick` and `actions-contextmenu`.

Jermolene:
> I suggest the action attributes take the form `"actions-"+event` for easier legibility, for example `actions-dblclick` and `actions-contextmenu`.

Good thinking.

PR #5087, opened 2020-11-21
Reveal widget allow re-positioning

saqimtiaz:
Update the **RevealWidget** so that popup position is updated when the **state tiddler** changes. 

This is needed if wanting to use a single `RevealWidget` to create a popup that can be re-used amongst a list of popup trigger items. Clicking a new trigger just re-positions and refreshes the open `RevealWidget` (if any). This behaviour is triggered by the new widget attribute `updatePopupPosition` and is therefore backwards compatible.

@Jermolene This is a tweak that I have been using for some time and gains greater importance for the core with the proposed addition of event delegation via `<$eventcatcher/>`

This PR also adds two new options for positioning the popup: `belowright` and `aboveleft`


Jermolene:
Thanks @saqimtiaz

PR #5088, opened 2020-11-21
Fixed bug with default value in lookup operator

### LINKED ISSUE #3445: Is the Lookup Operator "default" broken?
In an attempt to [Return a literal or new title from a filter?](https://groups.google.com/forum/?hl=en#!topic/tiddlywiki/in1d4QIXzHI) I just attempted to use the Loopup:default[nonexistantprefix] to get the filter to return "default". It did not work.

Looking at the [example](https://tiddlywiki.com/prerelease/#lookup%20Operator%20(Examples) both on tiddlywiki.com and tiddlywiki.com/prerelease, the second example is not working ie; is not (I believe) returning "show" for non existent visibility tiddlers.

Dependant upon the discussion  [Return a literal or new title from a filter?](https://groups.google.com/forum/?hl=en#!topic/tiddlywiki/in1d4QIXzHI) I am likely to raise another issue to improve the [title[tiddlername]] "filter".

Regards
Tony

saqimtiaz:
Apparently this bug has been around for a few years. The default value was never used properly.

Fixes #3445

### DIFF
--- a/core/modules/filters/lookup.js
+++ b/core/modules/filters/lookup.js
@@ -22,7 +22,7 @@ Export our filter function
 exports.lookup = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		results.push(options.wiki.getTiddlerText(operator.operand + title) || options.wiki.getTiddlerText(operator.operand + operator.suffix));
+		results.push(options.wiki.getTiddlerText(operator.operand + title) || operator.suffix);
 	});
 	return results;
 };
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -207,7 +207,7 @@ function runTests(wiki) {
 
 	it("should handle the lookup operator", function() {
 		expect(wiki.filterTiddlers("Six Seventh 8 +[lookup[Tiddler]]").join(",")).toBe("Missing inaction from TiddlerOne,,Tidd");
-		expect(wiki.filterTiddlers("Six Seventh 8 +[lookup:8[Tiddler]]").join(",")).toBe("Missing inaction from TiddlerOne,Tidd,Tidd");
+		expect(wiki.filterTiddlers("Six Seventh 8 +[lookup:8[Tiddler]]").join(",")).toBe("Missing inaction from TiddlerOne,8,Tidd");
 	});
 
 	it("should retrieve shadow tiddlers", function() {


saqimtiaz:
Investigating the failing test.

Jermolene:
Hi @saqimtiaz while this patch provides coherent behaviour, the current behaviour is intentional: the default value is used as-is, and not indirected via a lookup.

saqimtiaz:
> Hi @saqimtiaz while this patch provides coherent behaviour, the current behaviour is intentional: the default value is used as-is, and not indirected via a lookup.

Are we sure about that?

saqimtiaz:
Edited last comment. Will close if you are sure that is the intentional behaviour. Note this means the examples in documentation are not valid either.

Jermolene:
> > Hi @saqimtiaz while this patch provides coherent behaviour, the current behaviour is intentional: the default value is used as-is, and not indirected via a lookup.
> 
> Are we sure about that?
> 
> `results.push(options.wiki.getTiddlerText(operator.operand + title) || options.wiki.getTiddlerText(operator.operand + operator.suffix));`
> 
> Suggests that the default value is **NOT** used as is.

Apologies, I misread the diff!

So it's the documentation that's wrong, I think. I'd be inclined to correct the docs rather than the implementation.

Jermolene:
Judging by the original commit, it looks like I perhaps changed things after writing the docs: aa417fe899d0983ad8a3fe1fdfe9cac42364b2b6

saqimtiaz:
So basically the intended behaviour is:

If "operator.operand + title" does not exists a tiddler, return the text of "operator.operand + operator.suffix"?

Is that correct?
Are we using it in this manner in the core? It feels like returning the default value as is would be more useful, but if it is already in use then we cannot change it.

Jermolene:
> So basically the intended behaviour is:
> 
> If "operator.operand + title" does not exist as a tiddler, return the text of "operator.operand + operator.suffix"?
> 
> Is that correct?
> Are we using it in this manner in the core? It feels like returning the default value as is would be more useful, but if it is already in use then we cannot change it.

Hmmm I can't find any usages in the core apart from the example tiddlers. I do use it in a couple of my projects but it turns out that that all of those usages were aligned to the docs, not the implementation.

So, yes, perhaps it does make sense to change it.

saqimtiaz:
@Jermolene I looked and can't find the default in use anywhere either.

I will update the tests.

saqimtiaz:
@Jermolene I have updated the test. Code, tests and documentation are now consistent.

Jermolene:
Thanks @saqimtiaz and apologies for the confusion.

Issue #4958, opened 2020-11-01
[IDEA] Make keyboard shortcuts easier to discover


saqimtiaz:
In #4909 Jeremy suggested that: "There's a nice feature on the iPad where you can hold down a modifier key to see a HUD of all the key combinations that can be used. I wonder if we could do something similar: a global message handler that collects the active keyboard shortcuts and displays the HUD. We wouldn't be able to trigger it with a modifier key, we'd have to use F1 or ctrl-?. (It would be tricky to use a button to trigger it because the act of clicking the button will move the focus away)."

As we have added a fair few keyboard shortcuts to TW recently, a feature that makes it easier to discover and be reminded of keyboard shortcuts warrants consideration.

rmunn:
In addition to context-sensitive keyboard shortcuts, it might be nice if that HUD also had a link or button to open the Control Panel tiddler with the Keyboard Shortcuts tab already selected, so that the user can more easily discover all the TiddlyWiki keyboard shortcuts as well as the ones applying to the current context. (Is it possible to create a link to the Control Panel with a specific tab selected, or would it have to be a link to the Control Panel tiddler, and then the user would have to click on Keyboard Shortcuts by themselves?)

saqimtiaz:
@rmunn good idea and that should be doable. Selected tabs are controlled by state tiddlers so we can have a button that sets the appropriate state tiddlers and opens the Control Panel.

rmunn:
https://github.com/Jermolene/TiddlyWiki5/issues/4957#issuecomment-720980153 is also relevant to the discoverability of keyboard shortcuts.

MrDowntempo:
#5017 is another situation where a keyboard shortcut's discoverability (ESC) could help alleviate the issue.

PR #5089, opened 2020-11-22
Add Switcher modal

saqimtiaz:
This is an alternative take on #5003 

`tm-show-switcher` accepts a parameter `switch` which determines which switcher to show. Possible values are "theme", "palette", "language" and "layout".

The keyboard shortcut <kbd>Ctrl+Shift+L</kbd> is configured for the layout switcher.

A quick take on a CSS class for centered modals has been added and used. Our options here are somewhat limited by the JavaScript controlled animations and properties.

[Demo](https://github.com/Jermolene/TiddlyWiki5/files/5579395/switcher-demo.zip)

image

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -64,7 +64,7 @@ MissingTiddler/Hint: Missing tiddler "<$text text=<<currentTiddler>>/>" -- click
 No: No
 OfficialPluginLibrary: Official ~TiddlyWiki Plugin Library
 OfficialPluginLibrary/Hint: The official ~TiddlyWiki plugin library at tiddlywiki.com. Plugins, themes and language packs are maintained by the core team.
-PageTemplate/Description: the default tiddlywiki layout
+PageTemplate/Description: the default ~TiddlyWiki layout
 PageTemplate/Name: Default ~PageTemplate
 PluginReloadWarning: Please save {{$:/core/ui/Buttons/save-wiki}} and reload {{$:/core/ui/Buttons/refresh}} to allow changes to ~JavaScript plugins to take effect
 RecentChanges/DateFormat: DDth MMM YYYY
@@ -77,10 +77,10 @@ Shortcuts/Input/Tab-Left/Hint: Select the previous Tab
 Shortcuts/Input/Tab-Right/Hint: Select the next Tab
 Shortcuts/Input/Up/Hint: Select the previous item
 Shortcuts/SidebarLayout/Hint: Change the sidebar layout
-Switcher/Subtitle/theme: Switch themes
-Switcher/Subtitle/layout: Switch layouts
-Switcher/Subtitle/language: Switch languages
-Switcher/Subtitle/palette: Switch palettes
+Switcher/Subtitle/theme: Switch Theme
+Switcher/Subtitle/layout: Switch Layout
+Switcher/Subtitle/language: Switch Language
+Switcher/Subtitle/palette: Switch Palette
 SystemTiddler/Tooltip: This is a system tiddler
 SystemTiddlers/Include/Prompt: Include system tiddlers
 TagManager/Colour/Heading: Colour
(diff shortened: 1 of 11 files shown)


saqimtiaz:
@BurningTreeC This builds on your work. It was just easier to code it than explain the approach. 
Would love your input on the CSS, especially the transitions.

BurningTreeC:
@saqimtiaz , I understand your approach and I like it!

About the transition: Removing it does not change anything, I doubt that it has any effect

BurningTreeC:
I propose making the keyboard shortcut <kbd>ctrl-shift-L</kbd> because <kbd>alt-shift-L</kbd> is already taken for `linkify` in the text editor, which can cause confusion

saqimtiaz:
Keyboard shortcut changed to <kbd>Ctrl+shift+L</kbd> and CSS line for animation removed. Demo updated.

saqimtiaz:
@Jermolene I believe this meets your specs for how the switcher should be.

Should we decide not to have this in 5.1.23 for any reason, I suggest we at least add a simple switcher to the Control Panel.

core/language/en-GB/Misc.multids
```diff
@@ -63,6 +64,8 @@ MissingTiddler/Hint: Missing tiddler "<$text text=<<currentTiddler>>/>" -- click
 No: No
 OfficialPluginLibrary: Official ~TiddlyWiki Plugin Library
 OfficialPluginLibrary/Hint: The official ~TiddlyWiki plugin library at tiddlywiki.com. Plugins, themes and language packs are maintained by the core team.
+PageTemplate/Description: the default tiddlywiki layout
```
Jermolene, line 67: We should capitalise `~TiddlyWiki`

core/language/en-GB/Misc.multids
```diff
@@ -74,6 +77,10 @@ Shortcuts/Input/Tab-Left/Hint: Select the previous Tab
 Shortcuts/Input/Tab-Right/Hint: Select the next Tab
 Shortcuts/Input/Up/Hint: Select the previous item
 Shortcuts/SidebarLayout/Hint: Change the sidebar layout
+Switcher/Subtitle/theme: Switch themes
```
Jermolene, line 80: I think all of these should perhaps be singular "Switch theme", "Switch layout" etc. I'm not sure if we should capitalise the second word, but I'd be inclined to if it doesn't look odd

core/ui/Switcher.tid
```diff
@@ -0,0 +1,11 @@
+title: $:/core/ui/Switcher
```
Jermolene, line 1: Perhaps title this $:/core/ui/SwitcherModal

Jermolene (commented):
Thanks @saqimtiaz looks good, just a couple of minor wording issues

saqimtiaz:
All good suggestions @Jermolene and now resolved.

image

Jermolene:
Thanks @saqimtiaz

saqimtiaz:
@Jermolene any thoughts on where to document the shortcut <kbd>Ctrl+Shift+L</kbd> for the switcher?

Considering that it's intended to be a way to resolve a potentially broken layout we need to make sure it is an easy to discover shortcut.

Jermolene:
Hi @saqimtiaz in the first instance, I think we need some docs under "Customise TiddlyWiki" in the TOC talking about layout switching, both from the perspective of an end user and for people creating new layouts. But longer term, I'm really interested in the idea of a context sensitive "heads up display". (We can figure out the keyboard shortcuts prevailing for a DOM node by walking up the widget tree)

saqimtiaz:
@Jermolene HUD idea is great, we have a ticket to work on it after 5.1.23: #4958 

I'll work on some docs for the layout switching.

kookma:
@saqimtiaz 

Seems the $:/snippets/LayoutSwitcher needs a caption

image

saqimtiaz:
@kookma good catch, I was focused on the modal and not the ControlPanel. 

@BurningTreeC would you have time to add the missing caption? I'm a bit pressed for time this evening.

BurningTreeC:
yes @saqimtiaz , I'll add a caption. "Layout Switcher" should be good

saqimtiaz:
@BurningTreeC I think in this case we just need the caption "Layout". See the screenshot from @kookma above.

Thank you, much appreciated.

kookma:
Hi @Jermolene 
 Is it possible to have at least one other page layout in Tiddlywiki 5.1.23
 Do you think solo can be used here?

Jermolene:
Hi @kookma I don't think we have any candidate layouts of a sufficient quality at this point; part of the problem is that any layout that goes into the core has to be permanent. The idea of including the layout switcher in v5.1.23 is to give an opportunity for plugin writers to be able to get to work on new layouts.

PR #5090, opened 2020-11-22
Make type input field min-width the highest character length of selectable types

### LINKED ISSUE #2986: Width of content Type in Editor too narrow

BurningTreeC:
Fixes #2986

This PR counts the length of the longest selectable type and adds that length in characters as min-width to the type input field

It was mentioned somewhere (~~I cannot find the source~~ ... see #2986) that the type input field was too narrow. This PR fixes that problem

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1385,7 +1385,7 @@ html body.tc-body.tc-single-tiddler-window {
 	display: inline-block;
 }
 
-<<set-type-input-min-width>>
+<<set-type-selector-min-width>>
 
 .tc-edit-tags {
 	border: 1px solid <<colour tiddler-editor-border>>;
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -48,7 +48,7 @@ $else$
 \end
 
 \define set-type-selector-min-width()
-<$list filter="[all[shadows+tiddlers]removeprefix[$:/language/Docs/Types/]length[]] +[nsort[]] +[last[]]" variable="typeLength">
+<$set name="typeLength" value={{{ [all[shadows+tiddlers]prefix[$:/language/Docs/Types/]get[name]length[]maxall[]] }}}>
 
 	.tc-type-selector-dropdown-wrapper {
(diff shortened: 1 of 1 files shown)


themes/tiddlywiki/vanilla/base.tid
```diff
@@ -47,6 +47,20 @@ $else$
 </$reveal>
 \end
 
+\define set-type-selector-min-width()
+<$list filter="[all[shadows+tiddlers]removeprefix[$:/language/Docs/Types/]length[]] +[nsort[]] +[last[]]" variable="typeLength">
```
pmario, line 51: `[all[shadows+tiddlers]removeprefix[$:/language/Docs/Types/]length[]maxall[]]` does the same thing BUT, that's not the right size. the text is combined `name` + `description`  field if I remember right.

There probably is a much bigger performance problem, but I'm not sure without testing

BurningTreeC: @pmario , you're right, it's the `name` field that gets used

The performance impact cannot be that high (I hope) :smile:

BurningTreeC: Using the `maxall[]` operator on the length gives me `-Infinity` which is strange

BurningTreeC: Ah sorry! I found my error

BurningTreeC (commented):
I've updated the stylesheet to use `maxall[]` and the `name` field for calculating the length

pmario:
Have a look at: https://tiddlywiki.com/prerelease/#%24%3A%2Flanguage%2FDocs%2FTypes%2Fimage%2Fsvg%2Bxml .. It's `description` + "(" + `name` + ")"

BurningTreeC:
@pmario - yes, but to solve the issue #2986 it's sufficient to get the `name` because that will be shown in the input field. The issue is about the input field not being wide enough to show the whole **name**

pmario:
ahh. ok

Jermolene:
Thanks @BurningTreeC

PR #5091, opened 2020-11-22
Button and Droppable widgets: improve refresh handling for classes

saqimtiaz:
This PR improves the refresh handling for the `Button` and `Droppable` widgets so that if the `class` attribute changes, the DOM node classes are updated without calling `refreshSelf()`

This is a performance improvement and especially relevant when using multiple widgets, and allows using dynamic class names (via filtered attributes) to control appearance of contents via CSS.

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -27,33 +27,22 @@ ButtonWidget.prototype = new Widget();
 Render this widget into the DOM
 */
 ButtonWidget.prototype.render = function(parent,nextSibling) {
-	var self = this;
+	var self = this,
+		tag = "button",
+		domNode;
 	// Remember parent
 	this.parentDomNode = parent;
 	// Compute attributes and execute state
 	this.computeAttributes();
 	this.execute();
 	// Create element
-	var tag = "button";
 	if(this.buttonTag && $tw.config.htmlUnsafeElements.indexOf(this.buttonTag) === -1) {
 		tag = this.buttonTag;
 	}
-	var domNode = this.document.createElement(tag);
+	domNode = this.document.createElement(tag);
+	this.domNode = domNode;
 	// Assign classes
-	var classes = this["class"].split(" ") || [],
-		isPoppedUp = (this.popup || this.popupTitle) && this.isPoppedUp();
-	if(this.selectedClass) {
-		if((this.set || this.setTitle) && this.setTo && this.isSelected()) {
-			$tw.utils.pushTop(classes,this.selectedClass.split(" "));
-		}
-		if(isPoppedUp) {
-			$tw.utils.pushTop(classes,this.selectedClass.split(" "));
-		}
-	}
-	if(isPoppedUp) {
-		$tw.utils.pushTop(classes,"tc-popup-handle");
-	}
-	domNode.className = classes.join(" ");
+	this.assignDomNodeClasses();
 	// Assign other attributes
 	if(this.style) {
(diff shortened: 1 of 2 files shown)


core/modules/widgets/button.js
```diff
@@ -219,14 +207,33 @@ ButtonWidget.prototype.execute = function() {
...
+	if(changedAttributes.actions || changedAttributes.to || changedAttributes.message || changedAttributes.param || changedAttributes.set || changedAttributes.setTo || changedAttributes.popup || changedAttributes.hover || changedAttributes.selectedClass || changedAttributes.style || changedAttributes.dragFilter || changedAttributes.dragTiddler || (this.set && changedTiddlers[this.set]) || (this.popup && changedTiddlers[this.popup]) || (this.popupTitle && changedTiddlers[this.popupTitle]) || changedAttributes.setTitle || changedAttributes.setField || changedAttributes.setIndex || changedAttributes.popupTitle || changedAttributes.disabled) {
```
pmario, line 232: ` changedAttributes.selectedClass` will always be "undefined" since `this.computeAttributes()` uses the name `class`

pmario: A lot of other widgets have the same problem. eg: radio and chackbox. ... I'm creating PRs atm.

pmario: I also think constructions like: `(this.set && changedTiddlers[this.set])` should not be here and we should use better init values in the `.execute()` function, to avoid the `&&`

core/modules/widgets/button.js
```diff
@@ -200,7 +189,6 @@ ButtonWidget.prototype.execute = function() {
 	this.setTo = this.getAttribute("setTo");
 	this.popup = this.getAttribute("popup");
 	this.hover = this.getAttribute("hover");
-	this["class"] = this.getAttribute("class","");
```
pmario, line 203: I would like to keep this line, so it's clear that we didn't forget about the value. .. Later in your code you could use `this.class` instead of `this.getAttribute`

pmario (commented):
Just some thoughts

core/modules/widgets/button.js
saqimtiaz: > ` changedAttributes.selectedClass` will always be "undefined" since `this.computeAttributes()` uses the name `class`

You might be confusing `selectedClass` with `class`, they are two distinct widget attributes.

I've targeted these specific widgets because the cost of re-rendering them can be a lot higher as they can contain a lot of nested widgets (e.g when used with tag="div" for instance").

core/modules/widgets/button.js
saqimtiaz: > Later in your code you could use `this.class` instead of `this.getAttribute`

That will not work, as `this.class` will not be updated on refresh, which is why we need to use `this.getAttribute()` again in the refresh handler.

saqimtiaz: If really necessary I would favour adding a comment here rather than redundant code.

pmario: I'm OK with at comment.

core/modules/widgets/button.js
pmario: uups. I didn't expand the code, so I thought it's an issue.

saqimtiaz:
Resolved. This should be good to go.

I'm forcing myself to stop making new PRs so we can actually get 5.1.23 out the door soon!

Jermolene:
Thanks @saqimtiaz

PR #5097, opened 2020-11-23
Docs for eventcatcher widget

saqimtiaz:
I've tried to be as thorough as possible and have included an example which incorporates logging/debugging.

I think at least for those familiar with JavaScript events, this should be pretty straight forward to use.


Jermolene:
Thanks @saqimtiaz

PR #5099, opened 2020-11-23
Fix StaticRiver exporter width of static tiddlers

### LINKED ISSUE #5072: [IDEA] Exported HTML Tiddler Width Sidebar Options
If you export a tiddler and the side bar is open, the resulting html will display a blank space at right where the sidebar would be. 

If the tiddler is exported with the sidebar closed, the result will be a full width html tiddler.

Another option is to view in new window where the tiddler shows at full width.

I’m thinking it would be nice to either remove the sidebar width or to have an option to meep or remove it.

BurningTreeC:
Fixes #5072

This PR adds a class `tc-static-story-river` to the StaticRiver exporter, removes the right margin and adds a padding of 42px when we're above the sidebarbreakpoint

### DIFF
--- a/core/templates/exporters/StaticRiver.tid
+++ b/core/templates/exporters/StaticRiver.tid
@@ -26,7 +26,7 @@ extension: .html
 </head>
 <body class="tc-body">
 {{$:/StaticBanner||$:/core/templates/html-tiddler}}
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @BurningTreeC

PR #5100, opened 2020-11-23
Adding a MacOS - inspired dark color palette

BurningTreeC:
Bildschirmfoto_2020-11-23_19-09-13
Bildschirmfoto_2020-11-23_19-09-29


Jermolene:
Hank's @BurningTreeC looks good

Jermolene:
Hi @BurningTreeC apologies but on reflection I think we should avoid the word "macOS" and instead call it "Cupertino" (Apple's HQ location; there's a tradition to use it to name things inspired by Apple's designs (and "Redmond" for Microsoft)).

PR #5101, opened 2020-11-23
Update and rename MacOSDark.tid to CupertinoDark.tid

BurningTreeC:
This PR renames the MacOSDark palette to Cupertino dark and changes its description to `A Cupertino-inspired dark palette`

### DIFF
--- a/core/palettes/CupertinoDark.tid
+++ b/core/palettes/CupertinoDark.tid
@@ -1,7 +1,7 @@
 title: $:/palettes/CupertinoDark
 tags: $:/tags/palette
 name: Cupertino Dark
-description: A Cupertino-inspired dark palette
+description: A macOS inspired dark palette
 type: application/x-tiddler-dictionary
 
 alert-background: <<colour background>>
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC apologies, my fauly: we should keep macOS in the description, just not in the name.

BurningTreeC:
Changed @Jermolene

Jermolene:
Thanks @BurningTreeC

Jermolene:
Hi @BurningTreeC another small point: this palette recolours the "save changes" button to red, which undermines the red colouring we use to indicate unsaved changes. Perhaps it would be better to avoid recolouring buttons?

BurningTreeC:
Hi @Jermolene, as far as my testing goes, it's only colored red when there are unsaved changes (like the palette that has currently changed?)

BurningTreeC:
*recently

Jermolene:
> Hi @Jermolene, as far as my testing goes, it's only colored red when there are unsaved changes (like the palette that has currently changed?)

Doh, of course, thank you. Weirdly, on iPad the control panel icon is also coloured:

image

Parenthetically, I think the page toolbar button contrast is too low. Do the macOS guidelines specify minimum contrast between background and foreground? We should be aware of the impact on accessibility.

BurningTreeC:
>Parenthetically, I think the page toolbar button contrast is too low. Do the macOS guidelines specify minimum contrast between background and foreground? We should be aware of the impact on accessibility.

@Jermolene - noted. There will follow some updates for this palette

PR #5105, opened 2020-11-23
Updates for Cupertino Dark Palette

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

kookma:
Hi @BurningTreeC 
Thank you for this great palette. Below are few testing.

**Links**
I think in Tiddlywiki visited links have the same look as not vested ones

image

Cupertino

097_chrome

I may recommend to keep visited link in Cupertino green as non visited ones. Furthermore the green has better contrast versus purple on dark background

kookma:
**Test 2**

Input boxes look like 3D elements the bottom border color cause this 3D effect.

See search box in sidebar and menubar. I recommend a flat style and look


096_chrome

kookma:
**Test 3**

Editing image tiddler:

When you create a tiddler of type png (or other images) the canvas is white and does not follow the dark palette.


094_explorer

BurningTreeC:
@kookma I like the different coloring of visited links. I don't know what others think.

The 3D search box is Firefox specific (browser specific) and I don't know if we want to change it so that it looks consistently across browsers.

BurningTreeC:
About the canvas : I somehow expect a white canvas when I want to draw something

kookma:
Hi @BurningTreeC ,

> The 3D search box is Firefox specific (browser specific) and I don't know if we want to change it so that it looks consistently across browsers.

See the below images both in Chrome 86, Windows 10

100_chrome
099_chrome

I believe it is border shading and color makes a 3D effect like on inputbox

kookma:
> @kookma I like the different coloring of visited links. I don't know what others think.

I like so, but I have got used to vanilla and I thought it is better to follow the same rule.
What do you think if change the shades?

kookma:
> About the canvas : I somehow expect a white canvas when I want to draw something

I think the contrast is very high!

PR #5106, opened 2020-11-23
Lighter sidebar-controls-foreground for Cupertino Dark Palette

BurningTreeC:
This makes the sidebar and tiddler controls better visible


Jermolene:
Thanks @BurningTreeC

PR #5107, opened 2020-11-23
Update CupertinoDark.tid

BurningTreeC:
This makes the tiddler-controls color a bit darker than the sidebar-controls color since the tiddler-controls are on a darker background and their color is also used for the editor buttons


Jermolene:
Thanks @BurningTreeC

PR #5109, opened 2020-11-23
Docs for switching layouts

saqimtiaz:
This covers both how to switch layouts and what it would entail to create a new layout.

We can elaborate a bit more once we have a second example of a layout that ships with the core.


Jermolene:
Great stuff thanks @saqimtiaz

PR #5111, opened 2020-11-24
Update chinese translations

BramChen:


Jermolene:
Thanks @BramChen, and thank you for keeping up with all the v5.1.23 updates so assiduously, much appreciated!

Issue #4742, opened 2020-06-28
[BUG] Data loss on Node if one or more modified files isn't writable


hoelzro:
**Describe the bug**
See https://groups.google.com/forum/#!topic/tiddlywiki/Y0fLouDHCqU

If one or more files underneath the wiki's tiddlers directory isn't writable, and one of them is changed, TiddlyWiki will keep trying to write that file, and other files (which _are_ writable) won't be written as a result.

What's more, the server appears to have accepted the changes, so no error dialog shows up, and `[haschanged[]]` has no results, so the wiki appears non-dirty.  Also, since the changes are cached in Node's memory, reloading the wiki appears to be normal - it's only after you've reloaded the Node daemon that the problem truly becomes apparent.

**To Reproduce**

Run this script, or its like in the browser:

```shell
$ npm install tiddlywiki@latest
$ ./node_modules/.bin/tiddlywiki --version
5.1.22
$ ./node_modules/.bin/tiddlywiki wiki --init server
Copied edition 'server' to wiki
$ ./node_modules/.bin/tiddlywiki wiki --listen port=9091
$ cat tw.log
 syncer-server-filesystem: Dispatching 'save' task: $:/StoryList
Serving on http://127.0.0.1:9091
(press ctrl-C to exit)
$ curl -s -w "%{http_code}\n" -H 'X-Requested-With: TiddlyWiki' -X PUT -d '{"title":"test1","text":"abc123"}' http://localhost:9091/recipes/default/tiddlers/test1
204
$ cat wiki/tiddlers/test1.tid
created: 20200628215555511
modified: 20200628215555511
title: test1

abc123
$ curl -s -w "%{http_code}\n" -H 'X-Requested-With: TiddlyWiki' -X PUT -d '{"title":"test2","text":"def456"}' http://localhost:9091/recipes/default/tiddlers/test2
204
$ cat wiki/tiddlers/test2.tid
created: 20200628215557534
modified: 20200628215557534
title: test2

def456
$ chmod 400 wiki/tiddlers/test1.tid
$ curl -s -w "%{http_code}\n" -H 'X-Requested-With: TiddlyWiki' -X PUT -d '{"title":"test1","text":"abc123\n\nupdate"}' http://localhost:9091/recipes/default/tiddlers/test1
204
[...]

PR #5113, opened 2020-11-24
Fix filesystem adaptor

joshuafontany:
As mentioned in the other PR (https://github.com/Jermolene/TiddlyWiki5/pull/4630) I have completely refactored the updates to the `filesystemadaptor` and `filesystem` core files. I tried to "tread lightly" and hook everything back up to the correct callback-flow. Docs will come once testing says they are needed.

Primary changes:
- Draft tiddlers are always saved in "[wikifolder]/tiddlers/drafts" with their complex title "simplified" to remove pathing-slashes and non-filesystem-friendly characters.
  - This allows all drafts to be saved to the server with no worries about overwriting, or nested directory problems, or filesystem errors. (Draft titles have the currently logged in UserName appended to them, such as `Draft of 'tatiana-hordiienko-weylin-enright.jpg' by Joshua Fontany`.)
  - Draft tiddlers are always saved as *.json singlefiles if they have bad fields, but are otherwise saved as a *.tid or the tiddler `type` field's matching filetype+meta, such as *.txt+*.txt.meta for "text/plain". Drafts are NOT affected by the config tiddlers below (because no, just no).

- Allow filters in `$:/config/FileSystemPaths` to change a tiddler's location on disk (exception: tiddlers loaded via a `tiddlywiki.files`config with `isEditableFile` set to `true` have a static "basepath" location and base file-name set -- but see next bullet). First matching filter always applied (& it is recommended to have a final "catch all" filter - which is what conflicted with the isEditableFile flag).
- Allow filters in `$:/config/FileSystemExtensions` to change a tiddler file's extension and basic "file-type". Store jpgs as *.txt+*.txt.meta (or whatever). Tiddlers loaded via a `tiddlywiki.files`config with `isEditableFile` set to `true` will honor this custom extension.

- Setup the __server_side__ `$tw.syncer.fileInfo[title]` array to grab all the data from `$tw.boot.files[title]` at boot startup. 
  - This is so when a tiddler's location or extension changes, we can compare the two & delete any remainig "duplicate" files on disk once the "new" file has been saved.
  - This required rewiring a lot of callbacks to return the correct data. I think I got it right? Once a file save (& duplicate delete check) are done, the callback returns the content of `$tw.boot.files[title]` and updates the store in `$tw.sycer.fileInfo[title]` (cycle complete).

Renaming tiddlers is actually an internal "save new, then delete old" operation, but I have tested this a lot with other changes (to text and other field, tags, etc) to drive new FileSystemPaths or FileSystemExtension results (& thus trigger "moves" by writing then deleting, writing with new extension, etc).

I hope this is stable, so make a COPY of a wiki folder to test it on. Mahalo (thanks) for any feedback!!!

Best,
Joshua F


saqimtiaz:
@joshuafontany Unfortunately I'm a little under the weather since yesterday afternoon, but as soon as I'm a little better I will do some thorough testing.

 I have some thoughts on possibly simplifying things, but I'll wait to share them as they are uninformed until I get the chance to properly review the changes.

joshuafontany:
@saqimtiaz  No worries. I also just realized that the set of filters I am using for $:/config/FileSystemPaths testing may have overcomplicated a couple of things. I'll be able to check that tomorrow (& post more of my test setup). Mahalo (thanks)!

saqimtiaz:
> @saqimtiaz No worries. I also just realized that the set of filters I am using for $:/config/FileSystemPaths testing may have overcomplicated a couple of things. I'll be able to check that tomorrow (& post more of my test setup). Mahalo (thanks)!

@joshuafontany thank you, that will be helpful.

Jermolene:
Thanks @joshuafontany this looks very comprehensive. We need to assess whether we want to try to get it into v5.1.23 -- I'd be inclined to try. In any case, please could you mark it as draft until it's ready to merge.

core/modules/utils/filesystem.js
```diff
@@ -204,15 +204,19 @@ exports.deleteEmptyDirs = function(dirpath,callback) {
...
 	pathFilters: optional array of filters to be used to generate the base path
-	wiki: optional wiki for evaluating the pathFilters
+	extFilters: optional array of filters to be used to generate the base path
+	wiki: optional wiki for evaluating the pathFilters,
+	fileInfo: an existing fileInfo to check against
 */
 exports.generateTiddlerFileInfo = function(tiddler,options) {
-	var fileInfo = {};
+	var fileInfo = {}, metaExt;
+	if (!!options.fileInfo) fileInfo.basepath = options.fileInfo.basepath;
```
Jermolene, line 219: Very minor style issue, but we don't have whitespace between `if` and the opening parenthesis.

Jermolene: Also we always use braces around the body of an `if` statement, even if it is only one statement,

Jermolene (commented):
Thanks @joshuafontany just a couple of coding style points. I haven't reviewed the logic yet.

core/modules/utils/filesystem.js
joshuafontany: Thank you @Jermolene. I'm cleaning this up and fixing the 'isEditableFile' flow when generating fileInfo.filepath (as I mentioned to @saqimtiaz). Will clean these up as well. 👍

joshuafontany:
After some more testing, I think this is ready for review! One thing to note, is that if you use the 'isEditableFile' flag in a `tiddlywiki.files` config that points __outside__ of your wiki folder, you are giving permission to TW+node.js to recursively delete empty directories at that location (not a bug, but could be unexpected behavior).

@saqimtiaz Here is my testing setup. The wiki folder located at `/c/wikis/Empty` started out as an empty node.js wiki.
Note for FileSystemPath filters used in testing. The `tiddlywiki.files` configs I am using will tag tiddlers loaded that way as "externalnote" (if isEditableFile == true) or "external" (if not editable).

Current folder structure:
   .
   |-external (exteral, non-editable files loaded via tiddlywiki.files)
   |-externalnotes (external, Editable file loaded via tiddlywiki.files)
   |-tiddlers
   |---_system
   |-----config
   |---drafts
   |---wiki
   |-----external (empty aside from tiddlywiki.files)
   |-----externalnotes (empty aside from tiddlywiki.files)
   |-----json (where tiddlers with tag[.json] are saved as singlefile .json tiddlers)
   |-----New Tiddler (contains "New Tiddler/1", "New Tiddler/2" etc)

Config Files:
```
title: $:/config/FileSystemPaths
type: text/vnd.tiddlywiki

[is[system]removeprefix[$:/]addprefix[_system/]]
[!is[system]tag[.json]addprefix[wiki/json/]]
[!is[system]!tag[externalnote]addprefix[wiki/]]
```

```
description: Forces a meta file and custom file extention when running on node.js. Same format as $:/config/FileSystemPaths, the first matching line's output becomes the new file extension.
title: $:/config/FileSystemExtensions
type: text/vnd.tiddlywiki

[type[text/vnd.tiddlywiki]tag[.txt]then[.txt]]
[tag[.json]then[.json]]
[tag[.tid]then[.tid]]
```

joshuafontany:
Files are moved around correctly, saved with the correct extensions, etc over many tests last night. With the fix today, I was able to force `isEditableFile` files to be saved back to their "home" directory if they do not match any FileSystemPath filters (note my final filter has to avoid catching these, and I use a tag for this purpose). It will even re-create that folder and place the files back there if ALL files have been moved into the tiddlers folder via changes, and then those changes are reverted so the FileSystemPaths filters no longer generate a match. Reopening as ready for review.

core/modules/utils/filesystem.js
```diff
@@ -276,40 +333,50 @@ exports.generateTiddlerFilepath = function(title,options) {
...
-	if(!filepath) {
-		// ...then just use the character codes of the title
-		filepath = "";	
-		$tw.utils.each(title.split(""),function(char) {
-			if(filepath) {
-				filepath += "-";
+	// Check for an "isEditableFile" basepath if no pathFilter matches
+	isEditableFile = options.fileInfo ? !!options.fileInfo.basepath : false;
+	if(!filepath && isEditableFile) {
+		var parsed = path.parse(options.fileInfo.basepath);
```
saqimtiaz, line 339: I'll take a thorough look in the morning but two quick questions @joshuafontany 
- are we ignoring options.extension if no pathFilter matched?
- for all tiddlers where no pathFilter matched, could we not re-use options.fileInfo.filePath if it exists? This would mean that all tiddlers that do not match a filePath filter get saved back to their original location if there was one.

joshuafontany: - The file extension is determined via a $utils method call and then passed to generateTiddlerFileInfo. This allows them to be customized completely independently.

 - This depends on your intent for `isEditableFile`, especially in corner-cases where a tiddler starts as one of these tiddlers, gets changed and re-saved to disk, and then gets changed again multiple times until it fails to trigger any FileSystemPath tiddler. Should that one end up in `./tiddlers` or should it "remember" it was loaded via tiddlywiki.files as editable and end up back in the basepath? What if you "accidentally" delete a tiddler and then recreate it (via cloning a similar one)?
    - When a tiddler is deleted, I clear it from `$tw.syncer.tiddlerInfo[title]` but I do not clear it from $tw.boot.files[title] as I was getting some weirness in the callback chain when I tried where I thought it would go. But this also gives us the ability to restore that data if a tiddler with that name is created again while the server is running.
  - Some of my tests had image tiddlers hopping around disk, being saved as .json tiddlers (bad fieds), then as *.txt, etc. I also wanted to run the `isEditableFile` tiddlers through similar series of tests.

Thanks for your notes!

saqimtiaz:
> I was able to force `isEditableFile` files to be saved back to their "home" directory if they do not match any FileSystemPath filters (note my final filter has to avoid catching these, and I use a tag for this purpose).

@joshuafontany If I understand correctly, this means it is a requirement that the final filter needs to not catch the externally loaded files. I would rather try to simplify our code by making it a requirement that files loaded via `tiddlywiki.files` with `isEditableFile` set to true, also need a corresponding `pathFilter` set to save them back to the original location. I think this can be documented properly and would not be a problem. 

Would that help in simplifying our handling?

joshuafontany:
@saqimtiaz That would simplify the handling greatly. As these are separate optional (& advanced) features we should aim for simplicity. I agree, and I'll make those edits. Then we can treat `isEditableFile` tiddlers just like every other tiddler.

saqimtiaz:
> @saqimtiaz That would simplify the handling greatly. As these are separate optional (& advanced) features we should aim for simplicity and document very well how they overlap. I agree, and I'll make those edits. Then we can treat `isEditableFile` tiddlers just like every other tiddler.

Excellent. Thank you @joshuafontany. While the description of the current implementation sounded like it would have fit my needs - and I greatly appreciate that you tried to accommodate them - the complexity and exceptional handling required in the code was not sitting well with me.

I'm still a bit under the weather but let me know when you've had a chance to revise and I'll do my best to get some testing done promptly.

joshuafontany:
@saqimtiaz Updated. I couldn't figure out where in the callback flow to delete an object out of `$tw.boot.files[]` in the sync adaptor, so I placed the below in `syncer` next to where it deletes the title from `$tw.syncer.tiddlerInfo[]`. This should mean that after each save/delete cycle, both arrays should agree, and only list active tiddlers currently saved to disk.

```
		if ($tw.boot.files){
			// Remove the tiddler from $tw.boot.files
			delete $tw.boot.files[self.title];
		}
```

Ready for testing!

core/modules/syncer.js
```diff
@@ -633,6 +633,10 @@ DeleteTiddlerTask.prototype.run = function(callback) {
 		}
 		// Remove the info stored about this tiddler
 		delete self.syncer.tiddlerInfo[self.title];
+		if($tw.boot.files){
+			// Remove the tiddler from $tw.boot.files
+			delete self.boot.files[self.title];
```
saqimtiaz, line 638: I believe this should be `$tw.boot.files` ? otherwise it just generates a slew or fatal errors.

saqimtiaz:
@joshuafontany with the above change in place, I've done some quick testing. Specifying the extension via `$:/config/FileSystemExtensions` seems to work well for both regular tiddlers and those added via `tiddlywiki.files`

Getting tiddlers loaded via `tiddlywiki.files` with `isEditableFile` saved back to their original path works with `$:/config/FileSystemPaths`, and you can also override and have those move files move around according to the filepath filters just like other tiddlers. Good. 

I am unfortunately still unwell so I have not tested other scenarios, nor have I been up to looking at the code logic and flow. Overall I think this is a more logical approach rather than accommodating special cases.

I am concerned however that this needs more thorough testing and there might be edge cases we are missing that may prove problematic. I think it would be prudent for a few of us to run these changes on a daily basis for a week or two before including in an official TW release.

saqimtiaz:
It would be good to test the following:

- saving a new tiddler with extension .json
- adding to an existing tiddler a multiline field, should save as .json
- adding to an existing tiddler a multiline field, where the same tiddler is also configured to save as  .json via `$:/config/FileSystemExtensions`
  - I recall getting some odd behaviours with this on the previous implementation

I am also curious as to why the change in the handling for **draft** tiddlers? 

In particular this might clash with future changes we probably want to implement for storage of per user tiddlers (like the storylist), and tie us down with backwards compatibility concerns when making those changes.

pmario:
>In particular this might clash with future changes we probably want to implement for storage of per user tiddlers (like the storylist), and tie us down with backwards compatibility concerns when making those changes.

+1

core/modules/syncer.js
joshuafontany: Good catch, thanks.

joshuafontany:
> > I am also curious as to why the change in the handling for **draft** tiddlers?
> > In particular this might clash with future changes we probably want to implement for storage of per user tiddlers (like the storylist), and tie us down with backwards compatibility concerns when making those changes.
> 
> +1

@saqimtiaz @pmario I agree that Draft handling could be better handled, especially considering backwards-compatibility and future per-user storage ideas. The two issues I ran into so far prompted me to just separate them from the normal `generateTiddlerFileInfo` flow were due to their "nested" titles.

1) The 5.1.22 implementation when not using FileSystemPath filters (or when none match) is to scrub filepath separator characters out (replacing them with "_"). As soon as you add a FileSystemPath filter that matches drafts, such as "[has[draft.of]addprefix[drafts/]]" that check gets skipped, and drafts of __system__ tiddlers (for example "$:/config/FileSystemPaths") are saved like this into /tiddlers:
.
|--drafts/
|----Draft of '$_
|------config
|--------FileSystemPaths'by Joshua Fontany.tid

2) I experimented with catching this in the FileSystemPAth filters by using the new filter operators that @saqimtiaz has been working on, especially search-replace, but the current flow to use the FileSystemPath filters does not allow pragmas (/define XYZ()) before the filters, and I wasn't getting anywhere or hadn't read the docs fully for the new operators yet....

Now imagine a draft of a system shadow tiddler from a plugin that has 4 or 6 path-separator characters in the title.... yeah. So, if we can find a way to parse the draft titles with the filters into a more reasonable on-disk location, we can skip the !has[draft.of] check there.

saqimtiaz:
@joshuafontany So if I understand correctly, the problem with drafts is that it is difficult to customize the file path without excessive nesting.

Would something like this work:

`[is[draft]search-replace:g:regexp[/|\\],[_]]`

joshuafontany:
Excellent, I had a feeling I was messing up that receipt. Should the example in the docs include the "addorefix[drafts/]" part if the run. Also, "diagramming" all of the example filters would be a great use of the filter-syntax table macro you and @Jermolene had been discussing.

I'll strip out the !draft check & also run through those mukti-line field test you mentioned this afternoon.

saqimtiaz:
@joshuafontany the more documentation examples the better.

I would love to get the filter visualization working where its not only a macro easily invoked for examples in documentation, but also setup to work with the filters tab in advanced search for testing filters.

joshuafontany:
@saqimtiaz Testing updates:

 - Tiddlers with bad fields will never have a custom or default file-type set, and are always set to ".json" with NO meta file. I saw the root-cause of your previous bug and moved this around slightly to be sure that never happens.

 - I think I understand that the `tiddlywiki.files` tiddlers that you are loading, and wish to write back to the "same directory" are all inside that wiki's "tiddlers/" folder. All my tests around that scenario are working fine.

 - I ran into a major bug when I loaded text files from outside that wiki's "tiddlers/" directory and then tries to put them back by using addprefix[] to add a relative path ("../externalnotes/"). This due to some of the checks handling the filepath, generated a "_." directory inside the wiki's "tiddlers/" folder - which immediately caused Windows10 to complain any time I tried anything with that folder.
    - I had to resort to a bash command line "rm -f -v -d -r /c/wikis/Empty/tiddlers/_." to clean up my file system.
    - This confirmed that there was the correct "externalnotes/" folder & tiddler files (inaccessable from windows) within that directory.

So, should we support relative paths in custom FileSystemPaths? Or should we warn people to keep all `isEditableFile` files loaded via `tiddlywiki.files` within the wiki's "tiddlers/" folder to start with? Or third option that will need research?

I'll sketch out some solutions based on relative paths.

@Jermolene @pmario Any input on that issue? Thanks everybody. :)

saqimtiaz:
> So, should we support relative paths in custom FileSystemPaths? Or should we warn people to keep all `isEditableFile` files loaded via `tiddlywiki.files` within the wiki's "tiddlers/" folder to start with? 

@joshuafontany 
I considered these issues when testing and I think it's perfectly ok to limit isEditableFile to files loaded from within the wiki's tiddlers folder. It also meshes well with the idea that the wiki only has write access to the tiddlers folder.

saqimtiaz:
I expect relative paths outside the tiddler directory for any kind of tiddler will be and probably also currently is problematic. So we definitely need to document that.

joshuafontany:
All file paths currently generated are relative to the "tiddlers/" directory, and that is how the config files work as well. We cannot actually use an absolute-path as a result in $:/config/FileSystemPaths.... So I'm inclined to not have additional special handling around the directory paths in the `tiddlywiki.files` configs... If I can get the normal relative syntaxes to not give any errors, we should be good.

And yes, note-to-self to document this very well. :)

joshuafontany:
PR updated with relative filesystem path fix. I also added tilde `~` to the next check for "characters that don't play well with file systems" to replace with "_"/ Tilde has OS specific meanings, and they are a rabbit hole. :P

core/modules/utils/filesystem.js
```diff
@@ -238,26 +240,75 @@ exports.generateTiddlerFileInfo = function(tiddler,options) {
...
+			}
+		}
 	}
-	// Take the file extension from the tiddler content type
+	// Take the file extension from the tiddler content type or metaExt
 	var contentTypeInfo = $tw.config.contentTypeInfo[fileInfo.type] || {extension: ""};
 	// Generate the filepath
 	fileInfo.filepath = $tw.utils.generateTiddlerFilepath(tiddler.fields.title,{
-		extension: contentTypeInfo.extension,
+		draft: !!tiddler.fields["draft.of"],
```
saqimtiaz, line 270: Do we still need this?

joshuafontany: Nope. Good catch, thanks.

joshuafontany:
This is working well with all of the various tiddler-tyle and file-type tests so far. In fact, I can only reliably trigger 1 bug so far, which is pathing a tiddler to be saved in a location that is write-protected (in this instance the root C drive).

title:  ../../../../../../this_is_an_orphan

Results in the debug log output below looping during every sync, as that Save task can never finish correctly and always gets re-queued.:

syncer-server-filesystem: Dispatching 'save' task: ../../../../../../this_is_an_orphan 
<node_internals>/$:/core/modules/utils/logger.js:51
Sync error while processing save of '../../../../../../this_is_an_orphan': Error: EPERM: operation not permitted, open 'C:\this_is_an_orphan.tid'

This does leave the previous saved file at the last location, but any changes to that tiddler once this bug is triggered would be lost once the server is shutdown. Hmmm...

saqimtiaz:
> This is working well with all of the various tiddler-type and file-type tests so far. In fact, I can only reliably trigger 1 bug so far, which is pathing a tiddler to be saved in a location that is write-protected (in this instance the root C drive).

This reminds me of #4742 and probably needs a separate fix that addresses the issue of insufficient permissions when writing to disk.

joshuafontany:
Aaah. As-is, this is currently a huge Security Risk. Allowing the user to define arbitrary write-location to disk is not goint to work out well. I think as long as the filepath starts with the wiki-path passed to tiddlywiki.js on startup, it should be fine (meaning that you could access the tiddlywiki.info file, the "plugins/" folders, etc from the wiki as its running, but no write access to anything above it or elswewhere on disk).

I'm also trying to figure out a good flow to catch EPERM and EACCESS errors & retry the save.

joshuafontany:
@saqimtiaz @Jermolene 

I believe I have resolved the last line-items for this PR.

 - Limit write access to the current wiki's directory (where tiddlywiki.info resides). IF a tiddler is attempts to be written outside of this, simply `encodeURICompnent(filepath)` and write that file to the wiki's `/tiddlers` directory. Security hole patched, and it retains a lot of flexibility.

 - If a Save or Delete task fails on the server (saves can fail writing, or fail deleting the previous file if there is a new filepath, whereas deletes can fail if the file is write protected or other weirdness) check for error codes EPERM and EACCESS (permissions or access error).
    - In the case of a failed save, repeat the above failure catch by setting a flag that will also trigger  `encodeURICompnent(filepath)` and write that file to the wiki's `/tiddlers` directory the next time the Save task is run.
    - In the case of a failed delete, fail gracefully but warn the server log that the server is desychronized (may not start as expected next time). Fail gracefully by not passing the error up the chain of callbacks, thus not blocking or infinitely retrying the Delete task.

Jermolene:
Great, thank you @joshuafontany @saqimtiaz. If you can please flip this out of draft mode I'll merge it.

joshuafontany:
@Jermolene Awesome. I would want to give it a quick read thru today & 1 or 2 more tests. It was late when I finally got the error catching done, may have missed something. If everything there goes fine I'll move this out of draft later today. 👍

Jermolene:
Great @joshuafontany thank you.

joshuafontany:
Thought so. Caught some final bugs around the writeError flag and other weird corner cases. I guess 5 years in video game QA in the Engineer room paid off. ;)

@Jermolene  Draft is now ready. 👍

Jermolene:
Thanks @joshuafontany and congratulations, this is a tricky area to work on to say the least!

saqimtiaz:
@joshuafontany Thank you for your efforts on this. We should make sure we get some documentation in place before the release as well.

Jermolene:
Hi @joshuafontany there appears to be a regression here. I just used the `bin/serve.sh` script to edit the `tw5.com` wiki to make some documentation updates (specifically editing "Date Format" and "DateFields").

I found that the editing those tiddlers changed their path to:

```
TiddlyWiki5/editions/tw5.com/tiddlers/%2FUsers%2Fjeremyruston%2Fgit%2FJermolene%2FTiddlyWiki5%2Feditions%2Ftw5.com%2Ftiddlers%2FDate%20Fields.tid
TiddlyWiki5/editions/tw5.com/tiddlers/%2FUsers%2Fjeremyruston%2Fgit%2FJermolene%2FTiddlyWiki5%2Feditions%2Ftw5.com%2Ftiddlers%2FDateFormat.tid
```

The `tw5.com` edition does not define `$:/config/FileSystemPaths`.

saqimtiaz:
@joshuafontany we may have missed the fact that to maintain backwards compatibility, in the absence of file path filters, tiddlers need to save back to their original path.

In fact it may even be that tiddlers that don't match any defined file path filters - even if some are configured - also need to save back to the path they are loaded from, so that it's not a breaking change for users relying on that behaviour. I believe that is the current behaviour in 5.1.22

joshuafontany:
@Jermolene Oof, good catch. I believe I overlooked the "includeWikis" functionality that allows you to set a `default-tiddler-location` in your tiddlywiki.info file. In the case of the "./editions/tw5.com-server" edition, it has 
```
	"includeWikis": [
		"../tw5.com"
	],
	"config": {
		"default-tiddler-location": "../tw5.com/tiddlers"
	}
```
Let me look at the logic that loads tiddlers from those options, and I should be able to get everything squared away.

Jermolene:
Excellent, thanks @joshuafontany.

joshuafontany:
@Jermolene Took longer than I expected, but with the result of a big improvement to overall handling.

Patch is up @ https://github.com/Jermolene/TiddlyWiki5/pull/5176

PR #5114, opened 2020-11-24
Update modifier variable docs

saqimtiaz:
Listing the potential key combinations in a single line is hard to read and also redundant when we have the same information in the table below.


BurningTreeC:
Thanks @saqimtiaz

Jermolene:
Thanks @saqimtiaz @BurningTreeC

PR #5115, opened 2020-11-24
Refactor updating of classes for Button widgets

saqimtiaz:
@Jermolene the updating of the `Button` widget classes that I introduced in #5091 is suboptimal, and I realized it could cause problems in edge cases.

This simplifies things: if the `class` attribute changes, we remove any old classes on the domNode that were derived from the `class` attribute, and add the new ones instead.

We do not need to check `isPoppedUp()`/`isSelected()`. If other attributes have changed it will trigger a `refreshSelf()` anyway. This handling is for when only the `class` attribute has changed.


Jermolene:
Thanks @saqimtiaz

PR #5116, opened 2020-11-24
Delete PETTIL example due to bad link

mcomella:
The domain of the linked site, http://chitselb.com/files/tiddlypettil.html, has expired and now links to https://www.hugedomains.com/domain_profile.cfm?d=chitselb&e=com. Note that the linked source repository is still available https://github.com/chitselb/pettil

I'm not sure whether it's better to delete the whole example because the wiki is not available without rehosting or to leave it there for reference but I chose the former.

afaict, deleting examples from this page will remove them from the listing because it's aggregated via tags but I'm not sure and I have not tested this change because I don't know how and I'm short on time.


Jermolene:
Hi @mcomella thank you for spotting this, much appreciated. We do have a bureaucratic process of signing the CLA (see CONTRIBUTING.md in the root of this repo), which seems overkill for such a simple change, but I hope you'll be able to contribute more in the future.

As an aside, I have been thinking that it would be useful to make a link checker plugin for Node.js but this reminds me that we would need to not just check for 404s, but also track redirects, and perhaps have a block list of domains that should count as 404s.

jeremyredhead:
Even **better**, why not just swap out the link with a wayback snapshot?:  
https://web.archive.org/web/20171114121112/chitselb.com/files/tiddlypettil.html  
It looks to be mostly intact.

Alternately, since as you mentioned the project in question is still going, 
we could link to the current/new PETTIL TiddlyWiki: http://pettilmobile.com/tw/  
(The github README links to http://pettilmobile.com/files/tiddlypettil.html but that's a 404)

PR #5117, opened 2020-11-24
Add TiddlyDesktop colors to CupertinoDark.tid + small Adjustments

BurningTreeC:


Jermolene:
Thanks @BurningTreeC I haven't tried it in TiddlyDesktop, it'd be cool to expose this stuff to the user there.

PR #5118, opened 2020-11-24
Fix #5108 - vanilla/reset overrides system fonts

### LINKED ISSUE #5108: [BUG] - vanilla/reset overrides fonts from vanilla/base Stylesheet
**Describe the bug**
Changing the fonts in ControlPanel -> Appearance (current prerelease) doesn't have any effect because those fonts are overridden by the ones set in vanilla/reset on the body

**Solution**
We should remove the block that defines those fonts explicitely from the vanilla/reset Stylesheet

BurningTreeC:
This PR fixes #5108 where we've found out that vanilla/reset overrides the fonts in the vanilla/base stylesheet

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -90,7 +90,7 @@ html button {
 ** Basic element styles
 */
 
-html {
+html, body {
 	font-family: {{$:/themes/tiddlywiki/vanilla/settings/fontfamily}};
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @BurningTreeC

saqimtiaz:
@BurningTreeC we have a bug report that this might not have been fully resolved:

https://groups.google.com/g/tiddlywiki/c/dZmp1gQivJI/m/OpWflmgVAwAJ

Issue #5104, opened 2020-11-23
[BUG] Some characters in tiddler titles in pre-release are obscured.


saqimtiaz:
**The bottom of some characters in tiddler titles is obscured.**

![image](https://user-images.githubusercontent.com/68092/100004698-286eef00-2dc8-11eb-9e33-45010dee7264.png)

Note the bottom of the character `g` in Plugins.
From: https://tiddlywiki.com/prerelease/#Plugins

![image](https://user-images.githubusercontent.com/68092/100004909-78e64c80-2dc8-11eb-9cb3-55a8ad629f03.png)
Or the `j` and `y` characters at https://tiddlywiki.com/prerelease/#TiddlyWiki%20on%20Node.js

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome. Firefox has the same issue with the j character but only intermittently.

saqimtiaz:
@BurningTreeC do you know if we have made any CSS changes in this area?

BurningTreeC:
@saqimtiaz not that I know ... this is new to me

saqimtiaz:
Everything is fine on tiddlywiki.com

BurningTreeC:
@saqimtiaz I'll investigate - but I can't remember having changed anything in this direction. Maybe it's the new normalize.css we're using

saqimtiaz:
@BurningTreeC thanks! It occurred to me that for the layout switcher I only focused on the modal and just copied your switcher code verbatim without testing it. Going to give it a quick test now to confirm it switches layouts as expected.

BurningTreeC:
>It occurred to me that for the layout switcher I only focused on the modal and just copied your switcher code verbatim without testing it. Going to give it a quick test now to confirm it switches layouts as expected.

fingers crossed that it works @saqimtiaz

saqimtiaz:
Switching layouts doesn't work! But I think the problem is not with the switcher but the underlying code. Sigh.

Will investigate after dinner. Sorry for going offtopic here.

BurningTreeC:
@saqimtiaz - to this issue here: it may have something to do with the new font settings. They prefer the system-ui font now

saqimtiaz:
@BurningTreeC manually changing the font family to the same as 5.1.22 didn't seem to help. Can you confirm that?

BurningTreeC:
@saqimtiaz - I don't seem to have this issue

Linux, Google Chrome

[...]

PR #5122, opened 2020-11-25
fix titlebar line-height for chrome

pmario:
Increasing the line-height from initial PR wasn't big enough. 
This fixes: #5104 

@saqimtiaz can you check it with Chrome _and_ different browser ZOOM settings. I did test it with FF latest and Edge latest windows 10. I could recreate the problem with Edge.


saqimtiaz:
@pmario confirmed, problem is resolved with this change on all zoom levels on Chrome on Windows 10.

pmario:
The funny thing was. FF also had the issue using 120% and 140% (may be a rounding problem) .. But I'm using 130% or 150% for my screen and resolution, which where fine ;)

Jermolene:
Thanks @pmario @saqimtiaz

PR #5123, opened 2020-11-25
CodeMirror and Highlight.js: Add ability to define colors through Palette

BurningTreeC:
This PR adds the `tiddlywiki` codemirror theme which defines the colors by the `<<colour>>` macro. Like that we can define the codemirror colors through our palettes

It also makes the highlight.js plugin use the colors from the different palettes


Jermolene:
Hi @BurningTreeC I like the idea of aligning the highlight.js and CodeMirror palettes with the currently selected core palette.

However, this implementation suffers from having to bloat all the core palettes with largely identical settings, increasing the friction significantly for anyone making a new palette, and making the palette editor even more unwieldy. Can we instead map the required plugin specific entries to existing core palette entries?

I also have some concerns about the auto dark/mode work in #5085 which I'll comment there shortly. I wonder if we might need to rethink our approach to palettes/themes a little.

BurningTreeC:
Hi @Jermolene, I share your concerns and would also favor rethinking the palette-/theme approach. But we should postpone that to after the 5.1.23 release

Jermolene:
@BurningTreeC it might be worth making a ticket for this so we don't forget?

BurningTreeC:
@Jermolene, I've prepared a different PR adressing this, see #5128

PR #5126, opened 2020-11-27
Git saver : bypass cache to avoid 409 errors.

### LINKED ISSUE #5119: [BUG] GitHub saver often fails with a 409 Conflict
**Describe the bug**
I've set up the GitHub saver with auto-saving set to disabled. Saving succeeds sometimes, but quite often I get the "409 conflict" error message.

I'm using Safari on macOS Catalina.

**To Reproduce**
1. Setup the GitHub saver
2. Create a new tiddler
3. Save it
4. Create another one
5. Save it

**Expected behavior**
Saving is always successful

**Screenshots**
Here's the log in the network tab of the browser:

image

From here it looks like the failures always happen after the `contents` response is served from disk. Looking into that `contents` response, it comes back with a Cache-Control header of `private, max-age=60, s-maxage=60` so my guess is that when a save is done within 60s, `contents` is loaded from disk (and therefore has the same SHA) which then causes the subsequent PUT request to fail.

 - OS: macOS Catalina (latest update)
 - Browser: Safari (latest update for Catalina)

saqimtiaz:
Ensure that we always fetch the latest sha bypassing cache.

Fixes #5119 which is caused by the caching of the GET request if a subsequent save is within 60 seconds.

### DIFF
--- a/core/modules/savers/github.js
+++ b/core/modules/savers/github.js
@@ -31,7 +31,8 @@ GitHubSaver.prototype.save = function(text,method,callback) {
 		headers = {
 			"Accept": "application/vnd.github.v3+json",
 			"Content-Type": "application/json;charset=UTF-8",
-			"Authorization": "Basic " + window.btoa(username + ":" + password)
+			"Authorization": "Basic " + window.btoa(username + ":" + password),
+			"If-None-Match": ""
 		};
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz, improving the GitHub saver is double plus good.

PR #5127, opened 2020-11-28
Correctly add EventListener 'click' for popup-handling in new windows

BurningTreeC:
This PR changes how the eventListener for the click event in new windows that deletes popups is added. Before it was faulty


Jermolene:
Thanks @BurningTreeC

PR #5129, opened 2020-11-28
Make "tc-dirty" update also on external windows

BurningTreeC:
This PR makes externally opened windows (tiddler -> open external) update their dirty-status


Jermolene:
Thanks @BurningTreeC

PR #5131, opened 2020-11-28
Remove background-colors from framed engine

BurningTreeC:
This PR removes the extracted background-colors from the framed engine (reverts a PR from myself)


Jermolene:
Thanks @BurningTreeC

PR #5132, opened 2020-11-28
Make tiddler-editor iframe same color as tiddler background

BurningTreeC:
This PR needs #5131 and makes the tiddler-editor iframe appear in the same color as the tiddler. The tiddler-editor textarea still has the tiddler-editor-background color. Like this the 14px offset of the iframe become unvisible


Jermolene:
Thanks @BurningTreeC

PR #5133, opened 2020-11-28
Add TiddlySpot-ReadOnly Message-Box to Tiddlyspot Saving Panel

### LINKED ISSUE #5120: [BUG] Remove TiddlySpot tab from Control Panel?
With TiddlySpot in a read-only state and unlikely to return without major changes if at all, should we remove the saver tab for it from the Control Panel?

BurningTreeC:
This PR fixes #5120


Jermolene:
Thanks @BurningTreeC that's great.

PR #5134, opened 2020-11-28
Small update for better readability for CupertinoDark palette

BurningTreeC:
The muted dropdown text isn't readable. This makes the dropdown background darker


Jermolene:
Thanks @BurningTreeC

PR #5141, opened 2020-11-29
Remove outline border from buttons and remove ::-ms-clear cross from inputs

BurningTreeC:
This PR removes the outline borders from buttons which makes the page feel more like an application than a website (opinionated) and removes the clear-input-cross from inputs in internet explorer using `::-ms-clear { display: none; }`

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -84,6 +84,7 @@ html button {
 	color: <<colour button-foreground>>;
 	background: <<colour button-background>>;
 	border-color: <<colour button-border>>;
+	outline: none;
 }
 
 /*
@@ -228,6 +229,10 @@ input[type="search"]::-webkit-search-cancel-button,
 input[type="search"]::-webkit-search-results-button,
 input[type="search"]::-webkit-search-results-decoration {
 	-webkit-appearance:none;
+}
+
+::-ms-clear {
+	display: none;
 }
 
 .tc-muted {


pmario:
Before / After Screenshot?

BurningTreeC:
@pmario - here's an after-before GIF. I currently don't have an Internet Explorer by my hand to demonstrate the input-cross but I've tested `::-ms-clear { display: none }` well enough

Peek 2020-11-29 09-49

pmario:
Sorry I can't see anything.It changes way too fast. 2 still shots would allow me to see something

BurningTreeC:
@pmario see the outline on the add-tiddler-button (mouse is clicking on it):

BEFORE:

Bildschirmfoto_2020-11-29_10-55-11

AFTER:

Bildschirmfoto_2020-11-29_10-54-56

saqimtiaz:
@BurningTreeC +1 from me. I have always found those outlines annoying.

Jermolene:
Hi @BurningTreeC @saqimtiaz I share your frustration but this problem actually seems to just be a quirk with Chrome; there are no outlines in Safari or Firefox on macOS. The fix here will affect all browsers, and a particular concern is that messing with the outline style is one of the things that causes browsers to switch from using standard OS platform controls to their own non-standard implementations.

BurningTreeC:
>I share your frustration but this problem actually seems to just be a quirk with Chrome; there are no outlines in Safari or Firefox on macOS. The fix here will affect all browsers, and a particular concern is that messing with the outline style is one of the things that causes browsers to switch from using standard OS platform controls to their own non-standard implementations.

Ok @Jermolene - I can just use my own stylesheet to disable those outlines

Jermolene:
> Ok @Jermolene - I can just use my own stylesheet to disable those outlines

I'm not against trying to fix it, I use Chrome as my main development browser and it is very annoying. But perhaps we need to do some more digging.

saqimtiaz:
Interesting. I hadn't realized that Firefox does not show these outlines even on Windows.

@BurningTreeC I wonder if there is a webkit/chrome specific solution.

BurningTreeC:
@saqimtiaz no I couldn't find any webkit/chrome specific solution

BurningTreeC:
A possible solution would be to make the buttons in question <a> elements using `tag="a"` but that would involve a bigger overhaul of the vanilla/base stylesheet

Jermolene:
> A possible solution would be to make the buttons in question <a> elements using `tag="a"` but that would involve a bigger overhaul of the vanilla/base stylesheet

It would be a shame because the core tries to do things semantically, so buttons are for actions and links are for navigating; they aren't really interchangeable.

BurningTreeC:
>It would be a shame because the core tries to do things semantically, so buttons are for actions and links are for navigating; they aren't really interchangeable.

Understood. I'll not even try :smile:

pmario:
>Interesting. I hadn't realized that Firefox does not show these outlines even on Windows.

OK .. That's why I don't know, what it does. No Chrome installed on my system. I use Edge for "chrome testing" atm

kookma:
@BurningTreeC 
I can confirm this annoying outlines appear in Chrome/Edge/Opera on Windows 10.
But to my surprise you don't see them in Firefox.

I use `.tc-btn-invisible { outline:none; }` as a default css in my TWs.

Jermolene:
> > Interesting. I hadn't realized that Firefox does not show these outlines even on Windows.
> 
> OK .. That's why I don't know, what it does. No Chrome installed on my system. I use Edge for "chrome testing" atm

Are you seeing the outline issue in Edge? It occurs on Edge on macOS.

kookma:
> Are you seeing the outline issue in Edge? It occurs on Edge on macOS.

This Edge 87 (the latest on Windows 10)
image

jeremyredhead:
Please ***don't*** remove outlines from buttons just for aesthetics, those outlines are important for **accessibility**.
Without those outlines, someone browsing via keyboard can't tell where keyboard focus is on the page.
They *usually* don't show up until/unless you use `Tab` to navigate by keyboard, but maybe that's just a firefox thing.

### Instead...
- CSS could be refactored so buttons use same `:focus` styling as on `:hover`
- Different buttons "types"/classes could be given new subtler but "prettier" `:focus` styles
- Maybe [`:focus-visible`][mdn] could be used somehow. Though [browser support is still patchy][caniuse] unfortunately :/

Or most likely some mix of these methods. 

[mdn]: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
[caniuse]: https://caniuse.com/css-focus-visible

Jermolene:
Thanks @jeremyredhead my experience is that custom styling of buttons ends up being an all or nothing endeavour: one either has to take over everything, taking into account all accessibility considerations, in order to avoid unexpected interactions with browser default styling, or one has to avoid poking the bear entirely by avoiding extensive styling of buttons. I think the latter is a better approach for the core themes, but I'm always delighted to see plugins that go another way (eg using Material design).

PR #5143, opened 2020-11-29
Rename "references" to "backlinks" in the tiddler info panel

Jermolene:
TiddlyWiki 5 has always referred to "backlinks" as "references" in the user interface, despite the underlying filter operator being called "backlinks". It is proposed to switch to calling them "backlinks" for clarity.

I've made this a separate PR so that we can discuss the merits, and also in case we need to adjust any docs in the light of this change

### DIFF
--- a/core/language/en-GB/TiddlerInfo.multids
+++ b/core/language/en-GB/TiddlerInfo.multids
@@ -14,7 +14,7 @@ List/Caption: List
 List/Empty: This tiddler does not have a list
 Listed/Caption: Listed
 Listed/Empty: This tiddler is not listed by any others
-References/Caption: References
+References/Caption: Backlinks
 References/Empty: No tiddlers link to this one
 Tagging/Caption: Tagging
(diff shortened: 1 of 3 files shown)


saqimtiaz:
This is an interesting one. I actually think the term `references` is more intuitive for the lay person. 

However, the recent attention and focus on `backlinks` in other apps means that it is a term that is more familiar for many new users coming from those backgrounds. 

So the question for me is whether we think the awareness around the term **backlinks** is here to stay?

pmario:
At the moment "backlinks" are more visible. ... IMO we should change it for easier promotion!. .. We can change it back if we want. ...

BurningTreeC:
I also think that "backlinks" fits well. For me it's more intuitive

Jermolene:
Excellent, I'll merge this once I've updated the docs. I've found these tiddlers that need updating:

https://tiddlywiki.com/prerelease/#:InfoPanel%20%5B%5BUsing%20links%20to%20navigate%20between%20tiddlers%5D%5D

PR #5144, opened 2020-11-29
Create "Filter Run Prefix (Examples)" tiddler

rmunn:
Move the `+` vs `:intersection` example into a separate Examples tiddler which can be expanded on later. Also delete the mention of "temporary storage" as suggested in https://github.com/Jermolene/TiddlyWiki5/pull/5051#discussion_r524203257


Jermolene:
Thanks @rmunn

saqimtiaz:
@rmunn thank you!

PR #5146, opened 2020-11-30
Operator pad[] along with tests and docs

### LINKED ISSUE #5137: [IDEA] filter operator to pad a string to given length
I've needed this a few times and mostly recently for something that might be a candidate for the core. 

I am not sure what the best name for the operator would be suggest something of the form `[[...input...]pad[length],[optional-character-to-pad-with]]`

`[[12]pad[5]]` => `00012`

`[[12]pad[4],[-]]` => `--12`

saqimtiaz:
Fixes #5137 

This implements a string operator `pad[]` as described in #5137 and includes tests and docs.

The implementation is consistent with the behaviour of the JavaScript String methods padEnd and padStart.

`[[12]pad[5]]` => `00012`

`[[12]pad[4],[-]]` => `--12`

`[[12]pad[9],[abc]]` => `abcabca12`

`[[12]pad:suffix[9],[abc]]` => `12abcabca`

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -143,4 +143,33 @@ exports["search-replace"] = function(source,operator,options) {
 	return results;
 };
 
+exports.pad = function(source,operator,options) {
+	var results = [],
+		targetLength = operator.operand ? parseInt(operator.operand) : 0,
+		fill = operator.operands[1] || "0";
+
+	source(function(tiddler,title) {
+		if(title && title.length) {
+			if(title.length >= targetLength) {
+				results.push(title);
+			} else {
+				var padString = "",
+					padStringLength = targetLength - title.length;
+				while (padStringLength > padString.length) {
+					padString += fill;					
+				}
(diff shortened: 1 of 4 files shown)


kookma:
@saqimtiaz thank you for making this PR.
by `pad:suffix`, one expects to have also `pad:prefix`. In TW filter operators, while there are `after`, `'before` and `suffix`, `prefix` why not to have `padstart`, `padend` like JS or `padafter`, `padbefore`.

What do you think?

saqimtiaz:
@kookma `pad:prefix[]` is identical to `pad[]`

The names `prefix/suffix` are consistent with other operators like `trim[]`

kookma:
@saqimtiaz thank you for clarification

Jermolene:
Thanks @saqimtiaz (I think I might have merged the docs before the implementation, sorry!)

saqimtiaz:
@Jermolene all good, this PR was self-contained with code, docs and tests.

PR #5147, opened 2020-11-30
Tag trim operator as a string operator

rmunn:
With the new functionality of the `trim` operator in 5.1.23, it's now a general-purpose string manipulation tool. So it should be given the "String operators" tag so that people see it in the same list as `removeprefix` and `removesuffix`, etc.

**Note:** PR created against master branch rather than tiddlywiki-com branch because without the new 5.1.23 features, I'm not convinced that it needs the "String operators" tag, so this change should go live once 5.1.23 releases, not before. (And besides which, if I created this PR against tiddlywiki-com, it would likely end up causing a merge conflict once the 5.1.23 release happens due to the "tags:" line being so close to the "op-output:" line, which has been modified in master. So since the 5.1.23 release is so close, I decided this PR was better made against master than against tiddlywiki-com).

### DIFF
--- a/editions/tw5.com/tiddlers/filters/trim Operator.tid
+++ b/editions/tw5.com/tiddlers/filters/trim Operator.tid
@@ -1,12 +1,12 @@
 caption: trim
 created: 20190613153740241
-modified: 20190613153820282
+modified: 20201130040553874
 op-purpose: returns each item in the list with whitespace, or a given character string, trimmed from the start and/or end
 op-input: a [[selection of titles|Title Selection]]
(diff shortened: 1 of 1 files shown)


rmunn:
I've also documented the suffix in this PR since it's not merged yet.

rmunn:
I might want to change the suffixes from `prefix` / `suffix` for trimming from the front and back of the string, to just `front` / `back`. If so, that will be a new PR since I want to have this one be docs only. Thoughts?

Jermolene:
Thanks @rmunn

PR #5148, opened 2020-11-30
Update static.tiddler.html.tid with tc-static-story-river class

BurningTreeC:
This PR adds the `tc-static-story-river` class to the `static.tiddler.html` template which makes rendered tiddlers full-width


kookma:
@BurningTreeC 
Sometimes users adjust the story river width to certain values for specific purpose. With the current setting, each tiddler occupies the whole screen width in static html. Is it possible to change this PR somehow to have the story river in the center of page and keep the story width?

I think the best may be to have a config tiddler lets to set 100% available width or keep the story width. What do you think?

PR #5152, opened 2020-11-30
"dedupe" and "raw" suffixes for enlist-input operator

### LINKED ISSUE #5151: [BUG] New enlist-input operator needs "dedupe" and "raw" suffixes like enlist
**Describe the bug**
If the `list` field of a tiddler contains `1 2 2 3`, the filter `[[TiddlerName]get[list]enlist-input[]]` will return `1 2 3`, deduplicating the duplicate 2. This may or may not be what the user desires. I suggest adding `dedupe` and `raw` suffixes line the `enlist` operator has, so that the user can choose whether deduplication is desired or not.

As for which should be the default, my first thought was that the default behavior of `enlist-input` should be to *not* deduplicate, so that it defaults to be a no-op filter: what comes in is what goes out. OTOH, perhaps the default behavior should be to match the behavior of `enlist`, where dedupe is the default. Thoughts?

rmunn:
Fixes #5151. Includes unit tests.

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -61,7 +61,7 @@ exports.split = makeStringBinaryOperator(
 );
 
 exports["enlist-input"] = makeStringBinaryOperator(
-	function(a) {return $tw.utils.parseStringArray("" + a);}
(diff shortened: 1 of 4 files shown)


Jermolene:
Thanks @rmunn

PR #5154, opened 2020-11-30
Add radio actions, th-radio-variables hook and fix label refresh problem

pmario:
@Jermolene please add this to 5.1.23 .. I desperately need it for 1 of my plugins. 

Docs will follow if merged!

 - This PR adds an "actions" parameter to the widget 
 - It adds a `th-radio-variables` hook call, which allows plugin authors to pass additional variables to the actions. 
 - it fixes the "label" refresh problem see: [Google groups](https://groups.google.com/g/tiddlywiki/c/IUjHMUxxfKo/m/tbeGOu0oBAAJ) thread

Simple test code:

```
\define newText() value: $(actionValue)$

\define radioActions()
<$action-sendmessage $message="tm-notify" $param="test-radio-simple-template" value=<<actionValue>>
    current=<<currentTiddler>> />
<$action-createtiddler $basetitle="test-x" $overwrite="yes" text=<<newText>>/>
\end

<$tiddler tiddler="$:/temp/test">
	<$radio field="id" value="a" actions=<<radioActions>> > a </$radio>|
	<$radio field="id" value="b" actions=<<radioActions>> > b </$radio>|
	<$radio field="id" value="c" actions=<<radioActions>> > c </$radio>|
</$tiddler>

<$radio field="id" value="d"  actions=<<radioActions>> > d </$radio>|
<$radio field="id" value="e" actions=<<radioActions>> > e </$radio>|
```

test-radio-simple-template

```
value = <<value>>
```

### DIFF
--- a/core/modules/widgets/radio.js
+++ b/core/modules/widgets/radio.js
@@ -13,7 +13,6 @@ Set a field or index at a given tiddler via radio buttons
 "use strict";
 
 var Widget = require("$:/core/modules/widgets/widget.js").widget;
-
 var RadioWidget = function(parseTreeNode,options) {
 	this.initialise(parseTreeNode,options);
 };
@@ -37,8 +36,8 @@ RadioWidget.prototype.render = function(parent,nextSibling) {
 	// Create our elements
 	this.labelDomNode = this.document.createElement("label");
 	this.labelDomNode.setAttribute("class",
-   		"tc-radio " + this.radioClass + (isChecked ? " tc-radio-selected" : "")
-  	);
+		"tc-radio " + this.radioClass + (isChecked ? " tc-radio-selected" : "")
+	);
 	this.inputDomNode = this.document.createElement("input");
 	this.inputDomNode.setAttribute("type","radio");
 	if(isChecked) {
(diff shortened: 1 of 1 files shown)


pmario:
@Jermolene I'll add the same mechanism to the "range-widget" if you want?

pmario:
By default this PR only sends the `<<actionValue>>` variable to the actions function, which is consistent with all other functions that are used atm.

saqimtiaz:
@pmario this looks good for the most part and I think you should go ahead and add the docs to this PR for the `actions` parameter and the `actionValue` variable.

With regards to the new hook `th-radio-variables`, we probably have to consider that it will become a backwards compatibility concern later and something we have to maintain indefinitely. Especially if we want to revisit the original idea of the widget creating variables later this may not be prudent. 

I am not in favour of this hook unless perhaps it can somehow be implemented as a generic `get-action-variables` hook so that we do not need one such hook per triggering widget....

If we want to avoid the hook, one alternative would be for you @pmario to just subclass the radio widget for your plugin, and define your own `handleChangeEvent()` method. Minimal code and gives you a lot of flexibility.

pmario:
>I am not in favour of this hook unless perhaps it can somehow be implemented as a generic get-action-variables hook so that we do not need one such hook per triggering widget....

You are right. It probably should be implemented as a generic hook: `th-get-action-variables`. Since we also pass `this` along, it should be possible to do everything.

pmario:
@saqimtiaz .. I did have a closer look at the widgets, that allow us to have actions. The problem is, that the widget attribute init behaviour is fundamentally different. So the code for a generic hook will need complex code to make it work.

So I think using different hook names is way less overhead and will be much faster.

saqimtiaz:
@pmario not only will having so many hooks be confusing for developers/documentation, but consider though that when a plugin adds such a hook, it will change the behaviour for all `radio` widgets in the TiddlyWiki. This could lead to unexpected and unwelcome side effects with user defined variables getting overwritten etc. Especially if the same pattern is applied to other widgets.

Also adding hooks doesn't make sense if this is supposed to be a temporary fix until we have time to re-consider the original PR. Once we have the hook it is there to stay. We cannot remove it later.

Instead if you subclass the `radio` widget as say `wikilabs-radio`, you can customize the behaviour as you like and not clash with the expected behaviour of other radio widgets. Your custom behaviour will apply to only your custom subclassed radio widgets, not to all radio widgets. So I do think that subclassing widgets is the better approach for these situations.

```
/*\
title: $:/wikilabs-radio.js
type: application/javascript
module-type: widget-subclass
\*/
(function(){

/*global $tw: false */
"use strict";

exports.baseClass = "radio";

exports.name = "wikilabs-radio";

exports.constructor = function(parseTreeNode,options) {
}

exports.prototype = {};

exports.prototype.handleChangeEvent = function(event) {
	//customize variables here and handle change event
}

})();
```

pmario:
OK .. I did remove the hook

saqimtiaz:
@pmario I think this is good to go.

And we should create a ticket to continue the discussion on adding variables via attributes.

pmario:
New issue: widgets, that allow "actions" should be able to pass on "user defined" variables and attributes #5155

Jermolene:
Thanks @pmario @saqimtiaz

PR #5157, opened 2020-11-30
Update server/static.tiddler.html.tid with tc-static-story-river class

BurningTreeC:
This PR adds the `tc-static-story-river` class to the tiddler `$:/core/templates/server/static.tiddler.html`

@Jermolene - is this appropriate?

### DIFF
--- a/core/templates/server/static.tiddler.html.tid
+++ b/core/templates/server/static.tiddler.html.tid
@@ -19,7 +19,7 @@ title: $:/core/templates/server/static.tiddler.html
 </head>
 <body class="tc-body">
 <$transclude tiddler="$:/core/templates/server/static.sidebar.wikitext" mode="inline"/>
-<section class="tc-story-river">
+<section class="tc-story-river tc-static-story-river">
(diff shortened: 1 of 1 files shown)


BurningTreeC:
... Or is there a sidebar rendered?

Jermolene:
> @Jermolene - is this appropriate?

Is the intention to make the rendered single tiddler view not show the sidebar if it has been disabled in the wiki? Seems like a good idea.

Longer term, I think we need to split base.tid into smaller chunks so that we can completely exclude all the interactive styles from the static renderings.

BurningTreeC:
>Is the intention to make the rendered single tiddler view not show the sidebar if it has been disabled in the wiki? Seems like a good idea.

No, the intention here was making the story-river full width. But I forgot that in this case there's a rendered sidebar

>Longer term, I think we need to split base.tid into smaller chunks so that we can completely exclude all the interactive styles from the static renderings.

+1 for this

BurningTreeC:
I think I can close this PR

PR #5158, opened 2020-11-30
Add actions parameter to range-widget + docs!

pmario:
This is WIP. The docs contains an example, that creates a log-like temporary tiddler, where users can exactly see, what's going on and when the different actions fire!

This PR adds 4 actions:

 - actionsStart
 - actionsStop
 - actions

2 variables: `actionValue` and `actionValueHasChanged` See the docs!

It behaves the same way in modern browsers and contains some lines of code to behave well with IE11.
I did test it with IE11 wondows 10, FF latest, Edge latest. ..

[range-actions.zip](https://github.com/Jermolene/TiddlyWiki5/files/5618393/range-actions.zip)

@twMat ... Please test!

### DIFF
--- a/core/modules/widgets/range.js
+++ b/core/modules/widgets/range.js
@@ -8,125 +8,190 @@ Range widget
 \*/
 (function(){
 
-/*jslint node: true, browser: true */
-/*global $tw: false */
-"use strict";
-
-var Widget = require("$:/core/modules/widgets/widget.js").widget;
-
-var RangeWidget = function(parseTreeNode,options) {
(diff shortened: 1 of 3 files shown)


pmario:
It contains some extra code, to make it compatible with IE! Once IE is gone, the code can be simplified.

Jermolene:
Looks good thanks @pmario

pmario:
@Jermolene ... marked the PR as "ready for review".

Jermolene:
Thank you @pmario

PR #5159, opened 2020-11-30
Make buttons look like buttons in CupertinoDark, GruvboxDark and DesertSand palette

BurningTreeC:
This PR removes the background-color and border-color from buttons in the CupertinoDark, GruvboxDark and DesertSand palettes so that buttons look like buttons :grin:


Jermolene:
Thanks @BurningTreeC - I'm not against palettes that do recolour the buttons, I just think we should avoid it in Vanilla. Perhaps we should consider adding support for optionally completely taking over button styling, so that people can choose native vs. Material vs. Win95 buttons.

BurningTreeC:
That would be an idea worth a ticket @Jermolene +1

PR #5161, opened 2020-12-01
Docs for trim operator suffix should say "from version 5.1.23"

rmunn:
Trim suffix was introduced in version 5.1.23; docs should mention that.


saqimtiaz:
@Jermolene more docs for 5.1.23

Jermolene:
Thanks @rmunn @saqimtiaz

PR #5163, opened 2020-12-01
Fix #5162 - tiddler opening position is incorrect, bug in navigator.js

### LINKED ISSUE #5162: [BUG] Tiddler opening position in story is incorrect in 5.1.23 pre-release
As reported [here](https://groups.google.com/g/tiddlywiki/c/i_rzRBDZvn4), the setting for where in the story the tiddler is opened, is not respected.

I believe this bug was introduced in #4200 

`Navigator.prototype.addToStory` calls `Story.prototype.addToStory` with [4 arguments](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/navigator.js#L131) whereas it [accepts only 3](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/story.js#L37). 

Presumably the 4th argument needs to be merged into the 3rd as an options object. However it **does not seem like the 3rd argument as it is now is used at all**? 

I haven't had the time to fully read through and understand the logic of this PR, but it seems like it is not quite consistent and needs careful review.

@Jermolene @BurningTreeC @pmario

BurningTreeC:
This PR fixes #5162 - the error is that $tw.Story.addToStory accepts only 3 arguments but it was called with 4 where the fourth was the options object and the third was the (unnecessary) storyTitle

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -128,7 +128,7 @@ NavigatorWidget.prototype.replaceFirstTitleInStory = function(storyList,oldTitle
 
 NavigatorWidget.prototype.addToStory = function(title,fromTitle) {
 	if(this.storyTitle) {
-		this.story.addToStory(title,fromTitle,this.storyTitle,{
+		this.story.addToStory(title,fromTitle,{
 			openLinkFromInsideRiver: this.getAttribute("openLinkFromInsideRiver","top"),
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@BurningTreeC thank you.

BurningTreeC:
Ping @Jermolene

PR #5167, opened 2020-12-01
docs for filesystem changes

joshuafontany:
Docs for the filesystemAdaptor changes, PR to tiddlywiki-com branch.


saqimtiaz:
@joshuafontany looks good, thank you!

rmunn:
If this is documentation for #5113, then I believe this PR should be against `master` rather than `tiddlywiki-com`. The tiddlywiki-com branch is for updating the documentation that's currently on the site, whereas PRs against master will be added to the site only once 5.1.23 releases. If this documentation is for the changes that won't be released until 5.1.23, then it should be against `master` so as to not confuse 5.1.22 users who might wonder why the docs are telling them something that doesn't work.

joshuafontany:
@saqimtiaz Slight cleanup and good to go. :)

PR #5171, opened 2020-12-01
Fix typo in x-listops.js

### LINKED ISSUE #5170: [BUG] - x-listops.js throws Syntax Error on Internet Explorer
**Describe the bug**
Using IE 11 on startup there's a RSOD saying "Syntax Error in x-listops.js"
I can't track the problem down since the Konsole doesn't show anything, but maybe @saqimtiaz has an idea what could be the problem?

saqimtiaz:
Fixes #5170 

Oddly enough my jslint did not catch this.


Jermolene:
Thanks @BurningTreeC

PR #5172, opened 2020-12-01
Update CodeMirror to 5.58.3 - fix vim-mode cursor invisible

BurningTreeC:
This PR updates the codemirror plugin to v5.58.3 and fixes a problem where the cursor isn't visible using the vim keymap


Jermolene:
Thanks @BurningTreeC

PR #5173, opened 2020-12-01
Don't minify codemirror.js

BurningTreeC:
The vim-cursor problem that should have been solved by #5172 is only solved when codemirror.js is NOT minified. I've tested this several times and there seems to be a problem with the minified code. This PR adds the un-minified version of codemirror.js


Jermolene:
Wow, I've heard of such things. Have you reported upstream?

pmario:
But this makes the plugin unusable because of size!

pmario:
I don't use the vim keymap. So why should I want to carry the extra weight around?

BurningTreeC:
@pmario I'll see if after reporting this upstream there's a quick fix for this

PR #5174, opened 2020-12-02
Fix a couple typos in "deprecated feature" warning

rmunn:
One tiddler had been renamed since the deprecation warning tiddler was created, and there were also a couple of minor punctuation and/or spelling errors.

### DIFF
--- a/editions/tw5.com/tiddlers/system/Deprecated_-_What_does_it_mean.tid
+++ b/editions/tw5.com/tiddlers/system/Deprecated_-_What_does_it_mean.tid
@@ -7,9 +7,9 @@ First of all: ''Keep calm!''
 
 {{$:/deprecated}}
 
-For ~TiddlyWiki it means, that you should not use this mechanism for new content anymore! ''AND you should update your existing content''!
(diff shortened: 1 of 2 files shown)


rmunn:
In addition to correcting a couple minor English-language mistakes in the text of the tiddlers, I also corrected the spelling of a parameter in the deprecated macro ("superseeded" should have been "superseded"), because I didn't see it used by parameter name anywhere, only by position. But if you think it's likely that people have called that parameter by name in their own private wikis and changing the parameter name would make those calls fail, I can undo that part of the change. I don't think it's likely at all, though; I think the only place this macro will be used is in TW itself, so I think it's safe to fix the spelling of the parameter.

Jermolene:
> In addition to correcting a couple minor English-language mistakes in the text of the tiddlers, I also corrected the spelling of a parameter in the deprecated macro ("superseeded" should have been "superseded"), because I didn't see it used by parameter name anywhere, only by position. But if you think it's likely that people have called that parameter by name in their own private wikis and changing the parameter name would make those calls fail, I can undo that part of the change. I don't think it's likely at all, though; I think the only place this macro will be used is in TW itself, so I think it's safe to fix the spelling of the parameter.

Good catch, I think it's reasonable to correct it.

PR #5176, opened 2020-12-02
Fix filesystem regression

joshuafontany:
I was able to track down the bugs around loading multiple wikis via includeWikis config. It seems like @Jermolene had been thinking about these probable problems, so I used his implementation of the $:/config/OriginalTiddlerPaths to resolve a bunch of bugs.

I was also able to resolve a very confusing part of the use of the `isEditableFile` flag in `tiddlywiki.files` configs. This really simplifies using that feature, no longer requireing a $:/config/FileSystemPath filter to place them "back" into the correct directory.

Now, all files that are not located within the base writable `$tw.boot.wikiTiddlersPath`, all files loaded from wikis that have the config-flag `"retain-original-tiddler-path": true`, and all files loaded via `tiddlywiki.files` with `isEditableFile` set have their relative paths saved to **$:/config/OriginalTiddlerPaths**. This is the preffered path when no filter-matches from FileSystemPaths applies.

I ahve tested this with my filesystem Test Wiki "Empty/" and also the TW.com edition. Looks good. 👍


Jermolene:
Great, thanks @joshuafontany

pmario:
>I ahve tested this with my filesystem Test Wiki "Empty/" and also the TW.com edition. Looks good.

You tested it with tw5.com-server edition. right?

joshuafontany:
@pmario Yes. Before this I didn't really get how the editions were setup to serve via node.js and build a single file from different edition folders & how you can load multiple wiki from the tiddlywiki.info config.

Definitely something I'll be using going forwards.

PR #5180, opened 2020-12-02
Re-add minified codemirror.js, without minification bug

BurningTreeC:
There was a bug in the compression I used, I don't know which one, but this time I used another compression, the file is even a bit smaller in file-size and the vim-cursor works


Jermolene:
Hi @BurningTreeC 

> There was a bug in the compression I used, I don't know which one, but this time I used another compression, the file is even a bit smaller in file-size and the vim-cursor works

If we're doing the minifying ourselves then ideally we ought to automate that build step, too.

BurningTreeC:
>If we're doing the minifying ourselves then ideally we ought to automate that build step, too.

Yes if that's possible... :+1:

PR #5181, opened 2020-12-02
Small tweak for codemirror selected-linenumber

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

kookma:
@BurningTreeC 
The below image shows default theme and palette in TW 5.1.23.

image


I think having different background colors for body and line number make it more readable or at list a divider line.

What do you think?


Windows 10, Chrome 87, TW 5.1.23pre latest

BurningTreeC:
@kookma I don't understand what you'd like. Would you mind explaining it a bit better?

kookma:
@BurningTreeC 
I attached a screenshot from 5.1.22. Please see the different background colors used for line numbers (grey) and editor body (white)

image

So you can clearly distinguish the border between line numbers and body text. This is not the case with Code mirror 5.1.23 as I have attached above.

What I mean is use different background colors for line numbers and main body!

BurningTreeC:
Ok @kookma I understand

PR #5185, opened 2020-12-02
Extend Eventcatcher to handle multiple events

saqimtiaz:
@Jermolene this makes the following changes:
- replaces `type` attribute with `events`, which is a space separated list of events.
- replaces `actions` attribute with support for separate actions per event, where each event type is prefixed with `actions-` to get the attribute name for the actions. For example: `actions-dblclick`
- adds variables for `event.type` and `event.details` properties of the event
- updates documentation.

Code changes are minimal, the diff shows a lot of green due to a change in indent level for the code that is now inside a loop.

Background: https://github.com/Jermolene/TiddlyWiki5/pull/5086#issuecomment-737308560

### DIFF
--- a/core/modules/widgets/eventcatcher.js
+++ b/core/modules/widgets/eventcatcher.js
@@ -142,7 +142,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 EventWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.types || changedAttributes["tag"]) {
+	if(changedAttributes["events"] || changedAttributes["tag"]) {
 		this.refreshSelf();
 		return true;
 	} else if(changedAttributes["class"]) {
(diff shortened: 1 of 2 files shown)


saqimtiaz:
@Jermolene one thought was to allow plugins to customize what variables are created via a hook just before invocation of actions:

`variables = $tw.hooks.invokeHook("th-eventcatcher-getvars",event,variables,selectedNode,domNode);`

Jermolene:
Great thanks @saqimtiaz. I'll be looking forward to see what people do with this.

saqimtiaz:
@Jermolene just to be clear I didn't add the hook I mentioned at this time.

It could be useful if there are variables needed that we haven't thought of but can also be added later.

Jermolene:
Sorry I meant to say that here I'm not sure a hook is appropriate here. We generally use them when there's a clear use case that we have opted not to cater for in the core. Here the scope for what information might be exposed as variables is limited to what we have available, which is known in advance. So, if there's information that the current design doesn't expose then there's probably a case for doing so for everyone, rather than continuing to support a hook.

saqimtiaz:
@Jermolene that makes sense, I had mixed feelings about it myself.

Jermolene:
These are tricky decisions! Hooks feel like an unavoidable hostage to fortune sometimes.

PR #5191, opened 2020-12-04
Add rawunicode suffix to (json)stringify operators

### LINKED ISSUE #5189: [IDEA] Suffix for stringify and jsonstringify to not escape Unicode characters
The `stringify` and `jsonstringify` filter operators escape Unicode characters in addition to escaping characters like `"` and `\` and newlines. While this is useful for portability (it ensures that the resulting string contains only ASCII characters and is therefore safe to read in any context), it may be overkill for some use cases. I'd like to see a suffix added to these operators that allows them to pass Unicode codepoints U+0080 and higher unchanged. There are many times when I know the tool chain I'm using is handling Unicode correctly all the way through, and I don't need þ turned into `\u00fe` because I know it will come out as þ on the other side of my tool chain. This would be especially relevant for people working in languages that don't use the Latin alphabet, where nearly every letter of their tiddlers would be above U+0080, so escaping those would turn three bytes of UTF-8 into six bytes, or two bytes of UTF-16 into twelve bytes.

I suggest adding the suffix `rawunicode` for the `stringify` and `jsonstringify` filter operators to enable skipping codepoints U+0080 and up. The suffix `raw` would also work, but since we use that as the opposite of `dedupe` it might cause confusion, or it might cause people to think that characters U+001F and below would also be skipped (no; U+001F and below would still be encoded). So `rawunicode` would best communicate what the suffix does, I think.

Implementation would be very simple: add a Boolean parameter to the `stringify` and `jsonStringify` functions in utils.js, named something like 'rawUnicode', which will default to undefined (and therefore be considered false if not passed) for compatibility with existing calls to those functions. If it is true, then the final `.replace()` call that replaces characters from U+0080 and up is just not called.

Leaving it as-is is not necessarily a bad option, because I might be a rarity in wanting this. But since the implementation is simple and it will be simple to explain, I think it's worth considering.

rmunn:
Default with no suffix is pre-5.1.23 behavior, escaping all Unicode characters for maximum compatibility (avoids encoding issues). New `rawunicode`suffix allows passing through Unicode characters U+0080 and up unchanged, for cases where you know your tools are handling encoding correctly and you want less verbose escaping.

Fixes #5189.

### DIFF
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -77,7 +77,7 @@ exports.encodehtml = function(source,operator,options) {
 exports.stringify = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		results.push($tw.utils.stringify(title));
+		results.push($tw.utils.stringify(title,(operator.suffix === "rawunicode")));
 	});
 	return results;
 };
@@ -85,7 +85,7 @@ exports.stringify = function(source,operator,options) {
 exports.jsonstringify = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		results.push($tw.utils.jsonStringify(title));
+		results.push($tw.utils.jsonStringify(title,(operator.suffix === "rawunicode")));
 	});
 	return results;
 };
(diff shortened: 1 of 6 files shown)


Jermolene:
Thanks @rmunn

PR #5193, opened 2020-12-04
Reduce operator: second optional parameter instead of suffix

### LINKED ISSUE #5190: [IDEA] reduce operator: initial value from second operand.
Use an optional second operand for the initial value for reduce, so that variables and references can be used to assign it.

@Jermolene if we want to support this we should make the change before 5.1.23 goes live.

saqimtiaz:
Fixes #5190

Replaces the suffix for `reduce[]` with an optional second parameter so that the initial value can also be set via variables.


Jermolene:
Thanks @saqimtiaz

Issue #5130, opened 2020-11-28
Timeline for 5.1.23 release?


saqimtiaz:
@Jermolene perhaps now would be a good time to set a timeline for the 5.1.23 release so we can co-ordinate our efforts better?

I don't believe we have any blocking issues that need resolving. 

~~We should decided if we want to add an upgrader module for dealing with the journal tags: https://github.com/Jermolene/TiddlyWiki5/issues/4721~~

There are a couple of almost ready PRs that would be good to include but need a bit more work/testing, but even there a timeline might push things along:
- ~~#5113~~
- ~~#5026~~
- ~~#5052~~
- ~~#5120~~

Is there anything else that I am not thinking of @pmario @BurningTreeC @joshuafontany ?

BurningTreeC:
@saqimtiaz we should definitely add an upgrader module for dealing with #4721

Jermolene:
Hi @saqimtiaz @BurningTreeC I was also thinking about the four tickets that @saqimtiaz mentions in the OP, but I wasn't sure what we'd decided about #4721, and will leave a comment there.

My main task for this weekend is to bring the release note up to date and hopefully rearrange it a bit as the current categories aren't working so well with so many new items.

If all goes well with that, I should be able to send out the usual last call on the mailing list on Monday, and then perhaps release next weekend.

saqimtiaz:
@Jermolene perhaps you could also check if you can reproduce this issue on OSX:
https://github.com/Jermolene/TiddlyWiki5/commit/6c98bb706a15b353a92895529a69bece11f613ff#commitcomment-44594409

I was not able to reproduce it in Chrome on OSX, Windows or Linux.

Jermolene:
I'm just pinging other contributors of this release to let you all know our plans to finalise v5.1.23 shortly.

This is a good time to test your plugins etc against the prerelease, and bump any PRs you'd like to see merged for v5.1.23.

@adithya-badidey
@Arlen22
@bimlas
@BramChen
@BurningTreeC
@danielo515
@ento
@favadi
@fkohrt
@flibbles
@gera2ld
@ibnishak
@idotobi
@jdangerx
@jjduhamel
@Kamal-Habash
@Marxsal
@mocsa
@NicolasPeton
@passuf
@pmario
@rmunn
@saqimtiaz
@twMat
@default-kramer

rmunn:
[...]

PR #5194, opened 2020-12-04
Make sidebarsegments/search count-matches-button display the count of ALL search results (de-duplicated)

BurningTreeC:
This PR makes the count-matches-button of the sidebar search show the count of ALL matches but de-duplicated. Matches the behavior of 5.1.22


BurningTreeC:
@Jermolene @saqimtiaz @bimlas 

In https://github.com/Jermolene/TiddlyWiki5/issues/5130#issuecomment-738748166 @saqimtiaz mentioned to possibly trim the prefix of the entered search text so that (3) spaces don't trigger search results. Should we do that?

Jermolene:
> In https://github.com/Jermolene/TiddlyWiki5/issues/5130#issuecomment-738748166 @saqimtiaz mentioned to possibly trim the prefix of the entered search text so that (3) spaces don't trigger search results. Should we do that?

I'd be inclined not to do so. I can see how the current behaviour might cause frustration, but I'm not sure that the proposed solution might not have unexpected impacts. Right now it is actually quite useful occasionally to be able to search for leading whitespace (eg if you want to find a word that is also found embedded within other words).

PR #5195, opened 2020-12-04
Make the tag-picker tags-dropdown more performant

BurningTreeC:
This PR makes the tags-dropdown of the tag-picker more performant by replacing the `macrocalls` for the `tag-button` with a `transclusion` of `$:/core/ui/TagPickerTagTemplate` (title to discuss ? ) ... The tag-template-tiddler minimizes macrocalls, too and variables for the `<<contrastcolour>>` macro are defined as early as possible, so that some don't have to be re-read for each tag


Jermolene:
Thanks @BurningTreeC

PR #5196, opened 2020-12-04
:reduce filter run prefix

saqimtiaz:
@Jermolene this adds a filter run prefix equivalent to the `reduce[]` operator. 

`[tag[shopping]] :reduce[get[quantity]add<accumulator>]`

is equivalent to:

```
\define num-items() [get[quantity]add<accumulator>]

[tag[shopping]reduce<num-items>]
```

AND

`[tag[non-existent]] :reduce[get[price]multiply{!!quantity}add<accumulator>] :else[[0]]`

is equivalent to:

```
\define add-price() [get[price]multiply{!!quantity}add<accumulator>]

[tag[non-existent]reduce:0<add-price>]
```

If you want to leave this for after 5.1.23 that is OK for me.

### DIFF
--- a/core/modules/filterrunprefixes/reduce.js
+++ b/core/modules/filterrunprefixes/reduce.js
@@ -1,0 +1,49 @@
+/*\
+title: $:/core/modules/filterrunprefixes/reduce.js
+type: application/javascript
+module-type: filterrunprefix
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter prefix function
+*/
+exports.reduce = function(operationSubFunction) {
+	return function(results,source,widget) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @saqimtiaz I think it's very low impact so I'll go ahead and merge for v5.1.23

saqimtiaz:
@Jermolene I'll update the docs tomorrow. Thank you.

rmunn:
@saqimtiaz - Please comment on #5246. There's a difference of behavior between this implementation and the way the `reduce` operator is implemented, and I think it's important to get it sorted out before the release. I also wonder what we should do about the example of `=1 =2 =3 :reduce[multiply<accumulator>]` — I think people will be surprised when that produces a result of 0, and #5246 has some discussion of what to do about that.

PR #5197, opened 2020-12-04
EventCatcher: Ensure new variables are strings

saqimtiaz:
@Jermolene this just ensure the new variables I added in the last round of improvements are always strings.


Jermolene:
Thanks @saqimtiaz

PR #5200, opened 2020-12-04
Update CodeMirror styles to set selection colors based on foreground contrastcolour

BurningTreeC:
This PR calculates the selection color based on the foreground color and two colors from the solarized palette


BurningTreeC:
@Jermolene - can we merge this before the release?

Jermolene:
Yes apologies, I've been a bit behind today

PR #5202, opened 2020-12-05
Fix for filterrunprefixes using $tw.wiki

flibbles:
Fix for `and` filterrunprefix. It was using $tw.wiki instead of the wiki supplied in its widget, which meant that it was impossible to test any filters in a testing environment using temporary wiki objects.

A simple one line fix, but it did mean touching all the filter tests. Yech, but it's done now.

### DIFF
--- a/core/modules/filterrunprefixes/and.js
+++ b/core/modules/filterrunprefixes/and.js
@@ -16,10 +16,10 @@ Equivalent to + filter run prefix.
 /*
 Export our filter prefix function
 */
-exports.and = function(operationSubFunction) {
+exports.and = function(operationSubFunction,options) {
 	return function(results,source,widget) {
 		// This replaces all the elements of the array, but keeps the actual array so that references to it are preserved
-		source = widget.wiki.makeTiddlerIterator(results);
+		source = options.wiki.makeTiddlerIterator(results);
 		results.splice(0,results.length);
 		$tw.utils.pushTop(results,operationSubFunction(source,widget));
 	};
--- a/core/modules/filterrunprefixes/and.js
+++ b/core/modules/filterrunprefixes/and.js
@@ -19,7 +19,7 @@ Export our filter prefix function
 exports.and = function(operationSubFunction) {
 	return function(results,source,widget) {
 		// This replaces all the elements of the array, but keeps the actual array so that references to it are preserved
-		source = $tw.wiki.makeTiddlerIterator(results);
+		source = widget.wiki.makeTiddlerIterator(results);
 		results.splice(0,results.length);
 		$tw.utils.pushTop(results,operationSubFunction(source,widget));
 	};
--- a/core/modules/filterrunprefixes/reduce.js
+++ b/core/modules/filterrunprefixes/reduce.js
@@ -12,13 +12,13 @@ module-type: filterrunprefix
 /*
 Export our filter prefix function
 */
-exports.reduce = function(operationSubFunction) {
+exports.reduce = function(operationSubFunction,options) {
 	return function(results,source,widget) {
 		if(results.length > 0) {
 			var accumulator = "";
 			for(var index=0; index<results.length; index++) {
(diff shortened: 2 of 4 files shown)


flibbles:
If anybody has a problem with touching all the tests like this, wiki.filterTiddlers could be made to create dummy widget itself if one isn't supplied, though that would be a little janky.

flibbles:
Okay...

I just found out the hard way that even though I made all the tests pass, tiddlywiki still breaks down, because it expects that filterTiddlers widget argument to be optional in _many_ places.

Here is a new version of it which creates a dummy widget inside of filterTiddlers that's complete enough for all the filter operators, at least all the core ones, since they don't seem to look at it too closely.

saqimtiaz:
> that's complete enough for all the filter operators, at least all the core ones, since they don't seem to look at it too closely.

I am a little concerned by this part, from the point of view of plugins that implement filters and backwards compatibility. So I am still wondering if it may not be better to call the filter run prefixes as follows:
`filterRunPrefixes["or"](operationSubFunction,self)` instead of `filterRunPrefixes["or"](operationSubFunction)`
where `self` is the wiki object. 

@flibbles You've missed the reduce filter run prefix that I mentioned, which also uses `$tw.wiki.makeTiddlerIterator`

flibbles:
If we go that route, I might suggest `filterRunPrefixes["or"](operationSubFunction,options)` with options being `{wiki: self}`. It seems more in line with how tiddlywiki is designed.

I can fix the reduce prefix with this PR if you'd like. I missed it because there isn't a single unit test that touches it. is that... just okay around here?

saqimtiaz:
Personally I would favour the `filterRunPrefixes["or"](operationSubFunction,options)` approach.

@flibbles the `reduce` prefix PR was made yesterday and I had not expected it to get merged without further discussion. I had not pushed the tests because I realized that the `reduce[]` operator that it is based on also has no tests, so I was still working on those.  

I am working on a docs update for `reduce` this weekend and will include the tests. Feel free to leave out the `reduce` prefix if you like and I will patch it according to whatever gets decided here.

flibbles:
Oh! YOU wrote the reduce prefix. I'm sorry. Didn't mean to trashtalk it. I just jumped back into Tiddlywiki a day ago. Haven't been paying attention.

I'll switch over to the `filterRunPrefixes["or"](operationSubFunction,options)` system now.

Is the whole prefix thing completely new? Because I've been working on a LinkedList alteration to it to address performance issues with filters with large numbers of tiddlers.

saqimtiaz:
@flibbles no worries! Sometimes things move surprisingly fast in the TiddlyWiki realm and it can catch us by surprise.

Yes the filter run prefixes are new for 5.1.23, merged maybe 5-6 weeks ago? It is a refactoring of `compileFilter()` to make the prefixes extensible via their own `module-type`.

flibbles:
Okeydokey. Here is a new version that uses an options argument.

Now to finish up my LinkedList proposal. I'd love to get it in before the current fulterrunprefixes are written in stone. Currently `[all[tiddlers]!tag[myTag]]` can take up to 4 seconds if you have 30,000-40,000 tiddlers. If we use a LinkedList, it only takes 1/10 of a second. This was a problem that existed long before filterrunprefixes, but this might be the best time to address it.

saqimtiaz:
@flibbles looks good. I don't foresee a problem with making changes for better performance, as long as we still have the same flexibility as the filter run prefixes currently offer. 

Just in case you haven't seen the relevant posts, the TW 5.1.23 release is imminent in the next few days.

flibbles:
Here are my proposed Linked list changes: #5206.

It changes what gets sent to filterrunprefixes, so here's hoping it can get in before the current method signatures get codified, though I realize I'm showing up just as the store is closing.

As for this fix, I think it's good to merge.

Jermolene:
Looks good thank you @flibbles

PR #5204, opened 2020-12-05
Add selection-background/foreground to GruvboxDark and Nord palettes

BurningTreeC:
This PR adds the `selection-background` and `selection-foreground` to the GruvboxDark and Nord palettes


BurningTreeC:
@Jermolene - can we merge this before the release?

Issue #5188, opened 2020-12-03
[BUG] Compare Tiddlywiki 5.1.23pre and 5.1.22 for adding Tags


kookma:
**Describe the bug**
It seems TW 5.1.23 is slower in adding tags compared to TW 5.1.22. See below screen recording, the measured time is 4 seconds more for TW 5.123 (15 vs. 19).

**Screen Recordings**
![TW5 1 22](https://user-images.githubusercontent.com/830394/100976845-5a651b80-3555-11eb-914c-4d947b79fb3e.gif)
![TW5 1 23](https://user-images.githubusercontent.com/830394/100976912-75d02680-3555-11eb-95f3-33f3b1274a5e.gif)

Is there any way to speedup the process in TW5.1.23? This lag is considerable when you have alot of tags.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome 87
 - Version 5.122 and 5.123

BurningTreeC:
@kookma this lag depends on system performance and yes, it's more obvious when there are many tags. The problem is the tagpicker setting focus on the input field right after adding a tag, which makes the tag-dropdown reload the full tags-list

Jermolene:
Thanks @kookma @BurningTreeC this seems like it could impact a lot of users. I wonder if there's any potential mitigations?

My first thought is probably too complicated: that in the case that the new tag text input receives focus via the keyboard that we don't display the dropdown until the user types something, or clicks on the arrow.

I'll do some performance analysis when I get back to my desk.

BurningTreeC:
@Jermolene - my first thought was making it configurable if the tags-input should get focus right after adding a tag

BurningTreeC:
@kookma @Jermolene - there's still the `$:/config/Tags/MinLength` setting that would help here

Jermolene:
One consideration is that displaying the dropdown on focus is very useful when using the mouse.

BurningTreeC:
>One consideration is that displaying the dropdown on focus is very useful when using the mouse.

Yes, indeed.

BurningTreeC:
@kookma @Jermolene - I wonder if `retain="yes"` on the tags-dropdown would speed things up a little

BurningTreeC:
@kookma - could you try changing the line 73 (in $:/core/macros/tag-picker):

from: 

```
<$reveal state=<<qualify "$:/state/popup/tags-auto-complete">> type="nomatch" text="" default="">
```

to:

[...]

PR #5205, opened 2020-12-05
Cache result of parsing macro calls which don't use text subsitution

Jermolene:
After the [recent discussion](https://github.com/Jermolene/TiddlyWiki5/issues/5188#issuecomment-737896712) of the relatively poor performance of macros relative to transclusions, it occurred to me that there is a possible route to caching the results of parsing a macro. The "catch" is that it only kicks in for macros that do NOT use text substitution of parameters or variables (ie `$parameter$` or `$(variable)$`).

Testing it turned out to a challenge because nearly all of the core macros do still use text substitution. The exception is the `tree` macro. I made a test rig of a tiddler that displayed 1000 instances of the tree macro. The optimisation improved the initial render time by 20% which seems worthwhile (especially given the rough and ready benchmark being used).

We need to decide whether this can go into v5.1.23. It's a deep modification, but it doesn't seem especially risky. We already cache parse trees so we know that part of it is safe. I'd be interested in other views.

### DIFF
--- a/core/modules/widgets/macrocall.js
+++ b/core/modules/widgets/macrocall.js
@@ -55,9 +55,18 @@ MacroCallWidget.prototype.execute = function() {
 	// Are we rendering to HTML?
 	if(this.renderOutput === "text/html") {
 		// If so we'll return the parsed macro
-		var parser = this.wiki.parseText(this.parseType,text,
-							{parseAsInline: !this.parseTreeNode.isBlock});
-		parseTreeNodes = parser ? parser.tree : [];
+		// Check if we've already cached parsing this macro
+		var parser;
+		if(variableInfo.srcVariable && variableInfo.srcVariable.parser) {
+			parser = variableInfo.srcVariable.parser;
+		} else {
+			parser = this.wiki.parseText(this.parseType,text,
+								{parseAsInline: !this.parseTreeNode.isBlock});
+			if(variableInfo.isCacheable && variableInfo.srcVariable) {
+				// variableInfo.srcVariable.parser = parser;
+			}
+		}
+		var parseTreeNodes = parser ? parser.tree : [];
 		// Wrap the parse tree in a vars widget assigning the parameters to variables named "__paramname__"
 		var attributes = {};
 		$tw.utils.each(variableInfo.params,function(param) {
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -113,7 +113,8 @@ Widget.prototype.getVariableInfo = function(name,options) {
 	// Check for the variable defined in the parent widget (or an ancestor in the prototype chain)
 	if(parentWidget && name in parentWidget.variables) {
 		var variable = parentWidget.variables[name],
-			value = variable.value,
+			originalValue = variable.value,
+			value = originalValue,
 			params = this.resolveVariableParameters(variable.params,actualParams);
 		// Substitute any parameters specified in the definition
 		$tw.utils.each(params,function(param) {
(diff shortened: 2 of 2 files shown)


saqimtiaz:
@Jermolene I am all for performance optimization and I think this is a worthwhile change.

That said I feel we keep finding unanticipated consequences of core changes very often and there is very little time for user feedback and testing. So I would favour either delaying this until after 5.1.23 or pushing back the release. I think delaying this to after 5.1.23 would be OK especially considering that most macros do use text substitutions and therefore the savings would be minimal for most users.

On a related note, I have a copy of my Streams plugin with the entire TiddlyToolmap from Dynalist in there:
https://saqimtiaz.github.io/sq-tw/temp/streams-tiddlytoolmap%20-2.html

The initial render time was decreased almost 3 fold switching from using macros to transcluded tiddlers!
I would be curious to compare this approach as well performance wise, the macros in question do not do any text substitution.

kookma:
@Jermolene I think it is better to postpone this to after 5.1.23. I have one general reason!

My experience shows most users do not try prerelease and so  always one sees issue reporting after a major release and then you have to release a bug fix / minor release. 

So, as you planned, it would be nice to have 5.1.23 final release next week.

saqimtiaz:
@Jermolene shall we get this a go too since we are merging and testing other changes?

Jermolene:
Thanks @saqimtiaz

rmunn:
@Jermolene - This appears to have been merged in a non-working state:

```
			if(variableInfo.isCacheable && variableInfo.srcVariable) {
				// variableInfo.srcVariable.parser = parser;  // <-- This is commented out; why?
			}
```

Was that intentional? If I'm testing this, should I uncomment that line in my local copy?

Jermolene:
Thanks @rmunn I'd commented it out for some performance testing.

kookma:
@Jermolene would you please advice me how can I do some tests to see the effects of this improvement?
Also do you plan to add any documentation for this change?

saqimtiaz:
@kookma this change applies to macros that do not use text substitution via `$parameter$` or `$(variable)$` syntax. So you would need to set up a test tiddler that has a large number of such macros to test. 

I don't think there is any need to test the performance outright, rather what is helpful is just testing that existing macros work as they should.

With regards to documentation we can add a small note to the recent documentation about performance but there isn't much to document as it isn't a user facing feature but rather an internal change.

kookma:
@saqimtiaz much appreciated

core/modules/widgets/macrocall.js
```diff
@@ -55,9 +55,18 @@ MacroCallWidget.prototype.execute = function() {
...
-		parseTreeNodes = parser ? parser.tree : [];
+		// Check if we've already cached parsing this macro
+		var parser;
+		if(variableInfo.srcVariable && variableInfo.srcVariable.parser) {
+			parser = variableInfo.srcVariable.parser;
+		} else {
+			parser = this.wiki.parseText(this.parseType,text,
+								{parseAsInline: !this.parseTreeNode.isBlock});
+			if(variableInfo.isCacheable && variableInfo.srcVariable) {
+				// variableInfo.srcVariable.parser = parser;
```
pmario, line 66: @Jermolene .. Just found this line. Is it intended, to be a comment?

pmario: @Jermolene ... I'm still not sure, if this comment is intended?

Jermolene: I think that must be an outdated diff, the current version at https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/macrocall.js doesn't have it

PR #5206, opened 2020-12-05
use a linked list for filter runs.

flibbles:
I know V5.1.23 is just about to release, but hear me out.

Right now, filters use `$tw.utils.pushTop` to assemble filter results, but it's really not scalable. It uses an array, and every pushTop is an O(n) operation, and since we build filters using that, it means filter manipulation is an O(n^2) task.

Link lists though allow for O(1) manipulation, whether it's pushTop, or push, or remove, or anything. How much of a benefit are they? Take this:
```javascript
var wiki = new $tw.Wiki();
for (var i = 0; i < 30000; i++) {
    wiki.addTiddler({title: i.toString()});
}
$tw.testWiki = wiki;
```
Then, lets use my dinky laptop to run this filter: `$tw.testWiki.filterTiddlers("[!tag[A]] [!tag[B]]")`

Screen Shot 2020-12-05 at 10 46 45 AM

It took over 2 seconds. (This is on Chrome. On Firefox it's way worse, but the profiler doesn't work as well.) Now let's try it with a _linked list_.

Screen Shot 2020-12-05 at 10 50 33 AM

About 35 ms.

It's that unavoidable indexOf call in pushTop which bottlenecks things. I realize not a lot of people are going to have 30,000+ tiddlers in their project. I've only got about 6000, but I'm noticing that my filters are getting a little sluggish.

So I know that 5.1.23 is about to ship, which is why I finally decided to get this implemented now. I can understand if we don't get it in, but I'd hate to lock down the filterrunprefix methods with passing them arrays when something so much better exists.

### DIFF
--- a/core/modules/utils/linked-list.js
+++ b/core/modules/utils/linked-list.js
@@ -15,14 +15,15 @@ function LinkedList() {
 
 LinkedList.prototype.clear = function() {
 	this.index = Object.create(null);
+	// LinkedList performs the duty of both the head and tail node
 	this.next = this;
 	this.prev = this;
 	this.length = 0;
 };
 
 LinkedList.prototype.remove = function(value) {
-	if ($tw.utils.isArray(value)) {
-		for (var t=0; t<value.length; t++) {
+	if($tw.utils.isArray(value)) {
+		for(var t=0; t<value.length; t++) {
 			this._removeOne(value[t]);
 		}
 	} else {
@@ -32,7 +33,7 @@ LinkedList.prototype.remove = function(value) {
 
 LinkedList.prototype._removeOne = function(value) {
 	var node = this.index[value];
-	if (node) {
+	if(node) {
 		node.prev.next = node.next;
 		node.next.prev = node.prev;
 		this.length -= 1;
@@ -52,12 +53,12 @@ LinkedList.prototype._linkToEnd = function(node) {
 };
 
 LinkedList.prototype.push = function(/* values */) {
-	for (var i = 0; i < arguments.length; i++) {
+	for(var i = 0; i < arguments.length; i++) {
 		var value = arguments[i];
 		var node = {value: value};
 		var preexistingNode = this.index[value];
 		this._linkToEnd(node);
-		if (preexistingNode) {
+		if(preexistingNode) {
 			// We want to keep pointing to the first instance, but we want
 			// to have that instance (or chain of instances) point to the
 			// new one.
@@ -72,14 +73,14 @@ LinkedList.prototype.push = function(/* values */) {
 };
 
 LinkedList.prototype.pushTop = function(value) {
-	if ($tw.utils.isArray(value)) {
-		for (var t=0; t<value.length; t++) {
+	if($tw.utils.isArray(value)) {
+		for(var t=0; t<value.length; t++) {
 			this._removeOne(value[t]);
 		}
 		this.push.apply(this, value);
 	} else {
 		var node = this._removeOne(value);
-		if (!node) {
+		if(!node) {
 			node = {value: value};
 			this.index[value] = node;
 		}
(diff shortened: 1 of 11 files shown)


saqimtiaz:
@flibbles code looks good! The performance improvements are compelling.

Given that there are tests makes me more comfortable with merging this at such a late stage. 

Alternatively if we really do not want to merge with so little time remaining before 5.1.23 for user testing, we could document for prospective plugin writers that there will be changes coming to the internal implementations for filter run prefixes. (I am the only one that has written any so far or expressed any interest in doing so).

core/modules/utils/linked-list.js
```diff
@@ -0,0 +1,106 @@
...
+
+LinkedList.prototype.clear = function() {
+	this.index = Object.create(null);
+	this.next = this;
+	this.prev = this;
+	this.length = 0;
+};
+
+LinkedList.prototype.remove = function(value) {
+	if ($tw.utils.isArray(value)) {
```
saqimtiaz, line 24: in terms of syntax/code style @flibbles I believe we do not use spaces between the `if` and the `(`

flibbles: Ooh, right. I knew there was a javascript convention for TW somewhere. Couldn't find it though, so I guessed. I'll fix this.

Jermolene: That's correct

Jermolene:
Thank you @flibbles this is great stuff. Let's continue the discussion about whether to merge for v5.1.23 over at #5130

core/modules/utils/linked-list.js
```diff
@@ -0,0 +1,106 @@
...
+
+LinkedList.prototype.each = function(callback) {
+	for (var ptr = this.next; ptr !== this; ptr = ptr.next) {
+		callback(ptr.value);
+	}
+};
+
+LinkedList.prototype.toArray = function() {
+	var output = [];
+	for (var ptr = this.next; ptr !== this; ptr = ptr.next) {
```
saqimtiaz, line 98: `for(var ptr=this.next; ptr!==this; ptr=ptr.next) {`

saqimtiaz: There are a few for loops that need whitespace correction as well.

flibbles:
All right. This _should_ be good? I will feel better once there are a few tests for the `reduce` prefix. I guessed at how to adjust it for Link lists. I'm not sure how to even use it. I'm not sure what it even does.

saqimtiaz:
@flibbles I'm away from my desk but I'll push the tests tomorrow morning. In the meantime for a more informal confirmation, try the examples given here: https://github.com/Jermolene/TiddlyWiki5/pull/5196

The results should be the same using the `reduce[]` operator and the `:reduce` prefix.

pmario:
Does it improve the "default" core performance?

flibbles:
@pmario, Currently, I've only applied the link list to filters, but pretty much anywhere we're doing array manipulation could be improved by switching to the linked list. We use pushTop all over the place.

saqimtiaz:
@flibbles there may be a conflict to resolve here.

flibbles:
Conflict resolved. :)

saqimtiaz:
Thank you @flibbles 

@Jermolene shall we merge this so we can get some real world testing and feedback?

Jermolene:
Thank you everyone, this will be interesting.

flibbles:
Hey quick question for the people who helped bring in the linked lists:

Those underscore methods are meant to be private, but they're technically accessible. I don't know Tiddlywiki coding convention very well. Are they supposed to be hidden within the scope of that file, or is the underscore method understood?

saqimtiaz:
@flibbles I considered this briefly when reviewing the PR. I don't think TW has a specific convention for private methods but the intent behind underscored names are well understood to most. I would consider adding a comment in this regard at the beginning of the file.

We would need input from @Jermolene on whether that is sufficient or whether private methods should only have local scope.

Jermolene:
I think the only case where the core goes to any trouble to hide private methods is in the `$tw.Wiki` constructor in boot.js. There I wanted to make sure that developers didn't reach in and access the store hashmap directly, which was a particular risk given that was how TWC worked.

Generally, though, the core doesn't distinguish between private and public methods. It probably should do, if nothing else to simplify docs and learning (eg widget.js has many methods that are not generally and could conceivably be marked private).

pmario:
@flibbles For the tests I would have expected something, that compares the results of the linked-list with the array.push() or $tw.utils.pushTop ...

So $tw.utils.pushTop creates an array. The result of that function is used as the `list test - toEqual()`

eg: 

```
expect(list.toArray()).toEqual(<output from array function comes here>);
```

So we can be sure, that the results and behaviors are the same.

flibbles:
I'll make another PR where I tuck away the private methods more securely. It won't change any logic, so it won't possibly break anything. I'll also elaborate the tests a little to confirm that the LinkList push, pushTop, and remove behave exactly as their array counterparts.

PR #5207, opened 2020-12-05
An even faster version of the TagPickerTagTemplate

BurningTreeC:
In my tests (see https://github.com/Jermolene/TiddlyWiki5/issues/5188#issuecomment-739321726) this version of the TagPicker TagTemplate is a bit faster than the current version in the prerelease that uses the `tag-pill-inner` macrocall, even if this version uses the `wikify` widget


Jermolene:
Thanks @BurningTreeC

PR #5209, opened 2020-12-05
Tests for filter prefixes

saqimtiaz:
@Jermolene I wont have the time to finish up the docs for the `:reduce` prefix until tomorrow. I am pushing the tests I already have written so @flibbles can use them for testing his changes for #5206


flibbles:
I have tested the Linked List branch using these tests, and they all pass.

saqimtiaz:
@flibbles good to hear. I would have pushed them sooner but I wasn't at my desk.

Jermolene:
Thanks @saqimtiaz

PR #5210, opened 2020-12-05
Ensure jasmine-plugin startup module is executed in the right order

ento:
This is a follow-up to #4889

Ensures `editions/test/output/test.html` is written regardless of the order by which startup modules are initially loaded.

There is a chance that this scenario happens (although currently this doesn't seem to be the case):

- Run `bin/test.sh`
- `jasmine-plugin` startup module is executed
- There's a test failure
- Jasmine calls `process.exit(1)` (#4889 introduced this behavior)
- `--rendertiddler` command is not executed, i.e. `editions/test/output/test.html` does _not_ get written

If this ever happens, it'll be annoying but not a blocker, as it can be 'fixed' by fixing the failing test first. Also it'll only be an issue for local development, as CI currently doesn't run browser tests. 

Alternative fix: We discussed only setting `process.exitCode` by overriding the `process.exit` function (within Jasmine's context) in #4889 -- I was a little hesitant to do that, as overrides can make things harder to understand.

### DIFF
--- a/plugins/tiddlywiki/jasmine/jasmine-plugin.js
+++ b/plugins/tiddlywiki/jasmine/jasmine-plugin.js
@@ -14,6 +14,15 @@ The main module of the Jasmine test plugin for TiddlyWiki5
 
 var TEST_TIDDLER_FILTER = "[type[application/javascript]tag[$:/tags/test-spec]]";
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @ento looks good

PR #5214, opened 2020-12-06
Update GruvboxDark, CupertinoDark and Nord palettes to work better with TiddlyDesktop

BurningTreeC:


Jermolene:
Thanks @BurningTreeC

PR #5218, opened 2020-12-06
Make editTiddler in sidebarsegments/search available as variable

BurningTreeC:


saqimtiaz:
@BurningTreeC is the idea that custom search solutions can have access to this variable?

BurningTreeC:
> @BurningTreeC is the idea that custom search solutions can have access to this variable?

Yes, I'm planning to extend the documentation about custom search results

saqimtiaz:
@BurningTreeC excellent idea

Jermolene:
Thanks @BurningTreeC

Issue #5187, opened 2020-12-02
[BUG] RSOD with any call to $tw.utils.getEditionInfo in the browser


joshuafontany:
**Describe the bug**
Red-screen error "Uncaught TypeError: $tw.utils.getEditionInfo is not a function" whenever `[editions[]]` or `[editiondescriptions[]` is called (from the $:/AdvancedSearch Fitler tab, for example).

**To Reproduce**
1. Go to the $:/AdvancedSearch Filter tab.
2. Enter exactly "[editions[]]"

Results: RSOD Error above

It's weird because 
```
title: $:/core/modules/utils/edition-info.js
type: application/javascript
module-type: utils-node
```
definitely exists on disk, but `$tw.utils.getEditionInfo` does not exist in `$tw.utils` in the browser....

This occurs live on tiddlywiki.com and tiddlywiki.com/prerelease.

joshuafontany:
I just fond the documentation stating that these filter functions are not available in the web browser. In which case should be treated as unknown filters, i.e. [field:editions[]]/[field:editiondescriptions[]]?

Jermolene:
Hi @joshuafontany I guess the more consistent thing might be for those operators to return empty lists?

Jermolene:
> Hi @joshuafontany I guess the more consistent thing might be for those operators to return empty lists?

I meant to add "in the browser":


> I guess the more consistent thing might be for those operators to return empty lists in the browser?

joshuafontany:
@Jermolene Agreed & and easy fix. https://github.com/Jermolene/TiddlyWiki5/pull/5222

PR #5222, opened 2020-12-07
Edition filters will only run if $tw.node, else return empty set

joshuafontany:
Bug https://github.com/Jermolene/TiddlyWiki5/issues/5187


Jermolene:
Thanks @joshuafontany

PR #5223, opened 2020-12-07
Logging widgets :Fixed issue on Firefox where logging does not work

### LINKED ISSUE #5219: [BUG] $log widget producing empty output in Firefox
**Describe the bug**
The `$log` widget appears to be producing blank output in the Firefox console: the values of the logged variables aren't showing up.

**To Reproduce**
I created a tiddler named `Something`, and two related tiddlers called `Something/one` and `Something/two` containing the text `One` and `Two` respectively. (In my real project these would be subcategories of data that I want to show up in the main tiddler). Each of the "one" and "two" tiddlers had a `list` field containing tiddler titles, to simulate my use case.

The content of the `Something` tiddler looked like this:

```plaintext
Something goes here

<$vars foo={{{ [<currentTiddler>addsuffix[/one]get[list]then[ListTemplate]else[nonexistent]] }}}>
<<foo>>
<$log foo currentTiddler/>
<$tiddler tiddler={{{ [<currentTiddler>addsuffix[/one]] }}} >
<$transclude tiddler=<<foo>>/>
</$tiddler>
</$vars>
```

The `<<foo>>` line printed out `ListTemplate` as I expected, and the transclusion looked the way it should (I was doing the equivalent of `{{Something/one ||ListTemplate}}`). But the `<$log foo currentTiddler/>` line just underneath the `<<foo>>` line produced a table that showed rows for `foo` and `currentTiddler`, but the rows were empty. See screenshot below.

A simpler repro would be to just create a tiddler named `ABC` with the following contents:

```
<$log currentTiddler/>
<<currentTiddler>>
```

When I do this, I can see that the `currentTiddler` variable does indeed contain ABC as it should, but the console log is showing an empty value for currentTiddler.

**Expected behavior**
The console log contains actual values.

**Screenshots**
image

 - OS: Linux (Ubuntu 18.04)
 - Browser: Firefox
 - Version: 81.0

**Additional context**
I don't have time right now to test this under Chrome or other browsers. Once I find the time, I'll report back with those findings.

saqimtiaz:
Fixes #5219 

In Firefox variables were not being logged properly. It turns out that the second argument (column names) to console.logTable is only supported for arrays and not objects. In Chrome the second argument is ignored for objects but in Firefox it causes the data not be logged properly.

I've tested this on Chrome and Firefox on Windows 10 and Linux. 

@Jermolene please confirm on Safari/OSX. I assume the original implementation was failing gracefully for you as it does for webkit based browsers in Windows 10 and Linux.

@rmunn Confirmation that this resolves the problem would be great.

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -56,7 +56,7 @@ exports.warning = function(text) {
 /*
 Log a table of name: value pairs
 */
-exports.logTable = function(data,columnNames) {
+exports.logTable = function(data) {
 	if(console.table) {
 		console.table(data);
 	} else {
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -58,7 +58,7 @@ Log a table of name: value pairs
 */
 exports.logTable = function(data,columnNames) {
 	if(console.table) {
-		console.table(data,columnNames);
+		console.table(data);
 	} else {
 		$tw.utils.each(data,function(value,name) {
 			console.log(name + ": " + value);
--- a/core/modules/widgets/action-log.js
+++ b/core/modules/widgets/action-log.js
@@ -78,7 +78,7 @@ LogWidget.prototype.log = function() {
 
 	console.group(this.message);
 	if(dataCount > 0) {
-		$tw.utils.logTable(data,["name","value"]);
+		$tw.utils.logTable(data);
 	}
 	if(this.logAll || !dataCount) {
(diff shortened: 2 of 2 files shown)


rmunn:
Confirmed, this fixes #5219 for me.

Jermolene:
Thanks @saqimtiaz @rmunn

Jermolene:
I can confirm that this works in Safari and Firefox on OS X

PR #5224, opened 2020-12-07
fix viewtemplate quirks

joshuafontany:
@Jermolene  This fixes 2 quirks with the view template that I ran into while experimenting with my "custom view template injections" branch.

 - Add a $reveal to the Plugin section of the view template to match the rest of the view template sections.
 - Ads `mode="block" to the $:/core/ui/TiddlerInfo/Advanced tiddler, so that if other tiddlers are tagged "$:/tags/TiddlerInfo/Advanced" do not appear inline alongside the Advanced/ShadowInfo or Advanced/PluginInfo transclusions.


Jermolene:
Thanks @joshuafontany

PR #5225, opened 2020-12-07
Add documentation for how to add customized search-results using the new search mechanism

BurningTreeC:
This PR adds documentation about how to add customized search-results using the newly introduced search mechanism


Jermolene:
Thanks @BurningTreeC

PR #5230, opened 2020-12-07
Make improvement banner foreground color follow palette colours

BurningTreeC:
This PR makes the contribution-banner text foreground color follow the palette colours using the `<<contrastcolour>>` macro

### DIFF
--- a/editions/tw5.com/tiddlers/system/tw5.com-styles.tid
+++ b/editions/tw5.com/tiddlers/system/tw5.com-styles.tid
@@ -4,6 +4,21 @@ tags: $:/tags/Stylesheet
 title: $:/_tw5.com-styles
 type: text/vnd.tiddlywiki
 
+\define set-improvement-banner-color-css(colour,colourA,colourB)
+<$set name="color" value=<<contrastcolour target:"""$colour$""" fallbackTarget:"""""" colourA:"""$colourA$""" colourB:"""$colourB$""">>>
+.tc-improvement-banner {
+	color: <<color>>;
+	fill: <<color>>;
+}
+</$set>
+\end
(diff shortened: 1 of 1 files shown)


Jermolene:
Bravo @BurningTreeC. Perhaps this might be more elegant if make "contrastcolour" into one of the new-fangled multi-operand filter operators?

BurningTreeC:
>Bravo @BurningTreeC. Perhaps this might be more elegant if make "contrastcolour" into one of the new-fangled multi-operand filter operators?

Thanks @Jermolene, but I don't understand the second part :grin: 

What do you intend, what should we do?

Jermolene:
Hi @BurningTreeC apologies. It's just that here as in the tabs macro it's actually quite awkward to marshal the parameters to call the contrastcolour macro, and requires textual substitution. I was wondering if we could improve things by making a new filter operator version of contrastcolour so that one could just do `color: {{{ [[#fcc]contrastcolour[foreground][background]] }}}`, but on reflection the problem here is really that colours need to be resolved with wikification because of the way we nest the colour macro.

BurningTreeC:
>but on reflection the problem here is really that colours need to be resolved with wikification because of the way we nest the colour macro.

yes, that's a problem

Jermolene:
I have wondered in the past if we might make a special lightweight parser for colour definitions. It's expensive to spin up the parser and rendering process when in most cases we only need to be able to execute a single macro.

Conversely, we might also want to explore possibilities from using other macros within palette entries. For example, it might be nice if we could use the contrastcolour macro directly in a palette entry.

Another thought: it's a shame we didn't make palette entries be filters instead of wikitext.

PR #5232, opened 2020-12-07
Make search-matches button discoverable on dark palettes

BurningTreeC:
This PR makes `.tc-sidebar-lists button small` the foreground-color of the palette. Like that, the search-matches button becomes visible on dark  palettes, too


Jermolene:
Thanks @BurningTreeC

PR #5237, opened 2020-12-08
Rename $:/language/Exporters/TidFile

BurningTreeC:
from `".tid" file` to `".tid" file (the first match only)` ... so that it's clearer that only one file is exported


Jermolene:
Thanks @BurningTreeC, that got me thinking, we can go better and hide exporters that can't cope with the number of tiddlers being exported.  I'll commit momentarily.

PR #5238, opened 2020-12-08
Docs for :reduce filter run prefix

saqimtiaz:


saqimtiaz:
@Jermolene would be good to add these docs before the release.

Jermolene:
Thanks @saqimtiaz

PR #5239, opened 2020-12-08
Docs update for Set Widget

saqimtiaz:
@Jermolene This updates the docs for **Set Widget** to add a tip regarding a common source of misunderstanding for users, i.e. using _Filtered List Variable Assignment_ and being confused as to why the value of the variable is in double square brackets. This is compounded by the oft given erroneous advice on the google group, to then wikify the value to remove the double square brackets.


Jermolene:
Thanks @saqimtiaz

PR #5242, opened 2020-12-09
Rename control-panel-button to advanced-search-button in advanced-search button

BurningTreeC:
This PR renames the macrocall for the advanced-search button so that it is semantically correct


Jermolene:
Thanks @BurningTreeC

PR #5244, opened 2020-12-09
cursor:pointer on input[type=file] doesn't work in webkit browsers...

BurningTreeC:
...

```
::-webkit-file-upload-button {
	cursor:pointer;
}
```

does


Jermolene:
Thanks @BurningTreeC

PR #5251, opened 2020-12-10
Fix for LinkedLists when using bad filter prefix

flibbles:
Quick fix for bug introduced by Linked Lists when the filter prefix used doesn't exist.


Jermolene:
Thanks @flibbles

Issue #5166, opened 2020-12-01
[IDEA]  :sort prefix and extensible filter run prefixes


saqimtiaz:
Just as the `:filter` filter run prefix provides an inline alternative to the `filter` operator, it would be desirable to have a `:sort` filter run prefix where the list is sorted by comparing the results of applying the filter run to each list item. This would be an inline alternative to the `sortsub` operator.

`[tag[HelloThere]] :sort[get[text]length[]] `

instead of:

```
<$vars compare-by-title-length="[length[]]">
[tag[HelloThere]sortsub:number<compare-by-title-length>]
</$vars>
```

However, we currently support multiple types of sorting: `string`, `integer`, `number`, `date` and `version`

So either we would need to add 5 filter run prefixes, one for each of the sort types. `:sortstring`, `:sortinteger` etc...

Or we could consider supporting suffixes for filter runs. Example : `[tag[HelloThere]] :sort:string[........]` or `[tag[HelloThere]] :sort:number[........]`
While this second approach is a lot more flexible and powerful, the syntax might prove not so intuitive for new users.

Is there a third alternative that I am not thinking of?

I am interested in implementing this - even if only as a plugin - so I would appreciate any feedback or ideas.

rmunn:
Not only that, there's also `sortan` for alphanumeric sorting, there's reverse sorting (high to low instead of low to high)... it could get very complicated very fast.

I strongly dislike the idea of having suffixes for filter run prefixes (`:sort:number` and so on). Definite comprehension issues for new users. If there's enough benefit to add a `:sort` prefix, then it's worth adding all of the various sorting variants as their own prefix (`:sortn`, `:sortan`, `:revsort` and so on). But I don't think there is enough benefit. I think `<$vars sort-by="[some-filter[]]">` followed by `sortsub<some-filter>` is simple enough that the prefix version doesn't gain all that much, and when you take into account the dozen different sorting prefixes we would need to account for all possibilities, I think the complexity cost is more than what we'd gain.

saqimtiaz:
@rmunn the feeling is mutual about suffixes for filter runs and comprehension. The sort options I listed are the ones supported by the `sortsub` operator rather than the `sort` operator.

As for the utility of the sort prefix, I find filters written inline **far easier to read, comprehend and maintain**. 

> when you take into account the dozen different sorting prefixes we would need to account for all possibilities, I think the complexity cost is more than what we'd gain.

[...]

PR #5252, opened 2020-12-10
Allow suffixes for filter runs

saqimtiaz:
Allows for suffixes to filter run prefixes, as discussed in #5166
Also see [here](https://github.com/Jermolene/TiddlyWiki5/issues/5166#issuecomment-742656857)

@Jermolene Suggest we delay this for 5.1.24 since no core prefixes use this at the moment, and thus we can have more time to make a considered decision and test more thoroughly. Just pushing this PR now so I do not forget.


saqimtiaz:
An open question is whether we would want to support multiple suffixes?

saqimtiaz:
@Jermolene I thought as much and it doesn't hurt to have the support there even if we don't necessarily use it right away.

I will re-work this PR to support multiple suffixes post 5.1.23

rmunn:
One core named prefix, `:reduce`, might be able to benefit from suffixes so that users can specify the starting value for the accumulator (e.g., the `:reduce:1[multiply<accumulator>]` example). Though for `:reduce`, what it really wants is a second parameter the way the `reduce` operator takes a second parameter, because that would allow specifying the starting value from a variable or a transclusion. A suffix would only allow literal values to be specified, which helps when you want a starting accumulator value of 1 but doesn't help much when you want a starting value of `<currentTiddler>`.

But that's not a reason to squeeze this into 5.1.23. I agree with delaying it until 5.1.24.

saqimtiaz:
@Jermolene I have extended this to accommodate multiple filter run prefix suffixes as previously discussed. As a recap after [some discussion](https://github.com/Jermolene/TiddlyWiki5/issues/5166#issuecomment-742656857) we had agreed to implement suffixes for filter runs.

`:sort:suffix1:suffix2[... filter ...]` or `:sort:suffix1a,suffix1b:suffix2[... filter ...]` are now possible.

- The syntax supported is exactly the same as for the rich suffixes for filter operators.
- I've added test for named filter run prefixes, without and with suffixes
- Once this is merged I'll extend `:reduce` to use a suffix and also work on the `:sort` filter run prefixes. 
- Once we have examples in the core that use these suffixes I will update the documentation.

saqimtiaz:
With regards to the alternate proposal for a syntax like `:sort[... filter ...],[suffix] `, I think this is far less easier to read and comprehend. Given that filters can get rather long, the suffixes for a filter run prefix can end up really far away from their prefix, and you don't understand what the filter run is doing until after you get to the end. 

Compare the following (please disregard the logic of the filter, its just an example of a longer filter run):
`:subsort[get[myfield]enlist-input[]removeprefix[$:/sq/plugins/sort-criteria]then<1>else<2>],[number]`

`:subsort:number[get[myfield]enlist-input[]removeprefix[$:/sq/plugins/sort-criteria]then<1>else<2>]`

With the latter, it is immediately understood how the filter run will be applied, whereas with the former you don't realize this until after reading the entire run. While the former syntax might seem familiar to the developers amongst us as it resembles passing arguments to a function, I think for most of our users the latter will be far easier to understand.

As for being able to use variables as suffixes, currently our parser doesn't support either of the above two syntax. So support for variables isn't a given in either one, and could be implemented for either if so desired.

Also, I think it would be acceptable for each way of filtering to have its own strengths. For example comparing the `subsort[]` filter operator to a `:subsort` filter run prefix, the operator needs a variable for the filter, whereas the filter run prefix does not but cannot pass a variable to the suffix. Users can choose whichever is most convenient for their use case.

Jermolene:
Sounds good thanks @saqimtiaz, I'll have a look at the code.

Jermolene:
Fantastic, thanks @saqimtiaz

PR #5253, opened 2020-12-10
Is there a reason for the <$browse> widget not being within the <$button> scope?

BurningTreeC:
This PR moves the `<$browse>` widget from outside the button to within the button, which makes hover effects on the button work. The file input is triggered nevertheless in my tests (chrome,firefox and IE) ... is there a special reason that it's not been done like that?


Jermolene:
Hi @BurningTreeC the way we style the file input button dates back to the early days of TW5 in 2013, when the state of the art for styling file inputs was to use CSS to overlay a transparent file input over a properly styled button. There may be a newer approaches we could follow now that older IE is gone, but we would need to research further because I wouldn't want to unwittingly break any browser that currently works.

PR #5254, opened 2020-12-10
Fix browser testing

### LINKED ISSUE #5243: [BUG] Is anybody else noticing that jasmine doesn't work anymore on browser?
Jasmine still works with the core testing framework in `editions`, but ever since the fix for #5210 (Dec 4th) came in, jasmine no longer runs in the browser for any of my projects, including Relink.

I'm still looking into this, but I'm coming into this cold. Does anyone know why the single line: `exports.after = ["commands"];` would cause the tests to stop initializing?

ento:
Fixes #5243 by specifying the `exports.after = ["commands"]` condition only in Node.js.

image


Jermolene:
Thanks @ento

PR #5255, opened 2020-12-11
The reduce operator and :reduce prefix should handle empty input identically

### LINKED ISSUE #5246: [BUG] 5.1.23 Blocker: The reduce operator and :reduce run prefix behave differently on empty input
**5.1.23 blocker**, IMHO, though it shouldn't take too long to resolve.

**Describe the bug**
The `reduce` operator and `:reduce` prefix behave differently on empty input. One or the other should be adjusted so they behave the same. The operator will always produce output, even if its input is empty (the output is the value of the accumulator). The filter run prefix will produce an empty output if its input is empty, and will only produce output if there was at least one input item. This is because of the `if(results.length > 0)` check in the filter run prefix version, which is lacking in the operator version.

**To Reproduce**
1. Run the "nonexistent tag" example for the `reduce` operator
2. Run the "nonexistent tag" example for the `:reduce` filter run prefix
3. Note the different behavior with regard to the `else` operator (or `:else` prefix)

**Expected behavior**
Both the operator and the filter run prefix should have exactly the same behavior with regard to empty input. There are four ways I can think of to make this happen:
1. Make the operator return empty output when its input is empty, and don't change the way the filter run prefix acts.
1. Make the filter run prefix return the accumulator's initial value when its input is empty.
1. Make the filter run prefix return the accumulator's initial value when its input is empty, **and** additionally provide a way for filter run prefixes to take an extra argument the way operators do.
1. Entirely change the behavior of both `reduce` and `:reduce`: if no initial accumulator value given, use first list item for initial value. Use accumulator value only if given. (On empty input, return empty output, just like option 1).

Of those four, I strongly prefer the first one. I had to write extra documentation for `reduce` to make it plain that it will always produce output even when its input is empty, and although my reasons for that were based on the fact that that's how it works in functional programming, I'm re-evaluating that opinion. The fact that I had to write extra documentation just goes to show that it's non-intuitive for most people that empty input would produce non-empty output, and the right way for the `reduce` operator would have been for it to produce empty output when its input is empty, because that's what other operators do and that's what most people will expect. Thankfully, 5.1.23 isn't released yet so it's not too late to change this.

Option two is possible, but means that the `:reduce` prefix can never return, say, 0 when asked to sum up an empty list of numbers. And the current `:reduce` example (which uses `:else[[0]]` to specify a default value for empty input) would then be impossible.

Option three is a major syntax change, which I don't like so late in the release cycle. However, there's one good point in its favor. What would you expect to happen when you write `1 2 3 :reduce[multiply<accumulator>]`? You'd expect to get 6, right? But instead you get 0, because the initial value of the accumulator is the empty string, and math operators treat the empty string as 0. To get what you actually expect from `:reduce[multiply<accumulator>]`, you have to specify an initial accumulator value of 1. That's possible with the operator, but impossible with the filter run prefix!

Option four is not a major syntax change, but is a bit of a behavior change. But if we're going to do it, now's the time. Basically, right now `reduce` acts like the functional programming function called `fold`, *not* like the function called `reduce`! The rules for `reduce` in functional programming are "first item in list is initial accumulator value, and it's an error to pass in an empty list". The function where you specify an initial accumulator value *and* an input list, and the return value on empty input is the initial accumulator value, that one is called `fold` in functional programming. I do **not** suggest creating a `fold` operator, as most people will have no idea what it does! But I do think that it would be good for the `reduce` operator to be able to actually act like `reduce`, so that the `multiply<accumulator>` result will work the way people expect.

So I suggest that either we make the small change and let the `reduce` operator also return empty input (option one), or we take option four and let the `reduce` operator have two different behaviors, one for when no initial accumulator value is passed, and one for when an initial accumulator value is provided. Whichever one we do, we'd better do it soon, because right now there's an inconsistency that we don't want to allow into the release.

rmunn:
Now the `reduce` operator and `:reduce` filter run prefix will both return empty output when their input is empty, so that both can be chained together with the `else` operator or `:else` prefix.

Fixes #5246.

### DIFF
--- a/core/modules/filters/reduce.js
+++ b/core/modules/filters/reduce.js
@@ -48,7 +48,11 @@ exports.reduce = function(source,operator,options) {
 			accumulator = "" +  list[0];
 		}
 	}
-	return [accumulator];
+	if(results.length > 0) {
+		return [accumulator];
+	} else {
+		return [];
+	}
 };
 
 })();
--- a/editions/test/tiddlers/tests/test-prefixes-filter.js
+++ b/editions/test/tiddlers/tests/test-prefixes-filter.js
@@ -77,6 +77,8 @@ describe("'reduce' and 'intersection' filter prefix tests", function() {
 		expect(wiki.filterTiddlers("[tag[shopping]] :reduce[get[quantity]add<accumulator>]").join(",")).toBe("22");
 		expect(wiki.filterTiddlers("[tag[shopping]] :reduce[get[price]multiply{!!quantity}add<accumulator>]").join(",")).toBe("27.75");
 		expect(wiki.filterTiddlers("[tag[shopping]] :reduce[<index>compare:number:gt[0]then<accumulator>addsuffix[, ]addsuffix<currentTiddler>else<currentTiddler>]").join(",")).toBe("Brownies, Chick Peas, Milk, Rice Pudding");
+		// Empty input should become empty output
+		expect(wiki.filterTiddlers("[tag[non-existent]] :reduce[get[price]multiply{!!quantity}add<accumulator>]").length).toBe(0);
 		expect(wiki.filterTiddlers("[tag[non-existent]] :reduce[get[price]multiply{!!quantity}add<accumulator>] :else[[0]]").join(",")).toBe("0");
 	});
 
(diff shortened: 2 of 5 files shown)


Jermolene:
Thank you @rmunn

PR #5260, opened 2020-12-11
URGENT: Reveal widget - fixed problem with previous patch of refresh handling

saqimtiaz:
@Jermolene Please merge before 5.1.23!

Apologies, my previous patch was in haste and not well thought through. It did not properly account for the state tiddler changing in addition to the style and/or class. I have now tested this quite thoroughly.

It now follows the original logic of `refresh()`, but if we do not need to call `refreshSelf()` it also checks at the end whether the `style` and `class` need updating.


Jermolene:
Thanks @saqimtiaz

PR #5263, opened 2020-12-12
tabsState should hold a tiddler name rather than a list singleton

xcazin:
Before this fix, state tiddlers were created by the tab macro with names like `[[$:/state/my tab state--1845137525]]` instead of `$:/state/my tab state--1845137525`.

Not sure that this syntax is better than `<$set name="tabsState" filter="[<__explicitState__>minlength[1]] ~[<qualifiedState>]" select="0">` though. It just seemed more "modern" to me.

### DIFF
--- a/core/wiki/macros/tabs.tid
+++ b/core/wiki/macros/tabs.tid
@@ -3,7 +3,7 @@ tags: $:/tags/Macro
 
 \define tabs(tabsList,default,state:"$:/state/tab",class,template,buttonTemplate,retain,actions,explicitState)
 <$set name="qualifiedState" value=<<qualify "$state$">>>
-<$set name="tabsState" value={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@Jermolene we should resolve this before 5.1.23

@xcazin I would favour the vars widget in this situation:
`<$vars tabsState={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>`

xcazin:
> @xcazin I would favour the vars widget in this situation:
> `<$vars tabsState={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>`

Thanks @saqimtiaz. Pushed a new commit with the above syntax.

saqimtiaz:
Thank you @xcazin 
Looks good

saqimtiaz:
@Jermolene @BurningTreeC  just bumping this to make sure it doesn't slip through the cracks.

Jermolene:
Great, thanks @xcazin @saqimtiaz

PR #5267, opened 2020-12-12
Add a hook to allow modifying the the behaviour creating tiddler paths

inmysocks:
This is needed for Bob to use the core to generate tiddler fileInfo 

I don't know if this is the best way to make the hook, but it works for what I need

The hook can return a new path, which is used if it exists, otherwise it uses the new path with the uri encoding if that exists, otherwise it uses the default path

### DIFF
--- a/core/modules/utils/filesystem.js
+++ b/core/modules/utils/filesystem.js
@@ -382,13 +382,15 @@ exports.generateTiddlerFilepath = function(title,options) {
 		count++;
 	} while(fs.existsSync(fullPath));
 	//If the path does not start with the wikiPath directory or the wikiTiddlersPath directory, or if the last write failed
-	var encode = !(fullPath.indexOf(path.resolve($tw.boot.wikiPath)) == 0 || fullPath.indexOf($tw.boot.wikiTiddlersPath) == 0) || ((options.fileInfo || {writeError: false}).writeError == true);
+	var newPath;
+	var encode = !(fullPath.indexOf($tw.boot.wikiPath) == 0 || fullPath.indexOf($tw.boot.wikiTiddlersPath) == 0) || ((options.fileInfo || {writeError: false}).writeError == true);
 	if(encode){
 		//encodeURIComponent() and then resolve to tiddler directory
-		fullPath = path.resolve(directory, encodeURIComponent(fullPath));
+		newPath = path.resolve(directory, encodeURIComponent(fullPath));
 	}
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @inmysocks thank you, and apologies for the delay. There is a subtlety with the hook mechanism around how it handles multiple handler functions registered against the same hook. In order to allow multiple hooks to update the same data, the hook invocation code passes the return value of the previous hook function to the subsequent one. So, in this case, I think you should be passing the value in the first parameter. This was poorly documented so I've updated things here: 81947edd5c3a2daa8da44a520824b1867e40953e

Also, it would be appreciated if you could add a docs tiddler in the dev wiki along these lines: https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/dev/tiddlers/new/th-saving-tiddler.tid

Jermolene:
Thanks @inmysocks

Jermolene:
Hi @inmysocks I might have jumped the gun and merged before you had a chance to commit the docs, sorry you'll have to create a new PR.

inmysocks:
ok, I will make a new PR

joshuafontany:
After reading through @inmysocks 's `MultiWikiAdaptor` code, I think all I am missing to return the correct filepath is to test against a `path.resolve(directory)`, because that is passed in as the root & bob has already ensured that it exists. So something like:

```
	// If the last write failed with an error, or if path does not start with:
	//	the resolved options.directory, the resolved wikiPath directory, or the wikiTiddlersPath directory, 
	//	then encodeURIComponent() and resolve to tiddler directory
	var encode = (options.fileInfo || {writeError: false}).writeError == true;
	if(!encode){
		encode = !(fullPath.indexOf(path.resolve(directory)) == 0 ||
			fullPath.indexOf(path.resolve($tw.boot.wikiPath)) == 0 ||
			fullPath.indexOf($tw.boot.wikiTiddlersPath) == 0);
		}
	if(encode){
		fullPath = path.resolve(directory, encodeURIComponent(fullPath));
	}
	// Return the full path to the file
	return fullPath;
```

At this point, Bob is only creating these fileInfo objects for tiddlers that are just created (do not already have one), and passing the directory in with the function call. I think this will resolve Bob's filepath generation & I'd like to work with Jed so that Bob also calls `generateTiddlerFileInfo()` every time a tiddler is saved (to match the core and use the new FileSystemExtention filters as well).

Thoughts?

inmysocks:
I think that is a good solution. I feel a bit slow because I just understood what you meant when you mentioned a security concern before and that is going to be of particular concern for what I am making with the secure server.

PR #5273, opened 2020-12-13
Remove menubar from print

bimlas:
In the print view, a shadow appeared at the top of each page, and the first lines were not visible.

image

---

I hide the menubar in the print view.

image

### DIFF
--- a/plugins/tiddlywiki/menubar/styles.tid
+++ b/plugins/tiddlywiki/menubar/styles.tid
@@ -213,3 +213,10 @@ nav.tc-menubar .tc-more-sidebar > .tc-tab-set > .tc-tab-buttons > button {
 	<<set-sidebar-scrollable-top>>
 
 }
+@media print {
+
+	nav.tc-menubar {
+		display: none;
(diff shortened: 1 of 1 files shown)


bimlas:
If it will be merged, you will probably need to increase the version number of the plugin, I did not do that.

Jermolene:
Thanks @bimlas.

> If it will be merged, you will probably need to increase the version number of the plugin, I did not do that.

The core plugins don't carry a version number in their `plugin.info` file. Instead, the plugin loading process uses the core version number for any plugins that don't carry a version field.

https://github.com/Jermolene/TiddlyWiki5/blob/0911d99813f9123961116a08c136b157a72393bd/boot/boot.js#L1974-L1977

This was done in the early days but I think there's now a case for switching to independent versioning for each plugin. The challenge we face there is that we can't go back to a numerically earlier version number than the core without disrupting the upgrade process. So perhaps we'd bump all the plugins to v10.0.0, and then start versioning them in the usual way from there.

PR #5278, opened 2020-12-14
Include named prefixes in railroad diagram for Filter Expression tiddler

### LINKED ISSUE #5277: [BUG] Railroad diagram for Filter Expression tiddler is out-of-date
**Describe the bug**
The railroad diagram shown in the Filter Expression tiddler currently shows the prefixes `+`, `~`, `-`, and `=`, but does not show any named prefixes. This doesn't have to block the 5.1.23 release as the documentation could be updated post release, but it's probably a good idea to update the railroad diagram before the release.

**Expected behavior**
I suggest adding a single entry marked "named prefix" with a colon in front to the diagram, rather than adding all the valid named prefixes in as separate possible choices. Otherwise the sheer number of valid named prefixes will make the list unwieldy, and the fact that named prefixes are extensible by plugins will make it likely that the diagram would miss a valid prefix anyway.

rmunn:
Fixes #5277 and tries to be future-proof by building the list of valid named prefixes dynamically.

I had a little trouble deciding how to build the railroad diagram that lists all the named prefixes, because their actual names are determined by what the module puts into its exports (e.g., in `all.js` you get `exports.all = function(...)` and so that named prefix is called `all`). I could not find a way inside TW to say "What does this module export?", so in the end I decided to go with the name of the module as the name I put into the railroad diagram. This means that if someone in the future creates a named prefix called `foo` in a file called `bar.js`, it will list as `bar` in the diagram I wrote, when it should list as `foo`.

Also, it might be better to build the list by first trimming a `.js` suffix, and then using the `splitregexp` operator to split on non-word characters, then grab the last item of the resulting list and use that. That way if a user creates a module called "MyPrefixes/foo.js", it would be shown as "foo" in the resulting diagram instead of as "MyPrefixes/foo". I plan to add to this PR later on to improve that, but I'm running out of time for today so I'm submitting this as-is. It will at least be good for documenting the current valid prefixes in the core.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/syntax/Filter Expression.tid
+++ b/editions/tw5.com/tiddlers/filters/syntax/Filter Expression.tid
@@ -1,5 +1,5 @@
 created: 20150124182421000
-modified: 20201208185257875
+modified: 20201214053032397
 tags: [[Filter Syntax]]
 title: Filter Expression
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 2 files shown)


Jermolene:
Great thank you @rmunn

PR #5280, opened 2020-12-14
Fix refresh for all attributes of droppable widget. Add disabled class

saqimtiaz:
This PR fixes the refresh handling of the `DroppableWidget` so that the widget is also refreshed when the `action` or `effect` attributes change.

It also adds a `disabledClass` that is applied to the generated DOM node when drag and drop is disabled.


Jermolene:
Thanks @saqimtiaz

PR #5284, opened 2020-12-15
add missing tags for file-backups plugin info

pmario:
Just a little docs change: add missing tags for file-backups plugin info

@Jermolene ... We should really add some info about the "color codes" at: https://tiddlywiki.com/prerelease/#Saving ... It's super confusing even for me :/


Jermolene:
Thanks @pmario. This can actually go to the "tiddlywiki-com" branch and be published immediately, if you wouldn't mind changing the target?

PR #5287, opened 2020-12-15
add pragmacomment wikirule

pmario:
This PR allows us to use comments outside of macro definitions, which isn't possible with html-comments `<!-- comment -->`

You can [play with it at a demo site](https://wikilabs.github.io/editions/custom-markup/#Pragma%20Comment:%5B%5BPragma%20Comment%5D%5D%20%5B%5BDraft%20of%20'New%20Tiddler'%5D%5D)

Since the parse-tree is cached, those elements only need to be handled once. 

```
\\ comment comes here till the end of the line
\\ comments like this are fast since they are ignored and the result is cached!

\define test()
\\ This comment is as slow as the next line
<!-- HTML comment -->
\end

\\ The whole definition is deactivated. .... Works well for a few lines of code!
\\define test()
\\end

\\ if you have big junks of code, it's easier to rename the macro eg: testXXXX() as shown below!
\define testXXXX()
x
x
x
x
\end
```

### DIFF
--- a/core/modules/parsers/wikiparser/rules/pragmacomment.js
+++ b/core/modules/parsers/wikiparser/rules/pragmacomment.js
@@ -1,0 +1,59 @@
+/*\
+title: $:/core/modules/parsers/wikiparser/rules/pragmacomment.js
+type: application/javascript
+module-type: wikirule
+
+Wiki pragma rule for pragma comment specifications
+
+```
+\\ this is comment in pragma "space"
+\\ fast comment, since it doesn't produce a parse-tree element
+
+\\define xx()
+\\ slow comment because it WILL produce parse-tree elements
+\\end
+```
+
+\*/
+(function(){
+    /*jslint node: true, browser: true */
+    /*global $tw:false, exports:false*/
+    "use strict";
+    
+    exports.name = "pragmacomment";
+    exports.types = {
+        pragma: true
+    };
+    
+    /*
+    Instantiate parse rule
+    */
+    exports.init = function (parser) {
+        var self = this;
+        this.parser = parser;
+        // Regexp to match
+        this.matchRegExp = /^\\\\/mg;
+    };
+    
+    /*
+    Parse the most recent match
+    */
+    exports.parse = function () {
+        // Move past the pragma invocation
+        this.parser.pos = this.matchRegExp.lastIndex;
+        // Parse line terminated by a line break
+        var reMatch = /([^\n]*\S)|(\r?\n)/mg;
+    
+        reMatch.lastIndex = this.parser.pos;
+        var match = reMatch.exec(this.parser.source);
+        while (match && match.index === this.parser.pos) {
+            this.parser.pos = reMatch.lastIndex;
+            // Exit if we've got the line break
+            if (match[2]) {
+                break;
+            }
+        }
+        return [];
+    };
+    })();
+    


saqimtiaz:
This may not be entirely backwards compatible. 

I'd like for us to not merge this until after 5.1.23 so we can test this properly, including potential unit tests with content where the first line  after the pragma starts with `//`

Jermolene:
Hi @pmario I've also wondered about exactly this syntax for comments. My only concern is whether `\\` might already be in use in the wild as an ordinary part of text. However, that's a concern with almost all useful proposals to extend the syntax.  It's not sustainable to be prevented from updating the wikitext syntax. I'm hoping that something like #345 might help us maintain backwards compatibility.

rmunn:
I'll just comment (heh) to say that as a developer who does a lot of coding in languages that use `//` for comments, `\\` grates on me *hard* and I don't like it. Purely for aesthetic reasons — nothing wrong with it in principle, but as I was looking at your example I kept thinking "Ooh, that's totally wrong! He got backslashes and slashes mixed up!" Even though I knew that you were using `\\` on purpose, I still had that running through my head.

So while I kind of like the idea of having wiki syntax for a single-line comment, I'm a strong :-1: on the `\\` syntax specifically.

Jermolene:
> So while I kind of like the idea of having wiki syntax for a single-line comment, I'm a strong :-1: on the `\\` syntax specifically.

The `\\` syntax is because this is implemented as a pragmarule, all of which start with `\`. However, pragmarules can only be used at the start of a tiddler, and can't be used after ordinary tiddler content. I suspect that that will trip people up too. We may be able to get around it by overloading the rule so that it can behave both as a pragma and an ordinary block mode rule.

As I mentioned above, there's a lot of Groundhog Day moments when we consider adding new wikitext syntax; we are apt to get locked into concerns about whether a proposed rule is universal enough to warrant going into the core, and lots of discussion about the best syntax to use. Given the paucity of easily typed symbols on most keyboard layouts, we're fighting over diminishing resources.

pmario:
@rmunn 
>So while I kind of like the idea of having wiki syntax for a single-line comment, I'm a strong 👎 on the \\ syntax specifically.

As Jeremy pointed out, pragmas have to use a backslash. I could have used:

```
\comment .... but I think that's nonsense
\\ ... as a shortcut is much easier to type and remember. 
```

pmario:
@saqimtiaz 

> I'd like for us to not merge this until after 5.1.23 so we can test this properly, including potential unit tests with content where the first line after the pragma starts with \\

That's easy. Just test this

```
\\ comment
\define x() xxxx

\\ ADD a dot in front of this line and have a look at the "parse tree" preview
<<x>>
```

pmario:
> ... The \\ syntax is because this is implemented as a pragmarule, all of which start with \. However, pragmarules can only be used at the start of a tiddler, and can't be used after ordinary tiddler content.

From my point of view that's an advantage. We don't have a compatibility problem, if a users wants to use it for wikitext! They are free to do so.

pmario:
>..As I mentioned above, there's a lot of Groundhog Day moments when we consider adding new wikitext syntax; we are apt to get locked into concerns about whether a proposed rule is universal enough to warrant going into the core, and lots of discussion about the best syntax to use. Given the paucity of easily typed symbols on most keyboard layouts, we're fighting over diminishing resources.

That's true. The linked project in the OP is a very good example for the problems we face, if we want to add new wikitext "markers". In the discussion about "custom markup" we name those markers "glyphs". ... 

It's absolutely hard to find them. They should be easy to remember, type, read in plain text and be part of the default fonts in different OSes. ...

pmario:
I'm inclined to close the PR and publish the function as a "stand alone" plugin, instead of being part of a bigger plugin.

So users can decide, if they want to use it.

Jermolene:
> That's true. The linked project in the OP is a very good example for the problems we face, if we want to add new wikitext "markers". In the discussion about "custom markup" we name those markers "glyphs". ...

(Parenthetically, I don't think it's a good term to use because it already has a very clear established meaning in relation to text processing; perhaps best not to start using it in this way in core discussions).

> I'm inclined to close the PR and publish the function as a "stand alone" plugin, instead of being part of a bigger plugin.

Perhaps the takeaway for the core team is the broader question of how to cleanly open up the ability to extend wikitext syntax while striking a balance with backwards compatibility.

saqimtiaz:
> I'm inclined to close the PR and publish the function as a "stand alone" plugin, instead of being part of a bigger plugin.

I would suggest leaving the PR open and also publishing the plugin. The learnings from the plugin and its usage will help inform this PR and the discussions around it.

rmunn:
> > So while I kind of like the idea of having wiki syntax for a single-line comment, I'm a strong -1 on the `\\` syntax specifically.
> 
> The `\\` syntax is because this is implemented as a pragmarule, all of which start with `\`. However, pragmarules can only be used at the start of a tiddler, and can't be used after ordinary tiddler content. I suspect that that will trip people up too. We may be able to get around it by overloading the rule so that it can behave both as a pragma and an ordinary block mode rule.

Ah, I see. I hadn't read the PR code, so I thought the mention of "pragmacomment" in the PR description was because the implementation added a `\comment` pragma that could be turned on either locally in a single tiddler, or globally to apply to all tiddlers, and the `\\` syntax would only be parsed if the `\comment` pragma was turned on.

And, now that I think about that, that would actually be a really good way to implement this, if it's possible to have a global pragma. (How would that be defined? In a tiddler with the `$:/tags/GlobalPragmas` tag?) Or perhaps the `\comment` pragma can only be specified in a local tiddler, and there would be a global setting in one of the normal TW settings tabs for "single-line comments", which could be turned on, and then turned off in individual tiddlers if desired? Then you could even allow the user to specify what the marker for single-line comments would be, e.g. some might prefer `//` and others might want `#` or `;`. The Wiki text parser would only look for comments at the start of a line, to keep the parse cost small and to allow the normal use of those markers (especially `//` which is what I would recommend as the default) in most text. And for those tiddlers where the character you chose needs to appear at the start of a line, e.g. a tiddler where you write `//Some text//` on its own line, it's possible to put `\comment off` at the start of the tiddler to ensure that the `//` in that line won't be seen as a comment. (And if you wanted comments in that tiddler, you'd still have the HTML `<!-- comment -->` syntax to fall back on.)

A global setting would also solve the issue of fighting over diminishing resources, because users who are used to C# or Javascript can use `//`, users who are used to Python or Bash can use `#`, and so on. And users can just turn it off on a single-tiddler basis when they need to use the comment character(s) in their normal meaning in that particular tiddler.

Another possibility would be to add a rule (and carefully document it) that the comment marker *must* be followed by a space to have its comment meaning. Then you could write `#Foo` for a numbered list, and `# comment` for a comment, and the number of tiddlers where you have to use `\comment off` would be much fewer.

Jermolene:
Hi @rmunn we can already use the `\rules` pragma to disable parser rules on a per-tiddler basis, I think the `\comment` pragma would be a subset of that. My experience has been that `\rules` is very useful, but that it can be awkward when dealing with tiddlers from external sources, or when wants to, say, make bulk change the rules that are applied. That's what took me to the ideas in #345.

rmunn:
Speaking of the `\rules` pragma, is it documented anywhere? https://tiddlywiki.com/prerelease/#Pragma doesn't mention it, and https://tiddlywiki.com/prerelease/#Release%205.1.14 mentions a `wikiparserrules` operator which links to a non-existent tiddler (and there's no `wikiparserrules` operator listed anywhere in the "Filter Operators" tiddler either). I can read the code if I want to learn about it, of course, but non-programmers wouldn't be able to do that.

Jermolene:
Thanks @rmunn, that is indeed a hole in the docs. Might you be able to fill in the gaps?

pmario:
@Jermolene ... Please re-consider this PR for the core. It is a way to graatly improve "inline" macro documentation. And it's the only way to add comments in tiddler "parser area", without switching to wikitext parser.

Jermolene:
> @Jermolene ... Please re-consider this PR for the core. It is a way to graatly improve "inline" macro documentation. And it's the only way to add comments in tiddler "parser area", without switching to wikitext parser.

I'd certainly like to introduce a comment syntax, but I want to resolve the issues raised above, particularly the fact that it would be more useful if it was a general purpose rule, and not restricted to being a pragma rule.

pmario:
>I'd certainly like to introduce a comment syntax, but I want to resolve the issues raised above, particularly the fact that it would be more useful if it was a general purpose rule, and not restricted to being a pragma rule.

But imo that's not possible. As soon as you write "standard wikitext" you _end_ the pragma parser. So something like the following isn't possible

```
\define test() asdf
<!-- comment -->
\define xx() xx
```

IMO it has to be a pragma rule, if it is in the middle of pragmas.

And there is a `<!-- general purpose wikitext comment -->` already. Why should we have 2 of them?

pmario:
Thinking about different syntax elements, that could be used as a general purpose comment syntax. 

**Multi line comments from existing "programming" languages**

```
<!-- ... --> ... known from HTML ... WORKS in TW wikitext

/* comment start
comment end */ ... multi line comment known from JS  also works with CSS

--[[
print(10)
--]] ... LUA multi-line comment

%{
  %{
    (Example of a nested comment, indentation is for cosmetics (and ignored).)
  %}
%}  .... MATHLAB 

{- ... -} ... know as comment from Haskell

(* ... *)  ... known from Pascal  and 
{ ... } ... known from Pascal

<# Multi
   Line
   Comment #> ... PowerShell

```

** Single line comments **

```
' ... single quote / apostrophe is know as comment from VB .NET 
⍝ ... APL uses this sign as "single line comment"
!* ... Fortran 90
% ... MATHLAB
```

Not possible combinations since we use them for a different purpose.

```
// can't be used, since it is the indicator for italic wikitext
## can't be used, since it is "ordered list" indicator and it clashes with markdown syntax
; and ;; can't be used ... used as "definition" list indicator
-- ... is used as n-dash
--- ... is used as m-dash
! ... used as heading  
```

Most examples come from: https://en.wikipedia.org/wiki/Comment_(computer_programming)

pmario:
I could imagine: 

```
\\ ... As proposed by the PR for pragmas. ... But we may also use it in wikitext. 
<#  ... Inline rule till the end of the line / Single line 
!* ... inline / single line
```

@Jermolene @saqimtiaz  ... What do you think?

pmario:
I could also imagine this

```
/** inline rule till the end of the line / single line comment in pragma and wikitext

/* comment start
comment end */ ... multi line comment known from JS  also works with CSS

```

Where the second rule would also be possible for dynamic CSS wikitext

Jermolene:
Thanks @pmario.

Given our existing support for the HTML comment syntax I'm primarily interested in adding a single line comment syntax that can be used within pragmas and within ordinary wikitext.

As a JS developer, I'd favour `\\` as the comment marker.

But as ever with these sort of discussions, I think we'll continue to be constrained in how we can extend the wikitext syntax until we add a mechanism for custom vocabularies (or similar).

pmario:
I did a short test. The way the rule is implemented I'd need to add `block: true` to make it work

```
    exports.types = {
        pragma: true,
        block: true
    };
```

So the following code would work with adding 1 line to the existing PR ...  

```
\\ asdf
\define x() xxx

asdf

\\ hallo

test
```

saqimtiaz:
> I'm inclined to close the PR and publish the function as a "stand alone" plugin, instead of being part of a bigger plugin.
> 
> So users can decide, if they want to use it.

@pmario did you end up releasing this plugin? I would be very interested in the user feedback, in particular if or how often the proposed syntax clashed with existing wiki in people's wikis. In general, testing out ideas as plugins first can really help inform better solutions for the core when there is uncertainty.

One advantage of a customizable vocabulary as Jeremy suggests is that if the new syntax was problematic for someone's existing content they could just change the syntax.

pmario:
@saqimtiaz I did release the plugin: https://wikilabs.github.io/editions/pragma-comment/ ... But there was no official announcement in the GG. 

The syntax can not "clash" with their wikitext, because the plugin is a "pragma comment" only. .. As soon as you type the double backslash in the `\\` wikitext area, they will show up as double backslash. 

The latest test that I made was using the "same" double backslash as pragma-comment and wikitext-comment.

Both of them can be deactivated with the `\rules except pragmacomment` which will remove the function on a per tiddler basis. .. The same as for every other parser rule. 

This also shows, that I need to have 2 rules. 

 - 1 pragma comment, named: `pragmacomment`
 - 1 wikitext comment, named: `wikitextcomment`

They both can have the same functionality, but having 2 names will make it possible to separately disable them.

pmario:
I do use the plugin with the "custom-markup" plugin. .. @TiddlyTweeter uses this plugin "in production" and can tell, if he ever had a problem with the "pragma comments"?

saqimtiaz:
Regardless of whether we decide to merge this or not, it is important to acknowledge that **this isn't backwards compatible.** 

Any tiddler that starts with `\\` will have the first line not render. I have quite a few tiddlers exactly like this where I use `\\ text \\` as my own shorthand.

Personally I can adapt quite easily with a bulk search and replace, so it isn't a problem for me. However, it may be for others and we need to be clear that this is a change that breaks backwards compatibility and if it is to be introduced, this needs to be conveyed to end users. 

As @Jermolene states, backwards compatibility can't strictly be maintained if we are ever to extend wikitext. However, we should do so in awareness of and after weighing the consequences, and make sure this is communicated to end users.

pmario:
OK. ... What I want is the possibility to add performant comments into the pragma section, to make it possible to have some "explanation", what's going on. Performant commments means, they need to be outside the `\define ... \end` construct. 

So atm something like this is not possible

```
<!-- new comment in pragma "area" -->
\define xx()
..
\end
```

**What if we make that possible**. IMO it should be a backwards compatible change, since `<!-- comment -->` has always be possible at the beginning of the tiddler. There is nothing new. ... 

The only new thing would be, that the html-comment can be followed by `\define` or any other pragma rule now. @saqimtiaz ... I think, that should be OK for everyone ... right?

The code change seems to be simple. ... BUT a bit more testing will be needed.

TiddlyTweeter:
> @pmario wrote ... The syntax can not "clash" with their wikitext, because the plugin is a "pragma comment" only. .. As soon as you type the double backslash in the `\\` wikitext area, they will show up as double backslash.

The _problem_ with that is we end up with TWO methods of adding comments `\\` in macros and `<!-- ... -->` in wikitext.
I do think it would be better to have ONE method. :-)

TiddlyTweeter:
#### Comments On Pragma Comments & Custom Markup

> @pmario wrote I do use the plugin with the "custom-markup" plugin. .. @TiddlyTweeter uses this plugin "in production" and can tell, if he ever had a problem with the "pragma comments"?

I had NO problems with it. Extremely helpful.

**Upsides** ...
1. It is _Lightweight_. Minimal coding (just `\\`) to do it.
2.  It is _Output Lightweight_. Nothing appears in HTML render.
3. It is _great_ for users like me who don't always know what they are doing and always want to notate to remind themselves what is supposed to be going on. 

**Downsides ...**

1. You can _only_ use it in macro areas, not in wikitext.
2. At the moment it is a _plugin._ That is a problem because you could develop macros using it that otherwise are perfectly normal that will then _break_ if you copied them to a wiki without the comment pragma. 

In brief, this is one of the very few cases where I think that _Comments in general_ should be directly facilitated by the core. As an add-on it doesn't make full sense in any wiki that needs to ever share code.

My 2 cents
TT

TiddlyTweeter:
@pmario wrote ...
> ```
> 
> \define xx()
> ..
> \end
> ```
> 
> **What if we make that possible**. IMO it should be a backwards compatible change ...

Ha! Clever! (If possible).

pmario:
Close this one, will be replaced by a new one, which uses `<!-- html comment -->` syntax.

CodaCodr:
Late to the party again... 🙄 

This works. It's all I need. And using `regexps[]` and regex in general, relatively easy to parse (if you want to sift them out)

```
<!--$ myProc
  @param x <number> A number.
  @param y <number> Another number.
  @comments `myProc` adds two numbers.
$-->
\procedure myProc(x, y) 
...

But I dod like @pmario's in-situ `\\` or `//` if it can be achieved.


\end
```
:)

PR #5293, opened 2020-12-17
set initial tab size to 4

pmario:
related to:  how to change code block indent size？ #5290 

I did test it with windows 10 Edge, FireFox 84 (needs the prefix), Safari latest 

So a bit more testing would be nice.


Jermolene:
Thanks @pmario. Isn't the default tab size already 4 characters? It seems to be on all the browsers I've tried.

pmario:
For me it's 8 chars for all of them

PR #5294, opened 2020-12-17
add dynamic toolbar buttons + HowTo

pmario:
related to:  Proposal: Enable Toolbar Buttons to use dynamic disabled status #5281 

If you want to see, how it works "play with the HowTo". ... The tricky part is: Disabling the shortcut, if the button is visible _and_ disabled. 

Play with it at: https://pmario.github.io/kitchensink/5249-dynamic-toolbar-buttons

See the "in code" comment


core/ui/EditTemplate/body-toolbar-button.tid
```diff
@@ -35,22 +35,23 @@ title: $:/core/ui/EditTemplate/body/toolbar/button
...
 
   class="tc-btn-invisible $(buttonClasses)$"
   tooltip=<<tooltip-text>>
   actions={{!!actions}}
+  disabled=<<disabled>>
 
 ><span
 
-  data-tw-keyboard-shortcut={{!!shortcuts}}
+  data-tw-keyboard-shortcut={{{ [<disabled>compare:string:eq[yes]else{!!shortcuts}] +[trim[yes]] }}}
```
pmario, line 47: This is tricky: ... 

 - `[<disabled>` .. gets the info from `$set` in line 38
 - `compare:string:eq[yes]` ... returns "yes" if the button should be disabled
   - if the tiddler or the field doesn't exist ... I till follow the "else" path
 - `else[!!shortcuts]` .. will set the data-tw... attribute, which is needed for the keyboard handler
 - `+[trim[yes]]` is needed to return an empty stirng, if the shortcut is disabled

saqimtiaz: Just took a quick glance, but would this work, and if so be more legible?

`[<disabled>match[yes]then[]else{!!shortcuts}]`

pmario: Seems to work. Thx for the input!

pmario:
@Jermolene ... What do you think about it? 

 - It contains the code to create dynamic toolbar buttons and 
 - an example how-to create toolbar buttons and keyboard shortcuts. 

We don't have "hands on" docs like that atm.

Jermolene:
> @Jermolene ... What do you think about it?
> 
> * It contains the code to create dynamic toolbar buttons and
> * an example how-to create toolbar buttons and keyboard shortcuts.
> 
> We don't have "hands on" docs like that atm.

I think I commented on the original ticket that it was a good idea. We can consider this for post-v5.1.23, if you'd like to add it to the list.

pmario:
@Jermolene ... bump

pmario:
@Jermolene bump

Play with it at: https://pmario.github.io/kitchensink/5249-dynamic-toolbar-buttons

twMat:
Looks nice but it is not clear in the demo where such a button would be located? (In the sidebar>tools? In the editor toolbar?)

pmario:
It's an editor toolbar named "temporary Bold" So it's at the end of the row with the same function as the "bold" button.

twMat:
No, I mean where does the GUI for your mechanims show up? How/where does the user access your feature? Sorry if I misunderstand something.

Jermolene:
Hi @pmario this PR features lots of whitespace changes. To be mergeable, the changes in the PR need to be restricted to functional changes. I'm happy to accept a separate PR that fixes whitespace issues.

pmario:
@Jermolene ... Did remove the whitespace

pmario:
>No, I mean where does the GUI for your mechanims show up? How/where does the user access your feature? Sorry if I misunderstand something.

It's a "How to create dynamic **editor toolbar** buttons" ... So the button shows up in the editor. 

It also explains, how to create shortcuts for new toolbar buttons. 

And it adds a new feature to every existing toolbar button and every new toolbar button. .. It allows users to use the "disabled" parameter, that was introduced in 5.1.23

Jermolene:
Thanks @pmario

PR #5323, opened 2020-12-26
tabs actions: currentTiddler should remain visible

xcazin:
Actions are a great addition to the tabs macro, but **currentTiddler** (the tiddler outside the macro call) should be readily exposed to them, as **currentTab** is. 

Fortunately, it is only a matter of moving the value of the _actions_ attribute under the scope of **save-currentTiddler**.

Since **currentTab** remains exposed to _actions_ as well after this fix,  no harm was done to the 3 _actions_ calls I've found in search-related tiddlers. 

- https://github.com/Jermolene/TiddlyWiki5/blob/84479bc403de0dbcee775c5c9921c82f71224906/core/ui/SearchResults.tid#L11
- https://github.com/Jermolene/TiddlyWiki5/blob/2a7cdb22c03eb47b7a66ccbc77f2cfa8e615d503/core/ui/AdvancedSearch.tid#L6
- https://github.com/Jermolene/TiddlyWiki5/blob/4d85d267a195725ed84659d9ca2e7612c46a210a/core/ui/AdvancedSearch/Standard.tid#L49

### DIFF
--- a/core/wiki/macros/tabs.tid
+++ b/core/wiki/macros/tabs.tid
@@ -14,7 +14,7 @@ tags: $:/tags/Macro
 <$macrocall $name="currentTab" $type="text/plain" $output="text/plain"/>
 </$transclude>
 </$transclude>
-</$set></$tiddler>$actions$</$button></$tiddler></$set></$list>
+</$set>$actions$</$tiddler></$button></$tiddler></$set></$list>
 </div>
 <div class="tc-tab-divider $class$"/>
 <div class="tc-tab-content $class$">


xcazin:
Since I submitted this PR, I've been using actions intensively with this modified version of the tabs macro, and I've always been glad that they had access to the saved `<<currentTiddler>>`. I believe this fix would be beneficial in the context of 5.2.0, especially if a future _action-call_ functionnality extends the number of use cases for actions.

Jermolene:
Thanks @xcazin looks good, I'm just racking my brains whether there could be any unintended consequences for backwards compatibility. @saqimtiaz do you have any thoughts?

saqimtiaz:
This is a tricky one. If I am reading the wikitext correctly, with this change the value of `currentTiddler` changes from the value of `currentTab` to the value of `currentTiddler` **outside the tabs macro**. 

This isn't backwards compatible as there may be users whose code is relying on the value of `currentTiddler` to point to the `currentTab`. Note however that the value of `currentTiddler` from outside the tabs macro is available in the actions in the variable `save-currentTiddler`. Perhaps a better solution is to document this variable for the actions parameter of the tabs macro.

xcazin:
Hi Saq, thanks for weighting in,

> This is a tricky one. If I am reading the wikitext correctly, with this change the value of `currentTiddler` changes from the value of `currentTab` to the value of `currentTiddler` **outside the tabs macro**.

Indeed.

> This isn't backwards compatible as there may be users whose actions code is relying on the value of `currentTiddler` to point to the `currentTab`. 

Agreed, but this scenario might have happened only since actions have been introduced for tabs, that is 5.1.23. 

> Note however that the value of `currentTiddler` from outside the tabs macro is available in the actions in the variable `save-currentTiddler`. Perhaps a better solution is to document this variable for the actions parameter of the tabs macro.

To me, this rather sounds like a fallback solution; I believe that allowing actions to access both the versatile `currentTiddler` and `currentTab` is worth the arguably small compatibility risk. After all, TW version is about to bump to 5.2.0 :-)

xcazin:
Another way to look at this issue is to consider the tabs macro as part of the set of selection controls that also include _select_, _checkbox_, _radio_ and _range_. As of 5.1.24, they all feature _actions_ that have access to both the state that holds the selected value and the `currentTiddler` that hosts the control. Being the most expressive of the group, it would be consistent that tabs is treated at least equally :-)

That being said, modifying $:/core/macros/tabs is easy for an advanced user (the kind that uses actions in tabs) so that could also be a mean to maintain backwards compatibility.

saqimtiaz:
@xcazin what is relevant here in my opinion isn't which implementation is better. I would go as far as to say the current implementation is an oversight, and if this was an entirely new feature we would be in complete agreement. However, we do have a strict policy of backwards compatibility in TiddlyWiki and with good reason.

> Agreed, but this scenario might have happened only since actions have been introduced for tabs, that is 5.1.23.

Right, however that still means that this change isn't backwards compatible and can break people's code, even if only for a few users. I understand and share the frustration of not being able to clean up suboptimal implementations, but the backwards compatibility commitment is an important one. See #5684 for an example of an even more niche use case that was introduced in 5.1.23, but where we chose to work around the existing implementation rather than break backwards compatibility.

Note that 5.2.0 does not introduce break any backwards compatibility other than for external tools that work with TiddlyWiki HTML files.

I try to review PRs from the perspective of the project's stated goals and the community, rather than my own personal preferences, and thus find it important to point out that this change isn't backwards compatible and goes against our stated policy in that regard. Whether that compatibility risk is small enough to be worthwhile - which can easily become a slippery slope for other changes - is a judgement call that I think @Jermolene needs to make.

Jermolene:
Hi @xcazin @saqimtiaz

The key problem here is that since v5.1.23 people could very easily have used the `currentTiddler` variable mistakenly, and it would have worked. We can't break that behaviour now.

An alternative way around this is suggested by #5691, where we introduce the convention of a double dot prefix to indicate a variable from an outer scope. So, we'd expose the outer `currentTiddler` variable as `..currentTiddler`.

saqimtiaz:
@Jermolene I have been pondering the `..currentTiddler` nomenclature in this context as well. Note that `save-currentTiddler` does currently provide the same access to the outer `currentTiddler` if we want to take more time to think over how and were we are going to use `..currentTiddler`.

Jermolene:
>  Note however that the value of `currentTiddler` from outside the tabs macro is available in the actions in the variable `save-currentTiddler`. Perhaps a better solution is to document this variable for the actions parameter of the tabs macro.

I'd missed that, but yes, that seems the best solution for the moment.

xcazin:
Hi @saqimtiaz, @Jermolene, thank you for having taken the time to elaborate on this. I know that the priority given to backwards compatibilty has been cardinal in TW from the beginning, and obviously rightly so. 

As Saq noted, `save-currentTiddler` allows plugins to use actions in the tabs macro without having to overridde `$:/core/macros/tabs`, and this is the most important point.

Jermolene:
Thanks @xcazin

Jermolene:
> As Saq noted, `save-currentTiddler` allows plugins to use actions in the tabs macro without having to overridde `$:/core/macros/tabs`, and this is the most important point.

I still think it would be worth a docs PR to make sure this is more widely known (and to ensure that subsequent updates to the tabs macro continue to support that variable).

saqimtiaz:
Docs updated in #5722

PR #5324, opened 2020-12-27
Create Japanese (Japan) Edition.tid

donmor:
Added the Japanese version tiddler since the Japanese version of empty.html exists


Jermolene:
Hi @donmor thank you, that was an oversight.

PR #5329, opened 2020-12-28
always test ext in tiddler title and remove it

joshuafontany:
Fixes a few small bug with filesystem extensions. Adds a 32 character limit (avoids a fs vulernability),k and makes it so that you do not need to remove extensions from the tiddler title in FileSystemPaths filters.


Jermolene:
Thanks @joshuafontany

PR #5354, opened 2021-01-02
doc(Developer) syntax error

joebordes:
minor syntax error in documentation


Jermolene:
Good catch, thank you @joebordes. These kinds of little tweaks are much appreciated, they make a big difference to the perceived quality of things.

PR #5362, opened 2021-01-03
Taking advantage of linkedLists in all operator

flibbles:
Now that we have linked lists in Tiddlywiki, we can take advantage of them elsewhere. Here is a three line change to `all[]`. When I take a wiki with 30,000 tiddlers, made like so:
```javascript
var wiki = new $tw.Wiki();
for (var i = 0; i < 30000; i++) {
    wiki.addTiddler({title: i.toString()});
}
$tw.testWiki = wiki;
```
And then I run `$tw.testWiki.filterTiddlers("[all[orphans+tiddlers]]")`, those three lines change this:

Screen Shot 2021-01-03 at 3 00 39 PM

Into this:

Screen Shot 2021-01-03 at 3 03 18 PM

2.5 seconds becomes 200 milliseconds. There's more room for improvement by cutting down all that garbage collection by improving LinkList's memory management, but let's not worry about that right now.

There are a ton of places where we can use LinkedLists like this with minimal changes. Should I continue to make them each their own PR? Or should I just do them all in one go?

### DIFF
--- a/core/modules/filters/all.js
+++ b/core/modules/filters/all.js
@@ -31,7 +31,7 @@ exports.all = function(source,operator,options) {
 	// Get our suboperators
 	var allFilterOperators = getAllFilterOperators();
 	// Cycle through the suboperators accumulating their results
-	var results = [],
+	var results = new $tw.utils.LinkedList(),
 		subops = operator.operand.split("+");
 	// Check for common optimisations
 	if(subops.length === 1 && subops[0] === "") {
@@ -49,10 +49,10 @@ exports.all = function(source,operator,options) {
 	for(var t=0; t<subops.length; t++) {
 		var subop = allFilterOperators[subops[t]];
 		if(subop) {
-			$tw.utils.pushTop(results,subop(source,operator.prefix,options));
+			results.pushTop(subop(source,operator.prefix,options));
 		}
 	}
-	return results;
+	return results.toArray();
 };
 
 })();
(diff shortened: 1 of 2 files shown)


pmario:
Are you sure, the results are the same? So are both arrays equal?

pmario:
I think operators like this one may make sense to have 1 PR. ... But if it is complicated to creates separated PRs, because 1 PR will need an other 1 to be merged too, they should all go to 1 PR

I'm not sure, if the above sentence makes sense ;)

flibbles:
They do come out the same. LinkedList was tested back when I added it for filter prefixes. [This test module exists](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-linked-list.js) to ensure its behavior perfectly mimics the equivalent array operations. I can create tests specific for the `[all[]]` operator if it would make people feel more comfortable.

You're correct. That sentence doesn't make sense. Do you mean it makes sense to have 1 PR per operator? I don't mind. I enjoy profiling them individually, but if that's a hassle for everyone else, I can make a conglomerate PR.

saqimtiaz:
@flibbles thank you. Performance improvements are always welcome.

Separate PRs would be easier to review and test. Please do include tests where they make sense to confirm the behaviour has not changed.

flibbles:
All right. I'll write a test or two when none already exist.

Though I would like to say that I feel testing each invocation of LinkedList still behaves like the array it's replacing will be redundant. The framework already confirms that.

pmario:
>You're correct. That sentence doesn't make sense. Do you mean it makes sense to have 1 PR per operator? 

:) ... Yes .. Where it's easy for you to create 1 PR for 1 operator .. go that way. 

If it is easier to have 1 PR for several "heavily connected" operators that can be done too. ... But it may make the merge process longer.

saqimtiaz:
@flibbles don't think every change needs tests, which is why I said where it makes sense :)

flibbles:
All right. There wasn't a test that touched that part of the `all[]` code, so I wrote one. Can confirm it's the same result both with and without this improvement.

This PR is good to go.

Jermolene:
Great stuff, thanks @flibbles

Issue #5341, opened 2020-12-30
[BUG] Extra dashes in page title when there is no subtitle


kookma:
**Describe the bug**
When Tiddlywiki has no subtitle you see extra dashes!

**To Reproduce**
1. Go to https://tiddlywiki.com/
2. Open $:/ControlPanel
3. Delete the Site subtitle
4. See the page title in your browser

**Expected behavior**
Use a conditional display to remove extra dashes is there is no site subtitle

**Screenshots**
![image](https://user-images.githubusercontent.com/830394/103356301-cc357400-4ac5-11eb-9b53-bdd3f8ca75bc.png)

 - OS: Windows 10
 - Browser Chrome 87
 - Version 5.1.23

Jermolene:
Hi @kookma the tiddler `$:/core/wiki/title` is used to set the page title. By default, it is set to:

```
{{$:/SiteTitle}} --- {{$:/SiteSubtitle}}
```

You can edit `$:/core/wiki/title` to remove the dashes, or replace them with a different separator.

kookma:
Many thanks for Clarification!

kookma:
Hi @Jermolene 
Does it make sense to change `$:/core/wiki/title` to

```
<$list filter="[{$:/SiteSubtitle}trim[]minlength[1]]" emptyMessage="""{{$:/SiteTitle}}""">
{{$:/SiteTitle}} --- {{$:/SiteSubtitle}}
</$list>
```
Note: `trim[]` is used to ignore spaces (empty field but has spaces). The `!is[blank]` does not work with `$:/SiteSubtitle` contains spaces.

If so, please let me know if it worth to make a PR.

Jermolene:
Hi @kookma 

> Does it make sense to change `$:/core/wiki/title` to

That's not a bad idea. The current expectation is that one would override `$:/core/wiki/title` if one didn't want to have the dash separator but the `$:/core` prefix would (rightly) discourage people.

I think it could be simplified to:

```
{{$:/SiteTitle}}
<$list filter="[{$:/SiteSubtitle}trim[]minlength[1]]" variable="ignore">
--- {{$:/SiteSubtitle}}
</$list>
```

(I added the `variable="ignore"` because otherwise we'll be unintentionally setting the `currentTiddler` variable.

kookma:
This is lovely! Thank you @Jermolene 
I submit a PR within few minutes

PR #5370, opened 2021-01-05
Remove extra dashes in page title

kookma:
When Tiddlywiki has no subtitle you see extra dashes! See https://github.com/Jermolene/TiddlyWiki5/issues/5341

This PR smartly adds double dashes as separator when there is a site subtitle and show only site title when site subtitle is empty.


Jermolene:
Thanks @kookma

PR #5375, opened 2021-01-07
Update Formatting List Results as Tables with CSS - Specified Columns…

CodaCodr:
… Methods.tid

-moz-column* seems to have been dropped in Firefox. Added the non-prefixed, standard properties.


Jermolene:
Thanks @CodaCodr

PR #5376, opened 2021-01-07
Fixed issue where [lookup[]] could emit undefined

flibbles:
If [lookup[]] could not lookup the tiddler and it had no suffix, it would put `undefined` into the stream. Thus, a filter like: `[lookup[nonexistent]prefix[anything]]` would throw an exception, because prefixed tries to call `substr` on undefined. Here is a quick fix.

### DIFF
--- a/core/modules/filters/lookup.js
+++ b/core/modules/filters/lookup.js
@@ -22,7 +22,7 @@ Export our filter function
 exports.lookup = function(source,operator,options) {
 	var results = [];
 	source(function(tiddler,title) {
-		results.push(options.wiki.getTiddlerText(operator.operand + title) || operator.suffix);
+		results.push(options.wiki.getTiddlerText(operator.operand + title) || operator.suffix || '');
 	});
 	return results;
 };


flibbles:
@Jermolene, @pmario but my REAL reason for pushing this (and also the reason I found it), is because the memory efficient linkedList I'm working on really doesn't having undefined as an inputted value.

Am I correct in assuming that filters should never have undefined in their source at any point? And thus the behavior is undefined? Because undefined leads to problems all over the place, and maybe it's time we just do something like `title.toString()` so all null and undefined becomes "null" and "undefined", and it becomes easier to discover problems.

I will still endeavor to make my LinkedList at least someone robust against them though.

Jermolene:
Thanks @flibbles I fear there may be more bugs like this.

flibbles:
No problem, @Jermolene, but is `undefined` something LinkedList needs to take into consideration? Or can I just toString() any incoming list items?

Jermolene:
> No problem, @Jermolene, but is `undefined` something LinkedList needs to take into consideration? Or can I just toString() any incoming list items?

Pushing `undefined` would generally indicate a bug, and so we might want LinkedList to fail with an error. Doing so wouldn't be entirely backwards compatible but might help find other similar bugs.

pmario:
IMO throwing errors until we found those little creatures ;)

flibbles:
I cannot tell you how happy this makes me.

My linkedlist tests for end-of-list by looking for undefined.
BUT if it has to allow for undefined, it has to use $tw.utils.hop.
BUT it uses values as keys now too, so undefined is really "undefined"
So I'd have to prefix ACTUAL strings like null and undefined with `_`, so they're `_null` and `_undefined`.
So I'd have to prefix ALL string values that start with `_`. so `_undefined` would be `__undefined`, and `__undefined` would be `___undefined`.
And then I would have gone insane and eaten all the towels in my linen closet.

PR #5377, opened 2021-01-07
Cleaned up jasmine test suite output

flibbles:
Also testing for expected log messages, instead of just letting them print
to the console every single time, constantly making you think there's some
warning you need to worry about, and making all those dots not line up nicely.

Screen Shot 2021-01-07 at 2 32 45 PM
:(

Screen Shot 2021-01-07 at 2 31 46 PM
:)

### DIFF
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -14,6 +14,22 @@ Tests the filtering mechanism.
 /* global $tw, require */
 "use strict";
 
+// This wrapper method is used to collect warnings which should be emitted
+// by certain deprecated tests.
+function collectLog(block) {
+	var messages = [];
+	var oldLog = console.log;
+	console.log = function(a) {
+		messages.push(Array.prototype.join.call(arguments, ' '));
+	}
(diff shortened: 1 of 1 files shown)


editions/test/tiddlers/tests/test-filters.js
```diff
@@ -14,6 +14,22 @@ Tests the filtering mechanism.
 /* global $tw, require */
 "use strict";
 
+// This wrapper method is used to collect warnings which should be emitted
+// by certain deprecated tests.
+function collectLog(block) {
+	var messages = [];
+	var oldLog = console.log;
+	console.log = function(a) {
+		messages.push(Array.prototype.join.call(arguments, ' '));
```
Jermolene, line 23: A very small point, but we use double quotes for string literals (except for a few cases where the string contains double quotes).

Jermolene:
Thanks @flibbles

editions/test/tiddlers/tests/test-filters.js
flibbles: Switched over. Should be good to merge?

PR #5380, opened 2021-01-09
Memory efficient linked list

flibbles:
The more I convert Tiddlywiki over to using the Linked List, the more of a hungry hungry memory hippo it reveals itself to be. Currently, it's never used in a context where it will be retained, which is good, because using it with large (60K entries) can spike the memory several mB.

I've created a new Linked List which (usually) only uses three objects, instead of an object for every entry. It's less of a link list and more of a link table which relies on these assumptions

1. The only entries are strings. No undefined, null, objects, or anything like that. Exceptions are thrown in those cases.
2. Duplicate values must be supported, but they are very rare (<1% of use cases), and thus it's okay to be a little inefficient with them.

Since the previous Linked List was already pretty good, I had to ramp up the test to really outline the difference.
`for (var i = 0; i < 4; i++) {$tw.testWiki.filterTiddlers("[all[tiddlers+tiddlers+tiddlers+tiddlers]]")}` on 60K tiddlers:

## Old Linked list
Screen Shot 2021-01-08 at 7 53 11 PM
Large clumps of garbage collection wasting time to pick up all those little objects.

## New improved Linked list
Screen Shot 2021-01-08 at 7 54 12 PM
Hardly a garbage collection in sight.

(Meanwhile, the tw5.com website takes 90 seconds to do this.)

Granted, it was a pretty intense test case, but even with my old 30K testing suites, the new form was usually completed in about 70% of the old time. Moreover, this opens the door to allowing Linked Lists to persist without eating up memory.

I've also nearly doubled the number of tests since this implementation was a much trickier. But I'm pretty sure it's solid now. Let me know your thoughts.

### DIFF
--- a/core/modules/utils/linked-list.js
+++ b/core/modules/utils/linked-list.js
@@ -110,7 +110,6 @@ function _removeOne(list,value) {
 		if(typeof list.next[prev] === 'object') {
 			if(next === undefined) {
 				// Must have been last, and 'i' would be last element.
-				// Just pop instead
 				list.next[prev].pop();
 			} else {
 				var i = list.next[prev].indexOf(value);
@@ -128,7 +127,7 @@ function _removeOne(list,value) {
 	if(next !== undefined) {
 		if(typeof list.prev[next] === 'object') {
 			if(prev === undefined) {
-				// Must have been first, and 'i' would be 0. Just shift instead
+				// Must have been first, and 'i' would be 0.
 				list.prev[next].shift();
 			} else {
 				var i = list.prev[next].indexOf(value);
@@ -162,7 +161,8 @@ function _linkToEnd(list,value) {
 				list.next[value] = [list.next[value]];
 				list.prev[value] = [list.prev[value]];
 			} else if(typeof list.next[value] === 'undefined') {
-				// special case. The list already contains exacly 1 "value".
+				// list.next[value] must be undefined.
+				// Special case. List already has 1 value. It's at the end.
 				list.next[value] = [];
 				list.prev[value] = [list.prev[value]];
 			}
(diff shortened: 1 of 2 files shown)


core/modules/utils/linked-list.js
```diff
@@ -14,105 +14,181 @@ function LinkedList() {
...
+function _linkToEnd(list,value) {
+	if(list.first === undefined) {
+		list.first = value;
+	} else {
+		// Does it already exists?
+		if(list.first === value || list.prev[value] !== undefined) {
+			if(typeof list.next[value] === 'string') {
+				list.next[value] = [list.next[value]];
+				list.prev[value] = [list.prev[value]];
+			} else if(typeof list.next[value] === 'undefined') {
```
Jermolene, line 163: Apologies, I may have missed this in earlier commits, but our coding style is to always use double quotes for string literals (there are some instances where we use single quotes for strings that themselves contain double quotes).

Jermolene:
Thanks @flibbles. We should certainly pay attention to memory usage, and I'd be interested in improving other parts of the core with poor memory usage.

I've left a minor coding style comment, otherwise looks good.

flibbles:
@Jermolene, sure I can absolutely change those quotes over to doubles, but may I ask about that?

Why double quotes to singles? All the javascript modules get html escaped before being written to the output Tiddlywiki file. That means that all your double quotes become seven characters `\&quot;` while all your single quotes remain one `'`. Seems like single quotes would make a lot more sense for Tiddlywiki. At least to me.

flibbles:
I'm guessing the double quotes has something to do with how the JSON specifications?  Well, either way, the PR is switched over and ready to merge.

flibbles:
I discuss the double quote thing more in #5383, where I discuss if escaping them is even necessary.

PR #5382, opened 2021-01-09
Fix  #5308 - WidgetSubclassingMechanism not working with widgets that add EventListeners (or logic ?) in constructor

### LINKED ISSUE #5308: [BUG] WidgetSubclassingMechanism fails for widgets that set event-handlers
**Describe the bug**
When subclassing a widget which registers `eventHandlers` (e.g. `navigator, scrollable`) using the WidgetSubclassingMechanism (as described in https://tiddlywiki.com/dev/#WidgetSubclassingMechanism), one gets the error "this.eventListeners is undefined" because the underlying widget-constructor (which initializes the field) is called too late in the flow. 

**To Reproduce**
Create a widget-subclass (module-type: widget-subclass) implementation target to e.g. navigator like this:

```
(function(){
    /*global $tw: false */
    "use strict";
    exports.baseClass = "navigator";
    exports.name = "limited-stack-navigator";
    exports.constructor = function(parseTreeNode,options) {
    };
    exports.prototype = {};
    exports.prototype.handleNavigateEvent = function(event) {
        console.log("HANDLE");
        return false;
    };
})();
```
and restart the wiki. Exception-details appear in console.

**Expected behavior**
Navigator should be overwritten and navigate-event shouldn't be handled anymore. Instead a console.log stating "HANDLE" should appear.

 - OS: Win10
 - Browser Firefox
 - Version 84

BurningTreeC:
This PR is an attempt to fix  #5308

- The adding of EventListeners is moved from the constructor to render()
  - in the navigator widget
  - in the fieldmangler widget
  - in the scrollable widget
  - in the linkcatcher widget
- The additional adding of logic is also moved from the constructor to render()
  - in the edit-bitmap widget
  - in the list widget
  - in the vars widget. note the `Widget.call(this)` call - I'm not sure what it should do / if it can be removed entirely

### DIFF
--- a/core/modules/widgets/edit-bitmap.js
+++ b/core/modules/widgets/edit-bitmap.js
@@ -38,17 +38,17 @@ Render this widget into the DOM
 */
 EditBitmapWidget.prototype.render = function(parent,nextSibling) {
 	var self = this;
+	// Initialise the editor operations if they've not been done already
+	if(!this.editorOperations) {
+		EditBitmapWidget.prototype.editorOperations = {};
+		$tw.modules.applyMethods("bitmapeditoroperation",this.editorOperations);
+	}
 	// Save the parent dom node
 	this.parentDomNode = parent;
 	// Compute our attributes
 	this.computeAttributes();
 	// Execute our logic
 	this.execute();
-	// Initialise the editor operations if they've not been done already
-	if(!this.editorOperations) {
-		EditBitmapWidget.prototype.editorOperations = {};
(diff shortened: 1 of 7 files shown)


Jermolene:
Thanks @BurningTreeC I think it would be better to move the code to the very top of the render method, just in case there's a subtle dependency on, say, the execute method.

BurningTreeC:
>I think it would be better to move the code to the very top of the render method, just in case there's a subtle dependency on, say, the execute method.

Done, @Jermolene

Jermolene:
Thanks @BurningTreeC

PR #5383, opened 2021-01-09
double quotes are no longer escaped in html bodies

flibbles:
It's entirely cool if you don't merge this PR. If nothing else, I'd need to update the documentation on $view before it'd be ready.

This is to discuss double quotes in Tiddlywiki. We escape them _all_, _and_ it's part of the style guide. However if we _didn't_ escape them, then we shave nearly 300 kB off of the generated Tiddlywiki file (that's about 10%). I've done this with this PR, and it all works fine.

According to the [Google style guide](https://google.github.io/styleguide/htmlcssguide.html), they even recommend that you don't.

> Do not use entity references.
> 
> There is no need to use entity references like `&mdash;`, `&rdquo;`, or `&#x263a;`, assuming the same encoding (UTF-8) is used for files and editors as well as among teams.
> 
> The only exceptions apply to characters with special meaning in HTML (like < and &) as well as control or “invisible” characters (like no-break spaces).

As far as I can tell, in the textContent of HTML, the double quote has no special meaning.

The only trick about this is we need a separate htmltextencode keyword for view, because attributes _do_ still need to escape double quotes.

So, why not?

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -515,6 +515,15 @@ exports.htmlEncode = function(s) {
 	}
 };
 
+// Converts like htmlEncode, but forgets the double quote for brevity
+exports.htmlTextEncode = function(s) {
+	if(s) {
+		return s.toString().replace(/&/mg,"&amp;").replace(/</mg,"&lt;").replace(/>/mg,"&gt;");
+	} else {
+		return "";
+	}
+};
+
 // Converts all HTML entities to their character equivalents
 exports.entityDecode = function(s) {
 	var converter = String.fromCodePoint || String.fromCharCode,
--- a/core/modules/widgets/view.js
+++ b/core/modules/widgets/view.js
@@ -65,6 +65,9 @@ ViewWidget.prototype.execute = function() {
 		case "htmlencoded":
 			this.text = this.getValueAsHtmlEncoded();
 			break;
+		case "htmltextencoded":
+			this.text = this.getValueAsHtmlTextEncoded();
+			break;
 		case "urlencoded":
 			this.text = this.getValueAsUrlEncoded();
 			break;
@@ -158,6 +161,10 @@ ViewWidget.prototype.getValueAsHtmlEncodedPlainWikified = function(mode) {
 
 ViewWidget.prototype.getValueAsHtmlEncoded = function() {
 	return $tw.utils.htmlEncode(this.getValueAsText());
+};
+
+ViewWidget.prototype.getValueAsHtmlTextEncoded = function() {
+	return $tw.utils.htmlTextEncode(this.getValueAsText());
 };
 
 ViewWidget.prototype.getValueAsUrlEncoded = function() {
--- a/core/templates/html-div-tiddler.tid
+++ b/core/templates/html-div-tiddler.tid
@@ -5,5 +5,5 @@ title: $:/core/templates/html-div-tiddler
 This template is used for saving tiddlers as an HTML DIV tag with attributes representing the tiddler fields.
 
 -->`<div`<$fields template=' $name$="$encoded_value$"'></$fields>`>
-<pre>`<$view field="text" format="htmlencoded" />`</pre>
+<pre>`<$view field="text" format="htmltextencoded" />`</pre>
 </div>`
--- a/editions/tw5.com/tiddlers/widgets/ViewWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ViewWidget.tid
@@ -29,6 +29,7 @@ The following formats can be specified in the `format` attribute:
 |!Format |!Description |
 |''text'' |Plain text (default) |
 |''htmlencoded'' |The field is displayed with HTML encoding |
+|''htmltextencoded'' |The field is displayed with HTML encoding, only double quotes (") are not escaped. This creates a more compact htmlencoding appropriate for html text content, but //not// for attributes. |
 |''urlencoded'' |The field is displayed with URL encoding |
 |''doubleurlencoded'' |The field is displayed with double URL encoding |
 |''htmlwikified'' |The field is wikified according to the mode attribute and the resulting HTML returned as plain text (ie HTML elements will appear in plain text) |
--- a/plugins/tiddlywiki/tiddlyweb/html-div-tiddler.tid
+++ b/plugins/tiddlywiki/tiddlyweb/html-div-tiddler.tid
@@ -5,5 +5,5 @@ title: $:/core/templates/html-div-tiddler
 This template is used for saving tiddlers as an HTML DIV tag with attributes representing the tiddler fields. This version includes the tiddler changecount as the field `revision`.
 
 -->`<div`<$fields exclude='text revision bag' template=' $name$="$encoded_value$"'></$fields>` revision="`<<changecount>>`" bag="default">
-<pre>`<$view field="text" format="htmlencoded" />`</pre>
+<pre>`<$view field="text" format="htmltextencoded" />`</pre>
 </div>`


Jermolene:
Hi @flibbles I think there might be a couple of issues here.

We use double quotes for string literals for consistency with existing code; back in 2004 I used double quotes because I came from an (ancient) C/C++ background. Nowadays it's totally out of line with modern JavaScript practice, of course.

I'd missed the point that the (over)use of double quotes leads to an increase in the size of the JSON encoding of plugins.

Anyhow, in principle, I'd be happy to consider changing more or less any of our coding style guidelines. The catch is that we still have to try to keep the codebase consistent, and so the implication is that we'll need to do a bulk conversion. Ideally, such conversions should be accompanied by automated tests demonstrating that the before and after ASTs are the same.

In this particular case, there's an obvious saving to be had from changing our encoding, and so it may well be worth the effort. The only other factor for me is that I am very interested in adding support for GZip encoding of the content of plugins. That would be quite an undertaking as well, so we might not want to do both at the same time.

I don't recall exactly why our HTML encode function encodes double quotes, but I'm guessing it was so that we could use the same encoding function for attribute content. Your approach of a new function for the conversion without double quotes seems a reasonable way to assure backwards compatibility.

So, in summary, I'd be inclined to merge this, and to consider a separate PR for the first issue.

pmario:
I think it would be great if we could save 300kByte in TW size!

There are some other issues that would need to change the whole codebase, which will invalidate _all_ open PRs. But I think it would be worth it. 

eg: consistent settings for ESLint, fix leading and trailing spaces and may be more

pmario:
If we change our styleguide, I think 2 major goals needs to be. "Don't hurt readability" AND "Don't increase TW file size!"

flibbles:
So if we adopt the changes I suggest in this PR, the double quote vs single quote thing becomes much less relevant. Instead of 7 characters. Double quotes in javascript will only take 2 `\"`, as opposed to the single quote 1 `'`. And then switching from doubles to singles doesn't save enough to make invalidating every PR worth it.

The super low hanging fruit here is all the JSON encoded shadow tiddlers. It's `\&quot;` for miles. Take a look:

Screen Shot 2021-01-09 at 7 04 19 PM

There's a little more I want to do with this PR before it should be merged though.

pmario:
> The only other factor for me is that I am very interested in adding support for GZip encoding of the content of plugins. That would be quite an undertaking as well, so we might not want to do both at the same time.

I think most plugins are so small, that size doesn't matter too much. I think the "decompression" overhead is much higher than the win. 

I also think, that we have to make some experiments which "decompression" method is the best for our usecase. It can make a huge difference! See this article: https://snapcraft.io/blog/why-lzo-was-chosen-as-the-new-compression-method

The article just shows, that it is important to have some experiments and measure first.

flibbles:
I guess the only real question I have is are people okay with `<$view field="text" format="htmltextencoded" />`. Leaving `htmlencoded` untouched is probably best for backward compatibility, I think. So the new format name is `htmltextencoded`? That's cool?

pmario:
> ... <$view field="text" format="htmltextencoded" />

I think in reality it doesn't matter. Nobody really knows what to use. They experiment until they found the right one. .. At least that's true for me.

flibbles:
I agree with @pmario. At some point, is there enough gain from it? Because we already do stream compression when sending it over a network. Making browsers decompress plugin contents (AFTER 64bit decoding it) might end up being counter productive.

I know the idea of minifying javascript has been discussed and thrown out, but I've been considering writing a plugin (NOT core code) which would allow a NodeJS server to minify plugins before sending them to browsers. That, I think, is the only ground to be gained with plugin compression when stream compression is already happening.

> I think in reality it doesn't matter. Nobody really knows what to use. They experiment until they found the right one. .. At least that's true for me.

Yeah, I've done exactly this several times. But I'd still like the name to have had some forethought and consensus.

pmario:
The 1 feature I really like in  TW debugging is: It works out of the box. No extra tooling nonsense is needed 

So I think code obfuscation would be the worst thing we could do to TW.

pmario:
Did you ever try to find a problem at 1 of Tobi Beers plugins. It's the pure horror, because the code is minified. I don't even try to find anything. It's way too much effort for no bucks.

flibbles:
Yeah. That's why I'd make it an optional plugin. Mostly I got this idea after people started griping to me about how Relink is too big. And also, other plugin developers already minify their plugins, but they have use stupid non-tiddlywiki techniques like gulp with NPM.

Even a plugin that just stripped comments would make me much less adverse to commenting my code they way I'd prefer to do. For Relink for example, people who'd check out the project and run with NodeJS wouldn't get it minified, but people who downloaded from the plugin library, or who drag and dropped the download pill, would.

I agree that applying it to the core code wouldn't be great. But some plugin developers just _have_ to minify. It's in their blood.

flibbles:
The neat part would be how anyone who wanted everything minified would just need to include such a plugin in their NodeJS project, and it'd be minified by the time it reaches their browser. But it'd be their choice to include such a plugin, and if they hit a bug, then they'd disable the plugin, reload, and reproduce.

Just a thought. It was my next big idea after Relink.

saqimtiaz:
I did a lot of work with TiddlyWiki Classic in very low bandwidth environments, where even now 10+ years later TW5 would be a non-starter due to the file size of empty.html and the lack of control over how the file will be served.

So while I completely agree that the standard distribution of TW should remain unminified as it is now, there would still be value in the availability of a version with a minified core. I know at least one person currently considering using TiddlyWiki in an educational context in a bandwidth scarce environment but unsure if the file size is workable.

pmario:
may be OT: 

I think it should be possible to create a TiddlyWiki core browser AddOn, that can handle a content-only html-file. ... 

The problem is, that most users are "crazy" about installing way too many plugins at the beginning, which leads to all kind of problems in a setting like this.

pmario:
>... I know at least one person currently considering using TiddlyWiki in an educational context in a bandwidth scarce environment but likely to look elsewhere due to the file size.

Where does a bandwidth scarce environment exist in our days? If you install TW on a raspberry Pi with a little bit of housing and a power connection you have your own environment, for about 100 €s

saqimtiaz:
> > ... I know at least one person currently considering using TiddlyWiki in an educational context in a bandwidth scarce environment but likely to look elsewhere due to the file size.
> 
> Where does a bandwidth scarce environment exist in our days? If you install TW on a raspberry Pi with a little bit of housing and a power connection you have your own environment, for about 100 €s

That is a very privileged position to take. There are countless places in sub-Saharan Africa and South America where we work with education and healthcare - and have previously leveraged TiddlyWiki Classic to great effect - where connectivity is extremely bandwidth and time limited, and where €100 is an incredible chunk of available resources. In those circumstances the idea of a raspberry Pi is so far fetched that I don't even know what to say.

In fact I am currently considering whether to steer my colleague towards TiddlyWiki Classic purely because of the file size, which would be quite a shame as I'd rather flesh out the TW5 ecosystem. I've even been tempted to consider using TW5 in some of my own current projects in Central Asia, but bandwidth is also a big consideration there and the file size just isn't tenable.

flibbles:
It may be OT (over-thinking?) indeed. I'm not about to take a crack at it anytime soon. And if I do, as a plugin, it makes minifying something the core developers never have to think about again. Minifying would be there for anyone who wanted, and not present for most of us who don't.

Either way, I'll work on this double-quote fix. It's only gains.

flibbles:
Huh... can somebody who speaks French help me out here? Is there somewhere I go for this?

This is what I have for English
```
|''htmlencoded'' |The field is displayed with HTML encoding |
|''htmltextencoded'' |The field is displayed with HTML encoding, only double quotes (") are not escaped. This creates a more compact htmlencoding appropriate for html text content, but //not// for attributes. |
```
But this is what I have for French.
```
|''htmlencoded'' |Le champ est affiché jusqu'à l'encodage HTML |
|''htmlencoded'' |Le champ est affiché jusqu'à l'encodage HTML, uhh... Je suis Muzzy. Le //grand// Muzzy! |
```

saqimtiaz:
@flibbles I don't think it is a requirements for PRs to update the non-English documentation and I think we rely on the contributors that work on the translations to update it.

flibbles:
Oh! In that case, this PR is good to go.

**One quick note:** I did not update `html-div-tiddler-remove-prefix` to use htmltextencoded, because that's in the `classictools` plugin, and I have no idea whatsoever what that plugin is about. Should it be updated too? I updated the tiddlyweb div-tiddler after all.

flibbles:
Remember when I said I wasn't going to take a crack at an uglifying plugin anytime soon?

I lied. https://flibbles.github.io/tw5-uglify/

But still, this PR is good to go.

flibbles:
Of all the size optimizing PRs I've pushed, this remains the lowest and biggest hanging fruit. I don't see how it would cause problems, but I'd recommend merging it sooner rather than later to give time for people to test it out.

Also, then I can properly resolve the conflicts with this and #5386.

pmario:
@flibbles ... I think Jeremy is only notified if you ping with @Jermolene ?!

flibbles:
Really? I actually didn't know that. This explains several of my dusty PRs.

Do you know if there's some etiquette about pinging him? Because isn't he the only one who can merge these?

Jermolene:
> Of all the size optimizing PRs I've pushed, this remains the lowest and biggest hanging fruit. I don't see how it would cause problems, but I'd recommend merging it sooner rather than later to give time for people to test it out.
> 
> Also, then I can properly resolve the conflicts with this and #5386.

Apologies @flibbles @pmario, just to confirm I do get pings even if my username isn't mentioned.

I'll have a look at this as soon as I can.

Jermolene:
Thanks for your patience @flibbles.

Issue #5342, opened 2020-12-30
[BUG]


Arlen22:
**Describe the bug**

https://github.com/Jermolene/TiddlyWiki5/blob/7ec6d3703150469e73dabcf25655192767092a77/core/modules/server/routes/get-tiddlers-json.js#L31-L33

GET tiddlers.json (aka LIST tiddler) has an option to exclude system tiddlers from the list, in the spirit of restricting syncing of system tiddlers to the server, but GET tiddler and PUT tiddler do not check whether system tiddlers are allowed to be synced before saving or serving system tiddlers. This seems a little strange to me, and I would think that this should rather be checked in PUT tiddler to actually prevent a system tiddler from being modified, and the client should be able to detect on its own whether it is supposed to be changing system tiddlers or not. 

In other words LIST and GET should have identical behaviour, but PUT should definitely have a check to prevent tiddlers from being written. I don't know where this feature is used, but I'm assuming it is used in cases where there is a custom index file instead of the autogenerated one from `$:/core/save/all`. 

This also seems like a sort of strange feature as it seems like config options would not be saved back to the server. Is the idea here that system tiddlers on the client would be saved into a separate user store? I can certainly see this heading that direction.

saqimtiaz:
@Arlen22 I can provide a little background on what this config setting does and why it was introduced.

The new syncer introduced in 5.1.22 caused several problems and broke workflows for multi-user and multi-client situations by syncing from the server to the client, tiddlers such as `$:/StoryList` which previously were only synced from the client to the server.

It was decided to introduce this config setting as a temporary stop-gap measure to allow restoring the behaviour from pre 5.1.22, until a proper solution for configuring what system tiddlers to sync and how is formulated. Per user storage was indeed discussed but is not related to this config setting. See #4882 and #4868 

 When `$:/config/SyncSystemTiddlersFromServer` is set to "yes", system tiddlers are synced from the client to the server, but not from the server to the client. This restores the expected behaviour from before the syncer changes in 5.1.22, and means that one users $:/StoryList does not get synced to other clients.

It may be helpful moving forwards if you give this issue a title that makes it easier to find.

Arlen22:
@saqimtiaz So what would be the scenario where you would want to save a system tiddler back to the server but not have other clients see that change until they reload the page. `$:/StoryList` makes sense, but can't it be added to a list of ignored tiddlers somehow? I'm just asking to try to understand better.

saqimtiaz:
[...]

PR #5384, opened 2021-01-10
Add server/client sync filters

Arlen22:
This adds the server and client sync filters as proposed by me in #5342. 

The changes are mainly to the syncer, and apply to the file system adapter as well, but that does not break anything as write-only tiddlers will still be saved, and read-only tiddlers are ignored anyway because the file system adapter does not read the file system. In the future, whatever mechanism is used for bags will most likely overcome this problem.

### DIFF
--- a/plugins/tiddlywiki/tiddlyweb/config-tiddlers-filter.tid
+++ b/plugins/tiddlywiki/tiddlyweb/config-tiddlers-filter.tid
@@ -1,2 +1,2 @@
-title: $:/config/Server/ExternalFilters/[all[tiddlers]] -[[$:/isEncrypted]] -[prefix[$:/temp/]] -[prefix[$:/status/]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] -[[$:/library/sjcl.js]] -[[$:/core]]
+title: $:/config/Server/ExternalFilters/[is[tiddler]] -[[$:/core]] -[[$:/library/sjcl.js]] -[prefix[$:/boot/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/isEncrypted]] -[prefix[$:/status/]] -[prefix[$:/state/]] -[prefix[$:/temp/]]
 text: yes
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -19,7 +19,8 @@ Syncer.prototype.titleIsLoggedIn = "$:/status/IsLoggedIn";
 Syncer.prototype.titleIsAnonymous = "$:/status/IsAnonymous";
 Syncer.prototype.titleIsReadOnly = "$:/status/IsReadOnly";
 Syncer.prototype.titleUserName = "$:/status/UserName";
-Syncer.prototype.titleSyncFilter = "$:/config/SyncFilter";
+Syncer.prototype.titleSyncServerFilter = "$:/config/SyncServerFilter";
+Syncer.prototype.titleSyncClientFilter = "$:/config/SyncClientFilter";
 Syncer.prototype.titleSyncPollingInterval = "$:/config/SyncPollingInterval";
 Syncer.prototype.titleSyncDisableLazyLoading = "$:/config/SyncDisableLazyLoading";
 Syncer.prototype.titleSavedNotification = "$:/language/Notifications/Save/Done";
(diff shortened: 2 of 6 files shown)


Jermolene:
Thanks @Arlen22 I've had a quick look. What about backwards compatibility? That is the constraint that gave rise to the current implementation.

Arlen22:
It retains compatibility with the current implementation because if nothing is specified in the client sync filter (which is appended to the server sync filter) then it functions exactly the same. The only difference is that I changed the sync filter to match what the tiddlyweb adapter requests. I should probably take a look at that and also make sure I didn't miss any other uses of the sync filter.

Jermolene:
Hi @Arlen22 perhaps I'm reading the PR wrongly, but it seems that the syncer is no longer obeying `$:/config/SyncFilter`, which wouldn't be backwards compatible for users that have customised that tiddler.

plugins/tiddlywiki/tiddlyweb/config-tiddlers-filter.tid
```diff
@@ -1,2 +1,2 @@
-title: $:/config/Server/ExternalFilters/[all[tiddlers]] -[[$:/isEncrypted]] -[prefix[$:/temp/]] -[prefix[$:/status/]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] -[[$:/library/sjcl.js]] -[[$:/core]]
+title: $:/config/Server/ExternalFilters/[is[tiddler]] -[[$:/core]] -[[$:/library/sjcl.js]] -[prefix[$:/boot/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/isEncrypted]] -[prefix[$:/status/]] -[prefix[$:/state/]] -[prefix[$:/temp/]]
```
pmario, line 1: IMO a filter should reduce the number of elements as fast as possible. I'm sure there are more `[prfix[xxx]]` elements than `$:/core` in the list. Why did you change the order in that way?

Arlen22:
@Jermolene  and @pmario, The old version (red line) in the change @pmario commented on was what tiddlers.json was returning for the skinny tiddler list and the new version (green line) was what the `$:/config/SyncFilter` was set to (which I renamed `$:/config/SyncServerFilter`. In retrospect I think it would be easier at the moment to change the name back to SyncFilter because it is used elsewhere as well. 

So basically the only change I made here was that tiddlers.json using the SyncFilter instead of having its own separate filter that's hardcoded.

plugins/tiddlywiki/tiddlyweb/config-tiddlers-filter.tid
Arlen22: See my comment here: https://github.com/Jermolene/TiddlyWiki5/pull/5384#issuecomment-758172726

Arlen22:
The other change I made was to have a separate filter appended to the SyncFilter (which I named SyncClientFilter) to change the tiddlers that can be synced back to the server if desired. This is a more granular approach obviously than just excluding system tiddlers from tiddlers.json and also allows tiddlers to be made readonly by subtracting them in the filter. 

I explained it better at https://github.com/Jermolene/TiddlyWiki5/issues/5342#issuecomment-754251110 so I copied that here to try to explain it better. 

Basically, `SyncServerFilter` and `SyncClientFilter` work in tandom. `SyncServerFilter` would affect which tiddlers get synced to the client through all of the several sync mechanisms, and `SyncClientFilter` would affect which tiddlers get synced back to the server. I'm thinking that it would be chained to `SyncServerFilter` so it would add or remove tiddlers to either make tiddlers readonly from the server or write-only to the server. 

Here is an example of what I'm talking about. 

```txt
SyncServerFilter: [all[tiddler]] -[[$:/core]] -[[$:/StoryList]]
SyncClientFilter: [[$:/StoryList]] -[[$:/ServerStatus]] +[tag[Only Save Tiddlers Tagged With This!]]
```

I'm not sure exactly how that would get executed, but this is what I'm imagining. Additional tiddlers can be added or removed to `SyncClientFilter` as shown above. As you can see, adding a plus sign should modify everything prior to it in the filter, if I understand right. 

```js
let tiddlersToServe = `${SyncServerFilter}`;
let tiddlersToSave = `${SyncServerFilter} ${SyncClientFilter}`;
```

`tiddlersToServe` would be used to filter the list of tiddlers that goes from the server to the client (and also to filter the list of lazy-load requests), and `tiddlersToSave` would be used to filter which changes are saved from the client to the server. Obviously both the server and the client would have this check (the server in the PUT and DELETE tiddlers route and the client wherever needed).

pmario:
I think this doesn't work. With the default settings the `$:/StoryList` has to be synced in both directions. 

For me it would be OK, If I have a "per user" "add-list", that allows my to tell the server which tiddlers I want to be shipped to the client. The "add-list" may be a second request. I don't have a problem with 2 requests 1 big list with content 1 small list with additional content

Arlen22:
> I think this doesn't work. With the default settings the `$:/StoryList` has to be synced in both directions.

I believe this is correct and should still be the case. If someone doesn't want to sync $:/StoryList they would remove it from SyncFilter so it does not get sent from the server and then add it to SyncClientFilter so it does get saved back to the server.

pmario:
@Arlen22 ... Is this PR still something we should consider?

PR #5385, opened 2021-01-10
contain the long list inside a div which is 50% of vertical height

pmario:
contain the long list inside a div which is 50% of vertical height

### DIFF
--- a/editions/tw5.com/tiddlers/concepts/ShadowTiddlers.tid
+++ b/editions/tw5.com/tiddlers/concepts/ShadowTiddlers.tid
@@ -16,4 +16,4 @@ ShadowTiddlers can be overridden with an ordinary tiddler of the same name. If t
 
 !! Shadow Tiddlers
 
-Shadow tiddlers can be found in the <<click "Right Sidebar : More : Shadows tab">>
+A full list of shadow tiddlers can be found in the sidebar in the "More" -> "Shadows" tab
--- a/editions/tw5.com/tiddlers/concepts/ShadowTiddlers.tid
+++ b/editions/tw5.com/tiddlers/concepts/ShadowTiddlers.tid
@@ -1,15 +1,19 @@
-title: ShadowTiddlers
-tags: Concepts
-modified: 201308091623
+\define actions()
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @pmario I agree that the current listing is very unwieldy with 2400-odd entries. However, this change means that  with some browser/OS combinations one can no longer see that the list is truncated or scrollable:

image

This is Chrome on macOS.

On balance, I think I'd favour removing the list entirely. The rest of the tiddler could do with improvement, but the list really doesn't help.

pmario:
The new commit only shows the modified shadows and creates a button to "open" the shadows in the right sidebar.

pmario:
looks like this:

grafik

pmario:
@Jermolene bump. See new screenshot

pmario:
@Jermolene bump.

pmario:
@Jermolene .. this is a docs change only

editions/tw5.com/tiddlers/concepts/ShadowTiddlers.tid
```diff
@@ -1,11 +1,19 @@
...
 
-The current shadow tiddlers are:
+!! Overridden Shadow Tiddlers
 
-<$list filter="[all[shadows]sort[title]]"/>
+<<list-links "[is[tiddler]is[shadow]sort[title]]">>
+
+!! Shadow Tiddlers
+
+Shadow tiddlers can be found in the <<click "Right Sidebar : More : Shadows tab">>
```
Jermolene, line 19: Apologies, more consistency issues:

* The docs generally refer to the "sidebar" not the "right sidebar"
* The docs frequently refer to a specific sidebar tab without having a button to activate that tab. One might argue that this particular docs tiddler is enhanced by the button, but I think that consistency is more important. If we think that this is a worthwhile improvement then it needs to be implemented consistently across all the docs

I'd recommend removing the macros and changing the last line to:

```
A full list of shadow tiddlers can be found in the sidebar in the "More" -> "Shadows" tab
```

Jermolene (commented):
Thanks @pmario

PR #5397, opened 2021-01-12
Fix broken aria-label in $:/PaletteManager

NicolasPetton:
This PR fixes an `aria-label` attribute in `$:/PaletteManager`.


Jermolene:
Good catch, thanks @NicolasPetton

Issue #5305, opened 2020-12-22
GHSL-2020-362: ReDoS in TiddlyWiki


Jermolene:
I (@Jermolene) have just received the following notification from the GitHub Security Lab team and am posting it here for discussion.

---

# GitHub Security Lab (GHSL) Vulnerability Report: `GHSL-2020-362`

The [GitHub Security Lab](https://securitylab.github.com) team has identified potential security vulnerabilities in [TiddlyWiki](https://github.com/Jermolene/TiddlyWiki5).

We are committed to working with you to help resolve these issues. In this report you will find everything you need to effectively coordinate a resolution of these issues with the GHSL team.

If at any point you have concerns or questions about this process, please do not hesitate to reach out to us at `securitylab@github.com` (please include `GHSL-2020-362` as a reference).

If you are _NOT_ the correct point of contact for this report, please let us know!

## Summary

The project contains one or more regular expressions that are vulnerable to [ReDoS](https://en.wikipedia.org/wiki/ReDoS) (Regular Expression Denial of Service)

## Product

TiddlyWiki

## Tested Version

Latest commit at the time of reporting (December 21, 2020).

### ReDoS

ReDoS, or Regular Expression Denial of Service, is a vulnerability affecting poorly constructed and potentially inefficient regular expressions which can make them perform extremely badly given a creatively constructed input string.

For the specific regular expression reported, it is possible to force it to work with an `O(2^n)` runtime performance when there is [exponential backtracking](http://en.wikipedia.org/wiki/ReDoS#Exponential_backtracking).

ReDoS can be caused by ambiguity or overlapping between some regex clauses. These badly performing regular expressions can become a security issue if a user can control the input. For example if the project is an input validation library, then the project could be used by a server to validate untrusted user input. There is no one size fits all when it comes to fixing ReDoS. But in general it is about removing ambiguity/overlap inside the regular expression.

Before showing the vulnerable regex, it may be helpful to show some examples of regular expressions vulnerable to ReDoS and how to fix them. If you are familiar with this vulnerability and how to fix it, please skip this section.

--- 

```JavaScript
var reg = /<!--(.|\s)*?-->/g;
```

The above regular expression matches the start of an HTML comment, followed by any characters, followed by the end of a HTML comment.
The dot in the regular expression (`.`) matches any char except newlines, and `\s` matches any whitespace.
Both `.` and `\s` matches whitespace such as the space character.
[...]

PR #5401, opened 2021-01-14
Fix backtracking issue with regex

slaymaker1907:
There was an unnecessary qualifier on the regex which was causing exponential backtracking.

I believe the updated regex should not demonstrate this behavior and it does not take a long time when using the example of `..........................................................................` when tested with https://regex101.com/ while the previous version would timeout.

This is in relation to the bug #5305

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -78,7 +78,7 @@ function parseFilterOperation(operators,filterString,p) {
 					nextBracketPos = filterString.indexOf(">",p);
 					break;
 				case "/": // regexp brackets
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @slaymaker1907. With apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)). Many thanks.

PR #5413, opened 2021-01-15
getCacheForTiddler can cache falsy values

flibbles:
`getCacheForTiddler` should be able to cache falsy values just like `getGlobalCache` can. Otherwise it's inconsistent, and a little annoying when my initializer keeps reevaluating an empty string.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -844,7 +844,7 @@ exports.clearGlobalCache = function() {
 exports.getCacheForTiddler = function(title,cacheName,initializer) {
 	this.caches = this.caches || Object.create(null);
 	var caches = this.caches[title];
-	if(caches && caches[cacheName]) {
+	if(caches && $tw.utils.hop(caches,cacheName)) {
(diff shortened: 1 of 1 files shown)


flibbles:
If people like, I can change this from
```
if(caches && $tw.utils.hop(caches,cacheName)) {
    
```
to
```
if(caches && caches[cacheName] !== undefined) {
    ....
```
which will make this change not introduce any performance overhead, but I do strongly feel that this PR should be merged. I know I use the cache and frequently initialize a value like `""` or `null` all the time, except those values need to be initialized all the time.

flibbles:
Actually, the more I think about it, the more I vastly prefer `caches[cacheName] !== undefined`. It gives the initializer method an opportunity to say, "Don't cache my return value. My value isn't defined yet and I'll need to try again when called." by returning "undefined".

I'd might argue that the getGlobalCache method should do the same.

Either way, this PR is still good to go.

Jermolene:
Thanks @flibbles

PR #5414, opened 2021-01-16
Focus-Preservation over refresh cycles

BurningTreeC:
This PR preserves the focus over refresh cycles in

- render.js
- windows.js
- modal.js

It adds a `$tw.focusManager` module that does the handling
It stores selections if the widget owning the focused domNode is an editor widget and restores those selections after refresh
~~If widgets change their positions in the widget-tree over refresh-cycles this mechanism doesn't work~~


kookma:
Hi @BurningTreeC - I hope this resolve the issue of losing focus specially for keyboard shortcuts.

BurningTreeC:
I've found that it's more reliable to store the widgets' stateQualifier plus its `transclusionFootprint` like its generated in #4113 and to restore focus to the widget that has that same `stateQualifier` plus `transclusionFootprint` identifier - instead of generating a full `renderTreeFootprint` from the rootWidget to the affected widget

My finding is much more resource-hungry though because when looking for the affected widget after refresh it generates that identifier for each widget on its way to the widget that has the matching identifier. But it makes it possible - for example - to keep the focus within the editor when opening/closing the editor preview using the keyboard shortcut <kbd>alt-P</kbd>

Another thing that we need is a mechanism to set focus when there's no widget found
- which widget should get the focus?
- should we store a variable of the currently focused element in a state tiddler before refresh?

Jermolene:
Hi @BurningTreeC how is this going? I haven't analysed the code in detail at this point, but I'm very interested to hear what you're discovering.

BurningTreeC:
Hi @Jermolene, thanks for coming back to this

The current state is, that the code will always find a widget and always restore focus after refresh (but not necessarily to the same widget / domNode)
The widget that owns the domNode with focus is first stored in a `widgetInfo` object of the following form:
```
{
widget: widget,
index: domNodeIndex,
qualifier: the state qualifier of the widget + its generated renderTreeFootprint
widgetIndex: the index of the widget amongst the children of its parentWidget
}
```

If the found widget is an editor widget, also its selectionStart and selectionEnd values are stored

Then, the FULL `renderTreeFootprint` of the found widget is generated.
Widgets also have their own `generateRenderTree` method which generates a footprint up to the first found widget that changes the `transclusion` variable.
After refresh, we detect the widget/domNode that should get focus in two ways:
- first, we're looking for a widget that has the same state qualifier plus renderTreeFootprint as the widget found before refresh. This combination has the form: `-8349201932_0_0_2_0_1`. I found that it's more reliable to do this first.
- then, if no widget was found, we use the FULL `renderTreeFootprint` to look for a widget and in case that there's no widget found at that position in the render tree, we use the last found widget

Now the domNode of the found widget can still be problematic, it can be a hidden node, a text-node, a span of a keyboard widget or a widget without domNodes. While true, we look for other nearby widgets

When we have a widget that matches all criterias the focus gets restored to the domNode with the domNodeIndex or - if there's no such domNode - to the first domNode of its domNodes array. If the widget is an editor widget, its selections get restored.

BurningTreeC:
Hi @Jermolene, I'm a bit in a dead end here. Do you have any new ideas how this should be done?

Jermolene:
> Hi @Jermolene, I'm a bit in a dead end here. Do you have any new ideas how this should be done?

Hi @BurningTreeC great to see you back, I hope you are well.

Your logic sounds right, I'll give it a run through as soon as I can.

BurningTreeC:
Hi @Jermolene, yes I was away from TiddlyWiki for quite some time, thank you for welcoming me back :)

Issue #5366, opened 2021-01-04
[BUG] Crash on Node.js with ActionSetField when type field is given a value upon new tiddler creation


xcazin:
When the ActionSetField widget is used on a missing tiddler, it usually creates it, then sets the provided fields to their requested values. In Node.js mode, it instead crashes when given an explicit value for the type field.

1. Start a Node.js instance of TiddlyWiki 5.1.23 in a terminal, like `node ~/src/TiddlyWiki5/tiddlywiki.js ~/wikis/100000000000000/ --verbose --listen port=10000 host=0.0.0.0 tls-key=../tls/key.pem tls-cert=../tls/server.crt`
2. Create a tiddler with the following content: `<$button><$action-setfield $tiddler="A probably missing tiddler" type="text/plain"/>Crash test</$button>`
4. Click on the button, and experience the crash in the terminal:
``` 
...
syncer-server-filesystem: Dispatching 'save' task: A probably missing tiddler 
internal/fs/utils.js:781
  throw new ERR_INVALID_ARG_TYPE(
  ^

TypeError [ERR_INVALID_ARG_TYPE]: The "data" argument must be of type string or an instance of Buffer, TypedArray, or DataView. Received undefined
    at Object.writeFile (fs.js:1436:5)
    at Object.exports.saveTiddlerToFile ($:/core/modules/utils/filesystem.js:414:6)
    at $:/plugins/tiddlywiki/filesystem/filesystemadaptor.js:85:13
    at FileSystemAdaptor.getTiddlerFileInfo ($:/plugins/tiddlywiki/filesystem/filesystemadaptor.js:72:2)
    at FileSystemAdaptor.saveTiddler ($:/plugins/tiddlywiki/filesystem/filesystemadaptor.js:81:7)
    at SaveTiddlerTask.run ($:/core/modules/syncer.js:597:27)
    at Syncer.processTaskQueue ($:/core/modules/syncer.js:503:9)
    at Function.<anonymous> ($:/core/modules/syncer.js:90:9)
    at $tw.Wiki.exports.dispatchEvent ($:/core/modules/wiki.js:133:13)
    at $:/core/modules/wiki.js:166:10 {
  code: 'ERR_INVALID_ARG_TYPE'
}
...
```
FWIW, this has been tested with Node.js version v14.14.0 and v15.5.0

Jermolene:
[...]

PR #5415, opened 2021-01-16
Fix fs adaptor callbacks

joshuafontany:
A collection of fixes for the syncer and filesystemadaptor modules.

* ~~Fixes https://github.com/Jermolene/TiddlyWiki5/issues/5366, where a missing text field would cause a crash or a 0 byte file to be written.~~
* ~~Adds a Try/Catch block to saveTiddlerFileSync() so that it can return an error if it cannot save the file.~~
* Cleans up the method signatures for all the syncer task calls to the active syncadaptor so that the callback function is always last, and the syncer always sends the "current wiki" to the syncadaptor
`methodName(tiddlerOrTitle, optionsObj, callbackFunction(){})`
* Add catches to the stock syncadaptor modules to detect any parameters passed in the "older order" (title, callback, options) and convert them to the correct objects for the method.
* Adds a few callback functions that pass the `adaptorInfo` back to the syncer-callback that needs it.

This has seriously helped me start debugging the Bob MultiWikiAdaptor. @inmysocks was previously completely disregarding the $tw.syncer object and then manually changing the changeCount each time to prevent the syncer from entering a loop.

With these changes, I can "attach" a  new Syncer() to each of the `$tw.Bob.Wikis[index].wiki` objects as `$tw.Bob.Wikis[index].syncer`, and passing the `$tw.Bob.Wikis[index].wiki` object to its constructor. As these all call a single MultiWikiAdaptor object, I needed the syncer to identify the wiki it was syncing. I've made a lot of progress with Bob after these changes, but am not all the way through fixing Bob's filesystem model. As such, these changes could use more testing with the normal Node.js filesystemadaptor, tiddlywebadaptor, etc.


core/modules/utils/filesystem.js
```diff
@@ -459,8 +479,11 @@ Delete a file described by the fileInfo if it exits
 */
 exports.deleteTiddlerFile = function(fileInfo, callback) {
 	//Only attempt to delete files that exist on disk
-	if(!fileInfo.filepath || !fs.existsSync(fileInfo.filepath)) {
-		return callback(null);
+	var invalid = !fileInfo.filepath || !fs.existsSync(fileInfo.filepath);
+	if(invalid) {
```
pmario, line 483: did you add a variable here for easier debugging?

core/modules/utils/filesystem.js
```diff
@@ -491,20 +524,20 @@ exports.cleanupTiddlerFiles = function(options, callback) {
...
 	title = options.title || "undefined";
 	if(adaptorInfo.filepath && bootInfo.filepath && adaptorInfo.filepath !== bootInfo.filepath) {
-		return $tw.utils.deleteTiddlerFile(adaptorInfo, function(err){
+		$tw.utils.deleteTiddlerFile(adaptorInfo, function(err){
 			if(err) {
 				if ((err.code == "EPERM" || err.code == "EACCES") && err.syscall == "unlink") {
 					// Error deleting the previous file on disk, should fail gracefully
-					$tw.syncer.displayError("Server desynchronized. Error cleaning up previous file for tiddler: "+title, err);
-					return callback(null);
+					$tw.syncer.displayError("Server desynchronized. Error cleaning up previous file for '"+title+"'", err);
```
pmario, line 531: I think, the existing error message, makes it more obviour, that the `title` is dynamic. So I would prefer:

```
$tw.syncer.displayError("Server desynchronized. Error cleaning up previous file for tiddler: \""+title+"\"", err);
```

editions/dev/tiddlers/from tw5.com/moduletypes/SyncAdaptorModules.tid
```diff
@@ -151,8 +153,12 @@ Delete a tiddler from the server.
 
 |!Parameter |!Description |
 |title |Title of tiddler to be deleted |
-|callback |Callback function invoked with parameter `err` |
 |options |See below |
+|callback |Callback function invoked with parameter `err` |
+
+!!! Options
+
+<<.from-version "5.1.24">> The signature of syncadaptor functions that accept callbacks has been changed so that the callback is always the last argument. A check for the old order of arguments means that this change should be backwards compatible. The new order should be prefered when updating or writing new plugins.
```
pmario, line 161: +1

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -36,8 +36,9 @@ FileSystemAdaptor.prototype.isReady = function() {
 
 FileSystemAdaptor.prototype.getTiddlerInfo = function(tiddler) {
 	//Returns the existing fileInfo for the tiddler. To regenerate, call getTiddlerFileInfo().
+	var self = this;
 	var title = tiddler.fields.title;
-	return this.boot.files[title];
+	return self.boot.files[title];
```
pmario, line 41: hmmm, Doesn't make too much sense for me. IMO `this.` is good enough.

pmario: In [line 20](https://github.com/Jermolene/TiddlyWiki5/pull/5415/files#diff-ccebed3949d890544181387dc9b9cde2a9f2f81870103a6de069f480ccca97abL20) we should remove `var self = this`  since the variable isn't used and only wastes time and memory

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -46,12 +47,13 @@ Return a fileInfo object for a tiddler, creating it if necessary:
...
 
-The boot process populates this.boot.files for each of the tiddler files that it loads.
+The boot process populates self.boot.files for each of the tiddler files that it loads.
 The type is found by looking up the extension in $tw.config.fileExtensionInfo (eg "application/x-tiddler" for ".tid" files).
 
-It is the responsibility of the filesystem adaptor to update this.boot.files for new files that are created.
+It is the responsibility of the filesystem adaptor to update self.boot.files for new files that are created.
 */
 FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
+	var self = this;
```
pmario, line 56: here it's OK since the var is used later in the code.

pmario: BUT is it necessary @Jermolene what do you think?

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -61,51 +63,55 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
...
 			if(err) {
 				if ((err.code == "EPERM" || err.code == "EACCES") && err.syscall == "open") {
-					var bootInfo = self.boot.files[tiddler.fields.title];
-					bootInfo.writeError = true;
-					self.boot.files[tiddler.fields.title] = bootInfo;
-					$tw.syncer.displayError("Sync for tiddler [["+tiddler.fields.title+"]] will be retried with encoded filepath", encodeURIComponent(bootInfo.filepath));
+					fileInfo = fileInfo || self.boot.files[tiddler.fields.title];
+					fileInfo.writeError = true;
+					self.boot.files[tiddler.fields.title] = fileInfo;
+					$tw.syncer.logger.log("Sync failed for '"+tiddler.fields.title+"' and will be retried with encoded filepath", encodeURIComponent(fileInfo.filepath));
```
pmario, line 100: I personally would prefer escaped double quotes instead of single quotes.

plugins/tiddlywiki/savetrail/savetrail.js
```diff
@@ -99,7 +99,14 @@ SaveTrailSyncAdaptor.prototype.getTiddlerInfo = function(tiddler) {
...
 */
-SaveTrailSyncAdaptor.prototype.saveTiddler = function(tiddler,callback) {
+SaveTrailSyncAdaptor.prototype.saveTiddler = function(tiddler,options,callback) {
+	if(!!callback && typeof callback !== "function"){
+		var optionsArg = callback;
+	}
+	if(typeof options === "function"){
+		callback = options;
+		options = optionsArg || {};
+	}
```
pmario, line 109: Please add 1 or 2 lines of code-comments, that this is done for compatibility reasons. Right? Starting with 5.1.24.

plugins/tiddlywiki/savetrail/savetrail.js
```diff
@@ -112,14 +119,28 @@ SaveTrailSyncAdaptor.prototype.saveTiddler = function(tiddler,callback) {
...
 */
-SaveTrailSyncAdaptor.prototype.loadTiddler = function(title,callback) {
+SaveTrailSyncAdaptor.prototype.loadTiddler = function(title,options,callback) {
+	if(!!callback && typeof callback !== "function"){
+		var optionsArg = callback;
+	}
+	if(typeof options === "function"){
+		callback = options;
+		options = optionsArg || {};
+	}
```
pmario, line 129: same comment here pls

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -182,7 +182,14 @@ TiddlyWebAdaptor.prototype.getSkinnyTiddlers = function(callback) {
...
 */
-TiddlyWebAdaptor.prototype.saveTiddler = function(tiddler,callback) {
+TiddlyWebAdaptor.prototype.saveTiddler = function(tiddler,options,callback) {
+	if(!!callback && typeof callback !== "function"){
+		var optionsArg = callback;
+	}
+	if(typeof options === "function"){
+		callback = options;
+		options = optionsArg || {};
+	}
```
pmario, line 192: inline comment pls

pmario (changes requested):
Are we able to automatically test those functions?

plugins/tiddlywiki/filesystem/filesystemadaptor.js
pmario: **Did you add the `self.xxx` .. for consistency, to make it easier to copy/paste code??**

saqimtiaz:
> * Cleans up the method signatures for all the syncer task calls to the active syncadaptor so that the callback function is always last, and the syncer always sends the "current wiki" to the syncadaptor
>   `methodName(tiddlerOrTitle, optionsObj, callbackFunction(){})`

Is this for Bob compatibility? If so, I would strongly suggest having a separate PR for this to facilitate testing.

joshuafontany:
@pmario @saqimtiaz Thank you for the feedback! I'll be able to review it later today

joshuafontany:
@pmario @saqimtiaz @Jermolene 
I think handling the bullet points as separate PRs is the way to go. I have marked this one as draft to reference it while I create the others. Thanks!

core/modules/utils/filesystem.js
joshuafontany: Yes, I should put that back into the if, thank you.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
joshuafontany: Thank you. I think I was trying to catch an accidental "replace all". Agree.

joshuafontany:
> > * Cleans up the method signatures for all the syncer task calls to the active syncadaptor so that the callback function is always last, and the syncer always sends the "current wiki" to the syncadaptor
> >   `methodName(tiddlerOrTitle, optionsObj, callbackFunction(){})`
> 
> Is this for Bob compatibility? If so, I would strongly suggest having a separate PR for this to facilitate testing.

Yes. Would you prefer to see the `syncer` changes separate from the `syncadaptor` changes? I have moved the other bugfixes to a targeted PR.

Jermolene:
Hi @joshuafontany I'm late to this. I'm not a huge fan of the change to the sync adaptor method signatures. It seems to mostly so that we can pass a wiki reference to each of the methods, but given that the wiki won't change for the life of the syncadaptor, might it not make more sense to assign the wiki during syncadaptor initialisation? I think it's done this way for compatibility with Bob?

joshuafontany:
@Jermolene Thank you (no prob on being a bit late). Yes, this was an experiment that would allow the single instance of the "MultiWikiAdaptor" in Bob (which writes files to disk) to understand which wiki to query for the FileSystemPath/Extension tiddlers. I found a couple of other unrelated bug fixes, which I will spin out into another PR. I also think changing the order of the options & callback so that callback are always last is worth its own PR.

As to identifying the wiki on every syncadaptor call, I agree that in vanilla node this doesn't make sense. I would be open for a different approach to the unique Bob situation. While reading through @inmysocks 's code I noted he had to step around almost all of the syncer/syncadaptor logic to get his Websocket Connections to work - and each websocket message would identify the wiki affected to the server.

I have had success in cleaning up the task/message flow in Bob so that I can give each wiki in memory on the server its own Syncer instance. At the moment, all these point to a single MultiWikiAdaptor instance. Then, incoming websocket-messages use the native TW api to add/delete/load a tiddler, and the Syncer hands the file-writing task off to the MultiWikiAdaptor.

I suppose I could pull all of the websocket logic out of the "adaptor" class on the server and just instance a filesystemAdaptor for each wiki in memory - but my first instinct was to stick with @inmysocks 's desire to load native TW5 code as few times as possible when multiple wikis are in active memory on the server.

Definitely open to suggestions. :)

Jermolene:
> I suppose I could pull all of the websocket logic out of the "adaptor" class on the server and just instance a filesystemAdaptor for each wiki in memory - but my first instinct was to stick with @inmysocks 's desire to load native TW5 code as few times as possible when multiple wikis are in active memory on the server.

Sharing the same syncadaptor instance between multiple syncers seems like an optimisation that might constrain us in the future. Other syncadaptors that I've made have maintained state information that wouldn't work if they were shared.

Generally, I'd be happy to strip back and clean up more of the synchronisation implementation. In particular, it would be useful to be able to have multiple syncers attached to a wiki, each syncing a different subset of tiddlers.

joshuafontany:
> Sharing the same syncadaptor instance between multiple syncers seems like an optimisation that might constrain us in the future. Other syncadaptors that I've made have maintained state information that wouldn't work if they were shared.
> 
> Generally, I'd be happy to strip back and clean up more of the synchronisation implementation. In particular, it would be useful to be able to have multiple syncers attached to a wiki, each syncing a different subset of tiddlers.

@Jermolene Thanks. I have split off the bugfixes & the order of parameters change (leaving out the wiki obj) to separate PRs.

I am experimenting with a deep refactor of Bob's multi-wiki & multi-user logic. I hope to have the base of something scalable (as in later on updates may allow a large number of users connected to a cluster of Bob instances behind a load-balancer). That means considering things like user-sessions: tracking a user over multiple tabs, each with different wikis, er "rooms", open, each with multiple possible websocket connections in series - as they reload the tab or reconnect. I think a 1:1 syncer-syncadaptor pair would work better all around there.

joshuafontany:
I will probably close this once I have my 'bobdev' branch up to parity by merging the separate PRs & cleaning up.

Issue #4263, opened 2019-09-15
Call for thoughts on improving plugin handling


Jermolene:
**This ticket is a call for input. I'll update this post with any suggestions**

The work for adding support for dynamic loading/unloading of plugins in #4259 requires touching many parts of the core concerned with plugin handling. I'd like to take the opportunity to address a few other long standing issues/limitations that are in the same area.

For example, in the control panel plugin listing and plugin library we need to add an indication of whether a plugin contains JS modules and hence will require a restart if installed or removed. I'd like to take the opportunity to address one of the concerns in [Improve the plugin dropdown information](#4086) by elevating the portion of the title after the final slash to be the shorthand name of a plugin

Other existing issues that might be worth revisiting:

* [Support `plugins/publisher` in wiki folders](#1309)
* [Move plugin library from modal to tiddler](#3208)
* [Proposal: The $ControlPanel provides possibilities for plugins to promote important content](#4125)

pmario:
@Jermolene ... There is something wrong with your links.

Jermolene:
Thanks @pmario

twMat:
* [quote] "Move plugin library from modal to tiddler"<br>+1 ...and make it take substantially fewer steps to reach it.<br>...and, related, prepare plugin library for Federation i.e to show multiple libraries.
* I'd like a simple plugin packer. Drag'n drop titles onto a dropzone and they become a plugin. Plus some additional post-editing options (remove, add, edit shadows).
* I want update alerts, i.e to know if I have the latest version of a plugin ideally both when I look at the plugin list but also alerts
* (See image below)
  1. The open/close slider is unclear and easy to miss. One idea is to put the arrow to the left of the box shadow... (and make the button have a border itself?)
   2. The "technical" path title should not be seen here, it should be when the slider is opened
   3. Version should be much bigger
   4. It is confusing that clicking on this area is the same as clicking on the title. Clicking on this area should possibly open the slider, if anything. Only by clicking the blue link should one go to the actual tiddler. 

![image](https://user-images.githubusercontent.com/4534056/64974440-20a2e900-d8ad-11e9-9f9b-adf4cee8b674.png)

twMat:
[ref](https://groups.google.com/d/msg/tiddlywiki/Fdi955ap4a4/pTX_tSDYAQAJ)

Harmonize appearance of plugin library titles with local listing of titles, i.e 
[...]

Issue #4125, opened 2019-07-23
Proposal: The $ControlPanel provides possibilities for plugins to promote important content


pmario:
This is a "meta-issue" to track the progress for: **The $ControlPanel provides possibilities for plugins to promote important content**

Also see the discussion at: Captions should visually replace title everywhere #4081

 - [ ] clean up $:/ControlPanel: Settings tab. 
 - [ ] provide a prominent section, where plugin authors can promote their plugins
 - [ ] provide a possibility to get attention for important plugin related content.

AnthonyMuscio:
Mario,

Good initiative. In the absence of standards I have played myself with introducing control panel tabs.

What follows is what I have considered such that it may inform this discussion.

I introduced control panel tabs, such as 
- informational tabs (Control Panel > Info >tabname) on the Current Wikis Hosting, Storage Settings
- and one for my "Brand"  (Control Panel > Info > Brandname) , the idea of the brand tab is to display meta information such as PSaT plugins, macros and settings installed, this I keep the number of tabs low, IE not one for each plugin/solution.
- I am working on user related login and out and Have Used Control Panel > User
- Where necessary I would also consider placing solution specific tabs under  Control Panel > Settings > solutionname, including settings, config, debug and override settings etc... many of which would be repeated in my brand name tab.

Perhaps there could be a tab that contains a set of vertical tabs called Control Panels (3rd Party Control Panels), that any designer can use a tag to add their own control panels, Then have a search appear in and at the top of this panel a focused search including the propose keywords. 

A matching and cross referenced standard should also guide the use if the Tiddler Info Panel Tabs.

Regards
Tony

AnthonyMuscio:
Minor edit to previous

pmario:
I think every plugin should have its own edition, which contains documentation and examples. 

The documentation, that comes with the plugin needs to be as minimal as possible. So maximum here is a "reference - doc". 

plus a link to the plugin edition 
plus a link to the authors home page.

AnthonyMuscio:
Mario

I agree with the minimal as posible however there are cases where a little more information improves the use and functonality a great deal. I think the key is some simple standards that tells designers where they can expose additional information and the nature of this info. Eg cpanel vs info panel etc..

Yes
plus a link to the plugin edition
plus a link to the authors home page.

[...]

PR #5418, opened 2021-01-18
Translateable plugins - first try

pmario:
(This info looks much more complicated than it actually is)

If you download this branch you can do the following things: 

# Pluginlibrary Server - NEW

- **Edit 2022-04-03**: This is a new PR now: **Pluginlibrary local test edition** #6586 
  - go to `./editions/pluginlinlibrary-server` ... new directory ... CAN be a separated PR (@Jermolene ??)
    - call `npm test` ... this will create a local library version and run a server on http://localhost:8080
    - It will allow you to test load plugins to the server
    - There is a "Delete all "plugin modules"", which will remove all plugins.
      - see: http://localhost:8080/#Remove-all-plugin-modules
    - call `npm start` ... if you want to skip the "build library" step

# Switch language for plugins

 - You need to install the TW core "menubar" plugin  to test this branch
 - You need to install the TW core "de-DE" language to test this branch
 - This PR contains a POC translation to **de-DE**. 
   - see file: `menubar.multids`
   - ~~Only 2 tiddlers are translated atm.~~ 
   - Contents (caption) -> Inhalt ... So it's easy to test and see if it works, since the Contents text is seen by default top left.
   - Contents (description) -> This text is part of the **plugin : config** tab
     - more commits will follow to translate everything

# Structure

## initial language

 - The initial language is part of the plugin
   - That's important, so there always will be a "fallback" text, if any text translation is missing
   - Technically is a "language" 
 - I did use the directory name `_en-GB` the underscore guarantees, that the directory is listed first place. 
    - The directory name doesn't matter. 
    - It's only the "namespace", "author" and "plugin-name" that matters: eg:
      - `title: $:/language/tiddlywiki/menubar` 
      - `title: $:/language/<author>/<plugin-name>`
    - Important: `/<author>/<plugin-name>` are used for AddOns too!

## plugin: `plugin.info`

There is a new parameter, which indicates, that a plugin is "ready to be translated". 

  - `addons": ["language-addon"]` .. it's an array. ATM only `language-addon` is supported!
     -  as Jeremy suggested, it may be possible to create `theme-addon` plugins, **not part of this PR!**
  - This parameter is part of the plugin: "plugin-info"
  - This parameter is used in the add-plugin-UI to show a new button.

## language: `plugin-info`

The translation `plugin-type` must be `language-addon` and is part of the `languages` namespace. see:

{
	"title": "$:/languages/de-DE/tiddlywiki/menubar",
	"name": "Menü Toolbar/de-DE",
	"plugin-type": "language-addon",
	"description": "Top Menü Plugin",
	"author": "MarioPietsch aka pmario",
	"core-version": ">=5.1.24"
}

 - The **title** and **plugin-type** are important to identify the addon
 - The language plugin title is the ONLY tiddler, that contains the "language string" eg: **de-DE**
 - the `title` must have the structure: 
   - `$:/languages/de-DE/tiddlywiki/menubar`
   - `<language-string>/<author>/<plugin-name>`
     - the `<language-string>` is used by the language-switcher UI and is stored in [$:/language](http://localhost:8080/#%24%3A%2Flanguage) tiddler
     - It's used to identify the "root" language namespace. 
     - Everything after the `<language-string>` is part of the "addon-namespace"

## Video

https://youtu.be/yZMaZ4qeaKk


core/language/en-GB.tid
```diff
@@ -2,6 +2,8 @@ title: $:/languages/en-GB
 name: en-GB
 description: English (British)
 author: JeremyRuston
-core-version: >=5.0.0"
+core-version: >=5.0.0
+plugin-type: language
+type: application/json"
```
pmario, line 7: the last 2 elements are needed, otherwise the "pluginswitcher.js" doesn't activate addons for en-GB. .. Those dependencies are hardcoded atm

core/modules/startup/startup.js
```diff
@@ -91,6 +91,14 @@ exports.startup = function() {
...
 		}
 	});
+	$tw.languageAddonSwitcher = new $tw.PluginSwitcher({
+		wiki: $tw.wiki,
+		pluginType: "language-addon",
+		controllerTitle: "$:/language",
+		defaultPlugins: [
+			"$:/languages/en-GB"
+		]
+	});
```
pmario, line 101: new `pluginType` added, so the core can activate AddOns

core/ui/ControlPanel/Plugins.tid
```diff
@@ -16,4 +16,9 @@ caption: {{$:/language/ControlPanel/Plugins/Caption}}
 
 <<lingo Installed/Hint>>
 
-<$macrocall $name="tabs" tabsList="[[$:/core/ui/ControlPanel/Plugins/Installed/Plugins]] [[$:/core/ui/ControlPanel/Plugins/Installed/Themes]] [[$:/core/ui/ControlPanel/Plugins/Installed/Languages]]" default="$:/core/ui/ControlPanel/Plugins/Installed/Plugins" explicitState="$:/state/tab--86143343"/>
+<$macrocall $name="tabs"
+tabsList="[[$:/core/ui/ControlPanel/Plugins/Installed/Plugins]]
+[[$:/core/ui/ControlPanel/Plugins/Installed/Themes]]
+[[$:/core/ui/ControlPanel/Plugins/Installed/Languages]]
+[[$:/core/ui/ControlPanel/Plugins/Installed/LanguagesAddOns]]"
```
pmario, line 23: This tab will need a new expandable "group", so users can filter for a base language. eg: "de-??"

editions/pluginlibrary-server/tiddlers/$__DefaultTiddlers.tid
```diff
@@ -0,0 +1,6 @@
+created: 20131127215321439
+modified: 20210117002817063
+title: $:/DefaultTiddlers
+type: text/vnd.tiddlywiki
+
+[list[$:/StoryList]]
```
pmario, line 6: **System and Config tiddlers will NOT be part of the final PR. ... BUT the make testing this branch simpler!!!**

plugins/tiddlywiki/menubar/plugin.info
```diff
@@ -2,5 +2,7 @@
 	"title": "$:/plugins/tiddlywiki/menubar",
 	"name": "Menu Bar",
 	"description": "Menu Bar",
-	"list": "readme config"
+	"list": "readme config license",
+	"plugin-priority": "99999",
```
pmario, line 6: ATM plugins with an existing translation will need a `plugin-priority`, otherwise the add-on will not be activated. @Jermolene it would be nice to avoid this new parameter, but I couldn't find a good way.

pmario:
I did upload a video and update the OP

https://youtu.be/yZMaZ4qeaKk

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -2296,10 +2296,12 @@ html body.tc-body.tc-single-tiddler-window {
 	background: <<colour background>>;
 }
 
+.tc-plugin-info-sub-language-indicator,
```
pmario, line 2299: TODO - Don't forget to add them everywhere, where needed

languages/de-DE-menubar/menubar.multids
```diff
@@ -2,3 +2,18 @@ title: $:/language/tiddlywiki/menubar/
 
 items/contents/caption: Inhalt
 items/contents/description: Inhaltsverzeichnis
+items/hamburger/caption: Manü-Icon
```
BurningTreeC, line 5: M**e**nü?

core/modules/commands/savelibrarytiddlers.js
```diff
@@ -77,6 +77,9 @@ Command.prototype.execute = function() {
 			iconText = iconTiddler.text,
 			iconContent;
 		if(iconType && iconText) {
+			if (iconType === "image/svg+xml") {
+				iconText = iconText.replace("<svg ", "<svg xmlns=\"http://www.w3.org/2000/svg\" ")
```
Jermolene, line 81: Why is this needed?

core/modules/pluginswitcher.js
```diff
@@ -47,16 +47,40 @@ PluginSwitcher.prototype.switchPlugins = function() {
 	}
 	// Accumulate the titles of the plugins that we need to load
 	var plugins = [],
+		unregisteredTiddlers = [],
+		registeredTiddlers = [],
 		self = this,
+		accumulateLanguageAddOn = function(title) {
```
Jermolene, line 53: We use the pluginswitcher class for both themes and language switching. Can't we keep things generic so that we get theme addons for free?

Jermolene (commented):
Thanks @pmario.

core/modules/commands/savelibrarytiddlers.js
pmario: There is a super tricky problem: 

In line 79 there is `if(iconType && iconText) {` ... which tests the icon-tiddler for the `iconType`. .. BUT most TW icons don't have a `type` field, because we don't need it. .. So the plugin icon will _never ever_ be used in the "add plugins" modal dialogue. 

The attempt to fix this was: I did add a type field to my icons. ... BUT ... If icons have a type-field the existing "plugins tab" in the control panel doesn't show the icon anymore. 

So the "fix above" won't work anyway. ... I'll need to create a completely separate Issue for this one. 

----------

Adding a type field to svg-icons doesn't fix the problem. SVG's need to have the `xmlns=\"http://www.w3.org/2000/svg\"` attribute to `iconContent = $tw.utils.makeDataUri(iconText,iconType);` work as desired.  .... BUT ... I don't want to add this attribute. 

SO we have a "chicken / egg" problem here. That's why we need a new **independent issue**

core/modules/pluginswitcher.js
pmario: You are right: I also don't like this part of the new code too much. 

--- 

Themes and languages are technically  plugins, BUT the internal handling is slightly different. ... Those differences cause problems here. 

A language-add on will probably need different handling than themes. .. Themes are closer to plugin handling.

core/modules/commands/savelibrarytiddlers.js
pmario: see: https://github.com/Jermolene/TiddlyWiki5/pull/5418/commits/1c6e6e8f4a1d05aa2ae3dad97ce7ff3520cef849

linonetwo:
@pmario Will you consider reworking on this? Recently some of the plugin users request me to translate my plugins to English, but I don't know a good i18n framework in tw, I think this may be a hope. Thanks!

pmario:
Hi, 
I think the directory structure needed to create translatable plugins is pretty solid and simple enough to maintain it. 

The show stopper (for me) at the moment is the plugin import UI. The core UI for the plugin importer is one of the most complex wikitext code we have in the system. It wasn't designed to handle translatable plugins. 

The way it is implemented in this PR, feels a bit hacky, that's why it is still a draft. I would have to have a look, how and if it still works with the "cascade" mechanism. ... May be using cascades could help us to simplify the plugin importer UI. ... 

I think I can try to remove the conflicts, so you can download it and play with it, if it would work for you. ... 

I also have to have a closer look at this discussion at Talk: https://talk.tiddlywiki.org/t/translating-tw-documentation-workflow-proposal/2440/10 ... It may have overlapping "interests"

There are some other related discussions here at gitHub: 

- A new type of plugins for distributing contents like course notes, presentations, ordinary bundle of tiddlers #5462
- Call for thoughts on improving plugin handling #4263 
- Proposal: The $ControlPanel provides possibilities for plugins to promote important content #4125 

and probably more.

pmario:
I think the first step will need to check a possibility if we can "simplify" the existing UI in a backwards compatible way to make it more flexible, for the next steps ... I've done this type of things for the core tabs-macro, which is merged for testing now #6578.

Jermolene:
Hi @pmario @linonetwo I'm keen to have translatable plugins in the core.

There's a lot going on in the PR which makes it hard to consider each part separately. It seems to introduce a new plugin type, which seems a very intrusive way of adding the feature. I'd strongly suggest separating things out. What we need to see first is the basic mechanism for translations, and then perhaps worry about the impact on the plugin library user interface separately.

It's not clear whether the mechanism here allows for a plugin to contain integrated translations, rather than relying on a separate add-on plugin?

pmario:
See: **Pluginlibrary local test edition** https://github.com/Jermolene/TiddlyWiki5/pull/6586

pmario:
>It's not clear whether the mechanism here allows for a plugin to contain integrated translations, rather than relying on a separate add-on plugin?

It's a separated language-addOn. I personally only want to have languages installed, that I can read. Users have to decide which languages they want to install.

pmario:
> It seems to introduce a new plugin type, which seems a very intrusive way of adding the feature.

That's caused by the existing core code. The existing core plugin handling is _not_ generic. There are hardcoded differences between plugin, language and theme handling in the core code.  ... Links to the code will follow. I have to find them first

Jermolene:
> > It's not clear whether the mechanism here allows for a plugin to contain integrated translations, rather than relying on a separate add-on plugin?
> 
> It's a separated language-addOn. I personally only want to have languages installed, that I can read. IMO users have to decide which languages they want to install.

No, we need to be able to cope with both translations in separate add-on plugins and integrated directly into the original plugin.

> That's caused by the existing core code. The existing core plugin handling is _not_ generic. There are hardcoded differences between plugin, language and theme handling in the core code. ... Links to the code will follow. I have to find them first

Of course there are - there are clear functional differences in the circuitry used to switch out themes and languages vs. plugins. I think once again what you really want is a subtype of "plugin", not a new type of plugin.

linonetwo:
@pmario Hi Mario, do you have a tutorial about how to use this? Even it is not in the core, I can package it as a plugin and use it as dependency to manage translation addons of my plugin.

linonetwo:
I'm going to use this method, uses plain tiddler shadowing mechanism. Does this mean we don't need a translate framework in tw core?

https://talk.tiddlywiki.org/t/how-to-translate-3rd-party-plugin/5435/6?u=linonetwo

linonetwo:
I don't think `addons": ["language-addon"]` is necessary, because just like sub-plugin mechanism we already have, we can only describe `dependant` on sub-plugin's info file, and plugin library add this information automatically to the main plugin.

And currently some plugins already using sub-plugin to deliver translation, I think it is better only ask them to change sub-plugin's type from "plugin" to "language-addon", not ask them also add this info to the main plugin.

linonetwo:
A simple macro idea: https://talk.tiddlywiki.org/t/how-to-translate-3rd-party-plugin/5435/10?u=linonetwo

Pro: only add 4 lines of code (if add to the core)
Con: localization and other plugin content is in same plugin, no language-pack sub-plugin

PR #5419, opened 2021-01-18
removing redundant title fields from plugins

flibbles:
All the tiddlers described in a plugin tiddler go something like this:
```json
    ...
    "$:/language/ControlPanel/Basics/Tiddlers/Prompt": {
        "title": "$:/language/ControlPanel/Basics/Tiddlers/Prompt",
        "text": "Number of tiddlers"
    },
    ...
```
We use the title as a key, and then we repeat the title as a field. In an empty tiddlywiki, this amounts to 180K of wasted space. Turns out, removing that title field is actually super simple. At first I thought it'd be a serious backward compatibility issue, but then I found the place where we create tiddlers from these JSON objects:
https://github.com/Jermolene/TiddlyWiki5/blob/ae273a08f138dc470cb034cdc9867bb6defc2323/boot/boot.js#L1375-L1383

 The ``tiddler: new $tw.Tiddler(constituentTiddler,{title: constituentTitle}) `` line means we already put in the key as the title in case the field is missing. And we've been doing this as far back as February 2013 (probably farther, but I got tired of git blaming into the past).

This PR does two things:

1. It removes the title field when we're creating pluginInfo JSON string for writing.
2. It puts it back in to the plugInfo JSON object we retain in memory. This isn't actually necessary for the core. It never uses that pluginInfo as far as I can tell, but maybe some plugin somewhere depended on the title being among the 'tiddler' keys.

That's it! However, I completely fine if this PR seems problematic, and you don't want to merge it. I'm close to releasing my TW5-Uglify plugin which while cull these redundant title fields. It's just such an easy optimization that I figured maybe the core would like it too.

If not, it just means TW5-Uglify will seem all the better at compressing.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1289,7 +1289,15 @@ $tw.Wiki = function(options) {
 			var tiddler = tiddlers[title];
 			if(tiddler) {
 				if(tiddler.fields.type === "application/json" && tiddler.hasField("plugin-type") && tiddler.fields.text) {
-					pluginInfo[tiddler.fields.title] = JSON.parse(tiddler.fields.text);
+					var info = JSON.parse(tiddler.fields.text);
+					if (info.tiddlers) {
+						// Put the title in just in case it was omitted for
+						// brevity's sake.
+						for (var title in info.tiddlers) {
+							info.tiddlers[title].title = title;
+						}
+					}
+					pluginInfo[tiddler.fields.title] = info;
 					results.modifiedPlugins.push(tiddler.fields.title);
 				}
 			} else {
@@ -1962,8 +1970,17 @@ $tw.loadPluginFolder = function(filepath,excludeRegExp) {
 			var tiddlers = pluginFiles[f].tiddlers;
 			for(var t=0; t<tiddlers.length; t++) {
 				var tiddler= tiddlers[t];
+				var abridgedFields = Object.create(null);
 				if(tiddler.title) {
-					pluginInfo.tiddlers[tiddler.title] = tiddler;
+					// We copy the fields to abridged fields without the title>
+					// keeping title in pluginInfo would be redundant, since
+					// the object keys are the titles.
+					for(var field in tiddler) {
+						if(field !== 'title') {
+							abridgedFields[field] = tiddler[field];
+						}
+					}
+					pluginInfo.tiddlers[tiddler.title] = abridgedFields;
 				}
 			}
 		}
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1290,10 +1290,10 @@ $tw.Wiki = function(options) {
 			if(tiddler) {
 				if(tiddler.fields.type === "application/json" && tiddler.hasField("plugin-type") && tiddler.fields.text) {
 					var info = JSON.parse(tiddler.fields.text);
-					if (info.tiddlers) {
+					if(info.tiddlers) {
 						// Put the title in just in case it was omitted for
 						// brevity's sake.
-						for (var title in info.tiddlers) {
+						for(var title in info.tiddlers) {
 							info.tiddlers[title].title = title;
 						}
 					}


pmario:
IMO a nice win for mobiles!

kookma:
@Jermolene - please merger this PR.
@flibbles and @pmario - I use TW in my university classes! Most students uses the whole wiki all the time to email, so for us size is a big deal! In my plugins I try to save 1kB, so 180kB is alot! Please reopen this PR! I am sure @Jermolene was very busy and missed this!

kookma:
> The contents would be compatible with the JSON store area, and thus we would be able to use the same wikitext tools that we use for creating the JSON store area to create plugins.

Thanks a million @Jermolene 

This would be great! I think Tiddlywiki should be able to create plugins (also content plugins for distributing ebooks, course notes, slides, product brochures, ...) as simple as exporting bunch of tiddler from $:/AdvancedSearch page!
I have developed alternative to Tinka which is simpler to use specially for nontechnical people, still with recent features in TW 5.123, 5.2.0, I think this workflow can be simpler!

pmario:
>text/vnd.tiddlywiki. That's the default type already, I believe.

Yes. If there is no type-field it uses `text/vnd.tiddlywiki`

pmario:
>@Jermolene: Making them an array, would actually save less space, because the "title" field would be copies into each shadow tiddler. I also suspect that it would require more work trying to maintain backward compatibility. 

>The wonder of the change I'm proposing is that that TiddlyWiki as far back as 2013 can already handle the removal of the title field from the JSON objects.

@Jermolene ... I think this PR should be considered again after v5.2.1 is out. There are 2 good reasons to merge it as it is.

flibbles:
Hey @Jermolene, when you said...

> There is of course some further redundancy: for example, field names like "text" and "type" get repeated in full many times. I did explore ways of encoding fieldnames when I was working on the JSON store area but I think on balance such complex techniques introduce too much friction for external tools.

I take it you considered a "multids" section in plugins, so instead of:
```json
"tiddlers": {
	"$:/prefix/some_title": {
		"text": "...",
		"type": "...",
	},
	...
}
```
You would have...
```json
"tiddlers": {
	"$:/prefix/some_title": {
		"text": "...",
		"type": "...",
	},
	...
},
"multids": {
	"$:/prefix/": {
		"type": "...",
		"text": {
			"title_A": "...",
			"title_B": "...",
			...
		}
	},
	...
}
```
It'd be a section identical to the "tiddlers" section, except the "text" field would be an object of "title_suffix": "text_body" pairs instead of just the string content of a single tiddler.

It would allow all language tiddlers (or any tiddler described in a ".multids" file) to be packed down. Also, since it would allow specifying extra fields that would apply to all tiddlers within an entry, you could have this compact down files describe in a "tiddlywiki.files" file. A multids strategy could apply to many module-types, given how many verbose-but-identical fields they share.

The savings on ".multids" files alone might make it worth it, but also certain uglifying plugins could examine all tiddlers that have identical field profiles (all fields the same except for "text" and "title") and find optimal ways of compacting them. 

But this is a complication. So I'd understand if this was something you'd considered and rejected.

Jermolene:
Hi @flibbles 

> But this is a complication. So I'd understand if this was something you'd considered and rejected.

Almost. When I was working on the JSON store area I experimented with compression techniques. One technique that worked quite well was to make an array of all the unique strings in the tiddler array, and then reference the strings via their numeric index in the `tiddlers` object. It worked reasonably well, but in the end I felt that the increased burden for any tools reading or writing the files was not acceptable.

As I think I've said elsewhere, I am interested in compressible plugins, but I prefer using standard Zip compression to make things easier for external tooling.  `.twplugin` files would be zip files containing the JSON of the plugin.

flibbles:
Ah well, pity. Woulda been really nice to pack up all those language shadow tiddlers, especially given TiddlyWiki's recent push to deprecate data tiddlers. It does make unpacking a plugin kind of a pain.

pmario:
@Jermolene Bump

Jermolene:
> @Jermolene Bump

Thanks @pmario I've added this PR to #6266

flibbles:
Hmm. You're right. Took me forever to figure out how to replicate that bug, but it definitely is there. I could fix it so that the title fields are only removed during write instead of read, but like you said, it could still have unforeseen consequences.

Anyone who still wants this size reduction can always use Uglify, which because of the way it's designed, this problem won't ever come up there.

pmario:
@flibbles hmmm Closing with no fix :( ... I would like to have the space reduction, but I don't want to use uglify. Wouldn't it be simple to fix it?

flibbles:
> Wouldn't it be simple to fix it?

It would probably involve digging into the JSONWidget to intercept plugins as they're being written. I don't imagine it would be hard, but making sure it's done in all the places where a plugin can be written would be a pain (but at least not as error prone, since if we miss one, it's not as big of a deal). It's been over a year since I worked with that part of the code, so I'd have to load it all back into my brain to fix this, which I can do if there's strong interest, I just don't have time for that _right now_. Maybe for a future release.

> but I don't want to use uglify

Oof. I understand. Does it make a difference to know that you can uglify without compressing javascript or wikitext? (I know it doesn't. Don't worry.)

PR #5422, opened 2021-01-19
Fix blank favicon if root-tiddler=save-lazy-*

FlashSystems:
If the `root-tiddler` option is set to `save-lazy-images` or `save-lazy-all` on node.js the favicon is empty because the tiddler `$:/favicon.ico` is lazy loaded and is empty at first.

This patch excludes the `$:/favicon.ico` tiddler from lazy loading and restores the favicon functionality.


FlashSystems:
Crazy. The `is[system]` filter should exclude the favicon in the first place. I'have to reconsider this pull request. Sorry for the interruption.

PR #5423, opened 2021-01-19
Fix blank favicon if root-tiddler=$:/core/save/lazy-images is set

FlashSystems:
If image lazy loading is used with node.js the favicon is blank. The line `-[!is[system]is[image]]` excludes only non system images from begin saved as full tiddlers. But the `[is[image]]` in the `skinnySaveTiddlerFilter` includes all (even system) images as skinny tiddlers. The created HTML file has the favicon tlisted twice. And the skinny tiddler seems to win. This breaks the display of the favicon.

This patch fixes this issue by excluding system images from the skinny tiddlers list.


Jermolene:
Thanks @FlashSystems, that's great.

PR #5430, opened 2021-01-21
Rename TW5-firebase: TiddlyWiki5 for Google Firebase by Peter Neumark.tid

joshuafontany:
Rename file with illegal character in filename


joshuafontany:
This change is also needed in the "tiddlywiki-com" branch, but I'm not sure how to target that...

mklauber (approved):

joshuafontany:
@mklauber Thanks! @Jermolene As mentioned above, this change is also needed in the `tiddlywiki-com` branch.

Jermolene:
If I'm understanding correctly, the problem was the inclusion of `:` character in a filename. The same thing has happened before at least once, so I'm wondering if we should add something to the test suite.

PR #5440, opened 2021-01-24
Fix the right margin of tags used outside of the tags wrapper

NicolasPetton:
Only add a right margin to tag items rendered within the tags-wrapper.
This fixes the margin of tags used inline using for instance the `tag`
macro.

###### Before

Screenshot from 2021-01-24 20-30-13

###### After

Screenshot from 2021-01-24 20-29-52


Jermolene:
Great, thanks @NicolasPetton

PR #5444, opened 2021-01-25
Toolbar macro

Alzacon:
The macro works. It do it in the test of gist in https://github.com/Jermolene/TiddlyWiki5/discussions/5312#discussioncomment-287262 except in EditorToolbar because it needs conditions that they are evaluated in the template. It works in the editor tiddler.

But I will try to get it to work in the editor dropdown.

I have used $x$ for work with parameters, I don't know if there is the better option.

I hope don't break nothing while i do modifications, I never worked using git (and github).


Jermolene:
Hi @Alzacon this looks good. We need documentation, but also it would be nice to switch the existing core toolbars (apart from the editor toolbar) to use these macros.

Alzacon:
Hi @Jermolene 

I think that I explain it really bad.
- The toolbar macro works displaying the toolbars (ViewToolbar, EditToolbar, PageControls and EditorToolbar) and their respective buttons.
- The part of drop down has an error in _line 24_. I have to remove `variable="listItem"` then it.


While I was trying to prepare (and check) the macro implementations I discoverd that the buttons of tiddler controls don't work as expected. I work on it.

Yes, I had planned do the documentation, when the macro works without problems in the implementations and the making of custom toolbar.

Jermolene:
Hi @Alzacon no problem, just shout when it's ready for review.

Alzacon:
Finally, I fixed problems in the buttons. I add a class parameter for customization. There is a basic doc, I don't like much the point of the config parameter.

This file is a WIP of the implementation.
[toolbar-5.1.24.zip](https://github.com/Jermolene/TiddlyWiki5/files/6618871/toolbar-5.1.24.zip)

Alzacon:
Exampled moved to a new tiddler and created with doc macro _.example_

pmario:
@Alzacon ... your repo contains an invalid filename, so it can't be checked out on windows. Would it be possible, that you merge Jermolene/master, which fixed the problem already, to your master.

pmario:
The ZIP file version contains a "visual glitch" at 1 of the icons. 

grafik

Alzacon:
Invalid file name? :fearful: 

About the "visual glitch", I think that it is a problem with the rasterized in the browser and the font-size in the body of tiddler (15px) and the font-size in the class _tc-page-controls_ 1.5em. Have you do zoom? Because I only see it with 150% of zoom. The "original" PageControls are modified in that wiki, do you see anything there?

pmario:
You are right. I did use a browser zoom. On my laptop it's only there with 120% on my PC it's a different value. ... I'll need to think about it.

pmario:
>Invalid filename
It's `editions/tw5.com/tiddlers/community/resources/TW5-firebase: TiddlyWiki5 for Google Firebase by Peter Neumark.tid` that can't be handled by the windows filesystem. That's why I can't check out your branch to my PC. 

Jeremy did fix this problem in master, that's why I did suggest, that you may merge master to your branch to fix the problem.

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
+title: $:/core/macros/toolbar
+tags: $:/tags/Macro
+
+\define toolbar-filter-display(tag, config)
+<$set name="config" filter="[[$config$]match[]]"
```
Jermolene, line 5: We are moving away from using macro parameters via textual substitution (ie `$foo$` syntax) because the technique is brittle to parameter values containing quotes or square brackets.

The preferred alternative is to use the `__parameterName__` syntax, whereby each macro parameter is available as a variable with the same name wrapped in double underscores.

So, the line above could be written:

```
<$set name="config" filter="[<__config__>match[]]"
```

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
+title: $:/core/macros/toolbar
+tags: $:/tags/Macro
+
+\define toolbar-filter-display(tag, config)
```
Jermolene, line 4: The coding style for the core is not to insert spaces after commas in parameter lists

Jermolene (changes requested):
Many thanks @Alzacon, just a couple of points.

We also need to refactor all the existing toolbar buttons to use these macros. We try to avoid macros in the core that we don't use ourselves.

core/wiki/macros/toolbar.tid
Alzacon: Done. I thought i did it.

There are more textual substitution but when filters are declared as variables/macros that alternative doesn't work (for me).

Alzacon:
Changes done

> We also need to refactor all the existing toolbar buttons to use these macros. We try to avoid macros in the core that we don't use ourselves.

Yes, I agree with it.

I thought it best not to do a big pull request. It is the reason for which I prepared a list for refactor. In fact, I incluided it (and some changes) in the uploaded ZIP file.

I have no problem adding the changes in this PR or creating a new PR for them. The changes are half done, I need to fix some things and do the ones that are left undone. But I don't know how long it will take.

Alzacon:
Two question about coding styte becuase I usually use other spaces for readability:
- 2 tabs when it is a continuation of another line without brackets or parentheses without closing
- I left some spaces at the end of some lines in the above cases.
- I also left some spaces before closing some angle brackets but I did it halfway I have to fix it, for example I did it when the end of the call to a macro and the end of the label of a widget coincide (`>>>` to `>> >`)

Do I have to fix them? And how?

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
+title: $:/core/macros/toolbar
+tags: $:/tags/Macro
+
+\define toolbar-filter-display(tag,config)
```
pmario, line 4: needs a `\whitespae trim` here

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+<$list filter=<<toolbar-filter-display "$tag$" "$config$">> variable="listItem">
+	<$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]] [<__class__>]" >
+		<$transclude tiddler=<<listItem>>/>
+	</$set>
+</$list>
+\end
+
+\define buttons-with-template(tag,config,class,template)
+\whitespace trim
+<$list filter=<<toolbar-filter-display "$tag$" "$config$">> >
```
pmario, line 23: I think this `>> >` is OK for readabilty

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+	<$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]] [<__class__>]" >
+		<$transclude tiddler="$template$" />
```
pmario, line 25: The extea spaces at the end in the 2 lines above imo ar not needed.

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+	</$set>
+</$list>
+\end
+
+\define toolbar(tag,config,class,template)
+\whitespace trim
+<$set name="buttons" 
+		filter="[<__template__>!match[]]" 
+		value="""<$macrocall $name="buttons-with-template" tag=<<__tag__>> config= <<__config__>> class=<<__class__>> template=<<__template__>> />"""
```
pmario, line 34: `config= <<__config__>>`  should  have no space after the equal sign.

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+	<<buttons>>
+</$set>
+\end
+
+
+\define toolbar-filter-drop-down(tag,config)
+[all[shadows+tiddlers]tag[$tag$]!has[draft.of]] :filter[lookup:show[$config$]match[hide]]
+\end
+
+\define toolbar-basic-drop-down-list(tag,config,class)
```
pmario, line 45: whitespace trim is missing

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+<$list filter=<<toolbar-filter-drop-down "$tag$" "$config$">> variable="listItem">
+	<$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]] [<__class__>]" >
+		<$transclude tiddler=<<listItem>>/>
+	</$set>
+</$list>
+\end
+
+\define toolbar-drop-down-list-wt(tag,config,class,template)
```
pmario, line 53: whitespace trim

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+\end
+
+\define toolbar-drop-down-list(tag,config,class,template)
+\whitespace trim
+<$vars tv-config-toolbar-icons="yes" 
+		tv-config-toolbar-text="yes" 
+		tv-config-toolbar-class="tc-btn-invisible">
+    <$set name="drop-down" 
+			filter=" [<__template__>!match[]]" 
+			value="""<$macrocall $name="toolbar-drop-down-list-wt" tag=<<__tag__>> config= <<__config__>> class=<<__class__>> template=<<__template__>> />"""
```
pmario, line 68: This line of code seems to be a duplicate ... `config= <<__config__>>`  ... Is it possible to use a macro here?

core/wiki/macros/toolbar.tid
```diff
@@ -0,0 +1,73 @@
...
+	</$set>
+</$list>
+\end
+
+\define toolbar-drop-down-list-wt(tag,config,class,template)
+<$list filter=<<toolbar-filter-drop-down "$tag$" "$config$">> >
+    <$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]] [<__class__>]" >
+		<$transclude tiddler="$template$" />
+	</$set>
+</$list>
```
pmario, line 58: the lines above seem to mix tabs and spaces

core/wiki/macros/toolbar.tid
Alzacon: Yes, it is reused code (and typos...).

I think that it is possible but I also think that it may be complex and I don't know how many other times would be used. It would be a macro to differenciate other two macros by one(¿two, three, ...?) parameter(s) and to choose one macro if you call to one (¿two, three, ...?) differentiator parameter(s) It looks like a overkill for me.

core/wiki/macros/toolbar.tid
Alzacon: It is necessary? If this macro (and the following macro) is always used inside other macro with this pragma. I don't think that someone uses the macro `toolbar-basic-drop-down-list` directly.

I think I forgot to remove it on the lines 13 and 22.

Alzacon:
Fixes done except two `\whitespace trim` for now, see reason here-> https://github.com/Jermolene/TiddlyWiki5/pull/5444#discussion_r662152888

core/wiki/macros/toolbar.tid
pmario: OK.

editions/tw5.com/tiddlers/macros/ToolbarMacro.tid
```diff
@@ -0,0 +1,33 @@
...
+: Optionally, the config is the common prefix of config tiddlers where we can set state (show/hide) of the button. It uses to return which buttons will be displayed.
+
+;class
+: Optionally, additional [[CSS|Cascading Style Sheets]] classes for buttons. Multiple classes can be separated with spaces.
+
+;template
+: Optionally, the title of a tiddler to use as a [[template|TemplateTiddlers]] for transcluding the buttons.
+<br>
+
+[[Examples|Table-of-Contents Macros (Examples)]]
```
pmario, line 33: Links to the wrong examples ;)

editions/tw5.com/tiddlers/macros/examples/toolbar Macro (Examples).tid
```diff
@@ -0,0 +1,26 @@
...
+
+<$macrocall $name=".example" n="1" eg="""<div class="tc-page-controls">
+<<toolbar "$:/tags/PageControls" "$:/config/PageControlButtons/Visibility/">>
+</div>"""/>
+
+
+This example create a replica of ~TiddlyWiki ~PageControls with class "//tc-btn-rounded//", without change the general style for toolbar buttons:
+
+<$macrocall $name=".example" n="2" eg="""<div class="tc-page-controls">
+<<toolbar "$:/tags/PageControls" "$:/config/PageControlButtons/Visibility/" "tc-btn-rounded">>
```
pmario, line 16: IMO 1 of the examples should have "named parameters" eg: `tag:"$:/tags/PageControls"` and so on. 

_And_ the example for the `template` parameter is missing. ... IMO it would be interesting, to know, how it is meant to be used

pmario:
Except of my 2 remarks above it looks good to me!

editions/tw5.com/tiddlers/macros/ToolbarMacro.tid
Alzacon: :facepalm:

editions/tw5.com/tiddlers/macros/examples/toolbar Macro (Examples).tid
Alzacon: How you coud see i was following the doc of **toc** macro. The "named parameters" are not used there. And I think that there are multiple options to do a macro call.
- Unnamed and ordered parameters (which are used in examples)
- Named and ordered parameters (how you say it)
- Named and unordered parameters `<<toolbar config:"$:/config/PageControlButtons/Visibility/" tag:"$:/tags/PageControls">>`
- named and (un)ordered parameters mixed with unnamed and ordered parameters. `<<toolbar config:"$:/config/PageControlButtons/Visibility/" "$:/tags/PageControls">>`
But I think is better use the same syntax in the examples and maybe we can update the documentation of `Macro Call Syntax` and then we can add, in the tiddlers with examples, a link to (updated) documentation.

I agree. But I have several doubts/problems with this. I don't know how work with templates for buttons and the implementation of editor is WIP, the editor is in early step, its dropdown doesn't work. In short, I have doubts about if it is convenient create a ad hoc example and I don't know well how I can create it; and the alternative is in WIP and if it can't be confused to new users and few expert users can use it and they do it few times.

Alzacon: The following code works as example, but "more action" button in this toolbar doesn't work. I have to fix it.
```
<div  class="tc-editor-toolbar">
<<toolbar "$:/tags/EditorToolbar" "$:/config/EditorToolbarButtons/Visibility/" template:"$:/core/ui/EditTemplate/body/toolbar/button">>
</div>
```

Alzacon:
I found the problem with the editor dropdown. There is an extra filter (list) which works with a field `condition` of tiddler(/button). And now I have doubts about how to solve it. I doubt between two versions: A longer and more vesatile version or one with a minimal change that it can only be used in like editor toolbar. I don't know if the toolbar with button template will be used out of the editor toolbar.

Alzacon:
- Fixed problem with condition.
- Added the example for the template parameter by @pmario suggests
- Implementation done

There is a conflict, but i don't understand where is it.

Alzacon:
I forgot to comment:
- the pop story in toolbars was added to keep https://github.com/Jermolene/TiddlyWiki5/commit/3c86cf7d2e3e69fdda1f7a4353ff3a058f9d8b6b
- Add a warning in documentation about parameters class and template.
 
I don't understand why it detects a conflict in this file, I think that I do the same changes that in other files: remove config-title macro and add the respective toolbar macro. Also I add format with indents. 

I only "see" two possible causes for it: the "spaces" between (few) lines OR maybe i have edited the file without being in the branch then it causes the conflict

pmario:
There seems to be a problem in tiddler edit mode

see: 
grafik

pmario:
The more-dropdown has some CSS problems too. It looks completely different than at tiddlywiki.com

grafik

Alzacon:
Ouch! I forgot to comment the commits here. I tried to fix the conflict by reverting and making the implementation simpler and without changing the format, but it continues...

@pmario , it's true. They are problems known to me but I forgot fix them, because I focused on avoiding the conflict.
- In edit mode. I forgot remove a set variable widget unused in `$:/core/ui/EditTemplate`. This problem is also visible in the master build, when we open the preview of the edit mode in the tiddler `$:/core/ui/EditTemplate`. because with the set variable widget are added the following class `[[tc-btn-invisible tc-btn-%24%3A%2Fcore%2Fui%2FEditTemplate%2Fbody]]`
- I didn't add linebreaks, but the parser adds a container paragraph. However in the original tiddler with the same number of linebreak doesn't add this paragraph...

pmario:
> - I didn't add linebreaks, but the parser adds a container paragraph. However in the original tiddler with the same number of linebreak doesn't add this paragraph...

Yea, We know, that the core has some problems adding paragraphs, where they don't belong. ... There is a pending PR, but it is one which changes quite a lot of tiddlers. ... https://github.com/Jermolene/TiddlyWiki5/pull/4290

If you can point me to the offending code, I may have a solution. ...

Alzacon:
> If you can point me to the offending code, I may have a solution. ...

The offending code about?

I alredy had the solutions for the issues you pointed out.
- Remove this setwidget and its closing tag [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/EditTemplate.tid#L20#L22)
```
<$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]]">
</$set>
```
to
```
```

- Removing the linebreaks

If it is about the conflict, I don't understand why did it happens. I have only changed the blocks of code: I seled the original block and then paste the new (and if there was any indent I remove it to have the macro call in one line). Only the conflict ocurrs in the tiddler `core/ui/ViewTemplate/title.tid`

<hr>

OMG! That PR is a full revision of TW; Changes in js core, tests, styles and docs, and "_formatting in the code style_". Maybe it can be split in small and great PRs. The real PR about p generation is _around_ the wikiparser, I have to look how he solves it.

Alzacon:
@pmario , I think that I found the reason of the conflict. The file isn't updated. How can I solve it? (I am noob with git)

pmario:
>@pmario , I think that I found the reason of the conflict. The file isn't updated. How can I solve it? (I am noob with git)

I don't understand what you mean. "The file isn't updated". Can you be more specific?

Alzacon:
This file `core/ui/ViewTemplate/title.tid`. I think that i need undo changes, updated (¿fetch?) the repository and then I can do the changes.

pmario:
In your repo, there should be a button to fetch changes from jermolene/master ... If that's what you are after

grafik

Alzacon:
Sorry! I explain it too bad. I refer how I can do these steps (undo changes in the file, update it and do the changes) or others that solve the conflict. In other words, what are the commands that i have to use (and ther order)? 

If I use directly the button for fetch, it displays
in-fetch

pmario:
BACKUP - FIRST :)

It seems I do use a different workflow as you do. So I'm not sure. I never needed to do it that way. ..

But as is says: "Open a pull request." ... It seems you can open a pull request for your own branch. ... It should show you, what is different and hopefully where the conflicts are. You'll need to resolve those conflicts. ... 

After that you should be able to create a PR against jermolene/master again. ... 

As I wrote I don't know exactly, but I hope it will let you do something useful, without breaking stuff.

Alzacon:
There isn't way to avoid the conflict when I change the tiddler `title.tid`. The first of two commits don't have the conflict.

I don't know if I should close this PR and open a new one, or I revert changes in the file and do one PR for implementation in this file.

PR #5451, opened 2021-01-26
wikirules now use better macrocall parser

flibbles:
the macrocallinline and macrocallblock both implemented their own means of parsing macros, and both of them were inferior to the one used by the widget parser.

I've consolidated the macro parsing, so instead of there being **three** parsing logics, there's only one---the best one in parseutils.

Besides simplifying the code, this corrects the super annoying behavior where shorthand macro syntax couldn't handle wikitext arguments that also contained macrocalls. e.g.
```
<<mymacro arg:"""nested <<macro>> call""">>
```
There's no reason why that shouldn't work. Now it does!

Also, I added a bunch of tests. Also the code is smaller and less repetitive.

### DIFF
--- a/core/modules/parsers/parseutils.js
+++ b/core/modules/parsers/parseutils.js
@@ -132,7 +132,7 @@ exports.parseMacroParameter = function(source,pos) {
 		start: pos
 	};
 	// Define our regexp
-	var reMacroParameter = /(?:([A-Za-z0-9\-_]+)\s*:)?(?:\s*(?:"""([\s\S]*?)"""|"([^"]*)"|'([^']*)'|\[\[([^\]]*)\]\]|([^\s>"'=]+)))/g;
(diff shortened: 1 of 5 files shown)


flibbles:
Made one quick change to this. Found out while I was reconciling Relink with this that wikirule macrocalls can do **one** thing better than widget attribute macrocalls.
```
<<mymacro greater>than >> # This works

<$widget param=<<mymacro greater>than >> /> # This DOESN'T work
```
wikirule macrocall could allow single `>` sign in unquoted parameters, which the widget parser didn't. I fixed it so that the widget macrocall parameters also allow standalone `>`.

flibbles:
@Jermolene, the `=` sign was specifically forbidden in widget macro attributes, and I can't figure out why. Especially weird since they're allowed in normal macro calls.

```
<$text text=<<test myVar>=4>> /> <!-- This breaks. -->

<<test myVar>=4>>                <!-- This doesn't -->
```
I've push one more commit making `=` allowed in both cases. That way this PR doesn't break anyone's wikitext. Is there a reason I'm missing?

Jermolene:
Thanks @flibbles, this has been outstanding a long time and is a very worthwhile improvement.

> @Jermolene, the `=` sign was specifically forbidden in widget macro attributes, and I can't figure out why. Especially weird since they're allowed in normal macro calls.

It seems like that's a mistake, but I agree it's curious.

Issue #3306, opened 2018-05-21
<$list><$set><$set> nesting renders garbage closing tags


flibbles:
I'm noticing that this:
```
<$list filter="1 2 3 4 5">
<$set name="A" value="value">
<$set name="B" value="value">

</$set></$set></$list>
```
produces that:
```
</$list> </$list> </$list> </$list> </$list>
```
Am I doing something wrong?

In my project, I'm currently using a workaround of <$list><$list><$set> nesting.

Jermolene:
Hi @flibbles good spot. I think it's a bug, but we'll need to investigate further. In the prerelease you can use the preview type dropdown to choose the parse tree view which makes it easier to see.

Evolena:
It seems to be a inline/block issue. These codes don't add any list closing tag:

```
<$list filter="1 2 3 4 5">
<$set name="A" value="value">
<$set name="B" value="value">
</$set></$set></$list>
```

```
<$list filter="1 2 3 4 5">
<$set name="A" value="value">
<$set name="B" value="value">

</$set>
</$set></$list>
```

Or with something displayed inside:

```
<$list filter="1 2 3 4 5">
<$set name="A" value="value">
<$set name="B" value="value">
<<A>> (inline)
</$set></$set></$list>
```

```
<$list filter="1 2 3 4 5">

<$set name="A" value="value">
<$set name="B" value="value">
<<A>> (block)
</$set></$set></$list>
```

This one adds the extraneous `<\$list>`:

```
<$list filter="1 2 3 4 5">
<$set name="A" value="value">
<$set name="B" value="value">

<<A>>
</$set></$set></$list>
```

flibbles:
This is causing me more problems, so I looked at it a bit more.
```
<div><span><span>

Text
</span></span></div>
```
This also catches the problem. This is definitely an issue with the parser.

Putting in the following test in `test-wikitext-parser.js` (Line 96) replicates the problem:

```javascript
                expect(parse("<div><span><span>\n\nSome text</span></span></div>")).toEqual(

[...]

PR #5452, opened 2021-01-27
Fix for #3306, inline/block widget glitch

flibbles:
This bug kept biting me from time to time. After my work on macrocall rules, figured I'd fix this since I was in the neighborhood.

What was causing this was the html rule calling either parseBlocks or parseInlineRun. They were inconsistent about when they'd eat the run terminator.

* parseInlineRun never would, which is fine since the html rule eats it.
* parseBlock wouldn't if it downgraded to an inline rule, but it _would_ if it successfully parsed as a block. So the parseBlock and the html rule would both try to eat the run terminator. This was okay most of the time, since one of them would fail. Not in my use case where there happen to be two closing `</$set>` (or `</span>` in the test case)

So now parseInlineRun is passed `eatTerminator: true`, which means both parseInlineRun and parseBlock always eat it, so the html rule never has to.


Jermolene:
Thanks @flibbles good stuff

PR #5453, opened 2021-01-27
fix tiddler info area content bleeding on close animation

pmario:
This is a fix for the following problem: 

 - Set https://tiddlywiki.com/#%24%3A%2Fconfig%2FAnimationDuration  to eg: 2400
 - Open the tiddler **More : Info area**
 - You can see, that the "info text" becomes visible while it is still part of the main text area
 - The same "info text" "bleeding" also happens, if you close the info-area

This PR fixes the problem


Jermolene:
Thanks @pmario

PR #5458, opened 2021-02-01
Add throttling for changed tiddlers prefixed with $:/volatile/

NicolasPetton:
TiddlyWiki already supports rendering throttling when changed tiddlers have
`throttle.refresh`. Ensuring that the edited tiddler has that field is not very
convenient in many situations.

I believe that throttling is a very effective optimization that could be used in more situations if it was more convenient to use.
This PR adds support for throttling render updates when editing a tiddler that is prefixed with `$:/volatile/`.

The original idea of using `$:/volatile/` was from @Jermolene (I can't find the reference anymore though).

### DIFF
--- a/core/modules/startup/render.js
+++ b/core/modules/startup/render.js
@@ -82,7 +82,7 @@ exports.startup = function() {
 		var onlyThrottledTiddlersHaveChanged = true;
 		for(var title in changes) {
 			var tiddler = $tw.wiki.getTiddler(title);
-			if(!tiddler || !(tiddler.hasField("draft.of") || tiddler.hasField("throttle.refresh"))) {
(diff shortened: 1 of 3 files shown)


NicolasPetton:
Update: I've changed the prefix to `$:/temp/volatile/` so that existing filters filtering out temp tiddlers would work without any modification.

@Jermolene what do you think?

NicolasPetton:
@Jermolene friendly ping :-)

NicolasPetton:
@Jermolene This PR has been opened since February, any feedback on this?

Jermolene:
Perfect, thank you @NicolasPetton, apologies for missing these things. You have been very patient, but please do feel free to reach out directly to me if something falls through the cracks like this again.

Issue #5236, opened 2020-12-07
[IDEA] Tools for manipulating colors


saqimtiaz:
It would be helpful to provide some tools to complement the palette system, which would allow things like adjusting the brightness or transparency of existing colours in the palette. This would help limit the need for adding new palette entries every time a new colour is needed.

See https://github.com/Jermolene/TiddlyWiki5/issues/5220#issuecomment-739822421 and https://github.com/Jermolene/TiddlyWiki5/issues/5220#issuecomment-739830534

kookma:
@saqimtiaz - Thank you for this.
The current palette is very lengthy and many elements has been given there to be manually set. Maintaining these palettes is a little hard and I think  they should me more flexible.

One suggestion is: to have a set of main colors like 4 to 6 colors and all elements derive their colors from them programmatically like using alpha or contrast or some mixing macros.

As an example see https://www.canva.com/learn/100-color-combinations/ which most palettes are built based on four main colors.

I tried to use the current palette e.g. Vanilla to generate required color and shading for some my plugins but it is very difficult and semantically confusing. For example to set a background color of a div in particular tiddler I have to use border color of some element in Vanilla palette. That is why I said we need to have some main colors in each palette with semantic names.

I should emphasis that, the current palette seems to allow user set colors like red instead of #ff0000 so, I think for core palettes it should be a rule to have all colors in 6 digit hex and force main colors to be defined only in 6 hex digits.

MrDowntempo:
Maybe something like one of these:
https://ourcodeworld.com/articles/read/81/top-5-best-javascript-and-jquery-color-picker-plugins

pmario:
IMO this should be a "discussion" and no issue. @Jermolene could move it.

Jermolene:
> IMO this should be a "discussion" and no issue. @Jermolene could move it.

You're probably right. Before starting to do that though I wonder if we might try to figure out what our guidelines are for the what should be an issue and what should be a discussion.

At one extreme, perhaps issues should just be bug reports and specific, actionable proposals. At the other, perhaps issues are focussed on a single goal, while discussions can be more freewheeling.

saqimtiaz:
> At one extreme, perhaps issues should just be bug reports and specific, actionable proposals.

[...]

PR #5459, opened 2021-02-01
All colour codes use 6 digits now

pmario:
While working on the $:/PaletteManager I did stumble upon this problem again. 

Does not work

grafik

Would work ... I know that this PR is only a partial fix, but it is a baby step in the right direction and the PR already changes 16 files. So the chance to get it merged with 16 more is close to 0. 

grafik

<details><summary> Initial text for this PR </summary>
This PR is important for consistency reasons, to make it possible to create CSS settings as follows. 

To test the PR copy the following content to tiddlywiki.com and to your localhost, to see the difference!

```
<style>
.test-3 {
  background: <<colour "tag-background">>4;  /* this is the inconsistency that should be fixed */
}

.test-6 {
  background: <<colour "tag-background">>40;  /* it should always be possible like this */
}

.test-x {
  background: #eecc6640;
}

.test-full {
  background: #eecc66;
}
</style>

<div class="test-3">test text - works but is inconsistent</div>

<div class="test-6">test text - background color ''fails'' atm but should work</div>

<div class="test-x">test text - transparant background</div>

<div class="test-full">test text - full background</div>
```
</details>

### DIFF
--- a/core/palettes/Blanca.tid
+++ b/core/palettes/Blanca.tid
@@ -23,7 +23,7 @@ dragger-background: <<colour foreground>>
 dragger-foreground: <<colour background>>
 dropdown-background: <<colour background>>
 dropdown-border: <<colour muted-foreground>>
-dropdown-tab-background-selected: #fff
+dropdown-tab-background-selected: #ffffff
 dropdown-tab-background: #ececec
 dropzone-background: rgba(0,200,0,0.7)
 external-link-background-hover: inherit
@@ -53,7 +53,7 @@ select-tag-background:
 select-tag-foreground:
 sidebar-button-foreground: <<colour foreground>>
 sidebar-controls-foreground-hover: #000000
-sidebar-controls-foreground: #ccc
+sidebar-controls-foreground: #cccccc
 sidebar-foreground-shadow: rgba(255,255,255, 0.8)
 sidebar-foreground: #acacac
 sidebar-muted-foreground-hover: #444444
@@ -80,7 +80,7 @@ table-border: #dddddd
 table-footer-background: #a8a8a8
 table-header-background: #f0f0f0
 tag-background: #ffeedd
-tag-foreground: #000
+tag-foreground: #000000
 tiddler-background: <<colour background>>
 tiddler-border: #eee
 tiddler-controls-foreground-hover: #888888
--- a/core/palettes/Blue.tid
+++ b/core/palettes/Blue.tid
@@ -8,7 +8,7 @@ alert-background: #ffe476
 alert-border: #b99e2f
 alert-highlight: #881122
 alert-muted-foreground: #b99e2f
-background: #fff
+background: #ffffff
 blockquote-bar: <<colour muted-foreground>>
 button-background:
 button-foreground:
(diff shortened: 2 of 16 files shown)


pmario:
I'll fix the docs in the TW-docs branch if this will be merged. see: https://tiddlywiki.com/#ColourPalettes

OR

@Jermolene I can add the docs here if you want

kookma:
@pmario - thank you for this. It is good to add such requirement in Tiddlywiki documentation. I mean any new palettes should follow the six digits hex color code.

By the way I think if the palette itself needs a eight digits color code, then a call like `<<colour "tag-background">>40;` can be used.

One other recommendation is to define some main colors in a palette and derive all other colors from them. Right now there are too many colors in a palette.

pmario:
>... One other recommendation is to define some main colors in a palette and derive all other colors from them. Right now there are too many colors in a palette.

I think you are right. ... TW classic has been much simpler in this regard. ... (good ol' times ;)

pmario:
I would like to have something like this: 

```
<<colour "tag-background">><<lighter>>;
<<colour "tag-background">><<darker>>;

<<colour "tag-color">>;
<<colour-hover "tag-color">>;
```

`<<colour-hover>>` would use something like [contrastcolour macro](https://tiddlywiki.com/#%24%3A%2Fcore%2Fmodules%2Fmacros%2Fcontrastcolour.js) if needed. .. So if the "foreground/background" contrast is too low, the "colour" should be "inverse" or something like this

But this would be a completely different PR ;)

saqimtiaz:
This and related issues like using alpha channel values in colours was the motivation behind #5236 

Recommend reading the comments linked in the first post of that issue.

pmario:
@saqimtiaz ... Yea, For 1 of my plugins I wanted to use a predefined "palette colour". I couldn't find one, but I found one that was pretty close and also works for the purpose. So no weird naming confusions, but I wanted it to be "lighther". ... 

I did create a $:/StyleSheet tiddler and used `background: <<xxx>>50` ... but it didn't work. It needed to be `<<xxx>>5` ... BUT the problem now is, that different palettes may have different number of digits. 

That's the reason for this PR

Jermolene:
Hi @pmario it's a nice idea but sadly I don't think we can restrict the result of the `<<colour>>` macro to be a simple CSS hex colour. For example, one of my projects uses [wide colours](https://css-tricks.com/wide-gamut-color-in-css-with-display-p3/) and thus has colour definitions like:

```
background-color: color(display-p3 1 1 0);
```

Nor, of course, can we enforce that user generated palettes use any particular format.

pmario:
@Jermolene I didn't try to restrict the result of the colour-macro. This PR tries to make the existing palettes, that are shipped with the core consistent.

So that the macro can be used in a consistent way.  ... That's all. 

If we need to adjust the macros for other color formats IMO it would be a separated PR

pmario:
But you are right, I should probably change all "core settings" to CSS hex colours for consistency reasons. ... 

Our users most of the time use the mechanism, the core suggests. ... 

If not, it won't be compatible as long as someone makes it compatible. ... BUT at least the core is consistent!

pmario:
I did update [the OP](https://github.com/Jermolene/TiddlyWiki5/pull/5459#issue-798462426)

--------

@Jermolene  wrote

>Hi @pmario it's a nice idea but sadly I don't think we can restrict the result of the <<colour>> macro to be a simple CSS hex colour. For example, one of my projects uses [wide colours](https://css-tricks.com/wide-gamut-color-in-css-with-display-p3/) and thus has colour definitions like:
>
>`background-color: color(display-p3 1 1 0);`
>
>Nor, of course, can we enforce that user generated palettes use any particular format.

I don't want to restrict anybody from using any CSS colour settings they want to use. 

This PR should be 1 little step towards a broader fix. If this PR would be out of the way we could build on it and implement more possibilities.

pmario:
The new palette-manager code can handle this edge case. So close

PR #5464, opened 2021-02-02
Add th-before-importing hook mechanism

pmario:
This PR will allow plugin authors to do the following: 

```
	$tw.hooks.addHook("th-before-importing",function(importTiddler) {

// Do your thing here.

		return importTiddler;
	});
};
```

The `importTiddler` object contains information about the "checked / unchecked" state and other infos about the tiddlers that will be imported. 

This hook and the info object allows us to create a logging mechanism, that uses this info to write custom "import reports". I want to use it to create a "log-tiddler" that contains a **header text** in the log tiddler.

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -560,7 +560,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 		importData,
 		importReport = [];
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
-	// If you need to modify the import tiddler payload, consider th-importing-tiddler instead!
+	// If you need to modify the import tiddler payload then consider th-importing-tiddler instead!
 	importTiddler = $tw.hooks.invokeHook("th-before-importing",importTiddler);
 	importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
 	$tw.utils.each(importData.tiddlers,function(tiddlerFields) {
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -557,13 +557,14 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	var self = this,
 		importTiddler = this.wiki.getTiddler(event.param),
-		importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
+		importData,
 		importReport = [];
 
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
-	// th-before-importing _can not_ manipulate the importTiddler.fields.text element!
-	// Use th-importing-tiddler later in the process
+	// If you need to modify the import tiddler payload, consider th-importing-tiddler instead!
 	importTiddler = $tw.hooks.invokeHook("th-before-importing",importTiddler);
+	// Update importData, if a hook did manipulate $:/Import 
+	importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
 	$tw.utils.each(importData.tiddlers,function(tiddlerFields) {
 		var title = tiddlerFields.title;
 		if(title && importTiddler && importTiddler.fields["selection-" + title] !== "unchecked") {
@@ -573,7 +574,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 				var tiddler = new $tw.Tiddler(tiddlerFields);
 			}
 			// th-importing-tiddler doesn't allow user interaction by default
-			// If you want to use the core UI. Use: $:/core/modules/upgraders/ instead
+			// If you want to use the default UI, use: $:/core/modules/upgraders/ instead
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
 			// Add the tiddlers to the store
 			self.wiki.addTiddler(tiddler);
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -560,9 +560,10 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 		importTiddler = this.wiki.getTiddler(event.param),
 		importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
 		importReport = [];
-	// Add the tiddlers to the store
+
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
-	// Let the loggers do their work
+	// th-before-importing _can not_ manipulate the importTiddler.fields.text element!
+	// Use th-importing-tiddler later in the process
 	importTiddler = $tw.hooks.invokeHook("th-before-importing",importTiddler);
 	$tw.utils.each(importData.tiddlers,function(tiddlerFields) {
 		var title = tiddlerFields.title;
@@ -572,7 +573,10 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 			} else {
 				var tiddler = new $tw.Tiddler(tiddlerFields);
 			}
+			// th-importing-tiddler doesn't allow user interaction by default
+			// If you want to use the core UI. Use: $:/core/modules/upgraders/ instead
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
+			// Add the tiddlers to the store
 			self.wiki.addTiddler(tiddler);
 			importReport.push("# [[" + tiddler.fields.title + "]]");
 		}
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -572,7 +572,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 				var tiddler = new $tw.Tiddler(tiddlerFields);
 			}
 			// th-importing-tiddler doesn't allow user interaction by default
-			// If you want to use the default UI, use: $:/core/modules/upgraders/ instead
+			// If you want to use the default UI then use: $:/core/modules/upgraders/ instead
 			tiddler = $tw.hooks.invokeHook("th-importing-tiddler",tiddler);
 			// Add the tiddlers to the store
 			self.wiki.addTiddler(tiddler);
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -561,7 +561,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 		importReport = [];
 	// Add the tiddlers to the store
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
-	// Let the loggers do their work. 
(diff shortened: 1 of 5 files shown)


pmario:
Documentation will be added to the tiddlywiki/dev edition

Jermolene:
Thanks @pmario it's best to have the docs in the same PR unless that's inconvenient.

pmario:
@Jermolene ... Are you OK with the hook-name? ... I'll add the docs ASAP to this PR

pmario:
@Jermolene ... docs added **and** did also set the "theme tweaks" "sidebar layout" to "fluid story, fixed sidebar", so it looks more like tiddlywiki.com

editions/dev/tiddlers/new/Hook_ th-before-importing.tid
```diff
@@ -0,0 +1,62 @@
...
+
+* ''importTiddler'': an object, that contains information about "selected / unselected" tiddlers and more. 
+
+Return value:
+
+* ''importTiddler'': object
+
+The original importTiddler object can be returned unmodified by the hook, which is probably the case for most use-cases.
+
+If you want to manipulate the tiddlers, that should be imported use: [[Hook: th-importing-tiddler]] instead.
```
saqimtiaz, line 23: The comma in this sentence is unnecessary and makes it difficult to understand.

editions/dev/tiddlers/new/Hook_ th-before-importing.tid
```diff
@@ -0,0 +1,62 @@
+created: 20210203184649726
+modified: 20210203191622629
+tags: HookMechanism
+title: Hook: th-before-importing
+type: text/vnd.tiddlywiki
+
+This hook allows plugins to inspect or modify the `importTiddler` object ''before'' any tiddlers are imported. It is invoked when the final "Import" button is clicked, but ''before'' the selected tiddlers are being imported into the store.
```
saqimtiaz, line 7: Have you checked that modifying the importTiddler object is actually reflected in the tiddlers imported? I suspect this is not the case, as the importData variable has already been assigned.

core/modules/widgets/navigator.js
```diff
@@ -561,7 +561,7 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 		importReport = [];
 	// Add the tiddlers to the store
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
-	// Let the loggers do their work. 
+	// Let the loggers do their work
```
saqimtiaz, line 565: This comment is far too vague as we have no predefined concept of "loggers" in the TiddlyWiki code.

pmario: What do you suggest?

editions/dev/tiddlers/new/Hook_ th-before-importing.tid
pmario: >Have you checked that modifying the importTiddler object is actually reflected in the tiddlers imported?

Yes! The bundler also contains some upgraders, that are executed earlier in `NavigatorWidget.prototype.handleImportTiddlersEvent`, which makes it possible to manipulate the importTiddler data in this hook. 

At the moment I'm only using the hook, to write a "header" line into a log-tiddler. This doesn't need to manipulate the importTiddler data.

editions/dev/tiddlers/new/Hook_ th-before-importing.tid
pmario: will fix it

core/modules/widgets/navigator.js
```diff
@@ -561,6 +562,8 @@ NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 		importReport = [];
 	// Add the tiddlers to the store
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
+	// Let the loggers do their work
+	importTiddler = $tw.hooks.invokeHook("th-before-importing",importTiddler);
```
saqimtiaz, line 566: I recommend moving the variable assignment for importData from above to here.

`importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}})`

This would make the hook a lot more versatile as if needed, the raw JSON payload of the importTiddler could be manipulated.

pmario: >This would make the hook a lot more versatile as if needed, the raw JSON payload of the importTiddler could be manipulated.

IMO at this point it's way too late to manipulate tiddlers. Import tiddlers can be manipulated using the "upgraders" mechanism. You can manipulate on a "per tiddler" basis. It's simple

saqimtiaz: @pmario then your documentation and code are not coherent. If the code is to stay as is, your documentation update needs to reflect that the only the fields of the `importTiddler` other than the text field with the `JSON` data can be manipulated in the hook that you are adding. 

Furthermore, I don't really see a reason to artificially limit the utility of a hook in this manner. However, regardless of that please ensure that the docs update accurately reflects what the hook can and cannot be used for.

core/modules/widgets/navigator.js
```diff
@@ -527,6 +527,7 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	});
 	// Deselect any suppressed tiddlers
 	$tw.utils.each(importData.tiddlers,function(tiddler,title) {
+		// count the keys in the tiddler object
```
saqimtiaz, line 530: Unless we intend to add comments everywhere that core util functions are used, why do we need this comment for this specific instance of `$tw.utils.count` being called?

editions/dev/tiddlers/new/Hook_ th-before-importing.tid
saqimtiaz: This is incorrect and the docs are misleading. The documentation suggests that the `importTiddler` can be freely modified and the changes will be reflected in the imported tiddlers. However, any changes to the `JSON` payload are not reflected in the imported tiddlers.

core/modules/widgets/navigator.js
pmario: Because of this: #5466 "[IDEA] Pls let us rename `$tw.utils.count(tiddler)` to eg: `$tw.utils.cntObjectKeys(xxxx)` "

saqimtiaz: @pmario that is no reason to add a spurious comment for one particular instance of a $tw.util method being called. Either we do it consistently everywhere, or we don't. There is no reason this particular usage of that method is more confusing and in need of clarification than others.

core/modules/widgets/navigator.js
Jermolene: Actually there is already a $tw.Logger classs in the core (see syncer.js for example)

Jermolene: It seems like a more appropriate comment would be something like "Give hooks an opportunity to modify the tiddlers before we process them"

Jermolene (commented):
Thanks @pmario @saqimtiaz just a couple of minor comments while you resolve the other issues

pmario:
@saqimtiaz ... I did move the `importData` definition to the point that you suggested, which should make the code more robust. 

I did update the docs to describe the "intended" use-case. 

If a developer needs to manipulate the "payload", the docs and the code comments, show them 2 alternative ways. I personally prefer the alternatives, since they are much simpler to handle and less error prone.

core/modules/widgets/navigator.js
pmario: done! See lines below

core/modules/widgets/navigator.js
```diff
@@ -557,10 +557,14 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
...
 		importTiddler = this.wiki.getTiddler(event.param),
-		importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
+		importData,
 		importReport = [];
-	// Add the tiddlers to the store
+
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
+	// If you need to modify the import tiddler payload, consider th-importing-tiddler instead!
+	importTiddler = $tw.hooks.invokeHook("th-before-importing",importTiddler);
+	// Update importData, if a hook did manipulate $:/Import
```
saqimtiaz, line 566: @pmario we are assigning importData and not updating it, so the comment is confusing. I think we can skip this comment entirely as the flow seems rather intuitive.

saqimtiaz:
@pmario thank you. One minor comment that needs updating, otherwise this looks good.

pmario:
@Jermolene, @saqimtiaz ... I did change everything we discussed. .... BUT there is still the [fluid-fixed](https://github.com/Jermolene/TiddlyWiki5/pull/5464/files#diff-2d29afe28fd84fe18f1759e0722a78f94e07051b14dece2cb35420cf46e11f5a) Layout setting. It's the same as tiddlywiki.com now. I'd like to keep it

saqimtiaz:
@pmario do you want to change the default fluid-fixed setting? Apologies, I did not notice that change.

 If so, that really needs to be a separate PR. Firstly that makes for a more organized commit history where it is easier to see which changes originate in which PR. Also, this would allow proper discussion around it as others might have opinions on this as well. 
 
 Personally that setting is one of the first things I change in an empty TW but others may not feel the same. Also, there is a keyboard shortcut now to make it easier to switch.

Jermolene:
I'm happy with changing the sidebar layout on tiddlywiki.com/dev, as @pmario says it is consistent with tiddlywiki.com.

In typical TiddlyWiki fashion, when we introduced fluid-fixed layout we only kept fixed-fluid layout as the default for backwards compatibility reasons, despite it's obvious superiority (!). It's probably about time we talked about changing the default.

saqimtiaz:
I missed that the change was just for the **dev** edition, in which case I think it is fine to merge as is. 

I would even be in favour of starting a discussion about changing it for the empty edition.

pmario:
@pmario ... note to myself

core/modules/widgets/navigator.js
```diff
@@ -557,10 +557,12 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 NavigatorWidget.prototype.handlePerformImportEvent = function(event) {
 	var self = this,
 		importTiddler = this.wiki.getTiddler(event.param),
-		importData = this.wiki.getTiddlerDataCached(event.param,{tiddlers: {}}),
+		importData,
 		importReport = [];
-	// Add the tiddlers to the store
 	importReport.push($tw.language.getString("Import/Imported/Hint") + "\n");
+	// If you need to modify the import tiddler payload, consider th-importing-tiddler instead!
```
Jermolene, line 563: I'm sorry to keep raising the same issue, but the exclamation mark is not needed here. In English it would also be clearer to replace the comma with the word "then"

Jermolene (commented):
Thanks @pmario

core/modules/widgets/navigator.js
pmario: Will be updated

pmario:
I did update the comment and I did revert the "fluid-fixed" setting to the "old" default, since support for changing empty.html hasn't been a huge success. see: #5492

Jermolene:
Thanks @pmario

PR #5416, opened 2021-01-17
targeted fix for 5366

joshuafontany:
Fixes #5366

PR #5465, opened 2021-02-02
Fix filesystem

joshuafontany:
This pulls all the `filesystem.js` and related updates out of my experimental draft-PR. This should be merged alongside https://github.com/Jermolene/TiddlyWiki5/pull/5416

They key changes are:
* Ensure all callbacks recieve a second argument "fileInfo/adaptorInfo" & do not rely on global objects during async tasks
  * Makes sure that all code-paths return a valid second data-object if the method does not error.
  * Wrap an anon function around the callback when calling certain $tw.utils functions.
* Patch a vulnerability with the final filePath when writing to disk.
  * Hooks should be called first, then it should be tested for the "encode" logic to prevent writing outside the wiki directory.
* In `WikiFolderMaker.prototype.saveTiddler` use the path/ext filters and a try/catch block to test for `saveTiddlerToFileSync` errors and log to the console.
  * This prevents file-save errors from crashing the command.


core/modules/commands/savewikifolder.js
```diff
@@ -158,11 +158,26 @@ WikiFolderMaker.prototype.saveCustomPlugin = function(pluginTiddler) {
 };
 
 WikiFolderMaker.prototype.saveTiddler = function(directory,tiddler) {
+	var title = tiddler.fields.title, fileInfo, pathFilters, extFilters;
+	if(this.wiki.tiddlerExists("$:/config/FileSystemPaths")){
```
Jermolene, line 162: There should be a space between the closing parenthesis and the opening brace

core/modules/commands/savewikifolder.js
```diff
@@ -158,11 +158,26 @@ WikiFolderMaker.prototype.saveCustomPlugin = function(pluginTiddler) {
...
+		pathFilters: pathFilters,
+		extFilters: extFilters,
+		originalpath: this.wiki.extractTiddlerDataItem("$:/config/OriginalTiddlerPaths",title, "")
 	});
-	$tw.utils.saveTiddlerToFileSync(tiddler,fileInfo);
+	try {
+		$tw.utils.saveTiddlerToFileSync(tiddler,fileInfo);
+	} catch (err) {
+		console.log("SaveWikiFolder: Error saving file '" + fileInfo.filepath + "', tiddler: '" + tiddler.fields.title);
+		console.error(err);
```
Jermolene, line 179: We should, but we don't currently consistently send error messages to console.error. Perhaps that's something we should do in bulk as a separate PR?

core/modules/utils/filesystem.js
```diff
@@ -419,14 +417,29 @@ exports.saveTiddlerToFile = function(tiddler,fileInfo,callback) {
...
+					return callback(err);
+				}
+				return callback(null,fileInfo);
+			});
 		});
 	} else {
 		// Save the tiddler as a self contained templated file
 		if(fileInfo.type === "application/x-tiddler") {
-			fs.writeFile(fileInfo.filepath,tiddler.getFieldStringBlock({exclude: ["text","bag"]}) + (!!tiddler.fields.text ? "\n\n" + tiddler.fields.text : ""),"utf8",callback);
+			fs.writeFile(fileInfo.filepath,tiddler.getFieldStringBlock({exclude: ["text","bag"]}) + (!!tiddler.fields.text ? "\n\n" + tiddler.fields.text : ""),"utf8",function(err){
```
Jermolene, line 430: Again, there should be another space: `function(err) {`

Jermolene (changes requested):
Thanks @joshuafontany a couple of minor points

core/modules/commands/savewikifolder.js
joshuafontany: Agree. I think it would be more consistent to use $tw.logger.alert() here (& in similar cases), and that logger method should print errors to both node and browser console (along with the alert-tiddler logic). That should be a separate PR.

joshuafontany:
@Jermolene Thanks for catching those. :)

Jermolene:
Thanks @joshuafontany

PR #5467, opened 2021-02-02
Update sync methods

joshuafontany:
Targeted PR updates all the syncadaptor "save" "load" and "delete" methods to follow node.js callback-as-last-argument standard.

This will be useful for anyone trying to "promisify" code, for example for a node.js server that uses Promises.


Jermolene:
Thanks @joshuafontany I think you had a docs update in the previous PR too?

joshuafontany:
@Jermolene Thanks for the reminder, docs pushed! :)

joshuafontany:
@Jermolene Conflicts resolved.

joshuafontany:
@Jermolene @saqimtiaz Would this be a good candidate for the 5.2.0 version? I am still experimenting with Bob & also have some changes to the Listen command  server that would allow other syncers to be used with the main Node server - but thats better as a separate pr.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -75,7 +75,16 @@ FileSystemAdaptor.prototype.getTiddlerFileInfo = function(tiddler,callback) {
 /*
 Save a tiddler and invoke the callback with (err,adaptorInfo,revision)
 */
-FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
+FileSystemAdaptor.prototype.saveTiddler = function(tiddler,options,callback) {
+	// Starting with 5.1.24, all syncadptor method signatures follow the node.js
```
Jermolene, line 79: I'll do a search and replace for 5.1.24 -> 5.2.0 before release, but it might be worth updating it here for clarity. Also, it is preferred to prefix version numbers with a "v".

plugins/tiddlywiki/savetrail/savetrail.js
```diff
@@ -99,7 +99,16 @@ SaveTrailSyncAdaptor.prototype.getTiddlerInfo = function(tiddler) {
 /*
 Save a tiddler and invoke the callback with (err,adaptorInfo,revision)
 */
-SaveTrailSyncAdaptor.prototype.saveTiddler = function(tiddler,callback) {
+SaveTrailSyncAdaptor.prototype.saveTiddler = function(tiddler,options,callback) {
+	// Starting with 5.1.24, all syncadptor method signatures follow the node.js
```
Jermolene, line 103: Typo "syncadptor"

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -235,7 +253,16 @@ Delete a tiddler and invoke the callback with (err)
 options include:
 tiddlerInfo: the syncer's tiddlerInfo for this tiddler
 */
-TiddlyWebAdaptor.prototype.deleteTiddler = function(title,callback,options) {
+TiddlyWebAdaptor.prototype.deleteTiddler = function(title,options,callback) {
+	// Starting with 5.1.24, all syncadptor method signatures follow the node.js
+	// standard of callback as last argument. This catches the previous signature:
```
Jermolene, line 258: It might be worth trying to abbreviate the repeated code here; we probably don't need the comments.

Jermolene (commented):
Thanks @joshuafontany a few comments

Jermolene:
Thanks @joshuafontany

Jermolene:
Hi @joshuafontany somehow I didn't realise that this PR is not backwards compatible, which I appreciate means that I hadn't thought about it properly (the problem is that I saw the code that updated the syncadaptors to work with an old syncer, but of course there is no provision for the updated syncer to work with the old syncadaptor interface). I know that there are many 3rd party syncadaptors out there (it's worth [searching](https://github.com/search?q=%22syncadaptor%22&type=code))

It is categorically not acceptable for us to unilaterally change the syncadaptor interface and break all existing syncadaptors, my apologies for not figuring out what was going on earlier.

So, I'll need to revert this PR, but I will need some help understanding whether anything else will need changing/reverting in light of that. Is there anything else that you can think of?

joshuafontany:
Hi @Jermolene 

Oooooh... I don't think this will be a problem, we'll revert this PR. I have also been working with the syncer code a lot recently, and can see how I missed that backwards compatability issue when I first proposed this.

My experiments with a "live multiplayer" feature with a Bob branch recently had me reconsidering if this was neccessary. I see now where some of that feeling was coming from. I'll think in any possible halo-bugs - but feel free to undo this one.

Jermolene:
Thanks @joshuafontany, really no problem, the process has worked by the skin of its teeth!

PR #5469, opened 2021-02-03
add html5 elements and remove those, that are wrong

pmario:
This PR adds some missing "block level" HTML elements and removes some that are not block level.


pmario:
@Jermolene ... Is there anything, that blocks this PR?

Jermolene:
Thanks @pmario

Issue #5435, opened 2021-01-22
[BUG] Global use of pragma "\whitespace trim" remove specific space inside of text 


Alzacon:
**Describe the bug**
If I use the pragma `\whitespace trim` globally, it removes space inside the text when there is a change of style. 
>Control whether whitespace is trimmed from the start and end of text runs

Does "text run" just end with the line break?
I don´t know if it is a true bug, but it doesn´t work as expected for me.

**To Reproduce**
```
\whitespace trim

this ''text''
```
The result is:
> this**text**

**Expected behavior**
>this **text**

BlueGreenMagick:
```
\whitespace trim
this '' text''
```

The above example would be parsed as
```
[
  [type: "text" text: "this "],
  [type: "element" tag:"stong" children: [[text " text"]]
]
```

Because each text element would be trimmed, `"this "` is trimmed to `"this"`, `'' text''` to `text`.

pmario:
The first trim works with the body text
The second trim works inside the macro definition
```
\whitespace trim 

\define macro()
\whitespace trim 
xxxx  yyyy
\end

aaa

bbb
```

BUT

whitespace trim was developed to allow us to use code nesting in complicated macro definitions. The downside is, that it also removes whitespace in text. So it can only be used effectively, if the `<<lingo xx>>` macro is used for text

Alzacon:
Thank you very much to both, @pmario and @BlueGreenMagick 

Then it isn't a bug. I close the issue.

PR #5473, opened 2021-02-04
Fix ViewToolbar items inconsistent spacing

BlueGreenMagick:
image

On the attached screenshot, you can see that 'delete' and 'close' is not aligned with other items. This PR fixes the issue.


core/ui/EditToolbar/delete.tid
```diff
@@ -9,6 +9,9 @@ description: {{$:/language/Buttons/Delete/Hint}}
 {{$:/core/images/delete-button}}
 </$list>
 <$list filter="[<tv-config-toolbar-text>match[yes]]">
-<span class="tc-btn-text"><$text text={{$:/language/Buttons/Delete/Caption}}/></span>
+<span class="tc-btn-text">
+<$text text=" "/>
```
pmario, line 13: We don't want to use whitespace for spacing. There are some dedicated CSS settings

Alzacon:
I agree with @pmario 

I think something like this CSS could work:

``` CSS
svg + .tc-btn-text{
	margin-left: 1em;
}
```
We would need to adjust to the best margin value and then these hardcoded spaces, `<$text text=" "/>`, can be removed.

BlueGreenMagick:
I've tried @Alzacon's code and removed all hard coded whitespaces from other ViewToolbar items, but it fails to align the ViewToolbar item texts.

It looks like the reason there is a whitespace hard coded into many of ViewToolbar items is because whitespaces are inserted when the ViewToolbar items are defined and used as a macro.(For example, [core/ui/ViewToolbar/new-journal-here.tid](https://github.com/Jermolene/core/ui/ViewToolbar/new-journal-here.tid)) So `<$text text=" "/>` is inserted to align other non-macro ViewToolbar items with them.

I'm not sure if it is an intended behaviour of macros, or maybe it is a bug of whitespace trim pragma. But it will be much simpler to add whitespaces to the only two ViewToolbar items that isn't aligned.

pmario:
I think it's not a problem with 5.1.23 "edit toolbar" but in the dropdown ... We should have a closer look and fix the problem and not the symptoms

Alzacon:
@BlueGreenMagick , you're right

The `\whitespace` pragma don't run inside the macro if the pragma is outside of macro, it only works when you put one pragma whitespace inside.

I don't know how this pragma works and what its expected behaviour are.

In fact I have opened this issue (https://github.com/Jermolene/TiddlyWiki5/issues/5435) about behaviour of this pragma.

pmario:
I think we need to fix `tc-btn-text` and remove all the unnecessary `<$text text=" ">` widgets. They only increase siez and need rendering time. IMO CSS would be much more effective here. 

BUT 

we need to have a look, where `tc-btn-text` is used and check if there is no unwanted side effects

pmario:
As far as I can see `tc-btn-text` is used 54 times in 50 different tiddlers. It's not hardcoded in the JS code.

BlueGreenMagick:
This was a bigger change than I first thought it would be, but this should be good now.

Result:
imageimage

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -555,6 +555,10 @@ html body.tc-body .tc-btn-rounded:hover svg {
 	fill: <<colour muted-foreground>>;
 }
 
+svg + .tc-btn-text, .tc-image-button + .tc-btn-text{
+	margin-left: 0.3em;
```
pmario, line 559: 0.3em seems / feels a little bit close. Did you try 0.4 or 0.5. ... How does that feel?

pmario:
A global search in the project gave me 50 files. ... So there may be something missing

themes/tiddlywiki/vanilla/base.tid
pmario: does the CSS need to be so complicated?

Isn't it enough to use the rule below. ... IMO it should be

Alzacon: If we have it as you suggest then it will run always, with or without image. The sibling selector avoids it. https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator

And the rule `.tc-btn-text` looks unused.

If finally the value is `0.5em`, the selector maybe could be add in this rule 
https://github.com/BurningTreeC/TiddlyWiki5/blob/62eee8118368b7f728b00537d056cd885194ace5/themes/tiddlywiki/vanilla/base.tid#L2774

BlueGreenMagick: I think you are right that just using `.tc-btn-text` would be good. Currently, `.tc-btn-text` is always preceded by an icon, so I'll assume that it is a class that is used for button texts that comes after icons. It can simply be changed if someone uses the class for other purposes.

Also, setting margin-left to an `em` value did not look good on large font-sizes. (Which can be seen for ViewToolbar when you set `$:/config/Toolbar/Text` to true.) So switched to pixel value `7px` which is approximately  `0.4em` for the dropdowns.

BlueGreenMagick:
Many of the files do not require any change, as they are without spacing. In fact, if you look at the modified files of this PR, some are just modifying code style to make it consistent with other files.

themes/tiddlywiki/vanilla/base.tid
pmario: >I think you are right that just using .tc-btn-text would be good. Currently, .tc-btn-text is always preceded by an icon, so I'll assume that it is a class that is used for button texts that comes after icons. It can simply be changed if someone uses the class for other purposes.

You are right, it is exclusively used that way atm. So imo we can use it to add the margin-left, if possible. I would like to avoid CSS selectors that are too specific. If we can use `.tc-btn-text` we should use it.

pmario: We need to make sure, that the following setting looks good: https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FSettings%2FToolbarButtons

Check the "Include Text" box and you will see, what I mean. 

I personally think, that this always looks ugly. ... May be in a second PR we should try to make the text a bit smaller and see, how it looks like

pmario:
Also the **right sidebar: tools tab** needs to be checked .... 

and https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FSideBar%2FTools should be checked. The tiddler view isn't used by users, but if it also looks good, we know, that the setting is consistent and hopefully harder to break by "standard users"

pmario:
Those pages: https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FToolbars need to checked if they look good with this PR applied

BlueGreenMagick:
$:/core/ui/Sidebar/Tools, Sidebar view, Control panel

pmario:
@Jermolene ... Looks good to me. I did test the branch with FireFox latest Windows 10  and Edge Win10. Everything looks nice and alligned now

themes/tiddlywiki/vanilla/base.tid
BlueGreenMagick: I think the toolbar looking ugly when 'Include Text' is turned on, is a different issue and out of scope for this PR. It should be done in another PR, probably by someone who has more design sense than me :)

pmario:
@pmario ... note to myself. Check this again!

Jermolene:
Hi @BlueGreenMagick this looks good to go. Please could you resolve the conflicts?

BlueGreenMagick:
Resolved!

Jermolene:
Thanks @BlueGreenMagick

Issue #5013, opened 2020-11-09
[IDEA] Re-do the import UI


saqimtiaz:
There have been various suggestions related to improving the import UI, including being able to choose the deserializer being used, select all/select none buttons, post-import actions, etc.

We should create a flowchart, and consider designing a new import UI from the ground up.

I will attempt to work on this post 5.1.23, as time allows. My primary focus will be UI, so for other changes related to the internals of the import process it may be necessary to create separate issues to discuss and reach consensus.

Please feel free to link here any previous issues related to the import process.

References:
- #5012 
- #4695 
- #4416 
- #4401 
- #2240 
- #1487
- #1219
- #3031

kookma:
Option to set certain tags on import is useful.

AnthonyMuscio:
The pre-release allows rename during import however I would see a lot of value being able to add a prefix to all tiddlers.

This makes me wonder if a new import UI could provide methods to apply a **filter** both to selection within the import list including acting on each tiddler within as if it were a tiddler. For example we could apply a filter that would rename a fieldname within every tiddler. 

To support this we could make it easy to add a set prefix then import the tiddlers directly into a json file/plugin eg $:/import/(n)/tiddlername(s), (this is only matter of renaming and setting a field or two)  Further activity could then act on these shadowtiddlers and create real tiddlers, then subsequently optionally delete the json or plugin tiddler.

The point here is the import tools are also json tiddler tools, and moving between imports, plugins, json and tiddlers (Even zips)  should have a set of tools available as these are tiddlywiki objects. Mohammads Trash plugin moves deleted tiddlers to a JSON, but users could be given the tools to move tiddlers or back them up in any JSON or plugin, and extract them. Now plugins without js tiddlers need not reload the wiki this is even easier.  

Being able to have a data plugin containing the original tiddler data set allowing edit, made into tiddlers and either save back to or restore the data tiddler would be helpful to many different code patters or solutions.

Finally the ability to define tiddlers as containing a drop zone on which a preconfigured import will be used rather than the default would provide a method to create tiddlers with canonical, a username prefix, a date prefix etc... Another method I would like to see is a drop zone to drop the tiddlers, **including the current wikis tiddlers** on and subsequently pack them into a JSON/Plugin or zip.

Much of what I raise here can be done with specialist knowledge however I think JSON, PLugin, Drag and Drop, import and zip files should be first class citizens and have easy batch and interactive tools to manage them.

Tony

[...]

PR #5474, opened 2021-02-04
add possibility to import system tiddlers and allow the import list to be filtered

pmario:
# This PR does the following: 

 - Allows the user to re-activate "disabled" elements
 - Allows the user to filter the list. That's handy if you want to use the "select all" button
    - "select all" will only handle visible elements
 - If the list is filtered a hint will be shown
 - If a filtered list is shown **every selected** element will be imported. Even if it is invisible
 - $:/StoryList and $:/HistoryList are "prefixes" now. 
 - Every system tiddler is default **unchecked**
 - $:/core and $:/Import are blocked

grafik

# Filtered list

grafik

# Version 5.1.23

grafik

### DIFF
--- a/core/modules/upgraders/system.js
+++ b/core/modules/upgraders/system.js
@@ -12,8 +12,8 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 /*global $tw: false */
 "use strict";
 
-var DONT_IMPORT_LIST = ["$:/StoryList","$:/HistoryList"],
-	DONT_IMPORT_PREFIX_LIST = ["$:/temp/","$:/state/","$:/Import"],
+var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
+	DISABLE_SYSTEM_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
 	WARN_IMPORT_PREFIX_LIST = ["$:/core/modules/"];
 
 exports.upgrade = function(wiki,titles,tiddlers) {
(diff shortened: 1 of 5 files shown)


core/modules/upgraders/system.js
```diff
@@ -12,8 +12,8 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 /*global $tw: false */
 "use strict";
 
-var DONT_IMPORT_LIST = ["$:/StoryList","$:/HistoryList"],
-	DONT_IMPORT_PREFIX_LIST = ["$:/temp/","$:/state/","$:/Import"],
+var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
+	DISABLE_SYSTEM_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
```
pmario, line 16: @Jermolene ... It may be possible to make these settings configurable, So the user could define, which elements should be "unselected / disabled" by default.

saqimtiaz:
There are some significant usability concerns with the UX:
- the size and placement of the filter input are all suboptimal, particularly on narrower screens. The UI is already rather cluttered and needs a better layout and spacing.
- the use of the term `filtered` is confusing, as that has a specific meaning in TW
- the fact that selected but invisible tiddlers might be imported suggests the need for an extra step where the list of tiddlers to be imported is shown for verification. Otherwise there is the potential for data loss. 

Do note that I have slowly been working on re-designing the import UI and workflow from scratch for 5.1.24, based on the links to feedback and ideas gathered in the issue description at #5013, from all the previous discussions on this topic that I could find.

If that re-design is to be eventually merged, it will render the changes in this PR obsolete as the implementation and workflow is very different, while addressing many of the same needs and more. It may be prudent and a better use of resources to delay this PR for a while until my proposal for a re-design can be assessed. However if there is a strong desire to merge this now I have nothing against that, but the above mentioned UX concerns need to be addressed.

pmario:
a) Adding the filter was the easiest way to filter for eg: prefix $:/state and then use the "select all" or "unselect all" button. 

We could move the "preview" setting into a "settings" submenu represented by a "cog icon"
We could move the "cancel" and "import" buttons to the bottom.

That will free up space for the search input field at the top, which imo is common UI behaviour.

Instead of the "This list is filtered" I could add a row into the table that says: "... more" with a "down chevron". Clicking this button would clear the filter. ... Also kind of suboptimal. 

The problem with filtered lists is, that if we have long lists, we also need to see that it is filtered at the top of the list. The end of the list isn't good enough for long lists. 

May be we can add some colours. But colours can't be seen by everyone

pmario:
I will remove the UI changes, because what I really need are the code changes.

pmario:
@saqimtiaz ... I did create a new PR: "allow us to import system tiddlers #5479"

I will maintain this PR as "draft" and still push my UI changes. 

Once I do have something that works for my bundler plugin, I'll move the UI to the plugin and close this PR.

pmario:
My bundler-plugin edition will get UI updates. So closing this one!

PR #5475, opened 2021-02-04
Replace \s\s* with \s+ in regex

BlueGreenMagick:
Which achieves the same thing
\* is the 0 or more quanitifier
\+ is the 1 or more quantifier

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -101,7 +101,7 @@ Thanks to Steven Levithan, http://blog.stevenlevithan.com/archives/faster-trim-j
 */
 exports.trim = function(str) {
 	if(typeof str === "string") {
-		return str.replace(/^\s\s*/, '').replace(/\s\s*$/, '');
+		return str.replace(/^\s+/, '').replace(/\s+$/, '');
 	} else {
 		return str;
 	}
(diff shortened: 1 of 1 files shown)


pmario:
Please check your markdown text. IMO not everything is rendered as expected here.

BlueGreenMagick:
Thanks, didn't notice that. Fixed!

simonbaird:
This should be an instant merge IMO. What's the blocker?

simonbaird:
Oh, is it because [the blog post|https://blog.stevenlevithan.com/archives/faster-trim-javascript] says that \s\s* is faster?

That post if from 2007, so I very much doubt it's still relevant. Also I doubt the performance deltas would be noticable in any practical way.

simonbaird:
The benchmarks are Firefox 2 and IE 6...

Jermolene:
Hi @BlueGreenMagick @simonbaird @pmario apologies I missed this. Indeed we'd need more rationale before merging this. Otherwise we'll spend the next 20 years explaining to people why we use regexps wrong.

pmario:
FireFox 88 (latest) got some really cool extensions to regexp matching indexes. see: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/88#javascript and: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec Indices!

BlueGreenMagick:
No need to change what is working I guess!

PR #5478, opened 2021-02-05
rename to-be-defined methods in Story

BlueGreenMagick:
Having 'story' appended in front of all method names seem unnecessary- they are methods in Story class!

Also, do you still intend to move story manipulation to Story class, instead of NavigatorWidget? I've noticed that the code is 6 years old, and while I'd be willing to move the methods over, I wasn't sure of your intentions.

### DIFF
--- a/core/modules/story.js
+++ b/core/modules/story.js
@@ -100,35 +100,35 @@ Story.prototype.addToHistory = function(navigateTo,navigateFromClientRect) {
 	this.wiki.setTiddlerData(this.historyTitle,historyList,{"current-tiddler": titles[titles.length-1]});
 };
 
-Story.prototype.storyCloseTiddler = function(targetTitle) {
+Story.prototype.closeTiddler = function(targetTitle) {
 // TBD
 };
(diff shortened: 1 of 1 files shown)


core/modules/story.js
```diff
@@ -100,35 +100,35 @@ Story.prototype.addToHistory = function(navigateTo,navigateFromClientRect) {
...
 // TBD
 };
 
-Story.prototype.storyCancelTiddler = function(targetTitle) {
+Story.prototype.CancelTiddler = function(targetTitle) {
 // TBD
 };
 
-Story.prototype.storyNewTiddler = function(targetTitle) {
+Story.prototype.NewTiddler = function(targetTitle) {
```
simonbaird, line 131: FWIW it should be lowercase first letter on these to fix the standard naming conventions for method names.

Jermolene:
Hi @BlueGreenMagick once code has been released we are constrained to retain backwards compatibility, and so we can't in general refactor method names.

Issue #758, opened 2014-08-27
Ability for upgrader modules to non-destructively uncheck tiddlers


danielo515:
In order to let third-party developers create their own upgrader modules, this handler at the navigation widget should provide more flexibility. Currently the only way to uncheck a tiddler to be imported is deleting all its fields. This is maybe good for blocking purposes but it is too radical for other purposes. For example unchecking existing tiddlers. I think this can be an easy change with lots of useful possibilities.

pmario:
@Jermolene ... I think this can be closed now, since the desired behaviour will ship with 5.2.0

Jermolene:
Thanks @pmario

PR #5479, opened 2021-02-05
allow us to import formerly blocked system tiddlers

pmario:
# This PR does the following: 

 - Allows the user to re-activate "disabled" elements
 - `$:/StoryList` and `$:/HistoryList` are "prefixes" now. 
 - Every tiddler, that is part of the `UNSELECT_PREFIX_LIST` is default **unchecked**
 -  `$:/Import` is blocked

With TW 5.1.23 the "listed" tiddlers are completely removed from the import-tiddler. So they can't be reactivated anymore. The new import mechanism only disables them, so they can be activated if needed. 

# New Import tiddler

See:  https://github.com/Jermolene/TiddlyWiki5/pull/5479#issuecomment-847388102

### DIFF
--- a/core/modules/upgraders/system.js
+++ b/core/modules/upgraders/system.js
@@ -13,7 +13,7 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 "use strict";
 
 var DONT_IMPORT_LIST = ["$:/Import"],
-	DISABLE_PREFIX_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
+	UNSELECT_PREFIX_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
 	WARN_IMPORT_PREFIX_LIST = ["$:/core/modules/"];
 
 exports.upgrade = function(wiki,titles,tiddlers) {
@@ -26,10 +26,10 @@ exports.upgrade = function(wiki,titles,tiddlers) {
 			tiddlers[title] = Object.create(null);
 			messages[title] = $tw.language.getString("Import/Upgrader/System/Suppressed");
 		} else {
-			for(var t=0; t<DISABLE_PREFIX_LIST.length; t++) {
-				var prefix = DISABLE_PREFIX_LIST[t];
+			for(var t=0; t<UNSELECT_PREFIX_LIST.length; t++) {
+				var prefix = UNSELECT_PREFIX_LIST[t];
 				if(title.substr(0,prefix.length) === prefix) {
-					messages[title] = $tw.language.getString("Import/Upgrader/Disabled/Tiddler");
+					messages[title] = $tw.language.getString("Import/Upgrader/Tiddler/Unselected");
 				}
 			}
 			for(var t=0; t<WARN_IMPORT_PREFIX_LIST.length; t++) {
--- a/core/modules/upgraders/system.js
+++ b/core/modules/upgraders/system.js
@@ -13,7 +13,7 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 "use strict";
 
 var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
-	DISABLE_SYSTEM_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
+	DISABLE_PREFIX_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
 	WARN_IMPORT_PREFIX_LIST = ["$:/core/modules/"];
 
 exports.upgrade = function(wiki,titles,tiddlers) {
@@ -26,8 +26,8 @@ exports.upgrade = function(wiki,titles,tiddlers) {
 			tiddlers[title] = Object.create(null);
 			messages[title] = $tw.language.getString("Import/Upgrader/System/Suppressed");
 		} else {
-			for(var t=0; t<DISABLE_SYSTEM_LIST.length; t++) {
-				var prefix = DISABLE_SYSTEM_LIST[t];
+			for(var t=0; t<DISABLE_PREFIX_LIST.length; t++) {
+				var prefix = DISABLE_PREFIX_LIST[t];
 				if(title.substr(0,prefix.length) === prefix) {
 					messages[title] = $tw.language.getString("Import/Upgrader/System/Disabled");
 				}
--- a/core/modules/upgraders/system.js
+++ b/core/modules/upgraders/system.js
@@ -12,7 +12,7 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 /*global $tw: false */
 "use strict";
 
-var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
+var DONT_IMPORT_LIST = ["$:/Import"],
 	DISABLE_PREFIX_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
 	WARN_IMPORT_PREFIX_LIST = ["$:/core/modules/"];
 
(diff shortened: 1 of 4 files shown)


pmario:
@saqimtiaz ... Please check this one. No UI changes there ;)

core/modules/upgraders/system.js
```diff
@@ -12,8 +12,8 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 /*global $tw: false */
 "use strict";
 
-var DONT_IMPORT_LIST = ["$:/StoryList","$:/HistoryList"],
-	DONT_IMPORT_PREFIX_LIST = ["$:/temp/","$:/state/","$:/Import"],
+var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
+	DISABLE_SYSTEM_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
```
saqimtiaz, line 16: I would suggest renaming this to `DISABLE_SYSTEM_PREFIX_LIST`

saqimtiaz:
@pmario I don't have the opportunity at present to run the code and see it in action, but at a quick glance it looks OK.

The one concern is that some single file wikis can have large numbers of state tiddlers, since they are always saved. So this could lead to a very cluttered and long list when importing from a TW or upgrading. For instance a TW using dynaview can have hundreds of state tiddlers. 

Perhaps we need a way to toggle whether `state` and `temp` tiddlers are available for import? A simple checkbox?
If we don't want to work on the UI for that at present, it could be a config setting that is disabled by default.

saqimtiaz:
PS: personally I would be ok with tackling the UI changes separately, and I could make sure to take this into account with the UI work that I am doing.

pmario:
I'll add a configuration option

pmario:
>The one concern is that some single file wikis can have large numbers of state tiddlers, since they are always saved. So this could lead to a very cluttered and long list when importing from a TW or upgrading. For instance a TW using dynaview can have hundreds of state tiddlers.

That's the same behaviour as in 5.1.23. This PR allows us to "re-enable" them, if we need them. The "old" behaviour did show the elements, but you couldn't use them anymore. 

>Perhaps we need a way to toggle whether state and temp tiddlers are available for import? A simple checkbox?
If we don't want to work on the UI for that at present, it could be a config setting that is disabled by default.

The only way hiding elements with the current UI is physically remove them from the importData. ... But that's not the scope of this PR. 

I think we should go for a different PR that deals with the UI changes.

core/modules/upgraders/system.js
```diff
@@ -13,7 +13,7 @@ Upgrader module that suppresses certain system tiddlers that shouldn't be import
 "use strict";
 
 var DONT_IMPORT_LIST = ["$:/core","$:/Import"],
-	DISABLE_SYSTEM_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
+	DISABLE_PREFIX_LIST = ["$:/temp/","$:/state/","$:/StoryList","$:/HistoryList"],
```
pmario, line 16: @saqimtiaz ... The real name should be DISABLE_PREFIX_LIST. At the moment it's only system tiddlers that are disabled by default. ... But it can be any prefix. So I think that's the right name now.

pmario:
Latest import dialogue

grafik

pmario:
@saqimtiaz @Jermolene ... I think this PR can be merged now. Pls test!

saqimtiaz:
@pmario if I recall correctly, the message "Disabled tiddler" has so far been used for suppressed tiddlers that can no longer be imported. As such we shouldn't be using it for state and temp tiddlers now, and only for tiddlers not available for import. I would suggest having no message for these tiddlers and just have them unselected by default.

Could you please post a TW file for testing, or a screenshot that also shows a tiddler that is suppressed and not available for import? Thank you

pmario:
They are called: "Blocked system tiddler" atm.

pmario:
@saqimtiaz Here is an empty.html + a tiddlers.json, that can be used for testing

[import-test.zip](https://github.com/Jermolene/TiddlyWiki5/files/5972828/import-test.zip)

pmario:
This PR will also fix: Ability for upgrader modules to non-destructively uncheck tiddlers #758

Jermolene:
Thanks @pmario and apologies for the delayed response.

I think the OP might have got out of date: it mentions blocking `$:/core` but I don't see the code that does that.

saqimtiaz:
The rest looks good but we need to be clear on terminology. The use of the word "disabled" in the status message doesn't make sense since these tiddlers aren't disabled, just not selected by default, and changing the selection doesn't change the message.

Suggest dropping the message entirely for state and temp tiddlers, or replacing with messages that explain what state and temp tiddlers are, since many users never realise that they even exist.

pmario:
@Jermolene I did remove the $:/core handling from this PR since it has better handling in the "plugins" upgrader.

pmario:
@saqimtiaz wrote

>... The use of the word "disabled" in the status message doesn't make sense since these tiddlers aren't disabled, just not selected by default, and changing the selection doesn't change the message.

I can change the text to "not selected by default" ... but I have to use a text. The message text is used in the selection logic. 

This PR doesn't change the dynamic behavior of the UI. Changing the selection of any tiddler doesn't change the UI message and it never did.

We may change the UI behavior in a different PR

saqimtiaz:
@pmario just because we internally have to use a message text does not mean we have to show it in the UI. In #5059 we addressed user complaints about the status column being too busy and hard to follow. As a result we have hidden messages that were not relevant or useful. A similar approach to what is used there can be used to hide the message for state and temp tiddlers as well. (Alternatively in navigator.js, after setting the field for the selection state of a tiddler, you can remove the message if it is for a system or state tiddler. Personally I would prefer to do so at the UI level)

The problem with using a status message like "not selected by default" is that it is sure to lead to confusion for users as to why that is the case. However, even internally, that is a far better choice of message than using the word "disabled" since the point of this PR is that they can now be imported and are not disabled or suppressed.

> Changing the selection of any tiddler doesn't change the UI message and it never did.

We have also never had status messages that were related to the selection text, so it was never a problem. Messages like the warning for overwriting tiddlers already are dynamic in that they change if the user renames a tiddler.

I agree that any larger scale UI changes are better handled in a separate PR, but I would suggest that this PR needs to be self contained and include the minimal UI tweaks necessary to ensure a continued good user experience until we can land further changes.

pmario:
@pmario note to myself

Jermolene:
Hi @pmario I'm very keen to regain the ability to reselect disabled import items, and so would like to merge this for v5.2.0, but there's a lot more going on here. @saqimtiaz what do you think?

saqimtiaz:
The issue here is with the UI and the status messages displayed to the user:

- the message shown for state and and temp tiddlers with this PR is `disabled by default`. Except they aren't disabled, they just aren't selected by default. So this message is incorrect. Having no message would be best, as previous feedback from users has been that the messages are too cluttered and in #5059 we addressed these concerns and removed unnecessary and redundant messages.
- @pmario mentions that having a message text is important to select the state internally. However, this does not mean that we need to show it in the UI and it can be hidden at the UI/wikitext level. 
- The problem with using a status message like "not selected by default" is that it is sure to lead to confusion for users as to why that is the case. If we absolutely have to use such a message then it needs to update to reflect the selected/unselected state of the tiddler. We already have such conditional messages for warning about overwriting tiddlers, and this is all possible at the wikitext level.

Jermolene:
Thanks @saqimtiaz. @pmario we need to address those issues before this can be merged.

pmario:
The latest test version including the index.html and a JSON that can be imported is attached.

[import-test.zip](https://github.com/Jermolene/TiddlyWiki5/files/6535223/import-test.zip)

The import-test.JSON should give you this dialogue.

Import-05

---------

The import dialogue without libraries will look like this: 

Import-01

If the user selects an "unselected" element there will be a text "User selected." This allows us to identify elements that have been unselected by default.

Import-02

If elements already exist it looks like this: 

Import-03

Existing and "re-selected" elements will look like this:

Import-04

pmario:
Commit 50e03f7 is already part of the import-test.zip above.

pmario:
UUPs. ... I did include the "bundler-plugin" into my "import-test.JSON" and I did forget, that the bundler actually alters the Import-dialogue. ... 

So do _not_ import the bundler plugin ... atm. I will have to adjust it to 5.2.0 :/ 
Sorry

Jermolene:
Thanks @pmario

PR #5492, opened 2021-02-12
emtpy.html should also have fluid-fixed as default layout

pmario:
# New

grafik

### DIFF
--- a/editions/empty/tiddlers/$__themes_tiddlywiki_vanilla_options_sidebarlayout.tid
+++ b/editions/empty/tiddlers/$__themes_tiddlywiki_vanilla_options_sidebarlayout.tid
@@ -1,0 +1,6 @@
+created: 20210212152738698
+modified: 20210212152738712
+title: $:/themes/tiddlywiki/vanilla/options/sidebarlayout
+type: text/vnd.tiddlywiki
+
+fluid-fixed


pmario:
Just a backlink: https://groups.google.com/g/tiddlywikidev/c/oqDc6cE76xM/m/idX3SBbfBgAJ

NicolasPetton:
I actually find `fluid-fixed` to have quite poor readability on most displays as lines become too wide, so I would vote against it.

TiddlyTweeter:
> @NicolasPetton I actually find `fluid-fixed` to have quite poor readability on most displays as lines become too wide, so I would vote against it.

I agree. And don't. What I mean is that a centered river of determinate (user tweakable from a default, though also responsive to device context) width is optimal.

At the moment (with the defaults as they are) wikis often are cramped. SO. I think @pmario's PR is one, good, step towards a slight rethink of "what width a story river?" 

I do think, from a UI perspective, it does matter quite a lot. 
Something like that.

pmario:
Will close this one, in favour of a different approach described at the GG by Soren, which will need a differnt UI and some different values. 

 - Zone 1, used when viewport width is below minimum side-by-side width:
   - Sidebar above content; both sidebar and content take up full viewport width (as presently).

 - Zone 2, used when viewport width is above minimum side-by-side width but below maximum story river width + fixed sidebar width:
   - Sidebar to right of content; sidebar consumes a configurable fixed width, and story river consumes all remaining available space (as in fluid-fixed presently).

 - Zone 3, used when viewport width exceeds maximum story river width + fixed sidebar width:
   - Sidebar to right of content; story river consumes a configurable maximum width, and sidebar consumes all remaining available space (as in fixed-fluid presently).

PR #5503, opened 2021-02-17
Update TW Icons by morosanuae.tid

morosanuae:
@Jermolene Just a small modification to improve the visibility and a little clarification.


pmario:
Has been changed somewhere else. See: https://tiddlywiki.com/#%22TW%20Icons%22%20by%20morosanuae

Issue #5483, opened 2021-02-08
[BUG] $:/config/OriginalTiddlerPaths - paths outside the wiki folder are encoded.


joshuafontany:
**Describe the bug**
Over on GoogleGroups, there has been some confusion on how to use the new `isEditableFile` feature when importing files external to the Wiki directory. There is also a bug that encodes these filepaths, instead of using the "original" relative path found in `$:/config/OriginalTiddlerPaths`.

**To Reproduce**
1. Setup a node wiki folder as given in https://tiddlywiki.com/#tiddlywiki.files%20Files and https://tiddlywiki.com/#Customising%20Tiddler%20File%20Naming, with tiddlers being loaded from OUTSIDE the Wiki directory.
2. Edit and save an "external tiddler".

**Observed behavior**
The tiddler is saved to the Wiki's `tiddlers/` folder with an encoded filename if the tiddler was read from a file outside the Wiki directory.

**Expected behavior**
The tiddler is saved back to its "original" file location.

Here we have a security concern. If we only rely on `$:/config/OriginalTiddlerPaths` to store the "original" location and a user is able to change the contents of `$:/config/OriginalTiddlerPaths`, then they can acquire write access to anywhere on-disk.

My first instinct is to store and load that info in the internal $tw variables, and use `$:/config/OriginalTiddlerPaths` only as a read-only log of the files loaded and their relative locations to the Wiki directory. Then remove the tiddlers loaded with the `isEditableFile` from any of the $:/config/FileSystem* filters. If we enforce using the original filepath and extension with no way of editing it, that would minimize the security concerns. If the user would like that tiddler's data to be subject to the $:/config/FileSystem* filters, they can re-name the tiddler to a title that does not appear in `$:/config/OriginalTiddlerPaths`.

I think this will be important to implement before we merge in the PR(https://github.com/Jermolene/TiddlyWiki5/pull/5275) that will introduce the `searchSubdirectories` flag, as that will generate a lot more tiddlers that may have `isEditableFile` set...

Working on a PR now.

joshuafontany:
@saqimtiaz It would be great to get your input on this, I know you had a special case where, I think, you were loading .txt files as `type: text/vnd.tiddlywiki` tiddlers, but wanted to save them as *.txt and *.txt.meta files, right? Isn't that why we pass `isEditableTiddler` files to the $:/config/FileSystem* filters?

[...]

PR #5504, opened 2021-02-18
Fix 5483 & 3483

joshuafontany:
Fixes https://github.com/Jermolene/TiddlyWiki5/issues/5483 by storing a relative path in the $tw.boot.files fileInfo objects for any tiddler marked `isEditableFile`. This is replaced with '' (an empty string) if missing, and is then used in the tests that check the "new" filepath against the known writable locations. We have to test against both, so a new check has been added. The previous calls to the tiddler storing this info have been removed as a security vulnerability.

@Jermolene This is good to merge, along with https://github.com/Jermolene/TiddlyWiki5/pull/5416. Mahalo (thanks)!!!


joshuafontany:
This now includes fixes for https://github.com/Jermolene/TiddlyWiki5/issues/3483, and sanitizers for "reserved codes" like "com5" on Windows, or certain character-sets on *nix.

Jermolene:
Thanks @joshuafontany, intricate stuff!

PR #5512, opened 2021-02-22
Introduce browser messaging saver

Jermolene:
This PR introduces a new saver that uses window.postMessage(), enabling TiddlyWiki apps to be constructed that host TiddlyWiki documents in an `<iframe>` element, handling saving on their behalf.

General purpose publish/subscribe classes are introduced. As well as the saver, these classes are used to relay changes to the wiki page title and favicon. I'll add docs for /dev that includes a demonstration host.

This PR is prompted by the [ongoing work](https://groups.google.com/d/msgid/tiddlywiki/333e972b-3796-4595-bc09-af5f7dc659d4n%40googlegroups.com) to integrate TiddlyWiki with [Fission](https://fission.codes/), but is entirely generic, with no dependencies on Fission.

### DIFF
--- a/core/modules/savers/postmessage.js
+++ b/core/modules/savers/postmessage.js
@@ -1,0 +1,66 @@
+/*\
+title: $:/core/modules/savers/postmessage.js
+type: application/javascript
+module-type: saver
+
+Handles saving changes via window.postMessage() to the window.parent
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Select the appropriate saver module and set it up
+*/
+var PostMessageSaver = function(wiki) {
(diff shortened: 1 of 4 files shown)


core/modules/savers/postmessage.js
```diff
@@ -0,0 +1,66 @@
...
+type: application/javascript
+module-type: saver
+
+Handles saving changes via window.postMessage() to the window.parent
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
```
Leilei332, line 12: Function wrappers and some comments should be removed now. And we may also change the code to use class syntax.

Jermolene: Thanks @Leilei332 I’ll attend to this shortly.

PR #5514, opened 2021-02-23
Update chinese language files

BramChen:
* Add chinese descriptions for module-type `utils-browser`
* Fixed typos


Jermolene:
Thanks @BramChen

PR #5521, opened 2021-03-04
Fixed super minor issue with import pragma

flibbles:
When import pragma is alone in a text body...

```
\import something
```

...with no newline following it, then the filter doesn't get added to the parseTreeNode. I realize that this bug doesn't cause any issues for end users, but it may cause trouble for some plugins (*cough* Relink).


pmario:
@Jermolene ... I think this should be merged

Jermolene:
Thanks @flibbles @pmario

PR #5522, opened 2021-03-04
Fix 4461

joshuafontany:
Now uses $tw.syncadaptor.host OR "/".


joshuafontany:
@Jermolene this small fix is ready to go as well

Jermolene:
Thanks @joshuafontany

PR #5542, opened 2021-03-12
fix $tw.wiki.search options.invert init problem

pmario:
In line 1125 the `invert = !!options.invert;` variable is initialized. BUT it was never used. 

If the function is called from a plugin and `options.invert` is **missing** it is `undefined` which causes a problem with the search results. see line 1245

This PR fixes the problem because it uses the `invert` variable

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -1242,7 +1242,7 @@ exports.search = function(text,options) {
 	var results = [],
 		source = options.source || this.each;
 	source(function(tiddler,title) {
-		if(searchTiddler(title) !== options.invert) {
+		if(searchTiddler(title) !== invert) {
 			results.push(title);
 		}
 	});


Jermolene:
Thanks @pmario

pmario:
@Jermolene ... I'm using a call like this: `$tw.wiki.search(text,{caseSensitive:false, field: "title", invert:false, literal:true});` which is fixed to the **title field**. ... The code for the `wiki.search` function is super flexible, but it contains very much code, that isn't needed for "simple" title search. 

So I was thinking, if it would make sense to create a `wiki.searchTitle()` that would only contain the code, that is needed to search the title field. 

I could also modify the `wiki.search()` code, that it uses that function if possible. 

What do you think. Would you be willing to merge such a function to the core?

PR #5550, opened 2021-03-17
Refactored filter tests to use nifty spyOn method

flibbles:
This is a minor refactoring to the test-filters suite. I made that collectLog method a while ago, but while working on Relink, I have since discovered the wonderful little __spyOn__ method supplied by jasmine which does the same thing, only better.

I'm putting this change in so future developers will discover and use __spyOn__ instead of building their own adhoc monkey-patchers like I did.


Jermolene:
Thanks @flibbles

PR #5552, opened 2021-03-18
Change css-escape-polyfill to a tw uitility method

cdruan:
Convert css-escape-polyfill.js into a $tw.utils module so that "escapecss" operator can work under Node.js as well as the browser.

* Add `$tw.utils.escapeCSS()` method and invoke that function within the `escapecss` operator.

* Add test cases for the `escapecss` filter operator.

* Fix `$tw.boot.doesTaskMatchPlatform()` so it works as expected if a module's `export.platforms` contains more than one values


Jermolene:
Thanks @cdruan, an exemplary PR, much appreciated. I apologise for the delay in merging, it had somehow fallen through the cracks. Please don't be shy to ping me in such situations.

Issue #5544, opened 2021-03-13
[BUG] SelectWidget placeholder attribute missing


twMat:
**Describe the bug**
I find no way to show a *placeholder* in the SelectWidget. The underlying html tag can do this using `disabled selected` parameters in the `option` tag. Compare the following by pasting them into a tiddler and observing the behaviour:

```
<select>
<option disabled selected>Select your option</option>
<option>A Tale of Two Cities</option>
<option>A New Kind of Science</option>
<option>The Dice Man</option>
</select>

<$select>
<option disabled selected>Select your option</option>
<option>A Tale of Two Cities</option>
<option>A New Kind of Science</option>
<option>The Dice Man</option>
</$select>
```

You'll note in both setups that the `disabled` parameter does work (shows as gray) but in the SelectWidget it does not show as *selected*. (The SelectWidget `default` parameter also does not *show up* as a placeholder value but, regardless, a deafult is not the same as a placeholder.)

It is very useful with a placeholder to give example values or instructions, especially in a tight layout when there is no room for an instruction next to the widget.

**Desktop (please complete the following information):**
 - Chrome
 - Win10

twMat:
Add: Maybe - I'm not sure, please help me out - maybe it works with a `default` value after all because unless a value is selected, there is *no* value set anywhere. A deafult value does show before anything is selected, functioning as a placeholder.
Is this sufficient or does it risk messing something up? Are there cases where the default value can actually be transmitted as a value without it being actively selected (...which is normally the *point* with a default value). What, actually, *is* the point with a default value in the SelectWidget?

pmario:
with tw widgets, the default value is used, if the config tiddler doesn't exist. As soon as the tiddler exists, this value is obsolete.

twMat:
@pmario - can I read up on config tiddlers anywhere? I was not aware there is another tiddler involved with the SelectWidget selection.

Still, you say "the default value is used, if the config tiddler doesn't exist" - but if the user didn't actually select an option in the selectwidget to *set* a tiddler with some value... then how is the default value accessed by whatever uses the (supposedly) set value? I.e 

```
<$select default="true">
<option>false</option>
</$select>
```

How is that `true` value accessed? This isn't really my OP question but a general question about what the default value actually is/does in this widget.

Jermolene:
[...]

PR #5555, opened 2021-03-20
Update SelectWidget.tid

twMat:
Added instrux for placeholder value. [Ref.](https://github.com/Jermolene/TiddlyWiki5/issues/5544)


Jermolene:
Thanks @twMat

twMat:
@Jermolene  Whupps! Looking at https://tiddlywiki.com/#SelectWidget something is not working correctly. It seems like it is the quote marks for the special demo but even when that is correced, the intended "placeholder" doesn't show. Sorry if I've made a mistake somewhere. Could we revert the merge and I'll fiddle with this tomorrow unless you can tell what the problem is right away. Thank you.

Issue #1676, opened 2015-04-25
Improve core support for static site generation


Jermolene:
Both @welford and @dullroar have done considerable work on improving TiddlyWiki's support for static site generation (SSG). Their sites are http://www.phasersonkill.com and http://dullroar.com respectively.

I think we are now at the point where it would be useful to try to merge the best of @welford and @dullroar's work into the core to make it more accessible for others. I'm envisaging presenting the SSG functionality on tiddlywiki.com, not tiddlywiki.com/dev, and trying to make it as accessible as possible to non-developers.

As a first step, perhaps can we list out the possibilities for the feature set? (I can edit this list to reflect discussion):

* New page templates 
* New themes
* Generate Atom/RSS feed
* Generate `sitemap.xml`
* Stretch goal: A way of doing SSG entirely in the browser

What do others think?

welford:
SSG entirely in the browser would be fantastic. Requiring nodejs, a batch file and a tiddlywiki.info are big hurdles when recommending TW SSG to non programmers. 

I've tried as much as possible with my SSG to have everything available and editable in the actual tiddlywiki itself  and only do the necessary viewtemplate switching at export time.

Jermolene:
> necessary viewtemplate switching

I was curious about that - why not just render a completely independent template tree? A lot of the PageTemplate stuff doesn't apply in the SSG case, and we'd still be able to reuse the core view template.

welford:
I am using a custom template but currently i'm only switching  out $:/core/ui/ViewTemplate/subtitle and $:/core/ui/ViewTemplate/tags  to render slightly differently in my blog. 

It seemed easier to me to keep the rest of the viewtemplate stuff as it was ( and still call <$view tiddler="$:/core/ui/ViewTemplate" format="htmlwikified"/> from the template  ) than have to maintain a separate custom ViewTemplate for the blog.

Jermolene:
That makes sense. A useful refactoring might be to parameterise the view template via variables that can be set outside the transclusion, as we do in `$:/core/save/all` with the filter identifying the tiddlers to be saved. We might use the variables to disable specific view template segments.

dullroar:
On 4/25/2015 2:52 AM, Jeremy Ruston wrote:
> As a first step, perhaps can we list out the possibilities for the 
> feature set? (I can edit this list to reflect discussion):
>
>   * New page templates
>   * New themes
>   * Generate Atom/RSS feed
[...]

PR #5559, opened 2021-03-22
Introduce universal publishing framework for static/custom sites

Jermolene:
This PR aims to make static site generation user friendly and practical in the browser. I also plan to make the same framework function under Node.js both for generating static sites and for serving custom site configurations over HTTP.

The current functionality:

* A new control panel tab exposing the installed static file routes (see below for screenshots)
* A new page control button to initiate publishing through the current static file routes
* A new framework for executing long running tasks while displaying a progress bar, and without locking up the user interface
* A browser-based publisher module in the JSZip plugin that can publish an entire static site as a ZIP file that is automatically downloaded by the browser
* A new framework for TiddlyWiki's JavaScript modules to expose translatable strings. This is needed so that the publisher module [under development for Fission](https://github.com/Jermolene/TiddlyWiki-on-Fission) can be injected into a wiki and bring its customisation user interface with it

Everything is pretty basic at the moment. Planned improvements include:

* Making it possible to pass arbitrary variables to a route
* Refactoring the configuration tiddlers so that they can be reused more easily
* Update the link and image widgets so that they can optionally mangle their targets with a filter for static site generation (an updated version of the existing tv-wikilink-template variable)
* Improve the feedback report provided after publishing
* Improving the control panel UI so that entities can be created and removed as well as edited
* Integrate the progress bar with the existing modal dialogue mechanism
* Add a cancel button to the publishing modal
* Properly deal with errors raised by the publisher modules (at the moment it's best to keep an eye on the JS console)
* Add mechanism to remove files in the output folder that are no longer needed
* Add support for sitemaps
* Add support for RSS/Atom
* Add support for Node.js-based publisher modules, and a means to invoke them from the browser
* Extend the "listen" command to be able to use route configuration tiddlers
* Consider extending the save and render commands to use the publishing framework. The `--output <path>` command would become a synonym for `--publish filesystem <path>`
* Add a GitHub Pages publisher module
* Add an Amazon S3 publisher module
* Possibly add a generic POST/PUT publisher module

<details>
<summary>Screenshots</summary>
Page control button:
image
Control panel tab:
image
Progress bar:
image
</details>


Jermolene:
Note that this PR addresses #1676, which I opened in April 2015!

pmario:
Very interesting

kookma:
Hi @Jermolene 
Is it possible to publish searchable static website? I mean having a small JS script lets search in the static webpages generated by Tiddlywiki.

Jermolene:
> Is it possible to publish searchable static website? I mean having a small JS script lets search in the static webpages generated by Tiddlywiki.

Yes, that will be possible. Hopefully we'll gather a library of templates and components suitable for static publishing.

Jermolene:
I've [pushed](https://github.com/Jermolene/TiddlyWiki5/pull/5559/commits/509356c6964b1c04dbc920e5cb756dceb7f5c0fd) some significant updates.

## Data model refactoring

First, the tiddlers describing the publishing job have been refactored so that they can more easily be reused. Publishing job tiddlers define the tiddlers to be published and the publisher and sitemap to be used. They look like this:

```
title: $:/config/PublishingJobs/Default
tags: $:/tags/PublishingJob
caption: Demo static site
publisher: jszip
sitemap: $:/core/sitemaps/StaticSite
jszip-output-filename: myzipfile.zip
baseurl: https://example.com
enabled: yes
export-filter: [!is[image]!is[system]prefix[T]] [is[image]!is[system]]
```

(The `prefix[T]` is just to keep the size of the export down in my testing; it's still quite a few megabytes).

The sitemap tiddlers look like this:

```
title: $:/core/sitemaps/StaticSite
caption: Static site map
description: The original TiddlyWiki 5 static file layout
tags: $:/tags/SiteMap
list: $:/core/routes/StaticSite/Index $:/core/routes/StaticSite/HTML $:/core/routes/StaticSite/Images $:/core/routes/StaticSite/Styles
var-tv-wikilink-template-filter: [slugify[]addsuffix[.html]]
var-tv-image-template-filter: [slugify[]addprefix[../images/]addsuffix<extension>]
```

Note that fields prefixed `var-` are passed as variables to the rendering process, without the `var-` part.

The `list` field lists the routes that make up the sitemap. For example:

```
title: $:/core/routes/StaticSite/Images
caption: Images
tags: $:/tags/Route
route-type: save
route-path-filter: [slugify[]addprefix[images/]addsuffix<extension>]
route-tiddler-filter: [is[image]]
```

The idea is that sitemaps can be shared across different jobs, and routes can be shared across different sitemaps.

Previously, the JS code didn't know about any of that, it just supported messages for publishing individual routes, assembled in an elaborate action string. Now we do all of that in JS.

## Modal

The modal mechanism is extended to support an optional progress bar. At the moment, only JS code triggering the modal can update the progress bar, but that might be something we want to extend.

## Node.js support

This update adds initial support for Node.js. There are a number of rough edges to explore and fix. For example, the JSZip plugin ought to be able to work under Node.js but currently uses browser techniques to download the finished zip file.

There is also a new "filesystem" publisher module. There's currently no way to trigger it from the browser.

## Image URLs

The image widget is updated to support a new `tv-image-template-filter` filter for converting tiddler titles into URLs. For example, the sitemap above uses `[slugify[]addprefix[../images/]addsuffix<extension>]`.

I intend to make the sample change to the link widget as well.

Jermolene:
Another [update](https://github.com/Jermolene/TiddlyWiki5/pull/5559/commits/df2a3fdefdf9dd6e11695f0bd1a2b56a9a032518) that refactors things to allow the same sitemap tiddlers to drive the web server. The eventual goal is to enable complex static sites like [links.tiddlywiki.com](https://github.com/TiddlyWikiLinks/TiddlyWikiLinks) to function identically when saved as a static site, or when served dynamically.

Getting this working required changing the way that paths are specified. Previously, I'd been using filters like `[slugify[]addprefix[../images/]addsuffix<extension>]` to form an output path from the details of a tiddler. However, the trouble is that we can't run filters in reverse, so there's no way that the webserver can reverse things to find the tiddler corresponding to a particular path. Now, paths are specified as templated strings like `images/$title_slugify$.$type_extension$`, which can be easily converted into a regexp for matching against incoming paths.

Making slugified URLs work with the web server also requires an `unslugify` function that figures out the tiddler title corresponding to a slugified title. There doesn't seem to be a way around having to slugify every title in advance so that we can do the reverse mapping. We do as much caching as possible, but perhaps it will turn into a bottleneck when many thousands of tiddlers are being served.

This still very much a work in progress but now I think we've hit all the big functional goals, and so it should be a matter of cleaning up and fleshing out to get things finished.

I've updated the checklist in the OP.

kookma:
Hi Jeremy
Have you updated the https://tiddlywiki-pr-5559.tiddlyhost.com/ ?
May be it worth to add date or a number to realize which version we are testing at https://tiddlywiki-pr-5559.tiddlyhost.com/

Jermolene:
Hi @kookma 

> Have you updated the https://tiddlywiki-pr-5559.tiddlyhost.com/ ?

No, the latest changes only affect users running under Node.js, who can run it directly.

> May be it worth to add date or a number to realize which version we are testing at https://tiddlywiki-pr-5559.tiddlyhost.com/

I'll make sure to alert everyone here to any material updates.

kookma:
> No, the latest changes only affect users running under Node.js, who can run it directly.

Thank you! So, I have to download the publishing framework branch and test it under Node.JS. IS there any other requirements?



> The eventual goal is to enable complex static sites like [links.tiddlywiki.com](https://github.com/TiddlyWikiLinks/TiddlyWikiLinks) to function identically when saved as a static site, or when served dynamically.

Does this mean the search features of https://github.com/TiddlyWikiLinks/TiddlyWikiLinks will be available for static pages?

Jermolene:
> Thank you! So, I have to download the publishing framework branch and test it under Node.JS. IS there any other requirements?

Yes, you'll need to download the publishing-framework branch if you want to test it under Node.js. Have a look at the docs in the PR: there's a new `--publish` command and a new `sitemap` option for the `--listen` command. 

> Does this mean the search features of [links.tiddlywiki.com](https://github.com/TiddlyWikiLinks/TiddlyWikiLinks) will be available for static pages?

Yes, that's the idea. At the moment it's a pretty dumb search but I'd like to explore using something like https://fusejs.io to give us fuzzy matching.

The other thing I'd like to be able to do is seamless transitions from a static page to the corresponding view in a dynamic TiddlyWiki. Then we'd be able to have, say, column sorting buttons in the static view that transition to the dynamic TiddlyWiki view in order to carry it the sorting.

kookma:
>Yes, that's the idea. At the moment it's a pretty dumb search but I'd like to explore using something like https://fusejs.io to give us fuzzy matching.

Wonderful!

> The other thing I'd like to be able to do is seamless transitions from a static page to the corresponding view in a dynamic TiddlyWiki. Then we'd be able to have, say, column sorting buttons in the static view that transition to the dynamic TiddlyWiki view in order to carry it the sorting.

These are really exciting! I keep eyes open and follow this PR.

kookma:
> Yes, you'll need to download the publishing-framework branch if you want to test it under Node.js. Have a look at the docs in the PR: there's a new `--publish` command and a new `sitemap` option for the `--listen` command.

Having all the above down, starting the TW under node.js when I click the publish button on the sidebar using `filesystem` as publisher I got the following red screen of death error with below message:

`Uncaught Cannot find module named 'fs' required by module '$:/core/modules/publishers/filesystem.js', resolved to fs`

Using `jszip` as publisher works fine! Note i did not used --publish command from command prompt, but I start TW server on node.js and used the publish button!

kookma:
**Issue**

Using JSZIP if you publish an empty wiki (e.g [!is[system]!is[image]] returns nothing) creates the output e.g myzipfile.zip
Opening the index.html from myzipfile.zip results in RSOD with below message

`Uncaught TypeError: Cannot read property 'name' of undefined`

saqimtiaz:
@Jermolene I've been going through an unexpected busy period with work, however I've blocked out time this coming Thursday to go through this in detail.  Expect some questions and feedback at that time.

Jermolene:
> @Jermolene I've been going through an unexpected busy period with work, however I've blocked out time this coming Thursday to go through this in detail. Expect some questions and feedback at that time.

Great, thank you. I think things are in a reasonably coherent shape but will review.

core/modules/publisher-handler.js
```diff
@@ -0,0 +1,181 @@
...
+PublishingJob.prototype.publish = function(callback) {
+	var self = this;
+	// Get the job tiddler and check it is enabled
+	this.jobTiddler = this.publisherHandler.wiki.getTiddler(this.jobTitle);
+	if(this.jobTiddler && this.jobTiddler.fields.enabled === "yes") {
+		// Get the list of tiddlers to be exported, defaulting to all non-system tiddlers
+		this.exportList = this.publisherHandler.wiki.filterTiddlers(this.jobTiddler.fields["export-filter"] || "[!is[system]]");
+		// Get publisher
+		this.publisher = this.getPublisher(this.jobTiddler.fields.publisher);
+		if(this.publisher) {
```
saqimtiaz, line 65: we should also check that the sitemap field exists and refers to an existing tiddler.

core/modules/publisher-handler.js
```diff
@@ -0,0 +1,181 @@
...
+			this.sitemap = new $tw.Sitemap(this.jobTiddler.fields.sitemap,{
+				wiki: this.publisherHandler.wiki,
+				variables: this.publishVariables
+			});
+			this.sitemap.load();
+			// Get the output operations
+			this.operations = this.sitemap.getAllFileDetails(this.exportList);
+			// Display the progress modal
+			if($tw.modal) {
+				this.progressModal = $tw.modal.display(PUBLISHING_MODAL_TITLE,{
```
saqimtiaz, line 76: We may need a means for multiple jobs to use the same modal, if one already exists. Currently with multiple enabled jobs we get one modal over another.

saqimtiaz:
@Jermolene Just a heads up that I am working through this carefully focusing mostly on the architecture and doing a bit of testing of different scenarios along the way to confirm my understanding and assumptions. Will continue after lunch and post final findings and thoughts when I finish that process.

saqimtiaz:
@jermolene feedback and thoughts follow. I apologize if it seems a bit jumbled. I've tried to organize my thoughts but there is a lot to digest and consider. I have a new appreciation for the work that goes into reviewing and merging a large PR, this has taken most of the working day!

I've completed a careful review of the code changes in the PR accompanied by some testing of publishing from the browser and creating my own job for a static site. It all looks coherent and the architecture is logical. I am not seeing any red flags.

**I do think it would be good to clarify the architecture and the role of `publisher` modules and how that compares to `saver` modules.**

As per my understanding, a `publisher` module allows you to save files to a specific platform, the files to be saved are defined by a publisher `job` (which consists of a sitemap and routes).

A `saver` is responsible for saving the single file version of the wiki, and each saver saves to a specific platform.

How do the two relate to each other, and is there potential for refactoring so that the two can share code?

For instance we have a Github saver, and we plan to have a Github publisher module. Could we refactor the common functionality into a library that is used by both. **Or perhaps, the Github saver could be refactored to execute a publishing job that saves the single file version of the wiki?** Implementing the Github publisher would provide a good opportunity to explore these possibilities.

One of the areas we could do with more flexibility and affordances in TW is the ability to generate and save files to different storage backends (local file, zip, Github, S3 etc). Currently the savers are hardwired to save the single file version of the wiki. However, it can be useful to be able to save other files such as:
- a static version of the wiki
- a JSON representation of all the tiddlers
- an RSS file
- images
- backups (single file wiki but to a different file name)

The publishing mechanism seems well engineered and flexible enough to handle these use cases. It looks like existing templates can be re-used which is a big plus. In order to confirm that there aren't any hidden issues or caveats, I plan to test the following (though I likely wont get around to it until tomorrow or perhaps Saturday):
- saving images only via the zip publisher
- saving the single file wiki with a different file name. Apart from saving backups this could be a way to have two ways to save a wiki, e.g. to Github and localfile.
- saving a JSON representation of all of the tiddlers
- saving a single tiddler as a .tid file.

The JSZip publisher module:
- Currently this is hardcoded to trigger a zip file download in the browser.
- However, I can see it being widely useful for most publishers to be able to zip all or certain files from a job before saving them to their respective platform.
- We should explore a way to easily provide this functionality to the different publisher modules instead of having each of them implement this separately.

Other thoughts:
- I concur that being able to trigger nodejs publishers from the browser would be a nice feature. There might be an overlap here with invoking listen commands from the browser.
- Modals: we need a mechanism to use a single modal for showing the progress of multiple concurrent jobs.
- I am wondering if we can find a better name for sitemaps, something not so tied to the idea of static websites and less likely to get confused with XML sitemaps.

What I haven't tested so far is serving the static files using the same sitemap used to publish them.

PR #5567, opened 2021-03-24
Add another podcast link to history

FND:
> plus updated not-so-cool Changelog URL

phrasing is a bit awkward; feel free to change at will


Jermolene:
Thanks @FND, much appreciated.

As a pure docs change, if you can target the tiddlywiki-com branch then it can go live as soon as it is merged. Many thanks!

FND:
Done. It goes without saying that this latest podcast made me somewhat nostalgic (which is why I'll post to Discussions as well later).

Jermolene:
Thanks @FND I'll look forward to it!

Issue #5487, opened 2021-02-10
[BUG] `server-external-js` does not support `path-prefix`


rubaboo:
**Describe the bug**
Trying to run server-external-js edition with `path-prefix` parameter results in "Error: Cannot load tiddlywiki.js" and an empty page.

**To Reproduce**
Create a new external js edition wiki and start it:

```
cd /tmp ; mkdir newwiki || rm -rf newwiki/*
nodejs /path/to/tiddlywiki ./newwiki --init server-external-js
nodejs /path/to/tiddlywiki ./newwiki --listen 'port=8090' 'root-tiddler=$:/core/save/all-external-js'
```

Navigate to wiki at http://localhost:8090
Create a tiddler `$:/config/tiddlyweb/host` with text `$protocol$//$host$/extjs/`
Restart the wiki with path-prefix:

```
nodejs /path/to/tiddlywiki ./newwiki --listen 'port=8090' 'root-tiddler=$:/core/save/all-external-js' 'path-prefix=/extjs'
```
Navigate to http://localhost:8090/extjs.
Observe popup "Error: Cannot load tiddlywiki.js"
Click OK.
Observe empty page.

**Environment**
 - Debian 10
 - Node.js 10.23.1
 - TW 5.1.23

Issue #5343, opened 2020-12-30
server-external-js edition unable to save snapshot


fluxsmith:
The local html file produced when $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/save-snapshot is run is corrupt.  Initially I had missed the documented instructions for populating $:/config/SaveWikiButton/Template with $:/core/save/all-external-js.  Without this it the file produced is unexpectedly small.  After adding the forgoing the file produced is the expected size, but shows no data when opened in a browser.

saqimtiaz:
@fluxsmith which version of TiddlyWiki are you running?

fluxsmith:
5.1.23

Jermolene:
Thanks @fluxsmith do you have the external JS file in the right place for the saved HTML file to access it? Do you see any errors in the browser JabaScript console?

PR #3423, opened 2018-09-05
Add support for externalising TW's JavaScript

Jermolene:
I plan to add support to the web server for serving an externalised JavaScript variant

PR #5570, opened 2021-03-26
Fix faulty external-js single-file wiki

cdruan:
This PR details the modifications I made to create working copies of single-file wikis with external-js setting. The PR also fixes the broken snapshots saved by wikis with external-js setting enabled.

* Add/modfiy templates and helper tiddlers to exclude client-server plugins in the offline version and to specify a working URL for loading tiddlywiki5.js from local disk. Mirror save-all and save-offline templates from the regular TiddlyWeb server.

* To avoid confusion with node.js startup `tiddlywiki.js`, the core JS is renamed from `tiddlylwiki5.js` to `twcore-5.x.x.js` when downloaded to local disk.

* To minimize misconfiguration, the externalized core will preload a copy of `$:/config/SaveWikiButton/Template`  tiddler instructing tiddlywiki to save html using offline-external-js template.

* Documentation on how to acquire core JS for saved snapshots (single file wikis) with the external-js configuration.


editions/server-external-js/tiddlywiki.info
```diff
@@ -13,11 +13,14 @@
 		"listen": [
 			"--listen","root-tiddler=$:/core/save/all-external-js"],
 		"index": [
-			"--rendertiddler","$:/plugins/tiddlywiki/tiddlyweb/save/offline","index.html","text/plain"],
```
pmario, line 16: The name of this edition IMO should be external-js-server

pmario (commented):
I'll test it today.

pmario:
> - Saving and loading "externalized" single-file wikis with the browser

Did you use a browser addON? Which one?

pmario:
If I save it with FF and file-backups addON it saves a "big version" of the index.html and the fails to load. .. I'll investigate

cdruan:
Hi @pmario 

> > * Saving and loading "externalized" single-file wikis with the browser
> 
> Did you use a browser addON? Which one?

I just did direct browser load and save via the "Save Wiki" button (after taking an offline snapshots).

> If I save it with FF and file-backups addON it saves a "big version" of the index.html and the fails to load. .. I'll investigate

Can you please verify your `$:/config/SaveWikiButton/Template` tiddler is set to using this template:`$:/core/save/offline-external-js`? I have a feeling you're still using full-save template. Mine bare bone wiki is about 360KB, and tiddlywiki5.js is 2.2MB I'm unfamiliar with file-backups add-on, but I did a quick test and it worked for me.

pmario:
There is an open PR #3501 from me, which has this problem: https://github.com/Jermolene/TiddlyWiki5/pull/3501#discussion_r228585652 ... That you solved. 

I did solve the "file save problem". .... I don't want to save a snapshot, I want to save a wiki with an external core from a file:/// URI. 

My PR is from 2018, so I have to have a closer look. .... 

BUT it would be nice, if we could combine our 2 PRs. to fix both problems.

pmario:
We should try to rename `tiddlywiki.js`, since this is the same name as the tiddlywiki node project uses. So I think this is confusing.

pmario:
I did use `external-core.js` ... but I'm not really happy with it.

cdruan:
Hi @pmario

> I don't want to save a snapshot, I want to save a wiki with an external core from a file:/// URI.

Is there something not working the way you'd expect? You should be able to save/auto-save your wikis. If not, I'm curious to know what kind of problems you ran into. I added a pre-configured `$:/config/SaveWikiButton/Template` tiddler in this edition's `tiddlers` folder. That should make rolling out external-js easier.

As to renaming tiddlywiki5.js, how about `twcore.js` or `tw-core.js`? 

Thanks for taking the time to test this PR!

pmario:
Your configuration doesn't save. Wit the last commit it create a index.html file with 0 byte size. 

Have a look at my PR. It shows how a single file version needs to work. ... The only problem I had was: https://github.com/Jermolene/TiddlyWiki5/pull/3501#discussion_r228585652  

I'm using file-backups FF browser AddON with auto-save on. ... The wiki saves an empyt.html with about 87kByte and external-core.js with 1.9MByte.

cdruan:
Hi @pmario 

Ok, I'm at a loss... Missing template tiddlers? Could you please do me a favor by starting afresh? Type the following two commands at command line:

```
tiddlywiki.js <wiki-folder> --init server-external-js
```
Verify `tiddlers/SaveWikiButtonTemplate.tid` exists in the wiki-folder. Then run the command:

```
tiddlywiki.js <wiki-folder> --build index
```

The `output` folder should contain `index.html` and `tiddlywiki5.js`. Check index.html's file size before and after saving. My `index.html`'s file size is 357kB.

I've examined your #3501 PR commits. I think we've tried to do similar things with regard to saving. We both tried to instruct `saveWikit()` to use `$:/config/SaveWikiButtonTemplate`, and here's my mod:

```diff
core/modules/saver-handler.js
@@ -157,7 +157,8 @@ SaverHandler.prototype.saveWiki = function(options) {
                return false;
        }
        var     variables = options.variables || {},
-               template = options.template || "$:/core/save/all",
+               template = options.template || 
+                          this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template", "$:/core/save/all"),
                downloadType = options.downloadType || "text/plain",
```

Another difference is that I didn't pull out plugins into core JS because this could create complications when users add/delete plugins down the line.

Thanks!

pmario:
- I did --init and --build. 
 - I did open the index.html in FF
 - I did a "default" browser save back to output
 - file size is 0

pmario:
I'll have to have a closer view. ... My PR is old, so some core templates have been changed. ... But I don't have time for this atm

cdruan:
Hi @pmario 

I figured out! Apparently having a newline character at the end of `SaveWikiButtonTemplate.tid` file breaks the whole thing! I'm used to having `\n` at EOF so I added the char right before commit. Yikes. Sorry I didn't test more thoroughly. I ran through the whole process: git fetched/checked out PR; --init, and --build; save with and without FF addon. Seems good now.

Would you mind testing it again when you have the time :) Thanks!

pmario:
I'm sorry, but it doesn't work for me. Windows 10, FF latest, with or without file-backups AddOn. I always get 0 byte html file.

pmario:
Forget about my last comment. I did it wrong. I only did fetch the source from github, but I didn't do the --init and --build step again, which is important. 

Will test again ;)

pmario:
OK. It works now with or without file-backups. ... The only thing is the tiddlywiki5.js name now, which imo needs to be changed.

pmario:
>Only tested on the following platforms with the basic TW configuration:
>
> -   Saving and loading "externalized" single-file wikis with the browser

This works now for me.

> -   Saving and loading on Windows webdav server, including autosaving.

Which configuration did you use IIS?

> -   Saving and loading in client-server mode, including using custom path-prefix configuration.

Can you post your command ?

> -   Upgrading single-file wiki with the external-js template

What do you mean by "Upgrading"?

pmario:
> - Saving and loading in client-server mode, including using custom path-prefix configuration.

I did 

```
tiddlywiki ./test --init server-external-js
tiddlywiki ./test --build index
tiddlywiki ./test --listen
```

It does what it is supposed to do! ... BUT ... If I create a snapshot, it only saves the tiddlywiki.html file without the core. 

**So the question is**: Should the snapshot save tiddlywiki.htm + xxxx-core.js ... IMO it should

cdruan:
Hi @pmario

Yeah! Thanks for all your patience :)

> > * Saving and loading on Windows webdav server, including autosaving.
> 
> Which configuration did you use IIS?

Yes, IIS webdav server.

> > * Saving and loading in client-server mode, including using custom path-prefix configuration.
> 
> Can you post your command ?

I was testing #5487 and followed the steps posted. The issue was caused by missing the trailing slash in the url.

> > * Upgrading single-file wiki with the external-js template
> 
> What do you mean by "Upgrading"?

I wanted to find out whether an external-js single file wiki could be upgraded to a newer tiddlywiki version later on, so I experimented and came up with the steps. It's not streamlined, but does the job (I think). See my doc `core/language/en-GB/Snippets/ExtJSReadme.tid` in this PR (or read `$:/language/Snippets/ExtJSReadme` tiddler from any wiki generated by this PR) for details. The instruction actually reveals the method to convert a regular single-file wiki to an external core wiki (steps #2 and #3).

> It does what it is supposed to do! ... BUT ... If I create a snapshot, it only saves the tiddlywiki.html file without the core. 
>
> So the question is: Should the snapshot save tiddlywiki.htm + xxxx-core.js ... IMO it should

The doc mentioned above does explain to the users that they need to acquire the JS file. It also links to another doc `core/language/en-GB/Snippets/GetTiddlyWikiJS.tid` (or `$:/language/Snippets/GetTiddlyWikiJS` in the wiki) showing several methods to obtain the JS file.

As to the JS name, how about `tw-core.js`?

cdruan:
Hi @pmario 

By the way, in case you're reading the doc from the offline wiki, you won't see the "download tiddlywiki5.js (link provided) from the server" method--it only appears as a viable option if you're running in the client-server mode.

pmario:
I did test IIS WebDav and it works as expected.

core/language/en-GB/Snippets/ExtJSReadme.tid
```diff
@@ -0,0 +1,15 @@
+title: $:/language/Snippets/ExtJSReadme
```
pmario, line 1: I don't understand, why the additional docs is part of `$:/language/` without language code eg: en-GB here

core/language/en-GB/Snippets/GetTiddlyWikiJS.tid
```diff
@@ -0,0 +1,54 @@
+title: $:/language/Snippets/GetTiddlyWikiJS
```
pmario, line 1: I don't understand, why the additional docs is part of `$:/language/` ... same here. IMO it should be part of TiddlyWiki docs. right?

core/language/en-GB/Snippets/ExtJSReadme.tid
cdruan: The ExtJSReadme should eventually go to TW5.com edition. Since this is a draft PR, I feel it is easier to view the doc if everything is contained in this PR generated wikis.

GetTiddlyWikiJS is meant to be contained in every wiki to facilitate users in acquiring core JS depending on their particular setups. What would be a better title path?

cdruan:
> IMO not possible. Eg: I do have `index-V5.1.23.html` with a `tw-core.js`, which is V5.1.23. ... And I do have an `index-V5.1.24.html` with a `tw-core.js` ... Then that's the wrong one.

I'm a bit confused. Not possible because `tw-core.js` is taken or you wish to contain version number in the JS filename?

cdruan:
Hi @pmario 

Here's my latest round of commits. The main focus is renaming `tiddlywiki5.js` to `twcore-VERSION.js`. If running in client/server "external-js" mode, there is an additional dropdown item under the "cloud" button to help users download tiddlywiki's core JS.

I decided to use `twcore.js` to preload `$:/config/SaveWikiButton/Template` tiddler with the correct external-js template value because forgetting to set this tiddler correctly is the most common error in failed rendering of external-js html--especially when converting from an existing wiki.

cdruan:
Hi @pmario 

> ```
> tiddlywiki ./test --listen
> ```
It didn't catch my attention until now. Don't forget that if you want to start the server with the external-js template, either use `--build listen` target in the server-external-js edition, or use the native `--listen` command with the additional `root-tiddler=$:/core/save/all-external-js` parameter.

pmario:
>It didn't catch my attention until now. Don't forget that if you want to start the server with the external-js template, either use --build listen target in the server-external-js edition, or use the native --listen command with the additional root-tiddler=$:/core/save/all-external-js parameter.

Sorry for the late reply. I do like this trick: `tiddylwiki ./EDITION --build listen` ... Because it also works with windows powershell.

The powershell command with --listen needs to look like this:

```
tiddlywiki ./test --listen root-tiddler='$:/core/save/all-external-js'
```

See the "single quotes" used to cover the root-tiddler parameter **`**$:/...**`**  It's needed that way, otherwise the $ causes problems. .. For powershell, there is a difference between single and double-qutes.

pmario:
@cdruan Cindy, 

I did the following tests today, with the updated commits. 

 - `tiddlywiki ./test --init server-external-js`
 - `tiddlywiki ./test --build index`

   - Did read / write the `file:///` wiki from `./test/output` folder using the browser default save mechanism ... **OK**
   - Did read / write the `file:///` wiki from `./test/output` folder using the file-backups AddOn ... **OK**

 - `tiddlywiki ./test --build listen`
   - Did read / write the nodejs tiddlers ... **OK**
   
 - Did read / write WebDav using IIS ... **OK**
 - Did read / write github test using PUT saver ... **OK** see:  [repo](https://github.com/pmario/test) and [demo](https://pmario.github.io/test/)

----

@Jermolene 
The only thing that I think should be done is, that we create a new favicon.ico for the external-js version. ... It may be part of twcore-VERSION.js ... and may be a "green" one

What do you think?

pmario:
@cdruan ... Are you still with us? ... If you want to enable merging, you'll need to "promote" the PR from draft to "review" state.

cdruan:
> If you want to enable merging, you'll need to "promote" the PR from draft to "review" state.

I see. Thanks, @pmario!

pmario:
@Jermolene ... It would be nice to have this one in 5.2.0

plugins/tiddlywiki/tiddlyweb/syncer-actions-download-twcore.tid
```diff
@@ -0,0 +1,8 @@
+title: $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/download-twcore
+tags: $:/tags/SyncerDropdown
```
Jermolene, line 2: What's the purpose of presenting this menu item? Wouldn't the wiki fail to load if twcore.js was not set up correctly?

core/language/en-GB/Snippets/ExtJSReadme.tid
Jermolene: When docs updates are required to accompany a functional change, then we push the docs change to `master` along with the code changes. So, please refactor the docs to go into the `tw5.com` edition

Jermolene (commented):
Thanks @cdruan @pmario some minor comments.

plugins/tiddlywiki/tiddlyweb/syncer-actions-download-twcore.tid
cdruan: This is optional. The purpose of the menu item "Get twcore.js" is to help users obtain a copy of twcore.js onto their local disk for their html wiki. Due to the essential boot code no longer exists as tiddlers when tiddlywiki starts up with the external-js template, the core JS needs to come from the server directly. I achieve this via an external link to `$:/core/templates/twcore.js` with the download attribute. (Is there another way?)

Ideally, I'd like to fire up the save-file dialog box directly when users click on the menu item, but the "download" behavior is inconsistent across the platforms, so I opted to direct the users to the help page instead.

pmario:
@cdruan ... Is this ready to be reviewed again?

cdruan:
Hi @pmario,  @Jermolene

> ... Is this ready to be reviewed again?

Yes. Sorry, I'm very new to PR / github. I'll be clear next time.

pmario:
@Jermolene ... Are there still some concerns open?

core/language/en-GB/Snippets/GetTiddlyWikiJS.tid
Jermolene: Indeed, this is the wrong namespace, but more than that we don't have docs like this within the core. The only docs that are included are the command descriptions for the `--help` command.

core/templates/external-js/tiddlywiki.js.tid
```diff
@@ -1,4 +1,4 @@
-title: $:/core/templates/tiddlywiki5.js
+title: $:/core/templates/twcore.js
```
Jermolene, line 1: All these template name changes are not backwards compatible. Are they really needed?

plugins/tiddlywiki/tiddlyweb/syncer-actions-download-twcore.tid
Jermolene: OK I think we need to handle this differently. It's too confusing for ordinary users.

Jermolene (commented):
Thanks @cdruan @pmario 

I've noted some specific concerns.

plugins/tiddlywiki/tiddlyweb/syncer-actions-download-twcore.tid
pmario: The line is only added if a node-server where an external core is active. ... IMO It's not shown for ordinary users. It's shown for users that need this option. 

There is an option "Save Snapshot for offline use" ... which only downloads the "html store", without the core ... 

That's why I did request a visible possibility to also download the core. May be "**Save Core for offline use**" would be a less confusing option text.

core/language/en-GB/Snippets/GetTiddlyWikiJS.tid
cdruan: Hi @Jermolene,

Any preference for the location of this help doc? Would `$:/language/Help/external-js/getcore.tid` be OK?

This doc is meant to be contained in every wiki--so moving into tw5.com edition is not ideal. It guides users on how to acquire core JS based on their particular setups. For example, if a wiki is loaded by TiddlyWiki node.js, a link will be available to help users download the core from their servers, but the "Export TiddlylWiki Core" button will only be enabled if the wiki is not currently rendered with `external-js` template. But I think more importantly, this doc hopes to ensure the version of the core the users acquire will match their own wiki versions.

core/templates/external-js/tiddlywiki.js.tid
cdruan: Hi @Jermolene, @pmario 

Restoring the original template name is not a problem. Are we keeping `twcore-<VERSION>.js` as the name of the downloaded core JS file? The filename was changed from `tiddlywiki5.js` to avoid confusion with node.js startup `tiddlywiki.js`.

pmario: We have to keep the version in the core file, otherwise it's not possible to have 2 different files in one directory.

pmario:
Hi, With the commit 9f00f1e the template-renaming concern should be out of the way. ... I didn't test the version yet, but I will soon.

pmario:
@cdruan I will send you a PR with some changes, where I hope we have a better chance to get it merged. 

IMO the main problem with the new "syncer dropdown" element is, that the behaviour is completely new to TW. If a user clicks the button, they expect a "download dialogue". In my PR I'll fix that. 

The documentation problem imo can be solved by merging the info from "Using the external JavaScript template" and "$:/language/Snippets/GetTiddlyWikiJS" into 1 consistent tiddler. 

At the moment, if you read both of them Jeremy is right. It is confusing. I'll try to include a suggestion into my PR too.

cdruan:
Thanks @pmario

cdruan:
Hi @Jermolene (cc @pmario)

With the latest rounds of commits, here's what's changed:

* Syncer dropdown relabeled to **Save TiddlyWiki core for offline use** to make it clearer and it fires a "Save File" dialogue box&mdash;with a cavea that older browsers with likely see full JS displayed instead. It will _download_ the core from the server. As pmario pointed out, this menu item only shows up when the wiki is rendered with the external-js setting, so regular users won't see this at all.
* Tiddler `$:/language/Snippets/GetTiddlyWikiJS` is removed&mdash;so no longer has a namespace issue&mdash;and the contained button **Export tiddlywiki core** is moved to the sidebar > tools section. The button displays a brief description alongside: export tiddllywiki core for external-js wiki. This button _exports core_ from within the wiki itself and is disabled when the wiki has the core externalized. The purpose of this button is to assist users acquire the core without TiddlyWiki on Node.js, and is needed for upgrading single file external-js wiki in the future. (See doc in tw5.com edition) To make the button less conspicuous, it is tugged away, down at the bottom of the list.
* Updated doc in tw5.com edition (much thanks to pmario!)

Jermolene:
Thanks @cdruan much appreciated.

I have one final concern about backwards compatibility of the stylesheet changes. It looks like the concern is to ensure that disabled buttons get the correct text colour, is that correct?

@saqimtiaz I'd welcome your thoughts as well

saqimtiaz:
@Jermolene will review tomorrow when I am more clear-headed.

cdruan:
@Jermolene 

> I have one final concern about backwards compatibility of the stylesheet changes. It looks like the concern is to ensure that disabled buttons get the correct text colour, is that correct?

yes; the CSS affects the disabled sidebar>tools buttons (e.g. export-core-button). Without the modification, the icon would be grayed out but the text remains black. And if the disabled button was checked to appear on the sidebar, the button would change color when hovered.

saqimtiaz:
@cdruan @Jermolene this looks good. It works as advertised though it is important to note the extra `build index` step in initializing a new wiki. The CSS changes look like they should be backwards compatible.

cdruan:
@saqimtiaz @Jermolene @pmario 

> ...though it is important to note the extra build index step in initializing a new wiki. 

Sorry, it took me a while to get it. Actually, no, if you are going to start a new wiki via `--listen root-tiddler=...`, then you do not need the `build index` step. That optional step is for obtaining all required files for external-js version of standalone wiki.

(Did this comment come from reading the tw5.com external-js doc, "Init new wiki" section?) I think I'll edit it a bit...

pmario:
>(Did this comment come from reading the tw5.com external-js doc, "Init new wiki" section?) I think I'll edit it a bit...

If you change the docs, make sure you use single quotes to cover the `'root-tiddler=....'` parameters, otherwise it doesn't work in windows powershell.

cdruan:
User doc edited.

pmario:
I'm OK with the doc changes.

pmario:
@Jermolene ... bump ...

core/language/en-GB/Buttons.multids
```diff
@@ -32,6 +32,8 @@ ExportTiddler/Caption: export tiddler
 ExportTiddler/Hint: Export tiddler
 ExportTiddlers/Caption: export tiddlers
 ExportTiddlers/Hint: Export tiddlers
+ExportTWCore/Caption: export tiddlywiki core
```
Jermolene, line 35: "~TiddlyWiki" should be capitalised, with the tilde character to prevent camelcase linking

core/language/en-GB/Buttons.multids
```diff
@@ -32,6 +32,8 @@ ExportTiddler/Caption: export tiddler
...
+ExportTWCore/Hint: Export tiddlywiki core for external-js wiki
```
Jermolene, line 36: Perhaps "Export the ~TiddlyWiki core code for running with external ~JavaScript"

core/language/en-GB/Buttons.multids
Jermolene: The wording "TWCore" is awkward, I think it would better to spell it out as "ExportTiddlyWikiCore/Caption" etc.

core/templates/external-js/save-offline-external-js.tid
```diff
@@ -0,0 +1,8 @@
+title: $:/core/save/offline-external-js
+
+\import [[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]
+\define saveTiddlerFilter()
+[is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/core]] -[[$:/plugins/tiddlywiki/filesystem]] -[[$:/plugins/tiddlywiki/tiddlyweb]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
+\end
+\define coreURL() twcore-$(version)$.js
```
Jermolene, line 7: Again, I think it's better to spell out "tiddlywikicore"

core/ui/PageControls/export-twcore.tid
```diff
@@ -0,0 +1,19 @@
+title: $:/core/ui/Buttons/export-twcore
```
Jermolene, line 1: Again, "twcore" should be spelt out in full as "tiddlywikicore" throughout this file

core/wiki/config/PageControlButtons.multids
```diff
@@ -4,6 +4,7 @@ core/ui/Buttons/advanced-search: hide
 core/ui/Buttons/close-all: hide
 core/ui/Buttons/encryption: hide
 core/ui/Buttons/export-page: hide
+core/ui/Buttons/export-twcore: hide
```
Jermolene, line 7: Again, twcore -> tiddlywikicore

core/wiki/tags/PageControls.tid
```diff
@@ -1,2 +1,2 @@
...
+list: [[$:/core/ui/Buttons/home]] [[$:/core/ui/Buttons/close-all]] [[$:/core/ui/Buttons/fold-all]] [[$:/core/ui/Buttons/unfold-all]] [[$:/core/ui/Buttons/permaview]] [[$:/core/ui/Buttons/new-tiddler]] [[$:/core/ui/Buttons/new-journal]] [[$:/core/ui/Buttons/new-image]] [[$:/core/ui/Buttons/import]] [[$:/core/ui/Buttons/export-page]] [[$:/core/ui/Buttons/control-panel]] [[$:/core/ui/Buttons/advanced-search]] [[$:/core/ui/Buttons/manager]] [[$:/core/ui/Buttons/tag-manager]] [[$:/core/ui/Buttons/language]] [[$:/core/ui/Buttons/palette]] [[$:/core/ui/Buttons/theme]] [[$:/core/ui/Buttons/storyview]] [[$:/core/ui/Buttons/encryption]] [[$:/core/ui/Buttons/timestamp]] [[$:/core/ui/Buttons/full-screen]] [[$:/core/ui/Buttons/print]] [[$:/core/ui/Buttons/save-wiki]] [[$:/core/ui/Buttons/refresh]] [[$:/core/ui/Buttons/export-twcore]] [[$:/core/ui/Buttons/more-page-actions]]
```
Jermolene, line 2: Again, twcore -> tiddlywikicore

editions/server-external-js/tiddlywiki.info
```diff
@@ -11,13 +11,16 @@
...
+			"--rendertiddler","$:/core/save/offline-external-js","index.html","text/plain",
+			"--render","$:/core/templates/tiddlywiki5.js","[[twcore-]addsuffix<version>addsuffix[.js]]","text/plain"],
 		"static": [
 			"--rendertiddler","$:/core/templates/static.template.html","static.html","text/plain",
 			"--rendertiddler","$:/core/templates/alltiddlers.template.html","alltiddlers.html","text/plain",
 			"--rendertiddlers","[!is[system]]","$:/core/templates/static.tiddler.html","static","text/plain",
-			"--rendertiddler","$:/core/templates/static.template.css","static/static.css","text/plain"]
+			"--rendertiddler","$:/core/templates/static.template.css","static/static.css","text/plain"],
+		"twcore": [
+			"--render","$:/core/templates/tiddlywiki5.js","[[twcore-]addsuffix<version>addsuffix[.js]]","text/plain"]
```
Jermolene, line 24: Again, twcore -> tiddlywikicore

editions/tw5.com/tiddlers/webserver/Using the external JavaScript template.tid
```diff
@@ -1,23 +1,128 @@
 created: 20180905075846391
-modified: 20200814110131083
+modified: 20210611055708739
 tags: [[WebServer Guides]]
 title: Using the external JavaScript template
 type: text/vnd.tiddlywiki
 
-Setting the [[root-tiddler|WebServer Parameter: root-tiddler]] parameter to `$:/core/save/all-external-js` switches to a special template that externalises TiddlyWiki's core ~JavaScript into a separate file. For example:
+\define break-section()
```
Jermolene, line 7: It's really important that the TiddlyWiki documentation is consistent. Here you've introduced a new system for navigating within this tiddler. That may be a useful innovation, but only if it is used consistently; allowing each documentation tiddler to evolve their own non-standard navigation techniques makes things difficult for users.

So, here, either we should use one long tiddler, or we should break it up into separate tiddlers.

editions/tw5.com/tiddlywiki.info
```diff
@@ -52,8 +52,8 @@
 			"--rendertiddlers","[!is[system]]","$:/core/templates/static.tiddler.html","static","text/plain",
 			"--rendertiddler","$:/core/templates/static.template.css","static/static.css","text/plain"],
 		"external-js": [
-			"--rendertiddler","$:/core/save/all-external-js","index.html","text/plain",
-			"--rendertiddler","$:/core/templates/tiddlywiki5.js","tiddlywiki.js","text/plain"]
+			"--rendertiddler","$:/core/save/offline-external-js","index.html","text/plain",
+			"--render","$:/core/templates/tiddlywiki5.js","[[twcore-]addsuffix<version>addsuffix[.js]]","text/plain"]
```
Jermolene, line 56: Again, twcore -> tiddlywikicore

plugins/tiddlywiki/tiddlyweb/syncer-actions-download-twcore.tid
```diff
@@ -0,0 +1,7 @@
...
+
+\define twcoreFilename() twcore-$(version)$.js
```
Jermolene, line 4: Again, twcore -> tiddlywikicore

plugins/tiddlywiki/tiddlyweb/tags-syncerdropdown.tid
```diff
@@ -1,2 +1,2 @@
 title: $:/tags/SyncerDropdown
-list: $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/login-status $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/login $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/refresh $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/logout $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/save-snapshot $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/copy-logs
+list: $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/login-status $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/login $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/refresh $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/logout $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/save-snapshot $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/download-twcore $:/plugins/tiddlywiki/tiddlyweb/syncer-actions/copy-logs
```
Jermolene, line 2: Again, twcore -> tiddlywikicore

core/ui/PageControls/export-twcore.tid
```diff
@@ -0,0 +1,19 @@
...
+tags: $:/tags/PageControls
+caption: {{$:/core/images/star-filled}} {{$:/language/Buttons/ExportTWCore/Caption}}
+description: {{$:/language/Buttons/ExportTWCore/Hint}}
+
+\whitespace trim
+\define jsFileName() twcore-$(version)$.js
+
+<$set name=isButtonOff filter="[[$:/boot/boot.js]is[missing]then[yes]else[no]]">
+<$button tooltip={{$:/language/Buttons/ExportTWCore/Hint}} aria-label={{$:/language/Buttons/ExportTWCore/Caption}} class=<<tv-config-toolbar-class>> disabled=<<isButtonOff>>>
```
Jermolene, line 10: It introduces quite a lot of complexity to handle the disabled button status, is it really necessary?

Jermolene (commented):
Thanks @cdruan just a few points, but I'm afraid one is quite big: I've come to think that the abbreviation "twcore" is too generic, we should use "tiddlywikicore". Hopefully it's just a matter of some search and replace, but I appreciate it's a nuisance.

I apologise again that this process is so time consuming and intricate. The issue is that this PR involves a few deep changes to the core, and our commitment to backwards compatibility means that I and the other core maintainers will have to maintain this code for ever. That in turn means that we have to dive in just as deeply as if we'd written the code ourselves.

Anyhow, these are worthwhile changes that will make a difference to a quite a few users, and I thank you again for your patience as we work through things.

core/ui/PageControls/export-twcore.tid
cdruan: Hi @Jermolene 

```
\define jsFileName() tiddlywikicore-$(version)$.js

<$set name=isButtonOff filter="[[$:/boot/boot.js]is[missing]then[yes]else[no]]">
<$button tooltip={{$:/language/Buttons/ExportTWCore/Hint}} aria-label={{$:/language/Buttons/ExportTWCore/Caption}} class=<<tv-config-toolbar-class>> disabled=<<isButtonOff>>>
```
> It introduces quite a lot of complexity to handle the disabled button status, is it really necessary?

I don't quite follow your concerns here. What complexity are you referring to? If it's whether the button needs to be disabled in certain situations, then I would argue yes--if the wiki's core has been externalized; otherwise, the users may not realize the exported core from an externalized-core wiki is not usable.

saqimtiaz: If we do need to disable the button we should use a filtered transclusion instead of a set widget:

`<$button tooltip={{$:/language/Buttons/ExportTWCore/Hint}} aria-label={{$:/language/Buttons/ExportTWCore/Caption}} class=<<tv-config-toolbar-class>> disabled={{{[[$:/boot/boot.js]is[missing]then[yes]else[no]]}}}>`

Jermolene: >  Sorry, I don't quite follow your concerns here. What complexity are you referring to? If it's whether the button needs to be disabled in certain situations, then I would argue yes--if the wiki's core has been externalized; otherwise, the users may not realize the exported core from an externalized-core wiki is not usable.

The core doesn't currently use the disabled attribute on toolbar buttons, which is why you had to add a CSS class to help style it. Part of the reason that we don't use them is because testing shows that users can find it hard to distinguish disabled buttons from ordinary buttons, and that the behaviour of a disabled button is confusing because it feels like the mouse isn't working. If there's an error condition that prevents a button from being used then I think it's better to display an error message when the button is clicked.

The complexity I was referring to was in the implementation details, but perhaps a bigger concern is usability.

cdruan: > ... we should use a filtered transclusion instead of a set widget:

Thanks @saqimtiaz. I'll keep that mind for the future.

cdruan:
Hi @Jermolene 

I've addressed your most recent concerns:

* Renamed twcore to tiddlywikicore.
* Reformatted the user doc to be consistent with the rest of tw5.com documentation.
* Reworked "Export TiddlyWiki Core" button. Now it pops up a message when export-core cannot be done. No more disabled button state. (Thanks for the idea!)
* Reverted the commit that changed vanilla stylesheet.

Please let me know if you have any more issues.

Jermolene:
Thanks for all your hard work @cdruan, I know a lot of people will value this update.

Jermolene:
I am afraid that I had missed the fact that this PR introduces a new page control button that is visible to all users. Here is a screenshot from https://tiddlywiki.com/prerelease

I will have to fix this. It is obviously unacceptable to burden all users of TiddlyWiki with the cognitive burden of additional page control button that only applies to a tiny percentage of users.

@cdruan just so that I am not missing anything, please can you explain the rationale for including this button in the core? Why could it not be part of the tiddlyweb plugin? Why does it have to be a page control button at all?

Screenshot 2021-08-03 at 20 58 47

cdruan:
Hi @Jermolene 

The purposes of the "export TiddlyWiki core" button are:

* To assist users acquire the core *without* TiddlyWiki on Node.js
   A keen user may realize all you have to to is to follow latter part of the upgrade process to convert any single-file wiki to using the external-js template&mdash;without the need of going through the node.js route.
* To facilitate the users upgrade their single file external-js wikis in the future

I don't think it belongs to the tiddlyweb plugin. Ideally, you want to obtain the core JS from your own wiki to prevent version mismatch so the button should be available somewhere in their wiki. Originally, I had the button embedded in the ` $:/language/Snippets/GetTiddlyWikiJS` name space.

I've purposefully moved the button down to the bottom of the page-control buttons list to make it less obvious, but if you really want to remove this button, you would also need to modify the "upgrade single-file wiki" section of the "external-js" documentation. Perhaps, in place of *clicking the "extract TiddlyWiki core" button to obtain the new core*, instruct the users to issue the following command to extract the core from their single file wiki?

```
<$action-sendmessage $message="tm-download-file" $param="$:/core/templates/tiddlywiki5.js" filename="tiddlywikicore-5.2.0.js"/>
```
 Might be good to emphasize that it is necessary to obtain the core before they save their wiki with the `$:/config/SaveWikiButton/Template` tiddler updated to `$:/core/save/offline-external-js` value.

Commit "fc5fa28" of this PR was the instance that created this page control button. If revert doesn't work, at least it lists all of the related files.

pmario:
@Jermolene I think the external-js functionality is extremely useful for everyone that has restrained internet access. It allows users to have the TW core in the browser cache. It will considerably speed up the loading time of the wilki for those users.

It also is a big advantage for everyone that wants to provide a hosting platform like TiddlyHost. It helps the provider to save outbound bandwidth and it speeds up the wiki loading experience. 

It will also help everyone, that wants to use TiddlyWiki with new tech as the hyperdrive technology. Hyperdrive was DAT that the beaker browser uses. Hyperdrive uses a "copy on write" mechanism with an intrinsic versioning system. Saving the core, every time the content is changed only uses more harddisc space, without any other advantage. So using an external core will be an advantage for every user of this new technology. 

IMO the "timestamp" button is also a cognitive burden for every user and only used by a tiny fraction of the TW user base. It is known to cause problems if users activate it and then forget about it. We do have several requests in the GG that show that. But the button is still a page control button. My question is why?

Jermolene:
Thanks @cdruan @pmario 

> The purposes of the "export TiddlyWiki core" button are:
> 
> * To assist users acquire the core _without_ TiddlyWiki on Node.js
>   A keen user may realize all you have to to is to follow latter part of the upgrade process to convert any single-file wiki to using the external-js template—without the need of going through the node.js route.
> * To facilitate the users upgrade their single file external-js wikis in the future

I don't understand the decision to use a page control button to download the core, rather than, say, a link inside control panel?

> I don't think it belongs to the tiddlyweb plugin. Ideally, you want to obtain the core JS from your own wiki to prevent version mismatch so the button should be available somewhere in their wiki. Originally, I had the button embedded in the ` $:/language/Snippets/GetTiddlyWikiJS` name space.

The issues here are entirely my fault. I hadn't checked this PR carefully enough before merging. That's often a problem, because there is a lot of pressure to merge longstanding PRs. The truth is that merging a PR takes me more work and effort than it would to write the same functionality for myself. I've come to accept that as the price of this kind of collaboration.

In this case, I hadn't appreciated that the intent was to provide support for operating the single file configuration of TiddlyWiki with an externalised JS file. I thought it was just adding support for externalising JS in the client-server configuration. Hence my suggestion of putting in the tiddlyweb plugin.

> @Jermolene I think the external-js functionality is extremely useful for everyone that has restricted internet access. It allows users to have the TW core in the browser cache. It will considerably speed up the loading time of the wilki for those users.
> 
> It also is a big advantage for everyone that wants to provide a hosting platform like TiddlyHost. It helps the provider to save outbound bandwidth and it speeds up the wiki loading experience.
> 
> It will also help everyone, that wants to use TiddlyWiki with new tech as the hyperdrive technology. Hyperdrive was DAT that the beaker browser uses. Hyperdrive uses a "copy on write" mechanism with an intrinsic versioning system. Saving the core, every time the content is changed only uses more harddisc space, without any other advantage. So using an external core will be an advantage for every user of this new technology.

You make a reasonable case for why externalised JS can be useful, but that doesn't address the concerns that I raised above about why the download functionality needs to be a page control button.

In any case, the complexity of setting up the single file edition with externalised JS is demonstrated by the complexity of the discussion here. Given the difficulties many users have with saving changes in the usual way, it's hard not to imagine that this externalised JS configuration is going to be far too complex for most users.

The core is supposed to be universal and eternal. I don't think these features should be presented as being universally useful to all users because they make TW too brittle and hard to use. I don't want to have to support those users into the far future.

So, at the moment I am minded to revert everything here. I'd like to get to the point where people can use a separate plugin to get this functionality.

> IMO the "timestamp" button is also a cognitive burden for every user and only used by a tiny fraction of the TW user base. It is known to cause problems if users activate it and then forget about it. We do have several requests in the GG that show that. But the button is still a page control button. My question is why?

With respect, that's not a good argument here. If there's a problem with the "timestamp" button then we should address it directly, rather than using it as a rationale for accepting new features without question.

pmario:
IMO it will be easy to remove the button from the "Tools" tab. We only need to remove the tag that shows it there. The button will be still usable in the documentation, where we describe the functionality. In that part of the docs we can add a possibility for users to add the button back to the Tools tab. So only those users who want that function need to see the button.

pmario:
>So, at the moment I am minded to revert everything here. I'd like to get to the point where people can use a separate plugin to get this functionality.

I really beg you to think about this again. As I wrote, if the Tools button is that big of a problem, it will be easy to remove it. Just remove the tag.

We (those who need the functionality) can put it back by changing the TW docs only.

The functionality itself imo needs the core template changes. Especially as I think it has been implemented in a sensible way.

pmario:
>In any case, the complexity of setting up the single file edition with externalised JS is demonstrated by the complexity of the discussion here. 

You are right, it is more complex, than handling a single file wiki. But imo it is also much less complex than a node-js version. Node-js had a chance. IMO this functionality should also have a chance to be used and improved.

>Given the difficulties many users have with saving changes in the usual way, it's hard not to imagine that this externalised JS configuration is going to be far too complex for most users.

I think this implementation did move most of the complexity into the documentation and there we are able to improve it.

cdruan:
Hi @Jermolene 

Please keep in mind that this PR was meant to "fix" the existing external-js functionaly. Currently (as of version 5.1.23), the saved snapshots of an external-js client/server wiki cannot be opened successfully (#5343); furthermore, the documentation lacks instructions on how to obtain the core js to accompany the saved snapshot. 

> I don't understand the decision to use a page control button to download the core, rather than, say, a link inside control panel?

I'm not sure where, but it should be fine.

Jermolene:
Thanks @cdruan @pmario 

This is entirely my fault for not spending more time studying this PR before merging.

> Please keep in mind that this PR was meant to "fix" the existing external-js functionaly. Currently (as of version 5.1.23), the saved snapshots of an external-js client/server wiki cannot be opened successfully (#5343); furthermore, the documentation lacks instructions on how to obtain the core js to accompany the saved snapshot.

Yes indeed, that was my understanding when this PR was opened. The external-js functionality in #3423 was an experiment; I wanted to show users who were asking that it was fairly trivial to do, but it was fairly well hidden, and not directly exposed to casual users.

> You are right, it is more complex, than handling a single file wiki. But imo it is also much less complex than a node-js version. Node-js had a chance. IMO this functionality should also have a chance to be used and improved.

I don't agree with all of that.

The Node.js configuration of TiddlyWiki is conceptually as simple to use as any local web server based app. The only complexity over and above the single file version is the requirement to be able to use the command line to issue a command. Being comfortable on the command line isn't universal, but TiddlyWiki places the minimum possible burden on users who do have that familiarity.

The whole point of TiddlyWiki's simplicity is that it is a single file. Adding a dependency so that two files is required might seem like a harmless way of optimising storage usage, but it has a profound impact on usability.

Anybody who uses TiddlyWiki can understand the idea that their stuff is saved in a single file, because that's just how Word documents have worked for 30 years.

There is no precedent for most users for a dependency that must be present for a file to work. It requires a high level conceptual understanding of the web to understand how file dependencies work.

The failure mode of the external JS configuration is utterly disastrous from a user perspective: a blank screen, with no way of getting it to work.

So, I believe that TiddlyWiki should not support external JS as a standard configuration. It's just not worth the extra complexity, and the configuration breaks TiddlyWiki's fundamental principle of being self-contained.

> I think this implementation did move most of the complexity into the documentation and there we are able to improve it.

Actually, the docs in this PR are highly complex and unapproachable for all but the most determined users. Necessarily, because what they're putting across is complex and unintuitive.

core/modules/saver-handler.js
```diff
@@ -157,7 +157,8 @@ SaverHandler.prototype.saveWiki = function(options) {
 		return false;
 	}
 	var	variables = options.variables || {},
-		template = options.template || "$:/core/save/all",
+		template = (options.template || 
+		           this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template","$:/core/save/all")).trim(),
```
pmario, line 161: @Jermolene ... If you really want to remove the whole thing, imo these 2 lines are needed, because plugins have user facing way, to tell the core, which template to use. ... The rest of the code can go into a plugin.

Jermolene:
Hi @cdruan @pmario thanks for your patience. I'm about to commit the changes discussed above:

* Removed the `export-tiddlywikicore` page control button
* Removed the `download-tiddlywikicore` syncer action
* Added a new hidden UI for downloading the core JS in `$:/core/ui/ExportTiddlyWikiCore`

I've also overhauled the documentation tiddler. I'd appreciate your review if you have a chance.

Many thanks for your help.

cdruan:
Hi @Jermolene 

There is one more change you'd need to make:

In the last section of the documentation tiddler&mdash; **Upgrading your standalone wiki**, step 3&mdash;you would need to ask the users to visit ` $:/core/ui/ExportTiddlyWikiCore` instead.

PR #5574, opened 2021-03-29
This PR add tc-tiny-gap-xxx to the vanilla theme

pmario:
This PR add tc-tiny-gap-xxx to the vanilla theme.  It is similar to &nbsp; and should replace it in the TW UI


Jermolene:
Thanks @pmario

PR #5587, opened 2021-04-06
Merge Client Tweak page into Control Panel

donmor:
Once I saw [this issue](https://github.com/donmor/TiddloidLite/issues/6) and decided to make new versions of Tiddloid/Lite. To make configurations easy, I created this PR for a tweak page in Control Panel, and now a wiki that properly configured looks like this:
Screenshot_1617682448
Screenshot_1617682555

### DIFF
--- a/languages/de-DE/ControlPanel.multids
+++ b/languages/de-DE/ControlPanel.multids
@@ -25,6 +25,12 @@ Basics/Tiddlers/Prompt: Anzahl Tiddler
 Basics/Title/Prompt: Titel dieses ~TiddlyWikis
 Basics/Username/Prompt: Benutzersignatur zum Editieren
 Basics/Version/Prompt: ~TiddlyWiki Version
+Client/Caption: Client-Optimierungen
+Client/Hint: Hier können Sie das Erscheinungsbild Ihres Kunden anpassen.
+Client/Tiddloid/Caption: Tiddloid
+Client/Tiddloid/Hint: Diese Einstellungen gelten für den Tiddloid-Client.
+Client/Tiddloid/ApplyTheme/Description: Wenden Sie die Themenfarbe auf die Systemleisten an
+Client/Tiddloid/HideToolbar/Description: Ausblenden Symbolleiste auf der Ladevorgang abgeschlossen
 EditorTypes/Caption: Editor Typen
 EditorTypes/Editor/Caption: Editor
 EditorTypes/Hint: Diese Tiddler definieren, welcher Editor für bestimmte Tiddler Typen (MIME-Type) verwendet werden soll.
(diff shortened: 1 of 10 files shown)


languages/de-DE/ControlPanel.multids
```diff
@@ -25,6 +25,12 @@ Basics/Tiddlers/Prompt: Anzahl Tiddler
 Basics/Title/Prompt: Titel dieses ~TiddlyWikis
 Basics/Username/Prompt: Benutzersignatur zum Editieren
 Basics/Version/Prompt: ~TiddlyWiki Version
+Client/Caption: Client-Optimierungen
+Client/Hint: Hier können Sie das Erscheinungsbild Ihres Kunden anpassen.
```
pmario, line 29: I'm not sure if Client, should be translated to "Kunden" here.

languages/de-DE/ControlPanel.multids
```diff
@@ -25,6 +25,12 @@ Basics/Tiddlers/Prompt: Anzahl Tiddler
...
+Client/Tiddloid/Caption: Tiddloid
+Client/Tiddloid/Hint: Diese Einstellungen gelten für den Tiddloid-Client.
+Client/Tiddloid/ApplyTheme/Description: Wenden Sie die Themenfarbe auf die Systemleisten an
+Client/Tiddloid/HideToolbar/Description: Ausblenden Symbolleiste auf der Ladevorgang abgeschlossen
```
pmario, line 33: IMO this isn't proper German. ... And I'm not sure what is meant. Even the English text isn't clear to me

pmario:
What is "Tidloid Client"? ... Shouldn't this be a plugin?

languages/de-DE/ControlPanel.multids
donmor: Okay...I'm not German speaker XD I just uses google translate

donmor: The first checkbox is to hide the appbar of Tiddloid if the file is completely loaded, and the second is to let appbar and systembar be the same color as the background of the page, as shown in the picture.

donmor:
I think it is rather a framework, where configuration pages for other end-user apps could be added.
Or maybe it should be moved to `saving` tab? But these options are not related to saving.

donmor:
So what should I do to make the Tiddloid tweak page available in next TW5 update? If Appearance is not suitable, which tab should I place the panel in? Looking forward to your answer.

donmor:
I created another PR (#5682), which packs the contents into an independent plugin. It's expected to merge one of the two PRs and reject another.

Jermolene:
Hi @donmor this PR introduces a new Control Panel tab called "Client", which is a new usage of that term in the context of TiddlyWiki. I think in fact, Tiddloid is a saver, and its configuration should go under the "Saving" tab.

But a deeper problem is that this PR brings makes too strong a dependency between the core and Tiddloid. The core is intended to be "universal", by which I mean that everything in the core is potentially useful to all users. The exceptions are small: for years we've accepted third party savers for the core to make things convenient for the core, but clearly it's not scalable to move the UI of every saver into the core.

PR #5597, opened 2021-04-10
Extend Dropzone widget

saqimtiaz:
This PR extends the `<$dropzone>` widget to:

- optionally invoke actions after the `tm-import-tiddlers` message has been sent. This allows triggering an alternative UX and workflow to the default import process.
- specify an optional `contentTypesFilter` which determines which content types are accepted by the dropzone. (note that the parameter is intentionally named for consistency with existing documentation https://tiddlywiki.com/#ContentType rather than introducing the term mimeType for end-users). Using a filter offers flexibility such as using the following for all image types: `[prefix[image/]]`.

Note that the `filterByContentTypes` method does the filtering just before calling `tm-import-tiddlers`. Filtering here instead of earlier in the event handling allows:
- a consistent and unified handling of drop and paste events by the same method.
- removes the need to handle browser inconsistencies in the information provided in the drop and paste events.
- allows the logic in our deserializers to attempt to correctly identify the content types in the payload.
- prevents the default browser behaviour regardless of mimeType, such as replacing the TW file with the dropped content or opening the content in a new tab, which is otherwise disruptive to the user experience.

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -12,6 +12,8 @@ Dropzone widget
 /*global $tw: false */
 "use strict";
 
+var IMPORT_TITLE = "$:/Import";
+
 var Widget = require("$:/core/modules/widgets/widget.js").widget;
 
 var DropZoneWidget = function(parseTreeNode,options) {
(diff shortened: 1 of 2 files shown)


Jermolene:
Great, thanks @saqimtiaz

kookma:
> This PR extends the `<$dropzone>` widget to:

Does this PR allows dropping a copy of a bibtex entry from clipboard?

saqimtiaz:
@kookma this PR only allows you to restrict which content types a dropzone supports from what is already supported.

Have you tried creating a dropzone that specifically use the bibtex deserializer and then pasting bibtex entries?

kookma:
> @kookma this PR only allows you to restrict which content types a dropzone supports from what is already supported.
> 
> Have you tried creating a dropzone that specifically use the bibtex deserializer and then pasting bibtex entries?

I am trying to do that! I can create a specified drip zone by I do not know how to trigger bibtex importer then!

saqimtiaz:
@kookma it is possible that the dropzone logic only applies the specified deserializer for file drops and not for paste content. I can investigate later today.

Can you please provide an example bibtext file to import, as well as a bibtext entry that should work when pasted?

kookma:
Thank you Saq!

This is a sample bibtex entry and file attached!

```
@book{noauthor_numerical_2018,
	address = {New York, NY},
	title = {Numerical python},
	isbn = {9781484242452},
	publisher = {Springer Science+Business Media},
	year = {2018},
}

```
[mypapers.zip](https://github.com/Jermolene/TiddlyWiki5/files/6289770/mypapers.zip)

saqimtiaz:
@kookma I've created an issue for this #5600

PR #5598, opened 2021-04-10
Correct DropzoneWidget docs

saqimtiaz:
@Jermolene It was somehow easier to create a new PR than target the previous one at the tiddlywiki-com branch.

This is a correction in existing docs and can go live immediately.


Jermolene:
> @Jermolene It was somehow easier to create a new PR than target the previous one at the tiddlywiki-com branch.

I've had problems with that too, I guess it's one of those cases where the illusion of Git breaks down.

> This is a correction in existing docs and can go live immediately.

Thanks.

PR #5601, opened 2021-04-11
Extend dropzone to use the specified deserializer for string data

### LINKED ISSUE #5600: [IDEA] Dropzone should use deserializer for non-file data
Currently if the `<$dropzone>` widget has a deserializer specified, it attempts to use it for File objects in both the drop and paste event handlers. However, for non-file data (strings) there is no attempt to use the specified deserializer. This would be helpful for use cases like pasting or dragging and dropping the text of a bibtex entry into a dropzone configured with the bibtex deserializer.

It is relatively straight forward in the paste event handler to attempt to use the specified deserializer first, before falling back to importing as a plain text string. (Example code at the end of this issue.)

**Edit:** the handling of File objects always uses the deserializer specified and does not attempt to fall back to an alternative method to deserialize or parse the data. The handling for non-file data in the paste and drop events should similarly always attempt to use the deserializer specified, and fall back to default importTypes when no deserializer has been specified.

I don't have the opportunity to look into this properly at the moment, so am creating this issue either for myself to visit at a later date or for someone else to follow up on.

Suggestion for how to handle this in the paste event at first glance:
```
if(self.dropzoneDeserializer) {
	var tiddlerFields;
	tiddlerFields = self.wiki.deserializeTiddlers(null,str,{title: self.wiki.generateNewTitle("Untitled")},{deserializer:self.dropzoneDeserializer});
	if(tiddlerFields && tiddlerFields.length) {
		readFileCallback(tiddlerFields);							
	}
}
if(!self.dropzoneDeserializer || !tiddlerFields || !tiddlerFields.length) {
	var tiddlerFields = {
		title: self.wiki.generateNewTitle("Untitled"),
		text: str,
		type: type
	};
	if($tw.log.IMPORT) {
		console.log("Importing string '" + str + "', type: '" + type + "'");
	}
	readFileCallback([tiddlerFields]);
}
```

saqimtiaz:
Currently the `<$dropzone>` widget only uses the `deserializer` specified via the optional parameter for `File` objects being imported, but not for `string` data that comes in via drag and drop or paste. 

This PR makes the handling of strings imported via drag and drop or paste consistent with Files, i.e. if a `deserializer` is specified it should be used to parse the data.  The current behaviour of the dropzone in this regard seems like an oversight and this discrepancy is not explained in the documentation.

The behaviour is unchanged if no `deserializer` has been specified.

This enhacement allows for workflows like having a custom dropzone to drag and drop or paste text in `bibtex` format.

Fixes #5600

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -168,7 +168,23 @@ DropZoneWidget.prototype.handleDropEvent  = function(event) {
 	}
 	// Try to import the various data types we understand
 	if(numFiles === 0) {
-		$tw.utils.importDataTransfer(dataTransfer,this.wiki.generateNewTitle("Untitled"),readFileCallback);
+		var fallbackTitle = self.wiki.generateNewTitle("Untitled");
+		//Use the deserializer specified if any
+		if(this.dropzoneDeserializer) {
+			for(var t= 0; t<dataTransfer.items.length; t++) {
+				var item = dataTransfer.items[t];
+				if(item.kind === "string") {
+					item.getAsString(function(str){
+						var tiddlerFields = self.wiki.deserializeTiddlers(null,str,{title: fallbackTitle},{deserializer:self.dropzoneDeserializer});
+						if(tiddlerFields && tiddlerFields.length) {
+							readFileCallback(tiddlerFields);
+						}
+					})
+				}
+			}
+		} else {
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@Jermolene let me know if this requires any changes or if you have a different implementation in mind.

Jermolene:
Thanks @saqimtiaz looks good

PR #1330, opened 2015-01-06
implements list iterator variables

tobibeer:
provides `<<iterator>>`, `<<iterator-even>>` and `<<iterator-last>>` see #1328

Issue #3384, opened 2018-08-09
Provide an index value in lists


AnthonyMuscio:
When using the proposed [range and indices](https://github.com/Jermolene/TiddlyWiki5/pull/3346#issuecomment-411620178) we will now have a quick and easy way to access tiddlers or fields that have numbers in them. 

Have we a method available to users (I do not think so) to generate these? like an increment?

This would allow the the creation of the type of content the range and indices[5] filters would be good at finding.

By way of an example
Lets say I have 20+ tiddlers, and three sort fields on each. Perhaps I could use drag and drop to order them, then make a button with list to display them in this order and update a sort field with that number.

I could drag and drop reorder them, again and set a second order field, and again for a third.

Now I can use sort[sort-field1] to use the first order, or sort[sort-field2] and sort[sort-field3]  

**The Best solution?**

I would think implicitly or explicitly  providing a variable to each iteration of a list item. 

Note this is not simply a count as can be achieved with an ordered list, but an actual value that can be used. Inside the list. I am not sure if this could be valid in the set statement use of filter but perhaps the final value could be available eg `<<namecount>>` within the <$set></$set>

**Implicit**
Or not stated, just available in the widget

```
<$list filter="[tag[test]]">
<<index>>. <<currentTiddler>><br>
</$list>
```

Result

0. 1st item
1. 2nd item
2. 3rd item

```
<$list filter="[tag[test]]">
<<item>>. <<currentTiddler>><br>
</$list>
```

1. 1st item
2. 2nd item
3. 3rd item

**explicit**
We can name the variable to contain the index, item or count.

```
<$list filter="[tag[test]]" indexvar="var-name">
<<var-name>>. <<currentTiddler>><br>
</$list>

```
0. 1st item
1. 2nd item
2. 3rd item

```
<$list filter="[tag[test]]" itemvar="var-name">
<<var-name>>. <<currentTiddler>><br>
</$list>
```

1. 1st item
2. 2nd item
3. 3rd item

In a similar vein we could just return a count variable that is the total at the end.
so we need not have separate set of list functions, just to count something. Even better would be if the last item, count or index would be available outside the list. This may be a good reason for explicitly named variables.

EvanBalster:
Hello, Anthony —

[...]

Issue #1523, opened 2015-02-23
Clean-up List widget and add Story widget


Jermolene:
We recently attempted to extend the list widget with an additional "iterator" variable that exposes the numeric index of an item. See #1328.

The problem with that approach was that the new iterator variables defeat the list widget logic for selective updating. (This is the mechanism whereby when a new item is inserted in a list, only the new item is rendered, and the existing list items are preserved.)

Another related problem is also emerging with the list widget. It supports an optional "storyview" that is used to animate changes to the list. These storyviews require that there be a DOM element wrapping each element of the list. In practice, the list widget deliberately doesn't insert any DOM elements in the render tree - if it, say, inserted included a parent DIV and a DIV for each item, then those DIVs would interfere with using the list widget to generate UL/LI lists. So the various storyviews have to do error checking to ensure that there is a wrapper div around each item. That's worked OK, but now we're creating storyviews that need to apply styles to the wrapper of the list.

It is proposed to resolve these issues by:
- Introduce a new "story" widget that is based on the current list widget with the following differences:
  - Adds an intrinsic DIV wrapper and a DIV for each item (thus providing the anchor points for storyviews to style the list and its items)
- Modify the current "list" widget to:
  - Remove support for storyviews
  - Remove support for selective updating
  - Restore the iterator variable functionality

The backwards compatibility implications of this change are pretty specific: it would break any existing code that uses the list widget to render a story list. I don't think that that is a big problem because the list widget is seldom used in that way except by the core.

Jermolene:
Just to add that subsequently we could add drag and drop functionality to the new story widget. Adding drag and drop to the list widget as it stands would be rather messy because of the lack of intrinsic generated elements.

tobibeer:
Interesting issue, sounds a sensible approach.

> Just to add that subsequently we could add drag and drop functionality to the new story widget. Adding drag and drop to the list widget as it stands would be rather messy because of the lack of intrinsic generated elements.

I'm wondering if any future drag and drop does not imply (re)sorting the underyling story (list), meaning ordering and iterating capabilities.

pmario:
At the google groups, there is a discussion about dynamically generated TW CSS "code". Imo a js macro will be needed to create CSS code like this. There are 2 dynamic elements. One is an iterator and the second one is a calculation. The number of elements will be defined with tags. ... I don't talk about the calculation, but the iterator variable would be really nice here. 

```
img:nth-child(5){animation-delay:0s;}
[...]

Issue #1328, opened 2015-01-06
expose iterator to list widget


tobibeer:
Expose temporary iterator variables during list processing...

* `<<iterator>>`
    * set to the current index, e.g. `1` 
* `<<iterator-even>>`
    * set to `true` for an even index
* `<<iterator-last>>`
    * set to `true` for last index

Always expose the default iterator variables, but — especially for nested lists — allow to declare a custom iterator variable prefix via a new *iterator* attribute of the list widget, as below...

**Examples**

```
<$list filter="[tag[foo]]" iterator="my-foo">

<$reveal type=match default=1 text=<<my-foo>>>
only show for item at index 1, with 1 meaning "first" (no js lingo)
</$reveal>

<$reveal type=nomatch default=1 text=<<my-foo>>>
show for all except first index
</$reveal>

<$reveal type=match default=true text=<<my-foo-even>>>
only show at even indexes
</$reveal>

<$reveal type=nomatch default=true text=<<my-foo-even>>>
only show at uneven indexes
</$reveal>

<$reveal type=match default=true text=<<my-foo-last>>>
only show for last index
</$reveal>

</$list>
```

tobibeer:
If ok, I may go ahead and write some code...

tobibeer:
Demo at: http://tbdemo.tiddlyspot.com/#Iterator

tobibeer:
**done**, demo at: http://tbdemo.tiddlyspot.com/#Iterator

Jermolene:
Hi @tobibeer excellent stuff, well done. It's a great idea, and the implementation looks good. Please go ahead and make a pull request.

tobibeer:
done & closing...

PR #5611, opened 2021-04-17
Extend list widget with "index" attribute

Jermolene:
This PR extends the `<$list>` widget with an optional `index` parameter that specifies the name of a variable to receive the numeric index of each list item. For example:

```
<$list filter="[tag[About]sort[title]]" index="index">
<div>
<<index>>: ''<$text text=<<currentTiddler>>/>'' (is first: <<index-first>>, is last: <<index-last>>)
</div>
</$list>
```

Two additional variables are also set to indicate the first and last items in the list:

* `<index-variable-name>-first` is set to `yes` for the first entry in the list, `no` for the others
* `<index-variable-name>-last` is set to `yes` for the last entry in the list, `no` for the others

This PR fixes #1330 and #3384. It probably obviates #1523. It also finally replaces the reverted PR #1328 from 2015.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -230,6 +230,10 @@ ListWidget.prototype.handleListChanges = function(changedTiddlers) {
 					}
 					this.insertListItem(t,this.list[t]);
 					hasRefreshed = true;
+				} else {
+					// Refresh the item we're reusing
+					var refreshed = this.children[t].refresh(changedTiddlers);
+					hasRefreshed = hasRefreshed || refreshed;
 				}
 			}
 		} else {
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -61,6 +61,7 @@ ListWidget.prototype.execute = function() {
 	this.template = this.getAttribute("template");
 	this.editTemplate = this.getAttribute("editTemplate");
 	this.variableName = this.getAttribute("variable","currentTiddler");
+	this.indexName = this.getAttribute("index");
 	this.storyViewName = this.getAttribute("storyview");
 	this.historyTitle = this.getAttribute("history");
 	// Compose the list elements
@@ -72,7 +73,7 @@ ListWidget.prototype.execute = function() {
 		members = this.getEmptyMessage();
 	} else {
 		$tw.utils.each(this.list,function(title,index) {
-			members.push(self.makeItemTemplate(title));
+			members.push(self.makeItemTemplate(title,index));
 		});
 	}
 	// Construct the child widgets
@@ -105,7 +106,7 @@ ListWidget.prototype.getEmptyMessage = function() {
 /*
 Compose the template for a list item
 */
-ListWidget.prototype.makeItemTemplate = function(title) {
+ListWidget.prototype.makeItemTemplate = function(title,index) {
 	// Check if the tiddler is a draft
 	var tiddler = this.wiki.getTiddler(title),
 		isDraft = tiddler && tiddler.hasField("draft.of"),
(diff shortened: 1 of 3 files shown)


saqimtiaz:
@Jermolene this looks good and I favour this approach rather than introducing a new widget as has been previously discussed. 

We may want to draw more attention to the note about degrading performance, but we can fine tune the docs once this is merged.

Jermolene:
> @Jermolene this looks good and I favour this approach rather than introducing a new widget as has been previously discussed.

The result is a bit cleaner than I feared; the list widget is still a bit of a mess, but not too much worse!

core/modules/widgets/list.js
```diff
@@ -128,7 +129,14 @@ ListWidget.prototype.makeItemTemplate = function(title) {
 		}
 	}
 	// Return the list item
-	return {type: "listitem", itemTitle: title, variableName: this.variableName, children: templateTree};
+	var parseTreeNode = {type: "listitem", itemTitle: title, variableName: this.variableName, children: templateTree};
+	if(this.indexName) {
+		parseTreeNode.index = index.toString();
+		parseTreeNode.indexName = this.indexName;
+		parseTreeNode.isFirst = index === 0;
```
pmario, line 136: hmm, Should we use 0-based counting for user-faced numbers? eg [nth operator](https://tiddlywiki.com/#nth%20Operator) starts with 1  ?!

kookma:
> We may want to draw more attention to the note about degrading performance, but we can fine tune the docs once this is merged.

$list is an essential widget and I hope we have no degrading performance!

saqimtiaz:
@kookma the performance degradation only applies if you use the new index parameter (in which case there is no way around it). Otherwise the performance will be the same.

Jermolene:
An interesting experiment to assess the performance difference between the two variants of the list widget is to hack the code to always use the index attribute. I did it by splicing the following line in:

```js
	this.indexName = this.getAttribute("index");
	this.indexName = this.indexName || "unused-index"; // <<<<< ADD THIS LINE
	this.storyViewName = this.getAttribute("storyview");
```

Playing with the resulting wiki, as you'd expect the story river becomes very sluggish opening and closing tiddlers. The tab controls behave hilariously: every tab "pops" as you switch between them.

I uploaded a snapshot here for others to experiment with:

https://tw5-with-hacked-list-widget.tiddlyhost.com

core/modules/widgets/list.js
```diff
@@ -142,7 +150,7 @@ ListWidget.prototype.refresh = function(changedTiddlers) {
 		this.storyview.refreshStart(changedTiddlers,changedAttributes);
 	}
 	// Completely refresh if any of our attributes have changed
-	if(changedAttributes.filter || changedAttributes.template || changedAttributes.editTemplate || changedAttributes.emptyMessage || changedAttributes.storyview || changedAttributes.history) {
+	if(changedAttributes.filter || changedAttributes.variable || changedAttributes.index || changedAttributes.template || changedAttributes.editTemplate || changedAttributes.emptyMessage || changedAttributes.storyview || changedAttributes.history) {
```
pmario, line 153: IMO if there is a refresh needed, if any attribute changes, the mechanism that Saq uses is much more readable.

```
if (changedAttributes.length > 0) {
```

core/modules/widgets/list.js
Jermolene: The `reduce[]` operator uses zero-based indexes, and I think it's reasonable to continue. (The `nth` operator was part of one of those PRs that I've come to regret)

pmario: Then we should deprecate it and remove it in a year or so.

Jermolene:
Just to note that in 4a99e0cc7d4a6b9e7071c0b2a9a0f63c3c7d2492, I changed the name of the "index" attribute to "counter", and moved to 1-based counting.

PR #5612, opened 2021-04-17
Add "average" filter operator for arithmetic mean

Jermolene:
This PR introduces an "average" operator for calculating the arithmetic mean of a list of numbers. It is already possible to use the "reduce" operator (or prefix) to calculate averages, but a dedicated operator makes things easier.

Should we add other statistical operators like median, mode or standard deviation? Or is that a slippery slope to adding trigonometry functions?

### DIFF
--- a/core/modules/filters/math.js
+++ b/core/modules/filters/math.js
@@ -125,14 +125,6 @@ exports.minall = makeNumericReducingOperator(
 	Infinity // Initial value
 );
 
-exports.average = makeNumericReducingOperator(
-	function(accumulator,value) {return accumulator + value},
-	0, // Initial value
-	function(finalValue,numberOfValues) {
-		return finalValue/numberOfValues;
-	}
-);
-
 exports.median = makeNumericArrayOperator(
 	function(values) {
 		var len = values.length, median;
@@ -148,6 +140,39 @@ exports.median = makeNumericArrayOperator(
 	}
 );
 
+exports.average = makeNumericReducingOperator(
+	function(accumulator,value) {return accumulator + value},
+	0, // Initial value
+	function(finalValue,numberOfValues) {
+		return finalValue/numberOfValues;
+	}
+);
+
+exports.variance = makeNumericReducingOperator(
+	function(accumulator,value) {return accumulator + value},
+	0,
+	function(finalValue,numberOfValues,originalValues) {
+		return getVarianceFromArray(originalValues,finalValue/numberOfValues);
+	}
+);
+
+exports["standard-deviation"] = makeNumericReducingOperator(
+	function(accumulator,value) {return accumulator + value},
+	0,
+	function(finalValue,numberOfValues,originalValues) {
+		var variance = getVarianceFromArray(originalValues,finalValue/numberOfValues);
+		return Math.sqrt(variance);
+	}
+);
+
+//Calculate the variance of a population of numbers in an array given its mean
+function getVarianceFromArray(values,mean) {
+	var deviationTotal = values.reduce(function(accumulator,value) {
+		return accumulator + Math.pow(value - mean, 2);
+	},0);
+	return deviationTotal/values.length;
+};
+
 function makeNumericBinaryOperator(fnCalc) {
 	return function(source,operator,options) {
 		var result = [],
@@ -157,24 +182,24 @@ function makeNumericBinaryOperator(fnCalc) {
 		});
 		return result;
 	};
-}
+};
 
 function makeNumericReducingOperator(fnCalc,initialValue,fnFinal) {
 	initialValue = initialValue || 0;
 	return function(source,operator,options) {
 		var result = [];
 		source(function(tiddler,title) {
-			result.push(title);
+			result.push($tw.utils.parseNumber(title));
 		});
 		var value = result.reduce(function(accumulator,currentValue) {
-				return fnCalc(accumulator,$tw.utils.parseNumber(currentValue));
+				return fnCalc(accumulator,currentValue);
 			},initialValue);
 		if(fnFinal) {
-			value = fnFinal(value,result.length);
+			value = fnFinal(value,result.length,result);
 		}
 		return [$tw.utils.stringifyNumber(value)];
 	};
-}
+};
 
 function makeNumericArrayOperator(fnCalc) {
 	return function(source,operator,options) {
@@ -188,6 +213,6 @@ function makeNumericArrayOperator(fnCalc) {
 		});
 		return results;
 	};
-}
+};
 
 })();
--- a/core/modules/filters/math.js
+++ b/core/modules/filters/math.js
@@ -125,6 +125,14 @@ exports.minall = makeNumericReducingOperator(
 	Infinity // Initial value
 );
 
+exports.average = makeNumericReducingOperator(
+	function(accumulator,value) {return accumulator + value},
+	0, // Initial value
+	function(finalValue,numberOfValues) {
+		return finalValue/numberOfValues;
+	}
+);
+
 function makeNumericBinaryOperator(fnCalc) {
 	return function(source,operator,options) {
 		var result = [],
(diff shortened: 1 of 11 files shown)


saqimtiaz:
@Jermolene I considered operators for mean, median and mode when I added some math operators for the last release. In the end I had planned to do so but did not get around to it.

I think the key is deciding what belongs in the core, and what potentially belongs in a more specialized "Mathematics Filters plugin". Median and mode in particular are commonplace enough that I think a place in the core is warranted.

I believe @kookma has worked with calculating some basic statistics in TW and might have some useful insight as to what would be useful in the core versus in a specialized plugin.

kookma:
Thank you @saqimtiaz - I think bringing some basic statistical functions is good and useful! Hopefully TW has some of them like min/max minall/maxall.

I have implemented few macros in Shiraz. See [maths](https://kookma.github.io/TW-Shiraz/#%24%3A%2Fplugins%2Fkookma%2Fshiraz%2Fmacros%2Fdtables%2Fmaths) and https://kookma.github.io/TW-Shiraz/#demo%2Fdynamic-tables-numerical-summary

I may recommend: mean (average), median, sdv (standrd deviation), var (variance)

Jermolene:
I've added a median operator.

I'd welcome some help on the standard deviation and variance functions. The sample code I'm finding with casual searches usually employs the `reduce` function, which isn't available in ECMAScript 5 that we use for the core. I'm not super confident that I remember enough maths to be a safe pair of hands to wrangle the algorithm into the shape we need...

saqimtiaz:
@Jermolene I can take a crack at this after Thursday.

Jermolene:
> @Jermolene I can take a crack at this after Thursday.

Thank you! Much appreciated

Jermolene:
> What do you intend `average` to do if handed an empty list? Do you intend it to throw an exception, since there is no such thing as the average of 0 items? Or do you intend it to return 0, or allow the user to define some value that it will return for a list of 0 items? Because right now, the code will throw an exception: on an empty list, the `fnFinal` function is always called (with the final value being the initial value), and `fnFinal` for `average` will then divide by 0. If you want `average` to return the initial value unchanged on an empty list, then `average`'s implementation needs to change to `return numberOfValues == 0 ? finalValue : finalValue/numberOfValues`

Thanks @rmunn I'd overlooked that issue. As it turns out, JavaScripts weird semantics help: the current behaviour is not to crash, but to return the result "NaN" as a string. On reflection that might be a reasonable response. If so, I'll add a test and clarify things in the docs.

kookma:
NaN or n/a are good!

saqimtiaz:
> The sample code I'm finding with casual searches usually employs the `reduce` function, which isn't available in ECMAScript 5 that we use for the core.

@Jermolene it looks like `Array.prototype.reduce` was introduced in ECMAScript 5 and we do already use it in the core in a few places. So before I work on this, is there a nuance to this that I am missing here?

Jermolene:
> > The sample code I'm finding with casual searches usually employs the `reduce` function, which isn't available in ECMAScript 5 that we use for the core.
> 
> @Jermolene it looks like `Array.prototype.reduce` was introduced in ECMAScript 5 and we do already use it in the core in a few places. So before I work on this, is there a nuance to this that I am missing here?

Sorry, I always get the names of JS versions wrong.

The core avoids map/reduce/forEach because they weren't widely supported when we started in 2011/2. Any usages that have crept in are accidental on my part. I dimly remember that I did once merge map or reduce into boot.js, and quickly got feedback that it broke some old browser and so we removed it again.

Clicking some buttons on caniuse gives quite a helpful timeline of when reduce was introduced, but doesn't help me remember which browsers we were concerned about:

image

So, it might be time to rethink our JS subset. But of course now that I'm using modern JS on other projects I want `for(x of y)` and other even newer good stuff.

pmario:
>So, it might be time to rethink our JS subset. But of course now that I'm using modern JS on other projects I want for(x of y) and other even newer good stuff.

... modern JS is addictive ;)) as long as you don't need to use "babel and other js bloat-ware"

saqimtiaz:
@Jermolene just to be clear, we do already use `reduce()` even in the existing mathematics filters including in the [makeNumericReducingOperator](https://github.com/Jermolene/TiddlyWiki5/pull/5612/files#diff-e5a2113e5745d9ede34d0e3668edbe02f801dacb1bd1f54c361b5254ec923f46R169) method which this PR extends. 

Do we want to remove that as well? It also used in `filters/strings.js` and `filters/x-listops.js` and `utils.hashString`. The fact that we have had no complaints or error reports in this regard suggest to me that this isn't something we need to worry about.

> So, it might be time to rethink our JS subset. But of course now that I'm using modern JS on other projects I want `for(x of y)` and other even newer good stuff.

As someone that works extensively with people in remote and/or less privileged communities, older browsers really isn't an issue even in my work, whereas it used to be 10+ years ago. Bandwidth on the other hand continues to be a real limitation. So I would be in favour of moving to a more modern JS spec.

pmario:
>Do we want to remove that as well? It also used in filters/strings.js and filters/x-listops.js and utils.hashString. The fact that we have had no complaints or error reports in this regard suggest to me that this isn't something we need to worry about.

+1

Jermolene:
> @Jermolene just to be clear, we do already use `reduce()` even in the existing mathematics filters including in the [makeNumericReducingOperator](https://github.com/Jermolene/TiddlyWiki5/pull/5612/files#diff-e5a2113e5745d9ede34d0e3668edbe02f801dacb1bd1f54c361b5254ec923f46R169) method which this PR extends.

Of course, I'd forgotten. In fact I think we've already gone round this loop and decided that we might as well start using map and reduce. Apologies, my [original comment](https://github.com/Jermolene/TiddlyWiki5/pull/5612#issuecomment-823146646) was wrong.

> As someone that works extensively with people in remote and/or less privileged communities, older browsers really isn't an issue even in my work, whereas it used to be 10+ years ago. Bandwidth on the other hand continues to be a real limitation. So I would be in favour of moving to a more modern JS spec.

That is encouraging. The fun part is choosing our baseline. I'm interested in ES2017 to get async functions, which translates to browsers released in 2017 and later. In a year or two that might be reasonable, lining up with a potential v5.2.x.

Jermolene:
I'm also thinking that improving our static site publishing capabilities might reduce the pressure to support older browsers.

saqimtiaz:
@kookma for variance and standard deviation, does it make sense to calculate for [population data, rather than sample data](http://www.differencebetween.net/science/mathematics-statistics/difference-between-sample-variance-population-variance/#:~:text=Summary%3A%20Population%20variance%20refers%20to,variance%20calculated%20from%20sample%20data.&text=As%20a%20result%20both%20variance,found%20out%20from%20population%20data.)?

Or do we need to allow for both?

saqimtiaz:
#5645 adds variance and standard-deviation operators.

kookma:
> @kookma for variance and standard deviation, does it make sense to calculate for [population data, rather than sample data](http://www.differencebetween.net/science/mathematics-statistics/difference-between-sample-variance-population-variance/#:~:text=Summary%3A%20Population%20variance%20refers%20to,variance%20calculated%20from%20sample%20data.&text=As%20a%20result%20both%20variance,found%20out%20from%20population%20data.)?
> 
> Or do we need to allow for both?

For a complete solution I prefer both! From mathematical point of view the calculation is similar! The formula can be obtained for [KhanAcademi](https://www.khanacademy.org/math/statistics-probability/summarizing-quantitative-data/variance-standard-deviation-sample/a/population-and-sample-standard-deviation-review) or the one you have given above!

But as you said, population SD and Var may be the most use cases here in TW.

saqimtiaz:
@Jermolene I think this can be merged to master now. The examples I added include return `NaN` for empty input.

Jermolene:
Thanks @saqimtiaz @kookma

PR #5614, opened 2021-04-18
Fix an "update problem" and Improve new journal workflow 

pmario:
This PR fixes a problem and adds new functionality in a backwards compatible way. 

## Working wiki

https://pmario.github.io/kitchensink/5614-improved-journal-workflow.html

## Problem fixed: 

 - go to tiddlywiki.com
 - open the [Control Panel](https://tiddlywiki.com/#%24%3A%2FControlPanel)
   - Use this `DDth MMM YYYY 0hh:0mm:0ss` as the new journal **title**
 - Activate "New Journal Here" in the tiddler toolbar

grafik

 - click "New Journal Here" - and Save!
 - click "New Journal Here" .. -> Problem: The same journal tiddler will be opened. 

This PR fixes this problem. 

## New Functions

 - New Journal button (4) in the right sidebar has a new functionality, which can be activated by new option (3) Navigate 
   - This means, if (3) is checked and (4) is clicked, it will only open the journal tiddler in view mode
   - If the journal tiddler doesn't exist is will be created.
   - If the option tiddler (3) doesn't exist the default V5.1.23 behaviour happens
 - There are 2 new settings see: (2)
   - "new journal here" title and 
   - "new journal tag"

new-journal-01

---------

So if "New Journal Here" (2) is clicked this happens

new-journal-02

--------------

This also works with a "New Tiddler" if you use `<<currentTiddler>> - 0hh:0mm:0ss` as a "new journal here" title

new-journal-03


Jermolene:
Hi @pmario this doesn't appear to be backwards compatible; clicking the "new journal here" button on a tiddler doesn't seem to have the same effect as on an empty TiddlyWiki because the "new journal here" title template doesn't match the present default.

Jermolene:
(I should add that I've only had a quick look at this so far)

pmario:
>Hi @pmario this doesn't appear to be backwards compatible; clicking the "new journal here" button on a tiddler doesn't seem to have the same effect as on an empty TiddlyWiki because the "new journal here" title template doesn't match the present default.

That's right, New Journal Here is completely useless with the current default settings. 

Try this: 

 - go to tiddlywiki.com
 - Create a New Journal with the right sidebar
 - add some text in the text area
 - save
 - Create a New Journal Here with the tiddler toolbar
 - save
 - select any tiddler from the TOC
 - Create a New Journal Here with the tiddler toolbar
 - -> IMO this is a useless function

It only tags the Journal tiddler with new tiddlers that have no connection.

pmario:
>Hi @pmario this doesn't appear to be backwards compatible; clicking the "new journal here" button on a tiddler doesn't seem to have the same effect as on an empty TiddlyWiki because the "new journal here" title template doesn't match the present default.

If you set the config as follows, you get the same behaviour. ... But IMO that's useless

grafik

Jermolene:
> That's right, New Journal Here is completely useless with the current default settings. 
>
> It only tags the Journal tiddler with new tiddlers that have no connection.

That is the intended functionality of the "new journal here" button. An example of the workflow that it supports is to be able to link the tiddlers that one edits/reviews on a particular day to that days journal.

It is slightly discouraging that the first point mentioned in the description of the PR isn't entirely correct. PRs work best if they are really easy for me to merge; any inaccuracy makes me super cautious, and feel that I need to triple check everything.

pmario:
@Jermolene I did update the wiki. If you download empty.html from https://pmario.github.io/kitchensink/5614-improved-journal-workflow.html#GettingStarted you'll get the exact same behaviour as 5.1.23 with NewJournalHere now. 

I did completely remove the NewJournalHere conifg tiddlers. ... If they are missing the NewJournal configs are used
see commit: https://github.com/Jermolene/TiddlyWiki5/pull/5614/commits/671f55de30618649fc31f006bd8e613a0fdc9d06

Jermolene:
Thanks @pmario.

A couple of points from a preliminary look at the code:

* In several files there are whitespace changes which make it very hard to see what is actually changed (particularly the stylesheet). Please could you revert the pure whitespace changes
* I do not understand the purpose of the checkbox to open the journal tiddler in view mode

pmario:
> I do not understand the purpose of the checkbox to open the journal tiddler in view mode

I personally would like to use the NewJournal tiddler as a "Daly Dashboard", that is prefilled with some code from $:/config/NewJournal/Text  .... 

If the daily journal exists it is very frustrating, that the right sidebar opens the Journal in edit mode. The checkbox will allow a user to open the Journal "view mode".

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -2285,22 +2285,29 @@ html body.tc-body.tc-single-tiddler-window {
...
-.tc-control-panel table, .tc-control-panel table input, .tc-control-panel table textarea {
+.tc-control-panel table, 
+.tc-control-panel table input,
+.tc-control-panel table textarea {
 	width: 100%;
 }
 
+/* input type checkbox doesn't make sense with 100% width */
+.tc-control-panel table input[type=checkbox] {
+	width: inherit;
```
pmario, line 2296: This is the only addition I did make to the base.tid. 

... It's needed because of the rule above, where I did add 2 linebreaks to make it readable for humans. 

The rest is replacing spaces with tabs.

Jermolene:
> > I do not understand the purpose of the checkbox to open the journal tiddler in view mode
> 
> I personally would like to use the NewJournal tiddler as a "Daly Dashboard", that is prefilled with some code from $:/config/NewJournal/Text ....
> 
> If the daily journal exists it is very frustrating, that the right sidebar opens the Journal in edit mode. The checkbox will allow a user to open the Journal "view mode".

I sympathise, but my worry is that this PR still reflects your personal requirements where we should be aiming to make core functionality be universal. In the early days we were forced to use a lot of "sealed box magic" that didn't give much flexibility, but now we have the opportunity to make things very flexible without resorting to pages of checkboxes.

Perhaps a helpful way forwards would be to gather some more input, perhaps kicking off a thread in the main group to discuss how people use the journaling feature, and try to synthesise a proposal that helps more people.

Just as an example, a less disruptive approach to introducing the checkbox for opening the journal in view mode, one could just construct a link/button that linked to the current journal in view mode, rather than trying to wrap everything together in the same button. It's often easier to satisfy goals of different groups by slotting in new things rather than accreting more and more options on the things we already have.

themes/tiddlywiki/vanilla/base.tid
Jermolene: > The rest is replacing spaces with tabs.

We've had this issue before and I thought I'd made it clear why I don't like PRs that mix functional changes with whitespace changes.

AnthonyMuscio:
In the GG thread I mention Ands provide my enhancement to set a journal-date field, with new and new here journal, so its easy to sort on journal date or date time, regardless of the journal tiddlers title format used. I think this should be done along with the suggestions above (I have not separately reviewed). This also allows the created date to remain true if you wanted to pre-date or post date a journal.

pmario:
Close this one in favour of future PRs

PR #5615, opened 2021-04-19
freelinks - performance improvement

pmario:
This PR removes all CamelCase checks from the freelink logic. 

 - CamelCase has its own parser. So freelinks shouldn't deal with it. 
 - Especially since the CameCase parser is already done when freelinks starts
 - IMO freelinks can be "freed" from that work ;)

This allows us to reduce the `reparts` length from 1339 to about 950 for tiddlywiki.com which imo is huge. Since the regexp is used about 1400 times if 30 tiddlers are open.

I did the following test. 

 - Install freelinks plugin
 - Disable it 
 - Open 30 tiddlers from the Recent tab
 - Enabled freelinks
 - Enabled Pereformance Instrumentation
 - Save - Reload
 - Inspect
 - Switch tabs between Recent and Contents

I was able to reduce the time needed to check all open tiddlers from about 1.9 seconds to about 1.2 seconds. Using FF latest Windows 10 Pro.

### DIFF
--- a/plugins/tiddlywiki/freelinks/text.js
+++ b/plugins/tiddlywiki/freelinks/text.js
@@ -77,9 +77,10 @@ TextNodeWidget.prototype.execute = function() {
 					}
 				}),
 				titles = [],
-				reparts = [];
+				reparts = [],
+				isCC = new RegExp("^" + $tw.config.textPrimitives.unWikiLink + "?" + $tw.config.textPrimitives.wikiLink,"mg");
 			$tw.utils.each(sortedTitles,function(title) {
-				if(title.substring(0,3) !== "$:/") {
+				if ((title.substring(0,3) !== "$:/") && (title.match(isCC) === null)) {
 					titles.push(title);
 					reparts.push("(\\b" + $tw.utils.escapeRegExp(title) + "\\b)");
 				}
@@ -120,10 +121,10 @@ TextNodeWidget.prototype.execute = function() {
 						childParseTree[index] = {
 							type: "plain-text",
 							text: text.substring(matchEnd)
-						};					
+						};
 					}
 				}
-			} while(match && childParseTree[childParseTree.length - 1].type === "plain-text");			
+			} while(match && childParseTree[childParseTree.length - 1].type === "plain-text");
 		}
 	}
 	// Make the child widgets
@@ -158,7 +159,7 @@ TextNodeWidget.prototype.refresh = function(changedTiddlers) {
 		this.refreshSelf();
 		return true;
 	} else {
-		return false;	
+		return false;
 	}
 };
 


pmario:
That's the link I did test with: https://pmario.github.io/kitchensink/5615-improve-freelinks.html#%24%3A%2FControlPanel:[[TiddlyMap%20Plugin%20by%20Felix%20K%C3%BCppers]]%20[[TiddlyServer%20by%20Arlen%20Beiler]]%20[[TiddlyWiki%20extensions%20for%20Sublime%20Text%203%20by%20roma0104]]%20[[TiddlyWiki%20extensions%20for%20ViM]]%20[[TiddlyWiki%20Video%20Tutorials%20by%20Francis%20Meetze]]%20[[Timimi%3A%20WebExtension%20and%20Native%20Host%20by%20Riz]]%20[[Tinka%20by%20Andreas%20Hahn]]%20[[TW-Scripts%20by%20Mohammad]]%20[[TW5-SingleExecutable%20by%20Jed%20Carty]]%20TWCommunitySearch%20[[Twexe%3A%20Single%20File%20Tiddlywiki5%20executable]]%20[[twproxy%20by%20Steve%20Gattuso]]%20[[vis.js%20Timeline%20by%20emkay%2C%20revived%20by%20kixam]]%20[[Widdly%20by%20Opennota]]%20[[Projectify%20by%20Nicolas%20Petton]]%20[[%22Notebook%20theme%22%20by%20Nicolas%20Petton]]%20[[WebServer%20Parameter%3A%20sse-enabled]]%20[[TW5-firebase%3A%20TiddlyWiki5%20for%20Google%20Firebase%20by%20Peter%20Neumark]]%20[[%22TW%20Icons%22%20by%20morosanuae]]%20[[Tiddler%20Structure]]%20[[RangeWidget%20Example]]%20[[RadioWidget%20Example]]%20RadioWidget%20MessageCatcherWidget%20[[Community%20Links%20Aggregator]]%20Community%20ViewWidget%20Contributing%20DropzoneWidget%20[[Hidden%20Setting%3A%20HTML%20Parser%20Sandbox]]%20%24%3A%2FControlPanel

Jermolene:
Hi @pmario the observation that we don't need to scan check for camelcase titles is very valuable, and does indeed seem to offer a striking optimisation.

However, this implementation is dependent on the camel case parser rule being enabled and active. That's actually quite tricky to check because the `\rules` pragma allows it to be done on a local basis, and so given a fragment of the render tree there's no way to determine whether the camelcase parser rule was active.

An alternative approach might be to leave things in the hands of the user by introducing a new `[is[camelcase]]` operator that can be used in the `$:/config/Freelinks/TargetFilter` filter to exclude camelcase titles. The default at the moment is that `$:/config/Freelinks/TargetFilter` is empty (which has the same effect as `[!is[system]]`) but it could be something like `[!is[system]!is[camelcase]]`.

pmario:
>However, this implementation is dependent on the camel case parser rule being enabled and active. 

I don't understand, what you mean be "this implementation". ... My changes or the original code?

IMO it doesn't matter if the the user want's to render CamelCase or not. freelinks doesn't need to check for it. ... If the user disabled the wikilink.js parser globally, it indicates, that freelinks also doesn't need to render it. 

>That's actually quite tricky to check because the \rules pragma allows it to be done on a local basis, and so given a fragment of the render tree there's no way to determine whether the camelcase parser rule was active.

If wikilink.js is enabled, it already did parse them. So there will be _no_ render-tree text elements, that contain wikilinks. They have been converted already. So IMO freelinks doesn't need to know about CamelCase settings at all. 

We can just remove them

Jermolene:
> I don't understand, what you mean by "this implementation". ... My changes or the original code?

I meant the changes in this PR.

> IMO it doesn't matter if the the user wants to render CamelCase or not. freelinks doesn't need to check for it. ... If the user disabled the wikilink.js parser globally, it indicates, that freelinks also doesn't need to render it.
> 
> > That's actually quite tricky to check because the \rules pragma allows it to be done on a local basis, and so given a fragment of the render tree there's no way to determine whether the camelcase parser rule was active.
> 
> If wikilink.js is enabled, it already did parse them. So there will be _no_ render-tree text elements, that contain wikilinks. They have been converted already. So IMO freelinks doesn't need to know about CamelCase settings at all.
> 
> We can just remove them

Apologies, I didn't explain it clearly. The problem arises if a user has tiddlers with titles that happen to be camelcase but they have camelcase parsing switched off globally. In that situation, I think this PR would not link to the camelcase tiddlers. It's a scenario that might plausibly happen for users migrating from using camelcase to freelinks.

pmario:
>Apologies, I didn't explain it clearly. The problem arises if a user has tiddlers with titles that happen to be camelcase but they have camelcase parsing switched off globally.

OK. So the user switched off CamelCase linking. .. It's the users decision. 

>In that situation, I think this PR would not link to the camelcase tiddlers. It's a scenario that might plausibly happen for users migrating from using camelcase to freelinks.

Why should freelink link to CamelCases? ... The user switched it **off**. .. From my point of view, it should _not_ link.

Jermolene:
> > Apologies, I didn't explain it clearly. The problem arises if a user has tiddlers with titles that happen to be camelcase but they have camelcase parsing switched off globally.
> 
> OK. So the user switched off CamelCase linking. .. It's the users decision.
> 
> > In that situation, I think this PR would not link to the camelcase tiddlers. It's a scenario that might plausibly happen for users migrating from using camelcase to freelinks.
> 
> Why should freelink link to CamelCases? ... The user switched it **off**. .. From my point of view, it should _not_ link.

OK take a different scenario. A new user decides (a) they don't like camelcase so they turn it off globally and (b) to use Freelinks. Then later they create a tiddler called "McDonald". With this PR, that tiddler would mysteriously not be handled by the freelink plugin. I think that would be confusing to users.

pmario:
>OK take a different scenario. A new user decides (a) they don't like camelcase so they turn it off globally and (b) to use Freelinks. Then later they create a tiddler called "McDonald". With this PR, that tiddler would mysteriously not be handled by the freelink plugin. I think that would be confusing to users.

OK. That's valid ..  

>An alternative approach might be to leave things in the hands of the user by introducing a new [is[camelcase]] operator that can be used in the $:/config/Freelinks/TargetFilter filter to exclude camelcase titles. The default at the moment is that $:/config/Freelinks/TargetFilter is empty (which has the same effect as [!is[system]]) but it could be something like [!is[system]!is[camelcase]].

In the plugin `[!is[system]]` is not needed, since it is hardcoded in line 83

So we need an `[!is[camelcase]]` operator. ... Should it come with the freelinks plugin, or should it be part of the core?

Would an `[all[camelcase]]` make sense?

pmario:
Close in favour of: "[IDEA] a new camelcase operand may make the freelinks plugin faster" #5630

PR #5622, opened 2021-04-23
Add TiddlyHost saving info and revise TiddlySpot info

simonbaird:
I updated the English text only. Will need some help with the
translations.

Summary:
- Mention that TiddlySpot is deprecated and doesn't allow site
  creation any more. Suggest using TiddlyHost instead.
- Remove obsolete intructions about creating TiddlySpot sites.
- Misc editing/rewording/tweaking of existing TiddlySpot info
  for tidiness and clarity.
- Add new information about saving on TiddlyHost.
- Add logos because why not..

Note: I usually prefer the non-camel case versions of Tiddlyspot and
Tiddlyhost, but decided to go with the CamelCase WikiWords here to
fit in with the existing conventions.

### DIFF
--- a/editions/tw5.com/tiddlers/saving/Saving on TiddlyHost.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving on TiddlyHost.tid
@@ -1,0 +1,17 @@
+caption: ~TiddlyHost
+color: #29B6F6
+community-author: Simon Baird
+created: 20210422191232572
+delivery: Service
+description: Online service for creating and hosting TiddlyWikis
+method: save
+modified: 20210423003921468
+tags: Android Chrome Firefox [[Internet Explorer]] Linux Mac Opera PHP Safari Saving Windows iOS Edge
+title: Saving on TiddlyHost
+type: text/vnd.tiddlywiki
+
+[img width=140 [https://github.com/simonbaird/tiddlyhost/raw/main/rails/app/assets/images/logo-800.png]]
+
+[[TiddlyHost.com|https://tiddlyhost.com/]] is a hosting service for TiddlyWiki created by Simon Baird. Once you sign up and confirm your email you can create "sites", (i.e. ~TiddlyWikis), with support for online saving. Sites can be private or public, and you can optionally list them on the taggable and searchable [[TiddlyHost Hub|https://tiddlyhost.com/hub]] where they'll be discoverable by others.
+
+Unlike [[TiddlySpot|Saving on TiddlySpot]], [[TiddlyHost|https://tiddlyhost.com]] is secure, open source, and has proper support for TiddlyWiki5. It also allows uploading existing ~TiddlyWiki files, supports TiddlyWikiClassic, and lets you claim ownership of your ~TiddlySpot sites. For more information see the [[FAQ|https://github.com/simonbaird/tiddlyhost/wiki/FAQ]] and the [[About|https://tiddlyhost.com/about]] page.
--- a/editions/tw5.com/tiddlers/saving/Saving on TiddlySpot.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving on TiddlySpot.tid
@@ -1,56 +1,46 @@
 caption: ~TiddlySpot
 color: #29B6F6
-community-author: Simon Baird and Daniel Baird
+community-author: Simon Baird & Daniel Baird
 created: 20130825213500000
 delivery: Service
-description: Free online service for hosting TiddlyWiki files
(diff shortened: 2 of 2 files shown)


simonbaird:
Demo [here|https://mr-5622-tiddlyhost-saving-info.tiddlyhost.com/#Saving%20on%20TiddlySpot:GettingStarted%20%5B%5BSaving%20on%20TiddlySpot%5D%5D%20%5B%5BSaving%20on%20TiddlyHost%5D%5D].

Jermolene:
Thanks @simonbaird some minor comments coming up.

Jermolene:
> Thanks @simonbaird some minor comments coming up.

Nope, on more careful review, all looks good, except that since this is docs only it should target the `tiddlywiki-com` branch which will allow it to be deployed to tiddlywiki.com instantly.

saqimtiaz:
Do we want to update the TiddlySpot saver tab in the Control Panel as well for the next TW release?

Jermolene:
> Do we want to update the TiddlySpot saver tab in the Control Panel as well for the next TW release?

Yes, we definitely should. It's a shame that Tiddlyhost can't use the upload saver.

saqimtiaz:
I'm not entirely sure but I think TiddlyHost might still use the upload saver just without a password, so we would just need to tweak it accordingly. @simonbaird can confirm.

Jermolene:
> I'm not entirely sure but I think TiddlyHost might still use the upload saver just without a password, so we would just need to tweak the UI accordingly. @simonbaird can confirm.

Yes indeed, and it works great, but it needed a core modification, and so doesn't have good backwards compatibility.

simonbaird:
Hold on, I'll rebase...

simonbaird:
It's saying I want to merge 99 commits...

simonbaird:
That's better. Now rebased on tiddlywiki-com branch.

Jermolene:
> It's saying I want to merge 99 commits...

> That's better. Now rebased on tiddlywiki-com branch.

Great stuff, thanks

simonbaird:
Ps, I'm not following the upload saver discussion. Tiddlyhost does use the standard upload saver. There was a little tweak for it, see https://github.com/Jermolene/TiddlyWiki5/pull/5455 .

So I have an interest in seeing that tweak appear in a non-prerelease version on TW. At that point Tiddlyhost will stop using a pre-release build by default, and start using the official released build, presumably 5.1.24.

In the long term I think there'll be a new saver plugin specifically for Tiddlyhost, but it's not a big priority now, since upload.js is doing a perfectly good job currently.

simonbaird:
Oh, I think I get it, re "..update the TiddlySpot saver tab..". The wording could be revised a little to reflect current reality. I'll take a look at that soon.

Jermolene:
> Ps, I'm not following the upload saver discussion. Tiddlyhost does use the standard upload saver. There was a little tweak for it, see #5455 .

I was referring to the PUT saver:

https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/savers/put.js

The original idea was that it be a completely generic saver, but it has subsequently acquired some features that I think are specific to WebDAV servers.

> So I have an interest in seeing that tweak appear in a non-prerelease version on TW. At that point Tiddlyhost will stop using a pre-release build by default, and start using the official released build, presumably 5.1.24.

Yes indeed. Very soon we need to make a call whether to wait for the new publishing framework stuff before we release v5.1.24, which would imply a May or June release, but we could go earlier if we decided to defer the publishing stuff to v5.1.24.

> In the long term I think there'll be a new saver plugin specifically for Tiddlyhost, but it's not a big priority now, since upload.js is doing a perfectly good job currently.

Yes, I think a good place to get to would be to have basic functionality with the components in the core, and an optional plugin with extended platform capabilities (like access to backups etc) that could perhaps be injected into wikis.

simonbaird:
Hmm, I haven't consider the idea of Tiddlyhost using the put saver, but it may be doable. I'll have a think about it.

Jermolene:
> Hmm, I haven't consider the idea of Tiddlyhost using the put saver, but it may be doable. I'll have a think about it.

That's an oversight on my part, I meant to suggest it when the TiddlySpot saver mods first came up.

pmario:
IMO the PUT saver was intended to be generic for all servers, that allow PUT and have their own Auth system activated. Imo it was initially developed with a WebDav server as a backend. ...

It turned out, that ETag handling on the server side isn't as standardized as it should be. It is dependent on server side settings. So  409 conflict was the biggest problem, that came up. 

In the meantime, it seems those problems have been resolved. Mainly by adjusting server side settings. There are several threads in the GG.

PR #5623, opened 2021-04-23
List mode for CheckboxWidget

### LINKED ISSUE #5613: [IDEA] CheckboxWidget - generalize to also toggle individual items
The CheckboxWidget is the natural goto for toggling a field value. But, annoyingly, it does not feature toggling of individual items in a field. Unless it is the tags field. So the mechanism *does* exist to toggle individual values... but it just isn't applied for non-tags fields. 

So instead of a direct widget use, one has to construct this and fiddle with *listops* and *​icons*, etc. ​And note that there is no icon that looks like an actual checkbox so if one presents a list with checkboxes to toggle multiple fields (e.g both tags and other fields values) they would get an ugly *mix* of actual checkboxes and surrogate icons.

I wish the CheckboxWidget could be generalized to toggle any individual value in any field. Possily there could be limitations like "there must only be a single instance of the value in the field".

It would be fine to introduce a **new attribute** to specify that it is a single value that should be toggled and not the whole field value.

I did build http://toggle-in-field.tiddlyspot.com/ back in 2018 out of frustration. (BTW, you can see the icon problem there where I've made small png's to resemble checkboxes. Silly!)

rmunn:
Fixes #5613.

This PR will be in draft mode until it's ready to merge, but I'm seeking feedback already. I'll add documentation for the new list modes in the CheckboxWidget, then I'll add the feature itself once the documentation looks good.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
@@ -1,6 +1,7 @@
 caption: checkbox
 created: 20131024141900000
 modified: 20201109090732190
+colors: red orange yellow
 tags: Widgets
 title: CheckboxWidget
 type: text/vnd.tiddlywiki
@@ -21,6 +22,7 @@ The content of the `<$checkbox>` widget is displayed within an HTML `<label>` el
 |tag |The name of the tag to which the checkbox is bound |
 |invertTag |When set to ''yes'', flips the tag binding logic so that the absence of the tag causes the checkbox to be checked  |
 |field |The name of the field to which the checkbox is bound |
+|fieldMode |<<.from-version "5.1.24">> Optional, turns on field list mode if set to "list". Requires //field// to be defined |
 |index|<<.from-version "5.1.14">> The index of the //tiddler//, a [[DataTiddler|DataTiddlers]], to which the checkbox is bound<<.tip "be sure to set the //tiddler// correctly">>|
 |checked |The value of the field corresponding to the checkbox being checked |
 |unchecked |The value of the field corresponding to the checkbox being unchecked |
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
```diff
@@ -47,6 +49,16 @@ This example creates a checkbox that is checked if the field ''status'' is equal
 
 <<wikitext-example-without-html """<$checkbox field="status" checked="open" unchecked="closed" default="closed"> Is it open?</$checkbox><br>''status:'' {{!!status}}""">>
 
+!! Field List Mode
+
+Using the checkbox widget in field mode requires the ''fieldMode'' attribute to be "list", and the ''field'' attribute to specify the name of the field. The ''checked'' attribute specifies the value to be added to (or removed from) the list to correspond to its checked and unchecked states respectively. If ''checked'' is empty but ''unchecked'' is set (to a non-empty value), then the logic will be inverted so that the checkbox will be checked if the value is missing from the list, or unchecked if the value is present. Finally, if both ''checked'' and ''unchecked'' are present, the checkbox will work like a toggle, replacing the ''checked'' value with the ''unchecked'' value and vice-versa.
```
saqimtiaz, line 54: In the very first sentence I assume you mean "Using the checkbox widget in ~~field~~ list mode"

saqimtiaz:
@rmunn this seems logical and intuitive to me.

I am happy to do this piecemeal and discuss along the way. However, I would favour implementing the proposal for a `checkFilter` attribute as well before we merge, just so that we can properly assess and test any interactions that may crop up.

saqimtiaz:
Also, `fieldType` might be an alternative name to `fieldMode` for the new attribute.

editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
```diff
@@ -21,6 +22,7 @@ The content of the `<$checkbox>` widget is displayed within an HTML `<label>` el
 |tag |The name of the tag to which the checkbox is bound |
 |invertTag |When set to ''yes'', flips the tag binding logic so that the absence of the tag causes the checkbox to be checked  |
 |field |The name of the field to which the checkbox is bound |
+|fieldMode |<<.from-version "5.1.24">> Optional, turns on field list mode if set to "list". Requires //field// to be defined |
```
pmario, line 25: what's the name of the default mode?

editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
rmunn: Yep. Will fix.

editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
rmunn: I haven't given the default mode a name, partly because I couldn't come up with one. If the `fieldMode` (which might become `fieldType`) attribute is empty or anything other than "list" then it will work in default mode ("field mode" in the documentation).

pmario: OK. I could think of a "binary", "single" mode as the default. .. So we have "binary" and "list" mode. In the future there may be more. IMO checkbox has 3 possible modes: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#additional_attributes ... indeterminate

pmario: IMO indeterminate could be used for toggle, where neither `checked` nor `unchecked` value are part of the list

twMat:
bump :)

rmunn:
Superseded by #6561.

PR #5627, opened 2021-04-23
Two typo fixups related to surplus tilde chars

simonbaird:
One is a typo I missed in the previous commit.

The other is a similar but unrelated fix I noticed while looking at
the savers list at https://tiddlywiki.com/#GettingStarted


Jermolene:
Thanks @simonbaird

PR #5628, opened 2021-04-24
Revise the TiddlySpot Saver settings form

simonbaird:
Mention TiddlyHost and link to some documentation on the
configuration options.

Also remove the TiddlySpot control panel and backups links since
they no longer work.

Notes:
* The last three fields are no use for TiddlySpot or TiddlyHost, but
  it's possible that someone, somewhere is still using the old
  store.php from Bidix's UploadPlugin, and would miss them if they
  were removed.

  I'd be happy to remove them in a future PR, if it's decided they
  can be retired.

  (If they were removed, I could delete the last row here:)
  https://github.com/simonbaird/tiddlyhost/wiki/TiddlySpot-Saver-configuration-for-Tiddlyhost-and-Tiddlyspot

* It's still called "TiddlySpot Saver" which I think is fine for
  now. TiddlyHost might use a different saving method in future
  so keep the old name seems best.

### DIFF
--- a/core/language/en-GB/ControlPanel.multids
+++ b/core/language/en-GB/ControlPanel.multids
@@ -123,12 +123,12 @@ Saving/TiddlySpot/BackupDir: Backup Directory
 Saving/TiddlySpot/ControlPanel: ~TiddlySpot Control Panel
 Saving/TiddlySpot/Backups: Backups
 Saving/TiddlySpot/Caption: ~TiddlySpot Saver
-Saving/TiddlySpot/Description: These settings are only used when saving to http://tiddlyspot.com or a compatible remote server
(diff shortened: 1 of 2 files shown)


simonbaird:
As suggested in #5622. FYI @saqimtiaz @Jermolene .

simonbaird:
Demo [here|https://pr-5628-tiddlyspot-save-config.tiddlyhost.com/#%24%3A%2FControlPanel:%24%3A%2FControlPanel].

Jermolene:
Excellent, thank you @simonbaird

Jermolene:
> * The last three fields are no use for TiddlySpot or TiddlyHost, but
>   it's possible that someone, somewhere is still using the old
>   store.php from Bidix's UploadPlugin, and would miss them if they
>   were removed.

That seems quite likely, yes!

PR #5635, opened 2021-04-25
fix select refresh

pmario:
This PR fixes the following problem: The second select widget dosen't update on 5.1.23

 - Copy the following code into https://tiddlywiki.com/#New%20Tiddler
 - Click the "Create aa and bb" button
 - Change Select 1

Select 2 won't update. -> The PR fixes that.

 ```
<$button>
<$action-createtiddler $basetitle="aa" $overwrite=yes text="xx"/>
<$action-createtiddler $basetitle="bb" $overwrite=yes text="yy"/>
Create "aa" and "bb"
</$button>

<$select tiddler="test" >
		<option value="aa">aa should result in - xx - </option>
		<option value="bb">bb should result in - yy -</option>
</$select>
.
.
result: <$select tiddler={{test}} >
		<option value="xx">xx</option>
		<option value="yy">yy</option>
</$select>
```

### DIFF
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -170,7 +170,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 SelectWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
 	// If we're using a different tiddler/field/index then completely refresh ourselves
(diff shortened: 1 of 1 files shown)


pmario:
@Jermolene ... It would be nice if this fix could be merged fast.

Jermolene:
Thanks @pmario

PR #5638, opened 2021-04-25
Add 401 and 403 error messages for put saver

simonbaird:
Show the user a more useful message for the case where they're not
correctly authenticated and try to do a "put" save.

The context for this change is that I'm thinking about using the
put saver for TiddlyHost instead of the old upload saver, but it
should be generally useful.

Also includes a name change for the one existing error message. I
want to indicate these three messages are related, so I added the
"Put" prefix to make them appear next to each other in the language
files. (The Error/EditConflict string is used in just this one place
as far as I can tell.)

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -24,7 +24,6 @@ Encryption/RepeatPassword: Repeat password
 Encryption/PasswordNoMatch: Passwords do not match
 Encryption/SetPassword: Set password
 Error/Caption: Error
-Error/EditConflict: File changed on server
 Error/Filter: Filter error
 Error/FilterSyntax: Syntax error in filter expression
 Error/FilterRunPrefix: Filter Error: Unknown prefix for filter run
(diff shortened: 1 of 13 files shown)


simonbaird:
FYI @pmario @Jermolene . This will make the put saver more user friendly for TiddlyHost.

simonbaird:
See also https://github.com/simonbaird/tiddlyhost/issues/148 .

Jermolene:
Hi @simonbaird that looks great. You don't need to worry about the translations; the translators tool picks up core strings that haven't been translated into a particular language.

simonbaird:
The message could be more verbose message, e.g.:

    Permission denied. It appears you don't have permission to save this TiddlyWiki to {saveUrl}.

    Authentication required. You need to authenticate in order to save this TiddlyWiki at {saveUrl}.

simonbaird:
I also considered moving the messages to PutError/* instead of Error/* which would make sense if they were going to be specific to put saving.

simonbaird:
Ack, will push a revision soon without the translations

simonbaird:
Moving out of draft mode. There's a few changes since the original. Let me know if you don't like it.

Jermolene:
Thanks @simonbaird, great stuff.

simonbaird:
Awesome, thanks @Jermolene .

PR #5640, opened 2021-04-25
Fix issue with filter operator and variables

### LINKED ISSUE #5471: [BUG] `filter` filter step throws an uncaught exception or returns wrong results when used with a soft parameter
**Describe the bug**
The `filter` operator, when its parameter is a variable transclusion, throws the following error:

> Uncaught TypeError: Cannot read property 'getVariable' of undefined

This occurs on every render. If the side-by-side previewer is currently on, I additionally cannot get into the tiddler that contains the filter to fix it and have to edit from outside TiddlyWiki to be able to use the wiki again.

**To Reproduce**
1. Create a new wiki.
2. Paste the following MWE into a tiddler:

 ```
 \define fragment() [get[myfield]match<myvariable>]

 <$list filter="[tag[MyTag]filter<fragment>]">
   <<currentTiddler>><br>
 </$list>
 ```

3. Create a new tiddler tagged `MyTag` (the error only occurs when at least one input tiddler reaches the `filter` filter step).
4. The error appears as soon as the tag is added, if the first tiddler is still open.

Everything works as expected when `<myvalue>` is instead a hard parameter, e.g., `match[myvalue]`. When I use a field transclusion, there is no error, but it also doesn't appear to work correctly – with the filter fragment as `[get[myfield]match{!!myvalue}]`, I don't get any output even when the `myvalue` field on the listing tiddler is set to the same value as the `myfield` value on the tagged tiddler. As a sanity check, the filter `[tag[MyTag]get[myfield]match{!!myvalue}]`, starting in the same tiddler, does display one result (but obviously this filter doesn't do the same thing).

**Expected behavior**:
Either the filter works as expected (only tiddlers with tag `MyTag` whose `myfield` field matches the contents of the variable `myvariable` are listed, or (but I don't see any reason why this should be the case) TiddlyWiki at least recognizes that a variable transclusion for the parameter of the `filter` step is not supported and fails gracefully.

It does not make a difference whether the variable `myvariable` is actually defined. I didn't include any definition of the variable in the example above, but the error still occurs even if it has a value.

**Workaround**:
To get around this, you can use a hard parameter and use text substitution to get your desired match value into the fragment macro. This is annoying, but works fine.

Tested on TiddlyDesktop and in Chromium 87.0.4280.141 64-bit on Linux. Originally found the bug on 5.1.23, confirmed it is still present in prerelease as of today.

### LINKED ISSUE #5592: [BUG] RSOE from undefined variable in subfilter
**Describe the bug**
RSOE with `Uncaught TypeError: Cannot read property 'getVariable' of undefined`

**To Reproduce**
```
<$vars sub="[match<undefined>]">
<$list filter="[[arbitrary]filter<sub>]">

</$list>
</$vars>
```

saqimtiaz:
Fixes an issue where the `filter[]` operator throws an error when a variable is used in the filter operand. Added a test.

Fixes #5592
Fixes #5471


saqimtiaz:
@Jermolene this came up on the [GG again today](https://groups.google.com/g/tiddlywiki/c/Y3NnOMh0Qws/m/AIzow5QZAwAJ).

saqimtiaz:
@Jermolene this came up on the [GG again today](https://groups.google.com/g/tiddlywiki/c/Y3NnOMh0Qws/m/AIzow5QZAwAJ).

Jermolene:
Thanks @saqimtiaz

PR #5644, opened 2021-04-26
Fixes Action-listops bug

saqimtiaz:
Fixes ##5576 by avoiding stitching together filter expressions for the original list values in `<$action-listops>`. Optimized the refresh handling as well.


Jermolene:
Great, thanks @saqimtiaz

PR #5648, opened 2021-04-27
Extend keyboard widget

### LINKED ISSUE #5618: [IDEA] Make the keybaord widget more "hackable"
@Jermolene We do have widget subclassing now. I would like to restructure the keyboard widget a little bit, so it's possible to make it easier to change the `.handleChangeEvent()`

At the moment the keyboard widget can't pass variables on to the actions it calls. ... 

At the moment the `.render()` function is a big junk of code that also contains the "keydown" event

```
KeyboardWidget.prototype.render = function(parent,nextSibling) {
	var self = this;
	// Remember parent

// ------------- snip

	domNode.addEventListener("keydown",function (event) {
		if($tw.keyboardManager.checkKeyDescriptors(event,self.keyInfoArray)) {
			var handled = self.invokeActions(self,event);
			if(self.actions) {
				self.invokeActionString(self.actions,self,event);
			}
			self.dispatchMessage(event);
			if(handled || self.actions || self.message) {
				event.preventDefault();
				event.stopPropagation();
			}
			return true;
		}
		return false;
	},false);

// ------------- snip
};
```

I would like to have it similar to:

```
KeyboardWidget.prototype.render = function(parent,nextSibling) {
	var self = this;
	// Remember parent

// ------------- snip

	// Add a event handler
	$tw.utils.addEventListeners(this.inputDomNode,[
		{name: "keydown", handlerObject: this, handlerMethod: "handleChangeEvent"}
	]);

// ------------- snip
};
```

Is there a chance you'd merge it?

pmario:
This PR fixes #5618 [IDEA] Make the keybaord widget more "hackable"

Documentation and working POC can be found at: [https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html](https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html#KeyboardWidget:KeyboardWidget%20%5B%5BPossible%20Keys%5D%5D)

The `KeyboardWidget.prototype.handleChangeEvent = function(event) {` event listener is separated now, which makes it easy to create a "sub-widget"

There are 3 variables, that are available to the actions 

 - **key**   ... eg 1, 2, 3, a, b, c, Enter
 - **code** ... eg: Space Numpad1 ... is needed, otherwise Space, NumpadX and other "special keys" are hard to handle
 - **modifierKey** ... eg: shift, ctrl-shift, ctrl-alt-shift ..

The widget is super easy to use and very flexible now. ... I think it is a nice addition to the event-catcher widget.

The "simple" example is at: https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html#Possible%20Keys%20(Example)

A more complex example is: https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html#Possible%20Keys%20(Example%201) which gives the users a possibility to understand what's possible. 

The biggest part of the PR is the docs


saqimtiaz:
@pmario a few comments and a question:

**Variable names**
We should use different variables names. `key` and `code` both are common enough names that there is a good chance they could overwrite user defined variables and thus are not **backwards compatible**. 

With `modifier` we are somewhat constrained by precedence but please note that all our other widgets that provide information about the modifier key associated with an event use the variable name `modifier` and not `modifierKey`. In `<$eventcatcher>` we use the `event-` prefix for variables derived from event attributes and that might be an option here.


**Documentation**
The documentation is fairly difficult to follow, even for me as a developer and someone who understands what the new functionality is meant to do. I think a simple example of checking the variables for a particular key combination and then executing an action widget would be more helpful in aiding understanding of how to use these variables. I can see that a lot of effort has gone into the the documentation but I fear that as it is now, it is not very helpful and if anything raises more questions than it answers. There definitely also needs to be further explanation as to what `Possible keys` refers to in the KeyboardWidget tiddler. 


**Keyboard shortcut descriptors**
In #5618 I had pointed out that the keyboard widget was designed to be usable with `key descriptor syntax`, such as `((my-shortcut))`, ([which allows making shortcut keys configurable via the Control Panel](https://tiddlywiki.com/#Keyboard%20Shortcut%20Descriptor)). You seemed to indicate that user configurable shortcuts with this syntax would be usable in the new workflow your changes enable. I don't see any examples of this, please clarify. 

The appeal of this approach is being able to use a single keyboard widget for multiple unique shortcuts. It would be a pity if this was not possible for shortcuts defined via the Control Panel using keyboard shortcut descriptors. If indeed this will not be possible then (a) this exception needs to be explained in the documentation and (b) can we find a way to address this shortcoming now since we are already making changes? Proceeding without this capability now will make it harder to accommodate it later on as we will be even more constrained by backwards compatibility.

_I'm not opposed to merging this provided we can document how to make use of the new variables in a user friendly and approachable manner, and ensure backwards compatibility and consistency with other widgets._

pmario:
I think it will be simple, to rename the variables `event-key, event-code and modifier` ... I'll change them

pmario:
The main purpose of the Possible Keys docs is, that there hasn't been any documentation at all, other than the source code.

IMO testing is important. Eg: If you configure `up, down, left and right` as keys, that should be detected you'll get `UpArrow, DownArrow, LeftArrow and RightArrow` back. I didn't know that. ... So for me it was important to provide a possibility for users to test key strokes and see the results. 

I'll try to come up with some more examples, that show how some action logic can look like.

pmario:
The workflow, that I have in mind, is "inverse" to the current shortcut assignment. eg:

 - Tiddler in edit mode
 - I want a dropdown button, that can be activated with keybord eg: `ctrl-0` .. ((my-shortcut))
 - The popup contains a list of dynamically created command buttons 
    - numbered from `1-9` and `0`
    - and `a-z`
    - those are created and activated using the new keyboard widget functions

So there are up to 36 different commands, that can be reached with 3 key strokes. eg: `ctrl-0  7` will select the 7th command in the dropdown. IMO this is an extremely fast and easy way to select a lot of different commands.

The dropdown button can be configured with ((my-shortcut)). The rest of the configuration is free in a defined range.

I did describe 1 set which contains up to 36 commands. But there can be many sets. ... 

So the whole thing I have in mind is a new type of "shortcut configuration", where the shortcut "range" is defined, but the "commands" can be feely configured.

<details><summary>Dropdown Usage</summary>

grafik
</details>

<details><summary>Dropdown Configuration</summary>

grafik
</details>

pmario:
It is possible to handle `((xx-shortcut))` ... But it's kind of an artificial usecase. eg ((my-1)) is converted internally to an object 

`keyInfoArray[0] = { altKey: false, ​ctrlKey: false, ​keyCode: 49, ​metaKey: false, ​shiftKey: false` ... This info is detected by the keyboard-widget and sent to the action. ... 

The algorithm in the action has no info about `((my-1))`. It only gets `event-key: 1` and the `modifier: normal`

<details><summary>Test with ((my-1))</summary>

grafik
</details>

saqimtiaz:
Hi @pmario 

> The main purpose of the Possible Keys docs is, that there hasn't been any documentation at all, other than the source code.

I think we just need to add some explanatory test on the Keyboard Widget tiddler as to what the possible keys information is for. It wont be relevant for most users and comes across as intimidating. Some context will help.

> I'll try to come up with some more examples, that show how some action logic can look like.

Great thank you. Your intended workflow is probably an outlier in terms of complexity for most users, so a simpler example of how to use the variables to execute an action would be very helpful.

> I think it will be simple, to rename the variables `event-key, event-code and modifier` ... I'll change them

Great, thank you.

> So if we want to have the info in the action, we would need to add `((my-1))` to the `keyInfoArray`. Then we could give: "((my-1)) detected" variable to the action. ... BUT this would be completely new code.

Right. This could open the door to be able to use one <$keyboard> widget and one set of actions for multiple keyboard shortcuts that should be handled distinctly. In the actions you can check which shortcut triggered the action and respond accordingly. Thoughts @Jermolene ?

pmario:
I did update the PR and the demo https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html ... Just for the tests with the new names. ... 

There are no improved examples yet

pmario:
>Right. This could open the door to be able to use one <$keyboard> widget and one set of actions for multiple keyboard shortcuts that should be handled distinctly. In the actions you can check which shortcut triggered the action and respond accordingly. Thoughts @Jermolene ?

Yea, It would add an new layer of abstraction which would allow users to change shortcut settings, without the need to change the logic. ... 

BUT If someone renames the shortcut from `((my-1))` to `((my-X))` we will be stuck again. ... 

I personally would want to get this PR merged first and then we can see, what will be needed to add the "abstract" keyboard handling

saqimtiaz:
> BUT If someone renames the shortcut from `((my-1))` to `((my-X))` we will be stuck again. ...

That's not much different from how it is now for renaming shortcut descriptors and would still be a significant improvement on having no alternative to using multiple keyboard widgets for multiple shortcuts.

> I personally would want to get this PR merged first and then we can see, what will be needed to add the "abstract" keyboard handling

I think that's OK as long as we make a point to resolve this before the new release, so that we aren't further constrained by backwards compatibility in implementing it, and can make tweaks to the changes from this PR if need be (though unlikely).

@pmario we also need to add some explanation to the KeyboardWidget tiddler as to what information the new variables carry. See EventCatcherWidget and DroppableWidget tiddlers for examples.

pmario:
@saqimtiaz ... I did (update the  the docs](https://pmario.github.io/kitchensink/5618-improve-keyboard-widget.html#KeyboardWidget)

 - Linking should be improved
 - Examples only contain code that is important for the keyboard widget
 - boilerplate code has its own macro tiddler now
 
So IMO it should be much easier for users now, to see what's important and how they can use it.

pmario:
@Jermolene ... It would be nice, if you could have a look at this PR.

Jermolene:
Hi @pmario the changes look good, but the documentation needs some work.

* Instead of the phrase "Possible Keys", can we use "Keyboard Codes"
* The use of monospaced formatting in "Possible Keys" is confusing. I'd prefer a two level list, and using `<kbd>` to display each code
* "Possible Keys (Example)" is confusing when one first visits it because it is not clear what the label "Results In" means when it is blank
* I don't understand why the "play with me" button exists. Wouldn't it be simpler just to provide the text area with place holder text saying "Click here to try keyboard shortcuts"
* The input area doesn't seem to work properly: if one types quickly, it is possible to type multiple characters which seems to confuse it
* It's not clear what the section "Test Configuration" is intended to convey. Are those examples for users to try?
* The `<<showCode>>` macro uses the `<summary>/<details>` elements, which we've already discussed is not appropriate for the core
* Your text uses exclamation marks too liberally. In English there are very situations where we'd use an exclamation in documentation. In most cases, it would be better to use the warning icon we use elsewhere in the docs
* I don't understand the purpose of "Possible Keys (Example 2)" or "Possible Keys (Example 3)"

While I have no objection to the code changes, I think these docs changes are too ambitious, and do not succeed. Besides documenting the changes, I think the only the only other thing we really need is the table of key codes.

pmario:
I'll do the changes soon - Thaks for the feedback.

pmario:
> - The `<<showCode>>` macro uses the `<summary>/<details>` elements, which we've already discussed is not appropriate for the core

That's exactly the behaviour I want. If the tiddler is closed and reopened, the code section should be hidden again. I would have to implement it with a `<$reveal type=popup`, which makes it much more complicated, since the popup also closes, if I click outside the area. (That's _not_ what I want) So using reveal will be a lot more complicated to implement.

Details / Summary is convenient to use and does exactly what I want for this usecase.

pmario:
@Jermolene ... I think I did address all your suggestions. 

 - The `<details>` html element was removed and replaced by a button that keeps the state. 
 - Did change the wording according to your suggestions
 - The "Keyboard Codes" page uses `<kbd>` elements now.
 - The first example also shows ´<kbd>` elements
 - The second "advanced" examples has to use the "code view", since the configuration input, for the keyboard widget, is also used as documentation output to avoid code duplication
 - Removed exclamation marks
 - Removed (Example 2) since it can't be used that way properly. ... We will need a new PR
 - Removed the "Play with me" button

Jermolene:
Thank you @pmario

saqimtiaz:
@Jermolene @pmario was it intentional to keep this tiddler as part of the docs? 
https://tiddlywiki.com/prerelease/#Key%20Codes%20(Example%201)

Personally I find it confusing and don't understand what value it adds when we have the simpler example at https://tiddlywiki.com/prerelease/#Key%20Codes%20(Example)

pmario:
>@Jermolene @pmario was it intentional to keep this tiddler as part of the docs?
https://tiddlywiki.com/prerelease/#Key%20Codes%20(Example%201)

It shows the "real possibilities" with `ctrl-alt-shift` combinations and much more keys. ... I think it is important to see how they work. ... eg: Number-Pad, Enter, Backspace and so on. .. 

I think in general, we don't have enough examples, that show how our user facing functions do work..

Jermolene:
> @Jermolene ... I think I did address all your suggestions.

Not quite – as I said above:

> While I have no objection to the code changes, I think these docs changes are too ambitious, and do not succeed. Besides documenting the changes, I think the only the only other thing we really need is the table of key codes.

I've only just noticed that the additional docs updates are still present, so I'll remove them shortly.

I've no objection to re-introducing these docs, but they need improvement and then some testing with users.

PR #5653, opened 2021-05-01
:sort filter run prefix

### LINKED ISSUE #5166: [IDEA]  :sort prefix and extensible filter run prefixes

saqimtiaz:
This PR adds a `:sort` filter run prefix and fixes #5166 . It includes documentation and tests.

The exact syntax for the `:sort` prefix is:

`:sort:<sortType>:<casesensitive>,<reverse>[...filter...]`

To make it easier to review the syntax and documentation I have posted it [here](https://saqimtiaz.github.io/sq-tw/temp/sort-prefix.html).

This PR extends `$tw.utils.makeCompareFunction` in a backwards compatible manner to support alphanumeric sorting and a flag for case sensitivity. Since the `sortsub[]` operator uses the same method for sorting, `sortsub` also now supports alphanumeric sorting. I've added documentation and a test for this. I'll add a case sensitivity option to `sortsub[]` in a separate PR.


Jermolene:
Thanks @saqimtiaz, looks good. Let me know when you're ready to merge.

saqimtiaz:
@Jermolene  this is good to merge.

Jermolene:
Thanks @saqimtiaz

PR #5655, opened 2021-05-01
Extend <$action-popup> to allow creating floating popups

saqimtiaz:
This PR extends `<$action-popup>` with an optional attribute `$floating` to allow it to create floating popups that must be cleared manually. The popup mechanism already supports floating popups, we are just adding an affordance for creating one.


Jermolene:
Thanks @saqimtiaz good idea.

Issue #4816, opened 2020-08-14
[IDEA] Textify filter output


twMat:
(IMO, this borders on being a "bug")

I wish for a convenient way to display calculated output as **text**. Currently, filters always output *links* when the output quite often is not supposed to link to anything. It is often too messy and inconvenient to go the full route with a ListWidget+ViewWidget.
For example:

```
;Total time: <<total>> h.
:Physics {{{ [<total>multiply[.7]round[]] }}} h
:Chemistry {{{ [<total>multiply[.3]round[]] }}} h
```

Thus, I need a smooth way to "de-linkify" the output, ideally both for filtered transclusion and for the ListWidget.

One syntactic idea would be to prefix the filtered transclusion or even the full ListWidget with a `~`, e.g

`~{{{ ... }}}` and `~<$list ....`

This `~` would distribute onto all output items.

**Reflection**
I believe "delikification" has been neglected because filters, in the early days, were mostly limited to dealing with actual tiddler titles (i.e title links). As the filter language has become much more powerful it is now clear that we need this possibility.

pmario:
Tested on https://tiddlywiki.com 

**first version:** ... no links for whole tiddler

```
\define tv-wikilinks() no

{{{[tag[HelloThere]]}}}
```
docs: https://tiddlywiki.com/#tv-wikilinks%20Variable:%5B%5Btv-wikilinks%20Variable%5D%5D%20%5B%5Btv-wikilinks%20Variable%20(Examples)%5D%5D

**second**: per filtered transclusion

```
{{{[tag[HelloThere]]||noLinks}}}
```

```
title: noLinks

<$view format=text field=title/>, 
```

IMO that should do the tricks. Download the zip below. it contains the tids for easy testing :)

[filtered-transclusion-no-link.zip](https://github.com/Jermolene/TiddlyWiki5/files/5073386/filtered-transclusion-no-link.zip)

saqimtiaz:
I do something like this:
`{{{[tag[HelloThere]]||$:/noLinks}}}`

or

`<$text text={{{[tag[HelloThere]]}}}/>`

Jermolene:
Hi @twMat I'd definitely now consider it a mistake that the default template for the list widget generates links, and I appreciate the need for a concise means of rendering a text representation of the list output.

[...]

PR #5665, opened 2021-05-05
Format:titlelist operator

saqimtiaz:
This PR adds a `quotify` operator to wrap individual titles in square brackets if the title contains spaces, otherwise returning the title unchanged. Background discussion [here](https://github.com/Jermolene/TiddlyWiki5/discussions/5662#discussioncomment-687500).

**This is not a reducing operator,** as the ability to wrap individual titles in brackets is more versatile.

I think the name of the operator needs further consideration as the name may be misunderstood by users as implying wrapping a title in quotes characters. Suggestions are very welcome. (@rmunn @kookma @twMat )

I will work on the docs a bit more, pushing this now to facilitate collaboration with @pmario on what he is working on.


twMat:
(So I interpret this to be an invitation to comment here, in the PR thread)

Well, the obvious suggestion would be `brackify` even if it sounds a bit silly. At least something based on the term *brackets* because there's probably no other single word label that is more descriptive. One will have to read the docs to understand why it doesn't *always* result in a bracket wrapping. In that sense, a made up term like "brackify" might give indication that there is more to it than what a more established term like "brackets" would imply.

We might also try to project ahead if there is some other comparable operator that wraps titles, in which case the current operator might be better as a *suffix* to a more general "wrap" operator.

(Side note: I've previously called ( #4816  ) for some solution to get *text* output, not always links, from lists. Maybe the ListWidget could peek on the output and delinkify items wrapped with X and then remove the X wraps?)

Also, a general "wrap" operator might be an opportunity that would format the output to e.g third party services. Maybe service X requires lists with some other wrapping. (Hm... could one wrap things to somehow create a JSON format?)

saqimtiaz:
@twMat it's interesting that you mention wrap because I originally wrote this as a generic `wrap[]` operator that accepted two optional operands, and defaulted to wrapping in square brackets when no operands are provided.

However, a simple test case of wanting to wrap a title in wikitext for an image showed that there isn't any real benefit over using `addSuffix` and `addPrefix` as one has to use variables to define the  suffix and prefix.

Also the selective behaviour of only wrapping when the title contains spaces isn't intuitive for a generic `wrap` operator.

As such I decided to revert to a simpler operator that just wraps in square brackets when needed.

I'm inclined to consider `format:titlelist` again.

twMat:
>`format` operator

Ah, there we have it already. Yes, `:titlelist` makes sense. 

>I think the concept of a title list is rather fundamental in TiddlyWiki, so this could be an opportunity to draw further attention to it.

+1

Probably off topic but it would be nice if title lists could also be treated like arrays in a more traditional programming sense. Just sayin.

kookma:
> I'm inclined to consider `format:titlelist` again.

I think the `titlelist` is the semantic for this case!

rmunn:
I like the name `quotify`, as it would allow some expanded semantics by providing an argument. E.g., by default it wraps in square brackets, but if you give an argument then it can wrap things in other characters, e.g. `quotify[()]` would turn `Tiddler Two` into `(Tiddler Two)`. The semantics would be that the input string must have an even number of characters, and the first half will be the prefix while the second half would be the suffix. So `quotify[(**)]` would turn `Hello There` into `(*Hello There*)`. That would be versatile, allowing people to wrap things in double-quotes or single-quotes or even triple-quotes (though they'd have to use a variable to pass in the parameter in that case), or in `{}` or `()` for various needs that we haven't though of yet, like passing title lists to some other software with weird quoting semantics. The default behavior with no input would still be to wrap in brackets.

saqimtiaz:
@rmunn I had the same general idea at first. However, the logic of only wrapping when there are spaces as well as the need to assign variables means there isn't much benefit over using `addsuffix` and `addprefix`

saqimtiaz:
@Jermolene before I refactor this as a `format:titlelist[]` operator suffix and update docs accordingly, any thoughts?

In brief: 
- An operator that is non-reducing and operates on individual titles is more versatile
- generic `quotify` or `wrap` operator is not intuitive since we only wrap when there are spaces, and common use cases like wrapping a title in image syntax would require variables anyway due to square brackets, so one can use addsuffix and addprefix.

saqimtiaz:
> @Jermolene before I refactor this as a `format:titlelist[]` operator suffix and update docs accordingly, any thoughts?
> 
> In brief:
> 
> * An operator that is non-reducing and operates on individual titles is more versatile
> * generic `quotify` or `wrap` operator is not intuitive since we only wrap when there are spaces, and common use cases like wrapping a title in image syntax would require variables anyway due to square brackets, so one can use addsuffix and addprefix.

@Jermolene thoughts on whether you agree with implementing this as a non-reducing `format:titlelist[]` operator for the reasons given in the reply I have quoted? Or do we keep it as is in this PR?

Jermolene:
> > @Jermolene before I refactor this as a `format:titlelist[]` operator suffix and update docs accordingly, any thoughts?
> > In brief:
> > 
> > * An operator that is non-reducing and operates on individual titles is more versatile
> > * generic `quotify` or `wrap` operator is not intuitive since we only wrap when there are spaces, and common use cases like wrapping a title in image syntax would require variables anyway due to square brackets, so one can use addsuffix and addprefix.
> 
> @Jermolene thoughts on whether you agree with implementing this as a non-reducing `format:titlelist[]` operator for the reasons given in the reply I have quoted? Or do we keep it as is in this PR?

Thanks @saqimtiaz I agree that this should be a non-reducing `format:titlelist[]` operator, will be good to have.

saqimtiaz:
@Jermolene this has now been refactored as a `format:titlelist` operator. Tests and documentation are included.

Jermolene:
Thanks @saqimtiaz

PR #5682, opened 2021-05-12
Adding a new plugin "Tiddloid Tweaks"

donmor:
This PR creates a plugin that applies appearance tweaks to Tiddloid, and updates `Saving on Android` tiddler as well.

Currently it have these features:

* Hiding the toolbar when the file is fully loaded (Thanks to @jdjdjdjdjdjd's [idea](https://github.com/donmor/Tiddloid/issues/7))
* Making the toolbar (as well as system status bar and navigation bar) the same color to the page background

Here's some screenshots during the testing:

Screenshot_1620805108&nbsp;&nbsp;&nbsp;&nbsp;Screenshot_1620805139

There's also another PR (#5587), which integrates the contents into the core. It's expected to merge one of the two PRs and reject another.


Jermolene:
Hi @donmor packaging the tweaks for Tiddloid into a separate plugin is a good idea, but I don't think it's appropriate for the core official plugin library for the following reasons:

* The core plugin library is for plugins that have been adopted by the community. Tiddloid is a separate project, and its components should stay together
* The core plugin library is only updated when there's a new release of TiddlyWiki, so it will constrain Tiddloid to following TiddlyWiki's release schedule

donmor:
Okay. You mean this PR would be merged and closing the other, right?

donmor:
One more thing, is there an external plugin repository for all TiddlyWiki users that holds third-party plugins and provides quick plugin installation?

Jermolene:
> Okay. You mean the two PR are all not appropriate?

Yes, I'm afraid that's right.

> So is there another convinient way for users to install the plugin?

We do plan to make a third party plugin library (the links aggregator at https://links.tiddlywiki.com is a dry run for some of the techniques we intend to use). In the meantime, prolific plugin authors like @kookma maintain their own plugin librarys.

>I put a zip containing the plugin tid file along with the apk on my github release page, but there are users using android markets, and few of them tend to visit github or, due to internet restriction, can't access github.

I'm not familiar with how Tiddloid works, but the technique that Quine uses is to inject the JS functions and tiddlers into the wiki as it is loaded.

donmor:
Thanks for quick reply.
You mean Quine dynamically adding new tiddlers to the wiki? I know it is possible to create tiddlers using Javascript(tested some $tw stuff in the console of my browser), but the injected tiddlers will be saved into the file, even if users doesn't tend to do so. I don't want to make additional changes to the wikis if not necessary(TW classic is an exception, I found they won't work on Android's `content://` address during early development of v2 version, and I have to inject something to force saving configs internally instead of using cookies - Androidzs 'security' stuff again that prevents cookies to work).

linonetwo:
@donmor It is already on CPL, so don't forget to close this issue.

linonetwo:
And I suggest you make a standalone github repo for this plugin, so PR and release of it maybe easier. You can try https://github.com/tiddly-gittly/TiddlyWiki-WikiText-Plugin-Template

donmor:
Oh, I forgot this :-)

PR #5686, opened 2021-05-15
Bug fixes for Eventcatcher (FF and IE)

saqimtiaz:
These are targeted bug fixes for FF and IE.

On IE, `<$eventcatcher>` causes a red error screen as soon as an event is trapped. This is because IE does not support the `matches()` method but has its own equivalent `msMatchesSelector()` method. A utils method `domMatchesSelector()` has been introduced.

In FF there is a quirk where some events (such as `dragenter` and `dragleave`) can be fired for text nodes (nodeType 3) which are not proper Elements and therefore checking their attributes throws an error. To resolve this, a check for `instanceof Element` has been added.

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -22,6 +22,10 @@ exports.domContains = function(a,b) {
 		!!(a.compareDocumentPosition(b) & 16);
 };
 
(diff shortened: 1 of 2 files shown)


saqimtiaz:
Will open a new PR with a cleaner commit history.

Issue #5616, opened 2021-04-19
[BUG] Dropzone dragover state can incorrectly persist


saqimtiaz:
When a file is being dragged over a `<$dropzone>`, the DOM node created by the dropzone is given the class of `tc-dragover`. However, this class is not always correctly removed when the dragged file leaves the dropzone.

This particular issue surfaces on Firefox on both Windows 10 and Ubuntu 20.04 but not in Chrome (or at least not frequently enough that I can reproduce it). I've investigated the cause and while I do not have a clean solution for this yet, I will summarize my findings at the end of this issue.

The following video demonstrates the issue with a `<$dropzone>` with a custom class, with CSS to have a red border when it has the class `tc-dragover`.  Note how the red border around the dropzone persists even after the mouse is no longer over the dropzone's DOM Node, and even after the file has been dropped and an import triggered.

https://user-images.githubusercontent.com/68092/115244765-8b20a880-a124-11eb-9d54-b9310cad8c13.mp4

The file used for the video demonstration can be found here: https://saqimtiaz.github.io/sq-tw/temp/dropzone-bug.html

**While it takes some rather persistent erratic mouse movement here to trigger the bug, at times it can happen even the first time the file is dragged over a dropzone.**

All of the following contribute to the `dragleave` event firing less consistently and this problem cropping up more frequently:

- greater speed of drag movement
- no margin or padding between the nested DOM elements that we drag over
- entering and leaving the dropzone from different directions.
- the presence of an iframe or input element immediately inside the dropzone with no margin or padding.

`<$dropzone>` maintains an array of all DOM elements entered in the `dragenter` event, and then removes elements from this list in the `dragleave` event. The `dragleave` and `dragenter` events are fired for the DOM node created by the dropzone as well as all descendant DOM nodes. The logic is sound. 

The problem is that the browser implementation is such that the `dragleave` and `dragenter` events are not fired consistently on all descendant elements of the dropzone. So dragging across a dropzone with a child DIV inside it, you may get two `dragenter` events but only one `dragleave` event, in which case the `tc-dragover` class is never removed for the `<$dropzone>` widget. Alternatively you may get the same number of `dragenter` events as `dragleave`, but on different DOM nodes.

pmario:
I thin we did have a similar problem with the **list-links-draggable** macro. ... So we may have a closer look there. ... Just a guess!

saqimtiaz:
[...]

PR #5688, opened 2021-05-16
Dropzone: persistent dragover state fix

saqimtiaz:
This is a quick draft PR with a proposal for how to resolve the issues described in #5616 where the `dragover` state incorrectly persists in `<$dropzone>`. If we are amenable to this approach I will clean this up for merging.

**Summary of the problem:**
The `dragover` state incorrectly persists whenever the dragged object leaves the dropzone DOM node and the count of entered/exited nodes in `currentlyEntered` is inaccurate and not 0. The underlying reason is that the `dragleave` and `dragenter` events do not always fire for all DOM nodes, particularly in Firefox.

After quite extensive research and having tried a fair few things, I have finally identified an approach that can resolve this issue.

**Proposed workaround:**
- The first time `dragenter` fires within the dropzone, we attach `dragenter`, `dragleave` and `click` listeners to the `document.body`.
- In this `dragenter` listener, we check the `event.target` to see if it is a descendant of the dropzone DOM node. If it is not we know the drag has moved outside the dropzone, we reset the state of the dropzone, which includes removing the event listeners we had attached. This relies on the fact that whenever the drag leaves the dropzone a `dragenter` is fired on a DOM node outside the dropzone.
- This however doesn't cover the scenario where the dropzone covers nearly the entire viewport and the dragged file leaves the browser window. For this we check in the `dragleave` listener on `document.body` if the `relatedTarget` is null. If so, the drag has left the window and we can reset the state of the dropzone.
- There is a `resetState()` method in the widget that is called for all scenarios where the drag has concluded in the dropzone. This method removes the dragover class, sets `currentlyEntered` to an empty array and removes the extra event listeners that were added.
- The above accounts for all cases where the dragover state can incorrectly persist apart from using the <kbd>Escape</kbd> key to cancel the drag event in webkit browsers. This does not fire a dragend or dragleave event. As mitigation towards this, we have the `click` event listener which resets the dropzone.

I do not think there is an alternative fix that would not involve listening for events on the entire document. The extra event listeners are only present for the duration of the drag.

With this approach we could actually rewrite the entire dropzone widget and remove the need for tracking DOM nodes entered and left in `currentlyEntered[]`. However, the existing solution is more semantically correct and we are only having issues due to the HTML5 drag and drop spec being flawed and the browser implementations inconsistent. Therefore I would be in favour of adding these additional checks to the existing implementation as in the code in this PR. This will make it easier to remove this additional code once the browser implementation gets cleaned up (one can hope!).

Thoughts @Jermolene?

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -70,7 +70,7 @@ DropZoneWidget.prototype.handleEvent = function(event) {
 		}
 	} else if(event.type === "dragleave") {
 		// Check if drag left the window
-		if (event.relatedTarget === null || event.relatedTarget.nodeName === "HTML") {
+		if (event.relatedTarget === null || (event.relatedTarget && event.relatedTarget.nodeName === "HTML")) {
 			this.resetState();
 		}
 	}
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -70,7 +70,7 @@ DropZoneWidget.prototype.handleEvent = function(event) {
 		}
 	} else if(event.type === "dragleave") {
 		// Check if drag left the window
-		if (event.relatedTarget === null || (event.relatedTarget && event.relatedTarget.nodeName === "HTML")) {
+		if(event.relatedTarget === null || (event.relatedTarget && event.relatedTarget.nodeName === "HTML")) {
 			this.resetState();
 		}
 	}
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz that looks great (and would be terrific to get this fixed, it's been outstanding for a long time).

saqimtiaz:
@Jermolene thank you for the feedback. I'll clean up this up for merging later today.

saqimtiaz:
@Jermolene this is good to merge. I've tested on Chrome, Firefox and IE and fixed a couple of minor issues.

Let me know if you feel any code changes are needed.

Jermolene:
Thanks @saqimtiaz

PR #5691, opened 2021-05-17
CurrentTiddler variable consistency in subfilters and prefixes

### LINKED ISSUE #5683: currentTiddler variable and inconsistency in subfilter operators
Currently we do not have consistency in what the value of `currentTiddler` is set to for filters operators that implement subfilters which are invoked once for each input title:

- `filter[]` : unchanged value for `currentTiddler`
- `reduce[]` : the value of `currentTiddler` is set to the respective title each time, allowing use of syntax like `{!!text}` within the subfilter
- `sortsub[]` : the value of `currentTiddler` is set to the respective title each time, allowing use of syntax like `{!!text}` within the subfilter

Similarly the `:reduce` and `:sort` filter run prefixes set the value of `currentTiddler` to the respective title but `:filter` does not.

This inconsistency between `filter[]` and the other two subfilter operators is confusing. Can we reasonably still change this behaviour for `filter[]` to make it consistent with the other two operators? Note that until now using variables in the operand to the `filter[]` operator has not been possible due to the bug recently resolved in #5640 for TW 5.1.24. This should help with backwards compatibility concerns.

Alternatively we could use a filter operator suffix to control what to set the value of currentTiddler to.

Furthermore, having a way to reference the value of `currentTiddler` from the main filter run can still be useful. This could be made available in the subfilter under a different variable name. I haven't given much thought to names but something indicating that this is the value of `currentTiddler` outside the subfilter. A (not ideal) example is `outerCurrentTiddler`.

saqimtiaz:
Fixes #5683 

This PR makes the value of currentTiddler consistent in filters operators that implement subfilters which are invoked once for each input title, as well as their corresponding filter run prefix equivalents.

Documentation has also been updated.

### DIFF
--- a/core/modules/filterrunprefixes/filter.js
+++ b/core/modules/filterrunprefixes/filter.js
@@ -17,10 +17,21 @@ exports.filter = function(operationSubFunction,options) {
 	return function(results,source,widget) {
 		if(results.length > 0) {
 			var resultsToRemove = [];
-			results.each(function(result) {
-				var filtered = operationSubFunction(options.wiki.makeTiddlerIterator([result]),widget);
+			results.each(function(title) {
+				var filtered = operationSubFunction(options.wiki.makeTiddlerIterator([title]),{
+					getVariable: function(name) {
+						switch(name) {
+							case "currentTiddler":
+								return "" + title;
+							case "outerCurrentTiddler":
+								return widget.getVariable("currentTiddler");
+							default:
+								return widget.getVariable(name);
+						}
+					}
+				});
 				if(filtered.length === 0) {
-					resultsToRemove.push(result);
+					resultsToRemove.push(title);
 				}
 			});
 			results.remove(resultsToRemove);
--- a/core/modules/filterrunprefixes/filter.js
+++ b/core/modules/filterrunprefixes/filter.js
@@ -23,7 +23,7 @@ exports.filter = function(operationSubFunction,options) {
 						switch(name) {
 							case "currentTiddler":
 								return "" + title;
-							case "outerCurrentTiddler":
+							case "..currentTiddler":
 								return widget.getVariable("currentTiddler");
 							default:
 								return widget.getVariable(name);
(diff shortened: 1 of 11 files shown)


saqimtiaz:
@Jermolene apologies if this comes as an interruption, bumping PRs that are relatively straight forward and should be merged prior to 5.1.24

Jermolene:
I wonder if instead of using "outerCurrentTiddler" we might do something along the lines of the way that parsing macros create variables with the decorated name of each parameter. For example, "..currentTiddler"?

saqimtiaz:
@Jermolene I am not tied to the name `outerCurrentTiddler` but I worry that the distinction between something like `..currentTiddler` vs `currentTiddler` will be confusing for users, even though I personally find it appealing. 

Also, introducing a new syntax only makes sense if we can already think of other places the same concept will be used. Can you think of any?

Jermolene:
> @Jermolene I am not tied to the name `outerCurrentTiddler` but I worry that the distinction between something like `..currentTiddler` vs `currentTiddler` will be confusing for users, even though I personally find it appealing. 

It's definitely a geeky reference, but it's also a relatively advanced concept, so I'd be in favour of `..currentTiddler`

> Also, introducing a new syntax only makes sense if we can already think of other places the same concept will be used. Can you think of any?

Crazy, but perhaps we could go on to build this into the variable retrieval code, making a general purpose mechanism for getting at variables overridden further up the chain. It might not be practical now, but it would make a lot of sense for unrolling black box functionality like the "qualify" macro.

saqimtiaz:
Good points. I will refactor this to use `..currentTiddler` later today. Thank you.

Currently playing with your epub work to explore some ideas your demo inspired for leveraging some of those ideas in other domains.

saqimtiaz:
@Jermolene I've updated the code and docs to use `..currentTiddler` instead of `outerCurrentTiddler`.
This should be good to go.

Jermolene:
Excellent @saqimtiaz, thanks.

PR #5693, opened 2021-05-17
sort plugins, themes and languages in Control Panel

kookma:
This PR address the issue raised in discussion board!
See: https://github.com/Jermolene/TiddlyWiki5/discussions/5256

This PR does as below

- put the $:/core as first item in Plugins list in control panel
- it then lists **first the official plugins** [IMPORTANT]
- after official plugins it shows third party plugins
- it sorts plugins by name (unfortunately not all plugins have name! This PR did not touch this part)

@saqimtiaz I appreciate to have a look at the filter I used!

### DIFF
--- a/core/ui/ControlPanel/Plugins.tid
+++ b/core/ui/ControlPanel/Plugins.tid
@@ -7,7 +7,8 @@ caption: {{$:/language/ControlPanel/Plugins/Caption}}
 \define plugin-table(type)
 <$set name="plugin-type" value="""$type$""">
 <$set name="qualified-state" value=<<qualify "$:/state/plugin-info">>>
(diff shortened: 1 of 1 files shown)


kookma:
This is the result

1- Core
2- Official plugins
3- Third party plugins


image


I thought may be we could use a different  style for third party plugins!

saqimtiaz:
If you look at https://tiddlywiki.com/prerelease/ the plugins are already sorted alphabetically by name. I think that is more intuitive and user friendly unless we want to take the step of splitting it into separate lists for core and 3rd party plugins.

Personally I think one list makes sense and then alphabetical ordering of all the titles makes it easiest to find what you are looking for.

kookma:
Hi @saqimtiaz 

look at this

image

The issue is TW mixes up as not all third party plugins has correct name! and also one cannot distinguish which is official which is not!

kookma:
Should I close this PR?

kookma:
Thank you @Jermolene

PR #5695, opened 2021-05-17
Focus-Preservation over refresh cycles II

BurningTreeC:
This PR preserves the widget owning the domNode that has focus over refresh cycles.
With this PR, we're able to focus a tiddler (for example) by clicking it, then moving focus to another tiddler using a keyboard shortcut, editing the focused tiddler using a keyboard shortcut and so on

This is a draft PR for discussions and should replace #5414 but I don't want to close that PR yet

### DIFF
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -86,10 +86,12 @@ function FramedEngine(options) {
 		{name: "click",handlerObject: this,handlerMethod: "handleClickEvent"},
 		{name: "input",handlerObject: this,handlerMethod: "handleInputEvent"},
 		{name: "keydown",handlerObject: this.widget,handlerMethod: "handleKeydownEvent"},
-		{name: "focus",handlerObject: this,handlerMethod: "handleFocusEvent"}
+		{name: "focus",handlerObject: this,handlerMethod: "handleFocusEvent"},
+		{name: "blur",handlerObject: this, handlerMethod: "handleBlurEvent"}
 	]);
 	// Insert the element into the DOM
 	this.iframeDoc.body.appendChild(this.domNode);
+	this.widget.domNodes.push(this.domNode);
 }
 
 /*
@@ -105,6 +107,25 @@ FramedEngine.prototype.copyStyles = function() {
 	// In Chrome setting -webkit-text-fill-color overrides the placeholder text colour
 	this.domNode.style["-webkit-text-fill-color"] = "currentcolor";
 };
+
+/*
+Get an object containing the selectionStart and selectionEnd values
+*/
+FramedEngine.prototype.getSelectionRange = function() {
+	return {
+		selectionStart: this.domNode.selectionStart,
+		selectionEnd: this.domNode.selectionEnd
+	}
+};
+
+/*
+Set the selection-range
+*/
+FramedEngine.prototype.setSelectionRange = function(selectionStart,selectionEnd) {
+	this.domNode.selectionStart = selectionStart;
+	this.domNode.selectionEnd = selectionEnd;
+};
+
 
 /*
 Set the text of the engine if it doesn't currently have focus
@@ -171,6 +192,20 @@ Handle a focus event
 FramedEngine.prototype.handleFocusEvent = function(event) {
 	if(this.widget.editCancelPopups) {
 		$tw.popup.cancel(0);	
+	}
+	var currentTiddler = this.widget.document.querySelector('[data-tiddler-title="' + CSS.escape(this.widget.editTitle) + '"].tc-tiddler-frame');
+	if(currentTiddler) {
+		$tw.utils.addClass(currentTiddler,"tc-focused");
+	}
+};
+
+/*
+Handle a blur event
+*/
+FramedEngine.prototype.handleBlurEvent = function(event) {
+	var currentTiddler = this.widget.document.querySelector('[data-tiddler-title="' + CSS.escape(this.widget.editTitle) + '"].tc-tiddler-frame');
+	if(currentTiddler) {
+		$tw.utils.removeClass(currentTiddler,"tc-focused");
 	}
 };
 
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -234,6 +234,36 @@ Widget.prototype.hasVariable = function(name,value) {
 	}
 	return false;
 };
+	
+	/*
+Find the widget, walking up the widget-tree, that has a different transclusion variable than the current widget
+*/
+Widget.prototype.findParentTransclusionWidget = function() {
+	var node = this;
+	var transclusionVariable = this.getVariable("transclusion");
+	while(node && node.getVariable("transclusion") === transclusionVariable) {
+		node = node.parentWidget;
+	}
+	return (node !== this) ? node : null;
+};
+
+/*
+Generate the "transclusion-footprint" of the current widget in the widget-tree up to the next widget that changes the transclusion variable
+*/
+Widget.prototype.generateWidgetTreeFootprint = function() {
+	var parentTransclusionWidget = this.findParentTransclusionWidget(),
+	    node = this,
+	    footprint = node.parentWidget.children.indexOf(node);
+	while(node) {
+		node = node.parentWidget;
+		if(node === parentTransclusionWidget) {
+			break;
+		} else if(node.parentWidget && node.parentWidget.children) {
+			footprint = footprint + "-" + node.parentWidget.children.indexOf(node);
+		}
+	}
+	return footprint;
+};
 
 /*
 Construct a qualifying string based on a hash of concatenating the values of a given variable in the parent chain
--- a/core/ui/EditTemplate.tid
+++ b/core/ui/EditTemplate.tid
@@ -1,5 +1,16 @@
 title: $:/core/ui/EditTemplate
 
+\define get-focus-selector()
+[data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame
+\end
+\define focus-tiddler-actions(afterOrBefore)
+<$set name="nextTiddler" value={{{ [list<tv-story-list>$afterOrBefore$<currentTiddler>] }}}>
+<$set name="cssEscapedTitle" value={{{ [<nextTiddler>escapecss[]] }}}>
+<$action-navigate $to=<<nextTiddler>>/>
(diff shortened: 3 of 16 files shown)


core/modules/editor/engines/framed.js
```diff
@@ -172,6 +193,20 @@ FramedEngine.prototype.handleFocusEvent = function(event) {
 	if(this.widget.editCancelPopups) {
 		$tw.popup.cancel(0);	
 	}
+	var currentTiddler = this.widget.document.querySelector('[data-tiddler-title="' + CSS.escape(this.widget.editTitle) + '"].tc-tiddler-frame');
```
pmario, line 196: `data-tiddler-title` doesn't need to be excaped. That's the advantage of `data-tiddler`

core/modules/editor/engines/framed.js
```diff
@@ -172,6 +193,20 @@ FramedEngine.prototype.handleFocusEvent = function(event) {
...
+	if(currentTiddler) {
+		$tw.utils.addClass(currentTiddler,"tc-focused");
+	}
+};
+
+/*
+Handle a blur event
+*/
+FramedEngine.prototype.handleBlurEvent = function(event) {
+	var currentTiddler = this.widget.document.querySelector('[data-tiddler-title="' + CSS.escape(this.widget.editTitle) + '"].tc-tiddler-frame');
```
pmario, line 206: same as above.

core/ui/EditTemplate.tid
```diff
@@ -1,5 +1,16 @@
 title: $:/core/ui/EditTemplate
 
+\define get-focus-selector()
+[data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame
+\end
+\define focus-tiddler-actions(afterOrBefore)
+<$set name="nextTiddler" value={{{ [list<tv-story-list>$afterOrBefore$<currentTiddler>] }}}>
+<$set name="cssEscapedTitle" value={{{ [<nextTiddler>escapecss[]] }}}>
```
pmario, line 8: no css escaping needed!

core/ui/ViewTemplate.tid
```diff
@@ -3,8 +3,29 @@ title: $:/core/ui/ViewTemplate
 \define folded-state()
 $:/state/folded/$(currentTiddler)$
 \end
+\define get-focus-selector() [data-tiddler-title="$(cssEscapedTitle)$"].tc-tiddler-frame
+\define focus-tiddler-actions(afterOrBefore)
+<$set name="nextTiddler" value={{{ [list<tv-story-list>$afterOrBefore$<currentTiddler>] }}}>
+<$set name="cssEscapedTitle" value={{{ [<nextTiddler>escapecss[]] }}}>
```
pmario, line 9: and again

pmario (commented):
I think there is a conceptual problem with `data-tiddler-title`!!

core/modules/widgets/widget.js
```diff
@@ -234,6 +234,36 @@ Widget.prototype.hasVariable = function(name,value) {
 	}
 	return false;
 };
+	
+	/*
```
pmario, line 238: whitespace?

core/modules/editor/engines/framed.js
BurningTreeC: Imo the escaping is needed

core/ui/ViewTemplate.tid
BurningTreeC: Also here I think escaping is needed

core/modules/editor/engines/framed.js
pmario: The ViewTemplate doesn't do any escaping. So the code won't work https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FViewTemplate

Why do you think css-escaping is needed for data-xxx ?

pmario: the html-spec doesn't limit the "value" in any case: https://html.spec.whatwg.org/multipage/dom.html#attr-data-*

BurningTreeC:
Hi @Jermolene, this PR would be ready for a review :)

Jermolene:
Hi @BurningTreeC this is looking very promising indeed, but I think we're going to have to wait for a subsequent release for this.

The first test I tried was to try a self-referencing edit-text widget. I added the following to the tiddler `TiddlyWiki Pre-release`:

```
<$edit-text tiddler="TiddlyWiki Pre-release" field="myfield"/>

Current: <$view tiddler="TiddlyWiki Pre-release" field="myfield"/>
```

As expected, I found that editing the field worked correctly, without the loss of focus that one finds on tiddlywiki.com/prerelease.

The next thing I tried was to run the following code in the developer console – it rewrites the tiddler `$:/core/ui/PageTemplate` every 2s.

```
var tick = function() {$tw.wiki.addTiddler(new $tw.Tiddler($tw.wiki.getTiddler("$:/core/ui/PageTemplate"),{myfield: (new Date()).toString()}))};window.setInterval(tick,2000)
```

I was expecting that I would be able to use the edit-text widget from the first trial without being disrupted by the refresh events, but I found that the focus is lost on each refresh event.

Am I right in expecting that to have worked?

BurningTreeC:
Hi @Jermolene, thanks for testing this

Your second test fails because the `autoFocus` of the sidebar-search field steals the focus of the input field in the `TiddlyWiki Prerelease` tiddler. It's designed to behave like this. Whenever a widget with `autoFocus="yes"` asks for focus, the focusing of the domNode that had focus before the refresh is disabled

BurningTreeC:
@Jermolene - edited the above post

Jermolene:
> Hi @Jermolene, thanks for testing this
> 
> Your second test fails because the `autoFocus` of the sidebar-search field steals the focus of the input field in the `TiddlyWiki Prerelease` tiddler. It's designed to behave like this. Whenever a widget with `autoFocus="yes"` asks for focus, the focusing of the domNode that had focus before the refresh is disabled
> 
> Edit: If you close the sidebar you can see that the mechanism works

Ah, excellent. It does indeed work with the sidebar hidden.

However, as soon as one types enough text to cause the text input to scroll horizontally then we lose the scroll position on every refresh. I think we have discussed extending the techniques here to be able to track and restore scroll positions. Have you had a chance to think about that any more?

As you've raised before, we also need to figure out what to do about `autofocus=yes` controls. Perhaps we should mangle the attribute to `data-tiddlwiki-autofocus=yes` so that the browser ignores it, and then we could craft a selector that we'd use to choose an autofocus element if we were unable to retain the focus over the refresh cycle.

BurningTreeC:
> > Hi @Jermolene, thanks for testing this
> > Your second test fails because the `autoFocus` of the sidebar-search field steals the focus of the input field in the `TiddlyWiki Prerelease` tiddler. It's designed to behave like this. Whenever a widget with `autoFocus="yes"` asks for focus, the focusing of the domNode that had focus before the refresh is disabled
> > Edit: If you close the sidebar you can see that the mechanism works
> 
> Ah, excellent. It does indeed work with the sidebar hidden.
> 
> However, as soon as one types enough text to cause the text input to scroll horizontally then we lose the scroll position on every refresh. I think we have discussed extending the techniques here to be able to track and restore scroll positions. Have you had a chance to think about that any more?
> 

Now I've thought about it a little bit and came up with a simple solution which I hope works out

> As you've raised before, we also need to figure out what to do about `autofocus=yes` controls. Perhaps we should mangle the attribute to `data-tiddlwiki-autofocus=yes` so that the browser ignores it, and then we could craft a selector that we'd use to choose an autofocus element if we were unable to retain the focus over the refresh cycle.

I don't understand your idea here... could you please elaborate a bit on it?

Jermolene:
> Now I've thought about it a little bit and came up with a simple solution which I hope works out

Excellent!

> > As you've raised before, we also need to figure out what to do about `autofocus=yes` controls. Perhaps we should mangle the attribute to `data-tiddlwiki-autofocus=yes` so that the browser ignores it, and then we could craft a selector that we'd use to choose an autofocus element if we were unable to retain the focus over the refresh cycle.
> 
> @Jermolene I don't understand your idea here... could you please elaborate a bit on it?

I was thinking about the problem with the sidebar. One way to avoid it would be to make sure that there are no input elements with `autofocus=yes` so that the focus doesn't get stolen. We could do that by rewriting the autofocus attribute to be something different that we can still target with `querySelector()`, and thus use as part of our fallback chain if there isn't an element to restore the focus to.

Jermolene:
By the way, would you mind putting this PR into draft mode for the moment? Thank you.

BurningTreeC:
> > Now I've thought about it a little bit and came up with a simple solution which I hope works out
> 
> Excellent!
> 
> > > As you've raised before, we also need to figure out what to do about `autofocus=yes` controls. Perhaps we should mangle the attribute to `data-tiddlwiki-autofocus=yes` so that the browser ignores it, and then we could craft a selector that we'd use to choose an autofocus element if we were unable to retain the focus over the refresh cycle.
> > 
> > 
> > @Jermolene I don't understand your idea here... could you please elaborate a bit on it?
> 
> I was thinking about the problem with the sidebar. One way to avoid it would be to make sure that there are no input elements with `autofocus=yes` so that the focus doesn't get stolen. We could do that by rewriting the autofocus attribute to be something different that we can still target with `querySelector()`, and thus use as part of our fallback chain if there isn't an element to restore the focus to.

Hi @Jermolene,
The current code (almost) always finds an element to restore the focus to
It also depends on the autofocus attribute on the title-input-fields of tiddlers
- If we click on the new-tiddler button, focus would get restored on the new-tiddler-button
- But with the autofocus attribute on the title-input-field of the new tiddler, focus gets set on that

Maybe the whole code needs a redesign with a different approach?

core/modules/editor/engines/framed.js
Jermolene: I concur that the escaping is needed; otherwise we wouldn't be able to locate tiddlers with a single quote in their title.

BurningTreeC:
Hi @Jermolene, now the focus on the input doesn't get stolen anymore by the input in the sidebar or any other input with autofocus="yes" when using your 2-second-page-refresh method in the console.
Also the scroll position gets preserved over refresh cycles

saqimtiaz:
@BurningTreeC are you planning a new PR instead

BurningTreeC:
Hi @saqimtiaz , yes I'm planning a new PR

BurningTreeC:
@saqimtiaz ... #5736 has solved some problems that I was also trying to solve here, but in a really messy way

PR #5699, opened 2021-05-18
Drag and drop images in the editor to import and insert

saqimtiaz:
This PR adds the ability to **import images by dragging and dropping them into the editor**. The imported image tiddler is transcluded at the caret position in the editor.

I still need to add documentation. However, **I thought it would be helpful to get some feedback on the UX and implementation before I start on that.** Please keep your feedback directly pertinent to this PR.

https://user-images.githubusercontent.com/68092/118411885-3296e980-b697-11eb-82ce-146c08679090.mp4

I've posted a demo here:
https://saqimtiaz.github.io/sq-tw/temp/images-import-in-editor.html

**Implementation details:**
- `<$edit-text>` extended with an optional `fileDrop` attribute which when set to` yes` allows files to be dropped on the editor for import. This defaults to `no`, maintaining backwards compatibility.
- `framed engine` updated to propagate drop events to the parent document from the iframe.
- `codemirror engine` updated to allow drop events to be passed through.
- `<$dropzone>` widget updated with a `filesOnly` attribute which when set only allows the widget to accept `Files` via a drop or paste event, and not string data. We use the default browser handling for string data so that normal drag and drop and pasting of text in the editor still works.
- A new config tiddler called `$:/config/Editor/EnableImportFilter` determines whether dragging and dropping files in the editor works for a given tiddler or not.
  - this filter is used in such a manner that it respects the global drag and drop setting.
- A new config tiddler called `$:/config/Editor/ImportContentTypesFilter` which is used to set the attribute `contentTypesFilter` for the `<$dropzone/>` widget and determines which content types can be imported by dropping on the editor.
- `$:/core/ui/EditTemplate/body` updated to wrap a `<$dropzone>` widget around the editor.
- `$:/core/ui/EditTemplate/body/editor` updated to introduce a new tag `$:/tags/EditorTools` for transcluding non-button UI elements inside the `<$edit-text>` widget.

**Known issues:**
- The popup on dropping a file is vertically centred with CSS in the editor. This works well for most tiddlers. However, with very long tiddlers the popup can appear offscreen. Not using a popup and having the UI appear at the top or the bottom of the editor isn't any better as it could still appear offscreen in longer tiddlers.
 - Eventually, we could position the popup vertically at the caret position. This would need for us to implement support for positioning elements are the caret position as discussed in #5585. One facet of this which needs further consideration is how to to provide access to the caret co-ordinates and relevant variable for actions that are not triggered by an `<$edit-text>` widget, but that need not be discussed here.


saqimtiaz:
@Jermolene whenever you have the opportunity, I would appreciate some feedback before I invest time in adding documentation.

Jermolene:
Thanks @saqimtiaz this looks great, I'll look forward to it.

pmario:
The image should have created and modified fields, so it shows up in the Recent tab. ... This was confusing for me.

Jermolene:
> The image should have created and modified fields, so it shows up in the Recent tab. ... This was confusing for me.

That's a separate issue that exists with the import mechanism; this PR repurposes the existing mechanism.

saqimtiaz:
Correct, this PR is just using the existing core mechanism for importing images. If we make changes to how we handle image imports to add created and modified fields, the behaviour will be inherited here.

saqimtiaz:
Documentation added.

This is good to go @Jermolene

Jermolene:
Thanks @saqimtiaz

PR #5703, opened 2021-05-19
improve German Edtion - add a lot more content

pmario:
This PR will add a lot more content to the German Edition when it is finished.


pmario:
Closing this PR in favour of "German translation - pr against master" #5773

PR #5707, opened 2021-05-20
New text operation insert-text

saqimtiaz:
This PR adds a new text operation `insert-text` that can be used to insert the literal text specified in the param `text` at the caret position. If there is a selection, it is replaced.

Currently the only way to do something similar is to use `replace-selection` which results in a selection after the replacement. This is not always desirable. 

The new text operation `insert-text` has been used in this PR for the new feature for importing images via dropping in the editor. Previously, this resulted in the inserted image transclusion text being selected, which isn't consistent with using the editor tool for inserting a tiddler link for example. Now this behaviour is consistent.

Documentation is included.

### DIFF
--- a/core/modules/editor/operations/text/insert-text.js
+++ b/core/modules/editor/operations/text/insert-text.js
@@ -1,0 +1,23 @@
+/*\
+title: $:/core/modules/editor/operations/text/insert-text.js
+type: application/javascript
+module-type: texteditoroperation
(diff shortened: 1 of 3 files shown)


saqimtiaz:
Updated some docs, this is good to go.

Jermolene:
Thanks @saqimtiaz

PR #5708, opened 2021-05-20
Add support for JSON-formatted tiddler store, and make it the default

Jermolene:
This PR updates the way that tiddlers are represented within a TiddlyWiki standalone HTML file to use JSON in a `<script>` tag.

Up until now, the tiddler store area has been formatted according to a scheme devised for TiddlyWiki 1.0 back in 2004/5. Although based on HTML, it is a highly non-standard usage, and tricky to work with. The approach here of storing JSON in `<script>` tags is a standard technique widely used and understood by web developers.

The main rationale for the change is that it allows us to consider dropping all restrictions on field naming (ie, it becomes possible to use upper case letters and arbitrary symbols including whitespace within a field name). But it also makes it much easier for external tooling to extract or construct TiddlyWiki files.

The specific changes included are:

* Update boot.js to read a new tiddler store area attribute to determine the format of a store area, and to add support for the JSON format
* Update the `<<jsontiddlers>>` macro to optionally escape the `<` character, which would otherwise be unsafe within a `<script>` block
* Extends the templates for saving the single file configuration to allow the store area to use a variable `storeAreaFormat` that determines the store area format

Further changes will be required before this can be merged:

* Add support for importing/deserialising the new format

To try out the changes, the following command will create both a JSON and DIV formatted wiki:

```
tiddlywiki editions/prerelease/ \
    --verbose \
    --render $:/core/save/all index-json.html text/plain \
    --render $:/core/save/all index-div.html text/plain "" storeAreaFormat div
```

However, there is a serious impediment to merging this PR. It turns out that the JSON format is noticeably larger than the old DIV-based format:

* index-json.html – 8,063,237 bytes
* index-div.html – 7,833,254 bytes

I think it's actually only a 2.9% increase, but put another way, a single change that increases the size of tiddlywiki.com by 229,983 bytes doesn't seem worthwhile.

I investigated "compressing" the JSON into a different shape (in particular making a dictionary of unique strings, and replacing entries by the index into the dictionary). But I found that the resulting size was always pretty much the same, I think because the content of the strings have a bigger impact than the punctuation that surrounds them

I also investigated encoding schemes to allow us to extend the character set used for field names with the existing DIV-based format, but found it impossible to work around the fact that HTML retrieves attribute names as lowercase, regardless of how they are written.


saqimtiaz:
@Jermolene interesting idea, I definitely see the benefits that using JSON for the storeArea would bring. 

As you are well aware, I tend to have strong feelings about file size from the perspective of users in bandwidth scarce areas. However at the same time, size creep is inevitable if we are to keeping making progress in terms of features and usability. So I try to always assess changes that impact file size from the perspective of a cost-benefit analysis. 

In this particular case, a 3% size increase for a relatively large wiki at 8MB actually doesn't seem like it would make much of a practical difference for those with bandwidth concerns. Add on the fact that thanks to @flibbles' wizardry (#5383) we have shaved off 300kb of empty.html for 5.1.24, and I think the 3% size increase might well be an acceptable cost for the benefits it would bring.

core/modules/macros/jsontiddlers.js
```diff
@@ -20,14 +20,19 @@ exports.name = "jsontiddlers";
...
 ];
 
 /*
 Run the macro
 */
-exports.run = function(filter,spaces) {
-	return this.wiki.getTiddlersAsJson(filter,$tw.utils.parseInt(spaces));
+exports.run = function(filter,spaces,escapeUnsafeScriptCharacters) {
+	var json = this.wiki.getTiddlersAsJson(filter,{
+		spaces: $tw.utils.parseInt(spaces),
```
pmario, line 32: With this `.parseInt(space)` ... you hardcode "spaces" into the JSON output. I think it would be better to allow `\t` here. See description below. The spec would allow `\t`.
According to the macro docs, the default is 4 spaces (which I can't see?). ... While 1 `\t` would do the same thing. Saving 3 characters. 

>JSON.stringify()
>
>space Optional
    A **_String_** or **_Number_** object that's used to insert white space into the output JSON string for readability purposes.
>
>    If this is a Number, it indicates the number of space characters to use as white space; this number is capped at 10 (if it is greater, the value is just 10). Values less than 1 indicate that no space should be used.
>
>    If this is a String, the string (or the first 10 characters of the string, if it's longer than that) is used as white space. If this parameter is not provided (or is null), no white space is used.
>
> from: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

In line 706 of `core/modules/wiki.js` I think, the code is more flexible. IMO would allow us to use a `\t` tab.

core/modules/wiki.js
```diff
@@ -696,21 +696,44 @@ exports.getTiddlerAsJson = function(title) {
 	}
 };
 
-exports.getTiddlersAsJson = function(filter,spaces) {
+/*
+Options:
+spaces: number of spaces
```
pmario, line 701: According to `JSON.stringify(value, replacer, space)` ... space can be a **number** or a **string**. IMO we should use this possibility to reduce file-size but keeping the JSON readable, if needed.

Jermolene:
> In this particular case, a 3% size increase for a relatively large wiki at 8MB actually doesn't seem like it would make much of a practical difference for those with bandwidth concerns. Add on the fact that thanks to @flibbles' wizardry (#5383) we have shaved off 300kb from empty.html for 5.1.24, and I think the 3% size increase might well be an acceptable cost for the benefits it would bring.

Indeed, I think the case is very compelling (although I think moving to JSON would actually sidestep the improvements in #5383).

core/modules/macros/jsontiddlers.js
Jermolene: Good point @pmario, but the handling of the spaces parameter is unchanged in this PR, and so I think the point might be better made in a separate issue.

> According to the macro docs, the default is 4 spaces

I think that is wrong, the default is not to use whitespace.

Jermolene: I mean I think *the documentation* is wrong.

core/modules/wiki.js
Jermolene: The current implementation doesn't include whitespace in the JSON representation.

pmario:
@saqimtiaz 

> As you are well aware, I tend to have strong feelings about file size from the perspective of users in bandwidth scarce areas. 

You should definitely have a closer look at this PR: https://github.com/Jermolene/TiddlyWiki5/pull/5570

It allows me to have a local WebDav setup, with a 2.2MByte core, and a 300kByte html-wiki. Where empty.html is about 90kByte or something like this. 

The big advantage is, that the core can easily be "cached" by the browser. So only the hmtl file needs to be transferred. ... If it has changed. 

I know, that it will introduce a bit more complexity imo it would be worth a try, if you have bandwidth problems. IMO saving a file with some 100kByte will be less errorprone as saving a 2.6 MByte file. 

If I activate server side compression, there will be about 550kByte for the core and 130kByte for the html file which needs to go over the wire.

saqimtiaz:
> Indeed, I think the case is very compelling (although I think moving to JSON would actually sidestep the improvements in #5383).

@Jermolene right but when we compare file sizes now, are we not comparing against the version with Flibbles' improvements? So the size increase is about 300kb less than we think compared to 5.1.23

```
empty.html version 5.1.23 is 				2,420,592 bytes
empty.html version 5.1.24 with div store area is 	2,289,414 bytes
empty.html version 5.1.24 with JSON store area is 	2,497,083 bytes
```

Jermolene:
> @Jermolene right but when we compare file sizes now, are we not comparing against the version with Flibbles' improvements? So the size increase is about 300kb less than we think compared to 5.1.23

Yes, I see what you mean, excellent point, that does rather get us off the hook.

saqimtiaz:
> I know, that it will introduce a bit more complexity imo it would be worth a try, if you have bandwidth problems. IMO saving a file with some 100kByte will be less errorprone as saving a 2.6 MByte file.

@pmario we shouldn't get into this here as its off-topic. In brief, that just trades one problem for another. When dealing with low (digital) literacy the single file model is actually key, rather than needing to juggle and remember multiple files when emailing, copying to a USB stick etc.

Jermolene:
I'd like to merge this over the next couple of days, so I'd welcome any thoughts about the specifics of the implementation. We also need to consider how to communicate this change, because there's a good chance that it will end up impacting many people/projects. For example, I believe that TiddlyHost splices additional tiddlers into the HTML file (is that correct @simonbaird?), and if so will be broken as soon as we merge this.

Some thoughts/questions about the implementation:

* The JSON structure is currently an array of tiddler field objects, with each field being represented as a string. That means that we can't easily extend the object structure to embed additional information. I think that's OK because in such cases one would just make another JSON store area
* We should double-check that there aren't any commonly used conventions that we could be using. For example, from a distance one might think it would be useful to adopt JSON-LD, but I'm pretty sure that it would make the overall file size explode
* Something I've been interested in for a long time is to make it possible to add tiddlers to a TiddlyWiki file just by appending text to the end. In my experiments, a `<script>` tag placed after the closing `</html>` tag is still accessible through the DOM. The advantage is that it would enable really simple tooling to make a functional TW5 file: just take the empty.html, and append the tiddlers as JSON in a script tag. If we're updating the spec, there's a case to be made for building this in now.

```
<!doctype html>
<html>
<head>
</head>
<body>
This page demonstrates that it is possible to access script tags placed after the closing
<code>~&lt;html></code> tag of an HTML document.
<script type="text/javascript">
	
document.addEventListener("DOMContentLoaded",function(event) {
	alert("I'm here: " + document.getElementById("the-end").textContent);
},false)
</script>
</body>
</html>
<script id="the-end" type="tiddlers">
This is the tiddlers
</script>
```

Some thoughts about managing the change:

* We should make a pinned post on all our forums pointing to a GitHub issue that clearly explains what has changed and what developers might need to do about it
* We should attempt to pro-actively reach out directly to anyone actively maintaining TiddlyWiki tools or services that we anticipate would be impacted by the change
* We need formal "/dev" documentation for both the old and new formats
* To make it easier for developers to work with both the old and new formats, I'd like to create a small, new, modern JavaScript library called something like `@tiddlywiki/utils` containing basic methods for manipulating TiddlyWiki files (eg converting to and from TW's date format, reading the tiddlers from a file, splicing tiddlers into an HTML file etc)

saqimtiaz:
> For example, I believe that TiddlyHost splices additional tiddlers into the HTML file (is that correct @simonbaird?), and if so will be broken as soon as we merge this.

Not only that, but TH also supports retrieving individual tiddlers via an API. That code may need re-visiting as well.

Do any of the mobile OS specific apps for TW inject tiddlers via splicing? Anyone that manipulates the HTML file directly will need to update their code to be handle the old store format as well the new one.

[TiddlyClip](https://github.com/buggyj/tiddlyclip) I suspect is another one that will be impacted.

> * Something I've been interested in for a long time is to make it possible to add tiddlers to a TiddlyWiki file just by appending text to the end. In my experiments, a `<script>` tag placed after the closing `</html>` tag is still accessible through the DOM.

We would need to make sure that this is true in all major browsers. If so, this seems like an excellent idea. Another thought to consider is the possibility of adding tiddlers to a non-empty file by appending an additional `<script>` tag at the end of the file.

> * To make it easier for developers to work with both the old and new formats, I'd like to create a small, new, modern JavaScript library called something like `@tiddlywiki/utils` containing basic methods for manipulating TiddlyWiki files (eg converting to and from TW's date format, reading the tiddlers from a file, splicing tiddlers into an HTML file etc)

This would be fantastic and we probably should make it available via npm to make it easier for others to incorporate into their workflows and tooling.

saqimtiaz:
@Jermolene do we have an approximate timeline in mind for a 5.1.24 release? 
That would help in terms of knowing how much time we have to manage this change.

Jermolene:
> @Jermolene do we have an approximate timeline in mind for a 5.1.24 release? 
> 
> That would help in terms of knowing how much time we have to manage this change.

It depends whether we wait for the static publishing work to complete. But there's still a lot of uncertainty in that work, and so I'm now thinking we should defer it for v5.1.25, and get v5.1.24 out in the next few weeks.

saqimtiaz:
@Jermolene it would be good to have a longer period where the static publishing code is in the pre-release and gets some more real world testing and time for proper consideration.

Jermolene:
> Do any of the mobile OS specific apps for TW inject tiddlers via splicing? Anyone that manipulates the HTML file directly will need to update their code to be handle the old store format as well the new one.

The ones I can think of are Quine and Tiddloid.

> [TiddlyClip](https://github.com/buggyj/tiddlyclip) I suspect is another one that will be impacted.

Definitely, that's @buggyj 

> We would need to make sure that this is true in all major browsers. If so, this seems like an excellent idea. Another thought to consider is the possibility of adding tiddlers to a non-empty file by appending an additional `<script>` tag at the end of the file.

Yes indeed, I was thinking that the boot kernel would scoop up all the script tags it can find, with the later ones overriding any tiddlers of the same title in earlier script tags.

> This would be fantastic and we probably should make it available via npm to make it easier for others to incorporate into their workflows and tooling.

Absolutely, and we should start using an npm namespace for that.

Perhaps we should also expose a compatible API from require("tiddlywiki") as well.

Jermolene:
> @Jermolene it would be good to have a longer period where the static publishing code is in the pre-release and gets some more real world testing and time for proper consideration.

Yes indeed.

Jermolene:
Hi @saqimtiaz a couple of thoughts:

* We could somewhat ameliorate backwards compatibility issues by retaining support for reading tiddlers from the existing DIV structure, and including the empty DIV in the templates. That way, tools that splice tiddlers into empty.html would continue to work
* I wonder if we should bump the version number to v5.2.0 for this release. Previously I've thought of bumping the version to v5.2.x to mark some future fundamental refactoring of the core, but actually, perhaps after all these years now is the right moment, since this is the first material change from an external perspective since TW5 came out of beta. I think it would help reduce the cognitive load for everyone

saqimtiaz:
> * We could somewhat ameliorate backwards compatibility issues by retaining support for reading tiddlers from the existing DIV structure, and including the empty DIV in the templates. That way, tools that splice tiddlers into empty.html would continue to work

Very good thinking and I completely agree. We can document that we recommend moving over to using the JSON store area format and the div store area is deprecated, but in the code we continue to support both as well as mixing and matching the two. We just have to make it clear what the order is for precedence.

> * I wonder if we should bump the version number to v5.2.0 for this release. Previously I've thought of bumping the version to v5.2.x to mark some future fundamental refactoring of the core, but actually, perhaps after all these years now is the right moment, since this is the first material change from an external perspective since TW5 came out of beta. I think it would help reduce the cognitive load for everyone

I think we have always been very conservative with version number bumps. from a user perspective with 5.1.23 vs 5.1.24 it can be hard to even notice that the version number has changed. I would have been amenable to bumping to 5.2.0 even for the previous release which while not a fundamental refactoring was noticeable just for the sheer scope of development between releases.

So I am very much for bumping the version number to 5.2.0, especially given the store area changes.

I recommend we create an issue mentioning that we plan to release 5.2.0 in 2-3 weeks so that any PR's already under discussion can be resolved. I can take the lead on that if you feel it would helpful.

saqimtiaz:
@Jermolene let me know if there are areas I can be of assistance with in implementing this change.

With regards to documentation we have a bit of a start at https://tiddlywiki.com/dev/#Data-Storage which I can flesh out with more details on both store area formats, as well as precedence when both are used together, once the code changes are final.

Jermolene:
> @Jermolene let me know if there are areas I can be of assistance with in implementing this change.
> 
> With regards to documentation we have a bit of a start at https://tiddlywiki.com/dev/#Data-Storage which I can flesh out with more details on both store area formats, as well as precedence when both are used together, once the code changes are final.

Thanks @saqimtiaz. I'm sure the code will be fairly straightforward, but help with the communications plan and the docs much welcome.

> I think we have always been very conservative with version number bumps. from a user perspective with 5.1.23 vs 5.1.24 it can be hard to even notice that the version number has changed. I would have been amenable to bumping to 5.2.0 even for the previous release which while not a fundamental refactoring was noticeable just for the sheer scope of development between releases.
> 
> So I am very much for bumping the version number to 5.2.0, especially given the store area changes.
> 
> I recommend we create an issue mentioning that we plan to release 5.2.0 in 2-3 weeks so that any PR's already under discussion can be resolved. I can take the lead on that if you feel it would helpful.

Great, thank you, please do.

saqimtiaz:
@Jermolene On Monday, I'll write up a post about the store area changes and run it by you before posting that to the Google Group. 

For the docs I'll probably wait until the code changes are in place before I work on them much.

Jermolene:
The commit 16ca0e9c6d567a6a435cdce17ad5ae12f1e755d0:

* Moves JSON tiddlers into their own element, so that they can be provided alongside old-style `<div id="storeArea">` tiddlers
* Uses classes rather than an ID to find the JSON store area(s). This allows us to append new script tags to the end of the file to add tiddlers
* When using the JSON format, adds a dummy old-style store area for backwards compatibility
* Fixes encrypted wiki handling

As @saqimtiaz notes, we should double check whether append tiddlers as a script tag works in all browsers, but that is actually only a documentation issue. Otherwise, I think the code might be done.

Jermolene:
> The commit [16ca0e9](https://github.com/Jermolene/TiddlyWiki5/commit/16ca0e9c6d567a6a435cdce17ad5ae12f1e755d0):
> 
> * Moves JSON tiddlers into their own element, so that they can be provided alongside old-style `<div id="storeArea">` tiddlers
> * Uses classes rather than an ID to find the JSON store area(s). This allows us to append new script tags to the end of the file to add tiddlers
> * When using the JSON format, adds a dummy old-style store area for backwards compatibility
> * Fixes encrypted wiki handling
> 
> As @saqimtiaz notes, we should double check whether append tiddlers as a script tag works in all browsers, but that is actually only a documentation issue. Otherwise, I think the code might be done.

Famous last words. Commit 73da37c refactors things to use the existing deserialiser framework to decode the content. This allows us to do things like this:

```
<script class="tiddlywiki-tiddler-store" type="application/x-tiddler">
title: HelloThereEveryone

This is a tiddler
</script>
```

saqimtiaz:
@Jermolene looks good. I'll do some cross browser testing and start working on docs, likely tomorrow.

One caveat to keep in mind is that while these changes should make it very easy to append tiddlers to a TiddlyWiki file, it does make it a bit more involved to parse a file to extract tiddlers. One would have to allow for both formats and multiple tiddler store areas.

Jermolene:
> One caveat to keep in mind is that while these changes should make it very easy to append tiddlers to a TiddlyWiki file, it does make it a bit more involved to parse a file to extract tiddlers. One would have to allow for both formats and multiple tiddler store areas.

I'm starting on that now, wish me luck :smile:

core/modules/deserializers.js
```diff
@@ -142,7 +91,19 @@ function deserializeHtmlFile(text,fields) {
 	return [result];
 }
 
-function deserializeTiddlyWikiFile(text,storeAreaEnd,isTiddlyWiki5,fields) {
+function deserializeNewStoreArea(text,storeAreaEnd,type,fields) {
```
saqimtiaz, line 94: Would something along the lines of `deserializeJSONStoreArea` be a more appropriate name? 

I am just postulating a future in which we introduce yet another store area format and then "NewStoreArea" becomes ambiguous.

Jermolene: From the perspective of the deserializer, none of this is JSON-specific: we're adding support for embedding blocks for any supported deserializers.

I thought of "scriptStoreArea", but the core already supports tiddlers in script tags in a slightly different format for modules.

saqimtiaz:
@Jermolene I'm planning to do some cross browser testing for appending a script tag at the end of the file, as well as starting on documentation. However, I won't be able to get a start on that until late tonight or tomorrow.

It slipped my mind that it is a public holiday here today, which comes with its share of commitments and activities with the household.

Jermolene:
> @Jermolene I'm planning to do some cross browser testing for appending a script tag at the end of the file, as well as starting on documentation. However, I won't be able to get a start on that until late tonight or tomorrow.

I'm fairly confident that the behaviour relied upon is part of the HTML spec, but I've been meaning to check that too.

> It slipped my mind that it is a public holiday here today, which comes with its share of commitments and activities with the household.

That's a typical post-COVID bank holiday for me too.

saqimtiaz:
@Jermolene it is still early days, but I have a made a start at writing documentation for the `storeArea` formats beginning with the old formats:

https://saqimtiaz.github.io/sq-tw/temp/dev.html

With regards to formal documentation, is there a specific format or an example or the kind of structure you have in mind?

Jermolene:
> @Jermolene it is still early days, but I have a made a start at writing documentation for the `storeArea` formats beginning with the old formats:
> 
> https://saqimtiaz.github.io/sq-tw/temp/dev.html
> 
> With regards to formal documentation, is there a specific format or an example or the kind of structure you have in mind?

Great, thanks @saqimtiaz.

I wondered if the samples have the whitespace removed to match how TW files are actually written? ie

```
<div id="storeArea" style="display:none;"><div created="20140611153703343" modified="20140611153734589" tags="testTag" testfield="testvalue" title="TestTiddler" type="text/plain"><pre>testText</pre>
</div>
</div>
```

The other thing I think we need is a pair of guides for how to extract tiddlers from a TiddlyWiki HTML file, and how to splice in new tiddlers. That's where we'd need to cover how strict the requirements are that `<div id="storeArea" style="display:none;">` precisely matches etc.

saqimtiaz:
I've added docs on the new store format and have tried to ensure that the examples use the exact same whitespace as the actual TW files.

https://saqimtiaz.github.io/sq-tw/temp/dev.html

I've added a mention of appending a `script` tag to insert tiddlers.

**We still need to add documentation on extracting tiddlers.** 
What requirements should there be? Is it enough to have a `script` tag with `class` "tiddlywiki-store-area" or does the tag precisely need to match the following:
`<script class="tiddlywiki-tiddler-store" type="application/json">`

I expect tomorrow to be busy but I can work on this some more later in the week.

Jermolene:
> I've added docs on the new store format and have tried to ensure that the examples use the exact same whitespace as the actual TW files.
> I've added a mention of appending a `script` tag to insert tiddlers.

Great, thanks

> **We still need to add documentation on extracting tiddlers.** 
> 
> What requirements should there be? Is it enough to have a `script` tag with `class` "tiddlywiki-store-area" or does the tag precisely need to match the following:
> 
> `<script class="tiddlywiki-tiddler-store" type="application/json">`
> 
> 
> 
> I expect tomorrow to be busy but I can work on this some more later in the week.

The deserialisation process is more brittle than the boot process, so we should focus on the requirements for the former. I suspect it will end up being sanitised pseudocode of what the deserializer does.

simonbaird:
To confirm, yes Tiddlyhost read and insert tiddlers, so it does need to understand how the store works. I don't think it will be too hard to support the new format, but there will be some work needed.

simonbaird:
My 2 cents on a few things:
* Version numbering. Would you consider one more stable 5.1.x release with the old store format, then a 5.2.0 release with it?
* Supporting the old div store as well as the new. Consider not doing that. It seems like it may cause subtle and weird bugs and inconsistencies and have limited value. If I was the owner of some kind of tooling that depended on the old div based format, I'd rather just declare 100% incompatibility with 5.2.x, versus a hybrid mismash of two store formats that maybe works depending on the use case.

Jermolene:
Thanks @simonbaird 

> * Version numbering. Would you consider one more stable 5.1.x release with the old store format, then a 5.2.0 release with it?

What would be the advantage of that approach? The end user benefits of the JSON store area are significant so I'd be inclined to release it as soon as practical.

Before we get to the release, we need to manage the timing of merging of this PR (and hence the moment at which the new store area appears on https://tiddlywiki.com/prerelease). At this point we're gathering information and feedback, with no immediate timetable for merging this.
 
> * Supporting the old div store as well as the new. Consider not doing that. It seems like it may cause subtle and weird bugs and inconsistencies and have limited value. If I was the owner of some kind of tooling that depended on the old div based format, I'd rather just declare 100% incompatibility with 5.2.x, versus a hybrid mismash of two store formats that maybe works depending on the use case.

Those are very reasonable concerns, and I'm sympathetic. Our original plan was a straightforward switch along those lines. The main advantage of retaining both store areas is that existing tooling that splices new tiddlers into an empty TiddlyWiki file will continue to work correctly without modification. When a user subsequently save the file it will use the new JSON store area, leaving an empty old-style DIV.

The problems would occur with tooling that reads tiddlers from a TiddlyWiki HTML file; they will see a new-style JSON file as being empty, rather than being an invalid TiddlyWiki file. I think it's debatable which failure mode is preferable in general.

It might be worth noting that TW5 files have always supported multiple store areas of different types (both in the sense of encrypted vs. unencrypted, but also because of the way that the boot kernel and other system components are packaged in their own store areas). Of course, common usage has hitherto been able to ignore those complexities.

pmario:
@Jermolene ... Would it be possible for the new JSON-format to incorporate a SHA-value, that is calculated of "translatable text" elements to make it easier for translators to recognise "overall" changes. 

If you import changes within the same language one can use the "diff-view", which is provided by the import-dialogue, to see what changed. ... BUT that's not possible, with different languages.

Eg: I did create a draft-PR https://github.com/Jermolene/TiddlyWiki5/pull/5703 which contains about 500 changed and newly created tiddlers with TW-content / documentation. 

As I found out, working with it, it is extremely hard to keep track of changes between the German-German and especially the German-English versions. ... 

It is more or less a "trial and error" game, which is overwhelming and therefore somewhat frustrating. ... For both ... the contributor, which wants to help and the maintainer which needs to create a PR.

## Proposal

There should be a `content-digest` SHA256 value, with the main purpose to **track text content changes**, created by "all fields" **except:** 

 - the SHA-256 field itself
 - `modified, modifier, created, creator` .. Since core shadows don't have those fields
 - `type` ... most core tiddlers are `text/vnd.tiddlywiki` but have an **empty** type-field 
 - other "digest-values" - may be?!
 - server generated fields. eg: `bag, revision, _is_skinny` and alike

---------

I think, this possibility should be already part of the initial version, since it may be hard to implement it later.

Jermolene:
Hi @pmario this sounds a bit more fundamental than a change to the store area format. If I understand correctly, the proposal is that the core maintains a new field on every tiddler that contains a hash of the values of most fields of the tiddler. I don't quite understand why the core needs to maintain that hash; if any tooling wanted it, surely it could just calculate the hash from the tiddler values.

Perhaps it would help to understand the problem that you're trying to solve. It really sounds like this is an entirely separate issue to the format of the store area. The test would be whether you'd still be able to implement your proposal with the existing store area, which I'm pretty sure is the case. In that case, please open a new issue.

pmario:
After sending the post, I did think about it again. IMO it will be enough, that we define, how such a SHA value will be calculated and we have the function in the core. ... Keeping track of the value can be done on the "translators-side". ... I'll raise a new issue

Jermolene:
> After sending the post, I did think about it again. IMO it will be enough, that we define, how such a SHA value will be calculated and we have the function in the core. ... Keeping track of the value can be done on the "translators-side". ... I'll raise a new issue

If you can, please raise the issue as an exploration of the problem, and not just a proposal of the solution.

simonbaird:
> Thanks @simonbaird
> 
> > * Version numbering. Would you consider one more stable 5.1.x release with the old store format, then a 5.2.0 release with it?
> 
> What would be the advantage of that approach? The end user benefits of the JSON store area are significant so I'd be inclined to release it as soon as practical.

I guess I'm just hoping for a stable release of TiddlyWiki that includes !5455. If that stable release includes a new store format, then I won't be able to use it for Tiddlyhost right away, which means Tiddlyhost will remain stuck on an unsupported pre-release TW build that doesn't even have a version number.

> 
> Before we get to the release, we need to manage the timing of merging of this PR (and hence the moment at which the new store area appears on https://tiddlywiki.com/prerelease). At this point we're gathering information and feedback, with no immediate timetable for merging this.
> 
> > * Supporting the old div store as well as the new. Consider not doing that. It seems like it may cause subtle and weird bugs and inconsistencies and have limited value. If I was the owner of some kind of tooling that depended on the old div based format, I'd rather just declare 100% incompatibility with 5.2.x, versus a hybrid mismash of two store formats that maybe works depending on the use case.
> 
> Those are very reasonable concerns, and I'm sympathetic. Our original plan was a straightforward switch along those lines. The main advantage of retaining both store areas is that existing tooling that splices new tiddlers into an empty TiddlyWiki file will continue to work correctly without modification. When a user subsequently save the file it will use the new JSON store area, leaving an empty old-style DIV.
> 
> The problems would occur with tooling that reads tiddlers from a TiddlyWiki HTML file; they will see a new-style JSON file as being empty, rather than being an invalid TiddlyWiki file. I think it's debatable which failure mode is preferable in general.
> 
> It might be worth noting that TW5 files have always supported multiple store areas of different types (both in the sense of encrypted vs. unencrypted, but also because of the way that the boot kernel and other system components are packaged in their own store areas). Of course, common usage has hitherto been able to ignore those complexities.

Yeah, that's reasonable. I don't have a strong point to argue if it looks like supporting both formats can be done without too much trouble. Would you continue to support it forever, or deprecate and then remove it later?

Bonus question: Would a tiddler defined in the old format storeArea div take precedence over the same tiddler in the new store, or vice versa?

Jermolene:
> I guess I'm just hoping for a stable release of TiddlyWiki that includes #5455. If that stable release includes a new store format, then I won't be able to use it for Tiddlyhost right away, which means Tiddlyhost will remain stuck on an unsupported pre-release TW build that doesn't even have a version number.

Aha, that makes perfect sense.

Would a sequence like this work:

1. We add this branch to the continuous integration process so that a build will appear at https://tiddlywiki.com/prerelease-json
2. Update TiddlyHost to understand the new store area format without exposing it in the menu of editions on TiddlyHost
3. Merge this PR so that the prerelease has the JSON store area
4. Give everyone a chance to check that things are working
5. Release v5.2.0

> Yeah, that's reasonable. I don't have a strong point to argue if it looks like supporting both formats can be done without too much trouble.

The powerful point you've raised is that there might be unexpected, hard to debug consequences when trying to interoperate between tooling. I'd welcome other views.

> Would you continue to support it forever, or deprecate and then remove it later?

Also a good question. If we add this partially backwards compatible hack then I think we'll be constrained to stick with it forever...

> Bonus question: Would a tiddler defined in the old format storeArea div take precedence over the same tiddler in the new store, or vice versa?

The JSON store area takes precedence with the current code:

https://github.com/Jermolene/TiddlyWiki5/blob/aa9f9316ca9c70a442d949cd4eb59437f90a89fa/boot/boot.js#L1749-L1757

saqimtiaz:
> The JSON store area takes precedence with the current code:

@Jermolene I assumed the opposite when [writing the documentation](https://saqimtiaz.github.io/sq-tw/temp/dev.html) despite having earlier seen the code, as I am not sure it makes sense otherwise. 

As I understand it, the reasoning behind keeping support for the older div store area is to ensure that tools that add tiddlers to the HTML file still work. Which means they need to be able to add new tiddlers or override existing tiddlers. However, if the JSON store area takes precedence, then a tiddler added via such an external tool to the old store area can never override an existing tiddler. In other words, using both store area formats is going to be tricky and wont always yield the results that one expects.

In light of this and what Simon points out about the burden of having to maintain this support for both store area formats indefinitely, I think we should go ahead and drop support for the older store area format. 

We are bumping the version to 5.2.0 and it is therefore reasonable to expect that external tools will need to be updated. To some extent external tools will still have to support both store area formats depending on the TW version number, but at least then we can avoid the complexity and associated potential for edge case issues that would arise from mixing the two in the same file.

simonbaird:
>
> Would a sequence like this work:
> 
> 1. We add this branch to the continuous integration process so that a build will appear at https://tiddlywiki.com/prerelease-json
> 2. Update TiddlyHost to understand the new store area format without exposing it in the menu of editions on TiddlyHost
> 3. Merge this PR so that the prerelease has the JSON store area
> 4. Give everyone a chance to check that things are working
> 5. Release v5.2.0
> 

I think you could skip 1. I'm pretty sure I can check out the head of this MR and create an empty with it.

Ps, I'll create my own Tiddlyhost issue for tracking 2.

Jermolene:
> @Jermolene I assumed the opposite when [writing the documentation](https://saqimtiaz.github.io/sq-tw/temp/dev.html) despite having earlier seen the code, as I am not sure it makes sense otherwise. 

The code isn't super clear: the selectors later in the array are processed later, and so tiddlers that they yield overwrite tiddlers read earlier.

The reason it is this way round is so that tiddlers added through a JSON store area will overwrite any tiddlers in the body of the store area, which is I think the expected behaviour if appending tiddlers to a file.

> As I understand it, the reasoning behind keeping support for the older div store area is to ensure that tools that add tiddlers to the HTML file still work. Which means they need to be able to add new tiddlers or override existing tiddlers. However, if the JSON store area takes precedence, then a tiddler added via such an external tool to the old store area can never override an existing tiddler. In other words, using both store area formats is going to be tricky and wont always yield the results that one expects.

The arrangement I was particularly thinking about was a tool that injects tiddlers into empty.html; subsequently saving such a wiki will move all the tiddlers into the JSON area where they would take precedence.

So, I guess we've clarified things: the proposed arrangement allows old-style tools to add tiddlers to an empty file, but old-style tools cannot overwrite tiddlers in the new-style JSON store area.

> In light of this and what Simon points out about the burden of having to maintain this support for both store area formats indefinitely, I think we should go ahead and drop support for the older store area format. 

It may be worth clarifying that we're going to have to maintain support for reading old-style store areas for ever, in any case. The question is whether we support hybrid files. If we do, the burden is that we need to document the precedence rules. But, as I say, we already have precedence rules that affect boot.js reading tiddlers at startup.

> We are bumping the version to 5.2.0 and it is therefore reasonable to expect that external tools will need to be updated. To some extent external tools will still have to support both store area formats depending on the TW version number, but at least then we can avoid the complexity and associated potential for edge case issues that would arise from mixing the two in the same file.

I'm still not sure. The semantics are not difficult compared to (say) the semantics of filesystemadaptor tiddler file naming, nor is this kind of case unprecedented (it's actually the way I originally implemented shadow tiddlers), and the tricky semantics only arise in the presence of old-style tooling.

saqimtiaz:
> So, I guess we've clarified things: the proposed arrangement allows old-style tools to add tiddlers to an empty file, but old-style tools cannot overwrite tiddlers in the new-style JSON store area.

I have updated the docs to reflect this: https://saqimtiaz.github.io/sq-tw/temp/dev.html

simonbaird:
Is there any chance we could throw some line breaks into the json? I'm seeing a 2 million char line which is quite hostile to text editors and browsers doing a "view source".

simonbaird:
This may be off topic here, but what's the motivation for putting all of core.js into a plugin? Not to get too nostalgic, but hacking TW files directly, and being able to read code using view source was one of the fun things about TW back in the old days.

joshuafontany:
> This may be off topic here, but what's the motivation for putting all of core.js into a plugin? Not to get too nostalgic, but hacking TW files directly, and being able to read code using view source was one of the fun things about TW back in the old days.

https://tiddlywiki.com/dev/#TiddlyWiki%20Core%20Application

:)

pmario:
>This may be off topic here, but what's the motivation for putting all of core.js into a plugin? Not to get too nostalgic, but hacking TW files directly, and being able to read code using view source was one of the fun things about TW back in the old days.

If you open the dev-tools: debugger, you can still explore the source, set breakpoints and single step. IMO the new behaviour is much better then it was with TWc. 

<details><summary>see the image</summary>

grafik

</details>

saqimtiaz:
@Jermolene shall we merge this now?

 TiddlyHost now has support for the new format.

Jermolene:
> Is there any chance we could throw some line breaks into the json? I'm seeing a 2 million char line which is quite hostile to text editors and browsers doing a "view source".

That does make a lot of sense.

I'll investigate if there's a reasonably clean way to splice newlines between each tiddler.

Alternatively, we could use "newline delimited JSON", where we take the block of text from the script tag and split it by newline, and then process each line as a separate JSON object. So, instead of:

```
<script class="tiddlywiki-tiddler-store" type="application/json">[{"title": "A","text": "One"},{"title": "B","text": "Two"}]</script>
```

We'd have:


```
<script class="tiddlywiki-tiddler-store" type="application/json+newlinedelimited">
{"title": "A","text": "One"}
{"title": "B","text": "Two"}
</script>
```

While that's more readable, I'd be concerned that people might be confused that it wasn't proper JSON. We'd certainly need to change the MIME type; I made up "application/json+newlinedelimited"

> This may be off topic here, but what's the motivation for putting all of core.js into a plugin? Not to get too nostalgic, but hacking TW files directly, and being able to read code using view source was one of the fun things about TW back in the old days.

In the very early days of TW5 the core was a series of ordinary script tags that were executed directly by the browser rather than using `eval()`. There's still some vestigial support for JS modules that get loaded in that way, and the boot kernel is still loaded in that way. I ended up moving away from that architecture as the plugin mechanism and the module mechanism evolved because things were simpler if we had the consistency of the boot kernel treating all the plugins and modules in the same way on all platforms, without treating the core as special.

But to your deeper point, I think it would be fair to say that a big change from TWC to TW5 was that from a development perspective I have moved from thinking of the single file configuration as the canonical artefact to treating it as secondary output of the Node.js configuration.

Hackability of the single file edition is still very important, but I think now it's more a matter of making it easier to write tooling to read/modify/write the format than to be able to manually hack it in an editor. But that's not to say that I don't sometimes miss the days when tiddlywiki.com was a single file of 5,000 lines that I would edit code with a very primitive text editor (and edit the content using TiddlyWiki itself, of course)...

Jermolene:
> @Jermolene shall we merge this now?
> 
> TiddlyHost now has support for the new format.

I'd like to investigate @simonbaird's comment about adding some whitespace to the JSON output, and there's a bunch of docs changes.

I'd like to also get the release note completely updated before we merge this – I think that @OdinJorna kindly made a PR?

pmario:
>While that's more readable, I'd be concerned that people might be confused that it wasn't proper JSON. We'd certainly need to change the MIME type; I made up "application/json+newlinedelimited"

According to the Grammar in the spec https://datatracker.ietf.org/doc/html/rfc8259#section-2 it is perfectly fine to use line-breaks as you describe it. 

-------

According to [IANA](https://www.iana.org/assignments/media-types/media-types.xhtml) this isn't a registered JSON type. IMO we can't add something to `application/json` without registering it. That's not valid.

pmario:
uuups I didn't see, that you removed some elements. .. So it's no proper JSON anymore. 

```
<script class="tiddlywiki-tiddler-store" type="application/json">[
{"title":"A","text":"One"},
{"title":"B","text":"Two"},
]</script>
```

Would it rally hurt to make it that way? ... I know it will need more chars for the linebreaks, but I did remove the whitespace after the colons. So in sum my version should be smaller. 

And if you copy paste the stuff into an editor, it will be simple to call a "beautify" macro, to make it readable

simonbaird:
I think it's better if it's still parseable as standard json, so +1 to pmario's format suggestion.

I don't see a nice easy way to do it though, unfortunately. https://stackoverflow.com/questions/38832629/json-stringify-custom-formatting suggests using regex replace. I suppose you could also json stringify each tiddler individually then join them together with ",\n".

(FWIW I think this could be something to consider later. It's more of a nice-to-have rather than a blocker for this PR.)

Jermolene:
Hi @pmario @simonbaird the problem is indeed that it's surprisingly hard to come up with a robust and performant way to insert the spaces.

I should have explained the newline delimited JSON is already a thing: it's commonly used for streaming JSON to and from APIs:

http://ndjson.org

One minor thing is that if we decide to persist with no whitespace in the JSON, then it might be helpful to introduce a newline before and after it, so at least copying the JSON into an external tool is a bit easier.

Jermolene:
> >All serialized data MUST use the UTF8 encoding.
> 
> 
> 
> I think, that's a problem. IMO the advantage of the new storage format is, that we would be able to use unicode like this 

As I understand it, UTF-8 can encode any Unicode character by using a combination of bytes, so I don't think that's a problem. I'm pretty sure the JSON spec allows for emoji etc. in strings

boot/boot.js
```diff
@@ -1739,17 +1744,23 @@ $tw.modules.define("$:/boot/tiddlerdeserializer/dom","tiddlerdeserializer",{
...
+		// IDs for old-style v5.1.x tiddler stores
+		"#libraryModules",
+		"#modules",
+		"#bootKernelPrefix",
+		"#bootKernel",
+		"#styleArea",
+		"#storeArea",
+		"#systemArea",
+		// Classes for new-style v5.2.x JSON tiddler stores
+		".tiddlywiki-tiddler-store"
```
simonbaird, line 1757: Just a thought: Would it be better to be more specific here, e.g. `"html>body>script.tiddlywiki-tiddler-store"`?

I'm imagining some fun behaviour if someone manages to add that class somewhere it's not expected.

(It's out of scope for this PR, but the same question could be asked about the the other selectors. E.g. maybe `html>body>div#storeArea` is better than `#storeArea`.)

pmario: As I did understand that, it should be possible to simply add more SCRIPT tags to an existing html file by appending it at the end of the file. .. So those scripts will be outside the main HTML tabs. 

eg: 

```
cat newStuff >> ./path/tiddlywiki.html
```

Where `newStuff` contains: 

```
<script class="tiddlywiki-tiddler-store" type="application/json">[
{"title":"A","text":"One"},
{"title":"B","text":"Two"}
]</script>
```

@Jermolene Did I get this wrong?

simonbaird: Hmm, that's an interesting use case. There might be some disadvantages related to the doc being malformed, but I don't have a strong opinion on it. FWIW Tiddlyhost inserts its extra script element after the existing one.

Jermolene: If the user arrange a raw markup section containing a tag with the magic classes/IDs, then things could go wrong, but otherwise things seem safe.

btheado:
> Is there any chance we could throw some line breaks into the json? I'm seeing a 2 million char line which is quite hostile to text editors and browsers doing a "view source".

Lack of line breaks will also impact line-based tools such as `diff`. Currently I'm using a single file wiki with the webdav saver. Every time the file changes (detected using the file watching utility: http://eradman.com/entrproject/), I run `git commit` to keep a history of the changes. I find it nice that the `git diff` output is easy to read. I don't refer to the diffs very often, but it is a nice feature to have.

It will likely also have impact on the github saver. Similar to my use case, the diffs generated by the github saver checkins curretnly are very readable. Here is one for example: https://github.com/kookma/TW5-GitHub-Saver/commit/9bc60ac566d0ba6caed60124d635343d4822f1ad

I just wanted to point out this extra use case.

simonbaird:
Re inserting line breaks, see https://github.com/Jermolene/TiddlyWiki5/pull/5786 . Seems to work for me.

core/modules/wiki.js
```diff
@@ -700,21 +700,44 @@ exports.getTiddlerAsJson = function(title) {
...
+		function escapeUnsafeChars(unsafeChar) {
+			return {
+				"<":      "\\u003C",
+				">":      "\\u003E",
+				"/":      "\\u002F",
+				"\u2028": "\\u2028",
+				"\u2029": "\\u2029"
+			}[unsafeChar];
+		}
+		json = json.replace(/</g,escapeUnsafeChars);
```
simonbaird, line 738: It looks like only the < char is replaced. Did you intend to replace the other chars listed in the escapeUnsafeChars function as well?

Maybe like this?

    json.repace(/<|>|\/|\\u2028|\\u2029/,escapeUnsafeChars);

simonbaird: If not, then what's the point of listing them in the function?

Jermolene: Hi @simonbaird apologies, I should have clarified: I was/am a bit uncertain whether just escaping `<` is sufficient, and so implemented multiple character replacement, and then ran cross browser tests. Once we confirm that `<` is sufficient, then I'll clean up the code.

Jermolene:
> https://saqimtiaz.github.io/sq-tw/temp/dev.html

Thanks for that @saqimtiaz and apologies for the late reply.

I think we need to include the old and new encoding rules:

* Pre-v5.2.0, attribute values and tiddler text were both HTML encoded
* Post-v5.20, JSON store areas must escape the `<` character to `\u003C`

I think you may have already planned this, but it might be helpful to make tiddlers called "How to create a TiddlyWiki-compatible HTML file" and "How to read a TiddlyWiki-compatible HTML file" (or something) that walk through the process and pull out the gotchas (eg encoding).

A minor style point is that we prefix version numbers with a lowercase "v". There's also a typo "JavaScript Crypto Libary."

Jermolene:
> I just wanted to point out this extra use case.

Excellent point, thanks @btheado. #5786 from @simonbaird is now merged, so we now have each tiddler on its own line.

core/modules/wiki.js
simonbaird: Ack, thanks.

btheado:
> Excellent point, thanks @btheado. #5786 from @simonbaird is now merged, so we now have each tiddler on its own line.

Thanks @Jermolene and @simonbaird. I confirmed the diff output looks much better now. Changes inside of long tiddlers are still hard to see now that they are all on a single line. For that I installed https://github.com/mookid/diffr into my git  and I'm satisfied with the output. This tool gives word based highlighting.

twMat:
On topic but arguably out of scope for the PR... but I'm curious:
I predict this will be very useful to directly connect fields to (other) tiddlers, i.e by naming the field such. ...And such use will probably mean people will want to *linkify* the displayed field name, so they can access the corresponding tiddler. That doable?
Feel free to disregard this q if it is distracting in this thread.

Jermolene:
Hi @btheado

> Thanks @Jermolene and @simonbaird. I confirmed the diff output looks much better now. Changes inside of long tiddlers are still hard to see now that they are all on a single line. For that I installed https://github.com/mookid/diffr into my git  and I'm satisfied with the output. This tool gives word based highlighting.

Great, glad it helps. The fact that newlines in JSON strings are not represented as newlines is a big constraint on readability. It's a tricky balance; if the only benefit was the JSONness of the format then it wouldn't be worth such a big change: the driver is really the relaxing of field name constraints.

I'm interested to hear views on this before it's too late.

Jermolene:
> On topic but arguably out of scope for the PR... but I'm curious:
> 
> I predict this will be very useful to directly connect fields to (other) tiddlers, i.e by naming the field such. ...And such use will probably mean people will want to *linkify* the displayed field name, so they can access the corresponding tiddler. That doable?
> 
> Feel free to disregard this q if it is distracting in this thread.

It should be as easy as:

```
<$link to=<<field-name>>>
blah
</$link>
```

But I don't think the core would ever linkify fieldnames by default because it is not a universal usage. (To be fair, we might linkify fieldnames to bring up info about the field, but not to bring up a tiddler with the same name as the field).

saqimtiaz:
> > https://saqimtiaz.github.io/sq-tw/temp/dev.html

> I think we need to include the old and new encoding rules:
> 
> * Pre-v5.2.0, attribute values and tiddler text were both HTML encoded
> * Post-v5.20, JSON store areas must escape the `<` character to `\u003C`
> A minor style point is that we prefix version numbers with a lowercase "v". There's also a typo "JavaScript Crypto Libary."

@Jermolene I have addressed these concerns. I have also added information about the line breaks in the JSON and have updated the code snippets to match.

> I think you may have already planned this, but it might be helpful to make tiddlers called "How to create a TiddlyWiki-compatible HTML file" and "How to read a TiddlyWiki-compatible HTML file" (or something) that walk through the process and pull out the gotchas (eg encoding).

I was planning to add a tiddler covering how to extract content from a TiddlyWiki file. I think inserting content is now pretty easy using the described method of appending a `script` tag. Note however that I have come down with what is hopefully only  a cold, so I expect I wont be very productive this week.

Contributions to the documentation from others are very welcome.

Jermolene:
> @Jermolene I have addressed these concerns. I have also added information about the line breaks in the JSON and have updated the code snippets to match.

Thank you @saqimtiaz , looks good.

> I was planning to add a tiddler covering how to extract content from a TiddlyWiki file. I think inserting content is now pretty easy using the described method of appending a `script` tag. Note however that I have come down with what is hopefully only a cold, so I expect I wont be very productive this week.
> 
> Contributions to the documentation from others are very welcome.

Ouch! I hope you feel better swiftly.

saqimtiaz:
@Jermolene I may have thought of an issue with the relaxing of field name constraints while contemplating an idea for a new node.js file format for tiddlers. We haven't accounted for the fact that `.tid` files cannot deal with `:` characters in field names. The first `:` character is interpreted as the delimiter between field and value.

Can we encode this character in field names before saving? Otherwise, we already check tiddlers for fields with multi-line content and in their presence, save the tiddler as `.json`. So that could be one route to take:

https://github.com/Jermolene/TiddlyWiki5/blob/82ec63e71118f9ee788972371f713579bae67530/core/modules/utils/filesystem.js#L224-L237

Forcing **all** tiddlers to be saved as `.json` on node.js would be a big step backwards usability wise for many users, myself included. I would consider that a deal breaker for most of my personal TW usage, so I hope we can avoid going that route.

Jermolene:
> @Jermolene I may have thought of an issue with the relaxing of field name constraints while contemplating an idea for a new node.js file format for tiddlers. We haven't accounted for the fact that `.tid` files cannot deal with `:` characters in field names. The first `:` character is interpreted as the delimiter between field and value.
> 
> 
> 
> Can we encode this character in field names before saving? Otherwise, we already check tiddlers for fields with multi-line content and in their presence, save the tiddler as `.json`. So that could be one route to take:
> 
> 
> 
> https://github.com/Jermolene/TiddlyWiki5/blob/82ec63e71118f9ee788972371f713579bae67530/core/modules/utils/filesystem.js#L224-L237
> 
> 
> 
> Forcing **all** tiddlers to be saved as `.json` on node.js would be a big step backwards usability wise for many users, myself included. I would consider that a deal breaker for most of my personal TW usage, so I hope we can avoid going that route.

I'm definitely in favour of updating/extending the .tid format to work around the current limitations (ie field values not containing newlines, and field names not containing colons).

As you note, if one at present creates a multiline field value then TW will automatically save the tiddler in .json format instead of .tid. We may want to note somewhere that using colons in field names will also result in switching to .json format, but I hadn't expected that to be a problem.

It looks like the test for unsafe field names in $:/core/modules/utils/filesystem.js will need updating (at present it just checks for characters 0x00-0x1F to catch all the various newlines and tabs).

saqimtiaz:
> It looks like the test for unsafe field names in $:/core/modules/utils/filesystem.js will need updating (at present it just checks for characters 0x00-0x1F to catch all the various newlines and tabs).

We may wish to break this out into a utils method and also expose it via a filter operator. See #5809

Jermolene:
> We may wish to break this out into a utils method and also expose it via a filter operator. See #5809

For the moment I made a quick fix on master (even though users are prevented from interactively creating tiddlers with field names containing colons it can still be done via `<$action-createtiddler>`).

Jermolene:
I think this is now ready to merge, so I've taken it out of "draft" status. I'd be grateful for any last feedback; I'll merge in a few hours if all goes well.

Many thanks to @saqimtiaz @pmario @simonbaird @btheado @twMat for your with this.

saqimtiaz:
> > We may wish to break this out into a utils method and also expose it via a filter operator. See #5809
> 
> For the moment I made a quick fix on master (even though users are prevented from interactively creating tiddlers with field names containing colons it can still be done via `<$action-createtiddler>`).

I did not check this out and test it again but I believe we have removed that check and warning unless I am mistaken. Shouldn't be an issue though as it will just trigger saving as `.json` instead of `.tid`.

Otherwise this looks good to me.

pmario:
@Jermolene ... There seems to be something wrong with the JSON format. ... There is a lot of whitespace, which imo shouldn't be there. 

grafik

All of them contain spaces. I think the serialisation went wrong. I did use  `tiddlywiki .\editions\tw5.com\ --build` to build empt.html.

pmario:
It also seems the line-breaks as discussed at: https://github.com/Jermolene/TiddlyWiki5/pull/5708#issuecomment-857731442 haven't been implemented. So the whole storage area is a superlong string.

pmario:
It seems the plugin tiddlers have this problem in their "text":"{   the problem is here   }" area

simonbaird:
> It seems the plugin tiddlers have this problem in their "text":"{ the problem is here }" area

I think some formatting is expected there since the json content of the plugin tiddler does have line breaks and indenting. (Take a look at https://tiddlywiki.com/#%24%3A%2Fcore and click Edit for example.)

Perhaps some space could be saved by using a single space indent instead of a double space, but I don't that is specifically related to this PR.

pmario:
>Do you want to remove the trailing blank lines? (Not important.)

No the spaces. They are not needed and only bloat. For empty.html it's about 110kByte for tiddlywiki.hmtl with all plugins it will probably be 500kByte++ 

If we want to keep indentation we may add tabs instead of 4 chars per indent level, which should be super simple.

pmario:
>Perhaps some bytes could be saved by using a single space indent instead of a double space, but I don't think that is specifically related to this PR.

I think it is related. At the very beginning we did talk about the "size problems" that TW already has. .. If we want indentation in plugin tiddler source we should use "tabs" and save 3/4 of the chars that are needed atm.

Jermolene:
> @Jermolene ... There seems to be something wrong with the JSON format. ... There is a lot of whitespace, which imo shouldn't be there.

That's an image of the whitespace in the plugin tiddler `$:/core`. This PR doesn't change the format of the text field of plugins, it only changes how tiddlers are serialised into the HTML file.

You can see the same whitespace on tiddlywiki.com today:

image

pmario:
Yes, but we can save a lot of size, if we replace 4 chars with 1 tab. **I'll crate a new PR for this**. Since all plugins seem to be affected we can remove a lot of bloat, if they are encoded with tabs.

Jermolene:
> Yes, but we can save a lot of size, if we replace 4 chars with 1 tab. **I'll crate a new PR for this**. Since all plugins seem to be affected we can remove a lot of bloat, if they are encoded with tabs. 

Rather than replacing the spaces with tabs, perhaps better to just generate the JSON without any whitespace; there's no real benefit to it that I can see.

Jermolene:
Thanks @pmario

pmario:
There is something strange going on with the command `tiddlywiki .\editions\tw5.com-server\ --listen` with this branch. 

 - Start the server as described above
 - go to http://localhost:8080
 - F12   select the "Console" tab

There are 2000+ messages created. 

grafik

The network tab also shows a lot of communication. There seems to be something wrong with the "isDirty()" detection

grafik

Jermolene:
> There is something strange going on with the command `tiddlywiki .\editions\tw5.com-server\ --listen` with this branch. 

Ouch! Thank you @pmario, I'm investigating now.

core/templates/store.area.template.html.tid
```diff
@@ -1,14 +1,30 @@
...
-`</pre>`
-</$reveal>
\ No newline at end of file
+\whitespace trim
+<!-- Unencrypted -->
+<$list filter="[[$:/isEncrypted]get[text]else[no]match[no]]">
+    <$list filter="[[storeAreaFormat]is[variable]getvariable[]else[json]match[json]]">
+        <!-- New-style JSON store area, with an old-style store area for compatibility with v5.1.x tooling -->
+        `<script class="tiddlywiki-tiddler-store" type="application/json">`
+            <$macrocall $name="jsontiddlers" filter=<<saveTiddlerFilter>> spaces="-1" escapeUnsafeScriptCharacters="yes" $output="text/raw"/>
```
pmario, line 9: @Jermolene The `spaces="-1"` is very hacky. It causes problems, when I want to fix the "parseInt(spaces)" problem in the "jsontiddlers" macro. ... 

`JSON.stringify(value, replacer, space)` "space" can be a number OR a string eg: "\t"

`spaces` is used to transfer a completely unrelated information to `exports.getTiddlersAsJson = function(filter,options)` which already has an `option` that can be easily used to hold a new parameter. So imo there is no need to "misuse spaces" .. 

I would suggest, that the `jsontiddlers` macro gets a new parameter eg: `linebreakBetweenTiddlers`

Jermolene: Hi @pmario at one point I did have a separate parameter for "linebreakBetweekTiddlers", but the problem is that it is not independent of the "spaces" setting; it doesn't make sense to have "spaces=3" and "linebreakBetweekTiddlers=true". In any event, can't "spaces=\t" coexist with "spaces=-1"?

Jermolene: I mean apart from the parseInt() problem that you referred to

pmario: Calling the macro looks like this atm

```
<$macrocall $name="jsontiddlers" filter=<<exportFilter>> spaces={{{[charcode[9]]}}} $output="text/raw"/>
```

I need to do this in "jsontiddlers" which is (kind of) ugly. ... spaces can be any string 

```
exports.run = function(filter,spaces,escapeUnsafeScriptCharacters) {
	var jsonString = (typeof spaces === "string") ? spaces : null,
	    jsonString = (spaces === "-1") ? null : jsonString,
	    json = this.wiki.getTiddlersAsJson(filter,{
	    spaces: (jsonString) ? jsonString : $tw.utils.parseInt(spaces),
```

Jermolene: Hi @pmario why not just do this and let `this.wiki.getTiddlersAsJson` resolve the overloading of the spaces parameter?

```
exports.run = function(filter,spaces,escapeUnsafeScriptCharacters) {
	var json = this.wiki.getTiddlersAsJson(filter,{
		spaces: spaces,
		escapeUnsafeScriptCharacters: escapeUnsafeScriptCharacters === "yes"
	});
	return json;
};
```

pmario: uups the code doesn't work. ... So I do need a new parameter to get it going. .. What if the `linebreakBetweekTiddlers` option overwrites the "spaces" parameter. .. We do have many options, that take precedence over other params.

pmario:
I do have the size differences now.

 - empty.html built with 4 spaces (current default) - `\t` - no indent

```
                                                         diff 
4chars ... empty.html ... 2,27 MB (2 381 047 Bytes)   
tab ...... empty.html ... 2,22 MB (2 328 932 Bytes)    53_115
0 ........ empty.html ... 2,14 MB (2 248 642 Bytes)   132_405
```

 - The biggest win is, if we completely strip the whitespace form plugin tiddlers. .. The **plugin code will keep** the indentation

 - The win for tw5.com index.html is a little bit more than empty.html, since it only contains 6 plugins. 

 - The full.html edition, which contains a lot of plugins, languages and themes. 

```
4chars ... index.html ... 9,81 MB (10 296 275 Bytes) 
tab ...... index.html ... 9,39 MB (9 848 332 Bytes)     447_943
0 ........ index.html ... 8,73 MB (9 154 781 Bytes)   1_141_494 
```

So the average size improvement, if we use 0 indent for the new store, will be about 200+ kByte, for users with some plugins and may be a language.

Jermolene:
> This turns out to be the tip of a bigger iceberg.

Lazy loading should be fixed in e2ce4e11e8b4a7e7d0a096d46e247b6f55d1bbf3. I'd be grateful for any further testing

@pmario note that the new code no longer uses the jsontiddlers macro, so feel free to make a separate PR with the fixes you mentioned.

Jermolene:
> I do have the size differences now.

I've fixed this in 70e60cd93

pmario:
>> I do have the size differences now.
>
>I've fixed this in 70e60cd

For empty.html it's about 5.9% saving. ... 140kByte is a new 3rd party library + TW integration .. IMO that's not bad. 

There are a lot of other places, where JSON.stringify() has potential for improvements. Once this PR is merged I'll create a new PR.

simonbaird:
FYI the "Fix templates for lazy loading" adds a lot of inconsitent indenting, i.e. a mix of tabs and spaces.

simonbaird:
In `core/modules/widgets/jsontiddler.js`, would this work the same (and save a few bytes?)

````
-           json = json.replace(/</g,"\\u003C");
+           json = json.replace(/</g,"\\x3C");
````

**Comment by simonbaird** [comment]
> In `core/modules/widgets/jsontiddler.js`, would this work the same (and save a few bytes?)
> 
> ```
> -           json = json.replace(/</g,"\\u003C");
> +           json = json.replace(/</g,"\\x3C");
> ```

See https://github.com/Jermolene/TiddlyWiki5/pull/5840 if you like that idea.

**Comment by Jermolene** [comment]
> In `core/modules/widgets/jsontiddler.js`, would this work the same (and save a few bytes?)
> 
> 
> 
> ````
> 
> -           json = json.replace(/</g,"\\u003C");
> 
> +           json = json.replace(/</g,"\\x3C");
> 
> ````

Hi @simonbaird annoyingly, but the JSON spec only accepts the  `\u` formulation (see https://json.org/).

**Comment by pmario** [comment]
Hi Jeremy, 
Finally q:=)   ... Will you post some info in the GG, so users will test it with their configurations, so we get more feedback?

**Comment by Jermolene** [comment]
> Hi Jeremy,
> Finally, congratulation q:=) ... Will you post some info in the GG, so users will test it with their configurations, so we get more feedback?

Yes! Will press "send" shortly.

File: editions/test/tiddlers/tests/test-deserializers.js [hunk 7]
```diff
@@ -0,0 +1,39 @@
...
+		function executeTestCase(title,expectedOutput) {
+			it("test case " + title, function() {
+				var tiddler = $tw.wiki.getTiddler(title);
+				expect($tw.wiki.deserializeTiddlers(tiddler.fields.type,tiddler.fields.text)).toEqual(expectedOutput);
+			});
+		}
+	
+		executeTestCase("dezerializer test data case 1",[ { text: '<!doctype html>\n', type: 'text/html' } ]);
+
+		executeTestCase("dezerializer test data case 2",[ { text: '<!doctype html>\n<html lang="en">\n<head>\n\t<meta charset="utf-8">\n\t<title>Test Data</title>\n</head>\n<body>\n</body>\n</html>\n', type: 'text/html' } ]);
```
saqimtiaz, line 28: @Jermolene this and the above test are failing for me on the Windows 10 machine I am working on while away from home. My first guess is that it might be due to newline encoding?

@pmario could you perhaps confirm if you have the same problem? If it is something specific to my environment then it can be safely ignored.

image

Jermolene: > @Jermolene this and the above test are failing for me on the Windows 10 machine I am working on while away from home. My first guess is that it might be due to newline encoding?

That does sound very likely, might you have Git configured to convert LF to CRLF when it pulls?

saqimtiaz: @Jermolene that is probably it, thank you. I am so used to a *nix based dev environment I hadn't even realized that git converting line endings was a thing! The travails of working in an unfamiliar OS.

saqimtiaz:
I've run into an issue trying to append a script tag with a tiddler to the end of the pre-release empty.html and I would like to rule that this is specific to my environment. Odd because this was tested earlier while this PR was in progress and I recall clearly that it did work then.

@Jermolene  @pmario I would appreciate it someone could open this wiki and confirm that it does not contain the tiddler with title `Hello "There"`:
https://saqimtiaz.github.io/sq-tw/temp/empty-with-extra-script-store.html

I have created this file by downloading the current prerelease empty.html and appending this at the end of the file:
`<script class="tiddlywiki-tiddler-store" type="application/json">[{"title":"Hello \"There\"","text":"Protractor"}]</script>`

If I move this same tag to within `<head>` then the tiddler is successfully loaded from it. Moving the <script> tag just before the closing `</html>` is not working either.

pmario:
@saqimtiaz ... I can verify, that the wiki doesn't show the tiddler. .. _But_ it is shown in the DOM structure if I use F12

grafik

So FF 92 on Windows 10  picks it up and places it inside the BODY element as the last element.

Jermolene:
Thanks @saqimtiaz @pmario good catch.

The problem isn't that the script tag is hoisted into the body, it's that the boot kernel runs before the browser has parsed the HTML beyond that point, and so cannot load tiddlers from DOM elements that come later in the document. I don't know how this got through my testing.

A related problem that's been on my mind is that while it's easy enough for TiddlyWiki to process multiple tiddler store blocks, it's asking a lot of third party tools that want to read TiddlyWiki files to duplicate those semantics.

So, I think it makes sense to change from `script.tiddlywiki-tiddler-store` to `script#tiddlywiki-tiddler-store`, thus allowing only a single tiddler store in a document. I'll prepare a commit.

pmario:
>So, I think it makes sense to change from script.tiddlywiki-tiddler-store to script#tiddlywiki-tiddler-store, thus allowing only a single tiddler store in a document. I'll prepare a commit.

I think we should discuss this a bit more. There is a big advantage in allowing several "stores"

saqimtiaz:
@Jermolene yes I can confirm that extra script tag stores inserted before the boot kernel are loaded properly. 

I too think that there is a significant advantage in allowing several stores, especially in terms of how much easier it makes it to inject new tiddlers. Also, we would need another TiddlyHost update if we drop support for multiple store tags.

Jermolene:
> I too think that there is a significant advantage in allowing several stores, especially in terms of how much easier it makes it to inject new tiddlers. Also, we would need another TiddlyHost update if we drop support for multiple store tags.

The inconsistency at the moment is that if one imports your test file into tiddlywiki.com/prerelease, then the `Hello "There"` tiddler shows up, but that it doesn't when loading the file in a browser. That does seem problematic.

saqimtiaz:
For context, I discovered this issue while trying to inject tiddlers into a string representation of a TiddlyWiki file using wikitext. With a single store area I am not quire sure that would be possible.

Jermolene:
With a single store area one can still splice additional tiddlers onto the end of the array fairly easily. One doesn't have to worry about existing tiddlers with the same title because the later one in the array will win.

Jermolene:
> With a single store area one can still splice additional tiddlers onto the end of the array fairly easily. One doesn't have to worry about existing tiddlers with the same title because the later one in the array will win.

Just to add that the regexp is simplified by the fact that we escape any `<` characters in the JSON text.

Jermolene:
> Also, we would need another TiddlyHost update if we drop support for multiple store tags.

That's the killer argument for not changing things.

I noticed that the ability to have multiple tiddlywiki-tiddler-stores is currently undocumented, so perhaps all that needs to be done is to remove test case 5 (which tests a trailing script tag).

saqimtiaz:
> The inconsistency at the moment is that if one imports your test file into tiddlywiki.com/prerelease, then the `Hello "There"` tiddler shows up, but that it doesn't when loading the file in a browser. That does seem problematic.

Agreed. However it might be worth at least taking the time to consider alternatives. For example, additional script tag stores need to be adjacent to the original store? I realize that doesn't resolve the discrepancy with import, I am just thinking out loud.

> With a single store area one can still splice additional tiddlers onto the end of the array fairly easily. One doesn't have to worry about existing tiddlers with the same title because the later one in the array will win.

Trying to do so from wikitext feels brittle but perhaps I am just being too apprehensive.

> I noticed that the ability to have multiple tiddlywiki-tiddler-stores is currently undocumented

It is specifically documented in the section `Multiple store areas and precedence` at the end of this tiddler:
https://tiddlywiki.com/prerelease/dev/

pmario:
I did a short test and added the additional store at the beginning of the file. So instead of 

```
$ cat store.txt >> myWiki.html
```
I can do 

```
$ cat store.txt >> new.html
$ cat myWiki.html >> new.html
```

Which works as expected.

pmario:
>Which works as expected.

Except, that an existing tiddler in the wiki will win, if there is a naming conflict. ... Not sure if this is an advantage or not. ... May be depends on the usecase.

saqimtiaz:
If you add another script tag immediately after the default one, it will take precedence.

pmario:
>If you add another script tag immediately after the default one, it will take precedence.

Yea,
but I was thinking about a very simple CLI based system (similar to http://todotxt.org/) that reads the standard input, pipes it to a bash-script.
The bash-script creates a journal-like title eg. `2021.09.18-12:00:00-123` and 
uses the standard input as text
covers the whole thing with a "store" script tag and adds it to an "comment.txt" file. 
At some point a second script adds the store to the wiki and removes the file. 

Or something similar ;)

simonbaird:
> So, I think it makes sense to change from `script.tiddlywiki-tiddler-store` to `script#tiddlywiki-tiddler-store`, thus allowing only a single tiddler store in a document. I'll prepare a commit.

This will break Tiddlyhost's current support for 5.2. How about 'html > body > script.tiddlywiki-tiddler-store' ?

simonbaird:
FWIW the "add tiddlers by appending to the html file" seems like a weird idea to me. No one would expect that to work, so I don't see much value in supporting it.

saqimtiaz:
> How about 'html > body > script.tiddlywiki-tiddler-store' ?

That helps but isn't quite specific enough. Only script tag stores that are before the boot script in the HTML file will have their tiddlers loaded. We could address that through documentation.

PR #5711, opened 2021-05-21
Eventcatcher: bug fixes and new stopPropagation attribute

### LINKED ISSUE #5684: Eventcatcher: make it easier to trap events and do nothing
Currently if there are no actions defined for an event, the `<$eventcatcher>` widget [does not trap the event and allows it to propagate](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/eventcatcher.js#L87). This was presumably to guard against missing actions attributes.

However, if you just want to stop propagation for some events (and do not want to invoke actions) you have to use a dummy non blank action string. For example:

```
\define dummyActions() NULL

<$eventcatcher events="dragenter dragleave dragover" actions-dragenter=<<dummyActions>> actions-dragleave=<<dummyActions>> actions-dragover=<<dummyActions>> selector=".tc-editor-toolbar">
```

instead of something simpler like:

`<$eventcatcher events="dragenter dragleave dragover" selector=".tc-editor-toolbar">`

- Is it too late to change the current behaviour for when the actions are blank or not defined? 
  - If so can we try to make the above use case more elegant? 
- Perhaps an extra attribute that can be used to always trap events even if actions are not defined? `alwaysCatch="yes"` ?

An alternative approach - which might be useful in its own right - would be to fall back to an `actions` attribute (if present) if an event does not have a specific actions attribute defined for it. One would still need to define dummy non blank actions for the above use case, but the widget could be invoked more concisely:

```
\define dummyActions() NULL

<$eventcatcher events="dragenter dragleave dragover" actions=<<dummyActions>> selector=".tc-editor-toolbar">
```

saqimtiaz:
This PR addresses three issues in `<$eventcatcher>`:

1. IE bug fix: `<$eventcatcher>` causes a red error screen as soon as an event is trapped. This is because IE does not support the `matches()` method but has its own equivalent `msMatchesSelector()` method. A utils method `domMatchesSelector()` has been introduced.

2. FireFox bug fix: there is a quirk where some events (such as `dragenter` and `dragleave`) can be fired for text nodes (nodeType 3) which are not proper Elements and therefore checking their attributes throws an error. All events originating from text nodes are now treated as originating from their parentNode.

3. Introduced a new `stopPropagation` attribute which fixes #5684 and allows greater control over handling of event propagation. The 3 possible values are:
  - `onaction` (default) which stops event propagation if there are actions to invoke.
  - `never` - never stop event propagation
  - `always` - always stops event propagation even if no actions have been specified.

Documentation has also been updated.

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -22,6 +22,10 @@ exports.domContains = function(a,b) {
 		!!(a.compareDocumentPosition(b) & 16);
 };
 
+exports.domMatchesSelector = function(node,selector) {
+	return node.matches ? node.matches(selector) : node.msMatchesSelector(selector);
+};
+
 exports.removeChildren = function(node) {
 	while(node.hasChildNodes()) {
 		node.removeChild(node.firstChild);
--- a/core/modules/widgets/eventcatcher.js
+++ b/core/modules/widgets/eventcatcher.js
@@ -47,38 +47,46 @@ EventWidget.prototype.render = function(parent,nextSibling) {
 		domNode.addEventListener(type,function(event) {
 			var selector = self.getAttribute("selector"),
 				actions = self.getAttribute("actions-"+type),
+				stopPropagation = self.getAttribute("stopPropagation","onaction"),
 				selectedNode = event.target,
 				selectedNodeRect,
 				catcherNodeRect,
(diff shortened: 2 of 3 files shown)


Jermolene:
Thanks @saqimtiaz

PR #5726, opened 2021-05-27
Add "commentpragma" html style rule

pmario:
This PR changes `commentblock` ~~creates a new rule `commentpragma`~~ and adds a test. 
The rule can be used like this: 

```
<!-- This is a comment in the "pragma" area -->
\define xx()
some text
\end

<<xx>>
```

~~Having a completely new rule will allow us to "selectively" disable it with `\rules except / only commentpragma`
and in the control panel wikitext rules setting.~~

This PR shouldn't clash with existing wikitext tiddlers since the rule already exists for wikitext. It wasn't possible to use it together with pragmas ... now it is.


pmario:
@Jermelone @saqimtiaz ... If you think it can be used that way, I'll add the docs to the PR

pmario:
bump with test-version - [test-html-comment-pragma.zip](https://github.com/Jermolene/TiddlyWiki5/files/6569762/test-html-comment-pragma.zip)

saqimtiaz:
@pmario just to be clear, I wasn't against the proposal in #5287.  Wikitext changes are rarely going to be backwards compatible. However it is important to acknowledge changes that aren't and have a strategy for introducing them. Which could take the form of gathering such changes for a release like 5.3.0 that breaks wikitext backwards compatibility in some areas.

I like the idea of having the same comment syntax in pragma as well as the regular body of wikitext, otherwise the `\\` syntax is more appealing.

In terms of backwards compatibility this seems like a safer option. I can imagine some very edge cases where it would cause problems but I don't know realistic those are. Imagine the following text in a tiddler before and after your proposed change:

```
<!--this is my comment because I don't have any pragma in this tiddler-->
\define mymacro() 
here I have just written an example of a macro, which won't be treated as a macro since there is non-pragma text before it.
\end
```

In TW 5.1.23, there wont be a macro defined. With your changes, it would be. However this does seem like a rather unlikely edge case.

pmario:
>@pmario just to be clear, I wasn't against the proposal in #5287. 

I know, but you had a good argument, that you use `\\ some text \\` as a marker in line 1 of existing content tiddlers. ... 

Those elements would have been converted into "invisible comments". ... I have never thought about this case but I think it's a valid blocker. ... 

On the other hand, if someone uses `<!-- comment -->` in line 1 in an existing tiddler, the result would be the same. 

Except your edge-case example

```
<!--this is my comment because I don't have any pragma in this tiddler-->
\define mymacro() 
here I have just written an example of a macro, which won't be treated as a macro since there is non-pragma text before it.
\end
```

**I think we should make it clear in the docs**. ... I don't think, that an existing macro definition would look that way, because it wouldn't work AND it also produces an "ugly" output in 5.1.23 ... So I'm pretty sure we can neglect this one. 

If I would need to disable a macro it would look like this: 

```
<!-- temporarily idsabled 
\define mymacro() 
here I have just written an example of a macro, which won't be treated as a macro since there is non-pragma text before it.
\end
-->
```

pmario:
For me it's important to **have** the possibility to add comments in the pragma area. I don't care too much, how those comments look like. .. I'm OK with an `<!-- html comment -->` because it is only visible in edit mode or if the content is shown as `<code>` ... 

The biggest problem with "macro comments" atm is, that there aren't any, because they produce a performance hit. ... Especially if macros are called recursively. ... 

Every time a macro is "executed" a new Parser() instance is created, which does parse the comment again. ... ATM the macro parse-tree isn't cached, so this is a problem. ... at least for me.

IMO there are 2 possibilities to fix this.

 1. Integrate this PR as "step 1"
 2. Make macro parse-tree cacheable

I think this PR is a "low hanging fruit". ... 

Making the macro parse-tree cacheable in the future will also improve this code, which isn't the case with this PR

```
\define test()
<!-- html comment here -->
text text text
\end
```

pmario:
This PR will also make something like this possible, which doesn't work with 5.1.23

```
\define test()
<!-- html comment here -->
\whitespace trim
text text text
\end

<<test>>
```

saqimtiaz:
> Making the macro parse-tree cacheable in the future will also improve this code, which isn't the case with this PR

As an aside, we do have some caching of macro parse trees as introduced in #5205.

pmario:
>As an aside, we do have some caching of macro parse trees as introduced in #5205.

You are right. I did forget about it, even if I did comment there. ... That's better, because it will make pragma-comments from this PR possible and faster too ;)

Jermolene:
I'm happy with the idea, but I wonder if the implementation can be improved: it should be possible to have a single parse rule that is both an inline parse rule and a pragma parse rule. I appreciate that having two rules enables them to be independently enabled, but I don't think that that is particularly useful.

```
exports.types = {block: true, pragma: true};
```

We don't have any such parse rules at the moment, but it would be worth investigating whether it can be made to work.

pmario:
>We don't have any such parse rules at the moment, but it would be worth investigating whether it can be made to work.

hihi, ... I did have the same idea and found out, that we don't have "such a parser rule atm" ... I didn't dig deeper into `$tw.modules.createClassesFromModules = function(moduleType,subType,baseClass) {` .... because I wasn't very successful lately if my PRs also touched related core elements. ... 

I'll have a closer look.

pmario:
I did update the original post.

@Jermolene ... I did remove the duplicated code and added  `exports.types = {block: true, pragma: true};` to the commentblock pragma. 

I did add a little bit of documentation to the "HTML in Wikitext" tiddler.

pmario:
@Jermolene Bump. I think I did address your concerns. There is almost no code needed. Most of it is comments, tests and docs.

editions/tw5.com/tiddlers/wikitext/HTML in WikiText.tid
```diff
@@ -14,7 +16,36 @@ This is my nice and simple block of text. HelloThere
 </article>
 ```
 
-[[Widgets share the same syntax as HTML tags|Widgets in WikiText]], and so the following information applies to them, too.
+!! Comments
+
+A html-comment as described above has been possible since ~TiddlyWiki Version 5.1.0.
+
+<<.from-version 5.2.0>> It will be possible to use "performant" ''html-comments together with pragmas''. Comments as described below can be cached by the TW core, so they are performant.
```
Jermolene, line 23: We don't refer to TiddlyWiki as TW in the docs. This might be better expressed as:

```
 <<.from-version 5.2.0>> Comments can now be freely intermixed with pragmas as well as within the main body of a block of wikitext.
```

editions/tw5.com/tiddlers/wikitext/HTML in WikiText.tid
```diff
@@ -14,7 +16,36 @@ This is my nice and simple block of text. HelloThere
 </article>
 ```
 
-[[Widgets share the same syntax as HTML tags|Widgets in WikiText]], and so the following information applies to them, too.
+!! Comments
+
+A html-comment as described above has been possible since ~TiddlyWiki Version 5.1.0.
```
Jermolene, line 21: I think this line can be cut, it's not relevant for users

editions/tw5.com/tiddlers/wikitext/HTML in WikiText.tid
```diff
@@ -14,7 +16,36 @@ This is my nice and simple block of text. HelloThere
...
+
+<<.from-version 5.2.0>> It will be possible to use "performant" ''html-comments together with pragmas''. Comments as described below can be cached by the TW core, so they are performant.
+
+```
+<!-- NEW: Comment that describes the macro -->
+\define test()
+some text <!-- inline comment -->
+\end
+
+<!-- This has always been possible -->
```
Jermolene, line 31: Again I think it confuses matters to try to explain history in this way. All we need is the note above.

editions/tw5.com/tiddlers/wikitext/HTML in WikiText.tid
```diff
@@ -14,7 +16,36 @@ This is my nice and simple block of text. HelloThere
...
+<<test>>
+```
+
+If a macro use "text substitution" as described in [[Macro Definitions in WikiText]], you ''can use'' html-comments, ''but'' the will be ''slower'', since the core isn't allowed to cache them. Eg:
```
Jermolene, line 35: I don't think it's worth confusing this explanation with the discussion about caching. Comments are cached in the same way as other constructions, and so I don't see anything special that we need to communicate here.

Jermolene (changes requested):
Thanks @pmario minor comments

pmario:
@Jermolene ... Docs has been changed. Thx for the review

Jermolene:
Thank you @pmario

PR #5727, opened 2021-05-27
Add tc-small-gap-right to tag-pills in tags editTemplate

BurningTreeC:
This PR makes the tag-pills in the tags EditTemplate look like the tag-pills in the ViewTemplate, with a small gap between one tag and the other


Jermolene:
Thanks @BurningTreeC

PR #5728, opened 2021-05-28
Extend color parser to accept hex strings containing alpha

cdruan:
Currently, `parseCSSColor()` does not recognize 4-digit or 8-digit hex code. This PR will:

* Allow the parser to recognize alpha channel in the hex codes.

* Stricter parsing--returns null if a hex string contains illegal characters. For example, `parseCSSColor("#0xa")` now returns null instead of [ 0, 0, 170, 1 ].


pmario:
@cdruan .. Why did you close this one? IMO it would be an improvement.

cdruan:
The modification touches the code from another repository, and based on https://github.com/Jermolene/TiddlyWiki5/pull/5849#issuecomment-873564867 and the inactivity, I thought I would remove it to de-clutter the PR list.

PR #5732, opened 2021-05-29
i18n(EditorPreview) translate difference view caption

joebordes:
a couple of missing translations I ran into while testing the es-ES translation


Jermolene:
Thanks @joebordes, catching snags like that is really helpful.

PR #5733, opened 2021-05-29
i18n(en-GB) apply grammarly recommendations found while translating es-ES

joebordes:


Jermolene:
Great stuff, @joebordes thank you!

PR #5734, opened 2021-05-29
i18n(Tiddlyweb) setup basic translation support and translate two labels

joebordes:
I'm not sure this is the correct way to translate a plugin. I added a "translation string" multitiddler and used the values to translate the button and hint. Then I added a similar file to the es-ES plugin with the translation, reloaded and it worked.

But, if there is a better or "correct" way to translate plugins, let me know and I will try to adapt.

Note, I did look at the other plugins but couldn't find anything useful.

### DIFF
--- a/plugins/tiddlywiki/tiddlyweb/i18n.multids
+++ b/plugins/tiddlywiki/tiddlyweb/i18n.multids
@@ -1,0 +1,4 @@
+title: $:/config/plugins/tiddlywiki/tiddlyweb/language/
+
+ServerStatus: Server status
+ServerStatus/Description: Status of synchronisation with server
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @joebordes we don't yet have a standard approach for making the core plugins translatable. It's well overdue, and I'd be interested in nailing things down.

I think @pmario has been involved in earlier discussions about this. We actually have two issues:

* Making the core `$:/plugins/tiddlywiki/*` plugins translatable. In these cases we can pack the translations into the main `$:/languages/*` plugins
* Making 3rd party plugins translatable. In these cases the translations might be delivered within the plugin itself, or as one or more dependent plugins

pmario:
See: https://github.com/Jermolene/TiddlyWiki5/pull/5418 ... This is a global approach, that will allow core _and_ 3rd party plugins to be translated. ... I started to dig into this again yesterday, to see if the mechanism _and_ the code can be improved. 

@joebordes I did create a [short video](https://youtu.be/yZMaZ4qeaKk), how the mechanism is supposed to work.

pmario:
Close in favour of: Introducing bundled sub plugins #8957

PR #5742, opened 2021-06-01
Extend lookup operator to work with fields and indexes

joshuafontany:
I am creating a new PR for this, as the [old one](https://github.com/Jermolene/TiddlyWiki5/pull/5315) was reverted and no longer picks up changes from my branch. I have resolved the bug mentioned in that thread.

I have also setup a couple of Examples that test the new functionality, and would like input (@Jermolene @saqimtiaz) on how to structure tests in the `./editions/test` folder.

Also, I noticed that stock TW5 doesn't really _have_ a lot of json data tiddlers to test on that are valid "one level deep" json.

As such, the following test results in "missing-0" displayed twice, even thought the first ($:/HistoryList) **has** an object at index 0.
```
<<.operator-example 6 "HistoryList MissingHistoryList +[lookup:missing-0:index[$:/]]" "Retrieve index `0` from the `$:/HistoryList` and `$:/MissingHistoryList`.">>
```
Right now the `wiki.extractTiddlerDataItem ()` method will only return a value if it is a string or number type. I would prefer to JSON.stringify() all other non-null (!== null, !== undefined) types.

My idea would be to change the following code in `wiki.js`:
```
/*
Extract an indexed field from within a data tiddler
*/
exports.extractTiddlerDataItem = function(titleOrTiddler,index,defaultText) {
	var data = this.getTiddlerDataCached(titleOrTiddler,Object.create(null)),
		text;
	if(data && $tw.utils.hop(data,index)) {
		text = data[index];
	}
	if(typeof text === "string" || typeof text === "number") {
		return text.toString();
	} else if(text != null) {
		return JSON.stringify(text);
	} else {
		return defaultText;
	}
};
```

And thus, example 6 would render out as:

- {"title": "HelloThere"}
- missing-0

Thoughts?


saqimtiaz:
@joshuafontany For tests have a look at https://github.com/Jermolene/TiddlyWiki5/blob/652e8b1262786a3e7bcc01b51b4319749ff6674c/editions/test/tiddlers/tests/test-filters.js#L211-L213

You can add more tests here for the new features introduced for the lookup operator, as well as to test that the old behaviour remains unchanged for tiddlers that do not exist. You can also add more tiddlers to the fixtures in the [setupWiki](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-filters.js#L76) method.

With regards to examples for the documentation, I recommend adding a JSON data tiddler to tiddlywiki.com to be used for this purpose, rather than changing `extractTiddlerDataItem` which may have unforeseen knock on consequences.

pmario:
>... You can also add more tiddlers to the fixtures in the setupWiki method.

Yes, but be really careful adding new tiddlers there. It may have side-effects for other tests. So be prepared for some extra work!

AnthonyMuscio:
I was just reporting because my addition request of the same was deleted? As suggested I will add the "request here" in case its of any use.

- The lookup is the only operator in tiddlywiki that can transclude the result. However it is also almost the only transclusion related feature in tiddlywiki that does not allow us to specifiy a fieldname or index name.

- With the increased value of using the "filtered transclusions" there would be added value in being able to lookup field and index values with or without a tiddler prefix applied first.

- Enhance the Lookup operator to include parameters to also be able to return a fieldname/index value
- Ideal allow lookup to find a field or index in a single tiddlername without a prefix. eg: `[[tiddlername]lookup[],<index>]` 
- Thus lookup operator clearly illustrates retrieving information from tiddlers, tiddler fields and data tiddlers.
- For example; The addition of a second prefix eg `lookup:default:text[prefix]` to maintain backward compatibility text will be the default, field and index new values. In the case of field or index an additional parameter will be used `lookup:default:index[prefix],[indexname]` and also allow no prefix such as `lookup:field[],<fieldname>` which would simply look up the fieldname in the current title and transclude it.

Other filter operators can achieve a similar result such as `[tiddlername]addprefix[prefix]get[fieldname]]` how ever these are also used for obtaining intermediate results such as `[tiddlername]addprefix[prefix]get[fieldname]match[yes]]` extending the lookup operator would be easier and make clearer what a filter containing this operator is doing.


**Additional context**
- Almost all other transclusion widgets also permit the transclusion of the values of fields and indexes. Without this change the lookup operator remains another exception
- New users may find it easier to revert to the lookup operator before they fully come to terms with the wider filter technology.
- It would be easier to explain, self document and write filters that can use a simplified "lookup request".
- It took me ages to understand how to get the contents in data tiddlers when I started with TW5, this would have helped me a lot.

joshuafontany:
While I still feel that returning the Stringified value of any non-null/non-undefined property is the way to go, I can see that there might be backwards compatibility issues there....

As such, we leave that aside, and this PR is ready to go (tests have been added and run). @Jermolene @saqimtiaz I know there is a bug-fix release prioritized, so no worries if this is not gotten to right away...

Jermolene:
Hi @joshuafontany thanks for your patience, I had missed this.

But looks good, and makes sense, so I've added to #6266 for consideration after v5.2.1

I've also clarified the title

Jermolene:
Thanks @joshuafontany

PR #5743, opened 2021-06-01
Extend Server with 'required-plugins' option

joshuafontany:
@Jermolene This PR changes the TW Server to have a `required-plugins` defaultVariable, which can be overridden by passing in a `required-plugins` value in the constructor's options. This defaults to: `"tiddlywiki/filesystem", "tiddlywiki/tiddlyweb"`

The test for the required plugins has been generalized, and also the `servername` property is now set from `options.servername || this.wiki.getTiddlerText("$:/SiteTitle") || "TiddlyWiki5"`.

Finally, the initial `this.getStatus()` call has been moved outside of the `$tw.syncer` constructor, as this was causing custom `syncadaptor` code that referenced `$tw.syncer` to fail (as that had not been initialized yet). This call is moved to just after the syncer is initialized inside `core/modules/startup/startup.js`.

These changes have allowed me to extend a Server object for my Bob2.0 re-write, instead of writing a custom parallel server.

Thanks!
Joshua Fontany


core/modules/server/server.js
```diff
@@ -322,8 +326,16 @@ Server.prototype.listen = function(port,host,prefix) {
 		port = process.env[port] || 8080;
 	}
 	// Warn if required plugins are missing
-	if(!this.wiki.getTiddler("$:/plugins/tiddlywiki/tiddlyweb") || !this.wiki.getTiddler("$:/plugins/tiddlywiki/filesystem")) {
-		$tw.utils.warning("Warning: Plugins required for client-server operation (\"tiddlywiki/filesystem\" and \"tiddlywiki/tiddlyweb\") are missing from tiddlywiki.info file");
+	var missing = [];
+	for (let index = 0; index < this.variables["required-plugins"].length; index++) {
+		const r = this.variables["required-plugins"][index];
```
pmario, line 331: We don't use `let` or `const` atm and please give `r` a real name. ... For me the next line looked like a "broken" regexp ...

pmario: code style. `for` should look like

```
for (var index=0; index<this.variables["required-plugins"].length; index++) {
```

eg: no spaces between `index=0;`

core/modules/server/server.js
```diff
@@ -322,8 +326,16 @@ Server.prototype.listen = function(port,host,prefix) {
...
+		if (!this.wiki.getTiddler("$:/plugins/"+r)) {
+			missing.push(r);
+		}
+	}
+	if(missing.length > 0) {
+		var error = "Error: Plugin(s) required for client-server operation are missing from the command line or the tiddlywiki.info file: \"" + missing.join("\", \"") + "\"";
+		$tw.utils.error(error);
```
pmario, line 338: I'm not happy with the line length here, but I also don't have a good result. ... What about: 

```
	$tw.utils.error( ["Error: Plugin(s) required for client-server operation are missing from the command line ",
		"or the tiddlywiki.info file: \"", missing.join("\", \""), "\""].join(),
```

It's still "kind of ugly", but would be shorter. @Jermolene ... what do you think

core/modules/server/server.js
joshuafontany: Thanks @pmario. I have been bouncing back and forth between different code styles, and this is really helpful!

core/modules/server/server.js
joshuafontany: Tried a multi-line error.

pmario:
@joshuafontany ... This is still a "draft PR". Is this intentional. ... If you want it to get reviewed, you'd probably make it a real PR.

joshuafontany:
@Jermolene I have found a bug with the TW server basic with login, but will write that up as as a separate PR. Does this one look good for 5.2.0?

saqimtiaz:
I believe there are actual use cases for starting the server without the "tiddlywiki/filesystem", "tiddlywiki/tiddlyweb" plugins and that the `required-plugins` default is not backwards compatible.

pmario:
>I believe there are actual use cases for starting the server without the "tiddlywiki/filesystem", "tiddlywiki/tiddlyweb" plugins and that the required-plugins default is not backwards compatible.

I don't think so. If you start a server without those 2 plugins data-loss is guaranteed. 

I would be curious to here about those use-cases?

Jermolene:
I often use the listen command without either or both of the filesystem/tiddlyweb plugins. It's a quick way to get a read-only wiki for testing purposes, or to spin up a wiki that is intended to use other mechanisms for saving/syncing.

pmario:
>I often use the listen command without either or both of the filesystem/tiddlyweb plugins. It's a quick way to get a read-only wiki for testing purposes, or to spin up a wiki that is intended to use other mechanisms for saving/syncing.

But wouldn't it be better to implement a ´tiddlywiki myWiki --listen readonly´ option, instead of starting a server, where we know, that ordinary users will have a problem with?

Jermolene:
> But wouldn't it be better to implement a `tiddlywiki myWiki --listen readonly` option, instead of starting a server, where we know, that ordinary users will have a problem with?

The issue is backwards compatibility.

joshuafontany:
@Jermolene @saqimtiaz @pmario Thanks for reviewing this. We can make it backwards compatible by changing the $tw.utils.error to $tw.utils.warning (to hide that from the user-loaded UI), as it originally was.

pmario:
@joshuafontany ... I think the backwards compatibility concern is this: 

>This PR changes the TW Server to have a required-plugins defaultVariable, which can be overridden by passing in a required-plugins value in the constructor's options. **This defaults to**: "tiddlywiki/filesystem", "tiddlywiki/tiddlyweb"

I did mark it **bold** ... This defaults to ... It isn't default now that's why it may be problem

joshuafontany:
@pmario I'm not sure I understand that concern. The defaults exist in the code already, but they are card coded in-line (not passed via the options), line 352
`if(!this.wiki.getTiddler("$:/plugins/tiddlywiki/tiddlyweb") || !this.wiki.getTiddler("$:/plugins/tiddlywiki/filesystem"))`

The server can (now) be launched and function to just serve (not sync) a wiki without these "default plugins" - I'm just offering a way to customize the warning message (not prevent the server from starting).

joshuafontany:
@jeremy I think the concerns about passing in `options.required-plugins` are resolved. I specifically did not want a Bob `ws-server` user to see a false error if another plugin is actually handling syncing from the browser or from node.

joshuafontany:
@Jermolene I have also added a `tls-passphrase` server-variable that allows passing a passphrase to init the HTTPS server, if the cert was created with a passphrase. I'll update the docs here sortly.

Jermolene:
Thanks @joshuafontany 

I'm just wondering if it might be problematic that these changes introduce a server variable whose value is an array, rather than a string as at present. It means that one cannot usefully set the `required-plugins` variable on the command line, for example. We could make it be a comma delimited list instead of an array?

Alternatively, perhaps this can just be a property on the server object, and not a variable?

joshuafontany:
@Jermolene I like the idea of making it a server-object property. This hides the complication from the normal --listen command, and would allow custom commands to take in variables and setup the server as needed. I'll refactor with that in mind.

core/modules/server/server.js
```diff
@@ -171,7 +171,15 @@ Server.prototype.get = function(name) {
 };
 
 Server.prototype.addRoute = function(route) {
+	// Find out if the route exists
+	let index = this.routes.findIndex((thisRoute) => thisRoute.path.toString() === route.path.toString());
```
pmario, line 175: we don't use `let` and arrow functions. TW is still ES5 atm

core/modules/server/server.js
```diff
@@ -171,7 +171,15 @@ Server.prototype.get = function(name) {
...
+	if (index === -1) {
+	// Push the new route if not found
 	this.routes.push(route);
```
pmario, line 178: indent +1 also in the else path

core/modules/server/server.js
```diff
@@ -337,11 +337,11 @@ Server.prototype.listen = function(port,host,prefix) {
 	var missing = [];
 	for (var index=0; index<this.requiredPlugins.length; index++) {
 		var name = this.requiredPlugins[index];
-		if (!this.wiki.getTiddler("$:/plugins/"+name)) {
+		if (!this.wiki.getTiddler(name)) {
 			missing.push(name);
 		}
 	}
-	if(missing.length > 0) {
+	if(missing.length > 0) {debugger;
```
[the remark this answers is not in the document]
joshuafontany: @pmario Thanks. :) I'm marking this as a draft for now.

core/modules/server/server.js
Arlen22: This isn't correct, because different HTTP METHOD requests can have the same path. I'm pretty sure this isn't backward compatible. 

Furthermore, I think this should be left up to the calling code since the routes list is an instance variable and can certainly be modified by calling code (i.e. `listen` and `ws-server` commands) and child classes. So whoever is calling addRoute can do that check for themself. If this is for the default route loader, I don't think this is how we should fix that conundrum. While this may be a useful feature, I don't think this way of doing it really follows TiddlyWiki standard practice. I've listed below several ways that plugins can legitimately or illegitimately override already loaded routes.

If you are using the `--listen` command, then I would recommend making use of the `th-server-command-post-start` hook, which has the server instance as its first argument. If you wrote your own command you can access the server instance there. 

Here are the different options I can think of which plugins can currently use to remove routes which they want to replace. 

- overloading the addRoute function (`var original = server.addRoute; server.addRoute = function() { original(...arguments); }`), or manually removing the offending routes that you want to override (`server.routes.splice(index, 1)`). 
- overriding the shadow tiddler for those routes by including a tiddler with the same name in your plugin that either creates a new route or doesn't include the `module-type` field. This will override the shadow tiddler from any previous plugin according to standard plugin loading order. 

The second option is preferred for feature plugins that just add one feature. If I'm not mistaken, in this particular case this will be used for the core Bob plugin in the root datafolder so interfering with other plugins is not a concern.

editions/tw5.com/tiddlers/$__StoryList.tid
```diff
@@ -1,5 +0,0 @@
-created: 20210322152203906
-list: [[Documentation Macros]] HelloThere GettingStarted Community
-modified: 20210322152237613
-title: $:/StoryList
-type: text/vnd.tiddlywiki
```
Arlen22, line 5: @Jermolene @Is this supposed to be deleted? Am I missing something here?

editions/tw5.com/tiddlers/webserver/WebServer API_ Get All Tiddlers.tid
```diff
@@ -20,7 +20,7 @@ In order to avoid denial of service attacks with malformed filters in the defaul
 To enable a particular filter, create a tiddler with the title "$:/config/Server/ExternalFilters/" concatenated with the filter text, and the text field set to "yes". For example, the TiddlyWeb plugin includes the following shadow tiddler to enable the filter that it requires:
 
 ```
-title: $:/config/Server/ExternalFilters/[all[tiddlers]] -[[$:/isEncrypted]] -[prefix[$:/temp/]] -[prefix[$:/status/]]
+title: $:/config/Server/ExternalFilters/[all[tiddlers]] -[[$:/isEncrypted]] -[prefix[$:/temp/]] -[prefix[$:/status/]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] -[[$:/library/sjcl.js]] -[[$:/core]]
```
Arlen22, line 23: I don't see a change in any other file here that corrosponds to this change. Is this meant to bring the docs in line with the current default?

Arlen22 (changes requested):
Looked through this and have a few comments. Hopefully it all makes sense.

joshuafontany:
> Looked through this and have a few comments. Hopefully it all makes sense.

I really appreciate it! I got that route code from the Bob 1.x method that loads routes... I will refactor with the above in mind.

editions/tw5.com/tiddlers/webserver/WebServer API_ Get All Tiddlers.tid
joshuafontany: Yes.

editions/tw5.com/tiddlers/$__StoryList.tid
joshuafontany: I start lots of experimental servers, so sometimes the core editions get Storylists generated when those folders should not have that tiddler... but 8 may have deleted one in error here...

core/modules/server/server.js
joshuafontany: Reverted.

joshuafontany:
@Jermolene I believe I have reduced the changes here to the bare minimum needed for my Bob2.0 setup.

One thing to note is that I added an option to pass `options.boot` through to the syncadaptors and this the filesystem utils. This will allow multiple `boot` meta-data stores to exist when I do have Bob2.0 loading multiple wikis - each with their own syncer/filesystemadaptor-> filesystem logic.

joshuafontany:
Server options PR: https://github.com/Jermolene/TiddlyWiki5/pull/5899

filesystemAdaptor options PR: https://github.com/Jermolene/TiddlyWiki5/pull/5900

joshuafontany:
This got jumbled.. Closing this PR in favor of the separate ones.

PR #5747, opened 2021-06-01
Preserve the scroll-position over refresh cycles

BurningTreeC:
This PR preserves the scroll position in the root window and new windows over page-refresh cycles

### DIFF
--- a/core/modules/startup/render.js
+++ b/core/modules/startup/render.js
@@ -70,11 +70,16 @@ exports.startup = function() {
 	var deferredChanges = Object.create(null),
 		timerId;
 	function refresh() {
+		// Save the current scroll position
+		var scrollX = window.scrollX,
+			scrollY = window.scrollY;
 		// Process the refresh
 		$tw.hooks.invokeHook("th-page-refreshing");
 		$tw.pageWidgetNode.refresh(deferredChanges);
 		deferredChanges = Object.create(null);
 		$tw.hooks.invokeHook("th-page-refreshed");
+		// Restore the scroll position
+		window.scroll(scrollX,scrollY);
 	}
 	// Add the change event handler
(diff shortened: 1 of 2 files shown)


BurningTreeC:
Edited the OP

pmario:
I'm not sure, if I understand this. .. What happens, if the Story river after refresh is much shorter as the story river before?

BurningTreeC:
> I'm not sure, if I understand this. .. What happens, if the Story river after refresh is much shorter as the story river before?

@pmario, you mean if we remove a tiddler from the StoryRiver for example? As far as my testings go, nothing special happens

Jermolene:
Hi @BurningTreeC if we're going to try to do scroll position preservation then I think we need to adopt a technique that will also handle scroll positions within scrollable elements as well as the document body.

BurningTreeC:
> Hi @BurningTreeC if we're going to try to do scroll position preservation then I think we need to adopt a technique that will also handle scroll positions within scrollable elements as well as the document body.

Hi @Jermolene, how would such a technique look like? Would we store the scroll positions somewhere globally and the corresponding domNode when the page refresh starts and then restore it when it ends?

Jermolene:
> Hi @Jermolene, how would such a technique look like? Would we store the scroll positions somewhere globally and the corresponding domNode when the page refresh starts and then restore it when it ends?

Actually, aren't you doing just that as part of the focus preservation PR?

Jermolene:
Apologies @BurningTreeC by "just that" I meant "aren't you restoring scroll positions as part of your focus preservation PR"?

BurningTreeC:
> Apologies @BurningTreeC by "just that" I meant "aren't you restoring scroll positions as part of your focus preservation PR"?

Yes that's true @Jermolene

PR #5755, opened 2021-06-04
Make the insert- and remove-animations of storyviews work in new windows

BurningTreeC:
This PR makes the insert- and remove-animations of domNodes work in new windows, too, where `!(targetElement instanceOf Element)` fails


Jermolene:
Good catch, thanks @BurningTreeC

PR #5758, opened 2021-06-05
Keyboard widget: don't refresh when class changes

BurningTreeC:
This PR makes the keyboard widget just update the domNode's classes if the class attribute changes, instead of refreshing the whole thing


Jermolene:
Thanks @BurningTreeC

PR #5759, opened 2021-06-05
Make navigation in new windows work for storyviews

BurningTreeC:
This PR makes navigation via `tm-scroll` message work in new windows, too

The test `!(targetElement instanceOf Element)` fails in new windows


Jermolene:

PR #5765, opened 2021-06-06
Trigger actions when tiddlers change

BurningTreeC:
This PR allows to trigger actions when a specified tiddler changes

- We create a tiddler tagged `$:/tags/TiddlerChangeAction`
- The text field contains the action to be invoked
- The `change-tiddler` field contains the title of the tiddler that must change in order for the actions to be invoked

Sometimes tiddlers may change automatically, like the tiddler `$:/info/darkmode` and it would be suitable to invoke an action if they change

I also have in mind adding a mechanism like the one that detects dark mode, which detects if the window-width is lower than `$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint` ... but I'll have to ask @Jermolene if that's doable

### DIFF
--- a/core/modules/tiddler-change.js
+++ b/core/modules/tiddler-change.js
@@ -19,7 +19,7 @@ function TiddlerChangeManager(options) {
 	this.tiddlerActionList = [];
 	this.updateTiddlerChangeLists(this.getTiddlerList());
 	$tw.wiki.addEventListener("change",function(changes) {
-		self.handleShortcutChanges(changes);
+		self.handleTiddlerChanges(changes);
 	});
 }
 
@@ -52,27 +52,7 @@ TiddlerChangeManager.prototype.handleChangedTiddlerEvent = function(event) {
 	return false;
 };
 
-TiddlerChangeManager.prototype.detectNewShortcuts = function(changedTiddlers) {
-	var shortcutConfigTiddlers = [],
-		handled = false;
-	$tw.utils.each(this.lookupNames,function(platformDescriptor) {
-		var descriptorString = "$:/config/" + platformDescriptor + "/";
-		Object.keys(changedTiddlers).forEach(function(configTiddler) {
-			var configString = configTiddler.substr(0, configTiddler.lastIndexOf("/") + 1);
-			if(configString === descriptorString) {
-				shortcutConfigTiddlers.push(configTiddler);
-				handled = true;
-			}
-		});
-	});
-	if(handled) {
-		return $tw.utils.hopArray(changedTiddlers,shortcutConfigTiddlers);
-	} else {
-		return false;
-	}
-};
-
-TiddlerChangeManager.prototype.handleShortcutChanges = function(changedTiddlers) {
+TiddlerChangeManager.prototype.handleTiddlerChanges = function(changedTiddlers) {
 	var newList = this.getTiddlerList();
 	var hasChanged = $tw.utils.hopArray(changedTiddlers,this.actionTiddlers) ? true :
 		($tw.utils.hopArray(changedTiddlers,newList) ? true : false);
--- a/core/modules/tiddler-change.js
+++ b/core/modules/tiddler-change.js
@@ -2,9 +2,7 @@
 title: $:/core/modules/tiddler-change.js
 type: application/javascript
 module-type: global
-
 Tiddler change handling utilities
-
 \*/
 (function(){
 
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @BurningTreeC that's very interesting, and not something I'd given consideration to before.

One point is that we need to be very careful how we document the semantics. Given the asynchronous nature of the wiki "change" event, it seems like it would be possible for the listener to miss changes to rapidly changing tiddlers. (In other words, the tiddler value is changed again before the actions are triggered).

The present implementation seems like it's got some dependencies to keyboard shortcut handling which will need resolving.

But I think the main problem is that listening for changes to specific, named tiddlers doesn't cover all the common use cases. In particular, a pattern we often see in plugins is needing to keep track of changes to the results of a tiddler filter. For example, keeping track of tiddlers with a particular tag, and being able to process insertions, deletions and changes. With a primitive like that, we could potentially implement the storyview functionality in wikitext.

core/modules/tiddler-change.js
```diff
@@ -0,0 +1,87 @@
...
+TiddlerChangeManager.prototype.getTiddlerList = function() {
+	return $tw.wiki.getTiddlersWithTag("$:/tags/TiddlerChangeAction");
+};
+
+TiddlerChangeManager.prototype.updateTiddlerChangeLists = function(tiddlerList) {
+	this.actionTiddlers = tiddlerList;
+	for(var i=0; i<tiddlerList.length; i++) {
+		var title = tiddlerList[i],
+			tiddlerFields = $tw.wiki.getTiddler(title).fields;
+		this.tiddlerChangeList[i] = tiddlerFields["change-tiddler"] !== undefined ? tiddlerFields["change-tiddler"] : undefined;
```
pmario, line 35: I think, the field name should be different. IMO it contains the "triggering-tiddler" or may be "watch-tiddler" which I do like more.

BurningTreeC: I like them both @pmario

BurningTreeC:
> Hi @BurningTreeC that's very interesting, and not something I'd given consideration to before.

Do you think I should go ahead with this mechanism?

> One point is that we need to be very careful how we document the semantics. Given the asynchronous nature of the wiki "change" event, it seems like it would be possible for the listener to miss changes to rapidly changing tiddlers. (In other words, the tiddler value is changed again before the actions are triggered).

I'll try testing this behavior

> 
> The present implementation seems like it's got some dependencies to keyboard shortcut handling which will need resolving.

Yes, I just quickly copied things over from the keyboard shortcut handling mechanism, I'll resolve these things

> 
> But I think the main problem is that listening for changes to specific, named tiddlers doesn't cover all the common use cases. In particular, a pattern we often see in plugins is needing to keep track of changes to the results of a tiddler filter. For example, keeping track of tiddlers with a particular tag, and being able to process insertions, deletions and changes. With a primitive like that, we could potentially implement the storyview functionality in wikitext.

Could you make more detailed examples if I don't ask too much?

pmario:
>I also have in mind adding a mechanism like the one that detects dark mode, which detects if the window-width is lower than $:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint ... but I'll have to ask @Jermolene if that's doable

ATM `$:/info/browser/screen/width` isn't a dynamic variable. See some similar discussion at: https://github.com/Jermolene/TiddlyWiki5/pull/4374 ... So imo we would need a new namespace for the `$:/info/browser/` section. May be `$:/info/browser/active/xxx` ... BUT those active elements may fire at a rapid pace, which IMO would kill the nodejs syncer ... 

So those rapid firing tiddlers would need to be blocked from syncing and they may have the potential to create "race conditions" due to the "change" event being async. As Jeremy mentioned.

Jermolene:
> Could you make more detailed examples if I don't ask too much?

Hi @BurningTreeC I'm suggesting that a general purpose tiddler tracking mechanism would track changes to filters, not specific, individual tiddlers.

It could run in two different ways:

* Tracking changes to the result list itself, and thus running actions when a new title appears in the result list, or a title is removed
* Tracking changes to the tiddlers identified by the filter. There would have to be a way to specify which fields are considered when comparing tiddlers

I'm not sure if both would be required. On the face of it, the second behaviour could be emulated with some combination of filters that get the values of the tiddlers in the result list. For example, a filter like `[{$:/title/of/tiddler/to/be/tracked}]` could be used to track changes to the text field of a specific tiddler

Another example would be tracking `[tag[$:/tags/Post]]` to initiate a publishing job whenever a blog post is changed.

BurningTreeC:
> > Could you make more detailed examples if I don't ask too much?
> 
> Hi @BurningTreeC I'm suggesting that a general purpose tiddler tracking mechanism would track changes to filters, not specific, individual tiddlers.

Thanks @Jermolene,
> 
> It could run in two different ways:
> 
> * Tracking changes to the result list itself, and thus running actions when a new title appears in the result list, or a title is removed
> * Tracking changes to the tiddlers identified by the filter. There would have to be a way to specify which fields are considered when comparing tiddlers
> 
> I'm not sure if both would be required. On the face of it, the second behaviour could be emulated with some combination of filters that get the values of the tiddlers in the result list. For example, a filter like `[{$:/title/of/tiddler/to/be/tracked}]` could be used to track changes to the text field of a specific tiddler
> 
> Another example would be tracking `[tag[$:/tags/Post]]` to initiate a publishing job whenever a blog post is changed.

Thank you for your explanations, I'll try building something that tracks changes to the result lists and see if that works out

BurningTreeC:
ready for a quick review @Jermolene @pmario

BurningTreeC:
@Jermolene, any thoughts on this?

Jermolene:
Hi @BurningTreeC there's quite a lot still to think about:

* I don't have a good idea of the use cases for the technique
* I don't know whether tracking the filter results is sufficient (see my point above about detecting tiddler changes instead)
* The semantics seem awkward because of the interactions with the refresh mechanism. For example, we can't really guarantee to "see" all the values that a tiddler takes on over time

So, for me at the moment it remains an interesting new idea to think about, but I don't think we're close to being able to merge it.

PR #5766, opened 2021-06-06
Fix #5760 - tm-focus-selector doesn't work in new windows

### LINKED ISSUE #5760: tm-focus-selector doesn't work in new windows
**Describe the bug**
When I open the `$:/core/ui/PageTemplate` tiddler in a new window and try to focus the `sidebar-search` via `ctrl-shift-F`, `document.querySelector` finds the `sidebar-search` in the root tiddlywiki window, not the one in the newly opened window

**Expected behavior**
`ctrl-shift-F` in the new window focuses the sidebar-search of the new window

 - latest prerelease

BurningTreeC:
This PR fixes #5760

- in keyboard.js we now pass the original event to `invokeActionString`
- in rootWidget.js we now use event.event.target.ownerDocument or document as fallback if event.event is undefined

### DIFF
--- a/core/modules/keyboard.js
+++ b/core/modules/keyboard.js
@@ -324,7 +324,7 @@ KeyboardManager.prototype.handleKeydownEvent = function(event) {
 	if(key !== undefined) {
 		event.preventDefault();
 		event.stopPropagation();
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @BurningTreeC

PR #5769, opened 2021-06-07
Fix slide animations if closing while opening

BurningTreeC:
To reproduce the problem:
- Go to https://tiddlywiki.com/prerelease
- Set the AnimationDuration to something high, like 4000
- open the info panel of a tiddler and click somewhere outside the opening info panel so that it closes while it opens
- see that when it should be closed, it reopens for a short time-period

The problem is, that the `setTimeout` of the `slideOpen` animation is still working
This PR clears the slide-open-timeout when a slideClose animation is started and vice versa

### DIFF
--- a/core/modules/utils/dom/animations/slide.js
+++ b/core/modules/utils/dom/animations/slide.js
@@ -23,7 +23,7 @@ function slideOpen(domNode,options) {
 		currPaddingTop = parseInt(computedStyle.paddingTop,10),
 		currHeight = domNode.offsetHeight;
 	// Reset the margin once the transition is over
-	setTimeout(function() {
+	var slideOpenTimeout = setTimeout(function() {
 		$tw.utils.setStyle(domNode,[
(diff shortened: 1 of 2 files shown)


BurningTreeC:
@Jermolene, is this PR something to be considered?

Jermolene:
Thanks @BurningTreeC I think I've noticed this problem before and it's certainly worth fixing.

However, this approach wouldn't be my first choice because it adds a dependency from the generic "slide" utility to the internals of the reveal widget. As a point of style, in such situations it's often better to detect and call a method on the widget rather than reaching in to change a property directly.

But, in this case, I think a more robust approach would be for the animation framework to keep track of animations in progress, and to automatically cancel any previous animations for a particular DOM node before applying the new animation. It looks like that could be done mostly within $:/core/modules/utils/dom/animator.js, with the animator modules just being updated to return a function that cancels the animation.

PR #5771, opened 2021-06-08
Add test for event.event.target in rootwidget

BurningTreeC:
This PR adds a test for `event.event.target` being undefined in the tm-focus-selector listener


BurningTreeC:
@Jermolene, I think this PR should be considered for merging

Jermolene:
Thanks @BurningTreeC

PR #5776, opened 2021-06-10
Use event.event.view.confirm for confirmation messages in navigator.js and action-confirm.js

BurningTreeC:
This PR makes confirmation messages appear in the window where they were invoked


saqimtiaz:
Do we perhaps need a check for `event.event` in all instances before accessing `event.event.view`?

Maybe higher up in the method just define a variable for `view` and re-use that as needed.

saqimtiaz:
[action-confirm](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/widgets/action-confirm.js) would benefit from the same improvement.

Jermolene:

PR #5790, opened 2021-06-13
CodeMirror: include reference to edit widget

saqimtiaz:
Pass a reference to the edit widget to the CodeMirror instance. This is helpful for CodeMirror addons that need to render widgets or get variable values from the edit widget.


saqimtiaz:
Bump.

Jermolene:
Thanks @saqimtiaz

PR #5798, opened 2021-06-14
Focus-Preservation over refresh cycles III

BurningTreeC:
This PR preserves the focused domNode (the widget with the focused domNode) over refresh cycles
It also adds keyboard shortcuts for editing and closing the focused tiddler and navigating to the next/previous tiddler in the story river

This PR also includes something that adds `dom-` prefixed attributes to the keyboard widget. Something that should be addressed in another PR (and probably differently than I did) but it's needed in order for everything to work so here's my implementation of it

This should be considered a proof-of-concept PR and I hope to be able to discuss it here and to gather ideas


BurningTreeC:
Closing this PR in favor of #5997

PR #5812, opened 2021-06-20
Improve import cancel warning

### LINKED ISSUE #1219: Change alert for when cancelling import
[ Do you wish to delete the tiddler "$:/Import"? ]

"_Delete!?_ No, I just want to cancel! Ooow I probably ruined the whole TW importing mechanism now?"

quote from Winston Churchill, circa 1939.

pmario:
This PR fixes  #1219 "Change alert for when cancelling import" 

At the time when this issue was created we didn't have the `action-confirm` widget. The dialogue looks like this: 

grafik

### DIFF
--- a/core/ui/ViewTemplate/import.tid
+++ b/core/ui/ViewTemplate/import.tid
@@ -4,13 +4,10 @@ tags: $:/tags/ViewTemplate
 \define lingo-base() $:/language/Import/
 
 \define confirmCancel()
-<$wikify name="message" text="""{{||$:/language/Import/Listing/Cancel/Warning/1}}&nbsp;
-{{||$:/language/Import/Listing/Cancel/Warning/2}}""">
-<$action-confirm $message=<<message>> >
+<$action-confirm $message={{$:/language/Import/Listing/Cancel/Warning}} >
 <$action-deletetiddler $tiddler=<<currentTiddler>>/>
 <$action-sendmessage $message="tm-close-tiddler" title=<<currentTiddler>>/>
 </$action-confirm>
-</$wikify>
 \end
 
 \define buttons()
--- a/core/ui/ViewTemplate/import.tid
+++ b/core/ui/ViewTemplate/import.tid
@@ -4,7 +4,8 @@ tags: $:/tags/ViewTemplate
 \define lingo-base() $:/language/Import/
 
 \define confirmCancel()
-<$wikify name=message text={{$:/language/Import/Listing/Cancel/Warning}}>
+<$wikify name="message" text="""{{||$:/language/Import/Listing/Cancel/Warning/1}}&nbsp;
+{{||$:/language/Import/Listing/Cancel/Warning/2}}""">
 <$action-confirm $message=<<message>> >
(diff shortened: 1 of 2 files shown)


saqimtiaz:
Good idea. I would recommend rephrasing "Do you wish to cancel importing?" to "Do you wish to cancel the import?".

The part about deleting the tiddler will still be confusing for some users but I am not sure if we can drop it either.

By the way, passing a multiline macro to `<$action-confirm>` works in that you get a multiline message displayed which would be useful in this case. The problem is of course that the `.multitids` format does not support multiline strings. Can you think of a work around?

core/ui/ViewTemplate/import.tid
```diff
@@ -3,8 +3,17 @@ tags: $:/tags/ViewTemplate
 
 \define lingo-base() $:/language/Import/
 
+\define confirmCancel()
+<$wikify name=message text={{$:/language/Import/Listing/Cancel/Warning}}>
```
saqimtiaz, line 7: I think it would be preferable in the core to always quote attribute values: `name="message"`

pmario: changed it in the next commit

pmario:
>Good idea. I would recommend rephrasing "Do you wish to cancel importing?" to "Do you wish to cancel the import?".
>
>The part about deleting the tiddler will still be confusing for some users but I am not sure if we can drop it either.

IMO the problem is, that there can be a lot of work in the actual import settings. Eg: changed tiddler names, ... manually selected / unselected elements and so on. 

So I think it is important to have the "$:/Import will be deleted" info, even if it is a bit confusing. IMO it will save users some pain, even if they don't see it immediately.

pmario:
@saqimtiaz ... screenshot updated in the OP ... For some reason I need a [space] at the end of the 1st line, that it created the line-break. ... I used `$nbsp;` to make it clear, that the space shouldn't be removed.

Jermolene:
Thanks @pmario @saqimtiaz I think that the part about deleting $:/Import is confusing though, what is the rationale?

saqimtiaz:
I think the concern is that there might be data in the $:/Import tiddler that the user might inadvertently delete.  The problem is of course that for most users, the message about deleting $:/Import makes them think they are deleting something needed in the future for the import mechanism.

If we are going to keep it I suggest rephrasing as follows:

```
Do you wish to cancel the import?
(The tiddler $:/Import which holds the temporary data for this import will be deleted.)
```

It is far more verbose but hopefully less confusing for new users.

Jermolene:
Do we need users to understand that "cancelling the import" is implemented as "delete the tiddler $:/Import"?

If the concern is that cancelling the import is a destructive operation, I think that is adequately communicated by the presence of a confirmation dialogue.

saqimtiaz:
@Jermolene that is reasonable and I think we can drop the second sentence in the displayed message.

pmario:
@Jermolene @saqimtiaz ... I did remove the second line.

Jermolene:
Great stuff, thanks @pmario @saqimtiaz

PR #5813, opened 2021-06-21
:map filter run prefix with docs and tests

### LINKED ISSUE #5801: [IDEA] :map filter run prefix
**Proposal:** Introduce a new `:map` filter run prefix, which is invoked in turn for each input title with the output of the run replacing the input title. This will allow transforming each individual title with access to that title via the `currentTiddler` variable.

For example:

`[<...input titles...>] :map[get[caption]else{!!title}]`

Currently if one wants to do the same without changing the order of the input titles, the only option is a rather complicated usage of reduce:

`[<...input titles...>] :reduce[get[caption]else{!!title}format:titlelist[]addprefix[ ]addprefix<accumulator>] +[enlist-input[]]`

Other alternatives I have considered included a `map[]` operator, a suffix for the `get[]` operator which allows it to pass through unchanged titles that don't have the specified field, as well as a multi-operand form of `get[]` which returns the first field found in the tiddler: `get[field1],[field2],[field3]`. 

However I think the `:map` filter run prefix idea is more versatile.

saqimtiaz:
This PR adds a `:map` filter run prefix, as well as tests and docs.

Fixes #5801

### DIFF
--- a/core/modules/filterrunprefixes/map.js
+++ b/core/modules/filterrunprefixes/map.js
@@ -1,0 +1,39 @@
+/*\
+title: $:/core/modules/filterrunprefixes/map.js
+type: application/javascript
+module-type: filterrunprefix
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter prefix function
+*/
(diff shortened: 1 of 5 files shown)


Jermolene:
Great thanks @saqimtiaz

Issue #5815, opened 2021-06-21
[IDEA] Prevent the body from being scrolled when modal is open


ght:
It is really distracting to see the story river scrolling when being inside a modal and using the mouse wheel.

Scrolling should be contained in the modal. ~e.g. using the [`overscroll-behavior`](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) CSS property.~ (`overscroll-behavior: contain` isn't sufficient as was discovered in #5816: If the modal fits without scrolling, there is no overscroll to contain, so the body will still be scrollable.)

(This issue was created out of @felixhayashi's PR #1558 from 2015 to fix this, all credit to him. Since `overscroll-behavior` was not yet available at the time, that fix had unwanted side effects, so this issue was created for tracking the problem)

Jermolene:
Thanks @ght is the idea that the `overscroll-behavior` property is applied to the modal container?

pmario:
The intention is, to apply it to the DIV `tc-modal-body` with `overscroll-behavior: contain` setting

pmario:
I'll create a PR

ght:
> Thanks @ght is the idea that the overscroll-behavior property is applied to the modal container?

That's my understanding; `overscroll-behavior` must be set on the scrollable element for which you want to prevent "scroll-chaining", i.e. the element where if you're at the bottom and scroll further down, you don't want to scroll the element's ancestors.

I use TiddlyWiki daily for ~1year now, but my usage is pretty basic, just creating/linking/transcluding tiddlers; I don't think I've seen a modal in TiddlyWiki before, and I've only started to dig through the source, so I'm afraid I can't help with specifics. I just opened this to help bring down the number of lingering open PRs.

---

> I'll create a PR
Thanks @pmario

PR #1558, opened 2015-03-06
Prevent the body from being scrolled when modal is open

felixhayashi:
Hi @Jermolene 

it is really distracting to see the story river scrolling when being inside a modal and using the mouse wheel. A simple addition to the modal code fixes this.

-Felix

PR #5816, opened 2021-06-21
prevent scroll-chaining in modals

pmario:
This PR prevents the scroll-chaining, that usually happens to the main story-river if a list in a modal reaches the end, but the user keeps scrolling. 

This PR fixes: #5815 and #1558   ... (what a conincidence. See the swapped numbers ;)

### DIFF
--- a/editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
+++ b/editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
@@ -1,7 +1,6 @@
 created: 20140912145537860
 footer: <$button message="tm-close-tiddler">Close</$button>
-modified: 20210621223203244
-prevent-scroll-chain: yes
+modified: 20210627054504823
 subtitle: I'm a modal wizard
 title: SampleWizard
 type: text/vnd.tiddlywiki
--- a/editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
+++ b/editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
@@ -1,6 +1,7 @@
 created: 20140912145537860
 footer: <$button message="tm-close-tiddler">Close</$button>
-modified: 20140912145537861
+modified: 20210621223203244
+prevent-scroll-chain: yes
 subtitle: I'm a modal wizard
 title: SampleWizard
 type: text/vnd.tiddlywiki
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -243,6 +243,7 @@ Modal.prototype.adjustPageClass = function() {
 	if(windowContainer) {
 		$tw.utils.toggleClass(windowContainer,"tc-modal-displayed",this.modalCount > 0);
 	}
+	$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",this.modalCount > 0);
 };
 
 exports.Modal = Modal;
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -41,8 +41,6 @@ Modal.prototype.display = function(title,options) {
 	if(!tiddler) {
 		return;
 	}
-	// Make tiddler.fields available
-	this.tiddler = tiddler;
 	// Create the variables
 	var variables = $tw.utils.extend({
 			currentTiddler: title,
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -245,12 +245,7 @@ Modal.prototype.adjustPageClass = function() {
 	if(windowContainer) {
 		$tw.utils.toggleClass(windowContainer,"tc-modal-displayed",this.modalCount > 0);
 	}
-	if (this.tiddler.fields && this.tiddler.fields["prevent-scroll-chain"] === "yes") {
-		$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",this.modalCount > 0);
-	} else if (this.modalCount === 0) {
-		// if the last modal didn't contain a prevent-scroll-chain field
-		$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",false);
-	}
+	$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",this.modalCount > 0);
 };
 
 exports.Modal = Modal;
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -247,6 +247,9 @@ Modal.prototype.adjustPageClass = function() {
 	}
 	if (this.tiddler.fields && this.tiddler.fields["prevent-scroll-chain"] === "yes") {
 		$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",this.modalCount > 0);
+	} else if (this.modalCount === 0) {
+		// if the last modal didn't contain a prevent-scroll-chain field
+		$tw.utils.toggleClass(this.srcDocument.body,"tc-modal-prevent-scroll",false);
 	}
 };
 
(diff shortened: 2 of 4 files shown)


ght:
Looks good code-wise. Sorry noob question, but … what's an easy way to open a modal to test this?

pmario:
That's a very good question. ... It works with the ControlPanel : Plugins : Get More Plugins modal, which actually will scroll. ... 

BUT there still seems to be a problem with: https://tiddlywiki.com/#Modals ... There this PR does _not_ work as expected. ... Will need to investigate

pmario:
@Jermolene ... Felix was right. The only way to reliable prevent the background from scrolling is, to make it `overflow: hidden;`

When I was testing the behaviour, I found out, that for some modals it may be desirable to allow the background to be scrolled. .. That's why this PR defines a new modal tiddler-field: `prevent-scroll-chaining`

If the field is set to `yes` a class `tc-modal-prevent-scroll` is toggled on the ´srcDocument.body´ element. ... This behaviour also works for modals in a "new window"

pmario:
For nested modals ... The first modal that sets the variable to "yes" wins. ... That's also the modal, that re-activates the background scrolling behaviour

pmario:
@Jermolene ... IMO this can be reviewed now

editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
```diff
@@ -1,6 +1,7 @@
 created: 20140912145537860
 footer: <$button message="tm-close-tiddler">Close</$button>
-modified: 20140912145537861
+modified: 20210621223203244
+prevent-scroll-chain: yes
```
saqimtiaz, line 4: I would recommend using a field name that is more intuitive for end users, e.g. `capture-scrolling`

saqimtiaz:
I haven't tested this change but some docs would need to be added to the "Modals" tiddler.

editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
ght: "capture" is not related to the underlying `overscroll-behavior: contain` at all, maybe `contain-scroll`?

Not gonna bike-shed this though, go with what fits best, just please keep in mind that whenever you invent new terminology to help users who don't know the standards, you also force those who do to remember twice as much: how the standard calls it plus the new end-user friendly terms, so it's always a tradeoff.

saqimtiaz: The stated term was an example of an alternative and not a firm proposal. Note however that the implementation in this PR does not use `overscroll-behaviour`.

User demographic matters a great deal in such determinations and the vast majority of TW users are decidedly non-technical.

pmario: >The stated term was an example of an alternative and not a firm proposal. Note however that the implementation in this PR does not use overscroll-behaviour

Yea, .. It doesn't work reliably. ... The parameter is only active if there is something to scroll. .. So only if the content of the modal is bigger than the y-size of the modal. Since many modals have less content it doesn't work there.

pmario:
>I haven't tested this change but some docs would need to be added to the "Modals" tiddler.

That's right, but I also wasn't sure about the field-name

editions/tw5.com/tiddlers/demonstrations/SampleWizard.tid
pmario: The field-name used in the PR isn't carved in stone. I would be OK with the field-name: `contain-scroll` ... I did use a name related to the technical description at MDN.

@Jermolene what do you think?

>from [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior) 2nd paragraph
>
>You may also have noticed that when you have a dialog box with scrolling content on top of a page of scrolling content, once the dialog box's scroll boundary is reached, the underlying page will then start to scroll — this is called **scroll chaining**.

Jermolene:
> When I was testing the behaviour, I found out, that for some modals it may be desirable to allow the background to be scrolled. .. That's why this PR defines a new modal tiddler-field: `prevent-scroll-chaining`

Can you expand on the situations where one wouldn't want to block background scrolling while a modal is displayed?

pmario:
> Can you expand on the situations where one wouldn't want to block background scrolling while a modal is displayed?

grafik

I think the image is a good example. The Modals tiddler contains some information. The modal hides some of text. There may be modals that need it, that the "underlying" text can be read. .. So it will be nice to scroll the background. 

We can also "reverse" the mechanism. So the default is "scrolling forbidden" and the field says "allow-scrolling" it that is preferable.

ght:
> The Modals tiddler contains some information. The modal hides some of text. There may be modals that need it, that the "underlying" text can be read.

This is actually a great point for not using modals in the first place whenever possible :)

Are there actual use-cases that would break by always preventing scrolling the overlay-covered page in the background underneath the modal? I can't imagine anyone would rely on that. But then again, I just remembered [XKCD #1172](https://xkcd.com/1172) :)
<details>

XKCD 1172
</details>

pmario:
>This is actually a great point for not using modals in the first place whenever possible :)

I'm with you. I personally avoid modals completely. The only modal I'm interested in fixing is the "Get more plugins" modal in the ControlPanel.

ght:
I had just taken another look at precisely that, trying to un-modalify it. It actually *seems* to work fine when opened as a tiddler (I did it by creating a new tiddler with only `[[$:/core/ui/ControlPanel/Modals/AddPlugins]]`, then clicking that link).

IMHO the UX is much prefer that way; You can still browse the whole story river normally and go back to the plugin list at your own discretion. Especially for small screens and users relying on assistive technologies this would likely be a plus. Also, moving the dialog from a modal into a (usb)tab on the control panel plugin tab would also bring the plugin installation flow in line with the rest of the settings and make the experience consistent. No other setting uses a modal AFAICS, they are all right there in tabs/accordions.

Are there technical reasons for the plugin list to be a modal?

---

Getting offtopic, but one issue I noticed when using the plugin list as a plain tiddler is that when I then click the "close plugin library" button, I get an `Uncaught TypeError: iframe is undefined` RSOD (on current https://tiddlywiki.com/). Is that an expected consequence of using the modal tiddler out of its natural habitat, or should I open an issue for that?

pmario:
>Getting offtopic, but one issue I noticed when using the plugin list as a plain tiddler is that when I then click the "close plugin library" button, I get an Uncaught TypeError: iframe is undefined RSOD (on current https://tiddlywiki.com/). Is that an expected consequence of using the modal tiddler out of its natural habitat, or should I open an issue for that?

I can't replicate that behaviour. Can you replicate it at tiddlywiki.com or the /prerelease page too. .. I'd consider this an new issue

ght:
I can reliably reproduce this in my normal browser profile, which is quite locked down, but can't reproduce in a clean profile. I'll investigate more later, but it seems to be triggered by my setup.

pmario:
> I can reliably reproduce this in my normal browser profile, which is quite locked down, but can't reproduce in a clean profile. I'll investigate more later, but it seems to be triggered by my setup.

You should open a new issue with this problem. But if your profile prevents iframes to work properly, we may not be able to solve the problem. The plugin import mechanism needs an iframe, to be able to load plugins from servers.

Jermolene:
> I think the image is a good example. The Modals tiddler contains some information. The modal hides some of text. There may be modals that need it, that the "underlying" text can be read. .. So it will be nice to scroll the background.

I don't think it is a very a good example. The point of modals is to take over the users attention, forcing the content to take priority over whatever was displayed before. We keep the old content visible behind a dimmed curtain to help the user understand that it is still there, but we don't expect the user to be able to interact with content behind a modal.

> Are there technical reasons for the plugin list to be a modal?

The rationale was to avoid having two adjacent tabs in control panel that were almost identical looking, one showing the installed plugins, and one showing the plugins from the library. I think it would ask a lot of casual users to understand the difference between them.

pmario:
@Jermolene  I did remove the whole configuration field stuff. So scroll-chaining will be disabled for all modals and can't be switched off.

Jermolene:
Great, thanks @pmario

PR #5832, opened 2021-06-28
Update TiddlyWiki in the Sky for Dropbox.tid

GJRobert:
As [its site states](https://twcloud.github.io/):

> tw5-dropbox - This is the oldest and most dependable loader for TW5 files, but doesn’t support TiddlyWiki Classic files.
> dropbox - This is a newer loader which uses a different method to load files, but only seems to work on Chrome. Hardly anyone uses it, but you’re welcome to try it. It might work


Jermolene:
Thanks @GJRobert

PR #5835, opened 2021-06-28
Fixed bug introduced into transclusions for blank fields in #5736

### LINKED ISSUE #5834: There is a breaking change in the current master
Import the json into tiddlywiki.com and then into prerelease It has a different output. 

[breaking-change-5-23-5-24.zip](https://github.com/Jermolene/TiddlyWiki5/files/6726464/breaking-change-5-23-5-24.zip)

New Tiddler contains: 

```
<$tiddler tiddler=x>
<$transclude field=caption>

<$transclude field=title />

</$transclude>
</$tiddler>
```

x contains

```
title: x
caption: 

aa
```

saqimtiaz:
Fixes #5834

The underlying problem was that I missed a test for when there is a blank field, so naturally that had to be the part that wasn't backwards compatible!

As a precaution I've also added tests for blank indexes and blank fields in plugin subtiddlers as well.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -977,7 +977,7 @@ exports.getTextReferenceParserInfo = function(title,field,index,options) {
 		if(field === "title") {
 			parserInfo.sourceText = title;
 		} else if(tiddler && tiddler.fields) {
-			parserInfo.sourceText = tiddler.fields[field] ? tiddler.fields[field].toString() : null;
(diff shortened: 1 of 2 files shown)


pmario:
yea, it's hard to cover all code paths. .. I did test it and it works OK. 

I was testing my plugins if they are compatible with the new-json-store-area and found the problem.

Jermolene:
Thanks @saqimtiaz @pmario!

Issue #5830, opened 2021-06-27
[IDEA] Access to current date and time in filters.


saqimtiaz:
Currently there is no direct way to reference the current date and time in filters.
`<now>` returns the current date and time in `0hh:0mm, DDth MMM YYYY` format which makes it difficult to use with other operators.

The `format:date[]` operator supports formatting an input UTC date string (as used in the `created` and `modified` fields, i.e. `[UTC]YYYY0MM0DD0hh0mm0ssXXX`) according to a given template. If there was a macro or operator that provided the current date and time as a UTC date string, in combination with `format:date[]` this should be sufficient for most use cases.

Open to alternative ways to cater to the same use cases.

This came up for myself today when I needed to make sure an action was not fired more than once in a given period of time. There has also been a recent question on Discord by @sukima which involved checking if the current month was December or the 12th month. With access to the current date and time in `[UTC]YYYY0MM0DD0hh0mm0ssXXX` format this could be as simple as 
`[<now-timestamp>format:date[0MM]match[12]]`

Possible dupe of #5785 but I am not entirely sure what is being asked for there.

Jermolene:
Good idea. Perhaps just `[<timestamp>]` might be OK?

saqimtiaz:
@Jermolene Hah! `<timestamp>` is the working title I went with when I wrote a macro earlier today to use a stopgap until there was core support.

```
/*\
title: $:/core/modules/macros/timestamp.js
type: application/javascript
module-type: macro

Macro to return the current date and time as a [UTC]YYYY0MM0DD0hh0mm0ssXXX string.

\*/
(function(){

/*jslint node: true, browser: true */
/*global $tw: false */
"use strict";

/*
Information about this macro
*/

exports.name = "timestamp";
exports.params = [];

/*
Run the macro
*/
exports.run = function() {
	return $tw.utils.formatDateString(new Date(),"[UTC]YYYY0MM0DD0hh0mm0ssXXX");
};

})();

```

I will prepare a PR when I get the chance.

Jermolene:
Hmmm just thinking aloud but one issue is that this looks a lot like the existing `<<now>>` macro but with a different default when there are no parameters so that it can be used more easily in filters.

Two alternative approaches:

[...]

PR #5836, opened 2021-06-29
Support for macro params in filter operands

saqimtiaz:
@Jermolene this adds support for macro parameters in filters as discussed in #5830. 

Salient changes:
- refactored `$tw.utils.parseMacroInvocation()` to create `$tw.utils.parseMacroParameters()` which can be re-used for parsing macro parameters.
  - We have existing tests for parsing macros that show this is change is backwards compatible. 
- Introduced `$tw.utils.parseFilterVariable()` for which I have added tests.

### DIFF
--- a/core/modules/parsers/parseutils.js
+++ b/core/modules/parsers/parseutils.js
@@ -114,7 +114,7 @@ exports.parseStringLiteral = function(source,pos) {
 	var match = reString.exec(source);
 	if(match && match.index === pos) {
 		node.value = match[1] !== undefined ? match[1] :(
-			match[2] !== undefined ? match[2] : match[3] 
+			match[2] !== undefined ? match[2] : match[3]
 					);
 		node.end = pos + match[0].length;
 		return node;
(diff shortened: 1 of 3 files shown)


core/modules/parsers/parseutils.js
```diff
@@ -208,6 +215,29 @@ exports.parseMacroInvocation = function(source,pos) {
 	return node;
 };
 
+exports.parseFilterVariable = function(source) {
+	var node = {
+			name: "",
+			params: [],
+		},
+		pos = 0,
+		reName = /([^\s>"'=]+)/g;
```
Jermolene, line 224: Why do we use `=` as a terminator?

Jermolene (changes requested):
Thanks @saqimtiaz that was quick! Just a question, but also there's a bunch of whitespace changes which it would be great to remove for readability.

core/modules/parsers/parseutils.js
saqimtiaz: @Jermolene I re-used the regular expression from [parseMacroInvocation](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/parsers/parseutils.js#L174) and I spotted this but left it alone as I wasn't completely sure why we had it in that context either. Will revise.

saqimtiaz:
Apologies for the spurious white space changes, will resolve. Working on a different machine today and thus a new text editor, which has a mind of its own.

saqimtiaz:
- regular expression fixed.
- added more tests for edge cases.
- white space changes have been resolved (I really wonder why an editor would default to making whitespace changes unprompted!)

saqimtiaz:
Note that the following is also possible `[<now [UTC]YYYY0MM0DD0hh0mm0ssXXX>] `.

I've added a few tests specifically for square brackets in macro params.

pmario:
So this PR enables every macro to be used with params. right? Since it uses the macro parameter parser it's also possible to use named params?  eg: `[<now format:"[UTC]YYYY0MM0DD0hh0mm0ssXXX">] ` Is that possible?

saqimtiaz:
> So this PR enables every macro to be used with params. right? Since it uses the macro parameter parser it's also possible to use named params?  eg: `[<now format:"[UTC]YYYY0MM0DD0hh0mm0ssXXX">] ` Is that possible?

Yes. See tests.

Jermolene:
Thanks @saqimtiaz I think this might open up some interesting capabilities. It certainly makes it easier to bring custom JS code into filters.

PR #5837, opened 2021-06-29
Docs for macro parameters in filter operands

saqimtiaz:
I've used the term "literal parameters" to refer to parameters that are not variables or text references. Is there a different term we use for this in formal documentation? "String" felt too developer-centric and I wasn't sure if "hard" parameters would be understood.


Jermolene:
Thanks @saqimtiaz

> I've used the term "literal parameters" to refer to parameters that are not variables or text references. Is there a different term we use for this in formal documentation? "String" felt too developer-centric and I wasn't sure if "hard" parameters would be understood.

"Literal parameters" seems a good choice. The existing documentation manages to avoid the term, but I think it's good to establish concrete terms for wikitext phenomena – perhaps we should retrofit it into the docs for widget parameters?

PR #5840, opened 2021-06-30
Use ascii escaping instead of unicode in new json store area

simonbaird:
Includes also a commit to fix some bad indenting.

### DIFF
--- a/core/modules/widgets/jsontiddler.js
+++ b/core/modules/widgets/jsontiddler.js
@@ -49,7 +49,7 @@ JSONTiddlerWidget.prototype.render = function(parent,nextSibling) {
 	var json = JSON.stringify(fields);
 	// Escape unsafe script characters
 	if(this.attEscapeUnsafeScriptChars) {
-		json = json.replace(/</g,"\\u003C");
+		json = json.replace(/</g,"\\x3C");
 	}
 	// Update the DOM
 	var textNode = this.document.createTextNode(json);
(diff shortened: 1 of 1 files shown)


simonbaird:
Actually it seems like it doesn't work.. So maybe this is no good after all.

simonbaird:
If I run this with bin/serve.sh I get this in the browser:

    Uncaught SyntaxError: Unexpected token x in JSON at position 439

So I suppose it doesn't work like I expected it would.

I probably won't spend more time on this, but I'll leave the PR open for a bit. Maybe there's an obvious reason it doesn't work. Any one have an idea?

Feel free to cherry pick the whitespace fix if you want, or lmk if you want that in a separate PR.

pmario:
Simon. Did you see Jeremy's comment. https://github.com/Jermolene/TiddlyWiki5/pull/5708#issuecomment-871669306 It seems the JSON spec doesn't allow this

simonbaird:
> Simon. Did you see Jeremy's comment. [#5708 (comment)](https://github.com/Jermolene/TiddlyWiki5/pull/5708#issuecomment-871669306) It seems the JSON spec doesn't allow this

Yeah, I did. I was a little premature with the commit and MR!

I did something this and felt like it was certain to work.. :S
````
> console.log('\x3C\u003c', '\x3C' == '\u003c')
<< true
````

Oh well...

simonbaird:
Should have tried this:

````
> JSON.parse('"\\u003C"')
'<'
> JSON.parse('"\\x3C"')
Uncaught SyntaxError: Unexpected token x in JSON at position 2
> 
````

PR #5846, opened 2021-07-03
Add optional KaTeX support to markdown plugin

### LINKED ISSUE #2984: LaTeX plugin does not work in markdown tiddlers
The Katex and Markdown plugins don't work with one another. Most of my wiki ks in markdown and it is annoying to port individual docs into the Tiddly syntax to get equations to display.

Edit: [waylonflinn/markdown-it-katex](https://github.com/waylonflinn/markdown-it-katex) adds Katex support to Markdown, but I don't know if that's helpful.

rryan:
Motivation: The tiddlywiki/katex plugin is useful for including math notation in tiddlers, but it is not yet supported in markdown tiddlers.

This PR uses the remarkable-katex plugin 1.1.8 by Brad Howes to enable KaTeX support if the tiddlywiki/katex plugin is installed. Fixes #2984. remarkable-katex is distributed under the MIT license. I have included a copy of the license with attribution to Brad Howes in the tree and plugin description.

TESTED:

Created a test wiki with:
```
$ node tiddlywiki.js test --init markdowndemo
$ node tiddlywiki.js test --listen
```

* Verified markdown support works without the tiddlywiki/katex plugin enabled.

Screen Shot 2021-07-03 at 2 03 58 PM

* Verified markdown support works with the tiddlywiki/katex plugin enabled.
* Verified KaTeX (both inline and blocks) work as expected when the
tiddlywiki/katex plugin is enabled.

Screen Shot 2021-07-03 at 2 05 39 PM

### DIFF
--- a/plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
+++ b/plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2017 Brad Howes
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/plugins/tiddlywiki/markdown/files/remarkable-katex.js
+++ b/plugins/tiddlywiki/markdown/files/remarkable-katex.js
@@ -1,0 +1,20 @@
+"use strict";const rkatex=(md,options)=>{const backslash='\\';const dollar='$';const opts=options||{};const delimiter=opts.delimiter||dollar;if(delimiter.length!==1){throw new Error('invalid delimiter');}
+const katex=require("katex");const renderKatex=(source,displayMode)=>katex.renderToString(source,{displayMode:displayMode,throwOnError:false});const parseBlockKatex=(state,startLine,endLine)=>{let haveEndMarker=false;let pos=state.bMarks[startLine]+state.tShift[startLine];let max=state.eMarks[startLine];if(pos+1>max){return false;}
+const marker=state.src.charAt(pos);if(marker!==delimiter){return false;}
+let mem=pos;pos=state.skipChars(pos,marker);let len=pos-mem;if(len!==2){return false;}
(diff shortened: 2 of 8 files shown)


rryan:
Have not signed the CLA yet, working on it.

pmario:
I think there is a little problem with the naming conventions for plugins. ... But I'll have to have a closer look to be more accurate

plugins/tiddlywiki/markdown/files/remarkable-katex.js
```diff
@@ -0,0 +1,20 @@
+"use strict";const rkatex=(md,options)=>{const backslash='\\';const dollar='$';const opts=options||{};const delimiter=opts.delimiter||dollar;if(delimiter.length!==1){throw new Error('invalid delimiter');}
```
pmario, line 1: I think it would be good, to add a link to the license here (above line 1 ), so it will be included into the plugin and the github repo will be referenced, so we can easily find it. 

/*
License see: https://github.com/bradhowes/remarkable-katex/blob/master/LICENSE
*/

You can leave the license text from above in the repo. So everything will be fine.

pmario: Or you could just link to the repo..

plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
pmario, line 1: IMO we should add here: for: `remarkable-katex.js`

pmario: Or better: Add the reference to this plugin to the $:/plugins/tiddlywiki/markdown/readme  tiddler.

plugins/tiddlywiki/markdown/files/remarkable-katex.js
Jermolene: Ideally we do not modify source files bought in from external projects. Instead, we hack them via the tiddlywiki.files file, as below.

With other plugins that incorporate 3rd party libraries (eg JSZip) we do tend to expose the underlying license as a tiddler.

Jermolene (commented):
Thanks @rryan looks good if we can just include the license file.

plugins/tiddlywiki/markdown/files/remarkable-katex.js
rryan: Ahh ok -- one thing I did was to minify the JS from remarkable-katex. Should I include the un-minified JS instead?

plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
rryan: Mentioned the plugin usage in https://github.com/Jermolene/TiddlyWiki5/pull/5846/commits/fe013616f3ea8cdbc4621237aa75bde2fb461363

but since Remarkable's license wasn't mentioned there I didn't mention remarkable-katex's license either.  Did you mean I should mention the license there?

plugins/tiddlywiki/markdown/files/remarkable-katex.js
Jermolene: In that situation I would suggest including the unminified JS  in the files directory, but not pulling it in via `tiddlywiki.files`

plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
Jermolene: It would be great to bring in Remarkable's license, it's an oversight that it wasn't already included

rryan:
Included the license at `$:/plugins/tiddlywiki/markdown/remarkable-katex-license`:

Screen Shot 2021-07-04 at 11 32 38 AM

plugins/tiddlywiki/markdown/files/remarkable-katex-license.txt
rryan: Sure thing: https://github.com/Jermolene/TiddlyWiki5/pull/5846/commits/8b31c8567f9278206485708ca6babad1d3239adb

plugins/tiddlywiki/markdown/files/remarkable-katex.js
rryan: SG -- added the original and renamed the minified version.

Jermolene:
Thank you @rryan, great job!

PR #5848, opened 2021-07-04
markdown: Don't emit paragraph tags when a paragraph is "tight".

rryan:
Motivation: Since the upgrade to remarkable.js (#3876), lists are rendered as
HTML like this:

```
<ul>
  <li><p>One</p></li>
  <li><p>Two</p></li>
  <li><p>Three</p></li>
</ul>
```

Screen Shot 2021-07-03 at 2 43 01 PM

The paragraph nodes insert blocks that break the visual flow of the list and are
unexpected e.g. compared to WikiText markup's rendering of a bulleted list.

Solution: remarkable.js annotates certain paragraph nodes as "tight", and in the
bulleted list case, the paragraph nodes wrapping the text of each list item are
marked tight.

remarkable uses the tight property to [elide paragraph tags in its
renderer](https://github.com/jonschlinkert/remarkable/blob/58b6945f203ca7a0bb5a0785df90a3a6a8b9e59c/lib/rules.js#L136-L142).

This change implements the equivalent logic in TiddlyWiki's markdown rendering:
If a paragraph is marked tight, then we elide the `<p>` tag wrapping its children.

With this PR:

Screen Shot 2021-07-03 at 10 36 44 PM

### DIFF
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -94,7 +94,14 @@ function convertNodes(remarkableTree, isStartOfInline) {
 		var currentNode = remarkableTree[i];
 		switch (currentNode.type) {
 		case "paragraph_open":
-			i = wrappedElement("p", i, currentNode.level, "paragraph_close", remarkableTree);
+			// If the paragraph is a "tight" layout paragraph, don't wrap children in a <p> tag.
+			if (currentNode.tight) {
+				i = withChildren(i, currentNode.level, "paragraph_close", remarkableTree, function(children) {
+					out.push(...children);
+				});
+			} else {
+				i = wrappedElement("p", i, currentNode.level, "paragraph_close", remarkableTree);
+			}
 			break;
 
 		case "heading_open":
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -97,7 +97,7 @@ function convertNodes(remarkableTree, isStartOfInline) {
 			// If the paragraph is a "tight" layout paragraph, don't wrap children in a <p> tag.
 			if (currentNode.tight) {
 				i = withChildren(i, currentNode.level, "paragraph_close", remarkableTree, function(children) {
-					out.push(...children);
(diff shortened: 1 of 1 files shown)


plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -94,7 +94,14 @@ function convertNodes(remarkableTree, isStartOfInline) {
 		var currentNode = remarkableTree[i];
 		switch (currentNode.type) {
 		case "paragraph_open":
-			i = wrappedElement("p", i, currentNode.level, "paragraph_close", remarkableTree);
+			// If the paragraph is a "tight" layout paragraph, don't wrap children in a <p> tag.
+			if (currentNode.tight) {
+				i = withChildren(i, currentNode.level, "paragraph_close", remarkableTree, function(children) {
+					out.push(...children);
```
rryan, line 100: don't know whether the spread operator is allowed :)

pmario: No ... spread operator isn't allowed. TW core only allows ES5 atm.

Jermolene: Yes sorry about that, it's a nuisance but users find it useful to be able to run on older JS platforms.

Jermolene:
Thanks @rryan, there's a been lot of complaints about this so the fix is appreciated.

plugins/tiddlywiki/markdown/wrapper.js
rryan: Gotcha -- replaced with `Array.prototype.push`.

rryan:
Would you prefer I amend and force-push the existing single commit with review comments or should I add additional commits?

Jermolene:
> Would you prefer I amend and force-push the existing single commit with review comments or should I add additional commits?

I normally squash PRs when I merge them, so I think it shouldn't matter, but presumably just adding more commits is easiest?

rryan:
> I normally squash PRs when I merge them, so I think it shouldn't matter, but presumably just adding more commits is easiest?

Sounds good!

Jermolene:
Many thanks @rryan

PR #5849, opened 2021-07-04
Fix color string lookup that can potentiallly crash TW

cdruan:
I was working on a color filter and during testing, I accidentally pulled in all the tiddler titles in the wiki. Just so happened, I have a tiddler titled "Constructor", and the filter execution crashed with error "kCSSColorTable[str].slice is not a function".

So I replaced "in" operator with "hasOwnProperty()" in the color string lookup section of the code.

### DIFF
--- a/core/modules/utils/dom/csscolorparser.js
+++ b/core/modules/utils/dom/csscolorparser.js
@@ -133,7 +133,7 @@ function parseCSSColor(css_str) {
   var str = css_str.replace(/ /g, '').toLowerCase();
 
   // Color keywords (and transparent) lookup.
-  if (str in kCSSColorTable) return kCSSColorTable[str].slice();  // dup.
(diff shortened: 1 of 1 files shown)


Jermolene:
Good catch thanks @cdruan. The cssparser is taken from https://github.com/deanm/css-color-parser-js, so it would be good to apply the fix upstream if possible. I'm actually keen to adopt a newer, better colour parser, but the blocker for me is that none seem to yet support [P3 wide colours](https://css-tricks.com/wide-gamut-color-in-css-with-display-p3/)

Jermolene:
...or can we fix it by sanitising the data that we pass to the parser?

cdruan:
Hi @Jermolene 

Thanks for the explanation. The upstream project seems inactive at this point. Would it be prudent to fix potential bugs until a replacement is in place? Otherwise, I'll close this PR so not to clutter the PR pool.

pmario:
@Jermolene ... Since the upstream repo is inactive, IMO we should merge it here.

PR #5279, opened 2020-12-14
Add server sent events

NicolasPetton:
This is a follow-up to the original PR by @Arlen22 here: https://github.com/Jermolene/TiddlyWiki5/pull/4609.

---
#### Original PR description

This is related to Issue #4608 and also fixes Issue #4605.

To test this, run node tiddlywiki.js editions/server --listen then open two pages side-by-side and watch changes propagate in almost real time.

PR #5853, opened 2021-07-06
Improvements to SSE

Arlen22:
The main change to the server is that it keeps the last 5 minutes of events, and writes them to the client immediately when it reconnects if the client includes a last-event-id header (which EventSource does automatically). Obviously we have to debounce that for tiddlyweb, but it's part of the spec and makes the server properly usable for other plugins. If it cannot find the last-event-id that the client specified it returns a 409 Conflict, which cause the EventSource to throw an error and close. 

EventSource is the browser class that calls the server to recieve server-sent events.

The main change in the client is that whenever the EventSource throws an error, if the EventSource has closed, the client will now call syncFromServer to get the latest tiddlers when it successfully reopens the connection. The EventSource also throws an error when reconnecting automatically, but the readyState is set to CONNECTING so I just ignore it.

I also renamed the server parameter `sse-enabled` to `tiddlyweb-sse-enabled` since it is tiddlyweb specific and has no effect unless the tiddlyweb plugin is active. I renamed it in the status JSON as well. 

I compiled the server-sent-events.js file from Typescript and it didn't make sense not to include the source code. I'm not sure what formatting changes are needed to the JS file. Obviously the eval code surrounds it with a self-executing function so doing that in the file isn't required. If you reallly don't want the typescript file feel free to not include it.


pmario:
>I also renamed the server parameter sse-enabled to tiddlyweb-sse-enabled since it is tiddlyweb specific and has no effect unless the tiddlyweb plugin is active. I renamed it in the status JSON as well.

IMO instead of renaming, the sse-enabled function could check if all requirements are fulfilled and throw an error message or a warning if not. For me adding tiddlyweb to the name of the parameter doesn't indicate that it depends on a plugin. I expect it to work, if the parameter is set. ... If it doesn't work, it should tell me why.

Arlen22:
@Jermolene still working on this. Don't merge yet.

pmario:
You can make the PR a draft so Jeremy will see it but can't merge.

Arlen22:
I moved the sse-client.js file into the tiddlyweb adapter class. I don't know why I had it separate, but it works properly like this. It gets activated when the status JSON returns with `tiddlyweb_sse_enabled`. 

I also fixed up the code that writes the events and made sure that every main case is getting all the correct stuff. I added an initial record entry so clients reconnecting after a server restart immediately get an event id. 

There is also code in the Journal class for a second route (the emitter* functions) for clients to push events to the server to be broadcast, but this is not active in the tiddlyweb adapter. It is included however in case other plugins want to make use of that in some way. I did not test this much, but I did look over it and I expect it should work fine as it is very simple.

Arlen22:
I think it should start with tiddlyweb because it is specific to that protocol, not to all sse in general. Other plugins will have to individually choose to honor that variable as it is checked in the tiddlyweb plugin, not in core. I can still add a check for the plugin being enabled.

Arlen22:
Or are you saying the sse-enabled check should be global and be in the server-sent-events.js file?

Arlen22:
Actually, as far as the check, the listen command already checks for the tiddlyweb plugin and warns if it's not there, so unless I move the sse-enabled check to the server-sent-events.js file, that's already taken care of.

Arlen22:
Should I move the check to the server-sent-events Journal class in the core so it is a global flag?

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -102,6 +103,53 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
...
+		clearTimeout(timeout);
+		timeout = setTimeout(callback,$tw.syncer.throttleInterval);
+	};
+}
+
+TiddlyWebAdaptor.prototype.setupSSE = function (host, refresh) {
+	if (window.EventSource) {
+		var self = this;
+		if(this.eventsource && this.eventsource.readyState !== this.eventsource.CLOSED)
+			this.eventsource.close();
```
pmario, line 128: It should look like this: 

```
if(this.eventsource && this.eventsource.readyState !== this.eventsource.CLOSED) {
		this.eventsource.close();
	}
```

Otherwise it's super error prone if the source will ever be changed here.

plugins/tiddlywiki/tiddlyweb/sse-server.js
```diff
@@ -3,92 +3,70 @@ title: $:/plugins/tiddlywiki/tiddlyweb/sse-server.js
 type: application/javascript
 module-type: route
 
-GET /events/plugins/tiddlywiki/tiddlyweb
+GET /events/plugins/tiddlywiki/tiddlyweb/(channel)
 
 \*/
-(function(){
+(function () {
```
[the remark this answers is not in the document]
pmario: TW core uses tabs to indent, for file size reasons.

plugins/tiddlywiki/tiddlyweb/sse-server.js
```diff
@@ -3,92 +3,70 @@ title: $:/plugins/tiddlywiki/tiddlyweb/sse-server.js
...
+    function (request, response, state) {
+        if (state.server.get("tiddlyweb-sse-enabled") !== "yes"
+            || state.params[0] !== "wiki-change"
+        ) {
+            response.writeHead(404);
+            response.end();
+            return;
+        }
+        ensureChannelSetup(state.params[0], state.wiki);
+        eventServer.handler(request, response, state);
```
pmario, line 69: there should be no spaces between the parameters in function calls eg: `eventServer.handler(request,response,state);` ... That's valid for _all_ function calls

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -92,7 +92,8 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
 				self.isReadOnly = !!json["read_only"];
 				self.isAnonymous = !!json.anonymous;
 
-				var isSseEnabled = !!json.sse_enabled;
+				var isSseEnabled = !!json.tiddlyweb_sse_enabled && !!window.EventSource;
+				if(isSseEnabled) self.startSSE();
```
pmario, line 96: @Jermolene ... 1-liner?

plugins/tiddlywiki/tiddlyweb/sse-server.js
```diff
@@ -3,92 +3,70 @@ title: $:/plugins/tiddlywiki/tiddlyweb/sse-server.js
...
-		removeWikiConnection(state.wiki,connection);
-	});
+// this filter is called for the emitter route, which recieves 
+// messages from clients and forwards them to all listeners. It 
+// does not affect messages sent directly by the server. 
+// We don't use it in tiddlyweb so just set it to false
+eventServer.emitterFilter = function (sender) {
+    // do not allow clients to broadcast
+    // they can't anyway unless a route is specified
+    return function () { return false; };
```
pmario, line 52: `    return function() {return false;};`

pmario: line 49 .. no space between function and ()

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -102,6 +103,53 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
...
+			setTimeout(function () {
+				//call this function to set everything up again
+				self.setupSSE(host, true);
+			}, $tw.syncer.errorRetryInterval);
+		};
+		// call syncFromServer on open if necessary
+		if (refresh) source.onopen = function () {
+			//sync from server manually here to make sure we stay up to date
+			$tw.syncer.syncFromServer();
+		}
```
pmario, line 149: I think this construction is very error-prone. The closing brace belongs to the `function() {`, but it looks like if it would belong to the if statement. IMO the missing braces for the if should be added.

Jermolene:
Thanks @Arlen22. I'm sorry for any disappointment, but stepping back, I think these changes emphasise that the current SSE support is highly experimental and incomplete at this point. The trouble is that we're committed to backwards compatibility for anything that goes into the core, so there is a danger that we'll end up having to support features that we later come to regret.

At this point, I think the right thing to do is to completely revert all the current SSE commits for this release. What I'd like to do instead is add any necessary hooks so that these experiments can be done in plugins. I'd welcome other views @saqimtiaz @pmario

pmario:
TLDR; I'm already relying on that feature. I would be OK with a plugin. 

If it wouldn't be part of the next version at all, this would be a drastic setback for the German docs translation.

I didn't have a closer look at the code. I don't know if it's possible to use it as a plugin. @Arlen22 may know this.

------- 

long version

I did test the sse feature as implemented now and it worked well. I didn't have a look at the code, because I didn't need to. 

I did setup a client server installation on my own server to be able to collaborate for the German translation of the TW documentation. See draft PR "German translation of the tiddlywiki documentation." #5855

The draft PR contains 490 changed tiddlers, that I got from a new contributor named Frank. I did get a single-file wiki and it was a complete pain to move all the tiddlers into new directories, so the content will be manageable. The "directory info" is completely missing in the single file wiki. ... 

I want to use SSE to do collaborative editing, while we are connected via Skype. It is much more fun, if we change content like that and we can get immediate feedback. 

SSE as it is, is not perfect, but for 2 users at the same time, it is good enough. I know, which tiddler I'm editing, because it's in my story river, and I can see the tiddler which he is editing, because it's the red one on the bottom of the screen. That's it.

pmario:
ATM I'm preparing to test Arlen's code.

Jermolene:
Hi @pmario thank you, more testing will be helpful.

The real-time update propagation is clearly very desirable functionality; the issue is whether this implementation is sustainable for the long term.

pmario:
>The real-time update propagation is clearly very desirable functionality; the issue is whether this implementation is sustainable for the long term.

I do understand your concerns and support them. 

I think, if this functionality could be made a plugin, it would also be a good example for new plugins, that want to use the mechanism. ... 

I think, SSE is much easier to handle as web-sockets and it still can be the first stage to something the community desperately needs. Collaborative editing for smaller teams.

core/modules/server/server-sent-events.js
```diff
@@ -1,70 +1,224 @@
...
+        this.records = {};
+        this.responseHeaders = {};
+        this.repeaterFilter = function (conn) { return function (conn) { return true; }; };
+    }
+    Journal.prototype.cleanJournal = function (ts, channel) {
+        var maxage = ts - this.JOURNALAGE;
+        while (this.records[channel][0].Timestamp < maxage)
+            this.records[channel].shift();
+    };
+    Journal.prototype.initjournal = function (key) {
```
pmario, line 21: IMO it should be `initJournal` for consistency reasons

pmario:
@Arlen22 ... The problem I do have with your js-code is, that it contains a lot of unmaintainable code snippets, without the TS->JS transcoding stage. eg: 

```typescript
    initjournal(key: string) {
        if (!this.connections[key]) this.connections[key] = [];
        if (!this.records[key]) this.records[key] = [new JournalRecord("", "", 0, Date.now())];
        if (!this.entryIDs[key]) this.entryIDs[key] = 1;
    }
```

```javascript
    Journal.prototype.initjournal = function (key) {
        if (!this.connections[key])
            this.connections[key] = [];
        if (!this.records[key])
            this.records[key] = [new JournalRecord("", "", 0, Date.now())];
        if (!this.entryIDs[key])
            this.entryIDs[key] = 1;
    };
```

As you can see it adds a linebreak between the if-clause and the if-body text, without any curly-braces. .. Since a 1-liner would be OK in JS too, a 2-liner is extremely error-prone when body content, or an additional linebreak is added in the future. 

So from my point of view the JS-code is unmaintainable as it is at the moment.

Arlen22:
I found an easy solution for the braces problem (just add them in the Typescript source). I also figured out where my auto formatter settings were and reconfigured them. Hopefully I got it all correct. Does this look any better?

Arlen22:
The server-sent-events.js file is a very important piece and I think that should be in core once we know it's stable, because it is just a generic SSE server that all plugins can create their own instance of and mount under their own path. The rest of it can all be in a separate plugin, in which case we would not need the server variable either, because the plugin is specified manually. 

If we do add this line to the syncer.js file, that would make the rest of it very easy to do as a separate plugin. But we definitely should have the server-sent-events.js file in core at some point, because it's not plugin specific. 

https://github.com/Jermolene/TiddlyWiki5/blob/23fec9e390115c2985e59d2ec1bf6d3cbf5e9a88/core/modules/syncer.js#L271

                $tw.hooks.invokeHook("th-syncadaptor-status-response", Array.prototype.slice.apply(arguments));

Not sure if that's the right name for the hook, but that pretty much describes it. It will fire whenever the status is recieved in the syncer, which is good because it is called several places, such as login and logout, when you need to reload the connection anyway. So I think it makes sense to fire any time the status is recieved.

Arlen22:
Alright, 
- I moved the tiddlyweb related code into a plugin. 
- I added the hook to syncer.js right after the status tiddlers are set. 
- I renamed the SyncDisablePolling tiddler to be under `$:/status` instead of `$:/config` so it doesn't get saved to the server. 
- I removed the server variable since the plugin is specified manually for each datafolder. I also removed the docs. 
- I also removed it from the status and just set tiddlyweb to return false for isPollingDisabled. Other sync adaptors can return true if they like based on status or whatever, so I believe this is a really good idea which should be kept in.
- The plugin will not set SyncPollingDisabled to yes until it successfully connects to the SSE endpoint. Reloading the page should revert it to default until it successfully connects again, because it is a status tiddler rather than config. 
- I left the server-sent-events.js file in place for now to demonstrate the separation of concerns as I see it. And also because I believe the feature set in that file is basically stable, and I haven't needed to change anything more with it.

Arlen22:
So basically it now looks like this. If someone wants SSE enabled for the TiddlyWeb sync adaptor, they enable the tiddlyweb-sse plugin. The plugin has all the setup code for both the server and client. 

The server part consists of a route module which instantiates the Journal class, which handles all the business logic. Obviously many plugins can create instances of this class for their own purposes. 

The client part consists of a startup module and hooks into the syncer status response. It then attemps to connect to the SSE endpoint and disables polling on the client if successful by setting SyncDisablePolling to "yes". 

I change the hook to just pass in the Syncer instance and the status `!err` since everything is done on the wiki set on that instance, so anyone wanting to examine the syncer would really need to know which instance or at least which wiki the syncer was for. So syncer just made the most sense, in case there's more than one per wiki or who knows what else.

pmario:
I did review the code. It looks good to me. I'll test the code a bit later today. ... I did find a problem yesterday, that was already there with the old code and should be addressed now. A bit more testing is needed

Jermolene:
Thanks for your hard work and patience @Arlen22, I'll review in detail shortly. I'm not sure that the plugin should be a core plugin, simply because they have almost the same backwards compatibility constraints as the core itself (the difference is that we can deprecate a core plugin and replace it with a new one, as long as we retain support for the old one). The big advantage of not being a core plugin is that the releases won't have to be tied to the core releases.

plugins/tiddlywiki/tiddlyweb-sse/sse-client.js
```diff
@@ -0,0 +1,73 @@
...
+	$tw.hooks.addHook("th-syncer-status-response",function(syncer,success) {
+		// technically any syncer instance can be used, but we only use the global one for this
+		if(!success || syncer.syncadaptor.name !== "tiddlyweb" || $tw.syncer !== syncer) { return; }
+		// check if we have a previous one and close it if we do
+		if(source && source.readyState !== source.CLOSED) { source.close(); }
+		// Get the mount point in case a path prefix is used
+		var host = syncer.syncadaptor.getHost();
+		// Make sure it ends with a slash (it usually does)
+		if(host[host.length - 1] !== "/") { host += "/"; }
+		var endpoint = host + "events/plugins/tiddlywiki/tiddlyweb-sse/wiki-change";
```
pmario, line 31: IMO it should be `events/plugins/tiddlywiki/tiddlyweb-sse/wiki-changed` see the "d" at the end

pmario:
@Arlen22 here's the workflow issue I did fine, which imo isn't directly related to the PR. To reproduce. 

 - Open 2 windows with SSE enabled
 - I call them session: "left" and "right"
 - Open a "new tiddler" in the "right" session
 - The draft-tiddler will show up at the bottom of the "left" window
 - Click the red button, to open the draft-tiddler
 - Now save the tiddler on the "right" window. 
 - The draft-tiddler in the left window will be a "ghost", that can only be closed if the window is reloaded. 

As I wrote, that wasn't introduced with your PR. It has be there since the beginning of SSE

pmario:
I think the code that needs to be modified is at: https://github.com/Jermolene/TiddlyWiki5/blob/d373a187b865b9cdc241de728530f36085709706/core/modules/syncer.js#L335

IMO the behaviour should be: 

 - Remove the "Draft of " tiddler from the story-river and
 - replace it with the "New Tiddler".

The only problem I see is, if the tiddler has been renamed. ... I'm not sure if the "draft of" should be replaced with the renamed or only be removed. ... ?!

Arlen22:
@Jermolene I'm definitely fine with it not being a core plugin. I'd hate to retain whatever backward compatibility that plugin pretends to require. :)

@pmario I remember that this was maybe supposed to be fixed recently, but I'm not sure. The plugin calls syncFromServer, which is the method called for polling as well, so I think maybe it's a problem with the syncer or syncadaptor. But I thought Jeremy had gotten that fixed. I'll take a look at this. 

There is another feature called getUpdatedTiddlers, which would be more useful in an SSE context. It would be more realistic to have both of those and use one for manual refresh and the other for auto updates.

I also just remembered that I have no filtering on the server side yet either, so all changes get fed to the client directly even if they are not in the skinny tiddler list. So I'll need to get that yet. 

I would be fine with hosting the plugin on my GitHub account for now.

pmario:
@Arlen22 ... I did track it down already. ... When the "Draft tiddler" is removed from the server the event is propagated to the client and the client also deletes the tiddler. That's fine. 

The whole mechanism triggers a "change" event. The story.js receives the event, but it doesn't do anything, because on a normal wiki the "remove from story-river" is done by navigator.js. .... 

So IMO the easiest way, would be to modify the "change" handler in story.js to also remove "Draft of" tiddlers, from every storyriver that may be active.

pmario:
The 3rd possibility would be a new "hook" in the syncer, that will allow the plugin to deal with the situation.

pmario:
@Arlen22 ... If you have a closer look at the Dev-Console, you'll see that there is a periodic "wiki-change" event, that also triggers a "get-skinny-tiddlers" event. Even if nothing happens. 

IMO this shouldn't be the case. On my test system the skinny tiddler json is 35kByte, that just eat bandwidth.

Arlen22:
> It seems to come from: /tiddlyweb-sse/sse-client.js:48:16

That line is the debouncer, which means there is likely a change happening every two minutes. Can you click on the wiki-change line to see if there are any events. It's possibly a tiddler being saved every two minutes.

Arlen22:
> The whole mechanism triggers a "change" event. The story.js receives the event, but it doesn't do anything, because on a normal wiki the "remove from story-river" is done by navigator.js. ....

Whatever is happening here should also be happening with the skinny tiddler poll that happens every minute without this plugin, because this plugin callls the exact same method that is called every minute to do the poll, which is `syncer.syncFromServer`.

pmario:
>That line is the debouncer, which means there is likely a change happening every two minutes. Click on the wiki-change line in the network tab and click the event-stream tab in the detail. It will show you all the events that get emitted. It's possible a tiddler is being saved every two minutes.

There is _no_ change happening! There are no details - it's empty. I can reproduce it with FF and Edge the same problem. 
As long you do something with the wiki, you won't see it. You have to wait 2 minutes, doing nothing

pmario:
If I double-click it in Edge it says

```
id: 162575343374311
retry: 5000
```

But it doesn't finish.

pmario:
The Edge console says: 

```
:8080/events/plugins/tiddlywiki/tiddlyweb-sse/wiki-change:1 Failed to load resource: net::ERR_INCOMPLETE_CHUNKED_ENCODING
```

pmario:
Is it possible, that it messes some newlines?

pmario:
https://www.geeksforgeeks.org/node-js-http-server-timeout-property/ that's probably the problem. ... The default http module has a built in timeout of 2 minutes .. It seems?!

Arlen22:
Ok, not sure what that is, but I put a ten second setTimeout that closes the response to simulate the normal disconnection and reconnection and it does not do that for me. The chunked error is one that has something to do with the re-use of connections in HTTP/1.1. I ran into it the other day and I think it had something to do with making sure I immediately sent content or something like that.

Arlen22:
I moved all the files into a plugin for the moment, but did not edit any names. If you're using a datafolder to test this you do still need to put edit the tiddlywiki.info file, I think.

Arlen22:
There are some changes which do need to be made to core, including `isPollingDisabled` and the status hook, so I didn't revert those. I removed the server variable from the code and docs.

pmario:
@Arlen22 ... I did dig it up. All of them. .. I'm using nodejs 12.x where the built-in timeout is 120 sec. See: https://nodejs.org/api/http.html#http_server_settimeout_msecs_callback -> open the History "Details" tag. 

You can see, they changed this timeout to 0 in node 13. ... So if you use a later node version you don't see the problem.

pmario:
>net::ERR_INCOMPLETE_CHUNKED_ENCODING

Comes from the header that is used for SSEs. ... The SSE response header has `Transfer-Encoding: chunked` in the header, which is OK, since it is an open ended stream. It would need a slightly different encoding as we use. see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

But I don't think we should change it. .. I actually don't care too much as I know where it comes from. ... I'll only try to fix it as an experiment, just to see if I did understand it right.

pmario:
I did create a `th-server-command-post-start` hook,  that checks the active timeout and sets it to 0 if it is 120seconds. ... So the timeout won't happen anymore. .. 

There is only 1 new problem with https://nodejs.org/api/http.html#http_server_requesttimeout which must be set for node-version >14 ... if the server is directly connected to the web.

Arlen22:
Reconnections are part of the SSE protocol and should always be expected. If they are causing  a refresh then that means something is wrong. If you can hover over the line in the list, not in the detail, it should tell you if there was an error. 

Make sure you pull the latest changes just in case it's old code. But reconnections must be supported so I need to figure out why it's dong that. Can you give me anymore info about your setup? Are you running on localhost? And you have this happening in all browsers? They all run on the chrome engine now anyway, so probably not a big difference.

pmario:
>Reconnections are part of the SSE protocol and should always be expected. If they are causing a refresh then that means something is wrong. If you can hover over the line in the list, not in the detail, it should tell you if there was an error.

I know. 

>Make sure you pull the latest changes just in case it's old code. But reconnections must be supported so I need to figure out why it's dong that.

As I wrote, it was a problem from my nodejs version. I'm using 12.x ... Where there is a default timeout for http connections of 2 minutes. ... The client / browser does open a GET "wiki-change" connection, that shouldn't be closed ... But it was closed from the server. So the client initiated a reconnect. Starting from nodejs version 13 they did remove this 2minute timeout. 

So if you use node > 13 you don't have this problem. .. I'm testing with FF and Edge 

> Can you give me anymore info about your setup?

node 12.x, FF latest, Edge latest windows 10

>Are you running on localhost? 

Yes port 8080 .. ` node .\tiddlywiki.js .\editions\de-AT-server\ --listen gzip=yes use-browser-cache=yes`  atm. 

>And you have this happening in all browsers? 

Yes

>They all run on the chrome engine now anyway, so probably not a big difference.

:) My main browser is FF. So I'm using a different engine most of the time. .. But I do like Edge. I think they do have a good new chromium based browser now.

pmario:
@Arlen22 .. do you have a plugin repo already? Can you provide a link?

Arlen22:
Here is the repo. I already have a github.io website for the twcloud org, but I don't know what I should specifically do for this plugin to make it available to the Tiddlyverse. 

https://github.com/twcloud/tiddlyweb-sse

Jermolene:
Hi @Arlen22 apologies for all the work on this, but I think we need to make this two separate PRs in order to make ut easier for people to see what's going on:

1. First a PR that just reverts #5279. GitHub won't do that automatically because there have been subsequent changes, so this will need to be prepared manually
2. Another PR that adds the new hooks

Many thanks for your help

Arlen22:
> Hi @Arlen22 apologies for all the work on this, but I think we need to make this two separate PRs in order to make ut easier for people to see what's going on:
> 
> 1. First a PR that just reverts [Add server sent events #5279](https://github.com/Jermolene/TiddlyWiki5/pull/5279). GitHub won't do that automatically because there have been subsequent changes, so this will need to be prepared manually
> 2. Another PR that adds the new hooks
> 
> Many thanks for your help

Ok, I created both PRs, but the second one includes the commit from the first one so I'm not entirely sure how that works. It's up to you whether you want to merge it unsquashed or just merge the first one and then try the second one and if it doesn't work I'll rebase. 

Here's the second one, which contains both commits. https://github.com/Jermolene/TiddlyWiki5/pull/5881

Jermolene:
Thanks @Arlen22 much appreciated. Can this be closed?

Arlen22:
Alright I rebased after #5880 and opened #5882.

Arlen22:
> As I wrote, it was a problem from my nodejs version. I'm using 12.x ... Where there is a default timeout for http connections of 2 minutes. ... The client / browser does open a GET "wiki-change" connection, that shouldn't be closed ... But it was closed from the server. So the client initiated a reconnect. Starting from nodejs version 13 they did remove this 2minute timeout.

@pmario thank you. I was able to reproduce this with the same error in Node 14 by calling `request.setTimeout(10000)`, so I just added `request.setTimeout(0)` to the request handler in sse-server.js. 

> do you have a plugin repo already? Can you provide a link?

https://github.com/twcloud/tiddlyweb-sse

PR #5865, opened 2021-07-10
Add sidebar-slider to tiddlywiki core

BurningTreeC:
This PR adds the `sidebar-slider` like discussed in #5861


BurningTreeC:
This PR uses the `clamp()` CSS function heavily, which isn't supported in IE

saqimtiaz:
Some very quick thoughts off the top of my head that might be worth considering:
- can we have the eventcatcher only around the slider, or only listening to events from the slider?
  - It might be helpful here to consider using setPointerCapture (which we could expose via eventcatcher or a separate action-* widget)
- I was going to suggest using `user-select: none` but I see you did so already. 👍

BurningTreeC:
Hi @saqimtiaz,

> Some very quick thoughts off the top of my head that might be worth considering:
> 
> * can we have the eventcatcher only around the slider, or only listening to events from the slider?
>   
I tried that and it ended in the slider not working correctly. We will almost surely move the pointer out of the position of the slider while dragging and then it stops getting the `pointermove` event. So that's why I listen for the `pointermove` event on the pagecontainer div

>   * It might be helpful here to consider using setPointerCapture (which we could expose via eventcatcher or a separate action-* widget)

What exactly would `setPointerCapture` do?

> * I was going to suggest using `user-select: none` but I see you did so already. 👍

Thanks 😄

saqimtiaz:
> What exactly would `setPointerCapture` do?

It allows events to be targeted to a given element even if the pointer moves outside that element. The challenge is in finding an intuitive way to add a wikitext affordance for working with it. See https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

BurningTreeC:
> > What exactly would `setPointerCapture` do?
> 
> It allows events to be targeted to a given element even if the pointer moves outside that element. The challenge is in finding an intuitive way to add a wikitext affordance for working with it. See https://developer.mozilla.org/en-US/docs/Web/API/Element/setPointerCapture

Yes right @saqimtiaz , that would be the solution 👍

BurningTreeC:
Another thing I'd be interested to find a solution for is when the pointer leaves the window

In that case, I would need to trigger some actions, but I'm not able to get the `pointerleave` event and the `pointerout` event is triggered too often.

Is there a way to detect on which domNode the `pointerout` event was triggered and to dispatch some actions accordingly?

pmario:
I did just test it. Do you intend to change the tiddler values that are shown in the ControlPanel : Advanced : ThemeTweaks, so they reflect the settings of the slider.

BurningTreeC:
> I did just test it. Do you intend to change the tiddler values that are shown in the ControlPanel : Advanced : ThemeTweaks, so they reflect the settings of the slider.

Hi @pmario , no I didn't intend to change them since they're used for the limits (left/right) of the slider

BurningTreeC:
I have a problem in this PR with detecting when the pointer leaves the window. In that case I'd like to trigger some actions. But the `pointerout` event isn't very useful since it triggers too often (every time the pointer leaves the hit boundaries of any given element on the page). So there should be a way to detect when the pointer leaves the page...

Does anybody have any ideas?

pmario:
>I have a problem in this PR with detecting when the pointer leaves the window. In that case I'd like to trigger some actions. But the pointerout event isn't very useful since it triggers too often (every time the pointer leaves the hit boundaries of any given element on the page). So there should be a way to detect when the pointer leaves the page...

I think that's a general problem, that isn't solved. Most elements that can be dragged on other pages have that problem. ... I wouldn't invest too much time into that problem. 

Eg. Here at github you can make the input area bigger, but if the courser leaves the page and comes back the drag action is still going on. ... I don't see a problem in this behaviour

BurningTreeC:
> > I have a problem in this PR with detecting when the pointer leaves the window. In that case I'd like to trigger some actions. But the pointerout event isn't very useful since it triggers too often (every time the pointer leaves the hit boundaries of any given element on the page). So there should be a way to detect when the pointer leaves the page...
> 
> I think that's a general problem, that isn't solved. Most elements that can be dragged on other pages have that problem. ... I wouldn't invest too much time into that problem.
> 
> Eg. Here at github you can make the input area bigger, but if the cursor leaves the page and comes back the drag action is still going on. ... I don't see a problem in this behaviour

Ah ok, thanks for clarifying @pmario 👍

BurningTreeC:
@Jermolene - if this PR is considered for the core, I'd be in favor of finding a solution for the then obsolete fixed-fluid and fluid-fixed layouts

BurningTreeC:
This PR is now ready for a review 😺

pmario:
If I start it like this: ` tiddlywiki .\editions\tw5.com-server\ --listen` it doesn't work because your CSS settings are overwritten somewhere.

pmario:
```
.tc-sidebar-slider {

    left: calc(100% - 350px);
    left: clamp(770px, calc(px - px), calc(100% - 350px));
```

CSS looks like this, so  the second row is invalid. Seems the values are missing

core/ui/PageTemplate.tid
```diff
@@ -6,6 +6,9 @@ description: {{$:/language/PageTemplate/Description}}
 \define containerClasses()
 tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 \end
+\define actions-pointerdown() <$action-setfield $tiddler="$:/state/pointerdown" text="yes"/><$list filter="[{$:/state/dragging}match[yes]]"><$action-setfield $tiddler={{$:/state/dragging!!state-title}} event-fromselected-posx=<<event-fromselected-posx>> event-fromcatcher-posx=<<event-fromcatcher-posx>> event-fromselected-posy=<<event-fromselected-posy>> event-fromcatcher-posy=<<event-fromcatcher-posy>>/></$list>
+\define actions-pointerup() <$action-deletetiddler $tiddler="$:/state/dragging"/><$action-deletetiddler $tiddler="$:/state/pointerdown"/>
+\define actions-pointermove() <$list filter="[{$:/state/dragging}match[yes]]"><$list filter="[{$:/state/pointerdown}match[yes]]"><$action-setfield $tiddler={{$:/state/dragging!!state-title}} event-fromselected-posx=<<event-fromselected-posx>> event-fromcatcher-posx=<<event-fromcatcher-posx>> event-fromselected-posy=<<event-fromselected-posy>> event-fromcatcher-posy=<<event-fromcatcher-posy>>/></$list></$list>
```
pmario, line 11: It seems your state variable are missing, if you start from your PR

themes/tiddlywiki/vanilla/sidebar-slider.tid
```diff
@@ -0,0 +1,90 @@
...
+			top: 0;
+			height: 100%;
+			cursor: col-resize;
+		}
+
+		<$list filter="[{$:/themes/tiddlywiki/vanilla/options/sidebarlayout}match[fluid-fixed]]">
+
+			.tc-sidebar-slider {
+				left: calc(100% - {{$:/themes/tiddlywiki/vanilla/metrics/sidebarwidth}});
+				left: clamp({{$:/themes/tiddlywiki/vanilla/metrics/storywidth}}, calc({{$:/state/sidebar-slider!!event-fromcatcher-posx}}px - {{$:/state/sidebar-slider!!slider-event-fromselected-posx}}px), calc(100% - {{$:/themes/tiddlywiki/vanilla/metrics/sidebarwidth}}));
```
pmario, line 29: state vars are missing if they have never been initialized.

BurningTreeC: yes they are missing in the beginning, that's why the first line gets preferred over the invalid line

core/ui/PageTemplate.tid
BurningTreeC: Yes they are missing in the beginning but that should be no problem

BurningTreeC:
> ```
> .tc-sidebar-slider {
> 
>     left: calc(100% - 350px);
>     left: clamp(770px, calc(px - px), calc(100% - 350px));
> ```
> 
> CSS looks like this, so the second row is invalid. Seems the values are missing

When the second line is invalid, the first line is still valid and gets preferred. When we start using the slider then the second line becomes valid and gets preferred

BurningTreeC:
> If I start it like this: ` tiddlywiki .\editions\tw5.com-server\ --listen` it doesn't work because your CSS settings are overwritten somewhere.

Did you checkout my branch? Because there is missing the new eventcatcher syntax

pmario:
>Did you checkout my branch? Because there is missing the new eventcatcher syntax

I did checkout your patch-34 branch and imo that should do. Everything else is way to complicated.

pmario:
>Did you checkout my branch? Because there is missing the new eventcatcher syntax

How does this branch work on your machine?

BurningTreeC:
> > Did you checkout my branch? Because there is missing the new eventcatcher syntax
> 
> How does this branch work on your machine?

I had to merge master into the branch to get everything working

saqimtiaz:
@BurningTreeC I assume you are not using a commandline workflow. If so, go to this branch on github and fetch upstream changes:

image

BurningTreeC:
> @BurningTreeC I assume you are not using a commandline workflow. If so, go to this branch on github and fetch upstream changes:
> 
> image

Thanks @saqimtiaz - the commandline workflow is still an obscure thing for me 😄

pmario:
OK. Thx. Now it works as proposed

pmario:
But now we should probably move the width settings into the "hidden settings" area. ... Users have been confused by them anyway

BurningTreeC:
With this mechanism and a bit of work we could also build resizable divs or similar stuff:


https://user-images.githubusercontent.com/9407182/125416015-a6b8a058-4b8f-42c7-8899-23ef9946f2b2.mp4

BurningTreeC:
@Jermolene do you think this PR still needs some work to be ready for the core?

Jermolene:
Hi @BurningTreeC to go into the core I think we need to make it be a generic capability that authors can use in their own layouts, and not be tied to the current sidebar layout.

Jermolene:
But I should add that this is stunning stuff, and gets us a lot of the way there!

BurningTreeC:
Yes @Jermolene I already thought that you think so. I made the mechanism a bit generic so that it can be reused, just the CSS stuff is specific for the sidebar slider and the sidebar layout

BurningTreeC:
@Jermolene do you have an idea how to make this generic so that it can be part of the core?

Jermolene (commented):
> do you have an idea how to make this generic so that it can be part of the core?

It is definitely quite a challenge. You may want to explore making it as a plugin first, which would involve figuring out the required hooks. At the back of my mind is the idea that a more generic version of the `<$importvariables>` widget could be useful to pull in arbitrary chains of widgets into system templates, not just `<$set>` widgets. That way you could add components that are automatically wrapped around other components like the sidebar or story river. But equally, that's a bit of a research project, and you might want to just start by figuring out the minimal core tiddler overrides.

One aspect of this implementation that seems messy is the reliance on dynamically CSS, simply because that technique can be hard to scale to multiple instances of the component.

In terms of making this generic, it might be helpful to look at other UI component libraries and see how they expose this functionality. My guess is that it's easier to make a "splittable panel" component than a generic "splitter" component that can be applied to arbitrary other components.

core/ui/PageTemplate.tid
```diff
@@ -17,6 +20,8 @@ tc-page-container tc-page-view-$(storyviewTitle)$ tc-language-$(languageTitle)$
 	storyviewTitle={{$:/view}}
 	languageTitle={{{ [{$:/language}get[name]] }}}>
 
+<$eventcatcher selector=".tc-page-container" stopPropagation="never" events="pointerdown pointerup pointermove" $pointerdown=<<actions-pointerdown>> $pointerup=<<actions-pointerup>> $pointermove=<<actions-pointermove>>>
```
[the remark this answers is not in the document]
BurningTreeC: Oh, thanks @Jermolene - I missed that

saqimtiaz:
> At the back of my mind is the idea that a more generic version of the `<$importvariables>` widget could be useful to pull in arbitrary chains of widgets into system templates, not just `<$set>` widgets. That way you could add components that are automatically wrapped around other components like the sidebar or story river.

Such a feature would be very welcome (though it need not be tied to this PR). It is something I've wondered about a fair few times as a means to inject widgets into system templates without overriding shadow tiddlers and creating potential upgrade issues down the line.

BurningTreeC:
@Jermolene @saqimtiaz - I believe we can have a "splitter-panel" right now if we make it possible to assign inline styles to the DOM Node created by the `<$eventcatcher>` widget. I have a prototype which creates any number of flexbox panels which can be resized. Everything in wikitext. I'll post a gif

BurningTreeC:
My TiddlyWiki — a non-linear personal web notebook - Google Chrome 2021-07-17 08-39-10

PR #5867, opened 2021-07-11
Added counter-total variable to list widget

### LINKED ISSUE #5864: [IDEA] <counter-variable-name>-total variable in List widget
Complementing the variables introduced in the `<$list>` widget in #5611, it might be helpful to also introduce a variable `<counter-variable-name>-total` indicating the total number of list items. This would allow easily numbering lists as `x of total`.

Example:

```
<$list filter="[tag[About]sort[title]]" counter="counter">
<div>
<<counter>>/<<counter-total>>: ''<$text text=<<currentTiddler>>/>''
</div>
</$list>
```

saqimtiaz:
Fixes #5864

Adds a `<counter-variable-name>-total` variable when the List widget is used with the counter attribute.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -135,6 +135,7 @@ ListWidget.prototype.makeItemTemplate = function(title,index) {
 		parseTreeNode.counterName = this.counterName;
 		parseTreeNode.isFirst = index === 0;
 		parseTreeNode.isLast = index === this.list.length - 1;
+		parseTreeNode.counterTotal = this.list.length;
 	}
 	return parseTreeNode;
 };
@@ -341,6 +342,7 @@ ListItemWidget.prototype.execute = function() {
 		this.setVariable(this.parseTreeNode.counterName,this.parseTreeNode.counter);
 		this.setVariable(this.parseTreeNode.counterName + "-first",this.parseTreeNode.isFirst ? "yes" : "no");
 		this.setVariable(this.parseTreeNode.counterName + "-last",this.parseTreeNode.isLast ? "yes" : "no");
+		this.setVariable(this.parseTreeNode.counterName + "-total",this.parseTreeNode.counterTotal.toString());
 	}
 	// Construct the child widgets
 	this.makeChildWidgets();
--- a/editions/tw5.com/tiddlers/widgets/ListWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ListWidget.tid
@@ -1,6 +1,6 @@
 caption: list
 created: 20131024141900000
-modified: 20210416175333981
+modified: 20210711122847337
 tags: Widgets Lists
 title: ListWidget
 type: text/vnd.tiddlywiki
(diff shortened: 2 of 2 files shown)


Jermolene:
Hi @saqimtiaz I think there might be a problem in that when the list refreshes we reuse any unaffected, existing list item widgets, without updating the `-total` variable (or, come to think of it, the `-first`/`-last` variables).

https://github.com/Jermolene/TiddlyWiki5/blob/7204f442cdc84e25864f07f06934d50644967be7/core/modules/widgets/list.js#L225-L238

So, I think if the list gets longer on refresh, then the `-total` variable will only be accurate for the added entries. Annoyingly, all of this does severely limit our ability to optimise refreshes.

saqimtiaz:
@Jermolene aha! My mistake. I had not looked carefully at the revised refresh code and assumed we were removing and recreating all items when the counter variable was being set. 

I am not sure that the utility of this proposed `-total` variable outweighs the greater performance hit we would take in refreshes. However, the `-last` variable has the same problem. The `-first` variable should be OK since if the first item in the list changes, it and all subsequent items are recreated. 

To keep the `-last` variable updated and accurate, after removing extra items from the end of the list we would need to recreate the item left at the last position.

Something along these lines:

```javascript
// If there are items to remove then recreate the item that will now be at the last position
if(this.children.length > this.list.length && this.counterName){
	this.removeListItem(this.list.length-1);
	this.insertListItem(this.list.length-1,this.list[this.list.length-1]);
}
// Remove any left over items
for(t=this.children.length-1; t>=this.list.length; t--) {
	this.removeListItem(t);
	hasRefreshed = true;
}
```
Thoughts?

Jermolene:
Hi @saqimtiaz 

> @Jermolene aha! My mistake. I had not looked carefully at the revised refresh code and assumed we were removing and recreating all items when the counter variable was being set.
> 
> I am not sure that the utility of this proposed `-total` variable outweighs the greater performance hit we would take in refreshes. However, the `-last` variable has the same problem. The `-first` variable should be OK since if the first item in the list changes, it and all subsequent items are recreated.

Perhaps it's overkill, but we could perhaps make things simpler by having a separate attribute to specify the name of the total variable. Then we could more easily optimise for the situation where it is not being used. Conceivably the same holds for `-first` and `-last` 
 
> To keep the `-last` variable updated and accurate, after removing extra items from the end of the list we would need to recreate the item left at the last position.
> 
> Something along these lines:
> 
> ```js
> // If there are items to remove then recreate the item that will now be at the last position
> if(this.children.length > this.list.length && this.counterName){
> 	this.removeListItem(this.list.length-1);
> 	this.insertListItem(this.list.length-1,this.list[this.list.length-1]);
> }
> // Remove any left over items
> for(t=this.children.length-1; t>=this.list.length; t--) {
> 	this.removeListItem(t);
> 	hasRefreshed = true;
> }
> ```
> 
> Thoughts?

That looks right.

There are some tests for using the counter attribute with the list widget, and the tests allow for checking the refresh behaviouer, so it might be worth updating them.

pmario:
>Perhaps it's overkill, but we could perhaps make things simpler by having a separate attribute to specify the name of the total variable. Then we could more easily optimise for the situation where it is not being used. Conceivably the same holds for -first and -last

+1

PR #5869, opened 2021-07-11
Update MessageCatcher to simplify and unify syntax

saqimtiaz:
This PR removes the `type` and `actions` attribute for `$messagecatcher` resulting in a more consistent and easier to remember syntax, as [previously discussed](https://github.com/Jermolene/TiddlyWiki5/commit/f87b3bfcdba79b6ad198af286bd827c61044891f#commitcomment-53032635).

### DIFF
--- a/core/modules/widgets/messagecatcher.js
+++ b/core/modules/widgets/messagecatcher.js
@@ -63,9 +63,7 @@ MessageCatcherWidget.prototype.render = function(parent,nextSibling) {
 			);
 		}
 	}
-	// Add the main event handler
-	addEventHandler(this.getAttribute("type"),this.getAttribute("actions"));
-	// Add any other event handlers
+	// Add  event handlers
 	$tw.utils.each(this.attributes,function(value,key) {
 		if(key.charAt(0) === "$") {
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @saqimtiaz I am not sure about this. There are two arguments for the type/actions attributes:

* To make it possible to use a dynamically generated message type
* To use message types that can't be expressed as an attribute

The 2nd doesn't seem like a huge deal because I don't think there are any such messages types in the wild. But the first might be a consideration?

saqimtiaz:
> * To make it possible to use a dynamically generated message type

Yes this is a fair point. What I would suggest instead then is just rewording the docs a bit to suggest that the $message=actions syntax is the primary way to use the widget and explain in which circumstances the type/actions attributes are useful.

If that sounds reasonable I'll prepare a PR for it.

saqimtiaz:
In terms of refining the docs I am thinking of something along these lines:

image

Jermolene:
> In terms of refining the docs I am thinking of something along these lines:

That's great, thanks @saqimtiaz

PR #5882, opened 2021-07-14
Add syncer status hook and option to disable polling

Arlen22:
This adds another variable to the status response which tells the syncer whether the sync adapter expects it to poll for changes. This is useful if the syncadaptor has some other mechanism to get changes from the server, but still allows a manual refresh to be triggered. If this argument is not passed or is false, the default behavior will still happen, so it's backward compatible. 

**Update:** I removed the status response variable from the getStatus callback but kept the status tiddler. The status tiddler is checked every time to determine whether to set a timeout to schedule the next sync, meaning if the tiddler gets unset, syncFromServer must be called once to restart polling. The tiddler may also be set to yes at any time and polling will stop after the next call to syncFromServer (including by polling). 

It also adds a hook to the syncer which fires after the status callback is processed so other plugins can handle the status accordingly. Because the syncer puts everything into the wiki, and also because many syncer instances could theoretically exist, I chose to pass the syncer itself as the hook argument. The listener can then verify that the instance matches `$tw.syncer`.

The main reason I only pass a boolean for success is partly to prevent memory leakage and partly because it's the syncer's job to handle that stuff and just doesn't make sense to pass that out. However, a failure could still indicate something depending on the syncadaptor in use so I include it as a boolean. 

If the syncadaptor wants to reference the syncer, it can listen for the hook and then check if `syncer.syncadaptor` equals itself.

This was originally discussed in #5853 and #5279.

### DIFF
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -20,6 +20,7 @@ Syncer.prototype.titleIsAnonymous = "$:/status/IsAnonymous";
 Syncer.prototype.titleIsReadOnly = "$:/status/IsReadOnly";
 Syncer.prototype.titleUserName = "$:/status/UserName";
 Syncer.prototype.titleSyncFilter = "$:/config/SyncFilter";
+Syncer.prototype.titleSyncDisablePolling = "$:/status/SyncDisablePolling";
 Syncer.prototype.titleSyncPollingInterval = "$:/config/SyncPollingInterval";
 Syncer.prototype.titleSyncDisableLazyLoading = "$:/config/SyncDisableLazyLoading";
 Syncer.prototype.titleSavedNotification = "$:/language/Notifications/Save/Done";
@@ -89,7 +90,7 @@ function Syncer(options) {
 		if(filteredChanges.length > 0) {
 			self.processTaskQueue();
 		} else {
-			// Look for deletions of tiddlers we're already syncing	
+			// Look for deletions of tiddlers we're already syncing
 			var outstandingDeletion = false
 			$tw.utils.each(changes,function(change,title,object) {
 				if(change.deleted && $tw.utils.hop(self.tiddlerInfo,title)) {
@@ -121,7 +122,7 @@ function Syncer(options) {
 				self.login(username,password,function() {});
 			} else {
 				// No username and password, so we display a prompt
-				self.handleLoginEvent();				
+				self.handleLoginEvent();
 			}
 		});
 		$tw.rootWidget.addEventListener("tm-logout",function() {
@@ -138,7 +139,7 @@ function Syncer(options) {
 	if(!this.disableUI && this.wiki.getTiddlerText(this.titleSyncDisableLazyLoading) !== "yes") {
 		this.wiki.addEventListener("lazyLoad",function(title) {
 			self.handleLazyLoadEvent(title);
-		});		
+		});
 	}
 	// Get the login status
 	this.getStatus(function(err,isLoggedIn) {
@@ -173,8 +174,8 @@ Syncer.prototype.getTiddlerRevision = function(title) {
 	if(this.syncadaptor && this.syncadaptor.getTiddlerRevision) {
 		return this.syncadaptor.getTiddlerRevision(title);
 	} else {
-		return this.wiki.getTiddler(title).fields.revision;	
-	} 
+		return this.wiki.getTiddler(title).fields.revision;
+	}
 };
 
 /*
@@ -267,7 +268,7 @@ Syncer.prototype.getStatus = function(callback) {
 		// Mark us as not logged in
 		this.wiki.addTiddler({title: this.titleIsLoggedIn,text: "no"});
 		// Get login status
-		this.syncadaptor.getStatus(function(err,isLoggedIn,username,isReadOnly,isAnonymous) {
+		this.syncadaptor.getStatus(function(err,isLoggedIn,username,isReadOnly,isAnonymous,isPollingDisabled) {
 			if(err) {
 				self.logger.alert(err);
 			} else {
@@ -278,7 +279,11 @@ Syncer.prototype.getStatus = function(callback) {
 				if(isLoggedIn) {
 					self.wiki.addTiddler({title: self.titleUserName,text: username || ""});
 				}
+				if(isPollingDisabled) {
+					self.wiki.addTiddler({title: self.titleSyncDisablePolling, text: "yes"});
+				}
 			}
+			$tw.hooks.invokeHook("th-syncer-status-response",self,!err);
 			// Invoke the callback
 			if(callback) {
 				callback(err,isLoggedIn,username);
@@ -301,12 +306,15 @@ Syncer.prototype.syncFromServer = function() {
 			}
 		},
 		triggerNextSync = function() {
-			self.pollTimerId = setTimeout(function() {
-				self.pollTimerId = null;
-				self.syncFromServer.call(self);
-			},self.pollTimerInterval);
+			if(pollingEnabled) {
+				self.pollTimerId = setTimeout(function() {
+					self.pollTimerId = null;
+					self.syncFromServer.call(self);
+				},self.pollTimerInterval);
+			}
 		},
-		syncSystemFromServer = (self.wiki.getTiddlerText("$:/config/SyncSystemTiddlersFromServer") === "yes" ? true : false);
+		syncSystemFromServer = (self.wiki.getTiddlerText("$:/config/SyncSystemTiddlersFromServer") === "yes"),
+		pollingEnabled = (self.wiki.getTiddlerText(self.titleSyncDisablePolling) !== "yes");
 	if(this.syncadaptor && this.syncadaptor.getUpdatedTiddlers) {
 		this.logger.log("Retrieving updated tiddler list");
 		cancelNextSync();
@@ -329,7 +337,7 @@ Syncer.prototype.syncFromServer = function() {
 				});
 				if(updates.modifications.length > 0 || updates.deletions.length > 0) {
 					self.processTaskQueue();
-				}				
+				}
 			}
 		});
 	} else if(this.syncadaptor && this.syncadaptor.getSkinnyTiddlers) {
@@ -509,7 +517,7 @@ Syncer.prototype.processTaskQueue = function() {
 				} else {
 					self.updateDirtyStatus();
 					// Process the next task
-					self.processTaskQueue.call(self);					
+					self.processTaskQueue.call(self);
 				}
 			});
 		} else {
@@ -517,11 +525,11 @@ Syncer.prototype.processTaskQueue = function() {
 			this.updateDirtyStatus();
 			// And trigger a timeout if there is a pending task
 			if(task === true) {
-				this.triggerTimeout();				
+				this.triggerTimeout();
 			}
 		}
 	} else {
-		this.updateDirtyStatus();		
+		this.updateDirtyStatus();
 	}
 };
 
@@ -555,7 +563,7 @@ Syncer.prototype.chooseNextTask = function() {
 				isReadyToSave = !tiddlerInfo || !tiddlerInfo.timestampLastSaved || tiddlerInfo.timestampLastSaved < thresholdLastSaved;
 			if(hasChanged) {
 				if(isReadyToSave) {
-					return new SaveTiddlerTask(this,title); 					
+					return new SaveTiddlerTask(this,title);
 				} else {
 					havePending = true;
 				}
--- a/core/modules/syncer.js
+++ b/core/modules/syncer.js
@@ -268,7 +268,7 @@ Syncer.prototype.getStatus = function(callback) {
 		// Mark us as not logged in
 		this.wiki.addTiddler({title: this.titleIsLoggedIn,text: "no"});
 		// Get login status
-		this.syncadaptor.getStatus(function(err,isLoggedIn,username,isReadOnly,isAnonymous,isPollingDisabled) {
+		this.syncadaptor.getStatus(function(err,isLoggedIn,username,isReadOnly,isAnonymous) {
 			if(err) {
 				self.logger.alert(err);
 			} else {
@@ -278,9 +278,6 @@ Syncer.prototype.getStatus = function(callback) {
 				self.wiki.addTiddler({title: self.titleIsLoggedIn,text: isLoggedIn ? "yes" : "no"});
 				if(isLoggedIn) {
 					self.wiki.addTiddler({title: self.titleUserName,text: username || ""});
-				}
-				if(isPollingDisabled) {
-					self.wiki.addTiddler({title: self.titleSyncDisablePolling, text: "yes"});
 				}
 			}
 			$tw.hooks.invokeHook("th-syncer-status-response",self,!err);
--- a/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
+++ b/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
@@ -23,7 +23,6 @@ function TiddlyWebAdaptor(options) {
 	this.logger = new $tw.utils.Logger("TiddlyWebAdaptor");
 	this.isLoggedIn = false;
 	this.isReadOnly = false;
-	this.isPollingDisabled = false;
 }
 
 TiddlyWebAdaptor.prototype.name = "tiddlyweb";
@@ -92,15 +91,10 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
 				self.isLoggedIn = json.username !== "GUEST";
 				self.isReadOnly = !!json["read_only"];
 				self.isAnonymous = !!json.anonymous;
-
-				// set whether the syncer should be polling for changes
-				// this allows the syncer to still call for manual polls 
-				// when requested by the user
-				self.isPollingDisabled = false;
 			}
 			// Invoke the callback if present
 			if(callback) {
-				callback(null,self.isLoggedIn,json.username,self.isReadOnly,self.isAnonymous,self.isPollingDisabled);
+				callback(null,self.isLoggedIn,json.username,self.isReadOnly,self.isAnonymous);
 			}
 		}
 	});
--- a/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
+++ b/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
@@ -23,6 +23,7 @@ function TiddlyWebAdaptor(options) {
 	this.logger = new $tw.utils.Logger("TiddlyWebAdaptor");
 	this.isLoggedIn = false;
 	this.isReadOnly = false;
+	this.isPollingDisabled = false;
 }
 
 TiddlyWebAdaptor.prototype.name = "tiddlyweb";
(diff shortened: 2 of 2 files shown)


Jermolene:
Hi @Arlen22 for the record, please can you write the description so that it makes sense to people without them having to read the earlier threads. I think what you're trying to do here is let the tiddlywebadaptor disable syncer polling, but I'm not clear on why the status is being echoed to a tiddler, nor why you're reading changes from that tiddler, nor why it appears to be disabled by default with these changes.

It's also worth checking that you've checked that your changes don't affect the other places that the syncer is used? In particular, the filesystem plugin and the savetrails plugin.

Arlen22:
Okay, how's that. 

No other adapters should be affected as the default behavior is still the same. Pulling is disabled only if the variable is set to true. If it does not exist nothing changes.

Jermolene:
Hi @Arlen22 would it also meet your needs if the syncer were to call `syncadaptor.getUpdatedTiddlers()` and `syncadaptor.getSkinnyTiddlers()` with a new flag that indicated whether the sync was an automated poll or manually triggered? It seems like it might be a bit neater, and keep more of the logic out of the syncer.

Jermolene:
> Okay, how's that.

Apologies, I meant to say thank you, that makes things much easier.

joshuafontany:
@Arlen22 I really like this, this fits really well with my real-time sync experiments.

Jermolene:
Hi @joshuafontany would my alternate suggestion [above](https://github.com/Jermolene/TiddlyWiki5/pull/5882#issuecomment-882078608) also meet your needs?

joshuafontany:
Hi @Jermolene 

I will do another read thru of the syncer code, but I think an option to just turn off polling would work fine. I think I actually do not need to use the getUpdatedTiddler/getSkinnyTiddler code until I implement lazy loading.

The Yjs "conflict free replicated data" library I'm using means I actually do not need to track changeCount/revisions & compare them. All updates to the Y.Map storing each tiddler can be directly pushed to the wiki. (Debouncing the local edits makes sense in order to batch changes into a "transaction" during the Save/Delete Tiddker tasks, but we can apply incoming changes instantly due to how Yjs works.)

Jermolene:
Hi @joshuafontany @Arlen22 just to be clear, I am suggesting an alternative implementation to achieve the same objective as this PR. From the OP:

> This adds another variable to the status response which tells the syncer whether the sync adapter expects it to poll for changes. This is useful if the syncadaptor has some other mechanism to get changes from the server, but still allows a manual refresh to be triggered.

So, the objective is to accommodate syncadaptors that don't want to use polling, but do want to respond to syncer requests to refresh from the server. This PR introduces a lot of complexity in the form of another config tiddlers and associated state, and a bunch of glue.

My proposal is just to point out that if the syncer adds an argument to the polling methods indicating whether this is an automated or manual poll, then it becomes trivial for the syncadaptor to decide whether or not it wants to respond to polling requests. It seems a good deal simpler than this PR, and achieves the same goal.

Arlen22:
Part of the issue is that the syncadaptor class has no concept of scheduling. All of the intervals are handled in the syncer and the syncadaptor just does what the syncer calls for. As it is currently, there is no need to change anything about the syncadaptor, only the syncer. Changing the tiddlyweb syncadaptor at all is unnecessary in this case and only included for demonstration purposes. So this is mainly a change to the syncer system from my perspective.

Arlen22:
The other option would be to remove all the syncadaptor changes, and just have the status tiddler that sets whether to poll or not. That seems like the best of both worlds and allows us to kick the ball down the road on how to handle the syncadaptor, if we ever need to, which we probably don't. If the syncadaptor even cares whether it's polled or not, the plugin it is from can just include that status tiddler or it can set it at any time in the wiki since they share the same wiki.

Arlen22:
> The other option would be to remove all the syncadaptor changes, and just have the status tiddler that sets whether to poll or not. That seems like the best of both worlds and allows us to kick the ball down the road on how to handle the syncadaptor, if we ever need to, which we probably don't. If the syncadaptor even cares whether it's polled or not, the plugin it is from can just include that status tiddler or it can set it at any time in the wiki since they share the same wiki.

Ok, I removed the syncadaptor part of things so now it just checks the tiddler when scheduling the next sync and calls the status hook after a status is returned, because that's still important for login and stuff like that.

Jermolene:
Hi @Arlen22 apologies if I'm being slow, but please can you just explain why you prefer your solution over my proposal?

Arlen22:
> Hi @Arlen22 apologies if I'm being slow, but please can you just explain why you prefer your solution over my proposal?

The code changes required are a lot simpler. If we pass a flag to the syncadaptor then I have no choice but to override the getSkinnyTiddlers function at runtime to check whether the flag is set or not. It's a really hacky solution to me, and I would much rather just remove all the sync adaptor related code changes and just tell the syncer whether to poll or not. 

Plus I have to make sure I keep up to date with any method signature changes (in the overload) and we also have to worry about backward compatibility in the future, and adding a flag like that will have a domino effect throughout the entire server architecture because other tools and plugins will start depending on it as well. Since the tiddler-only solution has the exact same outcome and is much less brittle, I prefer it far more.

Jermolene:
> The code changes required are a lot simpler

A lot simpler for whom? We need to bias in favour of making the core as simple as possible because the complexity of the core is a tax on everyone using TW. And these changes introduce complexity that is to visible both to users and developers.

>because I just have to set a tiddler and hook into the status response. It has the exact same outcome and feels a lot less brittle.

It seems as though two things are being conflated here:

* Avoiding automatic polling when the syncadaptor doesn't need it, which it seems can be accomplished with the proposal I made above
* Concerns that arise from the way that the SSE work is piggybacking on the tiddlyweb syncadaptor (eg your point about not wanting to override another of the syncadaptor methods). The obvious way to accomplish that is by introducing a new syncadaptor that delegates some methods to an embedded instance of the TiddlyWeb adaptor

Arlen22:
> A lot simpler for whom? We need to bias in favour of making the core as simple as possible because the complexity of the core is a tax on everyone using TW. And these changes introduce complexity that is to visible both to users and developers.

A status tiddler with a simple toggle is a lot less complex and a lot more in line with TiddlyWiki than the following, which is what I am understanding you to say would be required for me to accomplish turning polling off. 

```js
$tw.hooks.addHook("syncer-status-result", function(syncer){
    if(syncer.syncadaptor.name !== "tiddlyweb") return;
    var original = syncer.syncadaptor.getSkinnyTiddlers;
    syncer.syncadaptor.getSkinnyTiddlers = function(isPoll, callback){
        if(isPoll) return;
        
        return original.call(syncer.syncadaptor, isPoll, callback);
    });
});
```

Where would you be planning to add the flag? Maybe I'm misunderstanding what you mean by flag. I'm imagining an isPoll variable as shown above. 

If you don't like the tiddler idea, an instance variable on the syncadaptor (adaptor.isPollingDisabled) would work equally well and be equally visible to anyone using the status hook. The syncer would just check that rather than checking the value of a tiddler. I thought that was the preferred TiddlyWiki way of communicating between different parts, so I'm rather confused. The code I wrote above does not line up at all with anything that I'm familiar with in core, so I don't understand what I'm missing, and maybe I'm missing something big, I don't know.

Arlen22:
If multiple plugins want polling disabled they all have to do it like this, and there's no way for anyone to check whether polling has been disabled or not.

Putting this after the call back isn't pretty but would work and would be backward compatible for the moment but I really don't want to introduce that as a backward compatibility restraint at this time because I know there's a much cleaner way to do this and since this is the last minute change I'm trying to keep it as simple and low-key as possible. I know that once I sit back and think over this I'll come up with a much better way of doing it then a flag that gets passed to the sync adapter.

Arlen22:
An instance variable on the sync adapter would work equally well because other plugins can just set it if needed.

Arlen22:
It also seems really strange that I can't just tell the syncer to stop polling. It really seems to break from the previous standard for the syncadaptor to just completely ignore it based on the flag. 

This isn't really how sync adaptors are intended to work because there is no way currently to push changes from the sync adaptor so it shouldn't matter to the sync adaptor whether it's being polled or not.

Jermolene:
Hi @Arlen22 

> A status tiddler with a simple toggle is a lot less complex and a lot more in line with TiddlyWiki than the following, which is what I am understanding you to say would be required for me to accomplish turning polling off.

Given the purpose of this PR, the goal is not to make the SSE experiments simpler, it is to make the minimum core changes to support those experiments.

> I don't even know if that's a backward compatible change now that I write it out like this because it changes the function signature and moves the callback.

We can do it in a backwards compatible way by checking the types of the function arguments. We'd add a new "options" first parameter that would contain the flag.

```
$tw.hooks.addHook("syncer-status-result", function(syncer){
    if(syncer.syncadaptor.name !== "tiddlyweb") return;
    var original = syncer.syncadaptor.getSkinnyTiddlers;
    syncer.syncadaptor.getSkinnyTiddlers = function(isPoll, callback){
        if(isPoll) return;
        
        return original.call(syncer.syncadaptor, isPoll, callback);
    });
});
```

It looks like you're using two techniques there: hooks and overriding methods of the syncadaptor. I think you're hitting the limits of those techniques in this case, and would be better minting a new syncadaptor that contains an instance of the tiddlywebadaptor to which it delegates some methods.

Hooks are attractive at first sight, but very brittle and present an enormous challenge in terms of backwards compatibility since their implementation is so closely coupled with the core. They need to be used with care.

> If multiple plugins want polling disabled they all have to do it like this, and there's no way for anyone to check whether polling has been disabled or not.

I think framing this as "how can we stop the syncer from polling" is unhelpful. Better to frame it as "how can we hack the tiddlyweb syncadaptor to extend it to use SSE".

> It also seems really strange that I can't just tell the syncer to stop polling.

If we did want to do that I would recommend adding support for disabling syncing when `$:/config/SyncPollingInterval` is zero, just as we do with disabling animations.

> It really seems to break from the previous standard for the syncadaptor to just completely ignore it based on the flag.

By "previous standard" do you mean the way that other configuration options are handled?

> This isn't really how sync adaptors are intended to work because there is no way currently to push changes from the sync adaptor so it shouldn't matter to the sync adaptor whether it's being polled or not.

The syncadaptor should be able to call the `storeTiddler()` method on the syncer.

Arlen22:
> > I don't even know if that's a backward compatible change now that I write it out like this because it changes the function signature and moves the callback.
> 
> We can do it in a backwards compatible way by checking the types of the function arguments. We'd add a new "options" first parameter that would contain the flag.

If we make any change to syncer that moves the callback we will break all existing sync adaptors. That's what I mean by backwards compatible and nothing of what I'm doing requires that at the moment.

> > If multiple plugins want polling disabled they all have to do it like this, and there's no way for anyone to check whether polling has been disabled or not.
> 
> I think framing this as "how can we stop the syncer from polling" is unhelpful. Better to frame it as "how can we hack the tiddlyweb syncadaptor to extend it to use SSE".

SSE code isn't related to the tiddlyweb sync adaptor at the moment. What we're really trying to get here is a clean way for any sync adaptor to stop polling from happening. The status hook seems important as well since the syncer calls it after every status change, making it useful for handling login and such things. 

> 
> > It also seems really strange that I can't just tell the syncer to stop polling.
> 
> If we did want to do that I would recommend adding support for disabling syncing when `$:/config/SyncPollingInterval` is zero, just as we do with disabling animations.

That's all I'm looking for. If that's the way you recommend to do it then that's what I'll do. I'll have to get to that later today hopefully. 

> > This isn't really how sync adaptors are intended to work because there is no way currently to push changes from the sync adaptor so it shouldn't matter to the sync adaptor whether it's being polled or not.
> 
> The syncadaptor should be able to call the `storeTiddler()` method on the syncer.

It currently doesn't have an instance specific way to do so. Most of these things appear to be implemented with the goal of supporting multiple wiki objects in one environment (so we can sync multiple wiki objects to mutliple backends, and I'm keeping that in mind for the future. Anything that accesses the $tw global for the syncer instance breaks that. But that's just something I see in the future. It's not related to this PR.

Jermolene:
Thanks for your patience @Arlen22 

> If we make any change to syncer that moves the callback we will break all existing sync adaptors. That's what I mean by backwards compatible and nothing of what I'm doing requires that at the moment.

Apologies, you're quite correct. I made the same mistake the other day. In this case we would indeed be constrained to adding a new 2nd argument, which would be backwards compatible, but would end up with a slightly unconventional functions signature.

> SSE code isn't related to the tiddlyweb sync adaptor at the moment.

Is the idea that the SSE stuff should work with other syncadaptors? That seems ambitious if so.

I can't quite see how the SSE code cannot be closely coupled with the tiddlyweb adaptor, or what is the benefit of it not being so.

As I say, at the moment, I'm seeing this as a subclass of the tiddlywebadaptor which seems obvious for precisely the reason that we're discussing: that the new code needs to modify the behaviour of the tiddlywebadaptor. I can see how you can also see it as modifying the behaviour of the syncer, but that line of reasoning leads you to these deep modifications to the syncer.

> What we're really trying to get here is a clean way for any plugin to stop polling from happening. 

Why would "any plugin" want to stop polling from happening?

> The status hook seems important as well since the syncer calls it after every status change, making it useful for handling login and such things.

> That's all I'm looking for. If that's the way you recommend to do it then that's what I'll do. I'll have to get to that later today hopefully.

No it's not. These changes go much further than that.

> It currently doesn't have an instance specific way to do so.

I'm fairly sure I've made it work in the past, are you sure that it is impossibe?

> Most of these things appear to be implemented with the goal of supporting multiple wiki objects in one environment (so we can sync multiple wiki objects to mutliple backends, and I'm keeping that in mind for the future. Anything that accesses the $tw global for the syncer instance breaks that. But that's just something I see in the future. It's not related to this PR.

That all sounds correct except I'm not sure exactly what you mean by "these things"?

Arlen22:
> > SSE code isn't related to the tiddlyweb sync adaptor at the moment.
> 
> Is the idea that the SSE stuff should work with other sync adaptors? That seems ambitious if so.

No, its probably more correct to say that adding a way to disable polling goes far beyond just tiddlyweb. 

The SSE stuff on the client side also goes far beyond tiddlyweb. This is just our test bed. I absolutely believe we will take another look at every other sync adaptor out there to see how we can take advantage of this ability as well, and the ability to monitor for signals from the outside and then trigger a sync manually is really big. 

No sync adaptor needs to actually get involved in that if it is setup like tiddlyweb. So if it's GitHub, Dropbox, or any other cloud service, we can listen for changes in a separate plugin then call syncFromServer when something changes. In most cases you can subscribe to a minimal endpoint that lets you know when something changes without needing to know what changed and then you just trigger the existing sync adaptor through the syncer. 

> > What we're really trying to get here is a clean way for any plugin to stop polling from happening. 
> 
> Why would "any plugin" want to stop polling from happening?

I mean we can just set it to poll every ten minutes or ten hours, I guess. If the polling is being triggered from outside the sync adaptor by another plugin then it wastes resources. I want to stop polling because I'm writing a plugin that adds additional functionality to the tiddlyweb adapter by making it trigger automatically instead of polling and the polling is just unnecessary. Many other plugins could take advantage of the same thing. But a flag to the syncadaptor is an unnecessary change that is pretty breaking. 

> > Most of these things appear to be implemented with the goal of supporting multiple wiki objects in one environment (so we can sync multiple wiki objects to mutliple backends, and I'm keeping that in mind for the future. Anything that accesses the $tw global for the syncer instance breaks that. But that's just something I see in the future. It's not related to this PR.
> 
> That all sounds correct except I'm not sure exactly what you mean by "these things"?

By "these things" I guess I would be referring to most of the things in core which are passed a wiki instance in the constructor, as the it allows multiple instances to operate simultaneously on multiple different Wiki instances.

Arlen22:
So in theory we could just set a really long poll, like 10 hours, and then just have the status hook in this PR. I'll wait for your response though.

joshuafontany:
I like the idea of disabling polling by setting the interval to 0.

Arlen22:
I'm also fine with not doing the status hook if it looks brittle to you because I can see ways around that for now since it does look like it could draw flies. So that one would be optional for me. But being able to turn off polling would be a positive in my mind. Which I guess means we could get away with not doing anything for right now.

Arlen22:
> Hooks are attractive at first sight, but very brittle and present an enormous challenge in terms of backwards compatibility since their implementation is so closely coupled with the core. They need to be used with care.

This is the reason I'm fine with dropping the hook, since you make it sound at least as bad as just overriding the syncer.getStatus function. Obviously overrides are never supported and are completely hacky, so I won't be using that in the final solution. I can listen for wiki change events and check for the username tiddler being modified as an alternative to the status hook. That would be pretty much identical. 

I'll wait for your opinion on this as well. Shall I remove the hook from the PR and just use a wiki listener instead in my plugin? The solution is identical and I had not thought of that earlier.

Arlen22:
We're really paring this thing down!

Arlen22:
@pmario I updated https://github.com/twcloud/tiddlyweb-sse to try to use the wiki change listener and a ten hour poll and it seems to work fine. Feel free to take a look. If you run `npm run serve` you should be able to immediately test it. You have to run `npm install` first if you haven't already. 

Obviously we should still have a way to completely disable the poll, because who knows what the ramifications of an unnecessary timeout are in some constrained environment.

Jermolene:
Apologies @Arlen22 I won't have a chance to review this properly until after the weekend, thank you for your continued patience.

pmario:
>@pmario I updated https://github.com/twcloud/tiddlyweb-sse to try to use the wiki change listener and a ten hour poll and it seems to work fine. Feel free to take a look. If you run npm run serve you should be able to immediately test it. You have to run npm install first if you haven't already.

Hi, This version contains a major regression. Try the following, which was possible with the "hooked" version. 

 - Start the SSE server
 - open 2 windows with the same wiki A and B
 - create a new tiddler in 1 wiki A and add some text
 - open the "red button" in window B, to see the  new tiddler from window A
 - add some text in window A -> it syncs
 - now add some text in window B -> it doesn't sync anymore. 

With the "hook" approach 2 way syncing worked. Now it doesn't 

I'm using windows 10 and FireFox for testing

Arlen22:
> > @pmario I updated https://github.com/twcloud/tiddlyweb-sse to try to use the wiki change listener and a ten hour poll and it seems to work fine. Feel free to take a look. If you run npm run serve you should be able to immediately test it. You have to run npm install first if you haven't already.
> 
> Hi, This version contains a major regression. Try the following, which was possible with the "hooked" version.
> 
> * Start the SSE server
> * open 2 windows with the same wiki **A** and **B**
> * create a new tiddler in wiki **A** and add some text
> * open the "red button" in window **B**, to see the  new tiddler from window A
> * add some text in window **A** -> it syncs
> * now add some text in window **B** -> it doesn't sync anymore.
> 
> With the "hook" approach 2 way syncing worked. Now it doesn't
> 
> I'm using windows 10 and FireFox for testing

I am able to reproduce this. However I am also able to reproduce this on the implementation that we reverted on master earlier. I can confirm the updated tiddler is being requested according to the network tab, so there must be some kind of edit conflict happening there that needs to be resolved in the syncer or the sync adapter. The SSE plugin just calls syncer.syncFromServer whenever there is a change (after waiting for the throttle interval).

Arlen22:
In other words, I believe this is an edit conflict situation which was already in code prior to any of this, and if you set the poll interval to something like 5 seconds (or anything longer) you should see the exact same behavior without SSE.

Arlen22:
@Jermolene I just realized that setting the syncer poll interval to zero is going to be a pain because as it currently stands it will immediately poll if set to zero, so it will be always making web requests. A separate tiddler would be preferable. The maximum setTimeout value is 24 days, according to [StackOverflow](https://stackoverflow.com/q/52294430/258482).

Jermolene:
I reviewed this for inclusion in v5.2.0 but am not merging it for the following reasons:

1. As @Arlen22 [noted above](https://github.com/Jermolene/TiddlyWiki5/pull/5882#issuecomment-883830066) the hook is not needed (it is still included in this PR at the moment)
2. This PR contains several pure whitespace changes that clutter things up. If whitespace needs adjusting for a file, we prefer a separate PR so that it is easier to review

We could still add the configuration tiddler, but it should be changed to `$:/status/SyncPollingDisable` for consistency with the existing configuration tiddler titles.

Arlen22:
I agree. I'll add a PR at some point, but I'm not sure how much it matters because a ten hour poll isn't bad. Thanks for your input on this.

PR #5883, opened 2021-07-15
Fix: bug with List Widget counter-last variable

saqimtiaz:
This is a fix for a bug with the current implementation of the list widget when using the counter variable. If the last item is removed with no earlier changes in the list, then the `<counter-variable-name>-last` variable is not updated. I've added a test for this scenario.

An alternative approach would be to make the `-last` variable a separate attribute that has to be specified and only do the refreshing for it then. This would result in slightly better performance when a user wants a `counter` variable but not the `-last` variable. I am not convinced those marginal gains are significant enough to warrant going this route.

I am just entering a busy period expected to last ~ 3-4 weeks during which I wont have the opportunity to contribute to the core. So if we want to explore a different fix, depending on the timing of the 5.2.0 release, someone else may need to pick up the work on this.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -236,6 +236,11 @@ ListWidget.prototype.handleListChanges = function(changedTiddlers) {
 					hasRefreshed = hasRefreshed || refreshed;
 				}
 			}
+			// If there are items to remove and we have not refreshed then recreate the item that will now be at the last position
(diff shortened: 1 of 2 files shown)


Jermolene:
Looks good thanks @saqimtiaz – thank you for all your help recently, and I hope you have a productive busy period!

PR #5538, opened 2021-03-10
Fix #5537

laomaiweng:
This PR implements the fix(es) proposed in issue #5537. For now, the most basic fix is simply to use `(authenticated)` as the default authorized username when only a credentials file is present.

If it seems appropriate, I can give a shot at adding a "role" (reader/writer) to the CSV credentials file. Please do not hesitate to suggest better approach as you see fit.

PR #5899, opened 2021-07-20
Fix server options

joshuafontany:
- ~~[Edit] Added a per-route authorizationType option to the requestHandler. This will allow different routes to have different readers/writers (multi-wiki+multi-user focus).~~
- Allows a requiredPlugins property to be set on the server constructor options.
- Adds a tlsPassphrase command line option to allow initializing certs that were created with a 'passphrase'
- Updates the docs to reflect the current default `WebServer API: Get All Tiddlers` settings
- Adds an 'admin' authorization principle.
- Adds a `.origin` and `.pathprefix` entry to the `this.boot` object when the sever starts.


pmario:
@joshuafontany ... Would you like to know, how to avoid those "Merge remote-tracking branch 'upstream/master'" commits in your PR`s?

joshuafontany:
> @joshuafontany ... Would you like to know, how to avoid those "Merge remote-tracking branch 'upstream/master'" commits in your PR`s?

I think you mean the "squash" workflow? Yeah, still learning about that...

pmario:
>I think you mean the "squash" workflow? Yeah, still learning about that...

No. 

I don't use my local "master" branch anymore. I directly use jermolene/master as my master branch. With that setting I also avoid to merge stuff into my master, because I can't push it to github. .. It would want to push it to Jeremy's branch, which I can't. ... So I'm sure I always create a new feature branch for new stuff. 

Since I use VSCode all the command line stuff is hidden by the UI. 

My "master-branch" setting in `.git/config` looks like this

```
[branch "master"]
	remote = jermolene
	merge = refs/heads/master
```

This is the only `[branch "master"]` setting!
So if I checkout "master" I do get the actual version from Jeremy. 

To get the latest version I need 

```
git pull --tags jermolene master
```

joshuafontany:
> > I think you mean the "squash" workflow? Yeah, still learning about that...
> 
> No.
> 
> I don't use my local "master" branch anymore. I directly use jermolene/master as my master branch. With that setting I also avoid to merge stuff into my master, because I can't push it to github. .. It would want to push it to Jeremy's branch, which I can't. ... So I'm sure I always create a new feature branch for new stuff.
> 
> Since I use VSCode all the command line stuff is hidden by the UI.
> 
> My "master-branch" setting in `.git/config` looks like this
> 
> ```
> [remote "jermolene"]
> 	url = https://github.com/Jermolene/TiddlyWiki5.git
> 	fetch = +refs/heads/*:refs/remotes/jermolene/*
> 
> ...
> 
> [branch "master"]
> 	remote = jermolene
> 	merge = refs/heads/master
> ```
> 
> This is the only `[branch "master"]` setting!
> So if I checkout "master" I do get the actual version from Jeremy.
> 
> To get the latest version I need
> 
> ```
> git pull --tags jermolene master
> ```
> 
> Since I use this workflow, I don't have a problem with the redundant merge-messages anymore.

Whoa! Thats really neat, thank you! I will definitely try to move to that workflow for the TW repo. Mahalo (thanks)!

joshuafontany:
@Jermolene Bumping for review (when you have a chance). Thanks!

joshuafontany:
@jermolene @saqimtiaz @pmario I have consolidated and cleaned up _most_ of my Server tweaks into this PR. I have also updated the documentation. Would love a review. My MultiServer and CORS experiments depend on these changes. Thanks!

saqimtiaz:
@joshuafontany will try to find time this weekend for a proper review. Did see some string literals in there... its server side code but is that something we allow in the core? Not sure.

saqimtiaz:
@joshuafontany apologies for the delay in reviewing this, I've had a particularly busy schedule recently.

I am a bit hesitant about the per path authorization changes in the sense that the server currently does not support such a concept, and if we are going to introduce it we probably want to do so with a bit more consideration as to how that is implemented and how flexible it is. So this might be best broken out into its own PR from my perspective. What do you think @Jermolene 

The server-settings file idea is an interesting one. Am I correct in thinking that the primary benefit is a convenience method for saving and using different server configurations, and the same could be achieved with different scripts used to start the wiki?

editions/tw5.com/tiddlers/WebServer Parameter_ admin.tid
```diff
@@ -0,0 +1,10 @@
+caption: admin
+created: 20211111015634164
+creator: Joshua Fontany
```
Jermolene, line 3: House style is not to have "creator" or "modifier" fields on docs tiddlers.

editions/tw5.com/tiddlers/WebServer Parameter_ server-settings.tid
```diff
@@ -0,0 +1,29 @@
+caption: server-settings
+created: 20211111003224763
+creator: Joshua Fontany
+modified: 20211111023427713
+modifier: Joshua Fontany
+tags: [[WebServer Parameters]]
+title: WebServer Parameter: server-settings
+type: text/vnd.tiddlywiki
+
+The [[web server configuration parameter|WebServer Parameters]] ''server-settings'' contains the pathname of a [[JSON file|https://en.wikipedia.org/wiki/JSON]] containing attribute-value pairs for //all other// [[web server configuration parameter|WebServer Parameters]].
```
Jermolene, line 10: Could we store this data in the tiddlywiki.info file?

Jermolene (commented):
Thanks @joshuafontany @saqimtiaz – these are worthwhile improvements.

joshuafontany:
> @joshuafontany apologies for the delay in reviewing this, I've had a particularly busy schedule recently.
> 
> I am a bit hesitant about the per path authorization changes in the sense that the server currently does not support such a concept, and if we are going to introduce it we probably want to do so with a bit more consideration as to how that is implemented and how flexible it is. So this might be best broken out into its own PR from my perspective. What do you think @Jermolene

I agree. That is the most radical change. Honestly, the "multi-wiki" feature is much lower on my priority list, and it has recently caused logic problems around global-plugins (what happens when a sub-wiki changes a globally served plugin???). 

I have pulled the Per-Path Auth changes out and I am setting that feature aside from the scope of this PR for now. We can revisit if it is worth allowing one "root" wiki-server to serve other editions at sub-paths, or if some other approach would be better.


> The server-settings file idea is an interesting one. Am I correct in thinking that the primary benefit is a convenience method for saving and using different server configurations, and the same could be achieved with different scripts used to start the wiki?

(See @Jermolene's comment about storing this data in the tiddlywiki.info file)

I had considered both of those approaches. I am building towards a "app-like" experience for end-users, but would still like end-users to be able to host their own server locally (that can then be found from a "directory" if listed as public). I found setting up a whole bunch of very different bash scripts (or VSCode "launch configurations") a bit rough to ensure the various test-case coverage I needed. I also wanted to separate it from the Wiki (tiddlywiki.info) being served so you could use the same config for different wikis.

Being able to reference a JSON filepath that has "human readable/editable defaults" and then override __that__ on a per-wiki or per-bash-launch-command basis was the most flexible. There are security concerns around storing __too much__ server startup data in a single flat file, but I hope that documentation and user feedback can help there.

saqimtiaz:
> Being able to reference a JSON filepath that has "human readable/editable defaults" and then override **that** on a per-wiki or per-bash-launch-command basis was the most flexible.

Right. However the tiddlywiki.info file is a JSON file as well. Could we could read defaults from there and allow for them to be overridden from the command line?  I am hesitant to introduce yet another configuration file for users to have to think about, given that the majority of even our node.js users do not have a technical background.

joshuafontany:
> > Being able to reference a JSON filepath that has "human readable/editable defaults" and then override **that** on a per-wiki or per-bash-launch-command basis was the most flexible.
> 
> Right. However the tiddlywiki.info file is a JSON file as well. Could we could read defaults from there and allow for them to be overridden from the command line? I am hesitant to introduce yet another configuration file for users to have to think about, given that the majority of even our node.js users do not have a technical background.

This would work, but pretty firmly ties a single server "setup" to that wiki-folder, and that seems like defining those in the "wrong place". Although, I just realized that if the properties are well structured, you could just `server-settings=./tiddlywiki.info`. That could be called out in the documentation.

editions/tw5.com/tiddlers/WebServer Parameter_ server-settings.tid
joshuafontany: @Jermolene This was brought up below by @saqimtiaz. While I like how this simplifies things for a single-wiki-server setup, I think it would be difficult in certain more complex cases (for example, I run a authenticated wiki server for my own multiple project wikis, and am often re-launching with the same server settings but pointing to a different wiki... but I also use a `writers=(anon)` wiki server whenever I update TW documentation). I would want point to a file outside of TW repository for cases where I am launching an edition from within the TW repository...

Hm, it would definitely give users not familiar with Node.js servers a "central settings file" to reduce cognitive load. That's a big point in favor. On the other hand, you could do `server-settings=./tiddlywiki.info` from the command line and achieve the same thing. This could be called out in the documentation. Thoughts?

joshuafontany:
Including the logic in https://github.com/Jermolene/TiddlyWiki5/pull/5538

joshuafontany:
@Jermolene @saqimtiaz Having read up on node.js and what you can do with the package.json, as well as other ways to setup command parameters, I have removed the `server-settings` option. This would just add confusion, and I can manage my different server settings in another fashion. At this point, I consider this ready for review/merging. Thanks everybody for your help!

joshuafontany:
I have moved the requiredPlugins functionality to a server-parameter. I am starting to see how useful hooks are in the core, and am moving over to using as much core code as possible and hooking in where required. For example I will use the `--listen` command instead of my own `--ws-listen` command, and hook in with `th-server-command-post-start`.

Jermolene:
Thanks for your patience @joshuafontany

joshuafontany:
> Thanks for your patience @joshuafontany

No problem! Sweet, gotta get to cleaning up my experiments! Thanks!

PR #5900, opened 2021-07-20
Fix adaptor boot option

joshuafontany:
Opens up the filesystemadaptor and filesystem utils to accept arbitrary `boot` objects in the options. This will allow one server to serve multiple wikis sourced from different wiki folders.


joshuafontany:
Closed in error. I need to review if this seems like the best way to allow one "root" TW instance to manage/serve other wikis. I think I did not fully understand the boot process when I sketched these changes out...

PR #5904, opened 2021-07-21
tabs-macro improve "default" parameter docs 

pmario:
This PR adds a bit more info to the tab-macro "default" parameter to make it clearer how it is supposed to work


pmario:
@Jermolene ... Bump ... It's docs only

Jermolene:
Thanks @pmario

PR #5909, opened 2021-07-22
Hide more-tiddler-actions button if all buttons are shown

BurningTreeC:
This PR hides the `more-tiddler-actions` button if ALL buttons are shown in the `View Toolbar` (excluding the fold-bar)

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -61,6 +61,12 @@ $else$
 </$set>
 \end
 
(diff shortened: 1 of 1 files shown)


pmario:
IMO that's not necessary. The "More" toolbar button has a "visibility" tiddler. If a user sets all other buttons as visible, but don't want to see the "More" button, they can do so, while they are in the menu.

Hiding elements with CSS, does still create the widget-tree and triggers the refresh mechanism for changes, that are not visible. So IMO this change wouldn't be efficient either.

PR #5910, opened 2021-07-23
Add pop storyview to ViewToolbar buttons

BurningTreeC:
This PR adds the `pop storyview` to the `ViewToolbar` buttons

### DIFF
--- a/core/ui/ViewTemplate/title.tid
+++ b/core/ui/ViewTemplate/title.tid
@@ -8,7 +8,7 @@ fill:$(foregroundColor)$;
 <div class="tc-tiddler-title">
 <div class="tc-titlebar">
 <span class="tc-tiddler-controls">
-<$list filter="[all[shadows+tiddlers]tag[$:/tags/ViewToolbar]!has[draft.of]addprefix[$:/config/ViewToolbarButtons/Visibility/]!contains:text[hide]removeprefix[$:/config/ViewToolbarButtons/Visibility/]]" storyview="pop" variable="listItem"><$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]]"><$transclude tiddler=<<listItem>>/></$set></$list>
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@BurningTreeC using the `lookup[]` operator here might be a good idea.

I think we need to invest some time in setting up our CI to automatically build and make available TiddlyWiki files for PRs, as that would make testing and providing feedback on PRs so much easier.

BurningTreeC:
@saqimtiaz I don't know how I should realize this with the lookup operator ...

saqimtiaz:
It would be something along these lines:

`[all[shadows+tiddlers]tag[$:/tags/ViewToolbar]!has[draft.of]] :filter[lookup[$:/config/ViewToolbarButtons/Visibility/]!match[hide]]
`

pmario:
I don't understand this PR. Why do we need a "storyview" button in the tiddler toolbar?

Jermolene:
Hi @BurningTreeC @pmario actually I like this: the pop animation makes it much more obvious what's happening when clicking the toolbar visibility checkboxes in the first tab of the tiddler info panel.

saqimtiaz:
> Hi @BurningTreeC @pmario actually I like this: the pop animation makes it much more obvious what's happening when clicking the toolbar visibility checkboxes in the first tab of the tiddler info panel.

Likewise. We should however confirm that removing the reveal widget does not have other consequences, for instance in terms of qualified state tiddlers etc. I think that should be OK but I don't have the opportunity to consider it properly at present.

pmario:
>Hi @BurningTreeC @pmario actually I like this: the pop animation makes it much more obvious what's happening when clicking the toolbar visibility checkboxes in the first tab of the tiddler info panel.

uups, I didn't have a look at the code. ... I did miss, that it's the animation that will be added and not the button.  ... So just ignore my comment. 

Which still should remind us, that PR request descriptions should be clear. This one wasn't .. At least for me :/

BurningTreeC:
@Jermolene , this PR would be ready for review

Jermolene:
Thanks @BurningTreeC @saqimtiaz I'm fairly confident that removing the reveal widget won't affect any generated qualifiers.

PR #5914, opened 2021-07-26
Fixes getStatus error alerts

joshuafontany:
All other connections errors lookup the localized Connection Error message and displays it in an alert. This aligns the getStatus function with that standard.


Jermolene:
Thanks @joshuafontany

PR #5915, opened 2021-07-26
Revise chinese translations

BramChen:
* Remove superfluous `~` from `ExportTiddlyWikiCore/Caption` in Buttons.multids


Jermolene:
Thanks @BramChen

Issue #1903, opened 2015-08-08
Simplify Widget creation


Drakor:
Hi,

when programming plugins or working with widgets in general, one often needs to add a small widgets or a small action widget, that doesn't do much. However the process to create a widget at the moment is tedious at best, since there is a larger boilerplate of code that has to be present and one usually has to write multipla functions, since the current ones don't always encapsulate the behaviour your want to change perfectly.

Therefore I have found myself to use something like this (yes there are other function not in this excerpt that are non-standard):

```
/*\
title: $:/plugins/ahahn/unibar/commonAction.js
type: application/javascript
module-type: widget

Cpmmon Action Widget

\*/
(function(){

/*jslint node: true, browser: true */
/*global $tw: false */
"use strict";

var Widget = require("$:/core/modules/widgets/widget.js").widget;

var CommonActionWidget = function(parseTreeNode,options) {
	this.initialise(parseTreeNode,options);
};

/*...*/
/*
Setup an action widget with these properties
*/
CommonActionWidget.prototype.setup = function(allowPropagation, doRenderChildren, preparedParams, refreshOnAttributeChange) {
	this.doRenderChildren = doRenderChildren;
	this.preparedParams = preparedParams;
	this.allowPropagation = allowPropagation;
	this.refreshOnAttributeChange = refreshOnAttributeChange;
}


/* ... */

})();
```

It allows you to preconfigure a widget, althoug it is currently geared towards action widgets in particular. Using this, a simple copy action widget, that copies a tiddler, might look like this:

```
/*\
title: $:/plugins/ahahn/unibar/copy.js
type: application/javascript
module-type: widget

Copy action widget

\*/
(function(){

/*jslint node: true, browser: true */
/*global $tw: false */
"use strict";

var CommonAction = require("$:/plugins/ahahn/unibar/commonAction.js").commonAction;

var CopyWidget = function(parseTreeNode,options) {
	this.initialise(parseTreeNode,options);
	this.setup(false, false, ["$src", "$dest", "$mode"], true);
};

/*
Inherit from the base widget class
*/
CopyWidget.prototype = new CommonAction();

CopyWidget.prototype.invokeAction= function(event) {
	if (this.param["$src"] && this.param["$dest"] && (this.param["$mode"] == "fields")) {
		var srcTid = this.wiki.getTiddler(this.param["$src"]);
		var destTid = new $tw.Tiddler(srcTid, {"title": this.param["$dest"]});
[...]

PR #5920, opened 2021-07-31
Assign domnode styles and classes to widgets that create domNodes

BurningTreeC:
This PR assigns classes and styles dynamically to the domNodes generated by our widgets

### DIFF
--- a/core/modules/widgets/browse.js
+++ b/core/modules/widgets/browse.js
@@ -98,7 +98,6 @@ BrowseWidget.prototype.execute = function() {
 
 BrowseWidget.prototype.assignDomNodeClasses = function() {
 	var classes = this.getAttribute("class","").split(" ");
-	classes.push("tc-eventcatcher");
 	this.domNode.className = classes.join(" ");
 };
 
--- a/core/modules/widgets/browse.js
+++ b/core/modules/widgets/browse.js
@@ -96,16 +96,6 @@ BrowseWidget.prototype.execute = function() {
 	this.nwdirectory = this.getAttribute("nwdirectory");
 };
 
-BrowseWidget.prototype.assignDomNodeClasses = function() {
-	var classes = this.getAttribute("class","").split(" ");
-	this.domNode.className = classes.join(" ");
-};
-
-BrowseWidget.prototype.assignDomNodeStyles = function() {
-	var styles = this.getAttribute("style");
-	this.domNode.style = styles;
-};
-
 /*
 Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
 */
--- a/core/modules/widgets/browse.js
+++ b/core/modules/widgets/browse.js
@@ -110,11 +110,11 @@ BrowseWidget.prototype.assignDomNodeStyles = function() {
 Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
 */
 BrowseWidget.prototype.refresh = function(changedTiddlers) {
-	var changedAttributes = this.computeAttributes(),
-		changedAttributesCount = $tw.utils.count(changedAttributes);
-	if(changedAttributesCount === 1 && changedAttributes["class"]) {
+	var changedAttributes = this.computeAttributes();
+	if(changedAttributes["class"]) {
 		this.assignDomNodeClasses();
-	} else if(changedAttributesCount === 1 && changedAttributes["style"]) {
+	}
+	if(changedAttributes["style"]) {
 		this.assignDomNodeStyles();
 	}
 	return false;
--- a/core/modules/widgets/browse.js
+++ b/core/modules/widgets/browse.js
@@ -70,6 +70,11 @@ BrowseWidget.prototype.render = function(parent,nextSibling) {
 		}
 		return false;
 	},false);
+	this.domNode = domNode;
+	// Assign classes
+	this.assignDomNodeClasses();
+	// Assign styles
+	this.assignDomNodeStyles();
 	// Insert element
 	parent.insertBefore(domNode,nextSibling);
 	this.renderChildren(domNode,null);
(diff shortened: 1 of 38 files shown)


core/modules/widgets/browse.js
```diff
@@ -91,10 +96,27 @@ BrowseWidget.prototype.execute = function() {
 	this.nwdirectory = this.getAttribute("nwdirectory");
 };
 
+BrowseWidget.prototype.assignDomNodeClasses = function() {
```
Jermolene, line 99: I wonder if we might make this a method in widget.js. It would need to be refactored to to take an options argument with a few options:

* an array of classnames to be added to those specified in the attribute
* an override of the name of the attribute, if we have any widgets that use a non-standard attribute name (I can't remember if we do)

core/modules/widgets/browse.js
```diff
@@ -91,10 +96,27 @@ BrowseWidget.prototype.execute = function() {
...
+	var classes = this.getAttribute("class","").split(" ");
+	this.domNode.className = classes.join(" ");
+};
+
+BrowseWidget.prototype.assignDomNodeStyles = function() {
```
Jermolene, line 104: Again, we can make this a method on the superclass in widget.js

Jermolene (commented):
Thanks @BurningTreeC this looks good so far. I've made a couple of suggestions.

I appreciate your help with this, it is intricate and big, and will need very careful review.

core/modules/widgets/browse.js
BurningTreeC: @Jermolene , I've refactored this method into widget.js
Now it accepts an array of classnames to be added to those specified in the attribute
As for the override of the name of the attribute I need some advice, I don't know what's meant with this

core/modules/widgets/browse.js
BurningTreeC: Refactored into widget.js @Jermolene

core/modules/widgets/select.js
```diff
@@ -138,7 +142,6 @@ SelectWidget.prototype.execute = function() {
 	this.selectTitle = this.getAttribute("tiddler",this.getVariable("currentTiddler"));
 	this.selectField = this.getAttribute("field","text");
 	this.selectIndex = this.getAttribute("index");
-	this.selectClass = this.getAttribute("class");
```
pmario, line 141: I don't think, that this should be moved to the widget.js ... If we would do it that way for every other widget parameter, it would overload widget.js. ... These elements have to be passed to `this.assignDomNodeClasses(options)` in the options parameter as Jeremy suggested. 

This issue is valid for every other widget, where you removed the parameter handling from `.execute`

Jermolene: Hi @pmario @BurningTreeC perhaps there is an issue here: by removing the line from `.execute()` it means that one can no longer see all the attributes at a glance. I think @pmario is suggesting that we keep lines like this, and pass `this.selectClass` to the new methods.

pmario: That's right and if we use an `options` object, we are able to expand the parameters in the future, without the need to change the function signature

BurningTreeC: Ok, the `assignDomNodeClasses` method calls `this.getAttribute` to get the class(es) ... I'm a bit confused what passing the class to the method should do in that regard.
The `assignDomNodeClasses` and `assignDomNodeStyles` methods call `this.getAttribute` because the styles and classes can change dynamically and we update them with low overhead.
Why should we want to get the `class` and `style` attributes in the `execute()` method and pass them to the two new methods if the methods themselves are already designed to handle that functionality?

pmario: the line: 

```
var classes = this.getAttribute("class",fallbackClass).split(" ");
```

should be removed from the widget.js file. It doesn't belong there. The widget base-class should provide generic functionality, which should be available in any widget, that inherits it. ... But it should not do widget specific logic. ... 

IMO reading parameters that belong to a widget is widget-specific logic. ... 

The same is true for the "style" parameter. ... 

---------

The second reason is, that it is extremely confusing if parameters are missing in the widget code, because they are handled in the base-widget. ... One of the first things, we do if we need to debug a problem with widgets, is to examine the `.execute()` function of the widget, to see if something is wrong with the parameters. 

But now some parameters are handled in a completely different part of the code tree. That's not consistent.  

If we would follow your example here, we could completely eliminate all widget files and handle all the logic in 1 monster widget.js file ... But that's not what we want.

pmario:
Why do we need a "styles" attribute for widgets. It will cause a lot of problems for theme creators!

core/modules/widgets/select.js
BurningTreeC: Yes @pmario I understand that, but it was requested to refactor the two methods into widget.js

Jermolene:
> Why do we need a "styles" attribute for widgets. It will cause a lot of problems for theme creators!

It's very useful in some situations; the button, reveal and edit-shortcut widgets already support a "style" attribute.

I don't see how it causes any more problems for theme creators than does, say, the "style" attribute on div or span elements. There's always been tension between CSS classes and direct styling, and that tension is part of HTML authoring. I don't think it's TiddlyWiki's job to resolve that tension one way or the other, but to allow authors to explore all the features of HTML.

core/modules/widgets/select.js
Jermolene: The problem here is that things like storing the domNode in the property "domNode" is not currently enforced by the widget baseclass; it's just a convention. The refactoring that I've suggested requires that the convention be baked into the baseclass.

I think all of this may be a good deal easier when we move to the proposed declarative approach for widget definition. Until then, I'm not sure we can do this without an unacceptable amount of code duplication.

BurningTreeC: @Jermolene , how does the declarative approach for the widget definition look like? The words themselves don't say anything to me

Jermolene: There's some recent discussion here: https://github.com/Jermolene/TiddlyWiki5/discussions/5265#discussioncomment-203399 and some history here https://github.com/Jermolene/TiddlyWiki5/issues/1903#issuecomment-145280856

BurningTreeC: Ok I think I now understand what the declarative approach is supposed to do

PR #5931, opened 2021-08-03
Do not add X-Requested-With header for simple requests

### LINKED ISSUE #4537: should / why is ??!!  _canonical_uri making an options request
after adding  _canonical_uri  field and setting url
leaving the content type field empty

after  confirming changes
a request is made to the resource location
with 
> Request method:OPTIONS

setting the type to text plain and confirming changes 
no request is made ?? =S

same as the above with content type  markdown

setting content type to html makes a get request
which returns 
content-type: text/plain; charset=utf-8
and displays a empty frame
if i view source on the frame i see the uri content 

i guess im ether doing something wrong !??
 or something is a bit squewiff!

=]

any one have any hints on
what im missing in this case 
or 
 exactly how/where the _canonical_uri  creates the request is made

saqimtiaz:
Fixes #4537 by not adding `X-Requested-With` header for [http requests that should not trigger a CORS preflight.](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS#simple_requests)

`$tw.utils.http()` currently always adds the `X-Requested-With` header to all requests, which in turn triggers an `OPTIONS` preflight even when it could be avoided. The changes herein modify the method so that the `X-Requested-With` header is only added for requests that would not satisfy the criteria for avoiding the CORS preflight.

The node.js server currently requires the `X-Requested-With` header for `POST` requests which with the above change would no longer have this header. `PUT` and `DELETE` requests have the same requirement but are unaffected by this change.

https://github.com/Jermolene/TiddlyWiki5/blob/82ec63e71118f9ee788972371f713579bae67530/core/modules/server/server.js#L242-L248

Rather than removing this requirement in the server, I have modified the tiddlyweb adaptor to manually add the `X-Requested-With` header for `POST` requests. 

The alternative would be to remove the requirement in the server for the `X-Requested-With `header for `POST` requests, which we probably do not want to do.

### DIFF
--- a/core/modules/utils/dom/http.js
+++ b/core/modules/utils/dom/http.js
@@ -34,6 +34,23 @@ exports.httpRequest = function(options) {
 			});
 			return result;
 		},
+		getHeader = function(targetHeader) {
+			return headers[targetHeader] || headers[targetHeader.toLowerCase()];
+		},
+		isSimpleRequest = function(type,headers) {
+			if(["GET","HEAD","POST"].indexOf(type) === -1) {
+				return false;
+			}
+			for(var header in headers) {
+				if(["accept","accept-language","content-language","content-type"].indexOf(header.toLowerCase()) === -1) {
+					return false;
+				}
+			}
+			if(hasHeader("Content-Type") && ["application/x-www-form-urlencoded","multipart/form-data","text/plain"].indexOf(getHeader["Content-Type"]) === -1) {
+				return false;
+			}
+			return true;	
+		},
 		returnProp = options.returnProp || "responseText",
 		request = new XMLHttpRequest(),
 		data = "",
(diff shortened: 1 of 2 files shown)


core/modules/utils/dom/http.js
```diff
@@ -34,6 +34,23 @@ exports.httpRequest = function(options) {
 			});
 			return result;
 		},
+		getHeader = function(targetHeader) {
+			return headers[targetHeader] || headers[targetHeader.toLowerCase()];
+		},
+		isSimpleRequest = function(type,headers) {
+			if(["GET","HEAD","POST"].indexOf(type) === -1) {
```
pmario, line 41: Should there be a `indexOf(type.toUpperCase())` ... just in case?

saqimtiaz: HTTP methods (verbs) are case sensitive, whereas headers are not.

Jermolene:
Thanks @saqimtiaz

Issue #4375, opened 2019-11-15
KaTeX global macros do not work as expected


KendrickLamarck:
There seem to be a couple of distinct but possibly related issues with defining macros that persist across math expressions with `\gdef` or `\global\def` using the KaTeX plugin (I tested mostly with `\gdef`).

## Issue 1:

`\gdef` only seems to work at all if the `macros` list is present in the options being passed to `katex.render`.

### Steps to reproduce

If this option is not present in the `options` list in **$:/plugins/tiddlywiki/katex/wrapper.js**, `\gdef` just seems to behave like `\def`, i.e. if we have a tiddler containing

```
$$
\gdef\RP{\R\mathrm P}
\RP
$$

$$
\RP
$$
```

only the first instance of `\RP` would render correctly and the second one would show as the red error text.

### Workaround

If we add `macros: {}` to the end of `options`in **$:/plugins/tiddlywiki/katex/wrapper.js**, it works, at least within the same tiddler. See **Issue 2**.

## Issue 2:

Macros defined with `\gdef` (using the workaround in **Issue 1**) only work with tiddlers that are in the story river at the same time.

### Steps to reproduce

Say we have Tiddler1 containing

```
$$\gdef\RP{\R\mathrm P}$$
```

and Tiddler2 containing

```
$$\RP$$
```

then, if we reload the page, Tiddler2 will only show the red error text.

### Workaround(s)

Tiddler2 will render correctly if we also open Tiddler1, click the edit button of Tiddler2 and then click the done button.

A workaround for this is to transclude Tiddler1 at the start of Tiddler2, but doing this for every Tiddler seems less than ideal.

## Conclusion

Having a mechanism to define macros globally would be extremely useful.

A first feature request would be to make it possible to define a macro list to be passed to `katex.render` without having to edit **$:/plugins/tiddlywiki/katex/wrapper.js**. An issue with this is that I don't think it's possible to define macros that take arguments in this way.

Secondly, it would be great if one could simply create a Tiddler containing all macros that one wants to use across the Wiki defined with `\gdef` and use them in all other Tiddlers without having to transclude.

Lastly, I apologize if these should be separate issues, but they seem connected. It also seems possible that at least **Issue 1** is an issue with KaTeX itself and not TiddlyWiki.

Jermolene:
Hi @KendrickLamarck thanks for the clear report. I think both issues are limitations of the integration with TiddlyWiki, rather than bugs with KaTeX.

[...]

PR #5933, opened 2021-08-04
Add custom macro editor for KaTeX plugin

8d1h:
...as requested in https://github.com/Jermolene/TiddlyWiki5/issues/4375, with a nice UI :)

katex

**Update.** Added a few tweaks: no need to refresh for new macros now, and macros with arguments can be added with `#` (`\dd#1#2` in the above example).

### DIFF
--- a/plugins/tiddlywiki/katex/config.tid
+++ b/plugins/tiddlywiki/katex/config.tid
@@ -1,0 +1,13 @@
+title: $:/plugins/tiddlywiki/katex/config
+
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/katex/config]!has[draft.of]]">
+
+<div>
+
+!! <$link><$transclude field="caption"/></$link>
+
+<$transclude>
+
+</div>
+
+</$list>
--- a/plugins/tiddlywiki/katex/config.tid
+++ b/plugins/tiddlywiki/katex/config.tid
@@ -1,6 +1,6 @@
 title: $:/plugins/tiddlywiki/katex/config
 
-<$list filter="[all[shadows+tiddlers]tag[$:/tags/katex/config]!has[draft.of]]">
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/KaTeX/Config]!has[draft.of]]">
 
 <div>
 
--- a/plugins/tiddlywiki/katex/config.tid
+++ b/plugins/tiddlywiki/katex/config.tid
@@ -1,5 +1,7 @@
 title: $:/plugins/tiddlywiki/katex/config
 
+<div class="tc-control-panel">
+
 <$list filter="[all[shadows+tiddlers]tag[$:/tags/KaTeX/Config]!has[draft.of]]">
 
 <div>
@@ -11,3 +13,5 @@ title: $:/plugins/tiddlywiki/katex/config
 </div>
 
 </$list>
+
+</div>
--- a/plugins/tiddlywiki/katex/wrapper.js
+++ b/plugins/tiddlywiki/katex/wrapper.js
@@ -16,6 +16,16 @@ var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js"),
     chemParse = require("$:/plugins/tiddlywiki/katex/mhchem.min.js"),
 	Widget = require("$:/core/modules/widgets/widget.js").widget;
 
+katex.macros = {};
+(function() {
+	var tid, m, v;
+	for (const t of $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro")) {
+		tid = $tw.wiki.getTiddler(t);
+		m = tid.fields["macro"];
+		v = tid.fields["text"];
+		katex.macros[m] = v;
+	};
+})();
 var KaTeXWidget = function(parseTreeNode,options) {
 	this.initialise(parseTreeNode,options);
 };
@@ -38,7 +48,7 @@ KaTeXWidget.prototype.render = function(parent,nextSibling) {
 	var displayMode = this.getAttribute("displayMode",this.parseTreeNode.displayMode || "false") === "true";
 	// Render it into a span
 	var span = this.document.createElement("span"),
-		options = {throwOnError: false, displayMode: displayMode};
+		options = {throwOnError: false, displayMode: displayMode, macros: katex.macros};
 	try {
 		if(!this.document.isTiddlyWikiFakeDom) {
 			katex.render(text,span,options);
--- a/plugins/tiddlywiki/katex/wrapper.js
+++ b/plugins/tiddlywiki/katex/wrapper.js
@@ -24,7 +24,7 @@ katex.updateMacros = function() {
 	for (var i=0; i < tiddlers.length; i++) {
 		tid = $tw.wiki.getTiddler(tiddlers[i]);
 		try {
-			macro = tid.fields["macro"];
+			macro = tid.fields["caption"];
 			macro = macro.replace(regex, "");
 			cmd = tid.fields["text"];
 			katex.macros[macro] = cmd;
--- a/plugins/tiddlywiki/katex/wrapper.js
+++ b/plugins/tiddlywiki/katex/wrapper.js
@@ -18,10 +18,11 @@ var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js"),
 
 katex.macros = {};
 katex.updateMacros = function() {
-	var tid, macro, cmd;
-	const regex = /#\d/g; // Remove the arguments like #1#2
-	for (const t of $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro")) {
-		tid = $tw.wiki.getTiddler(t);
+	var tiddlers = $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro"),
+		regex = /#\d/g, // Remove the arguments like #1#2
+		tid, macro, cmd;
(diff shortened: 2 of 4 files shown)


plugins/tiddlywiki/katex/wrapper.js
```diff
@@ -16,6 +16,19 @@ var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js"),
     chemParse = require("$:/plugins/tiddlywiki/katex/mhchem.min.js"),
 	Widget = require("$:/core/modules/widgets/widget.js").widget;
 
+katex.macros = {};
+katex.updateMacros = function() {
+	var tid, macro, cmd;
+	const regex = /#\d/g; // Remove the arguments like #1#2
+	for (const t of $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro")) {
```
pmario, line 23: In TW core we use ES5 javascript only. So you need to use `var` instead of `const` and the `for of` loop was introduced in ES6. So you need to use the "old way"

```
var tiddlers = $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro") ;
for (var i=0; i < tiddlers.length; i++) {
//your code changed a bit
}
```

pmario:
@8d1h did you sign the TW CLA already? https://github.com/Jermolene/TiddlyWiki5/blob/master/licenses/cla-individual.md If not, you should do so in a different PR.

plugins/tiddlywiki/katex/wrapper.js
8d1h: Thanks! It's fixed.

pmario:
@Jermolene Here is a version created with the latest pre-release code [katexdemo.zip](https://github.com/Jermolene/TiddlyWiki5/files/6938061/katexdemo.zip)

It seems to work as proposed. @8d1h You may create a post in the google group, so others users may test this. So we can see if there any problems

plugins/tiddlywiki/katex/config.tid
```diff
@@ -0,0 +1,13 @@
+title: $:/plugins/tiddlywiki/katex/config
+
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/katex/config]!has[draft.of]]">
```
Jermolene, line 3: For consistency, the tag should be `$:/tags/KaTeX/Config`

plugins/tiddlywiki/katex/wrapper.js
```diff
@@ -16,6 +16,23 @@ var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js"),
...
+katex.macros = {};
+katex.updateMacros = function() {
+	var tiddlers = $tw.wiki.getTiddlersWithTag("$:/tags/katex/macro"),
+		regex = /#\d/g, // Remove the arguments like #1#2
+		tid, macro, cmd;
+	for (var i=0; i < tiddlers.length; i++) {
+		tid = $tw.wiki.getTiddler(tiddlers[i]);
+		try {
+			macro = tid.fields["macro"];
+			macro = macro.replace(regex, "");
```
Jermolene, line 28: Why do we remove the `#1`, `#2` from the "macro" string?

Jermolene (commented):
Thank you @8d1h, it's fantastic when we get such complete submissions from new contributors.

I have a couple of minor questions below. More significantly, I'm wondering if we might not be better off taking advantage of our existing macro/variable system.

The idea would be to define a KaTeX macro in the same way as an ordinary wikitext macro:

```
\define \ZZ() \mathbb{Z}
```

Such macros could be defined locally or globally.

The KaTeX wrapper would then construct the macros from the in-scope variables whose name starts with a backslash.

This approach would mean that macros could be dynamically redefined without needing to reload.

8d1h:
> The KaTeX wrapper would then construct the macros from the in-scope variables whose name starts with a backslash.
> 
> This approach would mean that macros could be dynamically redefined without needing to reload.

Thanks! I think that is indeed a better choice in theory. However, after a bit of thought, this would basically mean that we need to write a parser for the LaTeX syntax itself.

For example, the second macro I listed could be used as `\dd{\sin{x}}{x}` for dsin(x)/dx. We will then need to parse the two parts `\sin{x}` and `x` to be able to generate the correct result `\frac{d\sin{x}}{d x}`. The macro calls could also be nested, like d(dsinx/dx)/dx, then things could get a lot more complicated. So right now I'm just letting KaTeX do all the heavy-lifting :)

In short, this editor is just a nicer interface for defining macros in KaTeX. It would require a lot more work to properly parse the LaTeX syntax, and probably not worth it since we are already using KaTeX to do this for us...

plugins/tiddlywiki/katex/config.tid
8d1h: Ok I will fix this.

plugins/tiddlywiki/katex/wrapper.js
8d1h: The KaTeX macro variable does not accept inputs like `\dd#1#2`, however I think it is more natural to allow this in the definition, since usually one writes `\def\dd#1#2{\frac...}`. So now both `\dd` and `\dd#1#2` can be put in the macro editor, and KaTeX will only get `\dd` in either case.

8d1h:
Maybe I can put the macro editor in the KaTeX editor toolbar, besides the snippets, then we won't need to make definitions in individual files and can just add them from the toolbar.

kookma:
@8d1h - each macro uses a new tiddler like 


`$:/plugins/tiddlywiki/katex/macros/macro`
`$:/plugins/tiddlywiki/katex/macros/macro 1`
`$:/plugins/tiddlywiki/katex/macros/macro 2`

I think in TW 5.0.2 you have options to create more meaningful name
also each tiddler has a new `macro` field. I may recommend to use caption which is a standard field here

I believe these tiddlers can be more semantic and tidy!

One crude proposal:
1. every macro has a name so you can use `$:/plugins/tiddlywiki/katex/macros/`+name
2. the KateX macro then can create from `$:/plugins/tiddlywiki/katex/macros/name` and what exists in text field!

8d1h:
> @8d1h - each macro uses a new tiddler like
> 
> `$:/plugins/tiddlywiki/katex/macros/macro`
> `$:/plugins/tiddlywiki/katex/macros/macro 1`
> `$:/plugins/tiddlywiki/katex/macros/macro 2`
> 
> I think in TW 5.0.2 you have options to create more meaningful name
> also each tiddler has a new `macro` field. I may recommend to use caption which is a standard field here
> 
> I believe these tiddlers can be more semantic and tidy!
> 
> One crude proposal:
> 
>     1. every macro has a name so you can use `$:/plugins/tiddlywiki/katex/macros/`+name
> 
>     2. the KaTeX macro then can be created from `$:/plugins/tiddlywiki/katex/macros/name` and what exists in text field!

Thanks for the comments! Yes you are right I was being lazy and trying to avoid naming them properly! :) They are now named properly after the macro, and I also changed the field from `macro` to `caption`.

pmario:
@8d1h ... If you intend to add functions and / or tweak the project, you should make the PR a "draft". 

grafik

Once you are finished and the PR should be merged you convert it back to a PR that should be reviewed. This button is near the end of the conversation 

grafik

Jermolene:
Thanks for your patience @8d1h

kookma:
@8d1h - with the release of official KaTeX 0.13.18 I would recommend to add documentation for custom macros to katex official plugin page!

Also see https://talk.tiddlywiki.org/t/tiddlywiki-5-2-and-katex-0-13-18/704/7

pmario:
> Also see https://talk.tiddlywiki.org/t/tiddlywiki-5-2-and-katex-0-13-18/704/7

I do like the "talk" links way more than GG links. They actually mean something.

8d1h:
> @8d1h - with the release of official KaTeX 0.13.18 I would recommend to add documentation for custom macros to katex official plugin page!

@kookma I just saw your TiddlyWiki of KaTeX demos, which is very nice! (I guess it's also useful for LaTeX users in general.) You should go ahead and update the official demo.

> Why not just `$:/plugins/tiddlywiki/katex/macros/ergun`? You can trim `\` from macro name and then add as suffix to `$:/plugins/tiddlywiki/katex/macros/`

Actually I was also thinking of doing that, but KaTeX actually supports redefining a single letter: e.g. you can create a macro that maps `a` (without backslash) to `bbbbb` (this is certainly a very bad practice), so I kept the backslash as part of the macro.

8d1h:
BTW in your macro example, I'd say for very complicated formula it's better to use the stamp functionality (the katex logo that can be seen from the editor), since one can insert the formula easily while still be able to edit it in place. The macros are most useful for small combinations of symbols that one uses a lot and might be tedious to type each time. For example I think this [wiki](https://mathematik.tiddlyspot.com/#%24%3A%2F_macros%2FKaTeX) has a very typical usecase

kookma:
@8d1h - when click `add` or `import` it is better to clear the input box specially if the addition is successful!
Right now when you work with a long list it is very confusing and you had to see all macros again to see if your entry is added!

8d1h:
Yes you are right. Maybe one better solution is to add the option to sort the macros by creation time? So the last added ones will be at the end of the list.

kookma:
> Yes you are right. Maybe one better solution is to add the option to sort the macros by creation time? So the last added ones will be at the end of the list.

Good! you can also add a button on header to sort them! by the way I may recommend to clear the box (here you have used a temp tiddler) after successful addition!

PR #5936, opened 2021-08-05
Fix syncer logout alert

joshuafontany:
Doing a deep dive into the syncer still for the Bob multiplayer feature.

Caught one more instance where it should check for a localized error message.


Jermolene:
Thanks @joshuafontany

Issue #1554, opened 2015-03-06
Importing should register in Recent


twMat:
Importing, say, an image via drag'n drop should be reflected in the Recent list.

sukima:
There is already an import list provided. Would this be to enhance that list because it needs to be recent or not.

I could see an counter argument that you don't want to change recent list during a import. But instead to add to recent as you click on your chosen tiddlers from the import summary page.

pmario:
imo both behaviours make sense for file based TiddlyWikis. ... The server version needs to update the modified field on import, for consistency reasons.

twMat:
IMO, the import list is of value - but it is not instead of the Recent listing and, as far as I understand, the import list gets replaced for each new import so it's not possible to see its history. (That may be desirable though?)

I try to look at things from a pragmatic perspective; the Recent list is there (I think) to easify access by using our "sense of chronolgy" for when it was we dealt with something. It would be (is!) odd to force the user to remember exactly *how* a tiddler came into existence; the ease of access should be regardless if it was created from scratch, copy-pasted field-by-field, or dragged in. What's the point in separating these?

>I could see an counter argument that you don't want to change recent list during a import. But instead to add to recent as you click on your chosen tiddlers from the import summary page.

I'm curious on which those counter arguments would be. And are you suggesting titles should be added to Recent merely by cliking on them (i.e not by editing them)?

sukima:
Good point. To clarify I was playing devil's advocate. So my personal opinion is this could be a useful feature worth looking into.

Jermolene:
> Importing, say, an image via drag'n drop should be reflected in the Recent list.

That implies that importing a tiddler would reset its modified time. That seems destructive; in some cases, the last modification date of a tiddler is useful metadata about when a tiddler was last updated.

We could perhaps set the created/creator/modified/modifier fields if they are not present on the incoming tiddler? But I'm concerned that there are other consequences too: importing a plugin would change the modified date, which would mean that the plugin tiddler would no longer be 100% the same as the original source.

inmysocks:
Would it be hard to add a tiddler that holds the titles of the previous n most recently imported tiddlers in its list field? That way there wouldn't be any changes to imported tiddlers.

You could add something like this to the import button to update the list:

<$set name=UpdatedList filter='[[$:/RecentlyImported]list[]]<<CurrentImportedFilterList>>+[last[n]]'>
[...]

PR #5947, opened 2021-08-09
partially-fix-1554-timeline-macro

pmario:
This PR will allow users to modify the Recent-tab without modify the core tiddler. See discussion in #1554

### DIFF
--- a/core/wiki/macros/timeline.tid
+++ b/core/wiki/macros/timeline.tid
@@ -3,14 +3,10 @@ modified: 20141212110330815
 tags: $:/tags/Macro
 title: $:/core/macros/timeline
 
-\define timeline-title()
-\whitespace trim
-<!-- Override this macro with a global macro 
-     of the same name if you need to change 
-     how titles are displayed on the timeline 
-     -->
-<$view field="title"/>
-\end
+<!-- Override one or both of the following 2 macros with a global macro -->
+<!-- of the same name if you need to change how titles are displayed on the timeline -->
+\define timeline-title() <$view field="title"/>
+\define timeline-link() <$link to={{!!title}}><<timeline-title>></$link>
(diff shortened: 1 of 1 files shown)


core/wiki/macros/timeline.tid
```diff
@@ -3,22 +3,18 @@ modified: 20141212110330815
...
-\define timeline-title()
-\whitespace trim
-<!-- Override this macro with a global macro 
-     of the same name if you need to change 
-     how titles are displayed on the timeline 
-     -->
-<$view field="title"/>
-\end
+<!-- Override one or both of the following 2 macros with a global macro -->
+<!-- of the same name if you need to change how titles are displayed on the timeline -->
```
Jermolene, line 7: Best to use a single comment, also a couple of tweaks:

```
<!-- Override one or both of the following two macros with a global or local macro of the same name if you need to change how titles are displayed on a timeline -->
```

core/wiki/macros/timeline.tid
```diff
@@ -3,22 +3,19 @@ modified: 20141212110330815
...
 
-\define timeline-title()
-\whitespace trim
-<!-- Override this macro with a global macro 
-     of the same name if you need to change 
-     how titles are displayed on the timeline 
-     -->
-<$view field="title"/>
-\end
+<!-- Override one or both of the following two macros with a global or local macro of the same name
```
Jermolene, line 6: I don't think one can override a macro definition from the same tiddler with a global macro definition. Does it definitely work?

pmario: Did just test this. It definitely works with the changes here, since user macros are initialized after core macros. I'll need to test the other PR

pmario:
@Jermolene ... I did test is just now with latest TW version and is does work as expected.

Jermolene:
Thank you @pmario, and apologies for the delay

PR #5948, opened 2021-08-09
make-open-bar-more-customizable

pmario:
This PR makes the Open-tab more customizable for users, without the need to modify the core macro

### DIFF
--- a/core/ui/SideBar/Open.tid
+++ b/core/ui/SideBar/Open.tid
@@ -23,10 +23,20 @@ $button$
 </$droppable>
 \end
 
+\define open-tab-close-tiddler-button()
+<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini tc-small-gap-right">{{$:/core/images/close-button}}</$button>
+\end
+
+<!-- Override one or both of the following 2 macros with a global macro -->
+<!-- of the same name if you need to change how elements are displayed  -->
+\define open-tab-link()<$link to={{!!title}}><$view field="title"/></$link>
+
+\define open-tab-element()<<open-tab-close-tiddler-button>><<open-tab-link>>
+
 <div class="tc-sidebar-tab-open">
 <$list filter="[list<tv-story-list>]" history=<<tv-history-list>> storyview="pop">
 <div class="tc-sidebar-tab-open-item">
-<$macrocall $name="droppable-item" button="""<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini tc-small-gap-right">{{$:/core/images/close-button}}</$button><$link to={{!!title}}><$view field="title"/></$link>"""/>
+<$macrocall $name="droppable-item" button="<<open-tab-element>>"/>
 </div>
 </$list>
 <$tiddler tiddler="">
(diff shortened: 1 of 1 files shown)


pmario:
Also see comment at: "Importing should register in Recent" -- https://github.com/Jermolene/TiddlyWiki5/issues/1554#issuecomment-895064392

core/ui/SideBar/Open.tid
```diff
@@ -23,10 +23,20 @@ $button$
 </$droppable>
 \end
 
+\define open-tab-close-tiddler-button()
+<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini tc-small-gap-right">{{$:/core/images/close-button}}</$button>
+\end
+
+<!-- Override one or both of the following 2 macros with a global macro -->
+<!-- of the same name if you need to change how elements are displayed  -->
```
Jermolene, line 31: Here I think it's cleaner to have a single comment:

```
<!--
Override one or both of the following 2 macros with a global macro of the same name if you need to change how elements are displayed  -->
```

Jermolene, line 31: Here I think it's cleaner to have a single comment with a couple of cleanups:

```
<!--
Override one or both of the following two macros with a global or local macro of the same name to change how list items are displayed  -->
```

Jermolene (commented):
Thanks @pmario

core/ui/SideBar/Open.tid
Jermolene: My mistake, of course here it only makes sense to override these macros globally.

core/ui/SideBar/Open.tid
```diff
@@ -23,10 +23,20 @@ $button$
...
+\define open-tab-link()<$link to={{!!title}}><$view field="title"/></$link>
```
Jermolene, line 32: We would normally have a space after the `open-tab-link()` and before the definition.

core/ui/SideBar/Open.tid
```diff
@@ -23,10 +23,20 @@ $button$
...
+
+\define open-tab-element()<<open-tab-close-tiddler-button>><<open-tab-link>>
+
 <div class="tc-sidebar-tab-open">
 <$list filter="[list<tv-story-list>]" history=<<tv-history-list>> storyview="pop">
 <div class="tc-sidebar-tab-open-item">
-<$macrocall $name="droppable-item" button="""<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini tc-small-gap-right">{{$:/core/images/close-button}}</$button><$link to={{!!title}}><$view field="title"/></$link>"""/>
+<$macrocall $name="droppable-item" button="<<open-tab-element>>"/>
```
Jermolene, line 39: I think the words "list-item" might be better than "element" here

pmario: The problem is, that we have to have global macros. So I would like to use a name, that works for the core, but doesn't clash if a user has a similar one. IMO `list-item` may be easy to clash with future user names.

pmario:
@Jermolene .. I did change the comments and the macro name as requested. The tests have failed because of some connection issues to github. So there shouldn't be  a problem with the code.

core/ui/SideBar/Open.tid
```diff
@@ -23,10 +23,20 @@ $button$
 </$droppable>
 \end
 
+\define open-tab-close-tiddler-button()
+<$button message="tm-close-tiddler" tooltip={{$:/language/Buttons/Close/Hint}} aria-label={{$:/language/Buttons/Close/Caption}} class="tc-btn-invisible tc-btn-mini tc-small-gap-right">{{$:/core/images/close-button}}</$button>
+\end
+
+<!-- Override one or both of the following 2 macros with a global macro of the same name
```
Jermolene, line 30: Is it actually possible to override a local macro definition with a global macro? I don't think it will work here.

pmario: You are right. This one doesn't work, since it is tagged: `$:/tags/SideBar` ... which I obviously did confuse with tags: `$:/tags/Macro`

pmario:
@Jermolene .. I'll convert this one to a draft, since it is more complicated as I thought. So it will have to wait :/

PR #5963, opened 2021-08-18
Remove stray characters in example of :map prefix

sobjornstad:
Nitpick fix in the new prerelease documentation – an extra, invalid `get[` is included in the alternative example, presumably due to a copy-paste error.


sobjornstad:
@Jermolene or anyone with access to trigger the workflow for the first time – have a chance to look at this? Should be a trivial fix.

Jermolene:
Thanks @sobjornstad delighted to welcome you as a contributor – apologies but your CLA signature needs to be in a separate PR targeting the tiddlywiki-com branch.

sobjornstad:
Opened #6049 and updated this PR to only include the documentation change.

Jermolene:
Thanks @sobjornstad

Jermolene:
Hi @sobjornstad apologies entirely my mistake: pure docs changes like this should be targetted to the `tiddlywiki-com` branch so that they go live immediately. I think it's OK in this case because v5.2.0 is just about to be released.

sobjornstad:
Ah, got it! Will do that next time.

PR #5967, opened 2021-08-20
Add `m` flag for multiline to search-replace filter operator

BurningTreeC:


saqimtiaz:
@BurningTreeC could you please add some tests?

BurningTreeC:
If I knew how to do that @saqimtiaz I would do so

saqimtiaz:
Looks like you found the tests.

You can run the tests locally with `node tiddlywiki editions/test`

Recommend adding two tests, one with the text being searched being a single line and one with multiple lines.

BurningTreeC:
> Looks like you found the tests.
> 
> You can run the tests locally with `node tiddlywiki editions/test`
> 
> Recommend adding two tests, one with the text being searched being a single line and one with multiple lines.

Yes I found them but I have a hard time getting the right results. But I'll make it :+1:

PR #5968, opened 2021-08-20
Add `m` flag for multiline matches to search-replace filter operator

BurningTreeC:
This PR adds the multiline flag to the search-replace operator
It also updates documentation and adds two tests

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -119,7 +119,7 @@ exports["search-replace"] = function(source,operator,options) {
 	var results = [],
 		suffixes = operator.suffixes || [],
 		flagSuffix = (suffixes[0] ? (suffixes[0][0] || "") : ""),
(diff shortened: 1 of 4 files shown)


BurningTreeC:
Is there anything that speaks against this PR?

saqimtiaz:
There are a few changes that I would like to see:

- more representative tests. For example, one test without the `m` flag that only replaces a word at the beginning of a string, and then the same test with the `m` flag that replaces the same word at the beginning of every string.
- additions to the documentation examples. You can most likely come up with one good example and use it both for documentation and for the tests I described above.

BurningTreeC:
@saqimtiaz - I've added / changed the tests like you recommended
Then I've also added an example in the documentation
I cannot reproduce that example as a test because it's using the HelloThere tiddler (it gets the text from it)
But I have a similar test where everything BUT a match is deleted

Jermolene:
Thanks @BurningTreeC

PR #5969, opened 2021-08-22
Make prefix and suffix for editor-toolbar stamps work for ...

BurningTreeC:
... shadow tiddlers

This PR makes the stamp snippets of the `stamp` editor-toolbar button that add a `prefix` and a `suffix` work with `shadow` tiddlers

### DIFF
--- a/core/ui/EditorToolbar/stamp-dropdown.tid
+++ b/core/ui/EditorToolbar/stamp-dropdown.tid
@@ -3,7 +3,7 @@ title: $:/core/ui/EditorToolbar/stamp-dropdown
 \define toolbar-button-stamp-inner()
 <$button tag="a">
 
-<$list filter="[[$(snippetTitle)$]addsuffix[/prefix]is[missing]removesuffix[/prefix]addsuffix[/suffix]is[missing]]">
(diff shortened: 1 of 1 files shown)


pmario:
Sorry. I have no idea, what you description says. Can you write down an example "before and after"?

BurningTreeC:
Hi @pmario 
There's this "hidden" functionality for editor stamps. We can create snippets that add a `prefix` and a `suffix` to the selection by first creating a tiddler tagged `$:/tags/TextEditor/Snippets` and then creating two tiddlers with the same title but with the suffix `/suffix` and the suffix `/prefix`. If those tiddlers exist, the text in the tiddler with the `/prefix` in the title gets inserted as prefix and the text in the tiddler with the `/suffix` in the title gets inserted as suffix.

The condition to test if those tiddlers with the suffix `/prefix` and the suffix `/suffix` in their title exist was by using the filter operator `!is[missing]` but if a plugin for example ships such a Snippet the condition fails because it's a shadow tiddler. This PR fixes this problem

BurningTreeC:
closing this in favour of #6003 which resolves the conflicts

PR #5973, opened 2021-08-22
this fixes some typos, how to hide viewtemplate toolbar buttons

pmario:
This PR fixes some typos, how to hide viewtemplate toolbar buttons. Since the text is English I was able to fix them in all the docs that use the wrong description.


pmario:
uuups, wrong path :/ close it

PR #5974, opened 2021-08-22
fix some typos in the ViewTolbar button docs

pmario:
This PR fixes a type in the ViewTemplate toolbar buttons docs. Since the configuration text is English I could fix all languages that contain the issue.


Jermolene:
Thanks @pmario

PR #5987, opened 2021-08-25
Make plugin-info states system tiddlers

BurningTreeC:
When we currently open the `plugin-info` of the `$:/core/ui/Components/plugin-info` tiddler (and plugins that use it as a template) an annoying state tiddler gets created that shows up under the `Recent` tab. The states of the `$:/core/ui/Components/plugin-info` aren't prefixed with `$:/state`

This PR solves this problem with an easy fix

### DIFF
--- a/core/ui/Components/plugin-info.tid
+++ b/core/ui/Components/plugin-info.tid
@@ -3,7 +3,7 @@ title: $:/core/ui/Components/plugin-info
 \define lingo-base() $:/language/ControlPanel/Plugins/
 
 \define popup-state-macro()
-$(qualified-state)$-$(currentTiddler)$
+$:/state$(qualified-state)$-$(currentTiddler)$
 \end
 
 \define tabs-state-macro()


saqimtiaz:
I cannot reproduce this in 5.1.23 when toggling the plugin info of a plugin. It only happens when interacting with `$:/core/ui/Components/plugin-info` directly which is not an intended use case.

BurningTreeC:
Hi @saqimtiaz - I can reproduce this using `{{$:/plugins/.../myplugin||$:/core/ui/Components/plugin-info}}`

saqimtiaz:
If you are using it in a custom template of your own then you need to set the necessary variables

```
<$set name="qualified-state" value=<<qualify "$:/state/plugin-info">>>
{{$:/plugins/.../myplugin||$:/core/ui/Components/plugin-info}}
</$set>
```

saqimtiaz:
Your suggested change will result in state tiddlers with titles starting with `$:/state$:/state/plugin-info-` when used from the core templates.

BurningTreeC:
> Your suggested change will result in state tiddlers with titles starting with `$:/state$:/state/plugin-info-` when used from the core templates.

Ok, that's bad. I'm going to close this PR and see how to use `{{||$:/core/ui/Components/plugin-info}}` correctly

PR #5988, opened 2021-08-25
Make ctrl-click in image-picker edit the image tiddler

BurningTreeC:
This PR makes <kbc>ctrl</kbd>-click open the image tiddler from the image-picker for editing


pmario:
IMO this needs to be configurable. Ctrl-click a link in view mode opens the tiddler without navigating to it. .. 

So for me this would be confusing. I do expect the link dropdown to work the same way as any other link dropdown. 

You can try this in the search dropdown. Search something. Click some tiddlers ... they move in the story river. Now ctrl-click. ... They don't move. 

So it needs to be configurable.

BurningTreeC:
Ok @pmario - this may have been an immature idea

PR #5989, opened 2021-08-26
Popup actions - actions triggered when popups open and close

BurningTreeC:
This PR is pretty extensive and is here for Discussion about its features. It includes:

- Buttons can have the new attributes `popup-open-actions` and `popup-close-actions`
- The Text-Editor toolbar buttons (dropdowns) `Stamp` and `Picture` can be navigated using the keyboard
  - This is possible by a keyboard-shortcut-backup mechanism that backs up all kb shortcuts that use `Up`, `Down`, `Escape` and `Enter` when the popup opens and makes them have no effect. Then custom shortcuts for `Up`, `Down`, `Escape` and `Enter` become active by tagging them with `$:/tags/KeyboardShortcut` ... And when the popup closes this is done in reverse
- When (if) at some point we have a marker for the currently viewed tiddler we can reuse this mechanism
- The `rootWidget` listens for `tm-external-text-operation` messages (that may be triggered by global keyboard shortcuts) and invokes the hook `th-external-text-operation` which is handled in factory.js
- A `current-widget-id` macro is added which is used to identify the input that should handle the `th-external-text-operation` hook
- I probably forgot some points, I'll update this when they come back to my memory


BurningTreeC:
I'm closing this PR in favour of #5991

PR #5995, opened 2021-08-28
Add global-popups mechanism

BurningTreeC:
This PR adds a mechanism that stores `popup-info` of button widgets when they are rendered in `$tw.popup.registeredPopups`

The `<$action-popup>` widget then looks in its `invokeAction` method if the popup that should be triggered is stored in `$tw.popup.registeredPopups` and triggers that if so. That allows for having access to the original `domNode` and positioning the popup accordingly.

If no such globally registered popup is found, the `<$action-popup>` widget behaves like before


linonetwo:
Shouldn't we use something like https://react.dev/reference/react-dom/createPortal for popup and model?

core/modules/utils/dom/popup.js
```diff
@@ -163,7 +170,9 @@ Popup.prototype.cancel = function(level) {
 				popup.wiki.deleteTiddler(popup.title);
 			} else {
 				popup.wiki.deleteTiddler($tw.utils.parseTextReference(popup.title).title);
-        		}
+			}
+			var index = this.findPopup(this.registeredPopups,popup.title);
```
pmario, line 174: I do have a bad feeline about the `index` variable. If `.findPopup()` returns `-1` it will remove the last element of `registeredPopups`. This may be intended, but from my point of view that's a "side-effect" and seems to be error prone and complex to maintain. ... BUT I'm not 100% sure what's intended.

pmario:
What should this functionality do?

BurningTreeC:
closing this PR - this is outdated and I don't know anymore why I wanted this

PR #5997, opened 2021-08-28
Add focus-preservation mechanism

BurningTreeC:
This PR adds the focus-preservation mechanism like already tried in #5798, #5695 and #5414

It doesn't add this mechanism for modals since I don't know how to test it in that case. I'd need a good test for that


pmario:
Is this draft still valid, or is it replaced by an other PR?

PR #5998, opened 2021-08-28
Make editor-preview open on a per-tiddler basis

BurningTreeC:
This PR makes the editor preview open on a per-tiddler basis
That means that the preview toggle toggles the preview for the current tiddler only, not for all open tiddlers
This reduces the pretty heavy refreshing of all open tiddler-editors if there are many open in the story river


BurningTreeC:
@Jermolene - is this something that can be considered?

PR #5999, opened 2021-08-29
Adds $tw.utils.decodeURISafe and $tw.utils.decodeURIComponentSafe

joshuafontany:
Replaces all instances of the direct calls to these functions with ones wrapped in try/catch that will return the raw string if the decoding fails.


Jermolene:
Thanks @joshuafontany looks good

pmario:
There are several utility functions in boot.js see: https://github.com/Jermolene/TiddlyWiki5/blob/d095aa0182faea4047dbc7de8789110f1bf84c6c/boot/boot.js#L43

Jermolene:
Thanks @joshuafontany – it would be most consistent to move the definition of those methods into boot.js.

PR #6000, opened 2021-08-29
Some love for the translink macro

BurningTreeC:
This PR changes the translink macro a bit

- It adds the classes `tc-translink-outer` and `tc-translink-inner` and sets styles in vanilla/base accordingly
- It adds a `transclusion-toggle-button`
- It uses the new `current-widget-id` macro for the toggle states so that we can have more translink macros per tiddler without toggling the preview-status for all at once


core/wiki/macros/translink.tid
```diff
@@ -2,14 +2,30 @@ title: $:/core/macros/translink
...
+<$link to=<<__title__>>>
+<$text text=<<__title__>>/>
 </$link>
-<div style="border:1px solid #ccc; padding: 0.5em; background: white; foreground; black;">
-<$transclude tiddler="""$title$""" mode="$mode$">
-"<$text text="""$title$"""/>" is missing
+<$set name="state" value=<<current-widget-id "$:/state/translink">>>
+<span style="float: right;">
+<$reveal state=<<state>> type="nomatch" text="show">
+<$button class=<<tv-config-toolbar-class>> set=<<state>> setTo="show">
```
pmario, line 12: hardcoded texts as this are not possible in the core. It needs to be language specific. Eg: new text in Buttons.multids

pmario:
The "translink" makro was new to me. .. IMO there should be more docs there too.

core/wiki/macros/translink.tid
BurningTreeC: there is no hardcoded text as far as I see @pmario

BurningTreeC: ... aaah now I've found it, well ther is hardcoded text and It's already been there

pmario: sorry. I did add the comment to the wrong place. It's line 25. the "is missing" text

BurningTreeC: I've now changed it to display the "Missing tiddler ViewTemplate" if the tiddler is missing. That should solve the problem since it's multilingual

BurningTreeC:
@pmario I'll add some docs

PR #6003, opened 2021-08-29
Make stamp-dropdown items that are suffixed and prefixed ...

BurningTreeC:
... work for shadow tiddlers, too

This PR is about stamp-dropdown snippets where a snippet with the tag `$:/tags/TextEditor/Snippet` exists and one or both of the tiddlers with the same title as the snippet-tiddler but with the suffix `/prefix` and the suffix `/suffix` exitst

Before this PR this worked only for non-shadow tiddlers. Now it also works for shadow tiddlers


Jermolene:
Thanks @BurningTreeC

PR #6004, opened 2021-08-29
Add focus-popups mechanism (II)

BurningTreeC:
This PR does the following:

- It adds a `<$focus>` widget that allows for focusing its created domNode on render time
- It uses the `<$focus>` widget when the `stamp` and the `picture` dropdowns are created
- Keyboard shortcuts can then be uses to navigate the popups up/down, select an entry with <kbd>Enter</kbd> and close the popup with <kbd>Escape</kbd>. There's also a `ctrl-Enter` shortcut in the picture dropdown to toggle system-images


BurningTreeC:
- This PR also adds a `focus-editor` text operation
- It also updated the `<$link>` widget to refresh when its `class` attribute changes (refresh problem)
- It uses the `focus-editor` text operation where appropriate

BurningTreeC:
@Jermolene, this PR would be ready to have a look at if you have time

PR #6013, opened 2021-09-01
Update link-widget refresh method

BurningTreeC:
This PR updates the `<$link>` widgets refresh method
Maybe there's a reason it didn't refresh so often but I felt like it should refresh when those attributes change

### DIFF
--- a/core/modules/widgets/link.js
+++ b/core/modules/widgets/link.js
@@ -207,7 +207,10 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 LinkWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip) {
+	var hideMissingLinks = (this.getVariable("tv-show-missing-links") || "yes") === "no";
+	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip ||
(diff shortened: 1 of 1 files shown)


core/modules/widgets/link.js
```diff
@@ -207,7 +207,10 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 LinkWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip) {
+	var hideMissingLinks = (this.getVariable("tv-show-missing-links") || "yes") === "no";
+	if(changedAttributes.to || changedTiddlers[this.to] || changedAttributes["aria-label"] || changedAttributes.tooltip ||
+		changedAttributes["class"] || changedAttributes.tabindex || changedAttributes.draggable || changedAttributes.tag ||
+		hideMissingLinks !== this.hideMissingLinks) {
```
pmario, line 213: I'm not sure about line 210 and 213. ... If the `tv-show-missing-links` variable changes, the "refresh" from eg: the vars or set-widget will be run. Right? ... So the link-widget should have been refreshed, with a new this.hideMissingLinks variable. So I think 210 and 213 should be not necessary. 

I didn't run tests, so it's only a guess

Jermolene: Hi @BurningTreeC that is correct: we don't need to worry about variable updates because a variable update will always refresh the widget tree from that point.

BurningTreeC: Ok thanks @pmario @Jermolene, I've changed it

PR #6015, opened 2021-09-01
Make image-picker in theme tweaks not dismiss popup ...

BurningTreeC:
when clicking the system-images checkbox

This PR adds `tc-popup-keep` to the popup created by the image chooser for the page background image
It dismisses the popup when choosing an image
Like this, the popup doesn't close when we click the "Include system tiddlers" checkbox


pmario:
I think that's an advantage. So +1 from me.

PR #6016, opened 2021-09-02
Better detection of scroll container and use of it in pop.js storyview

BurningTreeC:
This PR adds a better detection-method for the `scroll container` of a target Element.
It uses a solution proposed at https://stackoverflow.com/questions/35939886/find-first-scrollable-parent which is a pure JS port of the jQuery UI scrollParent method

In my tests this works more reliable

This PR also uses `$tw.utils.getScrollContainer(targetElement)` in the `pop storyview` to prevent overflowing of the scroll container in x-direction when the insert animation is done

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -66,13 +66,23 @@ exports.toggleClass = function(el,className,status) {
 
 /*
 Get the first parent element that has scrollbars or use the body as fallback.
+From https://stackoverflow.com/questions/35939886/find-first-scrollable-parent
 */
 exports.getScrollContainer = function(el) {
 	var doc = el.ownerDocument;
-	while(el.parentNode) {
-		el = el.parentNode;
-		if(el.scrollTop) {
-			return el;
+	var style = getComputedStyle(el);
+	var excludeStaticParent = style.position === "absolute";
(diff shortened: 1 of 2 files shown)


core/modules/utils/dom/dom.js
```diff
@@ -66,13 +66,23 @@ exports.toggleClass = function(el,className,status) {
...
-		el = el.parentNode;
-		if(el.scrollTop) {
-			return el;
+	var style = getComputedStyle(el);
+	var excludeStaticParent = style.position === "absolute";
+	var overflowRegex = includeHidden ? /(auto|scroll|hidden)/ : /(auto|scroll)/;
+	if(style.position === "fixed") {
+		return doc.body;
+	}
+	for(var parent=el;(parent = parent.parentElement);) {
```
pmario, line 79: For better readability IMO the line should look like this. No extra braces:
`for(var parent=el; parent=parent.parentElement;) {`

BurningTreeC: thanks @pmario, I'm about to fix this

pmario: Please also have a look at the whitespace. There is a space after `;` but no space before and after `=` 
See core for loops: https://github.com/Jermolene/TiddlyWiki5/blob/124b49456a4dc2b75ecab17e9be5180464e4d9b1/boot/boot.js#L83  .. They should look like this

PR #6019, opened 2021-09-05
Pagescroller: use scrollContainer for scrolling. + Scrollposition store/restore

BurningTreeC:
render.js and windows.js: store and restore scroll positions of all scrollable elements after a page refresh cycle

This PR does the following:

- In the pagescroller `scrollIntoView` method it detects the `scrollContainer` of the target element using `$tw.utils.getScrollContainer(element)` and scrolls that `scrollContainer` instead of using `window.scrollTo`. This approach is more flexible
- In `render.js` and `windows.js` it detects all scrollable elements before a refresh cycle and gets their scroll positions
- After a refresh, all scroll positions are restored

### DIFF
--- a/core/modules/startup/render.js
+++ b/core/modules/startup/render.js
@@ -70,11 +70,19 @@ exports.startup = function() {
 	var deferredChanges = Object.create(null),
 		timerId;
 	function refresh() {
+		var scrollableElements = $tw.utils.getScrollableElements(document);
+		var scrollPositions = [];
+		for(var i=0; i<scrollableElements.length; i++) {
+			scrollPositions.push($tw.utils.getScrollPosition(scrollableElements[i]));
+		}
 		// Process the refresh
 		$tw.hooks.invokeHook("th-page-refreshing");
 		$tw.pageWidgetNode.refresh(deferredChanges);
 		deferredChanges = Object.create(null);
 		$tw.hooks.invokeHook("th-page-refreshed");
+		for(var k=0; k<scrollableElements.length; k++) {
+			$tw.utils.setScrollPosition(scrollableElements[k],scrollPositions[k]);
+		}
 	}
 	// Add the change event handler
 	$tw.wiki.addEventListener("change",$tw.perf.report("mainRefresh",function(changes) {
--- a/core/modules/startup/windows.js
+++ b/core/modules/startup/windows.js
@@ -73,10 +73,18 @@ exports.startup = function() {
 		widgetNode.render(srcDocument.body,srcDocument.body.firstChild);
 		// Function to handle refreshes
 		refreshHandler = function(changes) {
+			var scrollableElements = $tw.utils.getScrollableElements(document);
+			var scrollPositions = [];
+			for(var i=0; i<scrollableElements.length; i++) {
+				scrollPositions.push($tw.utils.getScrollPosition(scrollableElements[i]));
+			}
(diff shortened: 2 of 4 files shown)


pmario:
Could you add a wiki with a testcase. From the source code changes it's relatively hard to create a working example and be sure I see the right code paths.

BurningTreeC:
@pmario , here's a demo wiki that demonstrates a case where the scroll position gets reset after a page refresh
Just open the file and go down to the last tiddler and save it. The scroll position will stay at the bottom even if it's a Macro tiddler
Do the same thing on https://tiddlywiki.com/prerelease and the scroll position will jump to the top of the page

[preserve-scroll-positions.zip](https://github.com/Jermolene/TiddlyWiki5/files/7111825/preserve-scroll-positions.zip)

BurningTreeC:
@pmario, for your convenience I've built the same wiki with the current prerelease, without those changes

[no-preserve-scroll-positions.zip](https://github.com/Jermolene/TiddlyWiki5/files/7111833/no-preserve-scroll-positions.zip)

BurningTreeC:
Also note, that scrolling works like it did before
The advantage now is, that we can also scroll elements where not the window is the scroll container (like in @Jermolene 's "Improved Page Layout" for example)

pmario:
Thanks for the example wikis. I do like the new behaviour

BurningTreeC:
You're welcome @pmario 
Maybe you have an explanation for this behavior:

- In my example wiki, I scroll the sidebar down
- Then I save the macro tiddler
- The sidebars scroll position gets restored (but it's not visible) because:
- After that the sidebar somehow gets replaced and appears to be scrolled to the top

I've put some console logs in my tests and I can detect the first restoration of the sidebar-scroll-position
Then I see something happening with the sidebar which I don't know what it is
Then another restoration of the scroll positions happens - but then the sidebars scroll position is already `(0,0)`

pmario:
I did some tests with the preserve-scroll-position.html 

- set animation duration to 4000
- edit tiddler
- cancel edit
- edit tiddler
- save 

does behave different. ... Did the same thing with 5.1.23 ... The animations behave different to your version

BurningTreeC:
I'll probably split this PR into two, one for the pagescroller and one for the scroll-preservation

BurningTreeC:
> I did some tests with the preserve-scroll-position.html
> 
> * set animation duration to 4000
> * edit tiddler
> * cancel edit
> * edit tiddler
> * save
> 
> does behave different. ... Did the same thing with 5.1.23 ... The animations behave different to your version.
> 
> May be there are some side effects.

@pmario , did you test this with the macro tiddler or with the other tiddlers also?
The other tiddlers should animate normally. The macro tiddler doesn't animate

pmario:
>@pmario , did you test this with the macro tiddler or with the other tiddlers also?
The other tiddlers should animate normally. The macro tiddler doesn't animate

I don't know anymore. ... BUT after reading you post above I did the following. 

 - Open the wiki 
 - Open Control Panel
 - close Control Panel 

There is a strange behaviour. The wiki I did use is attached in the ZIP here. It contains 4000ms animation delay. 
I'm using Win10 and FF 91.0.2

[preserve-scroll-positions.zip](https://github.com/Jermolene/TiddlyWiki5/files/7114653/preserve-scroll-positions.zip)

PR #6021, opened 2021-09-05
Fix slide-open and slide-close animations

BurningTreeC:
This PR tries a different approach than #5769 for fixing the slide-open and slide-close animations

The problem is (was) that the `setTimeout` call wasn't cleared when a new animation started while the other animation was still ongoing


core/modules/utils/dom/animations/slide.js
```diff
@@ -37,6 +37,7 @@ function slideOpen(domNode,options) {
 			options.callback();
 		}
 	},duration);
+	$tw.anim.animationId = id;
```
Jermolene, line 40: This is a red flag – `$tw.anim` is a shared instance that can be reused anywhere in the core, so we can't reach in and write to a property without interfering with other clients.

PR #6022, opened 2021-09-05
Fix the EditTabIndex value applied to DomNodes which was '1\n'

BurningTreeC:
This PR fixes the EditTabIndex applied to the input domNodes that use `$:/config/EditTabIndex`
Til now that tabindex was `1\n' even if that didn't cause problems


Jermolene:
Thanks @BurningTreeC

Issue #6017, opened 2021-09-02
[BUG] Node.js mode Does not Warn in UI of Sync errors


berney:
**Describe the bug**

I was using TiddlyWiki 5.1.23 in Node.js mode, using the FileSystem plugin to save tiddlers as individual files. I had TW5 running in a docker container with the volume bind mounted in so that the files would be saved to my host system, and persist after the container has been removed.

I found that if TiddlyWiki 5 encounters and error saving files to disk it does not notify the user in the browser. I was lucky that I noticed the files missing on disk. I nearly lost days of edits.

I hit a weird issue on Windows > WSL2 > Docker > TiddlyWiki5 where for some reason the bind mounted volume broke. In the docker-compose logs I could see this error repeatedly:

```
tiddlywiki_1  |  syncer-server-filesystem: Dispatching 'save' task: $:/DefaultTiddlers
tiddlywiki_1  | Sync error while processing save of '$:/DefaultTiddlers': Error: ENOENT: no such file or directory, open '/tiddlywiki/tiddlers/$__DefaultTiddlers.tid'
tiddlywiki_1  |  syncer-server-filesystem: Dispatching 'save' task: $:/DefaultTiddlers
tiddlywiki_1  | Sync error while processing save of '$:/DefaultTiddlers': Error: ENOENT: no such file or directory, open '/tiddlywiki/tiddlers/$__DefaultTiddlers.tid'
tiddlywiki_1  |  syncer-server-filesystem: Dispatching 'save' task: $:/DefaultTiddlers
tiddlywiki_1  | Sync error while processing save of '$:/DefaultTiddlers': Error: ENOENT: no such file or directory, open '/tiddlywiki/tiddlers/$__DefaultTiddlers.tid'
^CGracefully stopping... (press Ctrl+C again to force)
```

Despite the container logs showing the sync errors, in the browser the sync icon appeared like everything was happy, no errors, no warnings, and nothing in the console logs. It looked normal. All my Tiddlers were there, and everything looked saved.

If I had stopped these containers I would have lost all my work.

It would be nice if the sync error whilst processing the saves of tiddlers  could bubble back-up to the UI to indicate that something is wrong so that the user can investigate, or export all their tiddlers, before they suffer data loss.

**To Reproduce**

**Expected behavior**

[...]

Issue #5952, opened 2021-08-14
tiddlywiki by node.js can't write tiddlers to server


turtlekey:
Tiddlywiki5 is awesome👍👍👍, and I have used it for a long time🌹. But today is terrible, because I found some of my tiddlers have lost after rebooting tiddlywiki, really lost💔. Yes, now I have figured out that the reason is that some tiddlers are only in server's cache rather than write to the server, to the disk permanently, to the `tiddlers/` directory, and eventually they lost after rebooting.  Also what I have found is below:
  - I have lost some tiddlers, not all, some older tiddlers are in the directory `tiddlers/`.
  - I try deleting some tiddlers in the directory `tiddlers/`, and then writing tiddler from browser page, eventually the tiddler is writen to the `tiddlers/` directory successfully. So maybe there is size limit for the directory `tiddlers/`?
  - the available size of the server disk is enough (30G).
  - the authority mode of the directory `tiddlers/` is `775`.
  - the log file has 8.5 million rows, most of them are the error: `syncer-server-filesystem: Dispatching 'save' task: ××, Sync error while processing save of '××': Error: EACCES: permission denied, open '/home/myhome/turtlewiki/tiddlers/××.md`, it can be seen that the tiddlywiki tries to write the tiddler to the directory `tiddlers/` all the time, but fail all the time🤷‍♀️.

**TiddlyWiki Configuration (please complete the following information):**
 - Version:  v5.1.22
 - Saving mechanism: Node.js
 - Plugins installed: Highlight, Markdown, Sync changes to TW5 server or TiddlyWeb server, Synchronise changes to the local filesystem, TiddlyWiki5 core
 
**Desktop (please complete the following information):**
 - OS: centos7(server), window10(client)
 - Browser: chrome
 - Version: 92.0.4515.131

Please help me to resolve this problem.
I am looking forward your reply👀.
Thanks very much❤❤❤!

linonetwo:
Yes, this guy lost half a year of tiddlers. He is using the official nodejs setup.

One of the TiddlyGit users also reports that Windows sometimes prevent write into the filesystem, and TW just silently writes into the memory cache, while Typora will still somehow rise a Windows prompt to ask for permission to write to fs. https://github.com/tiddly-gittly/TiddlyGit-Desktop/issues/96

So force creates the file, or inform the user that fs is unavailable, is very important for nodejs setup, I think.

saqimtiaz:
[...]

PR #6026, opened 2021-09-08
Fix filesystem bugs

joshuafontany:
Fail more gracefully on filesystem errors (do not interrupt the server or send it into a retry loop). Instead, log the errors to a log tidder, and show the unseen errors in a tiddler tagged `$:/tags/Alert`. Is this a good method? The user must wait until the server is polled, and the tiddlers storing the errors must not be in the system namespace.

This is to address the recently filed filesystem bugs (to be linked here: https://github.com/Jermolene/TiddlyWiki5/issues/6017, https://github.com/Jermolene/TiddlyWiki5/issues/5952, https://github.com/Jermolene/TiddlyWiki5/issues/4742


core/modules/utils/filesystem.js
```diff
@@ -482,8 +482,10 @@ exports.deleteTiddlerFile = function(fileInfo,callback) {
 	//Only attempt to delete files that exist on disk
 	if(!fileInfo.filepath || !fs.existsSync(fileInfo.filepath)) {
 		//For some reason, the tiddler is only in memory or we can't modify the file at this path
-		$tw.syncer.displayError("Server deleteTiddlerFile task failed for filepath: "+fileInfo.filepath);
-		return callback(null,fileInfo);
+		let err = new Error("Non-existant filepath: "+fileInfo.filepath)
```
Jermolene, line 485: Does the wording "Non-existant filepath" capture all the possible reasons for an error here? For example, it might be a misleading message for access control errors.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -29,6 +29,70 @@ FileSystemAdaptor.prototype.name = "filesystem";
 
 FileSystemAdaptor.prototype.supportsLazyLoading = false;
 
+/*
+Show a filesystem errorlog alert
+*/
+FileSystemAdaptor.prototype.displayError = function(msg,err) {
```
Jermolene, line 35: There's a lot going on this method, does it need to be private to the file system adaptor? Also we already have a logging feature (see the tiddlyweb adaptor), can we not use that here?

Jermolene (commented):
Thanks @joshuafontany, a few comments.

core/modules/utils/filesystem.js
pmario, line 485: We need to use `var` instead of `let`

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -29,6 +29,70 @@ FileSystemAdaptor.prototype.name = "filesystem";
...
+		logTid = this.wiki.getTiddler(logTitle),
+		alertFields = alertTid? $tw.utils.extend({},alertTid.fields): null,
+		logFields = logTid? $tw.utils.extend({},logTid.fields): null,
+		existingCount,
+		text = "<span style='float:left;margin:2pt 4pt 8pt 4pt'>{{$:/core/images/warning}}</span> " + msg + "; " + err.toString();
+	this.logger.alert(Array.prototype.join.call([
+		$tw.language.getString("Error/WhileSaving") + ":",
+		msg + ":",
+		err.toString()
+	]," "));
```
pmario, line 50: IMO if the whole thing would be in 1 line it would be more readable. Especially, since the very last indent is wrong.

pmario: IMO this is valid for the whole function... Not 100% sure, but IMO it would be worth a try, if it is more readable

pmario:
@Jermolene ... Is there a way for the user to resolve the problem, without loosing data. .. If yes we should add a link to the docs, where we explain what to do. 

Eg: "Don't panic: Follow this link: xxxx"   It may be possible to point to AdvancedSearch Filter : `[haschanged[]]` or something similar. Then export.

Jermolene:
> @Jermolene ... Is there a way for the user to resolve the problem, without loosing data. .. If yes we should add a link to the docs, where we explain what to do.
> 
> Eg: "Don't panic: Follow this link: xxxx" It may be possible to point to AdvancedSearch Filter : `[haschanged[]]` or something similar. Then export.
> 
> There may be even a 1 click option, that does exactly that, on a "special docs tiddler"

I agree that could be helpful. We should standardise tiddlers tagged "Errors" on tiddlywiki.com with short, slug-like titles that are easy to link.

joshuafontany:
@Jermolene @pmario Thanks for the feedback, guys. This was just a first pass at a notification system for filesystem errors, mostly copied from the $tw.Logger code (because that does not generate `$:/tags/Alert` tiddlers on node, it only logs to the console). I will check out how the tiddlyweb adaptor handles logging.

Jermolene:
Thanks @joshuafontany

I remain concerned that the `FileSystemAdaptor.prototype.displayError()` method is really a generic logging component, and so ideally should be implemented within the existing logger framework.

core/modules/utils/filesystem.js
```diff
@@ -480,10 +480,12 @@ Delete a file described by the fileInfo if it exits
 */
 exports.deleteTiddlerFile = function(fileInfo,callback) {
 	//Only attempt to delete files that exist on disk
-	if(!fileInfo.filepath || !fs.existsSync(fileInfo.filepath)) {
+	if(!fileInfo || !fileInfo.filepath || !fs.existsSync(fileInfo.filepath)) {
 		//For some reason, the tiddler is only in memory or we can't modify the file at this path
-		$tw.syncer.displayError("Server deleteTiddlerFile task failed for filepath: "+fileInfo.filepath);
-		return callback(null,fileInfo);
+		var err = new Error("Invalid filepath `" + (fileInfo.filepath || "undefined") + "`");
```
simonbaird, line 485: The error message seems to incorrectly assume that fileInfo is present. Maybe you need something like this:

   var err = new Error("Invalid filepath `" + (fileInfo && fileInfo.filepath ? fileInfo.filepath) : "undefined") + "`");

saqimtiaz:
**bump**

pmario:
I think the bump should go to @joshuafontany  ... IMO there is an unresolved remark from Simon, which should be addressed first.

saqimtiaz:
@pmario the bump is meant for all of us actually as this is a pretty acute issue that needs resolution. If Joshua does not have time to work on this, hopefully one of us can pick it up.

saqimtiaz:
I think we need to extend the Logger class to be able to create $:/tags/Alert tiddlers on node.js for errors that need to be propagated to the client.

joshuafontany:
> I think we need to extend the Logger class to be able to create $:/tags/Alert tiddlers on node.js for errors that need to be propagated to the client.

That goes a long way to fix this issue. The other side of that is with the standard polling increment, the user would only see that Alert after about a minute.

saqimtiaz:
@joshuafontany right which again brings up the need to be able to propagate changes from the server to the client immediately. However even an alert after a minute will be far better than what we have now.

I am thinking of an options attribute for Logger that forces it to create alerts even on node. @joshuafontany would you have the opportunity in the next weeks to take a crack at this?

joshuafontany:
@Jermolene @pmario @saqimtiaz @simonbaird 

Hi, I've made time to get back to these filesystem bugs. I think the current draft goes a very long way to address these issues.

BUT, I am running into a conflict. Right now I am listing all server alerts/logs in the `$:/server/` namespace. Unfortunately, that means that the client/server-syncing changes made in https://github.com/Jermolene/TiddlyWiki5/pull/4987 are preventing the client from seeing these tiddlers unless they refresh the page. Which destroys the wiki-state in the browser before the server can warn the user there was a one or more problems changing the server state.

I can force-polling for these errors only, avoiding the `GET /recipes/default/tiddlers.json?filter=<filter>` route that forces system tiddlers to be removed from the updated tiddlers json:
```
	if(state.wiki.getTiddlerText("$:/config/SyncSystemTiddlersFromServer") === "no") {
		filter += "+[!is[system]]";
	}
```

I unfortunately think that the implementation in https://github.com/Jermolene/TiddlyWiki5/pull/4987 may be blocking us from resolving this unless I work around it. I'm out of ideas, and would love some input.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -132,7 +114,7 @@ FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 		}
 		$tw.utils.saveTiddlerToFile(tiddler,fileInfo,function(err,fileInfo) {
 			if(err) {
-				var message = "Save file failed for [["+tiddler.fields.title+"]]";
+				var message = "Save file failed for \'"+tiddler.fields.title+"\'";
```
pmario: Why is the single quote escaped here?

joshuafontany:
@Arlen22 I have seen some PRs from you trying to work around #4987... any input here?

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -124,7 +124,7 @@ FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 					return callback(err);
 				}
 			}
-			self.logger.log("Saved \'"+fileInfo.filepath+"\'");
+			self.logger.log("Saved \'"+fileInfo.filepath+"\'");debugger;
```
pmario: I think, that's a problem.

joshuafontany: Still testing, will remove that, thanks.

pmario:
@joshuafontany .. Just stumbled upon this Draft PR. .. It has never been "promoted" to active review state

Is it still valid? ... There are some open issues with link to this PR. .. For me it seems there would be value, since the issues mentioned talk about severe data-loss problems

Jermolene:
Hi @joshuafontany I noticed you closed this, is it no longer relevant?

PR #6028, opened 2021-09-10
Fix slide animations when opening/closing while animation is ongoing

BurningTreeC:
This PR fixes the slide animations when opening or closing while another animation is ongoing

I hope that this time I got how to do this right


pmario:
Did test it with Windows 10, FF 92 ... There is still a problem. 

To reproduce: 

 - Set animation time to 4000
 - Open tiddler info area
 - click outside within 1 second   ->  Works OK. 

but

 - Set animation time to 4000
 - Open tiddler info area
 - click outside within 3 second   ->  Works OK. 
 - The area starts to close
 - Click the Open Info button again while the closing animation is active
 - Now we have the same problem as we had with 5.1.23 for step 1

PR #6034, opened 2021-09-16
removed pushtop from untagged filter

flibbles:
I'm back to implementing the LinkedList into all the filters that use pushTop. I was going to start with [untagged[]], when I had the question:

Why on earth is [untagged[]] using pushtop? Nothing in the documentation indicates that it does. No filters that are like it do (such as tag, field, prefix, etc...). It seems to me that it should just be using push, because in the RARE occurrence that someone passes a source with duplicates to [untagged[]], it shouldn't be [untagged[]]'s responsibility to remove those duplicates. It seems to me like it's more like a bug than a feature. Who would want [untagged[]] doing that?

Anyway, here are the performance specs: Filtering with 30,000 input tiddlers on Chrome.

Using pushtop: 1685 ms
Just using push: 37 ms

### DIFF
--- a/core/modules/filters/untagged.js
+++ b/core/modules/filters/untagged.js
@@ -16,20 +16,13 @@ Filter operator returning all the selected tiddlers that are untagged
 Export our filter function
 */
 exports.untagged = function(source,operator,options) {
-	var results = [];
-	if(operator.prefix === "!") {
-		source(function(tiddler,title) {
-			if(tiddler && $tw.utils.isArray(tiddler.fields.tags) && tiddler.fields.tags.length > 0) {
-				$tw.utils.pushTop(results,title);
-			}
-		});
-	} else {
-		source(function(tiddler,title) {
-			if(!tiddler || !tiddler.hasField("tags") || ($tw.utils.isArray(tiddler.fields.tags) && tiddler.fields.tags.length === 0)) {
-				$tw.utils.pushTop(results,title);
-			}
-		});
-	}
+	var results = [],
+		expected = (operator.prefix === "!");
+	source(function(tiddler,title) {
+		if((tiddler && $tw.utils.isArray(tiddler.fields.tags) && tiddler.fields.tags.length > 0) === expected) {
+			results.push(title);
+		}
+	});
 	return results;
 };
 


flibbles:
Just to explain how it looks like I cut the existing code in half. The old way used to create two separate bodies, one for the standard `[untagged[]]` and the other for `[!untagged[]]`. I merged them by comparing an incoming tiddler's tag status to a variable boolean set with `operator.prefix === '!'`, rather than to true.

I realize someone must have considered this and rejected it long ago, since so many filters do this, but I contend that my way is better.

1. It's more brief. Less code duplication.
2. It probably introduces one more conditional for each incoming title, but whatevs. It's already doing, like, 6 conditionals and 3 function calls per title. My change from pushTop to push WAY improves the performance compared to this.
3. Maybe it's slightly harder to follow, but it's actually less error prone. Look at `[untagged[]]`'s old implementation and you'll see that whoever made this accidentally did not make the IF blocks in the corresponding bodies complements of each other. The bottom one had an extra conditional `!tiddler.hasField('tags')`. In this case, it didn't matter, since `isArray` already confirms the same thing, but this underlines my point that someone might not create a filter's true complement. My way guarantees that `[untagged[]]` and `[!untagged[]]` are complements of each other.

flibbles:
I'm not thrilled with my `expected` variable name though, so if someone has a better idea for that, let me know.

Jermolene:
Thanks @flibbles 

> Why on earth is `[untagged[]]` using pushtop?

I think because very early on in TW5's development I thought of filters as working with sets rather than lists, and so defaulted to using `pushTop`.

I think strict backwards compatibility requires us to retain the current behaviour, or are you suggesting that it would be a reasonable exception?

> I realize someone must have considered this and rejected it long ago, since so many filters do this, but I contend that my way is better.

I think that a few filters were written like that originally and then the rest were copied and pasted. I've no problem with the pattern you're using (it is used in some places in the core)

> I'm not thrilled with my `expected` variable name though, so if someone has a better idea for that, let me know.

It seems reasonable. I can't easily find the other places we use the same technique.

flibbles:
I figured this would be a reasonable case where backward compatibility wouldn't be a concern. If you have the following input.

['A', 'B', 'B', 'C', 'A']

or `A B =B C =A` as you'd input it as a filter string

And you ask `[!tag[myTag]]`, you might get back ['A', 'C', 'A']
But if you ask `[untagged[]]`, you might get back ['C', 'A']

I figured it wasn't untagged's responsibility to remove duplicates because its documentation doesn't indicate so, and you wouldn't assume as much given what it's supposed to be doing.

...But in writing this, I also noticed that `[tag[myTag]]` would return just ['B'], so `[tag[]]` and `[!tag[]]` aren't really true compliments either. It's very clear that TiddlyWiki filters were never originally intended to cope with duplicates in the filter.

Jermolene:
> It's very clear that TiddlyWiki filters were never originally intended to cope with duplicates in the filter.

It's definitely one of the areas I'd want to revisit for a future version that drops strict backwards compatibility.

PR #6035, opened 2021-09-16
filter ops don't LinkedList to array anymore

flibbles:
While investigating how to make filters lazy evaluators, I realized that filters don't have to return arrays. They can return methods that accept callbacks instead. So here's a little improvement to the way existing filters use LinkedList which removes the cloning of the LinkedList into an array before allowing the next filter step to use it.

Should be a little more time efficient, and a good deal more memory efficient.

Also, if people like my lazy filters, then this change makes transitioning easier.

### DIFF
--- a/core/modules/utils/linked-list.js
+++ b/core/modules/utils/linked-list.js
@@ -95,6 +95,15 @@ LinkedList.prototype.toArray = function() {
 	return output;
 };
 
+LinkedList.prototype.tiddlerIterator = function(wiki) {
+	var self = this;
+	return function(callback) {
+		self.each(function(title) {
+			callback(wiki.getTiddler(title),title);
+		});
+	};
+};
+
 function _removeOne(list,value) {
 	var prevEntry = list.prev[value],
 		nextEntry = list.next[value],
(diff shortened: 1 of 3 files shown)


core/modules/utils/linked-list.js
```diff
@@ -95,6 +95,15 @@ LinkedList.prototype.toArray = function() {
 	return output;
 };
 
+LinkedList.prototype.tiddlerIterator = function(wiki) {
```
Jermolene, line 98: I wonder if this method should be `makeTiddlerIterator` to match the corresponding wiki store method?

Jermolene (commented):
Thank you @flibbles terrific cunning, just one small point.

core/modules/utils/linked-list.js
flibbles: Fair enough. It does return a method which behaves identically (and must behave identically) to the $tw.wiki.makeTiddlerIterator. I'd considered making the wikimethod handle it instead, but I think it better fits the grain if LinkedList has its own, since it also has its own "pushTop" method.

Change is made, and PR is good to go.

Jermolene:
Thanks @flibbles

PR #6036, opened 2021-09-16
wiki.eachShadow returns overridden sources

### LINKED ISSUE #6033: [BUG] [all[shadows]] can cause filters to fetch shadowsource material
Here's what you do

1. Override the body of the shadow tiddler "$:/Acknowledgements" to be anything, let's say "OVERRIDDEN".
2. Open the "Advanced search" panel and switch to filters.

The following filters result in these results:
``[all[shadows]get[text]first[]]`` --> "TiddlyWiki incorporates code from these..."
``[all[shadows]first[]get[text]]`` --> "OVERRIDDEN"
``[all[shadows+tiddlers]get[text]first[]]`` --> "OVERRIDDEN"
``[all[shadows+tiddlers]first[]get[text]]`` --> "OVERRIDDEN"

What's happening is the `[all[shadows]]` filter is shortcutting to $tw.wiki.eachShadow, which calls the callback method with `callback(shadowTiddlers[title], title)`, whereas `$tw.wiki.each`, `$tw.wiki.eachShadowPlusTiddler`, nor `$tw.wiki.eachTiddlerPlusShadow` do that. The others all prefer the overridden tiddler.

However, if used in a filter, that shadow tiddler reference will get dropped if the filter passes all's contents through to anything else first. That's because every filter operator returns only the `title` as a string, and thus the next operator in the line gets  the `title, $tw.wiki.getTiddler(title)`

No. I did not find this organically. I'm looking into whether TiddlyWiki could benefit from having lazy evaluation with its filters and core tiddler iterating methods, but this inconsistency throws a wrench in that idea.

Was this inconsistency intended? Doesn't it seem weird that one of the four core tiddler iterator methods prefers the underlying shadow tiddler?

flibbles:
It was returning the underlying shadows before, even if overridden, which was causing some bugs.

Fixes #6033

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1212,8 +1212,12 @@ $tw.Wiki = function(options) {
 			index,titlesLength,title;
 		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
 			title = titles[index];
-			var shadowInfo = shadowTiddlers[title];
-			callback(shadowInfo.tiddler,title);
+			if(tiddlers[title]) {
+				callback(tiddlers[title],title);
+			} else {
+				var shadowInfo = shadowTiddlers[title];
+				callback(shadowInfo.tiddler,title);
+			}
 		}
 	};
 


Jermolene:
Great stuff, thank you @flibbles. I think this may explain some bugs I've experienced in the past.

PR #6037, opened 2021-09-17
Let's make TiddlyWiki lazy!!!

flibbles:
I know this is the feature no one asked for, but hear me out. I was blown away when I first got into TiddlyWiki development when I found out that its filters, which would lend themselves so wonderfully to lazy evaluation, actually accumulate full arrays of intermediate results between every single operation. I wanted to see if it was even possible to make it lazy now with so much backward compatibility necessary, and... actually it was pretty easy.

Check out my implementation. It's not meant for merging, but more for people to see. If people don't want to go this direction, I understand, but it's super nifty, you guys!

## tiddler iterating wikimethods
The first major change was taking wiki's methods `each`, `eachShadow`, `eachShadowAndTiddlers`, and `eachTiddlerAndShadows` and making them lazy while maintaining their existing use. If you call them and supply a callback method, they call that callback method like usual. If you don't supply a callback, they return an iterator. Boom. Done.

The super cool way I have it implemented right now makes these iterators forward compatible with ES6's `for...of` syntax, so you can do this...
```javascript
for(var title of $tw.wiki.each()) {
   // do stuff. Or don't. Finish, or break. Up to you.
}
```
Of course, TiddlyWiki doesn't allow `for...of`, but plugin developers don't have to concern themselves with supporting those people using those CRT screen Dell computers at their church to design their websites. We wouldn't _have_ to keep support for `for...of`, and if we don't, some of my code simplifies down.

To keep support, right now, $tw.wiki.each() et al are returning an iterator that returns `{value: title, done: false}` if there's a value and `{value: undefined, done: true}` for End-Of-Stream. BUT since our tiddler titles are always defined, and since filters never allow non-string values, we could have the simpler `title` on value, and `undefined` for End-Of-Stream.

## Filters

Filters don't have to change at all. This upgrade is fully backward compatible. However, we obviously don't get much benefit from this if I don't refactor the existing filters. Currently, I've only implemented `prefix` and `getindex`, because they both make good use cases.

There are several types of filters that won't get much benefit from this, and that's the ones that use pushTop. (I freakin hate that method). But many other filters do, especially ones like `[first[]]` and such.

Anyway, let me know what you guys think. Anyone interested in this? Should it maintain forward-compatibility or go with the slightly simpler form where it just passes around iterator methods? Check out the new code too. I littered it with heavy comments explaining stuff (that I'd take out before merge).

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1194,12 +1194,20 @@ $tw.Wiki = function(options) {
 	// Iterate through all tiddler titles
 	this.each = function(callback) {
 		var titles = getTiddlerTitles(),
-			index,titlesLength,title;
-		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
+			index = 0,
+			titlesLength,
+			title;
+		if(callback === undefined) {
+			return new $tw.utils.Iterator(function() {
+				return titles[index++];
+			});
+		}
+		for(titlesLength=titles.length; index < titlesLength; index++) {
 			title = titles[index];
 			callback(tiddlers[title],title);
 		}
 	};
+	this.each.iterable = true;
 
 	// Get an array of all shadow tiddler titles
 	this.allShadowTitles = function() {
@@ -1209,8 +1217,15 @@ $tw.Wiki = function(options) {
 	// Iterate through all shadow tiddler titles
 	this.eachShadow = function(callback) {
 		var titles = getShadowTiddlerTitles(),
-			index,titlesLength,title;
-		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
+			index = 0,
+			title,
+			titlesLength;
+		if(callback === undefined) {
+			return new $tw.utils.Iterator(function() {
+				return titles[index++];
+			});
+		}
+		for(titlesLength = titles.length; index < titlesLength; index++) {
 			title = titles[index];
 			if(tiddlers[title]) {
 				callback(tiddlers[title],title);
@@ -1220,31 +1235,75 @@ $tw.Wiki = function(options) {
 			}
 		}
 	};
+	this.eachShadow.iterable = true;
 
 	// Iterate through all tiddlers and then the shadows
 	this.eachTiddlerPlusShadows = function(callback) {
-		var index,titlesLength,title,
-			titles = getTiddlerTitles();
-		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
+		var index = 0,
+			titlesLength,title,
+			titles = getTiddlerTitles(),
+			shadows = getShadowTiddlerTitles();
+		if(callback === undefined) {
+			return new $tw.utils.Iterator(function() {
+				if (titles) {
+					if (index < titlesLength) {
+						return titles[index++];
+					}
+					titles = null;
+					index = 0;
+					titlesLength = shadows.length;
+				}
+				while (index < titlesLength) {
+					if(!tiddlers[title]) {
+						return shadows[index++];
+					}
+					index++;
+				}
+				return undefined;
+			});
+		}
+		for(titlesLength = titles.length; index < titlesLength; index++) {
 			title = titles[index];
 			callback(tiddlers[title],title);
 		}
-		titles = getShadowTiddlerTitles();
-		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
-			title = titles[index];
+		for(index = 0, titlesLength = shadows.length; index < titlesLength; index++) {
+			title = shadows[index];
 			if(!tiddlers[title]) {
 				var shadowInfo = shadowTiddlers[title];
 				callback(shadowInfo.tiddler,title);
 			}
 		}
 	};
+	this.eachTiddlerPlusShadows.iterable = true;
 
 	// Iterate through all the shadows and then the tiddlers
 	this.eachShadowPlusTiddlers = function(callback) {
-		var index,titlesLength,title,
-			titles = getShadowTiddlerTitles();
-		for(index = 0, titlesLength = titles.length; index < titlesLength; index++) {
-			title = titles[index];
+		var index = 0,
+			title,
+			shadows = getShadowTiddlerTitles(),
+			titles = getTiddlerTitles(),
+			titlesLength = shadows.length;
+		if(callback === undefined) {
+			return new $tw.utils.Iterator(function() {
+				if (shadows) {
+					if (index < titlesLength) {
+						return shadows[index++];
+					}
+					shadows = null;
+					index = 0;
+					titlesLength = titles.length;
+				}
+				while (index < titlesLength) {
+					if(!shadowTiddlers[title]) {
+						return titles[index++];
+					}
+					index++;
+				}
+				return undefined;
+			});
+		}
+		for(; index < titlesLength; index++) {
+			title = shadows[index];
 			if(tiddlers[title]) {
 				callback(tiddlers[title],title);
 			} else {
@@ -1260,6 +1319,7 @@ $tw.Wiki = function(options) {
 			}
 		}
 	};
+	this.eachShadowPlusTiddlers.iterable = true;
 
 	// Test for the existence of a tiddler (excludes shadow tiddlers)
 	this.tiddlerExists = function(title) {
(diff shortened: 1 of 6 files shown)


pmario (commented):
Hi,  I do like the concept. 

I personally would like to have it "forward compatible". ... But I do question myself, where does it lead to. ... Or better where "should" it lead us?

pmario:
Do you want to go into a direction similar to database "indexes, ranges and cursors"?

flibbles:
Hey @pmario,

I'm not sure what you mean by indexes, ranges and cursors. It's been a long time since I've worked databases.

flibbles:
By the way, while for...of is nifty, if we didn't go with it, then iterating the wiki's tiddlers would look something like this:

```javascript
for(var iter = $tw.wiki.each(), title = iter(); title !== undefined; title = iter()) {
    // Do stuff
}

// Or my personal preference
var iter = $tw.wiki.each(), title;
while((title = iter()) !== undefined) {
    // Do stuff
}
```

If we _kept_ for...of support, then the core (since it can't use for...of) would then look like this:

```javascript
for(var iter = $tw.wiki.each(), pair = iter.next(); !pair.done; pair = iter.next()) {
    var title = pair.value;
    // Do stuff
}

// Or my personal preference
var iter = $tw.wiki.each(), pair, title;
while((pair = iter.next()).done == false) {
    title = pair.value;
    // Do stuff
}
```

It's a little uglier, but I can't imagine we'd go through tiddlywiki core and switch over to using iterators rather than core anywhere except for in filters, since the callback will usually be just as effective outside the context of filters.

pmario:
>I'm not sure what you mean by indexes, ranges and cursors. It's been a long time since I've worked databases.

I was thinking about a filter-based "paging mechanism" without recalculating the list. Eg: Go to the right sidebar More : All tabs. .. There is a long list, which can't be shown on the screen. It's wasting CPU cycles and slows down the UI.

With `{{{[!is[system]limit[10]]}}}` we can create a list that only contains 10 elements (1-10). .. BUT what if I want to see the next 10 elements. Or previous 10 elements. We have to recalculate the list again and again ... And the filter syntax will be complicated. 

With a range-iterator imo it would be simple and effective. `.next(20)` will show the next 20 elements (11-30). `.next(-10)` will show element (1-10) again. `.next(5)` will show 5 elements (11-15) and so on. 

So the "iterator" calculates the index once and caches it
.next() request the range
.next(cursor) defines what to show and in which direction to navigate. 

I think it may be possible for filters to maintain state, which isn't possible atm. 

See: [range iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators) the fibonacci example.

Jermolene:
Thanks @flibbles this is a good idea, and we should consider it for post-v5.2.0. The changes are surprisingly clean, and I definitely like the compatibility with ES6 (which I'm already using a fair amount in non-TW projects).

flibbles:
@pmario, It wouldn't be too hard to create a seeking iterator. wiki.filterTiddlers would look like,
```javascript
exports.filterTiddlers = function(filterString,widget,source) {
    var fn = this.compileFilter(filterString);
    var iterator = fn.call(this,source,widget));
    return $tw.utils.slurpIterator(iterator); // or something like that
};
```
Meanwhile, things which require paging could instead just get the iterator, and then wrap that inside a seeking iterator which it would use.

Hmm... Well, we couldn't do exactly that, because too many plugins probably expect fn.call(...) to return an array, but you get the idea.

__Using__ that seeking iterator would be the hard part, and I don't even see how off the top of my head.

@Jermolene, Great! I'll hold off on developing this further until V1.2.0 is out. Converting over the other filters will be quick and simple. (Filterrunprefixes though, which currently work by manipulating an array passed to them, will be a trickier matter...)

flibbles:
Is this something people actually want to do? I mean, it's a *neat* idea, but more and more I'm seeing how TiddlyWiki filters have been developed (and continue to be developed) in a way that makes lazy evaluation pointless.

* Countless operators pushtop, which is about as lazy-unfriendly as it gets.
* Operators that should be good candidates, such ones that operate on inputs independently, will sort their output.
* The filter run prefixes work by taking an entire array, and then *altering that array in place* as a means of altering input. To make these lazy, you'd pretty much need to deprecate the module-type and create a new one (as far as I can tell).

Unless people are hungry for lazy filters, I might go ahead and close this so it'll stop irking me every time I see someone add a new filter operator which arbitrarily, and quite unnecessarily, sorts its output.

pmario:
> I see someone add a new filter operator which arbitrarily, and quite unnecessarily, sorts its output.

Which filters do you refer to?

flibbles:
Yesterday, I got a little bent out of shape over `moduleproperty[]`. It **should** have been a perfect little solution for a problem of mine, except it sorts. *Why does it sort?* Who looked this filter, which generates one-to-one output for each input, and thought, "this should be O(n log n). Ooh! And maybe it should completely throw away whatever order the input was in and make assumptions about what order it should be." And now it's impossible to make lazy.

I didn't "git blame" this, so I don't know who I'm disparaging right now. Yesterday was a bad day for me, so this little thing kinda set me off, but this frustration did make me consider that lazy filters may not be feasible. Tiddlywiki has an ever-growing pool of backward-compatibility requirements that make laziness difficult to make useful.

pmario:
>I got a little bent out of shape over moduleproperty[]. It should have been a perfect little solution for a problem of mine, except it sorts. Why does it sort? 

As far as I did understand it, from a discussion with Jeremy a long time ago. .. 

All the filters that have a hardcoded `.sort()` in it do gather their information from objects. The problem with default js-objects is, that the sort-order of their properties isn't explicitly specified in the ES5 JS specs. 

So it can depend on the browser vendor, how they implemented it. ... eg: If you call: `[[macro]modules[]moduleproperty[name]nth[1]]` we can't 100% guarantee, that it will always be the same thing. That's why some filters always sort. 

See: https://stackoverflow.com/a/5525820 ff
and about the property order: https://stackoverflow.com/a/38218582 ... it's complicated ... 

There are new object iterators, but we don't use them atm ;)

@Jermolene ... correct me if I'm wrong.

flibbles:
That would only matter if it were pulling all of the properties from a module object, but what this filter does is pull the same single property from a list of different modules, so the order would have been dependent on the order of the inputted modules. It'd be unrelated to the order of the properties in those modules' `export`.

Jermolene:
Hi @flibbles Modules don't have an ordering, or rather we treat the module ordering as an implementation detail that we hide. Is there a case where ordering of modules would be helpful?

flibbles:
Hey @Jermolene. It was a little beside the point, but I just needed to list some module properties in alphabetical order of the modules' names, **not** in alphabetical order of the values I was extracting from those modules. I got around it by writing my own allfilteroperator.

I was bringing this up because I've slowly been accepting that as neat as lazy filtering would be, I might be pushing it simply because it *is* neat, and not necessarily because Tiddlywiki needs it, and that there's a lot of filters and filterrunprefixes which have been designed in such a way that lazy evaluation would not benefit them unless they made backward-incompatible changes.

I'm still happy to pursue this if you all think this is a good idea.

Jermolene:
Hi @flibbles 

> I'm still happy to pursue this if you all think this is a good idea.

It's an interesting line of research that tackles an important problem. Perhaps it's only a temporary dead end; it may yet be that some new insight will kick start progress again.

But in terms of whether it's a good idea to continue, I guess it's competing with other areas that might be ripe for optimisation, and there's probably quite a few of those. Some experiments I've been doing with large 500MB wikis suggest that garbage collection is a big problem for the core.

pmario:
@flibbles ... Can we make this PR a draft, so it is "visible" as an experiment. (So it does not catch my attention everytime I visit open ummerged PRs)

flibbles:
@pmario, I think I'd be happy to, although I don't know what that is. Is that like a separate branch?

Might it be easier to close this as unmerged? Unless there are plans to put something like this into V5.4.x, I think we can safely assume there will never be plans to put something like this into TiddlyWiki.

pmario:
You should be able to convert your PRs into drafts, which means it is not finished yet. Once it is finished, the review process can be started again **by you**

See screenshot

image

flibbles:
Okay. I did it, though I still put forward the opinion that this feature will never get added. Filters were not designed with lazy in mind. For instance, any operator that is dominant appending cannot be lazy _at all_.

Even in the 4 years since I proposed this, additional features have been added which make it increasingly difficult to reap benefits from being lazy.

pmario:
OK. Then you can close it too.

PR #6046, opened 2021-09-18
katex-plugin: fix #6041 add automatic numbering reset to KaTex style-sheet

### LINKED ISSUE #6041: add automatic numbering reset to KaTex style-sheet
I did some experiments and it seems, a stylesheet (a tiddler tagged with `$:/tags/Stylesheet`) as below works

```
.tc-tiddler-body  { counter-reset: katexEqnNo; }
```

do you recommend this solution?

_Originally posted by @kookma in https://github.com/Jermolene/TiddlyWiki5/discussions/6039#discussioncomment-1345535_

pmario:
This PR fixes #6041 add automatic numbering reset to KaTex style-sheet, that has been discussed at: https://github.com/Jermolene/TiddlyWiki5/discussions/6039

Using `.tc-tiddler-frame` will also include elements that will be added to the ViewTemplate in the future.

### DIFF
--- a/plugins/tiddlywiki/katex/styles.tid
+++ b/plugins/tiddlywiki/katex/styles.tid
@@ -13,7 +13,7 @@ tags: [[$:/tags/Stylesheet]]
     text-rendering: auto;
 }
 
-/* Reset Automatic Numbering on a per tiddler body basis */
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario

PR #6047, opened 2021-09-18
Allow deleteTiddler to not modify fs

linonetwo:
When writing watch-fs plugin that monitor fs change, I found it difficult to handle delete, file delete can happened by calling `deleteTiddler`, or can be delete by the user using VSCode.

So I need this new option to ask it to not touch fs.

### DIFF
--- a/plugins/tiddlywiki/filesystem/filesystemadaptor.js
+++ b/plugins/tiddlywiki/filesystem/filesystemadaptor.js
@@ -154,7 +154,7 @@ Delete a tiddler in cache, without modifying file system.
 FileSystemAdaptor.prototype.deleteTiddlerInCache = function(title) {
 	// Only delete the tiddler if we have writable information for the file
 	if(this.boot.files[title]) {
-		delete self.boot.files[title];
+		delete this.boot.files[title];
 	};
 };
 
--- a/plugins/tiddlywiki/filesystem/filesystemadaptor.js
+++ b/plugins/tiddlywiki/filesystem/filesystemadaptor.js
@@ -82,12 +82,12 @@ FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 		if(err) {
 			return callback(err);
 		}
-		$tw.utils.saveTiddlerToFile(tiddler,fileInfo,function(err,fileInfo) {
+		$tw.utils.saveTiddlerToFile(tiddler,fileInfo, (err,fileInfo) => {
 			if(err) {
 				if ((err.code == "EPERM" || err.code == "EACCES") && err.syscall == "open") {
 					fileInfo = fileInfo || self.boot.files[tiddler.fields.title];
 					fileInfo.writeError = true;
-					self.boot.files[tiddler.fields.title] = fileInfo;
+					this.boot.files[tiddler.fields.title] = fileInfo;
 					$tw.syncer.logger.log("Sync failed for \""+tiddler.fields.title+"\" and will be retried with encoded filepath",encodeURIComponent(fileInfo.filepath));
 					return callback(err);
 				} else {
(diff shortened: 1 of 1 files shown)


plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -82,20 +82,20 @@ FileSystemAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 		if(err) {
 			return callback(err);
 		}
-		$tw.utils.saveTiddlerToFile(tiddler,fileInfo,function(err,fileInfo) {
+		$tw.utils.saveTiddlerToFile(tiddler,fileInfo, (err,fileInfo) => {
```
Jermolene, line 85: I'm afraid we can't use arrow functions in the core. We stick to ES5.1 (roughly) because it's important to retain compatibility with older browsers/JS runtimes.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -122,14 +122,22 @@ FileSystemAdaptor.prototype.loadTiddler = function(title,callback) {
 };
 
 /*
-Delete a tiddler and invoke the callback with (err)
+	Delete a tiddler and invoke the callback with (err)
+	@params title title of tiddler to delete
```
Jermolene, line 126: We don't use jsdoc style comments

plugins/tiddlywiki/filesystem/filesystemadaptor.js
```diff
@@ -122,14 +122,22 @@ FileSystemAdaptor.prototype.loadTiddler = function(title,callback) {
...
+	@params callback callback to execute after tiddler deletion
+	@params options `{ cacheOnly: boolean }` cacheOnly: if set cacheOnly as true, we won't touch fs, just delete file from this.boot.files
 */
 FileSystemAdaptor.prototype.deleteTiddler = function(title,callback,options) {
-	var self = this,
-		fileInfo = this.boot.files[title];
+	const fileInfo = this.boot.files[title];
 	// Only delete the tiddler if we have writable information for the file
+	var cacheOnly = options.cacheOnly || false;
+	if(cacheOnly) {
```
Jermolene, line 134: The "cacheOnly" flag isn't used by the core, so I assume it is intended for plugins? It seems like it would be simpler if it was a separate method

Jermolene (commented):
Thanks @linonetwo

plugins/tiddlywiki/filesystem/filesystemadaptor.js
linonetwo: I will keep that in mind in the following PRs.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
linonetwo: Can there be a post in Github Discussion talking about this? Currently $tw API doc is incomplete, which makes there being some obstacles in plugin development.

plugins/tiddlywiki/filesystem/filesystemadaptor.js
linonetwo: Ok, refactor complete!

linonetwo: Feel free to point out if you don't like the function name. I think code review is a good place that let people know more about owner's style.

joshuafontany:
@Jermolene @linonetwo Looking over existing file-system issues. This caught my eye. I feel that this is actually a file-system-permissions issue. I did many tests with loading tiddlers from "read-only" files, and what happens when TW Node tries to delete them as one of my "fail gracefully" tests.

At this point, there is no way for the TW Node server to set specific file permissions, so that would need to be done outside of TW code for the moment (or manually/on the command line). TW code like that leads into needing an "admin" authorization level to the server-flow, which I have started work on.

Good issue! Definitely worth investigating user-facing solutions.

linonetwo:
@Jermolene Can you reconsider this? Just add a new method to allow plugin touch file cache.

Jermolene (commented):
Looks good thanks @linonetwo

PR #6048, opened 2021-09-18
Eventcatcher: check for event properties before accessing

saqimtiaz:
Fixes a new bug in `<$eventcatcher>` for custom events where if the event does not have a `clientX` or `clientY` property, it results in an error. This was introduced after the changes in https://github.com/Jermolene/TiddlyWiki5/commit/e9613d7f1225a3fe726e1790cc46e3f6ca9ca934 

The other variables relying on these event properties were being assigned a value of `NaN` but were not causing an exception. Now, we only assign all of the variables relying on these event properties if the event has the relevant properties.


Jermolene:
Thanks @saqimtiaz

PR #6050, opened 2021-09-19
backlinks uses LinkedList

flibbles:
Backlinks. Here's another filter that benefits from using LinkedLists.

With an input of 10,000 tiddlers, This filter takes ~150 ms. With linked lists, it takes ~15 ms.


saqimtiaz:
This should be very well received with the proliferation in usage of backlinks. Thank you.

Jermolene:
Thanks @flibbles!

Issue #6042, opened 2021-09-17
Performance regression reported from v5.1.23


Jermolene:
From https://talk.tiddlywiki.org/t/tw-5-2-0-prerelease-significantly-slower-than-tw-5-1-23-instrumentation-data-included/776:

---

I’ve tested the 5.2.0-prerelease over the past few days both on mobile and on desktop.

I’ve noticed significantly degraded performance on mobile and did some performance instrumentation tests on my desktop. Results below.

I’ve primarily tested the basic TW operations such as navigating links (opening, closing tiddlers), selecting checkboxes, radio buttons, clicking buttons etc.

The overall performance is significantly slower in 5.2.0 compared to the same wiki running on 5.1.23. It appears that mainRenderer and mainRefresh see the biggest slowdowns.

Here’s some data from the two identical wikis running side by side, the figures on the left correspond to the 5.1.23 version and the ones on the right are from the 5.2.0-prerelease. In each “test run” I had all tiddlers closed beforehand and then performed identical navigation:

mainRenderer values

Opening tiddlers (variable levels of complexity)
151 ms vs 276ms
149 ms vs 230ms
242 ms vs 302ms
105 ms vs 196ms
398 ms vs 503ms

Opening Control Panel
90ms vs 190ms

Opening Advanced Search
83ms vs 195ms

Clicking on a checkbox that sets a value in a data tiddler
67ms vs 160ms

Pressing a button that sets a value in a data tiddler
175mm vs 278ms

Entering a password in an encrypted wiki, upon pressing enter:
mainRender (v5.1.23): 226.00m vs mainRender (v.5.2.0-prerelease): 350.00ms
mainRefresh (v5.1.23): 74.00ms vs mainRefresh (v.5.2.0-prerelease): 192.00ms

I’ve performed these tests on my desktop with an 8-Core Ryzen CPU clocked at 3.60 GHz and 32GB RAM. The performance on my midrange Motorola is at least 5 times slower. While desktop performance slowdowns on this system are negligible, mobile performance introduces major usability issues.

The values above are obviously system dependent and wiki dependent but the point I’m trying to make is that in relative terms, 5.2.0-prerelease is significantly slower than 5.1.23. I’ve made these tests on the same system, the same identical wiki, with the only difference being the TiddlyWiki version.

I’d like to hear your thoughts about this and also if there’s anything you could suggest. Could anyone perhaps perform similar tests on their end?

[...]

PR #6053, opened 2021-09-19
Optimise wiki.sortTiddlers()

Jermolene:
As discussed in #6042, there are some simple opportunities to make tiddler sorting more efficient.

These changes improve the startup time of a test wiki as follows:

* v5.1.23 mainRender: **5412.20ms**
* [575c233](https://github.com/Jermolene/TiddlyWiki5/commit/575c2335975cc6cc960d2061ba8fdfa726ed8618) mainRender: **11221.60ms**
* with this PR mainRender: **3862.00ms**
 
([575c233](https://github.com/Jermolene/TiddlyWiki5/commit/575c2335975cc6cc960d2061ba8fdfa726ed8618) is the last commit before reverting the problematic commit [dbd3f83](https://github.com/Jermolene/TiddlyWiki5/commit/dbd3f835bf8399ed1a3da7cc322ec9b6ab783d53))

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -365,49 +365,108 @@ Sort an array of tiddler titles by a specified field
 */
 exports.sortTiddlers = function(titles,sortField,isDescending,isCaseSensitive,isNumeric,isAlphaNumeric) {
 	var self = this;
-	titles.sort(function(a,b) {
-		var x,y,
-			compareNumbers = function(x,y) {
-				var result = 
-					isNaN(x) && !isNaN(y) ? (isDescending ? -1 : 1) :
-					!isNaN(x) && isNaN(y) ? (isDescending ? 1 : -1) :
-											(isDescending ? y - x :  x - y);
-				return result;
-			};
-		if(sortField !== "title") {
-			var tiddlerA = self.getTiddler(a),
-				tiddlerB = self.getTiddler(b);
-			if(tiddlerA) {
-				a = tiddlerA.fields[sortField] || "";
+	if(sortField === "title") {
+		if(!isNumeric && !isAlphaNumeric) {
+			if(isCaseSensitive) {
+				if(isDescending) {
+					titles.sort(function(a,b) {
+						return b.localeCompare(a);
+					});
+				} else {
+					titles.sort(function(a,b) {
+						return a.localeCompare(b);
+					});
+				}	
 			} else {
-				a = "";
+				if(isDescending) {
+					titles.sort(function(a,b) {
+						return b.toLowerCase().localeCompare(a.toLowerCase());
+					});
+				} else {
+					titles.sort(function(a,b) {
+						return a.toLowerCase().localeCompare(b.toLowerCase());
+					});
+				}	
 			}
-			if(tiddlerB) {
-				b = tiddlerB.fields[sortField] || "";
-			} else {
-				b = "";
-			}
-		}
-		x = Number(a);
-		y = Number(b);
-		if(isNumeric && (!isNaN(x) || !isNaN(y))) {
-			return compareNumbers(x,y);
-		} else if($tw.utils.isDate(a) && $tw.utils.isDate(b)) {
-			return isDescending ? b - a : a - b;
-		} else if(isAlphaNumeric) {
-			a = String(a);
-			b = String(b);
-			return isDescending ? b.localeCompare(a,undefined,{numeric: true,sensitivity: "base"}) : a.localeCompare(b,undefined,{numeric: true,sensitivity: "base"});
 		} else {
+			titles.sort(function(a,b) {
+				var x,y;
+				if(isNumeric) {
+					x = Number(a);
+					y = Number(b);
+					if(isNaN(x)) {
+						if(isNaN(y)) {
+							return 0;
+						} else {
+							return isDescending ? -1 : 1;
+						}
+					} else {
+						if(isNaN(y)) {
+							return isDescending ? 1 : -1;
+						} else {
+							return isDescending ? y - x :  x - y;
+						}
+					}
+				}
+				if(isAlphaNumeric) {
+					return isDescending ? b.localeCompare(a,undefined,{numeric: true,sensitivity: "base"}) : a.localeCompare(b,undefined,{numeric: true,sensitivity: "base"});
+				}
+				if(!isCaseSensitive) {
+					a = a.toLowerCase();
+					b = b.toLowerCase();
+				}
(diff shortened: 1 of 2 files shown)


pmario:
There seems to be something wrong. The "sameday" filter should be only active 2 times. Now it is 4 

grafik

Jermolene:
> There seems to be something wrong. The "sameday" filter should be only active 2 times. Now it is 4
> I'm using the test wiki from Hubert and this PR. Page reload with Recent tab selected

How would I replicate that test?

Jermolene:
These changes are quite hard to review, but I'd really appreciate another pair of eyes if anyone else can have a look.

saqimtiaz:
It is pretty tricky to review this as a diff/refactoring, so I looked through the new code on its own and it looks sound to me.

Jermolene:
Thanks @saqimtiaz it is devilishly difficult; the existing logic has some unexpected fallthroughs...

I'd like to track down @pmario's problem before merging this, and then asking users for some wider testing.

saqimtiaz:
> Thanks @saqimtiaz it is devilishly difficult; the existing logic has some unexpected fallthroughs...

If we don't already have tests for the edge cases then adding some would be one way to confirm that the new code behaves consistently with the old.

Jermolene:
> If we don't already have tests for the edge cases then adding some would be one way to confirm that the new code behaves consistently with the old.

I have added some more tests as I traced through the code in 0be39cfbc2829dbd8637f35110b33c4bd2416dc3.

pmario:
>How would I replicate that test?

 - I did download your `sort-optimisations` branch
 - created emtpy.html: `tiddlywiki .\editions\empty\ --build empty`
 - imported: `5.2.0-prerelease-perfm-5003_tids.html`
 - select: Recent tab
 - Save wiki
 - F12 dev console
 - F5 reload wiki
 - `$tw.perf.log()`

It's strange. To write this info I did do the steps and now I can't reproduce the problem anymore. 

But there is still the question, why is the filter executed twice instead of once directly after the startup.

grafik

core/modules/wiki.js
```diff
@@ -365,49 +365,108 @@ Sort an array of tiddler titles by a specified field
...
+				if(!isCaseSensitive) {
+					a = a.toLowerCase();
+					b = b.toLowerCase();
+				}
+				return isDescending ? b.localeCompare(a) : a.localeCompare(b);
+			});
+		}
+	} else {
+		titles.sort(function(a,b) {
+			var x,y;
```
pmario, line 423: if `a and b` would be initialized with `a="", b=""` here the else path in line 429 and 434 are redundant

pmario:
In the `if(sortField === "title") {` path you did most of the isXX logic outside the sort function
in the `else {` path you do most of this logic inside the sort function again. ... Is there a reason for this difference?
I think the first approach is faster, since it contains less code in the repeated sort code

core/modules/wiki.js
Jermolene: Thanks that is true.

Jermolene: Oh sorry, `a` and `b` are function parameters and so cannot be initialised like that.

Jermolene:
> In the `if(sortField === "title") {` path you did most of the isXX logic outside the sort function
> in the `else {` path you do most of this logic inside the sort function again. ... Is there a reason for this difference?
> I think the first approach is faster, since it contains less code in the repeated sort code

I'm not sure that I understand. The final code block is the complete one, the preceding ones isolate special cases.

Jermolene:
Thanks @pmario 

> But there is still the question, why is the filter executed twice instead of once directly after the startup.

Those per-filter timings are cumulative. If there were a subsequent refresh after the initial one then one would expect to see the number of executions of the filters to increase.

core/modules/wiki.js
pmario: uups you are right

pmario:
>Those per-filter timings are cumulative. If there were a subsequent refresh after the initial one then one would expect to see the number of executions of the filters to increase.

That's right. But the only thing that happens is: Reload page ... So I would expect it to be executed only once, but is shows 2

Jermolene:
> That's right. But the only thing that happens is: Reload page ... So I would expect it to be executed only once, but is shows 2

OK the question is whether you're seeing the same number of refreshes in the performance log for both of the test wikis.

Have you tried running the test rig? It does make this sort of thing a lot easier to manage.

hubertgk:
Hi @Jermolene et al, when these changes are merged so as to be accessible on the [prerelease page](https://tiddlywiki.com/prerelease/) for download in an empty edition I'll be happy to test the version manually as before if this can help in any way (I don't know JS so I'm unable to contribute otherwise).

pmario:
@hubertgk  ... I did upload a version with the new code for you to test here.
[empty.zip](https://github.com/Jermolene/TiddlyWiki5/files/7206003/empty.zip)

Jermolene:
At the moment the blocker for merging is that I can't reproduce the problem reported by @pmario [here](https://github.com/Jermolene/TiddlyWiki5/pull/6053#issuecomment-922933299), if anyone else can help attempt to duplicate the problem I'd be very grateful.

pmario:
>At the moment the blocker for merging is that I can't reproduce the problem reported by @pmario here, if anyone else can help attempt to duplicate the problem I'd be very grateful.

Sorry. This shouldn't be a blocker anymore. .. The problem I saw was the [4 times activation](https://github.com/Jermolene/TiddlyWiki5/pull/6053#issuecomment-922527140) of the filters. .. The 2 times are there since the performance log was activated. 

------

I'm only curious, why they are activated 2 times instead of once, if you do a page reload. .. But that shouldn't be a blocker

saqimtiaz:
I can confirm there is no discrepancy in the number of times in the filters are run between 5.1.23, the current pre-release and this branch.

Jermolene:
Thank you @pmario @saqimtiaz I'll merge this.

@hubertgk the changes will automatically appear on https://tiddlywiki.com/prerelease in 5-10 minutes.

hubertgk:
Thank you @Jermolene and @pmario

I've tested my own wiki running the updated 5.2.0-prerelease on mobile just now and can confirm it's as snappy as it's ever been, perhaps even faster now than 5.1.23!

Thank you @Jermolene @saqimtiaz @pmario and everyone involved for your work on this issue.

pmario:
>Thank you @Jermolene @saqimtiaz @pmario and everyone involved for your work on this issue.

Thank YOU for pointing out the issue ;)

Jermolene:
That's great, thank you @hubertgk!

PR #6055, opened 2021-09-19
Added modulesproperty filter operator and extended modules operator

### LINKED ISSUE #6009: moduleproperty[] and extended modules[] operators in core
I suggest that we include the `moduleproperty[]` and `modules[]` operators from the publishing framework in the core for 5.2.0:

* https://github.com/Jermolene/TiddlyWiki5/blob/bbdf9bae89a732d472e7569faded463e16839e56/core/modules/filters/moduleproperty.js
* https://github.com/Jermolene/TiddlyWiki5/blob/bbdf9bae89a732d472e7569faded463e16839e56/core/modules/filters/modules.js

Note that in the latter operator we may want to have another look to ensure that it is backwards compatible when there are no operands provided.

I am wrapping up my current round of work on the FileUploads plugin and while I don't have the opportunity to prepare a PR, I wanted to post this item from my notes lest it be forgotten.

In the FileUploads plugin I had to introduce an `uploaders[]` operator to get the module names of modules of the type uploader, this could have been avoided. Similarly, at the moment the Settings tiddler for the plugin transcludes the Settings tiddler of the selected uploader module, but this is based on naming convention and is brittle. It wont work if someone were to write an uploader module with a different plugin tiddler namespace. The two operators under discussion from the publishing framework would address both these use cases.

saqimtiaz:
Added `modulesproperty` filter operator and extended `modules` operator from the publishing framework.
Please see the examples in the included documentation for use cases.

Fixes #6009

### DIFF
--- a/core/modules/filters/moduleproperty.js
+++ b/core/modules/filters/moduleproperty.js
@@ -1,0 +1,30 @@
+/*\
+title: $:/core/modules/filters/moduleproperty.js
+type: application/javascript
+module-type: filteroperator
+
+Filter [[module-name]moduleproperty[name]] retrieve a module property
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports.moduleproperty = function(source,operator,options) {
+	var results = [];
+	source(function(tiddler,title) {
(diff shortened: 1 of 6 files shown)


Jermolene:
Thanks @saqimtiaz

PR #6056, opened 2021-09-20
Linked-List refactor

flibbles:
I was a little disappointed with some of the performance improvements I was seeing by switching to LinkedList in select cases. And then I had a brainstorm.

When I originally wrote LinkedList, I avoided using the newer Javascript Map because of backward compatibility, but it's a simple class to mimic. So I rewrote LinkedList to use a Map instead of an object, and for people using super-old versions of IE, they get a makeshift little map substitute.

...and it turns out that map substitute is faster than a map in most cases.

Maps are 30% slower on chrome across the board, whether you have 5000 objects or 30,000
They're moderately faster in Firefox once you start getting more than 15,000, but not by enough to really matter.

_I was disappointed._

However, I'm still pushing my overhaul using the adhoc map, because it **is** actually faster (or at least equivalent), the file is only 4 lines of code longer, and most importantly, **the code logic is way easier for someone else to understand**.

By the way, are we free to use Maps and Sets? They're supported as far back as IE 11, and there are a few possible cases where they may be the better replacement for arrays than this LinkedList would be.

### DIFF
--- a/core/modules/utils/linked-list.js
+++ b/core/modules/utils/linked-list.js
@@ -15,8 +15,8 @@ function LinkedList() {
 
 LinkedList.prototype.clear = function() {
 	// LinkedList performs the duty of both the head and tail node
-	this.next = Object.create(null);
-	this.prev = Object.create(null);
+	this.next = new OurMap();
+	this.prev = new OurMap();
 	this.first = undefined;
 	this.last = undefined;
 	this.length = 0;
@@ -77,7 +77,7 @@ LinkedList.prototype.each = function(callback) {
 		value = this.first;
 	while(value !== undefined) {
 		callback(value);
-		var next = this.next[value];
+		var next = this.next.get(value);
 		if(typeof next === "object") {
 			var i = visits[value] || 0;
 			visits[value] = i+1;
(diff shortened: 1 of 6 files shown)


core/modules/utils/linked-list.js
```diff
@@ -105,91 +107,79 @@ LinkedList.prototype.makeTiddlerIterator = function(wiki) {
...
+	var nextEntry = list.next.get(value);
+	if(nextEntry === undefined) {
+		return;
+	}
+	var prevEntry = list.prev.get(value),
 		prev = prevEntry,
-		next = nextEntry;
-	if(typeof nextEntry === "object") {
+		next = nextEntry,
+		ref;
```
pmario, line 117: ????? What is this `ref doing here?`

Jermolene:
> However, I'm still pushing my overhaul using the adhoc map, because it **is** actually faster (or at least equivalent), the file is only 4 lines of code longer, and most importantly, **the code logic is way easier for someone else to understand**.

Great, thank you @flibbles 

> By the way, are we free to use Maps and Sets? They're supported as far back as IE 11, and there are a few possible cases where they may be the better replacement for arrays than this LinkedList would be.

The core doesn't use them at the moment. To bring them in we'd need to push forwards our JS cut off date. For various reasons, I think the big win is to move as far as await/async, which I think would be ECMAScript 2018. That suggests that a policy of supporting the last 5 years of browser releases might work for us. To stress that that would be a big change from where we are today, but I think it would be reasonable balance of modernity and ubiquity...

flibbles:
After 11 months. Conflicts resolved, and this refactor is once again good to merge. It's simpler, shorter, faster, easier to understand, doesn't have the issue in #7059, and has absolutely no reason why it shouldn't merge.

Jermolene:
Thanks @flibbles

PR #6057, opened 2021-09-20
JSONTiddler a little more extensible

flibbles:
As per discussion #6043. This will allow plugins to inherit JSONTiddler and change a tiddlywiki's write-to-file behavior.


flibbles:
I realize I committed the sin of not providing context and instead just linking to a discussion. This PR will allow plugins such as (TW5-Uglify) to intercept a tiddler's fields before they're stringified, without the plugin needing to completely remake the widget.

Jermolene:

PR #6063, opened 2021-09-22
Fix for action-confirm: avoid errors if event is not defined

saqimtiaz:
This PR fixes a bug in the pre-release introduced in #5776  where the `<$action-confirm>` widget results in a red error screen when `event` is not defined.


saqimtiaz:
@Jermolene this needs attention before 5.2.0 ships as it is a new bug due to a PR that has been merged.

Jermolene:
Thanks @saqimtiaz, good spot

PR #6068, opened 2021-09-25
Update the Android section in Saving via WebDAV

xcazin:
_webdav server_ isn't available anymore on Google Play. For Android, I suggest the RCX app instead.


Jermolene:
Thanks @xcazin – this may be better directed to the "tiddlywiki-com" branch so that it can go live immediately.

xcazin:
Hi @Jermolene, never did that :-) After a bit of confusion, I guess I eventually managed to commit and request a pull to tiddlywiki-com 🤞

PR #6069, opened 2021-09-25
Update the Android section in Saving via WebDAV

xcazin:
_webdav server_ isn't available anymore on Google Play. For Android, I suggest the RCX app instead.


Jermolene:
Thanks @xcazin

PR #6076, opened 2021-09-29
Signing TW CLA

bmann:
Signing the CLA in order to make contributions.

### DIFF
--- a/licenses/cla-individual.md
+++ b/licenses/cla-individual.md
@@ -461,3 +461,5 @@ Frank Bruns-Ballhausen, @FBB-de, 2021/06/09
 RJ Skerry-Ryan, @rryan, 2021/07/03
 
 J. Song, @8d1h, 2021/08/05
+
+Boris Mann, @bmann, 2021/09/28


Jermolene:
Thanks @bmann – although perhaps inevitably given the recent discussions on the forum, there is a small problem: signatures for the CLA should target the "tiddlywiki-com" branch, not "master". GitHub offers a UI for doing this, but general experience seems to be that it doesn't always work, and so people often end up making an entirely fresh PR.

pmario:
@Jermolene There are 6 merged branches in the repo. I think it would be easier for users, if you delete them. see: https://github.com/Jermolene/TiddlyWiki5/branches/all

Jermolene:
> @Jermolene There are 6 merged branches in the repo. I think it would be easier for users, if you delete them. see: https://github.com/Jermolene/TiddlyWiki5/branches/all

Thanks @pmario, done.

bmann:
Sorry I'm rubbish at GitHub notifications. No problem, I can retarget. 

I looked, and I guess you deploy TW-com frequently and then just merge per release for master? So the contributors file then also ships per release?

Jermolene:
Thanks @bmann 

> I looked, and I guess you deploy TW-com frequently and then just merge per release for master? So the contributors file then also ships per release?

That's right – we also merge tiddlywiki-com into master at fairly regular intervals.

The rationale for having the CLA in the tiddlywiki-com branch is that it simplifies things for people contributing to the docs; they only have to deal with a single branch (albeit not the default branch).

pmario:
@Jermolene ... Since there is a new version on the horizon, IMO this PR could be merged to master.

Jermolene:
Thanks @pmario and @bmann

PR #6082, opened 2021-10-02
Add text to QR Code caption in config panel

linonetwo:
Before

截屏2021-10-03 01 05 45

After

截屏2021-10-03 01 06 14


Jermolene:
Thanks @linonetwo well spotted!

PR #6104, opened 2021-10-11
Improve $action-listops attribute docs

btheado:
Added documentation to the `$action-listops` widget to compare and contrast the `$filter`, `$subfilter`, and `$tags` attributes.

Also mentioned how the `$action-listops` widget compares to the `$action-setfield` widget and gave some guidance on when to use each.


Jermolene:
Thanks @btheado

PR #6112, opened 2021-10-13
add a core icon selector to the docs

pmario:
Code comes from https://talk.tiddlywiki.org/t/how-to-display-core-icons-as-thumbnails-and-copy-to-clipboard-on-click/1141/2

There is a screencast which shows how it works. I did also add a link to an other doc tiddler. Pls check the 1 liner for typos!

### DIFF
--- a/editions/tw5.com/tiddlers/demonstrations/ImageGallery Example.tid
+++ b/editions/tw5.com/tiddlers/demonstrations/ImageGallery Example.tid
@@ -1,0 +1,11 @@
+created: 20140809113603449
+modified: 20211022195248529
+tags: Learning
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @pmario a few thoughts:

* The new tiddler duplicated the existing "Image Gallery" with the addition of the copy-to-clipboard function. It seems confusing to have two different image galleries that look similar but have different functionality. I think this would be better either as an update to the existing "Image Gallery" tiddler
* Exposing the tiddler via the "ten things" tiddler makes it unlikely that people will find it. If it's a useful piece of functionality that it needs to be exposed appropriately

PR #6132, opened 2021-10-21
Extend HTML tag parser to maintain an ordered array of attribute names

Jermolene:
This PR extends the wikitext parse rule used for HTML elements to record the ordering of the attributes.

The goal is to enable a future variant of the `<$vars>` widget to respect the ordering when evaluating the attributes, such that variable assignments could depend on the results of previous assignments in the same widget.

Sadly, it's too late to make that change to the existing `<$vars>` widget itself without breaking backwards compatibility, but we could consider introducing a variant with a different name with the new functionality.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1950,6 +1950,20 @@ $tw.loadTiddlersFromSpecification = function(filepath,excludeRegExp) {
 			tiddlers.push({tiddlers: fileTiddlers});
 		}
 	};
+	// Helper to recursively search subdirectories
+	var getAllFiles = function(dirPath, recurse, arrayOfFiles) {
+		recurse = recurse || false;
+		arrayOfFiles = arrayOfFiles || [];
+		var files = fs.readdirSync(dirPath);
+		files.forEach(function(file) {
+			if (recurse && fs.statSync(dirPath + path.sep + file).isDirectory()) {
+				arrayOfFiles = getAllFiles(dirPath + path.sep + file, recurse, arrayOfFiles);
+			} else if(fs.statSync(dirPath + path.sep + file).isFile()){
+				arrayOfFiles.push(path.join(dirPath, path.sep, file));
+			}
+		});
+		return arrayOfFiles;
+	}
 	// Process the listed tiddlers
 	$tw.utils.each(filesInfo.tiddlers,function(tidInfo) {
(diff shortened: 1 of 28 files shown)


saqimtiaz:
@flibbles I think one of the main concerns is going to be that the change isn't backwards compatible, and therefore rather annoyingly this probably needs to be a new widget rather than an update of the vars widget.

flibbles:
The way I redesigned the $vars widget, `<$vars a={{{ [<a>add[1]] }}}>` would work fine, since $vars only returns variables once its widget has been fully computed. Otherwise it looks up to its parent as usual.

This means a $vars attribute can only pull variables from attributes preceding it, not itself or proceeding attributes.

However, if we do end up creating a new vars alternative, I propose the `<$let>` widget.

flibbles:
Right. So `<$vars a={{{ [<a>add[1]] }}}>` would still work, but things like `<$vars currentTiddler="Title" myVar={{!!field}} >` would behave differently.

Anyway, it's possible. I know this PR isn't actually about the $vars widget, so I'll drop the matter, but if people get interested in a better $vars-like widget, just say the word and I can throw together a nice PR.

Jermolene:
> Anyway, it's possible. I know this PR isn't actually about the $vars widget, so I'll drop the matter, but if people get interested in a better $vars-like widget, just say the word and I can throw together a nice PR.

We've established that backwards compatibility will require a new widget, and I like your suggestion of `<$let>`. If there's no objection, I think the next step is to merge this PR, and then I'd welcome a PR for the `<$let>` widget.

flibbles:
It would also be falling _away_ from the paradigm of how javascript's let functions.

Jermolene:
> I'm just really reluctant to have "orderedAttributes" contain so much extra info for every single widget, when only <$let> would make use of it, and only about 0.1% of the time.

The objects that would be in that "orderedAttributes" array already exist as properties of the "attributes" object and so the only "extra information" would be additional references to those objects.

flibbles:
All right. Fair enough. I'm just concerned that repeating attributes will introduce complexity which will be a pain to maintain. Particularly for how few situations it seems like it would be applicable to.

But however orderedAttributes ends up working, I'll make a $let widget that works with it. It'll handle repeating attributes if the core does.

PR #6135, opened 2021-10-23
Unified html syntax

nilslindemann:
This is the regeneration of #4290 on the current codebase.

[A wiki explaining what the pull request does is online](https://nilslindemann.github.io/TiddlyWiki5-unified-html-syntax--PR-Work-Folder/singlefile.html). There, the Tiddler [WikiText to HTML Examples](https://nilslindemann.github.io/TiddlyWiki5-unified-html-syntax--PR-Work-Folder/singlefile.html#WikiText%20to%20HTML%20Examples) explains the changes.

I also created a [repository](https://github.com/nilslindemann/TiddlyWiki5-unified-html-syntax--PR-Work-Folder) containing my work folder for this PR.

For more info, read the commit messages, and check out the tests I added to [test-wikitext-parser.js](https://github.com/nilslindemann/TiddlyWiki5/blob/unified-html-syntax/editions/test/tiddlers/tests/test-wikitext-parser.js#L25-L258).


nilslindemann:
Hi @pmario @flibbles 

Seems that I structured my commits here not too well.

I will generate a new PR making these changes, where I will be more conservative regarding addition / removal of newlines, in order to document better where this is needed.

I will then also only update that PR when there are breaking changes.

And I will structure the commits based on the type of change I made, instead of the location where I made it.

I will also leave editions other than tw5.com away, in order to make that PR smaller. I anyway didn't completely skim through these editions.

@Jermolene I have not forgotten your POV and I agree, we should do it using a mime type or a vocabulary parameter, or something else, to keep backwards compatibility. I will generate this new PR, so that interested people who don't want to wait and accept backwards breaking changes, can use it, until we figured out how to do it. I myself are not sure how to do it.

* A new mime type is simple to use and implement, but then we would have another WikiText type and that may irritate users. And what if we do another change to the parser? Add another mime type? And another?
* Vocabulary parameters are more flexible, but are not so simple to use for users, i guess, and it is not clear how to implement them.
* Maybe there is a third way, like making the [Parsing](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FParsing) tiddler in the Control panel behave comparable to the [Palettes](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FPalette) Tiddler. Aka, users can clone their individual "Parsing palette", set a default palette for new tiddlers, and can choose it from a drop-down in individual tiddlers.

pmario:
>Seems that I structured my commits here not too well.

Hi Nils, Sorry for causing trouble! ... I did write [my long comment first](https://github.com/Jermolene/TiddlyWiki5/pull/6135#issuecomment-951926077) and then [the shorter one](https://github.com/Jermolene/TiddlyWiki5/pull/6135#issuecomment-951981234) after experimenting and reading your documentation "line by line". 

The new rules that I didn't get right in my mind as I wrote the first post where:

>The syntaxes intuitively document where the paragraph wrapping happens.
>
>  **No linebreak** after the opening tag: The parent gets wrapped.
    **One linebreak** after the opening tag: Nothing gets wrapped.
    **Two linebreaks** after the opening tag: The child gets wrapped.
>
>As a result of these changes, the way self-closing and void elements get wrapped also changes. I document this later in the text.

Now I see, they make sense and are actually "easy" to handle, **if you know them ;)** ... You are right. They are different to what we are used to atm.

----

I think, I also found out that you added the new backslash `\` syntax in many places to make the html output the exact same as with TW 5.1.23. I think the new syntax is used mainly in DOC tiddlers. right? ... 

We could discuss if the "extra P-tags" added with `\` are needed in first place. ... 

----

I think, it's easier to get a PR merged, if it only changes the minimum tiddlers necessary.

nilslindemann:
@pmario No problem, it was a good thing to go through this code another time, it clarified a few more things to me.

The new pull request is #6156, closing this one.

PR #6148, opened 2021-10-28
The $let widget

flibbles:
Here is the $let widget, which takes advantage of orderedAttributes in order to allow for interdependent variables, as well as multiple definitions of the same variable.

I made a few design decisions while making this:

* I broke some code out of computeAttributes in widget.js" into a computeAttribute method, which is a zero-side-effect method which just returns the value of a passed attribute. I did this to cut down on repetition when $let makes its own computeAttributes method
* I opted not to have the core widget's computeAttributes evaluate attributes in order the way $let does. Since not all widgets have orderedAttributes, and not all attributes in widgets have their "name" member filled out, it made for some ugly logic. But since the order was (and remains) undefined, this can be changed later if desired.

Otherwise, this does everything it's promised to do. It took some special care to make sure it doesn't refresh unnecessarily, but it should all be good now. Also included some documentation (and I touched up varsWidget documentation to use the .wid and .wlink macros when appropriate).

Let me know what ya'll think.

### DIFF
--- a/core/modules/widgets/let.js
+++ b/core/modules/widgets/let.js
@@ -36,8 +36,6 @@ LetWidget.prototype = new Widget();
 Render this widget into the DOM
 */
 LetWidget.prototype.render = function(parent,nextSibling) {
-	// Call the constructor
-	Widget.call(this);
 	this.parentDomNode = parent;
 	this.computeAttributes();
 	this.execute();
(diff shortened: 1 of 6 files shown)


saqimtiaz:
@Jermolene look forward to your thoughts on this when you have the chance to review.

Jermolene:
Thanks @flibbles, great stuff. Apologies for the delay, I wanted to properly check this out.

PR #6150, opened 2021-10-29
Zero based zth[] operator and documentation

### LINKED ISSUE #6122: [IDEA] zero based nth[] operator
While the `nth[]` operator starts at 1, the `:reduce` (and soon `:map`) prefixes are zero based as is the `reduce[]` filter operator. This means you cannot use the `index` variable inside the reduce filter with the `nth[]` operator.

Suggestions:
* Either we introduce a new operator akin to `nth[]` that is zero based,
* or we extend `nth[]` to have an affordance to switch it to being zero based, keeping the existing behaviour as default

saqimtiaz:
This PR adds a `zth[]` operator, a zero based index equivalent for the `nth[]` operator.

Fixes #6122

### DIFF
--- a/core/modules/filters/listops.js
+++ b/core/modules/filters/listops.js
@@ -103,4 +103,16 @@ exports.nth = function(source,operator,options) {
 	return results.slice(count - 1,count);
 };
 
+/*
+The zero based nth member of the list
+*/
+exports.zth = function(source,operator,options) {
+	var count = $tw.utils.getInt(operator.operand,0),
+		results = [];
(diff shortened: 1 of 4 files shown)


kookma:
Is the difference with `nth` is only the base? base 0 vs. base 1?
Can't we use nth with some manipulation to reproduce zth behaviour?

saqimtiaz:
@kookma the linked issue provides an example for a situation where such manipulation is not possible.

Jermolene:
Thanks @saqimtiaz

PR #6151, opened 2021-10-29
Update getScrollContainer method to work more reliable

BurningTreeC:
This PR updates the `getScrollContainer` method in `utils/dom.js` to work more reliable
It now detects reliably the real scroll container of an element. Before it almost always detected the document.body

This PR uses https://stackoverflow.com/questions/35939886/find-first-scrollable-parent, a pure JS port of the jQuery UI `scrollParent` method and a fix for the `scrollTop Bug` - see https://dev.opera.com/articles/fixing-the-scrolltop-bug/

This PR then would allow us to use `getScrollContainer` in various places, like in the `pop` storyview to detect the `scrollContainer` of the inserted element and setting its x-overflow to `hidden` during the insert-animation and in the `pageScroller` `scrollIntoView` method where we could use the elements scrollContainer for scrolling - not `window.scrollTo`

I'd like to leave this here for discussion - there are two or three older PR's that tried to sneak this into the core but I believe they were premature

See: #6019 , #6016 , #6025

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -66,16 +66,39 @@ exports.toggleClass = function(el,className,status) {
 
 /*
 Get the first parent element that has scrollbars or use the body as fallback.
+From https://stackoverflow.com/questions/35939886/find-first-scrollable-parent/42543908#42543908
+Also, fix for scrollTop bug: https://dev.opera.com/articles/fixing-the-scrolltop-bug/
 */
-exports.getScrollContainer = function(el) {
+exports.getScrollContainer = function(el,includeHidden) {
 	var doc = el.ownerDocument;
-	while(el.parentNode) {
-		el = el.parentNode;
-		if(el.scrollTop) {
-			return el;
+	var style = getComputedStyle(el);
+	var excludeStaticParent = style.position === "absolute";
+	var overflowRegex = includeHidden ? /(auto|scroll|hidden)/ : /(auto|scroll)/;
+	if(style.position === "fixed") {
+		if("scrollingElement" in doc) {
+			return doc.scrollingElement;
+		}
+		if(navigator.userAgent.indexOf("WebKit") !== -1) {
+			return doc.body;
+		}
+		return doc.documentElement;
+	}
+	for(var parent=el; parent=parent.parentElement; ) {
+		style = getComputedStyle(parent);
+		if(excludeStaticParent && style.position === "static") {
+			continue;
+		}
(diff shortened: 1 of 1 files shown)


BurningTreeC:
Hello all, there's not been any interest in this...
But using this we could make the pageScroller work for scrollable div's anywhere on the page
Wouldn't that be a great benefit?

pmario:
I think, it's not about the interest. It just has been forgotten.

BurningTreeC:
does anybody have any thoughts on this? 🤣

pmario:
I think, it would be nice to have a working example, where it makes a difference. So may be some JSON tiddlers, that we can import at tiddlerwiki.com and see the current behaviour. 

A wiki with the patch applied using the same JSON config to see the difference. ... 

I did have a look, where `.getScrollContainer()` is used in the core. 

It's at used in the same file at:  `exports.resizeTextAreaToFit = function(domNode,minHeight) {` ... So we need to make sure it still works there.

Jermolene:
Hi @BurningTreeC 

> This PR updates the `getScrollContainer` method in `utils/dom.js` to work more reliable It now detects reliably the real scroll container of an element. Before it almost always detected the document.body
> 
> This PR uses https://stackoverflow.com/questions/35939886/find-first-scrollable-parent, a pure JS port of the jQuery UI `scrollParent` method and a fix for the `scrollTop Bug` - see https://dev.opera.com/articles/fixing-the-scrolltop-bug/

It's pretty gnarly code: regexps, and I really don't like the strange `for` loop.

It would be useful to see some other approaches the problem if possible.

> This PR then would allow us to use `getScrollContainer` in various places, like in the `pop` storyview to detect the `scrollContainer` of the inserted element and setting its x-overflow to `hidden` during the insert-animation

That's definitely a problem, but it is not an attractive approach for a widget to reach up and manipulate arbitrary DOM elements belonging to parent widgets.

> and in the `pageScroller` `scrollIntoView` method where we could use the elements scrollContainer for scrolling - not `window.scrollTo`

It would certainly be nice to improve scrollIntoView, but nowadays browsers have a pretty good native implementation. The only problem I've found is that they can only have one outstanding scroll animation at a time, which is a pain when working on multiple column story rivers :smile:

> I'd like to leave this here for discussion - there are two or three older PR's that tried to sneak this into the core but I believe they were premature
> 
> See: #6019 , #6016 , #6025

BurningTreeC:
Thanks @Jermolene 

How about this idea:

- We add the class `tc-scroll-container` to all scrollable elements
- In the pageScroller we detect the element's next ancestor with the class `tc-scroll-container`
- We use that ancestor for scrolling

PR #6153, opened 2021-10-30
Ensure the operator examples have unique id

btheado:
The first parameter of the operator examples macro is used for
constructing unique state tiddler titles. The cycle, log, and
match operators had duplicates, causing examples to share state
with each other.


btheado:
Maybe the operators example macro was written before the qualify macro existed? I didn't look very closely, but maybe this PR is just a band-aid and the proper fix is to change it to use the qualify macro instead?

Jermolene:
Thanks @btheado 

> Maybe the operators example macro was written before the qualify macro existed? I didn't look very closely, but maybe this PR is just a band-aid and the proper fix is to change it to use the qualify macro instead?

The qualify macro doesn't help in cases like this because the instances of the macro are all siblings, and so would share the same "recursion marker" (ie the magic hash derived from the list ancestors).

PR #6155, opened 2021-10-31
Add subfilter operator examples and mention when it is a constructor

btheado:
* Added 3 subfilter operator examples as well as some explanation for the examples.
* Mention when the subfilter operator does and does not operate as a selection constructor.


Jermolene:
Thanks @btheado, great stuff.

PR #6156, opened 2021-10-31
Refined html syntax

nilslindemann:
Regeneration of #6135 (which is the regeneration of #4290), this time with more conservative code changes, as requested by @pmario, and a better structure and self-documentation of the commits.

This pull request is not intended to be merged as it is, as @Jermolene [expressed his concerns about backwards compatibility](https://github.com/Jermolene/TiddlyWiki5/pull/6135#issuecomment-950128144). We need to consider that and find a smart solution.

The pull request exists

* to document what I intend to do
* to allow you to use it right now

[See the related online wiki](https://nilslindemann.github.io/TiddlyWiki5-unified-html-syntax--PR-Work-Folder/singlefile.html#Table%20of%20Contents), which has these changes applied to its source, so you can play around with them. [Its source can be downloaded here](https://github.com/nilslindemann/TiddlyWiki5-unified-html-syntax--PR-Work-Folder).

And [here is a clone of the official docs](https://nilslindemann.github.io/TiddlyWiki5-unified-html-syntax--PR-Work-Folder/tw5-singlefile.html), also with the changes in this pull request applied to its source.

Update 2021-11-1: Actually, above wikis are now based on an [independent "*-diff" branch](https://github.com/nilslindemann/TiddlyWiki5/tree/refined-html-syntax-diff), because I revert some things here.

PR #6157, opened 2021-10-31
fix - headings are block elements, and not allowed inside of paragraphs

pmario:
As @nilslindemann found out in: https://github.com/Jermolene/TiddlyWiki5/pull/6156/commits/6103fce5bd862ebe2ff2eb28428e18c1a4c0670f  headings shouldn't be wrapped in P-tags according to the html specs.

### DIFF
--- a/plugins/tiddlywiki/dynannotate/examples/combined.tid
+++ b/plugins/tiddlywiki/dynannotate/examples/combined.tid
@@ -58,11 +58,9 @@ Search: <$edit-text tiddler="$:/temp/search" tag="input"/>
 <div class="tc-drop-down-wrapper">
 <div class="tc-drop-down tc-popup-keep" style="max-width:550px;white-space: normal;overflow-y:hidden;">
 <$tiddler tiddler={{$:/temp/dynannotate/demo/annotation-title}}>
-<p>
 <h2>
 This is an annotation
 </h2>
-</p>
 <p>
 The annotation is stored in the tiddler:
 </p>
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @pmario – do the examples still work correctly without the `<p>` tags?

pmario:
yes, hx have their own margin-top and -bottom which is even more than the P-tag. See test and image: 

```
!! heading 2

some text 

<p>

!! heading 2
<p>
some text 
```

grafik

Jermolene:
HI @pmario the PR concerns the data used by the dynannotate plugin demo. Did you verify that the demo works correctly with this change applied?

pmario:
>Did you verify that the demo works correctly with this change applied?

Sorry I did not. I didn't think that the plugin UI uses the p-tag as a hack to apply formatting in the dropdown instead using CSS for styling. ... I think the whole dropdown needs a new PR. 

I will close this one

nilslindemann:
Yes, the dropdown should add padding, not the children margin.

I guess Jermolene did it like this, because dropdowns inside of dropdowns behave different regarding their children. Though I can currently not find an example of such an inner dropdown.

Interesting read: [The island object](https://csswizardry.com/2011/10/the-island-object/). Dropdowns, tiddlers, pre's, ..., can all be seen as some type of island, aka, they have a visible border, and space between the border and the children. In all these cases it is a good idea to let the isle add the padding and set the margin-bottom of the last child (or the margin-top of the first child) to 0.

pmario:
@nilslindemann Very interesting read and it would be a good approach for the plugin dialogues.

PR #6161, opened 2021-11-01
Add MessageHandlerWidgets and TriggeringWidgets tags

btheado:
Add two new Widget tags to the documentation. The `MessageHandlerWidgets` are those which receive and process widget messages, such as the NavigatorWidget and LinkCatcherWidget. The `TriggeringWidgets` are those which can trigger actions such as the ButtonWidget and CheckboxWidget.


Jermolene:
Thanks @btheado

PR #6162, opened 2021-11-01
fix: resolved search-replace operator bug 

### LINKED ISSUE #6129: [BUG] Search-Replace with $ in Search String
**Describe the bug**
The search-replace eats `$`

**To Reproduce**
2. Create a tiddler with content below

```
<$vars for="equation $$x$$ end." with="relation $$x$$  finish." source="This is equation $$x$$ end.">
{{{ [<source>search-replace<for>,<with>] }}}
</$vars>
```

See the result 

```
This is relation $x$ finish.
```

The `search-replace` operator eats one of `$`

NOTE: I also tested with `regexp-mode` with no success!

**Expected behavior**
Should keep both `$$`!

 - Version 5.2.0
 - Saving mechanism -
 - Plugins installed -
 
 - OS: Windows 10
 - Browser Chrome 90

saqimtiaz:
This PR resolves a bug in the `search-replace` operator when not used in regexp mode and there are `$` characters in the replacement strings. The `$` character needs to be escaped as `$$`.

The `new RegExp` call has been earlier so we can re-use the same regular expression object.

More tests have been added as well as more examples to the documentation.

Fixes #6129


Jermolene:
Thanks @saqimtiaz

PR #6164, opened 2021-11-02
Support macro parameters in filter run prefixes

### LINKED ISSUE #6124: [BUG] Some filter run prefixes do not correctly resolve macro params
Filter run prefixes such as `:map` and `:reduce` (and others) that provide a mock `getVariable()` method do not resolve macro parameters.

This was missed as it wasn't at all apparent to write tests using `:map[]` or `:reduce[]` prefixes for macro params in filters. I did test some prefixes but not those ones.

I will create a PR with tests and a fix when I get the opportunity. The fix is simple, `getVariable()` needs to accept and pass on a second argument `options`. For example for `:map` :

```javascript
getVariable: function(name,options) {
	switch(name) {
		case "currentTiddler":
			return "" + title;
		case "..currentTiddler":
			return widget.getVariable("currentTiddler");
		default:
			return widget.getVariable(name,options);
	}
}
```

saqimtiaz:
This PR implements:
1. literal macro parameters within filter run prefixes that define their own variables, this fixes #6124 
2. makes the variables set by the filter run - such as `index`, `length`, `currentTiddler`, `..currentTiddler` - available within the macros used in the filter run. 

The changes to the Widget base class are only necessary for the latter, making the variables assigned by the filter run available within the macros.

This opens the door for things like this, where the variables `index`, `length` and `currentTiddler` will be accurately substituted.
```
\define .printf(text) $text$

{{{ [tag[HelloThere]] :map[<.printf "$(index)$ of $(length)$ - $(currentTiddler)$">] }}}

```


saqimtiaz:
@Jermolene this draft PR illustrates the Widget base class changes that I mentioned.

Jermolene:
Thanks @saqimtiaz that makes sense, and I think the changes to the Widget base class are reasonable.

saqimtiaz:
@Jermolene I have extended this PR to cover all the filter run prefixes that set their own variables, using the pattern from :map which you previously reviewed.

Jermolene:
Thanks @saqimtiaz

Issue #5136, opened 2020-11-28
[IDEA] Prevent double rendering when $:/core/ui/PageTemplate is open in the story


saqimtiaz:
It can be quite disconcerting when a user has `$:/core/ui/PageTemplate` open in the story, that the page appears to have a double rendering and in some cases the sidebar becomes unusable. In particular this is a barrier to users that want to edit and customize the page template in the browser via the TiddlyWiki UI:

See for example this report:
https://github.com/Jermolene/TiddlyWiki5/commit/6c98bb706a15b353a92895529a69bece11f613ff#commitcomment-44594409

We should consider hiding/preventing the rendering of the `$:/core/ui/PageTemplate` tiddler when open in the story. One option would be to add the field `hide-body` with a value of `yes`.

Jermolene:
Hi @saqimtiaz I've been thinking about that too. While the double sidebar issue is unique to the page template, viewing many core templates is pretty unhelpful at present.  One potential solution is a generic mechanism for marking tiddlers so that the view template displays them with a different template.

The "obvious" solution would be an optional "view-body-template" field that is used instead of the default body template. I don't like the way that it ties tiddlers directly to their template. An alternative might be a new "view-role" field that contains a token that is looked up via a prefix to obtain the title of the template. Then we might define roles for, say, templates and filters with special affordances for working with those data types.

saqimtiaz:
> An alternative might be a new "view-role" field that contains a token that is looked up via a prefix to obtain the title of the template. Then we might define roles for, say, templates and filters with special affordances for working with those data types.

I like this idea. It would allow for instance, filter tiddlers to be displayed with the filter visualization tool we have been discussing.

I'll pick this thread up again after 5.1.23.

joshuafontany:
This is kind of what I was aiming for with PR https://github.com/Jermolene/TiddlyWiki5/pull/4681

I don't think this is the way to go (but it was a fun experiment). Still, having a way to customize the view and edit templates without relyting on a field on each tiddler would be excellent to have available in the core.

Jermolene:
> This is kind of what I was aiming for with PR https://github.com/Jermolene/TiddlyWiki5/pull/4681
> 
> 
> 
> I don't think this is the way to go (but it was a fun experiment). Still, having a way to customize the view and edit templates without relying on the limited 'hide-text' field on each tiddler functionality would be excellent to have available in the core.

[...]

PR #6168, opened 2021-11-04
Add "cascade" filter run prefix and use it to control view templates

Jermolene:
This PR introduces a new "cascade" filter run prefix (see [here](https://github.com/Jermolene/TiddlyWiki5/issues/5136#issuecomment-735435268) for the discussion that introduced the idea) that implements a generic version of the logic we use to [customise file naming under Node.js](https://tiddlywiki.com/prerelease/#Customising%20Tiddler%20File%20Naming)

The filter run is evaluated to yield a list of filters. Each of the input titles is evaluated against each of the filters in turn, and the input title is replaced with first result of the first filter that returns a non-empty list (or the empty string if none of the filters returns a result).

```
[[Sparkling water]] :cascade[all[shadows+tiddlers]tag[$:/tags/Filter]get[text]]
```

This PR also updates the view template body template to use the cascade filter run prefix to choose the template to use for the body. It incorporates the functionality of the existing checks that suppress the body for plugin tiddlers, and those that have a "hide-body" field set to "yes". It also adds a new "code" template that is used for certain system tiddlers (eg those prefixed `$:/core/ui/`, finally making it safe to view core templates).

* [x] cascade filter run prefix
* [x] use cascade filter run prefix to choose view template body template
* [x] use cascade filter run prefix to choose view vs. edit template
* [ ] add control panel UI for inspecting the current template cascades
* [x] tests
* [ ] docs


saqimtiaz:
First impression is very positive. Was extremely easy to set up a custom body view template for a given tiddler.

pmario:
There is an update problem with the "prefixed code tiddlers" 

 - open [this view at tiddlywiki.com](file:///D:/Mario/Downloads/tiddlywiki(5).html#%24%3A%2Ftemp%2Fadvancedsearch:%24%3A%2Ftemp%2Fadvancedsearch%20%24%3A%2FAdvancedSearch) and type in any search tab. `$:/temp/advancedsearch  ` is updated.
 - open the same configuration in the PR wiki -> The "code area" isn't updated.

saqimtiaz:
> There is an update problem with the "prefixed code tiddlers" 
> 
> To reproduce:
> 
>  - open [this view at tiddlywiki-com](https://tiddlywiki.com/#%24%3A%2Ftemp%2Fadvancedsearch:%24%3A%2Ftemp%2Fadvancedsearch%20%24%3A%2FAdvancedSearch) and type in any search tab. `$:/temp/advancedsearch  ` is updated.
>  - open the same configuration in the PR wiki -> The "code area" isn't updated. 

The $codeblock widget doesn't refresh itself, the refresh method just returns false. That should be viewed as a separate issue.

pmario:
>Use the cascade mechanism to choose between the edit and view templates

@Jermolene ... Very clever! ... I do like the new mechanism. 

Do you know about the performance implications? .. I did some filter `$tw.perf.log()` tests yesterday ... to test for something else ... and I couldn't see any negative impacts .... 

BUT I'm not sure if we are able to measure the "cascaded" filters. ... I could only see the filter wich contains the `xxxx :cascade[]` in the table.

Jermolene:
> @Jermolene ... Very clever! ... I do like the new mechanism.

Thank you, I'm glad it's come out so neatly.

> Do you know about the performance implications? .. I did some filter `$tw.perf.log()` tests yesterday ... to test for something else ... and I couldn't see any negative impact ....

This is definitely the time for performance regression testing. I suspect that the impact will be minimal because (a) these are not expensive filters and (b) typically there are 10s of entries in the story river, not 100s or 1000s.

> BUT I'm not sure if we are able to measure the "cascaded" filters. ... I could only see the filter wich contains the `xxxx :cascade[]` in the table.

Ouch! That could be a bug, I'll try to confirm.

core/wiki/config/ViewTemplateBodyFilters.multids
```diff
@@ -0,0 +1,7 @@
+title: $:/config/ViewTemplateBodyFilters/
+tags: $:/tags/ViewTemplateBodyFilter
+
+system: [prefix[$:/boot/]] [prefix[$:/config/]] [prefix[$:/core/]!field:title[$:/core/readme]!field:title[$:/core/icon]] [prefix[$:/info/]] [prefix[$:/language/]] [prefix[$:/languages/]] [prefix[$:/snippets/]] [prefix[$:/state/]] [prefix[$:/status/]] [prefix[$:/info/]] [prefix[$:/temp/]] +[limit[1]then[$:/core/ui/ViewTemplate/body/code]]
+plugin: [has[plugin-type]then[$:/core/ui/ViewTemplate/body/blank]]
```
pmario, line 5: IMO instead of showing an empty body here, it would be better to show: `$:/core/ui/ViewTemplate/plugin` and remove the tag: `$:/tags/ViewTemplate` that tiddler. 

Since the plugin-type will win here, there is no need in the ViewTemplate to process it again. ... (I'm not sure, if it's clear what I mean)

pmario:
The `hide-body` field was introduced with PR: **change filter to stop automatic view transclusion if field 'hide-me' is ... #826** because of the reply from @Jermolene [at the TiddlyWiki Dev group](https://groups.google.com/g/tiddlywikidev/c/ejTKd3OfopA/m/EZrfYwBpHUIJ)

>The core does something similar for the custom display of plugin tiddlers, suppressing the "$:/core/ui/ViewTemplate/body" segment whenever there is a `plugin-type` field. Then the "$:/core/ui/ViewTemplate/plugin" segment does the actual custom display.
>
>So, one simple way to proceed would be to also hide the body segment if the tiddler has a field with the name `hide-body`:

So the `hide-body` field was "side-loaded" to the `plugin-type` mechanism, because the ViewTemplate already did contain the code for `plugin-type` special handling. ... BUT now imo hide body is obsolete. ... 

Because what we have here in the PR is a generic way to replace the `[[$:/core/ui/ViewTemplate/body/default]]` with something different. 

**Proposal: Let's deprecate the `hide-body` field and describe the mechanism how it is done right now**

core/ui/TiddlerTemplate.tid
```diff
@@ -0,0 +1,7 @@
+title: $:/core/ui/TiddlerTemplate
+
+<$let template={{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] }}}>
+
+<$transclude tiddler={{{ [<template>!is[blank]] ~[[$:/core/ui/ViewTemplate]] }}}/>
```
saqimtiaz, line 5: @Jermolene I am curious as to why you prefer the above construction over the following:
```
<$let template={{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] :and[!is[blank]else[$:/core/ui/ViewTemplate]] }}}>

<$transclude tiddler=<<template>>/>
```

Or simplifying even further:
```
<$transclude tiddler={{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] :and[!is[blank]else[$:/core/ui/ViewTemplate]] }}} />
```

If the `:cascade` prefix returned **no** result instead of an empty string when there is no match, the filter could be simplified to:
```
{{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] :else[[$:/core/ui/ViewTemplate]] }}}
```

pmario:
IMO The ViewTemlate can be made much cleaner now, since we can remove stuff that doesn't belong there.

grafik

core/ui/TiddlerTemplate.tid
Jermolene: Thanks @saqimtiaz, that's an oversight on my part.

Jermolene: (Apologies I pressed send too early)

> I am curious as to why you prefer the above construction over the following:

Thanks @saqimtiaz, that's an oversight on my part.

> If the `:cascade` prefix returned no result instead of an empty string when there is no match, the filter could be simplified to:

The rationale for returning a blank result is that the cascade filter prefix allows multiple input titles to be fed through the cascade in one go. If we returned no result for the input titles that fell through the cascade then it wouldn't be possible to line up the results with the input. Equally, the ability to cascade multiple values at once doesn't seem especially useful...

Jermolene:
I've added a simple UI in control panel for inspecting the template cascades:

image

Jermolene:
> IMO The ViewTemlate can be made much cleaner now, since we can remove stuff that doesn't belong there.

Thanks @pmario that makes sense.

core/ui/TiddlerTemplate.tid
saqimtiaz: > The rationale for returning a blank result is that the cascade filter prefix allows multiple input titles to be fed through the cascade in one go. If we returned no result for the input titles that fell through the cascade then it wouldn't be possible to line up the results with the input. Equally, the ability to cascade multiple values at once doesn't seem especially useful...

That makes sense, though I am also struggling to think of a use case for cascading multiple values. Users frequently get tripped up by blank values from filters so my thinking was to avoid it if possible. Then again, primitives frequently get put to uses we never could have imagined, so there may well be some utility to being able to pass multiple input titles to a cascade filter run.

Jermolene:
> IMO The ViewTemlate can be made much cleaner now, since we can remove stuff that doesn't belong there.

Thanks @pmario that's now done in e0b7dcded16211392473fa552947df7e7ae55a55

saqimtiaz:
We either need to update the filters in `TiddlerTemplateFilters.multids` and elsewhere where we specify the ViewTemplate as default to retrieve the View and Edit template from their respective config tiddlers (backwards compatible), or we need to deprecate these config tiddlers:
- $:/config/ui/ViewTemplate
- $:/config/ui/EditTemplate

Jermolene:
> We either need to update the filters in `TiddlerTemplateFilters.multids` and elsewhere where we specify the ViewTemplate as default to retrieve the View and Edit template from their respective config tiddlers (backwards compatible), or we need to deprecate these config tiddlers:

Thanks @saqimtiaz. I think we might as well continue to support those config tiddlers. Fixed in 2a3a6dd55820b758c3e16ba9a59a000f891cfba6

core/ui/TiddlerTemplate.tid
```diff
@@ -1,3 +1,3 @@
 title: $:/core/ui/TiddlerTemplate
 
-<$transclude tiddler={{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] :and[!is[blank]else[$:/core/ui/ViewTemplate]] }}} />
+<$transclude tiddler={{{ [<currentTiddler>] :cascade[all[shadows+tiddlers]tag[$:/tags/TiddlerTemplateFilter]get[text]] :and[!is[blank]else{$:/core/ui/ViewTemplate}] }}} />
```
saqimtiaz, line 3: typo:
{$:/config/ui/ViewTemplate}, not $:/core

Jermolene: Thanks @saqimtiaz, fix upcoming

Jermolene:
When we come to refactoring the import process into wikitext primitives, it might also be useful to use a filter cascade in the import process to determine what happens to incoming tiddlers. The cascade could generate standard actions like "select", "deselect", "block", or user defined actions. In either case, the actions would be stored in tiddlers like `$:/core/config/ImportActions/select` etc.

saqimtiaz:
> When we come to refactoring the import process into wikitext primitives, it might also be useful to use a filter cascade in the import process to determine what happens to incoming tiddlers.

Good idea. Makes perfect sense.

Jermolene:
As part of the docs, I've [added](a624ed24e28e2d022a93133fc6765f68282767a6) a demo of a custom story tiddler template:

image

saqimtiaz:
Excellent @Jermolene. 

Between this and the $let widget it almost feels worth a 5.2.1 release, were it not the amount of time and effort that goes into making a release. Very exciting stuff.

Jermolene:
> Between this and the $let widget it almost feels worth a 5.2.1 release, were it not the amount of time and effort that goes into making a release. Very exciting stuff.

I've been thinking about that too. Perhaps a release in the next couple of months would make sense. If we do make a release, we should take the opportunity to sort out the file locations in the `tw5.com` edition.

saqimtiaz:
> I've been thinking about that too. Perhaps a release in the next couple of months would make sense.

Right, and a spring 5.2.1 release will give us a longer window for the other things we want to work on.



> If we do make a release, we should take the opportunity to sort out the file locations in the `tw5.com` edition.

Yes please. A good place for us to start might be to see how much of the existing file structure can be mapped by tag based filter runs.

pmario:
One element that should be considered to be moved out from: $:/core/ui/EditTemplate/body ... the `_cannonical_uri` info element.

https://github.com/Jermolene/TiddlyWiki5/blob/d095aa0182faea4047dbc7de8789110f1bf84c6c/core/ui/EditTemplate/body.tid#L13-L25

Jermolene:
> One element that should be considered to be moved out from: $:/core/ui/EditTemplate/body ... the `_cannonical_uri` info element.
> 
> https://github.com/Jermolene/TiddlyWiki5/blob/d095aa0182faea4047dbc7de8789110f1bf84c6c/core/ui/EditTemplate/body.tid#L13-L25

Thanks @pmario I hadn't considered that, makes good sense.

pmario:
>Thanks @pmario I hadn't considered that, makes good sense.

I think, there are even more places in the view and edit-templates, where the new mechanism will make the templates simpler.

core/ui/EditTemplate/body/default.tid
```diff
@@ -0,0 +1,38 @@
...
+\define config-visibility-title()
+$:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
+\end
+
+\define importFileActions()
+<$action-popup $state=<<importState>> $coords="(0,0,0,0)" $floating="yes"/>
+\end
+
+<$set name="edit-preview-state" value={{{ [{$:/config/ShowEditPreview/PerTiddler}!match[yes]then[$:/state/showeditpreview]] :else[<qualify "$:/state/showeditpreview">] }}}>
+<$vars importTitle=<<qualify $:/ImportImage>> importState=<<qualify $:/state/ImportImage>> >
```
saqimtiaz, line 12: We can combine the above two lines into a single $let widget now.

Jermolene: I just lifted and shifted the current code to make the diffs easier to follow.

In terms of core usage of `<$let>` I wonder if this might be the time for a systematic review of the core templates.

Jermolene:
> I think, there are even more places in the view and edit-templates, where the new mechanism will make the templates simpler.

Indeed! In 88d8057b383b9addbe2d75fbbd90c71703003989 I've added a cascade for displaying the tiddler title in the view template. The motivation is to make it easier for people using non-human readable titles such as UUIDs to use a template that displays the caption field if it exists instead of the title field.

Suggestions welcome for other places that might benefit. We're looking for situations where the core chooses between two or more mutually exclusive templates. Typically it's currently done via `<$reveal>` or `<$list>` widgets, with no easy to recognise pattern.

saqimtiaz:
We have had some recent requests to be able to display a different view toolbar for some tiddlers based on a filter. We may want to make the toolbar configurable.

Jermolene:
> We have had some recent requests to be able to display a different view toolbar for some tiddlers based on a filter. We may want to make the toolbar configurable.

I saw that discussion, but it's hard to tell whether those requirements could be addressed with our current technique of making individual toolbar buttons be conditionally visible.

pmario:
>Suggestions welcome

AT: https://talk.tiddlywiki.org/t/tiddler-icon-by-tag/1486 Mohammad asks

>To show an image alongside with tiddler title, one way is to add the icon field with proper image in the icon field!
>
>Is it possible to show the same icon for all tiddlers tagged say with People using view template or other method and not set icon field for all those tiddlers? I am looking for a solution without touching core tiddlers!

I'm not sure if this would be a usecase. I'm also not sure, how I would try to solve it.

pmario:
Just an info ... no action needed ... 

I'm involved in this discussion at Talk: https://talk.tiddlywiki.org/t/is-there-any-resource-in-tiddlywiki-that-helps-me-to-create-a-multilanguage-tiddlywiki/1454/16 - where the request is to edit 2 languages side by side in 1 tiddler. Both language texts should be contained in 1 tiddler. ... 

I'm almost finished to implement a "custom EditTemplate" and ViewTemplate for a more flexible multi-line editor for user selected fields. I want to use it for a different purpose ... 

Once I'm finished with the prototype I'll try to implement the same thing with this PR active. So I can see how it works out. ...

Jermolene:
> At: https://talk.tiddlywiki.org/t/tiddler-icon-by-tag/1486 Mohammad asks
> 
> > To show an image alongside with tiddler title, one way is to add the icon field with proper image in the icon field!
> > Is it possible to show the same icon for all tiddlers tagged say with People using view template or other method and not set icon field for all those tiddlers? I am looking for a solution without touching core tiddlers!
> 
> I'm not sure if this would be a usecase. I'm also not sure, how I would try to solve it.

It's an interesting idea, and does seem like a good fit for the cascade mechanism. I'll have a look.

Jermolene:
> It's an interesting idea, and does seem like a good fit for the cascade mechanism. I'll have a look.

I've added a cascade for dynamically choosing tiddler icons here and it works pretty well: 5f213567dd898e113c645cfe63d5e8c19e4d1eae

@pmario it has involved some changes to the tags macros: previously the icon was passed as a parameter, but now it is chosen dynamically using the cascade. I retained the macro parameters for backwards compatibility in case callers are using positional parameters, but they are now ignored.

@kookma there's a demo here where icons are assigned to tiddlers tagged "Table of Contents" and "Working with TiddlyWiki" - https://github.com/Jermolene/TiddlyWiki5/pull/6168/files#diff-bc620a06b53af2deacbbc679b45d98fc8fad1c666e759eb0da6c95f779a024bb

kookma:
> @kookma there's a demo here where icons are assigned to tiddlers tagged "Table of Contents" and "Working with TiddlyWiki" - https://github.com/Jermolene/TiddlyWiki5/pull/6168/files#diff-bc620a06b53af2deacbbc679b45d98fc8fad1c666e759eb0da6c95f779a024bb

Many thanks @Jermolene!
I am working to re-implement parts of Magic Tabs and Tiddlywiki Scholar from Alberto Molino (http://tw5.scholars.tiddlyspot.com/), there any category uses an icon for example all tiddlers have tag `People`, or all tiddler tagged `Theory`, where I have to add icon field to any tiddler and this waste space!
Your solution now simplify the task and save the space!

Thank you

joshuafontany:
This is looking VERY exciting. Wrapping up some of the back-end stuff atm, and am looking forward to playing with the new cascade filters.

saqimtiaz:
I am in favour of merging this now to make it easier to test.

It would also be very helpful with the ongoing work to create path filters for the tw-com edition. To test those filters easily an up to date copy of tw-com is needed that also supports the :cascade prefix, which is a bit more involved to set up at present than it need be.

Jermolene:
> I am in favour of merging this now to make it easier to test.

I'll do so now.

The docs are the big outstanding issue, I'll get to them as soon as I can.

saqimtiaz:
Thank you @Jermolene

kookma:
Thank you @Jermolene 

I am gonna to use this in my in progress Edition Mehregan and the also with Thinkup plugin (Smart Tabs inspired from Alberto Molani)!
I love the Tiddler Icon!

One question: Can I use this logic to set color or other parameters through field? For example I like to use color red for all tiddler tagged Task! I use cascade to set a nice icon for all tasks!

Jermolene:
> One question: Can I use this logic to set color or other parameters through field? For example I like to use color red for all tiddler tagged Task! I use cascade to set a nice icon for all tasks!

I think we could introduce a new cascade to choose the tiddler colour, I'll have a look at it

Jermolene:
> I think we could introduce a new cascade to choose the tiddler colour, I'll have a look at it

That's now done in 122de63b71a2454314365ac396e051ad1db77641

pmario:
I think we need to rewrite this info too: ([tiddlywiki.com reference on layout hackability](https://tiddlywiki.com/#Page%20and%20tiddler%20layout%20customisation)) ... It was mentioned in a "talk" [discussion](https://talk.tiddlywiki.org/t/idea-addition-to-core-hacking-button-actions/1601/2). ...

AnthonyMuscio:
@Jermolene I have come late to this discussion from here [button actions](https://talk.tiddlywiki.org/t/idea-addition-to-core-hacking-button-actions/1601/4?u=tw_tones)

I am very excited about the value of the cascade, I just wanted to ask if a little additional hackability could be added for us designers (not developers).

> It also adds a new "code" template that is used for certain system tiddlers (eg those prefixed $:/core/ui/, finally making it safe to view core templates).

Could we provide a user configurable way to make use of the **"code" template** on any tiddler?

Perhaps a **tag/or field or type** to use the **"code" template** on any tiddler?

Jermolene:
> Could we provide a user configurable way to make use of the "code" template on any tiddler?

Thanks @AnthonyMuscio that's reasonable. I'll explore adding support for a field `code-body` to match the existing support for the `hide-body` field

pmario:
I think `hide-body` was made obsolete with the cascade filter system. Since basically any field (anything) can be used to change the cascade. ... So the question here is, how we name the term, that will be part of the core templates. 

The target still is, to add functionality to the core, without changing the core tiddlers. IMO a field would make sense, if it could have different values. But that's not the case since we can't change the logic in the core tiddler. It has to be "hardcoded" upfront. 

If it is a binary decision, imo a tag may be the best option as proposed by Tony. 

Just some thoughts.

Jermolene:
Hi @pmario we've kept support for `hide-body` for backwards compatibility, and it's still useful to be able to manipulate a tiddler without having to add a cascade – think of it a as a generic step in the filter cascade that can be used whenever it is sufficient. Crucially, it also lets users hide the body of their tiddlers before they get to the point of understanding filter cascades, which are after all a fairly complex feature.

The trouble with using a tag is that they are visible in view mode, hence the proposal to use a `code-body` field. We could certainly extend the proposal so that there were other meaningful values besides `yes`, but I'd be inclined to think that departing too far from the precedent of `hide-body` would be a mistake.

kookma:
One point:
With 5.2.0 and earlier when a user defined viewtemplate is in use the text it displays is normally smaller than the text is displayed by core viewtemplate for text field!

I wish to know does this issue addressed with cascade? (See also [Talk](https://talk.tiddlywiki.org/t/section-editor-plugin-first-stable-release/1060/144?u=mohammad) )

Jermolene:
Hi @kookma just to clarify, are you saying that text is displayed at a different size within a custom view template body that has been displayed via the cascade compared to when the same template is displayed via overwriting `$:/core/ui/ViewTemplate/body`?

pmario:
> ...we've kept support for hide-body for backwards compatibility

I know, that's not a problem. ... I just thought that adding a new one wouldn't be preferable. 

> ... The trouble with using a tag is that they are visible in view mode, hence the proposal to use a code-body field.

I think it's preferable to be able to see the tags. The tag is added to the "user-tiddler" that is handled in a "special way". Having the tag indicates, that it belongs to a set of elements, that are treated in a special way. 

Using fields won't indicate, that a tiddler is special. IMO it would be as if `$:/tags/Stylesheet` would be a field. Nobody would see, that the tiddler will have a big impact. To see it, it would be needed to open the tiddler or search with a special search string in "advanced search" ... 

I'm not sure if this would be an advantage.

kookma:
> Hi @kookma just to clarify, are you saying that text is displayed at a different size within a custom view template body that has been displayed via the cascade compared to when the same template is displayed via overwriting `$:/core/ui/ViewTemplate/body`?

No! I am saying in 5.2.9 or 5.1.23, the text produced by a custom viewtemplate like

```
<$list filter="[all[current]tag[task]]">
The due date is {{!!due-date}}
</$list>
```

is smaller the one is displayed by tiddler body (text field) itself!

Jermolene:
Apologies @kookma I'm still not following. Are these the two situations that you are comparing?

1. The fragment above as a custom view template segment (ie tagged `$:/tags/ViewTemplate`)
2. The fragment above inserted directly into a tiddler body

Has the behaviour changed with this PR?

You're probably aware, but the default body template wraps the body in a class so that we can apply the body font size (which is different than the page default).

kookma:
> Has the behaviour changed with this PR?
No! 
I will prepare a demo to show you what I mean!

Just the text produced by cusom viewtemplate is smaller by TW tiddler body (text field)

saqimtiaz:
@kookma note what Jeremy has said about wrapping the body with a class to control the font size. If the custom templates want the same font-size they need to use the class `.tc-tiddler-body`

This is not related to this PR.

kookma:
> You're probably aware, but the default body template wraps the body in a class so that we can apply the body font size (which is different than the page default).

I think this is the source of difference

@saqimtiaz 
> This is not related to this PR.
No, it is not! I just asked to see if cascade do anything with what is displayed through user defined templates (those tagged with `$:/tags/ViewTemplate`
Thank you for clarification!

AnthonyMuscio:
Folks, At UTC +11 sometimes I end up quite separate from the conversation.

Is there somewhere I can experiment with the cascade filter?

I have not fully grasped from the above conversation the syntax and what the filters would look like that would be applied with the cascade. 

However I believe discussions have only related to changing the template to alter the body view and edit templates. I was keen to see a way to use the elements currently tagged $:/tags/ViewTemplate if nothing else was available or not at all, according to conditions. If this were the case an alternate tiddler view template could use alternate title, toolbar and subtitle definitions, ie making the whole viewTemplate via each component having their own cascade.

I am grasping for good examples,
- But lets say I use the object-type field, that a tiddler of object-type html-page and it contained html and such tiddlers only displayed the title and close button (Alternate tool bar), no subtitle or no $:/tags/ViewTemplate additions.
- Or tiddlers of object-type journal had an alternative edit body template but otherwise was standard
- Or tiddlers of type dashboard has no subtitle or edit button
I hope this makes sense, because I am not sure I can yet make sense of the use of the cascade filter in the view edit template process, and I am not sure yet if this is introducing the degree of habitability I desire.

pmario:
>Folks, At UTC +11 sometimes I end up quite separate from the conversation.

Hi Tony, ... I think this post should be used to start a new Discussion here at GitHub. ... The main problem is we don't have a documentation yet, other than the changes that have been made in the core tiddlers at the moment. 

The "thing" is also very new to us / me and as you wrote it's "tricky" but I thing also elegant. ... I don't have a good idea atm, to explain the whole thing in an easy way. ... 

So the discussion will probably help us all to find a way to describe this function in a "power user friendly" way.

AnthonyMuscio:
Mario
> Hi Tony, ... I think this post should be used to start a new Discussion here at GitHub. ... The main problem is we don't have a documentation yet, other than the changes that have been made in the core tiddlers at the moment.
> 
> The "thing" is also very new to us / me and as you wrote it's "tricky" but I think also elegant. ... I don't have a good idea atm, to explain the whole thing in an easy way. ...

I will hold off since the next release is urgent, I just stumbled on the key bug myself. Once I have the cascade doc'o and the next release includes new ActionSetMultipleFieldsWidget, new SetMultipleVariablesWidget and Let all if which could be helpful.

I am all for the cascade filter but I feel the view (no edit cascade?) template changes are a bit rushed. I have a lot of changes I wanted to see in this hack-ability, and its taking to time to see if they will solve, just help or hinder. The only risk is they hinder because of the backward compatibility, ie these changes may make it difficult for future changes if we need to keep backward compatibility. I am not finding the current pre-release info user friendly, even if I may find my way.


**Is it wrong for me to ask?, Can we add the fix and features without yet changing the view/body and template until it is reviewed more thoroughly?** I feel there are possibly broader, easier to understand ways to change the View and edit templating system.

saqimtiaz:
> Is there somewhere I can experiment with the cascade filter?

The pre-release.

AnthonyMuscio:
Thanks saq, but as of writing, the **Cascade Filter Run Prefix** is not documented at all. I can see its there from Info > Advanced > cascades.

Since I was not privy to or able to read the JavaScript I can only try and reverse engineer.

saqimtiaz:
Your question was about where you can test it @AnthonyMuscio, not about documentation. For someone comfortable with wikitext and interested in following, there is no JavaScript knowledge required to understand from looking at the examples in the core. If that is challenging for you, please wait for documentation.

Documentation will be written in due course and before release. There are higher priorities at the moment with resolving the drag and drop bug.

AnthonyMuscio:
Ok, if you feel people can test things before basic syntax is available then I have to live with it. I am interested even concerned.

I suppose my point is perhaps the drag and drop bug, drives an update with the drag and drop fixed. End of story. Must the cascade (as yet undocumented) implemented and also drive the changes to view templates be implemented in a hurry with the drag and drop issue?

This is what I said before, and seemed to offend, about developers just going a little further (not much), but enough that super users can engage. Ideally contributing before it is "all done an dusted". The idea is with a little more info the rest of the community can go a long way and release developers from the larger community documentation process.

Please try and place yourself in my shoes, if you can, because from here it is very disempowering (or however you spell it). Regardless, I think it fair that I ask for as much info as possible is available, basically I like a challenge, but not when they are unnecessary.

saqimtiaz:
As a developer, I feel no obligation to provide documentation and hand holding to end users during the active development process itself. If end users wants to engage in the active development process, then the onus is on them to learn the requisite skills to do so. 

We are in a bit of a unique situation this time around as Jeremy sensibly put out the call for testing the pre-release earlier than we normally would, in light of the urgency of resolving the drag and drop bug.

End user documentation will be available before release - as it always is - and there will be a window to review and provide feedback and raise any concerns at that time.

AnthonyMuscio:
That is clear then. Your attitude is the problem.

Jermolene:
Hi @AnthonyMuscio I don't think there is really a problem here.

It has been pointed out that the documentation is in progress. That's unfortunate, as I've acknowledged, but the situation warrants this expedited bug fix release. I understand your concerns about locking in the cascade mechanism before everyone has had a chance to understand its implications.

It's not clear what you're asking for, given that we're already working on the documentation that you need. I don't think it would save time to try to further explain the mechanism here; the effort is better spent on writing the documentation.

Jermolene:
> Ok, if you feel people can test things before basic syntax is available then I have to live with it. I am interested even concerned.

The basic syntax of the cascade filter run prefix is given in the OP of this PR. Understanding the syntax requires an understanding of how filter run prefixes work but that's about all.

saqimtiaz:
I'll also point out the examples that have been posted on the forum by another end user, working from the information present in this and related threads:

https://talk.tiddlywiki.org/t/lets-have-some-fun-with-tiddlywiki-5-2-0-default-tiddler-icon/1595

Jermolene:
My original comment [here](https://github.com/Jermolene/TiddlyWiki5/issues/5136#issuecomment-735435268) from November 2020 might also help:

> I wondered about picking up a trick from the way that we [customise the naming of tiddler files under Node.js](https://tiddlywiki.com/prerelease/#Customising%20Tiddler%20File%20Naming), and having a cascade of filters that are applied in turn until one of them returns a non-empty result, which is then used as the template for the tiddler.

> We would need a new operator `mapfirst` that takes a list of filters and returns the first non-empty result. For example:

```
\define filter1() [tag[mytag]then[$:/my-tag-template]]
\define filter2() [tag[anothertag]then[$:/another-tag-template]]
<$text text={{{ [<filter1>] [<filter2>] +[mapfirst[My Tiddler Title]] }}}/>
```

> For the view template, the filter list would be driven by a tag such as `$:/tags/TemplateMapper`. The core would include some mappers for handling view vs edit mode:

```
[is[draft]then[$:/core/ui/EditTemplate]]
[[$:/core/ui/ViewTemplate]]
```

> With careful use of "list-before" and "list-after" fields, plugin authors should be able to add their own templates for their own types of tiddlers.

kookma:
> I'll also point out the examples that have been posted on the forum by another end user

Yes, the example by this end user is a good point to experiment ;-)

AnthonyMuscio:
@Jermolene thanks for a little extra info. I understand the source of the cascade operator was in an area of Node implementations customising **customise the naming of tiddler files under Node.js**, this is in itself something I, and the vast majority of users would not use. It would take some research on my part to understand those issues. I understood the cascade operator was written and placed in the pre-release. I await the documentation, with enthusiasm.

- This thread seems to be specifically about **using the cascade operator to control viewTemplates**, and Imminent in the the pre release 5.2.1.
- As someone who uses and has discussed repeatedly the need for improved template handling I have substantial interest in this area.
- The Pre-release goes so far as to implement modifications to the view template mechanism and using an undocumented operator.
- Given a need to expedite the bug fix in 5.2.1 "perhaps as soon as this weekend" it appears it's release will include the **implement modifications to the view template mechanism** (Is this true?)
- Thus as a friendly and concerned member of the community, with a particular interest in this specific subject area. I have practically sought to understand these "modifications to the view template mechanism" whilst in pre-release.

I suppose the question is "Must **I simply wait passively and receive what I am given** or may I **contribute to or test serious changes to TiddlyWiki?**" If I am permitted to contribute to and test **tiddlywiki futures** I would have thought **here and now was the time and place to discuss controlling viewTemplates.** otherwise please tell me how and where.

I continue to glean what I can from **reverse engineering the pre-release** to understand what is proposed and imminent, and with all due respect it reads as "meaningful to those who built it" and "not so easy and meaningful to those who use it". Clearly there is no formal, User Acceptance Testing UAT in TiddlyWiki, so at least an informal an open conversation between advanced users (at least, if not users) and developers is a good idea?

Jermolene:
Hi @AnthonyMuscio 

>  I understand the source of the cascade operator was in an area of Node implementations customising customise the naming of tiddler files under Node.js, this is in itself something I, and the vast majority of users would not use

It is not necessary to understand the cascade used for tiddler file naming under Node.js. My [comment above](https://github.com/Jermolene/TiddlyWiki5/pull/6168#issuecomment-977917290) is free standing.

In particular, do these words make sense:

> having a cascade of filters that are applied in turn until one of them returns a non-empty result, which is then used as the template for the tiddler.

And these examples:

```
[is[draft]then[$:/core/ui/EditTemplate]]
[[$:/core/ui/ViewTemplate]]
```

> using an undocumented operator.

That's not really a fair way to describe it. We're talking about work in progress, that is in the process of being documented.

> it appears it's release will include the implement modifications to the view template mechanism (Is this true?)

Yes. The cascade is in the prerelease for v5.2.1, and that is what we intend to release, subject to further last minute fixes and changes.

> I have practically sought to understand these "modifications to the view template mechanism" whilst in pre-release.

I appreciate that you've stated that you don't understand the new stuff, but equally I don't think you've asked any specific questions about it, so it's been very hard to help you.

> Must I simply wait passively and receive what I am given or may I contribute to or test serious changes to TiddlyWiki?

You, or any other member of the community, may actively contribute to and test serious changes to TiddlyWiki, admittedly perhaps subject to having the necessary skills and knowledge to do so. Indeed you've already contributed substantially, despite not being a developer.

> If I am permitted to contribute to and test tiddlywiki futures I would have thought here and now was the time and place to discuss controlling viewTemplates. otherwise please tell me how and where.

The only reason why we're not discussing interesting stuff like controlling ViewTemplates is because we've stepped back to discuss the meta issue of the difficulties you face participating in the detailed development process as a non-developer.

As I say, I think your understanding would progress better if you read the material and then asked specific questions. That's helpful for us because it lets us see the specific misunderstandings that we need to address in the docs.

> I continue to glean what I can from reverse engineering the pre-release to understand what is proposed and imminent, and with all due respect it reads as "meaningful to those who built it" and "not so easy and meaningful to those who use it". Clearly there is no formal, User Acceptance Testing UAT in TiddlyWiki, so at least an informal an open conversation between advanced users (at least, if not users) and developers is a good idea?

The development process is continuing as it always does: the developers listen to users (in this case the discussions about customising the view template), come up with a solution, discuss it via a GitHub PR, implement it in the prerelease, listen to feedback from users, and then finally issue it as part of an official release.

The only thing that's changed this time around is that we're accelerating the process, and I am spending all my time responding to your messages here instead of finishing the documentation :smile:

AnthonyMuscio:
@Jermolene thanks for the response, 

I still feel I have being misunderstood (but it would waste time arguing that) and I believe I had something specific of value to add, but I did not want to raise it until I had a deeper understanding of what was "proposed". I was not to know the capacity to help me contribute, was not there.

However I have already identified a way to better handle View/Edit template irrespective of these changes, so I am no longer so concerned. I will share later some additional control of the view/edit templates, and see how it fits then.

Beyond that I am unlikely to comment further on this because you are all busy.

Jermolene:
> Beyond that I am unlikely to comment further on this because you are all busy.

I am encouraging you not to give up on understanding the new stuff. There's not much too it, and you're fairly knowledgeable. It would be particularly helpful if you could ask specific questions to help us see where the misunderstandings have arisen.

pmario:
There is a new discussion: **How the new cascade mechanism works #6280** that tries to explain how the new core cascade definitions work. 

The next step will be some more background info with links to the wikitext source and more info.

AnthonyMuscio:
Thanks Jeremy;

It is simpler to leave the miss-undertandings in the past, but on the actual cascade on core UI I have;

> ask[ed] specific questions

On Mario's new discussion: **How the new cascade mechanism works** #6280 where I ask some questions of clarification and any gaps I perceive.

PR #6175, opened 2021-11-06
Update the tiddler "Simple ways to write protect tiddlers" to use data-tags

Telumire:
The tiddler [[Simple ways to write protect tiddlers]] mention the deprecated method explained in [[How to apply custom styles by tag]]. This PR update the tiddler to use the new data-tags method and add the tag "How to apply custom styles" to improve discoverability.


Jermolene:
Thanks @Telumire

PR #6232, opened 2021-11-17
tm-open-window documentation improvements

saqimtiaz:
* Removes references to paramObject
* Corrects some typos
* Removes em units for new window size from example

### DIFF
--- a/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-open-window.tid
+++ b/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-open-window.tid
@@ -1,11 +1,11 @@
 caption: tm-open-window
 created: 20160424181447704
-modified: 20211117042202771
+modified: 20211117212929052
 tags: Messages
 title: WidgetMessage: tm-open-window
 type: text/vnd.tiddlywiki
 
-The `tm-open-window` [[message|Message]] opens a tiddler in a new //browser// window. If no parameters are specified, the current tiddler is opened in a new window. Similiar to `tm-modal` any additional parameters passed via the <<.param "paramObject">> are provided as variables to the new window.
+The `tm-open-window` [[message|Messages]] opens a tiddler in a new //browser// window. If no parameters are specified, the current tiddler is opened in a new window. Similar to `tm-modal` any additional parameters passed  are provided as variables to the new window.
 
 |!Name |!Description |
 |param |Title of the tiddler to be opened in a new browser window, defaults to <<.var "currentTiddler">> if empty |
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-open-window.tid
```diff
@@ -1,19 +1,19 @@
...
+modified: 20211117212929052
 tags: Messages
 title: WidgetMessage: tm-open-window
 type: text/vnd.tiddlywiki
 
-The `tm-open-window` [[message|Message]] opens a tiddler in a new //browser// window. If no parameters are specified, the current tiddler is opened in a new window. Similiar to `tm-modal` any additional parameters passed via the <<.param "paramObject">> are provided as variables to the new window.
+The `tm-open-window` [[message|Messages]] opens a tiddler in a new //browser// window. If no parameters are specified, the current tiddler is opened in a new window. Similar to `tm-modal` any additional parameters passed  are provided as variables to the new window.
 
 |!Name |!Description |
 |param |Title of the tiddler to be opened in a new browser window, defaults to <<.var "currentTiddler">> if empty |
```
Jermolene, line 11: The problem here is that we need to properly distinguish between the default parameter and the named parameters.

Jermolene (commented):
Thanks @saqimtiaz

editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-open-window.tid
saqimtiaz: Do you have any thoughts on how to do that? The problem here - and I thought of this when writing this PR - is that we are not documenting attributes for a widget but rather a Message which is never directly accessed via wikitext. Yet we need to somehow document it for use with wikitext via ActionSendMessage.

It feels like we need to do away with the entire table and try a different approach which doesn't mention the actual message parameters at all but only how they would be specified via ActionSendMessage.

Jermolene: > is that we are not documenting attributes for a widget but rather a Message which is never directly accessed via wikitext. Yet we need to somehow document it for use with wikitext via ActionSendMessage.

Yes, they are parameters not attributes.

Perhaps we can cover them like this:

```
|!Parameter Name |!Description |
|//{default param}// |Title of the tiddler to be opened in a new browser window, defaults to <<.var "currentTiddler">> if empty |
|`template` |Template in which the tiddler will be rendered in |
|`windowTitle` |Title string for the opened window |
|`width` |Width of the new browser window |
|`height` |Height of the new browser window |
|//{any other params}// |Any further parameters are made available as variables within the new window |
```

Then we'd just need the examples.

CodaCodr: > |template |Template in which the tiddler will be rendered in |

```
|`template` |Template within which the tiddler will be rendered |
```

PR #6241, opened 2021-11-18
Whitespace trim for incredible wikitext uglifying

flibbles:
This is just three `\whitespace trim`. They're just an example for something greater.

Throughout the core, \whitespace trim is *kinda* used when it can be, but it really should be used more. Whenever it is, it allows the wikiparser to toss out a whole bunch of unnecessary textWidgets which contain empty air. So it's a memory and speed improvement.

However, I have an ulterior motive. I authored [TW5-Uglify](https://flibbles.github.io/tw5-uglify/), which uglifies javascript and css in tiddlywikis, but I'm taking it to the next level: **wikitext**. You might think there isn't much to shave off in wikitext, and you'd mostly be right, but it actually adds up. More importantly, my Uglifier *really* benefits from whitespace trim pragma. Look at this macro from "$:/snippets/peek-stylesheets" (the file this pull request is touching):

```html
\define stylesheets-list()
<ol>
<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!has[draft.of]]">
<li>
<$link>
<$view field="title"/>
</$link>
<$set name="source" tiddler=<<currentTiddler>>>
<$wikify name="styles" text=<<source>>>
<pre>
<code>
<$text text=<<styles>>/>
</code>
</pre>
</$wikify>
</$set>
</li>
</$list>
</ol>
\end
```
If you uglify this as is, you get this:
```html
\define stylesheets-list()
<ol>
<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!has[draft.of]]">
<li>
<$link>
<$view field=title/>
</$link>
<$set name=source tiddler=<<currentTiddler>>>
<$wikify name=styles text=<<source>>>
<pre>
<code>
<$text text=<<styles>>/>
</code>
</pre>
</$wikify>
</$set>
</li>
</$list>

\end
```
Hardly anything, right? **But** if it has a whitespace trim right under the \define, then you get *this*:
```html
\define stylesheets-list()<ol><$list filter="
[all[shadows+tiddlers]tag[$:/tags
/Stylesheet]!has[draft.of]]"><li><$link><$view 
field=title/></$link><$set name=source tiddler=
<<currentTiddler>>><$wikify name=styles 
text=<<source>>><pre><code><$text text=<<styles>>/>
```
(I added some linebreaks to make it more readable)

Take "$:/snippets/peek-stylesheets"
* without pragma, it goes from 2969 to 2630, or 12% removed
* with the pragma, it goes from 3038 to 2225, or 25% removed

@Jermolene, If it's all right with you, I'd like to go through some of $:/core and add \whitespace trim where it's appropriate. (I've just written about 300+ unit tests about where it's allowed, so I've got a good idea now.)

That said, this pull request does have a visual effect. Because there is hidden whitespace in `<pre>\n<code>\n<$text .../>\n</code>\n</pre>`, the gap under the stylesheet preview was this
gap
but now it's this, without the gap:
nogap

Which I think was actually intended. If I go through the core like this, I will be careful, and I'd discuss any visual changes in separate PRs. I know when "\whitespace trim" can't be safely used (mostly around macro placeholders). This really is a win,win,win, I'd say. Each whitespace trim only adds a measly 23 bytes to the file, and those can easily be canceled out with a few quotation mark removals.

Whaddaya say?

### DIFF
--- a/core/wiki/peek-stylesheets.tid
+++ b/core/wiki/peek-stylesheets.tid
@@ -1,6 +1,7 @@
 title: $:/snippets/peek-stylesheets
 
 \define expandable-stylesheets-list()
+\whitespace trim
 <ol>
 <$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!has[draft.of]]">
 <$vars state=<<qualify "$:/state/peek-stylesheets/open/">>>
@@ -38,6 +39,7 @@ title: $:/snippets/peek-stylesheets
 \end
 
 \define stylesheets-list()
+\whitespace trim
 <ol>
 <$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!has[draft.of]]">
 <li>
@@ -57,6 +59,7 @@ title: $:/snippets/peek-stylesheets
 </$list>
 </ol>
 \end
+\whitespace trim
 
 <$vars modeState=<<qualify "$:/state/peek-stylesheets/mode/">>>
 


Jermolene:
Hi @flibbles indeed, using `whitespace trim` more consistently across the core has been on my background list for a while, and it would be great to get it done.

> That said, this pull request does have a visual effect
> ...
> Which I think was actually intended

Yes I think that is an improvement in this case, but in general we should indeed be aiming for full backwards compatibility.

pmario:
@flibbles ... I would like to help if you need it.

>Hardly anything, right? But if it has a whitespace trim right under the \define, then you get this:

So If I understand it right, you use the `\whitespace trim` pragma as an indicator, that the code can be "uglyfied". righ? ... That's a nice idea. 

I do know a lot of UI code, where the source code would profit from prettifying, to improve the readability and the modify-ability. .. eg: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/ui/EditTemplate/fields.tid is the pure horror to modify. 

Since you already started to create unit tests, **I would be keen to see them**. They also can help with prettyfication, since they will show changes also "the other way around"

I think, the way you assign uglify is unique and allows us to have a "smaller core" and on the other hand it allows us to switch the compression off if we need "readability" back. ...

flibbles:
Thanks, @Jermolene. I'll get started. Just a heads up, to use '\whitespace trim' in some places, I may need to sprinkle in a few `&#32;` so substitute spaces where they are still needed, like around //emphasis// markers.

Also, I hope it's okay to reduce a few of `attr="""$these$"""` into `attr=<<__these_>>` They're less error prone, and they use 8 less bytes in the file, so it should help me stay byte neutral while adding these whitespace pragma. (Also, the uglifier hates them. Placeholders make it very difficult to safely work with the code.)

@pmario, sure! [Here's what I have so far](https://github.com/flibbles/tw5-uglify/tree/wikitext/tiddlers/Test/wikitext). You can also build the uglifier wizard yourself and test using `tiddlywiki --build wizard`.

Currently, I'm still writing rules and tests (changing $set to $let is still buggy right now), but the aim is that the uglifier makes sure the uglified text produces *identical* html when rendered. It can still make improvements if \whitespace trim isn't there (like simplifying attribute quotation), but needing to keep around those invisible whitespace character between tags really hampers its best techniques.

And yeah, I agree. Almost the entire reason I built uglify is so users can be more liberal with their comments and tab-indenting and know that people who like it *tight* can still have it so. I hope that's what you mean by "prettification", because I haven't been working on a pretty-print function.

pmario:
>Also, I hope it's okay to reduce a few of attr="""$these$""" into attr=<<__these_>> 

You have to be careful here. ... That's not the same thing and can change the logic.

`$xx$` is a real text substitution before "executing" the macro code. `<<__xx__>>` is a variable, like the macro itself.

pmario:
> I hope that's what you mean by "prettification", because I haven't been working on a pretty-print function.

Yes. .. There are several elements, that need better indentation for readability. .. I was thinking about manually changing them ;)

flibbles:
Right, and in text, such as `<$reveal>$myplaceholder</$reveal>`, the placeholder can control whether it's block or inline, but `<$reveal><<myplaceholder</$reveal>` would always be inline.

However, I believe as attribute values, (e.g. `<$set value="""$myplaceholder$""">`) *should* be safe, since it gets passed to the widget handler as a simple string either way.

Except you are still right. Injection-like values are still a think. If myplaceholder=`value""" anotherAttr="""anotherVal`, thus *hacking* the macro, then the output would be different. I suppose you're right then. Despite how this technique creates indecipherable lunacy, I should leave those placeholders alone.

pmario:
I was thinking about constructions like this, which was part of the old TOC macro

`<$vars .....  excluded="""$exclude$ -[[$tag$]]"""` Where we have 2 text substitutions but the second one will break if replaced with `<<__tag__>>` ..

In the TOC macro all the text substitutions have been replaced. ... BUT I needed to change the logic that it worked, because of the above construction. .

I'm not sure if constructions like that are still part of the core logic, but they may exist  in 3rd party libs.

pmario:
I didn't test `<$vars ..... excluded="[subfilter<__exclude__>] -[<__tag__>]` .. It may work ... but

The TOC uses `<$set name="excluded" filter="""[enlist<__exclude__>] [<__tag__>]""">` now.

Jermolene:
> Thanks, @Jermolene. I'll get started. Just a heads up, to use '\whitespace trim' in some places, I may need to sprinkle in a few `&#32;` to substitute spaces where they are still needed, like around //emphasis// markers.

That's fine. I think you'll find a few `<$text text=" "/>` that could be similarly subsituted.

> Also, I hope it's okay to reduce a few of `attr="""$these$"""` into `attr=<<__these_>>` They're less error prone, and they use 8 less bytes in the file, so it should help me stay byte neutral while adding these whitespace pragma. (Also, the uglifier hates them. Placeholders make it very difficult to safely work with the code.)

As @pmario says, extreme caution is needed here 😄 – your plan to make multiple smaller PRs will help here.

PR #6242, opened 2021-11-18
Changed vanilla/reset.tid to be a css file

flibbles:
Was there a reason this css file in the vanilla theme was flagged as text/plain? It was preventing plugins like tiddlywiki/highlight and flibbles/uglify from doing anything with it.


Jermolene:
Seems like a typo, thanks @flibbles

PR #6257, opened 2021-11-21
Trim Saga: First batch of \whitespace trim

flibbles:
These files aren't big fish, but I'm doing them first because they're special cases. They contain \whitespace trim already, but they're inconsistent about it, and they also use $placeholders$, which can make trimming tricky.

1. I added trim in a few places.
2. I could *not* add trim in "add-tag-actions", because it contains a $placeholder$ in text-space. In order to trim whitespace, I manually did it, and I did something like `><$list`. It works. I do it sometimes. If you think it's weird, just say and I'll take them out.
3. I changed `"""` to `'` in cases where there was a nested `"` for these reasons:
   1. The typography rules allow for it.
   2. It's much more space efficient, `"""this"""` takes up 22 more bytes than this `'this'` because of how much quotes get escaped in the Tiddlywiki file. So each one compensates for a new `\whitespace trim`
   3. I'm trying to make my uglifier shine. If there's even a single apostrophe in wikitext content, it knows it's free to use them whenever appropriate, and they are much more space efficient than the double-quote.

I'll keep these whitespace trims limited to a handful of files at a time so a git bisect can locate any problem if I should accidentally introduce one. I tested this batch, and everything seems fine.

Also, let me know if you'd like me to limit my changes to `\whitespace trim` additions only.


kookma:
>  I changed `"""` to `'` in cases where there was a nested `"` for these reasons:

I see cases where parameters like `actions="""<$action...  ` which seems to be replaced by `actions='<$action...  ` 

> I'm trying to make my uglifier shine.

**Sidenote**: I would suggest to use `minifier` instead of `uglifier`. This make more sense specially for TW users which most are non-technical people (in term of web tech)

core/wiki/macros/tag-picker.tid
```diff
@@ -8,27 +8,34 @@ second-search-filter: [tags[]is[system]search:title<userInput>sort[]]
...
+	name="tag"
+	value={{{ [<__tiddler__>get[text]] }}}
+><$list
+	filter="[<saveTiddler>!contains:$tagField$<tag>!match[]]"
+	variable="ignore"
+	emptyMessage='<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter="-[<tag>]"/>'
+><$action-listops
+	$tiddler=<<saveTiddler>>
+	$field=<<__tagField__>>
+	$subfilter="[<tag>]"/>
```
pmario, line 21: I personally don't like constructions like this. It's the same as having the "semicolon in js at the beginning of the line"
I think it should look like this:

```
<$set name="tag" value={{{ [<__tiddler__>get[text]] }}}>
	<$list filter="[<saveTiddler>!contains:$tagField$<tag>!match[]]" variable="ignore" 
	emptyMessage='<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter="-[<tag>]"/>'>
		<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter="[<tag>]"/>

```
and so on. The proposal here is as unreadable for humans as if there wheren't any indentation at all

flibbles: All right. Fair enough. I've reverted the strange formatting. ...and now I'm being bold.

I've traced that macro back to the $keyboard widget it gets fed into. That wikitext is ONLY used for action widgets, and I cannot see how trimming the whitespace could ever cause a backward compatibility issue. I've gone ahead and added the whitespace trim to it.

flibbles:
@kookma 

> I see cases where parameters like `actions="""<$action...` which seems to be replaced by `actions='<$action...`

I'm not sure what you mean by this. Do you mean you've seen this pattern elsewhere in tiddlywiki and thus it's okay? Or are you saying you're seeing something which seems problematic to you, and I should revert it?

> I would suggest to use minifier instead of uglifier.

Oh, it's much too late for that. I'd have to make a new github project and everything.

kookma:
> I'm not sure what you mean by this. Do you mean you've seen this pattern elsewhere in tiddlywiki and thus it's okay? Or are you saying you're seeing something which seems problematic to you, and I should revert it?

I just mean to use `'` where possible instead of `"""`! but there is a discussion on Talk (https://talk.tiddlywiki.org/t/good-practice-quotes-in-scripts-single-double-or-triple-double-quotes/1620) seems people are not agree! So, ignore this!

flibbles:
Now that V5.2.1 is out, should these be examined and merged?

Jermolene:
Hi @flibbles we're collecting post-v5.2.1 PRs in #6266, if you could kindly list them there?

flibbles:
If part of the problem with this PR is the coding decisions I made to the quotation, I can remove those. What's important is the \whitespace trims.

Jermolene:
Hi @flibbles thanks for your patience.

Issue #6176, opened 2021-11-06
[BUG] CSS parsing failure in one stylesheet can break the whole wiki


Telumire:
If a tiddler with the tag `$:/tags/Stylesheet` has a curly bracket left open, all the following stylesheets (in the order of the dropdown given by the tag) will fail too. If the tiddler has the field list-before blank, this break the whole wiki styling.

**To Reproduce**

1. Create a tiddler
2. Add the tag `$:/tags/Stylesheet`
3. Type `{`
4. Add the field list-before

**Expected behavior**
Tiddlywiki should close the brackets left open.

This was tested on the current release and on the pre-release, on firefox.

Jermolene:
Thanks @Telumire. I thought we already had a ticket for this problem but I couldn't find it.

I believe that the best fix is for each stylesheet tiddler to be rendered to an independent `<style>` element. Not only will it prevent parsing errors leaking between stylesheets, it should also improve stylesheet refresh performance because it will only have to refresh stylesheet tiddlers that change.

Telumire:
@Jermolene I couldnt find it either ! Thanks for taking a look at this. If it's simply a matter of adding a <style> tag in the template, I should be able to make a PR. Can I go ahead and do it ?

Jermolene:
Hi @Telumire there's a bit more to it than that.

The following code in `$:/core/modules/startup/render.js` is responsible for stylesheet rendering/refreshing:

https://github.com/Jermolene/TiddlyWiki5/blob/9caba544eb04f56ed772340a2d8767974b2617c9/core/modules/startup/render.js#L41-L52

That code needs to be updated to track each stylesheet tiddler separately, and to maintain an independent rendering in a separate `<style>` element. On each refresh cycle, we'd read the list of tiddlers with the tag `$:/tags/Stylesheet` and determine if there is an associated rendering. If so, that rendering would be refreshed. Otherwise, a new style element would be created, and tracked for next time.

saqimtiaz:
@Telumire there is a bit more to it than that:

https://github.com/Jermolene/TiddlyWiki5/blob/9caba544eb04f56ed772340a2d8767974b2617c9/core/modules/startup/render.js#L41-L52

Telumire:
Hi @Jermolene. Thanks for this explanation ! I see, I understand now why editing $:/core/ui/PageStylesheet didnt suffice.. I'm afraid it's too complex for me to resolve for now, I hope someone will figure something out !

[...]

PR #6268, opened 2021-11-23
Add a snippet to visualise faulty stylesheets

Telumire:
As discussed in https://github.com/Jermolene/TiddlyWiki5/issues/6176, a faulty stylesheet can propagate and break several other stylesheets. This can be tricky to debug in a wiki with a lot of custom stylesheets.

This wikitext snippet display a list of the stylesheets created/modified by the user and will display a red cross next to the one causing a parsing error (updated with 09851dfa00ed021d89753c6d0b5518b44872baa4): 

> image

The tiddler "Broken stylesheet" contains an unclosed "{" and thus is listed with a red cross next to it's name.

EDIT: Note that the red cross will show up following any css parsing error, not just unclosed "{" (see my comment bellow)
EDIT2: Updated the code to display some guidance on how to fix the error (09851dfa00ed021d89753c6d0b5518b44872baa4)

### DIFF
--- a/core/wiki/debugstylesheets.tid
+++ b/core/wiki/debugstylesheets.tid
@@ -1,0 +1,12 @@
+title: $:/snippets/DebugStylesheets
+
+Broken stylesheet will have a red cross next to their name :
+
+<style>
+[class*="debug"] {display: block;}
+[class*="debug"]:before{content:"❌";}
+</style>
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]has[modified]]" counter=n>
+<style>{{!!text}}.debug<<n>>:before{content:"✔";}</style>
+<$link class={{{[[debug]][<n>]+[join[]]}}}/>
+</$list>
--- a/core/wiki/debugstylesheets.tid
+++ b/core/wiki/debugstylesheets.tid
@@ -1,12 +1,15 @@
 title: $:/snippets/DebugStylesheets
 
-Broken stylesheet will have a red cross next to their name :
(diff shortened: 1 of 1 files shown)


twMat:
>"This stylesheet is broken" contains an unclosed "{" and thus is listed with a red cross next to it's name.

If that is the exclusive trigger for the message, then I'd propose that the messages states this fact. It is otherwise unclear what to do about it. E.g some message like "*Stylesheet is missing "{" or "}" character*"

Telumire:
@twMat the red cross will be displayed if there is a css parsing error, it is not limited to curly brackets.

Here's how it works : 
This code loop trough all the stylesheets modified by the user, transclude the content wrapped by <style> tags and apply a css rule to display a green check. If there is an error in a stylesheet that make the css parsing fail, the green check will not be applied to the link and the default (red cross) will be displayed instead. Since the CSS is enclosed by <style> tags, the parsing error does not propagate to the rest of the loop and thus allows to pinpoint where there is an error.

This is not perfect by any means since if there is a lot of stylesheets, this may get a bit laggy, and the user still has to figure out what is the error, but this is a start ..

twMat:
Very clever! I misunderstood it to be limited to "unclosed {" because you mentioned that specific example but you have clearly made it much more powerful.  :-)

Jermolene:
Thank you @Telumire

Jermolene:
I've subsequently moved the documentation out of the core and into the tw5.com edition, the motivation being to avoid English language strings embedded in the core.

PR #6269, opened 2021-11-23
Trim saga: The big macros

flibbles:
Next batch of \whitelist trim. Only five files, cause there was a surprisingly lot to do in those give.

I also converted a few cases of <$vars> and <$sets> into combined <$let>. All is tested.


Jermolene:
Thanks @flibbles

PR #6270, opened 2021-11-23
Trim Saga: Another batch of medium-sized files

flibbles:
I'd tag these for V5.2.2 myself if I could. Stop me if all this is a hassle.

Prettied up EditTemplate a bit since we can now.


core/wiki/macros/tree.tid
```diff
@@ -6,28 +6,31 @@ tags: $:/tags/Macro
 \end
 
 \define leaf-node(prefix,chunk)
+\whitespace trim
 <li>
 <$list filter="[<__prefix__>addsuffix<__chunk__>is[shadow]] [<__prefix__>addsuffix<__chunk__>is[tiddler]]" variable="full-title">
 <$list filter="[<full-title>removeprefix<__prefix__>]" variable="chunk">
-<span>{{$:/core/images/file}}</span> <$macrocall $name="leaf-link" full-title=<<full-title>> chunk=<<chunk>>/>
+<span>{{$:/core/images/file}}</span>&#32;<$macrocall $name="leaf-link" full-title=<<full-title>> chunk=<<chunk>>/>
```
pmario, line 13: There are some `tc-xxx-gap` classes, that will allow you to add CSS gaps left, right and center if needed. So "artificial" spaces are not needed. 

This gap here can be added to `<$link to=<<__full-title__>>><$text text=<<__chunk__>>/></$link>` in the `leaf-link` macro from line 4. Like so: `<$link to=<<__full-title__>> class="tc-tiny-gap-left"><$text text=<<__chunk__>>/></$link>`

The other elements may need an additional span for proper "spacing".

flibbles: Is that the actual design preference to use those classes rather than `&#32;`? Or, as I saw used all over the place, `&nbsp;`? My aim when creating these PRs was to reduce changes to the resulting parse tree as much as possible, so a few deliberately placed spaces was better for that than introducing classes.

Is there a final decision on this? [I was following Jeremolene's remarks from here](https://github.com/Jermolene/TiddlyWiki5/pull/6241#issuecomment-973120667), but I don't know if he remembered the classes at the time either. I hadn't known about them.

pmario: The classes are not very old. They have been created, because whitespace shouldn't be used for styling. CSS should be used for styling. ... 

IMO styled SPANs are semantically correct HTML code. Whitespace is _not_.

pmario: OK. You are right.

flibbles:
> Every PR that I create will cause a conflict for your PR or for mine, if yours is merged first.

I wouldn't let that stop you. I've touching nearly 20 files right now with my Trim stuff... and I don't know when or if any of it will merge in. Those 20 files can't be considered "locked". As long as you maintain the consistent use of "\whitespace trim", I can just resolve conflicts by merging in favor of your changes.

core/wiki/macros/tree.tid
Jermolene: I'm not convinced about the new spacing classes. It seems to me that using whitespace for certain purposes like separating controls from their text is perfectly reasonable because a unit of whitespace has a clear semantic meaning.

Using `&#32;` is clever, but existing code tends to use `<$text text=" "/>`. I've no objection to a blanket change, but again it would require care. In the meantime, we should try to be consistent, or at least not make the inconsistencies worse.

flibbles: @Jermolene, So no worries about `<$text text=" "/>`. They have all already been removed as a part of #5473. The only one that remains is on inside the "menubar" plugin, and [it is literally having no effect](https://github.com/Jermolene/TiddlyWiki5/blob/4f42df8befd9afe9ff7647f027cd10355bb7f0a2/plugins/tiddlywiki/menubar/menu.tid#L17) (It's not being used with \whitespace trim).

@pmario, I hadn't considered the copy-paste implications of those spaces. That does seem an argument for removing them in favor of spacing classes, but I still of the opinion that it should be a separate issue from the "\whitespace trim" additions.

pmario: @flibbles You are right. I'm OK with `&#32;` here. I don't think, there is a copy paste side effect here.

flibbles:
Are these waiting on anything? I see we're approaching another release cycle 5.2.3. Sooner or later, these whitespace fixes are going to go stale. I'd say either merge 'em or close 'em.

flibbles:
Ugh. That was a pain. Resolved a conflict on this. Required me to change my authtoken settings in github.

flibbles:
DOUBLE UGH. Someone wrote tests for that tabs file SINCE I wrote this PR.

Reconciled and ready to merge.

Jermolene:
Thanks @flibbles

PR #6274, opened 2021-11-24
Feature server cors

joshuafontany:
Allows a wiki served from one domain to communicate with a back-end Node.js server serving from a different domain. Uses a "cascade" to determine if the incoming request's Origin domain is Whitelisted.

TODO:

* Adapt to the new Cascade filter prefix model


core/modules/server/server.js
```diff
@@ -226,6 +227,29 @@ Server.prototype.isAuthorized = function(authorizationType,username) {
 	return principals.indexOf("(anon)") !== -1 || (username && (principals.indexOf("(authenticated)") !== -1 || principals.indexOf(username) !== -1));
 }
 
+Server.prototype.isOriginWhitelisted = function(origin) {
+	// Check if any of the originFilters applies
+	var self = this,
+		valid = (this.boot.origin == origin),
```
pmario, line 233: Is it possible to have `valid = (this.boot.origin === origin),`  here?

core/modules/server/server.js
```diff
@@ -226,6 +227,29 @@ Server.prototype.isAuthorized = function(authorizationType,username) {
 	return principals.indexOf("(anon)") !== -1 || (username && (principals.indexOf("(authenticated)") !== -1 || principals.indexOf(username) !== -1));
 }
 
+Server.prototype.isOriginWhitelisted = function(origin) {
```
Jermolene, line 230: I try to avoid "whitelist" and "blacklist" in favour of "accept list" and "deny list" or "block list"

core/modules/server/server.js
```diff
@@ -226,6 +227,29 @@ Server.prototype.isAuthorized = function(authorizationType,username) {
...
+	// Check if any of the originFilters applies
+	var self = this,
+		valid = (this.boot.origin == origin),
+		originFilters = this.wiki.getTiddlerList(this.get("cors-tiddler"),"text");
+	// Optionally output debug info
+	if(this.get("debug-level") !== "none") {
+		$tw.utils.log(`CORS=${valid?'valid':'invalid'} boot.origin=${this.boot.origin} request.origin=${origin}`)
```
Jermolene, line 237: We're still stuck on ES5.1 and so can't used templated string literals.

Jermolene (commented):
Great stuff thanks @joshuafontany.

It would be helpful if you could flag PRs as "draft" until they are ready to merge

core/modules/server/server.js
joshuafontany: Good tip, will do.

PR #6276, opened 2021-11-24
Trim Saga: Edit and View templates

flibbles:
Also, a couple files which were supposed to be in the last PR, but didn't make it, because I'm lost in a labyrinth of git branches and Tiddlywiki editions.

I might take a break here and wait for some merges/feedback. That way, I'm not perpetuating a mistake throughout my whole trimming saga.


saqimtiaz:
@Jermolene these might be worth a look since there are already changes to the templates for 5.2.1

Jermolene:
> @Jermolene these might be worth a look since there are already changes to the templates for 5.2.1

I've been resisting that logic; the counter argument is that every additional change in v5.2.1 increases the risk of bugs. @flibbles work seems well thought through and tested but I think there's still a fair risk as we move to more complex examples.

saqimtiaz:
@Jermolene fair enough. Personally I think it shouldnt be an issue if there are consecutive releases with template changes, though I know that worries some users. Hopefully the extra flexibility of the cascades should render that a moot point.

Jermolene:
> @Jermolene fair enough. Personally I think it shouldnt be an issue if there are consecutive releases with template changes, though I know that worries some users. Hopefully the extra flexibility of the cascades should render that a moot point.

At this point the only one that's been merged is #6241, which was a fairly limited test case.

flibbles:
ready to merge again...

Jermolene:
Thanks @flibbles I think we can get all these done now.

PR #6278, opened 2021-11-25
Fix for CodeMirror issues with drop events in Chrome 96

### LINKED ISSUE #6273: [BUG] Urgent: Dragging and dropping tiddler titles or links into CodeMirror does not work as expected.
As a result of the same underlying changes in Chrome 96 that caused #6254, dragging tiddler links or external links into the CodeMirror editor does not insert the tiddler title or link URL as before. Instead we get text of the form:
```
[InternetShortcut]
URL=https://github.com/saqimtiaz/sq-tw
```

The way the CodeMirror plugin has always handled drop events is to let CodeMirror library handle it, which by default has always worked for us. Now however it is the underlying CodeMirror library handling itself that is not equipped to deal with this change.

Thoughts on how to proceed @Jermolene @pmario ?

Potential options:
- One option is for the CodeMirror plugin to take over the drop handling entirely, i.e. intercept the drop event, parse the text to insert and insert it at the correct position. However this may be significant work with a lot of edge cases to consider.
- We could also try catching the drop event and sending another drop event with a different dataTransfer payload, but that too will require some investigation as to how feasible it is.
- A cursory look does not show any options in the CodeMirror API to customize the drop handling but this might be worth looking into further.

This issue would need to be resolved before we ship TW v5.2.1

saqimtiaz:
This PR resolves the issues with CodeMirror's handling of drop events in Chrome 96 by handling the drop ourselves in those situations. Fixes #6273 

Note that there is a CodeMirror helper method copied verbatim, not sure if we would want to tweak that to abide by our coding style.

While the new handling should only apply when Chrome is doing its new thing with drop events, it is best to be safe and do some cross browser and OS testing: https://saqimtiaz.github.io/sq-tw/temp/cm-drop-patch.html

Things to test in CodeMirror - best to do this in a tiddler with some existing text:
- dragging and dropping tiddler links
- dragging and dropping external links
- dragging and dropping text
- whether the links or text are inserted at the correct place in the text
- dragging and dropping an image and importing it.

### DIFF
--- a/plugins/tiddlywiki/codemirror/engine.js
+++ b/plugins/tiddlywiki/codemirror/engine.js
@@ -130,6 +130,58 @@ function CodeMirrorEngine(options) {
 		if(!self.widget.isFileDropEnabled) {
 			event.stopPropagation(); // Otherwise TW's dropzone widget sees the drop event
 		}
+		// Detect if Chrome has added a pseudo File object to the dataTransfer
+		if(!$tw.utils.dragEventContainsFiles(event) && event.dataTransfer.files.length) {
+			//Make codemirror ignore the event as well will handle the drop ourselves
+			event.codemirrorIgnore = true;
+			event.preventDefault();
+
+			// from https://github.com/codemirror/CodeMirror/blob/master/src/measurement/position_measurement.js#L673
+			function posFromMouse(cm, e, liberal, forRect) {
+				let display = cm.display
+				if (!liberal && e_target(e).getAttribute("cm-not-content") == "true") return null
+
+				let x, y, space = display.lineSpace.getBoundingClientRect()
+				// Fails unpredictably on IE[67] when mouse is dragged around quickly.
+				try { x = e.clientX - space.left; y = e.clientY - space.top }
+				catch (e) { return null }
+				let coords = cm.coordsChar(cm, x, y), line
+				if (forRect && coords.xRel > 0 && (line = cm.getLine(cm.doc, coords.line).text).length == coords.ch) {
+					let colDiff = window.CodeMirror.countColumn(line, line.length, cm.options.tabSize) - line.length
+					coords = window.CodeMirror.Pos(coords.line, Math.max(0, Math.round((x - paddingH(cm.display).left) / charWidth(cm.display)) - colDiff))
+				}
+				return coords
+			}
+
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@Jermolene I would welcome your thoughts when you get the chance to review this approach. We could mark with the comments - on encapsulate in a new method - the extra code and eventually remove if the issue ever gets resolved in CodeMirror.

Jermolene:
> @Jermolene I would welcome your thoughts when you get the chance to review this approach. We could mark with the comments - on encapsulate in a new method - the extra code and eventually remove if the issue ever gets resolved in CodeMirror.

Marvellous thank you @saqimtiaz, much cleaner than I was expecting.

plugins/tiddlywiki/codemirror/engine.js
```diff
@@ -130,6 +130,58 @@ function CodeMirrorEngine(options) {
...
 			event.stopPropagation(); // Otherwise TW's dropzone widget sees the drop event
 		}
+		// Detect if Chrome has added a pseudo File object to the dataTransfer
+		if(!$tw.utils.dragEventContainsFiles(event) && event.dataTransfer.files.length) {
+			//Make codemirror ignore the event as we will handle the drop ourselves
+			event.codemirrorIgnore = true;
+			event.preventDefault();
+
+			// from https://github.com/codemirror/CodeMirror/blob/master/src/measurement/position_measurement.js#L673
+			function posFromMouse(cm, e, liberal, forRect) {
```
saqimtiaz, line 140: @Jermolene regarding the `posFromMouse()` inline method copied verbatim from CodeMirror, should it be kept as is or brought in line with our coding standards? It uses `let` and `destructuring assignments`, and semi-colons are nowhere to be seen.

Jermolene: I guess if this code is already in the CodeMirror plugin for v5.2.0 it won't cause any further compatibility problems, so I think we should leave as is.

saqimtiaz:
@Jermolene this should be good to go. Recommend merging soon so we can allow a day or two for testing in the pre-release, though I do not anticipate any problems.

Jermolene:
Thanks @saqimtiaz!

PR #6281, opened 2021-11-25
Fix duplicate example id for map filter run prefix

btheado:
The first parameter of the operator examples macro is used for constructing unique state tiddler titles. Two of the `:map` filter prefix examples had the same id, causing them to share state with each other. Chose a different id for one of them.


btheado:
This one I intentionally created the PR against the master branch :-). The corresponding code is new and not on the doc branch.

Jermolene:
Thanks @btheado, very attentive!

PR #6282, opened 2021-11-25
Added examples to compare :map filter run prefix to + prefix

btheado:
The functionality of the `:map` filter run prefix has some overlap with the `+` prefix. Add side-by-side examples showing cases when the results are the same and when they are different.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
+++ b/editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
@@ -1,5 +1,5 @@
 created: 20210618134753828
-modified: 20210618140945870
+modified: 20211125152755859
 tags: [[Filter Syntax]] [[Filter Run Prefix Examples]] [[Map Filter Run Prefix]]
 title: Map Filter Run Prefix (Examples)
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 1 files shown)


btheado:
If this PR passes muster, no hurry to merge it. The same tiddler is modified in the master branch and there will be a merge conflict when master is merged into the docs branch. If you want to wait until after 5.2.1, that's fine with me and I can take care of the conflict prior to PR merge.

editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
```diff
@@ -14,3 +14,28 @@ Replace the input titles with the caption field if it exists, otherwise preserve
 For each title in a shopping list, calculate the total cost of purchasing each item:
 
 <<.operator-example 2 "[tag[shopping]] :map[get[quantity]else[0]multiply{!!price}]">>
+
+!! Comparison between `:map` and `:and`/`+` filter run prefixes
```
saqimtiaz, line 18: We should consider the `+` name deprecated. I think it is enough to mention once in this tiddler that they are the same and then always use `:and` in both the descriptions and the examples.

saqimtiaz:
Excellent, thank you @btheado. You work on the documentation is greatly appreciated.

editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
btheado: Sure, I made that change.

Jermolene:
Thanks @btheado

PR #6283, opened 2021-11-26
Fix tiddler colour filters default

pmario:
There's probably a copy paste issue.   The current default filter doesn't return the value stored in `$:/config/DefaultTiddlerColour` 
This PR fixes this


Jermolene:
Thanks @pmario that is indeed a typo

PR #6285, opened 2021-11-26
Add NavigatorWidget example

btheado:
Add an example using the NavigatorWidget and ListWidget to embed a storyriver inside a tiddler. A search box is provided for an easy way to get links to click. One hard-coded link is also provided when the storyriver is empty.

### DIFF
--- a/editions/tw5.com/tiddlers/howtos/Creating SubStories.tid
+++ b/editions/tw5.com/tiddlers/howtos/Creating SubStories.tid
@@ -1,44 +1,32 @@
 created: 20140214215718317
-modified: 20140919161514318
+modified: 20211127032217828
 tags: Learning
 title: Creating SubStories
 type: text/vnd.tiddlywiki
 
-This example shows how to create a sub-story within a tiddler that is independent of the main story.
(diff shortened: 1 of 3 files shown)


kookma:
@btheado - very nice and useful example! Thank you indeed!

There is one point, if you change the storyview to zoomin, the two line description below searchbox disappeared.
I think the issue is not with your solution, but with zoomin!

Thank you again!

Jermolene:
Thanks @btheado we already have the tiddler https://tiddlywiki.com/#Creating%20SubStories, perhaps we should improve that, and link to it from the navigator widget docs?

btheado:
@kookma, that is strange behavior. I think I'll hard-code the example to have storyview=classic. I think for the embedded case classic makes the most sense.

@Jermolene, I wasn't aware of that tiddler. I'll do as you suggest.

btheado:
I moved the example to the Creating Substories tiddler (replacing the example there). I hard-coded the storyview to classic. I also added more description and links.

Jermolene:
Thanks @btheado

PR #6293, opened 2021-11-30
add "some" flag to search operator

### LINKED ISSUE #6289: [Proposal] Add a new flag to the "search operator" that returns something if "any" token matches
It's proposed to add a new flag to the search operator, that allows us to show _more_ results instead of _less_ as it is atm.

Eg: If I search for `[search:tags[one two]]` at the moment, the default behaviour needs to match both tokens to return an element

The proposal is to have a new flag eg: "any" that will return a result if at least **one** token matches. 

It would look like this: `[search:tags:any[one two]]` .......... I'm not sure if `any` is the best flag name .. It may be confused with search in "any field" if used without a field flag ... 

It will be possible to combine it with `anchored` and `casesensitive`

The implementation is simple. It's just the name of the flag and the docs that need to be done. I've done the code and tests already :)

What do you think?

pmario:
As proposed in #6289 "[Proposal] Add a new flag to the "search operator" that returns something if "some" token match" 

 - This PR implements the discussed feature. 
 - Tests are included
 - Docs has been changed
 - There are some more examples. 

This PR closes #6289

Link to a demo will follow soon

### DIFF
--- a/core/modules/filters/search.js
+++ b/core/modules/filters/search.js
@@ -40,6 +40,7 @@ exports.search = function(source,operator,options) {
 			invert: invert,
 			field: fields,
 			excludeField: excludeFields,
+			some: hasFlag("some"),
 			caseSensitive: hasFlag("casesensitive"),
 			literal: hasFlag("literal"),
 			whitespace: hasFlag("whitespace"),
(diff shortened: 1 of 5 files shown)


Jermolene:
Thanks @pmario

PR #6295, opened 2021-12-01
charcode-jsstring-macros-charcode-operator

pmario:
Link to the demo wiki with some info and examples will follow soon. 

There is no docs and no tests atm. 

FORGET about `morph-string` it only contains some code snippets.

This is a draft PR to experiment with the possibilities. The following lookup table is for experimenting atm, which is exposed as `$tw.config.lookup` it should probably be splitted into 2 parts. 1 for `charcode macro` and one for `jsstring` macro

```
// Character code lookups used in charcode filter and macro
exports.lookup = {
	"tab": 9,
	"\\t": 9,
	"lf": 10,
	"\\n": 10,
	"ff": 12,
	"\\f": 12,
	"cr": 13,
	"\\r": 13,
	"\\_": 32,
	"space": 32
}
```

### DIFF
--- a/core/modules/config.js
+++ b/core/modules/config.js
@@ -38,4 +38,19 @@ exports.htmlBlockElements = "address,article,aside,audio,blockquote,canvas,dd,de
 
 exports.htmlUnsafeElements = "script".split(",");
 
+
+// Character code lookups used in charcode filter and macro
+exports.lookup = {
+	"tab": 9,
+	"\\t": 9,
+	"lf": 10,
+	"\\n": 10,
+	"ff": 12,
+	"\\f": 12,
+	"cr": 13,
+	"\\r": 13,
+	"\\_": 32,
+	"space": 32
+}
+
 })();
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -175,10 +175,12 @@ exports.pad = function(source,operator,options) {
 }
 
 exports.charcode = function(source,operator,options) {
-	var chars = [];
+	var lookup = $tw.config.lookup,
+		chars = [];
 	$tw.utils.each(operator.operands,function(operand) {
-		if(operand !== "") {
-			chars.push(String.fromCharCode($tw.utils.parseInt(operand)));
+		var code = lookup[operand.toLowerCase()] || operand;
+		if(code !== "") {
+			chars.push(String.fromCharCode($tw.utils.parseInt(code)));
 		}
 	});
 	return [chars.join("")];
--- a/core/modules/macros/charcode.js
+++ b/core/modules/macros/charcode.js
@@ -1,0 +1,37 @@
+/*\
+title: $:/core/modules/macros/charcode.js
+type: application/javascript
+module-type: macro
+
+Macro to return a character using the character-code as input.
+If the code isn't found the character is returned.
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.name = "charcode";
+
+exports.params = [
+	{name: "codes"},
+];
+
+exports.run = function(codes) {
+	var lookup = $tw.config.lookup,
+		tokens = codes.trim().split(/\s+/),
+		results = [];
+	tokens.map(function(name) {
+		var code = lookup[name.toLowerCase()];
+		if (code) {
+			results.push(String.fromCharCode(code));
+		} else {
+			results.push(String.fromCharCode($tw.utils.parseInt(name)));
+		}
+	})
+	return results.join("");
+}
+
+})();
--- a/core/modules/macros/jsstring.js
+++ b/core/modules/macros/jsstring.js
@@ -1,0 +1,36 @@
+/*\
+title: $:/core/modules/macros/jsstring.js
+type: application/javascript
+module-type: macro
+
+Macro to returns a string and replaces \r, \n, \t strings with their character codes.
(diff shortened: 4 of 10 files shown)


pmario:
.

pmario:
@AnthonyMuscio 

I did update the demo-wiki: https://pmario.github.io/kitchensink/6295-charcode-jsstring-morph-macros.html

What do you think about the "morph" feature? ... (short answer pls ;)

AnthonyMuscio:
On Morph
As I understand it, With Morph the "from" relates to the staring position and the "to" first character in second the second set. 
So this is an an example that depends on the characters being contiguous sets, if that what one wants. This is very easy to use.
- I love the prefix feature to output the numbers (the reverse being the charcode operator?), I would love to test each way.
- Also this works as well for normal characters `<<morph "upper case" from:"a" to:"A">>` but perhaps that is what you mean.

On lookup tables and Morph
- If morph was to also handle the Lookup tables, and only if, perhaps _start_ and _offset_ should be used, so _to_ and _from_ could store the two lookup tables
- I presume we will consider the "lookup table separately"

When you say
>The to parameter should always be inserted as a character, to see if the OS / App can handle the character. 

I assume here you are only referring to a visual confirmation that it is being rendered? 
It would be harder to do a true test, however we could at least test for a (web)font we know that can display all or most Unicode characters is set in $:/themes/tiddlywiki/vanilla/settings/fontfamily

On Morph
As I understand it, With Morph the "from" relates to the staring position and the "to" first character in second the second set. 
So this is an an example that depends on the characters being contiguous sets, if that what one wants. This is very easy to use.
- I love the prefix feature to output the numbers (the reverse being the charcode operator?), I would love to test each way.
- Also this works as well for normal characters `<<morph "upper case" from:"a" to:"A">>` but perhaps that is what you mean.

On lookup tables and Morph
- If morph was to also handle the Lookup tables, and only if, perhaps _start_ and _offset_ should be used, so _to_ and _from_ could store the two lookup tables
- I presume we will consider the "lookup table separately"

When you say
>The to parameter should always be inserted as a character, to see if the OS / App can handle the character. 

I assume here you are only referring to a visual confirmation that it is being rendered? 
It would be harder to do a true test, however we could at least test for a (web)font we know that can display all or most Unicode characters is set in $:/themes/tiddlywiki/vanilla/settings/fontfamily

On JSString/Charcode 

>The lookup for the "placeholders" looks like this atm

Can this list be extended by the user?, or do we need to decide what is included to publish this?

AnthonyMuscio:
On Morph
As I understand it, With Morph the "from" relates to the staring position and the "to" first character in second the second set. 
So this is an an example that depends on the characters being contiguous sets, if that what one wants. This is very easy to use.
- I love the prefix feature to output the numbers (the reverse being the charcode operator?), I would love to test each way.
- Also this works as well for normal characters `<<morph "upper case" from:"a" to:"A">>` but perhaps that is what you mean.

On lookup tables and Morph
- If morph was to also handle the Lookup tables, and only if, perhaps _start_ and _offset_ should be used, so _to_ and _from_ could store the two lookup tables
- I presume we will consider the "lookup table separately"

When you say
>The to parameter should always be inserted as a character, to see if the OS / App can handle the character. 

I assume here you are only referring to a visual confirmation that it is being rendered? 
It would be harder to do a true test, however we could at least test for a (web)font we know that can display all or most Unicode characters is set in $:/themes/tiddlywiki/vanilla/settings/fontfamily

AnthonyMuscio:
Brought from other thread to here (just for the record)

In fact what I propose here, a variation of the morph macro by Mario, or the "recode macro" may actually be the best way to implement named character codes just have a list of character codes (the numeric representations, allowing multiples) then a separate string (code set) to map to. So `<<recode crlf   rom-set:"codename" to-set:"charcode">>`

I understood the morph does not work unless the two character sets are contiguous, and clearly some 0-9a-Z sets are not but that is OK, **some are**. It could even be used to encode text in a simple way, such as to conceal an answer.

Without using morph, or adding a new feature to it we can simply supply two strings (code set) and test for the matching position, thus making such non-contiguous sets, into contiguous ones.

math-0a-Z "𝟎𝟏𝟐𝟑𝟒𝟓𝟔𝟕𝟖𝟗𝐚𝐛𝐜𝐝𝐞𝐟𝐠𝐡𝐢𝐣𝐤𝐥𝐦𝐧𝐨𝐩𝐪𝐫𝐬𝐭𝐮𝐯𝐰𝐱𝐲𝐳𝐀𝐁𝐂𝐃𝐄𝐅𝐆𝐇𝐈𝐉𝐊𝐋𝐌𝐍𝐎𝐏𝐐𝐑𝐒𝐓𝐔𝐕𝐖𝐗𝐘𝐙"
0a-Z "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"
 
A macro (ideally JavaScript for performance) that would allow you to convert a string based on one set to another using such predefined sets would be more than adequate `<<recode "My string" from-set:"0a-Z" to-set:"math-0a-Z">>`

- However if we could also define a set based on a Unicode start position, and total items ALSO we need not even type such sets.
- We could also generate random character sets, and use them for simple concealment or encryption

I have succeeded in the past using split[] and using the position in a set select to get the matching character. If carefully designed rather than single character mappings we could also map one set of values OR titles to another set similar to the new $names and $values parameters in ActionSetMultipleFieldsWidget, new SetMultipleVariablesWidget and extended ActionSendMessageWidget

PR #6296, opened 2021-12-02
Feature server cors (postponed)

joshuafontany:
Redid this PR, as something went wrong with the last one.

This allows the Node.js TW server to respond to http(s) requests from OTHER Origins, if they include a valid 'Origin' header.

Valid 'Origin' headers are determined by a `:cascade` of all filters in tiddlers tagged `$:/tags/CorsFilter`.


joshuafontany:
HRM... This is unfortunate, but these options are completely bugging out when the node.js wiki is served from a local computer, and accessed from the local network with Basic-Authentication. Trying to allow all the options that a local user serving a wiki may set is not handled easily.

For example, set the host to `0.0.0.0` and the `origin` to the local IP address plus port 8080. Then try to access from the local machine via `http://127.0.0.1:8080` or `http://localhost:8080` instead of the computer's IP. The initial document fetch always succeeds, but then the lack of Authentication headers breaks everything.

As this is currently (probably) the most commonly used setup (local user, local machine, local node.js server), I feel this will break user-expectations. May come back to this.

joshuafontany:
Research for later:
https://portswigger.net/web-security/cors/access-control-allow-origin
https://portswigger.net/web-security/cors#server-generated-acao-header-from-client-specified-origin-header

Jermolene:
Thanks @joshuafontany – that makes sense. One could imagine a server config switch for selecting between personal/intranet mode and internet-exposed mode.

PR #6299, opened 2021-12-03
Remove the possible pair of extra curly brackets around the title in the imported bib file.

Gk0Wk:
I use Mendeley to manage papers. When I imported the Mendeley exported .bib file into TW, I noticed that the `title` field was enclosed in curly brackets, like this.

> **Title** {State-of-the-art in Automatic 3D Reconstruction of Structured Indoor Environments}

After observing the .bib file I found that the title field in the .bib file is surrounded by two pairs of curly brackets (see `example.bib.txt` for details. The GitHub does not allow to upload 'bib' files, so I renamed it), and the plugin only removes the outermost curly brackets, so it leaves the curly brackets inside.

Considering that Mendeley is also a popular documentation management software, and that curly brackets `{}` are non-displayable characters used to define the scope in TeX syntax, I think all the extra curly brackets should be removed. At the moment I only remove the possible curly brackets from the title field, but maybe this should be done for all fields as well?

[example.bib.txt](https://github.com/Jermolene/TiddlyWiki5/files/7646496/example.bib.txt)

### DIFF
--- a/languages/pl-PL/Exporters.multids
+++ b/languages/pl-PL/Exporters.multids
@@ -3,5 +3,5 @@ title: $:/language/Exporters/
 StaticRiver: Statyczny HTML
 JsonFile: Plik JSON
 CsvFile: Plik CSV
-TidFile: Plik ".tid"
+TidFile: Plik tekstowy TID
 
--- a/languages/pl-PL/Help/savewikifolder.tid
+++ b/languages/pl-PL/Help/savewikifolder.tid
@@ -19,6 +19,8 @@ Wspierane argumenty:
 ** `yes` rozdzieli wtyczki na osobne pliki tiddlerów i zapisze je do podfolderu z wtyczkami
 ** `no` każda wtyczka będzie zapisana jako jeden zbiorczy plik w formacie JSON w folderze z tiddlerami
 
+Obie wartości dla `explodePlugins` stworzą taką samą wiki. Różnica będzie jedynie w sposobie rozlokowania wtyczek.
+
 Typowe zastosowanie to konwersja pliku TiddlyWiki w formie pliku HTML do formatu folderu:
 
 ```


pmario:
Is there a standard spec for .bib files?

 At the moment I think the PR shouldn't be merged, since it seems to fix a user specific problem and isn't a generic fix. ... We need to know about the .bib rules and specification to see, if it is a Mendeley export problem and _not_ TW import problem. ...

Gk0Wk:
@pmario I read the bib file format specification described on [www.bibtex.com](www.bibtex.com) and found that the problem is a bit more complicated than I thought.

First, the Fields section of [BibTeX-Format](https://www.bibtex.com/g/bibtex-format/) points out the formatting specification for the field's value: if it is a string, then it needs to be enclosed in `""` or `{}`, while if it is a number it does not.

Further, for [title field](https://www.bibtex.com/f/title-field/), the rule becomes a bit more complicated. But as far as the conclusion is concerned, for the value of the title field enclosed by two pairs of `{}`, removing the two layers of `{}` is the right way to handle it. The specific rules can be read in their documentation.

pmario:
IMO there is still more to it, which doesn't seem to be dealt with at all. eg: 

```
% It is also possible to preserve the capitaliztion of single 
% letters and ot use LaTeX commands for special chars
title = "Revisiting {W}ittgenstein on {K}{\"o}hler and {G}estalt psychology"
```

In this case your code doesn't work at all atm. ... and `{K}{\"o}hler` will give some extra problems. It needs to be converted to `Köhler`

pmario:
>In this case your code doesn't work at all atm. ... and {K}{\"o}hler will give some extra problems. It needs to be converted to Köhler

It would be interesting, what Mendeley exports in this case. ... 

From your example "example.bib.txt" it seems they only use `{}` and `{{}}` to completely preserve the text styling. ... So a `"{{some text}}".replace(/{|}/g,"")` will be sufficient. .. **But** .. It still seems a bit hacky to me. 

@Jermolene @saqimtiaz what do you think?

pmario:
@Gk0Wk ... I did some tests .. Using the KaTex edition. It turns out, that the title actually needs the outer braces to be rendered in the right way. 

image

I think this PR can be closed. IMO it's needed to manually modify the imported content if used without the KaTex pluign. ... and if used with it, it seems to be needed

PR #6302, opened 2021-12-04
Add a cascade filter run example

btheado:
Add a concrete, detailed `cascade` filter run example. It shows the intermediate output of each of the cascaded filters in order to illustrate how it works.


Jermolene:
Thanks @btheado that's marvellous

PR #6307, opened 2021-12-04
Add documentation tiddlers for intersection filter run prefix

btheado:
Add separate documentation tiddlers for the `:intersection` filter run prefix functionality following the pattern set by the :sort and :map filter run prefixes.

Moved the existing `:intersection` examples from "Filter Run Prefix (Examples).tid" to "Intersection Filter Run Prefix (Examples)" and added new examples.


Jermolene:
Thanks @btheado that's great.

PR #6310, opened 2021-12-04
Add documentation tiddlers for filter filter run prefix

btheado:
Add separate documentation tiddlers for the `:filter` filter run prefix functionality following the pattern set by the `:sort` and `:map` filter run prefixes.

The examples were mostly taken from the `filter` operator while modifying them to take advantage of the ability to inline the subfilter variables.


Jermolene:
Great stuff, thanks @btheado

PR #6318, opened 2021-12-08
Correcting Links to Official Plugin Demos

ibnishak:


ibnishak:
Corrections for #6319

Jermolene:
Hi @ibnishak the problem is that those links appear in both the release at https://tiddlywiki.com/ and the prerelease at https://tiddlywiki.com/prerelease. The present arrangement was an attempt to make the links work correctly in either case, at the expense of them working offline. We could probably do better now with a macro that dynamically makes the links as appropriate.

ibnishak:
Something like a <<baseurl>>marcro? In that case I am closing this PR and opening a discussion

PR #6326, opened 2021-12-09
Acknowledgements for banner design 5.1.21-5.2.0

ibnishak:
No github/social media links for Frank B was found. Otherwise it is all there


Jermolene:
Thanks @ibnishak

PR #6327, opened 2021-12-10
findListingsOfTiddler should cache results

flibbles:
`wiki.findListingsOfTiddler` was not using the FieldIndexer when it was available. This made operators like `[listed[]]` slow as butt with large TiddlyWikis.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -638,14 +638,20 @@ exports.getTagMap = function() {
 Lookup a given tiddler and return a list of all the tiddlers that include it in the specified list field
 */
 exports.findListingsOfTiddler = function(targetTitle,fieldName) {
+	var titles,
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @flibbles – I think it might be worth adding some simple tests for this. One of the existing test files includes a framework for running the same tests with and without indexers.

flibbles:
You were correct. It turns out adding tests was a good idea. There were none for [listed[]], and so it didn't catch that the FieldIndexer just doesn't plain work for getListingsOfTiddler. I've put in a new implementation of that method which caches the results, so subsequent calls don't also run through O(N*M) operations.

Arguably, it'd be more appropriate to have another indexer for list fields, but since this method is only used by [listed[]], which is only used once in core, I imagine my quick caching is probably sufficient for now.

Jermolene:
Thanks @flibbles

PR #6329, opened 2021-12-11
Type coercion for EditTextWidget

FlashSystems:
As promised in discussion #6301 I created a pull request containing a PoC for the new inputtype modules.

This set of two patches adds infrastructure and documentation to do type coercion on input fields. It introduces a new module-type `inputtype` that can convert values between the internal TW5 representation and the representation used by the value attribute of the HTML-tag.

The comment on commit 1fa29ee8073ac2d6fafb15afd13c81be0cc22beb describes this mechanism in more detail.
The second commit (f8b628d1cbedab426ecff7774902fd723722a324) contains the documentation.
If there is a need for changes or more documenation, tests, etc. feel free to let me know.

### DIFF
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -32,6 +32,11 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	};
 
 	/*
+	Initialize the global List of type modules
+	*/
+	EditTextWidget.typeModules = $tw.modules.getModulesByTypeAsHashmap("inputtype", "type");
+
+	/*
 	Inherit from the base widget class
 	*/
 	EditTextWidget.prototype = new Widget();
@@ -134,12 +139,18 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 					updateFields = {
 						title: self.editTitle
 					};
+				if (self.typeModule && self.typeModule.fromValue) {
+					value = self.typeModule.fromValue.call(this,value);
+				}
 				updateFields[self.editField] = value;
 				self.wiki.addTiddler(new $tw.Tiddler(self.wiki.getCreationFields(),tiddler,updateFields,self.wiki.getModificationFields()));
 			};
 		}
 		if(this.editType) {
 			type = this.editType;
+		}
+		if(this.typeModule && this.typeModule.toValue) {
+			value = this.typeModule.toValue.call(this,value);
 		}
 		return {value: value || "", type: type, update: update};
 	};
@@ -205,6 +216,9 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		// Get the rest of our parameters
 		this.editTag = this.getAttribute("tag",tag) || "input";
 		this.editType = this.getAttribute("type",type);
+		// Fetch type module if any exists.
+		// This module takes kare of converting between TW5 and HTML5 representation of the value.
+		this.typeModule = EditTextWidget.typeModules[this.editType];
 		// Make the child widgets
 		this.makeChildWidgets();
 		// Determine whether to show the toolbar
--- a/core/modules/editor/types/date.js
+++ b/core/modules/editor/types/date.js
@@ -1,0 +1,27 @@
+/*\
+title: $:/core/modules/editor/types/date.js
+type: application/javascript
+module-type: inputtype
+
+Type module for HTML5 input elements with type="date"
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.type = "date";
+
+exports.toValue = function(twDateTimeString) {
+	var date = $tw.utils.parseDate(twDateTimeString);
+	return $tw.utils.isValidDate(date) ? date.toISOString().substring(0,10) : "";
+};
+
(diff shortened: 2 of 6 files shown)


pmario:
A fast feedback about 1 point, that immediately jump into my eye. ... You should have made a feature-branch instead of committing to the master-branch in your repo. ... 

If Jeremy merges content to Jermolene:master you may get conflicts, that can't be easily resolved. ...

editions/dev/tiddlers/new/InputTypeModules.tid
```diff
@@ -0,0 +1,47 @@
+created: 202112111013000000
+modified: 202112111013000000
+title: inputtype-modules
+type: text/vnd.tiddlywiki
+
+Modules of the type `inputtype` provide a way to convert the TiddlyWiki internal representation of a value into a value used within the HTML-value-attribute of an input tag and back.
+
+These modules are used to allow the date and time types of HTML5 input tags to function seamlessly with TiddlyWiki's date/time format.
+
+A module of type `inputtype` exports the `type` property that specifies on which type of `input` element they are applied to. In addition to that, they contain two exported functions `toValue` and `fromValue`. These functions are called each time a value needs to be converted from TW5`s representation into the HTML-value-attribute and vice versa.
```
pmario, line 10: There is a typo in ``converted from TW5`s representation`` See the backtick in ``TW`s`` It should be `TW's`

pmario:
I did use this wikitext code to test it:

```
<$edit-text tiddler="test" field="mydate" type=date/>

Value of ''mydate'' : {{test!!mydate}}

-----

<$edit-text tiddler="test" field="mydate-time" type="datetime-local"/>

Value of ''mydate-time'' : {{test!!mydate-time}}

{{{[[test]get[mydate-time]format:date[YYYY 0MM 0DD 0hh:0mm]]}}}
```

I think possibilities like this have been requested a 100 times in the past. Especially since most browsers natively support a date-input UI now. NO need to add heavy 3rd party libs anymore

grafik

pmario:
@FlashSystems ... Why did you use `datetime-local` instead of `datetime` or `date-time` ?

FlashSystems:
> A fast feedback about 1 point, that immediately jump into my eye. ... You should have made a feature-branch instead of committing to the master-branch in your repo. ...
> 
> If Jeremy merges content to Jermolene:master you may get conflicts, that can't be easily resolved. ...

Thank you @pmario for noticing. That's really embarrassing 🤦🏼‍♂️. I cherry picked the changed from my local repository where I experiment with the TiddlyWiki code and simply forgot to create a feature branch. I'll create one and switch this pull request over to the new branch.

...and I'll fix the misused back-tick.

> Why did you use datetime-local instead of datetime or date-time ?

W3C removed the input type `datetime` from the standard and replaced it with `datetime-local`. [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input/datetime) has more information about this.

Your comment sparked a though that is better explained within the discussion (#6301). I'll post it there as soon as I've though it through.

core/modules/editor/factory.js
```diff
@@ -205,6 +216,9 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		// Get the rest of our parameters
 		this.editTag = this.getAttribute("tag",tag) || "input";
 		this.editType = this.getAttribute("type",type);
+		// Fetch type module if any exists.
+		// This module takes kare of converting between TW5 and HTML5 representation of the value.
+		this.typeModule = EditTextWidget.typeModules[this.editType];
```
Jermolene, line 221: I don't think any other widgets instantiate submodules, but the pattern we use elsewhere is like this:

https://github.com/Jermolene/TiddlyWiki5/blob/54cfda76ee353190f1cf0210b9071894fb1a5690/core/modules/parsers/wikiparser/wikiparser.js#L62-L65

In otherwords, we avoid calling `$tw.modules.*` at parse time, we do it when the object is instantiated.

core/modules/utils/utils.js
```diff
@@ -336,6 +336,10 @@ exports.formatTitleString = function(template,options) {
 	return result;
 };
 
+exports.isValidDate = function(date) {
```
Jermolene, line 339: I assume that we can't use the existing `$tw.utils.isDate()` method?

FlashSystems:
Hm, it seems GitHub only allows switching the destination branch of a pull request, not the source. Would you mind if I close this pull request and create a new one?

core/modules/editor/types/date.js
```diff
@@ -0,0 +1,27 @@
+/*\
+title: $:/core/modules/editor/types/date.js
+type: application/javascript
+module-type: inputtype
```
Jermolene, line 4: Perhaps the module type should be `editmapping`? (the word "type" is a bit overloaded in the codebase)

Jermolene (commented):
Thanks @FlashSystems, some comments

core/modules/utils/utils.js
FlashSystems: Thank you for the hint. I haven't found an `isDate` method in `utils.js` and thought there wasn't any available. I'll switch over to that one.

FlashSystems:
Closed in favour of PR #6377

PR #6335, opened 2021-12-13
add polish translations to cascades

EvidentlyCube:
Added Polish translations for new translation strings introduced in #6168


Jermolene:
Thanks @EvidentlyCube

PR #6338, opened 2021-12-14
Clarify list location requirement

jc-ose:
This is to clarify that the list cannot be contained in the current tiddler where the macro is called. 

Edit: This is for the **list-links-draggable** macro documentation.


Jermolene:
Thanks @jc-ose

PR #6351, opened 2021-12-17
Docs: Filter Operators: indicate Selection Constructors

davout1806:
In 'Filter Operators' added column to indicate those operators that construct an entirely new selection i.e. Selection Constructors.

The one question for this was how to indicate the operators (all, subfilter) that may or may not act as a constructor based on their usage. I for all, subfilter I added a new tag 'Selection Constructors: Conditional'. This, understandingly, might not a desired direction. An alternative is to remove that tag and just make a general note such as "Please note even if operator is identified with C depending on usage, it might not actually create a new selection. Be sure to check individual operator documentation."

### DIFF
--- a/editions/tw5.com/tiddlers/filters/Filter Operators.tid
+++ b/editions/tw5.com/tiddlers/filters/Filter Operators.tid
@@ -1,5 +1,5 @@
 created: 20140410103123179
-modified: 20210830010231559
+modified: 20211217141224284
 tags: Filters
 title: Filter Operators
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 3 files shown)


ibnishak:
Alternate proposal: How about we remove the columns common, negatable, etc. Instead we add a single column named comments. In the column we add info like "negatable", "common", "new constructor" etc as tag pills. If each tag pill has a different colour, it might be easier to visually delineate too.

Advantage is we can future proof adding more info.

davout1806:
Works for me in theory. A few thoughts:

1. Are you suggesting adding new tags: "negatable", "common", "new constructor"? Or use the existing tags:  "Filter Operators", "Negatable Operators", "Selection Constructors"?
2. Do you have a concern that an operator with multiple qualities might "bloat" that cell? Distorting the table layout.
3. Could we add a field called say op-comment-short to all the tags we want to make a comment about ("Filter Operators", "Negatable Operators", "Selection Constructors", etc). Then in the table, for each operator, list the values of all the op-comment-short fields for all the tags the operator has. The op-comment-short would have either the check mark, !, or whatever short identifier. Above the table could be list of all unique op-comment-short value along with either title of the tag, link to tag to see explanation, or add another field say op-comment-description that describes what it means. This could make it more future, as in theory, the Filter Operators tiddler would not have to change to add a new comment type. 

Just thought up number 3 so there might be some gotchas I'm not recognizing. Maybe this was originally the level of future proof you were thinking about.

ibnishak:
> Are you suggesting adding new tags: "negatable", "common", "new constructor"? Or use the existing tags: "Filter Operators", "Negatable Operators", "Selection Constructors"?

No. Most of those tags are obvious. It is a table of "Filter operators", so that tag is pointless. Table is already subdivided to sections like "Order", "Mathematics" etc. So those tags are also unnecessary. The comments column need only highlight those qualities which are not obvious. At present these are only 3, Common, Negatable, Constructor. 

> Do you have a concern that an operator with multiple qualities might "bloat" that cell? Distorting the table layout.
Valid concern. This is won't be an issue currently  as there are only 3 short tag pills. However future proofing wise, two other issues need to be looked at here
1. Should documentation site bother to cater to small screen users? If yes, can horizontal scroll for wide tables be considered?
2. Especially in small screens, a descriptive column would be more suited than a symbol-key combination for long tables. Users can understand what is being conveyed without referring to a a key at the end of table

> The op-comment-short would have either the check mark, !, or whatever short identifier
Short identifier field sounds like a great idea. It would make the data collection for the table decentralized. 

After all this, I must admit I do see value in the current approach too. In some point in future, we can add make the table dynamic by adding ability to sort by columns.

davout1806:
OK so do you want me to work up a separate pull request using op-comment-short field?

ibnishak:
No. This is an open discussion. If more people chips in and gives their opinion, good. Ultimately Jeremy will guide you if you need to make any changes to the PR.

Jermolene:
Thanks @davout1806 @ibnishak and apologies for the delay.

I share the concern that this introduces a little more complexity into the table but it conveys useful information.

PR #6353, opened 2021-12-19
Using a different account now

nilslindemann:


Jermolene:
Thanks @nilslindemann – apologies, but it's probably better to do this on the `tiddlywiki-com` branch because we regularly merge that into `master`. Otherwise, the update won't appear on the main `tiddlywiki-com` branch until we do a release.

nilslindemann:
Sorry, I am just fighting to get rid of the other commits which I did not do. Have asked for help on [SO](https://stackoverflow.com/questions/70646809/undo-two-pull-requested-but-not-merged-commits).

PR #6355, opened 2021-12-19
Add ViewWidget cross references and examples

btheado:
Added links from the ViewWidget formats which have corresponding filter operators. Added example output for the non-date formats.


Jermolene:
Great stuff thanks @btheado, that's a good idea, very helpful.

PR #6361, opened 2021-12-21
Add section about dynamic links to 'Linking in WikiText'

btheado:
Explain what underlying widgets WikiText links translate to. Point out that it is easier to generate dynamic links using the LinkWidget directly.

### DIFF
--- a/editions/tw5.com/tiddlers/wikitext/Linking in WikiText.tid
+++ b/editions/tw5.com/tiddlers/wikitext/Linking in WikiText.tid
@@ -1,6 +1,6 @@
 caption: Linking
 created: 20131205155230596
-modified: 20211221221052922
+modified: 20211230145939554
 tags: WikiText
 title: Linking in WikiText
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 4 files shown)


editions/tw5.com/tiddlers/wikitext/Linking in WikiText.tid
```diff
@@ -80,3 +80,40 @@ You can also use the extended syntax to force an external link:
...
+
+For example, an attempt to use a [[reference|TextReference]] and a [[variable|Variables]] to dynamically generate a link like this:
+
+```
+[[{{!!caption}}|<<currentTiddler>>]]
+```
+is the same as trying to use the LinkWidget and TextWidget like this:
+
+```
+<$link to="<<currentTiddler>>">
```
Jermolene, line 107: I'm concerned that examples of what doesn't work or not to do might be misinterpreted by people scanning the text, and not reading it all properly.

We did attempt to devise a styling to indicate non-functional examples:

https://tiddlywiki.com/#Concatenating%20text%20and%20variables%20using%20macro%20substitution

It's not wildly successful, and has yet to be made into a macro, but perhaps this is the time to introduce a `<<.bad-example "">>` docs macro

Jermolene (commented):
Thanks for your patience @btheado

editions/tw5.com/tiddlers/wikitext/Linking in WikiText.tid
btheado: Thanks for the feedback and good idea. I created a `<<.bad-example "">>` docs macro as you suggested and I made use of it for the text in my original commit as well as for the text in the tiddler you referenced.

Jermolene:
Thanks for your patience again @btheado

Issue #6342, opened 2021-12-16
[IDEA] Short ascii alias to tiddlers with long, unicode titles


shardator:
**Is your feature request related to a problem? Please describe.**

I have a small wiki in Hungarian, and there are many accented letters in the language. I also want to have titles decorated with emojis, add commas, etc. But I'd also like to have a way to simply link them, without having to copy the whole title.

Sometimes, when I change the title to this new format, the references work, but sometimes they get broken. For example the link

```
#ArvizturoTukorfurogep
```

opens a new empty tiddler with the same title, but I'd like to change the title to

```
Árvíztűrő™ Tükörfúrógép®
```

sometimes this works, but more often it does not. **And when it does not, it is totally silent, the link gets lost, the new page gets orpaned, unreachable.** Clicking the link opens a new empty tiddler again.

**Describe the solution you'd like**

I'd simply like to have a way to specify **both** a visible title and a **wikiword** for the same tiddler.

**Describe alternatives you've considered**

None.

**Additional context**

None.

shardator:
OK, there is a solution:

> As for differently named links opening the same tiddler, Mario has made a uni-link plugin that does this using I believe aliases. I have not used it, but many others do.

shardator:
The only problem with UniLink that it is a bit tedious (one has to carefully set the aliases, they are not filled automatically clicking on the link for a non-existing page), **and does not work with markdown!**. Is there a way around the latter?

pmario:
>The only problem with UniLink that it is a bit tedious (one has to carefully set the aliases, they are not filled automatically clicking on the link for a non-existing page),

There has been a discussion lately, that the "Missing tiddler "xxxx" – click to create" template doesn't allow us to create markdown tiddlers. The default action will always create a standard TW tiddler. see: **tm-edit-tiddler should take (and use) a bag of fields like tm-new-tiddler** #6001 

The idea there was, that the markdown plugin could add a template, that would be show as an additional text eg: "Missing tiddler "xxxx" – click to create a markdown tiddler" 

The "edit tiddler" toolbar button would also need a possibility to create an MD or TW tiddler .. 

It may be possible, that uni-link also brings a "template" that shows "Create New Alias Tiddler" with a prefilled **aliases** field. ... 

I think this would be valid extension for the plugin.

pmario:
I did add a new issue to the uni-link repo: https://github.com/wikilabs/plugins/issues/108

pmario:
[...]

PR #6364, opened 2021-12-22
Missing tiddler cascade - To extend the "missing tiddler hint" in the missing tiddler body

pmario:
Many users create links in a tiddler and then click the missing link to create the new tiddlers. 

This workflow works well, if users use `text/vnd.tiddlywiki` type tiddlers. 

But if users mainly work with eg: `text/markdown` the mechanism fails, since the default buttons shown in the tiddler body and tiddler toolbar will create tiddlers with an **empty type field**. 

See: https://talk.tiddlywiki.org/t/how-do-i-make-the-click-edit-to-create-button-make-a-markdown-tiddler/436
See: https://groups.google.com/g/tiddlywiki/c/HtkGP4M1P8M/m/QMhmMXCiCAAJ
see: https://github.com/Jermolene/TiddlyWiki5/issues/6342

This PR prepares the UI to use a "missing tiddler cascade" so plugin authors and core plugins are more flexible to create an improved "Missing Tiddler Workflow". 

This PR doesn't change the default UI or the default workflow as it exists at the moment. It only creates a new "Missing Tiddler Cascade" which plugin authors can use to hook into. 

There will be a second PR, with a proposal, how this  new cascade can be used to improve the core markdown plugin.

The extended Cascase Info will look like this:
missing-tiddler-body-01

and

missing-tiddler-body-02


pmario:
.

Jermolene:
Thanks @pmario – if the markdown plugin needs to add its own missing tiddler template, surely it could just insert a new rule in the view template body cascade immediately before $:/config/ViewTemplateBodyFilters/default such as `[is[missing]then[$:/plugins/markdown/missingtiddler]]`?

The other way of approaching the original goal might be to introduce segments into the default $:/core/ui/ViewTemplate/body/default template using a tag such as $:/tags/MissingTiddlerSegment. That way, any amount of plugins can add their own custom UI.

pmario:
Just a note. Related to: https://talk.tiddlywiki.org/t/i-overcame-a-hurdle-choose-from-custom-new-tiddler-buttons-when-opening-a-missing-link/4799

pmario:
markdown plugin needs to handle this one

PR #6377, opened 2021-12-31
Extend the EditTextWidget and fields to allow Date-Pickers and other tags

FlashSystems:
This Pull request is the result of the discussion #6301.

It extends TiddlyWiki in multiple ways:

1. It allows transformation of the data stored in a field when it is used as the value of a HTML-tag. That way HTML5 date and time pickers can be used.
2. It adds the `parseDate` filter to allow processing of dates in ISO-8601 format.
3. It allows the same transformation mechanism to be used for input validation.
4. It allows the rendering of fields to be extended. Custom input tags can be used for tiddler fields.

Currently the documentation is missing. I'll add the documentation as soon as the technical discussion is resolved.

### DIFF
--- a/core/modules/editor/engines/simple.js
+++ b/core/modules/editor/engines/simple.js
@@ -82,6 +82,19 @@ SimpleEngine.prototype.setText = function(text,type) {
 };
 
 /*
+Set the validation status of the input field
+If the parameter is true, the field is valid. If it's false it is not valid
+and should be shown as such to the user.
+*/
+SimpleEngine.prototype.updateValidationStatus = function(isValid) {
+	if(isValid) {
+		$tw.utils.removeClass(this.domNode,this.widget.editErrorClass);
+	} else {
+		$tw.utils.addClass(this.domNode,this.widget.editErrorClass);
+	}
+}
+
+/*
 Update the DomNode with the new text
 */
 SimpleEngine.prototype.updateDomNodeText = function(text) {
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -221,7 +221,7 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.placeholder || changedAttributes.size || changedAttributes.autoHeight || changedAttributes.minHeight || changedAttributes.focusPopup ||  changedAttributes.rows || changedAttributes.tabindex || changedAttributes.cancelPopups || changedAttributes.inputActions || changedAttributes.refreshTitle || changedAttributes.autocomplete || changedTiddlers[HEIGHT_MODE_TITLE] || changedTiddlers[ENABLE_TOOLBAR_TITLE] || changedAttributes.disabled || changedAttributes.fileDrop) {
 			this.refreshSelf();
 			return true;
-		} else if (changedTiddlers[this.editRefreshTitle]) {
+		} else if(changedTiddlers[this.editRefreshTitle]) {
 			this.engine.updateDomNodeText(this.getEditInfo().value);
 		} else if(changedTiddlers[this.editTitle]) {
 			var editInfo = this.getEditInfo();
@@ -306,7 +306,7 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		var propertiesToCopy = propertiesToCopy || [],
 			newEvent = this.document.createEventObject ? this.document.createEventObject() : this.document.createEvent("Events");
 		if(newEvent.initEvent) {
-			newEvent.initEvent(event.type, true, true);
+			newEvent.initEvent(event.type,true,true);
 		}
 		$tw.utils.each(propertiesToCopy,function(prop){
 			newEvent[prop] = event[prop];
--- a/core/modules/editor/factory.js
+++ b/core/modules/editor/factory.js
@@ -134,12 +134,24 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 					updateFields = {
 						title: self.editTitle
 					};
-				updateFields[self.editField] = value;
-				self.wiki.addTiddler(new $tw.Tiddler(self.wiki.getCreationFields(),tiddler,updateFields,self.wiki.getModificationFields()));
+				if(self.editFromInputValueFilter) {
+					value = self.wiki.filterTiddlers(self.editFromInputValueFilter,self,self.wiki.makeTiddlerIterator([value]))[0];
+				}
+				if(value !== undefined) {
+					updateFields[self.editField] = value;
+					self.wiki.addTiddler(new $tw.Tiddler(self.wiki.getCreationFields(),tiddler,updateFields,self.wiki.getModificationFields()));
+					self.setValidationStatus(true);
+				} else {
+					self.setValidationStatus(false);
+				}
 			};
 		}
 		if(this.editType) {
 			type = this.editType;
+		}
+		if(this.editToInputValueFilter) {
+			var source = this.wiki.makeTiddlerIterator([value]);
+			value = this.wiki.filterTiddlers(this.editToInputValueFilter,self,source)[0];
 		}
 		return {value: value || "", type: type, update: update};
 	};
@@ -172,6 +184,7 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		this.editIndex = this.getAttribute("index");
 		this.editDefault = this.getAttribute("default");
 		this.editClass = this.getAttribute("class");
+		this.editErrorClass = this.getAttribute("errorClass","tw-input-error");
 		this.editPlaceholder = this.getAttribute("placeholder");
 		this.editSize = this.getAttribute("size");
 		this.editRows = this.getAttribute("rows");
@@ -185,6 +198,8 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		this.editInputActions = this.getAttribute("inputActions");
 		this.editRefreshTitle = this.getAttribute("refreshTitle");
 		this.editAutoComplete = this.getAttribute("autocomplete");
+		this.editToInputValueFilter = this.getAttribute("toInputValue");
+		this.editFromInputValueFilter = this.getAttribute("fromInputValue");
 		this.isDisabled = this.getAttribute("disabled","no");
 		this.isFileDropEnabled = this.getAttribute("fileDrop","no") === "yes";
 		// Get the default editor element tag and type
@@ -205,6 +220,8 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		// Get the rest of our parameters
 		this.editTag = this.getAttribute("tag",tag) || "input";
 		this.editType = this.getAttribute("type",type);
+		// Initialize global variables
+		this.isCurrentlyValid = true;
 		// Make the child widgets
 		this.makeChildWidgets();
 		// Determine whether to show the toolbar
@@ -213,19 +230,37 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 	};
 
 	/*
+	Called by different codepaths to set the validation status.
+	Only forwards the validation status to the engine (as an update) if it has changed. this minimizes update calls.
+	It is checked, that the engine supports `updateValidationStatus`. If it is not implemented, the call is essentially a NoOp.
+	*/
+	EditTextWidget.prototype.setValidationStatus = function(isValid) {
+		if(isValid !== this.isCurrentlyValid) {
+			this.isCurrentlyValid = isValid;
+			if(typeof this.engine.updateValidationStatus === "function") {
+				this.engine.updateValidationStatus(isValid)
+			}
+		}
+	}
+
+	/*
 	Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
 	*/
 	EditTextWidget.prototype.refresh = function(changedTiddlers) {
 		var changedAttributes = this.computeAttributes();
 		// Completely rerender if any of our attributes have changed
-		if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.placeholder || changedAttributes.size || changedAttributes.autoHeight || changedAttributes.minHeight || changedAttributes.focusPopup ||  changedAttributes.rows || changedAttributes.tabindex || changedAttributes.cancelPopups || changedAttributes.inputActions || changedAttributes.refreshTitle || changedAttributes.autocomplete || changedTiddlers[HEIGHT_MODE_TITLE] || changedTiddlers[ENABLE_TOOLBAR_TITLE] || changedAttributes.disabled || changedAttributes.fileDrop) {
+		if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.placeholder || changedAttributes.size || changedAttributes.autoHeight || changedAttributes.minHeight || changedAttributes.focusPopup ||  changedAttributes.rows || changedAttributes.tabindex || changedAttributes.cancelPopups || changedAttributes.inputActions || changedAttributes.refreshTitle || changedAttributes.autocomplete || changedTiddlers[HEIGHT_MODE_TITLE] || changedTiddlers[ENABLE_TOOLBAR_TITLE] || changedAttributes.disabled || changedAttributes.fileDrop || changedAttributes.toInputValue || changedAttributes.fromInputValue || changedAttributes.errorClass) {
 			this.refreshSelf();
 			return true;
 		} else if(changedTiddlers[this.editRefreshTitle]) {
 			this.engine.updateDomNodeText(this.getEditInfo().value);
+			// If an external value is set, it's always considered valid because it is already
+			// contained within the field.
+			this.setValidationStatus(true);
 		} else if(changedTiddlers[this.editTitle]) {
 			var editInfo = this.getEditInfo();
 			this.updateEditor(editInfo.value,editInfo.type);
+			this.setValidationStatus(true);
 		}
 		this.engine.fixHeight();
 		if(this.editShowToolbar) {
@@ -257,6 +292,11 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		var editInfo = this.getEditInfo();
 		if(text !== editInfo.value) {
 			editInfo.update(text);
+		} else {
+			// If the text matches the saved value exactly we've to remove
+			// the error-flag. There will be no call to update, but the input
+			// is valid, because ist was valid before.
+			this.setValidationStatus(true);
 		}
 	};
 
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -95,8 +95,8 @@ function FramedEngine(options) {
 			{name: "dragover",handlerObject: this.widget,handlerMethod: "handleDragOverEvent"},
 			{name: "dragleave",handlerObject: this.widget,handlerMethod: "handleDragLeaveEvent"},
 			{name: "dragend",handlerObject: this.widget,handlerMethod: "handleDragEndEvent"},
-			{name: "drop", handlerObject: this.widget,handlerMethod: "handleDropEvent"},
-			{name: "paste", handlerObject: this.widget,handlerMethod: "handlePasteEvent"},
+			{name: "drop",handlerObject: this.widget,handlerMethod: "handleDropEvent"},
+			{name: "paste",handlerObject: this.widget,handlerMethod: "handlePasteEvent"},
 			{name: "click",handlerObject: this.widget,handlerMethod: "handleClickEvent"}
 		]);
 	}
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -132,6 +132,19 @@ FramedEngine.prototype.setText = function(text,type) {
 };
 
 /*
+Set the validation status of the input field
+If the parameter is true, the field is valid. If it's false it is not valid
+and should be shown as such to the user.
+*/
+FramedEngine.prototype.updateValidationStatus = function(isValid) {
+	if(isValid) {
+		$tw.utils.removeClass(this.domNode,this.widget.editErrorClass);
+	} else {
+		$tw.utils.addClass(this.domNode,this.widget.editErrorClass);
+	}
+}
+
+/*
 Update the DomNode with the new text
 */
 FramedEngine.prototype.updateDomNodeText = function(text) {
--- a/editions/tw5.com/tiddlers/images/Value transformations.svg
+++ b/editions/tw5.com/tiddlers/images/Value transformations.svg
@@ -1,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- Created with Inkscape (http://www.inkscape.org/) -->
+
+<svg
+   width="64.157005mm"
+   height="29.740005mm"
+   viewBox="0 0 64.157005 29.740005"
+   version="1.1"
+   id="svg5"
+   xmlns="http://www.w3.org/2000/svg"
+   xmlns:svg="http://www.w3.org/2000/svg">
+  <defs
+     id="defs2">
+    <marker
+       style="overflow:visible"
+       id="Arrow1Send"
+       refX="0"
+       refY="0"
+       orient="auto">
+      <path
+         transform="matrix(-0.2,0,0,-0.2,-1.2,0)"
+         style="fill:context-stroke;fill-rule:evenodd;stroke:context-stroke;stroke-width:1pt"
+         d="M 0,0 5,-5 -12.5,0 5,5 Z"
+         id="path9562" />
+    </marker>
+    <marker
+       style="overflow:visible"
+       id="Arrow1Send-3"
+       refX="0"
+       refY="0"
+       orient="auto">
+      <path
+         transform="matrix(-0.2,0,0,-0.2,-1.2,0)"
+         style="fill:context-stroke;fill-rule:evenodd;stroke:context-stroke;stroke-width:1pt"
+         d="M 0,0 5,-5 -12.5,0 5,5 Z"
+         id="path9562-6" />
+    </marker>
+  </defs>
+  <g
+     id="layer1"
+     transform="translate(-16.643257,-18.311505)">
+    <text
+       xml:space="preserve"
+       style="font-size:4.23333px;line-height:1.25;font-family:sans-serif;stroke-width:0.264583"
+       x="37.550411"
+       y="21.527843"
+       id="text5090"><tspan
+         id="tspan5088"
+         style="font-size:4.23333px;stroke-width:0.264583"
+         x="37.550411"
+         y="21.527843">TW5 field</tspan></text>
+    <path
+       style="fill:none;fill-opacity:1;stroke:#000000;stroke-width:1;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;marker-end:url(#Arrow1Send)"
+       d="M 40.760143,24.493964 V 35.384493"
+       id="path9429" />
+    <path
+       style="fill:none;fill-opacity:1;stroke:#000000;stroke-width:1;stroke-linecap:butt;stroke-linejoin:miter;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1;marker-end:url(#Arrow1Send-3)"
+       d="M 52.42257,37.169834 V 26.279305"
+       id="path9429-7" />
+    <text
+       xml:space="preserve"
+       style="font-size:3.52777px;line-height:1.25;font-family:sans-serif;stroke-width:0.264583"
+       x="16.548517"
+       y="31.805136"
+       id="text10857"><tspan
+         id="tspan10855"
+         style="stroke-width:0.264583"
+         x="16.548517"
+         y="31.805136">toInputValue</tspan></text>
+    <text
+       xml:space="preserve"
+       style="font-size:3.52777px;line-height:1.25;font-family:sans-serif;stroke-width:0.264583"
+       x="53.956135"
+       y="31.805136"
+       id="text11301"><tspan
+         id="tspan11299"
+         style="stroke-width:0.264583"
+         x="53.956135"
+         y="31.805136">fromInputValue</tspan></text>
+    <g
+       id="g15663">
+      <rect
+         style="fill:#ffffff;fill-opacity:1;stroke:none;stroke-width:0.612694;stroke-linecap:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+         id="rect6690"
+         width="35.519291"
+         height="7.5835614"
+         x="29.564837"
+         y="40.161602" />
+      <path
+         d="m 65.084131,40.161603 v 7.583561 H 29.564837"
+         style="fill:none;stroke:#787878;stroke-width:0.612694;stroke-linecap:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+         id="path2552" />
+      <path
+         id="rect846"
+         style="fill:none;stroke:#282828;stroke-width:0.612694;stroke-linecap:round;stroke-miterlimit:4;stroke-dasharray:none;stroke-opacity:1"
+         d="M 29.564837,47.745164 V 40.161603 H 65.08413" />
+      <text
+         xml:space="preserve"
+         style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:4.23333px;line-height:0;font-family:sans-serif;-inkscape-font-specification:sans-serif;stroke-width:0.264583"
+         x="30.247374"
+         y="45.370037"
+         id="text13577"><tspan
+           id="tspan13575"
+           style="font-style:normal;font-variant:normal;font-weight:normal;font-stretch:normal;font-size:4.23333px;line-height:2;font-family:sans-serif;-inkscape-font-specification:sans-serif;stroke-width:0.264583"
+           x="30.247374"
+           y="45.370037">HTML element<tspan
+   style="font-size:4.58611px;line-height:2;letter-spacing:0px;baseline-shift:baseline"
+   id="tspan16786">|</tspan></tspan></text>
+    </g>
+  </g>
+</svg>
--- a/editions/tw5.com/tiddlers/images/Value transformations.svg.meta
+++ b/editions/tw5.com/tiddlers/images/Value transformations.svg.meta
@@ -1,0 +1,3 @@
+title: Value transformation.svg
+type: image/svg+xml
+tags: picture
--- a/editions/tw5.com/tiddlers/widgets/EditTextWidget Transformations.tid
+++ b/editions/tw5.com/tiddlers/widgets/EditTextWidget Transformations.tid
@@ -1,0 +1,44 @@
+created: 20220109200433409
+modified: 20220109200433409
+tags: Widgets
+title: EditTextWidget value transformations
+type: text/vnd.tiddlywiki
+
+By default, the [[<$edit-text> widget|EditTextWidget]] show the value of a field exactly like it is stored within the associated field. The `toInputValue` and `fromInputValue` attributes provide a way to change this behaviour.
+
+The filter assigned `toInputValue` attribute is used to modify the value after it was read from the field, before it is displayed to the user. If this filter returns no value, an empty value is assigned to the generated HTML editing element.
+
+The filter assigned to the `fromInputValue` attribute is used to modify the value that was entered by the user, before it is stored into the field. If this filter returns no value, the CSS class specified in the `errorClass` attribute is assigned to the generated HTML editing element.
+
+! Transforming values
+
+As an example for the transformation that can be done by using the `toInputValue` and `fromInputValue` attributes, let's build a date picker that uses the HTML5 "date" input type and stores its value inside a TW5 field.
+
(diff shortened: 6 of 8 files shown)


kookma:
@FlashSystems 
Based on MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input
The html5 supports several type of inputs! I did not go through the PR (lack of my JS knowledge) does your solution allows to use other type of inputs (I mean those not covered by Tiddlywiki widgets)

pmario:
@FlashSystems .. Can you post some example wikitext code here, how to use it?

FlashSystems:
> @FlashSystems .. Can you post some example wikitext code here, how to use it?

Hallo @pmario, I've build a [demo](https://flashsystems.github.io/FieldTemplatesDemo/) that uses all features. I hope this helps. If you have further questions, let me know.

FlashSystems:
> @FlashSystems Based on MDN docs: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input The html5 supports several type of inputs! I did not go through the PR (lack of my JS knowledge) does your solution allows to use other type of inputs (I mean those not covered by Tiddlywiki widgets)

Because the input transformation is generic (done via a filter expression). The solution should allow to use any type of input.

pmario:
I think there is something wrong. To reproduce the issue - Use the following wikitext.

```
<$edit-text field="mydateX" type="datetime-local" placeholder="Set your date" fromValue="[parseDate[ISO]]" />

<$view field="mydateX"/>
```

 - Create Date and Time with the dropdown
 - Now change the value-field in the field editor. -> The edit-text widget isn't updated but set to default  

I also think, that `ISO` isn't the right term.. TW format is not the same as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

FlashSystems:
Thank you @pmario for trying it out. The points you struggled with are a good hint for what should be made clear in the documentation. If the approach is not clear to an experienced TW user, an average user will surly have problems there, too.
> I think there is something wrong. To reproduce the issue - Use the following wikitext.
> [...]

The problem here is, that you haven't specified a `toValue` filter. TiddlyWiki does not know how to convert the changed field value into the representation needed for the input-tag. If you extend your `<$edit-text` line with `toValue="[format:date[YYYY-0MM-0DDT0hh:0mm:0ss]]"` everything will work as expected.Thank you @pmario for trying it out. The points you struggled with are a good hint for what should be made clear in the documentation. If the approach is not clear to an experienced TW user, an average user will surly have problems there, too.
> I think there is something wrong. To reproduce the issue - Use the following wikitext.

This is my only concern with this (much more flexible) approach. It's less discoverable than the approach in #6329. It's not of the "just works" type. The future user will have to read the documentation to understand how it works. And the documentation must contain some really good examples.

> 
> I also think, that `ISO` isn't the right term.. TW format is not the same as [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601)

I think ISO fits because `parseDate[ISO]` parses the date in ISO format and returns the date in TW format. It is used in the `fromValue` filter that converts the value from the the input-tag into the TW representation.

`fromValue`: Convert from input-tag to TW
`toValue`: Convert from TW to input-tag

`toValue` and `fromValue` are almost always used together. Only if `fromValue` is used as an input validator (returning the value unchanged if it was successfully validated) the `toValue` filter can be omitted.

pmario:
>This is my only concern with this (much more flexible) approach. 

I do like the possibilities. You are right, we need to have good examples, which makes it clear for users what's going on. 

----

... I did remove the toValue parameter after my first test failed. I thought it would use a default value ... May be we should have some "fallback / default" values here @Jermolene ?! 

With the suggested `toValue` from your response it works. -> OK

But the problem now is, that the `toValue` format `YYYY-0MM-0DDT0hh:0mm:0ss` has nothing to do, with what is shown on the screen. ... 

See screenshot:

image

I think `toInputValue` may be a better/different parameter term, because it is only used if `type="input"` .... I'm not sure here..

image

Thx for the work you put into this! I think it will be a great and well received new possibility. to create better "field UI".

FlashSystems:
> But the problem now is, that the toValue format YYYY-0MM-0DDT0hh:0mm:0ss has nothing to do, with what is shown on the screen.

That's actually by design. An input-Tag with type "date" for example expects its [value to be an ISO 8601 date](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#value). But the value is displayed according to the users locale. In my browser using a german locale the date input box looks like this:
image
But the internal format is the same (simplified) ISO 8601 that whould be shown as "13/01/2022" in your browser. The value the user sees is rendered according to the users locale and the style of the used browser.

core/modules/editor/factory.js
```diff
@@ -257,6 +277,11 @@ function editTextWidgetFactory(toolbarEngine,nonToolbarEngine) {
 		var editInfo = this.getEditInfo();
 		if(text !== editInfo.value) {
 			editInfo.update(text);
+		} else if(text === editInfo.value) {
```
Jermolene, line 280: The second "else if" could just be "else".

Are we certain that this code will never clear the error state without previously having set it? It seems as though there could be false fires here. We want to avoid that because removing the class means hitting the DOM.

core/modules/filters/parsedate.js
```diff
@@ -0,0 +1,38 @@
...
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports.parseDate = function(source,operator,options) {
```
Jermolene, line 18: Filter operator names do not use camel case.

In any case, I'm not sure that the operator is needed – the alternative is to add support to `$tw.utils.parseDate()`, and then the existing date operators will all be able to cope with ISO format.

Jermolene (commented):
Thanks @FlashSystems it's very helpful to see the code for this, and it looks good. Once we've addressed the points below I'd be happy to see docs if you get a chance.

core/modules/filters/parsedate.js
FlashSystems: Currently TiddlyWiki's date format is well [defined and documented](https://tiddlywiki.com/#Date%20Fields). Extending parseDate would allow the ISO format to be used everywhere. That can lead to a mix of both date formats in fields. Most of the time this will work. But sometimes plugins or user code, that depends on the date being in the default format will break in hard to debug ways.

In my opinion it would be better to stick with a fixed and well defined date format and use a conversion function.

core/modules/editor/factory.js
FlashSystems: `setError` would be called on every update of the value anyway (maybe `updateValidationStatus` would be a good name). If I read `addClass`/`removeClass` correctly this will only lead to a DOM update if the value changes. I think it should be the task of the engine to optimize these calls if this is necessary.

Are DOM reads so bad that it would be worthwhile to have a cache (an additional variable) inside the engine to prevent these? I normally hesitate to do such things because having the same state duplicated in two places has bitten me more than once.

PS: I'll convert the `else if` into an `else`. Sorry for the badly executed last minute change.

core/modules/editor/engines/simple.js
```diff
@@ -81,6 +81,17 @@ SimpleEngine.prototype.setText = function(text,type) {
 	}
 };
 
+/*
+Set the error state of the input field
+*/
+SimpleEngine.prototype.setError = function(hasError) {
```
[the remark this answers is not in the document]
FlashSystems: Regarding the duplication of `setError`: I think the function belongs into the engine and it's a "coincidence" that the method is currently the same for both engines. Currently only engines deal with the DOM. And `setError` has to update the DOM to convey the validation (error) state to the user. I see your point regarding duplicate code, but I think in this case it's cleaner (and better for extensibility) to leave the duplicate implementation within the engine.
Could you live with that?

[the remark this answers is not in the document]
Jermolene: Thanks @FlashSystems that makes sense.

core/modules/editor/factory.js
Jermolene: In general we try hard to minimise the work in the refresh method. Here I think I'm just not getting the logic; the concern is that we appear to selectively engage the error mode but disengage it unconditionally.

core/modules/filters/parsedate.js
Jermolene: Thanks @FlashSystems I agree.

core/ui/EditTemplate/fields.tid
```diff
@@ -15,7 +15,9 @@ $:/config/EditTemplateFields/Visibility/$(currentField)$
 \end
 
 \define new-field-actions()
-<$action-sendmessage $message="tm-add-field" $name={{{ [<newFieldNameTiddler>get[text]] }}} $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
+<$set name="newFieldValue" tiddler=<<newFieldValueTiddler>> field={{{ [<newFieldNameTiddler>get[text]] }}} >
+<$action-sendmessage $message="tm-add-field" $name={{{ [<newFieldNameTiddler>get[text]] }}} $value=<<newFieldValue>>/>
```
saqimtiaz, line 19: Why do we use a `$set` widget for `newFieldValue` instead of a filtered transclusion in `$action-sendmessage`?

core/ui/EditTemplate/fields.tid
```diff
@@ -126,16 +132,20 @@ $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
 </div>
 </$reveal>
 </div>
+<$vars currentTiddler=<<newFieldValueTiddler>> currentField={{{ [<newFieldNameTiddler>get[text]] }}}>
```
saqimtiaz, line 135: We should use $let instead of $vars.

editions/tw5.com/tiddlers/concepts/Customizing Field Rendering.tid
```diff
@@ -0,0 +1,31 @@
+created: 20220110205600000
+modified: 20220110205600000
+tags: Concepts
+title: Customizing field rendering
```
saqimtiaz, line 4: It isn't clear that this is about how fields are rendered in the edit template.

editions/tw5.com/tiddlers/concepts/Customizing Field Rendering.tid
```diff
@@ -0,0 +1,31 @@
...
+type: text/vnd.tiddlywiki
+
+Tiddler fields are normally rendered as a simple input boxes. To modify this behaviour, a tiddler with the name of the field can be created below `$:/core/config/EditTemplateFields/Templates`. The content of this tiddler is transcluded to represent the content of this field.
```
saqimtiaz, line 7: Should this be `$:/core/config/EditTemplateFields/Templates` or `$:/config/EditTemplateFields/Templates` ?

saqimtiaz:
There are a few things that need attention apart from my comments in the code:

- CodeMirror engine needs to be extended with the same validation feature
- the new filter operator needs tests

core/ui/EditTemplate/fields.tid
pmario: hmmm, From my point of view $let is only needed if you want to use a variable in a later parameter ... $vars is much less code to execute and imo much more lightweight and should be more performant

saqimtiaz: > hmmm, From my point of view $let is only needed if you want to use a variable in a later parameter ... $vars is much less code to execute and imo much more lightweight and should be more performant

I don't believe there is any real performance difference and in fact we only have a new widget for backwards compatibility. $vars should be considered deprecated and slowly phased out from the core.

Thoughts @jermolene ?

Jermolene: That's correct – the purpose of introducing the `<$let>` widget was to be able to deprecate the `<$vars>` widget.

It's not a great idea to only use `<$let>` when wanting to use a variable in a later parameter because it makes the complexity and cognitive load for users so much higher if they have to consider which to use each time. Much clearer and simpler to have a single, consistent recommendation.

core/ui/EditTemplate/fields.tid
FlashSystems: Hello @saqimtiaz, I actually need your help here: The `field` attribute on this `$set` widget is already a filtered transclusion. I use the `$set` widget because I need to get the field, named by the `newFieldName`-Tiddler's `text` field, from the `newFieldValueTiddler`. I have no idea how I do this within a single filtered transclusion.

saqimtiaz: If I understand correctly this should do the trick: 
`[<newFieldValueTiddler>get{newFieldNameTiddler}]`

otherwise:

`[<newFieldNameTiddler>get[text]] :map[<newFieldValueTiddler>get<currentTiddler>]`

core/ui/EditTemplate/fields.tid
FlashSystems: Should I add an additional commit replacing all occurrences of `<$vars>` in `fields.tid` with `<$let>` for consistency or should I only fix this one occurence?

saqimtiaz: We need to do a systematic review at some point and replace $vars with $let, so perhaps restrict your changes to the new code.

FlashSystems:
Hello all,

I'm currently preparing the next revision of this pull request addressing all the issues you mentioned. While doing this, I've encountered a few things I'd like to discuss:

1. I'm not using the CodeMirror editor in my regular TW installation, so it is quite possible that I missed something: I was unable to get the `<$edit-text>` widget to use the CodeMirror editor. If that's possible, please tell me how I can do it, so I can test the `updateValidationStatus` function I put into the CodeMirror plugin. If it is impossible to use CodeMirror with the `<$edit-text>` widget, I think we should put a dummy implementation of `updateValidationStatus` there, because it will never be called with any other value than "true" for `isValid`, because `fromInputValue` can not be set.
2. As said in 1. my experience with CodeMirror is 0 for all practical purposes. Currently, I just set the `errorClass` on the `<div>` containing the CodeMirror editor if `isValid` is false. Could someone with more CodeMirror experience tell me, if that's the right thing to do?
3. The "input validation" feature behaves "interesting" if more than one way exist to update the field. I've updated the [demo](https://flashsystems.github.io/FieldTemplatesDemo/) to demonstrate this. If you enter an invalid value in the last edit-box (that is unrestricted) the restricted ones will show this value as valid (no red underline). This is counterintuitive at first, but intentional. The value is already stored within the field. The red underline tells the user, that the value he entered was invalid and was **not** saved to the field (the output of `fromInputValue` was empty). Therefore, I reset the validation status to "valid" every time the value is reloaded. Because now we show a value, that's already saved - even if it's invalid. What do you think about this behavior?

core/modules/editor/factory.js
FlashSystems: Hello @Jermolene, please take a look at the new `updateValidationStatus` code. I've split the code into a `setValidationStatus` function within `factory.js` that can be called unconditionally and an `updateValidationStatus` function within the engine that is only called by `setValidationStatus` if the validation status has changed.

The use of the `setValidationStatus` function keeps the code clean, centralizing the change detection and the update of the state variable into one location.

This should minimize the work that is done is `saveChanges` by only calling the (potentially costly) `updateValidationStatus` function if the validation status has changed.

As an additional benefit `setValidationStatus` checks if the engine implements `updateValidationStatus` and skips the call if the function is not implemented. That should make it backward compatible with plugins that supply an engine, that does not implement `updateValidationStatus`.

FlashSystems:
I've pushed a new version of this patch set. Besides fixing the issues mentioned in the review comments, I've fixed/changed the following things:

* Renamed the JavaScript properties `fromInputValueFilter`/`toInputValueFilter` into `editFromInputValueFilter`/`editToInputValueFilter` for consistency.
* Implemented change detection for the validation status to only call `updateValidationStatus` if a change append.
* Only call `updateValidationStatus` if the engine implements it. This should make it backward compatible.
* Implemented Tests for the `parseDate` filter.
* Fixed the documentation about customizing field rendering to clarify that this work only within the EditTemplate.
* Fixed some grammar and spelling errors.

FlashSystems:
After using TW with this patch for a few weeks, I've updated this pull request:

- Bugfix: Value for a new field was not used when pressing return to add the field.
- Rebased on master to ease merging.

@Jermolene, @saqimtiaz: Is there anything more I can do to get this merged?

saqimtiaz:
@FlashSystems apologies for the delayed response. 

_Can you please clarify the intended behaviour in the situation where an invalid value is entered into the input field? Is the underlying tiddler updated or not?_ 

Users currently expect that any value entered into an input field generated by an $edit-text widget is saved to the underlying tiddler with no exceptions. This PR - as per my reading of the code - would introduce an inconsistency here that is undesirable from a user experience perspective.

I believe the above is the real critical issue that needs resolution, however I do have two other concerns:
* The engines [invoke optional action strings on input](https://github.com/Jermolene/TiddlyWiki5/blob/fb0f78a62b99c9605428e5fe0cc39b02b2f77842/core/modules/editor/engines/framed.js#L217-L224) and these actions are passed the value of the input field as the variable `actionValue`. We would need to also pass the value as transformed by the filter as another variable, as well as the raw value, and document it.
* We aim for feature parity between the different engines: framed, simple and codemirror. In the edit template for example the `$edit` widget calls the appropriate `$edit-*` widget depending on the value of a config tiddler, see for example `$:/config/EditorTypeMappings/text/vnd.tiddlywiki`
  * The codemirror engine would need to be updated with the same validation features if we were to go ahead with these changes.  It can be invoked via the `$edit-codemirror` widget provided by the codemirror core plugin.

FlashSystems:
> Can you please clarify the intended behaviour in the situation where an invalid value is entered into the input field? Is the underlying tiddler updated or not?

I tried to describe this in the documentation: "The `fromInputValue` filter can be used on its own to validate the input value. This works by assigning a filter to `fromInputValue` that returns the original value or nothing. If nothing is returned, the value will __not__ be saved to the field and the `errorClass` CSS class will be assigned to the generated editing element."

I can add a warning directly to the [third paragraph of the documentation](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/b6af5926f24d436e4ca620ccbd9b9baa4e3d996a#diff-634305af0a24a513587c6a9ad06e2e2a5271bb7dc47e94837f3bd1b9dc013386R11) to make it more obvious.

> Users currently expect that any value entered into an input field generated by an $edit-text widget is saved to the underlying tiddler with no exceptions. This PR - as per my reading of the code - would introduce an inconsistency here that is undesirable from a user experience perspective.

If we implement any kind of transformation for user input, we have to accept, that it is not possible to guarantee that all input will be saved to the underlying tiddler:
* If the filter fails, we could save the raw value anyway. In this case, anything that depends on the value being processed by the filter will fail. Making the user experience much worse.
* If the filter does not fail, but just returns an empty value every time, nothing will get saved as well. This can't be prevented. It's impossible to decide if the output of the filter represents the input in any way or not.

>  The engines [invoke optional action strings on input](https://github.com/Jermolene/TiddlyWiki5/blob/fb0f78a62b99c9605428e5fe0cc39b02b2f77842/core/modules/editor/engines/framed.js#L217-L224) and these actions are passed the value of the input field as the variable `actionValue`. We would need to also pass the value as transformed by the filter as another variable, as well as the raw value, and document it.

I don't think that this is necessary. The [documentation](https://tiddlywiki.com/#EditTextWidget) states: "The variable `actionValue` is available to the `inputActions` and contains the value of the input field." There is no mention of the value stored within the tiddler. It only mentions the "input field". The `inputActions` are implemented directly on top of the `input`-DOM event. Integrating this with the update logic (to pass the transformed value) would be a big rewrite. I'd rather see this as an additional feature - if it is at all necessary.

> The codemirror engine would need to be updated with the same validation features if we were to go ahead with these changes. It can be invoked via the `$edit-codemirror` widget provided by the codemirror core plugin.

All right, I can comprehend that. As I stated before, I've no experience with CodeMirror and I would be grateful if someone could tell me what the `updateValidationStatus` method for CodeMirror should do. How can I show an "invalid input"-marker to the user using CodeMirror?

Jermolene:
Thanks @FlashSystems, and apologies for the delay in getting back to this. I've now had an opportunity to play with the code properly – it's a neat PR, and working with it has been a pleasure so far.

However, I have run into a couple of fundamental problems with the architecture. They both relate to the fact that widgets must not store any state that cannot be reconstructed from the tiddler store; this is a consequence of the fact that the refresh mechanism can require any widget to be removed and recreated.

The following example from the docs can be used to trigger the first problem:

```
<$set name="digits-only" value="^[0-9]*$">
<$edit-text tiddler="HelloThere" field="mynumbers" placeholder="Input only numbers" fromInputValue="[regexp<digits-only>]"/><br/>
</$set>
 
<$view tiddler="HelloThere" field=mynumbers/>
```

Putting that in a new tiddler and then typing `foo` into the text box results in the text being underlined to indicate a validation error.

Now open the browser developer console and execute the following code. It triggers a refresh of the whole page by updating the tiddler `$:/core/ui/PageTemplate`.

```
$tw.wiki.addTiddler(new $tw.Tiddler($tw.wiki.getTiddler("$:/core/ui/PageTemplate"),$tw.wiki.getModificationFields()))
```

The result is that the underlining is removed from the text box.

The second issue can be shown with a small modification to the code above. There's now a hard dependency on the value of the "modified" field of `$:/core/ui/PageTemplate`, which causes the entire subtree to be discarded and recreated:

```
<$set name="temp" value={{$:/core/ui/PageTemplate!!modified}}>
<$set name="digits-only" value="^[0-9]*$">
<$edit-text tiddler="HelloThere" field="mynumbers" placeholder="Input only numbers" fromInputValue="[regexp<digits-only>]"/><br/>
</$set>
</$set>
 
<$view tiddler="HelloThere" field=mynumbers/>
```

Now, type something like `123foo` into the text box, and then again execute the snippet in the developer console. The result is that the text in the text box is changed to `123`, to reflect the underlying field value.

I think that both of these problems could be fixed by adding a new attribute to specify a state tiddler in which the currently entered text is stored. But, I don't think that's a good user experience: having to come up with another state tiddler title is confusing.

I'd like to give it some more thought, and would welcome yours.

FlashSystems:
Hello @Jermolene,

the first finding is a genuine bug. I should only reset the validation state, if a new text is set. In `refresh` I do this unconditionally. Moving the `this.setValidationStatus(true)` call into the `else if` statements that really update the text, fixes this problem. I've pushed an update for this pull request.

The second issue is really a hard one. As I already noted in my answer to @saqimtiaz remarks: If we want arbitrary filtering, it is impossible to guarantee that a round trip through a tiddler field is even possible. I've created the following example to illustrate that:

```html
<$set name="temp" value={{$:/core/ui/PageTemplate!!modified}}>
<$edit-text tiddler="HelloThere" field="xforu" placeholder="Input only numbers" fromInputValue="[search-replace:g[x],[u]]"/><br/>
</$set>
 
<$view tiddler="HelloThere" field=xforu/>
```
It replaces every `x` with a `u` in `fromInputValue`. As soon as the widget is refreshed, all `x` within the input text change to `u`.
Even if we use a second tiddler as a cache for the value of the input element. How should we decide when the cached value should be used?

I had some Ideas for "solving" this issue. For example, we could make the transformation "write only", remove the `toInputValue` filter and put the filtered value into an additional tiddler field (that must be specified as a parameter). But everything I came up with so far, would make this feature much less useful or flexible.

Thank @Jermolene and @saqimtiaz for your input and for the interesting discussion. I'm looking forward to hearing what you think.

Jermolene:
Hi @FlashSystems 

> But everything I came up with so far, would make this feature much less useful or flexible.

It is a conundrum, and I do apologise for being the one to send you off down this rabbit hole. As I mentioned before, your work here is of an exemplary standard and so it is a pleasure to explore the problems with you, but I'm still acutely conscious of the frustration if we end up deciding to go in a different direction. The fundamental requirement of supporting the entire set of HTML 5 input controls remains very solid. The problem at hand is string transformation, and so filters still seem like a promising approach. I think we'll have to give it some more thought.

pmario:
>The fundamental requirement of supporting the entire set of HTML 5 input controls remains very solid. The problem at hand is string transformation, and so filters still seem like a promising approach. I think we'll have to give it some more thought.

I think the main problem we have here is, that our widget at the moment is EditText - ... The HTML element used is INPUT, which happens to [have 22 different types atm](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input). 

So depending on the input-type our users expect a slightly different behaviour. ... and validation is a completely different thing. 

We do have 3 types of input behaviour in TW at the moment:

 1. Editing a tiddler ... Which in reality edits the "Draft of 'tiddler'" ... so we have an undo feature
 2. Editing a $:/temp/tiddler ... which is copied over to the "real" tiddler with a button-action
 3. Naive editing: Directly write to a tiddler field. ... Which imo is the worst case scenario, but the one beginners intuitively use. 

IMO the PR mixes 2. and 3. at the moment. ... 

It's possible to directly "connect" to a tiddler field like in 3. but the validation mechanism "blocks" the input, if the "validation" fails, which is more like the behaviour in 2. where the "target tiddler" isn't changed until an action happens ... 

This leads to inconsistent behaviour, if the widget is refreshed ... @Jermolene  pointed out. 

--------

[MDN input-type: date](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date) has a lot of different elements that can be assigned to the date-picker. ... Most of them have not much to do with **input-type: text** ... 

The HTML validations are completely different between [date validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date#validation), [text validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/text#validation) and [number validation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#validation)

IMO if we want to have the full potential, we will need more different widgets, which give us access to the built-in HTML validations for **visual clues only**.

For **real validation** that is triggered or can trigger actions, there may be generic validators needed similar to `fromValue` / `toValue` as we have it in the PR now. .. BUT .. they probably have to be different for completely different types. eg type=color imo has  nothing to do with type=date. 

Just some thoughts. ... I'll post more once I did more experiments

FlashSystems:
Foremost, thank you, @Jermolene, @saqimtiaz and @pmario for your input. I appreciate that you work with me on this change and take the time to supply feedback.

Currently, my spare time is somewhat limited. But I'm committed to solving this puzzle.

My suggestion for going forward is splitting this pull request into three parts:
* A PR containing commit [9624c38](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/9624c3890bd1401c8f7e46edb77ed203f375fa5f), [5283f90](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/5283f902a0b171db8e90885a8a8e48b1f554e1a4) and [39a2468](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/39a24686ca88ec526f1524ca8bd754412dc1460d) implementing the customizable field rendering. This is of value by itself, because it allows fields to use drop-down-lists and checkboxes. Even date pickers can be used, but the date format within the field will be ISO.
* A PR containing commit  [95ecdde](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/95ecdde212d96033c603147862763908866d4b6f), [40edd59](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/40edd5951f8d1e13673c6d2977f3daca10f28c0b) and [a1257e9](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/a1257e909d2e71053b519b5d742474d3038d4009) that implement the `parseDate` filter. This filter allows using the date/time input type's date format.
* This PR will be left with commits [1ca1b06](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/1ca1b063153358897e55cdf35121b43336bef721), [7075c9f](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/7075c9fa550d08cc83c47d2ce6aee67d28ed1146) and [c71191c](https://github.com/Jermolene/TiddlyWiki5/pull/6377/commits/c71191cdb1d2433f3969ada7dba88abad16ace41) in it that extend the EditText widget. It needs more work and don't want to block the other two PRs. I think they supply value on their own.

What do you think? Should I extract the first two parts into their own PRs to get them merged? Then I can try to find a solution for the last one.

saqimtiaz:
@FlashSystems likewise, thank you for a well presented and thought out PR and ensuing discussion. I think splitting the PR into smaller pieces is definitely a good idea. It will allow a more considered review of the individual parts.

For instance, it is worth considering whether we want to introduce a new mechanism to customize field rendering as this PR does at present, or whether this might be a good use case for cascades which might be more flexible and also consistent with customizing other UI components.

FlashSystems:
I've extracted the two parts of this pull request that are independent of type coercion:

* #6511
* #6512

I've updated this PR to only contain the type coercion stuff. These are still the same commits and have the same problems. I've just removed the commits that belong to the other PRs.

FlashSystems:
Hello @Jermolene, @saqimtiaz, @pmario, I've been thinking about the problem with filtering the input of the EditText widget. As @pmario pointed out, TiddlyWiki was designed for three different editing models. Filtering the input, at the EditText widget level, isn't compatible with these editing models. Therefore, I went back to the drawing board and redesigned my approach.

@pmario wrote:
> Editing a $:/temp/tiddler ... which is copied over to the "real" tiddler with a button-action

This made me think: Couldn't we use this approach for filtering the input? I've created a new widget: The FieldFilter widget. This widget creates a temporary tiddler and sets it as the current tiddler for the (so called) inner context. It uses filters to pass the data for a given field from the inner to the outer context and vice versa. Instead of a button-action it monitors the inner and outer tiddlers field for updates. I've made a little drawing to clearify this:

FieldFilter

I also created another [demo](https://flashsystems.github.io/FieldTemplatesDemo/fieldfilter.html) for you to try this out.

This approach has multiple advantages in my opinion:

- It prevents the refresh problem by using a temporary tiddler to store the value of the edit field.
- It's much more versatile because it separates the translation logic from the edit widget.
- There is no need to modify the edit widget. This prevents any implications regarding the codemirror editor or other, custom editor types.

I'd like to get your feedback before I devote more time into making this into a proper PR.

pmario:
It looks promising to me. But I did manage to completely mess up my brain, by intentionally trying to break it ;)

The following code seems to work as intended and it is clear, what's going on. We need really good docs, otherwise our users will also mess it up. 

I used `asdf` tiddler `text` field and opened it in the story river, so I could see, what's going on
I am also showing a link to the temp tiddler and opened it to see what's in it

```
<$set name="digits" value="^[0-9]+$">
<$fieldfilter name=asdf field="text" toOuter="[regexp<digits>search-replace:g[9],[x]]" toInner="[search-replace:g[x],[9]]" state="StateTest">

	> This is inside the fieldfilter widget (inner context)<br/><br/>
	> <$edit-text  tag=input/><br/>Inner Contexts <<currentTiddler>> text field value: <$view />

</$fieldfilter>
<$reveal type="match" state="StateTest" text="failed">
<br/>Invalid input
<br/>This is revealed by the state widget specified for the fieldfilter widget.
</$reveal>
</$set>
```

With the existing code I did manage to crate several problems. .. 

 - There is an initialisation problem, when the `$:/temp/iofilter/xxx` tiddler doesn't exist. 
 - I did manage to create a "double" refresh, using wrong parameters
   - Not sure if I can reproduce that again :/ But I'll try ;)
 - `<$fieldfilter name=asdf` uses `name`  ... IMO it should use `tiddler`, which defaults to `currentTiddler` if no paramter is set
 - The js code itself uses some "forbidden" elements like `let`
 - I'm not 100% sure what `this.lastValue` does, but it seems to "manage state", which probably will cause problems. 
 - If I remove all numbers from the text input, the last value stays in "asdf", since empty value seems to be invalid
   - That's OK, but confusiong. So IMO the regexp needs to be adjusted, that "empty input" is valid -- may be?!

That's it for the moment. 
PS: I do like your drawing and the thinking behind it!

FlashSystems:
Hello @pmario, thank you for checking out my idea.

I've done some updates based on your feedback. The initialization problem and the double update should be fixed now. The parameter for the outer tiddler's name is now `tiddler`. The problem with empty values should be gone now, too.

Regarding the local state. This is necessary to prevent endless loops. I think I should go into more details here. It's hard to explain. Especially for a non-native speaker:

Let's build a really crazy field filter that appends an `i` on every call to the `inFilter` and an `o` on every call to the `outFilter`:
```
<$fieldfilter field="myfield" toOuter="[append[o]join[]]" toInner="[append[i]join[]]">
<$edit-text field="myfield"/>
</$fieldfilter>
```
This filter is not symmetric (outerField -- inFilter --> tempField -- outFilter --> **not** outerField). Without a little local state, this would lead to an endless loop. A change to the outer field will result in a change to the temp field. This will again result in a change to the outer field. Every call of the filters will add a new character, and it will loop forever. To break this loop, the widget tracks the value it has written to the outer/temp tiddler field and checks if the refresh call was because of this value. If it is, there is no need to call the filter again.
If we assume that the filters **are** symmetric, the call is unnecessary because the result would be the starting value. If the filter is **not** symmetric, the call must not be made to prevent an endless loop.
I hope this explanation helps to understand how the FieldFilter widget works and why some local state is necessary here. I had to think about it for some time to get it right. And there was still a bug in the version you tested🤦‍♂️.

pmario:
>I hope this explanation helps to understand how the FieldFilter widget works and why some local state is necessary here. I had to think about it for some time to get it right. And there was still a bug in the version you tested 

I wanted to point out my findings. If it looks "suspicious", there must be "something going on", which needs to be clarified.. Your explanation looks sensible to me. ... I'll have a closer look. ... 

We do have a similar "recursive transclusion protection" for templates. That's OK.

linonetwo:
I may need `<$edit-text field="startDate" tag="input" type="datetime-local"` in my [https://github.com/tiddly-gittly/tiddlywiki-calendar](https://github.com/tiddly-gittly/tiddlywiki-calendar) using your approach, @FlashSystems will you continue this?

FlashSystems:
Hello @linonetwo,

I'm sorry, but I think that this is currently a dead end. I'm inclined to close this PR.
The goal of this and some other pull requests was to make HTML5 date and time pickers usable with tiddly wiki. It is not as easy as setting the `type`  of the EditTextWidget to `datetime-local`. The date-format is not compatible with TiddlyWikis's date-format. Having two different date formats in TW5's fields would be the least intuitive solution and would make plugin development really hard.
Therefore, different solutions have been discussed and evaluated. I'm documenting them here for anyone who wants to pick up this endeavour. I'm currently out of ideas.

TL;DR: The only way that worked was extending the EditTextWidget, adding a special case for date/time pickers. All other solutions did not work reliably or were too complicated. Maybe this could be done with an additional widget that is a subclass of the `edit-text` widget.

# Modifying the edit-text widget
The first proposed solution was modifying the edit-text widget to detect the date types and convert the format on the fly. See https://github.com/Jermolene/TiddlyWiki5/discussions/6301#discussion-3721514
**Pros**: Simple to implement. Works transparently.
**Cons**: Not very flexible. Hardcoded translations that must be kept in sync when the W3C spec changes. The first approach used the deprecated field modules, but could also be implemented directly within the edit-text widget.
This approach was abandoned because it was not flexible enough. In hindsight, this was the only approach that really worked.

# Implement a fromValue/toValue filter

I tried implementing a fromValue/toValue filter for edit element. See https://github.com/Jermolene/TiddlyWiki5/discussions/6301#discussioncomment-1816642 and PR 6377
Pros: Flexible approach. Highly customizable.
Cons: Problems on refresh that I could not solve. It is not guaranteed that the filters are symmetric or produce any output at all. Easy to get wrong by the user. Tiddlywiki depends on the fact, that a value that is written into a widget can be read back later if it was not explicitly changed by the user.

# Wrapper widget

My last idea was to implement a wrapper widget that creates a temporary tiddler. This tiddler can be used as a backing store to solve the refresh problem.
Pros: Solves the refresh problem by storing the raw value in the temporary tiddler.
Cons: Creating a temporary tiddler is complex and if this solution is used in a loop with many tiddlers it will flood TW with temporary tiddlers. The management of temporary tiddlers is tricky. They must have unique but reproducible names, even in a loop. Cleaning up the temporary tiddlers is not possible because there is no way for a widget to know if it is permanently destroyed or just refreshed.
https://github.com/Jermolene/TiddlyWiki5/pull/6377#issuecomment-1072996526

Jermolene:
Thanks for the very clear summary @FlashSystems, and for all your work on this. It's a shame that we haven't managed to land this but I agree with your analysis that the only remaining viable approach seems to be an `<$edit-datetime>` widget.

pmario:
>TL;DR: The only way that worked was extending the EditTextWidget, adding a special case for date/time pickers. All other solutions did not work reliably or were too complicated. Maybe this could be done with an additional widget that is a subclass of the edit-text widget.

I think the idea to have extra widgets that will allow us to have "specialiced" HTML INPUT tags is a good idea. 

I did create a new colour-widget for colour inputs. It is a special type of "text-input", but there are "just enough" differences to cause severe problems if used with the edit-text - tag parameter. 

There is some dedicated code needed. So I think we should have widgets that do **one thing -- and do it well**

pmario:
The INPUT element can be used in 22 different ways. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

Not all of them make sense to be used with TW, but if we would like to have a consistent way to handle colour and date inputs we should have new dedicated widgets. 

Dedicated widgets would also allow us to "abstract" the browser native dialogues away. .. Eg the native color-dialogue for android browsers doesn't make any sense to be used with TW.

Similar things are true for date-inputs, which are very different for different browsers. Non of them create a good UX in a TiddlyWiki context

FlashSystems:
Is anyone opposed to closing this PR? It seems we came to the conclusion, that creating dedicated widgets is the way to go. I think these widgets will have their own PRs and I'd like to close this PR to reduce clutter.

Jermolene:
Thanks for all your help @FlashSystems

twMat:
A private discussion led me to look at this thread. @FlashSystems thingy really impressed me so the conclusion that "it should be done with dedicated widgets" is a bit discouraging considering that, well, is it actually going to be implemented at all when nobody is commited to it?

Jermolene:
Hi @twMat the problem here is not a lack of commitment. This PR represents many hours of work by @FlashSystems and the other contributors. The problem is that we have unsolved technical problems that we have been unable to surmount. See [Flashsystem's summary](https://github.com/Jermolene/TiddlyWiki5/pull/6377#issuecomment-1335634884) and [a comment of mine](https://github.com/Jermolene/TiddlyWiki5/pull/6377#issuecomment-1047988723) for more details.

twMat:
@Jermolene - thanks for your reply.
Just to be clear; I did in no way mean to diminish the previous commitment put into this, and your reply now also makes me understand that current "stalling" is due to technical difficulties rather than lack of commitment. Thank you all.

AnthonyMuscio:
I did make a solution using the input settings on the edit-text widget, it saves the date as a html date in a field and then uses an action to also set as tiddlywiki serial date. 

Perhaps this is a shorter path to addressing the requirements?

Was or should a new thread be started even just to explore further options?

pmario:
May be you can start a new discussion with a minimal test case so we can see what you described here.

PR #6381, opened 2022-01-03
Added sourceURL tags to boot js

flibbles:
These two lines will let browsers categorize the boot javascript into their own files, which is just a lot cleaner for anyone working with the browser debugger. Otherwise, methods in those files are listed under "(filename).html", and the shown source includes EVERYTHING, including CSS, the tiddler store, and those uglified encryption libraries. It makes it really slow. And it also makes line number stupid large.

This makes it cleaner.


Jermolene:
Excellent thank you @flibbles

Issue #6382, opened 2022-01-03
[BUG] Newly created tiddlers sort at the bottom of tag lists, by time created


sobjornstad:
**Describe the bug**: If no list information is available for a tiddler, that tiddler is listed in order of *creation* in tag pills (after all alphabetically sorted items) until the wiki is reloaded. [This isn't supposed to happen.](https://talk.tiddlywiki.org/t/are-tiddlers-supposed-to-sort-in-order-created-during-the-current-session/1989)

**To Reproduce**
1. Grab a copy of `empty.html` (or start in an existing wiki).
2. Create a tiddler `B` with tag `T`.
3. Create a tiddler `A` with tag `T`.
4. Click the tag pill for `T`.

**Expected behavior**: The tiddlers are listed in the order `A B`, since A comes alphabetically before B and there is no `list`, `list-before`, or `list-after` information field on any of the tiddlers.

**Actual behavior**: The tiddlers are listed in the order `B A`, by the order of their creation.

After saving and reloading the wiki, they're correctly listed in the order `A B`. If list information is provided (e.g., by setting `list-after` on tiddler A to `B`), then they're displayed in the correct order. Removing the list information without reloading the wiki restores the incorrect order.

The problem occurs regardless of whether the tag tiddler `T` exists.

**TiddlyWiki Configuration (please complete the following information):**
 - Version: v5.2.1
 - Saving mechanism: Tested on Node.js and the download saver.
 - Plugins installed: None required to exhibit the behavior (except core, obviously)
 
**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Tested Chromium and Vivaldi
 - Version: All recent versions (currently on Chromium 96)

Jermolene:
Thanks @sobjornstad I've created a draft PR with a failing test which I *think* corresponds to the issue – see #6383

sobjornstad:
Looks right to me.

pmario:
@sobjornstad -- Is the OP resolved? If yes please close.

There have been commits that should fix the problem.

pmario:
@Jermolene -- IMO it seems this bug is fixed already - please close

Jermolene:
Thanks @pmario @sobjornstad

PR #6383, opened 2022-01-03
Fix for tag ordering issue #6382

Jermolene:
A failing test for fixing #6382

[UPDATE 4th Jan 2022](https://github.com/Jermolene/TiddlyWiki5/pull/6383#issuecomment-1004640078) – Added a naive fix for the failing test, which breaks 121 other tests

[UPDATE 8th Jan 2022] The fix is now pretty reasonable. It ensures that tiddler ordering behaviour always matches that of a freshly loaded wiki.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -75,10 +75,10 @@ $tw.utils.insertSortedArray = function(array,value) {
 	var low = 0, high = array.length - 1, mid, cmp;
 	while(low <= high) {
 		mid = (low + high) >> 1;
-		cmp = array[mid];
-		if(value > cmp) {
+		cmp = value.localeCompare(array[mid]);
+		if(cmp > 0) {
 			low = mid + 1;
-		} else if(value < cmp) {
+		} else if(cmp < 0) {
 			high = mid - 1;
 		} else {
 			return array;
@@ -1101,7 +1101,7 @@ $tw.Wiki = function(options) {
 		tiddlerTitles = null, // Array of tiddler titles
 		getTiddlerTitles = function() {
 			if(!tiddlerTitles) {
-				tiddlerTitles = Object.keys(tiddlers).sort();
+				tiddlerTitles = Object.keys(tiddlers).sort(function(a,b) {return a.localeCompare(b);});
 			}
 			return tiddlerTitles;
 		},
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1111,7 +1111,7 @@ $tw.Wiki = function(options) {
 		shadowTiddlerTitles = null,
 		getShadowTiddlerTitles = function() {
 			if(!shadowTiddlerTitles) {
-				shadowTiddlerTitles = Object.keys(shadowTiddlers).sort();
+				shadowTiddlerTitles = Object.keys(shadowTiddlers);
 			}
 			return shadowTiddlerTitles;
 		},
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -69,6 +69,26 @@ $tw.utils.isArrayEqual = function(array1,array2) {
 };
 
 /*
+Add an entry to a sorted array if it doesn't already exist, while maintaining the sort order
+*/
+$tw.utils.insertSortedArray = function(array,value) {
+	var low = 0, high = array.length - 1, mid, cmp;
+	while(low <= high) {
+		mid = (low + high) >> 1;
+		cmp = array[mid];
+		if(value > cmp) {
+			low = mid + 1;
+		} else if(value < cmp) {
+			high = mid - 1;
+		} else {
+			return array;
+		}
+	}
+	array.splice(low,0,value);
+	return array;
+};
+
+/*
 Push entries onto an array, removing them first if they already exist in the array
 	array: array to modify (assumed to be free of duplicates)
 	value: a single value to push or an array of values to push
@@ -1134,10 +1154,8 @@ $tw.Wiki = function(options) {
 				}
 				// Save the new tiddler
 				tiddlers[title] = tiddler;
-				// Check we've got it's title
-				if(tiddlerTitles && tiddlerTitles.indexOf(title) === -1) {
-					tiddlerTitles.push(title);
-				}
+				// Check we've got the title
+				tiddlerTitles = $tw.utils.insertSortedArray(tiddlerTitles || [],title);
 				// Record the new tiddler state
 				updateDescriptor["new"] = {
 					tiddler: tiddler,
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1081,7 +1081,7 @@ $tw.Wiki = function(options) {
 		tiddlerTitles = null, // Array of tiddler titles
 		getTiddlerTitles = function() {
 			if(!tiddlerTitles) {
-				tiddlerTitles = Object.keys(tiddlers);
+				tiddlerTitles = Object.keys(tiddlers).sort();
 			}
 			return tiddlerTitles;
 		},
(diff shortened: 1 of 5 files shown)


Jermolene:
The underlying issue is that the wiki methods that iterate over tiddlers do so in order of insertion into the store. The last commit 51175616d37275c402a7a6f14a5e16d925f4a794 fixes the failing tests introduced in 32f79d16c680f1cf8c8a38b6bd2305bb2c961ba5, but it also breaks approximately 121 other tests. Examining a sampling shows that the tests currently have a dependency on the insertion order.

The fix in 51175616d37275c402a7a6f14a5e16d925f4a794 is naive – presumably it would be more efficient to maintain a sorted array of titles in addition to the store object itself – but it is easy to reason about, and gives us a sound basis for fixing the 121 affected tests.

I'd welcome other thoughts on this – @saqimtiaz @joshuafontany @clutterstack @flibbles

Jermolene:
Oops the fix needs more work – we're still pushing new tiddlers to the top of the previously sorted titles arrays.

Jermolene:
I've updated the fix to maintain the sorted order of the array. Now there are 122 other tests failing.

flibbles:
Wait, tag ordering isn't the same as tiddler store ordering.

Tags order themselves based on the tag's "list" field, and the tiddlers' declaration of "list-after" and "list-before". Outside of those directives, I always thought that the tag ordering was undefined. Sorting the underlying tiddler store would have much more of an effect (not to mention introduce performance overhead considering how often wiki.getTiddlerTitles is called. Wouldn't it be better to...

1. Just accept that tag ordering is undefined outside of "list", "list-after", and "list-before"?
2. Or have the tag indexer keep its internal lists sorted whenever it's updated?

flibbles:
I would think option 1 would be better. If someone actually cared about tag ordering, they'd use those directives, or they'd just add [sort[]] to their filters.

flibbles:
> In terms of performance, there's actually no impact on wiki.getTiddlerTitles because we only do the sorting once, and thereafter the burden falls on adding new tiddlers which require the binary search to find the correct insertion point. Tiddler creation is relatively rare (it can't affect the refresh performance, for instance).

Oh, it will? It looked like in https://github.com/Jermolene/TiddlyWiki5/commit/51175616d37275c402a7a6f14a5e16d925f4a794 that it was sorting Object.keys()'s return results every time. If that was just a naive approach, and we just sort whenever we add new tiddlers, then I think this may actually be preferable. I think Object.keys() allocates a new array every time, whereas if we kept a sorted array around, we could just return that, and it'll be more time and memory efficient.

flibbles:
Oh, I see what's happening. I didn't _see more context_. We're populating tiddlerTitles on request and caching it, and then purging it after any tiddler changes. I guess that works too.

Pity. I've written all my plugins under the assumption that [all[]] and [tag[]] returned an undefined order. Now they'll all be sorting unnecessarily.

Jermolene:
Hi @flibbles 

> This may cause plugin tests for many developers to break as well (myself included).

Indeed, it smashes a wrecking ball through the core tests too. But I find it hard to defend the current behaviour because it is so baffling for end users.

clutterstack:
> I've updated the fix to maintain the sorted order of the array. Now there are 122 other tests failing.

Github doesn't have enough emojis.

I have some fatigue in the way of my powers of comprehension at the moment, but if I'm understanding correctly:

* there's a mechanism that sorts tiddlers alphabetically in tag-pill lists, except for newly-created ones, which show up in order of creation until reload (disregarding `list-*`)
* this is actually only one example of how the order of filter results must actually be changing on reload?
* alphabetical order is preferred
* this can be enforced on tiddler creation by sorting the tiddlers in the store or keeping a sorted array of tiddler titles
* should not be a performance hit to do the latter
* but currently 122 tests depend on tiddlers being in the store in their order of insertion (does this indicate widespread dependency on this order in core, or is it limited to tests?)
* ? Any changes to plugins or unit tests that break would likely be limited to adding a sort() to some filters?

How it would affect me if you change this

* hahahaha I have no unit tests
* possibly some plugins I have written will break in some way I didn't know I was depending on that order
* that's fine. Only I use them, and they're always in some state of brokenness.

After all that: I don't have strong feelings on the approach. :)

It's rather pedal-to-the-metal at my new job or I would feel more comfortable committing to help adapt tests (or whatever they're indicating might need fixing in core, though my JS is rudimentary) if you wanted to press on.

Jermolene:
Thanks @clutterstack 

> I have some fatigue in the way of my powers of comprehension at the moment, but if I'm understanding correctly:

That's right.

> * but currently 122 tests depend on tiddlers being in the store in their order of insertion (does this indicate widespread dependency on this order in core, or is it limited to tests?)

It's just the tests. They create tiddlers in more or less random order, and then run filters to test the results. Very often the ordering reflects the default insertion order.

> * ? Any changes to plugins or unit tests that break would likely be limited to adding a sort() to some filters?

Yes, that will work for a lot of the tests. For the tests that are actually sensitive to ordering one would have to change the expected result that is being tested against.

> * possibly some plugins I have written will break in some way I didn't know I was depending on that order

If so, your users will likely already be seeing different results in a freshly loaded wiki vs. one in which they have dynamically created tiddlers.

> It's rather pedal-to-the-metal at my new job or I would feel more comfortable committing to help adapt tests (or whatever they're indicating might need fixing in core, though my JS is rudimentary) if you wanted to press on.

Thank you! One thing that we're going to need help with is explaining the change in a way that will make sense to end users.

Jermolene:
I've just pushed fixes for the failing tests.

In most cases, the underlying insertion order was showing through in the expected test results. Some of the cases are subtle – for example, the `each` and `eachday` operator are dependent on the store ordering.

I'd welcome another set of eyes on these.

Jermolene:
I've found another unexpected consequence of these changes: stylesheets are now sorted differently

<details>
<summary>Screenshots of the differences</summary>

The current prerelease:

Screenshot 2022-01-08 at 09 39 13

This branch:

Screenshot 2022-01-08 at 09 38 56

</detail>

A couple of thoughts while I give this more consideration:

* Changing stylesheet ordering is not acceptable, and will need to be fixed
* This may invalidate some of the changes to the expected test results in ee03ee57f59aa3a21a518c0190acbdd060497a49

saqimtiaz:
It suggest checking whether this is also impacts ordering of templates/cascades.

flibbles:
I haven't checked, but it's my understanding that stylesheet ordering was originally determined by plugin priority, THEN alphabetical order. It wouldn't be too hard to change the core to sort by plugin priority. It might make sense to do that too.

Otherwise there's the pre-existing problem where someone makes a stylesheet tiddler and gets the results they want, but then they reload, and because their tiddler now sorts before other plugins, the css applies differently.

Jermolene:
I've improved the fix to avoid the problem of the stylesheets being reordered. 

The order of tiddlers in the HTML file is sorted using localeCompare(), and that determines the insertion order in previous versions. So if we want to be compatible with older versions we have to use localeCompare() to order tiddlers, not a plain sort().

I've also removed the sorting of the shadow tiddlers; we should instead use the ordering determined by the plugin unpacking process.

We're now at 92 failing tests, which will need careful review.

Jermolene:
Hi @flibbles 

> I haven't checked, but it's my understanding that stylesheet ordering was originally determined by plugin priority, THEN alphabetical order. It wouldn't be too hard to change the core to sort by plugin priority. It might make sense to do that too.

That's correct, and that behaviour is now retained in this PR.

> Otherwise there's the pre-existing problem where someone makes a stylesheet tiddler and gets the results they want, but then they reload, and because their tiddler now sorts before other plugins, the css applies differently.

That problem should now be fixed in this PR.

PR #6388, opened 2022-01-06
Add support for directly setting style.* attributes on HTML elements

Jermolene:
## Introduction

It is currently possible to use the standard HTML "style" attribute to add styling to an HTML element:

```
<div style="background:red;">
```

It is proposed to add the ability to set individual style properties directly:

```
<div style.background="red">
```

Note that this extension is not part of the HTML standard.

This improvement makes it easier to use computed CSS property values:

```
<div style.background={{!!colour}}>
```

Instead of:

```
<div style={{{ [{!!colour}addprefix[background:]addsuffix[;]] }}}>
```

## Style property names

The present implementation accepts both the camelcase and dashed versions of CSS property names. Thus the following two examples will have the same effect:

```
<div style.backgroundColor="red">
<div style.background-color="red">
```

## Attribute ordering

There is a potential ambiguity in cases where the `style` attribute is specified alongside a direct style attribute. This is resolved by processing the attributes in the order that they appear. In the following example, the result is to set the background to blue and the foreground to green:

```
<div style="background:red;color:green;" style.background="blue">
```

Meanwhile, in this example, the background will be red and the foreground green:

```
<div style.background="blue" style="background:red;color:green;">
```

## Progress

* [x] Add support for direct style attributes to HTML element widget
* [ ] Add support for direct style attributes to other widgets

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -314,21 +314,28 @@ excludeEventAttributes: ignores attributes whose name begins with "on"
 Widget.prototype.assignAttributes = function(domNode,options) {
 	options = options || {};
 	var self = this;
-	$tw.utils.each(this.attributes,function(v,a) {
-		// Check exclusions
-		if(options.excludeEventAttributes && a.substr(0,2) === "on") {
-			v = undefined;
+	$tw.utils.each(this.parseTreeNode.orderedAttributes,function(attribute,index) {
+		var name = attribute.name, value = self.getAttribute(name);
+		// Check for excluded attribute names
+		if(options.excludeEventAttributes && name.substr(0,2) === "on") {
+			value = undefined;
 		}
-		if(v !== undefined) {
-			var b = a.split(":");
-			// Setting certain attributes can cause a DOM error (eg xmlns on the svg element)
-			try {
-				if (b.length == 2 && b[0] == "xlink"){
-					domNode.setAttributeNS("http://www.w3.org/1999/xlink",b[1],v);
-				} else {
-					domNode.setAttributeNS(null,a,v);
+		if(value !== undefined) {
+			// Handle the xlink: namespace
+			var namespace = null;
+			if(name.substr(0,6) === "xlink:" && name.length > 6) {
+				namespace = "http://www.w3.org/1999/xlink";
+				name = name.substr(6);
+			}
+			// Handle styles
+			if(name.substr(0,6) === "style." && name.length > 6) {
+				domNode.style[$tw.utils.unHyphenateCss(name.substr(6))] = value;
+			} else {
+				// Setting certain attributes can cause a DOM error (eg xmlns on the svg element)
+				try {
+					domNode.setAttributeNS(namespace,name,value);
+				} catch(e) {
 				}
-			} catch(e) {
 			}
 		}
 	});
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -341,11 +341,11 @@ Widget.prototype.assignAttributes = function(domNode,options) {
 	// Not all parse tree nodes have the orderedAttributes property
 	if(this.parseTreeNode.orderedAttributes) {
 		$tw.utils.each(this.parseTreeNode.orderedAttributes,function(attribute,index) {
-			assignAttribute(attribute.name,self.getAttribute(attribute.name));
+			assignAttribute(attribute.name,self.attributes[attribute.name]);
 		});	
 	} else {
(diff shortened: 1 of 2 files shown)


pmario:
I don't understand, why this is needed. We can use classes to style HTML elements. Hardcoded styles make theming extremely difficult, since the style attribute will overwrite every CSS setting. 

Most browsers support [CSS variables](https://caniuse.com/?search=css%20variables) now. We don't use any of those possibilities in the core yet. I think they are much more performant then using wikitext CSS variables, since they are optimized by the browser vendors. 

I can't find the proposed syntax in the HTML spec. Can you post a link?

Jermolene:
> I don't understand, why this is needed. We can use classes to style HTML elements. Hardcoded styles make theming extremely difficult, since the style attribute will overwrite every CSS setting.

There are several use cases:

* The new syntax is more concise and clear
* To make it easier to use computed values for styles (it's often not practical to use classes)
* To make it possible to trigger CSS animations (we have to be able to reset individual styles in order to animate them)

> Most browsers support [CSS variables](https://caniuse.com/?search=css%20variables) now. We don't use any of those possibilities in the core yet. I think they are much more performant then using wikitext CSS variables, since they are optimized by the browser vendors.

But to be useful we still need a way to inject wikitext values into those variables, which takes us back to stylesheets generated from wikitext.

> I can't find the proposed syntax in the HTML spec. Can you post a link?

The proposed syntax is not part of the HTML spec. I've clarified the OP.

saqimtiaz:
> The proposed syntax is not part of the HTML spec. I've clarified the OP.

I actually consider this a plus as it makes it very clear that something wikitext specific is going on. Our syntax for HTML is already not standards compliant due to the support for assigning attributes via variables and transclusions. 

I think a lot of the misunderstandings that lead users to expect things like the following to work, arise because they don't realize that HTML tags are implemented by widgets: `<a href="http://mysite/{{!!title}}">{{!!title]]</a>`. 
Therefore `<a href={{!!title}}>` looks to them like arbitrary concatenation and no different than the prior erroneous example.

twMat:
I know of a lot of changes for "styling" that I'd love to see but like pmario I'm not convinced about this idea. Maybe it is a matter of getting used to.

Q: Would this only be useful for setting a single style attribute in the html tag? Or could things be chained somehow?

...

(*"...they don't realize that HTML tags are implemented by widgets"* - Say whaaaaat!? No, they certainly had no clue about this!!! )

Jermolene:
Here's an example of why this change is so useful. Consider this fragment of wikitext, showing one way to assign computed CSS properties to a DIV:

```
\define lenticulate-image-style()
color: $(color)$;
left: $(x)$px;
top: $(y)$px;
\end

<$let
	color={{{ [<myvar>get[color]] }}}
	left={{{ [<myvar>get[left]divide<myfactor>] }}}
	top={{{ [<myvar>get[top]divide<myfactor>] }}}
>
	<div style=<<lenticulate-image-style>>>
	...
	</div>
</$let>
```

The trouble with that implementation is that the `<$let>` widget gets refreshed if the value of any of its variables changes, which means that the div gets entirely deleted and recreated, which stops CSS animations working.

The alternative with the new syntax avoids that problem and is more concise:

```
<div
	style.color={{{ [<myvar>get[color]] }}}
	style.left={{{ [<myvar>get[left]divide<myfactor>addsuffix[px]] }}}
	style.top={{{ [<myvar>get[top]divide<myfactor>addsuffix[px]] }}}
>
...
</div>
```

saqimtiaz:
Worth noting that the ability to specific style attributes already exist, this change just makes it much less cumbersome to do so. It is not suggested that inline styles be used instead of classes for styling, but for certain situations such as animations they are needed.

CodaCodr:
Very nice - I like this a lot.

Wondering if this might be considered:

```
<div
  with(style){
    color={{{ [<myvar>get[color]] }}}
    left={{{ [<myvar>get[left]divide<myfactor>addsuffix[px]] }}}
    top={{{ [<myvar>get[top]divide<myfactor>addsuffix[px]] }}}
  }
>
...
</div>
```
Obviously possible, but desirable? If yes, what about...

```
<div
  with(data){
    id=foo
    something=bar
  }
>
...
</div>

<!-- which produces ... -->

<div data-id=foo  data-something=bar> ...<div>
```
Probably pushing the boat out too far -- and no, I don't have a use-case.

Jermolene:
Thanks @CodaCodr 

> Wondering if this might be considered:

I'm not sure if it's worth it. Generally speaking, the widget format trades brevity for clarity (while the wikitext syntax forms are intended to optimise for ease of typing)

PR #6402, opened 2022-01-16
Compiled filters are cached

flibbles:
I'm having some real slowdown problems with my large projects, and [I'm not alone](https://groups.google.com/g/tiddlywiki/c/dEhq2V4ZYzg/m/vfMnTUqiAQAJ?utm_medium=email&utm_source=footer).

~95% of all filter usage is done through filterTiddlers(), which means we're constantly recompiling filters just to run them once. So why don't we cache the compiled filters? They are derived solely from their filterString, so we can cache them based on that.

I did ran some profiling. I went to tw5.com, opened 20 tiddlers, then I edited and saved one of them. This called "filterTiddlers()" 2960 times. I captured the arguments for all of those calls and then ran them independently as the test.

Screen Shot 2022-01-15 at 11 20 03 PM

I ran this four time without my change:

Avg Runtime: 329 ms
Avg time in compileFilter(): 40 ms
Avg time Garbage Collecting: 28 ms

And with my change:

Avg Runtime: 208 ms
Avg time in compileFilter(): 1.0 ms
Avg time Garbage Collecting: 15 ms

I think most of those gains are in memory improvement. On tw5.com, after you've visted a ton of pages and all the tabs, the filterCache levels out at around 200 cached filters. Here's a memory snapshot of it.
Screen Shot 2022-01-15 at 11 27 15 PM

288K retained for 200 filters. THAT may be where the gains are, because if we were calling compileFilter 2960 times before, that's about 4.2Mb that we were allocating with *each refresh* and then just tossing away. Time lost to memory management is hard to track, but even if we ignore this, Reducing compileFilter down to a negligible runtime cuts about 10% of the time off running filters in tw5.com.

@Jermolene, @pmario? Can anyone think of a reason why this wouldn't work?

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -218,6 +218,13 @@ source: an iterator function for the source tiddlers, called source(iterator), w
 widget: an optional widget node for retrieving the current tiddler etc.
 */
 exports.compileFilter = function(filterString) {
+	if(!this.filterCache) {
+		this.filterCache = Object.create(null);
+		this.filterCacheCount = 0;
+	}
+	if(this.filterCache[filterString] !== undefined) {
+		return this.filterCache[filterString];
+	}
 	var filterParseTree;
 	try {
 		filterParseTree = this.parseFilter(filterString);
@@ -318,7 +325,7 @@ exports.compileFilter = function(filterString) {
 		})());
 	});
 	// Return a function that applies the operations to a source iterator of tiddler titles
-	return $tw.perf.measure("filter: " + filterString,function filterFunction(source,widget) {
+	var compiled = $tw.perf.measure("filter: " + filterString,function filterFunction(source,widget) {
 		if(!source) {
 			source = self.each;
 		} else if(typeof source === "object") { // Array or hashmap
@@ -333,6 +340,19 @@ exports.compileFilter = function(filterString) {
 		});
 		return results.toArray();
 	});
+	if(this.filterCacheCount >= 5000) {
+		// I doubt anyone will come close to 5000 cached filters (~7 Mb)
+		// but if we do, it's because something is dynamically creating
+		// filters, in which case, letting the cache grow indefinitely
+		// is a memory leak. The simplest solution is to just clear it
+		// if it gets too large. This'll result in a recaching every
+		// minute or so, which is fine.
+		this.filterCache = Object.create(null);
+		this.filterCacheCount = 0;
+	}
+	this.filterCache[filterString] = compiled;
+	this.filterCacheCount++;
+	return compiled;
 };
 
 })();
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -229,6 +229,7 @@ exports.compileFilter = function(filterString) {
 	try {
 		filterParseTree = this.parseFilter(filterString);
 	} catch(e) {
+		// We do not cache this result, so it adjusts along with localization changes
 		return function(source,widget) {
 			return [$tw.language.getString("Error/Filter") + ": " + e];
 		};
@@ -340,13 +341,10 @@ exports.compileFilter = function(filterString) {
 		});
 		return results.toArray();
 	});
-	if(this.filterCacheCount >= 5000) {
-		// I doubt anyone will come close to 5000 cached filters (~7 Mb)
-		// but if we do, it's because something is dynamically creating
-		// filters, in which case, letting the cache grow indefinitely
-		// is a memory leak. The simplest solution is to just clear it
-		// if it gets too large. This'll result in a recaching every
-		// minute or so, which is fine.
+	if(this.filterCacheCount >= 2000) {
+		// To prevent memory leak, we maintain an upper limit for cache size.
+		// Reset if exceeded. This should give us 95% of the benefit
+		// that no cache limit would give us.
 		this.filterCache = Object.create(null);
 		this.filterCacheCount = 0;
 	}


pmario:
I'm not sure, how you measure your times?

I did my measurements with `$tw.perf.log()` and it seems there is an improvement of about 30% in filter `totalTime` between the 2 versioins, which imo is OK to use it. 

I don't see a reason, why the cache couldn't be used in a way as it is implemented.

flibbles:
> I'm not sure, how you measure your times?

I used the Chrome profiler to get my times. To this day, I've yet to figure out how to easily activate that perf log thing. Actually, one of my main concerns about this change is that it may make activating that log difficult after the wiki is loaded. All the filters will be cached with it off.

Two design choices I expect people may take issue with.

* My magic number 5000. I chose that more out of intuition than any logic. Should that be in some configurable location?
* Do we want this filterCache accessible by anything like it currently is? Would it be better to put something in boot.js so it can be hidden away and only reachable through a "getCacheForFilter" method or something?

pmario:
>My magic number 5000. I chose that more out of intuition than any logic. Should that be in some configurable location?

Hmm, At first I also thought it would be a problem. .. Then I did activate some tiddlers especially activating all tabs that I know of. I got 204 cached filters ... So I think the limit will be high enough. 

>Should that be in some configurable location?

I don't think so. Accessing a tiddler and reading the actual value eveytime you hit the function will probably eat the advantage away.

flibbles:
It occurred to me that when people type in filters into the "advanced search" Tiddler, it'll might actually fill up the cache. I still think this optimization is totally worth it, but the magic number perhaps should be smaller. Dunno. I guess it's about how you all feel about 0.3mb - 8mb being reserved, or 0.3mb - 1mb being reserved. Kinda just a preference really.

I think a magic number of 1000 would probably be enough too.

pmario:
>It occurred to me that when people type in filters into the "advanced search" Tiddler, 

IMO that's not a problem. Only valid filters will be cached.

Jermolene:
Thank you @flibbles. I've also wondered about caching filter compilation. The two obvious issues are the cache invalidation strategy, and concerns over whether any of the existing code might inadvertently mutate filter parse trees.

The simple strategy you're using here does seem attractive, but as you know it's proven very hard to reason reliably about such things. We might mitigate the risks by including diagnostics.

The concern about existing code is hard to assess. I've learned painfully not to make assumptions about how TiddlyWiki is used in the wild.

Perhaps we might try to make a "turbo" plugin for experimental speedups for large wikis, where we can test things out before rolling them into the core.

flibbles:
I would have figured that the "turbo" plugin is what we already call the "prerelease" build. Although this change could quite easily wrap the compileFilter method from a plugin, I don't know how you'd get a useful number of people to actually install it and keep it up to date.

I understand that this PR seems dangerous, I do contend that it's safe The filter parse trees cannot be accessed through the "compileFilter" method. It's locally scoped, and only a method is returned. The only two points of danger that I see are...

* Someone manipulates the parseFilter method to create something that's state dependent somehow. Don't know HOW, but such a thing would have to be through monkey patching, and would really redefine what filters are.
* Someone might start manipulating the compiledFilterCache. This isn't a back-compatibility issue, but it could be in the future if it isn't wrapped better. I would be *all for* wrapping the filter cache in a way that makes it directly untouchable (sort of like the tiddler stores). I'm just not sure how I'd do that without altering the boot.js file. If we're able to assume that no two wikis will have a different set of modules, then we could actually just make the cache a global untouchable variable inside file.

pmario:
>.. Someone might start manipulating the compiledFilterCache.

We do have the same problem with other existing caches eg: `$tw.wiki.caches`. I think easy access makes the very useful for debugging purposes. ... IMO all of them can be modified with js-code. ... But I think no developer is "crazy" enough to "patch" existing cache-values and assumes they still work.

pmario:
>I would have figured that the "turbo" plugin is what we already call the "prerelease" build. Although this change could quite easily wrap the compileFilter method from a plugin, I don't know how you'd get a useful number of people to actually install it and keep it up to date.

I think it could be a system-plugin and that plugin could be added to the "prerelease" by default. So everyone who uses the prerelease would automatically "test" it. ... and IF there is a problem these users could simple remove the plugin and get the "default filter" handling.

Jermolene:
Thanks @flibbles 

> I would have figured that the "turbo" plugin is what we already call the "prerelease" build

The prerelease is, by definition, always ready for release (modulo any housekeeping backlog with the release notes etc). If there was an urgent need for a new release (eg another change to Chrome) it's important that we can make a release quickly. That means that we generally don't merge changes into `master` unless they are ready for immediate release.

flibbles:
So @Jermolene, if I make a system plugin called "tiddlywiki/turbo", and I put this change on there and change the default tiddlywiki.info to include it, that would make this okay to merge?

I apologize for being somewhat pushy on this one, but the thing is my current Tiddlywiki project has about 8000 tiddlers now, and it has become so slow that working on TiddlyWiki has become tedious. And I agree with your earlier assessment that the problem is memory management.

This optimization is just the first I'll be proposing. I'm probably going to be pushing for lazy filters again once I've confirmed that there's improvement to be had there. Could those go on turbo too?

flibbles:
An alternative might also be a git branch with these changes. Keeping all optimizations (such as lazy filters) entirely isolated in a plugin wouldn't be impossible, but it might be ungainly in some cases. I still see value in a plugin though. Most people would never know such a git branch exists.

Jermolene:
> So @Jermolene, if I make a system plugin called "tiddlywiki/turbo", and I put this change on there and change the default tiddlywiki.info to include it, that would make this okay to merge?

We may want to give more thought to the name of the plugin – ideally it should also communicate riskiness, as well as raw speed.



> An alternative might also be a git branch with these changes. Keeping all optimizations (such as lazy filters) entirely isolated in a plugin wouldn't be impossible, but it might be ungainly in some cases. I still see value in a plugin though. Most people would never know such a git branch exists.

The advantage of a plugin is that end users can then easily enable/disable the optimisations to aid debugging if they run into difficulties. A separate branch would only be accessible to more sophisticated developers.

Jermolene:
> would make this okay to merge?

Sorry I didn't answer clearly: yes, I think so.

flibbles:
I thought about this all shower long.

tiddlywiki/optimize
tiddlywiki/revamp
tiddlywiki/finetune

These are my three ideas. My preference is tiddlywiki/optimize. It's clear that it's optional, about performance, and implies some risk (to developers anyway). Obviously these are just three spitballed ideas though. Choice is yours.

pmario:
I'd go with `tiddlywiki/filter-cache` or `tiddlywiki/cached-filters`

Jermolene:
> I'd go with `tiddlywiki/filter-cache` or `tiddlywiki/cached-filters`

I am proposing that we bundle these improvements with other enhancements that are helpful for extremely large wikis. If we took that approach so we'd want a name with a broader scope.

pmario:
may be `tiddlywiki/speedups`

Jermolene:
Hi @flibbles another alternative would be to publish this as a "flibbles" plugin. I think the priority is to get something together that Siniy-Kit and others can use and evaluate in the field.

pmario:
@Jermolene ... IMO if we want it to be tested, it needs to be part of the prerelease. Otherwise users will test it when it is part of the next "full version"

Jermolene:
> @Jermolene ... IMO if we want it to be tested, it needs to be part of the prerelease. Otherwise users will test it when it is part of the next "full version"

Eventually, yes, but not yet.

Jermolene:
@pmario see what I said above:

> The prerelease is, by definition, always ready for release (modulo any housekeeping backlog with the release notes etc). If there was an urgent need for a new release (eg another change to Chrome) it's important that we can make a release quickly. That means that we generally don't merge changes into master unless they are ready for immediate release.

In other words we do not merge things to master unless we're happy for them to be released.

flibbles:
So you want me to make my own 3rd-party "flibbles/speedups" plugin?

1. It won't contain any "features" which would incentivize someone to use it.
2. It won't be connected to core. People would have to go to my github page separate from TW core and download this.
2. All of its changes are implemented through monkey-patching.
3. It would be advertised as "Packaged separately because the core developers were concerned it might break stuff."

If we're looking to have these changes tested, this isn't it. The only tester for this stuff would be me (and I've already been using it for weeks now), and maybe @pmario because he might trust me enough to try my "3rd party optimization plugin". Oh, and maybe @kookma, but I can't imagine he's got TiddlyWikis with thousands of tiddlers which would warrant these optimizations.

Over the last few months, I've pushed nearly a dozen PRs that are now in limbo, and I'm starting to realize that maybe I have WAY more free time than you do, @Jermolene. And now you are asking me to go play in my own sandbox. If you'd like, I can throttle myself. All of the PRs I'm pushing are just to handle my own personal issues with TW which I know don't concern 99% of other TW users. I can keep my PRs to improvements which concern other people, not just myself.

kookma:
@flibbles - I really like your improvements, implementations! Thank you for all your efforts.

Jermolene:
> So you want me to make my own 3rd-party "flibbles/speedups" plugin?

No, I was raising it for discussion.

> 1. It won't contain any "features" which would incentivize someone to use it.

I think that the people running into performance issues are highly incentivised to try potential fixes.

> 2. It won't be connected to core. People would have to go to my github page separate from TW core and download this.
> 3. All of its changes are implemented through monkey-patching.
> 4. It would be advertised as "Packaged separately because the core developers were concerned it might break stuff."

Indeed there are disadvantages to that approach, as there are to minting a new core plugin.

> If we're looking to have these changes tested, this isn't it. The only tester for this stuff would be me (and I've already been using it for weeks now), and maybe @pmario because he might trust me enough to try my "3rd party optimization plugin". Oh, and maybe @kookma, but I can't imagine he's got TiddlyWikis with thousands of tiddlers which would warrant these optimizations.

I'd hope that Siniy-Kit would also be able to try them out. Beyond that we don't really know how many people would be interested.

> Over the last few months, I've pushed nearly a dozen PRs that are now in limbo, and I'm starting to realize that maybe I have WAY more free time than you do, @Jermolene.

I have zero free time! As you can imagine with a project like this, I have an infinite todo list, with most of the items put there by other people. It's frustrating not to be able to get everything done, but there is no amount of work I could put in that would be sufficient.

Hence why your contributions (and others) are so vital to the project.

> And now you are asking me to go play in my own sandbox.

I repeat that I am not asking you to do anything, and I apologise if it came across like. [My comment yesterday](https://github.com/Jermolene/TiddlyWiki5/pull/6402#issuecomment-1030806350) was simply to raise another option for proceeding.

> If you'd like, I can throttle myself. All of the PRs I'm pushing are just to handle my own personal issues with TW which I know don't concern 99% of other TW users. I can keep my PRs to improvements which concern other people, not just myself.

Please don't throttle, your work is of value, and I have learned a lot from your contributions.

flibbles:
I understand. Caution is important, but my challenge still stands.

Dynamic filters won't be the breaker though. If there is a plugin which does so, then the compiled-filter buffer will fill up to 5000, then reset back to zero, making all filters recompile. This will still result in major savings, because as it stands right now, all filters must constantly recompile anyway. If this buffer fill-up happens even once every second (which is unlikely), there are still countless filters which are run hundreds of times in a row, recompiling every time, which would no longer have to do so.

I'm just hoping my challenge incentivizes people to examine this change and satisfy all their cautious bones, because I'd _really love_ to see Tiddlywiki work faster when it's got 10,000+ tiddlers, for personal reasons.

pmario:
> The concern would then be the impact on devices with constrained memory.

The code contains a "magic number" of max 5000 stored filters, which may result in about 8MByte of additional memory. "flibbles" and my tests showed, that tiddlywiki.com uses about 200 different filters in normal use. And if the cache is reset, if it hits 5000 we should be save and 8MB isn't too much, even for low range devices. I'm sure the browser needs more memory showing 2 images on any given web page.

flibbles:
> and 8MB isn't too much, even for low range devices

Also, if that magic number were reduced to 1, we'd still see significant gains, I think. There are many filters which churn hard in a row. If we are concerned about memory still, I think bringing that number down to 500 would provide nearly identical performance improvement (~800 KB).

flibbles:
I'll be fair, and I'll point out the only two ways I could think of that this could break a 3rd party plugin.

1. That plugin overrides "wiki.parseFilter" to do something weird and dynamic.
2. That plugin manipulates "wiki.filterOperators" or "wiki.filterRunPrefixes", which are two methods which the filters.js file uses to cache all filter modules. If any plugin dynamically manipulates those, they'd have to do `wiki.filterCache = null` at the same time, or else their changes won't propagate.

But those are the only two ways, and since I said them first, no one else can claim them and say I have to fix a bug. I guess it's up to you all whether these two issues are too big to shoulder.

saqimtiaz:
I would like to see us move this forwards and to that end I have a few thoughts that might help clear up any misgivings:
* while the lookup in the cache for filters will almost surely be faster than recompiling the filter function even when the cache is more than half full, it would be helpful to have some performance metrics to back this up. I believe Jeremy has created a performance testing framework or wiki. Even some general non-TiddlyWiki specific metrics on how lookups by key in larger objects scales in modern day browsers would be helpful. I assume this is a non-issue but it would be helpful to have confirmation.
* Filters can return error strings that are localized. With caching in place, switching languages in a wiki will lead to error strings still being shown in the previous language until the cache is cleared. Not necessarily a deal breaker but worth being aware of and documenting.
* There are some long descriptive comments in the code that I assume were meant to facilitate discussion, can we clean those up please as part of cleaning this up for merging?

flibbles:
I'll do my best to clear them up.

> it would be helpful to have some performance metrics to back this up

Both @pmario and I have conducted performance tests (different kinds) and confirmed about a 30% improvement overall. I've performed a more general test today where I measure general usage on the tw5.com site as I click through all of the tabs on the "more" sidebar. Here are the results
Screen Shot 2022-05-17 at 12 48 15 PM
Caching is on the left. It seems like a pretty clear cut performance gain.

> Filters can return error strings that are localized.

No worries. Non issue. When filters return error strings, they either do it by _throwing them_, in which case they get caught, and they're never cached at all. Or they return a function which returns a localized string, in which case it will adjust accordingly. I have tested this today and studied the code. There are no cases (that I found) where caching can cause old localization results to linger.

> can we clean [the comments] up please as part of cleaning this up for merging?

Sure. Done. I've pushed a change where the comments are more long-term. I've reduce the cache size to 2000 also, just because I've long thought 5000 was excessive.

This pull request remains undefeated. It will fight on.

flibbles:
It seems to me that the only unaddressed concern people had with this is:

> concerns over whether any of the existing code might inadvertently mutate filter parse trees

So I looked into it, and can confirm this problem never happens. The only way that developers ever even get a filter parse tree is if they called parseFilter directly (I do it sometimes in Uglify). But if you do that, that filter will never actually become a filter. Because the compileFilter and filterTiddlers methods call parseTiddler never take a parseTree. They call parseFilter directly.

Thus the __only way__ parse tree mutation could ever occur is if someone used monkey-patching to inject alternative methods into this module, and anyone who does monkey-patching simply runs the risk of having their code require updating between releases.

Therefore, this PR is good to go. Or be rejected. One or the other.

I'll reiterate that I keep pushing for this because it's first of many low hanging fruit to speed up TiddlyWiki, because TiddlyWiki becomes __slow as heck__ once you have thousands of tiddlers. I'm grinding my teeth trying to use it on my own projects, and I have to add even more content. I _really_ don't want to have to migrate to another platform.

Jermolene:
Thanks @flibbles your patience and commitment is much appreciated, and I'll look forwards to what you've got up your sleeve.

Issue #6165, opened 2021-11-03
[BUG] EditTextWidget size=1 is not size=1


twMat:
EditTextWidget doc reads:

|size |The size of the input field (in characters)|

...but in practice the width of another 4-5 characters is added. So any width less than that doesn't work, for example:

`<$edit-text tag=input size=1/>`

shows an input field that is at least 5 characters wide.

Any applied `class` also seems to be overrun by this.

The incorrect minimal width setting is a problem because it takes up too much real estate. I sometimes make "editable tables" where each cell is intended to only hold a single or very few characters but because of this issue the table becomes too wide.

saqimtiaz:
The size attribute of $edittext sets the size attribute of the generated input element, how that is rendered is dependent on the browser, see https://www.w3schools.com/tags/att_input_size.asp

A documentation update to clarify this further would be welcome if you feel it is needed. Controlling the width more precisely would need to be done via applying a custom class and your own CSS that sets the width, perhaps in em.

twMat:
>dependent on the browser

Oh, that was surprising to hear. Thanks. And, to my (second) surprise it does work with CSS... I did test it before I posted but must have done... whatever, thank you :-)

I'll edit the docs.

[closing]

PR #6403, opened 2022-01-16
Update size attribute in EditTextWidget.tid

twMat:
ref: https://github.com/Jermolene/TiddlyWiki5/issues/6165


Jermolene:
Thanks @twMat

PR #6405, opened 2022-01-19
Update doc list-links-draggabe Macro.tid

twMat:
Unless I misunderstand something:
A quick test indicates that the referred to problem is resolved... I assume it is so because we can now directly edit fields in the current tiddler. This is a pretty significant improvement to list-links-draggable.

Maybe there are other macros similarly affected and where the docs should be updated?


Jermolene:
Good thinking, thanks @twMat

PR #6409, opened 2022-01-19
Add support for renderEnd method for storyviews

### LINKED ISSUE #6408: [IDEA] post-render hook for storyviews
Storyviews are a versatile way to add animation and behaviour to a `$list` widget. However currently there is no post-render hook for a storyview, which I am finding that I need in my use case to control the initial display state.

Could we extend the $list widget to call a `storyview.postRender()` method if it exists for the given storyview, at the end of `ListWidget.prototype.render()`?

saqimtiaz:
This PR adds support for an optional post render method for storyviews called `renderEnd()` that allows storyviews to control the initial display state of the list.

The name of the method is consistent with `refreshStart()` and `refreshEnd()` storyview methods.

Fixes #6408 

I note that we currently do not have developer docs for storyviews, and should add some when the opportunity presents itself.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -51,6 +51,9 @@ ListWidget.prototype.render = function(parent,nextSibling) {
 	} else {
 		this.storyview = null;
 	}
+	if(this.storyview && this.storyview.renderEnd) {
+		this.storyview.renderEnd();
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz

PR #6410, opened 2022-01-19
Add support to tm-scroll message for scrolling without animating

### LINKED ISSUE #6406: [IDEA] extend tm-scroll message, $scrollable and pagescroller to optionally scroll without animation
The widget message `tm-scroll` currently triggers a scroll that is animated using the duration as configured in `$:/config/AnimationDuration`. This makes it impossible to trigger a scroll with no animation without completely disabling animations in the wiki.

It is suggested to extend the widget message `tm-scroll` with an optional `animation-duration` parameter, which if present takes precedence over `$:/config/AnimationDuration` in the scroll handling in `$scrollable` and `PageScroller`.

Edit: alternatively we could have an `animate=yes/no` paramter for `tm-scroll`

saqimtiaz:
This PR extends the `tm-scroll` widget message to allow scrolling without animating by passing the optional attribute `animationDuration` as `0` or any other interval.

@Jermolene Note that I have left the dud `callback` argument in place in both `PageScroller` and `Scrollable`. My personal preference would be to clean it up and remove it since it is never used, can never be used as the code stands, and only serves to confuse the function signatures. I would appreciate feedback on your preference in this matter. Should I go ahead and clean that up?

Fixes #6406

### DIFF
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -49,10 +49,14 @@ Handle an event
 */
 PageScroller.prototype.handleEvent = function(event) {
 	if(event.type === "tm-scroll") {
+		var options = {};
+		if(event.paramObject && event.paramObject.hasOwnProperty("animationDuration")) {
+			options.animationDuration = event.paramObject.animationDuration;
+		}
 		if(event.paramObject && event.paramObject.selector) {
-			this.scrollSelectorIntoView(null,event.paramObject.selector);
+			this.scrollSelectorIntoView(null,event.paramObject.selector,null,options);
 		} else {
-			this.scrollIntoView(event.target);
+			this.scrollIntoView(event.target,null,options);
 		}
 		return false; // Event was handled
 	}
@@ -62,9 +66,9 @@ PageScroller.prototype.handleEvent = function(event) {
 /*
 Handle a scroll event hitting the page document
 */
-PageScroller.prototype.scrollIntoView = function(element,callback) {
+PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	var self = this,
-		duration = $tw.utils.getAnimationDuration(),
+		duration = (options && options.hasOwnProperty("animationDuration")) ? parseInt(options.animationDuration) : $tw.utils.getAnimationDuration(),
 	    srcWindow = element ? element.ownerDocument.defaultView : window;
 	// Now get ready to scroll the body
 	this.cancelScroll(srcWindow);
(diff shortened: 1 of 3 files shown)


saqimtiaz:
@Jermolene please review when you get the chance and note that I have updated the original PR description to make it clearer.

Note that I have:
- introduced an `options` argument instead of one specifically for `animationDuration` in case we need to  introduce any other options later on
- the code uses `hasOwnProperty` to check if the appropriate property is present to guard against the edge use case where a JavaScript developer may pass the int `0` as the value of `animationDuration`, which would resolve to `false` if checking with `if(options.animationDuration)`

core/modules/utils/dom/scroller.js
```diff
@@ -49,10 +49,14 @@ Handle an event
 */
 PageScroller.prototype.handleEvent = function(event) {
 	if(event.type === "tm-scroll") {
+		var options = {};
+		if(event.paramObject && event.paramObject.hasOwnProperty("animationDuration")) {
```
Jermolene, line 53: We've got a wrapper for hasOwnProperty `$tw.utils.hop()`

Jermolene (commented):
Thanks @saqimtiaz, just a minor comment

core/modules/utils/dom/scroller.js
saqimtiaz: Good catch, thank you. Now resolved.

saqimtiaz:
Thank you @Jermolene; I have resolved the issue and this should be good to go.

core/modules/utils/dom/scroller.js
```diff
@@ -50,7 +50,7 @@ Handle an event
 PageScroller.prototype.handleEvent = function(event) {
 	if(event.type === "tm-scroll") {
 		var options = {};
-		if(event.paramObject && event.paramObject.hasOwnProperty("animationDuration")) {
+		if(event.paramObject && $tw.utils.hop(event.paramObject,"animationDuration")) {
```
Jermolene, line 53: I should have pointed out that `$tw.utils.hop()` always returns false if the object is falsy, so the first condition can be removed here.

saqimtiaz: apologies for not catching this myself, it has been a while since I looked at that utility method.

saqimtiaz:
OK, I think we should be good now. Apologies for the oversights in this PR, time away from core work means that the utils methods aren't as fresh in my mind.

Jermolene:
Thanks @saqimtiaz

PR #6415, opened 2022-01-22
Add doc related to WikiText parser modes

btheado:
Added documentation and examples about WikiText parser modes and how the parser transitions between them.

Rendered version with all the new tiddlers opened by default can be seen at http://btheado.github.io/TiddlyWiki5/wikitext-parser-mode-docs.html


Jermolene:
Thank you @btheado, magnificent stuff. I really appreciate the detailed work that goes into docs updates like this. It might be worth posting to the group about these updates to get some wider feedback. I suspect that this is an area where many experienced TW users still lack confidence.

PR #6426, opened 2022-01-27
fix: bug with importvariables widget when importing set widgets

### LINKED ISSUE #6424: [BUG] ImportVariables does not reliably import $set widgets.
`$importvariables` does not import variables defined by `$set` widgets in the following three examples of tiddlers to import from, only the macro in Example 2 and 3 are successfully imported: (Windows 10, Chrome 97, TW 5.2.1)

Example 1:
```
<$set name="a" value="alpha"></$set>
```

Example 2:
```
\define b() beta

<$set name="a" value="alpha"></$set>
```

Example 3:
```
\define b() beta

<$set name="a" value="alpha"><$set name="c" value="charlie"></$set></$set>
```

A quick peek at the code suggests the problem is that we parse the tiddler in block mode and not inline, the loop finds a `p` element and aborts variable collection:
https://github.com/Jermolene/TiddlyWiki5/blob/6452eb56a9c498cae4ebdb14278fae011a2d8485/core/modules/widgets/importvariables.js#L47-L50

After changing line 49 to the following, all three test cases from above are successful:

`var parser = widgetPointer.wiki.parseTiddler(title,{parseAsInline:true});`

saqimtiaz:
Fixes bug with `$importvariables` when importing `$set` widgets by parsing in inline mode.

fixes #6424

### DIFF
--- a/core/modules/widgets/importvariables.js
+++ b/core/modules/widgets/importvariables.js
@@ -46,7 +46,7 @@ ImportVariablesWidget.prototype.execute = function(tiddlerList) {
 	this.tiddlerList = tiddlerList || this.wiki.filterTiddlers(this.filter,this);
 	// Accumulate the <$set> widgets from each tiddler
 	$tw.utils.each(this.tiddlerList,function(title) {
-		var parser = widgetPointer.wiki.parseTiddler(title);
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz

Issue #5961, opened 2021-08-17
Updating highlight.js to 11.2.0 - will add breaking changes. 


pmario:
I wanted to create a fast "build and copy / paste" update, but ... 

We use v9.18.1 at the moment, which is from March 2020 .. so about 18 months old. 

The actual version is: 11.2.0 ... 
With 10.x  they did introduce some breaking changes. especially: (no more IE11, etc.). (We're using ES2015 code now.)

Version 10 and 11 will also include some API changes. .. So it will need more than just a "fast build and copy some files" update for us.

pmario:
see: https://github.com/highlightjs/highlight.js/blob/main/VERSION_10_BREAKING_CHANGES.md

and: https://github.com/highlightjs/highlight.js/blob/main/VERSION_10_UPGRADE.md
and https://github.com/highlightjs/highlight.js/blob/main/VERSION_11_UPGRADE.md

kookma:
Hi @pmario - there is a critical fix for Fortran language which I and all my student use and this needs 11 above!
So I hope @Jermolene update the highlight.js!

nilslindemann:
[See here](https://github.com/Jermolene/TiddlyWiki5/discussions/6121#discussioncomment-1673861) how to adapt to the breaking changes in the highlight.js project.

Gk0Wk:
@pmario I upgraded the highlight.js version to 10+ at the beginning of the year and slightly modified `$:/plugins/tiddlywiki/highlight/highlightblock.js` to fit the new API. so far the new version of highlight.js is working well without problems. I recently upgraded it again to the latest 11.3.1 and there are no problems either.

My modified `$:/plugins/tiddlywiki/highlight/highlightblock.js` is as follows for reference.

```js
(function () {

    /*jslint node: true, browser: true */
    /*global $tw: false */
    "use strict";

    var TYPE_MAPPINGS_BASE = "$:/config/HighlightPlugin/TypeMappings/";
    var CodeBlockWidget    = require("$:/core/modules/widgets/codeblock.js").codeblock;
    var hljs               = require("$:/plugins/tiddlywiki/highlight/highlight.js");

    hljs.configure({
        tabReplace: "    "
    });

    CodeBlockWidget.prototype.postRender = function () {
        var domNode  = this.domNodes[0],
            language = this.language,
            tiddler  = this.wiki.getTiddler(TYPE_MAPPINGS_BASE + language);
        if (tiddler) {
[...]

PR #6427, opened 2022-01-28
Update highlight.js to v11.4.0

cdruan:
Since highlight.js is now using ES2015, the existing highlight (v9) plugin is moved to highlight-legacy. The new highlight plugin contains v11.4.0.

* ensure loading this ES6 plugin will not cause error on legacy browsers (tested on ipad3 and IE11)
* update the code to use new highlight.js APIs
* change class tagging to match more closely with highlight.js
* allow users to add language definitions as JS "highlight" modules

Ref: #5961


kookma:
Many thanks @cdruan!
v11 also address several issues, and one is the better support for Fortran, I use in my classes!

It seems the plugin size is double now! One question

* is it possible to trim languages to some base sets and then user can choose extra language when installed the plugin? May be  you could have some sub plugin like the ones in CodeMirror?
* The above suggestion is valid, id stripping down the list of brushes really make plugin smaller

By the way I really appreciate your efforts!

cdruan:
Hi @kookma 

Thanks for your CodeMirror idea! I added an option to allow users to add additional languages as JS tiddlers, similar to how CodeMirror sub-plugins are done. See howto for details. Would you be able to try it out?

Thanks!

plugins/tiddlywiki/highlight/files/tiddlywiki.files
```diff
@@ -2,6 +2,16 @@
 	"tiddlers": [
 		{
 			"file": "highlight.pack.js",
+			"fields": {
+				"type": "application/javascript",
+				"title": "$:/plugins/tiddlywiki/highlight/highlight-9.js",
+				"module-type": "library"
+			},
+			"prefix": "var hljs = require(\"$:/plugins/tiddlywiki/highlight/highlight-9.js\");\n",
+			"suffix": "\nexports.hljs = hljs;\n"
```
pmario, line 11: Why do you include both versions? ... 

IMO only 1 version should be part of the plugin!

plugins/tiddlywiki/highlight/highlightblock.js
```diff
@@ -16,9 +16,23 @@ var TYPE_MAPPINGS_BASE = "$:/config/HighlightPlugin/TypeMappings/";
...
+var useNew = true;
 
-hljs.configure({tabReplace: "    "});	
+if (!$tw.hljs) {
+	$tw.hljs = require("$:/plugins/tiddlywiki/highlight/highlight.js");
+	
+	if ($tw.hljs.getLanguage === undefined) {
+		$tw.hljs = require("$:/plugins/tiddlywiki/highlight/highlight-9.js");
+		$tw.hljs.configure({tabReplace: "    "});
+		useNew = false;
```
pmario, line 27: Same as above. I don't think this should be there anymore.

plugins/tiddlywiki/highlight/highlightblock.js
```diff
@@ -16,9 +16,23 @@ var TYPE_MAPPINGS_BASE = "$:/config/HighlightPlugin/TypeMappings/";
 
 var CodeBlockWidget = require("$:/core/modules/widgets/codeblock.js").codeblock;
 
-var hljs = require("$:/plugins/tiddlywiki/highlight/highlight.js");
+var useNew = true;
```
pmario, line 19: What's the reason to have this new global variable?

pmario:
I personally think, the initial languages should be the same (except coffeescript) as the old version, BUT the old v9.x code should be removed. 

If the package is about 350kByte imo it would be fine.

plugins/tiddlywiki/highlight/files/tiddlywiki.files
cdruan: > Since highlight.js is now using ES2015, I added conditional require to support both modern and legacy browsers. If this is not necessary, let me know and I can remove old dependencies. Legacy browsers tested: iOS9 on ipad3 and IE11.

I wasn't quite sure if TW is staying with ES5 as long as it can--as 5.2.1 still loads nicely on my ipad3--and I think highlight.js is quite useful, so I included both versions for backward compatibility. Obviously this increases the size of the plugin...

I can remove v9 easily.

plugins/tiddlywiki/highlight/highlightblock.js
cdruan: To determine which version of highlight.js is loaded so CodeBlockWidget can use the correct API calls.

pmario:
hmm. I personally don't use highlight.js ... So I'm probably not the right person to suggest how it should be implemented.

But I do care about the size of plugins, so I wouldn't want to install a plugin which uses 600kByte, if I knew it could be smaller.

@Jermolene Should we keep the "old" version highlight-legacy **and** add a newer version as eg: "highlight-latest" which we keep up to date. Plus add some info about browser compatibility to the plugin info text?

plugins/tiddlywiki/highlight/files/tiddlywiki.files
Jermolene: We are indeed trying to stick with ~ES5.x as long as possible, so it is certainly useful to have the older version available. I wonder if it might be easier to make them be entirely different plugins, offered alongside one another in the plugin library.

Jermolene:
Thanks @cdruan for your hard work. I do indeed favour splitting off a separate legacy version of the plugin, and I like the idea of splitting off the languages into sub-plugins. Making each language into a separate plugin seems clumsy, though, do you have any thoughts for how they might be bundled?

cdruan:
@Jermolene 

> Making each language into a separate plugin seems clumsy, though, do you have any thoughts for how they might be bundled?

You don't need to bundle them into sub-plugins. This PR teaches users to add additional languages by creating javascript tiddlers in their wiki. The JS tiddler template has been prepped for them; all they have to do is copy-and-paste the language file content into a text area. Users may choose to add one JS tiddler per language or lump all languages into one JS tiddler.

The way highlight.js prepares their language files makes it easy to register new languages, with one caveat. If you look at the end of a typical highlight.js language file, you will find this:

```js
hljs.registerLanguage("basic",E)})();
```

To load the file via `require()`--without making substantial changes to the language file--you'd need to somehow pass an `hljs` named instance into the language module scope. The easiest way I can think of is via global variable. So a "highlight" JS module will look like this:

```js
var hljs = $tw.hljs; // global variable

// one or more highlight.js language file content pasted below this line

(()=>{.....;hljs.registerLanguage("basic",E)})();
(()=>{.....;hljs.registerLanguage("lisp",E)})();
```

In summary, this PR:

* adds highlight engine as part of `$tw` object--`$tw.hljs`.
* looks for any javascript tiddler with module-type "highlight" and loads it as an additional language defintion
* prepends `var hljs = $tw.hljs;` to the language file content.

cdruan:
This is what pops up when a user clicks "Add New Language(s)" button, located in the "howto" tab of the plugin.

highlight-js

And links to highlight.js language files are also provided in the "howto" section.

kookma:
@cdruan - nice solution! It think it is simple to be used by regular users. You may like to have a look also at [https://kookma.github.io/TW-Scripts/](https://kookma.github.io/TW-Scripts/#Add%20Language%20Brushes%20to%20Highlight.js%20Using%20Startup%20Modules:%5B%5BAdd%20Language%20Brushes%20to%20Highlight.js%20Using%20Startup%20Modules%5D%5D%20%5B%5BAdd%20More%20Language%20Brushes%20to%20Highlight.js%20Plugin%5D%5D%20%5B%5BManually%20Add%20More%20Language%20Brushes%20to%20Highlight.js%5D%5D)

cdruan:
Hi @kookma

Thanks for the tip! By using`$tw.utils.evalSandboxed()` instead of `require()`, there is no longer a need to tweak the language files. It is now just direct copy-and-paste from a language file into a JS tiddler, i.e. no more:
```js
var hljs = $tw.hljs;
```
in front of every langauge-definition content.

cdruan:
Thanks to everyone's feedback. 

I've now moved v9 to highlight-legacy. Both highlight-legacy and highlight plugins will occupy the same `$:/plugins/tiddlywiki/highlight/` namespace to ensure only one module is loaded at any time.

Notable changes:

* Ensure loading this ES6 plugin will not cause error on legacy browsers (tested on ipad3 and IE11)
* Allow users to add language definitions by directly copy-and-paste the language file content into a new JS tiddler and set the tiddler's module-type to "highlight". A button has been created to facilitate the whole process.

I still need a final say on what languages shall be bundled in the plugin. For now, I bundled only the common languages as defined by highlight.js.

pmario:
>... bundled only the common languages as defined by highlight.js.

Do you have a link?

cdruan:
@pmario  

If you go to this [page](https://highlightjs.org/download/), it's all of the ones that are checked by default.

pmario:
I found it. Since it's now easier to add new settings we should have a small initial setup. 

I would like to have `Apache Config, Nginx config, Dockerfile, LaTex`  and I'm sure @kookma mentioned `Fortran`

Jermolene:
Thank you @cdruan @pmario.

> I still need a final say on what languages shall be bundled in the plugin. For now, I bundled only the common languages as defined by highlight.js.

That seems a reasonable place to start. Perhaps in the future it would be interesting to explore an automated build step to create plugins for each language.

cdruan:
Okay, bundled languages are: common set + apache config + nginx config + latex + dockerfile + fortran.

kookma:
@cdruan - many thanks for all your efforts! it looks nice and have all common language!
Thank for adding Fortran! It would be great to have also Python! but now user can add their language brush easiliy!

plugins/tiddlywiki/highlight-legacy/howto.tid
```diff
@@ -0,0 +1,10 @@
+title: $:/plugins/tiddlywiki/highlight/howto
```
Jermolene, line 1: I think either we should include this information in the existing "readme" or "usage" tiddlers. If it's better in it's own tab, perhaps it might be better called "languages"?

plugins/tiddlywiki/highlight-legacy/howto.tid
```diff
@@ -0,0 +1,10 @@
...
+
+! Supporting Additional Languages
+ 
+The [[highlight.js|https://github.com/highlightjs/highlight.js]] project supports many languages. Only a subset of these languages are supported by the plugin. It is possible for users to change the set of languages supported by the plugin by following these steps:
+ 
+# Go to the highlight.js project [[download page|https://highlightjs.org/download/]], select the language definitions to include, and press the Download button to download a zip archive containing customised support files for a highlight.js syntax highlighting server.
+# Locate the `highlight.pack.js` file in the highlight plugin -- on a stock Debian 8 system running Tiddlywiki5 under node-js it is located at `/usr/local/lib/node_modules/tiddlywiki/plugins/tiddlywiki/highlight/files/highlight.pack.js`.
+# Replace the plugin `highlight.pack.js` file located in step 2 with the one from the downloaded archive obtained in step 1.
+# Restart the Tiddlywiki server.
```
Jermolene, line 10: It would be great if we could write instructions for users who only use TiddlyWiki in the browser.

plugins/tiddlywiki/highlight-legacy/readme.tid
```diff
@@ -0,0 +1,81 @@
+title: $:/plugins/tiddlywiki/highlight/readme
+
+This plugin provides syntax highlighting of code blocks using v9.18.1 of [[highlight.js|https://github.com/isagalaev/highlight.js]] from Ivan Sagalaev.
```
Jermolene, line 3: The readme is what users will see in the plugin library listing, so it's important that the very first paragraph explains that this is the legacy version, and clearly explains the circumstances under which it should be used.

plugins/tiddlywiki/highlight-legacy/usage.tid
```diff
@@ -0,0 +1,16 @@
...
+! Usage
+
+Fenced code blocks can have a language specifier added to trigger highlighting in a specific language. Otherwise heuristics are used to detect the language.
+
+```
+ ```js
+ var a = b + c; // Highlighted as JavaScript
+ ```
+```
+! Adding Themes
```
Jermolene, line 12: Just noticed that "Adding Themes" is here in the "usage" tiddler, which does indeed suggest that "Adding Languages" should be here too.

plugins/tiddlywiki/highlight/howto.tid
```diff
@@ -2,9 +2,33 @@ title: $:/plugins/tiddlywiki/highlight/howto
...
+!! Installing Customized highlight.js library
+
+This is the original method. You replace the supplied highlight.js library with a custom version:
 
 # Go to the highlight.js project [[download page|https://highlightjs.org/download/]], select the language definitions to include, and press the Download button to download a zip archive containing customised support files for a highlight.js syntax highlighting server.
-# Locate the `highlight.pack.js` file in the highlight plugin -- on a stock Debian 8 system running Tiddlywiki5 under node-js it is located at `/usr/local/lib/node_modules/tiddlywiki/plugins/tiddlywiki/highlight/files/highlight.pack.js`.
-# Replace the plugin `highlight.pack.js` file located in step 2 with the one from the downloaded archive obtained in step 1.
+# Locate the `highlight.min.js` file in the highlight plugin -- on a stock Debian 8 system running Tiddlywiki5 under node-js it is located at `/usr/local/lib/node_modules/tiddlywiki/plugins/tiddlywiki/highlight/files/highlight.min.js`.
+# Replace the plugin `highlight.min.js` file located in step 2 with the one from the downloaded archive obtained in step 1.
 # Restart the Tiddlywiki server.
```
Jermolene, line 34: Again, it would be helpful to present the instructions in a form that works for browser-based users too.

plugins/tiddlywiki/highlight/readme.tid
```diff
@@ -1,6 +1,6 @@
 title: $:/plugins/tiddlywiki/highlight/readme
 
-This plugin provides syntax highlighting of code blocks using v9.18.1 of [[highlight.js|https://github.com/isagalaev/highlight.js]] from Ivan Sagalaev.
+This plugin provides syntax highlighting of code blocks using v11.4.0 of [[highlight.js|https://github.com/isagalaev/highlight.js]] from Ivan Sagalaev. If you need this plugin to work on legacy browsers, you would need to install the "Highlight Legacy" version instead.
```
Jermolene, line 3: Good stuff – maybe just spell out what is meant by a legacy browser.

Jermolene (commented):
Thanks for your patience @cdruan, this is intricate work, but will be appreciated by many in the community. I've left a number of minor comments.

plugins/tiddlywiki/highlight-legacy/howto.tid
cdruan: @Jermolene 

Just want to make sure we are talking about the legacy version. I didn't plan to touch its code base so the instructions stay unchanged. Is it worth the effort to backport browser-based method of adding languages? Users would also need to hunt down v9 language files that will be hard to locate as time goes by.

cdruan:
@kookma - python is already included; it's one of the common languages :)

kookma:
Great job! Thank you @cdruan!

plugins/tiddlywiki/highlight-legacy/howto.tid
Jermolene: Of course, I'd missed that, thank you @cdruan

Jermolene:
Thanks for your patience @cdruan. Would you be able to address the other feedback I gave above?

cdruan:
@Jermolene 

>... Would you be able to address the other feedback I gave above?

Sorry that it's not clear. I did address them. Specifically:
* The content of "howto" tab has been moved into the "usage" tab. (both editions)
* For the "modern" version of highlight plugin, the usage info of the browser-based language addition was already there.
* Explained what a legacy browser is in both editions and explained it is legacy-only in the first paragraph of the legacy version.
* Moved legacy plugin into its own `highlight-legacy' namespace instead of sharing the same `highlight` space, so  plugin-info would show the contents of both properly.

plugins/tiddlywiki/highlight-legacy/howto.tid
cdruan: moved to "usage"

plugins/tiddlywiki/highlight-legacy/readme.tid
cdruan: done

plugins/tiddlywiki/highlight-legacy/usage.tid
cdruan: done

plugins/tiddlywiki/highlight/howto.tid
cdruan: It was already there. Renamed the heading to "Browser-based method" to be more clear...

plugins/tiddlywiki/highlight/readme.tid
cdruan: done

Jermolene:
Thanks for your patience @cdruan

kookma:
@cdruan 
This is not an issue, but I realized in this update, `xml` has been dropped. While this plugin allows users to add language definitions as JS "highlight" modules, but if this does take little time, I would kindly ask to add `xml` as default to supported languages. May be you like to update to v11.6.0 also.

cdruan:
@kookma 

xml is already supported by the plugin.

kookma:
> xml is already supported by the plugin.

Ah my bad! 

Thank you @cdruan

kookma:
@cdruan 

If you planned to update the Highlight.JS, please also include the Matlab language as additional language. I think it has been dropped in the latest update of plugin.

PR #6428, opened 2022-01-29
Markdown: Let WikiText parsing handle the creation of LaTeX widgets.

rryan:
When embedding LaTeX snippets in inline HTML nodes, such as TiddlyRemember
macros or HTML tables, the parsing of latex nodes breaks the WikiText by
splitting it into pieces around the latex node.

This commit fixes the issue by converting the Remarkable katex nodes back to
text, using a newline to indicate a block katex snippet. This is then re-parsed
by the WikiText KaTeX plugin.

TESTED:

Created a test wiki with:
```
$ node tiddlywiki.js test --init markdowndemo
$ node tiddlywiki.js test --listen
```

* Verified markdown + KaTeX support still works as expected.
* Verified that embedding LaTeX snippets in inline HTML works (e.g. `<a
href="https://example.com/">$x^2$</a>`).
* Verified the markdown + KaTeX support works as expected with renderWikiText
set to `false`.

Without fix:
image

With fix:
Screen Shot 2022-01-28 at 3 46 25 PM

### DIFF
--- a/plugins/tiddlywiki/markdown/wrapper.js
+++ b/plugins/tiddlywiki/markdown/wrapper.js
@@ -37,25 +37,25 @@ var accumulatingTypes = {
 	"softbreak": true
 };
 // If rendering WikiText, we treat katex nodes as text.
-if (pluginOpts.renderWikiText) {
+if(pluginOpts.renderWikiText) {
 	accumulatingTypes["katex"] = true;
 }
 
 var md = new Remarkable(remarkableOpts);
 
 // If tiddlywiki/katex plugin is present, use remarkable-katex to enable katex support.
-if ($tw.modules.titles["$:/plugins/tiddlywiki/katex/katex.min.js"]) {
+if($tw.modules.titles["$:/plugins/tiddlywiki/katex/katex.min.js"]) {
 	var rk = require("$:/plugins/tiddlywiki/markdown/remarkable-katex.js");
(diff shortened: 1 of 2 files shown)


plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -36,6 +36,10 @@ var accumulatingTypes = {
 	"text": true,
 	"softbreak": true
 };
+// If rendering WikiText, we treat katex nodes as text.
+if (pluginOpts.renderWikiText) {
```
Jermolene, line 40: It's a really tiny thing, but our coding style does not put whitespace between `if` and the following parenthesis.

Jermolene (commented):
Thanks @rryan looks good, just a very minor point.

plugins/tiddlywiki/markdown/wrapper.js
rryan: Oops! It looks like this file uses the wrong style. Should I just change touched lines or the whole file?

Jermolene: Oops indeed! I think it would be best to make it consistent. Thank you!

rryan: No prob, done.

Jermolene:
Thank you @rryan

PR #6431, opened 2022-01-29
Add keyword-spacing rule to .eslintrc.yml to reflect the project code style.

rryan:
- Update .eslintignore to ignore "third party" code.
- Add a "lint" script to package.json so `npm run lint` runs eslint on the repo.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,15 +1,8 @@
-# Known minified files
+# Ignore "third party" code whose style we will not change.
 /boot/sjcl.js
+/core/modules/utils/base64-utf8/base64-utf8.module.js
 /core/modules/utils/base64-utf8/base64-utf8.module.min.js
 /core/modules/utils/diff-match-patch/diff_match_patch.js
-/plugins/tiddlywiki/async/files/async.min.v1.5.0.js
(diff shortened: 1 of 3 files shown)


rryan:
I wasn't sure exactly what the repo codestyle was so I just went with polling files in core/modules to get a sense of the implicit style.

There are 113 active errors with this config: https://paste.debian.net/1228822/

Jermolene:
Terrific, thank you @rryan. I think @pmario originally contributed the eslint stuff and may be able to comment. I'd certainly like to get to the point where our coding style was fully automated.

pmario:
It was @hoelzro who created https://github.com/rryan/TiddlyWiki5/commit/aa7a00d080da0179c7c6dbf6a1ab1f7de8efa55c with this comment.

>This adds eslint as a developer dependency, plus a generated eslint
config that doesn't take a very strong stance on much of anything.  The
goal here to get started using automated style checking, add eslint
checking to the testing flow, and gradually introduce stricter checks
over time.

I think most rules are switched of at the moment. 

I think TWs coding style is a bit special, since we don't run a minification step when building the wiki. So the code itself tries to be as compact as uncompressed JS can be. Saving some white-spaces here and there but try to be still readable.

PR #6433, opened 2022-01-30
Add clipboard button to code viewtemplate

ibnishak:


Jermolene:
Thanks @ibnishak. Interesting idea. The only downside would appear to be that it adds visual clutter to every code view tiddler, but I wonder if it is the right approach.

One alternative would be to add two new tiddler toolbar controls to copy the tiddler source and to copy the rendered output. I'd find that useful sometimes, particularly being able to grab the content of a tiddler without flicking to edit mode. It would perhaps be more consistent because the controls would apply to all tiddlers.

Another alternative might be to incorporate copy to clipboard in the generic `<$codeblock>` widget, so that it is always consistently available whenever a user sees a code block.

pmario:
>Another alternative might be to incorporate copy to clipboard in the generic <$codeblock> widget, so that it is always consistently available whenever a user sees a code block.

I would want that to be an option. It can be "on" by default, but it should be controllable.

pmario:
>One alternative would be to add two new tiddler toolbar controls to copy the tiddler source and to copy the rendered output. I'd find that useful sometimes, particularly being able to grab the content of a tiddler without flicking to edit mode. It would perhaps be more consistent because the controls would apply to all tiddlers.

I think, that's an interesting idea. .. But I also think this is already very close to the "Clone" button, which I use often if I need to play with some core code.

ibnishak:
> Another alternative might be to incorporate copy to clipboard in the generic <$codeblock> widget, so that it is always consistently available whenever a user sees a code block.

This would also help update all the document example macros with copy to clipboard buttons. The cost of an if-then branch in codeblock widget is fine I guess.

Issue #762, opened 2014-08-28
Drop support for regexp filter operands


Jermolene:
Back in https://github.com/Jermolene/TiddlyWiki5/issues/228 we added support for any filter operand to be specified as a regular expression by replacing the inner square brackets with slashes. For example:

```
[field:title/ello/] --> on tiddlywiki.com, returns "HelloThere"
```

The syntax implies that a regular expression operand can be used with any filter operator. In practice, we've only implemented support for the `field` operator:

https://github.com/Jermolene/TiddlyWiki5/blob/master/core%2Fmodules%2Ffilters%2Ffield.js

Trying to use the regular expression operand syntax with other filter operators has unexpected results. For example `[/ello/]` (which is an abbreviation of `[title/ello/]`) in the filter tab of advanced search shows a result count of 1 but doesn't display any results.

Rather than updating all the other filters to explicitly support regexp operands, I propose to remove support for regexp operands and instead introduce a new regexp operator:

```
[regexp:fieldname[/myregexp/gi]]
```

The new operator would apply the specified regular expression to the specified field of each tiddler in the current subresult.

PR #6438, opened 2022-02-01
Fixed issue: deprecated regexp could cause crash

flibbles:
Found this issue while writing unit tests for TW5-Uglify.

If you use the deprecated [regexp/operators/] in many contexts, it causes the RSoD. Even "[//]" crashes Tiddlywiki.

The problem occurs because so many filter operators assume there will always be at least one operand. Regexp operators were the exception. My change makes it so the regexp operand is included. Most operators will not test for "regexp" and will treat it like a string operand, but whatever. Better than crashing.

I've never understood how these operators were supposed to work, so let me know if this fix is wrong.

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -95,7 +95,11 @@ function parseFilterOperation(operators,filterString,p) {
 			if(nextBracketPos === -1) {
 				throw "Missing closing bracket in filter expression";
 			}
-			operand.text = filterString.substring(p,nextBracketPos);
+			if(operator.regexp) {
+				operand.text = "";
+			} else {
+				operand.text = filterString.substring(p,nextBracketPos);
+			}
 			operator.operands.push(operand);
 			p = nextBracketPos + 1;
 		}
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -95,10 +95,8 @@ function parseFilterOperation(operators,filterString,p) {
 			if(nextBracketPos === -1) {
 				throw "Missing closing bracket in filter expression";
 			}
-			if(!operator.regexp) {
-				operand.text = filterString.substring(p,nextBracketPos);
-				operator.operands.push(operand);
-			}
+			operand.text = filterString.substring(p,nextBracketPos);
+			operator.operands.push(operand);
 			p = nextBracketPos + 1;
 		}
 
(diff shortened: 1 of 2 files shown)


flibbles:
I actually put in a slightly different fix than what I had before. Now regexp operands are empty-strings instead of whatever was inside the regexp operand. What I had before worked fine for TiddlyWiki, but it would have broken Relink. This will hopefully bother plugins less.

pmario:
Hi. Do you have an example, to reproduce the problem?

I also couldn't find the "deprecated" keyword in any tiddler docs. .. We may also need to improve the docs.

flibbles:
Also, all of the tests I added to the framework are all examples of filter operators which cannot tolerate having _zero_ operands. ("all", "title", "prefix", ...). There are probably plenty more, including many made by plugins. These are just some of the ones I found.

pmario:
I think there shouldn't be any tests. .. @Jermolene That stuff should be removed. 

See: https://tiddlywiki.com/#Release%205.0.17-beta  It has been deprecated in beta-17! ... The warning is written to the dev-console, where absolutely nobody sees it: https://github.com/Jermolene/TiddlyWiki5/commit/d45c417c187fe3dcc35ee7a308e64feebc7b185b

The associated ticket is from 2014: https://github.com/Jermolene/TiddlyWiki5/issues/762 ... I think 8 years+ is enough time to keep deprecated code around.

flibbles:
Actually, I might agree with @pmario on this. I'm having a lot of trouble finding regexp operand use cases for my uglify tests that DON'T cause exceptions. I can't imagine anyone in the world is using these.

pmario:
> Currently the "field" operator is the only one that makes any use of regexp operands. 

Does this mean, the core is still using it? Can you point a link to the code?

pmario:
>...The "field" operator is just the only operator which has code to handle regexp operands in case some end user is using them.

@Jermolene 
OK, so we need a decision if the deprecated code should be removed _and_ if the field code should be adjusted too.

 ... The main problem for me is, that I'm completely unable to create a filter, that uses the function, so that it actually works. .. I only can produce error messages or it doesn't work at all.

Jermolene:
> @Jermolene OK, so we need a decision if the deprecated code should be removed _and_ if the field code should be adjusted too.

I don't think we can remove the regexp operand handling at this point, having kept it past the 5.0.x beta phase.

> ... The main problem for me is, that I'm completely unable to create a filter, that uses the function, so that it actually works. .. I only can produce error messages or it doesn't work at all.

Try this in the filter tab of the advanced search box of tiddlywiki.com:

```
[field:text/b[ea]d/]
```

Jermolene:
Thanks @flibbles

PR #6442, opened 2022-02-02
Is draft

flibbles:
Found another problem while writing Uglifier tests.

[is[draft]] is analogous to [has[draft.of]], but [!is[draft]] is _not_ analogous to [!has[draft.of]], nor is it the true complement of [is[draft]], nor was it ever tested.

I've added an adjustment to make [is[draft]] and [!is[draft]] proper complements of each other. This should not cause any trouble with end-users and backward compatibility, because it's actually really tough to make a tiddler with a blank "draft.of" field, which is the one place where they were incorrect.

### DIFF
--- a/core/modules/filters/is/draft.js
+++ b/core/modules/filters/is/draft.js
@@ -19,7 +19,7 @@ exports.draft = function(source,prefix,options) {
 	var results = [];
 	if(prefix === "!") {
 		source(function(tiddler,title) {
-			if(!tiddler || !$tw.utils.hop(tiddler.fields,"draft.of")) {
+			if(!tiddler || !$tw.utils.hop(tiddler.fields,"draft.of") || (tiddler.fields["draft.of"].length === 0)) {
 				results.push(title);
 			}
 		});
--- a/core/modules/filters/is/draft.js
+++ b/core/modules/filters/is/draft.js
@@ -19,13 +19,13 @@ exports.draft = function(source,prefix,options) {
 	var results = [];
 	if(prefix === "!") {
 		source(function(tiddler,title) {
-			if(!tiddler || !$tw.utils.hop(tiddler.fields,"draft.of") || (tiddler.fields["draft.of"].length === 0)) {
+			if(!tiddler || !tiddler.isDraft()) {
 				results.push(title);
 			}
 		});
 	} else {
 		source(function(tiddler,title) {
-			if(tiddler && $tw.utils.hop(tiddler.fields,"draft.of") && (tiddler.fields["draft.of"].length !== 0)) {
+			if(tiddler && tiddler.isDraft()) {
 				results.push(title);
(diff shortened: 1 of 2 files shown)


core/modules/filters/is/draft.js
```diff
@@ -19,7 +19,7 @@ exports.draft = function(source,prefix,options) {
 	var results = [];
 	if(prefix === "!") {
 		source(function(tiddler,title) {
-			if(!tiddler || !$tw.utils.hop(tiddler.fields,"draft.of")) {
+			if(!tiddler || !$tw.utils.hop(tiddler.fields,"draft.of") || (tiddler.fields["draft.of"].length === 0)) {
```
Jermolene, line 22: There is a `tiddler.isDraft()` method in the core that perhaps would be worth using here.

What is the rationale for the new test for the length of the draft not being zero? Is there a situation where a zero length `draft.of` field is likely to occur?

Jermolene (commented):

flibbles:
So this is getting into how my PRs concern me and not really anyone else.

zero-length `draft.of` is very difficult to create unless you use setField actions or something, which is why this probably won't affect anyone one way or another. I was just writing unit tests and I discovered that is[draft] and !is[draft] are not complements, because if someone _does_ have a zero-length `draft.of` field, neither of those two operators will pass it along.

tiddler.isDraft is absolutely viable. I just tried rewriting it and it's fine. However it isn't _exactly_ analogous to [has[draft.of]] like the documentation says. {"draft.of": ""}.isDraft() is true, but {"draft.of": ""}=>[has[draft.of}] is false. But again, this affects nobody but me. I just think [is[draft]] should either be analogous to isDraft() or to [has[draft.of]], but right now it's neither.

When I talked about throttling myself, I was suggesting that maybe I don't fill up the PR queue with super pedantic nonsense like this. Because really, I only found this because of my edge-case unit tests. I don't think this "bug" is really bothering anyone. Except for you. Because of me.

flibbles:
But I do agree with you. isDraft() and [is[draft]] should be synonymous. People would expect that. So I'd be happy to revise this PR so that [is[draft] uses isDraft(), and accept all the subtle implications that means (which won't affect anyone except for unit-testing neurotics).

Jermolene:
> But I do agree with you. isDraft() and [is[draft]] should be synonymous. People would expect that. So I'd be happy to revise this PR so that [is[draft] uses isDraft(), and accept all the subtle implications that means (which won't affect anyone except for unit-testing neurotics).

Thanks @flibbles, that would be great.

flibbles:
Done. Should be ready to merge.

saqimtiaz:
bump. This should be good to go @Jermolene

Jermolene:
Thanks @flibbles @saqimtiaz

PR #6443, opened 2022-02-02
Add mention of HTML inline vs. block to wikitext parser mode

btheado:
Add mention of HTML inline vs. block to wikitext parser mode based on the discussion with @AnthonyMuscio at https://talk.tiddlywiki.org/t/parsing-modes-inline-vs-block/1994/11

### DIFF
--- a/editions/tw5.com/tiddlers/wikitext/parser/WikiText Parser Modes.tid
+++ b/editions/tw5.com/tiddlers/wikitext/parser/WikiText Parser Modes.tid
@@ -1,13 +1,16 @@
 created: 20220110233944530
-modified: 20220122182842030
+modified: 20220202022415581
 tags: WikiText
 title: WikiText Parser Modes
 type: text/vnd.tiddlywiki
 
-The WikiText parser has three modes:
+In order to display Tiddlers (usually the text field), the WikiText parser reads and interprets the content and applies WikiText rules. The parser has three modes:
 
 * ''pragma mode'' - the parser will recognise only [[pragma mode WikiText|Pragma]] punctuation 
 * ''block mode'' - the parser will recognise only [[block mode WikiText|Block Mode WikiText]] punctuation
 * ''inline mode'' - the parser will recognise only [[inline mode WikiText|Inline Mode WikiText]]
 
 The parser [[transitions between these modes|WikiText parser mode transitions]] based on the text it encounters. In addition, there are  [[places where the parser ignores WikiText|Places where the parser ignores WikiText]] punctuation.
+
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/wikitext/parser/WikiText Parser Modes.tid
```diff
@@ -1,13 +1,16 @@
...
\ No newline at end of file
+The parser [[transitions between these modes|WikiText parser mode transitions]] based on the text it encounters. In addition, there are  [[places where the parser ignores WikiText|Places where the parser ignores WikiText]] punctuation.
+
+<<.tip "The concept of inline vs block also exists for standard HTML elements. For HTML, these [[two layout modes|https://www.w3schools.com/html/html_blocks.asp]] determine if the output flows together on the same line or not.
+<p>Most [[block mode WikiText|Block Mode WikiText]] corresponds to [[block level HTML elements|https://developer.mozilla.org/en-US/docs/Web/HTML/Block-level_elements]] and most [[inline mode WikiText|Inline Mode WikiText]] corresponds to [[inline level HTML elements|https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements]]. However, for Wikitext the two modes are just as important for determining which syntax will be recognised by the parser as they are for determining how the output will flow.</p>">>
```
pmario, line 16: There is no need to cover the text in a P element. Just add a new-line between the 2 paragraphs.

btheado: That's the first thing I tried and it didn't work. I didn't dive into the `.tip` macro to figure out why and I just added the P element instead.

btheado: Looks like the tip macro is written like this:

```
\define .tip(_) <div class="doc-icon-block"><div class="doc-block-icon">{{$:/core/images/tip}}</div>$_$</div>
```

which results in it being parsed inline. It would have to change to something like this to get wikitext paragraphs recognized:

```
\define .tip(_)
<div class="doc-icon-block">

<div class="doc-block-icon">{{$:/core/images/tip}}</div>$_$</div>
\end
```

pmario: Sorry. You are right. We do need it atm. ... I didn't see that the tip macro covered _all_ the text. I thought it ends with the first line. ...

Jermolene:
Thanks @btheado @pmario

PR #6447, opened 2022-02-06
Add TW5-browser-nativesaver to tw5.com docs

slaymaker1907:
Added [TW5-browser-nativesaver](https://github.com/slaymaker1907/TW5-browser-nativesaver) to the documentation as a saver option.

See below for what the new tiddler looks like when rendered.

image


Jermolene:
Thanks @slaymaker1907

Issue #6407, opened 2022-01-19
[IDEA] caseinsensitive behavior for prefix/suffix operators


saqimtiaz:
It would be helpful to have optional case-insensitive support in the `prefix[]`, `suffix[]` operators (and perhaps `removeprefix[]` and `removesuffix[]`).

One option would be to introduce a suffix for these operators, for example:

`prefix:ci[]`

Jermolene:
Good idea. I think that the existing operators that have case sensitivity/insensitivity flags are "match" and "search", and that they spell things out in full to `casesensitive` and `caseinsensitive`.

kookma:
@saqimtiaz having casein sensitive/insensitive support for prefix/suffix is absolutely useful!

PR #6468, opened 2022-02-18
Support case insensitive matching in prefix/suffix operators

benwebber:
Support `caseinsensitive`/`caseinsensitive` suffixes in the following filter operators:

  * `prefix`
  * `suffix`
  * `removeprefix`
  * `removesuffix`

The suffixes have the same behaviour as the `match` operator.

Closes: #6407

---

The existing case sensitivity blurb for the `match` operator works with for these filters as well. Maybe we can extract it as its own tiddler?


saqimtiaz:
Thank you @benwebber 
Could you please also add some tests for the operators that now have new suffixes? The existing tests for the prefix operator can be found here:
https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-filters.js#L270

The tests can be run via `tiddlywiki editions/test`

benwebber:
Ran into something adding the test cases. What is the expected behaviour of `removeprefix` and `removesuffix` if the parameter is the empty string? Currently they act differently.

`removeprefix` does not filter the input:

```
ABCDE abcde +[removeprefix[]]
```
```
ABCDE abcde
```

Whereas `removeprefix` discards all titles:

```
ABCDE abcde +[removesuffix[]]
```
```
```

Intuitively, I would expect `removesuffix` to behave like `removeprefix`.

Jermolene:
Thanks @benwebber @saqimtiaz 

>  I am personally tempted to correct the behaviour of removesuffix[] so that it does not filter the list and remove all items when called with an empty string as operand. Thoughts @Jermolene ?

I think that makes sense.

benwebber:
Great. `suffix` has the same behaviour, should we change it too?

saqimtiaz:
> Great. `suffix` has the same behaviour, should we change it too?

Yes please.

core/modules/filters/insertbefore.js
```diff
@@ -20,7 +20,7 @@ exports.insertbefore = function(source,operator,options) {
 	source(function(tiddler,title) {
 		results.push(title);
 	});
-	var target = options.widget && options.widget.getVariable(operator.suffix || "currentTiddler");
+	var target = operator.operands[1] || (options.widget && options.widget.getVariable(operator.suffix || "currentTiddler"));
```
saqimtiaz, line 23: I assume this was not meant to be included in this PR?

benwebber: Correct, one moment.

saqimtiaz:
bump

Jermolene:
Great stuff, thanks @benwebber

Issue #6133, opened 2021-10-22
Extend insertbefore[] to accept a second parameter in place of the suffix.


saqimtiaz:
The `insertbefore[]` operator currently has a weird pattern where you supply a variable name as the suffix, to specify which tiddler to use as the marker for insertion. It would be more intuitive in current day TW to use a second parameter for this. This would also allow text references to be used instead of a variable.

 Suggest extending this and any other operators with the same pattern with an optional second parameter, that takes precedence over the suffix if both are present.

PR #6477, opened 2022-02-23
Support specifying insertbefore position title as parameter

benwebber:
`insertbefore` now accepts the position title as a parameter, instead of as a variable name in the suffix. If both are specified, the parameter takes precedence.

I could not find any other filters which used a similar technique. I searched for similar `op-suffix` documentation strings as well as `getVariable()` calls that could reference the filter suffix.

I updated calls that used the legacy syntax in the core.

Closes: #6133

### DIFF
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -752,16 +752,27 @@ Tests the filtering mechanism.
 			rootWidget.setVariable("tidTitle","e");
 			rootWidget.setVariable("tidList","one tid with spaces");
 			
+			// Position title specified as suffix.
 			expect(wiki.filterTiddlers("a b c d e f +[insertbefore:myVar[f]]",anchorWidget).join(",")).toBe("a,b,f,c,d,e");
 			expect(wiki.filterTiddlers("a b c d e f +[insertbefore:myVar<tidTitle>]",anchorWidget).join(",")).toBe("a,b,e,c,d,f");
 			expect(wiki.filterTiddlers("a b c d e f +[insertbefore:myVar[gg gg]]",anchorWidget).join(",")).toBe("a,b,gg gg,c,d,e,f");
-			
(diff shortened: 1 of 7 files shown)


saqimtiaz:
@benwebber this looks good, thank you.

The change to also modify the core to use the newer syntax was unexpected, however I do think it makes the code more intuitive and readable.

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -752,16 +752,27 @@ Tests the filtering mechanism.
...
+			expect(wiki.filterTiddlers("a b c d e +[insertbefore[f],[a]]",anchorWidget).join(",")).toBe("f,a,b,c,d,e");
+			expect(wiki.filterTiddlers("a b c d e +[insertbefore[f],<myVar>]",anchorWidget).join(",")).toBe("a,b,f,c,d,e");
+
+			// Parameter takes precedence over suffix.
+			expect(wiki.filterTiddlers("a b c d e +[insertbefore:myVar[f],[a]]",anchorWidget).join(",")).toBe("f,a,b,c,d,e");
 	
-			// Next 2 tests do weired things, but will pass - there for compatibility reasons
+			// No position title.
 			expect(wiki.filterTiddlers("a b c [[with space]] +[insertbefore[b]]").join(",")).toBe("a,c,with space,b");
-			expect(wiki.filterTiddlers("a b c d e +[insertbefore:2[b]]").join(",")).toBe("a,c,d,e,b");
```
pmario, line 764: Why was this test deleted?

saqimtiaz: My understanding is that it has been replaced by the test on line 773 which tests the same thing, as part of better organizing the tests according to the preceding comments.

benwebber: That's correct.

saqimtiaz:
@Jermolene this is ready for review and looks good from my perspective.

Jermolene:
Thank you @benwebber

Issue #6119, opened 2021-10-14
[BUG] The untagged operator filters out nonexistent tiddlers


IBUzPE9:
**Describe the bug**
The `untagged` operator filters out nonexistent tiddlers. 

**To Reproduce**
Create new tiddler with `ListWidget`:
```
<$list filter="[enlist[one two three]untagged[]]"/>
```
TiddlyWiki 5.2.0 displays empty tiddler.

**Expected behavior**
TiddlyWiki 5.1.x displays `onetwothree`.
The description of the `untagged` operator says: "A tiddler is deemed to have no tags if it doesn't exist".

**TiddlyWiki Configuration:**
 - Version 5.2.0
 - Saving mechanism - browser
 - No plugins installed
 
**Desktop:**
 - Manjaro Linux
 - Firefox 93.0

PR #6478, opened 2022-02-23
Consider non-existent tiddlers untagged

benwebber:
Fixes regression introduced in 575c23359.

Fixes: #6119
See-also: 575c23359 (Update untagged filter to avoid $tw.utils.pushTop (#6034), 2021-09-18)


saqimtiaz:
Good catch.

PR #6480, opened 2022-02-23
Extend $draggable to support an optional drag handle

saqimtiaz:
This PR extends the $draggable widget to optionally use a CSS selector identifying the child DOM node to be used as the drag handle. 

This allows for use cases like dragging a "card" where only the header of the card is draggable but the drag image shows the entire card and you can still select text within the card and interact with it as normal.

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -16,21 +16,25 @@ Browser data transfer utilities, used with the clipboard and drag and drop
 Options:
 
 domNode: dom node to make draggable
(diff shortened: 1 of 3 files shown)


Jermolene:
Looks good, thanks @saqimtiaz

saqimtiaz:
@Jermolene I am leaving this as a draft until I can review with a fresher head. Will mark it ready for review and ping you once that is done.

saqimtiaz:
@Jermolene this is now ready, I have done some cross browser testing as well (Chrome, FF and IE11).

Jermolene:
Thanks @saqimtiaz

PR #6494, opened 2022-02-27
Improve readability for documentation macros

TiddlyVee:
While I was recording a video, I found out, that those macros are not readable by users in view mode.


pmario:
@Jermolene .. At the moment the `code-body` field doesn't exist. So for the meantime it would be nice to have the `<pre><$view field="text"/></pre> ` element at the end of some macros. 

It doesn't change the functionality, since all those tiddlers are tagged `$:/tags/Macro` so the body text content doesn't matter. It only improves readability. 

Once `code-body` is active we can change them back to use this field

Jermolene:
Thanks @pmario @TiddlyVee

PR #6495, opened 2022-02-27
Improve the docs for the sortan operator

TiddlyVee:
The "Apple" tiddler didn't have any body text. The example has some more info for a new user, what to test.


TiddlyVee:
The Apple tiddler looks like this now

image

The example looks like this

image

pmario:
bump

Jermolene:
Thanks @pmario

PR #6497, opened 2022-02-28
Fix issues in WidgetMessage: tm-open-window

CodaCodr:
Fixed error in example width and height settings (open window does not support CSS units).
Made param table text self-consistent.
Minor text changes.

### DIFF
--- a/editions/tw5.com/tiddlers/messages/SampleWindowTemplate.tid
+++ b/editions/tw5.com/tiddlers/messages/SampleWindowTemplate.tid
@@ -1,5 +1,5 @@
 created: 20211109165213041
-modified: 20211117224321325
+modified: 20220227211343988
 tags: [[Message Examples]]
 title: SampleWindowTemplate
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @CodaCodr

CodaCodr:
@Jermolene I notice that 5.2.2 prerelease has been updated with "New in 5.2.2" pieces but the changes were applied to the _older_ version before this PR. The erroneous width/height units (this time "em") are back and the weird "something" attr in the example, too.

saqimtiaz:
What has happened here is that the master branch on which the pre-release is based had updates to the same tiddler that received a docs update in the tw.com branch. I suspect trying to merge tw.com branch into master will not cause a conflict that will need to be manually resolved.

Jermolene:
Hi  @CodaCodr I've just merged tiddlywiki.com into master, so things should be up to date now.

PR #6498, opened 2022-02-28
Improve affordances for new browser windows

saqimtiaz:
This PR implements two changes:

1. Extends the support for `tm-open-window` to allow opening a tiddler in more than one browser window by specifying a unique `windowID` parameter.
2. Introduces `tm-close-window` and `tm-close-all-windows` to allow closing browser windows opened with `tm-open-window`.

### DIFF
--- a/core/modules/startup/windows.js
+++ b/core/modules/startup/windows.js
@@ -103,7 +103,7 @@ exports.startup = function() {
 	var closeAllWindows = function() {
 		$tw.utils.each($tw.windows,function(win) {
 			win.close();
-		});		
+		});
 	}
 	$tw.rootWidget.addEventListener("tm-close-all-windows",closeAllWindows);
 	// Close open windows when unloading main window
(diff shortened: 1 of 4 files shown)


saqimtiaz:
@Jermolene I have updated this PR and it should be good to go. 

New changes:
- a tiddler can be opened in as many new windows as desired by providing a unique identifier.
- this identifier is also used to specify a window to close with `tm-close-window`
- also introduced `tm-close-all-windows` message for closing all windows (removed this facility from `tm-close-window`).

kookma:
Hi @saqimtiaz
This is wonderful. One question: Is it possible to have side by side windows for editing and displaying the results?
That means have two windows for the same tiddler one in edit mode and other in the view mode?

One suggestion: In Mentat plugin one can align window to left or right and also keep it inside browser window, what do you think if TW core support this feature?

saqimtiaz:
> This is wonderful. One question: Is it possible to have side by side windows for editing and displaying the results?
> That means have two windows for the same tiddler one in edit mode and other in the view mode?

Yes.

Jermolene:
Thanks @saqimtiaz

PR #6501, opened 2022-03-01
Fix sorting of sidebar language, plugin and theme list

### LINKED ISSUE #6452: [IDEA] Sorting of plugins in sidebar list
The changes to the sidebar plugin listing in 5.2.0 are a definite usability improvement. However, the list of plugins is still sorted by their descriptions, while the first thing a reader now sees is the plugin name. The non alphabetical sorting of the names makes it harder to scan through the list and locate the correct plugin.

I suggest we sort the list by the plugin name.

image

saqimtiaz:
This PR changes the sidebar lists for language, plugin and theme to sort alphabetically by name rather than by description.

Fixes #6452

Before:
image

After:
image


Jermolene:
Thanks @saqimtiaz

PR #6508, opened 2022-03-04
[fr] New batch of documentation translations

tw-FRed:
In this batch:
* Update of documentation macros
* Update of EN -> FR glossary
* More documentation translation updates


Jermolene:
Thanks @tw-FRed

PR #6512, opened 2022-03-06
`parseDate` operator

FlashSystems:
This PR adds a filter operator for parsing ISO dates. This is useful for parsing the output of HTML5 input elements when the date-types (`date`, `datetime-local`, etc.) are used. It's also useful when dealing with JSON formatted data.

This PR adds the `parseDate` filter operator, tests and the necessary documentation.

This PR is the result of the discussion in PR https://github.com/Jermolene/TiddlyWiki5/pull/6377 and the suggestion to split this PR into multiple PRs for easier review.


core/modules/filters/parsedate.js
```diff
@@ -0,0 +1,38 @@
...
+exports.parsedate = function(source,operator,options) {
+    var results = [];
+    source(function(tiddler,title) {
+        var date = null;
+        switch (operator.operand) {
+            case "ISO":
+                date = new Date(title);
+                break;
+            default:
+                return ["parseDate: Unkown operator " + operator.operand];
```
pmario, line 27: I think tw system messages need to be translatable. see an example: https://github.com/Jermolene/TiddlyWiki5/blob/aaf1aa3821e470735ac24b2f21af58835bc5d2ba/core/modules/filters/format.js#L34

They are part of [misc.multids](https://github.com/Jermolene/TiddlyWiki5/blob/aaf1aa3821e470735ac24b2f21af58835bc5d2ba/core/language/en-GB/Misc.multids#L31) text file

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -1009,6 +1009,21 @@ Tests the filtering mechanism.
...
+			expect(wiki.filterTiddlers("[[2015-03-25T15:40Z]parsedate[ISO]]").join(" ")).toBe("20150325154000000");
+			expect(wiki.filterTiddlers("[[2015-03-25T15:40:32Z]parsedate[ISO]]").join(" ")).toBe("20150325154032000");
+			expect(wiki.filterTiddlers("[[2015-03-25T15:40:32Z]parsedate[ISO]]").join(" ")).toBe("20150325154032000");
+			expect(wiki.filterTiddlers("[[2015-03-25T15:40:32-01:00]parsedate[ISO]]").join(" ")).toBe("20150325164032000");
+			// Validate that a date in the local timezone is correctly parsed and represented as UTC
+			// This is tricky because we can not set the used timezone.
+			expect(wiki.filterTiddlers("[[2015-03-25T15:40]parsedate[ISO]]").join(" ").substr(8, 2)).toBe((new Date("2015-03-15T15:40:32")).getUTCHours().toString());
+			expect(wiki.filterTiddlers("[[nodate]parsedate[ISO]]").join(" ")).toBe("");
+			expect(wiki.filterTiddlers("[[2015]] [[2020]] +[parsedate[ISO]]").join(" ")).toBe("20150101000000000 20200101000000000");
+		});
```
pmario, line 1025: I think the `default` code path has no tests

core/modules/filters/parsedate.js
Jermolene: We need to consider carefully how to handle errors here.

We don't currently handle or report errors within filter operators; generally a filter operator encountering an error would just not return any results at all.

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -1009,6 +1009,21 @@ Tests the filtering mechanism.
 			expect(wiki.filterTiddlers("[title[HelloThere]] +[format:titlelist[]]").join(" ")).toBe("HelloThere");		
 		});
 	
+		it("should handle the parsedate operator", function() {
```
Jermolene, line 1012: We should also have tests for trying to convert invalid dates.

Jermolene (commented):
Thanks @FlashSystems some comments below

saqimtiaz:
I am a bit concerned that users could find the name of the operator misleading and expect it to parse the created and modified fields of tiddlers. Perhaps the name of the operator can indicate that it parses ISO format dates? `parseisodate[]` for example?

core/modules/filters/parsedate.js
FlashSystems: The `filter` operator returns an error message if the suffix is wrong. In case of this operator, I'd like to give an error message, because the list of possible values is quite short (actually 1 currently). A simple typo would be a hard to debug error, because you would not know if your input is just not parsable or if the operand is wrong.

Actually, I'd prefer an out-of-band method to give the user the error message. Like logging the error and returning an empty value. But this would be a new endeavour altogether.

FlashSystems:
> I am a bit concerned that users could find the name of the operator misleading and expect it to parse the created and modified fields of tiddlers. Perhaps the name of the operator can indicate that it parses ISO format dates? `parseisodate[]` for example?

There was a previous [discussion](https://github.com/Jermolene/TiddlyWiki5/discussions/6301#discussioncomment-1780909) where @Jermolene suggested a universal method for parsing foreign date formats. I liked this idea and therefore, this one has a parameter for specifying the format of the date. But I understand your concerns. What do you think of `parseForeignDate`, just `foreignDate`, `date2tw` or `dateToTW` to make it more clear what the purpose of this operator is.

editions/test/tiddlers/tests/test-filters.js
FlashSystems: There is already the test case `expect(wiki.filterTiddlers("[[nodate]parsedate[ISO]]").join(" ")).toBe("");` for an invalid date. I'm using the JavaScript engines `Date` constructor to parse the date. Testing the JavaScript engine by providing different invalid dates was out of scope in my opinion. I just want to test the error handling of TW's implementation of the operator and thought that one test with an (obviously invalid) date would be sufficient.

core/modules/filters/parsedate.js
Jermolene: > The `filter` operator returns an error message if the suffix is wrong.

Ouch! I hadn't noticed that. On that basis your current implementation makes sense.

editions/test/tiddlers/tests/test-filters.js
Jermolene: Thanks @FlashSystems

Jermolene:
Thanks @FlashSystems @saqimtiaz . I think I'm OK with the name; it's hard to make operator names completely self-explanatory, but "parse" and "date" seem like the essential properties of the operator.

I do have one last concern: we are directly exposing the functionality of the JS date parser, which is not recommended by MDN:

>  Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#parameters

pmario:
>Parsing of date strings with the Date constructor (and Date.parse(), which works the same way) is strongly discouraged due to browser differences and inconsistencies.

I wouldn't be too worried about that: It also says: 

> - Support for [RFC 2822](https://datatracker.ietf.org/doc/html/rfc2822) format strings is by convention only.
> - Support for ISO 8601 formats differs in that date-only strings (e.g. "1970-01-01") are treated as UTC, not local.

If we define ISO it's the same as TW uses by default for `modified` and `created`. So our users (should) know that `GMT+0100`,  `UTC` and `TZD` have a meaning: https://tiddlywiki.com/#DateFormat .. We only need to add ore examples, so they know, what they are doing. 

RFC2822 is supported by browser vendors by convention ... IMO that's good enough. If we wouldn't be allowed to use those functions, they shouldn't be there at all. ... 

If we invest some time into more docs I think we have a good chance to add great value, by exposing those functions

FlashSystems:
The ECMAScript Standard mandates a time Format for compliant implementations. I think the main source of confusion for the user originates from this part of the specification of [Date.parse](https://tc39.es/ecma262/#sec-date.parse):

> The function first attempts to parse the String according to the format described in Date Time String Format ([21.4.1.15](https://tc39.es/ecma262/#sec-date-time-string-format)), including expanded years. If the String does not conform to that format the function may fall back to any implementation-specific heuristics or implementation-specific date formats.

The format specified in [Chapter 21.4.1.15](https://tc39.es/ecma262/#sec-date-time-string-format) should be consistent between JavaScript implementations as long as they comply with the ECMAScript standard. But the user may use the fallback without noticing. And that may fail on another implementation/browser.
I'd suggest using a regular expression to validate the input to the `parseDate` filter to the subset specified in the standard. That way, we can make sure that it's interpreted the same for all complying implementations.

Maybe the format should be called "ISO-ECMA" or "JSON" and not just "ISO" to prevent confusion. The format specified in the ECMAScript standard is just a subset of ISO 8601.

What do you think?

FlashSystems:
I've gone ahead and reworked this PR. The reworked pull request contains the following changes:

- Fixed a bug that prevented returning the error message for an invalid filter parameter. Thank you, @pmario, for spotting the missing test case that hid this bug.
- Added a testcase for passing an invalid parameter.
- Made the error message for an invalid filter parameter translatable. @Jermolene: I hope I interpreted your reaction correctly and that it's OK to return an error message on an invalid parameter (format specifier).
- Added a regular expression to limit the date format to ECMAScript [Date Time String Format](https://tc39.es/ecma262/#sec-date-time-string-format) (without expanded years), to make sure parsing is consistent for all JavaScript implementations.
- Added testcases for the regular expression based on the [Test262: ECMAScript Test Suite](https://github.com/tc39/test262). See comment in test for more details.
- Added a reference to the ECMAScript specification to the documentation. I think it does a good job explaining the format and has some examples.
- Rebased on current master.

I hope I've addressed all of your concerns. Thank you for the constructive feedback. Please let me know if I missed something.

FlashSystems:
The main question here is, what ECMAScript specification does TiddlyWiki needs to support?
Since ECMAScript 2016 (ES7) the rules are clear:

> When the time zone offset is absent, date-only forms are interpreted as a UTC time and date-time forms are interpreted as a local time.

I will put this into the documentation as well.

ECMAScript 2015 (ES6) specifies the behaviour a little different:

> If the time zone offset is absent, the date-time is interpreted as a local time.

I can make the behaviour match the ES7 spec with a little trick: Just append "T00:00:00Z" to all date-only formats to make it behave as in ES7.

All major browsers (except IE 11) since 2017 support at least ECMAScript 2015 (ES6) as far as I can tell.
If ES5 support is required, all bets are off. Parsing must be done manually. Getting this right considering the local time zone is a nightmare. Not using the local time zone for parsing would be very irritating for the user. Especially if a date/time selector is used.

If you're ok with a minimum requirement of ECMAScript 2015 (ES6) I'll put the workaround (for ES7 compatibility) in place and extend the documentation with some more examples. Then this should be good to go.

@jermolene could you share your thoughts on this, please?

FlashSystems:
In the absence of feedback, I went ahead in finalizing the PR:
* I renamed the date-format from `ISO` to `JS` because only a subset of ISO 8601 is understood. This will prevent any name collision if a full ISO 8601 parse is added at any time later.
* I extended the examples to contain a chapter about time zone parsing and an example for using the HTML5 date-time-input-element.
* The parser should now work consistently on ECMAScript 2015 (ES6) and above. See [here](https://github.com/Jermolene/TiddlyWiki5/pull/6512#issuecomment-1078074194) for a detailed explanation.

Jermolene:
Hi @FlashSystems thanks for your patience, and sorry for the delay.

At the moment the core is intended to be compatible with ES5.1. I'm keen to modernise, but it must be done via a policy decision.

In the meantime, how bad would manual parsing be?

Another quick thing I noticed is that perhaps `parseECMAScriptDate()` should go into `utils.js` for reuse.

FlashSystems:
Hello @Jermolene, I think it might be almost impossible to do the parsing manually in ES5.1. The problem is, that the date and time must be parsed in the local timezone. This feature is intended to work with the input-element types for date/time input[^1]. These generate locate time without a time zone designator. To make matters worse, the time must not be parsed with the current time zone offset but with the offset active at the specified moment. Daylight saving time, changes in time offsets and even changes in international borders have to be accounted for correctly, to translate historical local date and time values to UTC. I don't think this could be done in ES5.1 without re-implementing the whole time zone database in TiddlyWiki.

Regarding moving `parseECMAScriptDate` into `utils.js`. If you think this function offers enough value to be put there, I'll move it there.

[^1]: I know, it's a bit weird that the parsing should be compatible with a feature (HTML5 date/time input) that was not around, when ES5.1 was standardized. But in my opinion, this would be the primary use case for this function at the moment.

linonetwo:
@FlashSystems Yes, I may use that method in my plugin if it is in the utils

pmario:
@FlashSystems ... Are you still with us?

>Regarding moving parseECMAScriptDate into utils.js. If you think this function offers enough value to be put there, I'll move it there.

As Jeremy pointed out at https://github.com/TiddlyWiki/TiddlyWiki5/pull/6512#issuecomment-1086653322 -> It should be moved to utils.js

The docs would need to be adjusted to `<<.from-version "5.4.0">>`

FlashSystems:
Sorry I stopped using TiddlyWiki over two years ago and have not done any development on it since then.

linonetwo:
I can take this, luckily I'm still using tiddlywiki :P

Issue #2807, opened 2017-03-16
Paste from clipboard


twMat:
I would like to be able to paste images from the clipboard/memory into tiddlers. Possibly onto the drawing canvases that you get by setting tiddler *type* to image,  but ideally straight into the default editor textarea. For a digital notebook, this is a reasonable feature IMO.

Maybe these are viable routes:

* https://codepen.io/netsi1964/pen/IoJbg
* https://gist.github.com/Buildstarted/4577472
* https://www.lucidchart.com/techblog/2014/12/02/definitive-guide-copying-pasting-javascript/

Jermolene:
Hi @twMat not sure if you know, but Chrome does already support pasting; click on the page background and then use the menu or keyboard shortcut to trigger paste, and the pasted item will be imported as a new tiddler.

Making things work in other browsers would indeed be useful, but last time I looked the prospects were not good. (Copying to the clipboard has got a lot better supported, though, and we should probably add a message or widget for it).

twMat:
Ah - I did not know this! Great!

And on my Win10 it actually works in both **Chrome, FF and Edge!** Not IE.

Thanks @Jermolene . Closing.

Jermolene:
Pretty neat that it works in Firefox and Edge, that's new. I've updated the docs.

ldorigo:
Hi! 

Sorry to reopen this old issue, but I cannot get this to work, neither on Chrome nor on any browser I tried (Firefox, Safari, Opera). What I'm doing is taking a screenshot (with macOS, that copies it to the clipboard) and trying to paste it into a new tiddler I'm making. Nothing seems to happen.

At the moment, the only way there is for me to add a screenshot is: 

Save screenshot to a file -> find file on my computer -> drag to TW -> confirm import -> open the new image tiddler -> edit -> change title to something more readable than the auto-generated screenshot name -> confirm -> go to original tiddler -> click on the 'insert image'  -> find the image in the list (which is not sorted by date - so it can be quite hard to find) -> finally resume editing. 

That's more than 11 steps for something that should be trivial. When using TiddlyWiki to take notes in class as I do, it's a huge waste of time which means I often fall behind what the teacher says just because I need to annotate an image. 

So either I'm doing it wrong, or something is broken with the paste feature :)

Plloi:
@ldorigo I just tested on windows 7 and it still works,  so I'd say open a new issue as it may be a mac specific problem.

Jermolene:
[...]

PR #6517, opened 2022-03-08
Add th-readFileCallback hook mechanism

FSpark:
This PR is meant to address this issue: [#2807](https://github.com/Jermolene/TiddlyWiki5/issues/2807)

The hook was initially added to enable automatic renaming of pasted files, as pasted files will only have default names, which is difficult to replace by relying on existing hooks, the exact reason for this is [here](https://github.com/Jermolene/TiddlyWiki5/issues/2807#issuecomment-1056945191). 

This PR will allow plugin authors to do the following:

```js
	$tw.hooks.addHook("th-readFileCallback",function(tiddlerFieldsArray, triggeringWidget, event) {

		// Do your thing here.

		return tiddlerFieldsArray;
	});
};
```

Hook function parameters:

* **tiddlerFieldsArray**: A reference to the `tiddlerFieldsArray`, where the tiddler title can be rewritten.

* **triggeringWidget**: To facilitate access to some parameters of the target widget, such as `filterByContentTypes` for dropzone.

* **event**: For easy identification of the source based on event.type, e.g. `drop`, `paste`, and `change` for the browse widget. And `File` is avaliable in `lastModified`, `size` and other attributes.

Return value:

* Modified `tiddlerFieldsArray`.

Or should we adopt the same strategy as `th-importing-file` and decide whether to continue based on the return value? So that a custom `readFileCallback` can be implemented.

```js
	if($tw.hooks.invokeHook("th-importing-file",{
		file: file,
		type: type,
		isBinary: isBinary,
		callback: callback
	}) !== true) {
		this.readFileContent(file,type,isBinary,options.deserializer,callback);
	}
```

Another option would be to add an event source to the `info` of [the `th-importing-file` hook](https://tiddlywiki.com/dev/#Hook%3A%20th-importing-file) and intercept it when it detects a type of `paste` and rewrite the file name. 

First time participant, not sure if I should set more parameters, or if I'm over-designing, or if such a Hook is necessary.

I hope to hear everyone's opinions first, and finally I will complete the documentation and comments.

### DIFF
--- a/core/modules/widgets/browse.js
+++ b/core/modules/widgets/browse.js
@@ -63,6 +63,7 @@ BrowseWidget.prototype.render = function(parent,nextSibling) {
 		} else {
 			self.wiki.readFiles(event.target.files,{
 				callback: function(tiddlerFieldsArray) {
+					tiddlerFieldsArray = $tw.hooks.invokeHook("th-readFileCallback",tiddlerFieldsArray,self,event);
 					self.dispatchEvent({type: "tm-import-tiddlers", param: JSON.stringify(tiddlerFieldsArray)});
 				},
 				deserializer: self.deserializer
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -181,6 +181,7 @@ DropZoneWidget.prototype.readFileCallback = function(tiddlerFieldsArray) {
 DropZoneWidget.prototype.handleDropEvent  = function(event) {
 	var self = this,
 		readFileCallback = function(tiddlerFieldsArray) {
+			tiddlerFieldsArray = $tw.hooks.invokeHook("th-readFileCallback",tiddlerFieldsArray,self,event);
 			self.readFileCallback(tiddlerFieldsArray);
 		};
 	this.leaveDrag(event);
@@ -234,6 +235,7 @@ DropZoneWidget.prototype.handleDropEvent  = function(event) {
 DropZoneWidget.prototype.handlePasteEvent  = function(event) {
 	var self = this,
 		readFileCallback = function(tiddlerFieldsArray) {
+			tiddlerFieldsArray = $tw.hooks.invokeHook("th-readFileCallback",tiddlerFieldsArray,self,event);
 			self.readFileCallback(tiddlerFieldsArray);
 		};
 	// Let the browser handle it if we're in a textarea or input box


saqimtiaz:
> Another option would be to add an event source to the `info` of [the `th-importing-file` hook](https://tiddlywiki.com/dev/#Hook%3A%20th-importing-file) and intercept it when it detects a type of `paste` and rewrite the file name.

I have considered whether whether we might want to detect if there is a file object received in a `paste` event and if so, overwrite the default name with a customized one that increments. Question is: is it possible to receive a file object in a `paste` event that does not have the browser default filename? Some testing is needed as well as reading of the spec.

When I did some limited testing in the past I did not find any scenario in which a file object received in the payload of a paste event actually has an original filename passed along. If we can be confident this is always the case, allowing customization of the default file name could be a nice core improvement.

FSpark:
@saqimtiaz 

Unfortunately, it seems impossible.

I've read the latest Chrome and Firefox source code and determined this to be the case. But I still don't know about other browsers and the previous situation.

If you are interested, you can read:

[data\_object\_item.cc - Chromium Code Search](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/clipboard/data_object_item.cc;drc=459dba1945dd88c8dd98c6fbfd617f84e62eba99;l=186)

[m:d:DataTransferItem::CreateFileFromInputStream (DataTransferItem.cpp - mozsearch)](https://searchfox.org/mozilla-central/source/dom/events/DataTransferItem.cpp#386)

saqimtiaz:
@FSpark if it is impossible to receive a filename other than the browser default filename in a paste event, then from our perspective that is a good thing. It means we can safely override the name.

I will follow up this weekend with an example of what I am thinking.

FSpark:
@saqimtiaz Yes indeed😂, waiting for you

FSpark:
Hi, @saqimtiaz.  👀 Any updates?

saqimtiaz:
Hi @FSpark, apologies for not getting back to you sooner, I have been pressed for time lately.

I did look into this but ran into an issue I had not considered and decided to take a step back and give it some thought.

Here is what I tested:
- modified `dropzone.handlePasteEvent()` to pass along a property of the `options` object to `wiki.readFile()` indicating that we need a unique title
- `wiki.readFile()` passes the options object to `wiki.readFileContent()`
- in `readFileContent()`, use `wiki.generateNewTitle()` to generate a tiddler title unique in the wiki. This method works by adding a suffix, so we get titles like Untitled1.png, Untitled2.png etc

This works and gives you unique tiddler titles when pasting images as long as you paste one image, complete the import and then paste the next image.

However, the problem that this does not solve is that if you paste more than one image without completing the import, the second image will overwrite the first and so on. For example if we pasted Untitled1.png but did not complete the import, the next image will also be Untitled1.png and will overwrite the first image in the $:/Import tiddler.

While we could also check the contents of the $:/Import tiddler to make sure the title is unique there as well, I am unsure of the merits of that and wanted to take a step back and give it some thought. Some of the questions on my mind are: is that a universal enough need to warrant such an approach, and whether there might be a more elegant solution.

I will revisit as soon as I get the opportunity. In the meantime I welcome your thoughts on this.

pmario:
>I will revisit as soon as I get the opportunity. In the meantime I welcome your thoughts on this.

I did have an idea about a name-pool while importing. ... I'll have to run some tests, then I'll post more.

pmario:
It seems the existing TW upgrader mechanism kicks in early enough, before something hits the $:/Import tiddler. At that time it's possible to modify everything. The title, to be created and the payload. ... 

The upgrade mechanism is "stackable", so it will allow plugin authors to easily add new possibilities. All of them can be active at the same time.  

The $:/Import tiddler will probably work fine, since we know that name and we can check it, if there is an image in the queue with the same name. 

The problem will be the `$:/ImportImage-xxxx` tiddlers. Since the upgrader doesn't know that name upfront, since every Draft tiddler will create its own importer. 

So we can detect, if the date from the clipboard is a PNG and we can generate a unique name. ... eg: 
{$:/config/image/importer/prefix} + <<now YYYY-0MM-0DD-0hh:0mm:0ss:0zzz>>.png

That should be relatively easy to implement. What do you think

pmario:
I do have a POC, which is described here: https://talk.tiddlywiki.org/t/changing-image-png-for-importing-images/2698/18?u=pmario

FSpark:
Sorry @saqimtiaz , it turns out this discussion moved to the Talk forum and I hadn't followed it before.

I think this override scenario is needed to avoid the hassle of cancelling and reintroducing it.

`wiki.generateNewTitle()` generates a title that only has a count and cannot use the date template.

My original intention for adding that Hook was actually to rewrite the filename as a unique identifier like a timestamp or UUID, but it seems @pmario has implemented it so far, good job!

pmario:
>My original intention for adding this Hook was actually to rewrite the filename as a unique identifier like a timestamp or UUID, but it seems @pmario has implemented it so far, good job!

Thx. There are 2 things I don't like with my solution. 

 1. I don't know, if an image comes from the clipboard as `image.png` or if someone drags a file named `image.png` into a wiki. Both of them would be automatically renamed. 
 2. I need a config tiddler, to identify the name that should be renamed, since browsers use different default names depending on the language setting. 

So for me it would be nice, if the drag & drop handler would give me some info, where the file comes from. So I know, I can safely rename it.

saqimtiaz:
Note that as of mid-2021, in Chrome you can now [copy a file from the operating system/file manager and it is available to paste](https://groups.google.com/a/chromium.org/g/blink-dev/c/MRDuNaR4Zd0/m/n7J2g6qNAQAJ) in the browser. As such, it cannot safely be assumed that all files being received in paste events have browser generated file names.

FSpark:
Sorry, I overlooked this situation as I mainly use FF.

It looks like things are getting more complicated as there doesn't seem to be a way to distinguish between the two cases in `event.clipboardData`.

Safari and Chrome currently support this feature, while Firefox has been discussed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1087675) for 8 years with no results.

Maybe we can really only improve the user interface and leave it up to the user?

Like a global option to force automatic renaming or a button to automatically rename individual items?

saqimtiaz:
> Maybe we can really only improve the user interface and leave it up to the user?
> 
> Like a global option to force automatic renaming or a button to automatically rename individual items?

This might be the most prudent way to approach this.

saqimtiaz:
@FSpark see #6560

pmario:
@FSpark ... Should this PR still be considered? ... There are some conflicts, which will  need to be fixed.

PR #6523, opened 2022-03-11
Eventcatcher widget: add "matchSelector" attribute

Jermolene:
This PR extends the eventcatcher widget with an optional new attribute to specify a "matchSelector" that must be matched by target DOM nodes in order for the event to be trapped. This makes it possible to selectively ignore events that occur within the descendent nodes of the root selector identified by the "selector" attribute.


saqimtiaz:
Looks good, changes look to be backwards compatible. 

We should perhaps mention in the docs that the `matchSelector` attribute is only respected if used in combination with the `selector` attribute.

Jermolene:
> We should perhaps mention in the docs that the `matchSelector` attribute is only respected if used in combination with the `selector` attribute.

Good thinking, but on reflection I wonder if it might be better to make them independent. One could imagine using `matchSelector` on its own if one didn't need the feature of copying DOM attributes from the root selector.

Issue #4702, opened 2020-06-11
[IDEA] Optional parsing of metadata inside markdown using frontmatter


xeor:
**Is your feature request related to a problem? Please describe.**
I use Bob and an external editor for editing most of my content outside tw5. I'm using some of the same files for a static-blog which can have metadata using frontmatter inside the markdown files. tw5 however, wants to store them outside of the markdown file in .meta files.

**Describe the solution you'd like**
Instead of using .meta files for markdown. There should be a way to use frontmatter for storing the metadata in `.md` files

**Describe alternatives you've considered**
* Using .tid files. But then the editor's markdown-plugins wouln't understand. The static-blog also won't get it's metadata.
* Using .meta files gives better editor support. But there are still 2 files, static blog still need frontmatter on top of files, and the bob-extension isnt good at syncing .meta files.

**Additional context**
Using frontmatter would make tw5 a very nice candidate for all markdown note-taking people out there. I think this addition would be a very welcoming feature.

kravlost:
Bump! This would be a huge improvement and halve the number of files in the tiddlers folder. It's one of the things that stops me from using Markdown format tiddlers, much as I'd prefer them so I don't need to learn two forms of markup.

kravlost:
@pmario That would be great, thanks!

pmario:
>...integrate a YAML parsing library into the core which is a high overhead for a feature likely to be used by a select number of users.

I think that's a no-go. ... If we would use the same mechanism, we use for .tid meta elements, that should be enough for a start.

joshuafontany:
> > ...integrate a YAML parsing library into the core which is a high overhead for a feature likely to be used by a select number of users.
> 
> I think that's a no-go. ... If we would use the same mechanism, we use for .tid meta elements, that should be enough for a start.

Look at how the core loads the *.js files in the core folders. As long as it has commented out front-matter with just name: value (*.tid style) fields listed. No *.meta file needed.

pmario:
>Look at how the core loads the *.js files in the core folders. As long as it has commented out front-matter with just name: value (*.tid style) fields listed it will be read as a tiddler. No *.meta file needed.

I think it will have problems with the start and end "-------" front-matter uses but the rest should be there.

pmario:
Close in favour of: https://github.com/TiddlyWiki/TiddlyWiki5/issues/7534

PR #6528, opened 2022-03-13
Revamp markdown plugin

cdruan:
Instead of merging markdown parse tree into TW tree, this PR feeds markdown-rendered HTML into TiddlyWiki for a final parse tree. This approach allows raw html inside the markdown text to be interpreted properly. To demo this PR or view test samples, visit [new-markdown-plugin demo wiki](https://cdruan.github.io/TiddlyWiki5/); for comparison, here's [the same wiki with the current markdown plugin](https://cdruan.github.io/TiddlyWiki5/old-markdown.html)

### Other details
 
- Replace `remarkable` with `markdown-it` parser for more robust parsing.
    
- Extend syntax: strikethrough, insertion, mark, superscript, subscript, footnotes, definition lists.
    
- Prevent accidental katex in expressions such as `$5 and $6`, `<$list>...</$list>`, `$:/config%/PageControlButtons/Visibility/$:/core/ui/Buttons/home`.
  
- Ensure links to tiddlers work properly under markdown image and link syntax.
    
- Add "syntax" section to user documentation.

- Add support for "text/markdown" MIME type.


pmario (commented):
Some code remarks

pmario:
I think most of the the "test tiddlers" from: https://cdruan.github.io/TiddlyWiki5 should go to the TW markdown edition.

pmario:
>Replace remarkable with markdown-it parser for more robust parsing.

What did make you to switch to markdown-it from remarkable.

plugins/tiddlywiki/markdown/markdown-it-tiddlywiki.js
```diff
@@ -0,0 +1,242 @@
...
+}
+
+/// post processing ///
+
+function wikify(state) {
+	var href, title, src, alt, replaceTag = undefined;
+
+	state.tokens.forEach(function(blockToken) {
+		if(blockToken.type === 'inline' && blockToken.children) {
+			for(let token of blockToken.children) {
```
pmario, line 169: we don't use `let` at the moment

core/modules/parsers/parseutils.js
```diff
@@ -313,6 +317,9 @@ exports.parseAttribute = function(source,pos) {
 		node.type = "string";
 		node.value = "true";
 	}
+	if(node.type === "string") {
```
Jermolene, line 320: This is a non-backwards compatible change that impacts all parsers that use the `parseAttribute` utility.

TW syntax trees require HTML entities to be expressed as a node with type=entity.

plugins/tiddlywiki/markdown/config.tid
```diff
@@ -0,0 +1,56 @@
+title: $:/plugins/tiddlywiki/markdown/config
+
+! Plugin Configuration
+
+|!Config |!Default |!Description |
+|<code>[[breaks|$:/config/markdown/breaks]]</code>|``false``|markdown-it library config: Convert '\n' in paragraphs into ``<br>`` |
```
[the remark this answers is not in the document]
Jermolene: I think it would be more consistent with other docs to drop the `<code>` tags

plugins/tiddlywiki/markdown/markdown-it-katex.js
```diff
@@ -0,0 +1,266 @@
...
+
+/*jslint node: true */
+'use strict';
+
+var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js");
+
+// Test if potential opening or closing delimieter
+// Assumes that there is a "$" at state.src[pos]
+function isValidDelim(state, pos) {
+    var prevChar, nextChar,
```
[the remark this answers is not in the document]
Jermolene: Our usual approach is to leave 3rd party files untouched, using the tiddlywiki.files mechanism to prefix or suffix any necessary glue code. Is that approach practical here?

plugins/tiddlywiki/markdown/markdown-it-tiddlywiki.js
Jermolene: That's correct - we use (roughly) ES5.1 to maintain compatibility with older browsers without using transpilation.

Jermolene (commented):
Hi @cdruan thank you, this is an ambitious bit of work, and it appears to provide a lot of useful new features that will be appreciated by many users.

I've left some minor comments but an overall concern is (as usual) backwards compatibility. Can we be confident that all users will find their existing content renders correctly?

I note the switch to markdown-it. We only switched to remarkable a couple of years ago in #3876. Are we confident that we've made the best choice this time? Are there any features of remarkable that are not supported in markdown-it that users may be depending on?

There's also a lot of new features here, and our usual backwards compatibility guarantees means that we will be obliged to support those features forever, which may constrain us. Perhaps some of the new features might be considered somewhat opinionated. For example, the checkbox plugin appears to render checkboxes through CSS fakery, rather than using a real checkbox element.

So, one question is whether it is best for this plugin to entirely replace the existing one. Perhaps we need to offer both so that we don't make life difficult for existing users. (There's some discussion of the issues in #3876).

I think it would also be worth trying to get some end user feedback. Perhaps a post to talk.tiddlywiki.org that lays out the improvements, and also makes clear any chances that will affect existing users.

pmario:
As Jeremy pointed out the checklist plugin only adds visual "bells" without real functionality and it fails to print. see screenshot

There is a plugin, which I can't find at the moment, which actually allows users to change the state. It modifies the tiddler content.

grafik

cdruan:
Hi @pmario and @Jermolene

Let me start by explaining why I made the switch to markdown-it. The decision was only made (reluctantly) after I found a series of problems with remarkable:

* Remarkable does not handle raw `<pre>` and `<style>` blocks correctly when the blocks contain blank lines. This does not meet [CommonMark's spec](https://spec.commonmark.org/0.30/#html-block). See [example #169](https://spec.commonmark.org/0.30/#example-169):
  ```html
  <pre language="haskell"><code>
  import Text.HTML.TagSoup

  main :: IO ()
  main = print $ parseTags tags
  </code></pre>
  okay
  ```
  The example rendered by remarkable:
  ```html
  <pre language="haskell"><code>
  import Text.HTML.TagSoup
  <p>main :: IO ()
  main = print $ parseTags tags
  </code></pre>
  okay</p>
  ```

  Notice the extra `<p>` inside `<pre>` block and the spacing is no longer preserved.

* In its `rules_block/deflist.js` module, the variable at line 98 should have been placed outside the loop. Checking with markdown-it-deflist confirms my finding. The consequence of that is the entire formatting of `<dl>` list is determined by the last term's spacing (tight or loose).

* In its `rules_block/htmlblock.js` module, its HTML_TAG regular expressions (line 6 & 7) do not take into account of digits. So `<h1>`..`<h6>` are not recognized as block tags. So if you enter this:

  ```html
  <h4>heading</h4><ul><li>item 1
  </li><li>item2</li></ul>
  ```

  Remarkable yields this:
  ```html
  <p><h4>heading</h4><ul><li>item 1</p>
  </li><li>item2</li></ul>
  ```

  Not only is the `<p>` undesirable, its end tag cuts off the the `<li>` expression.

These problems altogether present a worrisome picture of a no longer well maintained project. Then, on markdown-it's github page:

> markdown-it is the result of the decision of the authors who contributed to 99% of the Remarkable code to move to a project with the same authorship but new leadership (Vitaly and Alex). It's not a fork.

Having worked with both libraries, I can report that markdown-it has extra routines (such as Tokens and default token rendering) that make writing extensions less tedious.

---
As for the motivation of this PR, if you haven't done so already, please visit the [demo using CURRENT markdown plugin](https://cdruan.github.io/TiddlyWiki5/old-markdown.html) and examine the tiddlers under **Issues Fixed** section. You'll see the issues I uncovered.

This PR loads successfully on both IE and ipad3&mdash;after I removed the `let`. There are some math expressions that could not be rendered in IE. On the other hand, the current plugin does not load successfully on either platforms. I suspect it may have to do with the `remarkable-katex.js` module not being es5 compliant.

plugins/tiddlywiki/markdown/config.tid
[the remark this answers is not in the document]
cdruan: The body of text was inherited from previous commits. It's cleaned up now.

plugins/tiddlywiki/markdown/markdown-it-katex.js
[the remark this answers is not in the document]
cdruan: Yes, I purposefully left the spacing alone in case the code needs to be diffed. The code contains several modifications from the original, so I placed attributions at the top of the file and above `math_inline_bloc()`. Is there anything else I should do?

plugins/tiddlywiki/markdown/markdown-it-tiddlywiki.js
cdruan: fixed

core/modules/parsers/parseutils.js
cdruan: @Jermolene

There may be some misunderstandings... I thought `parseAttribute()` functions **within** an HTML node. Here's the kind of problem I'm describing:

If you create a regular tiddler with the following text:
```html
<$codeblock code="var s=&quot;It's tricky&quot;;" language="js"/>
```

Without the patch, the rendered output will be this:
```
var s=&quot;It's tricky&quot;;
```

With the patch, you get:
```
var s="It's tricky";
```

Another example:
```html
<span title="this &quot;has&quot; a title">hello world</span>
```

If you hover over the rendered text "hello world", you'll find the title (tooltip) not showing the expected text: `this "has" a title`.

I've traced the problem down to Widget's calling `domNode.setAttributeNS(namespace,name,value)`. If you don't DECODE entities before calling this function, any entity markup with be treated as literal text.

And I believe unescaping entities needs to happen in the parsing stage; otherwise, by the time you get to widget tree stage, it's difficult to discern where a `codeblock` widget's  `code` value comes from. If it comes from `<$codeblock>` text like above, then the value needs decoding; but if it comes from a fenced code block, the content should not be decoded.

plugins/tiddlywiki/markdown/syntax.tid
```diff
@@ -0,0 +1,207 @@
+
+title: $:/plugins/tiddlywiki/markdown/syntax
+
+! Standard Markdown Syntax
+
+To review standard Markdown syntax, see: [ext[CommonMark quick reference|https://commonmark.org/help/]].
+
+! Linking to Tiddlers
+
+Prepend `#` to tiddler titles to form link addresses. If a tiddler title contains spaces or other special characters, you must either URI-encode the link, or surround the #title with `<>`, backslash-escaping any `<` or `>` in the title.
```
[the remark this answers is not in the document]
cdruan: It's actually in the CommonMark's [link destination spec](https://spec.commonmark.org/0.30/#link-destination). Both remarkable and markdown-it understand this syntax.

pmario:
>As for the motivation of this PR, if you haven't done so already, please visit the [demo using CURRENT markdown plugin](https://cdruan.github.io/TiddlyWiki5/old-markdown.html) and examine the tiddlers under Issues Fixed section. You'll see the issues I uncovered.

I did see the fixes and I'm with you. There are definitely some problems with the remarkable library. Since I don't use markdown regularly, I don't have a problem to switch.

cdruan:
@pmario @Jermolene 

> Perhaps some of the new features might be considered somewhat opinionated. For example, the checkbox plugin appears to render checkboxes through CSS fakery, rather than using a real checkbox element.

> There is a tw5-checklist plugin which actually allows users to change the state.

Yes, the checklist is only a visual presentation of the underlying markdown text, but it's a popular request that even github's markdown has it. See:

- [ ] item 1
- [x] item 2

Most people (myself included) would rather see the above output than the following:
<ul><li><code>[ ]</code> item 1</li><li><code>[x]</code> item 2</li></ul>

(I had to escape the `[ ]` because github is clever enough to transform pure html expression into checklist!)

The checklist may have its place: Sometimes you just want to quickly whip up a checklist and/or you don't want accidental "clicks" that change the state of a checkbox.

And thanks, @pmario, for pointing out CSS styled checkbox doesn't get printed. I've re implemented the checkbox to use disabled `<input type="checkbox">`.

cdruan:
One more thing, the immediate backward incompatibility I can see is that I removed the `image` parse rule from the `renderWikiPragma`. There was a debate whether `prettylink` should be included in the pragma or not. Since `prettylink` is still not in there and `image` rule provides the equivalent syntax for images, I took it out of the default setting... Or maybe both should be IN together?

core/modules/parsers/parseutils.js
cdruan: Hi @Jermolene 

Have you had another chance to look into decoding entities in string literals during ATTRIBUTE parsing? Filtered transclusions and macro values are not impacted by the proposed change.

Jermolene: Hi @cdruan this stuff is subtle. It's correct that HTML decodes entities within attribute values, but that only happens within HTML source code. If you assign an attribute in JS using the DOM API then no such decoding takes place. So the following assigns the literal string `one &amp; two` to the attribute:

```
$0.setAttribute("data-my-attrib","one &amp; two")
```

In TiddlyWiki, we often use computed attributes, where we normally wouldn't want the automatic conversion. For example, the wikify widget has an attribute `text` that contains wikitext; it would break things if the HTML entities were decoded before the wikitext were parsed.

In Markdown, of course, perhaps you would want to decode entities in attributes because there is no mechanism like dynamic attributes to complicate things.

pmario:
@cdruan ... Why did you close this one? ... I think it would be worth to be included.

core/modules/parsers/parseutils.js
Jermolene: Thanks @cdruan

Jermolene:
Hi @cdruan 

> @cdruan ... Why did you close this one? ... I think it would be worth to be included.

See https://github.com/Jermolene/TiddlyWiki5/pull/6528#discussion_r837977350

cdruan:
@pmario and @Jermolene

I've rollbacked changes to `parseutils.js`. The entity decoding (of attribute values) is now performed on the TW parse-tree element and widget nodes&mdash;with the exception of `<$wikify>` and `<$entity>`&mdash;during the post-processing stage.

Other notable updates:

* Add tw tag block parsing (similar to html blocks) to properly parse tw tags separated by blank lines.
* Remove katex block parsing&mdash;I think inline display mode is sufficient. Block parsing only adds ability to include blank lines between `$$...$$`, but this feature is actually undesirable.
* Add [Extending Markdown](https://cdruan.github.io/TiddlyWiki5/#Extending%20Markdown) demo tiddler to demonstrate the ability to extend the Markdown plugin.
* Update/add few demo tiddlers.
* Update user doc, mainly to specify inline math parsing rules.

Please let me know if you have any concerns. Thanks.

Jermolene:
Thanks @cdruan much appreciated.

> I've rollbacked changes to `parseutils.js`. The entity decoding (of attribute values) is now performed on the TW parse-tree element and widget nodes—with the exception of `<$wikify>` and `<$entity>`—during the post-processing stage.

I'm sorry I didn't feedback quick enough on your earlier comments. The decoding of attributes must be applied consistently across all widgets. Arbitrary exceptions add significantly to the cognitive load of the system and future updates.

Is there another approach that doesn't affect wikitext parsing?

cdruan:
Hi @Jermolene 

Thanks for your feedback. I was hoping the disruption is deemed minimal because decoding is applied only to the string type of attribute values in **markdown** tiddlers. The likelihood of users issuing widget calls inside the markdown text is probably low.

Besides allowing consistent grammar across html elements and widgets, decoding the attributes allows me to pass LaTex and fence block markdown contents to TW via `<$codeblock>` and `<$latex>` widgets. (I can go into details on why I ultimately chose this strategy.) I also translate markdown links and images to `<$link>` and `<$image>` calls when appropriate and it'd be nice to have tooltips properly rendered when there are entities in the text. So I'm quite dependent on having a mechanism to decode entities in attribute string values.

I believe I found another widget, `<$macrocall>`, that requires delayed entity decoding... There are probably other exceptions.  Would it be acceptable if the behavior is clearly documented? Alternatively, I could apply decoding only to the attributes of the aforementioned widgets (and html elements) that this PR depends on. Would that work? Personally, it would be nice to have a sure way of escaping quotes in the attribute values. Although delimiting with `"""` minimizes conflicts, theoretically, you can still have a string that contains `"""`.

By the way, I just realized that current markdown plugin may be indirectly decoding entities in attribute values anyway. See [example](https://cdruan.github.io/TiddlyWiki5/old-markdown.html#%3C%24entity%3E:%3C%24entity%3E%20%3C%24codeblock%3E%20%24%3A%2Fplugins%2Ftiddlywiki%2Fmarkdown). For your reference, same markdown tiddlers rendered by [this PR](https://cdruan.github.io/TiddlyWiki5/#%3C%24entity%3E:%3C%24entity%3E%20%3C%24codeblock%3E%20%24%3A%2Fplugins%2Ftiddlywiki%2Fmarkdown).

plugins/tiddlywiki/markdown/wrapper.js
```diff
@@ -3,338 +3,144 @@ title: $:/plugins/tiddlywiki/markdown/wrapper.js
...
 
-exports["text/x-markdown"] = MarkdownParser;
+// to extend MarkdownIt outside of this module, do:
+//
+// md = $tw.Wiki.parsers["text/x-markdown"].prototype.md;
+// md.use(plugin[, options]);
+MarkdownParser.prototype.md = md;
 
-})();
+exports["text/x-markdown"] = MarkdownParser;
```
pmario, line 145: `text/markdown` is an official MIME type now. So it needs to export both, the old an the new

cdruan: Thanks @pmario. Done! 

`text/markdown` is now the default markdown type, but `text/x-markdown` remains recognized.

pmario:
@Jermolene ... This should also be considered for v5.2.3

Jermolene:
Thanks for your patience @cdruan

Jermolene:
Hi @cdruan - apologies for not noticing sooner – please could you resolve the merge conflict? Many thanks

saqimtiaz:
> > I've rollbacked changes to `parseutils.js`. The entity decoding (of attribute values) is now performed on the TW parse-tree element and widget nodes—with the exception of `<$wikify>` and `<$entity>`—during the post-processing stage.
> 
> I'm sorry I didn't feedback quick enough on your earlier comments. The decoding of attributes must be applied consistently across all widgets. Arbitrary exceptions add significantly to the cognitive load of the system and future updates.
> 
> Is there another approach that doesn't affect wikitext parsing?

@cdruan @Jermolene has this issue been resolved?

cdruan:
Sorry, I just saw the recent discussions.

I think we have yet to resolve the issue of decoding HTML entities inside widget and html tag's **attribute** string values. My current implementation is to decode entities inside attribute values of all html tags and widget tags, except `<$wikify>` and `<$entity>` and `<$macrocall>`. This only applies to markdown tiddlers.

However, I realized there's another approach if we could allow another string literal syntax, say: `e"....string content..."`, similar to a raw string syntax in python, etc. Any strings of this form will have their entities decoded during parsing. This way, we could maintain backward compatibility while allowing content like this:

```
<$codeblock code=e"var s = &quot;foobar&quot;;"/>
```

And there won't be any seemingly arbitrary "exceptions" as before.

Would this new approach be acceptable?

cdruan:
@pmario

> Why is <$codeblock code="""var s = "foobar";;"""/> not good enough? It works out of the box.

This PR needs to be able to programmatically pass any content of a fenced code block to `<$codeblock>` widget. So it needs to assume that `"""` could be in the content.

cdruan:
I've updated this PR with the approach described in this [comment](https://github.com/Jermolene/TiddlyWiki5/pull/6528#issuecomment-1189021856). I also made few additional updates to support the new "text/markdown" MIME type ([commit 8e687eb](https://github.com/Jermolene/TiddlyWiki5/pull/6528/commits/8e687eb62aa3c4d074a4b187d445f0367e5fe536)).

The demo pages [new](https://cdruan.github.io/TiddlyWiki5/) vs. [old](https://cdruan.github.io/TiddlyWiki5/old-markdown.html) have been updated as well, powered by 5.2.3-prerelease version.

nicolnt:
> * Extend syntax: strikethrough, insertion, mark, superscript, subscript, footnotes, definition lists.

Very appreciated update, thanks!

Is there any plan to support markdown checklists (TODOs), like here on GitHub?

```
- [ ] to be done
- [x] task is done
```

kravlost:
Will the revamp also allow the plug-in to store meta file data in markdown front matter instead of needing two files per markdown tiddler?

saqimtiaz:
> Will the revamp also allow the plug-in to store meta file data in markdown front matter instead of needing two files per markdown tiddler?

The markdown plugin handles how markdown tiddlers are rendered and displayed in TiddlyWiki, and is not related to loading/save files in the node.js configuration which is a core behaviour.

kravlost:
> > Will the revamp also allow the plug-in to store meta file data in markdown front matter instead of needing two files per markdown tiddler?
> 
> The markdown plugin handles how markdown tiddlers are rendered and displayed in TiddlyWiki, and is not related to loading/save files in the node.js configuration which is a core behaviour.

#4702 says it would be possible with the new markdown plugin by adding the markdown-it-frontmatter plugin to the tw5 plugin. It would be great if this was supported.

saqimtiaz:
> #4702 says it would be possible with the new markdown plugin by adding the markdown-it-frontmatter plugin to the tw5 plugin. It would be great if this was supported.

You are mistaken. As described in that comment, the markdown-it-frontmatter plugin can hide the frontmatter but this would have nothing to do with how the tiddlers are saved to disk and the .meta files created on node.js.

kravlost:
> > #4702 says it would be possible with the new markdown plugin by adding the markdown-it-frontmatter plugin to the tw5 plugin. It would be great if this was supported.
> 
> You are mistaken. As described in that comment, the markdown-it-frontmatter plugin can hide the frontmatter but this would have nothing to do with how the tiddlers are saved to disk and the .meta files created on node.js.

Ah, OK. My mistake.

MaxGyver83:
After playing around with markdown tiddlers, I can tell that this PR fixes #6916 and #6919. Are there any blockers?

Jermolene:
Thanks @MaxGyver83.

@cdruan is this PR ready for another review?

cdruan:
@Jermolene

> @cdruan is this PR ready for another review?

Yes

MaxGyver83:
I think I found a regression:

When some text follows a table without an empty line in between, it's not a problem in WikiText and (remarkable) Markdown tiddlers. But in the [new markdown plugin demo wiki](https://cdruan.github.io/TiddlyWiki5/), the text following the table gets integrated into the table.

Example:

```md
|foo|bar|
|-|-|
|x|y|
(x != y)
```

image

cdruan:
Hi @MaxGyver83

This PR uses markdown-it parser which follows GitHub Flavored Markdown table extension. I checked the spec and it actually says:

> The table is broken at the first empty line, or beginning of another **block-level** structure:

See also [Example 202](https://github.github.com/gfm/#example-202). So what you see is the expected behavior. GitHub's markdown behaves in the same way:

|foo|bar|
|-|-|
|x|y|
(x != y)

MaxGyver83:
@cdruan : No problem for me. But other users (which maybe don't read Markdown specs) might be surprised when (markdown) tiddlers that looked good before look broken after upgrading TiddlyWiki the next time. (Even when it's their own fault.)

MaxGyver83:
Maybe an explanation in the release notes would be sufficient. Or maybe empty lines should be inserted automatically into existing markdown tiddler where necessary after the upgrade?

Jermolene:
Hi @cdruan thank you for the updates. The rationale for the change is strong, but I am still concerned about backwards compatibility. I wonder whether it would be practical to package the previous markdown parser as a legacy plugin that people could opt back into if they needed to? It looks as though the old one would continue to work after applying this PR, so it should be a matter of renaming it.

cdruan:
Hi @Jermolene 

I've repackaged the existing markdown plugin as markdown-legacy, with updates so that it recognizes both `text/markdown` (new default) and `text/x-markdown` MIME types. I also took the liberty of updating the markdowndemo edition to reflect the coming changes by this PR.

The demos are up-to-date. Added two more test cases: "CamelCase Link Text" and "No Nested Links" in the issues section. Please let me know if you have any concerns.

Jermolene:
Thank you @cdruan, I am really grateful for your commitment.

I think the PR might need rebasing because it is currently showing all the changes since v5.2.3.

cdruan:
Hi @Jermolene 

I've sorted out the git mess. This latest push contains only the commits I made.

Jermolene:
Hi @cdruan thank you again for your patience. These are important improvements that I am sure will be welcomed by the community. I won't have a chance to update the release note for a day or two, but if you are able it would be great if you could post to talk.tiddlywiki.org telling users about the changes.

PR #6530, opened 2022-03-15
button-widget: This PR fixes a problem, that the selected class attribute in the DOM isn't refreshed

pmario:
This PR fixes a problem, that the selected class attribute in the DOM isn't refreshed, when the "default" parameter changes when used in the tabs-macro.

While working to improve the tabs-macro, where the button-widget and the refresh-widget have to work together, with the same "default" parameter settings. 

I found out, that the selected class of the button widget isn't updated, while the refresh-widget woks as expected, when the default-parameter kicks in. The result is an inconsistency between the tabs selection and the body content.

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -247,7 +247,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 ButtonWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario

pmario:
@Jermolene Please revert this one. I may have missed something with the parameter, or I did make a logical error. Need some single step debugging. Will have to create a new PR

yaisog:
@Jermolene @pmario: Was this ever reverted? If so, would it show here in this discussion?

pmario:
It has been fixed .. IMO with a second PR

saqimtiaz:
> @Jermolene @pmario: Was this ever reverted? If so, would it show here in this discussion? To me it looks like the commit is still in v5.2.5...

See https://github.com/Jermolene/TiddlyWiki5/pull/6531

PR #6533, opened 2022-03-16
Improve Filters doc and add "Learn more about how to use Filters" link to every operator reference tiddler

pmario:
This PR adds a link `[[Learn more about how to use Filters|Filters]]` to every operator reference tiddler. Some users think, filter expressions can be used similar to links. 

In reality they would need [Filtered Transclusions](https://saqimtiaz.github.io/tw5-docs-pr-maker/#Transclusion%20in%20WikiText), which is now linked in the Filters tiddler under "Find out more". 

The Filters tiddler also contains a "copy to clipboard" element with the `<<list-links "[tag[Filters]]">>` example, that reveals additional info about Filters. 

So in sum, the whole thing is better "interconnected" and should provide more relevant info. 

https://talk.tiddlywiki.org/t/getting-tiddler-field-isnt-working/2746 triggered this PR.

The AdvancedSearch link now opens the AdvancedSearch tiddler, selects the "Filter" tab and adds the right text to the search field. 

The "tagged" link opens the "tag Operator" tiddler. ..

### DIFF
--- a/editions/tw5.com/tiddlers/concepts/Filters.tid
+++ b/editions/tw5.com/tiddlers/concepts/Filters.tid
@@ -1,23 +1,27 @@
 created: 20130827080000000
-modified: 20150221225414000
+list: [[Introduction to filter notation]] [[Filter Syntax]]
+modified: 20220316120141494
 tags: Reference Concepts
 title: Filters
 type: text/vnd.tiddlywiki
-list: [[Introduction to filter notation]] [[Filter Syntax]]
+
+\define openAdvancedSearch()
+<$action-setfield $tiddler="$:/state/tab--1498284803" text="$:/core/ui/AdvancedSearch/Filter"/>
+<$action-setfield $tiddler="$:/temp/advancedsearch/input" text="[tag[Filters]]"/>
+\end
 
 You can think of TiddlyWiki as a database in which the records are tiddlers. A database typically provides a way of discovering which records match a given pattern, and in ~TiddlyWiki this is done with filters.
 
 A <<.def filter>> is a concise notation for selecting a particular [[set of tiddlers|Title Selection]], known as its <<.def "output">>. Whenever ~TiddlyWiki encounters a filter, it calculates the output. Further work can then be done with just those tiddlers, such as [[counting|CountWidget]] or [[listing|ListWidget]] them.
 
-The following example passes a filter to the <<.mlink list-links>> macro to display a list of all tiddlers whose titles start with the letter H:
+The following example passes a filter to the <<.mlink list-links>> macro to display a list of all tiddlers whose titles are <<.olink2 tagged tag>> with the word <<.word Filters>>:
 
-> `<<list-links "[prefix[H]]">>`
+<<wikitext-example-without-html """<<list-links "[tag[Filters]]">>""" >>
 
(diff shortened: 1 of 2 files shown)


pmario:
Filters -- now looks like that: 

grafik

pmario:
Operator references have a link to Filters tiddler now

grafik

Jermolene:
Thanks @pmario

twMat:
Good idea @pmario and well written. Just one thing:

I would like for the text to give a sense of *where filters are used* by including a quick overview of this and where your `list-links` example serves as an illustration of this. Something like this following the second paragraph:

```
Filters are typically used as //arguments// in the following contexts:

* [[widgets|Widgets in WikiText]] that have a `filter` attribute
* [[filtered transclusions|Transclusion in WikiText]] -- which always, and only, take a filter as argument
* macros -- depending on implementation

For example, the following passes a filter [....etc with Marios text...]
```

Thoughts?

pmario:
Sorry Mat ... I'm not really sure, where you want to have your text. 
You did mention `For example, the following passes a filter [....etc with Marios text...]` ... It contains an "etc"... How should I know, what you exactly you want to tell me. ... 

@twMat ... You already have a GitHub account and you did sign the CLA in 2017

 - Please try to open: https://saqimtiaz.github.io/tw5-docs-pr-maker/  
 - Create a personal access token: https://github.com/settings/tokens/new?scopes=public_repo
    - Give it a name
    - Limit it to 7 days, if you  are not sure
    - Copy it over to the form
 - Make your changes to the wiki
 - Backup ... using the buttons in the form
 - Create a PR using the form at: https://saqimtiaz.github.io/tw5-docs-pr-maker/#Make%20a%20PR

It's not possible to break anything. I did create some videos, that describe every step https://www.youtube.com/playlist?list=PLuiC_HFhI4Owy3oY_dZpYe7MGlW5JRGIY

twMat:
@pmario 

>Sorry Mat ... I'm not really sure, where you want to have your text.

Well, I did say: *"Something like this following the second paragraph"* - so there. And the initial words in your *third* paragraph would be somewhat modified per my proposal.

It would be desirable that the PR form could somehow show the tiddler text that I want to modify. 

Anyway, I believe I jumped through all needed hoops but end up with this:

image<br>[snip]
image

<hr>

Also, in your bullet list in the post here above, your *first* bullet instructs:

> * Please try to open: https://saqimtiaz.github.io/tw5-docs-pr-maker/

- but why? I don't see any link or button there to the form so it is good you provided the last bullet in your list with the link so I could access the actual PR form. (And what do you mean with the second last bullet about making backups. How? Save the whole wiki?)

saqimtiaz:
> This PR adds a link `[[Learn more about how to use Filters|Filters]]` to every operator reference tiddler. Some users think, filter expressions can be used similar to links.

Hi @pmario, can you please have a look to see if we have had a regression in how this is displayed? On tiddlywiki.com and the pre-release, the link is appearing on the same line as "New in 5.2.0" message, which is quite confusing.

See https://tiddlywiki5-4omqgzl9p-jermolene.vercel.app/#deserialize%20Operator

Screenshot:
image

pmario:
@saqimtiaz ... I think it has always been like this. The PR adds a link, which in turn creates an inline HTML A element. The `<<.tip` macro also creates an inline element. 

It's an oversight. I'll create a new PR, which will cover the template link into a block-element.

PR #6537, opened 2022-03-17
Fix: only refresh styles if the CSS has changed

saqimtiaz:
This PR improves the style refresh so that we only update the contents of the style tag when the CSS generated by the stylesheet tiddlers has actually changed. We now compare the new CSS to the already assigned CSS and do not update the style tag if they are identical.

Previously, in some situations CSS was being reassigned even when it had not changed - leading to a performance hit which in some scenarios was on every keystroke - as `$tw.styleWidgetNode.refresh()` was returning true due to the refresh handling of the widgets being used in the stylesheets. Some widgets refresh themselves and therefore return true, even though the `textContent` of the resultant DOM is identical.

Testing in FF and Chrome on Windows 10 shows that saving the previous value of the CSS is faster by the finest of margins than directly accessing it via the `innerHTML` of the style tag for comparison purposes.


Jermolene:
Thanks @saqimtiaz

PR #6539, opened 2022-03-18
Update chinese language files

BramChen:
* Add chinese translations for `FieldEditor/*` in ControlPanel.multids.


Jermolene:
Thanks @BramChen

PR #6540, opened 2022-03-18
Fix action-listops unsafe use of $tw.utils.stringifyList()

### LINKED ISSUE #6535: [BUG] ActionListOpsWidget breaks with input titles starting with a dash
The `<$action-listops>` widget breaks with input titles starting with a dash. For example, if the tiddler "target" contains the text `-3` then the following will not work as expected:

```
<$action-listops  $tiddler="target" $field="text" $subfilter="+[toggle[-3]]"/>
```

In this case, the text field of "target" is not left untouched, and the "-3" item is not removed.

But, the same experiment with "foobar" instead of "-3" works as expected.

The problem is here:

https://github.com/Jermolene/TiddlyWiki5/blob/c741978e959c2608fa01dcd881d47194ccdf79b8/core/modules/widgets/action-listops.js#L72-L76

The code makes the assumption that a stringified list can be evaluated as a filter and will get the same results. That assumption fails if the list item starts with a dash, as it is interpreted as the `:except` filter run prefix.

Jermolene:
Fixes #6535

### DIFF
--- a/core/modules/widgets/action-listops.js
+++ b/core/modules/widgets/action-listops.js
@@ -1,3 +1,5 @@
+const { makeTiddlerIterator } = require("../wiki");
+
 /*\
 title: $:/core/modules/widgets/action-listops.js
 type: application/javascript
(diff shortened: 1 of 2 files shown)


core/modules/widgets/action-listops.js
```diff
@@ -1,3 +1,5 @@
+const { makeTiddlerIterator } = require("../wiki");
```
pmario, line 1: Do we allow `const` now?

Jermolene: Wow!! No it's much weirder than that – I didn't (intentionally) write that line of code at all,  as far as I can figure out, it was automatically inserted by VSCode and/or GitHub Copilot.

Jermolene: I've pushed a fix in 681f22b66d1e37c1f4bbaf900ed41b3036e8ae81

pmario: hihi, .. Yea I did disable Copilot again. It did add too much unwanted code and the "autocomplete feature" was constantly interrupting my thought process while programming.

Jermolene: It's equally terrifying and hilarious. It does quite well for the kinds of things I might otherwise have to google, and sometimes thrills me by suggesting very TW-specific idioms, like `$tw.utils.stringifyList()`

PR #6553, opened 2022-03-23
Fix nesting of block elements inside inline elements in the edit template

saqimtiaz:
This PR fixes 3 instances in the edit template where we had block elements inside span elements generated by the $reveal widget.

In `core/ui/EditTemplate/body/default.tid` I have also explicitly changed the second $reveal widget to generate a DIV rather than relying on it being parsed in block mode. This should help avoid accidental regressions due to whitespace/formatting changes in the code.


saqimtiaz:
Before:
image


After:
image

Jermolene:
Thanks @saqimtiaz

PR #6554, opened 2022-03-23
German translation adjustments

pmario:
German translation adjustments


pmario:
We need to include this one into 5.2.2

PR #6555, opened 2022-03-23
Signing the CLA

damscal:


Jermolene:
Thanks @damscal and welcome to the project! Is there anything particularly you're interested in helping with? Do let me know if there's anything I can do to help.

damscal:
Hello Jeremy, thank you very much for the warm welcome.

At the moment I'm working on a few plugins, with these purposes in mind:

- frictionless UI, both keyboard and touchscreen friendly, through the following designes:
  - integration with [Vimium-C](https://github.com/gdh1995/vimium-c). The aim is to allow faster navigation and editing, without having to frequently switch between keyboard and mouse (or touch input). To integrate well with Vimium, it would be great if all clickable TW components had a detailed, descriptive and possibly unique text tooltip that Vimium can access. One issue I've found so far is that for some reasons Vimium doesn't recognize TW's multiline editors (both default and codemirror) as valid input boxes.
  - extensive autocompletion features in all input boxes.
  - custom templates to easily view and edit tiddlers referenced (not just by links and tags, but also macros and transclusions) by the one in focus. Ideally the aim would be that users can seamlessly switch editing from one tiddler to another with the overall feeling of being interacting with just one and structured composite entity, instead of having to manage manually and in detail each individual tiddler. 
- an extensive automation framework for TW, to easily automate repetitive tasks. This could be used to populate tiddler fields with values that are automatically derived from other fields, providing a form of flexible inheritance compatible with TW's Draft mechanism and overall philosophy.

These are ambitious projects, and my knowledge about TW internals is still not deep, so any collaboration with more experienced long time users would be great.
In the meantime, I could also help improving the documentation while I gradually get deeper into the code. I've found that it would be great if system tiddlers could be populated with extensive and quickly understandable comments for developers. It would also be great if some of the most intricate mechanisms where documented using flowcharts such as mermaids'. I could help creating and maintaining those flowcharts.

Thank you @Jermolene for your outstanding work.

Best Wishes,
Dam

pmario:
Hi Dam, ... Welcome!
Just to be sure. Have a look at the dev-wiki at: https://tiddlywiki.com/dev/ ... It contains mainly developer content, that we wanted to keep out of the user docs.

damscal:
Ciao Mario, thank you! I started diving into it.

PR #6556, opened 2022-03-23
Correct typo in Named Filter Run Prefix.tid

damscal:
At the end of file, filerrunprefix --> filterrunprefix


Jermolene:
Thanks @damscal well spotted!

Issue #5599, opened 2021-04-10
[BUG] list-links creates extra links from CamelCase words in caption


kookma:
**Describe the bug**
The list-links macro creates a link to tiddlers from input filter. It shows caption if it is existed instead of title.
The error is occurred when a tiddler caption has missing CamelCase link!

**To Reproduce**
1. Download the attached bundle (list-links-issue.zip) of three tiddlers 
2. Unzip list-links-issue.zip and drag and drop  to  https://tiddlywiki.com/prerelease/
3. Open the Test tiddler
4. See error (the word YouTube is a missing link)

**Expected behavior**
One link expected per title

It seems the in [list-links](https://tiddlywiki.com/prerelease/#%24%3A%2Fcore%2Fmacros%2Flist)

```
<$transclude field="caption">
<$view field="title"/>
</$transclude>
```

shall be replaced with

```
<$view field="caption">
<$view field="title"/>
</$view>
```

I can create a PR if this is correct solution!

**Screenshots**
![image](https://user-images.githubusercontent.com/830394/114272450-77da5280-9a2b-11eb-8257-e61b712a14b9.png)

**TiddlyWiki Configuration (please complete the following information):**
 - Version 5.1.24pre
 - Saving mechanism -
 
**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Windows 10
 - Browser chrome 89

[list-links-issue.zip](https://github.com/Jermolene/TiddlyWiki5/files/6289959/list-links-issue.zip)

sobjornstad:
Just want to add that this happens to me regularly and is really irritating. If the caption is all CamelCase, it is even impossible to go to the intended link target at all.

Jermolene:
I think the best solution here would be for the list-links macro to set the `tv-wikilinks` variable to `no` when rendering the link content, disabling any embedded camelcase links

> It seems the in [list-links](https://tiddlywiki.com/prerelease/#%24%3A%2Fcore%2Fmacros%2Flist)
> 
> ```
> <$transclude field="caption">
> <$view field="title"/>
> </$transclude>
> ```
> 
> shall be replaced with
> 
> ```
> <$view field="caption">
> <$view field="title"/>
> </$view>
> ```

By convention, the caption field is wikified when rendered.

flibbles:
This issue cannot be fixed the way TiddlyWiki is coded right now, so get used to disappointment. This is the world we made for ourselves.

[...]

PR #6558, opened 2022-03-26
Fixed minor issue with list-links and wikilinks

flibbles:
These should be view widgets, not transclude widgets.

If someone has a value like "MyTiddler" as a caption, and they have wikilinks enabled, they'll get these stupid double-nested `<a>` elements that are stupid and broken.

### DIFF
--- a/core/wiki/macros/list.tid
+++ b/core/wiki/macros/list.tid
@@ -7,9 +7,9 @@ tags: $:/tags/Macro
 <$list filter="$filter$" emptyMessage=<<__emptyMessage__>>>
 <$subtype$>
 <$link to={{!!title}}>
-<$transclude field="caption">
+<$view field="caption">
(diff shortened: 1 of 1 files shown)


flibbles:
Oh, and this is the fix for  #5599. Didn't see that I was the only one to call out stupidness when I saw it.

flibbles:
Well great. Never mind.

I'm trying to follow what you said in that other issue about using tv-wikilinks=no instead of disabling wikitext rendering for the caption, but tv-wikilinks disables ALL links, not just wikilinks. So that does not work.

\rules except wikilink also does not work unless you cram it into the caption.

So this issue is just not going to be fixed apparently. Fun.

Jermolene:
Hi @flibbles apologies I should have been clearer, I'll push the fix shortly.

PR #2710, opened 2017-01-13
adds class tc-checkbox-checked when checked

tobibeer:
* equivalent to #2182 (RadioWidget)
* also applies `tc-checkbox` to checkboxes by default, always

PR #6561, opened 2022-03-27
Checkbox widget: list and filter modes

### LINKED ISSUE #5613: [IDEA] CheckboxWidget - generalize to also toggle individual items

rmunn:
Fixes #5613.
Supersedes #5623.

This adds `listField` and `filter` attributes to the checkbox widget. They are described in the documentation tiddler changes included in this PR, but I'll describe them in more depth below:

## List mode

The `listField` attribute works like `field`, but where `field` is treated as a single value, the field whose name is in the `listField` attribute will be treated as a list. The `checked` and `unchecked` attributes will determine which value is looked for in the list to determine if the checkbox is checked, and that value will be either added to removed as the user checks or unchecks the checkbox. It works the way you probably already expect it to work: if the list is a list of colors, say "red yellow blue", then the following combinations of `checked` and `unchecked` will produce the following results:

**List is currently "red yellow blue"**
- `checked="green"`: checkbox is unchecked because "green" is not found in the list. When user checks the box, list becomes "red yellow blue green": the `checked` value is added
- `unchecked="red"`: checkbox is unchecked because "red" is found in the list. When user checks the box, list becomes "yellow blue": the `unchecked` value is removed
- `checked="green" unchecked="red"`: checkbox is unchecked because "red" is found in the list and "green" is not found in the list. When user checks the box, "red" is replaced with "green" so list becomes "green yellow blue"
- neither `checked` nor `unchecked` is specified: checkbox is checked because list is non-empty. An empty list would have produced an unchecked box. Checking the box will not change the list because neither `checked` nor `unchecked` were specified, so this is almost never useful. However, if you have specified `actions` (or `checkactions` and `uncheckactions`) that will modify the list, it might be possible to make this mode useful.

**List is currently "green yellow blue"**
- `checked="green"`: checkbox is checked because "green" is found in the list. When user unchecks the box, list becomes "yellow blue": the `checked` value is removed
- `unchecked="red"`: checkbox is checked because "red" is not found in the list. When user unchecks the box, list becomes "green yellow blue red": the `unchecked` value is added
- `checked="green" unchecked="red"`: checkbox is checked because "green" is found in the list and "red" is not found in the list. When user checks the box, "green" is replaced with "red" so list becomes "red yellow blue"
- neither `checked` nor `unchecked` is specified: as with the other example, checkbox is checked because list is non-empty, but the list field will not be automatically changed and actions must be used if change in the list field is desired.

As you might be able to tell from these examples, when a new value is added to the list it's added at the end. But when an existing value is replaced in the list (changing `green` to `red`) it's changed in-place, so that the ordering of the rest of the list is kept intact.

## Filter mode

The `filter` attribute will contain a filter expression whose output is evaluated just like in list mode: the `checked` value being found in the output means the checkbox is checked, `unchecked` found in the output means it's unchecked, and so on. (And neither one being found means that the filter's output being empty means unchecked, and non-empty means checked).

One difference from list mode is that in filter mode, no changes are automatically made to any field. Instead, the user must specify actions (or checkactions and uncheckactions) that will be run when the user clicks on the checkbox. Most of the time, those actions will involve a `listops` action widget with a subfilter like `subfilter="newValue -oldValue"` (or the reverse, removing `newValue` and adding `oldValue`, for uncheckactions). But it's up to the user to ensure that the actions will cause a change in the filter results, so that clicking on a checked box will end up causing it to be unchecked (or vice-versa) when the filter is re-evaluated after the actions have run.

## Indeterminate checkboxes

In my current implementation, both list and filter modes will produce "indeterminate" checkboxes if both `checked` and `unchecked` are specified yet neither one of them is found in the list. (See [the MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes) and scroll down to the example to see what indeterminate checkboxes look like.) That's definitely useful in filter mode, but I could see that being undesirable in list mode. So I'd like feedback. In the following scenario, what would you expect to happen:

- The `colors` field contains "yellow blue"
- `<$checkbox listField="colors" checked="green" unchecked="red" />`

Would you expect the checkbox to be unchecked? Or show an "indeterminate" checkbox rather than an unchecked one?


rmunn:
Differences between this and PR #5623, which this supersedes: in #5623, I planned to implement a `mode` attribute which could be `list` to turn the `field` attribute into list mode. This would also have worked with the `index` attribute. In this PR, I don't have a `mode` attribute, and the mode is instead chosen by whether the `field` or `fieldList` attribute is specified. And, since @Jermolene said in https://github.com/Jermolene/TiddlyWiki5/issues/6521#issue-1165338585 that he wants to deprecate the `index` attribute in widgets, I did not add a `listIndex` feature. Instead, anyone who wants to use lists kept in data tiddler indexes will be able to do so with the `filter` attribute, and will have to specify correct actions in `checkactions` and `uncheckactions` to ensure that the data tiddler gets the correct updated value.

rmunn:
Looks like I'm just too late to get this into 5.2.2. I'll change the "new in 5.2.2" markers to "new in 5.2.3".

editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
```diff
@@ -2,9 +2,25 @@ caption: checkbox
 created: 20131024141900000
 modified: 20211009121239806
 tags: Widgets TriggeringWidgets
+colors: red orange yellow blue
 title: CheckboxWidget
 type: text/vnd.tiddlywiki
 
+\define checkActions() <$action-listops $field="colors" $subfilter="-red green"/>
```
Jermolene, line 9: Having the macro definitions separated from the examples is awkward.

The wikitext-example-without-html macro creates a new parsing context, so it is possible to make the examples self-contained by including the necessary definitions within. Is that possible in this case?

editions/test/tiddlers/tests/test-widget.js
```diff
@@ -805,6 +822,350 @@ describe("Widget module", function() {
 		// the <<qualify>> widget to spit out something different.
 		expect(withA).toBe(withoutA);
 	});
+
```
Jermolene, line 825: I wonder if these tests should go in their own file. I added a test-action-widgets.js file recently, but this is really a hybrid. I think it would make sense to make a new `test-checkbox-widget.js` file.

Jermolene (commented):
Thanks @rmunn

> And, since @Jermolene said in https://github.com/Jermolene/TiddlyWiki5/issues/6521#issue-1165338585 that he wants to deprecate the index attribute in widgets

I think in this case consistency trumps that concern, and it would make sense to have the companion property for handling indexes.

editions/tw5.com/tiddlers/widgets/CheckboxWidget.tid
rmunn: Yeah, it certainly was awkward having to write it this way, with the macro definitions at the top of the tiddler. I'm glad to learn about the new parsing context (did not know that before); I'll rewrite these examples using that fact.

editions/test/tiddlers/tests/test-widget.js
rmunn: Sure. The tests were getting long enough, that even by refactoring them to get as much shared code de-duplicated as I could, they were starting to be a significant percentage of the file. I'll have to duplicate a little bit of the `test-widget` code, specifically the helper functions, but I agree that it makes sense for the checkbox tests to be in their own file.

rmunn:
> I think in this case consistency trumps that concern, and it would make sense to have the companion property for handling indexes.

Will do. You'll see another commit or two tomorrow. Possibly today, but I don't think I'll have enough time today to get to it.

twMat:
Great work @rmunn ! I must admit I had not comprehended the complexity of this. (I can't help but wonder how the existing checkbox implementation distinguishes from checkboxing individual items in the *tags* field from any other field? And will checkboxes behave differently for individual tags if I use this new solution vs the existing?)

>[...] when a new value is added to the list it's added at the end. But when an existing value is replaced in the list (changing green to red) it's changed in-place, so that the ordering of the rest of the list is kept intact.

I think this is a good design decision and how one expects it to behave.

>* The colors field contains "yellow blue"
>* <$checkbox listField="colors" checked="green" unchecked="red" />

Spontaneously, I would expect it to behave like the existing checkbox implementation does for this:

`<$checkbox field="myfield" checked="green" unchecked="red" />`

in other words it is left unchecked. Now it could be argued that the existing checkbox implementation should not behave like this.

Side note: The 6'th bullet in the OP should probably read "...because red is *not* found..."

tw-FRed:
Hi @rmunn, thanks for this great enhancement!

> In the following scenario, what would you expect to happen:
> 
>     The colors field contains "yellow blue"
>     <$checkbox listField="colors" checked="green" unchecked="red" />
> 
> Would you expect the checkbox to be unchecked? Or show an "indeterminate" checkbox rather than an unchecked one?

I would expect the checkbox to be in "indeterminate" state, although I find Firefox's appearance for "indeterminate" checkboxes disturbing.

rmunn:
There's another possible answer that I just thought of, and that's to have an `indeterminate="foo"` attribute alongside `checked` and `unchecked`. I.e., `checked="yes" unchecked="no" indeterminate="maybe"`. The `indeterminate` value (here, "maybe") can never be assigned to a field by TW's default behavior when the user checks or unchecks the box: that will always assign the `checked` or `unchecked` value (as the sole contents of the field for `field="something"`, or as part of a list for `listField="something"`). The only way for the indeterminate value to end up in a field would be as a result of action widgets like `listops` or `setfield` being triggered by other checkboxes. That's because the most common use for indeterminate checkboxes is things like this:

* [ ] First task
    * [ ] Sub-task 1
    * [ ] Sub-task 2
* [ ] Second task

If sub-task 1 is checked but sub-task 2 is not checked, then you'd want "first task" to be in an indeterminate state. Then you hook up logic to the checkboxes as follows: when "first task" is checked, sub-task 1 and sub-task 2 get checked. When "first task" is unchecked, sub-task 1 and sub-task 2 get unchecked. And when sub-task 1 or sub-task 2 are changed, they look at the list that they're a part of: if that list's length is 0 then "first task" becomes unchecked. If that list's length is the maximum it should be (in this case, 2) then "first task" becomes checked. But if the list's length is greater than 0 but less than 2, then "first task" gets assigned an indeterminate value.

With the design I currently have, it's actually a little tricky to implement that scenario, because you have to have "first task" be part of a list-mode checkbox. (Because with my current design, there's no way for a "normal", single-value checkbox to be indeterminate). Even if you really wanted just one task (and "second task" wasn't there at all), you'd still have to write your TW widget in such a way as to have the "first tasK" be a list-mode or filter-mode checkbox widget, where `checked="yes"` and `unchecked="no"` and then the sub-task actions would assign `"maybe"` to it if their group was only partly checked.

But by adding an `indeterminate="foo"` attribute, I gain the following benefits:

* Indeterminate values are explicitly opt-in: you'll never see them unless you ask for them
* Single-value checkboxes (with `field` or `index` attributes) can now have indeterminate values if you want
    * But becuase it's opt-in, existing code won't be affected
* No confusion about what to do when neither the `checked` nor `unchecked` values appear
    * In that situation, if `indeterminate` attribute is present, checkbox is indeterminate, otherwise checkbox is unchecked

So what I plan to do is *remove* the indeterminate behavior from this PR, and introduce it in a new PR. That way there can be two entries in the release notes: one about checkboxes acquiring list mode and filter mode, and a second release notes entry about checkboxes acquiring the `indeterminate` attribute. Because those are really two different features, and should have separate entries in the release notes.

rmunn:
I've created #6564 to discuss the design of adding an `indeterminate` attribute to the checkbox widget. My intent is to remove the "indeterminate" code from this PR, so that it only adds list and filter modes to checkboxes. Any situation that would cause a checkbox to be indeterminate will, in this PR, cause it to be false. Once people have discussed #6564 and there's broad agreement on how it should be done, I'll create a new PR that implements the design from #6564 and submit that PR.

One advantage of having just barely missed the release window for 5.2.2 is that there's probably going to be seveeral months before 5.2.3 is released, so there's no need to rush in implementing the "indeterminate" attribute. I can allow a week or so for discussion on #6564 and then implement it (which won't be hard as I've already done most of the work in this PR), and both features will still be able to land in 5.2.3.

rmunn:
@twMat wrote:
> Side note: The 6'th bullet in the OP should probably read "...because red is _not_ found..."

Good catch! Fixed.

rmunn:
I've written the tests and code for the `listIndex` attribute. I have **NOT** yet written documentation for it, but I'm running out of time to work on TW today. I'll continue this PR later in the week. @Jermolene, if you want to take a look at the code I wrote to implement `listIndex`, it's ready to be reviewed. It's not quite in its final form yet as I'm going to be removing the part that implements indeterminate checkboxes (which I've implemented as returning `undefined` from `getValue`). But that change will be straightforward, and shouldn't impair your ability to review the code as it stands now. Then if you want to request any changes, I can get a head start on them.

editions/test/tiddlers/tests/test-widget.js
rmunn: Done.

rmunn:
I've incorporated the change from https://github.com/Jermolene/TiddlyWiki5/pull/2710 into this PR, and made sure that @tobibeer gets credit for the work he did which I pulled into commit 07d001c. (I originally had created commit 4397291 but realized that I'd forgotten to use `git comit --author` to give @tobibeer credit, so I amended that commit and force-pushed the branch).

I'm going to make one more change, to incorporate the suggestion from https://github.com/Jermolene/TiddlyWiki5/pull/6561#discussion_r835882431, and then this PR will be ready to be reviewed one final time and then merged.

rmunn:
Experimenting with the `listIndex` example has shown me that there's a bug in my code so far: that example starts off with all three boxes unchecked, but as soon as you check one box, the other two spontaneously become checked as well. That's incorrect behavior, and I'm trying to track down the bug now.

**Please do not merge** this PR until I have nailed down that bug.

rmunn:
@Jermolene I'm happy with this PR now. I'm ready for you to review it and, if you're happy with it, merge it in. Then I'll create another PR to address #6564.

Jermolene:
Great thanks @rmunn

PR #6567, opened 2022-03-28
Fix regressions cause by PR #6511

FlashSystems:
This fixes the regressions introduced by PR #6511 

See https://talk.tiddlywiki.org/t/possible-field-editing-bug-in-5-2-2/2884 and https://talk.tiddlywiki.org/t/extra-space-between-fields-row-in-5-2-2-versus-5-2-1/2879/5

### DIFF
--- a/core/ui/EditTemplate.tid
+++ b/core/ui/EditTemplate.tid
@@ -5,7 +5,7 @@ title: $:/core/ui/EditTemplate
 \define save-tiddler-actions()
 \whitespace trim
 <$action-sendmessage $message="tm-add-tag" $param={{{ [<newTagNameTiddler>get[text]] }}}/>
-<$action-sendmessage $message="tm-add-field" $name={{{ [<newFieldNameTiddler>get[text]] }}} $value={{{ [<newFieldValueTiddler>get[text]] }}}/>
+<$action-sendmessage $message="tm-add-field" $name={{{ [<newFieldNameTiddler>get[text]] }}} $value={{{ [<newFieldNameTiddler>get[text]] :map[<newFieldValueTiddler>get<currentTiddler>] }}}/>
 <<delete-edittemplate-state-tiddlers>>
 <$action-sendmessage $message="tm-save-tiddler"/>
 \end
--- a/core/ui/EditTemplate/fieldEditor-default.tid
+++ b/core/ui/EditTemplate/fieldEditor-default.tid
@@ -1,3 +1,3 @@
 title: $:/core/ui/EditTemplate/fieldEditor/default
 
-<$edit-text tiddler=<<currentTiddler>> field=<<currentField>> default="" class="tc-edit-texteditor" placeholder={{$:/language/EditTemplate/Fields/Add/Value/Placeholder}} tabindex={{$:/config/EditTabIndex}} cancelPopups="yes"/>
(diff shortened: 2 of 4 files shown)


FlashSystems:
Some notes about the two bugs:

## Fix field edit bug

The `$:/core/ui/EditTemplate` contains the `save-tiddler-actions` that saves the pending field, if there is one. This happens by accessing `newFieldNameTiddler` and `newFieldValueTiddler` across a transclusion and the edit body cascade. I think this construction is prone to errors like this. Especially if someone is customizing the edit template. Maybe this should be refactored.

## Fix UI-style regression

I'm sorry, I could not come up with a better way to fix this issue, than adding another variable. I would have liked to keep the field edit template agnostic about its use (add or edit) to make it simpler to customize.

Jermolene:
Thanks @FlashSystems 

> ## Fix field edit bug
> The `$:/core/ui/EditTemplate` contains the `save-tiddler-actions` that saves the pending field, if there is one. This happens by accessing `newFieldNameTiddler` and `newFieldValueTiddler` across a transclusion and the edit body cascade. I think this construction is prone to errors like this. Especially if someone is customizing the edit template. Maybe this should be refactored.

I quite agree, it's really not good to have dependencies spread around like that. I'd certainly be open to refactoring.

> ## Fix UI-style regression
> I'm sorry, I could not come up with a better way to fix this issue, than adding another variable. I would have liked to keep the field edit template agnostic about its use (add or edit) to make it simpler to customize.

Eeek, that does seem clumsy.

I think the problem caused by adding the `tc-edit-texteditor` class is just the margins, padding and line height. Adopting the shaded background and lighter borders does actually seem more consistent. With that in mind, might it be possible to refactor things to avoid the new variable?

FlashSystems:
> Eeek, that does seem clumsy.

Full ack. I don't like this solution either. But it's the only solution that is completely regression safe, because the generated HTML is the same as in TW 5.2.1.

For the CSS based solution, I would use an additional class `tc-edit-fields` that is added to the edit elements that are there to edit field values. This makes sure no change is done to `tc-edit-texteditor` because this class is used on many occasions.

This is the original TW5 5.2.1 look in Chrome:
Original

This is the look in TW5 5.2.2:
Regression

This is the look with the CSS approach. Adding `tc-edit-fieldeditor` that resets `margin`, `line-height` and `padding` if the element is a child of an element with class `.tc-edit-fields`.
All Reset - Uneven
Please note, that the line height is uneven because the line-height does not work for the date-input and the select-element on chrome.

If the line-height is not reset. There spacing is a little wieder, but looks more even because the select-element is the same size as the input elements. The date-input is still a little higher (at least in Chrome).
This is a deviation from the original size nonetheless.
Keep Line-Height

@Jermolene: Please tell me which one you prefer and I'll update this pull request accordingly.

I would go with the last one, and not reset the line-height. Mostly because resetting the line height means copying the value from `reset.tid` creating another duplication and refactoring hazard.

pmario:
>I would go with the last one, and not reset the line-height. Mostly because resetting the line height means copying the value from reset.tid creating another duplication and refactoring hazard.

I would also favour this solution. So +1 from my side ;)

Jermolene:
Hi @FlashSystems I agree, the 3rd option seems the most attractive.

FlashSystems:
I've updated this pull request with the preferred solution. The new class is `tc-edit-fieldeditor`. I've also updated the documentation to mention this class for authors of new field editors.

I've testet the CSS on current Chrome and Firefox.

Jermolene:
Great stuff, thank you @FlashSystems

PR #6581, opened 2022-04-01
Add "enabled" attribute to draggable widget

BurningTreeC:
This PR adds the `enabled` attribute to the `draggable` widget
By default, the domNode will be drag-enabled
Only if `enabled="no"` the domNode won't be draggable

### DIFF
--- a/core/modules/widgets/draggable.js
+++ b/core/modules/widgets/draggable.js
@@ -82,7 +82,7 @@ DraggableWidget.prototype.execute = function() {
 	this.endActions = this.getAttribute("endactions");
 	this.dragImageType = this.getAttribute("dragimagetype");
 	this.dragHandleSelector = this.getAttribute("selector");
-	this.dragEnable = (this.getAttribute("enable") || "yes") === "yes";
+	this.dragEnable = this.getAttribute("enable","yes") === "yes";
 	// Make the child widgets
 	this.makeChildWidgets();
 };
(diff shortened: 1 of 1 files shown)


core/modules/widgets/draggable.js
```diff
@@ -89,7 +92,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 DraggableWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.tag || changedAttributes["class"]) {
+	if(changedAttributes.tag || changedAttributes["class"] || changedAttributes.enabled) {
```
saqimtiaz, line 95: I am noticing that our refresh handling is not complete from before. Could you change this line to:

if($tw.utils.count(changedAttributes > 0) {

so that we refresh when any of the attributes change.

BurningTreeC: Sure thing

BurningTreeC:
Now I've changed the attribute to be `enable` so that it's consistent with the `droppable` widget.
By default the domNode will be draggable. If the `enable` attribute is specified and it isn't "yes" then the domNode won't be draggable and also the class `tc-draggable` won't be applied

@Jermolene, I think this is good to go now. I will add the docs in a later step

BurningTreeC:
@Jermolene - this is ready to merge. For the docs I have to wait for #6582 to be merged, otherwise we get merge conflicts. I'll create a separate docs-PR then

Jermolene:
Thanks @BurningTreeC

PR #6583, opened 2022-04-01
Add dragenteractions and dragleaveactions ...

BurningTreeC:
... to droppable widget

The `refresh` method needs some work here, too?

### DIFF
--- a/core/modules/widgets/droppable.js
+++ b/core/modules/widgets/droppable.js
@@ -67,6 +67,9 @@ DroppableWidget.prototype.enterDrag = function(event) {
 	}
 	// If we're entering for the first time we need to apply highlighting
 	$tw.utils.addClass(this.domNodes[0],"tc-dragover");
+	if(this.dragEnterActions) {
+		this.invokeActionString(this.dragEnterActions,this,event);
+	}
 };
 
 DroppableWidget.prototype.leaveDrag = function(event) {
@@ -79,6 +82,9 @@ DroppableWidget.prototype.leaveDrag = function(event) {
 		this.currentlyEntered = [];
 		if(this.domNodes[0]) {
 			$tw.utils.removeClass(this.domNodes[0],"tc-dragover");
+			if(this.dragLeaveActions) {
+				this.invokeActionString(this.dragLeaveActions,this,event);
+			}
 		}
 	}
 };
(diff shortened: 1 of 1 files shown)


saqimtiaz:
One thing to consider here is that our detection of "dragleave" does not work reliably on Firefox and is sometimes not triggered (visually indicated by the dragover class and associated effects persisting even when an element is no longer being dragged over). The dropzone widget had the same problem, for which I implemented a workaround when working on image import in the editor. However, we still need to resolve it for the droppable widget.

In this context it will mean that dragleave actions will not reliably be triggered in Firefox. It may be prudent to resolve those underlying issues before we add support for dragenter/leave actions.

pmario:
https://stackoverflow.com/questions/10253663/how-to-detect-the-dragleave-event-in-firefox-when-dragging-outside-the-window

There seems to be a discussion at stackoverflow, which also points to some bugs at bugzilla. ..

saqimtiaz:
There is discussion of the issue as seen in the dropzone widget and the fix implemented in the PR description at https://github.com/Jermolene/TiddlyWiki5/pull/5688

BurningTreeC:
> One thing to consider here is that our detection of "dragleave" does not work reliably on Firefox and is sometimes not triggered (visually indicated by the dragover class and associated effects persisting even when an element is no longer being dragged over). The dropzone widget had the same problem, for which I implemented a workaround when working on image import in the editor. However, we still need to resolve it for the droppable widget.

Hi @saqimtiaz , is there a way to resolve it for the droppable widget?

> 
> In this context it will mean that dragleave actions will not reliably be triggered in Firefox. It may be prudent to resolve those underlying issues before we add support for dragenter/leave actions.

saqimtiaz:
> Hi @saqimtiaz , is there a way to resolve it for the droppable widget?

I posted a link above to how we resolved it for the dropzone widget. We would need to see if that technique is transferrable to the droppable widget.

BurningTreeC:
> > Hi @saqimtiaz , is there a way to resolve it for the droppable widget?
> 
> I posted a link above to how we resolved it for the dropzone widget. We would need to see if that technique is transferrable to the droppable widget.

Thank you @saqimtiaz , but I'm having a hard time to understand what's going on and how to implement this for the droppable widget

PR #6586, opened 2022-04-03
Pluginlibrary local test edition

pmario:
- This PR adds a possibility to test the plugin library import mechanism locally. 
- This PR can highly simplify the development of the import UI. 
- This PR was part of  **Translateable plugins - first try** #5418 

To test the whole configuration you need to do the following: 

- `cd /your/path/to/tiddlywiki5/editions/pluginlibrary-server`
- `npm test`

It will build the needed components and start a server at https://localhost:8080 where you can locally test the library import mechanism.

There is a tiddler, with a [Remove plugins](http://localhost:8080/#Remove-all-plugin-modules) button, that allows you to remove all imported plugins, to run a new test

grafik


Jermolene:
Hi @pmario this seems like a lot of complexity.

I think it would be simpler just to add a new option to the `--listen` command that allows the path to the `files` folder to be customised. Then one could simply clone the existing `.editions/pluginlibrary --build library` build target and add a listen command to the end to run the plugin server.

PR #6587, opened 2022-04-03
Trim Saga: Another batch of whitespace trims

flibbles:
Here's another load. Instead of doing these in batches of five, I'll just push this PR. Once this is merged, I'll move onto the next batch.


Jermolene:
Thanks @flibbles

PR #6588, opened 2022-04-03
pluginlibrary - minimal test server

pmario:
This PR provides a minimal server, to test the plugin import mechanism. 

To use it: 

The following commands will build the files needed **and** it starts a local test server at http://localhost:8888

```
cd /your/path/to/tw5/editions/pluginlibrary
tiddlywiki --build test-server
```

@Jermolene ... This is the minimal version I can come up with. From my point of view, the server has to be "read only". I don't want to risk unknown side effects, that could mess up the plugin library creation process


pmario:
@Jermolene ... bump for v5.2.3

pmario:
Original draft was: https://github.com/Jermolene/TiddlyWiki5/pull/6586 ... This PR was made after your comments at the draft

.gitignore
```diff
@@ -3,3 +3,4 @@
 tmp/
 output/
 node_modules/
+editions/pluginlibrary/files/
```
Jermolene, line 6: It might be better to include `tmp` or `output` in the path and thus avoid having to update this file at all.

editions/pluginlibrary/tiddlers/GettingStarted.tid
```diff
@@ -0,0 +1,11 @@
+title: GettingStarted
+
+This server let's you test the plugin import mechanism.
```
Jermolene, line 3: "let's" shouldn't have an apostrophe. Clearer to say:

This edition contains the components needed to set up a local plugin library for testing or development.

editions/pluginlibrary/tiddlers/GettingStarted.tid
```diff
@@ -0,0 +1,11 @@
+title: GettingStarted
+
+This server let's you test the plugin import mechanism. 
+
+!! Important
+
+''Content isn't saved back to the server. It's read only!''
```
Jermolene, line 7: Please avoid exclamation marks. They should not be used in technical documentation.

Also we don't say "isn't", we spell it out in full as "is not".

Clearer to say:

The default setup is read-only. Nothing is saved back to the server.

editions/pluginlibrary/tiddlers/PrereleaseLocalPluginLibrary.tid
```diff
@@ -0,0 +1,6 @@
+title: $:/config/LocalPluginLibrary
+tags: $:/tags/PluginLibrary
+url: http://localhost:8080/files/local/library/test/index.html
+caption: TIDDLYWIKI Plugin Library (Local)
```
Jermolene, line 4: We never capitalise TiddlyWiki.

editions/pluginlibrary/tiddlers/meta/SiteSubtitle.tid
```diff
@@ -0,0 +1,3 @@
+title: $:/SiteSubtitle
+
+<<version>> - SERVER
```
Jermolene, line 3: The capitalisation here looks ugly.

editions/pluginlibrary/tiddlers/meta/SiteTitle.tid
```diff
@@ -0,0 +1,3 @@
+title: $:/SiteTitle
+
+Pluginlibrary Test
```
Jermolene, line 3: Better to say:

A plugin library for testing and development

Jermolene (commented):
Thanks @pmario

.gitignore
pmario: The problem here is, that the following commands will start the nodejs test server. As far as I know, it only serves files from the `files` subdirectory. Or do I miss something? ... It would be simple to use the output dir, if we can server files from there. 

```
cd /your/path/to/tw5/editions/pluginlibrary
tiddlywiki --build test-server
```

pmario: @Jermolene ... I did create a new route, that lets us serve the test content from the `tmp` folder, relative to the `editions/pluginlibrary` folder. So it will be excluded with the existing .gitignore settings. ... 

But it duplicates the code from the existing $:/core/modules/server/routes/get-file.js file into the pluginlibrary edition. So it will only be available for this edition. 

... I think that approach is a bit hacky. ... I'm not sure about it, if our users start to use it for their own setups. If it's local only I would be OK, but if they expose it to the web ?!

What do you think?

pmario:
I did convert it to a "draft" because of the new route code.

pmario:
as written at: https://github.com/Jermolene/TiddlyWiki5/pull/6588#discussion_r917279790 -- I'm not sure if you saw it. So I open it for review again

-------

@Jermolene ... I did create a new route, that lets us serve the test content from the tmp folder, relative to the editions/pluginlibrary folder. So it will be excluded with the existing .gitignore settings. ...

But it duplicates the code from the existing $:/core/modules/server/routes/get-file.js file into the pluginlibrary edition. So it will only be available for this server edition.

... I think that approach is a bit hacky. ... I'm not sure about it, if our users start to use it for their own setups. If it's local only I would be OK, but if they expose it to the web ?!

What do you think?

Jermolene:
> @Jermolene ... I did create a new route, that lets us serve the test content from the tmp folder, relative to the editions/pluginlibrary folder. So it will be excluded with the existing .gitignore settings. ...

Is that necessary? My understanding was that our `.gitignore` setting will ignore any path that has `tmp` as a component, not necessarily just the root tmp folder. So I'd have expected to be able to serve from `editions/plugin-library/files/tmp/` and have all the output ignored by git.

Jermolene:
Hi @pmario sorry for not thinking of it earlier, but I think it is confusing to call the edition "pluginlibrary". Perhaps "test-plugin-library" would be better?

pmario:
@Jermolene ... Is there an option to make the `--output` parameter to be relative to the `tiddlywiki.info` file?

pmario:
@Jermolene ... bump

pmario:
@Jermolene bump

pmario:
@Jermolene pls merge or comment

Jermolene:
> @Jermolene ... I did rename the edition to `test-pluginlibrary`. I did add a bit more text to the GettingStarted tiddler.

Oh my apologies, I had completely forgotten that "pluginlibrary" is an existing edition, I am afraid that we shouldn't change the edition name because of backwards compatibility. I am sorry to ask you to change it back again.

> @Jermolene ... Is there an option to make the `--output` parameter to be relative to the `tiddlywiki.info` file?

No, but it is already relative to the `output` folder, which is at a fixed location relative to the `tiddlywiki.info` file. Isn't that sufficient?

Otherwise, I think this is good to go.

Jermolene:
Actually perhaps we should add a brief docs tiddler to the /dev edition called "How to run a local plugin library for testing".

pmario:
@Jermolene ... 

- edition is renamed back to "pluginlibrary" 
- GettingStarted text is improved a bit
- The "dev-edition" has a new "How to run a local plugin library for testing", which is very similar to GettingStarted. 

I did test the different commands and they did work for me.

It's important that the pluginlibrary uses port=8888

Jermolene:
Thanks @pmario

PR #6589, opened 2022-04-04
Show server response as error message in put saver

simonbaird:
I'd like to use this on Tiddlyhost so users can get more informative error messages if the put save fails for whatever reason.

This would make the put saver a viable replacement for the legacy upload saver, which is what Tiddlyhost uses currently.

I'm not sure what standard WebDAV servers do, but I would guess they don't provide any response body for put requests, and hence this patch would have no impact for a standard WebDAV server.

(That said, it would be a good idea to test it to make sure there aren't any unexpected regressions for WebDAV or other put saver compatible services.)

Also includes two related small tweaks. See commit messages for details.

### DIFF
--- a/core/modules/savers/put.js
+++ b/core/modules/savers/put.js
@@ -80,6 +80,7 @@ PutSaver.prototype.save = function(text,method,callback) {
 	if(this.etag) {
 		headers["If-Match"] = this.etag;
 	}
+	$tw.notifier.display("$:/language/Notifications/Save/Starting");
 	$tw.utils.httpRequest({
 		url: this.uri(),
 		type: "PUT",
--- a/core/modules/savers/put.js
+++ b/core/modules/savers/put.js
@@ -88,16 +88,20 @@ PutSaver.prototype.save = function(text,method,callback) {
 		callback: function(err,data,xhr) {
 			if(err) {
 				// response is textual: "XMLHttpRequest error code: 412"
-				var status = Number(err.substring(err.indexOf(':') + 2, err.length))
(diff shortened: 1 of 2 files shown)


simonbaird:
Second commit is kind of unrelated, so let me know if you prefer it removed from this PR.

Jermolene:
Thanks @simonbaird looks OK to me, but it would indeed be helpful to test it with WebDAV backends. I think @pmario and @saqimtiaz have done some WebDAV experiments in the last year or two?

saqimtiaz:
I use WebDAV pretty heavily and can do some testing this week. I don't foresee any issues though.

pmario:
I'm using WebDav for some of my wikis and IIS as backend. I'll test it and report back

simonbaird:
I added another commit to add a "Save started" notification for put saving. See commit message for details.

saqimtiaz:
> I'm not sure what standard WebDAV servers do, but I would guess they don't provide any response body for put requests, and hence this patch would have no impact for a standard WebDAV server.

I can confirm that a response to a PUT request from a WebDAV server has no response body for a successful request. A bit trickier to recreate error conditions, but will have another look later in the week.

Jermolene:
Thanks everyone I think that we can merge this now, but obviously further testing is welcome.

simonbaird:
If you want to see this in action on Tiddlyhost, follow the steps [here](https://github.com/simonbaird/tiddlyhost/issues/219#issuecomment-1089669645).

PR #6591, opened 2022-04-04
Make any scrollable element with the class ".tc-scroll-container" scroll

BurningTreeC:
This PR makes any element with the class `tc-scroll-container` scrollable by the page scroller

### DIFF
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -69,7 +69,13 @@ Handle a scroll event hitting the page document
 PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	var self = this,
 		duration = $tw.utils.hop(options,"animationDuration") ? parseInt(options.animationDuration) : $tw.utils.getAnimationDuration(),
-		srcWindow = element ? element.ownerDocument.defaultView : window;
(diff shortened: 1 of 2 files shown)


core/modules/utils/dom/scroller.js
```diff
@@ -69,7 +69,13 @@ Handle a scroll event hitting the page document
 PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	var self = this,
 		duration = $tw.utils.hop(options,"animationDuration") ? parseInt(options.animationDuration) : $tw.utils.getAnimationDuration(),
-		srcWindow = element ? element.ownerDocument.defaultView : window;
+		srcWindow = element ? element.ownerDocument.defaultView : window,
+		closestElement = element.closest(".tc-scroll-container");
```
saqimtiaz, line 73: I haven't had the opportunity to review the merits of the approach yet, but just a quick heads up that .closest() is not available in IE11 and cannot be used in the core. However I do not recommend putting more effort into this before you have had feedback.

BurningTreeC: Thanks @saqimtiaz , yes I'm aware that .closest() is not available in IE11 ...
There are possible workarounds, which are simple enough but as you say, I won't put more effort into this before I have had feedback

PR #6593, opened 2022-04-05
Allow checkboxes to be indeterminate

### LINKED ISSUE #6564: [IDEA] Checkboxes can be in an "indeterminate" state
In https://github.com/Jermolene/TiddlyWiki5/pull/6561#issuecomment-1080122266 I made the decision that #6561 will not allow checkboxes to be in an "indeterminate" state, and that that feature will live in a different PR. This issue is to track that feature; I'll discuss my planned design of the `indeterminate` attribute here, then create a new PR with that implementation once the design has settled.

The checkbox widget will grow a new `indeterminate` attribute. If that attribute is present and has a non-empty value, then the checkbox can be in [an indeterminate state](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox#indeterminate_state_checkboxes). If the attribute is absent, or is present but its value is the empty string, then the checkbox can only be true (checked) or false (unchecked).

The rules for indeterminate checkboxes are as follows:
* If the `indeterminate` attribute is present, *and* the `checked` and `unchecked` attributes are both present, *and* the value the checkbox is looking at (whether that's a `field`, an `index`, a `listField`, a `listIndex`, or a `filter` result) contains neither the `checked` nor `unchecked` value, then the checkbox will be indeterminate.
* Under *any other conditions*, the checkbox will have a determinate value, either checked or unchecked.

Note that TW's way of handling "bare" attributes (attributes with no equal sign) is to assign them the value of `"true"` (a string containing the letters "t", "r", "u", "e"). That means that `<$checkbox indeterminate checked="yes" unchecked="no">` would enable the indeterminate feature.

Note also that the value of the `indeterminate` attribute is **ignored** other than being checked for a non-empty string. That means that if you have `<$checkbox indeterminate checked="yes" unchecked="no">`, which is equivalent to `<$checkbox indeterminate="true" checked="yes" unchecked="no">`, then a value of `"maybe"` will set the checkbox to the indeterminate state. (As would any value other than `"yes"` or `"no"`). The reason for that is to make it easy to assign the result of a filter to the `indeterminate` attribute using `{{{ [filter-expression] }}}` syntax. There's already plenty of precedent in filters for interpreting an empty value as "false" and a non-empty value as "true", such as the behavior of the `else` operator. So I've decided that the `indeterminate` attribute on checkbox widgets should function the same way.

In #6561, I considered making `indeterminate` a feature of list mode and filter mode *only*, but ended up deciding to allow people to create indeterminate checkboxes in single-value mode as well. The only way to do that safely, without changing the behavior of existing wikis where people *will not* be expecting indeterminate values, is to make it an opt-in attribute. The fact that it will also simplify logic like the "sub-task 1 / sub-task 2" example in https://github.com/Jermolene/TiddlyWiki5/pull/6561#issuecomment-1080122266 (the most common reason for wanting indeterminate checkboxes) is a nice bonus.

rmunn:
Fixes #6564.

See the example at the bottom of the CheckboxWidget tiddler for why this is useful: it allows lists of categories, where checking items in a category will check the category box if the category is filled, uncheck it if the category is empty, and leave it in an "indeterminate" state if the category is only *partly* filled.

### DIFF
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -27,6 +27,7 @@ CheckboxWidget.prototype = new Widget();
 Render this widget into the DOM
 */
 CheckboxWidget.prototype.render = function(parent,nextSibling) {
+	var isChecked;
 	// Save the parent dom node
 	this.parentDomNode = parent;
 	// Compute our attributes
@@ -38,9 +39,13 @@ CheckboxWidget.prototype.render = function(parent,nextSibling) {
 	this.labelDomNode.setAttribute("class","tc-checkbox " + this.checkboxClass);
 	this.inputDomNode = this.document.createElement("input");
 	this.inputDomNode.setAttribute("type","checkbox");
-	if(this.getValue()) {
+	isChecked = this.getValue();
+	if(isChecked) {
 		this.inputDomNode.setAttribute("checked","true");
 		$tw.utils.addClass(this.labelDomNode,"tc-checkbox-checked");
+	}
+	if(isChecked === undefined && this.checkboxIndeterminate) {
+		this.inputDomNode.indeterminate = true;
 	}
 	if(this.isDisabled === "yes") {
 		this.inputDomNode.setAttribute("disabled",true);
@@ -93,6 +98,14 @@ CheckboxWidget.prototype.getValue = function() {
 			if(this.checkboxUnchecked && !this.checkboxChecked) {
 				return true; // Absence of unchecked value
 			}
+			if(this.checkboxChecked && this.checkboxUnchecked) {
+				// Both specified but neither found: indeterminate or false, depending
+				if(this.checkboxIndeterminate) {
+					return undefined;
+				} else {
+					return false;
+				}
+			}
 		}
 		if(this.checkboxListField || this.checkboxListIndex || this.checkboxFilter) {
 			// Same logic applies to lists and filters
@@ -122,7 +135,12 @@ CheckboxWidget.prototype.getValue = function() {
 				return true; // Absence of unchecked value
 			}
 			if(this.checkboxChecked && this.checkboxUnchecked) {
-				return false; // Both specified but neither found: default to false
+				// Both specified but neither found: indeterminate or false, depending
+				if(this.checkboxIndeterminate) {
+					return undefined;
+				} else {
+					return false;
+				}
 			}
 			// Neither specified, so empty list is false, non-empty is true
 			return !!list.length;
@@ -265,6 +283,7 @@ CheckboxWidget.prototype.execute = function() {
 	this.checkboxChecked = this.getAttribute("checked");
 	this.checkboxUnchecked = this.getAttribute("unchecked");
 	this.checkboxDefault = this.getAttribute("default");
+	this.checkboxIndeterminate = this.getAttribute("indeterminate");
 	this.checkboxClass = this.getAttribute("class","");
 	this.checkboxInvertTag = this.getAttribute("invertTag","");
 	this.isDisabled = this.getAttribute("disabled","no");
@@ -277,14 +296,19 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 CheckboxWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.tiddler || changedAttributes.tag || changedAttributes.invertTag || changedAttributes.field || changedAttributes.index || changedAttributes.listField || changedAttributes.filter || changedAttributes.checked || changedAttributes.unchecked || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.disabled) {
+	if(changedAttributes.tiddler || changedAttributes.tag || changedAttributes.invertTag || changedAttributes.field || changedAttributes.index || changedAttributes.listField || changedAttributes.listIndex || changedAttributes.filter || changedAttributes.checked || changedAttributes.unchecked || changedAttributes["default"] || changedAttributes.indeterminate || changedAttributes["class"] || changedAttributes.disabled) {
 		this.refreshSelf();
 		return true;
 	} else {
 		var refreshed = false;
 		if(changedTiddlers[this.checkboxTitle]) {
 			var isChecked = this.getValue();
-			this.inputDomNode.checked = isChecked;
+			this.inputDomNode.checked = !!isChecked;
+			if(isChecked === undefined) {
+				this.inputDomNode.indeterminate = true;
+			} else {
+				this.inputDomNode.indeterminate = false;
+			}
 			refreshed = true;
 			if(isChecked) {
 				$tw.utils.addClass(this.labelDomNode,"tc-checkbox-checked");
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -44,7 +44,7 @@ CheckboxWidget.prototype.render = function(parent,nextSibling) {
 		this.inputDomNode.setAttribute("checked","true");
 		$tw.utils.addClass(this.labelDomNode,"tc-checkbox-checked");
 	}
-	if(isChecked === undefined && this.checkboxIndeterminate) {
+	if(isChecked === undefined && this.checkboxIndeterminate === "yes") {
 		this.inputDomNode.indeterminate = true;
 	}
 	if(this.isDisabled === "yes") {
@@ -100,7 +100,7 @@ CheckboxWidget.prototype.getValue = function() {
 			}
 			if(this.checkboxChecked && this.checkboxUnchecked) {
 				// Both specified but neither found: indeterminate or false, depending
-				if(this.checkboxIndeterminate) {
+				if(this.checkboxIndeterminate === "yes") {
 					return undefined;
 				} else {
 					return false;
@@ -136,7 +136,7 @@ CheckboxWidget.prototype.getValue = function() {
 			}
 			if(this.checkboxChecked && this.checkboxUnchecked) {
 				// Both specified but neither found: indeterminate or false, depending
-				if(this.checkboxIndeterminate) {
+				if(this.checkboxIndeterminate === "yes") {
 					return undefined;
 				} else {
 					return false;
@@ -283,7 +283,7 @@ CheckboxWidget.prototype.execute = function() {
 	this.checkboxChecked = this.getAttribute("checked");
 	this.checkboxUnchecked = this.getAttribute("unchecked");
 	this.checkboxDefault = this.getAttribute("default");
-	this.checkboxIndeterminate = this.getAttribute("indeterminate");
+	this.checkboxIndeterminate = this.getAttribute("indeterminate","no");
 	this.checkboxClass = this.getAttribute("class","");
 	this.checkboxInvertTag = this.getAttribute("invertTag","");
 	this.isDisabled = this.getAttribute("disabled","no");
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -304,11 +304,7 @@ CheckboxWidget.prototype.refresh = function(changedTiddlers) {
 		if(changedTiddlers[this.checkboxTitle]) {
 			var isChecked = this.getValue();
 			this.inputDomNode.checked = !!isChecked;
-			if(isChecked === undefined) {
-				this.inputDomNode.indeterminate = true;
-			} else {
-				this.inputDomNode.indeterminate = false;
-			}
+			this.inputDomNode.indeterminate = (isChecked === undefined);
 			refreshed = true;
 			if(isChecked) {
 				$tw.utils.addClass(this.labelDomNode,"tc-checkbox-checked");
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -67,7 +67,7 @@ CheckboxWidget.prototype.getValue = function() {
 	var tiddler = this.wiki.getTiddler(this.checkboxTitle);
 	if(tiddler || this.checkboxFilter) {
 		if(this.checkboxTag) {
-			if(this.checkboxInvertTag) {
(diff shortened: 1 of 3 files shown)


rmunn:
I wonder if the indeterminate-checkboxes example is a little too intimidating as it stands? Perhaps I should reduce the number of HTML elements by turning it into this instead:

```html
<<wikitext-example-without-html """\define check-all(field-name:"items") <$action-listops $field="selected-$field-name$" $filter="[list[!!$field-name$]]" />
\define uncheck-all(field-name:"items") <$action-listops $field="selected-$field-name$" $filter="[[]]" />
<$checkbox filter="[list[!!selected-fruits]count[]]" checked={{{ [list[!!fruits]count[]] }}} unchecked="0" checkactions=<<check-all fruits>> uncheckactions=<<uncheck-all fruits>> indeterminate> fruits</$checkbox>
<ul style="list-style: none">
<$list variable="fruit" filter="[list[!!fruits]]">
<li><$checkbox listField="selected-fruits" checked=<<fruit>>> <<fruit>></$checkbox></li>
</$list>
</ul>
<$checkbox filter="[list[!!selected-vegetables]count[]]" checked={{{ [list[!!vegetables]count[]] }}} unchecked="0" checkactions=<<check-all vegetables>> uncheckactions=<<uncheck-all vegetables>> indeterminate> veggies</$checkbox>
<ul style="list-style: none">
<$list variable="veggie" filter="[list[!!vegetables]]">
<li><$checkbox listField="selected-vegetables" checked=<<veggie>>> <<veggie>></$checkbox></li>
</$list>
</ul>

<p>Selected veggies: {{!!selected-vegetables}}<br/>
Selected fruits: {{!!selected-fruits}}</p>""">>
```

For comparison's sake (and so you don't have to click through to the commits in the PR to find it), here's the current example:

```html
<<wikitext-example-without-html """\define check-all(field-name:"items") <$action-listops $field="selected-$field-name$" $filter="[list[!!$field-name$]]" />
\define uncheck-all(field-name:"items") <$action-listops $field="selected-$field-name$" $filter="[[]]" />
<ul style="list-style: none">
<li>
<$checkbox filter="[list[!!selected-fruits]count[]]" checked={{{ [list[!!fruits]count[]] }}} unchecked="0" checkactions=<<check-all fruits>> uncheckactions=<<uncheck-all fruits>> indeterminate> fruits</$checkbox>
<ul style="list-style: none">
<$list variable="fruit" filter="[list[!!fruits]]">
<li><$checkbox listField="selected-fruits" checked=<<fruit>>> <<fruit>></$checkbox></li>
</$list>
</ul>
</li>
<li>
<$checkbox filter="[list[!!selected-vegetables]count[]]" checked={{{ [list[!!vegetables]count[]] }}} unchecked="0" checkactions=<<check-all vegetables>> uncheckactions=<<uncheck-all vegetables>> indeterminate> veggies</$checkbox>
<ul style="list-style: none">
<$list variable="veggie" filter="[list[!!vegetables]]">
<li><$checkbox listField="selected-vegetables" checked=<<veggie>>> <<veggie>></$checkbox></li>
</$list>
</ul>
</li>
</ul>

<p>Selected veggies: {{!!selected-vegetables}}<br/>
Selected fruits: {{!!selected-fruits}}</p>""">>
```

The simplified example removes the outer layer of `<ul>` and `<li>` elements, which has the effect of reducing the length of the HTML code you have to read to make sense of it. The only visual effect is that the indentation of the category checkboxes is shifted left.

Thoughts? Which of these two examples, otherwise identical except for having two unordered lists in them, would you prefer to see?

P.S. Also, I see I should add a blank line between the two macro definitions and the rest of the example, to separate the macros visually and make it clearer where the example truly starts.

rmunn:
Decided to go ahead and push the simplified version of the example, so that it looks a little less intimidating. Now it's two macro definitions, two checkbox widgets at the upper level, and two list widgets that each contain `<li><$checkbox ...>...</$checkbox></li>`. Anyone who isn't afraid of writing a TW filter should be fine reading that example now.

core/modules/widgets/checkbox.js
```diff
@@ -277,14 +296,19 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 CheckboxWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.tiddler || changedAttributes.tag || changedAttributes.invertTag || changedAttributes.field || changedAttributes.index || changedAttributes.listField || changedAttributes.filter || changedAttributes.checked || changedAttributes.unchecked || changedAttributes["default"] || changedAttributes["class"] || changedAttributes.disabled) {
+	if(changedAttributes.tiddler || changedAttributes.tag || changedAttributes.invertTag || changedAttributes.field || changedAttributes.index || changedAttributes.listField || changedAttributes.listIndex || changedAttributes.filter || changedAttributes.checked || changedAttributes.unchecked || changedAttributes["default"] || changedAttributes.indeterminate || changedAttributes["class"] || changedAttributes.disabled) {
```
rmunn, line 299: BTW, this line contains a minor bugfix for #6561. I added the `listIndex` field in #6561, but forgot to include it in this list of changed attributes. So if a user were to make `listIndex` an attribute that comes from a text reference (for some use case I can't think of), then a change in the value referred to by that text reference would not cause the checkbox to re-render. Would be very rare to occur, but I noticed it while I was making this PR so I might as well fix it.

saqimtiaz: Are there any attributes that do not warrant a refresh? If not, we can use the following check:

`if($tw.utils.count(changedAttributes) > 0)`

core/modules/widgets/checkbox.js
```diff
@@ -277,14 +296,19 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
...
 		var refreshed = false;
 		if(changedTiddlers[this.checkboxTitle]) {
 			var isChecked = this.getValue();
-			this.inputDomNode.checked = isChecked;
+			this.inputDomNode.checked = !!isChecked;
+			if(isChecked === undefined) {
+				this.inputDomNode.indeterminate = true;
+			} else {
+				this.inputDomNode.indeterminate = false;
+			}
```
rmunn, line 311: Question for whoever reviews this: would you prefer for that five-line `if` block to be collapsed down into a single line, as follows?

```js
this.inputDomNode.indeterminate = (isChecked === undefined);
```

Personally I would find that easier to read, but someone people would find it harder to read.

pmario: We happily use constructions like  ` = !!isChecked;`, which is also cryptic for some programmers ... So I would be fine with the collapsed version, since it only contains 1 comparison. .. 

If the logic would be more complicated I'd prefer the long form. ... But it depends on @Jermolene ??

IMO if we collapse this one, we can also collapse some if's in the code above.

rmunn: I'm inclined to make one change per PR, i.e. if I'm going to change code that existed before the PR (like collapsing several of the `if` statements in the checkbox widget) then I think I'd prefer to do that in a separate PR. That way you avoid the situation where someone wants to merge change 1 from the PR but not change 2, and can't merge the PR as a whole because of that. So for now, I think I'll just collapse that one if-else block into a single line and leave the others alone until a later PR.

rmunn: Done in commit 3d04b8a.

rmunn:
I don't plan any further changes to this PR unless a reviewer requests them. So it's ready to be reviewed now, and hopefully approved and merged.

rmunn:
@Jermolene Would you mind taking a look at this PR? It's ready to merge unless you have changes you want to request.

Jermolene:
Hi @rmunn thanks for the nudge.

I think there is one minor issue that came to light when I studied the docs.

The value of the "indeterminate" attribute is read like this:

```
this.checkboxIndeterminate = this.getAttribute("indeterminate");
```

And then, later, it is treated as a boolean value.

This isn't consistent with how other boolean values are handled. We generally use "yes" and "no" (there are inconsistencies with some widgets, but those are the preferred values).

The core generally doesn't use attributes that don't have a value because they can't be controlled programmatically. Instead, where necessary we rely on the behaviour that omitting an attribute causes it to be treated as the default value.

rmunn:
@Jermolene wrote:
> The value of the "indeterminate" attribute is read like this:
> 
> ```
> this.checkboxIndeterminate = this.getAttribute("indeterminate");
> ```
> 
> And then, later, it is treated as a boolean value.
> 
> This isn't consistent with how other boolean values are handled. We generally use "yes" and "no" (there are inconsistencies with some widgets, but those are the preferred values).

The reason for this design is addressed in https://github.com/Jermolene/TiddlyWiki5/issues/6564#issue-1182811362, which I'll repeat here so you don't have to go through an extra click to see it:

@rmunn wrote:
> Note also that the value of the `indeterminate` attribute is **ignored** other than being checked for a non-empty string. That means that if you have `<$checkbox indeterminate checked="yes" unchecked="no">`, which is equivalent to `<$checkbox indeterminate="true" checked="yes" unchecked="no">`, then a value of `"maybe"` will set the checkbox to the indeterminate state. (As would any value other than `"yes"` or `"no"`). The reason for that is to make it easy to assign the result of a filter to the `indeterminate` attribute using `{{{ [filter-expression] }}}` syntax. There's already plenty of precedent in filters for interpreting an empty value as "false" and a non-empty value as "true", such as the behavior of the `else` operator. So I've decided that the `indeterminate` attribute on checkbox widgets should function the same way.

Now, I can change this to use "yes" and "no" as the values, which mean that if you want to set the value of `indeterminate` with a filter, `[filter-expression]` would become `[filter-expression]then[yes]else[no]`. As a programmer I wanted to follow the "empty filter means false, non-empty means true" convention, but perhaps non-programmers would rather write `<$checkbox indeterminate="yes" checked="...">` instead of `<$checkbox indeterminate checked="...">`, which comes naturally to me.

So if you want, I'll change the design to use "yes" and "no", and also accept "true" as meaning "yes" because TW's design means that attributes with no value are treated as containing the string "true". (And if so, I think I should also accept "false" as meaning "no"). That does leave unanswered the question of what should be done with `indeterminate="y"` or any other value that isn't "yes" or "no", but conformity with other attributes is important.

I just wanted you to know that it was a deliberate design choice, and not just a coding mistake, that had me getting the value and then treating it as a boolean. But if you want a different design, of course I'll change it.

Jermolene:
Thanks @rmunn 

> The reason for this design is addressed in https://github.com/Jermolene/TiddlyWiki5/issues/6564#issue-1182811362, which I'll repeat here so you don't have to go through an extra click to see it:

Thank you, I'd missed that discussion I'm afraid.

The fact that a standalone attribute without a value gets assigned the value "true" is an old bug. I think it would have been  better to have matched the behaviour of the DOM, and assign the empty string. Meanwhile, TW chose "yes"/"no" over "true"/"false" long ago in the belief that it is clearer for ordinary users.

The only widget that recognises the string "true" as "yes" is the edit-shortcut widget, which was an oversight.

> > Note also that the value of the `indeterminate` attribute is **ignored** other than being checked for a non-empty string. That means that if you have `<$checkbox indeterminate checked="yes" unchecked="no">`, which is equivalent to `<$checkbox indeterminate="true" checked="yes" unchecked="no">`, then a value of `"maybe"` will set the checkbox to the indeterminate state. (As would any value other than `"yes"` or `"no"`). The reason for that is to make it easy to assign the result of a filter to the `indeterminate` attribute using `{{{ [filter-expression] }}}` syntax. There's already plenty of precedent in filters for interpreting an empty value as "false" and a non-empty value as "true", such as the behavior of the `else` operator. So I've decided that the `indeterminate` attribute on checkbox widgets should function the same way.

I do understand the point about consistency with filters, but here I think we need to be more concerned with consistency with the existing widgets.

> Now, I can change this to use "yes" and "no" as the values, which mean that if you want to set the value of `indeterminate` with a filter, `[filter-expression]` would become `[filter-expression]then[yes]else[no]`.

That's intentional. I think the second form is much clearer and less ambiguous.

> As a programmer I wanted to follow the "empty filter means false, non-empty means true" convention, but perhaps non-programmers would rather write `<$checkbox indeterminate="yes" checked="...">` instead of `<$checkbox indeterminate checked="...">`, which comes naturally to me.

While a big chunk of TiddlyWiki users have a programming background, I'm most interested in the users who come to it without prior experience in programming. For those users, the shorthands and conventions that programmers take for granted can be a barrier to understanding. That perspective has influenced the design of TiddlyWiki a fair bit.

> So if you want, I'll change the design to use "yes" and "no", and also accept "true" as meaning "yes" because TW's design means that attributes with no value are treated as containing the string "true". (And if so, I think I should also accept "false" as meaning "no"). That does leave unanswered the question of what should be done with `indeterminate="y"` or any other value that isn't "yes" or "no", but conformity with other attributes is important.

As mentioned above, I'd prefer to stick to the simple "yes"/"no" formulation to keep the overall complexity of the system manageable.

> I just wanted you to know that it was a deliberate design choice, and not just a coding mistake, that had me getting the value and then treating it as a boolean. But if you want a different design, of course I'll change it.

Apologies, I had assumed the design was intentional, and it's entirely my fault for not catching the issue sooner.

rmunn:
@Jermolene wrote:
> As mentioned above, I'd prefer to stick to the simple "yes"/"no" formulation to keep the overall complexity of the system manageable.

Done in commit e12a8f5. In the same commit, I've also updated the text of the CheckboxWidget tiddler to mention "yes" and "no" as the valid values for `indeterminate`, and I've updated the unit tests as well.

P.S. In the process, I discovered one place where the `invertTag` attribute was being checked for Javascript truthiness instead of the value "yes", so I've gone ahead and fixed that in this PR as well. I haven't created a bug reproduction, but looking at the rest of the code it's clear that that one line was a mistake, which could have caused `invertTag="no"` to be treated as if it was `invertTag="yes"`. So now commit e0d62cf fixes that inconsistency with the `invertTag` attribute.

I don't have any more code I plan to add to this PR unless you request something else, so if you're happy with it, go ahead and merge.

Jermolene:
That's great, thank you @rmunn

PR #6597, opened 2022-04-05
Add static fallback content/links for tw.org

jeremyredhead:
Modified from the static.content template used in the tw5.com branch

I hope the filename is okay? (for that matter, perhaps next PR should be to organize the files into folders a little more)

I wasn't sure whether to transclude the TiddlyWiki or "Table of Contents" tiddlers, nor which order to put them in, so I omitted them. (It may be a good idea to include TiddlyWiki. Actually, it may be even better to transclude it into HelloThere in case someone new to TW stumbles onto tw.org instead of tw.com)


Jermolene:
Hi @jeremyredhead my apologies for missing this. Much appreciated thank you.

PR #6513, opened 2022-03-07
Update Mathematics Operators documentation tiddler

es-kha:
Section titles have been added to the documentation
The operators are now listed in alphabetical order and their description is included.

Rational: with current documentation it was not so easy to find the operator I was looking for.

PR #6598, opened 2022-04-05
Update Mathematics Operators documentation

es-kha:
Follow up of Pull Request #6513.

*    Text of second section updated based on @Jermolene comment.
*   "Mathematics Operators list" section now presented as a table (as in Filter Operators tiddler)


Jermolene:
Hi @es-kha thanks for your patience, I'm afraid I missed this.

PR #6599, opened 2022-04-06
Fix for broken style block wikitext

flibbles:
Style blocks broke between this release and the last (or maybe the one before that). Came up in my Uglify unit tests.

```
@@.myClass
<div>

Content
</div>
@@
```
This is broken. It no longer injects the class into the div element like it used to.

Reason: style block notation injects into parseTreeNode.attributes, but it was never made aware of parseTreeNode.orderedAttributes. This PR fixes that. Fixed for styles too.

@Jermolene: By the way, styleblocks are still kinda bad. They're using `addAttributeToParseTreeNode` instead of `addStyleToParseTreeNode` like you'd expect. Because of this, they blast away any style that already existed on the contained nodes. Also, because styleblocks look for "^@@(?:\\r?\\n)?" (where the newline is optional), they're impossible to nest. Is this deliberate or a bug?

### DIFF
--- a/core/modules/utils/parsetree.js
+++ b/core/modules/utils/parsetree.js
@@ -25,17 +25,27 @@ exports.getAttributeValueFromParseTreeNode = function(node,name,defaultValue) {
 };
 
 exports.addClassToParseTreeNode = function(node,classString) {
-	var classes = [];
+	var classes = [],
+		attribute;
 	node.attributes = node.attributes || {};
-	node.attributes["class"] = node.attributes["class"] || {type: "string", value: ""};
-	if(node.attributes["class"].type === "string") {
-		if(node.attributes["class"].value !== "") {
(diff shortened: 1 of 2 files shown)


flibbles:
I realize `^@@(?:\r?\n)?` makes the newline optional in case it's at the end of source, but other wikitext rules handle this with something like `^@@(?:\r?\n|$)`. This might be a backward compatibility issue if anyone actually wrote content after closing `@@` even though that's super weird. If you want to make the change though, I can put in the fix and test it.

Jermolene:
Thanks @flibbles, well spotted. As you note, the implementation of style blocks is gnarly to say the least (they were done very early). Because they weren't a good idea I'm not really too inclined to put more effort into fixing them (beyond this PR), but I'm open to other views.

PR #6600, opened 2022-04-06
Trim Saga: Control panel and Editor Toolbar

flibbles:
Shout out to the TiddlySpot tiddler macro for being completely broken, but also being completely unused, so no one noticed.


Jermolene:
Thanks @flibbles

PR #6602, opened 2022-04-06
Add "scrollContainerSelector" attribute to pageScroller

BurningTreeC:
This PR adds the `scrollContainerSelector` attribute to the `tm-scroll` message which makes the pageScroller query for that selector and use the found element as the scroll Container

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -86,12 +86,11 @@ Returns:
 		y: vertical scroll position in pixels
 	}
 */
-exports.getScrollPosition = function(srcWindow) {
-	var scrollWindow = srcWindow || window;
-	if("scrollX" in scrollWindow) {
-		return {x: scrollWindow.scrollX, y: scrollWindow.scrollY};
+exports.getScrollPosition = function(scrollContainer) {
+	if("scrollX" in scrollContainer) {
+		return {x: scrollContainer.scrollX, y: scrollContainer.scrollY};
 	} else {
-		return {x: scrollWindow.document.documentElement.scrollLeft, y: scrollWindow.document.documentElement.scrollTop};
+		return {x: scrollContainer.scrollLeft, y: scrollContainer.scrollTop};
 	}
 };
 
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -53,6 +53,9 @@ PageScroller.prototype.handleEvent = function(event) {
 		if($tw.utils.hop(event.paramObject,"animationDuration")) {
 			options.animationDuration = event.paramObject.animationDuration;
 		}
+		if($tw.utils.hop(event.paramObject,"scrollContainerSelector")) {
+			options.scrollContainerSelector = event.paramObject.scrollContainerSelector;
+		}
 		if(event.paramObject && event.paramObject.selector) {
 			this.scrollSelectorIntoView(null,event.paramObject.selector,null,options);
 		} else {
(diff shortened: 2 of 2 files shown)


saqimtiaz:
I think if we go this route we might want to also accommodate the use case of scrolling the container into view at the same time as we scroll within the container to scroll the target element into view.

BurningTreeC:
> I think if we go this route we might want to also accommodate the use case of scrolling the container into view at the same time as we scroll within the container to scroll the target element into view.

Hello @saqimtiaz - wouldn't that mean that we would be in an endless loop looking for the scrollContainer of the scrollContainer and so on? How do you intend to specify the scroll-container of the scroll-container?

Jermolene:
I think @saqimtiaz is right. Another way of putting it is that the page scroller should detect all ancestor scrollable containers, and adjust all of them to ensure that the target is in view.

In [TWPub](https://github.com/TWPUB/TWPUB-Tools), I actually use the native `element.scrollIntoView()` method:

https://github.com/TWPUB/TWPUB-Tools/blob/adbf6f34282e010f5259223dfbfd01068f82f99f/plugins/twpub-tools/%24/core/modules/utils/dom/scroller.js#L61

Here's a demo showing how it scrolls multiple nested scroll containers one after the other:

https://gist.github.com/Jermolene/3acf7240235d8d82d3a97c813b53cad0

We should move to using `.scrollIntoView()` in the core, and ideally have a polyfill that is fully compatible with the spec behaviour.

BurningTreeC:
Question 1:

How do we query for the scroll containers?

BurningTreeC:
Question 2:

How to specify the duration of the scroll?

saqimtiaz:
@BurningTreeC the native `element.scrollIntoView()` method takes care of scrolling any ancestor scroll containers as needed. Suggest testing the demo that Jeremy provided.

BurningTreeC:
> @BurningTreeC the native `element.scrollIntoView()` method takes care of scrolling any ancestor scroll containers as needed. Suggest testing the demo that Jeremy provided.

Thanks @saqimtiaz , I did so.
Why don't we just do it like @Jermolene did it in the twpub-tools then?

BurningTreeC:
@Jermolene @saqimtiaz 

what would be the route to choose if the `scrollBehavior` parameter isn't supported by the browser?
I found a solution that wraps the `element.scrollIntoView(scrollBehavior: "smooth")` into a `try { ... } catch(e) { ... }`
We could do that and fall back to the old behavior

Jermolene:
Hi @BurningTreeC sorry for the confusion. I think I'm actually talking about two slightly different potential resolutions:

1. That our current scroller be updated to handle scrolling intermediate containers
2. That we should switch to using the native `scrollIntoView()` method, but that we would need a polyfill for older browsers

It would only make sense to do one or the other of those, of course. The simplest would be (1), the purpose of pointing out (2) was partly so that you could look at the spec to see what it actually does in terms of intermediate scroll containers.

BurningTreeC:
Hello @Jermolene , no problem

to 1.
how should we detect intermediate containers?

to 2.
I understand that that's a different resolution and that there's not the full browser support

I would prefer resolution 1

Jermolene:
> to 1. how should we detect intermediate containers?

I'm not sure, that's where I hope the spec for `scrollIntoView()` might help.

> to 2. I understand that that's a different resolution and that there's not the full browser support
> 
> I would prefer resolution 1

Yes, I think that makes perfect sense for the moment. It is on my list for TWPub, but very happy if you want to have a look at it.

BurningTreeC:
The specs don't say anything about how it's done
My idea would be to query for all scrollable parents of the element that should be scrolled into view
Then we start with the outermost scroll-parent and scroll the next scroll-parent into view,
then the same with that scroll-parent, down to the element
The crucial point is to reliably detect a scrollable element

saqimtiaz:
Looking at the implementation details of a polyfill for `scrollIntoView()` would likely offer insight.

We may also want to consider supporting scrolling only if needed, as implemented in the proprietary `scrollIntoViewIfNeeded()` method on Chrome/Edge.

BurningTreeC:
> Looking at the implementation details of a polyfill for `scrollIntoView()` would likely offer insight.

Thanks @saqimtiaz , I haven't found a polyfill for `scrollIntoView()` but for the `scrollBehavior` I did. But it doesn't offer much insight.
> 
> We may also want to consider supporting scrolling only if needed, as implemented in the proprietary `scrollIntoViewIfNeeded()` method on Chrome/Edge.

Yes that's something we should care about

BurningTreeC:
I found [this article](https://matthiasott.com/notes/scrolling-elements-into-view) which states, that we can set the `scroll-behavior` by CSS to the scrollContainers like `scroll-behavior: smooth;` and do just `element.scrollIntoView()`

The upside of this is, that we can add a media-query for people that prefer `reduced motion` because of an illness or else (`@media (prefers-reduced-motion)`) and set the `scroll-behavior: auto;` in that case

CSS `scroll-behavior` only lacks support for IE

BurningTreeC:
Any thoughts someone? :smiley:

PR #6604, opened 2022-04-07
Trim Saga: Advanced Search and actions

flibbles:
A smaller chunk cause I took some liberties with quotation to reduce the overhead of the pragma I'm adding. Also, I just find "single quote inside double quote" a lot more consistent when all the rest of the neighboring quotes are doubles too.


Jermolene:
Thanks @flibbles

PR #6606, opened 2022-04-07
Add an optional delay to the tm-focus-selector action

BurningTreeC:
This PR makes the tm-focus-selector message optionally look for the specified element after a given delay in ms.
This is useful if we trigger some actions that show a div with an animation inside which the element is that should be focused

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -40,15 +40,25 @@ exports.startup = function() {
 	// Install the tm-focus-selector message
 	$tw.rootWidget.addEventListener("tm-focus-selector",function(event) {
 		var selector = event.param || "",
+		    	delay = event.paramObject.delay,
 			element,
 		    	doc = event.event && event.event.target ? event.event.target.ownerDocument : document;
-		try {
-			element = doc.querySelector(selector);
-		} catch(e) {
-			console.log("Error in selector: ",selector)
-		}
-		if(element && element.focus) {
-			element.focus(event.paramObject);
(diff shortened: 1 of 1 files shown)


BurningTreeC:
@Jermolene - do you have any thoughts about this?

Jermolene:
Hi @BurningTreeC there's a long tradition of using timers to workaround obscure bugs in web development, but it's not something I'd like to see in the core.

As discussed elsewhere, focus and selection should be managed as part of the refresh cycle but there is some further work needed before we have a viable approach for that.

BurningTreeC:
Thanks @Jermolene , yes I know it's a workaround.

But if we set a state Tiddler to yes and at the same time send a tm-focus-selector message to focus an element which is being animated into the viewport if that state Tiddler is yes, then there's no chance to do it other than setting a delay to the tm-focus-selector message. Or do you have an idea how to do it?

Jermolene:
> But if we set a state Tiddler to yes and at the same time send a tm-focus-selector message to focus an element which is being animated into the viewport if that state Tiddler is yes, then there's no chance to do it other than setting a delay to the tm-focus-selector message. Or do you have an idea how to do it?

There is the possibility that the value of that state tiddler gets overwritten before the refresh cycle runs.

I think we've discussed this before: the refresh cycle needs to extended to explicitly manage the focus (and indeed selections and scroll positions) as part of the DOM state. As I recall it, your previous PR addressed the preservation of focus but that we ran into questions when trying to programmatically change the focus.

PR #6611, opened 2022-04-08
Trim Saga: THE END

flibbles:
This is it. Unless a tiddler snuck in during the saga (which probably happened), all unnecessary whitespace is gone.

However, at this point, it's about maintaining proper trimming hygiene. Some tiddlers don't have "\whitespace trim" because they don't need them, but with even a slight modification, this would change. Take this:

```
<div>

<span>

Text

</span>

</div>

```
Seems like there's a lot of whitespace, but there is none. But take this:
```
<div>
<span>

Text

</span>

</div>

```
This has whitespace all over it, and needs a trim pragma. **I** know all the rules about when whitespace is and is not automatically trimmed because I wrote uglify. I probably know it better than anyone, and I know that the rules are fairly complicated.

My point is that I think even trying to maintain trim hygiene, with all these other developers, TiddlyWiki is going to backslide, so... have fun with that.

Anyway, at this point, it may be perfect. And most importantly, Uglify trims TiddlyWiki just great.

### DIFF
--- a/core/ui/AdvancedSearch.tid
+++ b/core/ui/AdvancedSearch.tid
@@ -2,6 +2,7 @@ title: $:/AdvancedSearch
 icon: $:/core/images/advanced-search-button
 color: #bbb
 
+\whitespace trim
 <div class="tc-advanced-search">
 <$macrocall $name="tabs" tabsList="[all[shadows+tiddlers]tag[$:/tags/AdvancedSearch]!has[draft.of]]" default="$:/core/ui/AdvancedSearch/System" actions="""<$action-setfield $tiddler="$:/state/advancedsearch/currentTab" text=<<currentTab>>/>""" explicitState="$:/state/tab--1498284803"/>
 </div>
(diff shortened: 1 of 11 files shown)


tw-FRed:
> I know all the rules about when whitespace is and is not automatically trimmed because I wrote uglify. I probably know it better than anyone, and I know that the rules are fairly complicated.
>
> My point is that I think even trying to maintain trim hygiene, with all these other developers, TiddlyWiki is going to backslide

Hi @flibbles 
Thanks for all the work you put in this series of PR!

I can't help but hope the precious knowledge you built through this work won't be lost. Could the "fairly complicated" rules be translated to CI tests in order to prevent TW code to backslide?

Jermolene:
Thanks @flibbles much appreciated.

flibbles:
We're almost done! Just this one still remains: https://github.com/Jermolene/TiddlyWiki5/pull/6604

PR #6613, opened 2022-04-09
Extend documentation for tm-close-window

saqimtiaz:
Adds documentation about the variable tv-window-id


Jermolene:
Thanks @saqimtiaz

PR #6616, opened 2022-04-10
Prevent browser storage plugin from skipping items when items are removed at startup

btheado:
While looping over all the browser storage items by index, the items
should not be removed because removing alters the index positions. The
item following the removed one will be skipped by the loop. The skipped
localstorage item will therefore not be loaded into the current tiddlywiki session.

Instead, accumulate a list of keys to remove and remove them after the
loop.

### DIFF
--- a/plugins/tiddlywiki/browser-storage/rawmarkup.js
+++ b/plugins/tiddlywiki/browser-storage/rawmarkup.js
@@ -26,6 +26,7 @@ if(Object.prototype.hasOwnProperty.call($tw.hooks.names,hookName)) {
 // Load tiddlers from browser storage
 function hookBootTiddlersLoaded() {
 	var url = window.location.pathname,
+		keysToDelete = [],
 		log = [];
 	// Check that browser storage is available
 	try {
@@ -55,7 +56,9 @@ function hookBootTiddlersLoaded() {
 							existingTiddler = $tw.wiki.getTiddler(title);
 						if(existingTiddler && existingTiddler.isEqual(incomingTiddler)) {
 							// If the incoming tiddler is the same as the existing then we can delete the local storage version
-							window.localStorage.removeItem(key);
+							// Defer deletion until after this loop, since deleting will shift the index and cause the
+							// index+1 item to be skipped.
+							keysToDelete.push(key);
 						} else {
 							$tw.wiki.addTiddler(incomingTiddler);
 							log.push(title);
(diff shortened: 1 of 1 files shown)


btheado:
Here is a complicated and maybe difficult to understand description on how to reproduce the issue. Looking at the code change is probably much simpler than trying to understand these steps. The key point here is that I duplicated the issue with these steps and after making the code change, I could no longer duplicate the issue:

1. Load TW from `editions/full` which has the browser storage plugin loaded and enabled
1. Navigate to the settings tab of the browser storage plugin
1. Click the "Clear browser storage" button and then click the "Use browser storage plugin" checkbox to put it in the "checked" state
1. In the "Browser Storage Save Filter" field, paste the filter: `$:/config/BrowserStorage/SaveFilter [regexp[^HelloThere$]]`
1. Make some change to the "HelloThere" tiddler.
1. From the browser console type `window.localStorage.key(0)` and `window.localStorage.key(1)`. The goal is to have the "HelloThere" tiddler be at an index (i.e. 0) and have another tiddler be at the following index (i.e. 1). I don't know how items are inserted into localstorage, but in my case the "HelloThere" tiddler was always at index 0 and the $:/config/BrowserStorage/SaveFilter tiddler was at index 1. If it doesn't happen that way for you, then you will need to add more tiddlers to the filter until you can get one to come after the "HelloThere" tiddler.
1. Restore the HelloThere tiddler to its exact, original contents. I did this by dragging the original HelloThere tiddler from a different, private/incognito window and import it over top of the changed tiddler.
1. At this point, the localstorage should contain the "SaveFilter" tiddler and an exact copy of the "HelloThere" tiddler. Refresh the web page.
1. Navigate to the settings tab of the browser storage plugin again. Look at the "Browser Storage Save Filter" box. Without the bug, it would contain the filter which was pasted in previous step, but due to the bug it will have reverted to the default filter (i.e. `[prefix[$:/state/]] -[prefix[$:/state/popup/]]`).
1. View the local storage via the browser developer tools and notice the "HelloThere" entry is gone (as expected) and the "SaveFilter" entry still has the value which we pasted in an earlier step (as expected, but not matching the state of the loaded tiddlywiki).
1. This illustrates the bug that localstorage items will be skipped when the prior item is removed at startup. Reloading the page again will restore the "SaveFilter" to the value from localstorage since this time no items are removed from localstorage during startup.

btheado:
The fix in this PR was also included in #6625, since the fix is a prerequisite for it.  That PR was merged first, so this one is obsolete.

Issue #6592, opened 2022-04-04
[BUG] pasting content from a webpage creates 2 elements in the "paste" event, TW dropzone overwrites the first one


pmario:
I did just implement a simple "rename Untitled" function and found an issue. The paste-event actually contains 2 elements. The first one is a `text/html` snippet and the second one is "text/plain" ... 

Since the default import handler in TW uses the "Untiltled" name for both elements the "text/plain" elements wins. ... That's what we expected. .. So the issue didn't come up. 

Now I do rename Untitled it looks as shown in the screenshot. ... The first element is an HTML snippet and the second one is text. 

I think, since we actually want the `text/plain` representation the "dropzone" paste-handler should throw the `text/html` representation away. ... BUT I'm not sure.

@Jermolene @saqimtiaz ... What do you think?

![image](https://user-images.githubusercontent.com/374655/161608034-d9093a52-a233-44bf-8e6e-8eb99c75fbfe.png)

saqimtiaz:
Do note that this isn't specific to pasting text, copying an image in the browser and pasting it also results in two elements being imported, the HTML fragment and the image. The dropzone in the texteditor doesn't exhibit this since it accepts only specific content types. The default global dropzone does exhibit this behaviour.

> I think, since we actually want the `text/plain` representation the "dropzone" paste-handler should throw the `text/html` representation away. ... BUT I'm not sure.

The tricky thing is, what happens if we are copying rendered HTML? Can we be sure that there is consistency amongst browsers and that we can always ignore `text/html` in the presence of `text/plain`, and that there is no scenario where the `text/html` might be useful??

pmario:
> .. and that there is no scenario where the text/html might be useful??

Oh, it is useful. We can convert HTML to wikitext. eg. If the fragment contains an H1 element, we could create a wikitext `! element` and so on. But other than that, I think it's junk, **if** there is a text/plain version.

Especially for the function I want to implement, where I would like to convert a `dot tid` like text into a real tiddler, instead of importing plain text.

pmario:
>.. The tricky thing is, what happens if we are copying rendered HTML?

I'll need to have a closer look, but I think it already works, since we didn't have a problem till now. The issue was always there, we just didn't see it.

pmario:
>The tricky thing is, what happens if we are copying rendered HTML?

As expected. There are 2 versions the HTML snippet contains HTML-encoded chars for `< and >` and the text/plain contains the text only. ... 

[...]

PR #6622, opened 2022-04-11
fix dropzone closure variable problem - ready to merge

pmario:
See "vercel" preview links below to find the PR-wikis, where you can test different browsers. At the moment it will create 2 tiddlers

The existing core code has an issue with the `type` variable assigned to the tiddler in [line 267](https://github.com/Jermolene/TiddlyWiki5/blob/73138b79aaaafa32b3a0abf8ba80c96eb3a18d56/core/modules/widgets/dropzone.js#L267) ... It is defined in [line 255](https://github.com/Jermolene/TiddlyWiki5/blob/73138b79aaaafa32b3a0abf8ba80c96eb3a18d56/core/modules/widgets/dropzone.js#L255), which will be defined by the last element in the for loop (line 244)

The `type` variable is passed to the `item.getAsString(function(str) {` **callback function as a closure variable, which is a problem in the existing code.**

The PR fixes that. So the `type` will be correctly assigned to every element, that comes from the clipboard, which in turn makes the behaviour for all desktop browsers I could test consistent. 

@Jermolene ... We need to decide, which element should be imported. ... From my point of view, it should be `text/plain`, since `text/html` creates a lot of overhead, that needs to be manually removed. 

As @saqimtiaz ... pointed out. Using `text/plain` as the default will be different to what Chrome does at the moment. 

It would be possible to check for the browser and "recreate the same behaviour" as it is now. .. But I personally think we should fix the problem in a consistent way.

-------

As shown at issue: **[BUG] pasting content from a webpage creates 2 elements in the "paste" event, TW dropzone overwrites the first one** #6592 

If any content from a webpage is copy / pasted into TW, we get 2 [dataTransfer.items](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/items) 

- One of them is `text/html` and the other one is `text/plain`
- This behaviour is consistent on desktop browsers
- **I did test** FireFox 99.0 Win 11, FF 99.0 Ubuntu 20.40, Edge Version 100.0.1185.36 (Offizielles Build) (64-Bit) Win 11, Edge Version 101.0.1210.10 (Official build) beta (64-bit) on Win 11, Chromium Version 100.0.4896.75 (Official Build) snap (64-bit) Ubuntu 20.04

I did try to test copy / paste on my phone, but I have no idea how to do that. I can't open 2 chrome windows in split screen which would allow drag & drop copy paste. 

FF for Android doesn't create  draggable content. 

Copying from other Android apps eg: Free-mail email-client seems to create `text/plain` elements only.

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -235,7 +235,29 @@ DropZoneWidget.prototype.handlePasteEvent  = function(event) {
 	var self = this,
 		readFileCallback = function(tiddlerFieldsArray) {
 			self.readFileCallback(tiddlerFieldsArray);
-		};
+		},
+		getItem = function(type) {
+			type = type || "text/plain";
+			return function(str) {
+				// Use the deserializer specified if any
+				if(self.dropzoneDeserializer) {
+					tiddlerFields = self.wiki.deserializeTiddlers(null,str,{title: self.wiki.generateNewTitle("Untitled " + type)},{deserializer:self.dropzoneDeserializer});
+					if(tiddlerFields && tiddlerFields.length) {
+						readFileCallback(tiddlerFields);
+					}
+				} else {
+					tiddlerFields = {
+						title: self.wiki.generateNewTitle("Untitled " + type),
+						text: str,
+						type: type
+					};
+					if($tw.log.IMPORT) {
+						console.log("Importing string '" + str + "', type: '" + type + "'");
+					}
+					readFileCallback([tiddlerFields]);
+				}
+			}
+		}
 	// Let the browser handle it if we're in a textarea or input box
 	if(["TEXTAREA","INPUT"].indexOf(event.target.tagName) == -1 && !event.target.isContentEditable) {
 		var self = this,
@@ -251,27 +273,10 @@ DropZoneWidget.prototype.handlePasteEvent  = function(event) {
 				});
 			} else if(item.kind === "string") {
 				// Create tiddlers from string items
-				var tiddlerFields,
-					type = item.type;
-				item.getAsString(function(str) {
-					// Use the deserializer specified if any
-					if(self.dropzoneDeserializer) {
-						tiddlerFields = self.wiki.deserializeTiddlers(null,str,{title: self.wiki.generateNewTitle("Untitled")},{deserializer:self.dropzoneDeserializer});
-						if(tiddlerFields && tiddlerFields.length) {
-							readFileCallback(tiddlerFields);
-						}
-					} else {
-						tiddlerFields = {
-							title: self.wiki.generateNewTitle("Untitled"),
-							text: str,
-							type: type
-						};
-						if($tw.log.IMPORT) {
-							console.log("Importing string '" + str + "', type: '" + type + "'");
-						}
-						readFileCallback([tiddlerFields]);
-					}
-				});
+				var tiddlerFields;
+				// It's important to give getAsString a closure with the right type
+				// So it can be added to the import queue
+				item.getAsString(getItem(item.type));
 			}
 		}
 		// Tell the browser that we've handled the paste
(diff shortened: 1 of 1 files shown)


pmario:
Test wiki: https://pmario.github.io/kitchensink/6622-paste-from-clipboard.html

pmario:
@Jermolene ... I know that this issue needs a bit of time to review, but it fixes a really nasty bug in the current code. It will allow us to highly improve the behaviour of copy paste actions in the future. 

I'll make this PR a draft -- but -- please review it. It's important for some of my plugins.

pmario:
@saqimtiaz ... It would be nice if you would have a look at it too

Jermolene:
Hi @pmario this seems OK. We have lots of limitations with drag/drop/import. Ultimately, we ned to improve things to give us a multistep import process that enables users to explicitly and separately select the data items they want to import, and the mechanism used to convert them into tiddlers. It makes sense to apply minor fixes in the meantime, but bigger fixes (eg those requiring a change in the UI) should be reviewed against the option of revamping things properly.

saqimtiaz:
> We need to decide, which element should be imported. ... From my point of view, it should be `text/plain`, since `text/html` creates a lot of overhead, that needs to be manually removed.

The problem is that either choice leads to a break in backwards compatibility for users of one browser, either Firefox or Chrome. From my perspective, the only acceptable option is to allow users greater insight and control over what data items they want to import, and with which deserializer.

pmario:
From the other issue: https://github.com/Jermolene/TiddlyWiki5/issues/7050

So we need some new UI handling that will help both parties. ... But I didn't find out, how to do it yet. ... I think it's getting more and more clear now

    Show both or "all 3" examples
    Deselect all of them by default
    Let the user decide, what to import ..
        They explicitely need to select those, they want to import.
        may be it's all 3 or even more of them

pmario:
>... Ultimately, we ned to improve things to give us a multistep import process that enables users to explicitly and separately select the data items they want to import, and the mechanism used to convert them into tiddlers. It makes sense to apply minor fixes in the meantime, but bigger fixes (eg those requiring a change in the UI) should be reviewed against the option of revamping things properly.

I think this is a proper fix as you describe it here. The user is able to preview the content that is new. ... 

image

---

The only problem is, that the preview selection should also be possible on a per preview basis. ... But that's a completely new PR.

pmario:
>Jeremy wrote
>
>... Ultimately, we ned to improve things to give us a multistep import process that enables users to explicitly and separately select the data items they want to import, and the mechanism used to convert them into tiddlers.

IMO that's related to:  #7208 where the conclusion is ... It needs to be plugins that extend the import functionality. 

I do use the existing core-upgrader mechanism to manipulate incoming tiddlers rename eg: image.png to `image-<timestamp>.png` and eg: Untitled to `Clipboard-<timestamp>` and so on. 

This mechanism will also allow us to implement the mechanism discussed at Talk-TW [“Discussion examples” converted into tiddlers?](https://talk.tiddlywiki.org/t/discussion-examples-converted-into-tiddlers/2843)

GameDungeon:
Is it possible to get that type passed to a hook before it gets reset?

Jermolene:
Thanks @pmario just to verify the current state: Does this PR change the default behaviour on any browser that you know of?

pmario:
>Thanks @pmario just to verify the current state: Does this PR change the default behaviour on any browser that you know of?

It creates 2 import elements instead of 1. 

- For chrome-based browsers it did create the "Untitled text/html" version by default ... as shown in the screenshot below as the first tiddler. 

- For FF it did create the "Untitled text/plain" version by default .. second tiddler in the screenshot. 

With this PR active it creates both tiddlers by deafult, so the user can decide. 

image

Jermolene:
> It creates 2 import elements instead of 1.
> 
> * For chrome-based browsers it did create the "Untitled text/html" version by default ... as shown in the screenshot below as the first tiddler.
> * For FF it did create the "Untitled text/plain" version by default .. second tiddler in the screenshot.
> 
> With this PR active it creates both tiddlers by deafult, so the user can decide.

Thanks @pmario I think that's reasonable.

PR #6623, opened 2022-04-11
[fr-FR] Translation of Saving and all saving methods

tw-FRed:
- Remove old tiddlers already deleted from EN edition
- Start using EN edition paths for updated/new translations

Most of the deleted files are just updated translations moved to their new path, except:
- Saving on InternetExplorer
- Saving on Safari

These were leftovers from deleted tiddlers in the EN edition.


Jermolene:
Great stuff thank you @tw-FRed

PR #6624, opened 2022-04-11
Give plugin authors the chance to extend a palette

BurningTreeC:
This PR changes the `colour` macro so, that it transcludes a tiddler with the name of the current palette but with the suffix `/extension` on the second position

This gives plugin authors the chance to extend palettes

### DIFF
--- a/core/wiki/macros/CSS.tid
+++ b/core/wiki/macros/CSS.tid
@@ -2,6 +2,7 @@ title: $:/core/macros/CSS
 tags: $:/tags/Macro
 
 \define colour(name)
+\whitespace trim
 <$transclude tiddler={{$:/palette}} index=<<__name__>>>
 <$transclude tiddler="$:/palettes/Vanilla" index=<<__name__>>>
 <$transclude tiddler={{{ [[$:/config/DefaultColourPaletteMappings/]addsuffix<__palette__>addsuffix[/]addsuffix<__name__>] }}}/>
--- a/core/wiki/macros/CSS.tid
+++ b/core/wiki/macros/CSS.tid
@@ -2,10 +2,10 @@ title: $:/core/macros/CSS
 tags: $:/tags/Macro
 
 \define colour(name)
-<$transclude tiddler={{$:/palette}} index="$name$">
-<$transclude tiddler="$:/palettes/Vanilla" index="$name$">
+<$transclude tiddler={{$:/palette}} index=<<__name__>>>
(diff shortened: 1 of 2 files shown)


BurningTreeC:
bad idea @Jermolene ?

Jermolene:
Hi @BurningTreeC I think this approach only allows one plugin to provide extensions for any particular palette. We should always be trying to minimise the chance of conflicts between plugins.

An approach that allows plugins to provide entries for particular palettes without conflicting with one another might be to support config tiddlers of the form `$:/config/DefaultColourPaletteMappings/<palette>/<name>`:

```
<$transclude tiddler={{$:/palette}} index="$name$">
<$transclude tiddler="$:/palettes/Vanilla" index="$name$">
<$transclude tiddler={{{ [[$:/config/DefaultColourPaletteMappings/]addsuffix<__palette__>addsuffix[/]addsuffix<__name__>] }}}/>
<$transclude tiddler="$:/config/DefaultColourMappings/$name$"/>
</$transclude>
</$transclude>
</$transclude>
```

The only conflicts with that design is if two plugins try to set the same entry for the same palette, but that's an intrinsic conflict that we can't resolve.

BurningTreeC:
Hi @Jermolene , I've changed it as you said
I'll add some docs

Jermolene:
Thanks @BurningTreeC I meant to add that we should also take this opportunity to replace the text substitution usage of this macro, please.

BurningTreeC:
Thanks @Jermolene , I've updated everything accordingly

core/wiki/macros/CSS.tid
```diff
@@ -2,7 +2,13 @@ title: $:/core/macros/CSS
 tags: $:/tags/Macro
 
 \define colour(name)
-<$transclude tiddler={{$:/palette}} index="$name$"><$transclude tiddler="$:/palettes/Vanilla" index="$name$"><$transclude tiddler="$:/config/DefaultColourMappings/$name$"/></$transclude></$transclude>
```
Jermolene, line 5: I don't think it would be backwards compatible to introduce all that whitespace into the macro. Better to use `\whitespace trim`

Jermolene (commented):
Thanks @BurningTreeC just a minor point.

Jermolene:
Thanks @BurningTreeC

Jermolene:
@BurningTreeC I have reverted this commit because it completely breaks colour handling:

image

Did you test it before submission?

BurningTreeC:
Apologies @Jermolene - I tested it but there's one small typo

PR #6625, opened 2022-04-12
Allow browser storage plugin to delete existing tiddlers

btheado:
Previously, the browser storage plugin could only delete tiddlers present in the localstorage (i.e. new tiddlers added by the end user). With this pull request, even tiddlers from the loaded page can be deleted.

A deleted tiddler is represented by a blank value.

At startup, the localstorage tiddlers with blank values will be deleted from the wiki if they are still present. If they are already gone from the wiki, then the blank localstorage entry will be deleted.

Also document a better save filter than `[all[]]`.

This PR builds upon (contains) the bug fix from PR #6616.


Jermolene:
Thanks @btheado that's a great idea.

Issue #6566, opened 2022-03-28
[IDEA] charcode macro to compliment operator


saqimtiaz:
As discussed in https://github.com/Jermolene/TiddlyWiki5/issues/5309#issuecomment-1080522418 it might be useful to have a macro version of the `charcode[]` operator.

This would allow usage like this for example to add two closing square brackets:
`:map[addsuffix<charcode 93>addsuffix<charcode 93>]`

As opposed to the operator version:
`:map[charcode[93]addprefix<currentTiddler>] :map[charcode[93]addprefix<currentTiddler>]`

pmario:
IMO related to: https://github.com/Jermolene/TiddlyWiki5/issues/6284 and https://github.com/Jermolene/TiddlyWiki5/pull/6295

Leilei332:
Since functions are supported in v5.3.0, we may not need to create a javascript macro:

```
\function char.code(num) [charcode<num>]

<<char.code "93">>
```

yaisog:
Hi @Leilei332, this is exactly what I'm doing, only a few characters shorter:
```
\function .chr(n) [charcode<n>]
```
which can then be used to e.g. wrap something in brackets:
```
[[Some text]addprefix<.chr 91>addsuffix<.chr 93>]
```
or as a function even,
```
\function .brackets() [all[]] :map[<currentTiddler>addprefix<.chr 91>addsuffix<.chr 93>]

<$text text={{{ [[Some text].brackets[]] }}} />
```
@pmario: We might think about closing this one.

Issue #6284, opened 2021-11-26
[IDEA] Introduce charcode macro


Jermolene:
As discussed [here](https://talk.tiddlywiki.org/t/more-direct-way-to-insert-and-in-filters/1670/16?u=jeremyruston), it is proposed to add a new `charcode` JavaScript macro designed to make it easier to introduce sequences of control characters into filters:

```
[get[field]addprefix<charcode 93>]
[<currentTiddler>addsuffix<charcode 13 10>]
[<currentTiddler>addsuffix<charcode cr lf>]
[<currentTiddler>addsuffix<charcode [[>]
```

The `charcode` macro would take a variable number of parameters, each being either:

*  a numeric Unicode codepoint,
* a shortcut name for a control character or
* a single typed character (this feature would not work properly characters that could be misinterpreted as a quotation symbol like `"`, `'`)

AnthonyMuscio:
Jeremy,

If I may just add, Control characters are often very different to the other cases as they may not easily be copy/pasted, or read when in text/wiki text etc... This functionality is important. Here are a few notes that may help clarify or scope what you are proposing. If you do not expect to merge or address these in the same macro feel free to request spawn another discussion.

Sorry for my verbosity, but its all "pent up", and hopefully useful, and relevant to charcode design considerations.

The named special characters are a great idea but it would be helpful if it were extensible, where the user/designer could add named character sequences eg rather than cr lf, I could name a paragraph "cr lf cr lf" "13 10 13 10". We could call these named character codes. below is an example and how I may hope this feature to assist with;

`{{{ [<currentTiddler>addsuffix<charcode paragraph>] }}}`

But this is also important importing and exporting content and the ability to "give a character"/returning a code" may also be helpful. 

After all the problem with Control Characters is their visibility and remembering them. The ability to use multiple numbers an names (lf) to build a string to use in the filter is valuable. Arguably if the solution is extensible then its size can be minimised ie no need to document al the names, if people can make them as needed.

However as implied by your example, you are also suggesting charcode can be also used to introduce anything but "whole of filter delimiters" to a filter as you demonstrated with "[[" , this is often as a parameter to a filter operator eg 
`"[<input-string>prefix<charcode [>then[Filter provided in input]"` but this is no longer really a char**code** but actual characters or strings. 

Finally beyond control characters, named or otherwise, including special characters or strings there is the question of access to Unicode characters.  
[...]

Issue #5309, opened 2020-12-23
[IDEA] In editor, drop image-link to create an external link


twMat:
I wish to easily add links to external images, like how it is done in the github editor, i.e:

Drag'n drop an image file link (from local or web) into the editor and its link appears at that specific spot - and in the correct syntax `[img[...]]` so the image is directly seen when the tiddler is stored.

This would tremendously simplify adding of images in TW.

Currently, imagelinks dropped into the editor from the web don't get the surrounding `[img[...]]` and dropping links from local into the editor instead opens the image in a new tab (win10, chrome). 

**Background**
As a science teacher (for past 3 years), I wish to create *worksheets* for my pupils with physics questions etc. The idea is to tag the questions (topic, phenomena, marks, difficulty, question type, etc) so I can easily generate worksheets, exam papers, experiments, equipment-lists, etc. One question or task = one tiddler. Here's a typical example of one question:
 
![image](https://user-images.githubusercontent.com/4534056/102821116-cad5be80-43d6-11eb-8d60-a1e90ba8af85.png)

As seen, questions typically have complex content; it's an arbitrary mix of text, images, tables, numbering of problem, and more. 

One may argue that "the TW way" would be to split up the content into several tiddlers. This is not at all desirable because the content is typically *not* reused elsewhere. In fact, tiddly-fiddling *must* be kept to a minimum: There will be several hundreds (thousands?) of these questions. It is unavoidable to have to *type the text* and, elsewhere, *edit the images*. But the next step, to add/link the images, is a *deal breaker* with the current image workflow in TW. I'm hoping it doesn't have to be :-)

saqimtiaz:
> Drag'n drop an image file link (from local or web) into the editor and its link appears at that specific spot - and in the correct syntax `[img[...]]` so the image is directly seen when the tiddler is stored.

What is an image file link?
Dropping an image file into the Github editor does indeed insert the image with the correct syntax. The TW 5.2.0 pre-release now has something similar.

Dropping a URL to an image file inserts just the URL as plain unformatted text into the editor in both Github and TW. This is the best that a plain text area based editor can do.

twMat:
>What is an image file link?

You seem to get what I mean with "image file link" - and I must say *WHOA!* it is super smooth to DnD an image from a local folder! That is exactly what I meant. Saq, do I have you to thank both for the heads up *and* for the implementation? :-)

[...]

PR #6626, opened 2022-04-13
add charcode, jsstirng macros. allow lf, tab, sapce shortcuts in charcode 

pmario:
- The PR adds some "lookup config" elements to allow us to use `lf tab space \n \t` which are easier to handle for humans as `10 13`
- It adds a `charcode` **macro**
- It adds a `jsstring` **macro** ... name is up for discussion
- It changes the existing `charcode` **filter** to allow us to use `lf tab \n \t` -- same reason as above

-------

[There are some test-xxx tiddlers in the DEMO to show the possibilities](https://pmario.github.io/kitchensink/6566-charcode-macro.html#test-jsstring-charcode-side-by-side:test-jsstring-charcode-side-by-side%20test-6284-jeremy%20test-jsstring-with-linebreaks%20test-jsstring%20test-join-with-charcode-macro). Those tests should be converted to doc tiddlers in a second step.

This PR should fix:  

- #6566 .. by Saq
   - issue was created because of: https://github.com/Jermolene/TiddlyWiki5/issues/5309#issuecomment-1080522418
- #6284 .. by Jeremy
- #4697 .. by me

It uses the concept from draft PR:  **charcode-jsstring-macros-charcode-operator** #6295 .. by me.  

There are also some discussions at Talk-TiddlyWiki especially stobot's reply: https://talk.tiddlywiki.org/t/more-direct-way-to-insert-and-in-filters/1670/15 ... that imo requests a more generic way to manipulate strings. ... That's what the `jsstring` macro is used for.

### DIFF
--- a/core/modules/config.js
+++ b/core/modules/config.js
@@ -38,4 +38,18 @@ exports.htmlBlockElements = "address,article,aside,audio,blockquote,canvas,dd,de
 
 exports.htmlUnsafeElements = "script".split(",");
 
+// Character code lookups used in charcode operator and macro
+exports.lookup = {
+	"tab": 9,
+	"\\t": 9,
+	"lf": 10,
+	"\\n": 10,
+	"ff": 12,
+	"\\f": 12,
+	"cr": 13,
+	"\\r": 13,
+	"\\_": 32,
+	"space": 32
+}
+
 })();
--- a/core/modules/macros/charcode.js
+++ b/core/modules/macros/charcode.js
@@ -1,0 +1,37 @@
+/*\
+title: $:/core/modules/macros/charcode.js
+type: application/javascript
+module-type: macro
+
+Macro to return characters using the character-codes as input.
+If the code isn't found the character is returned.
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.name = "charcode";
+
+exports.params = [
+	{name: "codes"},
+];
+
+exports.run = function(codes) {
+	var lookup = $tw.config.lookup,
+		tokens = codes.trim().split(/\s+/),
+		results = [];
+	tokens.map(function(name) {
+		var code = lookup[name.toLowerCase()];
+		if (code) {
+			results.push(String.fromCharCode(code));
+		} else {
+			results.push(String.fromCharCode($tw.utils.parseInt(name)));
+		}
+	})
+	return results.join("");
+}
+
+})();
--- a/core/modules/macros/jsstring.js
+++ b/core/modules/macros/jsstring.js
@@ -1,0 +1,28 @@
+/*\
+title: $:/core/modules/macros/jsstring.js
+type: application/javascript
+module-type: macro
+
+Macro to return a string. Replaces \r, \n, \t strings with their character codes.
+All existing CRLF's are removed, which allows us to make the wikitext more readable.
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+exports.name = "jsstring";
+
+exports.params = [
+	{name: "text"}
+];
+
+// TODO .. may be it should also use the config.js lookup table
+
+exports.run = function(text, mode) {
+		return text.replace(/\\n/g,"\n").replace(/\\r/g,"\r").replace(/\\t/g,"\t");
+}
+
+})();
--- a/editions/tw5.com/tiddlers/$__config_ViewToolbarButtons_Visibility_$__core_ui_Buttons_info.tid
+++ b/editions/tw5.com/tiddlers/$__config_ViewToolbarButtons_Visibility_$__core_ui_Buttons_info.tid
@@ -1,0 +1,6 @@
+created: 20220413102320003
+modified: 20220413102320003
+title: $:/config/ViewToolbarButtons/Visibility/$:/core/ui/Buttons/info
+type: text/vnd.tiddlywiki
+
+show
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -175,10 +175,12 @@ exports.pad = function(source,operator,options) {
 }
 
 exports.charcode = function(source,operator,options) {
-	var chars = [];
+	var lookup = $tw.config.lookup,
+		chars = [];
 	$tw.utils.each(operator.operands,function(operand) {
-		if(operand !== "") {
-			chars.push(String.fromCharCode($tw.utils.parseInt(operand)));
+		var code = lookup[operand.toLowerCase()] || operand;
+		if(code !== "") {
+			chars.push(String.fromCharCode($tw.utils.parseInt(code)));
 		}
 	});
 	return [chars.join("")];
--- a/editions/tw5.com/tiddlers/$__StoryList.tid
+++ b/editions/tw5.com/tiddlers/$__StoryList.tid
@@ -1,5 +1,3 @@
-created: 20211126104006194
-list: [[Page and tiddler layout customisation]] [[Creating new buttons for the ViewToolbar and page controls]] [[Structuring TiddlyWiki]] Tagging [[Introduction to Lists]] [[Icon Gallery]] [[How to widen tiddlers (aka storyriver)]] [[How to turn off camel case linking]] [[How to put the last modification date in a banner]] [[How to hide the author's and other fields with CSS]] [[How to export tiddlers]] [[How to Customize TiddlyDesktop]] [[Editing Tiddlers with Vim]] [[Concatenating text and variables using macro substitution]] [[Demonstration: keyboard-driven-input Macro]] HelloThere GettingStarted Community
-modified: 20211126111221917
+list: test-jsstring-with-linebreaks test-join-with-charcode-macro test-jsstring-charcode-side-by-side test-jsstring
 title: $:/StoryList
 type: text/vnd.tiddlywiki
--- a/editions/tw5.com/tiddlers/$__StoryList.tid
+++ b/editions/tw5.com/tiddlers/$__StoryList.tid
@@ -1,3 +1,3 @@
-list: test-jsstring-with-linebreaks test-join-with-charcode-macro test-jsstring-charcode-side-by-side test-jsstring
+list: test-6284-jeremy test-jsstring-with-linebreaks test-join-with-charcode-macro test-jsstring-charcode-side-by-side test-jsstring
 title: $:/StoryList
 type: text/vnd.tiddlywiki
--- a/editions/tw5.com/tiddlers/test-6284-jeremy.tid
+++ b/editions/tw5.com/tiddlers/test-6284-jeremy.tid
@@ -1,0 +1,26 @@
+created: 20220413104641645
+modified: 20220413105228845
+myField: test
+tags: 
+title: test-6284-jeremy
+type: text/vnd.tiddlywiki
+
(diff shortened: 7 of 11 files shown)


Jermolene:
Hi @pmario would it be possible to use the existing `jsonstringify[]` operator rather than add a new one? It does a few substitutions though.

pmario:
I think none of my test-xx examples that use `jsstring` can be solved using the `jsonstringify[]` operator. I'll have a closer look

pmario:
eg: 
`[[Title with \n asdf]] +[stringify[]]` -> creates a link `[[Title with \\n asdf]]`

`<<jsstring "<pre>[[Title with \n asdf]]<pre>">>`  allows you to create this: 

image

core/modules/config.js
```diff
@@ -38,4 +38,18 @@ exports.htmlBlockElements = "address,article,aside,audio,blockquote,canvas,dd,de
...
+exports.lookup = {
+	"tab": 9,
+	"\\t": 9,
+	"lf": 10,
+	"\\n": 10,
+	"ff": 12,
+	"\\f": 12,
+	"cr": 13,
+	"\\r": 13,
+	"\\_": 32,
```
Jermolene, line 51: Why replace `\_` with a space? It doesn't seem to be a standard replacement.

And why are we only converting some of the JS character codes? Why not support all of the [standard ones](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#escape_sequences)?

core/modules/macros/charcode.js
```diff
@@ -0,0 +1,37 @@
...
+
+exports.params = [
+	{name: "codes"},
+];
+
+exports.run = function(codes) {
+	var lookup = $tw.config.lookup,
+		tokens = codes.trim().split(/\s+/),
+		results = [];
+	tokens.map(function(name) {
```
Jermolene, line 26: I know that some of them have crept in, but we don't currently use `.map` in the core.

core/modules/macros/jsstring.js
```diff
@@ -0,0 +1,28 @@
...
+exports.name = "jsstring";
+
+exports.params = [
+	{name: "text"}
+];
+
+// TODO .. may be it should also use the config.js lookup table
+
+exports.run = function(text, mode) {
+		return text.replace(/\\n/g,"\n").replace(/\\r/g,"\r").replace(/\\t/g,"\t");
```
Jermolene, line 25: I still don't understand the purpose of this macro, can you elaborate on some use cases?

Jermolene (commented):
Thanks @pmario some comments.

core/modules/config.js
```diff
@@ -38,4 +38,18 @@ exports.htmlBlockElements = "address,article,aside,audio,blockquote,canvas,dd,de
 
 exports.htmlUnsafeElements = "script".split(",");
 
+// Character code lookups used in charcode operator and macro
+exports.lookup = {
```
saqimtiaz, line 42: Would recommend use of a more explicit name that makes it clear what this is for, for example charcodeLookup rather than just lookup.

core/modules/config.js
pmario:
```
\0 | null character (U+0000 NULL)
```

I think the null-char will be **problematic** if saved **in its binary form** in eg: .TID files and opened with other editors. It's not really defined, what they do with that character. In the best case they will filter it out and remove it. But eg: notepad++ [will do weird things](https://stackoverflow.com/questions/58689285/why-does-the-notepad-null-character-not-paste).
 
```
\' | single quote (U+0027 APOSTROPHE)
\" | double quote (U+0022 QUOTATION MARK)
\\ | backslash (U+005C REVERSE SOLIDUS)
```

I thought. there is no real need to implement them, our users already know, that they can cover single quotes inside double quotes and vice versa. eg. `" ' "` and `' " '` or `""" " """` which IMO is an equivalent to `"\""`

I don't have a problem to implement the escaped quotes and backslash, but then we probably will need to implement them globally for **all operators, string functions, macros and widgets**. If that's intended, it's far beyond the intention of this PR.

```
\v | vertical tab (U+000B LINE TABULATION)
\b | backspace (U+0008 BACKSPACE)
```

The vertical tab imo makes no sense. I can't think of any usecase other than controlling printer behaviour. But since printers are not really controlled that way anymore, I think it would create a completely unpredictable behaviour.

BACKSPACE is a control character that will remove the last character. It can be losslessly substituted by not creating that character at all. So I don't really see a usecase for it.

pmario: >Why replace \_ with a space? It doesn't seem to be a standard replacement.

I think it would be useful, because it's more visible than ` ` if used with string manipulation.

core/modules/macros/charcode.js
pmario: I'll change that

core/modules/config.js
pmario: good idea

pmario:
>We should aim for consistency between the charcode macro and filter operator as much as possible. If we are going to support lookup tables in the macro, we need to introduce the same in the filter operator, perhaps enabled by use of a suffix.

I did think about that too

core/modules/config.js
Jermolene: @pmario 

> The vertical tab imo makes no sense. I can't think of any usecase other than controlling printer behaviour. But since printers are not really controlled that way anymore, I think it would create a completely unpredictable behaviour.

I distrust the logic here. If I've learned anything it is that I can't predict how people use the software that I build. It seems really weird to pick and choose which character codes we support.

> > Why replace _ with a space? It doesn't seem to be a standard replacement.
> 
> I think it would be useful, because it's more visible than ` ` <- there is a space if used with string manipulation.

It just means that we end up with a strange mix of standard and non-standard conversions, which is all highly opinionated.

pmario:
>@Jermolene wrote at: https://github.com/Jermolene/TiddlyWiki5/pull/6626#discussion_r851255004
>
>I still don't understand the purpose of this macro, can you elaborate on some use cases?

As I wrote in the OP there are several open GitHub issues and discussions at talk-tiddlywiki that discuss similar topics concerning "string manipulation" with "special characters" like `lf, \n, tab, \t` and so on. 

This PR should solve **all of them** in a simple to use way. 

IMO charcode as it is at the moment is difficult to use by ordinary users because it's very abstract. Most users (including me) have problems to know if `<<charcode 10>>` or `<<charcode 13>>` is the right thing to use. Users demanded and you suggested to be able to use `<<charcode lf>>` instead. 

At: https://github.com/Jermolene/TiddlyWiki5/issues/6284 you suggested: 

```
[get[field]addprefix<charcode 93>]
[<currentTiddler>addsuffix<charcode 13 10>]
[<currentTiddler>addsuffix<charcode cr lf>]
[<currentTiddler>addsuffix<charcode [[>]
```

Where the 3rd example is part of this PR and can be used like this: `[<currentTiddler>addsuffix<charcode "cr lf">]` or `[<currentTiddler>addsuffix<charcode "\r \n">]`. The former is more convenient for users and the later is common knowledge for programmers. The input for the charcode macro is a "space separated list" of "named" character codes or their "js-string" representation. 

The 4th example `[<currentTiddler>addsuffix<charcode [[>]` is impossible with `charcode`.  IMO `[[` should be used as a "string" and shouldn't be converted. ... BUT this now causes a problem. If `[[` should be used as a string we can't use `lf` as a name anymore, because it could also be a valid string. ... 

That's exactly what the `<<jsstring "[[">>` macro does. In addition jsstring can be used like this: `[[test]addprefix<jsstring "\n[[">addsuffix<jsstring "]]">]]` ... It can combine strings with special chars. jsstring can only use `\n` but not `lf`, for the same reason as above.

I know, there is a bit of a bit of an overlap between the "jsstring" macro and the "charcode" macro, but they nicely complement each other. 

I hope this makes it clearer. I honestly think we need both of them even if the functions overlap.

pmario:
@Jermolene ... After my last post, I did forget to switch from draft to review-mode. ... Done now

editions/tw5.com/tiddlers/$__config_ViewToolbarButtons_Visibility_$__core_ui_Buttons_info.tid
```diff
@@ -0,0 +1,6 @@
+created: 20220413102320003
+modified: 20220413102320003
+title: $:/config/ViewToolbarButtons/Visibility/$:/core/ui/Buttons/info
```
Jermolene, line 3: Did you mean to add this config tiddler?

editions/tw5.com/tiddlers/test-6284-jeremy-quotes.tid
```diff
@@ -0,0 +1,26 @@
+created: 20220421161059682
+modified: 20220421161120342
+tags: 
+title: test-6284-jeremy-quotes
```
Jermolene, line 4: Again, is this tiddler intended to be included?

Jermolene (commented):
Hi @pmario it doesn't look like this is quite ready; I'm seeing a number of TODO comments, too.

saqimtiaz:
@pmario a few areas of concern:
* are the changes in the charcode operator backwards compatible? Are any changes documented?
* we need documentation for the new macros and any changes to existing operators. Unless I am missing something I cannot see any documentation in this PR.

editions/tw5.com/tiddlers/$__config_ViewToolbarButtons_Visibility_$__core_ui_Buttons_info.tid
pmario: No will remove it

editions/tw5.com/tiddlers/test-6284-jeremy-quotes.tid
pmario: will be removed

pmario:
>- are the changes in the charcode operator backwards compatible? Are any changes documented?

Yes, I did only add new functionality. The old behaviour is still there. 

>- we need documentation for the new macros and any changes to existing operators. Unless I am missing something I cannot see any documentation in this PR.

The PR atm contains several text-xxx tiddlers, which I will convert to "operator xx (Example)" tiddlers. 

If the name for the new `jsstring` operator is OK I'll create a docs tiddler for it also using the existing text-xx tiddlers as examples. 

@Jermolene I'll convert this PR to a draft, till the new name `jsstring` is confirmed.

pmario:
>If the main motivation for the jsstring operator is to make it easier to include square brackets in filters, then I don't think it's really an improvement over the existing technique (ie \define brackets() [[).

That's only 1 usecase that was discussed at talk.tiddlywiki. .. I personally would use it to be able to create formatted plain text using tabs `\t` and linefeeds `\n` in combination with text like `<jsstring "\tsome text\n\n>` which is much more convenient than the following code. Especially if jsstring is only 1 element of a much more complex filter.

```
<$let tab="	" lf="
">
{{{ [[some text]addprefix<tab>addsuffix<lf>addsuffix<lf>] }}}
```

Jermolene:
> That's only 1 usecase that was discussed at talk.tiddlywiki. .. I personally would use it to be able to create formatted plain text using tabs `\t` and linefeeds `\n` in combination with text like `<jsstring "\tsome text\n\n>` which is much more convenient than the following code. Especially if jsstring is only 1 element of a much more complex filter.

That makes sense, but in that case I think that it needs to be considered as part of #6663, and should be removed from this PR.

Issue #6618, opened 2022-04-10
[BUG] Field Editor: Switching field name also clears the field value


AnthonyMuscio:
Unlike previous versions of tiddlywiki, in tiddler edit mode, if you paste a value into a new field dialogue then enter/select the fieldname, the value field is cleared and you have to provide it again. 

**To Reproduce**
1. Go to TiddlyWiki.com
2. Click on + New tiddler
3. Scroll down to "add new field", enter a value in the "field value"
4. Now enter or select a "field name".
5. Notice how the content of "field value" is erased (as soon as you type or select a field from the dropdown)

**Expected behavior**
The field value entered is retained until the user hits ''add'' or clears the content.

**Screenshots**
Not illustrative, please follow above instructions.

**TiddlyWiki Configuration **
 - Version v5.2.2
 
**Desktop (please complete the following information):**
 - Windows 11
 - Browser Chrome, Firefox and Edge
 - Latest general releases

**Smartphone (please complete the following information):**
 - expect same on all platforms

**Additional context**
I would expect this error should be fixed in the next release, however I would appreciate a workaround in the meantime. A workaround may forestall the "rushing to another release". Perhaps tiddlywiki.com should have  "errata" messages for such circumstances?

pmario:
I can confirm, that this problem wasn't there in v5.2.1 .... It's probably part of the keyboard handling for dropdowns

pmario:
@AnthonyMuscio .. It would be nice, if you add a proper title to your issue

pmario:
This is related to: https://github.com/Jermolene/TiddlyWiki5/pull/6511#issuecomment-1080693028

Jermolene:
Thanks @AnthonyMuscio, well spotted, and a very clear bug report.

It does indeed appear to be related to #6511. @FlashSystems could you comment?

pmario:
I think it's caused by the `new-field-actions()` ... 

But I think the underlaying problem is, that the keyboard handling uses 2 different temp-values to manage the "value" field. .. 
This causes "syncing problems" ... 

So **I think we need to fix the main problem**, otherwise we will only start to "fight" the symptoms all over the places. 

I needed to create a "hacky" workaround at: https://github.com/Jermolene/TiddlyWiki5/pull/6534/files ... which imo is caused by the same issue: 2 variables, that do 1 thing

pmario:
[...]

PR #6628, opened 2022-04-13
Fix for Bug #6618

FlashSystems:
As explained in bug #6618 switching fields lost the current value for the field. Fixing this could be done by using a single tiddler as the backing store for all values. But this can lead to the following surprising behaviour:

* The user types "not-a-date" into the field value of a simple text field.
* The user now selects a field name that uses a HTML5 date editor. The
  Editor will show no date because the value cannot be parsed.
* The user saves the tiddler by clicking the checkmark.

Now the date-field contains the value "not-a-date" but the user was not
aware that this will be added. The edit control showed no date (because
the value was invalid) and the user assumed the field value would be empty.

These commits fix the problem by using a different backing store for each type of field editor. The tiddlers for this are created as temporary tiddlers, and their name is derived from the name returned by the Field Editor Cascade. That way, each editor tiddler gets its own backing store.

If only the default tiddler is used. The behaviour exactly replicates that of TW 5.2.1. Even the quirk of the colour field (it's a text field until it is added) will be there again. Only if you add more field edit tiddlers to the cascade, the behaviour changes.

I've updated the documentation to clarify the details of how this works. You can try it out on the [Field Editor Cascade Demo](https://flashsystems.github.io/FieldTemplatesDemo/fieldeditorcascade.html).

### DIFF
--- a/core/ui/EditTemplate.tid
+++ b/core/ui/EditTemplate.tid
@@ -1,18 +1,21 @@
 title: $:/core/ui/EditTemplate
 
-\define delete-edittemplate-state-tiddlers() <$action-deletetiddler $filter="[<newFieldNameTiddler>] [<newFieldValueTiddler>] [<newFieldNameInputTiddler>] [<newFieldNameSelectionTiddler>] [<newTagNameTiddler>] [<newTagNameInputTiddler>] [<newTagNameSelectionTiddler>] [<typeInputTiddler>] [<typeSelectionTiddler>]"/>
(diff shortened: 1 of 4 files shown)


FlashSystems:
I've tested this solution for the last three weeks and it works quite well. If you do not use the Field Editor Cascade I found no difference to TW 5.2.1.

There is one minor issue, but I think that can not be solved at the moment: If you type a field name that changes the transcluded field editor and press tab fast you can switch to the old input element before the update happens. Then the focus of the input element is lost.

@Jermolene: I think this fix is ready to go. Could you please review this PR?

Jermolene:
Thank you @FlashSystems, apologies for the delay.

PR #6633, opened 2022-04-15
Droppable widget: dragenteractions and dragleaveations

BurningTreeC:
Hi all,
and especially @saqimtiaz 

I've prepared this PR which incorporates the modifications Saq made for the dropzone widget
I'd like to hear what Saq says to the implementation

### DIFF
--- a/core/modules/widgets/droppable.js
+++ b/core/modules/widgets/droppable.js
@@ -76,7 +76,7 @@ DroppableWidget.prototype.handleEvent = function(event) {
 	}
 };
 
-DroppableWidget.prototype.resetState = function(options) {
+DroppableWidget.prototype.resetState = function(options,event) {
 	options = options || {};
 	if(this.domNodes[0]) {
 		$tw.utils.removeClass(this.domNodes[0],"tc-dragover");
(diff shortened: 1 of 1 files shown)


saqimtiaz:
Hi @BurningTreeC, nice to see that you decided to take a crack at this. Unfortunately I am on leave until the end of the month and will not have the opportunity to properly review this before then.

 My recommendation is to try and create a test case that exhibits the original bug in Firefox, as signified by the persistent dragover state, and confirming that these changes resolves the issue. You will also need to test on other browsers to ensure that the behaviour there is unchanged.

BurningTreeC:
> Hi @BurningTreeC, nice to see that you decided to take a crack at this. Unfortunately I am on leave until the end of the month and will not have the opportunity to properly review this before then.
> 
> My recommendation is to try and create a test case that exhibits the original bug in Firefox, as signified by the persistent dragover state, and confirming that these changes resolves the issue. You will also need to test on other browsers to ensure that the behaviour there is unchanged.

Thanks @saqimtiaz , what would be a test case?
Would it be sufficient to use the "Open" tab without `\whitespace trim`?

BurningTreeC:
@Jermolene , may I prepare a PR for the dragenteractions only and have a look at this at a later point?

Jermolene:
> @Jermolene , may I prepare a PR for the dragenteractions only and have a look at this at a later point?

I think we should try to deal with them both at the same time.

pmario:
@BurningTreeC ... Is there a link to the description of the "problem" this PR should solve?

BurningTreeC:
Hi all,

I'd like to revive this PR

It does the following:

- adds dragenteractions
- adds dragleaveactions
- handles dragEnter and dragLeave like in dropzone.js

What do you think?

saqimtiaz:
Note also that while excessive dragenter events firing in edge cases are not a problem for the dropzone and were not handled by the fix I introduced there, they would be problematic here as they would lead to the extra invocation of actions.

PR #6635, opened 2022-04-16
Add renderTreeFootprint option to qualify

BurningTreeC:
The Problem:

if we have a tiddler opened twice in two separate story rivers the more-tiddler-actions dropdown, the tiddlerInfoState and the tiddler-folded-state are the same for the two tiddlers

The solution:

Using the `renderTreeFootprint` of the widget for those states

<hr>

What do you think?


saqimtiaz:
@BurningTreeC please develop the habit of having PRs that are actively being worked on in draft mode. Otherwise every change you make spams notifications and ultimately leads to your PRs being overlooked when they are finally ready for review.

saqimtiaz:
A PR should be **created** in draft mode to avoid the notification spam, just converting it into a draft is not sufficient.

BurningTreeC:
> A PR should be **created** in draft mode to avoid the notification spam, just converting it into a draft is not sufficient.

My sincere apologies @saqimtiaz

BurningTreeC:
This PR is now ready for review

Jermolene:
Hi @BurningTreeC 

> if we have a tiddler opened twice in two separate story rivers the more-tiddler-actions dropdown, the tiddlerInfoState and the tiddler-folded-state are the same for the two tiddlers

One problem is that the core view template explicitly uses an unqualified state tiddler title for the folded state: `$:/state/folded/$(currentTiddler)$`. If we want to have a separate folded state for each story river then we should use a qualified state tiddler title.

In the case of the tiddlerInfoState, the view template does already use a qualified state tiddler title. So therefore if you're finding that two story rivers are yielding the same state tiddler titles then you'd need to wrap each of the two story rivers in a transclusion (alternately, define the variable "transclusion" differently for each story river).

BurningTreeC:
Thanks for clarification @Jermolene ,
I always forget about this mechanism with the `transclusion` variable :'(

PR #6636, opened 2022-04-16
Fix: new field no longer got focus after adding a field

BurningTreeC:
When hitting enter in a new-field value the field gets added and it's supposed to focus the field-name input of the new field
This PR restores that behavior


FlashSystems:
Hello @BurningTreeC, thank you for correcting this issue, but it was already addressed in Commit [aed7c58](https://github.com/Jermolene/TiddlyWiki5/pull/6628/commits/aed7c58636d1e9ab16f8cf8af7326adf7410c068) of PR #6628 together with another problem created by PR #6511.

BurningTreeC:
Thanks @FlashSystems

PR #6638, opened 2022-04-16
Make folded-state variable qualified

BurningTreeC:
This PR makes the `folded-state` variable a `qualified` one through `<<qualify "$:/state/folded">>`

### DIFF
--- a/core/ui/ViewTemplate.tid
+++ b/core/ui/ViewTemplate.tid
@@ -1,12 +1,9 @@
 title: $:/core/ui/ViewTemplate
 
 \whitespace trim
-\define folded-state()
-$:/state/folded/$(currentTiddler)$
-\end
 \define cancel-delete-tiddler-actions(message) <$action-sendmessage $message="tm-$message$-tiddler"/>
 \import [all[shadows+tiddlers]tag[$:/tags/Macro/View]!has[draft.of]]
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @BurningTreeC perhaps there are backwards compatibility concerns here:

1. Users may have set up their wikis with certain tiddlers folded and wish to retain that state over an upgrade
2. Users may have custom code that folds/unfolds tiddlers

It may still be worth doing despite the incompatibility, but it is important that we don't break compatibility unknowingly.

Another consideration is that perhaps these state tiddlers should be based off the identity of the story river, not its location in the widget tree. Otherwise moving a story river (eg to rearrange columns) will change the names of all the state tiddlers.

pmario:
I'm not sure, why this is different to the existing behaviour. Tiddler titles need to be unique. So there isn't any difference. It's just not human readable anymore. I think, that's a regression.

saqimtiaz:
> 2\. Users may have custom code that folds/unfolds tiddlers

This is highly likely to be the case. The problem with qualified state tiddlers for folding is that they cannot be manipulated from action strings invoked elsewhere.

BurningTreeC:
Yes @pmario @saqimtiaz - I've now understood the point. I'm going to close this PR

PR #6639, opened 2022-04-16
Give plugin authors the chance to extend a palette (II)

BurningTreeC:
This is the working version, apologies for my previous mistake


BurningTreeC:
Hi @Jermolene , this would be ready to merge (and tested)

BurningTreeC:
@Jermolene , I think we should overthink the order in which colors get transcluded.
I would say it would make sense to get the new color-definitions first, then the DefaultColorMappings, then the colors from the palette and then the fallback Vanilla colors

PR #6641, opened 2022-04-17
add TiddlyMemo edition

oflg:
TiddlyMemo focuses on learning and memorization


Jermolene:
Thanks @oflg - now that we have the links aggregator I have not been making updates to the community links on tiddlywiki.com. I've no objection to accepting updates, but at some point we may remove the remaining community links from tiddlywiki.com.

However, updates for tiddlywiki.com should target the `tiddlywiki-com` branch so that CI will update the site immediately, without waiting for the next release.

oflg:
Thanks for pointing out my mistake, and I'll correct it.

PR #6643, opened 2022-04-17
Scrollable widget: don't trap tm-scroll message if a selector is given

BurningTreeC:
This PR removes the ability of the `<$scrollable>` widget to scroll a selector into view
Instead, the `tm-scroll` message isn't trapped anymore

This arises from the discussion at #6631

### DIFF
--- a/core/modules/widgets/scrollable.js
+++ b/core/modules/widgets/scrollable.js
@@ -34,6 +34,9 @@ ScrollableWidget.prototype.cancelScroll = function() {
 Handle a scroll event
 */
 ScrollableWidget.prototype.handleScrollEvent = function(event) {
+	if(event.paramObject && event.paramObject.selector) {
+		return true;
+	}
 	// Pass the scroll event through if our offsetsize is larger than our scrollsize
 	if(this.outerDomNode.scrollWidth <= this.outerDomNode.offsetWidth && this.outerDomNode.scrollHeight <= this.outerDomNode.offsetHeight && this.fallthrough === "yes") {
 		return true;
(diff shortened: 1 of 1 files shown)


saqimtiaz:
Have you tested and confirmed that this is backwards compatible?

If a tm-scroll message arises inside a scrollable widget and the selector targets a DOM element inside the scrollable, will the scrolling still happen inside the scrollable and not at the page scroller level?

Such a change may need to happen in combination with the proposed changes to the page scroller.

BurningTreeC:
Hi @saqimtiaz , I have tested this and in terms of backwards compatibility with the core I don't have any concerns
In terms of backwards compatibility with existing plugins I really don't know what to say

saqimtiaz:
Backwards compatibility applies not only to the core but to all wikitext usage patterns that users may realistically be applying. 

This change isn't backwards compatible with several of my wikis.

BurningTreeC:
>If a tm-scroll message arises inside a scrollable widget and the selector targets a DOM element inside the scrollable, will the scrolling still happen inside the scrollable and not at the page scroller level?

Yes @saqimtiaz - I don't have to test that to know that that will not happen anymore

What do you think about an additional parameter for the tm-scroll message that - if set to yes - makes scrolling happen on the pageScroller level?

saqimtiaz:
> I don't have to test that to know that that will not happen anymore

It is important to consider any such possibilities when making a PR and highlight any potential changes to the way that wikitext usage of widgets and messages will work. Otherwise we risk unknowingly breaking backwards compatibility and having regressions in the core.



> What do you think about an additional parameter for the tm-scroll message that - if set to yes - makes scrolling happen on the pageScroller level?

Some sort of passthrough parameter is indeed a possibility. I think we need to consider this holistically along with the changes to scrolling proposed in #6602. Changes to how we handle scrolling could in fact make a case for deprecating the scrollable widget.

PR #6644, opened 2022-04-17
Fix lazy all template with user defined macro cause error

### LINKED ISSUE #6637: [BUG] lazy-all with user defined macro cause TypeError: module is undefined
**Describe the bug**

截屏2022-04-16 19 18 07

when using root-tiddler=$:/core/save/lazy-all

**To Reproduce**
1. clone a nodejs wiki with some plugins https://github.com/tiddly-gittly/Tiddlywiki-NodeJS-Github-Template
2. run `node ./tiddlywiki.js +plugins/tiddlywiki/filesystem +plugins/tiddlywiki/tiddlyweb /xxx/wiki --listen root-tiddler=$:/core/save/lazy-all`
3. open http://127.0.0.1:8080

**Expected behavior**

It should load the frame, and lazy load tiddlers.

 - Version 5.2.2
 - Saving mechanism Node.js
 - Plugins installed: lots of
 
 - OS: Mac
 - Browser Firefox
 - Version 100

**Additional context**

I'm trying to speed up the loading...

linonetwo:
Fixes #6637

### DIFF
--- a/core/templates/save-lazy-all.tid
+++ b/core/templates/save-lazy-all.tid
@@ -1,7 +1,7 @@
 title: $:/core/save/lazy-all
 
 \define saveTiddlerFilter()
-[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] 
+[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] [is[tiddler]type[application/javascript]] +[sort[title]]
 \end
 \define skinnySaveTiddlerFilter()
 [!is[system]] -[type[application/javascript]]
--- a/core/templates/save-lazy-all.tid
+++ b/core/templates/save-lazy-all.tid
@@ -1,9 +1,9 @@
 title: $:/core/save/lazy-all
 
 \define saveTiddlerFilter()
-[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] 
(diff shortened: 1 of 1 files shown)


core/templates/save-lazy-all.tid
```diff
@@ -1,9 +1,9 @@
 title: $:/core/save/lazy-all
 
 \define saveTiddlerFilter()
-[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] 
+[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]]
```
Jermolene, line 4: The construction `-[type[application/javascript]library[yes]]` is used in the core as a rather clumsy way to exclude the SJCL library when saving. The same construction is in the usual `$:/core/save/all` filter too.

It's possible that we should review unintended side effects of that behaviour, but here we should leave it alone. I think the correct value for this filter should be:

```
[is[system]] -[prefix[$:/state/popup/]] -[[$:/HistoryList]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[is[tiddler]type[application/javascript]] +[sort[title]]
```

And the skinny filter should be as you have it:

```
[!is[system]] -[type[application/javascript]]
```

Jermolene (commented):
Thanks @linonetwo

core/templates/save-lazy-all.tid
linonetwo: Updated and tested locally.

And an interesting thing is, even you may misuse `+[[]]` to try to add a set to filter result, while the correct syntax is just `[[]]`, lol.

I always think using `+` as "pipe" operator is not a very good idea...maybe `=` is better, but at this point, we might not have a chance to change this.

Jermolene: > And an interesting thing is, even you may misuse +[[]] to try to add a set to filter result, while the correct syntax is just [[]], lol.

You're quite right, that's a typo, sorry!

> I always think using + as "pipe" operator is not a very good idea...maybe = is better, but at this point, we might not have a chance to change this.

Yes, I'm not sure that + was the best choice.

linonetwo:
@Jermolene ready to merge

Jermolene:
Thanks @linonetwo

PR #6646, opened 2022-04-17
Add "passThrough" parameter to scrollable.js

BurningTreeC:


core/modules/widgets/scrollable.js
```diff
@@ -34,6 +34,10 @@ ScrollableWidget.prototype.cancelScroll = function() {
 Handle a scroll event
 */
 ScrollableWidget.prototype.handleScrollEvent = function(event) {
+	// Pass the scroll event through if explicitly wanted
+	if(event.paramObject && event.paramObject.passThrough === "yes") {
```
Jermolene, line 38: We should check for the existing `selector` parameter

PR #6655, opened 2022-04-24
Update CupertinoDark.tid so that editor-bg-color isn't transparent

BurningTreeC:
This solves a problem if we apply `color-scheme: dark` to the root element by CSS, where the editor bg stays white and no text is visible


Jermolene:
Thanks @BurningTreeC

PR #6656, opened 2022-04-24
Write the right "color-scheme" meta tag to the iframe of the framed text-editor

BurningTreeC:
This PR writes the `color-scheme` **meta-tag** to the framed text-editor, corresponding to the `color-scheme` field of the currently selected palette.

This then enables us to set the following in CSS:

```
:root {
color-scheme: dark/light (corresponding to palette);
}
```


Jermolene:
Thanks @BurningTreeC

PR #6664, opened 2022-04-26
Added a reference to encrypted wiki

Telumire:
A user was wondering about how to add a background image to the login page of an encrypted wiki (https://talk.tiddlywiki.org/t/background-image-on-login-page/3145). It was not obvious that an encrypted wiki is considered as "loading", this PR update the tiddler "Creating a splash screen" to hint at this. 

A hint to the other system tags $:/tags/RawMarkup.. was also added since they can also be used for the same purpose.


Jermolene:
Thanks @Telumire

PR #6668, opened 2022-04-30
Fix: eventcatcher widget - variables can be undefined

BurningTreeC:
This fixes a nasty Bug introduced by myself where the variables can be undefined

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -292,14 +292,17 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 		$tw.utils.each(selectedNode.attributes,function(attribute) {
 			variables["dom-" + attribute.name] = attribute.value.toString();
 		});
-		// Add a variable with a popup coordinate string for the selected node
-		variables["tv-popup-coords"] = "(" + selectedNode.offsetLeft + "," + selectedNode.offsetTop +"," + selectedNode.offsetWidth + "," + selectedNode.offsetHeight + ")";
+		
+		if(selectedNode.offsetLeft && selectedNode.offsetTop && selectedNode.offsetWidth && selectedNode.offsetHeight) {
+			// Add a variable with a popup coordinate string for the selected node
+			variables["tv-popup-coords"] = "(" + selectedNode.offsetLeft + "," + selectedNode.offsetTop +"," + selectedNode.offsetWidth + "," + selectedNode.offsetHeight + ")";
 
-		// Add variables for offset of selected node
-		variables["tv-selectednode-posx"] = selectedNode.offsetLeft.toString();
-		variables["tv-selectednode-posy"] = selectedNode.offsetTop.toString();
-		variables["tv-selectednode-width"] = selectedNode.offsetWidth.toString();
-		variables["tv-selectednode-height"] = selectedNode.offsetHeight.toString();
+			// Add variables for offset of selected node
+			variables["tv-selectednode-posx"] = selectedNode.offsetLeft.toString();
+			variables["tv-selectednode-posy"] = selectedNode.offsetTop.toString();
+			variables["tv-selectednode-width"] = selectedNode.offsetWidth.toString();
+			variables["tv-selectednode-height"] = selectedNode.offsetHeight.toString();
+		}
 	}
 
 	if(event && event.clientX && event.clientY) {
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -293,7 +293,7 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 			variables["dom-" + attribute.name] = attribute.value.toString();
 		});
 		
-		if(selectedNode.offsetLeft && selectedNode.offsetTop && selectedNode.offsetWidth && selectedNode.offsetHeight) {
+		if(selectedNode.offsetLeft) {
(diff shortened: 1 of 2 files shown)


core/modules/widgets/eventcatcher.js
```diff
@@ -77,6 +77,9 @@ EventWidget.prototype.render = function(parent,nextSibling) {
 			}
 			// Execute our actions with the variables
 			if(actions) {
+				if(variables === undefined) {
+					variables = $tw.utils.collectDOMVariables(selectedNode,self.domNode,event);
+				}
```
Jermolene, line 82: Is this change related to the bug fix? We're already calling collectDOMVariables shortly before, which suggests that that call is in the wrong codepath.

core/modules/utils/dom/dom.js
```diff
@@ -292,14 +292,17 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 		$tw.utils.each(selectedNode.attributes,function(attribute) {
 			variables["dom-" + attribute.name] = attribute.value.toString();
 		});
-		// Add a variable with a popup coordinate string for the selected node
-		variables["tv-popup-coords"] = "(" + selectedNode.offsetLeft + "," + selectedNode.offsetTop +"," + selectedNode.offsetWidth + "," + selectedNode.offsetHeight + ")";
+		
+		if(selectedNode.offsetLeft && selectedNode.offsetTop && selectedNode.offsetWidth && selectedNode.offsetHeight) {
```
Jermolene, line 296: Do we need to check for all four? I think the spec probably suggests that if `offsetLeft` exists then you can assume that the others exist.

Jermolene (changes requested):
Thanks @BurningTreeC.

I am a little concerned that you may not be doing the kind of testing that I've been expecting. We've all seen how apparently innocuous small refactorings can be susceptible to surprising bugs. It's good practice to make sure that you've tested every code path before submitting the PR. If you don't have time to do that, no problem, just make sure that you say so in the PR description, and leave it as a draft.

I think you'll have heard me say that I frequently find merging PRs to be more work for me than if I had done the actual coding myself. Things get worse and worse the more I feel that I can't trust incoming PRs.

I know that you are often working in areas where we don't have good test harnesses which is something that I would like to address, but in the meantime I would really appreciate all your efforts to try to improve the quality of submissions.

BurningTreeC:
> Thanks @BurningTreeC.
> 
> I am a little concerned that you may not be doing the kind of testing that I've been expecting. We've all seen how apparently innocuous small refactorings can be susceptible to surprising bugs. It's good practice to make sure that you've tested every code path before submitting the PR. If you don't have time to do that, no problem, just make sure that you say so in the PR description, and leave it as a draft.
> 
> I think you'll have heard me say that I frequently find merging PRs to be more work for me than if I had done the actual coding myself. Things get worse and worse the more I feel that I can't trust incoming PRs.
> 
> I know that you are often working in areas where we don't have good test harnesses which is something that I would like to address, but in the meantime I would really appreciate all your efforts to try to improve the quality of submissions.

Thanks @Jermolene ,

I'm really sorry for my recent PR's which lack the testing one would expect
I'll change that with all my upcoming PR's. I was away of contributing for some time and lost my workflow

core/modules/widgets/eventcatcher.js
BurningTreeC: We're calling collectDOMVariables (which instantiates the "variables" variable) if there's a selector given.
If no selector is given we don't collect DOM variables and the "variables" variable is undefined at this point.
Now that I think about it better, it's probably the best to instantiate the `variables` variable at the top as an object and pass it to the `collectDOMVariables` method

core/modules/utils/dom/dom.js
BurningTreeC: Yes you're right it should be sufficient to check for `offsetLeft` only

Jermolene:
Thanks @BurningTreeC! As individuals we're all prone to errors, the amazing thing about working in open source is the way that we can fix them as a collective.

core/modules/widgets/eventcatcher.js
BurningTreeC: I've removed the second collection of the DOMVariables and fixed the instantiation of the variables variable

core/modules/utils/dom/dom.js
BurningTreeC: I've made the check short as it should be

BurningTreeC:
@Jermolene - in the light of this PR it may be interesting to add two new variables `tv-catchernode-width` and `tv-catchernode-height` which contain the width and the height of the eventcatcher-node itself

Jermolene:
Looks good thanks @BurningTreeC 

> @Jermolene - in the light of this PR it may be interesting to add two new variables `tv-catchernode-width` and `tv-catchernode-height` which contain the width and the height of the eventcatcher-node itself

Yes. It might be better to have a more generic term than "catcher" so that we can use it consistently with other widgets. Perhaps "tv-widgetnode-height`?

saqimtiaz:
I would recommend that this PR be a self-contained bug fix and that any new variables be added in a new PR.

BurningTreeC:
Ok then this is good to go

BurningTreeC:
@Jermolene - can this be merged?
Does this look good to you?

Jermolene:
Thanks @BurningTreeC

PR #6671, opened 2022-05-02
Update katex to latest 0.15.3

linonetwo:
Also update the readme, and add instruction for later developers to update this plugin.

I replace the whole `files` folder with the latest zip content.

About how to test this locally: https://github.com/Jermolene/TiddlyWiki5/pull/4133#issuecomment-1114444634

Snippets updated according to https://github.com/Jermolene/TiddlyWiki5/discussions/6670


Jermolene:
Hi @linonetwo it's a while since I worked on a KaTeX upgrade myself, but I notice that these changes dispense with the manually modified stylesheet "./plugins/tiddlywiki/katex/files/katex.without-font-face.min.css". The reason for those changes is because the font definitions cannot work in TiddlyWiki because they are trying to reference a file. So instead we have duplicate font definitions in the stylesheet "$:/plugins/tiddlywiki/katex/styles". Unless something has changed, I would expect to have to continue with that approach in order for the fonts to work with a standalone TW.

linonetwo:
Got it, updated. But I don't know how to test if those fonts worked.

I update the CSS by copying selected lines to the `katex.without-font-face.min.css` , and found nothing changed but the version number.

pmario:
I think, there is something wrong. The existing plugin uses 1 minified katex file. Now it seems there are 92 source files. 

What's the reasoning here. There also seems to be a webpack artefact in the code. I'm not sure about that one. Will have to have a closer look

linonetwo:
@pmario Most files are in contrib folder, and is ignored by the .files.

Do you want me to delete the untracked files?

pmario:
>Do you want me to delete the untracked files?

Not sure yet. Having a look at the moment.

linonetwo:
@pmario I just unzip all dist files from katex to /files folder, I think this will make future updates convenient and make less confusion.

pmario:
>@pmario I just unzip all dist files from katex to /files folder, I think this will make future updates convenient and make less confusion.

That's right. ... I was a bit irritated by the huge number of changed files. But most (60) of them are fonts. ... 

I'm OK with the redundant files in the contrib folder, but I think the redundant 1.2 MByte source code in the files folder should _not_ be there. (katex.js and katex.mjs)

-------

The developer update info needs a bit more work. There is a `katexdemo` edition, which should be used to test the plugin. .. I'll create a PR to your repo.

linonetwo:
No, I didn't saw the ImplementationNotes. Just see it now, because I didn't saw the katexdemo... It will be better to include it into the developer.tid in katex plugin folder...

pmario:
**I did create a PR for your repo**. I did check the CSS .. It should be OK. It seems there haven't been any changes.

linonetwo:
Merged, thanks for your review and modification!

pmario:
>I notice that this version of the plugin uses JS modules, and so is no longer compatible with ES5.1.

I think that's not true. The katex.min.js has an universal loader at the top, that checks for our require mechanism. ... The minified code doesn't contain any `import` or `.mjs` snippet in it.

**The question is**, if the .js and .mjs "plain text" code should be part of the plugin. IMO they should be removed, because they don't can be used. IMO they are only confusing.

pmario:
I did test the KaTeX demo with Edge in "Internet Explorer" mode and it worked fine.

pmario:
image

linonetwo:
Yes, the mjs is not used, just unzip there.

Jermolene:
> Yes, the mjs is not used, just unzip there.

Thanks for your patience @linonetwo. I think it would be better to only include the files that we need, as we have done up until now.

linonetwo:
@Jermolene Updated and tested, I understand this will make our repo necessarily smaller.

linonetwo:
By the way I'd like to make sure: I have delete all `*.ttf` and `*.woff2` font files, as they are not registered in the `tiddlywiki.files` , we only need `"filesRegExp": "^.*\\.woff$",` files.

Jermolene:
> By the way I'd like to make sure: I have delete all `*.ttf` and `*.woff2` font files, as they are not registered in the `tiddlywiki.files` , we only need `"filesRegExp": "^.*\\.woff$",` files.

Good thinking. I think we  originally kept the ttf and woff2 files in case we needed them for some obscure platform. But now it seems reasonable to remove them.

linonetwo:
@Jermolene seems ok to be merged now?

pmario:
@Jermolene ... this should be part of 5.2.3 too

Jermolene:
Thanks for your patience @linonetwo

kookma:
@linonetwo 
I was testing new TW-KaTeX 0.15.3

- open https://tiddlywiki.com/prerelease/plugins/tiddlywiki/katex/
- create a new tiddler
- put the below formula

```
$$
\ce{}
$$
```

See the error.

I then search the https://github.com/KaTeX/KaTeX/releases and found this error was in 0.15.3 and has been addressed in 0.15.5

Suggestion: Is it possible to update KaTex to 0.15.6 or 0.16.0 (the current stable) before release of TW 5.2.3?

linonetwo:
Yes, its possible, let me do this now.

linonetwo:
#6768

linonetwo:
I still have some WIP PR, but I'm recently doing a part-time job about react-three-fiber to learn about 3D games and earn some money by the way. So maybe I have to leave them to the next release if 5.2.3 is coming soon.

PR #6675, opened 2022-05-04
Expose isMobile to info

linonetwo:
Allow knowing if the user is on a mobile phone.

截屏2022-05-04 16 28 47

I personally more and more using tw on a phone, so need this to make changes to some of my plugins. Useful in places like 

https://github.com/NicolasPetton/Notebook/blob/465c832c885894ee3c97be76449cc524e1252eb0/plugins/notebook-mobile/js/notebookSidebarNav.js#L15-L18

### DIFF
--- a/plugins/tiddlywiki/browser-sniff/browser.js
+++ b/plugins/tiddlywiki/browser-sniff/browser.js
@@ -42,7 +42,8 @@ exports.getInfoTiddlerFields = function() {
 				["is/sailfish","sailfish"],
 				["is/android","android"],
 				["is/windowsphone","windowsphone"],
-				["is/firefoxos","firefoxos"]
+				["is/firefoxos","firefoxos"],
+				["is/mobile","mobile"]
 			];
 		$tw.utils.each(mappings,function(mapping) {
 			var value = bowser.browser[mapping[1]];
--- a/plugins/tiddlywiki/browser-sniff/browser.js
+++ b/plugins/tiddlywiki/browser-sniff/browser.js
@@ -45,7 +45,7 @@ exports.getInfoTiddlerFields = function() {
 				["is/firefoxos","firefoxos"],
 				["is/mobile","mobile"]
 			];
-		$tw.browser = $tw.utils.extend($tw.browser, bowser);
+		$tw.browser = $tw.utils.extend($tw.browser, bowser.browser);
 		$tw.utils.each(mappings,function(mapping) {
 			var value = bowser.browser[mapping[1]];
 			if(value === undefined) {
--- a/plugins/tiddlywiki/browser-sniff/browser.js
+++ b/plugins/tiddlywiki/browser-sniff/browser.js
@@ -45,7 +45,9 @@ exports.getInfoTiddlerFields = function() {
 				["is/firefoxos","firefoxos"],
 				["is/mobile","mobile"]
 			];
-		$tw.browser = $tw.utils.extend($tw.browser, bowser.browser);
+		$tw.browser = $tw.utils.extend($tw.browser, {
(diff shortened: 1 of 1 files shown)


plugins/tiddlywiki/browser-sniff/browser.js
```diff
@@ -42,8 +42,10 @@ exports.getInfoTiddlerFields = function() {
 				["is/sailfish","sailfish"],
 				["is/android","android"],
 				["is/windowsphone","windowsphone"],
-				["is/firefoxos","firefoxos"]
+				["is/firefoxos","firefoxos"],
+				["is/mobile","mobile"]
 			];
+		$tw.browser = $tw.utils.extend($tw.browser, bowser.browser);
```
Jermolene, line 48: `$tw.browser` is set by the startup code, and the core does not expect it to be changed subsequently. What's the idea here?

linonetwo: Maybe I can change it to `$tw.utils.extend(bowser.browser, $tw.browser);` ( let `$tw.browser` take priority ), so nothing will be overwritten (though it's not overwritten currently too).

The idea is to add more available info to it, as it already contains some browser info. So some js plugins can use it directly. If you think its an invariant, I can also delete this line, info are already existed in `$:/info/browser` tiddlers, just need some extra step to read them.

Jermolene: The trouble is that the properties of `bowser.browser` are not under our direct control, and so subsequent updates to Bowser might overwrite important properties of our own. I'd rather explicitly import the properties that we choose to support.

Jermolene: The trouble is that the properties of `bowser.browser` are not under our direct control, and so subsequent updates to Bowser might overwrite important properties of our own. I'd rather explicitly import the properties that we choose to support.

linonetwo: Ok, only export mobile, as that is only what I need.

plugins/tiddlywiki/browser-sniff/browser.js
```diff
@@ -42,8 +42,12 @@ exports.getInfoTiddlerFields = function() {
 				["is/sailfish","sailfish"],
 				["is/android","android"],
 				["is/windowsphone","windowsphone"],
-				["is/firefoxos","firefoxos"]
+				["is/firefoxos","firefoxos"],
+				["is/mobile","mobile"]
 			];
+		$tw.browser = $tw.utils.extend($tw.browser, {
+			mobile: bowser.browser.mobile,
```
linonetwo, line 49: @Jermolene Other people can pr if they need other variables

Jermolene: Thank you @linonetwo.

I notice that we give special treatment to `bowser.browser.mobile` by exposing it through `$tw.browser.mobile`. But perhaps it makes sense to expose all of the bowser properties that we've selected, perhaps as `$tw.browser.is[xxxx]` . For example `$tw.browser.is.firefoxos, and deprecate the existing `$tw.browser.isIE` and `$tw.browser.isFirefox`.

linonetwo:
Looks like this now.



About \``deprecate the existing` $tw.browser.isIE`and`$tw.browser.isFirefox\`, I will do that in [https://github.com/tiddly-gittly/TW5-Typed,](https://github.com/tiddly-gittly/TW5-Typed,) so we get tooltip in vscode

linonetwo:
@Jermolene can this be merged?

Jermolene:
Thanks @linonetwo

PR #6677, opened 2022-05-05
docs: mark out-of-box standalone apps

linonetwo:
New users from Obsidian and Notion may want to use a GUI app that is installable and work out-of-box, I add a new tag to mark them and filter them in the Saving tiddler.

I'm not sure \[\[Standalone App\]\] is a good tag, I actually mean GUI app that works like other note-taking apps, and is dedicated for note-taking (not versatile like Beaver browser).

### DIFF
--- a/editions/tw5.com/tiddlers/saving/Saving.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving.tid
@@ -1,6 +1,6 @@
 created: 20140912140651119
 modified: 20220401151525812
-saving-browser: Firefox Chrome Edge [[Internet Explorer]] Safari Opera
+saving-browser: Firefox Chrome Edge [[Internet Explorer]] Safari Opera [[Standalone App]]
 saving-os: Windows Mac Linux Android iOS
 tags: [[Working with TiddlyWiki]]
 title: Saving
--- a/editions/tw5.com/tiddlers/saving/Saving on Android.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving on Android.tid
@@ -6,7 +6,7 @@ delivery: App
 description: Android app for saving changes locally to device storage
 method: save
 modified: 20200507103926292
-tags: Saving Android
+tags: Saving Android [[Standalone App]]
 title: Saving on Android
 type: text/vnd.tiddlywiki
 url: https://github.com/donmor/Tiddloid
--- a/editions/tw5.com/tiddlers/saving/Saving on TidGi.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving on TidGi.tid
@@ -5,7 +5,7 @@ created: 20220221080637764
 delivery: App
 description: Desktop application for serving and syncing TiddlyWiki
 method: save
-tags: Saving Mac Windows Linux
+tags: Saving Mac Windows Linux [[Standalone App]]
 title: Saving on TidGi Desktop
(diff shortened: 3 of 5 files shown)


editions/tw5.com/tiddlers/saving/Saving.tid
```diff
@@ -1,6 +1,6 @@
 created: 20140912140651119
 modified: 20220401151525812
-saving-browser: Firefox Chrome Edge [[Internet Explorer]] Safari Opera
+saving-browser: Firefox Chrome Edge [[Internet Explorer]] Safari Opera [[Standalone App]]
```
pmario, line 3: I think, we should remove "Internet Explorer" from the list. Microsoft does the same: https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support

pmario:
uuups, ... I wasn't aware of that. ... But I'd support it.

Jermolene:
Thanks @linonetwo. A few thoughts:

* We shouldn't delete the IE material. We should add a note that IE is no longer supported, and perhaps only list it in "Saving" if a checkbox labelled "archived" is checked
* The GettingStarted tiddler is definitely unsatisfactory. It works quite well as a database of every available saving mechanism, but it is not helpful to guide new users towards making a sensible choice
* With that in mind, I worry that the solution here won't help make things easier for the people we're trying to reach. An alternative approach might be to keep the "Saving" tiddler as is, but stop transcluding it within "GettingStarted"
* An alternative approach for "GettingStarted" would be a wizard that guides users through a series of questions to help them decide on the best option. We've discussed this in a ticket before, see [here](https://github.com/Jermolene/TiddlyWiki5/pull/4730#issuecomment-650551710)

pmario:
We do have a [deprecated mechanism in place](https://tiddlywiki.com/#Deprecated%20-%20What%20does%20it%20mean:%5B%5BDeprecated%20-%20What%20does%20it%20mean%5D%5D%20%5B%5BHow%20to%20apply%20custom%20styles%20by%20tag%5D%5D%20%5B%5BTiddlyWiki%20in%20the%20Sky%20for%20TiddlyWeb%5D%5D) ... May be we could use that.

linonetwo:
Yesterday one man join my QQ group complain that he doesn't know anything on the page, later I know he just click on the big "download empty" button, and he don't know what to do next because he only uses other ordinary note apps before.

So I come up with this PR, to point out the "ordinary apps" in the list. (I also think there can be a "download app" button, but that will be too much changes.

linonetwo:
I have restored the IE relates things, and `$:/deprecated` doesn't work on this, so I'm not adding it.

linonetwo:
@Jermolene Will you merge this? The wizard can still utilize this tag, if we have a wizard later...

Jermolene:
Thanks @linonetwo it's good to go but I'm afraid I have only just noticed that this PR is targetting the "master" branch. Pure docs changes that are restricted to the `/editions` folder should target the "tiddlywiki-com" branch so that they get merged to tiddlywiki.com immediately, rather than waiting for the next release.

linonetwo:
@Jermolene Updated and rebased

Jermolene:
Thanks @linonetwo much appreciated!

Jermolene:
Thanks @linonetwo

PR #6685, opened 2022-05-10
Create config.yml

linonetwo:
Guide users to create new ideas on the Discussion or the talk forum, this will make issue only for bugs, and reduce the issue count.


linonetwo:
The old feature\_request.md is not deleted because someone may still need a feature request issue for its own PR or for adding the issue to a project.

Jermolene:
Thanks @linonetwo I had no idea GitHub had templated issues, that's great.

I'm sure there's a lot of other good GitHub and npm features that we're not yet taking advantage of. Suggestions always welcome.

PR #6689, opened 2022-05-11
Markdown: Add code block button + enable shortcut

MaxGyver83:
When using the markdown plugin, you can enter fenced code blocks in markdown tiddlers but there is no button in the editor toolbar and the usual shortcut (**ctrl-shift-M**) doesn't work either.

This pull requests adds a code block button (same as for regular tiddlers) and by that, the shortcut works without further code changes.

### DIFF
--- a/plugins/tiddlywiki/markdown/EditorToolbar/mono-block.tid
+++ b/plugins/tiddlywiki/markdown/EditorToolbar/mono-block.tid
@@ -1,0 +1,17 @@
+title: $:/plugins/tiddlywiki/markdown/EditorToolbar/mono-block
+list-after: $:/core/ui/EditorToolbar/mono-block
+tags: $:/tags/EditorToolbar
+icon: $:/core/images/mono-block
+caption: {{$:/language/Buttons/MonoBlock/Caption}} (Markdown)
+description: {{$:/language/Buttons/MonoBlock/Hint}}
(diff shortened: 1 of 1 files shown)


plugins/tiddlywiki/markdown/EditorToolbar/mono-block.tid
```diff
@@ -0,0 +1,17 @@
+title: $:/plugins/tiddlywiki/markdown/EditorToolbar/mono-block
+list-after: $:/core/ui/EditorToolbar/mono-block
+tags: $:/tags/EditorToolbar
+icon: $:/core/images/mono-block
+caption: {{$:/language/Buttons/MonoBlock/Caption}} (Markdown)
+description: {{$:/language/Buttons/MonoBlock/Hint}}
+condition: [<targetTiddler>type[text/x-markdown]]
```
pmario, line 7: Condition should be like this: `[<targetTiddler>type[text/x-markdown]] [<targetTiddler>type[text/markdown]]` since both types are possible now. Also see: https://github.com/Jermolene/TiddlyWiki5/pull/6528

MaxGyver83: OK, I have changed it.

Jermolene:
Great stuff, thanks @MaxGyver83

PR #6690, opened 2022-05-12
Add additional example of tm-new-tiddler

AnthonyMuscio:
Add additional example to create a new tiddler from a template with additional attributes:

Stemmed from this discussion "Cloning a tiddler" https://talk.tiddlywiki.org/t/cloning-a-tiddler/3301/4?u=tw_tones 

Unless documented users may feel forced to move to using the ActionCreateTiddlerWidget when in fact tm-new-tiddler permits a clone to include additional parameters to be specified. I just tested this with the above example on tiddlywiki.com v5.2.2 in a tiddler tagged $:/tags/ViewTemplate 

I just provided an additional example using the existing format.

### DIFF
--- a/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-new-tiddler.tid
+++ b/editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-new-tiddler.tid
@@ -36,3 +36,12 @@ To create a new tiddler with given attributes rather than from a template:
 New Tiddler
 </$button>
 ```
+
+To create a new tiddler from a template with additional attributes:
+
+```
+<$button>
+<$action-sendmessage $message="tm-new-tiddler" $param=<<currentTiddler>> fieldname="field value" />
+New Tiddler
+</$button>
+```


editions/tw5.com/tiddlers/messages/WidgetMessage_ tm-new-tiddler.tid
```diff
@@ -36,3 +36,12 @@ To create a new tiddler with given attributes rather than from a template:
 New Tiddler
 </$button>
 ```
+
+To create a new tiddler from a template with additional attributes:
```
Jermolene, line 40: Thanks @AnthonyMuscio, a couple of points:

* "template" has a special meaning in TiddlyWiki and so the docs try to avoid other usages of the word
* the term "additional attributes" might be misleading: what's going in is that we are using additional message parameters to specify field values that should be overridden

So perhaps we could say "To create a new tiddler by cloning an existing one. Parameters not starting with `$` are used to specify field values to be overridden"

AnthonyMuscio: Agreed, thank you for doing it.

pmario: In the right corner of the ui are 3 dots. You can click them, Edit the file and change it accordingly to the conversation. So there is no need to create a new PR. You can modify this one.

Jermolene: Thanks @AnthonyMuscio no worries I've committed this with some tweaks in 96cb66a27ee604e551a1c7417e6c5b9e3f5884d2

PR #6691, opened 2022-05-12
Markdown: Add link(ify) button and enable shortcut (ctrl-L)

MaxGyver83:
When using the markdown plugin, there is neither a _link_ nor a _linkify_ button in markdown tiddlers. The corresponding shortcuts  (**shift-alt-L** and **ctrl-L**) don't work.

This pull requests adds a _link_ button (same as for regular tiddlers) which works like the _linkify_ button for regular tiddlers but it inserts `[]()` instead of `[[]]`. This enables the default _link_ shortcut **ctrl-L**.

Probably the best solution would be to add both a _linkify_ button (which inserts `[]()`) and a _link_ button (which opens a popup menu in which you can search existing tiddlers). But I don't know how to create an icon (`[*]()`) matching the style of the existing icons (image) nor do I know how I could migrate the popup menu. So this PR is not the perfect solution but nevertheless an improvement of the current situation.

### DIFF
--- a/plugins/tiddlywiki/markdown/EditorToolbar/link.tid
+++ b/plugins/tiddlywiki/markdown/EditorToolbar/link.tid
@@ -1,0 +1,15 @@
+caption: {{$:/language/Buttons/Link/Caption}} (Markdown)
+condition: [<targetTiddler>type[text/x-markdown]] [<targetTiddler>type[text/markdown]]
+description: {{$:/language/Buttons/Link/Hint}}
+icon: $:/core/images/link
+list-after: $:/core/ui/EditorToolbar/link
+shortcuts: ((link))
+title: $:/plugins/tiddlywiki/markdown/EditorToolbar/link
+tags: $:/tags/EditorToolbar
+
+<$action-sendmessage
+	$message="tm-edit-text-operation"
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @MaxGyver83 

> Probably the best solution would be to add both a _linkify_ button (which inserts `[]()`) and a _link_ button (which opens a popup menu in which you can search existing tiddlers).

That sounds good.

> But I don't know how to create an icon (`[*]()`) matching the style of the existing icons (image)

I can probably stitch together something here. I may not be able to do it for a few days though.

> nor do I know how I could migrate the popup menu.

Have you run into any particular problems?

> So this PR is not the perfect solution but nevertheless an improvement of the current situation.

I think people do sometimes hold back from making PRs with modest improvements, but in fact of course the cumulative impact is immense, and they tend to be easier to process. So do please keep them coming.

MaxGyver83:
> Have you run into any particular problems?

The problem with the link popup (dropdown) menu (`link-dropdown.tid`) was that it creates only WikiText style links. It turned out, I just had to create a function `make-markdown-link` (analogous to `make-link`). See #6693.

PR #6698, opened 2022-05-17
Allow toggling wrap-lines text operations (like mono block)

MaxGyver83:
Whenever I apply a "wrap-lines" text operation (like a code/monospace block or a block quote) by accident (p.e. a line is missing in the selection), I can't easily undo it. **ctrl-Z** does not work. And using the (p.e. block quote) shortcut again results in multiple nested blocks quotes:

image

With this pull request merged, you can just undo such an operation by pressing the same button again (or using the same shortcut again).

TiddlyWiki toggle block quote


Jermolene:
Excellent stuff, thank you @MaxGyver83

PR #6699, opened 2022-05-19
feat: inform child widget to do some custom cleanup

linonetwo:
Implements for [https://github.com/Jermolene/TiddlyWiki5/discussions/5945](https://github.com/Jermolene/TiddlyWiki5/discussions/5945)

After this, we can do:

```ts
class SomeReactWidget extends Widget {
  destroy() {
    super.destroy();
    this.reactRoot?.unmount?.();
  }
```

TLDR: use `destroy()` instead of `removeChildDomNodes()` in your widgets.

The Problem: The parent widget (in this case is the storyriver's list widget, maybe) will directly remove child dom node from itself's childlist, and won't inform child widget, and just wait for GC. So event listeners and ReactJS instances will still exist on old child dom nodes, causing problems.

In this PR: A change to `removeChildDomNodes` has been made. So it is non-recursive by default. (You can pass an option like `this.removeChildDomNodes({ recursive: true })` to get old behavior).

**This will only break plugins with widget that doesn't have any dom node, but want to help child widgets do the `removeChildDomNodes`. This is very rare**, because normally every widget has its own dom node, and its child widget will attach the child dom node to the parent dom node.

And it is recommended to use the new `destroy()` method that will call `removeChildDomNodes()` only once, and it will recursively inform children about "parent has unmounted, you don't need to do anything but you can clean yourself up".

### DIFF
--- a/core/modules/storyviews/classic.js
+++ b/core/modules/storyviews/classic.js
@@ -80,7 +80,7 @@ ClassicStoryView.prototype.remove = function(widget) {
 	if(duration) {
 		var targetElement = widget.findFirstDomNode(),
 			removeElement = function() {
-				widget.removeChildDomNodes();
+				widget.destroy();
 			};
 		// Abandon if the list entry isn't a DOM element (it might be a text node)
 		if(!targetElement || targetElement.nodeType === Node.TEXT_NODE) {
@@ -112,7 +112,7 @@ ClassicStoryView.prototype.remove = function(widget) {
 			{opacity: "0.0"}
 		]);
 	} else {
-		widget.removeChildDomNodes();
+		widget.destroy();
 	}
 };
 
--- a/core/modules/storyviews/pop.js
+++ b/core/modules/storyviews/pop.js
@@ -73,7 +73,7 @@ PopStoryView.prototype.remove = function(widget) {
 		duration = $tw.utils.getAnimationDuration(),
 		removeElement = function() {
 			if(targetElement && targetElement.parentNode) {
-				widget.removeChildDomNodes();
+				widget.destroy();
 			}
 		};
 	// Abandon if the list entry isn't a DOM element (it might be a text node)
--- a/core/modules/storyviews/zoomin.js
+++ b/core/modules/storyviews/zoomin.js
@@ -144,7 +144,7 @@ ZoominListView.prototype.remove = function(widget) {
 	var targetElement = widget.findFirstDomNode(),
 		duration = $tw.utils.getAnimationDuration(),
 		removeElement = function() {
-			widget.removeChildDomNodes();
+			widget.destroy();
 		};
 	// Abandon if the list entry isn't a DOM element (it might be a text node)
 	if(!targetElement || targetElement.nodeType === Node.TEXT_NODE) {
--- a/core/modules/widgets/importvariables.js
+++ b/core/modules/widgets/importvariables.js
@@ -116,7 +116,7 @@ ImportVariablesWidget.prototype.refresh = function(changedTiddlers) {
 	}
 	if(changedAttributes.filter || !$tw.utils.isArrayEqual(this.tiddlerList,tiddlerList) || haveListedTiddlersChanged()) {
 		// Compute the filter
-		this.removeChildDomNodes();
+		this.destroy();
 		this.execute(tiddlerList);
 		this.renderChildren(this.parentDomNode,this.findNextSiblingDomNode());
 		return true;
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -219,7 +219,7 @@ ListWidget.prototype.handleListChanges = function(changedTiddlers) {
 	} else {
 		// If the list was empty then we need to remove the empty message
 		if(prevList.length === 0) {
-			this.removeChildDomNodes();
+			this.destroy();
 			this.children = [];
 		}
 		// If we are providing an counter variable then we must refresh the items, otherwise we can rearrange them
@@ -312,7 +312,7 @@ ListWidget.prototype.removeListItem = function(index) {
 	if(this.storyview && this.storyview.remove) {
 		this.storyview.remove(widget);
 	} else {
-		widget.removeChildDomNodes();
+		widget.destroy();
 	}
 	// Remove the child widget
 	this.children.splice(index,1);
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -615,7 +615,8 @@ Widget.prototype.removeChildDomNodes = function() {
 Inform widget that extends this widget and children widgets that widget tree is about to destroy, and dom nodes are being unmounted from the document.
 */
 Widget.prototype.destroy = function(options) {
-	var removeDom = options && options.removeDom
+	// removeDom by default
+	var removeDom = (options && options.removeDom) || true;
 	if (removeDom) {
 		// prepare options for children, if we have removed the dom, child don't need to remove their dom
 		removeDom = !this.removeChildDomNodes();
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -599,21 +599,19 @@ Widget.prototype.findFirstDomNode = function() {
 /*
 Remove any DOM nodes created by this widget or its children
 */
-Widget.prototype.removeChildDomNodes = function(parentRemoved) {
+Widget.prototype.removeChildDomNodes = function() {
 	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
-	// If parent has already detatch its dom node from the document, we don't need to do it again.
-	if(this.domNodes.length > 0 && !parentRemoved) {
+	if(this.domNodes.length > 0) {
 		$tw.utils.each(this.domNodes,function(domNode) {
 			domNode.parentNode.removeChild(domNode);
 		});
 		this.domNodes = [];
-		// inform child widget to do some custom cleanup in a override sub-class method, and tell child widget that parent has already done the update, so children don't need to do anything.
-		parentRemoved = true;
+	} else {
+		// Otherwise, ask the child widgets to delete their DOM nodes
+		$tw.utils.each(this.children,function(childWidget) {
+			childWidget.removeChildDomNodes();
+		});
 	}
-	// If parentRemoved is unset or false, will ask the child widgets to delete their DOM nodes
-	$tw.utils.each(this.children,function(childWidget) {
-		childWidget.removeChildDomNodes(parentRemoved);
-	});
 };
 
 /*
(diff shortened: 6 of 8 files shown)


pmario:
Are there tests for the existing function? ... Do they still pass?

linonetwo:
I've searched, no existing test for `removeChildDomNodes` . I've done a smoke test in my plugin to show it is working.

[https://github.com/tiddly-gittly/tw-react/blob/9804fcd9870777ebe341958248d60b227c27c931/src/widget.ts#L12-L31](https://github.com/tiddly-gittly/tw-react/blob/9804fcd9870777ebe341958248d60b227c27c931/src/widget.ts#L12-L31)

Do I need to add new test for `removeChildDomNodes`?

Jermolene:
Hi @linonetwo rather than overload `removeChildDomNodes()`, I think it might be clearer to introduce a new `destroy()` method for widgets.

buggyj:
There maybe some performance issues. e.g the search box results need to be tested as this an area where there is a lot of dynamic node creation.

linonetwo:
Please review again, I extract the `destroy()`

Currently, does `destroy` only means "dom has been destroyed"? Because we will have  `refreshSelf ->  removeChildDomNodes + render -> destroy` , so child widget will receive `destroy` when parent `refreshSelf` ...

This totally make sense for my React widgets, but don't know if this makes sense for you too?

linonetwo:
@buggyj 

> a lot of dynamic node creation.

There don't seem to be this?

buggyj:
> 
@linonetwo if you have a tw with a lot of content (like at tiddlywiki.com) then typing in the search box can produce a long list  rapidly changing list of matches.

linonetwo:
@buggyj  I think that can be dealt with by a debounced input. And nothing to do with the `destroy` method. What do you think?

buggyj:
@linonetwo the search box is just an area where I think there may be performance issues, obviously having a destroy method will take some cpu time to run, the question is will they be a perceivable slow down in the human interaction will the wiki?

linonetwo:
@buggyj Recursive calling empty function will have no human noticeable performance impact. The only big performance issue I see in tw are

1.  recursive filters that do calculations like tw5-locator bimlas/tw5-locator#10 
2.  frequent dom operation in the list, that is not debounced

this destroy method are not like them.

Jermolene:
Thanks @linonetwo. That's helpful, and shows up what might be a problem.

The issue is that this PR has `removeChildDomNodes()` calling `destroy()`, which is the wrong way round: it means that calling `destroy()` on a widget won't remove it's child dom nodes, as one would expect.

I think the fundamental problem is that `removeChildDomNodes()` has been treated as a de-facto equivalent of a `destroy()` method.

So, one possible resolution is to just accept the situation by declaring that `removeChildDomNodes()` **is** our way to destroy widgets, and then review the existing widgets to make sure that they are doing any expected cleanup.

linonetwo:
> `removeChildDomNodes()` has been treated as a de-facto equivalent of a `destroy()`

Yes, but as you said, this name is not clear enough, a method named `destroy` is better.

How about let `destroy` do the recursive, and call `removeChildDomNodes`? See my latest update for example, but this will not be as backward-compatible as my last commit.

linonetwo:
And you can see test passed again after I replace all existing `removeChildDomNodes()` to `destroy()`

linonetwo:
I found `removeChildDomNodes` can mean two things

1.  selectively refresh, don't unmount, you can reuse eventListeners after shadow dom diff, case is tiddler text updated by some event
2.  full unmount, because tiddler is closed

I personally do this in my widget:

```
  // in tw-react
  removeChildDomNodes(noUnmount?: boolean) {
    super.removeChildDomNodes();
    if (noUnmount !== true) {
      this.root?.unmount?.();
    }
  }
  
  // in slate-write, subclass of tw-react's widget
  refresh(changedTiddlers: IChangedTiddlers): boolean {
    this.noUnmount = true;
    this.refreshSelf();
  }

  /** sometimes we just want refresh selectly, no full unmount, like in case of update tiddler text from outside of editor */
  private noUnmount = false;
  removeChildDomNodes() {
    super.removeChildDomNodes(this.noUnmount);
    this.noUnmount = false;
  }
```

EvidentlyCube:
If the idea is to know in a child widget whether it still is used, can't you just check its dom nodes to see if their root node? And if it's not document then unmount them - this would require no monkey patching as far as I can tell, but that assumes you don't need to know about unmounting as soon as it happens.

linonetwo:
`check its dom nodes to see if their root node` what does this mean? If unmounted, its parent will be the document?

But even this works, I still need to get informed as far as destroy, because react need to unmount ASAP to prevent duplication of event listeners and dangling state updater.

linonetwo:
@Jermolene I did a lot of

```diff
- widget.removeChildDomNodes();
+ widget.destroy();
```

in core widgets. And this will potentially break some 3rd party plugins. Do you think this is OK? But I can't think of a better way to introduce a correct `destroy()` method without separate this functionality from `removeChildDomNodes()`.

linonetwo:
> The issue is that this PR has `removeChildDomNodes()` calling `destroy()`, which is the wrong way round: it means that calling `destroy()` on a widget won't remove it's child dom nodes, as one would expect.

This was an attempt to make it backward compatible...

EvidentlyCube:
> `check its dom nodes to see if their root node` what does this mean? If unmounted, its parent will be the document?

If you call `getRootNode()` on an elements that's not added to the document it will return either itself or whater is its topmost parent: 
image

You can use that to know if your widget is unmounted or not. But if, as you say, you need to know immediately about unmounting then it obviously won't work.
I am not that familiar with how widgets work in TW so I can't gauge how much damage the PR as it stands could do to existing PRs, but I am going to guess that TW's backwards-compatibility policy will make it impossible as it stands.

I'm wondering about an alternative, though, why not something like this:
```
// widget.js
Widget.prototype.removeChildDomNodes = function() {
    if (this.domNodes.length > 0) {
        $tw.utils.each(this.children,function(childWidget) {
            childWidget.unmount();
        });
        // Rest of the code remains the same
    }
};

Widget.prototype.unmount = function() {
    $tw.utils.each(this.children,function(childWidget) {
        childWidget.unmount();
    });
};
```

And then you can override `unmount()` and `removeChildDomNodes()` in your child widget to do whatever cleanup you need. Am I missing something?

linonetwo:
I was trying to do that, but https://github.com/Jermolene/TiddlyWiki5/pull/6699#issuecomment-1133634834

linonetwo:
Updated the description in this PR [https://github.com/Jermolene/TiddlyWiki5/pull/6699#issue-1241212442](https://github.com/Jermolene/TiddlyWiki5/pull/6699#issue-1241212442)

Just wait until we are on v5.3.0 prerelease

linonetwo:
> the destroy() method will not be called for any child widgets of the old widget

Yes, but `destroy()` is mostly for some new plugins like ReactJS lib plugin, those old plugins can still use the old `removeChildDomNodes()`. As I add in the description, the `removeChildDomNodes()` will still work for them in most cases.

pmario:
>..  "in most cases."

hmmmm, ... existing widgets work "in moste cases." 

I know it is true that software "works in most cases" due to bugs but I wouldn't want that paradigm to be a goal for the core development ;)

linonetwo:
Don't worry, I just mean "in most cases." plugin developers don't need to upgrade their plugin to adapt to this breaking change. (At least for all of my plugins...)

Jermolene:
> hmmmm, ... existing widgets work "in moste cases."
> 
> I know it is true that software only "works in most cases" due to bugs but I wouldn't want that paradigm to be a goal for the core development ;)

Is that a serious comment? The context here is that we are considering a deliberate break in backwards compatibility, and that was a discussion of the impact of making that change. The specific point is that only a few old widgets will fail in the face of the proposed changes. Does that help clarify what is being discussed?

linonetwo:
> refresh themselves

I think this is handled by `refreshSelf` or `render` method.

> destroy() called and sometimes not

This method is called by

1. story view widget, so it is called when tiddler removed from view, it is recursive, so always called for each child-widget.
1. parent widget on demand, if you write a custom widget, and want to inform your child widget

buggyj:
'old' widgets are detached by calling `this.makeChildWidgets(); ` (by `this.children=[]`)

usually we have this call sequence.

refreshSelf() -> refresh()-> execute()-> makeChildWidgets();

So if a widget refreshes itself due to a change to the store, for example, all child widgets get recreated.

Jermolene:
Hi @linonetwo @buggyj it might be helpful to try to make some improvements to the docs as a foundation for these investigations.

We have a tutorial covering widget creation:

https://tiddlywiki.com/dev/#Javascript%20Widget%20Tutorial

Perhaps we could add a section "Widget Lifecycle" and try to pull the information all together in one place.

linonetwo:
That tutorial is good, what a pity that I didn't read it, and have to guess the usage from the source code. I will try update it after we have `destroy` related PR merged.

---

> if a widget refreshes itself due to a change to the store, for example, all child widgets get recreated.

@buggyj That seems to be OK, when you call `refresh`, we want widget tree to be recreated. Even in ReactJS, when `render` is called, a new react node tree is created and returned and diffed.

In tw if you don't want the recreation, you can simply return false in the `refresh` method

And it is usually `refresh` -> `refreshSelf`, from tutorial https://tiddlywiki.com/dev/#Widget%20refresh%20tutorial%20part%20II

buggyj:
@linonetwo 
I mean to write the calling as

refreshSelf() -> render()-> execute()-> makeChildWidgets();

what I was trying to point out is that widget nodes are removed by makeChildWidgets() and this can be as a results of calling refreshSelf(), and so widgets can be removed without destroy() being called first.

linonetwo:
That's true, I have to add a call to `destry` between these lines

https://github.com/Jermolene/TiddlyWiki5/blob/8090079e789f691b5743efa903e5437f5ba5e70e/core/modules/widgets/widget.js#L381-L383

Jermolene:
Thanks @linonetwo

It would be nice to have an example of a widget that implements a custom destroy method.

buggyj:
How is this to be used - say I have a function 'tidyUp()' where do I call it from?

linonetwo:
@buggyj You can call it in the `destroy`, for example, I'm going to replace https://github.com/tiddly-gittly/tw-react/blob/711edd33df92d0288d75ec98d2532e4f983e760f/src/widget.ts#L104C1-L108 with this new method

```ts
  destroy() {
    super.destroy();
    this.root?.unmount?.();
  }
```

linonetwo:
My widget can be an example, will add it to the dev doc site after this released.

buggyj:
The great thing about the pre-release is we can test to make sure that things work as we expect.

Jermolene:
The core doesn’t use JS classes and so doesn’t have access to “super”. It would be useful to have an example of a core-compliant way of achieving the same thing.

PR #6703, opened 2022-05-23
Notes on navigation to Keyboard Shortcut tiddler

EvidentlyCube:
I've been adding some fancy keyboard shortcuts and consistently struggled with two quirks of how navigation interacts with them so I think it'd be great to codify that in the documentation.

Not sure if the way I wrote it is clear enough, perhaps adding a fully-fledged example would be better:

```
<!-- Navigate to the next open tiddler shortcut-->
<!-- The navigate action must be wrapped in <$navigator> so that they are processed -->
<$navigator story="$:/StoryList" history="$:/HistoryList" openLinkFromInsideRiver={{$:/config/Navigation/openLinkFromInsideRiver}} openLinkFromOutsideRiver={{$:/config/Navigation/openLinkFromOutsideRiver}} relinkOnRename={{$:/config/RelinkOnRename}}>
<!-- We need to strip square brackets from tiddler names to avoid navigating to non-existing tiddlers -->
  <$vars prefix="[[" suffix="]]" >
    <$list variable="target" filter="[[$:/HistoryList]get[current-tiddler]next[$:/StoryList]trim<prefix>trim<suffix>]" emptyMessage="""
      <!-- If we are at the last tiddler wrap to the beginning -->
      <$action-navigate $to={{{[list[$:/StoryList]first[]trim<prefix>trim<suffix>]}}}/>
    """>
      <$action-navigate $to=<<target>>/>
    </$list>
  </$vars>
</$navigator>
```


Jermolene:
Thanks @EvidentlyCube and apologies for the delay

PR #6708, opened 2022-05-24
Fix unit test for formatDateString UTC format

### LINKED ISSUE #6707: [BUG] One unit test fails in non-UTC timezones
### Describe the bug

Ever since https://github.com/Jermolene/TiddlyWiki5/pull/6615, [one unit test](https://github.com/Jermolene/TiddlyWiki5/blob/696c5c9c7a512cb2b146b2e4df129c20cebb6515/editions/test/tiddlers/tests/test-utils.js#L81) has been failing if your time zone is different from UTC (so @Jermolene may only notice this during the summer).

The reason the test fails is because of the use of `new Date(... integer values ...)`. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date says that in this mode, "the parameter values are all evaluated against the local time zone, rather than UTC."

### Expected behavior

The test passes.

### To Reproduce

1. Ensure your time zone is not UTC+0; e.g., mine is UTC+7
2. Run `node tiddlywiki.js editions/test`
3. See error: expected '20141109104128542' to be '20141109174128542'.

The hours value will be 17 minus your time zone offset. So in my case, 17 - +7 = 10. In a UTC-4 timezone, 17 - -4 = 21. In a UTC-7 timezone (or further west), the date will also be different because 17 - -7 = 24 and wraps around midnight.

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: `master` as of 2022-05-24, i.e. commit 696c5c9c7a512cb2b146b2e4df129c20cebb6515
- Saving mechanism: N/A
- Plugins installed: N/A

### Additional context

The correct fix is probably to have the unit test use `getUTCHours`, and possibly `getUTCDate` as well (in case it's being run in timezones of UTC-7 or further west), in order to calculate the expected string. I.e.,

```js
const day = d.getUTCDate();
const dayStr = ("" + day).padStart(2, '0');
const hours = d.getUTCHours();
const expectedUtcStr = `201411${dayStr}${hours}4128542`;
expect(fds(d,"[UTC]YYYY0MM0DD0hh0mm0ssXXX")).toBe(expectedUtcStr);
```

rmunn:
Fixes #6707.

Now the test should run correctly in any timezone.

### DIFF
--- a/editions/test/tiddlers/tests/test-utils.js
+++ b/editions/test/tiddlers/tests/test-utils.js
@@ -78,7 +78,11 @@ describe("Utility tests", function() {
 		expect(fds(d,"ddd hh mm ssss")).toBe("Sun 17 41 2828");
 		expect(fds(d,"MM0DD")).toBe("1109");
 		expect(fds(d,"MM0\\D\\D")).toBe("110DD");
-		expect(fds(d,"[UTC]YYYY0MM0DD0hh0mm0ssXXX")).toBe("20141109174128542");
+		const day = d.getUTCDate();
+		const dayStr = ("" + day).padStart(2, '0');
+		const hours = d.getUTCHours();
+		const expectedUtcStr = `201411${dayStr}${hours}4128542`;
+		expect(fds(d,"[UTC]YYYY0MM0DD0hh0mm0ssXXX")).toBe(expectedUtcStr);
 
 		// test some edge cases found at: https://en.wikipedia.org/wiki/ISO_week_date
 		// 2016-11-13 is Week 45 and it's a Sunday (month nr: 10)


Jermolene:
Great stuff, thanks @rmunn that had indeed been bothering me when I was running the tests in certain situations.

Jermolene:
Great stuff, thanks @rmunn that had indeed been bothering me when I was running the tests in certain situations.

rmunn:
@Jermolene I just noticed another issue: the hours could be a single-digit number if the date had wrapped around midnight (e.g., in UTC-7 timezones or ones further west than that). I was too late to get that fix into this PR before you merged it, so please also look at #6709 and merge it, otherwise the unit tests could fail in California, Alaska, Hawaii, and any other locations close to (and east of) the international date line.

PR #6729, opened 2022-06-20
Add Note formatting to an important paragraph in ScrollableWidget

pmario:
I did test something and completely missed that paragraph. So this PR makes it more visible.


Jermolene:
Thanks @pmario

Issue #6731, opened 2022-06-21
[BUG] fieldmangler doesn't get the right parameters if fields or tags are added


pmario:
### Describe the bug

Triggered by Issue #6725 ... I did single step through the fieldmangler.js code. 

-  It seems the UI doesn't send the right parameters with the tm-add-tag and tm-add-field anymore. 
-  It seems the tm-add-tag and tm-add-field aren't sent at all.

### To Reproduce

- crate a new tiddler eg: aaa
- add tag with a leading or trailing space
- hit add button 
- ---> Problem. 

---

- create a new tiddler 
- add tag with a leading or trailing space ... .BUT 
- DON'T click add. 
- Click "Done" in the tiddler toolbar 
- ---> No Problem

### TiddlyWiki Configuration

tiddlywiki.com was used to test

pmario:
@Jermolene ... I did find a fast "hack" as a temporary fix. ... But it's only fighting the symptoms and no real fix. 

It could be done with adding 2 `trim[]` parameters to the UI. But as soon as someone uses some 3rd party theme mods, the problem could pop up again. 

 - The hack requires 1 little PR

---

 - The real fix will probably require 3 or 4 PRs since I would want to 
   - add new tests to catch DOM structure changes
   - 1 PR to make the involved UI code (3 tiddlers) more readable 
     - no tests should fail
   - 1 PR to remove redundant DOM elements
     - adjust the test because of new DOM structure
   - fix the problem
     - tests should stay green

Jermolene:
Thanks @pmario 

> * It seems the UI doesn't send the right parameters with the tm-add-tag and tm-add-field anymore.
> * It seems the tm-add-tag and tm-add-field aren't sent at all.

Are you saying that the current UI no longer relies on the fieldmangler widget, but manipulates the fields directly? If so, I don't think that's a problem; I'd like to deprecate the fieldmangler widget.

> ### To Reproduce
> * crate a new tiddler eg: aaa
> * add tag with a leading or trailing space
> * hit add button
> * ---> Problem.

But this still needs fixing somehow.

> @Jermolene ... I did find a fast "hack" as a temporary fix. ... But it's only fighting the symptoms and no real fix.
> 
> It could be done with adding 2 `trim[]` parameters to the UI. But as soon as someone uses some 3rd party theme mods, the problem could pop up again.
> 
> * The hack requires 1 little PR
> 
> * The real fix will probably require 3 or 4 PRs since I would want to
[...]

PR #6740, opened 2022-06-24
Fix 6731 tag and field name trim

pmario:
This PR contains 3 commits that fix [#6731 ](https://github.com/Jermolene/TiddlyWiki5/issues/6731)

- 3dee2704ab917f46b74fce126567c2b31eca1706 adds 2 trims to the UI
- da99b265cc0e2d65877f76b6d6a95741beedb64f increases the chance, that users will spot future problems more easily. 
  - also see your commit message from 002d47b4d91f81b91bb9025a438c7f46c9ab3fe1
  - > Now long tags wrap onto multiple lines. I think they look quite good,
and we may want to consider using the same technique in the view
template.
- 02edf8580522ae3690085a52ee734c5e00696bfb adds an HTML end-tag, that was missing. 

Tested with FF on windows. 

image

Also tested with Edge on Windows

<details><summary>Click me -- Spaces are more visible now</summary>

image

</details>


pmario:
@Jermolene ... this should be part of 5.2.3

Jermolene:
Thanks @pmario

PR #6742, opened 2022-06-28
Various accessibility improvements

ndarilek:
**Is your PR related to a problem? Please describe.**
As a screen reader user, TiddlyWiki poses some challenges to me:
* It was unclear what and where the sidebar was.
* In the sidebar, it was unclear that various buttons were toggles/tabs.
* It was unclear that the story river dynamically added to the page. I couldn't figure out why clicking links didn't take me somewhere else.

**Describe the solution you are proposing**
* The sidebar now has the `region` role with an `aria-label` caption.
* Sidebar tabs now have the `switch` role, and `aria-checked="true"` is set on whichever is selected. To my screen reader, these now present as checkboxes/switches. I preferred this to ARIA's tabs because the latter implies changing keyboard navigation, and I felt this solved the problem I experienced in the simplest way.
* Each story now has the `article` role. This not only makes it easier to navigate between individual stories, but by announcing region transitions, screen readers now inform about the presence of multiple stories and their boundaries.
* The story river is surrounded by the `main` role, making it easier to jump to the river as a whole for screen readers that support it.
* The static templates have also been updated.

There really aren't any. We can certainly shuffle things around, and more can be added, but something that creates and updates walls of text really needs regions and labels to make it more navigable.

**Additional context**
I still need to do more, but this is a lot more usable and I intend to rebuild my wikis on this foundation as soon as it's merged.

I'm not sure how to check compliance with coding style, but I did try. Also not sure to what extent I should test the setting of `aria-checked`--I did set that attribute in a block that appears to trigger on selected state being true. Advice welcome.

## Checklist before requesting a review

- [X ] Illustrate any visual changes (however minor) with before/after screenshots
- [X ] Self-review of code
- [X ] Documentation updates (for user-visible changes)
- [ ] Tests (for core code changes)
- [ ] Complies with coding style guidelines (for JavaScript code)


Jermolene:
Thanks @ndarilek that's very helpful. Nice neat, small changes to the code, but it seems like they will have a big impact.

There are three tests for the HTML generated by the tabs macro that are currently failing – are you able to see the output of the CI run above?

ndarilek:
Yup, should be fixed now, ran them locally and everything passes.

Jermolene:
> Yup, should be fixed now, ran them locally and everything passes.

Great, thank you @ndarilek.

PR #6743, opened 2022-06-28
Add `alert` role to notifications so they're automatically spoken by screen readers.

ndarilek:
Sorry, this probably should have been in my previous PR--I thought it might require an additional DOM element and that the region might not present on first write. It didn't, however.

This just adds the `alert` role so my screen reader automatically speaks save notifications and such.

### DIFF
--- a/core/modules/utils/dom/notifier.js
+++ b/core/modules/utils/dom/notifier.js
@@ -36,8 +36,9 @@ Notifier.prototype.display = function(title,options) {
 	if(!tiddler) {
 		return;
 	}
-	// Add classes
+	// Add classes and roles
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @ndarilek

Jermolene:
> Awesome! Is the next release anytime soon? Tried merging some of these into 5.1.22 and had some conflicts--I'm sure I can resolve them but I think I'd rather just wait if it won't be very long.

Yes – I am behind in announcing plans, but I'm hoping to get v5.1.22 out in the next 3 or 4 weeks.

Jermolene:
> ?? v5.2.3 ?? right ;)

Oops, yes thanks, corrected above.

PR #6746, opened 2022-06-29
Add storyTiddler to render command

FSpark:
Enable transclusion to get the title of the currently rendered Tiddler, for example, it can be used on the banner of TiddlyWiki static pages to jump to specific pages. Of course I can implement this if needed.

### DIFF
--- a/core/modules/commands/render.js
+++ b/core/modules/commands/render.js
@@ -50,7 +50,7 @@ Render individual tiddlers and save the results to the specified files
 				console.log("Rendering \"" + title + "\" to \"" + filepath + "\"");
 			}
 			var parser = wiki.parseTiddler(template || title),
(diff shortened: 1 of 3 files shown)


pmario:
It's the same as `currentTiddler`. So why don't you use the currentTiddler variable in the templates?

Jermolene:
Hi @FSpark good idea thanks.

@pmario the idea is to be compatible with templates that we already use for the story river, which uses `storyTiddler` independently of `currentTiddler`.

FSpark:
> It's the same as `currentTiddler`. So why don't you use the currentTiddler variable in the templates?

Hi, of course I tried it, `currentTiddler` would be overridden in multiple Transclusions, in the above example `currentTiddler` would just show `$:/StaticBanner`.

PR #6747, opened 2022-06-29
add Polish translation for the newly introduced English i18n string

EvidentlyCube:
Adds Polish translation for the new string added in: f02bd2392fb40a54fe21bf64dd56aa3d53ce54d4


Jermolene:
Thanks @EvidentlyCube

PR #6750, opened 2022-07-02
Update version-macros.tid

AnthonyMuscio:
Documentation macro improvement.

Replace the SPAN used in the .from-version macro with the $button widget to navigate to the release notes
to={{{ [[Release $version$]] }}}

This allows one to click the version pill and open the relevant release notes.


Jermolene:
Thanks @AnthonyMuscio that's an excellent idea. I've got a slightly different implementation in mind which I will commit shortly, but thank you for the suggestion.

PR #6756, opened 2022-07-05
Svg foreign object failing test

pmario:
This PR is related to:  Fix svg foreignObject that contains DIVs #6755 

This test **fails** if it is used with TW **v5.2.2**. ... It passes if the code from PR #6755 is active

It fails with 

```
Failures:
1) Widget module should deal with foreignObject in SVG elements defined in elements.js
  Message:
    Expected 'http://www.w3.org/2000/svg' to be 'http://www.w3.org/1999/xhtml'.
  Stack:
```

because the `<div xmlns="http://www.w3.org/1999/xhtml" ... >`  is ignored with v5.2.2 code. 

The new tests will make sure that all `if` code paths are executed

### DIFF
--- a/editions/test/tiddlers/tests/test-widget.js
+++ b/editions/test/tiddlers/tests/test-widget.js
@@ -172,7 +172,7 @@ describe("Widget module", function() {
 		expect(wrapper.firstChild.namespaceURI).toBe("http://www.w3.org/2000/svg");
 	});
 
-	it("should deal with foreignObject in SVG elements", function() {
+	it("should deal with foreignObject in SVG elements defined in elements.js", function() {
 		var wiki = new $tw.Wiki();
 		// Construct the widget node
 		var text = '<svg width="260px" height="260px"><circle cx="150" cy="150" r="100" fill="lightblue" stoke="red"/><foreignObject x="70" y="110" width="150" height="180"><div xmlns="http://www.w3.org/1999/xhtml">Here is some text that requires a word wrap, and includes a [[link to a tiddler|HelloThere]].</div></foreignObject></svg>\n';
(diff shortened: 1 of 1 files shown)


pmario:
The new tests use the same SVG code snippet used to show the problem in PR #6755

Jermolene:
Thanks @pmario.

This is actually quite tricky. One problem at the moment is that the tests are run against the fakedom, and we don't really have adequate tests that the behaviour of the `namespaceURI` property is as per the spec. Do you get the same results running these tests in the browser?

I don't see an easy way around that problem; I think it might be the moment that we need to start to run the test suite under Puppeteer in `./bin/test.sh`.

Another style issue is that the SVG fragments in the tests do not look as though they have been minimised. The tests would be easier to follow if they could be simplified to the bare minimum.

Finally, we are going to need multiple namespace tests, so I think it would be worth refactoring things now to make that easier.

pmario:
>..Do you get the same results running these tests in the browser?

If I run this PR with TW v5.2.2 I do get the same result. 

**But** there are 6 failing test from `test-fiters.js` with and without this PR active.

Jermolene:
> **But** there are 6 failing test from `test-fiters.js` with and without this PR active.

We should deal with that with another issue and PR.

pmario:
I think I can make the text-element a bit shorter, so it only contains `[[link to a tiddler|HelloThere]]` but that's mainly it. 

I think, the goal is, that a reader can use the first SVG snippet and copy paste it into a tiddler at tiddlywiki.com and they will actually see something useful. 

--------

The structure of the test has to be that complex, so all the code lines in element.js are at least executed once. 

The outer SVG will be defined using the `tagNamespaces` object which activates the very first `if (this.namespace)` statement.

The CIRCLE element runs through both `else` paths and reads the namespace from the `namespace` variable

The `foreignObject` is the one which contains the DIV with the XMLNS attribute. It creates the new DIV element and activates the inner `if (this.hasAttribute("xmlns")) {`  ... So all code paths are executed at least once now. 

The HelloThere link in the TEXT element is there to test the conversion to "href", which is a special feature of TW SVGs. We can mix SVG code with TW wikitext links

The second CIRCLE element is there to make sure that the `namespace` variable scope switches back from the DIV HTML to SVG namespace when the `foreignObject` recursion has exited. 

-------

IMO testing SVG is a very high level test and it has to be complex to test this "edge case". As we can see, less complex tests didn't catch the problem. 

I know that `expect(wrapper.firstChild.children[1].firstChild.tag).toBe("div");` is very complex, but that's a problem from the fakeDOM, since it doesn't implement a function like `wrapper.getElementById()` which would make this line much simpler. 

`firstChild`, `children`, `tag` and `namespaceURI` seem to be compatible with the DOM elements available in the browser. 

------

Last but not least, I don't know, how to make `expect(wrapper.firstChild.children[1].firstChild.tag).toBe("div");`  simpler, without adding a lot of complex code to the fakeDOM and stay compatible with the browser DOM

pmario:
>Maybe we could explore running the tests a second time in the browser with the native document object.

I did do this with FF in single step mode while creating this test and the fix PR. So I'm sure all 3 code paths in element.js are executed. And the SVG output works in FF and Edge if the code snippet from the test is copy pasted into a prerelease tiddler.

Jermolene:
> I did do this with FF in single step mode while creating this test and the fix PR. So I'm sure all 3 code paths in element.js are executed. And the SVG output works in FF and Edge if the code snippet from the test is copy pasted into a prerelease tiddler.

Indeed, the problem is not the codepaths through `element.js`, but that we're still testing against $tw.fakeDocument, rather than using the real native document object.

pmario:
>Indeed, the problem is not the codepaths through element.js, but that we're still testing against $tw.fakeDocument, rather than using the real native document object.

As you wrote we may need to add puppeteer to our build environment. ... It seems [puppeteer works with Edge](https://docs.microsoft.com/en-us/microsoft-edge/puppeteer/) .. So I could use it, but I do have concerns about the dependencies that system wants to use. I completely lost my trust in npm with all the security problems lately. So my gut-feeling tells me "no" at the moment. 

I also don't know anything about puppeteer and jasmine. I found this: https://www.testim.io/blog/puppeteer-testing-introduction/ but hadn't time to read it yet.

Jermolene:
Hi @pmario

I've used Puppeteer in a number of projects recently: [TWPub](https://github.com/TWPUB/TWPUB-Tools/blob/main/epub2twpub/text-extractor.js), the new [performance test rig](https://github.com/Jermolene/tiddlywiki-performance-test-rig), and [the lenticulator](https://github.com/Jermolene/lenticulator) (the latter is the closest to what we'd need to do here).

I understand your concerns about npm, but I don't think that we can let them stop us from doing browser testing.

By the way, we should probably switch to Playwright; it's built by some of the same people as Puppeteer, but works with a wider variety of browser engines, including Firefox.

https://playwright.dev/

pmario:
>By the way, we should probably switch to Playwright; it's built by some of the same people as Puppeteer, but works with a wider variety of browser engines, including Firefox.
>
>https://playwright.dev/

That's interesting. I did have a look at the release history. It seems the got from beta 0.10.0 to v1.0.0 in just 3 months. .. That's impressive. Had a look at an [intro-video](https://www.youtube.com/watch?v=dxbh0bl4It8&t=1s) from last month. ... The code generator and the test-debugging tools look impressive too. ... It actually looks like fun to create tests. 

Coming back to the SVG test we have here. .. I can see how the visual tests can be implemented and since it seems to have all the DOM structure info, we are able to check that. ... Good

So it will complement the existing jasmine tests, and we should probably replace those tests. which need to use the fakeDOM at the moment ... right?

pmario:
@Jermolene ... Should we close this PR, since the "tests" are not real since the work with the fake-dom. ... We may create a "Proposal" issue that we may introduce the https://playwright.dev/ test-system to be able to test UI functionality. ...

Jermolene:
> @Jermolene ... Should we close this PR, since the "tests" are not real because they work with the fake-dom. ... We may create a "Proposal" issue that we may introduce the https://playwright.dev/ test-system to be able to test UI functionality. ...

I guess that makes sense, but thanks for your work on it. We should revisit this once we get Puppeteer/Playwright into our test suite.

pmario:
Closing in favour of: https://github.com/Jermolene/TiddlyWiki5/issues/6845

PR #6760, opened 2022-07-06
Month correction

fu-sen:
The month notation is strange in the date in the default Japanese display.


Jermolene:
Great thanks @fu-sen

PR #6763, opened 2022-07-07
add polish translation to a new string added in 92c4c820

EvidentlyCube:
This is the commit where the string was introduced: https://github.com/Jermolene/TiddlyWiki5/commit/92c4c820058fdf48fd6788ebb08128b05b9f94c7


Jermolene:
Thanks @EvidentlyCube

PR #6768, opened 2022-07-10
feat: update katex to 0.16 stable

linonetwo:
@kookma find out the last 0.15.x has a bug [https://github.com/Jermolene/TiddlyWiki5/pull/6671#issuecomment-1179657604](https://github.com/Jermolene/TiddlyWiki5/pull/6671#issuecomment-1179657604) , so here is the update to the 0.16 stable version that is possibly bug-free.


Jermolene:
Thanks @linonetwo

PR #6769, opened 2022-07-10
Update TiddlyWiki in the Sky for TiddlyWeb.tid

twMat:
For e.g the listing in 
https://tiddlywiki.com/prerelease/#Deprecated%20-%20What%20does%20it%20mean


Jermolene:
Thanks @twMat – good to see you around again.

PR #6666, opened 2022-04-26
Parameterised transclusions

Jermolene:
# Introduction

This PR introduces a number of improvements and new features related to some of TiddlyWiki's most fundamental components: macros, widgets, operators and transclusion.

The motivation is to fix one of TiddlyWiki 5's early design flaws: the reliance on macros using textual substitution as the primary way to modularise and reuse wikitext and filters.

Experience has shown that while macros are a good match for a small number of tasks, they are brittle and error prone for many common operations. Over the years we have introduced mitigations for the worst problems but these have come at a cost of increased complexity.

The changes in this PR provide powerful new ways to achieve common tasks, and unlock completely new capabilities that were previously impossible in wikitext.

* **Procedures**, which are essentially what macros should have been; they work in exactly the same way except that parameters are exposed as simple variables (without the double underscores) and no textual substitution takes place
* **Custom widgets**, allowing the creation of widgets in wikitext, and the redefinition of built-in widgets
* **Functions**, a new way to encapsulate filter expressions with named parameters
* **Custom Filter Operators**, allowing functions to be used as custom filter operators
* **Parameterised transclusions**, allowing strings and wikitext trees to be passed to transclusions

Some smaller improvements are also needed to enable these new features:

* ~~New filter operators for **reading JSON data**~~ – superseded by #6936 
* ~~A new **`<$genesis>`** widget that permits any other widget to be dynamically created, with dynamic attributes~~ – cherrypicked to #6961 
* ~~A new **`<$error>`** widget for displaying core error messages~~ – cherrypicked into #6970 
* ~~More effective protection against infinitely recursive functions and procedures~~ – cherrypicked into #6970 

All of these changes are intended to be backwards compatible, and should not affect existing functionality. While they represent a new field of opportunities for wikitext authors, equally it is entirely possible for authors to ignore all these new features and continue to use TiddlyWiki 5 in the way that they have always done.

# Background

TiddlyWiki 5 macros were originally based on the technique we call "textual substitution": the string values of the parameters provided when calling a macro would be plugged into the macro definition before it was wikified in the usual way.

A typical example of the approach in early versions of TiddlyWiki 5:

```
\define mymacro(title)
<$codeblock code={{$title$}}/>
\end
```

[...]

PR #6771, opened 2022-07-11
New insertafter filter operator

### LINKED ISSUE #6551: [IDEA] insertafter operator for symmetry with insertbefore
Most listops operators have a symmetrical counterpart: append and prepend, putafter and putbefore, and so on. But while there's an insertbefore operator, there's no insertafter operator.

Create an insertafter operator whose behavior would mirror that of insertbefore: given a title in the list, it would insert its operand immediately after that title.

When insertbefore can't find the target title it's searching for, it inserts its operand at the end of the list. To mirror that behavior, insertafter would, if the target title is not found at all, insert its operand at the beginning of the list. This seems a bit counterintuitive, but the behavor of insertbefore is well-established (it was added in 5.1.14) so there's no changing it now.

The insertafter operator could insert its operand at the end of the list if its target is not found, which would feel more intuitive. But then that's the same behavor as the insertbefore operator, whereas the rest of insertafter is a mirror of insertbefore. It's impossible to change the insertbefore behavior at this point without breaking backwards compatibility, so I think having insertafter mirror insertbefore in all respects is the better alternative.

rmunn:
This creates an `insertafter` filter operator that parallels the behavior of `insertbefore`, but inserts the new title just *after* the target title in the filter input, instead of just before it. If the target is not found, the new title is inserted at the end of the list, but a suffix "start" can be used to say "Insert at the start of the list if target not found" instead. (The suffix "end" is also available in case you want to explicitly select the default behavior). See discussion in https://github.com/Jermolene/TiddlyWiki5/issues/6551#issuecomment-1076508663 for the reasoning behind this. This PR also adds the suffixes "start" and "end" to the `insertbefore` operator, so that the two operators will continue to be mirror images of each other.

Fixes #6551.

### DIFF
--- a/core/modules/filters/insertafter.js
+++ b/core/modules/filters/insertafter.js
@@ -1,0 +1,41 @@
+/*\
+title: $:/core/modules/filters/insertafter.js
+type: application/javascript
+module-type: filteroperator
+
+Insert an item after another item in a list
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Order a list
+*/
+exports.insertafter = function(source,operator,options) {
+	var results = [];
+	source(function(tiddler,title) {
+		results.push(title);
+	});
+	var target = operator.operands[1] || (options.widget && options.widget.getVariable(operator.suffix || "currentTiddler"));
+	if(target !== operator.operand) {
+		// Remove the entry from the list if it is present
+		var pos = results.indexOf(operator.operand);
+		if(pos !== -1) {
+			results.splice(pos,1);
+		}
+		// Insert the entry after the target marker
+		pos = results.indexOf(target);
+		if(pos !== -1) {
+			results.splice(pos+1,0,operator.operand);
+		} else {
+			results.push(operator.operand);
+		}
+	}
+	return results;
+};
+
+})();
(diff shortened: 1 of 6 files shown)


rmunn:
https://github.com/Jermolene/TiddlyWiki5/issues/6551#issuecomment-1179935157 proposes adding a `skip` suffix to both insertbefore and insertafter, which would mean "If the target wasn't found in the list, don't insert anything at all". I haven't yet included that suffix in this PR because I'm not quite convinced yet that that behavior really belongs on `insertbefore` and `insertafter`, as it would be a bit of a semantics change enabled by that suffix. Currently these operators will always insert the item, and only its position will change if the target wasn't found. But if I add a `skip` suffix, then the operator becomes a "conditional insert" operator in the presence of that suffix. It's possible that that behavior belongs in a `maybeinsertbefore` filter operator instead; I'd like some feedback before I add a `skip` suffix here.

rmunn:
@Jermolene I'd appreciate a review of this PR, so that I can make any necessary changes before the 5.2.3 release. It shouldn't take too long.

Jermolene:
Thanks @rmunn I think we're good to go.

I'm hoping that #6666 will get to the point where we can implement `insertbefore` and `insertafter` as custom functions that operate in terms of lower level primitives. I think we'd just need a `:foreach[]` filter run prefix to make that possible, and perhaps complex implementations like that would be simpler if we also had local sub-functions.

PR #6778, opened 2022-07-13
Show stylesheets as rendered plain text

Jermolene:
As discussed [here](https://github.com/Jermolene/TiddlyWiki5/discussions/6772#discussioncomment-3138205), it is proposed to change the default view template body for stylesheets to show their rendered output as plain text.

At the moment, the raw text of stylesheets is displayed (ie including macros and other wikitext):

image

With this PR, the same stylesheet looks like this:

image

This is the same display format that is used in control panel -> info -> advanced -> stylesheets.

I think that the rendered view is useful because it makes it easier to see what the browser is actually seeing, but I recognise that it is non-backwards compatible change so I've prepared this PR for discussion.

### DIFF
--- a/core/ui/ViewTemplate/body/rendered-plain-text.tid
+++ b/core/ui/ViewTemplate/body/rendered-plain-text.tid
@@ -1,0 +1,10 @@
+title: $:/core/ui/ViewTemplate/body/rendered-plain-text
+
+\whitespace trim
+<$wikify name="text" text={{!!text}} type={{!!type}}>
+<pre>
+<code>
+<$text text=<<text>>/>
+</code>
+</pre>
+</$wikify>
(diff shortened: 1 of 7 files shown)


kookma:
> I think that the rendered view is useful because it makes it easier to see what the browser is actually seeing

As you suggested in the Discussion page, a toggle button may be useful here!
I myself use the preview mode, but that way you have to click on edit button, click on preview button, see and close, so three four clicks away!

Jermolene:
The stylesheet output looks much clearer now that there is support for syntax highlighting:

image

(Requires the highlight plugin to be installed)

pmario:
In most of the cases I'm only interested in the "code-body" view of templates and stylesheets. 

I personally like the idea of tabs to switch between the different views much more.

kookma:
> The stylesheet output looks much clearer now that there is support for syntax highlighting:

This looks very nice! Much more readable.  So please keep this. If no highlight.js installed then the plain CSS can be seen.

Jermolene:
> This looks very nice! Much more readable. So please keep this. If no highlight.js installed then the plain CSS can be seen.

The CSS highlighting makes it much easier to see errors, which is particularly useful when using wikitext to generate dynamic stylesheets.

74ddab75fb1a12fb2e0beb8a3c00baba1f79c557 extends the syntax highlighting to the control panel listing:

image

PR #6779, opened 2022-07-14
Split the tiddler subtitle into individual segments tagged $:/tags/ViewTemplate/Subtitle

### LINKED ISSUE #6775: [IDEA] Extend tiddler subtitle to be hackable with a tag for 5.2.3?
There are a number of occasions people want to modify or add to what appears in the tiddler subtitle
* however this demands editing the core tiddler [$:/core/ui/ViewTemplate/subtitle](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FViewTemplate%2Fsubtitle)

* Allow various solutions to place content in the subtitle distributed, without editing a core tiddler or clashing with other similar solutions.
* Move the current subtitle format to a new tiddler $:/core/ui/ViewTemplate/subtitle/default as follows
* Tag this new tiddler with $:/tags/tiddler-subtitle so it is again displayed by default.
```
<$link to={{!!modifier}} />
<$view field="modified" format="date" template={{$:/language/Tiddler/DateFormat}}/>
```
* Modify the contents of $:/core/ui/ViewTemplate/subtitle as follows;

```
\whitespace trim
<$reveal type="nomatch" stateTitle=<<folded-state>> text="hide" tag="div" retain="yes" animate="yes">
<div class="tc-subtitle">
<$list filter="[all[shadows+tiddlers]tag[$:/tags/ViewTemplate/Subtitle]!has[draft.of]]" variable=subtitle-tiddler><span title={{{ [<subtitle-tiddler>get[description]] }}}><$transclude tiddler=<<subtitle-tiddler>>/>&nbsp;</span></$list>
</div>
</$reveal>
```

Modifying the core tiddler as above.

**Additional context**
* The tag named in the above code $:/tags/tiddler-subtitle could be named differently
* The description field of each tiddler so tagged is used as a tooltip to the content displayed in the subtitle
  * Thus the new tiddler can be given the following description "Modifier and modified date"
* Another field name such as tooltip may be preferred over the description field.

**Examples of additional content in the subtitle**
* The inclusion of a characters or word count on the tiddler
* An indicator for different tiddler types, as a result of a tag or field
* Conditional help or more information buttons/Indicators
* Status indicator or information
* Other date and time information eg due-date

**Please note**
I have submitted this Issue: Feature request because I do not yet believe I have the capacity to submit PR's for the core, however if a I do secure the facility I plan to submit a PR. So please provide any feedback or submit a PR on my behalf thank you.

Jermolene:
Fixes #6775

### DIFF
--- a/core/ui/ViewTemplate/subtitle.tid
+++ b/core/ui/ViewTemplate/subtitle.tid
@@ -4,7 +4,11 @@ tags: $:/tags/ViewTemplate
 \whitespace trim
 <$reveal type="nomatch" stateTitle=<<folded-state>> text="hide" tag="div" retain="yes" animate="yes">
 <div class="tc-subtitle">
-<$link to={{!!modifier}} />
-<$view field="modified" format="date" template={{$:/language/Tiddler/DateFormat}}/>
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/ViewTemplate/Subtitle]!has[draft.of]]" variable="subtitleTiddler" counter="indexSubtitleTiddler">
+<$list filter="[<indexSubtitleTiddler-first>match[no]]" variable="ignore">
+&nbsp;
+</$list>
+<$transclude tiddler=<<subtitleTiddler>> mode="inline"/>
+</$list>
 </div>
 </$reveal>
--- a/core/ui/ViewTemplate/subtitle/modified.tid
+++ b/core/ui/ViewTemplate/subtitle/modified.tid
@@ -1,0 +1,4 @@
+title: $:/core/ui/ViewTemplate/subtitle/modified
+tags: $:/tags/ViewTemplate/Subtitle
+
+<$view field="modified" format="date" template={{$:/language/Tiddler/DateFormat}}/>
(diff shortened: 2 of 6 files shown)


core/ui/ViewTemplate/subtitle.tid
```diff
@@ -4,7 +4,11 @@ tags: $:/tags/ViewTemplate
 \whitespace trim
 <$reveal type="nomatch" stateTitle=<<folded-state>> text="hide" tag="div" retain="yes" animate="yes">
 <div class="tc-subtitle">
```
pmario, line 6: I think we can get red if this div, if we add the `class="tc-subtitle"` to the reveal widget. ... I didn't test it. So it's just a guess

pmario: hmmm, If we do already touch the structure, IMO we should really try to fix it too. 

I think there is enough new code and DOM elements now, that break backwards compatibility anyway.

core/ui/ViewTemplate/subtitle.tid
```diff
@@ -4,7 +4,11 @@ tags: $:/tags/ViewTemplate
...
-<$link to={{!!modifier}} />
-<$view field="modified" format="date" template={{$:/language/Tiddler/DateFormat}}/>
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/ViewTemplate/Subtitle]!has[draft.of]]" variable="subtitleTiddler" counter="indexSubtitleTiddler">
```
saqimtiaz, line 7: Could we start using the form `!is[draft]` instead of `!has[draft.of]` when introducing new changes to the core templates?

PR #6786, opened 2022-07-15
Fix: brittle selector implementation for draggable widget

### LINKED ISSUE #6595: [BUG] selector attribute implementation for $draggable is brittle
With the sincerest of apologies, I have to report that my `selector` attribute implementation for $draggable in #6480 was not fully thought through.

There are two issues both of which only occur when using the `selector` attribute:

- dragend actions do not fire. This is a simple fix, we need to check that the event.target is the dragHandle (and not the DOM node created by the widget).
- if the drag handle DOM node is destroyed and re-created, the drag handle is no longer draggable. This happens since we make the dragHandle draggable in the `render()` method of the draggable widget.

One potential way to resolve the latter issue is to check in `refresh()` for $draggable if `refreshChildren()` returned true. If it returns true we can check if we have a drag handle and if not create one by calling `$tw.utils.makeDraggable()`. 

However, I am unsure about the merits of adding such behaviour to the refresh handling and welcome other suggestions.

saqimtiaz:
This PR fixes the brittle implementation of the selector attribute for the $draggable widget, which allows specifying a drag handle to use for the draggable widget. 

Previously the event listener was attached to the drag handle and was therefore brittle if the contents of the $draggable widget were refreshed. This PR switches to an event delegation pattern and therefore is not impacted by widget refresh.

fixes #6595

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -25,14 +25,13 @@ widget: widget to use as the context for the filter
 exports.makeDraggable = function(options) {
 	var dragImageType = options.dragImageType || "dom",
 		dragImage,
-		domNode = options.domNode,
-		dragHandle = options.selector && domNode.querySelector(options.selector) || domNode;
+		domNode = options.domNode;
 	// Make the dom node draggable (not necessary for anchor tags)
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz

PR #6787, opened 2022-07-15
Feat: dynamically refresh class for draggable widget DOM node

saqimtiaz:
This PR improves the refresh handling of the $draggable widget when the class attribute changes. Instead of re-rendering the entire widget, the class name assigned to the DOM node is updated. The implementation is consistent with recent improvements in other widgets such as $button.


Jermolene:
Thanks @saqimtiaz

PR #6789, opened 2022-07-15
Fix: restore drag and drop on mobile chrome

### LINKED ISSUE #6617: [BUG] Drag and drop does not work in mobile Chrome
Drag and drop in mobile Chrome does not work in Chrome 100.0.4896.79 on Android. However, it did work at least as recently as  Chrome 99.0.4844.88.

The `dragstart` event fires but after that no drag related events are fired (`dragenter`, `dragover`, `dragleave`, `dragend`).
I have verified that this is due to the fact that we set a data URI as the value of the `URL` data transfer format in our `dragstart` event handler:

https://github.com/Jermolene/TiddlyWiki5/blob/73138b79aaaafa32b3a0abf8ba80c96eb3a18d56/core/modules/utils/dom/dragndrop.js#L105

Commenting out this line or removing the `data:` prefix from the value assigned restores functionality.

There is a [similar bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1293803) filed for Chrome on Android two months ago which describes any MIME type other than `text/plain` causing the exact same problem with drag and drop. 

Before we report this as a new or related bug, we should confirm that the specifications for the `URL` data transfer format/MIME type do not prohibit the `data:` prefix. In particular see items 6 and 9 in the specs for `getData` at https://html.spec.whatwg.org/multipage/dnd.html#dom-datatransfer-getdata and [RFC2483](https://datatracker.ietf.org/doc/html/rfc2483) for parsing `text/uri-list` data.

saqimtiaz:
This PR restores drag and drop behaviour in mobile Chrome by only assigning the URL data type on IE.

Tested on:
* Ubuntu and Windows 10 with the latest Firefox, Chrome and IE11
* OSX with Safari and Chrome
* iPhone and the latest Chrome and Safari

**Some iPad testing would be appreciated.** 

Test file:
https://saqimtiaz.github.io/sq-tw/temp/dnd-fix.html

Fixes #6617

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -107,8 +107,9 @@ exports.makeDraggable = function(options) {
 					dataTransfer.setData("text/vnd.tiddler",jsonData);
 					dataTransfer.setData("text/plain",titleString);
 					dataTransfer.setData("text/x-moz-url","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
+				} else {
+					dataTransfer.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
 				}
-				dataTransfer.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
 				dataTransfer.setData("Text",titleString);
 				event.stopPropagation();
 			}


Jermolene:
Thank you @saqimtiaz

PR #6790, opened 2022-07-15
Update Release notes

saqimtiaz:
Removed mention of global keyboard shortcuts support for framed editor, which has since been reverted and the new PR is still pending review.


Jermolene:
Thanks @saqimtiaz

PR #6793, opened 2022-07-16
foreach filter run prefix

saqimtiaz:
This PR adds a draft for a minimalistic approach to a `:foreach` filter run prefix.
@Jermolene I would appreciate some feedback before I prepare docs and tests.


saqimtiaz:
The alternate approach would be to have the map and foreach prefixes in the same file and refactor the common code into a utility function available to both.

saqimtiaz:
Closed in favour of #6806

PR #6801, opened 2022-07-20
Make insertafter docs more similar to insertbefore

rmunn:
Now both the `insertafter` and `insertbefore` filter operators have similar wording in their documentation tiddlers, making it more clear that the "start" and "end" suffixes are only available in the two-parameter form.


Jermolene:
Thanks @rmunn

PR #6806, opened 2022-07-20
Extend :map filter run prefix with mapflat option.

saqimtiaz:
This PR extends the `:map` filter run prefix with an optional suffix `flat` to enable returning all titles returned by the run, as opposed to the default behaviour of only returning the first result.

Docs and tests have been updated.

### DIFF
--- a/core/modules/filterrunprefixes/map.js
+++ b/core/modules/filterrunprefixes/map.js
@@ -16,7 +16,9 @@ exports.map = function(operationSubFunction,options) {
 	return function(results,source,widget) {
 		if(results.length > 0) {
 			var inputTitles = results.toArray(),
-				index = 0;
+				index = 0,
+				suffixes = options.suffixes,
+				flatten = (suffixes[0] && suffixes[0][0] === "flat") ? true : false;
 			results.clear();
 			$tw.utils.each(inputTitles,function(title) {
 				var filtered = operationSubFunction(options.wiki.makeTiddlerIterator([title]),{
@@ -36,7 +38,13 @@ exports.map = function(operationSubFunction,options) {
 						}
 					}
 				});
-				results.push(filtered[0] || "");
+				if(filtered.length && flatten) {
+					$tw.utils.each(filtered,function(value) {
+						results.push(value);
+					})
+				} else {
+					results.push(filtered[0]||"");
+				}
 				++index;
 			});
 		}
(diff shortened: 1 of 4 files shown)


Jermolene:
Thanks @saqimtiaz

CodaCodr:
For those non-programmer types learning to "code" in TW, I think something like `all` might be better.

Jermolene:
The rationale for using "flat" is that it matches the common computer science terminology for functional programming primitives (and the terminology used by JS for `Array.prototype.flatMap`). Using filter/reduce/map involves a lot of learning for new users, and so it's helpful for them to be able to use existing documentation through simple Googling.

CodaCodr:
> The rationale for using "flat" is ...

"fair" :)

rmunn:
The unit test for `:map:flat` demonstrates that it does not de-duplicate results (it doesn't use the "dominant append" method), but the documentation doesn't mention that it can return multiple copies of the same list item. And the https://tiddlywiki.com/prerelease/#Map%20Filter%20Run%20Prefix%20(Examples) example doesn't show that very clearly either since the `unique[]` operator later deduplicates the results. It's good that the `unique[]` operator is in that list, since that makes it a good usage example, but I think it might be wise to also construct an example that shows `:map:flat` being used *without* being followed by `unique[]`, and how that can put duplicate items in the results. That way, people will be expecting it.

saqimtiaz:
Hi @rmunn, the :map filter run prefix never de-duplicates, with or without the new suffix. Part of the challenge with creating good examples is trying to work within the bounds of tiddlers available in TiddlyWiki.com. Can you think of an appropriate example other than just reusing the same one without the `unique[]` operator?

rmunn:
Something like what I did in the CheckboxWidget example might work, where you create a field on the example tiddler, and put the data the example needs into that field. I.e., when I updated the CheckboxWidget.tid file in https://github.com/Jermolene/TiddlyWiki5/pull/6561/files#diff-e56093cf3f33aa1a6e496c46c769cb3ce1ad20cc27849accebe0d97afeeb058f I added a `colors` field containing "red orange yellow blue", and used that field in some of the examples.

Or another example that might get the point across is something like `1 2 3 4 :map:flat[range<currentTiddler>]`, which will produce `1 1 2 1 2 3 1 2 3 4`.

PR #6807, opened 2022-07-20
Add reference to Window in Widget base class for convenience

saqimtiaz:
This PR adds a reference to the Window object in the widget base class. This is convenient when writing widgets that should work in new window as well as the main window.

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -42,6 +42,7 @@ Widget.prototype.initialise = function(parseTreeNode,options) {
 	this.variablesConstructor.prototype = this.parentWidget ? this.parentWidget.variables : {};
 	this.variables = new this.variablesConstructor();
 	this.document = options.document;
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @saqimtiaz good idea. I've no objection to trying to get this in v5.2.3 if it's ready?

saqimtiaz:
@Jermolene the tests fail under node.js as window is undefined and I haven't had a chance to consider the best way to address that. I welcome any thoughts you may have on that.

Jermolene:
> @Jermolene the tests fail under node.js as window is undefined and I haven't had a chance to consider the best way to address that. I welcome any thoughts you may have on that.

Ah! Good point. Perhaps we need a *$tw.fakeWindow* as well as our existing *$tw.fakeDocument*... The alternative would be to set it null on Node.js, and expect code to check it before usage, but that seems inconsistent with the document implementation.

PR #6809, opened 2022-07-21
Image lazy loading

saqimtiaz:
This PR extends the $image widget with support for the attribute `loading` which can be set to `lazy` to enable browser based lazy loading for images loaded from an external URI.

Unfortunately lazy loading does not seem to work for embedded base64 images.


Jermolene:
Great thanks @saqimtiaz

PR #6817, opened 2022-07-22
[doc] Update GettingStarted.tid on tw.com

twMat:
The *ultimate* goal for this tiddler probably requires a fully new approach - but meanwhile, this PR can be seen as a small improvement of what we have.

If the PR is not liked, please point out what is not good.

Note: I would like for this tiddler to also inform that native TW does have a built-in saver and that the other savers are really optional and for convenience (which is not clear in the current GettingStarted!) but I am not sure how to phrase this. This particular aspect is also related to #6810

### DIFF
--- a/editions/tw5.com/tiddlers/gettingstarted/GettingStarted.tid
+++ b/editions/tw5.com/tiddlers/gettingstarted/GettingStarted.tid
@@ -4,9 +4,14 @@ tags: [[Working with TiddlyWiki]]
 title: GettingStarted
 type: text/vnd.tiddlywiki
 
-Click here to download an empty copy of TiddlyWiki: {{$:/editions/tw5.com/snippets/download-empty-button}}
+!!!1. Download an empty copy of TiddlyWiki
 
-The next step is to choose a method for saving changes. There's a wide variety of methods available, with different features and limitations. Click on the badge for a method to see more information about it. You can also click on one of the platform filters to restrict the listing to methods that work with that platform.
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/gettingstarted/GettingStarted.tid
```diff
@@ -4,9 +4,14 @@ tags: [[Working with TiddlyWiki]]
 title: GettingStarted
 type: text/vnd.tiddlywiki
 
-Click here to download an empty copy of TiddlyWiki: {{$:/editions/tw5.com/snippets/download-empty-button}}
+!!!1. Download an empty copy of TiddlyWiki
 
-The next step is to choose a method for saving changes. There's a wide variety of methods available, with different features and limitations. Click on the badge for a method to see more information about it. You can also click on one of the platform filters to restrict the listing to methods that work with that platform.
+<div style="display: flex; justify-content: center;">{{$:/editions/tw5.com/snippets/download-empty-button}}</div>
```
twMat, line 9: OK, I read elsewhere that we categorically don't use inline styling ever in the docs. I'll change this (...but I'd still appreciate feedback if there's anything else that is not *better* than the current doc.)

Jermolene: The comment you read was about inline style blocks (ie `<style>` elements in ordinary tiddlers), not about ordinary inline styling.

twMat:
I note the most important part of this PR, i.e to mention Tiddlyhost, has been dealt with so I'm closing this.
Thank you.

PR #6821, opened 2022-07-23
Add description and examples comparing :map vs. :map:flat

btheado:
Added more description of how :map compares to :map:flat and added several more examples.

Most of the examples are based on the suggestion by @rmunn in the comments of #6806

### DIFF
--- a/editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
+++ b/editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
@@ -1,5 +1,5 @@
 created: 20210618134753828
-modified: 20220720191457421
+modified: 20220723142748506
 tags: [[Filter Syntax]] [[Filter Run Prefix Examples]] [[Map Filter Run Prefix]]
 title: Map Filter Run Prefix (Examples)
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@btheado thank you, much appreciated.

I think we need to clarify the terminology a bit as the added comparison makes it sound like `map:flat` is a separate filter run prefix, which it is not. `flat` is a suffix for the `:map` filter run prefix.

btheado:
@saqimtiaz, thanks for the feedback and I agree. I am finding it more cumbersome to write about the "`:map` filter prefix with the `:flat` suffix" and  the "`:map` filter prefix without the `:flat` suffix".

I will work on it.

btheado:
@saqimtiaz, I rewrote the description. The way I understand the code, my rewritten mention of the first operator is accurate. If you think it still adds confusion I can remove it.

editions/tw5.com/tiddlers/filters/syntax/Map Filter Run Prefix (Examples).tid
```diff
@@ -20,12 +20,27 @@ Get the tags of all tiddlers tagged `Widget:`
 <<.operator-example 3 "[tag[Widgets]] :map:flat[tagging[]] :and[!is[blank]unique[]]">>
 <<.tip "Without the `flat` suffix the `:map` filter run only returns the first result for each input title">>
 
+!! Comparison between `:map` with and without the `flat` suffix
+
+The `:map` filter run will return at least as many outputs as given in the input.  By default one input item will result in exactly one output item. When the filter run transforms an input item into an empty result, the output for that item will be an empty string. When the filter run transforms an input item into multiple items, only the first item will appear in the output. This behavior can be overridden by providing the `flat` suffix. The `flat` suffix will cause all the items to appear in the output.
+
+Any `:map` filter run can be expressed with `:map:flat` by adding a call to the [[first operator|first Operator]] at the end of the filter run.
```
saqimtiaz, line 27: @btheado the rest looks good but this line is extremely confusing and I think should be dropped.

btheado: Removed

Jermolene:
Thank you @btheado @yrashk, great stuff.

PR #6823, opened 2022-07-24
feat: support arbitrary DOM attributes on DOM nodes generated by widgets

saqimtiaz:
This PR is a proof of concept to facilitate discussion of a new affordance to specify DOM node attributes on DOM nodes generated by widgets.

Any attributes prefixed with `dom-` would be assigned to the DOM node after removal of the `dom-` prefix.

`<$button tag="div" dom-data-item="abc" dom-style.border="1px solid red" dom-style.height="500px"/>`

renders as:
`<div data-item="abc" class="" style="border: 1px solid red; height: 500px;"></div>`

_Note that attributes prefixed with `dom-style.` are being treated as we do `style.` prefixed attributes on HTML elements. This support could be omitted, or made consistent with the implementation for HTML elements._

**Salient code changes**:
* extends `Widget.prototype.assignAttributes` with the option to only assign attributes with a specified prefix
* extends ButtonWidget as an example of using `widget.assignAttributes` to assign attributes to the DOM node generated.
* any attributes assigned by explicit widget attributes take priority.

Widgets such as `$scrollable` that generate more than one DOM node can choose to use different prefixes to support attributes on each DOM node.

### DIFF
--- a/core/modules/widgets/button.js
+++ b/core/modules/widgets/button.js
@@ -40,6 +40,7 @@ ButtonWidget.prototype.render = function(parent,nextSibling) {
 		tag = this.buttonTag;
 	}
 	domNode = this.document.createElement(tag);
+	this.assignAttributes(domNode,{requirePrefix: "dom-"});
 	this.domNode = domNode;
 	// Assign classes
 	var classes = this["class"].split(" ") || [],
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @saqimtiaz – looks right, do we need to add refresh handling?

saqimtiaz:
Closing in favour of #7769

PR #6824, opened 2022-07-25
Fix copy-and-paste mistake in insertafter docs

rmunn:
In #6801 I copied a section of text from the `insertbefore` operator to use in the `insertafter` docs, and in *most* of the copied text I replaced `.place B` with `.place A` since in `insertafter`, the parameter is called A rather than B. But I just noticed that there's one that I missed; this PR fixes that.


Jermolene:
Thanks @rmunn

PR #6825, opened 2022-07-25
fix #6782, remove hardcoded spaces and replace with SPAN and proper class

### LINKED ISSUE #6782: [BUG] toc-macros produce whitespace in the DOM because \whitespace trim is missing
### Describe the bug

toc-macros produce whitespace in the DOM because \whitespace trim is missing

See the screenshot. It's a bit hard to see, but there is some leading whitespace, that causes the underline on hover to be too long. 

### Expected behavior

_No response_

### To Reproduce

Visit: https://tiddlywiki.com/#Table-of-Contents%20Macros%20(Examples)
and select the Expadable tab. See screenshot

### Screenshots

image

### TiddlyWiki Configuration

v5.2.2 windows FF

### Additional context

_No response_

pmario:
fix #6782

The issue was first raised at Talk: https://talk.tiddlywiki.org/t/tiddler-link-underline-too-long/3913 ... Since the toc-macros have already been changed since 5.2.2, the fix is different to what was suggested at Talk. ... It's actually simpler now. 

- The PR removes hardcoded spaces and replaces them with SPAN and a proper class `tc-tiny-gap-left`
- The PR also adds a `tc-toc-caption` class, which didn't exist yet. So users have a class, they can use to style the caption if needed. 
- The PR also improves the possibility to copy-paste tiddler titles from the TOC since there will be no leading space in the clipboard, which can cause problems when pasted 

Before

image

After

image

### DIFF
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -3,11 +3,13 @@ tags: $:/tags/Macro
 
 \define toc-caption()
 \whitespace trim
+<span class="tc-toc-caption tc-tiny-gap-left">
 <$set name="tv-wikilinks" value="no">
   <$transclude field="caption">
     <$view field="title"/>
   </$transclude>
 </$set>
+</span>
 \end
 
 \define toc-body(tag,sort:"",itemClassFilter,exclude,path)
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario

PR #6828, opened 2022-07-26
Wikitext docs overhaul first step

pmario:
- The WikiText tag now creates a more logical sort order
- There is a button to open all wikitext rules in the StoryRiver
- Formatting in WikiText uses a table now and has more info
- There is a new wikitext-template tiddler, that adds a footer to every tiddler tagged WikiText. The footer links back to the WikiText tiddler. 

There may be more steps, but this could be a start. This [discussion at Talk](https://talk.tiddlywiki.org/t/a-better-documentation-for-wikitext/4066) triggered the PR.

### DIFF
--- a/editions/tw5.com/tiddlers/concepts/WikiText.tid
+++ b/editions/tw5.com/tiddlers/concepts/WikiText.tid
@@ -1,15 +1,23 @@
 created: 20131205155227468
-modified: 20140919191220377
+list: [[Headings in WikiText]] [[Paragraphs in WikiText]] [[Hard Linebreaks in WikiText]] [[Formatting in WikiText]] [[Dashes in WikiText]] [[Block Quotes in WikiText]] [[Lists in WikiText]] [[Definitions in WikiText]] [[Tables in WikiText]] [[Code Blocks in WikiText]] [[Typed Blocks in WikiText]] [[Images in WikiText]] [[Horizontal Rules in WikiText]] [[Linking in WikiText]] [[Variables in WikiText]] [[Macro Definitions in WikiText]] [[Macro Calls in WikiText]] [[Macros in WikiText]] [[Transclusion in WikiText]] [[Transclusion and Substitution]] [[Styles and Classes in WikiText]] [[HTML in WikiText]] [[Widgets in WikiText]] [[WikiText Parser Modes]]
+modified: 20220726132727329
 tags: Concepts Reference
 title: WikiText
 type: text/vnd.tiddlywiki
+
+\define openAll()
+<$set name=wList filter="[[WikiText]] [enlist{WikiText!!list}]">
+<$action-setfield $tiddler="$:/StoryList" list=<<wList>>/>
+\end
 
 ~WikiText is a concise, expressive way of typing a wide range of text formatting, hypertext and interactive features. It allows you to focus on writing without a complex user interface getting in the way. It is designed to be familiar for users of [[MarkDown|http://daringfireball.net/projects/markdown/]], but with more of a focus on linking and the interactive features.
 
 ~WikiText can also be inserted to the text field using the [[Editor toolbar]].
 
-See [[Formatting text in TiddlyWiki]] for an introduction to WikiText.
+The following elements of WikiText syntax are built into the core. 
 
-The following elements of WikiText syntax are built into the core:
+<$button actions=<<openAll>> >Open all tiddlers from the list below</$button>
 
 <<list-links "[tag[WikiText]]">>
+
+Also see [[Formatting text in TiddlyWiki]] for an introduction to WikiText.
--- a/editions/tw5.com/tiddlers/system/$__editions_tw5.com_wikitext-template.tid
+++ b/editions/tw5.com/tiddlers/system/$__editions_tw5.com_wikitext-template.tid
@@ -1,0 +1,13 @@
+code-body: yes
+created: 20220726132935807
+list-after: $:/core/ui/ViewTemplate/body
+modified: 20220726134215137
+tags: $:/tags/ViewTemplate
+title: $:/editions/tw5.com/wikitext-template
+
+<$list filter="[all[current]tag[WikiText]]" variable="listItem">
+
+---
+
+[[All options to format wikitext can be found at WikiText|WikiText]]
+</$list>
--- a/editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
+++ b/editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
@@ -1,19 +1,34 @@
 caption: Formatting
 created: 20131205155959399
-modified: 20220513120211686
+modified: 20220726132504220
 tags: WikiText
 title: Formatting in WikiText
 type: text/vnd.tiddlywiki
 
+<style>
+.no-code-border > code {
+  border: initial;
+  background-color: initial;
+}
+</style>
 Available character formatting includes:
 
-* <code>&#96;backticks&#96;</code> for `code` (<<.icon $:/core/images/mono-line>>)
-** Alternatively, <code>&#96;&#96;double backticks allows &#96;embedded&#96; backticks&#96;&#96;</code><br><br>
-* `''bold''` for ''bold text'' (<<.icon $:/core/images/bold>>)<br><br>
-* `//italic//` for //italic text// (<<.icon $:/core/images/italic>>)<br><br>
-* `__underscore__` for __underscored text__ (<<.icon $:/core/images/underline>>)<br><br>
-* `^^superscript^^` for ^^superscripted^^ text (<<.icon $:/core/images/superscript>>)<br><br>
-* `,,subscript,,` for ,,subscripted,, text (<<.icon $:/core/images/subscript>>)<br><br>
-* `~~strikethrough~~` for ~~strikethrough~~ text (<<.icon $:/core/images/strikethrough>>)
+<div class="no-code-border">
+
+|Wikitext | Editor Button |HTML |Rendered Output |h
+|```backticks` ``are used for  for code | (<<.icon $:/core/images/mono-line>>) |`<code>backticks</code>` are used for code|`backticks` are used for  for code |
+|2 single hyphens are used for `''bold text''`| (<<.icon $:/core/images/bold>>) |2 single hyphens are used for `<strong>bold text</strong>` |2 single hyphens are used for ''bold text''|
+|2 slashes are used for `//italic text//`| (<<.icon $:/core/images/italic>>) |2 slashes are used for `<em>italic text</em>` |2 slashes are used for //italic text//|
+|2 underscores are used for `__underscored text__`| (<<.icon $:/core/images/underline>>) |2 underscores are used for `<u>underscored text<u>` |2 underscores are used for __underscored text__|
+|2 circumflex accents are used for `^^superscripted^^` text | (<<.icon $:/core/images/superscript>>) |2 circumflex accents are used for `<sup>superscripted</sup>` text |2 circumflex accents are used for ^^superscripted^^ text |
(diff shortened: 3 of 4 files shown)


pmario:
There is a JSON file, which can be imported, to see what's going on

[wikitext-pr-step-1.zip](https://github.com/Jermolene/TiddlyWiki5/files/9190383/wikitext-pr-step-1.zip)

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -1,19 +1,34 @@
...
-** Alternatively, <code>&#96;&#96;double backticks allows &#96;embedded&#96; backticks&#96;&#96;</code><br><br>
-* `''bold''` for ''bold text'' (<<.icon $:/core/images/bold>>)<br><br>
-* `//italic//` for //italic text// (<<.icon $:/core/images/italic>>)<br><br>
-* `__underscore__` for __underscored text__ (<<.icon $:/core/images/underline>>)<br><br>
-* `^^superscript^^` for ^^superscripted^^ text (<<.icon $:/core/images/superscript>>)<br><br>
-* `,,subscript,,` for ,,subscripted,, text (<<.icon $:/core/images/subscript>>)<br><br>
-* `~~strikethrough~~` for ~~strikethrough~~ text (<<.icon $:/core/images/strikethrough>>)
+<div class="no-code-border">
+
+|Wikitext | Editor Button |HTML |Rendered Output |h
```
twMat, line 18: Second col has centered headline text.

Maybe change third col heading "HTML" into "HTML equivalent" or "rendered HTML" or some such because the people reading this whole tid are typically beginners and may otherwise believe that they need to know this (i.e learn html)

Overall for the table, I imagine an English teacher complaining about using "2" instead of "Two" or "Double".

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -1,19 +1,34 @@
...
+|```backticks` ``are used for  for code | (<<.icon $:/core/images/mono-line>>) |`<code>backticks</code>` are used for code|`backticks` are used for  for code |
```
twMat, line 19: IMO the output looks better with `` `backticks` ``, i.e symmetrical spaces around it. Also note the double "for" in both the first and the last table columns.

twMat:
IMO the output from a simple list-links macro is not particularly elegant for longer lists. I would prefer either a more hierarchical list with proper headings and subheadings (e.g "Complex elements" listing e.g tables and lists etc). Even more readable would be a ToC widget presenting these items, instead of a plain list.

twMat:
>[All options to format wikitext can be found at WikiText](https://tiddlywiki.com/prerelease/#WikiText)

Maybe instead:

>See [WikiText](https://tiddlywiki.com/prerelease/#WikiText) for the full list of formatting options.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
pmario: >Second col has centered headline text.

That's intentional. The buttons are centred too. It looks better.

>Maybe change third col heading "HTML" into "HTML equivalent" or "rendered HTML" or some such because the people reading this whole tid are typically beginners and may otherwise believe that they need to know this (i.e learn html)

I think I'll remove that column and add the HTML output to the end of the tiddler. eg: For those who are interested. 

>"2" instead of "Two" or "Double".  

That's intentional 2 ... IMO the number is much more visible as the word two. It's especially needed for `''bold''`

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
pmario: My CSS settings got messed up. It should look like this: 

grafik

pmario:
@twMat

I think the following table is much more useful as the first one:

editions/tw5.com/tiddlers/concepts/WikiText.tid
```diff
@@ -1,15 +1,23 @@
 created: 20131205155227468
-modified: 20140919191220377
+list: [[Headings in WikiText]] [[Paragraphs in WikiText]] [[Hard Linebreaks in WikiText]] [[Formatting in WikiText]] [[Dashes in WikiText]] [[Block Quotes in WikiText]] [[Lists in WikiText]] [[Definitions in WikiText]] [[Tables in WikiText]] [[Code Blocks in WikiText]] [[Typed Blocks in WikiText]] [[Images in WikiText]] [[Horizontal Rules in WikiText]] [[Linking in WikiText]] [[Variables in WikiText]] [[Macro Definitions in WikiText]] [[Macro Calls in WikiText]] [[Macros in WikiText]] [[Transclusion in WikiText]] [[Transclusion and Substitution]] [[Styles and Classes in WikiText]] [[HTML in WikiText]] [[Widgets in WikiText]] [[WikiText Parser Modes]]
```
Jermolene, line 2: I'm not sure that we should rearrange these items. The disadvantage is that it makes it much harder to visually scan the list of children to look for a specific entry.

As a general rule, I think it's only advisable to have manual ordering for lists of less than about 7 items.

editions/tw5.com/tiddlers/concepts/WikiText.tid
```diff
@@ -1,15 +1,23 @@
...
+modified: 20220726132727329
 tags: Concepts Reference
 title: WikiText
 type: text/vnd.tiddlywiki
 
+\define openAll()
```
Jermolene, line 8: If this is a useful feature then it should be made universally available, and not just buried in one particular docs tiddler.

One of our key goals with the documentation is consistency. The value of consistency across the docs more than outweighs the advantages of a random, isolated improvement like this.

editions/tw5.com/tiddlers/concepts/WikiText.tid
```diff
@@ -1,15 +1,23 @@
...
+<$set name=wList filter="[[WikiText]] [enlist{WikiText!!list}]">
+<$action-setfield $tiddler="$:/StoryList" list=<<wList>>/>
+\end
+
 ~WikiText is a concise, expressive way of typing a wide range of text formatting, hypertext and interactive features. It allows you to focus on writing without a complex user interface getting in the way. It is designed to be familiar for users of [[MarkDown|http://daringfireball.net/projects/markdown/]], but with more of a focus on linking and the interactive features.
 
 ~WikiText can also be inserted to the text field using the [[Editor toolbar]].
 
-See [[Formatting text in TiddlyWiki]] for an introduction to WikiText.
+The following elements of WikiText syntax are built into the core.
```
Jermolene, line 17: Do we use the phrase "elements of wikitext syntax" elsewhere in the docs? Again, we should try to be consistent with terminology.

We could say the "The following WikiText syntax rules are built into the core", but perhaps there is another collective noun/phraser that we already use.

editions/tw5.com/tiddlers/system/$__editions_tw5.com_wikitext-template.tid
```diff
@@ -0,0 +1,13 @@
+code-body: yes
+created: 20220726132935807
+list-after: $:/core/ui/ViewTemplate/body
+modified: 20220726134215137
+tags: $:/tags/ViewTemplate
+title: $:/editions/tw5.com/wikitext-template
+
+<$list filter="[all[current]tag[WikiText]]" variable="listItem">
```
Jermolene, line 8: Again this feels like a local, isolated fix to a universal issue. All we're doing here is providing a link back to one of the tags on the current tiddler. That's already possible using the tag pill. Randomly providing another means to do the same thing just makes it look like we don't trust the tagging mechanism as a means of user navigation.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -1,19 +1,34 @@
...
 tags: WikiText
 title: Formatting in WikiText
 type: text/vnd.tiddlywiki
 
+<style>
+.no-code-border > code {
+  border: initial;
+  background-color: initial;
+}
+</style>
```
Jermolene, line 13: (a) the core **never** uses local style blocks, nor do we recommend their use

(b) why aren't these styles in one of the docs stylesheets?

Jermolene (commented):
Thanks @pmario just a few minor comments.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
pmario: b) Because I forgot to move them sorry.

editions/tw5.com/tiddlers/system/$__editions_tw5.com_wikitext-template.tid
pmario: From the feedback we get from new users. It isn't enough, to have "hidden information" behind the tag-pill. Because new users don't know about it. 

If I search for "Formatting" in the sidebar search I may find: "Formatting in Wikitext" and open it. ... My expectation is, that I do get all the info. ... But the tiddler only contains one part of the info. ... That's why I think it makes sense to tell the users, that there is more.

Jermolene: That's all reasonable logic, but if it is true then it would actually be an argument for adding breadcrumbs to each docs tiddler, not for making a random isolated improvement for a single tag.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
Jermolene: Ha! No problem

editions/tw5.com/tiddlers/system/$__editions_tw5.com_wikitext-template.tid
pmario: OK. What about a "tip" eg: `<<.tip "The WikiText tag-pill at the top of the tiddler will show you all wikitext rules">> `

pmario: eg: IMO It's a call to action to "use the tagging system" and if they click the link the effect is similar

grafik

pmario:
>Again this feels like a local, isolated fix to a universal issue. All we're doing here is providing a link back to one of the tags on the current tiddler. That's already possible using the tag pill. Randomly providing another means to do the same thing just makes it look like we don't trust the tagging mechanism as a means of user navigation

Yes. I think we should be allowed to ask that question, because it's "hidden information"

I think Roam and others clearly show, that users these days want visual clues about back/linking. We've had that feature for almost 20 years and they can make a hype out of an ancient feature?! ... Sure because nobody saw it in our wikis. It's still hidden. 

I think [TagglyTagging](http://mptw.tiddlyspot.com/#TagglyTagging) is still a thing and it should be easy to implement it. Not at the top of the tiddler, but the bottom. In the footer.

I think the docs should have a **visible** backlinks or / and keywords area in the footer of every tiddler. ... But that's way above the scope of this little PR. 

Mehregan and Stroll are TW5 examples. 
Mehregan: 

grafik

Stroll

grafik

pmario:
If I click a tag in my wikis I do get this: 

grafik

twMat:
Tag pill ... hidden information ... Maybe tag pills could feature a small indicator?  image

pmario:
>Tag pill ... hidden information ... Maybe tag pills could feature a small indicator? image

This would be a core change. 

I think you should open a different issue for that change. It's outside the scope for this PR, which only changes documentation related tiddlers.

pmario:
This PR will be closed in favour of a new one

PR #6841, opened 2022-07-28
fix #6840 tab macro stat handling

### LINKED ISSUE #6840: [BUG] The tabs macro doesn't handle the state parameters safely anymore
### Describe the bug

In the tabs macro, the `tabState` that hosts the selected tab is [parsed as a filter](https://github.com/Jermolene/TiddlyWiki5/blob/ff42a9e4d44e659931372d4109852d266a18f82c/core/wiki/macros/tabs.tid#L50-L51) rather than a string:

```
<$qualify title=<<__state__>> name="qualifiedState">
<$set name="tabsState" filter={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>
```

As a result, should the `state` or `explicitState` strings trigger a filter parsing error, the  error string then replaces the state parameter value.

### Expected behavior

The title of the state tiddler set by a user tab selection should be the value of the `explicitState` parameter if it exists else the qualified version of the `state` parameter (aka `qualifiedState`).

### To Reproduce

2. Open a new tiddler
3. Paste the following code
```
\define myState() :state title that starts with a colon

> Should the title of the current tab be {{{ [<myState>get[text]] :else[[empty]] }}}?
> see <<list-links "[has[modified]!sort[modified]limit[3]]">>

<$macrocall $name=tabs 
      tabsList="[tag[sampletab]nsort[order]]" 
      explicitState=<<myState>> />
```
4. Save the tiddler and click on any tab.
5. See how the expected title for the current tab doesn't change and see the reason why: a state tiddler is created with the error string instead of the `explicitState` parameter.

### Screenshots

Capture d’écran 2022-07-28 à 15 58 24

### TiddlyWiki Configuration

- Version  v5.2.3
Note that the actions above give the expected result when using v5.2.2.

### Additional context

After editing https://tiddlywiki.com/prerelease/#%24%3A%2Fcore%2Fmacros%2Ftabs so that the `tabsState` variable is a string rather than a filter, as in:
```
<$set name="tabsState" value={{{ [<__explicitState__>minlength[1]] :else[<qualifiedState>] }}}>
```
the issue disappears.

Having gone through this, I realise that I should have  made a pull request rather than raise an issue, but I don't have a github repo handy, sorry!

pmario:
This pr fixes #6840 tab macro stat handling

code to test:

```
\define myState() :state title that starts with a colon

> Should the title of the current tab be {{{ [<myState>get[text]] :else[[empty]] }}}?
> see <<list-links "[has[modified]!sort[modified]limit[3]]">>

<$macrocall $name=tabs 
      tabsList="[tag[sampletab]nsort[order]]" 
      explicitState=<<myState>> />
```

### DIFF
--- a/core/wiki/macros/tabs.tid
+++ b/core/wiki/macros/tabs.tid
@@ -48,7 +48,7 @@ code-body: yes
 \define tabs(tabsList,default,state:"$:/state/tab",class,template,buttonTemplate,retain,actions,explicitState)
 \whitespace trim
 <$qualify title=<<__state__>> name="qualifiedState">
-	<$set name="tabsState" filter={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>
+	<$let tabsState={{{ [<__explicitState__>minlength[1]] ~[<qualifiedState>] }}}>
 		<div class={{{ [[tc-tab-set]addsuffix[ ]addsuffix<__class__>] }}}>
 			<div class={{{ [[tc-tab-buttons]addsuffix[ ]addsuffix<__class__>] }}}>
 				<<tabs-tab-list>>
@@ -58,6 +58,6 @@ code-body: yes
 				<<tabs-tab-body>>
 			</div>
 		</div>
-	</$set>
+	</$let>
 </$qualify>
 \end


Jermolene:
Thanks @pmario

PR #6844, opened 2022-07-28
Fix invalid file path in tiddlywiki-com branch

saqimtiaz:
This is the same fix applied to the master branch earlier.


pmario:
@saqimtiaz ... Is it possible that the problem comes from the pr-maker? May be if tiddlers don't have a default path assigned?

Jermolene:
Thanks @saqimtiaz

saqimtiaz:
> @saqimtiaz ... Is it possible that the problem comes from the pr-maker? May be if tiddlers don't have a default path assigned?

That is indeed the origin of the invalid path in this case. We never considered the use case of someone making a PR for system tiddlers using the pr-maker. An interim solution would be to strip the system prefix off of system tiddlers. Long term what we really need to do is finish the work on $:/config/FileSystemPaths for the tw-com edition that we started on Talk a while back.

Jermolene:
> Long term what we really need to do is finish the work on $:/config/FileSystemPaths for the tw-com edition that we started on Talk a while back.

Yes indeed it would be good to pick that up.

Issue #2228, opened 2016-01-16
Add wikitext for default highligt: @@highlight@@


twMat:
`@@background:yellow; content@@` works but there is no default option without parameters.

I propose `@@content@@` as a quick way to get the typical faint(!) yellow background on the content.

tobibeer:
I'd agree that such a basic highlight is useful.

twMat:
bump

linonetwo:
So you mean transform `@@background:yellow; content@@` to a `span` tag, and `@@content@@` to `mark` tag.

pmario:
@linonetwo ... `@@` is about [Styles and classes](https://tiddlywiki.com/#Styles%20and%20Classes%20in%20WikiText) ... imo not the same as `<mark>`

linonetwo:
You are right, `@@` only add style or class to the outer element, not adding a warper element.

Maybe `@@highlight@@ ` can add a special highlight class, that can be config in the palette.

pmario:
I think the OT is actionable and imo makes sense, so I'll bump the issue

Jermolene:
It's a good idea. I don't think there are any backwards compatibility considerations because `@@foobar@@` at present doesn't do anything useful.

twMat:
So no misunderstanding: I'm hoping it will work also for `@@strings with spaces@@`

pmario:
>So no misunderstanding: I'm hoping it will work also for @@strings with spaces@@

Yes is will. It's like any other "emphasis" rule we have eg: bold

pmario:
@twMat -- IMO this one can be closed -- The default colour has been implemented some time ago. See: https://tiddlywiki.com/#Formatting%20in%20WikiText

linonetwo:
I see, it generates `<span class="tc-inline-style">highlight</span>` where `tc-inline-style` can be moded.

PR #6877, opened 2022-08-04
add default settings for styled inline SPANs

pmario:
This PR fixes: **Add wikitext for default highlight: @@highlight@@** #2228 

- This PR adds `styled-inline-background` and `styled-inline-foreground` settings to all palettes. 
- It removes a redundant `.tc-inline-style` from a doc macro
- It adds `.tc-inline-style` CSS settings to "Vanilla base"

Docs 

I don't know if [Styles and Classes in WikiText](https://tiddlywiki.com/#Styles%20and%20Classes%20in%20WikiText) is enough documentation. ... 

If not we should create a new tiddler eg: "Styles and Classes for inline WikiText" which will be very similar to the first one.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/styleinline.js
+++ b/core/modules/parsers/wikiparser/rules/styleinline.js
@@ -45,12 +45,12 @@ exports.parse = function() {
 	};
 	if(classString) {
 		$tw.utils.addClassToParseTreeNode(node,classString);
-	} else {
-		$tw.utils.addClassToParseTreeNode(node,"tc-inline-style");
 	}
-
 	if(stylesString) {
 		$tw.utils.addAttributeToParseTreeNode(node,"style",stylesString);
+	}
+	if (!classString && !stylesString) {
+		$tw.utils.addClassToParseTreeNode(node,"tc-inline-style");
 	}
 	return [node];
 };
--- a/core/modules/parsers/wikiparser/rules/styleinline.js
+++ b/core/modules/parsers/wikiparser/rules/styleinline.js
@@ -41,14 +41,14 @@ exports.parse = function() {
 	var node = {
 		type: "element",
 		tag: "span",
-		attributes: {
-			"class": {type: "string", value: "tc-inline-style"}
-		},
 		children: tree
 	};
 	if(classString) {
 		$tw.utils.addClassToParseTreeNode(node,classString);
+	} else {
+		$tw.utils.addClassToParseTreeNode(node,"tc-inline-style");
 	}
+
 	if(stylesString) {
 		$tw.utils.addAttributeToParseTreeNode(node,"style",stylesString);
 	}
--- a/core/modules/parsers/wikiparser/rules/styleinline.js
+++ b/core/modules/parsers/wikiparser/rules/styleinline.js
@@ -49,7 +49,7 @@ exports.parse = function() {
 	if(stylesString) {
 		$tw.utils.addAttributeToParseTreeNode(node,"style",stylesString);
 	}
-	if (!classString && !stylesString) {
+	if(!classString && !stylesString) {
 		$tw.utils.addClassToParseTreeNode(node,"tc-inline-style");
 	}
 	return [node];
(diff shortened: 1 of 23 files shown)


kookma:
Hi Mario,
 Don't you think to use a consistent highlight color for different pallets?

Example: This Cupertino Dark

image

This palette has a highlight color, see below
image


I may recommend to use better color now!

Jermolene:
Thanks @pmario 

> * This PR adds `styled-inline-background` and `styled-inline-foreground` settings to all palettes.

I think it might make more sense to use semantic names: `highlight-background` and `highlight-foreground`

But the deeper problem is that this default styling must only be applied if there is no explicit styling applied after the opening `@@`. Otherwise I don't think this would be a backwards compatible change.

pmario:
>But the deeper problem is that this default styling must only be applied if there is no explicit styling applied after the opening @@. Otherwise I don't think this would be a backwards compatible change.

I did think about that too but I wasn't sure if I should implement it that way. .. So if there is a custom style the `tc-inline-style` should **not** be applied. ... I'll change that accordingly.

pmario:
>Don't you think to use a consistent highlight color for different pallets?

I did check all the palettes. ... The selection color and the highlight colour IMO should be different. In you second example the blue is a selection that can be copy / pasted. ... The yellow one is just highlighted. CTRL-insert will do nothing with it.

Jermolene:
> I did check all the palettes. ... The selection color and the highlight colour IMO should be different. In you second example the blue is a selection that can be copy / pasted. ... The yellow one is just highlighted. CTRL-insert will do nothing with it.

I think that's correct.

pmario:
@Jermolene I did address the concerns from https://github.com/Jermolene/TiddlyWiki5/pull/6877#issuecomment-1205606135 Should be good now.

core/modules/parsers/wikiparser/rules/styleinline.js
```diff
@@ -41,14 +41,14 @@ exports.parse = function() {
 	var node = {
 		type: "element",
 		tag: "span",
-		attributes: {
-			"class": {type: "string", value: "tc-inline-style"}
-		},
 		children: tree
 	};
 	if(classString) {
```
Jermolene, line 46: We should only apply "tc-inline-style" if there are no classes *or styles* assigned to the run.

pmario:
@Jermolene .. Updated code as requested and added new tests. They all pass

pmario:
bump

twMat:
bump

Jermolene:
Thanks for your patience @pmario

Jermolene:
I've reverted this PR due to failing tests. Please resubmit

pmario:
>I've reverted this PR due to failing tests. Please resubmit

@Jermolene 

I did just check out the branch an did run the tests and all of them work as expected.

see: git status, so the right TW branch is used. No tests failed for me

image

Jermolene:
Hi @pmario I wonder if it is a Windows vs. *nix thing. Here's the failing test when running under GitHub CI; I got the same errors under macOS.

https://github.com/Jermolene/TiddlyWiki5/actions/runs/3206431538/jobs/5240194686

I'm not sure if you'll have permission to see the detail:

5B98DD97-8637-4FBD-8F63-B6920FD0AD09

pmario:
I'll test again. I think I didn't merge the PR into the current master. May be did run with "old" master code.

pmario:
@Jermolene ... I did merge my PR to the actual master. Now there is an error. BUT the error doesn't come from my parser implementation, because the parse tree is right. ... The raw HTML output is wrong now. 

eg: `some @@color:green;.tc-inline-style 1 style and 1 class@@ text` gives the following pars tree, which is OK. So the code in the PR works. 

image

BUT the HTML output is wrong. So there must have been a change in the rendering mechanism, which is-not part of this PR.

image

pmario:
@Jermolene ... I did some more tests. If I check out the commit: https://github.com/Jermolene/TiddlyWiki5/commit/4e9267ea58c3081b9845dcebd96bbd25819d0143 and apply this PR it starts to fail

Jermolene:
Thanks @pmario, I've pushed a fix in 8f079e2d4566418afdfef16e594fc173423decc5 and then re-merged this PR in b531984f50cd82ebde3f167e157a840465d7a66a. All seems well so far.

PR #6881, opened 2022-08-05
Move the version number to the top of the upgrader

pmario:
This PR only moves the version number to the top area of the upgrader UI

**Start View**

image

**Import View**

image

### DIFF
--- a/plugins/tiddlywiki/upgrade/UpgradeWizard.tid
+++ b/plugins/tiddlywiki/upgrade/UpgradeWizard.tid
@@ -6,9 +6,9 @@ tags: $:/tags/AboveStory
 
 <div class="tc-upgrade-wizard">
 
-! ~TiddlyWiki Upgrade Wizard
+! ~TiddlyWiki version <<version>>
 
-!! version <<version>>
+!! Upgrade Wizard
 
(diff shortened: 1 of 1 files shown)


twMat:
Nitpick... possibly someone might interpret it to mean that it is the wizard itself that has this version number. Maybe it should explicitly state "TiddlyWiki version..." ?
As I said: nitpick.

pmario:
The wizard and the TW have that version number. So I don't see a problem here. The upgrader can only upgrade to that version number -- nothing else.

twMat:
OK, I just thought it was a bit clearer, especially for the group we turn to with this modification to begin with. But no biggie if you disagree.

pmario:
You are right. I did change it and did update the OT

twMat:
Your idea to put it as the *top* headline makes things much clearer! Good idea!

Jermolene:
Thanks @pmario I think it would be clearer if the heading "Upgrade Wizard" was the first and largest heading.

pmario:
@Jermolene Updated as requested. I did also change the images in the OP

pmario:
bump

twMat:
bump

pmario:
@Jermolene pls merge

Issue #6879, opened 2022-08-04
[IDEA] $:/Translators should directly expose the English prompts


twMat:
https://tiddlywiki.com/editions/translators/#%24%3A%2FTranslators shows a table where the translator is supposed to modify the content of several small edit fields into another language. 

**Problem**: The original English prompts are only accessible via popups activated with buttons next to the edit fields. But when translating something it is important to directly see what is being translated. It is very easy to tweak things so it quickly changes the original meaning. 

Less important, but still, it is also not optimal to have the prompt on the *side* of the editor. It is easier to gaze up and down at two typically very short strings when they are stacked vertically than horizontally.

**Describe the solution you'd like**
Keep the English prompt exposed, and either above or below the editor. Not to the side and not hidden in a button. (This would make the page longer.... but, so what?)

**Describe alternatives you've considered**
One idea is to make the prompt area foldable and provide a global toggle - expand all/none - with local state overwrite option.

Jermolene:
Hi @twMat that all makes sense. I'd welcome improvements to the translators edition, it doesn't get much attention.

pmario:
@twMat What do you think? .. I did use it with the German translation and did already find some minor issues in the translations. 

![image](https://user-images.githubusercontent.com/374655/183057690-45628ec5-221c-451b-9f53-26d775121882.png)

twMat:
Ah, good working! I think the spacing above and below the "Default text" string should be a bit smaller, but overall this makes it much simpler to translate stuff! Excellent idea to make the editors yellow so it is totally clear where to work.

Personally, I doubt anyone would need the option to use the checkbox and turn it off. I.e I would propose that your view is the default and only view and that the right hand arrow button feature is just removed. (Why would anyone want to use that?)

pmario:
@twMat This can be closed now. The PR was merged

twMat:
@pmario  - well done!

PR #6882, opened 2022-08-05
Translators edtion: Add global option to show default text below the edit textarea

pmario:
This PR fixes:  [IDEA] $:/Translators should directly expose the English prompts #6879 

image


Jermolene:
Thanks @pmario

PR #6883, opened 2022-08-05
Typo in the readme of Dynannotate plugin

oflg:
In addition, the Dynannotate plugin's Main Selection Tracker does not seem to work.

Is it because [json-ops](https://github.com/Jermolene/TiddlyWiki5/tree/json-ops) is not merged into the master branch?


Jermolene:
Thanks @oflg 



> In addition, the Dynannotate plugin's Main Selection Tracker does not seem to work.
> 
> Is it because [json-ops](https://github.com/Jermolene/TiddlyWiki5/tree/json-ops) is not merged into the master branch?

I think so. The latest version of the JSON operators are now included in the parameterised transclusion PR (#6666).

PR #6884, opened 2022-08-06
Minor adjustments to Japanese language files

fu-sen:
Since there was a mixture of "Desumasu-style"(ですます調) and "Dearu-style"(である調), the unification (that is unique to Japanese)


Jermolene:
Thanks @fu-sen much appreciated – it's great to see these improvements coming through.

PR #6887, opened 2022-08-07
Fix popup position if popup is triggered from within an offsetParent element

FlashSystems:
## Problem description

The popup mechanism does not take the `offsetParent` attribute of the element that initiates the popup into account. If the element initiating the popup and the declaration of the popup itself have different positioned ancestor elements, the popup will be shown in the wrong place. This can be easily reproduced with the following example:

```
| Popup 1 |<$button popup="$:/SamplePopupState">Pop me up!</$button>|
| Popup 2 |<$button popup="$:/SamplePopupState">Pop me up!</$button>|

<$reveal type="popup" state="$:/SamplePopupState">
<div class="tc-drop-down">
Popup
</div>
</$reveal>
```

Popup Wrong

The table cell is a positioned ancestor element (see [HTMLElement.offsetParent](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetParent)) and the popup is declared outside the table. This makes the popup show up in the upper-left corner of the tiddler, because that is the nearest positioned ancestor element. The position is measured relative to the table cell and put into the StateTiddler. But the Popup is positioned relative to the tiddler because the tiddler is the nearest positioned ancestor element. This leads to the popup to show up in the wrong place.

## What this PR does

* To fix the described issue, this pull request changes how the coordinates for the popup-mechanism are stored. Currently, they are stored relative to the nearest positioned ancestor elements. After applying the PR, the coordinates are stored relative to the body of the document and translated into the coordinate system of the popup when it's shown by the reveal widget.
* The second commit unifies the declaration of the regular expression used to parse the StateTiddler for the popup. It was declared in `popup.js` and `reveal.js`. I found this while debugging the problem and though the code would be more maintenance friendly if the regex is only declared in `popup.js` and used in `reveal.js`. If this change should be dropped or get its own PR, please let me know.

After applying the PR the popup is shown in the right place:

Popup Correct

## Possible regressions

If a plugin is relying directly on the content of the popup's StateTiddler, the new coordinate system will lead to problems.

I don't know if the content of the popup's StateTiddler is contractual or an implementation detail. If it is considdered an implementation detail and `Popup.show` is the official API, this PR would be safe to apply because the behaviour of `Popup.show` will not change in a backward incompatible way. I cannot think of any case in which the current behaviour regarding the "offsetParent" might be desired.

## Testing

I've used TiddlyWiki 5.2.2 with these commits applied for more than a month and have not noticed any problems.

### DIFF
--- a/core/modules/utils/dom/popup.js
+++ b/core/modules/utils/dom/popup.js
@@ -23,6 +23,12 @@ var Popup = function(options) {
 };
 
 /*
+Global regular expression for parsing the location of a popup.
+This is also used by the Reveal widget.
+*/
+Popup.popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/
+
+/*
 Trigger a popup open or closed. Parameters are in a hashmap:
 	title: title of the tiddler where the popup details are stored
 	domNode: dom node to which the popup will be positioned (one of domNode or domNodeRect is required)
@@ -185,9 +191,31 @@ Popup.prototype.cancel = function(level) {
 Returns true if the specified title and text identifies an active popup
 */
 Popup.prototype.readPopupState = function(text) {
-	var popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/;
-	return popupLocationRegExp.test(text);
+	return Popup.popupLocationRegExp.test(text);
 };
+
+/*
+Parses a coordinate string in the format `(x,y,w,h)` or `@(x,y,z,h)` and returns
+an object containing the position, width and height. The absolute-Mark is boolean
+value that indicates the coordinate system of the coordinates. If they start with
+an `@`, `absolute` is set and the coordinates are relative to the root element. If
+the initial `@` is missing, they are relative to the offset parent element and
+`absoute` is false.
+*/
+Popup.prototype.parseCoordinates = function(coordinates) {
+	var match = Popup.popupLocationRegExp.exec(coordinates);
+	if(match) {
+		return {
+			absolute: (match[1] === "@"),
+			left: parseFloat(match[2]),
+			top: parseFloat(match[3]),
+			width: parseFloat(match[4]),
+			height: parseFloat(match[5])
+		};
+	} else {
+		return false;
+	}
+}
 
 exports.Popup = Popup;
 
--- a/core/modules/utils/dom/popup.js
+++ b/core/modules/utils/dom/popup.js
@@ -29,6 +29,13 @@ This is also used by the Reveal widget.
 Popup.popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/
 
 /*
+Objekt containing the available prefixes for coordinates build with the `buildCoordinates` function:
+ - csOffsetParent: Uses a coordinate system based on the offset parent (no prefix).
+ - csAbsolute: Use an absolute coordinate system (prefix "@").
+*/
+Popup.prototype.coordinatePrefix = { csOffsetParent: "", csAbsolute: "@" }
+
+/*
 Trigger a popup open or closed. Parameters are in a hashmap:
 	title: title of the tiddler where the popup details are stored
 	domNode: dom node to which the popup will be positioned (one of domNode or domNodeRect is required)
@@ -152,8 +159,7 @@ Popup.prototype.show = function(options) {
 			currentNode = currentNode.offsetParent;
 		}
 	}
-	var popupRect = (options.absolute ? "@(" : "(") + rect.left + "," + rect.top + "," +
-				rect.width + "," + rect.height + ")";
+	var popupRect = $tw.popup.buildCoordinates(options.absolute?$tw.popup.coordinatePrefix.csAbsolute:$tw.popup.coordinatePrefix.csOffsetParent,rect);
 	if(options.noStateReference) {
 		options.wiki.setText(options.title,"text",undefined,popupRect);
 	} else {
@@ -214,6 +220,23 @@ Popup.prototype.parseCoordinates = function(coordinates) {
 		};
 	} else {
 		return false;
+	}
+}
+
+/*
+Builds a coordinate string from a coordinate system identifier and an object
+containing the left, top, width and height values.
+Use constants defined in the coordinatePrefix property to specify a coordinate
+system.
+If one of the parameters is invalid for building a coordinate string `(0,0,0,0)`
+will be returned.
+*/
+Popup.prototype.buildCoordinates = function(prefix,position) {
+	var coord = prefix + "(" + position.left + "," + position.top + "," + position.width + "," + position.height + ")";
+	if (Popup.popupLocationRegExp.test(coord)) {
+		return coord;
+	} else {
+		return "(0,0,0,0)";
 	}
 }
 
--- a/core/modules/utils/dom/popup.js
+++ b/core/modules/utils/dom/popup.js
@@ -136,7 +136,17 @@ Popup.prototype.show = function(options) {
 			height: options.domNode.offsetHeight
 		};
 	}
-	var popupRect = "(" + rect.left + "," + rect.top + "," + 
+	if(options.absolute && options.domNode) {
+		// Walk the offsetParent chain and add the position of the offsetParents to make
+		// the position absolute to the root node of the page.
+		var currentNode = options.domNode.offsetParent;
+		while(currentNode) {
+			rect.left += currentNode.offsetLeft;
+			rect.top += currentNode.offsetTop;
+			currentNode = currentNode.offsetParent;
+		}
+	}
+	var popupRect = (options.absolute ? "@(" : "(") + rect.left + "," + rect.top + "," +
 				rect.width + "," + rect.height + ")";
 	if(options.noStateReference) {
 		options.wiki.setText(options.title,"text",undefined,popupRect);
@@ -175,7 +185,7 @@ Popup.prototype.cancel = function(level) {
 Returns true if the specified title and text identifies an active popup
 */
 Popup.prototype.readPopupState = function(text) {
-	var popupLocationRegExp = /^\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/;
+	var popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/;
 	return popupLocationRegExp.test(text);
 };
 
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -94,6 +94,13 @@ RevealWidget.prototype.positionPopup = function(domNode) {
 		left = Math.max(0,left);
 		top = Math.max(0,top);
 	}
+	if (this.popup.absolute) {
+		// Traverse the offsetParent chain and correct the offset to make it relative to the parent node.
+		for (var offsetParentDomNode = domNode.offsetParent; offsetParentDomNode; offsetParentDomNode = offsetParentDomNode.offsetParent) {
+			left -= offsetParentDomNode.offsetLeft;
+			top -= offsetParentDomNode.offsetTop;
+		}
+	}
 	domNode.style.left = left + "px";
 	domNode.style.top = top + "px";
 };
@@ -183,7 +190,7 @@ RevealWidget.prototype.compareStateText = function(state) {
 };
 
 RevealWidget.prototype.readPopupState = function(state) {
-	var popupLocationRegExp = /^\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/,
+	var popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/,
 		match = popupLocationRegExp.exec(state);
 	// Check if the state matches the location regexp
 	if(match) {
@@ -191,10 +198,11 @@ RevealWidget.prototype.readPopupState = function(state) {
 		this.isOpen = true;
 		// Get the location
 		this.popup = {
-			left: parseFloat(match[1]),
-			top: parseFloat(match[2]),
-			width: parseFloat(match[3]),
-			height: parseFloat(match[4])
+			absolute: (match[1] === "@"),
+			left: parseFloat(match[2]),
+			top: parseFloat(match[3]),
+			width: parseFloat(match[4]),
+			height: parseFloat(match[5])
 		};
 	} else {
 		// If not, we're closed
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -190,20 +190,11 @@ RevealWidget.prototype.compareStateText = function(state) {
 };
 
 RevealWidget.prototype.readPopupState = function(state) {
-	var popupLocationRegExp = /^(@?)\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/,
-		match = popupLocationRegExp.exec(state);
+	this.popup = $tw.popup.parseCoordinates(state);
 	// Check if the state matches the location regexp
-	if(match) {
+	if(this.popup) {
 		// If so, we're open
 		this.isOpen = true;
-		// Get the location
-		this.popup = {
-			absolute: (match[1] === "@"),
-			left: parseFloat(match[2]),
-			top: parseFloat(match[3]),
-			width: parseFloat(match[4]),
-			height: parseFloat(match[5])
-		};
 	} else {
 		// If not, we're closed
 		this.isOpen = false;
--- a/editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
+++ b/editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
@@ -25,7 +25,7 @@ The relative coordinate system works flawless most of the time. Problems occure 
 
 Absolute coordinates can fix this problem by using the root element of the page (the upper-left corner of the page) as the origin of the coordinate system. Absolute coordinates are expressed in the form ''@(x,y,w,h)''. Where ''x'' and ''y'' represent the position and ''w'' and ''h'' the width and height of the element. The leading ''@''-symbol marks these coordinates as absolute.
 
-The ButtonWidget has an option (''popupAbsCoords'') to put absolute coordinates into the state tiddler.
+The ButtonWidget has an option (''popupAbsCoords'') to put absolute coordinates into the state tiddler. The DraggableWidget and the EventCatcherWidget provide the absolute coordinate of an event within the attribute `tv-popup-abs-coords`.
 
 
 !! Example
--- a/editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
+++ b/editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
@@ -1,0 +1,42 @@
+created: 20220810201659784
+modified: 20220810201659784
+tags: Concepts
+title: Coordinate Systems
+type: text/vnd.tiddlywiki
+
+TiddlyWiki (primarily the RevealWidget) supports two coordinate systems for positioning popups (see PopupMechanism to learn more about popups).
+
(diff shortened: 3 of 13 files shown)


saqimtiaz:
> To fix the described issue, this pull request changes how the coordinates for the popup-mechanism are stored.

I don't think we can make this without breaking backwards compatibility. $action-popup and $eventcatcher rely on the coordinates format, as might other third party widgets.

FlashSystems:
Hello @saqimtiaz,
thank you for your response. As I wrote: If the format (or better phrased the origin of the coordinate system) for the popups is contractual, there is no way for this pull request to get accepted without breaking existing code :(

I see the problem with `$action-popup`: The user can specify the coordinates of the popup. But the documentation does not define the origin of these coordinates. Currently, they are relative to the offset parent of the `$reveal` widget. So for `$action-popup` the origin of the coordinate system changes depending on where the popup was defined. If it was defined within a tiddler, the origin is the div with the `.tc-tiddler-frame` class because this class sets the CSS property `position: relative` making it an offset parent.

For the example within the pull request, the problem is hat the button and the popup have different offset parents and therefore can not agree on the same coordinate system.

Maybe we should first try to clarify the correct coordinate system for the popup mechanism. Then I can rework this PR to let the button use it and fix the problem described.

PS: For the `$eventcatcher` widget, I did not find any reference to the popup mechanism. Could you please give me a hint?

saqimtiaz:
What we need to be careful about is that existing usage of $action-popup should not break or change as a result of any changes.

$eventcatcher provides a variable with coordinates that can be used with $action-popup.

Jermolene:
Hi @FlashSystems it would be great to fix popups within tables, it's something that consistently trips up users.

As @saqimtiaz notes, we do have a problem with existing wikis that create/manipulate coordinate strings. Perhaps we need to add a new attribute to the popup-related widgets to explicitly select the coordinate system in use.

FlashSystems:
I've implemented a new approach to this problem. The RevealWidget now understands two coordinate formats:

| Coordinates | Description |
|---|---|
| `(x,y,w,h)` | The classic format, that is relative to the offset parent of the element. |
| `@(x,y,w,h)` | The absolute format. This format is relative to the root element of the page. |

The ButtonWidget has a new parameter to switch to the absolute coordinates: `popupAbsCoords`. If this is set to "yes", absolute coordinates will be used and the popup will display at the correct location, even when the button is within a table. The ActionPopupWidget also understands both coordinate systems.

@saqimtiaz, @Jermolene: Before I update the documentation and give this some more days of testing, please give me feedback if you see any problems with this approach.

## Example

I've created a [demo wiki](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) for you to experiment with.

## Risk of regressions

The risk of regressions is quite low because using the new coordinates is opt-in. If the user is experiencing problems with an extension, it can simply remove the `popupAbsCoords` switch from the ButtonWidget and everything is as before (and will not work in tables 😉). New extensions can use this coordinate system as long as they make sure that the version of TW supports it.

core/modules/widgets/reveal.js
```diff
@@ -183,18 +193,18 @@ RevealWidget.prototype.compareStateText = function(state) {
 };
 
 RevealWidget.prototype.readPopupState = function(state) {
-	var popupLocationRegExp = /^\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/,
-		match = popupLocationRegExp.exec(state);
+	var match = $tw.utils.Popup.popupLocationRegExp.exec(state);
```
pmario, line 196: While I think the regexp definition should be done only once. I'm not so happy with the implementation. 

`$tw.utils.Popup.popupLocationRegExp.exec()` defines the "matching groups" and returns a match. 
but
the `if (match)` ... `match[1-2-3]` are done here ... So there is a hardcoded dependency between 2 files now. ... 

IMO the `$tw.utils.Popup` should also return an object that contains the `absolute, left, top ... ` values. So if changes to the group matches need to be made, they can be done in 1 file only. 

reveal.js and action-popup.js should call the functions and get the results they need, without the `parseFloat(match[x])` code

FlashSystems: You are right. My approach was incomplete. I've reworked the commit and put a `parseCoordinates` method into `popup.js`.

FlashSystems:
Hello @saqimtiaz, @pmario, @Jermolene,

thank you for your feedback. I've reworked this pull request and changed the following things:

- Reworked the commit for unifying the declaration of the RegEx for parsing the popup-position. There is now a `parseCoordinates` method that is used for parsing the coordinates and it returns a JS object containing `x` , `y`, `w`, `h` and the absolute flag.
- Added documentation for the coordinate systems and the formats

core/modules/utils/dom/popup.js
```diff
@@ -136,7 +142,20 @@ Popup.prototype.show = function(options) {
...
 		};
 	}
-	var popupRect = "(" + rect.left + "," + rect.top + "," + 
+
+	if (options.absolute && options.domNode) {
+		// Walk the offsetParent chain and add the position of the offsetParents to make
+		// the position absolute to the root node of the page.
+		var currentNode=options.domNode.offsetParent;
+		while (currentNode)
+		{
```
pmario, line 151: the opening brace belongs to the line above

core/modules/utils/dom/popup.js
```diff
@@ -175,10 +194,32 @@ Popup.prototype.cancel = function(level) {
...
+value that indicates the coordinate system of the coordinates. If they start with
+an `@`, `absolute` is set and the coordinates are relative to the root element. If
+the initial `@` is missing, they are relative to the offset parent element and
+`absoute` is false.
+*/
+Popup.prototype.parseCoordinates = function(coordinates) {
+	var match = Popup.popupLocationRegExp.exec(coordinates);
+	if(match) {
+		return {
+			absolute: (match[1].length > 0),
```
pmario, line 212: The first group is `(@?)` .. so it returns an @ or nothing. .. We should check for @, which will allow us to introduce other "symbols" in the future. Just in case we need them. ... `> 0` is not specific enough.

editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
```diff
@@ -0,0 +1,42 @@
+created: 20220810201659784
+modified: 20220810201659784
+tags: Concepts
+title: Coordinate Systems
+type: text/vnd.tiddlywiki
+
+TiddlyWiki (primarily the RevealWidget) supports two coordinate systems for positioning popups (see PopupMechanism to learn more about popups).
+
+Beware that absolute coordinates were added with TiddlyWiki 5.2.4 and that they may not work with all extensions and plugins. For maximum backwards compatibility, use absolute coordinates only where necessary.
```
pmario, line 9: We have a `.from-version` macro to introduce info, that works from version xxx
eg:

>`<<.from-version "5.2.4">>` We introduced absolute coordinates that work with all extensions and plugins. For maximum backwards compatibility, use absolute coordinates only where necessary.

editions/tw5.com/tiddlers/widgets/ButtonWidget.tid
```diff
@@ -37,6 +37,7 @@ The content of the `<$button>` widget is displayed within the button.
 |default |Default value if <<.attr set>> tiddler is missing for testing against <<.attr setTo>> to determine <<.attr selectedClass>> |
 |popup |Title of a state tiddler for a popup that is toggled when the button is clicked. See PopupMechanism for details |
 |popupTitle |Title of a state tiddler for a popup that is toggled when the button is clicked. In difference to the <<.attr popup>> attribute, ''no'' TextReference is used. See PopupMechanism for details |
+|popupAbsCoords |<<.from-version "5.1.24">> If set to ''yes'' writes absolute coordinates to the tiddler referenced by the <<.attr popup>>. If set to ''no'' (the default) uses relative coordinates. See [[Coordinate Systems]] for details |
```
pmario, line 40: it will be v5.2.4

pmario (commented):
IMO it looks much cleaner now. There seems to be a nice "separation of concerns" between the different functions

FlashSystems:
Hello @pmario,

I've fixed all review comments and rebased the code on the HEAD of the master branch.

core/modules/utils/dom/popup.js
```diff
@@ -175,10 +193,32 @@ Popup.prototype.cancel = function(level) {
...
+value that indicates the coordinate system of the coordinates. If they start with
+an `@`, `absolute` is set and the coordinates are relative to the root element. If
+the initial `@` is missing, they are relative to the offset parent element and
+`absoute` is false.
+*/
+Popup.prototype.parseCoordinates = function(coordinates) {
+	var match = Popup.popupLocationRegExp.exec(coordinates);
+	if(match) {
+		return {
+			absolute: (match[1] == "@"),
```
pmario, line 211: sorry for nagging again. We usually use tripple `=` for exact matches ...  `absolute: (match[1] === "@"),` ... I did make that mistake too in one of my last PRs ;)

FlashSystems: Hello  @pmario,

thank you for your patience, and sorry for not having checked which comparison operator is normally used in the code base. I've pushed an update.

core/modules/utils/dom/popup.js
```diff
@@ -136,7 +142,19 @@ Popup.prototype.show = function(options) {
 			height: options.domNode.offsetHeight
 		};
 	}
-	var popupRect = "(" + rect.left + "," + rect.top + "," + 
+
+	if (options.absolute && options.domNode) {
+		// Walk the offsetParent chain and add the position of the offsetParents to make
```
Jermolene, line 147: Is there any scope for making a global utility function to aid with creating popup coordinate strings?

core/modules/widgets/reveal.js
```diff
@@ -90,10 +90,20 @@ RevealWidget.prototype.positionPopup = function(domNode) {
 			top = this.popup.top + this.popup.height;
 			break;
 	}
+
```
Jermolene, line 93: Very minor point, but we try to avoid blank lines except for separating top level methods (there are two more below too).

Jermolene (commented):
Thanks @FlashSystems. This is a very worthwhile change, and I appreciate your care and attention to detail with the implementation. A couple of minor review points.

FlashSystems:
I've created a new iteration of this PR with the following changes:

- Removed superfluous newlines
- Introduced a `buildCooridnates` function in `popup.js`
- Migrated the locations where coordinate strings are created to `buildCoordinates`
- Added test cases for `parseCoordinates` and `buildCoordinates`

Regarding the migration to `buildCoordinates` I need some help. I modified `core/modules/utils/dom/dom.js` and `plugins/tiddlywiki/dynannotate/modules/dynannotate.js` to use the new function, but I don't know how to test these code paths. Could anyone please tell how to exercise these code paths or test by themselves if they still work?

I've updated the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) as well.

core/modules/utils/dom/popup.js
FlashSystems: The newest update to this PR contains a commit that implements a function for that. I like the symmetry between `buildCoordinates` and `parseCoordinates` 👍 Please have a look at my last comment for more details.

joshuafontany:
@FlashSystems Do you think this PR will fix the bug encountered by @theSherwood in Mentat that required a "Repopup plugin"?
```
Repopup

A quick hack of a plugin that attempts to reposition tiddlywiki popups so that they don't run off the screen. It's not perfect and it doesn't seem to do much to normal dropdowns. But it's an okay fix until the popup mechanism can get a comprehensive overhaul. This plugin has been designed with the Volant, Engine, and Mentat Plugins in mind. So if you use it for something else, your mileage may vary.
```

https://github.com/theSherwood/Mentat/blob/master/src/tiddlers/system/plugins/admls/repopup/startup/repopup.js

FlashSystems:
> @FlashSystems Do you think this PR will fix the bug encountered by @theSherwood in Mentat that required a "Repopup plugin"?
> 
> ```
> Repopup
> 
> A quick hack of a plugin that attempts to reposition tiddlywiki popups so that they don't run off the screen. It's not perfect and it doesn't seem to do much to normal dropdowns. But it's an okay fix until the popup mechanism can get a comprehensive overhaul. This plugin has been designed with the Volant, Engine, and Mentat Plugins in mind. So if you use it for something else, your mileage may vary.
> ```
> 
> https://github.com/theSherwood/Mentat/blob/master/src/tiddlers/system/plugins/admls/repopup/startup/repopup.js

Hello @joshuafontany,

I'm not using Repopup or Mentat. So take everything below with a grain of salt.

I've had a look at the plugin. I don't think this PR will make the Repopup-Plugin obsolete, because this PR does not make sure, that the popup stays on screen. It's still possible for the popup to hang off the screen. Maybe the plugin can be simplified by using the new absolute coordinates. As far as I understood the code, it works around the offset parent problem as well.

The new coordinate format is extensible. Maybe it would be worthwhile to extend it with an option to force the specified rectangle "on screen". But I think this should be a separate PR because I think that some discussion will be necessary.

joshuafontany:
@FlashSystems Thats the sense I got from your changes, it & how @theSherwood approached it were both working around the problem with different goals in mind. Perfect, I look forward to building on these changes.

FlashSystems:
Hello @Jermolene, @saqimtiaz,

i think I've addressed all concerns/suggestions in the last update and this pull request is ready for final review.

FlashSystems:
Pushing this up again for final review and merge.

Jermolene:
Thanks for your patience @FlashSystems – @saqimtiaz could you kindly have a look at this? I think it's good to go but would appreciate an extra set of eyes, and I believe you've done work in this area.

editions/tw5.com/tiddlers/widgets/ActionPopupWidget.tid
```diff
@@ -15,7 +15,7 @@ The ''action-popup'' widget is invisible. Any content within it is ignored.
 
 |!Attribute |!Description |
 |$state |The title of the state tiddler for the popup |
-|$coords |Optional coordinates for the handle to which popup is positioned (in the format `(x,y,w,h)`) |
+|$coords |Optional coordinates for the handle to which popup is positioned (see [[Coordinate Systems]] for the supported formats) |
 |$floating |<<.from-version "5.2.0">> Optional. Defaults to `no`. Set to `yes` to create a popup that must be closed explicitly. |
 
 <<.from-version "5.1.23">> If the ''$coords'' attribute is missing or empty then all popups are cancelled.
```
saqimtiaz, line 21: We should add a note here that from version 5.2.4 the absolute coordinate system is supported.

saqimtiaz:
@FlashSystems could you kindly explain how one would use the new coordinate system in actions triggered by $eventcatcher? If this is not currently possible, we should extend the variables provided by the widget with an additional variable for the new coordinate system.

editions/tw5.com/tiddlers/concepts/CoordinateSystems.tid
```diff
@@ -0,0 +1,42 @@
+created: 20220810201659784
+modified: 20220810201659784
+tags: Concepts
+title: Coordinate Systems
```
saqimtiaz, line 4: I wonder if it would be helpful to rename this tiddler to Popup Coordinate System. It should also be linked to from the tiddler "PopupMechanism".

FlashSystems: Hello @saqimtiaz, thank you for taking the time to review my PR.
Are the coordinates really limited to be used with popups? I would prefer a more generic name for the documentation because these coordinates could be used for other things as well. Currently they are used for popups but should we suggest in the documentation, that they are limited to this use case in the future?

FlashSystems:
> @FlashSystems could you kindly explain how one would use the new coordinate system in actions triggered by $eventcatcher? If this is not currently possible, we should extend the variables provided by the widget with an additional variable for the new coordinate system.

I did not realize how useful a `tc-popup-abs-coords` attribute would be when I updated the EventCatcherWidget to use `$tw.popup.buildCoordinates`. Thank you for pointing this out. I'll implement the attribute and update this PR.

FlashSystems:
Thank you, @saqimtiaz, for your review. I've updated this pull request to address your comments.

The following changes were made:
- Fixed some coding style issues.
- Added `tv-popup-abs-coords` to `collectDOMVariables` to enable this attribute in the `DraggableWidget` and the `EventCatcherWidget`.
- Updated the documentation:
  - Documented the new `tv-popup-abs-coords` attribute for the  `DraggableWidget` and the `EventCatcherWidget`.
  - Linked the "Coordinate Systems" tiddler from the "PopupMechanism" tiddler.
  - Added a note to the "PopupMechanism" tiddler that absolute coordinates are supported since version 5.2.4.
- Rebased on the master branch.

I've updated the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) as well to ease testing.

saqimtiaz:
@FlashSystems thank you, looks good to me.

pmario:
@FlashSystems ... Please have a look at: https://github.com/Jermolene/TiddlyWiki5/issues/7003

If you comment out `tc-block-dropdown-wrapper` as shown below in vanilla/base.tid .. Does it make a difference for you??

```
.tc-block-dropdown-wrapper {
<!--	position: relative; -->
}
```

FlashSystems:
I've updated this PR one more time to clean up some code in `dom.js`. No functional change, just converting a while-loop to a for-loop to be more concise.

Thank you, @saqimtiaz, for your review.

@pmario: I've looked at the proposed change in #7003 and found no interaction between this PR and the removal of `position: relative` from `.tc-block-dropdown-wrapper`. The popups created by `reveal.js` get the style `position: absolute` assigned in `positionPopup` regardless of the CSS.

I've included the change from PR #7003 in the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) for this PR, so everyone can test independently if these the two PRs have unwanted consequences.

I think this should be good to go now. If anyone wants to give this some more testing, feel free to try out the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) and report any findings here.

Jermolene:
Great, thanks @FlashSystems, @saqimtiaz, @pmario

Jermolene:
I'm going to revert this commit because although it passes tests, it fails when attempting to build the site. Here's the log:

```
$ tiddlywiki editions/tw5.com --verbose --build
Boot log:
  Startup task: load-modules
  Startup task: info after: load-modules before: startup
  Startup task: plugins after: load-modules
  Startup task: startup after: load-modules
  Startup task: story after: startup
  Startup task: commands platforms: node after: story
Executing command: build 
Executing command: rendertiddler $:/core/save/offline-external-js index.html text/plain
Executing command: render $:/core/templates/tiddlywiki5.js [[tiddlywikicore-]addsuffix<version>addsuffix[.js]] text/plain
Rendering "$:/core/templates/tiddlywiki5.js" to "/Users/jeremyruston/git/Jermolene/TiddlyWiki5/editions/tw5.com/output/tiddlywikicore-5.2.4-prerelease.js"
Executing command: rendertiddler $:/core/templates/static.template.html static.html text/plain
$:/core/modules/widgets/reveal.js:193
        this.popup = $tw.popup.parseCoordinates(state);
                               ^

TypeError: Cannot read properties of undefined (reading 'parseCoordinates')
    at RevealWidget.readPopupState ($:/core/modules/widgets/reveal.js:193:25)
    at RevealWidget.readState ($:/core/modules/widgets/reveal.js:165:9)
    at RevealWidget.execute ($:/core/modules/widgets/reveal.js:132:7)
    at RevealWidget.render ($:/core/modules/widgets/reveal.js:32:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:475:15)
    at SetWidget.render ($:/core/modules/widgets/set.js:33:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:475:15)
    at TranscludeWidget.render ($:/core/modules/widgets/transclude.js:33:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:475:15)
    at SetWidget.render ($:/core/modules/widgets/set.js:33:7)

Node.js v18.4.0
```

FlashSystems:
Hello @Jermolene, sorry for having caused a problem. I've tested the site using `tiddlywiki editions/tw5.com --verbose --build index` but haven't tested the static build. :disappointed:

The problem is, that the Popup class is not initialized in `startup.js` if `$tw.browser` is not true. After having consolidated the facilities for parsing coordinate strings into `popup.js` this breaks because the static build needs to parse coordinate stings even if no Popup module is initialized.

A solution (besides duplicating the code as it was before) would be to move the functions for handling coordinates (`readPopupState`, `parseCoordinates`, `buildCoordinates`) from `popup.js` into `utils.js`. That way they are accessible even if the Popup module is not initialized.

@Jermolene, @saqimtiaz, do you think this would be an acceptable solution?

Jermolene:
> @Jermolene, @saqimtiaz, do you think this would be an acceptable solution?

Yes – although the other interesting possibility might be to make them be static methods of the popup class.

Jermolene:
> Hello @Jermolene, sorry for having caused a problem. 

@FlashSystems I meant to add that please don't worry, absolutely not a problem. Mistakes are inevitable, and the goal isn't for us to be individually perfect, but rather for the overall process to be effective.

FlashSystems:
I've fixed the bug that led to the crash when building a static site and created a new PR #7013.

Issue #6886, opened 2022-08-06
[BUG] Nameless field can't be deleted


twMat:
### Describe the bug

I happened to create a nameless custom field. This can't be deleted via the field delete button.

### To Reproduce

1. Create a new tiddler so you're in edit view
2. ~~Type `<$edit-text field={{!!currentTiddler}} />` and loudly say *"Yes, this could happen to anyone at 1 am"*~~<br>(update) Type `<$edit-text field='' />` (i.e just a missing argument)
3. Save the tiddler
4. Back in view mode, enter some value in the little editor field
5. Open the editor and see the nameless field with the entered value
6. Attempt to delete the field by clicking the delete button

### TiddlyWiki Configuration

- Version [e.g. v5.1.24]
- Saving mechanism [e.g. Node.js, TiddlyDesktop, TiddlyHost etc]
- Plugins installed [e.g. Freelinks, TiddlyMap]

kookma:
Confirmed (TW 5.2.3 + Chrome 102).
This is a bug!
It makes no sense to set value to nameless field!

Jermolene:
This is an interesting one! The requirement for a field name is that it be a string of characters; a valid string is a zero length string. So, there's nothing in the design of TiddlyWiki at the moment to prevent creating fields with a zero length name; indeed, there's nothing inconsistent or irregular about them. Arguably, the issue is just that the UI doesn't permit these fields to be deleted.

This gets tricky because in the TW core JS it is not possible for a write to a tiddler to fail. That's super important because it avoids having to have error checking everywhere that we change the values of tiddlers. So, it's not clear what we should do when faced with HS code that creates such a field:

```js
wiki.addTiddler({title: "HelloThere, "": "Zero length field"})
```

Given that there's no mechanism for that call to return with an error, the question would be what we do with the zero length field name. We could change the field name to something else, or we could entirely ignore that field. But then the problem is that code would observe a different tiddler value than that which it had set.

So, overall, I think it would actually be a good deal easier to just fix the UI so that these fields can be handled as expected.

pmario:
>So, overall, I think it would actually be a good deal easier to just fix the UI so that these fields can be handled as expected.

The problem is that action-deletefield doesn't allow us to enter an empty `$field=""` parameter at the moment. So we'll need to change that widget.

pmario:
@Jermolene 

>So, overall, I think it would actually be a good deal easier to just fix the UI so that these fields can be handled as expected.

[...]

PR #6888, opened 2022-08-08
allow us to remove a field with empty key

pmario:
This PR is related to **[BUG] Nameless field can't be deleted #6886**

It will fix the UI related problem. A nameless field can be removed after this PR is merged.

### DIFF
--- a/core/modules/widgets/action-deletefield.js
+++ b/core/modules/widgets/action-deletefield.js
@@ -59,7 +59,7 @@ DeleteFieldWidget.prototype.invokeAction = function(triggeringWidget,event) {
 		tiddler = this.wiki.getTiddler(self.actionTiddler),
 		removeFields = {},
 		hasChanged = false;
-	if((this.actionField != null) && tiddler) {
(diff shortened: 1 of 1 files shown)


core/modules/widgets/action-deletefield.js
```diff
@@ -59,7 +59,7 @@ DeleteFieldWidget.prototype.invokeAction = function(triggeringWidget,event) {
 		tiddler = this.wiki.getTiddler(self.actionTiddler),
 		removeFields = {},
 		hasChanged = false;
-	if(this.actionField && tiddler) {
```
Jermolene, line 62: We prefer !== unless the != operator is specifically required, in which case we should add a comment explaining why.

Jermolene (commented):
Thanks @pmario

core/modules/widgets/action-deletefield.js
pmario: uuups

pmario:
@Jermolene .. should be OK now

Jermolene:
Thank you @pmario

PR #6895, opened 2022-08-10
Update chinese language files

BramChen:
* Change core GettingStarted tiddler table layout
* Revise wording


Jermolene:
Thanks @BramChen

Issue #6871, opened 2022-08-03
[BUG] Internal JavaScript Error : TypeError (..) when setting the checkbox widget attribute listField to a "default" field (list, tags,..)


Telumire:
Use this wikitext to reproduce : 

```
* <$checkbox listField="list" checked="A"/>
* <$checkbox listField="list" checked="B"/>
* <$checkbox listField="list" checked="C"/>
```
Error message when checking more than one of the checkboxes : 

"`TypeError: can't define array index property past the end of an array with non-writable length`"

An error also occur when using a list widget like this : 

```
<$list filter="[is[current]fields[]]" variable="field">
 <$checkbox listField=<<field>> checked="A"/>
</$list>
```
The error message this time is "`TypeError: can't access property "indexOf", list is null`"

### TiddlyWiki Configuration

- Tested on v5.2.3 and pre-release 5.2.4

pmario:
I can confirm this problem

yaisog:
See also https://talk.tiddlywiki.org/t/new-checkbox-listfield-bug-or-user-error/4229

rmunn:
The cause of this problem is that when I implemented the `listField` feature, I tried to be efficient by saying "if the field is already stored as a list internally, just modify it in place". But TiddlyWiki's internal tiddler objects are "frozen" against being modified in-place. And my tests didn't catch that because most fields are stored as text internally, but the `list` field is, for obvious reasons, stored as a list internally.

I'm gotten a fix written for the bug (which involves making a copy of the list, modifying the copy, and then telling TW to please update the `list` field with this new value), and now I'm working on adding to the tests for the checkbox widget so that it will test the built-in "list" field in addition to any custom fields.

Until the bugfix is released, you can work around this issue by using custom fields, rather than the built-in `list` field, with the `listField` property.

rmunn:
[...]

PR #6897, opened 2022-08-11
Fix bug when using built-in `list` field as listField parameter to checkbox widget

rmunn:
This fixes the `TypeError: can't define array index property past the end of an array with non-writable length` error from #6871. The cause of this problem is that when I implemented the `listField` feature, I tried to be efficient by saying "if the field is already stored as a list internally, just modify it in place". But TiddlyWiki's internal tiddler objects are "frozen" against being modified in-place. And my tests didn't catch that because most fields are stored as text internally, but the `list` field is, for obvious reasons, stored as a list internally.

This PR also adds regression tests so that the bug won't happen again. The bug would have manifested with the `tags` field as well, but since the checkbox widget already has a `tag` input property, most people aren't going to use `listField="tags"` so they won't run into that version of the bug. Still, I added regression tests to test both the `list` and `tags` fields, just to make sure that possibility was covered as well.

### DIFF
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -212,6 +212,10 @@ CheckboxWidget.prototype.handleChangeEvent = function(event) {
 		} else {
 			listContents = $tw.utils.parseStringArray(this.wiki.extractTiddlerDataItem(this.checkboxTitle,this.checkboxListIndex) || "") || [];
 		}
+		if(Object.isFrozen(listContents)) {
+			// Can happen if using, say, the built-in "list" field which is frozen against modifications
+			listContents = listContents.slice(0);
+		}
 		oldPos = notValue ? listContents.indexOf(notValue) : -1;
 		newPos = value ? listContents.indexOf(value) : -1;
 		if(oldPos === -1 && newPos !== -1) {
--- a/editions/test/tiddlers/tests/test-checkbox-widget.js
+++ b/editions/test/tiddlers/tests/test-checkbox-widget.js
@@ -234,6 +234,38 @@ Tests the checkbox widget thoroughly.
             },
         ];
 
+        // https://github.com/Jermolene/TiddlyWiki5/issues/6871
+        const listModeTestsWithListField = (
+            listModeTests
+            .filter(data => data.widgetText.includes("listField='colors'"))
+            .map(data => {
+                const newData = {
+                    ...data,
+                    tiddlers: data.tiddlers.map(tiddler => ({...tiddler, list: tiddler.colors, colors: undefined})),
+                    widgetText: data.widgetText.replace("listField='colors'", "listField='list'"),
+                    expectedChange: {
+                        "Colors": { list: data.expectedChange.Colors.colors.split(' ') }
+                    },
+                }
+                return newData;
+            })
+        );
+        const listModeTestsWithTagsField = (
+            listModeTests
+            .filter(data => data.widgetText.includes("listField='colors'"))
+            .map(data => {
+                const newData = {
+                    ...data,
+                    tiddlers: data.tiddlers.map(tiddler => ({...tiddler, tags: tiddler.colors, colors: undefined})),
+                    widgetText: data.widgetText.replace("listField='colors'", "listField='tags'"),
+                    expectedChange: {
+                        "Colors": { tags: data.expectedChange.Colors.colors.split(' ') }
+                    },
+                }
+                return newData;
+            })
+        );
+
         const indexListModeTests = listModeTests.map(data => {
             const newData = {...data};
             const newName = data.testName.replace('list mode', 'index list mode');
@@ -453,6 +485,8 @@ Tests the checkbox widget thoroughly.
         const checkboxTestData = fieldModeTests.concat(
             indexModeTests,
             listModeTests,
+            listModeTestsWithListField,
+            listModeTestsWithTagsField,
             indexListModeTests,
             filterModeTests,
         );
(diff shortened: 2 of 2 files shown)


rmunn:
Note that #6871 should not yet be closed, as I have not yet fixed the second half of that issue report, where you could get an error saying "can't access property `indexOf`, list is null". The cause of that one is trying to use a built-in field like `created` (stored as a date) as a list field. The fix for that one is less obvious, because I need to decide what kind of error (if any) to show if the user does that.

editions/test/tiddlers/tests/test-checkbox-widget.js
```diff
@@ -234,6 +234,38 @@ Tests the checkbox widget thoroughly.
             },
         ];
 
+        // https://github.com/Jermolene/TiddlyWiki5/issues/6871
+        const listModeTestsWithListField = (
```
pmario, line 238: @Jermolene ... Should we allow const, arrow functions and ... rest in tests. .. **I'd say yes**, because our test equipment should be new enough ;)

rmunn: I've actually already used `const` and arrow functions in the rest of the test-checkbox-widget.js code from #6561. So if that worked, then this will also work. If `const` and arrow functions in the tests are going to cause problems, the problems would probably have shown up already. :-)

marad:
Hi! I've stumblied on this issue today. When is it going to be merged?

core/modules/widgets/checkbox.js
```diff
@@ -212,6 +212,10 @@ CheckboxWidget.prototype.handleChangeEvent = function(event) {
 		} else {
 			listContents = $tw.utils.parseStringArray(this.wiki.extractTiddlerDataItem(this.checkboxTitle,this.checkboxListIndex) || "") || [];
 		}
+		if(Object.isFrozen(listContents)) {
```
Jermolene, line 215: I wonder if this is the best fix.

Back in line 211, in the codepath for `if(this.checkboxListField)` being true, the value is retrieved with `tiddler.getFieldList()`, which always returns an immutable value. Meanwhile, in the other codepath, `wiki.extractTiddlerDataItem()` also reuses results from earlier calls thanks to the caching mechanism (although there's no freezing, so the immutability is not enforced).

As a result, we cannot safely mutate either of the returned lists, so it would be better to do the cloning without the `isFrozen()` test.

editions/test/tiddlers/tests/test-checkbox-widget.js
Jermolene: I think we should avoid new-fangled JS in the test suite because it runs in the browser, and we want to be able to run our tests in the oldest browsers that we support.

Jermolene (commented):
Thanks @rmunn

rmunn:
For some reason (don't know why) the GitHub interface won't let me reply to @Jermolene's comment in https://github.com/Jermolene/TiddlyWiki5/pull/6897#discussion_r948011899_ that reads:

> I think we should avoid new-fangled JS in the test suite because it runs in the browser, and we want to be able to run our tests in the oldest browsers that we support.

So I'm replying here. @Jermolene, do you want me to remove the `const` and arrow-function code that has already been released in 5.1.23? Because I wasn't aware of that requirement when I wrote #6561, so I included some uses of `const` and arrow-function definitions. I thought that the unit tests only run in Node environments, and that therefore I didn't need to be careful about IE support in the `editions/test` directory tree. (I was careful not including any newfangled JS in the main code, of course).

core/modules/widgets/checkbox.js
rmunn: If I clone the list every time, then in any usage scenario where a custom list is used (which I expect to be most of the time), the clone will be unnecessary since parseStringArray creates a new list. I think a better solution will be to extract the field without calling `getFieldList`, then examine the field I get. If it's a list, clone it before modifying. If it's a string, use `parseStringArray` and then don't clone the results as we know that that's a brand-new list.

I'll implement that sometime soon (not right now, as I'm about to run out the door for an appointment).

editions/test/tiddlers/tests/test-checkbox-widget.js
rmunn: So do you want me to remove the `const` and arrow-function code that has already been released in 5.1.23? Because I wasn't aware of that requirement when I wrote https://github.com/Jermolene/TiddlyWiki5/pull/6561, so I included some uses of `const` and arrow-function definitions. I thought that the unit tests only run in Node environments, and that therefore I didn't need to be careful about IE support in the `editions/test` directory tree. (I was careful not including any newfangled JS in the main code, of course).

core/modules/widgets/checkbox.js
Jermolene: Thanks for your patience @rmunn – your proposed solution makes sense, and I think would be preferable.

editions/test/tiddlers/tests/test-checkbox-widget.js
Jermolene: I think it would be worth removing them in a separate PR, if you can. Much appreciated thank you

core/modules/widgets/checkbox.js
rmunn: Sorry it took a week for me to get to this; I've been traveling without good Internet access until now. I've implemented that suggestion (just make a defensive copy if the field was already a list, whether or not it was frozen). I also fixed another bug mentioned in #6871, where a non-existent tiddler would result in `tiddler is null` in this same piece of logic.

There may be one more check I should make, which is to see if the field contains a data type (such as a date) that should not be turned into a list. (See https://github.com/Jermolene/TiddlyWiki5/issues/6871#issuecomment-1211578823 for discussion of this issue). I'm running out of time to implement that today, so I'll do that tomorrow. Then the PR will be ready for review again.

Jermolene:
Thanks @rmunn and apologies for the delay.

Issue #6926, opened 2022-08-21
[IDEA] Proposal to extend the palette editor with an "element" preview field


pmario:
Since the palette editor has some more space on the right side. The idea is to add a new table field, that shows a preview of the element, that can can be defined in the editor. 

There is a very very basic mockup, which 

- shows the $:/core/ui/AlertTemplate 
- with some fake data
- and a border radius of 4 px to actually see something

![image](https://user-images.githubusercontent.com/374655/185797191-a94f813b-1bb9-46dd-ade4-2a52767d300f.png)

Palettes are named: `$:/palettes/xxxx` tagged: `$:/tags/Palette`

They are data-tiddlers. It should be possible to use a second data-tiddler with the same keys and the value could be an "example tiddler" eg: $:/examples/palettes

```
alert-background: $:/example/palettes/alert
alert-border: $:/example/palettes/alert
alert-highlight: $:/example/palettes/alert
alert-muted-foreground: $:/example/palettes/alert
```

It seems to be possible to use a prefix ... So in the future it may be possible to work with the palette tiddler to dynamically create the example-tiddler-names. .. But the first idea is a second tiddler with all the keys and values 

What do you think?

There is no code yet, but working with the palettes lately I saw there are a lot of problems in the existing palettes, because users don't know what they are defining.

PR #6928, opened 2022-08-24
make palette manager readable

pmario:
@Jermolene .. This PR should make the $:/PaletteManager more readable to be able to add info for palette creators as mentioned at: ** [IDEA] Proposal to extend the palette editor with an "element" preview field** #6926 

**It adds whitespace only** 

Tested with FF latest, Edge on Windows and FF Android

It would be nice if it could be merged soon, since I'd need it to extend the PaletteManger as described

### DIFF
--- a/core/ui/PaletteManager.tid
+++ b/core/ui/PaletteManager.tid
@@ -13,51 +13,53 @@ title: $:/PaletteManager
 \import $:/core/macros/utils
 \whitespace trim
 <$wikify name="name" text="""$macrocall$""">
-<<name>>
+   <<name>>
 </$wikify>
 \end
 
 \define delete-colour-index-actions() <$action-setfield $index=<<colourName>>/>
+
 \define palette-manager-colour-row-segment()
 \whitespace trim
 <$edit-text index=<<colourName>> tag="input" placeholder=<<edit-colour-placeholder>> default=""/>
 <br>
 <$edit-text index=<<colourName>> type="color" tag="input" class="tc-palette-manager-colour-input"/>
 <$list filter="[<currentTiddler>getindex<colourName>removeprefix[<<]removesuffix[>>]] [<currentTiddler>getindex<colourName>removeprefix[<$]removesuffix[/>]]" variable="ignore">
-<$set name="state" value={{{ [[$:/state/palettemanager/]addsuffix<currentTiddler>addsuffix[/]addsuffix<colourName>] }}}>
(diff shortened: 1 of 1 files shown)


pmario:
I started a discussion at: https://talk.tiddlywiki.org/t/palette-manager-teaser/4385

grafik

pmario:
@Jermolene ... Pls merge or at least comment

Jermolene:
Hi @pmario this is an impressive piece of work; I've always wanted to have meaningful previews in the palette editor.

However, I am concerned the result is quite a lot more complex for new users. I wonder if it might improve things to split each section up into a separate vertical tab, for example. I'm tempted to suggest making it available as a plugin for the moment, but equally I would really like to see this in the core. Perhaps the best thing might be to try to do both...

pmario:
The current version of the plugin is described at: https://talk.tiddlywiki.org/t/palette-manager-feature-complete-preview/5095

Yes. I think I'll make a docs edition and implement it as a plugin for the time being. ... There is still some code cleaning needed ... There may be some new texts for translation ... and so on. 

All in all it already used about 10 times the amount of time I wanted to spend with it and it still isn't finished. :/ ... 

But on the way I did try to create a new colour-widget, that is capable to resolve "colour-macros" and should be able to handle all types of macros in the future. ... We'll see.

Jermolene:
> Yes. I think I'll make a docs edition and implement it as a plugin for the time being. ... There is still some code cleaning needed ... There may be some new texts for translation ... and so on.

Great, I think it is already very powerful, and will be helpful for a lot of users.

> All in all it already used about 10 times the amount of time I wanted to spend with it and it still isn't finished. :/ ...

Isn't it always the way? But I'm sure much useful learning as well.

> But on the way I did try to create a new colour-widget, that is capable to resolve "colour-macros" and should be able to handle all types of macros in the future. ... We'll see.

I'm still keen to integrate a colour library that can parse CSS colours and do interpolation and colours space conversion – I think we have a ticket for that?

pmario:
>Isn't it always the way?

It seems so ..

>But I'm sure much useful learning as well.

Absolutely :) ... Especially working with the cascade filter. 

>I'm still keen to integrate a colour library that can parse CSS colours and do interpolation and colours space conversion – I think we have a ticket for that?

Yes we do have a pending issue. ... That's also a reason to keep it as a plugin atm. There may be some changes needed to the colour-widget and the colour-picker-resolved macro.

pmario:
@Jermolene ... It would still be nice to merge this PR. It only adds code indentation, which may make future PRs easier to read.

pmario:
@Jermolene ... Bump. It only contains code indentation for readability.

PR #6932, opened 2022-08-30
Polymorphic Filter Operators

Jermolene:
# Introduction

Filter operators in TiddlyWiki work with lists of string items. This PR extends the internal implementation to allow data items in filters to be of any type, not just strings. It is planned to go further, and support variable values of any type too.

 These changes are an implementation optimisation that is intended to be invisible to end users. They can still conceptualise the filters as working with strings; the fact that some intermediate values may be represented as other data types is hidden by the automatic conversion to and from string format.

The motivation is to improve the efficiency of the implementation of the JSON filter operators. The current implementation can be seen in #6936. It is based on passing intermediate JSON results as strings between operators, with consequent frequent conversion between string and JSON data. This PR seeks to improve matters by allowing intermediate filter results to be handled as any convenient data type, avoiding most of those conversions.

For example, with the changes in this PR, the maths operators return numbers instead of strings. But if the result of a maths operator is passed to an operator that only works with strings then the result is automatically converted from a number to a string.

This PR introduces a new module type to be used for filter operators that can handle polymorphic data. Existing filter operators with the existing module type will continue to see data in string form, with automatic conversion if necessary.

# New JSON Operators

* **jsontypecheck** – verify that a filter item is of a particular type

# Other Changes

* All maths operators are now polymorphic, and return numeric results instead of strings
* Count operator is now polymorphic, returning the result as a number instead of a string

# Implementation Details

The effect of these changes is that the value of items within filters can be of any type supported by JavaScript:

* Strings
* Numbers
* Booleans
* Dates
* JSON-compatible objects and arrays (ie no circular references)

Polymorphic filter operators are responsible for coercing the values they receive to the desired type. Utility functions are provided to help with the conversion. Polymorphic operators return results as a list of items of any supported type, and are not restricted to returning strings.

An important feature of those conversions is that operators that expect JSON-compatible objects will process incoming string values by attempting to parse them as JSON.

# Progress

- [ ] Fix performance regressions
    - [ ] There is an impact on the performance of existing string-based filters because there is now the burden of looping through each intermediate filter result and checking whether it is a string or not. We could perhaps avoid that by assuming that the output of an old-school string-based filter operator will always be a string. So, along with the accumulator we need to track whether or not the accumulator is (or rather could be) polymorphic, and do our conversions accordingly
    - [ ] Improve simple-list class for efficiency, and perhaps to handle removal of non-string items
- [ ] Replace the module type "newfilteroperator" with "polymorphicfilteroperator" or perhaps "filteroperatorpolymorphic" is better so that the modules sort together
- [ ] Update filter run prefixes to only see string values and introduce new type of filter run prefix module that can handle polymorphic values
- [ ] Update format filter operator to be polymorphic; add support for "polymorphicformatfilteroperator" module type that accepts polymorphic values, and move format:json to the new model
- [ ] "filter" operator should be polymorphic
- [ ] Polymorphic version of filterTiddlers should return an indicator of whether the results are polymorphic vs. strings.


Jermolene:
Abandoning because of the performance impact.

PR #6934, opened 2022-08-31
[fr-FR] Fix tag pills style

tw-FRed:
In French tiddlywiki.com edition, tag pills style is skewed. This fix reverts old changes to core tiddler allowing titles translation, and provides a new solution.


Jermolene:
Thanks @tw-FRed

PR #6935, opened 2022-09-01
Update chinese translations

BramChen:
* Add Hint and Caption of the delete button in the AdvancedSearch filter results
* Add confirm messages for the above deleting action


Jermolene:
Thanks @BramChen

Issue #4194, opened 2019-08-23
Why does Jasmine run eagerly?


flibbles:
I'm trying to finalize the demo site for my own plugin, and I have to keep Jasmine separated as its own site because of how eagerly it runs tests.

### Problem

If you have a tiddlywiki Node.JS project with Jasmine added as a plugin, and you use the command line to render tiddlers, run a server, check the version, _or even look at the help menu_, all tests run.

If you have a tiddlywiki website with Jasmine as a plugin, you can't open the page without all the tests running. If your tests take a while, it means your browser will take a while to load the page, which isn't savory.

### Solution?

Why not instead make Jasmine less eager? If on command line, there could be a `tiddlywiki --test` command. If in the browser, why not have a widget that renders as a button, that runs the tests when clicked? This way, the tiddlywiki page loads just fine, and any delays from testing occur after hitting the button (which could now show progress output, maybe as dots) instead of just stalling.

I realize this introduces one extra click into running browser tests, but I really think it would be preferable.

If people agree, I could probably make the outlines of this myself.

Jermolene:
Hi @flibbles good question. The current behaviour is probably because I use a script that opens the test.html in my default browser, and I find it quite handy that I can see the results immediately. We could perhaps make it configurable. Now that we have startup actions we could make the tests be triggered by a message defined in a startup module, and then the `test` edition could include a startup action for auto-running the tests for backwards compatibiltiy.

flibbles:
Sure, @Jermolene, that sounds like it could work, but before I start poking, is this something worth doing? Between this and your response in #3970, I get the impression that, before me, you were the only person who used Jasmine. If that's the case, it should work the way _you_ want it to work.

Jermolene:
> Sure, @Jermolene, that sounds like it could work, but before I start poking, is this something worth doing? Between this and your response in #3970, I get the impression that, before me, you were the only person who used Jasmine. If that's the case, it should work the way you want it to work.

I've no objection to a PR that makes the behaviour configurable while retaining the current behaviour for the "test" edition.

flibbles:
I'd just like to confirm that this would still be desirable. This eager running has always been a thorn to me.

1. I can't debug with Chrome, because the test framework runs too soon, and the debug js doesn't load in time.
2. I can't run a server from Node.js if jasmine is installed, because jasmine hard quits TiddlyWiki once it's done.
3. I can't do *anything* from Node.js without running every single test.
[...]

Issue #6845, opened 2022-07-28
[Proposal] It is proposed to implement UI tests using: https://playwright.dev/


pmario:
This is a reminder that: It is proposed to implement UI tests using: https://playwright.dev/

From: https://github.com/Jermolene/TiddlyWiki5/pull/6756#issuecomment-1176481173

>>By the way, we should probably switch to Playwright; it's built by some of the same people as Puppeteer, but works with a wider variety of browser engines, including Firefox.
>>
>>    https://playwright.dev/
>
>That's interesting. I did have a look at the release history. It seems the got from beta 0.10.0 to v1.0.0 in just 3 months. .. That's impressive. Had a look at an [intro-video](https://www.youtube.com/watch?v=dxbh0bl4It8&t=1s) from last month. ... The code generator and the test-debugging tools look impressive too. ... It actually looks like fun to create tests.
>
>Coming back to the SVG test we have here. .. I can see how the visual tests can be implemented and since it seems to have all the DOM structure info, we are able to check that. ... Good
>
>So it will complement the existing jasmine tests, and we should probably replace those tests. which need to use the fakeDOM at the moment ... right?

https://github.com/Jermolene/TiddlyWiki5/pull/6756 should be re-visited for the first experiments.

Jermolene:
> So it will complement the existing jasmine tests, and we should probably replace those tests. which need to use the fakeDOM at the moment ... right?

I suspect we should run all the tests under both Node.js and the browser for maximum coverage.

rmunn:
I've had some experience with Playwright. The one drawback to Playwright is that it'll be significantly slower than the Jasmine tests we have already. That's because Playwright loads a fresh browser instance and a fresh copy of your page/site for every test, which is good from a test-isolation standpoint. But that means every Playwright test will load a fresh TiddlyWiki instance, and have to wait several seconds for TW to start up, before it can run its test. Each Jasmine test runs in milliseconds, but each Playwright test is going to run in seconds, and the total runtime for the tests is going to be much longer.

I definitely like Playwright — it's been a very good experience to use so far — but there is also a place for simpler tests using frameworks like Jasmine that don't need to start up a browser every time.

[...]

PR #6944, opened 2022-09-13
Jasmine command

flibbles:
This is the answer to issue #4194. It introduces the "--test" command for executing jasmine on command line rather than having it run automatically.

It does **not** change the behavior of browser-jasmine. I know we talked about making an event to run it, but here's the thing. Jasmine really *doesn't* want to be run more than once. And while I was able to get it so jasmine didn't run until you clicked a "<$button/>", it would always fall apart if you pressed the button twice. So for now, I'm only introducing the node.js side.

A few decisions I made.

* I made the command "--test", which might be too generic?
* It takes a second argument, which is a regular expression to specify specific tests or suites. This is the jasmine way of narrowing testing scope, not the *Tiddlywiki* way, which would be to use a filter. I can do either one (or both), but I implemented the jasmine testSpec filter because...
  1. It's how it works in the browser (Like when you click on failed tests, and it refreshes to run only that test.)
  2. You couldn't run individual tests using a Tiddlywiki filter. You could only specify which tiddlers run.
* Other than a lot of added documentation, there isn't currently a way to indicate to upgrading developers that they now need to add "--test" to their tiddlywiki.info files. I couldn't think of a way to do this that didn't introduce obnoxious problems.

If there are any ways you'd like me to do it differently, let me know. I'm happy to tweak whatever.


plugins/tiddlywiki/jasmine/help.tid
```diff
@@ -0,0 +1,18 @@
+title: $:/language/Help/test
+description: Run the jasmine test suite
+
+This runs all of the tests in tiddlers tagged with `$:/tags/test-spec`.
+
+```
+--test [<specFilter>]
+```
+
+The "specFilter" is an optional argument to run only specific suites or tests. It's a regular expression which will be matched against the full name of every test to see whether to run it. The full name of a test is its assigned name prepended with the names of all containing suites.
```
pmario, line 10: It would be nice, to have an example, which only runs some of existing tests.

Jermolene:
Thanks @flibbles 

> This is the answer to issue #4194. It introduces the "--test" command for executing jasmine on command line rather than having it run automatically.

I am concerned about backwards compatibility for users runnings the test suite their own CI. Perhaps we could automatically run the tests if the jasmine plugin is loaded but no commands have been specified? I appreciate that may not be straightforward as things stand.

> It does **not** change the behavior of browser-jasmine. I know we talked about making an event to run it, but here's the thing. Jasmine really _doesn't_ want to be run more than once. And while I was able to get it so jasmine didn't run until you clicked a "<$button/>", it would always fall apart if you pressed the button twice. So for now, I'm only introducing the node.js side.

That makes sense.

> * I made the command "--test", which might be too generic?

That's a good point. For instance, I've considered adding support for end user wikitext tests (so that users can write tests for their own macros etc), and we might well want to use the `--test` command for such a thing.

Perhaps `--runtest` or `--runjasmine`?

> * It takes a second argument, which is a regular expression to specify specific tests or suites. This is the jasmine way of narrowing testing scope, not the _Tiddlywiki_ way, which would be to use a filter. I can do either one (or both), but I implemented the jasmine testSpec filter because...

A regexp makes sense here, I agree.

> * Other than a lot of added documentation, there isn't currently a way to indicate to upgrading developers that they now need to add "--test" to their tiddlywiki.info files. I couldn't think of a way to do this that didn't introduce obnoxious problems.

 Indeed, see my suggestion above.

One other thought is that it might be worth using the named arguments we introduced for the listen command to give us a bit of a backwards compatibility cushion.

flibbles:
@pmario - Done. More documentation added.

@Jermolene - Done.

* I changed it from "--test" to "--jasmine". Seems succinct.
* The command takes a named parameter. (i.e. "--jasmine spec="^Framework")
* You suggested I have the tests run automatically if no commands are given. I did something different. Instead, if someone has jasmine installed, and they run TW on command line without running tests, they get a notice:
  `Jasmine: no "--jasmine" command given, so skipping tests`
  Here's why:
  1. Running tiddlywiki without commands vomits up the entire "usage" message, so I don't think anyone is doing this. Instead, anyone who runs tests regularly has probably created a build configuration rather than always running `tiddlywiki`.
  1. I think  most people using jasmine just copied the "run tests and make test file" command set that you use in the "test" edition. (That's why I did.) Which means they would not benefit from automatically running tests unless they were designed to automatically run if the "--jasmine" command isn't called. And that would defeat the whole point of making it a command.
  3. Instead, a notice is printed. Yes, you'll see this notice whenever you use a jasmine-installed tiddlywiki project to do anything besides run tests, but it's just a single line. Maybe in five years, once everyone has had a chance to migrate, it can be removed.

pmario:
>I changed it from "--test" to "--jasmine". Seems succinct.

If we may change the testsuite in the future, the name may be different. I'd prefer `--test use="jasmine"` or something similar. ... IMO `--test` is more future proof

Also see: https://github.com/Jermolene/TiddlyWiki5/issues/6845 ... which would use an additional framework

flibbles:
You raise some good points, @pmario.

@Jermolene, I can implement this any way we settle on, just let me know.

flibbles:
Well, I've thought about it more. I would say either `--test` or `--jasmine`.

`--test use=jasmine` is tricky, because it implies multiple plugins all having testing frameworks, (like `use=macro`, or `use=mocha`). But the question is, if the `--test` command isn't part of the core, which plugin is introducing it? (And I don't think it should be part of core.)

`--test` is probably fine and what everyone would be familiar with. Other js testing frameworks use `npm test` mostly. Also, it seems unlikely that someone would have two javascript testing plugins installed at once. (Why have both jasmine and mocha?) And a testing framework for wikitext macros doesn't actually sound like something end-users would test on command-line.

Maybe I'm wrong. If we're still concern about collision, `--jasmine` would also work, because I'm guessing nearly every node.js developer will have a tiddlywiki.file build task called "test" which would run stuff like `--jasmine --run-macro-tests --rendertiddler`, so they'll be running `tiddlywiki --build test` anyway, regardless of what we pick.

flibbles:
I've switched back to `--test` in expectation that we may settle on that, but that doesn't mean I'm hard set on `--test`. I'm totally open to discuss it more. It's easy to change again. I'm just trying to line this up for a merge.

Jermolene:
Thanks @flibbles I am still concerned about the backwards compatibility impact of changing the way that the tests are run, but I accept that there isn't a good alternative at the moment. The message is a reasonable mitigation.

I've come around to using `--test` as the command name.

flibbles:
That's understandable. The PR is here if/when you're ready.

I, for one, will be using this PR on my personal fork since I am long since tired of having all the tests run when I so much as run `tiddlywiki --help`.

Jermolene:

PR #6947, opened 2022-09-14
log message not to have spaces and <empty string>

flibbles:
This removes the empty space before and after all log messages.

More importantly, it removes the "<empty string>" that firefox likes to put before and after every log message.


Jermolene:
Thanks @flibbles

PR #6969, opened 2022-09-30
Update Styles and Classes in WikiText.tid

twMat:
Add tag to make this tiddler included in the list in the tiddler

https://tiddlywiki.com/prerelease/#How%20to%20apply%20custom%20styles


Jermolene:
Thanks @twMat

PR #6973, opened 2022-10-04
Use deprecated macro for deprecated commands

hoelzro:
This just brings the documentation for these commands more inline with other deprecations, plus offers a link to the recommended alternative as well as the explanation for deprecation

### DIFF
--- a/editions/tw5.com/tiddlers/commands/RenderTiddlerCommand.tid
+++ b/editions/tw5.com/tiddlers/commands/RenderTiddlerCommand.tid
@@ -2,4 +2,6 @@ title: RenderTiddlerCommand
 tags: Commands
 caption: rendertiddler
 
+<<.deprecated-since "5.1.15" "RenderCommand">>.
+
 {{$:/language/Help/rendertiddler}}
--- a/editions/tw5.com/tiddlers/commands/RenderTiddlersCommand.tid
+++ b/editions/tw5.com/tiddlers/commands/RenderTiddlersCommand.tid
@@ -2,4 +2,6 @@ title: RenderTiddlersCommand
 tags: Commands
 caption: rendertiddlers
 
+<<.deprecated-since "5.1.15" "RenderCommand">>.
+
 {{$:/language/Help/rendertiddlers}}
--- a/editions/tw5.com/tiddlers/commands/SaveTiddlerCommand.tid
+++ b/editions/tw5.com/tiddlers/commands/SaveTiddlerCommand.tid
@@ -4,4 +4,6 @@ created: 20131218121606089
 modified: 20131218121606089
 caption: savetiddler
 
+<<.deprecated-since "5.1.15" "SaveCommand">>.
+
 {{$:/language/Help/savetiddler}}
--- a/editions/tw5.com/tiddlers/commands/SaveTiddlersCommand.tid
+++ b/editions/tw5.com/tiddlers/commands/SaveTiddlersCommand.tid
@@ -4,4 +4,6 @@ created: 20140609121606089
 modified: 20140609121606089
 caption: savetiddlers
 
+<<.deprecated-since "5.1.15" "SaveCommand">>.
+
 {{$:/language/Help/savetiddlers}}


Jermolene:
Thank you @hoelzro – unfortunately, those `$:/core/language/en-GB/Help/*` tiddlers are part of the core, which doesn't have the **deprecated-since** macro. But in editions/tw5.com we have wrapper tiddlers to make them show up regular searches; we could perhaps add the deprecated macros there? We'd still need to keep the existing deprecation text because those language strings are used by the `--help` command under Node.js.

hoelzro:
@Jermolene Do you mean `editions/tw5.com/tiddlers/commands/RenderTiddlerCommand.tid` and others?

core/language/en-GB/Help/rendertiddler.tid
```diff
@@ -1,7 +1,7 @@
 title: $:/language/Help/rendertiddler
 description: Render an individual tiddler as a specified ContentType
 
-(Note: The `--rendertiddler` command is deprecated in favour of the new, more flexible `--render` command)
+<<.deprecated-since "5.1.15" "RenderCommand">>.
```
pmario, line 4: IMO the text is wrong. It's the `--rendertiddler` command that's deprecated and not the "RenderCommand". ... Also the link to the new preferred command is missing now.

I think this problem is valid for all the changes made

Jermolene: @pmario the point is that we can't use that macro within core language strings. Therefore if we want to use it we must put it in the non-system, non-shadow wrapper tiddlers we have for these help tiddlers.

pmario: yea. That's the second problem. .. I just wanted to point out, that the text is wrong too.

Jermolene: @pmario which text is wrong?

pmario: The new text would be wrong too

hoelzro: @pmario I thought that the second argument to `deprecated-since` is the title of the tiddler denoting the replacement for the deprecated feature, rather than the deprecated feature's title?

hoelzro: > Also the link to the new preferred command is missing now.

hoelzro:
@Jermolene I moved the `deprecated-since` calls to the wrapper tiddlers in the tw5.com edition - let me know if that works for you!

pmario:
@hoelzro See: https://tiddlywiki.com/#How%20to%20apply%20custom%20styles%20by%20tag .. The second parameter for the  deprecated macro is intended to point to a tiddler that explains the deprecation. 

Either this new tiddler should then contain some info, what should be used instead. OR it should be the next sentence following the deprecation macro as shown in the link. 

As a user I would be confused, if the red button would link to the replacement. .. Because I do expect more info about the deprecation.

Jermolene:
Hi @pmario I may be misunderstanding, but it looks as though the PR is correct, with the expected parameters in the right order.

pmario:
>Hi @pmario I may be misunderstanding, but it looks as though the PR is correct, with the expected parameters in the right order.

The parameters are in the right order  but If I click the red deprecated button as a user I do expect to be linked to some info about the deprecated feature. 

At the moment the links point to the new functions, which I think is confusing. see: `<<.deprecated-since "5.1.15" "RenderCommand">>.` is part of the editions/tw5.com/tiddlers/commands/RenderTiddlerCommand.tid tiddler which is deprecated. 

But the link goes to to the [RenderCommand](https://tiddlywiki.com/#RenderCommand) which is the new one. So from my point of view the red button links to the wrong topic.

hoelzro:
@pmario Right now, what I see when I build the tw5.com edition and open the result, is "DEPRECATED SINCE 5.1.15", which links to an explaination of deprecation in general, and then "Use RenderCommand instead", which links to the `RenderCommand` tiddler.  If I understand you correctly, you'd like for me to add either a) some additional explanation to the `RenderCommand` tiddler about how it replaces `RenderTiddlerCommmand` and `RenderTiddlersCommand`, or b) add some additional explaination to `RenderTiddlerCommand` after the call to `deprecated-since` - is that correct?

Either way, what kind of additional detail did you have in mind?

pmario:
>@pmario Right now, what I see when I build the tw5.com edition and open the result, is "DEPRECATED SINCE 5.1.15", which links to an explaination of deprecation in general, and then "Use RenderCommand instead", which links to the RenderCommand tiddler. 

You are right. Your links are correct. ... Forget about my comments. .. **Sorry for causing confusion**

Jermolene:
Thanks @hoelzro @pmario

PR #6974, opened 2022-10-05
Add VScode user settings directory to git ignore

EvidentlyCube:
Like the title says, this PR just adds the VSCode user settings directory to `.gitignore`. It's not a directory that would ever want to be committed and it's slightly annoying to always see the files from it in my unstaged changes, hence I'd love it to be added to git ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .c9/
+.vscode/
 tmp/
 output/
 node_modules/


Jermolene:
Great, thank you @EvidentlyCube – I'm keen to do more to smooth integration with tooling like VSCode and npm.

Plloi:
>.> Personally I don't think IDE specific stuff belongs in a project git ignore that's not IDE specific  as git supports global patterns
https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

EvidentlyCube:
@Plloi Are there any arguments for that though? The gist just says it's a bad practice without saying why, and I can totally imagine scenarios where for some personal projects I'd want to track `.vscode` (I actually did so in the past with IntelliJ).

Plloi:
@EvidentlyCube It's a poor practice because my ide or editor choice should have nothing to do with the project. if you're going to exclude files like that where do you draw the line? if i exclusivly use vim, should i add vim's tmp/backup files to the gitignore?

 personal projects also suggests a different scope than a project shared with the world, your house your rules. When something is open the unneeded cruft is just more to maintain. 

Additionally in the case of not blacklist ide stuff, vscode has a mechanism for suggesting extensions that are related/helpful when working on a project.  It goes in a json file in the .vscode folder. so a full block of the folder isn't the best choice either.

Plloi:
@Jermolene You might like a read through of the workspace recommended extensions section here: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

EvidentlyCube:
> @EvidentlyCube It's a poor practice because my ide or editor choice should have nothing to do with the project.
Okay, but **why?** What's the actual consequence that will make you regret doing this if you do so? There aren't that many so popular IDEs to worry that it'll grow into hundreds of lines (and even then, unless it risks masking real files, so what?) Because right now it sounds like purity for the sake of purity with actual negative consequences:
 * Being more difficult to new developers
 * Being unfriendly to developers who for some reason can't or won't change their global `.gitignore`

Contrast that with the consequences of keeping the same information in repo's `.gitignore` which are... slightly larger file. 

> personal projects also suggests a different scope than a project shared with the world, your house your rules. When something is open the unneeded cruft is just more to maintain.
Sure, but if I already have that file ignored globally now I need to update my repo's local `.gitignore` to explicitly allow that path. Which wouldn't be a big deal except:
1. I have to know I can do this
2. I have to remember I am actually ignoring that path, which most likely is going to be "no" and I'll waste an hour or two trying to figure out why I can't commit that file, because it's such a rare occurrence I won't remember it.

> Additionally in the case of not blacklist ide stuff, vscode has a mechanism for suggesting extensions that are related/helpful when working on a project. It goes in a json file in the .vscode folder. so a full block of the folder isn't the best choice either.
In which case you can explicitly allow that one path.

But going back to this:
> if you're going to exclude files like that where do you draw the line?
Arbitrarily at a point, where I think the pros outweigh the cons. It just so happens that I only see cons of moving IDE-specific files to the global `.gitignore`. I'd be happy to include vim's tmp/backup directories if I had a maintainer/contributor who used it.

pmario:
>> @Plloi wrote
>>
>>if you're going to exclude files like that where do you draw the line?
>
>@EvidentlyCube wrote:
>
>Arbitrarily at a point, where I think the pros outweigh the cons. It just so happens that I only see cons of moving IDE-specific files to the global .gitignore. I'd be happy to include vim's tmp/backup directories if I had a maintainer/contributor who used it.

I also think there would be no problem to include the `tmp/backup` directory if it makes live easier for VIM contributors. 

There is a good argument about recommended extensions quoted below. But I still would like to exclude the .vscode directory from the repo, since I want to have my own settings and don't want to be forced to use someone else's settings. 

>[Workspace recommended extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions)
>
>A good set of extensions can make working with a particular workspace or programming language more productive and you'd often like to share this list with your team or colleagues. You can create a recommended list of extensions for a workspace with the Extensions: Configure Recommended Extensions (Workspace Folder) command.

Plloi:
> @EvidentlyCube Wrote:
I am simply genuinely confused about what the actual reasons are to prefer global .gitignore because from the way I look at it

I don't prefer global gitignore for this. I'm simply offering as an alternative, to what I see as polluting the repo, to address an annoyance made by your choice of dev environment:
> It's not a directory that would ever want to be committed and it's slightly annoying to always see the files from it in my unstaged changes

This statement is wrong because the are things that might make a project maintainer want to include the directory

> @pmario Wrote:
since I want to have my own settings and don't want to be forced to use someone else's settings.

Your settings shouldn't be in the project .vscode folder anyway, they should be in your user directory.
If the project wants to treat vscode as a supported aspect, then project settings should be limited to settings the align with the projects style guide, extension recommendations, and task runners. Assuming that's the case, .vscode shouldn't be ignored.

To be clear, my argument isn't for or against vscode files in the  repo, I just think adding the .vscode folder to  project .gitignore offers little to no benefit to a project. As a developer, I think it's your responsibility to keep unrelated files generated by your  choice of development tools out of the repo.

PR #6976, opened 2022-10-06
Extend page template with filter assigned classes

saqimtiaz:
This PR introduces the new system tags `$:/tags/ClassFilters/PageTemplate` and `$:/tags/ClassFilters/TiddlerTemplate`  to allow dynamically assigning CSS classes to the page template container DIV and the tiddler template container DIV respectively.

The assignment of classes is changed to:
* avoid the use of text substitution via a macro
* introduces the new system tags `$:/tags/ClassFilters/PageTemplate` and `$:/tags/ClassFilters/TiddlerTemplate`
  * the text field of all tiddlers with these tags are evaluated as filters and their output concatenated into the classes assigned.

Example use cases for adding/removing classes:
* when a drag is active, set a state tiddler which in turn sets a class on the page container div
* when an error is being displayed to the user and the class assigned can thus be removed when the error message is dismissed
* when a modal or lightbox is shown

### DIFF
--- a/editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_PageTemplate.tid
+++ b/editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_PageTemplate.tid
@@ -1,0 +1,9 @@
+caption: $:/tags/ClassFilters/PageTemplate
+created: 20221020035315795
+description: marks filters evaluated to dynamically add classes to the page template.
+modified: 20221020035945262
+tags: SystemTags
+title: SystemTag: $:/tags/ClassFilters/PageTemplate
+type: text/vnd.tiddlywiki
+
+The [[system tag|SystemTags]] `$:/tags/ClassFilters/PageTemplate` marks filters marks filters evaluated to dynamically add their output as CSS classes to the page template.
--- a/editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_TiddlerTemplate.tid
+++ b/editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_TiddlerTemplate.tid
@@ -1,0 +1,9 @@
+caption: $:/tags/ClassFilters/TiddlerTemplate
+created: 20221020035738692
+description: marks filters evaluated to dynamically add classes to the page template.
+modified: 20221020035933363
+tags: SystemTags
+title: SystemTag: $:/tags/ClassFilters/TiddlerTemplate
+type: text/vnd.tiddlywiki
+
+The [[system tag|SystemTags]] `$:/tags/ClassFilters/TiddlerTemplate` marks filters marks filters evaluated to dynamically add their output as CSS classes to the tiddler template.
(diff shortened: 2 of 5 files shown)


pmario:
@saqimtiaz 

>This PR is meant to facilitate discussion around extending the page template so that classes can be dynamically assigned to the tc-page-container div.

Is this PR meant to be a new discussion or does it have some connection to the current discussion about the PageTemplates at TW Talk? see: https://talk.tiddlywiki.org/t/how-can-i-overwrite-the-basic-theme-vanilla/4745

saqimtiaz:
@pmario I am hoping to start a new discussion here and the topic is distinct from the one at Talk.

The use case here is being able to dynamically add and remove classes from the page container div.

Example use cases for adding/removing classes:
- when a drag is active, set a state tiddler which in turn sets a class on the page container div
- when an error is being displayed to the user and the class assigned can thus be removed when the error message is dismissed
- when a modal or lightbox is shown

Jermolene:
Great thanks @saqimtiaz. I like the technique and we should consider extending it to classes on the view template etc.

saqimtiaz:
Any suggestions for the names of the system tags for specifying classes on the page container and the view template wrapper DOM elements respectively?

`$:/tags/PageTemplateClassFilters` and `$:/tags/TiddlerTemplateClassFilters` is what I have come up with so far.

saqimtiaz:
I have updated this PR to extend the same functionality to tiddler templates and have added documentation for the new system tags.

Jermolene:
I think this is good to go @saqimtiaz?

editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_PageTemplate.tid
```diff
@@ -0,0 +1,9 @@
+caption: $:/tags/ClassFilters/PageTemplate
+created: 20221020035315795
+description: marks filters evaluated to dynamically add classes to the page template.
+modified: 20221020035945262
+tags: SystemTags
+title: SystemTag: $:/tags/ClassFilters/PageTemplate
+type: text/vnd.tiddlywiki
+
+The [[system tag|SystemTags]] `$:/tags/ClassFilters/PageTemplate` marks filters marks filters evaluated to dynamically add their output as CSS classes to the page template.
```
twMat, line 9: "...marks filters marks filters..."

editions/tw5.com/tiddlers/systemtags/SystemTag_ $__tags_ClassFilters_TiddlerTemplate.tid
```diff
@@ -0,0 +1,9 @@
+caption: $:/tags/ClassFilters/TiddlerTemplate
+created: 20221020035738692
+description: marks filters evaluated to dynamically add classes to the page template.
+modified: 20221020035933363
+tags: SystemTags
+title: SystemTag: $:/tags/ClassFilters/TiddlerTemplate
+type: text/vnd.tiddlywiki
+
+The [[system tag|SystemTags]] `$:/tags/ClassFilters/TiddlerTemplate` marks filters marks filters evaluated to dynamically add their output as CSS classes to the tiddler template.
```
twMat, line 9: "...marks filters marks filters..."

Jermolene: Thanks @twMat – fixed in b097d2ec48105687028e4ab5db240292c78bb510

PR #6987, opened 2022-10-14
allow select widget class to update if it uses a filter and is output changes

### LINKED ISSUE #6986: [BUG] SelectWidget class doesn't update if it uses a filter and the expected filter result changes
### Describe the bug

Create a SelectWidget with the `class` attribute value being a filter that depends on the value set by the select.

### Expected behavior

The widget's class should update if the result of the filter changes.

### To Reproduce

1. Open https://tiddlywiki.com/empty.html
2. Create an empty tiddler with the contents being set to the contents of the first comment in this issue.
3. Observer that all Selects correctly start with black color.
4. Change any of the Selects and notice that they don't change color. The filter results are updated though, as can be seen in the bullet list.
5. Close and reopen the tiddler and notice that the colors have updated.

## Starting state, everything is correctly black:
image

## Colors have not changed despite filters returning new values:
image

## Colors are updated after closing and reopening the tiddler:
image

### TiddlyWiki Configuration

- **Version** 5.2.3
- **Saving mechanism:** None:
- **Plugins installed:** None

### Desktop 

- **OS:** Windows 10
- **Browser:** Chrome
- **Version:** Version 106.0.5249.103 (Official Build) (64-bit)

### Additional context

_No response_

EvidentlyCube:
This fixes #6986 

The issue was caused by `class` not being in the list of compared attributes.

That being said, I am wondering if it wouldn't be better to do something like this:

```
SelectWidget.prototype.refresh = function(changedTiddlers) {
	// If we're using a different tiddler/field/index then completely refresh ourselves
	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes.tooltip) {
		this.refreshSelf();
		return true;
	// If the target tiddler value has changed, just update setting and refresh the children
	} else {
                // ---> THIS HERE <----
		if (changedAttributes.class) {
			this.getSelectDomNode().className = this.getAttribute("class");            // EITHER THIS
                        this.getSelectDomNode().setAttribute("class",this.getAttribute("class"));  // OR THAT
		}
                // ---> TO HERE <----
		
		var childrenRefreshed = this.refreshChildren(changedTiddlers);
		if(changedTiddlers[this.selectTitle] || childrenRefreshed) {
			this.setSelectValue();
		} 
		return childrenRefreshed;
	}
};
```

This way the whole widget doesn't need to be refreshed. Probably the latter option would be more idiomatic given other code in TW?

### DIFF
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -170,11 +170,16 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 SelectWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
 	// If we're using a different tiddler/field/index then completely refresh ourselves
-	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes.tooltip || changedAttributes.class) {
+	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes.tooltip) {
 		this.refreshSelf();
 		return true;
 	// If the target tiddler value has changed, just update setting and refresh the children
 	} else {
+		if(changedAttributes.class) {
+			this.selectClass = this.getAttribute("class");
+			this.getSelectDomNode().setAttribute("class",this.selectClass); 
+		}
+		
 		var childrenRefreshed = this.refreshChildren(changedTiddlers);
 		if(changedTiddlers[this.selectTitle] || childrenRefreshed) {
 			this.setSelectValue();
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -170,7 +170,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 SelectWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
 	// If we're using a different tiddler/field/index then completely refresh ourselves
-	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes.tooltip) {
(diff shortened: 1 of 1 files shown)


saqimtiaz:
> That being said, I am wondering if it wouldn't be better to do something like this:

I agree that this would be better. I would perhaps also update `this.selectClass` as well, just for completeness and so that it doesn't trip us up in the future if that property gets reused elsewhere in the code.

EvidentlyCube:
@saqimtiaz I've pushed a commit with both changes.

saqimtiaz:
@EvidentlyCube thank you, that looks good to me.

EvidentlyCube:
I know there's a `tests` edition of TW that I am guessing runs tests. I have some Qs:
1. Is there any documentation on how those tests work and how they're run?
2. If not, how do they work and how to run them?
3. Can we run end-to-end tests that actually interact with the site? Because I'd love to add a test for my change

Jermolene:
Hi @EvidentlyCube 

> I know there's a `tests` edition of TW that I am guessing runs tests. I have some Qs:
> 
> 1. Is there any documentation on how those tests work and how they're run?

There are some brief docs here:

https://tiddlywiki.com/dev/#TestingMechanism

> 2. If not, how do they work and how to run them?

The tests are also run automatically on GitHub when PRs are updated or merged. We don't merge PRs unless the tests pass.

There's also a script `./bin/test.sh` for interactive use.

> 3. Can we run end-to-end tests that actually interact with the site? Because I'd love to add a test for my change

We do not yet do so but our ambition is to use Puppeteer or Playwright to run interactive tests.

We have a primitve performance management test rig that uses Puppeteer:

https://github.com/Jermolene/tiddlywiki-performance-test-rig

saqimtiaz:
@Jermolene I believe this PR can be merged.

Jermolene:
Thanks @EvidentlyCube @saqimtiaz

PR #6993, opened 2022-10-17
Backlinks that respect dynamically rendered links

Jermolene:
There has been [some discussion](https://talk.tiddlywiki.org/t/backlinks-that-respect-lists/4844) about backlinks not respecting dynamically rendered links (for example, the output of the `<<list-links>>` macro).

This has come up many times before, and I've always given the standard answer that performance would be a concern. However, I've never gone as far as experimenting to quantify the performance hit. As it turns out, it's quite usable.

This PR is a first attempt at an implementation of backlinks that fully renders tiddlers before extracting the links.

It works surprisingly well in my limited experiments. Using the tiddlywiki.com wiki, the first time I clicks on the "backlinks" tab I am experiencing a 2-4 second delay on my Mac. But thereafter things seem to be reasonable zippy thanks to the various levels of caching involved. I suspect that it consumes a good deal of memory too, but I haven't attempted to instrument it yet.

The implementation is very much a proof of concept, with some serious shortcomings in its current form:

* The new-style backlinks completely override the existing backlinks implementation. Presumably we'd want to make them available side-by-side with a switch to choose
* The global import of `[[tag[$:/tags/Macro]]` that we have to do to make globals available is repeated for each tiddler. It would be more efficient to cache the widget tree resulting from rendering the globals, and reuse it

### DIFF
--- a/core/modules/indexers/backlinks-index.js
+++ b/core/modules/indexers/backlinks-index.js
@@ -26,11 +26,23 @@ BacklinksIndexer.prototype.rebuild = function() {
 }
 
 BacklinksIndexer.prototype._getLinks = function(tiddler) {
-	var parser =  this.wiki.parseText(tiddler.fields.type, tiddler.fields.text, {});
-	if(parser) {
-		return this.wiki.extractLinks(parser.tree);
-	}
-	return [];
+	var parser =  this.wiki.parseText(tiddler.fields.type,tiddler.fields.text,{});
+	parser.tree = [{
+		type: "importvariables",
+		attributes: {
+			filter: {
+				name: "filter",
+				type: "string",
+				value: "[[$:/core/ui/PageMacros]] [all[shadows+tiddlers]tag[$:/tags/Macro]!has[draft.of]]"
+			}
+		},
+		isBlock: false,
+		children: parser.tree
+	}];
+	var widget = this.wiki.makeWidget(parser,{document: $tw.fakeDocument, parseAsInline: false, variables: {currentTiddler: tiddler.fields.title}});
+	var container = $tw.fakeDocument.createElement("div");
+	widget.render(container,null);
+	return this.wiki.extractLinksFromWidgetTree(widget);
 }
 
 BacklinksIndexer.prototype.update = function(updateDescriptor) {
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -22,7 +22,8 @@ Adds the following properties to the wiki object:
 /*global $tw: false */
 "use strict";
 
-var widget = require("$:/core/modules/widgets/widget.js");
+var widget = require("$:/core/modules/widgets/widget.js"),
+	LinkWidget = require("$:/core/modules/widgets/link.js").link;
 
 var USER_NAME_TITLE = "$:/status/UserName",
 	TIMESTAMP_DISABLE_TITLE = "$:/config/TimestampDisable";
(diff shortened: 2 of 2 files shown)


saqimtiaz:
I have used the same approach to implement a filter operator that returns a list of the images used in a tiddler, including those dynamically rendered via macros and list widgets. Works well in practice.

We could consider a more flexible version of the extractLinksFromWidgetTree method that can cater to more use cases, such as extracting images or external links.

yaisog:
Would it be possible to make the caching state available in a variable?
In very large wikis I'd assume the initial run will take _much_ longer than 2–4 seconds. One could imagine that backlinks would only be shown in the ViewTemplate if the initial run has already been performed, and otherwise hidden, depending on the value of that variable. There could then be a button somewhere to do the initial backlink analysis on demand (while grabbing a coffee or something).

Jermolene:
> Would it be possible to make the caching state available in a variable?
> In very large wikis I'd assume the initial run will take _much_ longer than 2–4 seconds. One could imagine that backlinks would only be shown in the ViewTemplate if the initial run has already been performed, and otherwise hidden, depending on the value of that variable. There could then be a button somewhere to do the initial backlink analysis on demand (while grabbing a coffee or something).

The challenge does indeed seem to be that we have no way of predicting how long the initial indexing will take. As you suggest, it could be done manually, or perhaps we could check the running time as we go and abort early if we find that the indexing is taking too long.

pmario:
>The challenge does indeed seem to be that we have no way of predicting how long the initial indexing will take. As you suggest, it could be done manually, or perhaps we could check the running time as we go and abort early if we find that the indexing is taking too long.

There may be a possibility to let a worker do the heavy lifting and have a status variable once the indexing is done. So it may take several seconds at startup time for the status to be OK but then it should be good to go. 

At the moment we clear the caches very often. There would need to be some better handling for cache updating instead of clearing it and start from scratch

Jermolene:
> There may be a possibility to let a worker thread do the heavy lifting and have a status variable once the indexing is done. So it may take several seconds at startup time for the status to be OK but then it should be good to go.

I think that the problem with offloading the work to a worker thread is that the worker thread would need a copy of the entire wiki in order to do the rendering. Meanwhile, we have to have the wiki in the main thread to handle interactive refreshes. So the implication is that we'd need to sync a copy of the wiki between the two threads.

linonetwo:
Having copy is ok I think. In nodejs setup, we also have copy on client and server side. And we might also have sqlite indexer in the future, that will have the whole wiki's tiddler inside of it too!

pmario:
>Having copy is ok I think. In nodejs setup, we also have copy on client and server side.

I think we cannot compare a client - server configuration with a wiki on a mobile phone, without any internet connection. The whole caching and indexing needs to be done on the phone. ... and it should still work well.

Jermolene:
Hi @linonetwo @pmario while it is true that the client server configuration requires a separate copy of the wiki on the client and the server, I don't think that that makes it reasonable to have two copies of the wiki stored on a client device. The fundamental constraint here is that the TiddlyWiki rendering code requires synchronous access to the tiddler store. That means that the most easily accessible architecture would be to run sqlite under WASM on the main client thread. The other interesting architecture might be not just to run the wiki store in a worker thread, but the entire server including the store and widget trees and rendering, and just have HTML diffs passed back to the UI thread.

PR #7004, opened 2022-10-20
Support nested macro definitions

Jermolene:
### Introduction

It is common for macros to require further "sub-macros" for their operation. For example, here is an outline of a macro that renders a special button that uses a further macro to contain the action string:

```
\define actions()
<$action-sendmessage ... />
\end

\define special-button()
<$button actions=<<actions>>>
Caption
</$button>
\end
```

That should work fine in most situations, but a particularly pernicious case arises when the macros are made global by being tagged `$:/tags/Macro`. The problem then is that not only does `special-button` get exposed as a global variable, so does the `actions` macro. That means that the `<<special-button>>` macro would not work correctly if it was invoked in a context where the variable `actions` has been redefined. For example:

### Proposal

It is proposed to extend the macro definition syntax to allow the macro name to be specified after the `\end` keyword. When parsing a macro, the end marker is taken to be whichever is found first of an empty `\end` marker or a matching `\end <macroname>` marker.

This makes it possible to nest macros inside one another, completely hiding the nested definitions from the outside world.

The example above could instead be written as:

```
\define special-button()
\define actions()
<$action-sendmessage ... />
\end actions
<$button actions=<<actions>>>
Caption
</$button>
\end special-button
```

Note that the following example does not nest as expected because when parsing the parent definition it will find the first `\end` marker and use that as the terminator:

```
\define parent ( )
 commands, comments, ...
 \define child ()
    ....
 \end
 ....
\end
```

In other words, when nesting macro definitions, only the innermost macro definition can omit the macro name from the `\end` marker.

### To do

- [x] Documentation

### DIFF
--- a/editions/tw5.com/tiddlers/wikitext/Macro Definitions in WikiText.tid
+++ b/editions/tw5.com/tiddlers/wikitext/Macro Definitions in WikiText.tid
@@ -1,13 +1,15 @@
 caption: Macro Definitions
 created: 20150220181617000
-modified: 20180820165115455
+modified: 20221022135909352
 tags: WikiText
 title: Macro Definitions in WikiText
 type: text/vnd.tiddlywiki
 
 A [[macro|Macros]] is defined using a `\define` [[pragma|Pragma]]. Like any pragma, this can only appear at the start of a tiddler.
 
-The first line of the definition specifies the macro name and any parameters. Each parameter has a name and, optionally, a default value that is used if no value is supplied on a particular call to the macro. The lines that follow contain the text of the macro text (i.e. the snippet represented by the macro name), until `\end` appears on a line by itself:
+The first line of the definition specifies the macro name and any parameters. Each parameter has a name and, optionally, a default value that is used if no value is supplied on a particular call to the macro.
+
+The lines that follow contain the text of the macro text (i.e. the snippet represented by the macro name), until `\end` appears on a line by itself:
 
 <$codeblock code={{$:/editions/tw5.com/macro-examples/say-hi}}/>
 
@@ -16,6 +18,20 @@ Alternatively, the entire definition can be presented on a single line without a
 ```
 \define sayhi(name:"Bugs Bunny") Hi, I'm $name$.
 ```
+
+Macro definitions can be nested by specifying the name of the macro in the `\end` marker. For example:
+
+<<wikitext-example-without-html src:"""\define special-button(caption:"Click me")
+\define actions()
+<$action-sendmessage $message="tm-notify" $param="HelloThere"/>
+\end actions
+<$button actions=<<actions>>>
+$caption$
+</$button>
+\end special-button
+
+<<special-button>>
+""">>
 
 A more formal [[presentation|Macro Definition Syntax]] of this syntax is also available.
 
--- a/core/modules/parsers/wikiparser/rules/macrodef.js
+++ b/core/modules/parsers/wikiparser/rules/macrodef.js
@@ -27,7 +27,7 @@ Instantiate parse rule
 exports.init = function(parser) {
 	this.parser = parser;
 	// Regexp to match
-	this.matchRegExp = /^\\define\s+([^(\s]+)\(\s*([^)]*)\)(\s*\r?\n)?/mg;
+	this.matchRegExp = /^(\\+)define\s+([^(\s]+)\(\s*([^)]*)\)(\s*\r?\n)?/mg;
 };
 
 /*
@@ -37,7 +37,7 @@ exports.parse = function() {
 	// Move past the macro name and parameters
 	this.parser.pos = this.matchRegExp.lastIndex;
 	// Parse the parameters
-	var paramString = this.match[2],
+	var paramString = this.match[3],
 		params = [];
 	if(paramString !== "") {
 		var reParam = /\s*([A-Za-z0-9\-_]+)(?:\s*:\s*(?:"""([\s\S]*?)"""|"([^"]*)"|'([^']*)'|\[\[([^\]]*)\]\]|([^"'\s]+)))?/mg,
@@ -56,9 +56,9 @@ exports.parse = function() {
 	}
 	// Is this a multiline definition?
 	var reEnd;
-	if(this.match[3]) {
+	if(this.match[4]) {
 		// If so, the end of the body is marked with \end
-		reEnd = /(\r?\n\\end[^\S\n\r]*(?:$|\r?\n))/mg;
+		reEnd = new RegExp("(\\r?\\n" + $tw.utils.repeat("\\\\",this.match[1].length) + "end[^\\S\\n\\r]*(?:$|\\r?\\n))","mg");
 	} else {
 		// Otherwise, the end of the definition is marked by the end of the line
 		reEnd = /($|\r?\n)/mg;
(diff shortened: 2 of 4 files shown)


kookma:
This is a good PR and has use cases as you said, or when one wants to encapsulate an object (properties, and methods) in a single macro with several nested macros.

As an example, in Matlab we have three types of function

* Main Function (the first function in a file and the name of file is the same as function, nothing appears before it except comments), visible to all other file and also working space
* Sub function (any function defined after main function) and is visible to all other function in the same file
* Nested function (any function inside other function

Matlab and Fortran and Python use a similar approach like

```
function parent ( )
 commands, comments, ...
 function child ()
    ....
 end 
 ....
end
```


This can be used in any number of nesting levels. Fortran, and Matlab also allow to add the name of function to `end` statement. This is for readability like

```
function parent ( )
 commands, comments, ...
 function child ()
    ....
 end child
 ....
end parent
```


I may suggest a similar approach for nesting

```
\define parent ( )
 commands, comments, ...
 \define child ()
    ....
 \end child
 ....
\end parent
```

Or simply 

```
\define parent ( )
 commands, comments, ...
 \define child ()
    ....
 \end
 ....
\end
```

If parser see `\define name()` in a line then looks for a match `\end` else raises an error

Jermolene:
Thank you @kookma using `\end <macroname>` is a 100x better than multiplying the backslashes. I'll amend this PR as soon as I get a chance.

kookma:
I would be happy to create some examples for docs, ones it is ready to go.

In 5.2.3 and older TW, I have used nested macro, as single line short macros have no requirement for end, so the below code works in 5.2.3

```
\define parent(name:"")
\define child() $name$ middle name is Jack
Hi $name$.
<$list filter="[<__name__>match[Motovun]]"><<child>></$list>
\end

<<parent Micky>>

<<parent "Motovun">>
```
This produces

```

Hi Micky.

Hi Motovun. Motovun middle name is Jack

```

saqimtiaz:
There is an interesting issue that has come up on the discussion forum related to nested macros within action strings: https://talk.tiddlywiki.org/t/a-macro-trap-in-case-of-side-effect-involved/4885

In short, the action strings within nested macro definitions are also executed. Considering the widget tree created one understands why this happens. The question is whether we should try to introduce some mitigation against this while adding better support for nested macros, or just try to document the behaviour?

Jermolene:
Hi @saqimtiaz 

> In short, the action strings within nested macro definitions are also executed. Considering the widget tree created one understands why this happens. The question is whether we should try to introduce some mitigation against this while adding better support for nested macros, or just try to document the behaviour?

I couldn't follow the issue in that thread. Could you give an example of the problem? Is it just the unexpected results of using a macro containing nested macro definitions as the value of an action attribute?

Jermolene:
I've updated the PR as per @kookma's [suggestion above](https://github.com/Jermolene/TiddlyWiki5/pull/7004#issuecomment-1286429236). I also added a note to explain why one of the examples given by @kookma still won't work quite as expected.

kookma:
I will do some experiments and if there is any issue I will share here!

Jermolene:
Thanks @kookma much appreciated.

kookma:
Hi @Jermolene 
I did several tests, and all passed except one!


The one fails

Case 1: No Nesting (code works)

```
\define new-tiddler-text() 
Go to code: <$link to="$(createTiddler-title)$" />
\end new-tiddler-text

\define create-actions()
<$action-createtiddler $basetitle=<<__tidname__>> tags="code" text=<<now>> >
  <$action-createtiddler $basetitle={{{ [<createTiddler-title>addsuffix[/remark]] }}} tags="remark" text=<<new-tiddler-text>> />
</$action-createtiddler>
\end create-actions


\define new-code(tidname, label)
<!-- main macro -->
<$button actions=<<create-actions>> >$label$</$button>

\end new-code

<<new-code tidname:"test001" label:"Make it">>
```

Case ii: Nesting (code fails). The one failes

```
\define new-code(tidname, label)
<!-- main macro -->

\define new-tiddler-text() 
Go to code: <$link to="$(createTiddler-title)$" />
\end new-tiddler-text

\define create-actions()
<$action-createtiddler $basetitle=<<__tidname__>> tags="code" text=<<now>> >
  <$action-createtiddler $basetitle={{{ [<createTiddler-title>addsuffix[/remark]] }}} tags="remark" text=<<new-tiddler-text>> />
</$action-createtiddler>
\end create-actions


<$button actions=<<create-actions>> >$label$</$button>

\end new-code

<<new-code tidname:"test001" label:"Make it">>
```

This line fails `Go to code: <$link to="$(createTiddler-title)$" />` and substitution does not work, an empty value inserted.

kookma:
Further update:

Consider a parent with two children.
The second child calls first one, where the first access to a vars from second child.
It seemes vars like `$()$`

pmario:
>Case ii: Nesting (code fails). The one failes

@kookma ... If something fails please create a new issue as you did for other issues. This makes it sure, that we don't miss it in the "wall of text" we already have with this PR.

kookma:
> > Case ii: Nesting (code fails). The one failes
> 
> @kookma ... If something fails please create a new issue as you did for other issues. This makes it sure, that we don't miss it in the "wall of text" we already have with this PR.

Sure. 

See  https://github.com/Jermolene/TiddlyWiki5/issues/7006

editions/tw5.com/tiddlers/wikitext/Macro Definitions in WikiText.tid
```diff
@@ -17,6 +19,20 @@ Alternatively, the entire definition can be presented on a single line without a
...
+<$action-sendmessage $message="tm-notify" $param="HelloThere"/>
+\end actions
+<$button actions=<<actions>>>
+$caption$
+</$button>
+\end special-button
+
+<<special-button>>
+""">>
+
```
kookma, line 35: Question 1: Is this a requirement to have all nested macros right after `\define parent()`?
Of course, the current logic is fine and makes codes simpler to follow.

Question ii: Is it possible to allow one tab, or few whitespaces before `\define` and `\end` for nested macros?
I understood pragma `\define` shall start from first column!

Jermolene: > Question 1: Is this a requirement to have all nested macros right after `\define parent()`? Of course, the current logic is fine and makes codes simpler to follow.

Yes – the contents of the parent macro forms a new parsing context that behaves just the same as when the entire body of a tiddler is parsed: pragmas still need to be at the top.

> Question ii: Is it possible to allow one tab, or few whitespaces before `\define` and `\end` for nested macros? I understood pragma `\define` shall start from first column!

I wondered about that, I guess it will be a popular request. I think we'd apply that to all pragmas if we were to change it.

kookma:
Hi @Jermolene 
I put some comments on nested macro docs. Seems I should comment here!  Sorry for inconveniences!

* https://github.com/FlashSystems/TiddlyWiki5/commit/bcb3b06d60a3b616083e9df36fc487cbbbf8abb7#commitcomment-91922651
* https://github.com/FlashSystems/TiddlyWiki5/commit/bcb3b06d60a3b616083e9df36fc487cbbbf8abb7#commitcomment-91922974

Jermolene:
Thanks @kookma, I've made some updates in 28c1e6bfc3b4a842323ab6cc03c34c17a93e45e9

PR #7007, opened 2022-10-23
Handle ZoomIn issues when using tiddler template that starts with a text node

### LINKED ISSUE #6992: [BUG] Error thrown in ZoomIn story view when first dom node is Text node
### Describe the bug

The setup is quite complicated. You need the following:

* Control Panel -> Appearance -> Story View set to **zoomin**
* A cascade `$:/tags/StoryTiddlerTemplateFilter` that causes a tiddler to use a new template
* A template where the first dom element is a text node (rather than a dom element)
* A tiddler that uses this template and another tiddler

Then you need the following setup:
* Open the tiddler using the template
* Open the tiddler that doesn't use the template
* Close the second tiddler

This causes `zoomin.js::remove()` to try to open the bugged tiddler and run this line of code:

```
$tw.utils.addClass(toWidgetDomNode,"tc-storyview-zoomin-tiddler");
```

`toWidgetDomNode` is a TextNode so it throws an exception inside because `getAttribute()` method does not exist on it.

### Expected behavior

No exception

### To Reproduce

1. Go to https://zoomin-story-template-bug.tiddlyhost.com/
2. In the right menu click on the **Bug** tiddler to open it
3. Close the tiddler

### Screenshots

_No response_

### TiddlyWiki Configuration

- **Version:** 5.2.3

### Additional context

_No response_

EvidentlyCube:
This fixes #6992.

Two changes were made here:

1. Anytime ZoomIn tries to interact with a tiddler that is rendered through `$:/tags/StoryTiddlerTemplateFilter` and starts with a text node, a warning alert appears.
2. Prevent TW getting stuck when such a tiddler is rendered, making it impossible to fix the issue.

The alert in question looks like this:
image

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -40,6 +40,7 @@ Error/RetrievingSkinny: Error retrieving skinny tiddler list
 Error/SavingToTWEdit: Error saving to TWEdit
 Error/WhileSaving: Error while saving
 Error/XMLHttpRequest: XMLHttpRequest error code
+Error/ZoominTextNode: `<h2>''Story View Error''</h2>It appears you tried to interact with a tiddler that displays in a custom container. This is most likely caused by using `$:/tags/StoryTiddlerTemplateFilter` with a template that contains text or whitespace at the start. Please use the pragma <code>\whitespace trim</code> and ensure the whole contents of the tiddler is wrapped in a single HTML element.<br><br>''The text that caused this issue:''`
 InternalJavaScriptError/Title: Internal JavaScript Error
 InternalJavaScriptError/Hint: Well, this is embarrassing. It is recommended that you restart TiddlyWiki by refreshing your browser
 LayoutSwitcher/Description: Open the layout switcher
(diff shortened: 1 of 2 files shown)


Jermolene:
My apologies @EvidentlyCube I missed this. It's a great idea, but I'm not sure about some of the implementation details.

The broader context is that we have a general issue with how to present error and debugging information to wikitext authors (without resorting to browser dev tools which are not available on mobile etc). In some situations such as filter evaluation we don't have a direct relationship with a DOM node in which to display the information.

Fixing all of that is a bigger issue, so I suggest for the moment that we strip this back to just doing the basic error detection, and log the error to the browser dev console. It's not perfect because there's clearly a big chance that the messages won't be seen, but it's better than the present situation, and would allow us to return to the broader issues when we're ready.

EvidentlyCube:
@Jermolene No worries, I took my time too. I've changed it to log to console rather than show an alert.

Jermolene:
Thank you @EvidentlyCube

Issue #7002, opened 2022-10-20
[BUG] Content in tables has the tendency to overflwo on small mobile screens


pmario:
### Describe the bug

This info is related to PR#6842 comment https://github.com/Jermolene/TiddlyWiki5/pull/6842#discussion_r1000384038

>Jeremy wrote
>
>One big problem is that the formatting breaks down entirely when viewed at mobile sizes. I really think that it is essential that the cheat sheet is usable on a small screen.

and https://github.com/Jermolene/TiddlyWiki5/pull/6842#discussion_r1000505286

>Mario wrote
>
>**I'll create a completely new PR for this issue.**
>
>That's a general problem with TW tables. This problem has always been there, the PR only makes it visible.

<details open><summary><b> Click to toggle screenshot</b> </summary>

![image](https://user-images.githubusercontent.com/374655/196974750-ace44c82-5fc2-445c-9297-7cf81b654796.png)

</details>

There are 2 elements responsible 

 1.    For small screens we need: `word-wrap `to be set to `word-wrap: anywhere `within tables, otherwise they create table overflows.
 3.   default margins for `blockquotes, OL-LI and UL-li `need to be smaller in tables too. Especially if they are nested as shown in the examples here. They create too much white space, which in turn create table overflow

### To Reproduce

- Go to tiddlywiki.com
- download empty
- Import attached JSON tiddlers TODO

### TiddlyWiki Configuration

TW 5.2.3

pmario:
**PR is on the way**

PR #6842, opened 2022-07-28
Wikitext docs overhaul first step

pmario:
- The WikiText tiddler overview tiddler shows a logical sort order stored in the field `user-story`
   - The list is ordered by complexity -- simple -> complex
   - The list is 4 columns and is responsive with 2 columns for small screens
- "Formatting in WikiText" uses a table now and has more info
- The PR moves the multi-colums CSS definition used in "Formatting List Results as Tables with CSS - Specified Columns Methods" to the $:/_tw5.com-styles tiddler, so it can be reused 
- Markdown is a new definition tiddler. I don't want to have external links, when we explain our TW syntax.
- "Widgets in Wikitext" ... links to Widgets now, which shows the list of all widgets. There is no need to have that list while explaining the wikitext syntax.

There may be more steps, but this could be a start. This [discussion at Talk](https://talk.tiddlywiki.org/t/a-better-documentation-for-wikitext/4066) triggered the PR.

PR #7010, opened 2022-10-25
fix table overflow for small screens

pmario:
This PR fixes:  **[BUG] Content in tables has the tendency to overflwo on small mobile screens** #7002 

To reproduce: 

- Open tiddlywiki.com
- import this file: [table-overflow-fix-example.zip](https://github.com/Jermolene/TiddlyWiki5/files/9859045/table-overflow-fix-example.zip)
- make your window small. .. there will be a table overflow -> PROBLEM
- Start with the fix in the PR
- Fix shown in the GIF below
- or
- open with a mobile phone in portrait mode

The indentation for all list elements **in tables** will be adjusted to half of their size to gain some more text space.

table-small-screen

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1155,7 +1155,7 @@ button.tc-btn-invisible.tc-remove-tag-button {
 	}
 
 	pre {
-		word-break: break-word; /* needed for Chrome and Safari on mobile */
+		word-break: break-word;
 	}
 
 	table blockquote {
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -148,7 +148,7 @@ pre {
 	display: block;
 	margin-top: 1em;
 	margin-bottom: 1em;
-	word-break: normal;
+	word-break: break-word;
 	word-wrap: break-word;
 	white-space: {{$:/themes/tiddlywiki/vanilla/options/codewrapping}};
 	background-color: <<colour pre-background>>;
@@ -1149,26 +1149,22 @@ button.tc-btn-invisible.tc-remove-tag-button {
 
 
 /* prevent overflow for table content for small screens*/
-@media (max-width: <<tiddler-width>>) {
-	tr {
+@media (max-width: <<sidebarbreakpoint>>) {
+	.tc-tiddler-frame tr {
 		word-wrap: anywhere;
 	}
 
-	pre {
-		word-break: break-word;
-	}
-
-	table blockquote {
+	.tc-tiddler-frame table blockquote {
 		margin-left: 12.5px;
 		margin-right: 12.5px; 
 	}
 
-	table dd {
+	.tc-tiddler-frame table dd {
 		margin-left: 12.5px;
 	}
 
-	table ol,
-	table ul {
+	.tc-tiddler-frame table ol,
+	.tc-tiddler-frame table ul {
 		padding-left: 20px;
 	}
 }
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1154,6 +1154,10 @@ button.tc-btn-invisible.tc-remove-tag-button {
 		word-wrap: anywhere;
 	}
 
+	pre {
+		word-break: break-word; /* needed for Chrome and Safari on mobile */
+	}
+
 	table blockquote {
 		margin-left: 12.5px;
 		margin-right: 12.5px; 
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -30,6 +30,10 @@ background-size:` {{$:/themes/tiddlywiki/vanilla/settings/backgroundimagesize}}`
 <$text text={{{ [{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}removesuffix[px]subtract[1]addsuffix[px]] ~[{$:/themes/tiddlywiki/vanilla/metrics/sidebarbreakpoint}] }}}/>
 \end
 
+\define tiddler-width()
+	<$text text={{$:/themes/tiddlywiki/vanilla/metrics/tiddlerwidth}}/>
+\end
+
 \define if-fluid-fixed(text,hiddenSidebarText)
 <$reveal state="$:/themes/tiddlywiki/vanilla/options/sidebarlayout" type="match" text="fluid-fixed">
 $text$
@@ -1140,6 +1144,28 @@ button.tc-btn-invisible.tc-remove-tag-button {
 
 	.tc-tiddler-info {
 		margin: 0 -42px 0 -42px;
+	}
+}
+
+
+/* prevent overflow for table content for small screens*/
+@media (max-width: <<tiddler-width>>) {
+	tr {
+		word-wrap: anywhere;
+	}
+
+	table blockquote {
+		margin-left: 12.5px;
+		margin-right: 12.5px; 
+	}
+
+	table dd {
+		margin-left: 12.5px;
+	}
+
+	table ol,
+	table ul {
+		padding-left: 20px;
 	}
 }
 
--- a/core/modules/filters/is/variable.js
+++ b/core/modules/filters/is/variable.js
@@ -19,13 +19,13 @@ exports.variable = function(source,prefix,options) {
 	var results = [];
 	if(prefix === "!") {
 		source(function(tiddler,title) {
-			if(!(title in options.widget.variables)) {
+			if(options.widget.getVariable(title) === undefined) {
 				results.push(title);
 			}
 		});
 	} else {
 		source(function(tiddler,title) {
-			if(title in options.widget.variables) {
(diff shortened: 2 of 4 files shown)


pmario:
@Jermolene ... It would be nice to handle this PR fast since the Sheat-sheet depends on it.

pmario:
I did test it with 

- FF on Windows 
- Edge on Win
- FF on Android
- Chrome on Android  --> uuups There seems to be a problem with that one.

pmario:
hmmm -- it's likely to be a problem with safari too: https://caniuse.com/?search=anywhere  it doesn't understand `word-wrap: anywhere`

@Jermolene **Please check for an alternative that we can put in front of it too** May be an other option would work too

pmario:
> Chrome on Android --> uuups There seems to be a problem with that one.

See commit https://github.com/Jermolene/TiddlyWiki5/pull/7010/commits/5e8f4ed66da2f3059f3d942bf98587fd9429d824 which seems to work for me on Chrome Android now.

pmario:
@Jermolene ... You mentioned, that v5.2.4 should be released in Nov. So I think my PRs should be merged or at least be commented on.

pmario:
@Jermolene ... bump

Jermolene:
Thanks for your patience @pmario. I am concerned that this is a bit brittle. Using a dynamic media is likely to have unexpected effects when the affected elements are used outside tiddlers. Do we have any idea if other sites or apps have this problem?

pmario:
I don't understand this question. In TW UI we don't use TABLEs outside the tiddler-frame. .. 

IMO The problem is related to the fact, that our sidebar is moved to the top of the tiddler frame using media queries. ... These existing media queries bring the CSS out of balance. Which in turn causes the table overflow problem. 

So for me it seems to be clear to also use a media query to fix it.

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1143,6 +1147,32 @@ button.tc-btn-invisible.tc-remove-tag-button {
 	}
 }
 
+
+/* prevent overflow for table content for small screens */
+@media (max-width: <<tiddler-width>>) {
+	tr {
```
Jermolene, line 1153: This definition globally redefines the `<tr>` element, regardless of whether it is used inside a tiddler. So that means that (say) adding a table to a new menubar would be hit by this rule.

Jermolene (commented):
Hi @pmario I was referring to the fact that this PR unconditionally redefines several elements, which affects their display in all contexts.

themes/tiddlywiki/vanilla/base.tid
pmario: Yes that's intentional for tables with small screens.

pmario: On small screens it also half's the indent distance for all list types, to improve the readability for "real" text instead of showing a lot of whitespace. See the animated gif

Jermolene: The trouble is that these definitions are very broad. They would hit content embedded in widgets by third party libraries, for example. I do appreciate that that is also true of our table CSS generally, but there's no reason to make that worse.

pmario:
@Jermolene .. I did change the CSS settings for small screens, so only the tables within tc-tiddler-frame are affected. 

I needed to change the PRE element globally, since it won't work for Edge browser without that change. .. It seems Edge and Chrome calcualte the spacing for tables differently than FF

Jermolene:
Thanks @pmario

Jermolene:
Hi @pmario I've just run into a serious issue with the changes in this PR. I was working on a simple wikitext table with three columns. Here's how it looks with the current "master" branch:

Screenshot 2023-01-15 at 20 15 11

Here's how the same table looks with `word-wrap: anywhere` commented out:

Screenshot 2023-01-15 at 20 14 46

As you can see, `word-wrap: anywhere` is splitting the words in the first column in order to try to make the middle column as wide as possible.

This is clearly unacceptable. I should never have merged this PR if I had properly read what `word-wrap: anywhere` actually does.

I am really sorry that we've run into so many problems with PRs like this. You are addressing significant problems that are worth fixing. The problem is that we need much stronger processes in place to understand the impact of these sort of CSS changes. In particular, we all need to understand the logic of the changes being proposed, and the trade-offs being made.

pmario:
IMO there is something wrong with the media query. `@media (max-width: <<sidebarbreakpoint>>) {` 

Your screenshot doesn't look like a small screen? what's the resolution?

pmario:
Did you try this with a different browser? What's your browser?

Jermolene:
> Did you try this with a different browser? What's your browser? ... May be you only need to fix that for Safari??

The screenshots are from Chrome for macOS. window.innerWidth is 949 (I'm using a widescreen with developer tools open).

pmario:
hmmm, It seems with all the back and forth I've used the the wrong media query ... `@media (max-width: <<sidebarbreakpoint>>) {` instead of `@media (max-width: <<tiddler-width>>) {`

Could your try `<<tiddler-width>>` ...  It's expected, that the table gets weird for small screens at some point.

pmario:
>The screenshots are from Chrome for macOS. window.innerWidth is 949 (I'm using a widescreen with developer tools open).

That's strange. Your images have a native resolution from 1802 x 1422 ... So that's nowhere near 949 ... I have no real clue ??

Jermolene:
> That's strange. Your images have a native resolution from 1802 x 1422 ... So that's nowhere near 949 ... I have no real clue ??

I cropped the image to just show the table. You should be able to duplicate the same effect. Here's the table source:

```

|!Name |!Description |!Default |
|''particleCount'' |The number of confetti to launch |50 |
|''angle'' |The angle in which to launch the confetti, in degrees (90 is straight up) |90 |
|''spread'' |How far off center the confetti can go, in degrees. 45 means the confetti will launch at the defined `angle` plus or minus 22.5 degrees |45 |
|''startVelocity'' |How fast the confetti will start going, in pixels |45 |
|''decay'' |How quickly the confetti will lose speed. Keep this number between 0 and 1, otherwise the confetti will gain speed |0.9 |
|''gravity'' |How quickly the particles are pulled down. 1 is full gravity, 0.5 is half gravity, etc. |1 |
|''drift'' |How much to the side the confetti will drift. The default is 0, meaning that they will fall straight down. Use a negative number for left and positive number for right |0 |
|''ticks'' |How many times the confetti will move (this is an abstract quantity; the designed recommends playing with it if the confetti disappears too quickly for you) |200 |
|''originX'' |The `x` position on the page, with `0` being the left edge and `1` being the right edge |0.5 |
|''originY'' |The `y` position on the page, with `0` being the top edge and `1` being the bottom edge |0.5 |
|''colors'' |A space separated list of color strings in hex format (eg `#bada55` or `#ce5`) | |
|''shapes'' |A space separated list of shapes for the confetti. The possible values are `square`, `circle`, and `star`. The default is to use both squares and circles in an even mix. To use a single shape, you can provide just one shape in the list, such as `star`. You can also change the mix by providing a value such as `circle circle square` to use two third circles and one third squares | |
|''scalar'' |Scale factor for each confetti particle. Use decimals to make the confetti smaller |1 |
|''zIndex'' |Z-index of confetti. Increase the value if the confetti is appearing behind other on-screen elements|100 |
|''disableForReducedMotion'' |Set to `yes` to entirely disable confetti for users that [[prefer reduced motion|https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-reduced-motion]] |`no` |
```

> Do you have a OS based zoom factor active. ... May be we need to completely adjust our media-query width calculations to take modern devices into account.

It is a HiDPI display, that's all. I think all current Macs are HiDPI.

pmario:
With browser zoom-level at 100% I can go to this, where the whole browser width is 963 px. So just before it activates the media query.

image

When the media query is active it starts to look weird ... even if the sidebar is gone and the table gets wider. 
So it should be `@media (max-width: <<tiddler-width>>) {` :/

The following screenshot is with `<<tiddler-width>>` and 942px browser width. 

image

It starts to degrade at: 678 px. .. But that's expected otherwise the tables can _not_ be contained in tiddlers. 
There is a problem with GitHub, because it shows all images a max width and not in native size. So you need to save the images to actually see, how they look like :/

image

pmario:
So the conclusion may be ... We cannot contain tables in small screens ... At least I don't know how.

Jermolene:
> When the media query is active it starts to look weird ... even if the sidebar is gone and the table gets wider. So it should be `@media (max-width: <<tiddler-width>>) {` :/

Surely such a rule would evaluate differently depending on whether the sidebar were visible?

> It starts to degrade at: 678 px. .. **But that's expected otherwise the tables can _not_ be contained in tiddlers**. 

In my tests, removing "word-wrap: anywhere" fixes things perfectly. With the sidebar hidden, it is only when the window.innerWidth gets down to 450px or so that the table starts to get difficult to read.

pmario:
>This is clearly unacceptable. I should never have merged this PR if I had properly read what word-wrap: anywhere actually does.

I'm sorry that it causes problems, but the reason why I did create this PR was this comment. https://github.com/Jermolene/TiddlyWiki5/pull/6842#discussion_r1000384038

PR #7013, opened 2022-10-25
Fixed PR to fix popup position if popup is triggered from within an offsetParent element

FlashSystems:
This PR is the successor of the reverted PR #6887.

It contains the following changes:
- Fixed the bug the led to the crash when building the static version of TW (see 73ebdf0cd3a4221208e92017d361b72d1d43360f for details).
- Added a comment to the static methods that are safe to call even if the Popup class was not initialized.

I've updated the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) as well to ease testing.

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -294,13 +294,21 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 		});
 		
 		if(selectedNode.offsetLeft) {
-			// Add a variable with a popup coordinate string for the selected node
-			variables["tv-popup-coords"] = $tw.popup.buildCoordinates($tw.popup.coordinatePrefix.csOffsetParent,{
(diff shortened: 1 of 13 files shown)


FlashSystems:
This PR was tested with:
- `tiddylwiki editions/tw5.com --verbose --build index`
- `tiddylwiki editions/tw5.com --verbose --build static`
- `tiddylwiki editions/test`
- The [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) page.

I'm not using the DynAnnotate plugin but had to modify one line that generates a coordinate string. I would be glad if someone could test this code path or tell me how to test it.

FlashSystems:
Hello @Jermolene,

I fixed the bug that made the static build fail. Could you please review and merge this again? I gave this a few more weeks of testing and did not notice any problems.

Jermolene:
Thanks for your patience @FlashSystems

Issue #4767, opened 2020-07-23
[BUG] Renaming tiddler with lazy-loading enabled can delete other tiddlers' bodies


flibbles:
When you rename a tiddler (say, from A to B), then any tiddler that references A in its `tags` or `lists` field which is still skinny (it hasn't been loaded yet) will have its `text` field deleted. This is because the relinker will update the `tags` or `lists` field, but when it goes to save the tiddler, it doesn't have the text (because the server never gave it to the browser), so the browser ends up telling the server that the new text is an empty string.

This is **especially problematic** if you have my *Relink* plugin installed, [since relinking is far more likely to occur](https://github.com/flibbles/tw5-relink/issues/20).

**Fix**

There are two ways to fix this as far as I can see:

1. **The easy way:** Program the browser to load any skinny tiddlers whose fields it must update before sending it back to the server to update. Easy, but I don't think a good solution.

2. **The hard-but-better-longterm-solution way:** Create a new SyncAdapter task, RenameTiddler. The renaming occurs on the server, and then maybe the server sends the browser a list of updated tiddlers. This seems like more work, but it's definitely **my** preferred solution, since it's the only way my *Relink* plugin could handle lazy-server renaming. *Relink* can change the bodies of tiddlers, but unless it eagerly loads *every* tiddler, it can't do its job on the browser. The server has to do it. I realize *Relink* isn't core Tiddlywiki's problem, but if we ever want to adopt better relinking in core, this is necessary.

Does anybody have any idea how feasible solution 2 is? I'm not that familiar with the syncAdapter.

-Flibbles

flibbles:
So... any thoughts? Because this is a pretty bad bug for anyone using lazy loading.

Jermolene:
Hi @flibbles apologies for the late response.  As you say, moving rename operations to the server would be the logical solution. One could imagine something along the lines of the remote command processing in #3461 for this. Having said that, I find that lazy loading introduces a lot more problems than it solves, and I would seriously consider dropping it entirely. I don't think that there is much practical benefit in most situations.

flibbles:
Understood. I'll have my plugin warn against use of lazy-loading. I suppose this bug should be closed as "Won't fix" then?

Jermolene:
Hi @flibbles while there's no easy solution there are definitely a few things we could do:

* Pick up the external tasks work that I referenced above
* Canvas opinion about dropping support for lazy loading

joshuafontany:
[...]

PR #7014, opened 2022-10-25
Fix for #4767: lazy-loading deletes tiddler bodies

flibbles:
This is a fix to mitigate the horribleness that is lazy-loading. Without this, lazy-loading, especially combined with some plugins, can delete entire swaths of the user's tiddlers. It's *bad*.

This fix makes it better. It should no longer delete people's tiddlers, but lazy-loading still has problems. Relink doesn't work with it since relinking occurs on the client (and there is no mechanism for relinking on the server), but at least Relink won't blast people's tiddlers away anymore.

Installing TiddlyMap also will no longer delete literally everything.

By the way, guys. Lazy-loading should really be deprecated. Even with this bug fix, it's still disastrously dangerous.

### DIFF
--- a/core/modules/server/routes/put-tiddler.js
+++ b/core/modules/server/routes/put-tiddler.js
@@ -30,6 +30,16 @@ exports.handler = function(request,response,state) {
 	if(fields.revision) {
 		delete fields.revision;
 	}
+	// If this is a skinny tiddler, it means the client never got the full
+	// version of the tiddler to edit. So we must preserve whatever text
+	// already exists on the server, or else we'll inadvertently delete it.
+	if(fields._is_skinny !== undefined) {
+		var tiddler = state.wiki.getTiddler(title);
+		if(tiddler) {
+			fields.text = tiddler.fields.text;
+		}
+		delete fields._is_skinny;
+	}
 	state.wiki.addTiddler(new $tw.Tiddler(fields,{title: title}));
 	var changeCount = state.wiki.getChangeCount(title).toString();
 	response.writeHead(204, "OK",{


Jermolene:
Thank you @flibbles that seems a reasonable workaround.

core/modules/server/routes/put-tiddler.js
```diff
@@ -30,6 +30,16 @@ exports.handler = function(request,response,state) {
 	if(fields.revision) {
 		delete fields.revision;
 	}
+	// If this is a skinny tiddler, it means the client never got the full
+	// version of the tiddler to edit. So we must preserve whatever text
+	// already exists on the server, or else we'll inadvertently delete it.
+	if(fields._is_skinny !== undefined) {
+		var tiddler = state.wiki.getTiddler(title);
+		if(tiddler) {
+			fields.text = tiddler.fields.text;
```
linonetwo, line 39: What about other fields, like `tags` or `tmap.id` ?

flibbles: What *about* them? They are always loaded, regardless of lazy loading. This fix was to prevent body deletion when either of those two fields are added or modified.

linonetwo: Oh, I forget that. Then lazy loading won't work for me... I have 10k+ tiddler without text, only calendar event related metadata fields...

pmario: If they don't have any text, then you shouldn't worry about size anyway, except if you store images in the fields.

PR #7020, opened 2022-11-01
Update + rename Importing Tiddlers.tid

twMat:
Other than some sprucing up, this proposal should clarify that it really is about importing arbitrary content, not just "tiddlers". For this reason I'm also changing the misleading title into just "Importing" (I was considering "Importing content" but what else is there to import?)

BTW, I hope my use of the term "*convert* into tiddlers" is correct here.

### DIFF
--- a/editions/tw5.com/tiddlers/features/Importing Tiddlers.tid
+++ b/editions/tw5.com/tiddlers/features/Importing Tiddlers.tid
@@ -1,26 +1,28 @@
 created: 20130825162000000
 modified: 20170328160211362
 tags: Features
-title: Importing Tiddlers
+title: Importing
 type: text/vnd.tiddlywiki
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @twMat 

Changing the titles of docs tiddlers is problematic. It breaks any existing links that might be out there. In this case, there are 4 backlinks to "Importing Tiddlers' that need to be updated too.

I'd recommend sticking with the existing title. If we really want to change it then we should (a) correct the backlinks and (b) keep a tiddler called "Importing Tiddlers" that just contains"See [[Importing]]" (that's the technique I've used in #6666 where I've made a number of rationalisations to the transclusion and macro docs.

pmario:
@twMat ... Are you able to adjust as requested at: https://github.com/TiddlyWiki/TiddlyWiki5/pull/7020#issuecomment-1314085332

PR #7028, opened 2022-11-09
Add parser to display MS Office documents inside tiddler body using iframe

talha131:
This PR adds a parser `DocxParser`

# What does it do?

1. If a user adds a tiddler with MS Office type like `application/msword`
2. AND sets its `_canonical_uri` to a word document

The the parser uses MS live.com embed feature to display those documents.

# Examples

<details>
<summary>
DOCX example
</summary>

</details>

<details>
<summary>
DOC example
</summary>

</details>

<details>
<summary>
XLSX example
</summary>

</details>

<details>
<summary>
XLS example
</summary>

</details>

<details>
<summary>
PPTX example
</summary>

</details>

<details>
<summary>
PPT example
</summary>

</details>

# Why does this PR change `BinaryParser`?

These changes are extremely minor. They export `warn` and `link` variables, so that other components can use them if needed.

In my case, `DocxParser` needed them to handle the following usecase,

1. User adds a word doc to the tiddlywiki
2. AND user has not set the `_canonical_uri`

In this case, the word doc is added as a binary data to the TiddlyWiki. MS Office live viewer cannot display word documents from the binary data. 

So the `DocxParser` shows binary data warning.

<details>
<summary>
Binary data warning example
</summary>

</details>

### DIFF
--- a/core/modules/parsers/binaryparser.js
+++ b/core/modules/parsers/binaryparser.js
@@ -14,37 +14,38 @@ The binary parser parses a binary tiddler into a warning message and download li
   var BINARY_WARNING_MESSAGE = "$:/core/ui/BinaryWarning";
   var EXPORT_BUTTON_IMAGE = "$:/core/images/export-button";
 
-  var BinaryParser = function (type, text, options) {
-    // Transclude the binary data tiddler warning message
-    var warn = {
-      type: "element",
-      tag: "p",
-      children: [
-        {
-          type: "transclude",
-          attributes: {
-            tiddler: { type: "string", value: BINARY_WARNING_MESSAGE },
-          },
+  // Transclude the binary data tiddler warning message
+  var warn = {
+    type: "element",
+    tag: "p",
+    children: [
+      {
+        type: "transclude",
+        attributes: {
+          tiddler: { type: "string", value: BINARY_WARNING_MESSAGE },
         },
-      ],
-    };
-    // Create download link based on binary tiddler title
-    var link = {
-      type: "element",
-      tag: "a",
-      attributes: {
-        title: { type: "indirect", textReference: "!!title" },
-        download: { type: "indirect", textReference: "!!title" },
       },
-      children: [
-        {
-          type: "transclude",
-          attributes: {
-            tiddler: { type: "string", value: EXPORT_BUTTON_IMAGE },
-          },
+    ],
+  };
+  // Create download link based on binary tiddler title
+  var link = {
+    type: "element",
+    tag: "a",
+    attributes: {
+      title: { type: "indirect", textReference: "!!title" },
+      download: { type: "indirect", textReference: "!!title" },
+    },
+    children: [
+      {
+        type: "transclude",
+        attributes: {
+          tiddler: { type: "string", value: EXPORT_BUTTON_IMAGE },
         },
-      ],
-    };
+      },
+    ],
+  };
+
+  var BinaryParser = function (type, text, options) {
     // Set the link href to external or internal data URI
     if (options._canonical_uri) {
       link.attributes.href = {
@@ -69,5 +70,7 @@ The binary parser parses a binary tiddler into a warning message and download li
     this.tree = [element];
   };
 
+  exports.BinaryParserWarn = warn;
+  exports.BinaryParserLink = link;
   exports["application/octet-stream"] = BinaryParser;
 })();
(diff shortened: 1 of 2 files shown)


saqimtiaz:
Hi @talha131, happy to see a PR from you. Personally my concern with this proposed change is that it introduces a dependency on an external service, i.e. view.officeapps.live.com. We tend to avoid such dependencies on the TiddlyWiki core from the perspective of long term maintainability and there are privacy concerns to consider as well.

You should wait to hear from @Jermolene, however my personal opinion is that this is better suited as a plugin. Plugins offer the possibility to explore things that aren't quite appropriate for the longer term vision for the core but still offer considerable utility.

talha131:
Thank you @saqimtiaz for taking the time to review.

I agree that this patch would add a dependency on an external service, which means this feature will not work offline and adds privacy concerns.

I did publish a [plugin](https://talk.tiddlywiki.org/t/a-macro-to-embed-docx-files-into-your-tiddlers/4899) first. (I need some free time to package it properly). But there was [some interest](https://talk.tiddlywiki.org/t/a-macro-to-embed-docx-files-into-your-tiddlers/4899/6?u=talha131) in having feature parity with the built-in PDF viewer, which spurred me to look into the parsers.

> Somehow with pdf, you don't have to do anything at all in the text field ("body" of tiddler); tiddlywiki swaps in the viewtemplate needed for pdf as soon as _canonical_uri is set. 



Let's see what @Jermolene has to say.

pmario:
@talha131 ... Looks very interesting. ... nice work!

A also think, it should be a plugin. .. @Jermolene ... I think it should be a core-plugin. ... MS still powers the majority of all desktops. I think it would add some value to the TW "product".

pmario:
> 2. It cannot work without _canonical_uri. There is no straightforward way to display MS Office docs from binary data.

I think this can be solved with documentation. .. The same is true for the "online dependency"

talha131:
So guys, especially @Jermolene what's the conclusion here?

1. Add this patch as is?
2. Add this patch as an official plugin?
3. Close this PR and publish it as an external plugin

Jermolene:
Hi @talha131 than you for your patience.

It is indeed not acceptable to have a dependency on an external web service in the core.

I think instead we need to look at extending the binary parser to introduce logic to allow users to customise the handling of particular mime types.

Jermolene:
A possible approach might be for the binary parser to generate a parse tree that looks for a macro/procedure titled `tv-binary-parser/` concatenated with the type, and if it exists then invoke it. If it doesn't exist, just render the existing skeleton.

So, the parse tree generated by the binary parser would be equivalent to:

```
<$let handlerName={{{ [<type>addprefix[tv-binary-parser/]] }}}>
<$list filter="[<handlerName>is[variable]]" variable="ignore">
<$macrocall $name=<<handlerName>> type=<<type>> text=<<text>> _canonical_uri=<<_canonical_uri>> 
</$list>
<$list filter="[<handlerName>!is[variable]]" variable="ignore">
<div class="tc-binary-warning>
<p>
<$transclude tiddler="$:/core/ui/BinaryWarning"/>
<$let dataURI={{{ [[data:]addsuffix<type>addsuffix[;base64,]addsuffix<text>] }}}>
<a title={{!!title}} download={{!!title}} href={{{ [<currentTiddler>has[_canonical_uri]then{!!_canonical_uri}else<dataURI>] }}}>
<$transclude tiddler="$:/core/images/export-button"/>
</a>
<$let>
</p>
</div>
</$list>
</$let>
```

To override DOCX file rendering, you'd create a global macro called `tv-binary-parser/application/msword` with parameters for the type, text and _canonical_uri.

pmario:
>To override DOCX file rendering, you'd create a global macro called tv-binary-parser/application/msword with parameters for the type, text and _canonical_uri.

@Jermolene ... is the performance hit for "global macros" mentioned here solved already?

pmario:
>I think you are referring to the implementation of global macros that was briefly part of https://github.com/Jermolene/TiddlyWiki5/pull/6666; it is no longer included in that PR. I did resolve the performance issues, but decided to remove the feature to keep the PR simpler to understand as the features are independent.

You are right. 

@talha131 .. sorry for derailing the communication here. IMO you can ignore the comment.

pmario:
Needs to be a plugin. live-dot-com dependency can only be used in plugins.

Issue #7037, opened 2022-11-15
[BUG] Issue in UI Text for New Journal Tiddlers and Default Tiddlers


kookma:
### Describe the bug

The latest prerelease of 2022.11.15 shows issue for input boxes of 

* Text for New Journal Tiddlers
* Default Tiddlers

It seems the width set to auto

### Expected behavior

They need to have full width of parent!

### To Reproduce

1. Go to https://tiddlywiki.com/prerelease/
2. Open $:/ControlPanel
3. Info -> Basic tab

### Screenshots

![image](https://user-images.githubusercontent.com/830394/201822458-3dd23f1f-8e68-4cda-bc01-3df10f65d759.png)

### TiddlyWiki Configuration

Tiddlywiki 5.24 pre
Edge 107
Win 10

Jermolene:
Thanks @kookma – @pmario is this is related to 10cb585dae7e71bd2c9969768307989828f56f52?

ericshulman:
In TW5.2.3, $:/core/ui/ControlPanel/Basics, the textarea controls have ` class="tc-edit-texteditor"`.

In TW5.2.4prerelease, these class parameters were removed by change [10cb585](https://github.com/Jermolene/TiddlyWiki5/commit/10cb585dae7e71bd2c9969768307989828f56f52).

As a result, these textareas no longer fill the width of their containing table cells, as reported by @kookma

Jermolene:
Oh dear. I'll revert #6912 and comment more there.

pmario:
That's probably not the problem. I did download master and merged this PR. It looks like this.

![image](https://user-images.githubusercontent.com/374655/201921074-411950a1-b4bc-42f6-8ea0-7d5817671e39.png)

pmario:
There seems to be a problem with the "hardcoded" styles that are applied to the text input area ... may be a codemirror problem needs more investigation

pmario:
@Jermolene ... The actual problem is, that the 2 text areas used in $:/core/ui/ControlPanel/Basics are converted to codemirror editors, if the plugin is installed. I personally think that's a bug, but for backwards compatibility reasons I'll leave it that way.

They use the `$edit` widget instead of the `$edit-text` widget, which the rest of the table uses. So they are converted to cm-editors ... 

PR fix will follow shortly

pmario:
@kookma -- IMO the problems described in the OP have been fixed. -- right?

kookma:
[...]

PR #6912, opened 2022-08-18
allow $:/core/ui/ControlPanel/Basics to work in the story river + some docs about utility classes

pmario:
## For reviewers

See: https://pmario.github.io/kitchensink/6912-table-utility-classes.html

- **Please check the wording in the docs text** 

## To reproduce: 

- open https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FBasics 
- the text input elements don't have the right size
- The second problem with tables in general. `code snippets` clash with table borders

<details open><summary>Screenshot -- Problem with ControlPanel table in story river</summary>

image

</details>

The second problem with tables in general

<details open><summary>Screenshot -- Code snippets in tables clash with table border</summary>

image

</details>

## This PR

- This PR allows the $:/core/ui/ControlPanel/Basics tiddler to work in the story river. 
- The screenshot also shows the Basics tiddler in the sidebar, to prove that the used classes also work there. 
- The PR also contains 3 doc tiddlers that have been updated with the info needed to understand the concept
- This PR should also deal with the second `code snippet` problem described above. 

- The tiddler:  editions/tw5.com/tiddlers/wikitext/Tables in WikiText Utility Classes.tid  contains a lot of whitespace, that will be removed. **It only makes editing with the preview ON easier.**

## Tested with 

- FF 105.03 Windows 11
- Edge Version 106.0.1370.37 Win 11
- FF 105.2.0 on Android 13
- Chrome on Android 13

This PR should also allow users to create table based "input forms" similar to those we use in the ControlPanel. IMO user forms should be consistent with the current theme. 

image

PR #7038, opened 2022-11-15
Table utility classes

pmario:
This PR fixes:  **[BUG] Issue in UI Text for New Journal Tiddlers and Default Tiddlers** #7037 

- this PR adds back the `tc-edit-texteditor` class
- it removes the `tag="textarea"`  parameter from the [<$edit widget](https://tiddlywiki.com/#EditWidget), since it does not support it and it causes confusion
- remove StoryList from PR

------

The rest stays the same as the original PR see: https://github.com/Jermolene/TiddlyWiki5/pull/6912

I did test it with 

- FF latest on windows
- Edge latest on windows
- FF on Android
- chrome on Android. 

Works for me.


editions/tw5.com/tiddlers/$__StoryList.tid
```diff
@@ -1,5 +0,0 @@
-created: 20220728145919904
-list: Brackets Base64 HelloThere GettingStarted Community
-modified: 20220728145919904
-title: $:/StoryList
-type: text/vnd.tiddlywiki
```
saqimtiaz, line 5: This should be a separate PR in order to have a cleaner and easier to follow commit history.

saqimtiaz:
> * it removes the `tag="textarea"`  parameter from the [<$edit widget](https://tiddlywiki.com/#EditWidget), since it does not support it and it causes confusion

Can you please explain what you mean here? The $edit widget passes on all attributes to the $edit-text widget it calls. 

Are you proposing that we not use textareas but use input fields instead in ControlPanel/Basics?

editions/tw5.com/tiddlers/$__StoryList.tid
pmario: So you say, I should add it back and pollute the commit history even more, just to remove it with a separate PR? IMO that doesn't make sense to me

saqimtiaz: Your original PR has been reverted and you are making a change unrelated to this PR. So yes, it does make sense. The story list tiddler needs to be amended to undo what is probably an erroneous commit by Jeremy but does not need to be deleted.

PR #7039, opened 2022-11-15
Allow ControlPanel/Basics to be opened in the story-river - Add table utility classes and some docs how to use them 

pmario:
This PR fixes:  **[BUG] Issue in UI Text for New Journal Tiddlers and Default Tiddlers** #7037 

- this PR adds back the `tc-edit-texteditor` class. Needed by the `$edit` widget.
- it removes the `tag="textarea"`  parameter from the [<$edit widget](https://tiddlywiki.com/#EditWidget). It's not needed since the specified field is text, so it creates a text area by default

------

The rest stays the same as the original PR see:  **allow $:/core/ui/ControlPanel/Basics to work in the story river + some docs about utility classes** #6912 

I did test it with 

- FF latest on windows
- Edge latest on windows
- FF on Android
- chrome on Android. 

Works for me.

### DIFF
--- a/core/ui/ControlPanel/Basics.tid
+++ b/core/ui/ControlPanel/Basics.tid
@@ -20,15 +20,16 @@ caption: {{$:/language/ControlPanel/Basics/Caption}}
 \end
 \whitespace trim
 
+|tc-max-width tc-edit-max-width|k
 |<<lingo Version/Prompt>> |''<<version>>'' |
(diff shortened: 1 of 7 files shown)


pmario:
@Jermolene bump

pmario:
@Jermolene ... It would be really great, if this PR would get into 5.2.4. .. The last problem reported was due to missing tests with CodeMirror. 

Please check again.

Jermolene:
Thanks @pmario

PR #7042, opened 2022-11-16
CSV parser improvements

EvidentlyCube:
Inspired by this thread: https://talk.tiddlywiki.org/t/csv-problem-with-strings-in-quotation-marks/5144/6

# Preview

**See it live here:** https://new-csv-parser-pr.tiddlyhost.com/
Since it was only a few files I created a Tiddlyhost page which contains the changes.

# Changes

I've added a new method for parsing CSV text that should be compatible with all the edge cases of [RFC 4180](https://www.rfc-editor.org/rfc/rfc4180), specifically:
 * Quoted cells
 * Quoted cells with quotes inside
 * Quoted cells with separator inside
 * Quoted cells with newline characters

Then I added test cases, changed the existing `parseCsvStringWithHeader` method to leverage the new one and made some improvements to how CSV tables are displayed:
* They display newlines
* They always have enough columns to fit the longest row
* They add blank cells when the row does not have enough columns

# Important notes

This makes some backwards incompatible changes because, well, CSV files will now be parsed correctly. If someone depended on the wrong behavior... I'd say "tough luck" and I doubt there are more than a handful of users whom it will affect negatively.

I wrote the parser myself from the scratch so it doesn't depend on any other parser - I briefly looked at `mafintosh/csv-parser` but spent a total of about a minute looking at its code before deciding I'd rather write it all by myself.

### DIFF
--- a/core/modules/utils/csv.js
+++ b/core/modules/utils/csv.js
@@ -12,35 +12,109 @@ A barebones CSV parser
 /*global $tw: false */
 "use strict";
 
+var QUOTE = '"';
+
+var getCellInfo = function(text, start, length, SEPARATOR) {
+	var isCellQuoted = text.charAt(start) === QUOTE;
+	var cellStart = isCellQuoted ? start + 1 : start;
+	
+	if (text.charAt(i) === SEPARATOR) {
+		return [cellStart, cellStart, false];
+	}
+	
+	for (var i = cellStart; i < length; i++) {
+		var cellCharacter = text.charAt(i);
+		var isEOL = cellCharacter === "\n" || cellCharacter === "\r";
+		
+		if (isEOL && !isCellQuoted) {
+			return [cellStart, i, false];
+			
+		} else if (cellCharacter === SEPARATOR && !isCellQuoted) {
+			return [cellStart, i, false];
+			
+		} else if (cellCharacter === QUOTE && isCellQuoted) {
+			var nextCharacter = i + 1 < length ? text.charAt(i + 1) : '';
+			if (nextCharacter !== QUOTE) {
+				return [cellStart, i, true];
+			} else {
+				i++;
+			}
+		}
+	}
+	
+	return [cellStart, i, isCellQuoted];
+}
+	
+exports.parseCsvString = function(text, options) {
+	options = options || {};
+	var SEPARATOR = options.separator || ",",
+		length = text.length,
+		rows = [],
+		nextRow = [];
+		
+	for (var i = 0; i < length; i++) {
+		var cellInfo = getCellInfo(text, i, length, SEPARATOR);
+		var cellText = text.substring(cellInfo[0], cellInfo[1]);
+		if (cellInfo[2]) {
+			cellText = cellText.replace(/""/g, '"');
+			cellInfo[1]++;
+		}
+		nextRow.push(cellText);
+		
+		i = cellInfo[1];
+		
+		var character = text.charAt(i);
+		var nextCharacter = i + 1 < length ? text.charAt(i + 1) : '';
+		
+		if (character === "\r" || character === "\n") {
+			// Edge case for empty rows
+			if (nextRow.length === 1 && nextRow[0] === '') {
+				nextRow.length = 0;
+			}
+			rows.push(nextRow);
+			nextRow = [];
+			
+			if (character === "\r") {
+				var nextCharacter = i + 1 < length ? text.charAt(i + 1) : '';
+				
+				if (nextCharacter === "\n") {
+					i++;
+				}
+			}
+		}
+	}
+	
+	// Special case if last cell in last row is an empty cell
+	if (text.charAt(length - 1) === SEPARATOR) {
+		nextRow.push("");
+	}
+	
+	rows.push(nextRow);
+	
+	return rows;
+}
+
 /*
 Parse a CSV string with a header row and return an array of hashmaps.
 */
 exports.parseCsvStringWithHeader = function(text,options) {
-	options = options || {};
-	var separator = options.separator || ",",
-		rows = text.split(/\r?\n/mg).map(function(row) {
-			return $tw.utils.trim(row);
-		}).filter(function(row) {
-			return row !== "";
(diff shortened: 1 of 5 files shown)


kookma:
Hi @EvidentlyCube 
Great PR. Just a question: is it possible to add few features like
* interactive table header to click and sort by column?
* apply formatting per cell?

Please ignore if not relevant. My ideas come from csv table in Shiraz.

editions/test/tiddlers/tests/data/csv-cases.tid
```diff
@@ -0,0 +1,202 @@
...
+        ],
+        "jsonWithHeaders": [
+            {"cell-11": "cell-21", "cell-12": "cell-22", "cell-13": "cell-23"},
+            {"cell-11": "cell-31", "cell-12": "cell-32", "cell-13": "cell-33"}
+        ]
+    },
+    {
+        "name": "Custom separator",
+        "options": {"separator": "\t"},
+        "csv": "cell-11\tcell-12\tcell-13\r\ncell-21\tcell-22\tcell-23\r\ncell-31\tcell-32\tcell-33",
```
cdaven, line 23: You could add some "," to the cells' contents, to make it clear that it's no longer a separator.

core/modules/utils/csv.js
```diff
@@ -12,35 +12,109 @@ A barebones CSV parser
...
+			}
+		}
+	}
+	
+	return [cellStart, i, isCellQuoted];
+}
+	
+exports.parseCsvString = function(text, options) {
+	options = options || {};
+	var SEPARATOR = options.separator || ",",
```
cdaven, line 50: Maybe completely unnecessary, but you _could_ throw an error if `options.separator` contains a double quote, since that is a hard-coded, "magic" character in the parsing.

cdaven: Also, what happens if you use \r or \n as separators?

cdaven:
It looks really well done!

Jermolene:
Thanks @EvidentlyCube great stuff. I wondered if it might be possible for the default to autodetect CSV vs. TSV, as I think Excel does?

To answer @kookma's question, it would be possible to extend the CSV parser to have sortable columns, but that would still mean that the data in the CSV is locked up and inaccessible. I think a better next step might be to explore filter operators that could give access to the raw data in CSV tiddlers, so that table sorting can be implemented in wikitext.

For example, we could have a `csvgetrow[rowno]` that retrieves a row as a list, or `csvgetcell[col][rowno]` to get the value of an individual cell. Behind the scenes, we could use the same caching mechanism as data tiddlers to improve efficiency.

kookma:
> For example, we could have a `csvgetrow[rowno]` that retrieves a row as a list, or `csvgetcell[col][rowno]` to get the value of an individual cell. Behind the scenes, we could use the same caching mechanism as data tiddlers to improve efficiency.

This is amazing and will give a lot of flexibility.

core/modules/utils/csv.js
EvidentlyCube: >  you could throw an error
I feel like TW's way is to just fail gracefully rather than throw exceptions, and in that context the best option would probably be to just parse it that way and return whatever mess you get. I'll add a test case for that just to make sure no exception is thrown :).

Should also add a test case for empty input and such!

editions/test/tiddlers/tests/data/csv-cases.tid
EvidentlyCube: Good idea!

EvidentlyCube:
I just pushed a few changes:
 * Support for Tab delimited mimetype
 * More test cases
 * Graceful handling of empty and null input

I've also updated the tiddlyhost preview with new changes.

pmario:
I only a brief view at the code. There seems to be a lot of CR `\r` handling. TW internally typically completely removes CRs and then hasn't do deal with it. .... Just some thoughts. 

All the output created by TW contains LF only

EvidentlyCube:
> There seems to be a lot of CR \r handling.

@pmario That's because the spec explicitly states CRLF is used for line breaks. I briefly had a version of the code that cleaned up the line endings to only LF but even if it made the code slightly simpler I wanted to work with the spec and not mess up someone's content (the line endings can appear inside the cell after all).

Are you sure all tiddlers text gets sanitized though? I just checked on latest empty and it stores `\r\n` as-is:
image

pmario:
>Are you sure all tiddlers text gets sanitized though? I just checked on latest empty and it stores 

Nothing gets sanitized on the Tiddler() level. But if exported or rendered as HTML there should be some filters in place. 

If CR would be part of the HTML file on windows it would be significantly bigger and the windows html and linius html would be different ... which shouldn't happen.

pmario:
> I wanted to work with the spec and not mess up someone's content 

That's an argument. OK

pmario:
Everything works as expected. If you use TW js functions to force CRLF into the text field, I think you expect them to be there after a file reload. 

Forget about the CR handling that I mentioned. I think it isn't important for your PR here. You want to preserve the CR character and it's OK that way. 

If the CVS is considered the "source of truth" it should be preserved.

Jermolene:
Thanks @EvidentlyCube

pmario:
@EvidentlyCube ... There seems to be a push to your repository, with some changes related to this PR **after** it has been merged by Jeremy. 

Did you ever create a new PR to add your changes to the core? Or is this PR what you pushed to your repo?

EvidentlyCube:
> @EvidentlyCube ... There seems to be a push to your repository, with some changes related to this PR **after** it has been merged by Jeremy.
> 
> Did you ever create a new PR to add your changes to the core? Or is this PR what you pushed to your repo?

I don't beleive there's been a change - [last commit is Novembber 17](https://github.com/EvidentlyCube/TiddlyWiki5/commits/csv-parser-improvs/):

image

And merge happened on November 20:
image

----

Are you perhaps referencing to this?

image

It seems someone has cherry picked my commit into their TW fork.

pmario:
Ahh. OK

Issue #280, opened 2013-12-14
suggestion: store modified, created dates as UTC ISO string


pmario:
see: https://groups.google.com/forum/?fromgroups=#!topic/tiddlywiki/SOEPWuxejF8

Would it make sense to store the TW dates as UTC ISO strings. They are human readable and precise down to the milisecond. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

what do you think?

pmario:
This issue was raised prior to TW v5.1.0 where we did not have many filter operators and changes like this one would have been possible. Now it would break backwards compatibility.

So closing

PR #7043, opened 2022-11-17
Add Timestamp to DateFormat

EvidentlyCube:
# What:

* Adds `TIMESTAMP` as a new option to Date Format which returns the timestamp. 
* Test case included

# Why:

* I got bit by the difference between UTC and non-UTC dates while developing something (the problem mentioned here: https://github.com/Jermolene/TiddlyWiki5/issues/280#issuecomment-36069270)
* Timestamp makes it possible to compute the difference between two dates
   * Which, for example, allows displaying dates in the relative format if someone wishes to
* Timestamps are ubiquitous and it's a weird hole in the capabilities of TiddlyWiki

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -354,6 +354,9 @@ exports.formatDateString = function(date,template) {
 	var result = "",
 		t = template,
 		matches = [
+			[/^TIMESTAMP/, function() {
+				return date.getTime();
+			}],
 			[/^0hh12/, function() {
 				return $tw.utils.pad($tw.utils.getHours12(date));
 			}],
--- a/editions/test/tiddlers/tests/test-utils.js
+++ b/editions/test/tiddlers/tests/test-utils.js
@@ -78,6 +78,7 @@ describe("Utility tests", function() {
 		expect(fds(d,"ddd hh mm ssss")).toBe("Sun 17 41 2828");
 		expect(fds(d,"MM0DD")).toBe("1109");
 		expect(fds(d,"MM0\\D\\D")).toBe("110DD");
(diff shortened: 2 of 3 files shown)


saqimtiaz:
I would recommend linking to an explanation of what the ECMAScript/Unix Epoch is, since the primary target audience is non-developers. It would also be very helpful to add some documentation about how to use a timestamp to compute the difference between dates, but that does not need to be a part of this PR.

Jermolene:
Thanks @EvidentlyCube - it would indeed be helpful to add that it is the number of milliseconds since 1 January 1970 UTC, and a link to Wikipedia.

EvidentlyCube:
**I've improved the documentation:**

* Timestamp now contains a "New in 5.2.4" pill
* Contains a link to description of epoch time
* And specifically tells what's the epoch date
image

And added an example on how to use timestamp to calculate the difference between two dates:
image
English is not my native language so there is likely room for improvement, especially around articles and punctuation

kookma:
Hi @EvidentlyCube 
Please correct the documentation (https://tiddlywiki.com/prerelease/#DateFormat) for `YYYYMMDDHHMMSSmmm` 
This is not the TiddlyWiki correct date format, for example milliseconds are `XXX`

Jermolene:
Hi @kookma I think the context here is not a TiddlyWiki date format string, but rather an attempt to describe the format itself in simple terms.

EvidentlyCube:
@kookma Is right, I mindlessly copied this from the comment above `$tw.utils.parseDate`. It makes more sense to use TW's Date Format here since we're in the page where the format can be inspected. I'll make a PR in a second.

PR #7045, opened 2022-11-18
Additional Polish translation tweaks

EvidentlyCube:
I went through the whole control panel fixing typos and improving language, especially with things I understand much better now.


Jermolene:
Thanks @EvidentlyCube

PR #7047, opened 2022-11-19
Sort completed tasks by 'modified' instead of 'created' field

talha131:
see details here 

https://talk.tiddlywiki.org/t/a-probable-mistake-in-the-official-documentations-task-management-example/5198

The test claims, completed tasks are listed in the order of completion. This can only be done if tasks are sorted by `modified`.


Jermolene:
Thanks @talha131

PR #7049, opened 2022-11-20
add a hidden config to disable Syncer logging to the console

EvidentlyCube:
# What:

* Adds a hidden/undocumented config option `$:/config/SyncLogging` that allows disabling Syncer's logging

# Why:

* Currently using Node server with plugins `plugins/tiddlywiki/filesystem` and `plugins/tiddlywiki/tiddlyweb` will happily log up to 3 console logs each time any tiddler is changed.
* This makes using console for plugin development a chore, as any relevant info you log will be interspersed with Syncer logs.
* I opted for a hidden setting because developer console is already something that non-developers are unlikely to use.

----

**Logs created after changing a typing a single character in a tiddler's title:**
image

### DIFF
--- a/core/modules/startup/startup.js
+++ b/core/modules/startup/startup.js
@@ -121,7 +121,11 @@ exports.startup = function() {
 	});
 	// Set up the syncer object if we've got a syncadaptor
 	if($tw.syncadaptor) {
-		$tw.syncer = new $tw.Syncer({wiki: $tw.wiki, syncadaptor: $tw.syncadaptor});
+		$tw.syncer = new $tw.Syncer({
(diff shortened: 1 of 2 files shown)


pmario:
The PR seems to contain a doc tiddler. So it's a "documented" but hidden-setting.

I'd highly support that feature. ... Will do some testing now

Jermolene:
Good stuff thanks @EvidentlyCube

pmario:
Cool Jeremy was faster ;) .. It's merged already .. did test it with FF on linux and it seems to do, what's expected. Thx

PR #7052, opened 2022-11-22
Update Filter Run Prefix (Examples)

twMat:
Added entry for `:map`

(Would be better if the text was generated but not sure how to show the title while it is between code-quotes, (e.g this `<$view field=title/>` will literally show like that when the desired output is that it shows the actual title styled like code).


Jermolene:
Thanks @twMat all of those tiddlers should indeed be listed here. As you note, the problem is that this listing is not automatically generated.

> (Would be better if the text was generated but not sure how to show the title while it is between code-quotes, (e.g this `<$view field=title/>` will literally show like that when the desired output is that it shows the actual title styled like code).

Try `<code><$view field="title"/></code>`.

twMat:
I'm retracting this PR because I think there are wider questions that need to be dealt with which, as outlined here:

https://talk.tiddlywiki.org/t/ranty-proposal-for-docs-on-filter-run-prefixes/5772

Issue #7056, opened 2022-11-23
[BUG] Visual regression in import table in 5.2.4 pre-release


saqimtiaz:
As reported on [talk,](https://talk.tiddlywiki.org/t/tiddlywiki-5-2-4p-issue-in-import-table/5259) the table rows in the $:/Import tiddler have gaps between them:
![image](https://user-images.githubusercontent.com/68092/203480777-52c799d7-48b2-4c23-ba98-49c26b23be7b.png)

The change seems to be caused by #7039 and in particular the following change in CSS:
https://github.com/Jermolene/TiddlyWiki5/pull/7039/files#diff-1608798fd26545169d67db97f5f4b2ed06e74bdf30c65b70d95482dddbbb06feR328-R331

Jermolene:
Thanks @saqimtiaz – I'll revert #7039 for the moment. @pmario could you have a look at an alternative implementation that doesn't change the styling of existing classes/elements?

pmario:
I think the visual integrity of all tables in general is much more important than the import dialogue, that is not used every day. ... I think the import table needs to be fixed instead of everytime a visual glitch comes up to revert the whole PR. 

eg: have a look at the initial screenshots at: https://github.com/Jermolene/TiddlyWiki5/pull/6912#issue-1342948905
which the PR does fix.

Jermolene:
> It's not a problem of the PR it's a problem of the import table that included invisible rows instead of _not_ creating them. The PR only makes the problem visible.

For better or worse, that is how the import table has always worked, and we might expect that users will have copied it, so backwards compatibility is a concern.

> So everytime a problem shows up, my fixes get reverted instead of having a closer look at the "real" problem

The reason I tend to revert PRs immediately that we see a problem reported is to avoid other people seeing the same problem and reporting it again. It also gives the submitter more time to fix things.

PR #7057, opened 2022-11-23
fix problem with hidden preview DOM structure

pmario:
@kookma @saqimtiaz please test the wiki in the ZIP .. It contains fixes to issue #7056 AND it also contains the changes PR #7039, which IMO have been reverted without any need. 

 [fix-7056-import-visual-glitch.zip](https://github.com/Jermolene/TiddlyWiki5/files/10075092/fix-7056-import-visual-glitch.zip)  <------------------ Test me

This is the fix to the real problem that my PR #7039 made visible. ... 

The problem was the "preview element" it added visibe TR and TD elements to the DOM with a hidden content DIV. Since the TD had some padding from #7039 it was visible. 

This PR fixes:  [BUG] Visual regression in import table in 5.2.4 pre-release #7056 .

---------
@Jermolene please reapply PR #7039 and then this PR. 

There should be no visual glitches anymore. My PR didn't cause the problem it made it visible

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -117,15 +117,15 @@ title: $:/core/ui/ImportListing
 		</div>
 	</td>
 </$reveal>
-<tr>
+<$reveal type="match" text="yes" state=<<previewPopupState>> tag="tr">
 <td colspan="3">
-<$reveal type="match" text="yes" state=<<previewPopupState>> tag="div">
 <$list filter="[{$:/state/importpreviewtype}has[text]]" variable="listItem" emptyMessage={{$:/core/ui/ImportPreviews/Text}}>
-<$transclude tiddler={{$:/state/importpreviewtype}}/>
+	<div>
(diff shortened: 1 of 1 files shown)


pmario:
The wiki in the ZIP should look like this: 

image

kookma:
Hi Mario

New fix 5.2.4p
image


Old 5.2.3
image


I think the row height is different and new fix creates taller rows. Other properties seem the same.

Jermolene:
Thanks @pmario @kookma 

> @Jermolene please reapply PR https://github.com/Jermolene/TiddlyWiki5/pull/7039 and then this PR.

Done

pmario:
>I think the row height is different and new fix creates taller rows. Other properties seem the same.

It's intended, that the row hight is higher. See the problems in the tables described at: https://github.com/Jermolene/TiddlyWiki5/pull/6912

Got to tiddlywiki.com and have a look at all the control-panel tables. All the spacing in the content edit inputs is way to small. It doesn't look good, which has been pointed out several times here at GH and at Talk

PR #7060, opened 2022-11-24
update documentation to fix a mistake with format used by `created` and `modified`

EvidentlyCube:
Fixes issue with documentation mentioned by @kookma in [here](https://github.com/Jermolene/TiddlyWiki5/pull/7043#issuecomment-1325958383).

I think it makes more sense to use TW's date format in the example on the Date Format than some arbitrary mental shortcut format that might be difficult to understand for non-developers.

The new format was taken from `utils.js::formatDateString()` so it hopefully should be correct!


Jermolene:
Thanks @EvidentlyCube

PR #7061, opened 2022-11-25
Fix redundant p tags

pmario:


pmario:
.

pmario:
.

pmario:
.

pmario:
.

pmario:
.

pmario:
This PR is related to: https://github.com/Jermolene/TiddlyWiki5/pull/6156

Alzacon:
The related issues to the change in the parse can be grouped in two point:
- Some widgets return directly text or inline element, their output should be wrapped in the nearest parent that has block element, if it doesn't have parent but it is between break line it would be wrapped in a block element. This could continue be a `p` element (or a `div`), how it works in the wiki syntax. The same case is the macros that generate link, for example the `.operator-examples` macro.
- `listwidget` is a special case, it can return a list of text elements or widgets. Its output should be wrapped in a `div` (to contain other generated block elements). If its contain will work how I comment in the previous point, there shouldn't be relative issues.


The other issues are related a current macros.
- `.tip` and `.warning` macros have the same issue but in the current version the margin of p, that is wrapping the link, hides the problem, but it can also appreciate the problem in the gray border. If the output of `.operator-examples` macro was wrapped in a p element, the issue is hidden again. The fix will be fit the min-height of `.doc-icon-block` to the image. The `thumbnail` macro has the same problem and fix.
- `wikitext-example` macro works because there is a weird behaviour of the empty p that is followed the pre, their margin collapse to 0.

rmunn:
I was just dealing with a problem trying to make a grid-based layout for editions/tw5.com, which includes the menubar plugin. The menubar is transcluded and has an extraneous `<p>` element that I could not get rid of, which is a problem because the rules for `layout: grid` say that the elements in the grid must be direct children. I was able to solve that in the browser dev tools by adding `display: contents` to the extraneous `<p>` tag, but of course it's not possible to use that as a generic solution (not all `<p>` tags added by TW should have `display: contents`, in fact most of them shouldn't).

But I added the 5-line change to wikiparser.js to my local copy of TW, and voila: the unwanted `<p>` tag was gone, and the menubar was part of the layout the way it should be (as a direct child of the grid element).

So this PR, if polished up to something that makes all the tests pass, would solve a real problem I'm having that I don't currently see any other way to solve.

pmario:
>So this PR, if polished up to something that makes all the tests pass, would solve a real problem I'm having that I don't currently see any other way to solve.

The tests show "hidden problems". The fixes are relatively straight forward. It needs to be fixed in a theme CSS.

But that fix will make the whole thing not backwards compatible. So it will most likely cause problems with the vanilla theme. 

It may be possible to create a plugin, that overwrites the block-mode parser, but that's more or less the same as starting a "patch set" for TW. 

On the other hand, it would be a description, what needs to be changed, to fix the problems. So it's probably worth considering.

Issue #1674, opened 2015-04-23
Make "tc-vertical" optional in $:/core/ui/SideBar/More


rubaboo:
I'd like to have the tabs inside SideBar/More arranged horizontally. I'd rather have that extra width available for the tab content. There is no easy CSS to achieve that, and I'd rather not override the shadow $:/core/ui/SideBar/More to remove "tc-vertical".

Jermolene:
Hi @nameanyone that's a good idea. Would you be interested in working on a pull request?

nameanyone:
Sorry, I'm not that skilled yet. I've done some trivial submits before.

Jermolene:
Hi @nameanyone no problem, your contributions are much appreciated as is!

tobibeer:
Would it be sensibe to have an optional configuration tiddler right under that namespace, e.g.:

```
title: $:/core/ui/SideBar/More/horizontal
```

...which would do the desired magic via conditional variable asignment?

pmario:
IMO a little bit of new CSS should be enough to achieve this behaviour. There is no need for a new configuration tiddler and conditional layout handling. ... If CSS is not enough, its a bug.

pmario:
I had a closer look and I still think, it should be possible to solve it using CSS. .. But messing with the tiddler code is much easier today, than it has been 7 years ago. Today it's pretty much a 2 liner. 

Creating the docs and may be a config tiddler is probably 10 times more work. 

If [$:/core/ui/SideBar/More](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FSideBar%2FMore) is changed to the following code and `$:/config/ui/SideBar/More/horizontal` tiddler contains "yes" then it will show a horizontal view. 

```
\whitespace trim
<div class={{{ [{$:/config/ui/SideBar/More/horizontal}match[yes]then[tc-sidebar-tabs]else[tc-more-sidebar]] }}}>
<$macrocall $name="tabs" tabsList="[all[shadows+tiddlers]tag[$:/tags/MoreSideBar]!has[draft.of]]" default={{$:/config/DefaultMoreSidebarTab}} state="$:/state/tab/moresidebar" class={{{ [{$:/config/ui/SideBar/More/horizontal}match[yes]then[tc-sidebar-tabs-more]else[tc-vertical tc-sidebar-tabs-more]] }}} explicitState="$:/state/tab/moresidebar-1850697562"/>
</div>
```

This feature request is still valid.

[...]

PR #7063, opened 2022-11-25
More sidebar tabs horizontal

pmario:
This PR fixes: **Make "tc-vertical" optional in $:/core/ui/SideBar/More** #1674 ... And can be merged after TW v5.2.4 has been released. I just wanted to close the issue ;)

Edit: This setting is a Hidden Setting now. 

image


Jermolene:
Thanks @pmario 

I don't think it's worth exposing a user interface for this setting because of the cost of the additional translation strings, and the absence of evidence that significant numbers of people don't like the vertical tabs. From my perspective, I'm only considering this PR because in the discussion in #1674 it became clear that it couldn't be done in CSS.

pmario:
So you think it should be a hidden setting, which only needs a docs tiddler and no translation

Jermolene:
> So you think it should be a hidden setting, which only needs a docs tiddler and no translation

Exactly that, thank you @pmario, and apologies for the confusion.

pmario:
@Jermolene ... I did update the PR so it is a "Hidden Setting" now

Jermolene:
> @Jermolene ... I did update the PR so it is a "Hidden Setting" now

Thank you @pmario

PR #7064, opened 2022-11-25
allow the big download button to be defined using the colour palette values

### LINKED ISSUE #7041: [BUG] tiddlywiki.com style $:/_tw5.com-styles contains hardcoded colour so palette-manager doesn't work
[$:/_tw5.com-styles](https://tiddlywiki.com/prerelease/#%24%3A%2F_tw5.com-styles) contains the styling for the "Big Downloads" button. 

The colour is hardcoded so it cannot be defined by the palette-manager. -> **That's a problem**

### At the end of the file it contains

```
.tc-btn-download {
...
	background: #1462ff;
}

.tc-btn-download:active {
	background: #1462ff;
}
```

### It should contain

```
.tc-btn-download {
...
	background: <<colour "download-background">>;
}

.tc-btn-download:active {
	background: <<colour "download-active">>;
}
```

As shown in the screenshot below. .. The default colour with the Vanilla-Palette is "green"

At tiddlywiki.com it's hardcoded to blue-ish -> The blue colour also is a problem for the ControlPanel Plugin overview.

<details><summary><b> show screenshot</b> </summary>

image

</details>

# Request

- We should change the blue button to a green button
- Which will enable palette handling
- The ControlPanel plugin overview will be much more readable as it is atm

image

@Jermolene

- Should I create a PR for the green-ish button and check all the existing palettes ?
- Do we want a green-ish button as the download button

IMO The blue colour causes a problem in the plugin overview, so it should not be blue ... just some thoughts

pmario:
This PR fixes #7041 

It allows the download button to be defined by the colour palette value

image

### DIFF
--- a/editions/tw5.com/tiddlers/system/tw5.com-styles.tid
+++ b/editions/tw5.com/tiddlers/system/tw5.com-styles.tid
@@ -183,17 +183,17 @@ div.content {
 .tc-btn-download {
 	padding: 10px 30px;
 	border-radius: 5px;
-	background: #1462ff;
-        border: none;
-	box-shadow: 0 2px 2px 0 #4a74c9;
+	background: <<colour "download-background">>;
+	border: none;
+	box-shadow: 1px 2px 2px 0 <<colour muted-foreground>>;
 	overflow: hidden;
 	cursor: pointer;
-        font-size: 1.2em;
-        line-height: 1.4em;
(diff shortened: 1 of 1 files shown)


EvidentlyCube:
Stuff looks good from my perspective 👍

Jermolene:
Thanks @pmario

PR #7065, opened 2022-11-26
Fix issue with linked list remove

### LINKED ISSUE #7059: [BUG] :filter filter run does not always remove non-matching entries (linked-list.js issue)
### Describe the bug

I expect the output of the filter expression `A =A B =A :filter[match[B]]` to be `B`, but the output is `B A`

Looks to me like an issue with `linked-list-.js` not removing items correctly. If I add this line to `editions/test/tiddlers/tests/test-linked-list.js`:

```js
compare(remove(newPair(["A", "A", "B", "A"]), ["A", "A", "A"])); // B
```

Then I get this test failure when running the test suite:

```
Failures:
1) LinkedList class tests can remove all instances of a multi-instance value
  Message:
    Expected $.length = 2 to equal 1.
    Unexpected $[1] = 'A' in array.
  Stack:
    Error: Expected $.length = 2 to equal 1.
    Unexpected $[1] = 'A' in array.
        at <Jasmine>
        at compare (test-linked-list.js:64:35)
        at UserContext.<anonymous> (test-linked-list.js:111:7)
        at <Jasmine>
  Message:
    Expected 2 to be 1.
  Stack:
    Error: Expected 2 to be 1.
        at <Jasmine>
        at compare (test-linked-list.js:65:32)
        at UserContext.<anonymous> (test-linked-list.js:111:7)
        at <Jasmine>
```

### Expected behavior

The `:filter` filter run should remove all non-matching entries from the input

### To Reproduce

1. Paste `A =A B =A :filter[match[B]]` into https://tiddlywiki.com/#%24%3A%2FAdvancedSearch filter tab
2. See the unexpected results `B A`

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version v5.2.3

### Additional context

_No response_

btheado:
Prior to this fix, the filter expression `A =A B =A :filter[match[B]]` resulted in `B A` instead of just `B`.

The prev pointer of the underlying double linked-list implementation was getting corrupted for this case. The code for `prev` was "symmetric" with the `next` code, even though for `prev` there can be an extra `undefined`  at the front of the list, but the same extra `undefined` entry is not present at the end of `next`.

Fixes #7059

### DIFF
--- a/core/modules/utils/linked-list.js
+++ b/core/modules/utils/linked-list.js
@@ -136,13 +136,9 @@ function _removeOne(list,value) {
 	// we need to know if the FIRST value is the last in the list, not the last.
 	if(next !== undefined) {
 		if(typeof list.prev[next] === "object") {
-			if(prev === undefined) {
-				// Must have been first, and 'i' would be 0.
-				list.prev[next].shift();
-			} else {
-				var i = list.prev[next].indexOf(value);
-				list.prev[next][i] = prev;
-			}
+			// Nothing special needed for first since list.prev[next][0] will be 'undefined'
+			var i = list.prev[next].indexOf(value);
+			list.prev[next][i] = prev;
 		} else {
 			list.prev[next] = prev;
 		}
--- a/editions/test/tiddlers/tests/test-linked-list.js
+++ b/editions/test/tiddlers/tests/test-linked-list.js
@@ -137,6 +137,21 @@ describe("LinkedList class tests", function() {
 		compare(pushTop(newPair(["C", "X", "A", "A"]), "X")); // CAAX
 	});
 
+	it("can remove all instances of a multi-instance value #7059", function() {
+		// Remove duplicate items when one or more items between the duplicates
+		// are not removed and the first of those duplicates is not the first item.
+		// These tests used to fail prior to the fix to #7059
+		compare(remove(newPair(["A", "A", "C", "B", "A"]), ["A", "C", "A", "A"])); // B
+		compare(remove(newPair(["A", "A", "C", "B", "A"]), ["C", "A", "A", "A"])); // B
+		compare(remove(newPair(["A", "A", "C", "B", "A"]), ["A", "A", "A"])); // CB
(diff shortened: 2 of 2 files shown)


Jermolene:
Thanks @btheado. @flibbles are you available to review this?

flibbles:
I tested it. It looks good. Sorry you guys had to dig through the linked-list yourselves. I know it's super logic heavy. This fix does seem to resolve the issue and remove the special exception which apparently is not needed (not sure how that happened.)

> Ouch thank you @btheado. The linked list code is quite challenging to work with, might you be able to investigate a PR?

@Jermolene, I have a refactor of the link-list class (https://github.com/Jermolene/TiddlyWiki5/pull/6056) which I still promote. It's a lot easier to understand, and it's a bit faster if I remember, and it doesn't suffer from the same bug (even if it had the same strange exception...)

Once this is merged, I'd be happy to fix any conflicts and re-prep my refactor.

btheado:
Thanks, @flibbles.

As mentioned in #7059, my initial fix was to just comment out the original line 141. It didn't seem right to me, but I was having trouble figuring out the code to know for sure.  Then I added some code to linked-list which would start at the end of the list and accumulate the items by following the prev links and then reverse the list once finished. I added that to the `compare` function in the tests. That way the arrays generated for comparison were fully using both the next and prev links.

Sure enough with this extra testing of the prev links and with my original fix some of the tests failed. However, with my final fix all these tests passed. This gives me more confidence in the fix.

I didn't think it would make sense to include this extra linked-list code I wrote, so I didn't. However, it occurs to me it should still be possible to write SOME tests using the existing linked-list methods which make more use of the prev links would have exposed that issue. I'll think about that some more.

flibbles:
I would think if a reverse iteration was revealing some corruption of the list, then it's worth having, even if it's not worth having in the link-list module itself. The test-linked-list file could inject it into LinkList.prototype and test there. That's just if we really don't want to add it, but I can think if instances where reverse iteration may be desired in the rest of core, so maybe add it?

Jermolene:
Thanks @btheado @flibbles

PR #7069, opened 2022-11-28
Make Chinese simplified translation more localized

oflg:


oflg:
Thanks @BramChen, I've been using your translations. 
Here I some translations changed to a more simplified Chinese usage.

linonetwo:
LGTM, but this means later @BramChen  can't help us update the simplified Chinese translation like before?

oflg:
> LGTM, but this means later @BramChen  can't help us update the simplified Chinese translation like before?

I'm just doing some side work, and BramChen can certainly do the same as before.

BramChen:
Many thanks @linonetwo, @oflg 

>
> I'm just doing some side work, and BramChen can certainly do the same as before

ACK @oflg

Jermolene:
Thanks @BramChen @linonetwo @oflg

Issue #7067, opened 2022-11-27
[BUG] Page Layout switch menu is not convenient


linonetwo:
### Describe the bug

1. After using keyboard shortcut `ctrl+shift+L`, we can't use arror key to select a layout and use enter to confirm now. This asks the user to use both the keyboard and mouse.
2. When menu is opened, can't close it (and confirm selected layout) by click outside
3. When the menu is opened, we can still open more menus, then the user must click OK multiple times to close it.
4. not working in FireFox

https://user-images.githubusercontent.com/3746270/204124863-72547f57-83fc-409c-8578-a712d1a10e87.mov

### Expected behavior

1. Allow open that menu using mouse, from a page control button. Or allow using arror key and enter key in that menu
2. Disable shortcut when menu is opened
3. Allow click on mask to close the menu
4. should allow switch in Firefox, not only Chrome.

### To Reproduce

1. Go to https://tiddlywiki.com/ or https://tiddly-gittly.github.io/tiddlywiki-whiteboard/
2. using keyboard shortcut `ctrl+shift+L`  x5
3. Try arrow key and click on dark mask
5. See error

### Screenshots

See my screen recording above.

### TiddlyWiki Configuration

Official website, 5.2.3

### Additional context

I'm creating a whiteboard layout.

Jermolene:
Thanks @linonetwo those would all be useful enhancements for the layout switcher.

pmario:
I think the keyboard handling would needed to be implemented in wikitext in the switcher template. 

Closing the modal by clicking on the mask is a new function, which imo will need some changes to the modal.js code.

linonetwo:
I don't know how to do keyboard handling using widget or macro, there seem to be a kbi widget, but is only for text input, not for list.

pmario:
>  4. not working in FireFox

I'm using FF all the time. The function worked there all the time. So I don't understand this issue

linonetwo:
oh, sorry it does worked...In [https://tiddly-gittly.github.io/tiddlywiki-calendar/](https://tiddly-gittly.github.io/tiddlywiki-calendar/) it is working.

But in [https://tiddly-gittly.github.io/tiddlywiki-whiteboard/](https://tiddly-gittly.github.io/tiddlywiki-whiteboard/) I can't switch in firefox, maybe it is a bug of my plugin.

linonetwo:
> Closing the modal by clicking on the mask

This has been implemented.

[...]

PR #7072, opened 2022-11-30
Allow click outside to close (layout) switchers' modal

linonetwo:
Fix problem 2 in #7067

Also, I'm using the modal widget in my new plugin, so I hope there is an option to make its mask closable.

Try ctrl+shift+L in the preview site, and click outside.

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -26,7 +26,7 @@ exports.startup = function() {
 		$tw.modal.display(event.param,{variables: event.paramObject, event: event});
 	});
 	$tw.rootWidget.addEventListener("tm-show-switcher",function(event) {
-		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event, maskClosable: true});
+		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event});
 	});
 	// Install the notification  mechanism
 	$tw.notifier = new $tw.utils.Notifier($tw.wiki);
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -26,7 +26,7 @@ exports.startup = function() {
 		$tw.modal.display(event.param,{variables: event.paramObject, event: event});
 	});
 	$tw.rootWidget.addEventListener("tm-show-switcher",function(event) {
-		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event});
+		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event, maskClosable: true});
 	});
 	// Install the notification  mechanism
 	$tw.notifier = new $tw.utils.Notifier($tw.wiki);
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -26,6 +26,7 @@ Display a modal dialogue
 	options: see below
 Options include:
 	downloadLink: Text of a big download link to include
+	maskClosable: Whether to close the modal dialog when the mask (area outside the modal) is clicked
 */
 Modal.prototype.display = function(title,options) {
 	options = options || {};
@@ -209,6 +210,9 @@ Modal.prototype.display = function(title,options) {
 	headerWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	bodyWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	footerWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
+	if(options.maskClosable) {
+		modalBackdrop.addEventListener('click', closeHandler, false);
+	}
 	// Set the initial styles for the message
 	$tw.utils.setStyle(modalBackdrop,[
 		{opacity: "0"}
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -116,7 +116,7 @@ Modal.prototype.display = function(title,options) {
 			attributes: {
 				text: {
 					type: "string",
-					value: title
+					value: tiddler.fields && tiddler.fields.caption || title
 		}}}],
 		parentWidget: navigatorWidgetNode,
 		document: this.srcDocument,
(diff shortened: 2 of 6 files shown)


linonetwo:
Also I hope we can provide modal title via "caption" field, but I don't know how to handle the case when "caption" field has `{{ }}` 

```
caption: {{$:/language/ControlPanel/LayoutSwitcher/Caption}}
```

I know this can be done in wikitext using `<$transclude field="caption"/>`, but don't know how in JS

core/modules/utils/dom/modal.js
```diff
@@ -115,7 +116,7 @@ Modal.prototype.display = function(title,options) {
 			attributes: {
 				text: {
 					type: "string",
-					value: title
+					value: tiddler.fields && tiddler.fields.caption || title
```
pmario, line 119: Does this change belong to this PR? What is it good for?

core/modules/utils/dom/modal.js
```diff
@@ -209,6 +210,9 @@ Modal.prototype.display = function(title,options) {
 	headerWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	bodyWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	footerWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
+	if(options.maskClosable) {
```
pmario, line 213: This only checks if the variable exists, but IMO it should also check for the value `=== true` 
I'm not sure. Is the maskClosable accessible from wikitext? .. If yes it should be "yes" instead of true. .. If it's only accessible with js functions true is OK

core/modules/startup/rootwidget.js
```diff
@@ -26,7 +26,7 @@ exports.startup = function() {
 		$tw.modal.display(event.param,{variables: event.paramObject, event: event});
 	});
 	$tw.rootWidget.addEventListener("tm-show-switcher",function(event) {
-		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event});
+		$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event, maskClosable: true});
```
pmario, line 29: I think `maskCloseable` is already part of the `paramObject` @Jermolene .. right?
So it doesn't need to be added in the root-widget, where it is hardcoded. IMO it needs to be configurable

core/modules/utils/dom/modal.js
linonetwo: Now I know it can...

core/modules/utils/dom/modal.js
linonetwo: Oh, I can use `subtitle` field to do so.

linonetwo:
Done, now the change is pretty small.

linonetwo:
Try this in the preview site

```
Your name: <$edit-text tiddler="$:/temp/yourName" tag="input" default="Your name"/>

Your message:
<$edit-text tiddler="$:/temp/yourMessage" default="Your message"/>

<$button>
<$action-sendmessage $message="tm-modal" $param="SampleModal" yourName={{$:/temp/yourName}} yourMessage={{$:/temp/yourMessage}} maskClosable=yes/>
Click me!
</$button>
```

core/modules/utils/dom/modal.js
linonetwo: Deleted

linonetwo:
Added a doc.

Feel free to point out anything I need to change.

core/ui/KeyboardShortcuts/switcher.tid
```diff
@@ -2,4 +2,4 @@ title: $:/core/ui/KeyboardShortcuts/switcher
 tags: $:/tags/KeyboardShortcut
 key: ((layout-switcher))
 
-<$action-sendmessage $message="tm-show-switcher" switch="layout"/>
\ No newline at end of file
+<$action-sendmessage $message="tm-show-switcher" switch="layout" maskClosable=yes/>
```
pmario, line 5: The convention is to have double quotes for strings in the TW core, to be consistent `maskClosable="yes"`. If we use it consistently, there is a higher chance that users also use them. It causes less issues.

pmario (commented):
Looks much simpler now.

Jermolene:
Thanks @linonetwo.

The core does not currently define any special members of the paramObject for the tm-modal message. Instead, we currently use fields of the modal tiddler to define the modal. Could we continue to use the same approach, using a "mask-closable" field on the modal tiddlers instead of as a message parameter?

core/modules/utils/dom/modal.js
```diff
@@ -212,7 +212,7 @@ Modal.prototype.display = function(title,options) {
 	bodyWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	footerWidgetNode.addEventListener("tm-close-tiddler",closeHandler,false);
 	// Whether to close the modal dialog when the mask (area outside the modal) is clicked
-	if(variables.maskClosable === "yes" || variables.maskClosable === "true") {
+	if(tiddler.fields && (tiddler.fields["mask-closable"] === "yes" || tiddler.fields["mask-closable"] === "true")) {
```
pmario, line 215: @Jermolene .. Is it needed to check for `.fields` in the `tiddler` object. I think it's not, since every tiddler has to have a title field. So it should be redundant to check for fields here. ??

linonetwo: You know, people are good at copying code...  
 

So actually I copied it from here:

https://github.com/Jermolene/TiddlyWiki5/blob/6f225cef634f8940410e92d04753ed4821d83a5a/core/modules/utils/dom/modal.js#L69

linonetwo: I don't know why it does this check, so I decided to do the same.

linonetwo: The vercel perview is so convenient.

linonetwo:
The vercel preview is so convenient.

core/modules/utils/dom/modal.js
Jermolene: Thanks @linonetwo I think it is right to be consistent here.

Jermolene (commented):
All looks good thank you @linonetwo

PR #7076, opened 2022-12-02
Add layout switcher page control button

linonetwo:
try address problem 1 in #7067


linonetwo:
ICON:

Layout by artworkbean from \<a href="https://thenounproject.com/browse/icons/term/layout/" target="\_blank" title="Layout Icons">Noun Project\</a>

linonetwo:
Icon is not working...How to create a proper icon...

linonetwo:
Use [https://morosanuae.github.io/tw-icons/](https://morosanuae.github.io/tw-icons/) instead, save lots of time dealing with icon!

linonetwo:
Why icon is not shown here?

linonetwo:
Ready to be merged I think.

pmario:
@saqimtiaz @Jermolene ... For me the layout-switcher was implemented as a "backup" functionality if users "brick" the default layout, so they have a possibility to switch back. 

It's functionality somewhat overlaps with the theme switcher, which imo should be used to deal with those type of changes. 

Adding more and more functions to the layout switcher, that it basically mimics the theme-switcher shouldn't happen. 

The only difference between the layout switcher and the theme-switcher is how they are activated. 

So instead of adding new functions and page-buttons to the layout switcher the functions from theme- and layout-switcher should be consolidated. ... if possible.

linonetwo:
I thought Theme was about beautify the layout using CSS, and Palette was about change variables in the CSS of theme, and Layout can be total change of layout using HTML.

I'm treating layout manager as a way to mimic https://heptabase.com/ and https://affine.pro/ , you can see my recent plugins are all about this.

saqimtiaz:
> For me the layout-switcher was implemented as a "backup" functionality if users "brick" the default layout, so they have a possibility to switch back.

The layout switcher was implemented as a modal to help users if they brick the layout and need a way to recover. However, the layout mechanism itself was meant to be at the same level as themes and palettes as another opportunity to customize the wiki. See https://tiddlywiki.com/#Alternative%20page%20layouts

 I do agree that the differences between a theme and a layout can get blurry. Personally I have a switcher that switches all three at the same time usually: layout, theme and palette. 



> Anyway, if Layout wasn't intended for doing so, I will have to find another way to add whiteboard and calendar "View" I think...

I think using layouts for this is entirely appropriate and see no problem with it.

linonetwo:
I think I will also want to keep using one theme (or actually I mean palette?), and switch between layouts very frequently, write something, put them on the whiteboard, and record some todo on the calendar.

BTW, what do you mean by "brick the layout"?

linonetwo:
I think we can have recommendation for developers, like: only change color in palette plugin, only add CSS in theme plugin, and mess up with anything in layout plugin...

saqimtiaz:
> BTW, what do you mean by "brick the layout"?

Modifying the layout in such a way that the wiki becomes unusable. In such situations, the layout switcher modal mechanism offers the possibility to switch to the default layout and fix the wiki.

linonetwo:
After this PR, I will add this button to my layouts.

And currently refresh the page will bring you back to default layout, so it is safe.

pmario:
>I do agree that the differences between a theme and a layout can get blurry. Personally I have a switcher that switches all three at the same time usually: layout, theme and palette.

So they should complement each others functionality. I can live with that.

pmario:
May be we could move layout just below the theme-switcher button

image

linonetwo:
Done, using list-after
截屏2022-12-02 19 38 04

core/images/layout-button.tid
```diff
@@ -0,0 +1,4 @@
+title: $:/core/images/layout-button
+tags: $:/tags/Image
+
+<svg width="22pt" height="22pt" class="tc-image-tbi-layout tc-image-button" viewBox="0 0 24 24" stroke-width="1" stroke="none"><path d="M0 0h24v24H0z" fill="none"/><rect x="2" y="2" width="7" height="7" rx="2"/><rect x="2" y="13" width="7" height="9" rx="2"/><rect x="12" y="2" width="10" height="20" rx="2"/></svg>
```
Jermolene, line 4: I'd expect the classes to be `tc-image-layout tc-image-button`

core/ui/PageControls/layout.tid
```diff
@@ -0,0 +1,23 @@
+title: $:/core/ui/Buttons/layout
+tags: $:/tags/PageControls
+caption: {{$:/core/images/layout-button}} {{$:/language/Buttons/LayoutSwitcher/Caption}}
+description: {{$:/language/LayoutSwitcher/Description}}
+list-after: $:/core/ui/Buttons/theme
```
Jermolene, line 5: We shouldn't have a list-after field here. The correct way to set the default ordering is via the tiddler https://github.com/Jermolene/TiddlyWiki5/blob/master/core/wiki/tags/PageControls.tid

core/ui/PageControls/layout.tid
```diff
@@ -0,0 +1,23 @@
...
+<$list filter="[<tv-config-toolbar-icons>match[yes]]">
+{{$:/core/images/layout-button}}
+</$list>
+<$list filter="[<tv-config-toolbar-text>match[yes]]">
+<span class="tc-btn-text"><$text text={{$:/language/Buttons/LayoutSwitcher/Caption}}/></span>
+</$list>
+</$button>
+\end
+
+<$list filter="[list<tv-story-list>] +[field:title[$:/LayoutSwitcher]]" emptyMessage=<<control-panel-button>>>
```
Jermolene, line 21: This code looks like it's copied from the page control button for control panel. The intention is to highlight the button if the control panel tiddler is opened in the story river. That doesn't work here because the button action is to trigger the layout switcher. So I think this button can actually look at lot more like a simple one such as $:/core/ui/Buttons/save-wiki

core/ui/PageControls/layout.tid
```diff
@@ -0,0 +1,23 @@
...
+tags: $:/tags/PageControls
+caption: {{$:/core/images/layout-button}} {{$:/language/Buttons/LayoutSwitcher/Caption}}
+description: {{$:/language/LayoutSwitcher/Description}}
+list-after: $:/core/ui/Buttons/theme
+
+\whitespace trim
+\define control-panel-button(class)
+\whitespace trim
+<$button tooltip={{$:/language/Buttons/LayoutSwitcher/Hint}} aria-label={{$:/language/Buttons/LayoutSwitcher/Caption}} class="""$(tv-config-toolbar-class)$ $class$""">
+{{$:/core/ui/KeyboardShortcuts/switcher}}
```
Jermolene, line 11: I think it would be clearer not to use a transclusion here, but instead include the action-sendmessage widget directly.

Jermolene (commented):
Thank you for your patience @linonetwo just a couple of points of feedback.

core/images/layout-button.tid
linonetwo: My bad, I copy paste this and forget to clean it up

core/ui/PageControls/layout.tid
linonetwo: Ok, I should check if its tag has a list field in the future.

core/ui/PageControls/layout.tid
linonetwo: Ok, I thought we should do abstractions like this. Maybe literal programming is different from traditional programming here.

core/ui/PageControls/layout.tid
linonetwo: Got it, interesting usage, I didn't understand it at first glance, I have to admit...

linonetwo (commented):
Updated

linonetwo:
Thank you, Mario, Working properly this time.



截屏2022-12-04 23 20 50
截屏2022-12-04 23 20 32
截屏2022-12-04 23 20 25

Jermolene:
Thanks @linonetwo 

> Ok, I thought we should do abstractions like this. Maybe literal programming is different from traditional programming here.

Absolutely, it's a reasonable thing to do, but I think it would need to be done consistently across all the keyboard shortcuts.

PR #7080, opened 2022-12-03
Added using the TextWidget to suppress soft and hard links

AnthonyMuscio:
Supporting a recent change to `[[Hard and Soft Links]]` tiddler.

The initial issue was discussed here https://talk.tiddlywiki.org/t/lost-in-brackets/5354?u=tw_tones

The TextWidget is often used to suppress hard and soft links so a reference to `[[Hard and Soft Links]]` in the documentation given.

A common use of the text widget is to suppress soft links as the result of "Filtered Transclusions" so a reference to `[[Filtered Transclusions|Transclusion in WikiText]]`

The above text also results in additional search results for "soft links" and "filtered transclusions" which show how the TextWidget is commonly used for this purpose.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/TextWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/TextWidget.tid
@@ -9,6 +9,9 @@ text-test: e=mc^^2^^
 
 The text widget displays plain text without parsing it as [[WikiText]], opposite of [[WikifyWidget]].
 
+* The ~TextWidget can be used to suppress a soft link from being recognised see [[Hard and Soft Links]]
+* The ~TextWidget is commonly used to suppress a soft link when using [[Filtered Transclusions|Transclusion in WikiText]]
+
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/widgets/TextWidget.tid
```diff
@@ -9,6 +9,9 @@ text-test: e=mc^^2^^
 
 The text widget displays plain text without parsing it as [[WikiText]], opposite of [[WikifyWidget]].
 
+* The ~TextWidget can be used to suppress a soft link from being recognised see [[Hard and Soft Links]]
```
saqimtiaz, line 12: Recommend that the words "soft link" be the link to the tiddler "Hard and Soft Links", dropping the text "See [[Hard and Soft Links]]".

pmario:
@AnthonyMuscio -- See remarks

Issue #7070, opened 2022-11-28
[IDEA] add focus="yes" parameter to $select widget


ericshulman:
In several places, I have a `$button` widget that triggers a popup via `<$reveal type="popup"... >` .  The popup contains a `$select` widget.  It would be a significant improvement for keyboard interaction if that `$select` widget could automatically get the focus as soon as it is rendered within the `$reveal` widget.  This same type of interaction is currently available for the $edit-text widget by using a parameter `focus="yes"`.  By adding handling for `focus="yes"` to the `$select` widget it would be possible to click (or press the `space` key) to trigger a `$button` that displays a listbox, and then immediately use the down-arrow key to scroll through the available listbox options to make a selection.

To add this optional behavior to the `$select` widget, just two small changes are needed in `$:/core/modules/widgets/select.js`:

In `SelectWidget.prototype.execute` add 1 line:
   `this.selectFocus = this.getAttribute("focus");`

In `SelectWidget.prototype.render` add 1 line:
   `if(this.selectFocus=="yes") { this.getSelectDomNode().focus(); }`

Jermolene:
Thanks @ericshulman that sounds like a good idea.

Would another contributor like to volunteer to take the baton from Eric and make a PR with documentation updates?

kookma:
@ericshulman 
The above proposed solution only works with focus="yes", what about focus="true"? I am not sure, but I see the `$edit-text` accepts both!

Jermolene:
Hi @kookma the convention in TW5 is to use "yes"/"no" rather than "true"/"false". Inconsistencies have crept in a few places like the `<$edit-text>` widget. Those are the result of well-meaning PRs that I didn't catch. It seems harmless and perhaps helpful to support both conventions, but in practice that kind of inconsistency breeds complexity at an alarming rate. For example, imagine in the future that we have the ability to reach into a widget and access its attributes. Any code that needs to check the status of the focus attribute will have to test it against twice as many strings.

So, I'd prefer that we identify all the situations where "true"/"false" are accepted and (a) where possible make sure that "yes"/"no" were accepted too and (b) mark the usages of "true"/"false" as deprecated in the docs.

kookma:
> So, I'd prefer that we identify all the situations where "true"/"false" are accepted and (a) where possible make sure that "yes"/"no" were accepted too and (b) mark the usages of "true"/"false" as deprecated in the docs.

[...]

PR #7081, opened 2022-12-03
Add new focus attribute to $select widget

kookma:
This PR addresses the request by @ericshulman. See https://github.com/Jermolene/TiddlyWiki5/issues/7070.

The docs also updated.

### DIFF
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -42,6 +42,9 @@ SelectWidget.prototype.render = function(parent,nextSibling) {
 	this.execute();
 	this.renderChildren(parent,nextSibling);
 	this.setSelectValue();
+	if(this.selectFocus == "yes") {
+		this.getSelectDomNode().focus();
+	}
 	$tw.utils.addEventListeners(this.getSelectDomNode(),[
 		{name: "change", handlerObject: this, handlerMethod: "handleChangeEvent"}
 	]);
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @kookma

pmario:
IMO there should also be some CSS, that makes the focus visible. eg: 

The following CSS should be near the `select` css definition at line 3101 in vanilla/base.tid

```
select:focus {
	outline: 2px solid <<colour primary>>;
	outline-offset: -2px;
}
```

Jermolene:
Hi @pmario is that a problem in a particular browser? The focus is already visible for the select widget in the browsers I tried:

image

pmario:
Yes. It's a problem in FF. I'm testing with FF on ubuntu atm. Will need a bit more investigation. 

We actually could add it to the existing definition at line: 235 ... I will have a closer look

Jermolene:
Hi @pmario I'm not seeing the problem in Firefox on macOS:

image

pmario:
The problem imo comes from the reset. I'll have a closer look, at the reset itself we are using. May be there is a new version. ... Or we have to contribute the reset upstream. 

I think the decisions made in upstream are wrong or at least should be discussed

pmario:
I think the difference between your 2 screenshots is significant. .. In FF the outline is barely visible. With my laptop I don't see it at all.

Jermolene:
> With my laptop I don't see it at all.

That would seem to be a general problem with Firefox, and so I'm not sure that the core should be specially accommodating it.

pmario:
As I wrote I'll have a discussion with the reset.css folks .. upstream. I did create a new issue. 

>That would seem to be a general problem with Firefox, and so I'm not sure that the core should be specially accommodating it.

As I wrote it's _not_ a firefox problem. It's a problem caused by the reset.css which the core uses. So IMO it _is_ a core problem and should be addressed there if upstream doesn't change it.

Jermolene:
Thanks @pmario it does seem to be a reset problem.

PR #7088, opened 2022-12-04
Correction of zindex for tc-card-ribbon-wrapper (new ribbon)

Telumire:
This fix avoid the new banner to go above the topnav (z-index: 850)

before fix:

image

after fix:

image


Jermolene:
Thank you @Telumire

PR #7089, opened 2022-12-05
Update chinese language files

BramChen:
* Add chinese translations for the new layout switcher page control button


Jermolene:
Thanks @BramChen

Issue #7054, opened 2022-11-22
[BUG] Reuse of $:/core/ui/EditTemplate/fields broken in v5.2.3


Jermolene:
### Describe the bug

The changes in #6511 have resulted in significant data loss for at least one user:

https://talk.tiddlywiki.org/t/is-that-a-bug-undefined-variable-on-the-core-ui-edittemplate-fields-cause-all-the-tiddlers-to-be-deleted-when-creating-a-new-field/5247

The problem is that the changes in #6628 are not backwards compatible. They require that additional variables are set before transcluding `$:/core/ui/EditTemplate/fields`

### TiddlyWiki Configuration

v5.2.3

Jermolene:
Do you have any thoughts about this @FlashSystems?

FlashSystems:
I've had a look at the code and found the problem: The clean-up-routine that deletes all the temporary tiddlers contains a prefix-filter to clean up all temporary tiddlers that were created for different types of input fields. These share a common prefix to make them easy to find. This prefix-filter uses the `newFieldValueTiddlerPrefix` variable. If this variable is empty, the prefix-filter matches all tiddlers :fearful:.
I did not expect anyone reaching directly into the core and calling this out of context. This was definitely not good defensive programming practice :disappointed:
I'll create a PR that creates a `safeNewFieldValueTiddlerPrefix` that makes sure that, if there is no value set, a dummy value will be created that matches nothing. I'll also limit the filter to tiddlers starting with `$:/temp/NewFieldValue` to make sure that a non-empty, but still bogus value for `newFieldValueTiddlerPrefix` can only cause very little damage.

Jermolene:
> I've had a look at the code and found the problem: The clean-up-routine that deletes all the temporary tiddlers contains a prefix-filter to clean up all temporary tiddlers that were created for different types of input fields. These share a common prefix to make them easy to find. This prefix-filter uses the `newFieldValueTiddlerPrefix` variable. If this variable is empty, the prefix-filter matches all tiddlers 😨.
> I did not expect anyone reaching directly into the core and calling this out of context. This was definitely not good defensive programming practice 😞
[...]

PR #7092, opened 2022-12-05
Fix for bug #7054

FlashSystems:
This PR fixes the bug #7054 and another problem that was discovered on the way.

The PR adds a default for `newFieldValueTiddlerPrefix` and a safe default in case the clean-up functions are called with an empty value for `newFieldValueTiddlerPrefix`. I think the risk that new regressions are caused by this changes is rather low. I tried to keep the changes to a minimum.

### DIFF
--- a/core/ui/EditTemplate/fields.tid
+++ b/core/ui/EditTemplate/fields.tid
@@ -17,7 +17,9 @@ $:/config/EditTemplateFields/Visibility/$(currentField)$
 \define new-field-actions()
 \whitespace trim
 <$action-sendmessage $message="tm-add-field" $name={{{ [<newFieldNameTiddler>get[text]] }}} $value={{{ [<newFieldNameTiddler>get[text]] :map[subfilter<get-field-value-tiddler-filter>get[text]] }}}/>
-<$action-deletetiddler $filter="[<newFieldNameTiddler>] [prefix<newFieldValueTiddlerPrefix>] [<storeTitle>] [<searchListState>]"/>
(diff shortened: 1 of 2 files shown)


core/ui/EditTemplate/fields.tid
```diff
@@ -10,14 +10,20 @@ $:/config/EditTemplateFields/Visibility/$(currentField)$
 [[hide]] -[title{$(config-title)$}]
 \end
 
+<!-- Beware this is duplicated from EditTemplate.tid. For details see bug #7054 -->
```
pmario, line 13: TW allows you to use the `\import` pragma. Would this be an option to avoid code duplication? https://tiddlywiki.com/prerelease/#Pragma

FlashSystems: Hello @pmario,

the problem would be that if a plugin or user has overwritten the source tiddler (in this case `FieldEditor.tid`) the `\import` paragma would not find the imported macro, and we would be back to square one.

This would only work if I create a completely new tiddler (that is extremely unlikely to be already overwritten) and import it into both of the destinations. But I think this is too much of a change for a bugfix.

Maybe it would be a good idea to define a path in TiddlyWiki that must not be overwritten by the user or plugins, where we can put utility macros to import without the risk that they vanish.

pmario: >the problem would be that if a plugin or user has overwritten the source tiddler (in this case FieldEditor.tid) the \import paragma would not find the imported macro, and we would be back to square one.

Ahhh ... OK

PR #7094, opened 2022-12-07
Updated and improved tag picker macro

Telumire:
**Code update**

- All the set and vars widgets are replaced with the let widget
- tag-picker-inner is rewritten to be more readable, parts of the macro that was repeated was separated into two smaller macro, tag-button and tagsAutoComplete

**Improvements**

- When a tag is added to a tiddler, it is not longer suggested in the drop down list, this help reduce clutter.
- The popup no longer appears when there is no results in the suggestion dropdown.
- Variables added into the `nonSystemTagsFilter` and `SystemTagsFilter` : `first-search-filter` and `second-search-filter`. These variables allow to edit the tag list suggestions with ease.

### DIFF
--- a/core/wiki/macros/tag-picker.tid
+++ b/core/wiki/macros/tag-picker.tid
@@ -9,12 +9,12 @@ second-search-filter: [tags[]is[system]search:title<userInput>sort[]]
 
 \define add-tag-actions(actions,tagField:"tags")
 \whitespace trim
-<$set name="tag" value={{{ [<__tiddler__>get[text]] }}}>
+<$let tag={{{ [<__tiddler__>get[text]] }}}>
 <$list filter="[<saveTiddler>!contains:$tagField$<tag>!match[]]" variable="ignore" emptyMessage="<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter='-[<tag>]'/>">
 <$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter="[<tag>trim[]]"/>
 $actions$
 </$list>
-</$set>
+</$let>
 <<delete-tag-state-tiddlers>>
 <$action-setfield $tiddler=<<refreshTitle>> text="yes"/>
 \end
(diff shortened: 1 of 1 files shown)


kookma:
> * The popup no longer appears when there is no results in the suggestion dropdown.

I cannot reproduce it, instead I see the old behavior see:

image

As you wrote above, I expect no popup for empty or no result

saqimtiaz:
@Telumire Have you considered using the new nested macros feature instead of introducing new global macros?

Telumire:
@saqimtiaz So I tried to nest the macros but it didn't work, here's what I did : 

```
\define tag-picker-inner(actions,tagField:"tags")
\whitespace trim
\define tag-button(classes)
<$let button-classes='tc-btn-invisible $classes$' currentTiddler=<<tag>>>{{||$:/core/ui/TagPickerTagTemplate}}</$let>
\end
\define tagsAutoComplete()
<$list filter=<<tagsAutoCompleteFilter>> emptyMessage=<<tagsAutoCompleteEmptyMessage>> variable="listItem">
<$list filter=<<tagsFilter>> variable="tag">
<$list filter="[<tag>addsuffix<suffix>] -[<tagSelectionState>get[text]]" emptyMessage=<<tag-button 'tc-tag-button-selected'>> variable="ignore">
<<tag-button>>
</$list>
</$list>
</$list>
\end
<$let 
newTagNameInputTiddlerQualified=<<qualify "$:/temp/NewTagName/input">> 
..
```

Did I miss something ?

Telumire:
@saqimtiaz I see, I should have read the doc more carefully ! Also @kookma the reason why the popup doesnt work is because I use the variable userInput before it's defined, I will correct that

pmario:
There is no code indentation. So if the code is changed completely, it should be human readable. 
So `\whitespace trim` will be needed a lot.

core/wiki/macros/tag-picker.tid
```diff
@@ -28,73 +28,125 @@ $actions$
...
-		focusPopup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-edit-texteditor tc-popup-handle" tabindex=<<tabIndex>> 
-		focus={{{ [{$:/config/AutoFocus}match[tags]then[true]] ~[[false]] }}} filterMinLength={{$:/config/Tags/MinLength}} 
-		cancelPopups=<<cancelPopups>> configTiddlerFilter="[[$:/core/macros/tag-picker]]"/>
-</span><$button popup=<<qualify "$:/state/popup/tags-auto-complete">> class="tc-btn-invisible tc-btn-dropdown" tooltip={{$:/language/EditTemplate/Tags/Dropdown/Hint}} aria-label={{$:/language/EditTemplate/Tags/Dropdown/Caption}}>{{$:/core/images/down-arrow}}</$button><$reveal state=<<storeTitle>> type="nomatch" text=""><$button class="tc-btn-invisible tc-small-gap tc-btn-dropdown" tooltip={{$:/language/EditTemplate/Tags/ClearInput/Hint}} aria-label={{$:/language/EditTemplate/Tags/ClearInput/Caption}}>{{$:/core/images/close-button}}<<delete-tag-state-tiddlers>></$button></$reveal><span class="tc-add-tag-button tc-small-gap-left">
-<$set name="tag" value={{{ [<newTagNameTiddler>get[text]] }}}>
-<$button set=<<newTagNameTiddler>> setTo="" class="">
+<$macrocall $name="keyboard-driven-input"
```
pmario, line 68: I personally don't like macrocalls, that create 1 line for every parameter. I think they don't improve readability at all. ... But it's just me. 

May be indentation will help

Telumire: I originally indented my code but I noticed that there was none on the tiddler so I removed the indents, I will add them then 👌

Telumire: > As you wrote above, I expect no popup for empty or no result
> 
> I assume the mechanism for displaying dropdown is to see if the input is focused or non-empty, while it can be changed to: show dropdown if there is any result. I think this is more semantic, but one may believe to show a message instead of above empty dropdown with horizontal line!

@kookma the popup should be fixed now. For the focus indication there is the blinking cursor, and the search doesnt show a popup when nothing is written so in my opinion it's a bit more consistent to have the same behavior in the tag picker, but I can revert this change if needed

Telumire: Hum seems like I made a mistake, I'm working on it

Telumire:
@pmario Apologies, I use VSCode and I didn't notice that it was converting my tabs to spaces. It's fixed now.

kookma:
@Telumire 
Thank you for all your efforts. The preview I see is 5.2.3-pre. I suspect the fork you have is old.

kookma:
@Telumire 
I have downloaded your fork at https://github.com/Telumire/TiddlyWiki5/tree/tag-picker-update and tested on my local machine!
It is an old fork. You need yo use the latest version of TW.

pmario:
The tag-picker macro has been completely refactored since this PR was created. So close

PR #7097, opened 2022-12-07
Improve readability for: Using the external JavaScript template

pmario:
This PR is DOCS only

- improves readability by removing most unnecessary WikiLinks to elements _not_ important for this doc info.
- It makes the core tiddler [$:/core/ui/ExportTiddlyWikiCore](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FExportTiddlyWikiCore) visible again. 

There is a new "Show the Download Info" button now. It's **intentional** that the workflow is a **2-click action**. Users should read the text in front of the button to know what's going on. If there is a "big green downlaod button" they usually click it, whithout reading anything. Even if the button is in the middle of the tiddler text

With the latest changes to the ViewTemplate the core/ui tiddlers are shown as text/plain. .. So the Download button doesn't work at the moment. 

[Link to DemoWiki -- Using the external Download template](https://tiddlywiki5-git-fork-pmario-improve-readabilit-c689a0-jermolene.vercel.app/#Using%20the%20external%20JavaScript%20template)

## Info Shown

image

### DIFF
--- a/editions/tw5.com/tiddlers/webserver/Using the external JavaScript template.tid
+++ b/editions/tw5.com/tiddlers/webserver/Using the external JavaScript template.tid
@@ -1,22 +1,35 @@
 created: 20180905075846391
-modified: 20210611055708739
+modified: 20221207112242775
 tags: [[WebServer Guides]]
 title: Using the external JavaScript template
 type: text/vnd.tiddlywiki
 
-You can use a special template to externalise TiddlyWiki's core code into a separate file. This configuration allows the browser to cache the core for improved efficiency. 
+\define showDownloadExternalButton()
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario hiding the noisy links seems like a good idea, but I'm not sure what we gain by putting the download button behind another button. 

The other point is that the intention here is not (just) to provide a way for the user to download the core JS file there and then, but also to provide instructions so that the user is able to do it at any arbitrary version of TW that they need.

The wording is clumsy with the button because it occurs in the middle of a sentence. I think it would probably make more sense to transclude the button in the usual way.

pmario:
I did remove the extra button and only transclude the green-button.

Jermolene:
Great thanks @pmario

PR #7098, opened 2022-12-07
fix long words in CODE element makes tables overflow [base css]

pmario:
Hi Jeremy, I'm sorry, but there is still one element that can make tables overflow if it contains long words. 

With the last PR I did fix PRE code-blocks, but I did miss the CODE element, because I wasn't exactly sure, what caused the problem. Now I know. It's the following 2 things. 

[code-and-pre-causing-table-overflow.zip](https://github.com/Jermolene/TiddlyWiki5/files/10176064/code-and-pre-causing-table-overflow.zip)

The coincidence is, that the URLs reach to the same position, which should trigger the word-break. So I didn't see the CODE element also causes a problem. While testing -- the PRE element was more obvious to cause problems.

image

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -166,6 +166,7 @@ code {
 	padding: 0 3px 2px;
 	border-radius: 3px;
 	font-family: {{$:/themes/tiddlywiki/vanilla/settings/codefontfamily}};
+	word-break: break-word;
 }
 
 blockquote {


pmario:
The problem with tables is, that there only needs to be 1 element somewhere in the column, that doesn't want to "wrap", the whole column is kept at that sice, which made it so hard to actually find the offending construction :/

Jermolene:
Hi @pmario once again I'm concerned that this might have unexpected effects on backwards compatibility with existing usage in the field.

Lets merge it after v5.2.4 is out so that users can try it in the prerelease. I've made a ticket to keep track: #7100

pmario:
bump

Jermolene:
Hi @pmario as I said above, the concern here is that this change is clearly not backwards compatible, and it is not at all clear what impact it will have on existing code.

pmario:
Close the PR ... renamed the PR it to make it easier to find in the future.

Issue #7003, opened 2022-10-20
[BUG] search dropdown is hidden on smaller screens


pmario:
### Describe the bug

see screenshot

The problem is in vanilla/base 

```
.tc-block-dropdown-wrapper {
	position: relative;
}
```

If commented out `position: relative` it seems to work. Since this line is part of the vanilla/base since 2015 and no changes have been made, it's not clear if there are any side effects. 

I didn't find any side effects if it is commented out like this: 

```
.tc-block-dropdown-wrapper {
<!--	position: relative; -->
}
```

### To Reproduce

- Go to tiddlywiki.com
- make the browser window small -> so sidebar is on top
- search for "test"
- see the problem

### Screenshots

![image](https://user-images.githubusercontent.com/374655/197000986-0c79876a-fca5-4293-b30f-b7edd9c17112.png)

### TiddlyWiki Configuration

tiddlywiki.com v5.2.3

pmario:
Did some testing. It does **not** interfere with https://tiddlywiki.com/#Customising%20search%20results

Jermolene:
Hi @pmario it's great that removing the `position: relative` fixes the overlap issue, but I am worried that it cannot be backwards compatible to globally change `tc-block-dropdown-wrapper`. I would feel safer adding a new class to the search results dropdown that overrides the `position` CSS property.

Alzacon:
I think the dropdown is hidden because it is inside the sidebar and this last has `overflow: auto;`
```
<div class="tc-sidebar-scrollable" style="overflow: auto;">
```
I don't find the origin of this inline style. The scrolable widget create other styles and they appears in the CSS class.

Its class `tc-sidebar-scrollable` is used in a media query (+960px) and it also have properties for the overflow, that aren't used in the desktop UI because there is this inline style.

If you disable this inline style in the browser and it also works.

FlashSystems:
Sorry for my late response. @pmario noticed me of this PR to check if it collides with PR #6887. While adding this change to the demo for PR #6887 to check for problems, I noticed that the proposed change is not correct. To comment out the style in `base.tid` a CSS comment should be used. I think the correct change should be:

```css
.tc-block-dropdown-wrapper {
	/* position: relative; */
}
```

I've added the change to the [demo](https://flashsystems.github.io/FieldTemplatesDemo/abspopup.html) for PR #6887 if anyone wants to verify it for themselves.

twMat:
[...]

Issue #4038, opened 2019-07-01
search dropdown is hidden on narrow screens


BurningTreeC:
to reproduce: use a narrow view on tiddlywiki.com, so that the sidebar shows on top. then search for something so that the search dropdown shows up (Hello for example). It is hidden below the story river which is a bit annoying

BurningTreeC:
I would prefer the sidebar not to overflow and to give it a z-index high enough so that its dropdowns show on top of the river

pmario:
https://wikilabs.github.io/editions/advanced-search-plus/ ... gives you a new tab in the sidebar, that doesn't create a popup.

twMat:
>give it a z-index high enough

Do you get this to work in the inspector tool? I don't.

One circumvention is to make sure you're on tab *Recent* which is longer than tab *Open*. Not a real solution of course.

BurningTreeC:
>Do you get this to work in the inspector tool? I don't.

No I'm not getting this to work with simple css solutions neither... :(

Jermolene:
This is indeed annoying. Here's a picture for posterity:

<img width="544" alt="image" src="https://user-images.githubusercontent.com/174761/60525518-66a4d380-9ce6-11e9-9681-b013ac974334.png">

I think the problem is the scrollable widget in `$:/core/ui/PageTemplate/sidebar`. We don't need the scrollable at all in compact mode. I wonder if we might be able to disable it via CSS? Otherwise we'll have to attack the wikitext template...

BurningTreeC:
My "solution" is revealing a `div` instead of the `scrollable` widget on narrow screens using the `dynaview plugin` ... but that's nothing for the core ...

amreus:
If this is related to #6440, can it be closed also?

Jermolene:
Thanks @pmario

PR #7099, opened 2022-12-07
make search dropdown visible on small screens

pmario:
fix for  [BUG] search dropdown is hidden on smaller screens #7003 
fix for  search dropdown is hidden on narrow screens #4038 

This PR adds a media-query, which checks if the search input is on top of the screen and modifies the CSS accordingly. 

Did test it with 

- FF latest on Windows, 
- Edge latest on Widows and 
- Android FF mobile
- Android Chrome on mobile

### To Reproduce

- Go to tiddlywiki.com
- make the browser window small -> so sidebar is on top
- search for "test"
- see the problem

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1999,6 +1999,15 @@ html body.tc-body.tc-single-tiddler-window {
 	margin: auto;
 }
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Bravo! Thank you @pmario that does seem to do the trick.

PR #7102, opened 2022-12-07
Use iframe to embed PDF

m42e:
`<iframe>` is kind of preferred over `<embed>` and widely supported.


Jermolene:
Thank you for your patience, @m42e.

PR #7103, opened 2022-12-08
Change first letter in two plugin-related buttons to be uppercase for consistency

EvidentlyCube:
# What?

Change the case of the first letter in two plugin-related buttons to be uppercase.

# Why?

* To be consistent with the earlier big button "Get more plugins"
* I think buttons starting with uppercase letter look better (100% subjective)

# Counterpoint

* We could consider change "Get more plugins" to "get more plugins" since most of the time it seems the text on buttons in TW is all lowercase.

# Screenshots

image


Jermolene:
Thanks @EvidentlyCube I think that makes sense

PR #7115, opened 2022-12-13
Remove "Saving" tag from "Saving on TiddlySpot"

simonbaird:
The content is still there since there are links to it from various places, but let's not have its card appear in the list of savers shown in the "Saving" tiddler any more.

Removing just the "Saving" tag would have been sufficient, but I think it makes sense to remove the other tags as well.

### DIFF
--- a/editions/tw5.com/tiddlers/saving/Saving on TiddlyHost.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving on TiddlyHost.tid
@@ -3,7 +3,7 @@ color: #29B6F6
 community-author: Simon Baird
 created: 20210422191232572
 delivery: Service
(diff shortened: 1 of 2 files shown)


editions/tw5.com/tiddlers/saving/Saving on TiddlyHost.tid
```diff
@@ -3,7 +3,7 @@ color: #29B6F6
 community-author: Simon Baird
 created: 20210422191232572
 delivery: Service
-description: Online service for creating and hosting TiddlyWikis
+description: Online service for creating and hosting ~TiddlyWikis
```
simonbaird, line 6: It's an unrelated formatting fixup.

Jermolene:
Thanks @simonbaird

Issue #7100, opened 2022-12-07
Tickets for consideration after v5.2.5


Jermolene:
This is a list of tickets that we intend to consider once v5.2.5 is released:

* [x] #7098 
* [x] #7102 
* [x] #7082 
* [x] #7128
* [x] #7121
* [x] #7130 
* [x] #7129 
* [ ] #7133 
* [x] #7124 
* [ ] #7142 
* [x] #7144 
* [x] #7145
* [x] #6528

PR #7121, opened 2022-12-14
correction of the toc and toc-body macro

Telumire:
- Tiddler without toc-link field can now be overriden
- Converted spaces to tabs

Copy paste the following in the [preview tiddlywiki](https://tiddlywiki5-git-fork-telumire-toc-macro-update-jermolene.vercel.app/) generated by vercel to test the changes :

```
<table>
<thead><th>caption = title</th><th>default</th></thead>
<tr>
<td><$let toc-caption="{{!!title}}"><<toc "Filter Operators">></$let></td>
<td><<toc "Filter Operators">></td>
</tr>
</table>
```

<details>
<summary>Old behavior</summary>

image

</details>

<details>
<summary>New behavior</summary>

image

</details>


Alzacon:
It is great idea. Could it be used in the line 148 too?

There are a minor issue in the default of the New behaviour. Its is visual: a extra separation from the list marker.

The `toc-caption` macro takes the margin from the class `.tc-tiny-gap-left` that was introduced in https://github.com/Jermolene/TiddlyWiki5/pull/6825 for the added container. But if the class is removed, I think it could create another visual issue with the buttons of toc macros. Then new selectors would be necessary but without added it in the html. They could be a kind of sibiling or child selector.

Jermolene:
Hi @Telumire thank you.

It's difficult to see what has actually changed here because of changing the indentation. I'd prefer that changes like that be put into a separate PR.

I am not clear if this is a bug fix or a new feature. If it is a new feature then a documentation change would usually also be needed.

Telumire:
> There are a minor issue in the default of the New behaviour. Its is visual: a extra separation from the list marker.

@Alzacon I'm not sure if this is an issue, this gap was there before but previously was only inserted on tiddlers with the field "toc-link" set to no, now it's displayed by default - but it can always be overiden. Unless the caption should be different for tiddlers with a toc-link field set to no ?

>It's difficult to see what has actually changed here because of changing the indentation. I'd prefer that changes like that be put into a separate PR.

@Jermolene ok, I will revert the tabs to spaces. I'm not sure either, toc-caption was already used in the toc macro but not in the empty message, basically I just changed the code from this : 

```
<$list filter="[all[current]toc-link[no]]" emptyMessage="<$link to={{{ [<currentTiddler>get[target]else<currentTiddler>] }}}><$view field='caption'><$view field='title'/></$view></$link>">
              <<toc-caption>>
</$list>
```

to this : 

```
<$list filter="[all[current]toc-link[no]]" emptyMessage="<$link to={{{ [<currentTiddler>get[target]else<currentTiddler>] }}}><<toc-caption>></$link>">
              <<toc-caption>>
</$list>
```

Telumire:
By the way, if you open the commit in the web vscode editor, it's much easier to see the changes made : 

Go to [`262625a` (#7121)](https://github.com/Jermolene/TiddlyWiki5/pull/7121/commits/262625aa62425d7e3b22d4c057e3cc75a170b749)

Then use the "." key to open the vs code editor. VS Code doesnt show differences for spaces/tabs : 

image

Alzacon:
> this gap was there before but previously was only inserted on tiddlers with the field "toc-link" set to no, now it's displayed by default
My wrong. I tried to point it, but it seems that I didn't explain well. Yes, the gap (via macro) was introduced but it was only used to fix hardcoded spaces but it wasn't extended its use to other places, as your PR does an improvement of the hackability.

I talked about find a possible solution, and now I think that could be it. The small gap is needed where there is a button, I think we could use the following selectors with the same propertie and its value that the class `.tc-tiny-gap-left` uses.
```
button + .tc-toc-caption,
button > .tc-toc-caption{
margin-left: .25em;
}
```
And then we will remove, in the `toc-caption` macro, the mentioned class to have
```
<span class="tc-toc-caption">
```

Jermolene:
Hi @Telumire thank you, that looks much better. I'll add this ticket to #7100 for merging after v5.2.5.

I would like to address @Alzacon's point though; this change should ideally not introduce any visible differences.

Telumire:
> I would like to address @Alzacon's point though; this change should ideally not introduce any visible differences.

Ok, so do I make a new PR with @Alzacon's new css code and remove the class .tc-tiny-gap-left from the macro as suggested, or do I do it in this PR ?

Jermolene:
> Ok, so do I make a new PR with @Alzacon's new css code and remove the class .tc-tiny-gap-left from the macro as suggested, or do I do it in this PR ?

I think a separate PR would be best, thank you @Telumire much appreciated.

Jermolene:
Thanks for your patience @Telumire

Telumire:
@Jermolene I forgot to remove the class and push a new PR, sorry ! I'll do it now

PR #7122, opened 2022-12-15
Fix html macro exception missing parameter value

pmario:
This PR fixes the problem where a macrocall with an empty parameter causes the RSOD 
Also see the discussion at [Talk](https://talk.tiddlywiki.org/t/bug-with-selecting-which-field-to-render-macro-results-in/5472)

To preproduce

- open tiddlywiki.com
- create a tiddler with this content: `<<list-links '[tag[HelloThere]]' field:>>`
- Activate the preview 
- The RSOD comes up. 

image

The problem is this line

```
var domNode = this.document.createElementNS(this.namespace,this.tag);
```

[.createElementNS()](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS) allows the second parameter `createElementNS(namespaceURI, qualifiedName)` to be a HTML tag name. 

**In this case `this.tag` -> `qualifiedName` contains the value `field:` ... where the `:` colon throws an exception. ...** 

This PR 

- creates an error message and 
- writes the JS-error string to the console. 

image

### DIFF
--- a/core/modules/widgets/element.js
+++ b/core/modules/widgets/element.js
@@ -71,7 +71,15 @@ ElementWidget.prototype.render = function(parent,nextSibling) {
 	// Make the child widgets
 	this.makeChildWidgets();
 	// Create the DOM node and render children
-	var domNode = this.document.createElementNS(this.namespace,this.tag);
+	try {
+		var domNode = this.document.createElementNS(this.namespace,this.tag);
+	} catch (error) {
+		this.makeChildWidgets([{type: "error", attributes: {
+			"$message": {type: "string", value: 'Parameter value is missing: "' + this.tag + '"' }
+		}}]);
+		this.renderChildren(this.parentDomNode,null);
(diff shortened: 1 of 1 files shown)


pmario:
@Jermolene ... Do _not_ merge yet, since we will probably need a new translatable error text

pmario:
If **add console-error to the Logger class #7125** is merged I'd like to use the `logger.error()` function to create a red error message on the dev console, instead of an almost invisible `log()` message

pmario:
@Jermolene Your latest fix for the genesis widget didn't fix this problem

Jermolene:
> @Jermolene Your latest fix for the genesis widget didn't fix this problem

I had not seen this ticket.

Jermolene:
Thanks @pmario I'm going with a slightly different fix which is to silently correct the tag name; it's consistent with what we already do with unsafe elements.

PR #7125, opened 2022-12-15
add console-error to the Logger class

pmario:
This PR allows the Logger class to use `console.error()` ... since `console.log()` doesn't add any color to messages in the browser dev-console. 

This is needed, if a core function throws an error and there is a `try/catch` function. If possible the `catch()` function should avoid data corruption. 

-----------

There is a `$tw.utils.error()` function which would write a message to the browser console. BUT it also creates the RSOD info. .. So it's not usable. 

`$tw.utils.log()` is not usable for my usecase, since it's not visible enough. 

----------

An alternative would be to create global `$tw.utils.errorToConsole()` utility function. .. BUT I think the implementation in the PR is more versatile. 

-------

If the PR is merged I'll add some docs to the tiddlywiki.com/dev edition. ... Because messing around with the naming inconsistencies did cost me 3 hours. 

So investing 2 more to create some docs aobut what I learned IMO would be OK

### DIFF
--- a/core/modules/utils/logger.js
+++ b/core/modules/utils/logger.js
@@ -52,7 +52,19 @@ Logger.prototype.log = function(/* args */) {
 			logMessage[logMessage.length-1] += $tw.utils.terminalColour();
 			return Function.apply.call(console.log, console, logMessage);
 		}
-	} 
+	}
+};
+
+/*
+Log an error message to console.error so it also gets a red colour in browsers
+*/
+Logger.prototype.error = function(/* args */) {
+	var self = this;
+	if(console !== undefined && console.error !== undefined) {
+		var logMessage = [$tw.utils.terminalColour(this.colour) + this.componentName + ":"].concat(Array.prototype.slice.call(arguments,0));
+		logMessage[logMessage.length-1] += $tw.utils.terminalColour();
+		return Function.apply.call(console.error, console, logMessage);
+	}
 };
 
 /*


pmario:
bump

Jermolene:
Hi @pmario I notice that the new `error()` method does not save output to the buffer as we do in the `log()` method. Is there a reason for this? It seems that with use of a factory function the two methods could share most of their implementation.

But I'm concerned at a deeper level at the proliferation of similar but different debugging methods.

What is the underlying problem you want to solve? If it is to add colour to log messages I'd rather we approached that by trying to make the browser and server more compatible in that regard.

> This is needed, if a core function throws an error and there is a `try/catch` function. If possible the `catch()` function should avoid data corruption.

I don't quite understand. Are you suggesting that there are places in the core that we need to add error trapping and use this proposed new method?

pmario:
>I notice that the new error() method does not save output to the buffer as we do in the log() method. Is there a reason for this

I did search the whole code-base and `logger.getBuffer()` is only used once in the syncer. 

There is no info about the "args" in the  `Logger.prototype.log = function(/* args */) {` .. so I don't really know, what the function exactly should do and which parameters can be assigned in which order ... 

Since the existing logger is unable to create a console.error() it was the only thing the new function should do. So I implemented it that way. 

I could have added a new parameter to the `.log(/*args*/)`, but I don't know how, without causing side effects.

pmario:
>> This is needed, if a core function throws an error and there is a try/catch function. If possible the catch() function should avoid data corruption.

>I don't quite understand. Are you suggesting that there are places in the core that we need to add error trapping and use this proposed new method?

No I wanted to have a log function for plugins to call if an error message should be logged to the console in the same way plugins can use the .log() method.

PR #7129, opened 2022-12-17
Fully committing core to the $genesis widget

flibbles:
I *love* the new genesis widget.

It's great. It provides a far better mechanism for dynamic widgets than the old way, which was to do placeholder widget like this: `\define macro(tag:div) <$tag$>...`

Placeholder widgets were **bad** for plugin developers like me. Take Uglify. It can't parse <$tag$>, so it actually saw it as plaintext containing a placeholder. And all it could do was stay away from it, because touching it was dangerous. As a result, Uglify becomes much less effective around them. But it *can* handle `<$genesis>` just fine. It's safer. It compresses better.

Even Relink prefers $genesis. If it encountered `<$tag$ attribute=...>`, it couldn't relink any of its attributes at all, because it couldn't parse it as a widget. With $genesis, it can. (It may still fail to identify some string-value attributes that require relinking in some cases, but it's still **far** better off with a $genesis widget).

In fact, my only complaint about the genesis widget is this:

### *Why aren't we fully embracing it?*

This PR addresses that by migrating the few remaining places in core where dangerous placeholder widgets were still used.

### DIFF
--- a/core/wiki/macros/list.tid
+++ b/core/wiki/macros/list.tid
@@ -26,7 +26,7 @@ tags: $:/tags/Macro
 \whitespace trim
 <span class="tc-links-draggable-list">
 <$vars targetTiddler="""$tiddler$""" targetField="""$field$""">
-<$type$ class="$class$">
+<$genesis $type=<<__type__>> class="$class$">
 <$list filter="[list[$tiddler$!!$field$]]" emptyMessage=<<__emptyMessage__>>>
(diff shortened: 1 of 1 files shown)


Jermolene:
That's great, thank you @flibbles. I'm deferring merging PRs until the bug fix release of v5.2.5, and in the meantime have added this PR to the list at #7100

Jermolene:
> I _love_ the new genesis widget.

I meant to say thank you for the kind words!

Jermolene:
Thank you for your patience @flibbles

Issue #6684, opened 2022-05-10
[IDEA] Last saved timestamp and what it may enable


AnthonyMuscio:
**Is your feature request related to a problem? Please describe.**
If we return to a TiddlyWiki in a tab after some time it is not possible to establish when it was last saved. This makes some solutions impossible to implement.

**Describe the solution you'd like**
The addition of a tiddler similar to $:/info/startup-timestamp perhaps named $:/info/save-request-timestamp which is updated with a Tiddlywiki serial number date/timestamp before saving just after the save button is selected OR Autosave is actioned but before the save takes place.

The name "save-request-timestamp" is selected intentionally to reflect the fact if the save is not completed due to user intervention or a failure to save it does not misrepresent the timestamps meaning.

**Describe alternatives you've considered**
Building my own package using message catcher or event catcher widgets but always needing to install it in a wiki before it becomes available is inconvenient. The tm-save-wiki is defined within a javascript module.

The value of such a timestamp is increased if it is reliably available in all wikis we come across. For example it allows us to determine how long since an update was applied (the wiki was saved) by the author.

**Additional context**
Having the save-request-timestamp provides information that can support
* Identifying the age since last updated
* Allow timed saves to be triggered after a period of time since last saved
* Determine for how long the wiki has being open since loading $:/info/startup-timestamp and/or since it was last saved.
* With publicly accessible sites it indicates to the Author the last update published time.
* Possibly trigger an action if a given wiki is found unsaved for an extended period. eg: Auto check in, retry checkout, refresh external content and more.
* Possibly warn the user if they attempt to save a Wiki they have had open in there browser for an extended period so they do not save a stale copy.
* Use in conjunction with the Local storage plugin which does not set/reset $:/info/save-request-timestamp identify how long has passed since a save committed local storage changes to disk.
  *  Perhaps local storage could set its own last saved timestamp as long as it succeeds.
* Be able to list tiddlers created modified since last save ie created/modified > $:/info/save-request-timestamp which is different to `[haschanged[]]`

Related feature;
If it were possible to confirm the save but if it fails restore the last value in $:/info/save-request-timestamp and provide information in the wiki eg; set a flag eg "Save mechanisium XYZ failed at timestamp"
* Store information about the specific "save mechanism" that completed the successful save process.

pmario:
>If it were possible to detect a user returning to a tab and trigger an action it would be possible programmatically compare "now" with loaded and saved timestamps and trigger further smart actions.
>
[...]

PR #7131, opened 2022-12-18
Record state saving timestamp

pmario:
This PR will fix: **[IDEA] Last saved timestamp and what it may enable #6684**

Every save action that saves a file-wiki will create 2 new `$:/state/` tiddlers. 

- $:/state/saving/timestamp
- $:/state/saving/lastModifiedSha

Link to more [detailed documentation](https://tiddlywiki5-git-fork-pmario-record-state-savin-654a0f-jermolene.vercel.app/#PreSaveActions)

In short. 

All tiddlers tagged: **$:/tags/PreSaveAction** are invoked, similar to the [StartupActions](https://tiddlywiki.com/#StartupActions)

**$:/config/presave/action/timestamp** creates `$:/state/saving/timestamp` that contains a unix Date [TIMESTAMP](https://tiddlywiki.com/#DateFormat)

```
<$action-setfield $tiddler="$:/state/saving/timestamp" text={{{ [<now "TIMESTAMP">] }}} />
```

**$:/config/presave/action/lastModifiedSha** creates `$:/state/saving/lastModifiedSha` which contains an information about the last modified tiddler as a SHA256. This makes sure that _no_ information about tiddlers with an encrypted wiki are leaked.

```
<$action-setfield $tiddler="$:/state/saving/lastModifiedSha" text={{{ [!is[system]!sort[modified]limit[1]] :map[get[modified]addsuffix<currentTiddler>sha256[64]] }}} />
```

- Those 2 tiddlers make it possible to identify if a file wiki is opened in 2 or more browser tabs. and
- Which one was  modified last
- The startup actions to detect this info are _not_ part of this PR.

**New Hidden Setting**

The PreSaveActions are active by default. If the actions should be disabled the tiddler $:/config/InvokePreSaveActions has to be set to no.

### DIFF
--- a/core/modules/saver-handler.js
+++ b/core/modules/saver-handler.js
@@ -114,6 +114,7 @@ function SaverHandler(options) {
 SaverHandler.prototype.titleSyncFilter = "$:/config/SaverFilter";
 SaverHandler.prototype.titleAutoSave = "$:/config/AutoSave";
 SaverHandler.prototype.titleSavedNotification = "$:/language/Notifications/Save/Done";
+SaverHandler.prototype.saveState = "$:/state/saving/timestamp";
 
 /*
 Select the appropriate saver modules and set them up
@@ -158,9 +159,14 @@ SaverHandler.prototype.saveWiki = function(options) {
 	}
 	var	variables = options.variables || {},
 		template = (options.template || 
-		           this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template","$:/core/save/all")).trim(),
-		downloadType = options.downloadType || "text/plain",
-		text = this.wiki.renderTiddler(downloadType,template,options),
+					this.wiki.getTiddlerText("$:/config/SaveWikiButton/Template","$:/core/save/all")).trim(),
+		downloadType = options.downloadType || "text/plain";
+
+	if (options.recordSaveTime || true) {
+		this.wiki.addTiddler(new $tw.Tiddler({title: this.saveState, text: "" + Date.now()}, this.wiki.getModificationFields()));
+	}
+
+	var text = this.wiki.renderTiddler(downloadType,template,options),
 		callback = function(err) {
 			if(err) {
 				alert($tw.language.getString("Error/WhileSaving") + ":\n\n" + err);
--- a/core/modules/saver-handler.js
+++ b/core/modules/saver-handler.js
@@ -147,7 +147,6 @@ Save the wiki contents. Options are:
 	method: "save", "autosave" or "download"
 	template: the tiddler containing the template to save
 	downloadType: the content type for the saved file
-	invokePreSaveActions: invoke actions tagged: "$:/tags/PreSaveAction". defaults to yes
 */
 SaverHandler.prototype.saveWiki = function(options) {
 	options = options || {};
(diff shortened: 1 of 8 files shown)


saqimtiaz:
I think that it would be better to separate the generic mechanism of pre-save actions into a separate PR, which can be considered independently of saving a timestamp for the last save.

One of the potential issues with pre-save actions is that depending on how they are used, they could modify the store, making the wiki state dirty or even trigger another save.

pmario:
>I think that it would be better to separate the generic mechanism of pre-save actions into a separate PR, which can be considered independently of saving a timestamp for the last save.

The timestamp tiddler for the last save was confirmed by Jeremy at: https://github.com/Jermolene/TiddlyWiki5/issues/6684#issuecomment-1133640239

This PR implements the core functions and the related action to create it.

>One of the potential issues with pre-save actions is that depending on how they are used, they could modify the store, making the wiki state dirty or even trigger another save.

For exactly that reason my first suggestion was hardcoded and therefore doesn't have that problem. See the comment: https://github.com/Jermolene/TiddlyWiki5/issues/6684#issuecomment-1351672650 and comment https://github.com/Jermolene/TiddlyWiki5/issues/6684#issuecomment-1351683249 which shows an image created with the hardcoded implementation.

----

@saqimtiaz 

This PR implements your idea from comment: https://github.com/Jermolene/TiddlyWiki5/issues/6684#issuecomment-1352224407 

>I think a more flexible affordance to invoke actions before saving via tagged tiddlers would be preferable, and would also satisfy the OP.
>
>For instance I would like to update a tiddler before the wiki is saved, specifying when and where a backup was last saved.

@Jermolene 

So the question now is: **Is it too dangerous for normal users to execute arbitrary actions prior to a save as implemented here.** 

If the answer is yes, then I'll close the PR and create a new one that only implements the hardcoded TIMESTAMP as requested and confirmed by issue: **[IDEA] Last saved timestamp and what it may enable** #6684

If the answer is no, we can use this PR to discuss the problems, because I think it contains enough info and test cases to play with the mechanism.

Jermolene:
Thanks @pmario and apologies, I realise that I hadn't replied here or on #6684 with an update on my thinking. I share @saqimtiaz's concern that modifying the tiddler store in response to the save request is brittle because of the possibility that the modification to the tiddler store triggers another save. This is an area in which I think we have to be cautious because there are so many different saver modules out in the field.

So, I'd prefer to add the capability to dynamically insert tiddlers into the save template that do not originate from the store. It would be similar to the way that skinny tiddlers are handled in the current save templates.

The great advantage of this approach is that the saved file no longer has to be a precise replica of the in-memory wiki. So if the save template included a tiddler with a dynamic title that includes a timestamp "Save at 11:32pm on 2023-01-23" then that timestamp tiddler would appear in the saved wiki but not in the in-memory wiki.

It's important to remember that browser semantics mean that in general, from the perspective of a wiki in which the user has clicked the save button, we will never be able to confirm that a save worked successfully except by inspecting the saved file.

pmario:
I'll close this PR and create a new one which injects new tiddlers when the wiki is saved according to Jeremy's response.

pmario:
new pr

- #7953

PR #7143, opened 2022-12-20
Fix miscellaneous typos

wincentbalin:
I ran `aspell --master en_GB --lang en_GB check file` over several `*.tid` files and found more typos.

Also, I found several tiddlers using US spelling (even in their names) and fixed them, but these are not a part of this pull request.


Jermolene:
Wow that's __terrific__ thank you @wincentbalin. A great idea, and a fabulous pay off to have found so many egregious errors! Reading through them I'm horrified how may typos fit a pattern that I often find in my work: "representating" is representative :smile:

> Also, I found several tiddlers using US spelling (even in their names) and fixed them, but these are not a part of this pull request.

Until we have an redirect feature in place for tiddlywiki.com, I'm keen to avoid changing tiddler titles because of the risk of breaking external links to tiddlywiki.com.

PR #7146, opened 2022-12-21
changecount macro change the wording ... just a bit

pmario:
changecount macro change the wording ... just a bit


Jermolene:
Thanks @pmario

PR #7147, opened 2022-12-21
changecount macro change the wording ... just a bit

pmario:
New `$let` widget added to variables are defined by. 

Make **`each snippet is a constant string`** bold. IMO it's the most common error new users make. So I think it should stand out a bit more.


Jermolene:
Thanks @pmario

PR #7152, opened 2022-12-25
Server timing for server-side performance measurement

linonetwo:
Extend $tw.perf to have easy-to-use timer, that can do fine-grained measure.

Result will show in the webDevTool's network panel's timing tab.

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -331,7 +331,7 @@ exports.compileFilter = function(filterString) {
 		})());
 	});
 	// Return a function that applies the operations to a source iterator of tiddler titles
-	var fnMeasured = $tw.perf.measure("filter: " + filterString,function filterFunction(source,widget) {
+	var fnMeasured = $tw.perf.measureFn("filter: " + filterString,function filterFunction(source,widget) {
 		if(!source) {
 			source = self.each;
 		} else if(typeof source === "object") { // Array or hashmap
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -331,7 +331,7 @@ exports.compileFilter = function(filterString) {
 		})());
 	});
 	// Return a function that applies the operations to a source iterator of tiddler titles
-	var fnMeasured = $tw.perf.measureFn("filter",filterString,function filterFunction(source,widget) {
+	var fnMeasured = $tw.perf.measureFn("filter"+filterString,filterString,function filterFunction(source,widget) {
 		if(!source) {
 			source = self.each;
 		} else if(typeof source === "object") { // Array or hashmap
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -331,7 +331,7 @@ exports.compileFilter = function(filterString) {
 		})());
 	});
 	// Return a function that applies the operations to a source iterator of tiddler titles
-	var fnMeasured = $tw.perf.measureFn("filter: " + filterString,function filterFunction(source,widget) {
+	var fnMeasured = $tw.perf.measureFn("filter",filterString,function filterFunction(source,widget) {
 		if(!source) {
 			source = self.each;
 		} else if(typeof source === "object") { // Array or hashmap
--- a/core/modules/server/routes/get-index.js
+++ b/core/modules/server/routes/get-index.js
@@ -17,9 +17,10 @@ exports.method = "GET";
 exports.path = /^\/$/;
 
 exports.handler = function(request,response,state) {
-	var text = state.wiki.renderTiddler(state.server.get("root-render-type"),state.server.get("root-tiddler")),
-		responseHeaders = {
-		"Content-Type": state.server.get("root-serve-type")
+	var text = state.wiki.renderTiddler(state.server.get("root-render-type"),state.server.get("root-tiddler"));
+	var responseHeaders = {
+		"Content-Type": state.server.get("root-serve-type"),
+		"Server-Timing": $tw.perf.generateHeader()
 	};
 	state.sendResponse(200,responseHeaders,text);
 };
--- a/core/modules/server/routes/get-index.js
+++ b/core/modules/server/routes/get-index.js
@@ -20,7 +20,7 @@ exports.handler = function(request,response,state) {
 	var text = state.wiki.renderTiddler(state.server.get("root-render-type"),state.server.get("root-tiddler"));
 	var responseHeaders = {
 		"Content-Type": state.server.get("root-serve-type"),
-		"Server-Timing": $tw.perf.generateHeader()
+		"Server-Timing": $tw.perf.generateHeader() || ""
 	};
 	state.sendResponse(200,responseHeaders,text);
 };
--- a/core/modules/server/routes/get-status.js
+++ b/core/modules/server/routes/get-status.js
@@ -17,6 +17,8 @@ exports.method = "GET";
 exports.path = /^\/status$/;
 
 exports.handler = function(request,response,state) {
+	$tw.perf.reset();
+	$tw.perf.timer("stringify", "JSON.stringify");
 	var text = JSON.stringify({
 		username: state.authenticatedUsername || state.server.get("anon-username") || "",
 		anonymous: !state.authenticatedUsername,
@@ -27,7 +29,8 @@ exports.handler = function(request,response,state) {
 		},
 		tiddlywiki_version: $tw.version
 	});
-	state.sendResponse(200,{"Content-Type": "application/json"},text,"utf8");
+	$tw.perf.timer("stringify");
+	state.sendResponse(200,{"Content-Type": "application/json", "Server-Timing": $tw.perf.generateHeader()},text,"utf8");
 };
 
 }());
--- a/core/modules/server/routes/get-status.js
+++ b/core/modules/server/routes/get-status.js
@@ -30,7 +30,7 @@ exports.handler = function(request,response,state) {
 		tiddlywiki_version: $tw.version
 	});
 	$tw.perf.timer("stringify");
-	state.sendResponse(200,{"Content-Type": "application/json", "Server-Timing": $tw.perf.generateHeader()},text,"utf8");
+	state.sendResponse(200,{"Content-Type": "application/json", "Server-Timing": $tw.perf.generateHeader() || ""},text,"utf8");
 };
 
 }());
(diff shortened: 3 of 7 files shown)


linonetwo:
I'm going to add more `timer` call in core code, that is executed during wiki initialization.

linonetwo:
@Jermolene I found it takes 20s to load a [Using the read-only single tiddler view](https://tiddlywiki.com/#Using%20the%20read-only%20single%20tiddler%20view) on nodejs wiki on my low-end Android phone with termux, and I don't know why. So I hope we can have this advanced server-side timing in the core, so its easier to improve performance in the future.

core/modules/server/routes/get-index.js
```diff
@@ -17,9 +17,10 @@ exports.method = "GET";
 exports.path = /^\/$/;
 
 exports.handler = function(request,response,state) {
-	var text = state.wiki.renderTiddler(state.server.get("root-render-type"),state.server.get("root-tiddler")),
-		responseHeaders = {
-		"Content-Type": state.server.get("root-serve-type")
```
linonetwo, line 22: Add "Server-Timing" here, separate to two `var`

core/modules/filters.js
```diff
@@ -331,7 +331,7 @@ exports.compileFilter = function(filterString) {
 		})());
 	});
 	// Return a function that applies the operations to a source iterator of tiddler titles
-	var fnMeasured = $tw.perf.measure("filter: " + filterString,function filterFunction(source,widget) {
+	var fnMeasured = $tw.perf.measureFn("filter"+filterString,filterString,function filterFunction(source,widget) {
```
linonetwo, line 334: make the name clear

core/modules/server/routes/get-tiddler.js
```diff
@@ -17,13 +17,17 @@ exports.method = "GET";
 exports.path = /^\/recipes\/default\/tiddlers\/(.+)$/;
 
 exports.handler = function(request,response,state) {
+	$tw.perf.reset();
```
linonetwo, line 20: And reset timer, clear measure logs in $tw.perf singleton before each route.

core/modules/utils/performance.js
```diff
@@ -68,14 +76,14 @@ Performance.prototype.log = function() {
 /*
 Wrap performance measurements around a subfunction
 */
-Performance.prototype.measure = function(name,fn) {
+Performance.prototype.measureFn = function(name, description,fn) {
 	var self = this;
 	if(this.enabled) {
-		return function() {
+		return function measureCallback() {
```
linonetwo, line 82: Add a name to the function so it has a name when debugging, this is a best practice from eslint.

core/modules/utils/performance.js
```diff
@@ -68,14 +76,14 @@ Performance.prototype.log = function() {
 /*
 Wrap performance measurements around a subfunction
 */
-Performance.prototype.measure = function(name,fn) {
+Performance.prototype.measureFn = function(name, description,fn) {
```
linonetwo, line 79: Accept and store description that will show in the Server timing

core/modules/utils/performance.js
```diff
@@ -86,6 +94,70 @@ Performance.prototype.measure = function(name,fn) {
 	}
 };
 
+/**
+ * Start a timer, add a record to `this.measures`.
+ * Name has to be slag-separated, like "image-processing"; description is optional.
+ */
+Performance.prototype.timerStart = function(name, description) {
```
linonetwo, line 101: methods that are similar to https://github.com/yosuke-furukawa/server-timing

linonetwo (commented):
Comment for reviewers

core/modules/filters.js
pmario: I think the text-prefix `filter` should be removed completely. It only makes automatic post-processing of the data more complicated.

pmario:
Just seeing the new code I think I don't like it, that the perf measurement adds so much noise to the business logic. The logging stuff is called all over the places. I do have a "bad feeling" .. 

I would prefer new routes eg: "/perf/reset instead of the additional function calls within the code 
May be it's possible to give the whole server-side perfomance testing new routes, that don't effect the "production" business logic at all

linonetwo:
Perf measurement is working like this. 

https://github.com/microsoft/vscode/blob/96bf637c101b8832e44a97a27a26305d8c2cf22e/src/vs/server/node/server.main.ts#L19-L20

Also I hope every user can know why a request is slow by enable perf in control panel, and get this info in developer tool. If adding new routes, this will be not that easy. And we have to maintain copies of code for each route in nodejs.

linonetwo:
And I'd like to add even more perf marks to frontend code, not only the server side. Watch [https://www.youtube.com/watch?v=r0OeHRUCCb4](https://www.youtube.com/watch?v=r0OeHRUCCb4) for how VSCode team speedup their product (to win the editor competition with Atom editor) by adding a lot of perf.mark

pmario:
>So trust me, this is not bad at all.

I can live with your arguments

core/modules/utils/performance.js
Jermolene: We can't rename existing methods without considering the backwards compatibility impact. In this case, I don't think it is worth the disruption.

core/modules/filters.js
Jermolene: Why has the prefix been changed to remove the colon? There is a backwards compatibility impact so we shouldn't change the text unless there's a good reason.

Removing the space around the "+" breaks our coding standards.

core/modules/server/routes/get-index.js
Jermolene: It looks like there are changes on line 20 to drop the semicolon and line 21 to add the "var". Those sort of changes make the PR harder to follow and should be avoided unless there is a good reason.

core/modules/server/routes/get-index.js
```diff
@@ -17,9 +17,10 @@ exports.method = "GET";
...
+	var text = state.wiki.renderTiddler(state.server.get("root-render-type"),state.server.get("root-tiddler"));
+	var responseHeaders = {
+		"Content-Type": state.server.get("root-serve-type"),
+		"Server-Timing": $tw.perf.generateHeader() || ""
```
Jermolene, line 23: I am not keen on  the approach of passing performance data back to the client in all responses. It exposes internal information about the server that could be used to attack it. I think it's safer for server logs to stay on the server in the usual way, and only expose them via a privileged admin interface.

core/modules/server/routes/get-status.js
```diff
@@ -17,6 +17,8 @@ exports.method = "GET";
 exports.path = /^\/status$/;
 
 exports.handler = function(request,response,state) {
+	$tw.perf.reset();
```
Jermolene, line 20: Resetting the performance statistics on each request makes it impossible to analyse timings over multiple runs.

core/modules/server/routes/put-tiddler.js
```diff
@@ -30,21 +32,29 @@ exports.handler = function(request,response,state) {
...
 		if(tiddler) {
 			fields.text = tiddler.fields.text;
 		}
 		delete fields._is_skinny;
+		$tw.perf.timer("get-skinny");
 	}
+	$tw.perf.timer("add-tiddler", "new Tiddler instance and add to wiki");
 	state.wiki.addTiddler(new $tw.Tiddler(fields,{title: title}));
+	$tw.perf.timer("add-tiddler");
+	$tw.perf.timer("count-change", "Construct etag using changed count of tiddler");
```
Jermolene, line 51: As @pmario notes, I do think that the performance code is much too intrusive. I think we might have to settle for measuring at the level of individual requests, even though that loses some granularity from what you have here.

But equally, some of these measurements don't make much sense. For example, `getChangeCount()` is really simple and so it's hard to justify measuring it in isolation.

Jermolene (commented):
Thank you @linonetwo I'm in favour of improving instrumentation but there are some issues with the present implementation.

core/modules/server/routes/get-index.js
linonetwo: This measurement header will only be sent when user toggle this feature in the control panel (`if(!this.enabled) return;`), which can only be opened by privileged admin.

And this is [server-timing](https://www.w3.org/TR/server-timing/), which is meant to send to the client. In this way, it is also easy to use.

core/modules/server/routes/get-status.js
linonetwo: Server timing is only about performance in a single request. For other perf measurements, we can add some other things for it later, which might only output logs on the server side.

core/modules/server/routes/put-tiddler.js
linonetwo: I'm not sure which part of the code is time-consuming, so I part them in this granularity.

We can merge some of them, if you think these codes usually don't consume much time. But we should add as much as possible, to get a result like the screenshot above.

> at the level of individual requests

We can already get this information in devtool, about how long a request will take. Server timing is about make this time more clear, by cut it into multiple parts.

linonetwo:
Comment about the usage of server timing. About code I will change it later, I'm writing a new task management plguin now.

core/modules/server/routes/get-status.js
linonetwo: Maybe you mean this might clear other measurement. Do you think I should move server timing related code to a new util, so it won't affect other perf things?

Jermolene:
Hi @linonetwo I still have the reservations discussed above. I don't see a path to merging this without major rearchitecting.

I wondered if you have any experience of using JS tracing tools. For example, [Google's web testing framework](https://google.github.io/tracing-framework/overview.html) appears to work without requiring source code modifications. I think there are a number of tools in this space that might be worth investigating.

It might make it easier to use those kind of tools if we made one simple change: ensuring that all our anonymous functions are labelled with a meaningful name. It's a big job, because there are a lot of anonymous functions in our code, but it would make a lot of debugging tasks easier. (We can [use eslint](https://eslint.org/docs/latest/rules/func-names) to find them).

linonetwo:
Okay, I can try that tool, if it works, it can save many time of manual labeling. Once I have experience on this, I can write a usage tutorial in dev wiki, to introduce a way for people to fix performance issue in tw.

linonetwo:
I think `import inspector from 'node:inspector';` do the job

https://github.com/tiddly-gittly/TidGi-Desktop/blob/00625ef88ed60ac01f822be4427a74f7c206a67b/src/services/wiki/wikiWorker.ts#L91-L96

PR #7156, opened 2022-12-27
Fixes reduce, filter, and sortsub operators undefined variable exception

btheado:
In some case when the subfilter for the `reduce`, `filter`, and `sortsub` operators contains reference to an undefined variable, a `Linked List only accepts string values, not undefined` exception is thrown.

The `getVariable` override defined in those filter operators was not passing the `{params:varTree.params,defaultValue: ""}` options through to the `widget.getVariable` call and due to that a value of `undefined` was being pushed onto the results array.

Fixes 7155.

Also before this fix, literal macro parameters inside of these three operators did not work correctly.

### DIFF
--- a/core/modules/filters/sortsub.js
+++ b/core/modules/filters/sortsub.js
@@ -26,14 +26,15 @@ exports.sortsub = function(source,operator,options) {
 		var r = filterFn.call(options.wiki,function(iterator) {
 			iterator(options.wiki.getTiddler(title),title);
 		},{
-			getVariable: function(name) {
+			getVariable: function(name,opts) {
+				opts = opts || {};
 				switch(name) {
 					case "currentTiddler":
 						return "" + title;
 					case "..currentTiddler":
 						return options.widget.getVariable("currentTiddler");
 					default:
-						return options.widget.getVariable(name);
+						return options.widget.getVariable(name, opts);
 				}
 			}
 		});
--- a/core/modules/filters/sortsub.js
+++ b/core/modules/filters/sortsub.js
@@ -34,7 +34,7 @@ exports.sortsub = function(source,operator,options) {
 					case "..currentTiddler":
 						return options.widget.getVariable("currentTiddler");
 					default:
-						return options.widget.getVariable(name, opts);
+						return options.widget.getVariable(name,opts);
 				}
 			}
 		});
--- a/core/modules/filters/filter.js
+++ b/core/modules/filters/filter.js
@@ -21,14 +21,15 @@ exports.filter = function(source,operator,options) {
 		target = operator.prefix !== "!";
 	source(function(tiddler,title) {
 		var list = filterFn.call(options.wiki,options.wiki.makeTiddlerIterator([title]),{
-				getVariable: function(name) {
+				getVariable: function(name,opts) {
+					opts = opts || {};
 					switch(name) {
 						case "currentTiddler":
 							return "" + title;
 						case "..currentTiddler":
 							return options.widget.getVariable("currentTiddler");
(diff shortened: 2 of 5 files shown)


pmario:
>In some case when the subfilter for the reduce, filter, and sortsub operators contains reference to an undefined variable, a Linked List only accepts string values, not undefined exception is thrown.

I think this is the hint, where to fix it. IMO it's the linked-list implementation, which is linkedlist.js .. It may / should be possible to fix the problem an it's root. .. IMO the PR is a workaround and not a real fix. ... BUT we'll have to have a closer look at the source first

pmario:
@flibbles ... Would it be possible to eliminate the `throw` from https://github.com/Jermolene/TiddlyWiki5/blob/9f69161709259dda51c01ce081850f6ec1dc3cb4/core/modules/utils/linked-list.js#L188 

Can we make sure, that the `value` "array" only contains strings. We could cast them when they are assigned. ...BUT this would change the logic of the whole thing ...

core/modules/filters/sortsub.js
```diff
@@ -26,14 +26,15 @@ exports.sortsub = function(source,operator,options) {
...
+			getVariable: function(name,opts) {
+				opts = opts || {};
 				switch(name) {
 					case "currentTiddler":
 						return "" + title;
 					case "..currentTiddler":
 						return options.widget.getVariable("currentTiddler");
 					default:
-						return options.widget.getVariable(name);
+						return options.widget.getVariable(name, opts);
```
saqimtiaz, line 37: Please remove the whitespace before the second argument for consistency.

saqimtiaz:
@btheado this looks good to me with the exception of the tiny whitespace correction that I have requested.

@pmario note that this PR brings the code for these filters in conformance with how the corresponding filter run prefixes are implemented for resolving variables. This approach is also needed to resolve literal parameters to macros used in the filter operand. Any potential discussion about changes to the LinkedList implementation are best discussed separately, though I am not sure that any changes are needed.

core/modules/filters/sortsub.js
btheado: Thanks, I fixed it.

btheado:
@pmario, @saqimtiaz, thanks for the feedback.

@saqimtiaz, based on your comment about this fix also being needed for resolving literal parameters, I added some tests which would have failed before the fix.

flibbles:
> @flibbles ... Would it be possible to eliminate the `throw` from

I guess we could. Like you say, it's more of a change in logic than a bug fix. I don't really know what the context here is, but I remember when I made linked-list, Jermolene and I talked about it. At the time, it was understood (but not always enforced) that filters should only deal with string values. Since then, this exception has come up a few times, but to me it always seemed like it was revealing a deficiency with some other filter operator.

Are we certain that isn't the case here? The Python language makes a habit of throwing exceptions when anything is slightly wrong. It's annoying, but it does help to keep code consistent (and to find bugs).

saqimtiaz:
> @saqimtiaz, based on your comment about this fix also being needed for resolving literal parameters, I added some tests which would have failed before the fix.

Thank you @btheado. Much appreciated. 



> At the time, it was understood (but not always enforced) that filters should only deal with string values. Since then, this exception has come up a few times, but to me it always seemed like it was revealing a deficiency with some other filter operator.

@flibbles filter operators should indeed only return strings as they are currently implemented, so I concur that this reveals a deficiency in the implementation of these filter operators. If we do at some point allow operators to return other types then we may need to revisit this at that time.

Jermolene:
Thank you @btheado much appreciated.

PR #7158, opened 2022-12-30
better content type handle

### LINKED ISSUE #7157: [BUG] import some binary file will corrupted the file (eg. mp3 file)
### Describe the bug

import a `.mp3` file, the file can not be play and the file is corrupted (content not encode in base64).

### Expected behavior

the file can be play normally, and is encoded in base64

### To Reproduce

1. get a empty wiki (`empty.html`)
2. drag a .mp3 file to wiki
3. import it
4. the imported file can not be played, and it's not encode in base64.

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: v5.1.24
- Saving mechanism: single file
- Plugins installed: none
- Browser: Chrome 108.0.5359.124 (64bit, Linux), Firefox 108.0 (64bit, Linux)

### Additional context

For `.mp3` file, MIME got from browser is `audio/mpeg` not `audio/mp3`
https://github.com/Jermolene/TiddlyWiki5/blob/0b1a4f3a4d80a58729ce5b399cf0d38b5f29a279/boot/boot.js#L2410

and `wiki.readFile()` will treat as text file, so the file will corrupted.
https://github.com/Jermolene/TiddlyWiki5/blob/35b0833e0cafc477e402309c006a163eb59a94ca/core/modules/wiki.js#L1464-L1465

Some other type will happened too,
so I think we should default to binary if the type is unknown.

I make a plugin using `th-importing-file` hook to fix this issue, but not fully fix,
the `th-importing-file` hook can not get the original `deserializer`,
maybe we should pass it to hook too?
https://github.com/Jermolene/TiddlyWiki5/blob/35b0833e0cafc477e402309c006a163eb59a94ca/core/modules/wiki.js#L1471-L1478

Here is the plugin for partial fix:
[$__plugins_tiddlywiki_fix-content-type-handle.json.gz](https://github.com/Jermolene/TiddlyWiki5/files/10317844/__plugins_tiddlywiki_fix-content-type-handle.json.gz)

Edit: update plugin and try to make a PR for fixing other related part in core

cs8425:
This should fix #7157.
Assume files via import will always have `type` field with a value,
if the type is unknown in `$tw.config.contentTypeInfo`,  then treat it as binary.
If tiddler which did not have `type` field, treat it as `text/vnd.tiddlywiki` just like before.

### DIFF
--- a/core/modules/utils/utils.js
+++ b/core/modules/utils/utils.js
@@ -858,8 +858,8 @@ Convert text and content type to a data URI
 */
 exports.makeDataUri = function(text,type,_canonical_uri) {
 	type = type || "text/vnd.tiddlywiki";
-	var typeInfo = $tw.config.contentTypeInfo[type] || $tw.config.contentTypeInfo["text/plain"],
-		isBase64 = typeInfo.encoding === "base64",
+	var typeInfo = $tw.config.contentTypeInfo[type],
+		isBase64 = (!!typeInfo && typeInfo.encoding === "base64") || !typeInfo,
 		parts = [];
 	if(_canonical_uri) {
 		parts.push(_canonical_uri);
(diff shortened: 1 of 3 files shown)


Jermolene:
Alternate fix in 0b39e47ce88c7620b7a66c1553a71efaff06edb9

PR #7169, opened 2023-01-01
Automatic sync of tiddler changes in browser storage when sync connection is re-established

fkmiec:
The intent of this change is to leverage the browser-storage plugin as a temporary offline storage mechanism when the sync adapter is unable to successfully sync to the server, and to automatically sync tiddlers saved offline in browser storage once connection is re-established. The use of browser storage allows for tiddlers saved offline to be synced with the server even in the case where the browser is closed or the user (purposely, accidentally, or automatically in the case of a mobile browser periodically refreshing) navigates away from the wiki. 

The two changes are:

1. To check browser storage when GetStatus is called and sync the tiddlers in browser storage with the server
2. To remove a tiddler from browser storage when successfully synced to the server

If the browser storage plugin is not present, or is disabled, the changes are inert. 

NOTE - Deletions that occur when offline are not synced. This could be added, but would be more complex and doesn't seem critical for the purpose, which is to prevent loss of information in cases where the user is temporarily disconnected from the server. If the user has deleted some tiddlers while temporarily offline, it is not nearly so frustrating to have to do the delete again as it would be to lose new or modified content.

### DIFF
--- a/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
+++ b/plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
@@ -76,6 +76,14 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
 			if(err) {
 				return callback(err);
 			}
+
+			//If Browser-Storage plugin is present, add pre-loaded tiddlers to the event queue to sync to the server 
+			if($tw.wiki.getTiddlerText("$:/config/BrowserStorage/Enabled") === "yes" && $tw.boot.preloadDirty != null) {
+				$tw.boot.preloadDirty.forEach(function(item){
+					$tw.wiki.enqueueTiddlerEvent(item, false);
+				});
+			}
+
 			// Decode the status JSON
 			var json = null;
 			try {
@@ -211,6 +219,12 @@ TiddlyWebAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 			if(err) {
 				return callback(err);
 			}
+
+			//If Browser-Storage plugin is present, remove tiddler from local storage after successful sync to the server
+			if($tw.wiki.getTiddlerText("$:/config/BrowserStorage/Enabled") === "yes") {
+				window.localStorage.removeItem("tw5#" + window.location.pathname + "#" + tiddler.fields.title);
+			}
+
 			// Save the details of the new revision of the tiddler
 			var etag = request.getResponseHeader("Etag");
 			if(!etag) {


Jermolene (commented):
Thanks @fkmiec, that's very helpful. I'd like to make the coupling between the plugins a bit cleaner. For example, the tiddlywebadaptor should not know about the format used for composing the keys used to store tiddlers in local storage. We might have the browser storage plugin install some methods on an `$tw.browserStorage` object that can be invoked by the tiddlywebadaptor.

Jermolene:
Apologies, I submitted before completing the review; I'll add some notes in a moment.

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -76,6 +76,14 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
 			if(err) {
 				return callback(err);
 			}
+
+			//If Browser-Storage plugin is present, add pre-loaded tiddlers to the event queue to sync to the server 
+			if($tw.wiki.getTiddlerText("$:/config/BrowserStorage/Enabled") === "yes" && $tw.boot.preloadDirty != null) {
```
Jermolene, line 81: With the changes proposed above, this test might be `if($tw.browserStorage && $tw.browserStorage.isEnabled())`

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -76,6 +76,14 @@ TiddlyWebAdaptor.prototype.getStatus = function(callback) {
...
+				$tw.boot.preloadDirty.forEach(function(item){
```
Jermolene, line 82: We use a relatively ancient dialect of approximately ECMAScript 5.1, and so cannot use `.forEach`. We have an alternative utility method:

```
$tw.utils.each($tw.boot.preloadDirty,function(item) {
...
```

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -211,6 +219,12 @@ TiddlyWebAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
 			if(err) {
 				return callback(err);
 			}
+
+			//If Browser-Storage plugin is present, remove tiddler from local storage after successful sync to the server
+			if($tw.wiki.getTiddlerText("$:/config/BrowserStorage/Enabled") === "yes") {
```
Jermolene, line 224: We can rewrite the test here as above.

plugins/tiddlywiki/tiddlyweb/tiddlywebadaptor.js
```diff
@@ -211,6 +219,12 @@ TiddlyWebAdaptor.prototype.saveTiddler = function(tiddler,callback,options) {
...
+				window.localStorage.removeItem("tw5#" + window.location.pathname + "#" + tiddler.fields.title);
```
Jermolene, line 225: Here we can invoke `$tw.browserStorage.removeTiddler(tiddler.fields.title)`

Jermolene (commented):
Thanks @fkmiec

fkmiec:
All make sense. Happy to make the suggested changes. For the proposed
changes to Browser Storage plugin to add a $tw.browserStorage object, can
you point me to a similar pattern to follow in the current codebase?

Jermolene:
Hi @fkmiec 

> All make sense. Happy to make the suggested changes. For the proposed changes to Browser Storage plugin to add a $tw.browserStorage object, can you point me to a similar pattern to follow in the current codebase?

The Dynannotate plugin sets up a `$tw.dynannotate` object for other components to use. In this case we'd be instantiating an instance of a BrowserStorageUtilities class, into which we'd refactor the helper methods currently in `$:/plugins/tiddlywiki/browser-storage/startup.js`.

btheado:
@fkmiec could you explain the steps to follow in order to duplicate the issue you are fixing? The issue I was able to find based on your description doesn't appear to have any better behavior using your code.

I'm using an edition configured via this tiddlywiki.info file:

```json
{
    "description": "Browser storage plugin combined with nodejs sync",
    "plugins": [
        "tiddlywiki/browser-storage",
        "tiddlywiki/filesystem",
        "tiddlywiki/tiddlyweb"
    ],
    "themes": [
        "tiddlywiki/vanilla"
    ],
    "build": {
        "index": [
            "--rendertiddler","$:/core/save/all","index.html","text/plain"]
    }
}
```
Then I follow these steps:

1. Start the nodejs server and load the wiki in the browser.
2. Ensure browser storage is enabled and configure the save filter to `[all[]] -[is[shadow]!is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]]`
3. Stop the nodejs server in order to simulate network disconnect
4. Click the + button to create a new tiddler. Save the New Tiddler. The syncer saver complains that it can't save.
5. Check the browser storage and see the New Tiddler is successfully saved there
6. Close the tab or navigate to some other page
7. Start the server again and visit the page
8. The New Tiddler is no longer present in the wiki. The javascript console  has the message: "syncer-browser-tiddlyweb: Deleting tiddler missing from server: New Tiddler" and "browser-storage: Blanking New Tiddler"

The data from that new tiddler is lost. I thought that might be what you are tring to fix, but when I perform the same steps from your branch, I see the same behavior.

Am I testing your fix wrong?

fkmiec:
@btheado - Sorry. You are testing correctly. It is not working for new tiddlers. I think you'll find that if you try it for a modification, rather than a new tiddler, it will sync the changes from browser storage. The syncer.js code is deleting the new tiddler because it is not in the list returned from getSkinnyTiddlers. I had originally modified the syncer.js code because I was aware of that logic, but I wanted to limit changes to the tiddlywebadaptor.js code. 

I posted an update that fixes that issue by holding the local storage tiddlers temporarily and loading them at the end of getSkinnyTiddlers, after the syncer.js logic is complete. This is, of course, just to get it working. The proposals from Jeremy still have to be implemented.

Jermolene:
Thank you @fkmiec – feel free to ask any questions.

btheado:
> @btheado - Sorry. You are testing correctly. It is not working for new tiddlers. I think you'll find that if you try it for a modification, rather than a new tiddler, it will sync the changes from browser storage.

Thanks. I do see a modification while in disconnected state does persist even after navigating away. However, I'm not seeing the change getting synced to the server. It lives only in browser storage until the next time that tiddler gets modified. At least the change does not get lost, but is that the behavior you intend?

> I posted an update that fixes that issue by holding the local storage tiddlers temporarily and loading them at the end of getSkinnyTiddlers, after the syncer.js logic is complete. This is, of course, just to get it working. The proposals from Jeremy still have to be implemented.

I pulled the latest code, but this still isn't working for me for the new tiddler case.

btheado:
Just thinking out loud here. I wonder if the network disconnected state could be persisted to some tiddler (one which would be saved to browser storage but not to the server).  At startup that tiddler can be used to see if the last known network state was "disconnected". If so, the syncer can trigger different initialization behavior.

Maybe this is already similar to what you are doing? Just that you are triggering the behavior by explicitly looking to see if the browser storage plugin is enabled rather than checking some generic network state tiddler? I'm not sure.

Also, resyncing of tiddlers after the network is reconnected already seems to work pretty well. Is it possible at startup to get the wiki restored from the server without touching the browser storage tiddlers and then trigger the resync functionality?

fkmiec:
@btheado - Thank you. Starting with a clean install, I saw the same behavior as you and was quite confused. Turns out that the Browser Storage plugin creates the $:/config/BrowserStorage/Enabled tiddler only if you actually click the checkbox (even if checked by default initially). So, the conditional check was skipping all the browser storage sync logic. If we are going to be making changes to Browser Storage plugin per some of Jeremy's suggestions, then I'll fix that at the same time so that the config tiddler exists when it's being checked. 

I'm attaching my [console output](https://github.com/Jermolene/TiddlyWiki5/files/10333712/browser-storage-sync-console.txt) with some additional logging that I added showing what I'm seeing and expecting when adding a new tiddler and modifying another offline and then reconnecting. 

> I wonder if the network disconnected state could be persisted to some tiddler (one which would be saved to browser storage but not to the server). At startup that tiddler can be used to see if the last known network state was "disconnected". If so, the syncer can trigger different initialization behavior.

It seems to me that GetStatus is only called when re-establishing the connection, so the sync adapter already picks up on the fact that it had been disconnected. 

> Is it possible at startup to get the wiki restored from the server without touching the browser storage tiddlers and then trigger the resync functionality?

This is effectively what the last change does. It waits for the syncFromServer to complete and then adds the tiddlers that were in browser storage.

fkmiec:
Question - I was originally calling a function to queue events to sync the local tiddlers, but it was necessary to call addTiddler for new tiddlers to sync, so I just call it for all of them. It seems to work. Is there any reason the code should also call the function to queue events for those same tiddlers?

btheado:
> Turns out that the Browser Storage plugin creates the $:/config/BrowserStorage/Enabled tiddler only if you actually click the checkbox (even if checked by default initially).

Good find and that explains my issues. After I un-check and then re-check the browser storage checkbox, it works like I expect for both the create tiddler while disconnected and modify tiddler while disconnected.

Jermolene:
Thanks for your patience @fkmiec. Apologies for the bureaucracy but I've just noticed that I don't think you've signed the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

fkmiec:
> Thanks for your patience @fkmiec. Apologies for the bureaucracy but I've just noticed that I don't think you've signed the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

Perhaps I missed something? I followed the instructions for the individual contributor. Screenshot of cla-individual.md, below with my entry highlighted. 

image

fkmiec:
Regarding the pull request, I believe I've updated the code per your suggestions. I think it's much cleaner. Appreciate any feedback. 

We could leave it as is, which is helpful for not losing new or modified tiddlers while temporarily offline. As noted at the top, it doesn't account for deletions while offline. In fact, the browser storage plugin does account for them, but it is not coordinated with the syncer syncFromServer logic, so it does not have the intended effect when combined with the tiddlyweb adapter and syncer. I think if there were an event or callback from the syncer to signal the end of all actions related to the syncFromServer, it would be possible to coordinate the actions and handle deletions. Thoughts on that idea, and if it's interesting, how best to incorporate such an event in the syncer?

plugins/tiddlywiki/browser-storage/startup.js
```diff
@@ -19,19 +19,28 @@ exports.after = ["startup"];
...
-	DEFAULT_QUOTA_EXCEEDED_ALERT_SUFFIX = "` in browser local storage";
+	SAVE_FILTER_TITLE = "$:/config/BrowserStorage/SaveFilter";
+
+var BrowserStorageUtil = require("$:/plugins/tiddlywiki/browser-storage/util.js").BrowserStorageUtil;
 
 exports.startup = function() {
 	var self = this;
+	
+	// Add ENABLED tiddler with default value "yes"
+	$tw.wiki.addTiddler({title: ENABLED_TITLE, text: "yes"});
```
pmario, line 30: - Did you test, if the configuration can be disabled?
- Is it possible to remove the configuration tiddler?

At the moment the browser storage plugin is disabled by default. @Jermolene ... Do we want to change that?

I think the plugin has to work with sensible defaults if _no_ config tiddler exists at all. That's the standard behaviour for all config tiddlers atm.

plugins/tiddlywiki/browser-storage/util.js
```diff
@@ -0,0 +1,106 @@
...
+type: application/javascript
+module-type: library
+
+Utility methods for browser-storage plugin
+
+\*/
+
+(function(){
+
+    /*jslint node: true, browser: true */
```
Jermolene, line 12: There's one extra level of indentation than we would normally use for a JS module.

Jermolene (commented):
> Perhaps I missed something? I followed the instructions for the individual contributor. Screenshot of cla-individual.md, below with my entry highlighted.

My sincere apologies @fkmiec I was of course looking at the wrong branch.

> We could leave it as is, which is helpful for not losing new or modified tiddlers while temporarily offline. As noted at the top, it doesn't account for deletions while offline. In fact, the browser storage plugin does account for them, but it is not coordinated with the syncer syncFromServer logic, so it does not have the intended effect when combined with the tiddlyweb adapter and syncer. I think if there were an event or callback from the syncer to signal the end of all actions related to the syncFromServer, it would be possible to coordinate the actions and handle deletions. Thoughts on that idea, and if it's interesting, how best to incorporate such an event in the syncer?

That's a promising line of thought, and those would be welcome improvements if you were able to pursue them.

plugins/tiddlywiki/browser-storage/startup.js
fkmiec: Good catch. I added check for existence of the config tiddler. The browser storage plugin is actually enabled by default, but the save filter is very limited in scope. The current default value for the "Use browser local storage" checkbox is "yes", so the config tiddler is created to match. If you delete the config tiddler, the default for the checkbox once again applies.

plugins/tiddlywiki/browser-storage/util.js
fkmiec: I removed the indentation.

fkmiec: I will likely look into the additional change to enable coordinating with the syncer logic, but in my opinion the current pull request meets my original intent. Would it be better to create a separate pull request for the additional change or create a branch from this pull request and change the base branch later, if and when it is suitable for review?

Jermolene: Thanks @fkmiec, I think a separate PR would be easier to manage.

Jermolene:
Thanks @fkmiec and welcome – this is a very consequential first PR, great to see these improvements.

Jermolene:
By the way @fkmiec we need to make some documentation changes to let people know about the new feature, and update the readmes of both plugins. Would you be able to work on that?

fkmiec:
Hi @Jermolene. Sure. I assume that you mean to modify the readme.tid files for the plugins. That should be done as a separate pull request? Or two separate pull requests?

Jermolene:
Thanks @fkmiec 

> Hi @Jermolene. Sure. I assume that you mean to modify the readme.tid files for the plugins.

Yes, that's right. A brief note in the tiddlyweb plugin readme briefly explaining the benefits, and mentioning the browser-storage plugin. Then the browser-storage plugin readme will probably need a more detailed explanation of exactly how things are handled behind the scenes. Perhaps we should also point out the issues around deleting tiddlers when offline.

> That should be done as a separate pull request? Or two separate pull requests?

One PR would be best, and then we can tie it back to the original commit. Our preferred approach is to include docs updates in the original PR but quite often that doesn't happen for various reasons.

PR #7176, opened 2023-01-04
Allow innerwiki $data widget to override tiddler fields

### LINKED ISSUE #7171: [BUG] Innerwiki $data widget with $tiddler attribute does not allow attribute override
### Describe the bug

The documentation for the `$data` widget with the `$tiddler` attribute seems to say that remaining attributes will override the attributes from the original tiddler:
https://github.com/Jermolene/TiddlyWiki5/blob/04810667e6e5cb788b52652e6adb84a1ae0330bc/plugins/tiddlywiki/innerwiki/doc/docs.tid#L35

However, on https://tiddlywiki.com/plugins/tiddlywiki/innerwiki/ if I create the tiddler with this content:

```
<$innerwiki width="1024" height="400" style="width:100%;">
  <$data $tiddler="HelloThere" text=""/>
  <$data title="$:/DefaultTiddlers" text="HelloThere"/>
</$innerwiki>
```
then the HelloThere tiddler in the innerwiki has the text field copied from the outer wiki instead of blanked out by the `text=""` attribute.

The code looks like it is giving precedence to the fields in the original tiddler fields rather than the attributes in the `$data` widget:
https://github.com/Jermolene/TiddlyWiki5/blob/04810667e6e5cb788b52652e6adb84a1ae0330bc/plugins/tiddlywiki/innerwiki/innerwiki.js#L292-L299

### Expected behavior

It should be possible to override the tiddler fields when using `<$data $tiddler/>`

### To Reproduce

1. Go to https://tiddlywiki.com/plugins/tiddlywiki/innerwiki
2. Add a new tiddler with the content from this issue's description
3. The body of the HelloThere tiddler should be blank in the innerwiki but it is not.

### Screenshots

_No response_

### TiddlyWiki Configuration

browser TW version 5.2.5

### Additional context

_No response_

btheado:
Change the innerwiki plugin's $data widget so it allows the fields of an existing tiddler to be overridden using attributes, as documented.

Fixes #7171

### DIFF
--- a/plugins/tiddlywiki/innerwiki/innerwiki.js
+++ b/plugins/tiddlywiki/innerwiki/innerwiki.js
@@ -295,7 +295,7 @@ InnerWikiWidget.prototype.readTiddlerDataWidget = function(dataWidget) {
 			if(tiddler) {
 				fields = tiddler.getFieldStrings();
 			}
-			results.push($tw.utils.extend({},item,fields));
+			results.push($tw.utils.extend({},fields,item));
 		})
 		return results;
 	} else {


btheado:
@Jermolene, any thoughts on this PR? It is the same change you suggested and in my testing it worked for me.

Jermolene:
Thanks for the reminder @btheado.

PR #7180, opened 2023-01-04
Use British English

wincentbalin:
Fixed some typos too.


wincentbalin:
Like I suggested in #7143, but without renaming tiddlers as well as links pointing to them.

Jermolene:
Thank you @wincentbalin and apologies for the delay thanks to the holidays. This is another very much appreciated improvement; the docs more consistent is really helpful.

PR #7185, opened 2023-01-06
Adds support for indenting pragmas with whitespace

saqimtiaz:
This PR adds support for indenting pragmas with whitespace, which helps readability particularly in nested macros.

With the way that pragmas are parsed with a preceding call to skipWhitespace(), it turned out that the only change needed to the regular expressions was to drop the requirement to start with a new line. 

https://github.com/Jermolene/TiddlyWiki5/blob/f3ebb258f79fdd03478bd5a27060f9ac4ec8a59d/core/modules/parsers/wikiparser/wikiparser.js#L197-L203

@Jermolene Are there any caveats here that you can think of?

### DIFF
--- a/core/modules/parsers/wikiparser/rules/import.js
+++ b/core/modules/parsers/wikiparser/rules/import.js
@@ -25,7 +25,7 @@ Instantiate parse rule
 exports.init = function(parser) {
 	this.parser = parser;
 	// Regexp to match
-	this.matchRegExp = /^\\import[^\S\n]/mg;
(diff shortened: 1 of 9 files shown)


Jermolene:
Thanks @saqimtiaz, looks good.

saqimtiaz:
I have added a brief note to the documentation as well. @Jermolene this is ready for review.

Jermolene:
Thanks @saqimtiaz

PR #7187, opened 2023-01-07
Remove the tag to itself on the "Open collective" tiddler

AnthonyMuscio:
Remove [[Open Collective]] from the tags of itself.

This breaks the heirachy of the TOC and produces an entry that will not open its own list see HelloThere > Open Collective

Mentioned here https://talk.tiddlywiki.org/t/filter-question-return-all-toc-tiddler-titles-as-a-plain-list/5752/9?u=tw_tones


Jermolene:
Thanks @AnthonyMuscio

PR #7190, opened 2023-01-07
Add a debug-log filter operator

### LINKED ISSUE #7184: [IDEA] Debug Logging for Filters
I'm sorry if this is a duplicate, but I have searched and found nothing.

I would _really_ like a logging functionality for filters, akin to `$log` and `$action-log`. I could imagine a `log[]` filter operator that does nothing but print the list of input tiddlers to the console. If one wanted to go further, the log could take variable names (like `currentTiddler` or `..currentTiddler`) as parameter(s) and log their value, too.

Filters can be a complex beast, especially with subfilters, reduce, map, especially where operators are defined elsewhere in the code. Such a debug logger could help a lot.

Have a nice day
Yaisog

yaisog:
The `debug-log` filter operator will output its input and one optional variable value (passed as parameter) to the console, to aid in debugging more complex filter runs.

The output is a numbered list of entries. I initially thought about stringifying and concatenating the output, so that it looks more like that of $log and $action-log and goes into a single row, but then empty elements would get dropped. I think it is _very_ important that \<empty value\>s are kept for debugging.

The filter operator name seems to be still open for debate. Docs will be added.

Closes #7184


yaisog:
Be gentle with me, this is my first PR.

linonetwo:
I think this is cool, can you also add some doc to `editions/tw5.com` ? Like other filter operators?

yaisog:
> can you also add some doc to editions/tw5.com ?

Yes, I will do that ASAP.

kookma:
@yaisog where is the demo to give a try?

Jermolene:
Hi @kookma for reasons I don't understand, the Vercel preview deployment hasn't occurred for this PR.

@yaisog perhaps you could try pushing another commit to this branch to try to trigger the build process?

yaisog:
I'll try to get the docs done by Thursday at the latest, which will be a new commit.

core/modules/filters/debug-log-js
```diff
@@ -0,0 +1,36 @@
...
+	var results = [], data = {};
+	
+	if (operator.operand)
+		data[operator.operand] = options.widget.getVariable(operator.operand,{defaultValue:""});
+
+	source(function(tiddler,title) {
+		results.push(title);
+	});
+	
+	console.group("filter-debug");
```
pmario, line 28: IMO if new logging functionality is needed, $:/core/modules/utils/logger.js needs to be extended. IMO `console` could be "undefined"

pmario:
@yaisog ... IMO the preview doesn't work for me. It seems the new operator is missing. I can't find it in browser debug mode.

core/modules/filters/debug-log-js
linonetwo: Yes, seems we can use core logger here, and add new group method to the core logger.

linonetwo: Yes, seems we can use core logger here, and add new group method to the core logger.

yaisog:
Duh, I had a typo in the filename. Fixed.
Like I said in the issue, this will be and is a copy-paste job. I copied the logging code from $log and $action-log. At this point I don't feel able or have the time to make deeper changes to fundamental core libraries. But do feel free to take it from here.
For me it does work reliably and well, but I can only test on Firefox / Windows.

yaisog:
@linonetwo, @pmario, @kookma: Just a ping that the demo site works now.

pmario:
looks good to me

core/modules/filters/debug-log.js
```diff
@@ -0,0 +1,36 @@
...
+	var results = [], data = {};
+
+	if (operator.operand)
+		data[operator.operand] = options.widget.getVariable(operator.operand,{defaultValue:""});
+
+	source(function(tiddler,title) {
+		results.push(title);
+	});
+
+	console.group("filter-debug");
```
pmario, line 28: may be the group should also be named `debug-log` so we can see a connection

editions/tw5.com/tiddlers/filters/debug-log.tid
```diff
@@ -0,0 +1,13 @@
...
+op-input: a [[selection of titles|Title Selection]]
+op-output: the original input titles
+op-parameter: name of a variable
+op-parameter-name: V
+op-purpose: prints a table of the input titles to the console; if optional parameter <<.place V>> is specified, also prints the value of the variable with that name
+tags: [[Conditional Operators]] [[Filter Operators]]
+title: debug-log Operator
+type: text/vnd.tiddlywiki
+
+<<.from-version "5.2.6">>
```
pmario, line 13: IMO there should be some more text, that the `debug-log` operator should be removed from the working filter expression. Otherwise it will only slow down the execution and pollute the dev console

core/modules/filters/debug-log.js
yaisog: That sounds sensible.

core/modules/filters/debug-log-js
linonetwo: @yaisog will you do this? Do you need help on this?

editions/tw5.com/tiddlers/filters/debug-log.tid
yaisog: I have added a longer explanation of the filter's purpose that leans on the description of $action-log. I hope it's not too verbose.
It also contains a reminder to clean up.

core/modules/filters/debug-log-js
yaisog: I don't think I can do this. Maybe with some help? I have no idea what this is about.

core/modules/filters/debug-log.js
```diff
@@ -0,0 +1,36 @@
...
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports["debug-log"] = function(source,operator,options) {
+	var results = [], data = {};
+
+	if (operator.operand)
```
Jermolene, line 21: There should be no whitespace between `if` and the following parenthesis.

core/modules/filters/debug-log.js
```diff
@@ -0,0 +1,36 @@
...
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+/*
+Export our filter function
+*/
+exports["debug-log"] = function(source,operator,options) {
+	var results = [], data = {};
+
```
Jermolene, line 20: Please avoid blank lines unless they are needed to improve legibility.

core/modules/filters/debug-log.js
```diff
@@ -0,0 +1,36 @@
...
+	if (operator.operand)
+		data[operator.operand] = options.widget.getVariable(operator.operand,{defaultValue:""});
+
+	source(function(tiddler,title) {
+		results.push(title);
+	});
+
+	console.group("debug-log");
+	$tw.utils.logTable(results);
+	if (operator.operand) $tw.utils.logTable(data);
```
Jermolene, line 30: We always use braces with if statements:

```
 	if(operator.operand) {
		$tw.utils.logTable(data);
	}
```

Jermolene (commented):
Thanks @yaisog just a couple of points about coding style – we use some formatting that is now considered slightly eccentric but it's important to try to keep the codebase consistent.

Jermolene:
Hi @yaisog I'd like to pause before merging this. While it's a nice small change that doesn't have much impact on existing code, it does suffer from a couple of disadvantages:

* The requirement to add the `debug-log` operator to a filter before it can be debugged
* The requirement to remove it again afterwards

One alternative approach that would avoid having to directly modify filters would be to instead have the user register the text of filter expressions that are to be debugged. When a matching filter is compiled, it would add additional code to log the values after every step of the filter.

There are also some wider questions about debugging features that it would be good to think through, particularly how to make log output available to users with browsers that don't support the developer console (such as iPhones/iPads). We might want to consider adding a "debug mode" that displays an interactive copy of the widget tree in a separate window.

Generally, this is an area where I think we could really benefit from some improvements. The challenge is how to explore new techniques without committing ourselves to lifetime support in the core.

pmario:
>One alternative approach that would avoid having to directly modify filters would be to instead have the user register the text of filter expressions that are to be debugged. When a matching filter is compiled, it would add additional code to log the values after every step of the filter.

I think that won't help much if we want to debug a filter string that is "popular". Everytime such a filter is triggered somewhere in the UI the info will be logged. eg: `[tag[$:/tags/xxx]]`

That's exactly the problem we have at the moment using breakpoints in the core [filter compiler code](https://github.com/pmario/TiddlyWiki5/blob/fdade8d232945d5c84c42d8f0181586c6852cac3/core/modules/filters.js#L222). Not only is it close to impossible to debug the code there because the stack mainly shows anonymous function calls ... It also triggers 100 times, which makes this approach completely useless. 

**IMO the `debug-log[]` operator has a very clear advantage here.** It's easy to set a breakpoint in the module code and be certain to be triggered with the data needed and the exact location, where the operator is used.

Jermolene:
> I think that won't help much if we want to debug a filter string that is "popular". Everytime such a filter is triggered somewhere in the UI the info will be logged. eg: `[tag[$:/tags/xxx]]`

At worst, one could hack the filter string to make it unique. That's no worse than adding the debug-log filter operator, and considerably less invasive.

> Not only is it close to impossible to debug the code there because the stack mainly shows anonymous function calls ...

I've no objection to adding some function names to aid debugging.

> It also triggers 100 times, which makes this approach completely useless.

I do it all the time. The key is to use conditional breakpoints to test for the data of interest.

pmario:
>One alternative approach that would avoid having to directly modify filters would be to instead have the user register the text of filter expressions that are to be debugged. 

A second concern I do have here is the extra weight, which is always there but not needed in production.

The current PR only executes the code if it is part of the filter string. For everyone reading the string it will be obvious, what's going on. .. If the registration is in a config tiddler 99% of the user won't see that limitation at all.

yaisog:
@Jermolene: I appreciate your input and I'll update the code according to the guidelines when the PR is "unpaused".

Regarding the workflow: I think this is very close to what `$log` and `$action-log` do, where you get console output until you remove the log lines from the code. So I'm used to that kind of workflow. It is also very straightforward and not difficult to explain to non-power users. Personally, I don't want to have to register some string somewhere else to see what my filter is doing. And re-register everything every time I make a change to the filter in the debugging process? This is exactly why I made this.

Also, is there a reasonable number of users coding WikiText on an iPhone? I wouldn't, but maybe I'm just old-school...

Anyway, I'm OK with it if you want to pause this. I got it running in my wikis, and I'm using it regularly, together with its cousins, the $log brothers. Just ping me if you change your mind.

Have a nice day
Yaisog

Jermolene:
One thing we notice with other debugging environments is that they tend to support debugging commands both intertwingled with the code, and via an alternate out-of-band technique. For example, in JS breakpoints can be triggered with the `debugger;` or via the UI. Those non-invasive techniques are very useful in a lot of situations.

> Personally, I don't want to have to register some string somewhere else to see what my filter is doing. And re-register everything every time I make a change to the filter in the debugging process? This is exactly why I made this.

My point in raising that alternative was not to suggest that we pick between the two proposals, but just to illustrate that different approaches were possible.

But I don't think it needs to be as painful as you assume. Imagine, for instance, a "debug" tab in control panel with a checkbox to switch on debugging mode. When enabled, it could show a list of all of the filters that have been encountered, and let the user interact with them directly to mark them for debugging, or to copy them to $:/AdvancedSearch for more investigation. There would be a lot of useful capabilities there that would be hard to replicate with `debug-log`.

> Also, is there a reasonable number of users coding WikiText on an iPhone? I wouldn't, but maybe I'm just old-school...

For me it would be more a matter of the iPad, but I know that there are people out there who use their phone as their main coding device. (I get a taste of that when I connect my phone to a Bluetooth keyboard; it has a biggish screen, and so I end up with more usable screenspace than I had on desktop computers 25 years ago).

AnthonyMuscio:
If you do extend the debugging I made a set of tools that may inform the changes to concider;

First have a convention of $:/config/debug-mode = yes to activate, then special features and in wiki code can be written to respond to the debug status.
* Ideally the aforementioned operator could respond to this, so its fine if its left in the code. 
* This would be a useful convention to add other "mode" config tiddlers eg wiki, author, design and other modes, we could call this a mode or settings cascade
  * $:/config/debug-mode = yes
  * A corresponding field on any tiddler eg debug-mode=yes could override the global $:/config/debug-mode=no
  * A corresponding variable `<<debug-mode>>` could override the one set in the current tiddler field.

Dump variables is useful for debugging but it is too verbose. This version of min makes it collapsible and indicates the byte size of the variable, this too could respond to the above debug-mode - mode or settings cascade.

[my-custom-dumpvariables.zip](https://github.com/Jermolene/TiddlyWiki5/files/10452226/my-custom-dumpvariables.zip)

I mention this here because of the idea of extending the debug process was raised.

pmario:
>...or to copy them to $:/AdvancedSearch for more investigation

That would be interesting especially for filters that contain variables. .. About 40% of the tiddlers used in TW contain variables.

pmario:
I did test a filter, which is complex and where I think there is potential for optimisation. 

```
[all[tiddlers+shadows]tag[$:/tags/MenuBar]] +[debug-log[]] -[all[tiddlers+shadows]prefix[$:/config/plugins/menubar/MenuItems/Visibility/]regexp:text[hide]debug-log[]removeprefix[$:/config/plugins/menubar/MenuItems/Visibility/]] +[debug-log[]] -[all[tiddlers+shadows]tag[$:/tags/TopLeftBar]debug-log[]limit[1]then[]else[$:/plugins/tiddlywiki/menubar/items/topleftbar]] +[debug-log[]] -[all[tiddlers+shadows]tag[$:/tags/TopRightBar]debug-log[]limit[1]then[$:/plugins/tiddlywiki/menubar/items/toprightbar]] +[debug-log[]] -$:/plugins/tiddlywiki/menubar/items/hamburger +[debug-log[]] +[limit[1]] +[debug-log[]] 
```

I needed a lot of `debug-log[]` **points** to try to make sense of the construction. ... The problem with the output in the console.log is, that there is no relation between the output text and the position of the "log trigger" 

So I would like 2 things. 

- I would like to have a **suffix** to be a "name" that is used in the console to identify the context eg: `debug-log:XX[]`
  - So the console would output XX instead of `debug-log` table
- I would like to have a shorter version for the `debug-log[]` eg: `DL[]` ... so it would be much less to type, but still stand out much more 

The debug-filter with my proposal would look as follows. IMO it's much more obvious where the debug points are. 

```
[all[tiddlers+shadows]tag[$:/tags/MenuBar]] +[DL:X1[]] -[all[tiddlers+shadows]prefix[$:/config/plugins/menubar/MenuItems/Visibility/]regexp:text[hide]DL:X2[]removeprefix[$:/config/plugins/menubar/MenuItems/Visibility/]] +[DL:X3[]] -[all[tiddlers+shadows]tag[$:/tags/TopLeftBar]DL:X4[]limit[1]then[]else[$:/plugins/tiddlywiki/menubar/items/topleftbar]] +[DL:X5[]] -[all[tiddlers+shadows]tag[$:/tags/TopRightBar]DL:X6[]limit[1]then[$:/plugins/tiddlywiki/menubar/items/toprightbar]] +[DL:X7[]] -$:/plugins/tiddlywiki/menubar/items/hamburger +[DL:X8[]] +[limit[1]] +[DL:X9[]] 
```

@Jermolene .... It would be interesting how you would want to solve a configuration as discussed here with a global "registration" option. ... IMO it would completely flood the console.

pmario:
A version, that stands out even more would be `[XX:NAME[variable]]`

`[all[tiddlers+shadows]tag[$:/tags/MenuBar]] +[XX:X1[]] -[all[tiddlers+shadows]prefix[$:/config/plugins/menubar/MenuItems/Visibility/]regexp:text[hide]XX:X2[]removeprefix[$:/config/plugins/menubar/MenuItems/Visibility/]] +[XX:X3[]] -[all[tiddlers+shadows]tag[$:/tags/TopLeftBar]XX:X4[]limit[1]then[]else[$:/plugins/tiddlywiki/menubar/items/topleftbar]] +[XX:X5[]] -[all[tiddlers+shadows]tag[$:/tags/TopRightBar]XX:X6[]limit[1]then[$:/plugins/tiddlywiki/menubar/items/toprightbar]] +[XX:X7[]] -$:/plugins/tiddlywiki/menubar/items/hamburger +[XX:X8[]] +[limit[1]] +[XX:X9[]] `

Jermolene:
> @Jermolene .... It would be interesting how you would want to solve a configuration as discussed here with a global "registration" option. ... IMO it would completely flood the console.

The proposal doesn't say anything about the console; in fact one of the rationales that I mentioned for the idea was that it did not use the developer console. The idea would be provide a GUI.

yaisog:
> I would like to have a **suffix** to be a "name" that is used in the console to identify the context eg: `debug-log:XX[]`
> So the console would output XX instead of `debug-log` as the name of the console.group
That's a great idea! I've also wondered how to best tell multiple debug-logs apart. I never thought about the suffix, because often the suffix has to be from a list of very few fixed options. I'll implement this ASAP.
> I would like to have a shorter version for the `debug-log[]` eg: `DL[]` ... so it would be much less to type, but still stand out much more
I get what you are saying, but filter operator names so far have all been lowercase. The `$action-log` widget is also not just called `$AL` in order to stand out. Could I just export an alias in the filtermodule JS? Something like `exports.DL = debug-log`... sort of like an "insider" shortcut.

kookma:
> I get what you are saying, but filter operator names so far have all been lowercase. The `$action-log` widget is also not just called

I support you here! The filters are all lowercase with full name. So, it is better to stick to current rule.

pmario:
>I support you here! The filters are all lowercase with full name. So, it is better to stick to current rule.

Yes, but the intention of an uppercase DL operator is to stand out. It needs to be removed after it has been used. So it needs to stand out to easily find it again. See: https://github.com/Jermolene/TiddlyWiki5/pull/7190#issuecomment-1396890098

yaisog:
@Jermolene: After using this for a while, I think you should maybe reconsider your position on not wanting to merge this.

I found the debug filter extremely useful on multiple occasions when writing up more complex filter runs (and even some short ones). It's a bit of a _bird in the hand_ thing, I guess. I wish I would have had this thing years ago, could have saved me so much trouble. Who knows how many years we'll have to wait for your alternative to materialize? I don't see this being a priority in the list of 1 k open issues. Meanwhile, people are struggling with the intricacies of filter programming...

Also, if you're doing this to move away from the console as a debugging tool, you'd have to provide alternatives for `$action-log` and `$log`, too. Whatever these are, together with a filter debugger they will address mostly mobile users. As a desktop user, I'll always prefer my debugging output on a second screen instead of somewhere in the wiki where it takes up screen real estate. And the console is always only one F12 away. So, these mobile-friendly debuggers might be just a plugin in the end, since they address only a small group of users?

Jermolene:
> @Jermolene: After using this for a while, I think you should maybe reconsider your position on not wanting to merge this.

Thanks @yaisog I do understand, but the issue here is not whether the debug-log filter is useful, it's whether it is the best we can do for the core.

> Also, if you're doing this to move away from the console as a debugging tool, you'd have to provide alternatives for `$action-log` and `$log`, too. Whatever these are, together with a filter debugger they will address mostly mobile users. As a desktop user, I'll always prefer my debugging output on a second screen instead of somewhere in the wiki where it takes up screen real estate. And the console is always only one F12 away. So, these mobile-friendly debuggers might be just a plugin in the end, since they address only a small group of users?

Our primary audience for TiddlyWiki is not people who already have software development experience.

If we were to move the debug infrastructure into TiddlyWiki itself just as important a motivation would be that it would make the debugging infrastructure hackable and customisable; the developer console isn't really customisable by regular users.

yaisog:
OK, I guess you're more the _two in the bush_ type, then. ;-P

I guess I'll just promote the debug-log filter plugin on TWTalk a bit better (right now it's hidden in an unrelated thread), so that at least the folks frequenting there have something they can work with _now_...

Do you have something like a roadmap or master plan for these bigger features and their priority? I know, currently it's all PR 6666, but thereafter?

Jermolene:
> OK, I guess you're more the _two in the bush_ type, then. ;-P

That characterisation ignores the brutal calculus of backwards compatibility.

> I guess I'll just promote the debug-log filter plugin on TWTalk a bit better (right now it's hidden in an unrelated thread), so that at least the folks frequenting there have something they can work with _now_...

Plugins are the lifeblood of the community, and how functionality is intended to be shared.

Sometimes people comment on a new development by saying "that's great, it should be in the core", as if being in the core were a mark of distinction. But, from the developers perspective moving things into the core makes the process inevitably slower and limits the ability to make improvements that are not backwards compatible.

The criteria for inclusion in the core isn't excellence, it's *universality*.
 
> Do you have something like a roadmap or master plan for these bigger features and their priority? I know, currently it's all PR 6666, but thereafter?

Like anything else, these improvements emerge from dialogue within the community. We're working on it now with this conversation.

yaisog:
> That characterisation ignores the brutal calculus of backwards compatibility.

That is the best way I've seen this put in. Ever. I need to copy that into my bag of quotes... =)

> Plugins are the lifeblood of the community, and how functionality is intended to be shared.

I'm sorry if I'm not totally up to speed here, but I am under the impression that plugin repositories are very distributed, with no central listing, and it's hard to find something if you're not sure who wrote it.

Take Sublime with its Package Control, for example. That lists many hundreds of packages (I didn't find an exact count), and for everything you might want to do, there seem to be at least three different packages that do that, which is a different problem. TW.com lists 63 plugins, not one of them a filter operator. How would users find this "lifeblood of the community"? TWTalk also has no comprehensive list, though you can find some threads with lists of personal repositories.

> The criteria for inclusion in the core isn't excellence, it's _universality_.

That's another great quote. I get the feeling you had a bunch of previous discussions just like this one before...
But also, I'm not arguing because I want my code in the core just for the sake of it. I actually do think that this would help many people, some of which might never find it on TWTalk. Building filters was hard when I started out, and I would've taken any help I could get. Back then there was only Google Groups. But I get that you would not want the core spammed will all kinds of filter operators, so I'll leave you alone already.

Jermolene:
Thank you @yaisog, much appreciated. These policies are abraded out of the rock face by the wind; I'm hardly aware of them until they are fully formed.

We could probably do a better job of writing them policies down (feel free to have a hack at contributing.md to see if you can add some of the answers that you'd have found useful (you'll need to edit the wikitext sources, not the MD file)). 

As you note, one big thing that the TiddlyWiki community needs is a better way to find and share plugins. But that's orthogonal to the issues around what goes into the core.

linonetwo:
This debug-log plugin is very useful during development. I now always install them in my dev environment (for example https://github.com/tiddly-gittly/TiddlyWiki-TS-Plugin-Template and https://github.com/tiddly-gittly/Modern.TiddlyDev ). And I can always find it in the CPL.

So I'm not very care about if it is in the core now...Thought it is super useful.

yaisog:
Thanks for your nice words. As you can read above, Jeremy has other plans to make debugging filters comfortable.

That reminds me that I did want to make a separate post about the debugging filter on TWTalk - right now it is buried in an unrelated discussion somewhere, where almost noone can find it if he/she is not specifically searching for it...

yaisog:
Hi @Jermolene, should I close this? If your "no" is hard and final, I don't think it makes sense for this PR to linger around. If you want to meditate on it a bit more, I can leave it open, though...

yaisog:
I am closing this in favor of @Jermolene's upcoming implementation of a filter debugging UI.

linonetwo:
BTY, debug log filter could be download from https://yaisog.tiddlyhost.com/

I find it hard to debug filter when writing PR for tw. I have to use this.

pmario:
@yaisog ... The debug-log operator may need an update to deal with multi-value-variables in filters. https://tiddlywiki.com/prerelease/#Multi-Valued%20Variables eg:  `[tag[HelloThere]] =>x [debug-log(x)]` needs to show all tiddlers tagged HelloThere.

linonetwo:
@pmario It already shows all values.

图片

When I'm testing https://github.com/TiddlyWiki/TiddlyWiki5/discussions/9464

yaisog:
> The debug-log operator may need an update to deal with multi-value-variables in filters

The parameter of `debug-log[]` is a variable name. If that variable contains a list, that list is shown. That already works with MVVs.

> [tag[HelloThere]] =>x [debug-log(x)]

That doesn't work like you probably think it does. To show all tiddler titles stored in `x`, you'd use either `[(x)debug-log[]]` or `[debug-log[x]` (without parentheses). Your syntax would presumably try to show the contents of all variables with names that correspond to tiddlers tagged "HelloThere". Though it would currently only show the first one, I guess.

A useful (?) extension to MVVs would be to be able to put a list of variable names into another variable, e.g. "variableList", and then show the contents of all those variables at the same time using `[debug-log(variableList)`]. This might be useful in some situations, when e.g. defining a global debugging watchlist with `$let`, though I don't think this will be used much. It's probably quicker to just chain a few `XXX[variable1]XXX[variable2]` operators back-to-back.

PR #7191, opened 2023-01-07
Added documentation for neglected core fields

flibbles:
This is just documentation for a few tiddler fields which are used explicitly by the core code, or as part of plugins which seem to have "slipped the net".

My reason for adding this documentation is also slightly selfish. Relink-fieldnames uses the $:/language/Docs/Fields/ tiddlers to figure out which fields are built-in fields, and thus should never be reported or relinked (because relinking would be destructive in those cases).


Jermolene:
Thank you @flibbles, always good to improve consistency and completeness.

PR #7197, opened 2023-01-10
Feat/button default layout

linonetwo:
I'm creating several layout plugins, and I want a standard way for users of alternative layouts to back to the default layout, or switch between other layouts.

I introduce a new tag `$:/tags/LayoutSwitcherButton` for a plugin author to get these buttons. These buttons can even be used to go to a specific state of a layout (for example, 3-day view of the calendar layout or a week view of the calendar layout).

Though not forced, but I hope every layout author can add these buttons to the new layout.


linonetwo:
Why this button is enabled by default?

Jermolene:
> Why this button is enabled by default?

Buttons that are to be disabled by default must be listed in https://github.com/Jermolene/TiddlyWiki5/blob/master/core/wiki/config/PageControlButtons.multids with the value "hide".

linonetwo:
Done here, also add doc about this new system tag.

Do you think the icon is OK? I pick this one because it looks like leaving the custom layout.

Jermolene:
Hi @linonetwo it's a good idea. The wording should perhaps be "Reset layout to default".

But I wonder if we might go further and make this a general reset button, that doesn't just reset the layout but also the theme and palette etc. We might give it a keyboard shortcut too.

If that doesn't work, then my concern would be that if we introduce a "reset layout" button then we can also make a case for "reset theme", "reset palette" etc., all of which risks making the page control button list get much longer, which makes it a lot harder for users to scan and understand.

linonetwo:
Those can be separate buttons, and hide by default. But they won't be used as frequently as resetLayout button. Because people may stick to a palette and theme for months, but will switch between layout (default -> calendar -> whiteboard -> default) several times in a hour.

linonetwo:
截屏2023-01-11 11 44 36
截屏2023-01-11 11 44 32

Jermolene:
Hi @linonetwo can you explain the reason for introducing the new `$:/tags/LayoutSwitcherButton` tag? It doesn't seem to be needed by the core itself.

Is the overall objective here to build an alternative layout switcher that features image buttons for each available layout?

Please understand that there is quite a high bar for adding a new page control button. TiddlyWiki would become unusable if there were too many of them. It's not clear how many is "too many" so we are super cautious about adding new ones.

linonetwo:
> reason for introducing the new `$:/tags/LayoutSwitcherButton` tag

In an alternative layout, we might need a dedicated nav bar to switch between layouts. So it is easier to draw something after add new event in the calendar.

We can't just use PageControlButtons in alternative layout, because some of them may not work or functionality is undefined, like the home button and controlpanel button will goto nowhere in calendar (Though I still hope I can use them to go back to the default layout before they work).

So I hope I can create a new bar using `[tag[$:/tags/LayoutSwitcherButton]]` in every alternative layout, only show these buttons.

linonetwo:
> there is quite a high bar for adding a new page control button

That is true, and only few of them will be added to the PageControl. What about make these kind of setting list become a paginated & filtered list, like https://morosanuae.github.io/tw-icons/

截屏2023-01-12 01 44 17
截屏2023-01-12 01 44 24

We can solve the problem of having massive items using these UI elements, what do you think?

Jermolene:
> What about make these kind of setting list become a paginated & filtered list, like https://morosanuae.github.io/tw-icons/ , which will hide most of the unimportant items in the end of the list.

> We can solve the problem of having massive items using these UI elements, what do you think?

It doesn't really solve the problem. Hiding some of the buttons behind a button click just makes it harder to visually scan the buttons to find the one you want.

> And the `$:/tags/LayoutSwitcherButton` can also solves this, if we will have 4-10 layouts in the future, adding all of them to pagecontrol button list maybe unacceptable. So we can use this new tag to create a new UI for them.

I think that at this point this functionality should stay as a plugin. The logic of resetting the layout is trivial, so there's little to be gained in terms of avoiding code duplication. The risk of doing this is that we might later come up with a better approach than `$:/tags/LayoutSwitcherButton`, and yet if we implement it now we'll be stuck with it forever. The context for all of this is that it is relatively early days for the use of layouts in TiddlyWiki, and so we might expect to have a better idea of how to improve these features once they are more widely used.

pmario:
> I introduce a new tag $:/tags/LayoutSwitcherButton for a plugin author to get these buttons. These buttons can even be used to go to a specific state of a layout (for example, 3-day view of the calendar layout or a week view of the calendar layout).

I think the proposed tag should be part of the `$:/tags/Layout/` namespace eg: `$:/tags/Layout/Switcher` ... 

IMO `xxxxxButton` ... is already too specific, because in the future there my be other elements to switch the layout. ...

I didn't have a look at the code yet. So think about this with a grain of salt.

core/ui/PageControls/reset-layout.tid
```diff
@@ -0,0 +1,15 @@
+title: $:/core/ui/Buttons/reset-layout
+tags: $:/tags/PageControls $:/tags/LayoutSwitcherButton
+caption: {{$:/core/images/reset-layout-button}} {{$:/language/Buttons/ResetLayout/Caption}}
+description: {{$:/language/Buttons/ResetLayout/Hint}}
+
+\whitespace trim
```
pmario, line 6: Since there is `\whitespace trim` the code itself should use indentation. .. We use tabs to indent code.

linonetwo: Ok, Seems I was using 2 space in some of previous code...

Jermolene:
Hi @linonetwo as I said above, I think that at this point this functionality should stay as a plugin.

The logic of resetting the layout is trivial, so there's little to be gained in terms of avoiding code duplication. The risk of doing this is that we might later come up with a better approach than `$:/tags/LayoutSwitcherButton`, and yet if we implement it now we'll be stuck with it forever. The context for all of this is that it is relatively early days for the use of layouts in TiddlyWiki, and so we might expect to have a better idea of how to improve these features once they are more widely used.

linonetwo:
Ok, I am making some plugin for that.

So i think we can merge this button only now? I should had done some cleanup.

Jermolene:
> So i think we can merge this button only now? I should had done some cleanup.

Apologies for the confusion @linonetwo. I said that I think this functionality does not belong in the core. Users already have a means to switch layouts (including via a keyboard shortcut), so I don't see the case for an additional page control button.

linonetwo:
That's OK, so how about this change, so we can encourage every layout plugin include an icon to it, and we can use `[tag[$:/tags/Layout]]` to get all layouts' icon?

This may make it possible for plugins to explore many alternative ways to switch layouts. (keyboard shortcut won't work on mobile, and the model window is not friendly on mobile)

linonetwo:
截屏2023-01-16 14 02 59

Jermolene:
Great idea @linonetwo I'll post a fix along those lines shortly.

linonetwo:
@Jermolene Do you think `$:/core/images/layout-button` is the correct icon for the default layout?

I hope there can be a dedicated icon for the default layout, so `$:/core/images/layout-button` be the icon for "layouts launcher" or "layout menu".

Currently it is confusing for user if I put `$:/core/images/layout-button` as a button that return to default layout. And `$:/core/images/layout-button` is also the button for layouts menu.

I suggest use these button for default layout (storyview layout) https://morosanuae.github.io/tw-icons/#%24%3A%2Fimages%2Fboxicons%2Fregular%2Fbx-card:%5B%5BTW%20Icons%5D%5D%20%24%3A%2Fimages%2Fboxicons%2Fregular%2Fbx-card%20%24%3A%2Fimages%2Ftemplarian-material-design%2Fcards-variant

Or reuse the storyview icon.

Jermolene:
Hi @linonetwo I agree that there is confusion from using the same button for both purposes, and I would be happy to adopt a new icon for the default page layout. My preference would be to introduce a new icon rather than re-use one of the existing ones.

PR #7201, opened 2023-01-11
Update Macro Definitions in WikiText.tid

kookma:
Ref: https://talk.tiddlywiki.org/t/requesting-macro-massage-willing-to-donate-50-to-tw/5833/17

Add explanation for single line nested macro


Jermolene:
Thanks @kookma

PR #7204, opened 2023-01-12
Fix the share plugin so it can decode tiddlers from the URI

btheado:
Reverts the part of #5999 which changed the URI decoding in the share plugin. The URI decoding for the share plugin happens in rawmarkup before `$tw.utils` exists. In addition, the "safe" version of the function is not required since a try/catch is already present in the rawmarkup.

Without this change, the share plugin is only able to generate url payloads. It cannot receive and render any url payload. When the page is loaded the following message is in the console: `Share plugin: Error decoding location hash ReferenceError: $tw is not defined`


Jermolene:
Great, thank you @btheado

PR #7207, opened 2023-01-12
Fixes share plugin stale url and prevents wikification of generated json

### LINKED ISSUE #7205: [BUG] Share plugin generates stale urls
### Describe the bug

The `$:/plugins/tiddlywiki/share/wizard` tiddler is rendered above the story one time at startup. The action widgets in the "Generate sharing link" are rendered at the same time and are never refreshed unless the sharing filter gets changed.

That means the list of tiddlers and the tiddler content to be shared (i.e. the share url) is generated only at startup. Any changes made to tiddlers matching the filter will not be included in the url when the "Generate sharing link" button is clicked.

### Expected behavior

The content of the share link should be refreshed whenever any of the content in the share filter changes.

### To Reproduce

1. Go to https://tiddlywiki.com/prerelease/share.html
2. Create and save a "New Tiddler"
3. Click the "Generate sharing link" button to copy the link to clipboard
4. Paste the clipboard contents somewhere
5. Notice the "New Tiddler" is not present in the pasted url. This is the bug.
6. Trigger a refresh by adding and then removing a space character to/from the sharing filter text box
7. Repeat steps 3 and 4.
8. Notice the "New Tiddler" is now present in the pasted url. This is a workaround.

### Screenshots

_No response_

### TiddlyWiki Configuration

https://tiddlywiki.com/prerelease/share.html - 5.2.6-prerelease

### Additional context

_No response_

btheado:
Use the `tv-action-refresh-policy` variable to ensure the share url is up-to-date when the "Generate share url" button is clicked. Also use the macrocall widgets `$output="text/raw"` to ensure the generated json doesn't get modified by wikification.

Fixes #7205

### DIFF
--- a/plugins/tiddlywiki/share/wizard.tid
+++ b/plugins/tiddlywiki/share/wizard.tid
@@ -1,7 +1,7 @@
 title: $:/plugins/tiddlywiki/share/wizard
 
 \define generate-json()
-<$macrocall $name="jsontiddlers" filter=<<share-filter>> spaces="" $output="text/vnd.tiddlywiki"/>
+<$macrocall $name="jsontiddlers" filter=<<share-filter>> spaces="" $output="text/raw"/>
 \end
 
 \define inner-share-actions()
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @btheado

PR #7210, opened 2023-01-15
Allow boot code to remain as tiddlers in tiddlywikicore.js

cdruan:
This PR retains boot tiddlers in `tiddlywikicore.js`; therefore, making it possible to easily save an external-js wiki as a standard single-file wiki. Exporting tiddlywikicore from an external-js wiki is no longer a problem.

**Important:**

The last part of a "library" tiddler's title determines how the module can be accessed from the global scope. For example, `$:/library/sjcl.js` module is loaded into `window.sjcl`.

### DIFF
--- a/core/ui/DownloadFullWiki.tid
+++ b/core/ui/DownloadFullWiki.tid
@@ -1,0 +1,17 @@
+title: $:/core/ui/DownloadFullWiki
+
+\whitespace trim
+\rules except wikilink
+
+To download the standard single-file version of your wiki:
+
+<$wikify name="site-title" text={{$:/config/SaveWikiButton/Filename}}>
+<$let publishFilter="""-[[$:/config/SaveWikiButton/Template]] -[[$:/plugins/tiddlywiki/filesystem]] -[[$:/plugins/tiddlywiki/tiddlyweb]]""">
+<$button tooltip="Download fully standalone wiki" aria-label="download full wiki" class="tc-btn-big-green">
+<$action-sendmessage $message="tm-download-file" $param="$:/core/save/all" publishFilter=<<publishFilter>> filename=<<site-title>>/>
+{{$:/core/images/download-button}}
+&#32;
+Download full wiki
+</$button>
+</$let>
+</$wikify>
--- a/core/ui/ExportTiddlyWikiCore.tid
+++ b/core/ui/ExportTiddlyWikiCore.tid
@@ -1,11 +1,6 @@
 title: $:/core/ui/ExportTiddlyWikiCore
 
 \define jsFileName() tiddlywikicore-$(version)$.js
-\define noExportMsg()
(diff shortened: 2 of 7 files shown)


core/ui/DownloadFullWiki.tid
```diff
@@ -0,0 +1,17 @@
...
+\rules except wikilink
+
+To download the standard single-file version of your wiki:
+
+<$wikify name="site-title" text={{$:/config/SaveWikiButton/Filename}}>
+<$let publishFilter="""-[[$:/config/SaveWikiButton/Template]] -[[$:/plugins/tiddlywiki/filesystem]] -[[$:/plugins/tiddlywiki/tiddlyweb]]""">
+<$button tooltip="Download fully standalone wiki" aria-label="download full wiki" class="tc-btn-big-green">
+<$action-sendmessage $message="tm-download-file" $param="$:/core/save/all" publishFilter=<<publishFilter>> filename=<<site-title>>/>
+{{$:/core/images/download-button}}
+&#32;
```
pmario, line 13: please do not user whitespace to change the layout. Use `<span class="tc-tiny-gap-left">` instead. It will allow us to modify gaps globally in the future if we need to. See: utility classes can be found here: https://github.com/pmario/TiddlyWiki5/blob/7a7d83784c91eaa3c385918918491c1416c31fba/themes/tiddlywiki/vanilla/base.tid#L3046-L3091

core/ui/ExportTiddlyWikiCore.tid
```diff
@@ -17,13 +12,7 @@ Export the TiddlyWiki core JavaScript code for running with external JavaScript:
...
-<$list
-	filter="[[$:/boot/boot.js]is[missing]]"
-	variable="ignore"
-	emptyMessage="<$action-sendmessage $message='tm-download-file' $param='$:/core/templates/tiddlywiki5.js' filename=<<jsFileName>>/>" >
-<$action-setfield $tiddler=<<qualify "$:/temp/alert">> text=<<noExportMsg>> subtitle="Export ~TiddllyWiki Core"/>
-<$action-sendmessage $message="tm-modal" $param=<<qualify "$:/temp/alert">>/>
-</$list>
+<$action-sendmessage $message='tm-download-file' $param='$:/core/templates/tiddlywiki5.js' filename=<<jsFileName>>/>
 {{$:/core/images/download-button}}
 &#32;
```
pmario, line 17: IMO should be fixed too

core/ui/DownloadFullWiki.tid
cdruan: Thanks @pmario. Done

core/ui/ExportTiddlyWikiCore.tid
cdruan: Done.

Jermolene:
Thanks @cdruan looks good.

PR #7213, opened 2023-01-16
Update is.tid

twMat:
Ref [discussion](https://talk.tiddlywiki.org/t/filter-q-is-tiddler-or-shadow-how/5888)


Jermolene:
Thanks @twMat

PR #7214, opened 2023-01-17
Improve \parsermode documentation

yaisog:
The link and tag to [[Pragma]] were corrected.
Added a note that this cannot be used in the main body of text.
Made the examples look like those in [[Pragma]] and appended the \parsermode info there.

Do we have a good definition for what "at the start of text" means? From experience, this also covers the start of macro definitions and limits the scope of the pragma to the macro, but this is nowhere documented.


Jermolene:
Thanks @yaisog

PR #7215, opened 2023-01-17
Check Content-Type of Pasted Tiddlers

GameDungeon:
Currently, if you paste-import a tiddler it assumes you are pasting in text (or a file). The issue is, this isn't always true. You could be pasting in, for example, a URI.

This PR fixes that by checking the content type and handling that data accordingly.

For testing reasons, [here is a code pen](https://codepen.io/gamedungeon/pen/ZEjXrwd) that will put a tiddler onto your clipboard.

I tested it in Chrome, Edge, and Firefox.

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -159,7 +159,7 @@ exports.importDataTransfer = function(dataTransfer,fallbackTitle,callback) {
 		if(!$tw.browser.isIE || importDataTypes[t].IECompatible) {
 			// Get the data
 			var dataType = importDataTypes[t];
-				var data = dataTransfer.getData(dataType.type);
+			var data = dataTransfer.getData(dataType.type);
 			// Import the tiddlers in the data
 			if(data !== "" && data !== null) {
 				if($tw.log.IMPORT) {
@@ -172,6 +172,34 @@ exports.importDataTransfer = function(dataTransfer,fallbackTitle,callback) {
 		}
 	}
 };
+
+exports.importPaste = function(item,fallbackTitle,callback) {
+	// Try each provided data type in turn
+	for(var t=0; t<importDataTypes.length; t++) {
+		if(item.type === importDataTypes[t].type) {
+			// Get the data
+			var dataType = importDataTypes[t];
+
+			item.getAsString(function(data){
+				if($tw.log.IMPORT) {
+					console.log("Importing data type '" + dataType.type + "', data: '" + data + "'")
+				}
+				var tiddlerFields = dataType.toTiddlerFieldsArray(data,fallbackTitle);
+				callback(tiddlerFields);
+			});
+			return;
+		}
+	}
+};
+
+exports.itemHasValidDataType = function(item) {
+	for(var t=0; t<importDataTypes.length; t++) {
+		if(!$tw.browser.isIE || importDataTypes[t].IECompatible) {
+			return true;
+		}
+	}
+	return false;
+}
 
 var importDataTypes = [
 	{type: "text/vnd.tiddler", IECompatible: false, toTiddlerFieldsArray: function(data,fallbackTitle) {
@@ -205,7 +233,13 @@ var importDataTypes = [
 		return [{title: fallbackTitle, text: data}];
 	}},
 	{type: "text/uri-list", IECompatible: false, toTiddlerFieldsArray: function(data,fallbackTitle) {
-		return [{title: fallbackTitle, text: data}];
+		// Check for tiddler data URI
+		var match = $tw.utils.decodeURIComponentSafe(data).match(/^data\:text\/vnd\.tiddler,(.*)/i);
+		if(match) {
+			return parseJSONTiddlers(match[1],fallbackTitle);
+		} else {
+			return [{title: fallbackTitle, text: data}]; // As URL string
+		}
 	}}
 ];
 
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -195,8 +195,7 @@ exports.importPaste = function(item,fallbackTitle,callback) {
 exports.itemHasValidDataType = function(item) {
 	for(var t=0; t<importDataTypes.length; t++) {
 		if(!$tw.browser.isIE || importDataTypes[t].IECompatible) {
-			if(item.type === importDataTypes[t].type)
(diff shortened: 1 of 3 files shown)


GameDungeon:
Does TiddlyWiki still support IE? I was trying to test my changes there, but I couldn't as Microsoft won't even let me download it. Also, there is a load of IE support code in the area I changed. 

My changes do work on, Chrome, Firefox, and Edge.

pmario:
```
async function copy() {
  const data = {
    text: "I love TiddlyWiki",
    title: "Hello There!",                 //<-- It does not overwrite the HelloThere tiddler but I actually needed to look twice.
  };
```

IMO that title should be more different. I immediately hesitated to import that tiddler, because it's too close to HelloThere. .. Even if I tested it with the preview, the name made me stop the import. 

May be "Hello World" .. even if it is a cliche

pmario:
About `text/vnd.tiddler` ... Is it intentional that you use a new mime type here, or just a typo?

TWs mime type is `text/vnd.tiddlywiki` ... which imo is related to the `.tid` file format. ... So it may be the right thing to give it a different type. ... not sure about that

pmario:
@GameDungeon I think, I did find a problem. To reproduce. 

- go to [here is a code pen](https://codepen.io/gamedungeon/pen/ZEjXrwd)
- click the Copy Button
- open tiddlywiki.com
- open a New Tiddler in **edit mode**
- paste the string to the text-area ---> works

Now open the [Vercle Preview](https://tiddlywiki5-git-fork-gamedungeon-copypasteimportfix-jermolene.vercel.app/)

- open a New Tiddler in edit mode
- paste the content ----> Nothing happens

Tested with FF on Widows.

GameDungeon:
Ok, I changed the title in the test :)

I didn't add "text/vnd.tiddler" As far as I can tell, it was already there.

Also, for me, it doesn't paste in either tiddlywiki.com or the preview. That is expected behavior, as if you are in an edit box it does not import, but the stuff on your clipboard isn't text, it's a tiddler URI.

pmario:
>Also, for me, it doesn't paste in either tiddlywiki.com or the preview. That is expected behavior, as if you are in an edit box it does not import, but the stuff on your clipboard isn't text, it's a tiddler URI, so it can't paste either.

You are right. I did already remove the post. It was wrong.

GameDungeon:
Oh, also I now see what you meant by "text/vnd.tiddler" I honestly got that from the dev docs, and it works, so I didn't think twice.

pmario:
>I didn't add "text/vnd.tiddler" As far as I can tell, it was already there.

Where?

GameDungeon:
When I said that, I thought you were talking about in the importdatatypes array in the dragndrop.js. In there it's the same, so I think it's just the right format for imports? Or some old code never got updated.

pmario:
The docs is at: https://tiddlywiki.com/#DropzoneWidget .. I actually didn't know that. I must have missed this info or forgot about it :( ... So it seems to be the right thing to do.

GameDungeon:
This is the one that has the correct info. Took me forever to actually find lol https://tiddlywiki.com/dev/static/TiddlyWiki%2520Drag%2520and%2520Drop%2520Interoperability.html

Jermolene:
Thanks @GameDungeon and welcome! With apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

GameDungeon:
Why is Vercel failing all the sudden? I made a small change, and it builds fine for me.

Jermolene:
> Why is Vercel failing all the sudden? I made a small change, and it builds fine for me.

I'm not sure; all the recent deployments have been failing. I'm cleaning up build-site.sh to help debug things.

GameDungeon:
Ok. Was just making sure I didn't mess up the pull request

Jermolene:
> Ok. Was just making sure I didn't mess up the pull request

The most recent commits to master did trigger to run Vercel successfully, so it may be worth pushing to this branch as well.

Jermolene:
You may need to merge the changes from master as well.

GameDungeon:
Ok, merged.

pmario:
@Jermolene ... This post contains some more thoughts and **not** necessarily a problem with the PR. I think it's browser spec?!?

I did play a bit with the new code. It seems it's not possible to copy from Edge -> FF and vice versa. ... 

According to: https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Recommended_drag_types#link

> It is recommended to always add data of the text/plain type as a fallback for applications or drop targets that do not support other types, unless there is no logical text alternative. Always add this text/plain type last, as it is the least specific and shouldn't be preferred.

`text/plain` should always be included in the clipboard, so if a browser doesn't understand or allow one datatype the text can be used as a fallback. 

So the code that creates the clipboard data should be similar to this: 

```
    e.clipboardData.setData("text/uri-list", "data:text/vnd.tiddler," + encodeURIComponent(JSON.stringify(data)));
    e.clipboardData.setData("text/plain", JSON.stringify(data));
```

IMO it would be even better if the second element would create a .tid file format eg: 

```
title: something
tags: aa bb

text
```

So it would be possible to create a TW upgrader module that is able to convert this text into a real tiddler.

GameDungeon:
I noticed that too. It seems not possible to copy between firefox and other browsers. Of course, it works individually though

GameDungeon:
Is there anything else needed for this pull or is it just currently stuck in limbo?

Jermolene:
Thanks for your patience @GameDungeon 

I did some brief tests on Chrome, Firefox and Safari on macOS. Things seem to work as expected on Chrome and Firefox, but pasting the sample does not work in Safari (pasting fragments text works as usual though).

I assume it's difficult for you to debug on Safari? I'll try to make some time to have a look through it.

GameDungeon:
Sadly, I don't own a mac, so I can't test Safari. Try checking the pasted mime-type to see if safari is just doing something weird.

Jermolene:
No worries thank you @GameDungeon I'll add this to my list and spend some time on it as soon as I can.

Jermolene:
Thanks for your patience @GameDungeon – I've updated the drag-drop interop example to handle copying to the clipboard to make this easier to test

PR #7219, opened 2023-01-19
small gap fix for the updated toc macro (alternative to tc-tiny-gap-left)

Telumire:
As discussed here : https://github.com/Jermolene/TiddlyWiki5/pull/7121

This PR fix the css for the TOC macro


Jermolene:
Thanks @Telumire

PR #7222, opened 2023-01-22
Add focusSelect attribute to <$edit-text> widget

### LINKED ISSUE #7221: [IDEA] it is proposed to add a "focusSelect" parameter to the edit-text widget
At [Talk](https://talk.tiddlywiki.org/t/focus-on-the-body-when-entered-into-edit/5955/11) it is proposed to add a **focusSelect** parameter to the edit-text widget

From Eric Shulman at Talk

-------------

Currently, the $edit-text widget supports a focus="yes" parameter, which is handled by

$:/core/modules/editor/engines/simple.js, here:

```
SimpleEngine.prototype.focus  = function() {
	if(this.domNode.focus && this.domNode.select) {
		this.domNode.focus();
		this.domNode.select();
	}
};
```

and by $:/core/modules/editor/engines/framed.js, here:

```
FramedEngine.prototype.focus  = function() {
	if(this.domNode.focus && this.domNode.select) {
		this.domNode.focus();
		this.domNode.select();
	}
};
```

Note that these functions not only set the input focus on the domNode, but they also select the domNode contents.

In order to support the proposed new behavior, this handling would need to be re-written. I suggest adding a new select="no" parameter to the $edit-text text widget so that the current hard-coded behavior can be bypassed if desired.

Note that, for backward compatibility, the default when this new parameter is omitted must be to still select the domNode contents as it presently does.

Jermolene:
This PR adds a new "focusSelect" attribute to the `<$edit-text>` widget to control how text is selected when the "focus" attribute is enabled.

Fixes #7221


saqimtiaz:
This is a very welcome change and would allow me to deprecate the [extension of the $edit-text widget that is a part of the Streams plugin](https://github.com/saqimtiaz/streams/blob/main/plugins/streams/core-tweaks/simple-engine-tweak.js).

We should also consider adding support for selecting a specific range in the text. While I do not have an immediate need for it, I can imagine scenarios in which it would be useful and implementing it now would allow for a more cohesive implementation.

Potential approaches for specifying the range to select could be based on start and index indexes, or prefix and suffix text strings that identify the selection.

Jermolene:
> We should also consider adding support for selecting a specific range in the text. While I do not have an immediate need for it, I can imagine scenarios in which it would be useful and implementing it now would allow for a more cohesive implementation.

That makes sense. Perhaps we should split the "focusSelect" attribute into two separate attributes: "focusSelectStart" and "focusSelectEnd". Both attributes would take numeric values, with two subtleties:

* If both attributes are omitted, "focusSelectStart" defaults to 0 and "focusSelectEnd" defaults to the end of the string
* If one attribute is provided but the other is omitted, then that one value is applied to both properties. So, just setting "focusSelectStart=0" would make a zero width selection at the start of the string
* Negative numbers would be taken from the end of the string: the value -1 would be equivalent to the string length, -2 to the string length minus 1, etc. So, setting "focusSelectEnd=-1" would make a zero width selection at the end of the string

The handling of negative indexes seems clumsy and confusing but I am struggling to see an alternative.

saqimtiaz:
> The handling of negative indexes seems clumsy and confusing but I am struggling to see an alternative.

I cannot think of a better alternative at present either. This will however be an advanced feature not relevant for most users so that helps to some extent.

pmario:
I think `selectStart` and `selectEnd` is enough, if we add to the docs that those parameters are only active if "focus" attribute is set to "yes"

>The handling of negative indexes seems clumsy and confusing but I am struggling to see an alternative.

-1 may be an option for programmers, but IMO not for our users, even advanced users. I think the only problem is with `focusEnd` since we don't know the length of the string.  I think as soon as `focusEnd` exists it should be active. So `focusEnd` with emptyString or "0" should be the end of the string.

`focusEnd="-1"` should be length-1 already. I think that would be consistent with js-string-functions eg: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice

So the start of the selection can also start with 0 since it should be in front of the first character. That should also make sense for our users.

yaisog:
I like Mario's zero-based proposal. How about we rename the parameters `focusFromStart` and `focusFromEnd`, or just `fromStart` and `fromEnd`? This would make their meaning immediately clear.
Setting only one of the parameters will make a zero-length selection at the specified location, either counted from the beginning or end of the contents. To select everything from/to some position, set the other parameter to 0. Nice and flexible.

I can imagine a nice example with some buttons that select different parts of a string in an $edit-text. For that it would be necessary to amend the tm-focus message to also include these parameters. Which should be done anyways, for consistency.

Jermolene:
> I like Mario's zero-based proposal. How about we rename the parameters `focusFromStart` and `focusFromEnd`, or just `fromStart` and `fromEnd`? This would make their meaning immediately clear. Setting only one of the parameters will make a zero-length selection at the specified location, either counted from the beginning or end of the contents. To select everything from/to some position, set the other parameter to 0. Nice and flexible.

That's a great idea. I would favour `focusFromStart` and `focusFromEnd` for maximum clarity.

> I can imagine a nice example with some buttons that select different parts of a string in an $edit-text. For that it would be necessary to amend the tm-focus message to also include these parameters. Which should be done anyways, for consistency.

That would be helpful.

Would anyone like to prepare a PR for this?

Jermolene:
> Would anyone like to prepare a PR for this?

Ooops! A bit of contextual drift on my part, I missed that this is a thread on one of my own PRs rather than a discussion thread... I'll make the updates soon

yaisog:
That's probably not right:
image
We're not quite at 5.2.26 ... yet.

yaisog:
Also, would you mind if I created an example for the docs? Although I wouldn't know how to squeeze it into this PR...

pmario:
Since the PR is still open, you only need to commit to this branch and push it again. It will be added automatically here

yaisog:
Hmm, GitHub created a new PR #7286. I was hoping that wouldn't happen. But this is the way, I guess.

pmario:
>Hmm, GitHub created a new PR https://github.com/Jermolene/TiddlyWiki5/pull/7286. I was hoping that wouldn't happen. But this is the way, I guess.

uups. Sorry. I thought is was one of your PRs but it's Jeremy's ... So you cannot push to this one. :/

saqimtiaz:
We should merge this to allow for testing before the release of v5.2.6

Jermolene:
Thanks @saqimtiaz

PR #7230, opened 2023-01-24
Fixed issue where $genesis didn't pass isBlock

flibbles:
The $genesis widget wasn't passing the parseTreeNode.isBlock attribute along to the generated widget. Not all widgets make use of it, but some do. I think $transclude is one of them.

I used $reveal to do the test because it's one of the only widgets that makes a visible difference whether it's block or inline without needing extra sorcery. Hope that's all right. (I used it a bunch in tests for TW-Uglify.)

### DIFF
--- a/core/modules/widgets/genesis.js
+++ b/core/modules/widgets/genesis.js
@@ -46,6 +46,7 @@ GenesisWidget.prototype.execute = function() {
 	this.genesisRemappable = this.getAttribute("$remappable","yes") === "yes";
 	this.genesisNames = this.getAttribute("$names","");
 	this.genesisValues = this.getAttribute("$values","");
+	this.genesisIsBlock = this.getAttribute("$mode",this.parseTreeNode.isBlock && "block") === "block";
 	// Do not create a child widget if the $type attribute is missing or blank
 	if(!this.genesisType) {
 		this.makeChildWidgets(this.parseTreeNode.children);
@@ -60,7 +61,7 @@ GenesisWidget.prototype.execute = function() {
 		tag: nodeTag,
 		attributes: {},
 		orderedAttributes: [],
-		isBlock: this.parseTreeNode.isBlock,
+		isBlock: this.genesisIsBlock,
 		children: this.parseTreeNode.children || [],
 		isNotRemappable: !this.genesisRemappable
 	}];
--- a/core/modules/widgets/genesis.js
+++ b/core/modules/widgets/genesis.js
@@ -60,6 +60,7 @@ GenesisWidget.prototype.execute = function() {
 		tag: nodeTag,
 		attributes: {},
 		orderedAttributes: [],
+		isBlock: this.parseTreeNode.isBlock,
 		children: this.parseTreeNode.children || [],
 		isNotRemappable: !this.genesisRemappable
 	}];
--- a/editions/test/tiddlers/tests/data/genesis-widget/Block.tid
+++ b/editions/test/tiddlers/tests/data/genesis-widget/Block.tid
@@ -1,0 +1,18 @@
+title: Genesis/Block
+description: genesis widget distinguishes between block and inline
+type: text/vnd.tiddlywiki-multiple
+tags: [[$:/tags/wiki-test-spec]]
+
+title: Output
+
+\whitespace trim
+<$genesis $type="$reveal" type=nomatch>
+
+Block
+</$genesis>
+
+<$genesis $type=$reveal type=nomatch>Inline</$genesis>
++
+title: ExpectedResult
+
+<div class=" tc-reveal"><p>Block</p></div><p><span class=" tc-reveal">Inline</span></p>
--- a/editions/test/tiddlers/tests/data/genesis-widget/Block.tid
+++ b/editions/test/tiddlers/tests/data/genesis-widget/Block.tid
@@ -11,8 +11,20 @@ title: Output
 Block
 </$genesis>
 
+<$genesis $type="$reveal" type=nomatch $mode=block>
+
+Block forced block
+</$genesis>
+
+<$genesis $type="$reveal" type=nomatch $mode=inline>
+
+Block forced inline
+</$genesis>
+
 <$genesis $type=$reveal type=nomatch>Inline</$genesis>
+<$genesis $type=$reveal type=nomatch $mode=block>Inline forced block</$genesis>
+<$genesis $type=$reveal type=nomatch $mode=inline>Inline forced inline</$genesis>
 +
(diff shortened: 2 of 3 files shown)


Jermolene:
Hi @flibbles I wonder if in such cases we should also provide an attribute that allows the setting to be overridden (`$isBlock` in this case), with the default behaviour being to pass the setting through from the parse tree node. I think we may find other raw properties  in the parse tree that we also want to cover.

flibbles:
Actually now that you mention it, it would be nice to be able to reengage block behavior in some cases while retaining surrounding structure. I've made some pretty ugly wikitext in the past to be able to mimic that sort of thing. I'd be happy to modify this PR and add the documentation. Give me a day or two.

Jermolene:
Thanks @flibbles.

flibbles:
@Jermolene, now that I'm working on it, wouldn't it make more sense to have a `$mode` attribute instead with accepted attributes of `block` and `inline`? That seems more consistent with widgets like `$transclude` and `$wikify`.

flibbles:
Here is an implementation of it. I used `$mode` attribute instead of an `$isBlock`, since there is no precedent of isBlock in Tiddlywiki that I can find. If you'd like me to change it, I'd be happy to. It would take me about 10 minutes.

Jermolene:
Thanks for your patience @flibbles. The `$mode` attribute does make sense but I wanted to review the other raw attributes used by the parsers:

* The dash parser and the entity parser put the entity string in a parse tree node property called `entity`
* The html parser uses a parse tree node property called `isSelfClosing`, but as far as I can tell it is unused in the rest of the core
* The macrocall parser uses a parse tree node property called `params` for the parameters and one called `name` for the name of the macro
* The macrodef parser uses a parse tree node property called `params` for the parameters and one called `isMacroDefinition` to mark the variable definition as a macro
* The text parser uses a parse tree node property called `text` for the text to be rendered (the associated widget also accepts the text as an attribute)

I think it makes sense to address all of these omissions at once. I'd propose that alongside `$mode` we also add:

* `$entity` for entities
* `$params` for the parameters for macro calls and definitions, both in slightly different JSON formats
* `$isMacroDefinition` for macro calls

I'm not sure whether the ability to dynamically compose macro definitions is really going to be useful, but it seems like it might be worth giving people a chance to find out.

flibbles:
Are you sure you want to introduce irreversible support for all of these?

`Entity` could easily be accomplished with `<$genesis $type=$entity entity="&nbsp;" />` instead of the proposed `<$genesis $type=$entity $entity="&nbsp;" />`. Honestly, it seems kinda legacy that there's an entity value on parseTreeNodes at all, and maybe the dash and entity parser should originally have just set a `parseTreeNode.attribute.entity`, rather than have a dedicated `parseTreeNode.entity` value.

For the rest, (`$params` and `$isMacroDefinition`). Is it a good idea to burden ourselves with supporting them forever onward if no one has yet come up with a plausible use-case for when they'd be used? I sure can't think of one.

Maybe it's better to wait until someone asks for these features before adding them, and this PR be kept to the scope of `$mode`?

flibbles:
I can sort of see some value to `isSelfClosing`, even if I can't think of a use-case off the top of my head where you'd specify it rather than just... *closing* the widget. What would that look like?

```
<$genesis $type=$transclude ... isSelfClosing={{{ conditional filter }}}>

Content that gets conditionally excluded????
</$genesis>
```

I can definitely put this one in, but I'd just like to ask ourselves first if it's feature creeping the genesis widget.

Jermolene:
Hi @flibbles I'd [forgotten](https://github.com/Jermolene/TiddlyWiki5/commit/2adf09129dddd4df231d18b4198fb2a9663fdf5c#diff-c90d07b1d8f7c937d8ae55251c086106add29a3916acc877e9ae111e98c18dd3) that the entity widget can now accept an attribute, so I agree we don't need that one.

I don't think `isSelfClosing` does anything useful, but I might be wrong.

> For the rest, (`$params` and `$isMacroDefinition`). Is it a good idea to burden ourselves with supporting them forever onward if no one has yet come up with a plausible use-case for when they'd be used? I sure can't think of one.

Being able to do dynamic function definitions is useful in other functional languages and so I would expect it to be useful here.

> Honestly, it seems kinda legacy that there's an entity value on parseTreeNodes at all

That's correct, but we still have the backwards compatibility concerns.

> I can definitely put this one in, but I'd just like to ask ourselves first if it's feature creeping the genesis widget.

Not really. The purpose of the genesis widget is to be able to dynamically create any valid widget. The current inability to specify all the properties of the parse tree node that are used in the core is a shortcoming.

flibbles:
If we really want to allow for dynamic function definitions, doesn't it make **a lot** more sense to put the `params` and `isMacroDefinition` attribute on `$set` directly? It's the only widget that actually uses those, so it seems silly to force users to resort to creating `$set` through `$genesis $type=$set`. This way...

1. Dynamic functions don't have an extra layer of complexity by having to work through `$genesis`.
2. `$genesis` doesn't have to support attributes which only make sense for one particular widget.
3. This PR can stick to the scope of `$mode`, because all this would make more sense as a separate issue.

flibbles:
I would like to add that if/when/where `isMacroDefinition` is implemented, either on `$set` or `$genesis`, it shouldn't be called that. `isMacroDefinition` may be the name of the internal parseTreeNode member, and must remain for legacy, by introducing this as an attribute, isMacroDefinition no longer necessarily means the widget is a macro.

A better name would be something like `substitute=yes` or `placeholders=yes` or something like that. All it does is enable substitution of of `$(variablesLikeThis)$`.

pmario:
> ... `$(variablesLikeThis)$`

I thought we should avoid and deprecate "variables like the above one"

flibbles:
Placeholders in general? Sure, but you've got to take that up with Jermolene. I'm just trying to fix an inline/block bug. :)

pmario:
>Placeholders in general? Sure, but you've got to take that up with Jermolene. I'm just trying to fix an inline/block bug. :)

Yea, I also think this PR should be merged and the other stuff can be added later. If we wait for a consensus on everything nothing will get merged in the near future.

Jermolene:
Hi @flibbles 

>  doesn't it make a lot more sense to put the params and isMacroDefinition attribute on $set directly?

Yes, that would  be a more important change to make first.

> I would like to add that if/when/where isMacroDefinition is implemented, either on $set or $genesis, it shouldn't be called that

Indeed, the code in #6666 adds additional flags for procedures, functions etc. so probably would make sense to do that after #6666 is merged

Issue #7223, opened 2023-01-23
[BUG] But in browse widget


AnthonyMuscio:
### Describe the bug

Normally a widget needs to be closed as a rule, either with `</$widdgetname>` or the short form `/>`.

The browse widget, when used to set the deserialiser such as here does not work.
 
`<$browse multiple deserializer="text/plain"/>`

The only form that works seems to be the incorrect form

`<$browse multiple deserializer="text/plain">`

### Expected behaviour

For the following form to work, and perhaps not the unclosed form.

`<$browse multiple deserializer="text/plain"/>`

However since the unclosed form may have being the "only working form" for some time perhaps in still needs to be supported.

### To Reproduce

Use the above examples and attempt to import a JSON tiddler file

The Correct form will not even display a Button.

### Screenshots

N/A

### TiddlyWiki Configuration

- Version v5.2.3 v5.2.
- Saving mechanism - N/A on tiddlywiki.com
- Plugins installed - N/A on tiddlywiki.com

### Desktop (please complete the following information):

- OS: Windows 11 latest
- Browser chrome and FireFox
- Version Latest General Availability

### Smartphone (please complete the following information):

Not tested 
Assumed the same

### Additional context

Matter arising when looking at improving the documentation and handling of JSON in tiddlywiki https://talk.tiddlywiki.org/t/conversion-tools-between-the-various-tiddler-and-json-formats/5990?u=tw_tones

ericshulman:
Using Chrome, I entered `<$browse multiple deserializer="text/plain"/>` into a "New Tiddler" on https://TiddlyWiki.com.  When rendered, it showed a "choose file" button as expected.  Clicking the button opened a system "select a file/folder" dialog.  Selecting a file and pressing "open" button displayed the "$:/Import" tiddler with the selected file listed.  Pressing "import" created the correct tiddler.

It seems to be working properly.  Can you provide more detailed information to help reproduce the problem you observed?

AnthonyMuscio:
That's very odd. It is no longer occurring for me either, I will see if I can replicate it and see what caused it (tomorrow) And close if I cant reproduce it.
* It occurred in the two browsers

Ahh - just found it coincidence in my test

```
<$browse deserializer="text/plain">

<$browse multiple deserializer="text/plain"/>
```

The second only fails if in the above order, if reversed they both work. Just an unfriendly result of the malformed/unclosed widget.

PR #7231, opened 2023-01-24
Update BrowseWidget.tid

AnthonyMuscio:
Made sure all references to use of the `<$browse>` widget were changed to `<$browse/>`

Which possibly caused this problem, now closed https://github.com/Jermolene/TiddlyWiki5/issues/7223

If not closed other widgets may work

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/BrowseWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/BrowseWidget.tid
@@ -11,7 +11,7 @@ The browse widget displays an HTML file browser button that allows the user to c
 
 ! Content and Attributes
 
-The content of the `<$browse>` widget is ignored.
+The content of the `<$browse/>` widget is ignored.
 
 |!Attribute |!Description |
 |multiple |Set to "multiple" to select multiple file upload |
@@ -26,10 +26,10 @@ On iPhone/iPad choosing the multiple option will remove the ability to take phot
 ''e.g.''
 
 ```
-<$browse>
+<$browse/>
 ```
 
 renders as:
 
-<$browse>
+<$browse/>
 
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/widgets/BrowseWidget.tid
```diff
@@ -11,7 +11,7 @@ The browse widget displays an HTML file browser button that allows the user to c
 
 ! Content and Attributes
 
-The content of the `<$browse>` widget is ignored.
+The content of the `<$browse/>` widget is ignored.
```
Jermolene, line 14: A lot of inconsistency has crept in, but for textual references to widgets within their own definition should use the `<<.wid>>` macro:

```
The content of the <<.wid BrowseWidget>> widget is ignored
```

(References to widgets that should be linked use the `<<.wlink>>` macro).

Jermolene (commented):
Thanks @AnthonyMuscio just a small point.

editions/tw5.com/tiddlers/widgets/BrowseWidget.tid
AnthonyMuscio: Noted: 
How can I see if you made this change and accepted it?
If not how do I go about doing this?
* Cancel and resubmit?

btheado: Are you asking how to make the requested changes? It should work if you go to https://github.com/AnthonyMuscio/TiddlyWiki5/blob/patch-24/editions/tw5.com/tiddlers/widgets/BrowseWidget.tid and click the edit button and make the changes.

I haven't edited files via the github website before, but I expect making the change in that way will cause a new commit to the same branch (patch-24). Since that branch is attached to this pull request, it will appear here.

Jermolene:
Thanks for your patience @AnthonyMuscio

PR #7232, opened 2023-01-24
Make save wiki button be accessible to users without colour vision

### LINKED ISSUE #7145: [BUG] Save wiki button status not accessible for colour blind users
The save wiki button indicates "dirty" status by colouring the icon red. The transition between red and black is not accessible for some colour blind users.

It is proposed to:

* make the icon additionally change shape to reflect the dirty status
* ensure that the tooltip for the save wiki button also reflects the dirty status

The challenge is that internally the dirty status is implemented as the addition of the class "tc-dirty" to the body element. This means that only CSS can be used to implement the status differences.

We cannot use a state tiddler to indicate the dirty status because changing that state tiddler would itself change the dirty status of the wiki.

Jermolene:
This PR introduces a new icon for the save button that changes shape when when the wiki is dirty to make it distinguishable by colour blind users by replacing the tick with a blob. It is implemented as a variant of the `$:/core/images/save-button` image called `$:/core/images/save-button-dynamic`:

Screenshot 2023-01-24 at 11 10 46

The appearance remains the same when the wiki is not dirty:

Screenshot 2023-01-24 at 11 12 17

I also explored a more complex variant that incorporates an exclamation mark but decided that it was confusing at small sizes:

Screenshot 2023-01-24 at 11 12 27

Fixes #7145


pmario:
IMO it's not possible to change the core save button in that way. It will break every plugin, that uses that button in a different context. 

eg: If my script manager stuff is used with the PR preview it looks like this

image

But it should look like this, because the save button in that context doesn's use a colour change, because of the dirty status. So it shouldn't change form too

image

pmario:
IMO there should be a new button with the desired behaviour. may be $:/core/images/save-dirty-button

Jermolene:
> IMO there should be a new button with the desired behaviour. may be $:/core/images/save-dirty-button

That's my conclusion too. I'll push a fox later.

Jermolene:
I've pushed an update renaming the dynamic version of the save button to "$:/core/images/save-button-dynamic".

pmario:
I think an image like image would be known by most users. 

I recognised the image used atm as a "record" button on a camera app.

Jermolene:
Hi @pmario the constraints are:

* we cannot change the basic "checkmark in a circle" icon used for saving because it is referenced in so much documentation and screenshots
* the "dirty state" icon cannot be totally different from the ordinary one; we want users to perceive the icon as having transformed somehow, and not being swapped for something completely different

So I don't see a way that we can adopt the standard download icon for our save wiki button. In any case, it seems particularly tied to the download saver, and doesn't seem so appropriate for other saver mechanisms.

I agree about the resemblence to a record button, but I don't think that's a problem. It's impossible to design a simple icon that doesn't resemble other simple icons. What we're aiming for is a set of icons that are harmonious and easily distinguishable.

pmario:
>I agree about the resemblence to a record button, but I don't think that's a problem. It's impossible to design a simple icon that doesn't resemble other simple icons. What we're aiming for is a set of icons that are harmonious and easily distinguishable.

I'm OK with the icon ... just thinking out loud.

Jermolene:
I'm open to suggestions for better alternatives for the graphic but I'm going to go ahead and merge this so that we get some more feedback.

mateuszwilczek:
Here are some icon ideas I have come up with:

unsaved-icon-draft

B1 is `$:/core/images/cancel-button` for comparison, to make sure that the dirty icon is not too similar to it, as it does the exact opposite thing – discards changes.

I think A5 is the best. It gives a clear message – altered tick icon, not just an altered circle which is not specific to the save button. It is readable in low resolution, A6, B5, B6 are worse in this regard. The style of the "crossed out" icons A3–B4 is often used for settings or toggles that one can switch in both ways, and they may add confusion being a little similar to B1.

Jermolene:
Thank you @mateuszwilczek those ideas are very helpful. I share your view that A5 works best. The only thing I'm unsure about is that it comprises three elements, which perhaps makes it a little harder to parse visually.

I think all of the examples illustrate very well the particular shortcoming of the current A2 icon – the white inner ring ends up being visually prominent. Offsetting the smaller circle works well to break that effect.

pmario:
I think we should try it as it is implemented by the PR. We'll see what the feedback will be. 

@Jermolene .. you wrote, that you may want to release 5.2.6 in Mach .. So we may tell the community, that they should start to test tiddlywiki.com/prelrelease

ericshulman:
Note:

For plugin code that is updated to use the new `$:/core/images/save-button-dynamic` image defined in v5.2.6, you can replace:

`{{$:/core/images/save-button}}`
with
`<$transclude tiddler="$:/core/images/save-button-dynamic">{{$:/core/images/save-button}}</$transclude>`

This will preserve backward-compatibility so the updated plugin will fallback to using the previous image when installed in v5.2.5 or earlier.

Jermolene:
Based on a [discussion at talk.tiddlywiki.org](https://talk.tiddlywiki.org/t/now-is-the-time-to-help-with-testing-the-tiddlywiki-v5-2-6-prerelease/6405/60), I've made an [update](https://github.com/Jermolene/TiddlyWiki5/commit/95f987544cf13c2acc9e1664db1f42eba5a9cc55) to the "dirty" shape as follows:

Screenshot 2023-03-14 at 14 44 00 Screenshot 2023-03-14 at 14 44 09

pmario:
I'm OK with that change. I'm using the "now outdated" version for quite some time now. It works OK. So the diamond should work too. We will only need to update the banner - competitions / voting :)

PR #7233, opened 2023-01-24
add depth first destroy method 

buggyj:
This is another take on #6699.

A widget would add a method like this:

```
TextNodeWidget.prototype.destroy = function() {
	var text = this.getAttribute("text",this.parseTreeNode.text || "");
	text = text.replace(/\r/mg,"");
	console.log("destroy " + text );
};
```

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -617,7 +617,7 @@ Widget.prototype.waitDestroy = false;
 /*
 Remove any DOM nodes created by this widget or its children
 */
-Widget.prototype.removeChildDomNodes = function() {
+Widget.prototype.removeChildDomNodes = function(noDestroy) {
 	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
 	if(this.domNodes.length > 0) {
 		$tw.utils.each(this.domNodes,function(domNode) {
@@ -627,10 +627,10 @@ Widget.prototype.removeChildDomNodes = function() {
 	} else {
 		// Otherwise, ask the child widgets to delete their DOM nodes
 		$tw.utils.each(this.children,function(childWidget) {
-			childWidget.removeChildDomNodes();
+			childWidget.removeChildDomNodes(true); 
 		});
 	}
-	if (!this.waitDestroy) this.desendDestroy();
+	if (!noDestroy && !this.waitDestroy) this.desendDestroy();
 };
 
 /*
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -596,6 +596,24 @@ Widget.prototype.findFirstDomNode = function() {
 	return null;
 };
 
+
+
+/*
+Depth first destroy
+*/
+Widget.prototype.desendDestroy = function() {
+	this.waitDestroy = true; //for blocking repeat calls to this method from removeChildDomNodes
+	$tw.utils.each(this.children,function(childWidget) {
+		childWidget.desendDestroy();
+	});
+	
+	if (this.destroy) {
+		this.destroy();
+	}
+	this.waitDestroy = false;
+};
+
+Widget.prototype.waitDestroy = false;
 /*
 Remove any DOM nodes created by this widget or its children
 */
@@ -612,6 +630,7 @@ Widget.prototype.removeChildDomNodes = function() {
 			childWidget.removeChildDomNodes();
 		});
 	}
+	if (!this.waitDestroy) this.desendDestroy();
 };
 
 /*


buggyj:
needs a bit more work

pmario:
It would be nice to have a little ZIP file with a custom widget that does call the destroy function as described in the OP. So we could import and test it with the vercel preview.

buggyj:
So basically after removeChildDomNodes() has detached the descending dom nodes, all destroy() methods will be called in a depth first manner.  There is no change to the behavior of the widget as destroy() is not called directly. The widget developer may call removeChildDomNodes()  from there destroy() method and so there is a guard variable (waitDestroy) to stop  the destroy process being started more than once unintentionally .

core/modules/widgets/widget.js
```diff
@@ -596,10 +596,28 @@ Widget.prototype.findFirstDomNode = function() {
...
+	}
+	this.waitDestroy = false;
+};
+
+Widget.prototype.waitDestroy = false;
 /*
 Remove any DOM nodes created by this widget or its children
 */
-Widget.prototype.removeChildDomNodes = function() {
+Widget.prototype.removeChildDomNodes = function(noDestroy) {
```
pmario, line 620: what would be the opposite of "noDestroy" .... I think I do have a problem with the "inverted" logic in combination with the inverted variable names.

buggyj: All widgets call removeChildDomNodes() with no parameter, which is falsy, and I cannot pass in 'false'

pmario: Yea, I think I know how it works, but it needs a bit of mind-twisting to get it. So if there would be "easier" names it may help

Jermolene: > All widgets call removeChildDomNodes() with no parameter, which is falsy, and I cannot pass in 'false'

Perhaps we could check explicitly for undefined, and not just treat it as falsy?

buggyj: we could pass in an options object, like this:

```

Widget.prototype.removeChildDomNodes = function(options) {
	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
	if(this.domNodes.length > 0) {
		$tw.utils.each(this.domNodes,function(domNode) {
			domNode.parentNode.removeChild(domNode);
		});
		this.domNodes = [];
	} else {
		// Otherwise, ask the child widgets to delete their DOM nodes
		$tw.utils.each(this.children,function(childWidget) {
			childWidget.removeChildDomNodes({destroy:true}); 
		});
	}
	if ((options && options.destroy && !this.waitDestroy) this.desendDestroy();
};
```

core/modules/widgets/widget.js
```diff
@@ -615,10 +615,11 @@ Widget.prototype.desendDestroy = function() {
 
 Widget.prototype.waitDestroy = false;
 /*
-Remove any DOM nodes created by this widget or its children
+Remove any DOM nodes created by this widget or its children, and start destroy procedure
 */
-Widget.prototype.removeChildDomNodes = function(noDestroy) {
-	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
+Widget.prototype.removeChildDomNodes = function(destroy) {
+     var destroy = (destroy !== undefined ? destroy : true);
```
pmario, line 621: You should be able to use `destroy = destroy || true;` .. see there is no var

pmario: would it be better to use an `options = options || {}` with an object here. It would make it possible to add parameters, without the need to change the function signature again.

buggyj: with options it would be:

```
Widget.prototype.removeChildDomNodes = function(options) {
	var options = options || {destroy:true};
	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
	if(this.domNodes.length > 0) {
		$tw.utils.each(this.domNodes,function(domNode) {
			domNode.parentNode.removeChild(domNode);
		});
		this.domNodes = [];
	} else {
		// Otherwise, ask the child widgets to delete their DOM nodes
		$tw.utils.each(this.children,function(childWidget) {
			childWidget.removeChildDomNodes({destroy:false}); 
		});
	}
	if ((options.destroy  ===  true) && !this.waitDestroy) this.desendDestroy();
}
```

I don't think any of the 3 versions are better or worse.

core/modules/widgets/widget.js
```diff
@@ -596,11 +596,30 @@ Widget.prototype.findFirstDomNode = function() {
...
+
+
 /*
-Remove any DOM nodes created by this widget or its children
+Depth first destroy
 */
-Widget.prototype.removeChildDomNodes = function() {
-	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
+Widget.prototype.desendDestroy = function() {
+	this.waitDestroy = true; //for blocking repeat calls to this method from removeChildDomNodes
```
linonetwo, line 605: No need for this lock, because this method won't get run until it complete its running. JS is run-to-end.

buggyj: some developer may create a destroy() like this:

aWidget.prototype.destroy = function() {
	this.removeChildDomNodes();
        //etc....
};

ittayd:
The original use case was for event listeners cleanup. So maybe instead of changing the destruction flow, just add a notification flow that the dom was removed? A widget will still have reference to its own dom node and can do any cleanup it wants.

linonetwo:
@buggyj are you interested in make a new PR about this? I think with some test case and perf, it will be better this time. Or I can take time for one.

buggyj:
@linonetwo maybe @saqimtiaz  has something to contribute

https://github.com/Jermolene/TiddlyWiki5/pull/7468#issuecomment-1665659922

pmario:
@buggyj In the light of:  **Adds a widget destroy method** #9097 -- Is this PR still relavant?

If not please close.

PR #7236, opened 2023-01-25
fix 7229 add innerClass to checkbox widget

### LINKED ISSUE #7229: [IDEA] The checkbox widget should get a second class for the "inner text"
## Proposal

It is proposed to add a second class eg: `inner-class` or `text-class` attribute to the checkbox-widget. This class should be assigned to the "native" SPAN element that is created by the widget. 

## Reasoning

As shown in the HTML code. If we assign a class to the following wikitext, the class is assigned to the "outer" label element. That's OK. 

BUT since we have whitespace trim the inner text is covered in a SPAN by the widget without any spacing. 

```
\whitespace trim

<$checkbox tag="done" class="tc-tiny-gap-right">Is it done?</$checkbox>
```

```
<p><label class="tc-checkbox tc-tiny-gap-right"><input type="checkbox"><span>Is it done?</span></label></p>
```

----------

The fix is like: 

```
\whitespace trim

<$checkbox tag="done" class="tc-tiny-gap-right"><span class="tc-tiny-gap-right">Is it done?<span></$checkbox>
```

BUT it creates 2 SPANS for the TEXT element. 

```
<p><label class="tc-checkbox tc-tiny-gap-right"><input type="checkbox"><span><span class="tc-tiny-gap-right">Is it done?<span>&lt;/$checkbox&gt;</span></span></span></label></p>
```

Screenshot **as it is now**

image

pmario:
fix #7229 add innerClass to checkbox widget

- Add 2 new tests
- add docs for new "innerClass" parameter
- add class attribute to inner SPAN in the DOM **only** if a class attribute is set

### DIFF
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -52,6 +52,9 @@ CheckboxWidget.prototype.render = function(parent,nextSibling) {
 	}
 	this.labelDomNode.appendChild(this.inputDomNode);
 	this.spanDomNode = this.document.createElement("span");
+	if(this.checkboxInnerClass !== "") {	// do not add an empty class attribute to the DOM
+		this.spanDomNode.setAttribute("class","" + this.checkboxInnerClass);
+	}
 	this.labelDomNode.appendChild(this.spanDomNode);
 	// Add a click event handler
 	$tw.utils.addEventListeners(this.inputDomNode,[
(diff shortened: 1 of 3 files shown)


pmario:
converting this PR in to a draft because new thoughts and experiments are needed.

pmario:
Don't see a good way to fix this. So close.

PR #7241, opened 2023-01-28
Add a th-modal-refreshed hook

yaisog:
Since the th-page-refreshed hook fires before modals are updated, it is not possible to react directly to changes in the modals' content.
This PR adds a th-modal-refreshed hook that triggers when modals are refreshed (but not when they're initially created).

Relates to [https://talk.tiddlywiki.org/t/does-th-page-refreshed-fire-before-modals-are-updated/6026](https://talk.tiddlywiki.org/t/does-th-page-refreshed-fire-before-modals-are-updated/6026)

### DIFF
--- a/core/modules/utils/dom/modal.js
+++ b/core/modules/utils/dom/modal.js
@@ -180,6 +180,7 @@ Modal.prototype.display = function(title,options) {
 		headerWidgetNode.refresh(changes,modalHeader,null);
 		bodyWidgetNode.refresh(changes,modalBody,null);
 		footerWidgetNode.refresh(changes,modalFooterButtons,null);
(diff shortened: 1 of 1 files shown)


yaisog:
Does this need to be documented somewhere? Hooks are not discussed on TW5.com, but only in the developer documentation. How are these edited?

pmario:
All the hooks are documented at: https://tiddlywiki.com/dev/ .... search for "th-"

Jermolene:
Thanks @yaisog the developer docs are in the edition `editions/dev`

yaisog:
My need for this PR has been made superfluous by the Dynannotate plugin. Therefore I withdrew the PR.

Issue #519, opened 2014-03-31
tiddler info: Advanced tab - basically all tabs should contain propper sentences. 


pmario:
```
Shadow Status
The tiddler $:/core/ui/TiddlerInfo/Advanced/ShadowInfo is a shadow tiddler
It is defined in the plugin $:/core
It is overridden by an ordinary tiddler
```

imo should be

```
Shadow Status
The tiddler $:/core/ui/TiddlerInfo/Advanced/ShadowInfo is a shadow tiddler.
It is defined in the plugin $:/core.
It is overridden by an ordinary tiddler.
```

PR #7243, opened 2023-01-29
Fix importTitle and autoOpenOnImport in the tm-import-tiddler message

### LINKED ISSUE #7234: [BUG] tm-import-tiddlers handler can't receive parameters from action-sendmessage widget
### Describe the bug

#4741 extended the tm-import-tiddlers handler to allow customizations via `importTitle` and `autoOpenOnImport`. However, the handler looks for these values at the top-level of the `event` object rather than the standard way of using paramObject subkeys.

The `$action-sendmessage` widget can only send values via the paramObject field. That means the customized `importTitle` and `autoOpenOnImport` values sent by `action-sendmessage` will be ignored by the tm-import-tiddlers handler.

The `$dropzone` widget code is calling dispatchEvent directly and it is passing the two fields as `event` top-level keys. Therefore setting these values via `$dropzone` works fine (I didn't test, just assuming based on reading the code).

I think the tm-import-tiddlers handler in navigator.js can be changed to look for the fields in paramObject instead of at the top level. A matching change in the `$dropzone` widget would be required. It would need to use paramObject in its call to dispatchEvent

### Expected behavior

It should be possible to customize `importTitle` and `autoOpenOnImport` using the `$action-sendmessage` widget.

### To Reproduce

1. Go to https://tiddlywiki.com
2. Create a new tiddler with the following content
```
<$let  json=<<jsontiddlers filter:"[tag[Core Macros]first[2]]">>>

<pre><$text text=<<json>>/></pre>

<$button
  actions="""
    <$action-sendmessage
      $message="tm-import-tiddlers"
      $param=<<json>>
      importTitle="import-test"
      autoOpenOnImport="no"
    />
  """
>
Import tiddlers to the "import-test" tiddler
</$button>

`$:/Import`: <pre><$text text=<<jsontiddlers filter:"$:/Import">>/></pre>

`import-test`: <pre><$text text=<<jsontiddlers filter:"import-test">>/></pre>
</$let>
```
4. Click on the "Import tiddlers..." button
5. Notice the tiddlers are imported into `$:/Import` instead of `import-test` and the `$:/Import` tiddler is auto-opened contrary to the custom attributes.

### Screenshots

_No response_

### TiddlyWiki Configuration

tiddlywiki.com version 5.2.5

### Additional context

_No response_

btheado:
Since the action-sendmessage widget can only pass parameters using `event.paramObject`, change the tm-import-tiddlers message handler to look in both `event.paramObject` and `event` for the options. Fixes #7234.

Added tests which fail before the fix and pass after the fix. Also added a test which passes both before and after the fix.

### DIFF
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -540,7 +540,7 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Save the $:/Import tiddler
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
-	// Update the story and history details.
+	// Update the story and history details
 	var autoOpenOnImport = event.autoOpenOnImport || paramObject.autoOpenOnImport || this.getVariable("tv-auto-open-on-import");
 	if(autoOpenOnImport !== "no") {
 		var storyList = this.getStoryList(),
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -499,7 +499,8 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Get the tiddlers
 	var tiddlers = $tw.utils.parseJSONSafe(event.param,[]);
 	// Get the current $:/Import tiddler
-	var importTitle = event.importTitle ? event.importTitle : IMPORT_TITLE,
+	var paramObject = event.paramObject || event,
+		importTitle = paramObject.importTitle ? paramObject.importTitle : IMPORT_TITLE,
 		importTiddler = this.wiki.getTiddler(importTitle),
 		importData = this.wiki.getTiddlerData(importTitle,{}),
 		newFields = new Object({
@@ -540,7 +541,7 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
 	// Update the story and history details
-	var autoOpenOnImport = event.autoOpenOnImport ? event.autoOpenOnImport : this.getVariable("tv-auto-open-on-import");  
+	var autoOpenOnImport = paramObject.autoOpenOnImport ? paramObject.autoOpenOnImport : this.getVariable("tv-auto-open-on-import");  
 	if(autoOpenOnImport !== "no") {
 		var storyList = this.getStoryList(),
 			history = [];
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -540,7 +540,7 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Save the $:/Import tiddler
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
-	// Update the story and history details
+	// Update the story and history details. Note: paramObject is either event or event.paramObject
 	var autoOpenOnImport = paramObject.autoOpenOnImport ? paramObject.autoOpenOnImport : this.getVariable("tv-auto-open-on-import");  
 	if(autoOpenOnImport !== "no") {
 		var storyList = this.getStoryList(),
--- a/core/modules/widgets/navigator.js
+++ b/core/modules/widgets/navigator.js
@@ -499,8 +499,8 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Get the tiddlers
 	var tiddlers = $tw.utils.parseJSONSafe(event.param,[]);
 	// Get the current $:/Import tiddler
-	var paramObject = event.paramObject || event,
-		importTitle = paramObject.importTitle ? paramObject.importTitle : IMPORT_TITLE,
+	var paramObject = event.paramObject || {},
+		importTitle = event.importTitle || paramObject.importTitle || IMPORT_TITLE,
 		importTiddler = this.wiki.getTiddler(importTitle),
 		importData = this.wiki.getTiddlerData(importTitle,{}),
 		newFields = new Object({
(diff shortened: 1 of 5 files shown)


core/modules/widgets/navigator.js
```diff
@@ -499,7 +499,8 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Get the tiddlers
 	var tiddlers = $tw.utils.parseJSONSafe(event.param,[]);
 	// Get the current $:/Import tiddler
-	var importTitle = event.importTitle ? event.importTitle : IMPORT_TITLE,
+	var paramObject = event.paramObject || event,
```
btheado, line 502: With this change, the code looks both at the toplevel of event or in event.paramObject. I could have changed it to only look in event.paramObject. With that approach, dropzone.js would also need to be changed to pass parameters in paramObject.

pmario:
I did test the code from https://github.com/Jermolene/TiddlyWiki5/issues/7234 with the vercel preview wiki -- It does not do anything. ... Does the code need to be changed?

core/modules/widgets/navigator.js
```diff
@@ -540,7 +541,7 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
 	// Update the story and history details
```
pmario, line 543: May be there should be 1 line of comment here that the `paramObject` could be `event` from above. I think than it's OK the event changed the name.

btheado:
> I did test the code from #7234 with the vercel preview wiki -- It does not do anything. ... Does the code need to be changed?

@pmario thanks for testing! I updated the example. With the updated example it should be easier to see the behavior difference between tiddlywiki.com and the vercel preview.

core/modules/widgets/navigator.js
btheado: Thanks. I added a comment

pmario:
Did more testing. IMO setting autoOpenOnImport would have done the trick too. `autoOpenOnImport="yes"` ;)

core/modules/widgets/navigator.js
```diff
@@ -539,8 +540,8 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Save the $:/Import tiddler
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
-	// Update the story and history details
-	var autoOpenOnImport = event.autoOpenOnImport ? event.autoOpenOnImport : this.getVariable("tv-auto-open-on-import");  
+	// Update the story and history details. Note: paramObject is either event or event.paramObject
```
Jermolene, line 543: I don't think the logic is quite right here. The value of paramObject only falls back to the event object if the paramObject is entirely missing. We need to do that check at the next level down: something like `event.importTitle || event.paramObject.importTitle || IMPORT_TITLE".

Jermolene (commented):
Thanks @btheado

core/modules/widgets/navigator.js
btheado: Thanks, @Jermolene. This would be to guard against things going wrong when dispatchEvent is called directly, right?

Here is how dropzone.js is calling it:

```js
this.dispatchEvent({
  type: "tm-import-tiddlers",
  param: JSON.stringify(tiddlerFieldsArray),
  autoOpenOnImport: this.autoOpenOnImport,
  importTitle: this.importTitle
});
```

I think if instead someone did the following (last line is different), then your suggested change would be required for it to work.

```js
this.dispatchEvent({
  type: "tm-import-tiddlers",
  param: JSON.stringify(tiddlerFieldsArray),
  autoOpenOnImport: this.autoOpenOnImport,
  paramObject: {importTitle: this.importTitle}
});
```

Jermolene: Thanks @btheado, that's correct.

We may well want in the future to introduce the ability to dispatch events with custom properties on the root of the event object because it would be required for full backwards compatibility as we move more functionality from JavaScript into wikitext (eg the navigator widget).

btheado: Thanks, @Jermolene. I changed the code. Now the parameters can be split between the top level event object and paramObject and it should still work.

btheado:
In addition to the automated test cases I added, I also manually tested the following:

1. Dropping json tiddlers on the default dropzone
2. Dropping json tiddlers on a manually constructed dropzone (`<$dropzone importTitle="mytestimport" autoOpenOnImport=no>`)
3. Trigger an import by copying/pasting text
4. Import a file via the browse widget

It all looked good to me.

I did briefly explore trying to add a test action widget which would allow mixing some javascript calls in with the wiki-based tests hoping it would allow adding wiki-based tests for things like the dropzone widget and the browse widget. I thought the action widget could make calls to `this.dispatchEvent` to trigger "drop" and "paste" events for dropzone.

Something like this:

```js
[...snip...]
/*
Invoke the action associated with this widget
*/
EvalJavascriptWidget.prototype.invokeAction = function(triggeringWidget,event) {

	eval(this.javascript);
	return true; // Action was invoked
};

exports["testaction-evaljavascript"] = EvalJavascriptWidget;
```

For example, I was able to use it to test the "mixed" parameters on `event` and `paramObject`
```
<$testaction-evaljavascript
  javascript='this.dispatchEvent({
      "type": "tm-import-tiddlers",
      "param": "[{\"title\": \"Elephants\"}, {\"title\": \"Eagles\"}]",
      "paramObject": {"autoOpenOnImport": "no"},
      "importTitle": "MyCustomTitle"
});'/>
```

However, for the dropzone widget, I would need to construct a proper DragEvent with a proper DataTransfer field. I wasn't sure if that was a wise thing to attempt or not. Or whether the `testaction-evaljavascript` is a good idea or not. So I left it out and didn't pursue it further.

pmario:
> Or whether the testaction-evaljavascript is a good idea or not.

IMO calling "eval()" when users are able to define the content will never be a good idea.

core/modules/widgets/navigator.js
```diff
@@ -539,8 +540,8 @@ NavigatorWidget.prototype.handleImportTiddlersEvent = function(event) {
 	// Save the $:/Import tiddler
 	newFields.text = JSON.stringify(importData,null,$tw.config.preferences.jsonSpaces);
 	this.wiki.addTiddler(new $tw.Tiddler(importTiddler,newFields));
-	// Update the story and history details
-	var autoOpenOnImport = event.autoOpenOnImport ? event.autoOpenOnImport : this.getVariable("tv-auto-open-on-import");  
+	// Update the story and history details.
```
pmario, line 543: most of the time full-stops are missing intentionally

See: https://github.com/Jermolene/TiddlyWiki5/issues/519
and: https://en.wikipedia.org/wiki/Full_stop#In_text_messages

btheado: Thanks. That was a mistake. I fixed it.

btheado:
> IMO calling "eval()" when users are able to define the content will never be a good idea.

This was not a core widget. Just one I created at `editions/test/tiddlers/tests/data/modules/widgets`. What "users" would be able to define content there? I guess the edition is published at https://tiddlywiki.com/test. Is that what you are worried about?

Jermolene:
Thanks for your patience @btheado, great stuff.

PR #7252, opened 2023-02-03
Zoomin: Avoid using broken tc-storyview-zoomin-tiddler class

### LINKED ISSUE #7247: [BUG] Tag change removes tc-storyview-zoomin-tiddler class from tiddler frame
### Describe the bug

When tags are added via `$fieldmangler`, the `tc-storyview-zoomin-tiddler` class which is present for all tiddlers in the zoomin storyview is removed from the current tiddler.
Switching back and forth between tiddlers adds the class again.
Removing the class this way can lead to unwanted changes in the UI if there is CSS defined on `tc-storyview-zoomin-tiddler`.

### Expected behavior

Only the class relating to the removed tag should be removed, not the class `tc-storyview-zoomin-tiddler`.

### To Reproduce

Go to tiddlywiki.com and create a tiddler with this content:
```
<$fieldmangler>
<$button message="tm-add-tag" param="HelloThere" class="tc-remove-tag-button">Add</$button>
<$button message="tm-remove-tag" param="HelloThere" class="tc-remove-tag-button">Remove</$button>
</$fieldmangler>
```
Click one of the buttons and observe the classes on the div.tc-tiddler-frame.

### Screenshots

_No response_

### TiddlyWiki Configuration

TW 5.2.5

### Additional context

_No response_

Jermolene:
The Zoomin storyview uses a brittle technique for setting the tc-storyview-zoomin-tiddler class on the tiddler frames in the story view. This is required for the tiddler widths to be set correctly. This PR uses the newer storyview classes on the page container instead.

Fixes #7247

### DIFF
--- a/core/modules/storyviews/zoomin.js
+++ b/core/modules/storyviews/zoomin.js
@@ -35,7 +35,6 @@ var ZoominListView = function(listWidget) {
 		} else {
 			self.currentTiddlerDomNode = domNode;
 		}
-		$tw.utils.addClass(domNode,"tc-storyview-zoomin-tiddler");
 	});
 };
 
@@ -52,7 +51,6 @@ ZoominListView.prototype.navigateTo = function(historyInfo) {
 		return;
 	}
 	// Make the new tiddler be position absolute and visible so that we can measure it
-	$tw.utils.addClass(targetElement,"tc-storyview-zoomin-tiddler");
 	$tw.utils.setStyle(targetElement,[
 		{display: "block"},
 		{transformOrigin: "0 0"},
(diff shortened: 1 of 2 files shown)


yaisog:
I see you didn't change the ViewTemplate. Is that so that custom ones are not affected and receive this fix without having to be changed?

Jermolene:
> I see you didn't change the ViewTemplate. Is that so that custom ones are not affected and receive this fix without having to be changed?

Because we already have a class on the page template that reflects the current storyview, and so it would be redundant to introduce another class along the same lines at the tiddler frame level.

But I am concerned about the backwards compatibility impact – it might be safer to keep the code in zoomin.js just in case plugins are using that class.

yaisog:
> it might be safer to keep the code in zoomin.js just in case plugins are using that class

So, deprecate it first, tell people to switch to the page template classes, and remove it later?

PR #7260, opened 2023-02-10
Add additional searchModes to Dynannotate

### LINKED ISSUE #7256: [IDEA] Make Dynannotate searchMode parameter consistent with search[] filter suffix
### Dynannotate
The search modes supported by the `searchMode` attribute are:
- **normal** - a literal string of plain text to match
- **regexp** - a JavaScript-style regular expression (without the quoting backslashes and flags)
- **whitespace** - a literal string to match while normalising runs of whitespace. This allows `a. b` to match `a.  b`

### `search` filter
Search mode - the first to be set of the following flags determines the type of search that is performed:
- **literal**: considers the search string to be a literal string, and requires an exact match
- **whitespace**: considers the search string to be a literal string, but will consider all runs of whitespace to be equivalent to a single space. Thus `A B` matches `A   B`
- **regexp**: treats the search string as a regular expression. Note that the regexp option obviates the need for the older [regexp](https://tiddlywiki.com/prerelease/#regexp%20Operator) operator.
- **words**: (the default) treats the search string as a list of tokens separated by whitespace, and matches if all of the tokens appear in the string (regardless of ordering and whether there is other text in between)
- **some**: treats the search string as a list of tokens separated by whitespace, and matches if at least one of the tokens appear in the string

Dynannotate's `normal` should be renamed to `literal`. I don't think it makes sense for Dynannotate to differentiate between `words` and `some`. Since `words` has been there longer, I would use that. That should also be the default mode in Dynannotate. Its function would be to highlight all words in the search string separately, not just when there is a literal match. That functionality is currently missing.

yaisog:
This PR adds three additional `searchMode`s to Dynannotate, to make it more consistent with the suffixes of the `search` filter operator:

- `literal` is introduced as an alias of `normal`
- `some` and `words` operate identically and will treat the search string as a list of tokens separated by whitespace, and match all tokens (regardless of ordering and whether there is other text in between)

Closes #7256. 

This PR also adds or changes some plugin examples.


yaisog:
Here's some code I used for testing, if anyone wants to try it out in the preview without a lot of typing:

```
<$dynannotate search={{!!key}} searchMode={{!!mode}} searchCaseSensitive="no" searchMinLength=3>

one two three four five six seven eight nine ten

one two three four five six seven eight nine ten

one two three four five six seven eight nine ten
</$dynannotate>

<$edit-text field="key" />
<$select field="mode">
<option value="">none</option>
<option value="literal">literal</option>
<option value="regexp">regexp</option>
<option value="whitespace">whitespace</option>
<option value="some">some</option>
<option value="words">words</option>
</$select>
```

Jermolene:
Thanks @yaisog – I'll look at the code properly shortly, but I wonder if that example you gave might be useful to add to the demos?

yaisog:
I'll have a look tomorrow. I'm a big fan of demos and example code, after all.

pmario:
I did test it with the example code provided in the OP with all mode options. It works as advertised in the OP

I think `casesensitive` and `anchored` are missing for feature parity with the [search-operator](https://tiddlywiki.com/#search%20Operator)


@Jermolene 
There seems to be something wrong with the `$:/plugins/tiddlywiki/dynannotate/examples/viewtemplate` example. It seems to create a cascade config tiddler with the wrong tag: 

image

IMO it should be ` tags="$:/tags/ViewTemplateBodyFilter"` as shown in the code below. ... atm it is ` tags="$:/tags/ViewTemplateTitleFilter"`,

```
<$button>
<$action-setfield $tiddler="$:/config/ViewTemplateTitleFilters/dynannotate" tags="$:/tags/ViewTemplateBodyFilter" text="[[$:/plugins/tiddlywiki/dynannotate/examples/viewtemplate/text]]" list-before="$:/config/ViewTemplateBodyFilters/default"/>
Click here to use the custom view template
</$button>
```

But I'm not 100% sure how that example should actually work.

yaisog:
@Jermolene: Here ya go. I combined all the _Simple_ search examples into a single example with interactive attributes, as inspired by the code above.
The corresponding code view is generated on-the-fly according to the user's selection. Because of the `$codeblock`, I had to use `$wikify` for that, which I never do. So, if you have some way of doing that without it, let me know.
The default search term should produce different results with the various searchModes / case sensitivities. Users can enter their own, though.
Also, it demonstrates the different predefined searchClasses (at least the ones I know of). So, maybe you don't have to write up something extra about them (except maybe in the readme).

pmario:
If you are in your TiddlyWiki5 directory you can use the following command to start the core dev-server edition with the plugin for testing. .. That's how I did test it

```
cd /to/your/TiddlyWiki5
tiddlywiki.cmd +plugins/tiddlywiki/dynannotate .\editions\tw5.com-server\ --listen
```

pmario:
>Here ya go. I combined all the Simple search examples into a single example with interactive attributes, as inspired by the code above. ...

I do like it. The docs and functionality now looks good to me

yaisog:
> If you are in your TiddlyWiki5 directory you can use the following command to start the core dev-server edition with the plugin for testing. .. That's how I did test it

I'm a bit embarrassed to admit that I'm working with the single-file wikis and vercel deployments. That _does_ make it harder to push commits, though. That's always a lot of copy/paste. I imagine a node.js wiki combined with GitHub desktop can streamline that workflow considerably...

plugins/tiddlywiki/dynannotate/examples/searching-in-tiddlywiki.tid
```diff
@@ -0,0 +1,19 @@
+title: $:/plugins/tiddlywiki/dynannotate/searching-in-tiddlywiki
```
Jermolene, line 1: I think this tiddler is just here as test data. It is a bit confusing that that cannot be determined by reading it. It would be clearer if the text of it were titled `$:/plugins/tiddlywiki/dynannotate/example-tiddler` tiddler and the content explicitly said "This tiddler is test data for use in the Dynannotate demos".

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
```diff
@@ -1,28 +1,72 @@
...
 
 <$button>
-<$action-deletetiddler $tiddler="$:/config/ViewTemplateTitleFilters/dynannotate"/>
+<$action-deletetiddler $tiddler="$:/config/ViewTemplateBodyFilters/dynannotate"/>
 Click here to revert to the core view template
 </$button>
 
-<$codeblock code={{$:/plugins/tiddlywiki/dynannotate/examples/viewtemplate/text}}/>
\ No newline at end of file
+<div style="border: 1px solid black; border-radius: 20px; padding: 1em;">
```
Jermolene, line 38: Could we use a class here?

Jermolene:
Hi @yaisog I hadn't realised how useless the Vercel preview is in these cases where one is working on a plugin that isn't included in the tw5.com wiki. I'd recommend adding Dynnanotate to the tiddlywiki.info file for the tw5.com edition so that we can easily try things out.

plugins/tiddlywiki/dynannotate/examples/simple.tid
```diff
@@ -10,75 +10,68 @@ caption: Simple
...
-</$dynannotate>
-""">>
 
 <div class="tc-dynannotation-example-info">
 
-!! Normalised whitespace searching
+!! Highlights with `search`
 
-We use the `mode` attribute set to `whitespace` to search for a string with whitespace normalised (ie runs of whitespace are collapsed to a single space for matching purposes):
+We use the `search` attribute to specify a search string for highlighting.<br>
```
Jermolene, line 50: What is the purpose of using `<br>` here? Why can't we rely on ordinary paragraph formatting?

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
```diff
@@ -1,28 +1,72 @@
...
+
+{{$:/plugins/tiddlywiki/dynannotate/searching-in-tiddlywiki}}
```
Jermolene, line 40: Why is the demo text included here? The view template demo doesn't work here so it's not obvious what it's doing here.

Jermolene (commented):
Thanks @yaisog

It took me a long time to work through this PR because there are so many different, somewhat unrelated things going on here that make it very hard to track:

* Functional improvements to the code
* Docs wording changes
* Fixing the view template demo
* Adding the new interactive demo
* etc

To keep things moving quicker it's much easier if future PRs can be more focussed.

yaisog:
> I'd recommend adding Dynnanotate to the tiddlywiki.info file for the tw5.com edition so that we can easily try things out.

Is this something you'd like _me_ to do? I wouldn't know where to start...

plugins/tiddlywiki/dynannotate/docs/readme.tid
```diff
@@ -95,10 +96,10 @@ Note that using the `annotate-tiddler` field to associate an annotation with the
 
 !! Selection Trackers
 
-The following configuration tiddlers can be used to control whether the selection trackers are enabled when the following configuration tiddlers are set to ''yes'' (the default).
+The selection trackers are enabled when the following configuration tiddlers are set to ''yes''.
 
-* $:/config/Dynannotate/SelectionTracker/Enable for the main selection tracker
-* $:/config/Dynannotate/LegacySelectionTracker/Enable for the legacy selection tracker
+* `$:/config/Dynannotate/SelectionTracker/Enable` for the main selection tracker
```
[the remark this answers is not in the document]
yaisog: Sounds reasonable.

plugins/tiddlywiki/dynannotate/examples/simple.tid
```diff
@@ -10,75 +10,68 @@ caption: Simple
...
-
-!! Simple annotation
-
-We use the `target*` attributes to specify a target string for the annotation and optionally a prefix and suffix for disambiguating multiple occurances.
-
-</div>
-
-<<show-example """
+\define search-example-code()
+\rules only filteredtranscludeinline
```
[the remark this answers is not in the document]
yaisog: Wait, what? Did you just suggest using macro substitutions? 😱 
Just kidding. But do we really need to have _more_ of those around?
I'd like to just add a short explanation to the code instead, something like
```
<!-- The rules pragma is used here to evaluate only the inline filters depending on the chosen configuration, but leave everything else untouched, to be displayed within the example code block.
This way the displayed code updates to reflect the choices the user has made in the example UI. -->
```

plugins/tiddlywiki/dynannotate/examples/searching-in-tiddlywiki.tid
yaisog: I'm open to use something completely different here, some text unrelated to TW maybe. Like in the _simple_ examples. I did want to transclude a "real" tiddler, since it's a demonstration of the view template. Initially I _did_ just transclude "Searching in TiddlyWiki", but then I realized that this tiddler won't exist for most people who install the plugin in their wikis, so I made a copy of it...
I'm not really married to the idea, so if you have a better example (which preferrably uses a bunch of features of WikiText)...

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
yaisog: I was afraid you'd say that. There's just no reward for being lazy... 😉 
Done and done.

plugins/tiddlywiki/dynannotate/examples/simple.tid
yaisog: I thought the space between the lines was too large with paragraphs. Changed it, though.

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
yaisog: Why do you think the view template doesn't work there? Once activated, it works on all open tiddlers and all of their body content.

However, you are right that it makes no sense to have the example text transcluded there a second time (additionally to the second view template example). The first example can be tested on the whole wiki and needs no extra playground.
You did have the source of $:/plugins/tiddlywiki/dynannotate/examples/viewtemplate/text transcluded in that place in a codeblock in its entirety. I felt this was too much content and too abstract, so I used a shorter text that people might relate to better. But in reality there is no example text needed at all...

plugins/tiddlywiki/dynannotate/examples/simple.tid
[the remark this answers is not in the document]
Jermolene: I'd be happy with the addition of a comment. I don't think we should compromise the readability of the example output to make the wikitext easier to read, but otherwise anything you can do is appreciated.

plugins/tiddlywiki/dynannotate/examples/searching-in-tiddlywiki.tid
Jermolene: Maybe we can transclude the control panel basic tab?

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
Jermolene: No problem, let's leave things like this.

Jermolene:
> I hope I did that right... Seemed to have worked, anyway.

Looks good!

> And here I thought reusing this PR for the view template stuff was a clever idea... 😞 Should I split this up for the cleanup or just power through it as is?

Let's power through!

plugins/tiddlywiki/dynannotate/examples/searching-in-tiddlywiki.tid
yaisog: I used the tiddler fields tab from _Advanced_, because it's not interactive und thus not potentially confusing. Wouldn't someone to mess up their wiki because they think they're just playing with an example...

plugins/tiddlywiki/dynannotate/examples/viewtemplate.tid
yaisog: Too late, that thing was gone already. I do think it's better this way, less clutter.

yaisog:
Hi @Jermolene, just an aside: I just installed GitHub Desktop which somehow automagically integrated with my Sublime Text, and it was just eye-opening! All my work before was completely browser-based and clumsy. Is there a tutorial somewhere for aspiring TW contributors how to contribute _efficiently_? I wish I had seen something like this a while ago. I've read contributing.md, but there was nothing.
I can imagine that using the node.js version of TW has even more benefits, but I have no experience with that.

Jermolene:
> Hi @Jermolene, just an aside: I just installed GitHub Desktop which somehow automagically integrated with my Sublime Text, and it was just eye-opening! All my work before was completely browser-based and clumsy. Is there a tutorial somewhere for aspiring TW contributors how to contribute _efficiently_? I wish I had seen something like this a while ago. I've read contributing.md, but there was nothing.

@pmario has made some tutorial videos that may be helpful, they are linked on tiddlywiki.com.

For a long time, I used GitHub Desktop along with Sublime Text. I switched editors to VSCode a couple of years ago because of the momentum around VSCode. VSCode actually has a perfectly serviceable Git client built-in but I still use GitHub Desktop because I find it clearer and easier to use. I am not in any sense a power user of these tools; I pretty much use the defaults, and don't customise the environment or configuration.

I suspect that nowadays the easiest way to work for many people is to use GitHub Codespaces or CodeSandbox.io, both of which offer a free Node.js environment in the cloud, with a browser-based development environment. That way there is no need for any local installation, making it possible to work from a tablet.

> I can imagine that using the node.js version of TW has even more benefits, but I have no experience with that.

Conceptually, the Node.js version is pretty simple: it allows one to edit TW as a bunch of separate files, and then run a single command to pack it all up into an index.html file. It's the missing link between the file layout you see on github.com/Jermolene/TiddlyWiki5 and the index.html file on tiddlywiki.com.

Jermolene:
(I'm travelling today but will review your changes later).

yaisog:
> @pmario has made some tutorial videos that may be helpful, they are linked on tiddlywiki.com.

In the video: 123 open issues, 89 open PRs. 😆

pmario:
>I can imagine that using the node.js version of TW has even more benefits, but I have no experience with that.

I did create a video series lately, that talks about TW and plugins. It contains several videos, which are very basic. Eg: How to install Node.js LTS version  ... and so on.

https://www.youtube.com/playlist?list=PLuiC_HFhI4OxO7BAiKMp7psz9g0PBQBd-

----

There is a How to work with a TiddlyWiki Dev Environment, but it's already 5 years old and uses Brackets as an editor. I did like it quite a bit, but VSCode took the dev-community by storm. .. So I would have to do an update there. 

https://www.youtube.com/playlist?list=PLuiC_HFhI4OxtThoQf92ws875R9zgNGSd

yaisog:
@Jermolene: Do you want to merge this for 5.2.6?

Jermolene:
Hi @yaisog thanks for all your work on this. We should leave this until after v5.2.6; I've added it to the list at #7345.

Jermolene:
Thanks for your patience @yaisog

Issue #7262, opened 2023-02-11
[IDEA] Should tm-close-tiddler (optionally) navigate to the previous tiddler?


yaisog:
After `tm-close-tiddler`, the tiddler referenced in the `current-tiddler` field in $:/HistoryList is no longer in the story. That seems a bit contradictory. An easy solution would be to navigate back to the previous tiddler in tv-history-list -- but only if the closed tiddler was the `current-tiddler`.

That would change the default behavior of the story, though, and I can already see some people running for their pitchforks. So maybe make that an option, via a parameter to the message?

yaisog:
In WikiText, it might look a bit like this:
```
<$list filter="[all[current]match{$:/HistoryList!!current-tiddler}]">
	<$let previousItem={{{ [{$:/HistoryList}jsonindexes[]] :map[{$:/HistoryList}jsonget<currentTiddler>,[title]] :intersection[list<tv-story-list>] +[!match<currentTiddler>reverse[]first[]] :else[list<tv-story-list>before<currentTiddler>] :else[list<tv-story-list>after<currentTiddler>]" }}}>
		<!-- HistoryList aktualisieren, damit die Meta-Informationen und menubar wieder stimmen -->
		<$action-navigate $to=<<previousItem>> />
	</$let>
</$list>
```
where I wasn't really consistent with the source of the history list. Also, `current-tiddler` is not cleared when the story is empty, because somewhere in the back of my mind, something's yelling "Never mess with $:/HistoryList!"

yaisog:
Same goes for `tm-delete-tiddler`...?

pmario:
IMO similar to: When closing, navigate to previous tiddler #2776 .. which has been closed lately because of [a comment form Saq at Talk](https://talk.tiddlywiki.org/t/a-peculiar-pr-request/5234/20)

>Implementing anything like this is intricately tied with the work of opening up the core to be expressed more so in wikitext and thus hackable. In particular we need to provide affordances to replace the current [JavaScript only hooks 1](https://tiddlywiki.com/dev/#HookMechanism) with equivalent wikitext based affordances. This is a non-trivial amount of work. Parameterized transclusions are the first step in this direction.
>
Trying to address the expressed needs independently of implementing wikitext affordances for hooks would likely lead to two similar but conflicting mechanisms for customizing core behaviour.

yaisog:
@pmario: Not only similar, but pretty much congruent. I did search before opening the issue, but only through open PRs...

[...]

PR #7268, opened 2023-02-15
Modals in WikiText

### LINKED ISSUE #7267: [IDEA] Re-implement modals in WikiText
It was suggested by one @Jermolene that it might be prudent to re-engineer TW modals, preferrably in WikiText.

I have tried to create a baseline version (posted as a PR shortly) that implements all the documented functionality (except for `rootwindow`). I'll write about the limitations and differences to the current modals in the PR, to make that a bit self-contained.

Feel free to play around with the preview and offer improvements.

Also, since the message handling is written in WikiText, too, I have not found a great way to transfer the hashmap of parameters to variables inside the modal. Currently, I'm using `$setmultiplefields` to put them into temporary tiddlers, and then `$setmultiplevariables` to assign the indexes to modal variables. The latter handling of `tm-modal` in JS will probably take care of this, I hope.

yaisog:
**Update 2023-02-21**:
I have pushed a major update. The changes are explained in [this comment](https://github.com/Jermolene/TiddlyWiki5/pull/7268#issuecomment-1438360819).

---

This is a first implementation of modals in WikiText, as discussed in #7267 

- Most of the documented functionality has been implemented (except `rootwindow`).
- There are no CSS animations. Everything is static for now.
- The `tm-modal` and `tm-close-tiddler` messages are also handled in WikiText. For that I did insert the extensible `$messagecatcher` from #7262 into the PageTemplate. This will not catch `tm-modal` calls from global keyboard shortcuts, though. Messages originating from inside the modal are intercepted by a separate `$messagecatcher`.
- I have not found a great way to transfer the hashmap of parameters to variables inside the modal. Currently, I'm using `$setmultiplefields` to put them into temporary tiddlers, and then `$setmultiplevariables` to assign the indexes to modal variables.
- Modals can be stacked, even in loops. So the example in the "Modals" tiddler works exactly as on tiddlywiki.com.
- The state tiddler that keeps the list of opened modals is currently hardcoded to "$:/state/modal", but could be made flexible. But then, where would it be set?
- The state tiddler can be used to check for opened modals. Deleting the state tiddler closes all modals. There was a request for this somewhere (not only by me).
- The `mask-closable` field only checks for "yes", not for "true".
- The bit of CSS in the modal will eventually be stuffed into the global stylesheet.

I'm not sure where to go from here, except for collecting ideas and requests. Currently, it is (mostly) backwards compatible. I'd probably could use some help with the JS implementation of handling the pertinent messages, and also the CSS. I'm not sure if the animations can be implemented without JS. `pop` storyview doesn't work right out of the box.

Closes #7267


pmario:
It seems there are no examples in the PR?

yaisog:
For now there are only the examples from tiddlywiki.com, i.e. the "Modals" doc tiddler.

I will add some more when I get the chance...

pmario:
> the "Modals" doc tiddler.

Those modals use the new mechanism? You wrote there are no animations?

core/ui/PageTemplate/MessageHandler/tm-modal.tid
```diff
@@ -0,0 +1,10 @@
+title: $:/yaisog/ui/PageTemplate/MessageHandler/tm-modal
```
pmario, line 1: :) ... IMO needs to be `/core/ui`

yaisog: This is temporary, since I'd assume the $messagecatcher will not be permanently used to redirect tm-modal calls. I didn't want to imply that this should be added to the core. The title has no meaning for functionality.

yaisog:
> Those modals use the new mechanism? You wrote there are no animations?

Sorry, my mistake. I didn't account for all the tiddlers being shadows in the preview deployment. The filter of the $genesis widget in the PageTemplate only checked for tiddlers, and thus did not find anything overrides.

This time I checked that it works. 😄

core/ui/PageTemplate/modals.tid
```diff
@@ -0,0 +1,87 @@
...
+	<!-- Show all modal tiddlers in the order they are listed in the state tiddler -->
+	<$list filter="[enlist:raw{$:/state/modal}]" variable="currentModal" counter="modalCounter" >
+	<!-- We need special handling for tm-close-tiddler and tm-modal from inside the modal (the latter to be handled by the core eventually) -->
+	<$messagecatcher $tm-close-tiddler=<<tm-close-handler>> $tm-modal=<<tm-modal-handler>> >
+	<!-- Retrieve the paramObject hashmap in case a modal was called with it -->
+	<$setmultiplevariables $names="[[$:/state/modal/paramObject/]addsuffix<modalCounter>indexes[]]" $values="[[$:/state/modal/paramObject/]addsuffix<modalCounter>indexes[]getindex[]]">
+	<!-- Implements mask-closable (only checks for "yes") -->
+	<$eventcatcher $click={{{ [<currentModal>get[mask-closable]match[yes]then<close-current-modal>] }}} selector=".tc-modal-backdrop" matchSelector=".tc-modal-backdrop" tag="div">
+	<!-- Display modal content -->
+	<div class="tc-modal-backdrop">
```
pmario, line 58: This structure is wrong. .. `tc-modal-backdrop` must not cover the `tc-modal`. As the name says it is a "backdrop". The modals need to be on top of it. 

That means the `matchSelector` of the eventcatcher needs to be changed too. I'll have a closer look. 

If `tc-modal-backdrop` covers the whole thing and it's style is set to `opacity: .7;` as it has to be, it will also make the modal semi-transparent.

pmario:
I'm not overly happy with the modal state-handling. I think there should be 1 state-tiddler per modal. Especially `<$action-listops` should not be needed at all. nowhere

pmario:
With the screenshot I did remove the backdrop to be able to see the state/modal tiddler content for debugging

I saw that there are 2 modals atm. 

grafik

core/ui/PageTemplate/modals.tid
yaisog: You're right. `tc-modal-backdrop` and `tc-modal` must be siblings. I'll change that. I think `matchSelector` might not be needed in this case at all...

Like I said, I didn't touch the styles yet. The opacity is set via JavaScript in modal.js, to fade out the background in an animation.

pmario:
>Like I said, I didn't touch the styles yet. The opacity is set via JavaScript in modal.js, to fade out the background in an animation.

Yea. I see. I did temporarily remove the backdrop to see the state tiddler content for testing. I just thought, the whole modal handling should just be a different story. So modals can be standard tiddlers. I do hate the special handling for modals. 

The concept feels completely wrong. .. Just a gut-feeling

yaisog:
> I'm not overly happy with the modal state-handling. I think there should be 1 state-tiddler per modal. Especially `<$action-listops` should not be needed at all. nowhere
> 
> Also `targetCount` may be redundant if there is 1 state-tiddler per modal.

I thought about this a bit before starting, but couldn't decide if these state tiddlers should be identified by a special field, tag or title prefix. I eventually decided against it, since I couldn't come up with a great way of tracking their order, like a list can. BUT, the individual state tiddlers would be able to replace the extra tiddlers currently needed to set the variables (but later maybe not anymore?). I found out about the need for those only after I had implemented the single-tiddler list version...

I think that the cleanest way, without adding extra tags or fields to the core, would be increasing title prefixes, `$:/state/modal/1`, `$:/state/modal/2`, etc. and take care of ordering in that way. What do you think?

yaisog:
> The concept feels completely wrong. .. Just a gut-feeling

Yeah, that came up in a discussion I had with Jeremy, too. Nobody seems to like that. He invited me to work on an alternative, which I'm doing here.

I really appreciate all your feedback so far.

yaisog:
> I saw that there are 2 modals atm.

I couldn't recreate that on the preview site, after I cleared the background property on `tc-modal-backdrop` and made the `tc-modal` semi-transparent, it looked like this:
image

The DOM also doesn't contain anything extra...

pmario:
>I think that the cleanest way, without adding extra tags or fields to the core, would be increasing title prefixes, $:/state/modal/1, $:/state/modal/2, etc. and take care of ordering in that way. What do you think?

I was thinking about `$:/state/modal/<tiddler-name>`  The modals already have tiddlers associated. eg: `SampleWizard` for the first demo wizzard. `SampleWizzard 2` and so on. 

>BUT, the individual state tiddlers would be able to replace the extra tiddlers currently needed to set the variables (but later maybe not anymore?). I found out about the need for those only after I had implemented the single-tiddler list version...

It may be possible to use the "text-field" of the new state tiddlers and use JSON datatiddlers to store the variables. I think the `paramObject/1` tiddler already is like that, but I couldn't see it yet. 

--> That's also a problem. Modals are extremely "nasty" to debug, because one cannot interact with the underlaying wiki. So opening state tiddlers and so on is very cumbersome. 

Internally the modal widget already opens a completely new story. ... Why not make it a real story with full functionality. 

https://github.com/Jermolene/TiddlyWiki5/blob/0d1667c4676dea47c58d0fece005b27a690d4caf/core/modules/utils/dom/modal.js#L54-L103

pmario:
>I couldn't recreate that on the preview site, after I cleared the background property on tc-modal-backdrop and made the tc-modal semi-transparent, it looked like this:

I did completely remove the `tc-modal-backdrop` DIV from the template. Then there have been 2 modals. .. I think 1 modal is from the modalwidget in the default TW story river. 

The second modal is from your new template. .. I think that isn't the biggest problem. .. 

The whole debugging complexity bothers me much more. 

What if the modal widget just creates a new "layout" on top of a `tc-modal-backdrop` ... Where the backdrop contains a navigator-widget with a new `$:/StoryList/modals` and `$:/HistoryList/modals` ... 

Modals would have their own ViewTemplate similar to the one they have now, but without the hardcoded stuff in modal.js, which steps on everyone's feet anyway.

yaisog:
> I was thinking about `$:/state/modal/<tiddler-name>`

That would break the example in the "Modals" tiddler, because "SampleWizard 1" and "SampleWizard 2" call each other recursively ad infinitum. That why I the trouble with list de-duplication – that messed with the recursiveness (recursivity?). If that's a useful feature may be debatable, but it would surely break backwards compatibility...
Also, how would you track the order then? Have a field in each state tiddler that lists the predecessor? I really don't want to add new fields to the core if I don't have to.

> [...] use JSON datatiddlers to store the variables. I think the paramObject/1 tiddler already is like that, but I couldn't see it yet.

You're exactly right. The `paramObject` state tiddlers are JSON data tiddlers already.

> Where the backdrop contains a navigator-widget with a new $:/StoryList/modals and $:/HistoryList/modals ...

I did use the story and history of the context calling the modal, in order to be able for the modal to open tiddlers in the underlying story. This is important for my personal application of modals, which perform a search and open the results back in the original story.

> Modals would have their own ViewTemplate similar to the one they have now [...]

I like the idea. Makes it much more flexible. It would be a matter of exporting the hardcoded structure into a tag-based ViewTemplate one. This would add an extra tag for the modal-specific ViewTemplate parts. One could call it `$:/tags/ModalTemplate`. The original template would have to look like it does now.

Also, one thing that bothers me about the current ViewTemplate is that I cannot replace parts of it with my own without touching the original template core tiddlers (to remove the `$:/tags/ViewTemplate` tag). I would like a `$:/config` option to hide them...

pmario:
>Also, one thing that bothers me about the current ViewTemplate is that I cannot replace parts of it with my own without touching the original template core tiddlers (to remove the $:/tags/ViewTemplate tag). I would like a $:/config option to hide them...

I think that already happened with the ViewTemplate cascades. It's actually easy to replace everything in the story-river with the Story Tiddler cascade already. 

But cascades are a "brain-twister" on their own. It's very important to get the naming right

grafik

pmario:
>I did use the story and history of the context calling the modal, in order to be able for the modal to open tiddlers in the underlying story. This is important for my personal application of modals, which perform a search and open the results back in the original story.

pmario:
>I did use the story and history of the context calling the modal, in order to be able for the modal to open tiddlers in the underlying story. This is important for my personal application of modals, which perform a search and open the results back in the original story.

Yea, that's why every message that wants to update the story should send a `tv-story-list` and a `tv-history-list` with the params -- and -- the navigator should use this info to modify the story and history accordingly. So the user can decide which story should be modified, by changing those vars. 

With that in place it would be possible to handle 100 different stories with only 1 navigator widget. There is absolutely no need to have more than 1 navigator widget in the whole wiki. 

It would be super easy to to make multi-column stories, where the use can decide which link opens a new tiddler and where. 

I think 80% of the needed fixes are in place already. .. BUT as you know the last 20% are the hardest.

pmario:
>That would break the example in the "Modals" tiddler, because "SampleWizard 1" and "SampleWizard 2" call each other recursively ad infinitum. That why I the trouble with list de-duplication – that messed with the recursiveness (recursivity?). If that's a useful feature may be debatable, but it would surely break backwards compatibility...

I think we should let the existing mechanism alone. The `tm-modal` message is handled by the "root-widget". 

As modals are used now, it needs to be "outside" the PageTemplate, because if users mess up the page template with broken themes or broken layouts, it's possible to switch back to the default layout. .. If there is one. 

The modals that I'm envisioning are part of the current theme. They are a different layout... kind of. So theoretically it should be possible to switch between the modal-layout and the default-layout where both of them keep their state. (dev mode only!!!!)

That would make the development of modals much easier, because it could be done at the same time. ...

If modals have their own history, it should be straight forward to keep track of them. It would even be possible to open a modal in a tiddler, for testing. Instead of switching layouts. 

----------

I personally have to say that I am __not a fan__ of modals. I think most of the time they are used wrong. 

<details><summary>What I would want</summary>

I personally would like to have a classic story river, where I could "disable all tiddler interactions" except for 1 tiddler. 
All tiddlers are "disabled" and "dimmed down" like the backdrop. 
Interacting with them is not possible, but it is still possible to read them. 
Only 1 tiddler has full possibilities.
.. This would allow me to have full readability for the whole content, but an author could guide users to interact with the 1 tiddler that has focus. 

</details>

yaisog:
> The modals that I'm envisioning [...]

What you are advocating is some brand new functionality that might be added to TW some day. I don't think that should be called _modals_, though. That is not where this PR is headed, but it may build on it. I'm just trying to bring the modal code into WikiText, to make it flexible and easily extensible / customizable. A separate ViewTemplate is a good idea in that direction.

For me, modals are a one-task, focused, popup-like mechanism. Like the layout switcher, or a search. Does anyone ever use the wizard-like chaining of multiple modals? And should "close" then really go back to the previous one? For the sake of flexibility, let alone backwards compatibility, the answer must be "yes", I guess. Let the modal creator handle all the logic of what to show when...

> As modals are used now, it needs to be "outside" the PageTemplate

I don't think that's possible with a WikiText-based implementation. I think we'll have to go with the good old-fashioned "reload a previous save" way of recovery... That's what I do when working on this modal implementation, anyways.

> If modals have their own history, it should be straight forward to keep track of them.

I would run into the problem of making the param hashMap available as variables: I need state tiddlers to hold them, but since I might open the same tiddler multiple times with different hashMaps, the state tiddlers cannot be based on the tiddler title. They must be sequentially numbered. If I already have that, it doesn't make sense to track a history on top of that.

> Modals are extremely "nasty" to debug, because one cannot interact with the underlaying wiki. So opening state tiddlers and so on is very cumbersome.

I guess that is the definition of a modal – interaction with underlying content is disabled.

I think what I'll try next is
- Change from a list-based modal `$list` to a prefix-based, numbered list of JSON data tiddlers, `$:/state/modal/<n>`. The title of the modal tiddler is either put into a field or into the JSON data, along with the hashMap.
- Make the layout flexible by creating a separate configurable ViewTemplate, `$:/tags/ModalTemplate`. By default it will look like current modals. This will result in a lot of layout core tiddlers.
- For debugging, I'll add a switch that removes the backdrop from the modal template, so interaction with the background wiki remains possible. This will turn the modal into a popup, basically.

pmario:
>> As modals are used now, it needs to be "outside" the PageTemplate
>
>I don't think that's possible with a WikiText-based implementation. I think we'll have to go with the good old-fashioned "reload a previous save" way of recovery... That's what I do when working on this modal implementation, anyways.

Sorry I haven't been clear enough. Saq wrote it much clearer at the other thread: https://github.com/Jermolene/TiddlyWiki5/issues/7267#issuecomment-1434342912 

That's why I wrote

>I think we should let the existing mechanism alone. The tm-modal message is handled by the "root-widget".

It is implemented like this: 

https://github.com/Jermolene/TiddlyWiki5/blob/0d1667c4676dea47c58d0fece005b27a690d4caf/core/modules/startup/rootwidget.js#L28-L30

As you can see the `$tw.modal.display("$:/core/ui/SwitcherModal",{variables: event.paramObject, event: event});` "links" the event hashmap variables to the variables object of the modal, which makes them available. 

I wanted to suggest, that there should be a new `tm-wikimodal` as you named it that activates a much simpler modal structure

As shown at: https://github.com/Jermolene/TiddlyWiki5/pull/7268#issuecomment-1434147727, the structure that is created at the moment is hardcoded, complex and not flexible. 

What I would like is much more lightweight. ... 

- It only should contain the "backdrop" 
  - which optionally disables the interaction with the wiki below (for easy debugging)
  - optionally dims the background (for easy debugging)
- loads a new modal-story-template
  - This template may contain your dynamic message catcher
  - a list-widget that allows us to define a "single tiddler" story which displays the modals. 
  - There should be a new modal-template which looks the same as the existing ones but is not hardcoded

Not sure if that makes sense. ... But .. what I want is that modals are normal tiddlers with a different ViewTemplate. They should **not be special**, because special handling creates extra complexity.

yaisog:
I just committed a major overhaul of the implementation.

A major change is the alternative way of opening modals by calling the `$action-modal` widget. I think this is preferrable over the clunky message mechanism, and since I didn't design it as a replacement for the old modals (where the `$root` widget will get some action) but as a side-by-side, an action widget is the more elegant solution.

The widget is fairly simple and only creates the state (data) tiddler from the paramObject.

The `tm-wikimodal` message mechanism still works, though. It does create the same state tiddler.

There is now a `$debug="yes"` (widget) or `debug="yes"` (message) attribute to remove the backdrop and make the underlying wiki remain interactive.

I have added plenty of examples that can be tested from the "ModalExamples" tiddler, which should open at the top upon loading.

The modals use a ModalTemplate similar to the ViewTemplate, fully configurable via cascade and tags. A custom template can also be forced by using the `$template` / `template` attribute (see examples). If the modal tiddler has any of the "old" fields, a legacy template is shown that looks like the `tm-modal` one. The new template always shows a close button in the upper right corner, so that you cannot brick your wiki by using a footer that has no close functionality.

There still are no animations.

yaisog:
These WikiText modal will probably never be able to fully replace the current modals, because they must originate in the PageTemplate (think e.g. PageTemplateChooser). Therefore, they will likely co-exist with the current implementation, should they ever make it into the core. I am therefore quite inclined to drop the baggage that the current implementation brings, with all kinds of parameters that are settable via configuration fields.

Especially, if we could agree on calling the modal via an `$action-modal` widget, I would like to configure it via attributes instead.

However, I don't know if there is a reason for `tm-modal` except maybe the need to pass it up all the way to $root. Which in this parallel implementation would not be necessary. Anything else I'm missing?

core/ui/PageTemplate/modals.tid
```diff
@@ -0,0 +1,93 @@
...
+\end
+
+\define tm-close-handler()
+<!-- Close the modal unless a different tiddler was given in the message parameter -->
+<$list filter="[<event-param>!is[blank]!match<currentTiddler>]" emptyMessage=<<close-current-modal>> >
+<$action-sendmessage $message="tm-close-tiddler" $param=<<event-param>>/>
+</$list>
+</$let>
+\end
+
```
pmario, line 18: IMO Instead of using `<$let stateTiddler` inside the macro you could define a function outside the macro

```
\function stateTiddler() [removeprefix[$:/state/modal/]sortan[]add[1]else[0]addprefix[$:/state/modal/]]
```

pmario (commented):
This are issues I did see reading the code here. I did not do any tests with the wiki preview version yet.

PR #7272, opened 2023-02-17
Improve documentation of WikiText formatting

mateuszwilczek:
Add information about `@@highlight@@` syntax


Jermolene:
Thank you @mateuszwilczek these are very good quality improvements, and the material you've added is clearly written and concise. Much appreciated! Documentation updates aren't glamorous, but they make an enormous difference.

PR #7281, opened 2023-02-23
fix: correctly identify paste events to be handled by editor widgets

### LINKED ISSUE #7244: [BUG] application/json tiddler in edit mode -- copy paste to text area also creates an $:/Import tiddler
### Describe the bug

As written in the title, "application/json tiddler in edit mode -- copy paste to text area also creates an $:/Import tiddler"

It seems the problem only comes up with CM installed. 

### Expected behavior

text should only be copied into the text area

### To Reproduce

- go to: https://paste.tiddlyhost.com/
- open the Json Test tiddler
- copy some text from the text area to the clipboard
- Paste the text. -> Problem

-----

Testing the same thing with tiddlywiki.com there is **no problem**

### Screenshots

image

### TiddlyWiki Configuration

TW v5.2.5 

### Additional context

> from [TTalk thread](https://talk.tiddlywiki.org/t/disable-import-on-paste/5943)
>
>... While I don’t fully grok the issue the PR fixes, in my case it doesn’t seem to be the type of the pasted element, but the fact that the dropzone that receives the event doesn’t filter by that type. I’m basically editing text and yet the dropzone behaves as if any sort of paste should be handled as an import. This manifests itself for me since the changes I made to CM mean that the event target is no longer a textarea.
>
>In an empty TW without these settings, if the content type of the tiddler that is being edited is empty, a dropzone with the contentTypeFilter is called, but because it doesn’t do anything, since the target is a textarea, the event gets propagated to another dropzone which again does nothing for the same condition. If the content type is for javascript, only the second dropzone is called and it does nothing.
>
>In my TW, when the tiddler’s content type is empty, the first dropzone gets called, it enters the “!textarea” condition but in there it calls readFileCallback that does nothing due to the content filter. But since it is in the condition body, it stops propagation so the second dropzone is not called. When the content type is for javascript, only the second dropzone is called, it goes into the condition, but since it doesn’t have a filtertype, it handles the import.
>
>I don’t know how to do it, but I think the first dropzone should remain when the tiddler that is edited has a content type of json/javascript (or other text formats). While were at the subject, it seems to me that import should trigger when the content type filter doesn’t match the pasted type (it should be a whitelist), and not the way it is now (when it acts as a blacklist).

saqimtiaz:
This PR allows paste events that should be handled by an editor widget to be correctly identified by the dropzone widget and fixes #7244 

The attribute `twEditor` is added to the paste event by the codemirror plugin, which is used as part of the check in $dropzone to determine whether the event originated from an editor. Other editor widgets would be able to use the same attribute if/when necessary and this is more robust than checking for a specific event target or DOM structure.

@pmario @ittayd can you please check at https://tiddlywiki5-8gsvrvjx2-jermolene.vercel.app/ if this resolves the problem for you and report back as to your browser and OS? I have tested on the latest Chrome on Windows and Ubuntu. In particular this needs testing on Firefox.

### DIFF
--- a/core/modules/widgets/dropzone.js
+++ b/core/modules/widgets/dropzone.js
@@ -259,7 +259,7 @@ DropZoneWidget.prototype.handlePasteEvent  = function(event) {
 			}
 		};
 	// Let the browser handle it if we're in a textarea or input box
-	if(["TEXTAREA","INPUT"].indexOf(event.target.tagName) == -1 && !event.target.isContentEditable) {
+	if(["TEXTAREA","INPUT"].indexOf(event.target.tagName) == -1 && !event.target.isContentEditable && !event.twEditor) {
 		var self = this,
 			items = event.clipboardData.items;
 		// Enumerate the clipboard items
--- a/plugins/tiddlywiki/codemirror/engine.js
+++ b/plugins/tiddlywiki/codemirror/engine.js
@@ -220,9 +220,15 @@ function CodeMirrorEngine(options) {
 			}
 		});
 		this.cm.on("paste",function(cm,event) {
+			event["twEditor"] = true;
 			self.widget.handlePasteEvent.call(self.widget,event);
 		});
+	} else {
+		this.cm.on("paste",function(cm,event){
+			event["twEditor"] = true;
+		});
 	}
+;
 }
 
 /*


Jermolene:
Thanks @saqimtiaz. The technique of adding custom properties to an event generated by the DOM is interesting, and I know from experience that it works, but I did go down a brief rabbit hole to try to discover if it is reliable and considered good practice. The MDN docs seem to suggest creating a custom event with the additional data in the 'details' property. I found a few Stack Overflow questions on the topic, but none with decent answers. I'm very happy to stick with this approach as it is much clearer than messing around with custom event objects, but would be interested in citations about it.

saqimtiaz:
> The technique of adding custom properties to an event generated by the DOM is interesting, and I know from experience that it works, but I did go down a brief rabbit hole to try to discover if it is reliable and considered good practice.

@Jermolene I went down the same rabbit hole earlier today as well and was not able to find anything definitive as to whether this is recommended practice. I do know from prior experience that it does work across all major browsers. I went with this approach in the interest of simplicity, but it would not be difficult to switch to creating a custom event if there was reason to do so.

Jermolene:
Great @saqimtiaz I'm glad it's not just me! Let's leave things as they are for the moment.

pmario:
@saqimtiaz @Jermolene ... I did test that branch with codemirror active. It looks good for me.


- FF latest and Windows 11 ... Works as is should
- Edge latest on Windows 11 ... Copy paste works

but spell-checking seems to be ignored by Edge, FF is OK. .. BUT that's a different thing to explore.

pmario:
@saqimtiaz ... I think you can remove the "Draft" flag

saqimtiaz:
Thank you @pmario, much appreciated.

@Jermolene this should be good to go.

Jermolene:
I can test this on Safari later today – what are the steps for testing it?

pmario:
To reproduce the problem.

- Go to https://paste.tiddlyhost.com/
- edit JSON Test
- type eg: asdf  -> Should show the browser spellchecking info .. red underline
- select any text
- CTRL C .. copy to clipboard
- CTRL V .. insert  -> problem .. Opens the $:/Import tiddler. 

With this PR: 

Open TW like so: 

```
tiddlywiki.cmd +plugins/tiddlywiki/codemirror .\editions\tw5.com-server\ --list
```

- import the 3 tiddlers from the ZIP
- [test-cm-contenteditable-spellcheck.zip](https://github.com/Jermolene/TiddlyWiki5/files/10833943/test-cm-contenteditable-spellcheck.zip)
- open JSON Test
- copy / paste action should not open the $:/Import tiddler.
- browser spellchecking should work. 

It's important that the edited tiddler is `type: application/json`

image

saqimtiaz:
@Jermolene I have tested this PR on Safari 16.3 (latest) on OSX and confirm that it works as expected.

Jermolene:
Great thanks @saqimtiaz

PR #7284, opened 2023-02-24
markdown-plugin: Handle non-string attributes in TW-style image format

cdruan:
Overlooked that tiddlywiki-style image syntax can contain non-string attributes such as transclusions and macro calls.

This PR fixes that so content such as `[img width={{!!mywidth}} class=<<image-classes>> [Motovun Jack.jpg]]` will be rendered correctly in markdown tiddlers.


Jermolene:
Thanks @cdruan

Issue #7277, opened 2023-02-22
[IDEA] Delay the backlink indexer


linonetwo:
**Is your feature request related to a problem? Please describe.**

It consumes too much time to re-parse tiddlers.

![截屏2023-02-22 13 44 54](https://user-images.githubusercontent.com/3746270/220533776-b1c4b471-1be5-406c-b473-362437727c5d.png)

The direct cause is a list widget's filter uses backlink filter. And this list widget is on a ViewTemplate that show under every tiddler.

**Describe the solution you'd like**

Delay execution of this indexer, or ask this widget lazy render it self after FCP: do these heavy job only after First Contentful Paint.

https://github.com/Jermolene/TiddlyWiki5/blob/95dc56d8508109a76c232cfbcd858889d32916a9/core/modules/startup/render.js#L64-L66

For example, just return in the `render()`, and when `$tw.hooks.invokeHook("th-page-refreshed");` fired, render those lazy widgets again?

I'm not sure if this is possible, need to ask @Jermolene 

![截屏2023-02-22 13 46 38](https://user-images.githubusercontent.com/3746270/220534103-fc69e2b3-4c1d-463c-9209-0b231079beec.png)

**Describe alternatives you've considered**

Cache the inder parse result in 

```
BacklinksIndexer.prototype._getLinks = function(tiddler) {
// need cache here, so it won't need to prepare it on each page refresh.
	var parser =  this.wiki.parseText(tiddler.fields.type, tiddler.fields.text, {});
	if(parser) {
		return this.wiki.extractLinks(parser.tree);
	}
	return [];
}
```

**Additional context**

Still trying to fix #7151

pmario:
>Delay execution of this indexer, or ask this widget lazy render it self after FCP

What is FCP?

pmario:
>The direct cause is a list widget's filter uses backlink filter. And this list widget is on a ViewTemplate that show under every tiddler.

Why don't you hide it and show after user interaction?

linonetwo:
> > Delay execution of this indexer, or ask this widget lazy render it self after FCP
> 
> What is FCP?

It is a web vital https://web.dev/fcp/

linonetwo:
> > The direct cause is a list widget's filter uses backlink filter. And this list widget is on a ViewTemplate that show under every tiddler.
> 
> Why don't you hide it and show after user interaction?

Good idea, I can create a lazy widget to hide it , and show it when `th-page-refreshed` message received.

[...]

PR #7285, opened 2023-02-24
wip parseutils add callback, html.js avoid extensive backtracking

pmario:
## Test - Work in progress

@linonetwo ... Can you test this branch with your wiki mentioned at: https://github.com/Jermolene/TiddlyWiki5/issues/7277
And send me chrome profile. I want to see the differences

This PR tries to improve html and widget parsing for big blocks of html-like wikitext, **which has no line-breaks**

The TW "isBlock" detection uses this regexp `([^\S\n\r]*\r?\n(?:[^\S\n\r]*\r?\n|$))` which can need a lot of time if there are no linebreaks till the end of the tiddler. So the regexp "fails late". 

The example in the screenshot shows an UL block with 3 LI blocks. Where the detection finds the `\n\n` after 64 regexp iterations. Every iteration (for 1 char in the test text) needs 4 steps. ... The regexp is OK it can not be improved. 

image

- This regexp is evaluated for every TW block-detection. see [wikiparser/rules/html.js](https://github.com/Jermolene/TiddlyWiki5/blob/2271f6885af8b6301696d3aa33f82472f691ba72/core/modules/parsers/wikiparser/rules/html.js#L51)
- It finds the end of the UL block and throws away this info
- It searches again for the first LI and throws away this info
- 2nd LI and throws away this info
- 3rd LI and throws away this info

So it needs 1 UL iteration to find the `\n\n` and there are 3 LI iterations, where we know the end of the block already, but we throw this info away. ... 

The new code returns the info from the utility function `$tw.utils.parseTokenRegExp(this.parser.source,this.parser.pos,/([^\S\n\r]*\r?\n(?:[^\S\n\r]*\r?\n|$))/g)` into the calling function **with a callback(),** so the result can be reused. ... 

For long HTML blocks with no linebreaks this optimisation can save a lot of time. In the testcase from issue **[IDEA] Delay the backlink indexer #7277** it's a 30% improvement at TW startup-time. ... 

**OLD performance profile** can be inspected with Chrome or Edge

[tw-loading-profiling.json.zip](https://github.com/Jermolene/TiddlyWiki5/files/10833239/tw-loading-profiling.json.zip)

image

**NEW performance profile** can be inspected with Chrome or Edge

[wiki-backlink-withmodification.json.zip](https://github.com/Jermolene/TiddlyWiki5/files/10833238/wiki-backlink-withmodification.json.zip)

image

### DIFF
--- a/core/modules/parsers/parseutils.js
+++ b/core/modules/parsers/parseutils.js
@@ -86,13 +86,15 @@ exports.parseTokenString = function(source,pos,token) {
 /*
 Look for a token matching a regex. Returns null if not found, otherwise returns {type: "regexp", match:, start:, end:,}
 */
-exports.parseTokenRegExp = function(source,pos,reToken) {
+exports.parseTokenRegExp = function(source,pos,reToken,options) {
+	options = options || {};
 	var node = {
 		type: "regexp",
 		start: pos
 	};
 	reToken.lastIndex = pos;
 	node.match = reToken.exec(source);
+	(options.cb) ? options.cb(node) : null;
 	if(node.match && node.match.index === pos) {
 		node.end = pos + node.match[0].length;
 		return node;
--- a/core/modules/parsers/wikiparser/rules/html.js
+++ b/core/modules/parsers/wikiparser/rules/html.js
@@ -46,7 +46,7 @@ exports.parse = function() {
 	// Retrieve the most recent match so that recursive calls don't overwrite it
 	var self = this;
 	var options = {};
-	var hasLineBreak;
+	var hasLineBreak=false;
 	var tag = this.nextTag;
 	this.nextTag = null;
 	// Advance the parser position to past the tag
--- a/core/modules/parsers/wikiparser/rules/html.js
+++ b/core/modules/parsers/wikiparser/rules/html.js
@@ -28,6 +28,7 @@ exports.types = {inline: true, block: true};
 
 exports.init = function(parser) {
 	this.parser = parser;
+	this.matchInfo = {match:{}};
 };
 
 exports.findNextMatch = function(startPos) {
@@ -43,12 +44,27 @@ Parse the most recent match
 */
 exports.parse = function() {
 	// Retrieve the most recent match so that recursive calls don't overwrite it
+	var self = this;
+	var options = {};
+	var hasLineBreak;
 	var tag = this.nextTag;
 	this.nextTag = null;
 	// Advance the parser position to past the tag
 	this.parser.pos = tag.end;
-	// Check for an immediately following double linebreak
-	var hasLineBreak = !tag.isSelfClosing && !!$tw.utils.parseTokenRegExp(this.parser.source,this.parser.pos,/([^\S\n\r]*\r?\n(?:[^\S\n\r]*\r?\n|$))/g);
+
+	// Crecate a callback, to have access to node.match.index
+	options.cb = function(node) {
+		self.matchInfo = node;
+	}
+
+	// Check for an immediately following double linebreak if we don't know if it is a block
+	// BP: this.parser.source.match("<!-- xxxxxx -->")[0] === "<!-- xxxxxx -->"
+	var doubleLinebreakAt = (this.matchInfo.match && this.matchInfo.match.index)? this.matchInfo.match.index : 0; // for easy debugging
+	if (this.parser.pos >= doubleLinebreakAt) {
+		// TODO if we could guarantee, that our source code does not have \r we could make this detection 25% faster
+		var node = $tw.utils.parseTokenRegExp(this.parser.source,this.parser.pos,/([^\S\n\r]*\r?\n(?:[^\S\n\r]*\r?\n|$))/g,options);
+		hasLineBreak = !tag.isSelfClosing && !!node;
+	}
 	// Set whether we're in block mode
 	tag.isBlock = this.is.block || hasLineBreak;
 	// Parse the body if we need to


Jermolene:
Thanks @pmario – I don't wish to pick on you, but from my perspective it is a good example of a poor quality PR:

* The title that doesn't explain what it does or its purpose
* The OP is not a self-contained explanation of the purpose of the changes

I appreciate that you probably only made the PR as a convenient way of getting the Vercel preview, but please consider the impact on others of working like this. A new PR is visible to all collaborators, and so best practice is to make sure that even the first draft version of the PR communicates clearly to everyone.

I came across [this post](https://github.com/Jermolene/TiddlyWiki5/discussions/5484) that I wrote two years ago that explains why this is a problem:

> Quite a few of the PRs we get are not of the highest quality, even sometimes from regular contributors. (My pet bugbear is PRs that are not self contained, and require reading through a Google Groups discussion to understand). I generally still try to engage with such PRs but it makes more work for me, hits my motivation, and slows things down. In an ideal world, PRs would be put together with primary consideration for ease of merging. That means making a detailed description, rationale, screenshots, and anything else to help people understand the proposal

pmario:
Sorry -- It's a _draft-PR_ and says wip as the first word. ... I was a bit in a hurry and just wanted that @linonetwo can pull a branch for testing. I could not reproduce the exact problem .. only having the profile-log for guessing. 

I'll fix the OP later today.

Jermolene:
> Sorry -- It's a _draft-PR_ and says wip as the first word. ... I was a bit in a hurry and just wanted that @linonetwo can pull a branch for testing. I could not reproduce the exact problem .. only having the profile-log for guessing.
> 
> I'll fix the OP later today.

That's what I thought. But the problem is that GitHub is a shared space, and so everything we do is visible to everyone else. Therefore we have to take the trouble to make sure that we communicate clearly what's going on.

linonetwo:
Thanks, @pmario , but it didn't change much.

[wiki-backlink-withmodification.json.zip](https://github.com/Jermolene/TiddlyWiki5/files/10829811/wiki-backlink-withmodification.json.zip)


@Jermolene I will create a separate repo to place a Huge wiki edition for this kind of tests.

pmario:
>Thanks, @pmario , but it didn't change much.

If it shows the same tiddlers at startup, there is a big difference. IMO initial time did go up from 9 sec to 6 sec, which in my book is a 30% improvement. 

The tiddlers after first-paint have gone from 1sec+ to max 400ms, which is less of half the time. .. So imo it is a big improvement. 

The code in the PR removed the big red "parseInlineRunTerminated" ... blocks. ... That was intended. So there has been success. 

But there is still a problem with the TW "is block mode" detection, which is the brown regexp `([^\S\n\r]*\r?\n(?:[^\S\n\r]*\r?\n|$))` that needs much time without doing anything. 

**Is it possible that I can have access to the wiki?** So I can see, what is parsed. .. It seems the regexp is only wasting time with no result. I'm pretty sure all of this unneeded work can be avoided.

pmario:
@Jermolene .. I did update the OP with some more info. Hope that makes it clearer. The code till now could improve the parsing times quite a bit. ... BUT there is still something wrong with the "brown" blocks seen in the screenshots.

Jermolene:
Hi @pmario I think there may be a much simpler solution here. The html parser is trying to check for a double line break immediately after the opening tag. However, the `parseTokenRegExp()` method doesn't really do that: it searches for the next double line break, and then checks if it was found at the position in question. So, I think the way to approach this is to tweak `parseTokenRegExp()` so that it only looks for a match at the specified position, rather than searching from that position.

Jermolene:
@pmario @linonetwo I meant to add that I'm delighted to see profiling being used to track down performance issues like this, and I'd love to see more of this.

pmario:
>Hi @pmario I think there may be a much simpler solution here. The html parser is trying to check for a double line break immediately after the opening tag. However, the parseTokenRegExp() method doesn't really do that: it searches for the next double line break, and then checks if it was found at the position in question. So, I think the way to approach this is to tweak parseTokenRegExp() so that it only looks for a match at the specified position, rather than searching from that position.

The `parseTokenRegExp()` is a utility function which is used several times in some plugins with a slightly different context. So I did not want to mess with it ... hence the callback, which also isn't the best solution, but good for testing. 

>@pmario @linonetwo I meant to add that I'm delighted to see profiling being used to track down performance issues like this, and I'd love to see more of this.

The problem with those profiles is, that w do see, which widgets are called and if you know how TW UI is built we even have some hints what's going on. But there are several `anonymous()` function calls where we actually don't know what's going on. 

-----------

Slightly OK-- Especially the boot sequence has a lot of anonymous function calls. 

image

Jermolene:
> The `parseTokenRegExp()` is a utility function which is used several times in some plugins with a slightly different context. So I did not want to mess with it ... hence the callback, which also isn't the best solution, but good for testing.

The semantics are very clear, so I think the change I have outlined is safe.

> The problem with those profiles is, that w do see, which widgets are called and if you know how TW UI is built we even have some hints what's going on. But there are several `anonymous()` function calls where we actually don't know what's going on.

I think I've mentioned this elsewhere, but I'd welcome PRs that add meaningful short names to these anonymous functions.

linonetwo:
@pmario  I've created a https://github.com/tiddly-gittly/HugeWikiExample , but not able to reproduce problem in my wiki (Though I try to add many wikilink, PPT, PDF, and images)

So I'm going to send you a snapshot of my private wiki in HTML format, you can extract it to nodejs wiki if you'd like.

pmario:
See: https://github.com/tiddly-gittly/HugeWikiExample/issues/2 ... IMO you should serve the huge PDFs and PPTs from the files directory outside of tiddlers/ !!

Jermolene:
Hi @pmario see #7297 for a different approach.

pmario:
https://github.com/Jermolene/TiddlyWiki5/pull/7297 .. That's nice. I was not aware of the y-flag even if it seems to be part of the browsers for a long time. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp

Anyway. I'm glad I could help with the experiments. Even better if they improve the overall performance of TW \o/

Jermolene:
> #7297 .. That's nice. I was not aware of the y-flag even if it seems to be part of the browsers for a long time. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp
> 
> Anyway. I'm glad I could help with the experiments. Even better if they improve the overall performance of TW \o/

Thanks @pmario getting that fixed was thanks to this thread.

PR #7286, opened 2023-02-24
Add two examples for text selection

yaisog:
I added two examples in the doc tiddler for the EditTextWidget for the new text selection functionality added in PR #7222.

The second example is a bit advanced and shows off some intermediate filtered transclusion functionality (concatenation of strings, calling macros with parameters from within filters). Depending on your point-of-view, this may either be welcome or confusing. Personally, I like it as a demonstration of good practice.


Jermolene:
Thanks @yaisog

PR #7287, opened 2023-02-24
Signing the CLA

andrigamerita:


Jermolene:
Thanks @andrigamerita and welcome to the project! Is there anything particularly you're interested in helping with? Do let me know if there's anything I can do to help.

andrigamerita:
Thank you @Jermolene. After that patch right now I don't have any other work-in-progress or planned contribution to the project, but if something comes to mind I'll surely let you and other contributors know.

Or, well... actually yesterday I started some light work on trying to more or less modify TiddlyWiki's core, and I don't really know if this could only really develop into a fork or it's something the TiddlyWiki project as a whole can make direct use of, but what I'm trying to achieve is to assemble a variant of this software that can weigh a bit less, both as in file size and performance. Just by partially compressing all the JavaScript I got empty.html v5.2.5 from 2.3 to 2.0 MB, but I already know what I can do to go even smaller without removing features.

Anyways, I saw the announcement for v5.3.0 just now, very excited for the release! Keep up the excellent work :+1:

Jermolene:
Thanks @andrigamerita 

> what I'm trying to achieve is to assemble a variant of this software that can weigh a bit less, both as in file size and performance

In terms of file size, there is a lot of low hanging fruit because it has not had much attention historically.

One simple change to the development process that I'd like to do is to add some file size metrics to the test suite so that we can add a report to each PR giving the absolute and percentage file size increase that it means for empty.html, (and for each individual plugin, too, but it should only display figures that have changed). That would at least raise awareness of the impact of changes, and encourage more consideration of the impact on file size.

A couple of bits of historical background:

* I am not in favour of minifying the core JavaScript because it makes debugging and hacking so much harder (especially since sourcemaps are not convenient with a single file unhosted system like TiddlyWiki)
* TiddlyWiki compresses very well with GZip compression (2.4MB -> 400KB in my tests), and so most of these considerations are moot when working online over HTTP with GZip compression enabled
* In the future we may be able to deprecate some older functionality, moving the relevant code to an optional plugin

Nonetheless, we consistently get feedback from users concerned with file size, and so I'm interested to explore what we might be able to do

andrigamerita:
@Jermolene I think that would be a good idea, better to have those metrics than not.

I never noticed before that github.io has compression, because unfortunately gitlab.io doesn't, and I have my wiki served from that... Maybe I should just point my domain to Github, as my wiki is already also on it for backup reasons, and my workflow would remain the same.

Anyway, one feature I feel TiddlyWiki lacks now that is comes to my mind is tiddler compression/decompression. My text tiddlers are all pretty small for now, but I have many plugins, some of them really big. I think automatic compression of tiddlers on wiki saving (maybe with zstd, the most efficient algorithm around), and decompression on startup, can be implemented in a similar way to how encryption is done, speaking of the technical side.
This would make it easier to download, save, or share TiddlyWikis without having to manually enclose the HTML file, and then (even worse) extract it, from an archive that needs external programs.

Jermolene:
Thanks @andrigamerita I've added a ticket for the size checks: #7291 

I'm not sure that there is a compelling case for integrating compression into the core: again, the issue is that the alternative of applying compression to the entire wiki is so well supported by modern toolchains that it seems redundant to duplicate that functionality within TiddlyWiki. So perhaps it's something that should be explored via plugins in the first instance.

I hadn't seen zstd (https://github.com/OneIdentity/zstd-js), it's interesting but it seems to be quite large if one wants both compression and decompression. We do already have JSZip integrated which I think could be an alternative.

PR #7290, opened 2023-02-25
Introduce new levenshtein, makepatches, applypatches operators

Jermolene:
These new operators expose additional functionality from the [diff-match-patch](https://github.com/google/diff-match-patch) library that TiddlyWiki already integrates.

The new operators are as follows:

* `[<a>levenshtein<b>]` - computes the number of inserted, deleted or substituted characters to turn one string into another. Often used as a measure of similarity between strings
* `[<a>makepatches<b>]` - computes the patches needed to turn one string into another
* `[<a>applypatches<b>]` - applies the patches in the operand to the input strings

See the docs of diff-match-patch for the gory details: https://github.com/google/diff-match-patch/wiki/API

To do:

* [ ] Add documentation and examples

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -112,7 +112,7 @@ function diffPartsToChars(text1,text2,mode) {
 	        if(mode === "words") {
                 regexpResult = searchRegexp.exec(text);
                 lineEnd = searchRegexp.lastIndex;
-                if(lineEnd === null) {
+                if(regexpResult === null) {
                 lineEnd = text.length;
                 }
                 lineEnd = --lineEnd;
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -74,6 +74,32 @@ exports.join = makeStringReducingOperator(
 	},null
 );
 
+var dmp = require("$:/core/modules/utils/diff-match-patch/diff_match_patch.js");
+
+exports.levenshtein = makeStringBinaryOperator(
+	function(a,b) {
+		var dmpObject = new dmp.diff_match_patch(),
+			diffs = dmpObject.diff_main(a,b);
+		return [dmpObject.diff_levenshtein(diffs) + ""];
+	}
+);
+
+exports.makepatches = makeStringBinaryOperator(
+	function(a,b) {
+		var dmpObject = new dmp.diff_match_patch(),
+			patches = dmpObject.patch_make(a,b);
+		return [dmpObject.patch_toText(patches)];
+	}
+);
+
+exports.applypatches = makeStringBinaryOperator(
+	function(a,b) {
+		var dmpObject = new dmp.diff_match_patch(),
+			patches = dmpObject.patch_fromText(b);
+		return [dmpObject.patch_apply(patches,a)[0]];
+	}
+);
+
 function makeStringBinaryOperator(fnCalc) {
 	return function(source,operator,options) {
 		var result = [];
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -84,13 +84,86 @@ exports.levenshtein = makeStringBinaryOperator(
 	}
 );
 
-exports.makepatches = makeStringBinaryOperator(
-	function(a,b) {
-		var dmpObject = new dmp.diff_match_patch(),
-			patches = dmpObject.patch_make(a,b);
-		return [dmpObject.patch_toText(patches)];
-	}
-);
+// these two functions are adapted from https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs
+function diffLineWordMode(text1,text2,mode) {
+	var dmpObject = new dmp.diff_match_patch();
+	var a = diffPartsToChars(text1,text2,mode);
+	var lineText1 = a.chars1;
+	var lineText2 = a.chars2;
+	var lineArray = a.lineArray;
+	var diffs = dmpObject.diff_main(lineText1,lineText2,false);
+	dmpObject.diff_charsToLines_(diffs,lineArray);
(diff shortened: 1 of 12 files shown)


yaisog:
I've played around a bit with it and got some RSoEs for invalid patch strings. These strings should probably be checked for fitness.

In one case I had inline transclusion filter with makepatches in a tiddler (text), and transcluded that tiddler into the applypatches call. The error then was:
```
Error: Invalid patch string: {{{ [{Text 1}makepatches{Text 2}] }}}
```
i.e. the filter didn't get evaluated before being used. I'm not sure if this is supposed to work in the first place, though.

yaisog:
I can also offer some help with the docs, if you'd like.

pmario:
At Talk I also did some experiments. https://talk.tiddlywiki.org/t/what-if-we-would-implement-a-3-way-merge-system-into-tw/6248/12?u=pmario which may be usable in the docs

yaisog:
The coordinates of the resulting patches are wrapped in `@@`, which activates highlighting when tiddlers with such info are viewed in the story. Not a problem, really, but maybe unexpected.

pmario:
@Jermolene ... I think you can merge this one. We will get the docs going if it is merged

pmario:
@yaisog The https://github.com/Jermolene/TiddlyWiki5/pull/7290#issuecomment-1453188356  imo should be 2 new issues. They are a bit OT here.

yaisog:
@Jermolene, could you see a way to use the position information in the patches data to highlight these changes in a displayed tiddler, possibly in combination with Dynannotate? That would be pretty neat. I don't mean a `diff` view, but simply a highlight over the changed text / word / line, depending on the mode...

Jermolene:
> I can also offer some help with the docs, if you'd like.

Thanks @yaisog that would be much appreciated.

> It would be great if we could implement the `line` and `word` modes described here: https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs These could be made available with an operator suffix.

Good idea. Perhaps we'd also need to add line/word support to the `<$diff-text>` widget.

Are you interested in volunteering? Otherwise I think it would be reasonable to commit the current implementation (once we have the docs) and add `line` and `word` modes in a later version.

> Also, would it be a lot of work to implement the fuzzy search provided by the library's `match_main(),` maybe as an extension of the `search` operator, like `search::fuzzy`? I think this would be a great feature (but also maybe a different PR). I have no idea how to provide the expected location, though, or what a good default for that would be...

Yes, definitely worth exploring. I'd agree it should be a separate PR.

> @Jermolene, could you see a way to use the position information in the patches data to highlight these changes in a displayed tiddler, possibly in combination with Dynannotate? That would be pretty neat. I don't mean a `diff` view, but simply a highlight over the changed text / word / line, depending on the mode...

Good thinking. Perhaps Dynannotate could have a new searchMode attribute value that interprets the search string as a patch. Thinking about it, it might also make sense to extend the `<$diff-text>` widget to take one chunk of text and a set of patches, instead of two chunks of text.

yaisog:
> Are you interested in volunteering?

Perhaps I can get something done over the weekend. The GitHub site has JS example code that does that, so this shouldn't be too much work beyond defining the suffixes, etc.
The would be a PR to this branch, again, right?

> Perhaps Dynannotate could have a new searchMode attribute value that interprets the search string as a patch.

That's not something I can deliver, though, since I'm not very well versed with the Dynannotate code. I was rather amazed what's possible in this direction nowadays...

Jermolene:
> Perhaps I can get something done over the weekend. The GitHub site has JS example code that does that, so this shouldn't be too much work beyond defining the suffixes, etc. The would be a PR to this branch, again, right?

Great, thank you! And yes, a PR to this branch should work.

> That's not something I can deliver, though, since I'm not very well versed with the Dynannotate code. I was rather amazed what's possible in this direction nowadays...

I think it actually probably makes more sense to do address it via the `<$diff-text>` widget; Dynannotate is fairly specific to the implementation detail of applying highlights via an overlay layer. In this case, I think `<$diff-text>`-style annotations are adequate, and simpler to work with.

yaisog:
> I think it actually probably makes more sense to do address it via the `<$diff-text>` widget

You're probably right. The use case I had in mind is to use patches to track changes to tiddlers over time, and then display older versions with highlights where something has changed with the information from these saved patches, i.e. not the exact changes themselves as `$diff-text` does now. Since one would likely show the un-rendered source text of the tiddler anyway, `$diff-text` might very well do. I'm not too familiar with it (yet), though.

Jermolene:
> You're probably right. The use case I had in mind is to use patches to track changes to tiddlers over time, and then display older versions with highlights where something has changed with the information from these saved patches, i.e. not the exact changes themselves as `$diff-text` does now. Since one would likely show the un-rendered source text of the tiddler anyway, `$diff-text` might very well do. I'm not too familiar with it (yet), though.

Great stuff, that would be interesting to investigate.

yaisog:
> Dynannotate is fairly specific to the implementation detail of applying highlights via an overlay layer.

I only saw a certain resemblence between the location numbers that the patch data starts with and the calculations that Dynannotate does to determine the overlay locations. That was just from a quick glance, though.

When this PR is merged, I'll make an issue to the the `$diff-text` idea...

yaisog:
I did a quick speed test on the speed test text that is part of diff-match-patch:
no suffix: 239 ms
words: 18 ms
lines: 2 ms
(all times are median values of 100 filter evaluations)
I will put a summarizing note about speed in the docs if nobody objects.

Jermolene:
Thanks @yaisog please do, those are interesting results.

Jermolene:
Thanks for all your help @yaisog

PR #7292, opened 2023-02-26
Add the timestamp suffix to the format operator

### LINKED ISSUE #7120: [IDEA] add `format:timestamp[dateformat]` filter operator
As of TW5.2.4, a "TIMESTAMP" filter operand has been added: `format:date[TIMESTAMP]` to convert a date input value into a "Unix Epoch" timestamp (number of milliseconds since "January 1, 1970 00:00:00.000")

The following code implements `format:timestamp[dateformat]` to provide the inverse conversion (i.e., from a Unix Epoch timestamp to TWCore date-formatted string.

* First, it checks the input to make sure it is an integer value.  Note that negative integer values are permitted and specify milliseconds BEFORE the beginning of the "Unix Epoch" start date.
* Then, it applies the filter operand to format the output. If no operand is specified, it defaults to a 17-digit UTC TWCore serial date format ("`[UTC]YYYY0MM0DD0hh0mm0ss0XXX`").
* If the input is not a valid signed integer, it returns "Invalid timestamp value"

```text
/*\
title: $:/core/modules/filters/format/timestamp.js
type: application/javascript
module-type: formatfilteroperator
\*/
(function(){

/*global $tw: false */
"use strict";

/*
Export our filter function
*/
exports.timestamp = function(source,operand,options) {
	var results = [];
		if (title.match(/^-?\d+$/)) {
			var value = new Date(Number(title));
			results.push($tw.utils.formatDateString(value,operand || "[UTC]YYYY0MM0DD0hh0mm0ss0XXX"));
		} else{ 
			results.push("Invalid timestamp value");
		}
	});
	return results;
};
})();
```

yaisog:
This is the complementary operation to the `format[TIMESTAMP]` syntax and allows values that represent the Unix epoch to be converted back to any TW date format.

Invalid input titles will be passed through unchanged.

Initial commit taken verbatim from @ericshulman.

Closes #7120.

To do:
- [x] Add tests
- [x] Add more tests
- [x] Correct handling of invalid input


saqimtiaz:
It would be nice to add some tests for the new suffix to the format operator.

yaisog:
> It would be nice to add some tests for the new suffix to the format operator.

Honestly, I have no idea how the automated testing works. If you have some instructions, I might be able to whip something up, maybe...

btheado:
If you have node installed, then this command from the root of the git clone will run the tests: `node tiddlywiki.js editions/test --test`

`editions/test/tiddlers/tests/test-filters.js` has some tests for `format:titlelist` and `editions/test/tiddlers/tests/test-json-filters.js` has tests for `format:json`. Maybe you can figure out from those examples how to do it?

yaisog:
Hi @btheado, thanks for the pointers. I added a test to `test-filters.js`. Since I don't have a node.js setup, I'm letting vercel run these tests for me.

@Jermolene: I'm guessing it's too late to be included in 5.2.6?

saqimtiaz:
@yaisog could be so kind as to also add tests for:
- a valid input timestamp other than 0
- an invalid input title that is passed through

yaisog:
Hi @saqimtiaz, done and done. 

I noticed that `format:date[]` removes invalid input instead of passing it through. Is this intended?

editions/test/tiddlers/tests/test-filters.js
```diff
@@ -1060,6 +1060,9 @@ Tests the filtering mechanism.
 			expect(wiki.filterTiddlers("[[Hello There]] [[GettingStarted]] +[format:titlelist[]]").join(" ")).toBe("[[Hello There]] GettingStarted");
 			expect(wiki.filterTiddlers("[title[Hello There]] +[format:titlelist[]]").join(" ")).toBe("[[Hello There]]");
 			expect(wiki.filterTiddlers("[title[HelloThere]] +[format:titlelist[]]").join(" ")).toBe("HelloThere");		
+			expect(wiki.filterTiddlers("0 +[format:timestamp[]]").join(",")).toBe("19700101000000000");
+			expect(wiki.filterTiddlers("1603188514443 +[format:timestamp[]]").join(",")).toBe("20201020100834443");
```
pmario, line 1064: Something like this would be nice too `"1603188514443" +[format:timestamp[YYYY-0MM-0DD 0hh:0mm:0ss.0XXX]]`

saqimtiaz:
> I noticed that format:date[] removes invalid input instead of passing it through. Is this intended?

It is. This would be good area in which to establish some norms, i.e. what should the default behaviour of operators be when given invalid input. Thoughts @Jermolene ?

We should also consider what the current behaviour of other operators is in this regard.

pmario:
>I noticed that format:date[] removes invalid input instead of passing it through. Is this intended?

Could you be more specific. I think it works as intended passing through stuff that it doesn't know

image

editions/test/tiddlers/tests/test-filters.js
yaisog: I'm not sure that any formats without "[UTC]" in their definition should be used for these checks, as the results will be time-zone dependent...
Putting "[UTC]" in the date string requires an external variable, which I rather not bothered with.

Jermolene:
> It is. This would be good area in which to establish some norms, i.e. what should the default behaviour of operators be when given invalid input. Thoughts @Jermolene ?

This is a tricky one. As you might expect, different operators currently behave differently:

* Arithmetic operators interpret invalid numbers as the value zero
* JSON operators interpret invalid JSON as a single JSON literal string with the same content
* The various sub-operators of the format operator all drop invalid input

At this point, I think we will have to consider each case on a case-by-case basis.

In this particular case, it would be consistent for format:timestamp to drop invalid input.

yaisog:
> Could you be more specific.

I'm not talking about unknown tokens. To them this applies: "Note that other text is passed through unchanged, allowing commas, colons or other separators to be used." I'm talking about invalid inputs:

image

yaisog:
> In this particular case, it would be consistent for format:timestamp to drop invalid input.

Of course. 😭 
I'll make the necessary changes post-haste.

yaisog:
OK, done. I also added a sentence about the dropped input to the `format` docs.

Would anyone like to see anything else while I'm at it?

pmario:
@yaisog ... Is this PR ready for review again?

yaisog:
> @yaisog ... Is this PR ready for review again?

Yes, this should be good to go (into review).

Jermolene:
Thanks @yaisog

PR #7293, opened 2023-02-26
Make some cosmetic changes to the DateFormat and Date Fields documentation

yaisog:
First of all, I inserted the format template for wiki fields ([UTC]YYYY0MM0DD0hh0mm0ss0XXX) into both doc tiddlers. This information was way too hard to find before, measured by its relevance.

I did remove the resulting duplication from the TIMESTAMP example and made the explanation for the value 86,400,000 clearer.

Worth discussing is maybe the change from inline code display to the `.value` and `.field` macros for field names and (some) values. Especially the TIMESTAMP heading was just screaming at readers (uppercase red text). However, there is now an inconsistency between the listing of tokens in the table (code) and in some other places (macro). Since I hate inconsistencies as much as the next guy, I'll gladly make changes once we agree on a direction.


Jermolene:
Thanks @yaisog – as this is a docs only change it should target the "tiddlywiki-com" branch which will mean that it will be included in tiddlywiki.com as soon as it is merged, without waiting for the next release.

yaisog:
Just changing the base pulled in a whole bunch of commits, which is probably not right. I'll close this PR and try to make a new one...

Jermolene:
Thanks @yaisog

PR #7296, opened 2023-02-26
Removes datauri triggered save instructions modal

btheado:
At startup if the code in rootwidget.js detected a data uri in the browser location url, then a modal dialog box containing save instructions was displayed. Based on the discussion at #7294, these instructions were for browsers which did not support the download attribute on the anchor tag.

According to https://caniuse.com/download, all supported browsers have had the feature since at least 2019, so triggering of the modal has been removed.


btheado:
I tested chrome and firefox on Linux with and without data uri before and after. After, the modal is gone and I don't see any issues.

I also grepped all the files in the git repo for references to the `SaveInstructions.tid` tiddler and didn't find any.

Jermolene:
Great, thanks @btheado

yaisog:
One less modal that needs to be called from the $root widget... 👍

PR #7297, opened 2023-02-27
Use sticky flag to improve regexp search performance

Jermolene:
The internal function `$tw.utils.parseTokenRegExp()` is used to check whether a regexp matches at a specified position in a string.

Using a regexp with the global flag results in the search being performed for the entire remainder of the string, not just at the position specified.

It appears that using the new Y (sticky) flag allows us to restrict the matching to the specified position, avoiding searching the rest of the string.

### DIFF
--- a/core/modules/parsers/parseutils.js
+++ b/core/modules/parsers/parseutils.js
@@ -84,7 +84,8 @@ exports.parseTokenString = function(source,pos,token) {
 };
 
 /*
-Look for a token matching a regex. Returns null if not found, otherwise returns {type: "regexp", match:, start:, end:,}
+Look for a token matching a regex at a specified position. Returns null if not found, otherwise returns {type: "regexp", match:, start:, end:,}
+Use the "Y" (sticky) flag to avoid searching the entire rest of the string
 */
 exports.parseTokenRegExp = function(source,pos,reToken) {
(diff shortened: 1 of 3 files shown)


linonetwo:
I can confirm it is much faster!

截屏2023-02-27 22 04 42

[7297.json.zip](https://github.com/Jermolene/TiddlyWiki5/files/10840381/7297.json.zip)

Jermolene:
Thanks @linonetwo I'm going to go ahead and merge this so that we get some more people testing it.

pmario:
@linonetwo ... The performance profile seems to be a completely different one. It contains a lot e-chart stuff which imo should be kept out from startup at all. .. **I'm out** IMO the wiki startup UI is littered with a lot of plugins that I personally would never include because of js code size. ... Sorry.

linonetwo:
Never mind. I use the local nodejs version so size is not a problem.

Jermolene:
@cdruan points out in #7350 that the new regexp sticky flag is not supported by the dialect of JavaScript that TW5 targets. In response, I've made a ticket about moving to a new dialect in #7353.

But I am afraid that if we are going to be consistent about backwards compatibility, then we're going to have to revert this PR for the moment.

PR #7299, opened 2023-02-27
fix: use `isShadowTiddler` insteadof `allTitles`

linonetwo:
About https://github.com/Jermolene/TiddlyWiki5/discussions/7278#discussioncomment-5076129 @pmario @Jermolene 

Before

截屏2023-02-27 22 39 24

After

截屏2023-02-27 22 43 31

### DIFF
--- a/core/modules/indexers/field-indexer.js
+++ b/core/modules/indexers/field-indexer.js
@@ -32,12 +32,14 @@ FieldIndexer.prototype.setMaxIndexedValueLength = function(length) {
 
 FieldIndexer.prototype.addIndexMethods = function() {
 	var self = this;
+	// get all tiddlers, including those overwrite shadow tiddlers
 	this.wiki.each.byField = function(name,value) {
 		var lookup = self.lookup(name,value);
 		return lookup && lookup.filter(function(title) {
-			return !self.wiki.isShadowTiddler(title)
+			return self.wiki.tiddlerExists(title)
 		});
 	};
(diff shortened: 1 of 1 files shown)


linonetwo:
And this method is checking isShadow like before

截屏2023-02-27 22 48 01

截屏2023-02-27 22 47 52

Jermolene:
Hi @linonetwo I don't think this works properly because of the possibility that a tiddler is both a shadow tiddler and an ordinary tiddler (this can happen if the shadow is overridden). In these cases, isShadowTiddler() returns true, but there can still be an ordinary tiddler overriding it.

linonetwo:
Thanks, I use `tiddlerExists` instead, which is 

> Test for the existence of a tiddler (excludes shadow tiddlers)

Jermolene:
Great, thanks @linonetwo I think this a worthwhile improvement.

pmario:
@Jermolene .. it seems we use `tiddlers.indexOf(title)` 3 more times. 

IMO we should check, if they can optimized too. 

image


`indexof(...== -1` .. is used 136 times. Should we have a closer look there too?

Jermolene:
None of those matches are relevant here. The key thing was that the code was searching the array returned by `wiki.allTitles()` instead of using `wiki.tiddlerExists()`.

pmario:
Yea. I've just seen that too. ... :)

PR #7301, opened 2023-02-27
Update ActionListopsWidget documentation including notes on de-duplication

### LINKED ISSUE #7266: [BUG] $action-listops de-duplicates the output of its $filter attribute
### Describe the bug

Filter output is de-duplicated before being written to the target field.

### Expected behavior

No de-duplication.

### To Reproduce

1. Go to tiddlywiki.com
2. Create a tiddler with the text `<$button actions="""<$action-listops $tiddler="$:/temp/listops" $filter="[[1]] =[[1]]" />""">Click</$button>`
3. Click the button
4. Look at the `list` field of `$:/temp/listops`

### Screenshots

_No response_

### TiddlyWiki Configuration

5.2.5

### Additional context

_No response_

yaisog:
De-duplication happens automatically when working with `$action-listops` on the `list` field.

Also, the input to the `$subfilter` of `$action-listops` is de-duplicated without notice, so `$subfilter` cannot be used when the list possibly contains duplicates, and one currently has to resort to `$filter="[enlist:raw[...]"` to enlist the current list.

Corresponding notes have been added to the doc. I also streamlined it a bit and moved the table of listops filters to the "Extended Listops Filters" tiddler.

Two points worth discussing:

1. I renamed "The Extended Listops Filters" to "Extended Listops Filters" and updated all links to it. This might break links from outside, though, so I'll probably have to go back on that.
2. I would like to move the example section for ActionListopsWidget to its own tiddler, like for most filter operators (but no widgets). Would this be OK?

Closes #7266

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
@@ -58,11 +58,11 @@ is mostly equivalent to using `$subfilter` along with "tags" for the value of `$
 
 !! Comparison to [[ActionSetFieldWidget]]
 
-In general, [[ActionSetFieldWidget]] is better for setting multiple fields at once and for replacing the value of a field, which can also be a list. The ActionListopsWidget is better for modifying a list field based on the existing list and for using a [[Filter Expression]] to derive the value of the field.
+In general, ActionSetFieldWidget is better for setting multiple fields at once and for replacing the value of a field, which can also be a list. The ActionListopsWidget is better for modifying a list field based on the existing list and for using a [[Filter Expression]] to derive the value of the field.
 
-The ActionSetFieldWidget sets the value of a field using `$field` and `$value` attribute pairs or attributes that do not start with a `$`. A single ActionSetFieldWidget can be used to set any number of fields of a single tiddler.
+The ~ActionSetFieldWidget sets the value of a field using `$field` and `$value` attribute pairs or attributes that do not start with a `$`. A single ~ActionSetFieldWidget can be used to set any number of fields of a single tiddler.
 
-The ActionListopsWidget replaces or modifies a single field's value using filter expressions.
+The ~ActionListopsWidget replaces or modifies a single field's value using filter expressions.
 
 The following widgets are functionally equivalent:
 
--- a/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
@@ -1,7 +1,7 @@
 caption: action-listops
 created: 20141025120850184
 list: efg hlm pqr
-modified: 20211115091121133
+modified: 20230227170803513
 myfield: 
 revision: 0
 tags: ActionWidgets Widgets
@@ -36,7 +36,7 @@ The ''action-listops'' widget is invisible. Any content within it is ignored.
 |$subfilter |An optional subfilter expression, which takes the list being manipulated as input, and saves the modified list back to the field/index being manipulated |
 |$tags |An optional subfilter expression, which takes the 'tags' field of the target tiddler as input, and saves the modified list of tags back to the 'tags' field |
 
-! $filter vs. $subfilter
+!! Using $filter or $subfilter
 
 Standalone use of the `$subfilter` attribute can be replaced by using a (more complicated) `$filter` attribute value.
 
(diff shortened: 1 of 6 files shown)


Jermolene:
Thanks @yaisog 
> 1. I renamed "The Extended Listops Filters" to "Extended Listops Filters" and updated all links to it. This might break links from outside, though, so I'll probably have to go back on that.

All we need to do is recreate the tiddler "The Extended Listops Filters" with the text `See [[Extended Lisops Filters]]`. It shouldn't have any of the tags that were originally applied.

> 2. I would like to move the example section for ActionListopsWidget to its own tiddler, like for most filter operators (but no widgets). Would this be OK?

Great, please go ahead.

yaisog:
> All we need to do is recreate the tiddler "The Extended Listops Filters" with the text `See [[Extended Listops Filters]]`.

Are you sure you want this? I'm not sure if it's worse having this shell tiddler or using the uncharacteristically named "_The_ Extended Listops Filter" tiddler... Also, we could set the text to `{{Extended Listops Filters}}` and previous links would remain working without additional clicks.

Jermolene:
> Are you sure you want this, @Jermolene?

It's how I've been handling this problem for a while (particularly in #6666 where there is a lot of documentation rejigging). In the future we may replace these stubs with a proper redirection mechanism.

> I'm not sure if it's worse having this shell tiddler or using the uncharacteristically named "_The_ Extended Listops Filter" tiddler... Also, we could set the text to `{{Extended Listops Filters}}` and previous links would remain working without additional clicks.

If we transclude the content then it is not so easy for users to tell that they've ended up at a deprecated title.

yaisog:
I left the "Examples" link as H1 heading. Is that too much?

yaisog:
> In the future we may replace these stubs with a proper redirection mechanism.

Hi @Jermolene, an intermediate solution that would probably not scorch the earth for future mechanisms could be to have a dictionary tiddler (!) with old:new pairs, and to add a data lookup to `$:/core/ui/ViewTemplate/body/default`. On success, give a "Please update your bookmarks to (new)" message and on fail, show the "Missing Tiddler" hint. The advantage would be that a single system tiddler could replace the clutter of ghost tiddlers. The data tiddler content could later be used for the "proper redirection mechanism".

But I'm sure you already have your own ideas on how to go about this...

Jermolene:
> But I'm sure you already have your own ideas on how to go about this...

No, that's very close to what I meant. The further step might be to auto-redirect navigation attempts, but I'm not sure that that's worth it.

yaisog:
> The further step might be to auto-redirect navigation attempts

The major hurdle being the timeout and an automatic launch of actions when such a tiddler is opened?

yaisog:
Also, I keep saying "data tiddlers" and I'm not hearing "almost deprecated". What gives? 😉

Jermolene:
> Also, I keep saying "data tiddlers" and I'm not hearing "almost deprecated". What gives? 😉

It's more "planned for deprecation", we're not yet at the point where all the use cases are covered by alternative implementations. The data representation part of dictionary tiddlers isn't controversial; it's just an alternate form of JSON tiddler, which are not going anywhere. The part that most needs deprecating is the `title##index` nomenclature. It's OK for some simple cases where the index is known in advance, but it's not sufficiently general to do very much.

yaisog:
I feel this is the wrong place for this topic, but since we're already on it:
I cannot find a filter operator that does something like this
```
[‹index name›]operator[‹dictionary title›]]
```
i.e. that takes an index and looks it up in the tiddler that is given as parameter to the operator? I can only find awkward constructs with the `:map` filter prefix or setting variables to do something similar. Did I overlook it?
I feel that for dictionary tiddlers to be used as such, something like this would be necessary.

saqimtiaz:
> I feel this is the wrong place for this topic, but since we're already on it, let's power through:

It would actually be highly appreciated if discussions in Github Issues and PRs are kept on topic. It makes it incredibly hard in the long run to review proposed core changes when the conversation digresses, and can often lead to the issue or PR in question stalling and not getting merged.

Could you be so kind as to post your question on talk.tiddlywiki.org instead?

Jermolene:
Thanks @yaisog

editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
```diff
@@ -49,12 +35,12 @@ For example, the items "abc" and "123" can be appended to the field `myfield` us
 The same can be achieved using the `$filter` attribute and prepending the [[Filter Run]] `[all[current]get[myfield]enlist-input[]]` to the [[Filter Expression]]:
```
btheado, line 35: Sorry for the late comment, but this line should also have been changed to use `[enlist{!!myfield}]` so it matches the change you made in the next line.

Jermolene: Another PR would be welcome, of course

yaisog: 🙃

PR #7303, opened 2023-02-28
Change separators for "words" and "some" suffixes of the search operator

### LINKED ISSUE #7263: [BUG] "words" and "some" suffixes for the search operator do not work as documented
### Describe the bug

Docs say tokens in the search string are separated by whitespace, but the code only looks for spaces.
I believe the relevant code is in `$:/core/modules/wiki.js`:
https://github.com/Jermolene/TiddlyWiki5/blob/95dc56d8508109a76c232cfbcd858889d32916a9/core/modules/wiki.js#L1259
https://github.com/Jermolene/TiddlyWiki5/blob/95dc56d8508109a76c232cfbcd858889d32916a9/core/modules/wiki.js#L1270

### Expected behavior

Any whitespace character (except maybe with the exception of the non-breaking space) should work as a token separator.

### To Reproduce

_No response_

### Screenshots

_No response_

### TiddlyWiki Configuration

5.2.6-prerelease

### Additional context

_No response_

yaisog:
The doc for the `words` and `some` suffixes says that the operator "treats the search string as a list of tokens separated by whitespace". The current code only treats spaces as separators. This PR will change that to use any whitespace character except the non-breaking space.

This change is not strictly backwards compatible.

Also, the separator used in Dynannotate for `words` and `some` is `\s`. Whatever is eventually decided here should also be implemented in #7260.

Closes #7263

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -1256,7 +1256,7 @@ exports.search = function(text,options) {
 			console.log("Regexp error parsing /(" + text + ")/" + flags + ": ",e);
 		}
 	} else if(options.some) {
-		terms = text.trim().split(/ +/);
+		terms = text.trim().split(/[^\S\xA0]+/);
 		if(terms.length === 1 && terms[0] === "") {
 			searchTermsRegExps = null;
 		} else {
@@ -1267,7 +1267,7 @@ exports.search = function(text,options) {
 			searchTermsRegExps.push(new RegExp("(" + regExpStr + ")",flags));
 		}
 	} else { // default: words
-		terms = text.split(/ +/);
+		terms = text.split(/[^\S\xA0]+/);
 		if(terms.length === 1 && terms[0] === "") {
 			searchTermsRegExps = null;
 		} else {


pmario:
So we replace `/ +/` 

which means

```
" " +       // Match the character “ ” literally
   "+"         // Between one and unlimited times, as many times as possible, giving back as needed (greedy)
```

with `[^\S\xA0]+`

```
"[^" +         // Match any single character NOT present in the list below
   "\\S" +         // Any character that is NOT a “whitespace character” (any Unicode separator, tab, line feed, carriage return, vertical tab, form feed)
   "\\xA0" +       // The character with position 0xA0 (160 decimal) in the character set
"]" +
   "+"            // Between one and unlimited times, as many times as possible, giving back as needed (greedy)
```

I'm OK with the change if we are aware what it means and if we really want that.

saqimtiaz:
It needs to be pointed out here that the change is not strictly backwards compatible and this should be pointed out as such in the PR so that an appropriate intentional decision can be made regarding the backwards compatibility implications.

yaisog:
> It needs to be pointed out here that the change is not strictly backwards compatible and this should be pointed out as such in the PR

Done.

pmario:
@kookma Can you try the Vercel preview version with your "section editor" plugin? .. Does it use the search operator? .. If not forget about it.

Jermolene:
Thank you @yaisog and apologies for the delay – I think this got left behind in the push to release v5.2.7

PR #7306, opened 2023-02-28
Add a comparison table when existing tiddler is about to be imported

### LINKED ISSUE #7211: [IDEA] Show last modified date and tiddler size when tiddler already exists on import
If one has multiple wikis and regularly copies some tiddlers from one to the other, it's hard to keep track of which is the newest version.
`Import` only notifies that a tiddler already exists.

Every file copy tool I know of will give the user some more information in that situation; at least the last modified date and object size, sometimes even a content comparison. That would be a good way to go for TW, too. I can imagine using a button pop-up or accordion thing to show this information only on demand and to not take up too much space in the UI. Personally, I'd be content with `modified` in a human-readably format and tiddler size in bytes (or maybe lines?).

This could also be useful in the upgrade UI where people often get confused by the various messages. A date comparison could go a long way to clear that confusion.

yaisog:
This PR adds a table that compares the date and size (in lines) of the existing and incoming tiddler in those cases where an incoming tiddler already exists. It also gives a verbal indication which tiddler is older, newer, smaller, larger.

If the incoming tiddler is older, a red border is additionally drawn.

Closes #7211

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -7,13 +7,6 @@ title: $:/core/ui/ImportListing
 \define payloadTitleFilter() [<currentTiddler>get<renameField>minlength[1]else<payloadTiddler>]
 
 \define overWriteWarning()
-\whitespace trim
-<$list filter="[<currentTiddler>!has<suppressedField>]">
-<$text text={{{[subfilter<payloadTitleFilter>!is[tiddler]then[]] ~[<lingo-base>addsuffix[Listing/Rename/OverwriteWarning]get[text]]}}}/>
-</$list>
-\end
-
-\define overWriteWarning()
 \define date-format() YYYY-0MM-0DD 0hh:0mm:0ss
 <$list filter="[<currentTiddler>!has<suppressedField>]">
 <$list filter="[subfilter<payloadTitleFilter>is[tiddler]]" variable="void">
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -13,6 +13,34 @@ title: $:/core/ui/ImportListing
 </$list>
 \end
 
+\define overWriteWarning()
+\define date-format() YYYY-0MM-0DD 0hh:0mm:0ss
+<$list filter="[<currentTiddler>!has<suppressedField>]">
+<$list filter="[subfilter<payloadTitleFilter>is[tiddler]]" variable="void">
+<$text text={{{ [<lingo-base>addsuffix[Listing/Rename/OverwriteWarning]get[text]] }}}/>
+<$set name="incomingDate" tiddler=<<currentTiddler>> subtiddler=<<payloadTiddler>> field="modified">
+<$set name="incomingText" tiddler=<<currentTiddler>> subtiddler=<<payloadTiddler>> field="text">
+<$let incomingLines={{{ [<incomingText>splitregexp[\n]count[]] }}}
+			currentLines={{{ [<payloadTiddler>get[text]splitregexp[\n]count[]] }}}
+			currentDate={{{ [<payloadTiddler>get[modified]] }}}
+			incomingDescription={{{ [<incomingDate>compare:date:gt<currentDate>then[newer]] [<incomingLines>compare:number:gt<currentLines>then[larger]] +[!is[blank]join[, ]addprefix[(]addsuffix[)]] }}}
+			currentDescription={{{ [<incomingDate>compare:date:lt<currentDate>then[newer]] [<incomingLines>compare:number:lt<currentLines>then[larger]] +[!is[blank]join[, ]addprefix[(]addsuffix[)]] }}}
+			>
+
+<div class={{{ [<incomingDate>compare:date:lt<currentDate>then[tc-import-older]] }}}>
+
+|tc-center tc-import-comparison |k
+|!existing <$text text=<<currentDescription>> />|!incoming <$text text=<<incomingDescription>> />|
+| <$text text={{{ [<currentDate>format:date<date-format>] }}}/> | <$text text={{{ [<incomingDate>format:date<date-format>] }}}/> |
+| <$text text=<<currentLines>> /> lines | <$text text=<<incomingLines>> /> lines |
+</div>
+</$let>
+</$set>
+</$set>
+</$list>
+</$list>
+\end
+
 \define selectionInfo()
 \whitespace trim
 <$set name="escUnselected" value={{{[{$:/language/Import/Upgrader/Tiddler/Unselected}escaperegexp[]addprefix[(?g)]]}}}>
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -2357,6 +2357,24 @@ html body.tc-body.tc-single-tiddler-window {
 	background: <<colour diff-delete-background>>50;
 }
 
+.tc-import-table .tc-import-comparison {
+	min-width: 80%;
+	max-width: 100%;
+	box-sizing: content-box;
+	padding: 0 1em;
+}
+
+.tc-import-table .tc-import-comparison th {
+	width: 50%;
+	padding: 0 1em;
+	background: none;
+}
+
+.tc-import-older table.tc-import-comparison {
+	border: 2px solid red;
+}
+
+
 /*
 ** Alerts
 */


saqimtiaz:
I would strongly suggest that we explicitly solicit end user feedback for these changes over the course of a release cycle before making any changes.

We have previously worked on making the import related messages more concise in light of user feedback that they were too long and this was diminishing their usefulness. We should also consider whether redesigning the import UI is a better way to proceed rather than shoehorning more information into the already cluttered existing UI.

pmario:
hmmm. Now that I see it, I would like to make it much less intrusive and easier to manipulate. .. 

What if we create a tiddler eg: `$:/core/ui/ImportListing/Overwrite/xxxx` and transclude it into the overWriteWarning() macro using a list 

```
<$list filter="[subfilter<payloadTitleFilter>is[tiddler]]" variable="void">
	<$list filter="[all[shadows+tiddlers]tag[$:/tags/ImportInfo]]">
		<$transclude mode=block/>
	</$list>
</$list>
```

image

So it will be possible, to add additional info to the overwrite text on demand.

pmario:
It's exactly the same, so the info is redundant and way to much. 

image

pmario:
So as Tony wrote. If there is a verbose detection it should be much more verbose. Listing exactly, what has changed. Otherwise it does not contain new info.

yaisog:
> So as Tony wrote. If there is a verbose detection it should be much more verbose. Listing exactly, what has changed. Otherwise it does not contain new info.

Tony's is a different use case. I'm trying to sync two wikis manually, by import. I'll be importing a bunch of tiddlers and need an at-a-glance check if I'm about to do something stupid. No? Fine. Go import. 

Sure, there may be cases when I'd like to know the exact differences, and there already exists a detailed import comparison tool.

Yes, the layout can be optimized, the CSS is just a rough draft. The columns could be drawn above-and-below for narrower screens. I obviously designed it on a larger screen, where it looks fine:
image

And also yes, if the files are identical, we can leave out the details. That's a great idea, even though it sounded a little like criticism.

I also like the idea of making it configurable and extensible via tagged tiddlers. Then, it should be shipped with at least the original (default) and verbose versions, and a dropdown should be added above the status column that lets users select their favorite choice (which is subsequently saved in a $:/config tiddler for permanence).

yaisog:
If the utility of this PR is not self-evident, I think I'm wasting my time trying to convince people.

Mario's idea should be another PR for another day.

Jermolene:
> If the utility of this PR is not self-evident, I think I'm wasting my time trying to convince people.
> 
> Mario's idea should be another PR for another day.

Perhaps it's worth having a discussion on talk.tiddlywiki.org about the options. There has been a fair amount of discussion about the underlying problem and so I'd still be interested in exploring whether there are some simple improvements.

yaisog:
I will post my code for those who need this enough to be willing to overwrite a core tiddler, like I did. In that context we can maybe get some more opinions. To me this functionality – akin to probably every file manager out there – is so basic that I don't really see the point of discussing this further.

Jermolene:
> I will post my code for those who need this enough to be willing to overwrite a core tiddler, like I did. In that context we can maybe get some more opinions. To me this functionality – akin to probably every file manager out there – is so basic that I don't really see the point of discussing this further.

The issue here is the perennial one that we want to ensure that any updates to the core are the best we can do. Sometimes that does risk making "perfect be the enemy of good", of course.

saqimtiaz:
> Yes, the layout can be optimized, the CSS is just a rough draft. Let's not get too hung up on that. At this point it is more about function than about form. The columns could be drawn above-and-below for narrower screens. I obviously designed it on a larger screen, where it looks fine:

If this indeed an early draft of the implementation, this should have been created as a draft PR with a note that is it meant to facilitate discussion and it would then have been reviewed differently. PRs for core changes should be self-sufficient, that is there should not be an intent that a second PR will be required to fix a first implementation to address usability issues that are already evident.

PRs are for concrete implementation that the author considers appropriate to be merged into the core as is, unless otherwise specified, and are reviewed as such.



> To me this functionality – akin to probably every file manager out there – is so basic that I don't really see the point of discussing this further.

I find this a very counter productive way of looking at things. No one has questioned the underlying need, or the proposed idea for improved functionality, however the implementation definitely needs discussion, user feedback and further improvement before it can be considered for the core.

yaisog:
> this should have been created as a draft PR

I am sorry, but I am not yet as well versed with the intricacies of GitHub as I maybe should be. The draft status mainly concerns the CSS. I'm afraid if finely polished CSS is required to submit a PR, you will see no further ones from me.

> No one has questioned the underlying need [...]

I am again sorry, but I experienced this differently. I felt nothing but blowback from the minute I opened issue #7211, also and especially pertaining to the fundamental necessity of the proposed feature. This was draining too much of my time and energy, which eventually led me to close this PR.

pmario:
>If the utility of this PR is not self-evident, I think I'm wasting my time trying to convince people.
>
>Mario's idea should be another PR for another day.

I'm not against it. Just seeing it implemented I don't like it anymore. At the other thread #7211 I thought, that we would need additional help from JS code. So I did no explore it further. ... Your experiments / PR here shows that there is much more usable info in the import tiddler as I expected.

My experiments with https://talk.tiddlywiki.org/t/howto-compare-the-content-of-a-shadow-tiddler-with-the-content-of-an-overwritten-shadow/6191 also helped me to get a better understand of the possibilities diff-match-patch gives us. 

Having a look at the new images taken by Jeremy and myself, I just think that if there is new visual overhead it needs to provide a lot more new information. 

1. We do have the info that "A tiddler with that title already exists"
2. Then there is a table with almost the same info with the date
3. Then there is some more info "1 lines" that tells us again that there is a difference of 1 line ... But nothing about the real differences. 

IMO that should be improved in the way that #3 has the biggest potential to contain useful info, which should be easy to access for the user.

image

---------

>I think I'm wasting my time trying to convince people.

I can see your frustration, but I can ensure you, you are _not_ wasting time. In the contrary, it's inspiring that you keep pushing. 

I think this type of experiments are needed to get things going and even if this PR is not merged, I think there is a high chance, that in the end we will have a good solution, we all will be fine with. 

just some thoughts.

pmario:
> yaisog [wrote](https://github.com/Jermolene/TiddlyWiki5/pull/7306#issuecomment-1449661721)
>
>I will post my code for those who need this enough to be willing to overwrite a core tiddler, like I did. In that context we can maybe get some more opinions. 

Users do not like to overwrite core tiddlers, because that's usually problematic with updating. .. That's the whole reason why we did create the "cascade" filter stuff. 

With my suggestions posted at: https://github.com/Jermolene/TiddlyWiki5/pull/7306#issuecomment-1448599450 I think it would be possible to keep the things backwards compatible and also have the possibility to create experiments and plugins that can show more info. Even overwriting the core 1 line message, if needed. 

As I wrote in the light of your extension this 1-liner imo is redundant and having the possibility to remove it, will reduce visual complexity.

saqimtiaz:
> I'm afraid if finely polished CSS is required to submit a PR, you will see no further ones from me.

From my perspective a PR needs to be polished enough in all aspects before it can be merged, so that a holistic judgement can be made as to whether it is a good implementation. That does not however mean that the original PR author need do it entirely by themselves. Through discussion and collaboration others can contribute and help get a PR ready for merging, though this requires a certain degree of patience.

pmario:
I did a bit more experiments. I do find something like this much more convenient, because the info is easier to understand. 
At least for me. 

image


I'll create a PR for this one. It will _not_ contain the verbose elements. Just the backwards compatible but more hackable stuff.

Jermolene:
> I am again sorry, but I experienced this differently. I felt nothing but blowback from the minute I opened issue https://github.com/Jermolene/TiddlyWiki5/issues/7211, also and especially pertaining to the fundamental necessity of the proposed feature. This was draining too much of my time and energy, which eventually led me to close this PR.

I am sad to hear that. Reading the thread back, perhaps my terse comment implied disapproval too, which was not my intent.

As @saqimtiaz points out, it can be frustrating dealing with incomplete PRs ([here's an example of me talking about it](https://github.com/Jermolene/TiddlyWiki5/discussions/5484)), but that has to be balanced with our desire to help all contributors (whether newcomers or not) to navigate the complexities of the process. Our aim is for the community to be supportive and understanding.

Happily, you've already shown considerable success in getting your work into the core, and I hope you'll be able to continue to do so.

pmario:
I did one more experiment. ... It clearly shows that there is still room for improvement. For me it's extremely important, that I can import state- and temp-tiddlers. ... 

But as shown in the image below with a realistic import tiddler with existing, blocked system, blocked plugin and a "user selected" state tiddler. .. There is a problem with the $:/temp/info-plugin .. It does not have a created or modified info. ... 

As you can see, there is a lot going on. Especially with redundant info that only makes interpreting the info more difficult.

There is only 1 info box, that actually contains "relevant" info, where the existing is "newer" than importing. IMO only this info should stand out. The rest should _not_ be there at all

image

just more thoughts.

Issue #7279, opened 2023-02-22
Planning for the release of v5.2.6 and v5.3.0


Jermolene:
The following roadmap is proposed for the upcoming v5.2.6 release and the following v5.3.0 release:

* 6th March 2023 : final date for merging  pull requests for v5.2.6 
* 20th March 2023: release of v5.2.6
* (20th - 27th March 2023: window for release of v5.2.7 if required due to bugs in v5.2.6)
* 27th March 2023: merge #6666 into prerelease of v5.3.0
* 15th May 2023: final date for merging pull requests for v5.3.0
* 29th May 2023: release of v5.3.0

Jermolene:
Due to #7371 we are going ahead with a v5.2.7 bug fix release on Sunday 26th March. See #7372 for the planning discussion.

pmario:
Related:  Tickets to be reviewed/merged after release of v5.2.6 #7345

Jermolene:
I've updated the roadmap to reflect the delay caused by the v5.2.7 release.

Jermolene:
I've updated the roadmap now that #6666 has been merged.

rmunn:
I'd like to get https://github.com/Jermolene/TiddlyWiki5/pull/7448 reviewed so that the checkbox listField bug can be fixed before 5.3.0. See https://github.com/Jermolene/TiddlyWiki5/issues/6871 for the details of the bug; #7448 fixes the last outstanding issue, where the `created` and `modified` fields could be messed up if you pass them into a checkbox widget's `listField` parameter.

pmario:
@Jermolene .. Since the "closing date for PRs" is near, imo the **Tickets to be reviewed/merged after release of v5.2.7
#7345** needs to be reviewed again. 

Especially, there is the PR: **Bidirectional text improvements #4541** which seems to be a big improvement, but is not yet merged. IMO @kookma should give a statement about that one.

pmario:
@Jermolene ... Does this plan still stand?

Jermolene:
I've updated the OP to change the PR freeze date to today, 24th June 2023, and to change the final release date from 28th June to 30th June.

Tagging @twMat because he was asking elsewhere about changes to the release timetable.

pmario:
Should we open a "Final Test Request for v5.3.0" at Talk, which I could globally pin for the next week?

pmario:
.. and I'll update the German texts today :)

Jermolene:
[...]

PR #7308, opened 2023-03-01
make import listing more configurable

pmario:
This PR is an alternative approach to closed PR #7306 
It does 2 things. 

- Create a new system tag: `$:/tags/ImportOverwrite`
- Moves the default text out from the overWriteWarning() macro to a new tiddler: `$:/core/ui/ImportListing/overwrite/default-text`

So users have the ability to extend the import message area with their own info by tagging their template with new system-tag.

This attachment contains a sample configuration to play with. This info is _not_ part of the PR. 
[import-listing-verbose.zip](https://github.com/Jermolene/TiddlyWiki5/files/10861321/import-listing-verbose.zip)

There should be no visual difference between this PR and the v5.2.5
You need to import the JSON above to see a difference. 

There is an import payload attachment, which creates a more realistic import dialogue with "state", "temp" and "blocked" tiddlers. 
[test-import.zip](https://github.com/Jermolene/TiddlyWiki5/files/10861344/test-import.zip)

@Jermolene if it can be merged, I'll add some docs about the new system-tag and how to use the new possibilities

### DIFF
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -10,7 +10,10 @@ title: $:/core/ui/ImportListing
 \whitespace trim
 <$list filter="[<currentTiddler>!has<suppressedField>]">
 	<$list filter="[subfilter<payloadTitleFilter>is[tiddler]]" variable="ignore">
-		<$list filter="[all[shadows+tiddlers]tag[$:/tags/ImportOverwrite]]" variable="overwrite">
+		<$list filter="[all[shadows+tiddlers]!is[draft]tag[$:/tags/ImportOverwrite]]" 
+			variable="overwrite"
+			emptyMessage={{$:/core/ui/ImportListing/overwrite/default-text}}
+		>
 			<$transclude tiddler=<<overwrite>> mode="block"/>
 		</$list>
 	</$list>
--- a/core/ui/ImportListing.tid
+++ b/core/ui/ImportListing.tid
@@ -9,7 +9,11 @@ title: $:/core/ui/ImportListing
 \define overWriteWarning()
 \whitespace trim
 <$list filter="[<currentTiddler>!has<suppressedField>]">
-<$text text={{{[subfilter<payloadTitleFilter>!is[tiddler]then[]] ~[<lingo-base>addsuffix[Listing/Rename/OverwriteWarning]get[text]]}}}/>
+	<$list filter="[subfilter<payloadTitleFilter>is[tiddler]]" variable="ignore">
+		<$list filter="[all[shadows+tiddlers]tag[$:/tags/ImportOverwrite]]" variable="overwrite">
+			<$transclude tiddler=<<overwrite>> mode="block"/>
+		</$list>
+	</$list>
 </$list>
 \end
 
--- a/core/ui/ImportListing/overwrite/default-text.tid
+++ b/core/ui/ImportListing/overwrite/default-text.tid
@@ -1,4 +1,3 @@
 title: $:/core/ui/ImportListing/overwrite/default-text
-tags: $:/tags/ImportOverwrite
 
 <$text text={{{ [<lingo-base>addsuffix[Listing/Rename/OverwriteWarning]get[text]] }}}/>
--- a/core/ui/ImportListing/overwrite/default-text.tid
+++ b/core/ui/ImportListing/overwrite/default-text.tid
@@ -1,0 +1,4 @@
+title: $:/core/ui/ImportListing/overwrite/default-text
+tags: $:/tags/ImportOverwrite
+
+<$text text={{{ [<lingo-base>addsuffix[Listing/Rename/OverwriteWarning]get[text]] }}}/>


pmario:
@yaisog .. You may have a look at it.

yaisog:
I like the configurability via the tag, as I mentioned in the other PR. This makes it modular and configurable concerning the information displayed.

Could you make it so that the original message goes into the `emptyMessage` of the `$list` instead of the the "default-text" tiddler? This way one could remove the default message without having to remove the tag from a system tiddler. I can imagine that some people find the message superfluous when they have the table. Adding the message to a configuration tiddler to get it back would be simple.
I'm also thinking of replacing the default message with a "A (newer|older) tiddler with this name already exists." if newer/older can be determined.

PS: You could also add `!is[draft]` to the list filter.

pmario:
>I'm also thinking of replacing the default message with a "A (newer|older) tiddler with this name already exists." if newer/older can be determined.

Be aware, that the default message is translatable at the moment. .. So if you can reuse translated text you should do that.

yaisog:
> Be aware, that the default message is translatable at the moment.

Yes, I noticed. While the verbal indication as part of the message might be useful, it would add two new language strings... 🤔

yaisog:
I have created an alternative overwrite message for those who can live with a little more verbosity than in Mario's version:
image

There are a number of checks to only show relevant information (and hide metadata which is identical for both tiddlers), and a separate message when tiddlers seem to be identical ("seem" because I can only check the `text` field but not all possible fields, as they cannot be accessed by filter operators, as far as I know). Also, tiddlers with no modified date are output as such instead of blank lines.

Personally, I find it helpful to add an extra highlight to critical tiddlers, i.e. those cases where something bad is likely to happen should the user just click "Import" without thinking twice.

The definition tiddlers are here: 
[import-listing-yaisog.zip](https://github.com/Jermolene/TiddlyWiki5/files/10874892/import-listing-yaisog.zip)

I am by no means suggesting that these should become part of the core or this PR. It is simply a demonstration of how this PR could be leveraged to make the import table a little more useful (and which would make a nice plugin).

pmario:
>I am by no means suggesting that these should become part of the core or this PR. It is simply a demonstration of how this PR could be leveraged to make the import table a little more useful (and which would make a nice plugin).

That was the reason for this PR. It should be possible now to make more experiments, without the need to directly include more verbose info into the core. So users who want more info can use plugins. 

In the past we had some user feedback that the import info is overwhelming. That's way we did reduce it to the bare minimum. 

This PR should make it more flexible again. .. And it's the users choice if they want to see more info.

pmario:
@Jermolene .. I would like to have this one in v5.2.6 .. Could you have a look?

saqimtiaz:
> I have created an alternative overwrite message for those who can live with a little more verbosity than in Mario's version:

It would be worth exploring an approach that allows use of the full table width as the preview mechanism does, to ensure a viable approach for smaller screen sizes.

In fact, it would be worth exploring the kinds of messages shown here as previews, with perhaps the facility to automatically show the preview for certain types of tiddlers/conflicts.

Jermolene:
Hi @pmario I'm inclined to leave this for the next release. I understand the underlying need, but I think I'd rather collect together all the outstanding import-related requests and make a plan.

pmario:
@Jermolene  **Please review again.** 

Quote from the OP
>This attachment contains a sample configuration to play with. This info is _not_ part of the PR. 
[import-listing-verbose.zip](https://github.com/Jermolene/TiddlyWiki5/files/10861321/import-listing-verbose.zip)
>
>There should be no visual difference between this PR and the v5.2.5
You need to import the JSON above to see a difference. 
>
There is an import payload attachment, which creates a more realistic import dialogue with "state", "temp" and "blocked" tiddlers. 
[test-import.zip](https://github.com/Jermolene/TiddlyWiki5/files/10861344/test-import.zip)

**TW Default Import**

image

**Custom Import Initial**

If import-listing-verbose.JSON is imported **and** the test-import.JSON is imported the import dialogue will look like this: 

image

**Custom Import - existing files** 

If all tiddlers are imported already and imported again

image

pmario:
Bump in the light of:  Planning for the release of v5.3.0 #7279

pmario:
@Jermolene Bump for review. It does not change the existing Import dialogue. It adds the possibility for plugin authors to modify it. 

**Description and ZIP files needed for test testing** are at: https://github.com/Jermolene/TiddlyWiki5/pull/7308#issuecomment-1537140800

saqimtiaz:
Hi @pmario, have you considered using the already extensible preview mechanism for the same purpose? For example a preview could be created that compares the metadata such as the modified date of tiddlers that would be overwritten by the import.

Not only is that an existing mechanism that could be reused, but it would also allow for the entire width of the table to be used to display that information rather than trying to fit it into that small last column.

The import mechanism is overdue for a complete rewrite in the near future and I would be hesitant to add further complexity and backwards compatibility constraints to it at this point, which might curtail our options for the rewrite.

pmario:
You are right. I'll rewrite it as a plugin. So close.

PR #7310, opened 2023-03-01
Update chinese language files

BramChen:
* Add chinese translation for the text of `Error/ZoominTextNode` in Misc.multids


Jermolene:
Thank you @BramChen

PR #7312, opened 2023-03-02
Serving via a mini web server

hffqyd:
Saving via a mini web server, and uploading files/images for use like [img[images_folder/example.jpg]].

### DIFF
--- a/editions/tw5.com/tiddlers/saving/Saving via a Minimal Web Server.tid
+++ b/editions/tw5.com/tiddlers/saving/Saving via a Minimal Web Server.tid
@@ -1,0 +1,43 @@
+caption: tw5-server
+color: #70c9a0
+community-author: hffqyd
+created: 20230302011710789
+delivery: Server-side Script
+description: Web server for saving and uploading
+method: save
+modified: 20230302055929311
+tags: Android Chrome Firefox [[Internet Explorer]] Linux Mac Opera Safari Saving Windows iOS Edge
+title: Saving via a Minimal Web Server
+type: text/vnd.tiddlywiki
+
+A local server for TiddlyWiki5 that saves and backups wikis, inspired by
+[[tw5-server.rb | https://gist.github.com/jimfoltz/ee791c1bdd30ce137bc23cce826096da]].
+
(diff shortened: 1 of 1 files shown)


Jermolene:
Many thanks @hffqyd for your contribution! With apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

pmario:
Is it possible to upload any file-type with this server? If yes, this is an extremely dangerous thing. 

I'm not a python guy, but I can't see any code, that filters file types. 
I'm also not sure, if it is possible to access files outside the initial serving directory. 

I do have security concerns.

hffqyd:
> Is it possible to upload any file-type with this server?

Yes ..., if uploaded executable files might cause damage, I'll consider filter out such files, if it's a big concern.

> if it is possible to access files outside the initial serving directory

No, as least in my tests.

Jermolene:
> Is it possible to upload any file-type with this server? If yes, this is an extremely dangerous thing.

Whether is dangerous to upload arbitrary file types depends on the context and threat models being considered. I think this concern is better raised separately, it's not really a blocker to merging this PR.

> I'm not a python guy, but I can't see any code, that filters file types. I'm also not sure, if it is possible to access files outside the initial serving directory.
> 
> I do have security concerns.

I also don't have enough Python to verify the operation of @hffqyd's code. But I think this is more of a policy issue: we can't do a security review of every third party saver/syncer that we list. I'm not sure the best way to address the concerns you've raised:

* Perhaps we should emphasise that third party savers have no warranty, and haven't been reviewed from a security perspective. But that seems overly broad: some third savers are surely safe to use, and we shouldn't scare users off
* Perhaps we should introduce a security checklist for third party savers as a way to encourage developers to consider the steps that they have taken. That's a lot of work, and might well just serve to scare users

As ever, security can really only be addressed by educating users. That's not entirely practical, so perhaps the best we can do is to clearly signpost the solutions that are generally recommended.

PR #7315, opened 2023-03-03
Saving via a Minimal Web Server

hffqyd:
A standalone mini web server enables support tiddlywiki as well as upload files/images, for use like `[img[images/uploaded.jpg]]`.


Jermolene:
Thanks @hffqyd – I may have missed an earlier post, but I think it would also be helpful if you could also make a post to https://talk.tiddlywiki.org/ introducing this new saver

Issue #7142, opened 2022-12-20
[BUG] difference in outline styles v5.2.3 vs v5.2.5


ericshulman:
### Describe the bug

I ran a diff on $:/themes/tiddlywiki/vanilla/base, and the following two rules that were in v5.2.3 are omitted from v5.2.5:
```
input[type="search"] { outline-offset: initial; }
button, textarea, input, select { outline-color: <<colour primary>>; }
```
* Omitting the first rule results in the sidebar search focus outline offset being inset by -2px from the input field border instead of matching the input border.
* Omitting the second rule results in focus outlines shown using the current foreground color (e.g. black) instead of using the primary color (e.g., blue=#5778d8 when using default Vanilla palette).

Note, however, that even after restoring these two rules, ***the second rule does not seem to work correctly when the selected palette specifies a different primary color.***  For example, if `$:/palettes/SolarizedDark` is selected, the value of `<<colour primary>>` is `#859900` -- a green-ish/yellow-ish color -- , but the outlines appear bright white!

Using the following alternative rule seems to correct this:
```
* :focus-visible { outline:2px solid <<colour primary>>; outline-offset:-2px; border-radius:0.25em; }
```
Note that, for inexplicable reasons, all of the above styles must be included in this alternative rule in order to achieve the desired effect.  Regrettably, my expertise with CSS (and especially with regard to the `outline` attributes) is not sufficient to be certain as to why this is the case.

I suggest restoring the first two rules in order to revert to the v5.2.3 appearance, and then experimenting further with the use of the alternative `* :focus-visible` rule to hopefully find the minimal set of CSS attributes to fix the outline color issue noted above.

### TiddlyWiki Configuration

- Version: TW v5.2.5
- OS: Windows10
- Browser: Chrome Version 108.0.5359.125 (Official Build) (64-bit)

Jermolene:
Thank you @ericshulman. I'll prepare a PR to revert those changes as you suggest. I've added this ticket to the list at #7100.

kookma:
Is the below difference also related to this issue?

Tested on: Chrome 108 / Edge 108 - Windows 10 

* TW 5.2.3
![791_msedge](https://user-images.githubusercontent.com/830394/209458532-d12bef32-2f3c-406d-9d6d-83ecc7008f5f.gif)


* TW 5.2.5
[...]

PR #7317, opened 2023-03-03
fix 7142 search outline in chrome like browsers

pmario:
This PR fixes: [BUG] difference in outline styles v5.2.3 vs v5.2.5 #7142

using the code snippet suggested at: https://github.com/Jermolene/TiddlyWiki5/issues/7142#issuecomment-1364632955

It needs to be tested with Chrome or Edge to see the difference between old and new. 

I did test it with 

- FF on Windows
- Edge on Widows 
- FF an Android

NEW in PR

image

OLD

image


pmario:
I did also change the "primary colour" in the palette-manager. It will be assigned now as requested.

Jermolene:
Hi @pmario I am not keen on overriding built-in browser styles. The versions of TiddlyWiki that we make today are not just targeting the browsers of today, but those for the next 20 years. This PR treats current browser styling as broken, and attempts to fix it by overriding a built-in style in ab opinionated way.

The "before" screenshot above doesn't look great, but I'd like to see what it looks like with browser native styling before we go any further.

pmario:
It is broken, because reset.css breaks it.
In line 262ff it says. 

```
/**
1. Correct the odd appearance in Chrome and Safari.
2. Correct the outline style in Safari.
*/

[type='search'] {
  -webkit-appearance: textfield; /* 1 */
  outline-offset: -2px; /* 2 */
}
```

which draws the outline inside the search text input with -2px offset. According to reset it is a fix for Safari, which I cannot test. 

base.css tries to fix it because everyone complains.

```
input[type="search"] {
	outline-offset: initial;
}
```

As Eric requested in #7142 ... I don't know if this breaks anything in Safari, but it seems to fix the problem in Chrome. There is no "ouline offset problem" in FireFox. 

BUT 

reset.css in line 220ff destroys outlines for FF. Everything has been fine until we did update to the latest reset - 3 years ago.

```
/**
Restore the focus styles unset by the previous rule.
*/

:-moz-focusring {
  outline: 1px dotted ButtonText;
}
```

While it is the browser default formatting it does not make any sense. 1px dotted is almost invisible, especially for butons and tabs if selected with the keyboard. 

-------

If it is OK to fix problematic formatting for Chrome and Safari, I think it should also be OK to fix it for FireFox users too.

pmario:
As Eric wrote at: #7142

>I ran a diff on $:/themes/tiddlywiki/vanilla/base, and the following two rules that were in v5.2.3 are omitted from v5.2.5:
>
>```
>input[type="search"] { outline-offset: initial; }
>button, textarea, input, select { outline-color: <<colour primary>>; }
>```

So they are back now. Those 2 rules fix the problem that the reset.css causes for Chrome and Edge users

The following rule fixes the problem the reset.css causes for FireFox users.

```
:-moz-focusring {
	outline: 2px solid <<colour primary>>;
	outline-offset: -2px;
}
```

Jermolene:
> It is broken, because reset.css breaks it. In line 262ff it says.

Great. Then we fix it by fixing reset.css, not by layering more CSS on top.

pmario:
>Great. Then we fix it by fixing reset.css, not by layering more CSS on top.

IMO that's problematic when we do an other update. ... I'm pretty sure we will miss that fix and will have the same problem.

Most of the definitions in reset.css are overwritten by the base.css 

The very first rules in reset are: 

```
*,
*::before,
*::after {
  box-sizing: border-box;
}
```

The first rules in base.css are:

```
{{$:/themes/tiddlywiki/vanilla/reset}}

*, input[type="search"] {
	box-sizing: border-box;
	-moz-box-sizing: border-box;
	-webkit-box-sizing: border-box;
}
```

That's the same thing .. So we can completely remove it. `box-sizing` is supported by [every browser](https://caniuse.com/?search=box-sizing)

There are probably many other elements that are overwritten by base.css

Jermolene:
Hi @pmario it sound like we should drop the CSS reset entirely. I think now it has become too opinionated to be useful.

pmario:
>Hi @pmario it sound like we should drop the CSS reset entirely. I think now it has become too opinionated to be useful.

I did think about that too, but if someone wants to start a new "base-like" CSS they should be able to start from something. .. It's a pain point, that we still need browser CSS normalisation. .. There is much [more cooperation these days](https://web.dev/interop-2022/) as it has been in the past. So the reset.css imo is much smaller as normalize.css was. 

The problem with the reset.css settings is, that nobody really knows why they are there. There are some comments, but imo it's way too much work to dig into the rabbit hole. 

I also think that the reset `box-sizing` is the right version and obvious duplicates should be removed from base.css 

```
*,
*::before,
*::after {
  box-sizing: border-box;
}
```

-------

I did just compare our reset.css with normalize.css .... They are basically the same, but normalize still has more IE10 fixes and is less opinionated about font settings. 

eg: reset.css defines  font-family for body, code, kbd, samp and pre ... which normalize does not. 
On the other hand reset.css does global box-sizing and normalize does it per element. .. But that's already about it. .. 

---------

According to the FireFox outline. The browser default is `outline: 1px dotted ButtonText` .. **but** I think that's the completely wrong choice. It's almost invisible and against every accessibility guidline. 

At: https://developer.mozilla.org/en-US/docs/Learn/Accessibility/CSS_and_JavaScript if you scroll down to links it shows an image like this one, where the outline is blue. and somewhat visible. 

image

Every example input at MDN even has 2 outlines: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input and they are light-blue and blue. ... 

I think it's the wrong way to go with the default FF setting even if it is browser default.

pmario:
I did have a closer look at every element in the reset.css file. ... It's not true, that most things are overwritten. .. It's just the font-family elements that I did have in mind. Because in TW we have user configurable fonts .... So they have to be overwritten. 

code, kbd, samp and pre  have font-family definitions in reset ... 
only code and pre are overwritten by base. ... So imo it's ok that the others are initialized in a sensible way. 

There are some initialisations for HR, ABBR, STRONG, B, SMALL, SUB, SUP ... and so on, where we do not have any overwrites in base. So we rely on those in reset. 

The rest are the "exotic ones" ... Where the `:-moz-focusring {outline: 1px dotted ButtonText;}` is one of them. Which I did explain in the other post. 

---------

base.css has a 26 `-webkit-` specific adjustments all over the places. ... I can not see, why those adjustments are no problem but `-moz-focusring` is one?

pmario:
@Jermolene -- I did merge master into this PR which fixes https://github.com/Jermolene/TiddlyWiki5/issues/7142 which is a "long standing bug now"

As I wrote, I cannot test it with Safari.  So it would be nice, if you could post 2 screenshots, so I can see it. 

**Broken**: Edge Windows 11 **v5.3.4** -- Outline and colour do not work

image

image

----

**Broken**: FireFox Windows 11 **v5.3.4**

image

image

---

**Fixed:** This is Edge Windows 11 **with PR**

image

image

----

**Fixed:** This is FF Windows 11 **with PR**

image

and

image

Jermolene:
Hi @pmario let me know if these aren't what you need:

image

image


For the second one, I had to turn on Safari's setting for making all buttons and links keyboard accessible:

image

pmario:
I think that looks OKish. IMO the "Search" button is a bit too close to the input. 
I think it should be a bit further away. 

Should I create a new PR for the search button spacing or should I add it to this PR?

Jermolene:
> I think that looks OKish. IMO the "Search" button is a bit too close to the input. I think it should be a bit further away.

Agreed on both points.

> Should I create a new PR for the search button spacing or should I add it to this PR?

I think it's reasonable to add it here.

pmario:
@Jermolene -- I did fix the distance between the search test input and the Advanced Search Button. So IMO this one can be merged now

Leilei332:
@Jermolene I think this one can be merged, so that the strange black outline of the search box on focus in Chrome-like browsers can be replaced by the original one with palette color.

Jermolene:
Thanks @pmario @Leilei332

Leilei332:
Looks like there still exists some problems. After I switched to a palette with `color-scheme` set to `dark` , the outline color became white.



It displayed the palette color after I added the following code to the stylesheet:

```css
* :focus-visible { outline:2px solid <<colour primary>>; outline-offset:-2px; border-radius:0.25em; }
```

PR #7320, opened 2023-03-03
improve Formatting in Wikitext

pmario:
This PR should improve the initial information about WikiText basic formatting documentation

- It shows examples
- It shows the toolbar buttons
- and it shows the keyboard shortcuts. 

The second part of the tiddler shows the created HTML structure, which may be important for advanced users.


yaisog:
Since `''text''` is rendered as `<strong>`, should we also refer to it as "strong importance". It is not the same as bold, see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/strong#b_vs._strong

The same goes for `<em>` and _italic_: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#i_vs._em

Mozilla calls `__text__` underline, not underscore.

Should we use this PR to clean up the terms? Since then the indications on the buttons would no longer match, we might put the "correct" terms in parentheses? Well, maybe it's enough to show the resulting HTML code, too. I would change "underscore" to "underline", though.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -1,19 +1,43 @@
...
-* `__underscore__` for __underscored__ text (<<.icon $:/core/images/underline>>)<br><br>
-* `^^superscript^^` for ^^superscripted^^ text (<<.icon $:/core/images/superscript>>)<br><br>
-* `,,subscript,,` for ,,subscripted,, text (<<.icon $:/core/images/subscript>>)<br><br>
-* `~~strikethrough~~` for ~~strikethrough~~ text (<<.icon $:/core/images/strikethrough>>)<br><br>
-* `@@highlight@@` for @@highlighted@@ text (no editor toolbar button), see [[Styles and Classes in WikiText]] for more formatting options
+Available character formatting in WikiText includes:
 
-See also: [[Code Blocks in WikiText]]
+|Wikitext | Button | Shortcut |Rendered Output |h
+|2 single hyphens are used for `''bold text''`| <kbd><<.icon $:/core/images/bold>></kbd> | <kbd><<displayshortcuts "((bold))">></kbd> |2 single hyphens are used for ''bold text'' |
```
Jermolene, line 13: It's a subtlety that hasn't made it into our docs style guide, but here we should spell out numbers rather than using digits. In this case, I think we should use "Double hyphens" etc., not "two".

https://www.grammarly.com/blog/when-to-spell-out-numbers/

Jermolene (commented):
Thanks @pmario

Jermolene:
> Should we use this PR to clean up the terms?

Excellent point. Thinking about "strong" and "emphasis", we should be guided by the needs of our users more than adherence to specifications; users are still thinking in terms of bold and italics. Maybe a footnote explaining the HTML spec angle on things. As to underscore vs. underline, I definitely agree with [Microsoft's style guide](https://learn.microsoft.com/en-us/style-guide/a-z-word-list-term-collections/u/underline-vs-underscore) on this:

> Use underline to describe text formatting that puts a line under the characters. Use underscore to refer to the underscore character ( _ ).

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
pmario: The problem here is that it actually is  2 single quotes  ... So you suggest `Double single quotes` .. right?

Jermolene: > The problem here is that it actually is 2 single quotes ... So you suggest `Double single quotes` .. right?

Yes, that's right.

pmario: ok

pmario:
I did split this tiddler into 2 now. The first one tells the basics, the second one give more advanced info. 

underscore has been changed to underline

New info about

- bold vs. strong
- underline vs. underscore
- emphasis vs italics ... 

with reference links.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -1,19 +1,35 @@
...
-* `,,subscript,,` for ,,subscripted,, text (<<.icon $:/core/images/subscript>>)<br><br>
-* `~~strikethrough~~` for ~~strikethrough~~ text (<<.icon $:/core/images/strikethrough>>)<br><br>
-* `@@highlight@@` for @@highlighted@@ text (no editor toolbar button), see [[Styles and Classes in WikiText]] for more formatting options
+Available character formatting in WikiText includes:
 
-See also: [[Code Blocks in WikiText]]
+|Wikitext | Button | Shortcut |Rendered Output |h
+|Double single quotes are used for `''bold text''`| <kbd><<.icon $:/core/images/bold>></kbd> | <kbd><<displayshortcuts "((bold))">></kbd> |Double single quotes are used for ''bold text'' |
+|Double slashes are used for `//italic text//`| <kbd><<.icon $:/core/images/italic>></kbd> | <kbd><<displayshortcuts "((italic))">></kbd> |Double slashes are used for //italic text//|
+|Double underlines are used for `__underlined text__`| <kbd><<.icon $:/core/images/underline>></kbd> | <kbd><<displayshortcuts "((underline))">></kbd>  |Double underlines are used for __underlined text__ |
```
Jermolene, line 15: The *character* should be referred to as "underscore" and the text decoration feature should be referred to as "underline"

editions/tw5.com/tiddlers/wikitext/Formatting in Wikitext Semantics.tid
```diff
@@ -0,0 +1,37 @@
+created: 20230303205650613
+modified: 20230303211541555
+tags: [[Formatting in WikiText]]
+title: Formatting in Wikitext Semantics
```
Jermolene, line 4: I don't think splitting off a separate tiddler about semantics works here. It makes it harder to understand overall because one needs to jump between the two tiddlers to understand.

As I said above, I think the HTML spec stuff about strong/emphasis should just be addressed via brief note.

Jermolene (commented):
Thanks @pmario

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
```diff
@@ -30,6 +30,31 @@ See [[Styles and Classes in WikiText]] for more formatting options
...
-See [[Formatting in Wikitext Semantics]]
\ No newline at end of file
+|Wikitext | Resulting HTML Code  |h
+|Double single quotes are used for `''bold text''`|Double single quotes are used for `<strong>bold text</strong>` |
+|Double slashes are used for `//italic text//`|Double slashes are used for `<em>italic text</em>` |
+|Double underscores are used for `__underlined text__`|Double underscores are used for `<u>underlined text<u>` |
+|Double circumflex accents are used for `^^superscripted^^` text |Double circumflex accents are used for `<sup>superscripted</sup>` text |
+|Double commas are used for `,,subscripted,,` text |Double commas are used for `<sub>subscripted</sub>` text |
+|Double tilde signs are used for `~~strikethrough~~` text |Double tilde signs are used for `<strike>strikethrough</strike>` text |
+|Single backtick is used for ```code` `` |Single backtick is used for `<code>code</code>` |
```
Jermolene, line 42: It would be better to say "Single backticks are used..." – super weird, I know, and I'm afraid I'm not familiar enough with grammar to cite the relevant rule!

Jermolene (commented):
Thanks @pmario much appreciated.

editions/tw5.com/tiddlers/wikitext/Formatting in WikiText.tid
pmario: OK done

Jermolene:
Thanks @pmario

PR #7321, opened 2023-03-03
Add HTML Entities to WikiText info

pmario:
This PR adds the HTML Entities which was part of the TWclassic documentation to the WikiText overview documentation. 

There is a problem with the `&dollar;` sign, but that needs to be a different PR changing js code.

### DIFF
--- a/editions/tw5.com/tiddlers/wikitext/HTML Entities.tid
+++ b/editions/tw5.com/tiddlers/wikitext/HTML Entities.tid
@@ -1,0 +1,66 @@
+created: 20130204165019000
+modified: 20221025124615415
+tags: WikiText
+title: HTML Entities
+type: text/vnd.tiddlywiki
+
+!! Summary
+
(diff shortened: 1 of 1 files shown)


Jermolene:
> This PR adds the HTML Entities which was part of the TWclassic documentation to the WikiText overview documentation.

Thanks @pmario 

> There is a problem with the `&dollar;` sign, but that needs to be a different PR changing js code.

What's the problem there?

pmario:
>What's the problem there?

The dollar sign is not part of the https://github.com/Jermolene/TiddlyWiki5/blob/ba2c9f44b07c46ae847d16480162a9be873dd988/core/modules/config.js#L33

I'll create a new PR to fix this one. The $ sign usually is part of keyboard layouts. So it may be an oversight.

Jermolene:
> I'll create a new PR to fix this one. The $ sign usually is part of keyboard layouts. So it may be an oversight.

Thank you @pmario that is definitely an oversight, and I'd appreciate a PR. I'll merge this in the meantime.

PR #7323, opened 2023-03-03
add dollar sign to HTML entities

pmario:
add dollar sign to HTML entities, It's the second element `dollar:36` ...


Jermolene:
Thanks @pmario

PR #7327, opened 2023-03-04
Prevent infinite loop for single-word texts

yaisog:
With regexpResult === null, lineEnd was always 0, so the comparison to null failed.


Jermolene:
Thanks @yaisog

PR #7328, opened 2023-03-04
Add docs and examples for the new operators

yaisog:
The `applypatches` examples result in pretty long lines when written as a single filter expression.
Alternatively, I could make something similar to the ActionWidget examples: Multiple lines and a button to initiate, or even leave out the button and just show the result, as in the second `levenshtein` example (will use more space upon opening the tiddler).
In this case I'd use a `$let` to put the patches into a variable and reference that in the filter that generates the actual output. This is closer to what one would do in a "real-world" wiki...

### DIFF
--- a/editions/tw5.com/tiddlers/filters/applypatches Operator.tid
+++ b/editions/tw5.com/tiddlers/filters/applypatches Operator.tid
@@ -1,0 +1,15 @@
+caption: applypatches
+created: 20230304154824762
+modified: 20230304154826621
+op-purpose: applies a set of patches to transform the input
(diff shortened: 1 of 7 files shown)


yaisog:
I gotta say ... writing the docs does take a lot more time than writing the filters. 😅

yaisog:
> Excellent idea to use the Hamlet demos.

That idea was Mario's. I did credit the _original_ source, though.

One thing I was wondering was if we should mention or link to the diff-match-patch library on GitHub, since that is not "our" code...

Issue #7302, opened 2023-02-28
[IDEA] hook indexer and tiddler store to load cache (from sqlite)


linonetwo:
**Is your feature request related to a problem? Please describe.**

I'd like to implement sqlite and indexdb cache to speedup initial loading time even further.

I think this can be implemented as plugins, so i need a way in the core to allow my plugs to hook into.

**Describe the solution you'd like**

maybe add a new type "cache provider", load at the beginning of booting, and let indexer and core require them for initialization.

And this behavior can be disabled by passing a boot parameter.

And This mechanism should be async, so there can also be a provider in the client side that load all the state from the server side.

**Describe alternatives you've considered**

maybe store state as new $:/president-state tiddler, on the fs (they should be ignored by git, Because they will be huge.), And they can be sent to the clap as normal tiddlers. (But all following Process has to wait for this tiddler loaded before doing any indexer init.

**Additional context**

Part of solving tiddler 10k problem.

Also this can be a good start to make tw async.

pmario:
Adding sqlite alone would need to include a new query language into TW. Both to create new indexes and then to query them. So what if there will be a better database as a backend in the future?

IMO to make that implementation future prove we would need the possibility to plug in different storage engines, which ideally use the TW filter syntax as a query language. We may need some more "backend independent" query operators, but that should be OK.

From the discussion at: https://github.com/Jermolene/TiddlyWiki5/discussions/7282 I did find a very promising graph-database implementation that also supports wasm. 

[CozoDB](https://github.com/cozodb/cozo), which has a somewhat TW filter-like query language which would allow us to perform graph-like queries on the low-level side of things. The tutorial shows the possibilities very well: https://docs.cozodb.org/en/latest/tutorial.html

The main concept of Cozo queries is to return "lists of lists", which is very similar to TW filters. 

CozoDB itself can use different storage engines like SQLite, RocksDB and others. [Cozo runs almost everywhere.](https://docs.cozodb.org/en/latest/releases/v0.2.html)

It has a lot of built in [functions and operations](https://docs.cozodb.org/en/latest/functions.html) that should be straight forward to be mapped to TW filters. ... at least as far as I can see it from working through the tutorial last weekend.

@Jermolene 

Even if the query language has a learning curve, especially for complex queries that use stored-configurations IMO it would be worth a look.

linonetwo:
As long as SQLite has sync version of getter like localstorage

[...]

PR #7329, opened 2023-03-04
Developer demo of alternate wiki store implementation

Jermolene:
This is the very barebones beginnings of a demo implementation of an alternate tiddler store. It is not functional. If using the Vercel builds, open developer tools in the browser to see it failing due to the absence of basic wiki methods.

The plan is to build it up into the smallest possible plain JS wiki store implementation, sharing as much implementation as possible with the existing core implementation with as little code duplication as possible. It could then serve as the basis for future experiments with wiki stores based on SQLite (@linonetwo), or a custom append only database (@yaisog).

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1107,7 +1107,7 @@ Wiki constructor. State is stored in private members that only a small number of
 options include:
 enableIndexers - Array of indexer names to enable, or null to use all available indexers
 */
-$tw.Wiki = function(options) {
+$tw.Wiki = $tw.Wiki || function(options) {
 	options = options || {};
 	var self = this,
 		tiddlers = Object.create(null), // Hashmap of tiddlers
@@ -1466,10 +1466,10 @@ $tw.Wiki = function(options) {
 // Dummy methods that will be filled in after boot
 $tw.Wiki.prototype.clearCache =
 $tw.Wiki.prototype.clearGlobalCache =
-$tw.Wiki.prototype.enqueueTiddlerEvent = function() {};
+$tw.Wiki.prototype.enqueueTiddlerEvent = $tw.Wiki.prototype.enqueueTiddlerEvent || function() {};
 
 // Add an array of tiddlers
-$tw.Wiki.prototype.addTiddlers = function(tiddlers) {
+$tw.Wiki.prototype.addTiddlers = $tw.Wiki.prototype.addTiddlers || function(tiddlers) {
 	for(var t=0; t<tiddlers.length; t++) {
 		this.addTiddler(tiddlers[t]);
 	}
@@ -1478,7 +1478,7 @@ $tw.Wiki.prototype.addTiddlers = function(tiddlers) {
 /*
 Define all modules stored in ordinary tiddlers
 */
-$tw.Wiki.prototype.defineTiddlerModules = function() {
+$tw.Wiki.prototype.defineTiddlerModules = $tw.Wiki.prototype.defineTiddlerModules || function() {
 	this.each(function(tiddler,title) {
 		if(tiddler.hasField("module-type")) {
 			switch (tiddler.fields.type) {
@@ -1502,7 +1502,7 @@ $tw.Wiki.prototype.defineTiddlerModules = function() {
 /*
 Register all the module tiddlers that have a module type
 */
-$tw.Wiki.prototype.defineShadowModules = function() {
+$tw.Wiki.prototype.defineShadowModules = $tw.Wiki.prototype.defineShadowModules || function() {
 	var self = this;
 	this.eachShadow(function(tiddler,title) {
 		// Don't define the module if it is overidden by an ordinary tiddler
@@ -1516,7 +1516,7 @@ $tw.Wiki.prototype.defineShadowModules = function() {
 /*
 Enable safe mode by deleting any tiddlers that override a shadow tiddler
 */
-$tw.Wiki.prototype.processSafeMode = function() {
+$tw.Wiki.prototype.processSafeMode = $tw.Wiki.prototype.processSafeMode || function() {
 	var self = this,
 		overrides = [];
 	// Find the overriding tiddlers
@@ -1547,7 +1547,7 @@ $tw.Wiki.prototype.processSafeMode = function() {
 /*
 Extracts tiddlers from a typed block of text, specifying default field values
 */
-$tw.Wiki.prototype.deserializeTiddlers = function(type,text,srcFields,options) {
+$tw.Wiki.prototype.deserializeTiddlers = $tw.Wiki.prototype.deserializeTiddlers || function(type,text,srcFields,options) {
 	srcFields = srcFields || Object.create(null);
 	options = options || {};
 	var deserializer = $tw.Wiki.tiddlerDeserializerModules[options.deserializer],
--- a/editions/prerelease/tiddlywiki.info
+++ b/editions/prerelease/tiddlywiki.info
@@ -14,7 +14,8 @@
 		"tiddlywiki/dynannotate",
 		"tiddlywiki/codemirror",
 		"tiddlywiki/menubar",
-		"tiddlywiki/jszip"
+		"tiddlywiki/jszip",
+		"tiddlywiki/demo-alternate-store"
 	],
 	"themes": [
 		"tiddlywiki/vanilla",
--- a/editions/tw5.com/tiddlywiki.info
+++ b/editions/tw5.com/tiddlywiki.info
@@ -6,7 +6,8 @@
 		"tiddlywiki/railroad",
 		"tiddlywiki/evernote",
 		"tiddlywiki/internals",
-		"tiddlywiki/menubar"
+		"tiddlywiki/menubar",
+		"tiddlywiki/demo-alternate-store"
 	],
 	"themes": [
 		"tiddlywiki/vanilla",
--- a/plugins/tiddlywiki/demo-alternate-store/plugin.info
+++ b/plugins/tiddlywiki/demo-alternate-store/plugin.info
@@ -1,0 +1,6 @@
+{
+	"title": "$:/plugins/tiddlywiki/demo-alternate-store",
+	"name": "Demo alternate store",
+	"description": "Developer demo of an alternate wiki store implementation",
+	"list": "readme"
+}
--- a/plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
+++ b/plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
@@ -1,0 +1,22 @@
+/*\
+title: $:/plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
+type: application/javascript
+module-type: library
+
+Startup code injected as raw markup
+
+\*/
+
+(function() {
+
+// Need to initialise these because we run before bootprefix.js and boot.js
+$tw = window.$tw || Object.create(null);
+$tw.hooks = $tw.hooks || { names: {}};
+$tw.boot = $tw.boot || {};
+$tw.boot.preloadDirty = $tw.boot.preloadDirty || [];
+
+$tw.Wiki = function() {
+	
+};
+
+})();
--- a/plugins/tiddlywiki/demo-alternate-store/rawmarkup.tid
+++ b/plugins/tiddlywiki/demo-alternate-store/rawmarkup.tid
@@ -1,0 +1,6 @@
+title: $:/plugins/tiddlywiki/demo-alternate-store/rawmarkup
+tags: $:/tags/RawMarkupWikified
+
+`<script>`
+{{$:/plugins/tiddlywiki/demo-alternate-store/rawmarkup.js}}
+`</script>`
--- a/plugins/tiddlywiki/demo-alternate-store/readme.tid
+++ b/plugins/tiddlywiki/demo-alternate-store/readme.tid
@@ -1,0 +1,3 @@
+title: $:/plugins/tiddlywiki/demo-alternate-store/readme
+
+Developer demo of an alternate wiki store implementation


boot/boot.js
```diff
@@ -1466,10 +1466,10 @@ $tw.Wiki = function(options) {
 // Dummy methods that will be filled in after boot
 $tw.Wiki.prototype.clearCache =
 $tw.Wiki.prototype.clearGlobalCache =
-$tw.Wiki.prototype.enqueueTiddlerEvent = function() {};
+$tw.Wiki.prototype.enqueueTiddlerEvent = $tw.Wiki.prototype.enqueueTiddlerEvent || function() {};
```
btheado, line 1469: Is this change and all the following in this file so these methods can be optionally be defined in the code of the alternate store implementation?

The comment says they "will be filled in after boot". Won't that overwrite the method if it is implemented in the alternate store?

Also with the above it looks like clearCache and clearGlobalCache will also have the same function as enqueueTiddlerEvent if it has an alternate store implementation. Is that intended?

pmario:
I always thought, that the TW concept could have "additional" $tw.wiki stores if needed. ... 

With the implementation like this there can only be 1 type of store. Is this intended or is it just an experiment?

linonetwo:
In plugin, we always call `$tw.wiki`, so a singleton is better.

Just need some way to hook into existing $tw.wiki

boot/boot.js
Jermolene: > The comment says they "will be filled in after boot". Won't that overwrite the method if it is implemented in the alternate store?

These three methods all have an implementation loaded from wiki.js, but need to be invoked before wiki.js is fully loaded, so we provide this dummy implementation.

> Is this change and all the following in this file so these methods can be optionally be defined in the code of the alternate store implementation?

That's the idea. At the moment the additional methods in wiki.js are loaded unconditionally, and so can't be overridden by alternate stores, which is OK for the moment but might need to be revisited

> Also with the above it looks like clearCache and clearGlobalCache will also have the same function as enqueueTiddlerEvent if it has an alternate store implementation. Is that intended?

These methods are currently overwritten by the implementations in wiki.js, and as above there's currently no support for alternate stores to override those methods.

The wiki store implementation in boot.js has always used a closure for key state information like the hashmap of tiddlers and shadow tiddlers, which makes clear which wiki methods actually have and need access to the raw tiddler store, and to encapsulate the implementation details of the store. That means that it should be possible to just override those methods that have access to the private data in the closure, and seamlessly take advantage of the implementation of the higher level methods in wiki.js.

Jermolene:
> I always thought, that the TW concept could have "additional" $tw.wiki stores if needed. ...

I think that's an orthogonal goal, which would be accomplished by reclassifying the existing `$tw.wiki` instance as the root admin wiki, and allow for a dynamic set of child wikis that can be created and deleted without a restart.

> With the implementation like this there can only be 1 type of store. Is this intended or is it just an experiment?

An alternate store along these lines could still expose its own constructor so that further instances could be created.

boot/boot.js
btheado: Sounds reasonable and I might still be missing part of your point.

By my thinking these 3 lines should have either remained unchanged since overwriting them would only last until wiki.js is loaded and therefore isn't very useful:

```js
$tw.Wiki.prototype.clearCache =
$tw.Wiki.prototype.clearGlobalCache =
$tw.Wiki.prototype.enqueueTiddlerEvent = function() {};
```

or each should have their own separate overwrite:

```js
$tw.Wiki.prototype.clearCache = $tw.Wiki.prototype.clearCache || function();
$tw.Wiki.prototype.clearGlobalCache = $tw.Wiki.prototype.clearGlobalCache || function();
$tw.Wiki.prototype.enqueueTiddlerEvent = $tw.Wiki.prototype.enqueueTiddlerEvent || function();
```

Jermolene: Hi @btheado I think you're right that there's no point in providing an underride for these methods because they will be overwritten later.

joshuafontany:
Thanks for starting to sketch this out. I think this would allow tighter integration between TW and data-stores like the Yjs CRDT document(s). :)

linonetwo:
About how to use sqlite and if it should be a plugin https://github.com/Jermolene/TiddlyWiki5/discussions/7531

linonetwo:
How to support multiple plugins?

I want to create a sqlite plugin that works on nodejs server side, loading tiddler and indexer from sqlite instead of fs, then do swr from fs.
I also want to create a plugin that work on client side, to get indexer&tiddlers using ipc, so we can skip most of parsing on client booting.
And there might be a plugin you want to create, work on client side, using wasm sqlite to cache content...

linonetwo:
How to support multiple plugins?

I want to create a sqlite plugin that works on nodejs server side, loading tiddler and indexer from sqlite instead of fs, then do swr from fs.
I also want to create a plugin that work on client side, to get indexer&tiddlers using ipc, so we can skip most of parsing on client booting.
And there might be a plugin you want to create, work on client side, using wasm sqlite to cache content...

Jermolene:
Hi @linonetwo the "demo-alternate-store" plugin in this PR is intended to be a starting point for people wanting to write their own custom alternate store plugins. The current implementation only works in the browser but will be extended to work under Node.js as well.

To support multiple alternate store plugins being loaded at the same time we will also need a mechanism for selecting between them.

Jermolene:
I've updated the PR to include the start of a sqlite3 wiki store plugin. At this point, the sqlite3 engine is initialised and tested with some test queries. The next step is to wire it up into the wiki object.

btheado:
> At this point, the sqlite3 engine is initialised and tested with some test queries. The next step is to wire it up into the wiki object.

Nice work. I especially like the code showing how to embed wasm into a tiddlywiki. If someone else wants to embed something wasm, then using your code as a starting point should make it easy.

It looks like you are using in memory database. Do you plan to experiment with persistence?

An alternate wiki store which has persistent backing seems like it would be useful, but I'm confused about how it would work. All tiddlers in the html store area will be loaded into the store. If that store persists, then you wouldn't want to load the tiddlers from the html store the next time the page is loaded. Otherwise they might overwrite any changes which were made.

Or am I on the wrong track here? Maybe an automatically persisting store is not what you have in mind for these alternate stores?

linonetwo:
> Otherwise they might overwrite any changes which were made.

@btheado we can use swr, try search stale-while-revalidate in [HTTP RFC 5861](https://datatracker.ietf.org/doc/html/rfc5861)

Jermolene:
> Nice work. I especially like the code showing how to embed wasm into a tiddlywiki. If someone else wants to embed something wasm, then using your code as a starting point should make it easy.

Thank you! I am indeed interested in making it as easy as possible to incorporate Wasm functionality. It would be useful to have another example – perhaps something similar to https://squoosh.app/ for image compression.

> It looks like you are using in memory database. Do you plan to experiment with persistence?

At this point I am exploring using sqlite3 as a volatile store just for its querying and indexing capabilities. I am also deliberately starting by integrating sqlite in the browser where the persistence options are more limited than under Node.js

> An alternate wiki store which has persistent backing seems like it would be useful, but I'm confused about how it would work. All tiddlers in the html store area will be loaded into the store. If that store persists, then you wouldn't want to load the tiddlers from the html store the next time the page is loaded. Otherwise they might overwrite any changes which were made.

Yes, exactly. It would perhaps be more of a matter of importing an external TiddlyWiki HTML file, rather than opening it in place in the traditional way.

> Or am I on the wrong track here? Maybe an automatically persisting store is not what you have in mind for these alternate stores?

I do absolutely think there are some interesting options there. For example, we might establish a new format based on sqlite3 for TiddlyWiki files that only contain tiddlers, and rely on an external app to work.

plugins/tiddlywiki/sqlite3store/sql-functions.js
```diff
@@ -0,0 +1,217 @@
...
+		union
+			select title, meta, text
+			from tiddlers t3
+			where t3.shadow = 0
+				and exists (
+					select 1
+					from tiddlers t4
+					where t4.title = t3.title
+						and t4.shadow = 1
+				);
```
btheado, line 156: This looks like a syntax error with the semicolon there.

Also, I've rarely seen use of exists and not exists in SQL. The not exists can be replaced with left outer join and a normal join for exists like this:

```sql
/* shadow tiddlers which do not have any overriding tiddler */
select t1.title as title, t1.meta as meta, t1.text as text
from tiddlers t1
left join tiddlers t2 on t1.title = t2.title and t2.shadow = 0
where t1.shadow = 1 and t2.title is null
union
/* normal tiddlers which are overriding a shadow tiddler */
select t3.title, t3.meta, t3.text
from tiddlers t3
join tiddlers t4 on t3.title = t4.title and t4.shadow = 1
where t3.shadow = 0
order by title;
```
Not sure if you had specific reason to go with the exist/not exists approach, but I thought I'd share this alternative.

plugins/tiddlywiki/sqlite3store/sql-functions.js
```diff
@@ -0,0 +1,217 @@
...
+			text TEXT NOT NULL,
+			PRIMARY KEY(title,shadow)
+		);
+		CREATE INDEX tiddlers_title_index ON tiddlers(title);
+		`
+	});
+	/*
+	Save a tiddler. shadowSource should be falsy for ordinary tiddlers, or the source plugin title for shadow tiddlers
+	*/
+	var statementSaveTiddler = self.db.prepare("replace into tiddlers(title,shadow,shadowsource,meta,text) values ($title,$shadow,$shadowsource,$meta,$text);");
```
btheado, line 37: https://sqlite.org/wasm/doc/trunk/api-oo1.md#stmt
> It is important that statements be finalized in a timely manner, else clients risk introducing locking errors later on in their apps.

I don't have experience with long-lived sqlite connections, so I don't know the behavior from experience, but the above stood out to me.

plugins/tiddlywiki/sqlite3store/sql-functions.js
Jermolene: > This looks like a syntax error with the semicolon there.

Thanks, fix coming up.

> Also, I've rarely seen use of exists and not exists in SQL. The not exists can be replaced with left outer join and a normal join for exists like this:

I've never really got to grips with complex SQL and so I'm ashamed to say that ChatGPT helped with these queries. It's probably related to the wording of my prompt: "get all the rows where either the shadow field is 1 and there is no row with the same title and a shadow field value of zero, or the shadow field is zero and there also exists a row with the same title and a shadow field value of 1". I think perhaps my use of the word "exists" triggered the use of that construction.

In terms of clarity, as a non-SQL person, I find the exists/not exists fairly readable, but that probably is because I need to get more familiar with the principles of relational algebra. It would be interesting to know which has better performance.

> Also, it might be worthwhile to create views for these long queries? Just an idea.

That makes sense. I've little experience with SQL since Microsoft Access days and so would welcome help on this.

plugins/tiddlywiki/sqlite3store/sql-functions.js
Jermolene: My first implementation used db.exec() but it was super slow - more than a minute to boot the prerelease in Chrome on my machine. Switching to bind got performance to where it is now; very slow but bearable for working with it. But I'm sure there will be better approaches, I'm still feeling my way with sqlite.

plugins/tiddlywiki/sqlite3store/sql-functions.js
btheado: > In terms of clarity, as a non-SQL person, I find the exists/not exists fairly readable, but that probably is because I need to get more familiar with the principles of relational algebra. It would be interesting to know which has better performance.

I've written a lot of complex SQL over the years, but only sporadically and much less in recent years. By no means am I a SQL expert. A quick google search seemed to indicate that not exists might perform better. Most search results I found were related to microsoft sql server. Probably the only way to know for sure is to test, if you ever get to that point of needing to know.

> That makes sense. I've little experience with SQL since Microsoft Access days and so would welcome help on this.

This part is simple. It is just `create view <yourviewname> as <your sql query>`

pmario:
@Jermolene ... I did switch to this branch and I did create an empty.html which is a 3.992KB in size. From your comments I saw that's expected. 

Starting the file wiki with FF 115 needed about 24 seconds. .. Is that expected. 

The console showed the following. Is that expected?

```
Ignoring inability to install OPFS sqlite3_vfs: Cannot install OPFS: Missing SharedArrayBuffer and/or Atomics. The server must
emit the COOP/COEP response headers to enable those. See https://sqlite.org/wasm/doc/trunk/persistence.md#coop-coep
```

The wiki loaded, but then everything as in every key-stroke needs about 4-5 seconds to do anything. Clicking the NewTiddler button about 10 seconds. It shows a tiddler. 
Editing every keystroke 3-5 seconds. 

I did test with Windows Edge. Exactly the same thing. Except Edge pops up a window to kill the process, where I have to select "Wait". 

I think that behaviour is not expected?

Jermolene:
> Starting the file wiki with FF 115 needed about 24 seconds. .. Is that expected.

Yes. At the moment everything is super slow because I'm using a custom JS collation function so that we can precisely match the ordering performed by the original wiki store. The plan is to have a switch to turn that off.

Also, there's currently no indexing happening, so getting all tiddlers with a particular tag does a full scan of the wiki.

> The console showed the following. Is that expected?

Yes as far as I can tell.

> The wiki loaded, but then everything as in every key-stroke needs about 4-5 seconds to do anything. Clicking the NewTiddler button about 10 seconds. It shows a tiddler. Editing every keystroke 3-5 seconds.
> 
> I did test with Windows Edge. Exactly the same thing. Except Edge pops up a window to kill the process, where I have to select "Wait".
> 
> I think that behaviour is not expected?

Yes that is expected at this point.

pmario:
OK. Thx for the info. I only did skim the code. ... So every future performance improvement will actually feel like one. .. clever ;)

Jermolene:
Thanks @pmario – and no problem, those were good questions that I was happy to answer, and hopefully will be helpful for others too.

> So every future performance improvement will actually feel like one.

That's the hope! The ultimate goal is to compile TW filter expressions into SQL queries...

Jermolene:
I've updated the OP with the latest progress on the sqlite3 store

Jermolene:
I've done some work to simplify the schema that we use. At this point everything works, but performance is terrible because it doesn't yet have the tag optimisations that we previously had.

Jermolene:
I've now completed a major refactoring consisting of the following steps:

* c52014c66ffbf460392ae9ed5b824ad4153265e0 simplifies the SQL schema by shifting more of the work of handling shadow tiddlers from read operations to write operations like saving/deleting tiddlers. This update also temporarily removed most of the optimisations for reading tiddlers by tag
* d4dec0ca65d349c0f02aa6de997dc9e0cd6b2cea switches from using NULL as a special plugin title indicating a non-shadow tiddler to using the empty string. Using NULL was working against the grain of SQL, and using strings consistently allows the code to be much simpler
* f49b9faab0730234883a938aac386842fe42ba11 restores the additional tables used for more efficient tag handling
* 12c6cb35a0336ee44a61a26a643db368b019e45e adds indexes to all the columns used in joins. I didn't notice much impact on performance

Performance is better than before, but not decisively so. I plan now to spend some time profiling the code to find the bottlenecks.

I'd welcome any help with any analysis or optimisations. As ever, questions and suggestions welcome.

core/modules/filters.js
```diff
@@ -171,15 +171,15 @@ exports.parseFilter = function(filterString) {
...
+						$tw.utils.each(match[3].split(":"),function(subsuffix) {
 						operation.suffixes.push([]);
 						$tw.utils.each(subsuffix.split(","),function(entry) {
 							entry = $tw.utils.trim(entry);
 							if(entry) {
 								operation.suffixes[operation.suffixes.length -1].push(entry);
 							}
 						});
-					 });
+						});
```
pmario, line 182: I think the indent-level here is wrong

Jermolene: Thanks @pmario fixed in 4897248809d74e59dc6164c337a6553c8c4607c5

plugins/tiddlywiki/sqlite3store/init-sqlite3.js
```diff
@@ -0,0 +1,66 @@
...
+var thisPlugin = getTiddler("$:/plugins/tiddlywiki/sqlite3store"),
+	thisPluginTiddlers = JSON.parse(thisPlugin.text).tiddlers;
+// Execute the sqlite3 module
+var sqlite3js = thisPluginTiddlers["$:/plugins/tiddlywiki/sqlite3store/sqlite3.js"].text,
+	context = {
+		exports: {}
+	};
+$tw.utils.evalSandboxed(sqlite3js,context,"$:/plugins/tiddlywiki/sqlite3store/sqlite3.js",true);
+// Create a Blob URL for the wasm data
+var sqlite3wasm = thisPluginTiddlers["$:/plugins/tiddlywiki/sqlite3store/sqlite3.wasm"].text;
```
linonetwo, line 37: Why this, instead of `$tw.wiki.getTiddlerText` ?

Context is https://github.com/Jermolene/TiddlyWiki5/discussions/7947

plugins/tiddlywiki/sqlite3store/init-sqlite3.js
```diff
@@ -0,0 +1,66 @@
...
+var decodedData = window.atob(sqlite3wasm),
```
linonetwo, line 38: I get

```
Uncaught InvalidCharacterError: Failed to execute 'atob' on 'Window': The string to be decoded contains characters outside of the Latin1 range.
```

when mimicking your usage.

btheado: @linonetwo, notice Jeremy's tiddlywiki.info is encoding the file as base64. The `atob` can't decode it if it wasn't encoded.

```json
		{
			"file": "sqlite3.wasm",
			"encoding": "base64",
			"fields": {
				"type": "application/wasm",
				"title": "$:/plugins/tiddlywiki/sqlite3store/sqlite3.wasm"
			}
		}
```

linonetwo: @btheado I have a few tries, even I change encoding to utf8, when I use `$tw.wiki.getTiddlerText`, I always get the base64 encoded test. Same even set `"type": "text/text"`.

I tried this because 30MB wasm become 80MB of string after become a tiddler and I want to reduce it.

boot/boot.js
```diff
@@ -2438,6 +2471,7 @@ $tw.boot.initStartup = function(options) {
 	$tw.utils.registerFileType("image/svg+xml","utf8",".svg",{flags:["image"]});
 	$tw.utils.registerFileType("image/vnd.microsoft.icon","base64",".ico",{flags:["image"]});
 	$tw.utils.registerFileType("image/x-icon","base64",".ico",{flags:["image"]});
+	$tw.utils.registerFileType("application/wasm","base64",".wasm");
```
linonetwo, line 2474: This is important. Without this, wasm tiddler will be 2x larger, and can't properly loaded.

plugins/tiddlywiki/sqlite3store/init-sqlite3.js
Jermolene: To be clear, TiddlyWiki doesn't currently support true binary tiddlers; we only support base64 encoded pseudo-binary tiddlers.

linonetwo:
I want to send server-side built backlink-indexer to the client to solve #7302 , and I think it don't necessarily need sqlite. Simply serialize the indexer as json and send to client will save tons of time scanning tiddlers again to build indexer. Also allows client that uses lazy-all to show backlinks.

Jermolene:
> I want to send server-side built backlink-indexer to the client to solve #7302 , and I think it don't necessarily need sqlite. Simply serialize the indexer as json and send to client will save tons of time scanning tiddlers again to build indexer. Also allows client that uses lazy-all to show backlinks.

Hi @linonetwo that's a reasonable idea. A plugin could completely replace the existing indexer modules with replacements that retrieve the index in the background. I don't see any obvious core changes required to do this, but would be happy to explore any that crop up.

PR #7330, opened 2023-03-05
Markdown Plugin: Ensure WikiParser is loaded before MarkdownParser

cdruan:
This PR updates the wrapper to require `wikiparser.js` module, in case the markdown parser module is loaded before the wikiparser module. The scenario could happen when the wrapper shadow tiddler is overwritten by the users.


Jermolene:
Thanks @cdruan

PR #7332, opened 2023-03-06
Improve the StoryTiddlerTemplate fallback

### LINKED ISSUE #7311: [IDEA] Add fallback to StoryTiddlerTemplate
Both the `$:/config/StoryTiddlerTemplateFilters/default` cascade component as well as the current fallback in `$:/core/ui/StoryTiddlerTemplate` transclude the content of `$:/config/ui/ViewTemplate`. If the title set there doesn't exist, the wiki is bricked. Changing the PageTemplate will not bring it back.

I suggest adding this filter run at the end of `$:/core/ui/StoryTiddlerTemplate`:
```
:and[has[title]else[$:/core/ui/ViewTemplate]]
```
i.e. if the referenced tiddler does not exist as shadow or tiddller, fall back to the original core ViewTemplate.

yaisog:
I'm marking this as a draft PR, because this may be worth discussing.

Instead of adding _another_ fallback to the StoryTiddlerTemplate cascade, I _changed_ the existing fallback.

I believe those things could go wrong with the StoryTiddlerTemplate cascade due to user experimenting:
1) `draft` cascade entry is rendered unusable
2) `default` cascade entry is rendered unusable
3) `$:/config/ui/EditTemplate` is rendered unusable
4) `$:/config/ui/ViewTemplate` is rendered unusable

In the list, 3 will lead to 1 and 4 will lead to 2, because they are transcluded respectively. Also, 1 and 3 will lead to a ViewTemplate be used for the EditTemplate. In cases 1 and 2 the template falls back to `$:/config/ui/ViewTemplate`.

In all these cases, falling back to `$:/core/ui/ViewTemplate` would enable the user to delete faulty templates and configs to go back to the shadow versions, while currently 4 will brick the wiki.

Since falling back to the current `$:/config/ui/ViewTemplate` helps only in the edge case that 2 occurs without 4 occuring, I would _replace_ it to keep the logic simple.

Optionally, additionally, we could add a fallback to both cascades with the correct type (EditTemplate for `draft`, etc.), to catch problems with the config tiddlers.

Did I miss something?
Should we rather have two fallbacks, as in the initial commit?

I would also like to amend the cascade intro in the control panel to:
"These global rules are used to dynamically choose certain templates. The result of the cascade is the result of the first filter in the sequence that returns a result. _If there are problems with custom cascade filters, change their order in the corresponding tag dropdown or delete the changed cascade tiddlers to return to their shadow versions._

Closes #7311

### DIFF
--- a/core/language/en-GB/ControlPanel.multids
+++ b/core/language/en-GB/ControlPanel.multids
@@ -28,7 +28,7 @@ Basics/Title/Prompt: Title of this ~TiddlyWiki
 Basics/Username/Prompt: Username for signing edits
 Basics/Version/Prompt: ~TiddlyWiki version
 Cascades/Caption: Cascades
-Cascades/Hint: These global rules are used to dynamically choose certain templates. The result of the cascade is the result of the first filter in the sequence that returns a result. In case of problems with custom cascade filters, change their order in the corresponding tag dropdown or delete the changed cascade tiddlers to return to their original shadow versions.
+Cascades/Hint: These global rules are used to dynamically choose certain templates. The result of the cascade is the result of the first filter in the sequence that returns a result
 Cascades/TagPrompt: Filters tagged <$macrocall $name="tag" tag=<<currentTiddler>>/>
 EditorTypes/Caption: Editor Types
 EditorTypes/Editor/Caption: Editor
--- a/core/language/en-GB/ControlPanel.multids
+++ b/core/language/en-GB/ControlPanel.multids
@@ -28,7 +28,7 @@ Basics/Title/Prompt: Title of this ~TiddlyWiki
 Basics/Username/Prompt: Username for signing edits
 Basics/Version/Prompt: ~TiddlyWiki version
 Cascades/Caption: Cascades
-Cascades/Hint: These global rules are used to dynamically choose certain templates. The result of the cascade is the result of the first filter in the sequence that returns a result
+Cascades/Hint: These global rules are used to dynamically choose certain templates. The result of the cascade is the result of the first filter in the sequence that returns a result. In case of problems with custom cascade filters, change their order in the corresponding tag dropdown or delete the changed cascade tiddlers to return to their original shadow versions.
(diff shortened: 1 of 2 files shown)


saqimtiaz:
Another consideration is whether in safe mode the cascade should also default to the core templates, providing another means for recovering a wiki in case of broken templates. In particular, a broken edit template can make it very difficult to make other changes to make the wiki operable.

yaisog:
Hi @saqimtiaz, I don't know how safe mode works and would have to be done.

Falling back to the default templates in safe mode sounds like a reasonable idea, though.

> In particular, a broken edit template can make it very difficult to make other changes to make the wiki operable.

I noticed that during my tests with this PR. That's why I want to add the note from the OP to the cascade list. Deleting broken cascade filters should work with only the ViewTemplate intact, from the view toolbar.

Jermolene:
Hi @yaisog I think this a good idea, and generally would welcome efforts to make the cascades more robust.

yaisog:
> Another consideration is whether in safe mode the cascade should also default to the core templates, providing another means for recovering a wiki in case of broken templates. In particular, a broken edit template can make it very difficult to make other changes to make the wiki operable.

This should probably be implemented as part of #7357.

saqimtiaz:
> This should probably be implemented as part of #7357.

How so? None of the core cascades are used in that layout.

yaisog:
> How so? None of the core cascades are used in that layout.

I'm sorry, I think I misunderstood you. How would the cascade "default to the core templates" if it is not used in safe mode?

saqimtiaz:
> How would the cascade "default to the core templates" if it is not used in safe mode?

The draft PR in question proposes a new layout that is automatically triggered in safe mode. However users can still switch to another layout for further affordances for debugging.

 Whether or not we want the cascades to default to standard values in safe mode is an independent question that we need to consider. If there is a consensus that this is desirable, this PR would be the ideal place to make that change.

PR #7333, opened 2023-03-06
Add a description and examples for "strict" results from getindex

### LINKED ISSUE #7331: [IDEA] Add 'strict' mode to getindex to return empty values
If a data tiddler contains an index, but that index has no value, `getindex[]` doesn't return an empty string, but nothing. It is therefore impossible to distinguish between the data tiddler not containing the index and the index having no value.

For backwards compatibility, I propose to add a `strict` suffix to `getindex`, which when set returns an empty string in such cases.

I think the corresponding code is in `$:/core/modules/filters/getindex.js`:
```
data = options.wiki.extractTiddlerDataItem(tiddler,operator.operand);
if(data) {
	results.push(data);
}
```
I think the condition would have to be changed to check for `undefined` (or whatever is returned from `extractTiddlerDataItem` when the index doesn't exist).

yaisog:
The operator `getindex` does not return empty strings when an index is defined, but not set. There is a workaround using `:filter` and `:and` / `:map` which is described as a short tip and with some additional examples.

Closes #7331

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/getindex.tid
+++ b/editions/tw5.com/tiddlers/filters/examples/getindex.tid
@@ -1,8 +1,18 @@
 created: 20150203140000000
-modified: 20170608150301791
+modified: 20230306140231272
 tags: [[getindex Operator]] [[Operator Examples]]
 title: getindex Operator (Examples)
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @yaisog

PR #7337, opened 2023-03-07
fix some typos and add missing tiddlers - German Edition

pmario:
fix some typos and add missing tiddlers


Jermolene:
Thanks @pmario

PR #7340, opened 2023-03-07
Add a configurable `$messagecatcher` to the PageTemplate and go back in history upon close/delete

### LINKED ISSUE #7262: [IDEA] Should tm-close-tiddler (optionally) navigate to the previous tiddler?

yaisog:
I want to use the vercel preview of this draft PR to see how it feels with classic storyview and in general.

This PR will add a conditional `$messagecatcher` just inside the `$navigator` in the PageTemplate. Only if any tiddlers with the tag `$:/tags/PageTemplateMessageHandler` exist, the catcher is generated by `$genesis` and assigns the actions in those tiddlers to the messages they correspond to (via the `message` field). Otherwise there will be no difference to a vanilla TW.

I have also added two such message handlers for `tm-close-tiddler` and `tm-delete-tiddler`, which navigate to the last tiddler in the HistoryList that is also currently in the StoryList. If none are found, the preceding tiddler in the story is navigated to, and if there is none then the succeeding tiddler. These last two navigation actions are needed in a single-tiddler story view (which is based on the `current-tiddler` field of the HistoryList) to keep the `current-tiddler` field current.

The action of these handler tiddlers would later be controlled via a ControlPanel setting, which allows to switch between the back-in-history or current (default) behavior. I'm not sure how this would be best implemented, though.

Resolves #7262

### DIFF
--- a/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler.tid
+++ b/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler.tid
@@ -1,0 +1,7 @@
+message: tm-delete-tiddler
+tags: $:/tags/PageTemplateMessageHandler
+title: $:/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler
+
+\import [[$:/core/ui/PageTemplate/MessageHandler/tm-close-tiddler]]
+
+<<navigate-and-forward>>
--- a/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler.tid
+++ b/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler.tid
@@ -2,6 +2,21 @@ message: tm-delete-tiddler
 tags: $:/tags/PageTemplateMessageHandler
 title: $:/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler
 
-\import [[$:/core/ui/PageTemplate/MessageHandler/tm-close-tiddler]]
+\define navigate-and-forward()
+\define tv-action-refresh-policy() always
+<$let targetTiddler={{{ [<event-param>!is[blank]else<event-tiddlerTitle>] }}} history={{{ [<tv-history-list>get[text]] }}}>
+    <!-- forward the original message -->
+    <$action-sendmessage $message=<<event-type>> $param=<<targetTiddler>> />
+    <!-- only go back to previous tiddler if delete was not canceled -->
+    <$list filter="[<targetTiddler>!is[tiddler]]" variable="void">
+        <$list filter="[<tv-history-list>get[current-tiddler]match<targetTiddler>]" variable="void">
+            <!-- go the the last tiddler in history which is also currently open, and if none are found then to the previous / next tiddler in the story -->
+            <$let previousItem={{{ [<history>jsonindexes[]] :map[<history>jsonget<currentTiddler>,[title]] :intersection[list<tv-story-list>] +[!match<targetTiddler>reverse[]first[]] :else[list<tv-story-list>before<targetTiddler>] :else[list<tv-story-list>after<targetTiddler>] }}}>
+                <$action-navigate $to=<<previousItem>> />
+            </$let>
+        </$list>
+	  </$list>
+</$let>
+\end
 
 <<navigate-and-forward>>
--- a/core/ui/PageTemplate.tid
+++ b/core/ui/PageTemplate.tid
@@ -19,6 +19,10 @@ icon: $:/core/images/layout-button
 
 <$navigator story="$:/StoryList" history="$:/HistoryList" openLinkFromInsideRiver={{$:/config/Navigation/openLinkFromInsideRiver}} openLinkFromOutsideRiver={{$:/config/Navigation/openLinkFromOutsideRiver}} relinkOnRename={{$:/config/RelinkOnRename}}>
 
+<$genesis $type="$messagecatcher" 
+          $names="[all[shadows+tiddlers]tag[$:/tags/PageTemplateMessageHandler]!is[draft]has[message]get[message]addprefix[$]]"
+          $values="[all[shadows+tiddlers]tag[$:/tags/PageTemplateMessageHandler]!is[draft]has[message]get[text]]">
+
 <$dropzone enable=<<tv-enable-drag-and-drop>>>
 
 <$list filter="[all[shadows+tiddlers]tag[$:/tags/PageTemplate]!has[draft.of]]" variable="listItem">
@@ -28,6 +32,8 @@ icon: $:/core/images/layout-button
 </$list>
 
 </$dropzone>
+
+</$genesis>
 
 </$navigator>
 
--- a/core/ui/PageTemplate/MessageHandler/tm-close-tiddler.tid
+++ b/core/ui/PageTemplate/MessageHandler/tm-close-tiddler.tid
@@ -1,0 +1,18 @@
+message: tm-close-tiddler
+tags: $:/tags/PageTemplateMessageHandler
+title: $:/core/ui/PageTemplate/MessageHandler/tm-close-tiddler
+
+\define navigate-and-forward()
+<$let targetTiddler={{{ [<event-param>!is[blank]else<event-tiddlerTitle>] }}} history={{{ [<tv-history-list>get[text]] }}}>
+	<$list filter="[<tv-history-list>get[current-tiddler]match<targetTiddler>]" variable="void">
+		<!-- go the the last tiddler in history which is also currently open, and if none are found then to the previous / next tiddler in the story -->
+		<$let previousItem={{{ [<history>jsonindexes[]] :map[<history>jsonget<currentTiddler>,[title]] :intersection[list<tv-story-list>] +[!match<targetTiddler>reverse[]first[]] :else[list<tv-story-list>before<targetTiddler>] :else[list<tv-story-list>after<targetTiddler>] }}}>
+			<$action-navigate $to=<<previousItem>> />
+		</$let>
+	</$list>
+	<!-- forward the original message -->
+	<$action-sendmessage $message=<<event-type>> $param=<<targetTiddler>> />
+</$let>
+\end
+
+<<navigate-and-forward>>


Jermolene:
> Does anyone know how the `$messagecatcher`, or any widget for that matter, handles multiple instances of the same attribute? What would happen in the following case:
> 
> ```
> <$messagecatcher $tm-navigate=<<actions1>> $tm-navigate=<<actions2>> >
> ```

Generally, the last attribute with a given name wins, so `<$text text="Elephant" text="Rhinoceros"/>` displays "Rhinoceros".

The wrinkle is that when we implemented the `<$let>` widget we added an alternative way for widgets to access their attributes as an array property called `orderedAttributes` alongside the usual hashmap in the `attributes` property. This array does permit multiple attributes with the same name.

I think it would be reasonable to update the `<$messagecatcher>` widget to use the ordered attributes array, and thus to permit multiple handlers for the same message. The implementation is likely to be quite fiddly I think.

Stepping back, I think there might be an alternative, more generic way to handle this problem. I'd like to add a more generic version of the `<$importvariables>` widget that can import arbitrary parse tree nodes, not just `<$set>` widgets. We might call it "filtered transclusion":

```
<$transclude $filter="[all[shadows+tiddlers]tag[$:/tags/PageTemplateImports]]">
...rest of page template...
</$transclude>
```

Each tiddler returned by the filter would be transcluded, with the next one being inserted into the position marked by a `<$fill>` widget. The contents of the `<$transclude>` widget would be inserted into the final transcluded tiddler.

I'm not sure if it should perhaps be a separate widget or if it makes more sense to implement it within the current `<$transclude>` widget. 

The advantage of using this mechanism to extend the page template is that it would allow users to insert both message and event catcher widgets, or custom widgets.

yaisog:
Sounds reasonable to make the flexible insertion of items more generic. I remember that @saqimtiaz made a similar proposition in the issue related to this PR.

> Each tiddler returned by the filter would be transcluded, with the next one being inserted into the position marked by a `<$fill>` widget. The contents of the `<$transclude>` widget would be inserted into the final transcluded tiddler.

I'm having a bit of a hard time imagining what this would look like, unfortunately. What would the contents of a tagged tiddler as returned by the filter be exactly? A catcher widget always needs a closing tag. How is that placed with the `$fill` mechanism? Could you please provide a short pseudo-code example, so it'll become clearer?

saqimtiaz:
@yaisog we already import set widgets despite their closing tag.

For example, given a tiddler `myvariable` with the content:

```
<$set name="name" value="Saq">
</$set>

```

The following in another tiddler will wrap a set widget around the content (you can use preview in the TW editor to see the widget tree it creates):
```
<$importvariables filter="[[myvariable]]">

Hello There <<name>>.
</$importvariables>
```

For only importing Set Widgets this is relatively simple in that each subsequent widget goes inside the previous one. For larger parse trees being imported as would be the case here, we could use the slot/fill mechanism to specify where the subsequent import should be inserted.

yaisog:
Hi @saqimtiaz, many thanks. I was not familiar with the `$importvariables` widget yet.

> For larger parse trees being imported as would be the case here, we could use the slot/fill mechanism to specify where the subsequent import should be inserted.

You mean that the `$fill` slot for the next tiddler would be defined somewhere within the previous tiddler that was imported this way? Their order would then be determined by the usual tag ordering?

I really like the idea, because it would include event catching, too. I'm afraid that I cannot be the one implementing this, as much as I would like to though, as this is beyond my skill level of core programming... 😢

saqimtiaz:
> I really like this generic implementation idea, because it would include event catching, too. I'm afraid that I cannot be the one implementing this, as much as I would like to though, as this is beyond my skill level of core programming...

No worries @yaisog. It has been extremely pleasing to see your contributions to the core, and thank you for this PR which will hopefully serve as the impetus for Jeremy or myself to work on this for 5.3.0. Please do leave the PR open until we have a PR for a better implementation.

btheado:
> Each tiddler returned by the filter would be transcluded, with the next one being inserted into the position marked by a `<$fill>` widget. The contents of the `<$transclude>` widget would be inserted into the final transcluded tiddler.

Is this backwards? The way I read the slot/fill documentation it seems `$slot` is used to mark positions and `$fill` is used to fill the marked positions.

So in the hypothetical new functionality, the following:

```
<$transclude $filter="A B C">
...rest of page template...
</$transclude>
```
could be equivalent to something like this:

```html
<$transclude $tiddler=A>
  <$fill $name="some-well-known-name">
    <$transclude $tiddler=B>
      <$fill $name="some-well-known-name">
        <$transclude $tiddler=C>
          <$fill $name="some-well-known-name">
          ...rest of page template...
          </$fill>
        </$transclude>
      </$fill>
    </$transclude>
  </$fill>
</$transclude>
```

The A, B, and C tiddlers presumably would each have `<$slot $name="some-well-known-name"/>` somewhere inside them to mark the insertion position.

btheado:
> So in the hypothetical new functionality, the following: [...] could be equivalent to something like this [...]

I wrote a rough version of that hypothetical functionality using a recursive custom widget called `$$nested-transclude`. For the "well-known" slot name, I just used the already defined `ts-raw`.

Here is the code:

```
\widget $$nested-transclude(filter, n:1)
<$list filter="[subfilter<filter>nth<n>]" emptyMessage="<$slot $name=ts-raw $depth={{{[<n>multiply[2]subtract[1]]}}}/>">

<$transclude $tiddler=<<currentTiddler>>>

<$$nested-transclude filter=<<filter>> n={{{[<n>add[1]]}}}/>
</$transclude>
</$list>
\end

<$$nested-transclude filter="A B =A">
...rest of page template...
</$$nested-transclude>
```

This way of implementing recursion feels a little clumsy and inefficient to me. With every recursive step it re-evaluates the entire filter and walks one step further in the results using `nth`.  The recursion "stop" condition is triggered when n goes beyond the end of the filter results.  Maybe there are better ways to do recursion?

My test A and B tiddlers look like this:

```
AAA
<$slot $name=ts-raw/>
/AAA
```

```
BBB
<$slot $name=ts-raw/>
/BBB
```

The output looks like this:

```
AAA

BBB

AAA

...rest of page template...
/AAA
/BBB
/AAA
```

I'm thinking the `<$slot $name=ts-raw $depth={{{[<n>multiply[2]subtract[1]]}}}/>` emptyMessage might break if tiddlers A or B themselves introduce another level of transclusion.

You can play with the above functionality at this [share site link](https://tiddlywiki.com/prerelease/share##%5B%7B%22created%22%3A%2220230425013312315%22%2C%22title%22%3A%22%24%3A%2Fstate%2Fshoweditpreview%22%2C%22text%22%3A%22yes%22%2C%22modified%22%3A%2220230425013312315%22%7D%2C%7B%22created%22%3A%2220230425013317955%22%2C%22title%22%3A%22%24%3A%2Fstate%2Fsidebar%22%2C%22text%22%3A%22no%22%2C%22modified%22%3A%2220230425013317955%22%7D%2C%7B%22created%22%3A%2220230425013303292%22%2C%22title%22%3A%22%24%3A%2FStoryList%22%2C%22text%22%3A%22%22%2C%22list%22%3A%22A%20B%20%5B%5BDraft%20of%20%27nested-transclude%27%5D%5D%22%2C%22modified%22%3A%2220230425013332379%22%7D%2C%7B%22created%22%3A%2220230425010754711%22%2C%22text%22%3A%22AAA%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FAAA%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22A%22%2C%22modified%22%3A%2220230425012821112%22%7D%2C%7B%22created%22%3A%2220230425010831153%22%2C%22text%22%3A%22BBB%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FBBB%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22B%22%2C%22modified%22%3A%2220230425012843797%22%7D%2C%7B%22text%22%3A%22%5C%5Cwidget%20%24%24nested-transclude%28filter%2C%20n%3A1%29%5Cn%3C%24list%20filter%3D%5C%22%5Bsubfilter%3Cfilter%3Enth%3Cn%3E%5D%5C%22%20emptyMessage%3D%5C%22%3C%24slot%20%24name%3Dts-raw%20%24depth%3D%7B%7B%7B%5B%3Cn%3Emultiply%5B2%5Dsubtract%5B1%5D%5D%7D%7D%7D%2F%3E%5C%22%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%3C%3CcurrentTiddler%3E%3E%3E%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%3C%3Cfilter%3E%3E%20n%3D%7B%7B%7B%5B%3Cn%3Eadd%5B1%5D%5D%7D%7D%7D%2F%3E%5Cn%3C%2F%24transclude%3E%5Cn%3C%2F%24list%3E%5Cn%5C%5Cend%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%5C%22A%20B%20%3DA%5C%22%3E%5Cn...rest%20of%20page%20template...%5Cn%3C%2F%24%24nested-transclude%3E%22%2C%22created%22%3A%2220230424212905248%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22Draft%20of%20%27nested-transclude%27%22%2C%22modified%22%3A%2220230425013309371%22%2C%22draft.title%22%3A%22nested-transclude%22%2C%22draft.of%22%3A%22nested-transclude%22%7D%2C%7B%22created%22%3A%2220230424212905248%22%2C%22text%22%3A%22%5C%5Cwidget%20%24%24nested-transclude%28filter%2C%20n%3A1%29%5Cn%3C%24list%20filter%3D%5C%22%5Bsubfilter%3Cfilter%3Enth%3Cn%3E%5D%5C%22%20emptyMessage%3D%5C%22%3C%24slot%20%24name%3Dts-raw%20%24depth%3D%7B%7B%7B%5B%3Cn%3Emultiply%5B2%5Dsubtract%5B1%5D%5D%7D%7D%7D%2F%3E%5C%22%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%3C%3CcurrentTiddler%3E%3E%3E%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%3C%3Cfilter%3E%3E%20n%3D%7B%7B%7B%5B%3Cn%3Eadd%5B1%5D%5D%7D%7D%7D%2F%3E%5Cn%3C%2F%24transclude%3E%5Cn%3C%2F%24list%3E%5Cn%5C%5Cend%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%5C%22A%20B%20%3DA%5C%22%3E%5Cn...rest%20of%20page%20template...%5Cn%3C%2F%24%24nested-transclude%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22nested-transclude%22%2C%22modified%22%3A%2220230425012929228%22%7D%5D)

btheado:
> I wrote a rough version of that hypothetical functionality using a recursive custom widget called $$nested-transclude.

For a lot of cases, including this one, the parameterized transclusion functionality can achieve the same results as custom widgets.

When placed in a tiddler named "transclude wrapper", the follow can give the same results as the custom widget code I previously shared:

```
\parameters(filter, n:1)
<$list filter="[subfilter<filter>nth<n>]" emptyMessage="<$slot $name=ts-raw $depth={{{[<n>multiply[2]subtract[1]]}}}/>">

<$transclude $tiddler=<<currentTiddler>>>

<$transclude $tiddler="transclude wrapper" filter=<<filter>> n={{{[<n>add[1]]}}}/>
</$transclude>
</$list>
```
The above can be invoked like this:

```
<$transclude $tiddler="transclude wrapper" filter="A B =A">
...rest of stuff...
</$transclude>
```
Demo of both of these implementations at this [share site link](https://tiddlywiki.com/prerelease/share##%5B%7B%22created%22%3A%2220230501170455003%22%2C%22title%22%3A%22%24%3A%2FStoryList%22%2C%22text%22%3A%22%22%2C%22list%22%3A%22%5B%5Btransclude%20wrapper%20demo%5D%5D%20%5B%5Bnested-transclude%20widget%20demo%5D%5D%22%2C%22modified%22%3A%2220230501174833226%22%7D%2C%7B%22created%22%3A%2220230425010754711%22%2C%22text%22%3A%22AAA%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FAAA%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22A%22%2C%22modified%22%3A%2220230501170648175%22%7D%2C%7B%22created%22%3A%2220230425010831153%22%2C%22text%22%3A%22BBB%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FBBB%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22B%22%2C%22modified%22%3A%2220230425012843797%22%7D%2C%7B%22created%22%3A%2220230424212905248%22%2C%22text%22%3A%22%5C%5Cwidget%20%24%24nested-transclude%28filter%2C%20n%3A1%29%5Cn%3C%24list%20filter%3D%5C%22%5Bsubfilter%3Cfilter%3Enth%3Cn%3E%5D%5C%22%20emptyMessage%3D%5C%22%3C%24slot%20%24name%3Dts-raw%20%24depth%3D%7B%7B%7B%5B%3Cn%3Emultiply%5B2%5Dsubtract%5B1%5D%5D%7D%7D%7D%2F%3E%5C%22%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%3C%3CcurrentTiddler%3E%3E%3E%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%3C%3Cfilter%3E%3E%20n%3D%7B%7B%7B%5B%3Cn%3Eadd%5B1%5D%5D%7D%7D%7D%2F%3E%5Cn%3C%2F%24transclude%3E%5Cn%3C%2F%24list%3E%5Cn%5C%5Cend%5Cn%5Cn%3C%24%24nested-transclude%20filter%3D%5C%22A%20B%20%3DA%5C%22%3E%5Cn...rest%20of%20page%20template...%5Cn%3C%2F%24%24nested-transclude%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22nested-transclude%22%2C%22modified%22%3A%2220230501173426760%22%7D%2C%7B%22created%22%3A%2220230501173150913%22%2C%22text%22%3A%22%5C%5Cdefine%20c%28t%29%5Cn%21%21%20%3C%3C__t__%3E%3E%5Cn%3C%24tiddler%20tiddler%3D%3C%3C__t__%3E%3E%3E%3C%24transclude%20%24tiddler%3D%5C%22%24%3A%2Fcore%2Fui%2FViewTemplate%2Fbody%2Fcode%5C%22%2F%3E%3C%2F%24tiddler%3E%5Cn%5C%5Cend%5Cn%21%20Rendered%20output%5Cn%7B%7Bnested-transclude%7D%7D%5Cn%21%20Code%5Cn%3C%3Cc%20%5C%22nested-transclude%5C%22%3E%3E%5Cn%3C%3Cc%20%5C%22A%5C%22%3E%3E%5Cn%3C%3Cc%20%5C%22B%5C%22%3E%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22nested-transclude%20widget%20demo%22%2C%22modified%22%3A%2220230501173436409%22%7D%2C%7B%22created%22%3A%2220230430015149540%22%2C%22text%22%3A%22%5C%5Cparameters%28filter%2C%20n%3A1%29%5Cn%3C%24list%20filter%3D%5C%22%5Bsubfilter%3Cfilter%3Enth%3Cn%3E%5D%5C%22%20emptyMessage%3D%5C%22%3C%24slot%20%24name%3Dts-raw%20%24depth%3D%7B%7B%7B%5B%3Cn%3Emultiply%5B2%5Dsubtract%5B1%5D%5D%7D%7D%7D%2F%3E%5C%22%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%3C%3CcurrentTiddler%3E%3E%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%5C%22transclude%20wrapper%5C%22%20filter%3D%3C%3Cfilter%3E%3E%20n%3D%7B%7B%7B%5B%3Cn%3Eadd%5B1%5D%5D%7D%7D%7D%2F%3E%5Cn%3C%2F%24transclude%3E%5Cn%3C%2F%24list%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22transclude%20wrapper%22%2C%22modified%22%3A%2220230501172741943%22%7D%2C%7B%22created%22%3A%2220230501170827840%22%2C%22text%22%3A%22%5C%5Cdefine%20c%28t%29%5Cn%21%21%20%3C%3C__t__%3E%3E%5Cn%3C%24tiddler%20tiddler%3D%3C%3C__t__%3E%3E%3E%3C%24transclude%20%24tiddler%3D%5C%22%24%3A%2Fcore%2Fui%2FViewTemplate%2Fbody%2Fcode%5C%22%2F%3E%3C%2F%24tiddler%3E%5Cn%5C%5Cend%5Cn%21%20Rendered%20output%5Cn%7B%7Btransclude%20wrapper%20test%7D%7D%5Cn%21%20Code%5Cn%3C%3Cc%20%5C%22transclude%20wrapper%20test%5C%22%3E%3E%5Cn%3C%3Cc%20%5C%22transclude%20wrapper%5C%22%3E%3E%5Cn%3C%3Cc%20%5C%22A%5C%22%3E%3E%5Cn%3C%3Cc%20%5C%22B%5C%22%3E%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22transclude%20wrapper%20demo%22%2C%22modified%22%3A%2220230501172833019%22%7D%2C%7B%22created%22%3A%2220230430015447274%22%2C%22text%22%3A%22%3C%24transclude%20%24tiddler%3D%5C%22transclude%20wrapper%5C%22%20filter%3D%5C%22A%20B%20%3DA%5C%22%3E%5Cn...rest%20of%20stuff...%5Cn%3C%2F%24transclude%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22transclude%20wrapper%20test%22%2C%22modified%22%3A%2220230501170623894%22%7D%5D).

yaisog:
Hi @btheado, I've finally managed to wrap my head around your code. Seems that some of the syntax you used is not (yet) documented, which made it a bit harder (e.g. the `$depth` attribute of `$slot` and that `ts-raw` need not be defined via a `$fill`). But now that I figured it out, I have to say that it is pretty awesome WikiText.

I changed it a little bit so that the wrapper is defined in a macro, nay, procedure, which could be made global if these widget insertion points were to be sprinkled throughout.

It sure would be nice if one didn't need the `$depth` attribute when the `$slot`/`$fill` IDs are unique, i.e. that they are passed down to the next level of transclusion if they are not redefined, as is usually the case with `ts-raw`. Then, the actual content could be wrapped in such a unique `$fill` and used in the emptyMessage without needing to worry about the correct `$depth`.

Anyway, here's my current working code, for reference:
```
\procedure dynamic-insert(filter, depth:1)
<$list filter="[subfilter<filter>nth<depth>]" emptyMessage="""<$slot $name="ts-raw" $depth={{{ [<depth>multiply[2]subtract[1]] }}} />""">
	<$transclude $tiddler=<<currentTiddler>> $mode="block">
		<$transclude $variable="dynamic-insert" filter=<<filter>> depth={{{ [<depth>add[1]] }}} $mode="block"/>
	</$transclude>
</$list>
\end

<$transclude $variable="dynamic-insert" filter="[tag[$:/tags/TranscludeTag]]">
	...actual content...
</$transclude>
```
where `TranscludeTag` would be defined according to the location, e.g. `PageTemplateWidgets` or something.
The widget-to-be-inserted is then defined in a tiddler with the respective tag, like so:
```
<$messagecatcher $tm-notify="""<$action-sendmessage $message="tm-notify" $param="SampleNotification"/>""" >
	<$slot $name=ts-raw />
</$messagecatcher>
```
After all, this would be a pretty nice method to wrap an arbitrary number of widget around any piece of a WikiText template. I wonder what @saqimtiaz or @Jermolene think about it.

btheado:
> I changed it a little bit so that the wrapper is defined in a macro, nay, procedure, which could be made global if these widget insertion points were to be sprinkled throughout.

Yeah, I'm not sure how to decide between tiddler or variable transclude. The advantage to the tiddler transclude is that you don't need to do anything to make it global. It is automatically global. But that can also be a disadvantage.

Here is a [share site url with and alternate page template](https://tiddlywiki.com/prerelease/share##%5B%7B%22created%22%3A%2220230430234258391%22%2C%22title%22%3A%22%24%3A%2Flayout%22%2C%22text%22%3A%22alternate%20page%20template%22%2C%22modified%22%3A%2220230430234504482%22%7D%2C%7B%22created%22%3A%2220230504211914533%22%2C%22title%22%3A%22%24%3A%2FStoryList%22%2C%22text%22%3A%22%22%2C%22list%22%3A%22%5B%5Balternate%20page%20template%5D%5D%20%5B%5BLink%20click%20move%20to%20top%5D%5D%20A%20B%22%2C%22modified%22%3A%2220230504214123424%22%7D%2C%7B%22created%22%3A%2220230425010754711%22%2C%22text%22%3A%22AAA%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FAAA%22%2C%22tags%22%3A%22%5B%5Balternate%20page%20template%5D%5D%22%2C%22title%22%3A%22A%22%2C%22modified%22%3A%2220230430234607660%22%7D%2C%7B%22created%22%3A%2220230430233603455%22%2C%22text%22%3A%22%5C%5Cwhitespace%20trim%5Cn%5C%5Cimport%20%5B%5B%24%3A%2Fcore%2Fui%2FPageMacros%5D%5D%20%5Ball%5Bshadows%2Btiddlers%5Dtag%5B%24%3A%2Ftags%2FMacro%5D%21has%5Bdraft.of%5D%5D%5Cn%5Cn%3C%24vars%5Cn%5Cttv-config-toolbar-icons%3D%7B%7B%24%3A%2Fconfig%2FToolbar%2FIcons%7D%7D%5Cn%5Cttv-config-toolbar-text%3D%7B%7B%24%3A%2Fconfig%2FToolbar%2FText%7D%7D%5Cn%5Cttv-config-toolbar-class%3D%7B%7B%24%3A%2Fconfig%2FToolbar%2FButtonClass%7D%7D%5Cn%5Cttv-enable-drag-and-drop%3D%7B%7B%24%3A%2Fconfig%2FDragAndDrop%2FEnable%7D%7D%5Cn%5Cttv-show-missing-links%3D%7B%7B%24%3A%2Fconfig%2FMissingLinks%7D%7D%5Cn%5CtstoryviewTitle%3D%7B%7B%24%3A%2Fview%7D%7D%5Cn%5CtlanguageTitle%3D%7B%7B%7B%20%5B%7B%24%3A%2Flanguage%7Dget%5Bname%5D%5D%20%7D%7D%7D%3E%5Cn%5Cn%3Cdiv%20class%3D%7B%7B%7B%20%5Ball%5Bshadows%2Btiddlers%5Dtag%5B%24%3A%2Ftags%2FClassFilters%2FPageTemplate%5D%21is%5Bdraft%5D%5D%20%3Amap%3Aflat%5Bsubfilter%7B%21%21text%7D%5D%20tc-page-container%20%5B%5Btc-page-view-%5Daddsuffix%3CstoryviewTitle%3E%5D%20%5B%5Btc-language-%5Daddsuffix%3ClanguageTitle%3E%5D%20%3Aand%5Bunique%5B%5Djoin%5B%20%5D%5D%20%7D%7D%7D%20%3E%5Cn%5Cn%3C%24navigator%20story%3D%5C%22%24%3A%2FStoryList%5C%22%20history%3D%5C%22%24%3A%2FHistoryList%5C%22%20openLinkFromInsideRiver%3D%7B%7B%24%3A%2Fconfig%2FNavigation%2FopenLinkFromInsideRiver%7D%7D%20openLinkFromOutsideRiver%3D%7B%7B%24%3A%2Fconfig%2FNavigation%2FopenLinkFromOutsideRiver%7D%7D%20relinkOnRename%3D%7B%7B%24%3A%2Fconfig%2FRelinkOnRename%7D%7D%3E%5Cn%5Cn%3C%24dropzone%20enable%3D%3C%3Ctv-enable-drag-and-drop%3E%3E%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%5C%22transclude%20wrapper%5C%22%20filter%3D%5C%22%5Btag%5Balternate%20page%20template%5D%21has%5Bdraft.of%5D%5D%5C%22%3E%5Cn%3C%24list%20filter%3D%5C%22%5Ball%5Bshadows%2Btiddlers%5Dtag%5B%24%3A%2Ftags%2FPageTemplate%5D%21has%5Bdraft.of%5D%5D%5C%22%20variable%3D%5C%22listItem%5C%22%3E%5Cn%5Cn%3C%24transclude%20tiddler%3D%3C%3ClistItem%3E%3E%2F%3E%5Cn%5Cn%3C%2F%24list%3E%5Cn%5Cn%3C%2F%24transclude%3E%5Cn%5Cn%3C%2F%24dropzone%3E%5Cn%5Cn%3C%2F%24navigator%3E%5Cn%5Cn%3C%2Fdiv%3E%5Cn%5Cn%3C%2F%24vars%3E%5Cn%22%2C%22title%22%3A%22alternate%20page%20template%22%2C%22name%22%3A%22transclude%20wrapper%20test%22%2C%22description%22%3A%22transclude%20wrapper%20test%22%2C%22icon%22%3A%22%24%3A%2Fcore%2Fimages%2Flayout-button%22%2C%22tags%22%3A%22%24%3A%2Ftags%2FLayout%22%2C%22modified%22%3A%2220230504212116729%22%2C%22code-body%22%3A%22yes%22%7D%2C%7B%22created%22%3A%2220230425010831153%22%2C%22text%22%3A%22BBB%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%2FBBB%22%2C%22tags%22%3A%22%5B%5Balternate%20page%20template%5D%5D%22%2C%22title%22%3A%22B%22%2C%22modified%22%3A%2220230504212035257%22%7D%2C%7B%22created%22%3A%2220230504214002011%22%2C%22text%22%3A%22%3C%24linkcatcher%20message%3D%5C%22tm-navigate%5C%22%20actions%3D%5C%22%5C%22%5C%22%5Cn%3C%24action-listops%20%24tiddler%3D%5C%22%24%3A%2FStoryList%5C%22%20%24subfilter%3D%5C%22%5B%3CnavigateTo%3E%5D%20%2B%5Bputfirst%5B%5Dlimit%5B15%5D%5D%5C%22%2F%3E%5Cn%5C%22%5C%22%5C%22%3E%5Cn%3C%24slot%20%24name%3Dts-raw%2F%3E%5Cn%3C%2F%24linkcatcher%3E%22%2C%22tags%22%3A%22%5B%5Balternate%20page%20template%5D%5D%22%2C%22title%22%3A%22Link%20click%20move%20to%20top%22%2C%22modified%22%3A%2220230504214048093%22%7D%2C%7B%22created%22%3A%2220230430015149540%22%2C%22text%22%3A%22%5C%5Cparameters%28filter%2C%20n%3A1%29%5Cn%3C%24list%20filter%3D%5C%22%5Bsubfilter%3Cfilter%3Enth%3Cn%3E%5D%5C%22%20emptyMessage%3D%5C%22%3C%24slot%20%24name%3Dts-raw%20%24depth%3D%7B%7B%7B%5B%3Cn%3Emultiply%5B2%5Dsubtract%5B1%5D%5D%7D%7D%7D%2F%3E%5C%22%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%3C%3CcurrentTiddler%3E%3E%3E%5Cn%5Cn%3C%24transclude%20%24tiddler%3D%5C%22transclude%20wrapper%5C%22%20filter%3D%3C%3Cfilter%3E%3E%20n%3D%7B%7B%7B%5B%3Cn%3Eadd%5B1%5D%5D%7D%7D%7D%2F%3E%5Cn%3C%2F%24transclude%3E%5Cn%3C%2F%24list%3E%22%2C%22tags%22%3A%22%22%2C%22title%22%3A%22transclude%20wrapper%22%2C%22modified%22%3A%2220230430115621174%22%7D%5D) defined which makes use of my tiddler transclusion version. Any tiddler tagged "alternate page template" is used for wrapping. I include 3 example wrapping tiddlers. Two just insert text at the top and bottom and the third intercepts link clicks to move the clicked tiddler to the top of the storylist.

yaisog:
> I think it would be prudent to explore a JS implementation and compare performance before making a decision.

Hi @saqimtiaz, thank you for the kind words. I am absolutely not opposed to a comparison with a JS implementation (by someone else). However, I'm currently not very worried about performance. For every insertion point and any tiddler with widgets that is inserted, about one filter evaluation is necessary. Typically, that would be the `tag[]` filter which by now is lightning fast. So even sprinkling the necessary TranscludeWidgets throughout the templates should not generate an appreciable overhead -- if we don't overdo it. But, it wouldn't be the first time I'm wrong about such things... 😉 

Be that as it may, it would be nice to agree beforehand on the format that the tiddlers with the injected widgets should have, though. This way, we could implement the WikiText solution now, and later seemlessly switch to the JS version if that makes it into the core instead, without rewriting the insertions. As a guideline for my version, I used your example way above where a tiddler with a `$set` widget is imported by `$importvariables`, basically only adding the necessary `$slot`. @btheado's version is practically identical. I don't even know if there is another sensible way. It should be possible to add local macros/procedures to such tiddlers, though I haven't tested that yet.

yaisog:
Maybe someone with more JS development knowledge could do some timing tests with the current WikiText implementation?
I can offer to measure some timings with Maurycy's _Advanced Performance_ plugin in a rather large wiki with tons of tags, once I implement this WikiText solution, but I don't really know what to do with flame graphs and whatnot.

saqimtiaz:
> Be that as it may, it would be nice to agree beforehand on the format that the tiddlers with the injected widgets should have, though. This way, we could implement the WikiText solution now, and later seemlessly switch to the JS version if that makes it into the core instead

We should be able to explore and compare the two options before the release of v5.3.0

yaisog:
So, I put three MessagecatcherWidgets and a double EventcatcherWidget in 4 tiddlers that are dynamically inserted into the PageTemplate as described above, and then navigated around in the wiki. _Advanced Performance_ was not able to measure the filter time for this, as the plugin is probably loaded after the PageTemplate is initialized, and the PageTemplate is never refreshed during regular navigation. So the overhead for that is indeed negligible. I'll try later to put these in the ViewTemplate (though they don't belong there) to measure _something_.

I did notice, though, that when I forgot to put in the `ts-raw` slot (because I modified pre-existing code and forgot), the whole wiki is immediately bricked, unless you have a secondary PageTemplate to change back to. I wonder if we should somehow check for the presence of the slot in the transcluded tiddlers. We cannot make 100% sure that it is put in correctly, but we can at least guard against forgetfulness. Or is this something that we should leave to the user?

btheado:
> the whole wiki is immediately bricked, unless you have a secondary PageTemplate to change back to

If PR  #7357 moves forward it might be helpful for recovering from this.

>  I wonder if we should somehow check for the presence of the slot in the transcluded tiddlers.

The body text of the slot widget is used as a fallback. Maybe something useful can be put there. Maybe a separate template tiddler containing the slot widget with fallback content which is generally useful for all cases? Then encourage authors who are creating these "wrapper tiddlers" to use this template instead of directly using the slot widget. Seems a little flimsy to me, but maybe it can help.

pmario:
>@jermolene wrote
>
> ... Stepping back, I think there might be an alternative, more generic way to handle this problem. I'd like to add a more generic version of the <$importvariables> widget that can import arbitrary parse tree nodes, not just <$set> widgets. We might call it "filtered transclusion":

Jeremy is there any news / thoughts on this one?

yaisog:
> The proposed functionality, the ability to import widgets/parsetrees is potentially powerful and would allow us to make core templates extensible in a manner that has previously only been possible by overriding shadow tiddlers. As such, it would be desirable that the technique be usable without worries about performance repercussions. A JavaScript widget that implemented this functionality would possibly be more performant, especially considering how many times the filter needs to be evaluated in the wikitext solution. I think it would be prudent to explore a JS implementation and compare performance before making a decision.

Hi @saqimtiaz, I recently updated my implementation to better make use of some of the v5.3 features as well as multi-valued variables from the extension PR #9487. The result is a quite simple custom widget that takes a filter expression as argument, evaluates this filter once and then recursively wrapes the filtered tiddlers around the its content:
```
\widget $filtered.transclude($filter)
	\procedure transcludeRecursively(transcludeList,depth:3)
		<$transclude $tiddler={{{ [title:test<transcludeList>first[]] }}} >
			<$transclude $variable="transcludeRecursively" transcludeList=((( [<transcludeList>butfirst[]] ))) depth={{{ [<depth>add[1]] }}} $fillignore="yes" />
			<$fill $name="ts-missing"><$slot $name="ts-raw" $depth=<<depth>> /></$fill>
		</$transclude>
	\end transcludeRecursively
	<$transclude $variable="transcludeRecursively" transcludeList=((( [subfilter<$filter>] ))) />
\end
```
In the _$:/core/ui/PageTemplate_ it would be used like so:
```
<$filtered.transclude $filter="[tag[$:/tags/PageTemplateImports]!has[draft.of]]" >
	<$list filter="[all[shadows+tiddlers]tag[$:/tags/PageTemplate]!has[draft.of]]" variable="listItem">
	</$list>
</$filtered.transclude>
```
to include all tiddlers with the tag _$:/tags/PageTemplateImports_, similar to what @Jermolene proposed in [his comment](https://github.com/TiddlyWiki/TiddlyWiki5/pull/7340#issuecomment-1480909807).

The included tiddlers only must have a SlotWidget with `<$slot $name=ts-raw />` where the content should be inserted.

Since the filter is only evaluated once, there should not be much to be gained by translating this to JavaScript. Also, it would be a neat example for custom widgets.

core/ui/PageTemplate/MessageHandler/tm-delete-tiddler.tid
```diff
@@ -0,0 +1,22 @@
...
+title: $:/core/ui/PageTemplate/MessageHandler/tm-delete-tiddler
+
+\define navigate-and-forward()
+\define tv-action-refresh-policy() always
+<$let targetTiddler={{{ [<event-param>!is[blank]else<event-tiddlerTitle>] }}} history={{{ [<tv-history-list>get[text]] }}}>
+    <!-- forward the original message -->
+    <$action-sendmessage $message=<<event-type>> $param=<<targetTiddler>> />
+    <!-- only go back to previous tiddler if delete was not canceled -->
+    <$list filter="[<targetTiddler>!is[tiddler]]" variable="void">
+        <$list filter="[<tv-history-list>get[current-tiddler]match<targetTiddler>]" variable="void">
```
pmario, line 12: Indentation should be TABs for all files

yaisog: Ah, right. But this is only a draft PR for exploratory purposes which I didn't lint. Also, this is the original, obsolete implementation which has been superseded as I wrote in the new PR description.

saqimtiaz:
Hi @yaisog, thank you for persisting with this work. The introduction of multi-valued variables does address most of my earlier concerns with a wikitext-based approach.

My remaining concern is around potential brittleness, and the risk of effectively bricking a wiki if `<$slot $name="ts-raw" />` is accidentally omitted from a transcluded tiddler. Am I correct in understanding that the fill widget you include is intended as a fallback for this scenario?

`<$fill $name="ts-missing"><$slot $name="ts-raw" $depth=<<depth>> /></$fill>`

If so, that helps mitigate the risk, but it is still something we should consider carefully.

A good next step would be to think about where dynamic wrapper transclusions should be permitted, and what the corresponding system tag names ought to be. I can imagine one around the story river, and perhaps another for the page template as a whole. Should we also consider one for the tiddler template itself?

Finally, we will need an alternative name for the custom widget, as "filtered transclusion" already has an established meaning in TiddlyWiki. Thanks again for the thoughtful iteration on this.

yaisog:
Hi @saqimtiaz, your concern is warranted. A missing `ts-raw` slot will brick a wiki. To mitigate this, we could at least check for the presence of this slot in the source code before transcluding a tiddler. This is not 100% foolproof, but probably helps. Also, we would probably create a new PageTemplate for this and leave the old one in. The layout switcher would still work, so there would be a way to recover. We might not even make the new PageTemplate the default, so that (power) users must switch this feature on explicitly. This way, nobody's wiki is bricked by an upgrade.

> Am I correct in understanding that the fill widget you include is intended as a fallback for this scenario?

I think that only guards against missing transcluded tiddlers (or an empty list).

> think about where dynamic wrapper transclusions should be permitted

I think the PageTemplate, the ViewTemplate and the EditTemplate might be natural candidates. The StoryRiver maybe, if you want to exclude the SideBar and Menus from PageTemplate. There shouldn't be too many, though, because of the brittleness. I make do with only the PageTemplate, since I can filter out sources within the transcluded tiddlers if need be. Putting it elsewhere also makes it harder to recover since there is no LayoutSwitcher equivalent for the other templates.

"Filtered transclusion" was picked up from a comment by @Jermolene, since that describes what it does. But I agree that this concept is used already for something else.

yaisog:
I think I've taken this thing as far as it'll go. Can't work around the missing fallback. There's probably no way around a JS widget for that.

saqimtiaz:
@yaisog I feel like with the availability of MVV's this solution is compelling enough that it seems unfortunate to need a custom JavaScript widget. Perhaps the way to address the potential brittleness of a missing fallback is through the safety hack of a recovery layout. See #7357. This should realistically be possible for v5.5.0.

PR #7349, opened 2023-03-09
Improve documentation for `get` and `getindex`

yaisog:
I apologize, but the filter originally given in the "tip" for `getindex` worked only correct with a single input tiddler, which I corrected. Also, I felt that the _Examples_ were a better container for this information, so I moved it there.

Since `get` and `getindex` operate very similar, I made their descriptions and examples much more congruent, and also added an example of how to optionally return an empty string with `get`.

I refined the wording and formatting to better align with the _Documentation Style Guide_ and make better use of the _Documentation Macros_.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/get.tid
+++ b/editions/tw5.com/tiddlers/filters/examples/get.tid
@@ -1,9 +1,19 @@
 created: 20150118134611000
-modified: 20150124201839000
+modified: 20230309180521925
+myfield: 
 tags: [[get Operator]] [[Operator Examples]]
 title: get Operator (Examples)
 type: text/vnd.tiddlywiki
 
 <<.operator-example 1 "[all[current]get[draft.of]]" "the title of the tiddler of which the current tiddler is a draft">>
-<<.operator-example 2 "[get[tags]]">>
-<<.operator-example 3 "[each[tags]get[tags]]">>
+
+<<.operator-example 2 "[get[tags]]" "returns the tags of all tiddlers without de-duplication">>
+
+<<.operator-example 3 "[get[tags]unique[]]" "returns the tags of all tiddlers with de-duplication">>
+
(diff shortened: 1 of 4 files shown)


kookma:
I think `[all[current]] :filter[has:field[myfield]] :map[get[myfield]]` in **get Operator (Examples)** is equivalent to `[all[current]get[myfield]]` I would recommend a better example here. 

It is very good to go beyond simple examples.

yaisog:
Hi @kookma, thanks for the response.
The difference between my example and `[all[current]get[myfield]]` is that in my example an empty string is returned when the field exists, but is empty, while `get[]` returns nothing / an empty list. That's exactly what this example was supposed to show.

I like your example, too. In my opinion, there can never be too many examples. Any opportunity for copy/pasting solutions is always a great help in my own "programming" experience. I'll add something inspired by your proposal later.

kookma:
> is that in my example an empty string is returned when the field exists, but is empty, while `get[]` returns nothing / an empty list.

Ah, I did not note to this! so go ahead!

Jermolene:
Thanks @yaisog @kookma

Jermolene:
Hi @yaisog yes I'm afraid so! Thank you.

yaisog:
Well @Jermolene, I guess we're lucky then that making PRs is my favorite pastime... 😁

saqimtiaz:
@yaisog for documentation PRs for existing features, you might find this useful: https://saqimtiaz.github.io/tw5-docs-pr-maker/

yaisog:
Hi @saqimtiaz, I've used your PR maker before. It's an awesome piece of software! 👍 

I just recently discovered GitHub Desktop and how nicely it integrates with Sublime Text, so I've been using that a bit. But for docs I'm definitely going back to the PR maker for its ease-of-use and preview abilities...

PR #7351, opened 2023-03-10
Add another example for the `get` operator

yaisog:
This example inspired by @kookma demonstrates two different methods of getting the value of a field within a filter expression.


Jermolene:
Thanks @yaisog

PR #7357, opened 2023-03-12
Safe mode layout

Jermolene:
I came across this draft PR from a couple of years ago which seems relevant to some recent discussions about "safe mode".

This PR introduces a new safe mode layout that essentially consists of the tiddler manager (`$:/Manager`) plus the page buttons, and the layout switcher. The tiddler manager is extended to provide an editor and a "delete" button.

The new layout is engaged automatically with safe mode. The idea is that it gives users enough functionality to be able to edit or delete problematic tiddlers and then save the wiki.

image

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -1542,6 +1542,8 @@ $tw.Wiki.prototype.processSafeMode = function() {
 	this.addTiddler(new $tw.Tiddler({title: titleReportTiddler, text: report.join("\n\n")}));
 	// Set $:/DefaultTiddlers to point to our report
 	this.addTiddler(new $tw.Tiddler({title: "$:/DefaultTiddlers", text: "[[" + titleReportTiddler + "]]"}));
+	// Switch to the safe mode layout
+	this.addTiddler({title: "$:/layout", text: "$:/core/ui/SafeLayout"});
 };
 
 /*
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -56,6 +56,7 @@ Manager/Controls/Show/Option/Tiddlers: tiddlers
 Manager/Controls/Show/Prompt: Show:
 Manager/Controls/Sort/Prompt: Sort by:
 Manager/Item/Colour: Colour
+Manager/Item/Editor: Editor
 Manager/Item/Fields: Fields
 Manager/Item/Icon/None: (none)
 Manager/Item/Icon: Icon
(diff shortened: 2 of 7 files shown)


pmario:
Will this layout be always visible in the "layout chooser" or is it only active if safe-mode is active?

Jermolene:
> Will this layout be always visible in the "layout chooser" or is it only active if safe-mode is active?

At the moment the safe layout is always visible in the layout chooser. I think that makes sense because it's useful to be able to use it without engaging all of the features of safe mode. It is possible that it should be renamed to reflect that it is more general purpose, perhaps "Simple Layout" or "Minimal Layout".

pmario:
>..perhaps "Simple Layout" or "Minimal Layout".

From my point of view it is OK, that it's always visible. ... We do have some requests at Talk, which request "batch processing" from time to time. So this layout may help users with the most basic tasks.

linonetwo:
Can there be a standard to open a layout with URL query param? It's useful for launcher apps (Tiddloid TidGi) to open a wiki in safe mode layout or calendar layout directly.

I tried to add a button to open wiki with `#:safe` on TidGi desktop before, and it was not always working, so I removed it later.

Jermolene:
> Can there be a standard to open a layout with URL query param? It's useful for launcher apps (Tiddloid TidGi) to open a wiki in safe mode layout or calendar layout directly.

Hi @linonetwo TiddlyWiki doesn't use query parameters because they are reserved for the server. But we could use the location hash which would require the ability to customise the handling of the location hash. We've discussed that elsewhere and it remains a desirable enhancement.

However, for the case of launcher/hoster apps perhaps a more flexible approach would be to be able to use `window.postMessage()` to inject tiddlers directly into the hosted wiki.

linonetwo:
> inject tiddlers directly into the hosted wiki.

Thanks, this will work, I can listen on `tidgi://` scheme and onReceive I use API like `injectedJavaScript` `executeJavascript` to run `$tw.wiki.addTiddler('$:/layout', xxx)`. This will work after safe mode become a layout.

saqimtiaz:
Suggest renaming to Recovery Layout.

Jermolene:
> Suggest renaming to Recovery Layout.

That's much more self explanatory. However, `..#:recovery` is a bit clumsier than `..#:safe`. Maybe we could adopt something like `..#:rm` (for "recovery mode") as a shortcut alongside the full spelling. More controversially, we could use `..##` (without anything following the double hash) and still retain`..##...` (with a string after the double hash) for future expansion.

saqimtiaz:
We don't necessarily need to rename safe mode, rather just that the layout should have a name that indicates in the layout chooser modal that it has utility outside of safe mode, hence "recovery layout"

Issue #7348, opened 2023-03-09
[BUG] let vs vars if is-variable is evaluated in a filter


pmario:
### Describe the bug

To reproduce $let-widget does the wrong thing  -- $vars works as expected

let-widget always shows aa even if "default" is _not_ a variable

```
<$let default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$let>

<$vars default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$vars>
```

### Screenshots

![image](https://user-images.githubusercontent.com/374655/224071683-063026a0-007d-405d-9c87-14ce581aad67.png)

### TiddlyWiki Configuration

v5.2.5-prerelease

### Additional context

Related to: [BUG] editions/server-external-js is broken #7346 

The following code will always set coreURL to an empty string if coreURL does not exist

https://github.com/Jermolene/TiddlyWiki5/blob/b16d6953863c144bcae63374bf779bcb6a6bc78b/core/templates/external-js/save-offline-external-js.tid#L12

pmario:
@Jermolene .. I did update the OP .. had a copy paste problem

kookma:
Hi Mario,
For me it works with no error. TW 5.2.5

```
\define default() hi

<$let default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$let>

<$vars default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$vars>
```

produces

aa

aa

pmario:
That's what I wrote `\define default() hi` is _not_ allowed

Jermolene:
I think @pmario is pointing out that there is an inconsistency if the variable "default" is not defined:

```
<$let default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$let>

<$vars default={{{ [[default]is[variable]then[aa]else[bb]] }}} >
<<default>>
</$vars>
```

The result in v5.2.6 is certainly anomalous:

```
aa

bb
```

What is going on is that the let widget allows assignments to reference the values of variables that are set by previous attributes on the same let widget. This is what makes it possible to swap two variables like this:

```
\define showvalues() (<$text text=<<a>>/>|<$text text=<<b>>/>)

<$let a="elephant" b="giraffe">

<<showvalues>>

<$let tmp=<<a>> a=<<b>> b=<<tmp>>>

[...]

Issue #7346, opened 2023-03-09
[BUG] editions/server-external-js is broken


pmario:
### Describe the bug

@Jermolene ... PR https://github.com/Jermolene/TiddlyWiki5/commit/bef11fe6a25fb849dee40c4aa4337d6a30daf0b4 breaks 5.2.6 prerelease

To replicate: 

```
tiddlywiki ./myNewWiki --init server-external-js
tiddlywiki ./myNewWiki --build listen
```

![image](https://user-images.githubusercontent.com/374655/223953332-230617f1-a19d-45dd-b20b-836cd5ee8dc0.png)

That's how the index.html looks like.

```
<!--~~ Load external JavaScripts ~~-->
<script src="" onerror="alert('Error: Cannot load ');"></script>           <-- src="" .. 
</body>
</html>
```

That's how the template looks like `$:/core/templates/tiddlywiki5-external-js.html`

```
<!--~~ Load external JavaScripts ~~-->
<script src="`{{{ [<coreURL>] }}}`" onerror="alert('Error: Cannot load `{{{ [<coreURL>] }}}`');"></script>
</body>
</html>`
</$set>
```

--------

I'm experimenting atm. .. More info will follow

cdruan:
This issue is fixed by commit 007b759 and can be closed.

Jermolene:
Thanks @cdruan

Issue #7345, opened 2023-03-09
Tickets to be reviewed/merged after release of v5.2.7


Jermolene:
We've now entered the freeze period for v5.2.6 and so are no longer merging feature PRs (we will however continue to merge bug fixes for the new features included in v5.2.6).

These are tickets that will need reviewing/merging after v5.2.6 is released:

* #7253 
* #7260
* #7361
* #7215
* #7182
* #7365
* #7292
* #7332
* #7380
* #7378 
* #7400
* #7401 
* #4541
* #7409
* #7413 
* #4624
* #7422
* #6699 

See #7279 for the roadmap for the v5.2.6 and subsequent releases.

saqimtiaz:
We should consider merging #7297 after making a decision about ECMAScript version we will support from v5.3.0 onwards (#7353),

linonetwo:
How about https://github.com/Jermolene/TiddlyWiki5/pull/6699

pmario:
@Jermolene ... this PR has been delayed since 5.2.6 --  **make import listing more configurable #7308** .. It would be nice to give it a new review. 

I've updated the PR comments with more info

pmario:
IMO this one should be considered for a review:  **table-of-content macros -- make "exclude" an official macro parameter #7417**

saqimtiaz:
https://github.com/Jermolene/TiddlyWiki5/pull/7468 should be reviewed and merged.

Jermolene:
All the listed tickets have now been merged (including #7468).

PR #7361, opened 2023-03-14
external-js file naming adjusted. import docs form tw5.com. make extrnal-js more visible

pmario:
This PR fixes the issue: **[BUG] server-external-js does not have enough information to crate a working snapshot** #7366 

---

<details><summary>This PR does</summary>

- server-external-js -- tiddlywiki.info
  - renames index.html to `external-<version>.html`, to make it clear that there is something special going on
  - removed the highlight plugin .. The HTML file should be as small as possible
- imported "Using the external JavaScript template" from tw5-com -> **no duplication**
  - This tiddler also allows users to save the external core using the green button
- $:/DefaultTiddlers shows , GettingStarted and "Using the external JavaScript template" at startup

I did test

```
tiddlywiki ./myNewWiki --init server-external-js
tiddlywiki ./myNewWiki --build listen
```

- change some tiddlers
- reload page to see if they are saved -> OK
- Use the green-button to download `tiddlywikicore-5.2.6-prerelease.js`
- Use the "save snapshot" action to download `external-5.2.6-prerelease.html`
  - Open the .html in the browser
  - change some tiddlers
  - save
  - reload -> works

</details>

### DIFF
--- a/core/templates/external-js/save-all-external-js.tid
+++ b/core/templates/external-js/save-all-external-js.tid
@@ -5,8 +5,8 @@ title: $:/core/save/all-external-js
 \define saveTiddlerFilter()
 [is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/core]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
 \end
-<!-- It's important to fully understand this PR https://github.com/Jermolene/TiddlyWiki5/pull/5570 -->
-<!-- and what tiddlers: $:/core/templates/tiddlywiki5.js and $:/core/templates/tiddlywiki5.js/tiddlers do -->
+
+<!-- Important: core library is provided by serving URI encoded $:/core/templates/tiddlywiki5.js -->
 \define defaultCoreURL() %24%3A%2Fcore%2Ftemplates%2Ftiddlywiki5.js
 
 <$let coreURL={{{ [[coreURL]is[variable]then<coreURL>else<defaultCoreURL>] }}}>
--- a/core/templates/external-js/save-all-external-js.tid
+++ b/core/templates/external-js/save-all-external-js.tid
@@ -5,7 +5,10 @@ title: $:/core/save/all-external-js
 \define saveTiddlerFilter()
 [is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/core]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
 \end
+<!-- It's important to fully understand this PR https://github.com/Jermolene/TiddlyWiki5/pull/5570 -->
+<!-- and what tiddlers: $:/core/templates/tiddlywiki5.js and $:/core/templates/tiddlywiki5.js/tiddlers do -->
 \define defaultCoreURL() %24%3A%2Fcore%2Ftemplates%2Ftiddlywiki5.js
+
 <$let coreURL={{{ [[coreURL]is[variable]then<coreURL>else<defaultCoreURL>] }}}>
 	{{$:/core/templates/tiddlywiki5-external-js.html}}
 </$let>
--- a/editions/server-external-js/tiddlers/external/tiddlywiki.files
+++ b/editions/server-external-js/tiddlers/external/tiddlywiki.files
@@ -1,0 +1,8 @@
+{
+	"tiddlers": [
+		{
+			"file": "../../../tw5.com/tiddlers/webserver/Using the external JavaScript template.tid",
+			"isTiddlerFile": true
+		}
+	]
+}
--- a/editions/server-external-js/tiddlers/system/$__SiteTitle.tid
+++ b/editions/server-external-js/tiddlers/system/$__SiteTitle.tid
@@ -1,0 +1,6 @@
+created: 20230314153323056
+modified: 20230314153504240
+title: $:/SiteTitle
+type: text/vnd.tiddlywiki
+
+{{$:/config/SaveWikiButton/Filename}}
--- a/editions/server-external-js/tiddlers/system/$__SiteTitle.tid
+++ b/editions/server-external-js/tiddlers/system/$__SiteTitle.tid
@@ -1,6 +1,0 @@
-created: 20230314153323056
-modified: 20230314153504240
-title: $:/SiteTitle
-type: text/vnd.tiddlywiki
-
-{{$:/config/SaveWikiButton/Filename}}
--- a/editions/server-external-js/tiddlywiki.info
+++ b/editions/server-external-js/tiddlywiki.info
@@ -2,8 +2,7 @@
 	"description": "Client-server edition with external tiddlywiki.js",
 	"plugins": [
 		"tiddlywiki/tiddlyweb",
-		"tiddlywiki/filesystem",
-		"tiddlywiki/highlight"
+		"tiddlywiki/filesystem"
 	],
 	"themes": [
 		"tiddlywiki/vanilla",
@@ -13,7 +12,7 @@
 		"listen": [
 			"--listen","root-tiddler=$:/core/save/all-external-js","use-browser-cache=yes"],
 		"index": [
-			"--render","$:/core/save/offline-external-js","index.html","text/plain",
+			"--render","$:/core/save/offline-external-js","[[external-]addsuffix<version>addsuffix[.html]]","text/plain",
 			"--render","$:/core/templates/tiddlywiki5.js","[[tiddlywikicore-]addsuffix<version>addsuffix[.js]]","text/plain"],
 		"static": [
 			"--render","$:/core/templates/static.template.html","static.html","text/plain",
--- a/editions/server-external-js/tiddlers/config/$__DefaultTiddlers.tid
+++ b/editions/server-external-js/tiddlers/config/$__DefaultTiddlers.tid
@@ -1,0 +1,7 @@
+created: 20230314153132081
+modified: 20230314153243008
+title: $:/DefaultTiddlers
+type: text/vnd.tiddlywiki
+
+GettingStarted
+[[Using the external JavaScript template]]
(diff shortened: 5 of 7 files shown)


Jermolene:
Thanks @pmario pinging @cdruan who has worked on this area recently too, and might want to review this as well.

core/templates/external-js/save-all-external-js.tid
```diff
@@ -5,7 +5,10 @@ title: $:/core/save/all-external-js
 \define saveTiddlerFilter()
 [is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]] -[status[pending]plugin-type[import]] -[[$:/core]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]] $(publishFilter)$
 \end
+<!-- It's important to fully understand this PR https://github.com/Jermolene/TiddlyWiki5/pull/5570 -->
+<!-- and what tiddlers: $:/core/templates/tiddlywiki5.js and $:/core/templates/tiddlywiki5.js/tiddlers do -->
```
Jermolene, line 9: These comments are very cryptic. What are we trying to accomplish here? If there's a relevant piece of information I'd rather we provided it directly

editions/server-external-js/tiddlywiki.info
```diff
@@ -2,8 +2,7 @@
 	"description": "Client-server edition with external tiddlywiki.js",
 	"plugins": [
 		"tiddlywiki/tiddlyweb",
-		"tiddlywiki/filesystem",
-		"tiddlywiki/highlight"
```
Jermolene, line 6: Why has the highlight plugin been dropped here?

editions/server-external-js/tiddlers/external/tiddlywiki.files
```diff
@@ -0,0 +1,8 @@
+{
+	"tiddlers": [
+		{
+			"file": "../../../tw5.com/tiddlers/webserver/Using the external JavaScript template.tid",
```
Jermolene, line 4: I know we do things like this already, but this does strike me as super brittle. In particular, it won't be clear to anyone subsequently performing maintenance on that docs file that it is pulled in here. Perhaps add a comment to the top of the docs file specifically calling out that it is included elsewhere.

Jermolene (commented):
Thanks @pmario a couple of points.

editions/server-external-js/tiddlywiki.info
pmario: IMO it's useless. ... The goal of the external-js edition is to make the HTML file as small as possible. The highlight.js plugin IMO is a copy/paste artefact from tw5.com-server and should have never been there.

editions/server-external-js/tiddlers/external/tiddlywiki.files
pmario: Should I add some info as html comment eg: `<!-- this tiddler is imported into the server-external-js edtion -->` .. Or should I use a visible text

Jermolene: I was thinking of a comment; the concern is for people editing that tiddler in the future:

```
<!--
This tiddler is also included in the "server-external-js" edition. Take care before editing or moving it.
-->
```

core/templates/external-js/save-all-external-js.tid
pmario: Line 10 has been introduced with the PR #5570 and it's not obvious, what it does. The variable is named `defaultCoreURL` and the tiddler is named `core/templates/tiddlywiki5.js` ... Templates are usually used to create HTML output. .. So it's important to understand what's going on here. ... 

Additionally there are some inconsistencies between filenames and the tiddler titles ... So all of this has potential for confusion. ... Finding the connection between the different inconsistencies did cost me several hours of reverse engineering. The link to the PR and the 2 tiddler names are the best starting point.

editions/server-external-js/tiddlywiki.info
cdruan: The original intent is to have seamless transition to/from standard server edition, so that the users would not see any discrepancies (in rendered tiddlers) between two editions.

pmario: The size of the .html file created from the PR is 105kByte ... The size with the highlight plugin is trippled to 315kByte. 

This makes a significant difference if the file needs to be saved over the net to eg: github. Going from saving the file from 1 second to about 3 .... highly depending on the quality of the internet connection. 

If a user actively chooses to "burn" bandwidth, it's their choice. But IMO it's a "suboptimal experience" if it is forced upon users.

core/templates/external-js/save-all-external-js.tid
pmario: I could shorten the comment to 1 line eg:
`<!-- see and understand $:/core/templates/tiddlywiki5.js, $:/core/templates/tiddlywiki5.js/tiddlers and PR #5570 ! -->`

But IMO that's even more cryptic.

cdruan (commented):
Confirmed external-js TW ran as expected in server and html modes.

editions/server-external-js/tiddlers/system/$__SiteTitle.tid
```diff
@@ -0,0 +1,6 @@
+created: 20230314153323056
+modified: 20230314153504240
+title: $:/SiteTitle
+type: text/vnd.tiddlywiki
+
+{{$:/config/SaveWikiButton/Filename}}
```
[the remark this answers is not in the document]
pmario: OK. Then I'll remove it again

pmario:
@cdruan .. I did change the PR as you suggested. Thx.

Jermolene:
Thanks @pmario @cdruan is there any pressing reason to try to get this into v5.2.6?

pmario:
For everyone, which wants to work with a WebDav server and use the right filenames .. It is important. It's mainly a consistency problem which can cause confusion. ... The more time elapses till we fix it, the more backwards compatibility problems can come up once we do fix it.

saqimtiaz:
The lack of a summary as to what this PR accomplishes or what problem it addresses makes it harder to follow that it need be.

Note that the tiddler "Using the external JavaScript template" seems to already be in the wrong place in the docs directory structure which makes including it by path even more brittle.  This tiddler is tagged with "WebServer Guides" and does not fit with the other content under that tag, all of which has to do with the node.js server. We should consider moving this docs tiddler now and updating this PR accordingly.

pmario:
**Initially this section has been the last section of the reply**. .. But I moved it to the front. 
May be it's a TLDR; now.

If you **would try the steps** I did describe in **I did test** in the OP  with TW v5.2.**5** 

 - you will see that the workflow won't work, because
 - this PR is needed to get a smooth experience if users want to try out the external-js edition

The external-js setup is already somewhat complex, but the lack of documentation that actually works makes the situation even worse. 

I did try to fix the problem and I did find several severe problems on the way. ... Now this PR seems to be so unnecessary that it may be delayed ... 

I really don't see, what I've done wrong with the PR and the OP. Usually if I use too many external links in the OP I get rants about that. So with this OP I did try to make it as self-contained as I could.

I would have made a video, that would have shown the improvements the PR can make for testing external-js. .. But Jeremy does not watch them.

---------------

>The lack of a summary as to what this PR accomplishes or what problem it addresses makes it harder to follow that it need be.

The UL list in the OP IMO exactly points out, what the PR does. ... OK here is the rest as requested. 

- As I was testing the exernal-js version I did find this issue: https://github.com/Jermolene/TiddlyWiki5/issues/7346
- Which could be tracked down to this issue: https://github.com/Jermolene/TiddlyWiki5/issues/7348
  - Which Jeremy fixed with Fixed in https://github.com/Jermolene/TiddlyWiki5/commit/007b75942166e1ff138a46893250d4e62e85c85d
- Further testing revealed this issue: https://github.com/Jermolene/TiddlyWiki5/pull/7356
  - It go merged, because overwriting the single file index.html with an external index.html with the same name should not happen
 
Now that #7356 is merged .. This PR becomes necessary for consistency reasons. 

We could live with the problem that server-external-js now causes, because it does not cause any immediate damage wich #7356 did. .. **but** I think

 - It will cause maintenance problems in the future and 
 - it does cause documentation problems now. .. . **Which this PR should fix.** 

All in all I needed to invest about 1 day worth of work to find and point out problems, that are indirectly related to the topic I'm actually interested in. ... Which is the external-js functionality working with my private Node.js and WebDav setup.

--------

>This tiddler is tagged with "WebServer Guides" and does not fit with the other content under that tag, all of which has to do with the node.js server.

I think it works well with the Node.js version, which my "I did test" paragraph in the OP shows. ... Using the created files for WebDav could be seen as an extra feature. 

The `build --listen` command in `tiddlywiki.info` activates the browser cache header, which is what it is designed for. IMO it does save a lot of bandwidth if used with Node.js. 

image

>.. We should consider moving this docs tiddler now and updating this PR accordingly.

As I wrote. I would not know, what the new tag should be.

----------

Sorry for the wall of text.

saqimtiaz:
> The UL list in the OP IMO exactly points out, what the PR does. ... OK here is the rest as requested.

Thank you @pmario. What I found difficult from my perspective is that there is absolutely no context in the OP as to **why** these changes are desirable. It should not be necessary to read through the changes and guess what prompted them. Even with your reply I needed to click through several external links to understand what problem the PR is meant to address. The easier it is to understand the goal of a PR, the quicker it is to review accurately.

What I would expect in a PR is an introductory sentence along the lines of:

> This PR changes the file name of wikis built with the external-js template as well as the external js file to bring it into line with the changes made in #7536, which prevents the normal build target from overwriting the external-js version of the wiki.

See the following from https://tiddlywiki.com/#Contributing, emphasis is mine:

> Adequate explanation in the body of the PR for the motivation and implementation of the change. Focus on the **why** and what, rather than the how

pmario:
Would this have been better?

>The text describes changes made to TiddlyWiki's server-external-js plugin, which renames index.html to external-<version>.html and removes the highlight plugin to reduce the file size. Additionally, a tiddler called "Using the external JavaScript template" was imported from tw5-com, which allows users to save the external core and is included in the startup tiddlers list. The author tested the changes by creating a new TiddlyWiki, changing some tiddlers, downloading the external core and saving a snapshot, which worked successfully.

Summary created by ChatGPT using my OP as input.

Jermolene:
> Would this have been better?

No. The problem is that the OP describes the changes that are being made but doesn't explain why these changes are needed. Summarising the OP doesn't solve that problem.

pmario:
I did update the OP with a link to a new issue, which this PR should fix. I did **not** copy the text from the issue over to the PR.

Jermolene:
> bump

Hi @pmario this PR is already listed in #7345. I am away for a few days but will work through the outstanding PRs soon.

Jermolene:
Thanks @pmario

Jermolene:
@pmario I'm still struggling to phrase the purpose of this PR for the release note. What is the description of the purpose of these changes in one sentence?

PR #7362, opened 2023-03-14
Restructure and optimize `CheckboxWidget` docs

yaisog:
The doc tiddler for the CheckboxWidget was getting longer and longer with nearly every release for a while and has become a bit hard to navigate. Therefore, I put the various "modes" into tabs. Since this is a precedent for widget documentation, I'll mark this PR as a draft, in case that there arises the need for discussion.

I also put in Documentation Macros where appropriate, and changed the text (not the attributes) from "indexes" of DataTiddlers to "properties", as had been introduced in the `getindex` and `indexes` filter operator descriptions.

Furthermore, I added a table to `listField` and `listIndex` that summarizes how the checked and unchecked states are calculated for various attribute combinations. While the textual description is comprehensive and absolutely correct, by itself it may be a bit difficult to grasp.

I don't really like the horizontal line delimiting the tabs from the text below, but without that it was nearly impossible to discern where the tabs content ended. Any suggestions how to improve this are very welcome.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/CheckboxWidget (field Mode).tid
+++ b/editions/tw5.com/tiddlers/widgets/CheckboxWidget (field Mode).tid
@@ -1,0 +1,15 @@
+caption: ''field'' Mode
+created: 20230314171223911
+modified: 20230314200046885
+tags: CheckboxWidget
+title: CheckboxWidget (field Mode)
+
+!! <<.attr field>> Mode
+
+Using the checkbox widget in field mode requires the <<.attr field>> attribute to specify the name of the field. The <<.attr checked>> and <<.attr unchecked>> attributes specify the values to be assigned to the field to correspond to its checked and unchecked states respectively. The <<.attr default>> attribute is used as a fallback value if the field is missing or contains a value that does not correspond to the value of the <<.attr checked>> or <<.attr unchecked>> attributes.
+
+This example creates a checkbox that is checked if the field <<.field status>> is equal to <<.value open>> and unchecked if the field is equal to <<.value closed>>. If the field value is undefined then it defaults to <<.value closed>>.
+
+<<wikitext-example-without-html """<$checkbox field="status" checked="open" unchecked="closed" default="closed"> Is it open?</$checkbox>
+
+''status:'' {{!!status}}""">>
--- a/editions/tw5.com/tiddlers/widgets/CheckboxWidget (filter Mode).tid
+++ b/editions/tw5.com/tiddlers/widgets/CheckboxWidget (filter Mode).tid
@@ -1,0 +1,22 @@
+caption: ''filter'' Mode
+created: 20230314174505003
+modified: 20230314195819839
+tags: CheckboxWidget
+title: CheckboxWidget (filter Mode)
+
+!! <<..attr filter>> Mode
+
+Using the checkbox widget in filter mode requires the <<.attr filter>> attribute to contain a filter whose output will determine the checked state of the checkbox. In filter mode, checking the checkbox will not automatically make changes to any field of any tiddler. Instead, you can use the <<.attr actions>> attribute (or the <<.attr checkactions>> and <<.attr uncheckactions>> attributes) to specify what should happen when the checkbox is toggled. It is your responsibility to make sure the actions cause changes to the tiddlers or fields that the filter results depend on, so that the checkbox becomes properly checked or unchecked after the actions have triggered and the filter has updated.
+
+If the filter returns an empty result, the checkbox will be unchecked. Otherwise, if the filter result is non-empty, the checkbox will be checked.
+
+However, if either the <<.attr checked>> or <<.attr unchecked>> attributes (or both) are specified, then their values will be looked for in the filter result, instead of considering any non-empty value to mean <<.value checked>>.
+
+This example creates the same checkbox as in the <$button class="tc-btn-invisible tc-tiddlylink" set="$:/state/tab/CheckboxWidget" setTo="CheckboxWidget (listField Mode)">list mode example</$button>, selecting between <<.value red>> and <<.value green>> in the <<.field colors>> list field, but using filters and actions to make the change.
+
+<<wikitext-example-without-html """\define checkActions() <$action-listops $field="colors" $subfilter="-red green"/>
+\define uncheckActions() <$action-listops $field="colors" $subfilter="red -green"/>
+<$checkbox filter="[list[!!colors]]" checked="green" unchecked="red" default="red" checkactions=<<checkActions>> uncheckactions=<<uncheckActions>> > Is "green" in colors?</$checkbox>
+
+''colors:'' {{!!colors}}
+""">>
--- a/editions/tw5.com/tiddlers/widgets/CheckboxWidget (index Mode).tid
+++ b/editions/tw5.com/tiddlers/widgets/CheckboxWidget (index Mode).tid
@@ -1,0 +1,13 @@
+caption: ''index'' Mode
+created: 20230314171351122
+modified: 20230314195837258
+tags: CheckboxWidget
+title: CheckboxWidget (index Mode)
+
+!! <<.attr index>> Mode
+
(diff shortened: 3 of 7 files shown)


yaisog:
@rmunn @tobibeer @Jermolene : Since you made the most recent changes to these docs, I'd appreciate it if you could check that I didn't mangle them too badly...

Jermolene:
Thank you @yaisog this is good stuff. I very much like the updates to use the docs macros, they do seem to have a positive impact on readability.

I'm happy with the precedent of adding tabs to widget documentation; I think quite a few widgets might benefit from the same treatment.

As you note, the presence of the section on indeterminate checkboxes immediately after the tabs is confusing. It would be preferable for the tabs to be the last thing in the tiddler. Perhaps we could move that material into a tab?

I think it would be helpful if the heading within each tab was a link to the underlying tiddler.

pmario:
I think the section with "Indeterminate checkboxes" is always below the tabs. That's confusing. IMO it should be a link to a different tiddler and there should be no content below the tabs, that does not change with if the tab changes.

yaisog:
Hello @Jermolene, thanks for the kind words and your ideas for further improvement.

I did put the `indeterminate` section into its own tab and renamed all tab captions to just the attribute names. The tabs can now also be opened directly from the attributes table. Since the buttons that do that look like links, the user might however expect something different, like a new tiddler opening?

To make the tabs stand out a bit more, I made their text bigger than normal -- they are now the same size as the surrounding tiddler text -- and gave them a padding, so that they have room to breathe. This is controlled by giving the tabs the class `doc-tabs`, so that may be reused for other reference tiddlers.

> I very much like the updates to use the docs macros, they do seem to have a positive impact on readability.

Few people seem to have read the _Documentation Style Guide_. 😞

pmario:
>Few people seem to have read the Documentation Style Guide. 😞

Yea, but even if you did read it, it very easy to "forget about the existing styles" even if you would know they exist. ... And some documentation is much older than the doc-macros. ... So it's still a "work in progress" ;)

pmario:
I did click through the new tabs. It's definitely a much better structure than it is at the moment. I personally think that tabs look a bit strange, but they are definitely more visible. ...

Looks good to me.

pmario:
OT -----------

@yaisog ... I'm preparing a PR to improve the filter docs at the moment. So everything tagged Filter, Filter Syntax and everything which has a "railroad diagram" in it will probably be changed. 

Just to let you know.

yaisog:
> I'm preparing a PR to improve the filter docs at the moment.

👍 
That's going to be a lot of work. Let me know if you want help.

Jermolene:
Thanks @yaisog I've added this PR to the list for consideration post-v5.2.6 – #7345

> I did put the `indeterminate` section into its own tab and renamed all tab captions to just the attribute names. The tabs can now also be opened directly from the attributes table. Since the buttons that do that look like links, the user might however expect something different, like a new tiddler opening?

Those are great improvements, thank you.

I did quite like the way that the previous version formatted the name of the attributes in the tabs and a qualifying word (I forget; was it "mode"?). I assume the reason for the change was to make the tabs fit better on small screens. Perhaps we could restore the formatting? (But not the qualifying word).

> To make the tabs stand out a bit more, I made their text bigger than normal -- they are now the same size as the surrounding tiddler text -- and gave them a padding, so that they have room to breathe. This is controlled by giving the tabs the class `doc-tabs`, so that may be reused for other reference tiddlers.

That's good, and generally useful. We might consider putting the associated classes into the core.

> I'm changing this PR to _Ready for review_, so you may merge it at your leisure. I am still available for further improvements, however, if there are more ideas...
> 
> > I very much like the updates to use the docs macros, they do seem to have a positive impact on readability.
> 
> Few people seem to have read the _Documentation Style Guide_. 😞

Indeed! Using the right macros is a superpower, and makes things easier to read.

yaisog:
> I did quite like the way that the previous version formatted the name of the attributes in the tabs and a qualifying word (I forget; was it "mode"?). I assume the reason for the change was to make the tabs fit better on small screens. Perhaps we could restore the formatting? (But not the qualifying word).

Hi @Jermolene, you're exactly right. Since we're setting precedents here that may find use in more docs tiddlers, I was afraid that the tabs would get too wide on small screens when there are many of them. "Mode" was indeed the qualifying word, which I removed because there was no "indeterminate" _mode_. So I went back to using the attribute name only.

I did originally kept using the `.attr` macro, because it was the natural thing to do. But it looked awkward in larger sizes, and was too small in its original size (Consolas font on windows). I tried bold and regular weights and different shades of gray (not the sleazy movie).
Let me play around a bit more with it.

PS: I always wondered why the attribute names in the overview table don't use the `.attr` formatting macro...

Jermolene:
Great stuff thanks @yaisog. As part of the precedent setting, perhaps we might add some material about widget description tiddlers to the documentation styleguide? Improving the docs would make it easier for others to help out with updating the existing WDTs.

Consistently using the `.attr` macro in the widget attribute tables would be very helpful.

yaisog:
> I was expecting that the "indeterminate" tab would not use the .attr macro but perhaps you were concerned about consistency?

Right again. Since all tab labels describe attributes, I used the `.attr` macro for all of them.
So, which of the three font options do you like best?

yaisog:
There are three new macros that should make working with tabs in widget documentation much easier. I tried to make them universal. Since they're short, I'm posting them here for an at-a-glance review:
```
\define .widget-tabs() <$macrocall $name="tabs" tabsList="[tag<currentTiddler>]" default={{{ [tag<currentTiddler>first[]] }}} explicitState={{{ [<currentTiddler>addprefix[$:/state/tab/]] }}} class="doc-tabs" />

\define .widget-tab-link(text, target, tooltip) <$button class="tc-btn-invisible tc-tiddlylink" set={{{ [<currentTiddler>addprefix[$:/state/tab/]] }}} setTo=<<__target__>> tooltip=<<__tooltip__>>><<__text__>></$button>

\define .widget-attr-link(text, target) <$macrocall $name=".widget-tab-link" text={{{ [[<code class="doc-attr">]] [<__text__>] [[</code>]] +[join[]] }}} target=<<__target__>> tooltip="Show more information about the 'tag' attribute" />
```
The first macro will show the tabs for a WDT, set a default tiddler and explicitState that'll be reused later. It does not need parameters.
_Do we need some way to distinguish tab content tiddlers from other tiddlers that may be tagged with the WDT?_

The second and third macros define a shortcut to setting the current tab (one for regular text, the other with `.attr` formatting).

If these are acceptable, I'd extend the "Documentation Macros" tiddler appropriately.

yaisog:
On the topic of styling the "tab links" (clickable attributes in the table that open the corresponding tab):
By a serendipitous accident I didn't commit some style tinkering in the developer console to the stylesheet, so that there was no link coloring (blue) for these links, but they were still underlined on hover. I quite liked this. Users might only stumble over the link functionality by accident (hover), but now they don't disrupt the table so much. May be problematic for mobile, though, where there is no hover...

pmario:
I do like the first mock-up the best.

Jermolene:
Great stuff, thank you @yaisog I think this is good to go.

Jermolene:
Sorry @yaisog I missed your list of outstanding issues in the OP.

>  Finalize formatting for tab buttons

I'm happy with the current state.

> Decide on visual distinction of "tab links" in the attribute table

I think it would be sufficient to give the links the primary colour, and an underline on hover, just like ordinary tiddler links.

>  Decide on using tm-scroll as part of "tab links" buttons (scroll to the tabs)

I think that's a good idea.

>  Rename macros to reflect their general use

I think it's reasonable to use semantic names for these macros (ie name them for what they are for, rather than what they do)

yaisog:
Hi @Jermolene, this PR should be good to go now.

yaisog:
Hey @Jermolene, is there something holding this up? I'd be happy to do what's necessary to get this out of the door. I'd also like to look at some other widget docs, but I may want to use the stuff from this PR.

Jermolene:
> Hey @Jermolene, is there something holding this up? I'd be happy to do what's necessary to get this out of the door. I'd also like to look at some other widget docs, but I may want to use the stuff from this PR.

Hi @yaisog no I don't think there is – I had erroneously listed this ticket as part of the post-v5.2.7 backlog in #7345 but of course it's docs only and so shouldn't be affected by the moratorium. I'll review it again now.

PR #7365, opened 2023-03-18
Prevents shadow tiddlers from being saved to local storage

### LINKED ISSUE #7364: [BUG] Browser storage plugin does not remove deleted overwritten shadow tiddlers from storage
### Describe the bug

When using the browser storage plugin and a shadow tiddler is overwritten and that tiddler is written to browser storage, then deleting the tiddler does not remove it from local storage. The only way to remove it is to use developer tools and manually remove the local storage entry.

### Expected behavior

When an overwritten shadow tiddler is deleted, it should also be deleted from local storage.

I haven't fully thought through how to fix yet. It might work to check if the changed tiddler only exists as a shadow tiddler, then its key should be deleted from local storage.

### To Reproduce

1. Go to https://tiddlywiki.com/editions/full/#%24%3A%2Fconfig%2FBrowserStorage%2FSaveFilter:%24%3A%2Fplugins%2Ftiddlywiki%2Fbrowser-storage%20%24%3A%2Fconfig%2FBrowserStorage%2FSaveFilter%20%24%3A%2Fconfig%2FDefaultMoreSidebarTab
2. Change the browser storage save filter to be more liberal with the tiddlers it saves: `[all[]] -[is[shadow]!is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]]`
3. Make some change the value in the `$:/config/DefaultMoreSidebarTab` shadow tiddler. This will cause the overwritten shadow tiddler to be saved to local storage
4. Delete the overwritten shadow tiddler
5. Refresh the page and notice the overwritten shadow tiddler comes back in spite of its previous deletion.

### Screenshots

_No response_

### TiddlyWiki Configuration

5.2.6 prerelease (the issue is nothing new to this version...it has always been there)

### Additional context

_No response_

btheado:
If a tiddler is only a shadow tiddler, then do not save it to local storage. Otherwise when installing a plugin, each individual tiddler will be expanded and saved individually. In order to uninstall such a plugin, the plugin tiddler and each tiddler needs to be deleted.

This can be prevented by including a run like `-[is[shadow]!is[tiddler]]` in the save filter, but that has the side effect of preventing overwritten shadow tiddlers from being deleted from local storage.

Fixes #7364

### DIFF
--- a/plugins/tiddlywiki/browser-storage/settings.tid
+++ b/plugins/tiddlywiki/browser-storage/settings.tid
@@ -17,8 +17,8 @@ Click this button to clear browser storage and disable its use:
 This filter determines which tiddlers will be saved to local storage.
 
 * `[prefix[$:/state/]] -[prefix[$:/state/popup/]]` - the default value. Save state tiddlers except popup state tiddlers, thus preserving selected tabs and the open/closed status of table of contents entries. Any other tiddlers created or changed will be lost after reloading the page.
-* `[all[]]` - attempt to save all changed tiddlers. This means even popup state tiddlers and temporary tiddlers will be saved. In addition, when a plugin is installed, all the shadow tiddlers are individually "exploded" into local storage. Deleting the plugin requires deleting all the tiddlers individually. Not recommended unless these issues are unimportant.
-* `[all[]] -[is[shadow]!is[tiddler]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]]` - save all tiddlers except unmodified shadow tiddlers, popup state tiddlers, temp tiddlers and the history list. Solves the aforementioned issues with `[all[]]`. Recommended.
+* `[all[]]` - attempt to save all changed tiddlers. This means even popup state tiddlers and temporary tiddlers will be saved.
+* `[all[]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]]` - save all tiddlers except popup state tiddlers, temp tiddlers and the history list.
 
 <$link to="$:/config/BrowserStorage/SaveFilter">Browser Storage Save Filter</$link>: <$edit-text tiddler="$:/config/BrowserStorage/SaveFilter" default="" tag="input" size="50"/>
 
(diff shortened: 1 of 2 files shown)


btheado:
The diff isn't showing very cleanly when all I did was add an if and else (and changed indent accordingly). Here is the diff with `-b` option which is easier to read:

```diff
diff --git a/plugins/tiddlywiki/browser-storage/util.js b/plugins/tiddlywiki/browser-storage/util.js
index a88c358f6..7474a367c 100644
--- a/plugins/tiddlywiki/browser-storage/util.js
+++ b/plugins/tiddlywiki/browser-storage/util.js
@@ -53,6 +53,8 @@ BrowserStorageUtil.prototype.saveTiddlerToLocalStorage = function(title) {
     // Get the tiddler
     var tiddler = $tw.wiki.getTiddler(title);
     if(tiddler) {
+        if (this.wiki.tiddlerExists(title)) {
+            // This is not a shadow tiddler
             console.log("browser-storage: Saving",title);
             // Get the JSON of the tiddler
             var json = JSON.stringify(tiddler.getFieldStrings());
@@ -72,6 +74,11 @@ BrowserStorageUtil.prototype.saveTiddlerToLocalStorage = function(title) {
                     console.log("Browser-storage error:",e);
                 }
             }
+        } else {
+            // Shadow tiddler which is no longer overwritten (or never was)
+            // Ensure it is not in local storage
+            this.removeTiddlerFromLocalStorage(title);
+        }
     } else {
         // In local storage, use the special value of empty string to mark the tiddler as deleted
         // On future page loads, if the tiddler is already gone from startup then the blank entry
```

btheado:
Using this save filter: `[all[]] -[prefix[$:/state/popup/]] -[prefix[$:/temp/]] -[prefix[$:/HistoryList]]`, I tested installing a plugin and just the plugin tiddler being added to local storage (as desired). I also edited a shadow tiddler and it was successfully stored to local storage and restored upon reload. I deleted the overwritten shadow tiddler and it was removed from local storage as desired. Deleting normal tiddler by blanking the value in local storage still works as it did before.

Jermolene:
Thanks @btheado that looks reasonable. I'll add this to the list for post-v5.2.6

Jermolene:
Thanks for your patience @btheado

PR #7371, opened 2023-03-23
Fixed issue with trailing newlines in macrodefs

flibbles:
[This commit concerning indentation of pragma](https://github.com/Jermolene/TiddlyWiki5/commit/dd6e00687b218ecc3091af2612c46593933d57ac) introduced a bug into the wikiparser where macrodefs would eat up all the newline characters at the end of its definition. This changed parser behavior, since trailing newlines can be important for distinguishing block from inline. Take this example:

```
\define macro()

<$macrocall $name=other />

\end
```

In v5.2.5 and earlier, that macrocall would be processed as a block. In v5.2.6, it cannot be in that circumstance. This problem is particularly concerning for my Uglify macro, which took advantage of removing unneeded closing elements (e.g. `</$reveal>`.

The following fixes the problem and tests. Instead of eating all whitespace at the end of a definition, it excludes newlines.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/macrodef.js
+++ b/core/modules/parsers/wikiparser/rules/macrodef.js
@@ -58,7 +58,7 @@ exports.parse = function() {
 	var reEnd;
 	if(this.match[3]) {
 		// If so, the end of the body is marked with \end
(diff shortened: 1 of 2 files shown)


Jermolene:
Thank you @flibbles. We need to consider whether the impact of this problem is severe enough to warrant a v5.2.7 bug fix release; I'm inclined to think it is.

PR #7376, opened 2023-03-25
Tweak some examples for get and getindex

yaisog:
The examples unnecessarily used a `:filter` run prefix when a simple filter step would have done.
The explanation has also been extended somewhat to make it clearer what's going on.


Jermolene:
Thanks @yaisog

PR #7378, opened 2023-03-27
Fix missing extension for audio/mpeg

donmor:
There should be multiple extensions for audio/mpeg like `.mp3`, `.mp2` and so on.
The missing extensions raised exception in Tiddloid, breaking the tweaks plugin. I'll patch it in case of another bug like this.


Jermolene:
Thanks for your patience @donmor

PR #7380, opened 2023-03-27
Add the querySelectorSafe() function and replace querySelector() usage

### LINKED ISSUE #7375: [BUG] RSoE with `tm-scroll` when selector is invalid
### Describe the bug

With generated selectors, `tm-scroll` may generate an RSoE when the selector contains invalid characters because `escapecss[]` was not used.
I this particular case, `encodeuricomponent[]` was used with superficially valid results, but it looks like % in the selector attribute then led to the RSoE.

### Expected behavior

No scrolling, but also no RSoE when the selector contains invalid characters.

### To Reproduce

1. Go to tiddlywiki.com/prerelease
2. Create new tiddler
3. Paste this as text and save tiddler:
```
<$button>
    Click
    <$action-sendmessage $message="tm-scroll" selector=".doc-tabs-New%20Tiddler" />
</$button>
```
4. Click button

### Screenshots

_No response_

### TiddlyWiki Configuration

5.2.7-prerelease

### Additional context

_No response_

yaisog:
The argument to various `querySelector()` is never checked for validity which can lead to DOMExceptions when the selector is invalid.

To alleviate the potential problems, the `querySelectorSafe()` and `querySelectorAllSafe()` utility functions are introduced. Calls to `querySelector()` are replaced where appropriate.

Closes #7375 

- [x] Test implementation with Preview
- [x] Move new functions into `$:/core/modules/utils/dom.js`
- [x] Make sure the authorities approve of the new functions
- [x] Implement fallback to _document_
- [x] Find and replace remaining uses of `querySelector()`
- [x] Decide on usage replacement with hardcoded selectors

### DIFF
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -128,7 +128,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 
 PageScroller.prototype.scrollSelectorIntoView = function(baseElement,selector,callback,options) {
 	baseElement = baseElement || document.body;
-	var element = baseElement.querySelector(selector);
+	var element = $tw.utils.querySelectorSafe(selector,baseElement);
 	if(element) {
 		this.scrollIntoView(element,callback,options);
 	}
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -127,7 +127,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 };
 
 PageScroller.prototype.scrollSelectorIntoView = function(baseElement,selector,callback,options) {
-	baseElement = baseElement || document.body;
+	baseElement = baseElement || document;
 	var element = $tw.utils.querySelectorSafe(selector,baseElement);
 	if(element) {
 		this.scrollIntoView(element,callback,options);
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -74,7 +74,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	this.cancelScroll(srcWindow);
 	this.startTime = Date.now();
 	// Get the height of any position:fixed toolbars
-	var toolbar = $tw.utils.querySelectorSafe(".tc-adjust-top-of-scroll",srcWindow.document),
+	var toolbar = srcWindow.document.querySelector(".tc-adjust-top-of-scroll"),
 		offset = 0;
 	if(toolbar) {
 		offset = toolbar.offsetHeight;
--- a/core/modules/utils/dom/scroller.js
+++ b/core/modules/utils/dom/scroller.js
@@ -74,7 +74,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	this.cancelScroll(srcWindow);
 	this.startTime = Date.now();
 	// Get the height of any position:fixed toolbars
-	var toolbar = srcWindow.document.querySelector(".tc-adjust-top-of-scroll"),
+	var toolbar = $tw.utils.querySelectorSafe(".tc-adjust-top-of-scroll",srcWindow.document),
 		offset = 0;
 	if(toolbar) {
 		offset = toolbar.offsetHeight;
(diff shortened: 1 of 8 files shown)


core/modules/utils/utils.js
```diff
@@ -685,19 +685,9 @@ exports.escapeRegExp = function(s) {
...
-	return encodeURIComponent(s).replace(/[!'()*]/g,function(c) {
-		return "%" + c.charCodeAt(0).toString(16).toUpperCase();
-	});
-};
-
 // Checks whether a link target is external, i.e. not a tiddler title
 exports.isLinkExternal = function(to) {
-	var externalRegExp = /^(?:file|http|https|mailto|ftp|irc|news|obsidian|data|skype):[^\s<>{}\[\]`|"\\^]+(?:\/|\b)/i;
+	var externalRegExp = /^(?:file|http|https|mailto|ftp|irc|news|data|skype):[^\s<>{}\[\]`|"\\^]+(?:\/|\b)/i;
 	return externalRegExp.test(to);
```
saqimtiaz, line 691: is this an intentional change in this PR?

yaisog: No, something went definitely wrong during transfer on my end. Pushed a fix already.

core/modules/utils/dom/scroller.js
```diff
@@ -128,7 +128,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 
 PageScroller.prototype.scrollSelectorIntoView = function(baseElement,selector,callback,options) {
 	baseElement = baseElement || document.body;
```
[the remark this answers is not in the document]
yaisog: For `tm-focus-selector` in the $rootwidget, we have
```
doc = event.event && event.event.target ? event.event.target.ownerDocument : document;
```
So, which is the better default, `document` or `document.body`? Does it even make a difference?

core/modules/utils/utils.js
yaisog: Yes, you may be right. Probably my GitHub Desktop was not yet fully synced.

core/modules/utils/dom/scroller.js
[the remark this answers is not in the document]
saqimtiaz: I would normally go with `document` since the method is a property of the Document object. Thoughts @Jermolene ?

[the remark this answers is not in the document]
yaisog: Agreed. But let's see what Jeremy thinks.
Any objections to changing the default also in `PageScroller.prototype.scrollSelectorIntoView` if we all agree on `document`?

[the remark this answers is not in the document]
Jermolene: Thanks @yaisog @saqimtiaz I think `document` would be appropriate here.

yaisog:
The fallback to _document_ is implemented and I replaced any other instanced of `querySelector*` in the core. I didn't want to touch plugins that still use the old ways, because that would make them dependent on v5.2.8.

I'm marking this PR as "ready for review".

core/modules/utils/dom/scroller.js
```diff
@@ -74,7 +74,7 @@ PageScroller.prototype.scrollIntoView = function(element,callback,options) {
 	this.cancelScroll(srcWindow);
 	this.startTime = Date.now();
 	// Get the height of any position:fixed toolbars
-	var toolbar = srcWindow.document.querySelector(".tc-adjust-top-of-scroll"),
+	var toolbar = $tw.utils.querySelectorSafe(".tc-adjust-top-of-scroll",srcWindow.document),
```
saqimtiaz, line 77: I am not sure that we need to use the `querySelectorSafe` method where we are providing the selector since we know that it is valid, though I admit that it cannot hurt. Recommend we get input from others on this.

yaisog: I did think so, too, initially, but consistency won out in the end. At least for me. Other opinions are welcome, of course. Which "others" did you have in mind, @Jermolene?

Jermolene: Thanks @yaisog @saqimtiaz I would be inclined not to use `$tw.utils.querySelectorSafe()` for the reason given by Saq

Jermolene:
Thanks @yaisog I've added this PR to #7345 for merging soon.

Jermolene:
Thanks @yaisog

PR #7386, opened 2023-04-01
Add the `words` and `lines` modes to `$diff-text`

yaisog:
Similar to their implementation in the new `makepatches` operator, `$diff-text`, using the same library, could also use the `words` and `lines`modes. I believe there are quite a few cases that will be more intelligible if `words` mode were used. Similarly, highlighting whole lines may also be beneficial in some cases.

This PR adds the same `diffLineWordMode` and `diffPartsToChars` functions to `$diff-text` that were also added to `strings.js` as part of the implementation for the `makepatches` operator. This duplication feels wrong. Should these functions maybe added to the diff-match-patch.js (minimized) file or some other file?

You can check out the differences by going to the DiffTextWidget documentation tiddler and clicking the corresponding buttons at the bottom. This doc tiddler extension is not finished yet. I just wanted to add a short demo somewhere for now.

- [ ] Decide what to do with the auxiliary functions
- [ ] Finish the doc tiddler

### DIFF
--- a/core/modules/widgets/diff-text.js
+++ b/core/modules/widgets/diff-text.js
@@ -39,19 +39,25 @@ DiffTextWidget.prototype.render = function(parent,nextSibling) {
 	this.execute();
 	// Create the diff
 	var dmpObject = new dmp.diff_match_patch(),
+		mode = this.getAttribute("mode") || "chars",
+		diffs;
+	if(mode === "lines" || mode === "words") {
+	  diffs = diffLineWordMode(this.getAttribute("source"),this.getAttribute("dest"),mode);
+	} else {
 		diffs = dmpObject.diff_main(this.getAttribute("source"),this.getAttribute("dest"));
-	// Apply required cleanup
-	switch(this.getAttribute("cleanup","semantic")) {
-		case "none":
-			// No cleanup
-			break;
-		case "efficiency":
-			dmpObject.diff_cleanupEfficiency(diffs);
-			break;
-		default: // case "semantic"
-			dmpObject.diff_cleanupSemantic(diffs);
-			break;
-	}
+		// Apply required cleanup
+		switch(this.getAttribute("cleanup","semantic")) {
+			case "none":
+				// No cleanup
+				break;
+			case "efficiency":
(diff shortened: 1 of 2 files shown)


Jermolene:
Looks good, thank you @yaisog

core/modules/widgets/diff-text.js
```diff
@@ -35,9 +35,15 @@ DiffTextWidget.prototype.render = function(parent,nextSibling) {
 	this.computeAttributes();
 	this.execute();
 	// Create the diff object
-	var dmpObject = new dmp.diff_match_patch();
-		dmpObject.Diff_EditCost = $tw.utils.parseNumber(this.getAttribute("editcost","4"));
-	var	diffs = dmpObject.diff_main(this.getAttribute("source",""),this.getAttribute("dest",""));
+	var dmpObject = new dmp.diff_match_patch(),
```
Leilei332, line 38: Note that the diff-match-patch APIs have been changed since #9511 is merged.

core/modules/widgets/diff-text.js
```diff
@@ -133,12 +139,74 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
...
-	if(changedAttributes.source || changedAttributes.dest || changedAttributes.cleanup || changedAttributes.editcost) {
+	if(changedAttributes.source || changedAttributes.dest || changedAttributes.cleanup || changedAttributes.mode || changedAttributes.editcost) {
 		this.refreshSelf();
 		return true;
 	} else {
 		return this.refreshChildren(changedTiddlers);
 	}
 };
 
+// these two functions are adapted from https://github.com/google/diff-match-patch/wiki/Line-or-Word-Diffs
```
Leilei332, line 150: The same code also exists in https://github.com/TiddlyWiki/TiddlyWiki5/blob/master/core/modules/filters/strings.js . IMO it may be better to move it to a `util` module to avoid code duplication.

yaisog:
To update this cleanly to the post-#9511 API for diff-match-patch, I made a new PR #9551 and will close this one.

Issue #7387, opened 2023-04-03
Official Docker image?


Nezteb:
**Is your feature request related to a problem? Please describe.**
I like to host things on https://fly.io/, using only a single `Dockerfile` and their custom `fly.toml` config. At the moment, this doesn't seem easily doable with TiddlyWiki.

**Describe the solution you'd like**
I'd like a simple `Dockerfile` to be available for deploying TiddlyWiki to something like Fly.io or AWS ECS/EKS.

**Describe alternatives you've considered**
There is a Docker image, but it seems like an undocumented community-driven effort? https://hub.docker.com/r/elasticdog/tiddlywiki

**Additional context**
I searched GitHub issues but didn't find anything: https://github.com/Jermolene/TiddlyWiki5/issues?q=in%3Atitle+docker

Jermolene:
Hi @Nezteb I'd be very happy if we had an official Docker image that was automatically published by our build scripts. I'm not a Docker user myself, and am not familiar with it at all, so I would need some assistance.

Jermolene:
> I'll work on it and make a PR!

Great, thank you @Nezteb much appreciated

PR #7393, opened 2023-04-03
Add initial TiddlyWiki Dockerfile and docs

Nezteb:
Resolves https://github.com/Jermolene/TiddlyWiki5/issues/7387

I'm still not super familiar with TiddlyWiki (especially the documentation format and how to run it securely with auth), so let me know if I need to update the docs I wrote. 😅 

One other question for you Jeremy: Do you want to publish this to Docker Hub under an "official" TiddlyWiki org? Similar to https://hub.docker.com/r/elasticdog/tiddlywiki. I wouldn't be able to do those parts, but I could guide you through them if needed.

The other option is to just require users to build their own images for TiddlyWiki.

Let me know what you think. 😄

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,21 @@
+# Adapted from: https://github.com/elasticdog/tiddlywiki-docker/blob/master/5/Dockerfile
+
+ARG NODE_LTS_VERSION=18
+
+FROM node:${NODE_LTS_VERSION}-alpine
+
+ARG TIDDLYWIKI_VERSION=5.2.7
+ARG TIDDLYWIKI_DATA_DIR=/tiddlywiki
+ARG TIDDLYWIKI_PORT=8080
+
+# https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals
+RUN apk add --no-cache tini
+RUN npm install -g tiddlywiki@${TIDDLYWIKI_VERSION}
+
+EXPOSE ${TIDDLYWIKI_PORT}
+
+VOLUME ${TIDDLYWIKI_DATA_DIR}
+WORKDIR ${TIDDLYWIKI_DATA_DIR}
+
+ENTRYPOINT ["/sbin/tini", "--", "tiddlywiki"]
+CMD ["--help"]
--- a/editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
+++ b/editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
@@ -1,0 +1,39 @@
+created: 20230403211230516
+modified: 20230403211230516
+tags: [[TiddlyWiki with Docker]]
+title: Using TiddlyWiki with Docker
+type: text/vnd.tiddlywiki
+
+~TiddlyWiki5 can also be run using Docker. See the [[Dockerfile]] for details.
+
+The image can be built using the following command in the root of the TiddlyWiki git repo:
+```
+docker build --tag tiddlywiki .
+```
+
+It can be run like so:
+```
+# Create a data directory
+mkdir ~/tiddlywiki
+
+# Initialize the data directory
+docker run \
+    --rm \
+    --name tiddlywiki \
+    --volume ~/tiddlywiki:/tiddlywiki \
+    tiddlywiki \
+    mynewwiki --init server
+
+# Start the server with custom arguments
+docker run \
+    --rm \
+    --name tiddlywiki \
+    --volume ~/tiddlywiki:/tiddlywiki \
+    --publish 8080:8080 \
+    tiddlywiki \
+    mynewwiki --listen host=0.0.0.0 port=8080 "readers=(anon)" "writers=(authenticated)" admin=noah username=noah password=noah
+```
+
+If you'd prefer a pre-built image, there is a community-maintained image on Docker Hub: https://hub.docker.com/r/elasticdog/tiddlywiki/
+
+See [[Commands]] for a full listing of the available commands.
--- a/editions/tw5.com/tiddlers/readme/ReadMe.tid
+++ b/editions/tw5.com/tiddlers/readme/ReadMe.tid
@@ -32,6 +32,10 @@ Developer documentation is in progress at https://tiddlywiki.com/dev/
 
 {{Upgrading TiddlyWiki on Node.js}}
 
+! Using TiddlyWiki with Docker
+
+{{Using TiddlyWiki with Docker}}
+
 ! Also see
 
 <<list-links "[tag[TiddlyWiki on Node.js]] -[[Installing TiddlyWiki on Node.js]] -[[Using TiddlyWiki on Node.js]] -[[Upgrading TiddlyWiki on Node.js]]">>


pmario:
I think this docker-file is _not_ sufficient enough. The minimum requirement for a TW instance that could face the internet is the existing "username / password" authentication. ... 

IMO the dockerfile must make sure, that minimal security settings are active if the instance is started. There should be options to deactivate every setting on a per setting level. ... So if the hosting-provider does provide additional security like Let's-Encrypt it can be used at the provider level. 

Providers usually also have some kind of DOS or DDOS protection ... and so on. 

just some thoughts. ... There has been a discussion at Talk or her at GH lately. I can't find the thread atm. ... But the docker setup used was more detailed.

pmario:
This is the Talk: https://talk.tiddlywiki.org/t/how-to-upgrade-tw-inside-of-docker-container/6317
This is the repo: https://gitlab.com/nicolaw/tiddlywiki

I did not run any tests with that repo, but I thin it has potential

pmario:
The image mentioned in the OP is stored in the GH repo: https://github.com/elasticdog/tiddlywiki-docker

Dockerfile
```diff
@@ -0,0 +1,21 @@
+# Adapted from: https://github.com/elasticdog/tiddlywiki-docker/blob/master/5/Dockerfile
+
+ARG NODE_LTS_VERSION=18
+
+FROM node:${NODE_LTS_VERSION}-alpine
+
+ARG TIDDLYWIKI_VERSION=5.2.7
```
Jermolene, line 7: Forgive my ignorance; can the user override the values that are specified via ARG? I'm thinking it might be useful to give users the option to install the most recent version of TW? Or is it safer to clamp things to a specific version? It may also be useful to be able to specify "master" or another branch name.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ARG TIDDLYWIKI_DATA_DIR=/tiddlywiki
+ARG TIDDLYWIKI_PORT=8080
+
+# https://github.com/nodejs/docker-node/blob/master/docs/BestPractices.md#handling-kernel-signals
```
Jermolene, line 11: Thanks for the link, most of it is over my head, but the warning about running as root did jump out. Is that something we could address?

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
```diff
@@ -0,0 +1,39 @@
+created: 20230403211230516
+modified: 20230403211230516
+tags: [[TiddlyWiki with Docker]]
```
Jermolene, line 3: The tiddler 'TiddlyWiki with Docker' doesn't exist, and so that means that this docs tiddler won't appear in the table of contents. Perhaps it makes more sense to remove this tag and instead tag this tiddler `[[TiddlyWiki on Node.js]]`.

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
```diff
@@ -0,0 +1,39 @@
...
+title: Using TiddlyWiki with Docker
+type: text/vnd.tiddlywiki
+
+~TiddlyWiki5 can also be run using Docker. See the [[Dockerfile]] for details.
```
Jermolene, line 7: The file `Dockerfile` isn't part of the build process and so this link won't work. Perhaps we could make it:

```
[[Dockerfile|https://github.com/Jermolene/TiddlyWiki5/blob/master/Dockerfile]]
```

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
```diff
@@ -0,0 +1,39 @@
...
+
+The image can be built using the following command in the root of the TiddlyWiki git repo:
+```
+docker build --tag tiddlywiki .
+```
+
+It can be run like so:
+```
+# Create a data directory
+mkdir ~/tiddlywiki
```
Jermolene, line 17: Perhaps it would be clearer for the directory name to be something like `tiddlywiki-data`?

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
```diff
@@ -0,0 +1,39 @@
...
+type: text/vnd.tiddlywiki
+
+~TiddlyWiki5 can also be run using Docker. See the [[Dockerfile]] for details.
+
+The image can be built using the following command in the root of the TiddlyWiki git repo:
+```
+docker build --tag tiddlywiki .
+```
+
+It can be run like so:
```
Jermolene, line 14: It would be worth explicitly noting the placeholder strings `mynewwiki` and `noah` to remind the user to change them as required

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
```diff
@@ -0,0 +1,39 @@
...
+docker run \
+    --rm \
+    --name tiddlywiki \
+    --volume ~/tiddlywiki:/tiddlywiki \
+    --publish 8080:8080 \
+    tiddlywiki \
+    mynewwiki --listen host=0.0.0.0 port=8080 "readers=(anon)" "writers=(authenticated)" admin=noah username=noah password=noah
+```
+
+If you'd prefer a pre-built image, there is a community-maintained image on Docker Hub: https://hub.docker.com/r/elasticdog/tiddlywiki/
```
Jermolene, line 37: I would like to publish an official TiddlyWiki image. I'm guessing we'd need a script `./bin/docker-publish.sh` which would be used similarly to the existing `./bin/npm-publish.sh`. We'd also need to update https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/dev/tiddlers/build/Releasing%20a%20new%20version%20of%20TiddlyWiki.tid which is the checklist I follow for making a new release.

Jermolene (commented):
Thank you @Nezteb – with apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

Dockerfile
Nezteb: Yeah a user could do that like so:
```bash
docker build -t tiddlywiki --build-arg TIDDLYWIKI_VERSION=x.y.z .
```

I thought about defaulting to latest but then forgot about it; I'll set that up instead!

Regarding "is it safer", I personally always pin just in case, but I won't force my own paranoia on users! 😂

Dockerfile
Nezteb: I can address that! In general if you are using a minimal image (like [most `alpine` images](https://www.alpinelinux.org/about/)) there's not a lot a root user can do in the container if they gained access. Nevertheless, I agree because of "principle of least privilege" and all that. 😄

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
Nezteb: Yeah I made these files manually because I'm still new to TiddlyWiki and its markup. I'll fix that!

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
Nezteb: I can do that!

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
Nezteb: Agreed, I'll change that!

editions/tw5.com/tiddlers/docker/Using TiddlyWiki with Docker.tid
Nezteb: Okay, in that case I'll do some of the security stuff and cleanup that @pmario suggested and create that script!

Nezteb:
> with apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md?rgh-link-date=2023-04-05T20%3A18%3A21Z)).

No apologies necessary! I didn't read the contributing doc, oops. 🤦‍♂️ I'll do that!

Jermolene:
Thank you @Nezteb, that's great and much appreciated.

> Also if you'd like, once all of this is set up and ready to go I can walk you through the setup and such on a video call (or I'll screen record it and email it to you).
> 
> I originally started this because I'm going to deploy my own TiddlyWiki on https://fly.io/ via Docker, so I can show you that process as a demo. 😄

Terrific thank you – if you're able, it would be very useful to make a brief screencast for inclusion in the documentation.

Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN apk add --no-cache tini
+RUN npm install -g tiddlywiki@${TIDDLYWIKI_VERSION}
```
btheado, line 13: I wonder if it would be better to not have a dependency on npm here. Since this Dockerfile is being included with the git repository all the source code is already available, I would expect the docker image to be built using this local source (i.e. using the docker COPY command) rather than from whatever happens to be in npm.

Especially if some github actions CI job were created to build the docker image. It doesn't seem as useful to checkout the entire TW git repository, but then use none of the source code except for the Dockerfile?

But the result from using `COPY` would require invoking it as `nodejs tiddlywiki.js` rather than just `tiddlywiki`. Maybe that is ok or maybe there is some simple way around this.

I'm just thinking out loud here. Any other thoughts?

Nezteb:
(Closing as my free time has become more limited since I started this. 😓🙏)

PR #7394, opened 2023-04-05
Use Prettier to ensure consistent code formatting

### LINKED ISSUE #4625: [IDEA] Use Prettier

The formatting of JavaScript in TIddlyWiki is rather unconventional, and discussing formatting within a community is always a headache.

I think the current formatting can be quite off-putting and as a consequence be a barrier to contributions (it certainly was for me when I first looked at the code of TW).

Prettier (https://prettier.io/) is a formatting tool that has been widely adopted by OSS projects.
The main advantage I see it that there's no bike-shedding when using prettier, as it has (on purpose) almost no options.

The formatted code also follows widespread conventions, it has plugins for most text editors, can be applied automatically on a codebase, and can be setup as a CI job.

**What would TW JS code look like?**

Today TW code looks like this:

```js
var Commander = function(commandTokens,callback,wiki,streams) {
	var path = require("path");
	this.commandTokens = commandTokens;
	this.nextToken = 0;
	this.callback = callback;
	this.wiki = wiki;
	this.streams = streams;
	this.outputPath = path.resolve($tw.boot.wikiPath,$tw.config.wikiOutputSubDir);
};
```

With Prettier, it would look like the following:

```js
var Commander = function (commandTokens, callback, wiki, streams) {
  var path = require("path");
  this.commandTokens = commandTokens;
  this.nextToken = 0;
  this.callback = callback;
  this.wiki = wiki;
  this.streams = streams;
  this.outputPath = path.resolve(
    $tw.boot.wikiPath,
    $tw.config.wikiOutputSubDir
  );
};
```

**How can I help?**

If accepted, I would be willing to contribute by:
* setting up the repository to use Prettier
* fix linting rules is needed
* setting up Prettier to be run as part of CI jobs

Jermolene:
This commit installs and configures Prettier, and adds it to our test suite.

We do not yet apply Prettier. To see the changes that it would make, run the command `npm run prettier-write`.

Fixes #4625


linonetwo:
Prettier is good, but many web frontend already switch to [dprint](https://github.com/dprint/dprint#readme)! And use https://github.com/so1ve/eslint-plugin-dprint-integration for eslint.

Jermolene:
Thanks @linonetwo that's very helpful.

The main criteria for the choice is popularity because we want to maximise the chances that contributors will be familiar with whatever we choose. The secondary criteria is stability: we're here for the long term, and so we want to make choices that will last through the decades.

It looks like maybe it is still early days for dprint but if the advantages over Prettier are compelling we should definitely consider it. Would you be able to prepare a PR like this one but for dprint to make it easier to evaluate?

pmario:
I think we should configure eslint properly and skip prettier. We do have a very basic eslint configuration. We only need configure it in the way we want it and then enforce it.

Jermolene:
> I think we should configure eslint properly and skip prettier. We do have a very basic eslint configuration. We only need configure it in the way we want it and then enforce it.

Does eslint have an equivalent to `npm run prettier-write` though?

pmario:
yes, there are the --fix and --fix-dryrun parameters ... https://eslint.org/docs/latest/use/command-line-interface#fix-problems

using --fix-dry-run in combination with --format it would be possible to custom formatters.

-----

As I wrote somewhere else I think the airbnb ES5 rules are already very close to the existing TW js-style. https://github.com/airbnb/javascript/blob/es5-deprecated/es5/README.md

We could use their existing docs (which imo is very good) with a view changes to fit our needs.

linonetwo:
with eslint, you still need prettier. using eslint-prettier plugin. eslint itself deprecate formatting https://typescript-eslint.io/linting/troubleshooting/formatting/ because it is slow.

And it is correct to use eslint --fix to run eslint-prettier.

linonetwo:
I'm busy doing AI stuff, so maybe I will provide dpring example a few days later, and we can compare the speed of them.

prettier is like webpack, it was thriving, but now we think it is too slow, so we turn to rust based solutions like swc and dprint.

There is no much thing a formatter can do, so it can be regarded as matured.

pmario:
related comments, that are hard to find: https://github.com/Jermolene/TiddlyWiki5/commit/19c052d1380499ddbf75c78d7a878800a4ce8a39#comments

Jermolene:
> I'm busy doing AI stuff, I think AI is the key to make tiddlywiki popular, 

Good to hear.  I want to be able to run an LLM locally on my own computer, train it on my notes and interact with it by chat and autocomplete. 

> so maybe I will provide dpring example a few days later, and we can compare the speed of them in the CI.

Many thanks that would be much appreciated.

> prettier is like webpack, it was thriving, but now we think it is too slow, so we turn to rust based solutions like swc and dprint. And there is no much thing a formatter can do, so it can be regarded as matured.

I hadn't appreciated that dprint was written in Rust, that certainly makes it more attractive.

Jermolene:
I think this is obsolete with @pmario's recent experiments.

PR #7398, opened 2023-04-08
Request permission to protect local storage from eviction

### LINKED ISSUE #7369: [IDEA] Change the browser storage plugin to request persistence

Due to the possibility of [eviction](https://web.dev/storage-for-the-web/#eviction), localstorage used by the browser storage plugin can't be considered dependable.

According to [this page](https://web.dev/persistent-storage/):
> Thankfully, research by the Chrome team shows that data is very rarely cleared automatically by Chrome. It is far more common for users to manually clear storage. 

But "rarely" is not particularly reassuring. Also the [eviction](https://web.dev/storage-for-the-web/#eviction) link alarmingly says:

> Starting in iOS and iPadOS 13.4 and Safari 13.1 on macOS, there is a seven-day cap on all script writable storage, including IndexedDB, service worker registration, and the Cache API. This means Safari will evict all content from the cache after seven days of Safari use if the user does not interact with the site. This eviction policy does not apply to installed PWAs that have been added to the home screen.

The way I read that, the browser storage plugin has far too much risk of data loss to use on Safari.

According to https://web.dev/persistent-storage/ there is a browser API which can be called to request permission for the local storage data to be protected from eviction. The result of the request can be stored in a tiddler. If the request was denied then the code in the plugin would know not to request permission again. That tiddler should be one which is saved into local storage, but is ignored by the saver. Otherwise the saved wiki would be polluted by the original response.

I was unable to think of any other alternatives.

btheado:
Add code to the browser storage plugin to request permission for the local storage to be protected from eviction.

Some browsers such as firefox will prompt the user. To make the decision easier for the user only prompt them when they click the save button on a tiddler which will be stored to localstorage.

Other browsers will not prompt the user and will instead automatically approve or deny the request based on other factors.
 
Resolves #7369.

### DIFF
--- a/plugins/tiddlywiki/browser-storage/startup.js
+++ b/plugins/tiddlywiki/browser-storage/startup.js
@@ -53,6 +53,24 @@ exports.startup = function() {
 		$tw.wiki.addTiddler({title: ENABLED_TITLE, text: "no"});
 		$tw.browserStorage.clearLocalStorage();
 	});
+	// Request the browser to never evict the localstorage. Some browsers such as firefox
+	// will prompt the user. To make the decision easier for the user only prompt them
+	// when they click the save button on a tiddler which will be stored to localstorage.
+	var persistPermissionRequested = false;
+	$tw.hooks.addHook("th-saving-tiddler", function(tiddler) {
+		if (navigator.storage && navigator.storage.persist && !persistPermissionRequested) {
+			var filteredChanges = filterFn.call($tw.wiki, function(iterator) {
+				iterator(tiddler,tiddler.getFieldString("title"));
+			});
+			if (filteredChanges.length > 0) {
+				navigator.storage.persist().then(function(persisted) {
+					console.log("Request for persisted storage " + (persisted ? "granted" : "denied"));
+				});
+				persistPermissionRequested = true;
+			}
+		}
+		return tiddler;
+	});
 	// Track tiddler changes
 	$tw.wiki.addEventListener("change",function(changes) {
 		// Bail if browser storage is disabled
--- a/plugins/tiddlywiki/browser-storage/startup.js
+++ b/plugins/tiddlywiki/browser-storage/startup.js
@@ -54,36 +54,47 @@ exports.startup = function() {
 		$tw.wiki.addTiddler({title: ENABLED_TITLE, text: "no"});
 		$tw.browserStorage.clearLocalStorage();
 	});
+	// Helpers for protecting storage from eviction
+	var setPersistedState = function(state) {
+			$tw.wiki.addTiddler({title: PERSISTED_STATE_TITLE, text: state});
+		},
+		requestPersistence = function() {
+			setPersistedState("requested");
+			navigator.storage.persist().then(function(persisted) {
+				console.log("Request for persisted storage " + (persisted ? "granted" : "denied"));
+				setPersistedState(persisted ? "granted" : "denied");
+			});
+		},
+		persistPermissionRequested = false,
+		requestPersistenceOnFirstSave = function() {
+			$tw.hooks.addHook("th-saving-tiddler", function(tiddler) {
+				if (!persistPermissionRequested) {
+					var filteredChanges = filterFn.call($tw.wiki, function(iterator) {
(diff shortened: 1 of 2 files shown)


btheado:
The wiki at https://btheado.github.io/tw-browser-storage/ contains the code from this pull request if you want to test it without building it yourself.

plugins/tiddlywiki/browser-storage/startup.js
```diff
@@ -53,6 +54,48 @@ exports.startup = function() {
...
+			$tw.wiki.addTiddler({title: PERSISTED_STATE_TITLE, text: state});
+		},
+		requestPersistence = function() {
+			setPersistedState("requested");
+			navigator.storage.persist().then(function(persisted) {
+				console.log("Request for persisted storage " + (persisted ? "granted" : "denied"));
+				setPersistedState(persisted ? "granted" : "denied");
+			});
+		},
+		persistPermissionRequested = false,
```
pmario, line 68: I would prefer a `var` in front of all the variables and functions here. IMO it makes it easier to understand. ...

plugins/tiddlywiki/browser-storage/startup.js
```diff
@@ -53,6 +54,48 @@ exports.startup = function() {
...
 		$tw.browserStorage.clearLocalStorage();
 	});
+	// Helpers for protecting storage from eviction
+	var setPersistedState = function(state) {
+			$tw.wiki.addTiddler({title: PERSISTED_STATE_TITLE, text: state});
+		},
+		requestPersistence = function() {
+			setPersistedState("requested");
+			navigator.storage.persist().then(function(persisted) {
+				console.log("Request for persisted storage " + (persisted ? "granted" : "denied"));
```
pmario, line 64: IMO logging is a bit excessive. TW has a logger utility function. may be this one can be used. .. But I do not really know how to use it. https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/utils/logger.js 

@Jermolene Is there some docs, how it should be used?

plugins/tiddlywiki/browser-storage/startup.js
btheado: Thanks @pmario. I thought I was following the convention I've seen in the other tiddlywiki code. All the variables seem to be initialized using a single `var`. Are you thinking this is a special case because there are multiple function definitions involved here?

plugins/tiddlywiki/browser-storage/startup.js
btheado: Are you suggesting I remove the log statement? The state is being captured in a tiddler, so it is redundant and maybe not needed?

I see elsewhere in this file, the logger.js you mention is already being instantiated:

```js
	var logger = new $tw.utils.Logger("browser-storage",{
			colour: "cyan"
		});
```
But the only place that logger is being used is when browser storage gets a `QuotaExceededError`. There are many other places in browser storage where console.log is being called directly.

plugins/tiddlywiki/browser-storage/startup.js
Jermolene: The core does indeed generally use a single `var` to declare multiple variables

plugins/tiddlywiki/browser-storage/startup.js
Jermolene: Hi @pmario @btheado I think that given that the browser-storage plugin already makes liberal use of `console.log` it's reasonable to continue to use it.

pmario: I'm OK with liberal use of console.log ... but I want to deactivate is somehow. It's a pain to debug eg: the server version, because every minute it spams the log with new info with no real value.

Jermolene: > I'm OK with liberal use of console.log ... but I want to deactivate is somehow. It's a pain to debug eg: the server version, because every minute it spams the log with new info with no real value.

I don't think that that concern applies with the browser storage plugin because it doesn't run under Node.js

btheado:
Is this pull request a candidate for the upcoming release?

Jermolene:
Thanks @btheado

PR #7406, opened 2023-04-19
Add geospatial plugin

Jermolene:
This is an early version of the new Geospatial Plugin for TiddlyWiki.

The goal is to add new primitives to the TiddlyWiki platform that enable non-developers to build sophisticated interactive geospatial applications.

[Click here to see the preview of the demo edition for the Geospatial Plugin](https://tiddlywiki5-git-geospatial-plugin-jermolene.vercel.app/plugins/tiddlywiki/geospatial/index.html)

The Geospatial Plugin incorporates a number of third party libraries and online services:

* Integrating the open source [Leaflet.js](https://leafletjs.com/) library to display interactive maps
* Integrating the open source [Turf.js](https://turfjs.org/) library to perform geospatial calculations with [GeoJSON](https://en.wikipedia.org/wiki/GeoJSON) objects
* Integrating the commercial API [TravelTime](https://traveltime.com/) for [geocoding](https://traveltime.com/features/geocoding), [routing](https://traveltime.com/features/multi-modal-routing) and [isochrones](https://traveltime.com/features/isochrones)
* Integrating the free API from [Flickr](https://www.flickr.com/services/api/) for retrieving geotagged photographs
* Integrating Google's open source [OpenLocationCode](https://github.com/google/open-location-code) library for converting to and from Open Location Codes (also known as [PlusCodes](https://maps.google.com/pluscodes/))

This PR also includes a number of core improvements that will be cherry picked into the core before merging. See the [preview](https://tiddlywiki5-git-geospatial-plugin-jermolene.vercel.app/) to explore those changes


core/ui/TestCases/DefaultTemplate.tid
```diff
@@ -0,0 +1,22 @@
+title: $:/core/ui/testcases/DefaultTemplate
+
+\whitespace trim
```
pmario, line 3: You used `\whitespace trim` but there is no indentation for improved readability. 

IMO it would be nice to have indented templates. 

There are many tiddlers with no indent.

core/modules/widgets/data.js
```diff
@@ -0,0 +1,144 @@
...
+*/
+DataWidget.prototype.extractCompoundTiddler = function(title) {
+	var tiddler = this.wiki.getTiddler(title);
+	if(tiddler && tiddler.fields.type === "text/vnd.tiddlywiki-multiple") {
+		var text = tiddler.fields.text || "",
+			rawTiddlers = text.split("\n+\n"),
+			tiddlers = [];
+		$tw.utils.each(rawTiddlers,function(rawTiddler) {
+			var fields = Object.create(null),
+				split = rawTiddler.split(/\r?\n\r?\n/mg);
```
pmario, line 117: `rawTiddlers` are split with 2 or more NLs ... `text.split("\n+\n")` 
can `rawTiddler` be split with 2 new-lines again `rawTiddler.split(/\r?\n\r?\n/mg);` ... Did I misread something here?

editions/tw5.com/tiddlers/____ Authoring Tools ____.tid
```diff
@@ -0,0 +1,115 @@
+created: 20221009000000000
+modified: 20221019081352266
+tags: 
+title: **** Authoring Tools ****
```
pmario, line 4: What's this tiddler about?

core/modules/widgets/data.js
Jermolene: Thanks @pmario – `text.split("\n+\n")` is not a regular expression, it's matching a plus sign surrounded by newlines. I think you're right that we should include support for carriage returns.

editions/tw5.com/tiddlers/widgets/TestCaseWidget.tid
```diff
@@ -0,0 +1,46 @@
...
+
+<$testcase>
+<$data $tiddler="$:/core/ui/testcases/DefaultTemplate"/>
+<$data $tiddler="$:/core/ui/testcases/DefaultTemplate/Source"/>
+<$data title="Description" text="Simple example of a test case"/>
+<$data title="Output" text="""<$testcase>
+<$data title="Description" text="How to calculate 2 plus 2"/>
+<$data title="Output" text="<$text text={{{ =2 =[add[2]] }}}/>"/>
+</$testcase>
+"""/>
```
btheado, line 45: I notice at https://tiddlywiki5-git-geospatial-plugin-jermolene.vercel.app/index.html#TestCaseWidget if I edit any text in the Output tab, then after every refresh, a new "How to calculate 2 plus 2" card gets appended on the right side. I expect the refreshed output to replace the "How to calculate 2 plus 2".

Also, is it intentional your filter does not give "4" as the result (it gives "2")? If not, then I think you need `=2 :and[add[2]] ` or `[[2]add[2]]`

Jermolene: Thanks @btheado I've pushed a fix to the refresh problem in 03b6413c7a519429ea75353faf32c0a0baf2b88e and the other issue in a91a6f602c16a8fb48a489608d5e3d5868b7e427.

editions/tw5.com/tiddlers/widgets/TestCaseWidget.tid
```diff
@@ -0,0 +1,46 @@
...
+! Introduction
+
+The testcase widget creates an independent subwiki loaded with the specified payload tiddlers and then renders a specified template from within the subwiki.The default template displays a split view with the source tiddlers on the left and the rendered tiddler titled `Output` on the right. It also displays the tiddler titled `Description` as the heading. This makes it possible to run independent tests that also serve as documentation examples.
+
+! Content and Attributes
+
+The content of the `<$testcase>` widget is not displayed but instead is scanned for <<.wlink DataWidget>> widgets that define the payload tiddlers to be included in the test case.
+
+|!Attribute |!Description |
+|<<.attr template>> |Optional title of the template used to display the testcase (defaults to $:/core/ui/testcases/DefaultTemplate) |
```
btheado, line 17: IMO, this description would be better if it mentioned that the template tiddler must be present in the subwiki. It took me a while to figure out that I needed to use `$data` widget to load my template into the subwiki.

Or maybe it would be better to automatically load the template tiddler if it isn't already present in the subwiki (i.e. because of the hard-coded `loadTiddler("$:/core");`)?

core/ui/TestCases/DefaultTemplate.tid
```diff
@@ -0,0 +1,22 @@
...
+<$let
+	state={{{ [<qualify "$:/state/testcase">] }}}
+>
+<div class="tc-testcase-wrapper">
+<div class="tc-testcase-header">
+<h2><$transclude tiddler="Description" mode="inline"/></h2>
+</div>
+<div class="tc-testcase-panes">
+<div class="tc-testcase-source">
+<$macrocall $name="tabs" tabsList="[all[tiddlers]sort[]] -Description -ExpectedResult -Output Output +[putfirst[]] -[has[plugin-type]]" state=<<state>> default="Output" template="$:/core/ui/testcases/DefaultTemplate/Source"/>
```
btheado, line 13: Add `-[prefix<state>]` to this filter? That way a new tab like `$:/state/testcase--590634553--590634553` doesn't get created after switching tabs.

Or maybe even more should be excluded with: `-[prefix[$:/state/]] -[prefix[$:/temp/]]`

btheado: I agree. That's why my first suggestion was to exclude just the tab state tiddlers with `-[prefix<state>]`. Thinking from the perspective of using this widget in the tiddlywiki.com docs, it seems like > 99.9% of the time the tab state tiddler will be clutter as you say.

If a more aggressive filter like my second suggestion is made the default, then maybe some or all of the filter can be made configurable via a config tiddler somehow.

I know a custom template can be provided to achieve full customization, but it seems like wanting to customize just the filter will be a common use case.

Jermolene: Thanks @btheado, you're quite right, `-[prefix<state>]` works well, I'll push that momentarily.

PR #7411, opened 2023-04-22
Documents different ways of defining and invoking variables

btheado:
- Ways to define variables and parameters
- Behavior of invoked variables depends on how the variable was declared
- Namespaces


Jermolene:
Fantastic, thank you @btheado

PR #7421, opened 2023-04-28
Add a `:boolean` filter run prefix

yaisog:
I propose to add a `:boolean` filter run prefix that allows for various boolean operations on filter runs, or more accurately on the result of all previous filter runs and the prefixed filter run. For boolean operations, a run is _true_ if it contains at least one item, possibly an empty string, and _false_ if it is empty.
Consequently, the output of the `:boolean` filter run is either the string "true" or nothing. Thus, nesting is also possible, e.g. "A and (B or C)" via subfilters.
As with all filter operations that take a filter expression as parameter, it is cleaner to use new filter run prefixes than new operators.
The boolean operation is passed via a suffix (`:or`, `:nor`, `:xor`, `:xnor`, `:and`, `:nand`) and defaults to `:and`.
Advanced usage could look like this:
```
\define subcondition() [tag[2]] :boolean:or[tag[3]]

{{{ [<currentTiddler>tag[1]] :boolean:and[<currentTiddler>subfilter<subcondition>] }}}
```
with some tags 1, 2 and 3.

For some more background, refer to [this TW Talk discussion](https://talk.tiddlywiki.org/t/learning-from-my-mistakes-logical-and/6849).

@Jermolene: Should this not be deemed universal enough for core inclusion, I'll be happy to just make a plugin out of it.

To-do:
- [x] Create documentation with examples
- [ ] Possibly consider other labels than `:boolean` to be more intuitive for non-programmers
- [x] Implement short-circuit functionality where applicable (and, or, ...)
- [ ] Add a short-circuit example to the docs
- [ ] Add inline documentation styles to doc-styles
- [ ] Add tests
- (currently not possible) Integrate with existing general documentation on FRPs
- (currently not possible) Consolidate docs with #7368 after merging


yaisog:
I apologize in advance for not making an issue first, but I would like to avoid the retyping invariably resulting from duplicating the discussion on TW Talk.

Jermolene:
Hi @yaisog it's OK to jump straight to producing a PR without a preceding ticket, and it often makes sense as the clearest way to communicate the intent of a change. The only downside is that it slightly increases the risk that the work will be rejected.

In this case, I like the idea of offering clear access to all the boolean operations. However I have a couple of concerns:

* It is confusing to add `boolean:and` and `boolean:or` when we already have plain "and" and "or" filter run prefixes. Are the semantics different? If the semantics are the same then perhaps we could promote all these boolean operators to be separate filter run prefixes
* Having to repeatedly include the word `boolean` feels clumsy

yaisog:
> Are the semantics different?

The input to an `:and` run is the output of all previous runs. The input to `:boolean:and` should be all tiddlers. The runs are not supposed to depend on one another. So yes, the semantics are different. These two filter expressions will give different outputs on the preview site (even when the output is "translated" to boolean):
```
{{{ [tag[WikiText]] :boolean:or[tag[TableOfContents]] }}}
{{{ [tag[WikiText]] :and[tag[TableOfContents]] }}}
```
The first filter checks if there are tiddlers with the tag _WikiText_ and if there are tiddlers with the tag _TableOfContents_.
The second filter checks if there are tiddlers that have both tags (there are none).

For `:or` and `:boolean:or` only the output is different. The Boolean operation will output _true_ while `:or` will output a list of all titles that fulfill either condition. This can be important when used in a `$list` widget when its purpose is not to iterate over a list of tiddlers, but only to conditionally show or hide some content. In the `:or` case a `compare:number:gt[0]` may be required.

Basically, the `:boolean` filter run prefix is optimized for Boolean operations where one may not want to worry about the number of results that may be output.

> Having to repeatedly include the word `boolean` feels clumsy

True. I'm all open to suggestions. Maybe `:bool` would be at least a bit shorter?

yaisog:
I just added a rather lengthy doc tiddler and an example tiddler.

The doc tiddler uses the overview table from #7368, but since that is not yet merged, I had to duplicate the table code. This must be taken out when #7368 is merged to avoid duplicate tables.
Also, for the same reason I cannot add this run prefix to the general run prefix documentation.
I used two additional (inline) styles that I might want to add to the doc-styles when I do the cleanup described above.

yaisog:
The run prefix will now make use of JavaScript short-circuiting, so that the prefixed filter run is not evaluated if unnecessary. This might speed up operation in a number of cases, in contrast with previous workarounds, as described in the TWTalk thread linked above.

yaisog:
Hi @Jermolene, before I start adding tests, I'd like to finalize the run prefix name. Since you felt that `:boolean:op` was too clumsy, maybe you have an alternative in mind? Using `:bool:op` would at least be shorter, so that's a candidate?

If possible I'd like to keep the suffix instead of creating six different prefixes. I think it's a case where the suffix functionality can really shine. Currently, leaving out the suffix corresponds to `:boolean:and` which I felt was the most common application.

Others' proposals are also welcome, of course. @saqimtiaz, @pmario?

Jermolene:
Hi @yaisog apologies, I should have provided some feedback sooner.

I think this should be explored as a plugin in the first instance. My concern is that we already have a convention for representing boolean values, as seen with the "then" and "else" filter operators, but this filter run prefix works with a different convention. I think that makes the documentation and learning process for TiddlyWiki as a whole significantly more complex.

Publishing it as a plugin will help us gather feedback on whether the new approach is useful for our users.

Jermolene:
> Hi @Jermolene, before I start adding tests, I'd like to finalize the run prefix name. Since you felt that `:boolean:op` was too clumsy, maybe you have an alternative in mind? Using `:bool:op` would at least be shorter, so that's a candidate?

I think it's probably best to use "boolean" at first; it's very clear and you can always add an abbreviation later.

> If possible I'd like to keep the suffix instead of creating six different prefixes. I think it's a case where the suffix functionality can really shine. Currently, leaving out the suffix corresponds to `:boolean:and` which I felt was the most common application.

I think it might be worth encouraging authors to use an explicit suffix. While "add" might well be the most common suffix, the meaning of `:boolean[...]` would not be at all obvious.

yaisog:
> but this filter run prefix works with a different convention.

Hi @Jermolene, actually it doesn't. Just like `:then` and `:else`, a non-empty list represents true. Only the output is different: for `:then` and `:else` it's the result of the prefixed filter run, for `then[]` and `else[]` it's the filter parameter, and here it is the fixed string `true` or nothing. Which is the only sensible option, since e.g. an element-wise Boolean operation on two sets is not really what one wants when checking if two separate conditions are true.

Anyway, I'll make this into a plugin one of these days, like I suggested in the OP. I'm going to pick up your suggestion and leave the usage without suffix out of the docs.

PR #7422, opened 2023-04-29
Add tm-http-request message for making HTTP requests

Jermolene:
This is a simple way to trigger HTTP requests via a message. Headers and query string parameters can be specified and action strings can be specified to be invoked with progress reports and at completion.

To think about:

* Do we need logging of HTTP requests within the wiki, or do we expect users to rely on the network tab of developer tools?
* Do we need support for binary request and response bodies?
* It is possible for a completion handler to initiate another HTTP request; this chaining is necessary for APIs that use pagination (see the include Zotero demo for example). However, it also sets up the conditions where it is very easy to accidentally trigger an infinite loop. We could address this by providing an "interrupt button" that cancels any pending HTTP requests with an error code. Ideally the button would only appear during HTTP transfers

# Documentation

<div class="tc-tiddler-body tc-reveal"><p>The <strong>tm-http-request</strong> message is used to make an HTTP request to a server.</p><p>It uses the following properties on the <code>event</code> object:</p><table><tbody><tr class="evenRow"><th align="left">Name</th><th align="left">Description</th></tr><tr class="oddRow"><td align="left">param</td><td align="left">Not used</td></tr><tr class="evenRow"><td align="left">paramObject</td><td align="left">Hashmap of parameters (see below)</td></tr></tbody></table><p>The following parameters are used:</p><table><tbody><tr class="evenRow"><th align="left">Name</th><th align="left">Description</th></tr><tr class="oddRow"><td align="left">method</td><td align="left">HTTP method (eg "GET", "POST")</td></tr><tr class="evenRow"><td align="left">body</td><td align="left">String data to be sent with the request</td></tr><tr class="oddRow"><td align="left">query-*</td><td align="left">Query string parameters with string values</td></tr><tr class="evenRow"><td align="left">header-*</td><td align="left">Headers with string values</td></tr><tr class="oddRow"><td align="left">password-header-*</td><td align="left">Headers with values taken from the password store</td></tr><tr class="evenRow"><td align="left">password-query-*</td><td align="left">Query string parameters with values taken from the password store</td></tr><tr class="oddRow"><td align="left">var-*</td><td align="left">Variables to be passed to the completion and progress handlers (without the "var-" prefix)</td></tr><tr class="evenRow"><td align="left">bind-status</td><td align="left">Title of tiddler to which the status of the request ("pending", "complete", "error") should be bound</td></tr><tr class="oddRow"><td align="left">bind-progress</td><td align="left">Title of tiddler to which the progress of the request (0 to 100) should be bound</td></tr><tr class="evenRow"><td align="left">oncompletion</td><td align="left">Action strings to be executed when the request completes</td></tr><tr class="oddRow"><td align="left">onprogress</td><td align="left">Action strings to be executed when progress is reported</td></tr></tbody></table><p>The following variables are passed to the completion handler:</p><table><tbody><tr class="evenRow"><th align="left">Name</th><th align="left">Description</th></tr><tr class="oddRow"><td align="left">status</td><td align="left">HTTP result status code (see <a class="tc-tiddlylink-external" href="https://developer.mozilla.org/en-US/docs/Web/HTTP/Status" rel="noopener noreferrer" target="_blank">MDN</a>)</td></tr><tr class="evenRow"><td align="left">statusText</td><td align="left">HTTP result status text</td></tr><tr class="oddRow"><td align="left">error</td><td align="left">Error string</td></tr><tr class="evenRow"><td align="left">data</td><td align="left">Returned data</td></tr><tr class="oddRow"><td align="left">headers</td><td align="left">Response headers as a JSON object</td></tr></tbody></table><p>The following variables are passed to the progress handler:</p><table><tbody><tr class="evenRow"><th align="left">Name</th><th align="left">Description</th></tr><tr class="oddRow"><td align="left">lengthComputable</td><td align="left">Whether the progress loaded and total figures are valid - "yes" or "no"</td></tr><tr class="evenRow"><td align="left">loaded</td><td align="left">Number of bytes loaded so far</td></tr><tr class="oddRow"><td align="left">total</td><td align="left">Total number bytes to be loaded</td></tr></tbody></table></div>


Jermolene:
I've updated the OP to add a note about providing an interrupt button for chained HTTP requests:

> * It is possible for a completion handler to initiate another HTTP request; this chaining is necessary for APIs that use pagination (see the include Zotero demo for example). However, it also sets up the conditions where it is very easy to accidentally trigger an infinite loop. We could address this by providing an "interrupt button" that cancels any pending HTTP requests with an error code. Ideally the button would only appear during HTTP transfers

pmario:
I think "interrupt" a running request is important anyway. Not only for endless loops. Timeouts usually are at about 90 seconds. I'm pretty sure our users will fire the "retry" button 20 times before that timeout. IMO there needs to be a possibility to "fix" open / pending responses

Jermolene:
I've added a new page control button that indicates network activity, and allows outstanding requests to be cancelled.

It looks like this in the inactive state:

image

This is the active state:

image

Use the [Zotero demo](https://tiddlywiki5-git-tm-http-request-message-jermolene.vercel.app/#WidgetMessage%3A%20tm-http-request%20Example%20-%20Zotero) to try it out.

core/modules/utils/dom/http.js
```diff
@@ -13,11 +13,201 @@ Browser HTTP support
...
+				self.wiki.addTiddler(new $tw.Tiddler(self.wiki.getTiddler(requestTrackerTitle),{
+					status: completionCode,
+				}));
+				self.wiki.invokeActionString(self.completionActions,undefined,$tw.utils.extend({},self.variables,resultVariables),{parentWidget: $tw.rootWidget});
+				callback(hasSucceeded ? null : xhr.statusText);
+				// console.log("Back!",err,data,xhr);
+			},
+			progress: function(lengthComputable,loaded,total) {
+				if(lengthComputable) {
+					setBinding(bindProgress,"" + Math.floor((loaded/total) * 100))
```
btheado, line 184: The zotero pos252 example is giving me `Uncaught ReferenceError: bindProgress is not defined`. Maybe you mean this?

```suggestion
					setBinding(self.bindProgress,"" + Math.floor((loaded/total) * 100))
```

saqimtiaz:
> I've added a new page control button that indicates network activity, and allows outstanding requests to be cancelled.

I imagine that we eventually want something akin to a network connectivity tiddler, which shows all ongoing requests and allows users to cancel or debug them. I mention this here since the icon used in this PR might be appropriate in that context as well and we would need the two icons to be distinct enough.

@Jermolene have you considered using the state tiddler title as a means to identify which request to cancel, so that requests could be selectively cancelled instead of cancelling all outstanding requests?

I haven't delved too deeply into the implementation for the cancelling requests however the following does bring up a red "Internal JavaScript Error" in the [zotero demo](https://tiddlywiki5-git-tm-http-request-message-jermolene.vercel.app/#WidgetMessage%3A%20tm-http-request%20Example%20-%20Zotero) with the message `"Cannot read properties of undefined (reading 'request')"`:
- Double click "Start import from Zotero group"
- Immediately click "Cancel all HTTP requests".

I have not investigated but I wonder if this is prompted by a request completing while the cancel operation is in progress.

Jermolene:
> I imagine that we eventually want something akin to a network connectivity tiddler, which shows all ongoing requests and allows users to cancel or debug them. I mention this here since the icon used in this PR might be appropriate in that context as well and we would need the two icons to be distinct enough.

That's the path of the work I've been doing, but I'm deferring working on logging and debugging requests. The network activity button currently cancels all requests but in the future it could bring up a dialogue showing the details of the outstanding requests.

The HTTP infrastructure currently tracks the number of outstanding requests in a tiddler which is how the network activity button switches between states.

I'm not sure what you mean about the icon design; the icon indicates network activity.

> @Jermolene have you considered using the state tiddler title as a means to identify which request to cancel, so that requests could be selectively cancelled instead of cancelling all outstanding requests?

The code includes the support for cancelling individual requests but there's no UI at the moment. I'm not sure that it would be particularly useful anyway; the key motivation for this UI is to give users a way to cancel runaway recursive requests. Cancelling individual requests seems like a function for very advanced users.

> I haven't delved too deeply into the implementation for the cancelling requests however the following does bring up a red "Internal JavaScript Error" in the [zotero demo](https://tiddlywiki5-git-tm-http-request-message-jermolene.vercel.app/#WidgetMessage%3A%20tm-http-request%20Example%20-%20Zotero) with the message `"Cannot read properties of undefined (reading 'request')"`:
> 
> * Double click "Start import from Zotero group"
> * Immediately click "Cancel all HTTP requests".
> 
> I have not investigated but I wonder if this is prompted by a request completing while the cancel operation is in progress.

No, it's more basic than that; the double click generates two outstanding requests and there was a bug with cancelling multiple requests in one go. Fixed now.

saqimtiaz:
> The network activity button currently cancels all requests but in the future it could bring up a dialogue showing the details of the outstanding requests.

This is the area that I was trying to draw attention to. Either we need to accept any potential confusion that might come with the same button invoking a different behaviour in the future, or we would need a different icon for the network activity dialog once that is implemented. I can imagine users integrating the page control button in a different UI and there being backwards compatibility constraints to changing the behaviour of the button.

Jermolene:
> the double click generates two outstanding requests

This is a pervasive problem in TiddlyWiki but particularly acute in the context of long-lived chains of HTTP requests.

One approach to fixing it would be to disable the request initiation button as soon as it is clicked, and then only re-enable it once the request has completed. The trouble with that solution is that it wouldn't work if there were two copies of the request initiation button visible at one time.

An alternative approach would be to require each HTTP request to be given a name. If a request of the same name is already outstanding, then we fail the second request. In some situations it might be useful to cancel the request in progress and initiate the request afresh. Perhaps an additional parameter could be used to select the desired failure mode.

A common pattern when dealing with paginated APIs is to initiate the next HTTP request within the completion handler of the previous one. To account for this, requests with the same name would be allowed to proceed if initiated inside the completion handler of a request with the same name, but would trigger the failure processing if initiated outside the completion handler.

It would also be useful to be able to disable these request initiation buttons while the request is in progress. Perhaps we could offer a filter operator that returns a hashmap of outstanding HTTP requests indexed by their name:

```
<$button disabled={{{ [jsonactivehttprequests[]jsonget[req-zotero-demo-1]then[yes]else[no]] }}} ...>
```

Requiring a unique name for each chain of HTTP requests would also have the benefit of allowing better debugging information to be provided (requests are currently identified internally by a numeric ID).

saqimtiaz:
> Requiring a unique name for each chain of HTTP requests would also have the benefit of allowing better debugging information to be provided (requests are currently identified internally by a numeric ID).

In my own implementation prior to adopting yours for carrying out HTTP requests from wikitext, each HTTP request was tied to a state tiddler which identified the request and also was used for the bind status. The same state tiddler was also queried via filters to determine if the button should be enabled or disabled, as well as to show progress messages, etc. 

Is there a compelling reason not to use state tiddlers to identify requests?

Jermolene:
> This is the area that I was trying to draw attention to. Either we need to accept any potential confusion that might come with the same button invoking a different behaviour in the future, or we would need a different icon for the network activity dialog once that is implemented. I can imagine users integrating the page control button in a different UI and there being backwards compatibility constraints to changing the behaviour of the button.

Interesting point. Backwards compatibility constraints mustn't prevent ongoing improvements. I think that if we were to decide to change the UI of this button to bring up a dropdown with additional options (including a button that cancels all requests), it should just be considered a routine improvement. Backwards compatibility is retained as long as the improvement doesn't disrupt existing wikitext and JS modules in the community.

One concern I have with exposing the full details of outstanding requests is that it will necessarily expose the information to the entire wiki. At the moment, it is not possible to write wikitext that exfiltrates passwords or other secrets stored with the `<$password>` widget (it can be done with JS, of course). Perhaps the fix is simple; in outgoing requests we could just hide the secrets in the JSON blobs exposed to the wiki. We wouldn't be able to hide any secrets that might appear in the responses of course.

For the moment, I think the best way to avoid the confusion might be to update the network activity button to make it bring up a dropdown menu with sub-options. Initially the only option might be to cancel all requests. (The interface I am describing exists in the client-server edition when clicking on the cloud icon – I felt that the radio mast icon was a good match for the cloud icon when used alongside one another).

Jermolene:
> In my own implementation prior to adopting yours for carrying out HTTP requests from wikitext, each HTTP request was tied to a state tiddler which identified the request and also was used for the bind status. The same state tiddler was also queried via filters to determine if the button should be enabled or disabled, as well as to show progress messages, etc.
> 
> Is there a compelling reason not to use state tiddlers to identify requests?

No, it's just that the implementation currently only has a very crude implementation of state tiddlers. The current implementation focusses on JS representation of entities because that will always be required; I would think of the state tiddlers as mirroring information from the JS domain. I think it gives a better experience for authors: writing to these state tiddlers has no meaning and so the expected result of doing so is that it reverts to the previous, correct value. That's how the `$:/status/UserName` tiddler is handled by the tiddlywebadaptor, for example.

PR #7425, opened 2023-05-05
Improve translink macro, fixes 4877

mateuszwilczek:
Changes:
- Change styles to adapt colours to the current palette, instead of hard-coded white inside black box with grey borders.
- Simplify the design to a single div (inspired by the Community links on tiddlywiki.com), instead of two nested contrasting divs.

Comments:
- I found pre-border and pre-background to be the most consistently readable across the default palettes (better than relying on e.g. tiddler-border, tiddler-background, page-background, code-border, code-background). Cupertino Dark, Gruvbox Dark, Nord, Solar Flare, Spartan Day, Spartan Night, Spartan Day, and Twilight have same or very similar pre-border and pre-background colours, so if the macro is nested (translinking a tiddler that contains a translink), the inner frames are not distinguishable, but I think it's an acceptable edge case.
- Padding is defined separately for block and inline modes of the macro to make it appear consistent.

### DIFF
--- a/core/wiki/macros/translink.tid
+++ b/core/wiki/macros/translink.tid
@@ -5,7 +5,7 @@ tags: $:/tags/Macro
 \whitespace trim
 <div class={{{ [[tc-translink]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[join[ ]] }}}>
 <$link to="""$title$""">
-<h1><$text text="""$title$"""/></h1>
+<h1 style="font-weight: bold; font-size: 1.2em;"><$text text="""$title$"""/></h1>
 </$link>
 <$transclude tiddler="""$title$""" mode="$mode$">
 "<$text text="""$title$"""/>" is missing
@@ -22,9 +22,6 @@ tags: $:/tags/Macro
 	padding: 0em 1em;
 }.tc-translink-inline {
 	padding: 0em 1em 1em;
-}.tc-translink h1 {
-	font-weight: bold;
-	font-size: 1.2em;
 }
 </style>
 \end
--- a/core/wiki/macros/translink.tid
+++ b/core/wiki/macros/translink.tid
@@ -3,7 +3,7 @@ tags: $:/tags/Macro
 
 \define translink(title,mode:"block")
 \whitespace trim
-<div class={{{ [[tc-translink]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[join[ ]] }}}>
+<div class={{{ [[tc-translink]] [[$title$]is[missing]then[tc-translink-inline]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[first[2]] +[join[ ]] }}}>
 <$link to="""$title$""">
 <h1 style="font-weight: bold; font-size: 1.2em;"><$text text="""$title$"""/></h1>
 </$link>
--- a/core/wiki/macros/translink.tid
+++ b/core/wiki/macros/translink.tid
@@ -3,12 +3,14 @@ tags: $:/tags/Macro
 
 \define translink(title,mode:"block")
 \whitespace trim
-<div class="pre-border" style="padding: 0 1em;">
+<div class="tc-pre-background">
+<div class="tc-big-v-gap tc-big-gap">
 <$link to="""$title$""">
 <h1 class="tc-big-bold"><$text text="""$title$"""/></h1>
 </$link>
-<p><$transclude tiddler="""$title$""" mode="$mode$">
+<$transclude tiddler="""$title$""" mode="$mode$">
 <$set name="currentTiddler" value="""$title$"""><$transclude tiddler="$:/language/MissingTiddler/Hint"/></$set>
-</$transclude></p>
+</$transclude>
+</div>
 </div>
 \end
--- a/core/wiki/macros/translink.tid
+++ b/core/wiki/macros/translink.tid
@@ -3,25 +3,13 @@ tags: $:/tags/Macro
 
 \define translink(title,mode:"block")
 \whitespace trim
-<div class={{{ [[tc-translink]] [[$title$]is[missing]then[tc-translink-inline]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[first[2]] +[join[ ]] }}}>
+<div class="pre-border" style="padding: 0 1em;">
 <$link to="""$title$""">
-<h1 style="font-weight: bold; font-size: 1.2em;"><$text text="""$title$"""/></h1>
+<h1 class="tc-big-bold"><$text text="""$title$"""/></h1>
 </$link>
 <$transclude tiddler="""$title$""" mode="$mode$">
 "<$text text="""$title$"""/>" is missing
 </$transclude>
+<$list filter="[<__title__>is[missing]] ~[<__mode__>match[inline]]"><p></p></$list>
 </div>
-<style>
-.tc-translink {
-    border: 1px solid;
-    border-color: <<colour "pre-border">>;
-    border-radius: 2px;
-    background: <<colour "pre-background">>;
-	padding: <<verticalPadding>> 1em;
-}.tc-translink-block {
-	padding: 0em 1em;
-}.tc-translink-inline {
-	padding: 0em 1em 1em;
-}
-</style>
 \end
--- a/core/wiki/macros/translink.tid
+++ b/core/wiki/macros/translink.tid
@@ -20,11 +20,9 @@ tags: $:/tags/Macro
 <$link to="""$title$""">
 <$text text="""$title$"""/>
 </$link>
-<span>
-<$transclude tiddler="""$title$""" mode="inline">
+&#32;(<$transclude tiddler="""$title$""" mode="inline">
 <$set name="currentTiddler" value="""$title$"""><$transclude tiddler="$:/language/MissingTiddler/Hint"/></$set>
-</$transclude>
-</span>
+</$transclude>)
 </span>
 </$list>
 \end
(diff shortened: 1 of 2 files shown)


core/wiki/macros/translink.tid
```diff
@@ -3,14 +3,25 @@ tags: $:/tags/Macro
 
 \define translink(title,mode:"block")
 \whitespace trim
-<div style="border:1px solid #ccc; padding: 0.5em; background: black; foreground; white;">
+<div class={{{ [[tc-translink]] [[$title$]is[missing]then[tc-translink-inline]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[first[2]] +[join[ ]] }}}>
```
Jermolene, line 6: We avoid using the textual substitution feature of macros unless absolutely necessary. In this case, using textual substitution means that tiddler titles containing `]]` will not behave correctly.

The alternative is to use the double underscore syntax to access the parameters as variables:

```
 <div class={{{ [[tc-translink]] [<__title__>]is[missing]then[tc-translink-inline]] [<__mode__>match[block]then[tc-translink-block]] [<__mode__>match[inline]then[tc-translink-inline]] +[first[2]] +[join[ ]] }}}>
```

core/wiki/macros/translink.tid
```diff
@@ -3,14 +3,25 @@ tags: $:/tags/Macro
...
 <$link to="""$title$""">
-<$text text="""$title$"""/>
+<h1 style="font-weight: bold; font-size: 1.2em;"><$text text="""$title$"""/></h1>
```
Jermolene, line 8: Would it make sense for these inline styles to go into the stylesheet?

core/wiki/macros/translink.tid
```diff
@@ -3,14 +3,25 @@ tags: $:/tags/Macro
...
 </$link>
-<div style="border:1px solid #ccc; padding: 0.5em; background: white; foreground; black;">
 <$transclude tiddler="""$title$""" mode="$mode$">
 "<$text text="""$title$"""/>" is missing
 </$transclude>
 </div>
-</div>
+<style>
```
Jermolene, line 14: The core doesn’t use stylesheets embedded in HTML like this. One problem with this approach is that if there are many repetitions of this macro displayed at once then the browser will get a bunch of redundant stylesheets.

The core macros that need classes can define them in the base.tid stylesheet in the Vanilla theme.

Jermolene (commented):
Thank you @mateuszwilczek.

mateuszwilczek:
Thank you @Jermolene for your comments, I have made changes to resolve the issues you mentioned.

I have changed the way the consistent padding between block/inline/missing situations is achieved. An empty paragraph `<p></p>` is introduced after the inline text that would otherwise (inline mode or missing tiddler message) not be a paragraph and not receive enough spacing. I think it makes more sense than fixing the spacing with different styles as I did before.

I have moved the class definitions to vanilla/base.tid:

- I have split the `pre` style definition, to be able to reuse part of it (border, border-radius, background-color) as the newly created class `tc-pre-border`.
- I have created `tc-big-bold` at the end under _Other utility classes_, as it could make sense to reuse it in other contexts (see Community links on tiddlywiki.com).

mateuszwilczek:
One more thing to improve is the lacking localization of the alternative missing message, I will work on that.

core/wiki/macros/translink.tid
```diff
@@ -3,25 +3,13 @@ tags: $:/tags/Macro
...
-<div class={{{ [[tc-translink]] [[$title$]is[missing]then[tc-translink-inline]] [[$mode$]match[block]then[tc-translink-block]] [[$mode$]match[inline]then[tc-translink-inline]] +[first[2]] +[join[ ]] }}}>
+<div class="pre-border" style="padding: 0 1em;">
 <$link to="""$title$""">
-<h1 style="font-weight: bold; font-size: 1.2em;"><$text text="""$title$"""/></h1>
+<h1 class="tc-big-bold"><$text text="""$title$"""/></h1>
 </$link>
 <$transclude tiddler="""$title$""" mode="$mode$">
 "<$text text="""$title$"""/>" is missing
 </$transclude>
+<$list filter="[<__title__>is[missing]] ~[<__mode__>match[inline]]"><p></p></$list>
```
kookma: Shouldn't be used `:else[<__mode__>match[inline]]` instead of `~[<__mode__>match[inline]]`?

pmario: `<$list filter="[<__title__>is[missing]] ~[<__mode__>match[inline]]"><p></p></$list>` .. what is the `<p></p>` good for. It will always be empty. So **if it is used to introduce vertical spacing that's the wrong way to do it**. There should a class assigend to the right elements.

mateuszwilczek: It would be clearer, yes, but in the meantime I have found a simpler solution that does not require this list widget at all.

mateuszwilczek: I have just changed it, the spacing is achieved the proper way now.

mateuszwilczek:
I have resolved all the issues mentioned by @Jermolene, @kookma, and @pmario.

Explanation of the changes to the vanilla/base.tid stylesheet:
- Some properties of the `pre` element are available under the new class `tc-pre-background`. This class is assigned to the outer `div` of translink macro, and could be used for block as well as inline elements to give them background and border same as `pre`.
- `tc-big-v-gap` among the *Other utility classes/Vertical gaps* follows the naming and margin sizes of analouge already existing horizontal spacing class `tc-big-gap`.
- `tc-big-bold` is a new class, placed under *Other utility classes/Font styles*. It is assigned to the `h1` with translinked tiddler's title. One example where this class could be useful other than this macro are the titles of the Community links on tiddlywiki.com (from which I took inspiration on styling the whole translink macro). I think that simply styling this particular `h1` element without defining a new class would be just as reasonable, unless it is recommended to style by assigning classes only.

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -3234,4 +3237,16 @@ select {
 
 .tc-tiny-v-gap-bottom {
 	margin-bottom: 3px;
-}
\ No newline at end of file
+}
+
+.tc-big-v-gap {
+	margin-top: 1em;
```
Jermolene, line 3243: I'm not sure if it's an artefact of github.com, but it looks like there's some uneven indenting here, and also on line 3250.

core/wiki/macros/translink.tid
```diff
@@ -3,13 +3,13 @@ tags: $:/tags/Macro
 
 \define translink(title,mode:"block")
 \whitespace trim
-<div style="border:1px solid #ccc; padding: 0.5em; background: black; foreground; white;">
+<div class="tc-pre-background">
```
Jermolene, line 6: Rather than reusing generic classes, I wonder if we might be better off introducing a new class for the translink macro "tc-translink", and then targetting the child DIV and the h1 with CSS selectors. The motivation is that it would make it easier for end users to restyle the macro output without having to alter the macro itself.

Jermolene (commented):
Thank you @mateuszwilczek much appreciated.

themes/tiddlywiki/vanilla/base.tid
mateuszwilczek: This was my fault, some lines were spaced using four spaces instead of a tab, it's fixed now.

mateuszwilczek:
@Jermolene I have improved the use of CSS classes as you suggested.

Initially I have intended this PR to fix just the visual aspect of the macro. However, I'm wondering more and more about how it's working regarding the block/inline mode.

Currently setting the mode parameter of the macro is simply passed to the transclude widget, so if inline is passed instead of the default block, the text of the transcluded widget is be inline, but the whole macro is a block element anyway. I see only a limited use for this.

I think it could be equally, if not more, useful, to make the whole macro inline, so that a short tiddler could be translinked as an inline element, see screenshot:

translink-macro-inline

This raises some questions:

- Is this functionality universal enough to be included in the core?
- If so, how should it be introduced:
  - Add a new possible value of the mode parameter of the macro, e.g. `macro-inline`, this is backwards-compatible, but raises complexity.
  - Replace the current `inline` functionality with the new one. This would introduce an incompatibility (as far as visuals are concerned). This might be not of such a concern, and maybe thought of rather as a "fix". This macro is probably not in wide spread use in the inline mode anyway, the existence of the second parameter was not mentioned anywhere in the documentation, the only way to know it would be to read macro definition.

Jermolene:
Thank you @mateuszwilczek – I think your logic is sound, and it's better to tweak the behaviour of the `inline` functionality. Will you do that in this PR?

Jermolene:
Thanks for your patience @mateuszwilczek, let me know when you're ready.

mateuszwilczek:
@Jermolene The new tweaked inline mode is ready now. The macro consists simply of two list widgets, one for each mode. There are some repeating parts, namely the links to title and text transclusions, but I don't know if it makes sense to modularise them in helper macros.

The parentheses around the transcluded text in inline mode are introduced using CSS pseudo-elements, so they can be modified using stylesheet only. I think "Title (Text)" is the most universal solution, "Title: Text" or "Title - Text" could get confusing easily if there is similar puctuation in the text.

Jermolene:
Looks good thanks @mateuszwilczek – please remove the "draft" status if this is ready to merge.

Much appreciated – the translink macro was a bit of a proof of concept, and it's great to see it getting some love.

mateuszwilczek:
I think it's ready for the final review now.

My only concern is that the inline mode doesn't always look ideal if it spans multiple lines. This is why I adjusted the vertical padding, so that the borders from neighboring lines do not overlap. This is dependent on the font, zoom level, and possibly screen resolution, so getting it perfect in all situations is not necessarily worth it. It still looks way better than it used to, and if the transcluded text spans multiple lines it is probably better to use the macro in block mode anyway.

Jermolene:
Thanks @mateuszwilczek – that all makes sense. Overall these are worthwhile improvements and it will be interesting to see how users take to them. It may be worth you making a post to talk.tiddlywiki.org to elicit some feedback. I suspect many users have never investigated the possibilities.

PR #7426, opened 2023-05-05
Add documentation on translink macro

mateuszwilczek:


Jermolene:
Thanks @mateuszwilczek. It looks like this is independent of #7425 so I'm going ahead and merging.

PR #7432, opened 2023-05-06
Correct the caption for Procedure Call tiddler

kookma:
**Documentation improvement**

This PR corrects the wrong caption for Procedure Call.tid


Jermolene:
Thank you @kookma

PR #7433, opened 2023-05-06
thisTiddler description and examples improved

kookma:
This PR address the issue in https://github.com/Jermolene/TiddlyWiki5/issues/7431

* It expands the description for `thisTiddler` 
* It adds another example to show how `thisTiddler` works for transclusion
* It moves `thisTiddler` and example doc tiddlers to proper folder

PR #7434, opened 2023-05-07
Update thisTiddler Variable (Examples) for using unique parameter n in .example macro

kookma:
This PR address issue in https://github.com/Jermolene/TiddlyWiki5/pull/7433#discussion_r1186744704


kookma:
Hi @Jermolene 
This PR is a small correction to call `.example`. I think  this can be merged.

Issue #7443, opened 2023-05-10
[BUG] Doc examples should use best-practice for: button- and action-widgets


pmario:
Everytime someone at eg: [Talk posts examples](https://talk.tiddlywiki.org/t/macro-called-from-buttonwidget-doesnt-work-as-expected/6956) using the button-widget and actions they do it wrong. ... Because our documentation at the moment does provide examples, that do it wrong. 

IMO this should be changed. It will save us a lot of maintenance. eg: [ActionConfirmWidget](https://tiddlywiki.com/#ActionConfirmWidget) .. Even if the examples are "simple" they should use best-practice. 

**Status quo**

```
<$button>
<$action-confirm $message="Do you wish to delete the caption and tags?">
<$action-deletefield caption tags/>
Delete "caption" and "tags"
</$action-confirm>
</$button>
```

**Proposed**

```
\define testActions()
<$action-confirm $message="Do you wish to delete the caption and tags?">
	<$action-sendmessage $message="tm-notify" $param="SampleNotification"/>
</$action-confirm>
\end

<$button actions=<<testActions>> >Show a Notification</$button>
```

Jermolene:
Thanks @pmario this is a tricky one. In many situations it makes sense to use a macro for the action string instead of embedded action widgets. But it does add to the complexity of the demos which might make them harder to understand.

One new development in relation to examples is that as part of the Geospatial plugin work I have been developing a new `<$testcase>` widget that dramatically improves the presentation of complex examples. The testcase widget spins up a separate wiki object into which it inserts the required tiddlers. Then it renders a standard template against the wiki that shows the individual tiddlers in tabs on the left, and the output on the right (inspired by the way that tools like Codesandbox do this kind of thing).

It makes it much easier to understand complex examples that involve multiple tiddlers, and keeps the results of interacting with the demos out of the main wiki.

It's also worth noting that the tiddlers can be directly edited on the left (but the changes will be overwritten at the next refresh).

Have a look at the examples under the "docs" tab [in the demo](https://tiddlywiki5-git-geospatial-plugin-jermolene.vercel.app/plugins/tiddlywiki/geospatial/index.html#%24%3A%2Fplugins%2Ftiddlywiki%2Fgeospatial):

![IMG_2118](https://github.com/Jermolene/TiddlyWiki5/assets/174761/676f1a8f-83e2-4bee-b9a6-485eb04d4d66)

[...]

PR #7444, opened 2023-05-10
docs ActionConfirmWidget, ActionCreateTiddlerWidget best practice examples

pmario:
docs ActionConfirmWidget, ActionCreateTiddlerWidget best practice examples

As discussed at: **[BUG] Doc examples should use best-practice for: button- and action-widgets** #7443

This PR contains changes for 2 widgets ActionConfirmWidget and ActionCreateTiddlerWidget  to make the example code easier to maintain. 

* Every example works if the buttons are clicked in the "renders as" section and
* if copy & pasted into a new tiddler, the example text still makes sense and works.


editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
```diff
@@ -20,8 +20,8 @@ The ''action-createtiddler'' widget is invisible.
 
 |!Attribute |!Description |
 |$basetitle |The initial title that will be attempted. If a tiddler with that title already exists, then a numerical counter is added to the title and incremented until it is unique|
-|$savetitle |<<.deprecated-since "5.1.20" "ActionCreateTiddlerWidget Example 5">> A text reference identifying a field or index into which the title of the newly created tiddler will be stored after it is created |
```
Jermolene, line 23: Why the change in the deprecation cutoff version?

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
```diff
@@ -32,42 +32,131 @@ The ''action-createtiddler'' widget is invisible.
 <<.from-version "5.2.0">> The content of the <<.wid "action-createtiddler">> widget is executed after the new tiddler has been created. The title of the newly created tiddler is stored in the variable <<.var "createTiddler-title">>.
 
 |!Variables |!Description |
-|`createTiddler-title` |The tittle of the tiddler that has been created. See [[ActionCreateTiddlerWidget Example 5]] |
+|`createTiddler-title` |The tittle of the tiddler that has been created. See the 5th example |
```
Jermolene, line 35: Why replace the link with the clumsy textual reference?

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
```diff
@@ -32,42 +32,131 @@ The ''action-createtiddler'' widget is invisible.
...
 |`createTiddler-draftTitle` |This variable only exists to have feature parity with the deprecated parameters. It contains the title of a "draft tiddler" |
 
 ! Examples
 
 <<<
-<$macrocall $name='wikitext-example-without-html' src={{ActionCreateTiddlerWidget Example}}/>
+1) This example will create a ''new tiddler'' every time the button is clicked.
```
Jermolene, line 41: Using hardcoded list numbering is not a practice I would want to recommend. If the worst comes to the worst, it would be better to use HTML syntax for `<ol>/<li>`.

Jermolene (commented):
Thanks @pmario. I'm not sure that some of these changes are improvements. Why have you got rid of the separate example tiddlers? It seems more useful to be able to link to them than asking the user to count their way to them.

Jermolene (commented):
Thanks @pmario. I'm not sure that some of these changes are improvements. Why have you got rid of the separate example tiddlers? It seems more useful to be able to link to them than asking the user to count their way to them.

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
pmario: See: https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.20.html#ActionCreateTiddlerWidget `$savedrafttitle` was new in 5.1.20
See: https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.23.html#ActionCreateTiddlerWidget where both settings where deprecated. 

It probably was a "copy paste" error using the version number from `$savedrafttitle` as deprecation version, which is wrong.

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
pmario: Usually action-widget doc tiddlers contain their examples in the doc tiddler. ... That's why I did remove them for consistency reasons.

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
pmario: With examples we do have a lot of hardcoded numbering. eg: https://github.com/Jermolene/TiddlyWiki5/pull/7426/files#diff-ae90cbf3713502ed3ebecc9f345a9e5c95dbb06986e618697daa323025611ead from some docs recently merged.

pmario:
>... Why have you got rid of the separate example tiddlers?

As I wrote ... For consistency with all other widget docs.

> It seems more useful to be able to link to them than asking the user to count their way to them.

I did implement the example tiddlers some time ago fro the wrong reason. I thought it would not be possible to have `wikitext-example-without-html` with `\define macro()` in it. ... I was wrong there. 

-------

I also don't like the search results showing the examples. It looks messy.

image

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
pmario: Also the numbering does not work well, if the order should be changed. If we do have 1 tiddler per example they should have better names -- other than 1, 2, 3

editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget.tid
pmario: updated the link to v5.2.0  https://tiddlywiki.com/archive/full/TiddlyWiki-5.2.0.html#ActionCreateTiddlerWidget

PR #7445, opened 2023-05-10
test-with-mobile-chrome

pmario:
@saqimtiaz ... I'm just curious. This version did a drag & drop action with my chrome on Android .. But I actually do not know, what you did. 

Can you test the Vercel preview, if the problem is still there?

Can you describe how you did drag & drop with a mobile chrome browser for me to recreate the problem.

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -80,7 +80,7 @@ exports.makeDraggable = function(options) {
 				if(dataTransfer.setDragImage) {
 					if(dragImageType === "pill") {
 						dataTransfer.setDragImage(dragImage.firstChild,-16,-16);
-					} else if (dragImageType === "blank") {
+					} else if(dragImageType === "blank") {
 						dragImage.removeChild(dragImage.firstChild);
 						dataTransfer.setDragImage(dragImage,0,0);
 					} else {
@@ -107,8 +107,9 @@ exports.makeDraggable = function(options) {
 					dataTransfer.setData("text/plain",titleString);
 					dataTransfer.setData("text/x-moz-url","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
 				} else {
-					dataTransfer.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
+					// dataTransfer.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
 				}
+				dataTransfer.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(jsonData));
 				dataTransfer.setData("Text",titleString);
 				event.stopPropagation();
 			}


saqimtiaz:
I am unable to debug or test at the moment but as far I remember dragging and dropping was broken when using the draggable and droppable widgets when a URL type was present in the dataTransfer object with a `data:` prefix.

Streams is one of the places the problem surfaced, but I recall it was reproducible using simpler constructs with the droppable and draggable widgets. I have a vague memory that the entire `data:` prefixed string was used as the tiddler title on drop.

saqimtiaz:
The broken drag and drop behaviour in Android had to do with dragging within the wiki, so the draggable and droppable widgets. It is just not possible to test with tiddler links since Android triggers the contextmenu instead of a drag event, so you need a draggable widget creating an element other than a link to drag.

saqimtiaz:
Hi @pmario, the key with the drag and drop issues on Android Chrome is that they are related to dropping on the droppable widget, not the dropzone widget.

Here is a minimal test case:
```

\define droppable-actions()
<$action-setfield $tiddler="$:/temp/droptest" text=<<actionTiddler>>/>
\end

<$draggable tag="div" tiddler="Hello There">
<p style="height:50px;width:400px;border: 1px solid red; line-height:50px;">
Drag me
</p>
</$draggable>

<$droppable tag="div" actions=<<droppable-actions>> >
<div style="height:100px;width:200px;border: 1px solid orange;">

drop here

The dropped tiddler is: {{$:/temp/droptest}}

</div>
</$droppable>
```

On master this works correctly, however on this vercel preview that reverts #6789, the actionTiddler variable is incorrect and rather than the tiddler title you get the entire data URI from the data transfer object. See screenshots below:

On master:
Screenshot_20230605_082739_com android chrome_edit_1754192238270347

On this vercel build:
Screenshot_20230605_082705_com android chrome_edit_1754182557778161

saqimtiaz:
Debugging on Android Chrome shows that when we assign the `URL` data type prefixed with `data:,` the resultant dataTransfer object received in the droppable widget only has the `text/plain` type but its value is the data URI we assigned to the `URL` type.

It is almost as if the `data:` prefixed value is rejected as being valid for the `URL` type and assigned as `text/plain` instead. I am going to try re-ordering the assignment of the data types to see if that changes the outcome.

pmario:
@saqimtiaz ... I did discover a very strange behaviour with Chrome on Android. 

I did use your testcase with the "reverted" code. 

To reproduce: 

- Start dragging the "Drag me" element
- move "Drag me" down 
- hover outside of "drop here" 
- release -> **sometimes works** as expected.

That's super strange. If I try it many times it switches between problem and between OK.

pmario:
An other test, where it works as expected consistently. 

- Tiddler with the testcase code is the only tiddler open
- phone in portrait mode
- select "drag me" element
- drag down -> phone native "reload page" animation starts
- drop element -> It consistently works ... strange

PR #7447, opened 2023-05-11
Update thisTiddler Variable.tid

pmario:
Changing the caption of thisTiddler.tid to `caption: thisTiddler`


Jermolene:
Thanks @pmario

PR #7448, opened 2023-05-11
Checkbox widget should not affect date fields

### LINKED ISSUE #6871: [BUG] Internal JavaScript Error : TypeError (..) when setting the checkbox widget attribute listField to a "default" field (list, tags,..)

rmunn:
If a date field like `created` or `modified` is passed as the listField attribute of the checkbox widget, it will be left unchanged and actions on the checkbox widget will *not* fire.

Includes unit tests to ensure that the "created" and "modified" fields will not be touched by checkboxes and will continue to be Date objects.

Fixes #6871.

### DIFF
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -112,7 +112,7 @@ CheckboxWidget.prototype.getValue = function() {
 			var list;
 			if(this.checkboxListField) {
 				if($tw.utils.hop(tiddler.fields,this.checkboxListField)) {
-					list = tiddler.getFieldList(this.checkboxListField);
+					list = tiddler.getFieldList(this.checkboxListField) || [];
 				} else {
 					list = $tw.utils.parseStringArray(this.checkboxDefault || "") || [];
 				}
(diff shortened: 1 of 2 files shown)


editions/test/tiddlers/tests/test-checkbox-widget.js
```diff
@@ -119,6 +119,25 @@ Tests the checkbox widget thoroughly.
             return newData;
         });
 
+        const listModeTestsForDateFields = [
```
pmario, line 122: We do not use `const` yet so go with `var`. ... We still use ES5 atm.

rmunn: I have `const` throughout that file. I'll switch it all to `var` in a new commit, so this PR will end up touching a lot more lines than just to fix the bug. However, I'll keep that const->var switch in its own commit, so it's easy to back off if the decision ends up to leave it as-is.

pmario: oh... It's the tests. I didn't see that. .. May be it's OK to use if for tests, since that does not make a difference for the users.

rmunn: Done in commit eb3e31dc5.

rmunn: Note that there are several plugins using `const`: the highlight.js syntax highlighting plugin has `const` throughout the library file that it pulls in, and the Twitter Archivist plugin also has `const` sprinkled throughout the code. So if we want to really be strict about ES5 targeting, we'd need to change those as well. https://caniuse.com/?search=const suggests that it's only a problem for people using IE 10 or below, so hopefully nobody at all. Still, I changed it as requested.

Jermolene:
Thanks @rmunn much appreciated.

core/modules/widgets/checkbox.js
```diff
@@ -208,16 +208,20 @@ CheckboxWidget.prototype.handleChangeEvent = function(event) {
...
-			fieldContents = tiddler ? tiddler.fields[this.checkboxListField] : undefined;
+			fieldContents = (tiddler ? tiddler.fields[this.checkboxListField] : undefined) || [];
 		} else {
 			fieldContents = this.wiki.extractTiddlerDataItem(this.checkboxTitle,this.checkboxListIndex);
 		}
 		if($tw.utils.isArray(fieldContents)) {
 			// Make a copy so we can modify it without changing original that's refrenced elsewhere
 			listContents = fieldContents.slice(0);
-		} else {
-			listContents = $tw.utils.parseStringArray(fieldContents) || [];
```
pmario, line 219: @rmunn @Jermolene This change causes this problem: https://talk.tiddlywiki.org/t/checkboxwidget-listindex-mode-no-longer-working-in-5-3-0/7460

PR #7463, opened 2023-05-18
Update Procedure Definitions.tid

kookma:
Change macro to procedure!


Jermolene:
Great thank you @kookma

PR #7467, opened 2023-05-19
Procedure definition example should not say macro

rmunn:
The word "macro" appears in the text of the procedure definition example, but it should be "procedure" instead.

This is a minor fix that https://github.com/Jermolene/TiddlyWiki5/pull/7463 missed.


Jermolene:
Thanks @rmunn

Issue #7460, opened 2023-05-16
[BUG] widget destroy() method fails testing.


buggyj:
### Describe the bug

I have written some test scripts to test my destroy plugin (for v5.2.7) 

These tests aim to show:

- when a widget is to be removed its destroy() function is called

- when a widget's destroy() function is called it is able to access its dom nodes

I have tried these tests with prerelease of 5.3.0 and 2 of the first 3 tests fail (and so the later tests cannot be performed).

The test wiki is here, (includes descriptions of the tests and test widgets):

https://destroytests530prerelease.tiddlyhost.com/

for reference the tests with my plugin (all tests passing) are here:

https://destroytest.tiddlyhost.com/

### TiddlyWiki Configuration

- Version [e.g. v5.3.0prerelease]

Jermolene:
Hi @buggyj I am not familiar with your "destroy" plugin:

https://github.com/buggyj/destroy/blob/main/common/destroyfunction.js

It looks like a superior implementation of the "destroy" method that was merged in #6699, is that right? Would it be feasible for the core to adopt your implementation?

buggyj:
yes of course. I think the code could be simplified a bit further - I will put in a pull request.

Jermolene:
Thank you @buggyj – I've reverted #6699 for the moment

linonetwo:
I read the code, functionality is basically the same  （at least for tw-react plugin), we can use this simpler one for maintainability.

One thing different is I omit the dom removal of child widget https://github.com/Jermolene/TiddlyWiki5/pull/6699/files#diff-923154c1ac214c303ed3a97ac47050febdbea3c32b9b246c53e015b32b2641cdR631

Because when you remove parent dom node, you don't need to remove children. This can save some (not calculated yet) expensive dom operations.

pmario:
@buggyj I think this one can be closed now
@linonetwo There should be a new issue handling the `.destroy()` communication ... or open the old one.

pmario:
IMO related to: Adds a widget destroy method #9097

PR #7468, opened 2023-05-19
implemented destroy() function

buggyj:
I think this should work for https://github.com/Jermolene/TiddlyWiki5/issues/7460

I wonder if it would be better left as a plugin given that the core do not have a need for destroy()....

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -751,22 +751,43 @@ Widget.prototype.findFirstDomNode = function() {
 };
 
 /*
-Remove any DOM nodes created by this widget or its children
+Entry into destroy procedure
+*/
+Widget.prototype.destroyChildren = function() {
+	$tw.utils.each(this.children,function(childWidget) {
+		childWidget.destroy();
+	});
+};
+/*
+Legacy entry into destroy procedure
 */
 Widget.prototype.removeChildDomNodes = function() {
-	// If this widget has directly created DOM nodes, delete them and exit. This assumes that any child widgets are contained within the created DOM nodes, which would normally be the case
+	this.destroy();
+};
+/*
+Default destroy
+*/
+Widget.prototype.destroy = function() {
+	// call children to remove their resources
+	this.destroyChildren();
+	// remove our resources
+	this.children = [];
+	this.removeLocalDomNodes();	
+};
+
+/*
+Remove any DOM nodes created by this widget 
+*/
+Widget.prototype.removeLocalDomNodes = function() {
+	// If this widget has directly created DOM nodes, delete them and exit.
 	if(this.domNodes.length > 0) {
 		$tw.utils.each(this.domNodes,function(domNode) {
 			domNode.parentNode.removeChild(domNode);
 		});
 		this.domNodes = [];
-	} else {
-		// Otherwise, ask the child widgets to delete their DOM nodes
-		$tw.utils.each(this.children,function(childWidget) {
-			childWidget.removeChildDomNodes();
-		});
 	}
 };
+
 
 /*
 Invoke the action widgets that are descendents of the current widget.


Jermolene:
Many thanks @buggyj. It would be very helpful to include a documentation update in `editions/dev` that tells widget developers what has changed and they need to do as a result. Perhaps a new tiddler called something like "New widget destroy method introduced in v5.3.0", which should be linked from "HelloThere" and any existing widget documentation tiddlers. If there are any backwards compatibility implications then we need to cover those in the release note as well.

linonetwo:
@buggyj why not add those tests to this repo?

Seems your version don't need to change existing core widgets?

buggyj:
@linonetwo I do not foresee any problem with backwards compatibility.

linonetwo:
That's good.



> @buggyj why not add those tests to this repo?
> 
> Seems your version don't need to change existing core widgets?

These are two separate questions. So can you add those tests you created to this repo?

linonetwo:
> tells widget developers what has changed and they need to do as a result.

So this is not needed, as long as this PR has backwards compatibility.

And I can write the usage doc.

linonetwo:
@buggyj I haven't test it, will this method be called when switch from edit mode to view mode? https://github.com/Jermolene/TiddlyWiki5/discussions/5945#discussioncomment-2775060

saqimtiaz:
Docs are in #7508

Jermolene:
Thank you @buggyj

saqimtiaz:
@Jermolene it would be good to check whether this impacts performance in a measurable way. The new `destroy()` method also means that when a widget is destroyed, we recursively remove the DOM nodes created by each child rather than just removing the parent DOM node and thereby its children with it.

linonetwo:
Yes, my previous PR considers this, but it made it complex.

But I think just open performance tab of developer tool, in [HugeWiki](https://github.com/tiddly-gittly/HugeWikiExample) can reflect this.

saqimtiaz:
@linonetwo if you have the opportunity to do a quick performance test of the master branch vs the current TiddlyWiki release that would be very welcome

linonetwo:
I have updated hugewiki demo site with latest prerelease. But I have to do the check later.

Jermolene:
> @Jermolene it would be good to check whether this impacts performance in a measurable way. The new `destroy()` method also means that when a widget is destroyed, we recursively remove the DOM nodes created by each child rather than just removing the parent DOM node and thereby its children with it.

I tried a very simple experiment which suggests that there is indeed the expected performance impact. I just ran the following code in the developer console of both tiddlywiki.com and tiddlywiki.com/prerelease. Before running the test on the prerelease I manually closed the tiddler "TiddlyWiki Pre-release" so that the same tiddlers were displayed in both cases.

```
console.time();
$tw.rootWidget.removeChildDomNodes()
console.timeEnd();
```

The results were:

* https://tiddlywiki.com/ - 1.14ms
* https://tiddlywiki.com/prerelease - 3.65ms

It's hard to see a way around the performance hit. It's possible that those figures are misleading because manually destroying DOM nodes might improve the performance of garbage collection.

pmario:
>It's hard to see a way around the performance hit. It's possible that those figures are misleading because manually destroying DOM nodes might improve the performance of garbage collection.

**TW v5.2.7**

I did some test and the garbage collection mainly kicks in when we create a lot of elements. eg: Switching the sidebar from Open to More -> Tags will causes 16 minor GCs from 90µs up to 1.7ms and 6 GCs with 2ms each. 

Switching from More to Open tab causes _no_ GS at all. ... 
Switching needs 22ms

Using FF on Windows.

**TW v5.3.0**

Switch from Open to More -> Tags ... 18 minor GS from 90µs up to 2ms ... and 7 GCs 2ms each. 

Switching from More to Open causes no GS
Switching needs 32.6 ms

all .destroy functions seem to need about 6.6ms

saqimtiaz:
I have done some testing today working on the assumption that a `destroy` method is desirable. As such I have compared the current master branch to a tweaked implementation of `destroy()` that does not recursively remove child DOM nodes if a parent widget will be removing its DOM node (and thus all descendant nodes).

I tested the refresh time when switching sidebar tabs, and also ran the test outlined by @Jermolene above.

Executing $tw.rootWidget.destroy the results were (average of 10 executions):
* master branch - 3.84ms
* with a selective destroy method - 2.24ms

However, when switching from the `More/All` sidebar tab to `More/Tags` and back, the refresh times were comparable between both. So while there may not be a real world impact in performance, in the interest of being cautious I suggest we revert this change in v5.3.1 and revisit this for a future release after getting a more thorough performance test in place.

Jermolene:
> I have done some testing today working on the assumption that a `destroy` method is desirable. As such I have compared the current master branch to a tweaked implementation of `destroy()` that does not recursively remove child DOM nodes if a parent widget will be removing its DOM node (and thus all descendant nodes).

Does it still call the destroy method for all child widgets? The primary need that has driven this change is for widgets to be able to implement a destroy method that is guaranteed to be called when the widget is deleted from the widget tree, and so I wouldn't want to undermine that use case.

> However, when switching from the `More/All` sidebar tab to `More/Tags` and back, the refresh times were comparable between both. So while there may not be a real world impact in performance, in the interest of being cautious I suggest we revert this change in v5.3.1 and revisit this for a future release after getting a more thorough performance test in place.

I am happy to err on the side of caution here.

saqimtiaz:
> Does it still call the destroy method for all child widgets? The primary need that has driven this change is for widgets to be able to implement a destroy method that is guaranteed to be called when the widget is deleted from the widget tree, and so I wouldn't want to undermine that use case.

Yes it does, with a flag indicating whether an ancestor node is going to be removed. I designed the performance comparison in this manner to take into account the tree traversal time for the destroy method which needs to happen regardless of whether child DOM nodes are removed one at a time or not.

I second erring on the side of caution and reverting this PR, with the aim to introduce some performance tests and subsequently in a later release introduce a refined version of the destroy method that only removes child nodes when needed.

Jermolene:
Thanks @saqimtiaz I've reverted it in 160cc0e9a90535b7210a2ef021a5389c38ae9917

buggyj:
When I think of a destroy method I image that it will remove all the resources that were created. We could have a 'final' widget method that would not remove all the resources that the widget created, eg dom nodes and child widgets, these could be removed from within the 'refreshshelf' method (domnodes) and removal of childwidget from within the 'makechildwidgets' as happened before version 5.3.0 and leave the 'final' method to cancel timers etc. Unless major changes are made to the widget structure, it would still be necessary to visit all child widgets to call final() whenever it appears - so there will always be some performance price to pay.

linonetwo:
@buggyj 

> remove all the resources that were created

we can just do it in `destroy` method.

But I think it should be

> remove all the resources that were created, excluding DOM nodes

because DOM nodes deletion can be better managed by browser. So we don't need `final` and `destroy`, this will make it confusing.

buggyj:
@linonetwo @Jermolene @saqimtiaz The reason for reverting is fuzzy. It is clear that some cpu time is required to perform the destroy, and that is has caused enough concern to have it removed. However there is no criteria that describes an acceptable performance.  So until we have an acceptance criteria it is not possible to fix this issue.

linonetwo:
@buggyj Actually, informing child widgets without removing child widgets' dom node is possible. My first implementation considers this https://github.com/Jermolene/TiddlyWiki5/pull/6699 (although it breaks other things somehow?)

For this PR, maybe just a small modification will be sufficient.

saqimtiaz:
> informing child widgets without removing child widgets' dom node is possible.

I wrote such an implementation that was backwards compatible and found it measurably faster than the one we merged. However, I didn't have the opportunity for proper testing before the previous release. I am unfortunately away from home for the next couple of weeks but can push that implementation as a draft when I return if this issue is still pending.

linonetwo:
> push that implementation as a draft when I return if this issue is still pending.

That's OK, I might be able to run test on them.

core/modules/widgets/widget.js
```diff
@@ -751,23 +751,44 @@ Widget.prototype.findFirstDomNode = function() {
...
+};
+
+/*
+Remove any DOM nodes created by this widget 
+*/
+Widget.prototype.removeLocalDomNodes = function() {
+	// If this widget has directly created DOM nodes, delete them and exit.
 	if(this.domNodes.length > 0) {
 		$tw.utils.each(this.domNodes,function(domNode) {
 			domNode.parentNode.removeChild(domNode);
```
linonetwo, line 785: Should check here `domNode.parentNode?.removeChild?.(domNode);`

Otherwise `Uncaught TypeError: Cannot read properties of null (reading 'removeChild')`

linonetwo: A userside workaround is add `this.domNodes = [];` in `destroy`, so this line won't get executed.

linonetwo:
I'm using this in tw-react v0.5.1 and so its dependent slate-write v0.5.0

linonetwo:
@saqimtiaz If you can't opensource your solution, I may have to modify base on this PR.

saqimtiaz:
I will find time to share it this coming week, thank you for the reminder.

linonetwo:
@saqimtiaz how is it?

linonetwo:
@saqimtiaz Hi, how are you?

dongrentianyu:
@saqimtiaz how is it?

linonetwo:
I'm using https://github.com/wessberg/connection-observer as described in https://github.com/Jermolene/TiddlyWiki5/discussions/5945 in several plugins. Each add 32k to the file.

PR #7471, opened 2023-05-21
feat: decode url-encoded user name in authenticate request header to allow CJK username in header

linonetwo:
Chinese Japanese Koren character is not allowed in request header

So they have to be encoded before added to the header. So on server side, they need to be decode to restore.

The usecase is https://github.com/Jermolene/TiddlyWiki5/discussions/7469

### DIFF
--- a/core/language/en-GB/Help/listen.tid
+++ b/core/language/en-GB/Help/listen.tid
@@ -18,7 +18,7 @@ All parameters are optional with safe defaults, and can be specified in any orde
 * ''anon-username'' - the username for signing edits for anonymous users
 * ''username'' - optional username for basic authentication
 * ''password'' - optional password for basic authentication
-* ''authenticated-user-header'' - optional name of header to be used for trusted authentication
+* ''authenticated-user-header'' - optional name of header field to be used for trusted authentication.
 * ''readers'' - comma-separated list of principals allowed to read from this wiki
 * ''writers'' - comma-separated list of principals allowed to write to this wiki
 * ''csrf-disable'' - set to "yes" to disable CSRF checks (defaults to "no")
--- a/core/language/en-GB/Help/listen.tid
+++ b/core/language/en-GB/Help/listen.tid
@@ -18,7 +18,7 @@ All parameters are optional with safe defaults, and can be specified in any orde
 * ''anon-username'' - the username for signing edits for anonymous users
 * ''username'' - optional username for basic authentication
 * ''password'' - optional password for basic authentication
-* ''authenticated-user-header'' - optional name of header field to be used for trusted authentication.
+* ''authenticated-user-header'' - optional name of request header to be used for trusted authentication.
 * ''readers'' - comma-separated list of principals allowed to read from this wiki
 * ''writers'' - comma-separated list of principals allowed to write to this wiki
 * ''csrf-disable'' - set to "yes" to disable CSRF checks (defaults to "no")
--- a/editions/tw5.com/tiddlers/webserver/WebServer Header Authentication.tid
+++ b/editions/tw5.com/tiddlers/webserver/WebServer Header Authentication.tid
@@ -4,6 +4,10 @@ tags: [[WebServer Authentication]]
 title: WebServer Header Authentication
 type: text/vnd.tiddlywiki
 
-Header authentication is a web integration technique enabling external entities to securely pass details of the authenticated user to an application. It is commonly used for "single sign on" in corporate environments.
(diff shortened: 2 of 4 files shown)


Jermolene:
Thanks @linonetwo I notice that there are a few other places in the core code where we access `request.headers` without decoding the values. I think it would be worth fixing all of them in one PR.

We have a utility method `$tw.utils.decodeURIComponentSafe()` that should be used for the URI decoding.

linonetwo:
Ok, I checked other usages

- if-none-match: only contains ASCII, so no needed to decodeURI
- accept-encoding: only standard enums, no needed too\
- x-requested-with: will only be "Tiddlywiki"
- request.headers.authorization: this should have needed, but you already do `token = header.split(/\s+/).pop()` then `$tw.utils.base64Decode(token),` on it.

Jermolene:
Hi @linonetwo I've just been doing some more checking. As far as I can tell, it seems that HTTP headers are not URI encoded before transit, and so do not need URI decoding before use. Because header values are restricted to ASCII characters, URI encoding is sometimes used with particular headers to allow extended characters to be included.

So I think what you're actually proposing here is not a bug fix, but rather it is a new feature: supporting URI encoding in the authenticated username header so that usernames with extended characters can be used. It's not strictly backwards compatible.

On that basis, I think it would be worth adding a note to [the docs](https://tiddlywiki.com/#WebServer%20Parameter%3A%20authenticated-user-header) to explicitly note that since v5.3.0 the header value is URI decoded before use.

linonetwo:
Updated the doc, I was really confused by this doc, so I add more detail to them based on what I learned yesterday on this topic.

core/language/en-GB/Help/listen.tid
```diff
@@ -18,7 +18,7 @@ All parameters are optional with safe defaults, and can be specified in any orde
 * ''anon-username'' - the username for signing edits for anonymous users
 * ''username'' - optional username for basic authentication
 * ''password'' - optional password for basic authentication
-* ''authenticated-user-header'' - optional name of header to be used for trusted authentication
+* ''authenticated-user-header'' - optional name of header field to be used for trusted authentication.
```
Jermolene, line 21: The docs do not use the term "field" for headers to avoid confusion with tiddler fields. We could say "request header" perhaps?

editions/tw5.com/tiddlers/webserver/WebServer Header Authentication.tid
```diff
@@ -4,6 +4,10 @@ tags: [[WebServer Authentication]]
...
 type: text/vnd.tiddlywiki
 
-Header authentication is a web integration technique enabling external entities to securely pass details of the authenticated user to an application. It is commonly used for "single sign on" in corporate environments.
+Header authentication is a web integration technique enabling external entities to securely pass details of the authenticated user to an application.
 
 Header authentication is activated if is configured via the [[authenticated-user-header|WebServer Parameter: authenticated-user-header]]
+
+!! Usage in SSO
+
+It is commonly used for "single sign on" in corporate environments. When doing header authentication, the user is not prompted for a username and password on TiddlyWiki. Instead, user is required to login at a SSO proxy server. When the user authenticate themselves to a SSO proxy server, proxy server redirect user request to TiddlyWiki server with this additional header field. Then tiddlywiki server is able to use the value of this header field to identify the user.
```
Jermolene, line 13: Minor corrections:

Header authentication is commonly used for "single sign on" in corporate environments. When doing header authentication, the user is not prompted for a username and password on TiddlyWiki. Instead, the user is required to login at a SSO proxy server. When the user authenticates themselves to the SSO proxy server, the proxy server redirects the user request to the TiddlyWiki server with this additional header field containing the username. Then TiddlyWiki server is able to use the value of this header field to identify the user.

Jermolene (commented):
Thanks @linonetwo

Jermolene (commented):
Thanks @linonetwo

linonetwo:
Updated, this concept involves 3 level: authenticated-user-header → request header field → user name , it take a while for me to understand how to utilize it.

linonetwo:
preview:

https://tiddlywiki5-git-fork-linonetwo-fix-header-username-jermolene.vercel.app/#WebServer%20Parameter%3A%20authenticated-user-header:ListenCommand%20%5B%5BWebServer%20Parameter%3A%20authenticated-user-header%5D%5D%20%5B%5BWebServer%20Header%20Authentication%5D%5D

Jermolene:
Thanks @linonetwo

PR #7473, opened 2023-05-22
Update Filter Run Prefix (Examples).tid

twMat:
This catches all filter run prefixes. This is superior to the current hardcoded text that doesn't give any clues about non-documented prefixes.

My *actual* wish is to implement the more elaborate changes outlined [here](https://talk.tiddlywiki.org/t/ranty-proposal-for-docs-on-filter-run-prefixes/5772)... but this current PR is at least an improvement meanwhile.


twMat:
@pmario - I'm fine with your PRs being used instead of mine assuming they capture what I'm after in my PR. (BTW, I put effort into imitating the existing text that I'm replacing in that tiddler, but I frankly don't think it was a nice display so I'm happy if things end up better). Does any of your PRs **dynamically aggregate** all filter prefixes, including undocumented ones? I think it is wise to do this as also dead links are superior to no mention at all because it (1) shows a prefix *exists* and (2) "provokes" authoring. Currently, it is almost impossible to find out the "existance" of the undocumented prefixes.

PR #7474, opened 2023-05-22
Use DPrint to ensure consistent code formatting

linonetwo:
As an alternative to https://github.com/Jermolene/TiddlyWiki5/pull/7394

I suggest dprint because it is very fast (rust) and secure (see dprint.json, plugins are wasm in sandbox).

And I also suggest using pnpm because it can reuse packages across projects and is fast. (optional if you prefer npm, but I don't know why can't get npm install things on my Mac today, its been a while since last using it.)

No need to run dprint/prettier alone, they can be integrated into eslint and use `eslint . --fix` instead. Or we can do this file-by-file, when modifying a file in the future.

There are lots of style to be fixed, due to `(function(){ })()` on many js file that should introduce indentation but there was no indentation.
(I'm not sure this wrapper is necessary, I never write them in my plugins and still works fine.)

I also config dprint errors as warning in eslint, so in VSCode it looks softer:

图片

Use alt+shift+G to fix eslint error on a file, as well as dprint/prettier errors:

图片

### DIFF
--- a/dprint.json
+++ b/dprint.json
@@ -2,9 +2,14 @@
   "lineWidth": 120,
   "typescript": {
     "quoteProps": "asNeeded",
+    "parameters.spaceAround": false,
+    "ifStatement.spaceAfterIfKeyword": false,
+    "forStatement.spaceAfterForKeyword": false,
+    "whileStatement.spaceAfterWhileKeyword": false,
+    "spaceSurroundingProperties": false,
     "useTabs": true,
     "indentWidth": 1,
-    "quoteStyle": "preferSingle",
+    "quoteStyle": "preferDouble",
     "binaryExpression.operatorPosition": "sameLine"
   },
   "json": {},


dprint.json
```diff
@@ -0,0 +1,23 @@
+{
+  "lineWidth": 120,
+  "typescript": {
```
linonetwo, line 3: See https://dprint.dev/plugins/typescript/config/ , this also works for JS, not only TS

Jermolene:
Thanks @linonetwo I won't be able to try this out for a few days, but it looks good.

* I'm not familiar with pnpm, perhaps it's worth opening another issue to make the case and gather the evidence. We generally use npm because it is included in the default install from nodejs.org, and doesn't need to be separately installed
* The extra `(function(){ })()` wrapper around every module is redundant, and we should remove them all

pmario:
>There are lots of style to be fixed, due to (function(){ })() on many js file that should introduce indentation but there was no indentation.
(I'm not sure this wrapper is necessary, I never write them in my plugins and still works fine.)

at https://github.com/Jermolene/TiddlyWiki5/commit/19c052d1380499ddbf75c78d7a878800a4ce8a39#commitcomment-107771231 Jeremy wrote

>> pmario wrote
>>
>> IMO If we want to use that, the redundant wrapper function has to go.
>
>I'd like to remove the redundant wrappers in any case.
>
> Jeremy wrote

So we would be able to get rid of all the `(function(){ })()` wrappers ... 

The problem is even removing `\r` from the TW file reduced the file size quite a bit. So adding more characters for indentation as there is at the moment will have a negative impact on the file size.

So @linonetwo .. Did you compare the "vanilla" TW file size before and after you did use dprint?

linonetwo:
@pmario After removing `(function(){ })()`,  there won't be much changes. It won't add a lot indentations

图片

linonetwo:
I didn't see a rule to save this space, this might add a lot of spaces

图片

pmario:
>I didn't see a rule to save this space, this might add a lot of spaces

This one is a tricky one. From my point of view, in general the extra space would improve human readability. But it has a negative impact on file size. 

-----

You did open a "feature request" at the dprint repo. right? ... We'll see what they say.

linonetwo:
dprint wasm might work in browser too, so we might be able to write a Rust formatted compile to wasm and run with drpint as a tw codemirror plugin

linonetwo:
@pmario size compare  https://github.com/linonetwo/TiddlyWiki5/pull/3

Before: 1,803,178 B (1.8MB)

After : 1,803,407 B (1.8MB)

linonetwo:
If this is merged, other developers can run npx exlint . --fix before pr, so there will no be conflict.

linonetwo:
Sorry, I just right-click the `core/` folder or tiddlywiki to show its info for size. So it does not include the verbose HTML syntax in HTML.

pmario:
OK. But still. What do I need to do to activate dprint for all files so I can create a new file afterwards.

linonetwo:
See https://github.com/linonetwo/TiddlyWiki5/pull/3#issuecomment-1590617178 , but also a pr to add it to npm script so no one need to ask about this https://github.com/Jermolene/TiddlyWiki5/pull/7539

pmario:
Very interesting. 

- The log reports 13 errors and 
- modified 68 files. 

That's almost nothing. I did expect much more problems. I'll have to have a closer look, what actually has been changed.

linonetwo:
> IMO in [#7474 (comment)](https://github.com/Jermolene/TiddlyWiki5/pull/7474#issuecomment-1590485079) there is a typo:
> 
> IMO it should be `npx eslint . --fix` ... because exlint wants to install something and then throws an error

yes, updated.

About error, I think we can merge the auto fixed version, then take spare time to fix other errors one PR by one PR.

pmario:
Did you find a configuration to ignore the "additional indentation" cause by the wrapper functions?

linonetwo:
> Did you find a configuration to ignore the "additional indentation" cause by the wrapper functions?

Not yet, maybe we can just accept the space? Or I have to add that rule.

But spaces is not so much, as far as we tested.

linonetwo:
let's use biome

pmario:
We may need to go back to this approach using ESLint and dprint, because of

- #9113 and
- Jeremy's comment about parameter spacing

linonetwo:
Sure, I'm still using eslint + dprint in my projects.

Due to https://github.com/so1ve/eslint-plugin-dprint-integration/issues/112 , now use https://github.com/antfu/eslint-plugin-format instead.

pmario:
Can you revoke a PR like this one, since it seems we can not use biome, because there is no possibility to check for ES2017. With ESLint this should be possible. 

With dprint it is possible to change the configuration, so it formats the code as Jeremy wants to have it. While I think the biome configuration is more human readable, dprint is more configurable. 

We only need to hope, that the project does not go away in the future.

pmario:
@linonetwo -- IMO we should reactivate this one if possible.

linonetwo:
OK I will try add https://github.com/antfu/eslint-plugin-format .

I notice that eslint in tw repo is not formatting space to tab, maybe this can solve it.

PR #7478, opened 2023-05-22
Fix check for globalThis

Arlen22:
Untested, but this should bypass the error in old node versions where globalThis is not available. I don't understand why this wasn't an issue before, but hopefully this fixes it.


buggyj:
when you put in a polyfil for 'globalThis' the-global-object will contain a key (called 'globalThis') that references 'the-global-object'. So `Object.keys(globalThis).length` will be at least 1

Arlen22:
Ok, I see what you're saying. Let me take a look at it.

PR #7488, opened 2023-05-24
Update Field Editor Cascade.tid

twMat:


pmario:
I think I would prefer a button, that opens the **ControlPanel -> Info -> Advanced -> Cascades -> xxxx** tabs, because the show the "real" configuration that is active at the moment.

grafik

twMat:
@pmario - the problem is that the reader of the docs needs to study how the cascades are used, in practice. The tab you refer to does not link to any such tiddlers to be studied.

But I do agree it woudl be better if the doc could *generate* a list of actual tiddlers instead of hard coded. Is there a filter to identify these shadow titles?

pmario:
>.. the problem is that the reader of the docs needs to study how the cascades are used, in practice. The tab you refer to does not link to any such tiddlers to be studied.

Ahh. OK. You are right.

>But I do agree it woudl be better if the doc could generate a list of actual tiddlers instead of hard coded. Is there a filter to identify these shadow titles?

I don't think so. ... 

@Jermolene ... I do support the additions Mat made with those PRs.

Jermolene:
Thanks @twMat

PR #7490, opened 2023-05-24
fix #7472 add option explodeplugins

### LINKED ISSUE #7472: [IDEA] When extract HTML wiki into nodejs wiki, place plugins in tiddlers/ folder in JSON format

https://github.com/tiddly-gittly/TidGi-Desktop/issues/324

Some TidGi user report that using GUI to extract HTML wiki, extracted wiki's plugin can't be managed from GUI (delete in control panel and upgrade in plugin library is not working), which is very confusing.

While installing new plugins from plugin library and delete newly installed plugin still works, which is very weird.

But later they found out why: 

When running `wikiInstance.boot.argv = ['--load', htmlWikiPath, '--savewikifolder', saveWikiFolderPath];`
Extracted plugins will be placed into `plugins/` folder and `tiddlywiki.info` file, which can't be managed from wiki GUI.

I hope there can be an option to place all things inside tiddlers/ folder, so they can be managed from GUI.

https://github.com/tiddly-gittly/TidGi-Desktop/blob/3094eb764b1ed6507be73fdfef7e8fda65c3d080/src/services/wiki/wikiWorker.ts#L201

I think the deeper reason cause this is that 

> https://talk.tiddlywiki.org/t/the-right-way-to-install-the-shiraz-plugin-on-nodejs/6961/9?u=linonetwo
> 
> I guess previously, only geek use nodejs wiki, because it requires user to access terminal command line shell, so adding plugin in a geek way is not a problem.
>
> But now with TidGi app, even normal user can use TW nodejs wiki in an Obsidian-like way, so I think this rule will need to change when target user is TidGi app user.
> 
> (If you are using terminal to start nodejs wiki…You can keep the geek way)

Or we may need to implement more logic to manage plugins in plugins folder from control panel, and logic to upgrade them.

**Additional context**

To make more "normal" people use tiddlywiki, I want nodejs wiki to be easy&self-contain as HTML wiki, while still being able to access each single file from fs (for "geek user"'s git backup or vscode edit)

Zacharia2:
hi @Jermolene:
I tried to add an option for explodeplugins. Hope this works.

https://github.com/Jermolene/TiddlyWiki5/issues/7472#issuecomment-1556789379

https://github.com/tiddly-gittly/TidGi-Desktop/issues/324

### DIFF
--- a/core/language/en-GB/Help/savewikifolder.tid
+++ b/core/language/en-GB/Help/savewikifolder.tid
@@ -4,7 +4,7 @@ description: Saves a wiki to a new wiki folder
 <<.from-version "5.1.20">> Saves the current wiki as a wiki folder, including tiddlers, plugins and configuration:
 
 ```
---savewikifolder <wikifolderpath> [<filter>]
+--savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
 ```
 
 * The target wiki folder must be empty or non-existent
@@ -12,8 +12,19 @@ description: Saves a wiki to a new wiki folder
 * Plugins from the official plugin library are replaced with references to those plugins in the `tiddlywiki.info` file
 * Custom plugins are unpacked into their own folder
 
+The following options are supported:
+
+* ''filter'': a filter expression that defines the tiddlers to include in the output.
+* ''explodePlugins'': set to "no" to save plugins to the tiddlers directory of the target wiki folder, to suppress breaking plugins into their constituent shadow tiddlers (default is "yes").
+
 A common usage is to convert a TiddlyWiki HTML file into a wiki folder:
 
 ```
 tiddlywiki --load ./mywiki.html --savewikifolder ./mywikifolder
+```
+
+Save the plugin to the tiddlers directory of the target wiki folder:
+
+```
+tiddlywiki --load ./mywiki.html --savewikifolder ./mywikifolder explodePlugins=no
 ```
--- a/core/modules/commands/savewikifolder.js
+++ b/core/modules/commands/savewikifolder.js
@@ -5,7 +5,14 @@ module-type: command
 
 Command to save the current wiki as a wiki folder
 
---savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
+--savewikifolder <wikifolderpath> [ [<name>=<value>] ]*
+
+The following options are supported:
+
+* ''filter'': a filter expression defining the tiddlers to be included in the output
+* ''explodePlugins'': set to "no" to suppress exploding plugins into their constituent shadow tiddlers (defaults to "yes")
+
+Supports backward compatibility with --savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
 
 \*/
 (function(){
@@ -40,15 +47,12 @@ Command.prototype.execute = function() {
 		tiddlerFilter,
 		explodePlugins;
 	if (regFilter.test(this.params[1])) {  
-		// --savewikifolder <wikifolderpath> [ [<name>=<value>] ]*
 		namedParames = this.commander.extractNamedParameters(this.params.slice(1));
 		tiddlerFilter = namedParames.filter || "[all[tiddlers]]";
 	} else {
-		// --savewikifolder <wikifolderpath> <filter> [ [<name>=<value>] ]*
 		namedParames = this.commander.extractNamedParameters(this.params.slice(2));
 		tiddlerFilter = this.params[1];
 	}
-	// --savewikifolder <wikifolderpath> [<filter>] 
 	tiddlerFilter = tiddlerFilter || "[all[tiddlers]]";
 	explodePlugins = namedParames.explodePlugins || "yes";
 	var wikifoldermaker = new WikiFolderMaker(this.params[0],tiddlerFilter,this.commander,explodePlugins);
@@ -59,7 +63,7 @@ function WikiFolderMaker(wikiFolderPath,wikiFilter,commander,explodePlugins) {
 	this.wikiFolderPath = wikiFolderPath;
 	this.wikiFilter = wikiFilter;
 	this.commander = commander;
-	this.explodePlugins = explodePlugins; // default: write plugins to the plugins folder
+	this.explodePlugins = explodePlugins;
 	this.wiki = commander.wiki;
 	this.savedPaths = []; // So that we can detect filename clashes
 }
--- a/core/modules/commands/savewikifolder.js
+++ b/core/modules/commands/savewikifolder.js
@@ -5,7 +5,7 @@ module-type: command
 
 Command to save the current wiki as a wiki folder
 
---savewikifolder <wikifolderpath> [<filter>]
+--savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
 
 \*/
 (function(){
@@ -35,14 +35,31 @@ Command.prototype.execute = function() {
 	if(this.params.length < 1) {
 		return "Missing wiki folder path";
 	}
-	var wikifoldermaker = new WikiFolderMaker(this.params[0],this.params[1],this.commander);
+	var regFilter = /^[a-zA-Z0-9\.\-_]+=/g,  // dynamic parameters
+		namedParames,
+		tiddlerFilter,
+		explodePlugins;
+	if (regFilter.test(this.params[1])) {  
+		// --savewikifolder <wikifolderpath> [ [<name>=<value>] ]*
+		namedParames = this.commander.extractNamedParameters(this.params.slice(1));
+		tiddlerFilter = namedParames.filter || "[all[tiddlers]]";
+	} else {
+		// --savewikifolder <wikifolderpath> <filter> [ [<name>=<value>] ]*
+		namedParames = this.commander.extractNamedParameters(this.params.slice(2));
+		tiddlerFilter = this.params[1];
+	}
+	// --savewikifolder <wikifolderpath> [<filter>] 
+	tiddlerFilter = tiddlerFilter || "[all[tiddlers]]";
+	explodePlugins = namedParames.explodePlugins || "yes";
+	var wikifoldermaker = new WikiFolderMaker(this.params[0],tiddlerFilter,this.commander,explodePlugins);
 	return wikifoldermaker.save();
 };
 
-function WikiFolderMaker(wikiFolderPath,wikiFilter,commander) {
+function WikiFolderMaker(wikiFolderPath,wikiFilter,commander,explodePlugins) {
 	this.wikiFolderPath = wikiFolderPath;
-	this.wikiFilter = wikiFilter || "[all[tiddlers]]";
+	this.wikiFilter = wikiFilter;
 	this.commander = commander;
+	this.explodePlugins = explodePlugins; // default: write plugins to the plugins folder
 	this.wiki = commander.wiki;
 	this.savedPaths = []; // So that we can detect filename clashes
 }
@@ -93,10 +110,13 @@ WikiFolderMaker.prototype.save = function() {
 						self.log("Adding built-in plugin: " + libraryDetails.name);
 						newWikiInfo[libraryDetails.type] = newWikiInfo[libraryDetails.type]  || [];
 						$tw.utils.pushTop(newWikiInfo[libraryDetails.type],libraryDetails.name);
-					} else {
+					} else if(self.explodePlugins == "yes") {
 						// A custom plugin
 						self.log("Processing custom plugin: " + title);
 						self.saveCustomPlugin(tiddler);
+					} else if(self.explodePlugins == "no") {
+						self.log("Processing custom plugin to tiddlders folder: " + title);
+						self.saveTiddler("tiddlers", tiddler);
 					}
 				} else {
 					// Ordinary tiddler
--- a/core/modules/commands/savewikifolder.js
+++ b/core/modules/commands/savewikifolder.js
@@ -45,7 +45,7 @@ Command.prototype.execute = function() {
 	var regFilter = /^[a-zA-Z0-9\.\-_]+=/g,  // dynamic parameters
 		namedParames,
 		tiddlerFilter,
-		explodePlugins;
+		options = {};
 	if (regFilter.test(this.params[1])) {  
 		namedParames = this.commander.extractNamedParameters(this.params.slice(1));
 		tiddlerFilter = namedParames.filter || "[all[tiddlers]]";
@@ -54,16 +54,16 @@ Command.prototype.execute = function() {
 		tiddlerFilter = this.params[1];
 	}
 	tiddlerFilter = tiddlerFilter || "[all[tiddlers]]";
-	explodePlugins = namedParames.explodePlugins || "yes";
-	var wikifoldermaker = new WikiFolderMaker(this.params[0],tiddlerFilter,this.commander,explodePlugins);
+	options.explodePlugins = namedParames.explodePlugins || "yes";
+	var wikifoldermaker = new WikiFolderMaker(this.params[0],tiddlerFilter,this.commander,options);
 	return wikifoldermaker.save();
 };
 
-function WikiFolderMaker(wikiFolderPath,wikiFilter,commander,explodePlugins) {
+function WikiFolderMaker(wikiFolderPath,wikiFilter,commander,options) {
 	this.wikiFolderPath = wikiFolderPath;
 	this.wikiFilter = wikiFilter;
 	this.commander = commander;
-	this.explodePlugins = explodePlugins;
+	this.explodePlugins = options.explodePlugins;
 	this.wiki = commander.wiki;
 	this.savedPaths = []; // So that we can detect filename clashes
 }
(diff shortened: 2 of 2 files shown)


Zacharia2:
After testing, it is working properly.

core/modules/commands/savewikifolder.js
```diff
@@ -35,13 +35,27 @@ Command.prototype.execute = function() {
...
+		case 3:
+			this.wikiFilter = this.params[1];
+			this.explodeplugins = this.params[2] === "explodeplugins=no" ? true : false;
+	}
+	var wikifoldermaker = new WikiFolderMaker(this.params[0],this.wikiFilter,this.explodeplugins,this.commander);
 	return wikifoldermaker.save();
 };
 
-function WikiFolderMaker(wikiFolderPath,wikiFilter,commander) {
+function WikiFolderMaker(wikiFolderPath,wikiFilter,explodeplugins,commander) {
```
pmario, line 55: IMO that's not backwards compatible. The new function parameter needs to be at the end. As is it can invalidate code, that imports the Command.

Zacharia2: Okay, I'll revise it.

linonetwo:
Do we have a method to create meta file/ json tiddler in the core? I only know it might be in filesystem plugin, not in the core, so can't be used here.

core/modules/commands/savewikifolder.js
```diff
@@ -149,12 +163,26 @@ WikiFolderMaker.prototype.saveCustomPlugin = function(pluginTiddler) {
...
+		});
+	} else {
+		var directory = path.resolve(this.wikiFolderPath,"tiddlers"),
+			pluginInfo = pluginTiddler.getFieldStrings({exclude: ["text","created","creator","modified","modifier","bag","tags"]}),
+			pluginInfoArr = [],
+			pluginJson = $tw.utils.parseJSONSafe(pluginTiddler.fields.text),
+			pluginName = pluginTitle.replace(/[/:]/g, '_') + ".json",
+			pluginMetaName = pluginName + ".meta";
+		this.saveJSONFile(directory + path.sep + pluginName, pluginJson);
+		for (let key in pluginInfo) {
```
pmario, line 181: We do not use `let` at the moment

Zacharia2: ok，Changed

Zacharia2:
@Jermolene Do you have any suggestions about the #7472 issue and this pr?

Jermolene:
Hi @Zacharia2 many thanks. I'm reviewing the PR now.

Apologies for the bureaucracy, but please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

Zacharia2:
@Jermolene Ok, I've signed the contributor agreement and submitted pr to the specified branch.

Jermolene:
My apologies @Zacharia2 – I have a couple of thoughts.

I wonder if the handling of the new option should be a bit more general purpose. We already have support for commands like "listen" that use name=value arguments like `--listen host=8080`.The core parses the arguments and makes them available via the `param` parameter of the Command object (instead of the usual array). We could do something similar but for commands that have a fixed number of arguments followed by a variable number of name=value options. For example, the savewikifolder command would have a single fixed argument followed by its options.

Anyhow, I think that should be addressed separately, and need not hold up this PR.

The other concern is that I think it might make things easier if the conditional logic were in the `WikiFolderMaker.prototype.save()` method. At the moment this PR duplicates a lot of logic concerned with writing the tiddler file. The alternative would be to extend the logic in the following line to test for the use of the flag:

https://github.com/Jermolene/TiddlyWiki5/blob/cce23ac6cddbccc88a848dcc5c456e57c01b2c20/core/modules/commands/savewikifolder.js#L88

That would trigger the call to `self.saveTiddler("tiddlers",tiddler)` which contains the logic for writing the tiddler file.

Zacharia2:
Hi, @Jermolene I looked into the `--listen` argument passing, and looked at other commands and arguments, and I found that there are roughly two types of argument passing, key-value pairs and non-key-value pairs. At the moment I have two or three ways, so I hope I can hear your opinion.

Way one: destructive command, SaveWikiFolderCommand to take the full key-value pairs, similar to: `-savewikifolder wikifolderpath=value explodeplugins=value filter=value` , so the advantage is relatively simple to implement.

Way two, good compatibility, take the non-key-value pair approach, similar to: `-savewikifolder <wikifolderpath> [<filter>] [noexplode]`, I have not figured out how to implement this approach.


Way three, command compatibility is good, but seems to be destructive to the rest of the code, as this way seems to be disallowed in the `extractNamedParameters` method of the `$:/core/modules/commander.js` class. 
https://github.com/Jermolene/TiddlyWiki5/blob/9b8db5dbb1baa4d94c857d30f8ed6e539eba72c3/core/modules/commander.js#L140-L146
This approach takes a combination of key-value pairs and non-key-value pairs, something like: `--savewikifolder <wikifolderpath> [<name>=<value>]... `, although there doesn't seem to be a code implementation of this approach at the moment. So if you have a better suggestion for this approach, I would appreciate if you could provide me with some ideas.



These are some of my thoughts and options on the explodeplugins option, what is your opinion?

Zacharia2:
Well, if you think the first option works, then I can quickly get the PR modified, for example by removing the redundant code and adding the judgement logic here:

https://github.com/Jermolene/TiddlyWiki5/blob/cce23ac6cddbccc88a848dcc5c456e57c01b2c20/core/modules/commands/savewikifolder.js#L96-L99

and use the `self.saveTiddler("tiddlers", tiddler);` method to save the plugin in JSON format to the tiddlers directory.

core/modules/commands/savewikifolder.js
```diff
@@ -32,17 +33,23 @@ var Command = function(params,commander,callback) {
 };
 
 Command.prototype.execute = function() {
-	if(this.params.length < 1) {
+	if(typeof(this.params.folder) == "undefined") {
 		return "Missing wiki folder path";
 	}
-	var wikifoldermaker = new WikiFolderMaker(this.params[0],this.params[1],this.commander);
+	if(this.params.explodeplugins == "no" || typeof(this.params.explodeplugins) == "undefined"){
```
pmario, line 39: usually we use `===` to avoid type-casting errors ... also for the `== "undefined"` in line 36

Zacharia2: ok

Jermolene:
Thank you @Zacharia2 

> At the moment I have two or three ways, so I hope I can hear your opinion.

> Way one: destructive command, SaveWikiFolderCommand to take the full key-value pairs, similar to: `-savewikifolder wikifolderpath=value explodeplugins=value filter=value` , so the advantage is relatively simple to implement.

The trouble with this approach is that it is not backwards compatible.

> Way two, good compatibility, take the non-key-value pair approach, similar to: `-savewikifolder <wikifolderpath> [<filter>] [noexplode]`, I have not figured out how to implement this approach.

The trouble with this approach is that it doesn't scale well if further options are added in the future

> Way three, command compatibility is good, but seems to be destructive to the rest of the code, as this way seems to be disallowed in the `extractNamedParameters` method of the `$:/core/modules/commander.js` class.

This is the approach that I was thinking of. It would indeed require some refactoring of existing code.

Zacharia2:
Hi @Jermolene  , I have found the simplest possible implementation, check out the submitted code for details. If you think it's ok, I'll add the full documentation. Hopefully there will be more suggestions.

Zacharia2:
This approach fits your description perfectly `--savewikifolder <wikifolderpath> [<name>=<value>]` .

I tried the code logic and it works fine, but more testing is needed, maybe this way of writing is not the norm, if you have better suggestions, you are very welcome

core/modules/commands/savewikifolder.js
```diff
@@ -5,7 +5,7 @@ module-type: command
 
 Command to save the current wiki as a wiki folder
 
---savewikifolder <wikifolderpath> [<filter>]
+--savewikifolder <wikifolderpath> [<name>=<value>]
```
Jermolene, line 8: It breaks backwards compatibility to remove the existing optional `<filter>` parameter. I think the new syntax should be:

```
--savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
```

Jermolene (commented):
Thank you @Zacharia2 the current approach breaks backwards compatibility

core/modules/commands/savewikifolder.js
Zacharia2: Yes, this breaks backward compatibility, but it also adds complexity because the filter is also an optional parameter. Can we put the filter inside `[<name>=<value>]`?

Zacharia2: <wikifolderpath> [<filter>] [ [<name>=<value>] ]* There doesn't seem to be a way to do this, what I can think of is that unless the filter is made mandatory so that it is accurate and simple to get, we may need to test one by one if the parameter is a filter parameter.

Zacharia2: Backward compatibility support I solved

Zacharia2:
@Jermolene 

I'm very sorry, I can't fix the backward compatibility issue because I can't get the filter parameters to stay correct and the second position of the array is not necessarily the filter parameter.

This is the limit of my ability.

core/modules/commands/savewikifolder.js
Zacharia2: Well, it's been solved

Zacharia2:
@Jermolene I have solved the backward compatibility problem. 

It can now be used in this way.

```
--savewikifolder <wikifolderpath>
--savewikifolder <wikifolderpath> [<filter>]
--savewikifolder <wikifolderpath> [ [<name>=<value>] ]*
--savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*

```

Jermolene:
Thank you @Zacharia2 that's great.

The regexp used for detecting filter expressions might be a bit too broad. It currently matches a command parameter that contains an opening square bracket followed by arbitrary characters and a closing square bracket. That means that it would match a command parameter such as `exclude=[1,2,3]`.

I think it might be more robust to use a regular expression to test for a new-style named parameter such as `/^[a-zA-Z0-9\.\-_]+=/g`. If params[1] does not match that pattern then prefix it with `filter=`.

That would also make the documentation simpler:

```
--savewikifolder <wikifolderpath> [ [<name>=<value>] ]*

...

The following options are supported:

* ''filter'': a filter expression defining the tiddlers to be included in the output
* ''explodePlugins'': set to "no" to suppress exploding plugins into their constituent shadow tiddlers (defaults to "yes")
```

Finally, would you be able to update the documentation? Many thanks.

Zacharia2:
hi, @Jermolene   I've done all the work, including refining and translating the documentation, and I've also fulfilled all your expectations about the form of the input parameters, such as 

```
--savewikifolder <wikifolderpath> filter=[all[tiddlers]] [ [<name>=<value>] ]* 
--savewikifolder <wikifolderpath> [all[tiddlers]] [ [<name>=<value>] ]* 

--savewikifolder <wikifolderpath> [ [<name>=<value>] ]* 
--savewikifolder <wikifolderpath>
--savewikifolder <wikifolderpath> [<filter>]
--savewikifolder <wikifolderpath> [<filter>] [ [<name>=<value>] ]*
```
Hope you'll like it

Zacharia2:
Hi, good evening Mr. @Jermolene 

First of all, thank you very much for taking time out of your busy schedule to reply to me, and for following this pr and giving me some very good advice.

I will listen to your suggestions and keep improving it to meet your expectations so that it can be merged into the main line.

I really need your feedback to tweak it, so let me know if you have any ideas, including dropping the pr. I will respect your idea to tweak it, or just drop the pr and do something else.

Again, thank you very much for your help.

core/modules/commands/savewikifolder.js
```diff
@@ -93,10 +114,13 @@ WikiFolderMaker.prototype.save = function() {
 						self.log("Adding built-in plugin: " + libraryDetails.name);
 						newWikiInfo[libraryDetails.type] = newWikiInfo[libraryDetails.type]  || [];
 						$tw.utils.pushTop(newWikiInfo[libraryDetails.type],libraryDetails.name);
-					} else {
+					} else if(self.explodePlugins == "yes") {
 						// A custom plugin
 						self.log("Processing custom plugin: " + title);
 						self.saveCustomPlugin(tiddler);
+					} else if(self.explodePlugins == "no") {
```
Jermolene, line 121: It is problematic to test for both "yes" and "no" because of ambiguity when the value is neither "yes" nor "no".  In these situations, I would normally only test for the non-default value, so we'd be testing `if(self.explodePlugins !== "no")` etc.

Also please note that the core coding style is to use triple equals signs for comparisons.

Jermolene (commented):
Thank you @Zacharia2, your work on this PR has been of a very high quality, and I appreciate your dedication.

I have just a couple of very minor points and then I think we can merge this for v5.3.0

core/modules/commands/savewikifolder.js
```diff
@@ -35,14 +42,28 @@ Command.prototype.execute = function() {
...
+		tiddlerFilter = this.params[1];
+	}
+	tiddlerFilter = tiddlerFilter || "[all[tiddlers]]";
+	explodePlugins = namedParames.explodePlugins || "yes";
+	var wikifoldermaker = new WikiFolderMaker(this.params[0],tiddlerFilter,this.commander,explodePlugins);
 	return wikifoldermaker.save();
 };
 
-function WikiFolderMaker(wikiFolderPath,wikiFilter,commander) {
+function WikiFolderMaker(wikiFolderPath,wikiFilter,commander,explodePlugins) {
```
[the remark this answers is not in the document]
Zacharia2: okk

core/modules/commands/savewikifolder.js
Zacharia2: okk

Zacharia2:
Thank you Mr. @Jermolene . I have modified these few small points. And tested without problems.

Jermolene:
Thank you @Zacharia2 for your patience and dedication. Are there other areas you would like to contribute?

Jermolene:
That's great to hear, thank you @Zacharia2.

Should we add a link to https://tw-cn.netlify.app/ to https://tiddlywiki.com?

Zacharia2:
Haha, you're welcome, I like a lot of the design of tiddlywiki. About that, I have no problem with it, but currently it's @linonetwo who is managing the Chinese community, so maybe we can ask him for his opinion.


hi，@linonetwo Do you have any ideas or want to add?

linonetwo:
@Zacharia2 Yes we can add a "friend link" in "Other Resources" tag, you can arrange another PR to add a new tiddler to tw5.com edition to add this.

linonetwo:
I've publish a prelease `@tiddlygit/tiddlywiki@5.3.0-prerelease-2023-06-30`, you can `pnpm i` in TidGi to use this feature in TidGi, and I can publish a prerelease of TidGi after your PR.

core/language/en-GB/Help/savewikifolder.tid
```diff
@@ -4,16 +4,27 @@ description: Saves a wiki to a new wiki folder
...
 
 * The target wiki folder must be empty or non-existent
 * The filter specifies which tiddlers should be included. It is optional, defaulting to `[all[tiddlers]]`
 * Plugins from the official plugin library are replaced with references to those plugins in the `tiddlywiki.info` file
 * Custom plugins are unpacked into their own folder
 
+The following options are supported:
+
+* ''filter'': a filter expression that defines the tiddlers to include in the output.
+* ''explodePlugins'': set to "no" to save plugins to the tiddlers directory of the target wiki folder, to suppress breaking plugins into their constituent shadow tiddlers (default is "yes").
```
pmario, line 18: @Zacharia2 ... I did just translate this info to German, but it does not make sense to me. 

explodePlugins .. yes means ... convert prlugins to shadows ... BUT the text above says "no" will do so.  -> confused!

-------------

IMO the whole thins should be re-worded

```
* ''explodePlugins'': defaults to "yes"
** ''yes'' will "explode" plugins into shadow tiddlers and save them to the tiddlers directory of the target wiki.
** ''no'' will suppress breaking plugins into their constituent shadow tiddlers. 
```

That's how I think it should be written. That's how I think it works. Except it does not work that way. ..

core/language/en-GB/Help/savewikifolder.tid
```diff
@@ -4,16 +4,27 @@ description: Saves a wiki to a new wiki folder
...
 A common usage is to convert a TiddlyWiki HTML file into a wiki folder:
 
 ```
 tiddlywiki --load ./mywiki.html --savewikifolder ./mywikifolder
 ```
+
+Save the plugin to the tiddlers directory of the target wiki folder:
+
+```
+tiddlywiki --load ./mywiki.html --savewikifolder ./mywikifolder explodePlugins=no
```
pmario, line 29: Does this command save the plugin as a **single** tiddler to the wiki-folder? If yes my first post / understanding is wrong.

Zacharia2:
@pmario  Yes. The only difference between them is which directory they are written to. If they are written to the tiddlers directory, the plugin will exist as a json file, if they are written to the plugins directory, the plugin will exist as multiple tid files and will be stored in the same way as it was originally stored in the plugins directory.

core/language/en-GB/Help/savewikifolder.tid
Jermolene: Hi @pmario

> explodePlugins .. yes means ... convert prlugins to shadows

the option relates to the savewikifolder command, so the wording refers to how plugins are handled by the savewikifolder command. The current behaviour is that tiddlers containing plugins are exploded into plugin folders containing individual tiddler files. The new option allows that behaviour to be suppressed, so that plugins are saved as JSON tiddler files.

core/language/en-GB/Help/savewikifolder.tid
Jermolene: Yes that's right, any plugins in the wiki will be saved as JSON files, and not exploded into the wiki plugins folder.

core/language/en-GB/Help/savewikifolder.tid
pmario: OK. Then it still should be written differently. eg:

```
* ''explodePlugins'': defaults to "yes"
** ''yes'' will "explode" plugins into shadow tiddlers and save them to a tiddlers plugin-directory
** ''no'' will suppress breaking plugins into their constituent shadow tiddlers. It will save the plugin as a single JSON tiddler
```

IMO that's much easier to read and understand. At least for me.

linonetwo: Yes, this is clearer.

Issue #7465, opened 2023-05-18
[BUG] nodejs build fails with message ""Global assignment is not allowed within modules on node.""


buggyj:
### Describe the bug

this has to do with https://github.com/Jermolene/TiddlyWiki5/pull/4624

'globalThis' is not defined in nodejs until version 12. ubuntu is still on version 10

I suggest adding something like this to the boot:

```
var nodeMajorVersion = Number(((process.version).split(".")[0]).slice(1));

$tw.utils.evalSandboxed(moduleInfo.definition,sandbox,tiddler.fields.title,(nodeMajorVersion < 12));
```

### TiddlyWiki Configuration

prerelease 5.3.0

buggyj:
the code needs a guard for nodejs only

```
var nodeMajorVersion;
if ($tw.node) nodeMajorVersion = Number(((process.version).split(".")[0]).slice(1));
```

pmario:
The latest Node.js LTS version is 18 .. So I think we should _not_ change anything in TW to maintain compatibility for Node 10 or even 12. Both of them are "dead" -> **Not maintained and therefore highly unsecure**. https://en.wikipedia.org/wiki/Node.js

There is no problem to update Node versions in ubuntu. So you should use a newer version

rmunn:
In fact, Node 20 just came out last month: https://nodejs.org/en/blog/announcements/v20-release-announce

I do agree with adding a check for the Node version, and refusing to run if it's too low (perhaps with a link to https://github.com/nodejs/release or https://nodejs.org/en/download/current or something). According to the chart at https://github.com/nodejs/release, Node 14 is just about to fall out of Maintenance mode, so by the time TW 5.3.0 releases, the earliest supported version of Node will be 16.

buggyj:
There will be tens of million of people running older versions of node - that is their choice. Many will not have permission to upgrade their computer installation or will be unwilling or unable to do it. I do not believe that tiddlywiki should police which versions of node it will run with - it did not do it in the past.

On a related note: I believe that the checking 'globalThis' should be configurable as there may be third party nodejs modules that will fail to run as they modify the globalThis.

Jermolene:
Thanks @buggyj – please could you make a PR with the changes you propose?

I do not think TW should fail with an error if running on an old version of Node.js, for the reasons that @buggyj mentions.  The wider points about baseline compatibility should be part of the ongoing discussions in #7353 about bumping the JS dialect we target.

[...]

PR #7492, opened 2023-05-24
Polyfill is not needed, only globalCheck

Arlen22:
Per discussion in https://github.com/Jermolene/TiddlyWiki5/issues/7465 and some additional testing, I discovered (or rather realized) that the polyfill is not needed for checking the global and the temp variable is sufficient for conducting the check. 

Furthermore, because we are in an isolated context, neither `global` nor `globalThis` are supposed to be defined.

So this removes the polyfill and uses the temp variable directly instead of assigning it to globalThis.


Jermolene:
Thanks @Arlen22

PR #7493, opened 2023-05-24
Add codemirror option to show trailing space

tavin:
~So, can be turned on by a shadow tiddler or a plugin etc.~

~Might be nice to add a UI settings toggle but of course it also requires CSS.
I have it working with a local custom theme.~

Bundles the relevant codemirror addon, with a UI toggle (default on), and some default CSS.


pmario:
I think there is something wrong. ... It seems this addon is added to the core cm-plugin. It probably should be a stand alone addOn. I'll have to have a closer look.

pmario:
@tavin ... Have a closer look at the cm-closetag AddOn. https://github.com/Jermolene/TiddlyWiki5/tree/master/plugins/tiddlywiki/codemirror-closetag

That's the way it should be done. CSS files are in the `files/addon/edit/trailingspace.css` directory. At the same level as the `trailingspace.js` file. 

cm-addons have a special plugin.info file: https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/codemirror-closetag/plugin.info  so they can be shown in the plugin import UI.

pmario:
Hi @tavin ... I'm sorry. I've been commenting a lot on your PR, without first welcoming you as a new contributor. ... So let's do that. 

**Welcome! ;)**

There is also a bit of administartion needed. 
See the [Contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md) and the 
[Contributors License Agreement](https://github.com/Jermolene/TiddlyWiki5/blob/master/licenses/cla-individual.md) which you need to **sign in a separate PR**

tavin:
> I think there is something wrong. ... It seems this addon is added to the core cm-plugin.

Hi @pmario, personally I felt it belongs in the core, but sure I can separate it.

Regarding the CSS, I guess I wasn't sure if this would be the responsibility of themes or whatnot. I will try adding a simplistic style though: `.cm-trailingspace { background-color: red; }`

Cheers.

tavin:
@pmario ok I've updated my PR as per your suggestions, but using the simple red background for now. I'll try to think about how to improve upon that. Actually I like it but I can appreciate that it won't work for everybody.

pmario:
I do like it, but I just saw that we may have a problem with palette settings too. I did test `#ff000020` which imo is already better.
But let me test different palettes. We may need to use a variable setting. 

grafik

pmario:
There is an other problem with a colour-indicator only. With TW v5.2.7 we introduced a new "save-button" for colour-blind users. ... So we will have the same problem here. 

So there should be a visual indication in addition to the color ... So IMO we should stick with the background image, until we do have a better option.

tavin:
@pmario I have copied the background image styles from the codemirror demo.

pmario:
@tavin, @Jermolene .. It looks good to me. 

- I did test all palettes and even if the background is hardcoded it works well with all default palettes.
- I did also test the enable/disable setting, which works OK.

Jermolene:
Thanks @tavin this is a worthwhile improvement, and is appreciated.

tavin:
Cheers @Jermolene you're welcome, and thanks for the guidance @pmario.

PR #7494, opened 2023-05-25
Sortby add field suffix

linonetwo:
I want to sortby tmo_priority field, which is High, Medium, Low

I want to use it like this

```js
[filtersToGetSomeTitles[]sortby:tmo_priority[High Medium Low]]
```

### DIFF
--- a/core/modules/filters/x-listops.js
+++ b/core/modules/filters/x-listops.js
@@ -162,10 +162,18 @@ Extended filter operators to manipulate the current list.
 	/*
 	Returns all items from the current list sorted in the order of the items in the operand array
 	*/
-	exports.sortby = function (source, operator) {
+	exports.sortby = function (source, operator, options) {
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@linonetwo you should be able to accomplish this with the sortsub operator: https://tiddlywiki.com/#sortsub%20Operator

Something along these lines (untested):
```

\define compareByPriority() [get[tmo_priority]match[High]then[3]] :else[get[tmo_priority]match[Medium]then[2]] :else[get[tmo_priority]match[Low]then[1]]

[filtersToGetSomeTitles[]sortsub:number<compareByPriority>]

```

linonetwo:
Thanks @saqimtiaz , I already looked at all sort-related operator/prefixs. But I wasn't aware of your `match then :else` usage, which works like switch-case statement.

PR #7499, opened 2023-05-26
ignore visual studio files

joshuafontany:
I plan on diving into some dev work in both VS Code and Visual Studio. This will prevent the git-history from getting spammed.


Jermolene:
Thanks @joshuafontany

PR #7505, opened 2023-05-31
Use functions for DateFormat calculation example

pmario:
Use functions for DateFormat calculations. The code is cleaner now.


pmario:
closed because it used the wrong target-branch

PR #7506, opened 2023-05-31
use functions for dateformat example

pmario:
Use functions for DateFormat v5.3.0 example. 

image

### DIFF
--- a/editions/tw5.com/tiddlers/features/DateFormat.tid
+++ b/editions/tw5.com/tiddlers/features/DateFormat.tid
@@ -1,5 +1,5 @@
 created: 20140418142957325
-modified: 20230226144359284
+modified: 20230531084305631
 tags: Features
 title: DateFormat
 type: text/vnd.tiddlywiki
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario what was your objective here? I'm not sure that the use of functions makes the example any clearer. The goal with examples that illustrate particular features is to make the smallest examples that demonstrate the functionality at hand. I think the updates here make things more complicated (for instance, introducing `modified-formatted` but not `created-formatted`).

pmario:
>... what was your objective here?

The current example uses: `<$text text={{{ [{!!created}format:date[YYYY-0MM-0DD]] }}}/>` because the intuitive version `{{{ [{!!created}format:date[YYYY-0MM-0DD]] }}}` creates a link, which is not wanted here. 

I think the definition `\function modified-formatted() [{!!modified}format:date[YYYY-0MM-0DD]]` and 
usage: `<<modifed-formatted>>` is a 100 times easier to understand and use.

The old format is still there, because I thought users should know about it and decide themself what is easier to use. 

>The goal with examples that illustrate particular features is to make the smallest examples that demonstrate the functionality at hand.

The existing example at tiddlywiki.com is not small, nor is it easy. IMO it's there because TW misses intuitive date calculation options. 

I think using `\function` definitions where ever we use the transclude-list workaround `{{{ [[filter]] }}}`, to evaluate filters, will result in much easier definitions and easier to understand code. 

>...  (for instance, introducing modified-formatted but not created-formatted).

I think at the moment it's needed for our users to have both versions documented, because we do use the text-widget workaround `<$text text={{{ [{!!created}format:date[YYYY-0MM-0DD]] }}}/>` a lot in the TW UI. 

**But** I can remove the text-widget version from the example, if you think it's confusing.

Jermolene:
Hi @pmario the example isn't working for me at the moment:

image

The trouble is that this is not a good example of a custom function because it only uses implicit parameters such as `{!!modified}` rather than properly parameterising the values that are to be processed.

We could improve the function usage like this:

```
\function date.to.human(d) [<d>format:date[YYYY-0MM-0DD]]
\function date.to.timestamp(d) [<d>format:date[TIMESTAMP]]

\function timestamp-modified() [date.to.timestamp{!!modified}]
\function timestamp-created() [date.to.timestamp{!!created}]
\function difference-days() [<timestamp-modified>subtract<timestamp-created>divide[86400000]floor[]]

* ''Modified date:'' <$text text={{{ [date.to.human{!!modified}] }}}/>
* ''Created date:'' <$text text={{{ [date.to.human{!!created}] }}}/>
* ''Difference in days:'' <<difference-days>> days
```

I know that one of the things you're trying to avoid is the `<$text text={{{ ... }}}/>` construction, but I think that that belongs in demos like this because it currently is the recommended way to output the result of a filter. If that's a problem we want to fix, then we must fix it globally, rather than just trying to fix it in the context of this one example.

pmario:
>Hi @pmario the example isn't working for me at the moment:

pmario:
>Hi @pmario the example isn't working for me at the moment:

It only works with v5.3.0-prerelease because it uses the `\function` pragma. I'll have a closer look to your suggestion

pmario:
From my point of view `{{{ [date.to.human{!!modified}] }}}` is a workaround and
the outer text-widget is a workaround to fix a workaround. `<$text text={{{ [date.to.human{!!modified}] }}}/>` 

I'll close the issue and need to think more about it.

PR #7511, opened 2023-06-02
Adds a deserialize filter operator

saqimtiaz:
This PR adds a `deserialize[]` filter operator that can extract the JSON representation of tiddlers from the input string using the deserializer module specified in the operand. This was previously only possible via JavaScript.

In particular, this will enable new workflows in combination with #7422, such as importing another wiki and extracting tiddlers from it via wikitext.

In order for this operator to be useful for our end users, it is necessary to provide documentation on the core deserializer modules and what they can be used for.  I have added a draft of this as a table in the documentation for the deserialize filter operator. [Preview here](https://tiddlywiki5-4omqgzl9p-jermolene.vercel.app/#deserialize%20Operator:%5B%5Bdeserialize%20Operator%5D%5D%20%5B%5Bdeserialize%20Operator%20(Examples)%5D%5D).

 However, this information would be helpful in other places as well and it might be prudent to move it to its own tiddler. I welcome feedback on how to organize and improve this documentation and where it should live.

Other primitives related to deserializers:
* Currently the `deserializers[]` filter operator can be used to see the names of the available deserializer modules in a wiki.
* Both the $dropzone and $browse widgets accept a deserializer attribute

### DIFF
--- a/core/language/en-GB/Misc.multids
+++ b/core/language/en-GB/Misc.multids
@@ -25,6 +25,8 @@ Encryption/RepeatPassword: Repeat password
 Encryption/PasswordNoMatch: Passwords do not match
 Encryption/SetPassword: Set password
 Error/Caption: Error
+Error/DeserializeOperator/MissingOperand: Filter Error: Missing operand for 'deserializer' operator
+Error/DeserializeOperator/UnknownDeserializer: Filter Error: Unknown deserializer provided as operand for the 'deserialize' operator
 Error/Filter: Filter error
 Error/FilterSyntax: Syntax error in filter expression
 Error/FilterRunPrefix: Filter Error: Unknown prefix for filter run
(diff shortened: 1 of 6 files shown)


btheado:
@saqimtiaz, very nice! And the output can be used directly with the `tm-import-tiddlers` message. For example:

```
<$button
  actions="""
    <$action-sendmessage
      $message="tm-import-tiddlers"
      $param={{{[<html-data>deserialize[text/html]]}}}
      importTitle="import-tiddlers-from-html"
    />
  """
>
Import tiddlers from html
</$button>
```

saqimtiaz:
> very nice! And the output can be used directly with the `tm-import-tiddlers` message.

Thank you @btheado. You will find an example of this pattern in some code I posted on Talk a while back to import tiddlers from another wiki on startup, in which I manipulate the import tiddler created for a silent import without user interaction: https://springer-prerelease.tiddlyhost.com/#LoadRemoteTiddlersActions

Jermolene:
Thanks @saqimtiaz

PR #7513, opened 2023-06-03
Switch off CamelCase linking by default

Jermolene:
It is proposed to switch CamelCase linking off by default for v5.3.0. The rationale is that CamelCase linking is confusing for new users and is not widely used.

As well as disabling the CamelCase parse rule by default, this PR includes a new parse rule that matches suppressed CamelCase links (eg `~TiddlyWiki`), and renders them as plain text without the tilde character.

Note that for backwards compatibility CamelCase linking is currently explicitly switched on for the documentation editions such as `tw5.com` and `dev`.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/wikilinkprefix.js
+++ b/core/modules/parsers/wikiparser/rules/wikilinkprefix.js
@@ -1,0 +1,40 @@
+/*\
+title: $:/core/modules/parsers/wikiparser/rules/wikilinkprefix.js
+type: application/javascript
+module-type: wikirule
+
+Wiki text inline rule for suppressed wiki links. For example:
+
+```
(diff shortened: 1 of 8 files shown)


pmario:
... and is not widely used. //anymore//

While I do have a nostalgic view about it, I think it should be deactivated by default. 

-------

I think the test for the new rule is missing ?!

pmario:
We do have the `tv-wikilinks` variable. Does it overrule the global definition? I think it should.

pmario:
May be we need a new `tv-camelcase` that locally overrules the global setting

pmario:
You are right. Thinking about it again, it's easy to create a `[[CamelCase]]` link. That's enough control, without the internal complexity to mess with variables.

PR #7515, opened 2023-06-03
Adds an addtiddlers action widget

saqimtiaz:
This PR is a first draft at an `$action-addtiddlers` action widget that accepts a JSON payload as an attribute, and adds any tiddlers in the payload to the wiki. This is intended to be used for example with the new `deserialize[]` operator, or with data retrieved via HTTP requests from backends that provide access to just the tiddlers as JSON in a separate file (such as TiddlyHost).

This new primitive should allow for entirely wikitext based - and thus easily customizable - workflows for parsing data to extract tiddlers and adding said tiddlers to a wiki.

Example syntax:
`<$action-addtiddlers $json=<<JSONData>> />`

Before I work on this further I welcome input on whether the following features should be baked into the widget, or handled using wikitext to iterate over and manipulate the JSON data before passing it to the widget:

1. **Logging of tiddlers added.** Should a log tiddler be created by the action widget of the tiddlers that are added? Alternatively the creation of a log tiddler could be handled with wikitext by extracting the titles of the incoming tiddlers from the JSON using the JSON operators. This would be a bit more work for users but would be much more flexible.

2. **Finer control over what gets added from JSON.** It might be desirable to have support for optionally excluding system tiddlers from being imported from the JSON data, as well as specifying whether or not existing tiddlers with the same title should be overwritten. (We could also allow filtering the tiddlers in the JSON data using a filter expression) However, given a `jsonset[]` operator, most of this could also be achieved by manipulating the JSON payload before passing it to the `$action-addtiddlers` widget. .

My personal inclination is that handling these usecases via wikitext is preferable for the flexibility it will offer. For example, given a `jsonset[]` operator a user could also decide to add a field to each tiddler added.

Update: JSON parsing errors now fail silently. Edited questions for clarity.

### DIFF
--- a/core/modules/widgets/action-addtiddlers.js
+++ b/core/modules/widgets/action-addtiddlers.js
@@ -56,10 +56,8 @@ Invoke the action associated with this widget
 AddTiddlersWidget.prototype.invokeAction = function(triggeringWidget,event) {
 	var  self = this,
 		importData = $tw.utils.parseJSONSafe(this.JSONData,function(err) {
-			return [{
-				title: "JSON error: " + err,
-				text: ""
-			}];
+			console.log("JSON error: " + err);
+			return [];
 		});
 	if(!$tw.utils.isArray(importData)) {
 		importData = [importData];
--- a/core/modules/widgets/action-addtiddlers.js
+++ b/core/modules/widgets/action-addtiddlers.js
@@ -1,0 +1,80 @@
+/*\
+title: $:/core/modules/widgets/action-addtiddlers.js
+type: application/javascript
+module-type: widget
+
+Action widget to send a message
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+var Widget = require("$:/core/modules/widgets/widget.js").widget;
+
+var AddTiddlersWidget = function(parseTreeNode,options) {
+	this.initialise(parseTreeNode,options);
+};
+
+/*
+Inherit from the base widget class
+*/
+AddTiddlersWidget.prototype = new Widget();
+
+/*
+Render this widget into the DOM
+*/
+AddTiddlersWidget.prototype.render = function(parent,nextSibling) {
+	this.computeAttributes();
+	this.execute();
+};
+
+/*
+Compute the internal state of the widget
+*/
+AddTiddlersWidget.prototype.execute = function() {
+	this.JSONData = this.getAttribute("$json");
+};
+
+/*
+Refresh the widget by ensuring our attributes are up to date
+*/
+AddTiddlersWidget.prototype.refresh = function(changedTiddlers) {
+	var changedAttributes = this.computeAttributes();
+	if(Object.keys(changedAttributes).length) {
+		this.refreshSelf();
+		return true;
+	}
+	return this.refreshChildren(changedTiddlers);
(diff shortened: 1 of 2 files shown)


saqimtiaz:
@jermolene absolutely no hurry on this but would appreciate some input on the questions stated above when you have the opportunity.

pmario:
I did move add 3 to the top after writing feedback for 1) and 2)

add 3) **Finer control over what gets added from JSON**
Reading this info and thinking about my concerns from 2), I thought: **Why does the new action-widget not use the default "js import workflow", instead of creating a new one.** 

If the existing workflow is used, all the existing hooks can be used and we can implement a fully automatic or a fully user controlled workflow. .. The plugin author or the user can decide. 

---

add 1) **Error handling**
I think it should fail silently as long as we can be sure, that the wiki still works. 
As soon as we cannot guarantee functionality, we should throw the RSOD. I know that's hard, but better save than sorry.

add 2) **Logging of tiddlers added**
IMO tiddlers imported over HTTP**S**, that contain executable code like libraries and `<script>` tags should be imported using the default "non silent" $:/Import mechanism. So users are aware and have the possibility to decide, if they want to import the content. 

If the default import mechanism is used it will be possible for plugins to have **full control** over the functionality. 

IMO importing tiddlers over HTTP:// should be blocked by default. There may be an option named "UNSECURE-something" to disable the default behaviour for debugging.

Just my thoughts
Mario

pmario:
>3) Finer control over what gets added from JSON. I suggest adding support for excluding system tiddlers from being imported from the JSON data, 

I personally need the possibility to import state-tiddlers of all types. 

> ... as well as specifying whether or not existing tiddlers with the same title should be overwritten

So the same "problems" do come up we already have with the "user driven" $:/Import mechanism. ... Specifically this functionality triggered my thought. Why not use the existing stuff with all the hooks in place.

saqimtiaz:
Hi @pmario, this PR proposes a widget that provides an affordance for adding tiddlers from a JSON payload, there is no HTTP involved. As such, I don't see the relevance or merit of discussing here whether or not imports over HTTP should be allowed.

While it is indeed possible to utilize the existing import mechanism to import tiddlers from a JSON payload silently ([example here](https://springer-prerelease.tiddlyhost.com/#LoadRemoteTiddlersActions)), the process is convoluted and needs an intimate understanding of the structure of the import tiddler and how to manipulate it. The purpose behind this exploratory PR, in combination with #7511, is to try and provide a more user friendly and configurable **wikitext based** approach to adding tiddlers from different payloads. From this perspective, the necessity to use JavaScript hooks to tweak the default import process is a shortcoming that we need to try to improve upon.

saqimtiaz:
Having had the chance to reflect on this, I believe the idea behind this action widget is dated and there are better alternatives now available to us. This idea arose as a result of discussing with fellow developers some code that was written before we had JSON operators in the core. The introduction of the JSON operators means that we no longer need to use the import mechanism to add tiddlers from JSON to a wiki, we can directly iterate over the JSON and use the `$action-setmultiplefields` widget to create the corresponding tiddlers.

However, there is a question of whether we can make the `$action-setmultiplefields` widget more robust and user friendly for such use cases, as discussed in #7519

Jermolene:
That was a very interesting experiment thank you @saqimtiaz

PR #7518, opened 2023-06-04
Update ShadowTiddlers.tid (updated)

twMat:
main change is added section about "overriding shadow tids"


pmario:
I think that's a good improvement. There can not be enough understandable docs about the internal mechanisms.

Jermolene:
Thanks @twMat

PR #7521, opened 2023-06-05
Pr for demo alternate store

joshuafontany:
Not sure if I should have merged main into demo-alternate-store before working, but this is the result.

The actual changes are in: `plugins/tiddlywiki/demo-alternate-store/rawmarkup.js`

I just pulled the bare minimum functions from boot.js into here to test. serves tw5.com-server and can write tiddlers back to disk, but many of the tests in ./bin/test.sh are failing (not sure why atm).

Mahalo (thanks) for starting this.


saqimtiaz:
@joshuafontany you may need to rebase the PR, currently it is shown a lot of unrelated commits which makes it difficult to review.

joshuafontany:
> @joshuafontany you may need to rebase the PR, currently it is shown a lot of unrelated commits which makes it difficult to review.

Thanks @Saq. That helps, as now the demo code is clearly readable.

pmario:
What's the purpose of this PR?

saqimtiaz:
@pmario this is an extension of the work in https://github.com/Jermolene/TiddlyWiki5/pull/7329

pmario:
I know, but what's the purpose of this implementation. Is it just an experiment? 
If yes it should be a draft. If no, there should be an advantage over the existing implementation

saqimtiaz:
> If no, there should be an advantage over the existing implementation

We do not have an existing working implementation of an alternate store and this furthers the so far incomplete work in #7329.

Yes, this might be more suitable as a draft PR at this time. However I would hope that we could endeavour to request clarification on such things in a manner that would be less prone to coming across as hostile to contributors.

pmario:
>What's the purpose of this PR?

It is meant as a question. I wanted to get more information about the plugin. That's it. No pun intended.

joshuafontany:
Hi @pmario @saqimtiaz. Thanks for clarifying. Yes, this has taken the changes that @Jermolene made in his PR in `boot/boot.js`, which opens up defining a $tw.Wiki module before booting the whole TW code. The new example code in `plugins/tiddlywiki/demo-alternate-store/rawmarkup.js` is a straight up copy/paste out of boot.js, with a few required helper methods from $tw.utils copied over and renamed to allow the $tw.Wiki parallel code to work.

Jermolene:
Thank you @joshuafontany this is exactly what I was planning to do in #7329 before I got distracted. Could you retarget this PR to target the branch for #7329? I think that would make the history easier to follow.

Jermolene:
By the way @joshuafontany would you be interested in working on an SQLite implementation? It was that idea that triggered #7329.

linonetwo:
Latest prerelease of TidGi https://github.com/tiddly-gittly/TidGi-Desktop/releases/tag/v0.8.0-prerelease11

has a [better-sqlite3](https://github.com/WiseLibs/better-sqlite3/blob/master/docs/api.md) instance at `$tw.utils.Sqlite`. Which is accessable for server side plugin. (TidGi is a launcher for nodejs wiki)

And there is a predefined table https://github.com/tiddly-gittly/TidGi-Desktop/blob/5651ea70c6301bbabacf45126b84a60b630b3a4b/src/services/database/index.ts#L47-L59

```sql
CREATE TABLE IF NOT EXISTS tiddlers (
          title TEXT PRIMARY KEY,
          text TEXT,
          type TEXT,
          created INTEGER,
          modified INTEGER,
          tags TEXT,
          fields TEXT,
          creator TEXT,
          modifier TEXT
          );
```

Usage is like

```js
// pure nodejs context
console.log(`## Hi!`);
let aaa = 3;

/** tw */

// now in worker_thread with $tw wiki access

let filterExample = $tw.wiki.filterOperators.sum((callback) => {callback({}, String(aaa));callback({}, String(aaa))})
console.log(filterExample)

// use Sqlite
const db = $tw.utils.Sqlite
const stmt = db.prepare('SELECT title, text FROM tiddlers');
const result = stmt.pluck().all()
console.log(result)
```

plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
```diff
@@ -0,0 +1,426 @@
...
+type: application/javascript
+module-type: library
+
+Startup code injected as raw markup
+
+\*/
+
+(function() {
+
+// Need to initialise these because we run before bootprefix.js and boot.js
```
linonetwo, line 12: Why not do these in bootprefix.js or boot.js, seems you init them but not used.

plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
```diff
@@ -0,0 +1,426 @@
...
+                    var key = keys[f];
+                    next = callback(object[key],key,object);
+                    if(next === false) {
+                        break;
+                    }
+                }
+            }
+        }
+    },
+    hop = function(object,property) {
```
linonetwo, line 65: Are we redefining too many things? I thought we just need to define some **hooks** to do "[hydration](https://react.dev/reference/react-dom/client/hydrateRoot)"

plugins/tiddlywiki/demo-alternate-store/rawmarkup.js
```diff
@@ -0,0 +1,426 @@
...
+		indexersByName[name] = indexer;
+		indexer.init();
+	};
+
+	this.getIndexer = function(name) {
+		return indexersByName[name] || null;
+	};
+
+	// Add a tiddler to the store
+	this.addTiddler = function(tiddler) {
```
linonetwo, line 111: I think we only need to define these fuctions in the plugin?

Jermolene:
Hi @joshuafontany I hope you don't mind, but I've merged your updated plain JS wiki implementation in 1d0b9280d8b50c39f958fd81b7994d41a5ebc891

PR #7524, opened 2023-06-06
Add Tabs To Settings

### LINKED ISSUE #5520: [BUG] Settings tab and third party plugins
**Describe the bug**
Several of kookma plugins use a Settings tab appears in $:/ControlPanel / Settings. It is desired to see these setting pages side by side of Tiddlywiki settings.

Unfortunately it seems this does not happen until you first install the CodeMirror plugin. CodeMirror overwrite the core tiddler [$:/core/ui/ControlPanel/Settings](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FSettings). If you install third party plugins with Settings page after you have installed CodeMirror then you see the Setting Pages!

**To Reproduce**
2. Install a third part plugin with Settings page e.g [TW-Vote](https://kookma.github.io/TW-Vote/#%24%3A%2Fplugins%2Fkookma%2Fvote)
3. Open $:/ControlPanel and then setings
4. You do not see Vote settings page  (see image 1)
5. Now install CodeMirror and you see the Settings tab (see image 2)

**Expected behavior**
The official TW shall support adding Settings tab from third party plugin. It seems $:/core/ui/ControlPanel/Settings needs to be revised and also prevent CodeMirror plugin overwriting shadow tiddler.

**Screenshots**

## Fig 1.  What you get if use empty.html with no codemirror plugin
258_chrome

## Fig 2. Expected behaviour
image

 - Version [e.g. v5.1.24]
 - Saving mechanism [e.g. Node.js, TiddlyDesktop, TiddlyHost etc]
 - Plugins installed [e.g. Freelinks, TiddlyMap]
 
 - OS: Windows 10
 - Browser Chrome 89
 - Version TW 5.1.23 and 5.1.24p

### LINKED ISSUE #7522: [IDEA] Standardize Settings Tabs
Currently, the code mirror plugin overrides the control panel's settings menu to add tabs. It then creates a tab for itself to allow you to select between changing tiddlywiki and code mirror settings. Other plugins use this same method of overwriting the settings menu and creating the tabs as well. 

I propose that those overrides become the default functionality, making the settings menu tabs there by default (of course only with a single tab), with the ability to add more tabs through tags.

This has two main benefits:

- It prevents two plugins from overriding the settings menu in two slightly different ways creating a conflict and breaking a plugin.
- It creates a standard place where users can expect to settings pages for installed plugins.

GameDungeon:
The title says it all.

Closes #7522 
Closes #5520


GameDungeon:
Ok, the changes seem to have worked. Ready for review

kookma:
Thank you @GameDungeon.
I checked with kookma library where most of plugins in this library used Settings tab. It works as expected.

I think @pmario the author behind CodeMirror plugin should check the changes. The reason is CodeMirror does not overwrite the Settings Tab and also does not create a second Settings tab. So, an update for CodeMirror ro TW 5.3.0 is needed.

One minor suggestion: 
Is it required to add `list-before` with empty value to `$:/core/ui/ControlPanel/Settings/TiddlyWiki` to have it the first tab all the time or do we need to add `sort[]` to the tablist in `$:/core/ui/ControlPanel/Settings`?

GameDungeon:
Well, that code is literally just the code mirror code moved into the core. I've also never seen it list the tiddlywiki tab second, so it seems to not be required.

core/ui/ControlPanel/Settings.tid
```diff
@@ -2,18 +2,6 @@ title: $:/core/ui/ControlPanel/Settings
 tags: $:/tags/ControlPanel
 caption: {{$:/language/ControlPanel/Settings/Caption}}
```
GameDungeon, line 4:
```suggestion
list-before:

```

core/ui/ControlPanel/Settings.tid
```diff
@@ -1,19 +1,8 @@
 title: $:/core/ui/ControlPanel/Settings
 tags: $:/tags/ControlPanel
 caption: {{$:/language/ControlPanel/Settings/Caption}}
+list-before:
```
GameDungeon, line 4:
```suggestion

```

kookma:
I also tested with CodeMirror installed. It works fine, but CodeMirror overwrite core tab!
@pmario I would recommend pushing an update for CodeMirror after this PR merged.

twMat:
I just posted a much [more general idea](https://github.com/Jermolene/TiddlyWiki5/issues/7522#issuecomment-1585559175) for this. IMO we should go with that idea instead of this PR.

pmario:
>kookma wrote
>
>I also tested with CodeMirror installed. It works fine, but CodeMirror overwrites core tab!
@pmario I would recommend pushing an update for CodeMirror after this PR merged.

@kookma ... The PR removes the core tab from the CM plugin. ... So it will fix that problem.

@twMat 
I think having the TiddlyWiki tab to be the only tab in the default setting is OK.
I'm in favour of the possibility to say: **Go to the ControlPanel -> Settings -> TiddlyWiki tab.**

With this PR the CM plugin only brings the new tab that will be added to the Settings tab. So there will be "TiddlyWiki" and "CodeMirror" Every other plugin can do the same thing as CM does from now on. With this PR there is no need to overwrite the core code anymore.

@Jermolene ... As far as I can see, the PR is good to be merged.

Jermolene:
Thanks @GameDungeon

PR #7526, opened 2023-06-07
Adds Text substitution support in widget attributes

### LINKED ISSUE #6663: Add support for string literal attributes with textual substitution
A common use case for macros is to construct a string from component parts. For example:

```
\define my-url()
https://$(host)$/$(path)$
\end

<$let host="example.com" path="/files">
  <a href=<<my-url>>>
    ...
  </a>
</$let>
```

It is proposed to introduce a shortcut syntax to cover these simple use cases.

The shortcut syntax uses backticks to quote the string, with the following substitutions:

* `$(variable-name)$` is replaced with the value of the variable
* `${filter}$` is replaced with the first result of evaluating the filter

The example above could be expressed as follows using the new syntax:

```
<$let host="example.com" path="/files">
  <a href=`https://$(host)$/$(path)$`>
    ...
  </a>
</$let>
```

An example using filters:

```
<div style.width=`${ [<width>multiply[3]] }$px`>
...
```

An example of transcluding a tiddler:

```
<div style.width=`${ [{mytiddler!!myfieldfield}] }$px`>
```

saqimtiaz:
This PR adds text substitution support as discussed in #6663. 

The `substitute[]` operator and backtick quoted attributes support the following substitutions in this order:
1. embedded filters using `${ filter }$` syntax
2. operands using the syntax `$n$` where `n` denotes the position of the operand (starting from 1).
3. variables using the syntax `$(varname)$`

Fixes #6663 

To do:
- [x] documentation for substitute filter operator
- [x] documentation for substituted widget attributes 
- [x] more tests for edge cases
- [x] support for using triple backtick characters to delineate the substituted attributes
- [x] refactor new widget method as a wiki method


Jermolene (commented):
> Should we perhaps prefix these variable names with the word `parameter`? So for example the first parameter would be available as the variable `parameter1` for text substitution purposes.

The loss of expressiveness does seem like a shame, but on the other hand, overwriting variables with such short names seems presumptuous.

Another possibility might be to change things around so that the positional parameters of the `substitute` operator are indexes into the list of variable substitutions contained within the text. For example:

```
<div attr=`Something about $(subject)$ at $(location)$`>
<$text text={{{ [<attr>substitute[Maths],[the Library]] }}}/>
</div>
```

Which would render: `Something about Maths at the Library`.

core/modules/widgets/widget.js
```diff
@@ -373,6 +373,21 @@ Widget.prototype.evaluateVariable = function(name,options) {
 	}
 };
 
+
+Widget.prototype.getSubstitutedText = function(text,options) {
```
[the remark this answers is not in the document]
saqimtiaz: @Jermolene I will adapt once you have refactored the method for making a fake widget. This issue occurred to me while writing the code but I figured it made sense to just follow your lead on how you resolve the issue with the `makeFakeWidgetWithVariables()`, since I need to call that method here.

saqimtiaz:
> Another possibility might be to change things around so that the positional parameters of the `substitute` operator are indexes into the list of variable substitutions contained within the text.

I fear that might get rather confusing when there are substitutions taking place from variables and parameters. 

Example where the first and third substitutions are from operands and the middle one from a variable:

```
<let time="morning">
<$text text={{{ [[Something about $(subject)$ in the $(time)$ at $(location)$]substitute[Maths],[],[the Library]] }}}/>
</let>
```

vs

```
<let time="morning">
<$text text={{{ [[Something about $(param1)$ in the $(time)$ at $(param2)$]substitute[Maths],[the Library]] }}}/>
</let>
```

saqimtiaz:
> Or perhaps the syntax parameters could be taken from the existing macro syntax:
> 
> ```
> <let time="morning">
> <$text text={{{ [[Something about $1$ in the $(time)$ at $2$]substitute[Maths],[the Library]] }}}/>
> </let>
> ```

This is the syntax I use in my own operator that does text substitution and is in fact exactly what an earlier iteration of the code in this PR did before we made parameters available as variables for the purpose of substitution. This approach does solve the problem of overriding variables with short names such as `1` but on the other hand it requires that users are aware of two different syntaxes for substitution of variables and parameters.

If we are going to have separate syntax for parameters and variables, consistency with the macro syntax would be preferable rather than introducing something completely new for users like double dollars. 

The current choices are between having the same syntax for variables and parameters at the cost of inelegant variable names like `parameter1`, or having different syntax for variables and parameters such as `$(varname)$` and `$1$` respectively. Alternatively, since this is new functionality might be expect that users will know to avoid using digits are their own variable names?

Jermolene:
> The current choices are between having the same syntax for variables and parameters at the cost of inelegant variable names like `parameter1`, or having different syntax for variables and parameters such as `$(varname)$` and `$1$` respectively. Alternatively, since this is new functionality might be expect that users will know to avoid using digits as their own variable names?

I am strongly in favour of aligning with the existing macro implementation. Even though it will be deprecated, it still exists and needs to be documented, and of course we have a generation of TW users familiar with the convention.

saqimtiaz:
Thank you @Jermolene, I will revert to the earlier implementation - which uses the macro syntax - when I next work on this.

Jermolene:
Thanks @saqimtiaz that will be marvellous.

saqimtiaz:
I have updated the OP with the current status of this work. The functionality is largely in place and the majority of the remaining work is documentation which I will work on this week.

 It would be helpful to get some input and testing at this time @btheado @yaisog @pmario @kookma

pmario:
The example in the https://github.com/Jermolene/TiddlyWiki5/pull/7526#issuecomment-1581273459 gave me a hard time with some headbanging. The typo was so obvious I could not see it. 

```
<$let time="morning">
<$text text={{{ [[Something about $1$ in the $(time)$ at $2$]substitute[Maths],[the Library]] }}}/>
</$let>
```

IMO you should also fix the typos `<let` -> `<$let` in the comment linked.

pmario:
ATM I'm testing some possible examples that can be explained. I'll post them here once I'm happy. 

-----

Just thinking.

The goal may be to "convert" the [final test](https://raw.githubusercontent.com/Jermolene/TiddlyWiki5/2ee4b4ed74fd5589432231cb6d14e31b6cdcb28e/editions/test/tiddlers/tests/data/filters/substitute.tid) into documentation and explain it ;)

saqimtiaz:
@pmario and @kookma more examples are extremely welcome and will be useful for the documentation. I am planning to work on that tomorrow and Thursday.

I highly recommend using the tests as guidance for syntax and **not** the examples in this discussion, as we have gone through a few different variations on the syntax.

saqimtiaz:
I have added some [documentation](https://tiddlywiki5-qkwhs0pmm-jermolene.vercel.app/#substitute%20Operator:%5B%5Bsubstitute%20Operator%5D%5D%20%5B%5Bsubstitute%20Operator%20(Examples)%5D%5D) for the `substitute[]` operator. More examples to add would be very welcome.

 Note that the "try it now" style macros for operator examples cannot be used for most of the examples as they process the variable substitutions since the syntax is the same. At some point we need to rewrite the macros for operator examples are procedures.

kookma:
@saqimtiaz 
Runing last example in doc results in error if I directly use it like below

```
<$text text={{{ [[This tiddler was last modified on ${[{!!modified}format:date[DDth MMM YYYY]]}$]substitute[]] }}} />
```

saqimtiaz:
@kookma that is expected, you cannot directly embed a filter in the input title to the operator, where that input title itself is wrapped in square brackets. This is the same reason we use variables as parameters to the subfilter operator.

I will add that as a tip to the documentation.

kookma:
I also recommend to have example with variables like below:

```
<$let
  nTiddlers= {{{ [all[tiddlers+shadows]count[]] }}}
  nShadows=  {{{ [all[shadows]count[]] }}} >

<$text text={{{ [[The total number of tiddlers in this TiddlyWiki are $1$ while $2$ are shadows.]substitute<nTiddlers>,<nShadows>] }}} />

</$let>

```

Results in:

The total number of tiddlers in this TiddlyWiki are 4157 while 2471 are shadows.

saqimtiaz:
I have added docs for [Substituted Attributes](https://tiddlywiki5-dbkddgbfj-jermolene.vercel.app/#Substituted%20Attribute%20Values) and have refactored the docs for widget attributes to [their own tiddler](https://tiddlywiki5-dbkddgbfj-jermolene.vercel.app/#Widget%20Attributes).

I am happy to add more examples if we can think of some that illustrate alternative usage patterns.

---

A small annoyance that has cropped up is that it is not possible to write an inline code run example with the triple backtick form of the substituted attribute. For single backticks quoted attributes, we can use the double backtick form of the code run markers (see #305.

kookma:
Wow, I can create Beautyful styling like below:

```
<$let color="aqua red blue orange purple Teal yellow magenta green" 
          nTids={{{ [enlist<color>count[]] }}} >

<$list filter="[tag[Learning]first<nTids>]" counter=n>
<span style=`color: ${ [enlist<color>nth<n>] }$`>
<$text text=<<currentTiddler>> /><br>
</span>

</$list>
</$let>
```

kookma:
Saq,
An example using triple backticks may be useful

```
<$text text=```This example contains backticks `$(currentTiddler)$` inside triple backticks.```/>
```

By the way I could not find a meaningful example

saqimtiaz:
I have added two more examples that use substituted widget attributes to create links, which I think is a fairly common use case for end users: https://tiddlywiki5-8e40is3rm-jermolene.vercel.app/#Substituted%20Attribute%20Values

saqimtiaz:
@Jermolene this is ready for review. Any suggestions for further improvement of the documentation would be very welcome.

btheado:
> Note that the "try it now" style macros for operator examples cannot be used for most of the examples

I rewrote `.operator-example` to not use textual substitution and it seems to work for your examples. It was a pretty straightforward translation. I wrote two versions. The first I kept as a `\define` and used `<<__var__>>`. The second I used `\procedure`.

I hope to create a PR for it sometime this weekend (I still need to do more testing). I assume the `\procedure` version is preferred? Any contrary opinions?

saqimtiaz:
> I rewrote `.operator-example` to not use textual substitution and it seems to work for your examples.

I suggest carefully verifying that the raw wikitext shown by the example is accurate and not substituting variables. It can be hard to spot. I think a macro approach will always run into this issue which suggest a procedure would be preferred.

btheado:
> I suggest carefully verifying that the raw wikitext shown by the example is accurate and not substituting variables

Good advice. I will be sure to test carefully. I pushed my changes and you can see them here: https://github.com/Jermolene/TiddlyWiki5/compare/master...btheado:TiddlyWiki5:remove-text-subst-from-operator-example-macro

I committed the `\define` first and then used a second commit to convert to `\procedure`. The conversion to procedure did nothing other than remove double underscores from variable names, so I think if one is right, the other is also.

All the input variable references are either on widget attributes or filter parameters, so I feel somewhat confident it is good. Will create a PR once I've taken the time to properly test.

editions/tw5.com/tiddlers/filters/examples/substitute Operator (Examples).tid
```diff
@@ -0,0 +1,37 @@
+created: 20230614225302905
+modified: 20230614233448662
+tags: [[Operator Examples]] [[substitute Operator]]
+title: substitute Operator (Examples)
+type: text/vnd.tiddlywiki
+
+\define time() morning
+\define field() modified
+\define sentence() This tiddler was last $(field)$ on ${[{!!modified}format:date[DDth MMM YYYY]]}$
```
btheado, line 9:
```suggestion
\procedure sentence() This tiddler was last $(field)$ on ${[{!!modified}format:date[DDth MMM YYYY]]}$
```

Without the above, I'm seeing the following in the display:
```
This tiddler was last modified on ${[{!!modified}format:date[DDth MMM YYYY]]}$
```

saqimtiaz: thanks @btheado

pmario:
I know the following is just cosmetics, but I think It would be much more readable if the "substitute Operator" placeholder table would look like this on a full HD screen.

**expected**

image

**as it is atm**

image

editions/tw5.com/tiddlers/filters/examples/substitute Operator (Examples).tid
```diff
@@ -6,7 +6,7 @@ type: text/vnd.tiddlywiki
 
 \define time() morning
 \define field() modified
-\define sentence() This tiddler was last $(field)$ on ${[{!!modified}format:date[DDth MMM YYYY]]}$
+\procedure sentence() This tiddler was last $(field)$ on ${[{!!modified}format:date[DDth MMM YYYY]]}$
```
kookma, line 9: Perhaps you mean `<<field>>`

saqimtiaz: @kookma no, the variable reference is intentional to show how it is further processed for variable substitution.

saqimtiaz:
> I know the following is just cosmetics, but I think It would be much more readable if the "substitute Operator" placeholder table would look like this on a full HD screen.

I am unsure about the merits of inserting manual line breaks to optimize for a specific screen size, however if we have an agreed upon policy for such things I will be happy to adopt it.

saqimtiaz:
@Jermolene I welcome any input on the documentation, otherwise this should be good to go.

editions/tw5.com/tiddlers/filters/examples/substitute Operator (Examples).tid
```diff
@@ -0,0 +1,37 @@
...
+!Substitute <<.op substitute[]>> operator parameters
+<<.operator-example 1 "[[Hi, I'm $1$ and I live in $2$]substitute[Bugs Bunny],[Rabbit Hole Hill]]">>
+
+!Substitute variables
+This example uses the following variables:
+
+* name: <$codeblock code=<<name>>/>
+* address: <$codeblock code=<<address>>/>
+
+<<.inline-operator-example "[[Hi, I'm $(name)$ and I live in $(address)$]substitute[]]">>
```
btheado, line 22: #7550 is in master now. After you merge master into your branch it should work to use `.operator-example`.

kookma:
Saq,
In documentation, I think the first column last row is a little confusing for newbies. I would recommend having `${filter expression}` in one line without that automatic line wrap (now it is in two line). I checked this both on FHD and HD screen.

saqimtiaz:
@Jermolene if you are able to review this in the next day or two that would be helpful, allowing me to make any necessary changes before I am offline for two weeks.

Jermolene:
Thanks @saqimtiaz

Issue #7528, opened 2023-06-08
[BUG] data-tag-title is not active in edit-mode


pmario:
### Describe the bug

From: https://talk.tiddlywiki.org/t/how-can-i-keep-the-style-i-defined-by-data-tags-even-in-the-editor/7217

... I have this piece of code in a stylesheet tiddler:

```
[data-tag-title^="#"] .tc-tag-label {
   border-radius: 3px;
}
```

This causes all tags prefixed with “#” to have a different border-radius property than the default. ([Custom styles by data-tags 1](https://tiddlywiki.com/#Custom%20styles%20by%20data-tags)).

But this doesn’t persist if I go to the tiddler editor tagged with the tag prefixed with “#”.
How can I make the custom style persist also in the tiddler editor?

### TiddlyWiki Configuration

v5.2.7

Jermolene:
Hi @pmario I am not sure what the problem is here. I think `data-tag-title` is assigned correctly to tiddlers rendered with the edit template:

https://github.com/Jermolene/TiddlyWiki5/blob/cce23ac6cddbccc88a848dcc5c456e57c01b2c20/core/ui/EditTemplate.tid#L30

Perhaps something else is going wrong?

pmario:
The `data-tag-title` is directly part of the tag-pill in view-mode, but missing in edit-mode. ... Here's the commit that added it. https://github.com/Jermolene/TiddlyWiki5/commit/e157d16b724172f752da0ff714847e0c0ca9664d ... I'll have to investigate more

pmario:
We need to add the info to 

- EditTemplate [$:/core/ui/EditTemplate/tags](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FEditTemplate%2Ftags)
- Tag Picker Template [$:/core/ui/TagPickerTagTemplate](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FTagPickerTagTemplate)

PR follows

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/e31c0937-93bf-4fa9-b666-152a1401cb6f)

pmario:
Are there any other places?

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/c1528090-c622-4037-8b16-f6a11af74548)

PR #7530, opened 2023-06-08
Add data tag title where needed

pmario:
- This PR fixes:   #7528 
- It also contains a new docs tiddler "Custom tap-pill styles by data-tag-title"
  - The naming is consistent with the existing data- attribute doc tiddlers. .. But it's a bit strange .. So improvements are welcome.

The screenshot below uses the following CSS settings **which are not active in the PR** but are part of the docs.

So if you want to test it, you'll need to create a stylesheet tiddler.

```
[data-tag-title^="#"] .tc-tag-label,
[data-tag-title^="#"].tc-tag-label {
   border-radius: 3px;
}
```

image


editions/tw5.com/tiddlers/howtos/Custom tag-pill styles by data-tag-title.tid
```diff
@@ -0,0 +1,32 @@
+created: 20230608121519758
+modified: 20230608123444591
+tags: [[How to apply custom styles]]
+title: Custom tap-pill styles by data-tag-title
```
Jermolene, line 4: Should be "tag", not "tap". Also don't need the dash. And I'm not sure that the title needs to mention "data-tag-title".

So, perhaps just `Custom tag pill styles`

Jermolene (commented):
Thanks @pmario just  a small point

Jermolene:
Thanks @pmario

PR #7533, opened 2023-06-09
Update Jasmine website

carlo-colombo:
pivotal.github.io/jasmine website does not exists anymore,
replaced with the current Jasmine website


Jermolene:
Thanks @carlo-colombo

PR #7538, opened 2023-06-14
Remove duplicate release note

rmunn:
https://github.com/Jermolene/TiddlyWiki5/commit/7b9915b5c237a577bc6931cd1619d5a615b35f39 added a release note for PR #7448, but it already had a release note, so now it has two. Let's remove one.

It's kind of arbitrary which one we should remove, so I chose to remove the one that I added a couple weeks ago, assuming that the one that @Jermolene added in commit 7b9915b5c237a577bc6931cd1619d5a615b35f39 reflects his preferred wording.


rmunn:
@Jermolene, if you decide you'd rather remove the release note line that you added and keep the one that I added, I'm happy to switch this PR around to make that change, to save you time. Just let me know which one you'd prefer.

Jermolene:
Hi @rmunn I think your wording is clearer actually, so I'll delete my one.

PR #7542, opened 2023-06-14
Add PR template for further automation

linonetwo:
Learn from https://github.com/CleverRaven/Cataclysm-DDA/blob/master/.github/pull_request_template.md


Jermolene:
Great stuff, thank you @linonetwo. I'm very keen to apply more automation to PR and issue handling, and agree that we can learn from other projects.

It would be great if PRs could include the update to the release note. The challenge there is that the name of the current release note changes with each release (the current one can be found in `editions/prerelease/tiddlers/Release 5.3.0.tid`).

Maybe we should switch to using a generic filename for the current release note (eg `editions/prerelease/tiddlers/Release.tid`), but keep the tiddler title as it is now.

linonetwo:
And you always add some non-standardized part at the top, so editor's message needs to be separated from auto-generated part, and transcluded back together...

截屏2023-06-14 17 24 37

linonetwo:
@Jermolene Can you merge this first? Other automation can be add step by step.

Jermolene:
Thanks @linonetwo this is good but I'd like to make some minor language tweaks which I'll push directly to this branch.

linonetwo:
> “Please grant permission for repository maintainers to edit your PR”

This make sure you can edit the text. that will be used in the release note! (also answer the first question).

It is very convenient for collaboration. I also enable it in this PR. You can try directly edit this PR to see if it works.

linonetwo:
> Why is it important that the comments in the template are not deleted?

The only possible reason maybe preventing only delete half of comment, cause render issue. Maybe some contributer is clumsy and cause this kind of trouble before.

You can delete it if you think people here will feel uncomfortable with it.

I don't know the real reason, but this template is fork from a opensource game with 20k PR, maybe they have encountered every possible kinds of strange situations...

linonetwo:
> a PR failing validation

You can still merge it.

We can also reflect this in action run title, for example, "PR style check", which feels very optional.

linonetwo:
> automatically compile the release note

We can have a separate github action that runs only when you git-tag vx.x.x, and generate release note tid from PRs between two tags.

Anyway, we will do it later.

linonetwo:
@Jermolene Ping

linonetwo:
@pmario what do you think about this one.

Jermolene:
Hi @linonetwo I’ve recently pushed some refinements to the release note implementation, and am working on some further improvements. Under the new model, a release note is comprised of one or more change notes, which may have one or more impact notes. The change note describes the change, including a change type along the lines here. The impact notes describes the impact of the change, things like pluginisation, or backwards compatibility break. By modelling these things as separate tiddlers we can do things like pull out all of the deprecations between two versions.

Once the new system is established it would be good to revisit the automation proposed here.

linonetwo:
Use #9407 instead

Issue #7428, opened 2023-05-05
[IDEA] Dot character as signifier for custom widgets and operators in v5.3.0


saqimtiaz:
In the current implementation of custom widgets and filter operators in v5.3.0, we use the `$$` prefix to identify custom widgets and the `.` prefix to identify custom operators. It would be desirable to have consistency between naming of the two types of custom affordances that users can create.

Custom widgets are quite similar to custom [Web Components](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) which are required to have a `-` character in the element name. While it would be desirable to adopt the same convention for custom widgets and operators in TW5, that is not possible as we already have operators with the dash character in the name. 

We should consider using the dot character in the same manner, that is for both custom widgets and operators the requirement would be that a dot character is present in the definition name.

joshuafontany:
Would it make sense for all custom widgets to start `$-` or `$.`? Possibly dot to match custom operators? I like enforcing this as a leading character...

Jermolene:
Thanks @saqimtiaz @joshuafontany I'm going to go ahead and implement this change, I think it makes a lot of sense to have a consistent convention between custom widgets and custom operators.

PR #7546, opened 2023-06-15
Remove stale reference to double dollar signs for custom widgets

btheado:
As of #7428  double dollar signs are no  longer used for custom widget names. Removed stale reference to this feature.


Jermolene:
Brilliant, thank you @btheado I was concerned that I hadn't found all the needed updates.

PR #7548, opened 2023-06-16
Fix and extend tag picker functions 

### LINKED ISSUE #7461: [BUG] calling tag-picker macro with a tagFields=foo shows the tags field instead of foo
### Describe the bug

calling tag-picker macro with a tagFields=foo shows the tags field instead of foo

To reproduce

- got to tiddlywiki.com
- copy paste the following code
- if the input is selected it shows the list using the `tags ` field instead of a list of all "foo-fields"

```
<<tag-picker tagField:foo>> 
```

image

### Expected behavior

It should use the content of all the foo-fields

pmario:
fix #7461  [BUG] calling tag-picker macro with a tagFields=foo shows the tags field instead

This PR works in the way that @ericshulman described in the https://github.com/Jermolene/TiddlyWiki5/issues/7461#issuecomment-1557541831 and should be 100% backwards compatible. If it's not that's a bug ;)

I did add 2 new parameters. 

- **tiddler** ... target tiddler where the tag should be set. .. (This parameter was completely missing)
- **tagListFilter** ... A filter that creates a list of tiddler-titles, that will be shown in the dropdown. 

There is a new: **tag-picker Macro (Examples)**  docs tiddler, which was missing. All parameters are shown, how the can be used.


pmario:
@ericshulman ... please test once the Vercel preview is built.

ericshulman:
Small typo in documentation:
```
The following examples us a temporary tiddler
```
should be:
```
The following examples use a temporary tiddler
```

pmario:
@ericshulman ... Are you OK with the implementation. Any incompatibilities?

ericshulman:
> @ericshulman ... Are you OK with the implementation. Any incompatibilities?

One additional change (maybe?)...

Currently, if two (or more) instances of `<<tag-picker>>` are used in the same tiddler, their edit fields all share the same temporary input tiddler: ``$:/temp/NewTagName`.  Perhaps this tiddler should be differentiated based on the target `tagField` value so that `<<tag-picker tagField:foo>>` and `<<tag-picker tagField:bar>>` have separate input values?

To accomplish this, in `tag-picker()` definition, replace:
```
<$set name="newTagNameTiddler" value=<<qualify "$:/temp/NewTagName">>>
```
with
```
<$set name="newTagNameTiddler" value={{{ [[$:/temp/NewTagName]] [<__tagField__>!match[tags]] +[join[/]] [<qualify>] +[join[]] }}}>
```
Note that this will only produce a different `$:/temp` tiddler title if a non-default `tagField` parameter is used (i.e., something other than "tags").  This should preserve backward compatibility with the existing default handling.

pmario:
@ericshulman should work now. 

Testing the tag-picker I found out that `$:/core/ui/EditTemplate/tags` needs to use the new params. So it will be possible to use the whole element with a different "targetTiddler" and a custom "tagField"

There have been several threads at Talk, that wanted to have the UI for custom field names. The code is almost finished. I will create a new PR soon.

pmario:
@Jermolene ... bump

AnthonyMuscio:
I did some work for my own reason on this here https://talk.tiddlywiki.org/t/how-to-make-it-possible-to-select-other-field-when-i-using-gen-tags-plugin/7880/4?u=tw_tones

I made it easy to stop the override of the core tag-picker-inner macro because the existing behaviour may have been appropriate in some cases.
* Someone may have used the tag picker macro and been happy for it to select only from the existing tags list to populate the alternative tag field. How likely is this? Not sure but such a change, although viewed as a bug, could damage backward compatibility.
* Perhaps if this may be an issue perhaps a switch or config setting?

If it proves of use I have attached my modifications including a new version of the core macro in $:/PSaT/alt-tags/tag-picker-inner-macro (the rest is just doco)
* It qualifies based on the tagField
* and modified the placeholder as well.
* it does separate the popup from the use on alt tag fields, however it does not stop the popup overlapping if the same tagField is used twice in the same tiddler, as is the case with the core tags anyway.
  * this is a bug that people change the transclusion variable for.
* If more extensive changes are to be implemented, I have also extended (in another tool) the tags bar both in the viewTemplate and an option in the editTemplate to allow the dropping of titles on the tag bar of a tiddler to tag with that title.
  * This is arguably an improvement that introduces a powerful usability enhancement, especially for handing user data.

[alt-tags.zip](https://github.com/Jermolene/TiddlyWiki5/files/12464921/alt-tags.zip)

pmario:
>- Someone may have used the tag picker macro and been happy for it to select only from the existing tags list to populate the alternative tag field. How likely is this? Not sure but such a change, although viewed as a bug, could damage backward compatibility.

The behaviour of the PR is 100% backwards compatible. 

Please read the docs and [read all the info for the examples.](https://tiddlywiki5-git-fork-pmario-7461-tag-picker-jermolene.vercel.app/#tag-picker%20Macro:%5B%5Btag-picker%20Macro%5D%5D%20%5B%5Btag-picker%20Macro%20(Examples)%5D%5D)

AnthonyMuscio:
Actually its even better because we can restrict the tags available to the tag regular tag picker.

How ever two tag pickers on the tags field but with different tagListFilter share popups.

CodaCodr:
@pmario Would be nice if this was <s>fixed</s> provided for, too: https://github.com/Jermolene/TiddlyWiki5/issues/7854

AnthonyMuscio:
> Please read the docs and [read all the info for the examples.](https://tiddlywiki5-git-fork-pmario-7461-tag-picker-jermolene.vercel.app/#tag-picker%20Macro:%5B%5Btag-picker%20Macro%5D%5D%20%5B%5Btag-picker%20Macro%20(Examples)%5D%5D)

I opened this wiki and have started to test;

I decided to add a tag field relating to the Features as follows;
```
<<tag-picker tagField:Features tagListFilter:"[tag[Features]]">>
```

The result in the dropdown is tag pills,  However you can click to add them to the features field, but there is no indicator that it has happened, such as closing the drop down etc...
* I don't know if there is a way without creating a custom version of the edit tags editor line?

The currently selected titles are not removed from the selection list if you use the tag list filter.
* We could add a more complex filter or build this into the tag picker.

pmario:
This PR is the first step for an improvement of the whole tag-picker line -- But this improvement is not possible without this PR. 

It has to be done in small steps to have a chance to get it merged.

CodaCodr:
> but there is no indicator that it has happened, such as closing the drop down etc...

I think that's a good thing. Having to reopen the dropdown adding six tags would be tedious.

Jermolene:
Hi @pmario on reflection, I'd like to suggest that we go further than this, and rewrite the macros properly as procedures before we attempt to add new features. I'd be really keen to revisit all the core macros for the next release.

pmario:
TL;DR
I'll close this PR since the main issue #7461 it wants to fix is completely different to the new goal set with comment: https://github.com/Jermolene/TiddlyWiki5/pull/7548#issuecomment-1825458523

---

@Jermolene 

- at the 17th of May a user opened a thread at Talk: [Tag-picker is not working independently](https://talk.tiddlywiki.org/t/tag-picker-is-not-working-independently/7029). 
- I did create an issue  #7461 the same day
- Eric Shulman presented a possible fix the same day - **but** Eric does not send PRs to GH
- At the GH issue #7461 there was some discussion about the implementation details.
- I did first suggest a new behaviour, that was not 100% backwards compatible 
- With a bit of back and forth Eric convinced me, that it can be implemented in a backwards compatible way. 
- May 22nd the discussion stopped. 
- June 10th I did create this PR according to the discussion going on at Talk and at the issue.
  - Everything was prepared to go into TW v5.3.0
- July 19th I did bump the PR the first time
  - Already missed v5.3.0
- August 20th v.5.3.1 went live -> Missing that one too :/
- 31th Aug. minor bug was discovered and fixed the same day. 
  - IMO this bug could have been found much earlier, if this PR would have ever been part of the TW preview
- Nov. 24th You decided to give this PR a completely new goal

The goal now is: 

>I'd like to suggest that we go further than this, and rewrite the macros properly as procedures before we attempt to add new features. I'd be really keen to revisit all the core macros for the next release.

**For me** the new goal has nothing to do with the issue raised at 17th of May, that I tried to fix **on behalf of someone else**. 

My motivation for this PR was to create a fix for a **different problem**, that is of interest for myself.

**What I actually want** to achieve is: Make the $:/core/ui/EditTemplate/tags -- edit-tags-template input **work with every field-name** instead of tags only. 

The `<<edit-tags-template` macro calls the `tag-picker`-macro.
So this PR would have been the first step needed to make **what I really want** happen. 

I'm sorry but with the new goal my motivation to invest more time into this PR and the new goal goes to zero.

What I really want to use is something completely different. 

--------

Changing existing macros in a 100% backwards compatible way from `\define` to `\procedures` and `\functions` is an incredibly complex task, because there are a lot of possible hickups. 

Especially, since most core templates need to be made "human readable" first. Most of them are not readable and also produce a lot of "[redundant P tags](WIP Improve redundant p tags #7061)" which is an even bigger problem for me. Where I do not see any light (without a hard fork) at all. 

----------

TL;DR
I'll close this PR since the main issue #7461 it wants to fix is completely different to the new goal set with comment: https://github.com/Jermolene/TiddlyWiki5/pull/7548#issuecomment-1825458523

Jermolene:
Hi @pmario I'm very happy to reconsider this for the next release.

The goal of make the edit-tags-template input work with every field-name instead of tags only is reasonable. The problem is only that we're now in the position that the core macros are mostly quite archaic, and don't take advantage of the latest features to improve conciseness and readability. Extending these macros without bringing them up to date makes the implementation more complex and hard to understand.

Perhaps what we need is a PR that modernises the existing global macros: making them be procedures, using $:/tags/Global, removing the remaining textual substitution, and fixing any remaining indentation issues. That would give us a better basis for moving forward with extensions to the current implementation.

AnthonyMuscio:
I support your approach @Jermolene however also think there may be value in exploring what kind of improvements can inform the rewrite.
* My feeling is we should implement some of these changes now and place the rewrite and some features in the future unless the rewrite can be achieved quickly. 

For example I have added a label parameter to the tag macro, so for example I can make a tag pill called "tagging" that lists all items with the current tag. I place this in the subtitle and it allows users to quickly find the tiddlers the current one is tagging.
* This needed two simple modifications.

Although not yet raised in this thread [RFC: Conditional UI elements Tabs/buttons and more](https://talk.tiddlywiki.org/t/rfc-conditional-ui-elements-tabs-buttons-and-more/8475) is it a simple change to also allow tag pills to respond to a condition field in tag tiddlers.
* This is also a somewhat simple modification to the existing code.

Whichever approach is taken I hope we can identify changes and features that will inform future tag macros, and also consider the introduction of other macros and names.
* In a related activity I have a very powerful implementation of a "filter pill" which was built on top of the current tags macro(s)
* Perhaps this could also inform the tag macro rewrite?

PR #7551, opened 2023-06-18
make all list-* widgets readable for easier future improvements

pmario:
>make all list-* widgets readable for easier future improvements

This PR adds indentation only. ... I would like to make the macros more consistent in the future. 

eg: list-links and list-links draggable use `tv-wikilinks=no` to render the `<$view field=title>` .... list-tagged-draggable does not, which prevents optimization. 

I would also like to make it easy to add a "search filter" to all list-* widgets without the need to change those macros again.

@Jermolene ... It would be nice if this PR could be merged.

### DIFF
--- a/core/wiki/macros/list.tid
+++ b/core/wiki/macros/list.tid
@@ -3,18 +3,18 @@ tags: $:/tags/Macro
 
 \define list-links(filter,type:"ul",subtype:"li",class:"",emptyMessage,field:"caption")
 \whitespace trim
-<$genesis $type=<<__type__>> class=<<__class__>>>
-<$list filter=<<__filter__>> emptyMessage=<<__emptyMessage__>>>
-<$genesis $type=<<__subtype__>>>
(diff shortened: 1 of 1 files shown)


core/wiki/macros/list.tid
```diff
@@ -3,18 +3,18 @@ tags: $:/tags/Macro
...
-<$link to={{!!title}}>
-<$let tv-wikilinks="no">
-<$transclude field=<<__field__>>>
-<$view field="title"/>
-</$transclude>
-</$let>
-</$link>
-</$genesis>
-</$list>
+<$genesis $type=<<__type__>> class=<<__class__>> >
```
Jermolene, line 6: Why is there an extra space inserted before the closing `>`? We are striving for consistency and it's unhelpful to adopt a different convention.

pmario:
@Jermolene .. I did update the PR as requested

Jermolene:
Thanks @pmario

Issue #7545, opened 2023-06-14
[BUG] CodeMirror plugin not working in HTA mode since TiddlyWiki v5.2.1


whjou:
### Describe the bug

Since https://github.com/Jermolene/TiddlyWiki5/issues/6702 where it was reported that CodeMirror plugin was not working  using Windows HTA Hack in TiddlyWiki v5.2.2, further investigation were done as documented in https://talk.tiddlywiki.org/t/hta-hack-and-codemirror/7001 and the issue was narrowed down to 1 line in https://github.com/Jermolene/TiddlyWiki5/pull/6278 which is using the arrow function that seems to be unsupported by the older Trident MSHTML engine on Windows; when it first appeared (TiddlyWiki v5.2.1); and that it is still present in TiddlyWiki v5.2.7.

### Expected behavior

Running latest version of TiddlyWiki (v5.2.7) using Windows HTA Hack with CodeMirror plugin enabled is expected to work.

### To Reproduce

1. Start with the empty edition at https://tiddlywiki.com/empty.html
2. Save it and rename file extension to .hta
3. Execute on a Windows PC (e.g. Windows 11)
4. From the Official TiddlyWiki Plugin Library, install the CodeMirror plugin.
6. Save and reload
7. See errors.

### Screenshots

![image](https://github.com/Jermolene/TiddlyWiki5/assets/696695/ca6cb9c2-af78-4eb2-8813-ab2b716d98e5)
![image](https://github.com/Jermolene/TiddlyWiki5/assets/696695/a983fb12-ece3-4c78-9168-84ee0cc3cabe)

### TiddlyWiki Configuration

- Version - 5.2.7
- Saving mechanism - Windows HTA Hack
- Plugins installed - Empty edition with CodeMirror plugin

### Desktop (please complete the following information):

- OS: Windows 11
- Browser : n/a (HTAs are rendered using Trident MSHTML engine on Windows)
- Version: n/a

### Additional context

The suggested fix to CodeMirror plugin is to rewrite the use of arrow function from

    setTimeout(() => cm.display.input.focus(), 20);

to the non-arrow function equivalent

    setTimeout(function() {cm.display.input.focus();}, 20);

pmario:
@Jermolene @saqimtiaz ... I think the journey @whjou has gone through for digging up the issue in #6702 deserves a fix. 

[...]

PR #7552, opened 2023-06-18
change arrow-function to function call for HTA

pmario:
This PR fixes:  **[BUG] CodeMirror plugin not working in HTA mode since TiddlyWiki v5.2.1** #7545 

This PR changes the arrow function call to a normal function call. The code itself is only executed in Chrome like browsers. 

I did the following tests as described in the original PR: https://github.com/Jermolene/TiddlyWiki5/pull/6278#issue-1063248880

>Things to test in CodeMirror - best to do this in a tiddler with some existing text:
>
>- dragging and dropping tiddler links
>- dragging and dropping external links
>- dragging and dropping text
>- whether the links or text are inserted at the correct place in the text
>- dragging and dropping an image and importing it.

It seems to work in Edge and FF ... I also did test it with HTA ... Should work again.


Jermolene:
Thanks @pmario

PR #4541, opened 2020-04-07
Bidirectional text improvements

Jermolene:
TiddlyWiki's language plugin mechanism includes a primitive mechanism for setting the HTML 5 "dir" attribute on the document body to "RTL", thus switching the direction of the user interface. This mechanism is used by languages such as Farsi and Hebrew.

Some of the issues with the current approach is that it is hard to work with tiddlers that have a different directionality to the main interface. For example, the document direction is "LTR" when using the default "en-GB" language plugin, but pasting (say) Farsi into the title, tags, or text fields of a tiddler doesn't switch direction to "RTL".

Prompted by [discussions](https://github.com/Jermolene/TiddlyWiki5/issues/2976#issuecomment-609932541) with @kookma, this experimental PR explores two separate mechanisms that help when working with text of a different directionality than the main interface.

The first is a new `\dir rtl` pragma that can be used at the top of a tiddler to specify that it should be rendered as RTL. This setting is respected if the tiddler content is transcluded elsewhere.

(A shortcoming at present is that the `\dir` pragma must be used after any `\define` pragmas if those macro definitions are to be imported elsewhere with the `\import' pragma).

The second change in this PR is support for a new $:/config/DefaultTextDirection hidden setting. It can be set to "rtl", "ltr" or "auto". It sets the "dir" attribute on several different parts of the view template and edit template.

The interesting thing is the setting "auto", part of the HTML 5 spec. It applies a simple heuristic to detect the appropriate direction for a fragment of text. The default value of $:/config/DefaultTextDirection is "auto". With this setting in place it becomes possible to paste fragments of RTL text into titles, tags, or text fields, and have things work as expected.

(A shortcoming that won't be going away is that the CodeMirror engine doesn't support "auto", just the explicit "RTL" and "LTR" settings. So I think we'll have to handle that entirely differently.

The change that I haven't explored here is adding a "_direction" field on individual tiddlers. That would work much better with the CodeMirror editor, but the disadvantage is that transclusion wouldn't respect the setting.

Issue #2976, opened 2017-09-14
How can I help to translate Tiddlywiki to Persian (Farsi)?


amirography:
Hi. I see that there have been developments on **RTL**, and Hebrew. I'm a native **Persian** user. I would love to be able to help with translating TW. How can I help? Where can I start?

amirography:
Ok, I found the 5.1.14 version of translator file. I'm at 191 out of 859. I'm not sure how it works. But I hope I just can do a delta translation for the pre-release, if possible.

pmario:
Hi, ... Welcome to the club :)

There are 2 ways to create translations. 

 a) using the [translators edition](http://tiddlywiki.com/prerelease/editions/translators/)
 b) using a [developers workflow](http://tiddlywiki.com/dev/#How%20to%20create%20a%20translation%20for%20TiddlyWiki)

For option a) ... Just download the translators edition ... change the english texts ... save and send the result to Jeremy :)

If you are familiar with nodejs and github, you can use option b)

pmario:
... :) ... just saw your second post .... 

> I'm not sure how it works. But I hope I just can do a delta translation for the pre-release, if possible.

Just translate the english texts. ... When finished just zip the version and drop the zip file here

amirography:
Sure thing! I'm working fast. I think I can knock it off in two days. Maybe a day for proofreading is needed. 

How about the delta thing? Can I just translate the difference between pre-release and current versions? Or should I copy everything manually?

pmario:
>How about the delta thing? 
Deltas are easier. ... 

>Can I just translate the difference between pre-release and current versions? 

You should go with the prerelease now. .... There's a dropdown "Show all strings" which will let you go for the new texts

>Or should I copy everything manually?

no. .... When we have a Persian version you'll be able to select it from the language dropdown, and just change the new texts.

pmario:
If there are just a view changes, it may be even possible to edit them directly in guthub. You can directly change the  .multids files. 

example: 

core english version for button texts: https://github.com/Jermolene/TiddlyWiki5/blob/master/core/language/en-GB/Buttons.multids
german version, which I contributed: https://github.com/Jermolene/TiddlyWiki5/blob/master/languages/de-DE/Buttons.multids

With the next build, they will be implemented. ... If some texts are missing in the translation, the core texts are shown. ... So the wiki should be still usable.

amirography:
[...]

PR #7557, opened 2023-06-21
Bidirectional text improvements

Jermolene:
<strong>This PR is a recreation of #4541 which was [merged](f90eb386ae7cd7da0f18b8d53fb0e454d9486d75) and then later [reverted](a05302da10de8404ca5aff793c661260673c4dfd) when it was discovered that it [caused a serious performance regression](https://github.com/Jermolene/TiddlyWiki5/pull/4541#issuecomment-1601137223).</strong>

---

TiddlyWiki's language plugin mechanism includes a primitive mechanism for setting the HTML 5 "dir" attribute on the document body to "RTL", thus switching the direction of the user interface. This mechanism is used by languages such as Farsi and Hebrew.

Some of the issues with the current approach is that it is hard to work with tiddlers that have a different directionality to the main interface. For example, the document direction is "LTR" when using the default "en-GB" language plugin, but pasting (say) Farsi into the title, tags, or text fields of a tiddler doesn't switch direction to "RTL".

Prompted by [discussions](https://github.com/Jermolene/TiddlyWiki5/issues/2976#issuecomment-609932541) with @kookma, this experimental PR explores two separate mechanisms that help when working with text of a different directionality than the main interface.

The first is a new `\dir rtl` pragma that can be used at the top of a tiddler to specify that it should be rendered as RTL. This setting is respected if the tiddler content is transcluded elsewhere.

(A shortcoming at present is that the `\dir` pragma must be used after any `\define` pragmas if those macro definitions are to be imported elsewhere with the `\import' pragma).

The second change in this PR is support for a new $:/config/DefaultTextDirection hidden setting. It can be set to "rtl", "ltr" or "auto". It sets the "dir" attribute on several different parts of the view template and edit template.

The interesting thing is the setting "auto", part of the HTML 5 spec. It applies a simple heuristic to detect the appropriate direction for a fragment of text. The default value of $:/config/DefaultTextDirection is "auto". With this setting in place it becomes possible to paste fragments of RTL text into titles, tags, or text fields, and have things work as expected.

(A shortcoming that won't be going away is that the CodeMirror engine doesn't support "auto", just the explicit "RTL" and "LTR" settings. So I think we'll have to handle that entirely differently.

The change that I haven't explored here is adding a "_direction" field on individual tiddlers. That would work much better with the CodeMirror editor, but the disadvantage is that transclusion wouldn't respect the setting.


kookma:
@Jermolene 
Is there any way to update this branch with the latest version of TiddlyWiki. I use the rtl features implemented in this branch. The latest version I have working with this branch is TW 5.3.0

kookma:
@Jermolene 
Is there any way to update this branch with the latest version of TiddlyWiki. I use the rtl features implemented in this branch. The latest version I have working with this branch is TW 5.3.0

Jermolene:
Hi @kookma I've updated this branch. Let me know if you run into anything.

pmario:
@Jermolene -- I did just have a look at the code here and I could not find any difference to the original "reverted" PR. So does this one still have the performance penalty?

Jermolene:
> @Jermolene -- I did just have a look at the code here and I could not find any difference to the original "reverted" PR. So does this one still have the performance penalty?

As I said in that original comment "Sadly, this PR will have to be reverted. I'll also create a new PR with the changes so that we can continue investigations". So at this point this is a duplicate of the original PR so that we can continue to experiment; we can't use #4541 because that has already been merged.

kookma:
> Hi @kookma I've updated this branch. Let me know if you run into anything.

Thank you, Jeremy! This works like TW-5.3.0pre.

kookma:
Hi @Jermolene ,
 Many thanks for all your efforts. The latest bidirectional version works great! It automatically switches to left and right based on the content.
I remember you said there is a performance impact for bidi version, and I assume most of it because of alot of `<p dir=auto`.
I tested TW 5.3.7p (the normal version NOT the bidi version) and I added below CSS

```
* { /* Or what selector you need */
  unicode-bidi: plaintext;
  text-align: start; /* Align text automatically left/right*/
}
```

and it works for BIDI as expected. So, I think the `dir=auto` can be dropped. I hope other user can check and confirm this!

kookma:
Hi @Jermolene 
Does this last commit mean the RTL/LTR will be supported by master branch?

Jermolene:
> Does this last commit mean the RTL/LTR will be supported by master branch?

Hi @kookma - no, not yet. What I did yesterday was update this PR to the latest "master" branch. It needs doing regularly with open PRs to ensure that it will be easier to merge when the time comes.

kookma:
Thank you very much! Sorry for inconveniences.

Jermolene:
> Thank you very much! Sorry for inconveniences.

Not at all, in fact thank *you* for being such a great advocate for these improvements, and your help has been invaluable in figuring out the details.

kookma:
Hi @Jermolene 
I would appreciate to update this branch to latest TiddlyWiki (perhaps 5.4.0).

kookma:
Hi @Jermolene 
Much appreciated. Works as expected.
It is worth to note that, the version showed on the demo wiki is **5.3.8-prerelease.**
I think it should be **5.4.0-prerelease**.

Thank you

Leilei332:
> Hi @Jermolene 
> Much appreciated. Works as expected.
> It is worth to note that, the version showed on the demo wiki is **5.3.8-prerelease.**
> I think it should be **5.4.0-prerelease**.
> 
> Thank you

This is because there is a backward incompatible PR was merged in v5.3.7, so v5.3.8 was created for a hotfix release.

kookma:
Is there a chance to have RTL support in TiddlyWiki 5.4?

pmario:
@Jermolene .. I did just read: https://evilmartians.com/chronicles/600-million-people-write-right-to-left-2-fixes-your-app-needs ... Which is interesting and it reminded me, that we already had something merged, but it got reverted. 

So I did tell Claude to run an assessment again. But instead of Puppeteer it used Playwright with Edge as target browser.

It found out, that the performance regression should have been gone with latest browsers. 

That's the summary: 

>The 45% regression that got #4541 reverted (2701ms to 3935ms) was **browser `mainRender`** cost, not Node/build: the tiddlywiki5-performance-test-rig builds the wiki, loads it under Puppeteer/Chromium, and reads `$tw.perf` — so the cost was 2023 Chromium resolving `dir="auto"` across the UI during layout.
>
>Mechanism: `dir="auto"` applied broadly. #4541 set it via `$:/config/DefaultTextDirection` (default `auto`) on ~5 elements per tiddler in the view/edit templates. The current #7557 branch (tip `ace2d2d51`, still ~5.3.8) ADDITIONALLY adds it to every `<p>` in `wikiparser.js parseBlock()` and to transclude function results.
>
>Measured 2026-06-06 (Opus): Node/fakedom ~0 cost (synthetic 5000 ¶: 25.98 vs 25.68ms; #4541 exact commits 12f7b98c4 vs f90eb386a identical). Modern Edge 149 also ~0 at scale — authentic rig repro at 400 tiddlers / 2030 live `dir="auto"` elements: mainRender 1794ms (before) vs 1647ms (after, within noise). Same TW diff thus gives +45% on 2023 Chromium but ~0 on current Edge: the variable was the browser engine, so the blocker is obsolete on today's browsers. Re-run the rig on current Chromium against a rebased branch before retiring #7557's perf objection; reproduce via `git worktree` at the two adjacent commits + build tw5.com with `$:/config/Performance/Instrumentation=yes`. See [[reference_playwright_local_setup]] and [[reference_visual_regression]].

Jermolene:
Hi @pmario

> It found out, that the performance regression should have been gone with latest browsers.

Good to know. We are planing to include this PR in v5.5.0.

pmario:
>Good to know. We are planing to include this PR in v5.5.0.

I think that would be nice, but IMO the main problem today is, that our "main" tester is not available at the moment, due to geopolitical problems.

kookma:
Hi @pmario 
I am available. The restriction with internet connection has gone. Not all websites are accessible, but the good news is I have access to GitHub and also tiddlywiki.com

I would be happy to help.

Jermolene:
> Hi @pmario I am available. The restriction with internet connection has gone. Not all websites are accessible, but the good news is I have access to GitHub and also tiddlywiki.com

Welcome back @kookma. You have been in my thoughts and I am delighted to hear that you are safe.

PR #7550, opened 2023-06-18
Remove text subst from operator example macro

btheado:
The textual substitution in the `.operator-example` macro prevents the macro from being used for some of the examples in the `substitute` operator (#7526). Remove the textual substitution syntax and convert from a macro to a procedure.

PR #7561, opened 2023-06-25
Use .operator-example for all substitute operator examples

btheado:
Take advantage of the macro improvements implemented in #7550 by converting all the substitute operator examples to use `.operator-example` instead of `.inline-operator-example`.


Jermolene:
Thanks @btheado

PR #7562, opened 2023-06-25
Various fr-FR translation improvements

xcazin:
Just caught up on the fr-FR translations before 5.3.0!


Jermolene:
Thanks @xcazin

PR #7582, opened 2023-07-03
Update Widget Attributes.tid

AnthonyMuscio:
Fixing error in 
* [[a transclusion of a macro/variable|Transcluded Attribute Values]]

Which should be;

* [[a transclusion of a macro/variable|Variable Attribute Values]]


AnthonyMuscio:
I checked the vercel version and the change is corrent

Jermolene:
Thanks @AnthonyMuscio – your fix looks correct but worth checking with @saqimtiaz who originally write this docs tiddler.

PR #7584, opened 2023-07-03
Update WidgetMessage_ tm-http-request Example Zotero.tid

stevesunypoly:
* corrected zotero groups available for import
* added field zotero-group to imported tiddlers


Jermolene:
Thanks @stevesunypoly

PR #7594, opened 2023-07-08
Add `tabindex` to SelectWidget and docs

### LINKED ISSUE #7593: [BUG] SelectWidgets don't get tabindex set
### Describe the bug

`SelectWidget`s don't get `tabindex` set.

### Expected behavior

The `SelectWidget` should pass through a `tabindex` property to the DOM node. 

A complete example is in https://talk.tiddlywiki.org/t/7482

### To Reproduce

1. Create a tiddler with some tab-able components
2. Include a `<$select ...>` widget among them
3. Set the tabindex on those components
4. Save the tiddler
5. Try to tab through the controls; note that the `<$select ...>` widget is skipped.

(My particular use-case was in using a `<$select ...>` as a custom field editor.  See https://talk.tiddlywiki.org/t/7482.)

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: 5.3.0 (but probably any earlier)
- Saving mechanism: Node.js (but presumably any)
- Plugins installed: none

- OS: Windows/Ubuntu
- Browser: Chrome/Brave/Firefox
- Version: any

- Device: Pixel 7
- OS: Android 13
- Browser: Brave/Firefox
- Version: any

### Additional context

_No response_

CrossEye:
Closes #7593 by adding a `tabindex` property to the `<$select ...>` widget and passing it through to the DOM.

Three questions:

  - The order of the fields in the code seems fairly arbitrary.  I put it between "size' and "tooltip", thinking that if these were ever alphabetized, this would already be in place.  Is there some other scheme I didn't notice?
  - I found one location to document the change: `editions/tw5.com/tiddlers/widgets/SelectWidget.tid` .  Are there others I should also be updating?
  - I added `"<<.from-version "5.3.1">>"`.  Is this presumptuous?  Should I not add anything and wait for approval and change it then?
 
And more generally, what might I, a newcomer to this source, do better in a (simple!) pull request?

### DIFF
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -145,6 +145,7 @@ SelectWidget.prototype.execute = function() {
 	this.selectDefault = this.getAttribute("default");
 	this.selectMultiple = this.getAttribute("multiple", false);
 	this.selectSize = this.getAttribute("size");
+	this.selectTabindex = this.getAttribute("tabindex");
 	this.selectTooltip = this.getAttribute("tooltip");
 	this.selectFocus = this.getAttribute("focus");
 	// Make the child widgets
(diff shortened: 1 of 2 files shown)


Jermolene:
Great stuff thanks @CrossEye

PR #7599, opened 2023-07-10
katex \label and \eqref

tavin:
What do people think of this? It probably needs a little fit & finish but... it works.

    $$
    \begin{align}
        \label{one} e^{i\pi} + 1 = 0 \\
        \label{two} E = mc^2 \\
    \end{align}
    $$

    If you were stranded on a desert island would you rather have $\eqref{one}$, $\eqref{two}$, or aspirin?

### DIFF
--- a/plugins/tiddlywiki/katex/wrapper.js
+++ b/plugins/tiddlywiki/katex/wrapper.js
@@ -16,7 +16,11 @@ var katex = require("$:/plugins/tiddlywiki/katex/katex.min.js"),
     chemParse = require("$:/plugins/tiddlywiki/katex/mhchem.min.js"),
 	Widget = require("$:/core/modules/widgets/widget.js").widget;
 
-katex.macros = {};
+katex.macros = {
+	'\\label': '\\htmlId{###1}{}',
+	'\\eqref': '\\htmlClass{katex-eqref}{(\\href{#####1}{\\htmlData{katex-label=#1}{\\text{#1}}})}',
+};
+
 katex.updateMacros = function() {
 	var tiddlers = $tw.wiki.getTiddlersWithTag("$:/tags/KaTeX/Macro"),
 		regex = /#\d/g, // Remove the arguments like #1#2
@@ -57,6 +61,9 @@ KaTeXWidget.prototype.render = function(parent,nextSibling) {
 	// Render it into a span
 	var span = this.document.createElement("span"),
 		options = {throwOnError: false, displayMode: displayMode, macros: katex.macros};
+	options.trust = function (ctx) {
+		return ctx.command == '\\href' || ctx.command == '\\htmlClass' || ctx.command == '\\htmlData' || ctx.command == '\\htmlId' && ctx.id[0] == '#';
+	};
 	try {
 		if(!this.document.isTiddlyWikiFakeDom) {
 			katex.render(text,span,options);
@@ -70,6 +77,23 @@ KaTeXWidget.prototype.render = function(parent,nextSibling) {
 	// Insert it into the DOM
 	parent.insertBefore(span,nextSibling);
 	this.domNodes.push(span);
+	var tiddlerFrame;
+	span.querySelectorAll('.katex-eqref [data-katex-label]:not([data-katex-eqnum])').forEach(element => {
+		// the tiddlyway requires a leading hash to jump inside a tiddler
+		var katexLabel = this.document.getElementById('#' + element.dataset.katexLabel);
+		if (!katexLabel) {
+			element.dataset.katexEqnum = element.dataset.katexLabel + '?';
+			return;
+		}
+		// everything is a <span> and the span holding a row doesn't even have a class
+		var katexRow = katexLabel.closest('.vlist > *');
+		var katexPos = Array.prototype.indexOf.call(katexRow.parentElement.children, katexRow);		// which row has the label
+		var katexEqn = katexRow.closest('.katex-html').querySelectorAll('.tag .eqn-num')[katexPos];	// which eqn num corresponds
+		// nodejs doesn't support ||= here ... wtf?
+		if (!tiddlerFrame) tiddlerFrame = span.closest('.tc-tiddler-frame');
+		if (!tiddlerFrame._katex_eqn_num_elements) tiddlerFrame._katex_eqn_num_elements = tiddlerFrame.getElementsByClassName('eqn-num');
+		element.dataset.katexEqnum = 1 + Array.prototype.indexOf.call(tiddlerFrame._katex_eqn_num_elements, katexEqn);
+	});
 };
 
 /*
@@ -88,7 +112,7 @@ KaTeXWidget.prototype.refresh = function(changedTiddlers) {
 		this.refreshSelf();
 		return true;
 	} else {
-		return false;	
+		return false;
 	}
 };
 
--- a/plugins/tiddlywiki/katex/styles.tid
+++ b/plugins/tiddlywiki/katex/styles.tid
@@ -19,6 +19,16 @@ tags: [[$:/tags/Stylesheet]]
 	counter-reset: katexEqnNo;
 }
 
+/* Replace \eqref placeholder with generated eqnum */
+
+.katex-eqref [data-katex-eqnum] > * {
+	display: none;
+}
+
+.katex-eqref [data-katex-eqnum]::after {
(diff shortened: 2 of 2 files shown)


plugins/tiddlywiki/katex/wrapper.js
```diff
@@ -70,6 +77,23 @@ KaTeXWidget.prototype.render = function(parent,nextSibling) {
 	// Insert it into the DOM
 	parent.insertBefore(span,nextSibling);
 	this.domNodes.push(span);
+	var tiddlerFrame;
+	span.querySelectorAll('.katex-eqref [data-katex-label]:not([data-katex-eqnum])').forEach(element => {
```
Jermolene, line 81: For backwards compatibility, the TiddlyWiki core restricts itself to ECMAScript 5.1. Instead of the `forEach` method the core has a `$tw.utils.each()` helper for iteration.

plugins/tiddlywiki/katex/wrapper.js
```diff
@@ -70,6 +77,23 @@ KaTeXWidget.prototype.render = function(parent,nextSibling) {
...
+		// the tiddlyway requires a leading hash to jump inside a tiddler
+		var katexLabel = this.document.getElementById('#' + element.dataset.katexLabel);
```
Jermolene, line 83: It looks like this solution uses DOM IDs to identify KaTeX formulae. The core avoids using IDs because the restriction a given ID can only be used once is problematic when in TiddlyWiki it is possible to open a tiddler more than once. For example, given a tiddler "Formulae" that contains a bunch of KaTeX formulae, it would be possible to both open that tiddler and alongside it display another tiddler that transcludes "Formulae".

Jermolene (commented):
Thanks @tavin a couple of points. But I think the big problem is that this approach will not work when generating KaTeX on the server under Node.js. If you're running Node.js you can test the static output with:

```
tiddlywiki editions/katexdemo --verbose --build static
```

pmario:
As Jeremy pointed out 2 tiddlers with the same formula could be open at the same time in a different text context, which may be confusing.

If the `\label{name}` is unique for a specific formula. The reference with the right name would jump to the right formula, even if it is open 2 times.

The bigger problem I see, is that CSS numbering is used, which is (1), (2), (3) and the references links are also (1), (2), (3). 

So if 2 tiddlers are open at the same time with **different formulae**, the numbering would be confusing for the reader, since they have the same numbering. 

So from my point of view label should use a text like reference similar to the `\tag{text}` format. 
Just some thoughts.

pmario:
@tavin Do you know that the katex-plugin has a built in macro editor. https://tiddlywiki.com/plugins/tiddlywiki/katex/#%24%3A%2Fplugins%2Ftiddlywiki%2Fkatex%2Fui%2Fconfig%2Fmacro

In the code you hardcoded 2 new macros. Would it be possible to use the editor instead?

Jermolene:
Hi @tavin the output from running `tiddlywiki editions/katexdemo --verbose --build static` is as follows:

```

Rendering "$:/core/templates/static.template.html" to "/Users/jeremyruston/git/Jermolene/TiddlyWiki5/editions/katexdemo/output/static.html"
$:/plugins/tiddlywiki/katex/wrapper.js:81
        var tiddlerFrame = parent.closest('.tc-tiddler-frame');
                                  ^

TypeError: parent.closest is not a function
    at KaTeXWidget.render ($:/plugins/tiddlywiki/katex/wrapper.js:81:28)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:631:15)
    at ElementWidget.render ($:/core/modules/widgets/element.js:83:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:631:15)
    at TranscludeWidget.render ($:/core/modules/widgets/transclude.js:35:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:631:15)
    at TranscludeWidget.render ($:/core/modules/widgets/transclude.js:35:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:631:15)
    at RevealWidget.render ($:/core/modules/widgets/reveal.js:48:7)
    at Widget.renderChildren ($:/core/modules/widgets/widget.js:631:15)

Node.js v18.4.0
```

The problem is that on the server we use a minimal DOM implementation that doesn't support many DOM features, including selectors. You can see the same problem in the browser by attempting to export as HTML a tiddler containing KaTeX formulae.

tavin:
> The problem is that on the server we use a minimal DOM implementation that doesn't support many DOM features, including selectors. You can see the same problem in the browser by attempting to export as HTML a tiddler containing KaTeX formulae.

Thanks @Jermolene I guess everything I am doing with DOM functions can be done another way, or maybe a polyfill is possible. That said, when can the server's DOM implementation be upgraded?

So far I think I've at least fixed ecmascript 5 compatibility and the unique ID issues.

tavin:
Hi @pmario when latex `\eqref` appears it *must* be replaced by a number. Well, unless the referenced equation has a custom `\tag`, but I haven't gotten that far yet.

As for the katex macro editor, sure I can use this, but if this work can be polished and accepted as a feature, I think we don't want to tell users they have to copypaste these macros as a secondary install step?

tavin:
Hi @Jermolene, @pmario. Well I found the `fakedom.js`. So I guess it's better to abandon any hope of transforming katex html in that code path. Therefore I am thinking of recasting my work as a browser only plugin.

What I'm not sure about is the right way to hook into the katex plugin's render method. Do I just reassign `$tw.modules.types.widget['$:/plugins/tiddlywiki/katex/wrapper.js'].exports.katex.prototype.render`? Or does TW have a mechanism for this?

Jermolene:
Hi @tavin thanks for your patience on this.

> Hi @Jermolene, @pmario. Well I found the `fakedom.js`. So I guess it's better to abandon any hope of transforming katex html in that code path. Therefore I am thinking of recasting my work as a browser only plugin.

I'm afraid that's also the conclusion that I had reached.

But we should give some consideration to the possibility of upgrading fakedom.js; even modest improvements would enable a lot more plugins to work on the server and for static exports. I wouldn't like to undertake support for selectors, but if the code was refactored to walk the DOM manually then it is conceivable that we could get it working.

> What I'm not sure about is the right way to hook into the katex plugin's render method. Do I just reassign `$tw.modules.types.widget['$:/plugins/tiddlywiki/katex/wrapper.js'].exports.katex.prototype.render`? Or does TW have a mechanism for this?

Ideally we would add some kind of hook for plugins that need to do this kind of thing. There is an existing hook mechanism in the core that may be suitable, or perhaps we'd introduce a new module type for KaTeX extensions, and have the KaTeX plugin track those extension modules, and invoke them at key points in the lifecycle.

tavin:
> But we should give some consideration to the possibility of upgrading fakedom.js; even modest improvements would enable a lot more plugins to work on the server and for static exports. I wouldn't like to undertake support for selectors, but if the code was refactored to walk the DOM manually then it is conceivable that we could get it working.

Well adding stuff like `parentElement` and `getElementsByTagName()` should be doable. How much is missing to be able to use `katex.render()` instead of `katex.renderToString()`?

A lot changes in 10 years... there are now various dom implementations that can run under nodejs. There might be reasons against pulling them into TW (I don't know). But even sticking with fakedom, maybe selectors are yours for free? I came across e.g. https://github.com/dperini/nwmatcher and https://github.com/dperini/nwsapi.

Jermolene:
Hi @tavin 

> Well adding stuff like `parentElement` and `getElementsByTagName()` should be doable.

Yes indeed, many DOM methods are fairly trivial and should be quite doable.

> How much is missing to be able to use `katex.render()` instead of `katex.renderToString()`?

Interesting question, it's a long time since I looked at that. But one would expect that rendering would not need to use selectors because they are mainly used for locating parts of existing DOM structures for subsequent manipulation. It would definitely be worth experimenting with.

> A lot changes in 10 years... there are now various dom implementations that can run under nodejs. There might be reasons against pulling them into TW (I don't know). But even sticking with fakedom, maybe selectors are yours for free? I came across e.g. https://github.com/dperini/nwmatcher and https://github.com/dperini/nwsapi.

The concern would be the size and performance of those libraries (our current fakedom is quite quick because it is so simple). One possibility might be to ship something like JSDom as an optional plugin that replaces the fakedom with a fuller implementation for users who need it.

PR #7600, opened 2023-07-11
Update Anchor Links using HTML.tid

AnthonyMuscio:
Removed broken link that is out of place.

<a href="##Suggestions%0">Here are some suggestions of tags you can use.</a>

As raised here by stobot https://talk.tiddlywiki.org/t/small-typo-on-tiddlywiki-com/7404?u=tw_tones


Jermolene:
Thank you @AnthonyMuscio – the link is actually an internal link within the same tiddler, but it has been misformatted. I'll post an updated fix momentarily.

PR #7603, opened 2023-07-12
Make the "Contents" menu item caption configurable in MenuBar plugin

tw-FRed:
The purpose of this PR is to enable "Contents" menu item caption to be configurable by adding a "caption" field to the custom ToC tag. A "description" field can also be added to the tag for the plugin configuration tiddler.

### DIFF
--- a/plugins/tiddlywiki/menubar/config.tid
+++ b/plugins/tiddlywiki/menubar/config.tid
@@ -1,6 +1,8 @@
-title: $:/plugins/tiddlywiki/menubar/config
-tags: $:/tags/ControlPanel/Toolbars
 caption: Menu Bar
+created: 20230519012218346
+modified: 20230712224614730
+tags: $:/tags/ControlPanel/Toolbars
+title: $:/plugins/tiddlywiki/menubar/config
 
 \define config-base() $:/config/plugins/menubar/MenuItems/Visibility/
 
@@ -28,7 +30,7 @@ The breakpoint position between narrow and wide screens. Should include CSS unit
 
 !! Contents Tag
 
-The tag for the ~TableOfContents used in the Contents dropdown
+The tag for the ~TableOfContents used in the Contents dropdown. Add //caption// and //description// fields to the tag tiddler to change ~TableOfContents' menu label.
 
 <$edit-text tiddler="$:/config/plugins/menubar/TableOfContents/Tag" default="" tag="input"/>
 
--- a/plugins/tiddlywiki/menubar/items/contents.tid
+++ b/plugins/tiddlywiki/menubar/items/contents.tid
@@ -1,8 +1,10 @@
-title: $:/plugins/tiddlywiki/menubar/items/contents
-caption: Contents
-description: Table of Contents
+caption: <$text text={{{ [{$:/config/plugins/menubar/TableOfContents/Tag}get[caption]else[Contents]] }}}/>
+created: 20230519010904497
+description: <$text text={{{ [{$:/config/plugins/menubar/TableOfContents/Tag}get[description]else[Table of Contents]] }}}/>
 is-dropdown: yes
+modified: 20230712225429073
 tags: $:/tags/MenuBar
+title: $:/plugins/tiddlywiki/menubar/items/contents
 
 <div class="tc-table-of-contents">
 


plugins/tiddlywiki/menubar/config.tid
```diff
@@ -28,7 +30,7 @@ The breakpoint position between narrow and wide screens. Should include CSS unit
 
 !! Contents Tag
 
-The tag for the ~TableOfContents used in the Contents dropdown
```
Jermolene, line 31: Would it be worth including text editor widgets that allow the caption and description fields to be directly inspected/edited?

plugins/tiddlywiki/menubar/config.tid
```diff
@@ -1,6 +1,8 @@
-title: $:/plugins/tiddlywiki/menubar/config
-tags: $:/tags/ControlPanel/Toolbars
 caption: Menu Bar
+created: 20230519012218346
```
Jermolene, line 2: We avoid created/modified fields on shadow tiddlers.

Jermolene (commented):
Thanks @tw-FRed

The main effect of these changes appears to be to override the naming of the "Table of Contents" item in the list of available menu items in the menu bar configuration control panel tab.

My concern is that changing the labelling of a user interface item makes it harder to write the documentation, because the entries described by the docs no longer match what the user sees. Perhaps the labelling of the item should stay as "Table of Contents" with the caption/description field being put in parenthesis immediately afterwards.

tw-FRed:
Thanks @Jermolene for your review 🙂
Looking at the PR result in Vercel I see it's not working as intended.
This is because the default TableOfContent tag already has a caption value of `{{$:/language/SideBar/Contents/Caption}}`. This caption should already allow translation of the menu item in the menu bar, but currently this is not working in the French edition of TiddlyWiki, although the translation exists in $:/language/SideBar/Contents/Caption.
You can see the menu item isn't translated in https://tiddlywiki.com/languages/fr-FR/index.html (should show "Sommaire" instead of "Contents").
image

I marked the PR as Draft and I will rework it.

Fred

tw-FRed:
TL;DR: @Jermolene I've got a quick fix for the menu item translation bug (explained below), and a more thorough and user-friendly solution (test full fix [here](https://tiddlywiki5-72ybt3s74-jermolene.vercel.app/#%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fitems%2Fcontents:%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fitems%2Fcontents%20%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fconfig)). Which solution do you prefer ?

---

The initial problem:

When I use the Contents menu of MenuBar plugin with French language, the menu item caption is "Contents" and I'd like to have it translated. 

Current situation:

* Default configuration of MenuBar uses TableOfContent tag
* TableOfContent tag tiddler caption field is defined to `{{$:/language/SideBar/Contents/Caption}}`
* In French this caption is translated to "Sommaire", which is correct
* The translation is used for the SideBar tab caption, but **not** for the MenuBar Contents menu item caption
* The MenuBar Contents menu item caption is hardcoded to "Contents" in `$:/plugins/tiddlywiki/menubar/items/contents` caption field

I wasn't satisfied with this situation so I initiated this PR, but my (now corrected) solution is quite involved for the benefits. And now I realize there's a quick fix for just this problem.

Quick fix:
* Simple: change `$:/plugins/tiddlywiki/menubar/items/contents` caption field to `{{$:/language/SideBar/Contents/Caption}}`
* Just works, on tiddlywiki.com and any complete language plugin, with a sane fallback of "Contents"
* But if one wants to change the menu caption, they must override a system tiddler
* The "Table of Contents" description, appearing in the menu items list of the MenuBar config tiddler, isn't translated (I can live with that, and fixing it would imply adding a new translation tiddler)

Full solution:
* Fixes the initial translation problem (falls back to the translation tiddler)
* Adds more flexibility by allowing full caption + description configuration (stored in the chosen tag tiddler) with configuration UI
* Puts more code than I'd like into the caption field of `$:/plugins/tiddlywiki/menubar/items/contents`

What solution do you prefer? I pushed a new commit with the full solution so you can test it with the Vercel build, see [here](https://tiddlywiki5-72ybt3s74-jermolene.vercel.app/#%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fitems%2Fcontents:%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fitems%2Fcontents%20%24%3A%2Fplugins%2Ftiddlywiki%2Fmenubar%2Fconfig).

Fred

plugins/tiddlywiki/menubar/config.tid
tw-FRed: Done in my next commit

plugins/tiddlywiki/menubar/config.tid
tw-FRed: Done in my next commit

PR #7605, opened 2023-07-13
Update Saving on Browser with TiddlyStow.tid

twMat:
Prevent CC linking in the GettingStarted card display


Jermolene:
Thanks @twMat

PR #7606, opened 2023-07-14
Allow attribute substitution to handle variables containing non-word characters

### LINKED ISSUE #7604: [BUG] Attribute substitution doesn't substitute variables whose names contain dashes
### Describe the bug

```
\define test-dash() dash
\procedure test() nodash

<$text text=`$(test-dash)$ - $(test)$`/>
```
renders as:

`$(test-dash)$ - nodash`

### Expected behavior

I expect it to render as 

`dash - nodash`

### To Reproduce

paste the above example into tiddlywiki.com

### Screenshots

_No response_

### TiddlyWiki Configuration

tiddlywiki.com 5.3.0

### Additional context

It seems to me the regexp here:
https://github.com/Jermolene/TiddlyWiki5/blob/f517497fe7cc2b9aec70e2a8390ebaf67fdee39f/core/modules/wiki.js#L1089

should be the same as the regexp here:
https://github.com/Jermolene/TiddlyWiki5/blob/f517497fe7cc2b9aec70e2a8390ebaf67fdee39f/core/modules/widgets/widget.js#L235

btheado:
Use the same regexp for attribute substitution (and the substitute filter operator) as is already used for textual substitution in macros. This change allows substitution to work for variable names containing spaces, dashes, and any character other than `$` and `)`.

Fixes #7604.

### DIFF
--- a/core/modules/wiki.js
+++ b/core/modules/wiki.js
@@ -1086,7 +1086,7 @@ exports.getSubstitutedText = function(text,widget,options) {
 		output = $tw.utils.replaceString(output,new RegExp("\\$" + $tw.utils.escapeRegExp(substitute.name) + "\\$","mg"),substitute.value);
 	});
 	// Substitute any variable references with their values
-	return output.replace(/\$\((\w+)\)\$/g, function(match,varname) {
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @btheado

PR #7609, opened 2023-07-15
Update substitute Operator.tid

pmario:
Improve wording for `$( filter expression )$` placeholder

### DIFF
--- a/editions/tw5.com/tiddlers/filters/substitute Operator.tid
+++ b/editions/tw5.com/tiddlers/filters/substitute Operator.tid
@@ -20,7 +20,7 @@ The <<.op substitute>> operator replaces any placeholders in the input titles in
 |placeholder syntax|description|h
 |`$n$`|Text substitution of a parameter provided to the operator, where n is the position of the parameter starting with 1 for the first parameter. Unmatched placeholders pass through unchanged.|
 |`$(varname)$`|Text substitution of a variable. Undefined variables are replaced with an empty string.|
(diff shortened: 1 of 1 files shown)


editions/tw5.com/tiddlers/filters/substitute Operator.tid
```diff
@@ -20,8 +20,8 @@ The <<.op substitute>> operator replaces any placeholders in the input titles in
 |placeholder syntax|description|h
 |`$n$`|Text substitution of a parameter provided to the operator, where n is the position of the parameter starting with 1 for the first parameter. Unmatched placeholders pass through unchanged.|
 |`$(varname)$`|Text substitution of a variable. Undefined variables are replaced with an empty string.|
-|`${ filter expression }$`|Text substitution with the first result of evaluating the filter expression.  |
+|`${ filter expression }$`|Text substitution with the first result of evaluating the filter expression. So if the expression returns multiple titles, only the first one can be used. |
```
saqimtiaz, line 23: I would suggest:
"If the filter expressions returns multiple titles only the first title is used".

Also note that the current wording is consistent with the tiddler "Filtered Attribute Values", see:
https://tiddlywiki.com/#Filtered%20Attribute%20Values

Jermolene:
Thanks @pmario I've pushed a fix with slightly updated wording. See 3a90c37816ff39104ee2851e1c6c1732c4528763

PR #7611, opened 2023-07-16
Trim whitespace when importing variables in $importvariables

saqimtiaz:
@Jermolene can you think of any undesirable consequences of trimming whitespace when parsing tiddlers to import variables?

It would make importing $set widgets much less error prone and would allow the following to be imported reliably:
```
<$set name="one" value="elephant">
<$set name="two" value="giraffe">
</$set>
</$set>

```


Jermolene:
Hi @saqimtiaz I think that makes sense, I can't see a problem at the moment.

Jermolene:
It would be worth adding a test along the lines of the one I added yesterday ff7214ff561d479e7563dce82a525004c556b615

saqimtiaz:
@Jermolene I am planning to both add a test for this specific use case but also some tests that cover importing a mixture of macro definitions and set widgets.

Jermolene:
Thanks @saqimtiaz

PR #7615, opened 2023-07-19
Remove full rebuilds for tag indexer with every keystroke or UI click

### LINKED ISSUE #7455: [Issue] Something strange is going on with the TagIndexer
https://github.com/Jermolene/TiddlyWiki5/blob/fb10b1ee9183f1b9c0e5dc05ef7b586f26bab403/core/modules/indexers/tag-indexer.js#L37-L41

- I did define a log-point at line 39 to log the variable `subIndexer`
- If I open a tiddler and switch to edit-mode I do get the log info 4 times for every subindex
   - shown in the screenshot
- `subindexer.update()` does this -> It destroys the index. 4 times ?!

https://github.com/Jermolene/TiddlyWiki5/blob/2e377b1f48ac012aca2cf71b2b433e72d3291242/core/modules/indexers/tag-indexer.js#L72C1-L74

So I think there is something strange going on. ... **Why does the `update()` function kill the index**? It should update the index. 

IMO we are wasting a lot of time here. 

image

A similar thing is going on with the `.rebuild()` function https://github.com/Jermolene/TiddlyWiki5/blob/2e377b1f48ac012aca2cf71b2b433e72d3291242/core/modules/indexers/tag-indexer.js#L56 ... 

If I change the Sidebar Tab, all the indices are rebuilt -> **Why?**

pmario:
This PR will fix #7455

As issue #7455 shows, the tag indexes get destroyed with every keystroke in edit mode and with every user UI interaction. IMO that should not be the case. 

This PR also contains the performance measurement logic, so Vercel builds a test-version. Once the preview-version is built I'll update the PR and remove that code. 

Since the updateDescriptor object is nested and ES5 needs a lot of boilerplate code to avoid "undefined" exceptions I did create several helper variables, which should make the code much more understandable. 

As the screenshots at the issue shows, the CSS "rebuilding" speed can be doubled.
I did some test with 2000 tags and the relative performance improvement is the same.


pmario:
The following text is copied over from #7455 

--------

I did some initial test with this simple logic, which tests if the `tags` field exists. No further processing. tiddlywiki.com has 388 tags atm. ... So I would need to test it with more tags 2000 or more. 

<details><summary>show code <--- </summary>

```
TagIndexer.prototype.update = function(updateDescriptor) {
	if( (!updateDescriptor.new.tiddler?.fields?.tags && updateDescriptor.old.tiddler === undefined) ||
		(!updateDescriptor.old.tiddler?.fields?.tags && updateDescriptor.new.tiddler === undefined) ||
		(!updateDescriptor.old.tiddler?.fields?.tags && !updateDescriptor.new.tiddler?.fields?.tags)) {
			return;
		} else {
			$tw.utils.each(this.subIndexers,function(subIndexer) {
				subIndexer.update(updateDescriptor);
			});
		}
};
```
</details>

Then I did measure, how long a rebuild takes. 

<details><summary>show code <--- </summary>

```
TagSubIndexer.prototype.rebuild = function() {
	var self = this;
	// Hashmap by tag of array of {isSorted:, titles:[]}
const t0 = $tw.utils.timer();
	this.index = Object.create(null);
	// Add all the tags
	this.indexer.wiki[this.iteratorMethod](function(tiddler,title) {
		$tw.utils.each(tiddler.fields.tags,function(tag) {
			if(!self.index[tag]) {
				self.index[tag] = {isSorted: false, titles: [title]};
			} else {
				self.index[tag].titles.push(title);
			}
		});
	});
console.log("--------------- rebuild", $tw.utils.timer(t0));
};
```
</details>

The TW startup shows this. Every rebuild on my machine needs between 1ms and 3ms (worst case)

image

That's not as much as I thought, but every "styleRefresh" needs about 2.5ms now. **That is about 2ms better with basically every click on the UI.** 

image

v5.2.7 code needs about double the time. 

image

-----

My machine is an  **Intel(R) Core(TM) i9-9900K CPU @ 3.60GHz   3.60 GHz** ... Not the newest machine but probably 10 times faster than an average mobile device.

linonetwo:
Great, in most of the times, `there is no new tiddler` will be the case.

core/modules/indexers/tag-indexer.js
```diff
@@ -35,9 +35,27 @@ TagIndexer.prototype.rebuild = function() {
 };
 
 TagIndexer.prototype.update = function(updateDescriptor) {
-	$tw.utils.each(this.subIndexers,function(subIndexer) {
-		subIndexer.update(updateDescriptor);
-	});
+	var newTid = updateDescriptor.new.tiddler;
+	var oldTid = updateDescriptor.old.tiddler;
+	var noOldTid = oldTid === undefined;
+	var noNewTid = newTid === undefined;
```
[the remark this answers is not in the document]
pmario: We cannot return here. The if-clause is needed to know if we should reset the indexers. The vars a, b and c are there for increased readability. I don not want to pre-mature optimize them away. It does not matter.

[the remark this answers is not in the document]
linonetwo: But you have noNewTid in the if below, and is in "or" statement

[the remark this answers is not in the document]
linonetwo: > The vars a, b and c are there for increased readability

Put them in separate if statement with comment does the same

pmario:
>Great, in most of the times, there is no new tiddler will be the case.

I see what you mean. I'll ad an extra `if (noNewTid) return` after the first variable definitions to avoid the a,b,c evaluation

pmario:
@linonetwo .. I did implement the "return early" logic and it works for me. Can you test and also make some measurements. If it makes a difference for your projects?

The code I used for measurements is documented at: https://github.com/Jermolene/TiddlyWiki5/pull/7615#issuecomment-1642078800 behind the "show code buttons"

core/modules/indexers/tag-indexer.js
```diff
@@ -35,9 +35,27 @@ TagIndexer.prototype.rebuild = function() {
...
+
+	// Index should be kept when there is no new tiddler
+	if (noNewTid) {
+		return;	// early
+	} //  -> or
+	// a) new tiddler has no tags && no old tiddler -> or
+	// b) new tiddler has no tags && old tiddler has no tags
+	var a=((newTid && newTid.fields && newTid.fields.tags === undefined) && noOldTid),
```
linonetwo, line 49: You can return early whenever possible. Here just 

```js
// a) new tiddler has no tags && no old tiddler -> or
if ((newTid && newTid.fields && newTid.fields.tags === undefined) && noOldTid) return;
// b) new tiddler has no tags && old tiddler has no tags
if ((newTid && newTid.fields && newTid.fields.tags === undefined) && (oldTid && oldTid.fields && oldTid.fields.tags === undefined) return;

File: core/modules/indexers/tag-indexer.js [hunk 3]
```diff
@@ -35,9 +35,27 @@ TagIndexer.prototype.rebuild = function() {
...
+		b=((newTid && newTid.fields && newTid.fields.tags === undefined) && (oldTid && oldTid.fields && oldTid.fields.tags === undefined));
+
+	if( a || b ) {
+			return;
+		} else {
```
linonetwo, line 54: https://eslint.org/docs/latest/rules/no-else-return

I thought we already have setup eslint? Did you get the warning?

linonetwo:
I don't know where I can test it better than you, I think code is good, I can put the prerelease into TidGi after this is merged.

core/modules/indexers/tag-indexer.js
pmario: These micro optimisations do not make sense. I did add this code to improve readability and make it easier to understand the logic.

core/modules/indexers/tag-indexer.js
pmario: I do not get any warning. The eslint configuration we discussed for pretty printing is not active yet.

Jermolene:
Hi @pmario looking at this again, I think the logic is faulty. The function bails early if it receives a deletion, which seems wrong. Let's look at this again after v5.3.2

pmario:
converted this PR into a draft, since it contains `console.log()` code, which should not be part of the PR

pmario:
The ZIP file contains 2 single-file wikis. 

- `tag-index.html` contains the "old code" + console.log() - which updated the tag-caches with every keystroke.
- `tag-optimized.html` contains the "new code" + console.log() - but it avoids tag-cache updates for most UI interactions. 

- both wikis contain 20000 new tiddlers with 20000 new tags
  - **WARNING** If you open the Right Sidebar -> More -> Tags tab it may brick your browser. 
  - It needs about 12 seconds on my machine. 

[tag-indexer-performance-measurements.zip](https://github.com/Jermolene/TiddlyWiki5/files/13476638/tag-indexer-performance-measurements.zip)

The best way to see the differences is, to open the browser dev-tools with F12 and interact with the UI. 

- tag-index.html will activate a lot more cache-updates
- styleRefresh will need about 90% more time with every interaction
- At browser reload tag-index.html on my machine needs about 80ms more time because of unnecessary cache refreshes

PR #7619, opened 2023-07-20
Fix tiddler icon size

Jermolene:
As reported [in the forum](https://talk.tiddlywiki.org/t/tiddler-icon-in-viewtemplate-has-fixed-size-not-relative-to-text-size/7612):

> This observation is nitpicky and of low priority if considered a bug at all, but looks like an overlook, so I decided to ask.

> The icon displayed in front of the tiddler title in the ViewTemplate has fixed, not relative, size. It stays the same size regardless of the title font size.

> This is not visible at all on the default font size `$:/themes/tiddlywiki/vanilla/metrics/fontsize` of 14px, because the fixed size happens to be the same or very similar to the view toolbar icons. It is also not very apparent on small changes of the font size. Bigger changes of the font size, as in screenshots below, demonstrate the problem:

### DIFF
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1173,8 +1173,8 @@ button.tc-btn-invisible.tc-remove-tag-button {
 }
 
 .tc-tiddler-title-icon svg {
-	width: 0.75em;
-	height: 0.75em;
+	width: 0.9em;
+	height: 0.9em;
 }
 
 .tc-system-title-prefix {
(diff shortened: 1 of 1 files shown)


pmario:
IMO that's not a fix. It breaks the existing default behaviour. 

See: tiddlywiki.com -- HelloThere  -- 100% browser zoom

image


Se preview -- HelloThere -- 100% browser zoom 

image

Jermolene:
Thanks @pmario fix upcoming. We seem to need 0.9em to match the existing size

pmario:
>Thanks @pmario fix upcoming. We seem to need 0.9em to match the existing size

Yea, that's the problem with `em` CSS scaling if there are several elements that use em it gets unpredictable values. 
That's why everyone prefers to use `rem` for new CSS projects now, because it always starts at the initial value.

Issue #7440, opened 2023-05-08
[BUG] Drag and drop from Google Chrome to Firefox fails


pearson:
### Describe the bug

Dragging and dropping a plugin link from Google Chrome to Firefox on a Windows 11 PC fails to correctly import the plugin.

This problem only occurs when dragging and dropping a plugin from Google Chrome to a wiki in Firefox. It does not happen when dragging and dropping a plugin "in the opposite direction" from Firefox to a wiki Google Chrome, or when dragging and dropping between two windows of the same browser.

Edit: As might be expected, this problem also happens when dragging and dropping a plugin from Microsoft's Edge to a wiki in Firefox. It also works correctly when dragging and dropping a plugin from Firefox to a wiki in Edge.

### Expected behavior

I would expect the plugin to be correctly imported.

### To Reproduce

1. In Google Chrome, open a plugin page such as [Section Editor](https://kookma.github.io/TW-Section/#%24%3A%2Fplugins%2Fkookma%2Fsection)
2. In Firefox, open a Tiddlywiki wiki, such as `empty.html` from [this link](https://tiddlywiki.com/#GettingStarted)
3. Drag the plugin link from the Google Chrome page to the wiki in Firefox, and drop the link once the "Drop now" banner appears
4. Notice that the import title is "Untitled" and not "$:/plugins/kookma/section"
5. Click the down arrow next to "Untitled" and notice that there is no code shown: ![chrome to firefox drag and drop example - with down arrow clicked](https://user-images.githubusercontent.com/656880/236949683-1deda94f-0977-44e8-a3ed-e913b58578db.png)
6. Click "Import" and notice that no plugin was imported. (Check Control Panel -> Settings -> Plugins, for example.)

### Screenshots

The expected result of this drag and drop is shown in this screenshot:

![firefox to firefox drag and drop example - with down arrow clicked](https://user-images.githubusercontent.com/656880/236949800-2590d52c-1d7e-43af-bbe0-a32895216c08.png)

### TiddlyWiki Configuration

- The default `empty.html` v5.2.7
- Saving mechanism: nothing changed from the default `empty.html`
- Plugins installed: nothing changed from the default `empty.html`

### Desktop (please complete the following information):

- OS: Windows 11
- Google Chrome v112.0.5615.138 64-bit 
- Firefox v112.0.2 64-bit

Jermolene:
[...]

Issue #7305, opened 2023-02-28
[BUG] erratic drag and drop issue in Safari on OSX


saqimtiaz:
### Describe the bug

With some combinations of source and target wiki, dragging and dropping a tiddler link from one wiki to another does not work as expected, resulting in an Untitled tiddler where the content of the tiddler is the title of the dragged tiddler.

I would appreciate any assistance in understanding why this bug only shows up in certain scenarios to allow for further debugging. The issue seems to predate the v5.2.5 release.

### To Reproduce

1. In separate windows in Safari open tiddlywiki.com and tiddlywiki.org
2. Drag the link to the tiddler HelloThere from tiddlywiki.org to tiddlywiki.com
3. Note how the $:/Import tiddler does not contain the correct data.

Note that:
* Using tiddlywiki.com/prerelease or tiddlywiki.com/prerelease/empty.html instead of tiddlywiki.com results in the same error as long as you drag from tiddlywiki.org
* Dragging from paste.tiddlyshost.com to tiddlywiki.com/prerelease exhibits the same issue, so the issue is not isolated to tiddlywiki.org
* Dragging from tiddlywiki.com to tiddlywiki.com/prerelease works as expected

Under the covers it seems like that in the instances when the drag and drop fails to work as expected, the `event.dataTransfer `object only contains the type `text/plain`. What is perplexing is why the types in the event vary even between different wiki instances that are the same TiddlyWiki version.

### TiddlyWiki Configuration

- Version [e.g. v5.1.24]
- Saving mechanism [e.g. Node.js, TiddlyDesktop, TiddlyHost etc]
- Plugins installed [e.g. Freelinks, TiddlyMap]

saqimtiaz:
The simplest way to reproduce the issue is to download https://tiddlywiki.com/empty.html and open it in two different Safari windows, create a tiddler in one window and try to drag it to the other.

Also very confounding is that on the last master, if I run the tw5-com edition on the node.js server and open it in two Safari windows, it works just fine to drag and drop between them. If I save a snapshot of the wiki and open that in two Safari windows, the problem surfaces.

Jermolene:
Thank you @saqimtiaz. I've spent some time on this preparatory to the v5.2.6 release, and I'm just as stumped.

[...]

PR #7622, opened 2023-07-20
fix drag and drop from chrome-like browsers to FireFox

pmario:
I did test it with the code @saqimtiaz [provided ](https://github.com/Jermolene/TiddlyWiki5/pull/7445#issuecomment-1576149933)for testing. 

```
\define droppable-actions()
<$action-setfield $tiddler="temp/droptest" text=<<actionTiddler>>/>
\end


<$droppable tag="div" actions=<<droppable-actions>> >
<div style="height:100px;width:200px;border: 1px solid orange;">

drop here

The dropped tiddler is: {{temp/droptest}}

</div>
</$droppable>

<$draggable tag="div" tiddler="Hello There">
<p style="height:50px;width:400px;border: 1px solid red; line-height:50px;">
Drag me
</p>
</$draggable>
```


pmario:
uups OP was created too early. 

I did test the PR with FireFox, Edge on Windows.

FF, Edge and Chrome on Android. 

All of them work now. 

This PR fixes: https://github.com/Jermolene/TiddlyWiki5/issues/7440 

@saqimtiaz Can you test it with your device?
@Jermolene .. Can you test it with Safari and iDevices?

saqimtiaz:
@pmario I will endeavour to test later today.

I have strong reservations about using browser sniffing but also do not see an alternative. On a personal note, the broken drag and drop in mobile Chrome renders two different TW editions that I maintain for a small group of users next to useless unless I override the core drag and drop code.

pmario:
>I have strong reservations about using browser sniffing but also do not see an alternative. 

Me too as I did comment at https://github.com/Jermolene/TiddlyWiki5/issues/7440#issuecomment-1642154461

We do heavily rely on browser sniffing as the $tw.browser.XXX functions show. IMO sometimes there is no other way.

From my point of view, keeping the `.isIE` code paths in tact, but not allowing ourself to fix new problems would be a poor decision. 

>On a personal note, the broken drag and drop in mobile Chrome renders two different TW editions that I maintain for a small group of users next to useless unless I override the core drag and drop code.

My question is. Does this PR fix the problem?

I know it's not perfect, but [it seem to be better](https://talk.tiddlywiki.org/t/change-in-drag-and-drop-between-chrome-and-firefox/4410/14?u=pmario) than the broken state we do have atm. 

I heavily rely on the possibility to drag pure wikitext-plugins to the browser bookmarks toolbar and then import them into other wikis from there. This PR will make that possible again.

core/modules/startup/startup.js
```diff
@@ -30,7 +30,7 @@ exports.startup = function() {
 		// 2023-07-21 Edge returns UA below. So we use "isChromeLike"
 		//'mozilla/5.0 (windows nt 10.0; win64; x64) applewebkit/537.36 (khtml, like gecko) chrome/114.0.0.0 safari/537.36 edg/114.0.1823.82'
 		$tw.browser.isChromeLike = navigator.userAgent.toLowerCase().indexOf("chrome") > -1;
-		$tw.browser.hasTouch = window.matchMedia("(pointer: coarse)").matches;
+		$tw.browser.hasTouch = !!window.matchMedia && window.matchMedia("(pointer: coarse)").matches;
```
saqimtiaz, line 33: Recommend refactoring the check for Chrome and Touch as utility functions.

saqimtiaz:
> My question is. Does this PR fix that problem? IMO it should.

With the vercel build from this PR I can confirm that drag and drop on mobile Chrome is functional.

core/modules/startup/startup.js
pmario: I did implement them as 3 new $tw.utils.browser.XXX functions. So it's easy to modify them in a plugin one by one if needed.

pmario:
Please test again with the new "utils" functions

core/modules/utils/dom/sniff-chrome-touch.js
```diff
@@ -0,0 +1,38 @@
+/*\
+title: $:/core/modules/utils/dom/sniff-chrome-touch.js
+type: application/javascript
+module-type: utils
+
+Utility functions to detect chrome-like browsers and touch input
```
Jermolene, line 6: We already have $tw.browser.* flags being applied in `$:/core/modules/startup.js`, I don't like the idea of a separate set of methods elsewhere.

I'd prefer to extend the definitions in `$:/core/modules/startup.js` to include these three new ones. It would be nice to move the $tw.browser.* stuff out into a specific module type but we can leave that for another time.

Jermolene (commented):
Thanks @pmario just a minor point

core/modules/utils/dom/sniff-chrome-touch.js
pmario: OK I did update it as requested.

pmario:
@Jermolene Can you test it with Safari on mobile devices. The test code is in the OP.

Jermolene:
> @Jermolene Can you test it with Safari on mobile devices. The test code is in the OP.

Hi @pmario I've tested it on Safari 16.5.2 on Ventura 13.4.1 and iPadOS 17 beta and the test works properly in both browsers.

pmario:
>Hi @pmario I've tested it on Safari 16.5.2 on Ventura 13.4.1 and iPadOS 17 beta and the test works properly in both browsers.

That's good. Can you also test a drag & drop import of a plugin from Safari to FireFox or Chrome and Chrome to Safari or FF?

Jermolene:
Hi @pmario I tried the following combinations for dragging a plugin

* Safari to Firefox - fails
* Firefox to Safari - succeeds
* Safari to Chrome - fails
* Chrome to Safari - succceeds

pmario:
hmmm, That's strange .. What happens if you remove the `if(!($tw.browser.isMobileChrome)) {`

It would be the same code that we had for TW v5.2.2 .. So can you test again with

- with: https://tiddlywiki.com/archive/full/TiddlyWiki-5.2.2.html 
- and the `if(!($tw.browser.isMobileChrome)) {` line removed from [core/modules/utils/dom/dragndrop.js](https://github.com/Jermolene/TiddlyWiki5/pull/7622/files#diff-8561008eab6f71c9fcce2eeed6acd4c7a63c08fd42ee9f7185bee2172c7e083a)

Jermolene:
Hi @pmario would you be able to create or update a PR for the test, just to make sure that we're running the same code?

pmario:
OK . .. In the mean while please  test v5.2.2 from the archive

Jermolene:
Hi @pmario exactly the same results for v5.2.2:

* Safari to Firefox - fails
* Firefox to Safari - succeeds
* Safari to Chrome - fails
* Chrome to Safari - succceeds

pmario:
Did you test with a laptop or on iPad / iPhone too?

pmario:
Since v5.2.2 and v5.3.0 are the same, I do expect, that this PR does not influence the behaviour. It probably has always been that way. So there seems to be a new difference between the `dataTransfer` object that Safari creates and the other browsers. 

Just for curiosity. Does Safari to Safari work?

pmario:
Does Safari have something like a "Bookmarklet bar", where you can drag & drop a plugin tiddler to the bookmarklet bar. 

Then importing it into other wikis dragging it from the bar to the wiki?

Jermolene:
> Did you test with a laptop or on iPad / iPhone too?

I could not repeat the tests for dragging to/from Chrome and Firefox, because there are no other browsers on iPad/iPhone.

> Since v5.2.2 and v5.3.0 are the same, I do expect, that this PR does not influence the behaviour. It probably has always been that way. So there seems to be a new difference between the `dataTransfer` object that Safari creates and the other browsers.
> 
> Just for curiosity. Does Safari to Safari work?

Yes, it works on v5.2.2, or did you mean v5.3.0?

> Does Safari have something like a "Bookmarklet bar", where you can drag & drop a plugin tiddler to the bookmarklet bar.

Yes, Safari supports bookmarklets on all platforms.

pmario:
>Yes, Safari supports bookmarklets on all platforms.

So creating bookmarklets from plugins should be as easy as dragging a plugin-tiddler to the bookmarklets bar and then from there to a new wiki. 

Can you test creating bookmarklets with Safari only with this PR?
Can you also test it with with Safari only and v5.1.23?

This behaviour worked till v5.1.23 and should work again with this PR. At least it does for me. 
(I just found out that it was broken with v5.2.0 a view minutes ago while preparing this post)

I'm not sure what happened to drag & drop with the update v5.1.23 to v5.2.0 ... Would need to investigate.

Jermolene:
> So creating bookmarklets from plugins should be as easy as dragging a plugin-tiddler to the bookmarklets bar and then from there to a new wiki.

Are you suggesting that we change the recommended way to install plugins?

> Can you test creating bookmarklets with Safari only with this PR? Can you also test it with with Safari only and v5.1.23?
> 
> This behaviour worked till v5.1.23 and should work again with this PR. At least it does for me. (I just found out that it was broken with v5.2.0 a view minutes ago while preparing this post)
> 
> I'm not sure what happened to drag & drop with the update v5.1.23 to v5.2.0 ... Would need to investigate.

I can't do further tests at the moment, but will return to this when I can. It would help enormously if you could prepare the builds that need testing.

pmario:
>Are you suggesting that we change the recommended way to install plugins?

No --- I just want that function back. 

It's an extremely convenient way to install wikitex-only plugins into any other wiki that I open. Eg: I can reload tiddlywiki.com and then drag my [link-to-tabs plugin](https://wikilabs.github.io/editions/link-to-tabs/) into the wiki. It will be installed and works.

pmario:
>I can't do further tests at the moment, but will return to this when I can. It would help enormously if you could prepare the builds that need testing.

OK

I use the archive to run tests with older TW versions eg: 

- https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.23
- https://tiddlywiki.com/archive/full/TiddlyWiki-5.2.0
- https://tiddlywiki.com/archive/full/TiddlyWiki-5.2.2

Jermolene:
Thanks @pmario just to clarify, you'd like to test dragging between Safari to and from Firefox and Chrome across all three of those versions?

pmario:
On Windows we know, that we lost the ability to drag plugins **from** Chrome / Edge **to** FireFox with TW **v5.2.3**
The other way around seems to always work. 

In [the post where you tested with v5.2.2](https://github.com/Jermolene/TiddlyWiki5/pull/7622#issuecomment-1646559733) you wrote, that 

>Hi @pmario exactly the same results for v5.2.2:
>
>- Safari to Firefox - fails
>- Firefox to Safari - succeeds
>- Safari to Chrome - fails
>- Chrome to Safari - succeeds

So it seems Safari lost the drag & drop plugin function earlier. Since everything is fine with v5.2.2 on Windows. 

**So I think we first need:** 

- Chrome TW v5.2.**2** to FireFox ... that's missing from earlier tests
- Chrome TW v5.2.**3** to FireFox ... that's missing from earlier tests

  ---

- Safari v5.1.23 to Firefox v5.3.0
- Safari v5.1.23 to Chrome v5.3.0
- Chrome v5.1.23 to FireFox v5.3.0

If they succeed then we need to test 

- Safari v5.2.0 to Firefox v5.3.0
- Safari v5.2.0 to Chrome v5.3.0
- Chrome v5.2.0 to FireFox v5.3.0

-----

Just to be sure: You did your very first test https://github.com/Jermolene/TiddlyWiki5/pull/7622#issuecomment-1646555508 with [Vercel-Preview v5.3.1 from this PR](https://tiddlywiki5-git-fork-pmario-fix-dnd-for-mobile-touch-jermolene.vercel.app/). Right?

Jermolene:
> On Windows we know, that we lost the ability to drag plugins **from** Chrome / Edge **to** FireFox with TW **v5.2.3** The other way around seems to always work.

Is the problem restricted to plugins, or does it apply to any tiddler?

> So it seems Safari lost the drag & drop plugin function earlier. Since everything is fine with v5.2.2 on Windows.

Yes that makes sense.

> Just to be sure: You did your very first test [#7622 (comment)](https://github.com/Jermolene/TiddlyWiki5/pull/7622#issuecomment-1646555508) with [Vercel-Preview v5.3.1 from this PR](https://tiddlywiki5-git-fork-pmario-fix-dnd-for-mobile-touch-jermolene.vercel.app/). Right?

Yes, that's correct I used the Vercel preview for this PR as it was at the time.

pmario:
>Is the problem restricted to plugins, or does it apply to any tiddler?

It's every tiddler starting with v5.2.3 that will be shown as Untitled in the $:/Import dialogue. 

**This PR fixes the problem for Windows and Edge** 

We already know which commit causes the problem starting with v5.2.3: https://github.com/Jermolene/TiddlyWiki5/pull/6789

This PR fixes the "plugin-problem" and preserves the Mobile Chrome drag and drop functionality that Saq needs [and confirmed, that it works for him.](https://github.com/Jermolene/TiddlyWiki5/pull/7622#issuecomment-1645273381)

--------

**Conclusion** 

- We could merge this PR to fix 2 major problems that we did know, prior to today. 

**Suggestion**

- We could treat the Safari drag & drop problem as a completely new one, since we did not know it till today
- We also can treat the bookmarklet-problem as a new problem
  - I thought it is directly related to v5.2.3, but it seems it's related to an earlier change from v5.2.0
  - I did find out a simple workaround in the meantime to fix the bookmark-bar problem for FireFox
  - There still is something strange going on, but it needs more investigation

pmario:
At Talk Tony seems to be happy too: https://talk.tiddlywiki.org/t/change-in-drag-and-drop-between-chrome-and-firefox/4410/18?u=pmario

Jermolene:
> This PR fixes the "plugin-problem" and preserves the Mobile Chrome drag and drop functionality that Saq needs [and confirmed, that it works for him.](https://github.com/Jermolene/TiddlyWiki5/pull/7622#issuecomment-1645273381)

Why do you call it a "plugin-problem"? Is that just because installing plugins is the most common usage of drag and drop?

> * We could merge this PR to fix 2 major problems that we did know, prior to today.

Yes I agree.

> * We could treat the Safari drag & drop problem as a completely new one, since we did not know it till today

Agreed.

> * We also can treat the bookmarklet-problem as a new problem

What is the bookmarklet problem?

pmario:
>What is the bookmarklet problem?

Super strange one too. I thought is was directly related to the change in v5.2.3 but today I did learn that it started with the switch to v5.2.0.

To replicate on Windows and FireFox

- Use [v5.1.23 Controlpanel](https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.23.html#%24%3A%2FControlPanel:%24%3A%2FControlPanel%20HelloThere%20GettingStarted%20Community)
- Open the Plugins tab
- drag the Browser Sniff plugin onto the browser bookmarks bar
- it will show up like this: 

image

- Now drag the bookmarklet back to the wiki

image

Everything is fine. 

--------

- Now use https://tiddlywiki.com/archive/full/TiddlyWiki-5.2.0.html
- to create a new bookmarklet
- drag the new one back to the wiki -> RSOD
   - The RSOD goes away with v5.2.2 but shows Untitled in the $:/Import dialogue
- Drag the v5.1.23 one back -> OK

So this tells me, that there is something different between the bookmarklet encode between 5.1.23 and 5.2.0

**Now the funny thing / workaround**

- rename the offending bookmarklet from 5.2.0 to eg: **test**
- drag it to any tw version
- it works now.

Hope that makes sense.

Jermolene:
Hi @pmario I'm not sure that I do understand; the TW versions you mentioned were all older versions. Is something broken in the current master?

pmario:
Uuups. Sorry. **Yes the current master has the same problem** that started with v5.2.0. 

I did describe the tests to find out where to start searching.

pmario:
In Feb. Saq brought up  **[BUG] erratic drag and drop issue in Safari on OSX #7305**, which imo is related to the Safari d&d problem we discussed here today

Issue #7625, opened 2023-07-21
[BUG] removing tc-tiny-gap-left from the toc-caption element wikitext broke the whole TOC structure visually


pmario:
### Describe the bug

The following PR broke the TOC structure spacing completely: **small gap fix for the updated toc macro (alternative to tc-tiny-gap-left)** #7219 

Removing the `tc-tiny-gap-left` removes the gap between the chevrons and the text and also messes with the spacing of the different nested elements 

The CSS added to vanilla does nothing, since it's not applied to the DOM elements that are effected.

**So from my point of view the PR should be reverted** because [the requested change in line 148](https://github.com/Jermolene/TiddlyWiki5/pull/7121#issuecomment-1354048662) from Alzacon was not implemented anyway. 

>Alzacon wrote
>
>It is great idea. Could it be used in the line 148 too?

The requested change was not implemented, but the tc-tiny-gap-left was removed and the CSS added, which broke the toc. 

### Expected behavior

IMO the PR can be reverted without any side effects. 

With the PR reverted a custom TOC using captions and titles intermixed **should look like this:** 

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/7420a2a7-900f-4614-b5b7-0fe2d433b1d3)

BUT at the moment it looks like this

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/4e8ec6a5-a0fc-402e-b05c-31545360789b)

### TiddlyWiki Configuration

v5.3.0

pmario:
tiddlywiki.com is broken too. Open HelloThere and see the indentation of Examples and also see the space between the chevrons and the text

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/fbd236d6-c9f5-4e5b-85e5-de9341846e32)

pmario:
I did have a look for a CSS rule that seems to work, but not sure if there other side effects

```
.toc-item .tc-toc-caption {
  margin-left: .25em;
}
```

pmario:
I did even more testing and id apply the requested change to line 148 .. the empty message. 

It turns out this PR is not needed. There would have been no visual change with line 148 modified. As far as I can see atm. 

So forget about the CSS from the post above and let us revert the PR. It should be possible to modify line 148 with no visual change at all.

Jermolene:
Hi @pmario I agree, now that you've pointed it out, the TOC formatting is definitely broken. Would you be able to prepare a PR?

[...]

PR #7627, opened 2023-07-21
fix toc indentation problem

pmario:
This PR **fixes** issue:  *[BUG] removing tc-tiny-gap-left from the toc-caption element wikitext broke the whole TOC structure visually #7625*

- It does **revert** the CSS definitions from #7219, since they did not apply.
- It does **re-add** the `tc-tiny-gap-left` utility class to the `toc-caption` macro
- It adds the change to line 148, requested by [Alzacon](https://github.com/Jermolene/TiddlyWiki5/pull/7121#issuecomment-1354048662) and was [approved by Jeremy](https://github.com/Jermolene/TiddlyWiki5/pull/7121#issuecomment-1356829600)
  - It adds a `toc-link: no` field to the "SecondThreeThree" doc tiddler, so the code in line 148 will be active and can be tested
  - It adds the CSS definition needed for changes in line 148
- It adds captions to "Third" and "Fourth" tiddler so we can see caption handling

The TOC-macro example tiddler can be used to test all toc code-paths. It should look like this now: 

image

- The tiddlywiki-com table of content indentation works again. 
- Indentation alignment is OK now. 

image

### DIFF
--- a/core/wiki/macros/toc.tid
+++ b/core/wiki/macros/toc.tid
@@ -7,7 +7,7 @@ tags: $:/tags/Macro
 
 \define toc-caption()
 \whitespace trim
-<span class="tc-toc-caption">
+<span class="tc-toc-caption tc-tiny-gap-left">
 <$set name="tv-wikilinks" value="no">
   <$transclude field="caption">
     <$view field="title"/>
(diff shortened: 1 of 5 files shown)


Jermolene (commented):
Thanks @pmario please could you remove the modifications to the modified fields? I don't think these changes should cause them to rise to the top of the recent changes list.

pmario:
@Jermolene .. I did set the modified fields back to the old date

Jermolene:
> @Jermolene .. I did set the modified fields back to the old date

Thanks @pmario

PR #7634, opened 2023-07-23
Add usemap attribute to image widget

pille1842:
Allow for a usemap attribute on the $image macro call which is passed through to the resulting img tag. This makes the use of HTML image maps [1] possible.

[1]: <https://www.w3schools.com/html/html_images_imagemap.asp>

### DIFF
--- a/core/modules/widgets/image.js
+++ b/core/modules/widgets/image.js
@@ -99,6 +99,9 @@ ImageWidget.prototype.render = function(parent,nextSibling) {
 	if(this.imageClass) {
 		domNode.setAttribute("class",this.imageClass);
 	}
+	if(this.imageUsemap) {
+	    	domNode.setAttribute("usemap",this.imageUsemap);
+	}
 	if(this.imageWidth) {
 		domNode.setAttribute("width",this.imageWidth);
 	}
@@ -138,6 +141,7 @@ ImageWidget.prototype.execute = function() {
 	this.imageWidth = this.getAttribute("width");
 	this.imageHeight = this.getAttribute("height");
 	this.imageClass = this.getAttribute("class");
+    	this.imageUsemap = this.getAttribute("usemap");
 	this.imageTooltip = this.getAttribute("tooltip");
 	this.imageAlt = this.getAttribute("alt");
 	this.lazyLoading = this.getAttribute("loading");
@@ -148,7 +152,7 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 */
 ImageWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
-	if(changedAttributes.source || changedAttributes.width || changedAttributes.height || changedAttributes["class"] || changedAttributes.tooltip || changedTiddlers[this.imageSource]) {
+	if(changedAttributes.source || changedAttributes.width || changedAttributes.height || changedAttributes["class"] || changedAttributes.usemap || changedAttributes.tooltip || changedTiddlers[this.imageSource]) {
 		this.refreshSelf();
 		return true;
 	} else {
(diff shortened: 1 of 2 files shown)


Jermolene:
Thank you @pille1842 that's great. A change like this would also require a small documentation update. The new attribute should be marked with the `<<.from-version>>` macro referencing the next version to be released (currently 5.3.1); see the [button widget docs](https://github.com/Jermolene/TiddlyWiki5/blob/d0da1ef9d9274b96ec514bd3e2b939653a2508ed/editions/tw5.com/tiddlers/widgets/ButtonWidget.tid#L40) for an example.

pille1842:
Hi @Jermolene, thanks for your kind feedback. I added another commit documenting the change in the ImageWidget tiddler.

Jermolene:
Thank you @pille1842. We're now in the freeze period before the release of v5.3.1, so I've added this PR to the list for consideration afterwards: #7659

Jermolene:
Thanks @pille1842

Jermolene:
And thank you for your patience!

AnthonyMuscio:
I have come here from the pre-release. There is no information as to what the value is to be given to the usemap and what form the content can be;

A quick google would suggest;
```
<map name="workmap">
  <area shape="rect" coords="34,44,270,350" alt="Computer" href="computer.htm">
  <area shape="rect" coords="290,172,333,250" alt="Phone" href="phone.htm">
  <area shape="circle" coords="337,300,44" alt="Cup of coffee" href="coffee.htm">
</map>
```

If this is the case, or it varies I believe it is the first time something such has this has being used to reference an otherwise only html section so some simple documentation is needed to clarify without having to document the use of image maps in general, just how you can use them here.
* Although a edit tool to generate them would be nice.

AnthonyMuscio:
Please since this is in the prerelease now can we have even minimal documentation on what you need to do you define the map and use it within tiddlywiki.

Jermolene:
Hi @AnthonyMuscio I don't think the `usemap` attribute is particularly generally useful. It is an ancient HTML feature for overlaying arbitrarily shaped clickable links on top of an image. There's some examples in the [w3schools docs](https://www.w3schools.com/tags/att_usemap.asp). 

Nowadays it is often easier to make an SVG with clickable areas. For example see https://manuals.annafreud.org/ambit/#AMBIT%20Wheel

AnthonyMuscio:
@Jermolene I do know what it is and have more experience with that than svg however that is not the point. 

We just need to show the bit googling will not show, how you reference the usemap and what form it takes in tiddlywiki. 

For example, Is the usemap parameter a reference to a set of html area statements, or to a set wrapped in a usemap tag?

Jermolene:
> We just need to show the bit googling will not show, how you reference the usemap and what form it takes in tiddlywiki.

It is used within TiddlyWiki in the same way as in ordinary HTML, because TiddlyWiki doesn't apply any special processing here.

> For example, Is the usemap parameter a reference to a set of html area statements, or to a set wrapped in a usemap tag?

As the docs I linked to shows, the `usemap` attribute is `#` plus the ID of the associated `<map>` element.

AnthonyMuscio:
I will try and use it given this sparse information but a simple example would be helpful

For example from [here](https://www.w3schools.com/html/html_images_imagemap.asp) I tried;

```
<$image source="workplace.jpg" usemap="#workmap">

<map name="workmap" id=workmap>
  <area shape="rect" coords="34,44,270,350" alt="Computer" href="https://tiddlywiki.com/#ViewWidget">
  <area shape="rect" coords="290,172,333,250" alt="Phone" href="phone.htm">
  <area shape="circle" coords="337,300,44" alt="Coffee" href="coffee.htm">
</map>
```

I added the id in keeping with your instructions, but it does not work.

Jermolene:
Hi @AnthonyMuscio the issue here is that TiddlyWiki doesn't like those unterminated HTML tags; it parses the "map" element as being a child of the `<$image>` widget. Explicitly closing the tags makes it work:

```
<$image source="Blurry Lawn.jpg" usemap="#workmap"/>

<map name="workmap" id="workmap">
  <area shape="rect" coords="34,44,270,350" alt="Computer" href="https://tiddlywiki.com/#ViewWidget"/>
  <area shape="rect" coords="290,172,333,250" alt="Phone" href="phone.htm"/>
  <area shape="circle" coords="337,300,44" alt="Coffee" href="coffee.htm"/>
</map>
```

PR #7637, opened 2023-07-24
Fixed typos in documentation for new pragma

lilscribby:
Fixed the following two typos:
"expect" instead of "except" in \rules definition
"\define" instead of "\procedure" in procedure one-line format definition


Jermolene:
Marvellous, thank you @lilscribby much appreciated.

PR #7639, opened 2023-07-25
Add docs on achieving unique `tag` macro dropdowns inside `list` widget

mateuszwilczek:
There are recurring questions about this situation, see:

* https://talk.tiddlywiki.org/t/use-variable-transclusion-in-a-list-widget-to-get-unique-popups/3958
* https://talk.tiddlywiki.org/t/clicking-on-a-tag-widget-opens-pop-up-in-all-the-occurrences-of-the-tag-widget/5188
* https://talk.tiddlywiki.org/t/tag-macro-multiple-dropdowns-displayed-when-multiple-macros-present-bug/5652
* https://talk.tiddlywiki.org/t/how-to-get-the-dropdown-working-correctly-for-tag-pills-that-are-transcluded-multiple-times-within-a-tiddler/7660

The most common use case seems to be when multiple identical tag macros are generated by a list widget. Clicking one of the tag pills opens dropdowns on all of them. This can be easily circumvented, as discussed in the first thread linked above. I thought it is worth clarifying in the official docs, to hopefully prevent confusion in the future.

This is only a superficial explanation of how to solve this common issue. It could be improved by explaining the mechanisms behind it, or why `variable` is more performant than `counter`. I don't understand these mechanisms enough to be able to document it at the moment. In the meantime, I think that some basic documentation is better than none.


Jermolene:
Great stuff thank you @mateuszwilczek

PR #7640, opened 2023-07-25
Use 'translink' macro on widget attribute documentation

btheado:
The Widget Attributes tiddler contains links to separate tiddlers which explain the 5 different attribute syntaxes. I find that clicking so many links is enough friction that transcluding the content is a worthwhile thing to do.

Use the translink macro to transclude and link to the tiddlers.

### DIFF
--- a/editions/tw5.com/tiddlers/wikitext/Widget Attributes.tid
+++ b/editions/tw5.com/tiddlers/wikitext/Widget Attributes.tid
@@ -4,7 +4,7 @@ tags: WikiText
 title: Widget Attributes
 type: text/vnd.tiddlywiki
 
-Attributes of HTML elements and widgets can be specified in several different ways:
+Attributes of [[HTML elements|HTML in WikiText]] and widgets can be specified in several different ways:
 
 * [[a literal string|Literal Attribute Values]]
 * [[a transclusion of a textReference|Transcluded Attribute Values]]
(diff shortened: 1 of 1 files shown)


btheado:
* old: https://tiddlywiki.com/#Widget%20Attributes
* new: https://tiddlywiki5-git-fork-btheado-transclude-widget-c50590-jermolene.vercel.app/#Widget%20Attributes

mateuszwilczek:
@btheado thanks for taking your time to improve the documentation.

Technical remark: the commits to the `master` branch are reserved for the core code, the documentation should be committed to the `tiddlywiki-com` branch.

Opinion about the commit itself: I understand your motivation, but I think that using the translink macro is not the best solution in this case. The additional tiddlers, particularly the last one, contain a lot of formatting including code blocks and tables - these are much less readable inside the translink macro (side note: a reason to improve the translink styling perhaps?).
I think it is better left as it is, but let's hear what others say.

btheado:
> Technical remark: the commits to the `master` branch are reserved for the core code, the documentation should be committed to the `tiddlywiki-com` branch.

Thanks for catching this. Up until now I've caught it myself every time before I created the pull request.

> these are much less readable inside the translink macro 

I still favor some sort of transclusion. I find the friction of having to click so many links very high. Maybe it would help if the examples from the last one were factor out into a different tiddler. Then it wouldn't be so long.

> side note: a reason to improve the translink styling perhaps?

This macro was recently updated in 5.3.0. Seeing the PR go by was the only reason I was aware of it. Because it was recently refreshed doesn't mean it can' t be improved...I just wanted to point it out.

btheado:
> I think it is better left as it is, but let's hear what others say.

I agree. When creating this pull request, I was thinking it might be controversial.

Jermolene:
I agree that the way that one has to click links in `Widget Attributes` could be improved. I think I'd be inclined to manually transclude each of them, with a heading that links to the individual tiddler.

btheado:
Thanks, @Jermolene. I'll open a new PR against the proper branch and replace the translink with transcludes.

PR #7641, opened 2023-07-26
Improve docs on styles and classes in WikiText

mateuszwilczek:
- Fix minor inaccuracies about using styles and classes with `@@`.
- Add info about using classes with other WikiText elements (previously not documented).


Jermolene:
Thank you @mateuszwilczek

PR #7643, opened 2023-07-27
Signing CLA

catter-fly:


Jermolene:
Thanks @catter-fly and welcome to the project. I've pushed an update to fix the double entry created by the current bug in the PR submission tool.

PR #7647, opened 2023-07-29
Fix transclude inefficiency

### LINKED ISSUE #7592: [BUG] Export all on tiddlywiki.com 5.3.0 hangs the browser
### Describe the bug

Invoking the export all function on tiddlywiki.com 5.3.0 hangs the browser.

### Expected behavior

It should export in under a second like it did in 5.2.7

### To Reproduce

1. Go to [tiddlywiki.com](http://tiddlywiki.com/)
2. Click on the tools sidebar tab
3. Click the export all button and choose json (I didn’t try the others)
4. The browser is hung

### Screenshots

_No response_

### TiddlyWiki Configuration

tiddlywiki.com 5.3.0, firefox, linux. Others have duplicated on other OS.

### Additional context

Git bisect points the finger at this commit: https://github.com/Jermolene/TiddlyWiki5/pull/6666/commits/ec1ec8ccd8942adb33f6aa148219bc7759c0bca7. Go one commit earlier and the misbehavior is not there.

This issue was originally reported at https://talk.tiddlywiki.org/t/export-all-hangs-browser-in-5-3-0/7496/

Jermolene:
Fixes #7592

### DIFF
--- a/core/modules/widgets/transclude.js
+++ b/core/modules/widgets/transclude.js
@@ -41,8 +41,15 @@ TranscludeWidget.prototype.execute = function() {
 	this.collectAttributes();
 	this.collectStringParameters();
 	this.collectSlotFillParameters();
+	// Determine whether we're being used in inline or block mode
+	var parseAsInline = !this.parseTreeNode.isBlock;
+	if(this.transcludeMode === "inline") {
+		parseAsInline = true;
+	} else if(this.transcludeMode === "block") {
+		parseAsInline = false;
+	}
 	// Get the target text and parse tree nodes that we are transcluding
-	var target = this.getTransclusionTarget(),
+	var target = this.getTransclusionTarget(parseAsInline),
 		parseTreeNodes;
 	this.sourceText = target.text;
 	this.parserType = target.type;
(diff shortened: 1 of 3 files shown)


core/modules/widgets/transclude.js
```diff
@@ -41,28 +41,45 @@ TranscludeWidget.prototype.execute = function() {
...
 			parseTreeNodes = [{type: "text", text: this.sourceText}];
 			break;
 		default:
-			// text/plain
-			var plainText = this.wiki.renderText("text/plain",this.parserType,this.sourceText,{parentWidget: this});
-			parseTreeNodes = [{type: "text", text: plainText}];
+			// "text/plain" is the plain text result of wikifying the text
+			target = this.parseTransclusionTarget();
+			this.sourceText = target.text;
+			this.parserType = target.type;
```
pmario, line 75: The 2 lines above seem to be the same code for all 3 states. They may be moved above the switch statement. Or is it intended to be that way?

core/modules/widgets/transclude.js
```diff
@@ -41,28 +41,45 @@ TranscludeWidget.prototype.execute = function() {
...
+	var parseTreeNodes, target;
 	// Process the transclusion according to the output type
 	switch(this.transcludeOutput || "text/html") {
 		case "text/html":
-			// Return the parse tree nodes
+			// Return the parse tree nodes of the target
+			target = this.parseTransclusionTarget(parseAsInline);
+			this.sourceText = target.text;
+			this.parserType = target.type;
+			this.parseAsInline = target.parseAsInline;
```
pmario, line 61: It seems `this.parseAsInline` is only initialized if this case has been executed. Is that intentional? IMO undefined means "we do not know".  The variable `parseAsInline` is defined as true or false. Just a gut feeling.

core/modules/widgets/transclude.js
pmario: I think `this.parseAsInline` should also initialized and moved outside the cases.

Jermolene: > The 2 lines above seem to be the same code for all 3 states. They may be moved above the switch statement. Or is it intended to be that way?

Yes that can be made a bit more efficient.

> I think `this.parseAsInline` should also initialized and moved outside the cases.

I don't think so. It is not required on all code paths

core/modules/widgets/transclude.js
Jermolene: The code relies on this.parseAsInline defaulting to false

PR #7648, opened 2023-07-29
Fixed error "Global assignment is not allowed within modules on node"

### LINKED ISSUE #7585: [BUG] NodeJS fails with "Global assignment is not allowed within modules on node." since v5.3.0
### Describe the bug

Running on NodeJS on TiddlyWIki version 5.3.0. Running the battery of Relink Unit tests results in an aborted run with the warning "Global assignment is not allowed within modules on node.", which points to a semi-random file which clearly doesn't have the issue.

Something changed between TiddlyWiki v5.2.x and v5.3.0 which made NodeJS not function. A Google search on that error pulls up two results within the entire internet. One is issues #7465, the other is an issue submitted to TiddlyMap which I ignored until it went away.

### Expected behavior

_No response_

### To Reproduce

_No response_

### Screenshots

_No response_

### TiddlyWiki Configuration

NodeJS version: v15.5.1
TiddlyWiki version: v5.3.0

### Additional context

_No response_

Jermolene:
This error message was introduced in #4624 as:

```
Global assignment is not allowed within modules on node
```

Since #4624 was released in v5.3.0 it has been found that the error is raised by a number of popular existing plugins when they are run under Node.js.

The initial approach to fixing it taken here is to make it a warning to the console, but I'm wondering whether it might not be more useful to entirely remove the global check.

Fixes #7585

@Arlen22 @flibbles @saqimtiaz I'd welcome your thoughts.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -575,9 +575,8 @@ var globalCheck =[
 	"    configurable: true",
 	"  });",
 	"  if(Object.keys(__temp__).length){",
-	"    console.log(Object.keys(__temp__));",
+	"    console.log(\"Warning: Global assignment detected\",Object.keys(__temp__));",
 	"    delete Object.prototype.__temp__;",
-	"    throw \"Global assignment is not allowed within modules on node.\";",
 	"  }",
 	"  delete Object.prototype.__temp__;",
 ].join('\n');


Arlen22:
I don't understand why that would be happening. The code in boot.js looks like it already skips the global check if $tw.browser is set. Is that not getting set properly in some scenarios?

flibbles:
For me, all the problems I had with this were people trying to start up a nodeJS server, and they couldn't do anything: start a server, generate a standalone tiddlywiki, *import plugins to fix the problem*. The `tiddlywiki` command line tool always failed with an error.

Jermolene:
> @Arlen22 I don't understand why that would be happening. The code in boot.js looks like it already skips the global check if $tw.browser is set. Is that not getting set properly in some scenarios?

> @flibbles For me, all the problems I had with this were people trying to start up a nodeJS server, and they couldn't do anything: start a server, generate a standalone tiddlywiki, _import plugins to fix the problem_. The `tiddlywiki` command line tool always failed with an error.

Apologies the OP was wrong to say that the problem occurs on the browser. I've updated the OP with a link to #7585.

Given that, what do you think @Arlen22? What is the case for retaining the error?

Arlen22:
Well, if you don't think it's a problem, I have no other case to make. The main reason I kept it that way was that the original behavior was simply to isolate each module, which kept such JavaScript errors contained. 

I didn't know this was going to conflict with undeclared variables being assigned to global. In retrospect, I may have approached it more cautiously or recommended that the function code include `"use strict";` when loading modules, which would be even more breaking, although the stack trace is a lot better. 

Polluting the global namespace in Node is definitely not a standard practice and that is the only situation where this check is useful, so removing it isn't a big deal.

Arlen22:
@Jermolene Well, if you don't think it's a problem, I have no other case to make. The main reason I kept it that way was that the original behavior was simply to isolate each module, which kept such JavaScript errors contained. 

I didn't know this was going to conflict with undeclared variables being assigned to global. In retrospect, I may have approached it more cautiously or recommended that the function code include `"use strict";` when loading modules, which would be even more breaking, although the stack trace is a lot better. 

Polluting the global namespace in Node is definitely not a standard practice and that is the only situation where this check is useful, so removing it isn't a big deal.

PR #7650, opened 2023-07-31
Improve `jsonstringify` and `stringify` operators docs

mateuszwilczek:
Fix redundant double backward slashes in front of escaped characters, e.g. change incorrect escaped newline `\\n` to the correct `\n`.

In fixing the above issue I have discovered that the tiddlers `jsonstringify Operator` and `stringify Operator`, as well as their respective linked example tiddlers, have exactly the same content (except for the "Also see the other one" at the end). There is no transclusion in use, so any changes to one of those docs tiddlers has to be manually copied over to the corresponding tiddler for the other operator.

I see two possible solutions:

1. If the two filter operators `jsonstringify` and `stringify` work exactly the same (and the two different names exist for convenience or backwards compatibility only): full documentation should be written only for one of them, and the docs of the other one should only link to the first one. This way it should be clear that both operators are simply aliases for the same underlying function.

2. If there is a difference between these operators that is not yet documented: some sort of transclusion should be in use, along with at least a very brief description of the differences between them.

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/jsonstringify Operator (Examples).tid
+++ b/editions/tw5.com/tiddlers/filters/examples/jsonstringify Operator (Examples).tid
@@ -1,9 +1,0 @@
-created: 20171029155046637
-modified: 20171029155227382
-tags: [[Operator Examples]] [[stringify Operator]]
-title: jsonstringify Operator (Examples)
-type: text/vnd.tiddlywiki
-
-<<.operator-example 1 """[[Title with "double quotes" and single ' and \backslash]] +[jsonstringify[]]""">>
-<<.operator-example 2 """[[Accents and emojis -> äñøßπ ⌛🎄🍪🍓 without suffix]] +[jsonstringify[]]""">>
-<<.operator-example 3 """[[Accents and emojis -> äñøßπ ⌛🎄🍪🍓 with rawunicode suffix]] +[jsonstringify:rawunicode[]]""">>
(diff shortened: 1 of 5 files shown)


mateuszwilczek:
Thanks for confirming that @Jermolene.

What about the duplicated docs for these operators? I think this is a more important issue here. I have changed this PR to draft, so that we don't forget it.

I don't have much understanding of TW's JS internals, but from what I have explored, both operators, or associated JS functions, seem to be defined identically. `stringify` was intriduced in v5.1.14, `jsonstringify` in v5.1.15, the `rawunicode` suffix was added to both of them in v5.1.23. It seems to me that `jsonstringify` is just an unnecessary duplication of `stringify`. If I'm right, then this is an issue in itself (which I have not reported yet, as I may be wrong).

Since both operators seem to be functioning identically, I think it is unnecessary to keep identical docs under different titles for both of them, it adds to confusion.

Jermolene:
Hi @mateuszwilczek I am baffled as to how we ended up with two such similar utility methods and operators. I am also unclear why the escaped characters are slightly different in each case; there's a decent chance that there was a reasonable rationale for the near-duplication, but it is lost now.

These are worthwhile improvements. I wonder if it might be helpful to explicitly point out the similarity between the operators, and acknowledge it as a mistake.

mateuszwilczek:
Thanks @Jermolene for looking into this. I will try to de-duplicate the documentation, assuming for now that both operators work identically and that `stringify` is the first/main one. I will also ask on the Talk forum, maybe someone knows the difference between them, if there is any.

mateuszwilczek:
@Jermolene I have rearranged the docs assuming that the operators are identical and treating `stringify` as the main one. I have marked this PR as ready for review.

I will wait some time to see if anyone responds to the question on Talk. If we can't find the reason for these operators being separate entities, I'll create an issue and change the docs to clearly state that `jsonstringify` was a mistake.

Jermolene:
Thank you for your patience @mateuszwilczek

PR #7667, opened 2023-08-05
Adds a note on subfilter expressions to `ActionListopsWidget` docs

yaisog:
This is something that I never get to work right off the bat. The parameter description is correct, but one needs to think deeply on the syntax of (sub)filter expressions to get it to work. The main doc tiddler contains only examples that append items, which is much less worrisome. Only deep within the examples will one find something that points towards the correct way.
I think the `toggle` example might be a pretty regular but simple use case that should be easy to understand.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
@@ -1,6 +1,6 @@
 caption: action-listops
 created: 20141025120850184
-modified: 20230301183438774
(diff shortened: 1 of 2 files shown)


Jermolene:
That's very helpful thank you @yaisog

kookma:
Hi @yaisog 

I assume the https://tiddlywiki.com/prerelease/#ActionListopsWidget%20(Examples) tiddler should tagged with `ActionListopsWidget`. This is to be consistent with other example tiddlers.

PR #7671, opened 2023-08-06
Change favicon format from ICO to PNG, fix #7665

### LINKED ISSUE #7665: [IDEA] Use PNG favicon instead of ICO on tiddlywiki.com and related sites
The favicons used on tiddlywiki.com and related sites (prerelease, dev, upgrader, links), implemented using tiddler `$:/favicon.ico`, are in uncompressed ICO format. **This causes them to be bloated and causes minor compatibility issues.** The ICOs contain an inconsistent amount of embedded sizes:

| site | 16 | 32 | 48 | 64 | 128 |
| - | - | - | - | - | - |
| tiddlywiki.com | ✔ | ✔ |   | ✔ | ✔ |
| dev, static            | ✔ | ✔ |   |   |   |
| prerelease | ✔ | ✔ | ✔ | ✔ | ✔ |
| upgrader   |   | ✔ |   | ✔ |   |

I have looked at the files located [here in this repo](https://github.com/Jermolene/TiddlyWiki5/tree/34bc9c72c65200c6036ce6cb6ed40998900e8a06/editions/tw5.com/tiddlers/_tw_shared/favicons).

Use PNG favicons instead. This format has [support on _all_ major browsers](https://caniuse.com/link-icon-png), superseeding that of ICO or [SVG](https://caniuse.com/link-icon-svg). I suggest 192x192 size, but see additional context below.

* **Leave as is** (ICO with multiple embedded sizes, without compression)
  * Bloated icon size, e.g. 120 kB for prerelease.
  * Not displayed on Chrome/iOS (I don't know if this is issue with TW's implementation of ICO favicons, or with the browser; the second would be strange, because both Chrome on other platforms and iOS Safari engine support it).
  * No relevant advantage over PNG - ICO's only advantage over PNG is ability to store images in multiple sizes, including specifically pixel-adjusted small versions like 16x16. From what I have seen, this is not put to use or does not make much difference in case of the tiddlywiki.com favicon. The ICO (containing a 16x16 version) is only marginally more sharp than 128x128 PNG rescaled by the browser. I guess the 16x16 version wasn't in any way adjusted for 16x16, but exporting it from vector source in this resolution is marginally sharper than rescaling a 128x128 PNG.
* **Compressed ICO**
  * Using PNG compression inside the ICO causes the file to be around 10x lighter, e.g. tiddlywiki.com favicon is reduced from 100 kB to 10 kB, but
  * Not supported on Safari/macOS and on iOS.
* **SVG**
  * Technically ideal in terms of file size/ quality ratio, but
  * Not supported on Safari/macOS, iOS, and on Firefox/Android.

**Additional context**
The favicon support on different browsers and platforms [is a mess](https://stackoverflow.com/questions/4014823/does-a-favicon-have-to-be-32%C3%9732-or-16%C3%9716). Ideal representation on all browsers/platforms requires a couple of files in various formats and accompanying declarations/manifest. This is neither achievable using only `$:/favicon.ico` mechanism, nor really necessary IMO.

In our case, I think PNG is the best compromise:

* the widest compatibility,
* reasonable file size (much better than uncompressed ICO, only slightly more than SVG),
* decent quality (marginally worse than ICO for small sizes; obviously worse than SVG for much larger sizes, but these are rarely needed for a favicon).

From what I have read, the 192x192 size seems to be the largest one widely used by various browsers. A PNG of these dimensions would still weigh much less than the current uncompressed ICO.

**I'm ready to help in creating the proposed PNG favicons.** I have created an issue instead of PR, because:

* I would like a confirmation that my proposal makes sense first.
* I don't know what would be the correct way to make PR (change files in https://github.com/Jermolene/TiddlyWiki5/tree/34bc9c72c65200c6036ce6cb6ed40998900e8a06/editions/tw5.com/tiddlers/_tw_shared/favicons, push new `$:/favicon.ico` to tiddlywiki-com branch?).
* If a resolution larger than the existing 128x128 is chosen for the new PNGs, it would be easier to work on an existing vector source, which I couldn't find (it's easy to recreate though).

---

Edit: I know it's a low priority issue, but it's easy to solve, and I think tiddlywiki.com should give a good example.

mateuszwilczek:
This PR changes format of all ICO favicons to PNG. The ICO format used on tiddlywiki.com and related sites does not offer any relevant advantages over PNG, weighs more, and has minor incompatibilities. I have explained in detail the reasons for why I think PNG is the best choice in our case in this issue, which this PR fixes https://github.com/Jermolene/TiddlyWiki5/issues/7665.

The change results in smaller file sizes in _all_ cases, e.g. for tiddlywiki.com favicon from 106 kB to 5.63 kB. The differences are not so big for other sites, e.g. dev, because their ICO favicons had small resolutions like 32x32, they are now replaced with 128x128 PNG favicons that still weigh slightly less. All PNGs introduced in this PR have 128x128 resolution (except for classic.tiddlywiki.com). This is the largest resolution that any of the replaced files had.

I'm marking it as draft PR, because of the following points that need resolving:

- [x] I'm new to the repo's structure regarding all the editions and building system tiddlers such as `$:/favicon.ico`, **I'm not sure I have made all the right changes to make the new PNG favicons work**.
- [x] Perhaps this is an occasion to **clean up the redundant favicons from different language editions**. Many contain their own copy of the favicons, see all the identical files I have changed. Do I understand correctly, that the editions could "inherit" the favicons, as any other contents, from `_tw_shared`?
- [x] I would be glad to **add my SVG source file to the repo** (in case it is needed to generate images at other resolutions, or if we want to switch to SVG favicons in the future), but **I don't know what would be the right place for it**.
- [x] **Consider other resolutions**, although I think 128x128 is a reasonable compromise. Exporting PNGs at another resolution from my SVG template is easy if we want that. I've considered 192x192 - this resolution is the minimum recommended by Chrome/Android when making desktop shortcuts and is larger than 180x180 wished by Apple for iOS desktop shortcuts. File sizes would be slightly larger, e.g. for tiddlywiki.com 8.53 kB instead of 5.63 kB (but that's still a lot less than original 106 kB ICO). We could also go the other way, make the favicons smaller, e.g. 32x32, if we want to target only the miniatures in tabs, and not speed dial/ new tab/ mobile desktop shortcuts.


editions/ja-JP/tiddlywiki.info
```diff
@@ -30,8 +30,8 @@
 			"--rendertiddler","$:/editions/ja-JP/download-empty","empty.html","text/plain",
 			"--rendertiddler","$:/editions/ja-JP/download-empty","empty.hta","text/plain"],
 		"favicon": [
-			"--savetiddler","$:/favicon.ico","favicon.ico",
```
Jermolene, line 33: The "./favicon.ico" was historically hardwired into browsers, and so for backwards compatibility I'd be inclined to keep the `favicon.ico` output filename for the PNG file.

Jermolene (commented):
Thanks @mateuszwilczek this looks good to go, with only a minor comment below

mateuszwilczek:
@Jermolene I have reverted to the `favicon.ico` filenames, as you suggested.

I'm marking this PR as ready for review now.

Does it make sense that I add my source SVG file somewhere in the repo? It would be useful if anyone would like to tinker with the favicons in the future. I will of course be keeping it for now, in case some adjustments will be needed soon. This file is not supposed to be a part of any edition really, so I don't know what would be the right place for it in the repo. Of course this file is easy to recreate with the Motovun/download-button SVGs from the site and grabbing the colors from PNGs, but I'd like to spare someone's redundant future work if possible.

pmario:
>Does it make sense that I add my source SVG file somewhere in the repo?

Yes.

mateuszwilczek:
I have added `favicons.svg`, my template for creating the current PNG favicons, along with some helpful information about it in `favicons.svg.md` in `editions/tw5.com/tiddlers/_tw_shared/favicons/`, as this is where most of the favicons are collected.
These files are not supposed to be included anywhere in the wiki, I hope placing them here without associated `.meta` files will achieve that.
If there is a better place for these files, please suggest.

pmario:
>Or is this only a Vercel preview thing and the tiddlers will not show up in the actual wiki?

If the preview shows it the wiki will show it too. I'll have a closer look, if it's possible to create an "assets" folder and add an empty "tiddlywiki.files" to it. So all elements there should be excluded from the wiki itself. So they are only part of the the github repo. 

Not sure, if Jeremy wants that.

mateuszwilczek:
I have marked this as ready for review, so we can use the lighter PNG instead of ICO favicons.

The only unresolved thing was where to put the source SVG file, which is not supposed to be included anywhere in the wiki. I wanted it to be preserved for the future case of someone working further on the icons. Since we're in the process of reworking the Motovun/ icon design in general, this is no longer relevant.

Jermolene:
Thanks @mateuszwilczek looks good to go to me. I do think it would be worth including the source SVG file, perhaps we could just include in tw5.com as another example image.

mateuszwilczek:
I have added the source SVG as a tw-com tiddler. If that's acceptable, I'll add some information about it and links to it later, when this PR is merged. I prefer working on the documentation using Saq's PR Maker (from which I cannot push to this branch).

The file is unmodified version created by Inkscape (with all its metadata), so that it is most useful for generating all the various flavors of the favicon. Removing these metadata would make it not much more than the _Motovun Jack.svg_ overlaid on a circle. It weighs only 27 kB, and the ICO we're removing was 106 kB, so this whole PR will still make wiki lighter.

Jermolene:
Great thanks @mateuszwilczek

kookma:
@pmario 
In https://tiddlywiki.com/prerelease/ I see some tiddlers like

[/home/runner/work/TiddlyWiki5/TiddlyWiki5/editions/tw5.com/tiddlers/_tw_shared/favicons/favicons.svg](https://tiddlywiki.com/prerelease/#%2Fhome%2Frunner%2Fwork%2FTiddlyWiki5%2FTiddlyWiki5%2Feditions%2Ftw5.com%2Ftiddlers%2F_tw_shared%2Ffavicons%2Ffavicons.svg)
[/home/runner/work/TiddlyWiki5/TiddlyWiki5/editions/tw5.com/tiddlers/_tw_shared/favicons/favicons.svg.md](https://tiddlywiki.com/prerelease/#%2Fhome%2Frunner%2Fwork%2FTiddlyWiki5%2FTiddlyWiki5%2Feditions%2Ftw5.com%2Ftiddlers%2F_tw_shared%2Ffavicons%2Ffavicons.svg.md)

Search for favicon in searchbox.

I assume these are left over after merging this PR.

mateuszwilczek:
@kookma thanks for noticing. These were of course never supposed to show up in the wiki like this, I have overlooked that this was not solved and forgot to remove the files. I'll make a PR to fix this.

PR #7679, opened 2023-08-15
Fix checkbox widget when listIndex field undefined

### LINKED ISSUE #7678: [BUG] CheckboxWidget listIndex mode no longer working after date field fix
### Describe the bug

https://github.com/Jermolene/TiddlyWiki5/pull/7448 was intended to fix the issue with `created` or `modified` being passed as the listField attribute of the checkbox widget, but it broke `listIndex` mode. See https://talk.tiddlywiki.org/t/checkboxwidget-listindex-mode-no-longer-working-in-5-3-0/7460 for the bug report I'm reproducing here.

### Expected behavior

The `listIndex` attribute of checkbox widgets continues to work as it did in 5.2.7.

### To Reproduce

1. Go to https://tiddlywiki.com/#CheckboxWidget%20%28listIndex%20Mode%29
2. Click on the checkboxes in the example
3. Note how the "Colors list:" line is not updating

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: 5.3.0
- Saving mechanism: any
- Plugins installed: none

- OS: any
- Browser: any
- Version: N/A

- Device: any
- OS: any
- Browser: any
- Version: N/A

### Additional context

I plan to work on this immediately, to try to get it in before 5.3.1 releases.

rmunn:
When listIndex was pointing to an entry that didn't exist yet, so that extractTiddlerDataItem returned undefined, the checkbox widget was aborting and refusing to proceed. This is not the desired behavior: we want it to create the index if it didn't exist yet. So now we explicitly check for undefined and treat it the same as an empty list.

Also fixed one code-style issue in the same file while I was in here.

Fixes #7678.

### DIFF
--- a/core/modules/widgets/checkbox.js
+++ b/core/modules/widgets/checkbox.js
@@ -116,7 +116,7 @@ CheckboxWidget.prototype.getValue = function() {
 				} else {
 					list = $tw.utils.parseStringArray(this.checkboxDefault || "") || [];
 				}
-			} else if (this.checkboxListIndex) {
+			} else if(this.checkboxListIndex) {
 				list = $tw.utils.parseStringArray(this.wiki.extractTiddlerDataItem(tiddler,this.checkboxListIndex,this.checkboxDefault || "")) || [];
 			} else {
 				list = this.wiki.filterTiddlers(this.checkboxFilter,this) || [];
@@ -215,6 +215,8 @@ CheckboxWidget.prototype.handleChangeEvent = function(event) {
 		if($tw.utils.isArray(fieldContents)) {
 			// Make a copy so we can modify it without changing original that's refrenced elsewhere
 			listContents = fieldContents.slice(0);
+		} else if(fieldContents == undefined) {
+			listContents = [];
 		} else if(typeof fieldContents === "string") {
 			listContents = $tw.utils.parseStringArray(fieldContents);
 			// No need to copy since parseStringArray returns a fresh array, not refrenced elsewhere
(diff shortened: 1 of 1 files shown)


rmunn:
An alternative way to fix this bug would have been to change 

https://github.com/Jermolene/TiddlyWiki5/blob/641d92ef7b344718a5eb1c6d76722846fc939c98/core/modules/widgets/checkbox.js#L213

to be:

```js
 fieldContents = this.wiki.extractTiddlerDataItem(this.checkboxTitle,this.checkboxListIndex,[]);
```

@Jermolene, please let me know if you would prefer this one-line fix instead of the two-line fix I ultimately went with, and I'll change it. I decided in favor of the two-line fix because it would let me create a later PR that cleans up a bit of the code and makes one use case slightly (*very* slightly) more performant. (If you're wondering, it would let me stop making copies of an empty array, leading to one less small object that the browser needs to garbage-collect. A very, very small performance win, and not worth the added complexity in a PR just days before release; I'll get that PR into 5.3.2).

core/modules/widgets/checkbox.js
```diff
@@ -215,6 +215,8 @@ CheckboxWidget.prototype.handleChangeEvent = function(event) {
 		if($tw.utils.isArray(fieldContents)) {
 			// Make a copy so we can modify it without changing original that's refrenced elsewhere
 			listContents = fieldContents.slice(0);
+		} else if(fieldContents == undefined) {
```
pmario, line 218: we usually use `fieldContents === undefined` to avoid type casting.

rmunn: I wondered if `null` should also be considered as equivalent to an empty list. But I suppose if someone has gone to great effort to put a `null` into a TW field (which would be very hard unless you're trying to do it on purpose), they can just deal with the fact that it doesn't work with the checkbox widget. I'll change this to `===`.

rmunn:
@Jermolene This is a bugfix for a bug I introduced in 5.3.0, so I'd love to get it into 5.3.1 if possible. I know it's post-freeze, but since this is such an easy-to-review bugfix for a rather annoying bug (the `listIndex` feature basically doesn't work without this bugfix), I'd like to request a freeze exception for this PR, and ask you to look at it and merge it in before 5.3.1.

PR #7683, opened 2023-08-18
Fix encodebase64 and decodebase64 filters

### LINKED ISSUE #7626: Unexpected and undocumented behaviour for $tw.utils.base64Encode()
[As reported](https://github.com/Jermolene/TiddlyWiki5/commit/9b2af1359614f4ad5afd05be7cf9853909334592) by @Jermolene, the utility method `base64encode()` can yield unexpected results in some situations that lead to garbled and unusable data.

Investigation shows that the method `base64encode()` is not a replacement for the the `window.btoa()` method.

- `window.btoa()` cannot handle arbitrary Unicode text that might contain characters that occupy more than one byte , only binary strings (i.e., a string in which each character in the string is treated as a byte of binary data).
- to work around this and to be usable for arbitrary text, `$tw.utils.base64Encode()` first [encodes the input as UTF8](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/utils/base64-utf8/base64-utf8.module.js#L97). This fits the use case for things like the [GitHub Saver](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/savers/github.js#L73).
- However, it means that when you pass a binary string to `$tw.utils.base64Encode()`, the UTF8 encoding leads to a garbled output. Disabling the UTF8 encoding by commenting out the relevant line leads to the expected behaviour. 

We could potentially add a flag to the method for disabling the UTF8 encoding. We should document this for tiddlywiki.com/dev as well as in the code itself.

rmunn:
Fixes #7626.

The documentation for the `encodebase64` filter operator says that the input is treated as binary data, but in fact the input is being treated as text data, with an extra UTF-8 encoding step being performed first.

Likewise, the `decodebase64` documentation says that it outputs binary data, but in fact it will do a UTF-8 decoding step before producing output, which will in fact garble binary data.

This PR changes the behavior of encodebase64 and decodebase64 to match what the documentation says they do. It also adds an optional `text` suffix to both filters to keep the current behavior.

Finally, an optional `urlsafe` suffix is added to both filters to allow them to use the "URL-safe" variant of base64 (using `-` instead of `+` and `_` instead of `/`).

### DIFF
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -18,7 +18,7 @@ Export our filter functions
 
 exports.decodebase64 = function(source,operator,options) {
 	var results = [];
-	var binary = !operator.suffixes || operator.suffixes.indexOf("text") === -1;
+	var binary = operator.suffixes && operator.suffixes.indexOf("binary") !== -1;
 	var urlsafe = operator.suffixes && operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
 		results.push($tw.utils.base64Decode(title,binary,urlsafe));
@@ -28,7 +28,7 @@ exports.decodebase64 = function(source,operator,options) {
 
 exports.encodebase64 = function(source,operator,options) {
 	var results = [];
-	var binary = !operator.suffixes || operator.suffixes.indexOf("text") === -1;
+	var binary = operator.suffixes && operator.suffixes.indexOf("binary") !== -1;
 	var urlsafe = operator.suffixes && operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
 		results.push($tw.utils.base64Encode(title,binary,urlsafe));
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -18,8 +18,8 @@ Export our filter functions
 
 exports.decodebase64 = function(source,operator,options) {
 	var results = [];
-	var binary = operator.suffixes.indexOf("text") === -1;
-	var urlsafe = operator.suffixes.indexOf("urlsafe") !== -1;
+	var binary = !operator.suffixes || operator.suffixes.indexOf("text") === -1;
+	var urlsafe = operator.suffixes && operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
 		results.push($tw.utils.base64Decode(title,binary,urlsafe));
 	});
@@ -28,8 +28,8 @@ exports.decodebase64 = function(source,operator,options) {
 
 exports.encodebase64 = function(source,operator,options) {
 	var results = [];
-	var binary = operator.suffixes.indexOf("text") === -1;
-	var urlsafe = operator.suffixes.indexOf("urlsafe") !== -1;
+	var binary = !operator.suffixes || operator.suffixes.indexOf("text") === -1;
+	var urlsafe = operator.suffixes && operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
 		results.push($tw.utils.base64Encode(title,binary,urlsafe));
 	});
--- a/core/modules/filters/encodings.js
+++ b/core/modules/filters/encodings.js
@@ -18,16 +18,20 @@ Export our filter functions
 
 exports.decodebase64 = function(source,operator,options) {
 	var results = [];
+	var binary = operator.suffixes.indexOf("text") === -1;
+	var urlsafe = operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
-		results.push($tw.utils.base64Decode(title));
+		results.push($tw.utils.base64Decode(title,binary,urlsafe));
 	});
 	return results;
 };
 
 exports.encodebase64 = function(source,operator,options) {
 	var results = [];
+	var binary = operator.suffixes.indexOf("text") === -1;
+	var urlsafe = operator.suffixes.indexOf("urlsafe") !== -1;
 	source(function(tiddler,title) {
-		results.push($tw.utils.base64Encode(title));
+		results.push($tw.utils.base64Encode(title,binary,urlsafe));
 	});
 	return results;
 };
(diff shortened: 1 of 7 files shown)


saqimtiaz:
> This PR changes the behavior of encodebase64 and decodebase64 to match what the documentation says they do. It also adds an optional `text` suffix to both filters to keep the current behavior.

Hi @rmunn, I feel that we should preserve the existing behaviour to minimize the impact on end users and update the documentation accordingly.

rmunn:
> > This PR changes the behavior of encodebase64 and decodebase64 to match what the documentation says they do. It also adds an optional `text` suffix to both filters to keep the current behavior.
> 
> Hi @rmunn, I feel that we should preserve the existing behaviour to minimize the impact on end users and update the documentation accordingly.

Except that the existing behavior (expecting text input) is rarely useful AFAICT. Most uses of `encodebase64`, I suspect, will be taking binary data. So I thought binary data should be the default behavior.

saqimtiaz:
@rmunn we have a strict backwards compatibility policy to not break the behaviour of wikitext code that might exist in the wild. Is this change sufficiently important to break backward compatibility @Jermolene ?

rmunn:
Does the backwards compatibility mandate extend to bugs? Because when the documentation says the wiki does one thing, but the actual behavior is something else, I think that's a bug and the behavior should be changed to match what the documentation says.

rmunn:
P.S. Here's what I wrote in the related bug report at https://github.com/Jermolene/TiddlyWiki5/issues/7626#issuecomment-1683419779:

> Since the encodebase64 filter was just added in release 5.2.6 in March 2023, five months ago, I doubt that many people have started using it yet, especially because it's buggy at the moment. So I don't think there are a lot of wikis out there relying on the current behavior, and I think it's safe to change the behavior.

Jermolene:
Thanks @rmunn @saqimtiaz on balance I would prefer to maintain backwards compatibility here, and update the docs.

Generally, I don't think the quality of our docs is high enough for us to regard the docs as the single source of truth; for us, it is still the code.

rmunn:
Okay. When I get back around to this PR (I'll probably finish the if/then/else and list join PRs first), I'll switch the default to text mode and add a "binary" suffix for turning off the UTF-8 encoding step. That will preserve the existing behavior by default. Then I'll update the documentation accordingly.

rmunn:
@Jermolene - Got back around to fixing the encodebase64 and decodebase64 filters. Please let me know what you think. I struggled a little with writing the documentation (should I assume people know what UTF-8 is?), so if you have suggestions to improve it they would be welcome. But the code works, and I've included unit tests to ensure it continues to work.

Once this is merged, a couple places in the code that use `window.btoa` (the GitHub saver and the tm-http-request handler code) could be replaced with `$tw.utils.btoa` or `$tw.utils.base64Encode(str,true)` now, so that they will work in Node.js environments where `window.btoa` is not available.

Jermolene:
> @Jermolene - Got back around to fixing the encodebase64 and decodebase64 filters. Please let me know what you think. I struggled a little with writing the documentation (should I assume people know what UTF-8 is?), so if you have suggestions to improve it they would be welcome. But the code works, and I've included unit tests to ensure it continues to work.

Thank you @rmunn, looks good. The docs definitely don't need to explain base64 and utf8 but it might be useful to link to good, durable sources for further information.

> Once this is merged, a couple places in the code that use `window.btoa` (the GitHub saver and the tm-http-request handler code) could be replaced with `$tw.utils.btoa` or `$tw.utils.base64Encode(str,true)` now, so that they will work in Node.js environments where `window.btoa` is not available.

Excellent, I think it's probably reasonable include those changes in this PR?

pmario:
source for utf-8: https://developer.mozilla.org/en-US/docs/Glossary/UTF-8
source for base64: https://developer.mozilla.org/en-US/docs/Glossary/Base64

rmunn:
> Excellent, I think it's probably reasonable include those changes in this PR?

Done. The one in the GitHub saver now encodes in text mode (I double checked and that's the correct way to encode an `Authorization: Basic` header; the previous version would have failed if someone's password had contained an emoji like 📚), and the one in the `http.js` file now encodes in binary mode.

rmunn:
> Thank you @rmunn, looks good. The docs definitely don't need to explain base64 and utf8 but it might be useful to link to good, durable sources for further information.

Done. There was already a link to the Base64 definition in MDN's glossary so I just added a link to the UTF-8 definition.

Jermolene:
Hi @rmunn I just came across this article which would have helped me in the first place:

https://web.dev/articles/base64-encoding

Do we have tests to check for the lone surrogate pair problem that they mention?

Jermolene:
Hi @rmunn I am seeing the following errors when running the test suite in the browser which may be related to this PR.

Visit https://tiddlywiki.com/prerelease/test.html and then scroll down to see the errors:

```
[Utility tests](https://tiddlywiki.com/prerelease/test.html?spec=Utility%20tests) > [should handle base64 encoding binary data in URL-safe variant](https://tiddlywiki.com/prerelease/test.html?spec=Utility%20tests%20should%20handle%20base64%20encoding%20binary%20data%20in%20URL-safe%20variant)
TypeError: Can only call Window.btoa on instances of Window (line 851)
    at <Jasmine>
[Utility tests](https://tiddlywiki.com/prerelease/test.html?spec=Utility%20tests) > [should handle base64 encoding binary data](https://tiddlywiki.com/prerelease/test.html?spec=Utility%20tests%20should%20handle%20base64%20encoding%20binary%20data)
TypeError: Can only call Window.btoa on instances of Window (line 851)
    at <Jasmine>
```

rmunn:
@Jermolene - https://github.com/Jermolene/TiddlyWiki5/pull/7814 will fix the mistake I made. (I didn't realize that `window.btoa` needed the Window instance, and therefore `exports.btoa = window.btoa` doesn't work because `this` is null.) As for setting up Playwright to run the tests in a real browser during CI, I'm willing to try to work on that but I can't prioritize it very highly just now.

Jermolene:
> @Jermolene - #7814 will fix the mistake I made. (I didn't realize that `window.btoa` needed the Window instance, and therefore `exports.btoa = window.btoa` doesn't work because `this` is null.)

Thank you, much appreciated.

> As for setting up Playwright to run the tests in a real browser during CI, I'm willing to try to work on that but I can't prioritize it very highly just now.

No problem.

PR #7688, opened 2023-08-21
CopyTiddler Toolbar Button & Clipboard Action Widget

GameDungeon:
This PR adds a new view toolbar button "Copy Tiddler" which is implemented through a new action widget $action-clipboard.

There exists some infrastructure for handling the clipboard in TiddlyWiki, but it was inadequate for the use case, and this new action widget should allow additional flexibility for users as well.

### DIFF
--- a/core/modules/widgets/action-clipboard.js
+++ b/core/modules/widgets/action-clipboard.js
@@ -1,0 +1,76 @@
+/*\
+title: $:/core/modules/widgets/action-deletetiddler.js
+type: application/javascript
+module-type: widget
+
+Action widget to delete a tiddler.
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+var Widget = require("$:/core/modules/widgets/widget.js").widget;
+
+var ClipboardWidget = function(parseTreeNode,options) {
+    this.initialise(parseTreeNode,options);
+};
+
+/*
+Inherit from the base widget class
+*/
+ClipboardWidget.prototype = new Widget();
+
+/*
+Render this widget into the DOM
+*/
+ClipboardWidget.prototype.render = function(parent,nextSibling) {
+    this.computeAttributes();
+    this.execute();
+};
+
+/*
+Compute the internal state of the widget
+*/
+ClipboardWidget.prototype.execute = function() {
+    this.actionTiddler = this.getAttribute("$tiddler");
+    this.actionTiddler = this.getAttribute("$text");
+    this.actionTiddler = this.getAttribute("$format");
+};
+
+/*
+Refresh the widget by ensuring our attributes are up to date
+*/
+ClipboardWidget.prototype.refresh = function(changedTiddlers) {
+    var changedAttributes = this.computeAttributes();
+    if(changedAttributes["$tiddler"] || changedAttributes["$text"] || changedAttributes["$format"]) {
+        this.refreshSelf();
+        return true;
+    }
+    return this.refreshChildren(changedTiddlers);
+};
+
+/*
+Invoke the action associated with this widget
+*/
+ClipboardWidget.prototype.invokeAction = function(triggeringWidget,event) {
+
+    const data = {
+        text: "I love TiddlyWiki",
+        title: "Copy Paste Test",
+        tags: "HelloThere",
+        apple: "Here is a test field"
+    };
+
+    e.clipboardData.setData("URL", "data:text/vnd.tiddler," + encodeURIComponent(JSON.stringify(data)));
+    e.preventDefault();
+
+    return true; // Action was invoked
+};
+
+exports["action-clipboard"] = ClipboardWidget;
+
+})();
+    
--- a/core/ui/ViewToolbar/copy-tiddler.tid
+++ b/core/ui/ViewToolbar/copy-tiddler.tid
@@ -1,0 +1,27 @@
+title: $:/core/ui/Buttons/new-here
+tags: $:/tags/ViewToolbar
+caption: {{$:/core/images/copy-clipboard}} {{$:/language/Buttons/CopyTiddler/Caption}}
+description: {{$:/language/Buttons/CopyTiddler/Hint}}
+
+\whitespace trim
+\define copyTiddlerAction()
+\whitespace trim
+<$action-clipboard $tiddler=<<currentTiddler>>/>
+\end
+
+\define copyTiddlerButton()
+\whitespace trim
+<$button actions=<<CopyTiddlerActions>> tooltip={{$:/language/Buttons/CopyTiddler/Hint}} aria-label={{$:/language/Buttons/CopyTiddler/Caption}} class=<<tv-config-toolbar-class>>>
+<$list filter="[<tv-config-toolbar-icons>match[yes]]">
+{{$:/core/images/copy-clipboard}}
+</$list>
+<$list filter="[<tv-config-toolbar-text>match[yes]]">
+<span class="tc-btn-text">
+<$text text={{$:/language/Buttons/CopyTiddler/Caption}}/>
+</span>
+</$list>
+</$button>
+\end
+<<copyTiddlerButton>>
+
+
--- a/core/wiki/macros/copy-to-clipboard.tid
+++ b/core/wiki/macros/copy-to-clipboard.tid
@@ -1,6 +1,8 @@
 title: $:/core/macros/copy-to-clipboard
 tags: $:/tags/Macro
 
+<!-- These are possibly now deprecated in prefrence to the action widget -->
+
 \define copy-to-clipboard(src,class:"tc-btn-invisible",style)
 \whitespace trim
 <$button class=<<__class__>> style=<<__style__>> message="tm-copy-to-clipboard" param=<<__src__>> tooltip={{$:/language/Buttons/CopyToClipboard/Hint}}>


GameDungeon:
I'm going to sleep for now, and will pick this up tomorrow, but I don't know why this is logging undefined. For some reason the attributes don't seem to be working properly.

Jermolene:
Hi @GameDungeon I wouldn't expect that we would need another action widget here. We already have the "tm-copy-to-clipboard" message  for copying text to the clipboard:

https://tiddlywiki.com/#WidgetMessage%3A%20tm-copy-to-clipboard

GameDungeon:
I know that message exists, but it was lacking quite a few features and wasn't designed in an extensible way. Additionally, it was designed only for plain text, so I can't copy a tiddler with it.

core/ui/ViewToolbar/copy-tiddler.tid
```diff
@@ -0,0 +1,27 @@
+title: $:/core/ui/Buttons/new-here
```
pmario: Is this the right title?

.gitignore
```diff
@@ -5,4 +5,4 @@
 tmp/
 output/
 node_modules/
-
+mynewwiki/
```
pmario, line 8: That's not possible. IMO you just forgot to remove it

pmario:
I did test it with FF on Ubuntu 22.04 -- There is no notification and it does not work with FF -- I'll have a closer look with FF and Win soon

core/modules/widgets/action-clipboard.js
```diff
@@ -0,0 +1,83 @@
...
+
+    function listener(e) {
+        e.clipboardData.setData("URL","data:text/vnd.tiddler," + encodeURIComponent(JSON.stringify(tiddlerData)));
+        e.clipboardData.setData("text",JSON.stringify(tiddlerData));
+        e.preventDefault();
+    }
+
+    document.addEventListener("copy",listener);
+    //document.execCommand("copy");
+    document.removeEventListener("copy",listener);
```
pmario, line 75: IMO those 2/3 lines do not make sense to me

core/ui/ViewToolbar/copy-tiddler.tid
```diff
@@ -0,0 +1,23 @@
+title: $:/core/ui/Buttons/copy-tiddler
+tags: $:/tags/ViewToolbar
+caption: {{$:/core/images/copy-clipboard}} {{$:/language/Buttons/CopyTiddler/Caption}}
+description: {{$:/language/Buttons/CopyTiddler/Hint}}
+
+\define copyTiddlerButton()
+\whitespace trim
```
pmario, line 7: If `\whitespace trim` is active you can use indentation using tabs

core/ui/ViewToolbar/copy-tiddler.tid
```diff
@@ -0,0 +1,23 @@
...
+<$button tooltip={{$:/language/Buttons/CopyTiddler/Hint}} aria-label={{$:/language/Buttons/CopyTiddler/Caption}} class=<<tv-config-toolbar-class>>>
+<$action-clipboard $tiddler=<<currentTiddler>> $text="hello"/>
```
pmario, line 9: actions should never be activated inside the button body. It causes refresh problems. 
Best practice is: `<$button actions=<<your-action-procedure>>...>` instead.

core/ui/ViewToolbar/copy-tiddler.tid
```diff
@@ -0,0 +1,23 @@
+title: $:/core/ui/Buttons/copy-tiddler
+tags: $:/tags/ViewToolbar
+caption: {{$:/core/images/copy-clipboard}} {{$:/language/Buttons/CopyTiddler/Caption}}
+description: {{$:/language/Buttons/CopyTiddler/Hint}}
+
+\define copyTiddlerButton()
```
pmario, line 6: IMO `\precedure` should be used for new functionality now

core/wiki/macros/copy-to-clipboard.tid
```diff
@@ -1,6 +1,8 @@
 title: $:/core/macros/copy-to-clipboard
 tags: $:/tags/Macro
 
+<!-- These are possibly now deprecated in prefrence to the action widget -->
```
pmario, line 4: We need to decide if they are deprecated. If we decide they are we need to add some info to the docs and provide new examples with the new "best practice"

core/modules/widgets/action-clipboard.js
```diff
@@ -0,0 +1,83 @@
...
+    this.copyText = this.getAttribute("$text", null);
+    this.tiddlerFormat = this.getAttribute("$format", null);
+};
+
+/*
+Refresh the widget by ensuring our attributes are up to date
+*/
+ClipboardWidget.prototype.refresh = function(changedTiddlers) {
+    var changedAttributes = this.computeAttributes();
+	if(Object.keys(changedAttributes).length) {
```
pmario, line 48: TW has a utility function to count changes in an object `$tw.utils.count(changedAttributes)`, which is basically the same thing. But for consistency reasons it should be used that way. eg: 

https://github.com/Jermolene/TiddlyWiki5/blob/779ac28bd02defa158af470a1ce3e3f0125cba38/core/modules/widgets/let.js#L87 or https://github.com/Jermolene/TiddlyWiki5/blob/779ac28bd02defa158af470a1ce3e3f0125cba38/core/modules/widgets/element.js#L89-L91

saqimtiaz:
> I know that message exists, but it was lacking quite a few features

Can you please clarify what features are missing?

You should be able to wikify the tiddler using the appropriate template and then copy that resultant text to the clipboard with the facilities the core already provides.

GameDungeon:
@pmario, the reason it currently doesn't work is the PR is a draft. The feature isn't done yet.

> You should be able to wikify the tiddler using the appropriate template and then copy that resultant text to the clipboard with the facilities the core already provides

The idea is to literally copy the tiddler data to the clipboard. Not as text, but as data.

See: https://github.com/Jermolene/TiddlyWiki5/pull/7215

core/modules/widgets/action-clipboard.js
GameDungeon: This line was directly stolen from another action widget, so if it's incorrect here it's incorrect there too.

saqimtiaz:
@GameDungeon the core endeavours to provide low level affordances (widgets, filters, macros etc) for basic operations that can be combined via wikitext to address more complex needs. The `tm-copy-to-clipboard` message can be used to copy any text based data (that is a String) to the clipboard, whether it be the text field of a tiddler, the JSON representation of the entire tiddler, or any other format. It is up to the user to decide what to pass to the message to be copied to the clipboard.

The PR summary makes it difficult to understand what this new action widget provides in addition. After looking through your code I assume that you want to be able to set the **data type** when copying to the clipboard. I would suggest exploring possibilities for extending the `tm-copy-to-clipboard` message handling to support setting the data type(s).  However, I am unsure that using `clipboardData.setData()` will work as that method is usually only available within copy and cut event handlers.


---

In case it is of interest, here is an example of creating a button that copies the JSON representation of the tiddler HelloThere to the clipboard:
```
\procedure copyactions()
<$wikify name="dataToCopy" text="""<$macrocall $name="jsontiddlers" filter="HelloThere" $output="text/raw"/>""">
<$action-sendmessage $message="tm-copy-to-clipboard" $param=<<dataToCopy>>/>
</$wikify>
\end

<$button actions=<<copyactions>>>copy</$button>

```

pmario:
@saqimtiaz I think the intended usecase is to send a data-URL to the clipboard, that contains a tiddler representation. If CTRL-V is used in a TW it imports 1 tiddler. 

With the TW drag and drop functionality the content of the data-URL is defined when the drag starts. We do have no user exposed functions, that create the same possibilities. 

Eg: If a tag-pill is drag and dropped all the tiddlers listed in the tag-pill can ex-/imported. IMO there should be a possibility for action-widgets to reproduce the same behaviour as we have with d&d. This would be a big advantage for mobile devices, where drag and drop is hard to accomplish. 

I'm not sure if all the possibilities can be done with tm-copy-to-clipboard message.

pmario:
It looks relatively good to me. There are 2 things left for the basic function

1. Feedback, that the tiddler has been copied to the clipboard
2. I'm not sure if the `text/plain` should be JSON - I would like an option as described below


**Add 1:**
At your [$:/core/ui/Buttons/copy-tiddler](https://tiddlywiki5-git-fork-gamedungeon-copytoclipboard-jermolene.vercel.app/#%24%3A%2Fcore%2Fui%2FButtons%2Fcopy-tiddler)
You can add the following line to the `\procedure copyTiddlerAction()` after the clipboard action
`<$action-sendmessage $message="tm-notify" $param="$:/language/Notifications/CopiedToClipboard/Succeeded"/>`

**Add 2:**
I would like this to be an option. Eg: "e can quickly copy paste tiddler content to Talk that way. But it may be configurable and JSON should be the default. 

```
title: HelloThere
tags: TableOfContents

lorem ipsum 
```

Just some thoughts. 

---

@GameDungeon Here in the repo you have the possibility to click the <kbd>Resolve</kbd>button at my code inline comments. So they will be folded for everyone who reads the thread and it gives you some feedback if you missed something.

core/modules/widgets/action-clipboard.js
GameDungeon: Should I fix this for the action widget it was stolen from as well?

core/modules/widgets/action-clipboard.js
GameDungeon: There are better ways to make this work (as this way of triggering the event is technically deprecated), but I'm not sure the overall browser support is high enough for me to use them https://caniuse.com/async-clipboard

core/modules/widgets/action-clipboard.js
pmario: For consistency reasons I thin yes. Which widget was it?

GameDungeon:
A "feature" of tiddlywiki I dislike negatively impacts the copy-paste functionality quite a bit. Unlike most applications that read from the single best datatype from the clipboard, tiddlywiki imports everything on the clipboard. So both the plain text and tiddler versions of the copy are imported.

core/modules/widgets/action-clipboard.js
GameDungeon: the action-log widget

GameDungeon:
I notice that it renders last in the view toolbar. I know how to change this in my own wiki, but how would I change the default?

pmario:
> Unlike most applications that read from the single best datatype from the clipboard, tiddlywiki imports everything on the clipboard. 

Yes that's a feature. It's because of backwards compatibility. There was a bug in the drag & drop -> $:/Import function that did import **text/plain for FireFox** (that I use all of the time) and **text/html in Chrome-like and Safari** browsers. 

I did fix that bug not long ago and to not annoy everyone I did allow to import both of them. I would have done `text/plain` only. But sometimes it is possible users want to import the HTML source of something and not the text/plain representation of HTML content. 

So your opinion just proves, that designers cannot decide right. There will always be someone how does not like it. 

I could have made it configurable. But since most users use Chrome-like browsers, I'm pretty sure text/html would have been the default. But why should I shoot myself in my own foot? 

That's why both of them are selected by default and users now have the possibility to decide. -- May be we are able to catch that info in the future and display a message: "Remember this setting" or something similar. ... May be

pmario:
You can use the `list-before` field and give it one of the following titles

- $:/core/ui/Buttons/more-tiddler-actions
- $:/core/ui/Buttons/info
- $:/core/ui/Buttons/new-here
- $:/core/ui/Buttons/new-journal-here
- $:/core/ui/Buttons/clone
- $:/core/ui/Buttons/export-tiddler
- $:/core/ui/Buttons/delete
- $:/core/ui/Buttons/permalink
- $:/core/ui/Buttons/permaview
- $:/core/ui/Buttons/open-window
- $:/core/ui/Buttons/close-others
- $:/core/ui/Buttons/close
- $:/core/ui/Buttons/fold-others
- $:/core/ui/Buttons/fold
- $:/core/ui/Buttons/copy-tiddler
- $:/core/ui/Buttons/fold

image

GameDungeon:
> You can use the `list-before` field and give it one of the following titles

I thought about that, but I noticed that not a single one of the other toolbar items did that.

pmario:
>Should that be a separate PR, or bundled with this one? Probably seperate.

Definitely a separated PR. 

But be aware. There are a 1000 traps in the $:/Import mechanism. It goes through 3 different import hooks. If it did hit the $:/Import tiddler it's too late ;)

GameDungeon:
Currently, I am using an old, now deprecated method to copy stuff to the clipboard. This is the new API that replaced it. https://caniuse.com/async-clipboard (Please note that the browsers marked partial support fully support writing, but have not fully implemented reading outside of a paste event for security reasons) This API is preferable for several reasons, but I will let @Jermolene decide if the support is high enough that it should be used over the old method.

pmario:
I think you should also merge the current master-branch into this branch. Since for version v5.3.1 we have fixed the possibility to copy stuff from Edge to Firefox, which does not work with this version. So it won't be possible to fully test this PR without the changes to the master branch

GameDungeon:
It seems firefox generally dislikes people messing with the clipboard at all. I don't think it's possible to copy tiddler data to the clipboard on firefox (with the exception of users changing a setting to allow it). It is still possible to copy text and images. This is regardless of the API used. 

The Async Clipboard API has been supported on every large browser for 3+ (and in most cases 5) years now (Other than IE obviously, but I'm not sure what the current stance is on that) The browsers that have partial support either support writing fully (what needs to be used here), or happen to be firefox which is covered above. 

The reason I want to use it is it's deprecated and I dislike using deprecated functions. Additionally, the other version is really clunky and slow.

PR #7691, opened 2023-08-22
Add `$if`...`$then`...`$elseif`...`$else` widget

### LINKED ISSUE #7570: [IDEA] Conditional Blocks (If Statements)
Having conditional blocks (blocks of wikitext that are shown or hidden based on a filter) (if elseif and else) is an important part of making view templates and other advanced wikitext functions.

This is currently done with the list widget, but I find that very unintuitive. Additionally, it makes else and elseif functionality very verbose to write out, with each requiring us to list every previous check (unless there is some way to do it that I'm unaware of). Additionally, I struggle to believe that using a list like that is especially performant.  

I'm not quite sure of the ideal syntax of these blocks, and that is something for discussion. 

To throw out something, but I'm sure people can suggest something better:

```
@if "[filter]"
    content
@elseif "[filter]"
    content
@else
    content
@endif
```

rmunn:
The `<$reveal>` widget is useful when you want to show or hide some text (or widgets) based on a state tiddler, but what if you want to show or hide something based on the results of a filter? Enter the `<$if>` widget.

Usage example:

```
\define num() 2

<$if filter="[<num>compare:number:gt[3]]">
<$then><<num>> is greater than 3</$then>
<$elseif filter="[<num>compare:number:gt[2]]"><<num>> is greater than 2</$elseif>
<$elseif filter="[<num>compare:number:gt[1]]"><<num>> is greater than 1</$elseif>
<$elseif filter="[<num>compare:number:gt[0]]"><<num>> is greater than 0</$elseif>
<$else><<num>> is negative or zero</$else>
</$if>
```

Note that the `$if` widget is short-circuiting: the first filter to evaluate to a true (non-empty) result will cause that child widget to be selected, and the contents of all other child widgets will be ignored. If there were `<$log>` widgets in every single one of those `<$then>`, `<$elseif>`, and `<$else>` blocks, only one `<$log>` widget would have been rendered. To achieve this result, the `<$then>`, `<$elseif>`, and `<$else>` widgets are actually *pseudo-widgets*: they do not have separate widget definitions and cannot exist outside an `<$if>` widget. Instead, the `<$if>` widget works at the parse tree level, choosing just one child pseudo-widget to render and ignoring all the others. So if the `<$else>` block contains a widget that's particularly expensive to render, that cost will only be paid if the `<$if>` condition evaluates to false (i.e., its filter returns an empty result).

Resolves #7570.

### DIFF
--- a/core/modules/widgets/if.js
+++ b/core/modules/widgets/if.js
@@ -16,7 +16,6 @@ var Widget = require("$:/core/modules/widgets/widget.js").widget;
 
 var IfWidget = function(parseTreeNode,options) {
 	this.initialise(parseTreeNode,options);
-	this.conditions = [];
 };
 
 /*
@@ -25,91 +24,128 @@ Inherit from the base widget class
 IfWidget.prototype = new Widget();
 
 /*
-Render this widget into the DOM
+Prepare child then and/or else widgets before they are created
 */
-IfWidget.prototype.render = function(parent,nextSibling) {
-	this.parentDomNode = parent;
-	this.computeAttributes();
-	this.execute();
-	this.renderChildren(parent,nextSibling);
-};
-
-IfWidget.prototype.safelyComputeAttribute = function(attribute,defaultValue) {
-	return attribute ? this.computeAttribute(attribute) : defaultValue;
-}
-
-IfWidget.prototype.computeConditions = function() {
-	var self = this;
-	this.filterString = this.safelyComputeAttribute(this.parseTreeNode.attributes.filter,"");
-	this.conditions = [{filter: this.filterString, parseTree: undefined}];
+IfWidget.prototype.preprocessChildNodes = function() {
+	var filter = this.parseTreeNode.attributes.filter;
 	$tw.utils.each(this.parseTreeNode.children,function(childNode) {
 		if(childNode.type === "then") {
-			self.conditions[0].parseTree = childNode;
-		} else if(childNode.type === "elseif") {
-			var childFilter = self.safelyComputeAttribute(childNode.attributes && childNode.attributes.filter,"")
-			self.conditions.push({filter: childFilter, parseTree: childNode});
+			if(filter) {
+				$tw.utils.addAttributeToParseTreeNode(childNode,filter);
+			}
 		} else if(childNode.type === "else") {
-			self.conditions.push({filter: true, parseTree: childNode});
+			$tw.utils.addAttributeToParseTreeNode(childNode,"filter","yes");
 		}
+		// elseif widgets have their own filter and don't need to be pre-processed
 	});
-}
-
-IfWidget.prototype.findFirstCondition = function() {
-	var self = this;
-	var resultIdx = -1;
-	$tw.utils.each(this.conditions,function(condition,idx) {
-		var filter = condition && condition.filter || "";
-		if(filter === true) {
-			resultIdx = idx;
-			return false; // Short-circuit
-		}
-		var filterResult = self.wiki.filterTiddlers(filter,self) || [];
-		if(filterResult && filterResult.length) {
-			resultIdx = idx;
-			return false; // Short-circuit
-		}
-	})
-	return resultIdx;
 }
 
 /*
 Compute the internal state of the widget
 */
 IfWidget.prototype.execute = function() {
-	// Get our parameters
-	this.computeConditions();
-	// Choose the appropriate child widget to construct
-	this.activeConditionIdx = this.findFirstCondition();
-	if(this.activeConditionIdx < 0) {
-		this.activeChildNodes = [];
-	} else {
-		var condition = this.conditions[this.activeConditionIdx];
-		this.activeChildNodes = condition && condition.parseTree && condition.parseTree.children;
-	}
+	// Set up then and/or else filters in parse tree
+	this.preprocessChildNodes();
 	// Make the child widgets
-	this.makeChildWidgets(this.activeChildNodes || []);
+	this.makeChildWidgets();
+	// Ensure each condition widget knows its next sibling
+	this.linkConditions();
 };
 
-/*
-Selectively refreshes the widget if needed. Returns true if the widget or any of its children needed re-rendering
-*/
-IfWidget.prototype.refresh = function(changedTiddlers) {
-	var previousConditions = this.conditions;
-	this.computeConditions();
-	var changed = previousConditions.length !== this.conditions.length;
-	var previousActiveIdx = this.activeConditionIdx;
-	this.activeConditionIdx = this.findFirstCondition();
-	changed = this.activeConditionIdx !== previousActiveIdx || changed;
-	// Refresh if an attribute has changed, or the type associated with the target tiddler has changed
-	if(changed) {
-		this.refreshSelf();
-		return true;
-	} else {
-		return this.refreshChildren(changedTiddlers);
-	}
+IfWidget.prototype.linkConditions = function() {
+	var self = this,
+		conditions = [];
+	$tw.utils.each(this.children,function(childWidget) {
+		if(self.isConditionChild(childWidget)) {
+			conditions.push(childWidget);
+		}
+	});
+	$tw.utils.each(conditions,function(childWidget,idx) {
+		if(idx > 0) {
+			conditions[idx-1].nextCondition = childWidget;
+		}
+	});
+	// Last one will have nextCondition still undefined, which is what we want
+}
+
+IfWidget.prototype.isConditionChild = function(childWidget) {
+	var type = childWidget.parseTreeNode.type;
+	return (type === "then" || type === "else" || type === "elseif");
+}
+
+var ConditionWidget = function(parseTreeNode,options) {
+	this.oldShouldRender = false;
+	this.shouldRender = false;
+	this.shortCircuited = false;
+	this.nextCondition = undefined;
+	this.initialise(parseTreeNode,options);
 };
+ConditionWidget.prototype = new Widget();
+
+ConditionWidget.prototype.shortCircuit = function() {
+	this.oldShouldRender = this.shouldRender;
+	this.shouldRender = false;
+	this.shortCircuited = true;
+	if(this.nextCondition) {
+		this.nextCondition.shortCircuit();
+	}
+}
+
+ConditionWidget.prototype.calculateFilter = function() {
+	if(this.shortCircuited) {
+		return;
+	}
+	this.oldShouldRender = this.shouldRender;
+	var result = this.wiki.filterTiddlers(this.filter,this);
+	this.shouldRender = !!(result && result.length);
+	if(this.shouldRender && this.nextCondition) {
+		this.nextCondition.shortCircuit();
+	}
+}
+
+ConditionWidget.prototype.render = function(parent,nextSibling) {
+	this.parentDomNode = parent;
+	this.computeAttributes();
+	this.execute();
+	this.renderChildren(parent,nextSibling);
+};
+
+ConditionWidget.prototype.execute = function() {
+	this.filter = this.getAttribute("filter","");
+	this.calculateFilter();
+	this.shortCircuited = false;
+	var childNodes = this.shouldRender ? this.parseTreeNode.children : [];
+	this.makeChildWidgets(childNodes);
+};
+
+ConditionWidget.prototype.refresh = function(changedTiddlers) {
+	var changedAttributes = this.computeAttributes();
+	if(changedAttributes.filter) {
+		this.filter = this.getAttribute("filter","");
+	}
+	this.calculateFilter();
+	this.shortCircuited = false;
+	if(this.oldShouldRender === this.shouldRender) {
+		return this.refreshChildren(changedTiddlers);
+	} else {
+		if(this.shouldRender) {
+			// Went from false -> true, so we didn't have child widgets before
+			this.makeChildWidgets();
+			// Re-render self, but don't re-compute filter as it might be expensive and we already know the result
+			this.shortCircuited = true;
+			this.refreshSelf();
+			return true;
+		} else {
+			// Went from true -> false, so skip refreshSelf as there's no need for the this.findNextSiblingDomNode() step
+			this.removeChildDomNodes();
+			return true;
+		}
+	}
+}
 
 exports.if = IfWidget;
-// exports['if'] = IfWidget; // Uncomment if the above is a syntax error
+exports.then = ConditionWidget;
+exports.else = ConditionWidget;
+exports.elseif = ConditionWidget;
 
 })();
--- a/core/modules/widgets/if.js
+++ b/core/modules/widgets/if.js
@@ -1,0 +1,115 @@
+/*\
+title: $:/core/modules/widgets/if.js
+type: application/javascript
+module-type: widget
+
+If-then-elseif-else widget
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+var Widget = require("$:/core/modules/widgets/widget.js").widget;
+
+var IfWidget = function(parseTreeNode,options) {
+	this.initialise(parseTreeNode,options);
+	this.conditions = [];
+};
+
+/*
+Inherit from the base widget class
+*/
+IfWidget.prototype = new Widget();
+
+/*
+Render this widget into the DOM
+*/
+IfWidget.prototype.render = function(parent,nextSibling) {
+	this.parentDomNode = parent;
+	this.computeAttributes();
+	this.execute();
+	this.renderChildren(parent,nextSibling);
+};
+
+IfWidget.prototype.safelyComputeAttribute = function(attribute,defaultValue) {
+	return attribute ? this.computeAttribute(attribute) : defaultValue;
+}
+
+IfWidget.prototype.computeConditions = function() {
+	var self = this;
+	this.filterString = this.safelyComputeAttribute(this.parseTreeNode.attributes.filter,"");
+	this.conditions = [{filter: this.filterString, parseTree: undefined}];
+	$tw.utils.each(this.parseTreeNode.children,function(childNode) {
+		if(childNode.type === "then") {
+			self.conditions[0].parseTree = childNode;
+		} else if(childNode.type === "elseif") {
+			var childFilter = self.safelyComputeAttribute(childNode.attributes && childNode.attributes.filter,"")
+			self.conditions.push({filter: childFilter, parseTree: childNode});
+		} else if(childNode.type === "else") {
+			self.conditions.push({filter: true, parseTree: childNode});
+		}
+	});
+}
+
+IfWidget.prototype.findFirstCondition = function() {
+	var self = this;
+	var resultIdx = -1;
+	$tw.utils.each(this.conditions,function(condition,idx) {
+		var filter = condition && condition.filter || "";
+		if(filter === true) {
+			resultIdx = idx;
+			return false; // Short-circuit
+		}
+		var filterResult = self.wiki.filterTiddlers(filter,self) || [];
+		if(filterResult && filterResult.length) {
+			resultIdx = idx;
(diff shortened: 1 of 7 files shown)


rmunn:
Coming soon: unit tests.

Open design question: should the `<$then>`, `<$else>`, and `<$elseif>` widgets be real widgets with their own definitions (in `widgets/then.js` and so on) rather than pseudo-widgets that are never instantiated? Or is the parse-tree-based approach I took the best approach to this problem?

Jermolene:
Thanks @rmunn, great to see this take shape so quickly. It would be helpful to mark this PR as a draft until things get a bit further.

> Open design question: should the `<$then>`, `<$else>`, and `<$elseif>` widgets be real widgets with their own definitions (in `widgets/then.js` and so on) rather than pseudo-widgets that are never instantiated? Or is the parse-tree-based approach I took the best approach to this problem?

In the case of the `<$innerwiki>` widget, the `<$data>` that is used within it is a real widget. However, in that case we currently render the entire child widget tree, so we need the real widget so that it can live in the widget tree as well as the parse tree. In this case they could indeed be pseudo-widgets, with no corresponding real widget, but I think that we should define them as real widgets, if nothing else for the benefit of tooling that needs to know the core widgets.

A very minor point, but it's worth putting the `<<.from-version "5.3.2">>` macro into the documentation.

rmunn:
> Thanks @rmunn, great to see this take shape so quickly. It would be helpful to mark this PR as a draft until things get a bit further.

Done.

> In the case of the `<$innerwiki>` widget, the `<$data>` that is used within it is a real widget. However, in that case we currently render the entire child widget tree, so we need the real widget so that it can live in the widget tree as well as the parse tree. In this case they could indeed be pseudo-widgets, with no corresponding real widget, but I think that we should define them as real widgets, if nothing else for the benefit of tooling that needs to know the core widgets.

I'll work on unit tests first, so that I know that rewriting the code to make `<$then>` _et al_ real widgets doesn't introduce bugs. Then once I have a decent set of unit tests, I'll make `<$then>`, `<$else>`, and `<$elseif>` real widgets. I'll try to write `<$else>` in such a way that it could be used as part of other widgets as well, such as the `<$switch>`...`<$case>` widget I plan to write. I was planning to have a `<$default>` pseudo-widget, but if there's a real `<$else>` widget I might as well use it instead in `<$switch>`.

> A very minor point, but it's worth putting the `<<.from-version "5.3.2">>` macro into the documentation.

Done.

rmunn:
I've created very basic documentation for the `<$then>`, `<$else>`, and `<$elseif>` widgets, basically pointing to the IfWidget tiddler for all details. @Jermolene, if you think those widgets need their own documentation with an attribute list, etc., despite not being allowed to exist outside the `<$if>` widget, let me know and I'll flesh out those documentation tiddlers.

AnthonyMuscio:
@rmunn Robin,

I think this is a great initiative, but I would like to suggest the usability of this can be improved and handle a larger number of cases. The danger of implementing it prematurely is the following widgets;
> `<$if>`,`<$then>`, `<$else>`, and `<$elseif>`

These widget names will end up being used, and through backwards compatibility, it would be a challenge to introduce additional features using the same widget names in the future.

Given substantial improvements and increased possibilities I have found more than one effective way already to handle if and case statements more in line with existing features. I can share them in time but want to raise this now, so we do not rush this implementation of #7691 

The desire to add a fundamental features such as these to TiddlyWiki is clear, but it potentially has far reaching consequences for usability and understandability. Unfortunately this mean such a solution is also somewhat complex to discuss, even if the end result appears simple.

I have found a way to use procedures and can also see how to use existing or custom widgets which is quite elegant so a ''if then else else-if and case'' structures can already be created and we can make use of them until we have a refined version of your proposal.

### "I will Put my code where my mouth is"?

One example here, past it into tiddlywiki.com

```
\procedure if-then-proc(filter)
\procedure true() Show true '<<input-params>>'
\procedure false() Show False '<<input-params>>'
\procedure empty() Show Empty '<<input-params>>'
<$parameters filter="[all[current]]" $params=input-params>
<$list filter="""[subfilter<filter>] :filter[getvariable<currentTiddler>]""" emptyMessage="missing procedure in filter" variable=procedure>
<$macrocall $name=<<procedure>>/>
</$list>
</$parameters>
\end if-then-proc

# <<if-then-proc "[all[current]prefix[Case]then[true]else[false]]" more:"value">>
# <<if-then-proc filter:"[all[current]test[1]then[true]] [all[current]test[0]then[false]] :else[[empty]]" more:"value2">>
```
* This shows how a single filter can be used, the value returned in the name of a procedure internal or external to the if-then-proc procedure is used to generate the alternative outputs, and could be overridden by redefined procedures.
* I am still working on passing all but the first parameter (the conditions) to the "sub procedures".
* I use $macrocall for simplicity rather than `<$transclude $variable`
* This can be used as a case statement as well.
* I am confident we can put this in a custom widget and possibly also;
  * place the "sub procedures" in $fill widgets
  * detect the names of $fill widgets and respond accordingly

AnthonyMuscio:
@rmunn Robin,

I think this is a great initiative, but I would like to suggest the usability of this can be improved and handle a larger number of cases. The danger of implementing it prematurely is the following widgets;
> `<$if>`,`<$then>`, `<$else>`, and `<$elseif>`

These widget names will end up being used, and through backwards compatibility, it would be a challenge to introduce additional features using the same widget names in the future.

Given substantial improvements and increased possibilities I have found more than one effective way already to handle if and case statements more in line with existing features. I can share them in time but want to raise this now, so we do not rush this implementation of #7691 

The desire to add a fundamental features such as these to TiddlyWiki is clear, but it potentially has far reaching consequences for usability and understandability. Unfortunately this means such a solution is also somewhat complex to discuss, even if the end result appears simple.

For Example: I have found a way to use **procedures** and can also see how to use **existing or custom widgets** which is quite elegant so a ''if then else else-if and case'' structures can already be created and we can make use of them until we have a refined version of your proposal.

### "I will Put my code where my mouth is"?

One example here, paste it into tiddlywiki.com

```
\procedure if-then-proc(filter)
   \procedure true() Show true '<<input-params>>'
   \procedure false() Show False '<<input-params>>'
   \procedure empty() Show Empty '<<input-params>>'

<$parameters filter="[all[current]]" $params=input-params>
<$list filter="""[subfilter<filter>] :filter[getvariable<currentTiddler>]""" emptyMessage="missing procedure in filter" variable=procedure>
<$macrocall $name=<<procedure>>/>
</$list>
</$parameters>
\end if-then-proc

# <<if-then-proc "[all[current]prefix[Case]then[true]else[false]]" more:"value">>
# <<if-then-proc filter:"[all[current]test[1]then[true]] [all[current]test[0]then[false]] :else[[empty]]" more:"value2">>
```
* This shows how a single filter can be used, the value returned is the name of a procedure internal (or external below) to the if-then-proc procedure and is used to generate the alternative outputs, and could be overridden by redefined procedures.
* I am still working on passing all but the first parameters, named or otherwise, to the "sub procedures".
* I use $macrocall for simplicity rather than `<$transclude $variable` they work the same.
* This can be used as a case statement as well.
* I am confident we can put this in a custom widget and possibly also
  * place the "sub procedures" in $fill widgets

On reflection I think it best to keep the true/false and empty procedures outside of if-then-proc because;
* You can redefine them locally with `\procedure` or `$let` for example.
* Overriding the default in a global procedure

```
\procedure true() Show true '<<input-params>>'
\procedure false() Show False '<<input-params>>'
\procedure empty() Show Empty '<<input-params>>'

\procedure if-then-proc(filter)
<$parameters filter="[all[current]]" $params=input-params>
<$list filter="""[subfilter<filter>] :filter[getvariable<currentTiddler>]""" emptyMessage="missing procedure in filter" variable=procedure>
<$macrocall $name=<<procedure>>/>
</$list>
</$parameters>
\end if-then-proc
```

pmario:
@AnthonyMuscio We could do if-then-else constructions using 2 lists and filters already. 

The point of the if-then-elseif-else widgets is, that they are easy to read and understand by users. 

Complex list-filter constructions are not easy to understand. Your suggested construction is interesting, but also not easy to understand and impossible to write for new users. I would not want it to write myself. I would need to copy paste and modify it. 

What users want:

```
<$if filter="something"><$then>

Some plain text. 
<$else>

Other  text
</$if>
```

Which does not need any explanation and can be written by everyone.

AnthonyMuscio:
@pmario thanks for the feedback
 
> Which does not need any explanation and can be written by everyone.

I think your example may need to be;
```
<$if filter="condition">
<$then>then</$then>
<$else>else</$else>
</$if>
```

I was talking to a code aware environment and did not give an explanation here, and it is designed to illustrate the possibilities. In this case we can provide 1 to multiple conditions and result, all in a single filter, with the result using any name, I hope you can use your imagination and see the same code pattern implemented in the if widget;
* The above could be the default and assume the existence of the $then and $else;
  * as if you wrote the above which is equivalent to the following;
```
<$if filter="[condition..then[then]] ~[[else]]">
<$then>then</$then>
<$else>else</$else>
</$if>
```
However Imagin the ability to name any number of results;

```
<$if filter="[condition..then[then]] :else[condition]then[also]]" variable=result>
<$then>then</$then>
<$else>else is no value</$else>
<$also>else</$also>
</$if>
```

My key point here is using widget names such as "if else then elseif" is using valuable words and we could make a more expansive solution to handle other code patterns which are currently quite verbose in tiddlywiki.
* If a number of conditions and results equivalent to a "case statement"
* Allow full nesting.
* The section displayed will have the name "result"

I for shadow the possibility of more power and simpler to read code and have only hinted as to where I am going. I am seriously concerned it would be premature to implement the current proposal in the core, and using such important widget names.

AnthonyMuscio:
@pmario thanks for the feedback
 
> Which does not need any explanation and can be written by everyone.

I think your example may need to be;
```
<$if filter="condition">
<$then>then</$then>
<$else>else</$else>
</$if>
```

I was talking to a code aware environment and did not give an explanation here, and it is designed to illustrate the possibilities. In this case we can provide 1 to multiple conditions and result, all in a single filter, with the result using any name, I hope you can use your imagination and see the same code pattern implemented in the if widget;
* The above could be the default and assume the existence of the $then and $else;
  * as if you wrote the above which is equivalent to the following;
```
<$if filter="[condition..then[then]] ~[[else]]">
<$then>then</$then>
<$else>else</$else>
</$if>
```
However Imagin the ability to name any number of results;

```
<$if filter="[condition..then[then]] :else[condition]then[also]]" variable=result>
<$then>then</$then>
<$else>else is no value</$else>
<$also>else</$also>
</$if>
```

My key point here is using widget names such as "if else then elseif" is using valuable words and we could make a more expansive solution to handle other code patterns which are currently quite verbose in tiddlywiki.
* If a number of conditions and results equivalent to a "case statement"
* Allow full nesting.
* The section displayed will have the name "result"

I for shadow the possibility of more power and simpler to read code and have only hinted as to where I am going. I am seriously concerned it would be premature to implement the current proposal in the core, and using such importiant widget names.

rmunn:
@AnthonyMuscio - About your idea, I have some bad news and some good news. The bad news is that I won't be writing the `<$if>` widget the way you suggest. The good news is that you'll still be able to use `<$then>` and `<$else>` in your idea, if/when you write it in the future; only the `<$if>` name will be taken. More details below.

The reason I'm going to do `<$if>` the way I'm doing it is because that maps almost exactly one-to-one with the way if/then/else works in nearly every programming language out there, so this is what anyone will expect when they learn that TW has an `<$if>` widget. Your idea is a good one, but it should not be called `<$if>`. Perhaps `<$condition>` would be a good name for it.

Now, why did I say that you'll still be able to use `<$then>` and `<$else>` in your idea? Because the only way to write your widget (which I'm going to call `<$condition>`) is going to be to use what I call pseudo-widgets. If you look at how I first implemented `<$if>` in https://github.com/Jermolene/TiddlyWiki5/pull/7691/commits/b8a32019c948f6f6850746f0f371fe8c770a17a7, you might notice that there's no actual `<$then>` or `<$else>` widgets in that code! Instead, I have the `<$if>` widget analyse the parse tree nodes and decide which child node(s) to render based on its analysis. So the `<$then>` widget isn't actually created as a real widget at all; it remains a parse tree node. It's a pseudo-widget. And the only way to write your `<$condition>` widget is going to be to use pseudo-widgets.

Why is that? Because TiddlyWiki's widget system requires that the names of all possible widgets be known to the TW core *at startup time*. The startup code looks through the existing modules, finds all modules of type "widget", and compiles a list of known widgets. Your idea, on the other hand, needs widgets whose names are produced by a filter expression, meaning that the names won't be known until the filter expression is evaluated, long past wiki startup. Therefore you can't use conventional, "real" widgets for the branches of your `<$condition>` widget; you'll need to use the same technique I used in https://github.com/Jermolene/TiddlyWiki5/pull/7691/commits/b8a32019c948f6f6850746f0f371fe8c770a17a7 and analyse the parse tree nodes, selecting one (or possibly none) to render.

But the good news is that if you're analysing the parse tree nodes like that, you have complete freedom to override TW's widgets in the Javascript code, by simply not calling `makeChildWidgets`. That way your filters can still have a result called `then`, and your `<$condition>` widget would look through the parse tree nodes for `node.type === "then"`. As long as you handle it when it's a parse tree node, the fact that there's already a `<$then>` widget won't matter, as your parse tree node handling will never create the `<$then>` widget in the first place. And since parse tree node handling is the only way to implement your idea *anyway*, there's no downside.

So I'm going to go ahead and implement the `<$if>`, `<$then>`, `<$elseif>` and `<$else>` widgets as per my original plan, and that won't prevent you from using those names (except for `<$if>`) in the future to implement your idea.

AnthonyMuscio:
FYI: I can already make a custom widget that would be called like this;
* In fact I do below.

```
<$if filter="[condition..then[then]] ~[[else]]">
   <$fill $name=then>then</$then>
   <$fill $name=else>else</$else>
</$if>
```
I appreciate your work on this however this approach can be extended to multiple outcomes and parameters. but it does not have any of the limitations you mention.

Imagin if we had an alternate way to shorten the $fill widgets;
* This is a change that would improve tiddlywiki for both our needs.

```
<$if filter="[condition..then[then]] ~[[else]]">
   <$!then>then</$!then>
   <$!else>else</$!else>
</$if>
```
That is `<$fill $name=then>then</$fill>` could be rewritten as `<$!then>then</$!then>` making it mode readable.
* You would get what you want, and I would beable to use my recently discovered methods.


But I can already do this, it works
```
\widget $if.then()
<$parameters filter="[all[current]]" $params=all-params>
<$list filter=<<filter>> variable=slot>
   <$slot $name=<<slot>>/>
</$list>
</$parameters>
\end $if.then

<$if.then filter="[all[current]has[switch]then[then]] ~[[else]]">
   <$fill $name=then>then</$fill>
   <$fill $name=else>else</$fill>
</$if.then>
```
However this can be immediately extended to a case widget with different results in the filter matching the name of a $fill widget.
* However I can't nest them
* I need to use a "." in the name

rmunn:
@Jermolene - This PR is ready to be reviewed and merged now. The code is in its final form, barring any changes you ask me to make, and the documentation is as good as I can make it. Please let me know anything you'd like me to change; once this goes in, I may work on a `<$switch>`...`<$case>` widget that works on similar principles. Though I'd plan to finish all my other outstanding PRs first before I tackle that one.

kookma:
One quesion:

In example below

```
<$if filter="[all[current]tag[something]]">
Hello
<$then> Yes</$then>
</$if>
```

The word `Hello` is displayed irresectable of filter result! Is this by design?
Does it make sense to ignore the body and only show $else/$elseif content if available? In other words when filter is empty, TW looks for $else/$elseif if nothing found, just show nothing.

Jermolene:
Thanks @rmunn this is excellent work.

I do still have reservations that the syntax we've ended up with is clumsy to type; the need to repeat the `filter=` part seems particularly egregious. It is a lot of typing.

My other concern is the one raised in the [Talk thread](https://talk.tiddlywiki.org/t/proposed-if-widget/7882?u=jeremyruston): that this PR is incompatible with Evan Balster's Conditional plugin (see  a copy of it [here](http://chronicles.wiki/TiddlyWikiFormula/#%24%3A%2Fplugins%2Febalster%2Fcondition)). I think we have to take a broad view of backwards compatibility within the TiddlyWiki ecosystem, so that core updates should not break the behaviour of popular existing plugins.

The solution that I favour which resolves both those issues is to implement if/then/else as parser rules that generate list widgets. I don't think they can fit into any of the established syntax patterns used in TiddlyWiki, but rather should introduce a new category of syntax constructions that I'm currently calling "directives".

I'm working on a PR which uses the following syntax:

```
This is a {% if [<something>match[one]] %}Elephant{% else %}Crocodile{% endif %}, I think.
```

I intend to publish the PR later today so that we can evaluate the two approaches.

rmunn:
I just saw that [Talk thread](https://talk.tiddlywiki.org/t/proposed-if-widget/7882) myself. I've marked this PR as draft again while I re-evaluate the syntax.

Jermolene:
Thanks @rmunn I've pushed an initial implementation of a parse rule for a conditional shortcut syntax that expands to the list widget behind the scenes. See #7710.

kookma:
> This is a {% if [<something>match[one]] %}Elephant{% else %}Crocodile{% endif %}, I think.

Does not this means to learn new syntax?

I think the $if/$then/$else is much more in line with current TiddlyWiki widgets and learning them is easier. They use the same concepts of current common TW widgets like $list.
One more benefit in @rmunn implementation is when newcommers from C/Python/Fortran/Java world will find the $if/$then/$else by @rmunn is more familiar. This is my personal opinion, as I am in favor of simpler or more uniform syntax everywhere.

pmario:
>[. . .] newcommers from C/Python/Fortran/Java world will find the $if/$then/$else by @rmunn is more familiar. 

I think a common misconception from users about wikitext is, that it is a programming language. IMO it's not. It's a text/web templating engine. 

It should be compared to templating engines like [Jinja ](https://en.wikipedia.org/wiki/Jinja_(template_engine)) or [Nunjucks](https://mozilla.github.io/nunjucks/) or any of many [others](https://colorlib.com/wp/top-templating-engines-for-javascript/)

AnthonyMuscio:
@pmario in @kookma diagrams, what if where it says "execute block" would it be more meaningful if we said "render block"?

Having come from the days of classic tiddlywiki I must say I get this whole "render" process (although it took some time), triddlywiki's special architecture, which I would argue actually endows macros in tiddlywikis "text/web templating engine" with more power that it would be in other platforms.
* personally I feel it is closer to a programming language than would appear on the surface.

Perhaps we could revisit describing the platforms architecture in less technical terms and this would go some way to helping people understand it better.

kookma:
> @pmario in @kookma diagrams, what if where it says "execute block" would it be more meaningful if we said "render block"?

Hi @AnthonyMuscio 
I think the Robin implementation addresses what I mean. So I am in favor of current implementation.



> I think a common misconception from users about wikitext is, that it is a programming language. IMO it's not. It's a text/web templating engine.

Well, I mean the Robin implementation is more familiar to those programmers! while I respect the difference of TiddlyWiki with a programming language like C.

AnthonyMuscio:
> Hi @AnthonyMuscio 
> I think the Robin implementation addresses what I mean. So I am in favor of current implementation.
> 

I only said
> I feel it is closer to a programming language than would appear on the surface.

And state again;

>Perhaps we could revisit describing the platforms architecture in less technical terms and this would go some way to helping people understand it better.

Jermolene:
Thank you everyone, particularly @rmunn.

I intend to merge #7710 and not this PR for the following reasons:

* The conditional construction in #7710 is less verbose in all cases
* This PR is incompatible with the existing `<$if>` widget

I appreciate the concerns about introducing a new syntax with `<%if%>`. It's a big step, but I think it will give us a lot of flexibility for adding further shortcut syntaxes in the future. For example, I'd like to add a select/case construction.

rmunn:
Closing PR since #7710 has been decided on.

PR #7694, opened 2023-08-23
Add `join` attribute to list widget

### LINKED ISSUE #7693: [BUG] List widget counter-last not updated correctly when items added to list
### Describe the bug

When you use the counter attribute of a list widget, the `<<counter-last>>` value doesn't update correctly as items are added to the list. When a new item is added to the list, its `<<counter-last>>` value is correctly set to "yes", but the item that *used* to be last doesn't get its `<<counter-last>>` value set to "no".

### Expected behavior

The previously-last item should have `<<counter-last>>` set to "no".

### To Reproduce

Create a new tiddler and paste in the following content:

```
Items: <$edit field="items" /> <br/>
<$list filter="[enlist{!!items}]" counter="c" variable="item"><<c>>:<<item>><$text text={{{ [<c-last>match[yes]then[]else[, ]] }}} /></$list>
```

Open the preview pane and slowly type "a b c d e" into the text edit box in the preview (or in the field edit pane, that works too), pausing to watch the list update as you add each new item to the list. Note how there are no commas, because each list item still thinks it's the last item in the list.

Highlight the letter "c" in the text-edit box and change it to "x". Note how items x and d now realize they're not the last item and therefore have commas after them, but "a" and "b" still have no commas because they still think they're the last item.

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: 5.2.1
- Saving mechanism: any
- Plugins installed: none

- OS: any
- Browser: any
- Version: any

- Device: any
- OS: any
- Browser: any
- Version: any

### Additional context

I first encountered this while trying to add a `join` attribute to the list widget, as suggested in https://github.com/Jermolene/TiddlyWiki5/discussions/7674. I re-used the existing logic in `handleListChanges` to process the `join` attribute, adding the join text if the `isLast` property was false on the list item's parse tree node, and omitting it if `isLast` was true. I thought for a while that my logic was wrong until I ran my reproduction on https://tiddlywiki.com/ and found that the bug was in the existing code.

The fix is going to be in the `handleListChanges` code. It currently deletes and re-creates the last item in the list if items were removed from the list (because that particular item used to be not-last and has now become last). But if items are *added* to the list then `handleListChanges` also needs to delete and re-create the item that used to be last and now is not-last, so that its `isLast` property will now correctly be false.

rmunn:
As suggested in https://github.com/Jermolene/TiddlyWiki5/discussions/7674. The `join` attribute can take arbitrary wikitext, but incurs a relatively-costly parsing step to do that. For the simple cases where all you want to do is insert a comma and/or a space between list items, the `joinText` attribute is there to enable that scenario to skip the parsing step.

Resolves #7674.
Fixes #7693.


pmario:
I thin fixes and new features should not be mixed. Especially if new features are not sure to be merged.

pmario:
As I wrote at the other issue. **I'm not a fan of that feature. It can properly be done within the list-widget code**. So it should not bloat the list widget. 

IMO the next request will be that we need to make the list-widget faster because poor filters, the counter variable and other highly use-case related functions did make it slow.

IMO TW should stay with it's principles, that widgets should provide the minimally possible functions. So small building blocks can be used to create flexible and dynamic code.

rmunn:
I didn't know about the bug until halfway through my implementation of the feature. I wasn't going to let the PR contain a known bug when I could solve it easily. If the feature is rejected, I'll submit a new PR with just the bugfix.

rmunn:
> IMO TW should stay with it's principles, that widgets should provide the minimally possible functions. So small building blocks can be used to create flexible and dynamic code.

That ship sailed long ago; have you seen the checkbox widget? Or the button widget? Attributes galore, each of them for a different use case. Any specific use case will only use three or four attributes, but there are about twenty in total on the button widget and a dozen on the checkbox widget, and none of those attributes goes unused in any particular use case.

Yes, `join` is use-case dependent. But it's such a common use case that it deserves a nice simple syntax. I remember when I was first starting to use TW and was boggled by the lack of any way to say "Hey, put a comma and a space between these list items, thanks." Now it *can* be done, but not in a way that beginners will be able to understand. But with `join=", "`, they can write "<$list filter=<<fruits>> join=", "/>" and get "apples, bananas, oranges, grapes" the way they'd expect a list to be laid out in English.

rmunn:
I should also add, on the subject of performance, that the way I wrote this PR, the `join` attribute should perform a lot faster (at least on large lists, it won't be noticeable on lists of 3-4 items) than using the `counter` attribute to do the same thing. I'll have to figure out a way to test performance to prove this, but I'm almost certain that `join=", "` will be far faster than doing `counter="var"` followed by `{{{ [<var-last>]match[no]then[, ]] }}}`.

The reason why has to do with the list widget implementation. If there's a `counter` attribute, and an item in the middle of the list is removed (or added), the list widget has to re-render all the items after the one that changed, because their counters changed. They used to be "4: d, 5: e, 6: f" and they have become "5: d, 6: e, 7: f". It's not possible for TW to skip that re-rendering, because it's impossible to tell if the counter variable is used in the list template. (It might not show up in the immediate wikitext, but there are ways to get a variable whose name is determined dynamically using a filter, so it's impossible to rule out all edge cases.)

With the `join` attribute (as implemented in this PR), though, the only thing the list item widgets care about is whether or not they are the last item in the list, in order to decide whether to render the joiner text. So if an item is added or removed from the middle of a list and the other items are unchanged, the TW code that handles list changes (which is pretty efficient at dealing with those scenarios) only has to care about rerendering two items: the item that *used* to be the last in the list, because it might no longer be last; and the item that is now the last item in the list, because it might have not been the last item before. So instead of rerendering, on average, half the elements in the list (which takes time proportional to the length of the list multiplied by the time it takes to render each element), TW will only need to spend the time rerendering a constant number of items. If you're familiar with big-O notation from computer science, the `[<counter-last>match[no]then[, ]]` approach takes O(N) time to refresh the list widget when a single item is added or removed in the middle, while the `join=", "` approach takes O(1) time in that same case.

So performance is actually a good reason to do this, rather than a reason to hold off.

pmario:
>That ship sailed long ago; have you seen the checkbox widget?

Yes -- it's a candidate for a complete redesign. It's code is hard to read, hard to maintain and as it seems also a bit brittle too. 

We have to live with it now. But the checkbox widget is rarely used. It will not kill the performance of the core UI.

The list-widget is **the** hart of TW it will kill the TW UI if it collects more and more functions.

>Or the button widget?

The button widget was part of TW v5.1.0 and it was growing to a black box with way too many functions. 
That was the **main reason** why the **action-widgets** where designed. 

The only parameters a button-widget needs to have for an average user is: `<$button actions=<<myActions>> class="" aria-label="">Click Me</$button>` 
The params **to, message, param, set, setTittle, setField, setIndex, setTo** should all be deprecated.

pmario:
Also see: https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussioncomment-6801358

rmunn:
Got some performance measurements. Unfortunately, I haven't been able to get performance measurements for updating an item in the middle of a list yet: when I edited an item near the beginning of a 5,000-item list I got `mainRefresh: 0 ms` even though it visibly took several seconds (3-4 seconds by my rough count) to actually show the update in the list. (That's with the existing, counter-based solution. The `join=", "` solution was indeed faster, about 1-2 seconds to update the same item in the 5,000-item list).

I did, however, manage to get performance measurements by using a reveal widget to show or hide the list. That doesn't measure update speeds, but here's what I did manage to get. Time as measured by TW's mainRefresh:

Using join:

434 ms
364 ms
379 ms
372 ms
412 ms

Average 392.2 ms over five runs

Using counter:

456 ms
535 ms
417 ms
442 ms
516 ms

Average 473.2 ms over five runs

So with this very rough estimate, the join attribute code in this PR is running about 17% faster than the code using `counter-last`.

<details>
<summary>Tiddler I used for testing, in case you want to reproduce it:</summary>

```
<$reveal type="nomatch" state="$:/temp/joinexample" text="show">

<$button set="$:/temp/joinexample" setTo="show">Show join example</$button>

</$reveal>
<$reveal type="match" state="$:/temp/joinexample" text="show">

<$button set="$:/temp/joinexample" setTo="hide">Hide join example</$button>

Items: <$edit field="items" /> <br/>
<$list filter="[enlist{!!items}]" join=", " variable="item"><<item>></$list>

</$reveal>

<$reveal type="nomatch" state="$:/temp/counterexample" text="show">

<$button set="$:/temp/counterexample" setTo="show">Show counter example</$button>

</$reveal>
<$reveal type="match" state="$:/temp/counterexample" text="show">

<$button set="$:/temp/counterexample" setTo="hide">Hide counter example</$button>

Items: <$edit field="items" /> <br/>
<$list filter="[enlist{!!items}]" counter="c" variable="item"><<item>><$text text={{{ [<c-last>match[yes]then[]else[, ]] }}} /></$list>
</$reveal>

Create <$edit field="itemcount" /> items: <$button action>Create<$action-setfield items={{{ [range{!!itemcount}join[ ]] }}}></$action-setfield></$button>
```

</details>

rmunn:
Okay, I've gotten some results from measuring the list-update code, and they're dramatic. I didn't measure the time it takes to redraw the list, only the time spent in the `ListWidget.prototype.handleListChanges` function, by changing the `result = this.handleListChanges(changedTiddlers);` line in `ListWidget.prototype.refresh` to:

```
// Handle any changes to the list
var start, stop;
start = performance.now();
result = this.handleListChanges(changedTiddlers);
stop = performance.now();
if(this.list.length > 499) {
	console.log('handleListChanges took', stop - start, 'ms');
}
```

The check for lists of 500 items or more is to stop my console from filling up with reports of lists taking 0, 1, or 2 ms to update.

I created a list of 5000 (five thousand) items via my test tiddler from https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1690160850 and then updated a single item near the start of the list (changing "10" to "x" in a list containing the numbers 1 through 5000 in order). Here were the results:

Using `join` attribute:

```
handleListChanges took 12 ms
handleListChanges took 14 ms
handleListChanges took 17 ms
handleListChanges took 16 ms
```

Using `<<counter-last>>`:

```
handleListChanges took 1119 ms
handleListChanges took 1103 ms
handleListChanges took 1154 ms
handleListChanges took 1094 ms
```

There's simply no comparison. The `<<counter-last>>` code path had to remove and replace 4,990 items in the list, even though nearly all of them had not changed, because there was no way for that code path to *safely* know that none of those items had changed. (They might have been using the `<<counter>>` value, and there's no way to prove that they aren't). The `join=", "` code path, on the other hand, only had to replace *two* items in that list of 5,000: the one that changed, and the last one in the list. (And even there, I think that code path hit the optimization that let it skip the last item in the list, because the list length hadn't changed).

Conclusion: if you care about the performance of the `<$list>` widget, you *definitely* want this PR, or one like it, to be merged.

pmario:
Main refresh also depends on the rest of the UI. 
So we have to measure 

- opening the reveal widget and 
- **subtract** closing the widget, to see the real values. 

My testcase was: 

- The "test tiddler" is **the only tiddler open**
- Right sidebar is set to "**Open**" tab 
- Using this app: https://tiddlywiki5-git-fork-rmunn-feature-list-join-jermolene.vercel.app/
- I did save the wiki and than a reload before starting measurements.

## join 5000 elements

- I needed 6 samples to basically get the same results on open and close
- Open-Close: 223-10, 217-10, 196-9, 187-6, 187-7, 187-6 -> ~161ms average

## counter

- 6 samples
- Open-Close: 282-8, 276-7, 258-7, 233-8, 223-6, 222-5 -> ~242ms average

## function

- 6 samples
- open-close: 17-8, 9-6, 9-5, 7-5, 8-4, 8-6 -> ~4ms average

So I think using list for exactly this job is the wrong way to go. 

<details><summary>I did use the following code</summary>

```
\function list-join(filter, sep:", ") [subfilter<filter>join<sep>]

<$let temp="$:/temp/join/example">
<$reveal type="nomatch" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="show">Show join example</$button>
</$reveal>
<$reveal type="match" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="hide">Hide join example</$button>
	Items: <$edit field="items" /> <br/>
	<$list filter="[enlist{!!items}]" join=", " variable="item"><<item>></$list>
</$reveal>
</$let>

<$let temp="$:/temp/counter/example">
<$reveal type="nomatch" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="show">Show counter example</$button>
</$reveal>
<$reveal type="match" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="hide">Hide counter example</$button>
	Items: <$edit field="items" /> <br/>
	<$list filter="[enlist{!!items}]" counter="c" variable="item"><<item>><$text text={{{ [<c-last>match[yes]then[]else[, ]] }}} /></$list>
</$reveal>
</$let>

<$let temp="$:/temp/function/example">
<$reveal type="nomatch" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="show">Show function example</$button>
</$reveal>
<$reveal type="match" state=<<temp>> text="show">
	<$button set=<<temp>> setTo="hide">Hide function example</$button>
	Items: <$edit field="items" /> <br/>
	<<list-join "[enlist{!!items}]">>
</$reveal>
</$let>

Create <$edit field="itemcount" />
items: <$button action>Create<$action-setfield items={{{ [range{!!itemcount}join[ ]] }}}></$action-setfield></$button>
```
</details>

pmario:
To create links using the `items` field of the current tiddler, the algorithm is: 

```
\function list-join(filter, sep:", ") [subfilter<filter>join<sep>]
\function getTo(sep:",") [<item>removesuffix<sep>else<item>]

<$let joined-list=<<list-join "[enlist{!!items}]">> >
	<$list filter="[<joined-list>enlist-input[]]" variable="item" >
		<$link to=<<getTo>> ><<item>></$link>
	</$list>
</$let>
```

There may be more optimized versions, but to prove the point IMO it should be good. 

### Proposal

I think we should try to figure out how to work with `procedures, functions, parametrized transclusions, slots, fill, genesis` widgets  first before we add new usecase related functionality to the list widget.

rmunn:
@pmario - in https://github.com/Jermolene/TiddlyWiki5/pull/7691#issuecomment-1692757844, you wrote:
> What users want:
> 
> ```
> <$if filter="something"><$then>
> 
> Some plain text. 
> <$else>
> 
> Other  text
> </$if>
> ```
> 
> Which does not need any explanation and can be written by everyone.

On the `<$if>` widget, I entirely agree with the position you took there. I find it interesting that here you are taking the opposite position, arguing for a complex solution instead of the simple one. The one I'm proposing that is simple and can be written by anyone:

```
<$list filter="some filter" join=", " variable="item">
	<<item>>
<$list>
```

The syntax you're proposing:

```
\function list-join(filter, sep:", ") [subfilter<filter>join<sep>]

<$let joinedList=<<list-join "[enlist{!!items}]">> >
	<$list filter="[<joinedList>enlist-input[]]" variable="item" >
		<<<item>>
	</$list>
</$let>
```

Now, yours isn't awful, by any means. (I omitted the part with `<$link />` to make it a fair comparison, since my example was aimed at filling OP's need which was JSON output, where he wants plain text). But it doesn't fit the "can be written by anyone" metric, as it still requires being quite comfortable with filters: `subfilter` and `enlist` are not easy to grasp on someone's first encounter with TiddlyWiki. Even for someone with a background in programming, like I have, it's not immediately obvious what `enlist` does, and or what `enlist-input` does or why it's needed, until you have some experience with TiddlyWiki's filters and how they work. On the other hand, anyone can add a `join=", "` attribute and get the right thing right away, without needing to write special filter code to strip off the separator that was added. (And which might even do the wrong thing if the final item in the list just happened to end with that separator, which is a rare case but can never be ruled out entirely).

pmario:
>The syntax you're proposing:

No. That's just a prove of concept, that it can be done with existing core functions in a performant manner.

-------

My proposal would be a an abstraction like `<<makeJson filter>>`, shown below, where the only input would be a `filter` parameter and some other parameters needed. 

```
\procedure makeJson(filter, sep:", ")
	\function list-join() [subfilter<filter>join<sep>]
	<<list-join>>
\end

<<makeJson "[tag[HelloThere]]">>
```

------------

>[. . .] my example was aimed at filling OP's need which was JSON output [. . .]

The OP in issue #7674 **ListWidget join parameter** is not very precise. It only points to the list-widget and wants a `join` parameter. 

I'm not sure if that's needed. It think it's a case of "You've done it wrong". TW renders JSON output all the time since v5.1.0. And now we need `join[, ]` in the list widget to be able to do it. I cannot believe that. 

That's why I wrote we should explore other possibilities with the new features included in v5.3.x. 

-------

>[. . .] it's not immediately obvious what enlist does, and or what enlist-input does or why it's needed, until you have some experience with TiddlyWiki's filters and how they work. On the other hand, anyone can add a join=", " attribute and get the right thing right away, without needing to write special filter code to strip off the separator that was added. [. . .] 

I do not have a problem with a list-like widget that has a `join` parameter. 

I do have a problem that you want to add it to the list-widget, which is the core functionality in TW. I personally think adding the `counter` parameter was an error. It's misused by users. I've seen posts at Talk already, that contain problematic code, because it's easy to misuse.

pmario:
For Jeremy, I did modify the post: https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1693095709

>That's why I wrote we should explore other possibilities with the new features included in v5.3.x. 
>
>With more info about the real usecase it may be possible to give a better advice. The wikify-widget may be an option here and if not we should have a closer look if we need to expose more server `--render` like functions to the user.

Jermolene:
Thanks @rmunn the performance improvements are impressive and I think justify the additional complexity for the list widget.

I imagine that we would see a similar gain if we were to update the saving template to use it for generating the JSON store area. See the template [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/templates/html-json-tiddler.tid) and the surround list widget [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/templates/store.area.template.html.tid#L11).

editions/tw5.com/tiddlers/widgets/ListWidget.tid
```diff
@@ -117,10 +119,29 @@ Displays as:
...
 
 Setting `counter="transclusion"` is a handy way to make child elements for each list element be identified as unique. A common use case are multiple [[tag macros|tag Macro]] for the same tag generated by a list widget. Refer to [[tag macro examples|tag Macro (Examples)]] for more details.
 
+!! `join` and `joinText` attributes
+
+The optional `join` or `joinText` attributes allow you to insert something between each list item without needing to use the `counter` attribute, which can become quite slow if the list is updated frequently. The difference between them is that the `join` attribute will have its contents wikified, while `joinText` will not. If both `join` and `joinText` are given, `joinText` has precedence and `join` will be ignored.
+
+<<.tip """Wikifying the attribute value is a relatively slow process, so if all you need is a bit of plain text (say, a comma and a space) between list items, `joinText` will perform a lot better than `join`. But if you need some HTML element like `<br>` inserted between list items, then use the `join` attribute.""">>
```
pmario, line 130: Did you think about to do a different nameing. eg: `join` is the text/plain parameter and `joinWikified` the text/vnd.tiddlywiki one?

rmunn:
I pulled out the bugfix (with no changes to the list widget) into https://github.com/Jermolene/TiddlyWiki5/pull/7712.

editions/tw5.com/tiddlers/widgets/ListWidget.tid
rmunn: I did. In https://github.com/Jermolene/TiddlyWiki5/pull/7694/commits/71a4a9fc0acd791b0c01051ab21506c129e83c46 I had `join` be the plaintext paramenter, and `joinTemplate` (not a good name, `joinWikified` is better) be the wikified one. But then I thought that people will expect `join="<br/>"` to work and will be confused if it doesn't, so in the next commit (https://github.com/Jermolene/TiddlyWiki5/pull/7694/commits/d0c23ed5abd79c70151f7598b2dab3cfe501951b) I switched it back to `join` being the wikified one and `joinText` being the plain-text one.

I'm open to suggestions, though. I actually prefer `join` being plain text, but would need to hear opinions from users. Maybe I should start a thread on https://talk.tiddlywiki.org/.

pmario: >Maybe I should start a thread on https://talk.tiddlywiki.org/.

I do not think so. It will be 50/50. IMO @Jermolene should decide it and we are done.

rmunn:
@Jermolene - About this:

> I imagine that we would see a similar gain if we were to update the saving template to use it for generating the JSON store area. See the template [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/templates/html-json-tiddler.tid) and the surround list widget [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/templates/store.area.template.html.tid#L11).

I'm looking at those templates now, and while I think they won't save that much time if they aren't constantly being refreshed, they certainly will save some complexity by having a join attribute instead of the `<$list>` logic in the template itself. That leads me to a question, though. I noticed that core and the tiddlyweb plugins both have html-json-tiddler and html-json-skinny-tiddler templates, but they're inconsistent. In both cases, the tiddlyweb plugin adds `$revision=<<changecount>> $bag="default"`, but in the html-json-skinny-tiddler template, the tiddlyweb plugin also has `$_is_skinny=""` which the same template in core is lacking. Based on the name, I assume that the html-json-skinny-tiddler.tid template in core was also supposed to have the `$_is_skinny=""` attribute, and I'll add it while I'm fixing the other stuff in there.

Please let me know if I've misunderstood the intent behind that file, and if the core html-json-skinny-tiddler.tid template is **not** supposed to have `$_is_skinny=""` in it.

Jermolene:
Hi @rmunn one of the foundational improvements in #7710 is an extension to the list widget to allow both the item template and the empty message template to be specified via explicit subwidgets:

```
<$list filter=<<filter>>>
	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
</$list>
```

It would be natural to extend that syntax to cater for a join template too:

```
<$list filter=<<filter>>>
	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
	<$list-join>
		,
	</$list-join>
</$list>
```

rmunn:
About to rebase this on top of current master, then I'll be able to add code to support a `$list-join` pseudo-widget. However, the bugfix for #7804 is going to end up touching some of the same code, so I might wait until #7804 is fixed before I work on the `$list-join` code, otherwise I'll end up with merge conflicts.

rmunn:
I just did an experimental implementation of the `<$list-join>` pseudo-widget, basing it on top of the code I wrote for #7810. And it was so easy and simple (just five lines changed, though of course documentation and unit tests will add more) that I'm going to wait until #7810 is merged into master, then rebase this PR branch again before adding the `<$list-join>` feature. To avoid merge conflicts and complications, I won't add the code I wrote to this branch... but here's the diff for future reference, so that I can find it easily once it's time to work on it. (And so that you can see how simple it was):

```diff
diff --git a/core/modules/widgets/list.js b/core/modules/widgets/list.js
index cb97ed46e..8293269f0 100755
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -102,6 +102,9 @@ ListWidget.prototype.findExplicitTemplates = function() {
                        } else if(node.type === "list-empty") {
                                found = true;
                                self.explicitEmptyTemplate = node.children;
+                       } else if(node.type === "list-join") {
+                               found = true;
+                               self.explicitJoinTemplate = node.children;
                        } else if(node.type === "element" && node.tag === "p") {
                                var foundInChildren = searchChildren(node.children);
                                found = found || foundInChildren;
@@ -112,7 +115,7 @@ ListWidget.prototype.findExplicitTemplates = function() {
        var removeTemplates = function(childNodes) {
                var result = [];
                $tw.utils.each(childNodes,function(node) {
-                       if(node.type === "list-template" || node.type === "list-empty") {
+                       if(node.type === "list-template" || node.type === "list-empty" || node.type === "list-join") {
                                // Do nothing
                        } else if(node.type === "element" && node.tag === "p") {
                                var childrenWithoutTemplates = removeTemplates(node.children);
@@ -185,7 +188,7 @@ ListWidget.prototype.makeJoinTemplate = function() {
                joinText = this.getAttribute("join",""),
                joinTemplate = this.getAttribute("joinWikified","");
        if(joinText === "" && joinTemplate === "") {
-               return undefined;
+               return this.explicitJoinTemplate;
        }
        if(joinText) {
                // No need to call a heavyweight parser for the simple cases
```

rmunn:
I'm about to rebase this PR on top of the code for #7827. So there will be merge conflicts until #7827 is merged. But then as soon as #7827 is merged, this one will be ready to go. Hopefully it will be in time to make it into the 5.3.2 release.

rmunn:
Now that #7827 is merged, I've rebased this on top of `master` and solved the merge conflicts. @Jermolene - This PR is ready to be reviewed and merged.

rmunn:
I've found a bug: when the last two items of a list are *swapped*, the previously-last item will end up without its join text attached. I.e. if "1, 2, 3, 4" has 3 and 4 swapped, it becomes "1, 2, 43". Pushing two more commits: a failing test for that bug, and a commit that fixes the bug and makes the test pass.

rmunn:
Found another bug: if the last item in the list is moved somewhere else, like "1 2 3 4" becoming "1 4 2 3", it's not being updated. Just pushed a failing test and then a fix.

BTW, I'm very grateful for the Vercel deployments, as they're allowing me to test my code in real-world scenarios. E.g., the last two bugs were found by using the following tiddler:

```plaintext
<$list filter="[tag[Tag Operators]]" join=", " />
<br/>

<<list-tagged-draggable tag:"Tag Operators">>
```

And then rearranging the tag list while watching to see if the `<$list>` widget put join text in all the right places.

rmunn:
And another bug. I do think this is the last one, though, as I seem to have exhausted all other possible list rearrangement scenarios. So once the bugfix lands (coming up shortly), I think this will be ready.

rmunn:
Just found one more bug, but it might be a little tricky to create a unit test for it, so here's a manual reproduction until I can create a proper test.

1. Create a new tiddler called ListTest with the following content:
    ```
    <$list filter="[tag[Tag Operators]]" join=", " />
    <br>
    <<list-tagged-draggable tag:"Tag Operators">>
    ```
2. Edit the HelloThere tiddler, and add the "Tag Operators" tag to it. Do not save yet.
3. Before saving the edit, look at ListTest and observe that "Draft of 'HelloThere'" is in the list.
4. Save the HelloThere edit.
5. Rearrange the tags by dragging and dropping so that HelloThere is somewhere in the middle.
6. Edit the HelloThere tiddler and look at ListTest, which has both HelloThere and "Draft of 'HelloThere'" in it now.
7. Remove the "Tag Operators" tag from HelloThere. Do not save yet.
8. Before saving, look at ListTest and observe that HelloThere is still in the list, but "Draft of 'HelloThere'" is gone.
9. Save the HelloThere tag edits. Look at ListTest and observe that HelloThere is also gone from the list.
10. Edit the HelloThere tiddler *again*, and add the "Tag Operators" tag. Do not save yet.
11. Before saving, look at ListTest and observe that HelloThere is not yet in the list, but "Draft of 'HelloThere'" is there, at the end of the list.
12. Save the HelloThere tiddler edits.
13. Look at ListTest: HelloThere is in the middle, "Draft of 'HelloThere'" is gone from the end, *but* there is a stray comma that shouldn't be there at the end of the list: one joiner has gotten "stuck" to the end.

This does not affect the list widget code that uses `counter` attribute, only the code that uses `join`. I'll create a unit test as soon as I can figure out how to boil that behavior down to a short test, and then I'll push that failing test.

rmunn:
Simpler reproduction for what is probably the same bug. Make a ListTest tiddler with the following:

```
Join: <$list filter="[tag[foo]]" join=", " />
<br/>
Counter: <$list filter="[tag[foo]]" counter="c">
<$link /><% if [<c-last>match[no]] %>, <% endif %>
</$list>

<<list-tagged-draggable tag:"foo">>
```

Now start creating new tiddlers called "a", "b", "c". Tag them with "foo". Watch the ListTest tiddler. The list using counter will be correct, but the list using `join` will fail to add commas between the list items as they are added.

rmunn:
@Jermolene - I'm not finding any more bugs: every list edit scenario I can come up with, the `join` attribute is now handling correctly. So I'm done with bugfixes (unless someone else turns one up) and I'm ready for you to review, and hopefully merge, this.

Jermolene (commented):
Hi @rmunn my one remaining concern is the use of the `join` and `joinWikified` attributes. It feels a lot of additional complexity in the docs to cater for optimising a bit of a corner case – now that we have the `<$list-join>` widget it is easy to avoid the inefficiency of the unnecessary parsing.

rmunn:
> Hi @rmunn my one remaining concern is the use of the `join` and `joinWikified` attributes. It feels a lot of additional complexity in the docs to cater for optimising a bit of a corner case – now that we have the `<$list-join>` widget it is easy to avoid the inefficiency of the unnecessary parsing.

I was thinking that myself: should I get rid of `joinWikified` since it's nearly redundant with `$list-join`? The answer, I think, is yes. It's not *quite* redundant: `joinWikified` might be set from a filter or something, which is hard to do with `$list-join`. But if you want the joiner set from a filter *and* it needs to be wikitext, then you could use a `$set` widget outside the list and then `<$list-join><<calculatedJoiner>></$list-join>` inside the widget.

So yes, I'll remove `joinWikified`. That will clean up the docs a lot, and I can just say "Note that the `join` attribute will be treated as plain text and will not be wikified. If you need HTML like `<br>`, then put it in `<$list-join>` instead." That's a lot simpler than trying to explain how `joinWikified` has to be re-parsed each time and is inefficient.

Jermolene:
> I was thinking that myself: should I get rid of `joinWikified` since it's nearly redundant with `$list-join`? The answer, I think, is yes. It's not _quite_ redundant: `joinWikified` might be set from a filter or something, which is hard to do with `$list-join`. But if you want the joiner set from a filter _and_ it needs to be wikitext, then you could use a `$set` widget outside the list and then `<$list-join><<calculatedJoiner>></$list-join>` inside the widget.
> 
> So yes, I'll remove `joinWikified`. That will clean up the docs a lot, and I can just say "Note that the `join` attribute will be treated as plain text and will not be wikified. If you need HTML like `<br>`, then put it in `<$list-join>` instead." That's a lot simpler than trying to explain how `joinWikified` has to be re-parsed each time and is inefficient.

The trouble with fixing it that way is that we end up with an inconsistency between the `emptyMessage` attribute and the `join` attribute. I think it makes more sense to wikify the attribute in the usual way.

rmunn:
> > I was thinking that myself: should I get rid of `joinWikified` since it's nearly redundant with `$list-join`? The answer, I think, is yes. It's not _quite_ redundant: `joinWikified` might be set from a filter or something, which is hard to do with `$list-join`. But if you want the joiner set from a filter _and_ it needs to be wikitext, then you could use a `$set` widget outside the list and then `<$list-join><<calculatedJoiner>></$list-join>` inside the widget.
> > So yes, I'll remove `joinWikified`. That will clean up the docs a lot, and I can just say "Note that the `join` attribute will be treated as plain text and will not be wikified. If you need HTML like `<br>`, then put it in `<$list-join>` instead." That's a lot simpler than trying to explain how `joinWikified` has to be re-parsed each time and is inefficient.
> 
> The trouble with fixing it that way is that we end up with an inconsistency between the `emptyMessage` attribute and the `join` attribute. I think it makes more sense to wikify the attribute in the usual way.

:frowning_face: :cry: :frowning_face: 

That would end up throwing away most of the performance gains that `join` offers in the common cases. Unless there's some way to cache the parse results, the `join` attribute value will need to be parsed every time the list is rendered. (You can't even skip parsing it for an empty list: I discovered during my testing that if you do, then the joiners aren't present when they should be if the list went from empty to having two or more items in a single refresh cycle).

In https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussioncomment-6778283, you mentioned that the "emptyMessage" attribute is awkward, because it's often difficult to place arbitrary wikitext constructions in an attribute value. So I decided to optimize the "join" attribute for the most common case, that of wanting to put plain text like ", " or ", and " in between list items, and let the slightly-less common cases like `<br>` use the more awkward construction. This all started with [a request for `<$list join=",">`](https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussion-5505809), and that's the use case I think we should optimize for.

https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussioncomment-6778283

Jermolene:
> That would end up throwing away most of the performance gains that `join` offers in the common cases. 

To what extent can the performance gain be regained by switching to using `<$list-join>` instead of the `join` attribute? 

> In [#7674 (reply in thread)](https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussioncomment-6778283), you mentioned that the "emptyMessage" attribute is awkward, because it's often difficult to place arbitrary wikitext constructions in an attribute value. So I decided to optimize the "join" attribute for the most common case, that of wanting to put plain text like ", " or ", and " in between list items, and let the slightly-less common cases like `<br>` use the more awkward construction. This all started with [a request for `<$list join=",">`](https://github.com/Jermolene/TiddlyWiki5/discussions/7674#discussion-5505809), and that's the use case I think we should optimize for.

If the optimisation were needed it would be more consistent to have a "joinText" attribute that is not wikified along with the wikified "join" attribute.

I'm guessing that the situations where this optimisation would have the most impact would be generating JSON, a task that may in any case be amenable to more significant, task-specific optimisations.

rmunn:
Just implemented `join` as parsing WikiText and discovered something I hadn't turned up before in my testing. When I did `<$list filter="[range[3]]" join=" and " />`, the space after the word "and" remains, but the space before "and" gets stripped: the result looks like "1and 2and 3". In order to get the leading space to appear, I have to write `join="&nbsp;and "`, which will *not* be obvious to a casual user. I believe the cause is the parser skipping over leading whitespace, for some reason I have yet to identify (setting `configTrimWhiteSpace` to false did not fix the incorrect behavior).

I'll push the commit with that change so that Vercel will deploy it and you can verify that for yourself. But then I'll push another commit reverting the change, because that behavior is not what we want. It looks to me like "text, not wikified" does need to be the default for the join attribute, because people are going to care about those leading spaces in many common scenarios.

rmunn:
Okay, https://tiddlywiki5-5vs7wdg0a-jermolene.vercel.app/ is the deployment with the `join` attribute parsing WikiText by default, and https://tiddlywiki5-l2g3we14r-jermolene.vercel.app/ is the deployment where the `join` attribute is reverted back to plaintext by default. Try the following test tiddler in both deployments:

```
<$list filter="[range[3]]" join=" and " />
```

In the WikiText-by-default deployment, you get "1and 2and 3". In the plaintext-by-default deployment, you get "1 and 2 and 3".

rmunn:
@Jermolene - I believe I've found a way to optimise the `join` attribute so it's only parsed once (and again if it ever changes, though I expect few people are going to do `join={{SomeTransclusion}}` so it'll rarely change). However, since there are parsing issues around leading whitespace in the attribute (the `join=" and "` attribute), I think we need to offer a plaintext option. As I see it, we have the following choices:

1. Make `join` be plain text, and bring back `joinWikified`.
2. Make `join` be plain text, don't have `joinWikified`, and just use the `$list-join` widget for wikitext or HTML. (My preference)
3. Make `join` be wikified, and add a `joinText` attribute for plain text. (What I think you'd prefer, and how I originally wrote this).
4. Don't have a `join` attribute at all, just `joinText` for plain text, and use `$list-join` for wikitext.

Or, I suppose...

5. Do not include this feature in 5.3.2 because we're not sure what the best interface should be.

I would prefer not to miss the 5.3.2 merge window, because I spent quite a lot of effort on this and it's been basically ready for review for a month, just waiting on #7827 to get merged first.

I can implement any of those options quickly and easily: at this point, I'm so familiar with the code that it will take me less than an hour to implement whichever one of those you choose. Just let me know.

I should say one more thing. You mentioned concern about inconsistency with `emptyMessage`, which is wikified. But I expect most people won't be bothered by that inconsistency — I certainly wouldn't be — because the two attributes have rather different purposes. `emptyMessage` is likely to be long, and could quite easily have some HTML or widgets in it. (Say, a loading spinner to indicate that the data that goes here, which is being fetched from an HTTP request, hasn't arrived yet). But joiners between lists are very likely to be short bits of text, like commas or comma-newline sequences (for JSON output). I expect users to want plain text 90% of the time. And I think the *common* use case should be *easy*, while the *less common* use case should be *possible*. Making the less common use case easy while the common use case is possible (but less convenient) out of concern for consistency is, IMHO, not the best choice. I think a small inconsistency that's well documented is worth it for the sake of making the common use case easy.

But, as always, in the end it's your project, and I'll implement whichever one of those you choose. Just let me know your final decision and I'll get this PR ready to finally merge.

Jermolene:
> Just implemented `join` as parsing WikiText and discovered something I hadn't turned up before in my testing. When I did `<$list filter="[range[3]]" join=" and " />`, the space after the word "and" remains, but the space before "and" gets stripped: the result looks like "1and 2and 3". In order to get the leading space to appear, I have to write `join="&nbsp;and "`, which will _not_ be obvious to a casual user. I believe the cause is the parser skipping over leading whitespace, for some reason I have yet to identify (setting `configTrimWhiteSpace` to false did not fix the incorrect behavior).

Ouch! Presumably we get this same behaviour if we use the `<$list-join>` widget? (In the text below the word "and" is preceded and followed by whitespace).

```
<$list filter="...">
<$list-template>
...
</$list-template>
<$list-join>
 and 
</$list-join>
</$list>
```

If whitespace is trimmed in that situation as well, then perhaps the conclusion is that the list widget with these join capabilities is not actually a good match for situations where we're doing whitespace-sensitive plain text processing. It's the same situation we face in regular templates where we endlessly have to wrap plain text in backticks to ensure that it doesn't get parsed.

>As I see it, we have the following choices:

Performance is a factor here, so it would be helpful to understand the performance differences between using a `<$list-join>` widget and a joinText/joinWikified attribute. Perhaps we also need to experiment with the join text wrapped in backticks in a `<$list-join>` widget.

I've just realised that there might also be a case for a way to specify a tiddler title as the join template. It would be consistent with the `template` attribute, and would allow for caching of the parse tree.

Jermolene:
> I would prefer not to miss the 5.3.2 merge window, because I spent quite a lot of effort on this and it's been basically ready for review for a month, just waiting on https://github.com/Jermolene/TiddlyWiki5/pull/7827 to get merged first.

It's turned out to be controversial, but I think mainly because there has turned out to be a lot of unexpected inwardness to it.

Jermolene:
> But this produces "1 and 2 and 3":

That sounds like a block vs inline parsing issue...

rmunn:
I'm calling it with `this.wiki.parseText("text/vnd.tiddlywiki",join,{parseAsInline:true})`, just like the `emptyMessage`. Does that sound right? I don't know the parser implementation inside and out yet...

Jermolene:
> I'm calling it with `this.wiki.parseText("text/vnd.tiddlywiki",join,{parseAsInline:true})`, just like the `emptyMessage`. Does that sound right? I don't know the parser implementation inside and out yet...

I think it's the parsing of the item template that is the issue here.

rmunn:
Found the issue via judicious application of `console.log` statements in the wikiparser.js code. It's the wikiparser that's causing the join text to have the initial whitespace skipped. In the [WikiParser code](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/parsers/wikiparser/wikiparser.js#L87) there's a call to `this.parsePragmas()`, and the first thing that `parsePragmas` does is to call `this.skipWhitespace()`. The parse tree for the joiner " , " (a space, a comma, and another space) ends up being `[ { type: 'text', text: ', ', start: 1, end: 3 } ]` because by the time parseInlineRunUnterminated is called for the first time, this.pos is 1 because `parsePragmas` skipped over whitespace and didn't restore the previous position when it found no pragmas.

I'll push a new commit with this bug fixed.

rmunn:
@Jermolene - I've fixed the parsing problem with a two-line change in wikiparser.js; the tests all continue to pass with that change, but please take a look to see if you think it will cause issues. Basically, it ensures that parsing pragmas will only skip over whitespace found at the start of a pragma; any whitespace *after* pragmas (or if there are no pragmas found) will not be skipped.

Finally, in a separate commit (95ed91d) I've added a `joinTemplate` attribute similar to the `template` attribute for specifying the list template. I'm not a big fan of this since joins are usually small, so if you decide to take it out I'll be happy to revert that commit.

But as of now, the bugs with wikitext parsing of the `join` attribute are fixed, the `join` attribute only re-parses when it has changed (so it's nearly as efficient as `joinText`), and I'm finally happy with the result (except for `joinTemplate`, which I don't see a use for since `join={{JoinTemplateTiddler}}` would work just as well). But if you really do want to keep `joinTemplate`, it's there. Otherwise, feel free to revert commit 95ed91d and then merge.

Jermolene:
> @Jermolene - I've fixed the parsing problem with a two-line change in wikiparser.js; the tests all continue to pass with that change, but please take a look to see if you think it will cause issues. Basically, it ensures that parsing pragmas will only skip over whitespace found at the start of a pragma; any whitespace after pragmas (or if there are no pragmas found) will not be skipped.

Great work, this an important improvement. I think the fix should go into its own PR so that the subsequent commit history is crystal clear.

rmunn:
Okay, https://github.com/Jermolene/TiddlyWiki5/pull/7835 is the wikiparser.js fix for skipping whitespace. And since this PR has gotten a whole lot of commits, some of which are reversions of reversions, I'm going to squash this PR down to a single commit (with an appropriate commit message) so that it doesn't end up producing confusing commit history when merged.

rmunn:
Commit 0263fc3 contains this PR squashed down into a single commit, for easier comparison and merging. If you need the pre-squashed version, that's still available in commit 95ed91d and its ancestor commits.

Jermolene:
> Commit [0263fc3](https://github.com/Jermolene/TiddlyWiki5/commit/0263fc3b0efd1fe5c477d54cc868c3e6627bb5a3) contains this PR squashed down into a single commit, for easier comparison and merging. If you need the pre-squashed version, that's still available in commit [95ed91d](https://github.com/Jermolene/TiddlyWiki5/commit/95ed91d8476b9b4b0f63ef6bc5db5052ab67c069) and its ancestor commits.

Much appreciated. In fact we use GitHub's "squash and merge" option when we merge PRs, so it isn't necessary.

rmunn:
Is there anything else you need me to do to this PR? It's been ready to merge for a week, and the PR freeze is coming up, so if you want me to change anything else about it please let me know soon.

Jermolene:
> Is there anything else you need me to do to this PR? It's been ready to merge for a week, and the PR freeze is coming up, so if you want me to change anything else about it please let me know soon.

Thanks @rmunn all good, I just wanted to make some time to give it a proper review.

AnthonyMuscio:
I have come here from the Pre-release and using a filter in the following example;

```
<$list filter="[tag[TableOfContents]]" variable="item" join=", ">
<$text text=<<item>>/>
</$list>
```

This leaves a space before the comma, unless placed on a single line.
```
<$list filter="[tag[TableOfContents]]" variable="item" join=", "><$text text=<<item>>/></$list>
```

Is this the intended behaviour?, and perhaps the example needs to be change otherwise it generates an undesirable result.

rmunn:
@AnthonyMuscio -

What happens if you do `\whitespace off` at the top of that example?

AnthonyMuscio:
@rmunn I am sure that "whitespace trim" would work in standalone tiddlers but it does not help with the list widget, when it is just part of a longer tiddler.
* Without wanting to discourage the current initiative, I have written a few algorithms for different join requirements previously and think this can be improved further.
* For example commas between, period at end, using html tags `<br>` `<p></p>`, using `<ol>/<ul>` and `<li></li>`

rmunn:
The precise cause of the extra space between the list item and the joiner text is that there's a newline after the `<$text>` widget. The newline gets incorporated into the template since `\whitespace` is set to "notrim", and HTML interprets any series of spaces and newlines as a single space in the output (for good reasons that would take too long to get into here). If you write it like this:

```plaintext
<$list filter="[tag[TableOfContents]]" variable="item" join=", ">
<$text text=<<item>>/></$list>
```

then there's no space between the text and the joiner.

That's intended behavior AFAIK, but @Jermolene can tell us more.

rmunn:
@AnthonyMuscio - Forgot to add: please let me know precisely how you'd suggest improving it further. The more details, the better. I'll just mention that `<p></p>` and `<li></li>` are possible now without the join, and indeed cannot be done in the join attribute as it currently exists: you cannot start an HTML element in one joiner and finish it in the other. I.e. you couldn't do `join="</li><li>"` because each joiner must be a valid self-contained WikiText snippet, and having a trailing `<li>` with no closing `</li>` in the same snippet won't work.

AnthonyMuscio:
Sorry about the delay @rmunn 
```
\function comma.list() [join[, ]addsuffix[.]]

{{{ [tag[TableOfContents]comma.list[]] }}}
```
Result:
Home, HelloThere, Learning, Features, Languages, Editions, Plugins, Platforms, Reference, Paramatised template blocks.

Parametised
```
\function comma.list(join:", " end:".") [join<join>addsuffix<end>]

{{{ [tag[TableOfContents]comma.list[ - ],[ - eol -]] }}}
```

Or full list items with links
```
\function join.end(join:", " end:".") [<item>match<items>then<end>else<join>]

<$let items={{{ [tag[TableOfContents]count[]] }}}>

<$list filter="[tag[TableOfContents]]" counter=item>
<$link/><<join.end>>
</$list>

</$let>

Or <$link/><<join.end " - " " - end of list -">>
```

With all due respect these methods are superior to the one in the pre-release and I believe the pre-release is verbose and will distract us from solutions like mine.
* If "items" was an automatic standard variable it would be even easier to use (I called for this recently) or tv-items-count.
* Rather than a custom operator we could provided a new core one even simpler. 
* The proposed sub widget introduce something we do not use and in some-ways is a code pattern that will cause confusion with the current $fill and $slot widgets.
```
<$let items={{{ [tag[TableOfContents]count[]] }}}>
<$list filter="[tag[TableOfContents]]" counter=item>
<li><$link/><<join.end>></li>
</$list>
```
* Optionally wrap the whole list widget with `<ol> or <ul>`

Without presenting the details now I think I can use this and other new code patterns to take this a lot further. 
* For example inside a list template item.


Also consider
```
\function con.joiner(join:", " end:".") [<item>!match<items>then<join>] :else[<end>addsuffix[ (]addsuffix<items>addsuffix[)]]

<$let items={{{ [tag[TableOfContents]count[]] }}}>
<$list filter="[tag[TableOfContents]]" counter=item>
<$link/><<con.joiner>>
</$list>
```
This gives a total items at the end.

A lot more can be done with just these patterns like a header and detailed footer stored in a procedure and given as the value of `end`".

pmario:
@AnthonyMuscio Please read the discussions in this PR and also the linked threads in the OP.

IMO you start over a discussion, which we already head.

The main argument https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1690196498 is speed. IMO That's the one, which is the biggest advantage, which justifies this new option

AnthonyMuscio:
I will try and find other references but please Concider my above comments I put a lot of effort into it.

flibbles:
I won't pile on my own remarks about how we may be overbloating the List Widget. I've come to terms with TiddlyWiki's feature creep. I'll live with with.

But at the very least, can people not see that there is no reason in the world to add both a join-attribute AND a join-widget? Why add two features for the same thing that will both need to be supported in perpetuity afterward? And looking at this thread, it looks the join-widget came about literally from someone saying, "oh hey! We should add this too!" instead of anyone having asked for the feature.

Because who *would* want it? Who in the world actually needs joins complicated enough to require multiple lines, but is ALSO not familiar enough with TiddlyWiki to use \function and [join[]] and counter-first to figure out their own way?

The same could be said for the join-attribute, but it's definitely true about the join-widget.

Jermolene:
> But at the very least, can people not see that there is no reason in the world to add both a join-attribute AND a join-widget? Why add two features for the same thing that will both need to be supported in perpetuity afterward? And looking at this thread, it looks the join-widget came about literally from someone saying, "oh hey! We should add this too!" instead of anyone having asked for the feature.

The disadvantage of the "join" attribute is that its content has to be explicitly parsed by the list widget before it can be used. Using the `<$list-join>` widget avoids the parsing step.

flibbles:
@Jermolene. I appreciate that. Wikitext in attributes is clumsy. So that would argue that one should *only* add the list-join widget, and not the attribute as well. Except, I don't believe the list-join widget is very new-user friendly. A simple list is this:
```
<$list filter="Daffy Donald Howard Wade">
<$list-template><<currentTiddler>></$list-template>
<$list-join>, </$list-join>
</$list>
```
Or this if they're willing to cram it all on one line
```
<$list filter="Daffy Donald Howard Wade"><<currentTiddler>><$list-join>, </$list-join></$list>
```
I would think new users are going to instantly gravitate toward the attribute instead. I'm not saying that the widget is bad. Now that I've played around with it a bit, it ain't that terrible. My concern is with introducing *two* methods to do the same thing in the same release. Especially since it sounds like you think the attribute is inferior.

If they're both definitely going in, and your concerned about parsing attributes as wikitext, couldn't the attribute in theory just be plain-text? Honestly, that would cover probably 90%+ use cases for casual users. If they want anything more elaborate, then use the widget.

That's probably the most constructive thing I can say. If it were me, I'd only bring one of these in at a time and see if the first answers the need users have. (Or I'd tell them to use {{{ ...[join[, ]] }}} like the rest of us old-school users.) But it is not up to me, and I'm starting to realize that much of my negative reaction to this release stems from problems I have with the v5.3.0 release, and how I haven't finished migrating to that yet.

"That gave me `Daffy Donald Howard Wade`, which is a good start, but how do I get a comma-separated list?" *research* "Oh, I need to use something called `$list-join`. All right. Let's try it."

Jermolene:
Thanks @flibbles 

> @Jermolene. I appreciate that. Wikitext in attributes is clumsy. So that would argue that one should _only_ add the list-join widget, and not the attribute as well.

The trouble is that the ability to compute the attribute dynamically is very useful in some situations. It's the easiest way to compose wikitext from a string.

> If they're both definitely going in, and your concerned about parsing attributes as wikitext, couldn't the attribute in theory just be plain-text? Honestly, that would cover probably 90%+ use cases for casual users. If they want anything more elaborate, then use the widget.

@rmunn made the same suggestion, but I am concerned that it would be inconsistent. We ended up thinking about two attributes, one wikitext and one plaintext, but that seemed like too much. In any case, the overhead of parsing a simple string like `,` is fairly insignificant in any case.

rmunn:
I was also able to cache the results of parsing the attribute, so that it only has to be re-parsed if the attribute value changes or if the list widget is completely torn down and rebuilt. So a case like `<$list filter=<<something>> join=","/>` will only have to pay the cost of parsing the attribute once. Every time the result of the "something" filter changes, the list will change but the list joiner will not have to be re-parsed. (But if you do `join={{tiddler!!field}}` then it will re-parse the joiner wikitext every time that field changes, as it should).

So I ended up agreeing with @Jermolene that a `joinText` attribute was unnecessary. If there's demand for it, though, it can be added in 5.3.3 or later.

Jermolene:
It turns out that there's a problem with these changes when saving lazily loaded tiddlers - see #7896

saqimtiaz:
@rmunn @Jermolene just a heads up that despite the fixes in #7897, the changes herein to core templates are not backwards compatible and we should be aware that users may run into problems, though it feels too late to roll them back.

I spent well over an hour today debugging a problem that was due to the changes to `$:/core/templates/html-json-tiddler` and I expect that anyone reusing core templates is likely to run into similar problems.

Jermolene:
> @rmunn @Jermolene just a heads up that despite the fixes in #7897, the changes herein to core templates are not backwards compatible and we should be aware that users may run into problems, though it feels too late to roll them back.
> 
> I spent well over an hour today debugging a problem that was due to the changes to `$:/core/templates/html-json-tiddler` and I expect that anyone reusing core templates is likely to run into similar problems.

Ouch. That is a subtle problem that we have faced before. It would be a shame to roll back the changes to `$:/core/templates/store.area.template.html.tid`. In your use case, would it be enough just to restore the "counter" variable?

saqimtiaz:
I have resolved my issue by not using the core template anymore. When I get the chance, I will have a look to see if there any minimal steps we could take to decrease the likelihood that others will run into the same problem.

Issue #7696, opened 2023-08-24
[BUG] Transcluded functions do not dynamically refresh


pmario:
While arguing with user: rmunn about the join feature in list widgets I did find and update problem with `functions`. But I'm not sure if that's intended or a bug. 

To reproduce

- Copy the following code into a tiddler at tiddlywiki.com
- Open the Preveiw Panel
- Create 10 items with <kbd>Create</kbd> button
- -> **Problem: The `<<list-join "[enlist{!!items}]">>` is not update**
- Close preview
- Open prevew
- now it is updated. 
- add new elements into the top Items: input box
- -> Problem: **The links will be updated, but the text is not**

@Jermolene   **Is this behaviour intended?**

```
\function list-join(filter, sep:", ") [subfilter<filter>join<sep>]

Items: <$edit field="items" /> <br/>

<<list-join "[enlist{!!items}]">>

---

<$let joined-list=<<list-join "[enlist{!!items}]">> >
	<$list filter="[<joined-list>enlist-input[]]" variable="item" >
		<$link to={{{ [<item>removesuffix[,]else<item>] }}}><<item>></$link>
	</$list>
</$let>

Create <$edit field="itemcount" />
items <$button action>Create<$action-setfield items={{{ [range{!!itemcount}join[ ]] }}}></$action-setfield></$button>
```

Jermolene:
Hi @pmario good catch, I think this is indeed a bug.

Jermolene:
I made a failing test over at #7698

rmunn:
I left a comment at https://github.com/Jermolene/TiddlyWiki5/pull/7698#issuecomment-1702144333 with what I think is the first step towards a solution.

AnthonyMuscio:
And as mario saw, I reported something very similar

If you have a tiddler you are editing, that contains a function definition, that makes use of a field value, and you are viewing the result in the Preview > Output most other forms of definition and wiki text are immediately reflected in the preview.

However if you change the value in a field used by a function, it will not update until you either save and reopen the tiddler, or close and reopen the preview.

Is this a bug or a feature?_

A related piece of evidence is us you have the additional edit buttons, save but not close, will not result in the functions result updating. I thought that did do a refresh, because we lose the cursor position.

rmunn:
@AnthonyMuscio - Yes, that's a bug. Anytime a filter would refresh, a function containing that filter should also refresh.

[...]

PR #7698, opened 2023-08-25
Fix refreshing of transcluded functions

Jermolene:
This is currently just a failing test for #7696


rmunn:
There's a comment in the transclude widget that its refresh "Doesn't need to consider transcluded variables because a parent variable can't change once a widget has been created". But that's not quite true of functions: their *text* can't change once the widget has been created, but their *result* can. So I think the refresh method of transclude widgets needs to check srcVariable.isFunctionDefinition, and if it is, call getVariableInfo again to get the new results. Then compare the old results to the new results and, if they are different, the widget needs to refresh.

BTW, a quick note about comparing the old results to the new results. Functions, like any other filter, could potentially return hundreds, or even thousands, of results. The point of comparing the old results to the new results is to skip a rerender if the results are the same, but that optimisation might fail if the cost of comparing the results is more than the cost of rerendering. So if the length of the results list is large, say more than a hundred items or so, it might be more efficient to just skip the comparison and say "yes, rerender". Performance testing would be needed to see where the inflection point is. But certainly if the list of results is fewer than about twenty items or so, it'll be worth it to compare the result lists and skip a rerender if the function result didn't change.

So I suspect that the `!this.transcludeVariable` check in the refresh() method needs to be changed, with a `functionNeedsRefresh()` method added:

https://github.com/Jermolene/TiddlyWiki5/blob/fa9bfa07a095548eb2f8339b0b1b816d2e6794ef/core/modules/widgets/transclude.js#L436-L444

rmunn:
@Jermolene I opened https://github.com/Jermolene/TiddlyWiki5/pull/7755 to implement the `functionNeedsRefresh()` solution I proposed in my earlier comment. Note that PR #7755 merges into **this** PR's branch, i.e. if you merge #7755 it will add that commit to **this** PR, which can then be merged into master once you verify that the bug is indeed fixed. (I did it that way because I couldn't push a commit to your `fix-refreshing-transcluding-functions` branch directly as I don't have write access to your repo, so that was the simplest way I could find to add a commit to your PR branch).

My commit does, indeed, fix the bug, as verified by the test you added. Hopefully you can just merge it into this PR and save yourself some debugging work.

Jermolene:
Thanks @rmunn, #7755 is a good fix, much appreciated.

Jermolene:
Thanks @rmunn

PR #7702, opened 2023-08-29
Slightly speed up the `all` filter operator

rmunn:
The `all` filter operator has shortcuts to optimise common patterns like `[all[shadows+tiddlers]]` or `[all[tiddlers]]`. In those cases, the filter operator function returns early and never uses the `result` linked list that was created, so it's immediately garbage-collected. This PR attempts to speed up the result of calling `all` in those common cases by only creating that linked list if it's actually going to be used, avoiding that garbage collection in the common cases that are optimised to return early.

### DIFF
--- a/core/modules/filters/all.js
+++ b/core/modules/filters/all.js
@@ -28,12 +28,8 @@ function getAllFilterOperators() {
 Export our filter function
 */
 exports.all = function(source,operator,options) {
-	// Get our suboperators
-	var allFilterOperators = getAllFilterOperators();
-	// Cycle through the suboperators accumulating their results
-	var results = new $tw.utils.LinkedList(),
-		subops = operator.operand.split("+");
 	// Check for common optimisations
+	var subops = operator.operand.split("+");
 	if(subops.length === 1 && subops[0] === "") {
 		return source;
 	} else if(subops.length === 1 && subops[0] === "tiddlers") {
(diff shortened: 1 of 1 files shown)


rmunn:
All unit tests still pass with this PR. I have not yet figured out how to measure the performance benefits you get from this optimisation. I expect it to be relatively small, as creating a `$tw.utils.LinkedList` object allocates about three objects (the linked list and its `next` and `prev` properties, which are LLMap objects) with about 5-6 allocated properties total. That's not a huge amount of memory to churn: a couple dozen bytes at most. However, the `[all[shadows+tiddlers]]` filter is used all over the place in TiddlyWiki's internals, and every one of those filters is currently allocating a LinkedList and then immediately throwing it away without using it. So those dozens of bytes, and the work the Javascript garbage collector has to do to keep track of them, will eventually add up to some measurable amount of time, I just don't know how large a wiki will need to be before it's noticeable.

rmunn:
@flibbles - You originally created https://github.com/Jermolene/TiddlyWiki5/pull/5362 in order to optimise the performance of the `all` operator, and this PR is simply building on your work a little. In that PR, you wrote "There's more room for improvement by cutting down all that garbage collection ...", and I think this is a small step in that direction. Since you already have a test wiki set up for performance measuring (or at least, you did at the time you created #5362), would you be interested in using it to measure the speedup obtained by this small change?

flibbles:
I do not have a test wiki set up. Your change seems pretty simple. It either helps or it does nothing. It won't hurt. I'm not sure it needs a run of performance testing.

Jermolene:
Thanks for your patience @rmunn

Issue #2144, opened 2015-12-22
In-text semantic annotation of links


felixhayashi:
**Motivation:**

* This would bring us one step closer to a semantic wiki.
* Allows users to expose the meaning of links in a machine readable way.
  * …so e.g. plugins like TiddlyMap can make more sense of it.

**How others do it**

Semantic MediaWiki uses the following notation

```
Make [[Example property::link|alternate text]] appear in place of the link.
```

See https://www.semantic-mediawiki.org/wiki/Help:In-text_annotation

**What needs to be done:**

* Extend existing regular expression for parsing links
* Extend `$tw.wiki.getTiddlerLinks` or create a new function

**Related**

* https://github.com/felixhayashi/TW5-TiddlyMap/issues/119
* https://github.com/Jermolene/TiddlyWiki5/issues/1749

**People previously involed**

@tobibeer @Jermolene @ybabel

Jermolene:
Hi @felixhayashi seems like a reasonable idea.

ybabel:
If I'm not mistaken, SMW link are reversed from TW's : 

 * `[[Displayed Link Title|Tiddler Title]]`
 * `[[Example property::link|alternate text]]`

"alternate text" is equivalent to our "Displayed Link Title"

So I think the right syntax for us should be rather (in order to keep ascendant compatibility): 
 * `[[Property::Displayed Link Title|Tiddler Title]]`
 * `[[Displayed Link Title|Property::Tiddler Title]]`

not a big deal I think ;-) I'm just mentioning

felixhayashi:
> not a big deal I think ;-) I'm just mentioning 

good point! should be `[[Displayed Link Title|Property::Tiddler Title]]` then. `[[Property::Displayed Link Title|Tiddler Title]]` is not appropriate here as the property should be paired with the real tiddler title and not with the alias.

tobibeer:
>  the property should be paired with the real tiddler title and not with the alias

While that is true, I don't quite think I would overload the wikitext syntax either way. I would rather try and find an appropriate spot to extend on it, something like:

```
[[pretty|title]{property:value;prop:val}]
```

felixhayashi:
`[[pretty|title]{property:value;prop:val}]`

Why does a semantic annotation need `value` and what is its meaning?

[...]

PR #7703, opened 2023-08-29
added prettyanotatedlinks

m42e:
Start of implementation of https://github.com/Jermolene/TiddlyWiki5/issues/2144

This will allow the last discussed syntax. But I am not sure how to be able to use that in filters or get the attribute of the link in a filter expression. Could you give me a hint @Jermolene 

Thanks.


pmario:
I just stumbled upon a response, that I did start some time ago, but did not finish it. 
I think it's still valid but "not finished" 
So here it is. 

------

Hi Matthias @m42e, 
Nice work! -- I did a bit of testing and there are some things that I want to point out. 

- The filter operator names: `annotatedbacklinks` and `annotatedlinks` are really hard to read 
  - since we have [standard-deviation](https://tiddlywiki.com/#standard-deviation%20Operator), which was introduced by Jeremy with PR #5612, it should be possible to use the same mechanism for your new operators too
- @Jermolene -> I do suggest we use `exports["annotated-backlinks"]` and `exports["annotated-links"]` to make them more readable.

---

- In your code you use `function parseAnnotations(annotationstring){` 
  - which should be `function parseAnnotations(annotationString){` -- see `annotationString` for the function parameter for readability

----

- For TW atm we still use ES5, which does not allow `let`. I know you did use `let` for a reason, but the core code needs consistency.
- So use `var` instead of `let` for consistency. 

----

- ` }else{` should be ` } else {`
- In TW we use double-quotes for strings eg: `.split(",")`
   - **except** if we need a double-quote for user facing strings to avoid escape-srings in the code eg: `'We use "double quotes"'`

----

- There is a `split(",")` in `annotationstring.split(',').forEach(function(part){` in annotatebacklinks.js **but** there is no example for the link syntax when it should be needed. 
- I assume it is used for `[link asdf="asdf", test=test []]`
- **But** `[link test=asdf xxx=aaa [HelloThere]]` seems to work for me too ?! -> So there is something strange going on. 

IMO we should use the HTML parameter parser instead of `parseAnnotations()`

----

- `string.includes` IMO is ES6+ so we use `"aaa bbb ccc".indexOf("test") !== -1` to test for partial strings,
- or we use regexp's

----

- Calculating any type of backlinks is very performance heavy. So if possible backlinks should be cached.

PR #7708, opened 2023-08-31
tm-open-window set focus to existing window

### LINKED ISSUE #7707: [BUG] FireFox: tm-open-window does not focus the window if it already exists
Also see: https://talk.tiddlywiki.org/t/reopen-window-with-tm-open-window-message/7910

>I seem to recall being able to use a button with tm-open-window a second time to reopen the window again.
>
>- Anyone know how we could make this happen?
>- tm-close-window works so perhaps I could close then open again but that seems wasteful

It seems to be auto-focused in Chrome-like browsers, but FireFox needs the `window.focus()` function.

PR will be created soon

pmario:
This PR fixes #7707. It sets the focus to the existing window if it already exists but is not visible atm.

### DIFF
--- a/core/modules/startup/windows.js
+++ b/core/modules/startup/windows.js
@@ -40,7 +40,7 @@ exports.startup = function() {
 			variables = $tw.utils.extend({},paramObject,{currentTiddler: title, "tv-window-id": windowID});
 		// Open the window
 		var srcWindow,
-		    srcDocument;
+			srcDocument;
 		// In case that popup blockers deny opening a new window
 		try {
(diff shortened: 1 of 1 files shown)


pmario:
It seems to be auto-focused in Chrome-like browsers, but FireFox needs the window.focus() function.

Jermolene:
Looks good thanks @pmario. Perhaps it's worth a docs update just to note the new behaviour?

pmario:
@Jermolene ... I did update the docs

pmario:
bump. Docs has been updated over a month ago.

Jermolene:
Thanks @pmario

Issue #7680, opened 2023-08-16
[IDEA] Pragmas allowed everywhere, not just at start of parsing context


rmunn:
**Is your feature request related to a problem? Please describe.**
The idea of an if...else or if...then...else construct has been brought up several times. As part of that discussion, @Jermolene proposed an [`\if` ... `\else` ... `\endif` pragma](https://talk.tiddlywiki.org/t/switch-case-construct/2234/13), which would necessarily require that [pragmas would be allowed anywhere in wikitext](https://talk.tiddlywiki.org/t/switch-case-construct/2234/20), except for the `\rules` pragma which would still need to appear at the top of a tiddler.

There are several use cases that this would allow:
* `\if`...`\else`...`\endif`
* `\switch`...`\case`...`\case`...`\default`...`\endswitch`
* Defining procedures/functions/custom widgets in the middle of a tiddler
* Probably many others

Since this idea would be broadly useful, I think it's worth tracking the idea in a GitHub issue.

**Describe the solution you'd like**
The WikiText parser is rewritten to allow pragmas to appear anywhere, except for the ones that change the parser rules

**Describe alternatives you've considered**
The `\if` and `\swtich` use cases could be done with widgets instead:

```plaintext
<$if filter="condition">
  <$then>Value if true</$then>
  <$else>Value if false</$else>
</$if>

<$switch filter="value to test">
  <$case filter="one">Result if one</$case>
  <$case filter="two">Result if two</$case>
  <$default>Result if no match</$default>
</$switch>
```

See https://github.com/Jermolene/TiddlyWiki5/issues/6862#issuecomment-1200449657 for the source of this idea.

However, there is currently no way to define procedures, custom widgets, and so on in the middle of a tiddler, so no alternative exists for this one.

**Additional context**
Previous discussion of the if...else or switch widgets (or closely related ideas):
* https://github.com/Jermolene/TiddlyWiki5/issues/272
* https://github.com/Jermolene/TiddlyWiki5/issues/1248
* https://github.com/Jermolene/TiddlyWiki5/issues/3463
* https://github.com/Jermolene/TiddlyWiki5/issues/4064
* https://talk.tiddlywiki.org/t/switch-case-construct/2234
* https://github.com/Jermolene/TiddlyWiki5/issues/6862

[...]

PR #7710, opened 2023-09-01
Conditional Shortcut Syntax

Jermolene:
This PR adds a new shortcut syntax for expressing if/then/else logic. It takes a different approach to #7691. Instead of adding a new widget, it adds a new parser rule for a shortcut syntax that expands to a series of nested `<$list>` widgets. For example:

```text
\procedure test(animal)
<% if [<animal>match[Elephant]] %>
  It is an elephant
<% else %>
  <% if [<animal>match[Giraffe]] %>
    It is a giraffe
  <% else %>
    It is completely unknown
  <% endif %>
<% endif %>
\end

<<test "Giraffe">>

<<test "Elephant">>

<<test "Antelope">>
```

Behind the scenes, the `<$list>` widget is extended with two new features:

* The option to specify the "emptyMessage" via a `<$list-empty>` widget within the list widget, instead of using the attribute, and similarly to use a `<$list-template>` widget to define the template
* The option to limit the number of results returned from the filter

There are also some changes to the core parser to support the new syntax.

# Progress

* [x] if-else-endif
* [x] elseif
* [x] Add support for block mode content
* [ ] Decide on a name for the generic `<%...%>` syntax (ie the term we would also use for other constructions like switch/case that use the same syntax)
* [x] Improve documentation


pmario:
Is the `limit` parameter needed for the if/else shortcut, or is it just an extension of the list-widget

pmario:
Ignore the question. I found it in the code and in the parse- and widget-tree

pmario:
I think I found a bug. I think the following example should result in `123` as a result. 
I think `<$list-template>` and `$list-empty` should be optional.

Especially since `emptyMessage` is a "pain point" in the current list-widget implementation.

```
\whitespace trim

\procedure test(filter)
<$list filter=<<filter>> >
	<$text text=<<currentTiddler>>/>
	<$list-empty>
		None!
	</$list-empty>
</$list>
\end

<<test "1 2 3">>
```

image

rmunn:
Here's one test I wrote for the `<$if>` widget, which I removed from the PR because it was spamming the console log, but which might be useful in testing the `{@ if @}` syntax as well. It uses `<$log>` widgets to make sure that only one branch of the whole if...elseif...elseif...else chain is ever wikified:

```js
it("if widget should short-circuit child rendering, so only one child ever gets rendered", function() {
    var wiki = new $tw.Wiki();
    // Construct four widget nodes
    var text = "<$if filter='[[1]compare:number:eq[2]]'><$then>yes<$log message='then' /></$then><$elseif filter='[[3]compare:number:eq[4]]'>maybe<$log message='elseif1' /></$elseif><$elseif filter='[[5]compare:number:eq[6]]'>another<$log message='elseif2' /></$elseif><$else>no<$log message='else' /></$else></$if>";
    var falsey = text;
    var elseif2 = falsey.replace("eq[6]]","eq[5]]");
    var elseif1 = elseif2.replace("eq[4]]","eq[3]]"); // Both second and first elseifs are true, first one wins
    var truthy = elseif1.replace("eq[2]]","eq[1]]"); // The "main" if condition and both elseifs are true, "main" condition wins
    var widgetNodeF = createWidgetNode(parseText(falsey,wiki),wiki);
    var widgetNodeT = createWidgetNode(parseText(truthy,wiki),wiki);
    var widgetNodeE1 = createWidgetNode(parseText(elseif1,wiki),wiki);
    var widgetNodeE2 = createWidgetNode(parseText(elseif2,wiki),wiki);
    // Render four widget nodes to the DOM
    console.log('Rendering F');
    var wrapperF = renderWidgetNode(widgetNodeF);
    console.log('Rendering T');
    var wrapperT = renderWidgetNode(widgetNodeT);
    console.log('Rendering E1');
    var wrapperE1 = renderWidgetNode(widgetNodeE1);
    console.log('Rendering E2');
    var wrapperE2 = renderWidgetNode(widgetNodeE2);
    // Test the rendering of all four
    expect(wrapperF.innerHTML).toBe("<p>no</p>");
    expect(wrapperT.innerHTML).toBe("<p>yes</p>");
    expect(wrapperE1.innerHTML).toBe("<p>maybe</p>");
    expect(wrapperE2.innerHTML).toBe("<p>another</p>");
});
```

Expected output if all goes well is "Rendering F" followed by exactly *one* `<$log>` output, then "Rendering T" followed by exactly *one* `<$log>` output, and so on for E1 and E2. I chose the `<$log>` widget because that was the only way I could see to have something trigger when the widget is rendered; everything else (like action-setfield) only triggers on a button press. I would have preferred to test the widget by having an action change a field or set a variable, then checking the value of that variable after rendering the `<$if>` widget... but I couldn't find a way to do that. So I don't have an automated "this is short-circuiting properly" test, just one that spams the console log and then I look at the console log to make sure it printed the right thing.

Anyway, I hope this is somewhat helpful in building a similar test for the list widgets built with the `{@ if @}` syntax, to make sure they short-circuit properly and don't even build the widgets for the parse tree branches that aren't selected.

pmario:
The parse tree can be checked directly. See eg: [test-wikitext-parser.js](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-wikitext-parser.js)

rmunn:
Thanks, but what I'm trying to test is that certain parts of the parse tree do not get turned into widgets. Checking the parse tree won't help with that.

pmario:
It's also possible to check the parse-tree, which only contains the widgets, which will be active

kookma:
I am not sure if this question is relevant, but the underlaying $list widget shows the below error when an empty line is inserted:

```
\define filter() [tag[HelloThere]]


<$list filter=<<filter>>>

	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
</$list>
```

shows

```
Undefined widget 'list-template' Undefined widget 'list-empty' Undefined widget 'list-join'
```

but without extra empty line, it works (see below)

```
\define filter() [tag[HelloThere]]


<$list filter=<<filter>>>
	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
</$list>
```

It seems adding empty line between $list and $list-template, and $list-empty shows the above error.

Jermolene:
> I think I found a bug

Hi @pmario that is a case that I wasn't sure how to handle: what to do if the `<$list-empty>` widget is found, but not the `<$list-template>` widget. Perhaps in such cases it should behave as if there was an empty `<$list-template>` widget.

Jermolene:
> Here's one test I wrote for the `<$if>` widget, which I removed from the PR because it was spamming the console log, but which might be useful in testing the `{@ if @}` syntax as well.

Thanks @rmunn that is helpful.

> I would have preferred to test the widget by having an action change a field or set a variable, then checking the value of that variable after rendering the `<$if>` widget... but I couldn't find a way to do that.

I'm not sure that it is enough for our purposes here, but the new(ish) wiki-based test framework supports actions that are performed after the initial rendering to force a refresh.

https://github.com/Jermolene/TiddlyWiki5/blob/2c406da823559e4777968f328bba1cfa3d2dd9e0/editions/test/tiddlers/tests/data/transclude/Variable-Refreshing.tid#L1-L27

Jermolene:
> I am not sure if this question is relevant, but the underlaying $list widget shows the below error when an empty line is inserted:

Thanks @kookma this is an interesting one. At the moment, the list widget only looks at its immediate children when searching for the `<$list-template>` and `<$list-empty>` pseudo widgets. The extra blank line causes the content of the list widget to be parsed in block mode, which means that its content is wrapped in `<p>` tag, which means that the `<$list-template>` and `<$list-empty>`  are no longer immediate children of the list widget. Thus, they are not found, and the list widget assumes that the entire content of the list widget is the item template. The undefined widget errors are raised because the `<$list-template>` and `<$list-empty>` widget are only pseudo widgets, and are not recognised outside of their role within the list widget.

One possible fix would be to do a recursive search for the `<$list-template>` and `<$list-empty>` pseudo widgets, so that they would be found even if they are wrapped in a `<p>` tag.

The trouble with that approach is that it might have unexpected consequences. For example, here the inner list widget will always evaluate to an empty list, and thus its content will never be rendered. However, if we supported recursive searching then the outer list widget would still find the `<$list-template>` and `<$list-empty>` widgets. Now, we could make a special case for not searching within nested list widgets, but that doesn't allow for other widgets that may have similar functionality.

```
<$list filter="...">
<$list filter="[[never]match[something]]">
<$list-template>
This is a template
</$list-template>
</$list>
</$list>
```

Perhaps the best we can do is to make HTML elements explicitly transparent when searching for `<$list-template>` and `<$list-empty>` widgets, but make all other widgets be opaque.

We'd still be left with the problem that a stray `<$list-template>` or `<$list-empty>` widgetswidget will display an undefined widget error. We could improve the situation by defining those widgets as an alias of the existing `<$error>` widget.

rmunn:
> Hi @pmario that is a case that I wasn't sure how to handle: what to do if the `<$list-empty>` widget is found, but not the `<$list-template>` widget. Perhaps in such cases it should behave as if there was an empty `<$list-template>` widget.

Unless there's a reason not to, I'd suggest handling that case in line with the existing behavior: "When no template is specified, the body of the [ListWidget](https://tiddlywiki.com/#ListWidget) serves as the item template." If there's a `<$list-empty>` widget it's removed, then the rest of the body is treated as if it was in the `<$list-template>` widget.

As for the block-mode parsing issue that @kookma found,

> One possible fix would be to do a recursive search for the `<$list-template>` and `<$list-empty>` pseudo widgets, so that they would be found even if they are wrapped in a `<p>` tag.

I'd suggest being slightly more limited, and just do a search for either children, or grandchildren where the immediate parent is a `<p>` tag. That should be enough to cover this scenario. And since the `<$list-template>` documentation already says that it should be an immediate child of the ListWidget, there's no need to test for cases like "What if it's inside a `<strong>` or `<em>` tag?" because the documentation says not to do that, so anyone who does that shouldn't be surprised that it doesn't work. (The extra `<p>` element wrapped around it is a bit of an exception, because it's invisible in the WikiText syntax and so people wouldn't expect that they would be adding an element merely by inserting a blank line).

Jermolene:
> If there's a `<$list-empty>` widget it's removed, then the rest of the body is treated as if it was in the `<$list-template>` widget.

Hi @rmunn my hesitation with that approach is that we don't have a convenient way to remove parse tree nodes from a parse tree. It's complicated because caching means that parse trees are shared, and so modifications made to a parse tree by one client of the parser will be visible to other clients. We can get around all of that, but it's a lot of code (and tests) for situations that shouldn't really occur).

> I'd suggest being slightly more limited, and just do a search for either children, or grandchildren where the immediate parent is a `<p>` tag. 

Yes that may make more sense.

pmario:
>I'd suggest being slightly more limited, and just do a search for either children, or grandchildren where the immediate parent is a <p> tag.

Or we fix the P-tag problem and stop working around it.

Jermolene:
I've made a couple of updates in [a1acef7](https://github.com/Jermolene/TiddlyWiki5/pull/7710/commits/a1acef7324676768a07aa2e89d98c3f509e8d699) and [7758601](https://github.com/Jermolene/TiddlyWiki5/pull/7710/commits/775860110c4ae8ec748dcbd16f473e562867dd77):

* If the `<$list-empty>` widget is present but not the `<$list-template>` widget then it behaves as if an empty list template were present. This should fix the [bug](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1703163003) that @pmario reported
* The `<$list>` widget will now recursively search into `<p>` tags to locate the `<$list-template>` or `<$list-empty>` widgets. This should fix the [bug](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1711734198) that @kookma reported

rmunn:
> I've made a couple of updates in [a1acef7](https://github.com/Jermolene/TiddlyWiki5/pull/7710/commits/a1acef7324676768a07aa2e89d98c3f509e8d699) and [7758601](https://github.com/Jermolene/TiddlyWiki5/pull/7710/commits/775860110c4ae8ec748dcbd16f473e562867dd77):
> 
>     * If the `<$list-empty>` widget is present but not the `<$list-template>` widget then it behaves as if an empty list template were present. This should fix the [bug](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1703163003) that @pmario reported

That's counterintuitive. The current behavior, minus a `<$list-empty>` widget, is to say "the body of the list widget is the template". Which is why @pmario's said in his [bug report](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1703163003) that he thought the output should be "123". If a missing `<$list-template>` in the presence of a `<$list-empty>` option is treated as empty, that would make the output of @pmario's example empty rather than "123".

And I agree with @pmario. User expectation is going to be that these two usages of the list widget will produce the same result:

```plaintext
<$list filter="1 2 3" variable="item">
<<item>>
<$list>

<$list filter="1 2 3" variable="item">
<<item>>
<$list-empty>No items given</$list-empty>
<$list>
```

kookma:
> * The `<$list>` widget will now recursively search into `<p>` tags to locate the `<$list-template>` or `<$list-empty>` widgets. This should fix the [bug](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1711734198) that @kookma reported

Thank you! The new release works. Just note to the output (raw html preview):

Case (1) 
```
\define filter() [tag[HelloThere]]

<$list filter=<<filter>>>

	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
</$list>
```
produces
```

		A Gentle Guide to TiddlyWiki
	
		Discover TiddlyWiki
	
		Some of the things you can do with TiddlyWiki
	
		Ten reasons to switch to TiddlyWiki
	
		Examples
	
		What happened to the original TiddlyWiki?
	
		Funding TiddlyWiki
	
		Open Collective
	
```

Case (2) no line break between $list and $list-template
```
<$list filter=<<filter>> >
	<$list-template>
		<$text text=<<currentTiddler>>/>
	</$list-template>
	<$list-empty>
		None!
	</$list-empty>
</$list>
```
produces
```
<p>
		A Gentle Guide to TiddlyWiki
	
		Discover TiddlyWiki
	
		Some of the things you can do with TiddlyWiki
	
		Ten reasons to switch to TiddlyWiki
	
		Examples
	
		What happened to the original TiddlyWiki?
	
		Funding TiddlyWiki
	
		Open Collective
	</p>
```

So, with no line breaks the output is wrapped into extra pair of `<p>` (generate extra `<p>` tags).

But no error is reported.

Jermolene:
> Thank you! The new release works. Just note to the output (raw html preview):

Hi @kookma my apologies, yesterday I committed some of the changes that are in progress for the block mode handling. The behaviour you're seeing is a result of that change.

However, do bear in mind that that a double line break after an opening tag always triggers block more for the parsing of the content of the tag. So an example like the following will render differently on tiddlywiki.com depending upon whether there is a double line break after the opening `<$list>` tag:

```
<$list filter="1 2 3">
	<div>something</div>
</$list>
```

I'll complete the in-progress changes and post an update here.

Jermolene:
I've pushed an update to add support for block mode content within an if/else/elseif clause.

@kookma the [example you posted](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1718758154) still produces the same output, but as I say that is a consequence of how the list widget and its content is parsed, and is not caused by these changes.

kookma:
> @kookma the [example you posted](https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1718758154) still produces the same output, but as I say that is a consequence of how the list widget and its content is parsed, and is not caused by these changes.

Tested new preview and works as described. Thank you

linonetwo:
I found it hard to type `<%` because after typing `<`, it is hard to move eyes and right hands across have of keyboard to find `%` (left hand is always holding Shift!)

Can it be two same char just like `{{` `[[` `<<` ? For example `((` or `[[[`

Jermolene:
Hi @linonetwo 

> Can it be two same char just like `{{` `[[` `<<` ? For example `((` or `[[[`

Double parenthesis is currently unused and so might be a promising choice:

```
\procedure test(animal)
((if [<animal>match[Elephant]] ))
  It is an elephant
((else))
  ((if [<animal>match[Giraffe]] ))
    It is a giraffe
  ((else))
    It is completely unknown
  ((endif))
((endif))
\end

<<test "Giraffe">>

<<test "Elephant">>

<<test "Antelope">>
```

pmario:
>Double parenthesis is currently unused and so might be a promising choice:

I think standard double parenthesis are way to common in prose text. Especially if someone explains an algorithm with "pseudo syntax" and IMO it's very close to [LISP](https://lisp-lang.org/learn/functions) which may cause confusion.

linonetwo:
> are way to common in prose text

This is not common in daily writing, if encounter this, maybe can use """ to wrap the.

> explains an algorithm

Also can wrap them with ``` code block

I think this syntax won't work in a block.

linonetwo:
@Jermolene Have you considered `/if /endif` syntax? Currently fnprocdef import macrodef are all using this syntax and each has its own parser, so there can also be a new parser for it. (And there may be more in the future).

One drawback is these `pragma` can only be used on top of tiddler, discussed in https://github.com/Jermolene/TiddlyWiki5/issues/7680

Jermolene:
> I think standard double parenthesis are way to common in prose text.

Really? I would be interested to understand why that might be. I would not expect to encounter double parenthesis in general prose, but only in technical contextx.

> Especially if someone explains an algorithm with "pseudo syntax" and IMO it's very close to [LISP](https://lisp-lang.org/learn/functions) which may cause confusion.

Hmmm those needs are readily catered for with the codeblock syntax, or the triple double quote syntax. I don't think that a rare use case like that should drive the choice of syntax that would affect far more users.

> @Jermolene Have you considered `/if /endif` syntax? Currently fnprocdef import macrodef are all using this syntax and each has its own parser, so there can also be a new parser for it. (And there may be more in the future).
> 
> One drawback is these `pragma` can only be used on top of tiddler.

There are two problems: backslash at the start of a line indicates a pragma, and pragmas have the restriction that they must precede ordinary parsed text. The other issue is that there is no lineline syntax for pragmas.

rmunn:
> > I think standard double parenthesis are way to common in prose text.
> 
> Really? I would be interested to understand why that might be. I would not expect to encounter double parenthesis in general prose, but only in technical contextx.

I have sometimes seen double parentheses used as a way of making a note to self in some text to come back and fix later. E.g,, in a novel in progress, the author might write: 

> The detective examined the bullet wounds. "Looks like the victim was shot with a .22 pistol," he said. "Either a ((model)) or a ((model)) would be my guess, but forensics should tell us more." ((TODO: Figure out a couple of plausible models)).

Then later when he's done writing the scene, he'd go online and do a bit of research and decide which models are plausible for a criminal to be able to get hold of easily in ((name of town)). I know a couple people (online and in real life) who have written books, and some of them have mentioned that they use this approach a lot so they don't have to interrupt the flow of words to go and do fiddly research; they leave the research until a time when inspiration isn't hitting.

So I would agree with avoiding double parentheses for anything. It's used fairly often as far as I've seen, precisely because it doesn't occur in normal text but it's extremely easy to type, and easy to search for later when you're looking for your notes-to-self.

linonetwo:
> backslash at the start of a line indicates a pragma

Is it hard to relax this limitation? Or use `\\` for `\\if` ?

Jermolene:
> I have sometimes seen double parentheses used as a way of making a note to self in some text to come back and fix later. E.g,, in a novel in progress, the author might write:

I would actually take the fact that double parenthesis is used as ad-hoc markup as validation: it demonstrates that it is a construction that is rarely used in prose, and is not visually distracting.

> > The detective examined the bullet wounds. "Looks like the victim was shot with a .22 pistol," he said. "Either a ((model)) or a ((model)) would be my guess, but forensics should tell us more." ((TODO: Figure out a couple of plausible models)).

Of course, in TiddlyWiki we would suggest using `{{model}}`.

> Is it hard to relax this limitation? Or use `\\` for `\\if` ?

It is a limitation in the sense that it needs to be consistent with the new syntax. A syntax that closely resembles the pragma syntax but has subtly different rules seems like it would be confusing.

Jermolene:
I've cherry picked the list widget enhancements into a new PR in #7784

CodaCodr:
> I've ended up going for "shortcut syntax", meaning it's a shortcut for the underlying widget syntax.

IMO, that's not a good reason for choosing the name. Fresh eyes don't know the background or (perhaps) understand the reasoning. Fresh eyes certainly don't need to know that baggage to use it.

If, as you mentioned, there are more shortcuts to come, you (we) need a better name, IMO.

Jermolene:
Hi @CodaCodr the requirement here isn't to come up with a word or phrase that will be self explanatory. We just need to name the concept, which means a label that is unique/distinctive within the context. In this case, we have an opportunity to choose a name that leads the user towards a deeper understanding of what is going on.

pmario:
Try this. IMO your second example shows a CSS link so the formatting is overwritte. 

```
''<$list filter="1" > <<currentTiddler>> </$list>''
```

Have a closer look to the parse-tree preview. You'll see it's similar to the first example

CodaCodr:
I strongly urge you to reconsider (the effect on fresh eyes). This is the kind of thing that can foster cognitive overload in a newcomer -- "will I _ever_ understand this stuff?"

That's not to say that the background/underlying "truth" shouldn't be made available and attention drawn to it, but _elsewhere_, perhaps behind "Learn more" or similar. But to place that in its name? I don't need to be reminded (nor, least of all, do you) but a newcomer?

And does it scale? 50 new pence. Windows NT. 

I'm harping on because this stuff matters. :-|

PR #7712, opened 2023-09-02
Fix list widget bug with counter-last when appending items

### LINKED ISSUE #7693: [BUG] List widget counter-last not updated correctly when items added to list

rmunn:
This PR contains just the bugfix from https://github.com/Jermolene/TiddlyWiki5/pull/7694 with no new features added to the list widget.

Fixes #7693.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -225,6 +225,8 @@ ListWidget.prototype.handleListChanges = function(changedTiddlers) {
 		// If we are providing an counter variable then we must refresh the items, otherwise we can rearrange them
 		var hasRefreshed = false,t;
 		if(this.counterName) {
+			var mustRefreshOldLast = this.children.length < this.list.length;
+			var oldLastIdx = this.children.length-1;
 			// Cycle through the list and remove and re-insert the first item that has changed, and all the remaining items
 			for(t=0; t<this.list.length; t++) {
 				if(hasRefreshed || !this.children[t] || this.children[t].parseTreeNode.itemTitle !== this.list[t]) {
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @rmunn

PR #7714, opened 2023-09-02
Test if selectedNode.offsetLeft is 0 when dragging a dom node

BurningTreeC:
this adds a necessary test in the edge-case that the offsetLeft property of the selectedNode is 0. It can be zero. For domNode.offsetWidth this test should be superfluous

### DIFF
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -312,8 +312,9 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 		$tw.utils.each(selectedNode.attributes,function(attribute) {
 			variables["dom-" + attribute.name] = attribute.value.toString();
 		});
-		
-		if(selectedNode.offsetLeft) {
+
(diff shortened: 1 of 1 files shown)


pmario:
Is there a related issue?

BurningTreeC:
Hi @pmario ,

there's no GitHub issue that's related but I've encountered an issue with this on https://burningtreec.github.io/tiddlywiki-multi-columns

The problem can be reproduced if one configures the layout in the ControlPanel (Appearance > MultiColumn) so that the story-rivers have no padding (0)

In that case, when you drag a tiddler, at dragstart the variables don't get collected (like the selectedNode's height) and you can see that the dragover-animations don't work

BurningTreeC:
That is because `selectedNode.offsetLeft` reports 0 in that case and 0 gets interpreted as false

Jermolene:
Thanks @BurningTreeC I've pushed an alternate fix in de6b866f22a8262c8c531e8fbace999563148aa2

PR #7715, opened 2023-09-02
Fix a dangling link in filter run prefix docs

rmunn:
https://github.com/Jermolene/TiddlyWiki5/pull/6307 moved the docs for the difference between `+` and `:intersection` to a new tiddler, but there's a link in `Filter Expression.tid` that pointed to the old location which PR #6307 missed. This PR updates that link to point to the current location of that particular docs section.


Jermolene:
Great thanks @rmunn

PR #7716, opened 2023-09-03
Remove Dynannotate plugin's SelectionTracker configuration

oflg:
image

The plugin documentation says that the configuration defaults to "yes", but now it is "no".

### DIFF
--- a/plugins/tiddlywiki/dynannotate/config/LegacySelectionTracker.tid
+++ b/plugins/tiddlywiki/dynannotate/config/LegacySelectionTracker.tid
@@ -1,3 +1,0 @@
-title: $:/config/Dynannotate/LegacySelectionTracker/Enable
-
-no
--- a/plugins/tiddlywiki/dynannotate/config/SelectionTracker.tid
+++ b/plugins/tiddlywiki/dynannotate/config/SelectionTracker.tid
@@ -1,3 +1,0 @@
-title: $:/config/Dynannotate/SelectionTracker/Enable
-
-no


pmario:
The PR that changed the behaviour was: https://github.com/Jermolene/TiddlyWiki5/pull/7260 

@yaisog ... What was the reasoning behind that particular change? Do you know it "by hart"?
(I did not read the PR conversation at the PR yet. So I cannot remember it all to well)

pmario:
@oflg ... Did you have a problem with the current default setting to "no"?

yaisog:
@pmario: I don't really remember. Might be that there was a bug associated with them. If so, the fix would have been done by @Jermolene.

oflg:
> @oflg ... Did you have a problem with the current default setting to "no"?

First of all, the user should have to enable the plugin's functionality after installing it, rather than finding the configuration tiddler to enable it manually again.
Secondly, I developed a [plugin ](https://github.com/oflg/fishing-catchfish/blob/main/tiddlers/configs/annotationfilter.tid)(containing an entry with a configuration tiddler of 'yes') to enable the annotation feature of Dynannotate, but couldn't enable it because of an override of the same shadow tiddler in Dynannotate.

oflg:
I tried to find the priority of overwriting the same shadow entries in different plugins in the tiddlywiki documentation, but failed.

pmario:
>First of all, the user should have to enable the plugin's functionality after installing it, rather than finding the configuration tiddler to enable it manually again.

I think that every user can expect that a plugin works with **sensible defaults** after installing it. It needs to work out of the box. The user action of manually "installing a plugin" shows enough "intention", that the plugin needs to be enabled too. 

Do you really want to install an app on your phone, just to search settings to activate it?

---------

I think **the right way to go** in your case is to create a TW [startup-action](https://tiddlywiki.com/#StartupActions) and check if the config-tiddler is a shadow tiddler. If it is you can create it with your own value. If it is a normal tiddler, the user may have changed it, or your startup may have created it. If it exists as a normal tiddler leaf it alone.

IMO changing core plugins is _not_ the right way to go for a personal usecase.

**But**

I think it may be an oversight, that the config-tiddlers are part of the plugin. Usually TW plugins work with defaults set in the wikitext code. 

We'll have to have a closer look at the code. May be those config tiddlers should really be removed, but for different reasons. You still should go the start-action way, because there may be several plugins that need the same config-tiddler.

oflg:
> I think **the right way to go** in your case is to create a TW [startup-action](https://tiddlywiki.com/#StartupActions) and check if the config-tiddler is a shadow tiddler. If it is you can create it with your own value. If it is a normal tiddler, the user may have changed it, or your startup may have created it. If it exists as a normal tiddler leaf it alone.

Thanks, good idea, I hadn't thought of that method before.

oflg:
> IMO changing core plugins is _not_ the right way to go for a personal usecase.

This isn't just for personal use, it's just through personal use that I've found a problem. The current problem is that Dynannotate's configuration items do not match its documentation. I can only change the config item or the documentation, and my PR changed the config item.

pmario:
>The current problem is that Dynannotate's configuration items do not match its documentation. I can only change the config item or the documentation, and my PR changed the config item.

That's right, but we need to check if the changes have been made intentionally, or if they slipped in by accident. I have a suspicion that the later is the case.

linonetwo:
> Thanks, good idea, I hadn't thought of that method before.

Then you probably never use Modern.TiddlyDev before https://github.com/tiddly-gittly/Modern.TiddlyDev/blob/master/wiki/tiddlers/%24__Modern.TiddlyDev_Startup.tid

Jermolene:
Hi @oflg is this still a problem? I'd rather change the docs to fit the code than the other way around.

oflg:
My problem is gone.

PR #7717, opened 2023-09-04
Update TranscludeWidget.tid to correct procedure name

kookma:
The copy pasted `mymacro` from old docs is changed to `myproc` in https://tiddlywiki.com/#TranscludeWidget


Jermolene:
Thank you @kookma much appreciated

PR #7721, opened 2023-09-05
Update WidgetMessage_ tm-permalink.tid

AnthonyMuscio:
Addition of note that the permalink message;

"The resulting link will be copied to the clipboard."

If anyone wants to make a format or grammatical change please do so.


AnthonyMuscio:
Preview is as intended.

Jermolene:
Thank you @AnthonyMuscio

PR #7722, opened 2023-09-06
Update GenesisWidget.tid

AnthonyMuscio:
Add "or HTML Element" to the description of "attributes not starting with $"


AnthonyMuscio:
The change was not propagated to the preview, will submit again

PR #7729, opened 2023-09-10
Simplify Permalink/Permaview URLs

### LINKED ISSUE #7728: [IDEA] Simplify Permalink/Permaview URLs
Current Permalink URLs are ugly and hard to read.  The permalink to the target tiddler,`Foo Bar Baz` would look like: `https://tiddlywiki.com/#Foo%20Bar%20Baz`.  The permaview for that target tiddler plus the tiddlers `Qux` and `Corge Grault` would look like `https://tiddlywiki.com/#Foo%20Bar%20Baz:%5B%5BFoo%20Bar%20Baz%5D%5D%20Qux%20%5B%5BCorge%20Grault%5D%5D`.

We can replace the `%20` encoding with simple plus signs (`+`)., making that first URL `https://tiddlywiki.com/#Foo+Bar+Baz`, and we can further simplify the list notation for the open tiddlers by joining them with commas rather than wrapping them in double square brackets, making the second one `https://tiddlywiki.com/#Foo+Bar+Baz:Foo+Bar+Baz&Qux&Corge+Grault`.  This should not break existing urls, so `https://tiddlywiki.com/#Foo%20Bar%20Baz` should continue to work as it does today.

There was a long-ish [talk.tiddlywiki.org thread][th] discussing this idea, and it included a number of alternative suggestions.  One simple possibility would be to use a different character.  Although `+` is common, Wikipedia, for one, uses the underscore (`_`) for a similar role.  I would welcome that equally well.  Other things I'm less enthused about are 

  * Suggesting the users who want prettier URLs simply replace the spaces in their titles with underscores.
  * Automating that behavior with a button that creates a new tiddler whose title copies the original, replacing spaces with underscores and whose body simply transcludes the original.
  * Doing as I suggest, but incorporating some signal that this url is using the new syntax, e.g. a leading `!`: `https://tiddlywiki.com/#!Foo+Bar+Baz`
  * Generating a slug by removing punctuation, lower-casing, and replacing spaces with `-`.  This is the most interesting alternative, but I as I responded on the thread:
    > Although that would would be both the most familiar and more readable than my + solution, I can’t see how to resolve the ambiguities, especially given TW’s dynamic nature.
    > 
    > Would a visitor arriving with #more-coffee be offered the choice between the three existing tiddlers `More Coffee?`, `More Coffee!`, and `MORE COFFEE!!!`? Or would we arbitrarily choose one? It only gets worse with permaviews.

**Additional context**
There were concerns raised in that discussion thread about the legitimacy of using the `+` character to encode spaces, but there is substantial prior art doing so, and a close reading of [the specification][sp] says not that a space *must* be encoded as `%20`, only that such encoding is available.  Tiddlywiki is already changing the common meaning of the fragment identifier for Web URLs from a unique identifier for a DOM element to a representation of a list of open tiddlers and an optional focused one.  This proposal doesn't at all change the meaning of those fragments for TW, only how they are encoded.

  [sp]: https://www.rfc-editor.org/rfc/rfc3986#appendix-A
  [th]: https://talk.tiddlywiki.org/t/7955

CrossEye:
Closes #7728

We create new functions for encoding and decoding the fragment (hash) portion of a URL, so that what would previously have been encoded as `https://tiddlywiki.com/#Foo%20Bar%20Baz` can now be encoded as `https://tiddlywiki.com/#Foo+Bar+Baz` and `https://tiddlywiki.com/#Foo%20Bar%20Baz:%5B%5BFoo%20Bar%20Baz%5D%5D%20Qux%20%5B%5BCorge%20Grault%5D%5D` can become `https://tiddlywiki.com/#Foo+Bar+Baz:Foo+Bar+Baz&Qux&Corge+Grault`.

This does not try to capture any additional punctuation.  If it appears in a tiddler title, it will still be percent-encoded.  There is a reasonably strong argument for keeping intact any punctuation the [specification][sp] allows, excepting the `+`, `:`, and `,` this format claims. That would be a fairly simple extension of this PR.

**Note** that this is intended to not break any existing permalinks/permaviews.  Those should continue to work as they have.  There is one minor unavoidable caveat there.  Although TiddlyWiki would not have generated a perma-link/view that included a `+` sign, it's possible that someone hand-crafted such a url for a tiddler with a `+` in its title.  That link would no longer work.

### Open Questions ###

  * Do we want to extend to allowing other characters?  All the following characters could be added without breaking the implementation or the spec: `-`, `.`, `_`, `~`, `!`, `$`, `'`, `(`, `)`, `*`, `;`, `=`, and`@`.  This would mean many fewer ugly percent-encoded characters?
  * Should we switch the separator for the list portion from the likely-to-appear in titles, `&` to the much-less-likely one of `~`?
  * What other formats of fragment identifier does TW support?  This only tested for the target and a title list.  I have vague memories of seeing other filters used as fragment identifiers in the URL.  Am I remembering correctly?  If so, how are those formatted.
  * How well does this match TW's coding standards?  I don't *think* I've added any linting errors, although that's hard to tell at the moment.  But my personal style uses many more modern JS features than the code-base seems to.  Some of those may have slipped in.  One specific question: does this project still try to support ES3?  Do I need to replace `Array.prototype.map` and `Array.prototype.filter` with other implementations?
  * Do you have other suggestions for a contributor new to the code-base?  
 

  [sp]: https://www.rfc-editor.org/rfc/rfc3986#appendix-A


boot/boot.js
```diff
@@ -287,6 +287,42 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+/*
+Convert a URI List Component encoded string (with `+` as an allowed replacement 
+for `+`) to a string
+*/
+$tw.utils.decodeTWURIList = function(s) {
+	return $tw.utils.decodeURIComponentSafe(
+		s.split('&')
+		.map(function(s) {return s.replaceAll('+', ' ')})
+		.map(function(s) {return s.includes(' ') ? '[[' + s + ']]' : s})
+		.join( ' ')
```
pmario, line 300: Typo - no leading space inside the braces `.join(" ")` -- TW core code uses double quotes for strings.  -- Double quotes also needed in the code above and below for consistency reasons.

boot/boot.js
```diff
@@ -287,6 +287,42 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+
+/*
+Convert a URI List Component encoded string (with `+` as an allowed replacement 
+for `+`) to a string
+*/
+$tw.utils.decodeTWURIList = function(s) {
+	return $tw.utils.decodeURIComponentSafe(
+		s.split('&')
+		.map(function(s) {return s.replaceAll('+', ' ')})
+		.map(function(s) {return s.includes(' ') ? '[[' + s + ']]' : s})
```
pmario, line 299: `array.prototype.includes()` is ES2016 and `replaceAll()` is ES2021 ... TW atm uses ES5 only
see: http://kangax.github.io/compat-table/es2016plus/

pmario:
>- Do we want to extend to allowing other characters? All the following characters could be added without breaking the implementation or the spec: -, ., _, ~, !, $, ', (, ), *, ;, =, and@. This would mean many fewer ugly percent-encoded characters?

I would be in favour of more readable characters in the links if it creates more human readable URLs

boot/boot.js
CrossEye: Fixed in 3249069e6a84610c29ac74a92fe780d5423e7077

CrossEye:
> > Do we want to extend to allowing other characters? All the following characters could be added without breaking the implementation or the spec: -, ., _, ~, !, $, ', (, ), *, ;, =, and@. This would mean many fewer ugly percent-encoded characters?

> I would be in favour of more readable characters in the links if it creates more human readable URLs

The latest commit adds this feature.  Now instead of 

> `https://tiddlywiki.com/#A%20(surprising%3F)%20tiddler%20with%20various%20*special%2Funusual*%20characters%2C%20%26%20more!`

We can have

> `http://tiddlywiki.com/#A_(surprising?)_tiddler_with_various_*special/unusual*_characters,_&_more!`

I have tried this with `_` instead of `+` as @AnthonyMuscio  recommended, and I agree that I like this much better.  I also switched the `,` separator to a `;`, as I think commas are more likely to actually appear in titles.

So for instance, we might have this permaview:

> `http://tiddlywiki.com/#Foo_Bar_Baz:Foo_Bar_Baz;Qux;Corge_Grault`

which is decent; it's hard to imagine better.  But for permalinks, this is hard to beat:

> `http://tiddlywiki.com/#Working_with_TiddlyWiki`


Code
----

The code is structured a little oddly for TW.  Part of this is to enable us to easily try various combination of space characters and separators.:

```
/*
The character that will substitute for a space in the URL
*/
var SPACE = "_";
/*
The character that will separate out the list elements in the URL
*/
var CONJUNCTION = ";";
/*
```

These -- along with a list of punctuation characters allowed in fragments -- are then used to build a lookup object, and some regexes that are used to remove unwanted percent encoding, and to add our own space encoding.

But this involves having a number of non-function variables inside the `_boot` closure, to be used by these functions.  I would expect that the indirection (which I'm using now to make these values easy to change) will go away, and there will be fewer of these _boot-global variables, but they won't all go away.  The regexes and lookup object are extracted, I guess, mostly for performance, but this does not feel at all like an early optimization.  However, if we do decide to do this, and there is a real objection to having those, I can inline them once we decide on the space and conjunction values.


Note
----

I really didn't stress that this is not a panacea.  There are plenty of URLs that will still have percent-encoding in them.  Probably the worst offender will be the colon (`:`).  To keep backward compatibility, that has to remain the separator between the target tiddler and the list of open tiddlers.  But it's a key part of the title of *all* system tiddlers, and I think it's a common character in titles.  But we will still have to deal with percent-encoding:

> `https://tiddlywiki.com/#JavaScript%3A_The_Good_Parts`
> `https://tiddlywiki.com/#$%3A/DefaultTiddlers`

I really wish we could get around that, but I think the break in backward compatibility is too severe.

boot/boot.js
```diff
@@ -287,6 +287,92 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+var claimed= [SPACE, ":", CONJUNCTION];
+/*
+Non-alphanumeric characters allowed in a URL fragment
+More information at https://www.rfc-editor.org/rfc/rfc3986#appendix-A
+*/
+var pchar = "-._~!$&'()*+,;=:@/?";
+/*
+The subset of the pchars we will not percent-encode in permalinks/permaviews
+*/
+var substitutes = pchar.split('').filter(c => claimed.indexOf(c) == -1);
```
pmario, line 314: Arrow functions are not ES5. 
We try to compare with tripple equals whenever possible `.indexOf(c) === -1`
Strings, empty strings or chars use double quotes `""` ... TW should **not have single quotes**, except for eg: `a + '"'` which is probably easier to read than `a + "\""`

boot/boot.js
```diff
@@ -287,6 +287,92 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+/*
+An object mapping the percent encodings back to their source characters
+*/
+var pctCharMap = substitutes.reduce(function (a, c) {
+	a['%' + c.charCodeAt(0).toString(16).toUpperCase()] = c
+	return a
+}, {});
+
+/*
+Convert a URI List Component encoded string (with `+` as an allowed replacement
```
pmario, line 337: IMO you should use `(with SPACE as an allowed ...` instead of `+` -- Check all comments

boot/boot.js
```diff
@@ -287,6 +287,92 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+		s.split(",")
+		.map(function(s) {return s.replace(spaceMatch, " ")})
+		.map(function(s) {return s.indexOf(" ") >= 0 ? "[[" + s + "]]" : s})
+		.join(" ")
+    )
+};
+
+/*
+Convert a URI Target Component encoded string (with `+` as an allowed replacement 
+for `+`) to a string
```
pmario, line 351: IMO `SPACE` instead of `+`

boot/boot.js
```diff
@@ -287,6 +287,92 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+var CONJUNCTION = ";";
+/*
+Those of the allowed url characters claimed by TW
+*/
+var claimed= [SPACE, ":", CONJUNCTION];
+/*
+Non-alphanumeric characters allowed in a URL fragment
+More information at https://www.rfc-editor.org/rfc/rfc3986#appendix-A
+*/
+var pchar = "-._~!$&'()*+,;=:@/?";
```
pmario, line 310: I think `CLAIMED` and `PCHAR` could be pseudo-constants too. We use "pseudo constants" quite a bit. So IMO it would be consistent with the rest of the core code

boot/boot.js
CrossEye: Damn, I had cleaned out all the ES6+ stuff, then introduced more.  I'll try to fix all this and the issues below this evening.

Thanks for looking.

boot/boot.js
CrossEye: Do you mean

> (with `+` as an allowed replacement for SPACE)

This is decoding `Foo+Bar` to `Foo Bar`.

boot/boot.js
CrossEye: Sure.  I don't know if PCHAR (regardless of capitalization) is too technical.  It's meaningful only in context of the spec.  I just didn't find a good alternative name.

pmario: You could use 

```
var VALID_IN_URL="-._~!$&'()*+,;=:@/?".split();
```

Since `claimed` is an array. May be VALID_IN_URL should be an array too. It would save you the `.split()` later in the code. 
But that's only nit-picking.

CrossEye: I did switch to a variant of this name in the latest commit, but won't bother making it an array, as this should be thrown away after the bikeshedding on characters to use is done.

CrossEye:
The latest commit (0cabd09) fixes more of the ES6+ stuff added in the previous commit.

It does one other thing.  In the [forum thread][th] @AnthonyMuscio had this comment:

> * The following additional characters also valid when encoding with encodeURI ; / ? : @ & = + $ , # the + symbol and others are in this subset. So if we modify your proposed change to “not encode these additional characters” the permalinks will look even easier to read.
>   * However I have read that not all systems may consider this a valid URI even if only used in the fragment. Thus I believe we need to allow people to opt out or in.

While I didn't get the full details of what he expects to be a problem, I did recognize that trailing sentence-ending punctuation is often not considered part of a link by many programs, assuming that that is part of the surrounding sentence.

Thus 

> `This is a link to https://tiddlywiki.com/#Foo_Bar_Baz!`

acts like it has a link to `https://tiddlywiki.com/#Foo_Bar_Baz` and a separate `!` to end the sentence.  I correct this by adding a configured trailing character if the title ends in `.`, `?`,  `!`, or that trailing character itself.  When decoding I strip off one copy of that trailing character from any URL.

I originally used the tilde ( `~`) as the trailing character, but soon realized that I could reuse underscore (`_`) and it's prettier.  So now the title `Foo Bar Baz!` is encoded as `#Foo_Bar_Baz!_`.

I haven't tested all the remaining characters yet to see if there are others besides `.`/`?`/`!` which need this treatment.  But before I bother, does it make more sense to simply do this for all non-alphanumeric characters?

-----------

There's one other major decision.  Tony's concern above is probably not only about these sentence-enders.  He would like this to be opt-in or opt-out, so that the current URL generation is still available.

I understand the concern, but I'm loathe to do that.  It certainly is possible, but I think it adds needless complexity.

What do others think?

   [th]: https://talk.tiddlywiki.org/t/7955/41

CrossEye:
> I originally used the tilde ( `~`) as the trailing character, but soon realized that I could reuse underscore (`_`) and it's prettier.  So now the title `Foo Bar Baz!` is encoded as `#Foo_Bar_Baz!_`.

And I just realized that this fixes an issue with the current permalink generation, which, while it converts `Foo?` to `#Foo%3F`, also converts `Foo!` to `#Foo!`, which will fail when used most places that autolink urls, because they will treat the `!` as outside the link.  The code above will turn these to `#Foo?_`, and `#Foo!_`, which while not as pretty as we might like, are not bad.

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+Code to make the permalink/permaview URLs more readable.
+(These comments are just to help reviewers; they will eventually be removed.)
+***************/
+
+
+/*************
+Helpers for encoding/decoding URIs.  **Note** that these are throw-away, only
+used to temporarily build `charMatch`, `spaceMatch`, `sentenceEnding`, and
+`sentenceTrailing`, all of which will eventually be hard-coded.
+***************/
```
pmario, line 300: I think these 2 comments can be removed now. We know, what the PR is used for and the comments cannot stay in the merged PR.

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+
+/*
+The character that will substitute for a space in the URL
+*/
+var SPACE = "_";
+
+/*
+The character added to the end to avoid ending with `.`, `?`, `!` or the like
+*/
+var TRAILER = '_';
```
pmario, line 310: Every character and string definition in this file should use double quotes for consistency. 
We only use single quotes if there is no other chance or to make it easier for translators. Eg: for translation strings to avoid string escaping. But that's basically it.

boot/boot.js
pmario: IMO the CONSTANTS should stay in the code. The js-engine will optimize the code. Hardcoded characters will be much harder to read and understand.

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+/*
+The subset of the pchars we will not percent-encode in permalinks/permaviews
+*/
+var substitutes = VALID_IN_URL_FRAGMENT.split('').filter(function(c){return CLAIMED.indexOf(c) === -1});
+
+
+/*************
+The more permanent implementation.  These helpers (`charMatch`, `spaceMatch`,
+`sentenceEnding`, `sentenceTrailing`) will be hard-coded, not derived.
+***************/
```
pmario, line 337: IMO this comment **above** can go away
The comment **below** should stay **as is**

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+/*************
+Helpers for encoding/decoding URIs.  **Note** that these are throw-away, only
+used to temporarily build `charMatch`, `spaceMatch`, `sentenceEnding`, and
+`sentenceTrailing`, all of which will eventually be hard-coded.
+***************/
+
+/*
+The character that will substitute for a space in the URL
+*/
+var SPACE = "_";
```
pmario, line 305: IMO the name should be `SUBST_SPACE`, `SEPARATOR` or something similar
SPACE feels wrong because it's a char other than space.

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+
+/*
+The character added to the end to avoid ending with `.`, `?`, `!` or the like
+*/
+var TRAILER = '_';
+
+/*
+The character that will separate out the list elements in the URL
+*/
+var CONJUNCTION = ";";
```
pmario, line 315: To save a bit of vertical space **for all 1-line** variable declarations below the var comments could be inline like so: 

```
// The character that will separate out the list elements in the URL
var CONJUNCTION = ";";

// Those of the allowed url characters claimed by TW
var CLAIMED = [SPACE, ":", CONJUNCTION];
```

boot/boot.js
```diff
@@ -287,6 +287,136 @@ $tw.utils.decodeURIComponentSafe = function(s) {
...
+		.filter(Boolean)
+		.map(function(s) {return s.trim()})
+		.map(function(s) {return encodeURIComponent(s)})
+		.map(function(s) {return s.replace(/\%20/g, SPACE)})
+		.map(function(s) {return s.replace(charMatch, function(_, c) {
+			return pctCharMap[c]})
+		})
+		.join(CONJUNCTION)
+		
+};
```
pmario, line 414: remove empty line **above** `}` -- and 
remove comment below

pmario:
@CrossEye -- at [Talk you wrote](https://talk.tiddlywiki.org/t/url-encoding-using-a-instead-of-20-for-a-space/7955/45?u=pmario): 

>It’s not clear, though, if there is any support. This discussion has been mostly you and me, with some technical details from **@pmario**, some interesting interjections from **@oeyoews**, and a somewhat mixed-signals message from **@jeremyruston**. The only interactions on the actual pull request have been a few well-appreciated corrections and suggestions from **@pmario**. I’m not sure whether to take this as discouragement or only a sign that things may move slowly around here.

I can assure you, that this contribution is valuable and appreciated by everyone here. -- 

As you wrote. Thinks move slowly in the core, especially if a change will be as visible for the end-users as this one is. Once it's merged and published, we have to live with it for a long time. 

Making things go away will have side effects for the community. As can be seen with v5.3.0, which only lived for several weeks and v5.3.1 was needed. 

So I hope this comment is **encouraging** 

------

We will also need to extend the documentation at: https://tiddlywiki.com/#PermaLinks

saqimtiaz:
I would also add that core development does tend to slow down for a while after a release, a natural consequence of the work and time commitment that each release takes. Also, I would recommend that for the future, directly creating an issue for a core improvement will move things along faster than posting on the community forums.

I support the proposed change and will try to find time in the near future to review the implementation. Thank you for your work on this.

pmario:
There is 1 thing, we did not talk about yet: Non-Latin alphabets like Greek, Russian and so on. See the screenshot.

If a tiddler is titled: заметка (note) the current URL is: 
https://tiddlywiki5-2vg2v0leq-jermolene.vercel.app/#%D0%B7%D0%B0%D0%BC%D0%B5%D1%82%D0%BA%D0%B0

But using the existing "slugify" mechanism it could be: 
https://tiddlywiki5-2vg2v0leq-jermolene.vercel.app/#zametka

image

Since I do not speak Russian or Greek I cannot say if this makes sense. But visually there is a big difference. 

Any idea about this one?

Jermolene:
Thank you @CrossEye, this is an ingenious solution to a long term blemish on TiddlyWiki's usability, and it is a great achievement to do so whilst retaining backwards compatibility.

One minor point is that I think that new definitions in boot.js are not actually used until after all module loading has been completed, and so they can be moved to `$:/core/modules/utils/utils.js`

> * Do we want to extend to allowing other characters?  All the following characters could be added without breaking the implementation or the spec: `-`, `.`, `_`, `~`, `!`, `$`, `'`, `(`, `)`, `*`, `;`, `=`, and`@`.  This would mean many fewer ugly percent-encoded characters?

I think that would make sense, yes.

> * Should we switch the separator for the list portion from the likely-to-appear in titles, `&` to the much-less-likely one of `~`?

That's a tricky one. Using `&` is familiar and clear. The tilde character can be hard to distinguish from a dash.

Perhaps the difficulties of making such a change backwards compatible make it impractical?

> * What other formats of fragment identifier does TW support?  This only tested for the target and a title list.  I have vague memories of seeing other filters used as fragment identifiers in the URL.  Am I remembering correctly?  If so, how are those formatted.

The format for permalinks is the encoded title of the target tiddler followed by an optional colon and then a filter giving the tiddlers to display in the story (ie a permaview). The two parts can be given independently:

https://tiddlywiki.com/#:[search[jeremy]]

> * How well does this match TW's coding standards?  I don't _think_ I've added any linting errors, although that's hard to tell at the moment.  But my personal style uses many more modern JS features than the code-base seems to.  Some of those may have slipped in.  One specific question: does this project still try to support ES3?  Do I need to replace `Array.prototype.map` and `Array.prototype.filter` with other implementations?

We don't use map/filter/reduce, sadly. There has been some recent discussion about updating the JS dialect we use, but it has not been resolved yet.

> * Do you have other suggestions for a contributor new to the code-base?

This is a great PR, the top post is nice and clear, and the code has the hallmarks of being written for ease of comprehension, good stuff.

boot/boot.js
CrossEye: Sorry about that.

I swear this things proliferate on their own!  I'm quite sure that my decades of a different coding style of nothing to do with it.  Nothing at all.  😜   Last I tried, there were so many lint warnings that I wasn't able to identify ones I was adding.

boot/boot.js
CrossEye: Ok.  The multiline version just looked like house style... but I didn't look very hard.

CrossEye:
Damn, I wrote a long response to everyone, in great detail, and think I must have left it in Preview mode when my machine needed to reboot.  I will create it again soon, but I don't have the heart right now.

In short, though, thanks for the encouragement!  I will make (most of) the requested changes for my next commit.

CrossEye:
@pmario:

> I can assure you, that this contribution is valuable and appreciated by everyone here. --
> 
> So I hope this comment is **encouraging**

It is, very much so.  Thank you very much!

> We will also need to extend the documentation at: https://tiddlywiki.com/#PermaLinks

Oops.  I remembered to do this in my previous PR, but that was a tiny bug fix.  Can't believe I forgot it here.  I will include this in my next commit.  (Tomorrow, I'm hoping.)

CrossEye:
@saqimtiaz:

> I would also add that core development does tend to slow down for a while after a release, a natural consequence of the work and time commitment that each release takes.

I do understand.  I created and help maintain a large project with varied periods of frenetic activity and others of relative quiescence. 

> Also, I would recommend that for the future, directly creating an issue for a core improvement will move things along faster than posting on the community forums.

Ok, I will probably get to that point eventually.  But I don't have enough experience with the core to get a good sense of whether something is useful or has been considered and rejected or is extremely controversial.  I'll get there, though.

> I support the proposed change and will try to find time in the near future to review the implementation. Thank you for your work on this.

Thank you.

CrossEye:
@pmario: 

> There is 1 thing, we did not talk about yet: Non-Latin alphabets like Greek, Russian and so on. See the screenshot.
> 
> If a tiddler is titled: заметка (note) the current URL is:
> https://tiddlywiki5-2vg2v0leq-jermolene.vercel.app/#%D0%B7%D0%B0%D0%BC%D0%B5%D1%82%D0%BA%D0%B0
> 
> But using the existing "slugify" mechanism it could be:
> https://tiddlywiki5-2vg2v0leq-jermolene.vercel.app/#zametka

I think there's a real problem that this is not reversible.  If we had a tiddler named `заметка` and one named `zametka`, which one would we choose to show?  It's the same problem I mentioned on [talk][ta] which asked which of the following three existing tiddlers `#more-coffee` should like to: `More Coffee?`, "More Coffee!`, or `MORE COFFEE!!!`?

But this is a serious problem you raise.  My proposal will make for cleaner URLs, but only for those using the Latin alphabet.  That's related to the specifications for URL fragment strings.  It might be interesting to see if we can find a way to do this more readably, even if the generated URLs are not legally allowed, if they would work in any reasonable place that people use URLs.  I've been mostly focused on the space and other punctuation, but perhaps if we do just generate 

https://tiddlywiki.com/#заметка

and so long as the characters are letters, it would just work.  I will have to think about this more.

  [ta]: https://talk.tiddlywiki.org/t/7955/34

CrossEye:
> Thank you @CrossEye, this is an ingenious solution to a long term blemish on TiddlyWiki's usability, and it is a great achievement to do so whilst retaining backwards compatibility.

😊 Thank you.  I should reiterate that there is one backward incompatible situation: if the user has laying around a permalink/view that was handcrafted and contains an `_`, then we will fail to decode it properly.  This shouldn't affect permalinks actually generated by TW.  I think this is likely to be an extremely rare case, but we should note it.

> One minor point is that I think that new definitions in boot.js are not actually used until after all module loading has been completed, and so they can be moved to $:/core/modules/utils/utils.js

While I have no problem moving them, the existing code that I'm modifying/replacing is in `boot.js`.  I don't know the lifecycle well enough.  Clearly the encoding part will happen late enough, but what about the decoding?  Have we already loaded everything everything at the time we try to parse the incoming fragment?

>> Should we switch the separator for the list portion from the likely-to-appear in titles, `&` to the much-less-likely one of `~`?
>
> That's a tricky one. Using & is familiar and clear. The tilde character can be hard to distinguish from a dash.

The latest incarnation uses `_` for the space (also prepended to trailing sentence-ending punctuation) and uses a semicolon between list items.  I think the semicolon is much less likely in titles than the comma or ampersand, although I have no evidence to back that up.  The biggest disappointment is the colon.  That is much more likely to appear in titles, but we can't do anything about it without badly breaking backwards compatibility.

> The format for permalinks is the encoded title of the target tiddler followed by an optional colon and then a filter giving the tiddlers to display in the story (ie a permaview). The two parts can be given independently:
> 
> [https://tiddlywiki.com/#:[search[jeremy]]](https://tiddlywiki.com/#:%5Bsearch%5Bjeremy%5D%5D)

I will have to do more testing with these, but am not expecting any problems.

> We don't use map/filter/reduce, sadly. There has been some recent discussion about updating the JS dialect we use, but it has not been resolved yet.

Understood.  The [Ramda][ra] library I founded and still sporadically maintain supports ES3!

I guess I have some refactoring to do.  Are there TW functions I can use in place of these?  Could I write my own plain functional versions of them?  Or should I just resort to `for`/`while` loops?

> This is a great PR, the top post is nice and clear, and the code has the hallmarks of being written for ease of comprehension, good stuff.

Thank you.  I'm happy to hear it.



  [ra]: https://ramdajs.com

CrossEye:
## A different idea altogether ##

I will try to make another commit tonight to address some issues, but I just had a thought about an entirely different permalink/view approach and wanted to see what people thought.

I want to make it clear up front that I still prefer the version I've been coding.  But I think this alternative is compelling enought to deserve some discussion.

## Use a short SHA-1 hash ##

We can make simple URLs that are not transparent but are short, easy to type, and would work everywhere by using one-way hashes of the titles, and then shorten them to a useful length.  For instance, `The First Rule of Using TiddlyWiki` would convert with the SHA-1 algorithm to `d8d158cb0dc0b0ab8baf69a29cf2b33c328abfa9`, which we could shorten to `d8d158cb0d`.  (In a wiki with 5000 tiddler titles, the chance of there being any collisions at all is approximately one in one hundred thousand.  If that scared us, we could use more characters.)  With this, we might have a permalink like `https://tiddlywiki.com/#/d8d158cb0d`   If we wanted one a little easier to type, we could break it up with a punctuation character: `https://tiddlywiki.com/#/d8d-158-cb0d`

On loading with that URL, TW would scan tiddlers for one with a title whose SHA-1 title hash starts with that value.  (If we use this elsewhere besides on load we should pre-cache the values.)

This could be done in an entirely backwards-compatible manner if we signal this style url fragment with an initial character we wouldn't otherwise create in a URL.  Here `/` makes the most sense, as it's widely seen around the web for navigating within single-page apps.  (Ok, the same caveat applies to pre-existing hand-crafted URL fragments starting with `/`, but that's still ignorable.)

## Possible extension ##

A possible extension of this idea would be to store this slug as a field in the tiddler, to be updated when the title changes.  (I don't know just how fraught that might be, though.)  But this could also let users override that slug for custom needs.  (`https://tiddlywiki.com/#/first-rule`)  There are several complexities here, and I don't know if it would be worth it.


## Downsides ##

There are two significant downsides I see:

  * There is a noted lack of transparency: You're going *somewhere*, but who knows where?  If this slug is not found, you do not have the nice signal such as 
  
    > *Missing tiddler "The First Rule of Using TiddlyWik" – click  to create* 

    which might make it clear that you dropped a character when copying the URL.  And this is simply nicer to read:  `https://tiddlywiki.com/#The_First_Rule_of_Using_TiddlyWiki` than is any variant of ``https://tiddlywiki.com/#/d8d158cb0d``.

  * Although this could convert permalinks the way my current one does, with a hash for each title, there is no clear extension mechanism for an arbitrary filter.  It probably could not extend to covering something equivalent to `https://tiddlywiki.com/#:[tag[Reference]]` or, these days, `https://tiddlywiki.com/#:%5Btag%5BReference%5D%5D`


## Question ##

Is this worth pursuing?  I want to finish up the approach I'm working on.  But is there something more compelling about this alternative that makes it worth following up in parallel?

CrossEye:
I finally found time to work on this again.  I've addressed almost all the points above.  My biggest question is about this:

>> @Jermolene :
>> One minor point is that I think that new definitions in boot.js are not actually used until after all module loading has been completed, and so they can be moved to $:/core/modules/utils/utils.js
>
> @CrossEye:
> While I have no problem moving them, the existing code that I'm modifying/replacing is in `boot.js`. I don't know the lifecycle well enough. Clearly the encoding part will happen late enough, but what about the decoding? Have we already loaded everything everything at the time we try to parse the incoming fragment?

I can certainly move these, but not understanding the lifecycle well enough, I want to ensure that it won't break anything.

Also, I wrote my own `map` and `filter` functions and used them in the places I would otherwise have used `Array.prototype.map` and `Array.prototype.filter`.  I placed them in `boot.js`, in the namespace `$tw.utl`.  Is this appropriate?

CrossEye:
I removed the `WIP` prefix.  Subject to the questions in my last post, I think this is ready for consideration.

I still need to know if I need to move the bulk of the changes from `boot` to `util`; mostly here I want reassurances that this will not break anything, as the code I'm replacing lived in `boot`.  And I want to know if adding my `map` and `filter` functions is appropriate.

Oh, and I guess someone really should double-check the docs, which I don't think have had a review.

editions/tw5.com/tiddlers/concepts/PermaLinks.tid
```diff
@@ -40,6 +40,22 @@ There are technical restrictions on the legal characters in an URL fragment. To
 
 Both the target tiddler title and the story filter should be URL encoded (but not the separating colon). TiddlyWiki generates properly encoded URLs which can look quite ugly. However, in practice browsers will usually perfectly happily process arbitrary characters in URL fragments. Thus when creating permalinks manually you can choose to ignore URL encoding.
 
+!! Simpler URLS
+
+<<.from-version "5.2.3">> The URLs generated are simplified from the hard-to-read percent encoding when feasible.  Spaces are replaced with underscores (`_`), many punctuation characters are allowed to remain unencoded, and permaview filters receive a simpler encoding.  For example the tiddler "Hard Linebreaks with CSS - Example", which percent-encoded would look like
```
Jermolene, line 45: The version number should be 5.3.2

Jermolene (commented):
Thanks @CrossEye my usual policy with boot.js is to only include code that is required during the boot process, before JS modules have been loaded. Would it be possible in this case for these utilities to reside in utils.js?

editions/tw5.com/tiddlers/concepts/PermaLinks.tid
CrossEye: Oops!  Fixed in latest commit.

CrossEye:
@Jermolene:
 
> Would it be possible in this case for these utilities to reside in utils.js?

I was hoping *you* would be able to answer that! 😄   I wasn't quite sure of the timings.  But I just tried, and it was no issue.

The code introduces a number of closure-local variables.  That doesn't bother me in general, but I see no similar examples in `util.js`, and I'm wondering if I'm missing any important conventions here.  Also, I publicly exposed the `map` and `filter` functions I wrote.  I want to make sure that doesn't feel like clutter in the `$tw.util` namespace.  And is there any objection to the way they're written, as fully curried, data-last functions?  (That is, Ramda style rather than Underscore style, called like `map (square) ([1, 2, 3, 4])` rather than `map([1, 2, 3, 4], square)`.) 

I think this is reasonably complete, although I wouldn't be surprised if @pmario's fine-tooth comb catches something out-of-step with house style.  It does not try to do anything about Mario's musings on non-Latin alphabets.  I see no reasonable way to handle that here.

Is there clean-up necessary?  Should I merge my changes into a single commit or anything like that?

saqimtiaz:
> And is there any objection to the way they're written, as fully curried, data-last functions? (That is, Ramda style rather than Underscore style, called like `map (square) ([1, 2, 3, 4])` rather than `map([1, 2, 3, 4], square)`.)

I think they should be in the `map([1, 2, 3, 4], square)` style for consistency, see for example `$tw.utils.each`:
https://github.com/Jermolene/TiddlyWiki5/blob/master/boot/boot.js#L141-L162

@Jermolene I note that we already use `Array.reduce` in the core and `Array.map` in several official plugins.

CrossEye:
@saqimtiaz:

> I think they should be in the `map([1, 2, 3, 4], square)` style for consistency, see for example `$tw.utils.each`:
https://github.com/Jermolene/TiddlyWiki5/blob/master/boot/boot.js#L141-L162

Should I also do the extension that this version of `each` does, to handle Objects as well as Arrays?  For `map`, should I also include functions?  (The type `Function` is a [Functor][fu], and `map(g, f)` should be roughly equivalent to `(x) => f(g(x))`.)

So that is, besides the obvious

```js
map([1, 2, 3], square) //=> [1, 4, 6]
filter([8, 6, 7, 5, 3, 0, 9], n => n > 5) //=> [8, 6, 7, 9]
```

Should we also support

```js
map({a: 1, b: 2, c: 3}, square) //=> {a: 1, b: 4, c: 9}
filter({a: 8, b: 6, c: 7, d: 5, e: 3, f: 0, g: 9}) //=> {a: 8, b: 6, c: 7, g: 9}
```

and possibly

```js
map(square, add3)(5) //=> 28
map(add3, square)(5) //=> 64
```

> @Jermolene I note that we already use `Array.reduce` in the core and `Array.map` in several official plugins.

If that's the case, especially for the core, should we fix those sections with custom functions as I do here, or is it time -- in this day of evergreen browsers -- to say that more modern JS is acceptable, and document the newest version TW supports?

  [fu]: https://github.com/fantasyland/fantasy-land/#functor

saqimtiaz:
@CrossEye I suspect a change to more modern JS is imminent but we do not as of yet have a clear roadmap for it. Keeping this in mind, I would actually recommend that we not add utility functions for map and filter at this time and re-factor the code in this PR to use $tw.utils.each to iterate over an array to create a new mapped/filtered array as needed.

CrossEye:
@saqimtiaz:

The latest commit  switches to `each`.

But man, that code is starting to look damn ugly to me!  I hope you're right about

>  I suspect a change to more modern JS is imminent 

!!

core/modules/utils/utils.js
```diff
@@ -1005,4 +1005,100 @@ exports.makeCompareFunction = function(type,options) {
...
+// Those of the allowed url characters claimed by TW
+var CLAIMED = [SPACE_SUBSTITUTE, ":", CONJUNCTION];
+
+// Non-alphanumeric characters allowed in a URL fragment
+// More information at https://www.rfc-editor.org/rfc/rfc3986#appendix-A
+var VALID_IN_URL_FRAGMENT = "-._~!$&'()*+,;=:@/?".split("");
+
+// The subset of the pchars we will not percent-encode in permalinks/permaviews
+var SUBSTITUTES = []
+$tw.utils.each (VALID_IN_URL_FRAGMENT, function(c) {if(CLAIMED.indexOf(c) === -1) {SUBSTITUTES.push(c)}});
```
pmario, line 1026: This should be multilines, for better readability and there should be no space between `$tw.utils.each(..` and the open brace.

core/modules/utils/utils.js
```diff
@@ -1005,4 +1005,100 @@ exports.makeCompareFunction = function(type,options) {
...
+}, {});
+
+// Convert a URI List Component encoded string (with the `SPACE_SUBSTITUTE`
+// value as an allowed replacement for the space character) to a string
+exports.decodeTWURIList = function(s) {
+	var parts = s.replace(SENTENCE_TRAILING, "$1").split(CONJUNCTION);
+	var withSpaces = []
+	$tw.utils.each(parts, function(s) {withSpaces.push(s.replace(SPACE_MATCH, " "))});
+	var withBrackets  = []
+	$tw.utils.each(withSpaces, function(s) {withBrackets .push(s.indexOf(" ") >= 0 ? "[[" + s + "]]" : s)});
```
pmario, line 1060: please make the code readable

CrossEye: Can you explain what that would entail?  I feel like I've been slowly making the code less-and-less readable to meet the criteria here.  Should I undo that, or move to a still more imperative style?

To me, this version is *much* more readable:

```js
	const withSpaces = parts.map(s => s.replace(SPACE_MATCH, " "))
	const withBrackets = withSpaces.map(s => s.indexOf(" ") > = 0 ? "[[" + s + "]]" : s)
```

and this a step further toward readablity:

```js
	const withBrackets = parts
		.map(s => s.replace(SPACE_MATCH, " "))
		.map(s => s.indexOf(" ") > = 0 ? "[[" + s + "]]" : s)
```

But I'm pretty sure that's not what you want.

pmario: If you globally search for `$tw.utils.each(` in the TW source you can see how it's ususally used. For new code we try to stay consistent. 

```
$tw.utils.each(withSpaces,function(s){
	withBrackets.push(s.indexOf(" ") >= 0 ? "[[" + s + "]]" : s);
});
```

CrossEye: Updated to match this.  I'd argue the point about readability, but this is not the time or place!  😉

CrossEye:
Updated to address @pmario's suggestions.

I have a minor question.  How do people test their changes?  I've been doing

```sh
node tiddlywiki ./editions/tw5.com --listen
```

And it works fine for these use-cases, but I get the following warning


>Warning: Plugin(s) required for client-server operation are missing.
>"$:/plugins/tiddlywiki/filesystem", "$:/plugins/tiddlywiki/tiddlyweb"


I'm guessing that if I were testing anything to do with saving, this would turn into an issue.  Is there a quick fix?

pmario:
I use the server version

```
node tiddlywiki ./editions/tw5.com-server --listen
```

saqimtiaz:
@CrossEye I recommend we move the permalink handling into a separate file in the utils directory, which will provide that code with its own closure. @Jermolene can probably recommend an appropriate filename.

Jermolene:
> @CrossEye I recommend we move the permalink handling into a separate file in the utils directory, which will provide that code with its own closure. @Jermolene can probably recommend an appropriate filename.

Thanks @CrossEye @saqimtiaz – perhaps `$:/core/modules/utils/twuri-encoding.js`. Besides encapsulating the global constants, another benefit is that give we're introducing a new concept I think having it named in the filename is useful.

CrossEye:
@saqimtiaz, @Jermolene, @pmario:

>> @CrossEye I recommend we move the permalink handling into a separate file in the utils directory, which will provide that code with its own closure. @Jermolene can probably recommend an appropriate filename.
>
> Thanks @CrossEye @saqimtiaz – perhaps $:/core/modules/utils/twuri-encoding.js. Besides encapsulating the global constants, another benefit is that give we're introducing a new concept I think having it named in the filename is useful.

This change is in.

However, I noticed something odd, and am wondering if there's a major flaw.

I'd been doing all my testing on `.editions/TW5.com`.  Everything is fine.  Above I asked how people avoid an error in the console when starting this way, and @pmario suggested using `.editions/TW5.com-server`.

If I do this, when I start and load http://localhost:8080/, it converts the URL to http://localhost:8080/#HelloThere:HelloThere;GettingStarted;Community.  This correctly identifies the list of open tiddlers and the current tiddler.  But it shouldn't show up in the url.

I will try to dig into this soon, but was wondering if someone had a quick explanation for this difference between `TW5.com` and `TW5.com-server`.

Jermolene:
> This change is in.

Thanks @CrossEye

> However, I noticed something odd, and am wondering if there's a major flaw.

The problem is likely to be that the server edition has the setting "Navigation Address Bar" set to "Include the target tiddler and the current story sequence". See here:

https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/tw5.com-server/tiddlers/config-Navigation-UpdateAddressBar.tid

CrossEye:
Damn, I didn't see this response before digging into it myself and getting to the same place.

Seeing the comment `<!-- TODO: Remove before end of beta -->`, I added a commit that changed that from `permaview` to `no`. (and removed the comment.)  But I'm not sure if that's the correct solution.  Is this permaview-address-bar-on-load behavior expected from this edition?  Is there a different fix?  Or should I just ignore this problem and assume that if it's working well in the TW5.com edition, it's doing what's needed.  

The permalink fragment that it is navigating to is correct according to the changes here, but it feels very odd to me to autonavigate on load.

Jermolene:
> Damn, I didn't see this response before digging into it myself and getting to the same place.

Ouch! Sorry to hear that.

> Seeing the comment `<!-- TODO: Remove before end of beta -->`, I added a commit that changed that from `permaview` to `no`. (and removed the comment.) But I'm not sure if that's the correct solution.

It is not the correct solution. The new code needs to work with any combination of the existing configuration settings.

> Is this permaview-address-bar-on-load behavior expected from this edition? Is there a different fix?

The correct fix would be to make the new encoding work with all the relevant configuration options.

> Or should I just ignore this problem and assume that if it's working well in the TW5.com edition, it's doing what's needed.

No, it also needs to work in tw5.com-server.

> The permalink fragment that it is navigating to is correct according to the changes here, but it feels very odd to me to autonavigate on load.

It is a setting that some people prefer in some situations. It means that the browser address bar maintains a representation of the current story river, useful for bookmarking and sharing.

pmario:
>It is a setting that some people prefer in some situations. It means that the browser address bar maintains a representation of the current story river, useful for bookmarking and sharing.

It also allows us to use the browser backwards and forwards button.

pmario:
>The correct fix would be to make the new encoding work with all the relevant configuration options.

@Jermolene -- As far as I can see, it works as expected with this PR.

Jermolene:
> > The correct fix would be to make the new encoding work with all the relevant configuration options.
> 
> @Jermolene -- As far as I can see, it works as expected with this PR.

Apologies I had misunderstood, I thought that @CrossEye had discovered that the changes here weren't compatible with the setting.

CrossEye:
>> Damn, I didn't see this response before digging into it myself and getting to the same place.
> 
>Ouch! Sorry to hear that.

Oh no, not at all.  It's one way to learn more.

It does work in that configuration; I simply had never used that setting, and as it had to do with setting the URL, I thought I must have broken something.

Tonight, I'll back out that latest change (blocked from doing so at GigantiCorp) , and I think this is done.

I am curious, though, about the comment "`<!-- TODO: Remove before end of beta -->`".  Is that old, or does it have to do with the `5.3.2` release?

Jermolene:
Great thank you @CrossEye, and apologies for the confusion.

CrossEye:
@Jermolene 
> Great thank you @CrossEye, and apologies for the confusion.

It's my own misunderstanding that caused any issue at all.  But we all have to learn somehow.

@pmario:
> It also allows us to use the browser backwards and forwards button.

Put that way, I may investigate using it myself.  I've been using TW for years, and very heavily in the past year, but I still too often try the back button inappropriately.  Thanks!

CrossEye:
That commit is reverted.  AFAICT, this is complete.

I've only committed bug fixes before.  For something like this that should have release notes, who writes them?  IOW, should I write release notes for this?

Jermolene:
Thank you again for all your hard work on this @CrossEye

CrossEye:
@Jermolene:
> Thank you again for all your hard work on this @CrossEye

It was very much my pleasure.  I learned a lot about TW in doing it!

AnthonyMuscio:
I am afraid I could not follow this whole process and change, but I just took the preview to test it, and see the safe `_` was used. 

I just created a tiddler with the underscore in the name and it creates a separate tiddler, but a permalink will only open the one with spaces. And fail if that is deleted, even if the underscore version is available.
* At a minimum this needs to be well documented.
* My concern is this is technically not backward compatible.
  * if someone made extensive use of underscores in the titles for this very purpose, of readable URL's in the past it could be a disaster for their wiki. Although they could batch change them as long as they do not also have space separated titles.
  * At the very least we would should provide a way for them to defeat or turn off this new behaviour, even if it is hacky.

I have not explored the use of other characters at this point.

never the less thanks @CrossEye for your effort.

CrossEye:
@AnthonyMuscio:

> I just created a tiddler with the underscore in the name and it creates a separate tiddler, but a permalink will only open the one with spaces. And fail if that is deleted, even if the underscore version is available.

Yeah, this is a flaw I should have seen.  I will try a trick to alleviate it.  Probably tomorrow, but maybe not until Sunday.

We knew that this was not entirely backward compatible, but this is a bigger gap than expected.  I think I can get this close enough not to matter as a practical concern.

> At the very least we would should provide a way for them to defeat or turn off this new behaviour, even if it is hacky.

I'm quite loathe to do this, unless absolutely necessary.  Technically it would not be very difficult, but maintaining both behaviors feels like bloat.

pmario:
I think it should be easy to test, if a tiddler with spaces exists and then test if tiddlers with underscores exist. The only problem is, that every combination has to be tested. Starting with one which has underscores instead of spaces, which is the most likely one

Jermolene:
Hi @CrossEye. I'm afraid I'm going to revert this PR for the moment due to the backwards compatibility issues. We'll return to it for v5.3.3

Jermolene:
Reverted in 0716ed4c30134c3b3a8a4840b3894047d675e1e4

I should add that even if this were to be fixed now, I think it would be unwise to rush it into this release. I would also prefer to have tests for the encoding/decoding functions.

CrossEye:
@pmario:
> I think it should be easy to test, if a tiddler with spaces exists and then test if tiddlers with underscores exist. The only problem is, that every combination has to be tested. Starting with one which has underscores instead of spaces, which is the most likely one

I'd really rather avoid that.  I believe the URL conversion should not depend on the actual tiddler names in the wiki.  I'm thinking that I can double-up underscores to represent single underscores.  There is still a backward incompatibility, but it is a much less likely scenario: if you have spaces next to underscores in titles, then there is an ambiguity between `<underscore><space>` and `<space><underscore>`, which both would map to a triple underscore.  And any other combination of the two would have similar ambiguities.  But I think finding wikis where people are using combinations of these two is a vanishingly small set.

If we don't want to accept this incompatibility, then I would prefer to go with Tony's suggestion of a toggle for the behavior, so that you can revert to percent-encoding.  I don't like the idea of continuing to support both, but I'm far too new to the dev community to argue too hard against it.  And I would definitely prefer it to testing for the existence of any of the 2<sup>n</sup> potential matching tiddlers.

@Jermolene:
> Hi @CrossEye. I'm afraid I'm going to revert this PR for the moment due to the backwards compatibility issues. We'll return to it for v5.3.3

Of course.  I'm disappointed, but it's definitely the right call.

> I would also prefer to have tests for the encoding/decoding functions.

Yes, I did forget to ask about that.  I didn't see tests anywhere near the root of repo, so thought there was probably not much of a test suite.  Now I see there's a dedicated edition -- a fascinating concept!  When I revisit this, I'll convert my personal mocha tests to match the style.

AnthonyMuscio:
Would it not be adequate to simply test for a tiddler with underscores and open it, otherwise open the equivalent with spaces as it it were underscores?

Ie if someone has gone out of there way to use undersores in the tiddler title it will open them. If not they will have a link generated from new permalink (now containing underscores) to open the tiddler by the same name with spaces.

To me this is adequate if documented, it even allows underscore tiddlers to be generated specifically to intercept a link incoming with underscores if desired. As I suggested earlier this will allow you to maintain a destination url even when the tiddler title needs to change within the wiki. 
* Trust me this is of great value allowing a form of redirect for SOE and link maintenance.

CrossEye:
@AnthonyMuscio:
> Would it not be adequate to simply test for a tiddler with underscores and open it, otherwise open the equivalent with spaces as if it were underscores?

There's no perfect solution.  I would live with this if we had to, but as I replied to @pmario, I believe the URL conversion should not depend on the actual tiddler names in the wiki.

It also would have a problem with a tiddler title like `"Team decided to WONT_FIX the issue"`, as neither `"Team decided to WONT FIX the issue"` nor `"Team_decided_to_WON'T_FIX_the_issue"` actually exists.  But my main objection is simply philosophical.  The job of this code is to encode/decode between URLs and titles/title lists.  It should do the job the same way regardless of the actual tiddler titles in the wiki.

My current thinking, which seems quite possible, and I will try it soon, is to convert this to

```
Team_decided_to_WONT__FIX_the_issue
                     ^--- double underscore
``` 

and then possibly add the flag you were suggesting earlier, the one I didn't and still don't really like, but which would let people exit this new behavior if they have a need to mix underscores and spaces together or have titles with consecutive spaces.

> To me this is adequate if documented, it even allows underscore tiddlers to be generated specifically to intercept a link incoming with underscores if desired. As I suggested earlier this will allow you to maintain a destination url even when the tiddler title needs to change within the wiki.

This is a fascinating concept.  I can definitely see the uses for it.  But I think if we want this, we should have an explicit mechanism for it, and not depend upon a convention about spaces and underscores.

To me that mechanism sounds simple both to write and to configure.  It would probably run before the decode above and might use dictionary tiddlers something like this:

```
title: SEO Animals
tags: $:/tags/TitleMapping
type: application/x-tiddler-dictionary

Horses: Mammals - Equine
Dolpins: Mammals - Delphinidae
Dogs: Mammals - Canine 
```

or the equivalent in JSON.  (Please excuse any lapses in biological taxonomy!)

pmario:
>..and then possibly add the flag you were suggesting earlier, the one I didn't and still don't really like, but which would let people exit this new behavior if they have a need to mix underscores and spaces together or have titles with consecutive spaces.

There should be **no option to deactivate** the new behaviour. IMO it only creates confusion.

CrossEye:
@pmario:
> There should be **no option to deactivate** the new behaviour. IMO it only creates confusion.

That would be my preference too.  I would only be willing to do this if it's the only way to make this change widely acceptable.  Otherwise we should treat it as a break with past behavior. 

Here's there's significant backward compatibility because -- except in some narrow cases we've been discussing -- older permalinks will still keep working.  New ones would look different, true, but that doesn't cause issue with the existing ones.

pmario:
>Here's there's significant backward compatibility because -- except in some narrow cases we've been discussing -- older permalinks will still keep working. 

There would an other way to handle it. As soon as there is an underscore in the tiddler title use the "old behaviour" for this tiddler. 
It would create an ugly, but compatible link. 

So users with underscores in their titles can change them if they want or stay with the old behaviour. -> Done.

AnthonyMuscio:
This seems far simpler to me than the responses raise. If a url is given, then open the tiddler to which it refers whether it has underscores or not. Now if such a tiddler is not found as per this activity, look to open the same tiddler where the underscores are replaced with spaces.

A wiki already using underscored tiddlers (for whatever reason will still work), urls that "permalinks containing spaces, to one containing underscores" will; also open the "required tiddler.
* The only side effect is if you have both an underscored tiddler and a underscore permalink is the underscored tiddler wins.

This side effect is unlikely to cause problems because it relates to legacy wikis, only "if you upgrade to a wiki containing these new permalinks is it even a question?" and the preference is for backward compatibility.

However this side effect is a feature and can be called out as such a feature and documented, thereby not becoming a problem but a feature.
* Any url permalink, permaview generated and published is stored elsewhere is "outside the current wikis control", however it is within the wikis ability to maintain the effectiveness of such links, so no published link ends up being a dead link. If an existing underscored tiddler exists, or a permalink using this mechanism is used we want the underscored url to always be valid. 
* However if you use the new mechanism and publish an underscored link to the world to a space delimited tiddler, highly likely, there is a chance you will want to rename the tiddler one day, relink may even do it for you. And your published link is now a dead one, because it can no longer find the space delimited version anymore.
* However in such cases it is trivial to maintain a tiddler with the original title, where spaces are replaced with underscores and will be valid for eternity, because any incoming links will find this tiddler. This tiddler can redirect the user, present a note, or simply transclude the now renamed tiddler, or even give a deprecated link message.

My preference here is driven by a reasonably deep understanding of search Engin optimisation, maintaining valid links, website publishing and more so please take my appeal seriously, especially since this is a core change, that sets a standard, we need to maintain indefinatly, to support backward compatibility.

CrossEye:
@pmario:
> IMO there is an other way to handle it. As soon as there is an underscore in the tiddler title use the "old behaviour" for this tiddler.
> It would create an ugly, but compatible link.

I wish we could.  The reason that we're having  this discussion, though, is that the "old behavior" does not do anything to the underscore characters.  They were not percent-encoded.  So we would have no way to distinguish between `Foo Bar Baz` and `Foo_Bar_Baz`.

CrossEye:
@AnthonyMuscio:
> This seems far simpler to me than the responses raise. If a url is given, then open the tiddler to which it refers whether it has underscores or not. Now if such a tiddler is not found as per this activity, look to open the same tiddler where the underscores are replaced with spaces.

That reduces the problem, but doesn't eliminate it.  Again, if we were to try to load the url [https://tiddlywiki.com/#Team_decided_to_WONT_FIX_the_issue](https://tiddlywiki.com/#Team_decided_to_WONT_FIX_the_issue), we could well find that there is no tiddler titled `"Team_decided_to_WONT_FIX_the_issue"` and so try to load `"Team decided to WONT FIX the issue"`, which will also fail because the original tiddler was titled `"Team decided to WONT_FIX the issue"` (note the single underscore still used.)  Then we have to decide whether it's all-or-none, or whether we want to search for all 2<sup>6</sup>, or `64` possible combinations, including such monstrosities as `"Team_decided to WONT_FIX_the issue"`.

Again, maybe it's mostly a philosophical issue, but I think that we really would prefer our encoded fragment to represent a single title, and not multiple ones depending upon what titles happen to be present in the current wiki.

> This side effect is unlikely to cause problems because it relates to legacy wikis, only "if you upgrade to a wiki containing these new permalinks is it even a question?" and the preference is for backward compatibility.

The backward compatibility I've been striving for is simply that legacy permalinks/views should continue to properly load the appropriate tiddlers.  You have correctly demonstrated that there was a potentially larger class of permalinks for which this  would fail with my first code.  I'm simply trying to reduce the damage.   Your suggestion gets us part-way there.  It may be our best bet, but it introduces an intellectual overhead I'd like to avoid.

saqimtiaz:
@CrossEye I think the goal needs to be ensuring that a permalink only maps to a single tiddler title while maximizing backwards compatibility. We should reconsider the original proposal of using the `+` character as the encoding for a space.

Jermolene:
Hi @CrossEye thanks for your patience with this. To add to @saqimtiaz's comment, it is also very important that neither the encoding nor decoding should depend on the state of the tiddler store. In other words, encode/decode must be pure functions.

AnthonyMuscio:
If spaces are changed to underscore in the custom permalinks existing underscores remain untouched there is no complexity here.

Test for the incomming tiddler titles and if it does not exist translate it to the new permalink form and open that.

pmario:
@CrossEye -- I think this PR still has value. We should change the `SPACE_SUBSTITUTE` from underscore `_` back to plus `+`, which probably is more unlikely to be in a tiddler title. 

On the other hand, it will never be backwards compatible, since we only restrict 5 characters `| [ ] { }` from titles.

@Jermolene -- If we would like to go with `+` we should also add "+" to the warning in tiddler titles and see what happens. We may do that sooner than later, to get some feedback.

CrossEye:
@pmario:

> I think this PR still has value. We should change the SPACE_SUBSTITUTE from underscore _ back to plus +, which probably is more unlikely to be in a tiddler title.

Yes, that was my thought too.  I expect to get to this next weekend, if not before.

> On the other hand, it will never be backwards compatible, since we only restrict 5 characters | [ ] { } from titles.

Because the current mechanism converts `+` to `%2B`, the only backward incompatibility will be if someone handcrafted a URL to a tiddler using `+` rather than `%2B`.  For example, if I have a tiddler `Foo+Bar`, TW would give a permalink of https://tiddlywiki.com/#Foo%2BBar.  But if the user happened to hand-create the URL https://tiddlywiki.com/#Foo+Bar, TiddlyWiki currently would happily load `Foo+Bar`.  With this change, it would try to load `Foo Bar`.  

This is a much smaller issue that what happens with `_`, which isn't converted.

> If we would like to go with + we should also add "+" to the warning in tiddler titles and see what happens. We may do that sooner than later, to get some feedback.

That isn't necessary.  We will continue to encode `+` as `%2B`.  We are trying to make more readable as many URLs as possible, but there will still be percent encoding for many characters.

CrossEye:
New version in #7990

PR #7734, opened 2023-09-11
Introduce Tour Plugin and Confetti Plugin, improve Dynannotate Plugin

Jermolene:
This PR introduces the Tour Plugin that allows interactive learning tours to be created and presented in TiddlyWiki.

The PR also includes a new Confetti Plugin that allows animated bursts of confetti to be displayed, and improvements to the Dynannotate Plugin to add the ability to highlight screen elements using an animated spotlight effect.

A preview build is available [via Vercel](https://tiddlywiki5-git-confetti-plugin-jermolene.vercel.app/tour.html).

The goals of the Tour Plugin are:

* To break down learning experiences into a bite-sized steps, with actions required before the user can proceed to the next step
* To be able to have custom tours loaded into a wiki at once. The plugin includes a basic tour called "Introduction to TiddlyWiki".
* To be able to use the plugin to build a standalone guide at https://tiddlywiki.com/tour.html, a demo wiki that contains basic information about the solar system, and autostarts the "Introduction to TiddlyWiki" tour
* For people publishing with TiddlyWiki to be able to present custom tours to their own wiki

The plugin is currently independent of the existing [Introduction Edition](https://tiddlywiki.com/editions/introduction/) because it addresses different goals: the purpose of the introduction edition was to be able to produce the [introduction video](https://www.youtube.com/watch?v=KtCUr83XgyE), and it was not designed for reusability.

Screenshot 2023-09-11 at 09 37 16

Confetti effect:

Screen Recording 2023-09-11 at 09 55 45

Spotlight effect:

Screen Recording 2023-09-11 at 09 48 36

Progress:

* [x] UI for switching between different tours
* [ ] Cover tags in the introduction tour
* [ ] Cover wikitext in the introduction tour


Jermolene:
Updated the OP to add an animated GIFs of the spotlight effect and the confetti effect.

plugins/tiddlywiki/tour/globals.tid
```diff
@@ -36,6 +36,31 @@ text: longer description of the tour
 <$transclude $tiddler={{$:/state/tour/step}} $field="enter-actions"/>
 \end
 
+\procedure tour-chooser(filter:"[all[shadows+tiddlers]tag[$:/tags/Tour]]")
```
pmario, line 39: IMO `\whitespace trim` for every procedure or macros that uses indentation is missing. Did not check any other code. Just this one sprang into the eye

pmario: Ah. Good to know. I have missed that. Did we document that one?

Jermolene: I couldn't find any mention in the docs so made an update in bd99cf3385ef876cfeae8863997948821c7e7c13

plugins/tiddlywiki/tour/tours/introduction-to-tiddlywiki/steps/welcome.tid
```diff
@@ -4,8 +4,8 @@ display-mode: fullscreen
...
 ! Hello!
 
-!! An interactive tour of ~TiddlyWiki
+!! An interactive tour of {{$:/config/Tours/IntroductionToTiddlyWiki/ProductName}}
 
-Welcome to our tour of ~TiddlyWiki.
+Welcome to our tour of {{$:/config/Tours/IntroductionToTiddlyWiki/ProductName}}.
 
-We hope you'll find ~TiddlyWiki a helpful and supportive tool. Let's get started!
+We hope you'll find {{$:/config/Tours/IntroductionToTiddlyWiki/ProductName}} a helpful and supportive tool. Let's get started!
```
pmario, line 11: I'm not 100% sure, but would `{{$:/config/Tours/ProductName}}` not be enough?

Jermolene: The product name is a configuration option that is specific to the "Introduction to TiddlyWiki", it is not a general configuration option of the tour plugin.

pmario:
There is "Finding recent tiddlers", where the text in the green box is not complete.

image

Jermolene:
> There is "Finding recent tiddlers", where the text in the green box is not complete.
Thanks @pmario I've just pushed a fix in fd5c073

saqimtiaz:
@Jermolene I have run into a small issue while starting to update the documentation submission edition to use the updated tour plugin. One of the modifications I had made to the copy of the tour plugin used in that edition was to support a `condition` field for each step tiddler, where this field was evaluated as a filter to determine whether the step should be shown or skipped. Note that this is consistent with how we use the `condition` field other places in the core, such as for the display of editor toolbar buttons.

However, I note that the updated Tour plugin is now using the `condition` field as a success criteria of sorts to determine whether the step has been completed and whether the `next` button should be shown. I would like to add support for a field optionally providing a filter that determines whether a step is shown or skipped, and the most appropriate name for such a field in my opinion is `condition` in order to be consistent with other usage in the core.

Could we rename the current `condition` field in the Tour plugin implementation to `criteria` for instance?

saqimtiaz:
I welcome any suggestions for a better new name for what is currently the `condition` field in a tour step tiddler. I keep vacillating between `criteria` and `success`.

pmario:
I would favour `success` instead of criteria, since criteria and condition are way to similar for me as a non-native English speaker.

pmario:
The [thesaurus for success](https://www.thesaurus.com/browse/success) also suggests: `advance`

saqimtiaz:
I quite like `advance`. I'll wait to hear from @Jermolene, as the renaming is going to be a bit of a cumbersome process and I would rather not have to do it several times.

saqimtiaz:
Worth noting here that we have procedures with names like `tour-step-condition-satisfied` and `tour-step-no-condition` that would also need to be renamed for consistency. None of `criteria`, `success` or `advance` are thus a suitable direct replacement.

I am going to proceed with a working field name of `advance-criterion` until we have consensus.

saqimtiaz:
**List of desirable changes from my perspective:**

* rename `condition` field as currently used and introduce a condition field as used elsewhere in the core that allows skipping steps
* make it possible to suppress the automatic call to `<<tour-buttons>>` in each step allowing the author to determine where to invoke that procedure.
* make it possible to suppress the confetti automatically displayed on the last step of a tour.
* make it possible to suppress the tour chooser automatically displayed in the header of each step.
* optional additional class names for each tour specified via the `class` field in the tour definition tiddler.
* make the "show me a hint" text configurable. In earlier versions of the plugin this could be specified via a `hint` field in the step tiddler.

The two options that come to mind in terms of making things configurable are either optional fields in the tour definition tiddler that act as toggles, or the ability to have a tour specific variables tiddler that sets variables specific to that tour. Such variables could be used both to toggle optional elements, but also for example to override or extend the tour buttons for a given tour with a different navigation mechanism by redefining the relevant procedure(s).

**Other considerations:**
* consider not making all `tour-*` procedures global to avoid cluttering the global variable namespace and instead import them as needed
* consider how the `selector` field is hardcoded to be used by the `tm-spotlight` message. It would be desirable in the future to add more mechanisms for highlighting elements in tour steps (as opposed to showing a hint when requested) which could make the `selector` field name ambiguous. Could we perhaps rename this to `hint-selector` ?

Jermolene:
Even though there is room for improvement I'm going to merge this now because that's the best way to get further feedback. The plugin is entirely self contained and doesn't affect anything else.

PR #7735, opened 2023-09-12
remove ";" from $:/themes/tiddlywiki/vanilla/settings/fontfamily

BurningTreeC:
the tiddler gets transcluded in the stylesheets like so:

```
font-family: {{$:/themes/tiddlywiki/vanilla/settings/fontfamily}};
```

note - the semicolon at the end
So this semicolon is superfluous


pmario:
good catch

Jermolene:
Thanks @BurningTreeC

PR #6081, opened 2021-10-02
Basic Backtranscludes filter

linonetwo:
## Summary

Feature "Adding backtranscludes filter"

## Purpose of change

This fixes https://github.com/Jermolene/TiddlyWiki5/issues/847

Add filters and indexers similar to links.

## Describe the solution

I basically copy&paste the code and replace "link" to "transclude" :P

## Testing

Copy&paste the test from links...

~~But I don't know how to run tests locally, there is no npm script to run `npm run test` for.~~

I see tests are passing in the CI.

## Additional context

Demo: https://tiddlywiki5-6ivxvwmuq-jermolene.vercel.app/#transcludes%20Operator%20(Examples):%5B%5Btranscludes%20Operator%20(Examples)%5D%5D%20%5B%5Btranscludes%20Operator%5D%5D%20%5B%5Bbacktranscludes%20Operator%20(Examples)%5D%5D%20%5B%5Bbacktranscludes%20Operator%5D%5D

This can also closes https://github.com/Jermolene/TiddlyWiki5/pull/4766

Example screenshots:

截屏2021-10-02 下午6 52 42
截屏2021-10-02 下午6 51 56

With this widget:

```js
<$list filter="[all[current]backtranscludes[]first[]sort[title]]">
	<label style="margin-left:auto;margin-right:auto;width:fit-content;display:block;">Transclusion (other)-{}->(this)</label>
</$list>

<<inverse-link-and-folder_list-links filter:"[all[current]backtranscludes[]]" class:"horizontal-list-links" >>

<$list filter="[all[current]backlinks[]first[]sort[title]]">
	<label style="margin-left:auto;margin-right:auto;width:fit-content;display:block;">Link (other)-[]->(this)</label>
</$list>

<<inverse-link-and-folder_list-links filter:"[all[current]backlinks[]]" class:"horizontal-list-links" >>
```

P.S. I'm try using PR template from https://github.com/CleverRaven/Cataclysm-DDA/blob/master/.github/pull_request_template.md , they are generating formated release note based on Summary section of PR.

PR #7744, opened 2023-09-15
Feat: allow link to ^id anchor

linonetwo:
This is block-level link. Prepare for block-level backlink. ~~(wait for https://github.com/Jermolene/TiddlyWiki5/pull/6081 so I can continue modify back-indexer)~~

Add a new `^id` syntax, see https://github.com/Jermolene/TiddlyWiki5/discussions/7537 for details.

Also modify pretty wiki link's parsing rule to allow link to such an id. And modify link widget to allow id widget to react to link jumping.

(Currently, if `New Tiddler` is not currently opened, `hookFocusElementEvent` can't highlight the element, I have to use `setTimeout` to wait for the dom element actually show. Is there a better solution?)

And sadly, we can't use `#xxx` as ID, otherwise will encounter `Invalid selector:  ##BlockLevelLinksID1`

Demo: https://tiddlywiki5-git-fork-linonetwo-feat-d29eaa-jermolenes-projects.vercel.app/#Block%20Level%20Links%20in%20WikiText:%5B%5BBlock%20Level%20Links%20in%20WikiText%5D%5D%20%5B%5BLinking%20in%20WikiText%5D%5D

```tid
title: New Tiddler 1

[[New Tiddler^sdfd]]

[[New Tiddler^bbbb]]

[[ddd|New Tiddler^ddd]]

sdfa
```

```tid
title: New Tiddler

\```css
.main {
  display: none;
}
\```

^sdfd

asfd ^bbbb

* adfgd ^ccc
** dsaffda ^ddd
** adsfad
```


kookma:
This is a minor comment: The highlighted color would be nice to follow the palette in use.
This is another minor comment: a toolbar button/shortcut to create unique/sequence when creating section mark.

kookma:
Some use cases:

1. with footnotes, click to show the footnote at the tiddler bottom
2. with references like refnotes, click to highlight the reference in the bibliograpgy/references tiddler
3. partial transclusion when start and end of section is defined

linonetwo:
> The highlighted color would be nice to follow the palette in use.

Sure, I'm using the existing `<<colour highlight-background>>`, I'm not sure if there are some better configurable alternatives.

https://github.com/Jermolene/TiddlyWiki5/blob/db83401a69a6a72e69dd73c0f954280745910da6/themes/tiddlywiki/vanilla/base.tid#L2416-L2420

linonetwo:
Docs Added https://tiddlywiki5-git-fork-linonetwo-feat-section-mark-jermolene.vercel.app/#BlockIdWidget:BlockIdWidget%20%5B%5BBlock%20Level%20Links%20in%20WikiText%5D%5D%20%5B%5BLinking%20in%20WikiText%5D%5D

kookma:
> Sure, I'm using the existing `<<colour highlight-background>>`, I'm not sure if there are some better configurable alternatives.

I think this is good to go!



> I think Obsidian can only transclude a single block.

That is fine.

Thank you

kookma:
If a line started with space or tab then TW does not recognize the the `^id` For example

```
 ^a110 this is a test
```
Note to the space before `^a110`. So it seems leading white spaces on the beginning of a line started with an ^is` is an issue here.

kookma:
Suggestion:
1. Is there any limitation for choosing an id? If so please add these limitation to documentation
2. I would recommend to not allow special chars in id like `{}` or `<<` etc. Keep it simple

linonetwo:
The doc is at https://tiddlywiki5-git-fork-linonetwo-feat-section-mark-jermolene.vercel.app/#Block%20Level%20Links%20in%20WikiText

You can modify it by PR to my branch.

pmario:
@linonetwo In the discussion #7744 you wrote, that the IDs would be "namespaced". So if the same ID is used in 2 tiddlers that are visible at the same time it would be possible to deterministically link to them. 

In the preview that's **not** the case atm.

Both "instances" which render the content have enough information to make it happen.
eg: Both links using the code below need to work if they are visible at the same time.

But I think it can work. 

```
title: aaa

some text ^link1

[[ccc^link1]]
```

and 

```
title: ccc

some text ^link1

[[aaa^link1]]
```

----

1)  When the ID element `^link1` is rendered, the name of the tiddler is known. So the HTML `id` attribute can be created with a unique prefix based on the title of the tiddler. 
2) When the `[[aaa^link1]]` is created, it should have enough information to link to the same ID as described in 1.
3) Due to the [limiting rules for ID attribute values](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id)  we should do some prefix normalisation similar to the "qualify" mechanism. 


So I think it should be possible. I'm experimenting atm.

linonetwo:
Okay, I forgot this. I think simply prefixing the id with the tiddler title is enough.

pmario:
>But blockid ast node don't know itself is in which tiddler...So blocked widget don't know this too! Can you get title in the BlockIdWidget.prototype.render method?

Should be possible to get the `currentTiddler` or the `transclude` variable. ID's cannot be prefixed with the title since not every character is allowed. Have a look at my MDN link from the other post

linonetwo:
Ok, Now it works, I also update the doc to use same id in two tiddlers.

linonetwo:
Can this totally replace the [Linking within tiddlers - "anchor links"](https://tiddlywiki5-qvl6sz65c-jermolene.vercel.app/#Anchor%20Links%20using%20HTML) usage?

core/modules/widgets/blockid.js
```diff
@@ -54,6 +66,14 @@ BlockIdWidget.prototype.hookNavigatedEvent = function(event) {
 	return false;
 };
 
+BlockIdWidget.prototype.hookNavigationAddHistoryEvent = function(event) {
+	// DEBUG: console this._isNavigateToHere(event)
+	console.log(`this._isNavigateToHere(event)`, this._isNavigateToHere(event));
```
pmario, line 71: In the TW core we still use ES5 only. So JS template literals are not possible atm: `` `this._isNavigateToHere(event)` ``
So this code needs to be changed.

linonetwo: Sorry, this is debug statement, forget to delete...

linonetwo: Created by this in vscode

```
  "Debugger statement": {
    "prefix": "de",
    "scope": "javascript,javascriptreact,typescript,typescriptreact",
    "body": ["// DEBUG: debugger", "debugger;"],
    "description": "Plain debugger statement"
  },
  ```

**Comment by pmario** [comment]
@CrossEye -- Can you have a closer look at this one. IMO it's related to your implementation of the permalink creation.

**Review by Jermolene — COMMENTED** [review]
Thank you @linonetwo and apologies for the delay in reviewing this.

I have a number of concerns about the complexity and brittleness of this implementation:

* The use of hooks for widgets to communicate with one another is not a technique that we use in the core. Hooks are provided for plugin developers to be able to experiment with modifications to the core behaviour without having to modify the core code. For core improvements, we can make any changes we need, and so don't need the blunt instrument of hooks
* The way that the parse rule can refer to the previous or current block is deeply confusing
* I don't understand the need for the "blockid" widget; I would expect the syntax rule for an anchor just to generate a simple `<a>` element. I suspect that this is because of the way that block ids can apply to the previous block

I think it would be clearer to avoid the term "block"; in TiddlyWiki, we use the word when we're talking about parser modes, but we do not refer to blocks within tiddlers. It would be clearer to just call them "anchors".

**Comment by linonetwo** [comment]
- I will change it to "anchor" later
- I want the line or block to be highlighted, I put this logic in the newly added widget's code. If we don't need a new "Anchor widget", then this code will have to be in `core/modules/widgets/navigator.js`, is that OK? I think `navigator.js` is already very long.
- About "previous block", we can't add `^id` after codeblock, because codeblock's parse rule needs to be ended with `\n`. So we need to put `^id` in a new line, and define this situation as `refer to previous block`, this is how Obsidian handle this, I don't know a better way.

**Comment by Jermolene** [comment]
Thanks @linonetwo 

> * I will change it to "anchor" later

Thank you

> * I want the line or block to be highlighted, I put this logic in the newly added widget's code. If we don't need a new "Anchor widget", then this code will have to be in `core/modules/widgets/navigator.js`, is that OK? I think `navigator.js` is already very long.

Is the idea that the anchor is a container that contains all the content up to the next anchor?

> * About "previous block", we can't add `^id` after codeblock, because codeblock's parse rule needs to be ended with `\n`. So we need to put `^id` in a new line, and define this situation as `refer to previous block`, this is how Obsidian handle this, I don't know a better way.

I think the way to handle that would be to update the code block widget to allow the anchor syntax.

**Comment by pmario** [comment]
>... I will have to modify other block parser like block quote...But maybe it worth it? I will have a try.

hmmm, You need to make sue, that it's backwards compatible with plugins that rely on the existing parsing behaviour of "block rules". 

So I'm not sure if that's a good idea.

**Comment by linonetwo** [comment]
@Jermolene Please review this again! A qq group member ask for paragraph level transclusion, but I can't implement it without this syntax.
 And I want this to be official, so more plugins can interoperate on this.

**Comment by linonetwo** [comment]
with this PR, I can implement obsidian plugin like this

**Comment by pmario** [comment]
@Jermolene ... I think this PR would be a nice feature.

**Comment by CrossEye** [comment]
@pmario: Somehow I missed this comment months ago:

> @CrossEye -- Can you have a closer look at this one. IMO it's related to your implementation of the permalink creation.

While I will dig into the actual code when I'm not on mobile, I don't really see how this would be related.  At a feature level, they don't have much in common.  Are there similar implementation ideas?

**Comment by dongrentianyu** [comment]
@Jermolene Could you check this request again? I really wish tiddlywiki would make this possible.

**Comment by linonetwo** [comment]
@Jermolene Could you check this before 5.2.4?

**Comment by dongrentianyu** [comment]
@Jermolene we need you to check this pr.

**Comment by Jermolene** [comment]
Hi @linonetwo thank you for your continued work on this. I recognise that there is a strong user demand for anchors and I am keen to get them into the core.

Two areas that I think need more work is documentation and tests. There should definitely be tests for the new parse rules, and it may be possible to create a rig so that the navigation can be tested too.

I am also concerned how this implementation will work with static renderings. For example, anchor links should work within a static rendering such as https://tiddlywiki.com/alltiddlers.html. There are obviously some interactive features that we cannot support in static renderings, but basic link navigation needs to work as expected.

**Comment by linonetwo** [comment]
I've add basic ast test, will add more later.

> create a rig

What does this mean, where is it? What case other than ast test can I add?

> work with static renderings

Can you explain how https://tiddlywiki.com/alltiddlers.html#TiddlyWiki works? Will it automatically works when `#` points to a `h2` element?

**Comment by Jermolene** [comment]
> > create a rig
> 
> What does this mean, where is it?

It's quite a challenge, but it would be helpful to be able to have formal tests for the navigation primitives.

One approach might be like the [performance test rig](https://github.com/TiddlyWiki/tiddlywiki5-performance-test-rig) and use Puppeteer to test that navigation works as expected.

I think though that we can do a little better than that, and devise a way to run navigation tests within the testcase mechanism. We could certainly test things like the story list changing in response to a navigation, but perhaps we can go further and test things like the location hash handling by simulating a bit more of the browser.

> 
> > work with static renderings
> 
> Could you explain how https://tiddlywiki.com/alltiddlers.html#TiddlyWiki works? Will it automatically works when `#` points to a `h2` element?

These static renderings use anchor tags for links between tiddlers. We would need to devise a way for the tiddler anchors to co-exist with the section mark anchors. At the moment there is no name spacing applied to the anchors, and so there wouldn't be a way to create a section anchor that was guaranteed unique, and couldn't clash with a tiddler anchor.

**Comment by linonetwo** [comment]
If there is a test env already, then I can add it, just like add new cases into test edition, otherwise it will be too much for this PR. And I thought there was a Puppeteer related PR?

I find `<a href="#div_id">jump link</a>` can jump to `<div id="div_id"></div>`, and I will add tiddler title as name space prefixed in id.

And I'm going to rename the widget to `block-mark`, because its previous name `anchor` has conflict with `a` tag.

File: editions/tw5.com/tiddlers/widgets/LinkWidget.tid [hunk 2]
```diff
@@ -16,6 +16,7 @@ The content of the link widget is rendered within the `<a>` tag representing the
 
 |!Attribute |!Description |
 |to |The title of the target tiddler for the link (defaults to the [[current tiddler|Current Tiddler]]) |
+|toAnchor |<<.from-version "5.3.2">> Optional id of the anchor for [[Block Level Links in WikiText]] |
```
Leilei332, line 19: The version number should be changed to 5.3.6 now.

linonetwo:
How to test it,

`node ./tiddlywiki.js ./editions/tw5.com-server --build static --output ./output --verbose`

`node ./tiddlywiki.js ./editions/tw5.com-server --rendertiddler $:/core/templates/alltiddlers.template.html alltiddlers.html text/plain`

don't give any output html file.

linonetwo:
Search for `Anchor Links using HTML`, and click the link in `see Block Level Links in WikiText`.

[alltiddlers.html.zip](https://github.com/user-attachments/files/17230024/alltiddlers.html.zip)

pmario:
@linonetwo @Jermolene -- Did you see this: https://frontendmasters.com/blog/cool-people-link-to-text-fragments/

Jermolene:
> @linonetwo @Jermolene -- Did you see this: https://frontendmasters.com/blog/cool-people-link-to-text-fragments/

I'm familiar with text fragments, they are very useful. They work OK with TiddlyWiki because we explicitly filter them out of our own location hash processing:

https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.23#:~:text=organise%20your%20wedding.%20Record%20every%20thought%20that%20crosses%20your%20brain%2C%20or%20build%20a%20flexible%20and%20responsive%20website.


The challenge we have with TiddlyWiki is that it is not possible to combine a permalink|view with a text fragment, which means that text fragments can only target tiddlers within the default tiddlers filter.

linonetwo:
@pmario This PR is not very aimed at "jump to section", I'm trying to provide a new syntax that can be used to transclude a section / filter sections (with section mark prefixed with something). The jump to link implementation is just an example of usage. The key part is the syntax.

Once syntax is decided, those features could be provided by plugins. I'm not going to put too much things on the core.

pmario:
>The challenge we have with TiddlyWiki is that it is not possible to combine a permalink|view with a text fragment, which means that text fragments can only target tiddlers within the default tiddlers filter.

@Jermolene  Using FireFox Ubuntu 24.04 test this

https://tiddlywiki.com/#asdf:asdf%20Languages:~:text=language%20to%20the%20library%20by%20learning

Jermolene:
> @Jermolene Using FireFox Ubuntu 24.04 test this

Very good; that works correctly on the latest Firefox, Chrome and Safari on macOS.

linonetwo:
Whiteboard may have deeplink https://tldraw.dev/examples/basic/deep-links

This remaind me of the URL syntax to section mark, we could use `https://tiddlywiki.com/archive/full/TiddlyWiki-5.1.23#New Tiddler^sdfd` to jump to a section mark. And that may also works in a whiteboard tiddler.

dhjackal:
Adding to this thread.....I think this new feature would be a very powerful and useful addition to the code base

linonetwo:
> Using FireFox Ubuntu 24.04 test this
> 
> https://tiddlywiki.com/#asdf:asdf%20Languages:~:text=language%20to%20the%20library%20by%20learning

@pmario This doesn't works with Whiteboard. And once text change slightly, it fails. And this doesn't allow us to transclude a section.

pmario:
>This doesn't works with Whiteboard. And once text change slightly, it fails.

I do not know Whiteboard. -- Text fragments are a W3C specification. So, if Whiteboard can not handle it, it needs to be changed to make it work. 

You are right. If the text for "fragmented links" changes, the links break. So the whole thing is brittle. But as described at: https://developer.mozilla.org/en-US/docs/Web/URI/Reference/Fragment/Text_fragments there are ´text-start, text-end, prefix- and -suffix´ parameters, to make it more specific and less brittle. Text fragments have been designed for static web pages, that do not change often.

linonetwo:
Personal notes in Tiddlywiki changes a lot.

And the section backlinks, section transclud awaits. I could have write them 2 years ago. So what are we waiting for?

linonetwo:
I've adapt the #8258 . This feature works best with WYSIWYG editor. See how Obisidian works.

kookma:
@linonetwo  
Where can I find the link to the demo for testing?

linonetwo:
@kookma It is on first and second message on this PR. https://github.com/TiddlyWiki/TiddlyWiki5/pull/7744#issue-1899065955

The behavior here is just the minimal usage. The true usage of this is block level backlink and transclusion, that is TBD. Because I find the larger my PR, the slower the accept.

core/modules/widgets/navigator.js
```diff
@@ -148,7 +149,7 @@ NavigatorWidget.prototype.handleNavigateEvent = function(event) {
 	if(event.navigateTo) {
 		this.addToStory(event.navigateTo,event.navigateFromTitle);
 		if(!event.navigateSuppressNavigation) {
-			this.addToHistory(event.navigateTo,event.navigateFromClientRect);
+			this.addToHistory(event.navigateTo,event.navigateFromClientRect,event.toBlockMark);
```
pmario: IMO it would be better to add an `options` object to the `.addHistory()` prototype. So we will be more flexible in the future. We can use a stable API and add more options if needed. eg:

```
this.addToHistory(event.navigateTo,event.navigateFromClientRect,{anchor:event.toBlockMark});
```

core/modules/widgets/navigator.js
```diff
@@ -135,9 +135,10 @@ NavigatorWidget.prototype.addToStory = function(title,fromTitle) {
 Add a new record to the top of the history stack
 title: a title string or an array of title strings
 fromPageRect: page coordinates of the origin of the navigation
+anchor:optional anchor id in this tiddler
 */
-NavigatorWidget.prototype.addToHistory = function(title,fromPageRect) {
-	this.story.addToHistory(title,fromPageRect,this.historyTitle);
+NavigatorWidget.prototype.addToHistory = function(title,fromPageRect,anchor) {
```
pmario: IMO it should be 

```
NavigatorWidget.prototype.addToHistory = function(title,fromPageRect,options) {
```

pmario:
@linonetwo ... can you please make this a draft. It seems to break my unilink-plugin

pmario:
@kookma 
The latest build can be found at: https://deploy-preview-7744--tiddlywiki-previews.netlify.app/

linonetwo:
@pmario What's the error, let's fix it quickly.

pmario:
>@pmario What's the error, let's fix it quickly.

I am at it. It's probably not this PR, but the pre-release itself -- I'll create a new issue atm.

core/modules/widgets/navigator.js
linonetwo: Updated

linonetwo:
@Jermolene include this in 5.4.0 ?

Jermolene:
Hi @linonetwo as I’ve said, I think section links would fill in a very important gap in TiddlyWiki’s capabilities. However, it raises much broader architectural questions than many new features. That means that I want to give it more consideration than usual. 

There is also some feedback above that I think is still outstanding:

* Using the term “blocks”
* Anchors referring to the previous block rather than the next one. I appreciate that it is consistent with the behaviour of an anchor at the end of a line, but it feels counter-intuitive as a block level feature

I am also uneasy about allowing widgets to reach up the tree and address their parents and siblings. As things stand, widgets are only aware of their children. It makes things a bit simpler, and it allows for optimisations we might like to think about in the future, like lazy refreshing where we only refresh widgets that are actually on the screen.

I should add that there are many things that I like about this implementation, particularly the syntax of putting the anchor at the end of the line.

I think we also need to think about broader, more ambitious ways of implementing the feature. One of our design techniques is to try to relentlessly break things down to the most fundamental feature. An example of a broader mechanism would be to allow link anchors to be attached to any widget, and possibly for links to target any widget whether or not it has an anchor attached.

There’s also the question of whether to extend the transclusion mechanism to allow anchors to be targeted. I think this could be done much more naturally if anchors were containers (potentially nested) that contain the content up to the next anchor at the same level.

I hope it is not discouraging that this PR isn’t being merged in its present form. You will have seen that many of my own PRs languish for years until prerequisites are in place, or final architectural decsisions made.

core/modules/utils/dom/dom.js
```diff
@@ -11,6 +11,19 @@ Various static DOM-related utility functions.
 
 var Popup = require("$:/core/modules/utils/dom/popup.js");
 
+/*
+Determines whether element 'a' contains element 'b'
+Code thanks to John Resig, http://ejohn.org/blog/comparing-document-position/
+*/
+exports.domContains = function(a,b) {
```
Jermolene: We already have an implementation of this function in $:/core/modules/utils/deprecated.js

Jermolene: All of these utility functions appear to duplicate copies in deprecated.js

core/modules/utils/dom/dom.js
```diff
@@ -144,7 +190,7 @@ exports.getPassword = function(name) {
 Force layout of a dom node and its descendents
 */
 exports.forceLayout = function(element) {
-	void element.offsetWidth;
+	var dummy = element.offsetWidth;
```
Jermolene: The previous formulation was clearer. Do we need to adjust eslint rules to make it work?

core/modules/utils/dom/dom.js
```diff
@@ -246,19 +292,23 @@ exports.copyToClipboard = function(text,options) {
 	}
 	if(!options.doNotNotify) {
 		var successNotification = options.successNotification || "$:/language/Notifications/CopiedToClipboard/Succeeded",
-			failureNotification = options.failureNotification || "$:/language/Notifications/CopiedToClipboard/Failed";
+			failureNotification = options.failureNotification || "$:/language/Notifications/CopiedToClipboard/Failed"
```
Jermolene: How did this missing semi-colon occur?

Jermolene (commented):
Thanks @linonetwo for all your hard work and persistence. I think the implementation is now suffering from the complexity of the approach. In particular, postProcessBlockIds() and the related code breaks the layering of the parser mechanism.

I can see that the fundamental problem is that the parser doesn't currently have a means for a child node to set properties on its parent. One could imagine a general purpose form of such a mechanism, but the narrow implementation here will just become deadweight. In its present form it seems brittle because of the way that the mechanism makes assumptions about the parent node.

I'm wondering now if we might not need to take a different approach. I've been toying with the idea of an `<$anchor>` widget that contains the content that is being anchored. and an attribute for the ID. We would then resolve anchors in the widget tree, jumping across to the DOM tree when we find it. We could allow anchors to be nested, with IDs being formed hierarchically.

These ideas need more work, but I think it's possible that they might lead to a more harmonious implementation.

core/modules/utils/dom/dom.js
linonetwo: probably my local ESlint made mistake

core/modules/utils/dom/dom.js
linonetwo: I see, maybe I forget to delete during rebase. I encounter this `deprecated.js` several times today.

linonetwo:
@Jermolene This time, I refactor it to put `blockId` into attribute, instead of add a new widget. But yes, this requires `postProcessBlockIds` modifying parient.

> an `<$anchor>` widget that contains the content that is being anchored. and an attribute for the ID

I think that was the previous version of this PR?

But that requires what you said:

> widgets to reach up the tree and address their parents and siblings.

so when finding what it points to, it could find the real "block" (by block, I mean Block Mode WikiText).

I guess anchoring is not context free, so can't be done without parient, either do this when parsing, or do this when rendering?

If you mean

```tid
<$anchor id="myId">
A paragraph.
</$anchor>
```

instead of ` ^id`, it would be difficult to write by hand, only possible to be generated by editor.

Or let each parser scan the children for anchors, using an util, but it will need to modify every rules, and that info has to be hold on the parent node, so anchor could be restore on serialization.

@Jermolene Do you still want the word to be "anchor"? Like `[[XXX]backlinks:anchor[]]` `[<sourceTitle>anchors[TargetTitle]]` , adding ` ^id` is adding an anchor, and the `id` part is "anchor name"?

Jermolene:
Hi @linonetwo 

> I think that was the previous version of this PR?
> 
> But that requires what you said:
> 
> > widgets to reach up the tree and address their parents and siblings.
> 
> so when finding what it points to, it could find the real "block" (by block, I mean Block Mode WikiText).

I don't recall the details, but my current thinking is that we would only need to descend the widget tree to find the anchor widgets.

> If you mean
> 
> ```
> <$anchor id="myId">
> A paragraph.
> </$anchor>
> ```
> 
> instead of ` ^id`, it would be difficult to write by hand, only possible to be generated by editor.

The `^id` syntax would be a shortcut for the widget.

> @Jermolene Do you still want the word to be "anchor"? Like `[[XXX]backlinks:anchor[]]` `[<sourceTitle>anchors[TargetTitle]]` , adding ` ^id` is adding an anchor, and the `id` part is "anchor name"?

Yes, I think the current terminology works well.

linonetwo:
Now, Any block can be tagged with a trailing `^id` on its opening line:

```
A paragraph. ^para1

!! Heading ^h1

(code fence ` ` `)css ^codeBlock
.foo { color: red; }
(code fence ` ` `)

<<< some cite ^quoteBlock
Some quote.
<<<

* list item ^listItem
```

### How it works (two-layer parse)

1. Inline anchor rule — a new `wikirule` of type `inline` matches ` ^id` at end of a line and emits an **empty placeholder node** (`{type:"anchor", name:"true"}`). This is transparent to all existing block rules; they don't need to know about anchors.

2. `wrapAnchorsInTree()` — called once after `parseBlock()` returns. It recursively scans the result tree, finds placeholder nodes inside block elements (`p`, `h1`–`h6`, `li`, `blockquote`, `codeblock`, `void`), and wraps the enclosing block in a **target anchor container** (`{type:"anchor", target:"true", children:[block]}`).

### Transclusion

- Single block: `{{Title^anchorId}}`  
- Range: `{{Title^startId^endId}}` — `^` introduces the start anchor, second `^` introduces the end anchor (anchor name don't allow containing `^`)

### No big changes to existing rules, no accessing parient widget

`codeblock`, `typedblock`, `quoteblock`, `list` — their opening lines are now parsed with `parseInlineRun` (as quoteblock already did for cite text), which naturally picks up the inline anchor rule. Zero special-casing inside block rules.

And now `wrapAnchorsInTree` read child ast node, not modifying children ast node, not accessing parient. Only wrap a container node.
At rendering time, no interaction between widgets are required.

linonetwo:
@Jermolene Need a revisit, I should have reimplemented based on your advise. (I'm open to change the plan again if you still found some problem)

pmario:
@linonetwo ... @Jermolene ... I would really like to have that functionality.

linonetwo:
Best time to have it is 2 years ago, when we try to catch up with Obsidian. Now we not only need this, but also need to catch up Obsidian + Claude, the most important feature set is moving.
But I still think this is useful, as you could transclude lines from other tiddlers to make a precise promot tiddler.

If you are interested, I will spend time solve conflicts again.

PR #7747, opened 2023-09-17
Don't refresh the body editor when switching the preview on and off

Jermolene:
A quick experiment to try to address #7745.


BurningTreeC:
Hello @Jermolene - this PR would be good to go!
I've tested it and it works well

Jermolene:
Thanks @BurningTreeC

PR #7749, opened 2023-09-21
type Edittemplate delete button should also delete the typeInputTiddler

BurningTreeC:
If we don't delete the typeInputTiddler with the click on the "delete" button then the dropdown stays filtered - but the text input seems to be empty. This PR corrects this behavior


Jermolene:
Thanks @BurningTreeC

PR #7751, opened 2023-09-22
feat: Delete hooks from the hashmap

linonetwo:
I found if you register a hook in a widget, for example previously in 

https://github.com/Jermolene/TiddlyWiki5/pull/7744/commits/7200f73cdc0f1bd99e46b9333dc84ec2b4371366

The hook is not cleanup after widget is unmounted. So add this method to cleanup to prevent memory leak.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -2679,11 +2679,9 @@ Delete hooks from the hashmap
 */
 $tw.hooks.removeHook = function(hookName,definition) {
 	if($tw.utils.hop($tw.hooks.names,hookName)) {
-		var index = $tw.hooks.names[hookName].findIndex(function(hook) {
-			return hook === definition;
-		});
-		if(index !== -1) {
-			$tw.hooks.names[hookName].splice(index, 1);
+		var p = $tw.hooks.names[hookName].indexOf(definition);
+		if(p !== -1) {
+			$tw.hooks.names[hookName].splice(p, 1);
 		}
 	}
 };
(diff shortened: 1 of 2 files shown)


boot/boot.js
```diff
@@ -2674,6 +2674,20 @@ $tw.hooks.addHook = function(hookName,definition) {
 	}
 };
 
+/*
+Delete hooks from the hashmap
+*/
+$tw.hooks.removeHook = function(hookName,definition) {
+	if($tw.utils.hop($tw.hooks.names,hookName)) {
+		var index = $tw.hooks.names[hookName].findIndex(function(hook) {
```
Jermolene, line 2682: We can't use findIndex() in the core

Jermolene (commented):
Thanks @linonetwo it makes sense to add this, just one small issue with the implementation

boot/boot.js
linonetwo: Ok, I think

```js
$tw.hooks.names[hookName] = $tw.hooks.names[hookName].filter(function(hook) {
  return hook !== definition;
});
```

does the same, but I use the way identical to `removeEventListener`

Jermolene:
Thank you @linonetwo. Apologies it's only just occurred to me that we should update the docs at https://tiddlywiki.com/dev/#HookMechanism as well, if you could? We just need a very brief note saying that hooks can be removed with `$tw.hooks.removeHook(handler)`.

linonetwo:
Updated. And I found the dev doc's toc is not very useful, I'm going to update it in another PR.

Jermolene:
Thanks @linonetwo

PR #7755, opened 2023-09-27
Proposed fix for refreshing transcluded functions

rmunn:
This is a proposed addition to PR #7698, which is an attempt at fixing issue #7696. Note that this PR is **not** merging into `master`, but into the target branch for #7698. I'm doing it this way because I can't push a commit to @Jermolene's PR branch directly.

To fix the issue where transcluded functions don't refresh, when it's time to refresh the transclude widget, if it was transcluding a function then that function is called again and the result list is compared. If the result list has changed then the widget refreshes. (This unfortunately will end up calling the function a *second* time, but avoiding that second call would have involved a large refactor that I don't have time for).

This is basically implementing the suggestion I made at https://github.com/Jermolene/TiddlyWiki5/pull/7698#issuecomment-1702144333.

With this, all tests now pass, including the new test introduced in #7698 to test the function-transclusion bug.

### DIFF
--- a/core/modules/widgets/transclude.js
+++ b/core/modules/widgets/transclude.js
@@ -109,6 +109,7 @@ TranscludeWidget.prototype.collectAttributes = function() {
 		this.recursionMarker = this.getAttribute("recursionMarker","yes");
 	} else {
 		this.transcludeVariable = this.getAttribute("$variable");
+		this.transcludeVariableIsFunction = false;
 		this.transcludeType = this.getAttribute("$type");
 		this.transcludeOutput = this.getAttribute("$output","text/html");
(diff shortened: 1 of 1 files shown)


rmunn:
A quick note about this code. There's one bit of code where you might think I made a mistake, because `if(srcVariable && srcVariable.isFunctionDefinition)` is checked twice inside `parseTransclusionTarget`. That's because if we only did that check *inside* the `if(variableInfo.text)` block, then it would be skipped if a filter function happened to return an empty list, and then the fix would work for all functions *except* those that returned an empty list on first render.

Jermolene:
Great thanks, @rmunn looks good, much appreciated.

Issue #7754, opened 2023-09-26
[IDEA] Automatically Switch Between Dark / Light Palette Depending on Browser Setting


pmario:
Started at: https://talk.tiddlywiki.org/t/config-tiddler-title-for-default-light-dark-palette-e-g-config-palette-default-dark/8071

**It is proposed** to automatically change to light / dark palettes depending on the browser setting. 

- 2 config tiddlers should define the default palettes: 
   - `$:/config/palette/default-dark -> $:/palettes/GruvboxDark`
   - `$:/config/palette/default-light -> $:/palettes/Vanilla`
- It should be possible to Enable / Disable the automatic mode

**Prove of Concept** 

![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/855a24ce-cbd1-46ee-8048-a20d8c34d365)

**Known Problems**

- With the current version of TW a browser reload is needed 
- The detection at: https://github.com/Jermolene/TiddlyWiki5/blob/773c1f83f24f5da81224a0b8f8abbb759041a9dd/core/modules/info/platform.js#L40 misses some functions
- **Separated PR**: It should be possible to dynamically activate an action-tiddler when the browser setting is changed

pmario:
I'm preparing a PR atm

linonetwo:
How to set $:/config/palette/default-light to $:/palettes/Notion using UI? 

Some people in QQ group say its annoying to customize wiki when there is no build-in config form for them. (When talk about sidebar position)

Jermolene:
Thanks @pmario – how would backwards compatibility be handled? I think it's problematic to migrate from a single configuration tiddler to two interrelated ones.

One could imagine an arrangement such as "Use the palette titles from $:/config/palette/default-dark and $:/config/palette/default-light if they exist, otherwise use the title from $:/palette" but this would break any existing code using the $:/palette tiddler.

pmario:
>Thanks @pmario – how would backwards compatibility be handled?

`$:/palette` will be untouched. 

The `$:/config/palette/default-dark - light` tiddlers are only used if the 
`$:/config/palette/enable-light-dark-detection` is set to `yes`. 

The `enable` tiddler does **not exist by default**. (Except we decide to enable that setting by default)

pmario:
> How to set :/palettes/Notion using UI?

[...]

PR #7756, opened 2023-09-27
implement browser light / dark theme switching detection

pmario:
This PR fixes:  **[IDEA] Automatically Switch Between Dark / Light Palette Depending on Browser Setting** #7754 

- At the moment the setting can only be activated on browser startup. 
- Once this PR is merged, there can be new functionality that dynamically switches setting, when the browser themes are changed.

**The ControlPanel looks like this atm**

image

**If the Configuration Detection is on**

- Showing the configuration is animated, so users do not loose focus
- There should be enough explanatory text for users to understand what's going on
- The configuration detection got its own tiddler

image

**New Macros**

There are 2 new macros, that allow plugin authors and users to use the same UI as the core for configuring palette settings.

- `<<palette-picker tiddler:abc>>` ... Creates a dropdown dialogue that allows the user to define the tiddler which will contain the palette name. Defaults to: `$:/palette`
- `<<palette-info tiddler>>` ... Shows the selected palette swatches -> `$:/snippets/currpalettepreview`

**Internationalisation**

- All texts are translatable -- Wording can be improved. Just let me know.

### DIFF
--- a/core/wiki/macros/palette-picker.tid
+++ b/core/wiki/macros/palette-picker.tid
@@ -1,0 +1,36 @@
+tags: $:/tags/Global
+title: $:/core/macros/palette-picker
+
+\whitespace trim
+
+\procedure palette-info(tiddler)
+\define tv-wikilinks() no
+<$tiddler tiddler={{{ [<tiddler>get[text]] }}} >
+	''{{!!name}}'' <span class="tc-tiny-gap">-</span> {{!!description}}
+	{{||$:/snippets/currpalettepreview}}
+</$tiddler>
+\end
+
+\procedure palette-picker(tiddler)
+<div class="tc-popup-keep" style="position:relative;">
+	<$button
+		popup=`$:/state/popup/$(tiddler)$`
+		tooltip={{$:/language/Buttons/Palette/Hint}}
+		aria-label={{$:/language/Buttons/Palette/Caption}}
+		class=<<tv-config-toolbar-class>>
+		selectedClass="tc-selected"
+	>
+		<$list filter="[<tv-config-toolbar-icons>match[yes]]">
+			{{$:/core/images/palette}}
+		</$list>
+		<$list filter="[<tv-config-toolbar-text>match[yes]]">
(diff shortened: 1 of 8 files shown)


BurningTreeC:
Hi @pmario
Does this also take into account that the dark mode of the os can change while a wiki is open?

pmario:
>Does this also take into account that the dark mode of the os can change while a wiki is open?

This setting has to be managed by the browser. If the browser is set to 

- "Automatic", it will use the setting from the OS. 
- Light -- The light theme is fixed even if OS is dark
- Dark -- The dark theme is fixed even if OS is light

So -- Yes

**FireFox**

image


**Edge**

image

core/wiki/macros/palette-picker.tid
```diff
@@ -29,8 +29,15 @@ title: $:/core/macros/palette-picker
...
 	</$button>
-	<$reveal state=`$:/state/popup/$(tiddler)$` type="popup" position="belowleft" tag=div class="tc-drop-down" style="font-size:0.7em;">
-		<$transclude $tiddler="$:/snippets/paletteswitcher" palette=<<tiddler>> />
+	<$reveal state=`$:/state/popup/$(tiddler)$`
+		type="popup"
+		position="belowleft"
+		positionAllowNegative="yes"
+		tag="div"
+		class="tc-drop-down"
+		style=`$(style)$`
```
Jermolene, line 38: Is there a reason not to use `style=<<style>>`?

pmario:
Sorry for wasting everyone's time. I'll close this UI centric PR and the related issue.

I'll create a new proposal related to the core function that detects the browser ["preferes-color-scheme" setting](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme).

Jermolene:
@pmario I do not think that this work was a waste of time at all. It is a valuable experiment, and the feedback can help us iterate towards a better solution. I welcome your attention on this issue, it is definitely the kind of modernisation that I would like to be addressing at the moment

PR #7762, opened 2023-10-01
Allow ")" in parameters default values for pragma definitions

xcazin:
The `matchRegExp` in wiki pragma rules for _macros_, _parameters_, _function_, _procedure_ and _widget_ definitions doesn't allow the character ")" in default values for parameters. For example, any of the following pragma definitions in a new tiddler would break normal parsing:

```
\parameters (foo:"successful (or at least it should be)")

Value of the parameter `foo`: @@<<foo>>@@
```
```
\define mymacro(foo:"successful (hopefully)")
 <<__foo__>>
\end

Value of the parameter `foo`: @@<<mymacro>>@@
```
```
\function myfunc(foo:"successful (hopefully)")
 [<foo>]
\end

Value of the parameter `foo`: @@<<myfunc>>@@
```
```
\procedure myprocedure(foo:"successful (hopefully)")
 <<foo>>
\end

Value of the parameter `foo`: @@<<myprocedure>>@@
```
```
\widget $my.widget(foo:"successful (hopefully)")
 <<foo>>
\end

Value of the parameter `foo`: @@<$my.widget/>@@
```
In this PR, `matchRegExp` values are modified in order to allow the previously excluded ")" so that the tests above become successful: as the first stage of the definition parsing, `matchRegExp` only needs to  capture the whole payload string between the parameter parentheses.


xcazin:
it looks like my changes actually break something else, as the following construct with the closing parenthesis on the next line does not work anymore in the Vercel:

```
\function myfunc(foo:"not successful"
)
 [<foo>]
\end

Value of the parameter `foo`: @@<<myfunc>>@@
```
I'll try to track down this issue as time permits...

pmario:
> it looks like my changes actually break something else,

Did you find about that from an existing test? If not, it would be nice, if you could add a new test, that checks about this case

pmario:
For me it seems to detect all the elements -- **but** I did not have a closer look if the groups still match the code

xcazin:
> > it looks like my changes actually break something else,
> 
> Did you find about that from an existing test? If not, it would be nice, if you could add a new test, that checks about this case

The code in [the above comment](https://github.com/Jermolene/TiddlyWiki5/pull/7762#issuecomment-1742642960) works on http://tiddlywiki.com/prerelease, but not on the preview vercel : adding a newline just before the closing parenthesis of the parameters definition gives:

Capture d’écran 2023-10-02 à 15 14 54

> For me it seems to detect all the elements -- **but** I did not have a closer look if the groups still match the code
> 
Thanks for looking into it. Another quick hypothesis is that parsing of the block 3 or 4 stumbles upon the linefeed.

PR #7765, opened 2023-10-04
Update chinese language files

BramChen:
* Improve consistency of naming file types and content types in some UI hints.


Jermolene:
Thanks @BramChen

PR #7768, opened 2023-10-04
Ensure {{}} doesn't cause a recursion error

### LINKED ISSUE #7767: [BUG] Problem with empty transclusions
### Describe the bug

In a new tiddler, with preview/output on, Entering something followed by an empty transclusion `{{}}` (not the backticks used here) will generate a loop which appears to transclude the current tiddler text a large number of times.
* "Because the current tiddlers text transcludes the current tiddlers text"?
* A second copy of the same line in the same tiddler may result in a loss of the wiki stored in browser memory with an "Awe Snap" error.
* Something similar can occur when using automatic braces eg `{{` with create the closing pair and if preview output is open trigger this error. Please see the [talk.tiddlywiki discussion](https://talk.tiddlywiki.org/t/curious-and-potentialy-dangerouse-empty-transclusions/8141?u=tw_tones)

### Expected behavior

an empty transclusion would be expected to result in no output because there is no tiddler to transclude.

Previously we may have expected `{{}}` to transclude a tiddler with no title and not work.
* We needed to indicate the currentTiddler was to be used with `{{currenttitle||template}}` or {{||template}}

If you use the below instructions to reproduce but only include `{{}}` and nothing else you will see the red message `Recursive transclusion error in transclude widget `

### To Reproduce

1. Go to '.tiddlywiki.com'
2. add a tiddler
3. type the word something followed by {{}}
4. Close and view or open preview > output

### Screenshots

Snag_69eaad6

Snag_69ebe9d

### TiddlyWiki Configuration

- Version [v3.5.1]
- Not applicable, Tiddlywiki.com and Empty.html
- Plugins installed as per tiddlywiki.com and empty.html (none)

- OS:Windows 11
- Browser FireFox  / Chrome
- Version 118.0.1 (64-bit) / Version 117.0.5938.150 (Official Build) (64-bit)

Not tested

### Additional context

My suspicion is the default content on the shortcut transclude now defaults when empty to "transclude current tiddler"  rather than to an "empty template tiddler title", this is not surprising given the 5.3.x changes.

Jermolene:
Fixes #7767 

As [reported on talk.tiddlywiki.org](https://talk.tiddlywiki.org/t/curious-and-potentialy-dangerouse-empty-transclusions/8141), the following construction causes a recursion error:

```
something {{}}
```

The cause is that the blank transclusion generates a transclude widget with no attributes. This causes the transclude widget to transclude the current tiddler, which frequently leads to a recursion error.

This PR fixes things by ensuring that `{{}}` generates `<$transclude $tiddler=""/>`, which can be rendered without errors.

It is worth noting that this actually transcludes the tiddler titled with the empty string, if there is one. An existing flaw of the core is that it is actually possible to create such a tiddler.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/transcludeblock.js
+++ b/core/modules/parsers/wikiparser/rules/transcludeblock.js
@@ -81,6 +81,9 @@ exports.parse = function() {
 			}
 			return [tiddlerNode];
 		} else {
+			// No template or text reference is provided, so we'll use a blank target. Otherwise we'll generate
+			// a transclude widget that transcludes the current tiddler, often leading to recursion errors
+			transcludeNode.attributes["$tiddler"] = {name: "$tiddler", type: "string", value: ""};
 			return [transcludeNode];
 		}
 	}
--- a/core/modules/parsers/wikiparser/rules/transcludeinline.js
+++ b/core/modules/parsers/wikiparser/rules/transcludeinline.js
@@ -79,6 +79,9 @@ exports.parse = function() {
 			}
 			return [tiddlerNode];
 		} else {
+			// No template or text reference is provided, so we'll use a blank target. Otherwise we'll generate
+			// a transclude widget that transcludes the current tiddler, often leading to recursion errors
+			transcludeNode.attributes["$tiddler"] = {name: "$tiddler", type: "string", value: ""};
 			return [transcludeNode];
 		}
 	}


AnthonyMuscio:
I have tested the preview and it seems to correct the specific case of `{{}}` as desired, by ignoring it. But possibly not other cases.

The somewhat similar shortcut transclusions `{{currentTiddler}}` and `{{||currentTiddler}}` continue to respond appropriately with a transclusion error. 

I am surprised to see I can't replicate the problem any more on tiddlywiki.com or the prerelease?

A quick test of the equivalent in the long hand version shows it also enters a loop `<$transclude/>` with the preview output open.

More work needs to be done, but the initial fix seems a good start.

pmario:
@Jermolene ... The "main hickups" seem to be gone now -- as [feedback at Talk](https://talk.tiddlywiki.org/t/curious-and-potentialy-dangerouse-empty-transclusions/8141/19?u=pmario) and my own tests show. 

I think it can be merged

shadowrevenant:
#7777 The Issue with "Empty" Transclusion is fixed in 5.3.2 but if you transclude the current tiddler the same crash happens (IE: Make "New Tiddler" Then {{New Tiddler}}) Causes page crashing

shadowrevenant:
> #7777 The Issue with "Empty" Transclusion is fixed in 5.3.2 but if you transclude the current tiddler the same crash happens (IE: Make "New Tiddler" Then {{New Tiddler}}) Causes page crashing

(Not Sure if this is the right issue to comment on but the two seem to go hand-in-hand

Jermolene:
> > #7777 The Issue with "Empty" Transclusion is fixed in 5.3.2 but if you transclude the current tiddler the same crash happens (IE: Make "New Tiddler" Then {{New Tiddler}}) Causes page crashing
> 
> (Not Sure if this is the right issue to comment on but the two seem to go hand-in-hand

I've commented over at https://github.com/Jermolene/TiddlyWiki5/issues/7777#issuecomment-1752092500

flibbles:
Oops! Just found out about this testing for https://github.com/Jermolene/TiddlyWiki5/pull/7866.

Uglify was substituting `{{}}` in as a tighter shorthand for `<$transclude />`, but I'll update Uglify to consider the versions.

I agree that `{{}} {{}}` should not make our browsers permanently ponderous.

flibbles:
Actually @Jermolene, I might suggest that instead of getting rid of `{{}}` which some people may have been using, it may be better to improve the infinite recursion testing. I looked at your comment over at https://github.com/Jermolene/TiddlyWiki5/issues/7777#issuecomment-1752092500, which talks about why `{{}} {{}}` causes a freeze. (and subsequently, so does `<$transclude /> <$transclude />` even now).

You are pretty much right that it's causing 2^1000 leaf nodes, but it doesn't have to be. I just looked into it. It actually figures out that it's exceeding recursion depth pretty quickly, but then it goes on to determine that again for every single leaf node.

Right now, the bottom-most transclusion prints the error, and everything above it still tries to render. If instead the transclusion were to rely on exception handling, then the bottom-most transclusion could throw *up* to a higher transclusion, which could print the error and stop rendering, thus preventing a bunch of extra junk from printing to the window like it does now. Then not even `<$transclude /> <$transclude />` would cause a crash. And I think infinite recusion handling would generally be more graceful.

Also, we could keep `{{}}`.

Would you like me to dig into how this could be done, and maybe start a discussion before v5.3.2 goes out?

Jermolene:
Hi @flibbles I'd be very happy to see the recursion testing improved, and therefore I'll go ahead and revert this PR.

CodaCodr:
Not sure if comments here are moot now...

1. `{{}}` is very easy to produce using the CM editor.
2. I was editing some "live" CSS (I know, not a great idea, but still) and typed `{{` and CM responded with `}}` giving `{{}}`. Oh dear. ~4mins before I got enough control back to reposition cursor and hit backspace.

Surprising how long 4 minutes of time is... :/

TW: 5.3.3

flibbles:
I'm absolutely ready to get this good-to-go. I'd just like some feedback on whether the custom error should be in its own namespace, and if the method I've selected for testing recursion is suitable enough.

If people like, I can just make my own judgement calls on all this and call it ready for merging.

PR #7769, opened 2023-10-04
Add data attribute support to button and other widgets

Jermolene:
This PR adds support for data attributes and `style.*` attributes to the button, browse, checkbox, link, radio, range and select widgets.

It also adds a `data-tab-title` attribute to tab buttons so that they can be easily located with CSS selectors.


saqimtiaz:
Hi @Jermolene, in my own work I have also needed to patch the $draggable and $droppable widgets to add support for data attributes. As such it might be desirable to add support for adding such attributes to all widgets that create DOM nodes.

I explored this a bit in the draft PR #6823 but the sticky point was how to handle refresh when an attribute is both added explicitly via a widget parameter and also as an arbitrary parameter.  An option would be to only add support for arbitrary data attributes which would also remove the risk of any overlap with existing widget parameters. I would be happy to explore this further.

Jermolene:
I've pushed an update that extends the implementation to add support for data- attributes to the browse, checkbox, link, radio, range and select widgets.

Jermolene:
I've pushed an update that extends the existing support for `style.*` attributes on HTML elements to the button, browse, checkbox, link, radio, range and select widgets

core/modules/widgets/widget.js
```diff
@@ -432,28 +439,38 @@ Widget.prototype.assignAttributes = function(domNode,options) {
...
+		// Check for a style attribute
+		} else if(name.substr(0,6) === "style." && name.length > 6) {
+			domNode.style[$tw.utils.unHyphenateCss(name.substr(6))] = value;
+		}
+	});
+};
+
+/*
+Set the data- and style. attributes of a parse tree node according to the attributes of the widget
+domNode: optional domNode to be assigned
```
saqimtiaz, line 463: I assume this comment about the domNode no longer applies?

core/modules/widgets/widget.js
```diff
@@ -432,28 +439,38 @@ Widget.prototype.assignAttributes = function(domNode,options) {
...
-				}
+			// Setting certain attributes can cause a DOM error (eg xmlns on the svg element)
+			try {
+				domNode.setAttributeNS(namespace,name,value);
+			} catch(e) {
 			}
 		}
 	}
-	// Not all parse tree nodes have the orderedAttributes property
-	if(this.parseTreeNode.orderedAttributes) {
```
saqimtiaz, line 448: Are we absolutely sure we do not need to respect the `orderedAttributes` order if it exists? I cannot think of any reason we do need to care about it but just wanted to flag this as something to review. Let Widget is the only one I can think of where order matters and it isn't impacted by this change.

core/modules/widgets/widget.js
Jermolene: Thanks @saqimtiaz

saqimtiaz:
@Jermolene a heads up that the LinkWidget implementation for data/style attributes is broken. Baking a fix plus tests into a PR that I am working on.

Jermolene:
> @Jermolene a heads up that the LinkWidget implementation for data/style attributes is broken. Baking a fix plus tests into a PR that I am working on.

Great, thank you @saqimtiaz.

I looked briefly at the ordered attributes problem and of course it will need a fair bit of refactoring to account for the optimisation with "changedAttribues". I won't be able to do that for the next day or two, so I'd very happy if you got a chance to look at that too.

saqimtiaz:
@Jermolene I have just moved on to the tour plugin work, but will circle back if I can find the time later today or tomorrow. If I do get the chance to work on the ordered attributes, I will leave a comment here when I start so that we do not duplicate effort.

Jermolene:
Great thanks @saqimtiaz

core/modules/widgets/widget.js
```diff
@@ -413,16 +413,23 @@ Widget.prototype.getAttribute = function(name,defaultText) {
...
 */
 Widget.prototype.assignAttributes = function(domNode,options) {
-	options = options || {};
 	var self = this;
+	options = options || {};
+	var changedAttributes = options.changedAttributes || this.attributes;
+	var sourcePrefix = options.sourcePrefix || "";
+	var destPrefix = options.destPrefix || "";
+	var EVENT_ATTRIBUTE_PREFIX = "on";
 	var assignAttribute = function(name,value) {
```
saqimtiaz, line 430: Yikes. I am surprised we never caught this earlier. This function is never called and we are assign attributes directly to the DOM node with no sanitation of event attributes!

saqimtiaz: We should add a test that explicitly checks to make sure that event attributes are not supported.

core/modules/widgets/widget.js
```diff
@@ -432,28 +439,23 @@ Widget.prototype.assignAttributes = function(domNode,options) {
...
-			assignAttribute(name,self.attributes[name]);
-		});	
-	}
+	$tw.utils.each(changedAttributes,function(value,name) {
+		value = self.getAttribute(name);
+		// Check for a prefixed attribute
+		if(name.substr(0,sourcePrefix.length) === sourcePrefix) {
+			domNode.setAttribute(destPrefix + name.substr(sourcePrefix.length),value);
+		// Check for a style attribute
+		} else if(name.substr(0,6) === "style." && name.length > 6) {
```
saqimtiaz, line 455: This logic is flawed. If no sourcePrefix is specified, such as in the element widget, then style. attributes do not work.

saqimtiaz: See https://github.com/Jermolene/TiddlyWiki5/pull/7852

Jermolene: Thanks @saqimtiaz

core/modules/widgets/widget.js
Jermolene: Ouch! Thanks @saqimtiaz

saqimtiaz:
@Jermolene I think this is now ready to be merged so we can get some user testing and feedback.

Jermolene:
Great thanks @saqimtiaz

PR #7774, opened 2023-10-06
Clean up after #7671 (Change favicon format from ICO to PNG)

mateuszwilczek:
Remove two unnecessary leftover files from #7671


Jermolene:
Thanks @mateuszwilczek

PR #7775, opened 2023-10-06
Improve Archive Overview

pmario:
This PR changes the overview to make it more approachable. 

I did split the different major version jumps into Milestones. So it should have more structure. 

In a second PR I'll try to use this tiddler as an index.html tiddler to be auto-created and stored at: https://tiddlywiki.com/archive 
So we have a nice index page.

image

### DIFF
--- a/editions/tw5.com/tiddlers/about/Archive.tid
+++ b/editions/tw5.com/tiddlers/about/Archive.tid
@@ -3,42 +3,81 @@ created: 20231005205623086
 modified: 20231005210538879
 tags: About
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario I don't think the subheadings are helpful. It looks as though the string "Unrestricted Fieldnames..." is a heading that applies to the following 8 releases.

Perhaps instead we might add a one line description to each of the release tiddlers, and include that as an additional column.

I do like the way that the ordering is reversed with the most recent release on the top.

PR #7776, opened 2023-10-06
add archive HTML wikis, add index.html + css

pmario:
- This PR should add the archive versions to the `archive` directory
- It should create an `index.html` + `static.css` file in the archive dir

@Jermolene ... I do not know it works with your workflow of creating a new major release, because I do not know how you activate it. Please test if the files are created in the right directories.

### DIFF
--- a/bin/build-site.sh
+++ b/bin/build-site.sh
@@ -101,7 +101,7 @@ node $TW5_BUILD_TIDDLYWIKI \
 	--version \
 	--load $TW5_BUILD_OUTPUT/build.tid \
 	--output $TW5_BUILD_OUTPUT \
-	--build favicon static index \
+	--build favicon static index archive \
 	|| exit 1
 
 # /empty.html					Empty
--- a/bin/build-site.sh
+++ b/bin/build-site.sh
@@ -96,13 +96,13 @@ echo -e -n "title: $:/build\ncommit: $TW5_BUILD_COMMIT\n\n$TW5_BUILD_DETAILS\n" 
 #
 ######################################################
 
-# /index.html				Main site
-# /favicon.ico				Favicon for main site
-# /static.html				Static rendering of default tiddlers
-# /alltiddlers.html			Static rendering of all tiddlers
-# /static/*					Static single tiddlers
-# /static/static.css		Static stylesheet
-# /static/favicon.ico		Favicon for static pages
+# /index.html			Main site
+# /favicon.ico			Favicon for main site
+# /static.html			Static rendering of default tiddlers
+# /alltiddlers.html		Static rendering of all tiddlers
+# /static/*				Static single tiddlers
+# /static/static.css	Static stylesheet
+# /static/favicon.ico	Favicon for static pages
 # 
 # Conditionally build archive if $TW5_BUILD_ARCHIVE is set to "archive"
 #
--- a/bin/build-site.sh
+++ b/bin/build-site.sh
@@ -46,6 +46,13 @@ if [  -z "$TW5_BUILD_OUTPUT" ]; then
     TW5_BUILD_OUTPUT=./output
 fi
 
+# Chech if archive should be built
+
+if [[ "$TW5_BUILD_OUTPUT" = "./output"* ]]; then
+    echo 'Archive will be built'
+	TW5_BUILD_ARCHIVE=archive
+fi
+
 mkdir -p $TW5_BUILD_OUTPUT
 
 if [  ! -d "$TW5_BUILD_OUTPUT" ]; then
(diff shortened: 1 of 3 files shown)


Jermolene:
Hi @pmario we use the `./bin/build-site.sh` script both to build the root site and the copy in `/prerelease`. Ideally, we would not create `/prerelease/archive` etc. That's a bit awkward, I think it might require an additional parameter to the script...

pmario:
>  I think it might require an additional parameter to the script...

As far as I can see, there are no parameters. There are only environment variables. IMO the important one to store the built content is `TW5_BUILD_OUTPUT` which points to the `./output` directory. 

So how do you change that when your create a release or prerelease

Jermolene:
> As far as I can see, there are no parameters. There are only environment variables. IMO the important one to store the built content is `TW5_BUILD_OUTPUT` which points to the `./output` directory.
> 
> So how do you change that when your create a release or prerelease

Apologies, we do indeed pass parameters via environment variables. The `TW5_BUILD_OUTPUT` variable is assigned within the GitHub actions script [here](https://github.com/Jermolene/TiddlyWiki5/blob/106e1c68df37b23b118d0c9effd0a395ca51210f/.github/workflows/ci.yml#L25).

The script could check the value of `TW5_BUILD_OUTPUT` to determine whether to build the archive index file, but I'd prefer to use a separate environment variable `TW5_BUILD_ARCHIVE` which accepts the relative path where the archive index file should be written.

pmario:
@Jermolene ... This is ready for review and testing, since you do plan a new major version. It works locally on my ubuntu 22.04 system. 

But I cannot test the CI version

bin/build-site.sh
```diff
@@ -46,6 +46,14 @@ if [  -z "$TW5_BUILD_OUTPUT" ]; then
     TW5_BUILD_OUTPUT=./output
 fi
 
+# Check if archive should be built depending on environment variable TW5_BUILD_OUTPUT
+# see: https://github.com/Jermolene/TiddlyWiki5/blob/4b56cb42983d4134715eb7fe7b083fdcc04980f0/.github/workflows/ci.yml#L62
+
+if [[ $TW5_BUILD_OUTPUT = "./output" ]]; then
```
Jermolene, line 52: I don't like the hardcoding of a special parameter value.  I'd prefer to use a separate environment variable TW5_BUILD_ARCHIVE which accepts the relative path where the archive index file should be written, and no archive would be written if the variable were not set, or was blank.

pmario:
@Jermolene -- This is ready for review again. If TW_BUILD_ARCHIVE is set to `./output` the archive will be created. 

Since `--build archive` also contains a static index.html with the version overview the `archive` subdirectory needs to be configured within tiddlywiki.info. So TW_BUILD_ARCHIVE and TW_BUILD_OUTPUT need to be the same directory.

Jermolene:
Hi @pmario that seems a little confusing to have to keep the two parameters in sync like that. I was expecting TW_BUILD_ARCHIVE would just be a flag telling the script whether or not to build the archive.

pmario:
@Jermolene -- ready for testing again. As you know, I cannot test the CI function. If you push the next update the prerelease should not create an archive version.

Jermolene:
Thank you @pmario and apologies for the delay.

saqimtiaz:
@pmario is this meant to be run via CI and to automatically update the archive? I do not see that an entry has been created for v5.3.2

pmario:
@saqimtiaz Yes, It should create the archive automatically. -- But as I wrote in the OP I cannot test it, since the build process does not run with my repo. 

IMO it can only be tested, if @Jermolene creates a new version, since the prerelease should not create an archive version

Jermolene:
Hi @pmario we've had two releases of new versions of TiddlyWiki since this PR was merged; wouldn't that have been enough to trigger your script?

pmario:
>..wouldn't that have been enough to trigger your script? 

I do not have a GitHub CI bot. My setup and directory structure is completely different to yours. I have no idea, how to test the configuration with your settings.

Jermolene:
Hi @pmario as I understand it, this PR contains an update to our CI script to make it save the archive version. You weren't able to test it for reasons that I understand, but what I don't understand is that we've now got to the point where the script has been run many times, including two releases. That makes me think that we should now be in a position to (a) determine whether the modifications in this PR worked and (b) to fix them if not.

pmario:
I found them: https://github.com/Jermolene/jermolene.github.io/tree/master/archive

The output path was wrong. I'll create a new PR soon. 

The index.html works and should be updated once we update the [TiddlyWiki Archive](https://tiddlywiki.com/#TiddlyWiki%20Archive) tiddler as a "docs only" update

pmario:
index.html and static.css need to stay in the archive directory as an index file.

PR #7783, opened 2023-10-13
More image type support

Gk0Wk:
## What I did

I found that the `<$image>` widget hardcoded how the src of the image tiddler was generated. If it's a `pdf` it uses `base64`, if it's a `svg` it's encoded using `encodeURIComponent`, if it's any other type it just defaults to `base64`. This makes it inconvenient to expand.

I've changed it to generate the src based on the encoding of the typeInfo corresponding to the type (`base64` or something else, default is `utf8`).

## Why

The purpose of this is to make tw more extensible to image formats.

For example, my [drawio plugin](https://talk.tiddlywiki.org/t/the-most-powerful-all-in-one-diagram-editor-for-uml-circuit-flowchart-etc/8099) customizes the `application/vnd.drawio` type. This format is essentially still `image/svg+xml` but I need to define it as a different type because I need to specify the use of a particular editor that way.

And at the moment such images can only be imported with transclude, not as `[img[]]`, because ImageWidget tries to treat them as base64 (missing an `encodeURIComponent` encoding) and thus can't display the image properly.

> You can look at this image, the src of the following image is `data:application/vnd.drawio;base64,<svg xmlns=...`
>
> image

### DIFF
--- a/core/modules/widgets/image.js
+++ b/core/modules/widgets/image.js
@@ -61,19 +61,19 @@ ImageWidget.prototype.render = function(parent,nextSibling) {
 				_canonical_uri = tiddler.fields._canonical_uri;
 			// If the tiddler has body text then it doesn't need to be lazily loaded
 			if(text) {
-                // Render the appropriate element for the image type by looking up the encoding in the content type info
-                var typeInfo = $tw.config.contentTypeInfo[type];
-                var encoding = typeInfo.encoding || "utf8";
-                if (encoding === "base64") {
-                    // .pdf .png .jpg etc.
-                    src = "data:" + type + ";base64," + text;
-                    if (type === "application/pdf") {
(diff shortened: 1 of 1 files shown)


pmario:
>For example, my [drawio plugin](https://talk.tiddlywiki.org/t/the-most-powerful-all-in-one-diagram-editor-for-uml-circuit-flowchart-etc/8099) customizes the application/vnd.drawio type. This format is essentially still image/svg+xml but I need to define it as a different type because I need to specify the use of a particular editor that way.

I think that's sensible. 

So +1 from my side.

Gk0Wk:
Get it! I have signed the CLA at the `tiddlewiki-com` branch.

Jermolene:
Thank you @Gk0Wk and welcome to the project

PR #7784, opened 2023-10-14
Support for list-template and list-empty widgets for specifying list widget templates

Jermolene:
Adds the option to specify the "emptyMessage" via a`<$list-empty>` widget within the list widget, instead of using the attribute, and similarly to use a `<$list-template>` widget to define the template.

Also adds an explicit "limit" attribute.

Cherry picked from #7710.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -60,6 +60,7 @@ ListWidget.prototype.render = function(parent,nextSibling) {
 Compute the internal state of the widget
 */
 ListWidget.prototype.execute = function() {
+	var self = this;
 	// Get our attributes
 	this.template = this.getAttribute("template");
 	this.editTemplate = this.getAttribute("editTemplate");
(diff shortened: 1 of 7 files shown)


AnthonyMuscio:
I have come here from the pre-release. the list-template widget is not really that, it is more a list-item-template. The body of the listwidget is also the item list template if you use that term.

I would suggest;
* either we rename list-template to list-item-template
* or at least document that is is acting as the "list item template" in the case of using the Explicit Templates for the [ListWidget](https://tiddlywiki.com/prerelease/#ListWidget)

Jermolene:
Thanks @AnthonyMuscio. You are right that "list item template" would be a more descriptive name. We've also got to balance the desire for brevity so we chose the shorter version.

AnthonyMuscio:
So please spell it out in the documentation lest this be lost in time and result in confusion.

If its a custom widget, the content of which is used to generate the list item do we need to even call it template? such as $list-item widget in keeping with $list-empty which could also be called a template.

PR #7785, opened 2023-10-14
Fix: Evernote .enex image import

### LINKED ISSUE #3176: evernote plugin does not import resources
I use evernote plugin to import a note (or notes) from evernote with embedded resources (images, pdf etc.). The plugin does import the note, but it does not import the resources.

linonetwo:
fixes #3176

Tested with 200M+ .enex file full of website clipped by Evernote WebClipper...All images in website are preserved.

截屏2023-10-15 00 09 12

A .enex that has wikihow clipped for test:

[wikihow.enex.zip](https://github.com/Jermolene/TiddlyWiki5/files/12906677/wikihow.enex.zip)

Only thing to note is that you need to use https://github.com/vzhd1701/evernote-backup/pull/54 to export .enex file, otherwise .enex that resources don't have hash can't be processed properly.


linonetwo:
TODO: Import 250M .enex on HTML wiki is smooth, but on NodeJS it quickly become 100% usage on server side, I need to figure out what is consuming CPU later.

But this PR is OK to merge.

Jermolene (commented):
Hi @linonetwo this looks good, apart from one minor issue. The use of a random number in the title seems clumsy. I'm not sure what options there might be, but a simple one would be if we could use a meaningful string (perhaps the truncated and possibly slugified text field), with a hash of the text field as a disambiguator.

linonetwo:
> random number in the title

I forgot I can use the hash provided by Evernote.

pmario:
>but on NodeJS it quickly become 100% usage on server side, I need to figure out what is consuming CPU later.

It's probably a Node.js file IO problem that causes problems. I did create 20000+ tiddlers from the debug console with a node-server active and it caused problems after about 10-20 seconds on the filesystem side. 

The tiddlers itself where created immediately. May be there is a problem in the node.js server backend if many files are written at the same time. may be memory leak or something similar.

linonetwo:
Yes, and ENEX import can be a good reproducible example for this performance issue. I will look into it one day maybe.

I don't have more things to add to this PR. I've done using it importing all my Evernote content. Now I have 26k tiddlers, with size 2GB+ (including .git backup, hope tiddlywiki nodejs won't look into .git ...)

linonetwo:
> probably a Node.js file IO problem

No, I find it's smooth with or without `"tiddlywiki/filesystem",` and `"tiddlywiki/tiddlyweb",`. On Official wiki.

It is just on my wiki. Maybe is cause by some plugin, but I don't have time to [bi-sect it](https://github.com/Jermolene/TiddlyWiki5/discussions/7588#discussioncomment-6597787) by hand.

Jermolene:
Thanks @linonetwo

Issue #7771, opened 2023-10-05
[IDEA] - display: grid for editor and preview


BurningTreeC:
**Is your feature request related to a problem? Please describe.**
I always thought the width of 49% for the editor and the preview is a strange workaround
I tried making the parent div of the toolbar, editor and preview `display: grid` using the fix-body-editor-refresh branch and that works well and we can display them side by side without gap

Jermolene:
Hi @BurningTreeC I don't think we need grid here, flexbox should be sufficient.

BurningTreeC:
Hi @Jermolene yes I tried flexbox and it looked promising. Then I went fullscreen with the editor like I do in the [codemirror 6 plugin](https://burningtreec.github.io/tiddlywiki-codemirror-6/) and tried making the editor and the preview fill the available space while the toolbar stays full width and height auto.

I couldn't manage that using flexbox - but using grid I could

Jermolene:
Hi @BurningTreeC OK, we'd need one flex container for the toolbar and the main body, and another one for dividing the body into the editor and the preview pane, which I see isn't going to fly with the current HTML structure.

Would you be interested in drafting a PR?

BurningTreeC:
@Jermolene

Do you mean a PR that's using grid?
If so, then yes

Jermolene:
Thanks @BurningTreeC yes.

PR #7787, opened 2023-10-15
display: grid for editor toolbar, editor and preview

BurningTreeC:
This PR makes the div surrounding the toolbar, editor and preview `display: grid`
It also adds the necessary styles so that editor and preview are displayed full width, not 49% as before
It also takes into account that the bitmap editor has variable width and height

see Idea #7771

### DIFF
--- a/core/ui/EditTemplate/body/default.tid
+++ b/core/ui/EditTemplate/body/default.tid
@@ -18,7 +18,7 @@ $:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
 	importState=<<qualify $:/state/ImportImage>> >
 <$dropzone importTitle=<<importTitle>> autoOpenOnImport="no" contentTypesFilter={{$:/config/Editor/ImportContentTypesFilter}} class="tc-dropzone-editor" enable={{{ [{$:/config/DragAndDrop/Enable}match[no]] :else[subfilter{$:/config/Editor/EnableImportFilter}then[yes]else[no]] }}} filesOnly="yes" actions=<<importFileActions>> >
 <div>
-<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]else[tc-no-tiddler-preview]] [[tc-grid]] +[join[ ]] }}}>
+<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]else[tc-tiddler-preview-hidden]] [[tc-tiddler-editor]] +[join[ ]] }}}>
 
 <$transclude tiddler="$:/core/ui/EditTemplate/body/editor" mode="inline"/>
 
--- a/core/ui/EditTemplate/body/default.tid
+++ b/core/ui/EditTemplate/body/default.tid
@@ -18,7 +18,7 @@ $:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
 	importState=<<qualify $:/state/ImportImage>> >
 <$dropzone importTitle=<<importTitle>> autoOpenOnImport="no" contentTypesFilter={{$:/config/Editor/ImportContentTypesFilter}} class="tc-dropzone-editor" enable={{{ [{$:/config/DragAndDrop/Enable}match[no]] :else[subfilter{$:/config/Editor/EnableImportFilter}then[yes]else[no]] }}} filesOnly="yes" actions=<<importFileActions>> >
 <div>
-<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]] +[join[ ]] }}}>
+<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]else[tc-no-tiddler-preview]] [[tc-grid]] +[join[ ]] }}}>
 
 <$transclude tiddler="$:/core/ui/EditTemplate/body/editor" mode="inline"/>
 
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1580,8 +1580,11 @@ html body.tc-body.tc-single-tiddler-window {
 	overflow: auto;
 }
 
-.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview {
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div {
 	display: grid;
+}
+
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview {
 	grid-template-areas: 
 	"toolbar toolbar"
 	"editor preview";
@@ -1590,7 +1593,6 @@ html body.tc-body.tc-single-tiddler-window {
 }
 
 .tc-tiddler-edit-frame .tc-dropzone-editor > div > div:not(.tc-tiddler-preview) {
-	display: grid;
 	grid-template-areas: 
 	"toolbar"
 	"editor";
@@ -1624,7 +1626,7 @@ html body.tc-body.tc-single-tiddler-window {
 
 .tc-tiddler-frame .tc-tiddler-preview canvas.tc-edit-bitmapeditor {
 	grid-area: editor;
-	width: 100%;
+	max-width: 100%;
 }
 
 .tc-edit-fields {
--- a/themes/tiddlywiki/vanilla/base.tid
+++ b/themes/tiddlywiki/vanilla/base.tid
@@ -1365,6 +1365,10 @@ html body.tc-body.tc-single-tiddler-window {
 	margin-top: 8px;
 }
 
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview > .tc-editor-toolbar {
+	grid-area: toolbar;
+}
+
 .tc-editor-toolbar button {
 	vertical-align: middle;
 	background-color: <<colour tiddler-controls-foreground>>;
@@ -1576,9 +1580,29 @@ html body.tc-body.tc-single-tiddler-window {
 	overflow: auto;
 }
 
-.tc-tiddler-preview-preview {
-	float: right;
-	width: 49%;
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview {
+	display: grid;
+	grid-template-areas: 
+	"toolbar toolbar"
+	"editor preview";
+	grid-template-columns: 1fr 1fr;
+	grid-template-rows: auto 1fr;
+}
+
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div:not(.tc-tiddler-preview) {
+	display: grid;
+	grid-template-areas: 
+	"toolbar"
+	"editor";
+	grid-template-columns: 1fr;
+	grid-template-rows: auto 1fr;
+}
+
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview > .tc-tiddler-preview-preview {
+	grid-area: preview;
+	width: calc(100% - 3px);
+	overflow-wrap: anywhere;
+	word-break: normal;
 	border: 1px solid <<colour tiddler-editor-border>>;
 	margin: 4px 0 3px 3px;
 	padding: 3px 3px 3px 3px;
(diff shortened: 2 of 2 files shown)


BurningTreeC:
Should there also be a fullscreen button for the editor?

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1365,6 +1365,10 @@ html body.tc-body.tc-single-tiddler-window {
 	margin-top: 8px;
 }
 
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div > .tc-editor-toolbar {
```
pmario, line 1368: selectors like this one are very brittle if the DOM structure is changed. And they will cause backwards compatibility problems. So if someone needs to fix a bug in the DOM structure. This CSS will make it backwards incompatible, so it will be rejected

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1576,9 +1580,31 @@ html body.tc-body.tc-single-tiddler-window {
...
+
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div:not(.tc-tiddler-preview) {
+	grid-template-areas: 
+	"toolbar"
+	"editor";
+	grid-template-columns: 1fr;
+	grid-template-rows: auto 1fr;
+}
+
+.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview > .tc-tiddler-preview-preview {
```
pmario, line 1603: I do have a problem with all those CSS rulses. IMO they will fail in the future

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: Thanks @pmario 

yes if the DOM structure changes, these rules need to be adapted

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: you mean if the DOM structure changes?

BurningTreeC:
Hello @pmario - I created an alternative PR here: #7788

themes/tiddlywiki/vanilla/base.tid
pmario: IMO go to the DIV of the template and add a new class eg: `tc-grid` to that element. Then `.tc-tiddler-frame .tc-grid` should be enough to set the CSS reliably. -- No js code and no brittle configurations needed.

BurningTreeC: Hi @pmario, I've done that now

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: I'm now using not so brittle CSS selectors. Could you have a look @pmario ?

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1619,12 +1614,12 @@ html body.tc-body.tc-single-tiddler-window {
...
 """>>
 
-.tc-tiddler-edit-frame .tc-dropzone-editor > div > div > .tc-edit-texteditor {
+.tc-tiddler-frame .tc-grid.tc-tiddler-preview .tc-edit-texteditor, .tc-tiddler-frame .tc-grid.tc-no-tiddler-preview .tc-edit-texteditor {
 	grid-area: editor;
 	width: 100%;
 }
 
-.tc-tiddler-frame .tc-tiddler-preview canvas.tc-edit-bitmapeditor {
+.tc-tiddler-frame .tc-grid.tc-tiddler-preview canvas.tc-edit-bitmapeditor, .tc-tiddler-frame .tc-grid.tc-no-tiddler-preview canvas.tc-edit-bitmapeditor {
```
pmario, line 1621: also 2 lines here and the one above

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1365,7 +1365,7 @@ html body.tc-body.tc-single-tiddler-window {
 	margin-top: 8px;
 }
 
-.tc-tiddler-edit-frame .tc-dropzone-editor > div > div > .tc-editor-toolbar {
+.tc-tiddler-frame .tc-grid.tc-tiddler-preview .tc-editor-toolbar, .tc-tiddler-frame .tc-grid.tc-no-tiddler-preview .tc-editor-toolbar {
```
pmario, line 1368: please use 2 lines for better maintainability.

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1580,29 +1580,24 @@ html body.tc-body.tc-single-tiddler-window {
...
+.tc-grid {
 	display: grid;
 }
 
-.tc-tiddler-edit-frame .tc-dropzone-editor > div > div.tc-tiddler-preview {
-	grid-template-areas: 
-	"toolbar toolbar"
-	"editor preview";
+.tc-tiddler-frame .tc-grid.tc-tiddler-preview {
+	grid-template-areas: "toolbar toolbar" "editor preview";
```
pmario, line 1588: I do like `grid-template-area` --  I think for readability reasons we should use: 

```
.tc-tiddler-frame .tc-grid.tc-tiddler-preview {
	grid-template-areas: 
		"toolbar toolbar"
		"editor preview";
..
```

@Jermolene -- I know we normally do not use formatting like this, but I think it makes the "intent" with `grid-template-areas` so much easier to understand and maintain. -- What do you think?

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1580,29 +1580,24 @@ html body.tc-body.tc-single-tiddler-window {
...
 	grid-template-columns: 1fr 1fr;
 	grid-template-rows: auto 1fr;
 }
 
-.tc-tiddler-edit-frame .tc-dropzone-editor > div > div:not(.tc-tiddler-preview) {
-	grid-template-areas: 
-	"toolbar"
-	"editor";
+.tc-tiddler-frame .tc-grid.tc-no-tiddler-preview {
+	grid-template-areas: "toolbar" "editor";
```
pmario, line 1594: also here. IMO it's much easier to understand

```
	grid-template-areas:
		"toolbar"
		"editor";
```

pmario (commented):
I think that's much better now.

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: I cleaned that up

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: I cleaned that up

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: i made that more readable like you proposed

themes/tiddlywiki/vanilla/base.tid
BurningTreeC: I made that more readable, too

BurningTreeC:
Hi @Jermolene @pmario - I think this is ready for review

pmario:
I think, we also needs some documentation, that tells users about the new CSS `tc-grid` classes and the grid areas. 
May be we add this to the https://tiddlywiki.com/dev/ edition. 

It should make it easier for plugin authors to deal with the new grid areas

pmario:
@Jermolene -- It looks good to me now

core/ui/EditTemplate/body/default.tid
```diff
@@ -18,7 +18,7 @@ $:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
 	importState=<<qualify $:/state/ImportImage>> >
 <$dropzone importTitle=<<importTitle>> autoOpenOnImport="no" contentTypesFilter={{$:/config/Editor/ImportContentTypesFilter}} class="tc-dropzone-editor" enable={{{ [{$:/config/DragAndDrop/Enable}match[no]] :else[subfilter{$:/config/Editor/EnableImportFilter}then[yes]else[no]] }}} filesOnly="yes" actions=<<importFileActions>> >
 <div>
-<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]] +[join[ ]] }}}>
+<div class={{{ [function[edit-preview-state]match[yes]then[tc-tiddler-preview]else[tc-no-tiddler-preview]] [[tc-grid]] +[join[ ]] }}}>
```
Jermolene, line 21: We try to use semantic class names, and avoid functional names such as "tc-grid". I think it would be clearer to replace `tc-no-tiddler-preview` with `tc-tiddler-preview-hidden`, and to use `tc-tiddler-preview-frame` instead of `tc-grid`.

Jermolene (commented):
Thanks @BurningTreeC @pmario

core/ui/EditTemplate/body/default.tid
pmario: I think `tc-tiddler-preview-frame` is way to specific. `tc-grid` only contains 1 setting to use grid. It can be reused that way. 

`tc-tiddler-preview-frame` is completely misleading since it is not set to the preview-frame, but it's parent

pmario: If I want to define `tc-grid` to any other gird with a completely different context in the sidebar or control-panel `tc-tiddler-preview-frame` would be misleading. IMO it should be as generic as possible

Jermolene: @pmario if we use `tc-grid` and reuse it elsewhere then we'll be constrained that we can't change the styling of just the preview pane. As I said before, functional class names are an anti-pattern.

pmario:
>@pmario if we use tc-grid and reuse it elsewhere then we'll be constrained that we can't change the styling of just the preview pane.

That's not true. Have a closer look at the CSS code. 

`tc-grid` activates `display: grid` only. It may be used for some global styling for grids but I assume that this one will never be changed again. 

`tc-tiddler-preview`, `tc-tiddler-no-preview` and `tc-tiddler-preview-preview` and may be others are the classes that are responsible for the editor and preview definitions. 

IMO `tc-grid` is a utility calss, which defines a major usage pattern. It is a functional class name because there is no other good name for "grid".  

We could use `tc-framework` but that would be misleading. tc-framework could be grid or flexbox. But those 2 definitions are **not** interchangeable. So it's needed to stay with tc-grid or a tc-flex if you started with it. They are not compatible.

Jermolene:
Hi @pmario our CSS classes are designed not just to accomplish a particular visual layout but to provide a surface area for customisations. If we use a single `tc-grid` class then it becomes impossible for others to customise the layout of the preview pane without also affecting the other places that use `tc-grid`.

BurningTreeC:
Hi @pmario @Jermolene 

I went with `tc-tiddler-editor` and `tc-tiddler-preview-hidden`

Should be good now 👍

Jermolene:
Terrific, thank you @BurningTreeC

oeyoews:
This seems to conflict with the flexbox layout of cm6 (maybe), but cm5 does not have this situation. @Jermolene  @BurningTreeC 


https://github.com/Jermolene/TiddlyWiki5/assets/72405338/dfbe4e69-4e3e-43c5-bdcc-497c9333ce7d

Jermolene:
Hi @oeyoews are you saying that the new CSS grid layout of the editor pane interferes with CodeMirror 6? It's hard to see how that could happen if CM6 adapts correctly to the size of its container.

oeyoews:
The reason is really not easy to investigate, and I don’t have any clue. What is certain is that if the width of the element is specified, it will cause this problem, unless width: 100% is specified. @Jermolene 
https://github.com/Jermolene/TiddlyWiki5/assets/72405338/a2e58dd5-1bef-488c-9726-2e74c37db486

oeyoews:
No, it doesn't work

oeyoews:
image



After looking at the usage of grid-template, this should fix the problem, but what I don't understand is why just cm6 would cause this problem, I'm having a hard time finding what the difference is. @Jermolene 

```css
.tc-tiddler-preview {
grid-template-columns:  minmax(0px, 1fr) minmax(0px, 1fr) !important;
}
```

oeyoews:
https://github.com/Jermolene/TiddlyWiki5/pull/7922

pmario:
no `!important` is toxic. It will mess up all our stylesheets.

PR #7788, opened 2023-10-15
display: grid for toolbar, editor and preview - using inline styles

BurningTreeC:
This PR is an alternative to #7787 

it adds the necessary styles directly to the created dom nodes


pmario:
Sorry, but IMO that's even worse. "style" can not be overwritten by standard CSS. So users have no chance to ever change those settings again. See my comment on the other PR. https://github.com/Jermolene/TiddlyWiki5/pull/7787

Jermolene:
Hi @BurningTreeC I think @pmario is correct: we need to find a way to move all those CSS property assignments into stylesheets.

Presumably the key to being able to do that is to assign classes to the DOM nodes that you need to target?

BurningTreeC:
closing this one ^^

PR #7789, opened 2023-10-15
Update Basics.tid to not use fixed height in edit widgets

BurningTreeC:
The `Basics` tab in the ControlPanel uses two `edit` widgets which get the fixed height of the tiddler editor if there's such a fixed height set

This PR makes them height auto for all cases


pmario:
Please be more specific with your description. There is no obvious difference between tiddlywiki.com and the auto-created preview wiki.

So what does need to be changed and how to see a difference between tw-com and the preview.

Jermolene:
Thanks @BurningTreeC

@pmario the bug is that at present if one changes the editor height setting to fixed height then the same setting is applied to the text editors in the control panel.

pmario:
I forgot about, how to set the fixed height, because I never use it. So a link to: https://tiddlywiki.com/#Height%20of%20text%20editor would probably have been enough to make it clear.

Issue #7798, opened 2023-10-21
[BUG]  Focus abnormality


oeyoews:
### Describe the bug

when set autofocus to text, if input on type field, this focus alway focus on editor 

### To Reproduce

1. Go to [tiddlywiki autofocus](https://tiddlywiki.com/#%24%3A%2Fconfig%2FAutoFocus)
2. change title to text
3. create new tiddler, and input on type field
4. this focus alway focus the editor on typing

### TiddlyWiki Configuration

- Version [e.g. v5.1.24]
- Saving mechanism [e.g. Node.js, TiddlyDesktop, TiddlyHost etc]
- Plugins installed [e.g. Freelinks, TiddlyMap]

pmario:
IMO related to:  [IDEA] it is proposed to add a "focusSelect" parameter to the edit-text widget #7221 
and: Add focusSelect attribute to <$edit-text> widget #7222 

Check those issues and the related Talk threads. May be there is something useful.

Jermolene:
Thank you @oeyoews.

It occurs to me that this behaviour may be intentional - @BurningTreeC would you kindly be able to investigate?

pmario:
I did just replicate the problem. If I want to write a custom text into the type field, the focus jumps after typing the first character. If the type is selected again. The focus does not jump anymore. 

So if an existing element from the dropdown is selected, it may be intentional. 

But if the user wants to manually input a type-info, jumping after the first character seems to be a bug.

BurningTreeC:
I'll investigate

Jermolene:
Thank you @pmario @BurningTreeC much appreciated.

BurningTreeC:
Ok, the editor re-renders and sets auto-focus when it's expected
But it's confusing behavior
The intended behavior is to set auto-focus when the tiddler is opened for editing
I think we can do something here

BurningTreeC:
I've created a Draft PR that addesses this issue: #7800

BurningTreeC:
Let's name the problem, to understand it better.
The editor type changes and the editor gets redrawn and gets focus because of that redraw.
That steals the focus of the type input field.
We need a little mechanism to tell if an element has focus which should not get "stolen".
Maybe we add the class `tc-keep-focus` to the type input field and test in factory.js if the activeElement has that class.

BurningTreeC:
I've now added PR #7805 that should fix the issue correctly

oeyoews:
[...]

PR #7800, opened 2023-10-21
Update body-editor.tid not to set autofocus when ...

BurningTreeC:
... the text in the input field doesn't match an EditorTypeMapping

This PR is a try to address #7798

### DIFF
--- a/core/ui/EditTemplate/body-editor.tid
+++ b/core/ui/EditTemplate/body-editor.tid
@@ -1,5 +1,8 @@
 title: $:/core/ui/EditTemplate/body/editor
 
+\function type-match()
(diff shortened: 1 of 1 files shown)


pmario:
I did test it and it seems the "auto-focus" text field does not work anymore, if the editor is switched from view-mode to edit-mode.

BurningTreeC:
> The cursor does not automatically focus on the search box when the website is opened

I cannot reproduce that. It must be an issue not related to this PR

BurningTreeC:
> I did test it and it seems the "auto-focus" text field does not work anymore, if the editor is switched from view-mode to edit-mode.

You mean if you open a tiddler for editing? Then it should work fine...

BurningTreeC:
I'm closing this PR as I've seen that there are downsides that are unresolved

PR #7802, opened 2023-10-21
Tweak TOC to not show expander for sublist with all children excluded

CrossEye:
This fixes a minor bug [described in talk][ttw] like this:

> I’ve noticed what looks like a (minor) bug when using exclude with <<toc-selective-expandable ... >> If there are tiddlers with tags matching a node in the tree, but all of them are excluded by the filter, we still show the right arrow. Clicking it, gives you the down arrow, but no children.

All credits for the fix should go to Eric Shulman.

  [ttw]: https://talk.tiddlywiki.org/t/8162


Jermolene:
Thanks @CrossEye

PR #7803, opened 2023-10-22
Use the class `tc-keep-focus` to detect in factory.js if the activeElement should keep focus

BurningTreeC:
This PR adds the class `tc-keep-focus` to the type input field
In `factory.js` it detects if the activeElement has that class and doesn't set focus in that case

This PR addresses #7798


BurningTreeC:
Closing in favor of #7805

PR #7805, opened 2023-10-22
Fix type and editor fields autofocus: Test in factory.js if the activeElement should keep focus

BurningTreeC:
This PR tests in factory.js if there's the activeElement with the class `tc-keep-focus`
If so, the editor doesn't set autofocus to itself


BurningTreeC:
I don't know why Vercel doesn't deploy the Preview...

Jermolene:
Hi @BurningTreeC the Vercel preview is failing with an error:

image

BurningTreeC:
Ah OK, I'll fix this 👍

Jermolene:
Hi @BurningTreeC the error from the last run was this:

image

Issue #7801, opened 2023-10-21
[BUG] Field editor incorrectly assumes focus on delete of field


CodaCodr:
### Describe the bug

1. Tiddler in edit mode with many fields (in my case a dozen or more)
2. Aim to delete half of them.
3. Click the trashcan icon against field,
4. Focus leaps to the Add field editor.
5. Scroll back to the next field to be deleted.
6. goto 3.

Not the best UX.

### Expected behavior

No focus change on delete of field.

### TiddlyWiki Configuration

- Version 5.3x
- OS: Win10
- Browser FF latest

saqimtiaz:
Fixed in #7806

PR #7806, opened 2023-10-22
Don't set focus on field-name input field when deleting field using delete-field button

BurningTreeC:
This PR removes the functionality that, when clicking the delete-field button, the focus gets set to the next field-name input field.
This addresses #7801 
This PR does NOT remove the functionality when deleting the field using the keyboard shortcut `((delete-field))`
There needs to be a consensus on which field to set the focus when using the keyboard shortcut.
I believe, when using the keyboard shortcut, focus should be set somewhere...


Jermolene:
Thanks @BurningTreeC

PR #7809, opened 2023-10-24
Fix last filter operator when zero items selected

### LINKED ISSUE #7507: [BUG] Behavior of the last filter when zero items requested
### Describe the bug

When using a variable to decide how many list items the last filter operator should return, it can happen that the number calculated is zero. In this case the last operator returns the entire list not no items as would seem to be the correct action. 

(As expected !limit<n> for n=0 returns the whole list, since it is stated to be a synonym for last. Note though that it is not strictly a synonym since last[N] defaults to N=1 and !limit[N] defaults to N=0 according to tiddlywiki.com. !limit[] returns the entire list not no items as one would assume from the statement about the default value being 0 or the last item if it was fully synonymous with last.)

### Expected behavior

When n=0 last<n> should return nothing. It would make the most sense when one is requesting some number of list items that if the number is zero, then nothing is returned. Note that this would match the behavior of first<n> when n=0 which returns nothing. It would also match the behavior of limit[N] which should default to N=0 and limit[] returns nothing.

### To Reproduce

Create and display a tiddler with this wiki text-

NB hardcoded 0 was just done for completeness

\define zero() 0

\define theList()  A B C D E F G H I J

!Last:

nothing: {{{ [enlist<theList>last[]] }}}

<<zero>>: {{{ [enlist<theList>last<zero>] }}}

hardcoded 0: {{{ [enlist<theList>last[0]] }}}

!First:

2: {{{ [enlist<theList>first[2]] }}}

nothing: {{{ [enlist<theList>first[]] }}}

<<zero>>: {{{ [enlist<theList>first<zero>] }}}

hardcoded 0: {{{ [enlist<theList>first[0]] }}}

!"""!"""limit:

nothing: {{{ [enlist<theList>!limit[]] }}}

<<zero>>: {{{ [enlist<theList>!limit<zero>] }}}

hardcoded 0: {{{ [enlist<theList>!limit[0]] }}}

!limit:

nothing: {{{ [enlist<theList>limit[]] }}}

<<zero>>: {{{ [enlist<theList>limit<zero>] }}}

hardcoded 0: {{{ [enlist<theList>limit[0]] }}}

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version  v5.1.27
- Saving mechanism TiddlyDesktop
- Plugins installed none

- OS: MacOS 13.4
- Browser TiddlyDesktop
- Version 0.0.14

### Additional context

_No response_

rmunn:
Previously, `[last[0]]` was incorrectly returning the entire list. It now returns zero items as it should. A unit test verifying the bug fix is also included.

As a small bonus, it also skips gathering the results list if it knows it's just going to throw away the entire thing. This makes the `[last[0]]` case more efficient.

Fixes #7507.

### DIFF
--- a/core/modules/filters/listops.js
+++ b/core/modules/filters/listops.js
@@ -58,6 +58,7 @@ Last entry/entries in list
 exports.last = function(source,operator,options) {
 	var count = $tw.utils.getInt(operator.operand,1),
 		results = [];
+	if(count === 0) return results;
 	source(function(tiddler,title) {
 		results.push(title);
 	});
--- a/editions/test/tiddlers/tests/test-filters.js
+++ b/editions/test/tiddlers/tests/test-filters.js
@@ -365,6 +365,7 @@ Tests the filtering mechanism.
 			expect(wiki.filterTiddlers("[sort[title]first[8]]").join(",")).toBe("$:/ShadowPlugin,$:/TiddlerTwo,a fourth tiddler,filter regexp test,has filter,hasList,one,Tiddler Three");
 			expect(wiki.filterTiddlers("[sort[title]first[x]]").join(",")).toBe("$:/ShadowPlugin");
 			expect(wiki.filterTiddlers("[sort[title]last[]]").join(",")).toBe("TiddlerOne");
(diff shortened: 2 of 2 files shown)


Jermolene:
Thanks @rmunn

PR #7810, opened 2023-10-24
Fix list widget when widget body used as template while `$list-empty` specified

### LINKED ISSUE #7804: [BUG] List template syntax inconsistent
### Describe the bug

These two list widget invocations should be equivalent:

```plaintext
<$list filter="[all[tiddlers]sort[title]]" limit="3" emptyMessage="nothing">
Title: {{!!title}}<br/>
</$list>

<$list filter="[all[tiddlers]sort[title]]" limit="3">
Title: {{!!title}}<br/>
<$list-empty>nothing</$list-empty>
</$list>
```

However, switching from `emptyMessage` to `<$list-empty>` has the side effect of discarding the list template, because `<$list-empty>` has the (undocumented) effect of ignoring the list widget's body. So the second invocation ends up using the default template, which doesn't have the text "Title: " or the `<br/>` in it.

### Expected behavior

Both of those widget invocations would produce the same result. When `<$list-empty>` is present without a `template` attribute or a `<$list-template>` block, the contents of the `<$list>` widget (with the `<$list-empty>` section removed) should continue to be used as the template, just as they were in 5.3.1, for backwards compatibility reasons.

### To Reproduce

2. Create a new tiddler with the following content:

```plaintext
<$list filter="[all[tiddlers]sort[title]]" limit="3" emptyMessage="nothing">
Title: {{!!title}}<br/>
</$list>

<$list filter="[all[tiddlers]sort[title]]" limit="3">
Title: {{!!title}}<br/>
<$list-empty>nothing</$list-empty>
</$list>
```

3. Verify that the two lists are rendered with different templates, rather than the same template as they should have been.

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: v5.3.2-prerelease
- Saving mechanism: None
- Plugins installed: None

- OS: Linux
- Browser: Firefox
- Version: any

### Additional context

Technically this doesn't break backwards compatibility, as no existing wikis should have a `<$list-empty>` widget invocation in a list template. So this could instead be solved by documenting that if you use `<$list-empty>` you must also use `<$list-template>`. However, I believe that user expectation will be that `emptyMessage` and `<$list-empty>` will behave exactly alike (except, of course, that it's far easier to put wiki syntax into a `<$list-empty>` section than an `emptyMessage` attribute), so I believe the current behavior is a bug.

rmunn:
If `<$list-empty>` is specified but `<$list-template>` is not, and the contents of the list widget are used as the template instead, then the result should be the same as if the emptyMessage attribute had been used.

Fixes #7804.

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -114,9 +114,21 @@ ListWidget.prototype.findExplicitTemplates = function() {
 			if(node.type === "list-template" || node.type === "list-empty") {
 				// Do nothing
 			} else if(node.type === "element" && node.tag === "p") {
-				var nodeWithoutTemplates = $tw.utils.extend({}, node);
-				nodeWithoutTemplates.children = removeTemplates(node.children);
-				result.push(nodeWithoutTemplates);
+				var childrenWithoutTemplates = removeTemplates(node.children);
+				if(node.children && node.children.length === childrenWithoutTemplates.length) {
+					// No changes
+					result.push(node);
+				} else {
+					// Special case: if the only child of the <p> tag was an explicit template, we remove the <p> tag as well
+					// This avoids polluting the output with empty <p></p> elements which can mess up formatting
+					if(childrenWithoutTemplates.length === 0 && node.children && node.children.length !== 0) {
+						// Do nothing
+					} else {
+						var nodeWithoutTemplates = $tw.utils.extend({}, node);
+						nodeWithoutTemplates.children = childrenWithoutTemplates;
+						result.push(nodeWithoutTemplates);
+					}
+				}
 			} else {
 				result.push(node);
 			}
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -28,6 +28,18 @@ Inherit from the base widget class
 */
 ListWidget.prototype = new Widget();
 
+ListWidget.prototype.initialise = function(parseTreeNode,options) {
+	// Bail if parseTreeNode is undefined, meaning that the ListWidget constructor was called without any arguments so that it can be subclassed
+	if(parseTreeNode === undefined) {
+		return;
+	}
+	// First call parent constructor to set everything else up
+	Widget.prototype.initialise.call(this,parseTreeNode,options);
+	// Now look for <$list-template> and <$list-empty> widgets as immediate child widgets
+	// This is safe to do during initialization because parse trees never change after creation
+	this.findExplicitTemplates();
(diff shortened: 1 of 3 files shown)


pmario:
I think there is a problem with P-tag creation. IMO the code contains way too much handling for P-tag edge-cases. 

All the special handling in this PR is fighting symptoms and does not fix the underlaying problem the TW syntax has with redundant P-tag creation.

rmunn:
@Jermolene - I would appreciate a review of this PR, because it's blocking https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1778413275. That is, I could work further on #7694, adding a `<$list-join>` pseudo-widget ([as you suggested](https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1708174802)). But the code I would have to write to add it would be so much simpler after this PR is merged that I'd prefer to wait; if I do the work now against current `master` I'll end up dealing with merge conflicts later. So if you need me to change anything about this bugfix PR, please let me know and I'll work on it right away. Once this is merged then I'll be able to finish #7694 and mark it ready for review as well.

rmunn:
@pmario - You're right about the `<p>` tag being a problem. However, I wasn't about to make large structural changes to the TW parser in a PR that fixes a single corner-case bug. The `<p>` tag stuff you mentioned is the result of https://github.com/Jermolene/TiddlyWiki5/pull/7710#issuecomment-1711734198, where if the list widget is in block mode then it wasn't picking up on the `$list-template` and `$list-empty` children (because, indeed, of that extra `<p>` tag). So my PR has to deal with that as well.

So yes, TW's handling of paragraph tags could use some looking at. I don't have time to do it at the moment, though.

pmario:
>So yes, TW's handling of paragraph tags could use some looking at. I don't have time to do it at the moment, though.

I think it's about 5 lines in 1 file to fix the problem [core/modules/parsers/wikiparser/wikiparser.js](https://github.com/Jermolene/TiddlyWiki5/pull/7061/files#diff-5402436d2e258350a2a1e102d8ae8176ef556720666bffdf64ebc2373103b6b5). 

It fixes a lot of problems with the TW DOM structure, but it also makes some problems visible that have been covered by erroneous P-tags. The linked PR contains 112 test tiddlers where we need to have a closer look.

rmunn:
Now that #7827 is working, closing this one in favor of #7827 which @Jermolene prefers. Will reopen if @Jermolene asks me to, of course.

PR #7814, opened 2023-10-26
Fix window.btoa call in browser

rmunn:
Fixes https://github.com/Jermolene/TiddlyWiki5/pull/7683#issuecomment-1778762929.


rmunn:
https://tiddlywiki5-git-fork-rmunn-bugfix-window-btoa-jermolene.vercel.app/test.html shows all green, including the two tests that fail on https://tiddlywiki.com/prerelease/test.html.

Jermolene:
Thanks @rmunn

PR #7816, opened 2023-10-26
Don't steal focus to the type input field if autofocus is set to "text"

BurningTreeC:
Another try making this PR happen


BurningTreeC:
yay!

BurningTreeC:
Hello @Jermolene

This PR now builds fine
Are there occasions where we want autofocus if we're not in a browser?

Jermolene:
Hi @BurningTreeC I'm concerned that this approach is not backwards compatible. Would it be possible to use a dynamic value for the autofocus attribute in this case?

BurningTreeC:
Hello @Jermolene - what do you mean by dynamic value for the autofocus attribute?

Jermolene:
Hi @BurningTreeC it looks like this PR is attempting to apple an additional condition governing the setting of autofocus. I wonder if that can be done in wikitext instead.

PR #7817, opened 2023-10-26
Add <$testcase> widget

Jermolene:
This PR adds a new `<$testcase>` widget to try to solve a problem with the examples that we feature in the documentation. The existing macros are workable for simple, self-contained examples, but can be extremely hard to follow in cases where the examples are also using additional tiddlers. The `<$testcase>` widget solves that problem by displaying interactive examples that show the output together with a tabbed display of the constituent tiddlers that produce it:

image

Try out the [Vercel preview](https://tiddlywiki5-git-testcase-widget-jermolene.vercel.app/#TestCaseWidget:TestCaseWidget%20DataWidget)

The testcase widget is essentially a lightweight version of the existing `<$innerwiki>` plugin. It constructs an entirely separate wiki store object, fills it with the required tiddlers, and then renders a template that by default provides the tabbed view of the source tiddlers alongside the output.

The example above was generated with:

```
<$testcase>
<$data title="Description" text="How to calculate 2 plus 2"/>
<$data title="FirstNumber" text="2"/>
<$data title="SecondNumber" text="2"/>
<$data title="Output" text="<$text text={{{ [{FirstNumber}add{SecondNumber}] }}}/>"/>
</$testcase>
```

The `<$data>` widget that originated in the `<$innerwiki>` plugin is moved into the core and extended with support for reading tiddlers from "compound tiddlers", which are the special format used by the test suite for embedding multiple tiddlers within a single tiddler. [Here is an example](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/data/conditionals/Basic.tid).

Our test cases are a useful form of documentation. The ultimate goal of this work is to be able to include our wikitext tests suite within the main documentation of TiddlyWiki.  Advanced users wondering how `\function` works could dig down into potentially dozens of test cases to explore the behaviour of edge cases.

Progress:

* [ ] Update some of the existing tw5.com examples to use the `<$testcase>` widget

This PR was cherry picked from #7406.


core/ui/TestCases/DefaultTemplate.tid
```diff
@@ -0,0 +1,22 @@
+title: $:/core/ui/testcases/DefaultTemplate
+
+\whitespace trim
```
pmario, line 3: There is whitespace trim, but no indentation. IMO please use indentation

core/ui/TestCases/DefaultTemplateSource.tid
```diff
@@ -0,0 +1,19 @@
+title: $:/core/ui/testcases/DefaultTemplate/Source
```
pmario, line 1: there is indentation but no `\whitespace trim`

pmario:
Looks good. Will need to run some tests.

Jermolene:
Thanks @pmario I've pushed an update to fix those issues

Jermolene:
Thanks @CodaCodr I've just pushed a fix.

CodaCodr:
Look Mum! No ActionWidgets!

image

TheophileDesmedt:
CSS affect tiddlers outside of the testcase, is that a bug? 

image

Jermolene:
> CSS affect tiddlers outside of the testcase, is that intended?

The testcase widget shares the same rendering context as the main wiki and so CSS and other globla DOM properties are shared. In such cases the innerwiki widget can be used to create an entirely separate embedded wiki.

The purpose of the testcase widget is to be a lighter weight and more performant version of the innerwiki widget. It wouldn't be practical to use the innerwiki widget for hundreds of documentation examples; each of those iframes would overwhelm a mobile device.

The documentation needs to clarify the limitations and constraints of the testcase widget.

editions/tw5.com/tiddlers/widgets/DataWidget.tid
```diff
@@ -0,0 +1,65 @@
...
+
+The content of the data widget is rendered as if the data widget were not present. It supports the following attributes:
+
+|!Attribute |!Description |
+|<<.attr $tiddler>> |Optional title of a tiddler to be used as a payload tiddler (optional) |
+|<<.attr $filter>> |Optional filter string identifying tiddlers to be used as payload tiddlers (optional) |
+|<<.attr $compound-tiddler>> |Optional title of a tiddler containing payload tiddlers in `text/vnd.tiddlywiki-multiple` format (see below) |
+|//any attribute<br>not starting<br>with $// |Field values to be assigned to the payload tiddler(s) |
+
+The data widget is not rendered when used within the <<.wlink TestCaseWidget>> widget or the [[Innerwiki Plugin]] but for ease of testing, when used elsewhere it renders a JSON representation of the payload tiddlers.
```
btheado, line 22: > when used elsewhere it renders a JSON representation of the payload tiddlers

I'm not able to get this behavior to work. Outside of the testcase widget, I find the data widget to render empty.

editions/tw5.com/tiddlers/widgets/TestCaseWidget.tid
```diff
@@ -0,0 +1,46 @@
...
+! Introduction
+
+The testcase widget creates an independent subwiki loaded with the specified payload tiddlers and then renders a specified template from within the subwiki.The default template displays a split view with the source tiddlers on the left and the rendered tiddler titled `Output` on the right. It also displays the tiddler titled `Description` as the heading. This makes it possible to run independent tests that also serve as documentation examples.
+
+! Content and Attributes
+
+The content of the `<$testcase>` widget is not displayed but instead is scanned for <<.wlink DataWidget>> widgets that define the payload tiddlers to be included in the test case. The `$:/core` plugin is automatically included in the payload.
+
+|!Attribute |!Description |
+|<<.attr template>> |Optional title of the template used to display the testcase (defaults to $:/core/ui/testcases/DefaultTemplate). Note that custom templates will need to be explicitly added to the payload |
```
btheado, line 17: Mention the other core provided template `$:/core/ui/testcases/RawJSONTemplate` as being useful for troubleshooting `$data` widgets?

core/ui/TestCases/DefaultTemplate.tid
```diff
@@ -0,0 +1,22 @@
...
+		<div class="tc-testcase-header">
+			<h2><$transclude tiddler="Description" mode="inline"/></h2>
+		</div>
+		<div class="tc-testcase-panes">
+			<div class="tc-testcase-source">
+				<$macrocall $name="tabs" tabsList="[all[tiddlers]sort[]] -[prefix<state>] -Description -ExpectedResult -Output Output +[putfirst[]] -[has[plugin-type]]" state=<<state>> default="Output" template="$:/core/ui/testcases/DefaultTemplate/Source"/>
+			</div>
+			<div class="tc-testcase-divider">
+			</div>
+			<div class="tc-testcase-output">
```
btheado, line 17: It be useful to add a linkcatcher widget here and select the tab corresponding to the clicked link?

editions/tw5.com/tiddlers/widgets/DataWidget.tid
Jermolene: Thanks @btheado I had replaced that feature of the data widget with the ability for the testcase widget to display its payload in JSON via a special template. I've pushed an update with the docs in 0db4c449397c0b30498c5e052e9fcb80bed90541 that explains how to do that. I've also added the JSON output feature back to the data widget in 1d89c7925aa8392b450e7ef36c04fc3f360f9173 with tests in 0db4c449397c0b30498c5e052e9fcb80bed90541

core/ui/TestCases/DefaultTemplate.tid
Jermolene: Thank you @btheado I've belatedly implemented your suggestion in Thank you @btheado in 648855e8a50b1ee3bef3120b64b5713b69190c9b

btheado: Thanks. Works great!

btheado:
@Jermolene, nice improvements! I like that test case tiddler titles are links to open the tiddler separately. Very useful. I also like the new Narrative feature.

The test case tiddlers and test case widget seem like a really good replacement for the `wikitext-example-without-html` macro. Having the widget examples be live-editable will be a great improvement. One nice feature of the wikitext-example macro is the "copy to clipboard" button. Have any thoughts about the equivalent for the test case widget?

Being able to easily copy the text of the example is useful. In addition or instead, maybe an easy way to get the json for all the tiddlers in the test case widget?

Tagging @AnthonyMuscio as I've seen him advocate for more use of "copy to clipboard" in the past.

btheado:
Originally I was thinking all the uses of `wikitext-example-without-html` would be converted to test case tiddlers. Maybe that isn't necessary. A least as a first step is much easier to change the macro to use the test case widget.

For example, at https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#%24%3A%2Feditions%2Ftw5.com%2Fwikitext-macros

Change this

```
\procedure wikitext-example-without-html(src)
<div class="doc-example">
	<$macrocall $name="copy-to-clipboard-above-right" src=<<src>>/>
	<$codeblock code=<<src>>/>
	<p>
		That renders as:
	</p>
	<$transclude $variable="src" $mode="block"/>
</div>
\end
```

to this

```
\procedure wikitext-example-without-html(src)
    <$macrocall $name="copy-to-clipboard-above-right" src=<<src>>/>
    <$testcase>
        <$data title="Output" text=<<src>>/>
    </$testcase>
\end
```

btheado:
I'm seeing some strange rendering behavior regarding `<<currentTiddler>>` variable

With the following test case tiddler, the test case is passing. It must be finding `<<currentTiddler>>` to match `Output`

However, visibly I see it rendered as `TestCases/TestCaseTiddler/currentTiddler` rather than `Output`.

```
created: 20240510013024332
description: currentTiddler should be properly set
modified: 20240510013609588
tags: $:/tags/wiki-test-spec
title: TestCases/TestCaseTiddler/currentTiddler
type: text/vnd.tiddlywiki-multiple

title: Narrative

currentTiddler variable in Output tiddler should be "Output"
+
title: Output

<$text text=<<currentTiddler>>>
+
title: ExpectedResult

<p>Output</p>
```

Here is a json attachement of the tiddler: [TestCases_TestCaseTiddler_currentTiddler.json](https://github.com/Jermolene/TiddlyWiki5/files/15269363/TestCases_TestCaseTiddler_currentTiddler.json)

Jermolene:
Thanks @btheado 

> The test case tiddlers and test case widget seem like a really good replacement for the `wikitext-example-without-html` macro.

Yes, although given that these examples only have a single payload tiddler perhaps it would be nicer to use a more streamlined template.

> Having the widget examples be live-editable will be a great improvement. One nice feature of the wikitext-example macro is the "copy to clipboard" button. Have any thoughts about the equivalent for the test case widget?
> 
> Being able to easily copy the text of the example is useful. In addition or instead, maybe an easy way to get the json for all the tiddlers in the test case widget?

Great idea, I agree. We should have both a "copy to clipboard" button on each source tiddler, and a way to drag and drop the tiddlers to another wiki (or the host wiki), and to save the JSON directly.

> Originally I was thinking all the uses of `wikitext-example-without-html` would be converted to test case tiddlers. Maybe that isn't necessary. A least as a first step is much easier to change the macro to use the test case widget.

Yes, that's an excellent idea. We should still aim to convert the usages to test case tiddlers so that we can specify ExpectedResult, and hence run them as tests.

> I'm seeing some strange rendering behavior regarding `<<currentTiddler>>` variable

Ouch, I will investigate. I'm not surprised to see this kind of wrinkle; the testcase widget mixes widgets from different wikis in a way that hasn't been tested before.

btheado:
> Yes, although given that these examples only have a single payload tiddler perhaps it would be nicer to use a more streamlined template.

There are still many examples for which the default template is very instructive. Whenever an example such as the [RevealWidget](https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#RevealWidget) creates a tiddler, a new tab appears. For a learner, that makes it easy to see what changes are happening.

>> A least as a first step is much easier to change the macro to use the test case widget.

But even this approach requires caution. There are many examples which depend on tiddlers in the main wiki. For example (just a partial list):

* [DiffTextWidget](https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#DiffTextWidget) - uses SampleTiddlerFirst and SampleTiddlerSecond tiddlers
* [SetMultipleVariablesWidget](https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#SetMultipleVariablesWidget) - assumes presence of HelloThere tiddler
* [SetWidget](https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#SetWidget) - a few of these examples assume presence of tiddlers tagged HelloThere.
* [Conditional Shortcut Syntax](https://tiddlywiki5-git-testcase-widget-jermolenes-projects.vercel.app/#Conditional%20Shortcut%20Syntax) tiddler assumes the presence of `$:/info/url/protocol` tiddler, but no `[prefix[$:/info]]` tiddlers are propagated to the subwiki by default.

Jermolene:
> I'm seeing some strange rendering behavior regarding `<<currentTiddler>>` variable

Thanks @btheado this is now fixed in eb4e9d86ac02a0d29e63ab600a4fc93bf8f2f360

> There are still many examples for which the default template is very instructive

Yes, that's very true.

> But even this approach requires caution

Good point. My instinct is to merge this quite soon, and then we can tackle updating the examples gradually. One advantage of the structure introduced in this PR is that the testcases are part of the tw5.com wiki, and can thus be updated directly to tiddlywiki.com because they are part of the documentation.

btheado:
> Thanks @btheado this is now fixed in https://github.com/Jermolene/TiddlyWiki5/commit/eb4e9d86ac02a0d29e63ab600a4fc93bf8f2f360

Great, thanks.

> My instinct is to merge this quite soon, and then we can tackle updating the examples gradually

For sure, I agree.

PR #7819, opened 2023-10-26
Update markdown-it-ins.min.js 

oeyoews:


Jermolene:
Thank you @oeyoews. What is the purpose of this change?

oeyoews:
Okay, I checked the markdown-it-ins dist file again, and it also write `markdown-it-mark` (isn’t this another markdown-it plug-in?). I feel like they wrote it wrong.

Jermolene:
Hi @oeyoews our policy is not to modify 3rd party files bought in via tiddlywiki.files. It makes it much easier to handle subsequent upgrades. If this is a problem, it looks like it would be best tackled within markdown-it.

PR #7820, opened 2023-10-28
Add support for running in-browser tests via playwright in GitHub CLI

### LINKED ISSUE #7812: Browser-based tests should be run under continuous integration
Our test suite runs under both Node.js and in the browser. However, our continuous integration implementation only runs the Node.js tests. There [have been cases](https://github.com/Jermolene/TiddlyWiki5/pull/7683#issuecomment-1778762929) where the browser-based tests have been broken without anybody being aware.

I would very much like some help to get the browser-based tests running under CI, using something like Puppeteer or Playwright. The browser-based tests should be run simultaneously with the regular Node.js-based tests. We will need to figure out a way to reach into the test wiki and extract the test results.

Would anyone be able to take on this task?

Questions and suggestions welcome.

EvidentlyCube:
* `ci.yml` was updated to store the report so that it can be inspected on failure
* `ci-test.sh` was added as an expansion to `test.sh` which installs and runs playwright
* `playwright.spec.js` does the actual verification of opening the test TW edition in browser, waiting for the tests to finish and then verifying it has indeed passed
* `playwright.config.js` Playwrifht configuration

Fixes #7812

----

I can't verify right now that the Github actions and `.sh` changes work. Playwright itself works fine, I verified it by:
* Modify some test to always fail
* Build test edition
* Run `npx playwright test` and confirm the tests have failed
* Revert the change to the test
* Build test edition
* Run `npx playwright test` and confirm it now passes

The additional option in `ci.yml` was based on my [own repo](https://github.com/EvidentlyCube/TW5-PluginShowcase/blob/main/.github/workflows/playwright.yml) so I am fairly sure it should work.


EvidentlyCube:
Okay, the tests are passing now!

Jermolene:
Hi @EvidentlyCube thanks, I'll merge this now and report back and errors.

Jermolene:
...*any* errors.

Jermolene:
Hi @EvidentlyCube it seems to be running with no errors. I'll try making a test PR that contains an error and check that it is flagged.

Jermolene:
Hi @EvidentlyCube the PR with the failing browser tests is #7823. The error is definitely being trapped but I wonder if we could make the way that the errors are reported a little clearer.

image

The "details" [link](https://github.com/Jermolene/TiddlyWiki5/actions/runs/6691336192/job/18178372436?pr=7823) shows this for users who don't have permission:

image

On the other hand, I see the full error details:

image

Can we do a better job of reporting the errors? It would be great if we could update an HTML comment on the PR with more details of the test run, like Vercel does.

EvidentlyCube:
I've no clue, haven't done this type of work before, but I can certainly give it a try this week!

EvidentlyCube:
@Jermolene it might be possible to configure Vercel to better integrate with Playwright and the published report artifact, but that's something you'd have to work on I believe.

In the meantime in #7824 I'll be poking around to see for a way to publish a link to the artifact containing the generated HTML report

Issue #7797, opened 2023-10-20
[Inconsistency] Filtered transclusion syntax does not assign attributes to rendered HTML output


pmario:
While fixing **[BUG] Filtered transclusion that is, much later, followed by curly brace** #7701 -- I did find out that:

The docs in the js code for $:/core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js says:

```
{{{ [tag[docs]] }}}
{{{ [tag[docs]] |tooltip}}}
{{{ [tag[docs]] ||TemplateTitle}}}
{{{ [tag[docs]] |tooltip||TemplateTitle}}}
{{{ [tag[docs]] }}width:40;height:50;}.class.class
```

In the code all the attributes are assigned to the list-like output node. 

**But** only the "template" attribute is actually used by the rendered output
**All other** attributes "tooltip", "style" and "class" are ignored 

------

I think the problem is, that those additional attributes do not make too much sense, since it's not clear to which output element they should be added. 

@Jermolene 
IMO it may make sense to:

a. make them available as variables to the "template" if there is one
b. Ignore the issue and let them be as is for future use
c: Comment the unused `if clauses` and activate them, when we need them in the future. 

At the moment it's confusing for contributors and wasting device memory.

I would suggest **option c:**

@Jermolene -- What do you think?

Jermolene:
Hi @pmario I would be inclined to agree that we should completely removed the unused parts of the regex and the associated code.

pmario:
@Jermolene -- bump -- info needed

Jermolene:
Hi @pmario in these situations I think it's clearer to add the tests in with the PR that includes the fix

pmario:
The latest proposal from Jeremy is at: https://github.com/Jermolene/TiddlyWiki5/pull/7822#issuecomment-1826266336

Which can be changed to the following format, which gives us even more **future** possibilities. 

`vars` ... can be any text except `|` -- even linebreaks are possible
`params` ... similar to `vars` but with basically **no limitations**

```
{{{ [tag[docs]] }}}
{{{ [tag[docs]] | vars }}}
{{{ [tag[docs]] || TemplateTitle }}}
{{{ [tag[docs]] | vars || TemplateTitle }}}
{{{ [tag[docs]] }} param:"value" | another="value" }

{{{ [tag[docs]] }} param:"value" | another="value" | any text we want 
including line-breaks }
```

PR #7822, opened 2023-10-29
Fix 7701 filtered transclusion fails if "}" follows somewhere in the text

### LINKED ISSUE #7701: [BUG] Serious bug with filtered transclusion and style block
### Describe the bug

Typing in the following on tiddlywiki.com or any other TW

```
{{{ [[any filter]] }}} 

text to disappear

<div>
{ anything surrounded by curly brackets inside any html tags }
</div>
```
...outputs this:

image

If the html element is removed, the output seems correct. 

To my surprise I get the faulty behavious regardless of TW version. (How has this possibly passed????)

### Expected behavior

...

### To Reproduce

_No response_

### Screenshots

_No response_

### TiddlyWiki Configuration

Tested on both Chrome and FF

and on TW latest and several older versions

### Additional context

_No response_

pmario:
This PR fixes #7701 and #7797

It contains 4 new tests. 

- The first 2 test should check the standard filtered transclusion parser modes
- Test 3 & 4 are specific tests for the issues mentioned above

- It removes the unused code in [filteredtranscludeblock.js](https://github.com/Jermolene/TiddlyWiki5/compare/master...pmario:fix-7701-filtered-transclusion?expand=1#diff-52d36893ad904e4fc22413858d81478b764b9ed192d4c476b9c61042fa6c30fc) and [filteredtranscludeinline.js](https://github.com/Jermolene/TiddlyWiki5/compare/master...pmario:fix-7701-filtered-transclusion?expand=1#diff-d1419c3a14f675c7134df98c550c79bf45ffc582ed920285ed0f39eb57190b1c) and changes the regexps accordingly. 
- All tests pass from the CLI **and** in the browser

---

The new **block mode regexp is** `\{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}(?:\r?\n|$)`  

<details><summary>Block-mode full description</summary>

```
// \{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}(?:\r?\n|$)
// 
// Options: Case insensitive; Dot matches line breaks; ^$ don’t match at line breaks
// 
// Match the character “{” literally «\{»
// Match the character “{” literally «\{»
// Match the character “{” literally «\{»
// Match the regex below and capture its match into backreference number 1 «([^\|]+?)»
//    Match any character that is NOT a “|” «[^\|]+?»
//       Between one and unlimited times, as few times as possible, expanding as needed (lazy) «+?»
// Match the regular expression below «(?:\|\|([^\|\{\}]+))?»
//    Between zero and one times, as many times as possible, giving back as needed (greedy) «?»
//    Match the character “|” literally «\|»
//    Match the character “|” literally «\|»
//    Match the regex below and capture its match into backreference number 2 «([^\|\{\}]+)»
//       Match any single character NOT present in the list below «[^\|\{\}]+»
//          Between one and unlimited times, as many times as possible, giving back as needed (greedy) «+»
//          The literal character “|” «\|»
//          The literal character “{” «\{»
//          The literal character “}” «\}»
// Match the character “}” literally «\}»
// Match the character “}” literally «\}»
// Match the character “}” literally «\}»
// Match the regular expression below «(?:\r?\n|$)»
//    Match this alternative (attempting the next alternative only if this one fails) «\r?\n»
//       Match the carriage return character «\r?»
//          Between zero and one times, as many times as possible, giving back as needed (greedy) «?»
//       Match the line feed character «\n»
//    Or match this alternative (the entire group fails if this one fails to match) «$»
//       Assert position at the very end of the string «$»
```
</details> 

The new **Inline regexp is:**: ` \{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}`

<details><summary>Inline-mode full description</summary>

```
// \{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}
// 
// Options: Case insensitive; Dot matches line breaks; ^$ don’t match at line breaks
// 
// Match the character “{” literally «\{»
// Match the character “{” literally «\{»
// Match the character “{” literally «\{»
// Match the regex below and capture its match into backreference number 1 «([^\|]+?)»
//    Match any character that is NOT a “|” «[^\|]+?»
//       Between one and unlimited times, as few times as possible, expanding as needed (lazy) «+?»
// Match the regular expression below «(?:\|\|([^\|\{\}]+))?»
//    Between zero and one times, as many times as possible, giving back as needed (greedy) «?»
//    Match the character “|” literally «\|»
//    Match the character “|” literally «\|»
//    Match the regex below and capture its match into backreference number 2 «([^\|\{\}]+)»
//       Match any single character NOT present in the list below «[^\|\{\}]+»
//          Between one and unlimited times, as many times as possible, giving back as needed (greedy) «+»
//          The literal character “|” «\|»
//          The literal character “{” «\{»
//          The literal character “}” «\}»
// Match the character “}” literally «\}»
// Match the character “}” literally «\}»
// Match the character “}” literally «\}»
```
</details>

### DIFF
--- a/core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
+++ b/core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
@@ -7,10 +7,7 @@ Wiki text rule for block-level filtered transclusion. For example:
 
 ```
 {{{ [tag[docs]] }}}
-{{{ [tag[docs]] |tooltip}}}
 {{{ [tag[docs]] ||TemplateTitle}}}
-{{{ [tag[docs]] |tooltip||TemplateTitle}}}
-{{{ [tag[docs]] }}width:40;height:50;}.class.class
 ```
 
 \*/
@@ -26,7 +23,8 @@ exports.types = {block: true};
 exports.init = function(parser) {
 	this.parser = parser;
 	// Regexp to match
-	this.matchRegExp = /\{\{\{([^\|]+?)(?:\|([^\|\{\}]+))?(?:\|\|([^\|\{\}]+))?\}\}([^\}]*)\}(?:\.(\S+))?(?:\r?\n|$)/mg;
+	// this.matchRegExp = /\{\{\{([^\|]+?)(?:\|([^\|\{\}]+))?(?:\|\|([^\|\{\}]+))?\}\}([^\}]*)\}(?:\.(\S+))?(?:\r?\n|$)/mg;
+	this.matchRegExp = /\{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}(?:\r?\n|$)/mg;
 };
 
 exports.parse = function() {
@@ -34,10 +32,7 @@ exports.parse = function() {
 	this.parser.pos = this.matchRegExp.lastIndex;
 	// Get the match details
 	var filter = this.match[1],
-		tooltip = this.match[2],
-		template = $tw.utils.trim(this.match[3]),
-		style = this.match[4],
-		classes = this.match[5];
+		template = $tw.utils.trim(this.match[2]);
 	// Return the list widget
 	var node = {
 		type: "list",
@@ -46,17 +41,8 @@ exports.parse = function() {
 		},
 		isBlock: true
 	};
-	if(tooltip) {
-		node.attributes.tooltip = {type: "string", value: tooltip};
-	}
 	if(template) {
 		node.attributes.template = {type: "string", value: template};
-	}
-	if(style) {
-		node.attributes.style = {type: "string", value: style};
-	}
-	if(classes) {
-		node.attributes.itemClass = {type: "string", value: classes.split(".").join(" ")};
 	}
 	return [node];
 };
--- a/core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
+++ b/core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
@@ -7,7 +7,7 @@ Wiki text rule for block-level filtered transclusion. For example:
 
 ```
 {{{ [tag[docs]] }}}
-{{{ [tag[docs]] ||TemplateTitle}}}
+{{{ [tag[docs]] || TemplateTitle }}}
 ```
 
 \*/
@@ -22,8 +22,7 @@ exports.types = {block: true};
 
 exports.init = function(parser) {
 	this.parser = parser;
-	// Regexp to match
-	// this.matchRegExp = /\{\{\{([^\|]+?)(?:\|([^\|\{\}]+))?(?:\|\|([^\|\{\}]+))?\}\}([^\}]*)\}(?:\.(\S+))?(?:\r?\n|$)/mg;
+	// Regexp to match TW versions > v5.3.1
 	this.matchRegExp = /\{\{\{([^\|]+?)(?:\|\|([^\|\{\}]+))?\}\}\}(?:\r?\n|$)/mg;
 };
 
--- a/core/modules/parsers/wikiparser/rules/filteredtranscludeinline.js
+++ b/core/modules/parsers/wikiparser/rules/filteredtranscludeinline.js
@@ -7,10 +7,7 @@ Wiki text rule for inline filtered transclusion. For example:
 
 ```
 {{{ [tag[docs]] }}}
-{{{ [tag[docs]] |tooltip}}}
 {{{ [tag[docs]] ||TemplateTitle}}}
-{{{ [tag[docs]] |tooltip||TemplateTitle}}}
-{{{ [tag[docs]] }}width:40;height:50;}.class.class
 ```
 
 \*/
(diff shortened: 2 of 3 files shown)


Jermolene:
Hi @pmario I think the problem here is that the regexp is too greedy. Removing the unused syntactic structures does fix it, but only because it removes that part of the regular expression. It would be better to fix it by making the regexp not be greedy.

As it happens, I do think there are possibilities for us to make use of the unused parts of the syntax that we should consider:

```
{{{ [tag[docs]] }}}
{{{ [tag[docs]] |variablename}}}
{{{ [tag[docs]] ||TemplateTitle}}}
{{{ [tag[docs]] |variablename||TemplateTitle}}}
{{{ [tag[docs]] }} param:value another:value}
```

pmario:
**As a reminder**

**The existing regexp** does create matches for 

```
{{{ [tag[docs]] }}}
{{{ [tag[docs]] | tooltip }}}
{{{ [tag[docs]] || TemplateTitle }}}
{{{ [tag[docs]] | tooltip || TemplateTitle }}}
{{{ [tag[docs]] }} width:40;height:50; }.class.class
```

**The proposal is:** 

```
{{{ [tag[docs]] }}}
{{{ [tag[docs]] |variablename}}}
{{{ [tag[docs]] ||TemplateTitle}}}
{{{ [tag[docs]] |variablename||TemplateTitle}}}
{{{ [tag[docs]] }} param:value another:value}
```

- For readability IMO it needs to be possible to have whitespace in between the separators. 
- The new pattern is different
- Since it does not have the `.class.class` pattern at the end it may work -> More tests will be needed. 

----------

I wrote: https://github.com/Jermolene/TiddlyWiki5/issues/7701#issuecomment-1772762691

>@Jermolene ... The problem is caused by the trailing whitespace character after the closing braces `}}} ` in the OP.  Any other "trailing" character has the same effect. eg: `{{{ [[any filter]] }}}a`
>
>The **reason** is, that with a trailing character the last 2 braces `}}` and the following text **match group 4** -- The first closing brace `}` will be added to match group 1
>
>- if and only if there is an other closing brace `}` in the text that follows.
>- If there is no closing brace in the following text **the whole rule fails**. 
>
>------
>
>From a pragmatic point of view a trailing character makes the whole text block an **inline** filtered transclusion. .. **BUT since the block-rule creates some valid matches it does not switch to the inline parser.** 
>
>I do have some ideas, but all of them are only partial solutions. 

--------

Code that causes the problem from https://github.com/Jermolene/TiddlyWiki5/issues/7701#issue-1868247366

- The trailing space after `{{{ [[any filter]] }}} `  causes the problem 
- `{{{ [[any filter]] }}}x` will also cause the problem

```
{{{ [[any filter]] }}} 

this string disappeas and the succeeding curly brace becomes wikified, or maybe it is a brace from the filtered transclusion above

}
```

pmario:
@Jermolene .. with the latest commits, we do have _all_ the parameters back. I did rename them with more generic names. So we are able to rename them again in the future, once we know, how to use them.

I did add 2 MAX tests, which test all the parameters, so we should save for accidental bugs in the future.

pmario:
The regexp fix for block mode was actually much simpler as expected. -- I only needed to look at it with some time in between.

saqimtiaz:
If we do decide to extend the filtered transclusion syntax to support parameters, this proposed syntax `{{{ [tag[docs]] |variablename||TemplateTitle}}}`  is discordant with the existing transclusion syntax `{{MyTiddler||TemplateTitle|Parameter|SecondParameter}}` in terms of the positioning of the parameters relative to the template. This would likely cause confusion amongst users. Could we adopt the existing syntax, i.e. reverse the positions of the parameters and the template without breaking backwards compatibility?

pmario:
> Could we adopt the existing syntax, i.e. reverse the positions of the parameters and the template without breaking backwards compatibility?

I'm not really sure. The regexp is super tricky. May be it is possible. Since at the moment there is only code that assigns the `template` to any widget-tree -> DOM ... We should have no problem with the compatibility. 

At the moment only the parse-tree is populated, which should cause failing tests -- intentionally.

twMat:
Just a note to say that I have again been affected by the bug that this PR intends to resolve. Luckily I had a vague memory of it so I could find the discussion for it, and see the workaround (to remove a trailing space). Other users would not know about this.
If the PR is up to par, please merge @Jermolene

Jermolene:
Hi @twMat @pmario as well as the bug fix this PR includes some relatively controversial new features that need further discussion before we can consider merging them.

Would it be possible to make a PR with just the bug fix?

pmario:
TLDR;
The best way would be to agree on a new proposal first. Since it is the same amount of work and doing it twice does not make too much sense.

----

The current regexp is very complex and contains some "existing but unused features". 

- Completely removing them was no option
- I made a proposal, which is possible but not optimal since it is different to the transclusion-parameters as Saq found out: https://github.com/Jermolene/TiddlyWiki5/pull/7822#issuecomment-1919287164

So I will need to take that one into account and see if it is possible to come up with a new proposal, that can fix the problem and also allows us to add new features in the future without breaking everything.

As far as I can remember the `}.class.class` has to go, since we cannot fix the problem otherwise. I think it was the main reason for the problem as described at: https://github.com/Jermolene/TiddlyWiki5/pull/7822#issuecomment-1828062376

So: We need to agree on a new proposal first.

PR #7827, opened 2023-11-01
Alternate fix for inconsistent list template syntax

rmunn:
Alternative approach to fixing #7804, making `$list-template` and `$list-empty` into real widgets (which don't render anything) so that the `<$list-empty>` widget doesn't have to be removed when building the template from the `$list` widget's body. (As suggested in https://github.com/Jermolene/TiddlyWiki5/issues/7804#issuecomment-1779701517).

### DIFF
--- a/core/modules/widgets/list.js
+++ b/core/modules/widgets/list.js
@@ -160,7 +160,7 @@ ListWidget.prototype.makeItemTemplate = function(title,index) {
 			// Check for a <$list-item> widget
 			if(this.explicitListTemplate) {
 				templateTree = this.explicitListTemplate;
-			} else if (!this.explicitEmptyTemplate) {
+			} else {
 				templateTree = this.parseTreeNode.children;
 			}
 		}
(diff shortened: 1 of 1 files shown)


rmunn:
Note how the `ListWidget/WithMissingTemplate` test fails, because the `$list` body contains a `$list-empty` element and is therefore *not* empty, so the default template is not used. To fix this, we'll need to write code that searches the widget's body and ignores any `$list-empty` widgets (which might, incidentally, be inside a `<p>` tag), and only counts the body as non-empty if it finds at least one child node that is *not* a `$list-empty` widget. Which ends up looking quite similar to the approach taken in #7810.

rmunn:
With commit 53b8969, the unit tests pass again. But notice that we still had to do a clone-and-mutate operation to make the WithMissingTemplate test pass. I think there might be a more efficient way to do it, but I'm not sure if it will work. I'll try it and see if I can pull it off.

rmunn:
Found a way to do this efficiently, avoiding the clone-and-mutate approach altogether. I also moved `findExplicitTemplates()` into the `initialise()` step since parse trees are never modified during a widget's lifetime, which allows us to skip calling it again on every re-render. Once this is merged, it will fix #7804 and unblock #7694.

rmunn:
@Jermolene, is there anything you need me to change here? I'd like to get this merged as soon as possible because it's blocking https://github.com/Jermolene/TiddlyWiki5/pull/7694 which is otherwise ready to go, and I'd really like #7694 to make it in before the merge freeze in <del>ten</del> nine days.

Jermolene:
Thank you @rmunn, all good.

I wonder if it might be helpful to add a follow-up docs update explaining the behaviour?

PR #7829, opened 2023-11-01
Make unit tests run faster in Github Actions

### LINKED ISSUE #7828: [BUG] Running unit tests is slow now
### Describe the bug

The TW unit tests in CI, which run every time you push to a PR branch used to be fast, finishing in less than 10 seconds. Now that Playwright has been added, they take much longer: https://github.com/Jermolene/TiddlyWiki5/actions/runs/6717219522/job/18254693717?pr=7827 took over a minute to fail. If the `ci-test.sh` script were swapped to run `node ./tiddlywiki.js ./editions/test` first, and then do the `npm install playwright` and `npx playwright install` tests only later, then the CI test runs could finish fast when there's a test that fails under Node.js, and not have to go through the steps of downloading and installing Playwright dependencies. That way when a PR is in progress and still has failing tests, the submitter can get feedback in seconds instead of waiting over a minute to see the failure.

### Expected behavior

Failing tests fail fast, and only successful test runs under Node.js go through the process of downloading Playwright dependencies, installing browsers, and running the Playwright tests.

### To Reproduce

Rather than a reproduction, a suggested fix: move the `npm install playwright` and `npx playwright install` lines further down in the `ci-test.sh` script.

### Screenshots

_No response_

### TiddlyWiki Configuration

Not relevant

### Additional context

An alternative solution, that would involve more work, would be to split the ci.yml workflow into two steps: one that runs tests under Node.js, and a separate one that runs them in the browser. The default Github Actions config is to fail fast, so that if one step fails, any other still-running steps are cancelled. The advantage this would give is that *successful* tests would also show up fast: the "run tests under Node.js" check would show up in seconds, while the "run tests under Playwright" check would take a little bit longer — but the PR author would immediately have some confidence that the Playwright tests would, almost certainly, pass as well.

rmunn:
Fixes #7828.

### DIFF
--- a/bin/ci-test.sh
+++ b/bin/ci-test.sh
@@ -2,9 +2,6 @@
 
 # test TiddlyWiki5 for tiddlywiki.com
 
-npm install playwright @playwright/test
-npx playwright install chromium firefox --with-deps
-
 node ./tiddlywiki.js \
 	./editions/test \
 	--verbose \
@@ -12,5 +9,8 @@ node ./tiddlywiki.js \
 	--rendertiddler $:/core/save/all test.html text/plain \
 	--test \
 	|| exit 1
+
+npm install playwright @playwright/test
+npx playwright install chromium firefox --with-deps
 
 npx playwright test


Jermolene:
Thanks @rmunn

PR #7835, opened 2023-11-14
Avoid skipping extra whitespace in wikiparser.js

rmunn:
When wikiparser.js parses text looking for a pragma block, it skips whitespace before looking for the next pragma. If no pragma is found, we should return the parse position to the original location so that the skipped whitespace can be parsed as a text node. This allows the attribute `join=" and "` to parse as " and " (leading and trailing whitespace intact) rather than "and " (where the leading whitespace was skipped).


Jermolene:
Great, thanks @rmunn

PR #7844, opened 2023-11-20
Fixes Text Parser being impacted by overrides to codeblock widget

saqimtiaz:
This PR fixes an issue [reported in the forum](https://talk.tiddlywiki.org/t/how-highlight-js-works-on-codeblock/8083/10?u=saqimtiaz) where overriding the codeblock widget with a custom widget impacts transclusions and any other uses of the Text Parser.

The problem is caused by the transclude widget (and thus also by the macrocall widget which is used in the tabs macro) when used with the output type text/plain. The parser that wikifies the variable to retrieve the plain text output [wraps the variable in a codeblock](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/parsers/textparser.js#L15-L25)

This PR replaces the codeblock widget used in the Text Parser with a genesis widget with the remappable attribute set to "no", to create the codeblock widget, thus rendering it immune to custom widget overrides. A test has been added that fails before the PR is applied.

### DIFF
--- a/core/modules/parsers/textparser.js
+++ b/core/modules/parsers/textparser.js
@@ -8,28 +8,31 @@ The plain text parser processes blocks of source text into a degenerate parse tr
 \*/
 (function(){
 
-/*jslint node: true, browser: true */
-/*global $tw: false */
-"use strict";
-
-var TextParser = function(type,text,options) {
-	this.tree = [{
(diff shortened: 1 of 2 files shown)


Jermolene:
Great thanks @saqimtiaz

PR #7845, opened 2023-11-20
feat: add support for data attributes to Draggable and Droppable widgets

saqimtiaz:
This PR extends #7769 to add support for data and style attributes to the $draggable and $droppable widgets.

Tests for refresh handling are included.


Jermolene:
Marvellous thank you @saqimtiaz

PR #7846, opened 2023-11-20
Fixes refresh issues for checkbox and links widgets for data attributes

saqimtiaz:
This PR fixes refresh issues for the data attributes introduced in #7769

Checkbox widget: missing refresh for data attributes added
Link widget: fixed issue where the data attributes were only added when `tv-wikilinks` was set to no. Now  data attributes are supported in both when the variable is set to no or yes.

Bonus: refresh test for ButtonWidget.


Jermolene:
Great thanks @saqimtiaz

PR #7848, opened 2023-11-20
Refactors Select widget to directly create DOM node

saqimtiaz:
This PR refactors the select widget to directly create its DOM node and fixes support for data and style attributes introduced in #7769 

The `Widget.prototype.assignAttributesToParseTreeNode()` method has been removed.

### DIFF
--- a/core/modules/widgets/select.js
+++ b/core/modules/widgets/select.js
@@ -182,16 +182,21 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
 SelectWidget.prototype.refresh = function(changedTiddlers) {
 	var changedAttributes = this.computeAttributes();
 	// If we're using a different tiddler/field/index then completely refresh ourselves
-	if($tw.utils.count(changedAttributes) > 0) {
+	if(changedAttributes.tiddler || changedAttributes.field || changedAttributes.index || changedAttributes.tooltip || changedAttributes.tabindex) {
 		this.refreshSelf();
 		return true;
 	} else {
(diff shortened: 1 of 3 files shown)


Jermolene:
Thank you @saqimtiaz I think that's a great improvement. I wonder whether we should retain the `Widget.prototype.assignAttributesToParseTreeNode` method just in case there's a custom widget out there that started out life as a clone of the "element" widget. We could add a comment to note that it is no longer used by the core, and only retained for backwards compatibility.

saqimtiaz:
@Jermolene Unless I am misunderstanding your intent or have misread the diff, the `Widget.prototype.assignAttributesToParseTreeNode` method is first introduced in this PR so there cannot be a custom widget in the wild that uses it. So the method has been removed from this PR, but was never in the core in the first place.

The refresh handling needs improvement, will revisit this with fresher eyes before marking as ready for review.

Jermolene:
> @Jermolene Unless I am misunderstanding your intent or have misread the diff, the `Widget.prototype.assignAttributesToParseTreeNode` method is first introduced in this PR so there cannot be a custom widget in the wild that uses it. So the method has been removed from this PR, but was never in the core in the first place.

Of course, you are quite correct, my apologies.
> 
> The refresh handling needs improvement, will revisit this with fresher eyes before marking as ready for review.

saqimtiaz:
@Jermolene this one is ready for review.

Jermolene:
Great thanks @saqimtiaz

PR #7849, opened 2023-11-21
Allow negative indexes in json operators

### LINKED ISSUE #7839: [IDEA] Allow negative indexes in jsonget/set/etc operators for array access from end
Accessing the last item of an array is a pretty common task, but it's not easy to do with the existing `jsonget`, `jsonset`, etc. operators. You first have to get the last index of the array with `[<jsondata>jsonindices[]maxall[]]` and then store that in a variable (let's call it `lastIndex`), then do `[<jsondata>jsonget<lastIndex>]`. It would be nice to have an easy way to refer to "the last item in the array" without knowing the array's length ahead of time.

In any situation where one of the `jsonget` family of operators would count a numeric as an index into an array, any index that's a negative number would be counted as (index + array.length). So an index of -1 (negative one) would count as "array.length - 1", i.e. the last item in the array. An index of -2 (negative two) would count as the next-to-last item in the array, and so on.

An alternative could be to use a suffix like `fromend`, so that `jsonget:fromend[1]` would mean "the last item of the array". But there are already so many suffixes, and most programming languages already have the concept of "-1 means the last item" built in, so the concept is familiar to anyone who's done any programming before. Plus, it's not clear whether `fromend[1]` or `fromend[0]` should mean "the last item", whereas -1 meaning "last item" is pretty universal across languages and is therefore easy to remember.

**Additional context**
This idea came out of a discussion at https://github.com/Jermolene/TiddlyWiki5/pull/7742#issuecomment-1740595358.

rmunn:
Fixes #7839.

This PR allows negative indexes to be used as input to the JSON operators: -1 means last item of the array, -2 means next-to-last item, and so on. A negative index that would go "past" the start of the array will be treated as an invalid index, e.g. in an array of length 5, the index -5 would be the first item in the list, but trying to do `jsonget[-6]` would produce an empty result. Similarly, trying to do `jsonset[-6],[newValue]` would return the original JSON data unchanged, just as `jsonset` does with any other invalid index.

### DIFF
--- a/core/modules/filters/json-ops.js
+++ b/core/modules/filters/json-ops.js
@@ -213,6 +213,18 @@ function getDataItemType(data,indexes) {
 	}
 }
 
+function getItemAtIndex(item,index) {
+	if($tw.utils.hop(item,index)) {
+		return item[index];
+	} else if($tw.utils.isArray(item)) {
+		index = $tw.utils.parseInt(index);
+		if(index < 0) { index = index + item.length };
+		return item[index]; // Will be undefined if index was out-of-bounds
+	} else {
+		return undefined;
+	}
+}
+
 /*
 Given a JSON data structure and an array of index strings, return the value at the end of the index chain, or "undefined" if any of the index strings are invalid
 */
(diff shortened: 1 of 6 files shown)


Jermolene:
Looks good thanks @rmunn

PR #7852, opened 2023-11-22
Fixes issues in the PR "Button widget data attributes"

saqimtiaz:
Fixes:
* ensures that event attributes are excluded
* add tests for event attribute exclusion
* ensures that style. attributes are supported when no sourcePrefix is specified
* adds tests for style. attributes for the element widget
* extends fakedom to accurately assign style property of fake DOM node.
* ensures that attribute order is respected and adds a test for this


saqimtiaz:
@Jermolene when you get the opportunity some feedback would be appreciated. I discovered a lot of broken edge cases with the implementation in the original PR for which I have added tests.

Jermolene:
Fantastic, thank you @saqimtiaz

PR #7855, opened 2023-11-23
Core macros don't allow pragmas in action strings

Jermolene:
As reported by @CodaCodr [over at talk.tiddlywiki.org](https://talk.tiddlywiki.org/t/actions-attribute-of-tag-picker-macro-does-not-support-nested-procs-variables/8518), the core "tag-picker" macro does not permit pragmas within the "actions" parameter:

```
\procedure tag-actions()
  \procedure update()
    <!--  do things -->
  \end update
<<update>>
\end tag-actions

<$transclude $variable=tag-picker tagField="my-tags" actions=<<tag-actions>> />
```
The result is garbled:


core/wiki/macros/tag-picker.tid
```diff
@@ -16,7 +16,7 @@ second-search-filter: [tags[]is[system]search:title<userInput>sort[]]
 		emptyMessage="<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter='-[<tag>]'/>"
 	>
 		<$action-listops $tiddler=<<saveTiddler>> $field=<<__tagField__>> $subfilter="[<tag>trim[]]"/>
-		$actions$
+		<$macrocall $name="__actions__"/>
```
saqimtiaz, line 19: Suggest we use the transclude widget: `<$transclude $variable="__actions__"/>`

PR #7858, opened 2023-11-26
Updates help text for CommandsCommand to avoid deduplication of commands

### LINKED ISSUE #7857: [BUG] enlist cannot be used with `--commands` because removes duplicated entries, documentation
### Describe the bug

Tryting to read commands from a tiddler to pass to the `--commands` command I noticed that it skips to return one of the e `--render`.  The documentation suggests to use `enlist` https://tiddlywiki.com/#CommandsCommand

### Expected behavior

If this is the expected behaviour of `enlist` it should be removed from the documentation as it's common to have multiple `--renders` or templates in a list of commands to generate static files.

### To Reproduce

Given an `example-tiddler` tiddler with content `–verbose cool –verbose –render cool cool` a filter `[elist{example-tiddler}]` only returns the first occurence of each duplicate `--verbose cool --render`.

### Screenshots

Screenshot_20231126_111316

### TiddlyWiki Configuration

- Version [e.g. v5.3.1]

### Additional context

`splitregexp[\n]` seems to be a viable alternative to splits tokens from a tiddler with each command on a new line. `split[ ]` would only work for all the commands on a single line but also would split ``[[valid tiddlers]]`` that include a space

saqimtiaz:
Fixes #7857 where the given example of using `enlist[] `on commands retrieved from a text reference leads to deduplication of the commands.

### DIFF
--- a/core/language/en-GB/Help/commands.tid
+++ b/core/language/en-GB/Help/commands.tid
@@ -10,7 +10,7 @@ Sequentially run the command tokens returned from a filter
 Examples
 
 ```
---commands "[enlist{$:/build-commands-as-text}]"
+--commands "[enlist:raw{$:/build-commands-as-text}]"
 ```
 
 ```


Jermolene:
Thanks @saqimtiaz

PR #7861, opened 2023-11-27
Extends the tour plugin with a condition field

saqimtiaz:
This PR extends the Tour plugin introduced in #7734 with:

- [x]  a `condition` field that determines whether a step should be displayed or skipped
  - [x] refactors filter expressions to use a function that gets all tour tiddlers filtered by the condition field
- [x]  support for custom classes to be assigned to the tour wrapper via the field `class` in the tiddler defining the tour
- [x]  replaces an instance of a macro with a procedure
- [x] a fix for a typo in the CSS for fullscreen tours.
- [x] makes the "show me a hint" text configurable via `hint` field in the step tiddler.
- [x] renames `condition` field to `step-success-filter`
- [x] renames `hint` field to `hint-text`
- [x] renames `selector` field to `hint-selector`
- [x] makes all tour globals local variables that are imported as needed, to allow for greater room for future improvements without backwards compatibility concerns
- [x] extends docs tiddler to cover tour step tiddler fields 

**Future improvements:**
Moved to a new issue: #7876 

- introduce variables that can be used to suppress UI elements in a tour:
  - suppress the automatic call to `<<tour-buttons>>` in each step by means of a `show-tour-buttons` variable
  - suppress the confetti automatically displayed on the last step of a tour by means of a `show-confetti-on-completion` variable
  - suppress the tour-HUD by means of a `show-tour-hud` variable
- introduce support for a variables tiddler for each tour from which variable definitions are imported, specified via a field name in the tour definition tiddler (`import-variables-from`)


Jermolene:
Thanks @saqimtiaz looks good, I will review properly tomorrow.

PR #7865, opened 2023-11-28
Update forum link in update wizard

mateuszwilczek:
The update wizard linked to the old Google Groups forum after performing the upgrade.

This PR changes the link to talk.tiddlywiki.org

I also went ahead and updated the references to the forum in the few places I have found them in the es-ES and de-AT editions. Not sure if it's worth much, as these editions seem to be very out of date and rudimentary.


Jermolene:
Great, thank you @mateuszwilczek

PR #7866, opened 2023-11-29
Add start and end properties to WikiText AST nodes for all elements.

Gk0Wk:
### Summary

This pull request introduces a significant enhancement to the WikiText parser by adding `start` and `end` properties to the AST nodes that did not previously have them. This update includes nodes such as lists, codeblocks, and various other elements that were lacking these properties.

### Background

The ability to identify the exact substring range of a syntax subtree within the source file is crucial for advanced functionalities like section editing. While some progress had been made in the past by adding `start` and `end` properties to paragraph elements, other nodes were still missing this information.

### Changes

With this PR, I've implemented a uniform check across all syntax rules during the AST generation process. Now, every node produced by the parser will have its corresponding `start` and `end` properties, marking the range of the substring it represents in the source text.

### Benefits

- Enables the implementation of more complex features such as segment editing.
- Improves the accuracy and usefulness of the AST for further processing and manipulation.
- Provides consistency across different types of nodes in the AST.

### Testing

Before:

image

After:

image

I look forward to your feedback on these enhancements!


saqimtiaz:
@Gk0Wk I have not reviewed the code yet but this is a long overdue improvement, thank you. 

I note that the tests are failing and need to be updated.

Gk0Wk:
@Jermolene Yes, I completely agree. In fact, today I also wanted to do the same thing, that is, to consider the `rule` as an attribute of the node.

However, I'm not sure if this alone is sufficient to describe everything completely. What I mean is, can we reconstruct the original wiki string without any discrepancies solely based on these? Is it possible that we still need more context?

Nevertheless, this would indeed be better than before. The previous AST was geared towards widget generation, or rather, it was oriented towards rendering, and not for describing all the information.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -215,6 +215,8 @@ WikiParser.prototype.parsePragmas = function() {
 			// Set the start and end positions of the pragma rule if
 			if (subTree[0].start === undefined) subTree[0].start = start;
 			if (subTree[subTree.length - 1].end === undefined) subTree[subTree.length - 1].end = this.pos;
+			var rule = Object.getPrototypeOf(nextMatch.rule);
+			for (const node of subTree) node.rule = rule;
```
pmario, line 219: For TW  code we still use ES5 spec. So you need to use `var` instead of `const`
If you need to iterate over "own properties" of an object or array, TW has a utility function.
See: [$tw.utils.eache(obj, callback)](https://github.com/Gk0Wk/TiddlyWiki5/blob/4200495055b86e80ced6e1776080f51d4581623f/boot/boot.js#L141)

Gk0Wk: 👌

Gk0Wk:
> @Gk0Wk I have not reviewed the code yet but this is a long overdue improvement, thank you.
> 
> I note that the tests are failing and need to be updated.

I've looked at the error messages from the testing process, but I can't understand what they mean. For example:

```
1) WikiText parser tests should parse inline macro calls
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:208:56)
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:213:1[24](https://github.com/Jermolene/TiddlyWiki5/actions/runs/7032732431/job/19137077975?pr=7866#step:4:25))
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:218:48)
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:223:35)
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:2[28](https://github.com/Jermolene/TiddlyWiki5/actions/runs/7032732431/job/19137077975?pr=7866#step:4:29):37)
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:234:[30](https://github.com/Jermolene/TiddlyWiki5/actions/runs/7032732431/job/19137077975?pr=7866#step:4:31))
        at <Jasmine>
  Message:
    Expected $[0] not to have properties
        parseRule: null
  Stack:
    Error: Expected $[0] not to have properties
        parseRule: null
        at <Jasmine>
        at UserContext.<anonymous> (test-wikitext-parser.js:2[39](https://github.com/Jermolene/TiddlyWiki5/actions/runs/7032732431/job/19137077975?pr=7866#step:4:40):[55](https://github.com/Jermolene/TiddlyWiki5/actions/runs/7032732431/job/19137077975?pr=7866#step:4:56))
        at <Jasmine>
```

Can someone help explain them to me?

linonetwo:
> distinguishes a `<$list>` widget

We can do that in another PR.

Also, I've been long waiting for [macro call, distinguish from transclusion #7646](https://github.com/Jermolene/TiddlyWiki5/discussions/7646)

core/modules/parsers/wikiparser/wikiparser.js
linonetwo: Save this as GPTs custom prompt...

Jermolene:
> @Jermolene Yes, I completely agree. In fact, today I also wanted to do the same thing, that is, to consider the `rule` as an attribute of the node:

I think it would be more useful if the "rule" property contained a string. It's very useful to be able to save parse trees in JSON.

> However, I'm not sure if this alone is sufficient to describe everything completely. What I mean is, can we reconstruct the original wiki string without any discrepancies solely based on these? Is it possible that we still need more context?

I think we would indeed still have some more work to do, but it would be a huge improvement.

> Nevertheless, this would indeed be better than before. The previous AST was geared towards widget generation, or rather, it was oriented towards rendering, and not for describing all the information.

Yes, exactly.

> I've also noticed that some rules do not generate any nodes, such as `comment` and `parserule`, whose syntactic information is discarded. `styleblock` does not contain information about itself but only returns the block it is wrapped in, while `quoteblock`, `prettylink` are also missing some information about the substring part.

Yes, we would need to retain some ore information in the parse tree. That may well have performance implications of course.

> Meanwhile, plain text is treated as a paragraph element and does not correspond to any rule.

I think a reasonable initial goal might be to be able to convert parse trees to a kind of normalised wikitext translation, where we don't have 100% fidelity of the source, but the rendered output is identical.

> Is this because I need to further modify the expected results in test-wikitext-parser.js? It looks like there's a significant amount of work.

Yes, I am afraid so. It is a lot of work. It may well be that some significant refactoring might be needed.

For example, the quickest way to get the existing tests running again might be to hack the parse function that it uses to remove the new attributes. Of course then we'd need another set of tests for the new attributes, but we could use the newer wikitext test format for those.

pmario:
>For example, the quickest way to get the existing tests running again might be to hack the parse function that it uses to remove the new attributes. Of course then we'd need another set of tests for the new attributes, but we could use the newer wikitext test format for those.

I would be not happy with a test-hack. We do not have the best test coverage for a project that size and hacking tests IMO makes the unreliable. 

I did not run the tests with this PR yet, but I think basically all of the parser related tests fail now. -> Because they should.

**I'll have a closer look**

pmario:
>> distinguishes a <$list> widget
>
>We can do that in another PR, [. . .]

I also think so. IMO this PR should be merged as is **after the tests have been fixed** and in a second step we can add more info to the AST

linonetwo:
> convert parse trees to a kind of normalised wikitext translation

This is already done in https://github.com/tiddly-gittly/wikiast , should I PR it to the core? But it rely on unist library.

Gk0Wk:
I have modified all the expected results in test-wikitext-parser.js and verified the correctness of the expected results; the tests have passed.

Gk0Wk:
@Jermolene @pmario New changes for html aprser:

## Overview
Fixing a parsing issue within the WikiText HTML syntax parser. The original implementation incorrectly set the `start` and `end` properties of an HTML AST node to only cover the opening tag's substring, neglecting the full extent of the HTML element.

## Changes Made
- The `start` and `end` properties have been redefined to represent the boundaries of the entire HTML node substring, from the start of the opening tag to the end of the closing tag.
- Introduced four new properties to accurately track positions of opening and closing tags:
  - `openTagStart`: The index where the opening tag begins.
  - `openTagEnd`: The index where the opening tag ends.
  - `closeTagStart`: The index where the closing tag begins.
  - `closeTagEnd`: The index where the closing tag ends.

These modifications ensure that the AST node for HTML elements now correctly encapsulates the complete range of the HTML snippet, allowing for more precise manipulation and analysis.

core/modules/parsers/wikiparser/rules/html.js
```diff
@@ -60,6 +64,27 @@ exports.parse = function() {
...
+			while (tag.closeTagStart >= tag.start) {
+				var char = this.parser.source[tag.closeTagStart - 1];
+				if (char === '>') {
+					tag.closeTagStart = -1;
+					break;
+				}
+				if (char === '<') {
+					break;
+				}
+				tag.closeTagStart--;
```
pmario, line 82: We usually do _not_ use this C-type decrement, but we do use `-=` .. `tag.closeTagStart -= 1` which is a bit more to type, but not ambiguous

Jermolene:
Tagging @flibbles – would these changes to the parse tree be helpful and/or problematic for Uglify/Relink?

flibbles:
Tests all pass for this. It's additive rather than changing, so shouldn't be a problem.

I actually considered proposing something like this myself a while back. It would have made it much easier for my plugins to figure things out instead of having to count out start and stop for themselves after-the-fact.

saqimtiaz:
> Fixing a parsing issue within the WikiText HTML syntax parser. The original implementation incorrectly set the `start` and `end` properties of an HTML AST node to only cover the opening tag's substring, neglecting the full extent of the HTML element.

@Gk0Wk can we please add some tests for these cases?

saqimtiaz:
@Gk0Wk I would love to see this get merged and I think all that is missing now is some more tests for the issues that you fixed.

core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
```diff
@@ -31,6 +31,16 @@ exports.init = function(parser) {
...
+	var filterStart = this.parser.pos + 3;
+	var filterEnd = filterStart + this.match[1].length;
+	var toolTipStart = filterEnd + 1;
+	var toolTipEnd = toolTipStart + (this.match[2] ? this.match[2].length : 0);
+	var templateStart = toolTipEnd + 2;
+	var templateEnd = templateStart + (this.match[3] ? this.match[3].length : 0);
+	var styleStart = templateEnd + 2;
+	var styleEnd = styleStart + (this.match[4] ? this.match[4].length : 0);
+	var classesStart = styleEnd + 1;
+	var classesEnd = classesStart + (this.match[5] ? this.match[5].length : 0);
```
saqimtiaz, line 43: We always collapse multiple var statements into one:

```
var filterStart = this.parser.pos + 3,
	filterEnd = .....
```

This applies to all changes throughout this PR

core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
```diff
@@ -42,21 +52,21 @@ exports.parse = function() {
 	var node = {
 		type: "list",
 		attributes: {
-			filter: {type: "string", value: filter}
+			filter: {type: "string", value: filter, start: filterStart, end: filterEnd},
```
saqimtiaz, line 55: It might be more consistent and user friendly to use attributeStart and attributeEnd property names for all the attributes, this applies to all wikirules.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -233,13 +243,21 @@ WikiParser.prototype.parseBlock = function(terminatorRegExpString) {
...
+		}
+		$tw.utils.each(subTree, function (node) { node.rule = nextMatch.rule.name; });
+		return subTree;
 	}
 	// Treat it as a paragraph if we didn't find a block rule
 	var start = this.pos;
 	var children = this.parseInlineRun(terminatorRegExp);
 	var end = this.pos;
-	return [{type: "element", tag: "p", children: children, start: start, end: end }];
+	return [{type: "element", tag: "p", children: children, start: start, end: end, rule: null }];
```
saqimtiaz, line 260: @Jermolene flagging for your review whether there might be a more appropriate value than null for the rule property here.

core/modules/parsers/wikiparser/rules/html.js
```diff
@@ -60,6 +64,27 @@ exports.parse = function() {
...
+					tag.closeTagStart = -1;
+					break;
+				}
+				if (char === '<') break;
+				tag.closeTagStart -= 1;
+			}
+			if (tag.closeTagStart < closeTagMinPos) {
+				tag.closeTagStart = tag.end;
+			}
+		}
```
saqimtiaz, line 87: @Gk0Wk @Jermolene we should make sure that this PR does not introduce any significant performance regression.

saqimtiaz (commented):
I have reviewed this PR, and the tests look sufficient however we should make sure that there is not any significant performance regression introduced here.

There are a few code style issues that need attention.

core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
linonetwo: Does this apply while there is assignment?

Hope this can be done by eslint.

core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
linonetwo: You mean use `{type: "string", value: filter, filterStart, filterEnd}` here?

saqimtiaz: @linonetwo I must have misread the code this morning, this code is fine and needs no further changes, and this comment can be ignored.

core/modules/parsers/wikiparser/rules/filteredtranscludeblock.js
saqimtiaz: I think this would need to be resolved manually.

core/modules/parsers/wikiparser/rules/html.js
linonetwo: I think this won't cost much, because it just `while` from `>` to the `<` in `<div>`, won't step more than `<blockquote>`

But why would we need open/closeTagStart/End ? Can this feature be turn on when needed?

core/modules/parsers/wikiparser/wikiparser.js
Jermolene: My instinct would be to omit the "rule" property entirely in this case.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -209,8 +214,13 @@ WikiParser.prototype.parsePragmas = function() {
 			break;
 		}
 		// Process the pragma rule
+		var start = this.pos;
 		var subTree = nextMatch.rule.parse();
 		if(subTree.length > 0) {
+			// Set the start and end positions of the pragma rule if
+			if (subTree[0].start === undefined) subTree[0].start = start;
+			if (subTree[subTree.length - 1].end === undefined) subTree[subTree.length - 1].end = this.pos;
```
saqimtiaz, line 222: @Gk0Wk note that we always use curly braces with if statements.

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -330,7 +348,16 @@ WikiParser.prototype.parseInlineRunUnterminated = function(options) {
 			this.pos = nextMatch.matchIndex;
 		}
 		// Process the run rule
-		tree.push.apply(tree,nextMatch.rule.parse());
+		var start = this.pos;
+		var subTree = nextMatch.rule.parse();
+		// Set the start and end positions of the first and last child if they're not already set
+		if (subTree.length > 0) {
```
Jermolene, line 354: We don't have whitespace between `if` and `(`

core/modules/parsers/wikiparser/wikiparser.js
```diff
@@ -330,7 +348,16 @@ WikiParser.prototype.parseInlineRunUnterminated = function(options) {
...
+			// Set the start and end positions of the first and last child if they're not already set
+			if (subTree[0].start === undefined) subTree[0].start = start;
```
Jermolene, line 356: We always use braces and a new line for the body of if statements.

core/modules/parsers/wikiparser/rules/html.js
```diff
@@ -60,6 +64,27 @@ exports.parse = function() {
...
+			tag.closeTagStart = tag.closeTagEnd - 2;
+			var closeTagMinPos = tag.children.length > 0 ? tag.children[tag.children.length-1].end : tag.openTagEnd;
+			if (!Number.isSafeInteger(closeTagMinPos)) closeTagMinPos = tag.openTagEnd;
+			while (tag.closeTagStart >= closeTagMinPos) {
+				var char = this.parser.source[tag.closeTagStart];
+				if (char === '>') {
+					tag.closeTagStart = -1;
+					break;
+				}
+				if (char === '<') break;
```
Jermolene, line 81: We always use double quotes for string constants.

Jermolene (commented):
Thanks @Gk0Wk much appreciated. There are a few coding style issues that need cleaning up, and then I am keen to merge this as soon as possible.

Jermolene:
Merged in e4c682d04b5627b148c86ad9ae5c7ce97b84a5f2, with some minor tweaks in 1a57d08feb88c73a321a3914a667e3319190ccc1

core/modules/parsers/wikiparser/wikiparser.js
linonetwo: This seem to be a bad idea now, because I need this info to restore \n\n in #8258

PR #7869, opened 2023-12-02
Make $scrollable less jumpy

yaisog:
The setTimeout in the code for the position update could lead to jumpy behavior when doing multiple short scroll actions, because a delayed update could interfere with the next manual scroll action.

### DIFF
--- a/core/modules/widgets/scrollable.js
+++ b/core/modules/widgets/scrollable.js
@@ -255,7 +255,7 @@ ScrollableWidget.prototype.refresh = function(changedTiddlers) {
 	}
 	// If a new scroll position was written into the tiddler, update scroll position
 	if(changedTiddlers[this.getAttribute("bind")]) {
-		setTimeout(this.updateScrollPositionFromBoundTiddler.bind(this),50);
+		this.updateScrollPositionFromBoundTiddler.bind(this);
 	}
 	return this.refreshChildren(changedTiddlers);
 };


yaisog:
Hi @Jermolene, the `setTimeout` must remain in the code that handles when the bound tiddler is changed, because otherwise it does not update the scroll position on screen. I assume this is due to elements needing re-rendering, so that they might not yet be shown when the update fires?

However, the 50 ms delay seems a bit arbitrary. I copied it from your original `render` function where you describe it as a "delay for rendering". I wonder if this might be problematic for slower machines. Would it be sensible to make use of the `requestAnimationFrame` infrastructure that you created elsewhere in the widget code or is this something unrelated?

Jermolene:
Thanks @yaisog I'll push a slightly different fix in a moment that entirely gets rid of the setTimeouts. The key is that in the refresh method we shouldn't try to scroll until the children have been refreshed.

Jermolene:
Hi @yaisog 

> 1. Somewhere in the docs for $scrollable, recommend using a bound tiddler in the `$:/temp/volatile` namespace. This avoids a lot of refreshing during scroll operations, which can occur even with the debounce.

This would need to be worded carefully to explain the narrow conditions under which it is an appropriate optimisation. For example, it would break things if the bound tiddler were being directly or indirectly rendered, which does seem to be the majority of use cases.

> 2. Make the debounce interval customizable, similar to _$:/config/Drafts/TypingTimeout_, for basically the same reasons that the latter exists. A default of 100 ms seems reasonable on my system, but might vary depending on OS or whether smooth scrolling is active, etc.

Rather than being a global option, in this case it should be an attribute on the scrollable widget. I've no objection to adding that feature but perhaps we should defer it until after the release.

yaisog:
Hi @Jermolene,

1. You're probably right. I'll defer to your judgment. I'm not sure how it would "break" things, though. It'll just delay the visual update, same as the debounce does, just by a different amount of time.
2. My reasoning was that the best setting seems user-dependent, so it would be awkward if I had to set this manually in multiple places (each $scrollable that uses `bind`). A creator could of course manually tie this to a config tiddler with e.g. `debounce={{$:/config/Drafts/TypingTimeout}}`, where `debounce` is a possible future attribute. I do agree that we should push an implementation to a later release, to allow for more time for testing.

I will close this PR, then.

yaisog:
This has been completed via hotfix by @Jermolene.
A discussion about future improvements has been initiated in #7870.

PR #7873, opened 2023-12-04
Improve `jsonset` operator docs

mateuszwilczek:
Improve the documentation of the `jsonset` operator in the prerelease docs:

* use more uniform nomenclature (operator parameter, JSON object)
* move examples to a separate tiddler

### DIFF
--- a/editions/tw5.com/tiddlers/filters/examples/jsonset.tid
+++ b/editions/tw5.com/tiddlers/filters/examples/jsonset.tid
@@ -1,0 +1,59 @@
+created: 20231204112944341
+modified: 20231204115056732
+tags: [[Operator Examples]] [[jsonset Operator]]
+title: jsonset Operator (Examples)
+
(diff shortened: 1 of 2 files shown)


Jermolene:
Looks good thank you @mateuszwilczek

mateuszwilczek:
@Jermolene if you're the main author of this operator (I presume so from #7742), please take a look if my additions to the docs make sense. I tried to explain rules behind what's happening in some of the examples, but I'm not sure i got it right. I mean the following situations in particular:

* multiple objects as input (the last example)
* appending a nonexistent top level property `[<object-b>jsonset[id],[Antelope]]`
* ignoring a nonexistent nested property `[<object-b>jsonset[missing],[id],[Antelope]]`

If I've got it right, then this is ready to merge.

Jermolene:
Great, thank you @mateuszwilczek

PR #7874, opened 2023-12-05
Docs: Improve `.from-version` macro

mateuszwilczek:
As discussed in this thread: https://talk.tiddlywiki.org/t/are-stringify-and-jsonstringify-operators-duplicates/7760/13, quote from Jeremy:

>In `<<.from-version "5.3.2">>` we could change the annotation text and styling depending on the version number. We could highlight the most recent release with a bright “NEW IN v5.3.2” badge, but mark older versions with less eye catching “introduced in v5.2.3” text.

I have decided to set the threshold of what is considered "new" manually. Highlighting only the last release as new would be more difficult than simply comparing against an arbitrary threshold. It might also not make sense in all situations, e.g. if there is a bugfix release immediately after another release, it makes no sense to mark just the bugfix as new. I have set this threshold to v5.3.0 for now.

The colors of the new badge are the same as in the `.link-badge-added` macro. I imagine it could be dynamically based on primary or highlight colors from current palette, but then it we'd introduce new color combinations for documentation elements. 

---

### DIFF
--- a/editions/tw5.com/tiddlers/system/doc-styles.tid
+++ b/editions/tw5.com/tiddlers/system/doc-styles.tid
@@ -1,5 +1,5 @@
 created: 20150117152612000
-modified: 20230325101137075
+modified: 20240223123123497
 tags: $:/tags/Stylesheet
 title: $:/editions/tw5.com/doc-styles
 type: text/vnd.tiddlywiki
@@ -181,18 +181,18 @@ tr.doc-table-subheading {
 	fill: <<colour alert-highlight>>;
 }
 
-a.doc-from-version.tc-tiddlylink {
-    display: inline-block;
-    border-radius: 1em;
-	background: <<colour muted-foreground>>;
-	color: <<colour background>>;
-	fill: <<colour background>>;
-    padding: 0 0.4em;
-    font-size: 0.7em;
-    text-transform: uppercase;
+a.doc-from-version {
+    background-color: <<colour muted-foreground>>;
+    color: <$wikify name="background" text="<<colour muted-foreground>>" mode="inline"><$transclude $variable="contrastcolour" target=<<background>> colourA="#000000" colourB="#ffffff" /></$wikify>;
+    padding: 3px;
+    border-radius: 4px;
     font-weight: bold;
-    line-height: 1.5;
-    vertical-align: text-bottom;
+    font-size: 0.75em;
+}
+
+a.doc-from-version.doc-from-version-new {
+    background-color: <<colour highlight-background>>;
+    color: <<colour highlight-foreground>>;
 }
 
 a.doc-deprecated-version.tc-tiddlylink {
--- a/editions/tw5.com/tiddlers/system/doc-styles.tid
+++ b/editions/tw5.com/tiddlers/system/doc-styles.tid
@@ -195,6 +195,11 @@ a.doc-from-version.doc-from-version-new {
     color: <<colour highlight-foreground>>;
 }
 
+a.doc-from-version svg {
+    fill: currentColor;
+    vertical-align: sub;
+}
+
 a.doc-deprecated-version.tc-tiddlylink {
     display: inline-block;
     border-radius: 1em;
--- a/editions/tw5.com/tiddlers/system/version-macros.tid
+++ b/editions/tw5.com/tiddlers/system/version-macros.tid
@@ -1,12 +1,19 @@
 code-body: yes
 created: 20161008085627406
-modified: 20221007122259593
+modified: 20231206135257498
 tags: $:/tags/Macro
 title: $:/editions/tw5.com/version-macros
 type: text/vnd.tiddlywiki
 
-\define .from-version(version)
-<$link to={{{ [<__version__>addprefix[Release ]] }}} class="doc-from-version">{{$:/core/images/warning}} New in: <$text text=<<__version__>>/></$link>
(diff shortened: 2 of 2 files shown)


mateuszwilczek:
Before:
image

After:
image

editions/tw5.com/tiddlers/system/doc-styles.tid
```diff
@@ -189,12 +189,17 @@ a.doc-from-version.tc-tiddlylink {
...
     padding: 0 0.4em;
     font-size: 0.7em;
-    text-transform: uppercase;
     font-weight: bold;
     line-height: 1.5;
     vertical-align: text-bottom;
 }
 
+a.doc-from-version.doc-from-version-new.tc-tiddlylink {
+	background: #ffe246;
```
pmario, line 198: This is not possible, since it needs to use a palette value. I also do not like the specificity of the selector. IMO it can be much less specific.

editions/tw5.com/tiddlers/system/version-macros.tid
```diff
@@ -1,12 +1,19 @@
...
 type: text/vnd.tiddlywiki
 
-\define .from-version(version)
-<$link to={{{ [<__version__>addprefix[Release ]] }}} class="doc-from-version">{{$:/core/images/warning}} New in: <$text text=<<__version__>>/></$link>
+\procedure .from-version-template(class, text)
+<$link to={{{ [<version>addprefix[Release ]] }}} class=<<class>> >{{$:/core/images/info-button}} <<text>><$text text=<<version>>/></$link>
+\end
+
+\procedure .from-version(version)
+<$let lastNewVersion="5.3.0">
```
pmario, line 13: IMO there should be **no** manual modification needed, when versions change. So I think that does not really work out. TW has a `<<version>>` macro, which can be used to compare between versions.

editions/tw5.com/tiddlers/system/doc-styles.tid
mateuszwilczek: I'll see how it looks with highlight or some other palette colors.
Side note: if we aim to base all docs styles on palette values only there is surely a lot of work in other places.

editions/tw5.com/tiddlers/system/version-macros.tid
mateuszwilczek: What programmatic criteria about what is "new" do you suggest?

* Only the latest - less informative in situations such as now, where we have a bunch of new things from v5.3.0, but the bugfix v5.3.1 is the latest.
* Latest minor release - informative now, since v5.3.0 introduces a lot of new things, but might get less informative over time if we stay on v5.3.x for a long time.
* A fixed number of releases?
* A fixed time between releases? E.g. everything released in the last 6 months or so. "Release" tiddlers have a `released` field, so this should be easily doable.

pmario: A simpler version could be to use a global variable like.

```
\procedure .from-version-reference() 5.3.0
```

So it would be more obvious what it is used for. It still would be a manual change an extra maintenence burden. I'm not sure how often it would be needed to change that global variable

pmario:
IMO we should also change all the macros to use `\procedure` instead of mixing them with `\define`. 

@Jermolene 
May be the next TW version should be one with a focus to change the core code to v5.3.x spec

pmario:
I just saw this: https://tiddlywiki.com/prerelease/#TiddlyWiki5%20Versioning -- But still the version-macro should have a `format` option

mateuszwilczek:
image

* remove icons and use Sentence case for better readability
* make "badge" style consistent with .link-badge macros used in release notes
* use palette colors -- the problem is, I can't find any color combination that would look consistently readable both in vanilla and in all other palettes.

mateuszwilczek:
@Jermolene I hope it is acceptable now. I think that in any case it is more readable and informative than the current design.

* the "Introduced in" badge uses `muted-foreground` as background, and a contrasting color as foreground (either black or white, determined by the contrastcolour macro as you suggested)
* the "New in" badge uses highlight background and foreground colors, these should be contrasting in any palette.

I have verified that it looks readable on all core palettes.

pmario:
hmm, why did you remove the icons?

The warning sign may not be ideal, but I think there should be an icon, to make it stand out more for colour blind users. 

May be we could use the info-button $:/core/images/info-button icon.

mateuszwilczek:
I have restored the icons, but changed them to `$:/core/images/info-button`, as suggested, I too think that the previous warning icon is a bit too strong in this situation.

The current look:

image

---

@CrossEye I'm tagging you because I recall you have started the discussion about these version badges.

pmario:
Here are some tiddlers with "old" and "new" badges. I think it looks OK.

https://tiddlywiki5-plfldbez7-jermolene.vercel.app/#BrowseWidget:BrowseWidget%20DroppableWidget%20%5B%5Bjsonset%20Operator%5D%5D

Jermolene:
Thanks @mateuszwilczek I'm open to further tweaks but I think we can merge this now.

CrossEye:
@mateuszwilczek:
> @CrossEye I'm tagging you because I recall you have started the discussion about these version badges.

Yes, I did.  And I haven't been following along, but the latest version looks great to me!

@Jermolene:
> [I]t might be interesting to explore a global way to expose all JS macros. One possibility might be to allow them to be invoked with the "function" operator:

:100:

That would be fantastic!

PR #7877, opened 2023-12-07
Move list-join example onto single line

### LINKED ISSUE #7860: [BUG] List join example in 5.3.2 release notes is less than ideal
### Describe the bug

In https://github.com/Jermolene/TiddlyWiki5/pull/7694#issuecomment-1826641493, it was pointed out that the list-join example I put into the 5.3.2 release notes produces output like "TiddlerOne , Tiddler Two , TiddlerThree", i.e. there appears to be an extraneous space between the list item and the joining comma.

This is less than ideal for showing off the feature, because most people are going to expect the joiner to be flush with the preceding text.

### Expected behavior

The output would look like "TiddlerOne, Tiddler Two, TiddlerThree" with no space between the tiddler titles and the following comma.

### To Reproduce

2. Scroll down to the "Joiners for the ListWidget" section
3. Copy the `join=", "` example into a new tiddler

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version; 5.2.3-prerelease
- Saving mechanism: N/A
- Plugins installed: N/A

- OS: N/A
- Browser: N/A
- Version: N/A

### Additional context

There are several things we might do. Add a `\whitespace trim` to the example in the release notes, pull the closing `</$list>` tag up one line so it's immediately following the `<$text>` widget with no adjoining whitespace, make a note about whitespace in the release notes or in the ListWidget docs... It might be best to have a discussion about what solution looks best.

rmunn:
It's a little less readable this way, but avoids the whitespace issue.

Fixes #7860.

### DIFF
--- a/editions/prerelease/tiddlers/Release 5.3.2.tid
+++ b/editions/prerelease/tiddlers/Release 5.3.2.tid
@@ -62,18 +62,13 @@ Note that the <<.attr "emptyMessage">> and <<.attr "template">> attributes take 
 You can replace it with:
 
 ```
-<$list filter=<<filter>> variable="item" join=", ">
-<$text text=<<item>>/>
-</$list>
+<$list filter=<<filter>> variable="item" join=", "><$text text=<<item>>/></$list>
 ```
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @rmunn

PR #7880, opened 2023-12-10
Rewrite tag-picker macros as procedures

etardiff:
This PR attempts to modernize the tag-picker macros as procedures, as @Jermolene suggested in [#7548 (comment)](https://github.com/Jermolene/TiddlyWiki5/pull/7548#issuecomment-1825458523). I'm hoping it may allow @pmario to revisit his tag-picker extensions as proposed in #7548.

What I changed:

* rewrote macros as procedures/replaced text substitutions with variables
* replaced nested $set and $var widgets with $let widgets/functions
* minor rewrites to reduce redundant code
* some additional linebreaks for clarity

What I have *not* (yet?) done: 

* replace any $reveal or $list widgets with 5.3.2 conditional syntax

This is both my first time using GitHub and my first time suggesting a core change; sorry for any errors! I'm submitting as a draft; please let me know if further changes are required to bring it up to the current standard.


core/wiki/macros/tag-picker.tid
```diff
@@ -36,142 +37,131 @@ second-search-filter: [tags[]is[system]search:title<userInput>sort[]]
 
 \define clear-tags-actions()
```
pmario, line 38: It seems this one did escape ;) It's still a `\define`

pmario:
>This is both my first time using GitHub and my first time suggesting a core change; sorry for any errors! I'm submitting as a draft; please let me know if further changes are required to bring it up to the current standard.

Hi @etardiff -- Welcome!
I think you should be proud ;)
I've been impressed by the quality of your responses at the Talk discussion group.

I did just skim the code and "visually" it looks good. It seems you did use the latest wikitext coding style, which IMO improves readability quite a bit. It also helps to avoid extra long lines.

I think, I did see 2 redundant $let's, but I'll need to run some tests to be sure. 

All in all it looks good.

Jermolene:
Thank you @etardiff this looks exemplary so far, and nicely idiomatic with the use of functions.

core/wiki/macros/tag-picker.tid
etardiff: Oops, good catch!

etardiff:
@Jermolene, @pmario - That's very nice to hear!

> I think, I did see 2 redundant $let's, but I'll need to run some tests to be sure.

I did intentionally use a few "unnecessary" $lets---`<$let tag=<<tag>>>` and `<$let tag=<<userInput>>>`, both referring to the functions I'd defined. I've been assuming that it's marginally faster to use $let to define a variable that I'm going to reuse within a scope than to reevaluate the function each time it's needed, but I didn't test it in this case. Happy to remove if needed. :)

@Jermolene, if this reflects the sort of changes you were looking for, I'd be glad to keep working on updating core macros to the 5.3.1+ standards. Is there an internal priority list, or should I just start with whatever looks messiest?

pmario:
@etardiff ... I'm not sure why you added the colour-picker to the PR. IMO there should be only 1 bigger change like this in 1 PR. So the colour-picker should be its own PR.

etardiff:
@pmario - Sorry, I didn't mean to do that! As I said, I'm still trying to figure out how Github works; I didn't realize that changing a different file on my fork would add it to my existing PR. It it possible to split or remove it?

PR #7882, opened 2023-12-11
Introduced preliminary idea for infinite recurse exception

flibbles:
This is a rough idea of how transclusion recursion could be handled in a way that's more robust (and allow things like `{{}} {{}}` and `<$transclude /> <$transclude />` to not break your browser instance. This isn't complete. I'm mostly looking for feedback right now.

The idea is to throw an exception when the maximum depth is exceeded. Then catch that exception hopefully at the top of where the infinite looping began (but not higher).

1. I would make TranscludeRecursionError properly inherit from Error like it should
2. I would think custom TiddlyWiki errors should have their own module type and not use the `utils` moduletype
3. This current system I use checks transclusion signature from when the depth was exceeded with the signature at every step up until it finds a mismatch. This is more robust than what currently exists, but it's not perfect yet. Recursion loops with multiple different tiddlers in the loop would fool this, but I'm just keeping this simple for the moment.
4. I'm not sure where I'm catching the error is the most appropriate place.

It's still possible for inifinite loops like this accumulate garbage because all execution and rendering are interleaved together. So when the loop is detected while executing a transclude widget, we've already rendered all the transclude widgets above and maybe `insertBefore`ed a bunch of stuff into parent. Oh well. I'd suggest separating execution and rendering so that such side-effects couldn't occur, but I think that'd be a backward-compatibility nightmare. This at least isn't crashing the browser.

If you guys like this, I'd make a few more tests for some crazier cases, *handle* those crazier cases, and then finalize the error type.

Everyone's thoughts?

(And if this this goes through, can we revert #7768? `{{}}` being `<$transclude />` was expected behavior for me. May have been for some others.)


flibbles:
Also, I'll look into how that error rethrowing effects the stack trace of other errors. I remember javascript being really bad about maintaining those in rethrows.

flibbles:
I've created a simple heuristic to better determine where the top of the loop is. It looks at the bottom 50 transclude signatures and assumes those are the tiddlers involved in the loop.

* The advantage to this is the "infinite loop" error shows up in your tiddler's body in the correct place without blasting everything in the tiddler. This might make it easier to find what transclusion is finding the loop.
* The disadvantage is I can still think of some super-weird edge cases where this would fail. But it would work 99% of the time. Also, despite my comments, the logic of what I'm doing may seem strange to other developers.

Alternatively, we could set some variable at the ViewTemplate/body level, like `inTiddlerBody` to `yes`, then all we have to do is climb up until that isn't true anymore and put the error there. This may be even more foolproof, but I'm not as fond of it. It involves changing the ViewTemplates, which is not a great idea if there's any chance this is being snuck into v5.3.2.

Still not sure where to put the TranscludeRecursionError, or what the most TiddlyWiki way to define it is.

@pmario, I'll try and work on this quickly if you guys give me any feedback. I understand your concern that *some* fix for the `{{}} {{}}` needs to be in v5.3.2.

flibbles:
I uh... don't know why that test is failing on github. It works on my machine.

pmario:
I did test it and it seems to work. 

>I've created a simple heuristic to better determine where the top of the loop is. It looks at the bottom 50 transclude signatures and assumes those are the tiddlers involved in the loop.

I'm not really sure, what's the advantage of this one in difference to the very first version.

flibbles:
> I'm not really sure, what's the advantage of this one in difference to the very first version.

I could certainly go back to the first version. It was simpler, and still better than what currently exists, and would likely catch 90%+ of the infinite loops end-users might cause. The reason for the new one is to detect more types of loops. The first version could only detect something like this:
```
{{CurrentTiddler|$:/core/ui/ViewTemplate/body/default}}
{{ImportedTiddler|CustomTemplate}}
{{TiddlerA||}}
{{TiddlerA||}}
{{TiddlerA||}}
{{TiddlerA||}}
...
```
As in, it could only detect a loop that was a single node long. The testing framework already had an example of a scenario which was more like this:
```
{{CurrentTiddler|$:/core/ui/ViewTemplate/body/default}}
{{ImportedTiddler|CustomTemplate}}
{{TiddlerA||}}
{{TiddlerB||}}
{{TiddlerA||}}
{{TiddlerB||}}
...
```
This loop is two nodes long. The first version I wrote wouldn't trace back up this loop correctly.

Ordinarily this wouldn't be a big deal because most loops don't subdivide the way that `{{}} {{}}` does. So even if the error gets printed at the very bottom of the loop (like it has been), and then resumed rendering, the first thing the wikiparser will do is "climb back out of the pit", so to speak. But in the case of a subdividing loop, then it'll just keep exploring the pit.

My newer system does a better job of recognizing the loop and climbing the wikiparser out of the pit before resuming rendering because it can recognize loop chains of up to 50 nodes long.

It's still not a perfect system. For one, loops longer than 50 nodes will fail to climb out of the pit before resuming rendering. Also, "irrational" loops will trick this system. I use irrational here like an irrational number, where the sequence of numbers past the decimal point don't have a repeating sequence. The only "irrational" loops I can fathom involve obscure uses of the `\parameters` feature though, and even then, it'd take both an irrational and subdividing loop to freeze the browser, so I figure the system I'm proposing now is probably good enough that no one will submit another bug report for a very long time.

Alternatively, I considered a system where the loop isn't analyzed at all. I described the `inTiddlerBody` variable approach which just climbs to the very top every time. I described it more earlier.

pmario:
>My newer system does a better job of recognizing the loop and climbing the wikiparser out of the pit before resuming rendering because it can recognize loop chains of up to 50 nodes long.

OK - I see. I'll have to have a closer look at the code in single step mode. ... 

I think the real max widget depth we use in the current UI is about 200. IMO its somewhere in the right sidebar - more - tags or so.

flibbles:
@Jermolene, It's been a hot minute since I made this, so to recap. It's pretty much good to go, except maybe the way it uses a custom exception. Where I grew up, custom exceptions were expected if there was ever to be custom behavior, but this may not be the TiddlyWiki way. So I'm happy to change this to whatever you want. (Maybe at least move the exceptions to a module-type or something?)

Currently though, this solutions should deal with 99%+ all cases of infinite recursion an end-user will ever encounter, so no one will ever submit such an issue again. However, if this solution is more complicated than you'd like, it can simplify down somewhat in exchange for being somewhat less thorough.

pmario:
>Currently though, this solutions should deal with 99%+ all cases of infinite recursion an end-user will ever encounter,

That's not really true: `<$button><$transclude/>` will still brick your approach. So as I wrote at: https://github.com/Jermolene/TiddlyWiki5/issues/8174#issuecomment-2128914028 -- All 3 "fixes" need to be implemented to make the thing stable enough, to leave the preview open while typing "unfinished" code

Jermolene:
Thank you @flibbles 

> It's been a hot minute since I made this, so to recap. It's pretty much good to go, except maybe the way it uses a custom exception. Where I grew up, custom exceptions were expected if there was ever to be custom behavior, but this may not be the TiddlyWiki way. So I'm happy to change this to whatever you want. (Maybe at least move the exceptions to a module-type or something?)

I don't think TiddlyWiki has a fully formed approach to exceptions, pretty much just using them as a way to bail in the face of fatal errors. Our goal is for the code to be internally consistent, and so a consideration might be whether there are other places in the core where the same approach might be helpful.

I'll merge this now so that we get some testing in before the release.

flibbles:
> That's not really true: `<$button><$transclude/>` will still brick your approach.

@pmario, Woah. You're right. And that bricks super hard too, beyond RSoD. I never noticed that before. Yeah, that needs fixin'.

> Our goal is for the code to be internally consistent, and so a consideration might be whether there are other places in the core where the same approach might be helpful.

@Jermolene, Understood. While my experience (from my C++ days) would tell me this that is an appropriate time to use a custom exception, I realize the rest of core doesn't even so much as use the built-in `Error` type ever. It only throws strings. So right now, a full-on custom exception seems a little inconsistent. I'd love to hear what other people say, and I'm happy to implement any feedback myself, even with this already merged.

Jermolene:
> @pmario, Woah. You're right. And that bricks super hard too, beyond RSoD. I never noticed that before. Yeah, that needs fixin'.

Thank you @flibbles 

> @Jermolene, Understood. While my experience (from my C++ days) would tell me this that is an appropriate time to use a custom exception, I realize the rest of core doesn't even so much as use the built-in `Error` type ever. It only throws strings. So right now, a full-on custom exception seems a little inconsistent. I'd love to hear what other people say, and I'm happy to implement any feedback myself, even with this already merged.

The throwing strings thing is quite slapdash and I think does make debugging harder. I don't think that the core does much catching of exceptions, though, which I think is where the custom exception types would be really helpful.

It's only a hypothetical question, but what would the implementation of this PR look like if it were constrained to throw strings? Would it even be possible? If the answer is that the string throwing implementation would be much messier then it would perhaps be a strong argument for refactoring things.

flibbles:
>> @pmario, Woah. You're right. And that bricks super hard too, beyond RSoD. I never noticed that before. Yeah, that needs fixin'.
>
> Thank you @flibbles

I take it I'm late to the party? This is a well known problem by now? :-)

> It's only a hypothetical question, but what would the implementation of this PR look like if it were constrained to throw strings?

It could do that, though strings are unmodifiable, so it can't have attached properties. So in order for the $transclude widget to stop an "Infinite recursion" exception and pass up all other exceptions, it'd have to resort to string matching the error message. It's brittle, not too bad with a testing framework, but it could be done.

I guess another thing to consider is if Tiddlywiki might ever use custom exception handling any more *at all*. Like, handle "filter compile errors" differently from "filter runtime errors". Or maybe "server sync errors" might have special handling?

If the need may never arise again, then it might be worth figuring out how to make this infinite recursion exception string like everything else, otherwise, maybe a `wikiexception` module-type that fits into $tw.exceptions?

Jermolene:
> I take it I'm late to the party? This is a well known problem by now? :-)

We've known since the very early days that the recursion detection is inadequate, but only fairly recently have there been concerted efforts to mitigate the problem.

> It could do that, though strings are unmodifiable, so it can't have attached properties. So in order for the $transclude widget to stop an "Infinite recursion" exception and pass up all other exceptions, it'd have to resort to string matching the error message. It's brittle, not too bad with a testing framework, but it could be done.

I had a quick search through the TW core code and I am pretty sure that it never does anything with an exception object apart from an occasional implicit `.toString()`.

So, this is the first core code that has had to check an exception, and I think it is therefore logical that it should have to introduce a core mechanism for working with exceptions to make the code clean and robust.
 
> I guess another thing to consider is if Tiddlywiki might ever use custom exception handling any more _at all_. Like, handle "filter compile errors" differently from "filter runtime errors". Or maybe "server sync errors" might have special handling?

I suspect the other places where we might want to use custom exceptions in the future would all be similar to this: situations where we want to bail early from infinite/unreasonable recursion. Given that there remains a few

> If the need may never arise again, then it might be worth figuring out how to make this infinite recursion exception string like everything else, otherwise, maybe a `wikiexception` module-type that fits into $tw.exceptions?

Jermolene:
Sorry I pressed send too early, I was going to see that given that there remain quite a few ways to create infinite loops in the core I wouldn't be surprised if we did call on this technique again. I don't think we need to refactor this PR though, it takes a reasonable approach for the moment.

PR #2435, opened 2016-05-11
Replace spaces with margins in editor toolbar

nameanyone:
As discussed in #2429.

**Before:**
2016-05-11_01-55-46__before
**After:**
2016-05-11_01-59-07__after

PR #7888, opened 2023-12-14
fix edit toolbar button whitespace

pmario:
@Jermolene It's not really clear, where the new problem comes from. 

----

This PR fixes: fix edit toolbar button whitespace

Image from: https://talk.tiddlywiki.org/t/announcing-the-release-of-tiddlywiki-v5-3-2/8661/4?u=pmario

9de7de28f56a130d1dc52f17fa148a5650ee1796_2_1035x237

Problem as shown at: https://talk.tiddlywiki.org/t/announcing-the-release-of-tiddlywiki-v5-3-2/8661/5?u=pmario

d4e841b88c52eee528ba6600a98aa005458781fa

-------

**fixed**

image

### DIFF
--- a/core/ui/EditTemplate/body-editor.tid
+++ b/core/ui/EditTemplate/body-editor.tid
@@ -3,42 +3,29 @@ title: $:/core/ui/EditTemplate/body/editor
 \whitespace trim
 
 <$edit
-
-  field="text"
-  class="tc-edit-texteditor tc-edit-texteditor-body"
-  placeholder={{$:/language/EditTemplate/Body/Placeholder}}
-  tabindex={{$:/config/EditTabIndex}}
-  focus={{{ [{$:/config/AutoFocus}match[text]then[true]] ~[[false]] }}}
-  cancelPopups="yes"
-  fileDrop={{{ [{$:/config/DragAndDrop/Enable}match[no]] :else[subfilter{$:/config/Editor/EnableImportFilter}then[yes]else[no]] }}}
-
-><$set
-
-  name="targetTiddler"
-  value=<<currentTiddler>>
-
-><$list
-
-  filter="[all[shadows+tiddlers]tag[$:/tags/EditorToolbar]!has[draft.of]]"
-
-><$reveal
-
-  type="nomatch"
-  state=<<config-visibility-title>>
-  text="hide"
-  class="tc-text-editor-toolbar-item-wrapper"
-
-><$transclude
-
-  tiddler="$:/core/ui/EditTemplate/body/toolbar/button"
-  mode="inline"
-
-/></$reveal></$list><$list
-
-  filter="[all[shadows+tiddlers]tag[$:/tags/EditorTools]!has[draft.of]]"
-
-><$list
-	filter={{!!condition}}
-	variable="list-condition"
-><$transclude/>
-</$list></$list></$set></$edit>
+	field="text"
+	class="tc-edit-texteditor tc-edit-texteditor-body"
+	placeholder={{$:/language/EditTemplate/Body/Placeholder}}
+	tabindex={{$:/config/EditTabIndex}}
+	focus={{{ [{$:/config/AutoFocus}match[text]then[true]] ~[[false]] }}}
+	cancelPopups="yes"
+	fileDrop={{{ [{$:/config/DragAndDrop/Enable}match[no]] :else[subfilter{$:/config/Editor/EnableImportFilter}then[yes]else[no]] }}}
+>
+	<$set name="targetTiddler" value=<<currentTiddler>> >
+		<$list filter="[all[shadows+tiddlers]tag[$:/tags/EditorToolbar]!has[draft.of]]">
+			<$reveal
+				type="nomatch"
+				state=<<config-visibility-title>>
+				text="hide"
+				class="tc-text-editor-toolbar-item-wrapper"
+			>
+				<$transclude tiddler="$:/core/ui/EditTemplate/body/toolbar/button" mode="inline"/>
+			</$reveal>
+		</$list>
+		<$list filter="[all[shadows+tiddlers]tag[$:/tags/EditorTools]!has[draft.of]]">
+			<$list filter={{!!condition}} variable="list-condition">
+				<$transclude/>
+			</$list>
+		</$list>
+	</$set>
+</$edit>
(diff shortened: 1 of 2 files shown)


pmario:
I did find the PR, that causes the issue. PR https://github.com/Jermolene/TiddlyWiki5/pull/7835 fixes a parsing problem, that may have been hiding other problems, with the current template parsing. 

I'm not sure yet. More tests are needed. 

@rmunn can you have a closer look too. To understand, what's going on with the toolbar button rendering.

Jermolene:
Thanks @pmario. #7835 was quite an egregious bug: whitespace was being silently stripped from the start of tiddler text. Where it gets problematic is that that bug was what allowed tiddlers like `$:/core/ui/EditTemplate/body/toolbar/button` to have an extra blank line separating the final pragma from the text of the tiddler. So perhaps a better fix for the problem might be to explicitly strip any whitespace immediately following one or more pragmas.

pmario:
I think this PR  **Replace spaces with margins in editor toolbar #2435**  was using the "hidden parser problem" to fix the whitespace. 

After the real problem was fixed the the "workaround" creeped up again. 

Just a guess. More tests needed.

pmario:
The code from **Replace spaces with margins in editor toolbar** https://github.com/Jermolene/TiddlyWiki5/pull/2435 is completely unreadable.

So readability in $:/core/ui/EditTemplate/body/toolbar/button and $:/core/ui/EditTemplate/body/editor should be fixed anyway.

Jermolene:
Hi @pmario just to be clear, this problem will potentially affect any tiddler that has pragmas preceding the content. Therefore we have to revise the fix #7835, and not just fix the particular example in the editor toolbar.

pmario:
>Hi @pmario just to be clear, this problem will potentially affect any tiddler that has pragmas preceding the content. Therefore we have to revise the fix https://github.com/Jermolene/TiddlyWiki5/pull/7835, and not just fix the particular example in the editor toolbar.

I know. I'll have a look atm

pmario:
>[. . .] So perhaps a better fix for the problem might be to explicitly strip any whitespace immediately following one or more pragmas.

That's also not possible, since for tiddlers it always has been allowed to start with as many leading new-lines as wanted, even if there are no pragmas at all.

This leading whitespace has always been "eaten". ... 

**Hmmm**
For me it looks like as if we will need to revert #7835 because of a backwards compatibility problems. -- At least **I do not have a good idea atm**

----------

If the list join-parameter needs leading spaces, we'll need to use proper templates to create a separator. Which imo contradicts the simplicity aspect a bit :/

```
<$list filter="[tag[About]sort[title]]" join="&nbsp;, "/>""" variable="item"><<item>></$list>

or

<$list filter="[tag[About]sort[title]]" join="""<$text text=" , "/>""" variable="item"><<item>></$list>

or

<$list filter="[tag[About]sort[title]]" join="""<$text text=" and " />""" ><$link/></$list>
```

Jermolene:
Hi @pmario my suggestion [above](https://github.com/Jermolene/TiddlyWiki5/pull/7888#issuecomment-1855622798) is to revert to stripping whitespace separating pragmas from the tiddler body, but continue to not strip whitespace at the start of tiddlers that do not have any pragmas.

pmario:
>... but continue to not strip whitespace at the start of tiddlers that do not have any pragmas.

Yea, but imo that's not possible. We do have several tiddlers that are used as templates, that do have leading whitespace, but no pragmas. The will start to create the same problem as the edit toolbar buttons do now.

pmario:
The edit toolbar buttons as they are rendered atm do contain 3 new-lines.

image

Jermolene:
> Yea, but imo that's not possible. We do have several tiddlers that are used as templates, that do have leading whitespace, but no pragmas. The will start to create the same problem as the edit toolbar buttons do now.

Can you give some examples? I'm not sure that it would be a common occurrence.

Jermolene:
I made a quick hack in boot.js to detect the core tiddlers that have leading whitespace. There are only two:

* $:/core/ui/MoreSideBar/Plugins
* $:/core/macros/toc

I don't think there will be any impact with either of them.

Jermolene:
Hi @pmario 

> Especially if I think about usecases, where the whole TW content including their templates are created using tools like python or bash scripts. Till now, there was no need to take care of leading spaces or new-lines. I think this could be a problem.

The problem only occurs under very specific conditions: the tiddler has to be transcluded in inline mode, and the whitespace needs to have a visible impact. Very often it will just be merged into adjacent whitespace because of the parsing rules of HTML.

In any case, none of that is an argument against my proposal; it's an argument for entirely reverting the change and going back to the old behaviour. I don't think that is the right course because the whitespace stripping behaviour is an egregious bug. It needs to be fixed, and not indefinitely tolerated.

Perhaps I would argue differently if we get more reports of the impact of #7835, but at the moment they seem to me to be minimal.

flibbles:
I have come here from an Uglify bug report: https://github.com/flibbles/tw5-uglify/issues/12

Uglify is having trouble because of #7835 too, though I'm still investigating why. Just thought I'd add now before this gets merged that I think @Jermolene's proposed fix seems the most rational to me.

If someone has a tiddler with whitepace at the start, like
```
   and
```
It's very likely that whitespace is deliberate. While if they have whitespace separating pragma from the body, like
```
\define myMacro() ...

And here's my body
```
It's likely just formatting, and TiddlyWiki should probably cut it.

I might chime in again once I know why that PR is causing Uglify trouble. Probably something to do with how Uglify likes to remove `\whitespace trim` and simply trim the whitespace explicitly, but it can't if whitespace sneaks in between a pragma and a body. Somehow (and I don't know how), this is causing TiddlyWiki to fail to inject style attributes into rendered Tiddler `div`s.

flibbles:
Yeah, here we go. Here's a concrete example of something #7835 is changing
```
<$list filter="" emptyMessage="""
\define macro() Macro
The list is empty""" />

```
The emptyMessage is getting an extra newline character before it's output which can only be removed by including `\whitespace trim`. If an emptyMessage has any pragma, it becomes impossible not to have a newline before "The list is empty" (without a `\whitespace trim` anyway).

I still agree that if there isn't pragma, but there is leading whitespace, that whitespace is most likely deliberate, and should be kept.

Jermolene:
There's now a PR for the fix I suggested: #7895

pmario:
Close this one in favour of https://github.com/Jermolene/TiddlyWiki5/pull/7895

PR #7889, opened 2023-12-15
Fix spacing in Standard and Filter tabs of Advanced Search

### LINKED ISSUE #7884: [BUG] Spacing of standard search results in `$:/AdvancedSearch`
### Describe the bug

There is no spacing between the text "Title matches:" and "All matches:" in the Standard tab of `$:/AdvancedSearch` if a query that has no title matches is entered. Sidebar search results are not affected.

### Expected behavior

There should be spacing between "Title matches:" and "All matches:" in all cases, even if there are no title matches.

### To Reproduce

1. Go to [tiddlywiki.com](https://tiddlywiki.com)
2. Open `$:/AdvancedSearch`
3. Open Standard tab
4. Type a query that gives no title matches (e.g. `aaa`) or no matches at all (e.g. `rrr`).

### Screenshots

Sidebar search with proper spacing, even if no title matches found:
sb

Advanced Search without adequate spacing if no title matches found:
as

### TiddlyWiki Configuration

Seems to be in all versions since v5.1.5 (where distinction between title and all matches was introduced) up to v5.3.2-pre where distinction between title and all matches was introduced.

### Additional context

In the sidebar, each caption and each result list (if present) is a separate `<p>` element.
In Advanced Search, everything (both captions and all results) are contained in a single `<p>` element.

mateuszwilczek:
The modified tiddlers are too complicated for me to understand in the whole, I only analysed what has to be done to fix the spacing issues. I don't know if this PR can have some unexpected consequences somewhere other than Advanced Search.

## Standard

Fix #7884 :

* set `$mode="block"` on the transclusion of `$:/core/ui/DefaultSearchResultList`, because inline transclusion causes #7884,
* move the transclusion to the `<$list-empty>` widget (thanks v5.3.2 😃), because a block transclusion inside the `emptyMessage` parameter would break things.

Before (notice no spacing between "_Title matches:All matches:_"):
image

After:
image

## Filter

Make spacing in the Filter tab consistent with other Advanced Search tab by removing the last `\whitespace trim` and adding whitespace around `<<lingo Filter/Matches>>`.

Before (notice no margin below "_N matches_"):
image

After (spacing of "_N matches_" same as for analogue elements in other Advanced Search tabs):
image


pmario:
Did you try it with an `emptyMessage="n/a"` for the list-widgets, or something similar, instead of removing the `\whitespace trim`?

It is needed to enable the uglify-plugin.

mateuszwilczek:
@pmario for Filter, the same effect can be achieved without removing `\whitespace trim` and instead wrapping the `<<lingo Filter/Matches>>` in `<p>   </p>`, but I don't know if explicit `<p>` tags are an acceptable thing.

As to the Standard and removing `emptyMessage`/ staying with old syntax; it's probably possible by introducing two complimentary list widgets instead of one with complex both content and emptyMessage. I'll look into it.

pmario:
I think only the first list needs an emptyMessage. I think the second one can stay as it is but I did not make any tests with the code.

mateuszwilczek:
@pmario I made changes to remain compatible with older syntax/uglify plugin.

Changes in `$:/core/ui/AdvancedSearch/Standard`: add `mode="block"` to a certain transclude widget to enable proper spacing. Since block transclusion does not work inside `emptyMessage` parameter and using newer `<$list-empty>` syntax is to be avoided, I decided to split the the initial list widget into two list widgets with complimentary filters (only one of them is true at a time).

Changes in `$:/core/ui/AdvancedSearch/Filter`: achieve the proper spacing using explicit `<p>` tags around the text. The `\whitespace trim` remains untouched.

mateuszwilczek:
Sorry for the mess, I forgot that the PR Maker works for docs only and not for core changes, I'll fix the PR soon.

mateuszwilczek:
I think it's ready for review now, @pmario @Jermolene 
In the Vercel preview, everything looks as intended, like in the screenshots in the first post.

Jermolene:
Thanks @mateuszwilczek

Issue #763, opened 2014-08-28
Pop-up menu opens outside visible area


braykov:
Have a wiki UI adjusted to 80% screen for the tiddlers and leaving 20% screen for the sidebar.
Have the down-arrow shown on the sidebar.
Click the down-arrow icon ("More Actions") in the sidebar.

Result: the menu opens positioned in a way that its upper left corner is just bellow the down-arrow.
This way most of the menu is not seen (placed outside the browser visible area.

The sidebar gets a horizontal scrollbar at the bottom and one can scroll manually to be able to see the full popup menu area.
Expected: There is enough space in the browser and the menu should be displayed a bit to the left so it is all visible and no scrolling is required.
![image](https://cloud.githubusercontent.com/assets/8580578/4077757/e9531cea-2ec5-11e4-9de1-83d518f73156.png)

danielo515:
>  I For wide popups, that would have the effect of moving their left position to zero. Would that behaviour achieve what you want?

That is what my modification of the popup widget does. I'ts very simple:

```
/*\
title: $:/danielo515/modules/widgets/revealNoOffset.js
type: application/javascript
module-type: widget

Reveal widget without left offset 

\*/

(function(){

/*jslint node: true, browser: true */
/*global $tw: false */
"use strict";

var reveal = require("$:/core/modules/widgets/reveal.js").reveal;


var RevealNoOffset = function(parseTreeNode,options) {
    this.initialise(parseTreeNode,options);
};


RevealNoOffset .prototype = new reveal();
RevealNoOffset .prototype.parentexecute = RevealNoOffset.prototype.execute;

RevealNoOffset.prototype.execute = function (){
this.parentexecute();
this.noOffset =  this.getAttribute("offset","left");
};

RevealNoOffset .prototype.readPopupState = function(state) {
    var popupLocationRegExp = /^\((-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+),(-?[0-9\.E]+)\)$/,
        match = popupLocationRegExp.exec(state);
    // Check if the state matches the location regexp
    if(match) {
        // If so, we're open
        this.isOpen = true;
        // Get the location
        this.popup = {
            left: this.noOffset === "left" ? 0 : parseFloat(match[1]) ,
            top: this.noOffset === "top" ? 0 : parseFloat(match[2]),
            width: parseFloat(match[3]),
            height: parseFloat(match[4])
        };
    } else {
        // If not, we're closed
[...]

PR #7898, opened 2023-12-19
Enable popup clamping to container

yaisog:
This is a demo implementation in response to #763.

If necessary, a popup is moved so that it fits completely into its parent container.

A demo tiddler, [[Popup Clamping Demo]], is included to illustrate the effect.

### DIFF
--- a/core/modules/widgets/reveal.js
+++ b/core/modules/widgets/reveal.js
@@ -96,6 +96,21 @@ RevealWidget.prototype.positionPopup = function(domNode) {
 		left = Math.max(0,left);
 		top = Math.max(0,top);
 	}
+	if(this.clampToParent) {
+		var offsetParentDomNode = domNode.offsetParent,
+			parentWidth = offsetParentDomNode.offsetWidth,
+			parentHeight = offsetParentDomNode.offsetHeight,
+			right = left + domNode.offsetWidth,
+			bottom = top + domNode.offsetHeight;
+		if(domNode.offsetWidth <= parentWidth && right > parentWidth) {
+			left = parentWidth - domNode.offsetWidth;
+		}
+		if(domNode.offsetHeight <= parentHeight && bottom > parentHeight) {
+			top = parentHeight - domNode.offsetHeight;
+		}
+		// clamping on left and top sides is taken care of by positionAllowNegative
+	}
+	// TODO: clampToParent handling for absolute positions
 	if (this.popup.absolute) {
 		// Traverse the offsetParent chain and correct the offset to make it relative to the parent node.
 		for (var offsetParentDomNode = domNode.offsetParent; offsetParentDomNode; offsetParentDomNode = offsetParentDomNode.offsetParent) {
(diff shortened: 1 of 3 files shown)


pmario:
@yaisog -- That's a very good example tiddler. Thank you very much.

pmario:
So if I do understand it right, we will need a `.container` element with the right style definitions to get the fix going for core UI dropdowns. right?

yaisog:
The popup is an absolutely positioned element. MDN says about that:

> The element is removed from the normal document flow, and no space is created for the element in the page layout. The element is positioned relative to its closest positioned ancestor (if any) or to the initial [containing block](https://developer.mozilla.org/en-US/docs/Web/CSS/Containing_block#identifying_the_containing_block). Its final position is determined by the values of top, right, bottom, and left.

So, any positioned element (or the document body) can serve as the containing block. This applies e.g. to the sidebar. I believe there are a few other elements in the standard DOM that are positioned.

I used an explicit container to illustrate the clipping / clamping without filling the whole screen with popup buttons.

yaisog:
I've been using this for over a year now and found it sensible in some situations to be able to clamp only on the right _or_ the bottom edge of the container. So the `clamp` parameter can now have the values `right`, `bottom`, `both` or `none` (the default).

I'm setting it to be ready for review.

editions/tw5.com/tiddlers/system/DefaultTiddlers.tid
```diff
@@ -3,6 +3,7 @@ modified: 20140912135951542
 title: $:/DefaultTiddlers
 type: text/vnd.tiddlywiki
 
+[[Popup Clamping Demo]]
```
pmario: @yaisog ... This change needs to be removed, otherwise it can not be merged.

pmario:
How do we need to change https://deploy-preview-7898--tiddlywiki-previews.netlify.app/#%24%3A%2Fcore%2Fui%2FSideBarSegments%2Fsearch so that the search dropdown is moved to the left if there are longer results, that are outside the window?

yaisog:
> How do we need to change ...

We'd need to change a whole lot because of the structure of the page template and how the search results dropdown is implemented:

- The dropdown is not a popup, but of `type=nomatch`, so `clamp` doesn't do anything.
- If the RevealWidget would be changed to `popup` clamping wouldn't work right, because the dropdown is wrapped in another RevealWidget with `position:relative`. That is the container for the popup, which doesn't match the container of the element that triggers it (EditWidget or ButtonWidget), so it'll be placed wrongly due to coordinate mismatch. Maybe using absolute coordinates could help, at least with the ButtonWidget? Note that this is completely independent of the changes introduced in this PR.
- Removing the outer RevealWidget helps somewhat, except now the dropdown may not always have the same position, depending on what triggers it.
- If the content is too wide, clamping doesn't help anyway because the sidebar width limits the popup width. Unless `positionAllowNegative` is used, but then the story river will partially obscure the popup.

You'd have to make the popup appear in front of the story river while remaining in the same container as the EditWidget. I don't think that's easily done.

pmario:
OK. Thanks for the info.

Jermolene:
Thanks @yaisog is this good to go?

yaisog:
> Thanks @yaisog is this good to go?

Hi @Jermolene, it is.

Jermolene:
> Hi @Jermolene, it is.

Apologies, thanks to my tardiness there is now a merge conflict.

yaisog (commented):
Hi @Jermolene, the merge conflict is resolved. You can go ahead with merging if you want to.

Jermolene:
Thanks @yaisog I think this is a useful enhancement. It's things like this that can put TiddlyWiki into an uncanny valley where it *almost* behaves like a modern web application.

Issue #7893, opened 2023-12-16
[BUG] Preview broken when `$:/config/ShowEditPreview/PerTiddler` is set to `yes`


yaisog:
### Describe the bug

Previews don't work when the above setting is active.
This is likely related to #7747. The `qualify` macro results are not correct now that is has been moved into `$:/core/ui/EditorToolbar/preview`, so the wrong state tiddler is set.

### To Reproduce

1. Go to tiddlywiki.com
2. Enable the per-tiddler preview state
3. Edit a tiddler and turn on preview

### TiddlyWiki Configuration

v5.3.2

Jermolene:
Thanks @yaisog – would you be able to have a look at this @BurningTreeC?

AnthonyMuscio:
Possibly related to this is when the above is set to no or deleted, the preview works, however with both the "widget Tree Preview" and "Raw HTML" preview you can't see the edit panel and the previews are full width.
* I think there is a bug in the flex box, or columns for different screen sizes.

I am glad to know this bug exits, but it is quite substantial, so can we post about in talk.tiddlywiki and if possible, provide a temporary patch?
* Any tips?, I will create a simple patch
* I am just trying to evaluate [$:/core/ui/EditorToolbar/preview](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FEditorToolbar%2Fpreview) and `:else[<qualify "$:/state/showeditpreview">]`

I wasted hours going crazy by working on a set of advanced previews and thought I was going crazy with all sorts of weird behaviours not knowing it was not me, it was tiddlywiki.

Just now I noticed another bug;
* Open $:/core/ui/EditorToolbar/preview
* Open the preview and select RAW Html
* Resize the browser window to get the preview along side the editor
* I get the result below in FireFox , the preview is an HTML version of a SVG not the HTML of the content in the editor.

![Snag_1ddca396](https://github.com/Jermolene/TiddlyWiki5/assets/31584658/5c122eca-6ba8-4995-95ab-1c45d5b73c85)

pmario:
@yaisog -- IMO we can close this one. The PR has been merged

AnthonyMuscio:
> Possibly related to this is when the above is set to no or deleted, the preview works, however with both the "widget Tree Preview" and "Raw HTML" preview you can't see the edit panel and the previews are full width.

[...]

Issue #7856, opened 2023-11-25
Tickets for consideration after release of v5.3.2


Jermolene:
These are the tickets that were held over from v5.3.2, and need consideration for the next release:

* #7817
  * Needs investigation of tighter unification between the testcase widget and the innerwiki widget
* #7830
  * Needs investigation of the user interface for handling automatic dark mode
* #7822 
  * Rejected; see [comment](https://github.com/Jermolene/TiddlyWiki5/pull/7822#issuecomment-1826266336)
* #7615 
  * Needs more evidence of efficacy
* #7548 
  * Rejected in favour of a deeper refactoring
* #7729
  * Needs further work; see [comment](https://github.com/Jermolene/TiddlyWiki5/pull/7729#issuecomment-1825440169)

Jermolene:
Added a link to #7895, which fixes a bug introduced in #7835

saqimtiaz:
#7906 fixes a problem whereby functions in filter runs don't use variables assigned by the filter run.

linonetwo:
Would you mind if I ask, when my PRs will get merged? They can be find using https://github.com/Jermolene/TiddlyWiki5/pulls/linonetwo , I have more PRs to come, but I hope to investigate more after this stockpile gets clean.

CrossEye:
I wasn't paying attention, and didn't see this thread until just now.  #7729 was reworked and replaced.  I believe #7790 is complete.

PR #7900, opened 2023-12-20
Fix per-tiddler previews

yaisog:
This is a fix for #7893.

Basically, it reverts the parts of #7747 that use a function in the EditTemplate in favor of setting the variable `editPreviewStateTiddler` which is also referred to in the preview button code. Otherwise, the qualify macro yields different state tiddlers and previews cannot be activated.

In one fell swoop, I also streamlined the preview button code to make use of inline filters instead of multiple RevealWidgets. This reduces it to three lines. It could be rolled back, of course, if we want to be conservative there.

### DIFF
--- a/core/ui/EditTemplate/body/default.tid
+++ b/core/ui/EditTemplate/body/default.tid
@@ -1,9 +1,5 @@
 title: $:/core/ui/EditTemplate/body/default
 
-\function edit-preview-state()
-[{$:/config/ShowEditPreview/PerTiddler}!match[yes]then[$:/state/showeditpreview]] :else[<qualify "$:/state/showeditpreview">] +[get[text]] :else[[no]]
-\end
-
 \define config-visibility-title()
 $:/config/EditorToolbarButtons/Visibility/$(currentTiddler)$
 \end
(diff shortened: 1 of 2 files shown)


Jermolene:
Thank you @yaisog I've added this to #7856 where we're tracking the fixes for 5.3.3

pmario:
>Basically, it reverts the parts of https://github.com/Jermolene/TiddlyWiki5/pull/7747 that use a function in the EditTemplate in favor of setting the variable editPreviewStateTiddler which is also referred to in the preview button code. Otherwise, the qualify macro yields different state tiddlers and previews cannot be activated.

If it does revert #7747 it also re-introduces the problem for issue #7745 with the codemirror 6 plugin. 

So I think we should not use the qualify-macro to create a state name, IMO we should use the draft-tiddler-title to create a unique state tiddler. 

Since the button and the editor preview should know this title, it should be possible to toggle the preview without redrawing the editor section. 

---------

@BurningTreeC ... Please test this preview https://tiddlywiki5-c5ljtb9p0-jermolene.vercel.app/ with the CM-6 plugin. Is the #7745 problem back with this preview version?

yaisog:
If I put a `<$log />` into `$:/core/ui/EditTemplate/body/editor`, I can see in the console that the editor is redrawn upon turning the preview on/off. The behavior is the same in the PR preview as it is on tiddlywiki.com.

Since the class on the DIV surrounding the editor changes when the preview is toggled, I think it might not be possible to prevent the child redraw.

Maybe @Jermolene solved the problem by setting the focus back to the editor at the end of the button code? That is still in there, so it might work. I guess we'll have to wait for feedback from @BurningTreeC.

PR #7911, opened 2023-12-27
Allows whitespace-only macro/procedure to be closed by \end

### LINKED ISSUE #3460: Bug: empty macro definitions don't parse as expected
The `\end` of an empty macro definition is incorrectly parsed as its body.

For example:

```
\define max()
\end
Nothing
\end

Out: <<max>>
```

The correct output should be `Nothing \end Out:`, but the current output is `Out: \end Nothing`.

btheado:
Change the regular expression used by the parser to find the end of a macro/procedure/widget/function so it can properly terminate by `\end` when the body is empty (contains nothing except whitespace).

Previously, the regular expression required matching a `\n` prior to the `\end`, but the `\n` on the `\define` line (or `\procedure`, etc.) processed by the parser.

The new regular expression makes the `\n` optional by matching either `\n` or beginning of string.

Fixes #3460.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/fnprocdef.js
+++ b/core/modules/parsers/wikiparser/rules/fnprocdef.js
@@ -49,11 +49,11 @@ exports.parse = function() {
 	if(this.match[3]) {
 		params = $tw.utils.parseParameterDefinition(this.match[4]);
 	}
-	// Is this a multiline definition?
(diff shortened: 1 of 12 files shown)


Jermolene:
Thank you @btheado looks good. I've added this PR to the list at #7856

Jermolene:
Thank you @btheado

PR #7915, opened 2024-01-02
Introduce "Multi Wiki Server" Plugin

Jermolene:
**This PR has been retired. Please see https://github.com/TiddlyWiki/MultiWikiServer for the latest version of MWS**.

---

# Introduction

This PR introduces the "Multi Wiki Server" (MWS) plugin. It adds support for hosting multiple wikis at the same time. For performance and scalability, it uses SQLite to store tiddlers.

Hosted wikis are independent of the main wiki, which is still used for administration functions:

* storing configuration for managing the hosted wikis
* generating templates for static HTML routes

# Related Work

## SQLite Browser Wasm Experiments

Last year, I worked on #7329 which involved using the Wasm version of SQLite in the browser to make an alternate implementation of the wiki store object. The current status is that everything works but the overall performance is worse than the current JavaScript store implementation. I believe that it is possible to significantly improve on the JS performance by compiling filters directly into SQL. I hope to return to that work, but in the meantime I think we can get a lot of value from SQLite on the server without requiring the ability to execute filters against tiddlers in the database.

## Bob.Exe

MWS has similar goals to @inmysocks's Bob.Exe at https://github.com/OokTech/TW5-Bob with the key differences being:

* MWS uses SQLite for storing the content of the wikis
* MWS uses the recipe/bag model for sharing content

# Progress

* [x] New tiddler schema that uses a record for each field instead of trying to pack them into a single record
* [x] `SqlTiddlerStore` class for storing and retrieving tiddlers with bags and recipes
* [x] Tests for the `SqlTiddlerStore` class
* [x] New `/wiki/:recipe_name` HTTP route that serves a TW5 wiki template with the tiddlers from a given recipe spliced in
* [x] Support for hosted wikis to synchronise via the API
* [x] Use a persistent disk-based database
* [x] Improved syncing that has persistent tiddler revision numbers, and thus can cope with server restarts
* [x] Allow configuration of the hosted wikis via the main wiki (without needing a restart of the server) [much more to do, but the basics are there]
* [x] Support for use of _canonical_uri for tiddlers >10MB
* [x] Support for fields containing large blobs that are stored in a separate content addressable file store indexed by the MD5 hash of the content of the blob. This will allow seamless handling of very large tiddlers (eg video files)
* [ ] Support for authentication and authorisation
* [ ] Consider caching the wiki template
* [x] Instant syncing from server to browser
* [ ] Fix uploading the same attachment twice – in particular, might be problematic if done via two browsers that happen to send a different content type
* [ ] Fix editing tiddlers that contain attachments
* [x] Avoid making text tiddlers become attachments
* [ ] Support for deleting attachments when the associated tiddler is modified or deleted
* [ ] Update `--mws-save-archive` to handle attachments
* [ ] Apply length limits while processing multipart form data to prevent DDOS attacks
* [ ] Support HTTP range requests to allow streaming videos to support seeking within video


pmario:
@Jermolene -- Is this PR intended to collect early feedback, or just to track your own progress?

Jermolene:
> @Jermolene -- Is this PR intended to collect early feedback, or just to track your own progress?

It's no different from any other draft PR, which is to say that it is the product of working in the open, and is published so that others can review and provide feedback.

linonetwo:
Cool, can different bags belong to different folders?

(So they can be synced to different Github repos, some of them are private bags and in private Github repo, and some public will be used to build public gh pages blog).

Jermolene:
Hi @linonetwo 

> Cool, can different bags belong to different folders?

This is not a hierarchical structure; bags are flat and do not contain other bags.

Bag can appear in multiple recipes at the same time.

> (So they can be synced to different Github repos, some of them are private bags and in private Github repo, and some public will be used to build public gh pages blog).

That is correct, there are several motivations that might apply for placing a particular set of tiddlers in a bag of their own:

* To be able to independently include those tiddlers in any other recipe (ie to reuse them in other wikis)
* To make it easier to sync those tiddlers independently
* To be able to apply access controls to those tiddlers (bags represent access control boundaries)

linonetwo:
Sorry for the confusion,

> can different bags belong to different folders?

I want to ask about the folder structure on the file system, will it still be one `tiddlers/` folder?

I know `bag` is a field on the `.tid` file or `.meta` file, so maybe all of these "bags" are still on the same `tiddlers/` folder on the file system?

Jermolene:
> I want to ask about the folder structure on the file system, will it still be one `tiddlers/` folder?

The tiddlers for the hosted wikis are stored in a SQLite database, they are not stored as files in the file system.

editions/multiwikiserver/tiddlywiki.info
```diff
@@ -0,0 +1,22 @@
+{
+	"description": "Multiple wiki client-server edition",
+	"plugins": [
+		"tiddlywiki/tiddlyweb",
+		"tiddlywiki/filesystem",
+		"tiddlywiki/highlight",
```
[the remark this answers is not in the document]
Jermolene: No that was actually just a merge error. I'll remove it.

pmario:
@Jermolene At Talk, your wrote

>- Robust authentication and authorisation options

But on GitHub "Progress" those "todo's" are not checked yet. So the question is. Is there any authentication and authorisation already. I would be pretty interested in that one.

Jermolene:
> But on GitHub "Progress" those "todo's" are not checked yet. So the question is. Is there any authentication and authorisation already. I would be pretty interested in that one.

The checklist is correct; I haven't begun work on authentication or authorisation yet.

CrossEye:
@Jermolene 
>  I haven't begun work on authentication or authorisation yet.

Do you have any expected direction here?  Are you thinking that auth will be entirely internal or would you look to use OAuth to connect to external providers?

Is this an area you might be looking for help with?  I'm not an expert, not even close, but I have been doing Node apps for years, and have had to implement several auth schemes over that time.  I'm not certain I can make the time, but if you are actively seeking help, I can try hard.

Jermolene:
> Do you have any expected direction here? Are you thinking that auth will be entirely internal or would you look to use OAuth to connect to external providers?

I think that we need to do both, and perhaps a bit more:

* An integral user database, doing whatever is necessary to meet current best practice for storing hashed passwords etc. TiddlyWiki wouldn't generally have the ability to send email and so it would be difficult to meet user expectations for being able to reset passwords etc.
* Authenticating against external identity providers via OAuth
* Use of a trusted identity header (which we already support, and is very useful when using a hosted environment like Azure)
* It would also be interesting for TiddlyWikit to be able to be an OAuth provider for other services

> Is this an area you might be looking for help with? I'm not an expert, not even close, but I have been doing Node apps for years, and have had to implement several auth schemes over that time. I'm not certain I can make the time, but if you are actively seeking help, I can try hard.

Thank you, I would very much like help here. I don't have deep implementation experience in this area.

I am hoping that we can leverage existing libraries for the difficult stuff, and to that end I am keen to try to fit in with the existing Node.js middleware ecosystem, but am hampered by my lack of familiarity with that world.

At this point in the development, I have a number of working assumptions that I am hoping hold together enough to allow me to continue to focus on other areas:

* That most of the time I can treat authentication as a black box within the request handling that produces a unique identifier for each user
* That there will be a table in the database keyed on that identifier that will hold the username and other information about the account
* That we will be able to add access control around the bag, recipe, and tiddler tables. I am hoping that there is prior art for representing Unix or NT-style ACLs within a database
* I don't have a robust model for handling groups. I know that people using, say, Active Directory would want to administer group membership within AD, and so I am expecting that I would somehow get a list of tokens representing group membership

Reading between the lines, you can probably tell that the sort of scenarios driving this are:

* A household that has a shared TiddlyWiki and wants to enable family members to have their own accounts
* A club or association website/forum that needs users to be able to login, but doesn't want to manage passwords etc, and so can offload authentication to Google/Facebook etc.
* An IT department for a large company that wants to operate an internal wiki with a few hundred users, and connect to the existing company ActiveDirectory
* An IT department for a small company that lacks much infrastructure, and wants to operate a wiki portal for a few hundred external clients

So this would certainly be a great time to start trying to flesh this stuff out more formally. My ideas are pretty inchoate at the moment so any contribution would be helpful!

hoelzro:
@Jermolene I took a quick look at the SQLite parts of this today - I'm going to look deeper, but I wanted to mention a few things I noticed in my first pass:

  * I see you're using foreign keys - strangely enough, SQLite doesn't enforce foreign keys by default :disappointed:  You need to turn them on via the [foreign_keys pragma](https://sqlite.org/pragma.html#pragma_foreign_keys).
  * I'd say the schema looks pretty good, but a few thoughts:
    * I personally put `NOT NULL` on all of my columns unless I explicitly want to allow a `NULL` there, but that's just me!
    * Since some fields like `created`, `modified`, `type`, and `text` will be on almost every tiddler, you might want to consider having nullable columns for those in the `tiddlers` table - but having them in the `fields` table _does_ definitely simplify the logic!  Another approach could be to just store extra fields in a JSON blob and use the JSON operators/functions to work with those.
    * Since the number of distinct field names is probably going to be fairly small, you _might_ want to consider having a `field_names` table in the future - it might be overkill at this point, though.
  * One thing to be aware of: SQLite has this weird feature known as [type affinity](https://www.sqlite.org/datatype3.html#type_affinity), which (for example) allows you to insert a string into a column typed as `INTEGER`.  If you're targeting a more recent version of SQLite (3.37.0 and higher) and you don't want this behavior, you can use [strict tables](https://www.sqlite.org/stricttables.html).
  * When working with `TEXT` columns, you'll want to be mindful of [collations](https://sqlite.org/datatype3.html#collation), which can cause some comparison operations and things like distinctivness checking/ordering/grouping to be case-insensitive.

Jermolene:
Thanks @hoelzro much appreciated.

> * I see you're using foreign keys - strangely enough, SQLite doesn't enforce foreign keys by default 😞  You need to turn them on via the [foreign_keys pragma](https://sqlite.org/pragma.html#pragma_foreign_keys).

Interesting. If they're not doing anything at the moment it does rather suggest that we don't need the functionality of foreign keys, and they should be removed. 

>   * I personally put `NOT NULL` on all of my columns unless I explicitly want to allow a `NULL` there, but that's just me!

Good point, fixed in c26acfdb429960ef00c1d96be41f4d6c0ceec4e6

>   * Since some fields like `created`, `modified`, `type`, and `text` will be on almost every tiddler, you might want to consider having nullable columns for those in the `tiddlers` table - but having them in the `fields` table _does_ definitely simplify the logic!  Another approach could be to just store extra fields in a JSON blob and use the JSON operators/functions to work with those.

The earlier schema that I worked on over the summer in #7329 worked along exactly those lines. At some point I realised that the only way that we're going to get decent performance is by compiling filters into SQL queries, which I think would be much easier if we have a consistent storage format that treats all fields the same way.

>   * Since the number of distinct field names is probably going to be fairly small, you _might_ want to consider having a `field_names` table in the future - it might be overkill at this point, though.

Yes, that would make sense, and if we're going to do it I'd be inclined to try to get it in now. Would you like to try your hand at a PR? It would be close to the edge of my SQL query skills.

> * One thing to be aware of: SQLite has this weird feature known as [type affinity](https://www.sqlite.org/datatype3.html#type_affinity), which (for example) allows you to insert a string into a column typed as `INTEGER`.  If you're targeting a more recent version of SQLite (3.37.0 and higher) and you don't want this behavior, you can use [strict tables](https://www.sqlite.org/stricttables.html).

I have read about that, and quite like the approach coming from a JS background. I notice that page doesn't mention a performance impact. I guess if there isn't an appreciable impact we might as well enable them to get the benefit of the error checking.

> * When working with `TEXT` columns, you'll want to be mindful of [collations](https://sqlite.org/datatype3.html#collation), which can cause some comparison operations and things like distinctivness checking/ordering/grouping to be case-insensitive.

Yes indeed, back in #7329 I had to go to the trouble of creating a custom collation function to get the tests to pass.

I think we're going to end up needing a bunch of custom functions for SQLite, which I'm hoping we would write in C and compile with SQLite.

pmario:
>>One thing to be aware of: SQLite has this weird feature known as [type affinity](https://www.sqlite.org/datatype3.html#type_affinity), which (for example) allows you to insert a string into a column typed as INTEGER. If you're targeting a more recent version of SQLite (3.37.0 and higher) and you don't want this behavior, you can use [strict tables](https://www.sqlite.org/stricttables.html).
>
>I have read about that, and quite like the approach coming from a JS background. I notice that page doesn't mention a performance impact. I guess if there isn't an appreciable impact we might as well enable them to get the benefit of the error checking.

After asking some question at talk I also did listen so some interviews of Richard Hipp the main author fo SQLite and he said that it's a feature and was intended to be that way. ... I think it was [this interview](https://www.youtube.com/watch?v=2eaQzahCeh4), but I cannot find the timestamp atm. 

So I think we should not prematurely try to "optimize" things, if we do not need to. -- Just my thoughts

hoelzro:
Yes, Dr. Hipp originally wrote SQLite for Tcl, and he wanted the type system to share Tcl's flexibility.  I figured I'd mention it since it differs from most other SQL implementations, and in my opinion it's easier to start strict and relax things later than the other way around.

Jermolene:
Hi @hoelzro @pmario I think I've somewhat hardened my position to thinking that we should opt into strict tables now, keeping our options open for opting out later if there is a performance benefit. The benefit would be presumably that we'd get earlier and more useful error messages.

CrossEye:
@Jermolene:

> An integral user database, doing whatever is necessary to meet current best practice for storing hashed passwords etc.

As I understand it, best practices seem to be trending away from passwords altogether.  This option -- which I'm guessing a large number of  users would want -- is probably the most difficult to set up.


> TiddlyWiki wouldn't generally have the ability to send email and so it would be difficult to meet user expectations for being able to reset passwords etc.

Running on Node, it may well be able to do so, to send passwords, or or texts with a short-term sign-in key.  I haven't done this, but it sounds feasible.

> Authenticating against external identity providers via OAuth

I'm guessing this would be the easiest one to implement, but it would require owners of the wiki back-ends to do some setup to get individual keys for each of the providers they decide to use.

> Use of a trusted identity header

Presumably that's also straightforward.  Where is the current supporting code?

> It would also be interesting for TiddlyWikit to be able to be an OAuth provider for other services

I have a few reactions:

  * "Damn, that sounds **hard**!"
  * "Why in the world?"
  * "Later, man... much later!"
  * "WTF?"

I'd love to hear your reasoning, because I personally would not trust someone's $5/month Node environment to securely hold my credentials so they can share them with other registered wikis.  Or are you suggesting that the main site, tiddlywiki.com serve that role?  That could be interesting.  Or was that not a simple typo, and are you talking about some ancillary TiddlyWiki tool, "TiddlyWiki**t**", that I've never heard of?


> I am hoping that we can leverage existing libraries for the difficult stuff, and to that end I am keen to try to fit in with the existing Node.js middleware ecosystem, but am hampered by my lack of familiarity with that world.

Usually when you hear about middleware related to Node, it's talking about the most common web framework for Node, [Express][ex].  I haven't really looked at the new code yet, but if you're not already using Express, I doubt it's worth bringing in just for this.  If it's already in place, then yes, middleware is a great help for auth code.


> * That most of the time I can treat authentication as a black box within the request handling that produces a unique identifier for each user

Of course, and that probably means some sort of interface that the different auth mechanisms all present, with some configuration that says, "Yes, I want a password mechanism; yes I want OAuth, with Google, GitHub, and LinkedIn providers; no, I don't need trusted headers," or whatever.

That leads me to an interesting question:  Would you want the auth configuration to be at an installation level or at a wiki level?  Or is there some level in between to consider?

> * That there will be a table in the database keyed on that identifier that will hold the username and other information about the account

That would clearly be necessary for an in-house password system.  But it might or might not be the case for external providers, AD groups, or trusted headers.

> That we will be able to add access control around the bag, recipe, and tiddler tables. I am hoping that there is prior art for representing Unix or NT-style ACLs within a database

We can look for prior art, but I think we'll have to take great care in determining the proper granularity of ACLs.  One real concern would be whether we would want to grant a user read-only access to a bag in one recipe, but read-write access to the same bag in a different recipe.  It would be easiest just to forbid this, but I think people might well come up with reasonable scenarios we couldn't support.  And that would be a shame.

----------

I'll start thinking about this and reading the current code soon.  No promises that I can contribute usefully, but I would like to try.


  [ex]: https://expressjs.com/

hoelzro:
> Interesting. If they're not doing anything at the moment it does rather suggest that we don't need the functionality of foreign keys, and they should be removed.

Indeed - maybe they could be leveraged for some of the deletion logic by way of `CASCADE`, but the logic to clean up is already there, and it's nice and explicit in the code what's getting cleaned up when.

> Good point, fixed in https://github.com/Jermolene/TiddlyWiki5/commit/c26acfdb429960ef00c1d96be41f4d6c0ceec4e6

:+1: 

You might want to add `NOT NULL` to the primary keys, too (unless `AUTOINCREMENT` implies that already - I'm not sure!), as well as to `fields.tiddler_id`.

> Yes, that would make sense, and if we're going to do it I'd be inclined to try to get it in now. Would you like to try your hand at a PR? It would be close to the edge of my SQL query skills.

I could! I think @pmario has got a point with avoiding "premature optimization", though, and storing the field names directly is already working, so maybe what I'll do is trying to spin up a copy of my personal wiki using the SQLite database, do a little manual table manipulation, and then see what the storage impact would be.

hoelzro:
@Jermolene BTW I noticed the "Instant syncing from server to browser" feature as well - do you have any thoughts on how you'd like to approach that?  That feature certainly grabbed my attention!

Jermolene:
Thank you @hoelzro @CrossEye. Conscious that the discussion thread at #6666 quickly got unmanageable, I propose to spawn separate GitHub discussions for each of the major areas of discussion. I'll link to them from the OP here.

I wondered if anyone wanted to volunteer to be the lead editor on each of these threads? I'm envisaging the lead editor being the person who starts the thread, and keeps the top post up to date to reflect our current thinking as the discussion evolves.

The threads that we might need include:

* SQL schema - planned enhancements (attachments, versions, delete markers), and (in due course) performance improvements
* Authentication and authorisation
* Syncing
* File attachment handling (perhaps)

The lead editors might well want to thrash things out a bit more here before actually starting their threads.

I'd welcome volunteers – I should stress that it doesn't have to be people who are already involved in this discussion, either.

Jermolene:
> As I understand it, best practices seem to be trending away from passwords altogether. This option -- which I'm guessing a large number of users would want -- is probably the most difficult to set up.

Yes indeed. I remember in the old days that it wasn't that hard. One hashed the password into a text file (with a salt), and then matched the hashes to authenticate. I liked the way that the hashed passwords didn't need to be kept secret, and I'd rather like that in MWS: every users password should be visible as a tiddler containing a hash of the password. I won't put up much resistance if others want to argue me out of that position, but I do think it would be elegant.

> Running on Node, it may well be able to do so, to send passwords, or or texts with a short-term sign-in key. I haven't done this, but it sounds feasible.

My understanding now is that the big platforms will effectively only accept mail that originated with an approved email server (like gmail, icloud, hotmail etc). In other projects I have used Amazon's APIs to send emails and text messages, and it is very much commoditised as a commercial service, but I think nowadays very rarely available without charge.

> I'm guessing this would be the easiest one to implement, but it would require owners of the wiki back-ends to do some setup to get individual keys for each of the providers they decide to use.

I would hope that we'd build a gui for that kind of task that users would access through the admin wiki.

> Presumably that's also straightforward. Where is the current supporting code?

It is stunningly simple. The user can specify the name of a header that, if present, is taken by the server to be the name of the current user. In our case it ends up being copied to  the standard `$:/status/UserName`. See https://tiddlywiki.com/#WebServer%20Header%20Authentication for more details. The code is [here](https://github.com/Jermolene/TiddlyWiki5/blob/master/core/modules/server/authenticators/header.js)
> I have a few reactions:

Ha! I'm perhaps optimistically and naively hoping that we can find some off-the-shelf code that we can either use directly, or use as inspiration.

> I'd love to hear your reasoning, because I personally would not trust someone's $5/month Node environment to securely hold my credentials so they can share them with other registered wikis.

I already know of several companies that run a Node.js TiddlyWiki instance within their intranet, at least one of which uses the trusted header technique for authentication. But several do manage their own user database in a CSV file, as TiddlyWiki currently requires. This would often be people without traditional IT experience, but who are sufficiently motivated to use TiddlyWiki that they have mastered the steps needed to get TiddlyWiki on Node.js running.

So, if I think of somebody in that situation, who has gone to all that trouble to set up their user database (presumably in some new, secure manner as discussed above), then it's easy to imagine them wanting to leverage that investment by hanging other things off the side.

Obviously I'd bow to others as to the feasibility of the idea, but I think it would be neat and I'd like to explore it.

As a sidebar, something even madder that I have long wanted is to give TiddlyWiki under Node.js an IMAP interface so that rendered tiddlers are exposed as messages in mailboxes that correspond (say) to tags. The magic would be that every device on the planet has an IMAP client that works offline...

> Or are you suggesting that the main site, tiddlywiki.com serve that role? That could be interesting. Or was that not a simple typo, and are you talking about some ancillary TiddlyWiki tool, "TiddlyWiki**t**", that I've never heard of?

I hadn't thought of that, but it might be interesting when we get to the point of setting up a shared community instance of MWS.


> Usually when you hear about middleware related to Node, it's talking about the most common web framework for Node, [Express](https://expressjs.com/). I haven't really looked at the new code yet, but if you're not already using Express, I doubt it's worth bringing in just for this. If it's already in place, then yes, middleware is a great help for auth code.

I was hoping that we might be able to be sufficiently plug compatible with Express that we could reuse selected bits of middleware. Having said that, my general feeling is that we should continue to avoid the user of 3rd party libraries unless absolutely unavoidable. A subtle corollary is that when we do adopt such a library we need to reuse the heck out of it to make sure we get value for money.

> Of course, and that probably means some sort of interface that the different auth mechanisms all present, with some configuration that says, "Yes, I want a password mechanism; yes I want OAuth, with Google, GitHub, and LinkedIn providers; no, I don't need trusted headers," or whatever.

Exactly that.

> That leads me to an interesting question: Would you want the auth configuration to be at an installation level or at a wiki level? Or is there some level in between to consider?

Interesting. I had assumed the installation level without thinking it through too much, but it would be interesting to see if it was viable to do it at the wiki level.

> That would clearly be necessary for an in-house password system. But it might or might not be the case for external providers, AD groups, or trusted headers.

I suspect we'd want to keep some private information against each user. For example, each user would have a bag name, which might be exactly the same as their username if the authentication provider has different rules than our bag names. (I should add that I've conservatively constrained bag and recipe names to a small set of explicitly allowed characters. I realise we may need to expand things to cope with the inevitable demand for bags and recipes named with emojis).

> We can look for prior art, but I think we'll have to take great care in determining the proper granularity of ACLs. One real concern would be whether we would want to grant a user read-only access to a bag in one recipe, but read-write access to the same bag in a different recipe. It would be easiest just to forbid this, but I think people might well come up with reasonable scenarios we couldn't support. And that would be a shame.

The [original TiddlyWeb specification](https://tank.peermore.com/tanks/tiddlyweb/Auth%20Model) has ACLS independently on both recipes and bags, but I don't think it would allow for the scenario you have outlined.

I have lots of battle scars from projects involving authorisation over the years. I've found that, in this area, if you ask clients what they want they will freely invent incredibly complicated schemes. In a TiddlyWiki context, people frequently suggest ACLs at the tiddler level. However, in my experience, faced with an access control environment that is too complex, almost everyone will retreat to taking the easiest path, compromising the very security concerns that prompted the invention of the complex model in the first place.

We explored simplifying the TiddlyWeb model with TiddlySpace. It was just an instance of TiddlyWeb with a bunch of plugins that added a simpler model: users could create "spaces" which could contain both private and public tiddlers. The public tiddlers would be visible to other users, and the private tiddlers would just be visible to the user and anyone else they authorise. Private tiddlers were highlighted in pink and public tiddlers in blue.

The idea was to model a common situation: writing/painting/collecting/acquiring a bunch of stuff that starts out private but can be published at any time with a single click.

TiddlySpace was popular, way more popular than TiddlyWeb alone. However, I felt that the "spaces" model was not a success. The trouble is predictability: when I am editing a tiddler I need to concentrate on almost imperceptible details in order to understand whether it is private or public. That compromises the users ability to predict what will happen if they take an action in the UI.

I have ended up with the view I've expressed elsewhere: that the browser location should be sufficient to determine where edits are going, without having to consider different routes for individual tiddlers.
 
Anyhow, that is all very abstract, but underlines that this is a difficult area. The policies that I think will help to guide us are our usual ones: to have as much functionality as possible implemented in wikitext, and to deeply understand the needs of our users.

> I'll start thinking about this and reading the current code soon. No promises that I can contribute usefully, but I would like to try.

Fantastic, thank you, much appreciated. Just asking questions is helpful to draw out some of the issues.

Jermolene:
> Indeed - maybe they could be leveraged for some of the deletion logic by way of `CASCADE`, but the logic to clean up is already there, and it's nice and explicit in the code what's getting cleaned up when.

My instinct is also to make the logic as explicit as possible, and avoid features that overlay invisible magic. But I'm still ignorant enough about the tradeoffs to be confident that we shouldn't be using those features in order to give SQLite the maximum opportunity to apply its own optimisations.

> You might want to add `NOT NULL` to the primary keys, too (unless `AUTOINCREMENT` implies that already - I'm not sure!), as well as to `fields.tiddler_id`.

Ah, I wasn't sure about that either, but rather that row_id columns can't be NULL.

> I could! I think @pmario has got a point with avoiding "premature optimization", though, and storing the field names directly is already working, so maybe what I'll do is trying to spin up a copy of my personal wiki using the SQLite database, do a little manual table manipulation, and then see what the storage impact would be.

That would be great, thank you.

Jermolene:
> @Jermolene BTW I noticed the "Instant syncing from server to browser" feature as well - do you have any thoughts on how you'd like to approach that? That feature certainly grabbed my attention!

I've long been a fan of Server Sent Events because of the way that they can easily pass through most firewalls.

pmario:
>Jeremy wrote
>
>...users could create "spaces" which could contain both private and public tiddlers. The public tiddlers would be visible to other users, and the private tiddlers would just be visible to the user and anyone else they authorise. Private tiddlers were highlighted in pink and public tiddlers in blue.

I think there was 1 more option. If private-tiddlers where saved, they where written to the private bag where they came from (If the user had write access). The same was true for public tiddlers. But there was a configuration for "protected" bags. They where public in scope, but they did behave like shadow tiddlers. If overwritten, they went to the highest level public bag. 

As I see the current configuration that you described, where new tiddlers are always written to the "highest" level bag. I think every bag below that level is automatically treated as a "shadow" store.

Jermolene:
I have been exploring switching to using SQLite3 compiled to Wasm instead of the current native code implementation. See #7996

plugins/tiddlywiki/multiwikiserver/modules/command-mws-load-tiddlers.js
```diff
@@ -0,0 +1,40 @@
...
+
+Command.prototype.execute = function() {
+	var self = this;
+	// Check parameters
+	if(this.params.length < 2) {
+		return "Missing pathname and/or bagname";
+	}
+	var tiddlersPath = this.params[0],
+		bagName = this.params[1];
+	$tw.mws.store.saveTiddlersFromPath(tiddlersPath,bagName);
```
linonetwo, line 34: I think I can directly call this method, after a git sync from remote is done.

And call "save tiddlers" when user want to sync To remote. What is differences between load-archive and load-tiddlers? And seems there is no save-tiddlers?

Jermolene: > I think I can directly call this method, after a git sync from remote is done. And call "save tiddlers" when the user wants to sync To remote, or every 30min.

It is my intention provide an action widget equivalent of this command, so that we will be able to extend the new "filtered actions" mechanism in #7999 to trigger saving the archive at intervals if there have been changes.

> What are the differences between load-archive and load-tiddlers? And it seems there are no save-tiddlers?

`mws-load` and `mws-save` introduce a standard archive format for MWS, with a JSON file for each tiddler, recipe and bag. The `load-tiddlers` command loads a bag with the tiddlers from a tiddler folder, which is an ordinary folder containing `.tid`, `.multids`, `tiddlywiki.files` etc, the same as the "tiddlers" subfolder of a wiki folder.

I've no objection to a `save-tiddlers` command that saves tiddlers as `.tid` files. However, until we extend the format to allow multiline fields, some tiddlers would presumably end up as `.json` files.

I specifically want to avoid reusing our existing mechanism for writing changes back to a wiki folder. There are many edge cases that make it unreliable and brittle (for example, editing a tiddler from a `.multids` file).

linonetwo: This is OK, looks like `load-tiddlers` and (hopefully there can be a) `save-tiddlers` are compatible to the old nodejs folder, which is being used in the client side. And `mws-load` `mws-save` can be used in the Real server side.

`.multids` won't be a problem because most of client user won't touch the folder under the hood, they only use wiki UI. Same as multiline fields, it only generated by very few plugin, as long as not all tiddler become JSON, I think user will accept some tiddler become json.

pmario:
@Jermolene ... I think there is a type and the wrong start command in the OP

image

Jermolene:
> @Jermolene ... I think there is a typo and the wrong start command in the OP

Thanks @pmario, much appreciated, I've edited the OP.

Jermolene:
I've just completed some fairly major refactoring in preparation for beginning work on the improved synchronisation process:

* No longer using the main root wiki for admin functions
* Introduced new static index route listing bags and recipes and allowing them to be created/modified
* Stylesheets for static routes

The next step is to replace the "tiddlywebadaptor" with a new "mwsadaptor". That will allow the routes to be rationalised (avoiding the double recipe problem).

plugins/tiddlywiki/multiwikiclient/config-incoming-updates-filter.tid
```diff
@@ -0,0 +1,2 @@
+title: $:/config/multiwikiclient/incoming-updates-filter
+text: [all[]] -[[$:/isEncrypted]] -[prefix[$:/temp/]] -[prefix[$:/status/]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] -[[$:/library/sjcl.js]] -[[$:/core]] -[[$:/StoryList]] -[[$:/HistoryList]]
```
pmario, line 2: IMO to should be `-[prefix[$:/StoryList]]` and `-[prefix[$:/HistoryList]]`

Jermolene: Thanks @pmario, fixed in d51ad80f8001f774ec98371a542aa10eba3c3dff

Jermolene:
Thanks @PotOfCoffee2Go that's very strange. The tiddler `$:/config/multiwikiclient/recipe` is baked into the wiki as part of get-wiki.js:

https://github.com/Jermolene/TiddlyWiki5/blob/d51ad80f8001f774ec98371a542aa10eba3c3dff/plugins/tiddlywiki/multiwikiserver/modules/routes/handlers/get-wiki.js#L76-L79

If it isn't present in the wiki your browser is seeing, then one possibility that the repo might be corrupt, and it would be worth redownloading/cloning the whole thing. If you still get the error, could you write up the actions needed to reproduce.

PotOfCoffee2Go:
Thanks! @Jermolene 

> that's very strange. The tiddler $:/config/multiwikiclient/recipe is baked into the wiki as part of get-wiki.js

That's it ('get-wiki.js' sounded familiar) -  I boot MWS up in my app and had a bunch of pre-installed tiddlers that I was using to  investigate stuff (a week or more ago). So overriding with stale tiddlers - forgot all about them.

Will set up a 'pure' multi-wiki-support directory and compare my run with that in the future.

Jermolene:
Thanks @PotOfCoffee2Go.

plugins/tiddlywiki/multiwikiclient/info-segment.tid
```diff
@@ -0,0 +1,8 @@
+title: $:/plugins/tiddlywiki/multiwikiclient/info-segment
+tags: $:/tags/TiddlerInfoSegment
+
+<$reveal type="nomatch" state=<<folded-state>> text="hide" tag="div" retain="yes" animate="yes">
+<div class="tc-subtitle">
```
pmario, line 5: I think since reveal already creates an animated and retained DOM element, it's not needed to include DIV just to add a class. It should be OK to add the class to the reveal widget.

core/modules/widgets/image.js
```diff
@@ -163,4 +163,4 @@ ImageWidget.prototype.refresh = function(changedTiddlers) {
 
 exports.image = ImageWidget;
 
-})();
+})();
```
pmario, line 166: I thought, we can remove the "covering" function?

plugins/tiddlywiki/multiwikiclient/configOfficialPluginLibrary.tid
```diff
@@ -0,0 +1,7 @@
+title: $:/config/OfficialPluginLibrary
+tags: $:/tags/PluginLibrary
+url: https://tiddlywiki.com/library/v5.1.23/index.html
```
pmario, line 3: v5.1.23 ?? Is this version number intended?

Jermolene: Ouch – that's actually an issue with the core;  the same file is at https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/tiddlyweb/configOfficialPluginLibrary.tid

plugins/tiddlywiki/multiwikiclient/multiwikiclientadaptor.js
```diff
@@ -0,0 +1,341 @@
...
+A sync adaptor module for synchronising with MultiWikiServer-compatible servers
+
+\*/
+(function(){
+
+/*jslint node: true, browser: true */
+/*global $tw: false */
+"use strict";
+
+var CONFIG_HOST_TIDDLER = "$:/config/multiwikiclient/host",
```
pmario, line 15: In other code below you used `const` -- IMO it could be used here too.

plugins/tiddlywiki/multiwikiclient/save-wiki-button.tid
```diff
@@ -0,0 +1,26 @@
+title: $:/core/ui/Buttons/save-wiki
+tags: $:/tags/PageControls
+caption: {{$:/plugins/tiddlywiki/multiwikiclient/icon/cloud}} Server status
+description: Status of synchronisation with server
+
+\whitespace trim
+\define config-title()
+$:/config/PageControlButtons/Visibility/$(listItem)$
+\end
+<$button popup=<<qualify "$:/state/popup/save-wiki">> tooltip="Status of synchronisation with server" aria-label="Server status" class=<<tv-config-toolbar-class>> selectedClass="tc-selected">
```
pmario, line 10: IMO wikitext code should be indented to make it more readable

plugins/tiddlywiki/multiwikiclient/multiwikiclientadaptor.js
pmario: Just saw this video about const -- [Let Me Be: "const ... Is Stupid and You're Using It Wrong"](https://www.epicweb.dev/talks/let-me-be) 

I think it's worth viewing.

simonbaird:
Is MWS likely to have support for multi-user? Could it become the back end for a future version of Tiddlyhost perhaps?

Jermolene:
> Is MWS likely to have support for multi-user?

Yes, that is the plan. Servers have got powerful enough recently that a single VM should be able to serve many thousands of users. And it ought to be possible for deployments to be scaled horizontally by running multiple application servers synchronising changes between their integrated SQLite databases.

We had basic multiuser support 15 years ago in the TiddlyWeb/TiddlySpace days. MWS is designed to go further, and deal with the practical challenges we found with extending TiddlyWiki's single user model to the cloud.

For example, in single user TiddlyWiki it is very natural that drafts should be ordinary tiddlers that live alongside non-draft tiddlers. It means that drafts are stored within the context in which they will be used, and so are not apt to get lost.

There is an immediate problem that my "Draft of 'HelloThere'" might clash with your "Draft of 'HelloThere'". We got around that on TiddlySpace by including the username in the draft title: "Draft of 'HelloThere' by JoeBloggs".

A more insidious, subtler problem lies in the fact that draft tiddlers are visible to any user with read access to that bag. This categorically does not match the expections of users set by their encounters with other cloud services: in Instagram or TikTok draft posts are absolutely private until they are explicitly published.

So, the plan with MWS is to store draft tiddlers in a special, private bag that is provisioned for each user. By default, drafts would go into this bag, and perhaps things like all the `$:/state/` tiddlers that would be accumulated by clicking around somebody else's site.

The plan is that these special, private user bags will actually be the entirety of the user database. Provisioning a new user will be a matter of creating a bag with the attributes that mark a user bag. Because the bag is private it can also be used for security sensitive data like OAuth tokens.

>  Could it become the back end for a future version of Tiddlyhost perhaps?

If that's something you think might be interesting for you then there's no reason we couldn't start running an experiment now. You could use continuous deployment to make https://mws.tiddlyhost.com/ from prerelease tags I would make on the branch. We'd collectively learn a lot about database backups and setting up backups (and restores) and other issues that might not otherwise be on my radar for a while.

Of course, right now, there's no security on MWS, so it would necessarily be an unprotected public space. That could be fun, but would could also conceivably constitute a small liability as you would need to monitor it for bad content, and perhaps set up reporting.

plugins/tiddlywiki/multiwikiclient/multiwikiclientadaptor.js
Jermolene: > I think it's worth viewing.

I'm afraid I do not generally watch videos, could you summarise the key points?

MagicJF:
Hello, lets suppose a teacher and students using MWS.

Teacher puts a math equation in "tiddler A" within a shared bag to its students and gives as homework to solve it.

Student nº1 edits "tiddler A" to answer and its version is stored in student 1 private bag while student nº2 doesn't do any homework and no version is stored in its private bag.

Then, teacher adds more info and context to "tiddler A" math equation.

Student nº2 will be able to get the new info, but what about student nº1 who already has a modified copy?

Mirodin:
I want to thank @Jermolene and all contributors so very much for the many years I am able to use TiddlyWiki (is it really 13 already O.o)!

Since you invited comments I would like to bring up a topic that was brought up very shortly [over at the Forum](https://talk.tiddlywiki.org/t/announcing-the-multiwikiserver-plugin/9033/2): Subdomains

Right now I just run an instance of the Node version on my server for every topic (inventory, crm, recipes, documentation, ...). I really dig the new (old) design based on bags and recipes because I often had the same issue one knows from archiving their documents when a topic/tiddler fits in more than one wiki.

However for routing and management reasons (and pretty URLs), I would love to keep the default Node assumption that a wiki runs on webroot. The problem with serving wikis at their folder path is that a reverse proxy that handles also authentication can not rewrite URLs on the fly like it could with "normal" webpages because the server returns JSON on its API and not rendered HTML. So rewriting URLs that do not yet exist and get created at the frontend would break if the wiki is served via a subdomain.

I envision an (opinionated) setup like this:

- Main wiki/dasboard at https://admin.example.com (Behind authentication from reverse proxy so only admin can create/delete new instances)
- Bob's private wikis at https://bob.example.com (Only Bob has access)
- Alice's private wiki at https://alice.example.com (Only Alice has access)
- Family wiki at https://family.example.com (All family members have access)
- Public wiki at https://wiki.example.com (Everybody can read, but the top bag is a dummy/empty so edits do not propagate to other/included bags)

1) Leveraging a reverse proxy for authentication would also reduce the burden on TW implementing secure authentication/authorization and not force duplicate setup of credentials. Many applications use a header the reverse proxy can send to the upstream app to "disable" login for the app (i.e. Paperless-NGX).

2) Maybe I am wrong but the wiki path/folder is an implementation detail that when leaked would break inter wiki links setup in Node if one switches to MW. For example some tiddler or external site links to https://tiddlywiki.com/#HelloThere and we decide to migrate to MW on the backend that link would break because MW would introduce the folder path.

I hope my thoughts are somewhat understandable and I am happy to discuss this further if there is interest.

pmario:
> Hello, lets suppose a teacher and students using MWS.
> 
> Teacher puts a math equation in "tiddler A" within a shared bag to its students and gives as homework to solve it.
> 
> Student nº1 edits "tiddler A" to answer and its version is stored in student 1 private bag while student nº2 doesn't do any homework and no version is stored in its private bag.
> 
> Then, teacher adds more info and context to "tiddler A" math equation.
> 
> Student nº2 will be able to get the new info, but what about student nº1 who already has a modified copy?

I think that's a very good question. IMO that's a case, where the UI needs to make sure, that the "original" tiddler can always be read, so "teachers" notes will always be visible. 

In the current wiki, we do have an operator that is named [shadowsource](https://talk.tiddlywiki.org/t/howto-compare-the-content-of-a-shadow-tiddler-with-the-content-of-an-overwritten-shadow/6191), which lets us find out, where an eventually overwritten shadow tiddler comes from. 

We will need something similar for the "teachers" tiddler. In a multi-user multi-"space" system such an operator may give us a "trail" of overwritten shadows. 

Just a thought.

pmario:
>2. Maybe I am wrong but the wiki path/folder is an implementation detail that when leaked would break inter wiki links setup in Node if one switches to MW. For example some tiddler or external site links to https://tiddlywiki.com/#HelloThere and we decide to migrate to MW on the backend that link would break because MW would introduce the folder path.

I think there you are wrong. 

- The TW recipe in your words would create a subdomain. 
- A recipe can contain several bags
- Every bag can contain tiddlers with the same title
- The bag that comes last in the recipe will always win. 

That's the unique point of TW shadow tiddlers. In a single file context the "core" contains shadow tiddlers, that can get overwritten. If the user overwrites them, they will always win. -- But if the user messes up a shadow tiddler, deleting this tiddler will fix the problem and "reactivate" the core shadow tiddler. 

The bag mechanism is similar. Every bag, but the last one will contain shadow tiddlers. The last bag is the users "bag", where they have write access out of the box. If the mess something up, they can delete that tiddler and the tiddler with the same name from a "lower level" bag will take over again. 

Hope that makes sense.

MagicJF:
Just to extend the "teacher"'s tiddler problem.

Lets define a permisions hierarcy where:
Directors > Teachers > Students

If student solves equation in tiddlerA, solution is stored in its private bag and changes dont propagate to other students.

If teacherX edits tiddlerA adding more info, changes do propagate to all  teacherX's students, but not students of teacherY.

If director edits tiddlerA, changes propagate to teacherX, teacherY, teacherX's students, and teacherY's students.

Mirodin:
Hi Mario,

thank you for taking the time to answer.
However I think we are talking about different things:

I understand how the whole bag/recipe mechanic works. But my comment was not about that but rather about running the system and how to manage multiple wikis (read sysadmin).

If I run MW via the command given above I can access the main wiki via `http://example.com/`.
However, the docs wiki is accessed via `http://example.com/wiki/docs` and this is the problem I am talking about. One can not run the main wiki at i.e. `http://main.example.com` and the docs wiki at `http://docs.example.com`. The fact that the folder path is leaked prevents subdomains from being used because a reverse proxy can not rewrite URLs in the JSON payload (as it would in "normal" HTML pages) send from the server. Since there are no URLs in the payload yet i.e. `http://example.com/wiki/docs` to `http://docs.example.com`. The wikitext gets only parsed at the frontend - after passing the proxy - and this is where URLs are generated.

This is why I asked if we could keep the Node assumption that a wiki is accessed at the webroot.

Hope this clears things up.

Jermolene:
I have experienced sync problems when deploying MWS to cloud service providers. It turns out that the current, simplistic implementation of server sent events is the problem: specifically, when saving a tiddler it is possible for the SSE notification to be received by the client before the response to the PUT.

I believe that the best fix for this is quite fundamental, but is happily what I was planning to do anyway: to make the tiddler table record old versions of each tiddler, in effect turning it into an append only, write only store, which is known to be much easier to synchronise.

This morning [coderofsalvation over at talk.tiddlywiki.org](https://talk.tiddlywiki.org/t/multiple-use-tws-and-persistence/9861/2) posted a link to a post that outlines exactly the issues: https://www.event-sourcing.dev/explain-like-im-five-intro/

joshuafontany:
Very interesting article. Thanks. Still following your progress.

andjar:
> I think that's a very good question. IMO that's a case, where the UI needs to make sure, that the "original" tiddler can always be read, so "teachers" notes will always be visible.

I agree with you; a similar scenario would perhaps be if one starts using tiddlywiki to integrate data from various sources in different bags. Say I have one script that exports mails to the mail bag, or from logseq (note taking software somewhat similar to obisidan) to a logseq bag, or from my calendar to a calendar bag. Mails are unlikely to change, so they are fine. Logseq files may be edited externally, but if they are intended to be read-only in tiddlywiki, that's not a problem. Calendar events, however, may change and it's not unlikely that one starts writing about them in tiddlywiki in parallell - and then one would need to have access to the updated source tiddler.

PotOfCoffee2Go:
Not sure if this the place to report this - and it might be something I'm doing on my side. Or is working as expected.

Get a lot of '$__StoryList' tiddlers in the MWS tiddler directory. Trying  track it down.

I believe it happens when  database.sqlite is deleted (in order to create a new DB on MWS startup). 

Not a big deal - thought would mention it.

```
./mws
├── store
│   ├── database.sqlite
│   ├── database.sqlite-shm
│   └── database.sqlite-wal
├── tiddlers
│   ├── $__layout.tid
│   ├── $__StoryList_10.tid
│   ├── $__StoryList_11.tid
     .   .   .
│   ├── $__StoryList_4.tid
│   ├── $__StoryList_5.tid
│   ├── $__StoryList_6.tid
│   ├── $__StoryList_7.tid
│   ├── $__StoryList_8.tid
│   ├── $__StoryList_9.tid
│   ├── $__StoryList.tid
│   ├── A test tiddler.tid
│   └── configMultiWikiServerEngine.tid
└── tiddlywiki.info

```

pmario:
@PotOfCoffee2Go Which OS do you use? For me on Windows this happens, if the StoryRiver tiddler is blocked by a different thread. Eg: if I did open it in a text editor and then restart the server. Since the original tiddler is blocked by the editor the server creates a new one. The problem then is, that even if the editor does not show the tiddler anymore, the OS still blocks it. 

For me the problem goes away when I do close the server, then delete all story-tiddlers and restart the server. -- Just a guess

PotOfCoffee2Go:
@pmario

I'm using Linux Debian in this case. But as you mention it happens on Windows too. Running v5.3.5-prerelease (branch multi-wiki-support). I am running multiple instances of tiddlywiki that hit the same database using the $tw.mws.store functions.

Agree that has something to do with OS file locking.   "tiddlywiki/tiddlyweb" sends storylist a lot and "tiddlywiki/filesystem" is probably getting conflicts with the multiple tiddlywiki's running against the same mws/store/tiddler directory.

As an experiment, will try is to set up a tiddlywiki.info file for the second tiddlywiki instance without "tiddlywiki/tiddlyweb" or  "tiddlywiki/filesystem" since it just needs access to the database. Will have the OS do a symLink to the first instance's database.sqlite. (if that makes any sense?)

And the conflict should go away. That being said - the same problem can happen with 'server' edition - so is not just MWS.

PotOfCoffee2Go:
Hi @Jermolene 

Was wondering if you would be adverse to adding a error handler at the end of `modules/mws-server.js' which catches EADDRINUSE on the call to listen?

My app has a button to start MWS server as do not want to launch it all the time - or already running the server elsewhere standalone.

Currently causes a Node.js 'Uncaught Exception' which crashes the app - as Node.js does now-a-days on Uncaught Exceptions )   ;(

If the error is not EADDRINUSE - then the throw(err) in the 'else' will still throw as a Node.js Uncaught Exception because the handler is running in the scope of Node.js. Thus is backward compatible with the existing setup - the app will error and crash.

```
	server.on("error", (err) => {
	  if (err.code === "EADDRINUSE") {
		console.error(`MWS Server error code: EADDRINUSE - port ${port} is already in use`);
	  } else {
		throw(err);
	  }
	});

	// Listen
	return server.listen(port,host);
```

Please advise

Jermolene:
Hi @PotOfCoffee2Go 

> Was wondering if you would be adverse to adding a error handler at the end of `modules/mws-server.js' which catches EADDRINUSE on the call to listen?

My first reaction is that that behaviour would be an uncommon requirement. I would expect most users to want to see an error if the server could not start up properly.

It's hard to see a way to allow that behaviour to be optionally configured beyond the obvious approach of a `fail-on-addinuse=true` flag that could be used with the listen command.

Could your requirement be addressed by interrogating the server object to determine if it is already listening? Or are you concerned about instances in other processes?

PotOfCoffee2Go:
Hi @Jermolene 

> Or are you concerned about instances in other processes?

Yes - Querying data from other databases and internet resources, composing as tiddlers and storing on MWS in recipes/bags. Custom RESTful routes compose recipes into `{ tiddlers: [] }` format and delivers to TiddlyWiki clients using the upcoming (SSE / MWS ) when released and settles in.

Was initially accessing the same SQLite database using $tw.mws.store with multiple TiddlyWiki instances sharing the same directories. But have found it to be difficult to manage, somewhat brittle (OS file locking issues, see above), and am concerned about future SQLite record locking. So now favor just a single TW instance accessing the database and having multiple 'server' instances (using async wrappers) that are managed by the ServerManager.

> Was wondering if you would be adverse to adding a error handler at the end of `modules/mws-server.js' which catches EADDRINUSE on the call to listen?
>> My first reaction is that that behaviour would be an uncommon requirement. I would expect most users to want to see an error if the server could not start up properly.

Agree.

So after further consideration - the change I proposed has not solved the problem of catching the listen() error - which the app needs to do to in order to resolve the problem and attempt to listen again; without intervention.

My solution is to catch the error in the ServerManager which is simple to implement:

```
ServerManager.prototype.createServer = function(options) {
    const MWSServer = require("$:/plugins/tiddlywiki/multiwikiserver/mws-server.js").Server;
    try {
        const server = new MWSServer(options);
    } catch(err) {
        if (err.code === "EADDRINUSE") {
            ... resolve the listen error - and try again ...
        }
    }
    this.servers.push(server);
    return server;
}
```

The ServerManager is in 'startup.js' which is unique to the application - for example the current multi-wiki-support startup.js loads TW core editions and test recipe/bags. Mine of course loads editions for my app instead.

PotOfCoffee2Go:
A MWS use case.

While dabbing my toe in the test waters of MWS - wanted to come up with a system to export a recipe from a MWS database which can be sent as an email attachment to friends, family, and foes; they bring up empty.html and drag-n-drop the email attachment - and 'poof!' an exact copy of the recipe is loaded and running on empty (pun not intended).

The issue of plugins, themes, and languages is solved by exporting the tiddlers of the plugin - not as a plugin - but as a set of tiddlers which are the content of the plugin. Simple and effective. The JSON file produced is a flat array of all tiddlers in the wiki excluding TW core :

```json
[
  {"title": "tiddler1", "...": "..." },
  {"title": "tiddler2", "...": "..." },
  {"title": "tiddler3", "...": "..." },
      ...
]
```

Which '$:/import' happily imports. So let's say the 'server' edition has Shiraz as a plugin - the exported JSON will contain all the tiddlers of Shiraz but as individual tiddlers. When imported the  empty.html 'control panel->plugins' tab will only show 'TW core'. However, since the Shiraz plugin tiddlers are in the wiki - it works.

loader3

> Note: that if the JSON file has JavaSrcipt tiddlers, the import will popup requiring a local save/reload - as expected.

All works well, except MWS attachments - which I am still working on how to get them into the exported JSON file - mostly interested in images; not large media files.

pmario:
>The issue of plugins, themes, and languages is solved by exporting the tiddlers of the plugin - not as a plugin - but as a set of tiddlers which are the content of the plugin. Simple and effective. The JSON file produced is a flat array of all tiddlers in the wiki excluding TW core :

You should not extract plugin tiddlers and ship them as single tiddlers. There is a reason why plugin are packed.

On the other hand plugins are also tiddlers. So there is no problem to [ship plugins with other content tiddlers as a bundle](https://wikilabs.github.io/editions/bundler/#tiddlerlist-plus-plugin.bundle:tiddlerlist-plus-plugin.bundle%20GettingStarted%20doc-bundle-lists%20doc-filtered-bundles%20new-in-v1.0.0%20new-in-v1.1.0). 

So they can easily be imported as plugins again.

PotOfCoffee2Go:
> You should not extract plugin tiddlers and ship them as single tiddlers. There is a reason why plugin are packed.

Agree 1000%. This export I use to unpack my plugins when I am working on them. Drop the JSON into empty, make changes, and export back to JSON. A companion process reads the JSON and handles the details of repackaging the plugin. For example, if I add a tiddler it automatically includes it in the repackage, handles versioning, etc.

Did not know about Bundles. Thanks for the link - have a new shiny object to play with.

While testing MWS most is throw away code - since I was updating the export repackage to include the ''stability' field - "hey, I wonder if this would work on MWS?" - so just copy/pasted the export code and figured it would probably not work with the un-packaging logic. Changed to use $tw.mws.store functions to pull the recipe tiddlers - and was frankly quite surprised it worked! So using the database is effectively the same as getting tiddlers from a HTML tiddler store.

For email attachments - is when sending stuff to my (too many 'great') grandkids, parental controls block attachments - .zips, .exe, HTML that contain scripts, just about everything. Normally would just send the whole TiddlyWiki but it gets blocked. JSON is allowed, so I send it exported instead. They know how to drag-n-drop it (most were born holding a cell phone). Not great but works.

So my first 'production' use of MWS will be to hold my bimonthly picture books I write for the grandkids.

PotOfCoffee2Go:
Hi @Jermolene 

The editions are not loading. In 'commands/mws-load-wiki-folder.js'  the resolved path has doubled up the 'editions' sub-directory : 

```js
$tw.mws.store.createRecipe(options.recipeName,recipeList,options.recipeDescription);
console.dir(path.resolve($tw.boot.corePath,$tw.config.editionsPath,options.wikiPath,$tw.config.wikiTiddlersSubDir))
$tw.mws.store.saveTiddlersFromPath(path.resolve($tw.boot.corePath,$tw.config.editionsPath,options.wikiPath,$tw.config.wikiTiddlersSubDir),options.bagName);

```
The console.dir shows the duplicate

```
:~/Documents/sites/TiddlyWiki5 $ npm start

> tiddlywiki@5.3.5-prerelease start
> node ./tiddlywiki.js ./editions/multiwikiserver --mws-load-plugin-bags --build load-mws-demo-data --mws-listen

'/home/poc2gopi/Documents/sites/TiddlyWiki5/editions/editions/tw5.com/tiddlers'
'/home/poc2gopi/Documents/sites/TiddlyWiki5/editions/editions/dev/tiddlers'
'/home/poc2gopi/Documents/sites/TiddlyWiki5/editions/editions/tour/tiddlers'
syncer-server-filesystem: Dispatching 'save' task: $:/StoryList
Serving on http://127.0.0.1:8080
(press ctrl-C to exit)
```

I'm not familiar with the $tw paths - figured you would be able to identify the error fairly quickly.

thanks.

kookma:
I would suggest in the diagram add soft to indirect and variable like `soft indirect`

PotOfCoffee2Go:
Hi @Jermolene 

In testing attachments have discovered that they work really well with media software which accepts URIs as import. However most image, audio, and video software use the last name of the attachment URI as the default name of the content. Currently, that is 'blob', ie: `http://localhost:8080/bags/mwstest/tiddlers/components.png/blob`. 

Might want to change the route so the name of the content is the last element of the URI.

I can make that change in a PR if you want - just need know what you think the route URI should be.

Jermolene:
Hi @PotOfCoffee2Go apologies for the delayed response.

> My solution is to catch the error in the ServerManager which is simple to implement:

I'd prefer to have the try catch scope just be the method that is liable to fail.

> The issue of plugins, themes, and languages is solved by exporting the tiddlers of the plugin - not as a plugin - but as a set of tiddlers which are the content of the plugin. Simple and effective

I'm sure you'll have discovered that this approach might be brittle for some plugins. By exploding the plugins the payload tiddlers will no longer be shadows. This could cause problems if the plugin uses techniques like accessing its overridden shadow tiddler tiddlers directly via the "subtiddler" attribute. I would recommend shipping the plugins as normal.

> The editions are not loading. In 'commands/mws-load-wiki-folder.js' the resolved path has doubled up the 'editions' sub-directory :

Are things still broken? I can't reproduce the problem.

> In testing attachments have discovered that they work really well with media software which accepts URIs as import. However most image, audio, and video software use the last name of the attachment URI as the default name of the content. Currently, that is 'blob', ie: `http://localhost:8080/bags/mwstest/tiddlers/components.png/blob`.
> 
> Might want to change the route so the name of the content is the last element of the URI.
> 
> I can make that change in a PR if you want - just need know what you think the route URI should be.

Good point. It does need some thinking about. It's less than ideal that we have two routes to essentially the same resource (the difference is that the /blob URL returns the full attachment text, without the /blob you get the tiddler with a _canonical_uri field). Ideally we would use some form of content negotiation (perhaps based on a custom header).

I have been remiss in not documenting the routes that are currently supported. It may make sense to take some time to document them in order to make discussions about changing them easier.

PotOfCoffee2Go:
> we have two routes to essentially the same resource (the difference is that the /blob URL returns the full attachment text, without the /blob you get the tiddler with a _canonical_uri field).

For GIMP, Windows Media Player, VLC Media Player, etc.; they work using the blob URI. Although as mentioned, every attachment - be it an image, audio, or video - is named 'blob'.

What I have cobbled together that works well is :

`routes/handlers/get-bag-tiddler-blob.js` currently line 17
```js
exports.path = /^\/bags\/([^\/]+)\/attachment\/([^\/]+)$/;
```

`store/sql-tiddler-store.js` currently line 136
```js
_canonical_uri: `/bags/${$tw.utils.encodeURIComponentExtended(bag_name)}/attachment/${$tw.utils.encodeURIComponentExtended(tiddlerFields.title)}`
```

Which removes 'blob' and replaces 'tiddler' with 'attachment' - since the response is not a tiddler but a stream of the resource. The last element of the URI is the resource name which ends with .png, .jpg, .mpeg, mp3, etc. Allowing the media player to display the name of the resource along with the appropriate file extension.

PotOfCoffee2Go:
Hi @Jermolene 

There is a bug in the '--mws-save-archive' command `mws-save-archive.js` in the 'saveArchive(archivePath)' function, currently line 54.

```
const tiddlerInfo = $tw.mws.store.getBagTiddler(title,bagInfo.bag_name);
                                                ^^^^^
```
'title' is an object - the call requires a string.

```
const tiddlerInfo = $tw.mws.store.getBagTiddler(title.title,bagInfo.bag_name);
                                                ^^^^^^^^^^^
```
Would do a PR - but figured you could fix it next time working on the project.
 
thanks!

editions/multiwikiserver/tiddlers/$__StoryList_1.tid
```diff
@@ -0,0 +1,2 @@
+list: GettingStarted
+title: $:/StoryList
```
linonetwo, line 2: I don't like nodejs wiki creating `$__StoryList_1.tid` while I already git itnore the `$__StoryList.tid`. I ignore this because it cause too much git backup revision.

plugins/tiddlywiki/multiwikiserver/modules/commands/mws-load-wiki-folder.js
```diff
@@ -0,0 +1,93 @@
...
+				recipeList.push(makePluginBagName(type,publisher,name));
+			});	
+		};
+		processPlugins("plugin",wikiInfo.plugins);
+		processPlugins("theme",wikiInfo.themes);
+		processPlugins("language",wikiInfo.languages);
+		// Create the recipe
+		recipeList.push(options.bagName);
+		$tw.mws.store.createRecipe(options.recipeName,recipeList,options.recipeDescription);
+		$tw.mws.store.saveTiddlersFromPath(path.resolve($tw.boot.corePath,$tw.config.editionsPath,options.wikiPath,$tw.config.wikiTiddlersSubDir),options.bagName);
```
linonetwo, line 87: Is `saveTiddlersFromPath` defined? I can't find this method.
I wonder what it will do, can it add path info to the bag, so filesystem core plugin can use this info to handle save tiddler back to .tid file on the folder?

I also want to use this information to save different bags to different storages, some to public, some save to private.

plugins/tiddlywiki/multiwikiserver/modules/commands/mws-load-wiki-folder.js
```diff
@@ -0,0 +1,93 @@
...
+				}
+				recipeList.push(makePluginBagName(type,publisher,name));
+			});	
+		};
+		processPlugins("plugin",wikiInfo.plugins);
+		processPlugins("theme",wikiInfo.themes);
+		processPlugins("language",wikiInfo.languages);
+		// Create the recipe
+		recipeList.push(options.bagName);
+		$tw.mws.store.createRecipe(options.recipeName,recipeList,options.recipeDescription);
```
linonetwo, line 86: Seems this is in "create bag" command, why it create recipe here? Will it deduplicate when recipe exists by previous command?
Also I can't find `createRecipe` defined in the code.

editions/multiwikiserver/tiddlers/$__StoryList_1.tid
pmario: If it creates StoryList_1.tid, there is a bug in "file locking" mechanism. If the server is restarted and StroyList is still locked somehow, it will create the _1.tid file. 

**Instead of a work-around you should try to fix the underlaying problem**

pmario:
@Jermolene -- Which tools do you use to inspect the SQLite database. So I can see recipes, bags and tiddlers -- for dev testing

Jermolene:
> @Jermolene -- Which tools do you use to inspect the SQLite database. So I can see recipes, bags and tiddlers -- for dev testing

@pmario I use this: https://sqliteonline.com/

But I would love to spend a little time at some point making it possible to import arbitrary SQLite database files into TW as tiddlers, and then run SQL queries to extract/insert data as tiddlers. That would make a very nice extensible version of https://sqliteonline.com/ that users could customise with UI that directly reflects the schemas they are working with.

package.json
```diff
@@ -27,15 +27,23 @@
...
+    "mws-add-user": "node ./tiddlywiki.js ./editions/multiwikiserver --build load-mws-demo-data --mws-listen --build mws-add-user --quit",
+    "test": "npm run build:test-edition && npm run test:multiwikiserver-edition",
     "lint:fix": "eslint . --fix",
     "lint": "eslint ."
+  },
+  "dependencies": {
+    "@playwright/test": "^1.47.2",
+    "better-sqlite3": "^9.4.3",
+    "node-sqlite3-wasm": "^0.8.10",
+    "playwright": "^1.47.2"
```
pmario, line 47: Do we really need `@playwright/test` and `playwright` as project dependencies. I think they should be moved to devDependencies

package.json
```diff
@@ -27,15 +27,23 @@
     "eslint": "^9.12.0",
     "@eslint/js": "^9.12.0"
   },
-  "bundleDependencies": [],
   "license": "BSD",
   "engines": {
     "node": ">=0.8.2"
```
pmario, line 32: I think we should update minimum Node version. This one is "stone age"

bin/ci-test.sh
```diff
@@ -2,6 +2,8 @@
 
 # test TiddlyWiki5 for tiddlywiki.com
 
+npm install
```
pmario, line 5: In line 15 there is `npm install playwright ...` which is also added to package.json, which is installed here. -- I think line 15 is redundant

bin/readme-bld.sh
```diff
@@ -10,6 +10,7 @@ fi
 
 # tw5.com readmes
 node $TW5_BUILD_TIDDLYWIKI \
+	+plugins/tiddlywiki/multiwikiserver \
```
pmario, line 13: Why does readme-bld.sh need the multiwikiserver plugin?

package.json
Jermolene: That's very true, but needs to be addressed in "master", not here

bin/readme-bld.sh
Jermolene: It's a temporary measure while the PR is in development. It allows us to include content from the MWS plugin in the readme.

package.json
```diff
@@ -27,15 +27,23 @@
...
+    "test": "npm run build:test-edition && npm run test:multiwikiserver-edition",
     "lint:fix": "eslint . --fix",
     "lint": "eslint ."
+  },
+  "dependencies": {
+    "@playwright/test": "^1.47.2",
+    "better-sqlite3": "^9.4.3",
+    "node-sqlite3-wasm": "^0.8.10",
+    "playwright": "^1.47.2"
   }
```
pmario, line 48: better sqlite3 version does not install on my Windows machine using Node.js 22.10.0
Version 11.5.0 which is latest does install.

pmario: better-sqlite3 v11.0.0 -- dropped support for Node.js v21 -- see: https://github.com/WiseLibs/better-sqlite3/releases/tag/v11.0.0

plugins/tiddlywiki/multiwikiserver/modules/routes/handlers/manage-user.js
```diff
@@ -0,0 +1,80 @@
...
+	
+		// Render the html
+		var html = $tw.mws.store.adminWiki.renderTiddler("text/plain", "$:/plugins/tiddlywiki/multiwikiserver/templates/page", {
+			variables: {
+				"page-content": "$:/plugins/tiddlywiki/multiwikiserver/templates/manage-user",
+				"user": JSON.stringify(user),
+				"user-initials": user.username.split(" ").map(name => name[0]).join(""),
+				"user-role": JSON.stringify(userRole),
+				"all-roles": JSON.stringify(allRoles),
+				"is-current-user-profile": state.authenticatedUser && state.authenticatedUser.user_id === $tw.utils.parseInt(user_id, 10) ? "yes" : "no",
```
pmario, line 71: @webplusai - If you use `$tw.utils.parseInt(str)` there is no need to add the radix param. Since the utility function takes care of it. See: https://github.com/TiddlyWiki/TiddlyWiki5/blob/c7531e53abbe065d9d1a9fe9959f88b39d1a98f4/core/modules/utils/utils.js#L980-L982

CrossEye: JFTR, you don't need it with the language's `parseInt` either, if you're calling it as a global function, where the radix defaults to `10`.  Of course if you're passing a reference to it into a function that calls it with additional parameters, you can be in for a surprise or two.  The classic example is:

```js
['1.1', '2.2', '3.3'].map(parseFloat) //=> [1.1, 2.2, 3.3]
// but
['1', '2', '3'].map(parseInt) //=> [1, NaN, NaN]
// hmm
```

That is not an issue, though with the current usage, either of the native version or the TW one.

MagicJF:
> @Jermolene -- Which tools do you use to inspect the SQLite database. So I can see recipes, bags and tiddlers -- for dev testing

I found this book that seems a good resource regarding SQLite
https://lessonomicon.github.io/querynomicon/s02_core/index.html

HwWobbe:
It's nice to see this here and to have a chance to use and use this variant now.

pmario:
@Jermolene -- Where can I change the content which is published at mws.tiddlywiki.com? There are some commands missing at the Installation descriptions.

Jermolene:
> @Jermolene -- Where can I change the content which is published at mws.tiddlywiki.com? There are some commands missing at the Installation descriptions.

Hi @pmario mws.tiddlywiki.com is built from the edition at https://github.com/TiddlyWiki/TiddlyWiki5/tree/multi-wiki-support/editions/multiwikidocs

pmario:
@Jermolene -- This may be of interest: https://nodejs.org/docs/v22.12.0/api/sqlite.html#sqlite

Jermolene:
> @Jermolene -- This may be of interest: https://nodejs.org/docs/v22.12.0/api/sqlite.html#sqlite

Goodness it will make life easier to have sqlite integrated into Node.js, excellent. When is it scheduled to hit the release version of Node.js?

pmario:
At the moment it is [Stability Index: 1.1 -- Active Development](https://nodejs.org/api/sqlite.html#sqlite)

See: https://nodejs.org/api/documentation.html#stability-index

pmario:
I did a test with Deno's SQLite module, but it crashes -- The API seems to be incompatible.

Arlen22:
The MWS plugin has now been moved to https://github.com/TiddlyWiki/MultiWikiServer

Arlen22:
@Jermolene, what would it take to point mws.tiddlywiki.com at the new repo? Or are you wanting to keep it on the main site? Is that something you would be able to set up?

Jermolene:
Closing this PR in favour of https://github.com/TiddlyWiki/MultiWikiServer

Jermolene:
> @Jermolene, what would it take to point mws.tiddlywiki.com at the new repo? Or are you wanting to keep it on the main site? Is that something you would be able to set up?

Apologies @Arlen22 I missed this comment when I opened https://github.com/TiddlyWiki/MultiWikiServer/issues/26, I'll respond there

Issue #7917, opened 2024-01-05
[BUG] The core search result list should return relevant info for basic search terms - early


pmario:
As the title says: "The core search result list should return relevant info for basic search terms - early"

What are basic search terms for new users. 

**tiddler, wiki, tiddlywiki, filter, widget, macro, list, theme, procedure, template, tab, json, data, plugin, save, import, image, edition, link, field, tag, wikitext,** ... 

The following screenshots show 45 results for every basic term mentioned above. 

- Left: New proposed filter string `[!is[system]search:title<userInput>] :sort:integer[levenshtein<userInput>] :and[limit[250]]`
- Right: TW v5.3.3 alphabetically sorted

You can decide for yourself, what makes more sense. 

![01-tiddler](https://github.com/Jermolene/TiddlyWiki5/assets/374655/9b2f4db8-3eb8-450c-a55b-5206946f5678)
![02-wiki](https://github.com/Jermolene/TiddlyWiki5/assets/374655/8328f424-13df-4e31-a00b-248e0902b22a)
![03-tiddlywiki](https://github.com/Jermolene/TiddlyWiki5/assets/374655/0e1d7b9d-e9c0-48f2-be6c-df06bdaac61d)
![04-filter](https://github.com/Jermolene/TiddlyWiki5/assets/374655/3fb25d04-3017-4ab2-aa80-c2a87357971f)
![05-widget](https://github.com/Jermolene/TiddlyWiki5/assets/374655/114b47a5-ed78-4380-b015-209d75309a29)
![06-macro](https://github.com/Jermolene/TiddlyWiki5/assets/374655/02dc47d2-85f5-4fab-828d-6fb3eaccc28f)
![07-list](https://github.com/Jermolene/TiddlyWiki5/assets/374655/75c12680-887a-4815-be44-5c429d9186f0)
![08-theme](https://github.com/Jermolene/TiddlyWiki5/assets/374655/fdceb131-992b-4894-a140-62189636ceb5)
![09-procedure](https://github.com/Jermolene/TiddlyWiki5/assets/374655/15109b5f-4b37-4d12-aa17-d4e72211fa38)
![10-template](https://github.com/Jermolene/TiddlyWiki5/assets/374655/b98b5c91-86ff-4f5d-a58b-aba849c65ad8)
![11-tab](https://github.com/Jermolene/TiddlyWiki5/assets/374655/de7bace4-91df-40fd-951a-d4593e7cf1c8)
[...]

PR #7918, opened 2024-01-05
Use Levenshtein distance to order search results

pmario:
This PR fixes: **#7917 [BUG] The core search result list should return relevant info for basic search terms - early** 

It replaces the default filter string in `$:/core/ui/DefaultSearchResultList`

- replace: `[!is[system]search:title<userInput>sort[title]limit[250]]`
- with: `[!is[system]search:title<userInput>] :sort:integer[levenshtein<userInput>] :and[limit[250]]`

It may not be optimal, but I think it's a huge improvement. 

**All Example Screenshots** can be seen at: #7917 -- I'll only add one here for **tiddler**

01-tiddler

### DIFF
--- a/core/ui/DefaultSearchResultList.tid
+++ b/core/ui/DefaultSearchResultList.tid
@@ -1,7 +1,7 @@
 title: $:/core/ui/DefaultSearchResultList
 tags: $:/tags/SearchResults
 caption: {{$:/language/Search/DefaultResults/Caption}}
-first-search-filter: [!is[system]search:title<userInput>sort[title]limit[250]]
+first-search-filter: [!is[system]search:title<userInput>] :sort:integer[levenshtein<userInput>] :and[limit[250]]
 second-search-filter: [!is[system]search<userInput>sort[title]limit[250]]
 
 \define searchResultList()


Jermolene:
Thanks @pmario it is useful to be able to try it out.

I find that it improves the quality of the first few search results, but makes it much, much harder to visually scan the main body of results because they are no longer alphabetised.

image

I think that this Levenshtein technique might be useful but not sufficient to fix our search shortcomings. I wonder if it might be useful for picking out a small number of top matches, for example.

pmario:
>I find that it improves the quality of the first few search results, but makes it much, much harder to visually scan the main body of results because they are no longer alphabetised.

We do have a completely "contrary view" here. For me personally the shorter results are easier to scan and most of the time reflect the context. 

For me the alphabetical order destroys the context and makes the search results mainly useless.

mateuszwilczek:
I think we should look at two uses of search.

One is "navigating", that is, getting to a tiddler, which name we more or less know. This would be my most frequent use of search. In this case, having the closest matches on top is a great convenience.

The second would would be "browsing", looking through a list of tiddlers, but not precisely knowing which one we want to open. In this case, having a longer list of titles organized in some way (alphabetical) is useful. I rarely do this, and if so, then often using filter search rather than standard search. The "closest matches" loose their meaning a bit in this case.

I agree with Jeremy that looking through a longer Levensthein-sorted list is impractical. But I do not see it as a disadvantage of this sorting, because I expect the result of interest to be near the top anyway (or I refine the search until it is).

Maybe a viable compromise would be to split the results not in two, but in three segments:

1. *Top title matches*. First 5 closest matches with the Levenshtein sorting (or whatever better we conceive in the future).
2. *Further title matches*. The rest of title matches, sorted alphabetically. Should this part be very short, like 5, then make the first segment longer (up to 10) and do not show the second segment.
3. *All matches* without changes.

This should satisfy both conflicting needs quite well. If I know what I want to navigate to, I can get it to show in the top 5 result without much effort. If I want to browse a list of 50 titles, I think that I can get around reading the first 5 as a separate list before the next 45 alphabetically sorted ones.

It would be far from ideal, but possible right away, and still much better that the current search.

Jermolene:
Thanks @pmario @mateuszwilczek, I like @mateuszwilczek's proposal.

saqimtiaz:
> Maybe a viable compromise would be to split the results not in two, but in three segments:
> 
> 1. _Top title matches_. First 5 closest matches with the Levenshtein sorting (or whatever better we conceive in the future).
> 2. _Further title matches_. The rest of title matches, sorted alphabetically. Should this part be very short, like 5, then make the first segment longer (up to 10) and do not show the second segment.
> 3. _All matches_ without changes.

I second this proposal. I do something similar in my own notes wiki with exactly those three groups of matches, though using a custom sort instead of Levenshtein as it predates the operator and also because it priorities results that are prefixed with the search term.

pmario:
>1. Top title matches. First 5 closest matches with the Levenshtein sorting (or whatever better we conceive in the future).
>2. Further title matches. The rest of title matches, sorted alphabetically. Should this part be very short, like 5, then make the first segment longer (up to 10) and do not show the second segment.
>3. All matches without changes.

Changing the DefaultSearchResultList into 3 or more different segments will be a backwards incompatible change.

The current implementation of the [DefaultSearchResultList](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FDefaultSearchResultList) converts the "shadow" tiddler into a "real tiddler", if the one of the two search filter strings is changed. This prevents automatic core updates. 

So if users did ever change the  `first-search-filter` or `second-search-filter` fields to optimize their search results, they will have problems with a future update. 

--------

The default search dropdown already allows us to [configure additional search-result-lists](https://tiddlywiki.com/#Customising%20search%20results) using the `$:/tags/SearchResults`.

So as I see it, we'll need to go the plugin route, which adds such a new tab to the search-dropdown in a backwards compatible way. That's not pretty, but does not need any change. 

I'll close this PR.

-------

I may have a closer look about a new option for my own [field-search plugin](https://wikilabs.github.io/editions/field-search/), which will allow me to also search in fields and titles.

yaisog:
> So as I see it, we'll need to go the plugin route, which adds such a new tab to the search-dropdown in a backwards compatible way. That's not pretty, but does not need any change.

Hi @pmario, I know I'm more than a year late, but I only stumbled upon this PR from a recent discussion on TalkTW.

Firstly, when making any change to a shadow tiddler, the user risks having to patch in future updates manually. We all try to keep the number of such overridden shadows small in our wikis for that reason. That doesn't mean that there can be no progress in TW development, just because someone may have overwritten a shadow. We have enough worries about backwards compatibility without such considerations already.

Secondly, these filters should be moved to $:/config tiddlers, not hardcoded in the search result tiddler.

Thirdly, can't we implement your changes in a _new_ tiddler that is called something other than `$:/core/ui/DefaultSearchResultList` and remove the `$:/tags/SearchResults` tag from the original shadow tiddler? If someone has overridden the shadow, he/she will see tabs with both search results lists, all others will only see the new results. Those who have hardcoded the $:/core/ui/DefaultSearchResultList name somewhere will still see the "legacy" search results.

This change would be so beneficial for many users, even @Jermolene liked the concept of it. Please consider re-opening the PR.

yaisog:
Also, assuming that only _title matches_ would be considered for potential _top matches_: In this case, the `levenshtein` sorting is almost identical to sorting by title length. `levenshtein` will give the number of changes (in this case additions of characters) that are necessary, which is equal to the total number of characters minus the search term.

Sorting by title length should be much faster than calculating the Levenshtein distance, with pretty much the same results (if the search term(s) appear(s) in the title.

Jermolene:
Thanks @yaisog I do indeed still support @mateuszwilczek's proposal in https://github.com/TiddlyWiki/TiddlyWiki5/pull/7918#issuecomment-1879221858

PR #7919, opened 2024-01-06
fix archive output paths

pmario:
This PR should fix the archive output paths mentioned at: https://github.com/Jermolene/TiddlyWiki5/pull/7776#issuecomment-1879271524

Please double-check. 

@Jermolene ... It should be easy to move the existing files from https://github.com/Jermolene/jermolene.github.io/tree/master/archive to "/full and /empty" by hand. 

I did test them in the archive/ dir and they seem to work as intended -- only in the wrong dir.


pmario:
index.html and static.css need to stay in the archive directory as an index file.

Jermolene:
Thanks @pmario apologies for the delay.

PR #7927, opened 2024-01-11
fix: add doc xls ppt type

linonetwo:
Add type for doc ppt xls, so they are recognized by `[is[binary]]`

But I also find some xls will be type `application/vnd.ms-excel`. I can't assign it because its extension is also `.xls`, and `registerFileType` don't allow multiple type have same extension.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -2455,6 +2455,7 @@ $tw.boot.initStartup = function(options) {
 	$tw.utils.registerFileType("application/msword","base64",".doc");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.spreadsheetml.sheet","base64",".xlsx");
 	$tw.utils.registerFileType("application/excel","base64",".xls");
+	$tw.utils.registerFileType("application/vnd.ms-excel","base64",".xls");
 	$tw.utils.registerFileType("application/vnd.openxmlformats-officedocument.presentationml.presentation","base64",".pptx");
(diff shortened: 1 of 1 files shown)


saqimtiaz:
You can register multiple types for the same extension, see https://github.com/Jermolene/TiddlyWiki5/blob/e26175e9c54f93ee4ac05aeecb32d9572e97cb23/boot/boot.js#L2417-L2418C68

linonetwo:
@saqimtiaz I think that usage is wrong, the previous one will immediately be overwritten:

https://github.com/Jermolene/TiddlyWiki5/blob/e26175e9c54f93ee4ac05aeecb32d9572e97cb23/boot/boot.js#L548

pmario:
no it will not be overwirtten. The register function creates arrays of types

linonetwo:
It is a dict!

```
$tw.config.fileExtensionInfo[extension]
```

With typescript this will be crystal clear

```ts
fileExtensionInfo: Record<string, { type: string }>
```

saqimtiaz:
Try this in the console in TiddlyWiki:

`$tw.config.contentTypeInfo`

You will see that multiple types can have the same extension because the  **type** is the key, not the extension.

linonetwo:
I see, there are `fileExtensionInfo` and `contentTypeInfo`, I was only looking at `fileExtensionInfo`.

Anyway, I add it.

Jermolene:
Thanks @linonetwo

PR #7939, opened 2024-01-19
add unusedtitle macro tests - should have full code covery

pmario:
This PR adds new tests for the `<<unusedtitle>>` macro, to be able to implement future changes without breaking the existing behaviour.

PR #7940, opened 2024-01-19
moves some exsiting macro tests to their own folders

pmario:
This PR does only move some existing macro tests to their own folders for better readability.


pmario:
slightly related to:  **add unusedtitle macro tests - should have full code covery** #7939

Jermolene:
Thanks @pmario

PR #7943, opened 2024-01-22
Refreshes edit widget when the editor type has changed

saqimtiaz:
This PR improves the refresh handling of the edit widget so that changes in the corresponding `$:/config/EditorTypeMappings` tiddler redraws the editor. This allows for example creating an editor button that can toggle between the default text editor and codemirror.


saqimtiaz:
@Jermolene the previous implementation handled the case where the `type` field of the editor being edited was changed. However, if the editor type mapping tiddler was changed, any rendered editors were not refreshed. 

This PR is prompted by a community discussion around toggling between CM6 and the default editor. I am leaving this is a draft until there is feedback and confirmation that the change is helpful and has no untoward consequences.

oeyoews:
Thanks, I have implemented this function in cm6 through subclass @saqimtiaz

Jermolene:
Great thanks @saqimtiaz

Issue #7740, opened 2023-09-13
@rmunn's open pull requests


rmunn:
Since I have [several open PRs](https://github.com/Jermolene/TiddlyWiki5/pulls/rmunn), in various states of progress, I figured I'd use an issue to help keep track of all of them.

I will edit this issue description as the status of my PRs changes.

**Ready for review & merge**:

- https://github.com/Jermolene/TiddlyWiki5/pull/7690
- https://github.com/Jermolene/TiddlyWiki5/pull/7702
- https://github.com/Jermolene/TiddlyWiki5/pull/7712
- https://github.com/Jermolene/TiddlyWiki5/pull/7715

**More work planned**:

- https://github.com/Jermolene/TiddlyWiki5/pull/6896
- https://github.com/Jermolene/TiddlyWiki5/pull/7683
- https://github.com/Jermolene/TiddlyWiki5/pull/7694

**On hold pending discussion**:

- https://github.com/Jermolene/TiddlyWiki5/pull/7691

CodaCodr:
It's a useful technique we use in our shop. We prefix the title with [META] or [METABUG]. Mozilla (Bugzilla) does the same.

Jermolene:
Thanks @rmunn – I've merged #7703, #7712, and #7715, and raised a question at #7690.

#6896 and #7683 look good so far. @7694 will need a tiny bit of extra work if we opt for #7710 instead of #7691

rmunn:
#7690 is ready to merge now; your suggestion was a good one and I've implemented it.

I've also created #7755 which targets not `master`, but the target branch of your PR #7698. More details over at #7755.

pmario:
All PRs have been merged or closed. One of them is draft. So I'll close this one. You can create a new issue for your current PRs if you need it.

PR #7944, opened 2024-01-23
Add trimSelection parameter to text editor

pmario:
As discussed at: https://talk.tiddlywiki.org/t/ideas-to-improve-wrap-selection-and-wrap-lines-operations-tm-edit-text-operation/8935

This PR adds a new parameter to the "wrap-selection" editor button functions. 

```
title: $:/core/ui/EditorToolbar/bold
```
```
<$action-sendmessage
    $message="tm-edit-text-operation"
    $param="wrap-selection"
    prefix="''"
    suffix="''"
    trimSelection="yes"  <!-- no (default), yes, start, end - are possible -->
/>
```

wrap-selection-01


pmario:
I also did a short test with CodeMirro5 plugin -> Seems to work as expected.

mateuszwilczek:
@pmario it works as expected with the core toolbar buttons in default editor. However, I cannot make it work with my custom buttons that add markup other than the core buttons. I tried the following in the preview:

* adding `trimSelection="both"` to my custom button
* cloning and modifying a core button, e.g. bold, so that it adds custom markup, e.g. `@@` prefix and suffix
* I tried saving and reloading in each case

The "ignoring" of spaces does not work, so if `word ` is selected, then it is changed to `@@word @@`. However, removing of markup even if the inline markup is not selected does work, so if only `word ` inside `@@word @@` is selected, it will be changed to `word ` as expected.

Is your current solution by any chance hard coded to support only the markup symbols present in core toolbar buttons, but not any arbitrary prefix/suffix? The ignoring of spaces in selection does work if I change the prefix/suffix to something wild that uses "default" symbols only, e.g. `///'_,~`, but does not work with `@@`.

pmario:
@mateuszwilczek 

- go to https://tiddlywiki5-git-fork-pmario-trim-selection-tex-1760fa-jermolene.vercel.app/#How%20to%20create%20dynamic%20editor%20toolbar%20buttons
- click the "Create temporary bold button"
- change prefix and suffix to #7740 
- Save the tiddler

It should work as expected. I did just test it and it works for me.

mateuszwilczek:
@pmario I got where the problem is. There's a typo in the parameter name, `c` is missing in `trimSeletion="both"`.
In my attempts I was copying the parameter from the first post here `trimSelection="both"`, which has no typo.

mateuszwilczek:
Since the only problem I found has been identified, I think its a good improvement. I like that it works on all whitespace, not just the single trailing space selected by double click on Windows. So it also works in cases with a lot of different whitespace characters on both ends of the selection.
E.g. this image becomes this image


As to the name of the parameter and possible values, perhaps something like `skipWhitespace` would be more clear? As you pointed out, it does not "trim" or "remove" the spaces, just "skips" or "ignores" or "omits" the surrounding spaces when applying markup.
The possible values would be default `no` (for backwards compatibility) and `yes`, instead of `both`. I don't think that the specific start/end only options are necessary.

The only thing I could wish for, would be to identify markup to remove if it is *partially* selected. Core removes markup if all of it is selected, e.g. `//example//`. Your PR makes it so that markup is removed if only its "content" is selected, e.g. //`example`//.
It would be nice if partial selection, e.g. /`/example/`/ or //`example//` would also make the markup be removed. I realize it might be difficult to implement, and as said, this would be a luxury to have. The features that you have already implemented are much more important.

pmario:
>The possible values would be default no (for backwards compatibility) and yes, instead of both. I don't think that the specific start/end only options are necessary.

We could change `both` to `yes`. But I think `start` and `end` should stay. I'm pretty sure users will have usecases where they will make sense. We just don't know them yet.

hmmm `skipWhitespace` may be an option.

@Jermolene @saqimtiaz ... any ideas?

pmario:
I did change the prameter "both" to "yes" - Switch this PR to "ready for review" since there was no other feedback

core/ui/EditorToolbar/underline.tid
```diff
@@ -11,4 +11,5 @@ shortcuts: ((underline))
 	$param="wrap-selection"
 	prefix="__"
 	suffix="__"
+	trimSeletion="yes"
```
saqimtiaz, line 14: @pmario this typo repeats itself and makes it difficult to test.

pmario: The typo has been everywhere, so the function was working but with the typo in the parameter. I changed it. So it should be fixed now.

pmario:
@Jermolene -- please review again. IMO this PR is a nice quality of live improvement, which also fixes a constant annoyance if macro parameters are double-clicked and renamed.

core/modules/editor/operations/text/wrap-selection.js
```diff
@@ -13,37 +13,121 @@ Text editor operation to wrap the selection with the specified prefix and suffix
...
+	var o = operation,
+		prefix = event.paramObject.prefix,
+		prefixLength = prefix.length,
+		suffix = event.paramObject.suffix,
+		suffixLength = suffix.length,
+		trimSelection = event.paramObject.trimSelection || "no",
+		selLength = o.selEnd - o.selStart;
+
+	var trailingSpaceAt = function(sel) {
+		// returns "yes", "start", "end", "no"
```
Jermolene, line 25: It's a small point, but we normally capitalise comments. It does aid readability.

I think some more comments might be helpful here. What does this function do? What do the results mean? Particularly "no".

core/modules/editor/operations/text/wrap-selection.js
```diff
@@ -13,37 +13,121 @@ Text editor operation to wrap the selection with the specified prefix and suffix
 "use strict";
 
 exports["wrap-selection"] = function(event,operation) {
-	if(operation.selStart === operation.selEnd) {
-		// No selection; check if we're within the prefix/suffix
-		if(operation.text.substring(operation.selStart - event.paramObject.prefix.length,operation.selStart + event.paramObject.suffix.length) === event.paramObject.prefix + event.paramObject.suffix) {
+	var o = operation,
```
Jermolene, line 16: I'm not a huge fan of this abbreviation pattern. I understand that it saves a bit of space, but it is hard to apply consistently. For example, here one might wonder why "operation" is abbreviated but "prefix" is not.

On balance, I would prefer to find ways to refactor the code to avoid the repetition, rather than use this pattern.

core/modules/editor/operations/text/wrap-selection.js
```diff
@@ -13,37 +13,121 @@ Text editor operation to wrap the selection with the specified prefix and suffix
...
+	var trailingSpaceAt = function(sel) {
+		// returns "yes", "start", "end", "no"
+		var _start,
+			_end,
+			result;
+		// this evaluation takes the user configuration into account!
+		switch (trimSelection) {
+			case "end":
+				result = (sel.trimEnd().length !== selLength) ? "end" : "no";
+			break;
```
Jermolene, line 33: The "break" should be indented one more step

core/modules/editor/operations/text/wrap-selection.js
```diff
@@ -13,37 +13,121 @@ Text editor operation to wrap the selection with the specified prefix and suffix
...
+			default:
+				result = "no";
+			break;
+		}
+		return result;
+	}
+
+	function togglePrefixSuffix() {
+		// this was the only behavour till TW v5.3.3
```
Jermolene, line 50: I am not sure that comments like this are helpful. Most people looking at the code will have access to Git history if they have questions about the past. It is better to optimise the code readability experience for new people.

core/modules/editor/operations/text/wrap-selection.js
```diff
@@ -13,37 +13,121 @@ Text editor operation to wrap the selection with the specified prefix and suffix
...
+			o.newSelStart = o.selStart + prefixLength;
+			o.newSelEnd = o.newSelStart;
 		}
-	} else if(operation.text.substring(operation.selStart,operation.selStart + event.paramObject.prefix.length) === event.paramObject.prefix && operation.text.substring(operation.selEnd - event.paramObject.suffix.length,operation.selEnd) === event.paramObject.suffix) {
+	}
+
+	// options: prefixLen, suffixLen
+	function removePrefixSuffix(options) {
+		options = options || {};
+		var prefixLen = options.prefixLen || 0;
```
Jermolene, line 71: Using "Len" as a suffix is inconsistent with the code above. In a situation like this I'd prefer "lenPrefix".

Jermolene (commented):
Thanks @pmario just some coding style points

core/modules/editor/operations/text/wrap-selection.js
pmario: prefix is an abbreviation. It is the shorthand for `event.paramObject.prefix`

pmario: The "o" abbreviation is used 99 times and was introduced to make the code readable and understandable. If I add it back again, it will make the code completely unreadable. Reducing the file size by 800bytes is a side effect. 

IMO there is no redundant, repettitive code on the right side of any =. Every line, that defines o.cutStart, o.cutEnd, ... and so on, has a different pattern on the right side. -- So I do not really see a possility to optimize that any further.

pmario:
I did test it with the default and the codemirror editors. Both work in the same way.

Jermolene:
Thanks @pmario

PR #7949, opened 2024-01-27
fix: on Windows, tiddlywiki.files title source shoud use / instead of \

linonetwo:
On windows, `path.relative(rootPath, filename)` will produce `items\comestibles\raw_fruit.json`.

Add `.split(path.sep).join('/')` fix this to be `items/comestibles/raw_fruit.json`

Before:

图片

After:

图片

The `tiddlywiki.files` file I'm using:

https://github.com/tiddly-gittly/tw-gamification/blob/1d9eaf87019a6f4801ef4bdb09b555551d3bcbf8/src/scp-foundation-site-director/game/tiddlywiki.files


Jermolene:
Thanks @linonetwo looks good

Issue #4224, opened 2019-08-31
Update a timestamp tiddler when a wiki is saved


Jermolene:
As [discussed on the mailing list](https://groups.google.com/d/msgid/tiddlywiki/CAGUDtYE0xRv2xvyxXBvWaSO-WPJp8bf620vw8p27h15mRxSj8w%40mail.gmail.com), it would be useful if a system tiddler were automatically updated with the date and time of the last save.

AnthonyMuscio:
And display the result in $:/ControlPanel > Info > Basics
and/or $:/ControlPanel > Saving > General

hchaase:
Yes, this would be useful! 
To see when the last save was done I quite frequently check the more>recent tab.

I second the suggestion in the malign list: "I can envision that timestamp being transcluded in "Tool" right under the wiki version number AND in Control Panel along with number of tiddler & overwritten shadows."

**extra feature**: it could be handy (though not needed) if the modification date (as presented above) was a link to the last edited tiddler.

**edge case**: what would happen when you turn timestamp off and save a tiddler (e.g. to correct spelling), would that change the timestamp?

Jermolene:
> edge case: what would happen when you turn timestamp off and save a tiddler (e.g. to correct spelling), would that change the timestamp?

I would envisage the timestamp being generated at the instant of saving, independent of the state of the tiddler store.

> it could be handy (though not needed) if the modification date (as presented above) was a link to the last edited tiddler.

That would be possible, but I'm not sure very obvious?

pesho-ivanov:
Along with time stamping, consecutive "version numbers" may simplify spotting the problem of overwriting from an older wiki state (thus losing data).

AnthonyMuscio:
Is it possible to make this happen?, it will be a complement to the newish [$:/info/startup-timestamp](https://tiddlywiki.com/#%24%3A%2Finfo%2Fstartup-timestamp) because we can actually identify wikis opened, have changes and when and last saved. 

```
\function last-modified() [all[tiddlers]get[modified]!sort[]first[]format:date[0hh:0mm DDD, DD. MMM YYYY]]
```
* This is a function that determines the last modified tiddler (only) but I wonder if it would be better set somewhere in the core for performance reasons?
* The question remains what if I change a system tiddler?

The key issues are in my mind;
* Will it only update if the save is successful?
  * If a save was successful then we can determine the wiki is not read only
[...]

Issue #6979, opened 2022-10-10
[IDEA] Autosave frequency throttle


simonbaird:
**Is your feature request related to a problem? Please describe.**

I suspect that some brand new TiddlyWiki users on https://tiddlyhost.com don't realize they need to hit the save button to save their changes. I'd like to enable autosave for that reason, but with a throttle so that it autosaves only when it's more than (say) two minutes since the last save.

**Describe the solution you'd like**

I'm imagining a new configuration option related to autosave, called something like "Minimum seconds between autosaves" or "Autosave throttle (minimum seconds between autosaves)". The value would be an integer number of seconds stored in a tiddler named `$:/config/AutoSavePeriod` or something like that.

The Autosave mechanism would change so if that value is present and non-zero it would not trigger an autosave until the time since the last save is more than the threshold. Of course multiple autosave triggers would need to not queue up, so when the time comes only one save event happens.

**Describe alternatives you've considered**

My current workaround is to disable autosave on Tiddlyhost since I don't want every single change to trigger a save. The reasoning there is that: a). I think it would create a significantly higher load on Tiddlyhost's server, and b). since Tiddlyhost does a full file save every time there's a wait between "Save started..." and "Save complete". This is a bad UX at best and at worst it could result in race condition problems where different save request are happening at the same time.

I think this idea would be generally useful, rather than being a feature just for Tiddlyhost, so that's why I'm suggesting it as a TiddlyWiki feature.

**Additional context**

There may be some overlap with #6684 here.

Actually I'm not 100% this feature doesn't exist already, so please let me know this can be done already, or if there's a well known plugin that can do it.

simonbaird:
Hypothetically, if I did some work on implementing this feature, would it be considered for merge?

Jermolene:
Hi @simonbaird sorry for the delay, that's a great idea, and a PR would be appreciated.

simonbaird:
I noticed there's something on TiddlyTools that might be relevant here, see https://tiddlytools.com/#TiddlyTools%2FTime%2FAutoSaver .

saqimtiaz:
I have taken a quick crack at an implementation in #7995

simonbaird:
See also [this forum post](https://talk.tiddlywiki.org/t/making-autosave-smarter-testing-a-new-implementation/13099).

twMat:
[...]

PR #7953, opened 2024-01-31
6684 add last saved timestamp

### LINKED ISSUE #6684: [IDEA] Last saved timestamp and what it may enable

pmario:
This PR fixes #6684 and #4224

Initially I intended to add 2 new `$:/Info/` tiddlers to the wiki-store when the wiki is saved. 

- $:/info/last-saved/timestamp ... easy to process by software
- $:/info/last-saved/date ... easy to read for humans

While testing I found out, that those 2 tiddlers where also modified with `tiddlywiki . --build index` command **but** using the server timestamp. So I did add this info to 2 new tiddlers. **IMO** all 4 of them have value. 

- $:/info/server-created-wiki/timestamp
- $:/info/server-created-wiki/date

Screenshot of the new docs info which is part of the PR

image

I did test the mechanism 

- using a client-server configuration and
- create a snapshot
- using a `tiddlywiki . --build index` command and
- save the wiki as a single-file wiki

Everything worked fine for me. When the wiki is loaded, the new tiddlers become part of the wiki-store, but they will be overwritten with every automatic or manual save action.


Jermolene:
Hi @pmario given that this change doesn't satisfy the requirements of the original requester (see https://github.com/Jermolene/TiddlyWiki5/issues/4224#issuecomment-1922729285), I propose that we shelve this for the moment. It makes the saving mechanism significantly more complex for everyone who is trying to understand things, and it is not clear that it is needed by anyone.

pmario:
@AnthonyMuscio ... Did you ever try this preview? It seems you have missed this PR? I think it does meet your requirements.

Jermolene:
Hi @pmario it may be worth bringing this up at talk.tiddlywiki.org to see if there's any mileage in it. Without a clear use case I am not keen to merge this because of the complexity it introduces.

pmario:
I converted this one into a draft, since I do need to read the different threads, that are

-  here at GH (#6684) , 
- at Talk 
    - https://talk.tiddlywiki.org/t/action-on-return-to-tiddlywiki/3281
    - https://talk.tiddlywiki.org/t/technical-discussion-javascript-access-to-document-and-window-information/5870 
- in the [GoogleGroups](https://groups.google.com/g/tiddlywiki/c/RxJdciY7yZY/m/90eoO2yfAAAJ) mentioned in #4224

Also see: **[IDEA] Autosave frequency throttle #6979** is also a related request, that probably could be solved by this mechanism, or the filters I did mention at: https://github.com/Jermolene/TiddlyWiki5/issues/4224#issuecomment-1921051773

-------------

The very first PR: **Record state saving timestamp #7131**

- was initially hardcoded which was very usecase specific
   - See the very first commits in the PR 
- The action-based flexible version was prone to create endless save-loops

core/templates/tiddlywiki5.html.tid
```diff
@@ -41,6 +41,7 @@ title: $:/core/templates/tiddlywiki5.html
 </noscript>
 <!--~~ Ordinary tiddlers ~~-->
 `{{$:/core/templates/store.area.template.html}}`
+`{{$:/core/templates/store.info.template.html}}`
```
pmario, line 44: @Jermolene -- I would like to convert this PR into a plugin.

The absolute minimum required in the core would be this one line. So the whole mechanism would be similar to ` $:/tags/RawMarkup*` behaviour. 

To make it less error prone we should also keep the template `$:/core/templates/store.info.template.html` tiddler in the core, which defines how to handle "injected" tiddlers. 

What do you think?

Jermolene: Hi @pmario that makes sense. Rather than adding another hardcoded template title we could use a filtered transclusion to bring in any tiddlers with the appropriate tag?

pmario: IMO that's exactly what: https://github.com/Jermolene/TiddlyWiki5/blob/7b0f23b2cc05c0fead63ec7265a5d54d940b2d52/core/templates/store.info.template.html.tid does.

Jermolene: Hi @pmario if it was a generic mechanism for splicing tiddlers into saved wikis then it probably shouldn't have the name "info" in the title.

pmario: It basically is an independent "store-area" to minimise side effects, if users mess it up. So may be we could use `$:/core/templates/custom.store.template.html` or  `$:/core/templates/store.custom.template.html` if we want it to start with `/store.`
 
The tag at the moment is: `$:/tags/InfoTiddlerTemplate` which also would need to be renamed eg: `$:/tags/CustomStore`

So the code in the core template would look like this:

```
<!--~~ Ordinary tiddlers ~~-->
`{{$:/core/templates/store.area.template.html}}`
`{{$:/core/templates/custom.store.template.html}}`
```

Jermolene: Hi @pmario I did notice the use of a separate store area. I think that would mean that these tiddlers wouldn't be visible to the import process?

pmario: I'm not sure, what you mean with "import process". 

When the wiki is reloaded, they will be "merged" into the main wiki-store as standard tiddlers.

Jermolene: I meant that an attempt to import the saved wiki (eg via drag and drop) would not see the inserted tiddlers.

pmario: OK - I did not think about that. I'll need to test this.

pmario: - I did download the [preview wiki](https://tiddlywiki5-git-fork-pmario-6684-add-last-save-16b5df-jermolene.vercel.app/)
- I did save empty from the preview
- Open empty.html and -> save it so new tiddlers are injected
- Did open empty.html in **text editor**
- Main store: `last-saved/date: 2024-03-10 11:11:42.171 UTC +01:00` see: 11:11
- Custom store exists with `last-saved/date: 2024-03-10 11:16:31.158 UTC +01:00` see: 11:16
- Import into tiddlywiki.com -> **seems to work** as expected. 
- image

Jermolene: Thanks @pmario I had a look at the code in `$:/core/modules/deserializers.js` and it does indeed process multiple new style store areas.

pmario: So -- which "generic" names would you suggest?

PR #7961, opened 2024-02-03
Work In Progress - Toc v5.3.x rewrite

pmario:
This PR was created to create a NETLIFY Preview

**Edit**: Test build now works: https://deploy-preview-7961--tiddlywiki-previews.netlify.app/

### All TOC-macros Rewritten

The toc-macro now use **toc-include**, **toc-filter** and **toc-hide** fields as [discussed at Talk: ](https://talk.tiddlywiki.org/t/toc-macros-rewritten-a-lot-of-new-fuctionality-part-2/9071)

[Initial discussion at Talk (Part1)](https://talk.tiddlywiki.org/t/proposal-toc-macros-rewrite-using-if-functions-and-procedures-new-fuctionality/9046/6?u=pmario)

### CTRL-click Chevron Expands / Folds Tree Branches

- CTRL-clicking on a chevron should expand the whole tree
- CTRL-click again will close the whole tree again

**Advantage:** If everything is expanded, a "normal" click only fold the "current" branch

### New Node Fields

New tiddler fields allow the new macros to have fine-grained control over toc branches on a "per node" / "per branch" level. 

- `toc-filter`
   - Expects a **titlelist** or a **filter**
   - If a toc node contains a field named: `toc-filter`, that filter will **replace the standard filter** that normally creates the "child list" of that node. 
   - **nodes added** with toc-filter have a CSS **class="toc-item-filter"**
   - If you want to list nodes with the current tag you should add `[tag<currentTidder>!has[draft.of]]` if you do not want to list draft-tiddlers
   - **If there is no toc-filter**, the default filter is: `[all[shadows+tiddlers]tag<tag>!has[draft.of]$(sort)$]` where the last element is a text-substitution of the _sort_ parameter of the toc macro. 

- `toc-include`
  - Expects a **titlelist** of a **filter**.
  - Tiddlers **will be added** to the existing accumulated list. 
  - node added with toc-include have a CSS **class="toc-item-include"**

- `toc-hide`
  - If set to **yes**, the node will be removed from the child list
  - toc-hide **takes precedence** over all other fields. 
  - Expects a **titlelist** or a **filter**
  - If set to a filter-run  eg: `:and[prefix[todo]]` ... all accumulated tiddlers prefixed `todo` will be **removed**
  - Be aware: Without the `:and` filter-run-prefix, the filter `[prefix[todo]]` is equivalent to `[all[tiddlers]prefix[todo]]`
   - **parents** with a toc-hide field have a CSS **class="toc-item-hide"**
   - toc-hide nodes are internally evaluated in the same way as the _exclude_ parameter of the toc macro

- `toc-itemTemplate`
   - Expects a template tiddler-title
   - If itemTemplate is set, it uses the content of the template tiddler to display a toc node
      - If itemTemplate is used captionField (see info below) is set to `text`

All **titlelists** of **filters** added to the new fields are internally evaluated using `[subfilter[tf.toc-*]]`

The field names used are defined with global variables. So for expert users, they can be overwritten like this. 

```
\procedure toc-hideField()    toc-hide
\procedure toc-includeField() toc-include
\procedure toc-filterField()  toc-filter

<<toc tag:"Contents">>
</div>
```

------

### New Macro Parameters

- **captionField** (default: caption)
  - Expects a field-name. If set it uses this field instead of the caption-field
  - If the field is empty, or does not exists, it tries to use "caption" or falls back to the "tiddler-title"

- **tagField**
  - Uses any "tag-like" field instead of the "tags" field to create the "tag-like" tree
  - tagField takes precedence over parentField

- **parentField**
  - Instead of using tags, the toc tree is built using "parent" fields. 
  - There is no default, but usually it will be "parent"
  - This function uses the same filter behaviour as the tocP plugin

- **itemTemplate**
  - Expects a: template tiddler-title
  - by default toc macros use the `toc-defaultItemTemplate` procedure, which is specific to toc-macros
  - If itemTemplate is set, it uses the content of the template tiddler to display a toc node
     - If itemTemplate is used captionField is set to `text`
  - **Important:** If the `toc-itemTemplate` field is set in the current node it will take precedence over the macro parameter

The new fields should work with every toc-* macro.

-------

<details><summary>Test Styles are part of the JSON test-suite linked below ;)</summary>

```
.toc-item-filter> a,
.toc-item-filter> .tc-toc-caption {
  color: darkviolet;
}

.toc-item-filter> button > svg {
  fill: darkviolet;
}

.toc-item-include  > a,
.toc-item-include > .tc-toc-caption
{
  color: green;
}

.toc-item-include > button > svg {
  fill: green;
}

.toc-item-hide> a,
.toc-item-hide > .tc-toc-caption
{
  color: red;
}

.toc-item-hide> button > svg {
  fill: red;
}

```

</details>

-----

Requested (but not implemented yet) features from Part-1:

- Use a `$:/config/toc/`_currentTidder_ configuration to allow "per node" parameter change, whithout "touching" a user-created toc
  - This will allow plugin authors, or admins to modify toc-behaviour without the need to edit user-content.
- (If I forgot something -- let me know, below)

-----

### Test Data

The JSON attachment contains some `test-*` tiddlers that I use for testing atm. They are **chaotic**, but contain a bit of info and some links

* blue links are standard
* green links are `toc-include`d
* dark violet is `toc-filter`ed

JSON file can be found at: https://talk.tiddlywiki.org/t/toc-macros-rewritten-a-lot-of-new-fuctionality-part-2/9071?u=pmario


pmario:
To look at Files Changed at GitHub. Use the diff  "split" setting to see what's going on

image

pmario:
The existing "Table-of-Contents Macros" tiddler got an overhaul, to make it more consistent and easier to understand

pmario:
@Jermolene -- Do you think these new macros will have a chance to be merged into the core?

pmario:
@Jermolene -- The OP at GH now contains the "docs" info copied over from the talk thread part 2

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
 title: $:/core/macros/toc
 tags: $:/tags/Macro
 
+<!--
+Naming rules / pattern
```
Jermolene, line 5: Hi @pmario my main concern at the moment is with the comments that make up the documentation. They are written in a style that is brief and abbreviated (for example, why use a slash in the line above?) and therefore unnecessarily hard to understand.

I am not a fan of inventing new symbols like `->`. The important thing for a project like TiddlyWiki is consistency, and so it is not helpful if one file full of macros adopts different standards for documentation.

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+- functions: tf.toc-variableName, "tf." -> function prefix reservered for TW, "toc-" -> global macro prefix, "variableName"
+- procedures eg: toc-procedureName,
+- macro parameters: path, itemTemplate
+- derived macro parameters: _path, _itemTemplate -> derived variables start with an underscore
 
-\define toc-open-icon() $:/core/images/down-arrow
-\define toc-closed-icon() $:/core/images/right-arrow
+The toc macro code gets more and more complex from top to bottom.
```
Jermolene, line 11: "callse" typo

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+toc-tabbed-internal-nav calls toc-tabbed-external-nav, which callse toc-selective-expandable, which calls toc-expandable
+
+The very first toc macro is _not_ "reused". So it can be used as a "simple" reverence implementation,
```
Jermolene, line 14: The use of double quotes is unusual and a bit clumsy. What are they trying to accomplish?

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+because it is realatively easy to debug.
```
Jermolene, line 15: "realatively" typo

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+\function tf.toc-getItemTemplate() [<itemTemplate>!is[blank]then<itemTemplate>] :else[<currentTiddler>get[toc-itemTemplate]]
+
+<!-- Use macro itemTemplate or one defined as a toc-itemTemplate field -->
+\procedure _itemTemplate()
+<$transclude $tiddler=<<tf.toc-getItemTemplate>> >
+	<<toc-defaultItemTemplate>>
+</$transclude>
+\end
+
+<!-- v5.2.4 add "Custom Icons" -->
```
Jermolene, line 42: What does this comment mean? Is it just marking something that was added in v5.2.4? Why is it needed?

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+
+<!-- helper functions for li class definitions -->
+\function tf.toc-itemClassFilter() [<selectedTiddler>get[text]match<currentTiddler>then[toc-item-selected]else[toc-item]]
+\function tf.toc-includeClass()    [subfilter<tf.toc-include>match<currentTiddler>then[toc-item-include]]
+\function tf.toc-filterClass()     [subfilter<tf.toc-filter>match<currentTiddler>then[toc-item-filter]]
+\function tf.toc-hideClass()       [<tf.toc-hide>!is[blank]then[toc-item-hide]]
+<!-- join the CSS definitions from above -->
+\function tf.toc-itemClass()       [<tf.toc-itemClassFilter>] [<tf.toc-includeClass>] [<tf.toc-filterClass>] [<tf.toc-hideClass>] +[join[ ]]
+
+<!-- Special case, if toc-hide field is "yes", we need a currentTiddler *titlelist* -- otherwise return the content as text -->
```
Jermolene, line 58: I don't understand what this sentence is trying to communicate but the use of commas here is wrong. Better to say:

```
Special case: if the toc-hide field is "yes", we need a currentTiddler *titlelist* -- otherwise return the content as text
```

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
-            <<toc-caption>>
-          </$button>
-        </$reveal>
-      </$list>
-      <$reveal type="match" stateTitle=<<toc-state>> text="open">
-        <$macrocall $name="toc-selective-expandable" tag=<<currentTiddler>> sort=<<__sort__>> itemClassFilter=<<__itemClassFilter__>> exclude=<<__exclude__>> path=<<__path__>>/>
-      </$reveal>
-    </li>
-  </$set>
+<!-- ==========================================================
```
Jermolene, line 243: This is another non-standard new type of marker. We should be consistent.

core/wiki/macros/toc.tid
```diff
@@ -1,218 +1,422 @@
...
+		selectedTiddler=<<selectedTiddler>>
+		unselectedText=<<unselectedText>>
+		missingText=<<missingText>>
+		template=<<template>>
+		exclude=<<exclude>>
+		captionField=<<captionField>>
+		tagField=<<tagField>>
+		parentField=<<parentField>>
+		itemTemplate=<<itemTemplate>>
+	/>
```
Jermolene, line 420: This file seems like it is now unmanageably large.

Jermolene (commented):
Thanks @pmario this is a huge and welcome piece of work. I have some concerns about the size of the file, and I think the documentation comments need a little more work. I wonder if you could collaborate with another contributor on this?

pmario:
>@jermolene wrote
>
>This file seems like it is now unmanageably large.

As you can see in the <$macrocall $name="toc-tabbed-... directly above your comment. It uses 12 lines now instead of 1. IMO those 12 lines make the code much more readable and manageable. 

IMO the same is true for all other $marcrocalls and all $list-widget filters. They mainly use 4 lines now instead of 1. 
Changing that back would make the file 42 lines "shorter" but much harder to read for other mainainers than me. 

The file now contains 50+ lines of inline comment, which IMO also makes maintenence easier, since they point out, where to pay attention. 

The original toc-macros tiddler had about 220 lines of code. The new macors have about 300 real lines of code, but they also have a lot more functionality. 

- The toc-macros can now handle tag-based, field-based and parent-based tocs using some new macro parameters
- It can handle: per item-based "hide", "include" and "filter" fields for every one of the above new macro "types"
- It can handle: open / close a "full branch" with CTRL-click on the "more chevron"
- It also allows to create open / close "full branch" buttons added to a custom item-template
- It can handle: global item templates to show other info than caption or the tiddler title
- It can handle: custom templates on a "per item" basis 
- With the right templates toc-macros can now handle a tree that was created by the streams-plugin from Saq
- With the right settings and templates, they can handle trees created by my tocP-plugin, which can handle "parent based" tocs and Saq's streams

All that new functionality could be achieved by about a 100 lines of new code and a lot of code-reuse with some new settings.
I think it would have not been possible to do this with the "old" syntax. Especially \functions work great now.

Most of the ideas discussed at Talk: https://talk.tiddlywiki.org/t/proposal-toc-macros-rewrite-using-if-functions-and-procedures-new-fuctionality/9046/34 could be implemented with 100 lines of code. 

I think maintaining the 300 lines of code is much easier than having 4 files with 200 lines each, where 90% of the code is the same. Everytime the core changes 1 of them all 4 of them would need to be changed. 

Just my thoughts

core/wiki/macros/toc.tid
pmario: Yes. I think we should introduce something like this when we update core macros. It makes it much easier to visually separate the main macros from helper functions.

core/wiki/macros/toc.tid
pmario: They give me a visual hint about the imortant words in the text, since comments cannot be formatted.

core/wiki/macros/toc.tid
pmario: will change it

core/wiki/macros/toc.tid
pmario: I did add this info to see, that I cannot change those names for backwards compatibility reasons. Otherwise I may have changed them by accident

I'll change it to: `<!-- Allow custom icons (TW v5.2.4). Keep names for backwards compatibility -->`

core/wiki/macros/toc.tid
pmario: Will change it.

More info: toc-hide is special. It can be set to "yes" for ease of use, or it can be a filter. Rudimentary docs for the "toc-hide" field is at: https://talk.tiddlywiki.org/t/toc-macros-rewritten-a-lot-of-new-fuctionality-part-2/9071

core/wiki/macros/toc.tid
pmario: I did reply to this one, in the main thread

pmario:
@Jermolene .. The OP of this thread: https://github.com/Jermolene/TiddlyWiki5/pull/7961#issue-2116618606 also contains the preliminary docs of the new macro params and item fields. 

The current docs and the examples can be [tested with the latest Vercel preview](https://tiddlywiki5-git-fork-pmario-toc-v53x-rewrite-jermolene.vercel.app/#Table-of-Contents%20Macros:%5B%5BTable-of-Contents%20Macros%5D%5D%20%5B%5BTable-of-Contents%20Macros%20(Examples)%5D%5D)

The new paramters and item-fields have not been documented with TW yet.

pmario:
The test-date I use atm is this one. [toc-macros-new-test-tiddlers.zip](https://github.com/Jermolene/TiddlyWiki5/files/14229900/toc-macros-new-test-tiddlers.zip)

pmario:
>Thanks @pmario this is a huge and welcome piece of work. 

Thanks

>I have some concerns about the size of the file,

As I wrote. About 100 lines come from more readable macro-calls and inline docs. About 100 lines are new functionality. 

>and I think the documentation comments need a little more work. 

That's always true. ATM the inline docs makes sense for me, but may cause questions for others.

> I wonder if you could collaborate with another contributor on this?

Collaborate on the inline docs or the macros?

Jermolene:
Hi @pmario 

> Collaborate on the inline docs or the macros?

I meant the docs. We are exceptionally lucky to have so many willing contributors. The linguistic diversity of the community is a great strength, too, and helps us avoid the cardinal mistake of unknowingly creating English language-centric software.

There is however a higher barrier for documentation from contributors for whom English is not their first language. It generally falls to me to help contributors get the docs into shape, which is of course our usual bottleneck. Meanwhile, there are no shortage of members of the community who have the necessary English skills and so might be in a position to help.

pmario:
>Roma a PM at Talk
>
>I noticed that the menubar’s Contents dropdown disappears when I interact with any of the expand/collapse chevron buttons. The click clearly succeeds (when I re-invoke the dropdown, the submenus show/hide in modified way, as expected), but the current version fails to leave the drop-down open for further interaction after the chevron is clicked.
>
>No need to reply, unless you don’t see what I see and there’s more detail I can offer to help troubleshoot.

pmario:
Investigate this one:

>Mentioned at Talk: https://talk.tiddlywiki.org/t/optimal-toc-use-for-multiple-topics/9804/10?u=pmario
>
>I have noticed that the captionField does not seem to use the caption field in the absense of the new field set, I am unsure if this is just because I copy/pasted the $:/core/macros/toc tiddler to a blank tw or not, but I felt I should mention it here.

pmario:
>https://talk.tiddlywiki.org/t/optimal-toc-use-for-multiple-topics/9804/14?u=pmario
>
>... would it be difficult to adda tooltipField for example tooltipField=description would display as the tooltip the content of the description field on hover?, if using the caption they could provide the title as the tooltip field, or tooltiop etc…
>
> - The link widget (now) and Button widget (for a while) have the tooltip parameter, this can be a way to introduce other information, especlialy if its transcluded.

pmario:
Test Version 2026.04.02 

[toc-rewritten-full-docs.zip](https://github.com/user-attachments/files/26446586/toc-rewritten-full-docs.zip)

pmario:
@Jermolene ... I think I did find the problem, which causes test-build **fails** because of static rendering cause memory problems.

See: https://github.com/TiddlyWiki/TiddlyWiki5/pull/7961/commits/513ffea96c4617cb44bddc7b62ba46d457b1810e

I did create these changes and now the build process works again.

NETLIFY: Test build: https://deploy-preview-7961--tiddlywiki-previews.netlify.app/

Issue #7458, opened 2023-05-16
[BUG] The First Attribute of a Parameters Pragma is Overwritten when Is used Inside a Procedure


kookma:
### Describe the bug

If you use `\parameters` inside a `procedure` the value of first attribute of parameters will be overwritten by procedure first attribute

### To Reproduce

1. Go to ['...'](https://tiddlywiki.com/prerelease)
2. Create a new tiddler with below contents

```
\procedure eqn(x)
\parameters (a:2, b:5)
* x = <<x>>
* a = <<a>>
* b = <<b>>
\end

<<eqn 4>>
```

you will get

- x = 4
- a = 4
- b = 5

Note how `a` is overwritten by `x`

### TiddlyWiki Configuration

- Version 5.3.0

- OS: Win10
- Browser Chrome/Edge 112

kookma:
Second example

```
\procedure eqn()
\parameters (a:2, b:5)
* a = <<a>>
* b = <<b>>
\end

<<eqn 344>>
```

produces

- a = 344
- b = 5

kookma:
Third example

```
\procedure mymacro()
<$parameters name={{$:/SiteTitle}} age="21">
My name is <<name>> and my age is <<age>>.
</$parameters>
\end

<<mymacro GitHub 15>>
```

produces:  My name is [GitHub](https://tiddlywiki.com/prerelease/#GitHub) and my age is 15.

kookma:
Fourth example: here a **macro** is used

```
\define mymacro()
<$parameters name={{$:/SiteTitle}} age="21">
My name is <<name>> and my age is <<age>>.
</$parameters>
\end
```

produces: My name is GitHub and my age is 15.
<<mymacro GitHub 15>>
```

**Comment by pmario** [comment]
That's a tricky one. TW has "positional" and "named" parameters. So it's an implementation detail.

 .. @Jermolene may be it would be an option to use a "rest-like" syntax: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/rest_parameters 

```
\procedure eqn(x, y)
\parameters (... a:2, b:5)

```

pmario:
>It seems there is a bug with both \parameters pragma and <$parameters widget! whne are used inside procedures

It's more like a "misuse". It's not intended to be used that way. ... **But** for sure it's an issue. ... As I wrote IMO it's how "positional" parameters are resolved at the moment. 

[...]

Issue #7909, opened 2023-12-25
[BUG] Procedure defaults to \whitespace trim when imported


flibbles:
### Describe the bug

When you import procedures, they have whitespace trim enabled by default.

### Expected behavior

Expected: This is *fancy* content
Results: This is*fancy*content

Is this expected behavior?

### To Reproduce

```
title: Target

\procedure test()
This is //fancy// content.
\end
```

```
title: Test

\import Target
<<test>>
```

### TiddlyWiki Configuration

- TiddlyWiki v5.3.3

CodaCodr:
https://talk.tiddlywiki.org/t/announcing-the-release-of-tiddlywiki-v5-3-2/8661/14

@Jermolene  Is this the simple STR we need finally?

Jermolene:
Thanks @flibbles 

@CodaCodr – what is STR?

pmario:
If `\define` is used instead of `\procedure` it works as expected. I think "define" and "procedure" should create the same behaviour if possible

Jermolene:
> I think "define" and "procedure" should create the same behaviour if possible

There is one intentional difference here: procedures inherit the `\whitespace trim` setting from the tiddler in which they are defined. I suspect that it is the implementation of that behaviour that is causing the problem here.

CodaCodr:
STR - steps to reproduce.

Jermolene:
Hi @flibbles I do think this is a bug, but it fell off my (rather fallible) radar and I haven't had a chance to investigate. I'm pretty sure it will be an issue with the mechanism for inheriting whitespace trim settings from the containing tiddler into its procedure definitions.

pmario:
I'm changing the wikitext syntax for the docs-macros at the moment.

This issue prevents me from changing `\define` to `\procedure` for these examples: https://tiddlywiki.com/#Macro%20Calls%20in%20WikiText%20(Examples)

Jermolene:
It turns out that the fix for this is straightforward: the code for the import pragma was passing true for "configTrimWhiteSpace" when parsing the tiddlers, which meant that "whitespace trim" was being applied to all parsed tiddlers. That matters with procedure definitions because they inherit the whitespace trim setting from the parent parser.

PR #7962, opened 2024-02-04
$parameters widgets now claim index parameters sequentially

flibbles:
This PR is in reference to the post-merge conversation in #7458. This changes it so $parameters widgets rely on their parent $transclude widgets to decide what index to use, rather than using their own. This allows multiple $parameters widgets to exist and to claim unnamed parameters in sequence, rather than in overlap.

As per that discussion, it was decided that the `$depth` attribute of $parameters should be removed. I still think that, but *technically* I just implemented this in a way that removing it isn't required.

Since $transclude widgets would keep track of which of their own unnamed parameters have claimed, that means multiple children $parameters can claim them, and so can grandchildren $parameters, and so on.

So does it work, and do all tests pass? Yes.
Does making code that makes use of this a heinous perversion? Yes.

That's why I'm still in favor of removing depth. And if no one says anything, I'll cut it out when I make some extra tests for this in a few days, but I'm publishing this now in case anyone wants to discuss.


Jermolene:
Thank you @flibbles. It would be great to have some tests, and perhaps clarifications to the documentation.

flibbles:
Sure thing. This builds off of your branch which already had that one basic test, but there are a few other ways I can imagine this going awry that I'd like to see. I'll do up the documentation.

...and that's a yes to scrapping $depth?

Jermolene:
Thanks @flibbles, and yes, apologies I meant to say that I do think we should go ahead and scrap the `$depth` attribute. I could probably persuade myself that it might be worth just disabling the current implementation in the first instance, but I'm happy to leave that to your judgement.

flibbles:
All right.

* I elaborated the test you wrote to consider more cases. I think it should be good.
* I updated and added to the documentation in the places I could find that are relevant. (Let me know if my example is inappropriate.)
* I excised the $depth attribute and its associated test. This is on an isolated commit and would be easy to revert if need be.

This should be good to merge, I think. Let me know if you have any concerns.

pmario:
Thanks for your effort. I did some rudimentary tests and it seems to work as advertised.

Jermolene:
Hi @flibbles I think that there is a subtle problem with refreshing. I've added a failing test to illustrate the problem: the refresh mechanism means that the child parameters widgets may call `getTransclusionParameter` additional times, resulting in the wrong value for the `claimedIndices` property.

flibbles:
You're right. I hadn't considered that. This modified version of getTransclusionParameter remembers which parameters go to which index.

I *think* that solves it. Their are only a few cases where this might fail that I can think of:

* a child `$parameter` widget changes the name of one of its parameters, which I think could only happen if you're using a genesis widget.
* Or if a refresh causes *different* `$parameter` widgets to get encountered that hadn't been before. Such as a `<% if %>` conditional branching differently.

I might be able to make it more robust than that, but there may be some strange corner cases that get very hard to handle. I'll think about this a little more.

flibbles:
Okay. This new system should work for all cases. It's what I was doing the first time, except `$parameter` widgets will remember at what index they started at. If that would be different when it refreshes, then it knows it must rebuild.

It also remembers what index it ends at, because it effectively dry-runs index assignment during refresh, even if it doesn't change. This helps later `$parameter` widgets determine their own indices.

I think this is solid.
* Genesis sorcery doesn't trip this up
* Parameters sneaking in or out because of conditional logic during refresh should be accounted for
* Sequential `$parameter` that aren't nested should be fine too. They won't mess each other up.

I think this is good to go. Let me know if my code comments aren't clear, or if that getter/setter isn't the TiddlyWiki way.

Otherwise, good to merge?

editions/test/tiddlers/tests/data/transclude/Parameterised-Conditional-Parameters-Refreshed.tid
```diff
@@ -0,0 +1,39 @@
+title: Transclude/Parameterised/ConditionalParameters/Refreshed
+description: Parameterised transclusion when conditional parameters are changed by a refresh
+type: text/vnd.tiddlywiki-multiple
+tags: [[$:/tags/wiki-test-spec]]
+
+title: Output
+
+\procedure elephant(filler)
+\whitespace trim
```
pmario, line 9: procedures now inherit `\whitespace trim` from the tiddler scope. So only 1 of them is needed

flibbles: I took that whitespace trim out, and the test started failing. The expect() error messages shows that a ton of whitespace is now present in the output.

editions/test/tiddlers/tests/data/transclude/Parameterised-Conditional-Parameters-Refreshed.tid
```diff
@@ -0,0 +1,39 @@
...
+<$parameters A=defaultA>
+A/<$text text=<<filler>>/>/<$text text=<<A>>/>/<$text text=<<B>>/>
+</$parameters>
+<% else %>
+<$parameters B=defaultB>
+B/<$text text=<<filler>>/>/<$text text=<<A>>/>/<$text text=<<B>>/>
+</$parameters>
+<% endif %>
+\end elephant
+\whitespace trim
```
pmario, line 21: May be this one needs to be at the top of the tiddler

Jermolene: Yes, there has to be a whitespace trim before the procedure definitions in order for it to be inherited by the procedures.

flibbles:
Oh, I see. I had no idea how the whitespace pragma worked with procedures. For me, they've always behaved erratically (#7909), so I've just been very liberal with them.

Either way, I've removed the extra pragma and set this up the way you guys suggested.

flibbles:
@Jermolene, nudging this because I'll be releasing the next Relink version soon, and I'd like to know what behavior Relink should assume.

The whitespace pragma issue you guys raised is resolved. I think this should be good to merge. If not, let me know what you'd like improved. I'm happy to do so.

Jermolene:
Thanks @flibbles

flibbles:
...I just noticed that this PR was merged, but not into master.

@Jermolene, was this patch decided against? I was surprised to find the $parameters $depth attribute still exists as I was working. I still cannot fathom a use for it.

flibbles:
(Missed this one thanks to my spam filter too...)

I haven't touched this since, and no reason to touch it now. I have no reason to think it isn't good as is.

PR #7967, opened 2024-02-06
Added conditional to the allowed markdown wikitext

flibbles:
[Some people would like for the new conditionals to be available in markdown.](https://talk.tiddlywiki.org/t/how-to-use-nested-list-filter/9064/13)

This adds it to the renderWikiTextPragma file.


flibbles:
Nuts. This still doesn't work. Ignore this.

PR #7975, opened 2024-02-09
fix incorrect state reference in core/ui/AdvancedSearch/Standard

### LINKED ISSUE #7973: [BUG] Incorrect State Reference in AdvancedSearch Standard
### Describe the bug

`$:/core/ui/AdvancedSearch/Standard` sets `configTiddler` by referencing the state stored in `$:/state/search/currentTab` instead of `$:/state/advancedsearch/standard/currentTab`. 

Since the former is the state used in the sidebar by `$:/core/ui/SideBarSegments/search` this leads to weird behavior if you add additional search tabs which do not output the same set of results (e.g. have different filters). What actually happens is as follows:

* Changing the tab in advanced search *does not* update the results in advanced search.
* Changing the tab in the sidebar search *does* update the results in *both* the sidebar *and* advanced search

Offending line is [here](https://github.com/Jermolene/TiddlyWiki5/blob/0cf80e824f4dd8d557a6cb586dd3dc9ddcbc46dd/core/ui/AdvancedSearch/Standard.tid#L57)

### Expected behavior

* Search filters should change in advanced search when selecting a different search tab in the advanced search,.
* Search filters should not change in advanced search when seleting a tab in the sidebar search.

### To Reproduce

1. Clone `$:/core/ui/DefaultSearchResultList` to create a new search tab and alter the filter fields (e.g. `[prefix[T]!is[system]search:title<userInput>sort[title]limit[250]]`)
2. Enter a distingushing term in the sidebar (e.g. `Tiddler`) and observe that swapping tabs alters the results
3. Open the Advanced Search and enter the same term into the Standard search
4. Observe that swapping tabs *does not* alter the results
5. Enter any term in the sidebar search and observe that swapping tabs there *does* alter results in the Advanced Search

### Screenshots

_No response_

### TiddlyWiki Configuration

Bug is present in the latest version (5.3.3) and can be reproduced in official wiki.

### Additional context

_No response_

eschlon:
This PR fixes #7973

### DIFF
--- a/core/ui/AdvancedSearch/Standard.tid
+++ b/core/ui/AdvancedSearch/Standard.tid
@@ -54,7 +54,7 @@ caption: {{$:/language/Search/Standard/Caption}}
 	variable="listItem">
 <$vars
 	userInput={{{ [[$:/temp/advancedsearch]get[text]] }}}
-	configTiddler={{{ [[$:/state/search/currentTab]!is[missing]get[text]] ~[{$:/config/SearchResults/Default}] }}}
+	configTiddler={{{ [[$:/state/advancedsearch/standard/currentTab]!is[missing]get[text]] ~[{$:/config/SearchResults/Default}] }}}
 	searchListState="$:/temp/advancedsearch/selected-item">
 <$list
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @eschlon much appreciated.

PR #7883, opened 2023-12-11
Rewrite tag-picker macros as procedures

etardiff:
This PR modernizes the tag-picker macros as procedures, as @Jermolene suggested in [#7548 (comment)](https://github.com/Jermolene/TiddlyWiki5/pull/7548#issuecomment-1825458523).

What I changed:

* rewrote macros as procedures/replaced text substitutions with variables
* replaced nested $set and $var widgets with $let widgets/functions
* minor rewrites to reduce redundant code
* some additional linebreaks for clarity

*This is functionally identical to my initial PR, [#7880](https://github.com/Jermolene/TiddlyWiki5/pull/7880)*, minus the extra file I'd accidentally added. Thank you for bearing with me as I learn to navigate GitHub.

PR #7978, opened 2024-02-09
Tag picker rewritten using new v5.3.x syntax

pmario:
This PR closes  **[BUG] calling tag-picker macro with a tagFields=foo shows the tags field instead of foo #7461** 

@etardiff This PR replaces your: **Rewrite tag-picker macros as procedures #7883**

I did use your code as basis and did extend it with my code from:  **Fix and extend tag picker functions #7548**, since I need it to have a nice UI for tag-like fields. 

I did add a new Example tiddler, which was completely missing.


pmario:
Slightly related to: **Work In Progress - Toc v5.3.x rewrite #7961**  since the new tag-picker can now be used to easily "populate" tag-like fields

pmario:
Converted this PR into a draft. There seems to be a problem with the "focus selector" if used in combination with the `$:/core/ui/EditTemplate/tags` template.

Using keyboard handling to add tag-likes -> works
Using the mouse to click -> focus is completely lost. 

```
\procedure tagField() test

{{||$:/core/ui/EditTemplate/tags}}
```

image

pmario:
@BurningTreeC ... Are you able to help with the keyboard focus handling mentioned at: https://github.com/Jermolene/TiddlyWiki5/pull/7978#issuecomment-1937825065

BurningTreeC:
Hello @pmario ... I will have to take a look at what you are doing here. When I understand it, I'll happily help

BurningTreeC:
Hello @pmario - could you please describe the problem more precisely?

BurningTreeC:
Hi @pmario ... I found the problem.
In the procedure `get-tagpicker-focus-selector()` your using the reference to `$(thisTiddler)$` which somehow doesn't work.
First of all, I would like that `thisTiddler` gets CSSescaped first by the `escapecss[]` operator , so that it works correctly with all tiddler titles.
Then we need to find a way for the variable to be resolved within the procedure

BurningTreeC:
@pmario ... if you make the `get-tagpicker-focus-selector()` a macro then it works. I was away from tiddlywiki quite some time and don't know anything about procedures and functions yet. So I don't know if `$(variable)$` gets resolved within a procedure but it seems it doesn't.
<hr>
Another thing: I've seen that you added an id so that the dropdown works for different tag pickers within a single tiddler.
We should somehow reuse that id and add it to the `tc-edit-add-tag` class. So that in the `get-tagpicker-focus-selector` we can identify the correct tag picker

BurningTreeC:
Here's a working tag-picker with my small additions
[tag-picker.txt](https://github.com/Jermolene/TiddlyWiki5/files/14277958/tag-picker.txt)

pmario:
Here's a working tag-picker with my small additions
[tag-picker.txt](https://github.com/Jermolene/TiddlyWiki5/files/14277958/tag-picker.txt)

Thx, I'll have a closer look soon.

---------

>First of all, I would like that thisTiddler gets CSSescaped first by the escapecss[] operator , so that it works correctly with all tiddler titles.

Assigning a title to `data-tiddler-something` will work with all unicode characters. So no CSS escapes needed.

BurningTreeC:
> > Here's a working tag-picker with my small additions
> > [tag-picker.txt](https://github.com/Jermolene/TiddlyWiki5/files/14277958/tag-picker.txt)
> 
> Thx, I'll have a closer look soon.
> 
> > First of all, I would like that thisTiddler gets CSSescaped first by the escapecss[] operator , so that it works correctly with all tiddler titles.
> 
> Assigning a title to `data-tiddler-something` will work with all unicode characters. So no CSS escapes needed.

I made a different experience, but if you say so I'll have to take a closer look

pmario:
With the help of your actions-debug and your new code I did find a solution without any CSS escapes. We only have to take care, that the `tag-picker-dropdown-id` is really unique and added to the right DOM element. 

So the `get-tagpicker-focus-selector` can be really simple. ... Some more experiments needed, but I think I do understand the mechanism now. 

Thanks for your input \o/

pmario:
@Jermolene -- This one is ready for review now. 

All the tap-picker inputs should work with keyboard text inputs, keyboard down/up selection and mouse inputs now. 

If testing the "Examples" it's important to read the info texts, since example 3 is used to "populate" the field used for examples 4 and 5.

Jermolene:
Hi @pmario I've merged #7883 so this will need some adjustments, many thanks.

pmario:
I did update this PR. 

[Updated tag-picker docs and new "Examples" tiddler](https://tiddlywiki5-git-fork-pmario-tag-picker-new-syntax-jermolene.vercel.app/#tag-picker%20Macro:%5B%5Btag-picker%20Macro%5D%5D%20%5B%5Btag-picker%20Macro%20(Examples)%5D%5D)

Jermolene:
Thanks @pmario

PR #7981, opened 2024-02-11
$:/core/ui/EditTemplate/tags - rewrite to use v5.3.x syntax

pmario:
@Jermolene, @saqimtiaz  -- This PR rewrites $:/core/ui/EditTemplate/tags  and uses the new v5.3.x syntax. 

**There is 1 problem**:  I could not find a way to get rid of the `<$wikify` widget, to get the result of the `<<contrastcolour>>` macro. 

```
<$wikify name="foregroundColor" text="""<$macrocall $name="contrastcolour" target=<<colour>> fallbackTarget=<<fallbackTarget>> colourA=<<colourA>> colourB=<<colourB>> />""">
```

Help wanted.

### DIFF
--- a/core/ui/EditTemplate/tags.tid
+++ b/core/ui/EditTemplate/tags.tid
@@ -3,39 +3,50 @@ tags: $:/tags/EditTemplate
 
 \whitespace trim
 
-\define lingo-base() $:/language/EditTemplate/
+\procedure lingo-base() $:/language/EditTemplate/
 
-\define tag-styles()
-background-color:$(backgroundColor)$;
-fill:$(foregroundColor)$;
-color:$(foregroundColor)$;
+\function tag-styles() [[background-color:$(backgroundColor)$; fill:$(foregroundColor)$; color:$(foregroundColor)$;]substitute[]]
+
+\procedure tag-body-inner(colour,fallbackTarget,colourA,colourB,icon,tagField:"tags")
+<$wikify name="foregroundColor" text="""<$macrocall $name="contrastcolour" target=<<colour>> fallbackTarget=<<fallbackTarget>> colourA=<<colourA>> colourB=<<colourB>> />""">
+<$let backgroundColor=<<colour>> >
+	<span style=<<tag-styles>> class="tc-tag-label tc-tag-list-item tc-small-gap-right" data-tag-title=<<currentTiddler>>>
+		<$transclude tiddler=<<icon>>/>
+		<$view field="title" format="text"/>
+		<$button class="tc-btn-invisible tc-remove-tag-button" style=<<tag-styles>>>
+			<$action-listops $tiddler=<<saveTiddler>> $field=<<tagField>> $subfilter="-[{!!title}]"/>
+			{{$:/core/images/close-button}}
+		</$button>
+	</span>
+</$let>
+</$wikify>
 \end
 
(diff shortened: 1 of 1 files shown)


core/ui/EditTemplate/tags.tid
```diff
@@ -3,39 +3,50 @@ tags: $:/tags/EditTemplate
...
 
-\define lingo-base() $:/language/EditTemplate/
+\procedure lingo-base() $:/language/EditTemplate/
 
-\define tag-styles()
-background-color:$(backgroundColor)$;
-fill:$(foregroundColor)$;
-color:$(foregroundColor)$;
-\end
+\function tag-styles() [[background-color:$(backgroundColor)$; fill:$(foregroundColor)$; color:$(foregroundColor)$;]substitute[]]
```
Jermolene, line 8: This seems a little inelegant. Perhaps we could use the new attributes for `style.backgroundColor`, `style.fill`, `style.color`?

Jermolene (commented):
Thanks @pmario – the wikify widget is definitely needed here because of the way that colour palette values may themselves be further `<<colour>>` macros.

core/ui/EditTemplate/tags.tid
pmario: @Jermolene -- I have no clue what you mean here. Can you be more specific?

saqimtiaz: @pmario instead of creating and assign a single style attribute, assign the different style properties individually using the `style.property` syntax for attributes. For example, `style.fill=<<foregroundColor>>`

pmario: ATM the `tag-styles` function is assigned 2 times like `class=<<tag-styles>>`, so the proposed change will create duplicated code, which imo is harder to maintain than a function "string" even if it is a bit clunky.

I'll change it to the following code, which should be easier to read and to understand

```
\function tag-styles()
[[ color:$(foregroundColor)$;
   fill:$(foregroundColor)$;
   background-color:$(backgroundColor)$; ]substitute[]]
\end
```

Jermolene: Hi @pmario while readability is a consideration, this is also about robustness. Using string substitution means that things can break if unexpected characters are included in the substituted strings. This is a particular risk if the substituted string is provided by the user and not subject to validation.

pmario: hmm, The old code did use a `\define` instead of a `\function`. The concept is the same. The [tag-styles concept is 11 years old](https://github.com/Jermolene/TiddlyWiki5/blame/master/core/ui/EditTemplate/tags.tid#L8) and it seems users did not have any problems with it.

Jermolene: > it seems users did not have any problems with it.

I think that if we are going to do this sort of refactoring then our primary goal should be to demonstrate best practice. Many users will use core code as the basis for their own custom solutions, so we should be wary of inadvertently encouraging them to use techniques that we know not to be robust.

I do understand that there is an alternative perspective that the textual substitution approach is preferable for pedagogic purposes because it is so conceptually simple, but I think our collective experience with SQL (which has a simpler syntax than TW/HTML/CSS) suggests that this is a dangerous area.

pmario: I changed it as requested

Jermolene:
Thank you @pmario

PR #7985, opened 2024-02-14
remove overflow: auto from tiddler preview

BurningTreeC:
This removes `overflow: auto` from the tiddler preview. I'm not sure anymore why it was added.
It causes some problems in the preview with popups for example, see screenshots


Jermolene:
Thanks @BurningTreeC

PR #7986, opened 2024-02-14
Core-macro-diff-new-syntax

pmario:
- This PR changes the `$:/core/macros/diff` to the new v5.3.x syntax
- It also changes the tags to `$:/tags/Global`

Important for testing: This diff-macor is used with the `$:/Import` dialogue.


Jermolene:
Thanks @pmario

PR #7987, opened 2024-02-14
css-macro -- change tags to $:/tags/Global

pmario:
- This PR changes the $:/core/macros/CSS to use the new v5.3.x syntax
- sets the tags to $:/tags/Global

### DIFF
--- a/core/wiki/macros/CSS.tid
+++ b/core/wiki/macros/CSS.tid
@@ -1,76 +1,69 @@
 title: $:/core/macros/CSS
 tags: $:/tags/Macro
 
-\define colour(name)
-<$transclude tiddler={{$:/palette}} index="$name$"><$transclude tiddler="$:/palettes/Vanilla" index="$name$"><$transclude tiddler="$:/config/DefaultColourMappings/$name$"/></$transclude></$transclude>
+\procedure colour(name)
+\whitespace trim
+<$transclude $tiddler={{$:/palette}} $index=`$(name)$`>
+	<$transclude $tiddler="$:/palettes/Vanilla" $index=`$(name)$`>
+		<$transclude $tiddler=`$:/config/DefaultColourMappings/$(name)$`/>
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @pmario

Jermolene:
Hi @pmario I've discovered that there are problems with converting the CSS helpers from macros to functions. For example,  we previously had this macro:

```
\define box-shadow(shadow)
 ``
   -webkit-box-shadow: $shadow$;
      -moz-box-shadow: $shadow$;
           box-shadow: $shadow$;
 ``
 \end
```

This PR changes that macro to the following function:

```
\function box-shadow(shadow)
 [[ -webkit-box-shadow: $(shadow)$;
 	-moz-box-shadow: $(shadow)$;
 	box-shadow: $(shadow)$;]substitute[]]
 \end
```

The problem comes when the parameter that is passed requires wikifying. For example:

```
	<<box-shadow "1px 1px 3px <<colour red>>">>
```

The new function form does not apply wikification to the substituted value, but the previous implementation handled this OK.

I intend to revert these changes.

kookma:
> <<box-shadow "1px 1px 3px <<colour red>>">>

I consider this example hard to understand. I wish to not pass `"1px 1px 3px <<colour red>>"` as a macro argument. If so one can also expect `<<macro-name "a b <<another 'f g <<last>>'">>` etc.

Jermolene:
> I consider this example hard to understand. I wish to not pass `"1px 1px 3px <<colour red>>"` as a macro argument. If so one can also expect `<<macro-name "a b <<another 'f g <<last>>'">>` etc.

Macros like "box-shadow"  are actually no longer needed since the unprefixed properties are now supported on all browsers. But the point remains that we inadvertently broke backwards compatibility. There's no particular benefit to having made these macros be functions, so the obvious course is to revert back to a macro.

PR #7988, opened 2024-02-14
copy-to-clipboard - change tags to $:/tags/Global

pmario:
- This PR changes the macro: $:/core/macros/copy-to-clipboard
- tags: $:/tags/Global


Jermolene:
Thanks @pmario

PR #7990, opened 2024-02-15
Simplify permalinks

CrossEye:
This is a reworking of #7729, creating more user-friendly permalinks and permaviews.  There's not much to discuss here; you can see that closed PR for rationale and some interesting discussion.

The largest difference is that the previous version replaced spaces with underscore characters (`_`), whereas this one uses plus signs (`+`).  Although there was an attempt in the previous version to abstract that character into a single variable to allow a quick switch, there was still some need to alter the code because of the difference in how underscores and plus signs are converted in native url-encoding.  These were not huge changes, and the code is substantially the same.

This version includes a number of unit tests.  They are a little different from some of the others in the `$tw.utils` namespace, so I would appreciate it if someone can look them over to make sure they seem all right.  The difference is that I have a second level of suites to focus on each function by name, and then a number of specs under these.  The specs contain either a single assertion or a few very-closely related ones.  I kept with the majority trend in the unit tests by starting my descriptions with "should`", although I personally prefer to skip that.  Is there any other level of testing I should add?  These are used by `core/modules/startup/story`, and I didn't see other tests for that.  Should I be looking elsewhere?

### DIFF
--- a/core/modules/startup/story.js
+++ b/core/modules/startup/story.js
@@ -122,10 +122,10 @@ function openStartupTiddlers(options) {
 		var hash = $tw.locationHash.substr(1),
 			split = hash.indexOf(":");
 		if(split === -1) {
-			target = $tw.utils.decodeURIComponentSafe(hash.trim());
+			target = $tw.utils.decodeTWURITarget(hash.trim());
 		} else {
-			target = $tw.utils.decodeURIComponentSafe(hash.substr(0,split).trim());
-			storyFilter = $tw.utils.decodeURIComponentSafe(hash.substr(split + 1).trim());
+			target = $tw.utils.decodeTWURITarget(hash.substr(0,split).trim());
+			storyFilter = $tw.utils.decodeTWURITarget(hash.substr(split + 1).trim());
 		}
 	}
 	// If the story wasn't specified use the current tiddlers or a blank story
@@ -183,7 +183,234 @@ function updateLocationHash(options) {
 	var storyList = $tw.wiki.getTiddlerList(DEFAULT_STORY_TITLE),
 		historyList = $tw.wiki.getTiddlerData(DEFAULT_HISTORY_TITLE,[]),
 		targetTiddler = "";
-	if(options.targetTiddler) {
+	if(options.targetTiddler) {/*\
+	title: $:/core/modules/startup/story.js
+	type: application/javascript
+	module-type: startup
+	
+	Load core modules
+	
+	\*/
+	(function(){
+	
+	/*jslint node: true, browser: true */
+	/*global $tw: false */
+	"use strict";
+	
+	// Export name and synchronous status
+	exports.name = "story";
+	exports.after = ["startup"];
+	exports.synchronous = true;
+	
+	// Default story and history lists
+	var DEFAULT_STORY_TITLE = "$:/StoryList";
+	var DEFAULT_HISTORY_TITLE = "$:/HistoryList";
+	
+	// Default tiddlers
+	var DEFAULT_TIDDLERS_TITLE = "$:/DefaultTiddlers";
+	
+	// Config
+	var CONFIG_UPDATE_ADDRESS_BAR = "$:/config/Navigation/UpdateAddressBar"; // Can be "no", "permalink", "permaview"
+	var CONFIG_UPDATE_HISTORY = "$:/config/Navigation/UpdateHistory"; // Can be "yes" or "no"
+	var CONFIG_PERMALINKVIEW_COPY_TO_CLIPBOARD = "$:/config/Navigation/Permalinkview/CopyToClipboard"; // Can be "yes" (default) or "no"
+	var CONFIG_PERMALINKVIEW_UPDATE_ADDRESS_BAR = "$:/config/Navigation/Permalinkview/UpdateAddressBar"; // Can be "yes" (default) or "no"
+	
+	
+	// Links to help, if there is no param
+	var HELP_OPEN_EXTERNAL_WINDOW = "http://tiddlywiki.com/#WidgetMessage%3A%20tm-open-external-window";
+	
+	exports.startup = function() {
+		// Open startup tiddlers
+		openStartupTiddlers({
+			disableHistory: $tw.boot.disableStartupNavigation
+		});
+		if($tw.browser) {
+			// Set up location hash update
+			$tw.wiki.addEventListener("change",function(changes) {
+				if($tw.utils.hop(changes,DEFAULT_STORY_TITLE) || $tw.utils.hop(changes,DEFAULT_HISTORY_TITLE)) {
+					updateLocationHash({
+						updateAddressBar: $tw.wiki.getTiddlerText(CONFIG_UPDATE_ADDRESS_BAR,"permaview").trim(),
+						updateHistory: $tw.wiki.getTiddlerText(CONFIG_UPDATE_HISTORY,"no").trim()
+					});
+				}
+			});
+			// Listen for changes to the browser location hash
+			window.addEventListener("hashchange",function() {
+				var hash = $tw.utils.getLocationHash();
+				if(hash !== $tw.locationHash) {
+					$tw.locationHash = hash;
+					if(hash !== "#") {
+						openStartupTiddlers({defaultToCurrentStory: true});
+					}
+				}
+			},false);
+			// Listen for the tm-browser-refresh message
+			$tw.rootWidget.addEventListener("tm-browser-refresh",function(event) {
+				window.location.reload(true);
+			});
+			// Listen for tm-open-external-window message
+			$tw.rootWidget.addEventListener("tm-open-external-window",function(event) {
+				var paramObject = event.paramObject || {},
+					strUrl = event.param || HELP_OPEN_EXTERNAL_WINDOW,
+					strWindowName = paramObject.windowName,
+					strWindowFeatures = paramObject.windowFeatures;
+				window.open(strUrl, strWindowName, strWindowFeatures);
+			});
+			// Listen for the tm-print message
+			$tw.rootWidget.addEventListener("tm-print",function(event) {
+				(event.event.view || window).print();
+			});
+			// Listen for the tm-home message
+			$tw.rootWidget.addEventListener("tm-home",function(event) {
+				window.location.hash = "";
+				var storyFilter = $tw.wiki.getTiddlerText(DEFAULT_TIDDLERS_TITLE),
+					storyList = $tw.wiki.filterTiddlers(storyFilter);
+				//invoke any hooks that might change the default story list
+				storyList = $tw.hooks.invokeHook("th-opening-default-tiddlers-list",storyList);
+				$tw.wiki.addTiddler({title: DEFAULT_STORY_TITLE, text: "", list: storyList},$tw.wiki.getModificationFields());
+				if(storyList[0]) {
+					$tw.wiki.addToHistory(storyList[0]);
+				}
+			});
+			// Listen for the tm-permalink message
+			$tw.rootWidget.addEventListener("tm-permalink",function(event) {
+				updateLocationHash({
+					updateAddressBar: $tw.wiki.getTiddlerText(CONFIG_PERMALINKVIEW_UPDATE_ADDRESS_BAR,"yes").trim() === "yes" ? "permalink" : "none",
+					updateHistory: $tw.wiki.getTiddlerText(CONFIG_UPDATE_HISTORY,"no").trim(),
+					targetTiddler: event.param || event.tiddlerTitle,
+					copyToClipboard: $tw.wiki.getTiddlerText(CONFIG_PERMALINKVIEW_COPY_TO_CLIPBOARD,"yes").trim() === "yes" ? "permalink" : "none"
+				});
+			});
+			// Listen for the tm-permaview message
+			$tw.rootWidget.addEventListener("tm-permaview",function(event) {
+				updateLocationHash({
+					updateAddressBar: $tw.wiki.getTiddlerText(CONFIG_PERMALINKVIEW_UPDATE_ADDRESS_BAR,"yes").trim() === "yes" ? "permaview" : "none",
+					updateHistory: $tw.wiki.getTiddlerText(CONFIG_UPDATE_HISTORY,"no").trim(),
+					targetTiddler: event.param || event.tiddlerTitle,
+					copyToClipboard: $tw.wiki.getTiddlerText(CONFIG_PERMALINKVIEW_COPY_TO_CLIPBOARD,"yes").trim() === "yes" ? "permaview" : "none"
+				});
+			});
+		}
+	};
+	
+	/*
+	Process the location hash to open the specified tiddlers. Options:
+	disableHistory: if true $:/History is NOT updated
+	defaultToCurrentStory: If true, the current story is retained as the default, instead of opening the default tiddlers
+	*/
+	function openStartupTiddlers(options) {
+		options = options || {};
+		// Work out the target tiddler and the story filter. "null" means "unspecified"
+		var target = null,
+			storyFilter = null;
+		if($tw.locationHash.length > 1) {
+			var hash = $tw.locationHash.substr(1),
+				split = hash.indexOf(":");
+			if(split === -1) {
+				target = $tw.utils.decodeTWURITarget(hash.trim());
+			} else {
+				target = $tw.utils.decodeTWURITarget(hash.substr(0,split).trim());
+				storyFilter = $tw.utils.decodeTWURIList(hash.substr(split + 1).trim());
+			}
+		}
+		// If the story wasn't specified use the current tiddlers or a blank story
+		if(storyFilter === null) {
+			if(options.defaultToCurrentStory) {
+				var currStoryList = $tw.wiki.getTiddlerList(DEFAULT_STORY_TITLE);
+				storyFilter = $tw.utils.stringifyList(currStoryList);
+			} else {
+				if(target && target !== "") {
+					storyFilter = "";
+				} else {
+					storyFilter = $tw.wiki.getTiddlerText(DEFAULT_TIDDLERS_TITLE);
+				}
+			}
+		}
+		// Process the story filter to get the story list
+		var storyList = $tw.wiki.filterTiddlers(storyFilter);
+		// Invoke any hooks that want to change the default story list
+		storyList = $tw.hooks.invokeHook("th-opening-default-tiddlers-list",storyList);
+		// If the target tiddler isn't included then splice it in at the top
+		if(target && storyList.indexOf(target) === -1) {
+			storyList.unshift(target);
+		}
+		// Save the story list
+		$tw.wiki.addTiddler({title: DEFAULT_STORY_TITLE, text: "", list: storyList},$tw.wiki.getModificationFields());
+		// Update history
+		var story = new $tw.Story({
+			wiki: $tw.wiki,
+			storyTitle: DEFAULT_STORY_TITLE,
+			historyTitle: DEFAULT_HISTORY_TITLE
+		});
+		if(!options.disableHistory) {
+			// If a target tiddler was specified add it to the history stack
+			if(target && target !== "") {
+				// The target tiddler doesn't need double square brackets, but we'll silently remove them if they're present
+				if(target.indexOf("[[") === 0 && target.substr(-2) === "]]") {
+					target = target.substr(2,target.length - 4);
+				}
+				story.addToHistory(target);
+			} else if(storyList.length > 0) {
+				story.addToHistory(storyList[0]);
+			}
+		}
+	}
+	
+	/*
+	options: See below
+	options.updateAddressBar: "permalink", "permaview" or "no" (defaults to "permaview")
+	options.updateHistory: "yes" or "no" (defaults to "no")
+	options.copyToClipboard: "permalink", "permaview" or "no" (defaults to "no")
+	options.targetTiddler: optional title of target tiddler for permalink
+	*/
+	function updateLocationHash(options) {
+		// Get the story and the history stack
+		var storyList = $tw.wiki.getTiddlerList(DEFAULT_STORY_TITLE),
+			historyList = $tw.wiki.getTiddlerData(DEFAULT_HISTORY_TITLE,[]),
+			targetTiddler = "";
+		if(options.targetTiddler) {
+			targetTiddler = options.targetTiddler;
+		} else {
+			// The target tiddler is the one at the top of the stack
+			if(historyList.length > 0) {
+				targetTiddler = historyList[historyList.length-1].title;
+			}
+			// Blank the target tiddler if it isn't present in the story
+			if(storyList.indexOf(targetTiddler) === -1) {
+				targetTiddler = "";
+			}
+		}
+		// Assemble the location hash
+		switch(options.updateAddressBar) {
+			case "permalink":
+				$tw.locationHash = "#" + $tw.utils.encodeTiddlerTitle(targetTiddler);
+				break;
+			case "permaview":
+				$tw.locationHash = "#" + $tw.utils.encodeTiddlerTitle(targetTiddler) + ":" + $tw.utils.encodeTiddlerTitle($tw.utils.stringifyList(storyList));
+				break;
+		}
+		// Copy URL to the clipboard
+		switch(options.copyToClipboard) {
+			case "permalink":
+				$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + $tw.utils.encodeTiddlerTitle(targetTiddler));
+				break;
+			case "permaview":
+				$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + $tw.utils.encodeTiddlerTitle(targetTiddler) + ":" + $tw.utils.encodeTiddlerTitle($tw.utils.stringifyList(storyList)));
+				break;
+		}
+		// Only change the location hash if we must, thus avoiding unnecessary onhashchange events
+		if($tw.utils.getLocationHash() !== $tw.locationHash) {
+			if(options.updateHistory === "yes") {
+				// Assign the location hash so that history is updated
+				window.location.hash = $tw.locationHash;
+			} else {
+				// We use replace so that browser history isn't affected
+				window.location.replace(window.location.toString().split("#")[0] + $tw.locationHash);
+			}
+		}
+	}
+	
+	})();
 		targetTiddler = options.targetTiddler;
 	} else {
 		// The target tiddler is the one at the top of the stack
@@ -198,19 +425,19 @@ function updateLocationHash(options) {
 	// Assemble the location hash
 	switch(options.updateAddressBar) {
 		case "permalink":
-			$tw.locationHash = "#" + encodeURIComponent(targetTiddler);
+			$tw.locationHash = "#" + $tw.utils.encodeTiddlerTitle(targetTiddler);
 			break;
 		case "permaview":
-			$tw.locationHash = "#" + encodeURIComponent(targetTiddler) + ":" + encodeURIComponent($tw.utils.stringifyList(storyList));
+			$tw.locationHash = "#" + $tw.utils.encodeTiddlerTitle(targetTiddler) + ":" + $tw.utils.encodeFilterPath($tw.utils.stringifyList(storyList));
 			break;
 	}
 	// Copy URL to the clipboard
 	switch(options.copyToClipboard) {
 		case "permalink":
-			$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + encodeURIComponent(targetTiddler));
+			$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + $tw.utils.encodeTiddlerTitle(targetTiddler));
 			break;
 		case "permaview":
-			$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + encodeURIComponent(targetTiddler) + ":" + encodeURIComponent($tw.utils.stringifyList(storyList)));
+			$tw.utils.copyToClipboard($tw.utils.getLocationPath() + "#" + $tw.utils.encodeTiddlerTitle(targetTiddler) + ":" + $tw.utils.encodeFilterPath($tw.utils.stringifyList(storyList)));
 			break;
 	}
 	// Only change the location hash if we must, thus avoiding unnecessary onhashchange events
--- a/core/modules/utils/twuri-encoding.js
+++ b/core/modules/utils/twuri-encoding.js
@@ -1,0 +1,131 @@
+/*\
+title: $:/core/modules/utils/twuri-encoding.js
+type: application/javascript
+module-type: utils
+Utility functions related to permalink/permaview encoding/decoding.
+\*/
+(function(){
+
+	// The character that will substitute for a space in the URL
+	var SPACE_SUBSTITUTE = "+";
+
+	// The character added to the end to avoid ending with `.`, `?`, `!` or the like
+	var TRAILER = "+";
+
+	// The character that will separate out the list elements in the URL
+	var CONJUNCTION = ";";
+
+	// The encoded version of the space substitute as a regex
+	var ENCODED_SPACE_SUBSTITUTE = new RegExp(encodeURIComponent(SPACE_SUBSTITUTE).replace("%", "\\%"), "g");
+
+	// The encoded version of the trailer as a regex
+	var ENCODED_TRAILER = new RegExp(encodeURIComponent(TRAILER).replace("%", "\\%") + "$", "g");
+
+	// Those of the allowed url characters claimed by TW
+	var CLAIMED = [SPACE_SUBSTITUTE, ":", CONJUNCTION];
+
+	// Non-alphanumeric characters allowed in a URL fragment
+	// More information at https://www.rfc-editor.org/rfc/rfc3986#appendix-A
+	var VALID_IN_URL_FRAGMENT = "-._~!$&'()*+,;=:@/?".split("");
+
+	// The subset of the pchars we will not percent-encode in permalinks/permaviews
+	var SUBSTITUTES = []
+	$tw.utils.each(VALID_IN_URL_FRAGMENT, function(c) {
+		if (CLAIMED.indexOf(c) === -1) {
+			SUBSTITUTES.push(c)
+		}
+	});
+
+	// A regex to match the percent-encoded characters we will want to replace.
+	// Something similar to the following, depending on SPACE and CONJUNCTION
+	//     /(%2D|%2E|%7E|%21|%24|%26|%27|%28|%29|%2A|%2B|%3B|%3D|%40|%2F|%3F)/g
+
+	var CHAR_MATCH_STR = []
+	$tw.utils.each(SUBSTITUTES, function(c) {
+		CHAR_MATCH_STR.push("%" + c.charCodeAt(0).toString(16).toUpperCase())
+	})
+	var CHAR_MATCH = new RegExp("(" + CHAR_MATCH_STR.join("|") + ")", "g");
+
+	// A regex to match the SPACE_SUBSTITUTE character
+	var SPACE_MATCH = new RegExp("(\\" + SPACE_SUBSTITUTE + ")", "g");
+
+	// A regex to match URLs ending with sentence-ending punctuation
+	var SENTENCE_ENDING = new RegExp("(\\.|\\!|\\?|\\" + TRAILER + ")$", "g");
+
+	// A regex to match URLs ending with sentence-ending punctuation plus the TRAILER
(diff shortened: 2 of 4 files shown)


core/modules/startup/story.js
```diff
@@ -183,7 +183,234 @@ function updateLocationHash(options) {
 	var storyList = $tw.wiki.getTiddlerList(DEFAULT_STORY_TITLE),
 		historyList = $tw.wiki.getTiddlerData(DEFAULT_HISTORY_TITLE,[]),
 		targetTiddler = "";
-	if(options.targetTiddler) {
+	if(options.targetTiddler) {/*\
```
pmario, line 186: It seems there is a copy past problem here. In line 187 the source file seems to start again. You need to check that.

pmario:
The tests look OK and they are passing since Vercel runs them

core/modules/startup/story.js
CrossEye: Wow.  How did I miss that?!!  It was bedtime, and I was tired, but that's a large miss!

I will see if I can update from my phone.  Committing to public GitHub is not allowed here at GigantiCorp.  If not, I'll do it this evening from home.

CrossEye:
@pmario:
> The tests look OK and they are passing since Vercel runs them

Thanks for checking.  It's been some years since I did Jasmine.  I remember there were ways to focus on only a specific suite or spec.  I thought changing `describe` to `ddescribe`, `fdescribe` or `describe.only` would allow me to run just my own tests, but none of them worked.  Does anyone have a working technique for that?  Waiting for the full spec to run is annoying when I tend to do it after ever new assertion.

core/modules/startup/story.js
pmario: You should do it from home and test properly, prior to submitting a new change.

pmario:
@CrossEye .. The new preview seems to work and the automatic tests pass. But please test again at home. Just "to make me feel better" ;)

core/modules/startup/story.js
CrossEye: Too late.  I managed it from my phone.  This is bizarre, as it was running fine locally when I committed.  No clue how it was with that mess of a file.

I did some minimal testing on the Vercel preview, but I also can only see that from my phone.  I will check more thoroughly this evening.

CrossEye:
@pmario:

> @CrossEye .. The new preview seems to work and the automatic tests pass. But please test again at home. Just "to make me feel better" ;)

I've tested the preview and everything seems to work.  The test cases seem to capture most everything.  If I think of other edge cases, I'll add tests.

The only backward incompatibility I know about would happen when a user had manually created a URL for a title that included a `+`, and this hand-crafted URL used the `+` sign instead of the `%2B` that TW would generate.  This new code would interpret that as a space.  We've known about this from the start, and it seemed acceptable.

I did just consider one more ambiguous circumstance, but it seems already handled by TW.  This code adds a `+` to the end of the permalink for a title that ends in `.`, `?`, or `!`, and strips it out when decoding.  I thought that this might mean some ambiguity between `"Foo Bar! "` and `"Foo Bar!"` (with and without a trailing space in the title.)  But it seems that on tiddler save, TW trims off trailing space in the title, so I think we're okay here.

pmario:
@Jermolene ... can you review this one again? It should be backwards compatible now.

saqimtiaz:
@CrossEye Apologies for only realizing this so very late, but am I understanding correctly that multiple titles in the URL hash are now delimited by a semi-colon? The problem with this approach is that the permalink should be a valid filter string and the semicolons make that no longer the case.

I am also somewhat concerned that the logic used is convoluted enough that it would be quite difficult to implement the same logic in any third party code that wants to create valid permalinks for TiddlyWiki.

pmario:
>I am also somewhat concerned that the logic used is convoluted enough that it would be quite difficult to implement the same logic in any third party code that wants to create valid permalinks for TiddlyWiki.

But the old version still works. See the link below

https://tiddlywiki5-git-fork-crosseye-simplify-permalinks-jermolene.vercel.app/#Working%20with%20TiddlyWiki:%5B%5BWorking%20with%20TiddlyWiki%5D%5D%20Learning%20HelloThere%20%5B%5Btags%20Operator%5D%5D

Jermolene:
> But the old version still works. See the link below

It only works if none of the titles have any of the special characters that are handled by the new code.

CrossEye:
@saqimtiaz:

>  am I understanding correctly that multiple titles in the URL hash are now delimited by a semi-colon? The problem with this approach is that the permalink should be a valid filter string and the semicolons make that no longer the case.

Maybe I'm misunderstanding, but the current URLs are also not valid filter strings.  Think of this code as replacing `encodeURIComponent`/`decodeURIComponent` to have an nicer visible URL format.

Something has broken since I last tried this; I don't know if another change overlapped with this, if there's something going on with how this change was ported to Vercel, or something else..  I will look at it in the next few days to see what happened.  But when it's working, `#Filter+Syntax:Filter+Syntax;Filters;Dominant+Append;Filter+Operators` will be converted into a `target` field of `Filter Syntax` and a `storyFilter` field of `[[Filter Syntax]] Filters [[Dominant Append]] [[Filter Operators]]`, which I think is correct.

> I am also somewhat concerned that the logic used is convoluted enough that it would be quite difficult to implement the same logic in any third party code that wants to create valid permalinks for TiddlyWiki.

That's certainly not a scenario I considered.  As @pmario said, this shouldn't break any existing such tools.  The only backward incompatibility I know of (when this is actually working) is that if someone created a URL outside TW, and used a `+` sign instead of the `%2B` which would be generated by TW, then that URL would not be parsed correctly.  The new code would treat it as a space.

However, I can write a sample implementation for reconstructing this behavior in modern JS or a number of other languages.  It would be a lot simpler than the code that's here.  I can't quickly find the relevant discussion, but when I discussed removing many of the existing CONSTANTS in the code, and hard-coding their values where necessary, I was encouraged to keep them there as easier to follow.  But we could do it here, leading to reasonably straightforward code:

<details><summary>See code</summary>

```js
const CHAR_MATCH = /(%2D|%2E|%5F|%7E|%21|%24|%26|%27|%28|%29|%2A|%2C|%3D|%40|%2F|%3F)/g
const PCT_CHAR_MAP = {"%21": "!", "%24": "$", "%26": "&", "%27": "'", "%28": "(", "%29": ")", "%2A": "*", "%2C": ",", "%2D": "-", "%2E": ".", "%2F": "/", "%3D": "=", "%3F": "?", "%40": "@", "%5F": "_", "%7E": "~"}

const encodeTiddlerTitle = (s) =>
  encodeURIComponent(s.replace(/(\.|\!|\?|\+)$/g, "$1" + '+'))
    .replace(/\%20/g, '+').replace(/\%2B$/g, '+')
	.replace(CHAR_MATCH, ((_, c) => PCT_CHAR_MAP[c]))

const encodeFilterPath = (s) =>
  s.replace(/(\.|\!|\?|\+)$/g, "$1" + '+')
    .replace(/\[\[(.+?)\]\]/g, (_, t) =>  t.replace(/ /g, '+' ))
    .split(" ")
    .filter(Boolean)
    .map(s => s.trim())
    .map(encodeTiddlerTitle)
    .map(s => s.replace(/\%20/g, '+').replace(/\%2B/g, '+'))
    .map ((s) => s.replace(CHAR_MATCH, (_, c) => PCT_CHAR_MAP[c]))
    .join(';')

const encode = (title, filterPath) => [
  encodeTiddlerTitle(title),
  ...(filterPath ? [encodeFilterPath(filterPath)] : [])
].join(':')
```
### Demo ###

```js
console.log(encodeTiddlerTitle('A (...surprising?) tiddler with various *special/unusual* characters, & more'))
//=> A+(...surprising?)+tiddler+with+various+*special/unusual*+characters,+&+more

console.log(encodeFilterPath('[[Filter Syntax]] [[Conditional Operators]] Filters [[Dominant Append]] [[Filter Operators]]'))
//=> Filter+Syntax;Conditional+Operators;Filters;Dominant+Append;Filter+Operators

console.log(encode('Dominant Append'))
//=> Dominant+Append

console.log(encode('Dominant Append', '[[Filter Syntax]] [[Conditional Operators]] Filters [[Dominant Append]] [[Filter Operators]]'))
//=> Dominant+Append:Filter+Syntax;Conditional+Operators;Filters;Dominant+Append;Filter+Operators
```
</details>

@Jermolene:
>It only works if none of the titles have any of the special characters that are handled by the new code.

I will double-check when I get it working again, but I believe that the other special characters used (`:` and `;`) shouldn't cause any problem.  The other special characters in the source code are those that *don't* need percent encoding: by not doing so, we can use many other plain characters to get more readable URLs.  The complex example I use in the test cases has 

```
title: A (...surprising?) tiddler with various *special/unusual* characters, & more
current: https://tiddlywiki.com/#A%20(...surprising%3F)%20tiddler%20with%20various%20*special%2Funusual*%20characters%2C%20%26%20more
updated: https://tiddlywiki.com/#A+(...surprising?)+tiddler+with+various+*special/unusual*+characters,+&+more
```

CrossEye:
I redid this change against the trunk; there was a minor conflict with #8211.  [The result](https://github.com/CrossEye/TiddlyWiki5/commit/218af86d6be4921775723d03893e4d52b743f327) is for now in my fork.


Should I create a new pull request for this?  Should I make the minor updates to this one?  And in either case, should I do it now or wait until after 5.3.4 is released?

twMat:
>Should I create a new pull request for this? 

Please do, so things start afresh, and close this one.

The need for prettier permalinks is real.

CrossEye:
@pmario:

> @CrossEye -- There seems to be a problem with the code I tested locally by cloning your PR.

I am not getting the same behavior.  I will rebase against HEAD and see if that's happening now for me.

CrossEye:
@pmario, @saqimtiaz:

> I think that means, is a "wiki-text" plugin author wants to create "pretty URLs" they can not easily do so, because the responsible "filter operators" are missing.
> 
> JS-plugin authors can use the decode and encode $tw.utils.*TWURI* functions. So that should not be a problem.

OK, thank you for the interpretation.  I didn't follow it when first presented, but wasn't sure where this PR even stood, so I let it drop.

It certainly makes sense to offer these as filter operators.  But I am not certain about how far to take this.

This PR does not change the structure of how `core/modules/startup/story` constructs the permaviews/permalinks; it just changes the functions it calls to do so, replacing calls to `$tw.utils.decodeURIComponentSafe` with calls either to `$tw.utils.decodeTWURITarget` or `decodeTWURIList` and replacing calls to `encodeURIComponent` with calls either to `$tw.utils.encodeTiddlerTitle` or `$tw.utils.encodeFilterPath`.

I could just expose those four functions as filter operators.  That's the easier path.

The harder path would involve replicating the logic used to en/decode permalinks/views entirely.  Permalinks would be trivial (`[encode-pretty-uri<target>]`).  But permaviews would take more thinking.  Would it look like:

```text
[encode-pretty-uri<target>,<filter>]
```
or
```text
[encode-pretty-uti<target>,<titlelist>]
```

Which path seems right?

Another question if we're talking permalinks/permaviews is whether we're generating *fragments*:

```'text
#Customise+TiddlyWiki
#Customise+TiddlyWiki:Customise+TiddlyWiki;Working+with+TiddlyWiki
```

or *urls*:

```text
http://tiddlywiki.com/#Customise+TiddlyWiki
http://tiddlywiki.com/#Customise+TiddlyWiki:Customise+TiddlyWiki;Working+with+TiddlyWiki
```

ITSM that fragments would be more flexible, especially if your tool is generating URLs for a separate wiki.  But that might subvert users' expectations.

Suggestions on both of these questions are welcome!

pmario:
I though it would only need 2 new filter operators. 

`[<titleList>encode-pretty-uri[]]` and `[<pretty-uri-list>decode-pretty-uri[]]` ... That's it. So the input will be a list. It does not really matter, if it was produced by a filer or manually. And *-pretty-uri does not really need parameters. -- At least -- that's what I thought

pmario:
>Question 2
>
>Another question if we're talking permalinks/permaviews is whether we're generating fragments:

I think fragments are enough for the beginning. -- We will see what user feedback says.

CrossEye:
@pmario:

>> Another question if we're talking permalinks/permaviews is whether we're generating fragments:
> I think fragments are enough for the beginning. 

Ok, that's easier to do and I think is a better design as well.

CrossEye:
@pmario:

> I though it would only need 2 new filter operators.
> 
> `[<titleList>encode-pretty-uri[]]` and `[<pretty-uri-list>decode-pretty-uri[]]`

Right now, to encode a permaview, we need the `target` tiddler and the `storyList` of open tiddlers.

In this current URL:

https://tiddlywiki.com/#Features:HelloThere%20Learning%20Features%20Filters

We would have open, in order, 
  1. *HelloThere*
  2. *Learning*
  3. *Features*
  4. *Filters*

and we would be scrolled to and focused on the third one, *Features*

What's our data structure for the return from this?  Is is something like a titleList where the first entry means one thing (the target) and the remaining ones something else (the storyList)?  That strikes me as really awkward.

If we had separate functions to extract the target and the storyList, then this becomes a non-issue.  That's why I was thinking that we might need all four.  Do you have a cleaner suggestion?

pmario:
You are right. Permaview needs a target as a parameter. Do there may be 3 me operators

CrossEye:
The names seem to be growing long, but would these make sense?

> **titles**: [[Working with TiddlyWiki]] HelloThere [[Customise TiddlyWiki]] Learning
> **target**: Customise TiddlyWiki
> **fragment**: Customise+TiddlyWiki:Working+with+TiddlyWiki,HelloThere,Customise+TiddlyWiki,Learning
> 
> **[&lt;target>encode-pretty-uri-target[]]**: Customise+TiddlyWiki
> **[&lt;titles>encode-pretty-uri-titlelist[]]**: Working+with+TiddlyWiki,HelloThere,Customise+TiddlyWiki,Learning
> 
> **[&lt;fragment>decode-pretty-uri-target[]]**: [[Customise TiddlyWiki]]
> **[&lt;fragment>decode-pretty-uri-titlelist[]]**: [[Working with TiddlyWiki]] HelloThere [[Customise TiddlyWiki]] Learning

The penultimate one seems very questionable.  Should decode-target return `Customise TiddlyWiki` or `[[Customise TiddlyWiki]]` (brackets or not?)

Also, can we think of shorter names?

Jermolene:
Hi @CrossEye thank you again for your work on this.

I remain concerned that the encoding/decoding logic is now very complex. Not only does that make interoperability more difficult, it also makes it impossible for most people to run the logic in their heads. Another minor consequence of the complexity is that it is opinionated, making it more likely that users will find something that they disagree with.

So I wonder if this might be a good point to think about another approach that may actually be orthogonal: using explicit "slugs" associated with each tiddler. The idea is that the wiki author would explicitly assign a slug to a new tiddler, with the field being automatically generated on saving the tiddler if it hasn't been set manually. As well as using them for navigation, we could also use them for naming static HTML files rendered under Node.js.

Contemporary CMSs like Wordpress use slugs, so I think they are conceptually familiar to some of our users. They are a reasonably universal solution because authors can choose their own scheme for the slugs that makes sense for their content.

The algorithm you present here could then be used to make the default slugs. In effect, it just introduces another lay of indirection over the solution in this PR.

There would still be quite a lot to figure out with this approach. For example, the need to navigate to tiddlers that don't have a slug field would seem to imply that there would actually be a cascade, with navigation targets first being matched as a slug, and then as an encoded title as today.

CrossEye:
@Jermolene:

> I remain concerned that the encoding/decoding logic is now very complex. Not only does that make interoperability more difficult, it also makes it impossible for most people to run the logic in their heads.

Below, I'll describe the algorithms more explicitly, and how they would be used with the `$tw.utils` methods and the not-yet-written filter operators.  But first I want to address this:

> So I wonder if this might be a good point to think about another approach that may actually be orthogonal: using explicit "slugs" associated with each tiddler. 

We've discussed this before, and I've worried about this:

> Would a visitor arriving with #more-coffee be offered the choice between the three existing tiddlers More Coffee?, More Coffee!, and MORE COFFEE!!!? Or would we arbitrarily choose one? It only gets worse with permaviews.

What you're describing now allows us a way to deal with some of this: if `slug` becomes a field on most public tiddlers, we can then be explicit on how we want our tiddlers to appear in URLs.

But it raises some questions:

  * How does this work with permalinks?  While we could combine slugs with punctuation (for instance, `#customise+tiddlywiki:working-with-tiddlywiki,hellothere,customise-tiddlywiki,learning`, this wouldn't help with arbitrary filters, even one as simple as https://tiddlywiki.com/#:[tag[Filters]].  Would we keep a new wiki-wide structure that maps, say, `#all-filters` to that permalink?  That would solve [an issue][t1] I raised almost two years ago
  * How do we deal with conflicts?  If we have a title that would default to a slug which already appears on another tiddler, do we warn as we do about overwriting titles?  It's clearly less severe than that, but it could definitely cause problems.  Would we try to interrupt the `save` if this is not addressed?  Similarly, what happens if slugs of imported tiddlers match slugs on the current wiki?
  * If this can be overridden, how does a user do so?  Do they overwrite something like `$:/config/default-slug-generator`?  And what will they have to write to do so?  A procedure/macro?  A function? Some sort of Widget?  Could they use conditionals (`<$if [tag[Journal]] %> ... <% else %> ... <% endif %>`?)

> The algorithm you present here could then be used to make the default slugs. In effect, it just introduces another lay[er] of indirection over the solution in this PR.

If we do go down this path, that sounds like a good idea.

## Describing the algorithm ##

The implementation of the algorithm can probably be simplified if we settle on the separators/joiners required for the url.  The code is full of indirection through constants so that we could fiddle with the parts.  I almost -- almost -- wrote a code generator based on these constants to show simpler code.   But we can do that at any time once we're sure enough about the actual URLs it generates.

For now, we have a fairly simple usage.

### Encoding Permaviews with `$tw.util.*` methods or (still-to-come) filter operators ###

1. Call `encode-target` on the target
2. Call `encode-titlelist` on the titlelist
3. Join them with a `:`

Decoding Permaviews is similar.  And en/decoding Permalinks is much simpler.  I don't think that's much harder than the current mechanism.

Of course this buries some of the complexity of the implementation, which does something like this:

### Encoding Permaviews, underlying algorithm (slightly idealized) ###

1. For the target titile and every title in the story,
  a. Replace all characters that aren't space, a-z, A-Z, 0-9, or one of these {`-`, `.`, `_`, `~`, `!`, `$`, `&`, `'`, `(`, `)`, `*`, `,`, `=`, `@`, `/`, `?`, `"`} with their percent-encoded version.
  b. Replace all spaces with `+`
  c. If the title ends with one of {`.`, `?`, `!`}  add a single trailing `+`
2. Join the resulting strings for the story with `;`
3. Join the resulting string for the target and the result from step 2 with a `:`

(The idealization is that I actually replace *all* special character with their percent-encoding, then convert back the ones mentioned.  I think next pass will try to fix this.  Hey, it made sense at the time! 😉 )

  [t1]: https://talk.tiddlywiki.org/t/5331

CrossEye:
@Jermolene:

> I think there is a strong case for [slugs]

Agreed.  I had considered them before I started with my current approach.

But I also think there is a fairly strong usability case against them.  The issues can likely be worked out, but when a wiki is partly pieced-together out of content from many different users (plug-ins, themes, arbitrary tiddler-bearing JSON files) it seems to me that maintaining slug fields could become quite burdensome.  I'd love to hear about techniques that clean this up; I haven't been able to come up with them myself.

> [W]e can develop the feature in a way that accommodates your codec algorithm

That would be great.  I think this style permalink will be a distinct improvement.

> Another thing to throw into the mix is that we have a long term desire to move all the processing of the location hash into wikitext. 

Is this because we want to move *everything* we can that way, or is there something more specific to the hash?

I'm trying to determine if I can comfortably port my code into plain wikitext.  I'll think about it.

> Part of the processing might be to use a cascade to convert the incoming identifiers into tiddler titles.

...and back again, too, presumably.  Obviously, what I'm doing could integrate well with this.  I like the notion of a cascade here, possibly a dynamic one (if that's possible) based on settings: try the `#more-coffee` algorithm first, then `#More+Coffee?+`,  and if both fail to generate a new unique slug, default back to `#More%20Coffee%3F`, allowing custom implementations to fall before, after, or between those options.

----------

The biggest problem I have with slugs is that it sounds as though creating them could no longer be a pure function, but would depend also on the state of the tiddler store.  There were several discussions in this thread, the previous one, and on *talk.tiddlywiki* that discussed doing so, and I'm pretty sure all the core people involved agreed with me that this was not a good idea.  Any `more-coffee`-style slug is a lossy translation of `More Coffee?`, and so would probably need to refer to the tiddler store to note duplicates.  Clearly this is not the end of the world; we already need to do this for the `title` field.  But it is worrisome.

Jermolene:
Hi @CrossEye I do agree that there are issues to be worked out with the implementation of the edge cases for slugs and how to apply the default codec. That's why it might make sense to move the location hash processing into wikitext. Then it would be possible for a plugin to override the handling provided by the core.

It should be fairly straightforward to do, with several parts:

* Support for a new action string that is invoked at startup and when the location hash changes
* A new core location hash handler global procedure that implements the current behaviour of parsing the location hash, updating the story list and performing the associated navigation
* Something like a custom global function or a cascade that converts a target tiddler title and a list of story list tiddler titles into a location hash

I think this would be a good way forward. It would allow you to package this approach as a plugin, and allow everyone to experiment with new techniques.

CrossEye:
@Jermolene:
> @CrossEye I do agree that there are issues to be worked out with the implementation of the edge cases for slugs and how to apply the default codec. That's why it might make sense to move the location hash processing into wikitext. Then it would be possible for a plugin to override the handling provided by the core.

This sounds like a good idea.  Unfortunately, at the moment, a lot of the details below fall outside my current TW understanding.  I'm hoping you and others on the core team can have the patience to help me figure it out, as I would really like to contribute in this area.


> * Support for a new action string that is invoked at startup and when the location hash changes

What is meant here by "a new action string"?  Are there some examples, especially any that are tied to startup and/or to location hash changes?


> * A new core location hash handler global procedure that implements the current behaviour of parsing the location hash, updating the story list and performing the associated navigation

Again, it would be helpful to have examples of somewhat similar global procedures.  

> * Something like a custom global function or a cascade that converts a target tiddler title and a list of story list tiddler titles into a location hash

That I think I get, although I haven't looked at the details about whether the cascades I know are implemented in wikitext or in JS.

Am I wrong to think of cascades as just a type of function, with a top-level declarative `if-else`/`switch` construct?  Because that's the model I have in my mind.  Or more precisely, cascades *return* functions.  In plain JS it might look [like this][rr].

> I think this would be a good way forward. It would allow you to package this approach as a plugin, and allow everyone to experiment with new techniques.

My original hope was that this would start as a plugin, but when I thought I was changing `boot.js`, I gave it up.  If this allows me to do it that way, it useful to be able to test out if the idea is sound enough for most everyone before including it in the core.


  [rr]: https://ramdajs.com/repl/#?console.clear%28%29%0A%0Aconst%20cascade%20%3D%20%28pairs%29%20%3D%3E%20%28...input%29%20%3D%3E%20%0A%20%20%28pairs.find%28pair%20%3D%3E%20pair%5B0%5D%28...input%29%29%20%7C%7C%20%5Bnull%2C%20%28noop%29%20%3D%3E%20undefined%5D%29%5B1%5D%28...input%29%0A%0Aconst%20stageOfLife%20%3D%20cascade%20%28%5B%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%201%2C%20%28%7Bname%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20just%20a%20baby%60%5D%2C%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%203%2C%20%28%7Bname%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20a%20toddler%60%5D%2C%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%2013%2C%20%28%7Bname%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20a%20child%60%5D%2C%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%2018%2C%20%28%7Bname%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20a%20teenager%60%5D%2C%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%2021%2C%20%28%7Bname%2C%20age%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20%28technically%29%20an%20adult%2C%20but%20at%20%24%7Bage%7D%20is%20still%20below%20drinking%20age%60%5D%2C%0A%20%20%5B%28%7Bage%7D%29%20%3D%3E%20age%20%3C%2035%2C%20%28%7Bname%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20an%20adult%60%5D%2C%0A%20%20%5B%28%29%20%3D%3E%20true%2C%20%28%7Bname%2C%20age%7D%29%20%3D%3E%20%60%24%7Bname%7D%20is%20an%20adult%2C%20and%20at%20%24%7Bage%7D%20is%20eligible%20to%20be%20President%20of%20the%20USA%60%5D%2C%0A%5D%29%0A%0Aconst%20people%20%3D%20%5B%0A%20%20%7Bname%3A%20%27Alea%27%2C%20age%3A%2015%7D%2C%0A%20%20%7Bname%3A%20%27Alden%27%2C%20age%3A%2019%7D%2C%0A%20%20%7Bname%3A%20%27Ayden%27%2C%20age%3A%2021%7D%2C%0A%20%20%7Bname%3A%20%27Benji%27%2C%20age%3A%201%7D%2C%0A%20%20%7Bname%3A%20%27Kayleigh%27%2C%20age%3A%2024%7D%2C%0A%20%20%7Bname%3A%20%27Scott%27%2C%20age%3A%2057%7D%2C%0A%20%20%7Bname%3A%20%27Teddy%27%2C%20age%3A%2027%7D%2C%0A%5D%0A%0Apeople.forEach%28%28p%29%20%3D%3E%20console.log%28stageOfLife%28p%29%29%29%20

CrossEye:
I have a first pass at implementing these in wikitext, without the indirection through levels of constants.  It's not quite working yet.  There are still a few bugs, but I won't have time to look at this again until Tuesday or Wednesday.  But I think this shows that the algorithms won't be too hard:

It turned out to be relatively simple:

```
title: Simpler Permalinks

<!-- Helpers -->
\define cs() [^a-z0-9\-\.\_\~\!\$\&\'\(\)\*\,\=\@\/\? ]

\function uri.encode()
  [regexp<cs>encodeuricomponent[]] :else[all[]]
\end

<!-- Public functions -->
\function encode.uri.target()
  [search-replace:g:regexp[(\.|\!|\?|\+)$],[$1+]split[]] :map[uri.encode[]] +[join[]search-replace:g:regexp[ ],[+]search-replace:g:regexp[\%2B$],[+]]
\end

\function decode.uri.target()
  [search-replace:g:regexp[(\.|\!|\?|\+)],[$1]search-replace:g:regexp[(\+)],[ ]decodeuricomponent[]]
\end

\function encode.uri.list()
  [enlist-input[]] :map[encode.uri.target[]] +[join[;]]
\end

\function decode.uri.list()
  [enlist-input[]decode.uri.target[]split[;]] :map[format:titlelist[]] +[join[ ]]
\end


!! Converting to and from Permalinks/Permaviews


<$let title-list ="[[Foo Bar]] Baz [[Corge Grault++]]">

|!Input                                     |!Operator           |!Output                                                                                    |
|backtranscludes Operator (Examples) #1!    |encode.uri.target[] |<$text text={{{ [[backtranscludes Operator (Examples) #1!]encode.uri.target[]] }}} />      |
|backtranscludes+Operator+(Examples)+%231!+ |decode.uri.target[] |<$text text={{{ [[backtranscludes+Operator+(Examples)+%231!+]decode.uri.target[]] }}} />   |
|<$text text=<<title-list>> />              |encode.uri.list[]   |<$text text={{{ [<title-list>encode.uri.list[]] }}} />                                     |
|Foo+Bar;Baz;Corge+Grault%2B%2B+            |decode.uri.list[]   |<$text text={{{ [[Foo+Bar;Baz;Corge+Grault%2B%2B+]decode.uri.list[]] }}} />                |

</$let>    
```

CrossEye:
I would like to start looking at this again for `5.4.0`.  

I'm looking at the most recent comment by @Jermolene, and after some more months with TW, I understand most of it, but I still have several questions:

>  That's why it might make sense to move the location hash processing into wikitext. Then it would be possible for a plugin to override the handling provided by the core.

Ok, I have a mostly working version in my previous post.  I sure I can work out whatever kinks remain.


> It should be fairly straightforward to do, with several parts:
> 
>  * Support for a new action string that is invoked at startup and when the location hash changes

I'm still not sure what this means.  Are there examples of "action strings" that run at startup and on some other event?

>  * A new core location hash handler global procedure that implements the current behaviour of parsing the location hash, updating the story list and performing the associated navigation

This is clear enough to me.  But this is a *core* change, yes?  It's a core change that would allow us to use plugin/other custom code to override the default behavior.

>  * Something like a custom global function or a cascade that converts a target tiddler title and a list of story list tiddler titles into a location hash

Is the idea here for the "custom global function" option that the user/plugin could simply *override* that function?

A cascade makes more sense to me, as it would allow us to reuse existing handlers, say for slugs.  For instance,

```
[<target>has[_slug]then[slug-handler]]
[<storyList>count[]!match[0]then[custom-permaview-handler]]
[[custom-permalink-handler]]
```

(Please pardon me if the syntax isn't quite right; I assume the idea is clear.)


> I think this would be a good way forward. It would allow you to package this approach as a plugin, and allow everyone to experiment with new techniques.

I just want to be sure that what you expect would involve *core* changes that themselves would allow users to plug in their own versions of these functions.  Is that correct?  If not, and you think this can all be done as a plugin, could you give more details?  Because I don't yet see how.

----------

But the discussion here is about how to *create* permalinks/views.  Equally important is how to *parse* them 

The current code looks like this:

```

		if(split === -1) {
			target = $tw.utils.decodeURIComponentSafe(hash.trim());
		} else {
			target = $tw.utils.decodeURIComponentSafe(hash.substr(0,split).trim());
			storyFilter = $tw.utils.decodeURIComponentSafe(hash.substr(split + 1).trim());
		}
```

And my version above only changes the functions called.  I don't know of a simple way to do this with a single-function override.  Basically the only data structures we seem to deal with via wikitext are Strings and lists of Strings.  (Yes, those Strings can be interpreted as numbers, but that's not relevant here.)

What we'd really like to return from a custom function in this case is something like:

```
{
    target: String,
    storyList: [String]
}
```

Is there some simple way in wikitext to create such data structures?  We *could* prepend the target to the story list, but that's pretty awkward to do and hard to explain.

Or would we insist that anyone trying to override the perma-* handling has to implement all three of these?:

  * **`createPerma`** : `(target, storyList) -> urlFragment`
  * **`parseTarget`** : `urlFragment-> target`
  * **`parseStories`** : `urlFragment-> storyList`

That's certainly possible, but it's ugly.  Is that the best we can do?

ping @Jermolene, @pmario, @saqimtiaz

saqimtiaz:
Hi @CrossEye, I am on the mend from some knee surgery yesterday and will need a week or so before I can be at my desk for longer stretches, but then I would love to engage with you on this more thoroughly.

In short, we need to implement a mechanism that allows handing the URL hash entirely using wikitext, including:
- the default current handling of permalinks, rewritten as wikitext
- the ability to replace the current handling with support for alternative permalink formats such as what this PR proposes, entirely in wikitext. We may need to introduce new filter operators to make this more userfriendly.
- the ability to invoke [custom wikitext startup action](https://tiddlywiki.com/#StartupActions)s to handle keyword/value pairs, such as for instance `mode:safe` or `mode:readonly` or passing a username and loading the appropriate data into view.
- we would want something akin to a cascade, but perhaps different in that instead of stopping at the first match, we might want to allow multiple handlers to act on the hash. For example, one might determine the storylist to open and the other might set the wiki to read only.

A few thoughts on implementation:
- we should endeavour to minimize backwards compatiblity breaks as much as possible.
- one escape hatch is that any hash starting with # such as tiddlywiki.com##mything is not processed at all currently. The official "[share](https://github.com/TiddlyWiki/TiddlyWiki5/tree/master/plugins/tiddlywiki/share)"plugin uses this as an escape route to implement custom startup behaviour. So if we use this as an escape hatch to implement the new behaviour, we just need to make sure that we update the share plugin simultaneously so it continues to function.
- currently a hash in TiddlyWiki will never include a # character. So the presence of a single # character could also be used to trigger the new behaviour.

Apologies if these aren't the most coherent or well articulated thoughts, but hopefully they will give you something mullover and help kick start the thinking process.

CrossEye:
@saqimtiaz:
> I am on the mend from some knee surgery yesterday

Good luck with the recovery!  My wife has done one and will soon do the other.  I know there are all sorts of stories, good and bad, about such recoveries.  Hers was easy and quick; I hope yours is the same.

> will need a week or so before I can be at my desk for longer stretches, but then I would love to engage with you on this more thoroughly.

Ok, heal well.  I'm eager to try this, but can certainly wait during your recuperation.

> the default current handling of permalinks, rewritten as wikitext

I was thinking that what really needed to be overridable -- and hence in wikitext -- are the parts that parse the target and/or storyList/filter from a URL fragment and the part that creates such a fragment from the target and list.  Do you think there's more for wikitext?  I had also been assuming that the `:` separator between the part would be universal, and that splitting on the `:` might happen before the wikitext implementation is called, and that joining on `:` might happen after it's called.

But there's certainly an argument to be made to get more generic and turn over the whole process to overridable wikitext.  For instance I can easily imagine someone wanting a url like

```
https://mywiki.example.com/#View/123
```

which parses a tiddler such as the following to determine not just which tiddlers to load in the story river and which one to focus on, but also which layout to use:

```
title: $:/my-views/123
tags: StartupView
layout: $:/my/custom/layout
open: [[Tiddler Foo]] [[Tiddler Bar]] [tag[Baz]] +[!sort[modified]]
focus: Tiddler Bar
```

> the ability to replace the current handling with support for alternative permalink formats such as what this PR proposes, entirely in wikitext.

Right.  For the current one, I had a pretty good start at that.  I can't remember what flaws it still had, but I know they exist.  That implementation [is above][c1].  When finished, it could perhaps stand as a good example of the idea, as would a second slug-based version.

> the ability to invoke [custom wikitext startup action](https://tiddlywiki.com/#StartupActions)s to handle keyword/value pairs, such as for instance mode:safe or mode:readonly or passing a username and loading the appropriate data into view.

I seem to have a block on this.  I understand StartupActions, and use them often enough, But what would this mean here?  How would you picture them specified?  And, as I think Jeremy was alluding to above, they would need to run at other times, right, depending on the values of `$:/config/Navigation/UpdateAddressBar` and `$:/config/Navigation/UpdateHistory`?

> we would want something akin to a cascade, but perhaps different in that instead of stopping at the first match, we might want to allow multiple handlers to act on the hash. For example, one might determine the storylist to open and the other might set the wiki to read only.

This makes me think of middleware, something similar to [Express' version][em].

> we should endeavour to minimize backwards compatiblity breaks as much as possible.

Of course.  My approach above is built on that idea, with only a tiny known backward incompatibility.   I would assume we always want to do the same.

> one escape hatch is that any hash starting with # such as tiddlywiki.com##mything is not processed at all currently. 

That does lead to some interesting possibilities!

> currently a hash in TiddlyWiki will never include a # character. So the presence of a single # character could also be used to trigger the new behaviour.

Possibly.  I could change my version by replacing the semicolon divider with the octothorpe, and requiring it appear even if there is no storyList/filter:

```
http://tiddlywiki.com/#Customise+TiddlyWiki#
http://tiddlywiki.com/#Customise+TiddlyWiki#Customise+TiddlyWiki;Working+with+TiddlyWiki
```

But, unless we go middleware style, I think I would prefer that the wiki has only one perma-* handler active, and not have some dynamic switching between them based on the URL content.  I just don't like those complexities.

----------

> Apologies if these aren't the most coherent or well articulated thoughts, 

On the contrary, these are all clear and helpful.  While there are still discussions to be had, I follow everything here, except for the part about the startup actions.

It doesn't answer all the questions I had in my previous post, and I'm still struggling with the interface we would ask url-handling plugins to implement, but I think your ideas themselves are quite straightforward.


> but hopefully they will give you something mullover and help kick start the thinking process.

Absolutely.

Thank you.  I hope you recover quickly!


  [c1]: https://github.com/TiddlyWiki/TiddlyWiki5/pull/7990#issuecomment-2351020020
  [em]: https://expressjs.com/en/guide/using-middleware.html

CrossEye:
@saqimtiaz:  I hope your recovery is going well!

Please let me know where you think I might focus my efforts on this!

saqimtiaz:
@CrossEye I have been trying to flesh out my ideas a bit in terms of implementation details. I will open a new issue early next week, since the intended functionality goes beyond just permalinks.

Jermolene:
Hi @CrossEye thanks for your work on this. @saqimtiaz has now posted #9174 outlining a more flexible approach which I think is necessary here. I'm therefore closing this PR in favour of #9174, and I hope we can work on that soon. We will also need to take the learnings from this thread.

CrossEye:
Yes, I've been following that.  I'm worried that it's stalled, though.  But perhaps it will pick back up soon.

PR #7991, opened 2024-02-15
MWS: Use transactions when modifying multiple resources

hoelzro:
This prevents partial changes from entering the database, and also nets a nice speed-up.

See also [https://github.com/Jermolene/TiddlyWiki5/discussions/7970](https://github.com/Jermolene/TiddlyWiki5/discussions/7970#discussioncomment-8425204) for addtional context.

### DIFF
--- a/plugins/tiddlywiki/multiwikiserver/modules/sql-tiddler-database.js
+++ b/plugins/tiddlywiki/multiwikiserver/modules/sql-tiddler-database.js
@@ -452,6 +452,13 @@ SqlTiddlerDatabase.prototype.getRecipeBags = function(recipename) {
 	return rows.map(value => value.bag_name);
 };
 
+/*
+Execute the given function in a transaction, committing if successful but rolling back if an error occurs.  Returns whatever the given function returns.
+*/
+SqlTiddlerDatabase.prototype.transaction = function(fn) {
+	return this.db.transaction(fn)();
+};
+
 exports.SqlTiddlerDatabase = SqlTiddlerDatabase;
 
 })();
--- a/plugins/tiddlywiki/multiwikiserver/modules/sql-tiddler-database.js
+++ b/plugins/tiddlywiki/multiwikiserver/modules/sql-tiddler-database.js
@@ -25,6 +25,7 @@ function SqlTiddlerDatabase(options) {
 	}
 	var databasePath = options.databasePath || ":memory:";
 	this.db = new $tw.sqlite3.Database(databasePath,{verbose: undefined && console.log});
+	this.transactionDepth = 0;
 }
 
 SqlTiddlerDatabase.prototype.close = function() {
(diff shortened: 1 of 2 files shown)


hoelzro:
This passes tests and seems correct to me, but there are a few things I'd like to call attention to:

  * `deleteAllTiddlersInBag` is called from `saveTiddlersFromPath`, so its invocation of `sqlTiddlerDatabase.transaction` could result in a nested transaction.  I don't think this should cause a problem, but it's something I wanted to mention, especially if there's a desire to extend support to other SQL implementations in the future, some of which may not support nested transactions.
  * I'm not the most knowledgeable when it comes to Node and its event loop, but I imagine that transactions could result in some odd behavior if used with async APIs.  `better-sqlite3` seems to be completely synchronous, as is the `SqlTiddlerDatabase` API, so that's not currently an issue, but that's another thing to keep in mind for the future.

Jermolene:
Thanks @hoelzro looks good

> * `deleteAllTiddlersInBag` is called from `saveTiddlersFromPath`, so its invocation of `sqlTiddlerDatabase.transaction` could result in a nested transaction.  I don't think this should cause a problem, but it's something I wanted to mention, especially if there's a desire to extend support to other SQL implementations in the future, some of which may not support nested transactions.

If that is a risk, perhaps we just need to make sure that our wrapper for the SQLite `transaction()` method is capable of being refactored to maintain a depth counter, and only invoke the database engine method for top level transactions? I think that is probably already the case, but it maybe worth an explicit comment to clarify our intent.

> * I'm not the most knowledgeable when it comes to Node and its event loop, but I imagine that transactions could result in some odd behavior if used with async APIs. 

I think that as long as the MWS APIs remain resolutely synchronous then there should be no problem for asynchronous code to use them.

The work so far could have been approached differently in an asynchronous manner because the Node.js APIs that drive things are themselves async. But the synchronous behaviour will be important when we reuse this implementation in the browser, where we do not have the option of an async store without sacrificing interactive responsiveness.

> `better-sqlite3` seems to be completely synchronous, as is the `SqlTiddlerDatabase` API, so that's not currently an issue, but that's another thing to keep in mind for the future.

I was stunned that the dominant Node.js wrapper for SQLite was async. I very much chose better-sqlite3 because of their approach to this issue. To quote [their docs](https://github.com/WiseLibs/better-sqlite3?tab=readme-ov-file#why-should-i-use-this-instead-of-node-sqlite3):

> node-sqlite3 uses asynchronous APIs for tasks that are either CPU-bound or serialized. That's not only bad design, but it wastes tons of resources. It also causes [mutex thrashing](https://en.wikipedia.org/wiki/Resource_contention) which has devastating effects on performance.

hoelzro:
> If that is a risk, perhaps we just need to make sure that our wrapper for the SQLite transaction() method is capable of being refactored to maintain a depth counter, and only invoke the database engine method for top level transactions? I think that is probably already the case, but it maybe worth an explicit comment to clarify our intent.

hoelzro:
> If that is a risk, perhaps we just need to make sure that our wrapper for the SQLite transaction() method is capable of being refactored to maintain a depth counter, and only invoke the database engine method for top level transactions? I think that is probably already the case, but it maybe worth an explicit comment to clarify our intent.

Ok, added that!

> But the synchronous behaviour will be important when we reuse this implementation in the browser, where we do not have the option of an async store without sacrificing interactive responsiveness.

Sorry, I might be misintepreting something or have have gaps in my knowledge here, but I was under the impression that the browser _does_ offer asynchronous storage APIs (like IndexedDB) - and wouldn't we want to make use of those in the interest of responsiveness?  Then again, I have yet to acquaint myself with that "SQLite on WASM" experiment you did, so maybe I'm missing some important details!

> I was stunned that the dominant Node.js wrapper for SQLite was async

I find that unusual as well - I'm guessing it was done in the interest of conforming to Node's ecosystem, where async is the rule rather than the exception.

Jermolene:
Thanks @hoelzro much appreciated

PR #7992, opened 2024-02-16
Dumpvariables rewrite new syntax plus search

pmario:
This PR is a draft atm. It's "Work In Progress". These extensions where inspired by @AnthonyMuscio at [Talk](https://talk.tiddlywiki.org/t/scope-of-variables-operator-how-can-i-display-variables-come-from-certain-namespace-or-scope/3979/6?u=pmario)

- It improves the `dumpvariables` macro and uses v5.3.x syntax
- It adds a completely new `search-variables` macro
- It is proposed to add the new search-variables feature to the $:/AdvancedSearch

The search-variables macro allows us to filter the list with a search text-input

For **testing** the PR adds a new tab to the $:/AdvancedSeach dialogue

image

The list can be extended, which opens a second search input, that shows, where the macro text is used and/or where it is defined. While working with the macro it turned out to be a very useful feature. 

image

@Jermolene @saqimtiaz ... What do you think?

I'll start a new thread at Talk, to get some feedback there too, once Vercel built a preview.

### DIFF
--- a/core/wiki/macros/search-variables.tid
+++ b/core/wiki/macros/search-variables.tid
@@ -11,8 +11,8 @@ tags: $:/tags/Global
 \procedure DV-SORT-OPTIONS() $:/temp/varSearch/sort
 
 <!-- Filter strings selected by "Sort" dropdown -->
-\procedure DV-RAW-FILTER-STR() [variables:raw<_wlf.dv-type>]
-\procedure DV-FILTER-STR() [variables<_wlf.dv-type>]
+\procedure DV-RAW-FILTER-STR() [variables-wl:raw<_wlf.dv-type>]
+\procedure DV-FILTER-STR() [variables-wl<_wlf.dv-type>]
 
 <!-- State or temp variable base-names -->
 \procedure DV-FOLDED() $:/temp/varFolded


pmario:
The second input allows us to add more words to the filter string, to shorten the list 

image

pmario:
The dumpvariables macro also got a `subfilter` parameter

image

pmario:
Talk thread is at: https://talk.tiddlywiki.org/t/proposal-dumpvariables-and-search-variables-macros-are-available-as-a-gh-preview/9145

saqimtiaz:
@pmario I would highly recommend having separate PRs for new features, and for the syntax rewrite of existing macros. Especially for the latter that would make it much easier to review and merge them.

pmario:
Yes, I can do that, since it's only 11 lines of code. Then this PR will completely rewrite it.

Jermolene:
Hi @pmario this is an ambitious piece of work, thank you. I am concerned that it has grown both too complex and insufficiently universal for inclusion in the core.

It looks as though we would need some minor core tweaks before you could implement this functionality as a plugin. I was thinking of a new operator `getvariablejson` that would return a blob of JSON to describe a variable, in more or less the format returned by `widget.getVariableInfo()`. Would that be sufficient?

pmario:
>I am concerned that it has grown both too complex and insufficiently universal for inclusion in the core.

As it is now -- yes. 

The whole thing can be split up into 7 different PRs, where every one of them should be universal enough. 

1. I did find a bug in `widget.js`, which can be a separated PR
2. `getvariable` would need access to `.getVariableInfo()`, which you pointed out. I think it could be a new operator in a separated PR
3. The `variables` operator would get an "include" parameter similar to the "fields" operator
4. `utils.js` and the "format" operator would need to be expanded to allow us better formatting for the variable info.

With those changes in place, 

5. The following would be 1 PR
    - the **dumpvariables** and **search-variables** macro are both stored in **1 tiddler** 
    - `search.multids` would need the new text-elements for translation
    - 1 or 2 exclude-snippets should be part of tw-com docs (not the core)

The rest of the content in this **draft** is documentation and some test tiddlers, which would be changed to:

6. Some real tests, which remove the test-xx tiddlers from this PR
7. New documentation

So in sum there can be 7 nicely separated PRs, that lets us implement 

- dumpvariables in backwards compatible way, but with much better filtering capabilities
- search-varaibles as a stand alone macro which can be used for documentation within a tiddler using the `tpye, sort, subfilter` parameters
- 1 tiddler to include search-variables into the AdvancedSearch panel. 

I think we will need

- 1.) up to 4.) to be part of the core.
- 5) could be a core plugin
- 6) would be part of tw-com documentation edition

pmario:
@Jermolene I did create a `getvariablejson` operator, as you suggested. But it is not really convenient. Getting the data, that is needed in a nicely formatted way is a nightmare.

What I want is:

- Show a prefix `\function`
- show the function name: `.concat`
- show a list of parameters as it is defined: `(1, 2, 3, 4, 5)`
- show the definition of the function
- show the value of the function, if it has one. 

eg: grafik

To test if I can get the needed info with `getvariablejson` I did this:

```
<pre><$text text={{{[[.concat]getvariablejson[]]}}}/></pre>
```

And got this

```
{"text":"","params":[{"name":"1","value":""},{"name":"2","value":""},{"name":"3","value":""},{"name":"4","value":""},{"name":"5","value":""}],"resultList":[""],"srcVariable":{"value":"[[$(1)$$(2)$$(3)$$(4)$$(5)$]substitute[]]","params":[{"name":"1"},{"name":"2"},{"name":"3"},{"name":"4"},{"name":"5"}],"isMacroDefinition":false,"isFunctionDefinition":true,"isProcedureDefinition":false,"isWidgetDefinition":false,"configTrimWhiteSpace":false},"isCacheable":false}
```

That's useless for humans. 

To get something readable I need to do this: 

```
\function tab() [charcode[9]]

<pre><$text text={{{[[.concat]getvariablejson[]format:json<tab>]}}}/></pre>
```

Which returns the following output. It's better but also pretty much useless for ordinary users. 

```
{
	"text": "",
	"params": [
		{
			"name": "1",
			"value": ""
		},
		{
			"name": "2",
			"value": ""
		},
		{
			"name": "3",
			"value": ""
		},
		{
			"name": "4",
			"value": ""
		},
		{
			"name": "5",
			"value": ""
		}
	],
	"resultList": [
		""
	],
	"srcVariable": {
		"value": "[[$(1)$$(2)$$(3)$$(4)$$(5)$]substitute[]]",
		"params": [
			{
				"name": "1"
			},
			{
				"name": "2"
			},
			{
				"name": "3"
			},
			{
				"name": "4"
			},
			{
				"name": "5"
			}
		],
		"isMacroDefinition": false,
		"isFunctionDefinition": true,
		"isProcedureDefinition": false,
		"isWidgetDefinition": false,
		"configTrimWhiteSpace": false
	},
	"isCacheable": false
}
```

The only thing which can be used out of the box is the definition. 

```
<pre><$text text={{{[[.concat]getvariablejson[]jsonextract[srcVariable],[value]format:json<tab>]}}}/></pre>
```

Which returns

```
"[[$(1)$$(2)$$(3)$$(4)$$(5)$]substitute[]]"
```

- To get the `\function` prefix I need to write a complex logic, I do not even want to think about 
- To get all the `params` in a nicely formatted way using wikitext is similarly complex. 

To get params I would do this: 

```
<pre><$text text={{{[[.concat]getvariablejson[]jsonextract[srcVariable],[params]format:json<tab>]}}}/></pre>
```

which gives me

```
[
	{
		"name": "1"
	},
	{
		"name": "2"
	},
	{
		"name": "3"
	},
	{
		"name": "4"
	},
	{
		"name": "5"
	}
]
```

I have no Idea, how I could convert this info into `(1, 2, 3, 4, 5)` using wikitext. And there are not even `default:` values used here.

As I wrote it's a nightmare, **or I may be completely missing something**

### My Conclusion

- **As a developer** I think it's fine to get all the info as JSON and I can create a separated PR for this. **but**
- **As a user** I do expect something like this: 

```
<pre><$text text={{{[[.concat]format:variable[]]}}}/></pre>
```

Which should show this:

```
\function .concat(1, 2, 3, 4, 5) [[$(1)$$(2)$$(3)$$(4)$$(5)$]substitute[]]
```

pmario:
@Jermolene -- My experiments with getvariablejson can be found at: https://tiddlywiki5-git-fork-pmario-dumpvari-87399a-jermolenes-projects.vercel.app/#test-getvariablejson once Vercel is finished to build it

Jermolene:
Hi @pmario I haven’t commented on this PR, but I like the idea of exposing the global variables.

I don’t think it belongs in advanced search because it is _**too**_ advanced; it would only be useful to users who are hacking TiddlyWiki. At the moment, everything in advanced search is focussed on exploring the users own tiddlers, and helping users to understand the basic concepts of TiddlyWiki, particularly the distinction between ordinary tiddlers and system or shadow tiddlers. The filters tab is admittedly advanced but again it give the user the chance to get to know another fundamental feature (the canned searches make it useful even to users who haven’t reached the point of writing their own filters).

Perhaps it should be a new tab in the control panel under Info -> Advanced?

pmario:
@Jermolene --- uups, I did forget, that this branch is already a PR. I just wanted to remove the function wrappers and update the code in my origin-repo, so I can download it on my laptop.

I did not intend to trigger attention. Sorry. -- The code should be in a plugin.

PR #7995, opened 2024-02-19
Throttle autosave

saqimtiaz:
A quick take on implementing throttling of AutoSave so it does not occur any more frequently than a configured interval, to aid discussion of #6979 

SaverHandler keeps track of the timestamp when AutoSave was last initiated and any subsequent autosave requests before the specified interval are throttled.

Configuration tiddlers (optional):
- `$:/config/AutoSave/Throttle`: enables throttling of AutoSave, defaults to "no" and the throttling is disabled.
- `$:/config/AutoSave/ThrottleInterval`: specifies the minimum interval between AutoSave attempts, currently defaults to 60s if the config tiddler is missing

### DIFF
--- a/core/modules/saver-handler.js
+++ b/core/modules/saver-handler.js
@@ -12,7 +12,6 @@ The saver handler tracks changes to the store and handles saving the entire wiki
 /*global $tw: false */
 "use strict";
 
-
 /*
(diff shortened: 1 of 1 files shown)


simonbaird:
Cool. I'll aim to give it a test when I can.

PR #7996, opened 2024-02-20
MWS: Switching SQLite3 engine to Wasm instead of compiled C code

Jermolene:
# Introduction

This PR targets the [the Multi Wiki Server PR](https://github.com/Jermolene/TiddlyWiki5/pull/7915). It swaps the SQLite implementation from [better-sqlite3](https://github.com/WiseLibs/better-sqlite3) to [node-sqlite3-wasm](https://github.com/tndrle/node-sqlite3-wasm).

The difference is that better-sqlite3 consists of JavaScript bindings around a copy of the SQLite3 engine that has been compiled to native code while node-sqlite3-wasm is an adapation of [the official SQLite3 Wasm distribution](https://sqlite.org/wasm/doc/trunk/index.md) that has been modified to work on Node.js.

At the moment, this change introduces a very significant 100x performance degradation. The measured time for loading the initial bags and recipes is:

* better-sqlite3 - 0.41s
* node-sqlite3-wasm - 38.8s

The performance degradation needs to be analysed and addressed before we can consider merging this PR. One positive observation is that the performance degradation goes away when using a memory database, so there's a good chance that file I/O can be optimised further.

Note that this PR currently includes switching to a memory based database to make testing easier.

# Why switch to SQLite3 Wasm?

It fits with our long term plans of using the same SQLite3 engine both in the browser and under Node.js. The advantage is that it allows TiddlyWiki to continue to be self contained, without requiring a complex installation and build step.

More immediately, I have run into many problems getting better-sqlite3 running correctly on an Azure Web Service which are all due to binary compatibility issues. It's an area in which I have no expertise and I find myself constantly confused, particularly given the long turnaround times for making changes in cloud environments like Azure. If I am having difficulties with this then I think there's a good chance that these problems will impact ordinary users of TiddlyWiki as well. (The project where I have been exploring running TiddlyWiki on Azure can be seen at https://github.com/Jermolene/tiddlywiki-azure-ci).

In contrast, the Wasm implementation just works on Azure without any fuss and bother, and offers the promise of a single codebase for the browser and the server.

# Questions

* [ ] Can we address the performance degradation?
* [ ] Is it worth making the code switchable to work with either better-sqlite3 or node-sqlite3-wasm? There are only very minor differences between them, and MWS already has the necessary abstractions to allow the differences to be hidden away in one place

# Appreciation of node-sqlite3-wasm

I'd like to mention @tndrle to congratulate them on a nicely conceived and very neatly executed project. There was very little work involved in switching from better-sqlite3 to node-sqlite3-wasm (the commit is [here](https://github.com/Jermolene/TiddlyWiki5/commit/f7d5147efa56418e1e47aecc2ecf516002db588f)), and it's been a pleasure to work with it.

I hope that the SQLite team themselves might eventually address the Node.js gap themselves, but in the meantime @tndrle's work is much appreciated.

### DIFF
--- a/core/modules/server/server.js
+++ b/core/modules/server/server.js
@@ -84,7 +84,6 @@ function Server(options) {
 	});
 	// Load route handlers
 	$tw.modules.forEachModuleOfType("route", function(title,routeDefinition) {
-		console.log("Loading server route " + title);
 		self.addRoute(routeDefinition);
 	});
 	// Initialise the http vs https
--- a/core/modules/server/server.js
+++ b/core/modules/server/server.js
@@ -84,7 +84,7 @@ function Server(options) {
 	});
 	// Load route handlers
 	$tw.modules.forEachModuleOfType("route", function(title,routeDefinition) {
-		// console.log("Loading server route " + title);
+		console.log("Loading server route " + title);
 		self.addRoute(routeDefinition);
 	});
 	// Initialise the http vs https
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.3.4-prerelease",
       "license": "BSD",
       "dependencies": {
-        "better-sqlite3": "^9.4.2",
+        "better-sqlite3": "^9.4.3",
         "node-sqlite3-wasm": "^0.8.10"
       },
       "bin": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "5.3.4-prerelease",
       "license": "BSD",
       "dependencies": {
-        "better-sqlite3": "^9.4.1"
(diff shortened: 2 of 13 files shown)


Jermolene:
I've now added support for both better-sqlite3 and node-sqlite3-wasm at the same time. Switching between them requires a change to the default in `$:/plugins/tiddlywiki/multiwikiserver/sql-tiddler-database.js`, and is not something that I intend to offer as a feature long term.

Even with transactions, node-sqlite3-wasm is about 10 times slower than better-sqlite3, which is not encouraging. I had hoped that @hoelzro's transaction support would help but it appears not.

Anyhow, I think this is ready to merge. We can continue to experiment with the choice of engine, and explore improvements to node-sqlite3-wam

Issue #7976, opened 2024-02-09
[IDEA] TW setting items need a designated css class


twMat:
I want to style the "setting" items listed under `Ctrlpanel > Settings > TiddlyWiki` but there is no class applied to their divs, i.e [this tiddler](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FSettings%2FTiddlyWiki) generates the list like so:

```
<$list filter="[all[shadows+tiddlers]tag[$:/tags/ControlPanel/Settings]]">

<div style="border-top:1px solid #eee;">

!! <$link><$transclude field="caption"/></$link>

<$transclude/>

</div>

</$list>
```

So, instead of a proper class for each div, they're inline styled making it problematic to target.

**Describe the solution you'd like**
I am hoping a proper class can be applied instead.

**Describe alternatives you've considered**
I am currently using `div[style="border-top:1px solid #eee;"]` to target the items but this is not very good and a browser extension I use (to get browser "dark mode") interferes with this.

Thank you.

CodaCodr:
Agreed and there's quite a few more like that that could be done better. Anything that places inline style into elements forces me to use `!important` where I shouldn't need to. For a while now, I've been adding `@layer tiddlywiki-defaults` to the core stylesheet.

Alternative to your solution:

```
div[data-tiddler-title='$:/core/ui/ControlPanel/Settings/TiddlyWiki'] h2 > a {
 color:red;
}
```

![image](https://github.com/Jermolene/TiddlyWiki5/assets/436812/91e334b4-a32d-4bc3-9550-3e4a30ae6fdb)

Jermolene:
Hi @twMat @CodaCodr I would consider it a bug to have any DIVs that cannot be targeted with a reasonably straightforward selector, so I would welcome a PR to fix these ones.

pmario:
IMO something like this should work. @Jermolene: Are you OK with 2 new classes. `tc-control-panel-setting` and `tc-control-panel-accent`

I personally would like `tc-cp-setting` and `tc-cp-accent`, but there are some `tc-control-panel` definitions already. 

```
\define lingo-base() $:/language/ControlPanel/Settings/

<<lingo Hint>>

<$list filter="[all[shadows+tiddlers]tag[$:/tags/ControlPanel/Settings]]">

<div class="tc-control-panel-setting" data-tiddler-title=<<currentTiddler>> style="border-top:1px solid #eee;">

[...]

PR #8000, opened 2024-02-23
Add classes to ControlPanel setting tiddlers, to use standard CSS rules for individual styling

pmario:
This PR fixes: #7976  **[IDEA] TW setting items need a designated css class**

As discussed there there are 3 new classes and a data- attribute now. 

- tc-control-panel-setting ... for the whole section
  -  also gets a `data-setting-title` attribute with a value of `<<currentTiddler>>`
- tc-control-panel-accent ... for the h2
- tc-control-panel-item ... for the links to the setting tiddler 

With those new classes every settings element should be reachable with standard CSS rules. 

tc-control-panel-setting-01

tc-control-panel-setting-02

### DIFF
--- a/core/ui/ControlPanel/Saving/DownloadSaver.tid
+++ b/core/ui/ControlPanel/Saving/DownloadSaver.tid
@@ -2,10 +2,19 @@ title: $:/core/ui/ControlPanel/Saving/DownloadSaver
 tags: $:/tags/ControlPanel/Saving
 caption: {{$:/language/ControlPanel/Saving/DownloadSaver/Caption}}
 
+\whitespace trim
 \define lingo-base() $:/language/ControlPanel/Saving/DownloadSaver/
+
+<div class="tc-control-panel-saving" data-setting-title=<<currentTab>>>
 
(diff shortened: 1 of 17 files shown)


pmario:
If you use GitHub to view the changes, use the "Split" view: 

grafik

pmario:
@twMat ... Can you have a closer look at the preview. It should be much easier now to reach every individual element.

pmario:
This PR also contains 2 ControloPanel -> Saving tiddlers, that have the same pattern. 

They use 

- tc-control-panel-saving instead of tc-control-panel-setting

The rest is the same

twMat:
@pmario 

>@twMat ... Can you have a closer look at the preview. It should be much easier now to reach every individual element.

Sorry for my tardiness. Yes, the `data-setting-title` is great ;-)

I do note that each setting still has the inline `style="border-top:1px solid #eee;"` (which makes it impossible to override with a custom stylesheet) so I'm wondering if this should perhaps be moved into the introduced `class="tc-control-panel-setting"` ?

PR #8005, opened 2024-02-25
Update keyboard.js to make it possible to override a keyboard shortcut

BurningTreeC:
This PR makes it possible to override a keyboard shortcut like `((new-tiddler))` for example with another action in another Keyboard Shortcut tiddler with the same key descriptor if that Keyboard Shortcut tiddler has the field `override` set to `yes`

### DIFF
--- a/core/modules/keyboard.js
+++ b/core/modules/keyboard.js
@@ -321,7 +321,7 @@ KeyboardManager.prototype.updateShortcutLists = function(tiddlerList) {
 		this.shortcutActionList[i] = tiddlerFields.text;
 		this.shortcutParsedList[i] = this.shortcutKeysList[i] !== undefined ? this.parseKeyDescriptors(this.shortcutKeysList[i]) : undefined;
 		this.shortcutPriorityList[i] = tiddlerFields.priority === "yes" ? true : false;
-		if(tiddlerFields.override === "yes") {
+		if((tiddlerFields.override === "yes") && tiddlerFields["override-title"]) {
 			needsOverRide = true;
 		}
 	}
@@ -331,7 +331,7 @@ KeyboardManager.prototype.updateShortcutLists = function(tiddlerList) {
 				tiddlerFields = $tw.wiki.getTiddler(title).fields;
 			if(tiddlerFields.override === "yes") {
 				for(var k=0; k<tiddlerList.length; k++) {
-					if((k !== j) && (this.parseKeyDescriptors(this.shortcutKeysList[k]) === this.parseKeyDescriptors(this.shortcutKeysList[j]))) {
+					if((k !== j) && (tiddlerFields["override-title"] === tiddlerList[k])) {
 						this.shortcutKeysList.splice(k,1);
 						this.shortcutActionList.splice(k,1);
 						this.shortcutParsedList.splice(k,1);
--- a/core/modules/keyboard.js
+++ b/core/modules/keyboard.js
@@ -321,6 +321,21 @@ KeyboardManager.prototype.updateShortcutLists = function(tiddlerList) {
 		this.shortcutParsedList[i] = this.shortcutKeysList[i] !== undefined ? this.parseKeyDescriptors(this.shortcutKeysList[i]) : undefined;
 		this.shortcutPriorityList[i] = tiddlerFields.priority === "yes" ? true : false;
 	}
+	for(var j=0; j<tiddlerList.length; j++) {
+		var title = tiddlerList[j],
+			tiddlerFields = $tw.wiki.getTiddler(title).fields;
+		if(tiddlerFields.override === "yes") {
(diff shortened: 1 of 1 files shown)


saqimtiaz:
@BurningTreeC could you please provide an example use case that is supposed to address?

What if two shortcut tiddlers for the same descriptor have the field override set to yes? What is the expected behaviour then?

BurningTreeC:
> @BurningTreeC could you please provide an example use case that is supposed to address?

Hello @saqimtiaz , in my [TiddlyFlex Layout](https://burningtreec.github.io/TiddlyFlex/) I have keyboard shortcuts for `new-tiddler`, `new-journal` and `new-image` that should override the core keyboard shortcuts for `((new-tiddler))`, `((new-journal))` and `((new-image))`. I don't want to overwrite the core tiddlers and that's why I came up with this mechanism.
 
> What if two shortcut tiddlers for the same descriptor have the field override set to yes? What is the expected behaviour then?

One will win, I believe it's the one that's last in the filtered list. I know this isn't ideal, if you have a better idea I'm all ears :smiley_cat:

BurningTreeC:
Oh, now I figured out that I can use the cascade filter run prefix

pmario:
>I don't want to overwrite the core tiddlers and that's why I came up with this mechanism.

We made keyboard shortcuts configurable, so users can change them. If you do not want to overwrite the shortcuts from you plugin, then create docs that explains how the user can do it. If that's not good enough, then create a button, that does it for the user. So it's the users decision to adjust shortcuts and clicking 1 button should be a good user experience too.

BurningTreeC:
Can't we add a mechanism that excludes certain keyboard-shortcut tiddlers?
Like:

```
$tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcut]] :except:cascade[all[shadows+tiddlers]tag[$:/tags/KeyboardShortcutsExcludeFilter]!is[draft]get[text]] :and[has[title]]");
```

(even if my filter is not yet fully correct)

BurningTreeC:
Hello @pmario @saqimtiaz 

Now this PR overrides a keyboard shortcut only if the field `overrride` of a Keyboard Shortcut is set to `yes` AND if there's a field `override-title` and if that title matches one of the keyboard shortcut titles

pmario:
>I have keyboard shortcuts for new-tiddler, new-journal and new-image that should override the core keyboard shortcuts for ((new-tiddler)), ((new-journal)) and ((new-image)). I don't want to overwrite the core tiddlers and that's why I came up with this mechanism.

**What is the exact difference between your "new-tiddler" and the core new-tiddler keyboard shortcut?**

I think I do have an idea, but I'm not sure. 

I think the core should be able to deal with your requirements out of the box, but as I wrote somewhere else, the current "story-handling" is not really prepared to handle multiple story-rivers at the same time.

BurningTreeC:
> > I have keyboard shortcuts for new-tiddler, new-journal and new-image that should override the core keyboard shortcuts for ((new-tiddler)), ((new-journal)) and ((new-image)). I don't want to overwrite the core tiddlers and that's why I came up with this mechanism.
> 
> **What is the exact difference between your "new-tiddler" and the core new-tiddler keyboard shortcut?**
> 
> I think I do have an idea, but I'm not sure.
> 
> I think the core should be able to deal with your requirements out of the box, but as I wrote somewhere else, the current "story-handling" is not really prepared to handle multiple story-rivers at the same time.

@pmario 
Thank you, I'm very open for ideas 😄 ... at the moment I need one

pmario:
OK. That's what I thought. So that's the reason, why you need to overwrite keyboard.js and scroller.js too. right?

As I wrote in the other thread at talk. I do have some unpublished code, with some experiments. I did not publish it, because it was hacky. It used the LinkCatcher widget.

There has been no Message- or EventCatcher widgets at that time. ... But as far as I can remember I did not really need them in the end. **As far as I can remember** the navigator widget does not fully respect the `tv-story-list and tv-history-list` variables and that can cause problems with multi-stories. 

I'll have to have a closer look.

BurningTreeC:
> OK. That's what I thought. So that's the reason, why you need to overwrite keyboard.js and scroller.js too. right?

scroller.js because I need to use the javascript native `scrollIntoView()` and keyboard.js because I somehow need to overwrite some keyboard shortcuts without overwriting the core tiddlers
> 
> As I wrote in the other thread at talk. I do have some unpublished code, with some experiments. I did not publish it, because it was hacky. It used the LinkCatcher widget.
> 
> There has been no Message- or EventCatcher widgets at that time. ... But **as far as I can remember** I did not really need them in the end. The navigator widget does not fully respect the `tv-story-list and tv-history-list` variables and that can cause problems with multi-stories.
> 
> I'll have to have a closer look.

PR #8007, opened 2024-02-25
Fix "sjcl not found" errors in TiddlyDesktop

### LINKED ISSUE #8001: [BUG] sha256 filter fails on node.js
### Describe the bug

The sha256 filter contains the following code:

```js
return sjcl.codec.hex.fromBits(sjcl.hash.sha256.hash(text)).substr(0,length);
```

But on node.js (for example, when running in TiddlyDesktop with a wiki folder) there's no global sjcl object available until it's been `require()`d.

### Expected behavior

This can be solved by going into boot.js and adding `$tw.sjcl = sjcl;` right before the `this.setPassword = ...` line. Then edit core/modules/filters/crypto.js and change the two occurrences of `sjcl` to `$tw.sjcl`.

### To Reproduce

1. Install TiddlyDesktop
2. Create a new wiki folder
3. Open it
4. Hit F12 to open the developer console
5. Create a new tiddler

You'll see errors about sjcl not being found in the developer console.

### Screenshots

_No response_

### TiddlyWiki Configuration

- Version: v5.3.3
- Saving mechanism: TiddlyDesktop 0.18.0
- Plugins installed: None, default wiki folder in TiddlyDesktop

- OS: Linux
- Browser NW.js from TiddlyDesktop
- Version NW.js 0.84.0 (bug also occurs in NW.js 0.77.0 which is included in TiddlyDesktop 0.18.0)

### Additional context

I believe this may be the cause of https://github.com/TiddlyWiki/TiddlyDesktop/issues/285. I was experiencing that issue as well, but once I solved the sjcl issue, I was able to create new tiddlers just fine.

rmunn:
Fix #8001.

### DIFF
--- a/boot/boot.js
+++ b/boot/boot.js
@@ -786,6 +786,7 @@ $tw.utils.Crypto = function() {
 			}
 			return outputText;
 		};
+	$tw.sjcl = sjcl;
 	this.setPassword = function(newPassword) {
 		currentPassword = newPassword;
 		this.updateCryptoStateTiddler();
--- a/core/modules/filters/crypto.js
+++ b/core/modules/filters/crypto.js
@@ -16,7 +16,7 @@ exports.sha256 = function(source,operator,options) {
 	var results = [],
 		length = parseInt(operator.operand,10) || 20,
 		sha256 = function(text) {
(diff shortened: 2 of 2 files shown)


Jermolene:
Thanks @rmunn

PR #8008, opened 2024-02-26
Unconditionally decrement transaction depth

hoelzro:
…otherwise we may end up in a situation where we're always stuck in an "already in a transaction" state and often neglect to actually enter a real transaction!


Jermolene:
Thanks @hoelzro much appreciated.

Issue #8004, opened 2024-02-24
[IDEA] Documentation Macros - documentation should be improved


pmario:
The https://tiddlywiki.com/#Documentation%20Macros lists all the macros, but lacks example code, that is easy for users to copy / paste

![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/f433507d-0751-4801-a102-c80a2204d6ac)

pmario:
close -> PR has been merged

Issue #8003, opened 2024-02-24
[Bug] doc-macros .note, .tip and .warning inconsistencies 


pmario:
There are several inconsistencies caused by [doc macros](https://tiddlywiki.com/#%24%3A%2Feditions%2Ftw5.com%2Fdoc-macros) .note-, .tip- and .warning-macros

**Note Macro**

- The `<<.note >>` macro is implemented as a DL definition list, [which is semantically wrong](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dl). 
- `\define .note(_) @@.doc-note ;Note : $_$ @@ \end ` with some inline styling that makes it look like this:
- ![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/ff953d26-7db0-4a98-8ef1-00a34bce6ab6)
- Can be seen at: https://tiddlywiki.com/#Installing%20a%20plugin%20from%20the%20plugin%20library

**Tip Macro**

- IMO The `<<.tip>>` macro most of the time looks as follows, where the icon is much bigger than the text.
- It looks "out of order"
- ![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/814a4bd7-8927-4c93-a970-f28dd2de59c0)
- Can be seen at: https://tiddlywiki.com/#ActionLogWidget:ActionLogWidget%20%5B%5Bacos%20Operator%5D%5D

**Warning Macro**

- The `<<.warning>>` macro has a similar problem to the tip-macro
- ![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/95199871-54c8-4bc6-b6bb-e11d7947b35a)
- Can be seen at: https://tiddlywiki.com/#CheckboxWidget%20(index%20Mode):%5B%5BCheckboxWidget%20(index%20Mode)%5D%5D%20%5B%5BActionWidget%20Execution%20Modes%5D%5D

**Note and Warning together**

- IMO it does not look good
- ![grafik](https://github.com/Jermolene/TiddlyWiki5/assets/374655/684805ef-4589-4a0d-a6ad-9ecdec3d66c3)

This is an issue created to have a reference for an upcoming PR to rewrite the doc-macros with the v5.3.x syntax.

pmario:
close -> PR has been merged

PR #8009, opened 2024-02-26
Doc macros new syntax

pmario:
This PR is a draft atm, since: 

**It is proposed to move the doc-macros + their style tiddlers into a core macro**. Those macros are also handy for plugin authors. Having them as a core plugin would allow us to make reference documentation consistent for the whole ecosystem. 

@Jermolene -- What do you think about this proposal?

This PR does not contain the proposed plugin -- yet?!

-------------

This PR does convert the doc-macros into the new v5.3.x syntax. It may be possible to make some code look prettier, but I think they all still work in a backwards compatible way. 

This PR also fixes the issues

- #8003 and
- #8004

There are new "info boxes" now, which all use $:/palette colour variables to allow palette switching.

**Note boxes** look like this now

image

**Tips** look like this now

image

**Warnings** look like this now

image

**Note together with a Warning" looks like this now

image

**Documentation Macro** docs is improved

image

**Tiddler Info Area**

Tiddler info section will get a link to the https://saqimtiaz.github.io/tw5-docs-pr-maker/ edition.

image


kookma:
This is just a crude idea: 
What do you think if we could have an official plugin like author-tools, or utility, ...
and then move all related tools to it. This way we have a smaller core and a full bundle of tools for developers.

kookma:
> What else would you like to be part of an "author-tools" plugin?

I think for documentation and plugin developers we can collect useful tools and include here. Examples are boilerplate tiddlers, badges, cards, faqs, ...

I think documentation tools and utility tools can be extensively improved and yet shipped as a separate plugin!
I am always in favor of small core, even few tens of kB smaller.

pmario:
>It is proposed to move the doc-macros + their style tiddlers into a core macro. Those macros are also handy for plugin authors. Having them as a core plugin would allow us to make reference documentation consistent for the whole ecosystem.
>
>@Jermolene -- What do you think about this proposal?

@Jermolene -- any opinion?

Jermolene:
Hi @pmario it seems a reasonable idea but I do have some concerns:

* It would mean that the docs were no longer self-contained. Visiting tiddlywiki.com and exporting some or all tiddlers as JSON would produce an export that would not work on other wikis. There would be no easy way to explain to users what the problem is and how to fix it. (At the moment there is a dependency in the docs on the railroad plugin, but the consequences of it not being installed are obvious and localised; the dependency on the docs macros is much harder to manage because important functionality would fail without the docs macros - eg the examples)
* I also have plans that have come up recently on talk.tiddlywiki.org to do more work on getting the docs working properly as a plugin that can be installed alongside the users own content. In that situation, I think we'd want to have the macros integrated with the content so that it works everywhere

Perhaps instead we can offer a simple, well signposted way for users to download all the docs macros as a JSON file?

pmario:
I think your concerns are valid. 

>Perhaps instead we can offer a simple, well signposted way for users to download all the docs macros as a JSON file?

I think we can create a "Bundle doc-macros" button somewhere, which also creates a "bundle info-tiddler". The info tiddler IMO should contain the TW-version from which the bundle was exported and the date of export, including a list of all the tiddlers that have been exported. 

I'll have a closer look.

pmario:
tiddler info section will get a link to the https://saqimtiaz.github.io/tw5-docs-pr-maker/ edition.

image

pmario:
This draft will be closed, since it was replaced by PR:  **Doc macros and documentation tiddler** #8036  and  several others

pmario:
Close this one, since most of the other "smaller" PRs have been merged. see: https://github.com/Jermolene/TiddlyWiki5/pull/8037

PR #8013, opened 2024-02-26
add StoryList to gitignore

pmario:
@Jermolene 
This PR adds $__StoryList.tid to .gitignore, so we do not accidentally commit it.


Jermolene:
Thanks @pmario

PR #8016, opened 2024-02-28
Update chinese language files

BramChen:
* Add chinese captions for new emergency tiddlers download button


Jermolene:
Thanks @BramChen

PR #8021, opened 2024-03-01
MWS: store level tests

hoelzro:


Jermolene:
Looks good, thanks @hoelzro

hoelzro:
@Jermolene I just took this out of draft status - there are some more tests I'd like to add, but this covers most of the public interface for the store, so I think it's a good foundation to build from!

Jermolene:
Thank you @hoelzro

PR #8022, opened 2024-03-01
MWS: Add support for large tiddlers to be stored as attachment files

Jermolene:
Note: This PR targets the branch for [MultiWikiServer](#7915)

This PR adds support for storing large tiddlers as attachment files. There are several motivations:

* To be able to efficiently handle tiddlers larger than SQLite's [roughly 1GB limit on string size](https://www.sqlite.org/limits.html)
* To be able to efficiently stream large tiddlers directly over HTTP

# Questions

This PR includes a change to the schema. At this point, we can probably just merge it and instruct testers to delete their existing database files, if necessary using `--mws-save-archive` to first take a backup. But it is also a good opportunity to think through how we want to handle database upgrades in the future. Perhaps we should start versioning the schema and add a metadata table to the schema that includes the version number. Or maybe it would be more resilient to just make the creation of each table and field be conditional on whether it already exists.

# Progress

* [x] Extend schema to add an attachment reference to the tiddlers table
* [x] Use an attachment when saving a tiddler above a given size
* [x] Route to stream attachments back to the browser
* [x] Make the attachment size limit be configurable (currently fixed at 200KB)
* [x] Tweak design to allow ACL checks to be performed in the  `GET /attachments/:attachment_name` route. The plan is to identify attachments by the combination of bag name and title, which will allow the database to be consulted for security checks before allowing access
* [x] Streaming mutlipart form data directly to the attachment store so that we can efficiently upload large tiddlers
* [ ] Fix uploading the same file twice – in particular, might be problematic if done via two browsers that happen to send a different content type
* [ ] Avoid making text tiddlers become attachments
* [ ] Support for deleting attachments when the associated tiddler is modified or deleted
* [ ] Update `--mws-save-archive` to handle attachments
* [ ] Review error handling
* [ ] Apply length limits while processing multipart form data to prevent DDOS attacks
* [ ] Support HTTP range requests to allow streaming videos to support seeking within video


Jermolene:
These changes have now reached a stable point and so I'm going to merge into #7915, and carry across the todo items that are not complete

PR #8030, opened 2024-03-03
Doc version macros - improve code readability

pmario:
Doc version macros - improve code readability

@Jermolene This is a docs-only PR. which includes

- `<<.from-version` and
- `<<.deprecates-since` macros.

image

image


pmario:
close pr -> wrong target branch

PR #8031, opened 2024-03-03
doc-macros - variable-macros new v5.3.x syntax

pmario:
@Jermolene This is a docs-only PR. It contains 

- Variable macros syntax change to v5.3.x


Jermolene:
Thanks @pmario

PR #8032, opened 2024-03-03
docs - make .from-version and .deprecated-since macros more readable

pmario:
Doc version macros - improve code readability

@Jermolene This is a docs-only PR. which includes

- `<<.from-version` and
- `<<.deprecates-since` macros.

image

image


Jermolene:
Thanks @pmario

PR #8034, opened 2024-03-03
docs - download empty button new syntax - v5.3.x 

pmario:
@Jermolene, @saqimtiaz 

**This is a PR intended to discuss some things.** 

I want to keep stuff like this using the old syntax for the time being.

Since the `\define`s are direct preparations for core templates, I cannot predict any side effects at the moment. 

Also for "text-like" building blocks it still feels much better to use `\define savingEmpty() yes` instead of `\procedure savingEmpty() yes` ... `\procedure xx() something` just feels wrong.

### DIFF
--- a/editions/tw5.com/tiddlers/system/download-empty.tid
+++ b/editions/tw5.com/tiddlers/system/download-empty.tid
@@ -4,7 +4,7 @@ code-body: yes
 \define saveTiddlerFilter()
 [[$:/core]] [[$:/isEncrypted]] [[$:/themes/tiddlywiki/snowwhite]] [[$:/themes/tiddlywiki/vanilla]] -[[$:/boot/boot.css]] -[type[application/javascript]library[yes]] -[[$:/boot/boot.js]] -[[$:/boot/bootprefix.js]] +[sort[title]]
 \end
-\define savingEmpty()
-yes
-\end
+
+\define savingEmpty() yes
+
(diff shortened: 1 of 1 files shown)


saqimtiaz:
I am confident we can safely replace that define with a procedure as there is no textual substitution taking place. My viewpoint is that macros are due to be marked as deprecated as soon as they are no longer used by the core.

Jermolene:
Hi @pmario I agree that we should switch to a procedure; I don't see much point in this PR otherwise

pmario:
>I agree that we should switch to a procedure; I don't see much point in this PR otherwise

So you say, that we should completely remove the `\define` pragma from the core and replace it with `\procedure` even if the definitions are something like `\define savingEmpty() yes`?

As I wrote in the OP for me `\define` for usecases as this feels OK - but - `\procedure savingEmpty() yes` feels wrong.

Jermolene:
> As I wrote in the OP for me `\define` for usecases as this feels OK - but - `\procedure savingEmpty() yes` feels wrong.

Consistency is important, and so for that reason I would be inclined to make the change

`procedure` remains an unsatisfactory term for me, and I wish we had been able to come up with something better.

pmario:
@Jermolene -- I did change it as requested.

Jermolene:
Thanks @pmario

PR #8035, opened 2024-03-03
docs - operator macros, templates and filter-run template

pmario:
@Jermolene -- this is a docs only PR - It contains 3 tiddlers, that are highly connected, so they are part of 1 PR. 

This PR converts them to the new v5.3.x syntax.

They are responsible for elements as shown in the screenshot

image

PR #8033, opened 2024-03-03
docs - wikitext-macros new v5.3.x syntax

pmario:
@Jermolene This is a docs only PR:

- wikitext-macros new v5.3.x syntax

PR #8029, opened 2024-03-03
tiddler-info-source - change sources tab to new syntax and add link to pr-creator

pmario:
This PR 

- uses a prettylink to link to the tiddler at the GitHub repository. 
- adds a link to Saq Imitaz's PR-Maker edition for easier community contribution
- changes the code syntax to v5.3.x

**New**

image

**Old**

image

PR #8037, opened 2024-03-03
doc-macros, styles and documentation overview tiddler

pmario:
This PR does convert the doc-macros into the new v5.3.x syntax. It may be possible to make some code look prettier, but I think they all still work in a backwards compatible way. 

This PR also fixes the issues

- #8003 and
- #8004

There are new "info boxes" now, which all use $:/palette colour variables to allow palette switching.

**Note boxes** look like this now

image

**Tips** look like this now

image

**Warnings** look like this now

image

**Note together with a Warning" looks like this now

image

**Documentation Macro** docs is improved

image

-----

This PR replaces: **Doc macros new syntax #8009**, since it was splint into several smaller PRs.

### DIFF
--- a/editions/tw5.com/tiddlers/styleguide/Documentation Macros.tid
+++ b/editions/tw5.com/tiddlers/styleguide/Documentation Macros.tid
@@ -1,174 +1,199 @@
 created: 20150110182600000
-modified: 20230325161424684
+modified: 20240224170607731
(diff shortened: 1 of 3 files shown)


pmario:
Related PR are

- #8035
- #8034 
- #8033 
- #8032 
- #8031 
- #8029

Jermolene:
Thanks @pmario I like the visual improvements

PR #8043, opened 2024-03-08
add sha256() to $tw.utils

pmario:
The sha256 function only exists as a wikitext filter operator, but it is not available as a core utility function.

This PR fixes that problem. It adds the sha256() to $tw.utils and uses it for the sha256 filter function.

### DIFF
--- a/core/modules/filters/crypto.js
+++ b/core/modules/filters/crypto.js
@@ -14,12 +14,9 @@ Filter operators for cryptography, using the Stanford JavaScript library
 
 exports.sha256 = function(source,operator,options) {
 	var results = [],
-		length = parseInt(operator.operand,10) || 20,
-		sha256 = function(text) {
-			return $tw.sjcl.codec.hex.fromBits($tw.sjcl.hash.sha256.hash(text)).substr(0,length);
-		};
+		length = parseInt(operator.operand,10) || 20;
 	source(function(tiddler,title) {
-		results.push(sha256(title));
+		results.push($tw.utils.sha256(title,{length: length}));
 	});
 	return results;
 };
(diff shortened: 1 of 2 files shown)


linonetwo:
Hi, @pmario . Is sjcl globally available? I'm testing TidGi mobile on iOS, and I find it errors when creating tiddler: `ReferenceError: Can't find variable: sjcl`

截屏2024-03-19 18 31 56

I notice that you have refactor the usage of `sjcl` to `$tw.utils.sha256`. Have you encounter any problem using this variable? Thanks.

P.S. It has no error on Android, only on iOS.

pmario:
> notice that you have refactor the usage of sjcl to $tw.utils.sha256. Have you encounter any problem using this variable? Should we use $tw.sjcl here? Thanks.

It should work, since https://tiddlywiki.com/#sha256%20Operator%20(Examples) still work. And this filter uses the new utility function now. 

You should be able to use it the same way. $tw.utils is globally available.

pmario:
`sjcl` is globally available in TW. 

image

linonetwo:
Thank you, this maybe bug in mobile webview or TidGi, I should fix it on my side.

Jermolene:
Hi @linonetwo we've just recently fixed a longstanding bug in TiddlyDesktop whereby the "sjcl" global wasn't visible. Potentially that same bug was affecting you?

Here's the TiddlyDesktop bug report: https://github.com/TiddlyWiki/TiddlyDesktop/issues/285#issuecomment-1963182247

And here's the fix in TiddlyWiki 5: https://github.com/Jermolene/TiddlyWiki5/pull/8007

PR #8045, opened 2024-03-08
Update draggable.js - fix the call to updateDomNodeClasses

BurningTreeC:
…sses

If I'm not wrong then the "updateDomNodeClasses" is unused and the "assignDomNodeClasses" is missing

This renames "updateDomNodeClasses" to "assignDomNodeClasses"

### DIFF
--- a/core/modules/widgets/draggable.js
+++ b/core/modules/widgets/draggable.js
@@ -93,7 +93,7 @@ DraggableWidget.prototype.execute = function() {
 };
 
 
-DraggableWidget.prototype.updateDomNodeClasses = function() {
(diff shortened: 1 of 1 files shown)


pmario:
Why is this needed?

saqimtiaz:
The fix needs to be the other way around, instead of renaming the method, rename the call to the method. This is an error introduced in a recent round of widget refactoring.

BurningTreeC:
@saqimtiaz  @pmario I've now renamed the call to the method

Jermolene:
Thanks @BurningTreeC

PR #8049, opened 2024-03-09
Draggable widget: don't refresh if `enable` attr changes

BurningTreeC:
This PR changes the `draggable` widget and the `makeDraggable` util

In the `makeDraggable` util the EventHandlers are stored in variables that are passed to the widget
In the widget the EventHandlers are added or removed if the `enable` attr changes
Also, the `tc-draggable` class is added/removed and the `draggable` attribute is added/removed

### DIFF
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -25,8 +25,7 @@ widget: widget to use as the context for the filter
 exports.makeDraggable = function(options) {
 	var dragImageType = options.dragImageType || "dom",
 		dragImage,
-		domNode = options.domNode,
-		removeEventHandler = options.remove;
+		domNode = options.domNode;
 	// Make the dom node draggable (not necessary for anchor tags)
 	if(!options.selector && ((domNode.tagName || "").toLowerCase() !== "a")) {
 		domNode.setAttribute("draggable","true");
--- a/core/modules/utils/dom/dragndrop.js
+++ b/core/modules/utils/dom/dragndrop.js
@@ -25,127 +25,136 @@ widget: widget to use as the context for the filter
 exports.makeDraggable = function(options) {
 	var dragImageType = options.dragImageType || "dom",
 		dragImage,
-		domNode = options.domNode;
+		domNode = options.domNode,
+		removeEventHandler = options.remove;
 	// Make the dom node draggable (not necessary for anchor tags)
 	if(!options.selector && ((domNode.tagName || "").toLowerCase() !== "a")) {
 		domNode.setAttribute("draggable","true");
 	}
-	// Add event handlers
-	$tw.utils.addEventListeners(domNode,[
-		{name: "dragstart", handlerFunction: function(event) {
-			if(event.dataTransfer === undefined) {
-				return false;
+
+	var dragStartHandlerFunction = function(event) {
+		if(event.dataTransfer === undefined) {
+			return false;
+		}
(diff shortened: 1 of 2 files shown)


pmario:
also see: 

- #8047

BurningTreeC:
@Jermolene, this is now ready for review

core/modules/utils/dom/dragndrop.js
```diff
@@ -30,122 +30,130 @@ exports.makeDraggable = function(options) {
...
+			titles = dragTiddler ? [dragTiddler] : [],
+			startActions = options.startActions,
+			variables,
+			domNodeRect;
+		if(dragFilter) {
+			titles.push.apply(titles,options.widget.wiki.filterTiddlers(dragFilter,options.widget));
+		}
+		var titleString = $tw.utils.stringifyList(titles);
+		// Check that we've something to drag
+		if(titles.length > 0 && (options.selector && $tw.utils.domMatchesSelector(event.target,options.selector) || event.target === domNode)) {
```
pmario, line 50: If possible we should split the line to make it a bit shorter and easier to read. eg:
```
if((titles.length > 0) && 
	(options.selector && $tw.utils.domMatchesSelector(event.target,options.selector) || 
	event.target === domNode)) {
```

pmario:
@BurningTreeC -- did you test if dragging and dropping between FireFox <-> Chrome, Chrome <-> Chrome, FireFox <-> FireFox still works as expected? Putting Edge into the mix would be good too.

BurningTreeC:
> @BurningTreeC -- did you test if dragging and dropping between FireFox <-> Chrome, Chrome <-> Chrome, FireFox <-> FireFox in any direction still works as expected? Putting Edge into the mix would be good too.

Hi @pmario - in the drag&drop mechanism nothing has changed. The two functions are exactly the same as they were.
I'm using this in my TiddlyFlex Layout and can confirm that it works cross-browser

pmario:
>Hi @pmario - in the drag&drop mechanism nothing has changed. The two functions are exactly the same as they were.
I'm using this in my TiddlyFlex Layout and can confirm that it works cross-browser

That's good. 

I did just test Edge latest <-> FireFox latest and copy / pasting tiddlers, tag-pills and selected text **still works on Windows.**

core/modules/widgets/draggable.js
```diff
@@ -114,10 +126,19 @@ Selectively refreshes the widget if needed. Returns true if the widget or any of
...
-	if(changedAttributes.tag || changedAttributes.selector || changedAttributes.dragimagetype || changedAttributes.enable || changedAttributes.startactions || changedAttributes.endactions) {
+	if(changedAttributes.tag || changedAttributes.selector || changedAttributes.dragimagetype || changedAttributes.startactions || changedAttributes.endactions) {
 		this.refreshSelf();
 		return true;
 	} else {
+		if(changedAttributes.enable) {
+			this.dragEnable = this.getAttribute("enable","yes") === "yes";
+			this.makeDraggable(this.domNodes[0]);
+			if(!this.dragHandleSelector && this.dragEnable && !this.domNodes[0].classList.contains("tc-draggable")) {
+				this.domNodes[0].classList.add("tc-draggable");
```
saqimtiaz, line 137: We have a utility function for toggling classes that would be useful to use here. Also, it might be more robust for the widget to track its own internal state (enabled/disabled) rather than check the classes of the DOM node to ascertain this.

PR #8050, opened 2024-03-10
Let caption be displayed as main title

linonetwo:
Isn't this already the de facto standard? Many plugins in distributions patch this, and users using vanilla HTML ask how to do this.

Code is copy from https://talk.tiddlywiki.org/t/should-caption-be-displayed-as-main-title/4399/8?u=linonetwo

### DIFF
--- a/core/ui/ViewTemplate/title/default.tid
+++ b/core/ui/ViewTemplate/title/default.tid
@@ -2,5 +2,7 @@ title: $:/core/ui/ViewTemplate/title/default
 
 \whitespace trim
 <h2 class="tc-title">
-<$view field="title"/>
+  <$transclude field="caption">
+    <$view field="title"/>
+  </$transclude>
 </h2>


oeyoews:
In fact, doing so may cause confusion to users, such as jumping between tiddlers, and the title that users see is not a real title.

pmario:
I'm not happy about that one. The `caption` field was introduced to create **short** titles for ControlPanel tabs. So for me a caption has to be **short**

I think it should be "subtitle" -> "caption" -> "tiddler title". 

For me it's OK, if the TOC shows the caption, without any visual indication, that it's the caption-field instead of the tiddler title. 

BUT for the tiddler title I do want to have a visual clue, that what I'm seeing is not the title but the "subtitle" in my case.

pmario:
The other point is my tiddler titles look like this: `detect-if-assistive-tech-is-active`. My tiddlers have captions, if I need to show them in tabs. **But in no way** I want to see those captions as a tiddler title.

So from my point of view there needs to be an "opt in" configuration to change that behaviour.

Jermolene:
Hi @linonetwo while this is a useful customisation for some wikis, it is not universally applicable. It is like to introduce confusion for some users (eg it would make it possible to see two different tiddlers in the story river with the same apparent title).

The purpose of the caption field is to provide an abbreviated, non-unique name for a tiddler that is shown in settings where the full title is not needed (eg within table of contents).

linonetwo:
I understand your concern, I will wrap this as a plugin instead, and call the new procedure from the core (described in https://github.com/Jermolene/TiddlyWiki5/pull/8051#issuecomment-1987112424 ).

PR #8051, opened 2024-03-10
Use caption instead of title in the TOC, Sidebar All and Sidebar Recent

linonetwo:
Another PR to make `caption` field a standard field for display. So community plugins won't be diverse in the field name.

code is copy from https://talk.tiddlywiki.org/t/how-to-use-caption-instead-of-title-in-the-toc-sidebar-all-and-sidebar-recent/6072/7?u=linonetwo


linonetwo:
There can also be a procedure shared between this and https://github.com/Jermolene/TiddlyWiki5/pull/8050

pmario:
I personally would be OK with that change.

linonetwo:
Let me add that procedure instead, and use it in the TOC, closing this.

PR #8054, opened 2024-03-10
add code-body field to doc-styles tiddler

pmario:
This PR is docs only and adds the `code-body: yes`  to the doc-styles tiddler for better human readability. It also shows which macros are used, which is important.


Jermolene:
Thanks @pmario

PR #8055, opened 2024-03-10
deprecated the if-macro, since it is not needed anymore. Use <% if %> instead

pmario:
This PR is docs only. It add the $:/deprecated tag to the if-macro. In the core docs this macro should already be replaced by `<% if %>`.

But users may still use it in their docs. So it should stay in the docs for a little bit longer.

@jermolene -- Should we completely remove it already?

### DIFF
--- a/editions/tw5.com/tiddlers/system/if-macro.js
+++ b/editions/tw5.com/tiddlers/system/if-macro.js
@@ -1,7 +1,12 @@
 /*\
 title: $:/editions/tw5.com/if-macro.js
 type: application/javascript
+tags: $:/deprecated
 module-type: macro
+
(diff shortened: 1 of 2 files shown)


Jermolene:
Hi @pmario I'd prefer to completely remove `$:/editions/tw5.com/if-macro.js`

pmario:
I did a second check. -- Both the editions/fr-FR and the editions/dev (the core dev edition) seem to use the .if-macro. 

I'll create an "Export doc macros" button, which should make it easy to update both the dev and the French edition. 

Both editions did copy the core $:/editions/tw5.com/doc-macros and some other doc-tiddlers into their edition, but forgot about the if-macro. I'm not sure if it ever worked their, or if it was needed.

PR #8056, opened 2024-03-10
some more tv-macros changed to new syntax. Plus adjusted doc tiddlers

pmario:
This PR is docs only. It changes some more tv-macros to new syntax and adjusts the corresponding doc tiddlers if needed.


Jermolene:
Thanks @pmario

PR #8057, opened 2024-03-10
some minor adjsutemens to exsisting doc tiddlers without changing modified date

pmario:
This PR is docs only. It contains some minor changes for some "variable" and "concatination" related tiddlers. It's a mini preparation for bigger changes, which IMO are needed in the future


Jermolene:
Thanks @pmario

PR #8062, opened 2024-03-11
deprecate Macro call railroad syntax

pmario:
This PR adds the $:/deprecated tag to the Macro call syntax description


Jermolene:
Thanks @pmario

PR #8067, opened 2024-03-12
Improve filter run prefix docs

mateuszwilczek:
* Add missing `:flat` suffix to the railroad diagram of `:map` filter run prefix
* Add missing "empty" side rail (indicating that the entire prefix is optional, as in other prefixes) to `:sort` filter run prefix

This was inspired by this [recent discussion](https://talk.tiddlywiki.org/t/plea-for-a-new-as-filter-run-prefix/9313) and by me myself not being able to easily find about the flat suffix not so long ago. 

---


Jermolene:
Thanks @mateuszwilczek

PR #8071, opened 2024-03-15
fix wikitext-macros example block mode

pmario:
This PR fixes a problem with the macro: `\procedure wikitext-example-without-html(src)`, which does not use block-mode anymore.

 

---


pmario:
Before: https://tiddlywiki.com/#WidgetMessage%3A%20tm-modal

image


After: 

image

Jermolene:
Thanks @pmario, a worthwhile improvement

PR #8075, opened 2024-03-16
make github-fork-ribbon usable again

### LINKED ISSUE #8074: [Bug] The github-fork-ribbon has a lot of flaws, which makes it unusable
While updating the sidebar-tools bar code I saw, that it contains a hardcoded STYLE tag, [which was introduced ](https://github.com/Jermolene/TiddlyWiki5/commit/06cbc45dfa1d1fe519e8de179ecf828133ee65f8 )because the github-fork-ribbon stylesheet "bleeds" into the main UI. See old issue: https://github.com/Jermolene/TiddlyWiki5/issues/1031

1. IMO that's a problem. Plugin styles should not introduce side effects like this one. 
2. If the ribbon is set to the left it looks like in the screenshot below
3. If tested on tiddlywik.com it is partially hidden below the top menu

image

4. The descriptions says: "The ribbon can be positioned over any corner, and can incorporate user defined text, colours and a link."
5. There is **no** convenient way to configure the color, the URL or the text. 

PR is on the way.

pmario:
This PR fixes #8074 **github-fork-ribbon**

* The ribbon is a tiddler eg: $:/github-ribbon tagged: `$:/tags/PageTemplate` now. 
* Styles have been transcluded into the template.
* The creation of the tiddler and toggling the tag is interactive. So the UX should be much easier now. 
* All parameters are described
* It got a new "**usage**" page

**Plugin to test with**

Since the CI build does not contain the plugin I did add it to this PR as a ZIP file. 

* [$ _plugins_tiddlywiki_github-fork-ribbon.zip](https://github.com/Jermolene/TiddlyWiki5/files/14624220/_plugins_tiddlywiki_github-fork-ribbon.zip)

The screenshot has 2 banners, to show that the "left" one is at the right position now.

image

### DIFF
--- a/plugins/tiddlywiki/github-fork-ribbon/readme.tid
+++ b/plugins/tiddlywiki/github-fork-ribbon/readme.tid
@@ -4,8 +4,6 @@ This plugin provides a diagonal ribbon across the corner of the window. It resem
 
 The ribbon can be positioned over any corner, and can incorporate user defined text, colours and a link.
 
-The CSS stylesheet is adapted from work by Simon Whitaker:
+The CSS stylesheet is adapted from work by [[Simon Whitaker|https://github.com/simonwhitaker/github-fork-ribbon-css/]]
 
-https://github.com/simonwhitaker/github-fork-ribbon-css/
-
-[[Source code|https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/github-fork-ribbon]]
+[[Plugin source code|https://github.com/Jermolene/TiddlyWiki5/blob/master/plugins/tiddlywiki/github-fork-ribbon]]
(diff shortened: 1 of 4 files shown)


Telumire:
The rewrite doesn't solve https://github.com/Jermolene/TiddlyWiki5/pull/8096, so I'll keep this PR open for now

image

Telumire:
Can we use the :has() selector to target any container containing a ribbon and apply isolation:isolate on it ?

pmario:
IMO the core should only use `:has()` iff there is a working fallback. The :has CSS function is way to new.

There is some discussion going on to use ECMAscript standard 2018 or now 2019 for the TW core. We should use the same "state" for CSS features.

pmario:
>The rewrite doesn't solve https://github.com/Jermolene/TiddlyWiki5/pull/8096, so I'll keep this PR open for now

Yes. sorry I did misread the image and did confuse it with the github-fork-ribbon.

pmario:
@Jermolene -- This is a core plugin, that is rarely used and it forced a "hardcoded style" fix to the core UI, which is a problem for the core UI in general. The problem caused by the fix is always there, even if the plugin is not used at all. 

If this PR is merged, I'm able to remove the hardcoded style from the core UI, because the new plugin should behave much friendlier.

Jermolene:
Thanks @pmario

PR #8078, opened 2024-03-16
Sidebar-Tools tab new syntax and fix vertical spacing in story river

### LINKED ISSUE #8076: [Bug] Sidebar Tools tab opened in the story-river looks different than in the sidebar
- Sidebar Tools tab opened in the story-river looks different than in the sidebar
  - The vertical spacing is different
- All the other tabs are already fixed

image

PR is on the way

pmario:
This PR

- fixes #8076 
- adds `tc-sidebar-tools-item` class to the list-items
- adds  `data-title=<<listItem>>` to every list-item
   - This will improve UX for users who want to "hide" toolbar elements for "read only" modes, since they can use plain tiddler titles for styling
   - This will allow us to simplify the core "read only" theme some day in the future

After: with PR active

image

Before: 

image

### DIFF
--- a/core/ui/SideBar/Tools.tid
+++ b/core/ui/SideBar/Tools.tid
@@ -2,31 +2,26 @@ title: $:/core/ui/SideBar/Tools
 tags: $:/tags/SideBar
 caption: {{$:/language/SideBar/Tools/Caption}}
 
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @pmario

PR #8079, opened 2024-03-16
Fix indentation for tiddlers that set tv-config-toolbar-class

pmario:
This PR makes the code human readable for tiddlers, that do **define the variable ** tv-config-toolbar-class with an uriencoded listItem name. 

This PR is a preparation to add `data-title=<<listItem>>` to buttons for better UX defining a "read only" theme. The `data-title` attribute allows us to use plain text titles instead of uri-encoded classes, which nobody can read. 

-	modified:   core/ui/EditTemplate/controls.tid
-	modified:   core/ui/PageControls.tid
-	modified:   core/ui/PageControls/more-page-actions.tid
-	modified:   core/ui/ViewTemplate/title.tid
-	modified:   core/ui/ViewToolbar/more-tiddler-actions.tid
-	modified:   plugins/tiddlywiki/menubar/items/pagecontrols.tid

### DIFF
--- a/core/ui/EditTemplate/controls.tid
+++ b/core/ui/EditTemplate/controls.tid
@@ -1,12 +1,18 @@
 title: $:/core/ui/EditTemplate/controls
 tags: $:/tags/EditTemplate
 
-\define config-title()
-$:/config/EditToolbarButtons/Visibility/$(listItem)$
-\end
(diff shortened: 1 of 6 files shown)


pmario:
- This PR should only change indentation.
- No new code is added. 
- It only contains tiddlers, which already had `\whitespace trim` set. So the DOM should not be changed

Jermolene:
Thanks @pmario

PR #8082, opened 2024-03-17
fix hide-show button code needs to be transcluded mode=block

pmario:
@Jermolene -- this PR is docs only. 

It contains 1 more fix, where the example needs to be transcluded using $mode=block.


pmario:
@Jermolene -- docs only - please review

Jermolene:
Thanks @pmario

Issue #8069, opened 2024-03-13
[BUG] Non-system library tiddlers get deleted


flibbles:
### Describe the bug

Library tiddlers (as in tiddlers that have `library=yes`, like sjcl.js), are excluded from being written to the general tiddler store because the saveTiddlerFilter has `-[type[application/javascript]library[yes]]` in it. Instead, they are written into the special Library Modules block that gets executed immediately upon load. Its transclude filter uses `[is[system]type[application/javascript]library[yes]]`.

See the difference? `[is[system]]`. That means nonsystem library modules don't end up existing anywhere. If you make one, it will be gone after you save like any `$:/temp` tiddler would be.

This is super minor. I know I'm the only person to notice this. I only did because I'm looking into using library modules for Uglify as a means to more consistently alter the `//# sourceMappingURL=` directive that gets attached to modules, and thus allow for better sourceMapping.

I actually **do** want the non-persistent library tiddler feature, so I might make use of this, except it looks like a bug which might get fixed out from under me. If this is an oversight, I can fix it and instead use `$:/temp` directly.

### TiddlyWiki Configuration

All Tiddlywikis

pmario:
>I actually do want the non-persistent library tiddler feature, so I might make use of this, except it looks like a bug which might get fixed out from under me. If this is an oversight, I can fix it and instead use $:/temp directly.

hmmm, I would go with $:/temp, if it does not cause any problems.

I think the 2 filters you listed, actually are an "inconsistency". We did not have any "bug reports" I can remember, where users actually lost data.

Jermolene:
Thanks @flibbles, that is indeed a bug. This feature has been around since the very beginning, but has never been widely tested apart from the usage in the core. I do think it would be worth fixing for consistency and clarity if nothing else.

flibbles:
Sure, I can create a PR. The way I'd fix it is to make the `[is[system]]` consistent. These are my reasons:

* If someone knows enough about TiddlyWiki that they're designing their own boot libraries, then they're already familiar with system tiddlers.
* Also, having a non-system boot module would make it more visible to end-users, and they absolutely should not be edited in the browser. It could cause the wiki to stop loading.
* I can't imagine why, but it seems possible that some user somewhere might have a "javascript snippet collection" tiddlywiki, and if they use `library` as a field, and if they ever put it as a 'yes', they'll brick their tiddlywiki.
[...]

PR #8083, opened 2024-03-18
Made library boot module requirements consistent

flibbles:
Here is the minor fix for #8069. Tiddlers with `[library[yes]]` only get treated specially if they're system tiddlers.


joshuafontany:
@flibbles Nice catch.

Jermolene:
Thanks @flibbles, and apologies, I thought I had merged this.

PR #8084, opened 2024-03-18
add tag parameter to reveal-widget to fix info-area regression

pmario:
Fix problem introduced with #8079 

PR applied

image

Before: 

image

@BurningTreeC Thx for the "catch"


Jermolene:
Thanks @pmario

PR #8086, opened 2024-03-18
make $:/core/ui/PageTemplate/sidebar human readable

pmario:
This PR prepares $:/core/ui/PageTemplate/sidebar for new v5.3.x syntax

- This PR adds indentation
- It also **adds** a `tag="div"` to the first reveal-widget so the DOM stays the same


Jermolene:
Hi @pmario this change is not backwards compatible. Inspecting the DOM output, we've lost the `<p>` tag that should be the first child of `<div class="tc-dropzone tc-page-container-inner">`.

If that change was intentional then this PR should have made that clear.

pmario:
>Hi @pmario this change is not backwards compatible. Inspecting the DOM output, we've lost the <p> tag that should be the first child of <div class="tc-dropzone tc-page-container-inner">.

There is no `<div class="tc-dropzone tc-page-container-inner">` code in this PR. 

I did have a look at TW v5.3.3 down to v5.2.7 None of them has a P as the first child of this DIV.

Jermolene:
Apologies @pmario I was mistaken. Looks good to go

pmario:
> Apologies @pmario I was mistaken. Looks good to go

Hmm, The funny thing is, that this PR would have created an extra P tag, that would have caused formatting trouble, with the tiddler title and wiki title baseline. 

It seems we need the extra new-lines, which points out a fundamental "hiccup" in our parser. I think I know where it comes from, but I'm not 100% sure yet.

Jermolene:
Apologies @pmario somehow I've messed up this PR and now there are a lot of extraneous files in it. I guess it happened because of the force-push I had to do earlier. Let me know if you're able to fix it

pmario:
I think we can skip this one, since we needed to keep the extra new-lines. So we should touch this file again when we do move to TW v5.3.x syntax. -- And when we know how to avoid the extra P tag when the new-lines are removed. -- I'll create a new issue, so we do not forget about it.

PR #8087, opened 2024-03-18
New v5.3.x syntax for page template sidebar

pmario:
This PR depends on, so the other one should be merged first. (I hope GitHub allows it that way)

- #8086

It changes

- `\define` -> `\function`
- changes transclusion to $params
- removes a redundant DIV since the scrollable-widget adds it's own wrapper DIV.

### DIFF
--- a/core/ui/PageTemplate/sidebar.tid
+++ b/core/ui/PageTemplate/sidebar.tid
@@ -2,16 +2,18 @@ title: $:/core/ui/PageTemplate/sidebar
 tags: $:/tags/PageTemplate
 
 \whitespace trim
-\define config-title() $:/config/SideBarSegments/Visibility/$(listItem)$
+\function config-title() [[$:/config/SideBarSegments/Visibility/$(listItem)$]substitute[]]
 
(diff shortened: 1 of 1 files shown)


core/ui/PageTemplate/sidebar.tid
```diff
@@ -2,28 +2,18 @@ title: $:/core/ui/PageTemplate/sidebar
 tags: $:/tags/PageTemplate
 
 \whitespace trim
-\define config-title()
-$:/config/SideBarSegments/Visibility/$(listItem)$
-\end
+\function config-title() [[$:/config/SideBarSegments/Visibility/$(listItem)$]substitute[]]
```
Leilei332, line 5: For local macros like these, it may be better to use [substituted attribute values](https://tiddlywiki.com/#Substituted%20Attribute%20Values)

pmario: That's exactly, what this **function** does. But the main problem here is, that the new code adds even more redundant P tags, which I do **not want**

PR #8088, opened 2024-03-18
use v5.3.x syntax for $:/core/ui/PageTemplate/pagecontrols

pmario:
This PR 

- changes `\define config-title` to a `\function`
- it removes a redundant set-widget
   - The substitution `\function` works as the screenshot below shows
   - Enabling / disabling element from the sidebar -> Tools tab also works
- transclude uses $params
- Stylesheet
  - The `tc-page-controls` DIV currently is covered by an HTML P-tag, which has `margin-top` and `margin-bottom` set to 14px. 
  - `tc-page-controls` CSS currently **only** has `margin-top`: 14px -> This PR fixes that problem (for the future)

image

### DIFF
--- a/core/ui/PageControls.tid
+++ b/core/ui/PageControls.tid
@@ -1,16 +1,14 @@
 title: $:/core/ui/PageTemplate/pagecontrols
 
 \whitespace trim
-\define config-title() $:/config/PageControlButtons/Visibility/$(listItem)$
+\function config-title() [[$:/config/PageControlButtons/Visibility/$(listItem)$]substitute[]]
 
 <div class="tc-page-controls">
 	<$list filter="[all[shadows+tiddlers]tag[$:/tags/PageControls]!has[draft.of]]" variable="listItem">
-		<$set name="hidden" value=<<config-title>>>
-			<$list filter="[<hidden>!text[hide]]" storyview="pop" variable="ignore">
-				<$set name="tv-config-toolbar-class" filter="[<tv-config-toolbar-class>] [<listItem>encodeuricomponent[]addprefix[tc-btn-]]">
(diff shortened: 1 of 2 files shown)


pmario:
@Jermolene ... please review

Jermolene:
Thanks @pmario

yaisog:
Why didn't you use
```
\function config-value() [<listItem>lookup[$:/config/PageControlButtons/Visibility/]]
```
and then
```
<$list filter="[<config-value>!match[hide]]"
```
instead of `substitute`? That only takes 2/3 of time (though both are pretty fast).
I think this is exactly the kind of operation that the `lookup` operator was made for.

pmario:
>Why didn't you use ...

Because it would remove the `config-title` variable from the widget tree. There is a transclusion `<$transclude $tiddler=<<listItem>> $mode="inline"/>`

The `<<listItem>>` tiddler contains a template, that may or may not use `config-title` for something. Keeping the config-title variable has less risk to break 3rd party stuff. 

I did make "simple" changes for the existing UI and some of them "exploded" directly in front of my face.

----------

On the other hand with v5.4.0 we would expect / tolerate backwards compatibility breaks. So there would be a chance to completely remove the function definition and use `<%if filter...%>` instead of `<$list filter...`.

-------

We would need to check some popular 3rd party plugins, that add page controls, if they still work. So you would need to check Eric Shulmans and Mohammads libraries. Even back to some "exotic" plugins from Tobias Beer, which are super old, super brittle ... but still popular. 

We could only get rid of them, if we would have modern replacements for some of them.

yaisog:
> Because it would remove the `config-title` variable from the widget tree.

That would indeed be a good reason. 👍

PR #8095, opened 2024-03-20
Fix: Set a proper doctype for the open window template (chromium bug fix) 

Telumire:
When DOCTYPE is not set, on chromium based browsers, css operator like :has() behave unexpectedly:

See https://talk.tiddlywiki.org/t/bug-has-hover-not-working-in-single-tindler-window/9379

Snippet to reproduce the bug (solved by this PR):

```
<article>
<span>if this text is green when hovered, :has work</span>
</article>

<pre><style contenteditable style.display="block">article{
color: red;
&:has(:hover){
color: green;
}
}</style></pre>
```

### DIFF
--- a/core/modules/startup/windows.js
+++ b/core/modules/startup/windows.js
@@ -56,7 +56,7 @@ exports.startup = function() {
 			return;
 		}
 		// Initialise the document
-		srcDocument.write("<html><head></head><body class='tc-body tc-single-tiddler-window'></body></html>");
+		srcDocument.write("<!DOCTYPE html><head></head><body class='tc-body tc-single-tiddler-window'></body></html>");
 		srcDocument.close();
 		srcDocument.title = windowTitle;
 		srcWindow.addEventListener("beforeunload",function(event) {


pmario:
Did you try it with the prerelease: https://tiddlywiki.com/prerelease/

Telumire:
> Did you try it with the prerelease: https://tiddlywiki.com/prerelease/ -- I think it's already fixed and should come with the next release

Yes, the prerelease currently do not fix the bug:

bughas

pmario:
Prerelease works for me with FF. -- Make sure that your browser does show the latest version of the wiki. I do a SHIFT-reload, to make sure, that I do not see a cached version. 

image

Telumire:
> Prerelease works for me with FF. -- Make sure that your browser does show the latest version of the wiki. I do a SHIFT-reload, to make sure, that I do not see a cached version. 
> 
> image
> 

This bug is appearing on chromium based browsers, not Firefox :)

Telumire:
More info on this bug: https://developer.chrome.com/docs/lighthouse/best-practices/doctype

When the doctype is missing, chrome triggers the "quirks mode"

Jermolene:
Thanks @Telumire, that's great.

PR #8097, opened 2024-03-21
fix: import lingo macro when render title

linonetwo:
Tested locally

Before:

图片

After:

图片

### DIFF
--- a/core/modules/startup/render.js
+++ b/core/modules/startup/render.js
@@ -29,7 +29,11 @@ var THROTTLE_REFRESH_TIMEOUT = 400;
 
 exports.startup = function() {
 	// Set up the title
-	$tw.titleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_TITLE_TITLE,{document: $tw.fakeDocument, parseAsInline: true});
+	$tw.titleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_TITLE_TITLE, {
+		document: $tw.fakeDocument,
+		parseAsInline: true,
+		importPageMacros: true,
+	});
 	$tw.titleContainer = $tw.fakeDocument.createElement("div");
 	$tw.titleWidgetNode.render($tw.titleContainer,null);
 	document.title = $tw.titleContainer.textContent;


core/modules/startup/render.js
```diff
@@ -29,7 +29,11 @@ var THROTTLE_REFRESH_TIMEOUT = 400;
 
 exports.startup = function() {
 	// Set up the title
-	$tw.titleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_TITLE_TITLE,{document: $tw.fakeDocument, parseAsInline: true});
+	$tw.titleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_TITLE_TITLE, {
+		document: $tw.fakeDocument,
+		parseAsInline: true,
+		importPageMacros: true,
```
pmario, line 35: hmmm -- Do we really need the whole PageMacro content? I think there should be a more specific tag, to avoid the memory overhead, just to populate browser tabs once.

linonetwo: I know, just use `, importVariables: "[[$:/core/macros/lingo]]"`, it accept a filter.

pmario: Thanks, I think that makes sense.

linonetwo:
Simillarly, when you want to render field that may contains `lingo` procedure call:

```ts
export function renderTextWithCache(text: unknown, widget: Widget | undefined, variables?: Record<string, string>) {
  if (text === undefined || typeof text !== 'string') return '';
  return $tw.wiki.getGlobalCache(
    `wikify-${text}`,
    () => $tw.wiki.renderText('text/plain', 'text/vnd.tiddlywiki', `\\import [[$:/core/macros/lingo]]\n\n${text}`, { parentWidget: widget, variables }),
  );
}
```

core/modules/startup/render.js
Jermolene: Actually I think it makes more sense to include all the page macros. Otherwise it's confusing to users why some macros are available and not others.

Jermolene:
Hi @linonetwo I'm ready to merge this if you could kindly address https://github.com/Jermolene/TiddlyWiki5/pull/8097#discussion_r1625783649

core/modules/startup/render.js
linonetwo: @Jermolene Will number of imported macro/variables impact performance significantly?

linonetwo:
@Jermolene Updated

Jermolene:
Thanks @linonetwo

PR #8099, opened 2024-03-21
Fix/sjcl variable

linonetwo:
https://github.com/Jermolene/TiddlyWiki5/pull/8043#issuecomment-2009207102 does not fix my problem, I get error `Can't create duplicate variable: 'A'` on iOS webview on TidGi mobile when init TW.

And I find sjcl.js 's internal `this.A=xxx` leak to window, where `this` is `window`. And later it does `function A(){xxx}`, and cause error on iOS. See this on tw5-com

图片

In this PR I wrap IIFE to sjcl.js by using `tiddlywiki.files`. (I have to move sjcl.js to `/lib`, otherwise `tiddlywiki.files` will prevent  `boot.js` from loading, don't know why.)

I also change sha256 to use `$tw.sjcl`, but this is not necessary, because we have to mount sjcl to `window.sjcl` before booting, when `$tw` is not init yet. Unless we can init `$tw = {}` before library.


flibbles:
@linonetwo, I've committed a programmer sin; I [just opened a PR](https://github.com/Jermolene/TiddlyWiki5/pull/8196) for this fix without checking that anyone else already had. *I just assumed that nobody else would have noticed this problem than me, the Uglify maker.*

I actually think your fix is better, so I'll close my PR and vouch for yours.

Jermolene:
Thanks @linonetwo I think it would be better for backwards compatibility to move sjcl.js back into `/boot`. To avoid the problem you ran into we just need to explicitly list boot.css, boot.js and bootprefix.js in `tiddlywiki.files`

linonetwo:
I was about to write tiddlywiki galgame plugin today, but I think about it, if this PR only takes 5 min to refactor, then I should tend to this.

Jermolene:
Thanks @linonetwo much appreciated

pmario:
@Jermolene -- This PR breaks the RSOD popup.

I did test the following code with the latest master and it broke the popup See: screenshot

```

```

image

linonetwo:
截屏2024-05-27 18 15 47

Looks like some core css (.tc-error-form) is missing, maybe is caused by tiddlywiki.files added in last commit

linonetwo:
I forget to add style system tag to it, let me PR a quick fix.

PR #8106, opened 2024-03-26
Stylesheets: Use a single `<style>` tag for each

BurningTreeC:
This PR is related to the discussion #8105

It makes each stylesheet tiddler become a single `<style>` tag
If a stylesheet tiddler changes or a stylesheet gets added/removed all style tags get removed and readded. This shouldn't be problematic as that won't happen often.

### DIFF
--- a/core/modules/startup/render.js
+++ b/core/modules/startup/render.js
@@ -38,20 +38,62 @@ exports.startup = function() {
 			document.title = $tw.titleContainer.textContent;
 		}
 	});
-	// Set up the styles
-	$tw.styleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_STYLESHEET_TITLE,{document: $tw.fakeDocument});
-	$tw.styleContainer = $tw.fakeDocument.createElement("style");
-	$tw.styleWidgetNode.render($tw.styleContainer,null);
-	$tw.styleWidgetNode.assignedStyles = $tw.styleContainer.textContent;
-	$tw.styleElement = document.createElement("style");
-	$tw.styleElement.innerHTML = $tw.styleWidgetNode.assignedStyles;
-	document.head.insertBefore($tw.styleElement,document.head.firstChild);
+
+	function setStylesheets() {
+		// Set up the styles for each stylesheet Tiddler
+		for(var i=0; i<$tw.stylesheetTiddlers.length; i++) {
+			var styleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_STYLESHEET_TITLE,{document: $tw.fakeDocument, variables: {
+				stylesheet: $tw.wiki.getTiddlerText($tw.stylesheetTiddlers[i])
+			}});
+			$tw.styleWidgetNodes.push(styleWidgetNode);
+			var styleContainer = $tw.fakeDocument.createElement("style");
+			$tw.styleContainers.push(styleContainer);
+			styleWidgetNode.render(styleContainer,null);
+			styleWidgetNode.assignedStyles = styleContainer.textContent;
+			var styleElement = document.createElement("style");
+			$tw.styleElements.push(styleElement);
+			styleElement.innerHTML = styleWidgetNode.assignedStyles;
+			document.head.insertBefore(styleElement,document.head.firstChild);
+		}
+	}
+
+	function getStylesheets() {
+		// Get our stylesheets in reversed order
+		return $tw.wiki.filterTiddlers("[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!has[draft.of]reverse[]]");
+	}
+
+	// Get our stylesheets
+	$tw.stylesheetTiddlers = getStylesheets();
+	$tw.styleWidgetNodes = [];
+	$tw.styleContainers = [];
+	$tw.styleElements = [];
+
+	setStylesheets();
+
+	// Listen for changes: Added / removed stylesheet Tiddlers, changed stylesheet Tiddlers
+	$tw.wiki.addEventListener("change",function(changes) {
+		var stylesheetTiddlers = getStylesheets();
+		if(stylesheetTiddlers.length !== $tw.stylesheetTiddlers.length || $tw.utils.hopArray(changes,stylesheetTiddlers)) {
+			for(var i=0; i<$tw.stylesheetTiddlers.length; i++) {
+				document.head.removeChild($tw.styleElements[i]);
+			}
+			$tw.stylesheetTiddlers = stylesheetTiddlers;
+			$tw.styleWidgetNodes = [];
+			$tw.styleContainers = [];
+			$tw.styleElements = [];
+
+			setStylesheets();
+		}
+	});
+
 	$tw.wiki.addEventListener("change",$tw.perf.report("styleRefresh",function(changes) {
-		if($tw.styleWidgetNode.refresh(changes,$tw.styleContainer,null)) {
-			var newStyles = $tw.styleContainer.textContent;
-			if(newStyles !== $tw.styleWidgetNode.assignedStyles) {
-				$tw.styleWidgetNode.assignedStyles = newStyles;
-				$tw.styleElement.innerHTML = $tw.styleWidgetNode.assignedStyles;
+		for(var i=0; i<$tw.stylesheetTiddlers.length; i++) {
+			if($tw.styleWidgetNodes[i].refresh(changes,$tw.styleContainers[i],null)) {
+				var newStyles = $tw.styleContainers[i].textContent;
+				if(newStyles !== $tw.styleWidgetNodes[i].assignedStyles) {
+					$tw.styleWidgetNodes[i].assignedStyles = newStyles;
+					$tw.styleElements[i].innerHTML = $tw.styleWidgetNodes[i].assignedStyles;
+				}
 			}
 		}
 	}));
--- a/core/ui/PageStylesheet.tid
+++ b/core/ui/PageStylesheet.tid
@@ -8,9 +8,7 @@ code-body: yes
 
 <$set name="languageTitle" value={{!!name}}>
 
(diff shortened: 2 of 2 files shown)


BurningTreeC:
P.S. for window.js this is missing, so don't merge yet

BurningTreeC:
Now this PR is ready @Jermolene @pmario 
I will try splitting up the base stylesheet in a different PR if it's worthwhile

pmario:
With these changes my styleRefresh times go up from 3-6ms to 5-13ms - I do click right sidebar Open - Contents - Open - Contents and so on.

pmario:
--------- OT -------

@BurningTreeC .. You have a lot of "stalled" / unused branches in your repository. -- If I want to  switch to one of your branches I can see all of them. It would be nice if you **remove your closed and merged** branches. It will make testing your PRs much easier.

<details><summary>Click to show screenshot</summary>

grafik

</details>

Your branches can be managed at: https://github.com/BurningTreeC/TiddlyWiki5/branches

BurningTreeC:
> With these changes my styleRefresh times go up from 3-5ms to 5-7ms - I do click right sidebar Open - Contents - Open - Contents and so on.

Hi @pmario I will also test

BurningTreeC:
> --------- OT -------
> 
> @BurningTreeC .. You have a lot of "stalled" / unused branches in your repository. -- If I want to switch to one of your branches I can see all of them. It would be nice if you **remove your closed and merged** branches. It will make testing your PRs much easier.
> Click to show screenshot
> 
> Your branches can be managed at: https://github.com/BurningTreeC/TiddlyWiki5/branches

Yeah I know 😊 I will try cleaning up

CodaCodr:
@BurningTreeC  FWIW, this is how I code my css. Works fine in the preview 👍 

image

BurningTreeC:
> With these changes my styleRefresh times go up from 3-5ms to 5-7ms - I do click right sidebar Open - Contents - Open - Contents and so on.

Hi @pmario, can you test another time and maybe something that also changes a stylesheet, like closing/opening the sidebar?

BurningTreeC:
Hi @Jermolene @pmario @CodaCodr 

Now it correctly checks if the stylesheets array has changed (also the order of the stylesheets)
I've added a utility function `arraysEqual` to utils.js because I haven't found such in the core and I'm using it in render.js and windows.js

If a stylesheet tiddler is added or removed or the order of the stylesheets changes or a stylesheet tiddler changes it recreates all `<style>` tags which is OK I think. Doing otherwise is very brittle and error prone.

BurningTreeC:
The styleRefresh times should be better now IF a stylesheet's content changes through a changed tiddler, like when we close/open the sidebar. If we change the tabs in the sidebar (which doesn't cause a stylesheet's content to change) then the styleRefresh times go slightly up since it's checking all stylesheet tiddlers for refresh changes.

BurningTreeC:
I'm now detecting the tag `$:/tags/Stylesheet/Static` on Stylesheet tiddlers and don't refresh those tiddlers.
I also made the `reset` tiddler such a static Stylesheet tiddler and it's listed before the vanilla/base Stylesheet

Jermolene:
Thanks @BurningTreeC that is very helpful. Looking at the code reminds me of the issue I encountered when I explored this problem. In effect, this PR contains an alternate implementation of the list widget. It does the same thing of continuously evaluating a filter and then adding, removing and updating DOM elements to match the filter results. There are similar optimisations, but a relatively hefty amount of infrastructure to support them.

If possible, I'd prefer to use the list widget itself, and apply any optimisations we can to it. The idea would be to wrap each stylesheet in `$:/core/ui/PageStylesheet` with a separate `<style>` tag. Then we can write tests that let us explore improving refresh handling. There are some [existing tests](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-widget.js) that use the fake DOM's sequenceNumbers to be able to precisely check which elements have been refreshed.

BurningTreeC:
> Thanks @BurningTreeC that is very helpful. Looking at the code reminds me of the issue I encountered when I explored this problem. In effect, this PR contains an alternate implementation of the list widget. It does the same thing of continuously evaluating a filter and then adding, removing and updating DOM elements to match the filter results. There are similar optimisations, but a relatively hefty amount of infrastructure to support them.
> 
> If possible, I'd prefer to use the list widget itself, and apply any optimisations we can to it. The idea would be to wrap each stylesheet in `$:/core/ui/PageStylesheet` with a separate `<style>` tag. Then we can write tests that let us explore improving refresh handling. There are some [existing tests](https://github.com/Jermolene/TiddlyWiki5/blob/master/editions/test/tiddlers/tests/test-widget.js) that use the fake DOM's sequenceNumbers to be able to precisely check which elements have been refreshed.

Hello @Jermolene - may I add a new `$:/core/ui/RootStylesheet` for this purpose since the `$:/core/ui/PageStylesheet` is used by the exporter and may be used by users as it is?

Jermolene:
> Hello @Jermolene - may I add a new `$:/core/ui/RootStylesheet` for this purpose since the `$:/core/ui/PageStylesheet` is used by the exporter and may be used by users as it is?

That's a good point, and I agree with your suggestion.

BurningTreeC:
Hi @Jermolene - is your idea that the RootStylesheet looks something like this:

```
title: $:/core/ui/RootStylesheet
code-body: yes

\import [subfilter{$:/core/config/GlobalImportFilter}]
\whitespace trim

<$set name="currentTiddler" value={{$:/language}}>

<$set name="languageTitle" value={{!!name}}>

<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
<$text text='<style type="text/css">'/>
<$transclude mode="block"/>
<$text text='</style>'/>
</$list>

</$set>

</$set>
```

?

Jermolene:
Hi @BurningTreeC I was thinking of a slightly different approach, but it is possible that your approach might be workable too.

My idea was to redefine the rendering of the stylesheets to render to `text/html` instead of `text/plain`, and then define the root stylesheet something like this:

```
title: $:/core/ui/RootStylesheet
code-body: yes

\import [subfilter{$:/core/config/GlobalImportFilter}]
\whitespace trim

<$set name="currentTiddler" value={{$:/language}}>

<$set name="languageTitle" value={{!!name}}>

<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
<style type="text/css">
<$view format="htmlwikified"/>
</style>
</$list>

</$set>

</$set>
```

The benefit of this approach is that it would allow us to extend the view widget to handle your optimisation whereby the text node is not replaced unless it has changed.

BurningTreeC:
> Hi @BurningTreeC I was thinking of a slightly different approach, but it is possible that your approach might be workable too.
> 
> My idea was to redefine the rendering of the stylesheets to render to `text/html` instead of `text/plain`, and then define the root stylesheet something like this:
> 
> ```
> title: $:/core/ui/RootStylesheet
> code-body: yes
> 
> \import [subfilter{$:/core/config/GlobalImportFilter}]
> \whitespace trim
> 
> <$set name="currentTiddler" value={{$:/language}}>
> 
> <$set name="languageTitle" value={{!!name}}>
> 
> <$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
> <style type="text/css">
> <$view format="htmlwikified"/>
> </style>
> </$list>
> 
> </$set>
> 
> </$set>
> ```
> 
> The benefit of this approach is that it would allow us to extend the view widget to handle your optimisation whereby the text node is not replaced unless it has changed.

Hello @Jermolene,

I believe we need to use the `transclude` widget here.
If we use the `view` widget, the text doesn't get updated at all if something in a stylesheet changes like a macro or a list widget. Or am I wrong?
Also, the `htmlwikified` format includes all the `<p>` tags. `plainwikified` works better.

Jermolene:
Hi @BurningTreeC apologies I did mean `plainwikified`, not `htmlwikified`.

You are also right that the view widget as it stands doesn't have any refresh handling. The point is that the view widget encapsulates the logic of selectively updating a text node with the text node content of a rendering, and so if we apply the necessary optimisations there then they should be reusable.

pmario:
I do not understand the purpose of the PR. It does cause a significant regression for "styleRefresh", since it does not prevent styleRefresh at all. 

As I wrote, at the moment TW does recalculate all styles with basically every keystroke. If I do compare styleRefresh with a tiddlywiki.html downloaded from tiddlywiki.com, it does a styleRefresh in about 3-5ms. 

With this PR is needs: 6-9ms with every UI interaction, which is a regression.

If I do "show / hide" right sidebar this PR uses 8-14ms

tiddlywiki.com If I do "show / hide" right sidebar this PR uses 5-7ms

For me it looks like that this PR tries to fix the problem in the completely wrong places.

BurningTreeC:
@pmario potentially we can optimize stylesheets for browsers. Browsers do handle them better if they are split up to more `<style>` tags.
I'm aware that the PR currently doesn't make the styleRefresh times better but I'm also exploring ways to make these times better.
I'll see where this can get...

BurningTreeC:
@pmario 

Note that the styleRefresh is the time it needs to replace the `<style>` tag(s) that change.
But here the styleRefresh reports the time it needs to check EVERY styleWidgetNode if it has widgets that refresh. At the end it replaces only the text content of those that effectively refreshed. The whole process needs more time but the effective styleRefresh is NOT calculated here.

BurningTreeC:
Hello @Jermolene 

can we assume that the `$:/core/ui/RootStylesheet` never changes or can we always translude the shadow tiddler? Is there an option to do so?

pmario:
IMO Every tiddler in TW can be modified by users or plugin authors.

BurningTreeC:
> IMO Every tiddler in TW can be modified by users or plugin authors.

You're right @pmario

BurningTreeC:
Closing this PR in favor of #8118

Issue #7117, opened 2022-12-13
[IDEA] Enable listing of all controlpanel settings


twMat:
**Problem**: The Controlpanel design is not good. It is difficult to locate settings. It is not searchable. It is difficult to modify the controlpanel because of multiple tab levels, at arbitrary places, *and* in arbitrary combination with hard coded content. For generated content, the filters are embedded in the tiddler text so the filter cannot be extracted for reuse. The whole controlpanel also has an archaic appearance.

**Describe the solution you'd like**
I want a list of *all* settings seen in the Controlpanel. I've long pondered over the problems mentioned above and concluded that some way to *list all settings* is the **minimum requirement** for doing further work with the items, such as creating custom controlpanels, searchability, etc.

I am not sure *how* this should be implemented, but it seems the most direct way would be a common tag on all relevant tiddlers. If this is considered cluttery, then instead a common *field*.

**Describe alternatives you've considered**
An alternative to a single *common* tag or field, would be a tag or field build up with name spaces but this would (1) require more work and (2) still be arbitrary. One neat aspect then could be that shared name spaces could be used to auto-select e.g template for how to present the setting.

**Other notes**
There are many other systems of greater complexity that feature better controlpanels, such as various OS and browsers - and they don't use a static Controlpanel like TW.

twMat:
@pmario - you bring up the problem with findability for translations. My OP would definitely simplify to solve this problem, but your problem is very specific where my request is much more general. I want to give the end user access to manipulate and hack the controlpanel in a *general* way. As noted, I've concluded that the first thing needed is a list of the titles of the tiddlers that are displayed as settings in the controlpanel. (I've updated the OP to clarify this.)

Let me clarify by using your quote:

>I personally would start with: https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FControlPanel%2FSettings .. If someone is able to make that list searchable with a different language than English, I think there is a good chance to get the rest going too.

[...]

PR #8108, opened 2024-03-26
tabs-missing-explicitState-added

pmario:
This PR will allow us to "copy / paste" tabs macro definitions into new tiddlers, without the need to adjust the "explicitState" parameter in the macro. 

- It will be used from the tiddler field. 
- Since it is a field it can be dynamically changed by user code or 3rd party plugins.
- It will also allow us to read the "needed" state from fields instead using regexps to read it from the macro parameter


Jermolene:
That's a great idea, thanks @pmario

pmario:
@Jermolene -- Improved readability and did some more testing. Is ready to be reviewed now

pmario:
uups forgot the tw function prefix `tf.` ... switched it draft again

pmario:
@Jermolene -- once Vercel built the preview it should be ready for review

core/wiki/macros/functions.tid
```diff
@@ -0,0 +1,5 @@
+title: $:/core/macros/functions
+tags: $:/tags/Global
+
+<!-- This global function can be used when the eexplicitState parameter is used with tabs-macros -->
+\function tf.getExplicitState() [<currentTab>get[explicitState]] ~[<currentTiddler>get[explicitState]] ~[<thisTiddler>get[explicitState]]
```
Jermolene, line 5: Why is this cascade needed? It feels like it could very confusing.

I don't like having to introduce such a specific core function. It doesn't seem particularly useful outside the core.

Jermolene (commented):
Thanks @pmario

core/wiki/macros/functions.tid
pmario: >Why is this cascade needed? It feels like it could very confusing.

It is necessary. Consider the following configuration. The story river contains 

- [$ControlPanel](https://tiddlywiki5-git-fork-pmario-explicit-f880f8-jermolenes-projects.vercel.app/#%24%3A%2FControlPanel) -> Info -> Cascades
- [$:/core/ui/ControlPanel/Cascades](https://tiddlywiki5-git-fork-pmario-explicit-f880f8-jermolenes-projects.vercel.app/#%24%3A%2Fcore%2Fui%2FControlPanel%2FCascades)
- "New Tiddler" with this code: `{{||$:/core/ui/ControlPanel/Cascades}}`

With the code in the PR all 3 tabs show the same "selected tab" -- all visible tiddlers can be used to switch tabs. All of them use the right explicitState value.

If `~[<thisTiddler>get[explicitState]]` is missing from the cascade, the "New Tiddler" will be **disconnected**, which is not what we want.

pmario: >I don't like having to introduce such a specific core function. It doesn't seem particularly useful outside the core.

At the moment this "function" is a bit "lonely". But I would like to be able to do this: 

`<<switch-to-tab root:"$:/ControlPanel" path:"Info/Cascades/View Template Title">>` and the tab should be selected in the $:/ControlPanel tiddler. 

Every tiddler that contains a "tabs" macro will probably need more meta-data, but the functionality will be globally useful for everyone, that uses the tabs-tiddlers with that configuration.

pmario:
Converted this one to a draft again, since I would like to rename and move the "global function" tiddler.

core/wiki/macros/functions.tid
pmario: So after calling an action or a button it should look like this:

image

pmario:
A side effect if a mechanism like this one `<<switch-to-tab root:"$:/ControlPanel" path:"Info/Advanced/Cascades/View Template Title">>`  works is: 

We should be able to globally search for parameters in the control panel **and** we should be able to **select** the right tab.

pmario:
@Jermolene -- I did some more experiments, if the whole PR is worth it. -- I think it is. This PR would be the first **"baby step"**

-----------

As I wrote at: https://github.com/TiddlyWiki/TiddlyWiki5/pull/8108#discussion_r1750156356

>Every tiddler that contains a "tabs" macro will probably need more meta-data, but the functionality will be globally useful for everyone, that uses the tabs-tiddlers with that configuration.

If we add most of the hardcoded variables into the tiddler fields eg: How the **tabsList** is created, we do have enough information in the fields, that we can connect all configurations together and show them in different ways. 

<details><summary>

#### Show Experimental "ControlPanel/Appearance" Code
</summary>

```
title: $:/core/ui/ControlPanel/Appearance
tags: $:/tags/ControlPanel
caption: {{$:/language/ControlPanel/Appearance/Caption}}
explicitState: $:/state/tab--1963855381
toc-filter: [all[shadows+tiddlers]tag[$:/tags/ControlPanel/Appearance]!has[draft.of]]

\whitespace trim
{{$:/language/ControlPanel/Appearance/Hint}}

<div class="tc-control-panel">
	<$macrocall $name="tabs"
		tabsList=<<tf.tabs-getListFilter>>
		default="$:/core/ui/ControlPanel/Theme"
		explicitState=<<tf.tabs-getExplicitState>>
	/>
</div>
```

</details>

**Possibilities:** 

- A long list of all settings (See screenshot below)
- Creating a Tabbed TOC-like view, where the tabs are on the left side (see screenshot below)
- Filtering will be extremely tricky because all texts are "<<lingo ..>>" (not sure if it is doable) 
  - What always works is - **Native browser search**, if everything is shown in a loooooong list
- Probably much more, which I did not think of

The following info is highly related to:  **[IDEA] Enable listing of all controlpanel settings** #7117 

### Experimental TOC Screenshot

image

**One more thing**

It seems to be backwards compatible

Jermolene:
Hi @pmario I do have some general reservations about the technique here.

Currently, tiddler fields are used to store information associated with a tiddler. The core does not use tiddler fields to store information used to render the content of a tiddler. While it's a technique that can be useful, the tradeoff is that it makes it harder to copy and paste or move text from one tiddler to another because one has to also remember to carry over the fields.

The limitations of the mechanism show why we don't use it in the core: it only works for a single tabset within a tiddler.

The core needs to have universal solutions. This seems like a neat dodge that would work OK in some situations, but is far from universal.

pmario:
We do have a lot of tiddlers where fields are responsible to define how the tiddler is handled. 

eg: `code-body`, `color`, `first-search-filter`, `second-search-filter` and so on.

I do see no difference here. Especially the `first-search-filter` and `second-search-filter` serve the exact same purpose. 

>... While it's a technique that can be useful, the tradeoff is that it makes it harder to copy and paste or move text from one tiddler to another because one has to also remember to carry over the fields.

The goal is, that every "tab-tiddler" can use the same tabs-code to be rendered. At the moment every tabs-macrocall needs to be manually maintained. All the filters work the same way, with only a small difference. We do have a lot of duplicated code. 

If a new tab should be created it is extremely complicated to define "deep links" to open the right tab in the ControlPanel.

Eg. If we want to open: **ControlPanel -> Info -> Advanced -> Cascades -> Story Tiddler**  -- There are 4 state tiddlers involved. For normal users it is almost impossible to define deep links. 

The code in this PR can help to solve that problem. But it is only the first step, to do so.

Jermolene:
> We do have a lot of tiddlers where fields are responsible to define how the tiddler is handled.
> 
> eg: `code-body`, `color`, `first-search-filter`, `second-search-filter` and so on.
> 
> I do see no difference here. Especially the `first-search-filter` and `second-search-filter` serve the exact same purpose.

There are a few differences. `first-search-filter`, `second-search-filter`  are components of the search architecture, and not a general purpose mechanism.

The critical difference is that `code-body`, `color` etc affect how the tiddler itself is displayed; they are referenced by the view template, and not referenced by user code within the body of a tiddler.

> The goal is, that every "tab-tiddler" can use the same tabs-code to be rendered. At the moment every tabs-macrocall needs to be manually maintained. All the filters work the same way, with only a small difference. We do have a lot of duplicated code.
> 
> If a new tab should be created it is extremely complicated to define "deep links" to open the right tab in the ControlPanel.

I understand the problem, but this is a very limited, constrained solution that only works in some specific circumstances. I think we would do better to rethink the control panel than resort to hacks like this.

PR #8109, opened 2024-03-27
Add basic authentication support for tm-http-request

Jermolene:
This PR adds support for basic authentication to the tm-http-request.

It was already possible to create an "Authorisation" header of the correct format with code like this:

```
header-Authorization={{{ [<username>addsuffix[:]addsuffix<password>encodebase64[]addprefix[Basic ]] }}}
```

However, that will only work if the password is available to wikitext. It is preferred to store passwords via the `<$password>` widget so that they are not accessible to casual wikitext.

This PR adds four new parameters to the `tm-http-request`message that allow the username and password to be passed as plain text, or taken from named entries in the password store:

* **basic-auth-username** - Optional username for HTTP basic authentication
* **basic-auth-username-from-store** - Optional username for HTTP basic authentication, specified as the name of the entry in the password store containing the username
* **basic-auth-password** - Optional password for HTTP basic authentication
* **basic-auth-password-from-store** - Optional password for HTTP basic authentication, specified as the name of the entry in the password store containing the password

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -47,7 +47,11 @@ exports.startup = function() {
 			headers: getPropertiesWithPrefix(params,"header-"),
 			passwordHeaders: getPropertiesWithPrefix(params,"password-header-"),
 			queryStrings: getPropertiesWithPrefix(params,"query-"),
-			passwordQueryStrings: getPropertiesWithPrefix(params,"password-query-")
+			passwordQueryStrings: getPropertiesWithPrefix(params,"password-query-"),
+			basicAuthUsername: params["basic-auth-username"],
+			basicAuthUsernameFromStore: params["basic-auth-username-from-store"],
+			basicAuthPassword: params["basic-auth-password"],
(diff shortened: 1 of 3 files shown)


Jermolene:
@mklauber do you have any thoughts on this? It is intended to make it easier to store basic auth credentials in the TW password vault.

mklauber:
I don't have an strong feeling about it.  tm-http-request is becoming a very complex widgetMessage, but I'm not sure if there's a better way to implement it.  

If we wrote a Filter Operator `password[...name of password...]` that returned strings from the password store, you could probably rewrite the above as a macro instead of embedding it inside the `tm-http-request`.  I'm not sure if that's desirable.  

Something like:
`header-Authorization=<<BasicAuth username:"username" password:"password" >>`  or `header-Authorization=<<BasicAuth username-from-store:"username" password-from-store:"password" >>`

mklauber:
Overall, I wouldn't been concerned about this PR until you asked for input.  Since you asked, those are my thoughts, but I'm happy with this as it is.

Jermolene:
Thanks @mklauber

PR #8112, opened 2024-03-28
Format $:/core/ui/EditTemplate/type

linonetwo:
So it is easier to modify it.

I use GPT4 to format it.

### DIFF
--- a/core/ui/EditTemplate/type.tid
+++ b/core/ui/EditTemplate/type.tid
@@ -2,50 +2,63 @@ title: $:/core/ui/EditTemplate/type
 tags: $:/tags/EditTemplate
 first-search-filter: [all[shadows+tiddlers]prefix[$:/language/Docs/Types/]sort[description]sort[group-sort]removeprefix[$:/language/Docs/Types/]search<userInput>]
 
-\define lingo-base() $:/language/EditTemplate/
-\define input-cancel-actions() <$list filter="[<storeTitle>get[text]] [<currentTiddler>get[type]] +[limit[1]]" emptyMessage="""<<cancel-delete-tiddler-actions "cancel">>"""><$action-sendmessage $message="tm-remove-field" $param="type"/><$action-deletetiddler $filter="[<typeInputTiddler>] [<refreshTitle>] [<typeSelectionTiddler>]"/></$list>
 \whitespace trim
+\procedure lingo-base() $:/language/EditTemplate/
+\procedure input-cancel-actions()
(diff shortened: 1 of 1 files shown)


core/ui/EditTemplate/type.tid
```diff
@@ -2,36 +2,113 @@ title: $:/core/ui/EditTemplate/type
...
-</div>
-<$set name="userInput" value={{{ [<typeInputTiddler>get[text]] }}}>
-<$list filter="[all[shadows+tiddlers]prefix[$:/language/Docs/Types/]group{!!group}] +[sort[description]] +[removeprefix[$:/language/Docs/Types/]] +[search<userInput>]"><span class={{{ [<currentTiddler>addsuffix[-primaryList]] -[<typeSelectionTiddler>get[text]] +[then[]else[tc-list-item-selected]] }}}><$link to={{{ [<currentTiddler>addprefix[$:/language/Docs/Types/]get[name]] }}}><$view tiddler={{{ [<currentTiddler>addprefix[$:/language/Docs/Types/]] }}} field="description"/><$text text=" "/>(<$view tiddler={{{ [<currentTiddler>addprefix[$:/language/Docs/Types/]] }}} field="name"/>)</$link></span>
-</$list>
-</$set>
+\procedure lingo-base() $:/language/EditTemplate/
+\procedure input-cancel-actions()
+<$list
+  filter="[<storeTitle>get[text]] [<currentTiddler>get[type]] +[limit[1]]"
+  emptyMessage="""<<cancel-delete-tiddler-actions "cancel">>"""
```
pmario, line 10: With TW we use tabs as indentation

core/ui/EditTemplate/type.tid
```diff
@@ -2,36 +2,113 @@ title: $:/core/ui/EditTemplate/type
...
+            state=<<qualify "$:/state/popup/type-dropdown">>
+            type="nomatch"
+            text=""
+            default=""
+          >
+            <div class="tc-block-dropdown tc-edit-type-dropdown">
+              <$linkcatcher to="!!type">
+                <$list
+                  filter='[all[shadows+tiddlers]prefix[$:/language/Docs/Types/]each[group]sort[group-sort]]'
+                >
```
pmario, line 79: I think, if there is only 1 line, it should stay as 1 line. It should not be formatted that way

linonetwo:
We can have a linter when we have ast-to-string. Leave formatting task to that day.

PR #8120, opened 2024-03-29
Update vanilla/reset to use modern-normalize 2.0.0

BurningTreeC:
This PR updates the vanilla/reset stylesheet to use the newer `modern-normalize 2.0.0`


pmario:
I think nothing really changed, except tab indent. So I'm OK with the update

Jermolene:
Thanks @BurningTreeC

PR #8121, opened 2024-03-30
Add a "Refresh Browser" keyboard shortcut

BurningTreeC:
In TiddlyDesktop the <kbd>ctrl-R</kbd> shortcut doesn't work and this adds a keyboard shortcut for that purpose.
It's very cumbersome having to open the "Tools" tab every time and going down to the dedicated button.

### DIFF
--- a/core/ui/KeyboardShortcuts/refresh.tid
+++ b/core/ui/KeyboardShortcuts/refresh.tid
@@ -1,0 +1,4 @@
+title: $:/core/ui/KeyboardShortcuts/refresh
+tags: $:/tags/KeyboardShortcut
+key: ((refresh))
+
--- a/core/ui/KeyboardShortcuts/refresh.tid
+++ b/core/ui/KeyboardShortcuts/refresh.tid
@@ -2,3 +2,4 @@ title: $:/core/ui/KeyboardShortcuts/refresh
 tags: $:/tags/KeyboardShortcut
 key: ((refresh))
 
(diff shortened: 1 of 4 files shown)


Jermolene:
Thanks @BurningTreeC I can see that this would be useful for TiddlyDesktop, but I wonder if it is overkill for the

Jermolene:
Thanks @BurningTreeC I can see that this would be useful for TiddlyDesktop, but I wonder if it is overkill for the TW core, given that browsers generally provide their own shortcut.

Are there other keyboard shortcuts that browsers generally provide that are missing in TiddlyDesktop?

BurningTreeC:
Hi @Jermolene 
Yes, this might be shipped with TiddlyDesktop and not with the core.
About missing keyboard shortcuts in TiddlyDesktop at the moment only <kbd>ctrl-F</kbd> comes to my mind

Jermolene:
> About missing keyboard shortcuts in TiddlyDesktop at the moment only ctrl-F comes to my mind

The other browser keyboard shortcut that I sometimes miss in TD is <kbd>cmd +</kbd> and <kbd>cmd -</kbd> to increase/decrease the page font size.

But on reflection we don't have an easy way for TiddlyDesktop to add shortcuts to hosted wikis. TiddlyDesktop doesn't require any special plugins within hosted wikis.

While we could introduce a new, optional TiddlyDesktop client plugin that includes these shortcuts, that does seem a lot of complexity. On that basis I'm happy to merge this.

PR #8130, opened 2024-04-04
Root stylesheet refresh - Stylesheets in single `<style>` tags

BurningTreeC:
This PR is a follow-up to #8106 , #8118 and #8125 
It's also related to the discussion #8128

This puts every tiddler tagged `$:/tags/Stylesheet` into its own `<style>` tag at the `BOTTOM` of the `<head>` section.
Why at the bottom? The `list` widget in `$:/core/ui/RootStylesheet` handles inserting and removing its list items all by itself. It does so by appending at the end of the `<head>` section. That's why from the beginning in `render.js` and `windows.js` I render the `styleWidgetNode` into the end of the `<head>` section and let the `list` widget do its "natural" flow.

This uses a modified `view` widget that's included in this PR. I'm not sure if I got @Jermolene right how it should be done but that's my try on it. It updates the `textContent` of its generated `domNode` in case the format is `plainwikified`, `htmlwikified` or `htmlencodedplainwikified` and the content changes, through a contained macro for example that changes.

The `$:/core/ui/RootStylesheet` distinguishes between standard Stylesheet tiddlers and Stylesheet tiddlers tagged `$:/tags/Stylesheet/Static`. In the latter case they are rendered as text and not refreshed which (I hope) makes it possible to save refresh time if we put static content into static Stylesheet tiddlers.

I open this as a draft PR since it has Performance Instrumentation enabled and the code might change based on the feedback I get.


themes/tiddlywiki/vanilla/reset.tid
```diff
@@ -1,5 +1,7 @@
 title: $:/themes/tiddlywiki/vanilla/reset
 type: text/css
+tags: $:/tags/Stylesheet $:/tags/Stylesheet/Static
+list-before: $:/themes/tiddlywiki/vanilla/base
```
pmario, line 4: hmmm, IMO that's not possible. If a user modifies the sort order by hand this shadow tiddler is converted to a system-tiddler and the `list-before` field is removed. -> That can lead to future problems for the user, if we change the `reset.css` again, because it is **no shadow** anymore

We will probably need to rename the tiddler to `$:/themes/tiddlywiki/vanilla/_reset`, so it will be automatically listed in front. 

@Jermolene -- The **problem** is, that renaming is not 100% backwards compatible

core/ui/RootStylesheet.tid
```diff
@@ -0,0 +1,14 @@
...
+\whitespace trim
+<$set name="currentTiddler" value={{$:/language}}>
+<$set name="languageTitle" value={{!!name}}>
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
+<style type="text/css">
+<$view format={{{ [<currentTiddler>tag[$:/tags/Stylesheet/Static]then[text]else[plainwikified]] }}}/>
+</style>
+</$list>
+</$set>
+</$set>
```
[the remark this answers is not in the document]
pmario: [$:/core/ui/PageStylesheet](https://tiddlywiki.com/#%24%3A%2Fcore%2Fui%2FPageStylesheet) uses `<$transclude mode="block"` to render current stylesheet tiddlers. The view-widget code here will do so using inline-mode atm. Is that intended?

core/modules/widgets/view.js
```diff
@@ -18,19 +18,95 @@ var ViewWidget = function(parseTreeNode,options) {
 	this.initialise(parseTreeNode,options);
 };
 
+var ViewHandler = function(parseTreeNode,options) {
+	this.initialise(parseTreeNode,options);
```
pmario, line 22: `this.initialise(parseTreeNode,options);` handles the initialisation of the options-object. So here it is OK to use it that way

core/modules/widgets/view.js
```diff
@@ -18,19 +18,95 @@ var ViewWidget = function(parseTreeNode,options) {
...
 
+/*
+Inherit from the base widget class
+*/
+ViewHandler.prototype = new Widget();
+
+/*
+Render this widget into the DOM
+*/
+ViewHandler.prototype.render = function(parent,nextSibling,options) {
```
pmario, line 38: Using `options` here is a problem if the function is called without the options parameter, which is common, because "options" is optional. 

So we need to initialize it -> otherwise line 42ff will probably throw RSODs

`options = options || {}`   without any var

core/modules/widgets/view.js
```diff
@@ -18,19 +18,95 @@ var ViewWidget = function(parseTreeNode,options) {
...
+			text: this.text,
+			title: this.viewTitle,
+			subTiddler: this.viewSubtiddler,
+			field: this.viewField,
+			index: this.viewIndex,
+			format: this.viewFormat,
+			template: this.viewTemplate,
+			mode: this.viewMode
+		}
+		this.viewHandler = new ViewHandler(this.parseTreeNode,{
```
pmario, line 102: I do not understand this. What does ViewHandler do?

Why does the ViewWidget instantiate a ViewHandler-widget in JS code?
If it is a widget, shouldn't it be be used in wikitext? 

I have the feeling, the functionality of ViewHandler is way to "intertwingled" with the functionality of the ViewWidget. 

Is the ViewHandler a "special / edge" case handling of the ViewWidget. -- If so it should extend the view-widget code.

core/modules/widgets/view.js
```diff
@@ -18,19 +18,95 @@ var ViewWidget = function(parseTreeNode,options) {
...
+				newText = $tw.utils.htmlEncode(this.fakeNode.textContent);
+				break;
+		}
+		if(newText !== this.text) {
+			this.domNodes[0].textContent = newText;
+			this.text = newText;
+		}
+	}
+	return refreshed;
+};
```
pmario, line 82: Hmmm, My gut-feelings tells me, there is something strange going on. -- ViewHandler is a widget. So should be in it's own file?

pmario (commented):
Please take my ViewHandler comments with a "bit of salt", I think I'm not 100% sure what it actually should do. 

The wiki itself seems to work, and the styles are added to the HEAD element

core/modules/startup/render.js
```diff
@@ -38,22 +38,13 @@ exports.startup = function() {
...
-	// Set up the styles
-	$tw.styleWidgetNode = $tw.wiki.makeTranscludeWidget(PAGE_STYLESHEET_TITLE,{document: $tw.fakeDocument});
-	$tw.styleContainer = $tw.fakeDocument.createElement("style");
-	$tw.styleWidgetNode.render($tw.styleContainer,null);
-	$tw.styleWidgetNode.assignedStyles = $tw.styleContainer.textContent;
-	$tw.styleElement = document.createElement("style");
-	$tw.styleElement.innerHTML = $tw.styleWidgetNode.assignedStyles;
-	document.head.insertBefore($tw.styleElement,document.head.firstChild);
+
+	var styleParser = $tw.wiki.parseTiddler("$:/core/ui/RootStylesheet",{parseAsInline: true}),
```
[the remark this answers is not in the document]
BurningTreeC: Yes you're right, I forgot to change that. Good thinking about the new name

core/modules/widgets/view.js
BurningTreeC: Hi @pmario , now ViewHandler isn't a widget anymore

core/modules/widgets/view.js
BurningTreeC: The viewHandler should be something like an independent object as far as I understand. But I don't understand a lot in this case 😄

core/modules/widgets/view.js
BurningTreeC: I've changed the code.

core/modules/widgets/view.js
BurningTreeC: This is now changed

core/modules/startup/render.js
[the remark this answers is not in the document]
BurningTreeC: Hi @pmario - I wasn't sure about how to name the constant, so I left `PAGE_STYLESHEET_TITLE` untouched and added `ROOT_STYLESHEET_TITLE` and use that title

core/ui/RootStylesheet.tid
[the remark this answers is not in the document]
BurningTreeC: I've now changed it to use a `<$let>` widget and reformatted the tiddler

BurningTreeC:
This PR now uses the `view` widget changes from #8135 for testing purposes

core/ui/RootStylesheet.tid
```diff
@@ -3,12 +3,21 @@ code-body: yes
...
-<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
-<style type="text/css">
-<$view format={{{ [<currentTiddler>tag[$:/tags/Stylesheet/Static]then[text]else[plainwikified]] }}}/>
-</style>
-</$list>
-</$set>
+
+	<$set name="languageTitle" value={{!!name}}>
+
+		<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]]">
```
pmario: We should not need the extra newlines for the indentation. Is there a reason?

core/ui/RootStylesheet.tid
```diff
@@ -0,0 +1,19 @@
+title: $:/core/ui/RootStylesheet
+code-body: yes
+
+\import [subfilter{$:/core/config/GlobalImportFilter}]
+\whitespace trim
+
+<$let currentTiddler" value={{$:/language}} languageTitle={{!!name}}>
```
pmario, line 7: There seems to be a typo .. `currentTiddler={{$:/language}}` let-widget has no value parameter

BurningTreeC: Oh yes, sorry. Changed

core/modules/startup/render.js
```diff
@@ -21,6 +21,7 @@ exports.synchronous = true;
 // Default story and history lists
 var PAGE_TITLE_TITLE = "$:/core/wiki/title";
 var PAGE_STYLESHEET_TITLE = "$:/core/ui/PageStylesheet";
```
pmario, line 23: Did you make sure, that `$:/core/ui/PageStylesheet` is not used hardcoded anywhere else?

BurningTreeC:
It fails because of the view widget problem... I'll try creating a new one

BurningTreeC:
Hi @pmario - do you remember this PR? Now it's passing the tests with the date parsing!

core/ui/RootStylesheet.tid
```diff
@@ -0,0 +1,18 @@
...
+
+\import [subfilter{$:/core/config/GlobalImportFilter}]
+\whitespace trim
+<$let currentTiddler={{$:/language}} languageTitle={{!!name}}>
+	<style type="text/css">
+		<$view tiddler="$:/themes/tiddlywiki/vanilla/reset" format="text" mode="block"/>
+	</style>
+	<style type="text/css">
+		<$view tiddler="$:/themes/tiddlywiki/vanilla/base" format="plainwikified" mode="block"/>
+	</style>
```
Leilei332: I wonder why vanilla theme should be hardcoded here.

core/modules/widgets/view.js
```diff
@@ -11,190 +11,300 @@ View widget
...
-		default: // "text"
-			this.text = this.getValueAsText();
-			break;
+var WikifiedViewHandler = function(widget) {
+	ViewHandler.call(this, widget);
+	this.fakeWidget = null;
+	this.fakeNode = null;
+};
+
+WikifiedViewHandler.prototype = Object.create(ViewHandler.prototype);
```
Leilei332: Since ES2017 is baseline now ,it may be more readable to use `class` syntx here

core/modules/widgets/view.js
```diff
@@ -18,325 +18,313 @@ ViewHandler Base Class
...
-StripCommentsViewHandler.prototype.getValue = function() {
-	var lines = this.widget.getValueAsText().split("\n"),
-		out = [];
-	for(var line = 0; line < lines.length; line++) {
-		var text = lines[line];
-		if(!/^\s*\/\/#/.test(text)) {
-			out.push(text);
+	getValue() {
+		const lines = this.widget.getValueAsText().split("\n");
+		const out = [];
```
pmario, line 252: An array, that is designed to collect values IMO should be let and not const.

Leilei332: An array defined by `const` is still possible to push items into it by using `array.push`, only overwriting like `array = [0, 1]` is protected.

See https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/const#const_in_objects_and_arrays

pmario: You are right. There will be more `const` definitions in "new" code, but `let` has a meaning. 

- `let` indicates, that usually later in the code it will be changed
- That's exactly what `out.push()` does

So if we need to indicate, that something will be changed. IMO It needs to be let, so we know it when we need to maintain the code. 

Have a closer look at this info at Youtube. https://youtu.be/dqmtzHB2zTM?t=1228

Where **the conclusion is** (which I think makes sense): There are 3 types of assignments in JS. 

- Actual Constants -> `const USE_SCREAMING_CASE` .. for real constants
- Constant references -> `const someVariable`
- Variable references -> `let someVariable` .. that should be changed somewhere later in the code

core/ui/RootStylesheet.tid
```diff
@@ -0,0 +1,15 @@
+title: $:/core/ui/RootStylesheet
+code-body: yes
+
+\import [subfilter{$:/core/config/GlobalImportFilter}]
+\whitespace trim
+<$let currentTiddler={{$:/language}} languageTitle={{!!name}}>
+	<style type="text/css">
+		<$view tiddler="$:/themes/tiddlywiki/vanilla/reset" format="text" mode="block"/>
+	</style>
```
Leilei332, line 9: I think we shouldn't hardcode stylesheet here. If a user swtiches to a theme that isn't based on vanilla theme, the vanilla theme will be disabled and the `$:/themes/tiddlywiki/vanilla/reset` will not exist. So IMO hardcoding reset does nothing here.

core/ui/RootStylesheet.tid
```diff
@@ -0,0 +1,15 @@
...
+	<$list filter="[all[shadows+tiddlers]tag[$:/tags/Stylesheet]!is[draft]] -[[$:/themes/tiddlywiki/vanilla/reset]]">
+		<style type="text/css">
+			<$view format={{{ [<currentTiddler>tag[$:/tags/Stylesheet/Static]then[text]else[plainwikified]] }}} mode="block"/>
```
Leilei332, line 12: We may add a comment that includes the title of the stylesheet tiddler at the top, which is useful for debugging.

```suggestion
			/* <<currentTiddler>> */
			<$view format={{{ [<currentTiddler>tag[$:/tags/Stylesheet/Static]then[text]else[plainwikified]] }}} mode="block"/>
```

Jermolene:
Thanks @BurningTreeC @Leilei332 @pmario is this now ready to merge?

Jermolene:
I am closing and re-opening this PR in order to force the CI to re-run the new linting.

Jermolene:
Hi @BurningTreeC there are some linting issues, if you have a chance to have a look at them? Many thanks.

core/modules/widgets/view.js
```diff
@@ -9,215 +9,423 @@ View widget
...
-	this.execute();
-	if(this.text) {
-		var textNode = this.document.createTextNode(this.text);
-		parent.insertBefore(textNode,nextSibling);
-		this.domNodes.push(textNode);
-	} else {
-		this.makeChildWidgets();
-		this.renderChildren(parent,nextSibling);
+class TextViewHandler extends ViewHandler {
+	constructor(widget) {
```
pmario: See: https://eslint.org/docs/latest/rules/no-useless-constructor

core/modules/widgets/view.js
```diff
@@ -9,215 +9,423 @@ View widget
 
 "use strict";
```
pmario, line 10: See: https://eslint.org/docs/latest/rules/max-classes-per-file

@Jermolene ... I think we should **not** enforce this one

core/modules/widgets/view.js
```diff
@@ -9,215 +9,423 @@ View widget
...
+			const textNode = this.document.createTextNode(this.text);
+			parent.insertBefore(textNode, nextSibling);
+			this.widget.domNodes.push(textNode);
+		} else {
+			this.widget.makeChildWidgets();
+			this.widget.renderChildren(parent, nextSibling);
+		}
+	}
+
+	refresh(changedTiddlers) {
```
pmario: see: https://eslint.org/docs/latest/rules/class-methods-use-this

Leilei332: IMO the refresh function here is used to override. So the file needs to add the following comment: 

```js
/*eslint class-methods-use-this: ["error", { "exceptMethods": ["refresh"] }] */
```

Jermolene:
Do not be alarmed! I am closing and then immediately re-opening this PR in order to force the new linting CI to run. It lints just the changed lines in the PR, making sure that new code fits the new coding guidelines, without requiring contributors to fix the linting for the entire file.

pmario:
There seems to be an ESLINT config problem

BurningTreeC:
Hi all, sorry to be late here, what can I do?
I have no idea about what ESLint is / does and what I can do to solve this but I will read the necessary information if you can point me to it

BurningTreeC:
Is the ESLint config problem an ESLint issue or can I remove / add something to resolve?

Leilei332:
> Is the ESLint config problem an ESLint issue or can I remove / add something to resolve?

This problem is now solved. You can merge the master branch to solve it.

For `no-useless-constructor`, you need to remove the `constructor` function in the class.

themes/tiddlywiki/vanilla/base.tid
```diff
@@ -1,7 +1,7 @@
 title: $:/themes/tiddlywiki/vanilla/base
 tags: [[$:/tags/Stylesheet]]
-list-before:
 code-body: yes
+list-before:
```
pmario, line 3: IMO this change should be removed. So the file is not part of the PR

BurningTreeC: removed

Jermolene:
Thanks @BurningTreeC

core/ui/RootStylesheet.tid
Leilei332, line 12: There is no need to add a `$:/tgs/Stylesheet/Static` tag. Tiddlers with type field set to `text/css` are regarded as static stylesheets in the previous implementation.

saqimtiaz:
@BurningTreeC after careful consideration, we have decided to revert this PR. We recognise and appreciate the significant amount of work you have put into it, but there are a few concerns with the implementation that unfortunately were not identified earlier:

- The root stylesheet hardcodes `vanilla/reset`, even though this is not relevant when a different theme is selected.

- Requiring users to learn an additional system tag to distinguish between dynamic and static stylesheet tiddlers is not ideal, particularly when this distinction is already clearly expressed via the tiddler type.

- The extensions to the view widget are problematic. This is a widget that likely should be deprecated over time in favour of equivalent functionality provided through filters.

We would very much welcome further discussion on this direction, in particular exploring how much of the view widget’s existing functionality could be replaced by new filter operators, and identifying any remaining use cases that are not well served by filters today.

Issue #8129, opened 2024-04-03
[IDEA] Generalize the tree macro to support any field


andjar:
Hi! Currently, the tree macro is only looking at the title field. It may be useful to generalize it to support an additional field parameter that defines the path. A tiddler with fields
```html
title: /A/B/C
path1: /B/A/C
path2: /C/B/A
```

will then be placed at different location using the calls

```html
<<tree prefix:"/">> // Defaults to title
<<tree prefix:"/" field:"path1">>
<<tree prefix:"/" field:"path2">>
```

PR #8134, opened 2024-04-05
generalize the tree macro to support any field

andjar:
The tree macro is modified to support a `field` parameter, defaulting to title for backward compatibility. As suggested in #8129, a tiddler with fields

```
title: /A/B/C
path1: /B/A/C
path2: /C/B/A
```
will be placed at different locations in the tree using the calls
```
<<tree prefix:"/">> // Defaults to title
<<tree prefix:"/" field:"path1">>
<<tree prefix:"/" field:"path2">>
```

The performance seems suboptimal compared to the original implementation, possibly due to nested looping over all tiddlers in `tree_node`:

```
<$list filter="[all[shadows+tiddlers]get<__field__>prefix<__prefix__>removeprefix<__prefix__>splitbefore<__separator__>sort[]!suffix<__separator__>]" variable="chunk">
<$list filter="[all[shadows+tiddlers]has<__field__>sort[]] :filter[get<__field__>removeprefix<__prefix__>match<chunk>]" variable="title">
<$macrocall $name="leaf-node" prefix=<<__prefix__>> chunk=<<chunk>> title=<<title>> separator=<<__separator__>>/>
</$list>
</$list>
```

### DIFF
--- a/core/wiki/macros/tree.tid
+++ b/core/wiki/macros/tree.tid
@@ -1,22 +1,15 @@
 title: $:/core/macros/tree
 tags: $:/tags/Macro
 
-\define leaf-link(full-title,chunk,separator: "/")
-<$link to=<<__full-title__>>><$text text=<<__chunk__>>/></$link>
-\end
-
-\define leaf-node(prefix,chunk)
+\define leaf-node(chunk,title)
 \whitespace trim
 <li>
-<$list filter="[<__prefix__>addsuffix<__chunk__>is[shadow]] [<__prefix__>addsuffix<__chunk__>is[tiddler]]" variable="full-title">
-<$list filter="[<full-title>removeprefix<__prefix__>]" variable="chunk">
-<span>{{$:/core/images/file}}</span>&#32;<$macrocall $name="leaf-link" full-title=<<full-title>> chunk=<<chunk>>/>
-</$list>
-</$list>
+<span>{{$:/core/images/file}}</span>&#32;
+<$link to=<<__title__>>><$text text=<<__chunk__>>/></$link>
 </li>
 \end
 
-\define branch-node(prefix,chunk,separator: "/")
+\define branch-node(prefix,chunk,field,separator: "/")
 \whitespace trim
 <li>
 <$set name="reveal-state" value={{{ [[$:/state/tree/]addsuffix<__prefix__>addsuffix<__chunk__>] }}}>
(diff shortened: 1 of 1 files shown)


core/wiki/macros/tree.tid
```diff
@@ -1,22 +1,15 @@
...
-<span>{{$:/core/images/file}}</span>&#32;<$macrocall $name="leaf-link" full-title=<<full-title>> chunk=<<chunk>>/>
-</$list>
-</$list>
+<span>{{$:/core/images/file}}</span>&#32;
+<$link to=<<__title__>>><$text text=<<__chunk__>>/></$link>
 </li>
 \end
 
-\define branch-node(prefix,chunk,separator: "/")
+\define branch-node(prefix,chunk,field,separator: "/")
```
pmario, line 12: For backwards compatibility reasons, we cannot change a macro signature. **So the new parameter has to go to the end of every macro call**, because TW allows unnamed parameter calls, where order is important. So you change the order of the parameters here, which will invalidate existing user code.

It's not nice, but it has to be that way

core/wiki/macros/tree.tid
```diff
@@ -31,32 +24,34 @@ tags: $:/tags/Macro
...
-\define tree-node(prefix,separator: "/")
+\define tree-node(prefix,field,separator: "/")
 \whitespace trim
 <ol>
-<$list filter="[all[shadows+tiddlers]removeprefix<__prefix__>splitbefore<__separator__>sort[]!suffix<__separator__>]" variable="chunk">
-<$macrocall $name="leaf-node" prefix=<<__prefix__>> chunk=<<chunk>> separator=<<__separator__>>/>
+<$list filter="[all[shadows+tiddlers]get<__field__>prefix<__prefix__>removeprefix<__prefix__>splitbefore<__separator__>sort[]!suffix<__separator__>]" variable="chunk">
+<$list filter="[all[shadows+tiddlers]has<__field__>sort[]] :filter[get<__field__>removeprefix<__prefix__>match<chunk>]" variable="title">
+<$macrocall $name="leaf-node" prefix=<<__prefix__>> chunk=<<chunk>> title=<<title>> separator=<<__separator__>>/>
+</$list>
```
pmario, line 41: Having 2 lists that iterate over every shadow and every tiddler starting from new, may be the performance bottle neck. So if possible we may be able to "reuse" some results from the first list as input to the second one. -- **But** I did not have a look at the code at the moment.

core/wiki/macros/tree.tid
andjar: Thank you, this was a very good point!

core/wiki/macros/tree.tid
andjar: I agree that the performance must be improved. The challenge, I think, is that we need to (1) find and sort the list of all valid field values (i.e., those with a given prefix) and (2) find all tiddlers with the given field values

One strategy may be to

1. Find tiddlers with valid field values
2. Join field value and tiddler title with the separator
3. Sort the list of field value-tiddler title strings
4. Split by the separator and somehow extract the last element (tiddler title)

But that would require a larger re-write of the tree macro

CeeBeeTree:
Hi all - in developing a change for another issue - i.e. [allow new tabs to be added to the plugin info box #9002](https://github.com/TiddlyWiki/TiddlyWiki5/pull/9002) .. I realised the solution could also work for this issue / feature request.

I've added a filter parameter to the tree macro now that the latest updates to that (to rectify them as procedures) was committed to the pre-release. I've tested this including some performance tests and explanations on the usage that will append a path field to a title to build the tree  - its available [here](https://ceebee-demos.tiddlyhost.com/#%24%3A%2Fplugins%2Fceebee%2FPhack:%5B%5BTree%20Macro%20Procedure%20Update%5D%5D%20%24%3A%2Fplugins%2Fceebee%2FPhack)

PR #8135, opened 2024-04-06
View Widget View handler

BurningTreeC:
This PR adds a `ViewHandler` base class to the `view` widget and `subclasses` for each of the different views

The `ViewHandler` has methods `render`, `renderWikified`, `createTextNode`, `createWikifiedTextNode`, `createFakeWidget` and `refresh` - EDIT: and now also `refreshWikified`
These methods are used by more than one of the different subclasses.

The subclasses are initialised if they're needed.
There's still a big `switch` statement in the `ViewWidget`'s `getView` method to find the view and initialise the correct subclass, maybe there's a better method to do so but I haven't found one.

Each subclass has its own `getValue` method.
The wikified views also have dedicated `render` and `refresh` methods.

I hope I got this right this time :smile_cat: 

Please review and test and leave me some comments if this is the right approach and what could be simplified.
I had to read about subclassing for this and am not 100% sure if I got it right.

### DIFF
--- a/core/modules/widgets/view.js
+++ b/core/modules/widgets/view.js
@@ -75,6 +75,7 @@ ViewHandler.prototype.createFakeWidget = function() {
 		field: this.widget.viewField,
 		index: this.widget.viewIndex,
 		parseAsInline: this.widget.viewMode !== "block",
+		mode: this.widget.viewMode === "block" ? "block" : "inline",
 		parentWidget: this.widget,
 		subTiddler: this.widget.viewSubTiddler
 	});
--- a/core/modules/widgets/view.js
+++ b/core/modules/widgets/view.js
@@ -38,8 +38,8 @@ ViewHandler.prototype.render = function(parent,nextSibling) {
 Base ViewHandler render method for wikified views
 */
 ViewHandler.prototype.renderWikified = function(parent,nextSibling) {
-	this.text = this.getValue(this.widget.viewMode);
 	this.createFakeWidget();
+	this.text = this.getValue();
 	this.createWikifiedTextNode(parent,nextSibling);
 };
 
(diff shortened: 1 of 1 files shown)


Jermolene:
Thanks @BurningTreeC I am just linking this back to the earlier discussion in [#8125](https://github.com/Jermolene/TiddlyWiki5/pull/8125#issuecomment-2037490188).

Sorry for the delay. I am now a bit concerned that this is a lot of complexity to introduce in order to be able to attempt the optimisation proposed in #8105. However, I am not sure that at this point we have confidence that the eventual performance improvements will be worthwhile.

Does this PR as it stand give you enough material to be able to explore the performance impact on #8105?

BurningTreeC:
Hello @Jermolene 

sorry for the delay

I couldn't measure any real performance improvements with the stock TiddlyWiki
I just had this idea when playing with TiddlyFlex - I don't know if you know it, it's here: https://burningtreec.github.io/TiddlyFlex/

I thought it might be better performance-wise if only those chunks of CSS would get replaced that really refresh

Jermolene:
Thanks @BurningTreeC – I do think these are worthwhile improvements. Particularly making the refreshing of the view widget much more usable. So I am inclined to merge this. I'd appreciate a second set of eyes – @saqimtiaz do you have any thoughts?

BurningTreeC:
Hello @Jermolene @saqimtiaz , could we try see if this PR could be considered?

Jermolene:
Thanks @BurningTreeC

Jermolene:
Hi @BurningTreeC we've had reports of compatibility issues in #8611 so I have reverted this for the moment in 678c272979d2ca4813bca7bbb07ce45d82f9e7c2

pmario:
@Jermolene -- reverting makes it impossible for me to push the fix that I did test.

It needed more time, since I did create 9 new tests to make the function more testable. 9 tests are probably not enough, so I did not create a PR yet. 

@BurningTreeC can you create a new PR with exactly the same content as here. So I can crate a PR for your repo. So you can test my fix. 

It's relatively simple, but I was not sure, if there are any side effects. That's why I did create some tests. I did mark teh differences to your code with `// NEW`

```
/*
Initialise date mehods
*/
ViewWidget.prototype.initialiseDateView = function(View) {
	var self = this;
	View.prototype.getValue = function(format) {
		format = format || self.viewTemplate || "YYYY MM DD 0hh:0mm";  // NEW
		var value = $tw.utils.parseDate(self.getValue(format)); // NEW
		if(value && $tw.utils.isDate(value) && value.toString() !== "Invalid Date") {
			return $tw.utils.formatDateString(value,format);
		} else {
			return "";
		}
	};
	return View;
};
```

Jermolene:
Hi @pmario given the work on #8630 do you still need me to un-revert the previous PR?

pmario:
>Hi @pmario given the work on https://github.com/TiddlyWiki/TiddlyWiki5/pull/8630 do you still need me to un-revert the previous PR?

I think we will be able to sort it out. So no actions needed here.

Issue #7878, opened 2023-12-07
[BUG] v5.2.3 changed the "save" command to not overwrite files


Jermolene:
The PR #6650 by @joshuafontany which was included in v5.2.3 attempts to fix #6649 by using the filesystem utils `generateTiddlerFileInfo()` and `saveTiddlerToFileSync()`.

It turns out that this change has an unexpected side effect, which is that the save command will no longer overwrite existing files, instead appending an ascending numeric suffix.

In previous versions of TiddlyWiki, the save command would overwrite existing files if asked.

saqimtiaz:
Fixed in #8140

PR #8140, opened 2024-04-10
Fix bug 7878: Save command

joshuafontany:
This restores the previous behavior of the `save` command, as noted in bug https://github.com/Jermolene/TiddlyWiki5/issues/7878.

Should this be configurable with command arguments?

### DIFF
--- a/core/modules/commands/save.js
+++ b/core/modules/commands/save.js
@@ -43,7 +43,9 @@ Saves individual tiddlers in their raw text or binary format to the specified fi
 						directory: path.resolve(self.commander.outputPath),
 						pathFilters: [filenameFilter],
 						wiki: wiki,
-						fileInfo: {}
+						fileInfo: {
+							overwrite: true
+						}
 					});
 					if(self.commander.verbose) {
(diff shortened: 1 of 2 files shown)


Jermolene:
Thank you @joshuafontany 

> Should this be configurable with command arguments?

It would be quite useful, but I'm not sure that it would be very elegant given the way that the save command uses positional arguments.

One approach would be to invent a new uber-command that encompasses the functionality of both `--save` and `--render`, but uses the same command line parsing as `--listen` so that we can offer exhaustive options. And then we'd define the `--save` command and the `--render` command as simple, positional, aliases for the uber-command.

We haven't used `--write` so that might be a possibility. I quite like the resonance with `--listen`.

joshuafontany:
I agree and I can see myself using the server to generate static files for many more purposes now. Might be worth a separate issue. The big logic issue here was that the server saves filepath info for tiddlers loaded at startup, but has no context for files already in the output folder. A simple `overwrite` flag in the fileInfo object seemed cleanest fix for now.

@Jermolene I have fixed a typo and tested, works as intended now.

Jermolene:
Thanks @joshuafontany

Issue #8144, opened 2024-04-15
[BUG] Crash on empty transclusion syntax inside button widget 


bluepenguindeveloper:
### Describe the bug

Wikitext such as `<$button>{{}}</$button>` (or unclosed button `<$button>{{}}`) results in an immediate crash.

Found and tested with single file wiki on Windows 10 with TiddlyDesktop and the Brave browser (both Chromium-based)

### Expected behavior

Recursive transclusion error (without crashing TiddlyWiki)

### To Reproduce

1. Click Create New Tiddler button
2. Enter `<$button>{{}}</$button>` or `<$button>{{}}` into the text field
3. Turn on side by side preview
4. Note that TiddlyWiki crashes

Steps 2 and 3 can be done in reverse order, in which case, the crash will occur as soon as the second `}` is typed if typing manually.

### TiddlyWiki Configuration

- TW Version [e.g. v5.3.3]
- Saving mechanism: TiddlyDesktop, or none (occurs if opening single file wiki in the browser (Brave), and can be done in the TiddlyWiki documentation simply viewed in the browser); seems to be independent of saving mechanism

- OS: Windows 10
- Browser: Brave (Version 1.64.116; Chromium: 123.0.6312.105)

Jermolene:
I've prepared a PR that takes a different approach and resolves this at the parsing level, I'll commit the changes shortly.

pmario:
@Jermolene My idea can be seen in the screenshot. 

Since "button inside button inside a button ... " does not make much sense, we know that we should stop it early. In the screenshot it stops at ancestorCount() 334, which should not cause too many problems. But I would need to test this one with FF on ubuntu again. 

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/05ab7e63-e8ef-471a-953b-b459d9a40aa9)

The idea would need some refinement. Since I think "button in button" should only be allowed eg: 10 times for any weird usecase. 

I think `this.getAncestorCount()` could be overwritten by the button-widget to detect a "button inside button" situation and handle that one individually. 

Just some thoughts.

Jermolene:
Hi @pmario in my tests, your example of `<$button><$transclude>` doesn't crash the browser if I reduce MAX_WIDGET_TREE_DEPTH to 300. So I wonder if a reasonable solution with your example isn't just to reduce that constant. I chose 1000 as a reasonable guess, but there is now evidence that it is too high.

pmario:
[...]

PR #8145, opened 2024-04-15
Fix crash with empty transclusions

Jermolene:
This PR addresses infinite loops that can occur when `{{}}` is typed in the editor with the preview displayed. See #8144 for an example, but this is a long standing problem. See #7768 for an earlier attempt to fix this problem

The approach taken here is to modify the parse rules for both inline and block transclusions so that the transclusion is not recognised unless there is a non-blank target and/or template.

### DIFF
--- a/core/modules/parsers/wikiparser/rules/transcludeblock.js
+++ b/core/modules/parsers/wikiparser/rules/transcludeblock.js
@@ -26,6 +26,27 @@ exports.init = function(parser) {
 	this.matchRegExp = /\{\{([^\{\}\|]*)(?:\|\|([^\|\{\}]+))?(?:\|([^\{\}]+))?\}\}(?:\r?\n|$)/mg;
 };
 
+/*
+Reject the match if we don't have a template or text reference
+*/
+exports.findNextMatch = function(startPos) {
+	this.matchRegExp.lastIndex = startPos;
+	this.match = this.matchRegExp.exec(this.parser.source);
+	if(this.match) {
+		var template = $tw.utils.trim(this.match[2]),
+			textRef = $tw.utils.trim(this.match[1]);
+		// Bail if we don't have a template or text reference
+		if(!template && !textRef) {
+			return undefined;
+		} else {
+			return this.match.index;
+		}
+	} else {
+		return undefined;
+	}
+	return this.match ? this.match.index : undefined;
+};
+
 exports.parse = function() {
 	// Move past the match
 	this.parser.pos = this.matchRegExp.lastIndex;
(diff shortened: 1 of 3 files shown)


pmario:
hmmm, That's OK for 

```
```

But it is still a problem with "unfinished" code as follows. 

```
<$button><$transclude>
```

Jermolene:
> hmmm, That's OK for

@pmario yes this is only a fix for #8144. I agree we still have a problem with recursive transclusions, but as you well know that is a much, much subtler problem. The benefit of the fix here is that it doesn't even parse `{{}}` as a transclusion, which makes it clearer that it is an invalid construction (particularly if we had syntax highlighting), and we don't need to engage any recursion detection which will always of necessity be resource intensive.

pmario:
I do think that this PR is an improvement for many users already. I think it should be included.

But I also think, that we need a runtime fix for my code. From time to time I do have a problem with this type of code at runtime. Especially if variables have a typo and are therefore empty

bluepenguindeveloper:
It works as stated (doesn't fix a more longstanding recursion issue but does fix the specific issue in #8144 by not treating `{{}}` as a transclusion), but I do have a doubt: sometimes in templates or lists etc., we might want to transclude the text of the current tiddler, like `{{!!text}}`, but if `{{}}` was heretofore equivalent to `{{!!text}}` and if anyone is currently using `{{}}` instead of `{{!!text}}` before, then this might be considered a breaking change.

pmario:
@bluepenguindeveloper -- I am creating a second PR atm, which tries to solve the "runtime" side of that problem. From time to time I'm also a "victim" of that problem especially since development where the code is still buggy.

Jermolene:
> It works as stated (doesn't fix a more longstanding recursion issue but does fix the specific issue in #8144 by not treating `{{}}` as a transclusion), but I do have a doubt: sometimes in templates or lists etc., we might want to transclude the text of the current tiddler, like `{{!!text}}`, but if `{{}}` was heretofore equivalent to `{{!!text}}` and if anyone is currently using `{{}}` instead of `{{!!text}}` before, then this might be considered a breaking change.

Thanks @bluepenguindeveloper. This PR only changes the behaviour where the trimmed contents of the `{{}}` are empty, and doesn't affect the processing of `{{!!text}}`.

This is indeed not strictly backwards compatible because `{{}}` is currently valid as a synonym for `<$transclude/>`, which is in turn interpreted as transcluding the current tiddler. However `{{}}` isn't used by the core, and so I suspect is not a common usage.

Perhaps the best way to assess the backwards compatibility impact of this PR is to go ahead and merge it so that it gets tested in the prerelease.

saqimtiaz:
Suggest we delay this until v5.4 as it is not strictly backwards compatible.

Jermolene:
Thanks @saqimtiaz I agree

PR #8148, opened 2024-04-16
Remove the `x-requested-with` header from the tm-http-request message

mklauber:
Extraneous headers can interfere with CORS validation when hitting external APIs, due to tightly configured `Access-Control-Allow-Headers` headers.  Until there's a way to disable these headers, it's probably better (i.e. most flexible) to expect users to set them deliberately.  

In this case, I ran into an API that uses CORS and the above header, and was unable to make a request to it without making the changes in this PR.  I'm also open to modifying the message to allow disabling these headers, but this was the simplest solution I could come up with.

### DIFF
--- a/core/modules/utils/dom/http.js
+++ b/core/modules/utils/dom/http.js
@@ -245,20 +245,6 @@ exports.httpRequest = function(options) {
 		getHeader = function(targetHeader) {
 			return headers[targetHeader] || headers[targetHeader.toLowerCase()];
 		},
-		isSimpleRequest = function(type,headers) {
-			if(["GET","HEAD","POST"].indexOf(type) === -1) {
-				return false;
-			}
-			for(var header in headers) {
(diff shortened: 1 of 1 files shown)


Jermolene:
Hi @mklauber I think that is reasonable, however we would also need to adjust tiddlywebadaptor.js to manually set the `x-requested-with` header because it is required by the TiddlyWiki API.

I think that any existing user code hitting a TW-based API would fail with this change. Perhaps instead of removing the header we could add a flag for explicitly disabling it? Ideally, we'd also expose that flag to user code via the `tm-http-request` message.

mklauber:
I suggest setting a flag to only use explicit headers.  I can try to write that, it may take me a few days.

mklauber:
replaced with https://github.com/Jermolene/TiddlyWiki5/pull/8152

PR #8150, opened 2024-04-16
Return responseText for all 2XX response codes during handling of `tm-http-request` messages.

mklauber:
APIs especially use 2XX response codes outside of 200, 201, 204 for responding to responses.  Treat all "Successful" response codes (i.e. anything between 200-299) as successes, and pass the responseText.


Jermolene:
Thanks @mklauber I agree, the existing code is naive.

PR #8152, opened 2024-04-16
Add defaultHeaders flag that controls helpful default heders that can sometimes interfere with apis

mklauber:
Extraneous headers can interfere with CORS validation when hitting external APIs, due to tightly configured `Access-Control-Allow-Headers` headers. Add a way to disable these headers when they are causing issues.

### DIFF
--- a/core/modules/startup/rootwidget.js
+++ b/core/modules/startup/rootwidget.js
@@ -39,6 +39,7 @@ exports.startup = function() {
 			method: params.method,
 			body: params.body,
 			binary: params.binary,
+			defaultHeaders: params.defaultHeaders,
 			oncompletion: params.oncompletion,
 			onprogress: params.onprogress,
 			bindStatus: params["bind-status"],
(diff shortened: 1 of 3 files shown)


Jermolene:
Thanks @mklauber and apologies for the delay.

Jermolene:
Hi @mklauber I've reverted this PR because of an error [reported by](https://github.com/Jermolene/TiddlyWiki5/commit/6910be795f6c578094a4fd781283bc49f83c43eb#commitcomment-142509735) @pmario – the client-server edition fails to start.

As well as fixing the problem, I think it would be better to rename the parameter/variable to "useDefaultHeaders".

mklauber:
Understood.  I'll look into that error, and also update that variable.

mklauber:
New PR up here.  https://github.com/Jermolene/TiddlyWiki5/pull/8225

Jermolene:
Thanks @mklauber

PR #8157, opened 2024-04-19
add a search button to the control panel tiddler fields tab

pmario:
This PR adds a new column to the ControlPanel -> Advanced -> Tiddler Fields tab. 
It will copy the filter into the AdvancedSearch, so users get an overview. 

grafik


pmario:
There has been a discussion at Talk lately: https://talk.tiddlywiki.org/t/list-of-all-fields-used-by-the-core/9600

It should improve the situation a bit.

core/wiki/allfields.tid
```diff
@@ -1,13 +1,41 @@
...
+		//<$text text={{$:/language/Docs/Fields/$title$}}/>//
+	</td>
+	<td class="tc-view-field-links">
+		<<show-filter-count "[has[$title$]sort[]]">>
+	</td>
+</tr>
 \end
-\whitespace trim
+
+\define renderfieldHeader()
```
Jermolene, line 20: Would we be better of making these procedures?

core/wiki/allfields.tid
```diff
@@ -1,13 +1,41 @@
...
+<tr class="tc-view-field">
+	<th class="tc-view-field-name">
+		''Name'':
+	</th>
+	<th class="tc-view-field-value">
+		''Description''
+	</th>
+	<th class="tc-view-field-links">
+		''Links''
```
Jermolene, line 29: Perhaps "Show" would be clearer?

Jermolene (commented):
Thanks @pmario this is a good idea, just a couple of points.

core/wiki/allfields.tid
pmario: I did use procedures, but it breaks the "show-filter-count" macro, which is imported from [[$:/core/ui/ControlPanel/Basics]]. It expects the parameter as a `$title$`. 

I did not want to change all the stuff and make it backwards compatible, just to add this one button. Filters usually have to be used very different if we remove the "old substitution" style. 

That's why I used `\define`

pmario: I think I can make the header a procedure. But IMO `renderfield()` should stay a macro

pmario: I did change the macros to use procedures

pmario:
@Jermolene -- Updated the PR according to your suggestions

pmario:
@AnthonyMuscio  As a "side effect" this PR also makes search-filter-count a global macro as discussed at: 

**Introduce a new macro to the core, well not really - search-filter-count?** #7203

pmario:
@Jermolene -- bump

core/wiki/allfields.tid
```diff
@@ -1,13 +1,43 @@
...
+<tr class="tc-view-field">
+	<td class="tc-view-field-name">
+		''<$text text=<<title>>/>'':
+	</td>
+	<td class="tc-view-field-value">
+		//<<tf.getLingoText>>//
+	</td>
+	<td class="tc-view-field-links">
+		<!-- show-filter-count expects a filter string. It does no more processing -->
+		<$macrocall $name="show-filter-count" filter=`[has[$(title)$]sort[]]`>>
```
Jermolene, line 17: The use of textual substitution gives a subtle regression here; it means that renderfield will no longer work with a title that contains a closing square bracket.

core/wiki/macros/show-filter-count.tid
```diff
@@ -0,0 +1,16 @@
+title: $:/core/macros/show-filter-count
+tags: $:/tags/Macro $:/tags/Global
+
+\procedure show-filter-count(filter)
```
Jermolene, line 4: Now that it is reusable we should add documentation for this procedure.

Jermolene (commented):
Thanks @pmario

core/wiki/allfields.tid
pmario: IMO it can't be a regression, since the "show-filter-count" macro is new to this PR. The problem is, that the "substituted" filter is copied "as is" to the $AdvancedSearch input field, which needs a valid filter string. So if we would use `[has<title>sort[]]` this exact string would be copied into the input, which will not work. 

Eg: an `author]` field can not be witten into the AdvancedSearch at the moment. It would result in `[has[author]]sort[]]` 
At the moment AdvancedSearch can not handle `[has<title>sort[]]` yet.

pmario: I'll change the function "signature" to: `\procedure renderfield(field)`

pmario:
Changed according to your remarks:

grafik

core/language/en-GB/Fields.multids
```diff
@@ -1,5 +1,8 @@
 title: $:/language/Docs/Fields/
 
+table-header-name-20240804162913000: Name
```
Jermolene, line 3: What is the thinking behind including a timestamp in the fieldname?

Jermolene (commented):
Thanks @pmario

core/language/en-GB/Fields.multids
pmario: I wanted to have a unique field-name to avoid name clashes. `table-header-name` is a perfectly fine field-name that users could use. I thought it is very unlikely, that someone will use an "outdated" date for their field-names in the future. Firs I did use `-xyzabc`. It's unlikely to cause name clashes, but I was more confident with the date.

I'm open to change the name

pmario:
@Jermolene -- There is no table header in the current implementation, so I did remove it for this PR. It probably was a bad idea to add it in the first place.

editions/tw5.com/tiddlers/macros/show-filter-count.tid
```diff
@@ -0,0 +1,21 @@
+caption: colour
```
Jermolene, line 1: Just noticed that this caption is wrong

Jermolene (commented):
Thanks @pmario

pmario:
@Jermolene -- Merged master into this branch. It still works -- wrong "caption" field has been removed

Jermolene:
Thanks @pmario

PR #8163, opened 2024-04-22
add new classes to wikitext created emphasised elements to make them easily stylable

### LINKED ISSUE #8162: [Bug] wikitext formatting a link in italic works - bold does not work
```
//[[Internal link|example in italics]]//

''[[Internal link|example in italics]]''
```

- The current TW version does style the link in italic as it should, because the user defined it that way.
- The internal link is not styled in **bold**, because the link is "covered" in a STRONG element, which has a very low specificity.

We need a new style definition to define the font weight, if the user decides to format a link in bold. 

```
title: link-styles
tags: $:/tags/Stylesheet

strong .tc-tiddlylink {
  font-weight: bolder;
}
```

pmario:
This PR fixes  #8162 

- #8162


Jermolene:
Hi @pmario I am concerned that the CSS selector here may trigger some false positives. The most obvious anomaly would be something like this:

```
<strong>
This is bold
<div style="font-weight:normal;">
This is not bold
<$link>
This should not be bold
<$/link>
</div>
</strong>
```

pmario:
>I am concerned that the CSS selector here may trigger some false positives

You are right. We need to be more specific. I had a look. We should add a new `tc-emphasis tc-strong` to the STRONG element created by the wikitext bold-parser.

The same should be true for: `italic, strikethrough, subscript, subscript, superscript, underscore` and inline-`code`

pmario:
It seems to work fine now

```
<strong>
This is bold, 
<div style="font-weight:normal;">

This is not bold

<$link>This should not be bold</$link>

''<$link>This should be bold</$link>''

</div>
</strong>

//[[Internal link italics|example in italics]]// - It's italic also when the tiddler exists

''[[Internal link bold|example in bold]]'' - it's italic too, if tiddler does not exist
```

Jermolene:
Hi @pmario I am not convinced that the new `tc-emphasis`, `tc-strong` and `tc-italic` classes are valuable enough for the cruft that they generate. It looks like #8162 could be fixed without introducing these classes.

pmario:
I did not create this PR to "kill time". There where and are posts in the Talk forum, where users do have problems to assign colours to emphasised wikitext elements.

Users use "bold", "italic" and others for a reason, but we do not make them accessible with CSS easily.

This PR tries to change that in a straight forward way. CSS is complicated if users need to know specificity. With the classes here in the PR no additional knowledge about CSS specificity is needed. 

`tc-emphasis` is a way to globally assign CSS rules to all wikitext formatting rules at once.
`tc-strong`, `tc-italic` ... are more specializes for only 1 category. 

-----

Discussions at Talk about formatting 

- https://talk.tiddlywiki.org/t/formatting-links-e-g-adding-emphasis/9661
- https://talk.tiddlywiki.org/t/text-format-bold-and-italic-nok/7010

There are more, but I cannot find them atm.

Jermolene:
Hi @pmario thanks for reminding me. I'm happy to merge this.

Leilei332:
Looks like this PR is causing some test to fail. More tests should be updated to make it pass.

Jermolene:
Thanks @Leilei332 

While the tests are easy to fix, on reflection I would like to revert this PR so that we can experiment a little more.

The idea I am exploring is to make the element widget look at a variable `tv-add-emphasis-classes` (or something) to determine whether to add these classes. Thus wikitext authors would opt-in to the new behaviour if required. This might also involve adding a generic mechanism for setting global variables.

PR #8170, opened 2024-04-27
Improve Polish date translation

sarna:
I provided reasoning for bigger changes in commit messages.

One question: is Sunday always the first day of the week in TiddlyWiki, or does it depend on locale? In Poland, the week starts with Monday; however, in all translations I checked Sunday was the 0th day.


Jermolene:
Hi @sarna and welcome to the project. With apologies for the bureaucracy, please could you prepare a separate PR to the 'tiddlywiki-com' branch with your signature for the Contributor License Agreement (see [contributing.md](https://github.com/Jermolene/TiddlyWiki5/blob/master/contributing.md)).

> is Sunday always the first day of the week in TiddlyWiki, or does it depend on locale? In Poland, the week starts with Monday; however, in all translations I checked Sunday was the 0th day.

Yes, I think TiddlyWiki does indeed always assume that Sunday is day zero. I am not sure that we can fix it, though, without breaking backwards compatibiltiy.

Jermolene:
Thanks @sarna!

PR #8020, opened 2024-02-29
feat: new icon for default layout

linonetwo:
Fix concern in https://github.com/Jermolene/TiddlyWiki5/pull/7197#issuecomment-1971377584

I also add `tags: $:/tags/Layout` to default layout, so other plugins can use `tag[$:/tags/Layout]]` to get full list. And make storyview layout equal to other layouts.

But storyview layout is still the most important layout (we even have 3 different view mode for it), so it should have an outstanding icon, for layout plugins to provide a button that back to storyview layout.

I also suggest there can be a layout implementation guideline, that includes

1. layout should have at least a button including `$:/core/images/default-layout` or `$:/core/images/layout-button`, so a user can go back to the default layout on mobile
2. shouldn't add layout's shortcut button to page control, instead, use plugins typed "Layout launcher" to launch them. 

I think a simplest layout launcher is a toolbar like Page Control, but only shows layouts. I've developed at least 4 layout plugins, I sometimes imagine an Android launcher or Windows StartMenu for them.

PR #8172, opened 2024-04-29
Rename default layout to "Standard Layout"

linonetwo:
https://talk.tiddlywiki.org/t/what-is-the-name-of-default-layout-is-it-story-river/9296/2

Only change filename and translation, and `$:/core/images/default-layout` which is not yet released.

`$:/core/ui/PageTemplate` is untouched, it will influence too many things. This could be a todo for large version bump.


linonetwo:
@BramChen Note that `~Tiddlywiki` is translated to `太微` which is simillar to pronouncation in Chinese. I may apply this change to other translations in the future. But I'm not sure if this pronouncation also apply to Hong Kong Cantonese and Taiwan Hokkien, so I'm not gonna modify it. I will only modify zh-Hans in this case.

saqimtiaz:
@linonetwo thank you, I agree that renaming the "default layout" to "standard layout" would be helpful. However, renaming the images is problematic as it can break user code where the old tiddler titles are used. Could you please modify the PR to only change the layout name?

linonetwo:
@saqimtiaz That image was added during this prerelase, so won't break anything other than my plugins (I push the https://github.com/Jermolene/TiddlyWiki5/pull/8020 so I'm using it immediately), and I will update my plugins.

saqimtiaz:
Thank you @linonetwo. 

@Jermolene this looks good to be merged.

Jermolene:
Great thanks @linonetwo @saqimtiaz

PR #8175, opened 2024-05-03
table parser add data attributes for easier styling

pmario:
## Main Goal TLDR;

Having the `data-addr` attribute on every cell, it will allow custom js-macros to create very specific STYLEs to make it easy for users to manually style single cells eg: `<<tableBG green>>`, without the need to count rows or columns, which is needed at the moment.

------

This PR adds some new attributes to tables created with wikitext. This makes it easier to style columns and specific cells.

- `oddCol` or `evenCol` classes are added to every table-cell. 
  - colCount starts with 1
- `data-row` attribute is added to every TR element
  - rowCount starts with 0 for compatibility reasons and
  - TH element usually is row 0
- `data-col` attribute is added to every cell
- `data-addr` attribute added to every cell. The format is `<rowCount>/<colCount>`

This allows us to:

- Style a alternating columns based on CSS `.oddCol` and `.evenCol`
- Style a specific column based on CSS `[data-col="x"]` selector
- Style a specific row based on CSS `[data-row="y"]` selector
- Style a specific cell with CSS `[data-addr="row/col"]`

Related Talk threads:

- https://talk.tiddlywiki.org/t/coloring-table-cells-with-css/7430
- https://talk.tiddlywiki.org/t/highlight-the-background-color-of-any-cell-with-duplicate-content-in-any-row-or-column-in-a-table/9538
- https://talk.tiddlywiki.org/t/change-the-background-color-of-one-single-cell-easily/9751

### DIFF
--- a/core/modules/parsers/wikiparser/rules/table.js
+++ b/core/modules/parsers/wikiparser/rules/table.js
@@ -21,11 +21,13 @@ exports.init = function(parser) {
 	this.matchRegExp = /^\|(?:[^\n]*)\|(?:[fhck]?)\r?(?:\n|$)/mg;
 };
 
-var processRow = function(prevColumns) {
+var processRow = function(prevColumns,options) {
+	options = options || {};
 	var cellRegExp = /(?:\|([^\n\|]*)\|)|(\|[fhck]?\r?(?:\n|$))/mg,
 		cellTermRegExp = /((?:\x20*)\|)/mg,
 		tree = [],
 		col = 0,
+		colCount,
 		colSpanCount = 1,
 		prevCell,
 		vAlign;
@@ -103,9 +105,12 @@ var processRow = function(prevColumns) {
 			// Record information about this cell
 			prevCell = cell;
 			prevColumns[col] = {rowSpanCount:1,element:cell};
+			// column count is 1 based
+			colCount = col + 1;
 			// Check for a colspan
 			if(colSpanCount > 1) {
 				$tw.utils.addAttributeToParseTreeNode(cell,"colspan",colSpanCount);
+				colCount = colCount - 1;
 				colSpanCount = 1;
 			}
 			// Parse the cell
@@ -121,6 +126,9 @@ var processRow = function(prevColumns) {
 			}
 			// Move back to the closing `|`
 			this.parser.pos--;
+			$tw.utils.addClassToParseTreeNode(cell,colCount%2 ? "oddCol" : "evenCol");
+			$tw.utils.addAttributeToParseTreeNode(cell,"data-col",colCount);
+			$tw.utils.addAttributeToParseTreeNode(cell,"data-addr",options.rowCount + "/" + colCount);
 		}
 		col++;
 		cellRegExp.lastIndex = this.parser.pos;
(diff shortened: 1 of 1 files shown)


pmario:
The best way to test is

```
| A | B | C | D |h
|Cell1 |Cell2 |Cell3 |Cell4 |
|Cell5 |Cell6 |Cell7 |<|
|Cell5 |~|Cell7 |Cell8 |
|>|Cell9 |Cell10 |Cell11 |
```

pmario:
@Jermolene -- bump

TiddlyTweeter:
> This allows us to:
> 
> * Style a alternating columns based on CSS `.oddCol` and `.evenCol`
> * Style a specific column based on CSS `[data-col="x"]` selector
> * Style a specific row based on CSS `[data-row="y"]` selector
> * Style a specific cell with CSS `[data-addr="row/col"]`

Looks very good!

The Table parser in TW is very impressive already.
Adding these CSS support points would really take to a new level IMO.

Jermolene:
Thanks @pmario the resulting HTML is a lot more complex than before. I appreciate that we don't usually worry about the size of generated HTML but this is a lot more overhead for every table rendered in wikitext that would in practice only be relevant for a small number of users/installations. I feel that this level of markup should be opt-in via a new flag.

```
<table>
	<tbody>
		<tr class="evenRow" data-row="0">
			<th align="left" class="oddCol" data-col="1" data-addr="0/1">
				Cell1
			</th>
			<th align="left" class="evenCol" data-col="2" data-addr="0/2">
				Cell2
			</th>
		</tr>
		<tr class="oddRow" data-row="1">
			<td align="left" class="oddCol" data-col="1" data-addr="1/1">
				Cell3
			</td>
			<td align="left" class="evenCol" data-col="2" data-addr="1/2">
				Cell3
			</td>
		</tr>
	</tbody>
</table>

```

pmario:
In the sidebar resizer, you mentioned, that it would be nice to have a possibility to resize table columns. But there is no way at the moment to do so. This PR will give us the possibility.

Modifying `[data-col="1"]` CSS is exactly the way to do this. `[data-row=...]` may even be able to adjust row height.

Jermolene:
@pmario as I said, I don't think it is reasonable to add all of this extra markup to every single table generated by TiddlyWiki, with no way to prevent it.

Introducing a new table flag would be a clumsy way to achieve that. An alternative approach that wouldn't require manual markup would be to refactor the table parser to use a `<$table>` widget. The widget could have runtime logic to look at a configuration variable like `tv-table-markup` to determine whether to add the decorations.

pmario:
>Introducing a new table flag would be a clumsy way to achieve that. An alternative approach that wouldn't require manual markup would be to refactor the table parser to use a <$table> widget. The widget could have runtime logic to look at a configuration variable like tv-table-markup to determine whether to add the decorations.

I was thinking about a new utility class. Eg: `tc-table-verbose`
We already have some [utility classes for tables](https://tiddlywiki.com/#Utility%20Classes:%5B%5BUtility%20Classes%5D%5D%20%5B%5BTables%20in%20WikiText%20CSS%20Utility%20Classes%5D%5D), that define the visual behaviour of the table. If `tc-table-verbose` is available, it is easy to identify for users.

pmario:
@Jermolene -- I did use `tc-table-data` as a new utility class.

Even if we would have used a variable `tv-xyz`, we would have needed an additional class anyway, since we need an easy way to identify tables that have extended HTML data-attributes. So `tc-table-data` is the perfect trigger, which can be used at **parsing time**.

------

I do like the idea of a new `<$table>`-widget but imo that would need a complete rewrite of the table parser, which is a much bigger project. IMO that's way out of the behaviour I wanted to achieve with this PR.

I personally do not use TW wikitext tables. I mainly use them for TW documentation. 

For my own wikis I do create dynamic tables which transclude tiddler content.

Jermolene:
> @Jermolene -- I did use `tc-table-data` as a new utility class.

I think this is even more of a hack. We don't have any other situations where the addition of a CSS class triggers special behaviour. The mechanism introduced here is not extensible. This mechanism requires tables to be marked at authoring time, with no dynamic possibilities.

I can see the argument that this PR is simpler than the approach I suggested, but it loads the core with even more cruft that will get in the way in the future.

PR #8176, opened 2024-05-04
Fix 8166 html tag validation

### LINKED ISSUE #8166: [BUG] Internal JavaScript Error on giving invalid "tag" attribute values for certain widgets
### Describe the bug

~The code `<$droppable <$macrocall $name="tag" tag="$:/tags/Global"/>` results in the Internal JavaScript Error `Uncaught InvalidCharacterError: Failed to execute 'createElement' on 'Document': The tag name provided ('$:/tags/Global') is not a valid name.`~

Edit: Certain widgets that support overriding the default HTML tag for the widget cause uncaught JS errors when given an invalid value for the "tag" attribute.  For example, `<$edit-text tag="$"/>` or `<$droppable tag="$"/>`.

### Expected behavior

JS error is caught and a syntax error should be displayed

### To Reproduce

1. Go to tiddlywiki.com or open an Empty wiki
2. Open new tiddler
3. Open side-by-side preview pane
4. ~Paste  `<$droppable <$macrocall $name="tag" tag="$:/tags/Global"/>` into the editor~
    Edit: simply paste `<$droppable tag="$"/>`
5. Internal JavaScript Error message pops up

### Screenshots

image

### TiddlyWiki Configuration

- Version v5.3.3
- Saving mechanism TiddlyDesktop, or none

- OS: Windows 10
- Browser: Chromium-based (at least Brave and TiddlyDesktop)

### Additional context

Edit: The below behavior was part of how this was discovered and the cause determined, but now that the example of `<$edit-text tag="$"/>` is known, the below examples are no longer needed to understand the issue.

* The error also occurs if calling the macro using the  `<$transclude>` widget ( `<$droppable <$transclude $variable="tag" tag="$:/tags/Global"/>`)
* The error does *not* occur when using the wikitext macrocall syntax (`<$droppable <<tag "$:/tags/Global">>`)
* The error does *not* occur if the tag name only includes alphabetic characters (e.g., in the documentation, `<$droppable <$macrocall $name="tag" tag="Macros"/>`); it appears the dollar sign `$` in the tag name is partially the cause of the issue.

Note:  The original intention was to complete the full `<$droppable>` widget, adding it around an existing `<$macrocall>` widget.  But when typing `<$droppable>` one character at a time, you get to `<$droppable` and the error occurs before you can finish off with the `>`.  (The point being that this isn't merely what you can find by experimenting with invalid syntax in attempt to break TW; it first happened to me in the natural course of trying to write valid syntax with the preview pane open.)

pmario:
This PR fixes #8166

- #8166

It adds a new function to `$tw.utils.makeTagNameSafe(tag, defaultTag)`

- It checks for the existence of the tag and if it is a safe tag. 
- If defaultTag is undefined it uses SPAN as default

It replaces all the hardcoded checks using `if(complex logi)` with a simple to read function call

### DIFF
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -49,9 +49,6 @@ function FramedEngine(options) {
 	this.widget.domNodes.push(this.iframeNode);
 	// Construct the textarea or input node
 	var tag = $tw.utils.makeTagNameSafe(this.widget.editTag, "input");
-	// if($tw.config.htmlUnsafeElements.indexOf(tag) !== -1) {
-	// 	tag = "input";
-	// }
 	this.domNode = this.iframeDoc.createElement(tag);
 	// Set the text
 	if(this.widget.editTag === "textarea") {
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -48,7 +48,7 @@ function FramedEngine(options) {
 	this.iframeDoc.body.style.padding = "0";
 	this.widget.domNodes.push(this.iframeNode);
 	// Construct the textarea or input node
-	var tag = $tw.utils.isTagNameSafe(this.widget.editTag, "input");
+	var tag = $tw.utils.makeTagNameSafe(this.widget.editTag, "input");
 	// if($tw.config.htmlUnsafeElements.indexOf(tag) !== -1) {
 	// 	tag = "input";
 	// }
--- a/core/modules/editor/engines/framed.js
+++ b/core/modules/editor/engines/framed.js
@@ -48,10 +48,10 @@ function FramedEngine(options) {
 	this.iframeDoc.body.style.padding = "0";
 	this.widget.domNodes.push(this.iframeNode);
 	// Construct the textarea or input node
-	var tag = this.widget.editTag;
-	if($tw.config.htmlUnsafeElements.indexOf(tag) !== -1) {
-		tag = "input";
-	}
+	var tag = $tw.utils.isTagNameSafe(this.widget.editTag, "input");
+	// if($tw.config.htmlUnsafeElements.indexOf(tag) !== -1) {
+	// 	tag = "input";
+	// }
 	this.domNode = this.iframeDoc.createElement(tag);
 	// Set the text
 	if(this.widget.editTag === "textarea") {
@@ -200,7 +200,7 @@ FramedEngine.prototype.handleFocusEvent = function(event) {
 Handle a keydown event
  */
 FramedEngine.prototype.handleKeydownEvent = function(event) {
-	if ($tw.keyboardManager.handleKeydownEvent(event, {onlyPriority: true})) {
+	if($tw.keyboardManager.handleKeydownEvent(event, {onlyPriority: true})) {
 		return true;
 	}
 
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -389,10 +389,10 @@ exports.querySelectorAllSafe = function(selector,baseElement) {
 /*
 Make sure HTML tag names are valid
 */
-exports.isTagNameSafe = function(tag,defaultTag) {
+exports.makeTagNameSafe = function(tag,defaultTag) {
 	defaultTag = defaultTag || "SPAN";
-	// This implements standard DOM elements: https://html.spec.whatwg.org/#syntax-tag-name
-	// It does _not_ implement Custom Elements: https://html.spec.whatwg.org/#valid-custom-element-name
+	// This implements a check for standard DOM elements: https://html.spec.whatwg.org/#syntax-tag-name
+	// It does _not_ deal with Custom Elements: https://html.spec.whatwg.org/#valid-custom-element-name
 	var regexp = /[a-zA-Z0-9]/g;
 	if(tag && tag.match(regexp) && $tw.config.htmlUnsafeElements.indexOf(tag) === -1) {
 		return tag;
--- a/core/modules/utils/dom/dom.js
+++ b/core/modules/utils/dom/dom.js
@@ -289,7 +289,7 @@ exports.copyToClipboard = function(text,options) {
 	var succeeded = false;
 	try {
 		succeeded = document.execCommand("copy");
-	} catch (err) {
+	} catch(err) {
 	}
 	if(!options.doNotNotify) {
 		$tw.notifier.display(succeeded ? "$:/language/Notifications/CopiedToClipboard/Succeeded" : "$:/language/Notifications/CopiedToClipboard/Failed");
@@ -306,8 +306,8 @@ Collect DOM variables
 */
 exports.collectDOMVariables = function(selectedNode,domNode,event) {
 	var variables = {},
-	    selectedNodeRect,
-	    domNodeRect;
+		selectedNodeRect,
+		domNodeRect;
 	if(selectedNode) {
 		$tw.utils.each(selectedNode.attributes,function(attribute) {
 			variables["dom-" + attribute.name] = attribute.value.toString();
@@ -324,7 +324,7 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 			variables["tv-popup-coords"] = Popup.buildCoordinates(Popup.coordinatePrefix.csOffsetParent,nodeRect);
 
 			var absRect = $tw.utils.extend({}, nodeRect);
-			for (var currentNode = selectedNode.offsetParent; currentNode; currentNode = currentNode.offsetParent) {
+			for(var currentNode = selectedNode.offsetParent; currentNode; currentNode = currentNode.offsetParent) {
 				absRect.left += currentNode.offsetLeft;
 				absRect.top += currentNode.offsetTop;
 			}
@@ -366,7 +366,7 @@ exports.collectDOMVariables = function(selectedNode,domNode,event) {
 };
 
 /*
-Make sure the CSS selector is not invalid
+Make sure the CSS selector is valid
 */
 exports.querySelectorSafe = function(selector,baseElement) {
 	baseElement = baseElement || document;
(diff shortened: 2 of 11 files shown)


core/modules/widgets/element.js
```diff
@@ -31,9 +31,7 @@ ElementWidget.prototype.render = function(parent,nextSibling) {
 	this.computeAttributes();
 	// Neuter blacklisted elements
 	this.tag = this.parseTreeNode.tag;
-	if($tw.config.htmlUnsafeElements.indexOf(this.tag) !== -1) {
-		this.tag = "safe-" + this.tag;
-	}
+	this.tag = $tw.utils.makeTagNameSafe(this.tag, "safe" + this.tag);
```
Jermolene, line 34: The original code here was attempting to sanitise the tag name provided by the user. This is in contrast to the behaviour of the helper function which is to fall back to a default if the provided tag name is not valid.

I think one approach or there other should be used consistently across the board. I'd favour the sanitising approach.

core/modules/editor/engines/framed.js
```diff
@@ -200,7 +197,7 @@ FramedEngine.prototype.handleFocusEvent = function(event) {
 Handle a keydown event
  */
 FramedEngine.prototype.handleKeydownEvent = function(event) {
-	if ($tw.keyboardManager.handleKeydownEvent(event, {onlyPriority: true})) {
+	if($tw.keyboardManager.handleKeydownEvent(event, {onlyPriority: true})) {
```
Jermolene, line 200: There's quite a few whitespace changes in this PR that make it hard to follow. I'd welcome separate PRs to handle whitespace changes.

core/modules/editor/engines/framed.js
```diff
@@ -48,10 +48,7 @@ function FramedEngine(options) {
 	this.iframeDoc.body.style.padding = "0";
 	this.widget.domNodes.push(this.iframeNode);
 	// Construct the textarea or input node
-	var tag = this.widget.editTag;
-	if($tw.config.htmlUnsafeElements.indexOf(tag) !== -1) {
-		tag = "input";
-	}
+	var tag = $tw.utils.makeTagNameSafe(this.widget.editTag, "input");
```
Jermolene, line 51: There's unneeded whitespace before the `"input"`.

core/modules/utils/dom/dom.js
```diff
@@ -386,4 +386,19 @@ exports.querySelectorAllSafe = function(selector,baseElement) {
 	}
 };
 
+/*
+Make sure HTML tag names are valid
+*/
+exports.makeTagNameSafe = function(tag,defaultTag) {
+	defaultTag = defaultTag || "SPAN";
+	// This implements a check for standard DOM elements: https://html.spec.whatwg.org/#syntax-tag-name
+	// It does _not_ deal with Custom Elements: https://html.spec.whatwg.org/#valid-custom-element-name
```
Jermolene, line 395: I think we should allow for custom elements.

Jermolene (commented):
Thanks @pmario just some minor points

core/modules/editor/engines/framed.js
pmario: It think, they have been automatically changed by VSCode on file save because of the global eslint settings. This behaviour started for me as eslint was defined under devDependencies.

core/modules/widgets/element.js
pmario: The logic does only sanitise the tag if it is not valid. That's exactly, what the new code does

core/modules/utils/dom/dom.js
Jermolene: Hi @pmario have you had a chance to look at this?

pmario:
@Jermolene -- There is a problem with the SPEC - https://html.spec.whatwg.org/#prod-potentialcustomelementname allows the following unicode characters: 

>PCENChar ::=
    "-" | "." | [0-9] | "_" | [a-z] | #xB7 | [#xC0-#xD6] | [#xD8-#xF6] | [#xF8-#x37D] | [#x37F-#x1FFF] | [#x200C-#x200D] | [#x203F-#x2040] | [#x2070-#x218F] | [#x2C00-#x2FEF] | [#x3001-#xD7FF] | [#xF900-#xFDCF] | [#xFDF0-#xFFFD] | [#x10000-#xEFFFF]

99% of them can be converted to a js regexp, but the last range `[#x10000-#xEFFFF]` causes a problem. JS unicode can only handle 4byte unicode ranges. 

For 10000 and up we would need a poyfill see: 

- https://stackoverflow.com/questions/15389497/how-to-match-high-value-unicode-characters-using-a-regex

Which leads to a ES6 polyfill 

- https://github.com/paulmillr/es6-shim/blob/8e570a4b425a80f9b13ff027dbd28d65f201a319/es6-shim.js#L171-L183

I think this would cause a significant performance problem, since we do test every HTML element for validity. 

**I think** we could implement it **without** the last range or leave it as it is in this PR. 

Unicode table with easy to find ranges

- https://symbl.cc/en/unicode-table/

pmario:
@Jermolene -- I did a bit more investigations. The main info I found about custom web-components was that web-components need to have a hyphen in the name. Almost none of the docs except the spec itself mentioned the full regexp. 

So I thought about a validit check like this: 

1. check the string, if it is a valid html tag -- using this spec: https://html.spec.whatwg.org/#syntax-tag-name
2. extend 1. and allow hyphens: `-` 
3. implement a forbidden list according to https://html.spec.whatwg.org/#valid-custom-element-name
  3.1 `exports.htmlForbiddenTags = "annotation-xml,color-profile,font-face,font-face-src,font-face-uri,font-face-format,font-face-name,missing-glyph".split(",");` 
5. if 1., 2. and 3. pass return the HTML tag as a valid tag -- That should be fast and cover 90% of html tags users will use. 

---

6. if 5. does not match check for **invalid character ranges** up to `\uFFFF`. So we can detect problems in a range JS can handle. 
7. We **assume** that everything out of js-range is valid, which would be OK for `\u10000-\uEFFFF` according to the spec.

@Jermolene -- what do you think?

core/modules/widgets/element.js
pmario: The function allows developers to define a default tag. The tag **and** the default tag are checked if they are safe. 
```
	// Check for unsafe tag and unsafe defaultTag
	if($tw.config.htmlUnsafeElements.indexOf(result.toLowerCase()) !== -1) {
		result = ($tw.config.htmlUnsafeElements.indexOf(defaultTag.toLowerCase()) !== -1) ? "safe-" + defaultTag : defaultTag;
	}
```

core/modules/utils/dom/dom.js
pmario: That's the comment how the function works. So custom elements should be allowed.

```
/*
Sanitize HTML tag- and custom web component names
	1. Check the string, if it is a valid html tag - using this spec: https://html.spec.whatwg.org/#syntax-tag-name
		1.1 Tag names are "case insensitive"
    2. Extend 1. and allow hyphens: "-"
		2.1 Browsers allow "AA-AA", so do we. Be aware there may be styling problems

	3. Sanitize input parameters - spec: https://html.spec.whatwg.org/#valid-custom-element-name
    4. Implement a forbidden list: exports.htmlForbiddenTags - see: $:/core/modules/config.js
    5. Check function parameters for invalid character ranges up to \uFFFF. This detects problems in a range JS RegExp can handle
    6. We assume that everything out of js RegExp-range is valid, which is OK for \u10000-\uEFFFF according to the spec

	Unicode overview: https://symbl.cc/en/unicode-table/
*/
```

pmario:
@saqimtiaz @Jermolene ... Please review

core/modules/utils/dom/dom.js
```diff
@@ -392,3 +392,44 @@ exports.querySelectorAllSafe = function(selector,baseElement) {
 		console.log("Invalid selector: ",selector);
 	}
 };
+
+/*
+Sanitize HTML tag- and custom web component names
+	1. Check the string, if it is a valid html tag - using this spec: https://html.spec.whatwg.org/#syntax-tag-name
```
Jermolene: Rather than having this long comment block, couldn't the inline comments be expanded to cover any strictly necessary subtleties?

Jermolene:
Thanks @pmario

pmario:
@Jermolene ... Can you revert this one. I wanted to add some tests, but forgot to make the PR a draft. I found an edge-case, that slipped through. I'll create a new PR with some tests.

saqimtiaz:
@pmario I have reverted this in #9374 

What is the edge case that does not work?

pmario:
Stupid mistake. I did test the initial error `<$droppable tag="$"/>` again, which was fixed, while in progress. It seems I broke it and forgot to test it again. :/ 

I am currently creating tests, because I wanted to test for 100% code coverage for the dom.js API and the widgets. IMO getting HTML elements right is important and the unicode ranges are not really human readable.

pmario:
@Jermolene @saqimtiaz -- As I did create this PR in May 24, I did create [these config elements](https://github.com/pmario/TiddlyWiki5/blob/60009ba2bb560dad0fd5d0a7930235376d4ee0c1/core/modules/config.js#L43-L47), which are rather complex. 

It is used to validate HTML-elements and custom-element names. 

```
exports.htmlCustomPrimitives = {
	prefix: "[a-z]",
	validPCENChar: ".|[0-9]|_|[a-z]|\xB7|[\xC0-\xD6]|[\xD8-\xF6]|[\u00F8-\u037D]|[\u037F-\u1FFF]|[\u200C-\u200D]|[\u203F-\u2040]|[\u2070-\u218F]|[\u2C00-\u2FEF]|[\u3001-\uD7FF]|[\uF900-\uFDCF]|[\uFDF0-\uFFFD]",
	sanitizePCENChar: "[\x00-\x2C]|\x2F|[\x3A-\x40]|[\x5B-\x60]|[\x7B-\xB6]|[\xB8-\xBF]|\xD7|\xF7|\x37E|[\u2000\u200B]|[\u200E-\u203E]|[\u2041-\u206F]|[\u2190-\u2BFF]|[\u2FF0-\u3000]|[\uD800-\uF8FF]|[\uFDD0-\uFDEF]|[\uFFFE-\uFFFF]"
	};
```

It seems in the meantime they did change the spec to something like this: https://dom.spec.whatwg.org/#valid-element-local-name which is obviously much simpler to understand. It covers ranges, that the Javascript parser does not use atm. but anyway. 

```
/^(?:[A-Za-z][^\0\t\n\f\r\u0020/>]*|[:_\u0080-\u{10FFFF}][A-Za-z0-9-.:_\u0080-\u{10FFFF}]*)$/u
```

**My question now is** -- Should I use the **simple** syntax, which is less restrictive, but may cause RSODs 

or the SPEC from a year ago. ???

Jermolene:
> **My question now is** -- Should I use the **simple** syntax, which is less restrictive, but may cause RSODs

What are the situations where an RSOE can occur? I would have expected that a goal of the tag validation would be to avoid RSOEs.

PR #8179, opened 2024-05-05
Fix get variable info params

pmario:
This PR will fix a problem in widget.getVariableInfo(). It does not return `params` for procedures and widgets. 

The PR contains new tests, that check the params and the `text` and `value` for functions (which is unique to functions)


Jermolene:
Thanks @pmario

Issue #8181, opened 2024-05-07
Tag-pill import (drag and drop) should include the tiddler at tag's home node, if it exists


springerspandrel:
Dragging a tag pill from one wiki to another currently serves to copy over all the tiddlers _under_ that tag. However, it does not copy over the tag tiddler itself — the “home base” node for the tag — even when there is a tiddler there at that logically important location. 

*Problem*: This default behavior is a problem given that the tiddler at this location (home / root / brain / key / core / hub for tag) has increasingly become a natural and appropriate place to organize all sorts of germane info and resources. For example, this tiddler is the best place to include an explanation for a community-shared solution informally bundled under a tag, an teaching overview or note-to-self of why a tag matters, or substantial info about a type or genre or category within the wiki, perhaps with a custom overview list or table tailored to the fields associated with tiddlers under this tag. Further, order information for a tag is stored in this tiddler's list field — and leaving that list order behind when those details are vital (as for example, in a cascade-based solution) can cause confusion and headaches, especially for newcomers. Some people also care about retaining assigned tag color across multiple wikis, and are surprised when dragging a tag pill doesn't retain the color (since it's stored in the tiddler for the tag, which is by default left behind).

*Solution*: I'd like to see the tag-pill's default drag-and-drop behavior extended, so that the tag's home-base tiddler (if it exists) is included in the tag-pill's payload. I'd hope to see the tag's main tiddler displayed above all its "children" if that's easy, so the list looks a bit like it does on the tag-pill's default drop-down. (If the import process can't easily pull it to top of list, that's minor.) As with any import, users will have an opportunity to exclude this tiddler (or any other) at the arriving payload-confirmation dialog, with visual flag for conflicts.

*Alternatives?*: Workarounds include (a) custom filter-pill-style solutions and (b) self-referential tagging. The first (a) is a lot of work to accomplish something that's straightforward (and should be available to beginning users); the second (b) is bound to create mischief for projects that want simple tag-hierarchy structures, and to undermine other powerful uses of tags (such as to serve as criteria for view templates, cascades, etc.).

CrossEye:
Note that there is more discussion on this issue on [talk.tiddlywiki](https://talk.tiddlywiki.org/t/9778), including [a workaround](https://talk.tiddlywiki.org/t/9778/23) that might offer a useful suggestion for a fix:

[...]

PR #8182, opened 2024-05-08
add tag-tiddler to tag-pill drag & drop import, add hidden setting for TagPillDragFilter

pmario:
This PR fixes: #8181

- #8181

- It adds the tag-tiddler to tag-pill drag & drop import. 
- It adds a hidden setting for TagPillDragFilter
- It adds some relatively detailed docs about the hidden setting, what to do to revert to the "old" behaviour. 
  - The old behaviour was buggy, since it also exports "Draft tiddlers" even if they are not shown in the tag-dropdown.


pmario:
@Jermolene -- bump

Jermolene:
Hi @pmario as I mentioned [here](https://github.com/TiddlyWiki/TiddlyWiki5/issues/8181#issuecomment-2098977770), on balance I think that we should not change the default behaviour.

pmario:
@Jermolene -- I did change the default, so it dos **not** include the tag tiddler itself. Also updated the docs.

Jermolene:
Thanks @pmario

By the way, it might be useful to make a plugin that contains a UI for the hidden settings. The motivation for keeping these settings hidden is generally to avoid bloating the core with the relevant language strings, so putting the UI in a plugin would be reasonable.

PR #8184, opened 2024-05-09
Clarify handling of title lists in ActionListopsWidget documentation

yaisog:
Based on this discussion on Talk TW
https://talk.tiddlywiki.org/t/possible-error-in-action-listops-widget-documentation/9762
the _Note on subfilter expressions_ has been corrected so that the given example works as intended.
Also, an info box was added describing how titles with spaces or multiple titles can be handled, with references for further reading.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
+++ b/editions/tw5.com/tiddlers/widgets/ActionListopsWidget.tid
@@ -1,6 +1,6 @@
 caption: action-listops
 created: 20141025120850184
-modified: 20240509074011368
+modified: 20240509135041526
 myfield: 
 tags: ActionWidgets Widgets
 title: ActionListopsWidget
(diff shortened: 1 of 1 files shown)


bluepenguindeveloper:
"To remove titles containing spaces, _or to remove multiple titles_, the individual titles must be wrapped in double square brackets, usually via a variable assignment."

The "or to remove multiple titles" part isn't correct; if none of the titles in the list contain spaces, then `+[remove[Item1 Item2]]` works because that's a valid title list containing "Item1" and "Item2".

Additionally, elsewhere in the documentation, the term "soft parameter" is used to denote variable or transcluded filter parameters, usually linking here: https://tiddlywiki.com/#Filter%20Parameter.  So perhaps the above portion could be restated as:

"To remove titles containing spaces, a [soft parameter](https://tiddlywiki.com/#Filter%20Parameter) must be used to enable the individual titles to be wrapped in double square brackets as required by the title list format."

And the [extended listops operators page](https://tiddlywiki.com/#Extended%20Listops%20Filters) seems like another good place to link to for more details.  It says: "These filters are best used with a reference to an array, stored in a field or data index elsewhere in the wiki (they may be used with a simple list of items, provided the items do not include white space.)"

yaisog:
@bluepenguindeveloper Thanks for your input. I have rephrased the part about multiple titles to explicitly mention that it applies to titles with spaces. If one wanted to be pedantic, this would mention that only one of multiple titles needed to contain a space to make a soft parameter necessary.
I also added a link to the _Filter Parameters_ tiddler for the explanation of soft parameters.
I did not add the link to the extended listops operators, because this tiddler refers to a parameter "array", which might be confusing in this situation. Since I don't want to update all the listops pages in the context of this PR, I would defer mentioning that tiddler for now.

Jermolene:
Hi @yaisog since this is docs only it should really target the tiddlywiki-com branch, but since we're so close to the next release I'll just merge this to master now.

yaisog:
> it should really target the tiddlywiki-com branch

Sorry about that, sometimes I don't think of these things...

Issue #8183, opened 2024-05-09
[BUG]  Blockquote parsing fails on example


joshuafontany:
### Describe the bug

The Wiki Parser Rule for QuoteBlock does not parse the wikitext given in the example comments:
```
/*\
title: $:/core/modules/parsers/wikiparser/rules/quoteblock.js
type: application/javascript
module-type: wikirule

Wiki text rule for quote blocks. For example:

	<<<.optionalClass(es) optional cited from
	a quote
	<<<
	
	<<<.optionalClass(es)
	a quote
	<<< optional cited from

Quotes can be quoted by putting more <s


	<<<
	Quote Level 1
	
	<<<<
	QuoteLevel 2
	<<<<
	
	<<<
```

### Expected behavior

Block Quotes are parsed as expected.

### To Reproduce

1. Go to https://tiddlywiki.com
2. Click on New Tiddler
3. Paste the wikitext from the quoteblock.js header comments:
	<<<.optionalClass(es) optional cited from
	a quote
	<<<
	
	<<<.optionalClass(es)
	a quote
	<<< optional cited from

### TiddlyWiki Configuration

- Version [e.g. v5.1.24]
- Saving mechanism [e.g. Node.js, TiddlyDesktop, TiddlyHost etc]
- Plugins installed [e.g. Freelinks, TiddlyMap]

pmario:
I think it works as expected.  The only problem is, that it should not be indented. Blockquotes have to start at the start of the line. Tabs at the "end quotes" are not allowed.

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/5ca0ed99-1170-4006-8122-d832d4685e16)

Jermolene:
Hi @joshua the comment there is out of date and misleading in two respects: the use of indentation, and the fact that the blockquote syntax does not support nesting. I think the entire comment can be reduced to "Wiki text rule for quote blocks"; these comments are not the place to explain the syntax.

pmario:
>[. . .] and the fact that the blockquote syntax does not support nesting. 

I was surprised, but it actually works. 

![image](https://github.com/Jermolene/TiddlyWiki5/assets/374655/492be915-67db-41ff-aba9-6cfe0e1f35bd)

joshuafontany:
Ack, it was late when I found this. I am re-writing =my VSCode-TW5-Syntax "grammar", and this jumped out at me now that I can colorize wkitext better. The actual bug is the other example:
```
[...]

PR #8186, opened 2024-05-12
Block Quotes - fix parsing error with spaces before reEndString

joshuafontany:
- Fixes parsing error when spaces/tabs appear before the end `<<<+`
- Clarifies docs as to what to do to "return to block parsing mode" after paragraphs inside block quotes.
- Clarifies similar docs in the "Hard Line Breaks" (tested inside of a block quote).
- Additional "Advanced Wikitest and Block Quotes" examples.

Fixes bug: https://github.com/Jermolene/TiddlyWiki5/issues/8183

### DIFF
--- a/core/modules/parsers/wikiparser/rules/quoteblock.js
+++ b/core/modules/parsers/wikiparser/rules/quoteblock.js
@@ -27,7 +27,6 @@ exports.parse = function() {
 	var reEndString = "^\\s*" + this.match[1] + "(?!<)";
 	// Move past the <s
 	this.parser.pos = this.matchRegExp.lastIndex;
-	this.parser.skipWhitespace({treatNewlinesAsNonWhitespace: true});
 	// Parse any classes, whitespace and then the optional cite itself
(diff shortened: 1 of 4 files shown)


core/modules/parsers/wikiparser/rules/quoteblock.js
```diff
@@ -3,30 +3,7 @@ title: $:/core/modules/parsers/wikiparser/rules/quoteblock.js
...
-```
-	<<<
-	Quote Level 1
-	
-	<<<<
-	QuoteLevel 2
-	<<<<
-	
-	<<<
-```
```
pmario, line 29: IMO this should not be removed. Just remove the indentation, then it should work, if copy/pasted to a tiddler.

joshuafontany: Thanks @pmario. @Jermolene Now that I have identified the bug as parsing tabs/whitespace & clarified the docs, should we keep or remove the example in the parser?

Jermolene:
Thanks @joshuafontany looks good

Issue #6847, opened 2022-07-29
Add support for comments within filters


Jermolene:
In order to help make complex filters more readable, I'd like to explore the idea of supporting comments within filters via several alternate syntaxes:

```
<<mymacro """
[tag[One]] // This is a single line comment
/* this is a potentially multiline comment */ [tag[Two]]
<!-- and so is this another comment --> [tag[Three]]
""">
```

The idea would be that comments would act as a filter run, but not have any observable effects. Thus the following would not be recognised as containing a comment:

```
<<mymacro """
[tag[/* not a comment */HelloThere]]
""">
```

It's worth noting that the custom function support in #6666 can also help with making filters more readable by giving subfilters meaningful names and parameters. So perhaps a different approach would be to decompose complex filters into a chain of custom functions, each of which could be self documenting if the names are chosen carefully.

kookma:
Good idea @Jermolene 
It happens many times to me where a complex filter needs to be commented and I had to add comments before or after!

I like something like below:

```
<$list filter="""[all[tiddlers+shadows]tag[$:/tags/Macro]] <!-- select all global macros -->
                 :map[{!!caption}!is[blank]else{!!title}]  <!-- return caption where possible -->
.
.
.
```

* I think one consistent form would suffice this. I propose `<!--`  to be consistent with current commenting syntax
* Simple is better, so for me, the below is confusing

```
<<mymacro """
[tag[/* not a comment */HelloThere]]
""">
```

Jermolene:
> Simple is better, so for me, the below is confusing

It is a contrived example. I'm trying to show that the `/* comment */` syntax is not recognised within an operand. And so the above is literally saying `give me everything tagged "/* not a comment */HelloThere"`.

rmunn:
If there's a chance that the comment syntax could overlap with tiddler titles / tags / whatever that people have already made, then one possible solution could be to require opting in to filter comment syntax via `\pragma filtercomments` (or `\pragma filtercomments on`) at the top of the file.

But by requiring comments to only start in the filter-run position of the syntax, I think you avoid that issue entirely. The only drawback I can see is that sometimes I might want to insert comments in between steps of a single complex run... and since runs cannot have whitespace inside them, it would be hard to have meaningful comments in there anyway. So if a run is so complex that I want comments inside it, I should prefix it with a multi-line comment (or several single-line ones) explaining the run in detail.

Jermolene:
[...]

PR #8190, opened 2024-05-15
add filter run prefix :comment or // or # for testing

pmario:
related to: **Add support for comments within filters** #6847

This PR adds 

- a verbose `:comment` filter run prefix and 
- 2 shortcuts: `#` and `//` 

There should be only 1 shortcut in the final version, that's why this is a draft PR

image

image

image

image

### DIFF
--- a/core/modules/filterrunprefixes/comment.js
+++ b/core/modules/filterrunprefixes/comment.js
@@ -1,0 +1,23 @@
+/*\
+title: $:/core/modules/filterrunprefixes/comment.js
+type: application/javascript
+module-type: filterrunprefix
+
+Intersection of sets.
+Equivalent to + filter run prefix.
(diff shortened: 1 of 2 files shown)


pmario:
@Jermolene -- What do you think. Should we add some more possibilities for testing purposes?

Jermolene:
Thanks @pmario I don't think this is the best approach; see my comment at https://talk.tiddlywiki.org/t/how-to-add-comments-to-filter-expressions/9816/17

pmario:
I did move the related posts at Talk to a new topic: https://talk.tiddlywiki.org/t/filter-comments-proposals/9830

pmario:
Was ruled out so close this one. #8191 will need more work

PR #8191, opened 2024-05-15
add // filter comment till linebreak

pmario:
This draft PR allows the following comment syntax.

- `//` ... comment start
- `\n` ... new-line as comment end

**Disadvantages:**

- Comments do force multiline filter syntax
- It's not backwards compatible. 

Related to PR:  **add filter run prefix :comment or // or # for testing** #8190

### DIFF
--- a/core/modules/filters.js
+++ b/core/modules/filters.js
@@ -70,7 +70,7 @@ function parseFilterOperation(operators,filterString,p) {
 		operator.operands = [];
 		var parseOperand = function(bracketType) {
 			var operand = {};
-			switch (bracketType) {
+			switch(bracketType) {
 				case "{": // Curly brackets
 					operand.indirect = true;
 					nextBracketPos = filterString.indexOf("}",p);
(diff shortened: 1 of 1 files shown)


yaisog:
The requirement for multiline filter syntax may be a significant drawback but it is only necessary when adding comments between two filter runs. A single-line filter expression can just have a comment at the end. That may cover a lot of use cases.

BTW, I very much welcome the effort of adding comments to filters. This is especially useful in (one-line) functions and macros/procedures.

pmario:
>... This is especially useful in (one-line) functions and macros/procedures.

I think for 1 line filters, it's enough to comment right above the filter itself. I personally would use filter comments only for really complex multi-line filters. Even if filter functions are cached, comments need to be parsed. So having comments outside the functions is a minimal performance benefit.

pmario:
@Jermolene -- Did you decide, which version of the comment you consider to merge, so I can prepare a real PR. 

- There is this one, which is not 100% backwards compatible and
- "add filter run prefix :comment or // or # for testing" https://github.com/Jermolene/TiddlyWiki5/pull/8190

pmario:
Since I did not get any more feedback on the other PR https://github.com/Jermolene/TiddlyWiki5/pull/8190 I'll activate this one for review

Jermolene:
Thanks @pmario I'm closing this in favour of #9055

PR #8195, opened 2024-05-18
Fixed issue with fakedome TW_Node inheritence

flibbles:
I discovered this. fakedom TW_Element and TW_TextNode don't inherit from TW_Node correctly, so nodeType couldn't be properly compared.

### DIFF
--- a/core/modules/utils/fakedom.js
+++ b/core/modules/utils/fakedom.js
@@ -42,7 +42,7 @@ var TW_TextNode = function(text) {
 	this.textContent = text + "";
 };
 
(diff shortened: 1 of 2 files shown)


flibbles:
@Jermolene, should be good to merge. I checked. The core doesn't use ELEMENT_NODE and TEXT_NODE when using fakedom, which is why this was missed, but the testing framework does.

Though it was also missed this because it does `expect(node.nodeType).toBe(node.ELEMENT_NODE);` Both were undefined.

Jermolene:
Ouch thanks @flibbles

PR #8196, opened 2024-05-18
Fixed issue where sjcl pollutes window scope

flibbles:
This is minor, and it probably bothers nobody except people who program uglifiers, but the way the sjcl library is made, it pollutes the window namespace with a bunch of uglified variables.

[There's actually a longstanding bug on this in the sjcl github](https://github.com/bitwiseshiftleft/sjcl/issues/364), but seeing as how sjcl hasn't been maintained in over 6 years, maybe we should just put the fix on our version of sjcl?

I get it if you think this isn't worth the fix. It's just a fix that's been sitting on my version of TW for a while, and I figured I'd offer to put it on master.


flibbles:
In case anyone was wondering what the pollution is, `window` in Tiddlywiki contains a the following obscure methods in global scope: A, B, C, D, E, F, G, H, i, t, u, y, z

flibbles:
I've just noticed that @linonetwo has found a way to do this without editing the core sjcl.js file. Despite sjcl being an unmaintained project, it's still a better practice not to manipulate a minified file, so I recommend merging [his PR](https://github.com/Jermolene/TiddlyWiki5/pull/8099) rather than mine.

Jermolene:
Thanks @flibbles, it's my fault for not getting to #8099 quickly enough

PR #8198, opened 2024-05-20
Add plugin stability badges

Jermolene:
This PR adds a new standard field to plugins called `stability`. It can contain the following values:

* **STABILITY_0_DEPRECATED** - Deprecated. This plugin is not recommended for new projects
* **STABILITY_1_EXPERIMENTAL** - Experimental. Non-backward compatible changes or removal may occur in any future release. Use of the plugin is not recommended in production environments
* **STABILITY_2_STABLE**  - Stable.
* **STABILITY_3_LEGACY** - Legacy. Although this plugin is unlikely to be removed, it is no longer actively maintained, and other alternatives are available

These stability levels are taken from the Node.js project - https://nodejs.org/api/documentation.html#stability-index.

There are several goals for introducing this feature:

* To warn users away from deprecated and legacy plugins that are still in the library for backwards compatibility reasons
* To allow experimental plugins to be merged before they are fully finalised. For example, the [Geospatial plugin](https://github.com/Jermolene/TiddlyWiki5/pull/7406) is complete enough for users to start experimenting with it, but it is not yet appropriate to subject it to the same backwards compatibility constraints as the core itself. Releasing the plugin makes it more accessible for users to try it out, and thus will give us more feedback for improvements

Questions:

* The values of the "stability" field could be straightforward single words like "DEPRECATED", "EXPERIMENTAL" etc but I found it easier to navigate the code by using the full, more unique name. I'd be interested in other views
* Should we hide the "stable" badge in the control panel listings? It is arguably redundant, and perhaps makes it harder to spot the ones that are deprecated or experimental.

The stability level is shown in Control Panel as a small badge:

image

It is also visible in the "add new plugins" modal:

image


pmario:
>- Should we hide the "stable" badge in the control panel listings? It is arguably redundant, and perhaps makes it harder to spot the ones that are deprecated or experimental.

It think "stable" should also be displayed, otherwise users will question themself why there is no badge. Also "experimental" and "deprecated" have different colours and are bigger, so I think they should be visible enough.

Jermolene:
Thank you @flibbles good questions.

A third party plugin like Relink might have used "EXPERIMENTAL" for the first release until they are ready to lock down their backwards compatibility guarantees.

The point at which a third party plugin might be marked as "DEPRECATED" might be if you decided to replace a plugin with a new version that is not fully backwards compatible, and wanted users to gradually migrate from the old plugin to the new.

> Could it make more sense to attach this flag to the data that's part of the plugin library rather than the plugin itself?

That may well be needed as well. With this PR we and embedded data we have a way for the developer to express their intent, which may well be overridden by a hosting plugin library. For example, there could be a third party plugin that is no longer maintained but still has the "STABLE" label, even though it is obsolete and there are better ways to achieve the same functionality. A plugin library may well have additional metadata to mark the plugin as deprecated and lay out the upgrade path.

> Or would a work better if instead of specifying Relink as "STABILITY_2_STABLE", I specify the range it's tested for. Like 5.1.17-5.3.3. And if your version is outside of that range, it gets the deprecated flag?

I think that that metadata would be orthogonal and complementary to the stability index.

> Or is this the case where I only push a "Deprecated" version of Relink as its last version before I take up chicken farming? And all past versions of Relink would still have their stable flags?

Jermolene:
> Or is this the case where I only push a "Deprecated" version of Relink as its last version before I take up chicken farming? And all past versions of Relink would still have their stable flags?

Yes, I think that would be correct. In that case the plugin developer has an intent, and the stability index gives them a clear way to express it.

core/wiki/config/OfficialPluginLibrary.tid
```diff
@@ -1,6 +1,6 @@
-title: $:/config/OfficialPluginLibrary
-tags: $:/tags/PluginLibrary
-url: https://tiddlywiki.com/library/v5.3.3/index.html
 caption: {{$:/language/OfficialPluginLibrary}}
+tags: $:/tags/PluginLibrary
+title: $:/config/OfficialPluginLibrary
+url: https://tiddlywiki5-git-plugin-stability-badges-jermolenes-projects.vercel.app/library/v5.3.3/index.html
 
-{{$:/language/OfficialPluginLibrary/Hint}}
+Plugin library for https://tiddlywiki5-git-plugin-stability-badges-jermolenes-projects.vercel.app
```
pmario, line 6: @Jermolene -- I think there is a problem with this URL link.

editions/prerelease/tiddlers/system/PrereleaseOfficialPluginLibrary.tid
```diff
@@ -1,6 +1,6 @@
-title: $:/config/OfficialPluginLibrary
+caption: {{$:/language/OfficialPluginLibrary}}
 tags: $:/tags/PluginLibrary
-url: https://tiddlywiki.com/prerelease/library/v5.3.3/index.html
-caption: {{$:/language/OfficialPluginLibrary}} (Prerelease)
+title: $:/config/OfficialPluginLibrary
+url: https://tiddlywiki5-git-plugin-stability-badges-jermolenes-projects.vercel.app/library/v5.3.3/index.html
```
pmario, line 4: @Jermolene -- This one seems to be wrong too

core/wiki/config/OfficialPluginLibrary.tid
Jermolene: Oops thanks @pmario that was for the preview, I'll remove it

pmario:
@Jermolene ... For my plugins I need `STABILITY_4_BETA`

Jermolene:
> @Jermolene ... For my plugins I need `STABILITY_4_BETA`

Why not use STABILITY_1_EXPERIMENTAL? These levels are taken from the Node.js project. I am not keen to have any more choices than strictly needed; more levels translates into a greater cognitive load on users.

PR #8199, opened 2024-05-20
feat: allow render tiddler above root

linonetwo:
图片

To show command palette on every possible layout, I need a place that is transcend of the layout mechanism to mount it, where it is the `$:/core/ui/RootTemplate`.

### DIFF
--- a/core/ui/RootTemplate.tid
+++ b/core/ui/RootTemplate.tid
@@ -1,5 +1,12 @@
 title: $:/core/ui/RootTemplate
 code-body: yes
 
+<$list filter="[all[shadows+tiddlers]tag[$:/tags/AboveRoot]!has[draft.of]]">
+<$transclude/>
+</$list>
+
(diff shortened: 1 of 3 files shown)


linonetwo:
This make

```ts
this.dispatchEvent({
  type: 'tm-navigate',
  navigateTo: itemUrl,
  navigateFromNode: this,
});
```

not working, because it is out of `<$navigator` widget's scope.

linonetwo:
Maybe something like this will be better?

```tid
title: $:/core/ui/CommandPaletteTemplate
tags: $:/tags/PageTemplate

<$list filter="[all[shadows+tiddlers]tag[$:/tags/CommandPalette]!has[draft.of]]">
<$transclude/>
</$list>
```

And I will ask layout authors to bring `{{$:/core/ui/CommandPaletteTemplate}}` with them.

PR #8200, opened 2024-05-20
Update chinese language files

BramChen:
* Add chinese translations for the new `<$testcase>` widget


Jermolene:
Thanks @BramChen

PR #8201, opened 2024-05-20
enable loading new tiddlers from filesystem after plugin init

CyberFoxar:
This PR adds a new way to load tiddlers when running Tiddlywiki with NodeJS. See https://github.com/Jermolene/TiddlyWiki5/discussions/8156 for the full discussion and motivation behind this.

This PR in particular adds 2 things:
- the deferred loading mechanism
- an example plugin implementing it, and associated modifications in one of the editions to demo it

The goal is to reuse as much of the tiddlywiki existing code and mechanisms as possible whilst enabling a way to save and load tiddlers once plugins have been discovered and instanced.

This PR is marked as a draft as there are a number of questions still in the air, as well as a number of TODO items I need to take care of. It is in a working state however, and some feedback would help me finish it properly.

Questions:
1. How best should I save the various `tiddlerloaders` we have ? Using `$tw.deserializerParsers` feels unelegant. The aim was to replicate somewhat how the parsers are saved inside `$tw.wiki`.
2. How should we handle `deferredFiletype`, i.e. know which `tiddlerloader` to call upon ? Using custom logic to handle the file extensions seems weird and using a field to keep track of that feels unelegant too. Maybe add it to the `fileinfo` object that's saved in `$tw.boot.files` ?

Concerns:
3. Should we expose the filename to loader plugins and allow them to edit it willy-nilly ? This feels a bit invasive, but then again, you're installing the plugin in a NodeJS instance which has already access to your whole filesystem and internet, should they wish to. Allowing for edition of the filename would help with compatibility with some other note-taking systems (i.e. dendron) which store some of the tags inside the filenames.

Closing notes:
This is my first PR in this project, and to be frank, any OSS project. Please do tell me what I'm doing wrong.


editions/server/tiddlywiki.info
```diff
@@ -3,7 +3,9 @@
 	"plugins": [
 		"tiddlywiki/tiddlyweb",
 		"tiddlywiki/filesystem",
-		"tiddlywiki/highlight"
+		"tiddlywiki/highlight",
+		"tiddlywiki/markdown",
+		"cyberfoxar/fm-md"
```
pmario, line 8: I think it would be better, to create a new temporary edition eg: "server-deferred". You and use the server edition as a starting point. It's less error-prone and easier to clean up after testing. 

-----

eg: If someone runs `tiddlywiki --init server` with your branch, they will inherit these settings.

plugins/cyberfoxar/fm-md/fmloader.js
```diff
@@ -0,0 +1,101 @@
+/*\
+title: $:/plugins/cyberfoxar/fm-md/fmloader.js
+type: application/javascript
+module-type: tiddlerLoader
+
+Proof-of-concept plugin for a tiddlerloader.
+\*/
+
+(function(){
```
pmario, line 9: Just an info:

We intend to get rid of this function wrapper in the future. Because pretty-printers want to indent the whole code block and the function is not needed anyway. 

So usually we do not indent the code like this. We tend to ignore the outer function from a indentation point of view.

boot/boot.js
```diff
@@ -1963,14 +1963,74 @@ $tw.loadTiddlersFromPath = function(filepath,excludeRegExp) {
...
 };
 
+$tw.deferredDirSpecs = [];
+
 /*
 Load all the tiddlers defined by a `tiddlywiki.files` specification file
 filepath: pathname of the directory containing the specification file
 */
-$tw.loadTiddlersFromSpecification = function(filepath,excludeRegExp) {
+$tw.loadTiddlersFromSpecification = function(filepath,excludeRegExp,loadDeferred) {
```
pmario, line 1972: I would suggest that we use `function(filepath,excludeRegExp,options) {` as a new parameter. So we do not have to change the function-signature again. It's easy to add new options in the future. eg:

```
$tw.loadTiddlersFromSpecification = function(filepath,excludeRegExp,loadDeferred) {
	options = options || {};
```

So you can use `options.loadDeferred` -- Look at line 2259 ff ($tw.loadWikiTiddlers) to see how we document it.

pmario (commented):
I think overall it looks pretty good for a start. I did not run the code in detail. Just had a look at the structure.

I'll do more testing soon

CyberFoxar:
woops. Shouldn't have pushed without test x)
Sorry about that.

plugins/cyberfoxar/fm-md/fmloader.js
CyberFoxar: Noted.
As an aside, you're planning on moving to CommonJS Modules, then ? Or even ES2015 modules ? :o

pmario: TW uses CommonJS Modules atm. IMO ES-modules will not be so easy, without a complete rewrite. 

The extra `(function(){` wrapper was added to be absolutely sure that we do not pollute the global variable space, even if someone else uses TW modules. 

BUT

The core module system already wraps every TW module in such a function. So they are redundant for TW and they cause indentation problems with automatic pretty printing, which we do want to introduce in the future. So there should be no discussion about code-styling anymore. The pretty-printer can take care of this.

pmario:
>Ah ! That's "just" the new plugin field.
I should really look into how we run tests properly >w<

hihi. That just caught me by surprise too ;) With an extra twist. See: #8207

CyberFoxar:
Done ! I've updated to latest x)

I've taken all your comments in account. Still working on finding out how to make the defer system more flexible regarding file extensions.

plugins/cyberfoxar/fm-md/fmloader.js
```diff
@@ -0,0 +1,101 @@
+/*\
+title: $:/plugins/cyberfoxar/fm-md/fmloader.js
```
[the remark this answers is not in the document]
CyberFoxar: Changed to frontmatter-tiddler, which felt a bit more descriptive (as it does not even invoke markdown x) )

Jermolene:
Hi @CyberFoxar and thank you for your dedication on this. We are keen on PRs like this that make minor tweaks to the core in order to accommodate the needs of plugins.

One of the criteria is generally that there is no other way for the plugin to proceed. Belatedly, in this case I've realised that there might be an approach that wouldn't require any changes to the core. It takes advantage of the fact that if TiddlyWiki finds a `tiddlywiki.files` app in a directory then it won't apply normal processing to that folder, it will only perform any processing specified within the file. That means that a folder containing a `tiddlywiki.info` file with the dummy content `{}` will be totally ignored by TiddlyWiki when it is loading tiddlers.

The idea is that the plugin would require that any folders containing the specially processed Markdown files also contain a file called `tiddlywiki.files` containing `{"my-new-markdown-format": true}`. The plugin would contain a startup module that would scan the wiki folder a second time, looking for `tiddlywiki.files` files that contain the magic `my-new-markdown-format` marker, and process the files in that folder.

Does that make sense? Would it meet your needs?

CyberFoxar:
@Jermolene, I like the idea ! I'll do some testing using that to see how it would mesh with what I have in mind. Below are my initial thoughts on this.

> One of the criteria is generally that there is no other way for the plugin to proceed. Belatedly, in this case I've realised that there might be an approach that wouldn't require any changes to the core. It takes advantage of the fact that if TiddlyWiki finds a `tiddlywiki.files` app in a directory then it won't apply normal processing to that folder, it will only perform any processing specified within the file. That means that a folder containing a `tiddlywiki.info` file with the dummy content `{}` will be totally ignored by TiddlyWiki when it is loading tiddlers.

I kinda see how it would work. I've got some testing to do, but this seems more to be a side-effect of how wiki discovery is handled than an intended behavior. I'd be keen on "officializing" this behavior if we go this route as I think this would output a warning/error message in log for the dummy `tiddlywiki.info` file.

> The idea is that the plugin would require that any folders containing the specially processed Markdown files also contain a file called `tiddlywiki.files` containing `{"my-new-markdown-format": true}`. The plugin would contain a startup module that would scan the wiki folder a second time, looking for `tiddlywiki.files` files that contain the magic `my-new-markdown-format` marker, and process the files in that folder.

As for that, it's already kinda how I expect to implement it. A marker in the `tiddlywiki.files` to indicate what kind of special processing to do. 

---

Thought, I might have misread, so I'll try my best to summarize what I understood:

Idea is to kinda take advantage of the behavior of `$tw.loadWikiTiddlers` that bails when it finds a dummy `tiddlywiki.info` file. We'd either add a hook there that a module could take advantage of, or otherwise remember that we should re-scan it later.
Then, boot goes as normal, since the wiki wasn't loaded because it's a dummy wiki, the tiddlers in it arn't processed. 
Then, a startup module re-scans the directory and actually processes it correctly according to plugin rules.

I like the idea, though it is not terribly different from the current implementation, except that instead of changing `tw.loadTiddlersFromSpecification`, we'd change  `$tw.loadWikiTiddlers` and add some custom handling to `load-defer.js`.

My issue with this approach is that `load-defer` might get a lot of responsibility, because "re-scanning" might involve parsing some kind of spec file whilst still needing to invoke the `tiddlerloader` plugins. This would involve either keeping the current patched behavior of `tw.loadTiddlersFromSpecification` or duplicate some code to handle specification discovery.

My goal in patching into the spec discovery is to keep reusing as much of the tiddlywiki mechanism as possible. I like the idea of the custom fields, the regex-based approach to file loading, the path filters, etc. It seems that using the wiki-bypass approach would bypass all of those goodies or require duplicating code, or require cutting up big functions into a lot of smaller ones to reuse them (which I'm not opposed to, but it seems a bit out of the scope of this PR).

---

What we might look into, if we want to keep `boot.js` changed to a minimum, is maybe add a new hook to be invoked by the defer module, but I'm not sure what form this would take.

---

Anyhow, I've a lot of fiddling still to do. I'm having some fun hacking away at a proper dev project :P

Jermolene:
Hi @CyberFoxar the approach I was thinking of wouldn't require any changes to the core. The disadvantage would be that your plugin would need to scan the wiki folder looking for its markers independently of the previous scan that the core has already done. And if there were two such plugins then they would each be performing their own scan.

>  think this would output a warning/error message in log for the dummy tiddlywiki.info file.

There's currently no error message when a dummy tiddlywiki.info is encountered.

> As for that, it's already kinda how I expect to implement it. A marker in the tiddlywiki.files to indicate what kind of special processing to do.

Yes the user experience should be very much the same.

Jermolene:
>  Then, boot goes as normal, since the wiki wasn't loaded because it's a dummy wiki, the tiddlers in it arn't processed.

Just to be clear, the `tiddlywiki.files` trick applies to a subfolder, not the whole wiki. So a wiki folder could contain a subfolder containing markdown files, and another subfolder containing ordinary tiddlers.

> We'd either add a hook there that a module could take advantage of, or otherwise remember that we should re-scan it later. 

As I say, I am suggesting that the plugin do the scan for itself, even though there is a cost in efficiency.

CyberFoxar:
Hah, yeah, that would work.
I think I see what you mean. I'll test out some implementations.

I think I'll try building out the module to handle the re-scanning of directories, as that seems unlikely to be plugin-dependant (I'll probably add an override switch in any case).

There will still be some changes made to boot, just to expose some of the inner functions currently used in a few places to allow the module to reuse as much code as possible.

CyberFoxar:
@Jermolene, Re: loading via wiki
I'm not sure what I should change in my file structure to make the dummy `tiddlywiki.info` file work properly. Some assistance in getting an expected test wiki sertup would be great ^^

I'm also toying with the idea of (re)using filetypes and registering tiddlerdeserializers, but I'm a bit hazy on how I'll handle the re-serialization afterward. I'm tempted to extend the `$tw.config.contentTypeInfo` to add optional information about serialization too. Something like `contentTypeInfo[type].serializerType: "..."`

boot/boot.js
```diff
@@ -576,10 +579,27 @@ $tw.utils.registerFileType = function(type,encoding,extension,options) {
...
+ * @type {object}
+ * @property {string} type
+ */
+
+/**
+ * Given an extension, always access the $tw.config.fileExtensionInfo
+ * using a lowercase extension only.
+ * @param {string} ext - Extension to find a type for
+ * @returns {FileExtensionInfo | null | undefined}
+ */
```
CyberFoxar, line 602: Also, commenting on my own code there, but how do you like this style of comments ?

They're [JSDoc](https://jsdoc.app/) type information. It's quite nifty, though not all its features are available in all editors (as they tend to implement TypeScript type hints, which are only a subset of what JSDoc can express).

I don't plan on adding them on _everything_, but I'm adding them as I go and find them useful.

PR #8203, opened 2024-05-22
Fixed issue where $parameters did nothing with no $transclude

flibbles:
Discovered this problem when creating a new TiddlyMap release.

If you try to directly render a tiddler (such as by using `$tw.wiki.renderTiddler`) that has `\parameters` or `<$parameters>` in it, the parameters widget will do nothing. It'll look up its parent tree, find no transclude to possibly pull arguments from, and just disable itself.

I'm not sure this should be expected behavior. For instance, TiddlyMap now has to do a weird workaround for this just so it can render the `$:/core/images/tag-button` into a usable form for svg icons. I would think expected behavior for $parameters would be to use its default parameters in such a case. It certainly makes sense for icons.

### DIFF
--- a/core/modules/widgets/parameters.js
+++ b/core/modules/widgets/parameters.js
@@ -74,6 +74,18 @@ ParametersWidget.prototype.execute = function() {
 				self.setVariable(variableName,getValue(name));
 			}
 		});
+	} else {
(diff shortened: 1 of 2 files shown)


Jermolene:
Thank you @flibbles I agree, this is a very worthwhile fix.

PR #8206, opened 2024-05-23
add jsonvariables operator plus documentation

pmario:
@Jermolene -- **Important**

This PR implements a new json operator named: `jsonvariable`, which allows us to show the internal structure from 1 variable. Using jsonextract, it's possible to access the different elements. 

As suggested by you: https://github.com/Jermolene/TiddlyWiki5/pull/7992#issuecomment-2089838323

- **This PR** and 
- https://github.com/Jermolene/TiddlyWiki5/pull/8179

are needed to let me implement  **Dumpvariables rewrite new syntax plus search** #7992 as a plugin.


Jermolene:
Thanks @pmario I think it's problematic to use `jsonvariable` as the name of the operator because it implies it is one of the JSON operators. I prefer the name `getvariablejson` because it is consistent with the existing `getvariable` operator.

pmario:
OK. I will change the name.

pmario:
@Jermolene -- renamed to `getvariablejson`

pmario:
I think the text element is not internal. I think we should keep the `text` element, even if it is a duplicate of `getvariable`.

pmario:
@Jermolene

- I did change the code according to your suggestions and did add some tests. 
- I think the `text` variable is needed for completeness
- I did add an `isVariable` element for variables that are created with eg: let-widget. This can be used for easy logic.
- I did set the `is*` values to "yes" and "no" strings instead of `true and false`
- Since some definitions have `name:value` and others have `name:default` I did unify them to `name:default` for all params. This should avoid confusion for users

core/modules/filters/getvariablejson.js
```diff
@@ -0,0 +1,52 @@
...
+				return {"name":param.name, "default": param.value || param.default || ""}
+			});
+		};
+		if (info.srcVariable) {
+			out.value = info.srcVariable.value;
+			out.isFunctionDefinition = (info.srcVariable.isFunctionDefinition) ? "yes" : "no";
+			out.isProcedureDefinition = (info.srcVariable.isProcedureDefinition) ? "yes" : "no";
+			out.isWidgetDefinition = (info.srcVariable.isWidgetDefinition) ? "yes" : "no";
+			out.isMacroDefinition = (info.srcVariable.isMacroDefinition) ? "yes" : "no";
+			out.isVariable = (info.srcVariable.isFunctionDefinition ||
```
Jermolene, line 42: The name "isVariable" does not make sense; functions, widgets, macros and procedures are all variables, and so it would be confusing for "isVariable" to equal "no".

Perhaps "isPlain" or "isSimple"?

Or maybe we don't need this at all? What is the use case?

If the intent is to be able to distinguish plain variables set with the `<$set>` widget from definitions, then things will get complicated when/if we extend the set widget to be able to define procedures/widgets/functions.

On balance, I think we should remove this.

core/modules/filters/getvariablejson.js
```diff
@@ -0,0 +1,52 @@
...
+"use strict";
+
+/*
+Export our filter function
+*/
+exports.getvariablejson = function(source,operator,options) {
+	var results = [],
+		widget = options.widget;
+	// "replacer" must be defined, otherwise JSON.stringify will throw a circular reference error RSOD
+	var replacer= "params name value default text isFunctionDefinition isProcedureDefinition isWidgetDefinition isMacroDefinition isVariable".split(" ");
```
Jermolene, line 23: I really don't think we should be exposing the `text` property. It is the same as the value retrieved from the variable in the usual way. Worse, it exposes an implementation detail of the caching mechanism.

core/modules/filters/getvariablejson.js
```diff
@@ -0,0 +1,52 @@
...
+	source(function(tiddler,title) {
+		var text,
+			out = {},
+			info = widget.getVariableInfo(title, {});
+
+		out.text = info.text;
+		if (info.params) {
+			// unify parameter elements
+			out.params = info.params.map(function(param) {
```
Jermolene, line 32: Please can you explain what this is doing and why it is necessary?

Jermolene (commented):
Thanks @pmario

pmario:
>We might in the future implement a way to set variables via this same JSON structure, along special variables like functions and procedures to be programmatically created.

I think it's too early to implement the `getvariablejson` operator. Once the "set" widget is able to define "procedures" there may be a possibility to revoke this PR again. -- So I'll close it. 

-----

I'll implement what I do need for my plugin in a much simpler way.

Jermolene:
Hi @pmario that makes sense for the moment, thanks for exploring this. I do think we should come back to it.

core/modules/filters/getvariablejson.js
pmario: The map function is needed, because there is an inconsistency in the core code. One of the variable definitions uses `param.value` and the others use `param.default` -- Or vice versa. I cannot remember without debugging info. I think we cannot change that behaviour without compatibility problems

PR #8214, opened 2024-05-27
fix: boot.css missing

linonetwo:
#8099 's tiddlywiki.files ignore the css, and was not noticed by me because I don't know I should test it with

```tid

```

Thanks for @pmario to point it out.

Seems this css file don't need a style system tag, so I correct the config to find it.

I also tried let it still have `.tid` suffix, but failed to load it. And with `.tid` suffix, a drawback is that VSCode can't apply css highlighting to it.

截屏2024-05-27 18 26 23

### DIFF
--- a/boot/boot.css
+++ b/boot/boot.css
@@ -1,6 +1,3 @@
-title: $:/boot/boot.css
-type: text/css
-
 /*
 Basic styles used before we boot up the parsing engine
 */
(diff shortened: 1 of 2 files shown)


Jermolene:
Thanks @linonetwo

Jermolene:
Hi @linonetwo I've pushed a small enhancement to avoid renaming the boot.css file in order to improve the continuity of our GitHub history - see da8d4ecfae60a807d6c8d884b61699e15378f465

linonetwo:
I see, I find I was adding `"isTiddlerFile": true` into `fields` by mistake.
图片

Issue #8216, opened 2024-05-27
[BUG] Unavoidable RSOD Caused by RegExp


pmario:
This issue is a single stepped, deep dive and it's related to: https://github.com/Jermolene/TiddlyWiki5/issues/8213

As shown in #8213 there is a RSOD created by $:/core/modules/utils/linkedlist.js because it throws an error if it encounters "array elements" other than strings. -- IMO this behaviour is valid.

The problem now is that this RSOD is unavoidable with the current js implementation, even if every code element, wikitext and JS-code in itself is valid. 

#### Wikitext snippet from #8213 seems to be valid

```
\define re() (color)|(colour)ed
\define str() Something coloured
{{{ [<str>splitregexp<re>] }}}
```

- See TW docs: https://tiddlywiki.com/#splitregexp%20Operator - There are no documented limitations about the regexp.

#### JS code seems to be valid

- https://github.com/Jermolene/TiddlyWiki5/blob/dbe912ba5d6b51e542abd8a94db0d49cc7eee04e/core/modules/filters/strings.js#L220

- See MDN `string.split(regexp)`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/split#description -> Scroll down to 

>If separator is a regular expression with capturing groups, then each time separator matches, the captured groups (including any undefined results) are spliced into the output array. This behavior is specified by the regexp's [Symbol.split](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/split) method.

In the quote it says: *the captured groups (**including** any **undefined** results) are spliced into the output array.*

#### How the problem happens

In the code the regexp `/(code)|(colour)ed/` contains 2 capturing groups, which will return at least 1 `undefined` with `Something coloured` given as "source text".

- In `strings.js` the `result` variable is an array and contains the `undefined` value.

```
result: (4) […]
​​0: "Something "
​​1: undefined
​​2: "colour"
​​3: ""
​​length: 4
```

- Then the `result` variable will be passed to: `LinkedList.js` to the `LinkedList.prototype.pushTop(value)`, where it will be `_assert`ed
- https://github.com/Jermolene/TiddlyWiki5/blob/dbe912ba5d6b51e542abd8a94db0d49cc7eee04e/core/modules/utils/linked-list.js#L58-L63 
- **So line 62 throws an error which is unavoidable** at the moment

### Conclusion

[...]

Issue #8213, opened 2024-05-25
[BUG] Fix splitregexp operator JavaScript error on matches with undefined capture groups


bluepenguindeveloper:
### Describe the bug

When the regular expression parameter to the splitregexp argument contains a capture group that (due to a disjunction `|`) is undefined in a match, it throws an uncaught error `Linked List only accepts string values, not undefined`

### Expected behavior

The input should be split on the match (edit: and `undefined` capture groups should be safely handled, presumably by simply removing them from the output since unlike empty strings they wouldn't represent tiddler titles) with no "Red Screen of Embarrassment"

### To Reproduce

Copy and paste the following code into a new tiddler (with the preview pane open):
```
\define re() (color)|(colour)ed

\define str() Something coloured

{{{ [<str>splitregexp<re>] }}}
```

### Screenshots

![image](https://github.com/Jermolene/TiddlyWiki5/assets/166951090/40f1840c-5b7e-4780-8c0e-9fb568b73e74)

### TiddlyWiki Configuration

- Version v5.3.3
- Saving mechanism TiddlyDesktop, browser
- No plugins installed (except core and the en language)

### Desktop:

- OS: Windows 10
- Browser: Brave (Chromium)

pmario:
There is an unrelated question. 
What would you expect from this filter? How should the output look like?

joshuafontany:
The difference is in your capture groups: https://regexr.com/80r94
`(color)|(colour)ed` returns two capture groups, one of white will be null. Either $1 = `color` | $2 = `colour`.

pmario:
I do know what's going on. I just wanted to know what the expectations where. There seems to be a bug in the implementation. And the assumption what the regexp does is wrong too. But what we know is, that there should be no RSOD no matter what.

pmario:
RegExps are complex. That's why I was asking, what you expected as a result. With `splitregexp` capture groups are not really useful as it is implemented in the JS code. 

But as I wrote. There should be no RSOD anyway

Jermolene:
Thanks @bluepenguindeveloper

PR #8222, opened 2024-05-29
Make sure split(regex) returns an array of strings

pmario:
This PR makes sure split(regex) used by the `splitregexp` operator returns an array of strings. It fixes: #8216 and #8213

Code to test with

```
{{{ [<str>splitregexp<re>join[, ]] }}}

---

```

image

### DIFF
--- a/core/modules/filters/strings.js
+++ b/core/modules/filters/strings.js
@@ -127,7 +127,7 @@ function diffPartsToChars(text1,text2,mode) {
             if(lineHash.hasOwnProperty ? lineHash.hasOwnProperty(line) : (lineHash[line] !== undefined)) {
 				chars += String.fromCharCode(lineHash[line]);
             } else {
-                if (lineArrayLength == maxLines) {
+                if(lineArrayLength == maxLines) {
                   line = text.substring(lineStart);
                   lineEnd = text.length;
                 }
@@ -218,7 +218,7 @@ exports.splitregexp = function(source,operator,options) {
 	}
 	source(function(tiddler,title) {
 		var parts = title.split(regExp).map(function(part){
-			return part+"";	// make sure it's a string
+			return part || "";	// make sure it's a string
 		});
 		Array.prototype.push.apply(result,parts);
 	});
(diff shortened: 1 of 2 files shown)


core/modules/filters/strings.js
```diff
@@ -217,7 +217,10 @@ exports.splitregexp = function(source,operator,options) {
 		return ["RegExp error: " + ex];
 	}
 	source(function(tiddler,title) {
-		Array.prototype.push.apply(result,title.split(regExp));
+		var parts = title.split(regExp).map(function(part){
+			return part+"";	// make sure it's a string
```
Jermolene, line 221: Hi @pmario I think this should be `return part || ""`.

pmario: But then `undefined` would return an emtpyString which is inconsistent. Please have a closer look at the OP code examples. IMO `{{{ [<str>splitregexp<re>join[, ]] }}} and {{{ [<str>splitregexp<re>] }}}` should return the same number of elements. This would not be the case, if we use `part || ""`

saqimtiaz: Returning "undefined" as a filter result is user hostile for non-developers and has no precedence among filter operators. On the other hand there is precedence for operators returning an empty string when there is no matching result.

pmario: I'll change it as requested

pmario:
@Jermolene -- Please confirm that the following code should create the output as shown in the screenshtot.

```
{{{ [<str>splitregexp<re>join[, ]] }}}

---

```

image

Jermolene:
Hi @pmario the issue here is that the JS split operator is essentially useless when used with a regex that includes capture groups.

So, I think the best we can do is coerce undefined to an empty string, and add a warning to the docs for the split operator explaining why it should not be used with capture groups.

pmario:
>So, I think the best we can do is coerce undefined to an empty string, and add a warning to the docs for the split operator explaining why it should not be used with capture groups.

OK - I'll some info to the docs too.

pmario:
@Jermolene -- This should work as requested now + a bit more docs

Jermolene:
Thank you @pmario

PR #8224, opened 2024-05-30
Update chinese language files

BramChen:
* Improve traditional chinese translation of "Standard Layout"


Jermolene:
Thanks @BramChen

PR #8225, opened 2024-05-30
Add useDefaultHeaders flag to tm-http-request (V2)

mklauber:
Reworked the handling of useDefaultHeaders did better testing than last time.  

* Switched to a ternary operator for converting to boolean, as the previous logic didn't correctly parse possible value strings.

Hopefully I got it right this time.  I did try running `nodejs tiddlywiki.js editions/tw5.com-server --listen` and validating that both normal browsing and my new changes appear to work.


Jermolene:
Thanks @mklauber

PR #8229, opened 2024-06-01
typo in Procedures.tid

twMat:


twMat:
Actually, that whole statement needs rephrasing. It claims that the "doubled angled brackets" transcludes "the snippet" but, actually, the double angle brackets that the reader has seen at this point in the text is really just the parameter call. And "the snippet", as per the tiddlers initial sentence, means the procedure content itself.

Jermolene:
Hi @twMat I think this PR does fix the typo effectively, and so I am inclined to merge it.

PR #8231, opened 2024-06-01
Catchup with fr-FR translation

xcazin:


Jermolene:
Great, thank you @xcazin

PR #8232, opened 2024-06-02
Use the testcase widget for $action-createtiddler examples

btheado:
Translate the `$action-createtiddler` examples to use the `$testcase` widget.

Directly used the `$testcase` widget rather than test case tiddlers so the user can interactively click the buttons to see the behavior.

Converted macro `\define` to `\procedure` in the examples.

### DIFF
--- a/editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget Example 1.tid
+++ b/editions/tw5.com/tiddlers/widgets/ActionCreateTiddlerWidget Example 1.tid
@@ -1,26 +1,15 @@
 created: 20200131142401129
-modified: 20211113230406823
+modified: 20240602154417754
 tags: ActionCreateTiddlerWidget
 title: ActionCreateTiddlerWidget Example 1
 type: text/vnd.tiddlywiki
 
-\define testCreate()
-<$action-createtiddler />
-\end
-
-This example will create "New Tiddler", "New Tiddler 1", " New Tiddler 2" and so on
-
-```
-\define testCreate()
-<$action-createtiddler />
-\end
+<$testcase>
+<$data title=Description text="Default widget with no attributes"/>
+<$data title=Narrative text='The button in this example uses the $action-createtiddler widget to create "New Tiddler", "New Tiddler 1", " New Tiddler 2" and so on'/>
+<$data title=Output text="""\define testCreate() <$action-createtiddler />
 
 <$button actions=<<testCreate>> >
 Create Tiddler
-</$button>
-```
-
-<$button actions=<<testCreate>> >
-<$action-setfield $tiddler="$:/state/tab/sidebar--595412856" text="$:/core/ui/SideBar/Recent"/>
-Create Tiddler
-</$button> - Clicking this button, will also open the Right sidebar: Recent tab
+</$button>"""/>
+</$testcase>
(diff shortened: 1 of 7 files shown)


pmario:
Creating tiddlers should actually create tiddlers. It's important to be able to test different buttons while there are tiddlers with the same name already created. Especially the last example should also open newly created tiddlers.. 

I'm not happy with the changes, because it does completely change the behaviour of the examples.

btheado:
@pmario, thanks for the feedback.

> Creating tiddlers should actually create tiddlers

I don't understand what you mean. That's what the action-createtiddlers widget does and all the examples are using that widget.

> It's important to be able to test different buttons while there are tiddlers with the same name already created.

What buttons do you mean?

>  It should not create tiddlers, with the same name

What should not create tiddlers with the same name? Only one of the [examples](https://tiddlywiki5-git-fork-btheado-docs-us-128711-jermolenes-projects.vercel.app/#ActionCreateTiddlerWidget%20Example%202) in which the button creates tiddlers of the same name since it uses the overwrite attribute.

>  Especially the last example should also open newly created tiddlers..

It already opens the newly created tiddler. Do you mean you want the Output tiddler to have a navigator widget so a story river can be seen?

pmario:
The first example with the "Create new page control button" should create a "real" page control button, that is shown in the right sidebar. It should open the button in the story river. The testcase does not do any of that. So users cannot see, what the example code actually does. 

image

Example 2 creates a "New Tiddler" and "New Tiddler 1" if clicked several times. 
Example 3 should overwrite the existing "New Tiddler" created in example 2 -- The "right sidebar" shows this. So users can see, what's going on. 

The 4th and 5th example have the same base-name. So ex. 4 should create eg: "base", "base 1" - and so on. If button in ex. 5 is clicked it should create "base 2" and "base 3" - It should not start with "base" again, as it does with the testcase examples.  

The last example should create a new tiddler and it should open that tiddler in the story river. That the whole purpose of that example. Testcase does not open the tiddler in the story river. 

-------

IMO testcase can be used, if examples should be shown in an "isolated" environment. But the action-createdtiddler examples depend on each other. They are designed to "modify" the existing wiki state and the story river.

Jermolene:
Thanks @btheado these are looking good.

> The first example with the "Create new page control button" should create a "real" page control button, that is shown in the right sidebar. It should open the button in the story river. The testcase does not do any of that. So users cannot see, what the example code actually does.

I disagree. One of the confusing things about the current demo is the way that there is action at a distance: the user interacts with the example but has to look at an unrelated part of the screen to see the effect. That has always been deeply confusing, and really doesn't belong in a documentation example. It sounds like this should be one of the "How to..." guides, and not part of the documentation for the underlying primitive.

btheado:
@pmario thanks for the explanation. I didn't notice that examples 2 and 3 and examples 4 and 5 were meant to work together. If that is the case, I think combining the interdependent examples and using two buttons would make that more clear.

pmario:
>If that is the case, I think combining the interdependent examples and using two buttons would make that more clear.

Probably -- Yes

btheado:
I got a little stuck trying to satisfy the original intention of the examples and not "polluting" the example code with extraneous code (and making them harder to read). I created PR #8242 and #8243 as simpler first examples of converting to the testcase widget. Will come back to this one later.

Jermolene:
Thanks @btheado and apologies for the delay.

PR #8233, opened 2024-06-04
Fixes nested functions to resolve filter run variables

### LINKED ISSUE #8178: [BUG] filter prefixes don't properly set currentTiddler, I think
### Describe the bug

Filter prefixes are turning out to be something I use often, not so much because they're elegant, but because they're able to set variables in the middle of a filter expressions, which can be crucial for some operations.

But I'm noticing strange behavior. While a filter run may set currentTiddler within its own run, it doesn't set it within functions called within that run, or anything within that run really.

Is that expected behavior that currentTiddler is only set superficially?

### Expected behavior

```
old-current

new-current
```

When I would have thought it'd produce:

```
new-current

new-current
```

### To Reproduce

```
\define currentTiddler() old-current

\function .inner() [<currentTiddler>]
\function .outer() new-current :map[function[.inner]]

\function .alone() new-current :map[<currentTiddler>]


<<.outer>>
<<.alone>>
```

saqimtiaz:
Fixes #8178 

The `getVariable` method in `makeFakeWidgetWithVariables` needs to accommodate being called with variables in the opts argument, which happens with nested fake widget objects.

### DIFF
--- a/core/modules/widgets/widget.js
+++ b/core/modules/widgets/widget.js
@@ -317,7 +317,8 @@ Widget.prototype.getStateQualifier = function(name) {
 Make a fake widget with specified variables, suitable for variable lookup in filters
 */
 Widget.prototype.makeFakeWidgetWithVariables = function(variables) {
-	var self = this;
+	var self = this,
+		variables = variables || {};
 	return {
 		getVariable: function(name,opts) {
 			if($tw.utils.hop(variables,name)) {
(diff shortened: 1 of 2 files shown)


saqimtiaz:
@Jermolene I think this is good to merge and it will be good to get some real world testing in the pre-release.

Jermolene:
Thanks @saqimtiaz

PR #7661, opened 2023-08-03
20230803 - minor edits

catter-fly:
Added thumbnail macro parameters, as well as some examples thereabouts.

Did some link-dusting via the Missing tab.

PR #8238, opened 2024-06-06
Docs tweaks (originally #7661)

Jermolene:
This is an attempt to salvage the content from #7661. For reasons that I don't understand, when I merged that PR the entire master branch was merged, and so I had to revert.


Jermolene:
Since I cannot merge these changes to "tiddlywiki-com", I am merging them to "master" on the basis that there will be a new release in a couple of weeks and so the changes will make their way to release then.

Jermolene:
OK I was able to merge these changes in 789d64f768a45d1874664e20c4a59f2db96f0125

PR #8239, opened 2024-06-07
Modifies action-log to return text of functions

### LINKED ISSUE #8189: [BUG] $action-log always evaluates all functions in its context
### Describe the bug

When functions are defined in the context of an ActionLogWidget (usually globally defined functions), these are always evaluated, even if they are not used in the output. Also, their filter runs are given all[tiddlers] as input, so they might need much more computation time that they were designed for.

### Expected behavior

They should not be evaluated in order to keep things running fast.

### To Reproduce

1. Go to tiddlywiki.com
2. Create a tiddler tagged `$:/tags/Macro` with this content
```
\function .slow() [range[1000]] :map[all[tiddlers]search:*[TiddlyWiki]] +[first[]]
```
3. Create another tiddler with this content
```
<$button actions="""<$action-log message="Button pressed." />""">GO</$button>
```
4. Open browser console and press button
5. Wait

### Screenshots

_No response_

### TiddlyWiki Configuration

No configuration dependence.

### Additional context

_No response_

saqimtiaz:
This PR modifies the action-log widget to return the raw text of functions instead of evaluating them, which can at times be exceedingly slow depending on how the function was written.

Fixes #8189

### DIFF
--- a/core/modules/widgets/action-log.js
+++ b/core/modules/widgets/action-log.js
@@ -66,7 +66,12 @@ LogWidget.prototype.log = function() {
 	});
 
 	for(var v in this.variables) {
-		allVars[v] = this.getVariable(v,{defaultValue:""});
+		var variable = this.parentWidget && this.parentWidget.variables[v];
+		if(variable && variable.isFunctionDefinition) {
+			allVars[v] = variable.value;
+		} else {
+			allVars[v]  = this.getVariable(v,{defaultValue:""});
+		}
 	}
 	if(this.filter) {
 		filteredVars = this.wiki.compileFilter(this.filter).call(this.wiki,this.wiki.makeTiddlerIterator(allVars));


pmario:
@saqimtiaz -- What if we log both values, to be sure we cover every usecase.

pmario:
I did not read the initial issue. -- But action-log is a debugging feature, so if it is slow that does not really matter, if it is what the user needs. 

So if evaluating is needed, it should be possible. May be the action-log widget can get a new parameter. -- I did not look at the code.

saqimtiaz:
Hi @pmario. I considered the value of being able to log the evaluated value of a function. However, logging both would still result in the same performance issues.

Also, one can still log the value of a function as follows:

`<$action-log functionValue=funcName/>`

saqimtiaz:
Hi @pmario. I considered the value of being able to log the evaluated value of a function. However, logging both would still result in the same performance issues.

Also, one can still log the value of a function as follows:

`<$action-log functionValue=funcName/>`

pmario:
>Also, one can still log the value of a function as follows:
>
>`<$action-log functionValue=<<funcName>>/>`

That's right, but the very first thing I do is log everything, to see the context of all variables.

I personally prefer the evaluated value as default, because at the start of the code there is a function definition that does contain the code of the function. I do not need this info. I need the evaluated info for debugging.

saqimtiaz:
I am open to a different mode for $action-log that does not evaluate variables (functions or macros). 

However, it feels rather user unfriendly that the default mode be one that does evaluate variables and thereby can be exceedingly slow in wikis that have a lot of functions defined (globally or in the ancestor tree of the widget). Given that our debugging tools are already not very user friendly, the default should definitely not be one that risks a wiki getting stuck in a very long loop.

Thoughts @Jermolene?

pmario:
I have no problem if the default is fast. I would like to have the option, that returns a value that I need, instead a value that I know already.

Jermolene:
Thanks @saqimtiaz – I agree that the current behaviour is problematic, and I think this is a reasonable fix.

PR #8240, opened 2024-06-07
fix: fixes regression introduced in #3704

saqimtiaz:
Fixes a regression introduced in #3704 whereby the server is initialized too early.


Jermolene:
Thanks @saqimtiaz sorry I missed this. I ended up fixing it by reverting the PR in question; on closer examination it was unfinished and not particularly useful in that state.

PR #8243, opened 2024-06-08
Use the testcase widget for the $action-deletefield widget examples

btheado:
Translate the `$action-deletefield` examples to use the `$testcase` widget.

Directly used the `$testcase` widget rather than test case tiddlers so the user can interactively click the buttons to see the behavior.


btheado:
Direct [link](https://tiddlywiki5-git-fork-btheado-docs-us-9b9439-jermolenes-projects.vercel.app/#ActionDeleteFieldWidget) to the modified tiddler.

Jermolene:
Thanks @btheado

PR #8246, opened 2024-06-09
Link to correct plugin instructions for Node.js

rybak:
If a user of a Node.js (aka "client-server") installation of TiddlyWiki5
goes through instructions of "Installing a plugin from the plugin
library" [1], they will encounter a dead end, quote:

	The official plugin library is disabled when using the
	client-server configuration. Instead, plugins should be
	installed via the `tiddlywiki.info` file, as described here.

The word "here" links to the same tiddler [1], sending the user into a
endless loop.

Instead, link to "Installing official plugins on Node.js" [2], which
actually contains the instructions for editing `tiddlywiki.info` file.

[1] https://tiddlywiki.com/#Installing%20a%20plugin%20from%20the%20plugin%20library
[2] https://tiddlywiki.com/#Installing%20official%20plugins%20on%20Node.js

### DIFF
--- a/plugins/tiddlywiki/tiddlyweb/language/en-GB/Translations.multids
+++ b/plugins/tiddlywiki/tiddlyweb/language/en-GB/Translations.multids
@@ -1,6 +1,6 @@
 title: $:/plugins/tiddlywiki/tiddlyweb/language/en-GB/
 
-ConfigOfficialPluginLibrary: The official plugin library is disabled when using the client-server configuration. Instead, plugins should be installed via the `tiddlywiki.info` file, as described [[here|https://tiddlywiki.com/#Installing%20a%20plugin%20from%20the%20plugin%20library]].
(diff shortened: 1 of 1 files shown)


Jermolene:
Thank you @rybak that's great. Updates that only affect the docs in `editions/tw5.com` and `editions/dev` are usually  merged to the "tiddlywiki-com" branch which makes them immediately available on tiddlywiki.com. Changes made to "master" are built to https://tiddlywiki.com/prerelease.

If you're able to retarget this PR to the "tiddlywiki-com" branch that would be great, but we're close enough to the release of v5.3.4 that there would be little impact to merging this to "master".

rybak:
> If you're able to retarget this PR to the "tiddlywiki-com" branch that would be great, but we're close enough to the release of v5.3.4 that there would be little impact to merging this to "master".

Tried it, but there's a merge conflict due to changes in commit 78ace9968 ("Feat: translatable plugins and docs (#7821)", 2024-06-08). It's not a problem at all for me to resolve the conflict, but then the opposite conflict would appear when merging "tiddlywiki-com" down to "master". So let's just leave the PR as is, targeting "master".

Jermolene:
No problem, thank you @rybak

PR #8247, opened 2024-06-10
Improve docs of transcludes and backtranscludes operators

mateuszwilczek:
Improve docs of transcludes and backtranscludes operators:

* mention that the filters work only on hard transclusions
* create a concept tiddler about hard and soft transclusions
* update "Release 5.3.4" to also include the transcludes operator


Jermolene:
Great thank you @mateuszwilczek

PR #8256, opened 2024-06-12
Fix 8177 add action tiddler list to droppable

### LINKED ISSUE #8177: [IDEA] It is proposted to add a "listactions" parameter to the droppable widget which acts different to "actions"
There is a problem if a tag-pill with **hundreds of titles** is dropped onto a user defined droppable-widget which in turn uses the **action-confirm-widget**. The confirm dialogue will pop up hundreds of times, with no possibility to stop it. 

### Proposal

It is proposed to add a new parameter eg: `listactions` to the droppable-widget. The connected action should get a variable eg: `actionTiddlerList` which contains a TW TitleList 

- The existing **actions** parameter will still execute 1 action **per tiddler**
- The new **listactions** parameter will only execute **1 action for the whole TitleList**

This makes it possible to check the number of titles, that will be dropped with a filter eg: `[<actionTiddlerList>split[ ]count[]]`. If the counter is too high, the user can decide how to handle the long list. 

- The idea contains a new "action variable" `actionTiddlerList`
- The existing [actionTiddler](https://tiddlywiki.com/#actionTiddler%20Variable) variable has 2 different behaviours. 
  - I think we should change that in a different PR. 

@Jermolene -- What do you think?

PS: Related post at Talk: https://talk.tiddlywiki.org/t/drag-and-drop-tag-pills-is-dangerous-tag-pills-can-have-big-drag-payloads/9768
PPS: The code change is simple. 9 lines of new code.

pmario:
This PR closes #8177

- #8177 

It extends the DroppableWidget with:

- New parameter `listActions`
- A new action-variable: actionTiddlerList which contains a TitleList instead of a single tiddler title
- It activates the listActions only once

@AnthonyMuscio -- please test the preview

There is a new example No. 3 at: https://tiddlywiki5-git-fork-pmario-fix-8177-14d6a0-jermolenes-projects.vercel.app/#DroppableWidget which can be used for testing


pmario:
There is a new example No. 3 at: https://tiddlywiki5-git-fork-pmario-fix-8177-14d6a0-jermolenes-projects.vercel.app/#DroppableWidget which can be used for testing

Jermolene:
Thanks @pmario

